PR #609, opened 2022-01-26
[mode] removed math operations and integer comparations

### LINKED ISSUE #520: Refactor mode selection
Right now, selecting the language mode is based on a number which can change based on which frontends are shipped with the current release. Also, it doesn't help that in some parts we are doing math operations e.g ``mode += 2` to select the mode.

I suggest creating an enum class for the modes and create a HashMap<mode,bool> to represent if the language is built.

rafaelsamenezes:
This should close #520, as it was causing problems on other frontends.

### DIFF
--- a/src/langapi/mode.cpp
+++ b/src/langapi/mode.cpp
@@ -30,6 +30,17 @@ int get_mode(const std::string &str)
   return -1;
 }
 
+int get_old_frontend_mode(int current_mode)
+{
+  unsigned i;
+  auto expected = mode_table[current_mode++].name;
+  for(i = current_mode; mode_table[i].name != nullptr; i++)
+    if(expected == mode_table[i].name)
+      return i;
+
+  return -1;
+}
+
 int get_mode_filename(const std::string &filename)
 {
   const char *ext = strrchr(filename.c_str(), '.');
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

kunjsong01 (approved):

src/langapi/mode.cpp
```diff
@@ -30,6 +30,17 @@ int get_mode(const std::string &str)
   return -1;
 }
 
+int get_old_frontend_mode(int current_mode)
+{
+  unsigned i;
+  auto expected = mode_table[current_mode++].name;
+  for(i = current_mode; mode_table[i].name != nullptr; i++)
+    if(expected == mode_table[i].name)
```
fbrausse, line 38: This comparison on `const char *` relies on string literals getting the same address in the translation unit `src/esbmc/globals.cpp`. This is [not guaranteed](https://en.cppreference.com/w/cpp/language/string_literal):
> Whether string literals can overlap and whether successive evaluations of a string-literal yield the same object is unspecified. That means that identical string literals may or may not compare equal when compared by pointer.

rafaelsamenezes: Nice catch. I will use `std::string` then.

Issue #133, opened 2020-02-04
ESBMC can't find bounds access violation in flexible arrays


mikhailramalho:
```
#include<stdlib.h>
#include<assert.h>

struct A {
  char alloc;
  int b[];
} a;

int main() {
  struct A *e = calloc(1, sizeof(struct A) + 3*sizeof(int));
  (*e).b[3] = a.b[0];
  assert((*e).b[3] == 0);
  free(e);
}
```
The flexible array `b` in `struct A` has size 3, but we are accessing `(*e).b[3]` and ESBMC doesn't find the violation. There is no violation in `a.b[0]`.

lucasccordeiro:
Verified fixed on master (`2dd70f7ce0`, ESBMC 8.4.0) during an audit of open issues.

**Reported:** ESBMC did not find the out-of-bounds access at `(*e).b[3]` in the flexible array member.

**On master now:**

```
$ esbmc main.c --force-malloc-success
Violated property:
  file main.c line 11 column 3 function main
  dereference failure: array bounds violated: heap object
  CWE: CWE-122, CWE-125, CWE-129, CWE-131, CWE-193, CWE-787
VERIFICATION FAILED
```
Line 11 is `(*e).b[3] = a.b[0];` — the exact access the report names. (`--force-malloc-success` is needed only so the `calloc` NULL-return check does not fire first.)

Closing. A regression test pinning this behaviour will follow in a separate PR.

PR #616, opened 2022-01-29
Changes in incomplete_array behaviour

rafaelsamenezes:
I am opening this as a draft for more discussion.

@mikhailramalho has suggested me to look into the `incomplete_array` from CBMC. It seems that they actually removed it and use it as an zero-sized array.

Using this approach fixed most fam-related bugs that I was having, even using the fam as an array (`fam_false_6`). Now the only problem is:

- Bounds-check for dereference to fam (e.g. `fam_false_0`, `fam_false_4`).

I am still doing some cleanup on the code so its not ready for review.

@mikhailramalho, @lucasccordeiro and @fbrausse let me know what you think. If this is ok, then I will remove the incomplete_array type.  Also, I would love some testcases as it is hard to find good examples of FAM.

### DIFF
--- a/regression/esbmc/fam_true_8/main.c
+++ b/regression/esbmc/fam_true_8/main.c
@@ -1,0 +1,60 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct S {
+  int x;
+  int y[];
+};
+
+void dynamic_fam_test_1(int x) {
+  // Allocate FAM
+  struct S *s = (struct S*) malloc(sizeof(int) + sizeof(int) * x);
+
+  // This is probably checked by clang itself!
+  assert(sizeof(s) == 8);
+  assert(sizeof(s->x) == 4);
+  assert(sizeof(*s) == 4);
+
+  s->x = x;
+
+  // Initialize FAM with i
+  for(int i = 0; i < s->x; i++)
+    s->y[i] = i;
+
+  // Check if FAM was actually initialized (Check #1)
+  for(int i = 0; i < s->x; i++)
+    assert(s->y[i] == i);
+
+  free(s);
+}
+
+void dynamic_fam_test_2()
+{
+  int y[] = {0, 1, 2, 3};
+  int x = sizeof(y)/sizeof(y[0]);
+  assert(x == 4);
+
+  // Allocate FAM
+  struct S *s = (struct S*) malloc(sizeof(int) * (x + 1));
+
+  // Initialize it with Y
+
+  // Wrong test
+  memcpy(s->y, y, sizeof(s->y[0]) * x);
+
+  // Check
+  for(int i = 0; i < x; i++)
+    assert(s->y[i] == i);
+
+  memset(s->y, 0, sizeof(s->y[0]) * x);
+  for(int i = 0; i < x; i++)
+    assert(!s->y[i]);
+
+  free(s);
+}
+
+int main() {
+  dynamic_fam_test_1(3);
+  dynamic_fam_test_2();
+}
--- a/regression/esbmc/fam_true_8/main.c
+++ b/regression/esbmc/fam_true_8/main.c
@@ -22,7 +22,7 @@ void dynamic_fam_test_1(int x) {
   for(int i = 0; i < s->x; i++)
     s->y[i] = i;
 
-  // Check if FAM was actually initialized (Check #1)
+  // Check if FAM was actually initialized
   for(int i = 0; i < s->x; i++)
     assert(s->y[i] == i);
 
@@ -39,14 +39,13 @@ void dynamic_fam_test_2()
   struct S *s = (struct S*) malloc(sizeof(int) * (x + 1));
 
   // Initialize it with Y
-
-  // Wrong test
   memcpy(s->y, y, sizeof(s->y[0]) * x);
 
-  // Check
+  // Check if 's->y' is initialized with 'y'
   for(int i = 0; i < x; i++)
     assert(s->y[i] == i);
 
+  // Resets 's->y'
   memset(s->y, 0, sizeof(s->y[0]) * x);
   for(int i = 0; i < x; i++)
     assert(!s->y[i]);
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -873,7 +873,7 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
       return true;
 
     new_type = array_typet(sub_type, gen_zero(index_type()));
-    new_type.set("incomplete", "true");
+    new_type.set("incomplete", true);
     break;
   }
 
@@ -1891,16 +1891,16 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
     // If this is a struct, we might be dealing with a fam!
     if(
-      t.is_struct() && to_struct_union_type(t).components().back().type().get(
-                         "incomplete") == "true")
+      t.is_struct() && to_struct_union_type(t).components().back().type().get_bool(
+                         "incomplete"))
     {
       // We should update the size of the incomplete array to the size of the last init!
-      typet fam_member;
+      typet fam_member_type;
       if(get_type(
            (*init_stmt.getInit(init_stmt.getNumInits() - 1)).getType(),
-           fam_member))
+           fam_member_type))
         return true;
-      to_struct_union_type(t).components().back().set("type", fam_member);
+      to_struct_union_type(t).components().back().type() = fam_member_type;
     }
 
     exprt inits;
@@ -1928,7 +1928,7 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
           return true;
 
         typet elem_type;
-        if(t.is_struct() || t.is_union())
+        if(t.is_struct())
           elem_type = to_struct_union_type(t).components()[i].type();
 
         else if(t.is_array())
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1915,19 +1915,17 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
           return true;
 
         typet elem_type;
-        bool do_typecast = true;
         if(t.is_struct() || t.is_union())
           elem_type = to_struct_union_type(t).components()[i].type();
+
         else if(t.is_array())
-        {
           elem_type = to_array_type(t).subtype();
-          do_typecast = t.get("incomplete") != "true";
-        }
+
         else
           elem_type = to_vector_type(t).subtype();
 
-        if(do_typecast)
-          gen_typecast(ns, init, elem_type);
+        gen_typecast(ns, init, elem_type);
+
         inits.operands().at(i) = init;
       }
     }
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1889,8 +1889,17 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     if(get_type(init_stmt.getType(), t))
       return true;
 
-    exprt inits;
+    // If this is a struct, we might be dealing with a fam!
+    if(t.is_struct() && to_struct_union_type(t).components().back().type().get("incomplete") == "true")
+      {
+        // We should update the size of the incomplete array to the size of the last init!
+        typet fam_member;
+        if(get_type((*init_stmt.getInit(init_stmt.getNumInits()-1)).getType(), fam_member))
+          return true;
+        to_struct_union_type(t).components().back().set("type", fam_member);
+      }
 
+    exprt inits;
     // Structs/unions/arrays put the initializer on operands
     if(t.is_struct() || t.is_array() || t.is_vector())
     {
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -872,7 +872,8 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
     if(get_type(arr.getElementType(), sub_type))
       return true;
 
-    new_type = array_typet(sub_type, gen_one(index_type()));
+    new_type = array_typet(sub_type, gen_zero(index_type()));
+    new_type.set("incomplete", "true");
     break;
   }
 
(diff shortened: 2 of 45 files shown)


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1616,7 +1617,9 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
         else
           elem_type = to_array_type(t).subtype();
 
-        gen_typecast(ns, init, elem_type);
+        if(elem_type.get("incomplete") != "true")
+          gen_typecast(ns, init, elem_type);
```
mikhailramalho: I think we need to remove the incomplete flag and generate the cast as-is. Does it break esbmc somehow?

src/goto-programs/goto_check.cpp
```diff
@@ -347,7 +347,10 @@ void goto_checkt::bounds_check(
   assert(is_array_type(t) || is_string_type(t));
 
   // We can't check the upper bound of an infinite sized array
-  if(is_array_type(t) && to_array_type(t).size_is_infinite)
+  // or of FAMs
+  if(
+    is_array_type(t) &&
+    (to_array_type(t).size_is_infinite || to_array_type(t).fam()))
```
mikhailramalho: Can we change it to `!to_array_type(t).size` and remove the `fam()` call?

src/irep2/irep2_type.cpp
```diff
@@ -160,6 +160,18 @@ unsigned int array_type2t::get_width() const
...
+{
+  try
+  {
+    return get_width() == 0;
+  }
+  catch(...)
+  {
+    return true;
+  }
+}
```
mikhailramalho: I don't think the array has enough information to say it's a fam, it might be a function parameter, right? Can we remove this method?

src/pointer-analysis/dereference.cpp
```diff
@@ -1129,6 +1129,20 @@ void dereferencet::construct_from_const_struct_offset(
...
+        // This access is in the bounds of this member, but isn't at the start.
+        // XXX that might be an alignment error.
+        expr2tc memb = member2tc(it, value, struct_type.member_names[i]);
+        constant_int2tc new_offs(pointer_type2(), int_offset - m_offs);
+
+        // Extract.
+        build_reference_rec(memb, new_offs, type, guard, mode);
+        value = memb;
+        return;
+      }
```
mikhailramalho, line 1198: You probably need to check the array subtype to generate out-of-alignment access here. I'm OK with leaving it as a TODO given that alignment checks are not enabled by default anymore.

src/pointer-analysis/dereference.cpp
```diff
@@ -1259,6 +1273,12 @@ void dereferencet::construct_from_dyn_struct_offset(
     expr2tc lower_bound = greaterthanequal2tc(bits_offset, field_offs);
     expr2tc upper_bound = lessthan2tc(bits_offset, field_top);
     expr2tc field_guard = and2tc(lower_bound, upper_bound);
+    // This breaks FAM
+    if(is_array_type(it) && to_array_type(it).fam())
+    {
+      msg.warning("Dynamic index for FAM member detected. Upper bound will not be verified...");
+      field_guard = lower_bound;
+    }
```
mikhailramalho, line 1347: This one should be easier tho... is the lower bound the upper_bound of the previous member/pad?

src/pointer-analysis/dereference.cpp
```diff
@@ -1908,7 +1928,8 @@ expr2tc dereferencet::stitch_together_from_byte_array(
   const expr2tc &offset,
   const guardt & /* guard */)
 {
-  assert(num_bytes != 0);
+  // Corner case, 0 bytes can mean that we are dealing with a FAM
+  num_bytes = num_bytes ? num_bytes : 1;
```
mikhailramalho: dreadful, should we simply return here? `num_bytes` will probably be an expr in the future.

src/solvers/smt/smt_bitcast.cpp
```diff
@@ -197,7 +197,8 @@ smt_astt smt_convt::convert_bitcast(const expr2tc &expr)
       type2tc subtype = arr_type.subtype;
 
       // We shouldn't have any bit left behind
-      assert(new_from->type->get_width() % subtype->get_width() == 0);
+      // This will not work for FAMs!
+      //assert(new_from->type->get_width() % subtype->get_width() == 0);
```
mikhailramalho: D:

mikhailramalho (commented):
> @mikhailramalho has suggested me to look into the incomplete_array from CBMC. It seems that they actually removed it and use it as a zero-sized array.

Nice, that means we can save a type here.

Can you add this test case also, it uses dynamic allocation (should also try it with alloca?):
```
#include <assert.h>
#include <stdlib.h>
#include <string.h>

struct S {
  int x;
  int y[];
};

void foo(int x) {
  struct S *s = (struct S*) malloc(sizeof(int) + sizeof(int) * x);
  assert(sizeof(s) == 8);
  assert(sizeof(s->x) == 4);
  assert(sizeof(*s) == 4);

  s->x = x;
  for(int i = 0; i < s->x; i++)
    s->y[i] = i;
  for(int i = 0; i < s->x; i++)
    assert(s->y[i] == i);
  
  free(s);
}

void foo1()
{
  
  int y[] = {0, 1, 2, 3};
  int x = sizeof(y)/sizeof(y[0]);

  struct S *s = (struct S*) malloc(sizeof(int) * (x + 1));
  memcpy(s->y, y, sizeof(s->y[0]) * x);
  for(int i = 0; i < x; i++)
    assert(s->y[i] == i);

  memset(s->y, 0, sizeof(s->y[0]) * x);
  for(int i = 0; i < x; i++)
    assert(!s->y[i]);
  
  free(s);
}

int main() {
  foo(3);
  foo1();
}
```

src/clang-c-frontend/clang_c_convert.cpp
rafaelsamenezes: Generating this cast causes issues down the road (at smt_conv). I still haven't look into it though.

src/goto-programs/goto_check.cpp
rafaelsamenezes: Yes

src/pointer-analysis/dereference.cpp
rafaelsamenezes: This was reverted

src/pointer-analysis/dereference.cpp
rafaelsamenezes: I added the TODO

rafaelsamenezes:
I also just started a benchexec run for this PR.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1871,13 +1885,17 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
           return true;
 
         typet elem_type;
-        if(t.is_struct())
+        if(t.is_struct() || t.is_union())
```
fbrausse: `t` being a union cannot happen in this branch. Unions are handled outside.

regression/esbmc/fam_true_8/main.c
```diff
@@ -0,0 +1,60 @@
...
+  int x = sizeof(y)/sizeof(y[0]);
+  assert(x == 4);
+
+  // Allocate FAM
+  struct S *s = (struct S*) malloc(sizeof(int) * (x + 1));
+
+  // Initialize it with Y
+
+  // Wrong test
+  memcpy(s->y, y, sizeof(s->y[0]) * x);
```
fbrausse, line 42: What does "wrong test" mean here? Is `memcpy` considered a test? Why is it wrong?

src/pointer-analysis/dereference.cpp
```diff
@@ -1078,6 +1076,10 @@ void dereferencet::construct_from_array(
     unsigned int num_bytes = compute_num_bytes_to_extract(
       replaced_dyn_offset, type_byte_size_bits(type).to_uint64());
 
+    if(!num_bytes)
+      num_bytes = compute_num_bytes_to_extract(
+        offset, type_byte_size_bits(value->type).to_uint64());
```
fbrausse: Why is it OK to pass a possibly non-constant `offset` here in contrast to the `replaced_dyn_offset` above, which should be constant (after simplification) in more cases that `offset`? I fail to see what dynamic offsets have to do with FAMs.

src/pointer-analysis/dereference.cpp
```diff
@@ -1974,6 +2011,9 @@ expr2tc dereferencet::stitch_together_from_byte_array(
   simplify(offset_bytes);
 
   BigInt num_bits = type_byte_size_bits(type);
+  // If the destiny is a zero-sized array then we can just return anything
```
fbrausse: "destination", not "destiny" :)

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -865,7 +865,8 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
     if(get_type(arr.getElementType(), sub_type))
       return true;
 
-    new_type = array_typet(sub_type, gen_one(index_type()));
+    new_type = array_typet(sub_type, gen_zero(index_type()));
+    new_type.set("incomplete", "true");
```
mikhailramalho: why an incomplete array has size one?

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1845,8 +1846,21 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+      t.is_struct() && to_struct_union_type(t).components().back().type().get(
+                         "incomplete") == "true")
+    {
+      // We should update the size of the incomplete array to the size of the last init!
+      typet fam_member;
+      if(get_type(
+           (*init_stmt.getInit(init_stmt.getNumInits() - 1)).getType(),
+           fam_member))
+        return true;
+      to_struct_union_type(t).components().back().set("type", fam_member);
```
mikhailramalho: you can probably do back().type() = fam_member;

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1845,8 +1846,21 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
       return true;
 
-    exprt inits;
+    // If this is a struct, we might be dealing with a fam!
+    if(
+      t.is_struct() && to_struct_union_type(t).components().back().type().get(
+                         "incomplete") == "true")
+    {
+      // We should update the size of the incomplete array to the size of the last init!
+      typet fam_member;
```
mikhailramalho: fam_member_type?

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: this should be `new_type.set("incomplete", true);`

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1845,8 +1846,21 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     if(get_type(init_stmt.getType(), t))
       return true;
 
-    exprt inits;
+    // If this is a struct, we might be dealing with a fam!
+    if(
+      t.is_struct() && to_struct_union_type(t).components().back().type().get(
+                         "incomplete") == "true")
```
mikhailramalho: and this should be `type().get_bool("incomplete");`

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1871,13 +1885,17 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
           return true;
 
         typet elem_type;
-        if(t.is_struct())
+        if(t.is_struct() || t.is_union())
           elem_type = to_struct_union_type(t).components()[i].type();
```
mikhailramalho, line 1932: remove added whitespaces below

src/goto-programs/goto_check.cpp
```diff
@@ -342,16 +343,39 @@ void goto_checkt::bounds_check(
 
   assert(is_array_type(t) || is_string_type(t) || is_vector_type(t));
 
-  // We can't check the upper bound of an infinite sized array
```
mikhailramalho, line 550: I suggest keeping the early return for an infinite-sized array and adding another one for an incomplete array.

src/goto-programs/goto_check.cpp
```diff
@@ -342,16 +343,39 @@ void goto_checkt::bounds_check(
...
     is_array_type(t) ? to_array_type(t).array_size
     : is_vector_type(t)
       ? to_vector_type(t).array_size
       : constant_int2tc(get_uint32_type(), to_string_type(t).get_length());
 
+  // Are we dealing with infinity/incomplete arrays?
+  if(
+    is_array_type(t) &&
+    (to_array_type(t).size_is_infinite ||
+     (to_constant_int2t(to_array_type(t).array_size).value == 0)))
```
mikhailramalho: can we have an array member with size 0 inside a struct, that is not a fam?

mikhailramalho (changes requested):
It looks like we can't get away with not having an incomplete array after all.

My main concern is that we are using an array of size zero to represent it in irep2, I'm not sure this will break other code in esbmc. Maybe we need an "incomplete" flag like infinite size?

src/goto-programs/goto_check.cpp
rafaelsamenezes: There are zero-length arrays: https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

I didn't look too deep into it but the main difference seems to be that FAM allows static initialization of the array while zero-length does not.

src/pointer-analysis/dereference.cpp
rafaelsamenezes: Nice catch! This was some leftover of the previous approach where we would not update the array size at the frontend. I will remove it.

src/clang-c-frontend/clang_c_convert.cpp
rafaelsamenezes: I am not sure what you mean here. Are you referring to the whitespaces between "if" and "else"?

src/clang-c-frontend/clang_c_convert.cpp
rafaelsamenezes: It is zero. The previous implementation used to go with one.

regression/esbmc/fam_true_8/main.c
rafaelsamenezes: Ow, another leftover. Not sure what I was thinking when I wrote this though.

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: yes

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1845,8 +1846,21 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     if(get_type(init_stmt.getType(), t))
       return true;
 
-    exprt inits;
```
mikhailramalho, line 1891: can this be done in adjust?

src/pointer-analysis/dereference.cpp
```diff
@@ -1974,6 +2007,9 @@ expr2tc dereferencet::stitch_together_from_byte_array(
   simplify(offset_bytes);
 
   BigInt num_bits = type_byte_size_bits(type);
+  // If the destination is a zero-sized array then we can just return anything
+  if(!num_bits.compare(0))
```
mikhailramalho, line 2017: ` num_bits != 0`?

mikhailramalho (approved):
Small nits, otherwise LGTM.

mikhailramalho:

This PR has been open since 2023-05-16 with no activity. FAM / incomplete-array handling has changed substantially since — master now has fam_false_0..5 and fam_true_0..3 regression tests, and the dereference / SMT pipeline has been reworked. Issue #133 ("ESBMC can't find bounds access violation in flexible arrays") tracks remaining gaps.

Issue #204, opened 2020-06-11
Add type assertions when creating irep2s


mikhailramalho:
We are currently generating a number of `irep`/`irep2` expr's where the operands have different types. The SMT solvers don't like it, and even the C standard tries to avoid it by introducing type promotions.

The problem causes bugs like #148, where `get_ulong` is used to generate constants and will generate conflicting types in a number of operations depending on the arch (32 bits vs 64 bits).

We should add assertions to `irep2`'s constructors, so our regressions can catch these problems. Something like:
```
From c1a8540907402a890a4174e948ce9b580719c2e8 Mon Sep 17 00:00:00 2001
From: "Mikhail R. Gadelha" <mikhail.ramalho@gmail.com>
Date: Thu, 11 Jun 2020 19:07:47 -0400
Subject: [PATCH] Irep2 assertions

Signed-off-by: Mikhail R. Gadelha <mikhail.ramalho@gmail.com>
---
 src/util/irep2_expr.h | 18 ++++++++++++++++++
 1 file changed, 18 insertions(+)

diff --git a/src/util/irep2_expr.h b/src/util/irep2_expr.h
index fe0487421..837ba9619 100644
--- a/src/util/irep2_expr.h
+++ b/src/util/irep2_expr.h
@@ -307,6 +307,20 @@ public:
     const expr2tc &s2)
     : expr2t(t, id), side_1(s1), side_2(s2)
   {
+    if(is_pointer_type(s1) && is_pointer_type(s2))
+    {
+      assert(s1->type->get_width() == s2->type->get_width());
+    }
+    else
+    {
+      try
+      {
+        assert(s1->type == s2->type);
+      }
+      catch(array_type2t::inf_sized_array_excp *e)
+      {
+      }
+    }
   }
   relation_data(const relation_data &ref) = default;
 
@@ -456,6 +470,7 @@ public:
     const expr2tc &v2)
     : arith_ops(t, id), side_1(v1), side_2(v2)
   {
+    assert(v1->type == v2->type);
   }
   arith_2ops(const arith_2ops &ref) = default;
 
@@ -506,6 +521,7 @@ public:
     const expr2tc &rm)
     : arith_ops(t, id), rounding_mode(rm), side_1(v1), side_2(v2)
   {
+    assert(v1->type == v2->type);
   }
   ieee_arith_2ops(const ieee_arith_2ops &ref) = default;
 
@@ -539,6 +555,8 @@ public:
     const expr2tc &rm)
[...]

PR #619, opened 2022-01-31
irep2 consistency, part 1/n

fbrausse:
This PR introduces a few invariants in the construction of expr2t in line with #204:
- ~`constant_struct2t` checks consistency of provided member expressions with the struct type~ (*not yet*)
- `with2t` checks consistency of the updated value with its source, as well as that of the field expression
- `arith_ops2` enforces some relations between the resulting type and also its 2 operands (this required to derive `shl2t` and `ashr2t` from `bit_ops2` instead of `arith_ops2`)
- `member2t` enforces that the source is a struct or union and that it actually contains the member
- ~`array_type2t` enforces a non-nil `size` expression to be of unsigned type~ (*not yet*)

~This is still based on #615, for now, and not ready for review at least until that PR is dealt with.~


fbrausse:
- regression/esbmc/32_floppy fails a new consistency-assert because it multiplies a signed 32-bit array-size with an unsigned 64-bit size of the subtype in dereferencet::construct_struct_ref_from_dyn_offs_rec(). I don't see signed array sizes having any valid use case in expr2; is there any? Enforcing unsigned array sizes breaks in a lot of other regression tests though, would need to investigate. Should unsigned array sizes be part of this PR?
- regression/esbmc-unix/10_bicycle_04 fails on macOS because a value of type `struct timespec` is updated in symex-assign, but the type is broken/incomplete - it has no members. How can this happen?
  ```
  p lhs->dump()
  member
  * source_value : symbol
      * name : c:main.c@1964@F@cycling_thread@time_to_sleep
      * renamelev : Level 1
      * level1_num : 1
      * level2_num : 0
      * thread_num : 1
      * node_num : 0
      * type : struct
        * members : 
  
        * member_names : 
  
        * member_pretty_names : 
  
        * typename : struct timespec
        * packed : false
  * member_name : tv_sec
  * type : signedbv
    * width : 64
  ```
  however clang -E (with a [macOS-11.3 sysroot](https://github.com/phracker/MacOSX-SDKs.git)) produces the right definition which is not used for `time_to_sleep`:
  ```c
  # 33 "../../MacOSX-SDKs/MacOSX11.3.sdk/usr/include/sys/_types/_timespec.h" 3 4
  struct timespec
  {
   __darwin_time_t tv_sec;
   long tv_nsec;
  };
  ```

fbrausse:
> the type is broken/incomplete - it has no members. How can this happen? 

Never mind, got it: it happened because `struct timespec` was not declared globally before it was used in our <pthread.h>. Interesting, but consistent. Fixed.

lucasccordeiro:
> Should unsigned array sizes be part of this PR?

@fbrausse: could you please create another PR to deal with this issue?

fbrausse:
I have traced the incompatibly pre-processed regression/esbmc-unix/04_bzip2smp_01 back to a1d16931ae6a2351158cfcf14a9eb5bdbede07ce. Does anyone know what the original source of these "esmbc-pt" benchmarks was?

src/irep2/irep2_expr.cpp
```diff
@@ -350,6 +350,68 @@ expr2tc constant_string2t::to_array() const
   return final_val;
 }
 
+#ifndef NDEBUG
```
lucasccordeiro: Do we really need this? If so, can you add some comments for this?

src/irep2/irep2_expr.h
```diff
@@ -499,6 +499,29 @@ class arith_2ops : public arith_ops
     const expr2tc &v2)
     : arith_ops(t, id), side_1(v1), side_2(v2)
   {
+#ifndef NDEBUG
```
lucasccordeiro: The same here. I would recommend adding some comments.

src/irep2/irep2_expr.h
```diff
@@ -2807,6 +2832,9 @@ class with2t : public with_expr_methods
     const expr2tc &value)
     : with_expr_methods(type, with_id, source, field, value)
   {
+#if !defined(NDEBUG)
```
lucasccordeiro: Comments?

src/irep2/irep2_expr.h
```diff
@@ -2827,6 +2855,15 @@ class member2t : public member_expr_methods
   member2t(const type2tc &type, const expr2tc &source, const irep_idt &memb)
     : member_expr_methods(type, member_id, source, memb)
   {
+#ifndef NDEBUG
```
lucasccordeiro: Comments?

src/pointer-analysis/dereference.cpp
```diff
@@ -1262,7 +1262,7 @@ void dereferencet::construct_from_const_struct_offset(
       return;
     }
 
-    if(int_offset < (m_offs + m_size))
+    if(int_offset < m_offs + m_size)
```
lucasccordeiro: I would prefer the previous version for the readability's sake :-)

lucasccordeiro (approved):
Minor comments only.

src/irep2/irep2_expr.h
fbrausse: I would keep this in the `.h`, otherwise it'll slow down Release-builds unnecessarily.

src/irep2/irep2_expr.cpp
fbrausse: This one we don't need, I've removed it.

src/pointer-analysis/dereference.cpp
fbrausse: strange :) reverted.

src/irep2/irep2_expr.h
fbrausse: I've added a comment.

fbrausse:
SV-COMP'22 with a "Release" build (i.e. no assertions) on master (5cce74ed0381c0795ae14eda2742dfb93ea63536):
```
Statistics:          15645 Files
  correct:            6591
    correct true:     3548
    correct false:    3043
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9024
  Score:              9579 (max: 25561)
```
This PR (33258c5eb6d3b8ddc6c4c05ac56b29f2f872977f):
```
Statistics:          15645 Files
  correct:            6601
    correct true:     3551
    correct false:    3050
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9014
  Score:              9592 (max: 25561)
```

lucasccordeiro:
Many thanks, @fbrausse.

Issue #589, opened 2022-01-11
Dereference errors in arrays of structs with bit-fields


shmarovfedor:
I came across this issue recently:
```
#include <assert.h>

typedef struct str {
  unsigned short field1 : 4;
  unsigned short field2 : 4;
} str;

int main(void) {
  str array;
  str array_set[2];
  array_set[1] = array;
  assert(*(unsigned char *)&array == *(unsigned char *)&array_set[1]);
  return 0;
}
```
ESBMC returns `VERIFICATION FAILED` which should not be the case. However, when the size of each struct member is a multiple of 8 bits, the problem goes away.

fbrausse:
I believe this is related to item 2. in https://github.com/esbmc/esbmc/pull/590#issuecomment-1010309733: pointers to structs seem to be interpreted as pointers to the struct's first member.

fbrausse:
Fixed by #624.

Issue #575, opened 2021-12-22
Dereference error 


fbrausse:
This code
```c
#include <assert.h>
#include <stdlib.h>

typedef struct str {
    unsigned field1 : 15;
} str;

int main(void) {

  str array, *ptr = &array;
  int i = 0;
  ptr[i].field1 = i;
  unsigned n = ptr->field1;
  assert(n == i);

  return 0;
}
```
 is wrongly attributed a
```
Counterexample:

State 1 file fedor-bitfields-01_ptr_bf_1.c line 12 function main thread 0
----------------------------------------------------
  (ptr + (signed long int)i)->field1 = byte_update_little_endian(array.field1, 0, 0 | byte_extract_little_endian(array.field1, 0) & 128)

State 2 file fedor-bitfields-01_ptr_bf_1.c line 13 function main thread 0
----------------------------------------------------
  n = (unsigned int)array.field1

State 3 file fedor-bitfields-01_ptr_bf_1.c line 14 function main thread 0
----------------------------------------------------
Violated property:
  file fedor-bitfields-01_ptr_bf_1.c line 14 function main
  assertion n == i
  n == (unsigned int)i


VERIFICATION FAILED
```
Replacing `[i].` with either `[0].` or `->` fixes the issue, as does making the bitfield width a multiple of 8.

lucasccordeiro:
@shmarovfedor: Can you please take a look at this issue? It seems to be related to the bitfield support.

shmarovfedor:
Yea it looks like the case of bitfields dereferenced using a dynamic offset (introduced by `[i]`) is not handled correctly in the `dereference` module. I will be working on it.

fbrausse:
Fixed by #624.

Issue #448, opened 2021-05-27
ESBMC reports oversized field offset


lucasccordeiro:
ESBMC reports that this C program contains a deference failure, in particular, oversized field offset, which seems to be wrong. This issue requires further investigation.

@mikhailramalho and @rafaelsamenezes: do you have any thoughts?

```
#include <stdbool.h>
#include <stdint.h>


typedef struct str1_t {
    unsigned int field1 : 16;
    unsigned int field2 : 8;
    uint8_t field3[59];
} str1_t;


typedef struct str2_t {
    str1_t field1;
    uint32_t field2;
    uint64_t field3;
    uint32_t field4;
    uint32_t field5[4];
    uint32_t field6;
} str2_t;


int main(void) {

    str2_t var_data;
    str2_t *var_data_p = &var_data;
    uint64_t var_offset[1000]; // = 0xdeadbeef;
    
    for (uint16_t i=0; i < sizeof(str1_t)/sizeof(uint64_t); i++) {
        ((uint64_t *)&var_data_p->field1)[i] = ((uint64_t *)var_offset)[i];
    }
   return 0; 
}
```

rafaelsamenezes:
~I was not able to replicate this issue. Which branch and options are you using?~

This issue doesn't happen on the latest version of ESBMC. It has the following problem:

```
 ../src/pointer-analysis/dereference.cpp:1742: expr2tc* dereferencet::extract_bytes_from_array(const expr2tc&, unsigned int, const expr2tc&): Assertion `subtype->get_width() == 8 && "Can only extract bytes for stitching from byte array"' failed.
```

lucasccordeiro:
ucode branch 

Just run ESBMC file.c

lucasccordeiro:
@shmarovfedor: Should this issue be fixed by https://github.com/esbmc/esbmc/pull/590?

shmarovfedor:
@lucasccordeiro, I don't think so. I'm still getting the same error message as @rafaelsamenezes : 
```
Assertion failed: (subtype->get_width() == 8 && "Can only extract bytes for stitching from byte array"), function extract_bytes_from_array, file dereference.cpp, line 1799.
```
I will look into this issue.

lucasccordeiro:
@shmarovfedor: Can I ask you whether you have had a chance to look at this issue related to bitfield?

PR #624, opened 2022-02-07
Fixing recent issues with bit-fields

### LINKED ISSUE #552: ESBMC does not allow multiple unnamed bit-fields
Here is a simple example where a struct has multiple unnamed bitfields
```
#include <stdint.h>

int main(int argc, char **argv)
{
         union {
                 struct {
                         unsigned int s : 5;
                         unsigned short f : 7;
                         unsigned char : 2;
                         unsigned char : 4;
                         unsigned int c : 3;
                 };
                 uint32_t i;
         } v = { .c = 1 << 2 | 1, .s = 1 << 4 | 1, .f = 1 << 6 | 1 };
         return 0;
 }
```
and ESBMC returns
```
ESBMC version 6.8.0 64-bit x86_64 macos
file /Users/user/examples/bitfields/bitfields-6.c: Parsing
Converting
Generating GOTO Program
GOTO program creation time: 0.553s
Assertion failed: (0 && fmt::format( "Name \"{}\" matches more than one member\" in struct/union \"{}\"", comp, name) .c_str()), function get_component_number, file irep2.cpp, line 301.
Abort trap: 6
``` 

This is because of the naming convention ESBMC uses for unnamed members. Each of such members will be given the same name at parsing which results in failing at a further stage due to having multiple identifiers with the same name. This shouldn't be hard to fix as we will only need to make sure that all unnamed members get unique names in the ESBMC irep.

shmarovfedor:
This PR fixes #552, #575, #589 and some upcasting issues for arrays (i.e., correcting byte extracting). I will also address #448 in this PR.


lucasccordeiro:
@shmarovfedor: overall, this PR looks good to me.

Can I ask you whether you could run it over the SV-COMP 2022 benchmarks?

shmarovfedor:
@lucasccordeiro , yes I will do.

src/pointer-analysis/dereference.cpp
```diff
@@ -1381,8 +1384,10 @@ void dereferencet::construct_from_dyn_offset(
   if(!is_bv_type(value->type) && !is_fixedbv_type(value->type))
     value = bitcast2tc(get_uint_type(value->type->get_width()), value);
 
+  expr2tc replaced_dyn_offset = replace_dyn_offset_with_zero(offset);
+  simplify(replaced_dyn_offset);
   unsigned int num_bytes =
-    compute_num_bytes_to_extract(offset, type_byte_size_bits(type).to_uint64());
+    compute_num_bytes_to_extract(replaced_dyn_offset, type_byte_size_bits(type).to_uint64());
```
fbrausse: Why does `construct_from_dyn_offset()` compute #bytes to extract without taking symbolic offsets into account while `construct_from_array()`, `construct_from_const_offset()` do not? Are they allowed to and not doing it is an optimisation or are there special conditions for this particular case?

fbrausse: OK, `construct_from_const_offset()` seems clear - no symbols in constant expressions. But what about `construct_from_array()`?

shmarovfedor: Essentially the offset value is only important when we try to extract a value that is not aligned to a byte (e.g., suppose we have a `struct {unsigned field1 : 7; unsigned field2 : 10}`, and to correctly extract `field2` we need to extract 3 bytes as `field2` spans over 3 bytes of the struct). Otherwise, the total number of bytes is completely defined by its type size (i.e., `type_byte_size_bits`). Now when the offset is constant, we can directly pass it to `compute_num_bytes_to_extract`. The same cannot be done for dynamic offsets, but I think we can make some reasonable assumptions in this case. Since the symbolic part of the `dynamic_offset` cannot encode an address of a bit-field, I think we can safely assume that it is always aligned to a byte. So when we are dealing with a `dynamic_offset` to a bit-field we can replace the symbolic part with 0, thus obtaining the constant offset to the field within the inner-most struct. And this is all that's required to correctly calculate the number of bytes occupied by a bit-field member. Later when the actual byte extracting happens, we will use the original `dynamic_offset`. Also I think you are right that I overlooked the `construct_from_array()` method which can indeed have a dynamic offset as an input (I just did not come up with an example to test it).

shmarovfedor: Sorry just saw you later comment. Yes I agree I will need to do the same for `construct_from_array`

fbrausse: Many thanks for the explanations @shmarovfedor! Should we add this reasoning as a comment in the code?

shmarovfedor: Yes I think that would be very useful.

src/pointer-analysis/dereference.cpp
```diff
@@ -72,6 +72,39 @@ static inline expr2tc get_base_dereference(const expr2tc &e)
...
+      replace_dyn_offset_with_zero(to_sub2t(e).side_2));
+
+  if(is_mul2t(e))
+    return mul2tc(e->type, replace_dyn_offset_with_zero(to_mul2t(e).side_1),
+      replace_dyn_offset_with_zero(to_mul2t(e).side_2));
+
+  if(is_div2t(e))
+    return div2tc(e->type, replace_dyn_offset_with_zero(to_div2t(e).side_1),
+      replace_dyn_offset_with_zero(to_div2t(e).side_2));
+
```
mikhailramalho, line 114: You can also use `e->get_sub_expr(N)` to get the (n + 1)th sub expr, so the guards can be unified.

The current version is better for readability tho.

shmarovfedor:
The following are the results for SV-COMP:

master:
```
Statistics:          15457 Files
  correct:            6239
    correct true:     2992
    correct false:    3247
  incorrect:            28
    incorrect true:      4
    incorrect false:    24
  unknown:            9190
  Score:              8719 (max: 25029)
```
This PR:
```
Statistics:          15457 Files
  correct:            6241
    correct true:     2992
    correct false:    3249
  incorrect:            29
    incorrect true:      4
    incorrect false:    25
  unknown:            9187
  Score:              8705 (max: 25029)
```
This PR produces an additional incorrect outcome (see #633 for the reduced version of the failing benchmark), but resolves additional 2 benchmarks. At the same time, the master branch of ESBMC does not support these 3 benchmarks (aborting and returning `UNKNOWN`). In my opinion this PR can be merged into master, and issues like #633 and #448 can be resolved by another PR.

lucasccordeiro:
Thanks, @shmarovfedor. We agree with you. 

I'll merge this PR; if you can get the other issues fixed via other PRs, then that would be great.

PR #625, opened 2022-02-07
Restore old cpp frontend

kunjsong01:
Restore old cpp frontend to provide more examples for references.

### DIFF
--- a/src/ansi-c/c_main.cpp
+++ b/src/ansi-c/c_main.cpp
@@ -299,10 +299,12 @@ bool c_main(
 
   code_function_callt thread_start_call;
   thread_start_call.location() = symbol.location;
-  thread_start_call.function() = symbol_exprt("c:@F@pthread_start_main_hook");
+  thread_start_call.function() = symbol_exprt("c:@F@pthread_start_main_hook"); // DEBUG: fail
+  //thread_start_call.function() = symbol_exprt("pthread_start_main_hook"); // DEBUG: fixes
   code_function_callt thread_end_call;
   thread_end_call.location() = symbol.location;
-  thread_end_call.function() = symbol_exprt("c:@F@pthread_end_main_hook");
+  thread_end_call.function() = symbol_exprt("c:@F@pthread_end_main_hook"); // DEBUG: fail
+  //thread_end_call.function() = symbol_exprt("pthread_end_main_hook"); // DEBUG: fixes
 
   init_code.move_to_operands(thread_start_call);
   init_code.move_to_operands(call);
--- a/src/ansi-c/c_main.cpp
+++ b/src/ansi-c/c_main.cpp
@@ -297,14 +297,17 @@ bool c_main(
   // Call to main symbol is now in "call"; construct calls to thread library
   // hooks for main thread start and main thread end.
 
+  // For the old frontend, we just use pthread_start_main_hook instead of "c:@F@pthread_start_main_hook" in the old frontend.
+  // The latter is used by clang-c-frontend to verify multi-threaded programs and atomics.
+  // If we want to use the latter, we also need to deal with the symbols referred to from the body of c:@F@pthread_strat_main_hook,
+  // such as __ESBMC_num_threads_running, __ESBMC_num_total_threads, and __ESBMC_atomic_end .etc.
+  // This will become an independent task to add thread support. See clang_c_languaget::final function for example.
   code_function_callt thread_start_call;
   thread_start_call.location() = symbol.location;
-  thread_start_call.function() = symbol_exprt("c:@F@pthread_start_main_hook"); // DEBUG: fail
-  //thread_start_call.function() = symbol_exprt("pthread_start_main_hook"); // DEBUG: fixes
+  thread_start_call.function() = symbol_exprt("pthread_start_main_hook");
   code_function_callt thread_end_call;
   thread_end_call.location() = symbol.location;
-  thread_end_call.function() = symbol_exprt("c:@F@pthread_end_main_hook"); // DEBUG: fail
-  //thread_end_call.function() = symbol_exprt("pthread_end_main_hook"); // DEBUG: fixes
+  thread_end_call.function() = symbol_exprt("pthread_end_main_hook");
 
   init_code.move_to_operands(thread_start_call);
   init_code.move_to_operands(call);
--- a/src/clang-c-frontend/clang_c_language.cpp
+++ b/src/clang-c-frontend/clang_c_language.cpp
@@ -241,8 +241,6 @@ bool clang_c_languaget::typecheck(
 
   // Add symbols in new_context
   // The symbols here cover AST nodes from internals and source code
-  // DEBUG: new_context.symbols 0 -> 72
-  // DEBUG: context.symbols 0 -> 0
   clang_c_convertert converter(new_context, ASTs, msg);
   if(converter.convert())
     return true;
@@ -251,14 +249,10 @@ bool clang_c_languaget::typecheck(
   //  - overwrite macro type with real type
   //  - adjust side effect
   //  - add additional symbol e.g. c:@F@assert
-  // DEBUG: new_context.symbols 72 -> 73
-  // DEBUG: context.symbols 0 -> 0
   clang_c_adjust adjuster(new_context, msg);
   if(adjuster.adjust())
     return true;
 
-  // DEBUG: new_context.symbols: 73 -> 73
-  // DEBUG: context.symbols: 0 -> 73
   if(c_link(context, new_context, msg, module))
     return true;
 
@@ -285,10 +279,10 @@ bool clang_c_languaget::preprocess(
 
 bool clang_c_languaget::final(contextt &context, const messaget &msg)
 {
-  // This function adds the cprover builtin symbols and __ESBMC_main symbol.
+  // This function injects the cprover symbols, as well as the __ESBMC_main symbol.
   // The symbol table is updated following the steps below:
   //
-  // Step 1: add_cprover_library adds buildtin symbols:
+  // Step 1: add_cprover_library adds builtin symbols:
   //  - c:@__ESBMC_num_threads_running
   //  - c:builtin_libs.c@358@F@__ESBMC_sync_fetch_and_add@initial
   //  - c:builtin_libs.c@307@F@__ESBMC_sync_fetch_and_add@ptr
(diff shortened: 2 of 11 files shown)


mikhailramalho:
Heads up: I think @rafaelsamenezes is also changing the part that enables/disables the old frontends, I think you two should try to converge your efforts here.

kunjsong01:
With the help of the above patches, we are able to collect the logs `esbmc/regression/20220211_233115_esbmc-cpp_cpp`. 

We now get the big picture of the pass/fail rate and the error signatures for cpp old frontend to pass esbmc-cpp/cpp: 

To fully restore cpp olf frontend, we need to fix the following error signatures: 
```
VirtualBox$ egrep "Assertion|###ERROR:" * -rn | rev | cut -d ':' -f 1 | rev | sort | uniq -c
      5  Assertion `decl.get_statement() == "block"' failed.
      2  Assertion `endatptr != atstr.c_str()' failed.
     86  Assertion `type.is_not_nil()' failed.
     33  CONVERSION ERROR
    233  expression statement takes one operand
      7  Message from unrecognized k-induction child process
```

TODOs:
1. (This PR) Continue fixing the errors until we have 100% pass on `esbmc-cpp/cpp`. Start with "**expression statement takes one operand**", which causes 233 failures!
2. (Another PR in the future) Fully bring up `esbmc-cpp/cpp` benchmark with clang-cpp-frontend. Use the old cpp frontend flows to guide the development. 
3. (Another PR in the future) Add C++11 and/or C++14 benchmarks, and fully bring up these benchmarks with clang-cpp-frontend.

kunjsong01:
Update of 04/03/2022: 

54% passing rate in `esbmc-cpp/cpp`, 196 out of 366 tests in this regression suite is passing. 

The stats below shows the error signatures:

```
VirtualBox$ ls 20220304_133136_esbmc-cpp_cpp/* | wc -l
366
VirtualBox$ egrep 'VERIFICATION SUCCESSFUL|VERIFICATION FAILED' 20220304_133136_esbmc-cpp_cpp/ -rn | wc -l
196
VirtualBox$ egrep 'Assertion|###ERROR:' 20220304_133136_esbmc-cpp_cpp/ -rn | rev | cut -d ':' -f 1 | rev | sort | uniq -c
      5  Assertion `decl.get_statement() == "block"' failed.
      6  Assertion `endatptr != atstr.c_str()' failed.
      1  Assertion `is_constant_array2t(elem) && "Sub-member of constant array must be " "constant array"' failed.
     86  Assertion `type.is_not_nil()' failed.
      8  Assertion `type->type_id == trueval->type->type_id' failed.
     33  CONVERSION ERROR
     13 format("Failed to find symbol {} not found", id2string(name)) .c_str()' failed.
      7  Message from unrecognized k-induction child process
```

regression/testing_tool.py
```diff
@@ -34,6 +35,19 @@
...
+# The logs will be kept in the director named as "20220211_073815_<test_name>".
+NOW = datetime.now().strftime("%Y%m%d_%H%M%S")
+LOG_DIR = ""
+LOG_RETENTION = False
+
+# Use old C and C++ frontend
+OLD_FRONTEND = False
+
+# user defined operation mode directory
+OP_MODE_DIR = ""
```
rafaelsamenezes: Are you sure all those variables need to be global? Can't you refactor into another class? 

Also this could be a PR by itself

src/goto-programs/builtin_functions.cpp
```diff
@@ -515,8 +515,10 @@ void goto_convertt::do_function_call_symbol(
   const irep_idt &identifier = function.identifier();
 
   const symbolt *symbol;
+
   if(ns.lookup(identifier, symbol))
   {
+    ns.lookup(identifier, symbol);
```
rafaelsamenezes: Isn't this duplicated?

src/langapi/language_ui.cpp
```diff
@@ -31,24 +31,14 @@ bool language_uit::parse()
 
 bool language_uit::parse(const std::string &filename)
 {
-  int mode = get_mode_filename(filename);
+  int mode = get_mode_filename(filename, &(config.options));
```
rafaelsamenezes: I think that config is a global variable already

src/util/message/message.h
```diff
@@ -30,18 +30,18 @@ class messaget
...
-   * 
+   *
    * @param message string to be printed
    * @param location location where the message happened
    */
   virtual void
   error(const std::string &message, const std::string location = "") const
   {
-    print(VerbosityLevel::Error, message, location);
+    print(VerbosityLevel::Error, "###ERROR: " + message, location);
```
rafaelsamenezes: Thats nice, but it could be in a separate PR

rafaelsamenezes (changes requested):

src/util/message/message.h
kunjsong01: Agreed. I'll split this PR into multiple small PRs.

regression/testing_tool.py
kunjsong01: Agreed. I'll split this PR into multiple small PRs.

src/ansi-c/c_main.cpp
```diff
@@ -297,12 +297,17 @@ bool c_main(
   // Call to main symbol is now in "call"; construct calls to thread library
   // hooks for main thread start and main thread end.
 
+  // For the old frontend, we just use pthread_start_main_hook instead of "c:@F@pthread_start_main_hook" in the old frontend.
+  // The latter is used by clang-c-frontend to verify multi-threaded programs and atomics.
+  // If we want to use the latter, we also need to deal with the symbols referred to from the body of c:@F@pthread_strat_main_hook,
+  // such as __ESBMC_num_threads_running, __ESBMC_num_total_threads, and __ESBMC_atomic_end .etc.
+  // This will become an independent task to add thread support. See clang_c_languaget::final function for example.
```
mikhailramalho, line 304: Yeah, I think the problem here is that we use clang to gen the names. How do the programs with dynamic memory work with the old frontend? The symbols there will also have different names.

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -239,10 +239,16 @@ bool clang_c_languaget::typecheck(
...
   contextt new_context(msg);
 
+  // Add symbols in new_context
+  // The symbols here cover AST nodes from internals and source code
   clang_c_convertert converter(new_context, ASTs, msg);
   if(converter.convert())
     return true;
 
+  // Adjust symbol attributes, such as:
+  //  - overwrite macro type with real type
```
mikhailramalho, line 249: what's a macro type?

src/cpp/cpp_language.cpp
```diff
@@ -145,6 +146,7 @@ bool cpp_languaget::parse(
   std::ostringstream o_preprocessed;
 
   internal_additions(o_preprocessed);
+  //std::cout << o_preprocessed.str() << std::endl;
```
mikhailramalho: remove comments.

src/cpp/cpp_language.cpp
```diff
@@ -201,10 +203,10 @@ bool cpp_languaget::final(contextt &context, const messaget &message_handler)
 
 void cpp_languaget::show_parse(std::ostream &out)
 {
-  for(cpp_parse_treet::itemst::const_iterator it = cpp_parse_tree.items.begin();
-      it != cpp_parse_tree.items.end();
-      it++)
-    show_parse(out, *it);
+  for(const auto &it : cpp_parse_tree.items)
+  {
```
mikhailramalho: no need for curly brackets.

src/goto-programs/goto_convert.cpp
```diff
@@ -419,14 +419,8 @@ void goto_convertt::convert_block(const codet &code, goto_programt &dest)
...
-  if(code.operands().size() != 1)
-  {
-    err_location(code);
-    message_handler.error("expression statement takes one operand\n");
-    abort();
-  }
-
-  exprt expr = code.op0();
+  // Old frontend may have the target operand at op1 rather than op0.
+  exprt expr = (code.operands().size() != 1) ? code.op1() : code.op0();
```
mikhailramalho, line 423: can we change the old frontend to set the target in op0?

I'm not always in favor of fixing the old frontend but I don't think we should change the rest of esbmc because of it, as it might introduce bugs that we simply don't need to fix.

src/langapi/language_ui.cpp
mikhailramalho: yeah, no need to pass it around.

mikhailramalho (changes requested):

src/clang-c-frontend/clang_c_language.cpp
kunjsong01: A 'macro type' refers to the `symbol.is_macro` as being used in function `clang_c_adjust::adjust_type`: 
```
515     if(symbol.is_macro)
516       type = symbol.type; // overwrite
```
I'm not 100% understanding 'is_macro', hence just briefly explaining the purpose of adjuster in a general manner.
This comment does appear vague and a little misleading. 

Q: What does 'symbol.is_macro' actually do in the code? What's the meaning of it?

src/goto-programs/goto_convert.cpp
kunjsong01: Agreed. This might introduce some unexpected bugs in the future. I'll fix the old frontend instead of goto converter.

mikhailramalho: yeah, or you can simply ignore it! I mean, the purpose of re-enabling the old frontend is to use the symbol table as a reference, right? If so, you shouldn't worry about the verification being broken later in the verification pipeline.

regression/testing_tool.py
kunjsong01: Fixed and moved to PR#643.

kunjsong01:
Abandoned. 
We are not going to work on the old CPP frontend.

fbrausse:
> We are not going to work on the old CPP frontend.

Does that mean we can remove it?

mikhailramalho:
The only reason the old ones are still around is so that we have a
reference when writing the new ones.

I'm fine with removing them if you all think it's not needed.

It is the last step before killing the old irep/migrate code as well.

kunjsong01:
>Does that mean we can remove it?

We'll remove it when after we bring up the clang-cpp frontend. 
For the time being, the old frontends are just for reference.

PR #638, opened 2022-02-18
added __ESBMC_get_object_size intrinsic #636

### LINKED ISSUE #636: Getting the size of a dynamic object
Is there a way to reason about the size of a dynamic allocated object?

```c
void *foo() {
  return malloc(MAGIC_NUMBER);
}

int bar(void *ptr) {
  unsigned int size = __ESBMC_object_size(ptr); // Do we have something like this?
  assert(size == MAGIC_NUMBER);
}

int main() {
 bar(foo); 
}

```

rafaelsamenezes:
This intrinsic is used to get information about the size of objects.

Should close #636

### DIFF
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -1021,3 +1021,24 @@ void goto_symext::intrinsic_memset(
     bump_call();
   }
 }
+
+void goto_symext::intrinsic_get_object_size(
+  const code_function_call2t &func_call,
+  reachability_treet &)
+{
+  assert(func_call.operands.size() == 1 && "Wrong get_object_size signature");
+  auto ptr = func_call.operands[0];
+
+  // Work out what the ptr points at.
+  internal_deref_items.clear();
+  dereference2tc deref(get_empty_type(), ptr);
+  dereference(deref, dereferencet::INTERNAL);
+
+  assert(is_array_type(internal_deref_items.front().object->type));
+  auto obj_size =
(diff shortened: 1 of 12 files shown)


lucasccordeiro (approved):
LGTM.

mikhailramalho (approved):
LGTM

src/goto-symex/builtin_functions.cpp
```diff
@@ -1027,3 +1027,24 @@ void goto_symext::intrinsic_memset(
...
+  auto ptr = func_call.operands[0];
+
+  // Work out what the ptr points at.
+  internal_deref_items.clear();
+  dereference2tc deref(get_empty_type(), ptr);
+  dereference(deref, dereferencet::INTERNAL);
+
+  assert(is_array_type(internal_deref_items.front().object->type));
+  auto obj_size =
+    to_array_type(internal_deref_items.front().object->type).array_size;
```
fbrausse, line 1039: Is it only ever one/the first object?

rafaelsamenezes: It should be, the dereference should only return the memory symbol.

fbrausse: So the assumption are a) that `__ESBMC_get_object_size` is only applied to pointers (in)to objects returned from `malloc`, `new`, etc. and never to pointers to anything else and b) that the value-set of that pointer is a singleton? If so, could we maybe document these assumption (since they are not clear from the signature of the intrinsic)?

PR #640, opened 2022-02-25
Intrinsic memset optimization

### LINKED ISSUE #635: ESBMC intrinsic memset enhancement
ESBMC uses intrinsic `memset` implementation (`__ESBMC_memset(ptr, val, size)`) to perform a much faster zero'ing of the memory by assigning a zero `expr2tc` directly to the memory addressed by the given pointer `ptr` instead of performing unrolling that is used in the default `memset` implementation. However, at the moment this only works when the allocation size and the size of the pointer type match, and when we allocate a larger block of memory, ESBMC resorts to the default implementation of `memset`. For example, ESBMC will use the intrinsic `memset` in the following piece of code:
```
  1 #include <stdlib.h>
  2
  3 int main()
  4 {
  5   int arr[4];
  6   int *ptr = malloc(sizeof(arr));
  7   ptr = __ESBMC_memset(ptr, 0, sizeof(arr));
  8   return 0;
  9 }
``` 
while when the allocation size in line 6 is increased by 1 (e.g., `int *ptr = malloc(sizeof(arr)+1);`), ESBMC will resort to the default implementation of `memset` that involves unrolling. I think that it should be possible extend the implementation of `__ESBMC_memset` to these kind of scenarios as it significantly outperforms the default `memset` for large allocations.

rafaelsamenezes:
This PR will optimize memset operations by initializing the entire object with the result of the memset. This should avoid unrolling and speed up for memset of known bounds.

Closes #635 

We should review this carefully, as this is a critical part of esbmc. I will also run this on sv-benchmarks.


src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
   symex_assign(code_assign2tc(lhs, va_rhs), true);
 }
 
-void goto_symext::intrinsic_memset(
-  reachability_treet &art,
-  const code_function_call2t &func_call)
+// Computes the equivalent object value when considering a memset operation on it
```
lucasccordeiro, line 788: Can you also document this method in the .h file?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
-    // We're going to execute a function call, and that's going to mess with
-    // the program counter. Set it back *onto* pointing at this intrinsic, so
-    // symex_function_call calculates the right return address. Misery.
-    cur_state->source.pc--;
+  /**
+   * The idea of this expression is to compute the object value
+   * in the case where every byte `value` was set set up until num_of_bytes
+   * 
+   * @warning this function does not add any pointer/memory/bounds check!
+   *          they should be added before calling this function!
```
lucasccordeiro: How could we do it?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+   * 
+   * In summary, there are two main computations here:
+   * 
+   * A. Generate the byte representation, this is mostly through
+   *    the `result` expression. The expression is initialized with zero
+   *    and then, until the num_of_bytes is reached it will do a full byte
+   *    left-shift followed by an bitor operation with the byte value:
+   * 
+   *    Example, for a integer(4 bytes) with memset using 3 bytes and value 0xF1
```
lucasccordeiro: "Example" -> "For example"

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+   * 
+   *    step 1: 0x00000000 -- left-shift 8 -- 0x00000000 -- bitor -- 0x000000F1
+   *    step 2: 0x000000F1 -- left-shift 8 -- 0x0000F100 -- bitor -- 0x0000F1F1
+   *    step 3: 0x0000F1F1 -- left-shift 8 -- 0x00F1F100 -- bitor -- 0x00F1F1F1
+   *    
+   *    Since we only want 3 bytes, the initialized object value would be 0x00F1F1F1
+   * 
+   * B. Generate a mask of the bits that were not set, this is done because skipped bits
+   *    need to be returned back. The computation of this is simple, we initialize every
+   *    bit that was changed by the byte-representation computation with a 1. Which is then
```
lucasccordeiro: "with a 1. Which is then" -> "with a 1, which is then"

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+   * Note about offsets: To handle them, we apply left shifts to the remaining offset after
+   * the computation of the object-value and initial mask representation
+   * 
+   */
+
+  expr2tc result = gen_zero(type);
+  auto value_downcast = typecast2tc(get_uint8_type(), value);
+  auto value_upcast = typecast2tc(
+    type,
+    value_downcast); // so smt_conv will complain about the width of the type
```
lucasccordeiro, line 900: Why? Could we refactor the code?

lucasccordeiro (changes requested):

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
   symex_assign(code_assign2tc(lhs, va_rhs), true);
 }
 
-void goto_symext::intrinsic_memset(
-  reachability_treet &art,
-  const code_function_call2t &func_call)
+// Computes the equivalent object value when considering a memset operation on it
+expr2tc gen_byte_expression(
```
mikhailramalho: static inline?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+  result = bitor2tc(type, result, mask);
 
-  // Work out here whether we can construct an assignment for each thing
-  // pointed at by the ptr.
-  cur_state->rename(size);
-  bool can_construct = true;
-  for(const auto &item : internal_deref_items)
-  {
-    const expr2tc &offs = item.offset;
+  auto simplified = result->simplify();
```
mikhailramalho: I think we have a method simplify that can be used here instead, as it honors the --no-simplify flag.

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
-    {
-      continue;
-    }
-    else if(
-      !is_constant_int2t(offs) && is_constant_int2t(size) &&
-      to_constant_int2t(size).value == tmpsize)
-    {
-      continue;
-    }
+inline expr2tc gen_value_by_byte(
```
mikhailramalho: static inline?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+    expr2tc arg1 = func_call.operands[1];
+    expr2tc arg2 = func_call.operands[2];
+
+    internal_deref_items.clear();
+    dereference2tc deref(get_empty_type(), arg0);
+    dereference(deref, dereferencet::INTERNAL);
+
+    // Is this pointing to a valid place?
+    if(!internal_deref_items.size())
+      throw std::invalid_argument("Couldn't deref the object");
```
mikhailramalho: Can we remove the throws and simply guard against these error cases? Calling the code in the catch block when the if fails.

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,277 +748,389 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+   *        until a limit has arrived. Dynamic memory is dealt here.
+   * 3 - Structs/Union: these are the hardest as we have to take the alignment into
+   *        account when dealing with it. Hopefully the clang-frontend already give it
+   *        to us.
+   * 
+   */
+
+  // I am not sure if bitwise operations are valid for floats
+  if(is_floatbv_type(type) || is_fixedbv_type(type))
+    throw std::invalid_argument("This will not work for floats");
```
mikhailramalho: can we get rid of the throws here? You can return expr2tc() and check the returned value on the caller.

mikhailramalho (commented):
Overall it looks go, except for a few nits.

As always, I'd like to see the sv-comp results before merging, as this piece of code is quite performance critical.

rafaelsamenezes:
@mikhailramalho and @lucasccordeiro Thanks for the reviews! I will actually put it back as a draft, there are 2 features still missing: symbol support and multiple dereference objects. I will address your comments, implement those features and mark it as ready again.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: replaced throws with an empty expression

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: replaced throws with an empty expression

lucasccordeiro:
@rafaelsamenezes: What is the status of this PR?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,18 +748,378 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+  if(is_pointer_type(type))
+    return gen_byte_expression_byte_update(
+      type, src, value, num_of_bytes, offset);
+  expr2tc result = gen_zero(type);
+  auto value_downcast = typecast2tc(get_uint8_type(), value);
+  auto value_upcast = typecast2tc(
+    type,
+    value_downcast); // so smt_conv will complain about the width of the type
+
+  expr2tc mask = gen_zero(type);
```
fbrausse, line 902: I don't understand why `mask` is generated symbolically here. The loop iterations and therefore #shifts are known statically below. Instead something like this could be done keeping the expressions smaller and the simplifier less busy:
```c++
BigInt n, o;
n.setPower2(num_of_bytes * 8); /* or config.ansi_c.char_width if we want to be pedantic */
o.setPower2(offset * 8);
mask = constant_int2tc(type, (n-1) * o);
```

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,18 +748,378 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+
+  auto eight = constant_int2tc(int_type2(), BigInt((int)8));
+  auto one = constant_int2tc(int_type2(), BigInt((int)1));
```
fbrausse: I believe neither the `(int)` casts nor the explicit calls to `BigInt()` are necessary here.

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,18 +748,378 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+   * 1 - Primitives: these are simple: just generate the byte_expression directly
+   * 2 - Arrays: these are ok: just keep generating byte_expression for each member
+   *        until a limit has arrived. Dynamic memory is dealt here.
+   * 3 - Structs/Union: these are the hardest as we have to take the alignment into
+   *        account when dealing with it. Hopefully the clang-frontend already give it
+   *        to us.
+   * 
+   */
+
+  // I am not sure if bitwise operations are valid for floats
```
fbrausse: They are invalid in C and C++ and I hope none of the other frontends generate those. This constraint is also part of #204.

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,18 +748,378 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+     * each component
+     */
+    constant_struct2tc result = gen_zero(type);
+
+    auto bytes_left = num_of_bytes;
+    auto offset_left = offset;
+
+    for(unsigned i = 0; i < result->datatype_members.size(); i++)
+    {
+      auto name = to_struct_type(type).member_names[i];
```
fbrausse: Maybe it's just me, but the fewer uses of unnecessary `auto` the better when refactoring because the compiler will complain about all places that might need to be touched. Would you consider decreasing the usage of `auto`?

src/goto-symex/builtin_functions.cpp
```diff
@@ -748,18 +748,378 @@ void goto_symext::symex_va_arg(const expr2tc &lhs, const sideeffect2t &code)
...
+      if(offset_left >= current_member_size)
+      {
+        result->datatype_members[i] = local_member;
+        offset_left -= current_member_size;
+      }
+      else
+      {
+        assert(offset_left < current_member_size);
+        auto bytes_to_write =
+          bytes_left < current_member_size ? bytes_left : current_member_size;
```
fbrausse: `std::min`?

src/goto-symex/builtin_functions.cpp
```diff
@@ -780,246 +1140,181 @@ void goto_symext::intrinsic_memset(
...
     return;
   }
 
-  // Work out what the ptr points at.
-  internal_deref_items.clear();
-  dereference2tc deref(get_empty_type(), ptr);
-  dereference(deref, dereferencet::INTERNAL);
+  auto simplified = arg2->simplify();
+  if(simplified)
+    arg2 = simplified;
```
fbrausse: `simplify(arg2)`?

src/goto-symex/builtin_functions.cpp
```diff
@@ -780,246 +1140,181 @@ void goto_symext::intrinsic_memset(
...
+      bump_call();
+      return;
     }
-    else if(
-      !is_constant_int2t(offs) && is_constant_int2t(size) &&
-      to_constant_int2t(size).value == tmpsize)
+
+    auto offset_simplified = item_offset->simplify();
+    if(offset_simplified)
+      item_offset = offset_simplified;
```
fbrausse: same: `simplify(...)`?

src/goto-symex/builtin_functions.cpp
```diff
@@ -780,246 +1140,181 @@ void goto_symext::intrinsic_memset(
...
-          else
-          {
-            // Converting offset into bits here
-            expr2tc new_offset =
-              mul2tc(this_offs->type, this_offs, gen_ulong(8));
-            simplify(new_offset);
+      /* If we reached here, item_offset is not symbolic
+       * and we don't know what the actual value of it is...
+       *
+       * For now bum_call later expand our simplifier
```
fbrausse: typo

src/goto-symex/builtin_functions.cpp
```diff
@@ -780,246 +1140,181 @@ void goto_symext::intrinsic_memset(
...
+      bump_call();
+      return;
     }
 
-    // Alternately, we might be memsetting a field within a struct. Don't allow
-    // setting more than one field at a time, unaligned access, or the like.
-    // If you're setting a random run of bytes within a struct, best to use
-    // the C implementation.
-    if(is_struct_type(item.object->type) && is_constant_int2t(size))
+    // TODO: Why (X*Y)/Y is not X?
```
fbrausse: overflow maybe? not sure :) can you elaborate?

lucasccordeiro:
@rafaelsamenezes: What is the status of this PR?

lucasccordeiro:
@rafaelsamenezes: could you please address @fbrausse's comments?

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Kinda yes. 

However, this operations are happening for offset, which should be safe.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Ow, didn't know that we could use like this.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Same as above

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: ACK

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: I've tried this approach and it doesn't seems to be working. Anyway, the formula is simplified. It shouldn't add much cost.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Sure, no problem

lucasccordeiro:
@rafaelsamenezes: as discussed, we should move forward with this PR :-)

rafaelsamenezes:
This PR:

```
2023-05-26T02:27:40.8875950Z Statistics:          26459 Files
2023-05-26T02:27:40.8878371Z   correct:           14378
2023-05-26T02:27:40.8900102Z     correct true:     8154
2023-05-26T02:27:40.8901427Z     correct false:    6224
2023-05-26T02:27:40.8902410Z   incorrect:            62
2023-05-26T02:27:40.8903417Z     incorrect true:     31
2023-05-26T02:27:40.8904293Z     incorrect false:    31
2023-05-26T02:27:40.8904830Z   unknown:           12019
2023-05-26T02:27:40.8905378Z   Score:             21044 (max: 43043)
```

Master:

```
  correct:           14084
    correct true:     7852
    correct false:    6232
  incorrect:            88
    incorrect true:     42
    incorrect false:    46
  unknown:           12287
  Score:             19856 (max: 43043)
```

From the incorrect solved:

- 6 incorrect false (invalid-deref) became timeouts (goblint-coreutils)
- 15 incorrect in unreach became timeouts
- 4 incorrect true became correct false (in reachability): `aws_array_list_init_dynamic_harness_negated.yml, aws_hash_iter_begin_harness_negated.yml, aws_ring_buffer_init_harness_negated.yml, and ldv-validator-v0.6/linux-stable-4ed3cba-1-100_1a-drivers--usb--serial--qcserial.ko-entry_point.cil.out.yml`
- 2 incorrect became unknown in reachability: still checking why.

Might be worth investigating those 4 incorrect true as this patch shouldn`t improve soundness.

There is two new incorrect results: loops/ludcmp.yml (was correct), and ldv-linux-3.14/linux-3.14_complex_emg_linux-kernel-locking-spinlock_drivers-net-ethernet-dlink-dl2k.cil.yml (was timeout)

lucasccordeiro:
Many thanks, @rafaelsamenezes. Can I ask you to fix the code style and rebase this PR with the master?

lucasccordeiro (approved):
Many thanks, @rafaelsamenezes. Can I ask you to fix the code style and rebase this PR with the master?

rafaelsamenezes:
Latest PR results:

```
   correct:           14369
     correct true:     8148
     correct false:    6221
   incorrect:            61
     incorrect true:     30
     incorrect false:    31
   unknown:           12029
   Score:             21061 (max: 43043)
```

PR #650, opened 2022-03-18
Contractor

Mohannad-Aldughaim:
Hi all.

This is WIP for using contractors in bmc as described in [here](https://arxiv.org/abs/2012.11245). 

When reviewing, I suggest starting at `goto_contractort.h`

## Goto Contractor

The class 'goto_contractort' does the following on the goto program:
-Parse the properties in the code which are inside the assert. (for now, only one assert). This will parse the expressions from ESBMC format to IBEX format.
-Parse the assumes created by `--interval-analysis` option in ESBMC. This is for creating the intervals of the variables in the assert.
-Use the Forward-backward contractor in the IBEX library to reduce the intervals.
-Insert assumes reflecting the contracted intervals.

For now, it only works correctly in one case which is if the intervals contracted are at the end of a loop.

## IBEX

-IBEX is a C++ library for constraint processing over real numbers. It has an implementation of contractors and intervals arithmetic and set operations. Can be found [here](http://ibex-lib.org/)
-The main reason to include this library in ESBMC is to handle the intervals and contractors. 

## Example

The prime example that I have been working on is a benchmark from SV-comp which is [afnp2014.c](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/loop-lit/afnp2014.c)

## Support

- Only int. As `--interval-analysis` option does not support floats.
- One property. So only one constraint in IBEX. 
- One loop. 
- No support for function calls.
- Binary arithmetic operators supported: +, -, *, / 
- Binary comparison operators supported: >,>=,<,<= 

## Implementation Notes

-Added a new option in ESBMC called `--goto-contractor`.
-Should be used with `--incremental-bmc` for now.


src/esbmc/bmc.cpp
```diff
@@ -421,7 +421,7 @@ void bmct::bidirectional_search(
   // We should only analyse the inductive step's cex and we're running
   // in k-induction mode
   if(!(options.get_bool_option("inductive-step") &&
-       options.get_bool_option("k-induction")))
+       options.get_bool_option("k-induction") ))
```
mikhailramalho: remove extra space.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1580,6 +1580,12 @@ bool esbmc_parseoptionst::process_goto_program(
       goto_k_induction(goto_functions, msg);
     }
 
+    if(cmdline.isset("goto-contractor"))
+    {
+      interval_analysis(goto_functions, ns);
+      goto_contractort(goto_functions, msg);
+    }
```
mikhailramalho, line 1619: this will run the interval analysis twice if the `--interval-analysis` is used with the new flag.

You can change:
```
    if(cmdline.isset("interval-analysis"))
      interval_analysis(goto_functions, ns);
```
to
```
    if(cmdline.isset("interval-analysis") || cmdline.isset("goto-contractor"))
      interval_analysis(goto_functions, ns);
```
and remove the interval analysis from this branch.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
+//
+// Created by Mohannad Aldughaim on 09/01/2022.
+//
+
+#include "goto_contractor.h"
+
+using namespace std;
```
mikhailramalho: Remove this and use the `std` namespace explicitly.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+    it++;
+  }
+}
+
+void goto_contractort::get_intervals(goto_functionst goto_functions)
+{
+  //only main here
+  auto function = goto_functions.function_map.find("c:@F@main");
+  std::cout << "\n parsing assumes\nfound main: " << function->first << "\n";
```
mikhailramalho: You need to remove the cout and use our messaget interface for messages.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+  symbol2tc symbol;
+  double value;
+
+  if(!is_comp_expr(expr))
+    return;
+
+  std::shared_ptr<relation_data> rel;
+  rel = dynamic_pointer_cast<relation_data>(expr);
+
+  //expr->dump();
```
mikhailramalho: remove unnecessary comments.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+  //side1 1 is a symbol or typecast to symbol
+  auto side1 = rel->side_1;
+  //side 2 is always a number.
+  auto side2 = rel->side_2;
+
+  if(is_typecast2t(side1))
```
mikhailramalho: You can use `get_base_object` here, then check if the base object is a symbolt. If it isn't return.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    else
+      return;
+  else if(is_symbol2t(side1))
+    symbol = to_symbol2t(side1);
+  else
+    return;
+
+  if(!is_constant_int2t(side2) && !is_constant_floatbv2t(side2))
+    return;
+  if(is_constant_int2t(side2))
```
mikhailramalho: `value` is a double. You'll probably need to handle the values differently here.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    value = to_constant_int2t(side2).as_long();
+  else
+    return;
+
+  int index =
+    map->find((string)symbol->get_symbol_name().c_str());
+  if(index == -1) return;
```
mikhailramalho: move return to the next line.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+  if(!is_constant_int2t(side2) && !is_constant_floatbv2t(side2))
+    return;
+  if(is_constant_int2t(side2))
+    value = to_constant_int2t(side2).as_long();
+  else
+    return;
+
+  int index =
+    map->find((string)symbol->get_symbol_name().c_str());
```
mikhailramalho: you probably don't need this cast.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+  if(index == -1) return;
+  if(is_greaterthan2t(expr) || is_greaterthanequal2t(expr))
```
mikhailramalho: You can replace this with a switch over the expr type.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    map->update_lb_interval(value, index);
+    map->update_ub_interval(value, index);
+  }
+}
+
+void goto_contractort::insert_assume(
+  goto_functionst goto_functions,
+  IntervalVector new_intervals)
+{
+  std::cout << "Inserting assumes.. " << std::endl;
```
mikhailramalho: use messaget interface.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+  auto goto_function = it->second;
+
+  for(int i = 0; i < map->size(); i++)
+  {
+    symbol2tc X = map->symbols[i];
+    if(
+      isfinite(new_intervals[i].lb()) &&
+      map->intervals[i].lb() != new_intervals[i].lb())
+    {
+      if(is_signedbv_type(X))
```
mikhailramalho: remove empty if.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,359 @@
...
+      e->location = loop_exit->location;
+      goto_function.body.destructive_insert(loop_exit, tmp_e);
+    }
+  }
+}
+
+IntervalVector goto_contractort::contractor()
+{
+  NumConstraint c = *constraint;
+  std::cout << "My Constraint:" << c << std::endl;
```
mikhailramalho: use messaget interface.

mikhailramalho (changes requested):
I've seen the first few changes and there are already several changes needed. Please, fix them and I'll take another look.

lucasccordeiro:
@Mohannad-Aldughaim: @mikhailramalho's comments are very relevant here. Could you please address them before others can review your PR? Additionally, there are conflicting files in your PR. Could you please solve them?

Mohannad-Aldughaim:
Will do. Thanks for the comments.

src/esbmc/bmc.cpp
Mohannad-Aldughaim: done.

src/esbmc/esbmc_parseoptions.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done. Though I left some for reference later. Will be removed when I don't need them anymore.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Replaced all cout with messaget interface.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: This one is a bit complex. The library Ibex uses doubles in intervals. So I need to convert from int to double and back to int. Not sure yet of the safest way to do this.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/util/goto_expr_factory.cpp
```diff
@@ -0,0 +1,323 @@
+// Rafael Sá Menezes - 04/2020
```
lucasccordeiro: Should this header be updated?

src/util/goto_expr_factory.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+#include "goto_expr_factory.h"
+
+bool is_symbols_equal(expr2tc &v1, expr2tc &v2)
```
lucasccordeiro: Can I ask you to write some short documentation for each method?

src/util/goto_expr_factory.cpp
```diff
@@ -0,0 +1,323 @@
...
+  assert(v1->expr_id == expr2t::expr_ids::constant_int_id);
+  assert(v2->expr_id == expr2t::expr_ids::constant_int_id);
+
+  constant_int2tc value1 = v1;
+  constant_int2tc value2 = v2;
+
+  return value1->value.compare(value2->value) == 0;
+}
+
+symbol2tc create_unsigned_32_symbol_expr(std::string name)
```
lucasccordeiro: Why do you need 32? What about the other bit-widths? 16, 64, 128?

src/util/goto_expr_factory.cpp
```diff
@@ -0,0 +1,323 @@
...
+  typecast2tc lhs_typecast (get_int32_type(),lhs);
+  typecast2tc rhs_typecast (get_int32_type(),rhs);
+
+  lessthanequal2tc relation(lhs_typecast, rhs_typecast);
+  return relation;
+}
+
+greaterthanequal2tc create_greaterthanequal_relation(expr2tc &lhs, expr2tc &rhs)
+{
+  typecast2tc lhs_typecast (get_int32_type(),lhs);
```
lucasccordeiro: What about other bit-widths? 16, 64, 128?

src/util/goto_expr_factory.h
```diff
@@ -0,0 +1,98 @@
+/*******************************************************************\
+Module: Cache unit test utilities
+Author: Rafael Sá Menezes
```
lucasccordeiro, line 3: Can you please update this header?

src/solvers/smt/smt_conv.cpp
```diff
@@ -942,7 +942,12 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
     }
     else
     {
-      assert(is_signedbv_type(lte.side_1) && is_signedbv_type(lte.side_2));
+      if(!(is_signedbv_type(lte.side_1) && is_signedbv_type(lte.side_2)))
```
lucasccordeiro: Should this piece of code be considered as debug? If so, add something like #ifdef DEBUG .... #endif

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,168 @@
...
+  function_loopst function_loops;
+
+  messaget message_handler;
+
+  void get_constraints(goto_functionst functionst);
+  void get_intervals(goto_functionst functionst);
+
+  IntervalVector contractor();
+
+  ibex::CmpOp get_complement(ibex::CmpOp);
```
lucasccordeiro, line 223: Could you please write a short documentation for each method?

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,168 @@
...
+  {
+    message_handler = _message_handler;
+
+    run1();
+    if(function_loops.size())
+    {
+      //initialize map
+      map = new MyMap();
+      vars = new Variable(MAX_VAR);
+      //find properties
```
lucasccordeiro: These comments are poor.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,430 @@
...
+    //Abort contractor
+  }*/
+  return f;
+}
+
+int goto_contractort::create_variable_from_expr2t(irep_container<expr2t> expr)
+{
+  std::string var_name = to_symbol2t(expr).get_symbol_name().c_str();
+  int index = map->find(var_name);
+  if(index == -1)
```
lucasccordeiro: Why index == -1?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,430 @@
...
+    loop = function_loop;
+
+  auto loop_exit = loop.get_original_loop_exit();
+
+  goto_programt dest(message_handler);
+
+  auto it = goto_functions.function_map.find("c:@F@main");
+  auto goto_function = it->second;
+
+  for(int i = 0; i < map->size(); i++)
```
lucasccordeiro: should `size_t` be used here instead of `int`?

lucasccordeiro (changes requested):

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: changed to NOT_FOUND value to indicate that the variable does not exist in the map.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: changed to Doxygen format

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1584,6 +1590,15 @@ bool esbmc_parseoptionst::process_goto_program(
...
     }
 
+    if(cmdline.isset("goto-contractor"))
+    {
+      #ifdef ENABLE_GOTO_CONTRACTOR
+      goto_contractort(goto_functions, msg);
+      #else
+      msg.error("Current build does not support contractors");
+      #endif
+    }
```
mikhailramalho, line 1619: define indentation needs to be fixed here

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+{
+  //only main here
+  auto function = goto_functions.function_map.find("c:@F@main");
+  auto it = function->second.body.instructions.begin();
+
+  while(it != function->second.body.instructions.end())
+  {
+    if(it->is_assume())
+    {
+      it->dump();
```
mikhailramalho: remove dump

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+  std::shared_ptr<relation_data> rel;
+  rel = std::dynamic_pointer_cast<relation_data>(expr);
+
+  //side1 1 is a symbol or typecast to symbol
+  auto side1 = rel->side_1;
+  //side 2 is always a number.
+  auto side2 = rel->side_2;
+
+  auto obj = get_base_object(side1);
+  if(is_symbol2t(obj))
```
mikhailramalho: Can you do `if(!is_symbol(obj)) return;`?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+    symbol = to_symbol2t(obj);
+  else
+    return;
+  ///keeping this for later
```
mikhailramalho: Do we really need to keep this?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+  //side2->dump();
+  bool neg = false;
+  if(is_neg2t(side2))
+  {
+    neg = true;
+    side2 = to_neg2t(side2).value;
+  }
+  if(!is_constant_int2t(side2) && !is_constant_floatbv2t(side2))
+    return;
+  if(is_constant_int2t(side2))
```
mikhailramalho: Can you do `if(!is_constant_int2t(side2))` return;?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+    value = to_constant_int2t(side2).as_long() * (neg?-1:1);
+  else
+    return;
+
+  int index = map->find(symbol->get_symbol_name());
+  if(index == NOT_FOUND)
+    return;
```
mikhailramalho, line 76: what's not_found here?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+      e->inductive_step_instruction = false;
+      e->guard = cond;
+      e->location = loop_exit->location;
+      goto_function.body.destructive_insert(loop_exit, tmp_e);
+    }
+  }
+}
+
+IntervalVector goto_contractort::contractor()
+{
```
mikhailramalho, line 136: remove the unnecessary comments here

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+  std::shared_ptr<relation_data> rel;
+  rel = std::dynamic_pointer_cast<relation_data>(expr);
+
+  Function *f, *g;
+  f = create_function_from_expr2t(rel->side_1);
+  g = create_function_from_expr2t(rel->side_2);
+  switch(expr->expr_id)
+  {
+  case expr2t::expr_ids::greaterthanequal_id:
+    c = new NumConstraint(*vars, (*f)(*vars) >= (*g)(*vars));
```
mikhailramalho: do we call delete on these object anywhere?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+  case expr2t::expr_ids::lessthan_id:
+    c = new NumConstraint(*vars, (*f)(*vars) < (*g)(*vars));
+    break;
+  case expr2t::expr_ids::equality_id:
+    c = new NumConstraint(*vars, (*f)(*vars) = (*g)(*vars));
+    break;
+  default:
+    break;
+  }
+  ///keeping this for later
```
mikhailramalho: do we really need to keep this?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+
+Function *
+goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
+{
+  Function *f = nullptr;
+  Function *g, *h;
+
+  if(is_comp_expr(expr))
+  {
+    //Abort contractor
```
mikhailramalho: ?

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,453 @@
...
+  {
+    int new_index = map->add_var(var_name, to_symbol2t(expr));
+    if(new_index != NOT_FOUND)
+      return new_index;
+    return NOT_FOUND;
+  }
+  return index;
+}
+
+bool goto_contractort::run1()
```
mikhailramalho: can you rename this function to something that's more meaningful?

mikhailramalho (commented):
Overall I think it needs more cleanups but it seems like it's getting there.

src/util/goto_expr_factory.cpp
rafaelsamenezes: added a second parameter to work with any type

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: changed.

src/esbmc/esbmc_parseoptions.cpp
Mohannad-Aldughaim: fixed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: removed

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Not any more. Removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: changed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: It is defined to indicate that this symbol is not in the tracked symbols list. Its value is -1.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: removed

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Not anymore. Removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Created destructor.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Updated to return nullptr so we can resume verification if goto-contractor is not applicable.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Done.

src/goto-programs/goto_contractor.cpp
mikhailramalho: Using BigInt should do the trick... When you need to convert to double just call .toint64().

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,338 @@
...
+  symbol = to_symbol2t(obj);
+  bool neg = false;
+  if(is_neg2t(side2))
+  {
+    neg = true;
+    side2 = to_neg2t(side2).value;
+  }
+  if(!is_constant_int2t(side2))
+    return;
+  if(is_constant_int2t(side2))
```
mikhailramalho: At this point we know that side2 is a constant_int2tc, so no need to check it again.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,338 @@
...
+  case expr2t::expr_ids::lessthan_id:
+  case expr2t::expr_ids::lessthanequal_id:
+    map->update_ub_interval(value, index);
+    break;
+  case expr2t::expr_ids::equality_id:
+    map->update_lb_interval(value, index);
+    map->update_ub_interval(value, index);
+    break;
+  default:;
+    //error
```
mikhailramalho: you can use the message interface and call abort();

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,239 @@
...
+    intervals.resize(MAX_VAR);
+  }
+  int add_var(std::string name, symbol2t symbol)
+  {
+    if(find(name) == NOT_FOUND && n < MAX_VAR)
+    {
+      symbols[n] = symbol;
+      var_name[n] = name;
+
+      ///keep for later. set initial intervals
```
mikhailramalho: remove comment.

src/solvers/smt/smt_conv.cpp
mikhailramalho: yes, this needs to go away.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,239 @@
...
+        insert_assume(_goto_functions, new_intervals);
+      }
+    }
+  }
+  /**
+   * This destructor that will delete all objects allocated of type Function
+   * and NumConstraint
+   */
+  ~goto_contractort()
+  {
```
mikhailramalho: can you do something like:

for( auto c : list)
  delete(p);
list.clear();

I don't think you need to check if p is not a nullptr before calling the destructor, at least you don't need it for free.

mikhailramalho (commented):
Only minor comments now.

Please before the next review check the code style:
1. One line ifs/for should not use curly brackets.
2. Remove unnecessary comments.
3. Check code-style with the same clang-format we use in the CI.

mikhailramalho:
Also, please look at lucas' comment about a debug code.

scripts/cmake/FindIbex.cmake
```diff
@@ -0,0 +1,37 @@
...
+find_library(IBEX_LIB ibex libibex HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES lib bin)
+find_library(IBEX_LIB_GAOL gaol libgaol HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES lib/ibex/3rd bin)
+find_library(IBEX_LIB_GDTOA gdtoa libgdtoa HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES lib/ibex/3rd bin)
+find_library(IBEX_LIB_ULTIM ultim libultim HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES lib/ibex/3rd bin)
+find_library(IBEX_LIB_SOPLEX soplex libsoplex HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES lib/ibex/3rd bin)
+find_path(IBEX_INCLUDE_DIRS ibex.h HINTS ${IBEX_DIR} $ENV{HOME}/ibex PATH_SUFFIXES include)
+
+if(IBEX_INCLUDE_DIRS STREQUAL "IBEX_INCLUDE_DIRS-NOTFOUND")
+        message(FATAL_ERROR "Could not find ibex include headers, please check IBEX_DIR")
+    endif()
```
fbrausse: Please fix the indentation here.

src/esbmc/options.cpp
```diff
@@ -236,6 +236,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"forward-condition", NULL, "check the forward condition"},
     {"inductive-step", NULL, "check the inductive step"},
     {"k-induction", NULL, "prove by k-induction "},
+    {"goto-contractor", NULL, "prove by using contractors"},
```
fbrausse: Could we be a bit more precise here? Technically, contractors do not prove by themselves, just in concert with other techniques. I would suggest to label this option something akin "enable contractor-based interval refinements on goto level".

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,384 @@
+//
+// Created by Mohannad Aldughaim on 09/01/2022.
+//
+
+#include "goto_contractor.h"
```
fbrausse: Please use non-local #includes for uniformity, e.g. `#include <goto-programs/goto_contractor.h>`. The `src` subdirectory is an include-directory by default.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
+//
+// Created by Mohannad Aldughaim on 09/01/2022.
+//
+
+#ifndef ESBMC_GOTO_CONTRACTOR_H
+#define ESBMC_GOTO_CONTRACTOR_H
+
+#include "goto_k_induction.h"
```
fbrausse: Please use non-local #includes everywhere.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+#include <ibex/ibex_Interval.h>
+#include <ibex/ibex_Expr.h>
+#include <ibex/ibex_Ctc.h>
+#include "irep2/irep2.h"
+#include "util/type_byte_size.h"
+
+#define MAX_VAR 10
+#define NOT_FOUND -1
+
+using namespace ibex;
```
fbrausse: Please don't pollute the global namespace in headers by `using namespace`; just prefix the corresponding symbols with `ibex::`.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+    goto_functionst &_goto_functions,
+    const messaget &_message_handler)
+    : goto_functions_algorithm(_goto_functions, true)
+  {
+    message_handler = _message_handler;
+    initialize_main_function_loops();
+    if(function_loops.size())
+    {
+      map = new CspMap();
+      vars = new Variable(MAX_VAR);
```
fbrausse: Couldn't `map` and `vars` be plain sub-objects instead of being pointers to heap-allocated objects? Same question for other members of this class.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+  ///vars variable references to be used in Ibex formulas
+  Variable *vars;
+  ibex::Ctc *ctc;
+  /// map is where the variable references and intervals are stored.
+  CspMap *map;
+  /// constraint is where the constraint for CSP will be stored.
+  ibex::NumConstraint *constraint;
+
+  std::list<Function *> list_of_functions;
+  std::list<NumConstraint *> list_of_constraints;
```
fbrausse: Could we instead of a list of pointers use lists of objects, e.g. `std::list<ibex::Function>`?

src/solvers/smt/smt_conv.cpp
```diff
@@ -268,8 +268,8 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
   {
     // Convert /all the arguments/. Via magical delegates.
     unsigned int i = 0;
-    expr->foreach_operand(
-      [this, &args, &i](const expr2tc &e) { args[i++] = convert_ast(e); });
+    expr->foreach_operand([this, &args, &i](const expr2tc &e)
+                          { args[i++] = convert_ast(e); });
```
fbrausse: This formatting change is caused by a new version of clang, use clang-11 to get the original formatting.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1584,6 +1590,16 @@ bool esbmc_parseoptionst::process_goto_program(
       goto_k_induction(goto_functions, msg);
     }
 
+    if(cmdline.isset("goto-contractor"))
+    {
+#ifdef ENABLE_GOTO_CONTRACTOR
+      goto_contractort(goto_functions, msg);
```
fbrausse: Creating an anonymous temporary object of type `goto_contractort` is a perfectly legal way to run this functionality, but it is somewhat unconventional inside ESBMC. Would you consider making this (maybe including the #ifdef guards) a call to a proper function?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1584,6 +1590,16 @@ bool esbmc_parseoptionst::process_goto_program(
...
+#else
+      msg.error("Current build does not support contractors");
```
fbrausse: Please include a hint to the user how the build process should be altered to support contractors, referring to the corresponding CMake variables.

src/goto-programs/goto_contractor.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+void goto_contractort::get_constraints(goto_functionst goto_functions)
+{
+  auto function = goto_functions.function_map.find("c:@F@main");
+  auto it = function->second.body.instructions.begin();
+  while(it != function->second.body.instructions.end())
```
fbrausse: This loop and the one in the function below could be simplified by `for (const auto &ins : function->second.body.instruction) { ... }`.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+   * the variable name, operator is <=,>= depending if its an upper or a lower
+   * limit and value is the value of the interval limit.
+   * @param goto_functions goto program functions
+   * @param vector result from the contractor.
+   */
+  void insert_assume(goto_functionst goto_functions, IntervalVector vector);
+
+  Ctc *create_contractors_from_expr2t(irep_container<expr2t>);
+  NumConstraint *create_constraint_from_expr2t(irep_container<expr2t>);
+  Function *create_function_from_expr2t(irep_container<expr2t>);
```
fbrausse: Instead of a heap-allocated `Function` this function could return a [`std::optional<Function>`](https://en.cppreference.com/w/cpp/utility/optional) instead. Same for `create_constraint_from_expr2t` above. This way we would reduce the number of heap allocations and indirections in accessing the objects.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+#include <util/std_types.h>
+#include <ibex.h>
+#include <ibex/ibex_Interval.h>
+#include <ibex/ibex_Expr.h>
+#include <ibex/ibex_Ctc.h>
+#include "irep2/irep2.h"
+#include "util/type_byte_size.h"
+
+#define MAX_VAR 10
+#define NOT_FOUND -1
```
fbrausse: Please consider making these `static constexpr` members of the `goto_contractort` class since they are specific to that functionality.

src/goto-programs/goto_contractor.h
```diff
@@ -0,0 +1,257 @@
...
+    add_interval(lb, intervals[index].ub(), index);
+  }
+  void update_ub_interval(double ub, int index)
+  {
+    add_interval(intervals[index].lb(), ub, index);
+  }
+  int find(std::string name)
+  {
+    for(size_t i = 0; i < n; i++)
+      if(var_name[i] == name)
```
fbrausse: Isn't it possible to use a proper `std::map` or even `std::unordered_map` instead of this linear search using string comparisons? That way you might also be able to get rid of the rather arbitrary `MAX_VAR` limit.

fbrausse (commented):
I've added a few remarks and suggestions. Overall it seems there still are some things to do, but it is taking shape, nice!

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: > Using BigInt should do the trick... When you need to convert to double just call .toint64(). 

May I ask you to clarify? i didnt understand what you mean by using BigInt.

src/esbmc/esbmc_parseoptions.cpp
Mohannad-Aldughaim: May I ask you to clarify? I have never used such a thing before.

src/goto-programs/goto_contractor.cpp
mikhailramalho: BigInt is a type that can hold large integers and that can be converted to `int64_t` and `uin64_t`. Since you don't seem to be handling fp intervals, you can make `value` a `BigInt`, then call `value.to_int64()` when calling the `ibex` functions.

src/esbmc/esbmc_parseoptions.cpp
fbrausse: Well, you have done it here :)

What happens here is a call to the constructor `goto_contractort(...)`. This call creates an "anonymous" and in this case also "temporary" object, because it not assigned to anything. Here it is equivalent to code that would create a named object like so: `goto_contractort var(...)` and there is no difference between the two, because the the lifetime of this `var` object directly ends on the next line as does that of the temporary version.

So it runs constructor followed by destructor, ok, fine; but it requires a bit more thinking (at least for me) to understand what's going on. My suggestion was to wrap this into a stand-alone function and call that, maybe like so:
```c++
void goto_contractor(...)
{
#ifdef ENABLE_GOTO_CONTRACTOR
  goto_contractort(...); /* all functionality invoked by constructor */
#else
  msg.error(...);
  abort();
#endif
}
```
It's just a suggestion; it might already be enough to just add such a comment in order to minimize confusion. The advantage of having such a function is that you could also move the invocation of the algorithm (labelled by messages "1/5", "2/5", etc.) from the constructor to this function and would have a more object-oriented approach in that external algorithms could just work with an object of type `goto_contractort` instead of embedding the algorithm inside it.

src/solvers/smt/smt_conv.cpp
Mohannad-Aldughaim: changed to clang-11

src/solvers/smt/smt_conv.cpp
Mohannad-Aldughaim: reverted to original file

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: changed

src/esbmc/options.cpp
Mohannad-Aldughaim: I agree. changed

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: changed.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: changed CspMap. But vars remained because it is used with every Function object created in ibex.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: changed.

src/esbmc/esbmc_parseoptions.cpp
Mohannad-Aldughaim: Thanks for the feedback. I have changed it now to work like other options in ESBMC. Particularly, similar to k-induction where there's a function call called goto_contractor and inside it will crate the object goto_contractort. Also added the #ifdef guards.

mikhailramalho (approved):
LGTM! As long as the CI is green, I think we are good to go.

lucasccordeiro (approved):

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: I tried to do it using normal objects but ibex was not copying the values. so I kept it as pointers, for now, to be able to check if any of them failed by checking for nullptr.

PR #322, opened 2020-12-14
Builtin support

rafaelsamenezes:
This will add the following builtins:

- `constant_p`
- `unreachable`
- `sync_fetch_and_add`

This also creates an initial template for model unit test

PR #651, opened 2022-03-21
[builtin] added builtin_constant_p

rafaelsamenezes:
This will add the constant_p builtin into esbmc.

Still missing some corner cases though.

This was part of an old PR, I am still implementing some of the comments

- https://github.com/esbmc/esbmc/pull/322


mikhailramalho:
Do we need to handle it at symex? Can't it be handled in the frontend like `__builtin_isgreater` and the others builtins?

rafaelsamenezes:
> Do we need to handle it at symex? Can't it be handled in the frontend like __builtin_isgreater and the others builtins?

Actually yes. Constant p behaviour depends on optimizations and flows. You even commented here: https://github.com/esbmc/esbmc/pull/322#issuecomment-762485016

mikhailramalho:
Oh, I see!

I think we should drop the define and handle the language builtins separately from the esbmc builtins (I never liked how I've implemented the `__ESBMC_overflow_smul` and the others builtins).

Could you create another PR to handle `c:@F@__builtin` separately, similar to `c:@F@__ESBMC`?

mikhailramalho (approved):
LGTM!

Should we print a warning when running esbmc with `--no-propagation`?

rafaelsamenezes:
> Could you create another PR to handle c:@F@__builtin separately, similar to c:@F@__ESBMC?

I didn't understand the comment. Like having a builtin_intrinsics functions on goto-symex? We might open an issue to discuss this

mikhailramalho:
> > Could you create another PR to handle c:@f@__builtin separately, similar to c:@f@__ESBMC?
> 
> I didn't understand the comment. Like having a builtin_intrinsics functions on goto-symex? We might open an issue to discuss this

Yes, exactly. That way we can remove these defines and handle compiler builtins and esbmc's intrinsics separately.

fbrausse (commented):
LGTM except one detail: according to https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html the return type of this builtin is `int`, not `_Bool`. That distinction is relevant when the builtin is used with e.g. `_Generic()`.

fbrausse:
erm, so no `int` return type?

rafaelsamenezes:
Ow, I actually missread it. I will fix it on master

PR #653, opened 2022-03-21
[builtin] added an intrinsic function for stack allocation

rafaelsamenezes:
This adds the intrinsic `__ESBMC_alloca`, it will have the same behaviour as the `__builtin_alloca`. This is part of the solution for #652

### DIFF
--- a/src/clang-c-frontend/clang_c_language.cpp
+++ b/src/clang-c-frontend/clang_c_language.cpp
@@ -162,6 +162,8 @@ void clang_c_languaget::build_compiler_args(const std::string &tmp_dir)
 
   compiler_args.emplace_back("-D__builtin_memcpy=memcpy");
 
+  compiler_args.emplace_back("-D__ESBMC_alloca=__builtin_alloca");
+
   // Ignore ctype defined by the system
   compiler_args.emplace_back("-D__NO_CTYPE");
 


rafaelsamenezes:
Sorry everyone, don't know what I did. But it seems that I merged this somehow.

Let me still know the discussion. I will revert this back if needed

mikhailramalho:
Why do we need this? `__builtin_alloca` is already handled during the goto conversion.

rafaelsamenezes:
> Why do we need this?

Do we have a reliable and proper way to define this besides a C extension?

mikhailramalho:
I don't understand your question, can you clarify?

rafaelsamenezes:
> I don't understand your question, can you clarify?

I mean that  `__builtin` are C extensions that come from GCC and/or Clang. If codebase of the user comes from another compiler such as MSVC, it will have `_alloca`.  Which in summary, are not part of the C standard (right?).

We should have a unified way to say that you want a stack allocation, so that this is consistent between platforms. In this sense, the user will learn that independent of the platform `__ESBMC_alloca` is what they should use.

mikhailramalho:
I don't think ESBMC should care about if that's an extension or how it's defined, if the user writes `_alloca`, `__builtin_alloca` or `alloca`. If ESBMC finds such calls, we know that should be modeled dynamically and cleaned up at the end of the scope.

In this PR we are expecting the user to write `__ESBMC_alloca` in their benchmark but I think that's unrealistic... unless the user wants to annotate their code with ESBMC's intrinsics.

rafaelsamenezes:
> In this PR we are expecting the user to write __ESBMC_alloca in their benchmark but I think that's unrealistic... unless the user wants to annotate their code with ESBMC's intrinsics.

Which is the case. Besides, I will create a `esbmc.h` (https://github.com/esbmc/esbmc/issues/652) to help users create their tests with esbmc.

mikhailramalho:
I still don't get why we need this change.

PR #654, opened 2022-03-21
[output] added an option to save counterexample into a file

rafaelsamenezes:
This PR adds the option `--cex-output`, this will save the CE into a file.

PR #655, opened 2022-03-21
[options] added --file-output option to redirect messages

rafaelsamenezes:
This PR will add an option to redirect outputs of the message system into a file.


lucasccordeiro (approved):
@rafaelsamenezes: could you please add a regression test?

mikhailramalho (approved):
LGTM

fbrausse (commented):
> This PR will add an option to redirect outputs of the message system into a file.

I do not understand the reason for this change. a) the message system already outputs to the file `stdout`, b) traditionally `stderr` and `stdout` have a different purpose yet here they are mangled into one, and c) on all supported OS redirection to files is traditionally achieved via the shell. Why is it not enough?

Also I do not like the name `--file-output` of the option as it is too general. Can we find a more accurate name? E.g. `--log-redirect` would better describe the intent in my opinion.

rafaelsamenezes:
> Why is it not enough?

Because there users that do not use esbmc through a Unix Shell. In this specific case, they were running through a Gradle script, in which would be easier to add a flag than trying to redirect.

> Also I do not like the name --file-output of the option as it is too general. Can we find a more accurate name? E.g. --log-redirect would better describe the intent in my opinion.

This is not redirecting logs, it is redirecting messages. I am not sure what would be a good name though

fbrausse:
> This is not redirecting logs, it is redirecting messages. I am not sure what would be a good name though

The difference between "redirecting log" and "redirecting messages" is not entirely clear to me.

When esbmc on this branch is run with `--verbosity 7` (to enable debug output), we currently pass the flag `-v` to clang resulting in output to stderr, which should probably also go to the specified file. This could be achieved with `freopen()`.

rafaelsamenezes:
> The difference between "redirecting log" and "redirecting messages" is not entirely clear to me.

It is more conceptual really.

- We print messages through the message system, which could also have a logger, network, or gui backend. In this sense, I would consider [log](https://www.boost.org/doc/libs/1_56_0/libs/log/doc/html/index.html#log.moti) to be something clearer about the user actions and the program flow (with date, time, and OS information). 
- Right now, we only have the "file" backend, which sends to stdout and stderr. The intention of this PR is to make this backend use a file specified by the user.

> When esbmc on this branch is run with --verbosity 7 (to enable debug output), we currently pass the flag -v to clang resulting in output to stderr, which should probably also go to the specified file. This could be achieved with freopen().

Thats a bug. I will open an issue

fbrausse:
Well, by "log" I was referring to the human-readable text ESBMC prints to `stdout/stderr` out now and not some sequence of enriched and attributed messages possibly required by an analysis of software running as a daemon.¹ And with that concept in mind, the use-case you mentioned could be accomodated easily by using a wrapper script that does the intended redirection (Windows shells also support redirection via `> file 2>&1`) around the `esmbc` executable, similar to the macos-wrapper.sh in `scripts/`.

¹ If that is a goal for ESBMC, I cannot comment, yet; I'd need a more clear explanation of its purpose (reasons, problems to be solved).

rafaelsamenezes:
> And with that concept in mind, the use-case you mentioned could be accomodated easily by using a wrapper script that does the intended redirection (Windows shells also support redirection via > file 2>&1) around the esmbc executable

This is not portable at all. 

As I've said, the user is relying on a Gradle environment in which he cannot do the buffer redirection. Even using Python subprocess wouldn`t work unless you manually set the stdout/stderr or just ignore it and use the system shell through it.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -411,6 +412,18 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
 
 int esbmc_parseoptionst::doit()
 {
+  // Configure msg output
+
+  if(cmdline.isset("file-output"))
+  {
+    FILE *f = fopen(cmdline.getval("file-output"), "w+");
```
fbrausse, line 422: Is it possible to retain the coherence of lines written into this file when mixing `err` and `out` by setting it to line-buffered mode via `setvbuf(f, NULL, _IOLBF, 0)`?

*Edit*: sorry my bad, no need, you are using the same file.

fbrausse:
According to the [API](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html#org.gradle.api.tasks.Exec) it looks easy enough to handle the `standardOutput` e.g. by a [FileOutputStream](https://docs.oracle.com/javase/7/docs/api/java/io/FileOutputStream.html)...

Anyway, I still would like to suggest to rename the flag `--file-output` to something that at least refers to messages or logs, because there are other "file outputs" ESBMC can produce, with e.g. #654 adding yet another one.

Should errors still be printed to `stderr` even with this flag? If not, should the exit code be altered and/or included in the file? I'm sorry for being a little slow with understanding the semantics of this option.

rafaelsamenezes:
> According to the [API](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html#org.gradle.api.tasks.Exec) it looks easy enough to handle the standardOutput e.g. by a [FileOutputStream](https://docs.oracle.com/javase/7/docs/api/java/io/FileOutputStream.html)...

I  think we are having a miscommunication. I just said that the redirection approach is not portable and that the user cannot do that specific redirection at their flow. It is not a matter of being easy/hard to redirect it as a process.  As a side-note, they were having issues with the ammount of text that had to be redirected.

It is easy enough to add this support to esbmc anyway and we can prevent problems with any shell.

> Anyway, I still would like to suggest to rename the flag --file-output to something that at least refers to messages or logs

I agree, how about something as  `--redirect-message-output`?

> Should errors still be printed to stderr even with this flag? If not, should the exit code be altered and/or included in the file? I'm sorry for being a little slow with understanding the semantics of this option.

Throughout ESBMC we have message system which contains message handlers. 

```
MESSAGE_OBJ:
  - HANDLER_1
  - HANDLER_2
  - ...
```

The concept is that if you use something as `message_obj.status("HI!");`, it would push this message through all the handlers that are registered.

Right now, we only have one handler implemented, which is the `file_message_handler`. Which, by default, targets `stdout/stderr`. However, we can use any file for any level. During an esbmc run, we pass down a message object:

```
MESSAGE_OBJ:
  - file_message_hander
```

This patch is just configuring the default handler to use a user specified file instead of stdout/stderr. We could also add one for stdout/stderr if by some reason we want to have a colored CLI output and a basic one in a file, this patch shouldn't affect this case.

Process errors such as asserts and throws aren't to be affected. They should still go to stderr, only `msg.error` messages should be affected.

fbrausse:
Many thanks, Rafael. Maybe the miscommunication happened since the reason for requiring structured logging isn't clear.

>  As a side-note, they were having issues with the ammount of text that had to be redirected.

That's an important piece of information! Wouldn't the `--verbosity` option help? If not, is it not fine-grained enough? Anyway, there are probably reasons you chose this path, but I don't understand them. Sorry for the noise.

lucasccordeiro:
> That's an important piece of information! Wouldn't the --verbosity option help? If not, is it not fine-grained enough? Anyway, there are probably reasons you chose this path, but I don't understand them. Sorry for the noise.

An industrial partner requested this feature. The verification engineers are pretty happy with the current version of this implementation. Essentially, they have integrated ESBMC into their CI and wanted to obtain all the messages via a file while running ESBMC in their CI. I'm unsure what @fbrausse has in mind, but I would move forward with this PR if there is no major implementation flaw.

lucasccordeiro:
@rafaelsamenezes: This PR has a conflict. Could you please fix it?

fbrausse:
@lucasccordeiro what I had in mind was 1) a less general cmdline parameter name and 2) either using
```c
freopen(cmdline.get("message-redirect").c_str(), "w", stdout);
setvbuf(stdout, NULL, _IOLBF, 0); /* line buffered mode */
dup2(fileno(stdout), fileno(stderr)); /* use _dup2() on windows */
```
or just adding a script to run ESBMC with its output files redirected, e.g. `esmbc > file 2>&1`, as is already being done with the macos_wrapper (I do not see why scripts could not be OS-dependent). Either of these would also have taken care of #664 (as well as the output of the `--parse-tree-*` options).

PR #657, opened 2022-03-21
[options] added an option to not crash on non supported models

### LINKED ISSUE #656: Create an option to not crash on models extraction
When we don't support the model extraction for a specific type, we crash ESBMC. 

```cpp
msg.error(fmt::format(
      "Unimplemented type'd expression ({}) in smt get", type->type_id));
    abort();
```

Since this means that we found a bug, the violation itself can still be useful for the user. We should add an option `--non-supported-models-as-zero` which would replace this code with a `gen_zero`.

Let me know what you think

rafaelsamenezes:
Added an option to make every non supported model extraction return zero. Closes: #656 

This is a draft because I will still look for some test cases on sv-benchmarks

### DIFF
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -172,6 +172,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"bv", NULL, "use solver with bit-vector arithmetic"},
     {"ir", NULL, "use solver with integer/real arithmetic"},
     {"smtlib", NULL, "use SMT lib format"},
+    {"non-supported-models-as-zero", NULL, "if ESBMC does  as a zero"},
     {"smtlib-solver-prog",
 
      boost::program_options::value<std::string>(),
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -172,7 +172,10 @@ const struct group_opt_templ all_cmd_options[] = {
     {"bv", NULL, "use solver with bit-vector arithmetic"},
     {"ir", NULL, "use solver with integer/real arithmetic"},
     {"smtlib", NULL, "use SMT lib format"},
-    {"non-supported-models-as-zero", NULL, "if ESBMC does  as a zero"},
+    {"non-supported-models-as-zero",
+     NULL,
+     "if ESBMC can't extract a type/expression from the solver, then the value "
+     "will be set to zero"},
     {"smtlib-solver-prog",
 
      boost::program_options::value<std::string>(),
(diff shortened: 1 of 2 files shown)


src/solvers/smt/smt_conv.cpp
```diff
@@ -2275,9 +2275,19 @@ expr2tc smt_convt::get_by_ast(const type2tc &type, smt_astt a)
   }
 
   default:
-    msg.error(fmt::format(
-      "Unimplemented type'd expression ({}) in smt get", type->type_id));
-    abort();
+    if(options.get_bool_option("non-supported-models-as-zero"))
```
lucasccordeiro: Shall you add a negation here?

If the user sets `--non-supported-models-as-zero`, this method will return true.

lucasccordeiro (changes requested):

src/solvers/smt/smt_conv.cpp
rafaelsamenezes: Oops, yes!

lucasccordeiro (approved):
@rafaelsamenezes: can you please add a regression test here?

lucasccordeiro:
@rafaelsamenezes: This is a nice PR to evaluate with the SV-COMP benchmarks.

lucasccordeiro:
@rafaelsamenezes: could you please add some regression tests here?

src/esbmc/options.cpp
```diff
@@ -172,6 +172,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"bv", NULL, "use solver with bit-vector arithmetic"},
     {"ir", NULL, "use solver with integer/real arithmetic"},
     {"smtlib", NULL, "use SMT lib format"},
+    {"non-supported-models-as-zero", NULL, "if ESBMC does  as a zero"},
```
fbrausse: This message reads incomplete.

fbrausse (commented):
There are 3 get_by_* methods in smt_convt. You may also want to put this check in https://github.com/esbmc/esbmc/blob/8a30f2f28db294b6a334f332392e1fe22c82ae8b/src/solvers/smt/smt_conv.cpp#L2695

lucasccordeiro:
@rafaelsamenezes: could you please run this PR over the SV-COMP benchmarks?

rafaelsamenezes:
I've run this over sv-comp and it the only results that changed were noise. I will merge this PR.

PR #662, opened 2022-03-25
Fixed array size for multi-dimension array during flattening

### LINKED ISSUE #661: ESBMC can't deal with assumes on three (or more) dimensional arrays
Hi all,
I have the following issue (see code below). When I declare a three-dimensional non-deterministic array and put assumes on the values of its entries, ESBMC stops checking the validity of later asserts. As an example, the following code returns a successful verification result despite containing an `__ESBMC_assert(0, "");` instruction at the end.

```c
int main()
{
  int x[1][1][3];
  x[0][0][0] = nondet_int();
  x[0][0][1] = nondet_int();
  x[0][0][2] = nondet_int();
  __ESBMC_assume((x[0][0][0] >= 77) && (x[0][0][0] <= 87));
  __ESBMC_assume((x[0][0][1] >= -64) && (x[0][0][1] <= 64));
  __ESBMC_assume((x[0][0][2] >= -64) && (x[0][0][2] <= 64));
  __ESBMC_assert(0, "");
  return 0;
}
```

Additional notes:
- the same code with a two-dimensional array (i.e. `int x[1][3];`) yields a correct verification result (fail).
- the same code with only two entries in the array (i.e. `int x[1][1][2];` and corresponding assumes) yields a correct verification result (fail).
- I confirmed the results above with ESBMC 6.9, but earlier (more complex) experiments make me believe the issue may have existed since at least version 6.7
- the issue appears with a variety of ESBMC flags, or no flags at all (i.e. `~/esbmc <filename.c>`)

rafaelsamenezes:
Closes: #661


lucasccordeiro (approved):
LGTM.

kunjsong01:
commit 98776ba33 - The regressions appear to be cpp kind. 
How does fixing multi-dimentional array affect verification of cpp queue<\*> and stack<\*>? Does ESBMC lower them to array type in the backend?

rafaelsamenezes:
> How does fixing multi-dimentional array affect verification of cpp queue<*> and stack<*>? Does ESBMC lower them to array type in the backend?

Not sure. Have to look up at every level to know

lucasccordeiro:
>  Does ESBMC lower them to array type in the backend?

@kunjsong01: Could you please investigate it further since you're working on the verification of C++ programs?

kunjsong01:
> Could you please investigate it further since you're working on the verification of C++ programs?

Sure, running these regressions with vs. without this PR

rafaelsamenezes:
@kunjsong01 it was fixed for my environment only. So, it is the patch + something that only happens on my environment

PR #665, opened 2022-04-02
Add while loop grammar construct to solidity frontend

correcthorsebatterystaple:
This PR adds the while loop grammar for the solidity front end.


kunjsong01:
Could you please add a while-loop test case in src/regression and run clang-format?

lucasccordeiro:
We don’t use the iostream library anymore to print messages in the console. Please take a look at other classes (e.g., in goto-symex) to see how we print messages using the fmt library.

mikhailramalho (changes requested):
You need to fix the code style:
1. Curly brackets go into the next line
2. One-line ifs don't use curly brackets
3. Don't use iostream, use the messaget interface instead
4. the identations is two spaces, instead of tabs

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2030,3 +2052,11 @@ bool solidity_convertert::is_dyn_array(const nlohmann::json &json_in)
   }
   return false;
 }
+
+void solidity_convertert::print_json(nlohmann::json &json) {
+  std::cout << std::setw(2) << json << std::endl;
```
rafaelsamenezes: If you don't want to pass down the messaget object. You can use the `default_message` for it (since this is only a dump!).

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -11,6 +11,7 @@
 #include <regex>
 #include <util/message/format.h>
 #include <fstream>
+#include <iostream>
```
rafaelsamenezes: remove

src/solidity-frontend/solidity_convert.h
```diff
@@ -10,6 +10,7 @@
 #include <nlohmann/json.hpp>
 #include <solidity-frontend/solidity_grammar.h>
 #include <solidity-frontend/pattern_check.h>
+#include <iostream>
```
rafaelsamenezes: remove this

rafaelsamenezes (changes requested):

correcthorsebatterystaple:
Have removed print statements and methods as they were vestigial from gdb debugging

lucasccordeiro (approved):
Can you please add a test case that succeeds?

lucasccordeiro:
Could you please also rebase your PR with the master branch?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -687,6 +687,29 @@ bool solidity_convertert::get_statement(
     new_expr = if_expr;
     break;
   }
+  case SolidityGrammar::StatementT::WhileStatement:
+  {
+    exprt cond = true_exprt();
+    if(get_expr(stmt["condition"], cond))
+    {
```
rafaelsamenezes, line 694: remove brackets if the `if` contains only one instruction

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -687,6 +687,29 @@ bool solidity_convertert::get_statement(
...
+      return true;
+    }
+
+    codet body = codet();
+    if(get_block(stmt["body"], body))
+    {
```
rafaelsamenezes, line 700: remove brackets if the `if` contains only one instruction

rafaelsamenezes (changes requested):
Just some nitpicks. After that, LGTM!

lucasccordeiro:
@correcthorsebatterystaple: could you please rebase your branch with the master branch?

PR #667, opened 2022-04-05
Slicer improvements

rafaelsamenezes:
This PR:

- Adds documentation over the symex-slicer.
- Adds a couple of todos to the slicer
- Adds support to ignoring some symbols during the slicing.

Originally, the idea was to add an attribute for symbols that we'd like to be ignored. However, it was more practical to pass a list of variables that should be ignored. I think we should still add the attribute on a later patch.

To get the list of symbols, I added support to a JSON file. For now, it is only used for ignored symbols, but I'd think that it can expanded further as suggested in #477

### DIFF
--- a/src/esbmc/CMakeLists.txt
+++ b/src/esbmc/CMakeLists.txt
@@ -30,6 +30,6 @@ target_include_directories(esbmc
 
 target_link_libraries(esbmc ${OLD_FRONTEND_TARGETS} ${SOLIDITY_FRONTEND_TARGETS} clangcfrontend
   clangcppfrontend symex pointeranalysis langapi util_esbmc bigint
-  solvers clibs default_message gotoalgorithms nlohmann_json::nlohmann_json  ${Boost_LIBRARIES})
+  solvers clibs default_message gotoalgorithms nlohmann_json::nlohmann_json ${Boost_LIBRARIES})
 
 install(TARGETS esbmc DESTINATION bin)
--- a/src/esbmc/CMakeLists.txt
+++ b/src/esbmc/CMakeLists.txt
@@ -30,6 +30,6 @@ target_include_directories(esbmc
 
 target_link_libraries(esbmc ${OLD_FRONTEND_TARGETS} ${SOLIDITY_FRONTEND_TARGETS} clangcfrontend
   clangcppfrontend symex pointeranalysis langapi util_esbmc bigint
-  solvers clibs default_message gotoalgorithms ${Boost_LIBRARIES})
+  solvers clibs default_message gotoalgorithms nlohmann_json::nlohmann_json  ${Boost_LIBRARIES})
 
 install(TARGETS esbmc DESTINATION bin)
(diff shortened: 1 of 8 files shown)


src/esbmc/CMakeLists.txt
```diff
@@ -30,6 +30,6 @@ target_include_directories(esbmc
 
 target_link_libraries(esbmc ${OLD_FRONTEND_TARGETS} ${SOLIDITY_FRONTEND_TARGETS} clangcfrontend
   clangcppfrontend symex pointeranalysis langapi util_esbmc bigint
-  solvers clibs default_message gotoalgorithms ${Boost_LIBRARIES})
+  solvers clibs default_message gotoalgorithms nlohmann_json::nlohmann_json  ${Boost_LIBRARIES})
```
mikhailramalho: extra space?

src/esbmc/bmc.cpp
```diff
@@ -42,6 +42,10 @@ Authors: Daniel Kroening, kroening@kroening.com
 #include <util/migrate.h>
 #include <util/show_symbol_table.h>
 #include <util/time_stopping.h>
+#include <nlohmann/json.hpp>
+
+// For json parsing
+using json = nlohmann::json;
```
mikhailramalho: please use the scoped calls, not `using`.

src/esbmc/bmc.cpp
```diff
@@ -600,9 +604,22 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
     fine_timet slice_start = current_time();
     BigInt ignored;
     if(!options.get_bool_option("no-slice"))
-      ignored = slice(eq, options.get_bool_option("slice-assumes"));
+    {
+      std::unordered_set<std::string> ignored_symbols;
+      // TODO: This could be expanded to the whole ESBMC options and configuration system (#477)
```
mikhailramalho: this TODO seems odd since it's not really related to the code.

mikhailramalho (commented):
It looks fine but is not really implementing the attribute parsing, only the json handling.

My previous suggestion is probably very simple to implement:
1. Add an attribute to `symbolt`.
2. Change `clang_c_convertert::get_var` to parse the annotation and store the symbol.
3. Change `goto_symext::symex_assign_symbol` (where all symbol assignments eventually pass through) to read the symbol from the contextt, check the flag and do something like:
```
  bool no_slice = contextt.find(lhs).no_slice;

  // do the assignment
  target->assignment(
    tmp_guard.as_expr(),
    renamed_lhs,
    new_lhs,
    rhs,
    full_rhs,
    cur_state->source,
    cur_state->gen_stack_trace(),
    hidden || no_slice,                        // <- so we always override even if esbmc tries to slice it
    first_loop);
```

rafaelsamenezes:
> t looks fine but is not really implementing the attribute parsing, only the json handling.

Yes, I will also add the attribute into another PR. This PR is for when the user can't change the original source.

fbrausse:
Could we do the config-file in another PR? I am happy about the functional changes to allow overriding the slicer in a fine-grained way per symbol. Attribute is fine, user-specified is also fine. Even #781 could possibly make use of that by putting the attribute on `__ESBMC_alloc`. But to force the use of a JSON file for this functionality seems overkill at the moment. Why not just give this information via the cmdline? It could be a vector of symbol names. If users want to use a file, there is a pretty commonly used way to achieve that by passing a parameter `@file`, e.g. by the linker [ld](https://linux.die.net/man/1/ld).

rafaelsamenezes:
PR #814 supersedes this PR

PR #668, opened 2022-04-05
Changed constant_p return type

rafaelsamenezes:


fbrausse (commented):
Not sure whether it's required, but maybe also change `true`/`false` to `1`/`0` here: https://github.com/esbmc/esbmc/blob/58dedc85c15239e682a2a774ca04de92edef88ae/src/goto-symex/symex_main.cpp#L559

PR #670, opened 2022-04-07
[z3] Improved error message printing in Z3 error handler

kunjsong01:
### Summary: 
This patch improved Z3 error handler to print out more informative error message. 

### Error message printing *Before* the change:
```
Thread 0 file main.c line 4 function main
ASSIGNMENT ()
f?1!0&0#1 == (float)d?1!0&0#1

esbmc: ../src/solvers/z3/z3_conv.cpp:21: void error_handler(Z3_context, Z3_error_code): Assertion `0 && oss.str().c_str()' failed.
```
### Error message printing *After* the change:
```
Thread 0 file main.c line 4 function main
ASSIGNMENT ()
f?1!0&0#1 == (float)d?1!0&0#1

terminate called after throwing an instance of 'z3::exception'
  what():  Sorts Real and (_ FloatingPoint 8 24) are incompatible
Aborted (core dumped)
```

Now we can see the root cause of Z3 equation conversion - incompatible sorts.

### DIFF
--- a/src/solvers/z3/z3_conv.cpp
+++ b/src/solvers/z3/z3_conv.cpp
@@ -19,9 +19,7 @@ static void error_handler(Z3_context c, Z3_error_code e)
       << "\n";
   oss << Z3_get_error_msg(c, e);
   default_message msg;
-  // use msg.status + abort instead of msg.error
-  // In GDB, the latter does not show a backtrace of the exact failure line in this function
-  msg.status(oss.str());
+  msg.error(oss.str());
   abort();
 }
 
--- a/src/solvers/z3/z3_conv.cpp
+++ b/src/solvers/z3/z3_conv.cpp
@@ -12,13 +12,19 @@
 
 #define new_ast new_solver_ast<z3_smt_ast>
 
+#if !defined(Z3_THROW)
+#if __cpp_exceptions || _CPPUNWIND || __EXCEPTIONS
+#define Z3_THROW(x) throw x
+#else
+#define Z3_THROW(x)                                                            \
+  {                                                                            \
+  }
+#endif
+#endif // !defined(Z3_THROW)
+
 static void error_handler(Z3_context c, Z3_error_code e)
 {
-  std::ostringstream oss;
-  oss << "Z3 error " << e << " encountered"
-      << "\n";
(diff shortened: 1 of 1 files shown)


mikhailramalho (commented):
You can just print the message on `oss.str().c_str()` using our message interface and call `abort()`. Can you check if it works?

lucasccordeiro (approved):

kunjsong01:
> Can you check if it works?
Done. The error message printed by our msg interface is as following: 
```
Z3 error 3 encountered
fp sort expected

Program received signal SIGABRT, Aborted.
```

It shows the error code and preserves the root cause. In GDB the backtrace also shows the exact location of failure, which is debugging friendly.

mikhailramalho (approved):

src/solvers/z3/z3_conv.cpp
```diff
@@ -18,7 +18,11 @@ static void error_handler(Z3_context c, Z3_error_code e)
   oss << "Z3 error " << e << " encountered"
       << "\n";
   oss << Z3_get_error_msg(c, e);
-  assert(0 && oss.str().c_str());
+  default_message msg;
+  // use msg.status + assert instead of msg.error
+  // In GDB, the latter does not show a backtrack of the exact failure line in this function
```
fbrausse: I believe you meant `abort` instead of `assert`. Why does msg.error not work?

kunjsong01: Done. 

> Why does msg.error not work?

Given that `error_handler` is a callback function we passed into Z3_set_error_handler and `msg.error` does not exit on failure, my conjecture is that if we don't exit in `error_handler` ESBMC will continue in some Z3 library routine. However, GDB may not have the Z3 debugging symbols to map to the corresponding function/line in the source file.

kunjsong01: Changed to msg.error + abort and removed the comments. 

Sorry about the confusing comments. Previously, I had always been mistaken about `msg.error` aborting the execution automatically. The comment actually means using msg.error **alone** would not make the program abort and hence not showing a clear backtrace of the exact line.

rafaelsamenezes (approved):

fbrausse (approved):

Issue #677, opened 2022-04-14
Casting pointer through integer fails


rafaelsamenezes:
The following program:

```c
#include <inttypes.h>

#define _4KB        0x1000
#define _32KB       0x8000
 
#define _4KB_MASK   (~(_4KB - 1))

#define UINT32 uint32_t 
#define UINT64 uint64_t
void Test()
{
  void *Heap = __ESBMC_alloca(_32KB);
  __ESBMC_assume((UINT64)Heap < 0xFF); 
  
  UINT64 *TabPhy = (UINT64*)(((UINT64)Heap + _4KB) & _4KB_MASK);
  UINT64 *TabPhy_Entry = (UINT64*)(((UINT64)TabPhy + _4KB) & _4KB_MASK);
  
  TabPhy[0] = ((UINT64)TabPhy_Entry);  
  __ESBMC_assert(TabPhy == &TabPhy[0], "i-st of TabPhy is TabPhy[0]");
  int i = 0;
  __ESBMC_assert((*(UINT64 *)TabPhy) == TabPhy[i], "value of TabPhy is TabPhy[i] value");
}
```

Fails on the last assertion, even though ESBMC knows that they are pointing to the same place. There are also two strange behaviors:

1. If we remove the mask `& _4KB_MASK` then the assertion does not fail.
2. If we replace `TabPhy[i]` with `TabPhy[0],` then the assertion does not fail.

@fbrausse could you take a look into it?

fbrausse:
@rafaelsamenezes sure, I'll take a look.

> 1. If we remove the mask & _4KB_MASK then the assertion does not fail.

When you remove the `& _4KB_MASK` expressions the only ones left are additions, right? Support for `+` and `-` in the value-set analysis I added some time ago, but not other expressions. I agree that support for bit-ops should be added around https://github.com/esbmc/esbmc/blob/ee09a075517a6643ec3867710b65ce3c45e2f190/src/pointer-analysis/value_set.cpp#L549

> 2. If we replace TabPhy[i] with TabPhy[0], then the assertion does not fail.

In #575 we had the same issue.

mikhailramalho:
Maybe duplicate of #92?

fbrausse:
> Maybe duplicate of #92?

[...]

PR #679, opened 2022-04-15
[value-set] WIP: tentative fix for #677 (needs more work)

fbrausse:


lucasccordeiro:
@rafaelsamenezes: did you have a chance to run this PR over the sv-comp benchmarks?

fbrausse:
@lucasccordeiro this PR is not final, yet: the other binary operations should be addressed as well, including `xor2t`. There are also connections to #696. I'm looking into a solution for these.

PR #680, opened 2022-04-17
Fix/state var init

correcthorsebatterystaple:
Fix state variable initialization throwing errors because of a faulty key access on the AST


lucasccordeiro (changes requested):
@correcthorsebatterystaple: could you please add a regression test that produces `VERIFICATION FAILED`?

lucasccordeiro:
BTW, the recently added regression test fails in our CI:

```
The following tests FAILED:
	727 - regression/esbmc-solidity/github_state_init (Failed)
Errors while running CTest
```

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -213,22 +215,22 @@ bool solidity_convertert::get_var_decl(
...
+  bool is_state_var = ast_node["stateVariable"] == true;
+
   // 2. populate id and name
   std::string name, id;
-  if(ast_node["stateVariable"] == true)
+  if(is_state_var)
     get_state_var_decl_name(ast_node, name, id);
+  else if(current_functionDecl)
+  {
+    // location variable in function body
```
kunjsong01: local variable in function body?

regression/esbmc-solidity/github_state_init/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+contract.solast
```
kunjsong01: The JSON AST file is missing in the commit.

kunjsong01 (changes requested):

regression/esbmc-solidity/github_state_init/test.desc
```diff
@@ -0,0 +1,4 @@
...
+--function state_var_init --contract contract.sol
+^VERIFICATION SUCCESSFUL$
```
kunjsong01: Would it be possible to add another test case that reports "VERIFICATION FAILED"?

kunjsong01 (approved):

lucasccordeiro (approved):

PR #682, opened 2022-04-17
Feature/add bool literal

correcthorsebatterystaple:
Add boolean literal analysis for solidity frontend

### DIFF
--- a/regression/esbmc-solidity/github_bool_literal/test.desc
+++ b/regression/esbmc-solidity/github_bool_literal/test.desc
@@ -1,4 +1,0 @@
-CORE
-contract.solast
---function foo --contract contract.sol
-^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-solidity/github_bool_literal/test.desc
+++ b/regression/esbmc-solidity/github_bool_literal/test.desc
@@ -1,0 +1,4 @@
+CORE
+contract.solast
+--function foo --contract contract.sol
+^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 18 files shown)


src/solidity-frontend/solidity_convert_literals.cpp
```diff
@@ -38,6 +39,29 @@ bool solidity_convertert::convert_integer_literal(
...
+  std::string the_value,
+  exprt &dest)
+{
+  typet type;
+  if(get_type_description(bool_literal, type))
+    return true;
+
+  assert(type.is_bool());
+
+  if(the_value == "true")
```
mikhailramalho, line 53: This is correct but to make it consistent with the code style I suggest:
```
  if(the_value == "true")
  {
    dest = true_exprt();
    return false;
  }
  
  if(the_value == "false")
  {
    dest = false_exprt();
    return false;
  }

  return true;
}
```

lucasccordeiro (changes requested):
@correcthorsebatterystaple: could you also please add more regression tests (e.g., when y is `false`)?

regression/esbmc-solidity/github_bool_literal/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+contract.solast
+--function foo --contract contract.sol
+^VERIFICATION SUCCESSFUL$
```
kunjsong01: Would it be possible to add another regression test that reports "VERIFICATION FAIL"? 
Ideally, we woud like two test cases one reporting successful and the other reporting failed.

kunjsong01 (changes requested):

kunjsong01 (approved):

PR #683, opened 2022-04-20
[clang-c-frontend] Add padding for _ExtInt struct fields

georgerennie:
This adds support for padding/aligning _ExtInt fields in structs as discussed in #673.

This is implemented by adding anonymous padding fields after the _ExtInt fields to pad them out to the width of the least integer type that can contain them. There are some flaws to this approach currently, but I don't have the time free to work on addressing them.

The flaws arise because although the fields are padded in structs, the internal representation of ExtInts only stores their width, not the representation width. This is primarily an issue for pointer analysis, where I found the following issues:

- Dereferencing an element from an array of _ExtInts doesnt align elements correctly. The following example fails when it should succeed:
```c
#include <assert.h>
int main() {
	_ExtInt(9) arr[10];
	assert(*&arr[2] == arr[2]);
}
```

- Pointer analysis expects _ExtInt struct fields above 8 bytes in size to be aligned to the next power of 2 bytes, instead of to just 8 bytes (e.g. pointer analysis expects _ExtInt(120) to be aligned to 16 bytes, whereas _ExtInts should have a maximum alignment of 8 bytes). This is a fairly complex case though, as the proposed _BitInt spec also suggests that _BitInt/_ExtInts above 8 bytes should be represented as a struct of individual 8 byte fields

As these semantics are implementation defined behaviour anyway, I think it makes most sense to ensure that derefering direct references (e.g. `*&object`) works more than dereferencing calculated ones (e.g. `*(ptr + 2)`)


mikhailramalho (commented):
Interesting work!

What happens in cases where the `sizeof(_ExtInt(X)) == sizeof(int)` and other native types? Are these handled like native types?

I guess this is a good enough implementation until we have a proper standard definition. @rafaelsamenezes can you run this PR on SV-Comp benchmarks to make sure that we are not introducing new regressions?

georgerennie:
> What happens in cases where the sizeof(_ExtInt(X)) == sizeof(int)?

_ExtInts are being represented using (un)signedbv types internally, and so the only thing distinguishing them from native types as far as I am aware is that they have a `#extint` attribute on them introduced in this PR. This is just used to determine that they need to be padded in struct fields (you could theoretically do away with this attribute and just pad all (un)signedbv types regardless, because native integers would just not require any padding).

mikhailramalho:
> > What happens in cases where the sizeof(_ExtInt(X)) == sizeof(int)?
> 
> _ExtInts are being represented using (un)signedbv types internally, and so the only thing distinguishing them from native types as far as I am aware is that they have a `#extint` attribute on them introduced in this PR. This is just used to determine that they need to be padded in struct fields (you could theoretically do away with this attribute and just pad all (un)signedbv types regardless, because native integers would just not require any padding).

Got it, but I wanted to know if `sizeof(_ExtInt(X)) == sizeof(int)`, should the alignment/padding rules fallback to native types or should we always default to the 8 bytes alignment you mentioned before?

georgerennie:
Ahh so the alignment rules for ExtInts should generally be the same as for the native integers. They are represented in memory the same as the smallest integer type that can contain them. If they are wider than the largest integer type (`chunk_t`), the clang proposed abi represents then as a struct containing an array of the least number of `chunk_t`s required to hold the whole of the ExtInt.

>  The element at index ``Idx`` stores bits ``Idx * ChunkWidth`` through ``(Idx + 1) * ChunkWidth``

This means that these large ExtInts are stored as a little endian array of `chunk_t`s, although the endianness of the individual elements is the same as for the rest of the ints in the target architecture. Therefore the alignment of these large ExtInts is the same as `chunk_t`. I believe that the largest integer supported for most common architectures is 8 bytes (a long long), so therefore `chunk_t` would be an 8 byte integer.

This pull request doesn't implement these semantics for large ExtInts, instead just representing them as the nearest power of 2 bytes that can hold them, as this seems to be what pointer analysis currently expects.

I found an issue in my logic (was reading the wrong variable by mistake) while testing this just now so have fixed it.

mikhailramalho:
Well, since this is still a proposal I think we shouldn't mess with pointer analysis and risk having to change it twice.

That being said, I'm cool with this change as long as it doesn't add any new regression in SV-COMP. @lucasccordeiro, @rafaelsamenezes could you verify the SV-COMP benchmarks using this PR?

rafaelsamenezes:
This PR didn't cause any regression on SV-Comp (unreach-call).

Total: 9213
Correct True: 4177
Correct False: 1330
Incorrect True: 20
Incorrect False: 10

Score: 6224

lucasccordeiro:
@georgerennie: many thanks for contributing to our codebase!

PR #685, opened 2022-04-24
[clang-cpp-frontend] Added symbol mode

kunjsong01:
This PR fixed the symbol mode in clang-c/cpp-frontend. When verifying C++ programs, the symbol mode should be "C++", not "C". 

### Symbol printing before the fix:
```
Symbol......: c:@S@Test@F@increment_x#I#
Module......: main
Base name...: increment_x
Mode........: C (0)
``` 
### Symbol printing after the fix:
```
Symbol......: c:@S@Test@F@increment_x#I#
Module......: main
Base name...: increment_x
Mode........: C++ (1)
```

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -28,11 +28,13 @@
 clang_c_convertert::clang_c_convertert(
   contextt &_context,
   std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs,
+  const std::string &_mode,
   const messaget &msg)
   : ASTContext(nullptr),
     context(_context),
     ns(context),
     ASTs(_ASTs),
+    symbol_mode(_mode),
     msg(msg),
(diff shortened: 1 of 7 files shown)


mikhailramalho:
I actually removed that a few years ago, because there was no advantage in
having a different mode between C/C++ (I'll look for the commit when I get
home).

Maybe it's useful for linking somehow? If we can't find a need for 
it, I don't think that should be changed.

kunjsong01:
> Maybe it's useful for linking somehow? 

Nope, it's a purely cosmetic change. No functional benifits.  

> If we can't find a need for it, I don't think that should be changed.

Agreed. Actually, the intrinsics are still in "C" mode. A mix of "C" and "C++" modes in symbol table is probably more confusing to the users.

kunjsong01:
Based on the [above discussion](https://github.com/esbmc/esbmc/pull/685#issuecomment-1107880643), this PR is abandoned.

Issue #687, opened 2022-04-24
Missing case in smt_convt::convert_terminal to deal with expr2t::member_id


kunjsong01:
When converting a terminal, there is no case in our SMT converter to deal with `expr2t::member_id`. 

### Error message:
```
### ERROR:
Converting unrecognized terminal expr to SMT
```

### Preliminary investigations:
```
(gdb) p expr->expr_id
$5 = expr2t::member_id
(gdb) p expr.get()->dump()
member
* source_value : symbol
    * name : c:@real_if_function1_inlined_1
    * renamelev : Level 2 (global)
    * level1_num : 0
    * level2_num : 4
    * thread_num : 0
    * node_num : 0
    * type : struct
      * members :
        0: floatbv
          * fraction : 23
          * exponent : 8
        1: signedbv
          * width : 32

      * member_names :
        0: x
        1: z

      * member_pretty_names :
        0: x
        1: z

      * typename : __real_if_function
      * packed : false
* member_name : x
* type : floatbv
  * fraction : 23
  * exponent : 8
$6 = void
```

### Info to Re-produce the Bug:
ESBMC mainline commit ee09a0755. 

Test case: 
```c
#include <stdbool.h>
typedef struct {
	float x;
	bool z;
} myStruct;

int main(void) {
  myStruct testObj;
  testObj.x = 0.0;
  testObj.z = false;

  if (testObj.x > 10.0)
  {
    testObj.z = true;
  }

  assert(testObj.z); // fail
  return 0;
}
```

Command lines used: 
```
./esbmc main.c --ir --z3
``` 

This bug was revealed by commit e4dca2c0d30d827, after we updated the SMT conversion flow to encode the original fp term in real mode when using "--ir" and "--floatbv" together.

lucasccordeiro:
@kunjsong01: I believe @fbrausse could help you with this issue.

kunjsong01:
I've replaced the original test case with a much simpler one that is also capable of reproducing the bug.

### More investigations: 
```
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#1 == (double)testObj?1!0&0#2.x > 10.000000
```  
In the above SSA statement, ESBMC doesn't know how to encode struct member typecasting `(double)testObj.x` using real/int mode. Previously, our patch e4dca2c0d3 just fixeed the scalar variable typecasting encoding, e.g. `(double)x` where x is a float. 

[...]

PR #689, opened 2022-04-25
Fixes for `--ir`

fbrausse:
This PR
- fixes the crash reported in #687 by recursing using the general `convert_ast()` instead of `convert_terminal()` for `--ir --floatbv`, and
- avoids crashing in the final counter-example stage by opting out of interpreting `--ir --floatbv` values (rationals/reals) as `ieee_floatt`. This is of course not a real solution, but at least does not crash. I'd be happy to find a "real solution" once I understand the `--ir` semantics better.

### DIFF
--- a/regression/esbmc/github_687-1/main.c
+++ b/regression/esbmc/github_687-1/main.c
@@ -1,0 +1,253 @@
+#include <stdbool.h>
+#include <stdint.h>
+#include <assert.h>
+#include <math.h>
+
+// Struct definitions
+typedef struct {
+} __VerificationLoopDS;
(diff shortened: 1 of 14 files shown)


lucasccordeiro:
@fbrausse: could you please add some regression tests here?

fbrausse:
> @fbrausse: could you please add some regression tests here?

I've added the ones from #687 as well as from @kunjsong01's PR #690 about the same issue.

lucasccordeiro (approved):

lucasccordeiro:
Thanks, @fbrausse.

PR #696, opened 2022-05-05
Pointer arithmetic for non-pointer operands

### LINKED ISSUE #695: Pointer arithmetic without explicit casting fails
ESBMC fails with the following program:

```c
#define PRESENT  0x01
#define _4KB        0x1000
#define _32KB       0x8000

#define _4KB_MASK   (~(_4KB - 1))
#include <assert.h>
#define UINT32 unsigned long
#define UINT64 unsigned long long

int main() {
    void *Heap = __ESBMC_alloca(_32KB);
    __ESBMC_assume((UINT64)Heap < 0xFF);

    UINT64 *TabPhy = (UINT64*)(((UINT64)Heap + _4KB) );
    UINT64 var = (UINT64)Heap + _4KB;

    for (int i = 0; i < 4; i++)
    {
        UINT64 current_value = i*sizeof(UINT64);
        *((UINT64*)(current_value + var)) = i;
    }

    for (int i = 0; i < 4; i++)
        assert(TabPhy[i] == i);

    return 0;
}
```

At the first `for` loop if we change `  *((UINT64*)(current_value + var)) = i;` to:
1.  ` *((UINT64*)(current_value + (char*)var)) = i;`. The program is verified correctly
1.  ` *((UINT64*)((char*)current_value + var)) = i;`. The program still fails

rafaelsamenezes:
This PR should close #695


src/pointer-analysis/value_set.cpp
```diff
@@ -569,12 +569,32 @@ void value_sett::get_value_set_rec(
...
-      get_value_set_rec(op0, op0_set, "", op0->type, false);
-
     object_mapt op1_set;
-    if(!is_pointer_type(op0))
-      get_value_set_rec(op1, op1_set, "", op1->type, false);
+    get_value_set_rec(op0, op0_set, "", (op0)->type, false);
+    get_value_set_rec(op1, op1_set, "", (op1)->type, false);
+
+    /*
+     * The main idea here is to be able to detect if any address might be valid at all.
```
lucasccordeiro: "The main idea here is to detect if any address might be valid at all. If any side of the pointer arithmetic is not a valid object, we can assume that it is not the pointer operand."

src/pointer-analysis/value_set.cpp
```diff
@@ -569,12 +569,32 @@ void value_sett::get_value_set_rec(
...
+     * If any side of the pointer arithmetic is not a valid-object, then we can
+     * assume that it is not the pointer operand.
+     *
+     * As of now, there is one KNOWNBUG (esbmc/github_659_2_true), which is triggered by
```
lucasccordeiro: Should you open an issue for this knownbug?

lucasccordeiro (changes requested):

lucasccordeiro:
@rafaelsamenezes: do not forget to run this PR over the SV-COMP benchmarks.

rafaelsamenezes:
> @rafaelsamenezes: do not forget to run this PR over the SV-COMP benchmarks.

Running on: https://github.com/esbmc/esbmc/actions/runs/2274876929

mikhailramalho (approved):
LGTM! Can you just run on SV-COMP before merging?

src/pointer-analysis/value_set.cpp
```diff
@@ -569,12 +569,34 @@ void value_sett::get_value_set_rec(
...
+      {
+        auto expr = object_numbering[it.first];
+        if(is_symbol2t(expr) || is_constant_string2t(expr))
+          return true;
+      }
+
+      return false;
+    };
+
+    if(!op_contains_valid_address(op0_set))
```
fbrausse: I believe ignoring the value-set of an operand based on non-presence of something is not sound: in particular, in the KNOWNBUG case the value-set of a `mul2t` expression is being built. As there is no special handling for that (yet), an `unknown2t` expression is put in there. By virtue of being unknown, it could also point to a symbol or constant string...

src/pointer-analysis/value_set.cpp
```diff
@@ -569,12 +569,34 @@ void value_sett::get_value_set_rec(
...
+     *
+     * As of now, there is one KNOWNBUG (esbmc/github_659_2_true), which is triggered by
+     * a cast of the index as char ptr. This makes ESBMC think that both operands are
+     * valid, which is currently unsupported.
+     */
+    auto op_contains_valid_address = [this](value_sett::object_mapt o) -> bool {
+      for(const auto &it : o)
+      {
+        auto expr = object_numbering[it.first];
+        if(is_symbol2t(expr) || is_constant_string2t(expr))
```
fbrausse: Are symbols and constant strings the only kind of expressions that constitute "valid addresses"? What about constant structs or unions? *Edit*: or arrays or array_ofs.

fbrausse (changes requested):
Nice seeing this progress, but I believe this change also contains a case of unsoundness, see comments below.

In connection with #677 (*edit*: and the attempt #679), I believe we might need to adapt this code path to all binary arithmetic. IMO the as-of-yet unhandled case that both operands' value-sets are non-empty, a kind of cartesian product would need to get analyzed.

rafaelsamenezes:
>  LGTM! Can you just run on SV-COMP before merging?

Running at: https://github.com/esbmc/esbmc/actions/runs/2280965649

src/pointer-analysis/value_set.cpp
rafaelsamenezes: This is an issue of not being complete, right? The value-set not being handled properly by multiplication causes it to propagate unknown.

`unknown` by itself can be anything. So, it is sound to throw an error.

src/pointer-analysis/value_set.cpp
rafaelsamenezes: Hm, I am not sure. I will add more tests for complex data-types.

rafaelsamenezes:
> IMO the as-of-yet unhandled case that both operands' value-sets are non-empty, a kind of cartesian product would need to get analyzed.

Yes, I am still not sure how to handle it though. We should produce some complex examples of it and open a new issue for this.

src/pointer-analysis/value_set.cpp
rafaelsamenezes: @fbrausse I've added `github_695_3` and `github_695_4` which uses arrays of struct and unions. Is that what you meant?

fbrausse:
> > IMO the as-of-yet unhandled case that both operands' value-sets are non-empty, a kind of cartesian product would need to get analyzed.
> 
> Yes, I am still not sure how to handle it though. We should produce some complex examples of it and open a new issue for this.

Agreed, does not have to be in this PR.

src/pointer-analysis/value_set.cpp
fbrausse: > This is an issue of not being complete, right? The value-set not being handled properly by multiplication causes it to propagate unknown.

Yeah, `unknown` as a fallback is fine if we don't handle it properly.

> `unknown` by itself can be anything. So, it is sound to throw an error.

Not sure what you mean by "throw an error". Where is that done? If `unknown` is in one operands value-set, I guess the easiest sound thing to do would be to put `unknown` in the entire expression's value-set.

But that's not done at the moment: the `unknown` in either operand's value-set is ignored, if it does **also** not contain whatever constitutes a "valid address".

rafaelsamenezes: > Not sure what you mean by "throw an error".

Sorry, I meant at the KNOWNBUG example.

> If unknown is in one operands value-set, I guess the easiest sound thing to do would be to put unknown in the entire expression's value-set.

If the operand is the pointer operand, then yes we should propagate the unknown (which is what happens). If both operands do/don't have any valid place to go, then it will be unknown.

> But that's not done at the moment: the unknown in either operand's value-set is ignored, if it does also not contain whatever constitutes a "valid address".

This will be an issue only if the value-set has failed before. Even NULL pointer is considered a "valid address". I would guess that the main limitation is that if you have something like: `(int)(&var) + (int*)(nondet_int()))` or `nondet_int() + (int*)(nondet_int()))`  which will/should only propagate unknown.

fbrausse: Maybe we're talking about different things? This code https://github.com/esbmc/esbmc/blob/45f57c96ad26ae4bb198c5a82d17b0bb11c66603/src/pointer-analysis/value_set.cpp#L596-L599 will remove any possibly existing `unknown` in the corresponding operand's value-set based on non-existance of something else in that set as determined by `op_contains_valid_address` ("valid address" above). That is what I perceive as unsound, since if that something else does not exist, the `unknown` will be dropped. Whether or not this also happens for the other operand is irrelevant for that statement.

Now, if `unknown` gets dropped here, then we're claiming that this operand does not refer to anything and correspondingly whatever it might have been referring to (`unknown`) will not be part of the formula but instead [*edit*: ~will~ may] be encoded as an offset.

src/pointer-analysis/value_set.cpp
fbrausse: Hm, partially. For constant array I rather meant something like [gh-695_5_true.c.txt](https://github.com/esbmc/esbmc/files/8640281/gh-695_5_true.c.txt) where the `op0_set` now also gets emptied.

src/pointer-analysis/value_set.cpp
rafaelsamenezes: > Maybe we're talking about different things? 

We aren't. But we still do not understand each other. `valid-address`, `value set` and `unknown` are too abstract. I will try to say what I understood from you:

I am assuming the following:
1. `unknown` is the same thing as nondet. Which, as you said, might be a valid address to any place.
2. Operations between two valid pointers does not make sense. It should be solved to an unknown (or error).
3. Operations between two invalid pointers does not make sense. It should be solved to an unknown (or error).
4. Valid address means a symbol or string. Which includes NULL pointer 

From those, 
1.  If we are trying to dereference it, then it will be a dereference error or a nondeterministic value that we cannot reason about it, e.g. `A = *(unknown_0); B = *(unknown_0); (A == B) fails`. *Is this what you mean?* 
2. This should be solved by a later PR, as it can be related to phi-operations.
3. This is already propagated as unknown.

Now, from what I understood from your comment is that you have one operator that the value-set know that it might be a valid-adress with a nondeterministic one is dropping the nondeterministic. Which I don't think it is, as doing the operation as an offset will cause it to be unknown.

The other flow is when both aren't valid, which will also propagate the unknown.

If this isn't what you meant. Then I will need a concrete example to be able to visualize.

fbrausse: > `valid-address`, `value set` and `unknown` are too abstract.

- You introduced a small lambda and called it `op_contains_valid_address`. The condition `is_symbol2t(expr) || is_constant_string2t(expr)` in the check it performs is what I'm referring to, "valid address". Sure, this predicate is going to return `true` for value-sets containing `NULL` symbols, I know, and that's fine. But it will ignore `unknown2t` (and `invalid2t`) members of its argument, while I believe it should not.
- By `value-set (of an expression)` I am referring to what `get_value_set()` should compute in order for ESBMC to correctly reason about dereferencing. In particular, no possible pointed-to targets of the (however casted/arithmetically constructed) expression should be missed - unless `unknown2t` or `invalid2t` are elements of that set.
- `unknown` is a short-hand for `an unknown2t or invalid2t expression`.

> 1. If we are trying to dereference it, then it will be a dereference error or a nondeterministic value that we cannot reason about it, e.g. `A = *(unknown_0); B = *(unknown_0); (A == B) fails`. *Is this what you mean?*

Not quite, let's put this example in a `add2t` or `sub2t` context: `uintptr_t x, A = &x, B = (unknown_0); *(A+B) = 42`. This PR will call `op1_set.clear()` and then compute the value-set of `A+B` as the singleton `{ (x at an offset) }`. It is my understanding that instead it should either be the two-element set `{ (x at an offset), unknown2t }` or just `{ unknown2t }`. Is that understanding correct? (Related: the `plus-dlist.c` example in https://github.com/esbmc/esbmc/issues/677#issuecomment-1100605221 where expressions of the form `(&a+&b)-&a` result in addresses to unrelated to `a`.)

If so, then either the `op_contains_valid_address` predicate should not ignore `unknown2t` and `invalid2t` expressions or clearing the entire `op*_set` (which also removes these "placeholders" `unknown2t` and `invalid2t`) is too strong.

fbrausse: > I am assuming the following:
> 1.    unknown is the same thing as nondet. Which, as you said, might be a valid address to any place.
> 2.    Operations between two valid pointers does not make sense. It should be solved to an unknown (or error).
> 3.    Operations between two invalid pointers does not make sense. It should be solved to an unknown (or error).
> 4.    Valid address means a symbol or string. Which includes NULL pointer

I believe 2. and 4. are not true.

2: see `plus-dlist.c` from the comment mentioned above or just consider `a+b` where `void *c = NULL; intptr_t b = (intptr_t)c`. The C standard defines `(void *)0` to be a null pointer constant. `NULL` is defined by the c2goto library exactly as that expression.

4: see the types mentioned in my 2nd remark to this PR, e.g. the constant array example in comment about the condition in `op_contains_valid_address` below.

rafaelsamenezes:
SV-Comp results:

Master

```
  correct:            8155
    correct true:     5078
    correct false:    3077
  incorrect:            62
    incorrect true:     24
    incorrect false:    38
  unknown:            7431
  Score:             11857 (max: 25567)
```

This branch:
```
  correct:            8147
    correct true:     5068
    correct false:    3079
  incorrect:            64
    incorrect true:     24
    incorrect false:    40
  unknown:            7437
  Score:             11807 (max: 25567)
```

mikhailramalho:
Are the new wrong results a result of the patch or is it noise?

Are the new unknowns crashes or timeout?

lucasccordeiro:
@rafaelsamenezes: could you please answer @mikhailramalho's questions?

It would be great if we could progress with this PR.

src/pointer-analysis/value_set.cpp
fbrausse: In light of the recent discussion about compound literals, I suppose my example above does not need to be supported in this PR.

However, I would still like to understand why `is_symbol2t(expr) || is_constant_string2t(expr)` is the complete condition for `expr` to constitute a "valid address". Phrased differently: what is special about `constant_string2t` that does not hold for other non-symbol expressions?

lucasccordeiro:
@rafaelsamenezes: there was a huge discussion in this PR. You have addressed all comments, but we ended up not merging them.

Can I ask you whether you could rebase it?

lucasccordeiro:
SV-COMP run (30s) here: https://github.com/esbmc/esbmc/actions/runs/17698511564

pmatos:
@rafaelsamenezes — revisiting this during a stale-PR sweep.

Good news first: the core patched region in `value_set.cpp` is unchanged on `master`, so the change still applies cleanly, and #695 is still open (not superseded). Before it can land, though, a few things from the original review still need resolving:

- **Soundness (from @fbrausse):** `op_contains_valid_address` drops an operand's value-set — including any `unknown2t` / `invalid2t` members — when it doesn't find a symbol / constant-string / constant-array. That can under-approximate and hide real pointer targets. The sound fallback would be to propagate `unknown` into the whole expression's value-set instead of discarding it.
- **Design:** what exactly should count as a "valid address"? The current set (symbol / constant_string / constant_array) was questioned in review and never fully justified.
- **SV-COMP delta:** the self-reported regression (correct 8155 → 8147, incorrect 62 → 64) was never explained — worth a fresh SV-COMP run on current `master`.
- **CI / tests:** CI is currently red (build-unix + `static llvm-16 DebugOpt`), and there's a lingering `KNOWNBUG` test (`github_695_2`).

Is this something you'd like to pick back up? Happy to help review if so; otherwise we can close it and keep #695 as the tracker.

src/pointer-analysis/value_set.cpp
```diff
@@ -582,12 +582,37 @@ void value_sett::get_value_set_rec(
...
+     * As of now, there is one KNOWNBUG (esbmc/github_695_2_true), which is triggered by
+     * a cast of the index as char ptr. This makes ESBMC think that both operands are
+     * valid, which is currently unsupported.
+     */
+    auto op_contains_valid_address = [this](value_sett::object_mapt o) -> bool {
+      for (const auto &it : o)
+      {
+        auto expr = object_numbering[it.first];
+        if (
+          is_symbol2t(expr) || is_constant_string2t(expr) ||
```
lucasccordeiro, line 603: `dynamic_object2t` is missing, so a malloc/alloca pointee is treated as invalid and its set is cleared — an under-approximation. Add it.

src/pointer-analysis/value_set.cpp
```diff
@@ -582,12 +582,37 @@ void value_sett::get_value_set_rec(
...
+    /*
+     * The main idea here is to detect if any address might be valid at all. If any side
+     * of the pointer arithmetic is not a valid object, we can assume that it is not the
+     * pointer operand.
+     *
+     * As of now, there is one KNOWNBUG (esbmc/github_695_2_true), which is triggered by
+     * a cast of the index as char ptr. This makes ESBMC think that both operands are
+     * valid, which is currently unsupported.
+     */
+    auto op_contains_valid_address = [this](value_sett::object_mapt o) -> bool {
```
lucasccordeiro, line 598: `object_mapt` is taken by value, copying the whole map on every call in a hot path. Take it by const reference.

regression/esbmc/github_1590-generic/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: this PR breaks a passing test. Fix the value-set change rather than disabling it.

regression/esbmc/github_2512_11/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: a second passing test is disabled here. It must keep failing as expected.

lucasccordeiro (changes requested):

src/pointer-analysis/value_set.cpp
```diff
@@ -582,12 +582,37 @@ void value_sett::get_value_set_rec(
     // Find out what the pointer operand points at, and suck that data into
     // new object maps.
     object_mapt op0_set;
-    if (!is_pointer_type(op1))
-      get_value_set_rec(op0, op0_set, "", op0->type, false);
-
     object_mapt op1_set;
-    if (!is_pointer_type(op0))
-      get_value_set_rec(op1, op1_set, "", op1->type, false);
+    get_value_set_rec(op0, op0_set, "", (op0)->type, false);
```
lucasccordeiro, line 586: Master's #6905 (`retire_objectless_operand`, db172e19b9) already covers this site soundly and flipped github_426_2/3/4 and github_2512_11 to CORE. Rebase and re-derive against it.

src/pointer-analysis/value_set.cpp
```diff
@@ -582,12 +582,37 @@ void value_sett::get_value_set_rec(
...
+     * a cast of the index as char ptr. This makes ESBMC think that both operands are
+     * valid, which is currently unsupported.
+     */
+    auto op_contains_valid_address = [this](value_sett::object_mapt o) -> bool {
+      for (const auto &it : o)
+      {
+        auto expr = object_numbering[it.first];
+        if (
+          is_symbol2t(expr) || is_constant_string2t(expr) ||
+          is_constant_array2t(expr))
```
lucasccordeiro, line 604: The head commit drops `is_null_object2t`, which 67cca087 added specifically to restore NULL-dereference detection. Restore it or state why the removal is correct.

regression/esbmc/github_695_3_true/main.c
```diff
@@ -0,0 +1,50 @@
...
+    int a;
+    int b;
+};
+
+#define A_ELEMENT(o) o
+#define B_ELEMENT(o) o + sizeof(int)
+
+int main() {
+    struct obj A[4];
+    unsigned int obj_ptr = (unsigned int)&A[0];
```
lucasccordeiro, line 29: `(unsigned int)` truncates a 64-bit address, so this FAILS on truncation, not on the pointer arithmetic under test. Use `uintptr_t` (same in github_695_3_false and github_695_4_*).

regression/esbmc/github_695_2_true/test.desc
```diff
@@ -0,0 +1,4 @@
+KNOWNBUG
+main.c
+
```
lucasccordeiro, line 3: Empty flags line, while every sibling github_695_* test passes `--force-malloc-success`. Add the same flag so this KNOWNBUG stays comparable to the tests it was derived from.

regression/esbmc/github_695_0_false/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--force-malloc-success
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone matches any violated property, so this pins nothing. Add the expected violated-property line (same for the other `_false` tests).

PR #698, opened 2022-05-09
Fixed parse errors in operational mode

kunjsong01:
Fixed parse errors in operational mode to pass esbmc-cpp11/ch2_01.

### Before: 
```
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdio:23:13: error: cannot form a reference to 'void'
typedef void& FILE;
            ^
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdio:57:7: error: declaration of 'vsprintf' has a different language linkage
  int vsprintf(char * str, const char * format, va_list arg);
      ^
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdarg:23:5: note: previous declaration is here
int vsprintf(char * str, const char * format, va_list arg);
...
...
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdio:83:37: error: 'position' declared as a pointer to a reference of type 'fpos_t' (aka 'int &')
  int fgetpos(FILE * stream, fpos_t * position);
                                    ^
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdio:87:30: warning: 'const' qualifier on reference type 'fpos_t' (aka 'int &') has no effect [-Wignored-qualifiers]
  int fsetpos(FILE * stream, const fpos_t * pos);
                             ^~~~~~
/home/kunjian/ESBMC_Project/esbmc/src/cpp/library/cstdio:87:43: error: 'pos' declared as a pointer to a reference of type 'fpos_t' (aka 'int &')
  int fsetpos(FILE * stream, const fpos_t * pos);
                                          ^
### ERROR:
PARSING ERROR
```

### After: 
```
### ERROR:
Conversion of unsupported clang expr: "CXXTemporaryObjectExpr" to expression
CXXTemporaryObjectExpr 0x80a95f8 </home/kunjian/ESBMC_Project/esbmc/src/cpp/library/ios:148:10, col:19> 'std::ios_base' 'void ()'
```

### DIFF
--- a/src/cpp/library/cstdio
+++ b/src/cpp/library/cstdio
@@ -18,9 +18,9 @@
 #define EOF (-1)
 #endif
 
-typedef int& fpos_t;
+typedef int fpos_t;
 
-typedef void& FILE;
+typedef void FILE;
 
 typedef unsigned int TMP_MAX;
 
@@ -54,7 +54,7 @@ extern "C" {
 
   int vprintf(const char * format, va_list arg);
 
-  int vsprintf(char * str, const char * format, va_list arg);
+  //int vsprintf(char * str, const char * format, va_list arg); // removed as duplicate of vsprintf in cstdarg
 
   int fgetc(FILE * stream);
 
--- a/src/cpp/library/cstdio
+++ b/src/cpp/library/cstdio
@@ -20,7 +20,7 @@
 
 typedef int fpos_t;
 
-typedef void FILE;
+typedef void* FILE;
 
 typedef unsigned int TMP_MAX;
 
@@ -54,7 +54,7 @@ extern "C" {
 
   int vprintf(const char * format, va_list arg);
 
-  //int vsprintf(char * str, const char * format, va_list arg); // removed as duplicate of vsprintf in cstdarg
+  int vsprintf(char * str, const char * format, va_list arg);
 
   int fgetc(FILE * stream);
 
--- a/src/cpp/library/ostream
+++ b/src/cpp/library/ostream
@@ -129,7 +129,7 @@ namespace std
 namespace std {
   //Output manipulators
   ostream& endl ( ostream& os ) { return os; }
-  ostream& ends ( ostream& os ) {  }
+  ostream& ends ( ostream& os );
 }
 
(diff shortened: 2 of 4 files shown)


lucasccordeiro (approved):

mikhailramalho:
Shouldn't FILE be a void*?

src/cpp/library/ostream
```diff
@@ -129,7 +129,7 @@ namespace std
 namespace std {
   //Output manipulators
   ostream& endl ( ostream& os ) { return os; }
-  ostream& ends ( ostream& os ) {  }
+  ostream& ends ( ostream& os ) { return os; }
```
fbrausse: This function is supposed to alter `os` (by appending a `\0` character). Would it be possible instead of this no-op definition to have just a declaration similar to what you did for streambuf?

src/cpp/library/cstdio
```diff
@@ -54,7 +54,7 @@ extern "C" {
 
   int vprintf(const char * format, va_list arg);
 
-  int vsprintf(char * str, const char * format, va_list arg);
+  //int vsprintf(char * str, const char * format, va_list arg); // removed as duplicate of vsprintf in cstdarg
```
fbrausse: I suggest to fix it the other way round: remove the definition in cstdarg and use this one here, since this header is where it is [supposed](https://en.cppreference.com/w/cpp/io/c/vfprintf) to be.

fbrausse (commented):
I agree with Mikhail on `FILE`. Would it be possible eventually to import the definition from the C model? Aside from minor comments looks good to me.

src/cpp/library/cstdio
kunjsong01: Agreed. Done.

kunjsong01: Done.

src/cpp/library/ostream
kunjsong01: Done.

fbrausse (approved):
Thanks, LGTM!

PR #699, opened 2022-05-09
Domain paritioning v2

MatildeMark:


lucasccordeiro:
@MatildeMark: there are two files in conflict:

lucasccordeiro:
@MatildeMark: there are two files in conflict:

```
src/esbmc/esbmc_parseoptions.cpp
src/goto-programs/CMakeLists.txt
```

Could you please fix them?

pmatos:
Thanks for this contribution.

Revisiting it against current `master`, the branch no longer builds against the current codebase: `create_solver_factory` has been replaced by `create_solver` (the message parameter is gone), `runtime_solver` is now a `std::unique_ptr` (this PR assigns a `std::shared_ptr` to it), `bmct::do_cbmc` has been removed, and several signatures in the BMC / symex pipeline have changed. Reviving it would effectively mean reimplementing the pass against today's architecture rather than rebasing.

Given the branch has been inactive for a long while and would need a ground-up rework, I'm closing it. If variable-domain partitioning is something we want to pursue, it's best approached as a fresh design against the current BMC pipeline — note that `--parallel-solving` and multi-property checking now cover part of the parallelism need. Thanks again for the work.

PR #701, opened 2022-05-12
Skip assignment in value-set analysis for structure dereferences of incompatible types

### LINKED ISSUE #669: ESBMC crashes during value set with union or struct dereference
Runnig the program with: `esbmc main.i --function func1`

```c
// main.i
typedef union {
} a;
int b;
func1() { a c = *(a *)&b; }
```
Results in:

```
esbmc: ../src/pointer-analysis/value_set.cpp:1018: void value_sett::assign(const expr2tc&, const expr2tc&, bool): Assertion `lhs_type->type_id == rhs->type->type_id' failed.
Aborted (core dumped)
```

fbrausse:
This PR fixes #669 by avoiding a crash during assignment in value-set analysis. In particular, if a pointer to a non-structure is type-casted to a structure type, dereferenced and assigned like this
```c
typedef struct { [...] } a;
int y;
a x = *(a *)&y;
```
value_sett::assign() would crash as it does expect lhs and rhs to be of the same structure type - an error which the derefence code already caught by guarding. This happens when the structure (struct or union) `a` is non-empty. For empty `a` there is a special case during symex avoiding the crashing assignment, see https://github.com/esbmc/esbmc/issues/669#issuecomment-1091543699.

### DIFF
--- a/regression/esbmc/github_669-empty-struct/main.c
+++ b/regression/esbmc/github_669-empty-struct/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef struct {
+} a;
+int b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-empty-struct/test.desc
+++ b/regression/esbmc/github_669-empty-struct/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function func1
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_669-empty-union/main.c
+++ b/regression/esbmc/github_669-empty-union/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef union {
+} a;
+int b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-empty-union/test.desc
+++ b/regression/esbmc/github_669-empty-union/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function func1
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_669-struct-union/main.c
+++ b/regression/esbmc/github_669-struct-union/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef struct { int x;
+} a;
+union { int x; } b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-struct-union/test.desc
+++ b/regression/esbmc/github_669-struct-union/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function func1
+^\s*dereference failure:
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_669-struct/main.c
+++ b/regression/esbmc/github_669-struct/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef struct { float x;
+} a;
+int b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-struct/test.desc
+++ b/regression/esbmc/github_669-struct/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function func1
+^\s*dereference failure:
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_669-union-struct/main.c
+++ b/regression/esbmc/github_669-union-struct/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef union { int x;
+} a;
+struct { int x; } b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-union-struct/test.desc
+++ b/regression/esbmc/github_669-union-struct/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function func1
+^\s*dereference failure:
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_669-union/main.c
+++ b/regression/esbmc/github_669-union/main.c
@@ -1,0 +1,5 @@
+// main.i
+typedef union { float x;
+} a;
+int b;
+void func1() { a c = *(a *)&b; }
--- a/regression/esbmc/github_669-union/test.desc
+++ b/regression/esbmc/github_669-union/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function func1
+^\s*dereference failure:
+^VERIFICATION FAILED$
--- a/scripts/competitions/svcomp/stats-300s.txt
+++ b/scripts/competitions/svcomp/stats-300s.txt
@@ -1,9 +1,9 @@
 Statistics:          15648 Files
-  correct:            8448
-    correct true:     5130
-    correct false:    3318
+  correct:            8431
+    correct true:     5121
+    correct false:    3310
   incorrect:            53
     incorrect true:     24
     incorrect false:    29
-  unknown:            7147
-  Score:             12346 (max: 25567)
+  unknown:            7164
+  Score:             12320 (max: 25567)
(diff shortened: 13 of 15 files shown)


rafaelsamenezes (approved):
LGTM

lucasccordeiro (approved):

mikhailramalho (approved):
LGTM. I would just run on SV-Comp first and check if there are no regressions.

rafaelsamenezes:
> LGTM. I would just run on SV-Comp first and check if there are no regressions.

Right now all our servers are busy. I will run this PR tomorrow.

fbrausse:
Thanks for your reviews! What I'm not clear about here: is the current behaviour on empty structs/unions (VERIFICATION SUCCESSFUL) desired? Empty structs/unions are not covered well by the standards, and for `structs` probably not even allowed. The comment https://github.com/esbmc/esbmc/blob/e1e88ece939d51d89317f41a7adfef8b434cd40e/src/goto-symex/symex_assign.cpp#L150-L152 above the special case does not mention C types at all.

lucasccordeiro:
>  is the current behaviour on empty structs/unions (VERIFICATION SUCCESSFUL) desired?

I believe so.

@mikhailramalho and @rafaelsamenezes: do you have any thoughts?

lucasccordeiro:
@fbrausse: could you please run this PR over the SV-COMP benchmarks?

mikhailramalho:
> Thanks for your reviews! What I'm not clear about here: is the current behaviour on empty structs/unions (VERIFICATION SUCCESSFUL) desired? Empty structs/unions are not covered well by the standards, and for `structs` probably not even allowed. The comment
> 
> https://github.com/esbmc/esbmc/blob/e1e88ece939d51d89317f41a7adfef8b434cd40e/src/goto-symex/symex_assign.cpp#L150-L152
> 
> above the special case does not mention C types at all.

I think that's because only the C++ frontend produced such cases.

What does the address/memory sanitizer says when you try to do that? Or when you try to access the union/struct? Can you access a union/struct without fields?

What about classes?

fbrausse:
> I think that's because only the C++ frontend produced such cases.

Quite possible, but at least with the clang frontend they also come from empty C structs/unions.

> What does the address/memory sanitizer says when you try to do that? 

Nothing. The entire statement gets compiled out. Even without optimizations the assembly for the assignment and dereference statement is just empty.

> Or when you try to access the union/struct? Can you access a union/struct without fields?

I don't believe so. One cannot increase its size via `_Alignas` since that only applies to a member. Even this gets compiled to nothing by gcc and clang already without optimizations:
```c
typedef struct {} a;
typedef struct {
	_Alignas(16) a A;
	_Alignas(16) a B;
} b;
_Static_assert(sizeof(b) == 0, "");
int y;
int main()
{
	a x = ((b *)&y)->B;
}
```

> What about classes?

Zero-size classes make sense, they still have the `this` pointer, potentially overloaded assignment operators and so on. The comment about C++ POD types probably refers to those classes where assignment is not overloaded.

lucasccordeiro:
@fbrausse: could you also rebase this branch?

fbrausse:
> @fbrausse: could you also rebase this branch?

Done. I've also extended the comment for the special case.

fbrausse:
[This branch](https://github.com/esbmc/esbmc/actions/runs/2333134512):
```
  correct:            8431
    correct true:     5121
    correct false:    3310
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7164
```

fbrausse:
[master](https://github.com/esbmc/esbmc/actions/runs/2339312424):
```
  correct:            8438
    correct true:     5124
    correct false:    3314
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7157
  Score:             12330 (max: 25567)
```

lucasccordeiro:
@fbrausse: Can you please check which benchmarks we crash?

fbrausse:
Diffs:
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/gh-701/diff-old-new/unreach-call/results.2022-05-18_16-53-59.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/gh-701/diff-old-new/valid-memsafety/results.2022-05-18_16-54-08.diff.html)

fbrausse:
Except for [this](https://fbrausse.github.io/esbmc/svcomp/gh-701/diff-old-new/unreach-call/results.2022-05-18_16-53-59.diff.html#/table?filter=0(0*status*(category(in(unknown))))) where master did produce `unknown` and now it is OOM I only see noise. I'll run this one locally to check what exactly did crash before and maybe not anymore now.

lucasccordeiro:
okay, you get some timeouts...

lucasccordeiro:
Can you run some of these benchmarks locally with a larger timeout?

fbrausse:
> Except for [this](https://fbrausse.github.io/esbmc/svcomp/gh-701/diff-old-new/unreach-call/results.2022-05-18_16-53-59.diff.html#/table?filter=0(0*status*(category(in(unknown))))) where master did produce `unknown` and now it is OOM I only see noise. I'll run this one locally to check what exactly did crash before and maybe not anymore now.

In master that benchmark crashed due to the issue #704 fixed in PR #705. Backtrace:
```
#0  constant_string2t::to_array (this=0x555565395870)
    at esbmc/src/irep2/irep2_expr.cpp:326
#1  0x000055555573d5bb in get_arr_type (expr=...)
    at esbmc/src/pointer-analysis/dereference.cpp:44
#2  dereferencet::construct_from_array (this=0x7fffffffc6c0, value=..., offset=..., type=..., 
    guard=..., mode=dereferencet::READ, alignment=8)
    at esbmc/src/pointer-analysis/dereference.cpp:1021
#3  0x000055555573fbc5 in dereferencet::construct_from_dyn_struct_offset (this=0x7fffffffc6c0, 
    value=..., offset=..., type=..., guard=..., alignment=8, mode=dereferencet::READ, 
    failed_symbol=0x7fffffffb7e0) at esbmc/src/pointer-analysis/dereference.cpp:1346
```
Rebased on the recently merged fix for this issue, this benchmark no longer crashes there, but uses >14GiB memory in around 10 minutes on my laptop.

Still checking the other benchmarks with larger timeouts.

fbrausse:
Other OOM/timeout benchmarks locally on rebased branch:
- openssl/s3_srvr.blast.12.i.cil-1.c: correct result
  107.33user 0.60system 1:47.95elapsed 99%CPU (0avgtext+0avgdata 3395748maxresident)k
- eca-rers2012/Problem13_label54.c: correct result
  332.39user 7.76system 5:40.18elapsed 99%CPU (0avgtext+0avgdata 1963436maxresident)k
- eca-rers2012/Problem18_label26.c: correct result
  201.64user 0.71system 3:22.38elapsed 99%CPU (0avgtext+0avgdata 913888maxresident)k
- floats-cdfpl/newton_3_6.i: correct result
  107.20user 0.51system 1:47.72elapsed 99%CPU (0avgtext+0avgdata 342472maxresident)k
- loop-acceleration/array_2-2.i: correct result
  108.69user 3.58system 1:52.29elapsed 99%CPU (0avgtext+0avgdata 898956maxresident)k
- ldv-validator-v0.6/linux-stable-9ec4f65-1-110_1a-drivers--rtc--rtc-tegra.ko-entry_point.cil.out.i: correct result
  60.07user 1.72system 1:01.79elapsed 99%CPU (0avgtext+0avgdata 1552252maxresident)k
- ldv-challenges/linux-3.14_complex_emg_linux-kernel-locking-mutex_drivers-media-usb-dvb-usb-dvb-usb-m920x.cil.i: correct result
  79.22user 1.46system 1:20.69elapsed 99%CPU (0avgtext+0avgdata 1667600maxresident)k

PR #703, opened 2022-05-13
throw only statically allocated exception objects & catch by const-ref

fbrausse:
The reasons for this non-functional change are discussed at <https://github.com/esbmc/esbmc/discussions/702>.


mikhailramalho (approved):
Nice!

rafaelsamenezes (approved):

PR #707, opened 2022-05-17
Alive messages during decision procedure

maktheus:


src/util/simplify_expr2.cpp
```diff
@@ -8,6 +8,7 @@
 #include <irep2/irep2_utils.h>
 #include <util/type_byte_size.h>
 #include <util/message/default_message.h>
+#include <algorithm>
```
rafaelsamenezes, line 11: Why are you changing this file?

src/esbmc/bmc.cpp
```diff
@@ -42,7 +42,9 @@ Authors: Daniel Kroening, kroening@kroening.com
 #include <util/migrate.h>
 #include <util/show_symbol_table.h>
 #include <util/time_stopping.h>
+#include <thread>
 
+static bool bmc_interrupted = false;
```
rafaelsamenezes, line 47: No globals please. Pass this variable as a reference where it is used

src/esbmc/bmc.cpp
```diff
@@ -175,10 +195,13 @@ smt_convt::resultt bmct::run_decision_procedure(
 
   std::stringstream ss;
   ss << "Solving with solver " << smt_conv->solver_text();
+  std::thread t1(printAliveMessage, msg);
   msg.status(ss.str());
 
   fine_timet sat_start = current_time();
   smt_convt::resultt dec_result = smt_conv->dec_solve();
+  bmc_interrupted = true;
+  t1.join();
```
rafaelsamenezes, line 204: I don't think you should join... you will block the main thread for how much sleep there is left. Just detach it

src/esbmc/bmc.cpp
```diff
@@ -175,10 +195,13 @@ smt_convt::resultt bmct::run_decision_procedure(
 
   std::stringstream ss;
   ss << "Solving with solver " << smt_conv->solver_text();
+  std::thread t1(printAliveMessage, msg);
```
rafaelsamenezes, line 198: Name it with something like `print_alive`. Also this thread should only happen if the user calls for it (add an option).

src/esbmc/bmc.cpp
```diff
@@ -137,6 +139,24 @@ void bmct::error_trace(
...
+  while(!bmc_interrupted)
+  {
+    fine_timet sat_stop = current_time();
+
+    ss << "ESBMC is still alive.\n";
+    ss << "Total time:" << sat_stop - sat_start << "\n";
+    ss << "Press Ctrl-C to terminate ESBMC.\n";
+    msg.status(ss.str());
+    ss.str("");
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
rafaelsamenezes, line 156: You should get the interval value from the user

src/esbmc/bmc.cpp
```diff
@@ -137,6 +139,24 @@ void bmct::error_trace(
   msg.result(oss.str());
 }
 
+void printAliveMessage(const messaget &msg)
+{
+  std::stringstream ss;
```
rafaelsamenezes, line 144: Couldn't you put this inside the while? Then you wouldn't need to call `s.str("")` later.

regression/cbmc/01_cbmc_Fixedbv11/log.txt
```diff
@@ -0,0 +1,2 @@
+2022/05/12 17:45:59 Micro started
```
rafaelsamenezes, line 1: Remove this file

rafaelsamenezes (changes requested):
Few comments. Overall, the two options are still missing.

regression/cbmc/01_cbmc_Fixedbv11/log.txt
maktheus: done

src/esbmc/bmc.cpp
maktheus: done

src/esbmc/bmc.cpp
maktheus: done

src/esbmc/bmc.cpp
maktheus: done

src/util/simplify_expr2.cpp
maktheus: done

pmatos:
Thanks for this contribution, and apologies for the long delay in triaging it.

The "alive messages during the decision procedure" feature has since been implemented on `master` by #2147 (merged 2024-11-20), which prints periodic keep-alive messages via the `--enable-keep-alive` / `--keep-alive-interval` options and addresses the review feedback left here (member state instead of a global, a detached thread, a user-configurable interval).

The remaining arithmetic-simplification changes in this PR target `src/util/simplify_expr2.cpp`, which has since been removed — the simplifier now lives in `src/util/expr_simplifier.cpp`, where the corresponding folds are already implemented.

Since the feature is now available on `master`, I'm closing this as superseded by #2147. Thanks again for the work.

PR #713, opened 2022-05-26
[dereference] also descend into constant_union initializer when resolving dereferences

### LINKED ISSUE #709: Crash due to unresolved dereference in cast to union
Version: current master (`1f5b4ed3fbdb51a9280911f411b4766c4963a097`)

Command: 
``` bash
esbmc main.c --compact-trace
```
Input program:

``` c
#include <assert.h>

typedef struct {
  int data;
} s1;

typedef union {
  struct {
    int
      p0 : 1,
      p1 : 31;
  };
  int raw;
} s2;

int main() {
  s1 x;
  if (((s2)(&x)->data).p1) {}

  int var, ta, tb;
  ta = var;
  var = 0;
  tb = var;
  assert(ta == tb);

  return 0;
}
```

Output:
```
### ERROR:
Couldn't convert expression in unrecognised format
dereference
* pointer : address_of
    * pointer_obj : symbol
...
```

Note: The `--compact-trace` is used specifically for this small example. There's another larger example that also causes this similar issue without using the compact trace option.

fbrausse:
An attempt to fix #709.

Background: a dereference within a constant_union2t initializer is not being resolved in code like
```c
typedef struct { int data; } s1;
typedef union { int raw; } s2;
[...]
s1 x;
int var = ((s2)(&x)->data).raw;
```
This compiler-extension is referred to as a "cast to union", which ESBMC translates to a constant_union2t with the dereference as initializer.

The `.raw` leads to `dereferencet::dereference_expr_nonscalar()` being invoked, however this method is not prepared for constant unions, but deals only with type-casts, member, index and if2t expressions. 

This PR adds support for constant union expressions by resolving the initializer ~in the standard way (not via `nonscalar`, though in the above example the `(&x)->data` leads to another invocation of the `nonscalar` method)~.

~I can't quite pinpoint a justification for staying in the inner `nonscalar` recursion here, but there might be one. So this might not be the optimal solution - though my understanding of~ https://github.com/esbmc/esbmc/blob/ceeb86a60417d2333dea62bd374fa696231f5bfe/src/pointer-analysis/dereference.h#L291-L295 ~is that this inner recursion exists purely as an optimization.~

*Edit:* needed to stay in the inner recursion in order to properly handle the built-up `scalar_step_list`, as evidenced by a new test case.


src/pointer-analysis/dereference.cpp
```diff
@@ -423,71 +418,77 @@ expr2tc dereferencet::dereference_expr_nonscalar(
...
+      res2 = theif.true_value;
+
+    expr2tc fin = if2tc(res1->type, theif.cond, res1, res2);
+    return fin;
+  }
+
+  if(is_constant_union2t(expr))
+  {
+    constant_union2t &u = to_constant_union2t(expr);
+    assert(u.datatype_members.size() == 1);
```
lucasccordeiro, line 479: I'm unsure about this assertion. Could you please add some comments here?

regression/esbmc/github_709-2/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: Can you also add the safe version of this regression test?

regression/esbmc/github_709-1/test.desc
```diff
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 
 ^VERIFICATION FAILED$
```
lucasccordeiro, line 4: Can you also add the safe version of this regression test?

lucasccordeiro (changes requested):

lucasccordeiro:
@fbrausse: can you also run this PR over the SV-COMP benchmarks?

regression/esbmc/github_709-2/test.desc
fbrausse: Not sure, do you have a suggestion? This PR is about fixing a situation that has been missed (dereference inside a literal of union type). What did you have in mind?

lucasccordeiro: @fbrausse: what type of violation do you get here? Is it a dereference or an assertion violation?

I would suggest a test case that shows the 'correct behaviour'. For example, if the violation is related to the assert statement, what happens if we replace `assert(ta == tb);` with `assert(ta != tb);`

fbrausse:
Last run: https://github.com/esbmc/esbmc/blob/ceeb86a60417d2333dea62bd374fa696231f5bfe/scripts/competitions/svcomp/stats-300s.txt#L1-L9 This PR:
```
  correct:            8442
    correct true:     5130
    correct false:    3312
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7153
  Score:             12340 (max: 25567)
```

src/pointer-analysis/dereference.cpp
fbrausse: done.

regression/esbmc/github_709-1/test.desc
fbrausse: done

lucasccordeiro (approved):

fbrausse:
> @fbrausse: what type of violation do you get here? Is it a dereference or an assertion violation?

I'm getting an assertion violation:
```
Counterexample:

State 1 file main.c line 21 function main thread 0
  var = (signed int)((signed _ExtInt(31))((unsigned _ExtInt(31))(CONCAT(CONCAT(CONCAT(byte_extract_little_endian(x.data, 3), byte_extract_little_endian(x.data, 2)), byte_extract_little_endian(x.data, 1)), byte_extract_little_endian(x.data, 0)) >> 1) & 2147483647))

State 2 file main.c line 23 function main thread 0
  ta = 1 (00000000 00000000 00000000 00000001)

State 3 file main.c line 26 function main thread 0
Violated property:
  file main.c line 26 function main
  assertion ta == tb
  ta == tb
```
For `ta != tb` as well, since `x` and therefore also the value of the initializer `((s2)(&x)->d).p1` is nondet. If we init `x` with `0`, the verification will be successful.
*Edit:* that's what I did with regression/esbmc/github_709-3

Issue #716, opened 2022-05-30
Assertion `uni_type.members.size() != 0' failed


rafaelsamenezes:
ESBMC fails in the following program:

```c
#include <stdbool.h>
#include <stdint.h>
 
typedef union{
} union1;
 
typedef union{
    struct {
        uint32_t    union2_var:1;
    };
} union2;
extern union2 u2;
 
typedef struct {
    bool struct1_flag;
} struct1;
 
typedef struct {
    union1 u1;
    struct1 struct2_var1;
    struct1 struct2_var2;
 
} struct2;
 
typedef struct {
    struct2 struct3_var1;
} struct3;
 
extern struct3 struct3_extern;
 
void main(void) {
    bool flag;
 
    struct1 *struct1_p;
 
    if (flag) {
        struct1_p = &(struct3_extern.struct3_var1.struct2_var2);
    } else {
        struct1_p = &(struct3_extern.struct3_var1.struct2_var1);
    };
 
    struct1_p->struct1_flag = (u2.union2_var) != 0;
}
```

Reduced version:

```c
typedef union {
  struct {
    int a : 1
  };
} b;
b c;
main() { c.a; }
```

lucasccordeiro:
@fbrausse: could you please take a look at this issue?

fbrausse:
With neither of these examples I'm getting the error in the title with eff35eb1700838c5af8d619a70b6910ac84d4c0a.
The first version results in an error related to the empty union type `union1`. The "reduced" version gives me `VERIFICATION SUCCESSFUL`.

rafaelsamenezes:
I will confirm tomorrow if I did a mistake xD. As of now, the assertion itself is in `#ifndef NDEBUG` and the original issue had a `--no-pointer-check` (which shoudn't matter). So:

1. Could you confirm that you are running in a DEBUG build?
2. Could you try adding the `--no-pointer-check`?

fbrausse:
Yes to both :)

rafaelsamenezes:
Now I am using eff35eb1700838c5af8d619a70b6910ac84d4c0a 

The reduced version does not have the problem anymore. But now the original program still has: ```Assertion `uni_type.members.size() != 0' failed```

Issue #639, opened 2022-02-24
Memset over unions are not handled correctly


rafaelsamenezes:
The following program does not work properly when using esbmc while clang does not show any problem. Is this UB?

```c
#include <assert.h>

union my_obj {
  char a;
  int b;
  int c;
};

#define test_type union my_obj
#define test_size sizeof(union my_obj)

void manual_memset(void *ptr, int byte, unsigned N)
{
  char *dst = ptr;
  for(unsigned i = 0; i < N; i++)
    dst[i] = (unsigned char) byte;
}

void test_memset(test_type initial, int byte, unsigned N)
{ 
  test_type actual = initial;
  test_type expected = initial;
  memset(&actual, byte, N);
  manual_memset(&expected, byte, N);

  // byte by byte comparation then
  char *ptr1 = &actual;
  char *ptr2 = &expected;
  for(int i = 0; i < N; i++)
    assert(ptr1[i] == ptr2[i]); 
}
int main()
{
    int byte = 2;
    union my_obj test;
    test.b = 0;
    test_memset(test, byte, test_size);
}
```

lucasccordeiro:
@fbrausse: Could you please help @rafaelsamenezes here?

rafaelsamenezes:
It seems that the issue is from the derefence, it always gets the first member (even if it is not the largest)

By changing dereference.cpp at line 993 to:

```c
case flag_src_union | flag_dst_struct | flag_is_dyn_offs:
  {
    // Just perform an access to the first element thing.
    const union_type2t &uni_type = to_union_type(value->type);
    assert(uni_type.members.size() != 0);
    auto union_total_size = type_byte_size(value->type).to_uint64();
    // Let's find a member with the biggest size
    int selected_member_index;
    for(auto i = 0; i < to_union_type(value->type).members.size(); i++)
    {
      if(
        type_byte_size(to_union_type(value->type).members[i]).to_uint64() ==
        union_total_size)
      {
        selected_member_index = i;
        break;
      }
    }

    value = member2tc(uni_type.members[selected_member_index], value, uni_type.member_names[selected_member_index]);
    build_reference_rec(value, offset, type, guard, mode, alignment);
    break;
  }
```

Solves the issue. I am not sure if its a good solution though

fbrausse:
In the example I can see no UB. memset() usage is fine, only the casts `int` -> `unsigned char` -> `char` in `dst[i] = (unsigned char) byte;` are implementation-defined.
[...]

PR #717, opened 2022-05-31
[dereference] more union-related fixes

fbrausse:
This PR addresses #639 (using @rafaelsamenezes' solution) and #716. I'll start an SV-COMP benchmarks run as soon as the CI finished.


rafaelsamenezes (approved):

lucasccordeiro (changes requested):
@fbrausse: Can I ask you to add regression tests that produce "VERIFICATION FAILED"?

mikhailramalho (approved):
LGTM, we can merge it if there aren't new regressions on SV-COMP.

fbrausse:
stats diff: https://github.com/esbmc/esbmc/pull/717/commits/3fd87fc263136277ae7f302af8e2900c5e8420ef
detailed diff:
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-old-new/unreach-call/results.2022-06-01_14-09-28.diff.html)

In particular, one `unknown` got solved correctly now but 2 fast-ish correct ones now time out. I'm looking into the reasons.

fbrausse:
The 2 fast-ish ones are `pthread-deagle/circular_buffer_{bad,good}`. In a Debug build they both fail with `master` and this PR  at the same place https://github.com/esbmc/esbmc/blob/eff35eb1700838c5af8d619a70b6910ac84d4c0a/src/irep2/irep2_type.cpp#L266-L272 where `comp` is `"__lock"` and `name` is `"pthread_mutex_t"`. Thus the error did exist before, it just happened to work because our c2goto lib's struct layout has `__lock` as the first member of `pthread_mutex_t` - same as in the memory layout of the preprocessed SV-COMP .i file. However, in the .i it looks like
```c
struct __pthread_mutex_s
{
  int __lock;
  unsigned int __count;
  int __owner;
  unsigned int __nusers;
  int __kind;
  short __spins;
  short __elision;
  __pthread_list_t __list;
};
[...]
typedef union
{
  struct __pthread_mutex_s __data;
  char __size[40];
  long int __align;
} pthread_mutex_t;
```
and there it is not member of largest size (that would be `__data`). That's why this change showed the different behaviour.

Again, this is not a new error, it is just showing now. The cause lies in the preprocessing SV-COMP does. A merge request explaining this problem and giving a solution has been opened but not been acted upon: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1359

fbrausse:
PS: SV-COMP's `.c` sources corresponding to the `.i` work fine in both Debug and Release, as well as `master` and this PR.

lucasccordeiro (approved):
LGTM.

PR #721, opened 2022-05-31
[migrate] Improved assertion

kunjsong01:
made migrate name assertion more understandable as a result of Issue https://github.com/esbmc/esbmc/issues/720


mikhailramalho (changes requested):
I think when we want to print an error message, it should be:
1. print to cerr (using the message interface)
2. abort.

kunjsong01:
ACK
will replace it by a proper error message printing.

fbrausse:
Do we want to print a message here for this clearly internal (opposed to usage-related) error? Why is the assertion and maybe an accompanying comment not enough?

mikhailramalho:
I think error messages are always useful, even if it's only for the user to
report the error to us.

fbrausse:
>> clearly internal

> always

Are you suggesting to favour unconditionally enabled checks and messages over `assert()` and `NDEBUG` generally?

mikhailramalho:
Personally, yes.

In ESBMC, I think we use asserts in a number of places and always release using RelWithDebInfo (-O2 -g).

lucasccordeiro:
What is the conclusion here?

src/util/migrate.cpp
```diff
@@ -510,9 +510,9 @@ expr2tc sym_name_to_symbol(irep_idt init, type2tc type)
 
     char *endatptr, *endexmptr;
     level1_num = strtol(atstr.c_str(), &endatptr, 10);
-    assert(endatptr != atstr.c_str());
     thread_num = strtol(exmstr.c_str(), &endexmptr, 10);
-    assert(endexmptr != exmstr.c_str());
+    if(endatptr == atstr.c_str() || endexmptr == exmstr.c_str())
+      assert(!"Failed to mirgate symbol due to invalid identifier.");
```
fbrausse: Typo: migrate. If you're printing this, please also include the identifier.

fbrausse (commented):
For me a comment at the assertion would have sufficed, but appearently not-compiled-out checks and error messages to the user seem to find enough support. Fine with me here as this code is run relatively infrequently and the checks are not expensive. For things like #204 or similar however, I would suggest to keep assertions for performance reasons.

kunjsong01:
Based on the discussion above, I've done the following actions: 

1. Added a comment to make the assertion more understandable while avoiding additional checks (which might cause performance degradation)
2. Replaced the `assert` error reporting with our standard `msg` interface and abort.  

---
P.S. It's not the first time seeing error reporting using `assert`, e.g. https://github.com/esbmc/esbmc/pull/670/files. This is what makes me think that both `assert` and standard interfaces were acceptable for error reporting. But apparently they just exist the legacy code and should no longer be used.

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:
@mikhailramalho and @fbrausse: Can I ask you whether you still have comments for this PR?

src/util/migrate.cpp
```diff
@@ -510,9 +511,10 @@ expr2tc sym_name_to_symbol(irep_idt init, type2tc type)
 
     char *endatptr, *endexmptr;
     level1_num = strtol(atstr.c_str(), &endatptr, 10);
+    // assertion could fail due to invalid symbol identifier.
     assert(endatptr != atstr.c_str());
     thread_num = strtol(exmstr.c_str(), &endexmptr, 10);
-    assert(endexmptr != exmstr.c_str());
+    assert(endatptr != atstr.c_str());
```
fbrausse: Please restore the original check here. No use checking the same thing twice ;)

kunjsong01: Good spot! 
sorry, it's a copy-paste error...

mikhailramalho (approved):
LGTM.

PR #722, opened 2022-06-02
[smt] avoid creating zero-width bitvectors

fbrausse:
This PR does two things:
1. Impose a minimum of 1 on the width of bitvectors in smt_conv, specifically `mk_int_bv_sort()`, since BVs of zero width are not supported by (most, if not all) SMT solvers
2. Handle zero-width structs/unions (i.e. empty or only containing zero-width members) when flattening to bitvectors.

Together, both points fix #719 where zero-width unions are used and (for the counter-example) empty structs are bitcasted.

PR #734, opened 2022-06-07
[dereference] simplify resolving dereferences in non-scalar expressions

fbrausse:
`dereference_expr_nonscalar()` is invoked for member and index accesses and should perform all necessary dereferences inside such an expression. It does feature an optimization by which the intermediate arrays, structs or unions are not created but the target of right type is extracted from the intended offset. This was done through re-construction of the original expression but with the intermediate dereferences resolved. Building this re-constructed expression is not necessary.

This PR
- simplifies the computation by avoiding reconstruction of the basic expression and instead modifying the given one,
- introduces an assertion that index2t only operates on array- or string-typed source expressions (which already holds true), and
- thus removes all cases where the dereference_expr_*() checked for index2t applied to a pointer source operand.

I realize the last item is contrary to @mikhailramalho's suggestion in #725. I'm open to re-introduce these cases later. At least the `scalar_step_list` issue would be gone with these non-functional changes and they provide a basis for moving further on #725, I think.

### DIFF
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -362,7 +362,6 @@ expr2tc dereferencet::dereference_expr_nonscalar(
     // Check that either the base type that these steps are applied to matches
     // the type of the object we're wrapping in these steps. It's a type error
     // if there isn't a match.
-    // assert((*scalar_step_list.front()->get_sub_expr(0))->type == expr->type);
     type2tc base_of_steps_type = ns.follow(expr->type);
 
     if(dereference_type_compare(expr, base_of_steps_type))
@@ -400,12 +399,8 @@ expr2tc dereferencet::dereference_expr_nonscalar(
   }
 
   if(is_member2t(expr))
-  {
-    member2t &memb = to_member2t(expr);
-    expr2tc res =
-      dereference_expr_nonscalar(memb.source_value, guard, mode, base);
-    return res;
-  }
+    return dereference_expr_nonscalar(
+      to_member2t(expr).source_value, guard, mode, base);
 
   if(is_index2t(expr))
   {
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -432,6 +432,8 @@ expr2tc dereferencet::dereference_expr_nonscalar(
 
   if(is_if2t(expr) && !is_scalar_type(expr))
   {
+    assert(0);
+
     guardt g1 = guard, g2 = guard;
     const if2t &theif = to_if2t(expr);
     g1.add(theif.cond);
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -379,9 +379,32 @@ expr2tc dereferencet::dereference_expr_nonscalar(
 {
   if(is_dereference2t(expr))
   {
+    // Check that either the base type that these steps are applied to matches
+    // the type of the object we're wrapping in these steps. It's a type error
+    // if there isn't a match.
+    type2tc base_of_steps_type =
+      (*scalar_step_list.front()->get_sub_expr(0))->type;
+    base_of_steps_type = ns.follow(base_of_steps_type);
+
+    expr2tc size_check_expr = expr;
+    if(dereference_type_compare(size_check_expr, base_of_steps_type))
+    {
+      // We can just reconstruct this.
+      for(expr2tc tmp : scalar_step_list)
+      {
+        *tmp->get_sub_expr_nc(0) = size_check_expr;
+        size_check_expr = tmp;
+      }
+    }
+    else
+    {
+      // We can't reconstruct this. The base types are incompatible.
+      bad_base_type_failure(
+        guard, get_type_id(*expr->type), get_type_id(*base_of_steps_type));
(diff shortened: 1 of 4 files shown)


fbrausse:
SV-COMP stats diff: https://github.com/fbrausse/esbmc/commit/d0744050b715327591c89aabc4ec7068c13b3150
Detailed diff to b8f2507e805d69a7aac879dbf95a8333da2fb205:
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-old-new/unreach-call/results.2022-06-09_09-36-06.diff.html)

The 6 correct <-> timeout/oom are noise as they all are close to the time/mem-limits.
The one unknown -> OOM previously segfaulted in `flatten_to_bitvector_rec()` and I can't reproduce that with current master. #722 may likely have fixed that one without another benchmark run having been performed.

src/pointer-analysis/dereference.cpp
```diff
@@ -338,128 +347,83 @@ void dereferencet::dereference_addrof_expr(
...
+    index2t &index = to_index2t(expr);
+    dereference_expr(index.index, guard, dereferencet::READ);
+    expr2tc res =
+      dereference_expr_nonscalar(index.source_value, guard, mode, base);
     return res;
   }
 
   if(is_if2t(expr) && !is_scalar_type(expr))
   {
+    assert(0);
```
lucasccordeiro: Shall we keep this assertion here? I believe we are now outputting a message and then aborting.

lucasccordeiro (approved):
I have just one minor comment; otherwise, this PR looks good to me.

src/pointer-analysis/dereference.cpp
fbrausse: I tried but haven't succeeded to come up with an example that triggers this code path. Also the regression tests don't. It seems to be dead code. I'm ok with either removing the assertion in favour of a comment saying "dead: remove this branch", but I didn't want to test SV-COMP without it, yet. The assertion is compiled out for the SV-COMP benchmark runs, so in case it is used after all, at least the other (main) changes of this PR are not affected.

I could start another SV-COMP run without this branch as part of this PR, if you like.

lucasccordeiro: > I could start another SV-COMP run without this branch as part of this PR, if you like.

Can you please run another run over SV-COMP without this branch/assertion?

lucasccordeiro:
@fbrausse: do not forget to rebase this branch with master.

src/pointer-analysis/dereference.cpp
```diff
@@ -338,138 +347,70 @@ void dereferencet::dereference_addrof_expr(
...
+    /* The first expression we're called with is index2t, member2t or non-scalar
+     * if2t. Thus, expr differs from base. */
+    assert(expr != base);
+
+    // Check that either the base type that these steps are applied to matches
+    // the type of the object we're wrapping in these steps. It's a type error
+    // if there isn't a match.
+    type2tc base_of_steps_type = ns.follow(expr->type);
+
+    if(dereference_type_compare(expr, base_of_steps_type))
```
rafaelsamenezes: Can't you remove the else and negate the condition? `if(!dereference_type_compare(expr, base_of_steps_type))`?

src/pointer-analysis/dereference.cpp
```diff
@@ -338,138 +347,70 @@ void dereferencet::dereference_addrof_expr(
...
   {
-    dereference_expr(to_index2t(expr).index, guard, dereferencet::READ);
-    scalar_step_list.push_front(expr);
-    expr2tc res = dereference_expr_nonscalar(
-      to_index2t(expr).source_value, guard, mode, scalar_step_list);
-    scalar_step_list.pop_front();
+    index2t &index = to_index2t(expr);
+    dereference_expr(index.index, guard, dereferencet::READ);
+    expr2tc res =
+      dereference_expr_nonscalar(index.source_value, guard, mode, base);
```
rafaelsamenezes: Wouldn't `return dereference_expr_nonscalar(index.source_value, guard, mode, base);
` work?

rafaelsamenezes (changes requested):
Some nitpicks

src/pointer-analysis/dereference.cpp
fbrausse: Sure would; changed.

src/pointer-analysis/dereference.cpp
fbrausse: OK, done

rafaelsamenezes (approved):

src/pointer-analysis/dereference.cpp
fbrausse: Confirmed. The `if2t` case is also dead for SV-COMP: stats diff master/now: https://github.com/esbmc/esbmc/pull/734/files#diff-eaa56d75b0df4075b5b34397706a4e5cd062a446c7250304bb873d6068d00601

fbrausse:
> do not forget to rebase this branch with master.

done.

lucasccordeiro:
Excellent, @fbrausse. We'll merge this PR once the CI gets green.

mikhailramalho (approved):
Nice, I'm glad to get rid of `scalar_step_list`.

PR #740, opened 2022-06-09
Get executable path

### LINKED ISSUE #739: Add a way to get the esbmc binary folder
I'd like to be able to invoke another process through ESBMC. However, I will need a way to get the binary directory of esbmc path. I believe the best approach for this is:

1. Store `argc` and `argv` into `esbmc_parseoptions.h`
2. Create a function in `filesystem.h` to extract the canonical path (with boost).

rafaelsamenezes:
Closes #739

### DIFF
--- a/src/util/parseoptions.cpp
+++ b/src/util/parseoptions.cpp
@@ -17,13 +17,15 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/parseoptions.h>
 #include <util/signal_catcher.h>
 #include <boost/program_options.hpp>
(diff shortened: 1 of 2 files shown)


fbrausse (approved):
According to the docs, this constructor doesn't throw or otherwise error though they are not particularly detailed what happens when the locale (imbue) can't interpret the string. I've tested an UTF8-renamed "ähm-esbmc" executable in the C locale and the path appears to be copied verbatim w/o errors. LGTM.

lucasccordeiro (approved):
LGTM.

PR #741, opened 2022-06-09
[clang-cpp-frontend] Initializer list

kunjsong01:
Added support for C++ constructor initializer list and enabled the corresponding test cases.


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -910,6 +961,17 @@ bool clang_cpp_convertert::get_decl_ref(
 
     break;
   }
+  case clang::Decl::Field:
```
mikhailramalho, line 955: Is this repeated from the C code?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -703,7 +703,7 @@ void clang_cpp_convertert::build_member_from_component(
   // Add this pointer as first argument
   std::size_t address = reinterpret_cast<std::size_t>(fd.getFirstDecl());
 
-  this_mapt::iterator it;
+  this_mapt::iterator it = this_map.find(address);
```
mikhailramalho, line 707: I still think we should get rid of this map.

regression/esbmc-cpp11/constructors/Constructor1-3/main.cpp
```diff
@@ -5,13 +5,13 @@ class t2
...
   {
   }
 };
 
 int main()
 {
   t2 instance2;
-  assert(instance2.i == 2);
+  assert(instance2.i == 1);
 }
```
mikhailramalho: Maybe this test case doesn't need to be changed?

mikhailramalho (commented):
I remember having issues when we were initializing array members, can you add a test case for that?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -930,6 +992,7 @@ bool clang_cpp_convertert::get_decl_ref(
   {
     msg.error(fmt::format(
       "Conversion of unsupported clang decl ref: {}", decl.getDeclKindName()));
+    decl.dumpColor();
```
rafaelsamenezes: This will dump on stdout directly, could you use the same approach that is used for the c-frontend?

rafaelsamenezes (changes requested):

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: For `clang::Decl::Var` or `clang::Decl::Field`, yes. 
But for `clang::Decl::CXXConstructor`, no. E.g. We do some extra parsing for function parameters when parsing CXXCtor decl reference. 

`clang_c_convertert::get_decl_ref` just contains two cases - it only distinguishes `enum` from `value decl`: 
https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2203-L2205
https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2216-L2218 

Not sure how this partation would handle CXX decl kind. So I'm currently doing an experiment: In commit f61f0c102f, I'm defering everything else to `clang_c_convertert::get_decl_ref` in the `default` case. 

Regression results before commit f61f0c102f: 
One additional test case is passing `cpp/cpp_with_pthread_02-no-lib` in  [regression results](https://github.com/esbmc/esbmc/runs/6851613967?check_suite_focus=true).

Regression results after commit f61f0c102f: 
See [comment below](https://github.com/esbmc/esbmc/pull/741#discussion_r896652687).

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: ACK

regression/esbmc-cpp11/constructors/Constructor1-3/main.cpp
kunjsong01: ACK

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: ACK. 
will use `llvm::raw_os_ostream` for future clang AST node dumping.

fbrausse:
>> This will dump on stdout directly, could you use the same approach that is used for the c-frontend?

> will use llvm::raw_os_ostream for future clang AST node dumping.

I'd like to point out that the colors in Clang's AST dumps (on stderr, actually, cf #678) are quite helpful for debugging, at least for me. Please don't decolorize `--parse-tree-*`. Maybe we should find a way to enable Clang's color dumping, [`setShowColor()`](https://clang.llvm.org/doxygen/classclang_1_1DiagnosticsEngine.html#aceb1f204bfa928e73298f5fe3ec31009) could be useful.

src/clang-cpp-frontend/clang_cpp_convert.cpp
mikhailramalho: > For `clang::Decl::Var` or `clang::Decl::Field`, yes. But for `clang::Decl::CXXConstructor`, no. E.g. We do some extra parsing for function parameters when parsing CXXCtor decl reference.
> 
> `clang_c_convertert::get_decl_ref` just contains two cases - it only distinguishes `enum` from `value decl`:
> 
> https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2203-L2205
> 
> 
> https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2216-L2218
> 
> Not sure how this partation would handle CXX decl kind. So I'm currently doing an experiment: In commit [2929e5b](https://github.com/esbmc/esbmc/commit/2929e5b7d22456d7f7100c2817481a0aead5b146), I'm defering everything else to `clang_c_convertert::get_decl_ref` in the `default` case.

Yeah, that makes sense to me, let's handle the `clang::Decl::CXXConstructor` and forward the rest to the

mikhailramalho: > For `clang::Decl::Var` or `clang::Decl::Field`, yes. But for `clang::Decl::CXXConstructor`, no. E.g. We do some extra parsing for function parameters when parsing CXXCtor decl reference.
> 
> `clang_c_convertert::get_decl_ref` just contains two cases - it only distinguishes `enum` from `value decl`:
> 
> https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2203-L2205
> 
> 
> https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/clang-c-frontend/clang_c_convert.cpp#L2216-L2218
> 
> Not sure how this partation would handle CXX decl kind. So I'm currently doing an experiment: In commit [2929e5b](https://github.com/esbmc/esbmc/commit/2929e5b7d22456d7f7100c2817481a0aead5b146), I'm defering everything else to `clang_c_convertert::get_decl_ref` in the `default` case.

Yeah, that makes sense to me! Let's handle the constructor case in the C++ converter and forward the rest to the C one.

rafaelsamenezes:
> > > This will dump on stdout directly, could you use the same approach that is used for the c-frontend?
> 
> > will use llvm::raw_os_ostream for future clang AST node dumping.
> 
> I'd like to point out that the colors in Clang's AST dumps (on stderr, actually, cf #678) are quite helpful for debugging, at least for me. Please don't decolorize `--parse-tree-*`. Maybe we should find a way to enable Clang's color dumping, [`setShowColor()`](https://clang.llvm.org/doxygen/classclang_1_1DiagnosticsEngine.html#aceb1f204bfa928e73298f5fe3ec31009) could be useful.

Shall we open an issue for this?

lucasccordeiro:
@kunjsong01: can I ask you to rebase this PR with the master branch?

lucasccordeiro (approved):

kunjsong01:
> can I ask you to rebase this PR with the master branch?

Done.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -928,9 +974,16 @@ bool clang_cpp_convertert::get_decl_ref(
 
   default:
   {
-    msg.error(fmt::format(
-      "Conversion of unsupported clang decl ref: {}", decl.getDeclKindName()));
-    abort();
+    // If there is an unsupported clang decl, first try to forward it to clang_c_convertert::get_decl_ref.
+    // It might be already handled there.
```
fbrausse: I don't understand this comment. The code below prints an error and does not forward to clang_c_convertert::get_decl_ref() unless I'm overlooking something?

regression/esbmc-cpp11/constructors/Constructor1-4/main.cpp
```diff
@@ -7,12 +7,12 @@ class t3
...
-t3::t3() : i(3)
+t3::t3() : i(1)
 {
 }
 
 int main()
 {
   t3 instance3;
-  assert(instance3.i == 3);
+  assert(instance3.i == 1);
```
fbrausse, line 17: Does this test case not work with the constant `3`?

kunjsong01: It works. For consistency, I used either '1' or '0'.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: [CI regression results](https://github.com/esbmc/esbmc/actions/runs/2493590744) after forwarding looks ok.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: > The code below prints an error and does not forward to clang_c_convertert::get_decl_ref() 

The forwarding happens in line 954 - 957. The comment here hints that when we see this error message extending the cases in line 954 might resolve the issue. 

Here I'm coding in a defensive way. I don't want to blindly put the forwarding in `default` because I would like to explicitly distinguish `clang::decl::Foo` that can be handled by forwarding from `clang::decl::CXXFoo` that requires additional parsing.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -928,9 +974,16 @@ bool clang_cpp_convertert::get_decl_ref(
...
+    std::ostringstream oss;
+    llvm::raw_os_ostream ross(oss);
+    ross << "Conversion of unsupported clang decl ref for: "
+         << decl.getDeclKindName() << "\n";
+    decl.dump(ross);
+    ross.flush();
+    msg.error(oss.str());
```
rafaelsamenezes: Warning?

rafaelsamenezes (changes requested):

src/clang-cpp-frontend/clang_cpp_convert.cpp
fbrausse: Oh I see, this is a comment not about the way the code works, but about how to extend the code in case this error message is seen. Could you maybe clarify this by being explicit, e.g., "Cases not handled above are unknown clang decls; we print an error/warning. It might be possible to support them either here or in clang_c_frontend::get_decl_ref() depending on whether they are C++-specific or not."?

kunjsong01: I like this comment. Done. Added it in the code.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Done. Replaced error with warning.

src/clang-cpp-frontend/clang_cpp_convert.cpp
fbrausse: Many thanks!

fbrausse (approved):

kunjsong01:
> I remember having issues when we were initializing array members, can you add a test case for that?

Done. Added two test cases in commit 70914d99e5.

rafaelsamenezes (approved):

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: > I still think we should get rid of this map.

I agree. 
For the time being, I think I'll keep it when extending for CXXNewExpr and CXXDeleteExpr. The `this_map` might also be helpful in this  case. 

Be assured that I'm will completely remove the `this_map` in next PR ( for  CXXNewExpr and CXXDeleteExpr mentioned in #747).

Issue #538, opened 2021-11-02
Interval Analysis is always generating inductive instructions


rafaelsamenezes:
This is causing conflicts when running esbmc with incremental approach.

Issue #82, opened 2013-10-13
Investigate a 'switch' irep


jmorse:
Given a vector of guards and expressions, we can avoid encoding a very long if-chain by producing a free variable, then asserting values into the free variable using an implication. I.E in not-even-pseudo-code:

type something = new free variable;
assert guard-a-is-true -> something == value1;
assert guard-b-is-true -> something == value2;
...
assert guard-n-is-true -> something == valuen;

This will reduce the complexity of SMT AST that has to be manipulated by the solver. It might also just generally improve the performance of the solver and reduce state space (more assertions, etc). It'd also mean that variable consolidation at the end of a switch block is much simpler to encode (which will improve the performance of what Gennaro is doing right now).

ETA: not now.

PR #743, opened 2022-06-10
[dereference] add previously missed pointer_guard when value-set is not known to be exhaustive

fbrausse:
This PR is a request for comments on the way target expressions for pointer-dereferences are built.

The general strategy is:
1. get the value-set of the pointer (this is the set of all statically known expressions referring into existing symbols or `unknown2t`/`invalid2t`)
2. for each element of this value-set, reconstruct a canonical expression corresponding to the proper member/index/etc. of the referenced symbol
3. combine all these possible target expressions into a single (potentially big) if-then-else chain that looks something like this:
   ```
   pointer_guard ? deref-expression :
   pointer_guard' ? deref-expression' :
   [...] :
   pointer_guard'' ? deref-expression'' :
   deref-expression'''
   ```
   where each pointer-guard (usually) just says "ptr points into target object" (`same_object2t`).

In particular, if there are n > 0 "deref-expression"s in the expr constructed in step 3., then there are only (n-1) "pointer_guard" conditions. The pointer_guard''' for the last one is missing.

- Q1: Is this deliberate? If so, why is it OK to ignore this condition?

In case there are no deref-expressions constructed in step 3., the final result will only contain an "failed symbol". My understanding is that this is done in order for the formula to be well-typed, according to the comment.

**Edit:** @mikhailramalho answered Q1 essentially in positive: In case the value-set is known to be exhaustive and all (n-1) other conditions don't hold, the exhaustiveness guarantee implies that the n'th condition will and thus it does not need to be encoded.

However, the value-set is not known to be exhaustive in presence of unknown2t/invalid2t (whether just for unknown2t expressions is unclear as of yet, but under-approximating the cases when such a guarantee exists does not affect soundness). With this PR, the dereference code in 3. takes into account whether the value-set is known to be exhaustive and if not adds the failed-symbol unconditionally so the now constructed expression for 3. when the exhaustiveness is not guaranteed looks like:
   ```
   pointer_guard ? deref-expression :
   pointer_guard' ? deref-expression' :
   [...] :
   pointer_guard'' ? deref-expression'' :
   pointer_guard''' ? deref-expression''' : failed-symbol
   ```

### *Below considerations not relevant anymore.*

In the changes of this RFC I modified 3. to always include the last pointer_guard''' this way:
```
[...] pointer_guard''' ? deref-expression''' : failed-symbol
```

Obviously, this is where the timeouts on the regression test come from and why I wouldn't consider this PR for merging.

But it also uncovers
```
The following tests FAILED:
	126 - regression/esbmc-unix/03_bzip2smp_comb_01 (Failed)
	127 - regression/esbmc-unix/03_bzip2smp_comb_01_new (Failed)
```
and those I would argue should indeed fail (maybe more gracefully, but nonetheless), because of #584:
```
esbmc: /home/runner/work/esbmc/esbmc/src/pointer-analysis/value_set.cpp:1477: expr2tc value_sett::make_member(const expr2tc&, const irep_idt&): Assertion `is_struct_type(type) || is_union_type(type)' failed.
```
Here, a member into an integer is being constructed for the call to the variadic `fprintf()` - for `--64`, not for `--32`, because `va_list` is an integer on 32-bit but a struct on 64-bit. Currently, ESBMC does not discover this error.

- Q2: Should ESBMC discover this error and should these two tests actually fail at present given that #584 is still open?
- Q3: Could the detection of the above error be performed elsewhere without this performance impact?
- Q4: The commented check for compatibility of `code`-typed expressions https://github.com/esbmc/esbmc/blob/6cbaa832056ddd40dcb5c4b0447aa92ae5a7dff3/src/pointer-analysis/dereference.cpp#L545-L547 would result in further regression tests being identified as broken (see comment there). Should they?

### DIFF
--- a/scripts/competitions/svcomp/stats-300s.txt
+++ b/scripts/competitions/svcomp/stats-300s.txt
@@ -1,9 +1,9 @@
 Statistics:          15648 Files
-  correct:            8394
-    correct true:     5082
-    correct false:    3312
-  incorrect:            49
-    incorrect true:     23
-    incorrect false:    26
-  unknown:            7205
-  Score:             12324 (max: 25567)
+  correct:            8457
+    correct true:     5132
+    correct false:    3325
+  incorrect:            54
+    incorrect true:     25
+    incorrect false:    29
+  unknown:            7137
+  Score:             12325 (max: 25567)
--- a/scripts/competitions/svcomp/stats-300s.txt
+++ b/scripts/competitions/svcomp/stats-300s.txt
@@ -1,9 +1,9 @@
 Statistics:          15648 Files
-  correct:            8439
-    correct true:     5130
-    correct false:    3309
-  incorrect:            53
-    incorrect true:     24
-    incorrect false:    29
-  unknown:            7156
-  Score:             12337 (max: 25567)
+  correct:            8394
+    correct true:     5082
+    correct false:    3312
+  incorrect:            49
+    incorrect true:     23
+    incorrect false:    26
+  unknown:            7205
+  Score:             12324 (max: 25567)
--- a/src/irep2/irep2_utils.h
+++ b/src/irep2/irep2_utils.h
@@ -315,7 +315,7 @@ inline expr2tc gen_zero(const type2tc &type, bool array_as_array_of = false)
   }
 
   case type2t::pointer_id:
-    return symbol2tc(pointer_type2(), "NULL");
+    return symbol2tc(type, "NULL");
 
   case type2t::struct_id:
   {
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -452,15 +452,19 @@ expr2tc dereferencet::dereference(
   // now build big case split
   // only "good" objects
 
-  /* Fallback if dereference failes entirely: to make this a valid formula,
-   * return a failed symbol, so that this assignment gets a well typed free
-   * value. */
-  bool is_nonexhaustive = false;
+  /* If the value-set contains unknown or invalid, we cannot be sure it contains
+   * all possible values and we have to add a fallback symbol in case all guards
+   * evaluate to false. On the other hand when it is exhaustive, we only need to
+   * encode (n-1) guards for the n values in the if-then-else chain below. This
+   * is done by leaving 'value' initially empty.
+   *
+   * XXX fbrausse: get_value_set() should compute this information */
+  bool known_exhaustive = true;
   for(const expr2tc &target : points_to_set)
-    is_nonexhaustive |= is_unknown2t(target) || is_invalid2t(target);
+    known_exhaustive &= !(is_unknown2t(target) || is_invalid2t(target));
 
   expr2tc value;
-  if(is_nonexhaustive)
+  if(!known_exhaustive)
     value = make_failed_symbol(type);
 
   for(const expr2tc &target : points_to_set)
@@ -498,7 +502,12 @@ expr2tc dereferencet::dereference(
     internal_items.clear();
   }
   else if(is_nil_expr(value))
+  {
+    /* Fallback if dereference failes entirely: to make this a valid formula,
+     * return a failed symbol, so that this assignment gets a well typed free
+     * value. */
     value = make_failed_symbol(type);
+  }
 
   return value;
 }
@@ -550,28 +559,7 @@ bool dereferencet::dereference_type_compare(
   }
 
   if(is_code_type(object) && is_code_type(dereference_type))
-  {
-#if 1
     return true;
-#else /* fails (correctly so?) for some of the 01_pthread_* cases which pass a
-       * function with signature incompatible to that expected by
-       * pthread_create() */
-    const code_type2t &ot = to_code_type(object->type);
-    const code_type2t &dt = to_code_type(dereference_type);
-    if(ot.arguments.size() != dt.arguments.size())
-      return false;
-    /* arguments could be derived */
-    for(size_t i = 0; i < ot.arguments.size(); i++)
-    {
-      expr2tc dummy = gen_zero(ot.arguments[i]);
-      if(!dereference_type_compare(dummy, dt.arguments[i]))
-        return false;
-    }
-    /* return value could be derived */
-    expr2tc dummy = gen_zero(ot.ret_type);
-    return dereference_type_compare(dummy, dt.ret_type);
-#endif
-  }
 
   // check for struct prefixes
 
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -452,29 +452,32 @@ expr2tc dereferencet::dereference(
   // now build big case split
   // only "good" objects
 
-  expr2tc value;
+  /* Fallback if dereference failes entirely: to make this a valid formula,
+   * return a failed symbol, so that this assignment gets a well typed free
+   * value. */
+  expr2tc value = make_failed_symbol(type);
 
-  for(value_setst::valuest::const_iterator it = points_to_set.begin();
-      it != points_to_set.end();
-      it++)
+  for(const expr2tc &target : points_to_set)
   {
     expr2tc new_value, pointer_guard;
 
     new_value = build_reference_to(
-      *it, mode, src, type, guard, lexical_offset, pointer_guard);
+      target, mode, src, type, guard, lexical_offset, pointer_guard);
 
-    if(!is_nil_expr(new_value))
+    if(is_nil_expr(new_value))
+      continue;
+
+    if(!dereference_type_compare(new_value, type))
     {
-      if(is_nil_expr(value))
-      {
-        value = new_value;
-      }
-      else
-      {
-        // Chain a big if-then-else case.
-        value = if2tc(type, pointer_guard, new_value, value);
-      }
+      bad_base_type_failure(
+        guard, get_type_id(*type), get_type_id(*new_value->type));
+      continue;
     }
+
+    assert(!is_nil_expr(pointer_guard));
+
+    // Chain a big if-then-else case.
+    value = if2tc(type, pointer_guard, new_value, value);
   }
 
   if(mode == INTERNAL)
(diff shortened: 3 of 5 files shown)


mikhailramalho:
>
>    - Q1: Is this deliberate? If so, why is it OK to ignore this condition?
>
> I always thought that it was because the value-set doesn't hold failed
symbols as values, so if pointer_guard'' is false, then it had to be
deref-expression'''

-- 

Mikhail Ramalho.

fbrausse:
> I always thought that it was because the value-set doesn't hold failed
symbols as values, so if pointer_guard'' is false, then it had to be
deref-expression'''

Is that because the value-set is supposed to be semi-exhaustive in the sense that
- either it does not contain unknown/invalid, then all possible values are statically known and contained in the value-set
- or it does contain unknown/invalid, in which case `[...] : false ? value : [...]` will be part of the if-then-else result and a corresponding "invalid-pointer condition" will be asserted https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/pointer-analysis/dereference.cpp#L576-L580

?

mikhailramalho:
AFAIK, yes.

Also, I don't think we have a limit to how big the value-set gets; this usually introduces huge performance regressions. At some point, we wondered if stuff like #82 on the backend would improve performance in these cases.

fbrausse:
Oh wow, these are interesting semantics, thank you!

So when it contains unknown or invalid, the value-set is not necessarily exhaustive, yes? Isn't that a problem when a possible (valid) value is missing as the last pointer_guard''' is not checked? That's what I checked in this PR and it led to the newly found errors in `regression/esbmc-unix/03_bzip2smp_comb_01*` related to the `va_list`.

mikhailramalho:
Em sex., 10 de jun. de 2022 às 15:31, Franz Brauße ***@***.***>
escreveu:

> Oh wow, these are interesting semantics, thank you!
>
> So when it contains unknown or invalid, the value-set is not necessarily
> exhaustive, yes?
>

I think maybe only for unknowns? I don't really remember...


> Isn't that a problem when a possible (valid) value is missing as the last
> pointer_guard''' is not checked? That's what I checked in this PR and it
> led to the newly found errors in
> regression/esbmc-unix/03_bzip2smp_comb_01* related to the va_list.
>

Well, if it's an actual bug, then we should push the code to ESBMC.

-- 

Mikhail Ramalho.

fbrausse:
I did run the SV-COMP benchmarks over this in https://github.com/esbmc/esbmc/actions/runs/2469706300
Master:
```
  correct:            8439
    correct true:     5130
    correct false:    3309
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7156
  Score:             12337 (max: 25567)
```
This PR:
```
  correct:            8394
    correct true:     5082
    correct false:    3312
  incorrect:            49
    incorrect true:     23
    incorrect false:    26
  unknown:            7205
  Score:             12324 (max: 25567)
```
The performance impact is noticable, but we also get the number of incorrect ones down.

fbrausse:
Diffs:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-old-new/termination/results.2022-06-10_19-07-52.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-old-new/unreach-call/results.2022-06-10_19-07-54.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-old-new/valid-memsafety/results.2022-06-10_19-08-03.diff.html)

fbrausse:
After only adding the failed-symbol and missed pointer_guard when there is unknown/invalid in the value-set, the performance impact is much smaller, plus we get results for previously unknowns, but also new unknowns as well as one timeout -> wrong. I'm looking into those.

Stats diff to master: https://github.com/esbmc/esbmc/pull/743/files#diff-eaa56d75b0df4075b5b34397706a4e5cd062a446c7250304bb873d6068d00601
Detailed diff to master:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-old-new2/termination/results.2022-06-12_09-36-10.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-old-new2/unreach-call/results.2022-06-12_09-36-12.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-old-new2/valid-memsafety/results.2022-06-12_09-36-20.diff.html)

fbrausse:
1. `c/pthread-atomic/read_write_lock-2.i` timeout->wrong fails for a Debug build in both, master and this PR, due to pthread_mutex_t layout (member `__lock`)
   ```
   src/irep2/irep2_type.cpp:266: unsigned int struct_union_data::get_component_number(const irep_idt&) const: Assertion `0 && fmt::format( "Looking up index of nonexistant member \"{}\" in struct/union \"{}\"", comp, name) .c_str()' failed.
   ```
   and gives the wrong answer in both, master and this PR, for the unpreprocessed .c source.
2. `c/memsafety/test-0214.i` unknown -> timeout fails in Debug builds with both, master and this PR here (mentioning #538) https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/goto-symex/execution_state.cpp#L240-L244

   In the master Release build, this was `unknown` because ESBMC aborts after Boolector complained `"[boolector] boolector_cond: argument 'e_else' belongs to different Boolector instance\n\n"` in https://github.com/esbmc/esbmc/blob/4a5b329ac140257715d2e575733583b00b530532/src/solvers/smt/array_conv.cpp#L346

   Same for `c/memsafety/test-0217.i` and `c/memsafety/test-0218.i`.
3. `c/ntdrivers/floppy.i.cil-1.c` unknown -> correct fails in master/Debug with
   ```
   src/solvers/boolector/boolector_conv.cpp:633: virtual const smt_ast* boolector_convt::mk_ite(smt_astt, smt_astt, smt_astt): Assertion `t->sort->get_data_width() == f->sort->get_data_width()' failed.
   ```
   but works also in a Debug build of this PR. Maybe connected to item 2. above?

   Same for `c/ntdrivers/floppy.i.cil-3.c` unknown -> timeout
   and `c/loops-crafted-1/discover_list.c` unknown -> correct.
4. `c/ldv-linux-3.0/module_get_put-drivers-video-aty-aty128fb.ko.cil.out.i` unknown -> correct failed in master/Debug with
   ```
   src/irep2/irep2_type.cpp:266: unsigned int struct_union_data::get_component_number(const irep_idt&) const: Assertion `0 && fmt::format( "Looking up index of nonexistant member \"{}\" in struct/union \"{}\"", comp, name) .c_str()' failed.
   ```
   where `comp` is `"owner"` and `name` is `"struct aty128fb_par"`. A Debug build of this PR produces the correct result same as Release.
5. `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--iio--imu--inv_mpu6050--inv-mpu6050.ko-entry_point.cil.out.i` timeout -> unknown now aborts with
   ```
   ### ERROR: 
   Can't construct rvalue reference to array type during dereference
   ```
   Takes a very long time to run, so I'm still investigating what the cause is and whether on master it did work.
6. `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--staging--unisys--visorbus--visorbus.ko-entry_point.cil.out.i` unknown -> timeout failed with an obscure message on master:
   ```
   ### ERROR: 
   function call: argument "c:linux-4.2-rc1.tar.xz-08_1a-drivers--staging--unisys--visorbus--visorbus.ko-entry_point.cil.out.i@353789@F@controlvm_respond_physdev_changestate@state" type mismatch: got struct, expected struct
   ```
   However, Debug builds of both, master and this PR, show that an earlier assertion fails:
   ```
   src/irep2/irep2_type.cpp:266: unsigned int struct_union_data::get_component_number(const irep_idt&) const: Assertion `0 && fmt::format( "Looking up index of nonexistant member \"{}\" in struct/union \"{}\"", comp, name) .c_str()' failed.
   ```
   with `comp` being `"name"` and `name` being `"struct kobject"`. So no change in the validity of this benchmark result.
   Same for `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--staging--unisys--visorbus--visorbus.ko-entry_point.cil.out.i`
   and `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-43_2a-drivers--staging--unisys--visorbus--visorbus.ko-entry_point.cil.out.i`.
7. `c/list-ext-properties/test-0214_1.i` unknown -> timeout is similar to 2. (same error about different Boolector instances) on master/Release but errors earlier on master/Debug with
   ```
   src/solvers/boolector/boolector_conv.cpp:633: virtual const smt_ast* boolector_convt::mk_ite(smt_astt, smt_astt, smt_astt): Assertion `t->sort->get_data_width() == f->sort->get_data_width()' failed.
   ```
   Same for `c/list-ext-properties/test-0217_1.i` unknown -> correct.

fbrausse:
Item 5. above already fails on master with the same rvalue-to-array error when constructing a reference to type
```c
struct inv_mpu6050_platform_data {
   __s8 orientation[9U] ;
};
```
by recursing over the array-typed member.

lucasccordeiro (approved):
LGTM.

fbrausse:
I've reverted the changes to value_set::assign() which added support for the LHS to be an if2t. This was necessary in the [first attempt](https://github.com/esbmc/esbmc/pull/743/commits/f688907e3df2ae3848fdd5a6f831ae528b812c27) which always created (at least) one if-then-else, even for the LHS.

lucasccordeiro:
@mikhailramalho and @rafaelsamenezes: do you have comments for this PR?

mikhailramalho:
Should we at least try to fix `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--iio--imu--inv_mpu6050--inv-mpu6050.ko-entry_point.cil.out.i`?

IIUC, that error message is related to unions.

fbrausse:
> Should we at least try to fix `c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--iio--imu--inv_mpu6050--inv-mpu6050.ko-entry_point.cil.out.i`?
> 
> IIUC, that error message is related to unions.

See https://github.com/esbmc/esbmc/pull/743#issuecomment-1153579288, item 5. is not about unions but one of the `construct_*_from_*()` calling `build_reference_rec()` on a struct member that actually is an array.

We should fix all of these. ;) This PR does not change the outcome of this particular benchmark: it is the same on master, just timed out during the benchmark run.

mikhailramalho:
I see, it's fine by me then.

mikhailramalho (approved):
LGTM

PR #745, opened 2022-06-13
Display dereferenced pointer-arithmetic as `p[idx]`

fbrausse:
Addresses #725 which aimed at removing the frontend's translation of `p[i]` to `*(p+i)`.

During the discussion it has become clear that repurposing the `index2t` type for that breaks in a few places since currently it operates only on array- and string-typed source expressions. Thus, this PR keeps the translation in the frontend, and instead adjusts the counter-example presentation to recognize pointer-arithmetic under dereferences and display those in the index notation.

It does so recursively, e.g. `*((0+p)-(1-(1+3))` gets displayed as `p[0 - (1 - (1 + 3))]` and thus should address https://github.com/esbmc/esbmc/issues/725#issuecomment-1152436623 as well.

### DIFF
--- a/src/clang-c-frontend/expr2c.cpp
+++ b/src/clang-c-frontend/expr2c.cpp
@@ -2022,6 +2022,73 @@ std::string expr2ct::convert_extract(const exprt &src)
          std::to_string(lower) + ")";
 }
 
+/* Checks whether the expression `e` is one performing pointer-arithmetic, that
+ * is, addition/subtraction of an integer-typed expression to/from a
+ * pointer-typed expression.
+ *
+ * If so, `true` is returned and `ptr` holds the address of the inner-most
+ * pointer-typed expression while `idx` gets assigned a (newly constructed, in
+ * case of multiple levels of pointer-typed expressions) expression that
+ * corresponds to the index into `*ptr`.
+ *
+ * Note, just a pointer-typed symbol (or constant) is not recognized as pointer-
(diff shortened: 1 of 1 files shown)


mikhailramalho (approved):
LGTM, thanks @fbrausse!

lucasccordeiro (approved):

fbrausse:
Thanks for your reviews!

@mikhailramalho the `cmt_lvalue` changes in your patch in #725, are they independent and could be applied in addition with those in this PR? What is the meaning of `cmt_lvalue`?

mikhailramalho:
It was probably for typechecking in the old frontend, but it seems we don't use it anymore

rafaelsamenezes:
@fbrausse do you think that we could create some stable test cases for this in the CE?

lucasccordeiro:

fbrausse:
@rafaelsamenezes Not sure, I could of course encode the status-quo output (and maybe the previous one) into a regex for some particular cases, but if at some point we decide to reorder/simplify operands the test case would fail. Is that what you had in mind?

@lucasccordeiro done

rafaelsamenezes:
> @rafaelsamenezes Not sure, I could of course encode the status-quo output (and maybe the previous one) into a regex for some particular cases, but if at some point we decide to reorder/simplify operands the test case would fail. Is that what you had in mind?

Yes, and you are correct. It would be hard to make it stable.

Issue #736, opened 2022-06-08
[ESBMC++ Dev Tracking] Incorrect verification result for lvalue reference


kunjsong01:
Incorrect verification result for reference. 

### Test case (global variable reference): 
```cpp
  1 #include <cassert>
  2 int g;
  3 void function(int &t) { t = 2;}
  4
  5 int main()
  6 {
  7   int &r = g;
  8   r = 1;
  9   assert(g == 1);
 10   function(r);
 11   assert(g == 2);
 12 }
```

Output: 
```yaml
Counterexample:

State 1 file main.cpp line 9 function main thread 0
----------------------------------------------------
Violated property:
  file main.cpp line 9 function main // <--- should not fail
  assertion g == 1
  g == 1


VERIFICATION FAILED
```

### Test case (local variable reference): 
```cpp
  1 #include <cassert>
  2 void function(int &t) { t = 2;}
  3
  4 int main()
  5 {
  6   int g;
  7   int &r = g;
  8   r = 1;
  9   assert(g == 1);
 10   function(r);
 11   assert(g == 2);
 12 }
```

Output: 
```yaml
Counterexample:

State 2 file main_local.cpp line 11 function main thread 0
----------------------------------------------------
Violated property:
  file main_local.cpp line 11 function main
  assertion g == 2
  g == 2


VERIFICATION FAILED
```

### Info to reproduce the bug:
esbmc mainline 884a7ad184198 + `gh pr checkout 733` 

commandline: ./esbmc <program_under_test>.cpp

kunjsong01:
AST of `int &r = g;`: 

```cpp
    |-DeclStmt 0x8a20f58 <line:7:3, col:13>
    | `-VarDecl 0x8a20ed0 <col:3, col:12> col:8 used r 'int &' cinit
    |   `-DeclRefExpr 0x8a20f38 <col:12> 'int' lvalue Var 0x8a20e38 'g' 'int'
```

The frontend created irep tree: 
```yaml
    1: code
        * type: code
            * arguments: 
        * operands: 
          0: code
              * type: code
                  * arguments: 
              * operands: 
                0: symbol      # ---> this should be converted to "pointer" with subtype==signedbv
                    * type: signedbv      
                        * width: 32
                        * #cpp_type: signed_int
                    * name: r
                    * identifier: c:ref.cpp@76@F@main#@r
                1: symbol     # ---> this operand should wrapped in an "address_of" node
                    * type: signedbv    
                        * width: 32
[...]

PR #753, opened 2022-06-15
[clang-cpp-frontend] Fix cpp reference

kunjsong01:
Implemented generalized method to handle lvalue reference in Issue https://github.com/esbmc/esbmc/issues/736 

Also resolved problems for lvalue reference in function callsite (CallExpr) and initialisation (DeclStmt) documented in https://github.com/esbmc/esbmc/issues/736#issuecomment-1157590502 and https://github.com/esbmc/esbmc/issues/736#issuecomment-1157607950.

### DIFF
--- a/regression/esbmc-cpp11/reference/Reference1/main.cpp
+++ b/regression/esbmc-cpp11/reference/Reference1/main.cpp
@@ -1,0 +1,19 @@
+#include <cassert>
+
+class t1 {
+public:
+  int i;
+
+  t1(): i(0)
+  {
+  }
+};
+
+int main()
+{
+  t1 instance1;
+  t1 &r = instance1;
+  assert(r.i == 0); // pass
+  r.i = 1;
+  assert(r.i == 1); // fail
+}
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1225,10 +1225,8 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
   switch(stmt.getStmtClass())
   {
-  /*
-       The following enum values are the the expr of a program,
-       defined on the Expr class
-    */
+  //The following enum values are the the expr of a program,
+  //defined on the Expr class
 
   // Objects that are implicit defined on the code syntax.
   // One example is the gcc ternary operator, which can be:
@@ -1256,6 +1254,10 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
     if(get_decl_ref(dcl, new_expr))
       return true;
+
+    // if type is lvalue ref, convert to dereference subtree
+    if(is_lvalue_reference(dcl))
+      assert(!"TODO - convert to dereference subtree");
 
     break;
   }
@@ -3115,4 +3117,19 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
   }
 
   return nullptr;
+}
+
+bool clang_c_convertert::is_lvalue_reference(const clang::Decl &d)
+{
+  if(const auto *nd = llvm::dyn_cast<clang::ValueDecl>(&d))
+  {
+    const clang::QualType &q_type = nd->getType();
+    const clang::Type &the_type = *q_type.getTypePtrOrNull();
+    if(the_type.getTypeClass() == clang::Type::LValueReference)
+    {
+      return true;
+    }
+  }
+
+  return false;
 }
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -506,7 +506,19 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
     if(get_expr(*vd.getInit(), val))
       return true;
 
-    gen_typecast(ns, val, t);
+    if (is_lvalue_reference(vd))
+    {
+      // We have converted lvalue reference to pointer in a declaration statement with initialisation.
+      // As a result, generate address_of to match the pointer type.
+      exprt result_expr = exprt("address_of", t);
+      result_expr.move_to_operands(val);
+      val.swap(result_expr);
+    }
+    else
+    {
+      // just typecast to match the type of LHS
+      gen_typecast(ns, val, t);
+    }
 
     added_symbol.value = val;
     decl.operands().push_back(val);
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1257,7 +1257,18 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
     // if type is lvalue ref, convert to dereference subtree
     if(is_lvalue_reference(dcl))
-      assert(!"TODO - convert to dereference subtree");
+    {
+      // get type and make a dereference subtree
+      typet deref_type = new_expr.type().subtype();
+      exprt deref_expr = exprt("dereference", deref_type);
+      // the original subtree will be the operand of dereference subtree
+      deref_expr.move_to_operands(new_expr);
+
+      new_expr.swap(deref_expr);
+
+      // add location node
+      new_expr.location() = location;
+    }
 
     break;
   }
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -560,9 +560,9 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
   std::string id, name;
   get_decl_name(fd, name, id);
 
-  if(id == "c:@F@function#&I#")
+  if(id == "c:@F@main#")
   {
-    printf("Got function\n");
+    printf("Got main\n");
   }
 
   symbolt symbol;
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -506,22 +506,17 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
     if(get_expr(*vd.getInit(), val))
       return true;
 
-    if(is_lvalue_reference(vd))
-    {
-      // We have converted lvalue reference to pointer in a declaration statement with initialisation.
-      // As a result, generate address_of to match the pointer type.
-      exprt result_expr = exprt("address_of", t);
-      result_expr.move_to_operands(val);
-      val.swap(result_expr);
-    }
-    else
-    {
-      // just typecast to match the type of LHS
-      gen_typecast(ns, val, t);
-    }
+    // just typecast to match the type of LHS
+    gen_typecast(ns, val, t);
 
     added_symbol.value = val;
     decl.operands().push_back(val);
+
+    if(is_lvalue_reference(vd))
+    {
+      // mark the subtree for adjuster to convert lvalue reference decl-block
+      decl.set("reference", 1);
+    }
   }
 
   decl.location() = location_begin;
@@ -571,6 +566,11 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
 
   std::string id, name;
   get_decl_name(fd, name, id);
+
+  if(id == "c:@F@main#")
+  {
+    printf("Got main\n");
+  }
 
   symbolt symbol;
   get_default_symbol(
@@ -1243,7 +1243,7 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
   switch(stmt.getStmtClass())
   {
-  //The following enum values are the the expr of a program,
+  //The following enum values are the expr of a program,
   //defined on the Expr class
 
   // Objects that are implicit defined on the code syntax.
@@ -1276,6 +1276,9 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     // if type is lvalue ref, convert to dereference subtree
     if(is_lvalue_reference(dcl))
     {
+      // mark the subtree for adjuster to convert lvalue reference DeclRefExpr
+      printf("@@ Mark the irep and move conversion to adjuster\n");
+      /*
       // get type and make a dereference subtree
       typet deref_type = new_expr.type().subtype();
       exprt deref_expr = exprt("dereference", deref_type);
@@ -1286,6 +1289,7 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
       // add location node
       new_expr.location() = location;
+      */
     }
 
     break;
@@ -1503,12 +1507,15 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
             static_cast<const clang::Decl &>(*decl.getDecl());
           if(is_lvalue_reference(dcl))
           {
+            printf("@@ Mark the irep and move conversion to adjuster\n");
+            /*
             // we have converted lvalue ref to a dereference subtree
-            // To make it a function argument, we need to wrap it in a "&(.)" expresson
+            // To make it a function argument, we need to wrap it in a "&(.)" expression
             typet addrof_type = single_arg.operands().at(0).type();
             exprt addrof_expr = exprt("address_of", addrof_type);
             addrof_expr.operands().push_back(single_arg);
             single_arg.swap(addrof_expr);
+            */
           }
         }
       }
@@ -3174,11 +3181,9 @@ bool clang_c_convertert::is_lvalue_reference(const clang::Decl &d)
   if(const auto *nd = llvm::dyn_cast<clang::ValueDecl>(&d))
   {
     const clang::QualType &q_type = nd->getType();
-    const clang::Type &the_type = *q_type.getTypePtrOrNull();
-    if(the_type.getTypeClass() == clang::Type::LValueReference)
-    {
-      return true;
-    }
+    const clang::Type &the_type =
+      *q_type.getTypePtrOrNull(); // TODO: check nullptr before dereferencing???
+    return the_type.getTypeClass() == clang::Type::LValueReference;
   }
 
   return false;
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -572,11 +572,6 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
   std::string id, name;
   get_decl_name(fd, name, id);
 
-  if(id == "c:@F@main#")
-  {
-    printf("Got main\n");
-  }
-
   symbolt symbol;
   get_default_symbol(
     symbol,
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -982,6 +982,17 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
     break;
   }
 
+  case clang::Type::LValueReference:
+  {
+    const clang::LValueReferenceType &lvrt =
+      static_cast<const clang::LValueReferenceType &>(the_type);
+
+    if(get_type(lvrt.getPointeeTypeAsWritten(), new_type))
+      return true;
+
+    break;
+  }
+
   case clang::Type::MacroQualified:
   {
     const clang::MacroQualifiedType &macro =
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -506,7 +506,7 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
     if(get_expr(*vd.getInit(), val))
       return true;
 
-    if (is_lvalue_reference(vd))
+    if(is_lvalue_reference(vd))
     {
       // We have converted lvalue reference to pointer in a declaration statement with initialisation.
       // As a result, generate address_of to match the pointer type.
@@ -1496,6 +1496,27 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
       exprt single_arg;
       if(get_expr(*arg, single_arg))
         return true;
+
+      // if argument is lvalue ref, convert to address_of subtree
+      if(const auto *stmt = llvm::dyn_cast<clang::Stmt>(arg))
+      {
+        if(stmt->getStmtClass() == clang::Stmt::DeclRefExprClass)
+        {
+          const clang::DeclRefExpr &decl =
+            static_cast<const clang::DeclRefExpr &>(*stmt);
+          const clang::Decl &dcl =
+            static_cast<const clang::Decl &>(*decl.getDecl());
+          if(is_lvalue_reference(dcl))
+          {
+            // we have converted lvalue ref to a dereference subtree
+            // To make it a function argument, we need to wrap it in a "&(.)" expresson
+            typet addrof_type = single_arg.operands().at(0).type();
+            exprt addrof_expr = exprt("address_of", addrof_type);
+            addrof_expr.operands().push_back(single_arg);
+            single_arg.swap(addrof_expr);
+          }
+        }
+      }
 
       call.arguments().push_back(single_arg);
     }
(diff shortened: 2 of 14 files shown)


src/clang-c-frontend/clang_c_convert.h
```diff
@@ -206,6 +206,8 @@ class clang_c_convertert
   const clang::Decl *get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt);
 
   void gen_typecast_to_union(exprt &dest, const typet &type);
+
+  bool is_lvalue_reference(const clang::Decl &d);
```
fbrausse, line 210: Could this be made `static` to hint that this function doesn't depend on any members of `class clang_c_convert`?

kunjsong01:
### Failed Regression Investigation: 
`regression/esbmc-unix2/23_picosat-846_05` failed due to unsupported type: 
```
ESBMC version 6.10.0 64-bit x86_64 linux
Target: 32-bit little-endian x86_64-unknown-linux with esbmc libc
file app.c: Parsing
file picosat.c: Parsing
Converting
### ERROR:
Conversion of unsupported clang type: "LValueReference
LValueReferenceType 0x785d800 '__builtin_va_list &'
`-TypedefType 0x7653700 '__builtin_va_list' sugar
  |-Typedef 0x752f468 '__builtin_va_list'
  `-PointerType 0x752f420 'char *'
    `-BuiltinType 0x752e900 'char'

### ERROR:
CONVERSION ERROR
```

Supposedly, lvalue reference shall be only applicable to C++. The above error happened becuase commit b168349a5592fd5e9b8865ce5494126308f5f462 moved the `clang::Type::LValueReference` case from clang-c into clang-cpp converter. 

Strangely, `LValueReferenceType` node does not exist in the parse tree ([parse_tree.c.zip](https://github.com/esbmc/esbmc/files/8917963/parse_tree.c.zip))

### Fix:
commit 647a1c5db99cbcf0bc4bb2bf26c4007380071592 

### More investigations in the future:
1. Does `--parse-tree-*` show the entire tree when verifying multiple files? 
2. Why C conversion involves `LValueReferenceType` node?

kunjsong01:
We have an ongoing SV-COMP CI for this PR: https://github.com/esbmc/esbmc/actions/runs/2511001904

regression/esbmc-cpp11/reference/Reference1/main.cpp
```diff
@@ -0,0 +1,19 @@
...
+  }
+};
+
+int main()
+{
+  t1 instance1;
+  t1 &r = instance1;
+  assert(r.i == 0); // pass
+  r.i = 1;
+  assert(r.i == 1); // fail
```
lucasccordeiro, line 18: Should it fail? The test desc says that it should succeed instead.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3119,3 +3165,18 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
 
   return nullptr;
 }
+
+bool clang_c_convertert::is_lvalue_reference(const clang::Decl &d)
+{
+  if(const auto *nd = llvm::dyn_cast<clang::ValueDecl>(&d))
```
lucasccordeiro, line 3181: Add some comments?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -157,6 +157,19 @@ bool clang_cpp_convertert::get_type(
 {
   switch(the_type.getTypeClass())
   {
+  case clang::Type::LValueReference:
```
lucasccordeiro, line 160: Add a short explanation?

lucasccordeiro (approved):
I have just minor comments.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -486,7 +486,19 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
...
+    if(is_lvalue_reference(vd))
+    {
+      // We have converted lvalue reference to pointer in a declaration statement with initialisation.
+      // As a result, generate address_of to match the pointer type.
+      exprt result_expr = exprt("address_of", t);
+      result_expr.move_to_operands(val);
+      val.swap(result_expr);
+    }
+    else
+    {
```
rafaelsamenezes: remove the brackets

rafaelsamenezes (approved):
Just a syntax comment, after that LGTM.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3119,3 +3165,18 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
...
   return nullptr;
 }
+
+bool clang_c_convertert::is_lvalue_reference(const clang::Decl &d)
+{
+  if(const auto *nd = llvm::dyn_cast<clang::ValueDecl>(&d))
+  {
+    const clang::QualType &q_type = nd->getType();
+    const clang::Type &the_type = *q_type.getTypePtrOrNull();
+    if(the_type.getTypeClass() == clang::Type::LValueReference)
```
mikhailramalho: this can be written as `return the_type.getTypeClass() == clang::Type::LValueReference);`

mikhailramalho (changes requested):
@kunjsong01 I think this patch changes the usual flow of conversion; you're changing the IR while converting it: the two-step conversion (first convert then adjust) was designed to handle cases like these.

If you need to do these operations, you need to mark the irep while converting and handle it in the adjust. Maybe you're looking for a `.lvalue` flag?

I don't recommend to merge the PR as is.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1439,6 +1464,27 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+        if(stmt->getStmtClass() == clang::Stmt::DeclRefExprClass)
+        {
+          const clang::DeclRefExpr &decl =
+            static_cast<const clang::DeclRefExpr &>(*stmt);
+          const clang::Decl &dcl =
+            static_cast<const clang::Decl &>(*decl.getDecl());
+          if(is_lvalue_reference(dcl))
+          {
+            // we have converted lvalue ref to a dereference subtree
+            // To make it a function argument, we need to wrap it in a "&(.)" expresson
```
fbrausse: typo: expression

fbrausse (approved):
LGTM.

I'm wondering whether there are guarantees C++ makes on references (such as always being initialized and basically its value-set being constant) we could make use of in order to do less work on them.

lucasccordeiro:
This PR:

```
  correct:            6285
    correct true:     3799
    correct false:    2486
  incorrect:            32
    incorrect true:     15
    incorrect false:    17
  unknown:            9331
  Score:              9332 (max: 25567)
```

Master:

```
  correct:            8457
    correct true:     5137
    correct false:    3320
  incorrect:            54
    incorrect true:     25
    incorrect false:    29
  unknown:            7137
  Score:             12330 (max: 25567)
```

mikhailramalho:
Weird results.

lucasccordeiro:
Yes! Kunjian will need to check for the failing cases.

mikhailramalho:
@kunjsong01 did you check how it was implemented in v1? I've only found:
```
    case clang::Type::LValueReference:
    {
      const clang::LValueReferenceType &lvrt =
        static_cast<const clang::LValueReferenceType &>(the_type);

      const clang::QualType &pointee = lvrt.getPointeeTypeAsWritten();

      typet sub_type;
      if(get_type(pointee, sub_type))
        return true;

      // Special case, pointers to structs/unions/classes must not
      // have a copy of it, but a reference to the type
      if(sub_type.is_struct() || sub_type.is_union())
      {
        struct_union_typet t = to_struct_union_type(sub_type);
        sub_type = symbol_typet(t.tag().as_string());
      }

      new_type = gen_pointer_type(sub_type);
      new_type.set("#reference", true);
      break;
    }
```

Isn't it enough? Actually, I don't think we even need to set the "#reference" flag.

lucasccordeiro:
I have seen that @kunjsong01 has used 10 seconds as a timeout. That's why we get weird results.

I have started a new run with the standard 300 seconds:

https://github.com/esbmc/esbmc/actions/runs/2548183464

lucasccordeiro:
I have seen that @kunjsong01 has used 10 seconds as a timeout. That's why we get weird results.

I have started a new run with the standard 300 seconds:

https://github.com/esbmc/esbmc/actions/runs/2548183464

lucasccordeiro:
This PR with 300 seconds:

```
  correct:            8435
    correct true:     5124
    correct false:    3311
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7160
  Score:             12327 (max: 25567)
```

Master:

```
  correct:            8455
    correct true:     5128
    correct false:    3327
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7140
  Score:             12351 (max: 25567)
```

kunjsong01:
>@kunjsong01 I think this patch changes the usual flow of conversion; you're changing the IR while converting it: the two-step conversion (first convert then adjust) was designed to handle cases like these.

>@kunjsong01 did you check how it was implemented in v1? I've only found:

TODOs: 
1. ~~Check lvalue reference handler in v1. If this handler works, use this one~~  (done evaluation in PR #771)
2. If v1 handler does not work, split the patch into two-step conversion.

kunjsong01:
> @kunjsong01 did you check how it was implemented in v1? I've only found:
> 
> ```
>     case clang::Type::LValueReference:
>     {
>       const clang::LValueReferenceType &lvrt =
>         static_cast<const clang::LValueReferenceType &>(the_type);
> 
>       const clang::QualType &pointee = lvrt.getPointeeTypeAsWritten();
> 
>       typet sub_type;
>       if(get_type(pointee, sub_type))
>         return true;
> 
>       // Special case, pointers to structs/unions/classes must not
>       // have a copy of it, but a reference to the type
>       if(sub_type.is_struct() || sub_type.is_union())
>       {
>         struct_union_typet t = to_struct_union_type(sub_type);
>         sub_type = symbol_typet(t.tag().as_string());
>       }
> 
>       new_type = gen_pointer_type(sub_type);
>       new_type.set("#reference", true);
>       break;
>     }
> ```
> 
> Isn't it enough? Actually, I don't think we even need to set the "#reference" flag.

@mikhailramalho  No, it's not enough. The code above generates the incorrect conversion: 
```
704 Symbol......: c:@F@main#
705 Module......: main
706 Base name...: main
707 Mode........: C (0)
708 Type........: signed int ()
709 Value.......:
710 {
711 signed int * r=(signed int *)g;    // ---> original code: int &r = g; 
712
713 r = (signed int *)1;  // ---> orginal code: r = 1;
```

But we want: 
```
signed int * r=&g;

*r = 1;
```

kunjsong01:
> @kunjsong01 I think this patch changes the usual flow of conversion; you're changing the IR while converting it: the two-step conversion (first convert then adjust) was designed to handle cases like these.
> 
> If you need to do these operations, you need to mark the irep while converting and handle it in the adjust. Maybe you're looking for a `.lvalue` flag?
> 
> I don't recommend to merge the PR as is.

@mikhailramalho 
ACK. 
I thought of the two-step rule when making the changes, but I didn't go for it because I wanted to quickly test my idea as explained in https://github.com/esbmc/esbmc/issues/736#issuecomment-1156324386. 

I agree that we should stick to the two-step rule, otherwise the converter will become too messy. I'll move some conversion steps to the adjuster.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -505,7 +505,19 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
...
       return true;
 
-    gen_typecast(ns, val, t);
+    if(is_lvalue_reference(vd))
+    {
+      // We have converted lvalue reference to pointer in a declaration statement with initialisation.
+      // As a result, generate address_of to match the pointer type.
+      exprt result_expr = exprt("address_of", t);
+      result_expr.move_to_operands(val);
+      val.swap(result_expr);
```
kunjsong01: mark the irep and move this conversion to adjuster to handle lvalue reference initialisation, e.g. convertion `int& r = g;` into `int* r = &g;`

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1254,6 +1264,21 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     if(get_decl_ref(dcl, new_expr))
       return true;
 
+    // if type is lvalue ref, convert to dereference subtree
+    if(is_lvalue_reference(dcl))
+    {
+      // get type and make a dereference subtree
```
kunjsong01, line 1282: mark the irep tree and move this conversion to adjuster to handle DeclRefExpr, e.g. convert `r = 1;` (where r is an lvalue reference) to `*r = 1;`

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1458,6 +1483,27 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
       if(get_expr(*arg, single_arg))
         return true;
 
+      // if argument is lvalue ref, convert to address_of subtree
+      if(const auto *stmt = llvm::dyn_cast<clang::Stmt>(arg))
+      {
+        if(stmt->getStmtClass() == clang::Stmt::DeclRefExprClass)
```
kunjsong01, line 1502: mark the irep and move the conversion to adjuster to handle function callsite

mikhailramalho:
Btw @kunjsong01, here's the goto program generated by the last working version of the cpp frontend:
```
        // 25 file main.cpp line 15
        class t1 * r;
        // 26 file main.cpp line 15
        r=&instance1;
        // 27 file main.cpp line 16
        ASSERT r->cpp::t1::i == 0
        // 28 file main.cpp line 17
        r->cpp::t1::i=1;
        // 29 file main.cpp line 18
        ASSERT r->cpp::t1::i == 2
```

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: Done in 80d1afc383924e5b4aef7b6651939ceb73a25982

kunjsong01:
I tried to use the two-step conversion to deal with lvalue reference and only managed to complete it *partially* in commit 80d1afc383924e5b4aef7b6651939ceb73a25982 for lvalue reference initialisation and 4a3c7d4fd4f92f02ac6a7ab28d5f91f8385df8e7 for lvalue reference DeclExprRef in assignment, but really struggled with making further progression. 

My general feeling is that it really overcomplicates things a lot, not to mention the "reference: 1" marking gets lost during adjusting. 

Considering the test case below:
```cpp
     1	#include <cassert>
     2	void function(int &t) { t = 2;}
     3
     4	int main()
     5	{
     6	  int g;
     7	  int &r = g;
     8	  r = 1;
     9	  assert(g == 1);
    10	  function(r);
    11	  assert(r == 2);
    12	}
```

The [previous version of this patch](https://github.com/esbmc/esbmc/pull/777/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590) has a centralised place ([see line 1268 - 1280](https://github.com/esbmc/esbmc/pull/777/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590R1268-R1280)) to convert lvalue reference DeclExprRef and it generated the correct GOTO conversion: 
```c
        // 27 file main.cpp line 10 function main
     1: FUNCTION_CALL:  function(&(*r))   // ---> already converted 
        // 28 file main.cpp line 11 function main
        ASSERT *r == 2 // assertion r == 2    // ---> already converted
        // 29 file main.cpp line 11 function main
        IF !(*r == 2) THEN GOTO 2   // ---> already converted 
        // 30 file main.cpp line 11 function main
        0;
```

The current version of this patch generated the wrong GOTO program and we have to adjust more places for lvalue reference (probably more adjusting for complicated data structures in the future?)
```c
     1: FUNCTION_CALL:  function(r)    // ---> should be OK? 
        // 28 file main.cpp line 11 function main
        ASSERT (signed int)r == 2 // assertion r == 2    // ---> need to adjust typecast to dereference
        // 29 file main.cpp line 11 function main
        IF !((signed int)r == 2) THEN GOTO 2      // ---> need to adjust typecast to dereference
        // 30 file main.cpp line 11 function main
        0;
```

>@kunjsong01 I think this patch changes the usual flow of conversion; 

On second thought, we only change the usual flow of conversion for lvalue reference in the [previous version of this patch](https://github.com/esbmc/esbmc/pull/777/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590). Since all the changes are guarded by `if(is_lvalue_reference(dcl))`, it does not affect the conversion for other language constructs. 


@mikhailramalho do we have to stick to the two-step conversion strictly? It might work for other cases, but really overcomplicates things for lvalue reference. 
If we can have a centralised place to deal with lvalue referene ([e.g. just 12 lines to handle lvalue reference DeclRefExpr](https://github.com/esbmc/esbmc/pull/777/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590R1268-R1280)), why do we bother adding more lines of code in the adjuster?

kunjsong01:
> Btw @kunjsong01, here's the goto program generated by the last working version of the cpp frontend:
> 
> ```
>         // 25 file main.cpp line 15
>         class t1 * r;
>         // 26 file main.cpp line 15
>         r=&instance1;
>         // 27 file main.cpp line 16
>         ASSERT r->cpp::t1::i == 0
>         // 28 file main.cpp line 17
>         r->cpp::t1::i=1;
>         // 29 file main.cpp line 18
>         ASSERT r->cpp::t1::i == 2
> ```

The [previous version of this patch](https://github.com/esbmc/esbmc/pull/777) generates the following GOTO program: 
```c
main (c:@F@main#):
        // 20 no location
        t1 instance1;
        // 21 no location
        FUNCTION_CALL:  t1(&instance1)
        // 22 file main.cpp line 15 function main
         t1 { signed int i; } * r;
        // 23 file main.cpp line 15 function main
        r=&instance1;
        // 24 file main.cpp line 16 function main
        ASSERT r->i == 0 // assertion r.i == 0
        // 25 file main.cpp line 16 function main
        IF !(r->i == 0) THEN GOTO 1
        // 26 file main.cpp line 16 function main
        0;
```

@mikhailramalho Does it look about right? 
`t1 { signed int i; } * r;` appears quite strange. Ideally, it should be like "t1 * r;"

mikhailramalho:
> > Btw @kunjsong01, here's the goto program generated by the last working version of the cpp frontend:
> > ```
> >         // 25 file main.cpp line 15
> >         class t1 * r;
> >         // 26 file main.cpp line 15
> >         r=&instance1;
> >         // 27 file main.cpp line 16
> >         ASSERT r->cpp::t1::i == 0
> >         // 28 file main.cpp line 17
> >         r->cpp::t1::i=1;
> >         // 29 file main.cpp line 18
> >         ASSERT r->cpp::t1::i == 2
> > ```
> 
> The [previous version of this patch](https://github.com/esbmc/esbmc/pull/777) generates the following GOTO program:
> 
> ```c
> main (c:@F@main#):
>         // 20 no location
>         t1 instance1;
>         // 21 no location
>         FUNCTION_CALL:  t1(&instance1)
>         // 22 file main.cpp line 15 function main
>          t1 { signed int i; } * r;
>         // 23 file main.cpp line 15 function main
>         r=&instance1;
>         // 24 file main.cpp line 16 function main
>         ASSERT r->i == 0 // assertion r.i == 0
>         // 25 file main.cpp line 16 function main
>         IF !(r->i == 0) THEN GOTO 1
>         // 26 file main.cpp line 16 function main
>         0;
> ```
> 
> @mikhailramalho Does it look about right? `t1 { signed int i; } * r;` appears quite strange. Ideally, it should be like "t1 * r;"

The latter is wrong, it should gen a reference instruction. 

I just had a quick look and saw that ESBMC is using the expr2c class to dump c++ code, that's why you're not seeing the reference. This patch should fix it:  [0001-Fix-using-expr2c-to-print-c-code.patch.txt](https://github.com/esbmc/esbmc/files/9013741/0001-Fix-using-expr2c-to-print-c-code.patch.txt)  (it's just a quick fix, the mode should a member class for each frontend, and get_default_symbol should just call mode).

Regarding the two-step conversion, yes we should keep using it or it'll end up like the goto conversion, wherein a bunch of "adjustments" are happening while we convert stuff: it's now very hard to decouple the functionalities and change each code slightly breaks the other.

I suggest you take a look at the old frontend and how it handles the conversion, it goes:
```
cpp_typecheckt::convert_initializer
cpp_typecheckt::reference_initializer
cpp_typecheckt::reference_binding
```
where the address_of is created to the RHS.

You don't actually need to do all the typecheck performed by these functions, as they were already done by clang and we can assume that it's legal code, so focus only on the AST changes required.

kunjsong01:
> The latter is wrong, it should gen a reference instruction.

@mikhailramalho  Which 'reference instruction' are you referring to?

mikhailramalho:
> > The latter is wrong, it should gen a reference instruction.
> 
> @mikhailramalho Which 'reference instruction' are you referring to?

Sorry, I meant at symbol table level, with the attached patch we get:
```
Module......: main2
Base name...: main
Mode........: C++ (1)
Type........: int ()
Value.......: 
{
t1 instance1;
t1(&instance1);

struct t1 & r=instance1;
r->i == 0 ? 0 : __assert_fail(&"r.i == 0"[0], &"main2.cpp"[0], 16, &"int main()"[0]);
r->i = 1;
r->i == 2 ? 0 : __assert_fail(&"r.i == 2"[0], &"main2.cpp"[0], 18, &"int main()"[0]);
}
Flags.......: lvalue
Location....: file main2.cpp line 12
``` 

You can see the `struct t1 & r` there, now we need to check/adjust its initializer, esbmc will take care of the rest.

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: Tried 4a3c7d4fd4f92f02ac6a7ab28d5f91f8385df8e7 but it's making the adjust code structure too messy. 
Need to figure out a better way to adjust DeclRefExpr.

kunjsong01:
> I suggest you take a look at the old frontend and how it handles the conversion, it goes:
cpp_typecheckt::convert_initializer
cpp_typecheckt::reference_initializer
cpp_typecheckt::reference_binding

I looked into the above functions. They seem to use lots of ESBMC's existing methods, e.g. is_reference(symbol.type) . Also, the cpp_typecheckt::reference_binding  also deals with temporaries.

Compared to these methods, my idea appears really hacky to me. Because:
1. it does not deal with temporaries
2. it may not interface well with the existing ESBMC backend, i.e. may not be able to take full advantage of the C++ verification capabilities in the backend

@mikhailramalho I think I'm going to take your patch as a starting point and re-write the entire PR based on the old frotnend to avoid backfires due to the hacky method in this PR.

mikhailramalho:
> > I suggest you take a look at the old frontend and how it handles the conversion, it goes:
> > cpp_typecheckt::convert_initializer
> > cpp_typecheckt::reference_initializer
> > cpp_typecheckt::reference_binding
> 
> I looked into the above functions. They seem to use lots of ESBMC's existing methods, e.g. is_reference(symbol.type) . Also, the cpp_typecheckt::reference_binding also deals with temporaries.

`is_reference(symbol.type)` only checks if `#reference` flag is set to true.

> 
> Compared to these methods, my idea appears really hacky to me. Because:
> 
> 1. it does not deal with temporaries
> 2. it may not interface well with the existing ESBMC backend, i.e. may not be able to take full advantage of the C++ verification capabilities in the backend
> 
> @mikhailramalho I think I'm going to take your patch as a starting point and re-write the entire PR based on the old frotnend to avoid backfires due to the hacky method in this PR.

They do a bunch of stuff because they are actually type-checking the AST, which we don't need to do as clang did that already.

The flow should be as follow: under which scenarios `convert_initializer` is called? We should replicate the checks and the transformations the old frontend does to the AST.

You also don't need to have the same structure as the old frontend, if you compare the old C frontend and the clang one, the clang one is very simplified.

kunjsong01:
> They do a bunch of stuff because they are actually type-checking the AST, which we don't need to do as clang did that already.

> The flow should be as follow: under which scenarios convert_initializer is called? We should replicate the checks and the transformations the old frontend does to the AST.


TODO: 
1. Investigate and study the flow in `cpp_typecheckt::convert_initializer` and `cpp_typecheckt::reference_binding`. (Estimate: 0.5d) 
2. Replicate the old frontend flow and avoid unnecessary type-checking (Estimate: 2d)

kunjsong01:
replaced by https://github.com/esbmc/esbmc/pull/804

PR #754, opened 2022-06-16
Feature/add solidity uint types

correcthorsebatterystaple:
Add uint variants for solidity and ensure that all the types behave with literals when arithmetic operations are performed


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -272,8 +274,16 @@ bool solidity_convertert::get_var_decl(
     nlohmann::json init_value =
       is_state_var ? ast_node["value"] : ast_node["initialValue"];
 
+    nlohmann::json literal_type = nullptr;
+    if(
+      SolidityGrammar::get_expression_t(init_value) ==
+      SolidityGrammar::ExpressionT::Literal)
+    {
```
lucasccordeiro: There is no need for this bracket since we have just one instruction under the if-statement.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -12,6 +12,8 @@
 #include <util/message/format.h>
 #include <fstream>
 
+int DEBUG_COUNTER = 0;
```
lucasccordeiro: Do we really need it?

regression/esbmc-solidity/uint_overflow/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+contract.solast
+--function func --contract contract.sol
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro: Can you also add a version of this regression test that produces `VERIFICATION FAILED`?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -711,6 +721,14 @@ bool solidity_convertert::get_statement(
 }
 
 bool solidity_convertert::get_expr(const nlohmann::json &expr, exprt &new_expr)
+{
+  return get_expr(expr, nullptr, new_expr);
+}
+
+bool solidity_convertert::get_expr(
```
lucasccordeiro, line 744: Add some doxygen documentation?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1400,6 +1457,39 @@ bool solidity_convertert::get_array_to_pointer_type(
   return false;
 }
 
+bool get_elementary_type_name_uint(
```
lucasccordeiro: Add some documentation?

lucasccordeiro (changes requested):

src/solidity-frontend/solidity_grammar.h
```diff
@@ -50,6 +50,37 @@ enum ElementaryTypeNameT
 {
   // rule unsigned-integer-type
   UINT8 = 0,
+  UINT16 = 1,
```
kunjsong01: `= 1,` is not needed here. Same for the other items in this enum.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -65,21 +66,28 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
   if(type_name.contains("typeString"))
   {
     // for AST node that contains ["typeName"]["typeDescriptions"]
-    if(type_name["typeString"] == "uint8" || type_name["typeString"] == "bool")
+    std::set<std::string> uintTypeStringSet = {
```
kunjsong01: make it `static` in namespace SolidityGrammar?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -155,22 +163,34 @@ const char *type_name_to_str(TypeNameT type)
...
 ElementaryTypeNameT get_elementary_type_name_t(const nlohmann::json &type_name)
 {
+  std::string typeString = type_name["typeString"].get<std::string>();
   // rule unsigned-integer-type
-  if(type_name["typeString"] == "uint8")
-  {
-    return UINT8;
-  }
-  else if(type_name["typeString"] == "bool")
+  std::map<std::string, ElementaryTypeNameT> uintMap = {
```
kunjsong01: make it static in namespace SolidityGrammar?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1400,6 +1457,39 @@ bool solidity_convertert::get_array_to_pointer_type(
...
+  SolidityGrammar::ElementaryTypeNameT &type,
+  typet &out)
+{
+  std::map<SolidityGrammar::ElementaryTypeNameT, unsigned int> uintSizeMap = {
```
kunjsong01: make it `static` in namespace SolidityGrammar?

src/solidity-frontend/solidity_convert.cpp
kunjsong01: make it a private function in class solidity_convertert?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1400,6 +1457,39 @@ bool solidity_convertert::get_array_to_pointer_type(
...
+    {SolidityGrammar::UINT168, 168}, {SolidityGrammar::UINT176, 176},
+    {SolidityGrammar::UINT184, 184}, {SolidityGrammar::UINT192, 192},
+    {SolidityGrammar::UINT200, 200}, {SolidityGrammar::UINT208, 208},
+    {SolidityGrammar::UINT216, 216}, {SolidityGrammar::UINT224, 224},
+    {SolidityGrammar::UINT232, 232}, {SolidityGrammar::UINT240, 240},
+    {SolidityGrammar::UINT248, 248}, {SolidityGrammar::UINT256, 256},
+  };
+
+  if(!uintSizeMap.count(type))
+    return true;
```
kunjsong01: To make future debugging easier, would it be possible to report error and abort? e.g. `msg("type not found")`?

src/util/c_types.h
```diff
@@ -18,6 +18,7 @@ type2tc index_type2();
 typet int_type();
 type2tc int_type2();
 typet uint_type();
+typet uint_type(unsigned int);
```
kunjsong01: wrap it in "#ifdef ENABLE_SOLIDITY_FRONTEND" ?

kunjsong01 (changes requested):

src/solidity-frontend/solidity_convert.cpp
mikhailramalho: This needs to be removed.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -900,18 +953,22 @@ bool solidity_convertert::get_binary_operator_expr(
   exprt lhs, rhs;
   if(expr.contains("leftHandSide"))
   {
+    nlohmann::json literalType = expr["leftHandSide"]["typeDescriptions"];
+
     if(get_expr(expr["leftHandSide"], lhs))
       return true;
 
-    if(get_expr(expr["rightHandSide"], rhs))
+    if(get_expr(expr["rightHandSide"], literalType, rhs))
```
mikhailramalho, line 944: why are you passing the LHS type with the RHS expression? Shouldn't it be the expr's type?

src/util/c_types.cpp
```diff
@@ -111,6 +111,11 @@ typet uint_type()
   return unsignedbv_typet(config.ansi_c.int_width);
 }
 
+typet uint_type(unsigned int size)
```
mikhailramalho: This is not needed. Check how we build _ExtInt in the C frontend.

src/util/c_types.h
mikhailramalho: This is not needed. Check how we build _ExtInt in the C frontend.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -155,22 +163,34 @@ const char *type_name_to_str(TypeNameT type)
...
   }
-  else if(
-    type_name["typeString"].get<std::string>().find("int_const") !=
-    std::string::npos)
+  else if(typeString.find("int_const") != std::string::npos)
   {
     // For Literal, their typeString is like "int_const 100".
     // TODO: Fix me! For simplicity, we assume everything is unsigned int.
-    return UINT8;
+    return UINT256;
```
mikhailramalho: why 256? The comment seems to say otherwise.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -155,22 +163,34 @@ const char *type_name_to_str(TypeNameT type)
...
+    {"uint176", UINT176}, {"uint184", UINT184}, {"uint192", UINT192},
+    {"uint200", UINT200}, {"uint208", UINT208}, {"uint216", UINT216},
+    {"uint224", UINT224}, {"uint232", UINT232}, {"uint240", UINT240},
+    {"uint248", UINT248}, {"uint256", UINT256},
+  };
+  if(typeString.substr(0, 4) == "uint" && uintMap.count(typeString))
+  {
+    return uintMap[typeString];
+  }
+  else if(typeString == "bool")
```
mikhailramalho: there is no need for the elses here since every if returns.

mikhailramalho (changes requested):
TBH, I don't understand this PR:
1. `convert_integer_literal` takes an `const nlohmann::json &integer_literal` but somehow we are passing a `literal_type`?
2. Why do we need the type after all? Since you're passing the expr around, you should be able to retrieve it whenever you want.

src/solidity-frontend/solidity_convert.cpp
correcthorsebatterystaple: Has been overlook, will remove

src/solidity-frontend/solidity_grammar.h
correcthorsebatterystaple: Will amend

src/solidity-frontend/solidity_convert.cpp
correcthorsebatterystaple: The way the AST is structured is that if the expression is `uint8 i = 22`, this is passed in with the type on the RHS to be just a generic type of int_const without any indication to which variant of the int/uint it is. We make the assumption that the literal on the right will share the type of the rhs. which is why we pass in that literal type.

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: This will actually be removed as this PR fixes the defaulting

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: > TBH, I don't understand this PR:
> 
> 1. `convert_integer_literal` takes an `const nlohmann::json &integer_literal` but somehow we are passing a `literal_type`?
> 2. Why do we need the type after all? Since you're passing the expr around, you should be able to retrieve it whenever you want.

1. This is a result of the existing structure, `convert_integer_literal` seems to also handle the creating of of the variables of uint types. The way the AST is built is that in an initialization with a literal rvalue `uint8 i = 10` we get type definitions for the left side but for the right side we get a generic integer type without any indication for what subtype of the integer it is (unsigned or signed, bitsize) which then requires us to pass in the type from the left hand side which is what the literal_type stores here
2. This is true that we are passing the expression about but because literal expressions do not contain any specific type descriptions we need to pass them around from the lhs of the expression.

correcthorsebatterystaple:
@mikhailramalho 
> TBH, I don't understand this PR:
> 
> 1. `convert_integer_literal` takes an `const nlohmann::json &integer_literal` but somehow we are passing a `literal_type`?
> 2. Why do we need the type after all? Since you're passing the expr around, you should be able to retrieve it whenever you want.

1. This is a result of the existing structure, `convert_integer_literal` seems to also handle the creating of of the variables of uint types. The way the AST is built is that in an initialization with a literal rvalue `uint8 i = 10` we get type definitions for the left side but for the right side we get a generic integer type without any indication for what subtype of the integer it is (unsigned or signed, bitsize) which then requires us to pass in the type from the left hand side which is what the literal_type stores here
2. This is true that we are passing the expression about but because literal expressions do not contain any specific type descriptions we need to pass them around from the lhs of the expression.

src/util/c_types.cpp
correcthorsebatterystaple: Doesn't _ExtInt also do the same thing just directly with the `unsignedbv_typet`?

lucasccordeiro (approved):
LGTM.

lucasccordeiro:
@fbrausse and @rafaelsamenezes: could you please review this PR?

src/util/c_types.cpp
rafaelsamenezes: The issue here is that `c_types` regards... C base types. Here you are basically creating an alias. Which is not needed.

rafaelsamenezes (changes requested):

lucasccordeiro:
@correcthorsebatterystaple: could you please address @rafaelsamenezes' comments?

src/util/c_types.cpp
correcthorsebatterystaple: This makes sense, I have a redundant alias and as this is for solidity there is no reason for dependence on c_types. Will amend to use the unsignedbv_typet.

src/solidity-frontend/solidity_grammar.h
```diff
@@ -4,6 +4,7 @@
 #include <map>
 #include <string>
 #include <nlohmann/json.hpp>
+#include <set>
```
fbrausse: This seems unused in this header.

src/solidity-frontend/solidity_grammar.h
```diff
@@ -67,6 +101,24 @@ enum ElementaryTypeNameT
 };
 ElementaryTypeNameT get_elementary_type_name_t(const nlohmann::json &type_name);
 const char *elementary_type_name_to_str(ElementaryTypeNameT type);
+static std::map<ElementaryTypeNameT, unsigned int> uintSizeMap = {
```
fbrausse: While I understand exporting a way to map from the enum to the type might be necessary, making this static declaration in a header file is not a good approach. I have two issues with this:
1. not every mapping needs to be a `std::map`; as the enum is layed out, this could just as well be an array. Or a switch.
2. Having this `static` in the header means that every translation unit that #include-s it also gets a local `uintSizeMap` symbol. That is wasteful.

So, the suggestion is to move this symbol declaration to the place where it is needed and if exporting a way to obtain the size for one of these enum values is required, to add a function for that.

Also: if you keep the std::map, can it be `const`?

lucasccordeiro:
@correcthorsebatterystaple: could you please rebase this PR with the master branch?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -65,21 +85,19 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
   if(type_name.contains("typeString"))
   {
     // for AST node that contains ["typeName"]["typeDescriptions"]
-    if(type_name["typeString"] == "uint8" || type_name["typeString"] == "bool")
+    std::string typeString = type_name["typeString"].get<std::string>();
+
+    if(uintTypeStringSet.count(typeString) || typeString == "bool")
```
fbrausse: It seems that this is the only place `uintTypeStringSet` is used. Basically for checking whether the `typeString` matches one of these constants. Wouldn't it be possible to use `uintStringToTypeMap` for that purpose, since its keys are the same as in the set? Then the set could be removed.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -17,6 +17,26 @@ Author: Kunjian Song, kunjian.song@postgrad.manchester.ac.uk
...
 namespace SolidityGrammar
 {
+std::set<std::string> uintTypeStringSet = {
+  "uint8",   "uint16",  "uint24",  "uint32",  "uint40",  "uint48",  "uint56",
+  "uint64",  "uint72",  "uint80",  "uint88",  "uint96",  "uint104", "uint112",
+  "uint120", "uint128", "uint136", "uint144", "uint152", "uint160", "uint168",
+  "uint176", "uint184", "uint192", "uint200", "uint208", "uint216", "uint224",
+  "uint232", "uint240", "uint248", "uint256",
+};
+std::map<std::string, ElementaryTypeNameT> uintStringToTypeMap = {
```
fbrausse: Same as the comment in the header: can't this be `const`? Also, why is this symbol `uintStringToTypeMap` exported in the `SolidityGrammar` namespace? Couldn't it be local to this translation unit by defining it outside the namespace as `static`?

fbrausse (commented):
Some remarks about the declaration of symbols. As a general question: why did you choose `std::map` over `std::unordered_map`?

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: This is true, I will substitute the set checks with the map checks

rafaelsamenezes (approved):

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -198,6 +260,14 @@ const char *elementary_type_name_to_str(ElementaryTypeNameT type)
   }
 }
 
+unsigned int uint_type_name_to_size(ElementaryTypeNameT type)
+{
+  if(!uint_size_map.count(type))
+    assert(!"Size map missing for type");
+
+  return uint_size_map.at(type);
```
fbrausse, line 261: Since this throws an exception when `type` is not in the map, you don't need the `.count()` check above.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -252,6 +260,14 @@ const char *elementary_type_name_to_str(ElementaryTypeNameT type)
   }
 }
 
+unsigned int uint_type_name_to_size(ElementaryTypeNameT type)
+{
+  if(!uint_size_map.count(type))
+    assert(!"Size map missing for type");
```
mikhailramalho: when building with -DNDEBUG flag, this assertion will be removed, and the if guard will be applied to the return statement. Please use msg.error + abort here.

mikhailramalho (changes requested):

src/solidity-frontend/solidity_grammar.cpp
fbrausse: The `.at()` will throw in that case, so I don't think the msg.error + abort() is needed here.

src/solidity-frontend/solidity_convert.cpp
mikhailramalho: Right, but shouldn't you get the type of expression, instead of the RHS? Something like `expr["typeDescription"]`?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -65,21 +91,19 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
...
+
+    if(uint_string_to_type_map.count(typeString) || typeString == "bool")
     {
       // For state var declaration,
       return ElementaryTypeName;
     }
-    else if(
-      type_name["typeString"].get<std::string>().find("int_const") !=
-      std::string::npos)
+    else if(typeString.find("int_const") != std::string::npos)
```
fbrausse, line 101: Sorry, I'm not familiar with the Solidity language nor the translated JSON very much. Is it possible to define own type names there? Could I for instance define a `float` (if it exists) to be called `my_int_const`? If yes, then this condition using `.find()` will (probably wrongly) recognize my "typedef'ed" `float` as an `int_const`. Is that a problem that can occur with Solidity?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -65,21 +91,19 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
...
     {
       // For Literal, their typeString is like "int_const 100".
       return ElementaryTypeName;
     }
-    else if(
-      type_name["typeString"].get<std::string>().find("function") !=
-      std::string::npos)
+    else if(typeString.find("function") != std::string::npos)
```
fbrausse, line 106: Same as above, is there a way to define a type called `my_function` in Solidity?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -155,31 +179,35 @@ const char *type_name_to_str(TypeNameT type)
 // rule elementary-type-name
 ElementaryTypeNameT get_elementary_type_name_t(const nlohmann::json &type_name)
 {
+  std::string typeString = type_name["typeString"].get<std::string>();
   // rule unsigned-integer-type
-  if(type_name["typeString"] == "uint8")
+
+  if(
+    typeString.substr(0, 4) == "uint" &&
+    uint_string_to_type_map.count(typeString))
```
fbrausse: Why is the first condition needed when `.count()` already compares for equality (actually equivalence)?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -155,31 +179,35 @@ const char *type_name_to_str(TypeNameT type)
...
-  else if(
-    type_name["typeString"].get<std::string>().find("int_const") !=
-    std::string::npos)
-  {
-    // For Literal, their typeString is like "int_const 100".
-    // TODO: Fix me! For simplicity, we assume everything is unsigned int.
-    return UINT8;
-  }
-  else
+  if(typeString.find("int_const") != std::string::npos)
```
fbrausse, line 193: Again `.find()`: `my_int_const`?

fbrausse (commented):
Some questions about the type identifiers.

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: I believe solidity doesn't support type aliases, so these type strings are from a predefined set of keywords and cannot be user defined.

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: This is true the first condition is redundant, I will amend this.

fbrausse (approved):
> I believe solidity doesn't support type aliases, so these type strings are from a predefined set of keywords and cannot be user defined.

I see, thank you. Using equality instead of a substring test in these case seems more robust (this set of keywords might still change in the future) and consistent since all the `map.count()` checks use equality. As this PR is not about these conditions, I guess I'm fine with that.

LGTM.

correcthorsebatterystaple:
> > I believe solidity doesn't support type aliases, so these type strings are from a predefined set of keywords and cannot be user defined.
> 
> I see, thank you. Using equality instead of a substring test in these case seems more robust (this set of keywords might still change in the future) and consistent since all the `map.count()` checks use equality. As this PR is not about these conditions, I guess I'm fine with that.
> 
> LGTM.

To elaborate the type definitions on the AST show up as `int_const 4` for the literal integer 4, which is why there is a substring test as we assure that the type is a literal by checking that there is an int_const in there. I do, however, agree that this check should be done more robustly by either checking the parent node and verifying by that node that it is a literal. However, with the current architecture the type definition is passed around instead of the parent node which would require a refactor on how the type definitions are evaluated.

Issue #757, opened 2022-06-18
Segmentation fault in the serialization and de-serialization of goto program


ChenfengWei0:
Hi everyone,

I am working on fixing the goto program I/O interface, including 

- Serialization: writing goto program to binary file (new option `--output-goto`)
- De-serialization: read goto program from binary file  (existed option `--binary`)

Here are some bugs I have found and confirmed so far:

1. ~~During serialization, the `goto_program::instructiont` will be converted to `irept`. However, the property `irep.statement` was forgotten to be assigned. This can lead to a **segmentation fault** during the **migration** (`migrate_expr()`)~~
2. During de-serialization, esbmc will firstly read and rebuild the symbol table (`contextt`), yet the namespace (`namespacet`)  was forgotten to be initialed . This will lead to a **segmentation fault** during the **migration** (`migrate_expr()`) from `irept` to `goto_program::instructiont`, when executing the statement
    ```cpp
    migrate_namespace_lookup->lookup(init, sym) == false
    ```

ChenfengWei0:
I have tried to fix some of them. For now it can successfully pass some very basic test cases.

## Successful case

- Command
    ```sh
    $esbmc --output-goto test.c     # binary write
    $esbmc --binary test.goto       # binary read
    ```

- `test.c`
    ```cpp
    int main()
    {
        int x = 1;
        if (x < 2)
            x = 2;
        return x;
    }
    ```

- Result
    ```
    ESBMC version 6.10.0 64-bit x86_64 linux
    Target: 64-bit little-endian x86_64-unknown-linux with esbmc libc
    Reading GOTO program from file
    GOTO program creation time: 0.001s
    GOTO program processing time: 0.000s
    Starting Bounded Model Checking
    Symex completed in: 0.000s (12 assignments)
    Slicing time: 0.000s (removed 11 assignments)
    Generated 0 VCC(s), 0 remaining after simplification (1 assignments)
    BMC program time: 0.000s

    VERIFICATION SUCCESSFUL
    ```

## Failed case


- Command
  ```sh
    $csmith > random1.c
    $esbmc random1.c --output-goto
    $esbmc random1.goto --binary
    ```


Basically all C files generated by Csmith will cause some errors, inluding:

- **segmentation fault**

    ```sh
    ESBMC version 6.10.0 64-bit x86_64 linux
    Target: 64-bit little-endian x86_64-unknown-linux with esbmc libc
    Reading GOTO program from file
    GOTO program creation time: 0.059s
[...]

Issue #557, opened 2021-11-14
Support for GOTO input


rafaelsamenezes:
How is ESBMC behaving when dealing with GOTO input, from:

1. Our own c2goto, can we use ESBMC to export a GOTO program to be used later?
2. From CBMC, can we use the GOTO generated by/for it?

fbrausse:
c2goto patched (in order to also export the `__ESBMC_main` symbol) with
```diff
diff --git a/src/c2goto/c2goto.cpp b/src/c2goto/c2goto.cpp
index 5aa2dbb7c..c4e1dfebc 100644
--- a/src/c2goto/c2goto.cpp
+++ b/src/c2goto/c2goto.cpp
@@ -65,6 +65,8 @@ public:
       return 1;
     if(typecheck())
       return 1;
+    if(language_files.final(context))
+      return 1;
 
     std::ofstream out(
       cmdline.getval("output"), std::ios::out | std::ios::binary);
```
generates .goto files which can be read by `esmbc --binary`. However, due to c2goto serializing an empty goto-functions table (see below), esbmc fails on such inputs with
```
GOTO instruction type NO_INSTRUCTION_TYPE not handled in goto_symext::symex_step
```

Regarding CBMC, there are a few differences:

ESBMC's serialization format consists of 4 parts:
1. header `GBF`
2. GOTO-serialization version number 1
3. symbol table
4. goto-functions table

CBMC serialization:
1. headers `GBF`, `\x7fGBF`, `\x7fELF`, various other formats like ELF, Mach-O, ...
2. GOTO-serialization version number 5
3. symbol table
4. goto-functions table

I have not checked what differences there are regarding 3. and 4.

rafaelsamenezes:
Thanks for the report @fbrausse!

> c2goto patched (in order to also export the __ESBMC_main symbol) with

I don't think that the __ESBMC_main symbol is needed, as it can change based on the analysis type.

> I have not checked what differences there are regarding 3. and 4.

This will probably be the trick part.

fbrausse:
> This will probably be the trick part.

I agree. Is the CBMC commit known where ESBMC forked? That would make tracking these changes easier.

mikhailramalho:
On Sun, 2 Jan 2022, 10:02 Franz Brauße, ***@***.***> wrote:

> This will probably be the trick part.
>
> I agree. Is the CBMC commit known where ESBMC forked? That would make
> tracking these changes easier.
>

[...]

PR #758, opened 2022-06-18
Partially fixed goto program I/O interface

ChenfengWei0:
Fixes for #757 

This PR contains the following changes:
- Added a new option `--output-goto` for writing goto program to binary file.
- Fixed some existed bugs.

Both successful and failed test cases are provided in #757


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1684,6 +1685,21 @@ bool esbmc_parseoptionst::process_goto_program(
...
+      std::string gbfname = fullname.substr(0, lastindex) + ".goto";
+      std::ofstream oss(gbfname, std::ios::binary);
+      if(write_goto_binary(oss, context, goto_functions))
+      {
+        msg.error("fail to generate goto binary file");
+        abort();
+      };
+      return true;
+    }
     // show it?
```
mikhailramalho, line 1701: add extra line.

src/esbmc/options.cpp
```diff
@@ -15,6 +15,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"parse-tree-only", NULL, "only show parse tree"},
     {"parse-tree-too", NULL, "show parse tree and verify"},
     {"goto-functions-only", NULL, "only show goto program"},
+    {"output-goto", NULL, "writing goto program to binary file"},
```
mikhailramalho: can you move this closer to the option that we read a goto binary?

src/goto-programs/read_bin_goto_object.cpp
```diff
@@ -107,7 +107,8 @@ bool read_bin_goto_object(
     if(it == functions.function_map.end())
       functions.function_map.emplace(fname, message_handler);
     goto_functiont &f = functions.function_map.at(fname);
-    convert(t, f.body);
+    convert(t,f.body,context);
```
mikhailramalho: code style: add spaces after commas.

src/goto-programs/read_bin_goto_object.cpp
```diff
@@ -107,7 +107,8 @@ bool read_bin_goto_object(
...
+    //convert(t, f.body);
```
mikhailramalho: remove commented code.

src/goto-programs/goto_program_irep.cpp
```diff
@@ -78,7 +91,7 @@ void convert(const goto_programt &program, irept &irep)
   irep.hide(program.hide);
 }
 
-void convert(const irept &irep, goto_programt &program)
+void convert(const irept &irep, goto_programt &program, const contextt &context)
```
mikhailramalho: why do we need to pass the context here? It should be already filled when we reach the process_goto call.

mikhailramalho (changes requested):

src/goto-programs/goto_program_irep.cpp
ChenfengWei0: The context has indeed been filled when we reach process_goto call. The reason of passing it is to intitialize the namespacet pointer `migrate_namespace_lookup` in `migrate.cpp`.

lucasccordeiro:
@ChenfengWei0: could you please add some regression tests to this PR? You should add test cases that produce as output "VERIFICATION SUCCESSFUL" and "VERIFICATION FAILED".

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1684,6 +1685,22 @@ bool esbmc_parseoptionst::process_goto_program(
...
 
+    if(cmdline.isset("output-goto"))
+    {
+      msg.status("Writing GOTO program to file");
+      std::string fullname = cmdline.args[0];
+      // ! assume correct filename
+      size_t lastindex = fullname.find_last_of(".");
+      std::string gbfname = fullname.substr(0, lastindex) + ".goto";
+      std::ofstream oss(gbfname, std::ios::binary);
+      if(write_goto_binary(oss, context, goto_functions))
```
rafaelsamenezes, line 1693: Might be better to print the file output as well. `msg.status(fmt::format("Writing GOTO program into {}", gbfname));`

src/esbmc/options.cpp
```diff
@@ -84,6 +84,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"document-subgoals", NULL, "generate subgoals documentation"},
     {"no-arch", NULL, "don't set up an architecture"},
     {"no-library", NULL, "disable built-in abstract C library"},
+    {"output-goto", NULL, "writing goto program to binary file"},
```
rafaelsamenezes: I don't like this description. Maybe "export generated goto program" would be better.

rafaelsamenezes (changes requested):

src/util/migrate.cpp
```diff
@@ -533,6 +533,17 @@ expr2tc sym_name_to_symbol(irep_idt init, type2tc type)
...
 
+void migrate_expr2(
+  const exprt &expr,
+  expr2tc &new_expr_ref,
+  const contextt &context)
+{
+  // Initialize namespacet pointer
+  namespacet tmp(context);
+  migrate_namespace_lookup = &tmp;
+  migrate_expr(expr, new_expr_ref);
```
fbrausse: Please restore the previous value of `migrate_namespace_lookup` at the end of this function to avoid dangling pointers.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1684,6 +1685,22 @@ bool esbmc_parseoptionst::process_goto_program(
       return true;
     }
 
+    if(cmdline.isset("output-goto"))
+    {
+      msg.status("Writing GOTO program to file");
+      std::string fullname = cmdline.args[0];
+      // ! assume correct filename
+      size_t lastindex = fullname.find_last_of(".");
+      std::string gbfname = fullname.substr(0, lastindex) + ".goto";
```
fbrausse: Please don't modify the path the user passed as argument: you might be overwriting important data unrelated to this run of ESBMC. Generally speaking, making assumptions about filename extensions is not portable.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1684,6 +1685,22 @@ bool esbmc_parseoptionst::process_goto_program(
       return true;
     }
 
+    if(cmdline.isset("output-goto"))
+    {
+      msg.status("Writing GOTO program to file");
+      std::string fullname = cmdline.args[0];
```
fbrausse: Would it be possible instead of interpreting `args[0]` to give `--output-goto` a parameter that specifies the output path?

fbrausse (changes requested):
> Fixed some existed bugs.

Not sure this is what this PR does. To me it looks like incorporating the `c2goto` command's functionality. Why is this needed when there is this separate tool for goto conversion?

ChenfengWei0:
> @ChenfengWei0: could you please add some regression tests to this PR? You should add test cases that produce as output "VERIFICATION SUCCESSFUL" and "VERIFICATION FAILED".

Sorry about that! It was forgotten!

src/esbmc/options.cpp
ChenfengWei0: Thanks for your correction!

src/util/migrate.cpp
ChenfengWei0: Thank you for the correction!

ChenfengWei0:
> > Fixed some existed bugs.
> 
> Not sure this is what this PR does. To me it looks like incorporating the `c2goto` command's functionality. Why is this needed when there is this separate tool for goto conversion?

I will think about it. It appears that I have done some duplicated jobs.

src/esbmc/esbmc_parseoptions.cpp
ChenfengWei0: Could I ask are you suggest using a fixed filename? The implementation in `c2goto.cpp` is like
  ```cpp
    std::ofstream out(
      cmdline.getval("output"), std::ios::out | std::ios::binary);

    if(write_goto_binary(out, context, goto_functions))
    {
      msg.error("Failed to write C library to binary obj");
      return 1;
    }
  ```

fbrausse: No, don't use a fixed filename. I'm suggesting that the option
```c
{"output-goto", NULL, "writing goto program to binary file"},
```
instead of `NULL` takes a string-typed parameter to be provided by the user, e.g., `--output-goto my-file.goto`. You could then it use exactly like `cmdline.getval()` in the c2goto code you quoted, instead of deriving the name of the output file from `args[0]`.

Another reason I'm suggesting this is because ESBMC can take more than one input source file, e.g. `esbmc a.c b.c c.c`, in which case basing the output filename on `args[0]` seems like a (too) arbitrary choice.

ChenfengWei0: Oh, what a silly mistake I have made.  Thank you very much!

fbrausse:
> > > Fixed some existed bugs.
> > 
> > 
> > Not sure this is what this PR does. To me it looks like incorporating the `c2goto` command's functionality. Why is this needed when there is this separate tool for goto conversion?
> 
> I will think about it. It appears that I have done some duplicated jobs.

That's OK, maybe this way we may eventually be able to replace `c2goto` by `esbmc` in the build process and unify the interface of both in the future.

I'm just curious: do you get the same results when using c2goto instead of this new option to esbmc?

ChenfengWei0:
> > > > Fixed some existed bugs.
> > > 
> > > 
> > > Not sure this is what this PR does. To me it looks like incorporating the `c2goto` command's functionality. Why is this needed when there is this separate tool for goto conversion?
> > 
> > 
> > I will think about it. It appears that I have done some duplicated jobs.
> 
> That's OK, maybe this way we may eventually be able to replace `c2goto` by `esbmc` in the build process and unify the interface of both in the future.
> 
> I'm just curious: do you get the same results when using c2goto instead of this new option to esbmc?

No, I do not get the same results.




- `test.c`
  ```cpp
  #include<stdio.h>
  
  int main()
  {
      int x = 1;
      if (x < 2)
          x = 2;
      return x;
  }
  ```

  - The goto binary file generated by `c2goto` will cause a error. 
  - The `--output-goto` will not.

fbrausse:
@ChenfengWei0 Many thanks! I now remember that I had the same problem in https://github.com/esbmc/esbmc/issues/557#issuecomment-1003588122 where c2goto didn't include the `__ESBMC_main` symbol (which is fine) and `--binary` also didn't provide it. The latter might be considered a bug depending on what semantics we attach to it. This PR makes `--binary` be the counter-part to the new `--output-goto` while ESBMC cannot at present interpret the files c2goto outputs as programs to be verified due to the missing symbol.

lucasccordeiro:
@rafaelsamenezes: as we discussed today, could you please help @ChenfengWei0 with this PR?

ChenfengWei0:
Hi everyone,
In the newest commit, I think I have resolved the problems in the goto program I/O interface. More detailed explanations can be found here #757

fbrausse:
According to @rafaelsamenezes in https://github.com/esbmc/esbmc/issues/557#issuecomment-1003710083, the `__ESBMC_main` function's content can change based on the analysis type. `--output-goto` therefore encodes it in the output file which means changing it when running `--binary` is without effect. Maybe this behaviour (if it is desired) should be documented.

As an additional note: @ChenfengWei0 could you also put the .c sources the `main.goto` files were derived from into the regression tests, maybe including a comment (or a Makefile) what exactly the command-line of `esbmc --output-goto` was to create these .goto files?

ChenfengWei0:
> According to @rafaelsamenezes in [#557 (comment)](https://github.com/esbmc/esbmc/issues/557#issuecomment-1003710083), the `__ESBMC_main` function's content can change based on the analysis type. `--output-goto` therefore encodes it in the output file which means changing it when running `--binary` is without effect. Maybe this behaviour (if it is desired) should be documented.
> 
> As an additional note: @ChenfengWei0 could you also put the .c sources the `main.goto` files were derived from into the regression tests, maybe including a comment (or a Makefile) what exactly the command-line of `esbmc --output-goto` was to create these .goto files?

Sure. Thanks for the advice.

lucasccordeiro:
@ChenfengWei0: could you please rebase this PR with the master branch?

ChenfengWei0:
> @ChenfengWei0: could you please rebase this PR with the master branch?

Sure!

rafaelsamenezes (approved):

lucasccordeiro (changes requested):
@ChenfengWei0: could you please add the original C programs used to produce the goto files?

lucasccordeiro (approved):
LGTM.

lucasccordeiro:
@fbrausse: do you have further comments for this PR?

fbrausse (approved):
LGTM.

PR #759, opened 2022-06-21
[solidity] add int types

correcthorsebatterystaple:
Note:
This PR has branched off of (feature/add-solidity-uint-types) #754 which implements uint types and coercion for literal types to a specific uint type based on its context 

Changes:
- Add processing for int types in solidity
- Add regression tests for int types in solidity


lucasccordeiro (approved):

lucasccordeiro:
@rafaelsamenezes and @fbrausse: could you please review this PR?

fbrausse:
@lucasccordeiro I'd like to resolve the comments for the first uint-PR first. It seems this PR duplicates many of the declarations of the previous one and actually includes all changes of some version of the previous PR as well.

lucasccordeiro:
@correcthorsebatterystaple: could you please rebase this PR? Do you still have comments to implement?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -467,7 +475,7 @@ bool solidity_convertert::get_block(
   get_start_location_from_stmt(block, location);
 
   SolidityGrammar::BlockT type = SolidityGrammar::get_block_t(block);
-  msg.debug(fmt::format(
+  msg.status(fmt::format(
```
fbrausse: nah... debug is fine; btw. with `--verbosity 7` you can enable debug messages.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -819,6 +827,9 @@ bool solidity_convertert::get_expr(
     const nlohmann::json &literal = expr["typeDescriptions"];
     SolidityGrammar::ElementaryTypeNameT type_name =
       SolidityGrammar::get_elementary_type_name_t(literal);
+    msg.status(fmt::format(
```
fbrausse: `.debug()`, if at all; also below

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1040,26 +1051,28 @@ bool solidity_convertert::get_binary_operator_expr(
...
   exprt &new_expr)
 {
-  // 1. get type
+  // TODO: Fix me! Currently just support prefix == true,e.g. pre-increment
+  assert(expr["prefix"]);
+
+  // 1. get UnaryOperation opcode
+  SolidityGrammar::ExpressionT opcode =
+    SolidityGrammar::get_unary_expr_operator_t(expr, expr["prefix"]);
+  msg.status(fmt::format(
```
fbrausse: debug

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1357,7 +1375,7 @@ bool solidity_convertert::get_type_description(
   }
   default:
   {
-    msg.debug(fmt::format(
+    msg.status(fmt::format(
```
fbrausse: debug

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1522,7 +1598,7 @@ bool solidity_convertert::get_elementary_type_name(
...
```
fbrausse: debug

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -426,7 +491,35 @@ ExpressionT get_expression_t(const nlohmann::json &expr)
...
+  if(expr["operator"] == "-")
+  {
+    return UO_Minus;
+  }
+  assert(!((fmt::format(
+              "Got expression operator={}. Unsupported expression operator",
+              expr["operator"].get<std::string>()))
+             .c_str()));
+
+  return ExpressionTError; // make some old compilers happy
```
fbrausse: I don't understand. You're asserting `false` just above and now `make some old compilers happy`? What compilers are meant here? Also new compilers will happily complain if this function doesn't return anything despite the `assert`. Asserts get compiled out on `Release` builds, old and new compilers will not know what to do then.

Either
- leave as it is without the comment, or
- print the error message and `abort()` - old and new compilers will be happy, or
- use `__builtin_unreachable` (least preferrable option)

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -460,27 +553,9 @@ ExpressionT get_expr_operator_t(const nlohmann::json &expr, bool uo_pre)
...
-    {
-      return UO_PreDec;
-    }
-    else
-    {
-      assert(!"Unsupported - UO_PostDec");
-    }
-  }
-  else if(expr["operator"] == "++")
+  else if(expr["operator"] == "&&")
```
fbrausse, line 556: That seems to be the same case as just below.

fbrausse (changes requested):

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: This is due to a bad merge, will remove.

src/solidity-frontend/solidity_grammar.cpp
correcthorsebatterystaple: This is due to copy and paste from the source function which this has been extracted from. See below.

As the rest of the methods employ the assert I shall leave it in and remove the comment

fbrausse: Thanks!

src/solidity-frontend/solidity_convert_literals.cpp
```diff
@@ -21,19 +21,12 @@ bool solidity_convertert::convert_integer_literal(
...
-      integer2binary(z_ext_value, bv_width(type)),
-      integer2string(z_ext_value),
-      type);
-  }
-  else
-  {
-    assert(!"Unimplemented - Got signed type. Add sigend bv conversion");
-  }
+  // extract the value: unsigned
+  BigInt z_ext_value = string2integer(the_value);
```
fbrausse, line 25: Watch out here: `string2integer()` only deals with strings matching `[0-9]*`, a leading `-` is not interpreted correctly. We should eventually fix this.

fbrausse (changes requested):
Please revert all the msg.{debug => status} changes in this PR and also remove the `regression/int.patch` file.

src/solidity-frontend/solidity_convert_literals.cpp
correcthorsebatterystaple: Seems like this doesn't cause an issue and it detects the underflow on the int types correctly

fbrausse (approved):
LGTM!

Issue #752, opened 2022-06-15
Performance degradation for Z3


rafaelsamenezes:
The above is for the large array https://github.com/esbmc/esbmc/blob/c7191cda390244cabca0285d0c41a7ad2ffa673e/regression/esbmc/github_732-2-2/main.c

Other solvers do the `--smt-formula-only` much quicker and with less memory:
- Boolector 7.2s 144M
- CVC4 0.77s 152M
- Yices 7.1s 238M
- Bitwuzla 8.9s 146M
- Mathsat 1.4s 175M
- well... and Z3 116.7s 640M

_Originally posted by @fbrausse in https://github.com/esbmc/esbmc/issues/749#issuecomment-1156257025_

lucasccordeiro:
@fbrausse: can I ask you whether you could look at this issue? This is a low-priority issue.

fbrausse:
`perf report -g` output:
```
# Children      Self  Command  Shared Object                       Symbol                                                                                                                                                                                                       >
# ........  ........  .......  ..................................  .............................................................................................................................................................................................................>
#
    65.70%    65.65%  esbmc    libz3.so.4.8.17.0                   [.] ast_manager::~ast_manager
            |          
             --18.10%--ast_manager::~ast_manager

    19.85%    19.84%  esbmc    libz3.so.4.8.17.0                   [.] ast_mark::is_marked
            |          
             --4.55%--ast_mark::is_marked

    11.91%    11.91%  esbmc    libz3.so.4.8.17.0                   [.] ast_mark::mark
     5.11%     0.00%  esbmc    [unknown]                           [.] 0000000000000000
            |
            ---0
               |          
               |--2.64%--ast_mark::is_marked
               |          
                --2.05%--ast_manager::~ast_manager

     4.29%     0.00%  esbmc    [unknown]                           [.] 0x0000000000000001
            |
            ---0x1
               |          
               |--2.38%--ast_mark::is_marked
               |          
                --1.90%--ast_manager::~ast_manager

     0.43%     0.41%  esbmc    libz3.so.4.8.17.0                   [.] ast_manager::register_node_core
     0.35%     0.00%  esbmc    libc.so.6                           [.] __libc_start_call_main
     0.35%     0.00%  esbmc    esbmc                               [.] main
     0.35%     0.00%  esbmc    esbmc                               [.] parseoptions_baset::main
     0.35%     0.00%  esbmc    esbmc                               [.] esbmc_parseoptionst::doit
[...]

PR #760, opened 2022-06-21
[z3] delete all smt_asts before destroying the z3_ctx

fbrausse:
Fixes the performance regression noted in #752.

Some unknown interplay between Z3's context dtor mechanics and
the order in which we build some expressions, such as in concat_tree(),
leads to a huge slow-down of the Z3 context destruction.

By deleting all stored smt_asts early in ~z3_convt() we avoid these
mechanics and solving times becomes more normal. For --smt-formula-only
on regression/esbmc/github_732-2-2:

> Encoding to solver time: 0.486s
> 2.86user 0.17system 0:03.04elapsed 100%CPU (0avgtext+0avgdata 640052maxresident)k

Before it was 116 seconds and also 640M.

The memory issue still seems to be open.

### DIFF
--- a/src/solvers/z3/z3_conv.cpp
+++ b/src/solvers/z3/z3_conv.cpp
@@ -58,6 +58,12 @@ z3_convt::z3_convt(
   Z3_set_error_handler(z3_ctx, error_handler);
 }
 
+z3_convt::~z3_convt()
+{
+  // delete ASTs before destructing z3_ctx: this speeds up the latter, see #752
+  delete_all_asts();
+}
+
 void z3_convt::push_ctx()
 {
   smt_convt::push_ctx();
--- a/src/solvers/z3/z3_conv.h
+++ b/src/solvers/z3/z3_conv.h
@@ -29,6 +29,8 @@ public:
   void dump() const override;
 };
 
+/* Be sure to not make smt_convt a *virtual* base class: our dtor ~z3_convt()
+ * erases all smt_asts early. */
 class z3_convt : public smt_convt,
                  public tuple_iface,
                  public array_iface,
@@ -36,7 +38,7 @@ class z3_convt : public smt_convt,
 {
 public:
   z3_convt(const namespacet &ns, const optionst &options, const messaget &msg);
-  ~z3_convt() override = default;
+  ~z3_convt() override;
 
 public:
   void push_ctx() override;


lucasccordeiro:
Thanks, @fbrausse. Let's see how the CI will enjoy your PR ;-)

fbrausse:
Solving the same regression test with Z3 (without any --smt-formula-*) with this PR runs in 60s and uses 10.7G memory.

> VERIFICATION FAILED
Command exited with non-zero status 1
61.76user 1.99system 1:03.81elapsed 99%CPU (0avgtext+0avgdata 10724724maxresident)k
4736inputs+0outputs (30major+2927977minor)pagefaults 0swaps

On current master, it takes more than 20min (I stopped the process) and also 10.7G memory.

> VERIFICATION FAILED
^CCommand terminated by signal 2
1266.59user 2.10system 21:09.07elapsed 99%CPU (0avgtext+0avgdata 10724124maxresident)k
0inputs+0outputs (0major+2927941minor)pagefaults 0swaps

rafaelsamenezes:
@fbrausse Keep the test case as a THOROUGH it takes too much memory. Let's see the performance over all the other benchmarks.

mikhailramalho (approved):
Nice!

fbrausse:
To summarize so far:
In https://github.com/Z3Prover/z3/issues/6100 @NikolajBjorner confirmed that the kept-alive z3::exprs in `live_asts` constitute a bug (fixed by this PR) and @nunoplopes suggested that we should check whether any other z3::exprs remain when destroying the z3::context by using a debug build of Z3. I'll do that and report back.

fbrausse:
With the new https://github.com/Z3Prover/z3/commit/ab9aee189b7be07d6c0ad6eb218693c534b405d6 (probably going to be part of v4.8.18) both, mem and time performance of Z3 are much better:
```
$ /usr/bin/time src/esbmc/esbmc ../regression/esbmc/github_732-2-2/main.c --z3
[...]
Encoding to solver time: 0.498s
Runtime decision procedure: 4.566s
[...]
VERIFICATION FAILED
Command exited with non-zero status 1
5.23user 0.50system 0:05.73elapsed 99%CPU (0avgtext+0avgdata 1841016maxresident)k
0inputs+0outputs (0major+670948minor)pagefaults 0swaps
```
Thanks @NikolajBjorner!

There still is quite a huge discrepancy between the 1.8G used by ESBMC here and Z3 on the [log.smt2](https://github.com/esbmc/esbmc/files/8952977/log.smt2.txt) (**edit** Z3 only uses 140M in about the same time) generated when patched with
```diff
diff --git a/src/solvers/z3/z3_conv.cpp b/src/solvers/z3/z3_conv.cpp
index 038d64def9..ff64e54883 100644
--- a/src/solvers/z3/z3_conv.cpp
+++ b/src/solvers/z3/z3_conv.cpp
@@ -48,6 +48,7 @@ z3_convt::z3_convt(
             z3::tactic(z3_ctx, "simplify") & z3::tactic(z3_ctx, "smt"))
              .mk_solver())
 {
+  solver.set("smtlib2_log", "log.smt2");
   z3::params p(z3_ctx);
   p.set("relevancy", 0U);
   p.set("model", true);
```
but it is much better now.

@rafaelsamenezes is there a way to let the Windows CI run on Z3 master?

lucasccordeiro (approved):

Issue #637, opened 2022-02-18
recursive struct/union types


fbrausse:
```c
// struct T; /* uncomment for verification successful */

struct S {
	struct T *t;
};

struct T {
	struct S s;
};

int main()
{
	struct T t;
	t.s.t = &t;
}
```
aborts with
```
esbmc/src/pointer-analysis/value_set.cpp:1025: void value_sett::assign(const expr2tc&, const expr2tc&, bool): Assertion `base_type_eq(rhs->type, lhs_type, ns) || is_subclass_of(lhs_type, rhs->type, ns)' failed.
```
and
```c
// union U; /* uncomment for verification successful */

struct S {
	union U *u;
};

union U {
	struct S s;
};

int main()
{
	union U u;
	u.s.u = &u;
}
```
wrongly gives
```
State 1 file union-struct.c line 14 function main thread 0
----------------------------------------------------
Violated property:
  file union-struct.c line 14 function main
  dereference failure: array bounds violated


VERIFICATION FAILED
```
Both are related to how recursive struct/union types are handled in the clang_c_frontend: By the time the `union U *u` field declaration is resolved already an empty `struct S` symbol is sitting in the context, which gets copied into the corresponding member of `union U`.

fbrausse:
With #615 merged the second example now aborts at the same place as the first:
```
esbmc: esbmc/src/pointer-analysis/value_set.cpp:1050: void value_sett::assign(const expr2tc&, const expr2tc&, bool): Assertion `base_type_eq(rhs->type, lhs_type, ns) || is_subclass_of(lhs_type, rhs->type, ns)' failed.
Aborted
```

lucasccordeiro:
Current master is failing with this error message for the first program:

```
Assertion failed: (a == b), function assert_type_compat_for_with, file ../src/irep2/irep2_expr.cpp, line 383
```

fbrausse:
Both examples do. That's good: both problems are now caught a bit earlier. The reason is still the same though. For the first example:
```
(gdb) p a->dump()
struct
* members : 
  0: pointer
    * subtype : symbol
        * symbol_name : tag-struct T

* member_names : 
  0: t

* member_pretty_names : 
  0: t

* typename : struct S
* packed : false
```
and
```
(gdb) p b->dump()
struct
* members : 

* member_names : 

* member_pretty_names : 

* typename : struct S
* packed : false
```
[...]

PR #764, opened 2022-06-22
[clang-c-frontend] Use Clang to evaluate bit-field width

### LINKED ISSUE #732: Packed struct
Version: `75c79ad99d1adfcbedd57835b728ad9bc20c55c0`

Case 1:
> ``` bash
> esbmc example1.c
> Aborted (core dumped)
> ```

Case 2:
> ``` bash
> esbmc example2.c
> Segmentation fault (core dumped)
> ```

Case 3:
> ``` bash
> esbmc example2.c -D RANGE_CHECK_ERROR
> terminate called after throwing an instance of 'std::out_of_range'
>  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> Aborted (core dumped)
> ```

Example 1:
``` c
typedef union {
  struct {
    int
      x: 12,
      y: (52ULL - 44);
  };
  int z;
} s1;

typedef union {
  int x;
} s2;

typedef struct {
  s1 x;
} s3;

typedef struct __attribute__((__packed__)) {
  s2 x;
  char y;
  s3 z;
} s4;

int main() {
  s1 v1;
  v1.z = 0;
  v1.x = 0;

  s4 v4;
  s4* v4p = &v4;
  v4p->x.x = v1.z;

  if (v1.z != v4p->z.x.z) {
    v4p->z.x.z = v1.z;
  }

  return 0;
}
```

Example 2:
``` c
typedef union {
  int x;
} s1;

typedef union {
  struct {
    int x;
  };
  int y;
} s2;

typedef struct {
  s1 x;
} s3;

typedef struct __attribute((__packed__)){
  s2 x;
  char y;
  s3 z;
} s4;

typedef struct {
  char x[11712];
} s5 ;

typedef struct {
  union {
#ifndef RANGE_CHECK_ERROR
    s5 x;
#endif
  };
} s6;

typedef struct {
  s4 x;
  s6 y;
} s7;

s7 gv7;

int main() {
  s1 v1;
  s4* v4 = &gv7.x;

  v1.x = 0;
  v4->x.x = v1.x;

  if (v1.x != v4->z.x.x) {
    v4->z.x.x = v1.x;
  }

  return 0;
}
```

fbrausse:
**Edit:** Fixes the struct layout in case 1 of #732 where member `y` in
```c
  struct {
    int
      x: 12,
      y: (52ULL - 44);
  };
```
did not get assigned a width because `.cformat()` on such a sub-expr does not work. This caused the wrong struct layout mentioned in https://github.com/esbmc/esbmc/issues/732#issuecomment-1154904069.

This PR uses Clang to evaluate the width expression, which is guaranteed by C to be a constant one. Just in case Clang bugs out on this evaluation, we still print an error and fail gracefully.


fbrausse:
Coming to think about it, the packed struct access should actually be `VERIFICATION SUCCESSFUL` (and hence <https://github.com/esbmc/esbmc/pull/764/files#diff-bf5902eb02ddce360ecc8e4798fc865e1e50dafdee3864819d11a9ecb29fe242>
 a `KNOWNBUG`) as compilers are supposed to generate correctly aligned accesses for this, even without `--no-align-check`.

WDYT?

src/clang-c-frontend/padding.cpp
```diff
@@ -165,6 +165,9 @@ void add_padding(struct_typet &type, const namespacet &ns)
         it != components.end();
         it++)
     {
+      if(it->type().get_bool("#bitfield"))
+        assert(!it->type().width().empty());
+
```
mikhailramalho, line 171: you may want to rewrite this as `assert(it->type().get_bool("#bitfield") && !it->type().width().empty())` or merge the assertion into the next `if`: compiling with `-DNDEBUG` will remove this assertion from the TU.

mikhailramalho (approved):
LGTM, so a minor change is required.

src/clang-c-frontend/padding.cpp
fbrausse: `!A || B`, but sure, will do. WDYT about the packed struct access? Should it be KNOWNBUG and SUCCESSFUL?

fbrausse:
See for instance https://godbolt.org/z/5a45o8G1K where armv4 is the target which does not support unaligned large loads. GCC here translates `v4p->z.x.z` with repeated `ldrb`, i.e., single-byte loads. I'll mark the regression test `KNOWNBUG` and `SUCCESSFUL`.

src/clang-c-frontend/padding.cpp
fbrausse: Done, also moved some common sub-expressions to variables.

lucasccordeiro:
> See for instance https://godbolt.org/z/5a45o8G1K where armv4 is the target which does not support unaligned large loads. GCC here translates v4p->z.x.z with repeated ldrb, i.e., single-byte loads. 

@fbrausse: how does clang work?

fbrausse:
@lucasccordeiro [Same](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:28,endLineNumber:33,positionColumn:28,positionLineNumber:33,selectionStartColumn:28,selectionStartLineNumber:33,startColumn:28,startLineNumber:33),source:'typedef+union+%7B%0A++struct+%7B%0A++++int%0A++++++x:+12,%0A++++++y:+(52ULL+-+44)%3B%0A++%7D%3B%0A++int+z%3B%0A%7D+s1%3B%0A%0Atypedef+union+%7B%0A++int+x%3B%0A%7D+s2%3B%0A%0Atypedef+struct+%7B%0A++s1+x%3B%0A%7D+s3%3B%0A%0Atypedef+struct+__attribute__((__packed__))+%7B%0A++s2+x%3B%0A++char+y%3B%0A++s3+z%3B%0A%7D+s4%3B%0A%0Aint+foo(s4+*v4p)+%7B%0A++s1+v1%3B%0A++v1.z+%3D+0%3B%0A++v1.x+%3D+0%3B%0A/*%0A++s4+v4%3B%0A++s4*+v4p+%3D+%26v4%3B*/%0A++v4p-%3Ex.x+%3D+v1.z%3B%0A%0A++if+(v1.z+!!%3D+v4p-%3Ez.x.z)+%7B%0A++++v4p-%3Ez.x.z+%3D+v1.z%3B%0A++%7D%0A%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:armv7-clang1101,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'1',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O+-march%3Darmv4',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'armv7-a+clang+11.0.1+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:44.20106711597865,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+armv7-a+clang+11.0.1+(Compiler+%231)',t:'0')),k:22.465599550688005,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

lucasccordeiro (approved):

lucasccordeiro:
@fbrausse: shall we run this PR over the SV-COMP benchmarks?

fbrausse:
Running: https://github.com/esbmc/esbmc/actions/runs/2549886129

lucasccordeiro:
@fbrausse: could you also rebase this PR?

fbrausse:
done. Aborted previous run and running new: https://github.com/esbmc/esbmc/actions/runs/2550115859

fbrausse:
Didn't think it had any influence, but I was wrong.

Stats diff: https://github.com/esbmc/esbmc/pull/764/files#diff-eaa56d75b0df4075b5b34397706a4e5cd062a446c7250304bb873d6068d00601
Detailed diff:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-gh-762-gh-764/termination/results.2022-06-24_17-15-50.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-gh-762-gh-764/unreach-call/results.2022-06-24_17-15-54.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-gh-762-gh-764/valid-memsafety/results.2022-06-24_17-16-12.diff.html)

fbrausse:
I looked at the non-noise differences on my local machine for master (f7b75630d9a2b074d73621b324eeebf0648cde81) and this PR with both, Release and Debug builds:

| diff | note | master (Release / Debug) | This PR (Release / Debug) | path |
|---|---|---|---|---|
| timeout -> wrong | no change, same running time locally | wrong / AF1 | wrong / AF1 | `c/pthread-atomic/read_write_lock-2.i` |
| correct(248s) -> timeout | no change locally | correct(134s¹) / correct(136s¹) | correct(136s¹) / correct(136s¹) | `c/ldv-sets/test_mutex.i` |
| correct(216s) -> timeout | no change locally | correct(35s¹) / correct(40s¹) | correct(35s¹) / correct(40s¹) | `c/ldv-validator-v0.6/linux-stable-9ec4f65-1-110_1a-drivers--rtc--rtc-tegra.ko-entry_point.cil.out.i` |
| correct(40s) -> timeout | large time difference locally, need to take a closer look | correct(75s¹) / correct(80s¹) | correct(194s¹) / correct(200s¹) | `c/ldv-validator-v0.6/linux-stable-c0cc359-104_1a-drivers--usb--serial--qcserial.ko-entry_point.cil.out.i` |

¹ time measured locally on my laptop

Assertion Failures:
- AF1: pthread-related, item 1 in https://github.com/esbmc/esbmc/pull/743#issuecomment-1153116422
  ```
   esbmc/src/irep2/irep2_type.cpp:266: unsigned int struct_union_data::get_component_number(const irep_idt&) const: Assertion `0 && fmt::format( "Looking up index of nonexistant member \"{}\" in struct/union \"{}\"", comp, name) .c_str()' failed.
   ```
   with `name` being `"pthread_mutex_t"` and `comp` being `"__lock"`.

fbrausse:
Last item above seems to affect Boolector and Bitwuzla. With Yices, Z3 and CVC4 running time remains unchanged by this PR (2.9s¹, 23s¹ and 16s¹). Bitwuzla goes from 27.5s¹ (master) to 13.7s¹ (this PR).

Order of assignments get slightly changed for `--goto-functions-only`. Diff: [goto.diff.txt](https://github.com/esbmc/esbmc/files/8984571/goto.diff.txt)

In the SSA as well: `--symex-ssa-trace` diff: [symex-ssa.diff.txt](https://github.com/esbmc/esbmc/files/8984583/symex-ssa.diff.txt)

Differences for `--z3 --smt-formula-too`: [smt-formula.diff.txt](https://github.com/esbmc/esbmc/files/8984611/smt-formula.diff.txt)

I don't understand why the order of assignments would change. :(

fbrausse:
With this patch the GOTO reordering is gone and Boolector timings are much better (90s¹). Why? What side-effects does `get_start_location_from_stmt()` have that lead to this reordering of initializations inside `pthread_lib.c`?
```patch
index 75b3fa6487..26ad7536a8 100644
@@ -169,6 +169,9 @@ bool clang_c_convertert::get_decl(const clang::Decl &decl, exprt &new_expr)
         return true;
       }
 
+      locationt loc;
+      get_start_location_from_stmt(*fd.getBitWidth(), loc);
+
       comp.type().width(integer2string(result.Val.getInt().getSExtValue()));
       comp.type().set("#bitfield", true);
       comp.type().subtype() = t;
```

And with this one the Boolector timings are even completely back to master (75s¹):
```patch
index 75b3fa6487..a1880e77a2 100644
@@ -169,6 +169,10 @@ bool clang_c_convertert::get_decl(const clang::Decl &decl, exprt &new_expr)
         return true;
       }
 
+      exprt width;
+      if(get_expr(*fd.getBitWidth(), width))
+        return true;
+
       comp.type().width(integer2string(result.Val.getInt().getSExtValue()));
       comp.type().set("#bitfield", true);
       comp.type().subtype() = t;
```

fbrausse:
My guess would be that it's related to the way we handle type definitions and forward declarations #637.

fbrausse:
As discussed with @lucasccordeiro, I've restored the `get_expr()` call even though its result is not used anymore, just to get the Boolector timings back to master. There is a comment attached explaining the problem.

PR #767, opened 2022-06-24
Disable alignment check for packed structs

fbrausse:
This PR is on top of #764. It disables alignment checks during dereference when the dereferencing expression itself goes through a packed type. As this loosens the guarantees ESBMC provides (see also #766), I'll leave this a draft to decide whether this is the right way to go.

### DIFF
--- a/src/pointer-analysis/dereference.h
+++ b/src/pointer-analysis/dereference.h
@@ -199,13 +199,53 @@ public:
   virtual ~dereferencet() = default;
 
   /** The different ways in which a pointer may be accessed. */
-  typedef enum
+  struct modet
   {
-    READ,     /// The result of the expression is only read.
-    WRITE,    /// The result of the expression will be written to.
-    FREE,     /// The referred to object will be freed.
-    INTERNAL, /// Calling code only wants the internal value-set data.
-  } modet;
+    enum : unsigned
+    {
+      READ,     /// The result of the expression is only read.
+      WRITE,    /// The result of the expression will be written to.
+      FREE,     /// The referred to object will be freed.
+      INTERNAL, /// Calling code only wants the internal value-set data.
+    } op : 2;
+
+    /**
+     * Whether the access is performed in a non-standard, known-unaligned way
+     * as for, e.g., structures annotated with `__attribute__((packed))`.
+     */
+    unsigned unaligned : 1;
+
+    /* Intentionally provide no comparison operators in order to avoid counter-
+     * intuitive results when comparing, e.g., an unaligned mode with one of
+     * the (!unaligned) dereferencet::(READ|WRITE|FREE|INTERNAL) constants.
+     *
+     * For that the below predicates exist. */
+
+    friend bool is_read(const modet &m)
+    {
+      return m.op == READ;
+    }
+
+    friend bool is_write(const modet &m)
+    {
+      return m.op == WRITE;
+    }
+
+    friend bool is_free(const modet &m)
+    {
+      return m.op == FREE;
+    }
+
+    friend bool is_internal(const modet &m)
+    {
+      return m.op == INTERNAL;
+    }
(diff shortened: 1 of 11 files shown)


lucasccordeiro:
I think it is the right way to go since we want to enable ESBMC to verify C programs using packed-type support. This feature seems to be crucial to our collaborators at Intel.

@rafaelsamenezes and @mikhailramalho: what do you think?

src/pointer-analysis/dereference.h
```diff
@@ -199,13 +199,53 @@ class dereferencet
   virtual ~dereferencet() = default;
 
   /** The different ways in which a pointer may be accessed. */
-  typedef enum
+  struct modet
```
mikhailramalho, line 202: Can you make this a class with a proper constructor?

mikhailramalho (approved):
LGTM, just one change is required.

src/pointer-analysis/dereference.h
fbrausse: Sure, no problem with this one. May I ask why you rather have classes w/ constructors instead of structs, even if they're just POD/Trivial? Providing an own (non-constexpr) constructor makes those non-trivial and non-literal types, so that's easier to get wrong...

fbrausse: > Sure, no problem with this one.

Oh, I was wrong. It turns out making this a non-trivial type disallows defining the named constants as `constepxr` in the class. I'll either get
> error: 'constexpr dereferencet::modet::modet(dereferencet::modet::operationt, bool)' called in a constant expression before its definition is complete

due to the the call not being in a complete context; and moving the initializer out of the class does not work with constexpr:
> error: 'constexpr' static data member 'READ' must have an initializer

So an own constructor prevents having these constants neatly defined (unless I'm missing something).

lucasccordeiro (approved):

lucasccordeiro:

fbrausse:
@lucasccordeiro done. But I'll also add a few more regression checks from the examples mentioned at #766.

lucasccordeiro:
> @lucasccordeiro done. But I'll also add a few more regression checks from the examples mentioned at https://github.com/esbmc/esbmc/discussions/766.

Excellent!

src/pointer-analysis/dereference.h
mikhailramalho: > Sure, no problem with this one. May I ask why you rather have classes w/ constructors instead of structs, even if they're just POD/Trivial? Providing an own (non-constexpr) constructor makes those non-trivial and non-literal types, so that's easier to get wrong...

Mostly in the name of readability: a call like `{modet::READ, false};` to construct an object is easier to understand when we know what fields we are actually initializing. If it's too hard to create a constexpr constructor, I think we should at least use designated initializers to initialize the fields.

fbrausse: > designated initializers

Done. That's much better. I don't like pessimizing the classification of a type; it avoids optimization opportunities and (to me) is less clear.

fbrausse: Windows:
> error C7555: use of designated initializers requires at least '/std:c++20'

fbrausse:
@lucasccordeiro ok, done

@mikhailramalho sorry, I don't see a way to do be more explicit in C++14, designated initializers are also not supported, yet: <https://github.com/esbmc/esbmc/runs/7079020765?check_suite_focus=true#step:15:229>. But I believe it's not a big deal here, the type is literally just above the initialization; it's not a complicated type and members are documented.

lucasccordeiro:
@mikhailramalho: do you have further comments for this PR?

mikhailramalho:
> @mikhailramalho: do you have further comments for this PR?

Nope, I had already approved the PR when I left the comment.

PR #769, opened 2022-06-25
[config] rebased config into a Singleton

rafaelsamenezes:


src/util/config.cpp
```diff
@@ -8,10 +8,20 @@ Author: Daniel Kroening, kroening@kroening.com
...
 #include <regex>
+#include <thread>
 
 #include <util/config.h>
 
-configt config;
+configt *configt::config = nullptr;
+std::mutex configt::instance_mutex;
+
+configt *configt::get_instance()
```
lucasccordeiro, line 18: Can you write some comments about this method?

src/util/config.cpp
```diff
@@ -8,10 +8,20 @@ Author: Daniel Kroening, kroening@kroening.com
...
+{
+  std::lock_guard<std::mutex> lock(instance_mutex);
```
lucasccordeiro, line 20: Don't we have to call `unlock` somewhere?

lucasccordeiro (changes requested):

fbrausse:
Sorry Rafael, I don't understand the purpose of this PR.
- Why is there a mutex now? If you're aiming at multi-threading, wouldn't `thread_local` be the way to go?
- What benefit does invoking `*configt::get_instance()` have over plain `config`?
- How does hiding the configt constructor help unit-testing configt?

This patch makes the source clumsier and the compiled code slower.

src/util/config.cpp
rafaelsamenezes: No need, it is released automatically when the lock goes out of scope.

rafaelsamenezes:
> Why is there a mutex now? If you're aiming at multi-threading, wouldn't thread_local be the way to go?

The mutex is only to prevent a data race over the creation of the config. I can remove if needed.

> What benefit does invoking *configt::get_instance() have over plain config?

Using it should only be the lazy initialization. For regressions, it can help to check whether we initialized things properly before using them. They add guarantees that only one object is alive at any given time, help avoid the initialization order fiasco, and do not pollute the namespace. 

> How does hiding the configt constructor help unit-testing configt

It won't, they have a similar problem of globals regarding tests. It might make testing around it easier (same motives above).

> This patch makes the source clumsier and the compiled code slower.

I am not sure about the code slower part, it seems to have the same performance overall. What is your suggestion then? 

1. Leave as a global?
2. Remove the mutex?
3. Refactor it as an DI?
4. Other?

fbrausse:
Sorry for the long post, I tried to be precise on why I don't agree with the path chosen in this PR.

> > Why is there a mutex now? If you're aiming at multi-threading, wouldn't thread_local be the way to go?
> 
> The mutex is only to prevent a data race over the creation of the config. I can remove if needed.

We don't have multiple threads at the moment. This data race seems to be under the assumption that multiple threads would work on the same config. I don't see at this point why that would be the case. It is entirely feasible to run multiple threads checking code with different configs (architectures, endiannesses, etc.) simultaneously. Unless this option is ruled out, the notion of config being a "singleton" throughout ESBMC may not be true. I would definitely suggest to remove the mutex until we are sure what it protects against.

> > What benefit does invoking *configt::get_instance() have over plain config?
> 
> Using it should only be the lazy initialization.

Why?

> For regressions, it can help to check whether we initialized things properly before using them.

But this patch is not doing that. The result of `*configt::get_instance()` is as well-initialized as the original `config` object. In fact, the config cannot be in a valid state consistent with the rest of the run of ESBMC without external information like that passed to `configt::set()`. I would understand making this function the constructor to solve this problem, but again, it's not what this PR does.

> They add guarantees that only one object is alive at any given time,

There are reasons not to have this guarantee, e.g., the above about multi-threading. Seems speculative to enforce it now. In current master, there also is only one `config` object throughout ESBMC.

> help avoid the initialization order fiasco,

Could you be more concrete? The [init order fiasco](https://en.cppreference.com/w/cpp/language/siof) (as I understand it) refers to the unspecified order in which constructors of static-/thread-storage objects in different TUs are run. Where do we face this problem in ESBMC? As far as I can see, no other constructors of static storage objects rely on the `config` already being initialized. If that was the case, a first measure is to put the definition of those symbols into the same TU in the right order, and only if that doesn't help (e.g. mutually dependent objects) resort to convoluting the API.

> and do not pollute the namespace.

I agree that global namespace pollution should be avoided and this PR achieves hiding the `config` symbol symbol from it. However, none of those "should" advices is 100%. There is always a tradeoff. `config` is used pervasively throughout ESBMC and having it global (for easier access) to my mind outweighs the rather theoretical benefits of a non-polluted namespace. In some sense, one could even argue that `configt` should be called `esbmct` as it is the core dependency of all the modules this patch touches. My first idea about reducing global namespace pollution would be to put everything into a `namespace esbmc` to facilitate easier linking of `esbmc` against 3rd party libs.

> > How does hiding the configt constructor help unit-testing configt
> 
> It won't, they have a similar problem of globals regarding tests. It might make testing around it easier (same motives above).

Indeed, it's still a globally accessible object with this PR. That's why I don't understand this PR's purpose.

> > This patch makes the source clumsier and the compiled code slower.
> 
> I am not sure about the code slower part, it seems to have the same performance overall.

So you agree with clumsier? ;) As for slower: a) there is a mutex involved in every access, b) `get_instance()` returns a pointer and is not inlined anywhere, so unless we rely on LTO the compiler will not know where it points to and what is.

> What is your suggestion then?
>     1. Leave as a global?
>     2. Remove the mutex?
>     3. Refactor it as an DI?
>     4. Other?

Mostly 2. and 3. `messaget` already gets injected nearly everywhere. My vision for a more library-like ESBMC would be to make `messaget` part of `configt` (and many of the things you mentioned in #755, as they conceptually belong the "the current execution" controlled by the config), inject this `config` whereever needed and put everything with external linkage into an `esmbc` namespace.

If a more library-like ESBMC is the goal of #755, we should probably discuss that there. At present the goal is not clear to me.

rafaelsamenezes:
> It is entirely feasible to run multiple threads checking code with different configs (architectures, endiannesses, etc.) simultaneously.

Right now it is impossible and unfeasible.

If we want it to be feasible, then we should discard singleton and the global anyway.

> The [init order fiasco](https://en.cppreference.com/w/cpp/language/siof) (as I understand it) refers to the unspecified order in which constructors of static-/thread-storage objects in different TUs are run. Where do we face this problem in ESBMC

Right now, we do not have it anymore (search past commits). Having a global like this, can add the behavior. Which we should avoid, unless we put in a manual to warn users that they shouldn't rely on it.

> So you agree with clumsier? ;) 

I do not. :kissing_heart:

Obviously, I wouldn't add code that I think is clumsy.

> As for slower: a) there is a mutex involved in every access,

This can be removed

> get_instance() returns a pointer and is not inlined anywhere, so unless we rely on LTO the compiler will not know where it points to and what is.

I see. Not sure if this even makes a difference in the end though. 

> Mostly 2. and 3. messaget already gets injected nearly everywhere. My vision for a more library-like ESBMC would be to make messaget part of configt (and many of the things you mentioned in https://github.com/esbmc/esbmc/issues/755, as they conceptually belong the "the current execution" controlled by the config), inject this config whereever needed and put everything with external linkage into an esmbc namespace.

That sounds a better plan. The hard part is that there are lots of "C API" in ESBMC and some parts rely in globals directly. This separation will be harder than messaget. 

> If a more library-like ESBMC is the goal of https://github.com/esbmc/esbmc/issues/755, we should probably discuss that there.

Yes, I agree. These remarks would be great last week :wink:

rafaelsamenezes:
Closing it as this needs more discussion

fbrausse:
> If we want it to be feasible, then we should discard singleton and the global anyway.

Discard singleton yes, global not necessarily. One option requiring comparatively little changes (still not fully library-like, though) could be
```c++
extern thread_local configt config;
```

I've previously taken this approach with [another](https://github.com/fbrausse/iRRAM) framework -> library conversion.

> > [init order fiasco](https://en.cppreference.com/w/cpp/language/siof)
> 
> Right now, we do not have it anymore (search past commits). Having a global like this, can add the behavior. Which we should avoid, unless we put in a manual to warn users that they shouldn't rely on it.

I see, thanks for the pointer. Yeah, it is a subtle "feature" of the language. 

> These remarks would be great last week

You are right, I'm sorry. It must have taken lots of effort changing the config throughout.

PR #773, opened 2022-06-29
[crypto_hash] fix buffer over-read

fbrausse:
The SHA-1 hash backing crypto_hash only has 160 bits, not 256.
Removing the duplicated information about the hash size.
Fixes:

  ==31812== Invalid read of size 16
  ==31812==    at 0x2C3027: memcpy (string_fortified.h:29)
  ==31812==    by 0x2C3027: execution_statet::state_hashing_level2t::generate_l2_state_hash() const (execution_state.cpp:1335)
  ==31812==    by 0x2C3CEC: execution_statet::generate_hash() const (execution_state.cpp:1089)

### DIFF
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -1325,22 +1325,20 @@ crypto_hash
 execution_statet::state_hashing_level2t::generate_l2_state_hash() const
 {
   unsigned int total;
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):
Well spotted!

rafaelsamenezes (approved):

Issue #776, opened 2022-06-29
Support for overflow check with unsigned types and shift operations


correcthorsebatterystaple:
Currently ESBMC supports an option `--overflow-check`, this option prompts ESBMC to add assertions to check for overflows in number types. However, there is no support yet for unsigned types which can be confirmed by passing the following C code to ESBMC with the option `--overflow-check`.
```c
int main() {
  unsigned char i = 255;
  i++;
  return 0;
}
```
Which yeilds a `Verification Successful`.

Adding overflow check for unsigned types is not difficult as all is requires it to add the unsignedbv type as a valid type to add assertions for.

The issue is to do with shift operations, in that there exist conflicting tests in `regression/esbmc` where it is expected that both shift operations are not checked for overflows and that shift operations are checked for overflows.

- [Here](https://github.com/esbmc/esbmc/blob/master/regression/esbmc/overflow_17_crc_new/main.c#L62) it is assumed that overflow check DOESN’T add assertions for shift operators
- [Here](https://github.com/esbmc/esbmc/blob/master/regression/esbmc/github_578_fail4/main.c) it is assumed that overflow check DOES add assertions for shift operators

Also there exist tests where once we add the overflow check assertions for unsigned types, those tests fail because of multiplication operations and other other operations on unsigned types.

- [Here](https://github.com/esbmc/esbmc/blob/master/regression/esbmc/overflow_16_fir_new/main.c#L117) unsigned long overflow is ignored even though the test is intended to run with `--overflow-check`

As unsigned types can indeed overflow so it is correct to check for overflows on unsigned types, however, one of these tests will need to be changed or removed depending what is decided to do with "overflows" on shift operations.

Please advise how one should proceed.

fbrausse:
The difference between signed and unsigned overflow is that according to the C standard unsigned overflow is defined as wrap-around while signed overflow is undefined behaviour (UB). And compilers make use of this UB in assuming it can never happen. This UB is what the `--overflow-check` checks.

Indeed, as you argue, unsigned overflow can happen, too, and maybe it should be possible for ESBMC to recognize it. But this use-case to my knowledge has not yet been requested.

correcthorsebatterystaple:
@fbrausse For other frontends, namely solidity, they are both classed as overflows and I would like to be able to detect both via esbmc. What do you suggest I do in this case, should I add some sort of flag to check if the frontend is solidity and then perform the overflow check?

mikhailramalho:
[...]

PR #779, opened 2022-06-29
Feature/add solidity overflow check

correcthorsebatterystaple:
PR implements changes suggested in issue https://github.com/esbmc/esbmc/issues/776

Changes:
- Add language string to `configt`
- Add overflow checks for unsigned types for solidity only
- Add regression tests for automated overflow checks on both signed and unsigned types


lucasccordeiro:
@correcthorsebatterystaple: Could you please check these regression tests?

```
he following tests FAILED:
	789 - regression/esbmc-solidity/int_overflow_check (Failed)
	793 - regression/esbmc-solidity/uint_overflow_check (Failed)
```

fbrausse (commented):
Personally, I'm not a big fan of storing computed information in strings. An enum would be much nicer as it is easier to compare/`switch` on, avoids dynamic allocation, can be used as an index into arrays, has all possible values explicitely and centrally defined, etc.

Doing that properly would probably require cleaning up the `mode_table` handling, though, so I'm fine with this way now. We can change it later.

Could you add a comment to the new `configt::language` member explaining what it is?

For the failing tests you might want to enable the `--overflow-check`. ;)

correcthorsebatterystaple:
> Personally, I'm not a big fan of storing computed information in strings. An enum would be much nicer as it is easier to compare/`switch` on, avoids dynamic allocation, can be used as an index into arrays, has all possible values explicitely and centrally defined, etc.
> 
> 
> 
> Doing that properly would probably require cleaning up the `mode_table` handling, though, so I'm fine with this way now. We can change it later.

I must agree I too am not a big fan of magic strings. However, as you said would require a larger refactor to have that be consistent with the mode table implementation. 

> 
> 
> 
> Could you add a comment to the new `configt::language` member explaining what it is?

Will add

> For the failing tests you might want to enable the `--overflow-check`. ;)

Oops

lucasccordeiro (approved):
LGTM.

fbrausse (approved):

PR #780, opened 2022-06-30
[mode] factor out language-ID from the mode_table

fbrausse:
Give the supported languages proper IDs so we can use the language information
down the line. Before, the language/frontend was identified by a free-form
string only, however, in #776 it was deemed necessary to assign a meaning to
options like `--overflow-check` that depend on the concrete language we're
verifying over.

An additional benefit is that the name of frontends now is independent of the
language they're translating, allowing us to freely assign and show them in the
log later on.

The `mode_table` itself for now retains its weird semantics wrt. the old
frontends.

Further, we get rid of some globally exported symbols and mark others `const`
like `buildidstring_buf` and `esbmc_version_string`, so they do not count
toward global "state" #755.

### DIFF
--- a/src/c2goto/c2goto.cpp
+++ b/src/c2goto/c2goto.cpp
@@ -97,5 +97,5 @@ int main(int argc, const char **argv)
 }
 
 const mode_table_et mode_table[] = {
-  LANGAPI_HAVE_MODE_CLANG_C,
-  LANGAPI_HAVE_MODE_END};
+  LANGAPI_MODE_CLANG_C,
+  LANGAPI_MODE_END};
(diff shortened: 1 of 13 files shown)


lucasccordeiro (approved):

lucasccordeiro:
@fbrausse: Many thanks for writing this PR!

kunjsong01 (approved):
LGTM

mikhailramalho (commented):
I noticed that when creating languagest objects (e.g., during witness gen), we force always the language_idt::C, shouldn't we get that from the config file?

For instance, when generating C++ witnesses, references with be printed as pointer if we use the C expr dump (expr2c) instead of the cpp one (expr2cpp).

fbrausse:
@mikhailramalho good point, I did a mostly mechanical refactoring and not think about it in depth. Let's investigate this in #782.

Issue #775, opened 2022-06-29
Memory safety: addresses of free'd objects are excluded from the address space


fbrausse:
This affects the memsafety benchmark [c/memsafety/cmp-freed-ptr](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/memsafety/cmp-freed-ptr.yml) ([C source](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/memsafety/cmp-freed-ptr.c)) which ESBMC wrongly recognizes as `VERIFICATION SUCCESSFUL`:

```c
#include <stdlib.h>
#include <stdint.h>

int main() {

    struct T {
        struct T* next;
    };

    struct T* x = NULL;
    struct T* y = NULL;

    y = malloc(sizeof(*y));
    intptr_t adressY = (intptr_t) y;

    free(y);

    x = malloc(sizeof(*x));
    intptr_t adressX = (intptr_t) x;

    if (adressX == adressY)
    { // if the second malloc returns the same value as the first, I should get here
        free(x);
    }

    free(x);

    return 0;
}

// predator-regre/test-0238.c
// - comparing freed pointers
// - contributed by Ondra Lengal
```

> `esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --k-step 2 --floatbv --unlimited-k-steps sv-benchmarks/c/memsafety/cmp-freed-ptr.c --32  --memory-leak-check`

The formula generated contains
```smt2
(assert (or (bvult __ESBMC_ptr_obj_end_3 __ESBMC_ptr_obj_start_2)
    (bvugt __ESBMC_ptr_obj_start_3 __ESBMC_ptr_obj_end_2)))
```
asserting that the (free'd) object `*y` and the existing object `*x` do not overlap. A guarantee not given by `malloc()`. Indeed, removing this assertion from the formula makes it SAT and ESBMC thus correctly finding the memsafety violation.

This assertion is generated in `smt_convt::finalize_pointer_chain()` unconditionally for all previous objects. If it was guarded by a condition saying "object is still alive", I believe this would fix the issue.

lucasccordeiro:
>  If it was guarded by a condition saying "object is still alive", I believe this would fix the issue.

I agree.

fbrausse:
[...]

PR #781, opened 2022-06-30
[smt] plug free'd hole

fbrausse:
Quick'n'dirty check to see what SV-COMP says for this attempt on #775.


fbrausse:
Master:
```
  correct:            8454
    correct true:     5127
    correct false:    3327
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7141
```
This PR:
```
  correct:            8445
    correct true:     5132
    correct false:    3313
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7150
  Score:             12345 (max: 25567)
```
Diff:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-781/termination/results.2022-07-01_15-57-11.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-781/unreach-call/results.2022-07-01_15-57-13.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-781/valid-memsafety/results.2022-07-01_15-57-21.diff.html)

fbrausse:
This q&d patch indeed fixes the benchmark in #775 (and also takes care of my "steal address space" example there).

The reason why #incorrect did not go down is due to `c/pthread-atomic/read_write_lock-2.i` (timeout -> wrong). This is not new, see <https://github.com/esbmc/esbmc/pull/764#issuecomment-1166255017>.

src/solvers/smt/smt_memspace.cpp
```diff
@@ -435,7 +435,16 @@ void smt_convt::finalize_pointer_chain(unsigned int objnum)
     lessthan2tc lt1(end_i, start_j);
     greaterthan2tc gt1(start_i, end_j);
     or2tc or1(lt1, gt1);
-    assert_expr(or1);
+
+    expr2tc e = or1;
+    if(current_valid_objects_sym)
```
lucasccordeiro: Can you add some comments?

src/solvers/smt/smt_conv.cpp
```diff
@@ -1347,13 +1347,15 @@ smt_astt smt_convt::convert_terminal(const expr2tc &expr)
   }
   case expr2t::symbol_id:
   {
+    const symbol2t &sym = to_symbol2t(expr);
+    std::string name = sym.get_symbol_name();
+    if(sym.thename == "c:@__ESBMC_alloc")
```
lucasccordeiro, line 1357: Shall we also add comments here?

lucasccordeiro (approved):
I have just some minor comments.

lucasccordeiro:

lucasccordeiro:
My initial suggestion was to implement this via symex (https://github.com/esbmc/esbmc/issues/775).

src/solvers/smt/smt_conv.cpp
fbrausse: Yes, done.

src/solvers/smt/smt_memspace.cpp
fbrausse: Done.

fbrausse:
I believe this patch is quite fragile, as explained in the new comments. Maybe there is a way to make it less fragile, e.g., if `smt_convt` had access to the current version of the `__ESMBC_alloc` symbol where the new object allocation is to be registered with. I don't know how to obtain that symbol there, though. That's why this patch uses "the last one seen", which can easily break if we change the order of `convert_terminal()` calls. It is also hard to reason about due to the `smt_cache` which prevents `convert_terminal()` from being called...

And I would agree with Lucas that symex is a better place for the things `finalize_pointer_chain()` does. Any help / hints on those items would be greatly appreciated!

mikhailramalho:
> I believe this patch is quite fragile, as explained in the new comments. Maybe there is a way to make it less fragile, e.g., if `smt_convt` had access to the current version of the `__ESMBC_alloc` symbol where the new object allocation is to be registered with. I don't know how to obtain that symbol there, though. That's why this patch uses "the last one seen", which can easily break if we change the order of `convert_terminal()` calls. It is also hard to reason about due to the `smt_cache` which prevents `convert_terminal()` from being called...
> 
> And I would agree with Lucas that symex is a better place for the things `finalize_pointer_chain()` does. Any help / hints on those items would be greatly appreciated!

Yeah, it's a hard problem to solve because smt_conv (as the name implies) only converts stuff, and has no clue of the current state.

It's similar to some cases where we generate typecasts when building the cex (which also goes through smt_conv) and we need to know the value of the rounding mode.

fbrausse:
There is `VERIFICATION FAILED` right here: <https://github.com/esbmc/esbmc/runs/7162947063?check_suite_focus=true#step:17:6627> as expected by <https://github.com/esbmc/esbmc/blob/3eb0a4f154d49f9f67b825d950aed55244ff7c19/regression/csmith/csmith01/test.desc#L4>

Why is this test failing?

**Edit** oh, it was a KNOWBUG...

fbrausse:
I would expect the verdict with `--32 --force-malloc-success` when the objects are not free'd
```c
#include <stdlib.h>
#include <assert.h>

void steal_addr_space(size_t n)
{
	char *a = malloc(n);
	// free(a);
}

int main()
{
	size_t MiB = 1ULL << 20;
	/* 16 * 256 MiB = 4 GiB */
	for (int i=0; i<16; i++)
		steal_addr_space(256 * MiB);
	assert(0);
}
```
to be `VERIFICATION SUCCESSFUL`, because the assumption that `malloc()` could return more than is addressable is wrong here. Do you agree?

But now slicing makes a difference: it removes all object allocations (as they are unused here), s.t. with this PR esbmc reports `FAILED` at the `assert(0)`. With `--no-slice` it is `SUCCESSFUL`. Is this behaviour intended? What semantics does `--no-slice` have with respect to allocations and to `--force-malloc-success`?

fbrausse:
Adding an assert:
```c
#include <stdlib.h>
#include <assert.h>

void steal_addr_space(size_t n)
{
	char *a = malloc(n);
	assert(a);
	// free(a);
}

int main()
{
	size_t MiB = 1ULL << 20;
	/* 16 * 256 MiB = 4 GiB */
	for (int i=0; i<16; i++)
		steal_addr_space(256 * MiB);
	assert(0);
}
```
It is `SUCCESSFUL` when run with only `--32` and `FAILED` when run with `--32 --no-slice` :(

fbrausse:
Diff between `--no-slice` and not giving that option of `--z3 --smt-formula-only` for the second example: [diff.txt](https://github.com/esbmc/esbmc/files/9034429/diff.txt)

Appearently the slicer removes the `__ESBMC_alloc` symbol (among others) entirely. This is most likely a side-effect of implementing `finalize_pointer_chain()` in smt_conv instead of symex: the slicer cannot see that this symbol is used anywhere.

lucasccordeiro:
It would be great if we could move all those checks from finalize_pointer_chain to symex.

fbrausse:
>> Appearently the slicer removes the __ESBMC_alloc symbol (among others) entirely. [...] the slicer cannot see that this symbol is used anywhere.

This is in the example with `assert(a)`. Only the `symex_dynamic::dynamic_*_array` symbols remain. My understanding is that this is really a problem between the slicer and the way the dynamic allocations are tracked with symbols unrelated to each other. Slicing any of them out makes the tracking inconsistent.

Since this was quick-and-dirty from the start, another approach (just to think about) would be to teach the slicer that all the symbols related to tracking dynamic allocations belong together and to remove either none or all together.

Maybe this outcome could even easily be obtained by putting all this info into one symbol (array of structs) instead of keeping separate arrays; then the slicer does not need to know about symbols related dynamic allocation and cannot remove the struct updates unless they're really unused?

> It would be great if we could move all those checks from finalize_pointer_chain to symex.

Any of these approaches (including putting it in symex) would (compared to this PR) likely lead to more complicated formulas involving more symbols also in easy cases involving dynamic allocation ~- basically for the only reason to check whether all allocated objects fit into the address space. This reason seems unrelated to the initial example from #775 and to only concern the various "steal address space" examples. If so, it could be worth having these more expensive checks en-/disabled by an option.~

fbrausse:
> basically for the only reason to check whether all allocated objects fit into the address space.

This statement is wrong. When removing both `free(x)` from the initial example in #775:
```diff
--- a.c	2022-07-03 09:19:15.326467503 +0200
+++ b.c	2022-07-03 12:32:01.233539370 +0200
@@ -22,11 +22,9 @@
 
     if (adressX == adressY)
     { // if the second malloc returns the same value as the first, I should get here
-        free(x);
+        assert(0);
     }
 
-    free(x);
-
     return 0;
 }
```
Without `--no-slice` we get `VERIFICATION SUCCESSFUL` and with the option it is `FAILED`.

Please ignore my last paragraph in the previous comment.

fbrausse:
Closed in favour of #788.

PR #787, opened 2022-07-04
[dyn-mem] explicitely unify interpretations of "valid"/"alloc" and "dealloc" expressions

### LINKED ISSUE #774: Duplicated information in `__ESBMC_alloc` and `__ESBMC_deallocated`?
It seems that both arrays tracking whether objects have been de-/allocated store equivalent information, that is, `__ESBMC_alloc[i] == true` <-> `__ESBMC_deallocated[i] == false`.

They're updated at the same time as well: https://github.com/esbmc/esbmc/blob/e34eacf04f4103962d18d0cd37292072d9dbe489/src/goto-symex/builtin_functions.cpp#L243-L248 and https://github.com/esbmc/esbmc/blob/e34eacf04f4103962d18d0cd37292072d9dbe489/src/goto-symex/builtin_functions.cpp#L342-L347

Can we remove one of them? This would simplify the formulas ESBMC generates.

fbrausse:
"valid_object" and !"deallocated_obj" have been functionally equivalent
already as they're always updated together. Both have been backed by
symbolic arrays indexed by the object number.

The `__ESBMC_deallocated` symbol goes back to
7e1fb8e39c33fbe16a06799846d8786dec7fc845, where it was introduced appearently
to give these two expressions a more orthogonal meaning, presumably with the
intent to make stack symbols "valid objects" at some point. This never
happened: only dynamic objects are ever "valid".

Removes the `__ESBMC_deallocated` symbol and replace the "deallocated_obj"
expression in symex with the equivalent to "not(valid)": `!__ESBMC_alloc[obj]`.

Fixes #774.

SV-COMP benchmarks [running](https://github.com/esbmc/esbmc/actions/runs/2608375982) just to double-check that it really doesn't break anything.

### DIFF
--- a/src/ansi-c/ansi_c_language.cpp
+++ b/src/ansi-c/ansi_c_language.cpp
@@ -41,7 +41,6 @@ static void internal_additions(std::string &code)
     // malloc / alloca
     "unsigned __ESBMC_constant_infinity_uint;\n"
     "_Bool __ESBMC_alloc[__ESBMC_constant_infinity_uint];\n"
-    "_Bool __ESBMC_deallocated[__ESBMC_constant_infinity_uint];\n"
     "_Bool __ESBMC_is_dynamic[__ESBMC_constant_infinity_uint];\n"
     "unsigned long __ESBMC_alloc_size[__ESBMC_constant_infinity_uint];\n"
 
--- a/src/clang-c-frontend/clang_c_language.cpp
+++ b/src/clang-c-frontend/clang_c_language.cpp
@@ -309,9 +309,6 @@ __attribute__((annotate("__ESBMC_inf_size")))
 _Bool __ESBMC_alloc[1];
 
 __attribute__((annotate("__ESBMC_inf_size")))
-_Bool __ESBMC_deallocated[1];
-
-__attribute__((annotate("__ESBMC_inf_size")))
 _Bool __ESBMC_is_dynamic[1];
 
 __attribute__((annotate("__ESBMC_inf_size")))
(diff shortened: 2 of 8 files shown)


mikhailramalho (approved):
Nice!

lucasccordeiro (approved):

fbrausse:
I just noticed this adds an implicit requirement (which we appearently meet): Only for already allocated dynamic objects does `deallocated_obj2t` now return the correct result. We could translate it as `!valid /\ dynamic` instead to make this explicit. For the regression suite there is no difference, though. Do you think it's worth doing that? Is `deallocated_obj2t` ever used on non-allocated dynamic objects (if this notion even exists)?

fbrausse:
Master:
```
  correct:            8454
    correct true:     5127
    correct false:    3327
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7141
```
This PR:
```
  correct:            8456
    correct true:     5128
    correct false:    3328
  incorrect:            54
    incorrect true:     25
    incorrect false:    29
  unknown:            7138
```
Diff:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-787/termination/results.2022-07-05_09-53-19.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-787/unreach-call/results.2022-07-05_09-53-21.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-787/valid-memsafety/results.2022-07-05_09-53-30.diff.html)

"New" (timeout -> incorrect) is the [usual](https://github.com/esbmc/esbmc/pull/764#issuecomment-1166255017) suspect: `c/pthread-atomic/read_write_lock-2.i`

lucasccordeiro:
Many thanks for writing this PR, @fbrausse.

PR #788, opened 2022-07-04
[symex] plug free'd hole

### LINKED ISSUE #775: Memory safety: addresses of free'd objects are excluded from the address space

### LINKED ISSUE #785: Large object sizes get truncated to zero in SMT formula
Probably due to some missing `uint64_t` or `BigInt`, this functionally identical variation of https://github.com/esbmc/esbmc/issues/775#issuecomment-1172190902
```c
#include <stdlib.h>
#include <assert.h>

void steal_addr_space(size_t n)
{
        char *a = malloc(n);
        assert(__ESBMC_get_object_size(a) == n);
        free(a);
}

int main()
{
        size_t GiB = 1ULL << 30;
        for (int i=0; i<4; i++)
                steal_addr_space(1 * GiB);
        assert(0);
}
```
gives `VERIFICATION FAILED`. This is the correct result, but the wrong reason: the smt-formula encodes the object sizes as zero:
```
(assert (= (bvadd __ESBMC_ptr_obj_start_2 #x00000000) __ESBMC_ptr_obj_end_2))
(assert (= (bvadd __ESBMC_ptr_obj_start_3 #x00000000) __ESBMC_ptr_obj_end_3))
(assert (= (bvadd __ESBMC_ptr_obj_start_4 #x00000000) __ESBMC_ptr_obj_end_4))
(assert (= (bvadd __ESBMC_ptr_obj_start_5 #x00000000) __ESBMC_ptr_obj_end_5))
```
Same for 8 * 512 MiB. Starting at 16 * 256 MiB the sizes get encoded correctly:
```
(assert (= (bvadd __ESBMC_ptr_obj_start_16 #x10000000) __ESBMC_ptr_obj_end_16))
```

**Edit** this is with `--32 --force-malloc-success`.

fbrausse:
The address space layout constraints make the assumption that all (alive,
fixes #775) dynamic objects have disjoint addresses. This was ensured in
`smt_convt::finalize_pointer_chain()` before, but the validity of objects
could not be ensured there - when the slicer was active, see #781.
By assuming those conditions in symex, the validity of objects can be and
now is encoded.

This also fixes a wrong type for the `__ESBMC_alloc_size[obj]` expression.

I'm also [running](https://github.com/esbmc/esbmc/actions/runs/2609944134) this one over the benchmarks.


fbrausse:
I've also added a fix for #785.

fbrausse:
(Previous) Master (3eb0a4f154d49f9f67b825d950aed55244ff7c19 merge-base):
```
  correct:            8454
    correct true:     5127
    correct false:    3327
  incorrect:            53
    incorrect true:     24
    incorrect false:    29
  unknown:            7141
```
This PR (https://github.com/esbmc/esbmc/pull/788/commits/6710342f69a3762c12ea926484f2c06e749c4d52 just before the #785 fix):
```
  correct:            8378
    correct true:     5075
    correct false:    3303
  incorrect:            52
    incorrect true:     23
    incorrect false:    29
  unknown:            7218
  Score:             12253 (max: 25567)
```
Diff:
- [termination](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-788/termination/results.2022-07-05_14-55-34.diff.html)
- [unreach-call](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-788/unreach-call/results.2022-07-05_14-55-36.diff.html)
- [valid-memsafety](https://fbrausse.github.io/esbmc/svcomp/diff-pre-gh-781-gh-788/valid-memsafety/results.2022-07-05_14-55-44.diff.html)

We have 2x (timeout -> wrong) in unreach-call:
- `c/pthread-atomic/read_write_lock-2.i` [again](https://github.com/esbmc/esbmc/pull/764#issuecomment-1166255017) since it's the same merge-base for this run
- `c/aws-c-common/aws_hash_table_create_harness.i` (also wrong locally in merge-base, Debug build)

2x (wrong -> timeout) in unreach-call:
- `c/ldv-validator-v0.6/linux-stable-1dfa93a-1-100_1a-drivers--usb--serial--kobil_sct.ko-entry_point.cil.out.i` [wrong](https://github.com/esbmc/esbmc/pull/606#issuecomment-1019967089) already for some time
- `c/ldv-linux-3.14/linux-3.14_complex_emg_linux-kernel-locking-spinlock_drivers-net-ethernet-myricom-myri10ge-myri10ge.cil.i` was [wrong](https://github.com/esbmc/esbmc/pull/615#issuecomment-1027754299) already [before](https://github.com/esbmc/esbmc/pull/468#issuecomment-888566789) and locally on this PR/Debug is still wrong.

2x (correct -> unknown) in unreach-call:
- `c/aws-c-common/aws_byte_buf_reserve_harness.i`
- `c/aws-c-common/aws_byte_buf_reserve_relative_harness.i`
  
  Both fail with Debug builds of merge-base and this PR locally:
  ```
  esbmc/src/irep2/irep2_expr.cpp:383: void assert_type_compat_for_with(const type2tc&, const type2tc&): Assertion `a == b' failed.
  ```
  when storing the 64-bit BV symbol `c:aws_byte_buf_reserve_harness.i@137819@F@can_fail_realloc@newsize` in the array `__ESBMC_alloc_size` of 32-bit BVs. That can be fixed. The sizes should actually not be 32-bit when the address space is larger. Solving this correctly might require touching many parts of the code base.

and of course the 1x (wrong -> correct) from #775 in valid-memsafety.

Though nothing gets new incorrect verdicts, I have the feeling this PR is not using ESBMC's symex API correctly since also the regression suite runs slower now and even one timeout. It's understandable as now these constraints are local assumptions, not globally asserted facts anymore about the symbols smt_convt uses to track dynamically allocated objects.

Maybe using `assume()` in `goto_symext` is not the right way to add these global address space layout constraints? They affect the entire modelled heap. Is there a better way to achieve this? Any pointers as to how to supply these assumptions properly in symex are appreciated.

lucasccordeiro (commented):
Hello Franz,

We could investigate the two ideas:

1) Use an "address set analysis" to obtain an over-approximation of the address space layout constraints. For example, we could look at how we perform value set analysis in the pointer-analysis module.

2) Use a lightweight incremental SMT solving or the UNSAT core produced by the SMT solver to obtain information about the existing dynamic objects. 

What do you think?

fbrausse:
> What do you think?

These are interesting ideas for alternative approaches. I'll look into those.

In the meantime, any suggestions how to improve the approach in this PR are still very much welcome. It seems to be a cleaner approach than #781, which was making assumptions about the sequence of formulas given to smt_convt. Though even with this approach, there still are such assumptions in the smt_memspace: https://github.com/esbmc/esbmc/blob/f3d0b40e21ba08f32cfd2fea689b0585a16cd7aa/src/solvers/smt/smt_conv.cpp#L105-L110 used (only?) in https://github.com/esbmc/esbmc/blob/f3d0b40e21ba08f32cfd2fea689b0585a16cd7aa/src/solvers/smt/smt_memspace.cpp#L393-L403 Note that the corresponding symbol nowadays is called `c:@__ESBMC_is_dynamic`.

lucasccordeiro:
@fbrausse: can I ask you for the status of this PR?

fbrausse:
@lucasccordeiro To me it seems even less likely that this PR is the right approach as it is, see <https://github.com/esbmc/esbmc/issues/775#issuecomment-1179712955> for details. The smt-based #781 did have less performance impact and could be made to work with the slicer by e.g. #667 and/or with an extra expression type.

lucasccordeiro:
@fbrausse: shall we close this PR? From your comments, it seems this PR is not the way to move forward.

fbrausse:
Closed, since we merged the #781 solution in #905.

PR #789, opened 2022-07-05
[svcomp] update stats

fbrausse:
@lucasccordeiro was too fast for me :)

This is the stats update for #787.


lucasccordeiro:
:-)

PR #791, opened 2022-07-06
Feature/add solidity string type

correcthorsebatterystaple:
Changes:
- Add string literal to the grammar set for solidity
- Add support for both string memory and string storage declarations and assignments

Note: Solidity doesn't have comparison of strings as a language feature, this is often implemented by leveraging the hash functions which will be implemented in a future PR. Lack of comparison prevents us from forming assertions on the string which prevents us from getting and failed verification instances in the tests.


src/solidity-frontend/solidity_convert_literals.cpp
```diff
@@ -60,6 +60,24 @@ bool solidity_convertert::convert_bool_literal(
 
 // TODO: Character literal
 // TODO: String literal
```
rafaelsamenezes: Is this old comment still valid?

rafaelsamenezes (changes requested):

src/solidity-frontend/solidity_convert_literals.cpp
correcthorsebatterystaple: Nope, will change

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1596,6 +1605,18 @@ bool solidity_convertert::get_elementary_type_name(
     new_type.set("#cpp_type", c_type);
     break;
   }
+  case SolidityGrammar::ElementaryTypeNameT::STRING:
+  {
+    size_t value_length = 128;
```
kunjsong01, line 1610: Add some comments to explain why 128 is used?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1596,6 +1605,18 @@ bool solidity_convertert::get_elementary_type_name(
...
+
+    new_type = array_typet(
```
kunjsong01, line 1612: Add some comments to explain why array type is used? 
Are we converting "string" and "string memory" to  array type?

src/solidity-frontend/solidity_convert.cpp
kunjsong01, line 1610: Add some comments to explain why 128 is used?

src/solidity-frontend/solidity_convert.cpp
kunjsong01, line 1612: Add some comments to explain why array type is used?
Are we converting "string" and "string memory" to array type?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -846,6 +846,12 @@ bool solidity_convertert::get_expr(
         return true;
       break;
     }
+    case SolidityGrammar::ElementaryTypeNameT::STRING_LITERAL:
+    {
+      if(convert_string_literal(literal, the_value, new_expr))
```
kunjsong01: Add a comment to explain this line? Perhaps, "converting the RHS of `string s = "HELLO"`"?

kunjsong01 (commented):
LGTM, just some minor comments

kunjsong01:
@correcthorsebatterystaple A question on converting the initialization statement `string s = "HELLO";`

In function `bool solidity_convertert::get_var_decl`: 

https://github.com/esbmc/esbmc/blob/f3d0b40e21ba08f32cfd2fea689b0585a16cd7aa/src/solidity-frontend/solidity_convert.cpp#L282-L293

It seems that this part is only dealing with int literal for the time being. 
Should this be extended to handle string literal?

src/solidity-frontend/solidity_convert_literals.cpp
```diff
@@ -60,6 +60,24 @@ bool solidity_convertert::convert_bool_literal(
...
+bool solidity_convertert::convert_string_literal(
+  const nlohmann::json &string_literal,
```
kunjsong01: This parameter can be removed

src/solidity-frontend/solidity_convert_literals.cpp
```diff
@@ -60,6 +60,24 @@ bool solidity_convertert::convert_bool_literal(
...
+{
+  size_t string_size = the_value.size();
+  typet type = array_typet(
+    signed_char_type(),
+    constant_exprt(
+      integer2binary(string_size, bv_width(int_type())),
+      integer2string(string_size),
+      int_type()));
+
+  string_constantt string(the_value, type);
```
fbrausse, line 82: Though this ESBMC type has the right name for the usage here, it actually represents a (somewhat broken¹) C-style string literal, that is, it represents the given value with a trailing `\0'. This should probably be fixed. Is Solidity OK with having size+1?

¹ broken because it gets the size in bytes from clang_c_frontend as well as the Solidity one here, and adds +1 to it. For wide-character strings that results in a "size" of the string_constantt of e.g. 21=5*4+1 for the C string literal `L"hello"` if `sizeof(wchar_t) == 4`.

src/solidity-frontend/solidity_grammar.h
```diff
@@ -123,6 +123,8 @@ enum ElementaryTypeNameT
   // TODO: rule address
   // TODO: rule address payable
   // TODO: rule string
```
fbrausse: Is this comment, too, still valid?

src/solidity-frontend/solidity_convert_literals.cpp
correcthorsebatterystaple: What would you suggest as a fix for this?

> Is Solidity OK with having size+1?

I do not understand the question

fbrausse: You're giving `the_value` to the string_constantt. This will (down the line) be interpreted as the C-style literal `the_value`, including a `\0` at the end. My question: Do string literals in Solidity have a trailing `\0`? If not, does it matter for Solidity? For instance, is it possible to get a specific character from a string by index `str[n]`?

If string literals are supposed to be represented the same way as in C or strings can't be indexed nor their size taken, it's fine. If not, we might need to change `string_constantt`.

correcthorsebatterystaple: In solidity strings themselves are not able to be accessed via index, and they need to be cast to a bytes type before their length can be accessed

fbrausse: OK, thanks. So when we get to this string -> byte conversion the additional `\0` byte needs to be observed. Could you maybe add a TODO comment here?

lucasccordeiro:
@correcthorsebatterystaple: can I ask you whether you had a chance to address the reviewers comments?

correcthorsebatterystaple:
> @correcthorsebatterystaple: can I ask you whether you had a chance to address the reviewers comments?

I am adding the edits as I write this, rebasing building and testing is taking a little while as I had some booting issues with my machine.

Please expect changes to come through soon.

src/solidity-frontend/solidity_convert.cpp
correcthorsebatterystaple: I believe the method name is self explanatory when it comes to describing the purpose of this statement. I should, however, add doc comments to the method itself to better explain what it does

correcthorsebatterystaple:
> @correcthorsebatterystaple A question on converting the initialization statement `string s = "HELLO";`
> 
> In function `bool solidity_convertert::get_var_decl`:
> 
> https://github.com/esbmc/esbmc/blob/f3d0b40e21ba08f32cfd2fea689b0585a16cd7aa/src/solidity-frontend/solidity_convert.cpp#L282-L293
> 
> It seems that this part is only dealing with int literal for the time being. Should this be extended to handle string literal?

string literals being of only one type (unlike int having different bitwidths) we do not have to worry about the literal type of the string. Furthermore, I have written that bit of int literal code with much less care than I should have, this will update in the fute and do expect a refactor to clean up that structure a bit.

src/solidity-frontend/solidity_convert.cpp
correcthorsebatterystaple: This here is a TODO as in actuality it should be the size of the string literals but we have not passed that information here

src/solidity-frontend/solidity_convert.cpp
correcthorsebatterystaple: This is in following with how string literals are handled by the clang c front end. Yes both string and string memory type both have array types following the convention from clang c front end

src/solidity-frontend/AST/solidity_ast.h
```diff
@@ -0,0 +1,18 @@
+#include <util/message/default_message.h>
+
+// For json parsing
+using json = nlohmann::json;
+
+class solidity_ast
```
fbrausse: This class and in fact this new header file seems to be unused and unfinished. I would suggest removing it. If you want to keep it, please
- remove the global `using json = [...]` to reduce namespace pollution
- consider not dumping via `msg.debug` but `msg.status`, this makes debugging easier
- finish it (e.g. by adding a `to_string()` method) so it compiles when included somewhere

fbrausse (changes requested):

src/solidity-frontend/AST/solidity_ast.h
correcthorsebatterystaple: This has been committed by mistake, I will remove

fbrausse (approved):
LGTM.

lucasccordeiro (approved):
@correcthorsebatterystaple: could you please rebase this PR with the master branch?

src/solidity-frontend/solidity_convert.cpp
fbrausse: Well, it's impossible to tell whether "128" is a TODO or not ;). Please be explicit.

kunjsong01 (approved):
LGTM

PR #797, opened 2022-07-07
Clang-14 support

fbrausse:
In #790 @rafaelsamenezes noted that Clang-14 builds are broken because of an API change.

This PR fixes it by supporting the new and the old name, depending on the major version number (14). As discussed in the issue, this PR also clarifies "minimum" and "recommended" versions for building ESBMC manually in the build instructions.


rafaelsamenezes (approved):
LGTM

lucasccordeiro (approved):

PR #803, opened 2022-07-09
Deduplicate namespacet API and separate "merged context" functionality from it

fbrausse:
This PR
- deduplicates the `namespacet` API functions `bool lookup(name, sym)`, `const symbolt & lookup(name)` into a single `const symbolt * lookup(name)`,
- removes unused constructors and wrapper functions from the ansi-c and cpp frontends,
- moves the unrelated "merge two different contexts into one namespace" functionality from `namespacet` to the now only user `c_linkt`, and
- unifies both locations `migrate_namespace_lookup` were initialized at since 7e8a77fe6928696f5c6a02bdd9a3b42d2e264c42 into one, by doing it earlier in `get_goto_program()`.

The aim here is making it clear that namespaces like `migrate_namespace_lookup` do only talk about a single context as c_linkt's merged one is only used internally. In this case, it is `esbmc_parseoptiont`'s context. As this class is also responsible for initializing the `config`, it is now conceivable to put `migrate_namespace_lookup` into `config` and initialize it at the same time in `configt::set()`. This PR is not doing that, yet. This may need consensus as it affects globals, cf #755.


fbrausse:
As this PR moves the initialization up to before parsing, #588 does not segfault anymore but instead aborts with
```
src/util/migrate.cpp:511: expr2tc sym_name_to_symbol(irep_idt, type2tc): Assertion `endatptr != atstr.c_str()' failed.
```
since the `migrate_namespace_lookup` is only being populated after parsing.

mikhailramalho (approved):
LGTM.

lucasccordeiro (approved):

PR #804, opened 2022-07-11
[clang-cpp-frontend] Fix references

kunjsong01:
replaces https://github.com/esbmc/esbmc/pull/753

Replicating the old frontend's conversion flow for: 
```cpp
     1	#include <cassert>
     2
     3	int main()
     4	{
     5	  int g;
     6	  int &r = g;
     7	  r = 1;
     8	  assert(r == 1); // pass
     9	  assert(g == 0); // fail
    10	}
```


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -982,9 +987,21 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
...
       static_cast<const clang::LValueReferenceType &>(the_type);
 
-    if(get_type(lvrt.getPointeeTypeAsWritten(), new_type))
+    typet sub_type;
+    if(get_type(lvrt.getPointeeTypeAsWritten(), sub_type))
       return true;
 
+    // Special case, pointers to structs/unions/classes must not
+    // have a copy of it, but a reference to the type
+    // TODO: classes
```
mikhailramalho: Is this TODO still valid?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -995,3 +995,19 @@ bool clang_cpp_convertert::get_decl_ref(
...
+
+void clang_cpp_convertert::get_default_symbol(
+  symbolt &symbol,
+  irep_idt module_name,
+  typet type,
+  irep_idt name,
+  irep_idt id,
+  locationt location)
+{
+  symbol.mode = "C++";
```
mikhailramalho: can you make a new constexpr member called `mode` that initializes to the proper string when creating the frontend?

mikhailramalho (approved):
Minor changes are required. Also, remember to remove the debug messages.

Can you also add test cases for when function arguments are references? Or is it going to be part of another PR?

lucasccordeiro (approved):

kunjsong01:
Now we are getting the symbol table same as the old frontend: 
```yaml
Symbol......: c:@F@main#
Module......: ref
Base name...: main
Mode........: C++ (1)
Type........: int ()
Value.......: 
{
int g;

int & r=&g; # ---> turned into address_of

*r = 1; # ---> turned into dereference_of

(int)(*r) == 1 ? 0 : __assert_fail(&"r == 1"[0], &"ref.cpp"[0], 8, &"int main()"[0]); # ---> turned into dereference_of

g == 0 ? 0 : __assert_fail(&"g == 0"[0], &"ref.cpp"[0], 9, &"int main()"[0]);

}
```

Verification of the example in this PR: 
```yaml
Counterexample:

State 1 file ref.cpp line 6 function main thread 0
  r = &g

State 2 file ref.cpp line 7 function main thread 0
  *r = 1 (00000000 00000000 00000000 00000001)

State 3 file ref.cpp line 9 function main thread 0
Violated property:
  file ref.cpp line 9 function main
  assertion g == 0
  g == 0


VERIFICATION FAILED
```

kunjsong01:
@mikhailramalho 
>Also, remember to remove the debug messages.

ACK

>Can you also add test cases for when function arguments are references? Or is it going to be part of another PR?

This is going to be in another PR. This PR only fixes the problem of the test case in https://github.com/esbmc/esbmc/pull/804#issue-1300783337.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Done in commit d1c213c01f64dcb3e961004d2b089f6b9d4b32d8

kunjsong01:
requesting a re-review as a few more commits were made to adjust side_effect for reference symbol.

lucasccordeiro (approved):

kunjsong01:
### Dev notes: 

Need to be careful with the following operations: 
```cpp
operands().resize(1);
copy_to_operands(expr);
```

It does NOT reserve op0 and copy expr to op0. 

The resultant tree will have one more empty operand: 
```yaml
* operands:
  0:
    * type:  
  1: expr
     * type: ...
     * operands:...
     * location: ...
     * statement: ...
```

This also breaks the old cpp frontend: 
```
### ERROR:
expression statement takes one operand
```

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -150,7 +150,7 @@ class clang_c_convertert
 
   bool get_cast_expr(const clang::CastExpr &cast, exprt &new_expr);
 
-  void get_default_symbol(
+  virtual void get_default_symbol(
```
mikhailramalho: This doesn't have to be virtual anymore.

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -206,6 +206,8 @@ class clang_c_convertert
   const clang::Decl *get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt);
 
   void gen_typecast_to_union(exprt &dest, const typet &type);
+
+  static constexpr const char *mode = "C";
```
mikhailramalho: Can we make it a protected class member? That way we can initialize it with "C++" in the cpp frontend and get rid of the `get_default_symbol` overload.

mikhailramalho (approved):
LGTM.

src/clang-c-frontend/clang_c_convert.h
fbrausse: I believe this string constant has something to do with `language_desc(language_idt::C)->name`. https://github.com/esbmc/esbmc/blob/d0a50af8ed45a9aa324f7ffb768e61caf7a16289/src/langapi/mode.cpp#L24

kunjsong01: @mikhailramalho done in commit ec3d68fa7e9dc333dfa986a438f230d5d07c692b

@fbrausse do you mean getting the corresponding `language_desct` and using `language_desc_*.name` in each frontend?

mikhailramalho: @kunjsong01 there is no need to pass the string as an argument to the constructor, simply initialize it in the initializer list.

kunjsong01: @mikhailramalho How do we initialize clang_c_converter's `mode`  in clang_cpp_converter?  
or do we want a `mode` member in clang_c_converter and clang_cpp_converter respectively?

mikhailramalho: I meant something like:
```
index ed7ea6e6c..72873227e 100644
@@ -28,11 +28,13 @@
 
 clang_c_convertert::clang_c_convertert(
   contextt &_context,
-  std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs)
+  std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs,
+  const char *_mode)
   : ASTContext(nullptr),
     context(_context),
     ns(context),
     ASTs(_ASTs),
+    mode(_mode),
     anon_symbol("clang_c_convertert::"),
     current_scope_var_num(1),
     current_block(nullptr),
@@ -41,6 +43,13 @@ clang_c_convertert::clang_c_convertert(
 {
 }
 
+clang_c_convertert::clang_c_convertert(
+  contextt &_context,
+  std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs)
+  : clang_c_convertert(_context, _ASTs, "C")
+{
+}
+
 bool clang_c_convertert::convert()
 {
   if(convert_top_level_decl())
diff --git a/src/clang-c-frontend/clang_c_convert.h b/src/clang-c-frontend/clang_c_convert.h
index c0b7e20de..6bc94a16f 100644
--- a/src/clang-c-frontend/clang_c_convert.h
+++ b/src/clang-c-frontend/clang_c_convert.h
@@ -65,10 +65,16 @@ public:
   static std::string get_decl_name(const clang::NamedDecl &nd);
 
 protected:
+  clang_c_convertert(
+    contextt &_context,
+    std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs,
+    const char *_mode);
+
   clang::ASTContext *ASTContext;
   contextt &context;
   namespacet ns;
   std::vector<std::unique_ptr<clang::ASTUnit>> &ASTs;
+  const char *mode;
 
   symbol_generator anon_symbol;
 
index 079f62881..b3d23ff82 100644
@@ -26,7 +26,7 @@
 clang_cpp_convertert::clang_cpp_convertert(
   contextt &_context,
   std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs)
-  : clang_c_convertert(_context, _ASTs)
+  : clang_c_convertert(_context, _ASTs, "C++")
 {
 }
```

I'm not sure I like it better tho. @fbrausse what do you think?

mikhailramalho: Btw, I already think the PR is good to merge, this is just cosmetics.

fbrausse: @mikhailramalho's version is easier to understand I think: mere users of clang_c_convertt should not need to specify the mode of that converter.

> @fbrausse do you mean getting the corresponding language_desct and using language_desc_*.name in each frontend?

Yeah, but:

>  this is just cosmetics.

Agreed.

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -72,6 +73,7 @@ class clang_c_convertert
   namespacet ns;
   std::vector<std::unique_ptr<clang::ASTUnit>> &ASTs;
   const messaget &msg;
+  const char *mode;
```
fbrausse, line 76: Could you add a comment, e.g., "Default value for `symbolt::mode` specifying the definition's source language"? And if this member could be of type `irep_idt`, we'd save repeated lookups in the string_containert.

src/util/std_expr.h
```diff
@@ -812,6 +812,12 @@ class dereference_exprt : public exprt
     operands().resize(1);
   }
 
+  explicit dereference_exprt(const exprt &op, const typet &tp)
```
fbrausse, line 815: I don't like having this constructor next to the other one taking only a `typett`. Because they take different `typett`s. This one uses the tp's subtype, the other one does not. This needs at least a comment to warn about the difference, but ideally, this constructor should use its type parameter as the other one does to avoid confusion.

fbrausse (changes requested):
Looks good, though I would really change the behaviour of the new constructor.

PR #808, opened 2022-07-13
[clang-c*-frontend] Minor fixes

kunjsong01:
some minor fixes.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -30,7 +30,7 @@ clang_c_convertert::clang_c_convertert(
   contextt &_context,
   std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs,
   const messaget &msg,
-  const char *_mode)
+  irep_idt _mode)
   : ASTContext(nullptr),
     context(_context),
     ns(context),
--- a/src/clang-c-frontend/clang_c_convert.h
+++ b/src/clang-c-frontend/clang_c_convert.h
@@ -47,7 +47,7 @@ public:
     contextt &_context,
     std::vector<std::unique_ptr<clang::ASTUnit>> &_ASTs,
     const messaget &msg,
-    const char *_mode);
+    irep_idt _mode);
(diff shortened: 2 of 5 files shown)


lucasccordeiro (approved):

fbrausse (commented):
`irep_idt` parameters don't need to be `const` (or const-ref): it is just a fancy wrapper around an integer.

kunjsong01:
> irep_idt parameters don't need to be const

@fbrausse Why does it not need to be const? 

Is it because of:
1. `irep_idt` is `dstring`
2. the unsigned integer `dstring.no` is a private number
3. there is no public setter method to set `dstring.no`

So `dstring.no` is inherently immutable?

That's just my guess. I tend to write the code in a more defensive way, sometimes could be unnecessary, of course.

fbrausse:
> Is it because of: [123]

Actually nothing that involved. The previous version had a non-const parameter `const char *mode`. `mode` itself is a non-const variable passed by-value. And that's fine, since it's just a pointer. Same now: `irep_idt` is just an integer that can be passed by-value. It's being copied into the `::mode` member (or passed to follow-up ctors by-value) anyway, there is no need to qualify it in any way, it's just visual noise.

fbrausse:
I'm talking about the parameters to the constructors, the member `::mode` itself can be `const`. Sorry for the confusion.

kunjsong01:
>Same now: irep_idt is just an integer that can be passed by-value. It's being copied into the ::mode member (or passed to follow-up ctors by-value) anyway,

I see. Thanks for the answer. 
Removed the const qualifier.

fbrausse (approved):
Great, many thanks for addressing my comments!

kunjsong01:
> Great, many thanks for addressing my comments!

np thanks for the reviewing

kunjsong01:
CI: 
Windows build failed with step "install gnu tools"

fbrausse:
@kunjsong01 might be temporary; the CI on another PR seems to be able to do this. After all tasks here are finished, you can (probably?) restart the failed tasks here. "Details" -> "Re-run jobs" (upper right).

PR #811, opened 2022-07-14
[cpp] Fixed zombie operand

kunjsong01:
Fixed zombie operand in cpp frontend while following the conversion flow for new/delete keywords. Previously mentioned in https://github.com/esbmc/esbmc/pull/804#issuecomment-1181750529. 

### Test Case
```cpp
     1	#include <cassert>
     2	class t2
     3	{
     4	public:
     5	  int i;
     6
     7	  t2() : i(2)
     8	  {
     9	  }
    10	};
    11
    12	int main()
    13	{
    14	  t2 *p = new t2;
    15	  assert(p->i == 2);
    16	  delete p;
    17	}
```

### irep tree BEFORE the fix: 
```yaml
code
  * type: code
      * arguments:
  * operands:
    0:        # ---> zombie operand
      * type:
    1: sideeffect
        * type: signedbv
            * width: 32
```
### irep tree AFTER the fix:
```yaml
code
  * type: code
  * operands:
    0: sideeffect
        * type: signedbv
            * width: 32
``` 

### error message BEFORE the fix: 
```
expression statement takes one operand
```

### error message AFTER the fix: 
```
Generating GOTO Program
cannot remove side effect (cpp_delete)
terminate called after throwing an instance of 'int'
```
This is another issue for next PR.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -562,11 +562,6 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
   std::string id, name;
   get_decl_name(fd, name, id);
 
-  if (id == "c:@F@main#")
-  {
-    printf("Got main\n");
-  }
-
   symbolt symbol;
   get_default_symbol(
     symbol,
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -562,6 +562,11 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
   std::string id, name;
   get_decl_name(fd, name, id);
 
+  if (id == "c:@F@main#")
+  {
+    printf("Got main\n");
+  }
+
   symbolt symbol;
   get_default_symbol(
     symbol,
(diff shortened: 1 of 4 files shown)


kunjsong01:
N.B. 
I don't want to change the behaviour of `move_to_operand` or `code_expressiont` ctor as they are being used in other places. 

We will probably see the same error signature (zombie operand) again in the old frontend. We can fix them on the fly. 
If all of them are caused by `code_expressiont` ctor then we can fix it there. Anyway, the current patch is good for the time being.

rafaelsamenezes (approved):

fbrausse (approved):
> I don't want to change the behaviour of move_to_operand or code_expressiont ctor as they are being used in other places.
>
> We will probably see the same error signature (zombie operand) again in the old frontend. We can fix them on the fly.
If all of them are caused by code_expressiont ctor then we can fix it there. Anyway, the current patch is good for the time being.

A proper fix would be to first get rid of the zero-arg ctor and then reintroduce it in a sane way, if needed.

As the only other zero-arg user outside of the old frontends is https://github.com/esbmc/esbmc/blob/8a42c753e0c1ae0ef720a8dfca5089ff15c37f37/src/goto-programs/builtin_functions.cpp#L362 this PR looks good to me.

lucasccordeiro (approved):

kunjsong01:
>A proper fix would be to first get rid of the zero-arg ctor and then reintroduce it in a sane way, if needed.
As the only other zero-arg user outside of the old frontends is 

I concur. This is a quick fix to push forward. I'll keep note of the `code_expressiont` ctor callsite.

Issue #800, opened 2022-07-08
Config revamp


rafaelsamenezes:
Following the discussion from #755

The main idea is:

1. Config will still be global
2. Configt will hold: message state, command-line and architecture.
3. Add support to a configuration file (#477)
4. Options should be removed from config. Having global options would break k-parallel (@mikhailramalho can you confirm this?)

PR #814, opened 2022-07-14
Support not slicing particular symbols

fbrausse:
This PR is an alternative to #667. It implements @rafaelsamenezes idea to suppress slicing of SSA steps when the symbol is unused. Two methods of specifying these symbols are included:
1. via annotation `__attribute__((annotate("__ESBMC_no_slice")))`
2. via one of the new cmdline options `--no-slice-name <name>` and `--no-slice-id <id>`, where the latter allows more fine-grained control of the explicit renumbered symbol and the former applies to all instances of that symbol

I don't want to steal any ideas here, it just presents a different approach that also turned out to be helpful with the slicing problems in #781. Additionally, this PR includes a simplification of the slicer: the generality of `get_symbol(symb, fn)` is not needed since this internal function is only used with 2 different callbacks `fn` ever. Represent that by a `template <bool Add>`.

### DIFF
--- a/regression/esbmc/github_666/test.desc
+++ b/regression/esbmc/github_666/test.desc
@@ -1,0 +1,4 @@
+CORE
+no_slice.c
+--json-options-input config.json
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_666/test.desc
+++ b/regression/esbmc/github_666/test.desc
@@ -1,4 +1,4 @@
 CORE
 no_slice.c
---json-options-input config.json
+
 ^VERIFICATION FAILED$
--- a/regression/esbmc/github_666/test.desc
+++ b/regression/esbmc/github_666/test.desc
@@ -1,4 +1,6 @@
 CORE
 no_slice.c
 
+\bfoo = 42\b
+\bfoo = 15\b
 ^VERIFICATION FAILED$
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -447,23 +447,30 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
     return true;
 
   // Check if we annotated it to be have an infinity size
+  bool no_slice = false;
   if(vd.hasAttrs())
   {
     for(auto const &attr : vd.getAttrs())
     {
       if(const auto *a = llvm::dyn_cast<clang::AnnotateAttr>(attr))
       {
-        if(a->getAnnotation().str() == "__ESBMC_inf_size")
+        const std::string &name = a->getAnnotation().str();
+        if(name == "__ESBMC_inf_size")
         {
           assert(t.is_array());
           t.size(exprt("infinity", uint_type()));
         }
+        else if(name == "__ESBMC_no_slice")
+          no_slice = true;
       }
     }
   }
 
   std::string id, name;
   get_decl_name(vd, name, id);
+
+  if(no_slice)
+    config.no_slice_names.emplace(id);
 
   locationt location_begin;
   get_location_from_decl(vd, location_begin);
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -570,10 +570,28 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
   {
     fine_timet slice_start = current_time();
     BigInt ignored;
-    if(!options.get_bool_option("no-slice"))
-      ignored = slice(eq, options.get_bool_option("slice-assumes"));
-    else
+    if(options.get_bool_option("no-slice"))
       ignored = simple_slice(eq);
+    else
+    {
+      struct
+      {
+        const std::unordered_set<std::string> &names, &ids;
+
+        bool operator()(const symbol2t &sym) const
+        {
+          return names.count(sym.thename.as_string()) ||
+                 ids.count(sym.get_symbol_name());
+        }
+      } no_slice_symbol = {
+        config.no_slice_names,
+        config.no_slice_ids,
+      };
+
+      ignored =
+        slice(eq, options.get_bool_option("slice-assumes"), no_slice_symbol);
+    }
+
     fine_timet slice_stop = current_time();
 
     log_status(
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -569,28 +569,7 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
   try
   {
     fine_timet slice_start = current_time();
-    BigInt ignored;
-    if(options.get_bool_option("no-slice"))
-      ignored = simple_slice(eq);
-    else
-    {
-      struct
-      {
-        const std::unordered_set<std::string> &names, &ids;
-
-        bool operator()(const symbol2t &sym) const
-        {
-          return names.count(sym.thename.as_string()) ||
-                 ids.count(sym.get_symbol_name());
-        }
-      } no_slice_symbol = {
-        config.no_slice_names,
-        config.no_slice_ids,
-      };
-
-      ignored =
-        slice(eq, options.get_bool_option("slice-assumes"), no_slice_symbol);
-    }
+    BigInt ignored = slice(eq);
 
     fine_timet slice_stop = current_time();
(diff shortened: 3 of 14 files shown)


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -450,24 +450,31 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
...
+        else if(name == "__ESBMC_no_slice")
+          no_slice = true;
       }
     }
   }
 
   std::string id, name;
   get_decl_name(vd, name, id);
 
+  if(no_slice)
```
mikhailramalho, line 472: I prefer to add a flag to `symbolt` and set it here. That way we can see it when dumping the context/symbol table.

src/esbmc/bmc.cpp
```diff
@@ -606,10 +606,28 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
+          return names.count(sym.thename.as_string()) ||
+                 ids.count(sym.get_symbol_name());
+        }
+      } no_slice_symbol = {
+        config.no_slice_names,
+        config.no_slice_ids,
+      };
+
+      ignored =
+        slice(eq, options.get_bool_option("slice-assumes"), no_slice_symbol);
```
mikhailramalho: Can you pass the options object and get both the `slice-assumes` flag and the list inside the `symex_slicet` constructor?

src/util/config.cpp
```diff
@@ -123,6 +123,18 @@ bool configt::set(const cmdlinet &cmdline, const messaget &msg)
     return true;
   }
 
+  if(cmdline.isset("no-slice-name"))
```
mikhailramalho, line 118: If we save the list to the options object in `get_command_line_options` we don't need this here.

mikhailramalho (changes requested):
The idea looks alright, I would just move stuff around a bit.

src/util/config.cpp
fbrausse: I see, there seem to be 2 sites where cmdline is being processed. Unfortunately `optionst` does not support lists of strings (nor unordered_sets) and I didn't want to add this functionality to it since @rafaelsamenezes wants to remove `optionst` (which mostly seems to be a glorified wrapper around the cmdline) in #800.

lucasccordeiro (changes requested):
@fbrausse: Shall we add to this PR the test cases from https://github.com/esbmc/esbmc/pull/781?

src/clang-c-frontend/clang_c_convert.cpp
rafaelsamenezes: I think that using it directly at config seems easier and more practical. Also, we can check if a symbol id is in the `no_slice` if we want to print it.

src/esbmc/bmc.cpp
rafaelsamenezes: Or move this block into a separate function. Or to inside `config.h`.

src/util/config.cpp
rafaelsamenezes: > If we save the list to the options object 

I think that would break `k-parallel` though. The current mode (e.g., `inductive-step`) is set as an option.

> Unfortunately optionst does not support lists of strings (nor unordered_sets) and I didn't want to add this functionality to it since @rafaelsamenezes wants to remove optionst (which mostly seems to be a glorified wrapper around the cmdline) in https://github.com/esbmc/esbmc/issues/800.

Yes, they both have lots of things in common and/or some limitations.

regression/esbmc/github_666/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+no_slice.c
+
+^VERIFICATION FAILED$
```
rafaelsamenezes, line 6: Since this focus on no-slice, could you add some regex to show that the symbol is shown at the counter example?

rafaelsamenezes (changes requested):
I would like more regression tests.

Also it would be great to extend this to support symbol prefixes in another PR. Slicing the nondet gives us some trouble at test/sv COMP.

regression/esbmc/github_666/test.desc
fbrausse: Good catch, done. Would be a bit useless without.

fbrausse:
@lucasccordeiro This PR does not fix the address space constraints and those test cases would still fail. It only enables us to put the no-slice attribute on `__ESBMC_alloc` so that this symbol won't be affected by the slicer and available to smt_conv. That was the main technical problem of #781. I've added Rafael's example from #667 in a slightly modified form (no config.json).

> Also it would be great to extend this to support symbol prefixes in another PR. Slicing the nondet gives us some trouble at test/sv COMP.

@rafaelsamenezes Regex cmdline params?

rafaelsamenezes:
> @rafaelsamenezes Regex cmdline params?

I am not sure if we need full regex. Just a `has_prefix` should be enough

src/clang-c-frontend/clang_c_convert.cpp
fbrausse: My main motivation was to support a fully renumbered symbol-ids as #667 did. This would be tricky to do purely via flags on symbols.

src/esbmc/bmc.cpp
fbrausse: Well, in that case I could get the `"no-slice"` option as well to distinguish between slice() and simple_slice() and don't need the `no_slice` callback as a member. I've added a commit that does this and also moved the helper class `symex_slicet` into an anon namespace in the .cpp file.

src/util/config.cpp
mikhailramalho: Wait, are we removing optionst in favor of config? Wouldn't be easier to remove cmdline, since it's used in fewer places?

rafaelsamenezes: > Wait, are we removing optionst in favor of config?

No, I was wondering if a global options would affect k-parallel. The plan is to:

1. Make cmdline only for parsing
2. Extend options to be able to handle multiple values per option. That is the main limitation right now as we cannot get all the inputs.
3. Make config hold the options

lucasccordeiro:
@fbrausse: can I ask you for the status of this PR?

fbrausse:
As I understand, I should add some regression tests and @mikhailramalho (a) prefers a different approach where the "no-slice" information is not stored in the config, but attached to each symbol, and (b) would like the user-specified "no-slice" info to sit in options instead of config.

There are reasons not doing [(a)](https://github.com/esbmc/esbmc/pull/814#discussion_r922004300) (in general) or [(b)](https://github.com/esbmc/esbmc/pull/814#discussion_r921323557) (in this PR), as explained above. @mikhailramalho what do you think?

mikhailramalho:
> As I understand, I should add some regression tests and @mikhailramalho (a) prefers a different approach where the "no-slice" information is not stored in the config, but attached to each symbol, and (b) would like the user-specified "no-slice" info to sit in options instead of config.
> 
> There are reasons not doing [(a)](https://github.com/esbmc/esbmc/pull/814#discussion_r922004300) (in general) or [(b)](https://github.com/esbmc/esbmc/pull/814#discussion_r921323557) (in this PR), as explained above. @mikhailramalho what do you think?

I see your point, I agree with the changes.

mikhailramalho (approved):
LGTM.

fbrausse:
Thanks! I'll look into providing some more regression tests.

rafaelsamenezes (approved):

src/goto-symex/slice.cpp
```diff
@@ -170,3 +193,15 @@ BigInt simple_slice(std::shared_ptr<symex_target_equationt> &eq)
 
   return ignored;
 }
+
+BigInt slice(std::shared_ptr<symex_target_equationt> &eq)
```
lucasccordeiro, line 290: Add some comments here.

lucasccordeiro (approved):
I have just one minor comment.

src/goto-symex/slice.cpp
fbrausse: I've put the comments in the header file, OK?

fbrausse:
Btw. we should not forget the nice commenting work @rafaelsamenezes did in #667.

lucasccordeiro:
@fbrausse: is this PR ready to merge?

lucasccordeiro (approved):

PR #816, opened 2022-07-15
[clang-c-converter] fix initialization of static storage compound literals

fbrausse:
Earlier, the initialization order for those comp-lits was wrong, e.g.
```c
  int vsi = (int){42};
```
in file scope got translated to
```
  vsi=main.c:7$compound-literal$1;
  main.c:7$compound-literal$1=42;
```
The reason was the unintuitive but necessary behaviour of `get_var()` which
takes care of variable declarations including the initializer:
```c
  int x = x;
```
is allowed, so the RHS (in general, but not in the static case!) needs to be
able to access the symbol on the LHS.

This fix is by special-casing the static case, as the RHS there has to be a
constant expression and thus cannot refer to other symbols (with the exception
of compound literals). So first translate the RHS (pretending there is no
enclosing scope to accomodate
```c
  void f() {
    static int x = (int){42};
  }
```
by temporarily clearing the `current_block` until the static decl is finished)
and then create the symbol for the LHS.


mikhailramalho (commented):
Can we move this code to clang_c_adjust? clang_c_convert should only convert the ast.

fbrausse:
@mikhailramalho The issue fixed here has to do with the order new symbols are put into the context, since that order is used by `clang_c_main::static_lifetime_init()` for the initialization. AFAICT clang_c_convert (**edit** or clang_c_adjust) does not reorder symbol declarations in the context. Is that even possible? How would you suggest to change this?

rafaelsamenezes:
> Can we move this code to clang_c_adjust? clang_c_convert should only convert the ast.

This is a corner case related to global variable initialization (similar to the `int x= x + 1`). I guess it is easier and more maintanable to do it directly.

rafaelsamenezes:
@fbrausse could you add some test cases?

rafaelsamenezes (approved):

lucasccordeiro (approved):
LGTM.

fbrausse:
@kunjsong01 I believe this will need to be treated differently in C++ context than done here for C, so it would be nice if you could take a look.

In C++ the initializer could be a function call (lambdas, other function objects) and therefore have side effects. I'm not sure how either of the C++ frontends handles this; we may want to revisit it at some point.

**Edit:** There is a nice CppCon 2021 talk by Daisy Hollman where she dives into the intricacies (aka hell) of static local initialization in C++: https://youtu.be/15etE6WcvBY from 17:21 to 25:22.

PR #827, opened 2022-07-27
SSA caching

rafaelsamenezes:
This PR is a basis for the full caching framework. This adds:

1. SSA algorithms, which work similarly to the goto algorithms. This will help develop "passes" over SSA
2. A basic CRC caching. Checks if a guard/cond pair was already seen and replace the assertion with a trivial case. I am not sure if it is better to replace the `cond` with 1 or the `guard` with 0.

### DIFF
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -46,6 +46,15 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
       algorithms.emplace_back(std::make_unique<simple_slice>());
     else
       algorithms.emplace_back(std::make_unique<symex_slicet>(options));
+
+    // Run cache if user has specified the options
+    if(options.get_bool_option("cache-asserts"))
+    {
+      // Store the set between runs
+      static auto cache = assertion_cache::create_empty_set();
+      algorithms.emplace_back(std::make_unique<assertion_cache>(
+        cache, !options.get_bool_option("forward-condition")));
+    }
   }
 
   if(options.get_bool_option("smt-during-symex"))
@@ -598,15 +607,6 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       result->remaining_claims,
       BigInt(eq->SSA_steps.size()) - ignored);
 
-    // Run cache if user has specified the options
-    if(options.get_bool_option("cache-asserts"))
-    {
-      // Store the set between runs
-      static auto cache = crc_assert_cache::create_empty_set();
-      crc_assert_cache(
-        eq->SSA_steps, cache, !options.get_bool_option("forward-condition"))
-        .run();
-    }
     if(options.get_bool_option("document-subgoals"))
     {
       std::ostringstream oss;
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -51,7 +51,7 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
     if(options.get_bool_option("cache-asserts"))
       // Store the set between runs
       algorithms.emplace_back(std::make_unique<assertion_cache>(
-        config.caching_db, !options.get_bool_option("forward-condition")));
+        config.ssa_caching_db, !options.get_bool_option("forward-condition")));
   }
 
   if(options.get_bool_option("smt-during-symex"))
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -601,8 +601,10 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
     // CACHE
     if(options.get_bool_option("cache-asserts"))
     {
-     static auto cache = crc_assert_cache::create_empty_set();
-     crc_assert_cache(eq->SSA_steps, cache, options.get_bool_option("forward-condition")).run();
+      static auto cache = crc_assert_cache::create_empty_set();
+      crc_assert_cache(
+        eq->SSA_steps, cache, options.get_bool_option("forward-condition"))
+        .run();
     }
     if(options.get_bool_option("document-subgoals"))
     {
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -32,6 +32,7 @@
 #include <util/migrate.h>
 #include <util/show_symbol_table.h>
 #include <util/time_stopping.h>
+#include <util/cache.h>
 
 bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
   : options(opts), context(_context), ns(context)
@@ -597,6 +598,12 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       result->remaining_claims,
       BigInt(eq->SSA_steps.size()) - ignored);
 
+    // CACHE
+    if(options.get_bool_option("cache-asserts"))
+    {
+     static auto cache = crc_assert_cache::create_empty_set();
+     crc_assert_cache(eq->SSA_steps, cache, options.get_bool_option("forward-condition")).run();
+    }
     if(options.get_bool_option("document-subgoals"))
     {
       std::ostringstream oss;
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -49,12 +49,9 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
 
     // Run cache if user has specified the options
     if(options.get_bool_option("cache-asserts"))
-    {
       // Store the set between runs
-      static auto cache = assertion_cache::create_empty_set();
       algorithms.emplace_back(std::make_unique<assertion_cache>(
-        cache, !options.get_bool_option("forward-condition")));
-    }
+        config.caching_db, !options.get_bool_option("forward-condition")));
   }
 
   if(options.get_bool_option("smt-during-symex"))
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -47,7 +47,7 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
     else
       algorithms.emplace_back(std::make_unique<symex_slicet>(options));
 
-    // Run cache if user has specified the options
+    // Run cache if user has specified the option
     if(options.get_bool_option("cache-asserts"))
       // Store the set between runs
       algorithms.emplace_back(std::make_unique<assertion_cache>(
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -598,12 +598,13 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       result->remaining_claims,
       BigInt(eq->SSA_steps.size()) - ignored);
 
-    // CACHE
+    // Run cache if user has specified the options
     if(options.get_bool_option("cache-asserts"))
     {
+      // Store the set between runs
       static auto cache = crc_assert_cache::create_empty_set();
       crc_assert_cache(
-        eq->SSA_steps, cache, options.get_bool_option("forward-condition"))
+        eq->SSA_steps, cache, !options.get_bool_option("forward-condition"))
         .run();
     }
     if(options.get_bool_option("document-subgoals"))
--- a/src/util/cache.cpp
+++ b/src/util/cache.cpp
@@ -3,35 +3,27 @@
 #include <utility>
 #include <util/crypto_hash.h>
 
-namespace
-{
-std::string hash_irep2(const expr2tc e)
-{
-  crypto_hash h;
-  e->hash(h);
-  h.fin();
-  return h.to_string();
-}
-} // namespace
 void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
 {
-  auto cond_hash = hash_irep2(step.cond);
-  auto guard_hash = hash_irep2(step.guard);
-  assert_pair pair = std::make_pair(guard_hash, cond_hash);
-  auto it = assert_set.find(pair);
-  if(it == assert_set.end())
+  auto [it, ins] = db.emplace(std::make_pair(step.cond, step.guard));
+  if(!ins)
   {
-    log_debug("Cache missed");
-    assert_set.emplace(pair);
+    log_debug("Cache hits: {}", ++hits);
+    step.cond = constant_bool2tc(trivial_value);
   }
   else
-  {
-    if((guard_hash == it->first) && (cond_hash == it->second))
-    {
-      log_debug("Cache hits: {}", ++hits);
-      step.cond = constant_bool2tc(trivial_value);
-    }
-    else
-      log_debug("Cache had a hash collision");
-  }
+    log_debug("Cache missed");
+}
+
+bool assertion_cache::run(symex_target_equationt::SSA_stepst &eq)
+{
+  fine_timet algorithm_start = current_time();
+  for(auto &step : eq)
+    run_on_step(step);
+  fine_timet algorithm_stop = current_time();
+  log_status(
+    "Caching time: {}s (removed {} assignments)",
+    time2string(algorithm_stop - algorithm_start),
+    hits);
+  return true;
 }
--- a/src/util/cache.cpp
+++ b/src/util/cache.cpp
@@ -3,7 +3,9 @@
 #include <utility>
 void crc_assert_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
 {
-  crc_pair pair = std::make_pair(step.cond->do_crc(), step.guard->do_crc());
+  auto cond_crc = step.cond->do_crc();
+  auto guard_crc = step.guard->do_crc();
+  crc_pair pair = std::make_pair(guard_crc, cond_crc);
   auto it = crc_set.find(pair);
   if(it == crc_set.end())
   {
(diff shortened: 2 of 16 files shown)


src/esbmc/bmc.cpp
```diff
@@ -595,6 +596,14 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       result->remaining_claims,
       BigInt(eq->SSA_steps.size()) - ignored);
 
+    // CACHE
```
lucasccordeiro: Improve this comment :-)

src/util/algorithms.cpp
```diff
@@ -38,4 +38,37 @@ bool goto_functions_algorithm::runOnFunction(
   std::pair<const dstring, goto_functiont> &)
 {
   return true;
-}
\ No newline at end of file
+}
+
+bool ssa_algorithm::run()
```
lucasccordeiro: Can you please add some comments to explain this piece of code?

src/util/cache.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <set>
+#include <util/algorithms.h>
+
+using crc_hash = unsigned long long;
+using crc_pair = std::pair<crc_hash, crc_hash>;
+class crc_assert_cache : public ssa_algorithm
```
lucasccordeiro: Add some comments to document this class.

lucasccordeiro (changes requested):
@rafaelsamenezes: Could you please run this PR over the SV-COMP benchmarks? It would be good to see the results with and without the cache system.

src/esbmc/bmc.cpp
```diff
@@ -595,6 +596,15 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       result->remaining_claims,
       BigInt(eq->SSA_steps.size()) - ignored);
 
+    // Run cache if user has specified the options
```
lucasccordeiro: "options" -> "option".

lucasccordeiro (approved):
LGTM.

I would like to see the SV-COMP results before we merge this PR.

fbrausse (commented):
That's a nice idea. I have a few concerns about the current implementation:
1. Most importantly: hashing is not injective! There can be (unlikely, but possibly) hash collisions. Replacing a condition solely based on equality of hashes is dangerous. Can we include a check for equality of the actual terms?
2. What are all these new `static` local variables doing there? Are they necessary? Why are they necessarily unique?
3. Using a `std::map` with an enum as key is inefficient. As are all these virtual calls in `ssa_algorithm::run`. Clearly, this class is meant to be subclassed and make these derived implementations shorter. But it does so at the expense of runtime. Is it really that much more cumbersome to put the loop into the subclass? The number of SSA-steps in particular can be very large.

rafaelsamenezes:
> Most importantly: hashing is not injective! There can be (unlikely, but possibly) hash collisions. Replacing a condition solely based on equality of hashes is dangerous. Can we include a check for equality of the actual terms?

Nice catch, I will add it.

> What are all these new static local variables doing there? Are they necessary? Why are they necessarily unique?

One of those was a mistake. Right now I added only one static variable which is the cache itself. I am unsure whether I should keep it static or move it into config.

> Using a std::map with an enum as key is inefficient. 

I removed the map and just added a plain switch. 

> As are all these virtual calls in ssa_algorithm::run. Clearly, this class is meant to be subclassed and make these derived implementations shorter. But it does so at the expense of runtime. Is it really that much more cumbersome to put the loop into the subclass? The number of SSA-steps in particular can be very large.

@fbrausse I think that the runtime impact won't be that impactful. For instance, this was inspired by LLVM passes which also rely on virtual methods for the code transformation/analyses. I would argue the other way around, if an algorithm has to be more performant then we can just do it without the algorithm interface.

Edit: Using std::function members and initialization though the constructor would be better?

fbrausse:
That's not what I meant :) `do_crc()` computes a hash of the type2t or expr2t. Equal `do_crc()` hashes do not imply equal type2t or expr2t. Couldn't you just store the condition and guard in a unordered_set? Might need to specialize `std::hash` for type2t and expr2t...

The `static auto cache` in `bmct::run_thread()` is pretty confusing.to me, as is part of the rest of the API you're introducing.
I would've expected
- `cache` to be local state (a member) of `class crc_assert_cache`, not just a reference to an externally managed set, and
- `ssa_algorithm::run()` to be invoked with `SSA_steps` as parameter - that's what it should run on, isn't it?

If you really want to keep `cache` external to the "pass", wouldn't it make more sense as a member of `class bmct`?

As a side note: To me `do_crc` is a completely misleading name:
a) it caches its result, doesn't "do" anything
b) it recursively computes a hash, it has nothing to do with CRCs or checksums in general.

rafaelsamenezes:
> That's not what I meant :) do_crc() computes a hash of the type2t or expr2t. Equal do_crc() hashes do not imply equal type2t or expr2t. 

I see what you mean now. 

> Couldn't you just store the condition and guard in a unordered_set? 

I am not sure if I follow: you mean something such: `std::pair<const &expr2tc, const &expr2tc>`?

> Might need to specialize std::hash for type2t and expr2t

I think we have it somewhere in the code (we do state hashing after all). I will look for it

> cache to be local state (a member) of class crc_assert_cache, not just a reference to an externally managed set,

The thing is, how to pass this member between runs? We need some way to store/load it between `run_bmc` calls. In this case, I was thinking about kind and incremental runs as the same assertions are going to be verified again.

> ssa_algorithm::run() to be invoked with SSA_steps as parameter - that's what it should run on, isn't it?

I would say the run is just to... run it. When I came up with the idea of the algorithm interface I was thinking of being able to create a list of transformations/analyses and iterate over it with the `run` method (similar to a pass manager of LLVM).

> f you really want to keep cache external to the "pass", wouldn't it make more sense as a member of class bmct?

Kinda yes, however, I felt that would affect readability if we had a member of a class that is only used in a specific and optional path.

fbrausse:
> > Couldn't you just store the condition and guard in a unordered_set?
> 
> I am not sure if I follow: you mean something such: `std::pair<const &expr2tc, const &expr2tc>`?

Yes, maybe even without references: `std::unordered_set<std::pair<expr2tc,expr2tc>>`.

> > Might need to specialize std::hash for type2t and expr2t
> 
> I think we have it somewhere in the code (we do state hashing after all). I will look for it

Thanks, I also recall seeing a std::hash-compatible type somewhere.

> > cache to be local state (a member) of class crc_assert_cache, not just a reference to an externally managed set,
> 
> The thing is, how to pass this member between runs? We need some way to store/load it between `run_bmc` calls. In this case, I was thinking about kind and incremental runs as the same assertions are going to be verified again.

Well, you're now in `bmct::run_thread()` creating a temporary, anonymous object of type `crc_assert_cache` only to invoke `run()` and destroy it again. Wouldn't keeping around this `crc_assert_cache` object in `bmct` (or config, depending on whether we want to centralize "state") take care of passing the cache?

> > ssa_algorithm::run() to be invoked with SSA_steps as parameter - that's what it should run on, isn't it?
> 
> I would say the run is just to... run it.

`ssa_algorithm::run()` would seem to disagree by actually only iterating through the SSA-steps :)

> When I came up with the idea of the algorithm interface I was thinking of being able to create a list of transformations/analyses and iterate over it with the `run` method (similar to a pass manager of LLVM).

Disclaimer: personal opinion following, no technical reason, just a suggestion regarding API design. The types (`goto_functionst` and `SSA_stepst`) the 2 specific kinds of algorithms are operating on, respectively, are fixed by the architecture of ESBMC. We could move the `run()` method from the generic base class `algorithm` to the `goto_functions_algorithm` and `ssa_algorithm`, each taking as argument the respective type they work on. Then `algorithm` itself wouldn't be runnable¹, but that's OK again by the design of ESBMC: After symex, all information is transferred from GOTO to SSA. It doesn't make a lot of sense to postpone running a GOTO algorithm after an SSA one. Then we'd have 2 separate kinds of "passes".

¹ Alternatively, `algorithm` could be a class template taking the argument type of `run()` as template parameters.

> > f you really want to keep cache external to the "pass", wouldn't it make more sense as a member of class bmct?
> 
> Kinda yes, however, I felt that would affect readability if we had a member of a class that is only used in a specific and optional path.

In that case, maybe the config is a better place as it already keeps a bunch of optionally set global information around.

src/util/algorithms.cpp
```diff
@@ -38,4 +38,37 @@ bool goto_functions_algorithm::runOnFunction(
...
+        break;
+      case ssa_type::OUTPUT:
+        run_on_output(step);
+        break;
+      case ssa_type::RENUMBER:
+        run_on_renumber(step);
+        break;
+      case ssa_type::SKIP:
+        run_on_skip(step);
+        break;
```
kunjsong01: missing default?

src/util/cache.cpp
```diff
@@ -0,0 +1,25 @@
...
+  }
+  else
+  {
+    if((guard_crc == it->first) && (cond_crc == it->second))
+    {
+      log_debug("Cache hits: {}", ++hits);
+      step.cond = constant_bool2tc(trivial_value);
+    }
+    else
+      log_debug("Cache had a hash collision");
```
kunjsong01: should we report an error and abort here?

src/util/cache.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @brief This class stores all asserts conditions and guards
+ *        for a given SSA, if some pair was already added
+ *        then it will convert it to be a trivial case,
+ *        e.g. GUARD => COND ----> GUARD => 1
+ */
+class crc_assert_cache : public ssa_algorithm
+{
+public:
+  using crc_pair = std::pair<crc_hash, crc_hash>;
```
kunjsong01: Perhaps a comment to explain the pair?
e.g. // first refers to guard_crc, second refers to cond_crc?

src/esbmc/bmc.cpp
```diff
@@ -595,6 +596,15 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
+    if(options.get_bool_option("cache-asserts"))
+    {
+      // Store the set between runs
+      static auto cache = crc_assert_cache::create_empty_set();
```
kunjsong01: Why static?

kunjsong01 (changes requested):

kunjsong01 (approved):
LGTM, just left some minor comments above.

rafaelsamenezes:
> ssa_algorithm::run() would seem to disagree by actually only iterating through the SSA-steps :)

I could make every runOnX function be abstract, in this way you couldn't invoke ssa_algorithm directly. It will be more verbose for the inherited classes though.

rafaelsamenezes:
> Disclaimer: personal opinion following, no technical reason, just a suggestion regarding API design. The types (goto_functionst and SSA_stepst) the 2 specific kinds of algorithms are operating on, respectively, are fixed by the architecture of ESBMC. We could move the run() method from the generic base class algorithm to the goto_functions_algorithm and ssa_algorithm, each taking as argument the respective type they work on. Then algorithm itself wouldn't be runnable¹, but that's OK again by the design of ESBMC: After symex, all information is transferred from GOTO to SSA. It doesn't make a lot of sense to postpone running a GOTO algorithm after an SSA one. Then we'd have 2 separate kinds of "passes".

@fbrausse Could you elaborate more? By what I understood you mean removing the `run` method from `algorithms`. And, every inheritance would create it is own run, e.g. `class foo_algorithm : public algorithm { bool run(Foo &foo); };`. I don't mind changing this, it would solve the issue of the static variable. So:

1. Does still make sense to have the algorithms class?
2. When/Where should we initialize the algorithms? My guess is at bmc class

fbrausse:
> I could make every runOnX function be abstract, in this way you couldn't invoke ssa_algorithm directly. It will be more verbose for the inherited classes though.

Don't think that's needed.

> Could you elaborate more? By what I understood you mean removing the `run` method from `algorithms`. And, every inheritance would create it is own run, e.g. `class foo_algorithm : public algorithm { bool run(Foo &foo); };`.

Yes, that's exactly what I was thinking. 

> 1. Does still make sense to have the algorithms class?

There would be only `has_sideeffect()` left in `algorithm`. I can't find users of it though it seems potentially necessary information for classifying when to run a pass. Though an algorithm will have a hard time returning anything meaningful here without knowledge of the data it is being run on. In general this info is part of the run itself.

> 2. When/Where should we initialize the algorithms? My guess is at bmc class

I suppose. You mentioned the initial motivation was to have a list of passes to be run. Does this list exist anywhere, yet?

Btw. the slicer could also be an algorithm, though it obtains all its state from the config at the moment....

rafaelsamenezes:
Ok, from @fbrausse suggestions, I think this should be put on hold. First I will create a PR for:

1. Convert `algorithms` into a template class (where `run` will receive the specialization).
2. Adapt the current goto_functions algorithms to use this new flow.
3. At esbmc_parseoptions class, add a list of algorithm<goto_functionst> to be run during `process_goto_program`
4. Create the ssa_step_algorithm and convert the slicer into it, moving the list of ignored symbols into the algorithm itself.
5. At bmct  class, add list of algorithm<ssa_step> (or similar) to be executed at `run_thread`

After that, I will come back to this PR. What do you think?

fbrausse:
That sounds great!

src/util/algorithms.cpp
fbrausse: I don't think `default:` is missing: what would you do in there? The way it is now, we'll get a compiler warning when the `ssa_type` enum changes. IMHO that's better than either aborting (making the compiled code slower) or just `break`, which only hides the warning.

src/util/cache.cpp
rafaelsamenezes: I don't think it is an error. It just means that we shouldn't replace the assertion.

src/esbmc/bmc.cpp
rafaelsamenezes: I moved it into config

src/util/cache.cpp
```diff
@@ -0,0 +1,37 @@
...
+void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
+{
+  auto cond_hash = hash_irep2(step.cond);
+  auto guard_hash = hash_irep2(step.guard);
+  assert_pair pair = std::make_pair(guard_hash, cond_hash);
+  auto it = assert_set.find(pair);
+  if(it == assert_set.end())
+  {
+    log_debug("Cache missed");
+    assert_set.emplace(pair);
```
fbrausse: `emplace` only if not found is duplicate work. You can use `auto [it,ins] = assert_set.emplace(pair);`, then the boolean `ins` will hold the info whether `pair` was in this set before.

src/util/cache.cpp
```diff
@@ -0,0 +1,37 @@
...
+  e->hash(h);
+  h.fin();
+  return h.to_string();
+}
+} // namespace
+void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
+{
+  auto cond_hash = hash_irep2(step.cond);
+  auto guard_hash = hash_irep2(step.guard);
+  assert_pair pair = std::make_pair(guard_hash, cond_hash);
```
fbrausse: As discussed [earlier](https://github.com/esbmc/esbmc/pull/827#issuecomment-1198354788), could you put the expressions themselves into the `assert_set` instead of just the hashes? The `std::unordered_set` will take care of hash collisions.

src/util/config.h
```diff
@@ -140,6 +146,9 @@ class configt
   static std::string this_operating_system();
 
   static triple host();
+
+  // For caching
+  std::set<assert_pair> caching_db;
```
fbrausse: I think you want std::unordered_set as it is hash-based, std::set relies on the order.

fbrausse (changes requested):

rafaelsamenezes:
I will still work on this PR but the initial results look good:

300s, Master
```
  correct:            8456
    correct true:     5128
    correct false:    3328
  incorrect:            54
    incorrect true:     25
    incorrect false:    29
  unknown:            7138
```

300s, this PR
```
   correct:            8561
     correct true:     5169
     correct false:    3392
   incorrect:            56
     incorrect true:     25
    incorrect false:    31
   unknown:            7031
   Score:             12434 (max: 25567)
```

rafaelsamenezes:
@fbrausse would you mind taking another look? I had to create a new hashing function in order to use `unordered_set`

src/util/config.h
```diff
@@ -108,6 +108,9 @@ class configt
   static std::string this_operating_system();
 
   static triple host();
+
+  // For caching
```
fbrausse: Could you mention the "cache-asserts" option here in order to add some context?

src/util/crypto_hash.h
```diff
@@ -14,6 +14,23 @@ class crypto_hash
 
   bool operator<(const crypto_hash &h2) const;
 
+  std::array<unsigned int, 5> to_array() const
+  {
+    // TODO: CPP+20 will have a to_array function
```
fbrausse: C++20 is already released. It *does* have this function. As a free-standing function. Since the `hash` array is already publicly accessible, there is no need for this function in the `crypto_hash` class. Maybe move it to the unit test case?

src/util/crypto_hash.h
```diff
@@ -14,6 +14,23 @@ class crypto_hash
...
+    std::copy(hash, hash + 5, result.begin());
+    return result;
+  }
+
+  size_t to_size_t() const
+  {
+    size_t result = hash[0];
+    for(int i = 1; i < 5; i++)
+      // Do we care about overlaps?
+        result ^= (hash[i] << (sizeof(unsigned int)*i));
```
fbrausse: You're throwing away bits of the hash through this shift. As `crypto_hash` is already supposed to be a crypto-strength hash¹, the probability of `hash[0] ^ hash[1]` having any recognizeable pattern is the same as that of `hash[0] ^ (hash[1] << 4)`. Similarly the probability of h0,h1 is the same as that of h1,h0. You're not combining hash values of independent objects where the order matters. I would remove the shift.

¹ I am aware that this is SHA-1, which is broken. Not for hash-maps, though. Still, if that's a problem change the implementation to anything that satisfies the `crypto_hash` promise.

fbrausse (approved):
Very nice PR, thanks for implementing it! I only have a few minor comments about the hashing.

src/util/crypto_hash.h
rafaelsamenezes: I have changed the implementation to use only ^. I will also start another sv-comp run to see if it changes anything.

lucasccordeiro:
@rafaelsamenezes: I have just updated the stats file at https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/stats-300s.txt. 

Next time, before merging the PR, can I ask you to update this file?

PR #828, opened 2022-07-28
Jimple - Array of arrays support

rafaelsamenezes:
This adds support for:

- java.lang.Integer and its valueOf
- Array of array type, e.g. `int[][] var;`


lucasccordeiro (approved):
@rafaelsamenezes: could you please add a test case that produces ``VERIFICATION FAILED``?

Additionally, can you also create an issue to handle the intrinsic later?

PR #831, opened 2022-08-01
Revamp of run method for algorithms 

rafaelsamenezes:
As discussed by https://github.com/esbmc/esbmc/pull/827#issuecomment-1199631426

### DIFF
--- a/src/esbmc/esbmc_parseoptions.h
+++ b/src/esbmc/esbmc_parseoptions.h
@@ -7,6 +7,7 @@
 #include <util/cmdline.h>
 #include <util/options.h>
 #include <util/parseoptions.h>
+#include <util/algorithms.h>
 
 extern const struct group_opt_templ all_cmd_options[];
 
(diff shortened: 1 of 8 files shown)


src/esbmc/esbmc_parseoptions.h
```diff
@@ -76,6 +77,8 @@ class esbmc_parseoptionst : public parseoptions_baset, public language_uit
   FILE *out = stdout;
   FILE *err = stderr;
 
+  std::vector<goto_functions_algorithm> goto_preprocess_algorithms;
```
fbrausse: This should contain pointers to accomodate objects of derived types, std::unique_ptr should fit here.

fbrausse (changes requested):

src/esbmc/esbmc_parseoptions.h
rafaelsamenezes: Done

fbrausse (approved):

PR #833, opened 2022-08-03
SSA step algorithms

rafaelsamenezes:
As a partial requirement of https://github.com/esbmc/esbmc/pull/827#issuecomment-1199631426

This PR:

1. Creates the ssa_step_algorithm
2. Turns the slicer algorithms into ssa_step_algorithm
3. Adds a list of ssa_step_algorithms into bmct

### DIFF
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -39,7 +39,7 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
   interleaving_number = 0;
   interleaving_failed = 0;
 
-  // Set algorithms
+  // The next block will initialize the algorithms used for the analysis.
   {
     if(opts.get_bool_option("no-slice"))
       algorithms.emplace_back(std::make_unique<simple_slice>());
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -583,7 +583,6 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
       a->run(eq->SSA_steps);
       ignored += a->ignored();
     }
-    //slice(eq);
 
     if(
       options.get_bool_option("program-only") ||
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -44,8 +44,7 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
     if(opts.get_bool_option("no-slice"))
       algorithms.emplace_back(std::make_unique<simple_slice>());
     else
-      algorithms.emplace_back(
-        std::make_unique<symex_slicet>(opts.get_bool_option("slice-assumes")));
+      algorithms.emplace_back(std::make_unique<symex_slicet>(options));
   }
 
   if(options.get_bool_option("smt-during-symex"))
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -39,6 +39,15 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
   interleaving_number = 0;
   interleaving_failed = 0;
 
+  // Set algorithms
+  {
+    if(opts.get_bool_option("no-slice"))
+      algorithms.emplace_back(std::make_unique<simple_slice>());
+    else
+      algorithms.emplace_back(
+        std::make_unique<symex_slicet>(opts.get_bool_option("slice-assumes")));
+  }
+
   if(options.get_bool_option("smt-during-symex"))
   {
     runtime_solver = std::shared_ptr<smt_convt>(create_solver("", ns, options));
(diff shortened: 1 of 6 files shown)


src/esbmc/bmc.cpp
```diff
@@ -39,6 +39,15 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
   interleaving_number = 0;
   interleaving_failed = 0;
 
+  // Set algorithms
```
lucasccordeiro: Can you please improve this comment?

src/esbmc/bmc.cpp
```diff
@@ -568,15 +577,13 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
-      "Slicing time: {}s (removed {} assignments)",
-      time2string(slice_stop - slice_start),
-      ignored);
+    BigInt ignored;
+    for(auto &a : algorithms)
+    {
+      a->run(eq->SSA_steps);
+      ignored += a->ignored();
+    }
+    //slice(eq);
```
lucasccordeiro: Remove?

src/goto-symex/slice.h
```diff
@@ -2,22 +2,182 @@
...
+   * of the dependency symbols, then it will be ignored.
+   *
+   * @param eq symex formula to be sliced
+   */
+  bool run(symex_target_equationt::SSA_stepst &eq) override
+  {
+    fine_timet algorithm_start = current_time();
+    iterate(eq, false); // iterate in reverse
+    fine_timet algorithm_stop = current_time();
+    log_status(
```
lucasccordeiro, line 70: nice!

lucasccordeiro (approved):
I have minor comments only.

src/esbmc/bmc.cpp
```diff
@@ -39,6 +39,15 @@ bmct::bmct(goto_functionst &funcs, optionst &opts, contextt &_context)
   interleaving_number = 0;
   interleaving_failed = 0;
 
+  // Set algorithms
+  {
+    if(opts.get_bool_option("no-slice"))
+      algorithms.emplace_back(std::make_unique<simple_slice>());
+    else
+      algorithms.emplace_back(
+        std::make_unique<symex_slicet>(opts.get_bool_option("slice-assumes")));
```
mikhailramalho: Can you pass opts here instead? We can then query any option in the constructor.

src/esbmc/bmc.cpp
rafaelsamenezes: Improved it

src/esbmc/bmc.cpp
rafaelsamenezes: Done

src/util/algorithms.cpp
```diff
@@ -39,3 +41,46 @@ bool goto_functions_algorithm::runOnFunction(
...
+    run_on_skip(step);
+    break;
+  default:
+    log_warning("Invalid type for SSA step");
+  }
+}
+
+void ssa_step_algorithm::iterate(
+  symex_target_equationt::SSA_stepst &eq,
+  bool in_order)
```
fbrausse: Do we need this function? The contract is not really clear when `in_order` is false. It could be any order. An algorithm doesn't necessarily have to visit all steps in the list exactly once.

fbrausse (approved):
LGTM, only one minor comment.

src/util/algorithms.cpp
rafaelsamenezes: > Do we need this function? 

I think so.

> The contract is not really clear when in_order is false

I can try a better naming such as iterate_in_reverse or an enum class DIRECTION { FORWARD, BACKWARD }.

> An algorithm doesn't necessarily have to visit all steps in the list exactly once

Yeah, this is mostly a helper function though, we also have the `run_on_step` in case
that the algorithm needs to do a custom flow.

fbrausse: > we also have the run_on_step in case that the algorithm needs to do a custom flow.

Right, that's why this function seems pretty useless to me. Shortening a long `switch` could be nice, but a 2-line `for` loop? Not sure a function call to a base class makes that easier to read.

lucasccordeiro: In this PR, this method seems unnecessary since we always call the method `iterate` with the `in_order` set to false. So, I agree with @fbrausse here.

@rafaelsamenezes: Is there some other class in our codebase that needs to use this method by setting `in_order` to true?

rafaelsamenezes: It will be used here at PR https://github.com/esbmc/esbmc/pull/827

fbrausse: A function that branches off to entirely different behaviours based on a parameter value (**edit** with dubious semantics, see above) is not well-defined. There is no need to encapsulate a for-loop that just delegates to another function.

rafaelsamenezes: I removed the iterate method

PR #643, opened 2022-03-09
Improve regression

kunjsong01:
Based on the comments, split PR #625 into multiple small PRs. 

This PR contains the changes for regression scripts introducing some new features: 
1. Test log retention
2. Provide additional command line args for each test case generated

Usage: 

Use `--old-frontend` to run regression with old frontend:
```
python3 testing_tool.py \
--regression esbmc-cpp/cpp \
--tool ../../release/bin/esbmc --mode ALL \
--old-frontend
```

Add `--retain-logs` to to keep the logs for all test cases to obtain the passing rate and error signatures.

PR #834, opened 2022-08-03
Add a new CI workflow to bring up a benchmark

kunjsong01:
Adding a new CI workflow to enable user to run a random benchmark on Github ubuntu container and collect logs to generate stats about error signatures. 

### The new workflow: 
workflow

### Logs are available to download: 
new_artifact

### Example run: 
https://github.com/kunjsong01/esbmc/actions/runs/2793101981


lucasccordeiro (approved):

regression/CMakeLists.txt
```diff
@@ -9,12 +9,24 @@ set(ESBMC_REGRESSION_TOOL "${CMAKE_CURRENT_SOURCE_DIR}/testing_tool.py")
...
 function(add_esbmc_regression_test folder modes test)
     set(test_name "regression/${folder}/${test}")
-    add_test(NAME ${test_name}
-             COMMAND ${Python_EXECUTABLE} ${ESBMC_REGRESSION_TOOL}
-                     --tool=${ESBMC_BIN}
-                     --regression=${CMAKE_CURRENT_SOURCE_DIR}/${folder}
-                     --modes ${modes}
-                     --file=${test})
+
+    if(NOT BENCHBRINGUP)
```
rafaelsamenezes: Could you add this option to the Options.cmake script?

regression/CMakeLists.txt
```diff
@@ -9,12 +9,24 @@ set(ESBMC_REGRESSION_TOOL "${CMAKE_CURRENT_SOURCE_DIR}/testing_tool.py")
...
+                       --file=${test})
+    else()
+      add_test(NAME ${test_name}
+               COMMAND ${Python_EXECUTABLE} ${ESBMC_REGRESSION_TOOL}
+                       --tool=${ESBMC_BIN}
+                       --regression=${CMAKE_CURRENT_SOURCE_DIR}/${folder}
+                       --modes ${modes}
+                       --file=${test}
+                       --timeout=$ENV{TIMEOUT}
+                       --benchbringup) # logs will be save in the OUT_DIR directory
```
rafaelsamenezes: You can replace this IF/ELSE by creating a string "BENCHBRINGUP" option (similar to what we do for regression, e.g. ${REGRESSIONS_JIMPLE})

regression/CMakeLists.txt
```diff
@@ -53,13 +65,27 @@ elseif(WIN32)
...
     set(REGRESSIONS esbmc-unix esbmc-unix2 esbmc ${REGRESSIONS_SOLIDITY}
                     ${REGRESSIONS_OLD_FRONTEND}
                     cbmc cstd llvm floats floats-regression
                     k-induction esbmc-cpp/cpp esbmc-cpp/cbmc csmith
                     k-induction-parallel nonz3 ${REGRESSIONS_BITWUZLA}
                     incremental-smt esbmc-cpp11/cpp esbmc-cpp11/constructors esbmc-cpp11/new-delete esbmc-cpp11/reference
                     ${REGRESSIONS_GOTO_CONTRACTOR} ${REGRESSIONS_JIMPLE} extensions)
+  else()
+    if(NOT DEFINED ENV{BENCHMARK_TO_RUN} OR NOT DEFINED ENV{LOG_DIR})
+      message( FATAL_ERROR "ENV vars are not defined. Please specify which benchmark to run and the corresponding log directory." )
```
rafaelsamenezes: I think that this is misleading. Setting a LOG_DIR during the CMake configuration will not ensure that there will be one during ctest run. I suggest to either:

1. Add the timeout as a flag of the Python script.
2. Turn this FATAL_ERROR into a warning stating that you need these flag setted during ctest execution

.github/workflows/benchbringup.yml
```diff
@@ -0,0 +1,65 @@
...
+        required: true
+        default: '300'
+      options:
+        description: 'Command-line flags to pass via ESBMC_OPTS'
+        required: false
+        default: ''
+      benchmark:
+        description: 'Benchmark to run (e.g. any benchmark as in regression/<benchmark>)'
+        required: true
+        #default: 'esbmc-cpp/cpp'
```
fbrausse: Please remove this commented line.

regression/testing_tool.py
```diff
@@ -197,41 +195,11 @@ def from_file(test_dir: str, name: str) -> BaseTest:
             first_line = fp.readline().strip()
             return TestParser.MODES[TestParser.detect_mode_by_header(first_line)](test_dir, name)
 
-class TestControl:
```
rafaelsamenezes, line 200: Why are you removing this class? Can't you extend this class and just change the filename?

.github/workflows/benchbringup.yml
```diff
@@ -0,0 +1,65 @@
+name: Run a Benchmark
+on:
+  workflow_dispatch:
+    inputs:
+      strategy:
```
rafaelsamenezes: remove this

.github/workflows/benchbringup.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Fix dubious ownership
+        run: git config --global --add safe.directory /__w/esbmc/esbmc
+      - name: Confirm the benchmark
+        run: echo ${{ github.event.inputs.benchmark }}
+      - name: Configure CMake
+        env:
+          TIMEOUT: ${{ inputs.timeout }}
+          BENCHMARK_TO_RUN: ${{ inputs.benchmark }}
+          LOG_DIR: ${{ inputs.logdirectory }}
+        run: mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -GNinja -DClang_DIR=/workspace/clang -DLLVM_DIR=/workspace/clang -DBUILD_STATIC=On -DBoolector_DIR=/workspace/boolector-release -DZ3_DIR=/workspace/z3 -DENABLE_REGRESSION=On -DBENCHBRINGUP=On
```
rafaelsamenezes: You should do the build with every solver here

regression/testing_tool.py
```diff
@@ -61,9 +62,6 @@ def mark_test_as_knownbug(self, issue: str):
             to it"""
         raise NotImplementedError
 
-    def append_test_arg(self, arg: str):
```
rafaelsamenezes, line 64: Why are you removing this method?

rafaelsamenezes (changes requested):
@kunjsong01 could you explain more about the changes to the Python script?

regression/testing_tool.py
kunjsong01: because the additional test args could be passed via `ESBMC_OPTS` (https://github.com/esbmc/esbmc/pull/643#issuecomment-1069014994)

regression/testing_tool.py
kunjsong01: becuase It does not suit the purpose any more. Extending it would lead to overengineering. 

Now it's much simpler: 
The `ESBMC_OPTS` environment variable and `def set_old_frontend` - they are doing the same thing. 
`def set_log_retention` is replaced by `BENCHMARK_BRINGUP` in this script.

.github/workflows/benchbringup.yml
kunjsong01: Done

.github/workflows/benchbringup.yml
kunjsong01: Done

regression/CMakeLists.txt
kunjsong01: Done

regression/CMakeLists.txt
kunjsong01: Done

regression/CMakeLists.txt
kunjsong01: LOG_DIR is just a place for ctest to dump the output logs. It's not used to ensure one test during ctest run. Actually, we want the parallelism indeed. 

The new workflow works as following: 
1. At configure time, CMake creates an empty folder `LOG_DIR`. 
2. At build time, nothing happens
3. At test time, the python script invoked by ctest will dump the output in `LOG_DIR`
4. When all test completes, Github workflow will zip `LOG_DIR` and upload it as archive@v2 for the user to download. 

This workflow requires three ENV vars:
`BENCHMARK_TO_RUN` : user-specified benchmark
`LOG_DIR` : the place to dump test logs
`TEST_TIMEOUT`: time out for each test case

As for timeout flag, we are already passing it via `--timeout=$ENV{TIMEOUT}`.

kunjsong01:
> @kunjsong01 could you explain more about the changes to the Python script?

The new workflow works as following:

    At configure time, CMake creates an empty folder LOG_DIR.
    At build time, nothing happens
    At test time, the python script invoked by ctest will dump the output in LOG_DIR
    When all test completes, Github workflow will zip LOG_DIR and upload it as archive@v2 for the user to download.

This workflow requires three ENV vars:
- `BENCHMARK_TO_RUN` : user-specified benchmark
- `LOG_DIR` : the place to dump test logs
- `TEST_TIMEOUT`: time out for each test case

The changes to the Python script aim to: 
- Removed overengineered (and usued) code in #643
- Dump logs to `LOG_DIR`
- Set timeout for each test case to `TEST_TIMEOUT`

kunjsong01:
With commit 2c51eac5163a24c354d3fe496dba26fc677a00d6, each test is able to include the correct library. The operational models are passed to esbmc via  environment variable `ESBMC_OPTS`. 

Now the error message matches standalone runs: 

Screenshot 2022-08-04 at 14 52 20

.github/workflows/benchbringup.yml
```diff
@@ -0,0 +1,61 @@
...
+  workflow_dispatch:
+    inputs:
+      timeout:
+        description: 'Timeout for each test case (in seconds)'
+        required: true
+        default: '300'
+      options:
+        description: 'Command-line flags to pass via ESBMC_OPTS'
+        required: false
+        default: '-I /__w/esbmc/esbmc/src/cpp/library'
```
fbrausse, line 12: `-I` is most likely not the right position in the include-dir list: `-I` puts it pretty much at the beginning of the search paths, while it should definitely not override user-given `-I`. `-isystem` would be the correct place, I believe. See https://github.com/esbmc/esbmc/blob/b6aac2b30bce8039426f994ca7e970b9f36e1386/src/clang-c-frontend/clang_c_language.cpp#L38-L43 This brings me to another question: Shouldn't adding this include path be handled by the `clang_cpp_language` in a similar manner?

.github/workflows/benchbringup.yml
kunjsong01: Done

.github/workflows/benchbringup.yml
kunjsong01: > Shouldn't adding this include path be handled by the clang_cpp_language in a similar manner?

It would be nice to have a coherent way to include operational model for both C and C++. However, this will be out of the scope as this PR just implements what's necessary to collect logs and get stats about the error signatures to guide our development. 

For the time being, we are still following the way old ESBMC++ includes these operational models using `-I` (in such a way unfortunately we have to ask the user to be careful with the `-I` order since the order matters). e.g.: 
```
cmd: "-I /home/ks/ESBMC_Project/esbmc/src/cpp/library -I ../../lib_dummy_2 -I ../../lib_dummy_1"

#include <...> search starts here:
 /home/ks/ESBMC_Project/esbmc/src/cpp/library
 ../../lib_dummy_2
 ../../lib_dummy_1
 /tmp/esbmc.78af-7fd4-f2b8/headers
 /tmp/esbmc-headers-980a-77a4-3e54
 /../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9
 /../lib/gcc/x86_64-linux-gnu/9/../../../../include/x86_64-linux-gnu/c++/9
 /../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/backward
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
```

kunjsong01:
@rafaelsamenezes Based on the discussion, I've added commit c6e35168950539aada9e5d103c78e37a19f51645 to add the prefix `ERROR:` and make error message grap-able. 

Now the error message becomes: 
```
Converting
ERROR: inheritance is not supported in get_struct_union_class_methods
Aborted (core dumped)
```

kunjsong01:
With the new workflow, we are now able to get stats about the error signatures when bringing up a bechmark. 
See https://github.com/esbmc/esbmc/wiki/Clang-Based-Frontend-for-Cpp 

It appears that I also need to update the [CI CD wiki](https://github.com/esbmc/esbmc/wiki/CI-CD) after merging thie PR.

.github/workflows/benchbringup.yml
fbrausse: As you agree that it's wrong, could you add a TODO-like comment to the `-I` in order to remind us that it's not as intended, yet?

kunjsong01: Done in commit 07353ecff7d2a1616e250d1990e65b2c764e34f3

rafaelsamenezes (approved):

.github/workflows/benchbringup.yml
fbrausse: Not sure what it's doing there, but better some TODO than none TODO. ;)

fbrausse (approved):
LGTM.

src/util/message.h
```diff
@@ -42,8 +42,10 @@ struct messaget
   static inline class
   {
     template <typename... Args>
-    void println(FILE *f, Args &&...args) const
+    void println(FILE *f, VerbosityLevel lvl, Args &&...args) const
     {
+      if(lvl == VerbosityLevel::Error)
+        fmt::print(f, "ERROR: ");
```
fbrausse, line 48: Sorry, I missed this part. We decided not to do it precisely because of this weird `ERROR: PARSING ERROR` kind of messages in https://github.com/esbmc/esbmc/pull/796#issuecomment-1183143762. This needs some uniformity.

PR #838, opened 2022-08-12
[cpp] renaming irept parse tree node name

kunjsong01:
CPP old-frontend uses a different name for c size `sizeof-type` rather than `#c_sizeof_type`: 

**Fixed the error message below:** 
```
esbmc: ../src/cpp/cpp_typecheck_type.cpp:18: virtual void cpp_typecheckt::typecheck_type(typet&): Assertion `type.is_not_nil()' failed.

Program received signal SIGABRT, Aborted.
```

**New error message:** 
```
Converting
Generating GOTO Program
ERROR: expression statement takes one operand

Aborted (core dumped)
```

We've seen this before (https://github.com/esbmc/esbmc/pull/811), probably due to zombie operand.


kunjsong01:
Closed. This is not a viable fix to restore old cpp frontend for string. 

As a side note on opertional models: 
As discussed in PR https://github.com/esbmc/esbmc/pull/835, there are multiple operaitonal modesl for C++string, e.g. `string`, `cstring`, `string.h`. Some functions are overlapping with the C models. These operational models should be reviewed. Ideally, C++ model should contain only the functions that C model doesn't have.

PR #842, opened 2022-08-26
GOTO to C translator

shmarovfedor:
### Update

This PR introduces a new module `goto2c` for translating GOTO programs into C. It takes `namespacet` and `goto_functionst` object and produces a string representation for the latter in correct C syntax. The main purpose of this translator is allowing ESBMC producing an artifact that can be analysed by other verification and testing tools when ESBMC itself cannot handle the given task. This can be done inside ESBMC after various analyses and optimisations of the GOTO program. Similarly, one can construct a `goto_functionst` object from a symex or SSA trace and translate it into C accordingly.

Currently this module can be used in ESBMC by using the `--goto2c` flag. This will wait for ESBMC to parse and translate the input C program into GOTO, and then `goto2c` will translate it back into C and output the resulting program into standard output. Alternatively, the translation result can be saved into a file specified by the `--output` option.

Roughly, the entire translation process consists of two stages: 1) **preprocessing** (`goto2c_preprocess.cpp`) and 2) **translating** (`goto2c_translate.cpp`).

Some of the **preprocessing** steps involve:
- Extracting the information from the symbol table (`namespacet`) and distributing it into various separate data structures for easier access during translation.
- Extracting initialisers for global and static variables from `ESBMC_main`.
- Reconstructing scopes for the variables in each GOTO program by analysing pairs of `DECL` and `DEAD` instructions. Each instruction is assigned a number representing the corresponding scope.
- Sorting all compound types (i.e., `struct/union`) in the correct order they should be defined in the translated program.
- Adjusting some assignments to fit the correct C syntax (e.g., assignments to `struct/union` variables), and modifying assignments that are allowed in GOTO but not in C (e.g., assignments to array variables past their initialization).
- Removing ESBMC instrinsic instructions that are not supported for translation at this (e.g., assignments to `dynamic_size`).

The **translation** stage is straight-forward. It begins by translating all types, functions and global variables declarations. Then it performs a full pass over the preprocessed GOTO program (i.e., `goto_functionst`) and translates each instruction according to C syntax.

All expressions (`exprt`) and types (`typet`) are translated by `expr2ccodet`. It inherits from the `expr2ct` class and overrides the methods that do not produce translations in correct C syntax. In my opinion, in order to avoid any confusion in the future `expr2ct` should be renamed (e.g., `expr2stringt` or `expr2textt`) since it does not produce C code. While `expr2ccodet` should be renamed to `expr2ct`.

There is also another stage in plans `goto2c_check.cpp` for **checking** the “correctness” of the GOTO program. All such checks kept in one place can eventually help honing the semantics of the GOTO language.

Although, the `goto2c` module is fairly independent from the rest of ESBMC, this PR introduces several changes to the “core” ESBMC:
- The `column` variable in `locationt` is assigned a value now in `clang-c-frontend`.
- All unnamed `struct/union` types and variables are given explicit unique names that include their `locationt` information. (This significantly simplifies things for translation and does not seem to break anything in the regression.)
- Some temporary variables with missing GOTO declarations are now explicitly declared (via GOTO `DECL` instructions).
- Some missing `locationt` information across several parts of ESBMC has been recovered.
- `goto_programt::instructiont` class is extended with two new members `scope_id` and `parent_scope_id`.


lucasccordeiro:
@shmarovfedor: Can I ask you for the status of this PR?

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -206,8 +210,36 @@ void clang_c_languaget::force_file_type()
   compiler_args.push_back("c");
 }
 
+std::vector<std::string> get_list_of_headers(const std::string &path)
+{
+  std::ifstream input_file(path);
+  std::string line;
+  std::vector<std::string> res;
+  std::regex reg("^\\s*\\#\\s*include\\s*(?:<(.+)>|\"(.+)\")");
```
fbrausse: It's entirely legal to `#include MACRO` if that macro has the <...> or "..." syntax. I don't know of a way to obtain this list of headers without watching the pre-processor or analyzing the `#line` directives in pre-processed code. As this might be too much trouble for the initial version of this feature, I suggest to add a comment about that here.

src/clang-c-frontend/clang_c_adjust_expr.cpp
```diff
@@ -596,14 +596,10 @@ void clang_c_adjust::adjust_side_effect_function_call(
 
       new_symbol.id = identifier;
       new_symbol.name = f_op.name();
-      new_symbol.location = expr.location();
+      // Leaving the location empty for now
```
rafaelsamenezes: Why? Does that break things?

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -206,8 +210,36 @@ void clang_c_languaget::force_file_type()
...
+    if(std::regex_search(line, match, reg))
+    {
+      // This will contain the match for the "#include <...>" pattern
+      if(!match[1].str().empty())
+        res.push_back(match[1]);
+
+      // This will contain the match for the "#include "..."" pattern
+      //if(!match[2].str().empty())
+      //  res.push_back(match[2]);
+    }
```
rafaelsamenezes: Should you be ignoring the second match?

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+bool expr2ccodet::is_anonymous_tag(std::string tag)
+{
+  std::smatch m;
+  return std::regex_search(tag, m, std::regex(" __anon_type_at_"));
+}
+
+bool expr2ccodet::is_padding(std::string tag)
+{
+  std::smatch m;
+  return std::regex_search(tag, m, std::regex("anon_pad\\$\\d+")) ||
```
rafaelsamenezes: Maybe use has_prefix?

src/clang-c-frontend/clang_c_language.h
```diff
@@ -38,6 +38,11 @@ class clang_c_languaget : public languaget
   bool from_type(const typet &type, std::string &code, const namespacet &ns)
     override;
 
+  // This map stores <path, list> pairs, where
+  // "list" - list of \#include's (both using angled brackets
+  // and quotation marks) contained in the file at "path"
+  static std::map<std::string, std::vector<std::string>> includes_map;
```
fbrausse: As the comment here goes into the specifics how this map is built, a reference to the above TODO comment might be useful here in the future.

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -206,8 +210,36 @@ void clang_c_languaget::force_file_type()
...
+  }
+  return res;
+}
+
 bool clang_c_languaget::parse(const std::string &path)
 {
+  std::vector<std::string> includes = get_list_of_headers(path);
+  includes_map[path].insert(
+    includes_map[path].end(), includes.begin(), includes.end());
```
fbrausse: You're copying the `includes` vector here, via iterators, but it is not used elsewhere in this function. As a performance optimization, it would be preferable to just `.insert([...], get_list_of_headers(path))` and to not lookup `path` in the `includes_map` twice.

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+  if(pos != std::string::npos)
+    shorthand.erase(0, pos + 1);
+
+  return shorthand;
+}
+
+bool expr2ccodet::is_anonymous_tag(std::string tag)
+{
+  std::smatch m;
+  return std::regex_search(tag, m, std::regex(" __anon_type_at_"));
```
fbrausse: I couldn't find the constant substring `__anon_type_at_` anywhere else in the code. How/Where is it constructed? Also, are you sure the single space in front has to be an ASCII space character and not just $WHITESPACE, aka `\s` or `[[:space:]]` in regex?

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+    if(width == config.ansi_c.long_long_int_width)
+      return q + sign_str + "long long int" + d;
+
+    if(width == config.ansi_c.char_width)
+      return q + sign_str + "char" + d;
+
+    if(width == config.ansi_c.short_int_width)
+      return q + sign_str + "short int" + d;
+
+    return q + sign_str + "_ExtInt(" + std::to_string(width.to_uint64()) + ")" +
```
fbrausse, line 222: `_ExtInt` is a clang-only feature to my knowledge aimed at integer types with more than 64 bits. Would it be possible to include GCC here, as well for the most common situation this case is triggered: bitfields? I realize, this would require knowledge of whether or not this declaration is part of a struct/union definition. So if it's too much work to support for now, I would again suggest to add a comment about that here. Also, `_ExtInt` is actually [deprecated](https://reviews.llvm.org/D108643), though for compatibility reasons it might still be the best option for goto2c. This reasoning should be mentioned as well.

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+  else if(src.id() == "floatbv" || src.id() == "fixedbv")
+  {
+    BigInt width = string2integer(src.width().as_string());
+
+    if(width == config.ansi_c.single_width)
+      return q + "float" + d;
+    if(width == config.ansi_c.double_width)
+      return q + "double" + d;
+    else if(width == config.ansi_c.long_double_width)
+      return q + "long double" + d;
```
fbrausse, line 236: What happens otherwise? Should there be a warning, an error or an assertion if none of these cases happens?

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+      type_str += "int";
+
+    if(width == config.ansi_c.long_int_width)
+      type_str += "long";
+
+    if(width == config.ansi_c.char_width)
+      type_str += "char";
+
+    if(width == config.ansi_c.short_int_width)
+      type_str += "short";
```
fbrausse, line 1463: Is there a `long long` version of `nondet`? What about other bit-widths, can/do they occur?

src/clang-c-frontend/expr2ccode.cpp
```diff
@@ -0,0 +1,1509 @@
...
+  }
+  else if(src.type().id() == "floatbv" || src.type().id() == "fixedbv")
+  {
+    BigInt width = string2integer(src.type().width().as_string());
+
+    if(width == config.ansi_c.single_width)
+      type_str += "float";
+    if(width == config.ansi_c.double_width)
+      type_str += "double";
```
fbrausse, line 1472: Same as above: what about `long double` and other bit-widths? Maybe add an assertion and/or a comment.

src/goto2c/goto2c.h
```diff
@@ -0,0 +1,73 @@
+/*******************************************************************\
+
+Module:
+
+Author: 
+
+\*******************************************************************/
```
fbrausse: Should we remove such empty headers?

fbrausse (commented):
I've not checked the precedence levels and overall have comments mainly about commenting some incomplete cases in the code. Other than that it looks good to me. All in all, this PR is great because it adds a largely complete relationship between the semantics of GOTO, irep1 and C making irep1 much more understandable (for those that read the source of this translator and understand C).

src/clang-c-frontend/expr2ccode.h
```diff
@@ -0,0 +1,101 @@
+/*******************************************************************\
```
lucasccordeiro: We don't use this type of header anymore.

src/goto-programs/goto_program.h
```diff
@@ -383,6 +383,13 @@ class goto_programt
     //! This is -1 if it's not a target.
     unsigned target_number;
 
+    //! Id of the scope within which a variable is declared (i.e., DECL).
+    //! It does not have a lot of meaning for other types of instructions.
+    unsigned int scope_id = 0;
```
lucasccordeiro, line 388: Shall we use BigInt here?

src/goto2c/goto2c.h
lucasccordeiro: I fully agree.

lucasccordeiro (changes requested):
Good stuff, @shmarovfedor.

I have just a few comments here.

Can I ask you to add some unit and regression tests here?

src/clang-c-frontend/clang_c_language.cpp
shmarovfedor: I removed this code from this PR now as it is not needed anymore

src/clang-c-frontend/clang_c_language.h
shmarovfedor: I removed this code from this PR now as it is not needed anymore

src/clang-c-frontend/clang_c_language.cpp
shmarovfedor: I removed this code from this PR now as it is not needed anymore

src/goto2c/goto2c.h
shmarovfedor: Done

src/clang-c-frontend/expr2ccode.h
shmarovfedor: Done

src/goto-programs/goto_program.h
shmarovfedor: I don't think this is necessary since `scope_id` and `parent_scope_id` need to be unique only within a given function (in the current implementation of the algorithm for identifying scopes). I think `unsigned int` is enough to store this information.

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: Those seem to be the only cases of `nondet`s that appear in `esbmc_intrinsics.h`. Apart from `__VERIFIER_nondet_schar()` which never appears in the regression, and which seems to be the same as `__VERIFIER_nonder_char()`

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: This method was never used, so I removed it.

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: Updated as suggested.

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: Updated as suggested.

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: Same as above. There does not seem to be a case of a `nondet` for `long double` in `esbmc_intrinsics.h`. I've added a comment in the description on the method.

fbrausse:
Strange that the Windows test case keeps timeouting. I don't have Windows here, but a wine run with the w64-mingw32 headers is far away from the 1500s timeout.
```
$  time wine release-windows/bin/esbmc.exe -I /usr/x86_64-w64-mingw32/usr/include/ ../regression/esbmc/cache_true/main.c --verbosity 9 --cache-asserts --incremental-bmc --unlimited-k-steps
[...]
real	0m8,706s
user	0m0,023s
sys	0m0,098s
```

fbrausse (approved):
LGTM.

lucasccordeiro (approved):

rafaelsamenezes (approved):

src/clang-c-frontend/clang_c_adjust_expr.cpp
shmarovfedor: It used to but not anymore.

src/clang-c-frontend/expr2ccode.cpp
shmarovfedor: Done

lucasccordeiro:
Thanks for submitting this PR, @shmarovfedor.

PR #843, opened 2022-08-28
[cmake] added comments

kunjsong01:
Added comments in cmake configure files to better explain the build process for C operational models. 

(A corresponding wiki page has also been created: https://github.com/esbmc/esbmc/wiki/Understanding-C-operational-models )


src/c2goto/CMakeLists.txt
```diff
@@ -38,19 +38,21 @@ endif()
...
   file(GLOB c2goto_libm_files CONFIGURE_DEPENDS
     "${CMAKE_CURRENT_SOURCE_DIR}/library/libm/*.c")
   file(GLOB_RECURSE c2goto_header_files CONFIGURE_DEPENDS
     "${CMAKE_CURRENT_SOURCE_DIR}/headers/*.h")
+
+  # For each config, generate clib*.c and clib*.goto file in the current binary directory
   foreach(in_f ${ARGN})
     set (in_f "${${in_f}}")
-    list(GET in_f 0 in_file)
+    list(GET in_f 0 out_file)
```
rafaelsamenezes, line 53: Isn't this going to break relative paths? We should at least add an assertion that only absolute paths are being used in the list.

scripts/flail.py
```diff
@@ -12,60 +12,79 @@
...
 
         Example:
 
         Step 1: Object Dump
-
             #ifndef __ESBMC_HEADERS_STDARG_H_
             #define __ESBMC_HEADERS_STDARG_H_
             ...
-                    vvv
+            is represented by the code as follows:
```
rafaelsamenezes: This is a hex object, not a code.

scripts/flail.py
```diff
@@ -136,6 +155,25 @@ def run(self, output_file, header = None, macro : str = None):
...
+                -o ./libc.c \
+                --header ./libc.h \
+                esbmc/src/c2goto/library/builtin_libs.c src/c2goto/library/ctype.c
+
+    Ex.2 Convert a goto binary file (see last line of the CMD below) into an unsigned int arrays
+         After building ESBMC, find clib32.goto from the build tree
+         CMD:
+            python3 flail.py -o ./clib32.c \
+                    esbmc/build/src/c2goto/clibd32.goto
+    '''
```
rafaelsamenezes: I suggest either move this as a doc for the module itself or add it as as `help` options

rafaelsamenezes (changes requested):
Some minor comments.

About the Wiki page:
- There are some typos on the word "operational/operaitonal"
- Could you add a legend in the diagram?

scripts/flail.py
```diff
@@ -12,60 +12,79 @@
 
 class Flail:
     """
-        The flaing consists in converting a file to a C header containing all the contents
-        as an uint array
+        The input file can be either an object file (e.g. a goto binary file)
+        or a list of C source file. The flailng converts the content of an input file
+        into an uint array. The resulting array will be written in a C header.
```
fbrausse: This comment is misleading: The input to this script can be file(s) of any type. There is no interpretation happening, "list of C source file" and "goto binary file" are misleading.

What `flail.py` does (in a quite obscure way via external tools, Python could actually do it itself - maybe slower, don't know) is to generate either (depending on a command-line option) a single C source file (and possibly also a header for it) containing one symbol for each input file defining the binary content of the file (as `char` array) and one symbol defining its length (as `unsigned int`), or one corresponding C source file (possibly including header) for each input file.

src/c2goto/CMakeLists.txt
```diff
@@ -59,16 +61,21 @@ function(mangle_clib output)
...
-      COMMAND ${CMD}
+      COMMAND ${run_c2goto}
       DEPENDS c2goto ${c2goto_library_files} ${c2goto_libm_files} ${c2goto_header_files}
       WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
       COMMENT "Generating libc model ${out_goto}"
       VERBATIM
       )
 
+    # Convert the content of goto binary file (e.g. clib32.goto) into an unsigned int array at CMake build time
+    # The resulting array will be written in the correpsonding C file (e.g. clib32.c) in the current binary directory.
```
fbrausse: "into an unsigned int array" is misleading (since it's actually of type `char`): maybe just write "into a C source file containing the content of the binary".

scripts/flail.py
```diff
@@ -12,60 +12,79 @@
...
-                    vvv
+            ...
+            output:
+            ...
             35,105,102,110,100,101,102,32,95,95,69,83,66,77,67,95,
             72,69,65,68,69,82,83,95,83,84,68,65,82,71,95,72,
             95,13,10,35,100,101,102,105,110,101,32,95,95,69,83,66,
 
-        Step 6: Create a C header
+        Step 6: Create a C header containing the output of Step 5 as an uint array
```
fbrausse: It's a `char` array.

fbrausse (commented):
In addition to Rafael's suggestions, I have suggestions how to avoid some misleading info in the new comments.

scripts/flail.py
rafaelsamenezes: The `flail.py` is just a translation of an older script (`flail.sh`). I am actually trying to make it python only as one of the dependencies keeps causing problems in the Windows Build.

> a single C source file (and possibly also a header for it) containing one symbol for each input file defining the binary content of the file (as char array)

You mean accepting a list of files as inputs and then generating an amalgamated version?

fbrausse: > The `flail.py` is just a translation of an older script (`flail.sh`).

I suspected that much. It's good to get rid of using these external tools.

> You mean accepting a list of files as inputs and then generating an amalgamated version?

Yes, I added support for multiple inputs some time back in cb442ce6d14a6f3da08cd7165a0be5a6bdf3f8ba. It's used to bundle the c2goto/{library,header} files for use in <https://github.com/esbmc/esbmc/blob/005ad337a051760ce22ddaaa618620b3794e13c7/src/c2goto/cprover_library.cpp#L151>

fbrausse:
Looking at the current help message of `flail.py`, it seems the `--macro` parameter is documented somewhat wrongly. It should read
> --macro MACRO is only meaningful in combination with --header; if specified, the header file will contain invocations of MACRO(body, size, fpath) for each input file where body is the name of the extern char[] symbol defining the binary content of the input file, size is its size and fpath is the path to the input file as passed to this script. If MACRO is not defined, the header will define it before the invocations to declare the body and size symbols as extern to facilitate usage of the header like this in a C file:
```c
#undef MACRO /* no definition of MACRO */
#include <header.h> /* declare symbols */
#define MACRO(body, size, ...) binary_handler(body, size, #__VA_ARGS__); /* last param is the binary's path */
#include <header.h> /* calls binary_handler() for each input */
#undef MACRO
```

scripts/flail.py
kunjsong01: Aren't they the ASCII encoding of "#ifndef __ESBMC_HEADERS_STDARG_H_"?

src/c2goto/CMakeLists.txt
kunjsong01: >Isn't this going to break relative paths?

No, it won't break the path. 

The goal here is to create `clib32/64.c` in `build/src/c2goto/`. The `out_file` variable is pointing to `clib32/64.c` and the current binary directory is already `build/src/c2goto/`. I don't think we need the relative path or assertion for absolute path at all in the first place.

scripts/flail.py
kunjsong01: Done

scripts/flail.py
kunjsong01: Fixed.

scripts/flail.py
kunjsong01: Done

src/c2goto/CMakeLists.txt
kunjsong01: Fixed.

kunjsong01:
> Looking at the current help message of `flail.py`, it seems the `--macro` parameter is documented somewhat wrongly. It should read
> 
> > --macro MACRO is only meaningful in combination with --header; if specified, the header file will contain invocations of MACRO(body, size, fpath) for each input file where body is the name of the extern char[] symbol defining the binary content of the input file, size is its size and fpath is the path to the input file as passed to this script. If MACRO is not defined, the header will define it before the invocations to declare the body and size symbols as extern to facilitate usage of the header like this in a C file:
> 
> ```c
> #undef MACRO /* no definition of MACRO */
> #include <header.h> /* declare symbols */
> #define MACRO(body, size, ...) binary_handler(body, size, #__VA_ARGS__); /* last param is the binary's path */
> #include <header.h> /* calls binary_handler() for each input */
> #undef MACRO
> ```

Added in flail.py

fbrausse:
> Added in flail.py

Thank you!

With C23 (hopefully), we can replace `flail.py` with `#embed`: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2499.pdf>

src/c2goto/CMakeLists.txt
```diff
@@ -38,19 +38,21 @@ endif()
...
+  # For each config, generate clib*.c and clib*.goto file in the current binary directory
   foreach(in_f ${ARGN})
     set (in_f "${${in_f}}")
-    list(GET in_f 0 in_file)
+    list(GET in_f 0 out_file)
     list(SUBLIST in_f 1 -1 in_flags)
-    file(RELATIVE_PATH out_file ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/${in_file})
-    string(REGEX REPLACE .c .goto out_goto "${in_file}")
-    string(REGEX REPLACE .c "" barename "${in_file}")
+    string(REGEX REPLACE .c .goto out_goto "${out_file}")
```
fbrausse: Could we use `string(REPLACE .c .goto [...])` here instead of `REGEX REPLACE` since the dot `.` has special regex meaning?

fbrausse (commented):
LGTM.

fbrausse (approved):

lucasccordeiro (approved):

src/c2goto/CMakeLists.txt
kunjsong01: Done

fbrausse:
I've updated the wiki page, though I'm still struggling somewhat to understand the diagram, there's so much going on. :eyes:

kunjsong01:
> I've updated the wiki page, though I'm still struggling somewhat to understand the diagram, there's so much going on. 👀

Thanks! I'll see what I can do to make the diagram more intuitive.

PR #846, opened 2022-09-02
Fix returning instance of `void`

### LINKED ISSUE #783: return free();
ESBMC does not accept this C program:
```c
#include <stdlib.h>

void foo(void *p)
{
	return free(p);
}

int main()
{
	foo(malloc(5));
}
```
The error message is confusing to me:
```
Generating GOTO Program
file free.c line 5 function foo: free is expected not to have LHS
terminate called after throwing an instance of 'int'
Aborted
```

fbrausse:
This PR:
- fixes #783 by removing call-site checks (for LHS or number of parameters) from builtins on the GOTO level,
- uses `log_status()` instead of `log_debug()` for `dump()` in order to ease debugging, and
- removes unused file/line parameters from logging.


mikhailramalho (approved):
LGTM.

We should probably check the abort in the frontend/symex too, there are probably some weird ones there still.

lucasccordeiro (approved):

lucasccordeiro:
@rafaelsamenezes: Do you have any idea of why the CI building is failing?

rafaelsamenezes:
> Do you have any idea of why the CI building is failing?

It seems that it failed while downloading boost through vcpkg (which should never fail). Hopefully it will work by restarting the windows build (just did it).

fbrausse:
> We should probably check the abort in the frontend/symex too, there are probably some weird ones there still.

From my quick look through goto-symex/builtin_functions.cpp there are no explicit aborts, just some assertions about the signature/call site, e.g. https://github.com/esbmc/esbmc/blob/f6ee4920d4ada93a193b001dab4b3bc45d8bbfca/src/goto-symex/builtin_functions.cpp#L1011 https://github.com/esbmc/esbmc/blob/f6ee4920d4ada93a193b001dab4b3bc45d8bbfca/src/goto-symex/builtin_functions.cpp#L737

PR #849, opened 2022-09-12
[compiler] added a new header to center compiler definitions

rafaelsamenezes:
This will add a new header to put our OS/Compiler specific definitions. I mainly did this because it was causing an unneeded dependency between `config.h` and `irep2.h` which is causing difficulties while implementing comments from https://github.com/esbmc/esbmc/pull/827

### DIFF
--- a/src/util/compiler_defs.h
+++ b/src/util/compiler_defs.h
@@ -1,0 +1,64 @@
+#pragma once
+
+/// Header to define some properties that are OS or Compiler specific
+
+#ifndef GNUC_PREREQ
+#if defined(__GNUC__) && defined(__GNUC_MINOR__) && defined(__GNUC_PATCHLEVEL__)
+#define GNUC_PREREQ(maj, min, patch)                                           \
+  ((__GNUC__ << 20) + (__GNUC_MINOR__ << 10) + __GNUC_PATCHLEVEL__ >=          \
+   ((maj) << 20) + ((min) << 10) + (patch))
+#elif defined(__GNUC__) && defined(__GNUC_MINOR__)
+#define GNUC_PREREQ(maj, min, patch)                                           \
+  ((__GNUC__ << 20) + (__GNUC_MINOR__ << 10) >= ((maj) << 20) + ((min) << 10))
+#else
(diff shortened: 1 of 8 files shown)


lucasccordeiro (approved):
LGTM.

src/util/compiler_defs.h
```diff
@@ -0,0 +1,64 @@
...
+#define DUMP_METHOD ATTRIBUTE_NOINLINE ATTRIBUTE_USED
+#else
+#define DUMP_METHOD ATTRIBUTE_NOINLINE
+#endif
+
+// Add your losing compiler to this list.
+#if !defined bool &&                                                           \
+  (defined __SUNPRO_CC && (__SUNPRO_CC < 0x500 || __SUNPRO_CC_COMPAT < 5) ||   \
+   defined __xlC__ || defined __DECCXX && __DECCXX_VER < 60000000 ||           \
+   defined _MSC_VER && _MSC_VER < 1100)
```
fbrausse: I doubt that any compiler which doesn't support `bool` can successfully compile ESBMC. Can we remove this?

src/util/compiler_defs.h
```diff
@@ -0,0 +1,64 @@
...
+#undef true
+#define bool int
+#define false 0
+#define true 1
+#endif
+
+// Minor optimization for gcc on some intel platforms.
+#if !defined _fast
+#if defined __GNUC__ && defined __i386__ && defined NDEBUG
+#define _fast __attribute__((__regparm__(3), __stdcall__))
```
fbrausse: I would also remove `_fast` and `_fasta` - unless we care about such ABI tricks to save a few instructions on 32-bit x86.

fbrausse (approved):
LGTM. Instead of moving, I would *re*move a bunch of these definitions, though.

src/util/compiler_defs.h
rafaelsamenezes: Removed it

src/util/compiler_defs.h
rafaelsamenezes: Removed it

fbrausse:
Many thanks, @rafaelsamenezes!

PR #850, opened 2022-09-13
SSA caching for sub-expressions

rafaelsamenezes:
This PR intend to add caching for matching and simplifying expressions that were already seen (in the same formula). For example, the program:

```c
int main() {
	int a,b,c; // nondet-value
	assert(a); // can't cache
	assert(a && b); // should change to 'b'
	assert(a || c); // should change to '1'
}
```

Produces the following formula:

```
ASSERT
execution_statet::\guard_exec?0!0 => !(a?1!0&0#0 == 0)
assertion

Thread 0 file cache.c line 4 function main
ASSERT
execution_statet::\guard_exec?0!0 => !(a?1!0&0#0 == 0) && !(b?1!0&0#0 == 0)
assertion

Thread 0 file cache.c line 5 function main
ASSERT
execution_statet::\guard_exec?0!0 => !(a?1!0&0#0 == 0) || !(c?1!0&0#0 == 0)
assertion

Thread 0 file cache.c line 6 function main
ASSUME
0
```

With this PR, the resulting formula will be:

```
ASSERT
execution_statet::\guard_exec?0!0 => !(a?1!0&0#0 == 0)
assertion

Thread 0 file cache.c line 4 function main
ASSERT
execution_statet::\guard_exec?0!0 => !(b?1!0&0#0 == 0)
assertion

Thread 0 file cache.c line 5 function main
ASSERT
1
assertion

Thread 0 file cache.c line 6 function main
ASSUME
0
```


lucasccordeiro (approved):
My only recommendation is to run this PR over the SV-COMP benchmarks before you merged it into the master.

regression/esbmc/cache_sub_true/main.c
```diff
@@ -0,0 +1,8 @@
+int main() {
+	int a,b,c; // nondet-value
+        __ESBMC_assume(a != 0);
+	assert(a); // can't cache
+        __ESBMC_assume(b != 0);
+	assert(a || b); // should change to 'b'
```
fbrausse, line 6: Shouldn't the assert be on `a && b` for the comment to match?

src/util/cache.cpp
```diff
@@ -12,7 +12,32 @@ void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
...
+         * the new condition */
+        log_debug("Simplified a sub expression!");
+        simplify(step.cond);
+        auto [itt, inss] = db.emplace(std::make_pair(step.cond, step.guard));
+        if(!inss)
+        {
+          log_debug("Cache hits: {}", ++hits);
+          step.cond = constant_bool2tc(trivial_value);
+        }
+      };
```
fbrausse, line 38: `;` is confusing here :)

src/util/cache.cpp
```diff
@@ -27,3 +52,35 @@ bool assertion_cache::run(symex_target_equationt::SSA_stepst &eq)
...
+  if(logic_op)
+  {
+    if(is_in_cache(logic_op->side_1, guards))
+    {
+      matched = true;
+      log_debug("Sub-Cache hits: {}", ++sub_hits);
+      logic_op->side_1 = constant_bool2tc(trivial_value);
+    }
+    else
+      try_matching_sub_expression(logic_op->side_1, guards);
```
fbrausse, line 74: Shouldn't `matched` be set to true if this recursive call matched a subexpr?

src/util/cache.cpp
```diff
@@ -27,3 +52,35 @@ bool assertion_cache::run(symex_target_equationt::SSA_stepst &eq)
...
+
+    if(is_in_cache(logic_op->side_2, guards))
+    {
+      matched = true;
+      log_debug("Sub-Cache hits: {}", ++sub_hits);
+      logic_op->side_2 = constant_bool2tc(trivial_value);
+    }
+    else
+      try_matching_sub_expression(logic_op->side_2, guards);
```
fbrausse, line 83: Same question as above: `matched |= try_[...]();`?

src/util/cache.cpp
```diff
@@ -12,7 +12,32 @@ void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
...
     log_debug("Cache missed");
+    if(match_subexpressions && step.cond->expr_id == expr2t::implies_id)
+    {
+      /* Assertions are in the form of step.guard => (exec_guard => expr)
+       * meaning that we can cache the 'expr' as long as we don't touch 'exec_guard'
+       */
+      auto logic_op = std::dynamic_pointer_cast<logic_2ops>(step.cond);
+      // First, add current expr into DB
+      db.emplace(std::make_pair(logic_op->side_2, step.guard));
+      // Now, check whether any of expr operands
```
fbrausse, line 25: The sentence in this comment is not complete. Check what exactly? :)

src/util/cache.cpp
```diff
@@ -12,7 +12,32 @@ void assertion_cache::run_on_assert(symex_target_equationt::SSA_stept &step)
...
+      if(try_matching_sub_expression(logic_op->side_2, step.guard))
```
fbrausse, line 26: I assume `logic_op->side_2` refers to `(exec_guard => expr)` in the comment above, yes? If so, how does passing this implication to the algorithm performing the substitutions ensure
> that we can cache the 'expr' as long as we **don't touch 'exec_guard'**

as per the comment above (emphasis mine)? try_matching_sub_expression() could touch exec_guard, can it not?

fbrausse (commented):
Looks like a nice extension of the previous "caching", though I'm having trouble understanding a few things in this implementation.

PR #851, opened 2022-09-14
Green Reordering algorithm

rafaelsamenezes:
This is the first algorithm for the canonization (normalization and renaming are missing for now) process of expr2t, which is based on this [paper](https://dl.acm.org/doi/abs/10.1145/2393596.2393665). The idea is that given an arithmetic expression, e.g `e + a + 42 + b = b + a` will convert it into `a + b + e + 42 = b + a`.

In future this will be used for the caching.

### DIFF
--- a/src/util/green_reordering.cpp
+++ b/src/util/green_reordering.cpp
@@ -1,0 +1,217 @@
+#include <util/green_reordering.h>
+#include <array>
+
+bool expr_variable_reordering::run(expr2tc &e)
+{
+  switch(e->expr_id)
+  {
+    // BINARY OPS
+  case expr2t::expr_ids::add_id:
+  case expr2t::expr_ids::mul_id:
+    run_on_binop(e);
+    break;
+    // RELATIONS
+  case expr2t::expr_ids::equality_id:
+  case expr2t::expr_ids::notequal_id:
+  case expr2t::expr_ids::lessthan_id:
+  case expr2t::expr_ids::greaterthan_id:
+  case expr2t::expr_ids::lessthanequal_id:
+  case expr2t::expr_ids::greaterthanequal_id:
+    run_on_relation(e);
+    break;
+    // Negations
+  case expr2t::expr_ids::neg_id:
+  case expr2t::expr_ids::not_id:
+    run_on_negation(e);
+    break;
+  default:
+    break; // don't care
+  }
+  return true;
+}
+
+void expr_variable_reordering::run_on_binop(expr2tc &expr)
+{
+  std::shared_ptr<arith_2ops> arith;
+  arith = std::dynamic_pointer_cast<arith_2ops>(expr);
+  // If arith is false then the cast failed
+  assert(arith);
+
+  /** Reorder both sides recursively
+   *
+   * The idea of this algorithm is to parse all symbols and values of a
+   * expression with the same operator and rewrite it.
+   *
+   * 1. Create a list of symbols and values
+   * 2. Order it in crescent order
+   * 3. Change all values
+   */
+
+  run(arith->side_1);
+  run(arith->side_2);
+
+  symbols_vec symbols; // annotates all inner symbols
+  values_vec values;   // annotates all inner values
+
+  // 1. Create a list of symbols and values
+  this->transverse_read_binop(arith, symbols, values);
+
+  // 2. Order it in crescent order
+  sort(symbols.begin(), symbols.end(), [](const auto &lhs, const auto &rhs) {
+    return lhs->get_symbol_name() > rhs->get_symbol_name();
+  });
+
+  // A constant propagation should have been executed prior to this
+  assert(values.size() <= 1);
+
+  // 3. Change all values
+  this->transverse_replace_binop(arith, symbols, values);
+}
+
+void expr_variable_reordering::run_on_negation(expr2tc &expr)
+{
+  if(expr->expr_id == expr2t::expr_ids::neg_id)
+  {
+    std::shared_ptr<arith_1op> arith;
+    arith = std::dynamic_pointer_cast<arith_1op>(expr);
+    // If arith is false then the cast failed
+    assert(arith);
+    run(arith->value);
+  }
+  else if(expr->expr_id == expr2t::expr_ids::not_id)
+  {
+    std::shared_ptr<bool_1op> arith;
+    arith = std::dynamic_pointer_cast<bool_1op>(expr);
+    // If arith is false then the cast failed
+    assert(arith);
+    run(arith->value);
+  }
+}
+
+void expr_variable_reordering::run_on_relation(expr2tc &expr)
+{
+  std::shared_ptr<relation_data> relation;
+  relation = std::dynamic_pointer_cast<relation_data>(expr);
+
+  // If relation is false then the cast failed
+  assert(relation);
+
+  // Reorder both sides
+  run(relation->side_1);
+  run(relation->side_2);
+}
+
+inline void expr_variable_reordering::transverse_read_binop(
+  const std::shared_ptr<arith_2ops> op,
+  expr_variable_reordering::symbols_vec &symbols,
+  expr_variable_reordering::values_vec &values)
+{
+  transverse_binop(op, symbols, values, TRANSVERSE_MODE::READ);
+}
+
+inline void expr_variable_reordering::transverse_replace_binop(
+  const std::shared_ptr<arith_2ops> op,
+  expr_variable_reordering::symbols_vec &symbols,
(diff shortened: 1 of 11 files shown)


src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+  case expr2t::expr_ids::greaterthanequal_id:
+    run_on_relation(e);
+    break;
+    // Negations
+  case expr2t::expr_ids::neg_id:
+  case expr2t::expr_ids::not_id:
+    run_on_negation(e);
+    break;
+  default:
+    break; // don't care
```
lucasccordeiro, line 28: Shall we print some message here?

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+  return true;
+}
+
+void expr_variable_reordering::run_on_binop(expr2tc &expr)
+{
+  std::shared_ptr<arith_2ops> arith;
+  arith = std::dynamic_pointer_cast<arith_2ops>(expr);
+  // If arith is false then the cast failed
+  assert(arith);
```
lucasccordeiro, line 38: Shall we print a message here?

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+  // 1. Create a list of symbols and values
+  this->transverse_read_binop(arith, symbols, values);
+
+  // 2. Order it in crescent order
+  sort(symbols.begin(), symbols.end(), [](const auto &lhs, const auto &rhs) {
+    return lhs->get_symbol_name() > rhs->get_symbol_name();
+  });
+
+  // A constant propagation should have been executed prior to this
+  assert(values.size() <= 1);
```
lucasccordeiro, line 65: Shall we add a message here?

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  // 3. Change all values
+  this->transverse_replace_binop(arith, symbols, values);
+}
+
+void expr_variable_reordering::run_on_negation(expr2tc &expr)
+{
+  if(expr->expr_id == expr2t::expr_ids::neg_id)
```
lucasccordeiro, line 73: Add some comments.

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+  {
+    std::shared_ptr<arith_1op> arith;
+    arith = std::dynamic_pointer_cast<arith_1op>(expr);
+    // If arith is false then the cast failed
+    assert(arith);
```
lucasccordeiro, line 78: Shall we add a message here?

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+    run(arith->value);
+  }
+  else if(expr->expr_id == expr2t::expr_ids::not_id)
+  {
+    std::shared_ptr<bool_1op> arith;
+    arith = std::dynamic_pointer_cast<bool_1op>(expr);
+    // If arith is false then the cast failed
+    assert(arith);
```
lucasccordeiro, line 86: Shall we add a message here?

lucasccordeiro (changes requested):

src/util/green_reordering.cpp
mikhailramalho, line 73: replace with `is_neg2t(expr)`

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  // 3. Change all values
+  this->transverse_replace_binop(arith, symbols, values);
+}
+
+void expr_variable_reordering::run_on_negation(expr2tc &expr)
+{
+  if(expr->expr_id == expr2t::expr_ids::neg_id)
+  {
+    std::shared_ptr<arith_1op> arith;
```
mikhailramalho, line 75: why are you using the `std::shared_ptr<arith_1op>` directly? Can you replace this and all the similar objects with `expr2tc`?

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+    arith = std::dynamic_pointer_cast<arith_1op>(expr);
+    // If arith is false then the cast failed
+    assert(arith);
+    run(arith->value);
+  }
+  else if(expr->expr_id == expr2t::expr_ids::not_id)
```
mikhailramalho, line 81: replace with `is_not2tc(expr)`

mikhailramalho (changes requested):
Nice! This is similar to what I was trying to do in the simplify branch

My comments are only related to the use of irep2 in a verbose way

src/util/green_reordering.h
```diff
@@ -0,0 +1,179 @@
...
+   * @param symbols
+   * @param values
+   */
+  static void replace_value(
+    const std::shared_ptr<arith_2ops> op,
+    bool is_lhs,
+    symbols_vec &symbols,
+    values_vec &values);
+
+  enum class TRANSVERSE_MODE
```
fbrausse, line 110: Typo: "traverse" instead of "transverse"?

src/util/green_reordering.h
```diff
@@ -0,0 +1,179 @@
...
+    TRANSVERSE_MODE mode,
+    bool is_lhs);
+
+  /**
+   * Default strategy to execute an algorithm while transversing the expr
+   * @param op binary operation
+   * @param symbols vector of all symbols expressions
+   * @param values vector of all values expressions
+   */
+  void transverse_binop(
```
fbrausse, line 138: Typo. Other members below, too.

unit/irep2/cache_fuzz_utils.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <climits>
+namespace
+{
+inline char convert_char_to_letter(const char input)
+{
+  const char a = 'a';
+  const char z = 'z';
+  if(input >= a && input <= z)
+    return input;
+  const char range = z - a;
```
fbrausse, line 11: `z-a+1` as you appearently want to include `'z'`. Also, you could just make this function `static` and avoid all the anon namespace syntax.

unit/irep2/cache_fuzz_utils.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Helper class to be used in fuzzing targets,
+ * this will help create valid expr to validate the algorithms
+ *
+ * The main class will work by receiving a string and parsing it as follows:
+ *
+ * RELATION BINOP [SYMBOLS] % SYMBOLS
+ *
+ * 00abc%a => a + b + c == a => (EQUAL (ADD (A) (ADD) (B) (C))) (A))
```
fbrausse, line 15: This looks pretty weird to me. If you want to avoid precedence via brackets, there is [polish notation](https://en.wikipedia.org/wiki/Polish_notation) for that. It seems like you used a bit of that, but not quite. Also I don't understand what `% SYMBOLS` means. Please clarify the syntax of this string.

src/util/green_reordering.cpp
```diff
@@ -0,0 +1,217 @@
...
+  run(arith->side_1);
+  run(arith->side_2);
+
+  symbols_vec symbols; // annotates all inner symbols
+  values_vec values;   // annotates all inner values
+
+  // 1. Create a list of symbols and values
+  this->transverse_read_binop(arith, symbols, values);
+
+  // 2. Order it in crescent order
```
fbrausse, line 59: What do you mean by "crescent order"? To me it looks like descending order.

src/util/green_reordering.h
```diff
@@ -0,0 +1,179 @@
...
+ *
+ *  Supported Operators: [+, *]
+ *  Supported Types (constants/symbols): [Signed BV, Unsigned BV]
+ *  Supported Relations: [==, !=, <, >, <=, >=]
+ *
+ */
+class expr_variable_reordering : public expr2tc_algorithm
+{
+  typedef std::vector<symbol2tc> symbols_vec;
+  typedef std::vector<constant_int2tc> values_vec;
```
fbrausse, line 25: Please refrain from introducing unnecessary typedefs. It is hard to reason about code when types are not clear. The fact that you repeat these two types a few times, and every time together, is a hint that maybe values of both should be paired in a separate type.

fbrausse (changes requested):
While I get the benefits of term normalization, I'm not sure that adding even more algorithms that modify existing expressions is a good idea. @rafaelsamenezes Why did you choose this design instead of creating a new expression as a result of this transformation while keeping the source expression unchanged?

PR #852, opened 2022-09-14
Goto contractor: support for multiple operators.

Mohannad-Aldughaim:
Hello everyone,

This is an update to the --goto-contractor option. This update includes support for operators &&, ||, ==, !=, and multiple asserts. Regression tests show the new features. Instead of creating one constraint and one contractor, the algorithm will create multiple contractors and save their location in the program. This will enable the algorithm to take the intersection and union of multiple properties and create their complement. The saved location helps get the properties to a given point in the program. 

I am happy to answer any questions you have.

Regards,
-Mohannad

### DIFF
--- a/src/goto-programs/goto_contractor.cpp
+++ b/src/goto-programs/goto_contractor.cpp
@@ -39,7 +39,7 @@ void goto_contractort::get_intervals(goto_functionst goto_functions)
 void goto_contractort::parse_intervals(expr2tc expr)
 {
   symbol2tc symbol;
-  BigInt value;
+  double value;
 
   expr = get_base_object(expr);
 
@@ -72,10 +72,11 @@ void goto_contractort::parse_intervals(expr2tc expr)
     neg = true;
     side2 = to_neg2t(side2).value;
   }
-  if(!is_constant_int2t(side2))
-    return;
 
-  value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
+  if(is_constant_int2t(side2))
+    value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
+  else
+    return;
 
   if(map.find(symbol->get_symbol_name()) == CspMap::NOT_FOUND)
     return;
@@ -83,15 +84,15 @@ void goto_contractort::parse_intervals(expr2tc expr)
   {
   case expr2t::expr_ids::greaterthan_id:
   case expr2t::expr_ids::greaterthanequal_id:
-    map.update_lb_interval(value.to_int64(), symbol->get_symbol_name());
+    map.update_lb_interval(value, symbol->get_symbol_name());
     break;
   case expr2t::expr_ids::lessthan_id:
   case expr2t::expr_ids::lessthanequal_id:
-    map.update_ub_interval(value.to_int64(), symbol->get_symbol_name());
+    map.update_ub_interval(value, symbol->get_symbol_name());
     break;
   case expr2t::expr_ids::equality_id:
-    map.update_lb_interval(value.to_int64(), symbol->get_symbol_name());
-    map.update_ub_interval(value.to_int64(), symbol->get_symbol_name());
+    map.update_lb_interval(value, symbol->get_symbol_name());
+    map.update_ub_interval(value, symbol->get_symbol_name());
     break;
   default:;
   }
@@ -129,11 +130,12 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
   else
     for(auto const &var : map.var_map)
     {
+      //testing with updating both bounds after reduction.
       symbol2tc X = var.second.getSymbol();
       if(var.second.isIntervalChanged())
       {
         auto lb =
-          create_value_expr(var.second.getInterval().lb(), double_type2());
+          create_value_expr(var.second.getInterval().lb(), int_type2());
         auto cond = create_greaterthanequal_relation(X, lb);
         goto_programt tmp_e;
         auto e = tmp_e.add_instruction(ASSUME);
@@ -143,7 +145,7 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
         goto_function.body.destructive_insert(loop_exit, tmp_e);
 
         auto ub =
-          create_value_expr(var.second.getInterval().ub(), double_type2());
+          create_value_expr(var.second.getInterval().ub(), int_type2());
         auto cond2 = create_lessthanequal_relation(X, ub);
         goto_programt tmp_e2;
         auto e2 = tmp_e2.add_instruction(ASSUME);
@@ -159,7 +161,9 @@ void goto_contractort::apply_contractor()
 {
   Contractor *contractor = contractors.get_contractors();
 
+  //Take the intersection of all contractors to perform Algorithm 2
   ibex::CtcFixPoint c_out(*contractor->get_outer());
+  //Take the union of all contractors with complement constraints to perform Algorithm 3
   ibex::CtcFixPoint c_in(*contractor->get_inner());
 
   std::ostringstream oss;
@@ -222,9 +226,10 @@ ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
     }
     case expr2t::expr_ids::not_id:
       //TODO: implement "not" flag  and process.
+      parse_error(base_object);
       break;
     default:
-      //unsupported. you shouldn't be here.
+      parse_error(base_object);
       break;
     }
   }
@@ -238,15 +243,13 @@ goto_contractort::create_constraint_from_expr2t(const expr2tc &expr)
 
   if(is_unsupported_operator(expr))
   {
-    std::ostringstream oss;
-    oss << get_expr_id(expr);
-    oss << " Expression is complex, skipping this assert.\n";
-    log_debug("{}", oss.str());
+    parse_error(expr);
     return nullptr;
   }
   if(is_not2t(expr))
   {
     //TODO: implement "not" flag  and process.
+    parse_error(expr);
     return nullptr;
   }
 
@@ -298,6 +301,7 @@ goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
 
   if(is_comp_expr(expr))
   {
+    parse_error(expr);
     return nullptr;
   }
   switch(expr->expr_id)
@@ -367,13 +371,6 @@ goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
     f = new ibex::Function(*vars, c);
     break;
   }
-  case expr2t::expr_ids::constant_floatbv_id:
-  {
-    const ibex::ExprConstant &c = ibex::ExprConstant::new_scalar(
-      to_constant_floatbv2t(expr).value.to_double());
-    f = new ibex::Function(*vars, c);
-    break;
-  }
   default:
     f = nullptr;
   }
@@ -394,6 +391,14 @@ int goto_contractort::create_variable_from_expr2t(irep_container<expr2t> expr)
     return index;
   }
   return map.NOT_FOUND;
+}
+
+void goto_contractort::parse_error(const expr2tc &expr)
+{
+  std::ostringstream oss;
+  oss << get_expr_id(expr);
+  oss << " Expression is complex, skipping this assert.\n";
+  log_debug("{}", oss.str());
 }
 
 bool goto_contractort::initialize_main_function_loops()
--- a/src/goto-programs/goto_contractor.cpp
+++ b/src/goto-programs/goto_contractor.cpp
@@ -73,10 +73,10 @@ void goto_contractort::parse_intervals(expr2tc expr)
     side2 = to_neg2t(side2).value;
   }
 
-  if(is_constant_int2t(side2))
-    value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
-  else
+  if(!is_constant_int2t(side2))
     return;
+
+  value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1) + (neg ? -1 : 1);
 
   if(map.find(symbol->get_symbol_name()) == CspMap::NOT_FOUND)
     return;
@@ -91,8 +91,6 @@ void goto_contractort::parse_intervals(expr2tc expr)
     map.update_ub_interval(value, symbol->get_symbol_name());
     break;
   case expr2t::expr_ids::equality_id:
-    map.update_lb_interval(value, symbol->get_symbol_name());
-    map.update_ub_interval(value, symbol->get_symbol_name());
     break;
   default:;
   }
@@ -134,15 +132,6 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
       symbol2tc X = var.second.getSymbol();
       if(var.second.isIntervalChanged())
       {
-        auto lb = create_value_expr(var.second.getInterval().lb(), int_type2());
-        auto cond = create_greaterthanequal_relation(X, lb);
-        goto_programt tmp_e;
-        auto e = tmp_e.add_instruction(ASSUME);
-        e->inductive_step_instruction = false;
-        e->guard = cond;
-        e->location = loop_exit->location;
-        goto_function.body.destructive_insert(loop_exit, tmp_e);
-
         auto ub = create_value_expr(var.second.getInterval().ub(), int_type2());
         auto cond2 = create_lessthanequal_relation(X, ub);
         goto_programt tmp_e2;
@@ -185,11 +174,13 @@ ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
   ibex::Ctc *contractor = nullptr;
   auto base_object = get_base_object(expr);
 
-  if(
-    (is_comp_expr(base_object) && !is_notequal2t(base_object)) ||
-    is_arith_expr(base_object))
-    contractor =
-      new ibex::CtcFwdBwd(*create_constraint_from_expr2t(base_object));
+  if(is_unsupported_operator_in_contractor(expr))
+  {
+    auto cons = create_constraint_from_expr2t(base_object);
+    if(cons == nullptr)
+      return nullptr;
+    contractor = new ibex::CtcFwdBwd(*cons);
+  }
   else
   {
     std::shared_ptr<logic_2ops> logic_op;
@@ -198,15 +189,19 @@ ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
     switch(base_object->expr_id)
     {
     case expr2t::expr_ids::and_id:
-      contractor = new ibex::CtcCompo(
-        *create_contractor_from_expr2t(logic_op->side_1),
-        *create_contractor_from_expr2t(logic_op->side_2));
+    {
+      auto side1 = create_contractor_from_expr2t(logic_op->side_1);
+      auto side2 = create_contractor_from_expr2t(logic_op->side_2);
+      if(side1 != nullptr && side2 != nullptr)
+        contractor = new ibex::CtcCompo(*side1, *side2);
       break;
+    }
     case expr2t::expr_ids::or_id:
     {
       auto side1 = create_contractor_from_expr2t(logic_op->side_1);
       auto side2 = create_contractor_from_expr2t(logic_op->side_2);
-      contractor = new ibex::CtcUnion(*side1, *side2);
+      if(side1 != nullptr && side2 != nullptr)
+        contractor = new ibex::CtcUnion(*side1, *side2);
       break;
     }
     case expr2t::expr_ids::notequal_id:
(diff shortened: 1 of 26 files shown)


src/goto-programs/goto_contractor.cpp
```diff
@@ -129,11 +132,21 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
       symbol2tc X = var.second.getSymbol();
       if(var.second.isIntervalChanged())
       {
-        //only update upperbound
-        auto ub = create_value_expr(var.second.getInterval().ub(), int_type2());
+        auto lb =
```
lucasccordeiro: Can you please add some comments?

src/goto-programs/goto_contractor.cpp
```diff
@@ -142,24 +155,18 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
...
 
-  std::ostringstream oss;
-  oss << "\t- Constraint:" << c;
-  oss << "\n\t- Function:" << c.f;
-
-  auto complement = get_complement(c.op);
-
-  ibex::NumConstraint c2(c.f, complement);
-  oss << "\n\t- Complement: " << c2;
+  ibex::CtcFixPoint c_out(*contractor->get_outer());
```
lucasccordeiro, line 152: Can you relate this to the operators we describe in our paper? For example, add some short sentences to explain the inner and outer contractor.

src/goto-programs/goto_contractor.cpp
```diff
@@ -142,24 +155,18 @@ void goto_contractort::insert_assume(goto_functionst goto_functions)
...
+  ibex::CtcFixPoint c_in(*contractor->get_inner());
 
-  ibex::CtcFwdBwd c_out(c);
-  ibex::CtcFwdBwd c_in(c2);
+  std::ostringstream oss;
 
   domains = map.create_interval_vector();
 
+  oss << "Contractors applied:";
```
lucasccordeiro, line 160: Shouldn't we use the fmt library interface to output messages?

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,37 +178,77 @@ void goto_contractort::contractor()
...
 
-ibex::CmpOp goto_contractort::get_complement(ibex::CmpOp op)
+ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
 {
-  switch(op)
+  ibex::Ctc *contractor = nullptr;
+  auto base_object = get_base_object(expr);
+
+  if(
+    (is_comp_expr(base_object) && !is_notequal2t(base_object)) ||
```
lucasccordeiro: This if-condition looks weird to me. Can you add some comments?

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,37 +178,77 @@ void goto_contractort::contractor()
...
+      ibex::Function *g = create_function_from_expr2t(rel->side_2);
+      auto *side1 = new ibex::NumConstraint(*vars, (*f)(*vars) > (*g)(*vars));
+      auto *side2 = new ibex::NumConstraint(*vars, (*f)(*vars) < (*g)(*vars));
+      auto *c_side1 = new ibex::CtcFwdBwd(*side1);
+      auto *c_side2 = new ibex::CtcFwdBwd(*side2);
+      contractor = new ibex::CtcUnion(*c_side1, *c_side2);
+      break;
+    }
+    case expr2t::expr_ids::not_id:
+      //TODO: implement "not" flag  and process.
```
lucasccordeiro, line 223: You could print a message here.

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,37 +178,77 @@ void goto_contractort::contractor()
...
+      break;
+    default:
+      //unsupported. you shouldn't be here.
```
lucasccordeiro: You could print a message here.

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,37 +178,77 @@ void goto_contractort::contractor()
...
+goto_contractort::create_constraint_from_expr2t(const expr2tc &expr)
 {
   ibex::NumConstraint *c;
+
   if(is_unsupported_operator(expr))
   {
     std::ostringstream oss;
     oss << get_expr_id(expr);
-    oss << "Expression is complex, skipping this assert.\n";
+    oss << " Expression is complex, skipping this assert.\n";
```
lucasccordeiro: Can you define the ``complex`` term?

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,37 +178,77 @@ void goto_contractort::contractor()
...
     log_debug("{}", oss.str());
     return nullptr;
   }
+  if(is_not2t(expr))
+  {
+    //TODO: implement "not" flag  and process.
```
lucasccordeiro, line 252: You could print a message here.

src/goto-programs/goto_contractor.cpp
```diff
@@ -307,6 +367,13 @@ goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
     f = new ibex::Function(*vars, c);
     break;
   }
+  case expr2t::expr_ids::constant_floatbv_id:
```
lucasccordeiro: Do we really support float?

lucasccordeiro (changes requested):

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Previously, we only reduced the upper bound to avoid false negatives. I'm testing with both bounds this time and waiting on the results.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: I am using fmt. The `std::ostringstream oss;` is just to format the string.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: In Ibex, the expressions are structured more strictly than c. On the top level, we have logical operators such as && and ||. Then the comparison operators such as > and <. For example, we can have `(x > y || x < y)` and parse successfully. However, if the expression is valid in c `( x == y > 0 )`, it cannot be converted to ibex without reasoning. That's why we call them complex expressions.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: We don't yet. removed. I was trying to handle floats in this update, but I faced some problems with it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Added to the documentation of this function.

src/goto-programs/goto_contractor.h
```diff
@@ -41,6 +41,210 @@ class vart
...
+  unsigned int get_location() const
+  {
+    return location;
+  }
+
+private:
+  ibex::Ctc *get_complement_contractor(ibex::Ctc *c)
+  {
+    ibex::Ctc *complement_contractor = nullptr;
+    auto type = boost::core::demangle(typeid(*c).name());
```
mikhailramalho: doesn't ibex have a way of getting some kind of type id? IIRC, typeid.name() may not be portable.

src/goto-programs/goto_contractor.h
```diff
@@ -41,6 +41,210 @@ class vart
...
+        inner->add(*(ctc->get_inner()));
+      }
+
+    c->set_outer(new ibex::CtcCompo(*outer));
+    c->set_inner(new ibex::CtcUnion(*inner));
+    c->set_location(0);
+    return c;
+  }
+
+  Contractor *get_contractors(void)
```
mikhailramalho: no need for K&R style.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done

src/goto-programs/goto_contractor.h
```diff
@@ -153,23 +355,21 @@ class goto_contractort : public goto_functions_algorithm
...
-      log_status("1/4 - Parsing asserts to create CSP Constraints.");
-      get_constraints(_goto_functions);
-      if(constraint == nullptr)
+      log_debug("1/4 - Parsing asserts to create CSP Constraints.");
+      get_contractors(_goto_functions);
+      if(contractors.is_empty())
       {
-        log_status(
-          "Constraint expression not supported. Aborting goto-contractor");
+        log_status("Aborting goto-contractor, No Contractors were created.");
```
mikhailramalho: I suggest avoiding using  "abort" in status messages. Maybe esbmc can print something like "Contractors: No supported expression found".

src/goto-programs/goto_contractor.h
```diff
@@ -50,7 +254,7 @@ class vart
 class CspMap
 {
 public:
-  static constexpr int MAX_VAR = 10;
+  static constexpr int MAX_VAR = 26;
```
mikhailramalho, line 239: I think I've asked this before but do we really need this MAX_VAR constant?

src/goto-programs/goto_contractor.cpp
```diff
@@ -69,26 +72,27 @@ void goto_contractort::parse_intervals(expr2tc expr)
...
     side2 = to_neg2t(side2).value;
   }
-  if(!is_constant_int2t(side2))
-    return;
 
-  value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
+  if(is_constant_int2t(side2))
+    value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
+  else
+    return;
```
mikhailramalho, line 77: Can you revert this to the previous early return code?

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,35 +180,74 @@ void goto_contractort::contractor()
...
 }
 
-ibex::CmpOp goto_contractort::get_complement(ibex::CmpOp op)
+ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
 {
-  switch(op)
+  ibex::Ctc *contractor = nullptr;
+  auto base_object = get_base_object(expr);
+
+  if(
```
mikhailramalho: can you rewrite this as an early return? i.e, if it's none of the expect exprt, then return.

mikhailramalho: or better still: you can use the is_unsupported_exprt you created.

src/goto-programs/goto_contractor.cpp
```diff
@@ -316,13 +378,25 @@ goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
 
 int goto_contractort::create_variable_from_expr2t(irep_container<expr2t> expr)
```
mikhailramalho: use expr2tc instead of irep_container<expr2t>

src/goto-programs/goto_contractor.cpp
```diff
@@ -316,13 +378,25 @@ goto_contractort::create_function_from_expr2t(irep_container<expr2t> expr)
...
 {
-  std::string var_name = to_symbol2t(expr).get_symbol_name();
-  int index = map.find(var_name);
-  if(index == CspMap::NOT_FOUND)
+  if(is_symbol2t(expr))
   {
-    index = map.add_var(var_name, to_symbol2t(expr));
+    std::string var_name = to_symbol2t(expr).get_symbol_name();
+    int index = map.find(var_name);
```
mikhailramalho: you can probably just use add, instead of find + add.

mikhailramalho (changes requested):
Can you also check if the expressions are not leaking? I see a lot of `new` calls but no `delete`

lucasccordeiro:
@Mohannad-Aldughaim: Can I ask you for the status of this PR?

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: Unfortunately, that's the only way I found to determine the type in ibex.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: Not sure what you mean. is it the spacing between the functions?

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: removed

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Created a new one for contractors. Because some are supported on the contractor level but not on the constraint level.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Done

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done.

Mohannad-Aldughaim:
Many thanks for the comments and sorry for the delay.
In preparation for SV-Comp, May I ask you to review and merge this pull request? 

@mikhailramalho :
>  Can you also check if the expressions are not leaking? I see a lot of new calls but no delete.

It seems to me that ibex uses smart pointers. when I try to delete the object it raises an exception.

src/goto-programs/goto_contractor.h
mikhailramalho: it's the `function_name(void)`. Just leave the it empty, i.e., `function_name()`.

src/goto-programs/goto_contractor.cpp
```diff
@@ -69,26 +72,25 @@ void goto_contractort::parse_intervals(expr2tc expr)
     neg = true;
     side2 = to_neg2t(side2).value;
   }
+
   if(!is_constant_int2t(side2))
     return;
 
-  value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1);
+  value = to_constant_int2t(side2).as_long() * (neg ? -1 : 1) + (neg ? -1 : 1);
```
mikhailramalho, line 79: What does this do exactly?

src/goto-programs/goto_contractor.cpp
```diff
@@ -171,35 +169,89 @@ void goto_contractort::contractor()
...
-  case ibex::GT:
-    return ibex::LEQ;
-  case ibex::LEQ:
-    return ibex::GT;
-  case ibex::LT:
-    return ibex::GEQ;
-  default:
-    log_status("cant process equal");
-    break;
+    auto cons = create_constraint_from_expr2t(base_object);
```
mikhailramalho, line 179: is it's unsupported, shouldn't we just return nullptr here?

src/goto-programs/goto_contractor.cpp
```diff
@@ -225,28 +278,32 @@ goto_contractort::create_constraint_from_expr2t(expr2tc expr)
...
-bool goto_contractort::is_unsupported_operator(expr2tc expr)
+bool goto_contractort::is_unsupported_operator_in_constraint(
+  const expr2tc &expr)
 {
   expr2tc e = get_base_object(expr);
   return is_arith_expr(e) || is_constant_number(e) || is_symbol2t(e) ||
-         is_notequal2t(e) || is_equality2t(e) || is_not2t(e) ||
-         is_modulus2t(e) || is_or2t(e) || is_and2t(e);
+         is_notequal2t(e) || is_not2t(e) || is_modulus2t(e) || is_or2t(e) ||
+         is_and2t(e);
```
mikhailramalho, line 295: I think you've implemented the support for `and` and `notequal`, right?

src/goto-programs/goto_contractor.cpp
mikhailramalho: I meant in the function parameter

src/goto-programs/goto_contractor.h
mikhailramalho: I just had a look at the API and all these objects inherit from ctc, so you can use the dynamic_cast that you already implemented to convert them safety.

It will look something like: 
```
if(auto ctc_compo = dynamic_cast<ibex::CtcCompo *>(c)) {
      ibex::Array<ibex::Ctc> list_of_contractors;
      for(auto &it : ctc_compo->list)
        list_of_contractors.add(*get_complement_contractor(&it));
      return new ibex::CtcUnion(list_of_contractors);
} else if (auto ctc_union = dynamic_cast<ibex::CtcUnion *>(c)) {
      auto ctc_union = dynamic_cast<ibex::CtcUnion *>(c);
      ibex::Array<ibex::Ctc> list_of_contractors;
      for(auto &it : ctc_union->list)
        list_of_contractors.add(*get_complement_contractor(&it));
      return new ibex::CtcCompo(list_of_contractors);
} // etc

      std::ostringstream oss;
      oss << "Type: " << type;
      log_debug("{}", oss.str());

     return nullptr;
```

Note that you don't need `complement_contractor` at all here.

src/goto-programs/goto_contractor.h
mikhailramalho: You removed it but used the magic number in the code directly. Please, use the MAX_VAR constant (or any other constant name FWIW) instead of magic numbers.

My question is why do we need to limit the max number of constants, doesn't the code work with more than 26 variables?

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Correct. I changed the function name. sorry for the confusion.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: I changed the name of the function to show its purpose. Because if we encounter these operators we will create constraints, not contractors.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: When I use the `--interval-analysis` option it will insert assumes into the code. These assumes do not hold in some cases because of a difference of 1. For example:
```
int x=0;
int y=0;
while( y < 1000 )
{
   assume(y<=999) //inserted by interval analysis option
   x=x+y;
   y++;
}
assert(x >=y)
```
Here the interval analysis option says the upper bound is 999. However, when we reach the assert it will be 1000. that's why I added/subtracted 1 in that statement. It improved the results for some benchmarks.

I am currently working on a script that will extract accurate intervals using Frama-C.   This is just temporary.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: done. My bad I thought I changed it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: I made a new function that does the same as the condition of the `if` statement. its called `is_constraint_operator()`

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: I agree with you. It should not be arbitrary. I am still working on this. Because I can't change the size of the vector, I need to recreate it each time we find a new variable. The ibex documentation did not state any limits on the number of variables in a vector.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: Done. What I had in mind when writing it was something like java's `instanceof`. Thank you!

lucasccordeiro (approved):

src/goto-programs/goto_contractor.cpp
mikhailramalho: I see... The interval analysis is missing the loop termination condition that it's added but the k-induction transformations.

You can probably move that from the k-induction algorithm to a separate transformation step in a separate PR, so you can remove this code. WDYT @lucasccordeiro?

src/goto-programs/goto_contractor.h
```diff
@@ -41,6 +41,198 @@ class vart
...
+      oss << "inner :" << to_oss(c->get_inner()).str() << "\n";
+      oss << "location :" << c->get_location() << "\n";
+    }
+    log_debug("{}", oss.str());
+  }
+  std::ostringstream to_oss(ibex::Ctc *c)
+  {
+    std::ostringstream oss;
+    auto type = boost::core::demangle(typeid(*c).name());
+    if(type == "ibex::CtcCompo" || type == "ibex::CtcUnion")
```
mikhailramalho: you don't need the typeid here, just use dynamic_cast and check if the type != nullptr

mikhailramalho (approved):
Thanks for addressing my comments.

Please just remove the remaining typeid and this should be good to go

src/goto-programs/goto_contractor.cpp
lucasccordeiro: Yes, I agree that it could be a separate PR.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done

PR #853, opened 2022-10-07
Multi property Verification

rafaelsamenezes:
This adds a naive support for verifying all  the claims of a given bound. Basically it sequentially checks for one assertion at a time. Note that, ESBMC's output gets a mess. So in future we should work on making it less verbose.


src/goto-symex/slice.h
```diff
@@ -36,6 +36,27 @@ class simple_slice : public slicer
...
+ * Claim slicer: remove every other claim of the formula
+ * with the exception of the one to keep.
+ */
+class claim_slicer : public slicer
+{
+public:
+  explicit claim_slicer(const size_t claim_to_keep)
+    : claim_to_keep(claim_to_keep)
+  {
+    if(claim_to_keep == 0)
```
lucasccordeiro: Shall we replace it with `!claim_to_keep`?

src/esbmc/bmc.cpp
```diff
@@ -288,6 +289,30 @@ void bmct::report_trace(
   }
 }
 
+void bmct::report_multi_property_trace(
+  smt_convt::resultt &res,
+  const std::string &msg)
+{
+  assert(options.get_bool_option("base-case"));
```
lucasccordeiro, line 306: Why does it work only for the base case?

src/esbmc/bmc.cpp
```diff
@@ -658,3 +688,58 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
+  // As of now, it only makes sense to do this for the base-case
+  assert(
+    options.get_bool_option("base-case") &&
+    "Multi-property only supports base-case");
+  smt_convt::resultt final_result;
+
+  /* TODO: For coverage, We should store and remove all the claims that
+   *       were already verified
+   */
+  for(size_t i = 1; i <= remaining_claims; i++)
```
lucasccordeiro: don't we have a way to check all assertions at once with the SMT solvers?

lucasccordeiro (changes requested):
It would be nice to have a follow-up PR with a better UI.

src/esbmc/bmc.cpp
rafaelsamenezes: You mean using solvers API or running all of them in parallel? Either way, it should be possible. I would suggest this going in another PR though.

lucasccordeiro: Yes, I believe you should be able to do it via API.

I think this current solution is not good.

lucasccordeiro: @fbrausse and @mikhailramalho: what do you think?

src/esbmc/bmc.cpp
rafaelsamenezes: There is nothing preventing other modes. I just didn't implement the messages for them.

As of now, It didn't make sense for me. Maybe we could somehow reuse the unwinding assertions that were proven, that's beyond this PR though.

src/esbmc/bmc.cpp
```diff
@@ -658,3 +688,58 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
+        options.get_bool_option("no-slice") &&
+        !options.get_bool_option("compact-trace"))
+        is_compact_trace = false;
+
+      goto_tracet goto_trace;
+      build_goto_trace(eq, runtime_solver, goto_trace, is_compact_trace);
+      // TODO: Replace this with a test-case for coverage!
+      std::string output_file = options.get_option("cex-output");
+
+      if(output_file != "")
```
mikhailramalho: isEmpty()?

src/esbmc/bmc.cpp
```diff
@@ -637,7 +662,12 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
         std::shared_ptr<smt_convt>(create_solver("", ns, options));
     }
 
-    return run_decision_procedure(runtime_solver, eq);
+    if(
+      options.get_bool_option("multi-property") &&
+      options.get_bool_option("base-case"))
+      return multi_property_check(eq, result->remaining_claims);
+    else
```
mikhailramalho, line 679: No need for the else here

src/goto-symex/slice.cpp
```diff
@@ -141,3 +141,37 @@ bool simple_slice::run(symex_target_equationt::SSA_stepst &steps)
...
+  for(symex_target_equationt::SSA_stepst::iterator it = steps.begin();
+      it != steps.end();
+      it++)
+  {
+    // just find the next assertion
+    if(it->is_assert())
+    {
+      if(
+        counter++ ==
+        claim_to_keep) // this is the assertion that we should not skip!
```
mikhailramalho, line 158: move the comment up or down to improve code readability?

mikhailramalho (approved):
Oh, that's really cool! Only some nitpick about the code, otherwise, it LGTM.

Some questions/comments:

1. Do the logs overlap or do we just print several logs/cexs? I wonder if we should use the osyncstream to prevent overlapping.

2. I'd love to see some learning regarding unsat claims in a future patch!

3. Do we have some number from verifying unsat problems in SV-COMP?

src/esbmc/bmc.cpp
mikhailramalho: You can check the code in `symex_target_equationt::convert` to convert all the assertions at once, it should look something like:
```
    smt_conv.assert_ast(
      smt_conv.make_n_ary(&smt_conv, &smt_convt::mk_or, assertions));
```

src/esbmc/bmc.cpp
lucasccordeiro: Could you please add a TODO here?

src/esbmc/bmc.cpp
lucasccordeiro: Well spotted.

src/esbmc/bmc.cpp
lucasccordeiro: Given that @rafaelsamenezes has now parallelized the VCs check, I think we could consider @mikhailramalho's suggestions for a future PR.

src/esbmc/bmc.cpp
```diff
@@ -658,3 +694,70 @@ smt_convt::resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
...
+      {
+        std::ofstream out(fmt::format("{}-{}", output_file, ce_counter++));
+        show_goto_trace(out, ns, goto_trace);
+      }
+
+      final_result = result;
+    }
+  };
+
+  if(options.get_bool_option("parallel-solving"))
```
lucasccordeiro: Can you add some comments here?

lucasccordeiro (approved):
@rafaelsamenezes: could you please add some regression tests? Can you also run this solution over the SV-COMP benchmarks?

rafaelsamenezes:
This PR is no longer valid. See #1022

PR #858, opened 2022-11-04
Forward Condition Slicer

rafaelsamenezes:
This is a static slicer to be applied for the Forward Condition only. The main idea is to remove loops that do not contribute for the reachability of any assertions.  As this is added at goto program level, it will only work for explicit user assertions. Also, we can't simplify if there is any pointer unless I implement a points-to analysis.

This should make esbmc able to prove correctness in programs such as:

```c
int main() {
  while(1);
  assert(0);
}
```

The flow is the following:

1. Run an Abstract Interpretation over the program, using the dependencies for every variable at each state
2. For each function run a goto algorithm, first in backwards to store dependencies for IF, ASSERT, ASSUME, RETURN and FUNCTION_CALL
3. Same as 2, now forward... removing unecessary instructions
4. At each loop, remove if they are trivial (only contains skips).


mikhailramalho:
Interesting. Why not move it `esbmc_parseoptionst::process_goto_program` and do something like: 
`if(cmdline.isset("forward-condition") slicer.run(g);`, maybe before inlining?

There is a call to `remove_skip` on `esbmc_parseoptionst::process_goto_program`, so you don't have to.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1283,15 +1284,37 @@ int esbmc_parseoptionst::do_forward_condition(
 
   opts.set_option("no-unwinding-assertions", false);
   opts.set_option("partial-loops", false);
-
+  auto g = goto_functions;
+  if(opts.get_bool_option("forward-slicer"))
+  {
+    log_status("Trying to apply Forward Slicer...");
```
lucasccordeiro: Can you improve this message? "Try" is not an appropriate verb :-)

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1283,15 +1284,37 @@ int esbmc_parseoptionst::do_forward_condition(
...
+    namespacet ns(context);
+    forward_slicer slicer(ns);
+    try
+    {
+      slicer.run(g);
```
lucasccordeiro: Can you add comments to this block?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1283,15 +1284,37 @@ int esbmc_parseoptionst::do_forward_condition(
...
+      
+      std::ostringstream oss;
+      g.output(ns, oss);
+      log_debug("{}", oss.str());
+      log_status("Sliced");
+    }
+    catch(...)
+    {
+      g = goto_functions;
+      log_status("Unable to slice");
```
lucasccordeiro: Why? Can you give further explanation here?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1283,15 +1284,37 @@ int esbmc_parseoptionst::do_forward_condition(
...
+  }
   // We have to disable assertions in the forward condition but
   // restore the previous value after it
   bool no_assertions = opts.get_bool_option("no-assertions");
 
   // Turn assertions off
   opts.set_option("no-assertions", true);
 
-  bmct bmc(goto_functions, opts, context);
+  bmct bmc(g, opts, context);
```
lucasccordeiro: Shall we keep 'goto_functions'? It is better than just 'g'.

src/esbmc/options.cpp
```diff
@@ -149,6 +149,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"partial-loops", NULL, "permit paths with partial loops"},
     {"unroll-loops", NULL, ""},
     {"no-slice", NULL, "do not remove unused equations"},
+    {"forward-slicer", NULL, "attempt to slice loops that don't affect VCCs"},
```
lucasccordeiro, line 152: why "attempt"? we should use a better verb here. Something like:

`Slice loops with XXX that do not affect VCCs`

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
+#include <goto-programs/forward_slicer.h>
+#include <util/prefix.h>
+namespace
+{
+// TODO: Move this to irep2_utils
```
lucasccordeiro: Shall we move it to this PR?

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+void slicer_domaint::assign(const expr2tc &e)
+{
+  auto assignment = to_code_assign2t(e);
+
+  // We don't care about non-symbolic LHS
+  if(!is_symbol2t(assignment.target))
+    return;
+
+  // Hack: This should prevent the atexit handled (Needed as we don't support pointers)
```
lucasccordeiro: Hum... that's bad...  Don't we have a better solution here?

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+  if(has_prefix(to_symbol2t(assignment.target).get_symbol_name(), "c:stdlib.c"))
+     return;
+
+  // We do not support pointers
+  if(is_pointer_type(assignment.target->type))
```
lucasccordeiro: Shall we move this code to an earlier position in this method?

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+  {
+     log_debug("Forward slicer does not support pointers: {}", *e);
+     throw "TODO: points-to analysis";
```
lucasccordeiro: Shall we use our log function?

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+  }
+  
+  std::set<std::string> vars;
+  // Build-up dependencies over the RHS
+  get_symbols(assignment.source, vars);  
+  dependencies[to_symbol2t(assignment.target).get_symbol_name()] = vars;
+}
+
+void slicer_domaint::declaration(const expr2tc &e)
```
lucasccordeiro: Add some comments here.

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+  {
+    log_debug("Forward slicer does not support pointers: {}", *e);
+    throw "TODO: points-to analysis";
+  }
+
+  std::set<std::string> vars;
+  dependencies[A.value.as_string()] = vars;
+}
+
+void slicer_domaint::transform(
```
lucasccordeiro: Add some comments.

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  case ASSIGN:
+    assign(instruction.code);
+    break;
+
+  default:;
+  }
+}
+
+bool slicer_domaint::join(const slicer_domaint &b)
```
lucasccordeiro: Add some comments.

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+    for(auto &d : dependencies)
+      for(auto &item : d.second)
+        for(auto &item_2 : dependencies[item])
+          changed |= dependencies[d.first].insert(item_2).second;
+  }
+
+  return true;
+}
+
+bool forward_slicer::runOnProgram(goto_functionst &F)
```
lucasccordeiro: Add some comments.

src/goto-programs/forward_slicer.cpp
```diff
@@ -0,0 +1,246 @@
...
+  return false;
+}
+
+bool forward_slicer::should_skip_function(const std::string &func)
+{
+  // We should skip trying to slice our intrinsic functions
+  return has_prefix(func, "c:@F@__ESBMC_") || has_prefix(func, "c:@F@pthread_");
+}
+
+bool forward_slicer::runOnFunction(std::pair<const dstring, goto_functiont> &F)
```
lucasccordeiro: Add some comments.

lucasccordeiro (changes requested):
@rafaelsamenezes: could you also run this PR over the sv-comp benchmarks?

src/goto-programs/forward_slicer.cpp
rafaelsamenezes: These came from the default ait

src/goto-programs/forward_slicer.cpp
rafaelsamenezes: These came from the default ait

src/goto-programs/forward_slicer.cpp
rafaelsamenezes: These came from the default algorithmst

lucasccordeiro (approved):

mikhailramalho:

This PR has been open since 2023-04-30 with no activity. The forward slicer concept it introduces is tracked in issue #992 ("[slicer] Remove loops that do not contribute to checking a given property"), which references this PR. Closing the PR and deleting the branch — anyone reviving the work should start from a fresh implementation against current master (the slicer has moved on substantially since 2023).

PR #859, opened 2022-11-06
Atexit model implementation

mikhailramalho:
This PR implements the atexit model and changes ESBMC to always call it when defined by the standard.

The biggest change is that finish_formula now call assertion(), instead of creating the assertion itself. This is necessary so we can encode the memory cleanup check using the appropriate guards.

### DIFF
--- a/src/c2goto/library/stdlib.c
+++ b/src/c2goto/library/stdlib.c
@@ -26,9 +26,9 @@ __ESBMC_HIDE:;
   while(__ESBMC_atexits)
   {
     __ESBMC_atexits->atexit_func();
-    __ESBMC_atexit_key *tmp = __ESBMC_atexits->next;
+    __ESBMC_atexit_key *__ESBMC_tmp = __ESBMC_atexits->next;
     free(__ESBMC_atexits);
-    __ESBMC_atexits = tmp;
+    __ESBMC_atexits = __ESBMC_tmp;
   }
 }
 
--- a/src/c2goto/library/stdlib.c
+++ b/src/c2goto/library/stdlib.c
@@ -14,7 +14,7 @@ typedef struct atexit_key
   struct atexit_key *next;
 } __ESBMC_atexit_key;
 
-__ESBMC_atexit_key *__ESBMC_atexits = NULL;
+static __ESBMC_atexit_key *__ESBMC_atexits = NULL;
 
 void __atexit_handler()
 {
--- a/src/c2goto/library/stdlib.c
+++ b/src/c2goto/library/stdlib.c
@@ -16,7 +16,7 @@ typedef struct atexit_key
 
 static __ESBMC_atexit_key *head = NULL;
 
-void __ESBMC_atexit()
+void __atexit_handler()
 {
 __ESBMC_HIDE:;
   __ESBMC_atexit_key *tmp = head;
@@ -47,13 +47,13 @@ __ESBMC_HIDE:;
 #pragma GCC diagnostic ignored "-Winvalid-noreturn"
 void exit(int status)
 {
-  __ESBMC_atexit();
+  __atexit_handler();
   __ESBMC_assume(0);
 }
 
 void abort(void)
 {
-  __ESBMC_atexit();
+  __atexit_handler();
   __ESBMC_assume(0);
 }
 
--- a/src/c2goto/library/stdlib.c
+++ b/src/c2goto/library/stdlib.c
@@ -48,6 +48,7 @@ __ESBMC_HIDE:;
 #pragma GCC diagnostic ignored "-Winvalid-noreturn"
 void exit(int status)
 {
+__ESBMC_HIDE:;
   __atexit_handler();
   __ESBMC_finish_formula();
   __ESBMC_assume(0);
@@ -55,6 +56,7 @@ void exit(int status)
 
 void abort(void)
 {
+__ESBMC_HIDE:;
   __atexit_handler();
   __ESBMC_finish_formula();
   __ESBMC_assume(0);
@@ -62,6 +64,7 @@ void abort(void)
 
 void _Exit(int status)
 {
+__ESBMC_HIDE:;
   __ESBMC_finish_formula();
   __ESBMC_assume(0);
 }
--- a/src/c2goto/library/stdlib.c
+++ b/src/c2goto/library/stdlib.c
@@ -8,57 +8,42 @@
 #undef atol
 #undef getenv
 
-typedef struct atexit_key
-{
-  void (*atexit_func)();
-  struct atexit_key *next;
-} __ESBMC_atexit_key;
-
-static __ESBMC_atexit_key *head = NULL;
+unsigned short atexit_index = 0;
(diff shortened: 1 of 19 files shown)


mikhailramalho:
Some design decisions about the implementation: 
1. I decided to use a fixed array instead of a dynamic one (like the pthread destructor key) to make the model faster. I don't know how much slower a dynamic allocation array would perform.
2. We are fixed to 32 functions, since this is the minimum requirement.

lucasccordeiro (approved):
LGTM once we get the SV-COMP results from this PR.

mikhailramalho:
I found some performance regressions because the k-induction code havocs the new variables `unsigned short atexit_index = 0;` and `void (*atexit_func[32])();`.

A dynamically allocated array should fix this...

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -319,12 +319,16 @@ int __ESBMC_rounding_mode = 0;
 
 void *__ESBMC_memset(void *, int, unsigned int);
 
+void __ESBMC_finish_formula();
```
mikhailramalho: this is probably not needed.

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -319,12 +319,16 @@ int __ESBMC_rounding_mode = 0;
...
+
 // Forward decs for pthread main thread begin/end hooks. Because they're
 // pulled in from the C library, they need to be declared prior to pulling
 // them in, for type checking.
 void pthread_start_main_hook(void);
 void pthread_end_main_hook(void);
 
+void __atexit_handler(void);
```
mikhailramalho, line 330: can we get rid of it? Since it is added in clang_c_main?

src/util/simplify_expr2.cpp
```diff
@@ -1912,7 +1912,27 @@ static expr2tc simplify_relations(
...
+          return true;
+      }
+      return false;
+    };
+
+    std::function<int(expr2tc &)> get_value = [](expr2tc &) -> int {
+      return 0xbadbeef;
+    };
+
+    simpl_res = TFunctor<int>::simplify(
```
rafaelsamenezes, line 1935: Shoudn't this be moved to another PR?

src/goto-symex/symex_main.cpp
```diff
@@ -703,17 +707,8 @@ void goto_symext::finish_formula()
...
     cur_state->rename(eq);
-    target->assertion(
-      it.alloc_guard.as_expr(),
-      eq,
-      "dereference failure: forgotten memory: " + get_pretty_name(it.name),
-      cur_state->gen_stack_trace(),
-      cur_state->source,
-      first_loop);
-
-    total_claims++;
```
rafaelsamenezes, line 716: `goto_symext::assertion` does not add to total_claims.

src/c2goto/library/stdlib.c
```diff
@@ -8,15 +8,65 @@
...
+__ESBMC_HIDE:;
+  __atexit_handler();
+  __ESBMC_finish_formula();
   __ESBMC_assume(0);
 }
 
 void abort(void)
 {
+__ESBMC_HIDE:;
+  __atexit_handler();
```
rafaelsamenezes: Not sure about this.  From [Linux Die](https://linux.die.net/man/3/atexit)
```
The atexit() function registers the given function to be called at normal process termination, either via [exit](https://linux.die.net/man/3/exit)(3) or via return from the program's main().
```
I.e., abort should not call the atexit functions.

Did you do this for sv-comp?

src/clang-c-frontend/clang_c_language.cpp
rafaelsamenezes: I am not against having an intrinsic that makes the symex stop the current path. There is a use case that is different from `assume(0)`

```c
// some code
assume(0);
while(1)
 // some code
```

Is different from having an intrinsic that just stop the analysis. But we can move this to another PR

rafaelsamenezes (changes requested):

src/util/simplify_expr2.cpp
mikhailramalho: This is needed to prevent a performance regression! 
```
void __atexit_handler()
{
__ESBMC_HIDE:;
  // This is here to prevent k-induction from unwind the next loop unnecessarily
  if(__ESBMC_atexits == NULL)
    return;

  while(__ESBMC_atexits)
  {
    __ESBMC_atexits->atexit_func();
    __ESBMC_atexit_key *__ESBMC_tmp = __ESBMC_atexits->next;
    free(__ESBMC_atexits);
    __ESBMC_atexits = __ESBMC_tmp;
  }
}
```
Without it, we go ahead and try to unwind the loop

src/goto-symex/symex_main.cpp
mikhailramalho: `goto_symext::assertion` seems broken then.

src/c2goto/library/stdlib.c
mikhailramalho: actually, I'm not sure why I added that... I think we assume abort is a valid program termination in sv-comp? I guess we can remove that and re run the sv-comp and see if we get any regression.

mikhailramalho:
> I am not against having an intrinsic that makes the symex stop the current path. There is a use case that is different from assume(0)
> 
> // some code
> assume(0);
> while(1)
>  // some code
> Is different from having an intrinsic that just stop the analysis. But we can move this to another PR

Actually, this intrinsic only call symex::finish_formula which (despite its name) doesn't finish the formula per se, it only adds memory-leak checks, if it is enabled of course.

src/util/simplify_expr2.cpp
rafaelsamenezes: I see, could you update your comment to reflect this then? Also, could you add testcases with `--no-simplify`?

rafaelsamenezes:
> Actually, this intrinsic only call symex::finish_formula which (despite its name) doesn't finish the formula per se, it only adds memory-leak checks, if it is enabled of course.

I see, we should do a refactoring then. It goes beyond this PR though.

mikhailramalho:
> > Actually, this intrinsic only call symex::finish_formula which (despite its name) doesn't finish the formula per se, it only adds memory-leak checks, if it is enabled of course.
> 
> I see, we should do a refactoring then. It goes beyond this PR though.

yeah, although the documentation clearly states what this function is supposed to do, I think changing its name is desirable:
```
  /**
   *  Perform accounting checks / assertions at end of a program run.
   *  This should contain anything that must happen at the end of a program run,
   *  for example assertions about dynamic memory being freed.
   */
  void finish_formula();
```

src/util/simplify_expr2.cpp
mikhailramalho: sure!

mikhailramalho:
@rafaelsamenezes could you take another look? I implemented all you comments.

rafaelsamenezes (approved):

PR #861, opened 2022-11-08
Interval analysis fixes

rafaelsamenezes:
This PR changes how the interval analysis handles meetings for two operators, specifically how to merge intervals in a relation between two symbols. I will also run on sv-comp benchmarks to check the impact!

The following case is now solved:

```c
#define a (2)
extern unsigned int __VERIFIER_nondet_uint();
int main() { 
  int i, j=10, n=__VERIFIER_nondet_uint(), sn=0;  
  for(i=1; i<=n; i++) {
    if (i<j) 
    sn = sn + a;
    j--;
  }
  __VERIFIER_assert(sn==((long long) n)*a || sn == 0);
}
```

Because of the way intervals were computed (with intersection), it was causing `n` intervals to be computed incorrectly. 

This PR:

1. Adds a ton of documentations for the interval analysis
2. Add unit tests for the interval_templates
3. Reimplements the meet operation between two symbolic intervals by applying a contractor algorithm
4. Add support for floats


src/goto-programs/interval_domain.cpp
```diff
@@ -159,6 +147,15 @@ void interval_domaint::havoc_rec(const expr2tc &expr)
   {
     havoc_rec(to_typecast2t(expr).from);
   }
+  else if(is_code_decl2t(expr))
```
lucasccordeiro, line 187: Add some comments?

lucasccordeiro (approved):
Well done, @rafaelsamenezes. Just one small comment. I'm happy with this PR if you get consistent results over SV-COMP.

mikhailramalho (approved):
Ouch! What are the new invariants generated by this PR in the example you provided?

I'm fine as long as there is no regression in SV-COMP.

BTW since you are changing the interval analysis code, are you thinking about fixing it for floats too?

rafaelsamenezes:
> Ouch! What are the new invariants generated by this PR in the example you provided?
> 
> I'm fine as long as there is no regression in SV-COMP.
> 
> BTW since you are changing the interval analysis code, are you thinking about fixing it for floats too?

This example is not nice... the main issue was that we were restricting *N* (which is nondet) into the initial value of  *i*. So the resulting interval was 1 <= N <= 1, which was causing a false positive (N = 6 should fail).

I will also discuss more with @Mohannad-Aldughaim whether the current merging is reasonable. I can also patch for floats, the only limitation is that we will have to over-aproximate < into <=. This shouldn't cause any wrong results though.

src/goto-programs/interval_domain.cpp
rafaelsamenezes: Done

Issue #869, opened 2022-11-20
Adding CITATION.cff to repository


jossmoff:
See discussion [here](https://github.com/polywit/polywit/issues/21) between Lucas and I about adding this citation to the repo. 

**What?**
Add `CITATION.cff` to repository with the contents to produce the bibtex entry similar to the citation at http://esbmc.org/ 

**Why?**
People can directly copy either the APA/Bibtex entry straight from the repo. Example can be seen in the Polywit repository sidebar [here](https://github.com/polywit/polywit).

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/875.

PR #875, opened 2022-12-06
Citation.cff

tCm3nc:
An effort to address #869 .

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,16 +1,20 @@
 cff-version: 1.2.0
-message: "If you cite ESBMC >= 5.0 please cite us as follows"
+message: "If you cite ESBMC >= 5.0 please cite us as below."
 authors:
     - family-names: Gadelha
       given-names: Mikhail R.
+      orcid: https://orcid.org/0000-0001-6540-6587
     - family-names: Monteiro
       given-names: Felipe R.
+      orcid: https://orcid.org/0000-0001-9420-9056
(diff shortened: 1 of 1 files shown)


tCm3nc:
Hopefully I've got the right ORCID links and the right people, please let me know otherwise.
Is the citation as expected?

lucasccordeiro:
Hi guys,

Do you have any suggestions here to acknowledge ESBMC's developers?

Best,
Lucas

rafaelsamenezes:
> Do you have any suggestions here to acknowledge ESBMC's developers?

@lucasccordeiro ESBMC developers are acknowledge through the CREDITS file. 

For the future though, I would suggest changing the reference. A 4 page paper is not enough at all to be considered the work to be cited. Maybe we should consider writing a more in-depth work (journal or book chapter) which could be published/updated on arxiv (is this a valid option?).

rafaelsamenezes (approved):

lucasccordeiro:
We have a book chapter that we were expecting to publish as part of the SV-COMP effort. I mentioned to @mikhailramalho that we should upload this book chapter to arxiv, given that the publication of the book chapter is taking quite a long time (more than two years).

kunjsong01 (approved):

mikhailramalho (approved):
My ORCID is correct

shmarovfedor (approved):

Issue #877, opened 2022-12-12
Issues with casting of floating-point numbers


shmarovfedor:
The following code lists several scenarios where ESBMC crashes (`Assertion failed: (t->sort->get_data_width() == f->sort->get_data_width()), function mk_ite, file boolector_conv.cpp, line 628.`) when using Boolector as the SMT solver. However, the same program works OK with Z3. It seems that this issue with Boolector appears when you cast `c` to anything larger/small than the first cast that has been used to initialise it in `line 2`. 
```
  1 int func() {
  2   float b, c = (int)b;
  3   //assert((int)c > 0); // Works with Boolector                                                                                                                         
  4   //assert((unsigned int) c > 0); // Works with Boolector
  5   //assert((long)c > 0); // Crashes with Boolector 
  6   //assert((short)c > 0); // Crashes with Boolector 
  7   //assert((char)c > 0); // Crashes with Boolector 
  8   //assert((float)c > 0); // Works with Boolector
  9 }
```
Further investigation of the issue revealed that there are differences in the SMT formula generation for floating-point types depending on the chosen solver. Namely, it appears that this particular problem comes from the `smt_astt smt_convt::convert_typecast_from_fpbv(const typecast2t &cast)` method. It uses `mk_smt_typecast_from_fpbv_to_sbv(smt_astt from, std::size_t width)`  which has an implementation in `z3_conv` using the available Z3 API functions, but the Boolector API does not provide support for it. As the result, ESBMC uses the implementation of this method from the `solvers/smt/fp/fp_conv.cpp` module when Boolector is used. The implementation of `mk_smt_typecast_from_fpbv_to_sbv(smt_astt from, std::size_t width)` in `solvers/smt/fp/fp_conv.cpp` has a single call to `smt_astt fp_convt::mk_to_bv(smt_astt x, bool is_signed, std::size_t width)` which seems to be quite involved (about 150 lines of code, and it might be causing the described issue), and it looks like that it is responsible for the conversion of floating point types into bit vector types.

rafaelsamenezes:
@mikhailramalho Do you have any suggestions?

mikhailramalho:
I'll take a look later today.

mikhailramalho:
Fix in PR #878

PR #878, opened 2022-12-14
Fix float cast

mikhailramalho:
Fixes issue #877 

In the boolector backend we cache all the symbols created (otherwise, if we try to create a symbol with the same name as an already created symbol, boolector aborts); this causes a crash when verifying the program in issue because:

1. We first create UNSPEC_FP with sort width 32 (`float c = (int)b;`).
2. When converting `(long)c > 0`, we try again to create UNSPEC_FP, but the cache returns the old symbol.
3. The conversion fails because we were actually trying to create UNSPEC_FP with sort width 64.

The fix is simple, add the sort width to the symbol name, so we don't create symbols with the same name but different sort widths.

I also added an assertion to prevent this from happening in the future.

### DIFF
--- a/regression/esbmc/github_877/main.c
+++ b/regression/esbmc/github_877/main.c
@@ -1,0 +1,8 @@
+int main() {
+   float b = nondet_float();
+   float c = (int)b;
+   assert((long)c > 0); // Crashes with Boolector 
+   assert((short)c > 0); // Crashes with Boolector 
+   assert((char)c > 0); // Crashes with Boolector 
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):
Shall we add a regression test that produces VERIFICATION SUCCESSFUL?

mikhailramalho:
I don't think so since we are checking the conversion in this case, not the
result.

rafaelsamenezes (approved):

shmarovfedor (approved):

PR #881, opened 2023-01-16
Changes in the c_link and contextt to handle extern variables

### LINKED ISSUE #481: ESBMC can't handle extern arrays
If we define a zero-sized array in a header and then implement it in a library, ESBMC won't detect the correct bounds.

```c
// extern.h
extern int ASD[];

// extern.c
#include "extern.h"
int ASD[10];

// extern_main.c
#include "extern.h"
int main() {
    __ESBMC_assert(ASD[1] == 42,"Ok");
    return 0;
}
```

The program returns: 
```
Violated property:
  file extern_main.c line 4 function main
  array bounds violated: array `ASD' upper bound
  0
```

rafaelsamenezes:
This should close #481

### DIFF
--- a/src/util/c_link.cpp
+++ b/src/util/c_link.cpp
@@ -53,6 +53,18 @@ public:
 
   void typecheck() override;
 
+  /**
+   * @brief Merges extern symbols from old contexts
+   * 
+   * Checks whether a non-extern new_context symbol
+   * was an extern symbol in a previous context.
+   * 
+   * If it was, then merge it.
+   * 
+   * @param s 
+   */
+  void extern_fixup(symbolt &s);
+
 protected:
   void duplicate(symbolt &in_context, symbolt &new_symbol);
   void duplicate_type(symbolt &in_context, symbolt &new_symbol);
@@ -304,9 +316,26 @@ void c_linkt::duplicate_symbol(symbolt &in_context, symbolt &new_symbol)
   }
 }
 
+void c_linkt::extern_fixup(symbolt &s)
+{
+  if(!s.is_extern)
+  {
+    // Does the previous context have it?
+    auto prev = context.find_symbol(s.id);
+    if(prev)
+    {
+      // If current context is extern and previous wasnt?
+      if(!s.is_extern && prev->is_extern)
+        prev->swap(s);
+    }
+  }
+}
+
 void c_linkt::typecheck()
 {
   new_context.Foreach_operand([this](symbolt &s) {
+    // Check for externs
+    extern_fixup(s);
     // build module clash table
     if(s.file_local && known_modules.find(s.module) != known_modules.end())
     {
--- a/src/util/c_link.cpp
+++ b/src/util/c_link.cpp
@@ -55,13 +55,13 @@ public:
 
   /**
    * @brief Merges extern symbols from old contexts
-   * 
+   *
    * Checks whether a non-extern new_context symbol
    * was an extern symbol in a previous context.
-   * 
+   *
    * If it was, then merge it.
-   * 
-   * @param s 
+   *
+   * @param s
    */
   void extern_fixup(symbolt &s);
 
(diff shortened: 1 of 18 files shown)


src/util/context.cpp
```diff
@@ -135,6 +135,10 @@ symbolt *contextt::move_symbol_to_context(symbolt &symbol)
       if(symbol.type.is_not_nil() && !s->type.is_not_nil())
         s->swap(symbol);
     }
+    else if(s->is_extern && !symbol.is_extern)
+    {
```
lucasccordeiro: Remove bracket.

src/util/c_link.cpp
```diff
@@ -304,9 +316,26 @@ void c_linkt::duplicate_symbol(symbolt &in_context, symbolt &new_symbol)
...
+      if(!s.is_extern && prev->is_extern)
+        prev->swap(s);
+    }
+  }
+}
+
 void c_linkt::typecheck()
 {
   new_context.Foreach_operand([this](symbolt &s) {
+    // Check for externs
```
lucasccordeiro: Improve this comment :-)

lucasccordeiro (approved):
Minor changes.

mikhailramalho (approved):
LGTM.

The c_link is quite messy, I wonder if we could ever get rid of it... I remember seeing some code in clang to merge ASTs, if that's possible we could merge all ASTs before calling convert + adjust.

rafaelsamenezes:
> The c_link is quite messy, I wonder if we could ever get rid of it... I remember seeing some code in clang to merge ASTs, if that's possible we could merge all ASTs before calling convert + adjust.

I tried the [ASTImporter] (https://clang.llvm.org/docs/LibASTImporter.html) approach a bit. It is a bit tricky though, we have to create matchers for everything that needs merging. From what I understood, the `ASTs` that we have are for the current TU, so we would also need to refactor the frontend so that we have ASTs from every module.

mikhailramalho:
> > The c_link is quite messy, I wonder if we could ever get rid of it... I remember seeing some code in clang to merge ASTs, if that's possible we could merge all ASTs before calling convert + adjust.
> 
> I tried the [ASTImporter] (https://clang.llvm.org/docs/LibASTImporter.html) approach a bit. It is a bit tricky though, we have to create matchers for everything that needs merging.

I'm not an expert on the importer but maybe you only need to import the `TranslationUnitDecl` and everything inside will be merged: https://clang.llvm.org/doxygen/ASTMerge_8cpp_source.html

>  From what I understood, the `ASTs` that we have are for the current TU, so we would also need to refactor the frontend so that we have ASTs from every module.

Yeah, although I think it shouldn't be that big of a change. 

IIRC, we do the following:
1. `clang_c_languaget::parse`: generates clang AST.
2. `clang_c_languaget::typecheck` converts, adjust, and merge ASTs (using c_link).
3. `clang_c_languaget::final` adds model + some small changes to __ESBBMC_main.

My idea is:
1. 1. `clang_c_languaget::parse`: generates clang AST.
2. `clang_c_languaget::typecheck` merges clang AST using ASTImporter.
3. `clang_c_languaget::final` converts, adjust, adds model + some small changes to __ESBBMC_main.

That way we could get rid of the c_link, WDYT?

rafaelsamenezes:
> That way we could get rid of the c_link, WDYT?

I do think that removing `c_link` is a good idea as Clang is probably better at it. BTW, is this a good opportunity to think about a goto-linker? Maybe we don´t need a linker at language level.

For this PR though, I think that we should merge it (since it is blocking users) and convert your comment into an issue (as it requires more research/thinking).

mikhailramalho:
> > That way we could get rid of the c_link, WDYT?
> 
> I do think that removing `c_link` is a good idea as Clang is probably better at it. BTW, is this a good opportunity to think about a goto-linker? Maybe we don´t need a linker at language level.

Indeed, I also think that cbmc has a goto-linker, we could check there first.

> 
> For this PR though, I think that we should merge it (since it is blocking users) and convert your comment into an issue (as it requires more research/thinking).

I agree, I already approved this PR.

PR #854, opened 2022-10-13
[clang-cpp] Support for inheritance and polymorphism

kunjsong01:
Replaces https://github.com/esbmc/esbmc/pull/848. 

**Methodology:** 
We follow the old cpp frontend conversion flow to support inheritance and polymorphism. 
Added vtable and vptr conversions in clang-cpp-frontend. 

**Example:** 
```cpp
     3	class Vehicle
     4	{
     5	public:
     6	  Vehicle() {}
     7	  virtual int number_of_wheels() = 0;
     8	};
```

Symbols added for the above class: 
```
   1 c:@S@Vehicle@F@Vehicle#
   1 c:@S@Vehicle@F@Vehicle#this
   1 c:@S@Vehicle@F@number_of_wheels#
   1 c:@S@Vehicle@F@~Vehicle#
   1 c:@S@Vehicle@F@~Vehicle#this
   1 tag-Vehicle
   1 virtual_table::tag-Vehicle
   1 virtual_table::tag-Vehicle@tag-Vehicle
```

**Major Changes and Dev Notes:**
1. Since conversion of constructors rely on vtable and vptr, we have to use iterate over the declarators in a class in two loops:
   - First loop: we do everything but constructors and set up vtable and vptr in  in `clang_cpp_convertert::do_virtual_table`
   - Second loop: we do constructors
2. Symbol `virtual_table::tag-Vehicle` is added by `clang_c_convertert::get_virtual_method`.
3. Symbol `virtual_table::tag-Vehicle@tag-Vehicle` is added by `clang_cpp_convertert::do_virtual_table`. 

P.S. Revert commit 87f7b210dd5d060618166187df4ede3202abce9e before merging. This is the commit that contains debug lines

PR #882, opened 2023-01-16
Single inheritance

kunjsong01:
https://github.com/esbmc/esbmc/pull/854 is too large and messy for review. 

Based on previous discussion on slack, this is a PR separated from https://github.com/esbmc/esbmc/pull/854, adding support for single inheritance only. 

Features implemented: 
1. Populated class type's `components` and `methods` vectors (which will be used for polymorphism support, e.g. generating implicit ctor/dtor code for vptr initializations).
2. Added base initializer conversion and generate a type casted argument for base initializer.
3. Added the corresponding test cases.

### DIFF
--- a/regression/esbmc-cpp/inheritance_bringup/single_member/main.cpp
+++ b/regression/esbmc-cpp/inheritance_bringup/single_member/main.cpp
@@ -1,0 +1,25 @@
+#include <cassert>
+
+class Vehicle
+{
+public:
+  int data;
+  Vehicle() { data = 1; }
+};
+
+class Motorcycle: public Vehicle
+{
+public:
+  Motorcycle() : Vehicle() {}
+  //Motorcycle() {}
+};
+
+int main()
+{
+  Motorcycle m;
+  bool res1;
+  res1 = (m.data == 1); // PASS
+  assert(res1);
+  return 0;
+}
+
--- a/regression/esbmc-cpp/inheritance_bringup/single_member/main.cpp
+++ b/regression/esbmc-cpp/inheritance_bringup/single_member/main.cpp
@@ -11,7 +11,6 @@ class Motorcycle: public Vehicle
 {
 public:
   Motorcycle() : Vehicle() {}
-  //Motorcycle() {}
 };
 
 int main()
--- a/regression/esbmc-cpp/inheritance_bringup/single_member_fail/main.cpp
+++ b/regression/esbmc-cpp/inheritance_bringup/single_member_fail/main.cpp
@@ -1,0 +1,25 @@
+#include <cassert>
+
+class Vehicle
+{
+public:
+  int data;
+  Vehicle() { data = 1; }
+};
+
+class Motorcycle: public Vehicle
+{
+public:
+  Motorcycle() : Vehicle() {}
+  //Motorcycle() {}
+};
+
+int main()
+{
+  Motorcycle m;
+  bool res1;
+  res1 = (m.data == 0); // FAIL, should be 1
+  assert(res1);
+  return 0;
+}
+
--- a/regression/esbmc-cpp/inheritance_bringup/single_member_fail/main.cpp
+++ b/regression/esbmc-cpp/inheritance_bringup/single_member_fail/main.cpp
@@ -11,7 +11,6 @@ class Motorcycle: public Vehicle
 {
 public:
   Motorcycle() : Vehicle() {}
-  //Motorcycle() {}
 };
 
 int main()
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -303,7 +303,7 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   if(context.find_symbol(id) != nullptr)
     return false;
 
-  // TODO: Fix me when we have a test case using C++ unioin.
+  // TODO: Fix me when we have a test case using C++ union.
   //       A C++ union can have member functions but not virtual functions.
   //       Just use struct_typet for C++?
   struct_union_typet t;
@@ -312,12 +312,6 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   else
     t = struct_typet();
   t.tag(name);
-
-  if(rd.isClass())
-  {
-    t.set("#class", true);
-    t.set("name", id);
-  }
 
   locationt location_begin;
   get_location_from_decl(rd, location_begin);
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -383,6 +383,7 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   else
     add_padding(to_struct_type(t), ns);
 
+  t.location() = location_begin;
   added_symbol.type = t;
   return false;
 }
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -641,7 +641,7 @@ bool clang_c_convertert::get_function(
   // We need: a type, a name, and an optional body
   if(fd.hasBody())
   {
-    if(get_function_body(fd, added_symbol.value))
+    if(get_function_body(fd, added_symbol.value, type))
       return true;
   }
 
@@ -653,7 +653,8 @@ bool clang_c_convertert::get_function(
 
 bool clang_c_convertert::get_function_body(
   const clang::FunctionDecl &fd,
-  exprt &new_expr)
+  exprt &new_expr,
+  const code_typet &)
 {
   assert(fd.hasBody());
 
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -342,6 +342,8 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   // Now get the symbol back to continue the conversion
   symbolt &added_symbol = *context.find_symbol(symbol_name);
 
+  // We have to add fields before methods as the fields are likely to be used
+  // in the methods
   if(get_struct_union_class_fields(*rd_def, t))
     return true;
 
@@ -832,7 +834,7 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
     if(sub_type.is_struct() || sub_type.is_union())
     {
       struct_union_typet t = to_struct_union_type(sub_type);
-      sub_type = symbol_typet("tag-" + t.tag().as_string());
+      sub_type = symbol_typet(tag_prefix + t.tag().as_string());
     }
 
     new_type = gen_pointer_type(sub_type);
@@ -1058,7 +1060,7 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
     if(sub_type.is_struct() || sub_type.is_union())
     {
       struct_union_typet t = to_struct_union_type(sub_type);
-      sub_type = symbol_typet("tag-" + t.tag().as_string());
+      sub_type = symbol_typet(tag_prefix + t.tag().as_string());
     }
 
     new_type = gen_pointer_type(sub_type);
@@ -1547,7 +1549,7 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     if(size_type.is_struct() || size_type.is_union())
     {
       struct_union_typet t = to_struct_union_type(size_type);
-      size_type = symbol_typet("tag-" + t.tag().as_string());
+      size_type = symbol_typet(tag_prefix + t.tag().as_string());
     }
 
     new_expr.set("#c_sizeof_type", size_type);
@@ -3099,7 +3101,7 @@ void clang_c_convertert::get_decl_name(
   {
     const clang::RecordDecl &rd = static_cast<const clang::RecordDecl &>(nd);
     name = getFullyQualifiedName(ASTContext->getTagDeclType(&rd), *ASTContext);
-    id = "tag-" + name;
+    id = tag_prefix + name;
     return;
   }
 
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -3026,7 +3026,7 @@ std::string clang_c_convertert::get_decl_name(const clang::NamedDecl &nd)
   return rso.str();
 }
 
-static std::string
+inline std::string
 getFullyQualifiedName(const clang::QualType &t, const clang::ASTContext &c)
 {
   clang::PrintingPolicy Policy(c.getPrintingPolicy());
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -303,6 +303,9 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   if(context.find_symbol(id) != nullptr)
     return false;
 
+  // TODO: Fix me when we have a test case using C++ unioin.
+  //       A C++ union can have member functions but not virtual functions.
+  //       Just use struct_typet for C++?
   struct_union_typet t;
   if(rd.isUnion())
     t = union_typet();
@@ -375,7 +378,7 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
     }
   }
 
-  if(get_struct_union_class_methods(*rd_def, t))
+  if(get_struct_union_class_methods(*rd_def, to_struct_type(t)))
     return true;
 
   if(rd.isUnion())
@@ -415,7 +418,7 @@ bool clang_c_convertert::get_struct_union_class_fields(
 
 bool clang_c_convertert::get_struct_union_class_methods(
   const clang::RecordDecl &,
-  struct_union_typet &)
+  struct_typet &)
 {
   // We don't add methods to the struct in C
   return false;
@@ -558,7 +561,9 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
   return false;
 }
 
-bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
+bool clang_c_convertert::get_function(
+  const clang::FunctionDecl &fd,
+  exprt &new_expr)
 {
   // Don't convert if implicit, unless it's a constructor or destructor
   // A compiler-generated default ctor/dtor is considered implicit, but we have
@@ -631,6 +636,7 @@ bool clang_c_convertert::get_function(const clang::FunctionDecl &fd, exprt &)
     type.make_ellipsis();
 
   added_symbol.type = type;
+  new_expr.type() = type;
 
   // We need: a type, a name, and an optional body
   if(fd.hasBody())
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -3026,7 +3026,7 @@ std::string clang_c_convertert::get_decl_name(const clang::NamedDecl &nd)
   return rso.str();
 }
 
-inline std::string
+std::string
 getFullyQualifiedName(const clang::QualType &t, const clang::ASTContext &c)
 {
   clang::PrintingPolicy Policy(c.getPrintingPolicy());
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -394,44 +394,12 @@ bool clang_c_convertert::get_struct_union_class_fields(
       return true;
 
     // Check for alignment attributes
-    if(field->hasAttrs())
-    {
-      const auto &attrs = field->getAttrs();
-      for(const auto &attr : attrs)
-      {
-        if(attr->getKind() == clang::attr::Aligned)
-        {
-          const clang::AlignedAttr &aattr =
-            static_cast<const clang::AlignedAttr &>(*attr);
-
-          if(aattr.isAlignmentExpr())
-          {
-            // This is usually a constant
-            clang::Expr *alignExpr = aattr.getAlignmentExpr();
-            exprt alignment;
-            if(alignExpr && get_expr(*(aattr.getAlignmentExpr()), alignment))
-              return true;
-            comp.type().set("alignment", alignment);
-          }
-          else
-          {
-            // I was not able to find an example to test this, so abort for now
-            log_error("ESBMC currently does not support type alignments");
-            std::ostringstream oss;
-            llvm::raw_os_ostream ross(oss);
-            aattr.getAlignmentType()->getType()->dump(ross, *ASTContext);
-            ross.flush();
-            log_error("{}", oss.str());
-            return true;
-          }
-        }
-      }
-    }
+    if(check_alignment_attributes(field, comp))
+      return true;
 
     // Don't add fields that have global storage (e.g., static)
-    if(const clang::VarDecl *nd = llvm::dyn_cast<clang::VarDecl>(field))
-      if(nd->hasGlobalStorage())
-        continue;
+    if(is_field_global_storage(field))
+      continue;
 
     type.components().push_back(comp);
   }
@@ -3298,4 +3266,52 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
   }
 
   return nullptr;
+}
+
+bool clang_c_convertert::check_alignment_attributes(
+  const clang::FieldDecl *field,
+  struct_typet::componentt &comp)
+{
+  if(field->hasAttrs())
+  {
+    const auto &attrs = field->getAttrs();
+    for(const auto &attr : attrs)
+    {
+      if(attr->getKind() == clang::attr::Aligned)
+      {
+        const clang::AlignedAttr &aattr =
+          static_cast<const clang::AlignedAttr &>(*attr);
+
+        if(aattr.isAlignmentExpr())
+        {
+          // This is usually a constant
+          clang::Expr *alignExpr = aattr.getAlignmentExpr();
+          exprt alignment;
+          if(alignExpr && get_expr(*(aattr.getAlignmentExpr()), alignment))
+            return true;
+          comp.type().set("alignment", alignment);
+        }
+        else
+        {
+          // I was not able to find an example to test this, so abort for now
+          log_error("ESBMC currently does not support type alignments");
+          std::ostringstream oss;
+          llvm::raw_os_ostream ross(oss);
+          aattr.getAlignmentType()->getType()->dump(ross, *ASTContext);
+          ross.flush();
+          log_error("{}", oss.str());
+          return true;
+        }
+      }
+    }
+  }
+  return false;
+}
+
+bool clang_c_convertert::is_field_global_storage(const clang::FieldDecl *field)
+{
+  if(const clang::VarDecl *nd = llvm::dyn_cast<clang::VarDecl>(field))
+    return (nd->hasGlobalStorage());
+
+  return false;
 }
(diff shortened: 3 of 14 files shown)


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1097,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+}
+
+void clang_cpp_convertert::gen_typecast_base_ctor_call(
+  const exprt &callee_decl,
+  side_effect_expr_function_callt &call,
+  exprt &initializer)
+{
+  // sanity checks
+  assert(initializer.derived_this_arg() != "");
+  assert(initializer.base_ctor_derived());
```
kunjsong01, line 1206: @mikhailramalho 
This function would appear to be adjustment only. 
Shall we move it to adjuster?

regression/esbmc-cpp/inheritance_bringup/single_member_fail/main.cpp
```diff
@@ -0,0 +1,25 @@
...
+public:
+  int data;
+  Vehicle() { data = 1; }
+};
+
+class Motorcycle: public Vehicle
+{
+public:
+  Motorcycle() : Vehicle() {}
+  //Motorcycle() {}
```
lucasccordeiro: Remove this line.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3291,3 +3280,51 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
 
   return nullptr;
 }
+
+bool clang_c_convertert::check_alignment_attributes(
```
lucasccordeiro, line 3278: You should comment on this method.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3291,3 +3280,51 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
...
+  const clang::FieldDecl *field,
+  struct_typet::componentt &comp)
+{
+  if(field->hasAttrs())
+  {
+    const auto &attrs = field->getAttrs();
+    for(const auto &attr : attrs)
+    {
+      if(attr->getKind() == clang::attr::Aligned)
```
lucasccordeiro, line 3287: Add explanation.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -220,7 +221,19 @@ bool clang_cpp_convertert::get_function(
   if(fd.isDependentContext())
     return false;
 
-  return clang_c_convertert::get_function(fd, new_expr);
+  // Only convert non-trivial functions, e.g. non-trivial ctor/dtor, copy ctor, move ctor .etc
```
lucasccordeiro: Remove this block of comments.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -220,7 +221,19 @@ bool clang_cpp_convertert::get_function(
...
+  //if(fd.isTrivial())
+  //  return false;
+
+  if(clang_c_convertert::get_function(fd, new_expr))
+    return true;
+
+  // add additional annotations for class/struct/union methods
+  if(const auto *md = llvm::dyn_cast<clang::CXXMethodDecl>(&fd))
```
lucasccordeiro, line 228: Shall we write it as an AND statement?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -268,23 +282,60 @@ bool clang_cpp_convertert::get_struct_union_class_fields(
...
+      cxxrd->getNameAsString(),
+      cxxrd->getNumBases(),
+      cxxrd->getNumVBases());
+
+    // TODO-split: add methods from base class
+    for(auto base : cxxrd->bases())
+    {
+      // `base` type is clang::CXXBaseSpecifier
+      base.getType().dump();
+      //log_error("inheritance is not supported in {}", __func__);
```
lucasccordeiro: Shall we remove these lines?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -321,6 +372,10 @@ bool clang_cpp_convertert::get_struct_union_class_methods(
     {
       // Add only if it isn't static
       if(!cxxmd->isStatic())
+      {
```
lucasccordeiro: Remove brackets?

lucasccordeiro (changes requested):
Some minor comments.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: It checks to see if the operand is of the specified type, and if so, returns a pointer to it, otherwise returns a nullptr.  

This is part of the `clang_cpp_convertert::get_function` which is used to convert both ordinary functions and class methods. So we might get an ordinary function rather than the non-CXXMethodDecl for class method, in which case we'll get a nullptr.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Removed, this will be part of the next PR "split-polymorphism".

lucasccordeiro (approved):
LGTM.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -298,13 +303,22 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
...
+  //       A C++ union can have member functions but not virtual functions.
+  //       Just use struct_typet for C++?
   struct_union_typet t;
   if(rd.isUnion())
     t = union_typet();
   else
     t = struct_typet();
   t.tag(name);
 
+  if(rd.isClass())
```
mikhailramalho: is this needed?

regression/esbmc-cpp/inheritance_bringup/single_member/main.cpp
```diff
@@ -0,0 +1,25 @@
...
+public:
+  int data;
+  Vehicle() { data = 1; }
+};
+
+class Motorcycle: public Vehicle
+{
+public:
+  Motorcycle() : Vehicle() {}
+  //Motorcycle() {}
```
mikhailramalho: remove

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -433,7 +418,7 @@ bool clang_c_convertert::get_struct_union_class_fields(
 
 bool clang_c_convertert::get_struct_union_class_methods(
   const clang::RecordDecl &,
-  struct_union_typet &)
+  struct_typet &)
```
mikhailramalho, line 415: why this change? I think unions can have methods.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3038,7 +3027,7 @@ std::string clang_c_convertert::get_decl_name(const clang::NamedDecl &nd)
   return rso.str();
 }
 
-static std::string
+std::string
```
mikhailramalho, line 3024: this can be an inline static function, why you removed it?

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3291,3 +3280,51 @@ clang_c_convertert::get_top_FunctionDecl_from_Stmt(const clang::Stmt &stmt)
...
+          log_error("{}", oss.str());
+          return true;
+        }
+      }
+    }
+  }
+  return false;
+}
+
+bool clang_c_convertert::is_field_global_storage(const clang::FieldDecl *field)
```
mikhailramalho, line 3318: do we really need a separate function for this?

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -38,8 +38,12 @@ class StringLiteral;
 class IntegerLiteral;
 class FloatingLiteral;
 class TagDecl;
+class FieldDecl;
 } // namespace clang
 
+extern std::string
+getFullyQualifiedName(const clang::QualType &, const clang::ASTContext &);
```
mikhailramalho, line 45: This doesn't need to be extern, since its definition is in the cpp file. 

Also, why do you need this?

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: I see why you need it now: to reuse in the cpp convert. Please use it in clang_cpp_convertert::get_struct_union_class_fields, in the loop after `  // let's check for (virtual) base classes if this declaration is a CXXRecordDecl,`

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -268,23 +278,58 @@ bool clang_cpp_convertert::get_struct_union_class_fields(
...
+      "Class {} has {} bases, {} virtual bases",
+      cxxrd->getNameAsString(),
+      cxxrd->getNumBases(),
+      cxxrd->getNumVBases());
+
+    // TODO-split: add methods from base class
+    for(auto base : cxxrd->bases())
+    {
+      // `base` type is clang::CXXBaseSpecifier
+      base.getType().dump();
```
mikhailramalho: cleanup this.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -268,23 +278,58 @@ bool clang_cpp_convertert::get_struct_union_class_fields(
...
+
+    // Check for alignment attributes
+    if(check_alignment_attributes(field, comp))
+      return true;
+
+    // Don't add fields that have global storage (e.g., static)
+    if(is_field_global_storage(field))
+      continue;
+
+    // [C++ annotation]: set parent in component's type
```
mikhailramalho, line 296: do we need to set this field?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -807,8 +866,14 @@ bool clang_cpp_convertert::get_function_body(
         }
         else
         {
-          log_error("Base class initializer is not supported in {}", __func__);
-          abort();
+          // TODO-split: add base class initializer
+          init->getInit()->dump();
```
mikhailramalho: cleanup this

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -807,8 +866,14 @@ bool clang_cpp_convertert::get_function_body(
...
+          // Add additional annotation for `this` parameter
+          initializer.derived_this_arg(
+            ftype.arguments().at(0).get("#identifier"));
+          initializer.base_ctor_derived(true);
```
mikhailramalho, line 873: why do we need to call derived_this_arg and base_ctor_derived?

I think the get_expr in the next line will overwrite anything set in this initializer object.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+  std::string access;
+  if(get_access_from_decl(field, access))
+    return true;
+
+  // annotate access in component
+  comp.set_access(access);
+  return false;
+}
+
+bool clang_cpp_convertert::get_access_from_decl(
```
mikhailramalho, line 1105: do we actually need this? We are not doing any typecheck in the IR, so the access information is useless for us (clang already check if not code call violates access definitions).

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+  {
+    typet rtn_type("constructor");
+    component_type.return_type() = rtn_type;
+  }
+
+  // annotate parent
+  std::string parent_class_name = getFullyQualifiedName(
+    ASTContext->getTagDeclType(cxxmdd->getParent()), *ASTContext);
+  std::string parent_class_id = tag_prefix + parent_class_name;
+  component_type.set("#member_name", parent_class_id);
```
mikhailramalho, line 1158: do we need this information?

Also, instead of exporting getFullyQualifiedName, you should call `get_decl(cxxmdd->getParent())` and access its name.

I fail to see why we need this code tho.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+
+  // annotate name
+  std::string method_id, method_name;
+  clang_c_convertert::get_decl_name(*cxxmdd, method_name, method_id);
+  new_expr.name(method_id);
```
mikhailramalho, line 1163: do we need this?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+
+  // annotate access
+  std::string access;
+  if(get_access_from_decl(cxxmdd, access))
+    return true;
+  new_expr.set("access", access);
```
mikhailramalho, line 1169: do we need this?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+
+  // annotate base name
```
mikhailramalho, line 1171: These two lines feel wrong...

base_name = unique identifier
pretty_name = name printed when we use --goto-functions-*, --symbol-table-*, --program-* and others

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+  new_expr.base_name(parent_class_name);
+
+  // annotate pretty name
+  new_expr.pretty_name(parent_class_name);
+
+  // annotate inline
+  new_expr.set("is_inlined", cxxmdd->isInlined());
```
mikhailramalho, line 1178: do we need this?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1024,3 +1089,138 @@ bool clang_cpp_convertert::get_decl_ref(
...
+
+  // annotate location
+  locationt location_begin;
+  get_location_from_decl(*cxxmdd, location_begin);
+  new_expr.location() = location_begin;
+
+  // We still need to add a trivial ctor/dtor as a `component` in class symbol's type
+  // remove "statement: skip" otherwise it won't be added
+  // TODO-split: Any more elegant way to do it?
```
mikhailramalho, line 1187: Can you explain why we need this code?

src/clang-cpp-frontend/clang_cpp_convert.cpp
mikhailramalho: It's hard to tell... can you explain why we need this function?

src/util/irep.cpp
```diff
@@ -538,6 +538,8 @@ const irep_idt irept::a_object_type = dstring("object_type");
 const irep_idt irept::a_cmt_size = dstring("#size");
 const irep_idt irept::a_cmt_type = dstring("#type");
 const irep_idt irept::a_type_id = dstring("typeid");
+const irep_idt irept::a_derived_this_arg = dstring("#derived_this_arg");
```
mikhailramalho, line 541: why do we need this?

src/util/irep.h
```diff
@@ -264,6 +264,16 @@ class irept
     return get(a_pretty_name);
   }
 
+  inline const irep_idt &derived_this_arg() const
```
mikhailramalho, line 267: ditto

src/util/irep.h
```diff
@@ -940,6 +950,16 @@ class irept
     set(a_pretty_name, val);
   }
 
+  inline void derived_this_arg(const irep_idt &val)
```
mikhailramalho, line 953: ditto

src/util/irep.h
```diff
@@ -1232,6 +1252,8 @@ class irept
   static const irep_idt a_end_location, a_guard, a_label, a_lhs, a_location;
   static const irep_idt a_object_type, a_cmt_size, a_cmt, a_type_id;
   static const irep_idt a_cmt_type;
+  // annotations for typecasting derived derived class `this` to base class type
```
mikhailramalho, line 1255: ditto

mikhailramalho (changes requested):
@kunjsong01 I added several comments, including a lot of nitpick. Remember to remove unnecessary code (calls to .dump()) and run clang-format for the next version).

Please, double-check all the new attributes we are setting and if they are actually needed.

Your patches are getting increasingly harder to review (that's why I asked you to explain some of your decisions). This is a good sign.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -298,13 +303,22 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
   if(context.find_symbol(id) != nullptr)
     return false;
 
+  // TODO: Fix me when we have a test case using C++ unioin.
```
rafaelsamenezes: unioin =>  union

rafaelsamenezes (changes requested):
Could you add more test-cases? For: static fields, nested inheritance, and pointer casts .

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: I suppose so... just need to distinguish class from struct, making it more intuitive in the symbol table, otherwise it would mark everything `struct`:
Screenshot 2023-01-25 at 12 51 00

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: > why this change?

Need to use the `methods()` as in `type.methods().push_back(comp);`. Our `struct_union_typet` class doesn't have this function.

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: If it's just to print "class" instead of "struct", we don't really need it.

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: Do we need to add the methods to the type? i.e., is there anywhere in esbmc after the frontend (goto, ssa, smt), that we call methods() to get the list of methods?

If it's only to add the methods to the type, I'd say we don't need it.

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: done removed

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: Yes, to generate the implicit code for constructors for vptr initializations.

src/clang-c-frontend/clang_c_convert.h
kunjsong01: Done removed `extern`. 

>why do you need this?

Need to be used in clang-cpp-converter to get the parent_class_id

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: done

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: >Please use it in clang_cpp_convertert::get_struct_union_class_fields, in the loop after   // let's check for (virtual) base classes if this declaration is a CXXRecordDecl,

Yep, it's being used in the `//Parse the fields` loop after the that base loop.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Done removed

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: >why do we need to call derived_this_arg and base_ctor_derived?

In short, the purpose is to generate typecasted `this` argument when calling base ctor in derived ctor, something like:
```cpp
 91 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 92
 93 Motorcycle (c:@S@Motorcycle@F@Motorcycle#):
 94         // 32
 95         FUNCTION_CALL:  Vehicle(( Vehicle *)this)  // <---- typecasted `this` argument
 96         // 33 file main.cpp line 13 function Motorcycle
 97         END_FUNCTION // Motorcycle
 98 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
"derived_this_arg" == symbol id for Motorcycle `this`
"base_ctor_derived" means "we are implicitly calling the base ctor in derivied ctor" 
(sorry about the wacky naming as I wan't able to give it a more intuitive and compact name :sweat_smile:) 

>I think the get_expr in the next line will overwrite anything set in this initializer object.

Correct... but we'll use them before the overwriting (i.e. the swapping) in `clang_cpp_convertert::get_constructor_call`. 

P.S. Combined with the new function `clang_cpp_convertert::gen_typecast_base_ctor_call`, I found these flags work pretty well when dealing with multiple inheritances (we'll see more test cases enabled for multiple inheritance in https://github.com/esbmc/esbmc/pull/884 when this PR is merged). Basically it's able to generate: 
```cpp
127 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
128 // given Rectangle class derives from Shape and PaintCost classes
129 Rectangle (c:@S@Rectangle@F@Rectangle#):
130         // 40
131         FUNCTION_CALL:  Shape(( Shape *)this)
132         // 41
133         FUNCTION_CALL:  PaintCost(( PaintCost *)this)
134         // 42 file main.cpp line 28 function Rectangle
135         END_FUNCTION // Rectangle
136 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: I see... thanks very much for the explanations. I mistakes base_name for the parent class id. These annotations are added to match the symbol types generated by the old frontend. 


Now fixed: 
base_name = method id
pretty_name = method name

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Preparation for vptr and vtable extensions. We need to make some implicit code for vptr later - sorry it may not be obvious for the time being.

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: done fixed.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: >Also, instead of exporting getFullyQualifiedName, you should call get_decl(cxxmdd->getParent()) and access its name.

well I thought of using `get_decl(cxxmdd->getParent(), new_expr)` ...  but it would appear to go through lots of recursions, seems to be an overkill and might be quite head scratching when we need to debug in the future (e.g. dealing with 25 frames in gdb). Obviously, here we just want the to get the id and annotate `member_name`. Using the getFullyQualifiedName function appears to be a neat way to do so.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: >can you explain why we need this function?

to generate typecasted `this` argument when calling base ctor in derived ctor. 
Please refer to the comment https://github.com/esbmc/esbmc/pull/882#discussion_r1088164013

kunjsong01:
Hi @mikhailramalho , thanks a lot for the insightful reviews (and the explanations). I've fixed most of them. 

>do we need this? 

Good question.  
Honestly I also feel that I've added more annotations than possibly required. It would appear that adding the additional annotations might seem unnecessary, and the purpose is not quite obvious .

It's difficult to say for sure whether we need them or not. I just added these annotations to match the symbol table as much as I can. IIRC, some of them are needed to generate vtable and vptr. I may need them to generate implicit code in ctors/dtors for vptr stuff (this will become another PR after https://github.com/esbmc/esbmc/pull/884).

Apologies for not being able to give a concrete answer to this question. In order to speed up the develoment, I'm currently adopting a "learn-and-try" style development. As the development moves forward, it would become clearer. 

For the time being, I'd suggest let's leave it be and focus on pushing the development to pass more test cases (and yes we have more TCs passing in https://github.com/esbmc/esbmc/pull/884 :smiley:). But please be reassured that I'll purge these annotations after passing the virtual method test cases, and also please be assured that for any additional annotation to be added I'll try to confine them in a function so that it would be easy to indentify the redundant in the future.

kunjsong01:
> Could you add more test-cases? For: static fields, nested inheritance, and pointer casts .

Hi @rafaelsamenezes thanks very much for the review. 
Good point. But this would be out of scope for this PR. I've got more test cases passing in the next PR https://github.com/esbmc/esbmc/pull/884 (which is currently gated by this PR) :smiley:. Currently, the focus of development is to pass the inheritance test suite in esbmc-cpp/inheritance.

mikhailramalho:
> Hi @mikhailramalho , thanks a lot for the insightful reviews (and the explanations). I've fixed most of them.
> 
> > do we need this?
> 
> Good question. Honestly I also feel that I've added more annotations than possibly required. It would appear that adding the additional annotations might seem unnecessary, and the purpose is not quite obvious .
> 
> It's difficult to say for sure whether we need them or not. I just added these annotations to match the symbol table as much as I can. IIRC, some of them are needed to generate vtable and vptr. I may need them to generate implicit code in ctors/dtors for vptr stuff (this will become another PR after #884).
> 
> Apologies for not being able to give a concrete answer to this question. In order to speed up the develoment, I'm currently adopting a "learn-and-try" style development. As the development moves forward, it would become clearer.
> 
> For the time being, I'd suggest let's leave it be and focus on pushing the development to pass more test cases (and yes we have more TCs passing in #884 smiley). But please be reassured that I'll purge these annotations after passing the virtual method test cases, and also please be assured that for any additional annotation to be added I'll try to confine them in a function so that it would be easy to indentify the redundant in the future.

Alright, given that this is not a critical part of esbmc, I agree to accept it for now.

But please, remember that: 
(1) esbmc is not well optimized for memory usage, so every extra annotation you add that is not necessary, more memory we are using unnecessarily.
(2) The more annotations you add, the more trouble you'll have when rewriting the frontend to use irep2 in the future. For instance, there is no access annotation in irep2. They are not even migrated in the migration code. I strongly believe that these are not necessary, and you might waste time migrating them to irep2 before getting a PR rejected.

My point is: be wary of your time :)

mikhailramalho:
Also, it's weird that the code style bot didn't complain about the unordered headers, is this option disabled @rafaelsamenezes? Can we enable it?

rafaelsamenezes:
> Also, it's weird that the code style bot didn't complain about the unordered headers

What do you mean by sorted? As in category or alphabetical?

For categories there are some regex that we can add: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories

Alphabetical is a matter of adding [SortIncludes](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#sortincludes) into .clang-format file.

rafaelsamenezes (approved):

Issue #885, opened 2023-01-30
ESBMC crashs at smt_convt


rafaelsamenezes:
I’m getting this error:

esbmc: ../src/solvers/smt/smt_conv.cpp:349: const smt_ast* smt_convt::convert_ast(const expr2tc&): Assertion `src_expr->type == ut.members[c]' failed.

```c
//test_abort.c:
#include "test_abort.h"

 void func1 (int y);
void func1 (int y)
{
                func2(y);
}

 //test_abort.h:

 typedef union {

     unsigned int uint_val;

     float fl_val;

}float_u;

 typedef struct

{

    float_u s1;

}struct_1_t;

 struct_1_t struct_1_var = {

                                        .s1.fl_val =   0,

                                       };

 typedef enum {

    ENUM_1_1,

    ALL_ENUMS_1

} enum1_e;

 

typedef enum

{

    ENUM_2_1,

    ENUM_2_2 = ENUM_2_1,

    ALL_ENUM2_2

}enum2_e;

 

typedef struct {

    enum2_e s2;

}struct_2_t;

 

typedef struct

{

    const struct_2_t s3[ALL_ENUMS_1];

}struct_3_t;

 

extern struct_3_t struct3_var;

 

void func2(enum1_e int_id)

{

    int x1 = struct3_var.s3[int_id].s2;

}

```

Edit:

Extra info, we need to run with `--no-slice` and boolector.

lucasccordeiro:
@fbrausse : Can I ask you to check this issue to see whether we have a reasonable fix?

fbrausse:
Can't reproduce this, please give the command line. I'm getting
```
src/esbmc/esbmc gh-885/test_abort.c --function func1
[...]
Counterexample:

State 1 file test_abort.h line 74 function func2 thread 0
----------------------------------------------------
Violated property:
  file test_abort.h line 74 function func2
  array bounds violated: array lower bound
  (signed long int)int_id >= 0


VERIFICATION FAILED
```

lucasccordeiro:
It also works for me if I fix some parsing errors:

```
Parsing
union.c:7:17: warning: implicit declaration of function 'func2' is invalid in C99 [-Wimplicit-function-declaration]
                func2(y);
                ^
union.c:80:6: error: conflicting types for 'func2'
void func2(enum1_e int_id)
     ^
union.c:7:17: note: previous implicit declaration is here
                func2(y);
                ^
ERROR: PARSING ERROR
```

```
[...]

PR #904, opened 2023-02-16
Fix union initializer type

fbrausse:
Fixes the issue raised in #885.

The Clang frontend set the initialized field name of constant union expressions, but did not set the right type. This PR modifies the parser to handle union initializer lists different from struct/array/vector following the way the Clang AST is structured: either exactly one initializer or none (in case of empty union type). This differs from the way struct initializers are represented (by implicit init AST nodes) and thus warrants a different handling in the Clang frontend.

### DIFF
--- a/regression/esbmc/github_885-valid/test.desc
+++ b/regression/esbmc/github_885-valid/test.desc
@@ -1,0 +1,4 @@
+CORE
+test_abort.c
+--function func1 --no-slice
+^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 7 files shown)


lucasccordeiro (approved):
LGTM.

@fbrausse : can you please run clang-format for the test case code?

fbrausse:
> @fbrausse : can you please run clang-format for the test case code?

Done.

rafaelsamenezes (approved):

PR #907, opened 2023-02-21
[util] replace_symbolt learned to handle argumentt

### LINKED ISSUE #902: C frontend does handle Static functions properly
ESBMC does not handle the following programs properly:

```c
//main.c

#include "file1.h"

 int main()

{

    func1(32);

}

 //file1.c

#include "file1.h"

 //file1.h

typedef struct {

                int arr[33];

} struct_t;

 

struct_t tmp_struct;

 

static int func1(int i)

{

                return tmp_struct.arr[i];

}
```

By invoking `esbmc --function main main.c file1.c` esbmc gives an out of bounds. Removing the Static keyword or removing `file1.c` changes the output.

fbrausse:
Fixes #902.

Unlike symbols, arguments to functions are not to be replaced but updated.
That is because they are not symbols and their cmt_identifier() is used to
refer to the corresponding symbol's name. We update the argumentt by
replacing its cmt_identifier, type and location but otherwise leaving it
alone.

### DIFF
--- a/regression/esbmc/github_902-extern/file1.c
+++ b/regression/esbmc/github_902-extern/file1.c
@@ -1,0 +1,2 @@
+
+#include "file1.h"
--- a/regression/esbmc/github_902-extern/file1.h
+++ b/regression/esbmc/github_902-extern/file1.h
@@ -1,0 +1,20 @@
+
+typedef struct {
+
+                int arr[33];
+
+} struct_t;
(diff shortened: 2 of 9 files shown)


lucasccordeiro (approved):
LGTM.

rafaelsamenezes (approved):

PR #911, opened 2023-02-21
Fix WinCI

Anthonysdu:


lucasccordeiro (approved):

.github/workflows/release.yml
```diff
@@ -56,58 +56,50 @@ jobs:
...
-    - name: Configure CMake
-      run: |
-        mkdir build
-        cd build
-        cmake.exe .. -A x64 -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake -DENABLE_REGRESSION=On -DBUILD_TESTING=On -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBOOST_DLL_FILE="C:/deps/boost_filesystem-vc143-mt-x64-1_80.dll;C:/deps/boost_program_options-vc143-mt-x64-1_80.dll" -DCMAKE_INSTALL_PREFIX:PATH=C:/deps/esbmc -DENABLE_SOLIDITY_FRONTEND=On -DENABLE_JIMPLE_FRONTEND=On -DDOWNLOAD_WINDOWS_DEPENDENCIES=On
-    - name: Build ESBMC
-      run: |
-        cd build
-        cmake --build . --target INSTALL --config RelWithDebInfo
-        cpack -G ZIP -C RelWithDebInfo
```
rafaelsamenezes, line 105: Add back `cpack`

rafaelsamenezes (changes requested):
minor changes required

src/c2goto/CMakeLists.txt
```diff
@@ -18,22 +18,27 @@ target_link_libraries(c2goto
...
   set (clib32f_config clib32_fp.c --32 --floatbv)
   set (clib64f_config clib64_fp.c --64 --floatbv)
-  set (inputs clib32_config
-              clib64_config
-              clib32f_config
-              clib64f_config
-      )
+
+  if(DISABLE_32_BITS_LIBRARIES)
+    set (inputs clib64f_config)
```
fbrausse: There seems to be a mismatch here: disabling "the 32-bit libs" also disables the `clib64_config` (without `f`).

src/c2goto/cprover_library.cpp
```diff
@@ -10,15 +10,17 @@
...
 extern "C"
 {
+  extern const uint8_t clib64_fp_buf[];
+  extern const unsigned int clib64_fp_buf_size;
+
+#ifndef DISABLE_32_BITS_LIBRARIES
   extern const uint8_t clib32_buf[];
   extern const uint8_t clib64_buf[];
   extern const unsigned int clib32_buf_size;
   extern const unsigned int clib64_buf_size;
```
fbrausse: You'll also want to enable the corresponding symbols here...

src/c2goto/cprover_library.cpp
```diff
@@ -38,6 +41,13 @@ static const struct buffer
     {&clib32_fp_buf[0], clib32_fp_buf_size},
     {&clib64_fp_buf[0], clib64_fp_buf_size},
   },
+#else
+  {{NULL, 0}, {NULL, NULL}},
+  {
+    {NULL, 0},
+    {&clib64_fp_buf[0], clib64_fp_buf_size},
```
fbrausse: ... and here.

fbrausse:
What's the purpose of selectively disabling particular pre-built operational models like the 32-bit ones? Is it related to Windows in some way? If so, is it because 32-bit code is unsupported by ESBMC on Windows, in which case shouldn't the `--32` switch also be disabled there or is there another reason?

Anthonysdu:
> What's the purpose of selectively disabling particular pre-built operational models like the 32-bit ones? Is it related to Windows in some way? If so, is it because 32-bit code is unsupported by ESBMC on Windows, in which case shouldn't the `--32` switch also be disabled there or is there another reason?

@rafaelsamenezes Could you explain why we set this build option?

rafaelsamenezes:
> @rafaelsamenezes Could you explain why we set this build option?

@Anthonysdu In my previous branch was to build with LLVM 14 in Windows as it was causing issues. You can revert it, it won't affect this PR since you are still relying on LLVM 11.

> If so, is it because 32-bit code is unsupported by ESBMC on Windows, in which case shouldn't the --32 switch also be disabled there or is there another reason?

@fbrausse You are correct, the switch needs to be removed as well.

lucasccordeiro:
@Anthonysdu : can you please rebase this branch with the master branch?

lucasccordeiro:
@fbrausse and @rafaelsamenezes : can I ask you whether you have further comments for this PR?

fbrausse (approved):
LGTM.

@Anthonysdu maybe a small hint for future PRs: the git history is quite important for future development and fixing bugs, the more info about the actual changes and intention you put into the commit message, the better. The info what files are changed by a commit is not necessary there as it can easily be obtained with `git whatchanged`. You can also use `git revert` or `git rebase -i` to revert a previous commit.

PR #913, opened 2023-02-22
Add support for solidity “constructor” keyword

ChenfengWei0:
Add support for "constructor" keywords in solidity front-end.


lucasccordeiro (changes requested):
Can you add regression tests with assertions that rely on the constructors/destructors?

ChenfengWei0:
> Can you add regression tests with assertions that rely on the constructors/destructors?

Sure

regression/esbmc-solidity/constructor_1/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+contract.solast
+--function empty --contract contract.sol
+^VERIFICATION SUCCESSFUL$
```
kunjsong01, line 4: Could you add a test case for "VERIFICATION FAILED" please?

kunjsong01 (changes requested):

ChenfengWei0:
> Can you add regression tests with assertions that rely on the constructors/destructors?

ChenfengWei0:
I just found that some features of constructor are not fully supported. Likely to take more time to resolve. Sorry.

Issue #886, opened 2023-01-30
Add timestamp and location to LOG messages


rafaelsamenezes:
It would be very helpful if:

- `log_debug`: preappended with `__FILE__, __FUNCTION__, __LINE__, or/and __PRETTY_FUNCTION__`
- all log messages had a timestamp associated.

kushalnl7:
Hi, can I work on this?

lucasccordeiro:
That’s okay.Sent from my iPhoneOn 30 Jan 2023, at 18:46, Kushal Lahoti ***@***.***> wrote:﻿
Hi, can I work on this?

rafaelsamenezes:
Relevant file: https://github.com/esbmc/esbmc/blob/master/src/util/message.h

fbrausse:
For me, it would *not* be helpful, if
> - all log messages had a timestamp associated.

*edit* Can we make this opt-in via an envvar or a cmd-line flag?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2097.

PR #916, opened 2023-02-25
Added timestamp and location to LOG messages

eyaget:
Added timestamp and location to LOG messages (to address issue #886).


src/util/message.h
```diff
@@ -44,10 +45,57 @@ struct messaget
...
+    	     
+      	/** 
+		* Newly added code to address issue #886.
+			* It prepends a well formatted timestamp, file path, function name, line number and log level to each LOG message. 
+			* It is transparent to the existing logging function, i.e., log_message.
+			* TODO: It can be implemented using a separate function and invoked here and/or elsewhere when required. 		
+		*/		
+		//========Start===============================================================================================
+		struct tm _stime; //A time structure consisting of 9 members, such as tm_sec, tm_min, tm_hour, tm_mday, tm_mon, tm_year, etc. 
+		char _buffer[20]; //A variable to hold values of certain _stime members.
```
lucasccordeiro: Why do you declare this buffer with size 20?

src/util/message.h
```diff
@@ -44,10 +45,57 @@ struct messaget
...
+		char _vrb[20]; //A variable to hold the log level.
```
lucasccordeiro: Why do you declare this buffer with size 20?

src/util/message.h
```diff
@@ -44,10 +45,57 @@ struct messaget
...
+				printf("Incorrect or no verbosity!\n");
+		}
+		
+		fprintf(f, "%s [%s] [%s:%d] %s ", _buffer, __FILE__, __func__, __LINE__, _vrb); /* Prepends the timestamp, file name, file path, 
+		* function name, line number and log level to each log message. */
+		fmt::print(f, std::forward<Args>(args)...);
+	  	fmt::print(f, "\n");
+	  	//========End=================================================================================================
+      
+		/*
```
lucasccordeiro: You can remove this commented code.

src/util/message.h
```diff
@@ -44,10 +45,57 @@ struct messaget
...
+		char _buffer[20]; //A variable to hold values of certain _stime members.
+		char _vrb[20]; //A variable to hold the log level. 
+		
+		time_t _currenttime = time(0);
+		localtime_r (&_currenttime, &_stime); //"localtime_r" works for POSIX-based systems. Use "localtime_s", otherwise. 
+		strftime (_buffer, sizeof(_buffer), "%Y-%m-%d %H:%M:%S", &_stime); //"strftime" is to format the date/time structure. 
+	   
+		switch(lvl){//To prepend the respective log level to each log message.  
+			case VerbosityLevel::Error:
+				strncpy(_vrb, "[ERROR]:", sizeof(_vrb));
```
lucasccordeiro: Shall we use the string class?

lucasccordeiro (changes requested):
@eyaget: thanks for implementing this issue. 

Overall, the code looks like C :-) can you try to make it look like C++? :-)

You can also find some comments to improve your PR.

lucasccordeiro:
@eyaget: Our CI complains about your code:

```
Codacy Static Code Analysis — This pull request quality could be better.
```

Can I ask you to take a look at the CI report?

eyaget (commented):
string.h added.

src/util/message.h
```diff
@@ -44,10 +45,57 @@ struct messaget
...
+				strncpy(_vrb, "[PROGRESS]:", sizeof(_vrb));
+				break;
+			case VerbosityLevel::Status:
+				strncpy(_vrb, "[STATUS]:", sizeof(_vrb));
+				break;
+			case VerbosityLevel::Debug:
+				strncpy(_vrb, "[DEBUG]:", sizeof(_vrb));
+				break;
+			default:
+				printf("Incorrect or no verbosity!\n");
```
rafaelsamenezes: Anything else can be defaulted to Debug.

rafaelsamenezes (changes requested):

src/util/message.h
eyaget: The length of the date/time value (including the terminating null character) is 20. That is why I declared a fixed size character array of 20 bytes.

src/util/message.h
eyaget: The size is now corrected to 12 bytes to make it more memory efficient. 

**Update**: My initial implementation was in `c` where a string is defined using array of characters. In C++, we can use the string class to define a string or string literal. So, this variable is now changed to a string.

src/util/message.h
eyaget: Done.

src/util/message.h
eyaget: Yes for the strncpy() function. BTW, I have replaced  strncpy() with strcpy() in the updated version as it is useless to copy a string literal with known length. In any case, strncpy() is not also a safe function :) 

**Update**: Yes, I used a string class now. The above explanation was for the standard string functions in `C`.

src/util/message.h
eyaget: Corrected, thanks

lucasccordeiro:
@eyaget : could you please rebase this branch with the master branch?

src/util/message.h
```diff
@@ -56,37 +55,37 @@ struct messaget
 		//========Start===============================================================================================
 		struct tm* time_struct; //A time structure consisting of 9 members, such as tm_sec, tm_min, tm_hour, tm_mday, tm_mon, tm_year, etc. 
 		char time_info[20]; //A variable to hold the date and time values.
-		char log_level[12]; //A variable to hold the log level.
+		const char *log_level;//[12]; //A variable to hold the log level.
```
mikhailramalho: Use string here instead.

src/util/message.h
```diff
@@ -44,10 +45,47 @@ struct messaget
...
+			case VerbosityLevel::Warning:				
+				log_level = "[WARNING]:";
+				break;
+			case VerbosityLevel::Progress:				
+				log_level = "[PROGRESS]:";
+				break;
+			case VerbosityLevel::Status:				
+				log_level = "[STATUS]:";
+				break;
+			default:
```
mikhailramalho: Please, implement all the enum values in the switch, so we can remove this default.

src/util/message.h
```diff
@@ -44,10 +45,47 @@ struct messaget
...
+			* It prepends a well formatted timestamp, file path, function name, line number and log level to each LOG message. 
+			* It is transparent to the existing logging function, i.e., log_message.
+			* TODO: It can be implemented using a separate function and invoked here and/or elsewhere when required. 		
+		*/		
+		//========Start===============================================================================================
+		struct tm* time_struct; //A time structure consisting of 9 members, such as tm_sec, tm_min, tm_hour, tm_mday, tm_mon, tm_year, etc. 
+		char time_info[20]; //A variable to hold the date and time values.
+		const char *log_level;//[12]; //A variable to hold the log level.
+		
+		time_t current_time = time(0);
```
mikhailramalho: use `nullptr` instead of `0`.

src/util/message.h
```diff
@@ -44,10 +45,47 @@ struct messaget
...
+    	     
+      	/** 
+		* Newly added code to address issue #886.
+			* It prepends a well formatted timestamp, file path, function name, line number and log level to each LOG message. 
+			* It is transparent to the existing logging function, i.e., log_message.
+			* TODO: It can be implemented using a separate function and invoked here and/or elsewhere when required. 		
+		*/		
+		//========Start===============================================================================================
+		struct tm* time_struct; //A time structure consisting of 9 members, such as tm_sec, tm_min, tm_hour, tm_mday, tm_mon, tm_year, etc. 
+		char time_info[20]; //A variable to hold the date and time values.
```
mikhailramalho: You can define a constexpr string for the format and use its size to allocate the right time size, and as an argument in the strftime call.

mikhailramalho (changes requested):
Thanks for the PR, some comments:
1. Please, remember to always run clang-format.
2. Remove the too-obvious comments. Things like: `/`/A variable to hold the date and time values.` or `//"strftime" is to format the date/time structure.` don't really add any value here.
3. Also remove the `==== start====` and `====end====` coments.

src/util/message.h
```diff
@@ -44,10 +45,47 @@ struct messaget
...
+				log_level = "[PROGRESS]:";
+				break;
+			case VerbosityLevel::Status:				
+				log_level = "[STATUS]:";
+				break;
+			default:
+				log_level = "[DEBUG]:";
+		}
+		
+		std::fprintf(f, "%s [%s] [%s:%d] %s ", time_info, __FILE__, __FUNCTION__, __LINE__, log_level); /* Prepends the timestamp, file name, file path,
```
fbrausse: Unless I'm mistaken, the expansion of the macros `__FILE__`, `__FUNCTION__` and `__LINE__` here will always default to `message.h`, `println()` and `84`. That's not helpful. With C++20 we could've used [std::source_location](https://en.cppreference.com/w/cpp/utility/source_location), but as ESBMC is C++17, I think the only choice we have is to make the log_(error|warning|...) things preprocessor macros instead of functions below, which insert those macros appopriately in a call to `println()`. Alternatively, drop this source-location info from the message.

As an additional comment (more a feature request): would it be of interest to disable this meta data via an envvar or cmd-line flag? I would like to do that in order to remove clutter from the terminal window when debugging issues.

fbrausse (changes requested):

lucasccordeiro:
@eyaget : Can I ask you whether you would have news about this PR?

eyaget:
@lucasccordeiro The regression test for this PR has kept failing. However, according to the CI reports, none of the errors are related to my PR. All the errors are from other implementations. Could you please help to check?

eyaget:
Sorry, it was by mistake that I closed the PR :)

lucasccordeiro:
@eyaget: The CI is failing for
your PR. Why do you think these errors are not related to your PR?

eyaget:
@lucasccordeiro According to the CI report, the errors are from other implementations. For example, the following are a few of the reported errors. Please correct me if I am wrong. I am actually new to such regression test. 
 

- /__w/esbmc/esbmc/regression/esbmc/github_270_1/main.c:7:5: error: second parameter of 'main' (argument array) must be of type 'char **'

- /__w/esbmc/esbmc/regression/esbmc/github_270_1/main.c:7:5: error: third parameter of 'main' (environment) must be of type 'char **'

- __w/esbmc/esbmc/regression/esbmc/to_union_0_fail/main.c:8:11: error: cast to union type from type 'char' not present in union
    FOO = (union foo) c;
    

- /__w/esbmc/esbmc/regression/esbmc/to_union_1_fail/main.c:5:23: error: use of undeclared identifier 'FLT_MAX'
    FOO = (union foo) FLT_MAX;

- /__w/esbmc/esbmc/regression/esbmc/to_union_2_fail/main.c:12:17: error: cast to union type from type 'char' not present in union
    int v = bar((union foo) x);

- /__w/esbmc/esbmc/regression/cbmc/01_cbmc_ExternExtern2/main.c:5:14: error: extern declaration of 'i' follows non-extern declaration
  extern int i;

- /__w/esbmc/esbmc/regression/cbmc/01_cbmc_Forward_Declaration2/main.c:4:7: error: variable has incomplete type 'List' (aka 'struct list')
        List l;

mikhailramalho:
I suggest rebasing so the tests can run again.

src/util/message.h
```diff
@@ -44,8 +46,51 @@ struct messaget
...
     static void println(FILE *f, VerbosityLevel lvl, Args &&...args)
     {
-      if(lvl == VerbosityLevel::Error)
-        fmt::print(f, "ERROR: ");
+      /** 
+		* Newly added code to address issue #886.
+		* It prepends a well formatted timestamp and log level to each LOG message. 
+		* It is transparent to the existing logging function, i.e., log_message.
+		* TODO: It can be implemented using a separate function and invoked here and/or elsewhere when necessary. 		
+		*/
```
mikhailramalho: Fix the indentation of this comment block. I suggest replacing it with `//` so clang-format doesn't think this is a doxygen comment.

src/util/message.h
```diff
@@ -44,8 +46,51 @@ struct messaget
...
+      struct tm *
+        timestamp_struct; //A time structure consisting of 9 members, such as tm_sec, tm_min, tm_hour, tm_mday, tm_mon, tm_year, etc.
```
mikhailramalho: We can get rid of this comment.

src/util/message.h
```diff
@@ -44,8 +46,51 @@ struct messaget
...
+      constexpr auto timestamp_format = "YYYY-MM-DD HH:MM:SS";
+      constexpr int timestamp_size =
+        std::char_traits<char>::length(timestamp_format) + 1;
```
mikhailramalho: I suggest replacing these two lines with:
```
constexpr std::string timestamp_format {"YYYY-MM-DD HH:MM:SS"};
constexpr int timestamp_size = timestamp_format.size();
```

src/util/message.h
```diff
@@ -44,8 +46,51 @@ struct messaget
...
+      constexpr auto timestamp_eval = "%Y-%m-%d %H:%M:%S";
```
mikhailramalho: I suggest replacing this with `constexpr std::string timestamp_eval {"%Y-%m-%d %H:%M:%S"};`

mikhailramalho (changes requested):
It's weird that we need both `timestamp_format` and `timestamp_eval`, can you follow the example in https://en.cppreference.com/w/cpp/chrono/c/strftime?

lucasccordeiro:
@eyaget : could you please rebase this branch with the master branch?

src/util/message.h
eyaget: The same issue as above.

src/util/message.h
eyaget: @mikhailramalho Thank you for your suggestions. However, the `constexpr std::string timestamp_format {"YYYY-MM-DD HH:MM:SS"};` definition returns the following compile-time error. 
- error: the type ‘const string’ {aka ‘const std::__cxx11::basic_string<char>’} of ‘constexpr’ variable ‘timestamp_format’ is not literal

Although `constexpr` is efficient, it is usually problematic unless its compile-time memory allocation/deallocation is handled properly. Hence, I used `const` instead in the revised code.

mikhailramalho: sure, that's fine, but my question still stands: why do we have two formats? Both `"YYYY-MM-DD HH:MM:SS"` and `"%Y-%m-%d %H:%M:%S"` are present in your code.

mikhailramalho:
Also, please keep the switch instead of the if/else chain

src/util/message.h
eyaget: @mikhailramalho I used the two formats for different purposes. The former is just a template of the actual appearance of the timestamp (e.g., 2023-03-08 21:15:00). This allows us to define the size of the timestamp and also to pass it as an argument in the `strftime` function. The latter is a character string specifying the format of conversion in the `strftime` function (refer https://en.cppreference.com/w/cpp/chrono/c/strftime). But, the latter does not allow us to compute the size of the timestamp.

fbrausse: If I would suggest this way to accomodate both your remarks and mine: check the return value of `strftime` and if it 0 retry with a larger buffer. This has the following benefits over the current version:
1. we're not making assumptions on the formatting/size, which leads to
2. the time could be output in a locale-dependent way later on without changes
3. the time format string could optionally be overridden via an environment variable, allowing my use case above

If this way is chosen, I would recommend to keep the computed size of the formatted result or the buffer itself in a (static) member in order to avoid recomputing it on every log message.

eyaget: @fbrausse Thanks for your suggestion. I think your assumption is only partially correct. Because, the function of `strftime` is to convert the date/time information from a given calendar time to a character string according to a particular format string specified, e.g., `"%Y-%m-%d %H:%M:%S"`. This time format is locale-independent. It always produces the time in the form of `YYYY-MM-DD HH:MM:SS`. Furthermore, the format string is hardcoded, so it would not be overridden by an environment variable. So, we can assume that size of the resulting timestamp is fixed. However, we still define it using the string `"YYYY-MM-DD HH:MM:SS"`. This is because, the size of a character string is dependent on the architecture of the target machine.

I agree with your idea of keeping the computed size as a static member.

lucasccordeiro:
@eyaget : could you also please squash your commits?

pmatos:
Thanks for this, and sorry for the very slow follow-up.

Timestamps and source location in log messages were subsequently implemented on `master` (commit `ffe4c269d3`, "Added timestamp and location to LOG messages"), gated behind the `--log-message` option and using `std::chrono` / `std::put_time`. That implementation also adds the file/line information and follows the idiomatic-C++ approach requested in the review here. The logging code this PR patches (`src/util/message.h`) has also been substantially refactored since.

As the feature is now on `master`, I'm closing this as superseded by `ffe4c269d3`. Thanks again for the contribution.

PR #919, opened 2023-03-05
Polymorphism Part I

kunjsong01:
This patch enabled more TCs in `polymorphism_bringup` test suite. 
Current pass rate: 12/36 TCs are passing.  

### Dev notes for basic polymorphism - Part I:

`clang_cpp_convert_vft` adds:  
* vtable:
  * vtable type symbols
  * vtable var symbols
* thunk function symbols
* vtable pointer symbols

`clang_cpp_convert_bind` deals with:
  * MemberExpr that refers to virtual/overriding function
 
`clang_cpp_adjust_code_gen` adds: 
  * vtable pointer initializations in constructors
  
### New regression suite: 
* regression/esbmc-cpp/polymorphism_bringup


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3335,3 +3339,27 @@ bool clang_c_convertert::is_field_global_storage(const clang::FieldDecl *field)
...
+  return false;
+}
+
+bool clang_c_convertert::get_vft_binding_expr(
+  const clang::MemberExpr &,
+  exprt &)
+{
+  log_error(
+    "MemberExpr call to virtual/overriding function cannot happen in C");
+  abort();
```
mikhailramalho, line 3362: do we need to abort here? why is this different from check_member_expr_virtual_overriding?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1258,3 +1297,10 @@ bool clang_cpp_convertert::need_new_object(const clang::Stmt *parentStmt)
 
   return false;
 }
+
+symbolt *clang_cpp_convertert::get_fd_symbol(const clang::FunctionDecl &fd)
+{
+  std::string id, name;
+  get_decl_name(fd, name, id);
+  return (context.find_symbol(id));
```
mikhailramalho, line 1306: what if this fails?

mikhailramalho (commented):
Great to see this taking shape!

My main concern is the `need_vptr_init`: How does the process work before without it? Do we really need it?

kunjsong01:
>Great to see this taking shape!

Hi @mikhailramalho, thanks very much for the review and I was happy to work on the clang-cpp-converter/adjuster which appears to be very extendible and maintainable for more complicated source code.  

>My main concern is the need_vptr_init: How does the process work before without it? Do we really need it?

The old frontend adds vtable_ptr initialization in two steps: 

1. Add an intermediate IR (see vptr_ir_intermediate.txt attached below) after "typechecking" the class `cpp_declarationt` (the old parse tree node).
https://github.com/esbmc/esbmc/blob/ee3923a1a56639f4ced0282bc50f0df9468367b2/src/cpp/cpp_typecheck_declaration.cpp#L112

[vptr_ir_intermediate.txt](https://github.com/esbmc/esbmc/files/10913824/vptr_ir_intermediate.txt)

2. Then in `typecheck_function_bodies` the old frontend converts the intermediate IR to a meaningful IR to represent the statement `this->vptr@BLAH = vtable::BLAH;` (see vptr_ir_final.txt attached below)
https://github.com/esbmc/esbmc/blob/ee3923a1a56639f4ced0282bc50f0df9468367b2/src/cpp/cpp_typecheck_function_bodies.cpp#L34

[vptr_ir_final.txt](https://github.com/esbmc/esbmc/files/10913850/vptr_ir_final.txt)

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: Yep, it would be really weird if a C program contains: 
```
|   |     |-CXXMemberCallExpr 0x5649f692df60 <col:10, col:26> 'int'
|   |     | `-MemberExpr 0x5649f692df30 <col:10, col:13> '<bound member function type>' ->do_something 0x5649f6922cd8
```
 where `do_something 0x5649f6922cd8` points to a virtual function: 
```
| |-CXXMethodDecl 0x5649f6922cd8 <line:11:3, col:47> col:15 used do_something 'int ()' 
```
or an overriding function:
``
| |-CXXMethodDecl 0x5649f692b890 <line:17:5, col:49> col:17 used do_something 'int ()' virtual
| | |-Overrides: [ 0x5649f6922cd8 Bird::do_something 'int ()' ]
``

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: This function gives max freedom to the caller. 
It can return nullptr if fd symbol does not exist. But that would be a decision for the caller to make...
something like: 
```cpp
if(get_fd_symbol)
{
  // do_something if fd symbol exists
}
else
{
  // do_something if fd symbol doesn't exist
}
```

sometimes we just need the `if` block.

lucasccordeiro (approved):
Overall, this PR looks good to me ;-)

Well done, @kunjsong01.

lucasccordeiro:
@kunjsong01 : could you please rebase this branch with the master branch?

kunjsong01:
> @kunjsong01 : could you please rebase this branch with the master branch?

@lucasccordeiro done

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -231,6 +232,29 @@ class clang_c_convertert
...
+   */
+  bool is_ConstructorOrDestructor(const clang::FunctionDecl &fd);
+
+  /*
+   * Check whether a member function call refers to
+   * a virtual/overriding method.
+   * For C, it always return false.
+   * For C++, see the overriding method in clang_cpp_convert module
+   */
+  virtual bool check_member_expr_virtual_overriding(const clang::Decl &decl);
```
fbrausse: The "check" and "expr" parts in the name of this method are not clear to me. Could we rename this function, maybe something like `is_virtual_method_overridden` or similar? Not clear whether that's the intent, though.

It would also be good if the comment explained what this method is supposed to do here, and not in some derived class - the interface is defined here. (This remark also applies to the comment on `get_vft_binding_expr` below.)

src/clang-cpp-frontend/clang_cpp_adjust.h
```diff
@@ -5,15 +5,24 @@
...
 class clang_cpp_adjust : public clang_c_adjust
 {
 public:
   explicit clang_cpp_adjust(contextt &_context);
   virtual ~clang_cpp_adjust() = default;
 
+  /**
+   * methods for symbol adjustment
+   */
+  void adjust_symbol(symbolt &symbol);
```
fbrausse, line 26: This comment is just repeating what's in the method's name. That is quite useless. Also it is just a single method.

src/clang-cpp-frontend/clang_cpp_adjust.h
```diff
@@ -32,6 +41,37 @@ class clang_cpp_adjust : public clang_c_adjust
   void adjust_side_effect(side_effect_exprt &expr) override;
   void adjust_new(exprt &expr);
   void adjust_struct_method_call(member_exprt &expr);
+
+  /**
+   * methods for implicit GOTO code generation
+   */
+  void gen_vptr_initializations(symbolt &symbol);
```
fbrausse, line 50: How is "implicit GOTO code generation" related to "vptr initializations"? Could you be more concrete about the functionality here? Again, it's commenting just on a single method.

src/clang-cpp-frontend/clang_cpp_adjust.h
```diff
@@ -32,6 +41,37 @@ class clang_cpp_adjust : public clang_c_adjust
...
+  /*
+   * generate vptr initialization code for constructor:
+   *  this->VptrBLAH@BLAH = $vtable::BLAH
```
fbrausse: What do the multiple `BLAH` stand for in this comment? Do they refer to a function's identifier, some counter for uniqueness, the constructor's class's name or what?

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
```diff
@@ -0,0 +1,152 @@
+#include <clang-cpp-frontend/clang_cpp_adjust.h>
+#include <util/std_expr.h>
+
+void clang_cpp_adjust::gen_vptr_initializations(symbolt &symbol)
+{
+  /*
+   * This function traverses the vptr components of the correponding class type
+   * in which the constructor is decleared. Then complete the initializations
```
fbrausse: typo: "declared".

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
```diff
@@ -0,0 +1,152 @@
...
+   * The "to_code_*" conversion functions have done some sanity checks for us:
+   *  - a ctor symbol shall contain a function body modelled by code_blockt
+   *  - symbol should be of code type
+   */
+  code_typet &ctor_type = to_code_type(symbol.type);
+  code_blockt &ctor_body = to_code_block(to_code(symbol.value));
+
+  /*
+   *  vptr initializations shall be done in ctor
+   *  TODO: For the time being, we just add vptr int code in ctors.
```
fbrausse: typo: "init".

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
```diff
@@ -0,0 +1,152 @@
...
+  //typet vtable_type = symbol_typet(comp.type().subtype().id());
+  new_code.type() = comp.type();
+
+  // 2. LHS: generate the member pointer dereference expression
+  exprt lhs_expr;
+  gen_vptr_init_lhs(comp, lhs_expr, ctor_type);
+
+  // 3. RHS: generate the address of the target virtual pointer struct
+  exprt rhs_expr;
+  gen_vptr_init_rhs(comp, rhs_expr, ctor_type);
```
fbrausse: Why don't these functions `gen_vptr_init_[lr]hs` return the `exprt` they're supposed to compute? Return-by-ref-argument looks wrong in C++17.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -316,22 +330,18 @@ bool clang_cpp_convertert::get_struct_union_class_methods(
...
-  {
-    log_debug(
-      "Class {} has {} bases, {} virtual bases",
-      cxxrd->getNameAsString(),
-      cxxrd->getNumBases(),
-      cxxrd->getNumVBases());
+  /*
+   * Order of converting methods:
+   *  1. convert virtual methods. We also need to add:
+   *    a). add virtual table type
```
fbrausse: remove duplicated "add" please.

src/clang-cpp-frontend/clang_cpp_convert.cpp
fbrausse: In that case the documentation should say that `nullptr` is a valid result.

src/clang-cpp-frontend/clang_cpp_convert.h
```diff
@@ -148,6 +161,250 @@ class clang_cpp_convertert : public clang_c_convertert
...
+   * additional annotations for virtual or overriding methods
+   */
+  bool annotate_virtual_overriding_methods(
+    const clang::CXXMethodDecl *md,
+    struct_typet::componentt &comp);
+  /*
+   * Check the existence of virtual table type symbol.
+   * If it exists, return its pointer. Otherwise, return nullptr.
+   */
+  symbolt *check_vtable_type_symbol_existence(struct_typet &type);
```
fbrausse: I'd personally recommend against prefixing predicates with `check_`. Something like `is_` or `get_` is more common.

Also, why is the `type` parameter given by non-const reference? Is it to be modified?

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+void clang_cpp_convertert::add_vptr(
+  const clang::CXXMethodDecl *md,
+  struct_typet &type)
+{
+  /*
+   * We model the virtual pointer as a `component` to the parent class' type.
+   * This will be the vptr pointing to the vtable that contains the overriden functions.
+   *
+   * Vptr has the name in the form of `tag-BLAH@vtable_pointer`.
```
fbrausse: what is `BLAH` here? It might be easier on the eye if it was `TYPE_NAME` or similar.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -0,0 +1,585 @@
...
+  new_compo.set_name(thunk_func_symb.id);
+  type.methods().push_back(new_compo);
+}
+
+void clang_cpp_convertert::setup_vtable_struct_variables(
+  const clang::CXXRecordDecl *cxxrd,
+  const struct_typet &type)
+{
+  /*
+   * Recall that the virtual function table (VFT) is just a switch table, and
```
fbrausse: Does "switch table" mean "jump table"? If so, I would drop this analogy, "struct of function pointers" is more clear, since a jump table does not necessarily (actually, almost never) contain entry addresses of functions but usually points somewhere well inside the body of a given function.

(Also applies to usage of "switch table" in comments below.)

fbrausse (changes requested):
LGTM, I'd like to request mostly cosmetic changes.

src/clang-c-frontend/clang_c_convert.h
kunjsong01: Good point. 
Done. Renamed to `perform_virtual_dispatch`... hope this name is more meaningful and intuitive.

src/clang-cpp-frontend/clang_cpp_adjust.h
kunjsong01: It may look useless at first glance in this code snippet, however it's actually indicating "the functions below are used for symbol adjustment" and to be consistent with the structure of `src/clang-c-frontend/clang_c_adjust.h` (i.e. The `adjust_*` functions are grouped together in clang_cpp_adjust class. ). 
It would make a lot sense in `src/clang-cpp-frontend/clang_cpp_adjust.h`.

src/clang-cpp-frontend/clang_cpp_adjust.h
kunjsong01: > How is "implicit GOTO code generation" related to "vptr initializations"? Could you be more concrete about the functionality here?

Done, improved the comment.

src/clang-cpp-frontend/clang_cpp_adjust.h
kunjsong01: Good point. I've added comment to explain the meaning of 'BLAH' - constructor's class's name

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
kunjsong01: Done.

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
kunjsong01: Done

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
kunjsong01: To avoid creating temporaries/calling copy constructor stuff? 
I'm not an expert in C++17... so what could we do to make it more elegant? @fbrausse 
It appears that we use return-by-ref-argument quite often in the frontend.

src/clang-cpp-frontend/clang_cpp_convert.cpp
kunjsong01: Done

src/clang-cpp-frontend/clang_cpp_convert.h
kunjsong01: > Also, why is the type parameter given by non-const reference? Is it to be modified?

yes

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
kunjsong01: Fixed

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
kunjsong01: Done, just used `a struct of function pointers` to be precise.

src/clang-cpp-frontend/clang_cpp_adjust.h
fbrausse: Well, if "the functions below are used for symbol adjustment" is meant to be indicated, maybe it should just say that. :)
*edit* sorry for nitpicking, to me this comment just reads as the documentation for this specific method.

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
fbrausse: I see, you're copying the style it was written in before. In C++17 we have mandated [return value optimization](https://en.cppreference.com/w/cpp/language/copy_elision). That means if you have
```c++
class Big { char huge_amount_of_data[1234567]; };
Big create_Big() {
  Big b;
  /* maybe populate b.huge_amount_of_data */
  return b;
}
```
and call `create_Big()` somewhere to store the result, no copy of the internal `b` is being performed. Instead, `b` is constructed at the target, where the return value is assigned. There should be no performance problem (actually, it should be faster) when returning the `exprt` directly.

src/clang-cpp-frontend/clang_cpp_convert.h
fbrausse: > yes

Strange, the implementation
```c++
symbolt *
clang_cpp_convertert::check_vtable_type_symbol_existence(struct_typet &type)
{
  irep_idt vt_name = vtable_type_prefix + tag_prefix + type.tag().as_string();
  return context.find_symbol(vt_name);
}
```
does not seem to modify it and neither does the documentation hint to that.

kunjsong01: Spot on
Apologies for the confusing answer I provided earlier. It's probably a copy-and-past mistake I've made.
Now fixed. I've just added `const`.

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
kunjsong01: Thanks! 
Done. Fixed to use the benefit C++17 RVO provided.

fbrausse (approved):
Many thanks, @kunjsong01.

PR #927, opened 2023-03-14
Polymorphism Part IV

kunjsong01:
This patch added support for diamond problem, and enabled TC7. 
Total pass rate: 36/45 in `regression/esbmc-cpp/polymorphism_bringup` suite.

### Dev notes: 
1. Fixed recursive method to get overriden_map to avoid generating duplicate thunk functions in a more complicated scenario (multi-level polymorphism, diamond problem)
2. Fixed recursive method to get bases_map to avoid getting duplicate components/methods from base classes in a more complicated scenario (multi-level polymorphism, diamond problem)
3. Also update some TC descriptions in `regression/esbmc-cpp/polymorphism_bringup` test suite.


lucasccordeiro (approved):
LGTM.

rafaelsamenezes (approved):

PR #930, opened 2023-03-14
VFT fix

kunjsong01:
This patch enabled 2 more test cases - vmd_defined_outside_class and vmd_defined_outside_class_fail
Total pass rate: 38/45 in `regression/esbmc-cpp/polymorphism_bringup` suite. 

### Dev notes: 
- Fixed the condition in get_function method to get the type for virtual/overriding method. 
-- Because  we need the type of virtual/overriding method to figure out the number of arguments when creating thunk functions and vtable.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.h
+++ b/src/clang-c-frontend/clang_c_convert.h
@@ -253,7 +253,7 @@ protected:
   virtual bool perform_virtual_dispatch(const clang::MemberExpr &member);
 
   /*
-   * Function to for virtual function table dynamic binding for "->" operator
+   * Function to get the ESBMC IR representing a virtual function table dynamic binding for "->" operator
    * Turning
    *  x->F
    * into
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):
LGTM.

src/clang-c-frontend/clang_c_convert.h
```diff
@@ -267,6 +267,17 @@ class clang_c_convertert
...
+  /*
+   * Function to check whether a clang::FunctionDec represents a
+   * clang::CXXMethodDecl that is virtual OR overrides another function
+   *
+   * Params:
+   *  - fd: a clang::FunctionDec we are currently dealing with
+   *
+   * For C, it always return false.
+   */
+  virtual bool is_fd_virtual_or_overriding(const clang::FunctionDecl &fd);
```
rafaelsamenezes, line 280: Shouldn't this function be local to the CPP frontend then?

rafaelsamenezes (changes requested):

src/clang-c-frontend/clang_c_convert.h
kunjsong01: The `clang-c-convert::get_function` needs this function to continue for virtual/overriding method as we need its type. 

I also considered moving everything local tp CPP frontend.... But if this function was local to CPP frontend, we would need to add some duplicate code in `clang_cpp_convertert::get_function` that does the same job as the code in `clang_c_convertert::get_function`. I reckon adding one additional condition in clang_c_converter is better than adding some duplicate code in the  overriding function in clang_cpp_converter.

rafaelsamenezes: > we would need to add some duplicate code in clang_cpp_convertert::get_function that does the same job as the code in clang_c_convertert::get_function. 

You could refactor the duplicated part into another function that should belong to the C frontend. If get_function from C works differently from CPP then it is expected that they have different specializations.

kunjsong01: Somehow I missed this comment from @rafaelsamenezes in my email inbox. 

>You could refactor the duplicated part into another function that should belong to the C frontend.

True, it makes sense. This appear to be a better structure. I'll see what I can do when working on this function in future patch.

PR #937, opened 2023-03-20
[library] Add Cuda Operational Model

XLiZHI:
Library source from https://github.com/ssvlab/esbmc-gpu
Some bug fixes for Clang


XLiZHI:
This check may have failed for some other reason. Could you please re-run this check?

lucasccordeiro:
Yes, sure. It is running now.

fbrausse (commented):
This PR touches a lot of the existing C++ operational models for unclear reasons. I recommend splitting those changes off of this PR and add the CUDA headers separately. Also I believe that the CUDA headers are not meant to sit in the same directory as the C++ stdlib headers. Would it be possible to find another location for the CUDA headers?

lucasccordeiro:
@XLiZHI : I fully agree with @fbrausse comments above.

kunjsong01:
@XLiZHI I agree with @fbrausse comment. 

This PR also contains some system CUDA libraries, which shouldn't be in our repository. 

When splitting this PR, you may want to group some realistic and relatable changes in a sequence of smaller PRs, working upward the include tree towards passing a test case (usually start with a low-hanging fruit). We have some general (sort-of) guidlines and a few example PRs: 

- https://github.com/esbmc/esbmc/wiki/Guidelines-for-Fixing-Operational-Models-(OM)-in-ESBMC
- https://github.com/esbmc/esbmc/pull/943
- https://github.com/esbmc/esbmc/pull/944
- https://github.com/esbmc/esbmc/pull/948 

Hope it helps.

XLiZHI:
My apologise for late reply.

@fbrausse @lucasccordeiro @kunjsong01  Thank you very much for your advice, my modification of some already existing c++ headers is due to our clang front end reporting errors on these files during compilation, could you please let me know if I made a mistake?

@kunjsong01 thanks for the examples, what do you mean by "system cuda library"?

kunjsong01:
>@kunjsong01 thanks for the examples, what do you mean by "system cuda library"?

[src/cpp/library/device_functions_decls.h](https://github.com/esbmc/esbmc/pull/937/files#diff-3b68c1a39404e4f95357b6b353f3e4aff7206b018b42f89e9c4706034da1c803)

PR #947, opened 2023-03-29
Fixed class ids

kunjsong01:
This fix is needed to support exception OM (see https://github.com/esbmc/esbmc/pull/948).

### Dev notes: 
- Fixed class ids in various places for VFT and MemberExpr binding
   * Avoid using `tag_prefix + getNameAsString()` because it may not give us the full name, which results in symbol id mismatch, e.g. : 
Looking for 
```
virtual_table::tag-std::Bird@tag-std::Penguin
``` 
but symbol table has
```
virtual_table::tag-std::Bird@tag-Penguin
```


lucasccordeiro (approved):

fbrausse (approved):
Aside from the [comment](https://github.com/esbmc/esbmc/pull/948#discussion_r1152358963) about hard-coding `--memlimit 14000000 --timeout 900` in the regression tests in #948, LGTM.

PR #948, opened 2023-03-29
Fixed and added OM sanity check for exception

kunjsong01:
Continuation of https://github.com/esbmc/esbmc/pull/944

This PR addresses level-4 dependency `exception` OM as shown in the following include tree:
Screenshot 2023-03-28 at 20 30 26

### Dev notes:
1. Reviewed exception OM:
   * added reference links
   * fixed PARSE ERRORs reported by clang
2. Added the TC for OM sanity check

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/exception/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/exception/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+-I /__w/esbmc/esbmc/src/cpp/library --memlimit 14000000 --timeout 900
+
+^VERIFICATION SUCCESSFUL$
--- a/src/cpp/library/exception
+++ b/src/cpp/library/exception
@@ -2,8 +2,12 @@
  Module:
 
  Author: Felipe Rodrigues
+ Created in: September 2012
+ Updated in: March 2023 by Kunjian Song
 
- Date: September 2012
+ Ref:
+ https://en.cppreference.com/w/cpp/header/exception
+ https://cplusplus.com/reference/exception/
 
  \*******************************************************************/
 
@@ -19,18 +23,27 @@ const char* message;
 typedef void (*terminate_handler)();
 typedef void (*unexpected_handler)();
 
-bool uncaught_exception() throw ();
+bool uncaught_exception() throw();
 
 // Default terminate function: calls abort
-void default_terminate() throw ()
+void default_terminate() throw()
 {
   __ESBMC_assert(0, "Aborted");
   // Allow now further execution.
   __ESBMC_assume(0);
 }
 
+// Default terminate function for unexpected: calls abort
+// A quick workaround model for "throw"
+void default_terminate_unexpected_exception() throw()
+{
+  __ESBMC_assert(0, "Aborted - unexpected exceptions");
+  // Allow now further execution.
+  __ESBMC_assume(0);
+}
+
 // Default unexpected function: calls terminate
-void default_unexpected() throw ()
+void default_unexpected() throw()
 {
   default_terminate();
 }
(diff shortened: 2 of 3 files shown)


lucasccordeiro (approved):

regression/esbmc-cpp/OM_sanity_checks/exception/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
+main.cpp
+-I /__w/esbmc/esbmc/src/cpp/library --memlimit 14000000 --timeout 900
```
fbrausse, line 3: Not clear why the parameters `--memlimit 14000000 --timeout 900` matter for these regression tests. Appearently a lot of the C++ tests have the exact same ones. In general, I wouldn't hard-code those but rather rely on other means to set these limits, e.g. the envvar `ESBMC_OPTS` or ctest's `--timeout`, etc.

src/cpp/library/exception
```diff
@@ -19,41 +23,54 @@ const char* message;
...
+void default_terminate() throw()
 {
   __ESBMC_assert(0, "Aborted");
   // Allow now further execution.
   __ESBMC_assume(0);
 }
 
+// Default terminate function for unexpected: calls abort
+// A quick workaround model for "throw"
+void default_terminate_unexpected_exception() throw()
```
fbrausse, line 38: The name of this function pollutes the `std` namespace which we have no control over. Could it be renamed to start with `__ESBMC_` as the other internally used symbols do or alternatively be put into a namespace like `::__ESBMC`?

src/cpp/library/exception
```diff
@@ -19,41 +23,54 @@ const char* message;
...
 
 // Model for unexpected function: calls the current
 // unexpected handler function
 void unexpected()
 {
   unexpected_pf();
-  throw;
+  // TODO: For the time being, we use a terminate function as a workaround.
+  //   probably should rethrow the exception and just use throw.
+  default_terminate_unexpected_exception();
```
fbrausse, line 73: I don't understand this change. You're removing `throw;` but describe it as what should probably be done.

fbrausse (commented):
I have a few comments. It would be good to know what the concrete problems are you're trying to solve:
> Working towards passing regression/esbmc-cpp/stream/istream_get_1/main.cpp

kunjsong01:
>It would be good to know what the concrete problems are you're trying to solve:

This patch fixes the parse errors when including the exception OM. 
[parse_errors.txt](https://github.com/esbmc/esbmc/files/11109301/parse_errors.txt)

The new test case added in OM_sanity_check test suite aims to provide basic sanity checks for future code changes in an OM, making sure it's still "includable" without parse errors.

regression/esbmc-cpp/OM_sanity_checks/exception/test.desc
kunjsong01: IIRC these parameters are originally from the old C++ test cases.

src/cpp/library/exception
kunjsong01: `throw` would give: 
```
ERROR: Conversion of unsupported clang expr: "CXXThrowExpr" to expression
CXXThrowExpr 0x560340bc53b8 <library/exception:73:3> 'void'

ERROR: CONVERSION ERROR
```

Since our frontend currently doesn't support exception handling (which is another major feature but not necessarily needed for our verification goal), this is a quick workaround to avoid this error and deal with unexpected exception in a fairly decent way.

src/cpp/library/exception
kunjsong01: Good point. I added this function based on the existing one `default_terminate`. I also had some concerns this new function, as well as `default_terminate`, might contaminate std. 
But if `default_terminate` can be safely used to verify the existing test cases without giving any problem regarding namespace, then we probably have a good reason (empirically) to keep this code structure and naming conventions.

PR #949, opened 2023-03-31
Adding support for the solidity constructor and multiple contracts

ChenfengWei0:
Adding support for the solidity constructor and multiple contracts
- The solidity frontend now accepts a single file with multiple contracts inside
	- However, if there are functions with the same name in different contracts, only the last one will be verified
- The solidity frontend now supports constructor by:
	- Parsing the keyword `constructor` as a void function.
	- An implicit constructor will be added to the symbol table if no explicit constructor function is found.
	- The keyword `new` and the declaration for the contract type variable are now supported. i.e.
		- No params: `Base x = new Base();`
		- With params: `Base x = new Base(2);`
		- Noted that these two declarations will result in different symbol table structures, which is consistent with the logic of the `clang_cpp` frontend
	- Member function call is now supported.  i.e. `x.caller();`


src/solidity-frontend/solidity_convert.h
```diff
@@ -147,7 +163,9 @@ class solidity_convertert
     exprt &dest);
   bool convert_string_literal(std::string the_value, exprt &dest);
 
-  static constexpr const char *mode = "C";
+  static constexpr const char *mode = "C++";
+  std::string prefix = "tag-";
+  //exprt new_object_expr;
```
lucasccordeiro: Do we still need this line?

src/solidity-frontend/solidity_grammar.h
```diff
@@ -218,6 +221,10 @@ enum ExpressionT
   // Solidity does NOT provide such rule
   IndexAccess,
 
+  NewExpression,
```
lucasccordeiro, line 227: Can you add some documentation here?

src/solidity-frontend/solidity_grammar.h
```diff
@@ -218,6 +221,10 @@ enum ExpressionT
...
+
+  MemberCallClass,
```
lucasccordeiro, line 232: Can you add some documentation here?

src/solidity-frontend/solidity_grammar.h
```diff
@@ -264,6 +271,16 @@ enum ImplicitCastTypeT
 ImplicitCastTypeT get_implicit_cast_type_t(std::string cast);
 const char *implicit_cast_type_to_str(ImplicitCastTypeT type);
 
+enum VisibilityT
```
lucasccordeiro, line 281: Can you add some documentation here?

src/solidity-frontend/solidity_language.cpp
```diff
@@ -9,8 +9,8 @@ CC_DIAGNOSTIC_POP()
 #include <solidity-frontend/solidity_language.h>
 #include <solidity-frontend/solidity_convert.h>
 #include <clang-c-frontend/clang_c_main.h>
-#include <clang-c-frontend/clang_c_adjust.h>
-#include <clang-c-frontend/clang_c_convert.h>
+#include <clang-cpp-frontend/clang_cpp_adjust.h>
```
lucasccordeiro, line 12: Why did we move from clang-c to clang-cpp?

lucasccordeiro: We should provide some explanation here.

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -708,4 +726,32 @@ const char *implicit_cast_type_to_str(ImplicitCastTypeT type)
   }
   }
 }
-}; // namespace SolidityGrammar
+
+VisibilityT get_access_t(const nlohmann::json &ast_node)
```
lucasccordeiro, line 730: Can you add some documentation here?

lucasccordeiro (changes requested):

src/solidity-frontend/solidity_convert.h
ChenfengWei0: Sorry, I forgot to delete this comment.

src/solidity-frontend/solidity_grammar.h
ChenfengWei0: Yes, I will add some documentation to explain.

src/solidity-frontend/solidity_language.cpp
ChenfengWei0: The general idea of the migration from C to C++ is that we treat each contract as a C++ **class**. Since clang-c did not support the feature of class or "construct a temporary object" or constructor call, therefore we change it to C++. The `mode` info of each symbol will be change from `C(0)` to `C++(1)`.

As for the reason for the code change here, it's because we need the adjustment in `clang_cpp_adjust_expr.cpp` which will first adjust the "temporary_object" expression (i.e. Base x = new Base()) and then feed it to `clang_c_adjust`. The `clang_c_adjust` cannot adjust this expression by itself.

I will add some documentation in the file to explain

ChenfengWei0: But some changes might not be necessary. I will take a look and do some testing to see what must be changed.

lucasccordeiro:
Many thanks, Chenfeng!Sent from my iPhoneOn 1 Apr 2023, at 11:43, Chenfeng Wei ***@***.***> wrote:﻿
@ChenfengWei0 commented on this pull request.



In src/solidity-frontend/solidity_language.cpp:
> @@ -9,8 +9,8 @@ CC_DIAGNOSTIC_POP()
 #include <solidity-frontend/solidity_language.h>
 #include <solidity-frontend/solidity_convert.h>
 #include <clang-c-frontend/clang_c_main.h>
-#include <clang-c-frontend/clang_c_adjust.h>
-#include <clang-c-frontend/clang_c_convert.h>
+#include <clang-cpp-frontend/clang_cpp_adjust.h>

But some changes might not be necessary. I will take a look and do some testing to see what must be changed.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -295,12 +304,173 @@ bool solidity_convertert::get_var_decl(
...
+    {
+    case SolidityGrammar::ContractBodyElementT::StateVarDecl:
+    {
+      if(get_struct_class_fields(*itr, t))
+        return true;
+      break;
+    }
+    case SolidityGrammar::ContractBodyElementT::FunctionDef:
+    {
+      // if(get_struct_class_method(*itr, t))
```
kunjsong01: remove?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -494,6 +672,17 @@ bool solidity_convertert::get_block(
       if(get_statement(stmt_kv.value(), statement))
         return true;
 
+      // TODO: FIX ME.
+      // We are now mannually doing the adjustment for NewExpression
+      // by spilting into two expression.
```
kunjsong01: expressions

src/solidity-frontend/solidity_convert.h
```diff
@@ -7,6 +7,7 @@
 #include <util/context.h>
 #include <util/namespace.h>
 #include <util/std_types.h>
+#include <clang-c-frontend/padding.h>
```
kunjsong01: Do we need this header?

kunjsong01 (approved):
LGTM. 
just some comments about cosmetic changes.

src/solidity-frontend/solidity_convert.h
ChenfengWei0: The `add_padding` in `padding.h` function is used to deal with the alignment in struct or union (i.e. bitfield). It occurs in the `clang_c_convert.cpp`.

I'm not quite sure if the `struct` and `union` need to be packed and aligned as in C++, but I would say I shouldn't include this "padding" in this commit. Thanks for the suggestion.

lucasccordeiro (approved):

Issue #931, opened 2023-03-15
[CUDA]  ERROR: Declaration has an empty name: CXXConstructorDecl


XLiZHI:
esbmc datar.cu -I ~/libraries/

```CUDA
#include <cuda.h>
#define tid threadIdx.x
#define N 2

__device__ inline void inlined(int *A, int offset) {
int temp = A[tid + offset];
A[tid] += temp;
}

__global__ void inline_test(int *A, int offset) {
inlined(A, offset);
}

int main( ) {
int *a;
int *dev_a;
int size = N*sizeof(int);

cudaMalloc((void**)&dev_a, size);

a = (int*)malloc(N*size);

for (int i = 0; i < N; i++)
a[i] = i;

cudaMemcpy(dev_a,a,size, cudaMemcpyHostToDevice);

ESBMC_verify_kernel_intt(inline_test, 1, N, dev_a, 2);

cudaMemcpy(a,dev_a,size,cudaMemcpyDeviceToHost);
free(a);
cudaFree(dev_a);
return 0;
}
```
https://github.com/esbmc/esbmc/commit/75cdc93b24e1c221258c627d245fd84c8b5169b8

here is the version of libraries.

kunjsong01:
We support `case clang::Decl::CXXConstructor` in `clang_cpp_convertert::get_decl`. 

@XLiZHI 
Need more information to reproduce the error to investigate: 
1. Version of ESBMC
2. https://github.com/esbmc/esbmc/commit/75cdc93b24e1c221258c627d245fd84c8b5169b8 - this commit doesn't seem to belong to any branch on esbmc repository. Would it be possible to make a patch for the changes you've done in the commit and attach it to this issue?

XLiZHI:
Hi @kunjsong01 
Thanks for your reply.
1. The latest ESBMC: ESBMC version 7.1.0 64-bit x86_64 linux (Commits on Mar 14, 2023)
2. https://github.com/XLiZHI/esbmc/tree/cuda

> We support `case clang::Decl::CXXConstructor` in `clang_cpp_convertert::get_decl`.
> 
> @XLiZHI Need more information to reproduce the error to investigate:
> 
> 1. Version of ESBMC
[...]

PR #933, opened 2023-03-15
Ignore unamed implicit trivial ctor

kunjsong01:
A quick fix for https://github.com/esbmc/esbmc/issues/931 

See comment https://github.com/esbmc/esbmc/issues/931#issuecomment-1470541439 for error description

PR #956, opened 2023-04-04
[headers]Named implicit trivial ctor

XLiZHI:
Fixed error with uname implicit trivial constructor from Clang C++

https://github.com/esbmc/esbmc/issues/931#issuecomment-1470541439

Added a regression test for pthread.h


lucasccordeiro (approved):

kunjsong01 (changes requested):
Since the OM fix in this PR is an alternative to the clang-cpp-frontend patch PR https://github.com/esbmc/esbmc/pull/933 for issue https://github.com/esbmc/esbmc/issues/931,  I believe we first need to revert https://github.com/esbmc/esbmc/pull/933 then test whether the OM fix in this PR resolves the issue.

we want to make sure the new test case added in this PR still passes after reverting https://github.com/esbmc/esbmc/pull/933. 

Also, `esbmc-cpp/OM_sanity_checks` contains only test cases for C++ OMs from `src/cpp/library`. `pthread.h` is not specific to C++. I reckon `regression/esbmc` is the place to add the corresponding test cases.

kunjsong01 (approved):
LGTM

lucasccordeiro (approved):

lucasccordeiro:
> Since the OM fix in this PR is an alternative to the clang-cpp-frontend patch PR https://github.com/esbmc/esbmc/pull/933 for issue https://github.com/esbmc/esbmc/issues/931, I believe we first need to revert https://github.com/esbmc/esbmc/pull/933 then test whether the OM fix in this PR resolves the issue.

@kunjsong01: Should ESBMC crash because of a wrong OM? We should either fail in the parser or reject the AST. The fix to the OM is okay, but I don't think we should remove the previous fix, which prevents the crash in ESBMC.

kunjsong01:
> > Since the OM fix in this PR is an alternative to the clang-cpp-frontend patch PR #933 for issue #931, I believe we first need to revert #933 then test whether the OM fix in this PR resolves the issue.
> 
> @kunjsong01: Should ESBMC crash because of a wrong OM? We should either fail in the parser or reject the AST. The fix to the OM is okay, but I don't think we should remove the previous fix, which prevents the crash in ESBMC.

@lucasccordeiro ESBMC didn't crash in issue https://github.com/esbmc/esbmc/issues/931. ESBMC indeed failed in the parser and rejeted the AST, also reported the correct error message "ERROR: Declaration has an empty name: CXXConstructorDecl".

lucasccordeiro:
Thanks for this clarification, @kunjsong01.

PR #958, opened 2023-04-05
[OM] definitions, cassert and stdexcept

kunjsong01:
Continuation of https://github.com/esbmc/esbmc/pull/954

Fixed the error: 
```
ESBMC version 7.1.0 64-bit x86_64 linux
Parsing
In file included from stdexcept_sanity_check.cpp:1:
library/stdexcept:70:11: error: exception specification of overriding function is more lax than base version
  virtual ~logic_error() {
          ^
library/exception:90:11: note: overridden virtual function is here
  virtual ~exception() throw() {}
          ^
ERROR: PARSING ERROR
```

The destructor `~exception()` should not have exception specifier (c.f. [std::exception::~exception reference](https://en.cppreference.com/w/cpp/error/exception/%7Eexception))

This PR addresses a few level-3 dependency OMs as shown in the following include tree:
Screenshot 2023-04-05 at 12 35 08

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/definitions/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/definitions/main.cpp
@@ -1,0 +1,4 @@
+#include "definitions.h"
+int main () {
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cassert/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cassert/main.cpp
@@ -1,0 +1,4 @@
+#include <cassert>
+int main () {
+  return 0;
+}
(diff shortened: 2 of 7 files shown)


lucasccordeiro (approved):

regression/esbmc-cpp/OM_sanity_checks/definitions/main.cpp
```diff
@@ -0,0 +1,4 @@
+#include "definitions.h"
```
fbrausse, line 1: I don't understand what this file is testing. On a regular C++ implementation, this should fail to compile because `"definitions.h"` is not in the include-path.

kunjsong01: This test case includes `definitions.h` in the same way as other C++ OMs, e.g.
https://github.com/esbmc/esbmc/blob/master/src/cpp/library/algorithm
https://github.com/esbmc/esbmc/blob/master/src/cpp/library/cmath

`definitions.h` is in `-I <path>/src/cpp/library`.

fbrausse: Sure, but `"definitions.h"` is an implementation detail and the non-existence of it should not constitute a regression. If anything, it should not exist (without a prefix like `__ESBMC` or similar). Anyhow, since it's merged we can discuss files like this when cleaning up the ESBMC-provided C++ headers.

PR #960, opened 2023-04-05
[OM] cstring

kunjsong01:
Continuation of https://github.com/esbmc/esbmc/pull/958

N.B This PR needs rebasing when https://github.com/esbmc/esbmc/pull/958 is merged

Fixed the error:
```
ESBMC version 7.1.0 64-bit x86_64 linux
Parsing
In file included from cstdlib_sanity_check.cpp:1:
In file included from master_library/cstdlib:18:
master_library/cstring:141:12: warning: comparison between NULL and non-pointer ('char' and NULL) [-Wnull-arithmetic]
        while (*s != NULL) { //until the end of string
               ~~ ^  ~~~~
master_library/cstring:277:18: error: arithmetic on pointers to void
        if (dest - src >= n) {
            ~~~~ ^ ~~~
In file included from cstdlib_sanity_check.cpp:1:
master_library/cstdlib:115:9: warning: address of stack memory associated with local variable 'tmp' returned [-Wreturn-stack-address]
        return tmp;
               ^~~
ERROR: PARSING ERROR
```

This PR addresses a few level-3 dependency OM `cstring` as shown in the following include tree:

fix_2

### DIFF
--- a/src/cpp/library/cstring
+++ b/src/cpp/library/cstring
@@ -13,312 +13,312 @@
 #include "definitions.h"
 
 char *strcpy(char *d, const char *s) {
-	int i;
+  int i;
 
-	for (i = 0; s[i] != 0; i++) {
-		d[i] = s[i];
-	}
+  for (i = 0; s[i] != 0; i++) {
+    d[i] = s[i];
+  }
 
-	d[i] = '\0';
+  d[i] = '\0';
 
-	return d;
+  return d;
 }
 
 char *strncpy(char *d, const char *s, size_t n) {
 
-	size_t i=0;
-	char ch;
-	bool end;
+  size_t i=0;
+  char ch;
+  bool end;
 
-	for(end=0; i<n; i++)
-	{
-		ch=end?0:s[i];
-		d[i]=ch;
-		end=end || ch==(char)0;
-	}
-	return d;
+  for(end=0; i<n; i++)
+  {
+    ch=end?0:s[i];
+    d[i]=ch;
+    end=end || ch==(char)0;
+  }
+  return d;
 }
 
 char *strcat(char d[], const char s[]) {
-	int i, e;
+  int i, e;
 
-	for (e = 0; d[e] != '\0'; e++)
-		;
+  for (e = 0; d[e] != '\0'; e++)
+    ;
 
-	for (i = 0; s[i] != '\0'; i++) {
-		d[e + i] = s[i];
-	}
+  for (i = 0; s[i] != '\0'; i++) {
+    d[e + i] = s[i];
+  }
 
-	d[e + i] = '\0';
+  d[e + i] = '\0';
 
-	return d;
+  return d;
 }
 
 char *strncat(char d[], const char s[], size_t n) {
-	size_t i, e;
+  size_t i, e;
 
-	for (e = 0; d[e] != '\0'; e++)
-		;
+  for (e = 0; d[e] != '\0'; e++)
+    ;
 
-	for (i = 0; s[i] != '\0' && i < n; i++) {
-		d[e + i] = s[i];
-	}
+  for (i = 0; s[i] != '\0' && i < n; i++) {
+    d[e + i] = s[i];
+  }
 
-	d[e + i] = '\0';
+  d[e + i] = '\0';
 
-	return d;
+  return d;
 }
 
 int strcmp(const char *s1, const char *s2) {
-	int i;
+  int i;
 
-	for (i = 0;; i++) {
-		if (s1[i] == s2[i]) {
-			if (s1[i] == '\0')
-				return 0;
-			continue;
-		}
-		if (s1[i] == '\0')
-			return -1;
-		if (s2[i] == '\0')
-			return 1;
-		if (s1[i] < s2[i])
-			return -1;
-		else
-			return 1;
-	}
+  for (i = 0;; i++) {
+    if (s1[i] == s2[i]) {
+      if (s1[i] == '\0')
+        return 0;
+      continue;
+    }
+    if (s1[i] == '\0')
+      return -1;
+    if (s2[i] == '\0')
+      return 1;
+    if (s1[i] < s2[i])
+      return -1;
+    else
+      return 1;
+  }
 }
 
 int strncmp(const char *s1, const char *s2, size_t n) {
-	int i;
+  int i;
 
-	for (i = 0; i < n; i++) {
-		if (s1[i] == s2[i]) {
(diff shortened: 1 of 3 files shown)


src/cpp/library/cstring
```diff
@@ -263,18 +263,18 @@ char *strstr(const char *s1, const char *s2) {
 
 void *memcpy(void *dst, const void *src, size_t n) {
 	__ESBMC_HIDE:
-        char *cdst = static_cast<char *>(dst);
-        const char *csrc = static_cast<const char *>(src);
-        for (size_t i = 0; i < n; i++)
+  char *cdst = static_cast<char *>(dst);
```
mikhailramalho, line 266: fix code style.

src/cpp/library/cstring
```diff
@@ -263,18 +263,18 @@ char *strstr(const char *s1, const char *s2) {
...
 		cdst[i] = csrc[i];
 	return dst;
 }
 
 void *memmove(void *dest, const void *src, size_t n) {
 	__ESBMC_HIDE:
-        char *cdst = static_cast<char *>(dest);
-        const char *csrc = static_cast<const char *>(src);
-        if (dest - src >= n) {
+  char *cdst = static_cast<char *>(dest);
```
mikhailramalho, line 275: fix code style.

mikhailramalho (approved):
LGTM, just needs some code-style fixes.

mikhailramalho:
@kunjsong01, most of the code in these headers is duplicated (poorly as this PR shows) from our models.

Have you tried to delete the duplicated code and only add the function definitions to the headers? My guess is that esbmc would pick the implementation from our models.

lucasccordeiro (approved):
LGTM. 

@kunjsong01: could you please fix the code style issues?

src/cpp/library/cstring
kunjsong01: Done. 
This file had lots of tab indents, which may have a weird look with different tab-space settings in various text editors. So I just retab-ed the entire file in vim.

kunjsong01:
> @kunjsong01, most of the code in these headers is duplicated (poorly as this PR shows) from our models.
> 
> Have you tried to delete the duplicated code and only add the function definitions to the headers? My guess is that esbmc would pick the implementation from our models.

@mikhailramalho 
Sorry, just a quick confirm: 
Do you actually mean "only add the function definitions declarations to the headers"? something like: 
```cpp
// in C++ model for `cstring` 
char *strcpy(char *dst, const char *src);
int strcmp(const char *s1, const char *s2);
```

I experimented with this idea using the test case below. Because the API implementations here appear to be duplicates of our C models as in src/c2goto/library/string.c.  

```cpp
#include <cstring>
#include <cassert>
int main ()
{
  char str1[3];
  char str2[3];
  strcpy(str1,"Hi");
  strcpy(str2,"Hi");
  int ret;
  ret = strcmp(str1, str2);
  //assert(ret != 0); // FAIL
  assert(ret == 0);
  return 0;
}
```

ESBMC gives incorrect verification results if we just keep the declarations.  I didn't really investigate this issue. Not sure what the problem is.... 
But with the duplicated API implementations in `cstring` model, ESBMC gives correct results.

mikhailramalho:
> > @kunjsong01, most of the code in these headers is duplicated (poorly as this PR shows) from our models.
> > Have you tried to delete the duplicated code and only add the function definitions to the headers? My guess is that esbmc would pick the implementation from our models.
> 
> @mikhailramalho Sorry, just a quick confirm: Do you actually mean "only add the function definitions declarations to the headers"? something like:
> 
> ```c++
> // in C++ model for `cstring` 
> char *strcpy(char *dst, const char *src);
> int strcmp(const char *s1, const char *s2);
> ```
> 
> I experimented with this idea using the test case below. Because the API implementations here appear to be duplicates of our C models as in src/c2goto/library/string.c.
> 
> ```c++
> #include <cstring>
> #include <cassert>
> int main ()
> {
>   char str1[3];
>   char str2[3];
>   strcpy(str1,"Hi");
>   strcpy(str2,"Hi");
>   int ret;
>   ret = strcmp(str1, str2);
>   //assert(ret != 0); // FAIL
>   assert(ret == 0);
>   return 0;
> }
> ```
> 
> ESBMC gives incorrect verification results if we just keep the declarations. I didn't really investigate this issue. Not sure what the problem is.... But with the duplicated API implementations in `cstring` model, ESBMC gives correct results.

Bummer, maybe the mangled name is different and we can't pull the symbols from our models?

Can you do the following and test again:

1. remove all code from cstring and simply add #include <string.h> (it should look like the climits file)
2. delete string.h

?

If this works, we should delete all *.h from our model and make all the c* header simply include the standard c headers. Hopefully this can pull the symbols from our c model and remove all duplicated code.

kunjsong01:
>Can you do the following and test again:
>
>  1. remove all code from cstring and simply add #include <string.h> (it should look like the climits file)
>  2.  delete string.h

@mikhailramalho Thanks! It works, and passed the simple test case in https://github.com/esbmc/esbmc/pull/960#issuecomment-1498067434 

I used the following code in `cstring` OM: 
```cpp
#ifndef STL_CSTRING
#define STL_CSTRING
#include <string.h>
#endif
```

Looking at the GOTO code, it's definitely using our C operaitonal model in `src/c2goto/library/string.c` as confirmed by the line number and file name in each comment: 
```cpp
215 strcpy (c:@F@strcpy):
216         // 88 file string.c line 29 function strcpy
217         // Labels: __ESBMC_HIDE
218         signed char * cp;
219         // 89 file string.c line 29 function strcpy
220         cp=dst;
221         // 90 file string.c line 30 function strcpy
222      1: tmp$1=cp;
223         // 91 file string.c line 30 function strcpy
224         cp=cp + 1;
225         // 92 file string.c line 30 function strcpy
226         tmp$2=src;
...
233         // 96 file string.c line 30 function strcpy
234         GOTO 1
235         // 97 file string.c line 32 function strcpy
236      2: RETURN: dst
237         // 98 file string.c line 33 function strcpy
238         END_FUNCTION // strcpy
239 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

>If this works, we should delete all .h from our model and make all the c header simply include the standard c headers. Hopefully this can pull the symbols from our c model and remove all duplicated code.

I concur. This appears to be the way to go. I've raised an issue https://github.com/esbmc/esbmc/issues/965, and put it in my backlog.

lucasccordeiro (approved):

PR #962, opened 2023-04-11
[DO NOT MERGE] An under-approximated operational model for the scanf function 

lucasccordeiro:
The scanf function is unsafe since it disregards the length of the string typed by the user; so, it can accept a string longer than the array size that stores that respective string.

This PR adds an assertion that will fail with the message ```The `scanf' function is unsafe to use.```. 

So, this PR will allow one ESBMC user to advance their studies on using ESBMC to find software vulnerabilities in open-source applications.


lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/976.

PR #799, opened 2022-07-08
Update BUILDING.md

fbrausse:
There was a typo in Boolector's min-version, sorry. I've added "latest" as recommendation for clang/boost.

PR #967, opened 2023-04-12
Fix for Dockerfile smell DL3008

grosa1:
Hi!
The Dockerfile placed at "scripts/ESBMC-Linux.dockerfile" contains the best practice violation [DL3008](https://github.com/hadolint/hadolint/wiki/DL3008) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3008 occurs when the version pinning for the installed packages with apt is not specified. This could lead to unexpected behavior when building the Dockerfile.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, we use a heuristic approach that selects the most probable version tag for a given apt package corresponding to the latest version at the current date. The package versions are retrieved from the Canonical Launchpad APIs.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/scripts/ESBMC-Linux.dockerfile
+++ b/scripts/ESBMC-Linux.dockerfile
@@ -1,5 +1,5 @@
 FROM ubuntu:20.04
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y clang-tidy python-is-python3 csmith python3 git ccache unzip wget curl libcsmith-dev gperf libgmp-dev cmake bison flex gcc-multilib linux-libc-dev libboost-all-dev ninja-build python3-setuptools libtinfo-dev pkg-config python3-pip && pip install toml
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y clang-tidy=1:10.0-* python-is-python3=3.8.* csmith=2.3.* python3=3.8.* git=1:2.25.* ccache=3.7.* unzip=6.0-* wget=1.20.* curl=7.68.* libcsmith-dev=2.3.* gperf=3.1-* libgmp-dev=2:6.2.* cmake=3.16.* bison=2:3.5.* flex=2.6.* gcc-multilib=4:9.3.* linux-libc-dev=5.4.* libboost-all-dev=1.71.* ninja-build=1.10.* python3-setuptools=45.2.* libtinfo-dev=6.2-* pkg-config=0.29.* python3-pip=20.0.* && pip install toml 
 WORKDIR /workspace
 # Boolector 3.2.2
 RUN git clone --depth=1 --branch=3.2.2 https://github.com/boolector/boolector && cd boolector && ./contrib/setup-lingeling.sh && ./contrib/setup-btor2tools.sh && ./configure.sh --prefix $PWD/../boolector-release && cd build && make -j4 && make install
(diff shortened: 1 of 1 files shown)


fbrausse:
I am missing a bit more technical information in order to review this PR: What relevance does hadolint's best practice guide have for ESBMC and in what way is it universal enough to justify "a heuristic approach" for solving the perceived problem? What's the issue with keeping the dependency version numbers floating as maintained by the distribution underlying the docker image?

grosa1:
Hi @fbrausse,
thank you for the response. Most of the hadolint rules are based on the official guidelines, while some other are proposed by software developers. 
What I am doing with my tool is trying to fix automatically the rules detected by hadolint. In this specific case, a suggestion is to pin the version for apt packages in order to avoid future issues due to major changes for that specific packages. 
For example, assuming that the Dockerfile installs git, the tool (i) searches for the packages in the launchpad APIs, (ii) selects the latest version currently associated with git, and (iii) fixes the major and minor version numbers, putting a wildcard character for the patch (`<major>.<minor>.*`) in order to get always the latest patch. In the end, a test is executed installing the pinned package using the same base image to verify if it is working.

fbrausse:
Hi @grosa1, ESBMC supports a range of different version of its dependency and also is regularly tested with different versions of Clang, SMT solvers, etc. Hence, only minimum version requirements for the dependencies are specified (https://github.com/esbmc/esbmc/blob/master/BUILDING.md#dependency-overview). Not even recommending (let alone requiring) a specific version was a choice made earlier in light of Windows support (see https://github.com/esbmc/esbmc/pull/799#discussion_r916727599 and follow-up comments).

The reason you hinted at
> in order to avoid future issues due to major changes for that specific packages

for me has another side: What happens if we want to upgrade the distro version in the docker file? Would we need to un-pin versions an re-run your tool to obtain new heuristically chosen pinned versions for the new distro?

I personally don't see the benefit of pinning versions and for me it rather means we should investigate and possibly fix any incompatibility issues arising from newer (or older) dependency versions that are in use - on a case-by-case basis.

grosa1:
Hi @fbrausse,
thank you for the response. 

> for me has another side: What happens if we want to upgrade the distro version in the docker file? Would we need to un-pin versions an re-run your tool to obtain new heuristically chosen pinned versions for the new distro?

Yes, in the case of an upgrade you need to run pin the package versions again to get the updated version numbers. 

> I personally don't see the benefit of pinning versions and for me it rather means we should investigate and possibly fix any incompatibility issues arising from newer (or older) dependency versions that are in use - on a case-by-case basis.

That's a good point that for sure could be investigated. Maybe the version pinning could be done for the packages that could be at risk. What do you think?

fbrausse:
> Maybe the version pinning could be done for the packages that could be at risk. What do you think?

Maybe. I cannot tell though which of our packages could be considered at risk of breaking ESBMC builds. In the past year or two I mainly observe Windows dependencies breaking (actually, not because of API, but because of availability via vcpkg). The Linux dependencies haven't caused any problems I am aware of. @rafaelsamenezes Do you have thoughts about which of our dependencies regularly cause breakage with the Ubuntu docker image?

rafaelsamenezes:
> Do you have thoughts about which of our dependencies regularly cause breakage with the Ubuntu docker image?

Not that I can remember. Only compiler updates have caused trouble due to "warnings are errors".

fbrausse:
I see, thanks Rafael. In this case I would vote against the changes made by this PR - unless there are some arguments for version pinning that I missed.

fbrausse:
After reviewing all of https://github.com/search?q=%22Fix+for+Dockerfile+smell+DL3008%22&type=issues I believe this to be a semi-automatically generated PR not fixing any concrete issue of concrete people using ESBMC. Closing. @grosa1 please feel free to open an issue or another PR fixing one if it's about a concrete use case. Many thanks.

Issue #955, opened 2023-04-04
[Operational Model] check the scanf implementation 


lucasccordeiro:
ESBMC gives VERIFICATION SUCCESSFUL for this program:

```C
#include <stdio.h>
int main( int argc, char *argv[])
{       
        char reg_name[12];
        printf("Enter your username:");
        scanf("%s", reg_name); //Buffer overflow here
        printf("The program is now registered to %s.\n",reg_name);
        return 0;
}
```

lucasccordeiro:
@rafaelsamenezes : Can I ask you to check this issue since you're the author of the scanf OM?

lucasccordeiro:
This issue is related to this PR https://github.com/esbmc/esbmc/pull/872.

ChenfengWei0:
Ok, I will work on it.

ChenfengWei0:
The CBMC generates same result as ESBMC.
```sh
$ cbmc test.cpp
CBMC version 5.80.0 (cbmc-5.80.0) 64-bit x86_64 linux
Parsing 4.cpp
Converting
Type-checking 4
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 8 object bits, 56 offset bits (default)
Starting Bounded Model Checking
**** WARNING: no body for function scanf
Runtime Symex: 0.00139503s
size of program expression: 68 steps
simple slicing removed 0 assignments
Generated 0 VCC(s), 0 remaining after simplification
Runtime Postprocess Equation: 3.7952e-05s
VERIFICATION SUCCESSFUL
```

lucasccordeiro:
Thanks, @ChenfengWei0.

We should provide `VERIFICATION FAILED`. There is a buffer overflow here.

lucasccordeiro:
Note that CBMC reports:

```
**** WARNING: no body for function scanf
```

lucasccordeiro:
@ChenfengWei0 : you can also take a look at this issue https://github.com/esbmc/esbmc/issues/508.

ChenfengWei0:
A short update of my progress and thoughts:

- I think we should add a `goto_trace_stept::typet INPUT`  for `scanf/fscanf`, i.e.
```c++
  typedef enum
  {
    ASSIGNMENT,
    ASSUME,
    ASSERT,
    OUTPUT,
    INPUT,  // added. CBMC also has this type
    SKIP,
    RENUMBER
  } typet;
```
[...]

PR #976, opened 2023-04-19
Add overflow checks for scanf/fscanf

ChenfengWei0:
Add overflow checks for `scanf`/`fscanf` (to address issue #955 )

- check for buffer overflow. e.g. `test.c`:
```c++
#include <stdio.h>
int main( int argc, char *argv[])
{
        char reg_name[12];
        printf("Enter your username:");
        scanf("%s", reg_name); //Buffer overflow here
        scanf("%12s",  reg_name); // also a buffer overflow
        printf("The program is now registered to %s.\n",reg_name);
        return 0;
}
```

command: `esbmc test.c --overflow-check --64`

result:
```sh
Counterexample:
Enter your username:

State 4 file 2.c line 31 function main thread 0
Violated property:
  file 2.c line 31 function main
  buffer overflow on c:@F@__ESBMC_scanf
  0


VERIFICATION FAILED
```

- checks for integer overflow i.e.
```c++
int main()
{
    int x;
    scanf("%d", &x); // input 12345678901
    printf("%d", x); // output -539222987  overflow 
    
    int y;
    scanf("%10d", &y); // input 12345678901
    printf("%d", y); // output 1234567890
}
```

- fail the assertion if the format identifiers do not match with the arguments. e.g.
```c++
#include <stdio.h>
#include <assert.h>
int main(int argc, char *argv[])
{
    int m, n = 42;
    fscanf(stdin,
           "%10d",
           &m,
           &n);  // warning: too many arguments for format [-Wformat-extra-args]
    assert(n == 42);
}
```

result
```sh
esbmc: /root/workspace/ESBMC_Commit/esbmc/src/goto-programs/goto_check.cpp:299: void goto_checkt::input_overflow_check(const expr2tc &, const locationt &): 
Assertion `(limits.size() == arg_names.size()) && "the format specifiers do not match with the arguments"' failed.
```

- TODO：
  - check for float i.e. `scanf("%1.2f", &fl_var);`
  - symex_trace for input (`goto_trace_stept::INPUT`)
  - (?) symbolic execution for cutting down the input stream based on the length limit

### DIFF
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -127,11 +127,7 @@ void goto_checkt::float_overflow_check(
   const guardt &guard,
   const locationt &loc)
 {
-  if(!enable_overflow_check && !enable_ub_shift_check)
-    return;
-
-  // Don't check shift right
-  if(is_lshr2t(expr) || is_ashr2t(expr))
+  if(!enable_overflow_check)
     return;
 
   // First, check type.
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -23,7 +23,7 @@ public:
       disable_unlimited_scanf_check(
         options.get_bool_option("no-unlimited-scanf-check")),
       enable_overflow_check(options.get_bool_option("overflow-check")),
-      enable_shift_ub_check(options.get_bool_option("shift-ub-check")),
+      enable_ub_shift_check(options.get_bool_option("ub-shift-check")),
       enable_nan_check(options.get_bool_option("nan-check"))
   {
   }
@@ -90,7 +90,7 @@ protected:
   bool disable_pointer_relation_check;
   bool disable_unlimited_scanf_check;
   bool enable_overflow_check;
-  bool enable_shift_ub_check;
+  bool enable_ub_shift_check;
   bool enable_nan_check;
 };
 
@@ -127,7 +127,11 @@ void goto_checkt::float_overflow_check(
   const guardt &guard,
   const locationt &loc)
 {
-  if(!enable_overflow_check)
+  if(!enable_overflow_check && !enable_ub_shift_check)
+    return;
+
+  // Don't check shift right
+  if(is_lshr2t(expr) || is_ashr2t(expr))
     return;
 
   // First, check type.
@@ -206,7 +210,11 @@ void goto_checkt::overflow_check(
   const guardt &guard,
   const locationt &loc)
 {
-  if(!enable_overflow_check)
+  if(!enable_overflow_check && !enable_ub_shift_check)
+    return;
+
+  // Don't check shift right
+  if(is_lshr2t(expr) || is_ashr2t(expr))
     return;
 
   // First, check type.
@@ -387,24 +395,30 @@ void goto_checkt::shift_check(
   const guardt &guard,
   const locationt &loc)
 {
-  if(!enable_shift_ub_check)
+  if(!enable_ub_shift_check)
     return;
 
-  assert(is_shl2t(expr));
+  auto right_op = (*expr->get_sub_expr(1));
 
-  auto bv_type = (*expr->get_sub_expr(0))->type;
-  auto constant = (*expr->get_sub_expr(1));
-
-  expr2tc zero = gen_zero(constant->type);
+  expr2tc zero = gen_zero(right_op->type);
   assert(!is_nil_expr(zero));
 
-  greaterthanequal2tc const_gte_zero(constant, zero);
+  greaterthanequal2tc right_op_non_negative(right_op, zero);
 
-  expr2tc bv_type_size(
-    new constant_int2t(bv_type, BigInt(bv_type->get_width())));
-  lessthanequal2tc const_lte_type_size(constant, bv_type_size);
+  auto left_op = (*expr->get_sub_expr(0));
+  auto left_op_type = left_op->type;
+  expr2tc left_op_type_size(
+    new constant_int2t(left_op_type, BigInt(left_op_type->get_width())));
 
-  and2tc ub_check(const_gte_zero, const_lte_type_size);
+  lessthan2tc right_op_size_check(right_op, left_op_type_size);
+
+  and2tc ub_check(right_op_non_negative, right_op_size_check);
+
+  if(is_shl2t(expr))
+  {
+    greaterthanequal2tc left_op_non_negative(left_op, zero);
+    ub_check = and2tc(ub_check, left_op_non_negative);
+  }
 
   add_guarded_claim(
     ub_check,
@@ -688,6 +702,8 @@ void goto_checkt::check_rec(
     /* fallthrough */
 
   case expr2t::shl_id:
+  case expr2t::ashr_id:
+  case expr2t::lshr_id:
     shift_check(expr, guard, loc);
     /* fallthrough */
   case expr2t::neg_id:
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -300,7 +300,7 @@ void goto_checkt::input_overflow_check(
 
   // do checks
   bool buf_overflow = false;
-  for(int i = 0; i < arg_names.size(); i++)
+  for(long unsigned int i = 0; i < arg_names.size(); i++)
   {
     const symbolt &arg = *ns.lookup(arg_names.at(i));
     const irep_idt type_id = arg.type.id();
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -269,7 +269,7 @@ void goto_checkt::input_overflow_check(
 
   for(std::string tmp_str = ""; pos < fmt.length(); pos++)
   {
-    if(fmt[pos] == '%' and fmt[pos + 1] != '.')
+    if(fmt[pos] == '%' && fmt[pos + 1] != '.')
     {
       pos++;
       while(std::isdigit(fmt[pos]))
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -330,18 +330,25 @@ void goto_checkt::input_overflow_check(
       switch(stoi(width))
       {
       case 8:
+      {
         if(stoi(limits.at(i)) > 3)
           buf_overflow = true;
+      }
       case 16:
+      {
         if(stoi(limits.at(i)) > 5)
           buf_overflow = true;
+      }
       case 32:
+      {
         if(stoi(limits.at(i)) > 10)
           buf_overflow = true;
-
+      }
       case 64:
+      {
         if(stoi(limits.at(i)) > 19)
           buf_overflow = true;
+      }
       default:
         break;
       }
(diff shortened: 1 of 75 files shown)


src/goto-programs/goto_check.cpp
```diff
@@ -231,6 +234,147 @@ void goto_checkt::overflow_check(
...
+  const std::string fmt =
+    get_string_argument(func_call.operands[fmt_idx]).as_string();
+
+  // obtain the length limits in the format string
+  long unsigned int pos = 0;
+  std::vector<std::string> limits;
+
+  for(std::string tmp_str = ""; pos < fmt.length(); pos++)
+  {
+    if(fmt[pos] == '%' and fmt[pos + 1] != '.')
```
kunjsong01: replacing "and" with "&&" would probrably resolve the windows CI failure.

lucasccordeiro:
@ChenfengWei0 : can I ask you whether you still plan to implement new features in this PR?

ChenfengWei0:
> @ChenfengWei0 : can I ask you whether you still plan to implement new features in this PR?

I intend to implement the new features in another PR.

lucasccordeiro (approved):

mikhailramalho (changes requested):
I see you went with the INPUT approach and, while I don't mind following that, we don't need the goto_check implementation here: can you implement scanf (and fscanf) to follow the printf/output approach? 

You can even reuse the code in printf_formatter.cpp to parse the format string, instead of writing your own.

ChenfengWei0:
> I see you went with the INPUT approach and, while I don't mind following that, we don't need the goto_check implementation here: can you implement scanf (and fscanf) to follow the printf/output approach?
> 
> You can even reuse the code in printf_formatter.cpp to parse the format string, instead of writing your own.

Thanks for the suggestions. I will work on it.

ChenfengWei0:
> I see you went with the INPUT approach and, while I don't mind following that, we don't need the goto_check implementation here: can you implement scanf (and fscanf) to follow the printf/output approach?
> 
> You can even reuse the code in printf_formatter.cpp to parse the format string, instead of writing your own.

Do you suggest I should move the input checking to symex_scanf/fscanf (as a part of the symbolic execution)?

mikhailramalho:
yeah, just like printf: you'll need a new data type to represent the input (code_input2tc?).

Btw, how does cbmc does this check?

ChenfengWei0:
> yeah, just like printf: you'll need a new data type to represent the input (code_input2tc?).
> 
> Btw, how does cbmc does this check?

CBMC does not have a symbolic execution function for scanf. It shows warnings:
`**** WARNING: no body for function scanf `

mikhailramalho:
when does the INPUT instruction is generated then?

ChenfengWei0:
> when does the INPUT instruction is generated then?

Sorry I just made a mistake.
- The old version CBMC (e.g. 5.10) will generates the warning above. 
- But the newest cbmc-5.81 do have a `do_scanf` `do_fscanf` and `symex_input` function. And it will do the verification. The INPUT is for the `symex_input`. However, it does not check the scanf buffer overflow.

src/goto-programs/goto_check.cpp
```diff
@@ -231,6 +234,147 @@ void goto_checkt::overflow_check(
...
+
+  // obtain the format string
+  const std::string fmt =
+    get_string_argument(func_call.operands[fmt_idx]).as_string();
+
+  // obtain the length limits in the format string
+  long unsigned int pos = 0;
+  std::vector<std::string> limits;
+
+  for(std::string tmp_str = ""; pos < fmt.length(); pos++)
```
rafaelsamenezes, line 278: This can probably be generalized into its own class (e.g. printf). C uses format strings in many places. This could be left to another PR though

src/goto-programs/goto_check.cpp
```diff
@@ -231,6 +234,147 @@ void goto_checkt::overflow_check(
...
+  for(long unsigned int i = 0; i < arg_names.size(); i++)
+  {
+    const symbolt &arg = *ns.lookup(arg_names.at(i));
+    const irep_idt type_id = arg.type.id();
+    std::string width;
+
+    // if no length limits, then we treat it as a buffer overflow
+    if(limits.at(i) == "INF")
+    {
+      buf_overflow = true;
```
rafaelsamenezes, line 323: This should be optional (e.g. `--no-scanf-overflow-check`)

src/goto-symex/builtin_functions.cpp
```diff
@@ -355,6 +355,51 @@ void goto_symext::symex_printf(const expr2tc &, const expr2tc &rhs)
     cur_state->guard.as_expr(), cur_state->source, fmt.as_string(), args);
 }
 
+void goto_symext::symex_scanf(
```
rafaelsamenezes: This could be merged into fscanf function

rafaelsamenezes (changes requested):
Can you better  explain the need of the INPUT instruction?

lucasccordeiro:
@ChenfengWei0: once you address @rafaelsamenezes's comments, could you please run this PR over the SV-COMP benchmarks?

src/goto-symex/builtin_functions.cpp
ChenfengWei0: Thanks for the suggestions. I think I should merge them to a function like `symex_input()` .

ChenfengWei0:
> @ChenfengWei0: once you address @rafaelsamenezes's comments, could you please run this PR over the SV-COMP benchmarks?

Sure

src/goto-programs/goto_check.cpp
ChenfengWei0: Yes. I think we should have a `scanf_formattert`, just like `printf_formattert`.

src/goto-programs/goto_check.cpp
```diff
@@ -235,6 +241,151 @@ void goto_checkt::overflow_check(
...
+  for(long unsigned int i = 0; i < arg_names.size(); i++)
+  {
+    const symbolt &arg = *ns.lookup(arg_names.at(i));
+    const irep_idt type_id = arg.type.id();
+    std::string width;
+
+    // if no length limits, then we treat it as a buffer overflow
+    if(limits.at(i) == "INF")
+    {
+      if(disable_unlimited_scanf_check)
```
rafaelsamenezes, line 320: maybe move this to the beginning of this function/

rafaelsamenezes (changes requested):
I would suggest adding KNOWNBUG for the unsupported cases. Also, I think you are going to miss out errors over dynamic memory such as:

```c
void *ptr = malloc(sizeof(char));
scanf("%d", ptr);
```

You can solve it at the goto_check by using the value-set analysis. However, you will probably be better off doing at symex (as @mikhailramalho suggested).

Note that only your overflow check should be KNOWNBUG. Tests that use the variables of scanf should be working fine.

I am fine merging it as it is since its simple enough to deactivate and there are users that need it. Just add more dynamic memory testcases.

src/goto-programs/goto_check.cpp
ChenfengWei0: The code behind this if-statements would expect the `limit[i]` as a "integer", as we will call `stoi(limit[i])`. So I think we should always have a function to handle the "INF" case.

ChenfengWei0:
> Note that only your overflow check should be KNOWNBUG. Tests that use the variables of scanf should be working fine.

I have just added a test case:
```c++
#include <stdlib.h>
int main(int argc, char *argv[])
{
    int* ptr = (int*) malloc(sizeof(int));
    scanf("%12d", ptr);  // input 12345678901 overflow
    printf("%d",*ptr);   // output --539222987
}
```
when verifying with `--overflow-check`, ESBMC will not report the buffer overflow. So I think this is a KNOWBUG.

ChenfengWei0:
Update Summary:
- the `symex_scanf` and `symex_fscanf` have been merged in to `symex_input`.
- codes that related to `symex_tracet::input` have been deleted.
- add an option `--no-unlimited-scanf-check` which means we do not do overflow checks for `scanf/fscanf` with unlimited character width. E.g. 
```c++
int x;
scanf("%d", &x); // not a verification failed if  "--no-unlimited-scanf-check" enabled
```

---

An explanation of the changes in the `regression/esbmc-unix/10_bicycle_04/main.c`. The original codes are:
```c++
int input;
do{
...
input = 0;
scanf("%d", &input);

switch (input) {
  case 0: ...
  case 1: ...
  ...
}

}while(1);
```
- Originally, there was no function body for `scanf`, therefore the switch statement will always jump to `case 0`, leading to the "VERIFICATION SUCCESSFUL"
- However, after adding the symbolic execution for `scanf`, this regression test will keep running until reaching the timeout limits.  
  - During the symbolic execution, the variable `input` will be set as non-deterministic (`symex::nondet`) and then initialised by `replace_nondet()`
- If we change the `scanf` to `fscanf`, i.e. `fscanf(stdin, "%d", &input);` and use the previous ESBMC (before this PR), we will still encounter the same problem.
- If we have codes like : 
```c++
// scanf("%d", &input);

switch (rand()) {
  case 0: ...
  case 1: ...
  ...
}
```
and we will still encounter the same problem.

So for now I just set `switch (0)` to make it work as it used to.

-- 
Edit: fixed

src/goto-programs/goto_check.cpp
rafaelsamenezes: I meant the `if(disable_unlimited_scanf_check)`

We usually put those at the beginning of the check functions (or before they are called)

ChenfengWei0: Sorry, I am a bit confused.  

For now, we have 
```c++
if(enable_overflow_check)
  input_overflow_check(i.code, loc);
```

which is before the function called, 

and

```c++
if(limits.at(i) == "INF")
{
  if(disable_unlimited_scanf_check)
```

Do you suggest we should have a specific option like `--no-scanf-overflow-check` instead of using `--overflow-check`? Do we still need the `--no-unlimited-scanf-check`?

rafaelsamenezes (approved):

lucasccordeiro:
@ChenfengWei0: Can I ask you whether you would have the SV-COMP results of this PR?

ChenfengWei0:
> @ChenfengWei0: Can I ask you whether you would have the SV-COMP results of this PR?

I ran the Benchexec(SV-COMP) 11 hours ago and it's still in progress.

ChenfengWei0:
> @ChenfengWei0: Can I ask you whether you would have the SV-COMP results of this PR?

```
  correct:           14550
    correct true:     8319
    correct false:    6231
  incorrect:            96
    incorrect true:     47
    incorrect false:    49
  unknown:           11813
  Score:             20581 (max: 43043)
```

lucasccordeiro:
The results of the master are:

```
   correct:            14549
     correct true:     8316
     correct false:    6233
   incorrect:            96
     incorrect true:     47
    incorrect false:     49
   unknown:            11814
   Score:              20577 (max: 43043)
```

lucasccordeiro:
@ChenfengWei0: I will merge this PR for now.

Can you open an issue with the following tasks to improve this implementation?

lucasccordeiro (approved):

ChenfengWei0:
> @ChenfengWei0: I will merge this PR for now.
> 
> Can you open an issue with the following tasks to improve this implementation?

I just created a new issue #984

PR #977, opened 2023-04-19
[ub-shift-check] Add new checks and improve tests

### LINKED ISSUE #580: Develop a check for undefined behaviour
This issue is to remind us to develop a new option in ESBMC (e.g., `--undef-behaviour-check`) to check for undefined behaviour. In particular, this check would add an `assert(const >= 0)` or `assert(const <= type-bit-width)` statement for shift operations. It should be something similar to the `--overflow-check` (available in the `goto_check.cpp` file).

brcfarias:
- Add ub check on shift right operations

- Improve regression tests

- The C language standard specifies the behavior of the shift operators as below

image

### DIFF
--- a/regression/esbmc/00_bitshift_04/main.c
+++ b/regression/esbmc/00_bitshift_04/main.c
@@ -3,5 +3,10 @@
 int main() {
   int n = 256 << 1;
   assert(n == 512);
+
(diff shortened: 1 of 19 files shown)


lucasccordeiro (approved):
LGTM.

Issue #832, opened 2022-08-01
[C++ OM] Fix parse errors in string OM


kunjsong01:
Failed to include `string` OM due to 54x Clang errors reported: 
[parse_errors.txt](https://github.com/esbmc/esbmc/files/11304728/parse_errors.txt)
based on ESBMC (commit 1991d6fac6d92e07b202033eb498a078bd6df62b) clang invocation. 

Error category:
1. incorrect number of parameters for operator overloading as a member function. 
2. Default argument erorrs
3. local variable/label redefinition errors
4. 'i/ostream' type errors
5. non-const lvalue reference binding to a temporary
6. redefinition of default argument
7. global operator overloading errors, e.g.

kunjsong01:
Closed via https://github.com/esbmc/esbmc/pull/988. No parse errors. 
No further actions required for this issue.

PR #988, opened 2023-04-23
[OM] Fix parse errors in string

kunjsong01:
Fixed all parse errors in https://github.com/esbmc/esbmc/issues/832 

N.B. The string OM file needs a retab when approved.

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/string/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/string/main.cpp
@@ -1,0 +1,4 @@
+#include <string>
+int main () {
+  return 0;
+}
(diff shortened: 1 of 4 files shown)


lucasccordeiro (approved):
LGTM.

@kunjsong01: Can we enable existing regression tests with this PR?

kunjsong01:
> Can we enable existing regression tests with this PR?

@lucasccordeiro Not yet. After fixing all the parse errors, we got an assertion failure - https://github.com/esbmc/esbmc/issues/989

lucasccordeiro:
@kunjsong01: if you can now concentrate on this assertion failure, that would be great.

rafaelsamenezes (commented):
Why do you need these?

kunjsong01:
> Why do you need these?

This PR fixed the parse errors in `string` OM for C++ std::string stuff.

PR #993, opened 2023-04-26
[C++ OM] Fix cmath OM

XLiZHI:
When call `math.h`, the path is the standard library, and Clang currently gives us an error for that.
```
In file included from main.cpp:2:
/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/math.h:38:7: error: use of undeclared identifier 'std'
using std::abs;
      ^
/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/math.h:39:7: error: use of undeclared identifier 'std'
using std::acos;
      ^
```
Dependency was as follows:
```
. /usr/include/math.h
.. /libraries/cmath
```

This PR fixed some issues with cmath file, added math.h file.

### DIFF
--- a/src/cpp/library/cmath
+++ b/src/cpp/library/cmath
@@ -29,7 +29,7 @@ float sqrt(float n) {
 
 float sqrt(int n) {
 	__ESBMC_assert(n >= 0, "The number must be greater or equal than 0");
-	return nondet_int();
+	return nondet_float();
 }
 
 double sqrt(double n) {
@@ -48,7 +48,7 @@ double sqrt(double n) {
 
 unsigned sqrt(unsigned x) {
 	__ESBMC_assert(x >= 0, "The number must be greater or equal than 0");
-	return nondet_unsigned();
+	return nondet_uint();
 }
 
 float exp(float n) {
--- a/src/cpp/library/math.h
+++ b/src/cpp/library/math.h
@@ -1,0 +1,1 @@
+#include <cmath>
--- a/regression/esbmc-cpp/OM_sanity_checks/cmath/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cmath/main.cpp
@@ -1,0 +1,4 @@
+#include "cmath"
+int main () {
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cmath/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/cmath/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+-I /__w/esbmc/esbmc/src/cpp/library --memlimit 14000000 --timeout 900
+
+^VERIFICATION SUCCESSFUL$


src/cpp/library/cmath
```diff
@@ -29,7 +29,7 @@ float sqrt(float n) {
 
 float sqrt(int n) {
 	__ESBMC_assert(n >= 0, "The number must be greater or equal than 0");
-	return nondet_int();
```
XLiZHI, line 32: `float sqrt(int n)` Should the method return a float?

src/cpp/library/cmath
```diff
@@ -48,7 +48,7 @@ double sqrt(double n) {
 
 unsigned sqrt(unsigned x) {
 	__ESBMC_assert(x >= 0, "The number must be greater or equal than 0");
-	return nondet_unsigned();
```
lucasccordeiro, line 51: Shouldn't we keep it as nondet_unsigned() since the return is unsigned?

lucasccordeiro (changes requested):

src/cpp/library/cmath
XLiZHI, line 51: There is no declaration of `nondet_unsigned()` in `clang_c_language.cpp` , whether to use `nondet_int()` instead. Its behavior is the same.

src/cpp/library/cmath
lucasccordeiro: it should return nondet_float();

src/cpp/library/cmath
lucasccordeiro: You can declare:

unsigned nondet_unsigned();

kunjsong01: `nondet_uint`

XLiZHI: May I ask what is the difference between `unsigned` and `unsigned int`? In my tests, they seemed to perform the same.

src/cpp/library/math.h
```diff
@@ -0,0 +1 @@
+#include <cmath>
```
kunjsong01, line 1: Why do we need "math.h" ?

kunjsong01 (changes requested):

src/cpp/library/cmath
kunjsong01: >May I ask what is the difference between unsigned and unsigned int?

No difference in this case. 
But `nondet_int()` shouldn't be used here.

src/cpp/library/math.h
XLiZHI: I investigated the `math.h` file from the C++ standard library, which has `cmath` as a header file. The header file used in our COM is `math.h`. Should we add a `math.h` file to avoid the error I showed in the description?

src/cpp/library/cmath
XLiZHI: I made a mistake. I meant to use `nondet_uint`, but forgot 'u'.

kunjsong01 (approved):

lucasccordeiro (approved):

Issue #978, opened 2023-04-19
[SV-COMP] Evaluate ub-shift-check over the HW benchamrks


lucasccordeiro:
@brcfarias: Can I ask you to evaluate your implementations of the ub-shift-check over the HW benchmarks in SV-COMP 2023?

https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/tree/main/c/hardware-verification-array
https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/tree/main/c/hardware-verification-bv

You can use our GitHub actions: https://github.com/esbmc/esbmc/actions/workflows/benchexec.yml

You must update our wrapper scripts to enable the ub-shift-check: https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/esbmc-wrapper.py and update our bench.

@rafaelsamenezes or @Anthonysdu can support you here.

brcfarias:
Here are the results for the benchmark without --ub-shift-check

Benchmark execution: https://github.com/esbmc/esbmc/actions/runs/4758395502

**hardware-verification-array**
  timeout: 5
  correct-false: 1
  total: 6
  
**hardware-verification-bv**
  timeout: 1052
  correct-false: 87
  correct-true: 77
  total: 1216

brcfarias:
Here are the results with --ub-shift-check enabled

**hardware-verification-array**
  timeout: 5
  correct-false: 1
  total: 6
 
**hardware-verification-bv**
  timeout: 564
  correct-false: 127
  correct-true: 78
  wrong-false: 5
  out of memory: 439
  unknown: 3
  total: 1216

lucasccordeiro:
What is the conclusion?Sent from my iPhoneOn 26 Apr 2023, at 12:49, Bruno Farias ***@***.***> wrote:﻿
Here are the results with --ub-shift-check enabled
hardware-verification-array
timeout: 5
correct-false: 1
total: 6
hardware-verification-bv
timeout: 564
correct-false: 127
correct-true: 78
wrong-false: 5
out of memory: 439
unknown: 3
total: 1216

lucasccordeiro:
What about the 5 wrong results?

brcfarias:
The violations occurred in lines containing % operator, as detailed below

**1 - btor2c-lazyMod.collision.1.prop1-back-serstep.c** line 586 function main
```cpp
[...]

PR #996, opened 2023-04-27
[goto_check] Avoid shift check on divisions and modulo operator

brcfarias:
This PR addresses the issues found during the execution of svcomp hardware benchmark as detailed at https://github.com/esbmc/esbmc/issues/978#issuecomment-1526557969

### DIFF
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -394,6 +394,10 @@ void goto_checkt::shift_check(
   if(!enable_ub_shift_check)
     return;
 
+  assert(is_lshr2t(expr) || is_ashr2t(expr) || is_shl2t(expr));
+
+  overflow_check(expr, guard, loc);
+
   auto right_op = (*expr->get_sub_expr(1));
 
   expr2tc zero = gen_zero(right_op->type);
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

src/goto-programs/goto_check.cpp
```diff
@@ -394,6 +394,9 @@ void goto_checkt::shift_check(
   if(!enable_ub_shift_check)
     return;
 
+  if(!is_lshr2t(expr) && !is_ashr2t(expr) && !is_shl2t(expr))
```
rafaelsamenezes: Why are we arriving at this function with anything besides a shift? Shouldn't we:

- Add an assertion that this function is only reachable within shift operations
- Fix the flow that is calling a shift check on something that it shouldn't

rafaelsamenezes (changes requested):

lucasccordeiro:
@brcfarias: could you please address @rafaelsamenezes's comments here?

src/goto-programs/goto_check.cpp
brcfarias: >Why are we arriving at this function with anything besides a shift

We didn't have a break in the switch-case of goto_checkt::check_rec(...) after case expr2t::modulus_id, so shift_check was being called on this operation. Initially, I preferred a minimal change to reduce the chances of side effects, but I agree with you that it is better to adjust the flow, and it seems safe to modify it. I will update the commit. Thank you for the suggestion.

rafaelsamenezes (approved):
LGTM

lucasccordeiro:
Thanks, both!

As soon as CI is green, I'll merge this PR.

PR #1000, opened 2023-05-03
Fix union initializer for cast-to-union cases supported by Clang and GCC

### LINKED ISSUE #994: ToUnion does not support bitfields
The following code:

```c
#include <stdbool.h>

#include <stdint.h>



typedef struct {

                unsigned long int tmp1: 64;

} struct_1;



typedef union {

    unsigned int raw : 2;

    struct {

        unsigned int x1 : 1;

        unsigned int x2 : 1;

    };

} union_1;



void func(struct_1 *tmp_p);



void func(struct_1 *tmp_p) {

    union_1 tmp1 = (union_1)(uint32_t)(tmp_p->tmp1);

}
```

Fails with `ERROR: Couldn't map type  into the union`

This is mainly due to how ESBMC handles struct equivalence at https://github.com/esbmc/esbmc/blob/master/src/clang-c-frontend/typecast.cpp#L43

fbrausse:
Fixes #994. Before, the correct component of the union was selected based on equality of types. However, for instance GCC treats `unsigned x : 32` and `uint32_t` as equal while `unsigned y : 31` is not equal to `uint32_t` and thus disallowed for initializing a union. Clang however allows the second case and even multiple `unsigned z1 : 1, z2 : 2, z3 : 3` union components. I am not 100% sure which one it takes. With this PR the component is chosen based on type equality or equality of the type underlying the bitfield. In the latter case the matching component with the largest bit-width is taken.


mikhailramalho (approved):
LGTM. I agree that we should follow whatever clang is doing.

Nit: Do we really need the new `const namespacet &ns`? It seems to be empty anyway.

fbrausse:
> LGTM. I agree that we should follow whatever clang is doing.
> 
> Nit: Do we really need the new `const namespacet &ns`? It seems to be empty anyway.

Thanks! I've put the new namespacet parameter there in order to be able to call `gen_typecast(ns, e, type)` (which the previous version did not do). This new gen_typecast() call is relevant for https://github.com/esbmc/esbmc/pull/1000/files#diff-a2269609ce5998b6c969953927ed8b65fa74576766667ef57bab24322634954d where the union's component has less (24) bits than that of the initializing expression. Without the typecast the smt_conv layer breaks somewhere. That's why it put it here.

lucasccordeiro:
@fbrausse : Can I ask you to rebase this PR with the master branch?

lucasccordeiro (approved):
@fbrausse: can you please run this PR over the SV-COMP benchmarks?

fbrausse:
> @fbrausse : Can I ask you to rebase this PR with the master branch?

Done.



> @fbrausse: can you please run this PR over the SV-COMP benchmarks?

It's been started as <https://github.com/esbmc/esbmc/actions/runs/4873492694>. Do we have a `master` baseline?

lucasccordeiro:
Thanks, @fbrausse.

The baseline is here: https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/stats-30s.txt

lucasccordeiro:
Master for 30 seconds.

````
   correct:            14485
     correct true:     8267
     correct false:    6218
   incorrect:            88
     incorrect true:     39
    incorrect false:     49
   unknown:            11886
   Score:              20720 (max: 43043)
````

This PR:

````
2023-05-04T02:18:07.7132544Z Statistics:          26459 Files
2023-05-04T02:18:07.7143002Z   correct:           14488
2023-05-04T02:18:07.7143605Z     correct true:     8269
2023-05-04T02:18:07.7144457Z     correct false:    6219
2023-05-04T02:18:07.7145335Z   incorrect:            88
2023-05-04T02:18:07.7145874Z     incorrect true:     39
2023-05-04T02:18:07.7146616Z     incorrect false:    49
2023-05-04T02:18:07.7157791Z   unknown:           11883
2023-05-04T02:18:07.7158972Z   Score:             20725 (max: 43043)
````

mikhailramalho:
> > LGTM. I agree that we should follow whatever clang is doing.
> > Nit: Do we really need the new `const namespacet &ns`? It seems to be empty anyway.
> 
> Thanks! I've put the new namespacet parameter there in order to be able to call `gen_typecast(ns, e, type)` (which the previous version did not do). This new gen_typecast() call is relevant for https://github.com/esbmc/esbmc/pull/1000/files#diff-a2269609ce5998b6c969953927ed8b65fa74576766667ef57bab24322634954d where the union's component has less (24) bits than that of the initializing expression. Without the typecast the smt_conv layer breaks somewhere. That's why it put it here.

I see. I thought the fuzz test code was part of the clang frontend, that's why I thought it was empty.

lucasccordeiro:
`Many thanks, @fbrausse

PR #1001, opened 2023-05-03
[goto-check] Remove __ESBMC prefix

### LINKED ISSUE #999: Remove __ESBMC prefix when generating CE for intrinsic functions
The followng program:

```c
#include <stdio.h>

 

int main( int argc, char *argv[])

{

 char reg_name[12];

 printf("Enter your username:");

 scanf("%12s", reg_name); //Buffer overflow here

 printf("The program is now registered to %s.\n",reg_name);

 return 0;

}
```

Generates the following CE:

```c
State 4 file scanf.c line 9 function main thread 0

Violated property:

 file scanf.c line 9 function main

 buffer overflow on c:@F@__ESBMC_scanf

 0
```

Maybe we could replace `c:@F@__ESBMC_scanf` with `scanf`

brcfarias:
This PR fixes #999.

### DIFF
--- a/src/clang-c-frontend/expr2c.cpp
+++ b/src/clang-c-frontend/expr2c.cpp
@@ -924,6 +924,9 @@ std::string expr2ct::convert_symbol(const exprt &src, unsigned &)
   if(src.id() == "next_symbol")
     dest = "NEXT(" + dest + ")";
 
+  if(dest.find("__ESBMC_") == 0)
+    dest.erase(0, 8); // Remove __ESBMC_
+
   return dest;
 }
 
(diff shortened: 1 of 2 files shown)


brcfarias:
The sample code is available [here](https://github.com/esbmc/esbmc/blob/381413c5fa287296eb938e389fc3283b49b79a00/regression/cstd/scanf-bug/main.c)

mikhailramalho (changes requested):
We actually have code to do that in expr2c or expr2cpp.

They've been broken for a while, so we should fix and use them here instead.

lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:
Many thanks, @brcfarias.

PR #1002, opened 2023-05-04
Bug fixed for Solidity return value

ChenfengWei0:
Resolves an error when calling a member function to get the return value, e.g.

```solidity
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.5.0;

contract SolidityTest {
    uint8 x;
    uint8 y;

    constructor() {
        x = 0;
        y = 1;
    }

    function A() public returns (uint8) {
        return 0; 
    }

    function B() public returns (uint8) {
        return y; 
    }
}

contract Derived is SolidityTest {
    SolidityTest xx = new SolidityTest();

    function getResult() public {
        uint8 result = xx.A() > xx.B() ? xx.A() + 10 : xx.B() + 10;
        assert(xx.A() == 0);
        assert(xx.B() == 1);
        assert(result == 11);
    }
}

```

TODO:
- We only support unit8 return value for now. We should support all types of value 
- We do not support returning multiple values from a Solidity function.
	- This implementation might be tricky, as we cannot find a direct example from the clang_c front end. A possible idea would be to pass multiple return values through the symbol table


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2679,15 +2718,14 @@ nlohmann::json solidity_convertert::make_callexpr_return_type(
         type_descrpt["typeString"].get<std::string>().find(
           "returns (contract") != std::string::npos)
       {
-        // TODO: Fix me
+        // TODO: Fix me. We treats contract as void
```
lucasccordeiro: "treats" -> "treat".

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2120,6 +2156,8 @@ bool solidity_convertert::get_elementary_type_name(
   }
   case SolidityGrammar::ElementaryTypeNameT::INT_LITERAL:
   {
+    new_type = int_type();
```
lucasccordeiro, line 2162: Can you add some comments here?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1169,9 +1196,15 @@ bool solidity_convertert::get_expr(
     new_expr.cmt_lvalue(true);
     new_expr.name(name);
 
+    nlohmann::json callee_rtn_type =
```
lucasccordeiro, line 1201: Can you add some comments here?

lucasccordeiro (approved):
Minor comments.

lucasccordeiro:
@ChenfengWei0: could you please rebase this PR with the master branch?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1618,11 +1651,12 @@ bool solidity_convertert::get_conditional_operator_expr(
 
 bool solidity_convertert::get_cast_expr(
   const nlohmann::json &cast_expr,
-  exprt &new_expr)
+  exprt &new_expr,
+  const nlohmann::json &int_literal_type = nullptr)
```
kunjsong01: default parameter value should be in the header, not .cpp.

kunjsong01: Also, it's better to initialize const ref to refer to a valid object. If this parameter needs to be nullptr by default, we shall use a const pointer instead of const ref.

kunjsong01 (changes requested):

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: Thanks for the suggestion. I will use a const pointer

ChenfengWei0:
Fix all the issues mentioned above

kunjsong01 (approved):

lucasccordeiro (approved):

Issue #991, opened 2023-04-25
[C++ verification] Various issues in nested class members and object composition


kunjsong01:
TC:
```cpp
#include <cassert>

char x = 'X';

class string
{
public:
  class iterator
  {
  public:
    char* pointer;
    void do_something()
    {
      iterator buffer;
      buffer.pointer = &x;
      pointer = buffer.pointer;
    }
  };

  iterator itr;
};

int main()
{
  string myString;
  myString.itr.do_something();
  assert(myString.itr.pointer == &x);
  //assert(myString.itr.pointer != &x); // FAIL, as should be identical
  return 0;
}
```

ESBMC version:  5e6db517094c
cmd line: ./esbmc main.cpp -I library/

kunjsong01:
This issue blocks `string` OM include: 
```
Converting
Generating GOTO Program
ERROR: Looking up index of nonexistant member "pointer" in struct/union "std::string::iterator"
Aborted (core dumped)
```

kunjsong01:
The root cause is incomplete and inconsistent struct type in symbols, possibly due to the recursive calls in our conversion flows --- class types cannot be synced in this way. So only after executing `get_struct_union_class` can we get a complete class' type. But this is not always the case - the conversion flow would stop and return in the middle leaving an incomplete type which is also inconsistent with the actual class' type symbol.  
```cpp
// For DeclRefExpr type to a class `iterator`
0: symbol
* type: struct 
  * tag: string::iterator
  * components:  <--- components being empty as a result of recursive calls to `get_struct_union_class` and `get_type`
* identifier: c:main.cpp@163@S@string@S@iterator@F@do_something#@buffer
```

Symbols referring to a struct type should be simplified to: 
```
0: symbol
    * type: symbol <--- this refers to the type symbol for class `iterator`
        * identifier: tag-string::iterator 
    * identifier: c:main.cpp@163@S@string@S@iterator@F@do_something#@buffer
```

kunjsong01:
There are two issues found here: 

**1. Nested class member access issue:** 
Failed to access nested class member due to incomplete and inconsistent type as explained in https://github.com/esbmc/esbmc/issues/991#issuecomment-1521833420 and https://github.com/esbmc/esbmc/issues/991#issuecomment-1523413991. 

**2. Obejct composition issue:** 
Goto-symex failed to instantiate due to missing argument:
```
Counterexample:

State 1  thread 0
----------------------------------------------------
Violated property:

[...]

PR #1007, opened 2023-05-07
Fix object composition

kunjsong01:
Fixed 991 issue 2 as in https://github.com/esbmc/esbmc/issues/991#issuecomment-1525504379

Turn
```cpp
this->itr = iterator();  // no argument for iterator constructor
```
into: 
```cpp
iterator(&this->itr);
```

Also removed unnecessary recursive type getters if a class/struct type already exists in symbol table.


lucasccordeiro (approved):

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3417,3 +3416,16 @@ bool clang_c_convertert::get_vft_binding_expr(
   abort();
   return true;
 }
+
+void clang_c_convertert::get_ref_to_struct_type(typet &type)
+{
+  /*
+   * For some special cases, we need to get a symbol type referring to
```
rafaelsamenezes, line 3423: Do you have any example?

src/clang-cpp-frontend/clang_cpp_convert.h
```diff
@@ -171,6 +183,18 @@ class clang_cpp_convertert : public clang_c_convertert
...
+   * A default copy constructor would have the return type below:
+   * * return_type: constructor
+   *   #default_copy_cons: 1
+   *
+   * Arguments:
+   *  cxxmdd: clang AST node representing the ctor/dtor we are dealing with
+   *  rtn_type: the corresponding return type node
+   */
+  void
+  annotate_cdtor_rtn_type(const clang::CXXMethodDecl *cxxmdd, typet &rtn_type);
```
rafaelsamenezes: This needs a better name.

rafaelsamenezes (changes requested):
Just some nitpicks

src/clang-c-frontend/clang_c_convert.cpp
kunjsong01: Fixed - added an example in the comment.

src/clang-cpp-frontend/clang_cpp_convert.h
kunjsong01: Fixed - disambiguate and replaced the name with "annotate_ctor_dtor_rtn_type"

rafaelsamenezes (approved):

PR #1033, opened 2023-05-22
Lookup-based atoi, atol and atoll

### LINKED ISSUE #1025: [operational-model] create a lookup table for atol
Comments from @fbrausse (see https://github.com/esbmc/esbmc/pull/1014):

Did you consider a lookup-based method, for `atol` e.g.

```c
/* one plus the numeric value, rest is zero to also catch '\0' */
static const unsigned char MAP[256] = {
	['0'] = 1, ['1'] = 2, ['2'] = 3, ['3'] = 4, ['4'] = 5,
	['5'] = 6, ['6'] = 7, ['7'] = 8, ['8'] = 9, ['9'] = 10,
};
```

As far as I know also the ctype functions like `isdigit`, etc. are usually implemented with lookup tables.

fbrausse:
These methods are faster to check (on my machine) than the strtol() version. I've included a specification of atoi in the regressions, basically checking the equality between the two versions. But that one still is very slow and thus not enabled by default.

I'm not sure whether the style with the #define is acceptable for these operational models. If not, I can write them out.


lucasccordeiro (approved):
Many thanks, @fbrausse.

fbrausse:
Master (not sure which Github run id, it's the https://github.com/esbmc/esbmc/blob/67c734badee221c8fbe9ac367f15cfb254ecdddc/scripts/competitions/svcomp/stats-30s.txt):
```
   correct:            14077
     correct true:     7848
     correct false:    6229
   incorrect:            88
     incorrect true:     42
    incorrect false:     46
   unknown:            12294
   Score:              19845 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5043446306):
```
  correct:           14505
    correct true:     8269
    correct false:    6236
  incorrect:            88
    incorrect true:     42
    incorrect false:    46
  unknown:           11866
  Score:             20694 (max: 43043)

```

lucasccordeiro:
@fbrausse : can I ask you to fix the CI?

fbrausse:
> @fbrausse : can I ask you to fix the CI?

@lucasccordeiro I'm not sure how: The Codacy thingy complains about the "unused" jump labels `__ESBMC_HIDE:`...

lucasccordeiro:
@rafaelsamenezes: could you please check this issue?

rafaelsamenezes:
> @rafaelsamenezes: could you please check this issue?

Just accept it and let the change through. It should stop complaining after its on master.

Not using a label can be a source of issues and we should check for it in the OMs. The only exception is, of course, the __ESBMC_hide.  I am not sure if there is a way (or even if its a good idea) to start setting  the labels that static checkers should ignore.

PR #1036, opened 2023-05-23
Fix access to multi-dim array member of unions

fbrausse:
Just flattening the expression's type seems to work, though I'm not 100% sure of this solution. I'll run it over SV-COMP to see whether things break.

In addition, while adding some regression tests, I noticed that big-endian union handling seems to be broken (at least in this multi-dim array member case).

*Edit:* this is about #841.


fbrausse:
@lucasccordeiro This approach still needs some work, I'll mark it ready when it's working.

fbrausse:
This PR fixes #841. It:
- flattens multi-dim array types of members of unions in order to support bit-casting those to a long BV
- adds a bit-cast to flatten_to_bitvector() for fpbv and bool types in order to enable concatenation
- fixes the computation of the mantissa size in cvc_convt::get_fpbv() to take into account the sign bit
- adds an irep2 assertion that concat2t is constructed from bv types (cf. #204)
- makes more of the .dump() methods use log_status() instead of log_debug() because this makes debugging easier;.dump() is not supposed to be called during regular control flow anyway

I've started an sv-comp run since it touches the union encoding: https://github.com/esbmc/esbmc/actions/runs/5066911312

fbrausse:
Master:
```
   correct:            14077
     correct true:     7848
     correct false:    6229
   incorrect:            88
     incorrect true:     42
    incorrect false:     46
   unknown:            12294
   Score:              19845 (max: 43043)
```

This PR:
```
  correct:           14084
    correct true:     7852
    correct false:    6232
  incorrect:            88
    incorrect true:     42
    incorrect false:    46
  unknown:           12287
  Score:             19856 (max: 43043)
```

lucasccordeiro:
Thanks for writing this PR, @fbrausse.

lucasccordeiro (approved):
LGTM.

PR #1042, opened 2023-05-25
Added option to download and build dependencies during build

rafaelsamenezes:
This should make build environments easier to set up and change.

- [X] LLVM
- [x] Z3
- [x] Boolector
- [x] Bitwuzla
- [ ] ~CVC4~ will skip directly to CVC5
- [ ]  ~CVC5~ will make it in other PR
- [ ]  ~MathSAT~ we are not testing it currently
- [ ] ~Yices 2~ will make it in other PR
- [ ] ~IBEX~ will make it in other PR

Also, maybe its time to refactor the CI as well. To much repeated code that could be turned into multiple scripts:

- [X] Create build scripts
- [X] Matrix builds
- [X] Clean-up the scripts folder
- [ ] ~Sanitizer builds~ will make in other PR
- [ ] ~macOS build~ will make in other PR

~This is currently blocked due to #1055  as the include paths for C++ and CUDA are broken,~ Fixed by #1124


fbrausse:
Does it build on Linux and Windows without `-Werror`?

rafaelsamenezes:
> Does it build on Linux and Windows without -Werror?

Not yet, on Windows the Z3 link is wrong and ibex build seems to have an issue.

scripts/build.sh
```diff
@@ -1,8 +1,74 @@
...
+#!/bin/sh
+
+# Create build directory
 mkdir build && cd build
-cmake .. -DClang_DIR=$PWD/../clang -DLLVM_DIR=$PWD/../clang -DBUILD_STATIC=On -DBoolector_DIR=$PWD/../boolector-release -DCMAKE_INSTALL_PREFIX:PATH=$PWD/../release -DCMAKE_BUILD_TYPE=RelWithDebInfo
-make -j2 . && cpack
+
+# Set arguments that should be available for every OS
+BASE_ARGS="-DDOWNLOAD_DEPENDENCIES=On -GNinja -DENABLE_CSMITH=On -DBUILD_TESTING=On -DENABLE_REGRESSION=On -DENABLE_SOLIDITY_FRONTEND=On -DENABLE_JIMPLE_FRONTEND=On -DCMAKE_INSTALL_PREFIX:PATH=$PWD/../release"
+SOLVER_FLAGS="-DENABLE_BOOLECTOR=On -DENABLE_YICES=Off -DENABLE_CVC4=OFF  -DENABLE_BITWUZLA=OFF -DENABLE_GOTO_CONTRACTOR=OFF"
```
fbrausse, line 8: You mentioned that Bitwuzla was handled in this PR, but it is disabled here. Could you explain why? Yices and CVC4 were also enabled earlier, but you moved those to another PR, that's fine with me.

fbrausse (approved):
LGTM, just one question below. I've tested this with my local builds and they all work fine (with `DOWNLOAD_DEPENDENCIES` turned off).

lucasccordeiro (approved):

scripts/build.sh
rafaelsamenezes: It should be work right now by just enabling it, however I think it will be better enabling only the core esbmc. Then each PR will be to add the other solvers and platforms (macOS is currently disabled).

Issue #972, opened 2023-04-18
[C++ verification] Migrate to C++14 standard


kunjsong01:
Since our verification target uses upto C++14 standard, ESBMC shall also be upgraded to this level. The clang AST &ast;might&ast; contain more information to facilitate the conversion flows. 

### Evaluation: 
1. Benchmarks for core language features - `inheritance_bringup`, `polymorphism_bringup`
3. Benchmarks for OMs - `stream` or `string`

We'll probably need to migrate to C++14 as soon as we have a good passing rate in either `stream` or `string` benchmark. Further code extensions on IR conversion flows based on C++98/03 ASTs might give us surprises when migrating to C++14. 

### Preliminary Experiment:
For sure we have to deal with some new AST nodes, e.g. the implicit move constructor and rvalue reference...
```cpp
class Bird {
  public:
  virtual int do_something(void) { return 21; }
};
```
```cpp
254 |   `-CXXConstructorDecl 0x556bf45732c0 <col:7> col:7 implicit constexpr Penguin 'void (std::Penguin &&)' inline default noexcept-unevaluated 0x556bf45732c0
255 |     `-ParmVarDecl 0x556bf45733f0 <col:7> col:7 'std::Penguin &&'
```
```python
Converting
ERROR: Conversion of unsupported clang type: "RValueReference
RValueReferenceType 0x55b393235110 'std::Bird &&'
`-RecordType 0x55b393234c60 'std::Bird'
  `-CXXRecord 0x55b393234bc8 'Bird'
```

fbrausse:
I would recommend adding an option that specifies the given language including its version. It could be similar to `-x` supported by Clang and GCC and a second option similar to `-std=`.

kunjsong01:
This issue currently has the highest priority as there's verification request from our research partners. 

TODOs:
- Migrate to C++11 (or preferably C++14) standard 
- Add cstdint OM

kunjsong01:
For the time being, we can probably "bypass" those nodes related to move semantics in the AST, because there's no verification request using that feature. The rest of the conversion flow shall not be affected. 

Original estimate: 1w for this ticket.

kunjsong01:
closed via https://github.com/esbmc/esbmc/pull/1047

PR #1047, opened 2023-05-30
Added option for C++ standard

kunjsong01:
~~Let's work with AST generated by modern C++14 standard.~~ 
~~Force standard to be C++14 standard based on comments in https://github.com/esbmc/esbmc/issues/972~~

Edit: added an option for C++ standard. Default is C++03.


src/clang-cpp-frontend/clang_cpp_language.cpp
```diff
@@ -21,8 +21,8 @@ languaget *new_clang_cpp_language()
 
 void clang_cpp_languaget::force_file_type()
 {
-  // We also force the standard to be c++98
-  compiler_args.push_back("-std=c++98");
+  // We also force the standard to be c++14
+  compiler_args.push_back("-std=c++14");
```
rafaelsamenezes: This could be the default, but you should add a flag to set C++ standard as flag. Then you could duplicate the testcases with a C++98 passing and a C++14 KNOWNBUG

rafaelsamenezes (changes requested):

src/clang-cpp-frontend/clang_cpp_language.cpp
lucasccordeiro: I agree with @rafaelsamenezes' suggestion.

lucasccordeiro (approved):
Please consider @rafaelsamenezes' suggestion.

src/clang-cpp-frontend/clang_cpp_language.cpp
kunjsong01: Done.

src/esbmc/options.cpp
```diff
@@ -72,6 +72,9 @@ const struct group_opt_templ all_cmd_options[] = {
     {"warning,W",
      boost::program_options::value<std::vector<std::string>>(),
      ""},
+    {"cppstd",
```
rafaelsamenezes, line 75: In general, we consider flags with only one "-" to be passed through the frontend. Also, users integrate ESBMC by using a build system so they would prefer to pass `-std=c++11` than `--cppstd=11`. I would suggest to also look at the flags that we pass to the frontends.

src/esbmc/options.cpp
```diff
@@ -72,6 +72,9 @@ const struct group_opt_templ all_cmd_options[] = {
...
+     boost::program_options::value<std::string>()->value_name("version"),
+     "set C++ standard"},
```
rafaelsamenezes: I would change the message to `set C++ standard (available: 98, 03, 11, 14, or 17)`

lucasccordeiro (approved):
LGTM.

Many thanks, @kunjsong01.

rafaelsamenezes (changes requested):

src/esbmc/options.cpp
kunjsong01: good point.
Added

src/esbmc/options.cpp
kunjsong01: logged in https://github.com/esbmc/esbmc/issues/1050, which shall be done in another PR.

rafaelsamenezes (approved):

fbrausse (commented):
I agree with @rafaelsamenezes' comment regarding `-std=c++11`. Also, to check whether this switch actually makes a difference, I would suggest adding a test case like `regression/esbmc-cpp/bug_fixes/972_std14/main.cpp` but with `--cppstd 98` expecting a parsing error.

kunjsong01:
> I would suggest adding a test case like regression/esbmc-cpp/bug_fixes/972_std14/main.cpp but with --cppstd 98 expecting a parsing error.

Good point. I've added another test case using c++98

kunjsong01:
Regarding the option styling, `--cppstd 98` is pretty much in line with other frontend options, and other options taking a numerical value e.g. "--k-step nr" in ESBMC. There are 39 frontend options in total. 35 out of 39 options have the two hyphens "--" style.

[list_of_frontend_options.txt](https://github.com/esbmc/esbmc/files/11614793/list_of_frontend_options.txt)

The `-std=c++11` style is in line with the compilers. But we are not writing a compiler. As for the users, I don't see `--cppstd 11` is less intuitive than `-std=c++11` in any way.  

Do we have any existing option in the format `-option=<number>` ? If not, for now an issue might be just enough for aesthetic improvement?

fbrausse:
> The `-std=c++11` style is in line with the compilers. But we are not writing a compiler.

I tend to disagree on this stance: ESBMC has options that are known from (non-Windows) compilers, such as `-I`, `-D`, `--sysroot` even though some compilers only allow a single dash for the long options. I regard `-std=c++N` or `--std c++N` similarly universal. Whether it is one dash with an `=` symbol or two dashes and space separated in my opinion is no big deal.

> As for the users, I don't see `--cppstd 11` is less intuitive than `-std=c++11` in any way.

The difference between these two is the semantics the option has: do we want a similar option for each frontend language that supports multiple versions (C, in particular), or should the language and standard be part of the same single `-std` option? Some time ago @mikhailramalho said that atm. ESBMC does "not really" support multiple different input languages and if that fact remains, I would really opt for a `-std` style option. Even if not, I'd rather vote for context- (aka position-) sensitive option processing akin `-std=c99 file1.c -std=c++98 file2.cc` instead of specifying a standard language-wide.

fbrausse (approved):
LGTM.

kunjsong01:
>The difference between these two is the semantics the option has: do we want a similar option for each frontend language that supports multiple versions (C, in particular), or should the language and standard be part of the same single -std option?

I see... makes sense, indeed. 

If we opt for `-std` style, it would require a complete re-writing of this patch.

BTW, it doesn't seem straightforward with [boost program optoin without writing a bespoke parser](https://stackoverflow.com/questions/19422541/boost-program-option-force), unless we write a bespoke parser. 

Do we have any existing option in the format `-option=<number>`?  If so, it should be just a implementation with a snap. 
Otherwise, https://github.com/esbmc/esbmc/issues/1050 should do, and it's probably better to push this patch, as we have an external user relying on it.

fbrausse:
> If we opt for `-std` style, it would require a complete re-writing of this patch.

Indeed. Fortunately it's not invasive.

> Do we have any existing option in the format `-option=<number>`?

I'm not aware of any. Just go forward with this PR until #1050 sorts out how to generalize this to other languages.

PR #602, opened 2022-01-19
strict internal libc compat

fbrausse:
This PR enforces usage of headers of ESBMC's internal libc for `c2goto` as well as `esbmc` when parsing source files (with the clang-c-frontend). If this approach turns out to be successful (no clashes with system headers on any supported OS),
- we can remove unchanged copies of headers such as bits/pthreadtypes.h, see #600, and
- we will get compiler errors for incompatibilities between the OS headers and ESBMC's; thereby
- discrepancies between any pre-built operational models and the translated user sources should be minimized, that is, for any library we (want to) model (e.g., POSIX threads), the sources will be translated with types corresponding to ESBMC's libc.

~It is work-in-progress at the moment.~

PR #1049, opened 2023-05-31
[smt] smt_dump is no longer an error for Text

rafaelsamenezes:
Running --smtlib --output gives an error because smt_dump is not supported by the interface. ESBMC, however, requires all solvers to have it.

### DIFF
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -2735,8 +2735,7 @@ smt_astt smt_ast::project(
 
 void smt_convt::dump_smt()
 {
-  log_error("SMT dump not implemented for {}", solver_text());
-  abort();
+  log_status("SMT dump was sent to --output file");  
 }
 
 void smt_convt::print_model()
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -2735,7 +2735,8 @@ smt_astt smt_ast::project(
 
 void smt_convt::dump_smt()
 {
-  log_status("SMT dump was sent to --output file");  
+  log_error("SMT dump not implemented for {}", solver_text());
+  abort();
 }
 
 void smt_convt::print_model()
--- a/src/solvers/smtlib/smtlib_conv.cpp
+++ b/src/solvers/smtlib/smtlib_conv.cpp
@@ -105,12 +105,9 @@ smt_convt *create_new_smtlib_solver(
   fp_convt **fp_api)
 {
   if(!options.get_bool_option("smt-formula-only"))
-  {
-    log_error(
-      "[smtlib] the smtlib interface does not support solving for now. Please, "
-      "use it with --smt-formula-only");
-    abort();
-  }
+    log_warning(
+      "[smtlib] the smtlib interface solving is unstable. Please, "
+      "use it with --smt-formula-only for production");
   smtlib_convt *conv = new smtlib_convt(ns, options);
   *array_api = static_cast<array_iface *>(conv);
   *fp_api = static_cast<fp_convt *>(conv);
--- a/src/solvers/smtlib/smtlib_conv.cpp
+++ b/src/solvers/smtlib/smtlib_conv.cpp
@@ -909,13 +909,22 @@ smt_astt smtlib_convt::mk_sign_ext(smt_astt a, unsigned int topwidth)
 
 smt_astt smtlib_convt::mk_zero_ext(smt_astt a, unsigned int topwidth)
 {
+  log_debug("[smt_ast] mk_zero_ext with {} width", topwidth);
   smt_astt z = mk_smt_bv(0, mk_bv_sort(topwidth));
   return mk_concat(z, a);
 }
 
 smt_astt smtlib_convt::mk_concat(smt_astt a, smt_astt b)
 {
-  smtlib_smt_ast *ast = new smtlib_smt_ast(this, a->sort, SMT_FUNC_CONCAT);
+  /**
+   * (concat (_ BitVec i) (_ BitVec j) (_ BitVec m))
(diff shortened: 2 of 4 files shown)


lucasccordeiro (approved):

src/solvers/smt/smt_conv.cpp
```diff
@@ -2728,8 +2728,7 @@ smt_astt smt_ast::project(
 
 void smt_convt::dump_smt()
 {
-  log_error("SMT dump not implemented for {}", solver_text());
-  abort();
+  log_status("SMT dump was sent to --output file");
```
fbrausse: If this function is not overridden and executed, printing the message would factually be wrong since `dump_smt()` did not produce any output to the file. Assuming that every backend is supposed to implement `dump_smt()`, wouldn't the proper solution be to implement it for the smtlib backend? Am I missing something?

rafaelsamenezes: So just dump the formula at stdout?

fbrausse: That's what I would expect. *Edit:* unless `--output` was given. But again, the output management is a bit unclear right now since ESBMC can generate lots of outputs.

rafaelsamenezes: Yeah, the `--output` is a very misleading flag. Also, its extra confusing for smtlib output.

For other solvers, using `--smt-formula-<only/too>` means to invoke the `dump_smt()` method. For the smtlib, generating a file/output is a requirement. So its redundant to print it out again.

fbrausse: > For the smtlib, generating a file/output is a requirement. So its redundant to print it out again.

It only is required if no `--smtlib-solver-prog` is given. If the solver-prog is used, based just on the help message description I would expect `--output` to still write whatever formula the smtlib backend produces to the given file, however, it refuses: `ERROR: Can't solve SMTLIB output and write to a file, sorry`.

So on a non-Windows build right now, `smtlib` requires either the `--output` or the `--smtlib-solver-prog` option to be present. Both options result in the above error, none of them are `ERROR: Must specify an smtlib solver program in smtlib mode`. Windows builds don't support `--smtlib-solver-prog` because process spawning is not implemented. Are you referring to and is this PR fixing the Windows build?

As an aside, `--smtlib-solver-prog` does not support handling spaces or command-line parameters to the executable, which could be fixed by exec'ing `sh -c PROG`.

rafaelsamenezes: At this point, I would say that the correct solution would be:

1. Ignore --output at convertion and only use it while dumping the formula.
2. This PR do not aim to fix the windows build, it's just enabling the usage of --smtlib to output formulas.

> As an aside, --smtlib-solver-prog does not support handling spaces or command-line parameters to the executable, which could be fixed by exec'ing sh -c PR

Could you clarify or point a link? I am having trouble finding the man page describing what PR does.

fbrausse: > > As an aside, --smtlib-solver-prog does not support handling spaces or command-line parameters to the executable, which could be fixed by exec'ing sh -c PR
> 
> Could you clarify or point a link? I am having trouble finding the man page describing what PR does.

PROG in `sh -c PROG` was meant to stand for the parameter given to `--smtlib-solver-prog`. For instance, if I want to use Z3 as the external solver, I would like/need to give `z3` the parameter `-in` in order for it to read from `stdin`. But specifying `--smtlib-solver-prog "z3 -in"` does not work as the program exec'ed by the smtlib backend now is `z3 -in` as one string. Such an executable does not exist. For comparison, this is similar to what was done in https://github.com/esbmc/esbmc/pull/602/files#diff-6d77e207194f76d3f2d17e0ef4b551cbf8bcc60848012dfb4cdaba0a8efae610 with the `ESBMC_OPTS` envvar: an arbitrary string that gets interpreted as a sequence of separate options by some simple rules. The rules in the `sh -c PROG` example would be the shell-parsing rules. For `ESBMC_OPTS` I've added a small method that does the argument splitting on a reduced subset of those shell-parsing rules (e.g., basic `\`-escaping, unescaped white-space separates options, quoting via `"`). That method could either be re-used for consistency or we could just delegate to the shell.

fbrausse: > At this point, I would say that the correct solution would be:
>  1. Ignore --output at convertion and only use it while dumping the formula.

Agreed.

src/solvers/smtlib/smtlib_conv.cpp
```diff
@@ -104,12 +104,38 @@ smt_convt *create_new_smtlib_solver(
   array_iface **array_api,
   fp_convt **fp_api)
 {
+  if(!options.get_bool_option("smt-formula-only"))
+  {
+    log_error(
+      "[smtlib] the smtlib interface does not support solving for now. Please, "
+      "use it with --smt-formula-only");
+    abort();
```
fbrausse: Would it be possible to make this a warning instead? The `--smtlib-solver-prog` option still works even though parsing the replies from the external program is somewhat buggy. Making this a warning would help in debugging it.

fbrausse (approved):
Generally looks good to me, just one question.

src/solvers/smt/smt_conv.cpp
fbrausse: > As an aside, --smtlib-solver-prog does not support handling spaces or command-line parameters to the executable, which could be fixed by exec'ing sh -c PR

I'll submit a PR that does this after this one here is merged.

src/solvers/smtlib/smtlib_conv.cpp
rafaelsamenezes: Sure

rafaelsamenezes:
There is still a couple of issues and inconssistencies that need to be sort out (mainly shown during the inductive step). They will come in other PRs though.

PR #1055, opened 2023-06-04
Update regression suite

kunjsong01:
Now all enabled C++ TCs are checked by our CI. 

Fixed all test.desc files in esbmc-cpp/* and added all test suites. 

Overall: 70 ish TCs enabled

For template TCs: 
- cbmc: 8 (mostly template TCs)
- gcc-template-tests: 11
- template: 5

39 more TCs enabled in cpp test suite.


lucasccordeiro (approved):
LGTM.

Many thanks, @kunjsong01.

rafaelsamenezes (commented):
The `-I /__w/esbmc/esbmc/src/cpp/library/` should be removed from all test cases (and from the testing script). It was kept because I though it was better to change it at the script than to patch thousands of files. If we are going to patch, it should to remove this path.

kunjsong01:
>The -I /__w/esbmc/esbmc/src/cpp/library/ should be removed from all test cases (and from the testing script).

@rafaelsamenezes True. Hard-coding it doesn't appear to be the most elegant way. 

> It was kept because I though it was better to change it at the script than to patch thousands of files. 

Which script are you referring to? The github workflow yaml or `testing_tool.py` ?

rafaelsamenezes:
> Which script are you referring to? The github workflow yaml or testing_tool.py ?

The `testing_tool.py` has a flag to add include paths. The cmake should use it for common builds and the yaml when the tool is invoked manually.

kunjsong01:
> > Which script are you referring to? The github workflow yaml or testing_tool.py ?
> 
> The `testing_tool.py` has a flag to add include paths. The cmake should use it for common builds and the yaml when the tool is invoked manually.

looks like we need to tweak the function `add_esbmc_regression_test` in `regression/CMakeLists.txt` to add the "--I" argument for `testing_tool.py` for all `esbmc-cpp/*` test suites. 

I'll see whether cmake's regex is capable of achieving this, and if so I'll submit a patch for it in the near future.

fbrausse:
> looks like we need to tweak the function `add_esbmc_regression_test` in `regression/CMakeLists.txt` to add the "--I" argument for `testing_tool.py` for all `esbmc-cpp/*` test suites.

Depending on any external include path is not the ideal way to go either. In my opinion, the C++ headers and compiled-to-goto operational models should follow the way we do it for C: compile implementation via `c2goto`, extract headers during runtime to a temporary directory and provide the correct flag (it is `-isystem`, not `-I`) to the clang++ frontend.

PR #1057, opened 2023-06-06
[cpp-frontend] Fix segfault on virtual methods 

### LINKED ISSUE #924: [C++ Verification] Segfault verifying program containing `new`
I'm getting a segmentation fault when running ESBMC on
```c++
#include <memory>

int main()
{
	int *p = new int;
}
```

brcfarias:
Fixes #924

### DIFF
--- a/regression/esbmc-cpp/cpp/virtual_methods/main.cpp
+++ b/regression/esbmc-cpp/cpp/virtual_methods/main.cpp
@@ -1,0 +1,8 @@
+struct Foo {
+  virtual void bar();
+};
+
+int main(void) {
+    return 0;
+}
+
--- a/regression/esbmc-cpp/cpp/virtual_methods/test.desc
+++ b/regression/esbmc-cpp/cpp/virtual_methods/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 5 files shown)


lucasccordeiro:
The following tests FAILED:
	2136 - regression/esbmc-cpp/polymorphism_bringup/polymorphism_common_error_3 (Failed)
Errors while running CTest
Error: Process completed with exit code 8.

lucasccordeiro (changes requested):
@brcfarias : could you please check the regression test that is failing?

kunjsong01:
`2137 - regression/esbmc-cpp/polymorphism_bringup/polymorphism_common_error_3 (Failed)`

This test case tries to instantiate a base class that contains a virtual method without definition. The linker reports an error in this case.

I reckon we can probably disable the failed test case in CI, as it doesn't do much in a sense that who would verify a program that doesn't link?

kunjsong01:
Bruno and I discussed offline. That was an old TC from the regression suite. 
just disable that TC then we can push this patch. @lucasccordeiro what do you think?

lucasccordeiro:
Fine by me ;-)

lucasccordeiro (approved):

kunjsong01 (approved):

lucasccordeiro:
Many thanks, @brcfarias.

PR #1058, opened 2023-06-06
Fix overflow check for unsigned integer

### LINKED ISSUE #1056: overflow-check is not detecting unsigned integer overflow/underflow
When ESBMC is run on the following code snippet with --overflow-check switch, verification passes. However, I was expecting it to flag an underflow at the operation "delta = j - i". When I ran CBMC with --unsigned-overflow-check on the same snippet, CBMC indeed flagged that line. Am I missing anything while running ESMBC with --overflow-check?

#include <stdio.h>
#include <assert.h>

int main(int argc, char **argv)
{
  unsigned int delta = 0;
  unsigned int i=5;
  unsigned int j=3;

  delta = j - i;
  printf("\nDelta %u", delta);

  return 0;
}

brcfarias:
Fixes #1056


rafaelsamenezes (changes requested):
Unsigned overlaps are not errors (as they are not undefined behavior). Also, this patch will break sv-comp as overflow is defined as:

```
It can never happen that the resulting type of an operation is a signed integer type but the resulting value is not in the range of values that are representable by that type. 
``` 

However, I think that most people would consider it an error... so add a `--no-unsigned-overflow-check` option?

brcfarias:
@rafaelsamenezes
> Unsigned overlaps are not errors (as they are not undefined behavior).

You are correct. The C standard confirms this by saying: "A computation involving unsigned operands can never overflow,
because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type."

> However, I think that most people would consider it an error... so add a --no-unsigned-overflow-check option?

I don´t know if it would be useful. Maybe we should just close it without modifications.

rafaelsamenezes:
> I don´t know if it would be useful.

It is useful (LLVM UB sanitizer and CBMC implements this analysis), maybe just not as a default check for overflow. Either an `--unsigned-overflow-check` or `--no-unsigned-overflow-check` should be fine.

brcfarias:
> > I don´t know if it would be useful.
> 
> It is useful (LLVM UB sanitizer and CBMC implements this analysis), maybe just not as a default check for overflow. Either an `--unsigned-overflow-check` or `--no-unsigned-overflow-check` should be fine.

Yes, CBMC has an `--unsigned-overflow-check` as mentioned in #1056 . So ESBMC could have this same. Is it ok for you @lucasccordeiro ?

lucasccordeiro:
That's fine, @brcfarias. Please go ahead ;-)

lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #1045, opened 2023-05-28
[clang-cpp-frontend] Fix template member functions support

brcfarias:
Fix crash regarding definition of template functions in structs.
It is related to #924.

PR #1060, opened 2023-06-07
Add regression test for #924

brcfarias:


lucasccordeiro (approved):

fbrausse (approved):
LGTM, thanks.

**Edit:** sorry, didn't see the failing test. Any ideas why it fails?

brcfarias:
> **Edit:** sorry, didn't see the failing test. Any ideas why it fails?

Initially I pushed test.desc without -I parameter. Changed it in last commit, but same result. It is passing locally.
@kunjsong01,  any ideas?

fbrausse:
Well, there are a bunch of errors related to the definition of iterators in the file included by `<memory>`. I suspect this header has not yet been made fit for ESBMC++. From my point of view the regression test itself is not dependent on `<memory>` so I would be fine with removing the #include.

kunjsong01:
> this header has not yet been made fit for ESBMC++. 

True. `memory` operational model is still in my TODO list :-) 

>any ideas?

It's complaining about the parsing errors in our memory operational model (OM) in https://github.com/esbmc/esbmc/blob/master/src/cpp/library/memory. When we give "-I /__w/esbmc/esbmc/src/cpp/library/", ESBMC replaces the standard library with OMs in `src/cpp/library/`. 

Also, just a nit pick: could you move this test case to `regression/esbmc-cpp/bug_fixes` please?

brcfarias:
> Well, there are a bunch of errors related to the definition of iterators in the file included by `<memory>`. I suspect this header has not yet been made fit for ESBMC++. From my point of view the regression test itself is not dependent on `<memory>` so I would be fine with removing the #include.

The fixes  (#1045 and #1057) resolved the issues found without ESBMC's OM. I mean, prior to those commits, we were unable to verify with "esbmc new.cpp" (without -I flag)

My last force-push commit modified the test.desc including -I flag to check if it was affecting the CI result.

kunjsong01:
>My last force-push commit modified the test.desc including -I flag to check if it was affecting the CI result.

@brcfarias does it work if removing the -f flag? If so we can just remove it in test.dec. It also means we are now able to use a real library instead of an OM with the clang-based converter. (which is good!)

brcfarias:
> It also means we are now able to use a real library instead of an OM with the clang-based converter. (which is good!)

Exactly.

> does it work if removing the -f flag? If so we can just remove it in test.dec.

The CI failed with -I as I tested before.

fbrausse:
In contrast to the current OM-related errors now, the one that happened without `-I` [was this](https://github.com/esbmc/esbmc/actions/runs/5242510232/jobs/9465994054#step:10:4198):
```
esbmc: ../src/clang-cpp-frontend/clang_cpp_convert.cpp:330: virtual bool clang_cpp_convertert::get_struct_union_class_methods_decls(const clang::RecordDecl&, struct_typet&): Assertion `ftd->isThisDeclarationADefinition()' failed.
```
To me it points to some missing handling of ASTs in the clang-cpp-frontend.

Issue #1066, opened 2023-06-10
ESBMC crashes if large value is specified in scanf 


tihanyin:
In this small example, scanf() is used to take a large integer as input. Evidently, the code lacks safety and can lead to unpredictable outcomes, commonly referred to as "undefined behavior".
ESBMC 7.2 crashes in this example due to incorrect conversion.

```c
#include <stdio.h>
int main() {
    int n;
    printf("Enter an integer: ");
    scanf("%10000000000d", &n);
    printf("%d\n",n);
    return 0;
}
```

Crash:

`esbmc test.c --overflow
ESBMC version 7.2.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing
Converting
Generating GOTO Program
GOTO program creation time: 0.381s
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
Aborted (core dumped)
`

fbrausse:
> Evidently, the code lacks safety and can lead to unpredictable outcomes, commonly referred to as "undefined behavior".

Your program does not have UB. The ESBMC crash obviously should not happen, though.

lucasccordeiro:
@ChenfengWei0: could you please check this issue?

ChenfengWei0:
> @ChenfengWei0: could you please check this issue?

Yes, I will work on it.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/1076

lucasccordeiro:
Many thanks for fixing this issue, @ChenfengWei0.

PR #1076, opened 2023-06-12
Fix scanf bigint bug

ChenfengWei0:
Relative issue: #1066 

- the bug was caused by `stoi(very_large_int)`
- Instead, we should use `BigInt` and `string2integer()`

### DIFF
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -72,8 +72,10 @@ protected:
   /** check for the buffer overflow in scanf/fscanf */
   void input_overflow_check(const expr2tc &expr, const locationt &loc);
   /* check for signed/unsigned_bv */
-  void
-  input_overflow_check_int(std::string width, BigInt limit, bool &buf_overflow);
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

src/goto-programs/goto_check.cpp
```diff
@@ -252,44 +256,19 @@ void goto_checkt::overflow_check(
 }
 
 void goto_checkt::input_overflow_check_int(
-  int width,
-  int limit,
+  const std::string width,
```
rafaelsamenezes: Why not use the bigint here?

ChenfengWei0: Because we do not compare the `width` with the `limit` in this function. Instead, we just need to confirm which of 8/16/32/64 is the `width` of the variable

ChenfengWei0: But I do agree we should uniform the format. I have changed it to `BigInt`

rafaelsamenezes (approved):

PR #1084, opened 2023-06-14
[Simplifier] bottom-up

rafaelsamenezes:
This changes the simplifier to use a bottom-up. The main difference is that we can assume that the expression operands are already simplified.

This is mainly a rebase from @mikhailramalho simplify branch (still missing the simplifications bit though).


mikhailramalho:
The code I've tried to implement was based on the instruction combine (e.g., InstCombinerImpl::visitAdd) from LLVM.

The arithmetic rebalancing is broken on that branch tho :/ We should implement the algorithm in `InstCombinerImpl::SimplifyAssociativeOrCommutative` which should bring some performance improvement.

rafaelsamenezes:
Yeah, I will focus on getting the bottom-up merged first. Although, it is being worse performance-wise for some reason (maybe there is a simplication missing).

pmatos:
The bottom-up / operands-first simplifier this PR introduced has since landed on `master` via #4282, which ports the simplifier to an LLVM-style operands-first walk and additionally adds the reassociation pass and expanded peepholes that this PR noted were "still missing." The file this PR modifies (`src/util/simplify_expr2.cpp`) has been replaced by `src/util/expr_simplifier.cpp` and `src/util/expr_reassociate.cpp`.

Closing as superseded by #4282.

PR #1105, opened 2023-06-19
Inductive step now checks the next unwind

### LINKED ISSUE #1092: K-Induction is breaking assertions inside loops.
ESBMC gives the wrong veredict for (with k-induction):
```c
#define __VERIFIER_assert(cond)                                                \
  do                                                                           \
  {                                                                            \
    if(!(cond))                                                                \
    {                                                                          \
      __ESBMC_assert(0, "");                                                   \
      abort();                                                                 \
    }                                                                          \
  } while(0)

#include <stdlib.h>
extern void abort(void);
#include <assert.h>
void reach_error()
{
  assert(0);
}

int main()
{
  int status = 0;

  while(__VERIFIER_nondet_int())
  {
    if(!status)
    {
      status = 1;
    }
    else if(status == 1)
    {
      status = 2;
    }
    else if(status >= 2)
    {
      status = 3;
    }
  }
  __VERIFIER_assert(status != 3);
  return 0;
}
```

The issue here is the `do-while` loop. Replacing it with a direct call to `__ESBMC_assert` seems to fix it.

rafaelsamenezes:
Fixes #1092 

This PR does two things:

1. Forces the max unwind of loops to be of k+1 (also, applies to the Assertions during symex).

2. Propagates esbmc Asserts into assumes after their execution. I am not so sure about this fix as I did it because reach error does it (and this fix the issue). @mikhailramalho any idea on the reasoning here?

### DIFF
--- a/regression/k-induction/github_1092/main.c
+++ b/regression/k-induction/github_1092/main.c
@@ -1,0 +1,41 @@
+#define __VERIFIER_assert(cond)                                                \
+  do                                                                           \
+  {                                                                            \
+    if(!(cond))                                                                \
+    {                                                                          \
+      __ESBMC_assert(0, "");                                                   \
+      abort();                                                                 \
+    }                                                                          \
+  } while(0)
+
+#include <stdlib.h>
+extern int __VERIFIER_nondet_int(void);
+extern void abort(void);
+#include <assert.h>
+void reach_error()
+{
+  assert(0);
+}
+
+int main()
+{
+  int status = 0;
+
+  while(__VERIFIER_nondet_int())
+  {
+    if(!status)
+    {
+      status = 1;
+    }
+    else if(status == 1)
+    {
+      status = 2;
+    }
+    else if(status >= 2)
+    {
+      status = 3;
+    }
+  }
+  __VERIFIER_assert(status != 3);
+  return 0;
+}
--- a/regression/k-induction/github_1092/test.desc
+++ b/regression/k-induction/github_1092/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--k-induction
+^VERIFICATION FAILED$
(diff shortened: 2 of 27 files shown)


mikhailramalho:
Can you explain why this is needed? The theory suggests that we only need to unwind k+1 in the inductive step.

Adding the assume after the assert shouldn't be needed, as the solver should be able to prune the states after the assert.

rafaelsamenezes:
> Can you explain why this is needed? The theory suggests that we only need to unwind k+1 in the inductive step.

@mikhailramalho I was hoping that you could give some light in here. It's tricky to debub , from my understanding:

1. All asserts in while or do-while were converted to assumptions, resulting in 0 VCC.

2. About the `__ESBMC_assert` propagating assumptions: I have no idea why (is it part of the theory somehow?) but it is needed to find the vulnerabilities. This also results in 0 VCC.

mikhailramalho:
> > Can you explain why this is needed? The theory suggests that we only need to unwind k+1 in the inductive step.
> 
> @mikhailramalho I was hoping that you could give some light in here. It's tricky to debub , from my understanding:
> 
> 1. All asserts in while or do-while were converted to assumptions, resulting in 0 VCC.

This seems like a bug in conversion (probably in symex?).

The inductive step:
1. unwinds all the loops k times
2. For k-1, all assertions are converted into assumes
3. The last k unwind the assertions as is.

The idea is to assume an increasing number of conditions and check if they still hold at the end of the k loop unwinds.

> 2. About the `__ESBMC_assert` propagating assumptions: I have no idea why (is it part of the theory somehow?) but it is needed to find the vulnerabilities. This also results in 0 VCC.

Yeah, they should be assumption k-1 times.

The bottom line is: you'll need to debug the generated formula, there is no way around it. I would start by looking at the generated SSA and trying to figure out if it's correct.

rafaelsamenezes:
I will close as it needs more investigations

PR #1111, opened 2023-06-20
[CUDA] refactor/modify the COM

XLiZHI:
In the old model there was a circular dependency problem where `cuda.h` and `cuda_runtime_api.h` files contained each other, which should be avoided in C++.

Purpose: Including too much unused code can lead to a large AST, which is not good for debugging, and we need to separate different functions to ensure simplicity.

`cuda.h`:  Interact directly with CUDA hardware, such as memory management and device synchronization.

`cuda_runtime_api.h`: The runtime API automatically handles many common operations, such as memory allocation.


src/cpp/library/CUDA/vector_types.h
```diff
@@ -310,10 +310,10 @@ __cuda_builtin_vector_align8(float2, float x; float y;);
 #endif /* !__CUDACC__ && !__CUDABE__ && _WIN32 && !_WIN64 */
 
   /*******************************************************************************
-*                                                                              *
-*                                                                              *
-*                                                                              *
-*******************************************************************************/
+  *                                                                              *
```
lucasccordeiro: We should remove these headers.

src/cpp/library/CUDA/vector_types.h
```diff
@@ -365,10 +365,10 @@ __cuda_builtin_vector_align8(float2, float x; float y;);
   typedef __device_builtin__ struct double4 double4;
 
   /*******************************************************************************
-*                                                                              *
-*                                                                              *
-*                                                                              *
-*******************************************************************************/
+  *                                                                              *
```
lucasccordeiro: We should remove these headers.

lucasccordeiro (changes requested):

lucasccordeiro (approved):
Many thanks, @XLiZHI.

src/cpp/library/CUDA/cuda.h
```diff
@@ -104,9 +79,9 @@ void __threadfence()
 void *Address[10];
 int Counter = 0;
 
-cudaError_t cudaMalloc(void **devPtr, size_t size)
+cudaError cudaMalloc(void **devPtr, size_t size)
```
kunjsong01: Would it be possible to double check the return type for cudaXYZ APIs? 
They seem to return `cudaError_t` instead of `cudaError`. 

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g37d37965bfb4803b6d4e59ff26856356

src/cpp/library/CUDA/cuda.h
```diff
@@ -47,12 +46,11 @@ void cudaInsertThread(pthread_t threadAux)
...
 }
 
-cudaError_t cudaThreadSynchronize()
-{
-  cudaError_t tmp;
+cudaError lastError;
 
-  //	if (cudaThreadList == NULL)
-  //		return CUDA_ERROR_OUT_OF_MEMORY;
+cudaError cudaThreadSynchronize()
```
kunjsong01: Doesn't this API return cudaError_t rather than cudaError? 
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__THREAD__DEPRECATED.html#group__CUDART__THREAD__DEPRECATED_1g86ba4d3d51221f18a9a13663e6105fc1

kunjsong01 (changes requested):

src/cpp/library/CUDA/cuda.h
XLiZHI: Yes, it should return `cudaError_t`, but in the API `cudaError_t `and `cudaError` are same and they do exactly the same thing. I will change them back to be consistent with the official documentation. Thanks for your advises.

kunjsong01 (approved):

lucasccordeiro:
Many thanks, @XLiZHI.

Issue #1103, opened 2023-06-19
Zero initialization rule


kunjsong01:
`cbmc/Templates36` tests class template, UDT(class) for type parameter. 

Failed with 
```
State 1 file main.cpp line 30 function main thread 0
----------------------------------------------------
Violated property:
  file main.cpp line 30 function main
  assertion z.some==0
  z.some == 0


VERIFICATION FAILED
```

but test.desc says it should be a `VERIFICATION SUCCESSFUL` 

ESBMC version: https://github.com/esbmc/esbmc/commit/880903638fc5a49505549d84b72f7a33b180af39

kunjsong01:
Preliminary analysis: 

Nothing wrong with template conversion. As confirmed by the TC with explicit initialization `z.some = 0;`: 
```cpp
#include<cassert>

// The orering of the following matters!
class FF
{
public:
  typedef int f;
};

// declare Z<>
template <typename T>
class Z;

// make an instance of Z<>
typedef Z<FF> my_Z;

// now define Z<>
template <typename T>
class Z
{
public:
  typename T::f some;
};

// trigger elaboration of Z<FF>
my_Z z;

int main()
{
  z.some = 0; // explicitly initialize it!
  assert(z.some==0);
  return 0;
}

```

Just need to double check whether we need to zero-initialize a class member of built-in type for C++03/11/14.

kunjsong01:
Same problem found in `cbmc/Templates36`, `cbmc/Templates20` and `cbmc/Templates39`. 

TODO: 
The objective of these TCs should be updated to test * **ONLY** * template, rather than a mix of template and zero initialization. 

Also, v2.1 is inconsistent with zero initialization rule: 
v2.1 passed the following TC (the original cbmc/Templates36): 
```cpp
#include<cassert>

// The orering of the following matters!
class FF
{
public:
  typedef int f;
};

// declare Z<>
template <typename T>
class Z;

// make an instance of Z<>
typedef Z<FF> my_Z;

// now define Z<>
template <typename T>
class Z
{
public:
  typename T::f some;
};

// trigger elaboration of Z<FF>
my_Z z;

int main()
{
  assert(z.some==0);
  return 0;
}

```
But failed with: 
```cpp
#include<cassert>

class FF
{
public:
  int f;
};

int main()
{
  FF obj;
  assert(obj.f==0);
  return 0;
}
```

kunjsong01:
[...]

PR #1113, opened 2023-06-22
Decouple TC20/36/39 from zero initialization

kunjsong01:
Decouple TC20/36/39 from zero initialization, and enabled these TCs. 

The objective of TCs should focus * ONLY * on template, rather than a mix of template and zero initialization, which is another issue covered by https://github.com/esbmc/esbmc/issues/1103 in our backlog.

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/1103_TC20/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/1103_TC20/main.cpp
@@ -1,0 +1,43 @@
+// This is the original TC for cbmc/Templates20, WITH zero init assumption
+#include <cassert>
+
+struct A
+{
+  int i;
+  A():i(10){}
+
+private:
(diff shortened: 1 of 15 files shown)


lucasccordeiro (approved):

fbrausse (commented):
Shouldn't these test cases be duplicated instead of modified? You're removing tests.

kunjsong01:
> Shouldn't these test cases be duplicated instead of modified? You're removing tests.

Thanks. Duplicates are added in https://github.com/esbmc/esbmc/pull/1113/commits/d738b4c4e5f738dd38c3736adba07cb317afecb7. I planned to do it when starting issue https://github.com/esbmc/esbmc/issues/1103. 
Now done in this PR instead.

fbrausse (approved):
LGTM. Thanks @kunjsong01.

PR #1114, opened 2023-06-22
Restore some smtlib functionality with external solver

fbrausse:
This PR is based on #1049 and does the following:
- the string given to `--smtlib-solver-prog` is now interpreted by `$SHELL -c` (default: `sh` if SHELL is not set). This allows for easier interaction, such that e.g. the following examples are now usable:
  - `--smtlib-solver-prog 'z3 -in'`
  - `--smtlib-solver-prog 'cvc4 -L smt2'`
  - `--smtlib-solver-prog 'tee in | yices-smt2 | tee out'` to dump input and output to separate files
- fixes communication with yices-smt2, which expects setting `:produce-models` before `(set-logic [...])`
- fixes interpreting the `(get-info :version)` response and unquotes solver's name and version
- handles early death of external process more gracefully by reading its final output and throwing an exception with that string; this mainly affects Yices since it dies after writing an smtlib2 (error [...]) message.
- allows specification of both, `--output` and `--smtlib-solver-prog`, simultaneously; this is done by preventing the textual "rendering" of the ASTs to modify state (in this case, by removing unnecessary internal state) and writing the same text to both files.
- adds the `dump()` method for smtlib's AST nodes, this is a bit hacky but I provided my reasoning why the way it's done for the intended debugging usage still is OK in a comment.

The smtlib backend still needs further fixes to be entirely usable again.

### DIFF
--- a/src/solvers/smtlib/smtlib_conv.cpp
+++ b/src/solvers/smtlib/smtlib_conv.cpp
@@ -233,7 +233,6 @@ smtlib_convt::process_emitter::process_emitter(const std::string &cmd)
       abort();
     }
   }
-#endif
   // Execution continues as the parent ESBMC process. Child dying will
   // trigger SIGPIPE or an EOF eventually, which we'll be able to detect
   // and crash upon.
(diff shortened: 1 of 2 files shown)


lucasccordeiro:
This is very good, @fbrausse ;-)

rafaelsamenezes (approved):
LGTM

lucasccordeiro (approved):

lucasccordeiro:

Issue #1089, opened 2023-06-14
ESBMC crashes: ERROR: Couldn't convert expression in unrecognised format


tihanyin:
ESBMC 7.2 crashes for the following small C program:

```c
//FormAI DATASET v0.1 Category: System event logger ; Style: systematic
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

// Define the maximum length of the log message
#define MAX_LOG_MSG_LEN 100

// Define the structure for the log event
typedef struct LogEvent {
    time_t timestamp;
    char message[MAX_LOG_MSG_LEN];
} LogEvent;

// Define the function to log the event
void logEvent(char* message) {
    // Create the log event and set the timestamp
    LogEvent event;
    event.timestamp = time(NULL);

    // Copy the message into the log event, making sure it doesn't exceed MAX_LOG_MSG_LEN
    int len = snprintf(event.message, MAX_LOG_MSG_LEN, "%s", message);
    if (len >= MAX_LOG_MSG_LEN) {
        event.message[MAX_LOG_MSG_LEN - 1] = '\0';
    }

    // Open the log file for appending
    FILE* logFile = fopen("system.log", "a");
    if (logFile == NULL) {
        printf("Error opening log file.\n");
        return;
    }

    // Write the log event to the file
    fprintf(logFile, "[%ld]: %s\n", event.timestamp, event.message);

    // Close the log file
    fclose(logFile);
}

// Define the main function
int main() {
    // Log some example events
    logEvent("System started.");
    logEvent("User logged in.");
    logEvent("File saved successfully.");
    logEvent("Error: Unable to connect to database.");

    return 0;
}

```

ESBMC crash output:
```
esbmc  DATA8/FormAI_81911.c --overflow
ESBMC version 7.2.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing
Converting
Generating GOTO Program
GOTO program creation time: 0.120s
GOTO program processing time: 0.001s
Starting Bounded Model Checking
Symex completed in: 0.003s (135 assignments)
Slicing time: 0.001s (removed 67 assignments)
Generated 16 VCC(s), 12 remaining after simplification (68 assignments)
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
ERROR: Couldn't convert expression in unrecognised format
code_printf
* operands : 
  0: address_of
    * pointer_obj : index
        * source_value : member
[...]

PR #1115, opened 2023-06-22
Bug fixing for printf return value

ChenfengWei0:
Related issue: #1089 

example
```c++
char *data = "123456";
int x = printf("%s", data);
assert(x == 6); // should hold
```

- However, the return value is always zero in Windows, probably due to the different symbol names in Windows (e.g. `__stdio_common_vfprintf`). Also, the error log shows a function body missing:

```sh
no body for function __acrt_iob_func
```
- Related info: [The printf and scanf family of functions are now defined inline - Microsoft](https://learn.microsoft.com/en-us/cpp/porting/visual-cpp-change-history-2003-2015?redirectedfrom=MSDN&view=msvc-170#BK_CRT)

may be fixed in another PR


lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:

src/util/array2string.cpp
```diff
@@ -0,0 +1,21 @@
...
+  if(src.type.id() != irept::id_array)
+    return;
+
+  std::string value_str;
+  size_t cnt = 0;
+  forall_operands(it, src.value)
+  {
+    std::string op_str = it->cformat().as_string();
+    if(cnt < src.value.operands().size() - 1)
+      value_str.push_back(op_str[1]);
```
kunjsong01, line 14: Would it be possible to add a comment here explaining why "op_str[1]"?

regression/cstd/printf_bug_1/main.c
```diff
@@ -0,0 +1,10 @@
+#include <stdio.h>
+#include <assert.h>
+int main()
+{
+	char *s = "abcde123415";
```
kunjsong01, line 5: indentation formatting

kunjsong01 (commented):
just some nitpicks

src/util/array2string.cpp
ChenfengWei0: Sure. I will add some comments. Basically it's the reversal of the `string2array()`:
```c++
    if(ch >= 32 && ch <= 126)
    {
      char ch_str[2];
      ch_str[0] = ch;
      ch_str[1] = 0;

      op.cformat("'" + std::string(ch_str) + "'");
    }
```

Issue #1021, opened 2023-05-17
SMT2 interface


rafaelsamenezes:
What is the current status of our SMT2 output? It would be great if we could produce formulas that could be sent directly to other solvers via FILE or STDIN. In that way we don't rely on the solvers being compiled with ESBMC which would improve supporting solvers with restricted licenses, no Windows support, or no API. 

AFAIK we rely on the solvers to produce the formulas... however, it would be great if we could be solver agnostic here.

rafaelsamenezes:
There is also an issue that we are producing large files. Comparing the output of the same category has issues: 

- AWS:  https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_add_size_saturating_harness_negated.i (ok) / https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_iter_begin_harness_negated.i (crashes)
- ECA generates huge files: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/eca-rers2012/Problem03_label08.c (generated a 200MB file)

I have used the flags ` --smtlib --output file.smt2 --unwind 20  --smt-formula-only` in the smt-dump branch.

ibnyusuf:
You mention "smt-dump" branch. Does this mean that the ability to output smtlib (directly, not via Z3) is not available on `master`?

rafaelsamenezes:
> You mention "smt-dump" branch. Does this mean that the ability to output smtlib (directly, not via Z3) is not available on master?

It is available. However, it will crash if you don't give a supported solver. The smt-dump branch allows to use the interface to just generate the formula and stop.

PR #1117, opened 2023-06-23
Fix symbol names and emission of AST node types in the smtlib backend

fbrausse:
This PR
- fixes emitting quoted SMT-LIB2 symbols (replacing `/ -> //`, `\ -> /b` and `| -> /p`), and
- restores the correspondence between the enum defining the AST node's kind and the SMT-LIB2 symbol table.

Thereby it increases the ratio of passed regression tests with the default backend set to the smtlib with Z3 from [58%](https://github.com/esbmc/esbmc/files/11846920/base.txt) to [75%](https://github.com/esbmc/esbmc/files/11846781/master.txt) on my machine. These numbers come from local tests of a debug build as the CI does not test the smtlib backend. The tests were run with `env ESBMC_OPTS="--default-solver=smtlib --smtlib-solver-prog 'z3 -in'" ctest -j8 --timeout 300` and Z3 v4.11.2 on Linux. For reference, when using the Z3-backend directly as default, the ratio of passed tests is [86%](https://github.com/esbmc/esbmc/files/11847071/z3.txt).

A note of caution on these numbers: I observed the 8 parallel smtlib runs occasionally use all main memory on my machine leading it to do swapping and possibly to more timeouts than if run non-parallelly.

Related: #1021

### DIFF
--- a/src/solvers/smtlib/smtlib_conv.cpp
+++ b/src/solvers/smtlib/smtlib_conv.cpp
@@ -13,86 +13,103 @@
 
 /** Mapping of SMT function IDs to their names. */
 static const std::array smt_func_name_table = {
-  "hack_func_id",
-  "invalid_func_id",
-  "int_func_id",
-  "bool_func_id",
-  "bvint_func_id",
-  "real_func_id",
-  "symbol_func_id",
-  "+",
-  "bvadd",
-  "-",
-  "bvsub",
-  "*",
-  "bvmul",
-  "/",
-  "bvudiv",
-  "bvsdiv",
-  "%",
-  "bvsmod",
-  "bvurem",
-  "shl",
-  "bvshl",
-  "bvashr",
-  "-",
-  "bvneg",
-  "bvlshr",
-  "bvnot",
-  "bvnxor",
-  "bvnor",
-  "vnand",
-  "bvxor",
-  "bvor",
-  "bvand",
-  "=>",
-  "xor",
-  "or",
-  "and",
-  "not",
-  "<",
-  "bvslt",
-  "bvult",
-  ">",
-  "bvsgt",
-  "bvugt",
-  "<=",
-  "bvsle",
-  "bvule",
-  ">=",
-  "bvsge",
-  "bvuge",
-  "=",
-  "distinct",
-  "ite",
-  "store",
-  "select",
-  "concat",
-  "extract",
-  "int2real",
-  "real2int",
-  "is_int",
-  "fneg",
-  "fabs",
-  "fp.isZero",
-  "fp.isNaN",
-  "fp.isInfinite",
-  "fp.isNormal",
-  "fp.isNegative",
-  "fp.isPositive",
-  "fp.eq",
-  "fp.add",
-  "fp.sub",
-  "fp.mul",
-  "fp.div",
-  "fp.fma",
-  "fp.sqrt",
-  "RNE RoundingMode",
-  "RTZ RoundingMode",
-  "RTP RoundingMode",
-  "RTN RoundingMode",
-  "bv2fp_cast",
-  "fp2bv_cast",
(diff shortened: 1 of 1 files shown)


lucasccordeiro:
@fbrausse: shall we add some regression tests for the smtlib backend?

fbrausse:
@lucasccordeiro What did you have in mind? We could duplicate those that are solver-agnostic into a new `regression/smtlib/$solver_prog` directory and adjust their `test.desc` file, though the number of combinations with different external solvers (and potentially their cmdline options) is significant. Another option might be to extend the `regression/testing_tool.py` script, though I don't quite see the benefit given that we can just set the `ESBMC_OPTS` envvar as done in my tests above.

lucasccordeiro:
> We could duplicate those that are solver-agnostic into a new regression/smtlib/$solver_prog directory and adjust their test.desc file, though the number of combinations with different external solvers (and potentially their cmdline options) is significant. 

We could consider this approach with a few duplicated regression tests using the SMT solver Z3.

What do you think?

@rafaelsamenezes : do you have suggestions here?

Let's keep an SMT-lib backend working ;-)

fbrausse:
Another option could be to extend the CI scripts with the envvar, e.g. here https://github.com/esbmc/esbmc/blob/b9e72cc1363ab33d267f15f46d39170310e633f7/.github/workflows/build.yml#L69

rafaelsamenezes:
> Another option might be to extend the regression/testing_tool.py script, though I don't quite see the benefit given that we can just set the ESBMC_OPTS envvar as done in my tests above.

Hm, we could create a new CMake function for this kind of test: https://github.com/esbmc/esbmc/blob/master/regression/CMakeLists.txt#L25

This way we wouldn't need to duplicate tests (just move around) and it would work as usual from ctest.

lucasccordeiro:
Thanks, @rafaelsamenezes.

@fbrausse: What do you think about Rafael's suggestions? Can I ask you whether you could implement that in this PR?

It would be great to keep the regression of the SMT-LIB backend.

lucasccordeiro (approved):

lucasccordeiro:
We could also consider this regression suite in another PR.

fbrausse:
> We could also consider this regression suite in another PR.

That's a good idea. I actually don't have a concrete solution right now, that's why I opened #1118 to address these concerns. I hope that's OK.

lucasccordeiro:
Many thanks, @fbrausse.

fbrausse:
Just as a note, the corresponding numbers for Yices with this PR are [74%](https://github.com/esbmc/esbmc/files/11848564/master-yices.txt) for the smtlib backend and  [94%](https://github.com/esbmc/esbmc/files/11848571/yices.txt) with `--default-solver=yices`. So the Yices backend seems to behave better than Z3...

lucasccordeiro:
@fbrausse: That's very good; shall we document it somewhere?

fbrausse:
> @fbrausse: That's very good; shall we document it somewhere?

That would be good, yes. Though these numbers here are on my local machine with my local memory limitations, so they're probably not the right ones for the repository. Though I wouldn't want to put more things on @rafaelsamenezes's TODO list, to my mind the best place to reliable obtain such numbers are the FM servers. Unfortunately we can't trigger them to update these stats easily right now. Is there another way to reliably get these regression suite coverage stats from somewhere? The CI has in the past been somewhat unstable wrt. timeouts and such.

rafaelsamenezes:
> Is there another way to reliably get these regression suite coverage stats from somewhere?

Not really, and the FM proxy will probably still be shut for a few weeks at least. So, I will have to come up with a script to set up local jobs. 

Other way would be to configure another machine:
- A physical/vm machine.
- A cloud machine that lets us do some OS setup (to install benchexec).

Issue #1085, opened 2023-06-14
Level2 renaming of VLA array size creates invalid if2t expressions


fbrausse:
Investigating the error mentioned in https://github.com/esbmc/esbmc/pull/1083#issuecomment-1590955655 for Z3 turned out to be due to an invalid if2t expression being translated to SMT:
```
if
* cond : symbol
    * name : goto_symex::guard
    * renamelev : Level 2
    * level1_num : 0
    * level2_num : 12
    * thread_num : 0
    * node_num : 0
    * type : bool
* true_value : symbol
    * name : c:gh-1070.c@421@F@main@url
    * renamelev : Level 2
    * level1_num : 1
    * level2_num : 2
    * thread_num : 0
    * node_num : 0
    * type : array
      * subtype : signedbv
          * width : 8
      * array_size : symbol
          * name : c:@F@main::$tmp::return_value$_strlen$1
          * renamelev : Level 0
          * level1_num : 0
          * level2_num : 0
          * thread_num : 0
          * node_num : 0
          * type : unsignedbv
            * width : 64
      * size_is_infinite : false
* false_value : symbol
    * name : c:gh-1070.c@421@F@main@url
    * renamelev : Level 2
    * level1_num : 1
    * level2_num : 3
    * thread_num : 0
    * node_num : 0
    * type : array
      * subtype : signedbv
          * width : 8
      * array_size : symbol
          * name : c:@F@main::$tmp::return_value$_strlen$1
          * renamelev : Level 0
          * level1_num : 0
          * level2_num : 0
          * thread_num : 0
          * node_num : 0
          * type : unsignedbv
            * width : 64
      * size_is_infinite : false
* type : array
  * subtype : signedbv
      * width : 8
  * array_size : constant_int
      * value : 0
      * type : unsignedbv
        * width : 64
  * size_is_infinite : false
```
The type of the if2t-expression has been mangled and does no longer correspond to those of either branches. The if2t expression itself is constructed with properly aligned types, however, this mangling happens in https://github.com/esbmc/esbmc/blob/ee286d92a125aa7fc3a966fc4ca85dfcd057ef9b/src/goto-symex/renaming.cpp#L149 since `it->second.constant` is the non-nil
```
constant_int
* value : 0
* type : unsignedbv
  * width : 64
```

PR #905, opened 2023-02-20
Prepare CHERI-C support

fbrausse:
This PR prepares for support the CHERI-C semantics, which will be part of another PR. Essentially, this PR does:

- add support for CMake to recognize and build against [CHERI-Clang](https://github.com/CTSRD-CHERI/llvm-project), which is a fork of LLVM intended to support the new semantics for MIPS64 and RISC-V CHERI-enabled architectures. There is another [fork](https://git.morello-project.org/morello/llvm-project/-/tree/morello/dev/) by ARM that supports Aarch64 targets. These compilers add a few new API constructs such as `clang::BuiltinType::UIntCap` and `clang::CK_PointerToCHERICapability` which this PR handles - behind #ifdefs depending on whether ESBMC is built against CHERI-Clang or the regular one.
- extend the `configt` and the embedded `ansi_ct` to:
  - know about whether ESBMC is supposed to verify standard C or CHERI-C,
  - what mode of CHERI-C: `purecap` or `hybrid` (this is set by the new cmd-line option `--cheri (hybrid|purecap)`, default: off),
  - whether or not compression of CHERI capabilities is enabled (cmd-line flag `--cheri-uncompressed`, default: off)
  - derive the size of capability-like types from these settings (in CHERI-C they are usually twice the size of regular pointers), i.e. there now is a distinction between `pointer_width()` and `capability_width()`.
- fix various places throughout the code base that previously assumed pointer-related sizes to correspond to `word_size` or `int_width` or similar. Actually, there is no correspondence between any C types and `word_size`. Such assumption should be removed throughout: `sizeof(void *)` and `sizeof(size_t)` are no longer related.
- introduce new `typet` for the new types `[u]intcap_t` (the CHERI-capability equivalent for `[u]intptr_t`)
- add a flag `can_carry_provenance` to irep1 types that corresponds to the similarly named flag from CHERI-Clang; this flag basically says that operations on expressions of such types should copy the associated validity tag.
- add a new set of pre-compiled GOTO libraries for the the CHERI hybrid and purecap modes (now we have 12),
- passes a bunch of new flags to the Clang frontend when the CHERI support is enabled (these unfortunately vary between platforms, see `clang_c_language.cpp` for details)
- add a new pseudo expression `pointer_capability2t` that, applied to capability-typed sub-expression, should yield the capability metadata in symbolic form; this functionality is not used in this PR and will be extended in the next one
- add support for the new metadata to the `smt_memspace`, though this is not complete, yet

PR #1121, opened 2023-06-24
if2t's type can differ from either side's

fbrausse:
This addresses discrepancies in the translated SMT solver's types between that of the if2t and those of either of its side's. In regression/k-induction-parallel/invert_string for instance, the Z3 and CVC4 backends aborted due to such a construct where the if2t's array size is a small constant but either side's one is a symbol of a larger type.


lucasccordeiro (approved):
LGTM.

Thanks for submitting this PR, @fbrausse  ;-)

fbrausse:
Thanks @lucasccordeiro! I would suggest to run this over SV-COMP before merging, the added type-casts for if2t expressions will probably affect some things. For instance, in the regression test mentioned initially, a new equality is being asserted and for Boolector and other users of `array_convt` the terms generated for "cond ? array1 : array2" also changed.

rafaelsamenezes:
> I would suggest to run this over SV-COMP before merging

I will start a run Monday

fbrausse:
> > I would suggest to run this over SV-COMP before merging
> 
> I will start a run Monday

Thanks @rafaelsamenezes! If you could let me know how it's done with the current setup, I can also do it myself (need to run #905 over it as well).

lucasccordeiro:
Thanks, @fbrausse: are you still able to run PRs over the SV-COMP benchmarks?

rafaelsamenezes:
> Thanks @rafaelsamenezes! If you could let me know how it's done with the current setup

The current setup is quite limited and cannot be used remotely.

fbrausse:
> Thanks, @fbrausse: are you still able to run PRs over the SV-COMP benchmarks?

I think so. Will keep you updated.

mikhailramalho:
This should be fixed where the if is being created, not at the backend.

fbrausse:
@mikhailramalho I know at least one place where it's created, it's described in <https://github.com/esbmc/esbmc/issues/1085>. But unfortunately I don't know enough about the renaming code to fix it there. Seeing that the other solver backends (via array_conv) sidestep this problem basically by ignoring the type information in the `if2t` and instead only relying on that of the `true_value` branch, I thought this would be a better way to handle it. But this one is not important for my other work, so I'm fine with solving this issue another way.

fbrausse:
For the record, this didn't change SV-COMP stats.

Still closing this PR, since it's a workaround that was better solved with #1123. Thanks for reviewing.

Issue #1070, opened 2023-06-11
ESBMC crash - ERROR: Select operation applied to tuple


tihanyin:
ESBMC 7.2 crash for this small C program with `ERROR: Select operation applied to tuple
Aborted (core dumped)`

```c
//FormAI DATASET v0.1 Category: Building a HTTP Client ; Style: high level of detail
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/socket.h>
#include <arpa/inet.h>
#include <unistd.h>
#define BUFSIZE 4096

int main(int argc, char* argv[]) {
    if (argc < 2) {
        printf("Usage: %s <url>\n", argv[0]);
        return 1;
    }

    const char delimiter[2] = "/";
    char* host, *path;
    char url[strlen(argv[1])];
    strcpy(url, argv[1]);

    strtok(url, delimiter); // http:
    strtok(NULL, delimiter); // empty
    strtok(NULL, delimiter); // host

    host = strtok(NULL, delimiter); // path
    path = "/";
    while (host != NULL) {
        path = strcat(path, "/");
        path = strcat(path, host);
        host = strtok(NULL, delimiter);
    }

    char request[BUFSIZE];
    sprintf(request, "GET %s HTTP/1.1\r\n"
                      "Host: %s\r\n"
                      "Connection: close\r\n\r\n"
            , path, argv[1]);

    struct sockaddr_in server;
    memset(&server, 0, sizeof(server));
    server.sin_family = AF_INET;
    server.sin_addr.s_addr = inet_addr(argv[1]);
    server.sin_port = htons(80);

    int sock = socket(AF_INET, SOCK_STREAM, 0);
    if (sock < 0) {
        perror("Socket error.");
        return 1;
    }

    if (connect(sock, (struct sockaddr*) &server, sizeof(server)) < 0) {
        perror("Connection error.");
        return 1;
    }

    send(sock, request, strlen(request), 0);

    char buffer[BUFSIZE];
    while (recv(sock, buffer, BUFSIZE, 0) > 0) {
        printf("%s", buffer);
        memset(buffer, 0, BUFSIZE);
    }

    close(sock);
    return 0;
}
```

ESBMC output:

```
esbmc --unwind 1 DATA3/FormAI_31168.c
ESBMC version 7.2.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing
Converting
Generating GOTO Program
[...]

PR #1123, opened 2023-06-25
rename_type() should also recurse into operands

### LINKED ISSUE #60: terminate called after throwing an instance of 'array_type2t::dyn_sized_array_excp*'
ESBMC fails for this particular program when we call it as:

esbmc main.c --unwind 10
```
# include <stdio.h>
# include <stdlib.h>
# include <assert.h>

typedef int __nodetype;

typedef struct node {
    __nodetype key;
    struct node *next;
} NODE;

static unsigned int vertices;

int get_graph_size(char *argv[]) {
  unsigned int size=0;
  int a, b;
  FILE *fp;
  fp = fopen(argv[1],"r");
  if (!fp) {
    printf("Failed to open the file %s.\n",argv[1]);
    return -1;
  }
  while (!feof(fp)) {
    fscanf(fp,"%d%d", &a, &b);
    size = a > size ? a : size;
    size = b > size ? b : size;
  }

  ++size;
  printf("graph size: %d\n", size);

  return size;
}

void insert_node(NODE* list[], int a, int b){
    NODE* l = (NODE*)malloc(sizeof(NODE));
    if (list[a] == NULL) {
        l->key = b;
        l->next = NULL;
    } else {
        l->key = b;
        l->next = list[a];
    }
    list[a] = l;
}

void print_adjacent_list(NODE *list[]){
  int i;
  NODE *tmp;
  printf("\nPrinting adjacent list...\n\n");
  for(i=0; i<vertices; i++) {
    if (list[i]!=NULL) {
      printf("(%d) ==> %d ", i, list[i]->key) ;
      tmp = list[i]->next;
      while (tmp != NULL) {
        printf("==> %d  ", tmp->key);
        tmp = tmp->next;
      }
    }
    printf("\n");
  }
}

void print_adjacent_matrix(__nodetype matrix[vertices][vertices]) {
  int i, j;
  printf("\nPrinting adjacent matrix...\n\n");
  for(i=0; i<vertices; i++) {
    for(j=0; j<vertices; j++) {
      printf("%d ", matrix[i][j]);
    }
    printf("\n");
  }
}

int create_adjacent_list(char *argv[], NODE *list[]) {
  int i, a, b, prev_a, prev_b;
  FILE *fp;
  fp = fopen(argv[1],"r");

  for(i=0; i<vertices; i++)
    list[i]=NULL;

  if (!fp) {
    printf("Failed to open the file %s.\n",argv[1]);
    return -1;
  }

  while (!feof(fp)) {
    fscanf(fp,"%d%d", &a, &b);
    if (prev_a != a || prev_b != b) {
          printf("a: %d, b: %d\n", a, b);
      insert_node(list, a, b);
    }
    prev_a = a;
    prev_b = b;
  }
  return 0;
}

int create_adjacent_matrix(char *argv[], __nodetype matrix[vertices][vertices]) {
  int i, j, a, b, prev_a, prev_b;
  FILE *fp;
  fp = fopen(argv[1],"r");

  for(i=0; i<vertices; i++)
    for(j=0; j<vertices; j++)
      matrix[i][j] = 0;

  if (!fp) {
    printf("Failed to open the file %s.\n",argv[1]);
    return -1;
  }

  while (!feof(fp)) {
    fscanf(fp,"%d%d", &a, &b);
    if (prev_a != a || prev_b != b) {
          printf("a: %d, b: %d\n", a, b);
      matrix[a][b]=1;
    }
    prev_a = a;
    prev_b = b;
  }
  return 0;
}

  vertices = get_graph_size(argv);

  NODE *adjacent_list[vertices];
  __nodetype adjacent_matrix[vertices][vertices];

  create_adjacent_list(argv, adjacent_list);
  print_adjacent_list(adjacent_list);

  create_adjacent_matrix(argv, adjacent_matrix);
  print_adjacent_matrix(adjacent_matrix);

  return 0;
}
```

### LINKED ISSUE #1085: Level2 renaming of VLA array size creates invalid if2t expressions

fbrausse:
Another take on fixing the failing cases from #1070 as suggested in https://github.com/esbmc/esbmc/pull/1121#issuecomment-1605586220, and this time also a fix for #1085 and #60 (though --unwind 10 takes too long, I've reduced it to 3).

`goto_symex_statet::rename_type()` did only consider the expression's type and
its subtypes and not sub-expressions as well. This causes the sub-expressions
for instance in the if2t `cond ? array1 : array2`, to not be considered,
triggering issues with the Z3 and CVC4 backends due to mismatching bv-widths
for array sizes. Something that's supposedly fixed up by the internal
fixup_renamed_type() for other kinds of expressions, but for some reason that's
only invoked for symbols, not ifs, and for renaming exprs, not types.
So the the recursion is put into the public `rename_type()` interface instead.

The other backends using array_conv have not been affected because they simply
ignore the if2t's type annotation and go with that of the true_value sub-expr.
As now they act on arrays of different sizes, I would again suggest to test this on SV-COMP.

**Edit:** Turns out the other backends have been affected to some degree by the missing type renaming in sub-exprs.


mikhailramalho (commented):
Yeah, that makes more sense!

I wonder if we could get rid of the the rename_type call and force it whenever we call rename. There might be some performance penalty tho

lucasccordeiro (approved):

@rafaelsamenezes: can you run this PR over the SV-COMP benchmarks too?

fbrausse:
Master (ac9ec4d1362f86b5505095b90e1549e2992c9d96):
```
  correct:           14370
    correct true:     8158
    correct false:    6212
  incorrect:            60
    incorrect true:     30
    incorrect false:    30
  unknown:           12029
  Score:             21088 (max: 43043)
```
This PR (591b023de1264f758d8651ac9e793ce54e4348af):
```
  correct:           14374
    correct true:     8161
    correct false:    6213
  incorrect:            60
    incorrect true:     30
    incorrect false:    30
  unknown:           12025
  Score:             21095 (max: 43043)
```
**Edit:** This is for 30s (on the fm08 VM).

lucasccordeiro:
Great! Thanks for submitting this PR, @fbrausse.

PR #1124, opened 2023-06-25
Embed CPP headers into ESBMC

rafaelsamenezes:
This PR:

1. Adds a recursive functionality into mangle. This was needed as using `WILDCARD *` was adding folders (even with `LIST_DIRECTORIES false`.
2. Adds the directory into the include list for Clang CPP mode (unless --no-optimized-cpp-includes)
3. Remove the library hack from the testing_tool.
4. Apparently this also fixes a few CPP tests.


lucasccordeiro (approved):
Excellent, @rafaelsamenezes.

Thanks for submitting this PR ;-)

kunjsong01:
Does this patch mean we don't need to specify -I <cpp_lib>/libaries" in the test.desc file?

rafaelsamenezes:
> Does this patch mean we don't need to specify -I <cpp_lib>/libaries" in the test.desc file? 

Yes, we do not.

kunjsong01:
> > Does this patch mean we don't need to specify -I <cpp_lib>/libaries" in the test.desc file?
> 
> Yes, we do not.

cool thanks! 

Also, if `no-optimized-cpp-includes` is specified, then it means we use the system's headers instead of our own C++ OMs. Is that correct?

rafaelsamenezes:
> > > Does this patch mean we don't need to specify -I <cpp_lib>/libaries" in the test.desc file?
> > 
> > Yes, we do not.
> 
> cool thanks! 
> 
> Also, if `no-optimized-cpp-includes` is specified, then it means we use the system's headers instead of our own C++ OMs. Is that correct? 

That's correct

src/clang-cpp-frontend/clang_cpp_language.cpp
```diff
@@ -36,6 +38,9 @@ void clang_cpp_languaget::force_file_type()
   std::string clangstd = cppstd.empty() ? "-std=c++03" : "-std=c++" + cppstd;
   compiler_args.push_back(clangstd);
 
+  if(!config.options.get_bool_option("no-optimized-cpp-includes"))
+    compiler_args.push_back("-I" + esbmc_cpp_includes());
```
fbrausse: `-I` is not the right option, it should be `-isystem`. See also the discussion in <https://github.com/esbmc/esbmc/pull/834#discussion_r937853269>.

fbrausse: In my opinion, this should depend on `--no-library` like the OM header path for C does, see <https://github.com/esbmc/esbmc/blob/db4d28d4bfdbb8a80f118c15ba6fa4f5f70ad0ea/src/c2goto/cprover_libc_sources.cpp#L86-L90>

As another comment: I don't think the `force_file_type()` method is the right place for placing standard include paths, though I also don't see a better one right now.

src/clang-cpp-frontend/optimized_includes/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+
+mangle(mangled_cpp "${CMAKE_SOURCE_DIR}/src/cpp/library"
+    RECURSIVE 1
+    MACRO ESBMC_FLAIL
+    ACC_HEADERS_INTO ${CMAKE_CURRENT_BINARY_DIR}/cpp_includes.h
+    LIST_DIRECTORIES FALSE)
```
fbrausse: This is generating C symbols. They are required to be prefixed in order to be compatible with libraries, the system, etc. Please include a `PREFIX esbmc_libcpp_header_` or similar, see e.g. https://github.com/esbmc/esbmc/blob/db4d28d4bfdbb8a80f118c15ba6fa4f5f70ad0ea/src/c2goto/headers/CMakeLists.txt#L19-L22

fbrausse (changes requested):
Very nice to see this take shape! However, I don't quite understand what's "optimized" about these headers and I would like to request a few changes.

scripts/flail.py
```diff
@@ -126,7 +126,11 @@ def obtain_var_name(self):
                                       .replace('.goto', '_buf')
                                       .replace('.txt', '_buf')
                                       .replace('buildidobj', 'buildidstring')
-                                      .replace('-', '_'))
+                                      .replace('-', '_')
+                                      # CPP THINGS
+                                      .replace('vector', '__esbmc_om_vector') 
+                                      .replace('new', '__esbmc_om_new')
+                                      .replace('ctime', '__esbmc_om_ctime'))
```
fbrausse: Could you explain why these are necessary?

src/clang-cpp-frontend/clang_cpp_language.cpp
```diff
@@ -95,3 +100,30 @@ bool clang_cpp_languaget::from_type(
...
+const std::string &clang_cpp_languaget::esbmc_cpp_includes()
+{
+  // Dump clang headers into a temporary directory
+  static bool dumped = false;
+  /* About the path being static:
+   * The static member 'dumped' above is used to check whether the headers were
+   * ever extracted before. This guarantees that the same path is used
+   * during a run. And no more than one is required anyway */
+  static auto p =
+    file_operations::create_tmp_dir("esbmc-headers-%%%%-%%%%-%%%%");
```
fbrausse: Would it be possible to give this a less generic prefix like `esbmc-cpp-headers` or similar?

rafaelsamenezes:
> However, I don't quite understand what's "optimized" about these headers

Hm, @lucasccordeiro and @kunjsong01 might be able to give more details. In shorts, these OMs at include level are supposed to be "optimized" in the sense of adding pre and post conditions into the C++ frontend (and be simpler overall). See section 4 of this paper: https://onlinelibrary.wiley.com/doi/10.1002/stvr.1793

src/clang-cpp-frontend/clang_cpp_language.cpp
rafaelsamenezes: Done

rafaelsamenezes: I agree, I also didn't find a more suitable place.

Maybe we should change the function name to something else

scripts/flail.py
rafaelsamenezes: It isn't. I didn't notice the PREFIX bit of your previous comment

src/clang-cpp-frontend/clang_cpp_language.cpp
rafaelsamenezes: Done

src/clang-cpp-frontend/optimized_includes/CMakeLists.txt
rafaelsamenezes: Done

fbrausse:
> In shorts, these OMs at include level are supposed to be "optimized" in the sense of adding pre and post conditions into the C++ frontend (and be simpler overall). See section 4 of this paper: https://onlinelibrary.wiley.com/doi/10.1002/stvr.1793

Thanks @rafaelsamenezes, I was just wondering what "optimized" refers to since it could be a bit misleading. I read section 4 and while it does explain how the C++ stdlib is modelled, there seems to be no mentioning of it being optimized in some sense. Of course it could be considered more optimal for ESBMC's operation than using the compiler-provided headers, but that argument (hopefully) is true for each operational model we have - otherwise there is no reason to keep that particular OM.

src/clang-cpp-frontend/clang_cpp_language.cpp
fbrausse: > Maybe we should change the function name to something else

Probably it would be good to have the include path handling moved out. It is probably only going to get more involved later on. I'm fine with this here for now, though. Thanks.

lucasccordeiro:
@fbrausse: can I ask you whether you have further comments here?

scripts/cmake/Utils.cmake
```diff
@@ -48,9 +49,14 @@ function(mangle output_c dir)
     set(single_file_desc "${MANGLE_WILDCARD} in ${dir}/")
     file(GLOB inputs RELATIVE ${dir} CONFIGURE_DEPENDS ${dir}/${MANGLE_WILDCARD} ${MANGLE_UNPARSED_ARGUMENTS})
   else()
-    set(inputs ${MANGLE_UNPARSED_ARGUMENTS})
-    list(JOIN ", " inputs single_file_desc)
-    set(single_file_desc "${inputs} in ${dir}")
+    if(MANGLE_RECURSIVE)
+      set(single_file_desc "${MANGLE_WILDCARD} in ${dir}/")
+      file(GLOB_RECURSE inputs RELATIVE ${dir} CONFIGURE_DEPENDS ${MANGLE_UNPARSED_ARGUMENTS} "${dir}/*")
```
fbrausse: In this branch, MANGLE_WILDCARD is not set. If you keep this functionality here, I recommend to remove it from the `single_file_desc` and also to update the comment on the WILDCARD keyword parameter above to reflect what's going on. It looks a bit inconsistent to have globbing when no WILDCARD parameter is specified.

fbrausse (commented):
> can I ask you whether you have further comments here?

@lucasccordeiro None, besides the use of "optimized" and the one above.

lucasccordeiro:
We could replace `optmized` with `abstraction`.

The C++ operational models typically contain only pre- and post-conditions and some code to simulate their behaviour.

scripts/cmake/Utils.cmake
rafaelsamenezes: Hm, I will invert it (check for MANGLE_RECURSIVE first) and also use the wildcard there.

fbrausse: Many thanks!

fbrausse (approved):
LGTM, thanks @rafaelsamenezes. With this PR I can finally run the C++ regression tests.

lucasccordeiro:
Excellent!

Many thanks, @rafaelsamenezes.

kunjsong01:
@rafaelsamenezes What's the difference between `no-abstracted-cpp-includes` and `no-library`? 
should they be used simultaneously if I want to use system's headers? something like: `./esbmc main.cpp --no-abstracted-cpp-includes --no-library`

rafaelsamenezes:
The difference is that `--no-abstracted-cpp-includes` does not disable the c2goto models while `--no-library` disables both

kunjsong01:
> The difference is that `--no-abstracted-cpp-includes` does not disable the c2goto models while `--no-library` disables both

@rafaelsamenezes Thanks!

PR #1131, opened 2023-06-27
Added CUDA and Qt/QtCore include paths in CPP

rafaelsamenezes:
This PR:

1. Add more abstract includes (with -I) for CUDA and QT.
2. Adds a function to create the parent folder when needed. std::ofstream was not creating some of the headers because of this.
3. Removed all hard coded paths for -I.
4. Fixed wrong calling of "-isystem={folder}" to "-cxx-isystem {folder}"
5. Remove duplicated OMs for  cstring


lucasccordeiro (approved):

src/clang-cpp-frontend/clang_cpp_language.cpp
```diff
@@ -44,9 +44,9 @@ void clang_cpp_languaget::force_file_type()
     !config.options.get_bool_option("no-library"))
   {
     log_debug("[CPP] Adding CPP includes: {}", ESBMC::esbmc_cpp_includes());
-    compiler_args.push_back("-isystem" + ESBMC::esbmc_cpp_includes());
+    compiler_args.push_back("-isystem=" + ESBMC::esbmc_cpp_includes());
     compiler_args.push_back("-I" + ESBMC::esbmc_cpp_includes() + "/CUDA");
-    compiler_args.push_back("-I" + ESBMC::esbmc_cpp_includes() + "/Qt/QtCore");
+    compiler_args.push_back("-I" + ESBMC::esbmc_cpp_includes() + "/Qt");
```
fbrausse: I don't have the Qt version modelled here, but `pkg-config --cflags Qt5Core` gives me both paths:
```
-DQT_CORE_LIB -I/usr/include/qt5/QtCore -I/usr/include/qt5
```

rafaelsamenezes: I added the qtcore back. I deleted some repeated OMs though (they didn't had ay specific behaviour)

src/clang-cpp-frontend/esbmc_internal_cpp.h
```diff
@@ -0,0 +1,5 @@
+#include <sstream>
+namespace ESBMC
+{
+const std::string &esbmc_cpp_includes();
```
fbrausse: I don't think we need the `esbmc_` prefix inside an `ESBMC` namespace. Also, there already seems to be a `esbmct` namespace, e.g. in irep/irep2.h and irep2/irep2_meta_templates.h. Should we re-use it?

fbrausse (approved):
LGTM, just one small nitpick.

kunjsong01:
As discussed off line, got to make sure whatever in the tmp folder really refers to our C++ OMs, e.g. the parsing/conversion error line numbers in `/tmp/esbmc-cpp-headers-bf22-9b68-f3c9/string` should match `src/cpp/string`

src/clang-cpp-frontend/esbmc_internal_cpp.h
rafaelsamenezes: Sure. Replaced ESBMC with esbmct and replaced the esbmc_cpp_includes with abstract_cpp_includes

PR #1135, opened 2023-06-28
Mock kernel

Kai-zhichengzhou:
I create directory in c2goto for mocking kernel dependencies, and create kmalloc related functions( including checking for the allocation size and if the flag passed in valid). The functionality has been tested locally with expected outcomes.


src/c2goto/library/stdlib.c
```diff
@@ -213,24 +216,75 @@ __ESBMC_HIDE:;
...
+    return malloc(size);
 }
 
+
+void kfree(const void *ptr) {
+    free((void *)ptr);
+}
+
+
+// void *__kmalloc(size_t size, gfp_t flags)
```
lucasccordeiro: Remove these?

src/c2goto/library/stdlib.c
```diff
@@ -252,10 +306,10 @@ void *kcalloc(size_t n, size_t size, gfp_t flags)
   return calloc(n, size);
 }
 
-void kfree(void *objp)
-{
-  free(objp);
-}
+// void kfree(void *objp)
```
lucasccordeiro: Remove these?

lucasccordeiro (changes requested):
Can you add some regression tests?

lucasccordeiro:
You can see some examples of regression tests here: https://github.com/esbmc/esbmc/tree/master/regression.

You can create a new folder to add regression tests about the kernel verification.

src/c2goto/library/kernel/include/linux/gfp.h
```diff
@@ -0,0 +1,67 @@
+
+#define MAX_ALLOC_SIZE 1024*1024 // Maximum allocatable size set to 1MB for example
```
rafaelsamenezes: Is this a hard requirement? 

If not, then this should be symbolic or obtained through a intrinsic. A TODO is enough for now though

rafaelsamenezes (changes requested):

fbrausse:
It would be great if you could provide a few more details about this development.

> I create directory in c2goto for mocking kernel dependencies, and create kmalloc related functions( including checking for the allocation size and if the flag passed in valid). The functionality has been tested locally with expected outcomes.

I assume "kernel" refers to the Linux kernel, and the headers are copied from it (or modified based on an existing copy).
1. It would be good to know what version of the kernel and what the paths of the corresponding source is. Additionally, any changes to those should be mentioned, if we ever want to be able to update these definitions to a newer kernel version.
2. If the source indeed is Linux, I would suggest to rename the "kernel" directory to `linux-VERSION` where `VERSION` corresponds to the official release these headers have been taken from.
3. I don't understand why some of these files are empty. Is it because they are included, e.g.
   ```
   [...]
   #include <linux/percpu-refcount.h>
   ```
   in `src/c2goto/library/kernel/include/linux/slab.h`?
4. Related to 3., the path `src/c2goto/library/kernel/include` does not seem to put as an include directory for any command during the build, so am I right in assuming that the headers contained therein are actually not used during compilation of ESBMC?

Kai-zhichengzhou:
> It would be great if you could provide a few more details about this development.
> 
> > I create directory in c2goto for mocking kernel dependencies, and create kmalloc related functions( including checking for the allocation size and if the flag passed in valid). The functionality has been tested locally with expected outcomes.
> 
> I assume "kernel" refers to the Linux kernel, and the headers are copied from it (or modified based on an existing copy).
> 
> 1. It would be good to know what version of the kernel and what the paths of the corresponding source is. Additionally, any changes to those should be mentioned, if we ever want to be able to update these definitions to a newer kernel version.
> 2. If the source indeed is Linux, I would suggest to rename the "kernel" directory to `linux-VERSION` where `VERSION` corresponds to the official release these headers have been taken from.
> 3. I don't understand why some of these files are empty. Is it because they are included, e.g.
>    ```
>    [...]
>    #include <linux/percpu-refcount.h>
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    in `src/c2goto/library/kernel/include/linux/slab.h`?
> 4. Related to 3., the path `src/c2goto/library/kernel/include` does not seem to put as an include directory for any command during the build, so am I right in assuming that the headers contained therein are actually not used during compilation of ESBMC?

Hi, so here are the responses corresponding to your questions. 1. Since the dependencies of linux kernel is huge and nested deeply, it is kind of impossible to generate AST using the native linux kernel dependencies as a lot of headers to be included is not necessary for current verification and some of them are related to assembly. My approach here is to cut down the dependencies by mock the kernel dependency and behavior based on my verification scope, e.g. kmalloc and the dependencies Kmalloc needs here. By this way, we can achieve the purpose of static verification without building the whole AST related to kernel. When we need to implement more scope of verification in the future, we can add more codes into the mock library.  
 2. The version of kernel  I was trying to mock is the kernel version of ubuntu 20.04, which is 5.4.1.  
 3. .Some of the files are empty because I just put there to mock the dependency of kernel in case more mock codes need to be put in those headers in the future. But I can remove them at this stage if we need to do that.  
 4. The gfp.h file in the kernel/include/linux is included in the stdlib.c file, so I think the files are used and compiled since the functionality works as expected ?  Please let me know if there are more issues with this request and I will try to fix that.

fbrausse:
Many thanks!

> 2. The version of kernel I was trying to mock is the kernel version of ubuntu 20.04, which is 5.4.1.

It's highly unlikely that Ubuntu 20.04 still uses version 5.4.1 given that the Linux 5.4 LTS release now is at 5.4.249. Could you please double-check?

Regarding item 4., sure `kernel/include/linux/gfp.h` is included via "..." from stdlib.c, but that file does not include others, so I was wondering why you add all the other files.

> My approach here is to cut down the dependencies by mock the kernel dependency and behavior based on my verification scope, e.g. kmalloc and the dependencies Kmalloc needs here. By this way, we can achieve the purpose of static verification without building the whole AST related to kernel.

I wonder whether that actually works the way you implemented it in this PR: The headers are under src/c2goto/library, but that directory is supposed to contain only things that are internal to our operational models and which are supposed to be compiled with c2goto. That means, if you build the c2goto models (cmake var `ESBMC_BUNDLE_LIBC=ON`) and run the resulting ESBMC binary, those headers will be gone.

We already have code for a similar approach, but in contrast that is living in `src/c2goto/headers`. Those headers there are always¹ extracted and the to-be-verified program is translated using those (our) headers. If that is what you're trying to achieve, I believe the headers need to be moved to `src/c2goto/headers` instead.

¹ modulo the option `--no-library`.

XLiZHI:
Hi, 

You may want to run the following cmd in the `ESBMC_Project/esbmc` directory to pass the code style:

```
find src -iname *.h -o -iname *.c -o -iname *.cpp -o -iname *.hpp -iname *.hh | grep -v src/ansi-c/cpp | grep -v src/clang-c-frontend/headers/ | xargs .././clang11/bin/clang-format -style=file -i -fallback-style=none
```

src/c2goto/library/stdlib.c
```diff
@@ -213,24 +215,56 @@ __ESBMC_HIDE:;
 
 typedef unsigned int gfp_t;
 
+void check_gfp_flags(gfp_t flags)
```
fbrausse: Please make this internal function `static`.

src/c2goto/library/stdlib.c
```diff
@@ -213,24 +215,56 @@ __ESBMC_HIDE:;
...
+  assert((flags & ~valid_flags) == 0);
+}
 void *__kmalloc(size_t size, gfp_t flags)
 {
-  (void)flags;
   return malloc(size);
 }
 
-void *kmalloc(size_t size, gfp_t flags)
+void *__kmalloc_large(size_t size, gfp_t flags)
```
fbrausse: Same here, this is internal, please make it `static`.

src/c2goto/library/stdlib.c
```diff
@@ -213,24 +215,56 @@ __ESBMC_HIDE:;
...
 }
 
-void *kzalloc(size_t size, gfp_t flags)
+void *kmalloc(int size, int flags)
 {
-  (void)flags;
+  // Check size greater than  zero and less than max
+  assert(size > 0 && size <= MAX_ALLOC_SIZE);
+  //check flags greater than zero
+  assert(flags > 0);
```
fbrausse, line 249: From its name, I would have expected the purpose of `check_gfp_flags()` to be encoding these kinds of constraints on the flags, though I could be wrong if that function is a verbatim copy from the kernel. Please clarify.

src/c2goto/library/stdlib.c
```diff
@@ -3,6 +3,8 @@
 #include <ctype.h>
 #include <limits.h>
 #include <errno.h>
+#include "../headers/kernel/include/linux/gfp.h"
```
fbrausse: This should be an #include with \<angle\> brackets, not "quotes" in order to not rely on the path layout of OMs and their headers inside `src/c2goto`. The header directory is part of the include paths.

regression/kernel/kmalloc_invalid_allocation/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+test.c
+-I /home/kai/PGT/ESBMC_Project/esbmc/src/c2goto/library/
```
lucasccordeiro: We should avoid this. 

@rafaelsamenezes : can you suggest an alternative here?

src/c2goto/headers/kernel/include/linux/gfp.h
```diff
@@ -0,0 +1,68 @@
+// TODO: Replace hardcoded MAX_ALLOC_SIZE with a dynamically obtained value
```
lucasccordeiro: Shall we add the license of this code?

lucasccordeiro (changes requested):

fbrausse (changes requested):
There are still a few rough edges here to figure out:
- The directory is still called `kernel`, please rename it to something like `linux-5.4.x` or a combination of "ubuntu20.04" and the corresponding output from `uname -r` if you used the current kernel source package of this distribution as a template for the headers.
- In order for your new regression tests to be run, please include the directory name (see the previous item for naming "kernel") into the file `regression/CMakeLists.txt`. There are a few cases in there, the important ones here are `if(APPLE)`, `elseif(WINDOWS)` and `if(NOT BENCHBRINGUP)`. Your directory should be added to each applicable branch in the `REGRESSIONS` list.
- All added regression tests have an `-I` option with some path in their test.desc file. This needs to be removed since noone can be expected to have this path.
- All added regression tests include `"stdlib.c"`. This file is part of the **internal** implementation of the operational models and should not be accessed in generic programs that ESBMC verifies. Please include the corresponding header instead, that declares the `kmalloc()` prototype and its flags.

Again, the reasons I would like to be more specific about "kernel" is that the Linux kernel is not the only kernel in use, and ESBMC is used to verify more than just code from that particular (version of that) kernel. Additionally, in order to be able to in the future update this operational model you're providing, the version info would be quite helpful.

As a comment (not necessarily to be addressed in this PR), how about moving the implementation of the `kmalloc` family of functions out of `stdlib.c`? I would consider that file to be only about implementing the OMs for the C standard library.

regression/kernel/kmalloc_invalid_allocation/test.desc
fbrausse: Installing/Bundling the new headers in the c2goto directory should be enough to be able to clear this line.

regression/CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ else()
                     ${REGRESSIONS_JIMPLE}
                     extensions
                     cuda/COM_sanity_checks
+                    kernel
```
fbrausse: Should we rename this directory to `linux`?

src/c2goto/c2goto.cpp
```diff
@@ -61,7 +61,7 @@ class c2goto_parseopt : public parseoptions_baset, public language_uit
     if(config.set(cmdline))
       return 1;
     config.options.cmdline(cmdline);
-    messaget::state.verbosity = VerbosityLevel::Result;
+    messaget::state.verbosity = VerbosityLevel::Debug;
```
fbrausse: Could you remove this change? On a side note: I've also been using `Debug` here quite a lot, we should at some point in the future add a `--verbosity` flag for `c2goto`...

src/c2goto/library/stdlib.c
```diff
@@ -213,24 +216,56 @@ __ESBMC_HIDE:;
...
+    __GFP_REPEAT | __GFP_NOFAIL | __GFP_NORETRY | __GFP_MEMALLOC | __GFP_COMP |
+    __GFP_NO_KSWAPD | __GFP_OTHER_NODE | __GFP_WRITE | __GFP_HARDWALL |
+    __GFP_THISNODE | __GFP_ATOMIC | __GFP_ACCOUNT | __GFP_RECLAIM | __GFP_IO |
+    __GFP_FS | GFP_KERNEL | GFP_KERNEL_ACCOUNT | GFP_NOIO | GFP_NOFS |
+    GFP_USER | GFP_DMA | GFP_DMA32 | GFP_HIGHUSER;
+
+  // Check if any flag is set that is not in the list of valid flags
+  assert((flags & ~valid_flags) == 0);
+}
+static void *__kmalloc(size_t size, gfp_t flags)
```
fbrausse, line 233: Is `__kmalloc` actually used anywhere? If not, we could just remove it.

src/c2goto/library/stdlib.c
```diff
@@ -3,6 +3,9 @@
 #include <ctype.h>
 #include <limits.h>
 #include <errno.h>
+// #include <ubuntu20.04_5.15.0-73-generic/include/linux/slab.h>
```
fbrausse: Could you remove this line?

fbrausse (commented):
The path of both new .h files is missing the last component: `include/linux/gfp.h`. Besides that I would like to request just a few minor changes.

regression/kernel/kmalloc_invalid_allocation/test.desc
```diff
@@ -0,0 +1,3 @@
+CORE
+test.c
+^VERIFICATION FAILED$
```
fbrausse: Sorry, one more thing to change: Our testing script (regression/testing_tool.py) expects the first 3 lines in `test.desc` to be
1. tag (CORE, THOROUGH, KNOWNBUG, etc.)
2. file name of the main source
3. other source files and command-line options for ESBMC

Only starting at the 4. line are the regular expressions placed. Just add an empty 3rd line in the `test.desc` files, please.

Kai-zhichengzhou: > Sorry, one more thing to change: Our testing script (regression/testing_tool.py) expects the first 3 lines in `test.desc` to be
> 
> 1. tag (CORE, THOROUGH, KNOWNBUG, etc.)
> 2. file name of the main source
> 3. other source files and command-line options for ESBMC
> 
> Only starting at the 4. line are the regular expressions placed. Just add an empty 3rd line in the `test.desc` files, please.

hi，I have fixed the codes you mentioned. And for the path of gfp.h, the /linux path is not missing because the way I define the direcotry is ubuntuxxx/kernelxxx/include, since there would only be one folder in include directory, I use the /include for the directory of my headers indtead. Please let me know if there are further issues

fbrausse: I thought you wanted to model the directory structure of the Linux kernel. For instance, my gfp.h is here:
`/usr/src/linux/include/linux/gfp.h`.

Kai-zhichengzhou: > I thought you wanted to model the directory structure of the Linux kernel. For instance, my gfp.h is here: `/usr/src/linux/include/linux/gfp.h`.

hi, i have changed that to /include/linux. could you check :) ?

src/c2goto/headers/CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
-set(inputs limits.h pthread.h stdarg.h stdbool.h stddef.h
-           # Glibc
+set(inputs limits.h pthread.h stdarg.h stdbool.h stddef.h 
            bits/pthreadtypes.h bits/thread-shared-types.h
            # MacOSX10.x
            sys/_types/_pthread_attr_t.h
-           # MacOSX11.0 and above
+           # MacOSX11.0 and abovegit
```
fbrausse: Let's remove this change.

fbrausse (commented):
Just a minor nitpick.

fbrausse (approved):
Looks good to me.

lucasccordeiro (approved):

lucasccordeiro:
@Kai-zhichengzhou: could you please rebase this PR with the master branch?

lucasccordeiro:
Thanks for submitting this PR, @Kai-zhichengzhou.

PR #1136, opened 2023-06-29
 Automatic sysroot inclusion in macOS

rafaelsamenezes:
Small QoL fixes for macOS


fbrausse:
In #905 I'm doing something similar, but with a cmake variable (additionally touches Options.cmake and cmake_config.in). We actually already have the `C2GOTO_SYSROOT` which is used in the generated `macos-wrapper.sh`. Would you consider using (and defaulting) that variable instead of the hard-coded path? That way we could even remove generating the `macos-wrapper.sh` (if it isn't required for backwards compat).

rafaelsamenezes:
> Would you consider using (and defaulting) that variable instead of the hard-coded path?

Sure, good idea

lucasccordeiro (approved):

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -165,7 +165,14 @@ void clang_c_languaget::build_compiler_args(const std::string &tmp_dir)
 
   if(config.ansi_c.target.is_macos())
   {
+    /* NOTE: clang cannot deduce the sysroot automatically
+     *  we could invoke system's clang and try to find out
+     * the correct place... but that seems an overkill
+     */
+    if(sysroot.empty())
+      compiler_args.emplace_back("--sysroot=" ESBMC_C2GOTO_SYSROOT);
```
fbrausse: I believe this is now independent of `config.ansi_c.target.is_macos()`. Locally, my clang accepts an empty `--sysroot=` parameter without modifying the search paths.

fbrausse (approved):
LGTM! Thanks for addressing my comments, @rafaelsamenezes.

lucasccordeiro:

Issue #1140, opened 2023-06-30
C++ OM enablement coverage


kunjsong01:
We've cleared many corner cases (and possibly some edge cases) in the umbrella ticket for template support. 
Need to double check the coverage of OM enablement in the table https://github.com/esbmc/esbmc/wiki/OM-Workload-Estimate-and-Tracking. 

TODO: 
1. add sanity checks for all OMs
2. run a regression and get the number of enabeld OMs. 
3. Update the table in https://github.com/esbmc/esbmc/wiki/OM-Workload-Estimate-and-Tracking.

kunjsong01:
closed as all TODOs are completed, and the wiki page has been updated.

PR #1141, opened 2023-06-30
Added OM sanity check TCs

kunjsong01:
Done work for https://github.com/esbmc/esbmc/issues/1140

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/algorithm/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/algorithm/test.desc
@@ -1,0 +1,5 @@
+KNOWNBUG
+main.cpp
+-I /__w/esbmc/esbmc/src/cpp/library --memlimit 14000000 --timeout 900
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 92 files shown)


regression/esbmc-cpp/OM_sanity_checks/algorithm/test.desc
```diff
@@ -0,0 +1,5 @@
+KNOWNBUG
+main.cpp
+-I /__w/esbmc/esbmc/src/cpp/library --memlimit 14000000 --timeout 900
```
lucasccordeiro: Do we really need this?

lucasccordeiro (changes requested):

regression/esbmc-cpp/OM_sanity_checks/algorithm/test.desc
kunjsong01: removed

lucasccordeiro (approved):
Many thanks, @kunjsong01.

lucasccordeiro (approved):

PR #1142, opened 2023-06-30
Fixing arch for 32-bit arm analysis as armv6

rafaelsamenezes:
After latest M1 update https://github.com/esbmc/esbmc/pull/810 is happening again. This fixes it.


fbrausse (approved):
LGTM.

fbrausse:
Background, since @rafaelsamenezes and I have been debugging this quite a bit: In Clang-16 there was a change where `-target aarch64-unknown-macos -m32 -mfloat-abi=softfp` does no longer work while in Clang-11..15 did accept this. Appearently, Clang did earlier fall back on the internal `-target armv4t-unknown-macosx10.4.0` and still accept the softfp float ABI, while starting with Clang-16 it no longer does. So this approach sets the internal `-target armv6k-unknown-macosx10.4.0` instead where Clang does accept softfp.

lucasccordeiro (approved):

PR #1144, opened 2023-07-03
Automatically build the developer manual

rafaelsamenezes:
This PR auto-builds the docs/manual.tex file. Maybe we should start updating/improving it as some of its informations is too old or incomplete.


lucasccordeiro (approved):

fbrausse (approved):
> Maybe we should start updating/improving it as some of its informations is too old or incomplete.

I might need to read it at some point. ;) LGTM!

PR #1161, opened 2023-07-07
String OM fix for 1145

### LINKED ISSUE #1145: Qt OMs contains duplicate of C++ standard abstractions
Both the QtCore and QtGui folders contain an implementation of `string`. The implementation is different from the main abstract model (at cpp/library).

Is there any reason for those headers (a Qt only specific optimization) or we can just remove it?

kunjsong01:
Fixed https://github.com/esbmc/esbmc/issues/1145 to avoid interference with `cpp/library/string`

Just renamed it for future reference, just in case we want to run the qt test suite again. 
I'm also OK with removing them if required.


rafaelsamenezes:
Why don't just remove it? 

If you find that it broke something you can revert the PR.

kunjsong01:
> Why don't just remove it?
> 
> If you find that it broke something you can revert the PR.

removed

regression/esbmc-cpp/OM_sanity_checks/ciso646/main.cpp
```diff
@@ -1,4 +1,4 @@
-#include <cisco646>
+#include <ciso646>
```
rafaelsamenezes: Why this file is part of the PR?

rafaelsamenezes (changes requested):

regression/esbmc-cpp/OM_sanity_checks/ciso646/main.cpp
kunjsong01: removed

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @kunjsong01.

PR #1166, opened 2023-07-09
ESBMC is now able to build in Release mode with WERROR

rafaelsamenezes:


rafaelsamenezes:
Close for now. This will require multiple small PRs

PR #1177, opened 2023-07-10
Add some options to secondary tools

fbrausse:
This PR
- adds support for the `--verbosity` flag to the c2goto command, and
- adds a `--dry-run` to the esbmc-wrapper.py, which just prints the command-line it would execute without actually executing it.

I've used both locally for quite a while and think these options are generally useful for debugging purposes.


lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #1183, opened 2023-07-11
Fixed `Run a Benchmark` workflow

kunjsong01:
Fixed this workflow to run a user specified benchmark and get the pass rate stats, as well as error/warning signatures if there is any. 

Interesting findings: 
- it appears that build.sh doesn't catch the environment variable set in github container. Got to do it in the way like @rafaelsamenezes' werror patch.

### DIFF
--- a/.github/workflows/benchbringup.yml
+++ b/.github/workflows/benchbringup.yml
@@ -27,27 +27,17 @@ jobs:
     name: Build ESBMC (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: 720 # 12h
-    container: rafaelsamenezes/esbmc-ci:clang11
+    #container: rafaelsamenezes/esbmc-ci:clang11
     steps:
-      - name: Small test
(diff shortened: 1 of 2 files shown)


lucasccordeiro (approved):

kunjsong01:
https://github.com/esbmc/esbmc/actions/runs/5518916208/jobs/10063487056?pr=1183#step:9:2435 

@rafaelsamenezes any idea why this TC is failing? Is it a false alarm? as this patch is not even touching any source code....

fbrausse:
> https://github.com/esbmc/esbmc/actions/runs/5518916208/jobs/10063487056?pr=1183#step:9:2435
> 
> @rafaelsamenezes any idea why this TC is failing? Is it a false alarm? as this patch is not even touching any source code....

I'm also getting this error in #1184. It's about a missing declaration of `uint8_t`. Maybe this test case should explicitly `#include <stdint.h>`...

lucasccordeiro:
Thanks for submitting this PR, @kunjsong01.

Issue #1064, opened 2023-06-08
[frontend] ESBMC does not handle argc and argv correctly


lucasccordeiro:
ESBMC should not provide `VERIFICATION FAILED` for this example:

````C
#include <stdio.h>
#include <stdlib.h>
int main(int argc, char *argv[]){
        if (argc<2){
        fprintf(stderr, "Program name is: %s\n", argv[0]);
}
}
````

@mikhailramalho started fixing this issue in https://github.com/esbmc/esbmc/tree/argc_argv, but the implementation is not yet complete.

**Comment by lucasccordeiro** [comment]
@shmarovfedor: Can I ask you please to help @brcfarias with this issue?

**Comment by lucasccordeiro** [comment]
@shmarovfedor: Can I ask you whether you had a chance to check this issue?

**Comment by brcfarias** [comment]
For the program below:
```C
#include <stdio.h>

int main(int argc, char* argv[]) {
  if (argv[0]) {}
  return 0;
}
```
the failure is occuring at  https://github.com/esbmc/esbmc/blob/1eefb1f742f977bc10a53b91d694a5e7d1990e46/src/pointer-analysis/dereference.cpp#L2205

In this function we have the following expressions

offset:
```
mul
* side_1 : constant_int
    * value : 0
    * type : unsignedbv
      * width : 64
* side_2 : constant_int
    * value : 8
    * type : unsignedbv
      * width : 64
* type : unsignedbv
  * width : 64
```
unsigned_offset:
```
typecast
* from : div
    * side_1 : mul
        * side_1 : constant_int
            * value : 0
            * type : unsignedbv
              * width : 64
        * side_2 : constant_int
            * value : 8
            * type : unsignedbv
              * width : 64
        * type : unsignedbv
          * width : 64
    * side_2 : constant_int
        * value : 8
        * type : unsignedbv
          * width : 64
    * type : unsignedbv
      * width : 64
* rounding_mode : symbol
    * name : c:@__ESBMC_rounding_mode
    * renamelev : Level 0
    * level1_num : 0
    * level2_num : 0
    * thread_num : 0
    * node_num : 0
    * type : signedbv
      * width : 32
* type : unsignedbv
  * width : 32
```
access_size_e:
```
constant_int
* value : 8
* type : unsignedbv
  * width : 32
```
  
upper_byte:
```
add
* side_1 : typecast
    * from : div
        * side_1 : mul
            * side_1 : constant_int
                * value : 0
                * type : unsignedbv
[...]

Issue #1118, opened 2023-06-23
Cross-backend regression testing


fbrausse:
It would be good to not just run ESBMC with its default solver (Z3 on a Windows CI build, Boolector for other CI builds, depending on configure-time options otherwise) over the solver-agnostic regression suite, but to also test the other (built-in) backends, smtlib in particular, see https://github.com/esbmc/esbmc/pull/1117#issuecomment-1604053990 and the following comments.

This is technically entirely possible by passing `--default-solver=[...]` and potentially additional parameters by setting the `ESBMC_OPTS` environment variable, see e.g. #1117 again for `smtlib`. Some questions to address are:
- Should these be new targets for cmake/ctest?
- Should these be a hard-coded set of some/all solver-agnostic benchmarks for each backend?
- How are the arguments given to the program for `--smtlib-solver-prog` handled? They do influence the timeouts and they could also influence the results.
- Should these tests be run as part of the CI and if so, which combinations on which target platform?

rafaelsamenezes:
Hm, maybe its time to let the `testing_tool.py` die. It doesn't serve its original intent anymore and its just used in hacky way for invoking esbmc. We could:

- Revamp it assuming that CMake will just give a folder and some options
- Remove it and implement its functionality inside CMake.
- Use something else.

Regardless of the option... we could improve our usage of ctest (and cdash?) beyond the basics.

> Should these be new targets for cmake/ctest?

I would add it in a different configuration (see [add_test](https://cmake.org/cmake/help/latest/command/add_test.html)). This way we can have more control on which tests are being executed.

> Should these be a hard-coded set of some/all solver-agnostic benchmarks for each backend?

I don't think so. At most, we could expand the test.desc to also contain solvers that are incompatible with the test case. This would only be useful in situations that we rely on something like incremental-solving or some specific option.

For tests that rely on a specific solver... they should be moved to a specific folder.

> How are the arguments given to the program for --smtlib-solver-prog handled? They do influence the timeouts and they could also influence the results.

I am not sure. Depends on how python subprocess behaves.

>  Should these tests be run as part of the CI and if so, which combinations on which target platform?

As you mentioned in the other issue... we could try to do independent runs for Linux with each solver weekly/monthly.

PR #1184, opened 2023-07-11
Fix more type sizes

fbrausse:
This PR addresses problems related to the bit-width of certain types, that are used during dereference (offset) and for the intrinsics `__ESBMC_alloc_size`/`__ESMBC_get_object_size()` and `__ESBMC_POINTER_OBJECT`.

Basically, we have to stop using types of constant 64 or 32 bits arbitrarily for certain concepts. The size of an object is determined by config.ansi_c and, by extension, by size_type{,2}(). That also holds true for indices, the type of pointer-obj expressions (determined by the address space), and so on. That's what this PR (after the first CHERI-C one) continues. In particular, since @shmarovfedor changed the offsets in dereference to be measured in bits, we need to use a type that does not overflow when converting bytes to bits. This type is `bitsize_type2()`.

I've added a regression test case (in 32 and 64 bit versions), that came out of a debugging session of #1064 with @brcfarias. The 32-bit one in particular would not be `VERIFICATION SUCCESSFUL` without the changes to the type of offsets during calculation in dereference.


fbrausse:
A note of caution: This PR changes the offset-bits calculations in dereference to use bit-widths of 67 (64 + 3) or 35 (32 + 3) bits, depending on the config. That might have performance impacts, I'll run it over SV-COMP to be sure.

src/irep2/irep2_expr.h
```diff
@@ -2898,6 +2898,13 @@ class index2t : public index_expr_methods
     assert(
       is_array_type(source) || is_string_type(source) ||
       is_vector_type(source));
+    if(0 && is_array_type(source))
```
mikhailramalho: remove?

mikhailramalho (commented):
Yeah, I agree. These issues have been a pain for a long time in esbmc.

This has historically been a problem since we moved from 32bit to 64bit by default, maybe it's time to run our regressions twice? With and without `--32` passed to esbmc?

src/irep2/irep2_expr.h
fbrausse: done.

fbrausse:
> This has historically been a problem since we moved from 32bit to 64bit by default, maybe it's time to run our regressions twice? With and without `--32` passed to esbmc?

That would be great to have, I agree. But we have a few problems with making the regression suite that modular:
1. it already takes too long for a quick turnaround time (1.5h to wait until the CI says "error" so one can fix it is not great)
2. there are more dimensions to generalize the regression to: #1118 wants to run it with different external solvers via `--smtlib-solver-prog`, we might want to use different backends which the `--default-solver` is for, and, finally
3. our option handling is not completely compatible with overriding options such as `--32` with `--64`; we currently bail out if multiple of these are defined, which is exactly why the `--default-solver` exists in comparison to giving the actual solver with `--z3` and the like, see also the discussion in #784 about that.

Don't really know how to solve 1. and 2. Item 3. could be solved by introducing another dimension into the option processing. We already have the envvar `ESBMC_OPTS` and the cmdline. Adding another variable `ESMBC_OPTS_OVERRIDE` could be used to override any given options from the first two. The thing is, we don't support overriding options, yet: `--32 --64` lead to an error, those options are not being related to each other in any generic way right now.

src/pointer-analysis/dereference.cpp
```diff
@@ -714,7 +718,12 @@ expr2tc dereferencet::build_reference_to(
   // or index exprs, like foo->bar[3]. If bar is of integer type, we translate
   // that to be a dereference of foo + extra_offset, resulting in an integer.
   if(!is_nil_expr(lexical_offset))
+  {
+    /* lexical_offset is in bytes, not bits */
+    assert(lexical_offset->type != offset_type);
+    lexical_offset = typecast2tc(offset_type, lexical_offset);
     final_offset = add2tc(final_offset->type, final_offset, lexical_offset);
```
fbrausse: Sorry for the self-review but this is broken: can't add bytes to bits.

fbrausse: Fixed.

fbrausse:
stats-30s.txt:
```
  correct:           14414
    correct true:     8195
    correct false:    6219
  incorrect:            62
    incorrect true:     28
    incorrect false:    34
  unknown:           11983
  Score:             21169 (max: 43043)
```

[This PR](https://github.com/esbmc/esbmc/actions/runs/5531134576/jobs/10092005809) (4c4cbc549e8fbc7c927e4e37ef4933d11ad59274):
```
2023-07-12T23:55:46.4144530Z Statistics:          26459 Files
2023-07-12T23:55:46.4145419Z   correct:           14342
2023-07-12T23:55:46.4151946Z     correct true:     8147
2023-07-12T23:55:46.4152583Z     correct false:    6195
2023-07-12T23:55:46.4153187Z   incorrect:            64
2023-07-12T23:55:46.4153735Z     incorrect true:     29
2023-07-12T23:55:46.4154257Z     incorrect false:    35
2023-07-12T23:55:46.4169231Z   unknown:           12053
2023-07-12T23:55:46.4172378Z   Score:             21001 (max: 43043)
```
~I can't analyze the results, because I don't have access and they're not uploaded. @rafaelsamenezes could you please take a look? The `run-output.zip` is only 120KiB and contains just the results for `esbmc.2023-07-05_10-40-57.results.SV-COMP23_valid-memsafety.SoftwareSystems-DeviceDriversLinux64-MemSafety.xml.bz2`.~ *Edit:* Thanks for the quick help, @rafaelsamenezes.

fbrausse:
- 1x unknown -> incorrect true; this was throwing a dyn_sized_array_exp before
- 2x timeout -> incorrect false; these are incorrect already on master, just faster now
- a number of new unknown, which segfault due to NULL exrpt2tc elements in the member of a constant_array2t
- a few new unknowns throwing dyn_sized_array_exp

I'll take a look.

fbrausse:
The NULL exprs that cause segfaults are created as part of `intrinsic_memset()` here: https://github.com/esbmc/esbmc/blob/06a5435b8b33f7bb5affb4b429d1a0604533c324/src/goto-symex/builtin_functions.cpp#L1015-L1020 because some struct type has a member named `bit_field_pad$6`.

fbrausse:
Some of the dyn_sized_array_excp are thrown also in `intrinsic_memset()` here https://github.com/esbmc/esbmc/blob/06a5435b8b33f7bb5affb4b429d1a0604533c324/src/goto-symex/builtin_functions.cpp#L1309

fbrausse:
@rafaelsamenezes Could you please take a look whether intrinsic_memset() is being used inappropriately here for whatever assumptions it has? The NULL exprs you can reproduce with
```
esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps sv-benchmarks/c/ldv-linux-3.4-simple/32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--net--phy--spi_ks8995.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.i --64 --no-pointer-check --no-bounds-check --interval-analysis --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
```
and the dyn-sized exceptions with
```
esbmc  --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps sv-benchmarks/c/ldv-linux-3.14/linux-3.14_linux-kernel-locking-spinlock_drivers-net-ethernet-cisco-enic-enic.cil.i --64 --no-pointer-check --no-bounds-check --interval-analysis --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
```

fbrausse:
I've added a straight-forward fix for the errors assuming the intrinsic_memset() was called correctly. It fixes some cases, I'll check whether it fixes all cases via another sv-comp run.

lucasccordeiro:
Thanks for this update, @fbrausse.

fbrausse:
Not entirely unexpectedly this slows us down a bit.

[Master](https://github.com/esbmc/esbmc/actions/runs/5620760754) (4640d2ec1d68849dc295d39c9d6c269f2ada1344):
```
  correct:           14419
    correct true:     8201
    correct false:    6218
  incorrect:            63
    incorrect true:     29
    incorrect false:    34
  unknown:           11977
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5620745315) (2eb51ed96800dccb066d8d5649faf583d09b5e63):
```
  correct:           14347
    correct true:     8158
    correct false:    6189
  incorrect:            64
    incorrect true:     29
    incorrect false:    35
  unknown:           12048
  Score:             21017 (max: 43043)
```
The new incorrect one was timeout before but already fails on master with a larger timeout:
- c/goblint-coreutils/instrumented_cut_comb.i this is #584

There are also fixed unknown ones, and a bunch of new unknown ones in memsafety in particular.

lucasccordeiro:
Thanks, @fbrausse.

Can I ask you please to fix the conflicting file?

lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #1186, opened 2023-07-11
Experimental simplification for memsafety assertions

rafaelsamenezes:
This PR is to try out some simplifications over:

- same-object.
- deref over `alloc_size` array
- deref over `alloc` array
- Add an option to stop propagation of invalid-object

### DIFF
--- a/src/goto-symex/goto_symex.h
+++ b/src/goto-symex/goto_symex.h
@@ -968,6 +968,8 @@ protected:
 
   void rename(expr2tc &expr) override;
 
+  expr2tc constant_propagation(expr2tc &expr);
+
   void
   dump_internal_state(const std::list<struct internal_item> &data) override;
   bool is_live_variable(const expr2tc &sym) override;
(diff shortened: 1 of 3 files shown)


lucasccordeiro:
@rafaelsamenezes: this is an excellent initiative. I believe there are many low-hanging fruits to simplify our VCCs for programs with pointers and dynamic memory allocation.

pmatos:
This experimental draft's goals have largely been addressed on `master` through the proper channels since it was opened:

- `same_object2t::do_simplify()` handles same-object folding at the simplifier layer;
- the `pointer_object` / pointer-offset simplifiers and the `__builtin_object_size` intrinsic cover the alloc-size / dereference reasoning;
- interval-based guard/assertion pruning (`--interval-symex-guard`, on by default, and `--interval-symex-assert`) provides a principled memory-safety-assertion pruning mechanism.

The touched files have also moved on considerably (`simplify_expr2.cpp` → `expr_simplifier.cpp`; `dereference_failure` now takes `guard2tc&`), and the ad-hoc `a*b / a = b` rule here lacks the `a == 0` / overflow guards that the current simplifier deliberately enforces to preserve div-by-zero VCCs.

Closing this experimental draft as superseded. Happy to open a focused issue if there's a specific simplification here that isn't yet covered on `master`.

PR #1192, opened 2023-07-13
simulation for Kernel user space copy

Kai-zhichengzhou:
The pr implements the simulation for kernel functions copy_to_user and copy_from_user functions. Since these two functions also involved assembly part and complex interation with memory, the simulation can only do best-effort to simulate the behavior of the copying between kernel space and user space, and also achieved the verfication for valid pointers that are involved. The regression tests are created and tested.

### DIFF
--- a/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
+++ b/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
@@ -1,17 +1,10 @@
-#include "asm.h"
-#include "extable.h"
-#include "page.h"
-#include "smap.h"
-#include "../linux/compiler.h"
-#include "../linux/string.h"
-#include "../linux/kasan-checks.h"
 
 #define PAGE_SIZE 4096 // page size
-#define USER_MEMORY_SIZE 0x100000000 //only for simulate the user memory
-#define KERNEL_MEMORY_SIZE 0x100000000 // only for simulate the kernel memory
+#define USER_MEMORY_SPACE 10000 //only for simulate the user memory
+#define KERNEL_MEMORY_SPACE 10000// only for simulate the kernel memory
 
-char user_memory[USER_MEMORY_SIZE]; //mock user memory
-char kernel_memory[KERNEL_MEMORY_SIZE];//mock user memory
+char user_memory[USER_MEMORY_SPACE]; //mock user memory
+char kernel_memory[KERNEL_MEMORY_SPACE];//mock user memory
 
 //simulate copy_to_user function in kernel space 
 unsigned long copy_to_user(void* to, void* from, unsigned size);
--- a/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
+++ b/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
@@ -1,0 +1,21 @@
+#include "asm.h"
+#include "extable.h"
+#include "page.h"
+#include "smap.h"
+#include "../linux/compiler.h"
+#include "../linux/string.h"
+#include "../linux/kasan-checks.h"
+
+#define PAGE_SIZE 4096 // page size
+#define USER_MEMORY_SIZE 0x100000000 //only for simulate the user memory
+#define KERNEL_MEMORY_SIZE 0x100000000 // only for simulate the kernel memory
+
+char user_memory[USER_MEMORY_SIZE]; //mock user memory
+char kernel_memory[KERNEL_MEMORY_SIZE];//mock user memory
+
+//simulate copy_to_user function in kernel space 
+unsigned long copy_to_user(void* to, void* from, unsigned size);
+unsigned long copy_from_user(void* to, void* from, unsigned size);
+
+
+    
--- a/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
+++ b/src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
@@ -2,13 +2,11 @@
 #define PAGE_SIZE 4096 // page size
 #define USER_MEMORY_SPACE 10000 //only for simulate the user memory
 #define KERNEL_MEMORY_SPACE 10000// only for simulate the kernel memory
-
-char user_memory[USER_MEMORY_SPACE]; //mock user memory
-char kernel_memory[KERNEL_MEMORY_SPACE];//mock user memory
-
+extern char user_memory[USER_MEMORY_SPACE]; //mock user memory
+extern char kernel_memory[KERNEL_MEMORY_SPACE];//mock user memory
 //simulate copy_to_user function in kernel space 
-unsigned long copy_to_user(void* to, void* from, unsigned size);
-unsigned long copy_from_user(void* to, void* from, unsigned size);
+unsigned long copy_to_user(void* to, void* from, unsigned long size);
+unsigned long copy_from_user(void* to, void* from, unsigned long size);
 
 
     
--- a/src/c2goto/library/kernel.c
+++ b/src/c2goto/library/kernel.c
@@ -8,7 +8,6 @@
 
 #include <assert.h>
 
-
 typedef unsigned int gfp_t;
 
 static void check_gfp_flags(gfp_t flags)
@@ -79,8 +78,8 @@ unsigned long copy_to_user(void *to, void *from, unsigned long size)
   assert(from != NULL);
   assert(size <= PAGE_SIZE);
 
-  assert( (char*) to >= user_memory);
-  assert( (char*) from >= kernel_memory);
+  assert((char *)to >= user_memory);
+  assert((char *)from >= kernel_memory);
 
   //copy memory from kernel space to user space
   //simulate the copy operation by memcpy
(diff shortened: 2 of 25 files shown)


src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
```diff
@@ -0,0 +1,14 @@
...
+#define PAGE_SIZE 4096 // page size
+#define USER_MEMORY_SPACE 10000 //only for simulate the user memory
+#define KERNEL_MEMORY_SPACE 10000// only for simulate the kernel memory
+
+char user_memory[USER_MEMORY_SPACE]; //mock user memory
+char kernel_memory[KERNEL_MEMORY_SPACE];//mock user memory
+
+//simulate copy_to_user function in kernel space 
+unsigned long copy_to_user(void* to, void* from, unsigned size);
+unsigned long copy_from_user(void* to, void* from, unsigned size);
```
fbrausse: According to my (Linux) sources, both of these functions take their `size` parameter as `unsigned long`.

src/c2goto/library/stdlib.c
```diff
@@ -265,6 +266,51 @@ void kfree(const void *ptr)
...
+  assert(size <= PAGE_SIZE);
+
+  //check on kernel space memory boundry
+  assert(
+    (char *)to < kernel_memory + KERNEL_MEMORY_SPACE &&
+    (char *)to >= kernel_memory);
+  //check on user space memory boundry
+  assert(
+    (char *)from < user_memory + USER_MEMORY_SPACE &&
+    (char *)from >= user_memory);
```
fbrausse: Why are these asserts needed? ESBMC should be able to find array-out-of-bounds without them. Same for `copy_to_user()`.

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
```diff
@@ -0,0 +1,14 @@
+
+#define PAGE_SIZE 4096 // page size
+#define USER_MEMORY_SPACE 10000 //only for simulate the user memory
+#define KERNEL_MEMORY_SPACE 10000// only for simulate the kernel memory
+
+char user_memory[USER_MEMORY_SPACE]; //mock user memory
+char kernel_memory[KERNEL_MEMORY_SPACE];//mock user memory
```
fbrausse: The definitions of these two arrays should not be in the header.

fbrausse (changes requested):
I have a few remarks/questions below, and three here:
- What purpose do the (user|kernel)_memory arrays have for a program to be verified with ESBMC (e.g., the test cases you add)? There also are multiple definitions of them, which our linker should complain about when running the new regression tests. Does it?
- Unless I'm mistaken, you add a lot of unreferenced empty files in this PR. Why are they necessary and why are they not being bundled in the CMakeLists.txt?
- Since you're clearly adding more kernel-related functionality, would it be possible to move the definition of the kernel-related functions from stdlib.c to a new .c file that's unrelated to the libc?

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
Kai-zhichengzhou: hi, for the questions you mentioned, here are the thoughts for my approaches. The first question you mentioned is also the indeterminate points that i have for this simulation.So what i was trying to do using these two arrays is trying to mock the user and kernel memory space which is available for copying the data. The limitation here I encountered is that determining the user and kernel memory , or space involves complex mechanism in linux memory management, and different among different architecture. It can not be determined using static approach. So that's why i am trying to using a predefined array to mock the memory space available. I understand it could be seems unreasonable, so may i have some suggestions from you, or i should just remove them ?

src/c2goto/library/stdlib.c
Kai-zhichengzhou: these assertions are used for verifying if the simulation would cause the copying out of the available/legal space. Yes so just as you said, I could let ESBMC directly find the array-out-of-bounds problem ?

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
Kai-zhichengzhou: Also, the 2nd and 3rd question, i put those header file because in linux source codes, those are the dependency headers for uacess.h, although my headers did not use function or variables from those files. I did not put them in CmakeLIist since i feel bit weird to put those empty files in the bundle. So maybe I should remove them, or put them in Cmakelist? Yes for the 3rd question,so maybe I can create a new file in c2goto library directory and put all kernel functions in ?

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
Kai-zhichengzhou: Hi, if I remove these two arrays , then I think i also need to change some assertion in my function body. May i got more suggestions?  So my goal is to simulate the situations that the copy from user to kernel, or vice versa, may cause corruption illegal memory space

Kai-zhichengzhou: And also, I am really thankful for your suggestions:)

fbrausse: I meant that the header should just declare the arrays, not define them. In C each *definition* (like the arrays above) results in a new symbol being created, which you probably don't want: the regression tests should refer to *the exact same* arrays the implementation of copy_to/from_user do. To achieve that, C has a distinction between declaration and definition. Symbols should be defined just once (in a translation unit, aka a .c file), while they can be declared multiple times. C headers usually don't contain definitions of such symbols, they just declare that they exist and that they are defined somewhere else. A declaration of an array of 42 chars looks like this:
```c
extern char my_array[42];
```

src/c2goto/library/stdlib.c
fbrausse: > Yes so just as you said, I could let ESBMC directly find the array-out-of-bounds problem ?

Sounds good. Will you get the same results?

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
fbrausse: > I did not put them in CmakeLIist since i feel bit weird to put those empty files in the bundle. So maybe I should remove them, or put them in Cmakelist?

There is nothing wrong with bundling empty files, if that's how they are to be used. For headers it might make sense to do that in order to hide system headers with the same include path. But having empty files in ESBMC's source tree that are not used indeed feels weird.

fbrausse: > Yes for the 3rd question,so maybe I can create a new file in c2goto library directory and put all kernel functions in ?

Yes, that would be great. Just put it in the `c2goto/library` directory for now.

src/c2goto/headers/ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
Kai-zhichengzhou: hi，I have fixed all the issues we discussed, and pushed them. May I have a check on those changes to see if there is further issues? thanks !

src/c2goto/library/kernel.c
```diff
@@ -0,0 +1,106 @@
...
+unsigned long copy_to_user(void *to, void *from, unsigned long size)
+{
+  //checking on the passed parameters of kernel function
+  //the source in kernel space and destination in user space must be valid
+  assert(to != NULL);
+  assert(from != NULL);
+  assert(size <= PAGE_SIZE);
+
+  assert((char *)to >= user_memory);
+  assert((char *)from >= kernel_memory);
```
fbrausse, line 85: ~If these asserts hold, it doesn't mean we're copying from kernel_memory to user_memory. Could as well be from kernel to kernel or from user to user, depending on where these arrays are put in memory. We only know they're not overlapping because they're different objects.~

At the moment you don't have test cases copying from user to user or kernel to kernel. They should be failing. Could you add these two variants?

fbrausse (commented):
> hi，I have fixed all the issues we discussed, and pushed them. May I have a check on those changes to see if there is further issues? thanks !

Many thanks for addressing my comments. I think I understand now why you initially had the asserts in copy_to/from_user the way they were. ~The current version doesn't look right to me.~ **Edit:** Nevermind, it looks fine. The reason is that addresses of objects are chosen "non-deterministically", so if there is a way to lay out the objects such that the assertions fail, the verification will fail. That's why it's fine. Still, it would be good if you could add the user-to-user and kernel-to-kernel cases so we can be sure that those cases fail.

fbrausse (commented):
Another two comments: The test cases you add all copy the definition of `USER_MEMORY_SPACE` and the kernel variant of that. Shouldn't they #include the new uaccess.h header instead? Secondly, it's kind of equivalent from a functionality point of view, but defining the arrays holding specific types of memory in the test cases rather than in the new c2goto/library/kernel.c seems strange to me since the kernel.c is supposed to model the kernel, which has full control over memory. Did you have a specific reason to put the user/kernel_memory definitions into the test cases instead?

src/c2goto/library/kernel.c
Kai-zhichengzhou: hi，I have added the cases based on your suggestion, also i have modeled the memory in kernel.c instead of defining them in test files. May I have a check? Thank you

fbrausse (approved):
Looks good to me!

lucasccordeiro (approved):

lucasccordeiro:
@Kai-zhichengzhou: thanks for submitting this PR!

PR #1194, opened 2023-07-13
Improvements for removing unreachable code

shmarovfedor:
This pull request revisits the code (and some places around it) used for detecting and removing unreachable code in GOTO programs. This PR mostly deals with refactoring and adding some trivial cases.

Here is the list of things this PR addresses:

- Adds several simple criteria that qualify when an instruction cannot have a successor.
- Addresses some inconsistencies in the naming of the `is_skip` methods. One of them checks if the given instruction is of type `SKIP`, while the other one determines if the given instruction can be considered as `NO-OP` (which includes `SKIP` instructions and several other cases where the instruction type may not be a `SKIP`). As a result, the latter was renamed accordingly, and the related method `remove_skip` was also renamed to `remove_no_op` (as well as the names of the corresponding source files).
- Makes the `remove_unreachable` method to actually remove unreachable instructions instead of replacing them with `SKIP` and calling `remove_no_op` (former `remove_skip`; see above) on the modified program. This is a minor correction just to be more precise, since "unreachable" and "no-op" are not the same thing.
- Refactors the `get_goto_functions` method by splitting it into multiple methods for clarity.
- Introduces comments throughout all the code that is affected by this PR.

This PR does ***not*** address the removal of unused functions (suggested by @mikhailramalho). This will be handled via a different PR in the near future, since some preparation work needs to be done to better facilitate this feature.


mikhailramalho:
Interesting patch!

I suggest you also look at how CBMC removes function pointers when it can statically define them. We can further simplify the code (by removing the function pointers) and remove goto functions that are never called.

shmarovfedor:
Yes! That's exactly what I was planning to do next.

lucasccordeiro (approved):
@shmarovfedor: can I ask you to run this PR over the SV-COMP benchmarks with the removal of unreachable code enabled?

shmarovfedor:
Master this PR is based on (4640d2ec1d68849dc295d39c9d6c269f2ada1344)
```
  correct:           14419
    correct true:     8201
    correct false:    6218
  incorrect:            63
    incorrect true:     29
    incorrect false:    34
  unknown:           11977
```

This PR (a7f26ccbccfd6ac6b999dc32b3f8d48153e06188)
```
  correct:           14420
    correct true:     8198
    correct false:    6222
  incorrect:            63
    incorrect true:     29
    incorrect false:    34
  unknown:           11976
  Score:             21146 (max: 43043)
```

lucasccordeiro (approved):

rafaelsamenezes (approved):

src/goto-programs/remove_unreachable.cpp
```diff
@@ -1,40 +1,67 @@
...
     }
   }
 
-  // make all unreachable code a skip
-  // unless it's an 'end_function'
-
-  Forall_goto_program_instructions(it, goto_program)
+  // Removing all the instructions that do not appear in the constructed
+  // reachable set. Note that we never remove the END_FUNCTION instructions
+  // (even when they are unreachable)
```
fbrausse, line 58: I wonder whether you can say something about why the END_FUNCTION instructions are kept. Is it for syntactic reasons, i.e., to be able to more easily reconstruct the function boundaries?

fbrausse (approved):
LGTM. I just have a question out of interest.

lucasccordeiro:
@shmarovfedor: Thanks for submitting this PR ;-)

src/goto-programs/remove_unreachable.cpp
shmarovfedor: I believe so. I think a lot of times END_FUNCTION is also also used as a target for the "RETURN jumps". I will need to further investigate the importance of this kind of instructions in the future refactoring of the goto_programt, etc. classes.

fbrausse: Thanks, @shmarovfedor.

PR #1195, opened 2023-07-14
Don't interpret __builtin_unreachable() for SV-COMP

fbrausse:
SV-COMP doesn't define a body for it and it doesn't give rules about it either. See <https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1296>.
So there's no point in treating it as `assert(0)` like ESBMC does by default.
This triggered us to return an incorrect FALSE_REACH result for (at least)

	c/ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--mtd--devices--docg3.ko.cil.i

I'll run it over the benchmarks...

### DIFF
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -211,6 +211,8 @@ esbmc_path = "./esbmc "
 # ESBMC default commands: this is the same for every submission
 esbmc_dargs = "--no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets "
 esbmc_dargs += "--no-align-check --k-step 2 --floatbv --unlimited-k-steps "
+# <https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1296>
+esbmc_dargs += "-D'__builtin_unreachable()' "
 
 import re
 def check_if_benchmark_contains_pthread(benchmark):


fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/5540520772) (18d016a377ffe9e73b1b4ef40c9cabd90a068d39):
```
  correct:           14415
    correct true:     8194
    correct false:    6221
  incorrect:            61
    incorrect true:     28
    incorrect false:    33
  unknown:           11983
  Score:             21185 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5553936132) (992a3faf306c93ec7d312bb76ed84485614a57dd):
```
  correct:           14417
    correct true:     8198
    correct false:    6219
  incorrect:            61
    incorrect true:     28
    incorrect false:    33
  unknown:           11981
  Score:             21191 (max: 43043)
```
No noteworthy change for 30s. :/

fbrausse:
More details:
- The benchmark mentioned above is unchanged: timeout -> timeout
- 1x incorrect -> timeout, this is the kernel 3.14 analogue of the above case with the same __builtin_unreachable() usage
- 1x timeout -> incorrect, this was incorrect already on master

As far as I can see the rest is the usual fluctuation due to timeouts.

lucasccordeiro (approved):

PR #1112, opened 2023-06-21
Fixed missing argument in copy constructor

kunjsong01:
Fixed https://github.com/esbmc/esbmc/issues/1104

GOTO before:
```
        A return_value$_A$1;
        // 39 
        FUNCTION_CALL:  return_value$_A$1=A(&a2)
```

GOTO after:
```
        A return_value$_A$1;
        // 39
        FUNCTION_CALL:  A(&return_value$_A$1, &a2)
```

Enabled 1 TC: 
- `cbmc/Templates19`

PR #1196, opened 2023-07-15
Fix temporary object constructs for a single argument

### LINKED ISSUE #1176: Ctor with multiple default initializers failed
Originally posted in https://github.com/esbmc/esbmc/issues/1126#issuecomment-1607731941

Temporary objects
```
#include <cassert>
struct dim3
  {
    unsigned int x, y, z;
    dim3(unsigned int vx = 1, unsigned int vy = 1, unsigned int vz = 1)
    {
      x = vx;
      y = vy;
      z = vz;
    }
  };

int main() 
{ 
 dim3(1);
 
 return 0;
}
```

Cmd: `esbmc main.cpp`
Result:
```
Violated property:
  
VERIFICATION FAILED
```
source: https://github.com/XLiZHI/esbmc/blob/master/src/cpp/library/CUDA/call_kernel.h#L1241-L1245

version: https://github.com/esbmc/esbmc/commit/db4d28d4bfdbb8a80f118c15ba6fa4f5f70ad0ea

When we create temporary objects, the constructor should be able to accept any number of arguments from 0 to 3 since we define default initial values in the struct.

But verification identification occurs when we take only one argument, goto function is as follows:
```
main (c:@F@main#):
        // 32 
        FUNCTION_CALL:  dim3(( dim3 *)1, 1, 1)
```
The verification is successful under the remaining number of parameters:
```
        FUNCTION_CALL:  dim3((unsigned int)(&tmp$1), 1, 1, 1)
        // 33 
        tmp$1;
```
It is not known why creation of temporary objects fails with only one argument.

XLiZHI:
This PR optimizes the generation of temporary objects.

- [x] Correct the wrong exprt.
https://github.com/esbmc/esbmc/blob/91b38cf9f3e6529115c8a7c257e0cd16513c9629/src/clang-cpp-frontend/clang_cpp_convert.cpp#L928
In the current C++ frontend, we put the 'this' pointer to the end of the parameter, which caused us to get the wrong `expr`. We should put it at the beginning.

- [x] Fix missing parameters in the return value.
For TC: `assert(dim3(3).x == 3);`
This special case does not generate a temporary object, so we directly assign the value to the generated return value.

we adjust the statement:
```
        dim3 return_value$_dim3$1;
        // 35 
        FUNCTION_CALL:  return_value$_dim3$1=dim3(( dim3 *)3, 1, 1)
```
to
```
        // 34 file main2.cpp line 17 column 10 function main
        dim3 return_value$_dim3$1;
        // 35 
        FUNCTION_CALL:  dim3(&return_value$_dim3$1, ( dim3 *)3, 1, 1)
```

- [x] Wait for #1189 to be merged.


XLiZHI:
goto-functions
before:
```
main (c:@F@main#):
        // 32 
        FUNCTION_CALL:  dim3(( dim3 *)1, 1, 1)
```

now:
```
main (c:@F@main#):
        // 34 file main.cpp line 16 column 3 function main
        dim3 tmp$1;
        // 35 no location
        FUNCTION_CALL:  dim3((unsigned int)(&tmp$1), 1, 1, 1)
        // 36 
        tmp$1;
```

lucasccordeiro (approved):

kunjsong01:
> goto-functions before:
> 
> ```
> main (c:@F@main#):
>         // 32 
>         FUNCTION_CALL:  dim3(( dim3 *)1, 1, 1)
> ```
> 
> now:
> 
> ```
> main (c:@F@main#):
>         // 34 file main.cpp line 16 column 3 function main
>         dim3 tmp$1;
>         // 35 no location
>         FUNCTION_CALL:  dim3((unsigned int)(&tmp$1), 1, 1, 1)
>         // 36 
>         tmp$1;
> ```

The first parameter in the `FUNCTION_CALL` doesn't look quite right... I reckon it should be just `&tmp$1`

Could you double check with the `dim3`'s GOTO function signature please?

XLiZHI:
> The first parameter in the `FUNCTION_CALL` doesn't look quite right... I reckon it should be just `&tmp$1`

I did the same with before we deal with the `case clang: : Stmt: : CXXTemporaryObjectExprClass`, this means that there is a problem with the previous processing. Do you have any suggestions?

kunjsong01:
>Do you have any suggestions?

We need to know where the`(unsigned int)` typecasting comes from, and probably need to avoid it in this case.

regression/esbmc-cpp/bug_fixes/1176_issue/main.cpp
```diff
@@ -0,0 +1,21 @@
...
+    {
+      x = vx;
+      y = vy;
+      z = vz;
+    }
+};
+
+int main() 
+{
+  assert(dim3(3).x == 3);
```
kunjsong01, line 16: Is `assert(dim3(3).z == 1);` mimicking any syntax or semantics in CUDA OM that's blocking us? 

If not, I'd suggest close this PR, and probably better to add a representative test case in 1189 and just focus on that PR. 

It's probably not the best thing to do to overcomplicate the code based on some edge cases. This is going to affect our debugging, maintanence and future extensions.

kunjsong01 (changes requested):
1. Do we really need to support `assert(dim3(3).z == 1);` ?   
2. Even if we really want to support it, we probably should do it in the adjuster, instead of the converter. 

Screenshot 2023-07-17 at 17 59 17

The converter has done a pretty good job w.r.t temporaries. We just need to adjust the statements from
```
        dim3 return_value$_dim3$1;
        // 35 
        FUNCTION_CALL:  return_value$_dim3$1=dim3(( dim3 *)3, 1, 1)
```
to 
```
        dim3 return_value$_dim3$1;
        // 35 
        FUNCTION_CALL:  dim3(( dim3 *)(&return_value$_dim3$1), 3, 1, 1)
```

XLiZHI:
> 1. Do we really need to support `assert(dim3(3).z == 1);` ?   
> 2. Even if we really want to support it, we probably should do it in the adjuster, instead of the converter. 
> 
> Screenshot 2023-07-17 at 17 59 17
> 
> The converter has done a pretty good job w.r.t temporaries. We just need to adjust the statements from
> ```
>         dim3 return_value$_dim3$1;
>         // 35 
>         FUNCTION_CALL:  return_value$_dim3$1=dim3(( dim3 *)3, 1, 1)
> ```
> to 
> ```
>         dim3 return_value$_dim3$1;
>         // 35 
>         FUNCTION_CALL:  dim3(( dim3 *)(&return_value$_dim3$1), 3, 1, 1)
> ```
> 

Would you mind checking the difference between the current master and this patch here again, I think you may have labelled it incorrectly.

kunjsong01:
>Would you mind checking the difference between the current master and this patch here again, I think you may have labelled it incorrectly.

Nope, it's labelled correctly.
Since we had another merge to the master in past few hours, I'll just lablel it "master", instead of "current master". 

Here's the GOTO diffs between master vs. this patch(cc42cc49) and master vs. this patch (2de9367e1bc9, just one more commit from cc42cc49). 

Screenshot 2023-07-17 at 19 27 39

Screenshot 2023-07-17 at 17 59 17

What's really needs to be noticed here is that there's nothing wrong with the existing temporary creation for `CXXFunctionalCastExprClass` with a subexpression `CXXConstructExpr`. 

The current master is doing a pretty good job at it. Also, the temporary object has already been created as part of removing side effect flow in `goto_sideeffects.cpp` (we know `return_value$` is from the goto conversion right?). So we don't need to give a special case for `CXXFunctionalCastExprClass`, nor do we need to copy the conversion flow for `CXXTemporaryObjectExprClass`. 

If we really need to support the TCs in this PR and to do it in the ESBMC-way,  please consider doing some proper adjustments in either: 
1. adjuster
2. goto converter (probably as part of the removing side effect flow in goto_sideeffects.cpp)

Edit:
For 2, hope this PR could be helpful (https://github.com/esbmc/esbmc/pull/1112/files).

src/goto-programs/goto_sideeffects.cpp
```diff
@@ -770,28 +770,26 @@ void goto_convertt::remove_function_call(
 
   goto_programt tmp_program;
   const typet &ftype = call.function().type();
-  if(
-    ftype.return_type().id() == "constructor" &&
-    ftype.return_type().get_bool("#default_copy_cons"))
+  if(ftype.return_type().id() == "constructor")
   {
     // for copy constructor, we need to add the implicit `this` as the first argument,
```
kunjsong01: we probably need to update this comment too ;)

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -925,7 +925,7 @@ bool clang_cpp_convertert::get_function_this_pointer_param(
   this_param.cmt_identifier(id);
 
   // Add to the list of params
```
kunjsong01: It's probably better to update the comment here. "// Add `this` as the first parameter to the list of params". 
Or just move it if you wish, the code seems pretty self-explanatory.

kunjsong01 (changes requested):
Also, is there any new TCs to verify the code changes in this patch?

kunjsong01:
We should add back the TC with `assert(dim3(3).x == 3); ` and compare the GOTO programs generated. Once this is verified, the patch is good to go. Otherwise, we kinda lose track of what we want to fix with this patch.

regression/esbmc-cpp/bug_fixes/1176_issue/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
```
kunjsong01, line 4: The test case looks great! 
We just need one more TC for the "VERIFICATION FAILED" counterpart. `assert(dim3(3).z == 3); // should be 1` should do. 
Edit: ~~Also, would it be possible to add the old GOTO vs. new GOTO you've collected and shared in the slack channel? Those are pretty good information to show the manin purpose of this PR~~ Already done in the description.

kunjsong01 (changes requested):

kunjsong01 (approved):

lucasccordeiro:

PR #1205, opened 2023-07-19
Fix argc and argv handling

### LINKED ISSUE #1064: [frontend] ESBMC does not handle argc and argv correctly

brcfarias:
Adjust __ESBMC_alloc_size as argv is handled as dynamic array.
Fixes #1064


fbrausse (commented):
Could you please add the test cases from #1064 that this PR fixes?

fbrausse:
The following tests FAILED:
	116 - regression/esbmc-unix/01_broadcast_01 (Failed)
	183 - regression/esbmc-unix/03_boundedBuffer (Failed)
	1682 - regression/nonz3/github_1070 (Failed)
	1688 - regression/z3/github_1070 (Failed)

1682 and 1688 are the same. They assume there is a valid '\0'-terminated C string in `argv[1]`. This PR isn't providing all the `argc+1` elements of the `argv` array with valid C strings or NULL for the last one, yet. So that's expected. IMO we can update the test.desc regex to match this expected behaviour.

The expected verdict of 116 is `VERIFICATION FAILED` and master does that claiming `array bounds violated`, in this line:
```c
  printf("Enter Testcase - %s\n", argv[0]);
```
I think that's unreasonable. Accessing `argv[0]` should always be valid. We should update the verdict.

183 is interesting:
```
ERROR: __ESBMC_set_start_data received nonconstant thread id
```
Looking into it.

lucasccordeiro (changes requested):
@brcfarias : Could you please add the test cases from https://github.com/esbmc/esbmc/issues/1064?

Also, some regression tests are failing for this PR:

116 - regression/esbmc-unix/01_broadcast_01 (Failed)
183 - regression/esbmc-unix/03_boundedBuffer (Failed)
1682 - regression/nonz3/github_1070 (Failed)
1688 - regression/z3/github_1070 (Failed)

regression/esbmc-unix/03_boundedBuffer/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
fbrausse, line 1: The label claims this bug is known. It is not, though, at least to me. :)
@brcfarias Could you please open an issue so we know what it's about and where it came from?

fbrausse (approved):
LGTM, except we should open an issue.

shmarovfedor (approved):
LGTM

lucasccordeiro (approved):

lucasccordeiro:
@brcfarias: thanks for submitting this PR!

Issue #1171, opened 2023-07-10
Crash for memset on float types


fbrausse:
In a Debug build the regression test `regression/bitwuzla/memset-good` crashes with
```
$ src/esbmc/esbmc ../regression/bitwuzla/memset-good/memset-good.c --bitwuzla
[...]
esbmc: /home/kane/scorch/esbmc/src/goto-symex/builtin_functions.cpp:980: expr2tc gen_value_by_byte(const type2tc&, const expr2tc&, const expr2tc&, size_t, size_t): Assertion `!is_floatbv_type(type) && !is_fixedbv_type(type)' failed.
Aborted
```

PR #1206, opened 2023-07-20
Move add_padding() from convert to adjust

### LINKED ISSUE #637: recursive struct/union types

### LINKED ISSUE #1202: Move add_padding to adjust
add_padding needs to be moved from the converter to the adjuster. 

### Previous attempt: 
- https://github.com/esbmc/esbmc/tree/adjust_type
 
### Evaluation: 
- https://github.com/esbmc/esbmc/pull/1180#issuecomment-1640012394

### LINKED ISSUE #1210: Verdict for accessing symbols of incomplete type
In the context of #1206 I've been trying around with incomplete types and found this:
```c
#include <assert.h>
#include <stdlib.h>
#include <string.h>

struct incomplete;

extern struct incomplete JJ;

int main()
{
	int k = 42;
	memcpy(&JJ, &k, sizeof(k));
	int j;
	memcpy(&j, &JJ, sizeof(k));
	assert(j == 42);
}

// struct incomplete { short x, y; }; /* uncommment for VERIFICATION SUCCESSFUL */
```
Current master says `VERIFICATION FAILED` due to out-of-bounds access.

The technical reason for that is clear: `struct incomplete` is being treated as a complete, but empty struct, see:
```
Symbol......: c:@JJ
Module......: incomplete
Base name...: JJ
Mode........: C (0)
Type........:  struct incomplete { }
Value.......: 
Flags.......: lvalue static_lifetime extern
Location....: file incomplete.c line 8
```
However, according to C it is not. It depends on what we're linking with and it doesn't matter what name the struct type of the symbol `JJ` has in another C translation unit: Link this with it and it's fine (for a C compiler/linker):
```c
struct complete { int v; };

struct complete JJ;
```
It is even fine for ESBMC, though somewhere in the middle it does print
```
ERROR: error: conflicting definition for variable `JJ'
old definition:  struct complete { signed int v; }
Module: incomplete-lib
new definition:  struct incomplete { }
Module: incomplete
Location: file incomplete.c line 8 column 1
```
So for me it seems clear that making `struct incomplete` a *complete and empty struct type* `struct incomplete { }` in the contextt is wrong for the first translation unit.

Secondly I wonder whether ESBMC should give a verdict for the underspecified first "program" at all. It does choose to do so when calling a just-declared-but-not-defined function, which happens much more often, by assuming that function can be called and doesn't do anything violating conditions. Here similarly little information is known about the nature of `JJ` and its type but in contrast ESBMC assigns the most restrictive semantics to it. Should it? Why is the first program alone not `VERIFICATION UNKNOWN`?

fbrausse:
This PR is another attempt to solve #637 by the method discussed in #1202. It mostly touches the clang-c-frontend, and a bit in utils. I haven't yet checked compatibility with clang-cpp (hoping that the CI will find issues), but I would ask @kunjsong01 to look over it, too.

1. Extend the type_byte_size() family of functions with an optional namespace parameter. This allows it to give meaningful results also for symbolic types. Reminder: symbolic types are similar to typedefs, they refer (by identifier) to a named type symbol in the namespace's underlying context.
2. The "convert" part of the frontend is now generating only symbolic type references for structures (structs and unions) instead of copying the type of the symbol in the context. It also no longer pads the structures it generates thanks to a patch by @mikhailramalho, which this PR is based on.
3. The converter now generates "incomplete_struct" and "incomplete_union" types when Clang does not have a definition of the type in the current translation unit. This is in line with how C sees types: the types in one translation unit are completely independent of the types in another one. This is different from C++, so we'll need to be careful here. In particular migrate_type() does questionable things to incomplete types... Before, these were treated as completely defined empty structures, which they are not.
4. add_padding() is now being done on all complete structure types in the adjust_type() phase running during typecheck().
5. adjust_expr() takes care of adjusting the initializer of structures whose type has been padded. There should be no difference in representation afterwards (i.e., padding members are initialized to zero just as before).
6. I've added a comment about a wrong assumption in `is_aligned_member()` in the dereference module. I'm not attempting to fix it in this PR.


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1022,28 +1028,11 @@ bool clang_c_convertert::get_type(const clang::Type &the_type, typet &new_type)
...
-
-    // avoid unnecessary conversion if symbol already exists
-    if(!sym_already_exist)
-      if(get_struct_union_class(rd))
-        return true;
-
-    symbolt &s = *context.find_symbol(id);
-    new_type = s.type;
+    /* record in context if not already there */
+    get_struct_union_class(rd);
```
mikhailramalho, line 1034: Do we need this here?  `get_type` is only called when we try to get the type of an expr, which either has its type already defined or is being defined (like the C++ iterator example from @kunjsong01).

Can't we just the ` new_type = symbol_typet(id);`?

src/clang-c-frontend/symbolic_types.cpp
```diff
@@ -1,63 +1,37 @@
...
+    assert(complete_type->is_type);
+    assert(!complete_type->type.is_symbol());
+    type = complete_type->type;
+    get_complete_type(type, ns);
+    return true;
+  }
+  return false;
 }
 
 bool contains_symbolic_struct_types(
```
mikhailramalho: can we get rid of this method and use `get_complet_type` directly? I think this method is only used in one place anyway.

src/util/type_byte_size.cpp
```diff
@@ -28,33 +44,43 @@ BigInt member_offset_bits(const type2tc &type, const irep_idt &member)
...
+BigInt member_offset_bits(
+  const type2tc &type,
+  const irep_idt &member,
+  const namespacet *ns)
+{
+  return type_sizet(ns).member_offset_bits(type, member);
+}
+
+BigInt
+member_offset(const type2tc &type, const irep_idt &member, const namespacet *ns)
```
mikhailramalho, line 63: const pointer -> const reference?

src/util/type_byte_size.h
```diff
@@ -7,21 +7,35 @@
...
+  const type2tc &type,
+  const irep_idt &member,
+  const namespacet *ns = nullptr);
 
 /* These can throw array_type2t::inf_sized_array_excp or
  * array_type2t::dyn_sized_array_excp */
-BigInt type_byte_size_bits(const type2tc &type);
-BigInt type_byte_size(const type2tc &type);
-BigInt type_byte_size_default(const type2tc &type, const BigInt &defaultval);
+BigInt type_byte_size_bits(const type2tc &type, const namespacet *ns = nullptr);
```
mikhailramalho, line 21: I say we make the namespacet not default to nullptr and require it to always be passed to this functions (and related).

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -333,21 +330,38 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
...
+  struct_union_typet t(rd.isUnion() ? "union" : "struct");
+  t.tag(name);
+  t.location() = location_begin;
 
   // Don't continue to parse if it doesn't have a complete definition
   // Try to get the definition
   clang::RecordDecl *rd_def = rd.getDefinition();
   if(!rd_def)
+  {
+    t.id("incomplete_" + t.id().as_string());
```
mikhailramalho: do we need to change the id here? I'd assume setting incomplete(true) is enough

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1967,40 +1956,37 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
-    typet complete_type;
-    bool contains_symbolic =
-      contains_symbolic_struct_types(t, complete_type, ns);
-    if(contains_symbolic)
-      t = complete_type;
+    const typet &t_ref = ns.follow(t);
 
     // Structs/unions/arrays put the initializer on operands
-    if(t.is_struct() || t.is_array() || t.is_vector())
+    if(t_ref.is_struct() || t.is_array() || t.is_vector())
```
mikhailramalho: t_ref for array and vector as well?

src/clang-c-frontend/symbolic_types.cpp
```diff
@@ -1,63 +1,37 @@
...
-        // follow the type recursively to replace all symbolic types
-        get_complete_struct_type(to_struct_type(it_type), ns);
-      }
-    }
+    bool r = false;
+    for(auto &comp : to_struct_union_type(type).components())
+      r |= get_complete_type(comp.type(), ns);
+    return r;
   }
+  else if(type.is_array())
```
mikhailramalho: we can remove the `else` here if the previous `if` returns.

fbrausse:
@mikhailramalho Thank you for the comments! Let me get it working first, though ;)

src/clang-c-frontend/clang_c_convert.cpp
fbrausse: Unfortunately not. `type.is_struct()` or `type.id() == "struct"` (same for union) is used often and many of those won't work for incomplete structures. Before this PR, `incomplete()` was completely unused, even the old frontend didn't use it. So no, just setting incomplete(true) won't work. The rest of the code is just not expecting it.

fbrausse:
OK, quick heads up - or maybe rather heads down: While the CI is cancelling the Linux runs, I've run them locally and a **huge** amount of C++ regression tests are timeouting. Without having investigated deeply, my suspicion is that the C++ frontend is using ns.follow(). @kunjsong01 Can you confirm this? If so, then we may need an alternative solution, though I thought this one here was quite neat...

*Edit:* Suspicion because of that inside `get_struct_union_class()`:
```c
  /* The downside of this approach is that we can't use ns.follow() until this
   * symbol (and all type symbols up our callstack) are properly constructed:
   * it would loop forever. */
```

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: by rest of the code, you mean only adjust, right? Since there should be no incomplete type when we start goto conversion, either they are complete or they are symbolic_typet.

fbrausse: I don't mean adjust only. Everything else. Until SMT. Look at migrate_type() for instance, there is a case about incomplete_struct. It might be a remnant from the good old days™, but still: in my opinion an incomplete type (one that just has been declared, but never defined) is different from a symbolic type (which is basically a typedef). A pointer to an incomplete type can never be dereferenced. On the other hand a pointer to a symbolic type *could be*, depending on whether ns.follow(sym_type) is complete or not.

fbrausse: I have been too quick to reply, sorry. It's not until SMT, just until c_link. And we'll probably need it until there in order for C++ symbols to be resolved across translation units. Nonetheless, I still believe incomplete type != symbolic type.

kunjsong01:
>Without having investigated deeply, my suspicion is that the C++ frontend is using ns.follow(). @kunjsong01 Can you confirm this?

Hi @fbrausse, there is no `ns.follow()` explicitly used in C++ adjuster. But the C and C++ adjusters share lots of common adjustment procedures. So I did a quick grep in both adjuster source code and found 6 methods in the C adjuster explicitly using `ns.follow()`:
```
1. clang_c_adjust::adjust_expr_shifts 
  - for exprt with id "shl" and "shr"; 
  - no overriding in cpp adjuster
2. clang_c_adjust::adjust_float_arith
  - part of the adjust_expr_binary_arithmetic flow; 
  - no overriding in cpp adjuster
3. clang_c_adjust::adjust_dereference 
  - for deref expression; 
  - no overriding in cpp adjuster
4. clang_c_adjust::adjust_index 
  - for pointer deref;
  - no overriding in cpp adjuster
5. clang_c_adjust::do_special_functions
  - for built-in function stuff;
  - no overriding in cpp adjuster
6. clang_c_adjust::adjust_function_call_arguments
  - OVERRIDING in cpp adjuster
  - used in clang_cpp_adjust::adjust_side_effect to deal with temporary_object 
  - part of adjust_side_effect_function_call flow
    > no overriding in cpp adjuster
    > used in clang_cpp_adjust::adjust_side_effect_assign
```

Procedure 1-5 might be of no interest w.r.t the C++ regressions. 
6 might be worth looking into.

fbrausse:
@kunjsong01 Many thanks for looking into it! The adjuster should be fine to use ns.follow(), however, with this PR's approach we can only be sure about using it in the converter when we're outside of clang_c_convertert::get_struct_union_class().

With the C++ frontend ns.follow() is invoked in via gen_typecast() on types currently being constructed at a few locations, such as the `std::string::iterator` in clang_c_convertert::get_expr().

Maybe it's worth switching from recording a symbolic type to an incomplete one. Though then the adjuster or something during convert might need to update all the places that's been used as soon as the definition of the complete type is finished.

src/clang-c-frontend/clang_c_convert.cpp
fbrausse: That would lead to the definition of the anonymous union in `__mbstate_t` missing, see <https://github.com/esbmc/esbmc/pull/1180#issuecomment-1640037798>.

fbrausse:
**Edit:** Ignore this, it's superseded by the latest change.

91% tests passed, 303 tests failed; of the failed ones, these are frequencies for assert failures:
```
      2 esbmc: esbmc/src/clang-c-frontend/clang_c_adjust_expr.cpp:616: void clang_c_adjust::adjust_type(typet&): Assertion `sz % a == 0' failed.
      1 esbmc: esbmc/src/goto-symex/builtin_functions.cpp:984: expr2tc gen_value_by_byte(const type2tc&, const expr2tc&, const expr2tc&, size_t, size_t): Assertion `!is_floatbv_type(type) && !is_fixedbv_type(type)' failed.
    294 esbmc: esbmc/src/util/namespace.cpp:66: const typet& namespacet::follow(const typet&) const: Assertion `next != symbol && "cycle of length 1 in ns.follow()"' failed.
```
The 2 `sz % a` is a check whether the type size modulo its alignment is zero after add_padding(). @mikhailramalho Would you care to take a look why add_padding() fails for those? Another one where add_padding() behaves strangely is `llvm/struct_union`, I've marked it a KNOWNBUG in <https://github.com/esbmc/esbmc/pull/1206/commits/da1298bb94958cf81fd940d83cf436ba7ff0d386> to get the CI run Linux.

The 1 !fixed/!float one is known: #1171.

I'll look into the 294 ones with cycles in ns.follow(); they're mostly C++ tests. And also why the remaining (303-297) ones fail.

Edit: This is the [ctest.log](https://github.com/esbmc/esbmc/files/12124945/ctest.log).

fbrausse:
Some of the C++ ones are failing in adjust_type() at `context.find_symbol(id)` where `std::string id = "tag-" + type.tag().as_string();` but the types don't have `tag()`, only `name()`, e.g. `"virtual_table::tag-std::streambuf"`. @kunjsong01 Why don't these have a `tag()`?

src/util/type_byte_size.h
fbrausse: It's quite a lot of places that use these functions. I won't mandate a change of signature in this PR. If this PR ever works.

fbrausse:
OK, now it seems to work much better. The `ns.follow()` madness has been resolved. Besides fixed ones, there are new KNOWNBUG regression tests:
- assert(type.is_struct()) failed in clang-cpp frontend for union types:
  - cuda/COM_sanity_checks/cuda
  - cuda/COM_sanity_checks/curand_kernel
  - esbmc-cpp/cpp/ch20_6
- add_padding() produces a type whose size is not a multiple of its alignment:
  - esbmc-unix/github_1199-1
  - incremental-smt/incremental-28

src/clang-c-frontend/symbolic_types.cpp
fbrausse: Done.

src/clang-c-frontend/clang_c_convert.cpp
fbrausse: I've removed t_ref and updated t directly.

src/clang-c-frontend/symbolic_types.cpp
fbrausse: Done.

src/util/type_byte_size.cpp
fbrausse: If we're not going to start overloading all these functions (one with, one without namespace parameter), I need a way to say "optional". As the implementation would use a pointer anyway to represent this option (unless we go template-crazy), I thought this would be a cleaner API.

fbrausse:
Well, it seems `incomplete_struct` does survive even after c_link: The example from #1210 produces a segfault with this PR here: https://github.com/fbrausse/esbmc/blob/fb/adjust_type/src/pointer-analysis/dereference.cpp#L2187

I think that's actually good. It now allows us to assign more meaningful semantics to it.

fbrausse:
After adding support for #1210, there is a new issue in migrate_expr(): In the case `extern int JJ[];` and accesses to it in one file and `struct complete { int v; } JJ;` in another file, we're trying to build an index-expression into the struct...

*Edit:* Already happens on master.

src/util/type_byte_size.h
mikhailramalho: Since we are changing the behaviour of the function that will require the namespace, we should make it mandatory to pass the namespace. That way we can prevent clients from misusing the API, e.g., passing a symbol_typet but not passing the namespace.

mikhailramalho (commented):
@fbrausse, can you provide a test case where we absolutely need to use the incomplete struct/union, instead of a symbol type? Maybe we can work out a way of not needing it.

My motivation for not having it is as you said: it was never used before in the old frontends.

fbrausse:
> can you provide a test case where we absolutely need to use the incomplete struct/union, instead of a symbol type? Maybe we can work out a way of not needing it.
> 
> My motivation for not having it is as you said: it was never used before in the old frontends.

@mikhailramalho Just so I know we talk about the same thing: You are referring to the initial `id()` that's set for the new symbol in the context, yes? Here: https://github.com/esbmc/esbmc/blob/b0580ac8abb3151655bd173374cd883fd2788689/src/clang-c-frontend/clang_c_convert.cpp#L315

Because my comment about usage in the old frontend  was about `irept::incomplete()`, which is different from the `id()`. The old frontend did/does use `incomplete_struct` and `incomplete_union` in https://github.com/esbmc/esbmc/blob/b0580ac8abb3151655bd173374cd883fd2788689/src/ansi-c/parser_static.inc#L241-L242 and also in https://github.com/esbmc/esbmc/blob/b0580ac8abb3151655bd173374cd883fd2788689/src/ansi-c/parser.ypp#L966-L967 for declarations. 

That's basically what I'm trying to do with the `incomplete_*` id here as well: Until the type is fully defined, we only have a declaration. #1210 is about that.

Turns out I was wrong though and have overlooked one case, sorry for misleading you: the old frontend sets .incomplete() here, appearently for a function's type: https://github.com/esbmc/esbmc/blob/b0580ac8abb3151655bd173374cd883fd2788689/src/ansi-c/c_typecheck_expr.cpp#L1150

By using `symbol_typet()` instead of `incomplete_*` there are plenty of (especially C++) cases that run into infinite loops, due to `gen_typecast()` (on the `this` pointer for example). `gen_typecast()` needs to do ns.follow() to performs its operations, I think. At least I haven't been able to make it not do that. And ns.follow() will run into an infinite loop though, in 2 cases:
1. there is no complete definition of the type in the entire program, (*edit*: this is #1210), and/or
2. ns.follow() is being called *while* we're parsing a type definition (*edit:* this is #637, which I'm trying to solve here as well).

Besides #1210 I can try to find a C-only example that fails when going back to revision b4d22ab8169c6cad5f651ef5e13f25f3a067224c before I switched to `incomplete_*` in the context. My problem with the symbol_typet is that it's meant to be ns.follow'ed. If that's impossible because of loops it's a time bomb. If we go to the symbol-type, how can we avoid the loop?

src/util/type_byte_size.h
fbrausse: I don't understand. We are not changing the behaviour of any existing callers that use the default value for `ns`. They already know they can't pass symbol-types into these functions because that'll lead to abort(). And symbol-types should all be dealt with in thrash_type_symbols. No change here. If not all have been replaced or some part before thrash_type_symbols uses these functions, that's a semantic change of this PR we should catch, I agree. But not for those after. In some cases there not even is a namespace available, in simplify() for example. Changing each call site and by extension passing some(?) namespace to those that don't have one is too invasive for this PR.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/5620760754) (4640d2ec1d68849dc295d39c9d6c269f2ada1344):
```
  correct:           14419
    correct true:     8201
    correct false:    6218
  incorrect:            63
    incorrect true:     29
    incorrect false:    34
  unknown:           11977
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5638920414) (b0580ac8abb3151655bd173374cd883fd2788689):
```
  correct:           14375
    correct true:     8158
    correct false:    6217
  incorrect:            63
    incorrect true:     30
    incorrect false:    33
  unknown:           12021
  Score:             21045 (max: 43043)
```
We got a flip: 1x incorrect -> timeout and 1x timeout -> incorrect. Looking into it. More importantly, with the current approach benchmarked here, there are significant number of cases that are new unknowns, many of them due to symbolic types where they're not expected. These might be the cases @mikhailramalho wants to cover by making the namespace parameter mandatory, not sure. I'll check. Some of the new unknowns are also:
- `Looking up index of nonexistant member "anon_pad$1" in struct/union "struct cdev"` (or similar: pad members in some structs)
- `throwing an instance of 'array_type2t::inf_sized_array_excp'`

I'll mark this PR a draft until I've sorted out these new unknowns.

fbrausse:
Update: 6 unknowns fixed, they were due to
- get_width() on sym-types calling abort() instead of throwing the right exception as done in type_byte_size*()
- still false handling of Clang's RecordDecl info

Remaining ones are unknown mainly due to add_padding():

1. sv-benchmarks/c/ldv-linux-3.16-rc1/43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
   release build: ERROR: Looking up index of nonexistant member "anon_pad$16" in struct/union "struct tty_port"
   debug build: src/clang-c-frontend/clang_c_adjust_expr.cpp:582: void clang_c_adjust::adjust_type(typet&): Assertion `sz % a == 0' failed.
   <details>

   ```
   (gdb) p sz.to_uint64()
   $2 = 33
   (gdb) p a.to_uint64()
   $3 = 8
   (gdb) p type.dump()
   struct
     * tag: struct attribute___0
     * components: 
       0: component
           * type: pointer
               * subtype: signedbv
                   * width: 8
                   * #constant: 1
                   * #cpp_type: signed_char
           * name: name
           * pretty_name: name
           * #location: 
             * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
             * line: 3595
             * column: 4
       1: component
           * type: unsignedbv
               * width: 16
               * #cpp_type: unsigned_short
           * name: mode
           * pretty_name: mode
           * #location: 
             * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
             * line: 3596
             * column: 4
       2: component
           * type: bool
               * width: 1
               * subtype: bool
                   * #cpp_type: bool
               * #cpp_type: bool
               * #bitfield: 1
           * name: ignore_lockdep
           * pretty_name: ignore_lockdep
           * is_unnamed_bitfield: 0
           * #location: 
             * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
             * line: 3597
             * column: 4
       3: component
           * type: unsignedbv
               * width: 7
               * #bitfield: 1
           * name: anon_bit_field_pad$3
           * pretty_name: anon_bit_field_pad$3
           * is_padding: 1
       4: component
           * type: unsignedbv
               * width: 40
           * name: anon_pad$4
           * pretty_name: anon_pad$4
           * is_padding: 1
       5: component
           * type: pointer
               * subtype: symbol
                   * identifier: tag-struct lock_class_key
           * name: key
           * pretty_name: key
           * #location: 
             * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
             * line: 3598
             * column: 4
       6: component
           * type: symbol
               * identifier: tag-struct lock_class_key
           * name: skey
           * pretty_name: skey
           * #location: 
             * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
             * line: 3599
             * column: 4
     * #location: 
       * file: 43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--tty--isicom.ko-entry_point.cil.out.i
       * line: 3594
       * column: 1
   ```
   </details>

2. sv-benchmarks/c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--crypto--qat--qat_common--intel_qat.ko-entry_point.cil.out.i
   release build: Program received signal SIGSEGV, Segmentation fault.
   debug build: src/pointer-analysis/value_set.cpp:1044: void value_sett::assign(const expr2tc&, const expr2tc&, bool): Assertion `base_type_eq(rhs->type, lhs_type, ns) || is_subclass_of(lhs_type, rhs->type, ns)' failed.
   -> problem: add_padding() is not idempotent
   <details>
   diff between rhs->type->dump() and lhs_type->dump():

   ```diff
   --- a	2023-07-27 10:38:59.292304556 +0200
   +++ b	2023-07-27 10:39:06.604388273 +0200
   @@ -1,4 +1,4 @@
   -(gdb) p rhs->type->dump()
   +(gdb) p lhs_type->dump()
    struct
    * members : 
      0: unsignedbv
   @@ -377,13 +377,15 @@
    
            * typename : struct kobject
            * packed : false
   -      1: pointer
   +      1: unsignedbv
   +        * width : 24
   +      2: pointer
            * subtype : symbol
                * symbol_name : tag-struct module
   -      2: pointer
   +      3: pointer
            * subtype : symbol
                * symbol_name : tag-struct file_operations
   -      3: struct
   +      4: struct
            * members : 
              0: pointer
                * subtype : symbol
   @@ -402,30 +404,34 @@
    
            * typename : struct list_head
            * packed : false
   -      4: unsignedbv
   -        * width : 32
          5: unsignedbv
            * width : 32
   +      6: unsignedbv
   +        * width : 32
    
        * member_names : 
          0: kobj
   -      1: owner
   -      2: ops
   -      3: list
   -      4: dev
   -      5: count
   +      1: anon_pad$1
   +      2: owner
   +      3: ops
   +      4: list
   +      5: dev
   +      6: count
    
        * member_pretty_names : 
          0: kobj
   -      1: owner
   -      2: ops
   -      3: list
   -      4: dev
   -      5: count
   +      1: anon_pad$1
   +      2: owner
   +      3: ops
   +      4: list
   +      5: dev
   +      6: count
    
        * typename : struct cdev
        * packed : false
   -  3: pointer
   +  3: unsignedbv
   +    * width : 40
   +  4: pointer
        * subtype : symbol
            * symbol_name : tag-struct class
    
   @@ -433,13 +439,15 @@
      0: major
      1: anon_pad$1
      2: drv_cdev
   -  3: drv_class
   +  3: anon_pad$3
   +  4: drv_class
    
    * member_pretty_names : 
      0: major
      1: anon_pad$1
      2: drv_cdev
   -  3: drv_class
   +  3: anon_pad$3
   +  4: drv_class
    
    * typename : struct adf_ctl_drv_info
    * packed : false
   ```
   </details>

3. sv-benchmarks/c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--vfio--vfio.ko-entry_point.cil.out.i
   same as item 2. on "struct vfio": add_padding() inserts padding at different places

4. - sv-benchmarks/c/ldv-linux-3.14/linux-3.14_complex_emg_linux-kernel-locking-spinlock_drivers-media-dvb-core-dvb-core.cil.i
   - sv-benchmarks/c/ldv-linux-3.14/linux-3.14_linux-kernel-locking-spinlock_drivers-net-wireless-ipw2x00-ipw2200.cil.i

   same as item 1 on "struct attribute"

5. sv-benchmarks/c/ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--staging--ozwpan--ozwpan.ko.cil.i
   same as item 2. on "struct oz_cdev"

fbrausse:
> [Master](https://github.com/esbmc/esbmc/actions/runs/5620760754) ([4640d2e](https://github.com/esbmc/esbmc/commit/4640d2ec1d68849dc295d39c9d6c269f2ada1344)):
> 
> ```
> Statistics:          26459 Files
>   correct:           14419
>     correct true:     8201
>     correct false:    6218
>   incorrect:            63
>     incorrect true:     29
>     incorrect false:    34
>   unknown:           11977
>   Score:             21148 (max: 43043)
> ```

[This PR](https://github.com/esbmc/esbmc/actions/runs/5680833957) (05414ffe5b5c2fe7550aa60ed004b00628e3f035):
```
  correct:           14387
    correct true:     8169
    correct false:    6218
  incorrect:            63
    incorrect true:     30
    incorrect false:    33
  unknown:           12009
  Score:             21068 (max: 43043)
```
The incorrect -> timeout and timeout -> incorrect flip is just jitter, both don't change results when run with larger timeout. No new unknowns anymore.

lucasccordeiro (approved):

rafaelsamenezes (approved):

fbrausse:
Since the SV-COMP numbers above are against a relatively old master (8 days ago), we might not want to update the stats-30s.txt with them.

@mikhailramalho Could I ask whether you're OK with this PR? We can make the namespace parameter mandatory in another PR, if you think that's a better design. Regarding the symbol-type, as I said, I don't know how to avoid the cycles where a symbol in the context refers to itself making ns.follow() a time bomb. This affects C++ mostly. Such cycles are also described here: https://github.com/esbmc/esbmc/blob/b6f905f81032485bd08a1decbe8b33b6e0a94106/src/goto-programs/goto_convert_functions.cpp#L265-L273

kunjsong01:
Hi @fbrausse, the patch below will fix the failures of CUDA TCs. 
[patch_for_CUDA_TCs.txt](https://github.com/esbmc/esbmc/files/12192585/patch_for_CUDA_TCs.txt)

Edit: also fixed the failure in `regression/esbmc-cpp/cpp/ch20_6` :D

fbrausse:
Many thanks @kunjsong01!

lucasccordeiro:
@fbrausse : thanks for submitting this PR!

PR #1216, opened 2023-07-24
Changed smtlib linker flags

### LINKED ISSUE #1212: "ld: unknown option: -z" on OSX
I am getting this error when trying to compile ESBMC on OSX: `ld: unknown option: -z`. This happens when `cmake` tries to configure Z3:

`CMake Error at src/solvers/z3/CMakeLists.txt:34 (message):
  Could not use Z3:`

Also when I try compiling without Z3, this error appears eventually during the ESBMC compilation.

rafaelsamenezes:
Should close #1212

### DIFF
--- a/src/solvers/smtlib/CMakeLists.txt
+++ b/src/solvers/smtlib/CMakeLists.txt
@@ -9,7 +9,7 @@ configure_file(smtlib_tok.lpp  ${CMAKE_BINARY_DIR}/smtlib_tok.lpp COPYONLY)
 
 add_library(smtlib smtlib_conv.cpp ${CMAKE_CURRENT_BINARY_DIR}/smtlib.cpp ${CMAKE_CURRENT_BINARY_DIR}/lexer.cpp)
 target_link_libraries(smtlib fmt::fmt)
-target_include_directories(smtlib     
+target_include_directories(smtlib
     PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
     PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
     PRIVATE ${CMAKE_BINARY_DIR}/src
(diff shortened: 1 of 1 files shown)


src/solvers/smtlib/CMakeLists.txt
```diff
@@ -21,7 +21,11 @@ target_include_directories(smtlib
 # could be very large and the std::list destructors use a lot of stack space
 set(stacksize 20971520)  # 20 MB ought to be enough for everyone
 if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:${stacksize}" CACHE STRING "Linker flags" FORCE)
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:${stacksize}" CACHE STRING "Linker flags" FORCE)
```
fbrausse: Indentation.

src/solvers/smtlib/CMakeLists.txt
```diff
@@ -21,7 +21,11 @@ target_include_directories(smtlib
...
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
```
fbrausse, line 25: Are you sure we should do this for Clang regardless of whether it's Apple's or not?

~Clang supports different linkers via `-fuse-ld=(bfd|lld|...)`, not sure about the Apple one. Unfortunately in https://gitlab.kitware.com/cmake/cmake/-/issues/18209 they say CMake does not and probably will not support something like CMAKE_LINKER_ID. Instead they suggest to use `try_compile()`.~ *Edit:* Your solution in https://github.com/esbmc/esbmc/issues/1212#issuecomment-1647849951 sounds much better. Doesn't have to be in this PR though, since we need the software to compile on macos as well.

rafaelsamenezes: > Are you sure we should do this for Clang regardless of whether it's Apple's or not?

I will confirm this later.

fbrausse (approved):
Thanks, @rafaelsamenezes. LGTM.

shmarovfedor (approved):
LGTM

Issue #1155, opened 2023-07-06
scanf() and gets() -different behavior


tihanyin:
Consider the following small C program :

```c
#include <stdio.h>
#include <string.h>
int main(int argc, char **argv)
{
  char name[15];
  char name2[15];
  printf("Enter your first name: ");
  gets(name);
  printf("Enter your last name: ");
  scanf("%s",name2);
}
```

Obviously we have a buffer overflow in gets() in line number 8 and in scanf() line number 10. 

The ESBMC output shows only the exact line number for scanf() but not for gets(). To enhance automatic vulnerability detection and repair, it would be advantageous to implement the capability of the EEBMC output to provide the specific line number for occurrences of gets(), similar to the existing functionality for scanf(). 

Here is the two different output:

`gets():` The exact line number is missing and also the function name (only the line number from io.c can be seen ,which is not relevant for vulnerability repair and bug fixing) 

```c

----------------------------------------------------
Violated property:
  file io.c line 79 function gets
  dereference failure: array bounds violated
VERIFICATION FAILED

```

`scanf():` However for scanf() the exact line number can be seen which is the good behavior.
```c

----------------------------------------------------
Violated property:
  file test.c line 10 function main
  buffer overflow on scanf
  0

```

tihanyin:
@lucasccordeiro I think this is already solved and can be closed. Right?

lucasccordeiro:
Yes ;-) Thanks for remind us ;-)

PR #1217, opened 2023-07-25
Show the exact location of function call

ChenfengWei0:
- issue: #1155 

add an option `show-caller-loc`.

```sh
// Previous
Violated property:
  file io.c line 79 function gets
VERIFICATION FAILED


// Now (--show-caller-loc)
Violated property:
  file io.c line 79 column 3 function gets
  Calling at: file test.c line 8 column 3 function main
VERIFICATION FAILED
```

This PR extends the issue to the more general case where the function caller and body definition are not in the same file. E.g.

- `test.h`
```c++
#include <stdio.h>
#include <string.h>
#include <assert.h>

void tmp(int x)
{
    assert(x == 0);
    assert(x + 1 == 1);
}
```

- `test.c`
```c++
1 #include "test.h"
2 
3 int main(int argc, char **argv)
4  {
5     char name[15];
6     gets(name);
7     tmp(1);
8     assert(0);
9     tmp(0);
10 }
```

- command: `$esbmc test.c --k-induction --show-caller-loc --multi-property`
- result

```sh
Violated property:
  file io.c line 79 column 3 function gets
  Calling at: file test.c line 6 column 5 function main

Violated property:
  file test.h line 7 column 5 function tmp
  Calling at: file test.c line 7 column 5 function main
  assertion x == 0
  x == 0

Violated property:
  file test.h line 8 column 5 function tmp
  Calling at: file test.c line 7 column 5 function main
  assertion x + 1 == 1
  x + 1 == 1

Violated property:
  file test.c line 8 column 5 function main
  assertion 0
  0
```


lucasccordeiro (approved):

fbrausse:
As this changes the format of the printed `Violated property:` messages I have to ask: does any downstream tool we care about depend on the specifics of this format? Is the esbmc-wrapper.py affected by these changes?

ChenfengWei0:
> As this changes the format of the printed `Violated property:` messages I have to ask: does any downstream tool we care about depend on the specifics of this format? Is the esbmc-wrapper.py affected by these changes?

Thanks for the suggestions. Will check.

lucasccordeiro:
I fully agree with @fbrausse's comments.

mikhailramalho:
@ChenfengWei0 we already have a stack trace, can you try to use that instead? See commit 5275157467c1822f60f33c6236081d93e344e4b4 for when we removed it from the cex.

ChenfengWei0:
> As this changes the format of the printed `Violated property:` messages I have to ask: does any downstream tool we care about depend on the specifics of this format? Is the esbmc-wrapper.py affected by these changes?

I have checked the esbmc-wrapper.py and it seems the extra printed information won't affect the results. Besides, we can always disable the (--show-caller-loc) option if it does affect any downstream tools.

ChenfengWei0:
> @ChenfengWei0 we already have a stack trace, can you try to use that instead? See commit [5275157](https://github.com/esbmc/esbmc/commit/5275157467c1822f60f33c6236081d93e344e4b4) for when we removed it from the cex.

Thank you very much for the information, didn't know it could be done this way.

Here is a comparison
```sh
Violated property:
  file test.h line 7 column 5 function tmp
  Calling at: file test.c line 13 column 5 function main  // this PR
c:@F@tmp at test.c line 13     // previous approach
c:@F@main at  line
  assertion x == 0
  x == 0
```
I think the old approach is good enough. I will replace my code with it.

ChenfengWei0:
Some result examples from the new approach :
```c++
Violated property:
  file test.h line 8 column 5 function tmp
Stack trace:
  c:@F@tmp at file test.c line 13 column 5 function main
  c:@F@main
  assertion x + 1 == 1
  x + 1 == 1

Violated property:
  file io.c line 79 column 3 function gets
Stack trace:
  c:@F@gets at file test.c line 12 column 5 function main
  c:@F@main

Violated property:
  file test.h line 7 column 5 function tmp
Stack trace:
  c:@F@tmp at file test.c line 5 column 5 function tmp2
  c:@F@tmp2 at file test.c line 10 column 5 function main
  c:@F@main
  assertion x == 0
  x == 0

Violated property:
  file test.c line 14 column 5 function main
Stack trace:
  c:@F@main
  assertion 0
  0
```

mikhailramalho (approved):
LGTM, but I want to suggest a follow-up patch: we should fix the naming of the function, i.e., `c:@F@gets` and `c:@F@main` are not valid c identifiers.

I think there are two possible ways to fix it:
1. Fix how the names are being stored in the function object in the stack_trace, or,
2. Use expr2c to get the correct name.

I think 2 should be easier, especially now that we are rewriting it.

ChenfengWei0:
> LGTM, but I want to suggest a follow-up patch: we should fix the naming of the function, i.e., `c:@F@gets` and `c:@F@main` are not valid c identifiers.
> 
> I think there are two possible ways to fix it:
> 
>     1. Fix how the names are being stored in the function object in the stack_trace, or,
> 
>     2. Use expr2c to get the correct name.
> 
> 
> I think 2 should be easier, especially now that we are rewriting it.

Just created issue #1233

lucasccordeiro (approved):

src/esbmc/options.cpp
```diff
@@ -92,6 +92,9 @@ const struct group_opt_templ all_cmd_options[] = {
     {"show-loops", NULL, "show the loops in the program"},
     {"show-claims", NULL, "only show claims"},
     {"show-vcc", NULL, "show the verification conditions"},
+    {"show-caller-loc",
+     NULL,
+     "show the location of the function calling expression"},
```
fbrausse: Should we rename this to a more universally known `show-stacktrace` and mention the counter-example in the description of the option?

ChenfengWei0: Changed as suggested.

fbrausse (approved):
Thanks, @ChenfengWei0!

lucasccordeiro:
@ChenfengWei0 : thanks for submitting this PR!

PR #1227, opened 2023-07-26
fp_conv: fix gte: not(less) is not equiv to gte

### LINKED ISSUE #1226: Incorrect floating point comparison with NaN in pre-conditions
Hi all,

ESBMC returns a counterexample for the code snippet below. It claims that the assert can be triggered if x=NaN. This is correct since NaN==NaN is false. However, x should never be equal to NaN, since the assume would return false too (NaN >= 0 is false)!

```
int main()
{
    double x = nondet_float();
	
	__ESBMC_assume(x >= 0);
	
	__ESBMC_assert(x == x, "");

    return 0;
}
```

I also tried modifying the condition in the assume to the macro `isgreaterequal(x, 0)` defined in `math.h` which guarantees compliance to the IEEE 754 standard. However, this does not fix the issue.

Any idea?

fbrausse:
Fixes #1226.


lucasccordeiro (approved):

lucasccordeiro:
@fbrausse: Thanks for submitting this PR!

PR #1228, opened 2023-07-26
Add benchmarks for CUDA

XLiZHI:
- [x] The verification time is optimized, and the lookup table is used in OM to replace the complex calculation. This saves about two thirds of the verification time.

- [x] The current benchmark score is 122/154, around 79% accuracy (ESBMC-GPU 85%), most of which failed because data races were not detected.

- [x] Plan to add some benchmarks with shorter validation times to TC. https://github.com/XLiZHI/esbmc/actions/runs/5710876892
The current TCs adds about 10 minutes to the test time.

This PR ignores some CUDA symbols in the data race check and is not sure whether to apply this change and needs to be discussed.


lucasccordeiro:
Thanks, @XLiZHI.

> The current benchmark score is 121/154, around 79% accuracy (ESBMC-GPU 85%), most of which failed because data races were not detected.

It would be great if you could inspect this data race check, i.e., try to understand why we now fail to detect data races.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -50,6 +51,43 @@ cudaError_t cudaMemcpy(T1 dst, T2 src, size_t count, enum cudaMemcpyKind kind)
   return __cudaMemcpy(dst, src, count, kind);
 }
 
+void *Address[10];
```
lucasccordeiro: Do we really need this address array?

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -50,6 +51,43 @@ cudaError_t cudaMemcpy(T1 dst, T2 src, size_t count, enum cudaMemcpyKind kind)
...
+
+cudaError_t cudaMalloc(void **devPtr, size_t size)
+{
+  cudaError_t tmp;
+  //pre-conditions
+  __ESBMC_assert(size > 0, "Size to be allocated may not be less than zero");
+  *devPtr = malloc(size);
+
+  Address[Counter] = *devPtr;
+  Counter++;
```
lucasccordeiro: I think this counter with the address array could be removed.

src/goto-programs/rw_set.cpp
```diff
@@ -49,6 +49,37 @@ void rw_sett::read_write_rec(
       {
         return; // ignore for now
       }
+
+      //improvements for CUDA features
```
lucasccordeiro: Can you add a TODO here? We should investigate in another PR why our data race check is broken.

lucasccordeiro (changes requested):

kunjsong01 (changes requested):
This is good work! The patch looks good to me. 

Just need a bit more information regarding those headers:
1. What's version of CUDA kernel are they based on?
2. What's the system path of the CUDA kernel they are based on? 
3. If an OM is directly copied and modified from CUDA system's header, would it be possible to add some information about the base version and briefly explain what has been modified.

src/goto-programs/rw_set.cpp
```diff
@@ -49,6 +49,37 @@ void rw_sett::read_write_rec(
...
+        symbol->name.as_string().find("__ESBMC") != std::string::npos ||
+        symbol->name.as_string().find("cuda") != std::string::npos ||
+        symbol->name.as_string().find("itoa") != std::string::npos ||
+        symbol->name.as_string().find("mem") != std::string::npos ||
+        symbol->name.as_string().find("rev") != std::string::npos ||
+        symbol->name.as_string().find("search") != std::string::npos ||
+        symbol->name.as_string().find("std") != std::string::npos ||
+        symbol->name.as_string().find("str") != std::string::npos ||
+        symbol->name.as_string().find("verify_kernel") != std::string::npos ||
+        symbol->name.as_string().find("virtual_table") != std::string::npos ||
```
fbrausse: Neither of these `.find()` conditions is a good idea: for instance, a substring like `str` or `std` or `mem` is way too common in regular code. If there is any way to use the `__ESBMC` **prefix** that would help in making these exceptions to read_write_rec() apply only to known symbols instead of matching arbitrary unknown ones.

If that's not possible, could you explain why? In any case we should find a way to make it possible or otherwise disambiguate the symbol names.

fbrausse (changes requested):

lucasccordeiro (approved):
LGTM.

kunjsong01 (approved):
LGTM

src/cpp/library/CUDA/math_functions.h
```diff
@@ -0,0 +1,75 @@
...
+
+  if(y == 0)
+    return result;
+
+  if(y < 0)
+    return 1 / __powf_gpu(x, -y);
+  else
+  {
+    for(int i = 0; i < y; ++i)
+      result *= x;
```
fbrausse, line 49: Even though [the docs](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__SINGLE.html#group__CUDA__MATH__INTRINSIC__SINGLE_1g2c2b295816185f6ce2423471df529974) say it's supposed to calculate an approximation, this implements the power with integer exponent, not a real / float exponent. You might be able to use `powf()` from our `<math.h>`. If you don't want to import all the definitions from it, you could just make a forward-declaration of powf.

src/cpp/library/CUDA/math_functions.h
```diff
@@ -0,0 +1,75 @@
...
+int __mul24(int x, int y);
+
+unsigned int __umul24(unsigned int x, unsigned int y);
+
+int __mul24(int x, int y)
+{
+  int cx;
+  int cy;
+  cx = x & 0x0000000000111111;
+  cy = y & 0x0000000000111111;
```
fbrausse, line 64: Selecting every 4th bit in x and y looks very strange to me. According to the docs, this function computes the least significant 32 bit of the multiplication of the least significant 24 bit of each operand. Same comment applies to `__umul24` below.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -50,6 +51,37 @@ cudaError_t cudaMemcpy(T1 dst, T2 src, size_t count, enum cudaMemcpyKind kind)
   return __cudaMemcpy(dst, src, count, kind);
 }
 
+cudaError_t cudaMalloc(void **devPtr, size_t size)
```
fbrausse, line 54: As this function is defined in a header file, it will lead to a multiple definitions error in case 2 different translation units include this file. You might want to either make it inline or instead of a header to define it in a .c file. Same applies to all non-inline non-static function definitions in headers.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -50,6 +51,37 @@ cudaError_t cudaMemcpy(T1 dst, T2 src, size_t count, enum cudaMemcpyKind kind)
...
+{
+  cudaError_t tmp;
+  //pre-conditions
+  __ESBMC_assert(size > 0, "Size to be allocated may not be less than zero");
+  *devPtr = malloc(size);
+
+  if(*devPtr == NULL)
+  {
+    tmp = CUDA_ERROR_OUT_OF_MEMORY;
+    exit(1);
```
fbrausse, line 64: According to [the docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g37d37965bfb4803b6d4e59ff26856356):
> cudaMalloc() returns cudaErrorMemoryAllocation in case of failure.

Is there a reason to call `exit(1)` here?

fbrausse (commented):
I have a few comments about the implementation of some API functions.

lucasccordeiro:
@XLiZHI: Thanks for submitting this PR! This is a great achievement for the ESBMC-GPU tool!

XLiZHI:
> I have a few comments about the implementation of some API functions.

Hi @fbrausse , Thank you very much for your review, these files are from the old ESBMC-GPU, it does need some changes to be correct, I will fix it in the next PR.

fbrausse:
@XLiZHI OK, thank you.

src/cpp/library/CUDA/cuda_runtime_api.h
lucasccordeiro: I will fix this issue in this PR: https://github.com/esbmc/esbmc/pull/1245/files

PR #1230, opened 2023-07-27
[Refactor] expr to string modules

shmarovfedor:
In this pull request I perform some refactoring for the methods that are dealing with producing string representations for `exprt` and `typet`:
- Classes `expr2ct` and `expr2cppt` are renamed to `c_expr2stringt` and `cpp_expr2stringt` respectively to avoid confusion, as they do not actually produce string representations for expressions and types in valid C and C++ syntax (but rather something that is C/C++-like but not quite valid)
- Files with the `c_expr2stringt` (former `expr2ct`) and `cpp_expr2stringt` (former `expr2cppt`) classes were moved to the `util` directory from `clang-c-frontend` and `clang-cpp-frontend` since they work on `exprt` classes regardless of which frontend is used (old C, clang C or clang C++). Perhaps, the `util` directory can be further refactored in the future as it looks slightly less organised then the other directories in the project.
- Class `expr2ccodet` has been renamed `expr2ct` since that's what it does: creating string representations of `exprt` and `typet` in valid C syntax. The files containing its implementation have also been moved from `clang-c-frontend` to `goto2c` to reflect this fact.


rafaelsamenezes (approved):

lucasccordeiro (approved):

fbrausse (commented):
I'm generally happy with this PR, just a small nitpick: It seems inconsistent naming to me to have `expr2string` and `expr2cpp` (in util), where the latter special-cases the former. C++ is not a special case of "string". ;-) How about something like "expr2human-c" and "expr2human-cpp"? Or "cex" instead of "human" if applicability to counter-example is more appropriate.

shmarovfedor:
Thanks @fbrausse I completely agree. I've already thought about this before submitting this PR. So I might as well address it here. Other candidates for the new names for `expr2string` and `expr2cpp` are `c_expr2string` and `cpp_expr2string` to reflect the separation between the languages in `exprt` while keeping a neutral suffix `string` since these methods are used for all sorts of outputs. What do you think?

fbrausse:
> Other candidates for the new names for `expr2string` and `expr2cpp` are `c_expr2string` and `cpp_expr2string` to reflect the separation between the languages in `exprt` while keeping a neutral suffix `string` since these methods are used for all sorts of outputs. What do you think?

Those sound good and better reflect the purpose, I fully agree.

shmarovfedor:
Done

fbrausse (approved):

lucasccordeiro:
@shmarovfedor : thanks for submitting this PR!

PR #1231, opened 2023-07-27
Remove some migrations

rafaelsamenezes:
As we discussed in https://github.com/esbmc/esbmc/issues/888 there are a couple of steps that are needed to get rid of irep1.

In this PR I will be focusing in:

- trivial cases that migrate back is not needed and is easily replace
- remove some irep functions that aren't used.

Usages of irep1 left: namespacet (symbolt is still the default), serialization, and string representations (all goto2c and outputs). These will be done in other PRs.

### DIFF
--- a/src/util/namespace.h
+++ b/src/util/namespace.h
@@ -10,7 +10,6 @@ class namespacet
 public:
   virtual const symbolt *lookup(const irep_idt &name) const;
 
-
   const typet &follow(const typet &src) const;
   const type2tc follow(const type2tc &src) const
   {
--- a/src/util/namespace.h
+++ b/src/util/namespace.h
@@ -9,13 +9,7 @@ class namespacet
 {
 public:
   virtual const symbolt *lookup(const irep_idt &name) const;
-  virtual const symbolt *lookup(const expr2tc &name) const;
-  const symbolt *lookup(const irept &irep) const
-  {
-    return lookup(irep.identifier());
-  }
 
-  void follow_symbol(irept &irep) const;
 
   const typet &follow(const typet &src) const;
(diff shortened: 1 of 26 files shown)


lucasccordeiro (approved):

src/util/namespace.h
```diff
@@ -9,12 +9,6 @@ class namespacet
 {
 public:
   virtual const symbolt *lookup(const irep_idt &name) const;
-  const symbolt *lookup(const irept &irep) const
-  {
-    return lookup(irep.identifier());
-  }
-
-  void follow_symbol(irept &irep) const;
```
fbrausse, line 17: I don't understand the purpose of removing these 2 API functions here. The `lookup(irept)` one is definitely useful judging by the amount of `.identifier()` suffixes added throughout this PR. And could you explain why you're moving `follow_symbol()` out of namespacet?

rafaelsamenezes: In summary, I don't want people to use it.

- follow symbol is only used by the old frontend
- by removing the lookup for irept it will be easier to my next PR which is replacing the symbol with symbolt2t in namespacet.

mikhailramalho:

This PR has been open since 2023-07-28 with no activity. The underlying goal (eliminating remaining irep1 usages and migrate_back call sites) is still live and is being pursued differently in @mikhailramalho's `feat/irep2-flatten` branch — that work supersedes this PR's approach.

Issue #888 ("Make migration optional") which this PR referenced is closed. Filing a fresh issue tracking the remaining migrate_back call sites might be a better way to coordinate finishing this off, if helpful.

Issue #818, opened 2022-07-19
Update math functions


mikhailramalho:
I've found this repo that implements correctly rounded fp operations. Maybe we can port them to esbmc: https://gitlab.inria.fr/core-math/core-math/-/tree/master/

fbrausse:
Paul gave a presentation about it last fall: https://members.loria.fr/PZimmermann/talks/core-math-arith2022.pdf

lucasccordeiro:
@brcfarias: can I ask you to take care of this issue?

fbrausse:
Based on `frexp`, I played around a bit with an implementation of log(), log2() and exp(). log2() implemented like this (exp is similar):
```c
double log2(double x)
{
	switch (fpclassify(x)) {
	case FP_NAN:
		return x;
	case FP_INFINITE:
		return signbit(x) ? NAN : x;
	case FP_ZERO:
		return -HUGE_VAL;
	case FP_SUBNORMAL:
	case FP_NORMAL:
		if (signbit(x))
			return NAN;
		break;
	}

	int xe;
	double xm = frexp(x, &xe); /* xm in [0.5, 1) */
	if (xm < 2.0 / 3.0) {
		xm *= 2;
		xe--;
	}
	int n = 1;                 /* xm in [0.666..., 1.333...) */
	xm = sqrt(xm); n *= 2;     /* xm in [0.816..., 1.154...) */
	xm = sqrt(xm); n *= 2;     /* xm in [0.903..., 1.074...) */
	xm = sqrt(xm); n *= 2;     /* xm in [0.950..., 1.036...) */
	xm = sqrt(xm); n *= 2;     /* xm in [0.974..., 1.018...) */
	int xe2 = xe;
	return xe2 + n * log1p_taylor(xm - 1) / M_LN2;
}
```
log1p_taylor(x) basically uses an N-th order Taylor approximation of log(x+1) at x=0.

Here are some local running times for successfully verifying
```c
	double x = nondet_double();
	__ESBMC_assume(0 < x && isfinite(x));
	int xe;
	frexp(x, &xe);
	double l = log2(x);
	double d = fabs(l - xe);
	assert(d <= 1);
```
- 4x sqrt, Taylor order 8: 745.220s
- 3x sqrt, Taylor order 8: 566.800s
- 3x sqrt, Taylor order 7: 336.039s
- 3x sqrt, Taylor order 6: 281.983s
- 2x sqrt, Taylor order 6: 202.370s
[...]

Issue #183, opened 2020-05-13
Improve our libc and libm models


mikhailramalho:
A long time ago I started implementing the model of our libc and libm but I never finished it.

Today I found the [newlib](https://sourceware.org/newlib/), a BSD-licensed implementation of libc and libm, which we could use as a reference to implement our models.

For instance, I found a bug in our implementation of fmodf by using the fmodf implementation:
```
int main()
{
  float f1 = nondet_float();
  __VERIFIER_assume(!isnanf(f1));
  __VERIFIER_assume(!isinff(f1));

  float f2 = nondet_float();
  __VERIFIER_assume(!isnanf(f2));
  __VERIFIER_assume(!isinff(f2));

  float f3 = fmodf(f1, f2);
  __VERIFIER_assume(!isnanf(f3));

  float f4 = __ieee754_fmodf(f1, f2);
  __VERIFIER_assume(!isnanf(f4));  

  assert(f3 == f4);
}
```
ESBMC founds the following cex:
```
  float f1 = -30934646784.0f;
  float f2 = 2.272718908704677750869278130857839909411942662560701666762952299905009567737579345703125E-29f;
```
Which shows that our model of fmodf returns the wrong result, compared to the compiled code.

By simply copying and pasting their fmodf model (from newlib 3.2.0) to ESBMC, it started to correctly verify the program. Their fmodf model:
```
typedef union {
  double value;
  struct
  {
    __uint32_t lsw;
    __uint32_t msw;
  } parts;
} ieee_double_shape_type;
typedef union {
  float value;
  __uint32_t word;
} ieee_float_shape_type;

static const float one = 1.0, Zero[] = {
                                0.0,
                                -0.0,
};

float fmodf(float x, float y)
{
  __int32_t n, hx, hy, hz, ix, iy, sx, i;

  do
  {
    ieee_float_shape_type gf_u;
    gf_u.value = (x);
    (hx) = gf_u.word;
  } while(0);
  do
  {
    ieee_float_shape_type gf_u;
    gf_u.value = (y);
    (hy) = gf_u.word;
  } while(0);
  sx = hx & 0x80000000;
  hx ^= sx;
  hy &= 0x7fffffff;

  if(((hy) == 0) || !((hx) < 0x7f800000L) || ((hy) > 0x7f800000L))
    return (x * y) / (x * y);
  if(hx < hy)
    return x;
  if(hx == hy)
    return Zero[(__uint32_t)sx >> 31];

[...]

PR #1232, opened 2023-07-28
[libm] Replaced existing operational models with MUSL library

emanino:
The existing libm operation models were incomplete and, in at least two cases, incorrect. First, acos(x) was wrongly implemented as 1/cos(x) instead of the inverse of the cosine function. Second, pow(x, y) assumed that y was the floating-point representation of an integer number, hence pow(2.1, 3.4) would actually return pow(2.1, 3.0). Third, many functions used imprecise, iterative implementations (e.g. cos(x), sqrt(x)) which were slow to verify and not bit-precise. Fourth, many important functions in the families of exponentials, logarithms, hyperbolic, trigonometric, Bessel, Gaussian and Gamma were not implemented.

This pull request provides operational models for all libm functions in three variants: float, double and long double. The operational models are ported from the MUSL library and written in plain C. In particular, we chose the MUSL library over Newlib and other similar libraries because its implementation guarantees correct roundings and contains as few loops as possible. Finally, the MUSL library is released under the MIT license.

Related: #818, #183


fbrausse:
> In particular, we chose the MUSL library over Newlib and other similar libraries because its implementation guarantees correct roundings

Hi @emanino, thank you for this PR! I wonder whether you could point me to any additional resources that support this claim. In particular, do you know of verification efforts or correctness proofs regarding musl's libm?

emanino:
> > In particular, we chose the MUSL library over Newlib and other similar libraries because its implementation guarantees correct roundings
> 
> Hi @emanino, thank you for this PR! I wonder whether you could point me to any additional resources that support this claim. In particular, do you know of verification efforts or correctness proofs regarding musl's libm?

Hi Franz, thanks for the question. I do not think there has been verification efforts on MUSL. My claims comes from the following sources.

One of the authors of MUSL found bugs in GCC while working on a bit-precise implementation of sqrt(x). See:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52593

MUSL was designed to be compliant to the standard, and a lot of effort was put towards such goal. See:
https://wiki.musl-libc.org/mathematical-library.html

A comparison table between different libraries can be found at the following link. See "features comparison" and "C99 math library":
https://www.etalabs.net/compare_libcs.html

As far as I know, there is currently an academic effort towards correct rounding by the CORE-MATH project. Unfortunately their library is incomplete and contains some bugs (I had a wrong result using their exp(x) function). It may be worth keeping an eye on their progress in the near future. Still, I do not know if they verify or prove the correctness of their implementation. See:
https://core-math.gitlabpages.inria.fr/

fbrausse:
I know Core-math and Paul Zimmermann, he's one of the main authors of MPFR. Many thanks for the additional pointers @emanino!

mikhailramalho (changes requested):
Thanks for the PR @emanino, however, I have a few concerns that must be addressed before this PR can be approved:
1. ESBMC encodes some operations internally using the SMT API (from the top of my head, I remember sqrt and fma, but there might be others). Calling these functions directly should be faster than the code implemented here. Is there a reason to replace them, e.g., performance, correctness?
2. Some of the iterative algorithms were only used in fixedbv arithmetic (via the __ESBMC_FIXEDBV ifdef), does the new algorithms you've added still work with fixedbvs?
3. We need test cases for all new models.

rafaelsamenezes:
- Reachability run (this PR): https://github.com/esbmc/esbmc/actions/runs/5691144790
- Reachability run (from last week): https://github.com/esbmc/esbmc/actions/runs/5643389166

The PR is fixing some KNOWNBUGs:
- esbmc/math_exp01
- esbmc/math_exp02
- esbmc/math_exp03

@emanino could you replace the KNOWNBUG to CORE (inside test.desc)?

However, the build time is taking too long. Even the action job is being cancelled automatically for taking too long.

My guess is that c2goto (and esbmc) have the problem of having to parse everything sequentially. @emanino How hard would be to mix the files into a few big C file?

fbrausse:
> How hard would be to mix the files into a few big C file?

If that's the problem, we could just add one libm.c file which #includes all the other .c files.

rafaelsamenezes:
> If that's the problem, we could just add one libm.c file which #includes all the other .c files.

True, I will do a PR to test this out.

emanino:
> Thanks for the PR @emanino, however, I have a few concerns that must be addressed before this PR can be approved:
> 
>     1. ESBMC encodes some operations internally using the SMT API (from the top of my head, I remember sqrt and fma, but there might be others). Calling these functions directly should be faster than the code implemented here. Is there a reason to replace them, e.g., performance, correctness?
> 
>     2. Some of the iterative algorithms were only used in fixedbv arithmetic (via the __ESBMC_FIXEDBV ifdef), does the new algorithms you've added still work with fixedbvs?
> 
>     3. We need test cases for all new models.

Hi Mikhail, that is great feedback! I have some further questions though.

1. Is there any documentation on which operations are encoded internally? Even just a list would be great. Also, how does ESBMC decide between using the internal operation vs the external operational model? How is the conflict resolved?
2. Similarly, is there any documentation on fixedbv arithmetic? I am guessing it stands for "fixed bit vectors", but I have never encountered the term before. Is it meant to model some specific IEEE 754 behaviour? The MUSL library assumes 32-bit floats and 64-bit doubles. Long doubles are either 64-bit, 80-bit or 128-bit. See the documentation at https://wiki.musl-libc.org/mathematical-library.html
3. Until now, I trusted the code in the MUSL library, since it is part of several Linux distributions. Are there tests for the old operational models in ESBMC I can draw inspiration from?

Many thanks in advance.

mikhailramalho:
> 1. Is there any documentation on which operations are encoded internally? Even just a list would be great. Also, how does ESBMC decide between using the internal operation vs the external operational model? How is the conflict resolved?

I don't think so... there are a few places you can look for them: the irep2 types, the do_special_functions in the frontend, and the fp encoding on backend.

> 2. Similarly, is there any documentation on fixedbv arithmetic? I am guessing it stands for "fixed bit vectors", but I have never encountered the term before. Is it meant to model some specific IEEE 754 behaviour? The MUSL library assumes 32-bit floats and 64-bit doubles. Long doubles are either 64-bit, 80-bit or 128-bit. See the documentation at https://wiki.musl-libc.org/mathematical-library.html

fixedbv is another way of encoding real numbers that brings some advantages/disadvantages compared to IEEE floating points. You can find more information about them in Lucas' and mine thesis. 

> 3. Until now, I trusted the code in the MUSL library, since it is part of several Linux distributions. Are there tests for the old operational models in ESBMC I can draw inspiration from?

Yes, you can find them in regressions/floats and regressions/floats-regression. Basically, when I wrote them you need to work out a test case with constants and another with nondet values, that way we can check if the simplification is not doing anything wrong with the models.

Also, remember that it's not only about trusting the correctness of the code but also if esbmc can handle the code you're writing efficiently. We are not executing code, but encoding it in SMT, so code that would be faster in one case not necessarily be the best approach for the other. That's also why we always run the PRs on SV-Comp and check for regressions there too (wrong results, timeouts, etc).


> 
> Many thanks in advance.

lucasccordeiro:
This PR is no longer valid.

PR #1234, opened 2023-07-28
Make Interval Analysis reset state after pointer operation

rafaelsamenezes:
Currently, the `--interval-analysis` is unsound due to ignoring pointers. This means that verifying something like...

```c
int main()
{
 int a,*q;
 a = 1;
 q = &a;
 *q = 2;
 __ESBMC_assert(a != 2, "");
}
```

... will result in an incorrect true. 

In this PR, we will reset the abstract state after any assignment over a dereference. As this is now "sound", we can enable it globally (as long as it doesn't have any concurrency). 

In future, we will combine the IA with a points-to analysis to only reset/update the ones that are needed. For now this is needed for the interval-simplify option.


rafaelsamenezes:
Results 30s:

   correct:           14343
     correct true:     8143
     correct false:    6200
   incorrect:            60
     incorrect true:     30
    incorrect false:    30
   unknown:           12056
  Score:             21046 (max: 43043)

The diff mainly comes from the Reachability category that now has weaker proofs for kind.

lucasccordeiro:
@rafaelsamenezes: do we have the latest results for the master branch?

rafaelsamenezes:
Master: 
```
2023-07-30T05:19:34.0617017Z Statistics:          26459 Files
2023-07-30T05:19:34.0624062Z   correct:           14370
2023-07-30T05:19:34.0624811Z     correct true:     8167
2023-07-30T05:19:34.0625540Z     correct false:    6203
2023-07-30T05:19:34.0626174Z   incorrect:            59
2023-07-30T05:19:34.0626910Z     incorrect true:     29
2023-07-30T05:19:34.0627671Z     incorrect false:    30
2023-07-30T05:19:34.0632711Z   unknown:           12030
2023-07-30T05:19:34.0654691Z   Score:             21129 (max: 43043)
```

lucasccordeiro:
Thanks, @rafaelsamenezes.

Does the additional incorrect result in your Par also happen in the master if you increase the timeout?

rafaelsamenezes:
Apparently, this patch trigger an issue during k-induction conversion. I will just let this patch work for reachability as it gives me more reasonable results.

Master:

```
Correct True: 2748
Correct False: 1634
Incorrect True: 29
Incorrect False: 3
```

This PR:

```
Correct True: 2734
Correct False: 1633
Incorrect True: 29
Incorrect False: 3
```

From the diff:
- 3 unknowns, due to the kind bug.
- 15 timeouts, all of them are from the ldv-linux benchmarks
- the remaining benchmarks seems to be noise.

lucasccordeiro (changes requested):
Can you also add a regression test that succeeds?

lucasccordeiro (approved):

lucasccordeiro:
@rafaelsamenezes: After we merge this PR, as discussed, we should try to enable the k-induction for the memory safety category of SV-COMP.

Issue #447, opened 2021-05-24
Inconsistent counterexample values with Z3


rafaelsamenezes:
The program with `--function loop_init --k-induction`:

```c
#define UINT8  unsigned __int8
unsigned arr[100];
int tt;

int foo() {
	int x;
    tt = nondet_int();
	tt += 1;
	int y = 25;
	int res = tt + y;

	return res;
}

int loop_init() {
	//__ESBMC_init_var(&arr);
	int a = foo();
	while (a > 1)
		a = a / 2;
	__ESBMC_assert(a == 2, "Failed as expected");

	return 0;
}
```

Gives the following result with Z3:

```c
State 1 file test.c line 7 function foo thread 0
----------------------------------------------------
  tt = -22 (11111111 11111111 11111111 11101010)

State 2 file test.c line 8 function foo thread 0
----------------------------------------------------
  tt = -21 (11111111 11111111 11111111 11101011)

State 3 file test.c line 10 function foo thread 0
----------------------------------------------------
  res = 4 (00000000 00000000 00000000 00000100)

State 4 file test.c line 17 function loop_init thread 0
----------------------------------------------------
  a = -2147483644 (10000000 00000000 00000000 00000100)
```

`a` should have received the same value as `res`. 

This bug doesn't happen when using boolector

PR #1237, opened 2023-07-30
Bring smtlib backend on par with other backends

fbrausse:
Besides some small fixes in the smtlib backend, this PR also removes uninterpreted `--` delimiters(?) used in some old test.desc files in the regression suite.

Local results with Yices:

99% tests passed, 13 tests failed out of 3560
<details>
Total Test time (real) = 1246.46 sec

The following tests FAILED:
	104 - regression/esbmc-unix/00_race13 (Timeout)
	232 - regression/esbmc-unix/github_538_weaver_true (Timeout)
	292 - regression/esbmc-unix2/01_pthread59 (Timeout)
	294 - regression/esbmc-unix2/01_pthread60 (Timeout)
	295 - regression/esbmc-unix2/01_pthread61 (Timeout)
	300 - regression/esbmc-unix2/01_pthread66 (Failed)
	305 - regression/esbmc-unix2/01_pthread70 (Failed)
	307 - regression/esbmc-unix2/01_pthread72 (Failed)
	708 - regression/esbmc/github_447-1 (Failed)
	846 - regression/esbmc/multi_dimensional_array_1-success (Timeout)
	1717 - regression/nonz3/overflow_12_adpcm (Failed)
	1731 - regression/bitwuzla/memset-good (Failed)
	1767 - regression/incremental-smt/incremental-26 (Timeout)
</details>

Some of the failed ones aren't bugs in the smtlib backend but Yices returning `???` for a `(get-value (...))` query. Not sure for which kinds of expressions exactly this happens. *Edit:* I've now handled those in a more graceful way than aborting.

Of the remaining 6 incorrect non-timeout ones, 1731 is #1171, 4 others are known bugs for not just the smtlib backend: 300, 305 and 307 are #1197 and 1717 is #1238. That leaves 708, which produces the same verdict but has a too restrictive regex in my opinion. If true, then with this PR the smtlib backend would be on par with the other backends.


lucasccordeiro (approved):

fbrausse:
> That leaves 708, which produces the same verdict but has a too restrictive regex in my opinion. If true, then with this PR the smtlib backend would be on par with the other backends.

Indeed, the regex checks whether some RHS in the counter-example matches another one, which doesn't necessarily have to be the case. Since its a reduced example of #447 which I think has a correct regex checking the same condition on different code, instead of opening an issue to find a matching regex I've relaxed the test performed here.

lucasccordeiro:
> Indeed, the regex checks whether some RHS in the counter-example matches another one, which doesn't necessarily have to be the case. Since its a reduced example of https://github.com/esbmc/esbmc/issues/447 which I think has a correct regex checking the same condition on different code, instead of opening an issue to find a matching regex I've relaxed the test performed here.

That's fine by me.

lucasccordeiro:
@fbrausse: thanks for submitting this PR!

PR #1241, opened 2023-08-01
Fix overflows

### LINKED ISSUE #1238: Overflow-check regression with some backends
This issue concerns `regression/nonz3/overflow_12_adpcm`: it runs with `--overflow-check`, the expected verdict is SUCCESSFUL but it seems all backends besides --boolector and --bitwuzla find an overflow. For instance with --yices:
```
[Counterexample]


State 134 file main.c line 209 column 3 function decode thread 0
Violated property:
  file main.c line 209 column 3 function decode
  arithmetic overflow on mul
  !overflow("*", (signed long int)dec_detl, (signed long int)(qq6_code6_table[(signed long int)il]))


VERIFICATION FAILED
```
Same happens with the CVC4, Z3 (though for some reason this regression test is under nonz3/) and smtlib backends.

These counter-examples look a bit suspicious to me, because the two quantities multiplied here are "int"s and the test fails on a system where "signed long int" has 64 bit.

fbrausse:
This PR fixes issues related to overflows, both checking them and actual overflows in the backends:
- smt_overflow: `1ULL << n` where `n` has no bounds is never a good idea
- z3: creating BVs of widths > 64
- cvc, yices: creating BVs from constants of widths > 63

Fixes #1238 for all backends. I've added a reduced regression test for that issue.

### DIFF
--- a/regression/esbmc/overflow-long/ovfl3.c
+++ b/regression/esbmc/overflow-long/ovfl3.c
@@ -1,0 +1,7 @@
+
+int main()
+{
+int a = -136;
+int v = 32;
+int dl = (long)v * (long)a;
+}
--- a/regression/esbmc/overflow-long/test.desc
+++ b/regression/esbmc/overflow-long/test.desc
@@ -1,0 +1,4 @@
+CORE
+ovfl3.c
+--overflow-check
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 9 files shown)


lucasccordeiro (approved):
That's great, @fbrausse.

Issue #1213, opened 2023-07-21
[Concurrency]Simplifer in symex_goto produces different result


Anthonysdu:
Here is a safe program:
```
#include <assert.h>
#include <pthread.h> 

int X = 0;
int Y = 0;

void* thread() {
  if(Y)
    Y = 1;
}

int main() {
  pthread_t t1;
  pthread_create(&t1, NULL, thread, NULL);
  X = 1;
  assert(X==1);
  return 0;
}
```
If we run `esbmc main.c --ssa-trace --no-slice`:
There is a wrong ssa in the interleaving:
```
Thread 0 file main.c line 15 function main
ASSIGNMENT ()
X&32#2 == 1

Thread 1 file main.c line 9 function thread
ASSUME
1

Thread 1 file main.c line 10 function thread
ASSIGNMENT (HIDDEN)
X&34#3 == 0
```
As `X` shouldn't be merged after the context-switch. #1125 

If run `esbmc main.c --ssa-trace --no-slice --no-simplify --unwind 1`, the result is correct, and also it can be seen in this interleaving, `Y` is the variable to be merged.

```
Thread 0 file main.c line 15 function main
ASSIGNMENT ()
X&50#2 == 1

Thread 1 file main.c line 9 function thread
ASSUME
1

Thread 1 file main.c line 9 function thread
ASSIGNMENT ()
Y&52#2 == 1
```
As I tested, these two ` do_simplify()` call in `symex_goto.cpp` make the result go wrong:
https://github.com/esbmc/esbmc/blob/4640d2ec1d68849dc295d39c9d6c269f2ada1344/src/goto-symex/symex_goto.cpp#L21C2-L21C3
https://github.com/esbmc/esbmc/blob/4640d2ec1d68849dc295d39c9d6c269f2ada1344/src/goto-symex/symex_goto.cpp#L174C10-L174C10

lucasccordeiro:
@Anthonysdu: Just so you know. If you use the incremental solving option for guards, ESBMC can verify this program correctly:

$ esbmc thread.c --no-por  --smt-during-symex  --smt-thread-guard --no-slice

````
Thread interleavings 19
Symex completed in: 0.014s (46 assignments)
Slicing time: 0.000s (removed 0 assignments)
Generated 9 VCC(s), 0 remaining after simplification (46 assignments)
BMC program time: 0.014s

VERIFICATION SUCCESSFUL
````

lucasccordeiro:
[...]

PR #1242, opened 2023-08-02
[Concurrency]Guard fix

Anthonysdu:
There is a bug in executing guards, and can be exposed by `--no-por` in this regression test.

In the function` is_cur_state_guard_false()` in `src/goto-symex/execution_state.cpp`, when not using `smt_thread_guard`, it should return if the guard is false, but not the false value. 

This should be the root cause of this issue #1213

### DIFF
--- a/regression/esbmc-unix2/11_burckhardt-fig2/main.c
+++ b/regression/esbmc-unix2/11_burckhardt-fig2/main.c
@@ -1,0 +1,19 @@
+#include <assert.h>
+#include <pthread.h> 
+
+int X = 0;
+int Y = 0;
+
+void* thread() {
+  if(Y)
+    Y = 1;
+}
+
+int main() {
+  pthread_t t1;
+  pthread_create(&t1, NULL, thread, NULL);
+  X = 1;
+  assert(X==1);
+  return 0;
+}
+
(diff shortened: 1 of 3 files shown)


mikhailramalho (approved):
Interesting! This is good to go if there are no regressions on sv-comp.

lucasccordeiro (approved):

lucasccordeiro:
@Anthonysdu: Thanks for submitting this PR!

mikhailramalho:
@lucasccordeiro I suggest we first run this on the sv-comp benchmarks, are we certain that this won't cause regressions?

Anthonysdu:
I am adding a benchexec run:
https://github.com/Anthonysdu/esbmc/actions/runs/5740235721

lucasccordeiro:
I agree; I have overlooked @mikhailramalho's comment.

Let's have an SV-COMP run for this PR.

Anthonysdu:
[Results](https://github.com/esbmc/esbmc/actions/runs/5770441232/job/15643599765) in 30s:

```
  correct:             374
    correct true:      107
    correct false:     267
  incorrect:             3
    incorrect true:      3
    incorrect false:     0
  unknown:             312
  Score:               385 (max: 1036)

```

PR #1245, opened 2023-08-02
[C++ OM] fixed address of stack memory associated with local variable

### LINKED ISSUE #1243: [C++ OM]  address of stack memory associated with local variable 'tmp' returned
We must fix this warning:

/tmp/esbmc-cpp-headers-6ff2-2230-a9c5/cstdlib:115:9: warning: address of stack memory associated with local variable 'tmp' returned [-Wreturn-stack-address]
        return tmp;

```C
char * getenv(const char * name) {
	size_t size = strlen(name) + 1;
	char tmp[size];
	strcpy(tmp, name);
	return tmp;
}
````

### PR DESCRIPTION
**Pull request by lucasccordeiro**
This PR should fix https://github.com/esbmc/esbmc/issues/1243.

### CODE REVIEW

**Review by kunjsong01 — APPROVED** [review]
LGTM

File: src/cpp/library/cstdlib [hunk 1]
```diff
@@ -108,11 +108,15 @@ long int strtol(const char * str, char ** endptr, int base);
...
-char * getenv(const char * name) {
-	size_t size = strlen(name) + 1;
-	char tmp[size];
-	strcpy(tmp, name);
-	return tmp;
+char *getenv(const char *name) {
+  if(name == NULL || *name == '\0')
+    return NULL; // Invalid input
+  size_t size = strlen(name) + 1;
+  char *tmp = (char *)malloc(sizeof(char) * size);
```
rafaelsamenezes: Won't that trigger memleak checks? 

I guess the proper way is to use an infinite array or just use an alloca which doesn't add a dead STMT.

rafaelsamenezes (changes requested):

src/cpp/library/cstdlib
lucasccordeiro: Fair point: I replaced `malloc` with `__ESBMC_alloca`.

rafaelsamenezes (approved):

mikhailramalho:
alloca should also be freed at the end of the scope, so this will fail too.

rafaelsamenezes:
> alloca should also be freed at the end of the scope, so this will fail too.

It depends on how we are modeling ALLOCA:

1. If we are considering what compilers do (push into the stack), then it should become an invalid dereference not a memleak. AFAIK the function stack is going to be removed when it goes out of scope, resulting in a dangling pointer (am I correct?). Also, is this intrinsic even part of the standard?
2. If we are considering how `__ESBMC_alloca` is used in production... its just a way of allocating a piece of memory for free. This is often used to initialize some byte arrays or to fake registers addresses for the harness.

I don't think that the `__ESBMC_alloca` should trigger any memleak or dangling pointer check. It should be equivalent of having a DECL without DEAD.

lucasccordeiro:
Thanks, @rafaelsamenezes.

We have many places in our OM, where we use `malloc` without calling `free`. This might be the cause of why we don't check for memory leaks in SV-COMP since we return too many incorrect results.

I wonder whether we should replace `malloc` with `__ESBMC_alloca` whenever possible in our OMs.

fbrausse:
> > alloca should also be freed at the end of the scope, so this will fail too.
> 
> It depends on how we are modeling ALLOCA:

We model it similar to malloc(), but: there is no memory-leak check, and we're checking that its return value is not free'd.

> 1. If we are considering what compilers do (push into the stack), then it should become an invalid dereference not a memleak. AFAIK the function stack is going to be removed when it goes out of scope, resulting in a dangling pointer (am I correct?). Also, is this intrinsic even part of the standard?

You're right about the dangling pointer. It's not part of any standard I know of. It is documented for GCC, however, as a builtin: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005falloca See also the man-page about <alloca.h>: https://linux.die.net/man/3/alloca

> 2. If we are considering how `__ESBMC_alloca` is used in production... its just a way of allocating a piece of memory for free. This is often used to initialize some byte arrays or to fake registers addresses for the harness.
> 
> 
> I don't think that the `__ESBMC_alloca` should trigger any memleak or dangling pointer check. It should be equivalent of having a DECL without DEAD.

alloca'd allocations are skipped for the memleak-check. About dangling pointers I don't know right now, but I disagree: leaving the function's scope that called alloca() and still using its return value is a bug.

rafaelsamenezes:
> leaving the function's scope that called alloca() and still using its return value is a bug.

True, but that should be how we implement `alloca` not `__ESBMC_alloca`. If you think that this might lead to user confusion... we could go for another name ( `__ESBMC_static_alloca`  ? ).

kunjsong01 (approved):

fbrausse:
> > leaving the function's scope that called alloca() and still using its return value is a bug.
> 
> True, but that should be how we implement `alloca` not `__ESBMC_alloca`. If you think that this might lead to user confusion... we could go for another name ( `__ESBMC_static_alloca` ? ).

Oh, I see. Sorry, I missed that. Yes, there's a difference. At the moment we have `-D__ESBMC_alloca=__builtin_alloca`. I was just talking with Lucas about introducing a new ESBMC intrinsic, but we might also be able to reuse this one. You said:

> 2. If we are considering how `__ESBMC_alloca` is used in production... its just a way of allocating a piece of memory for free. This is often used to initialize some byte arrays or to fake registers addresses for the harness.

Given that we define it as __builtin_alloca which is assigned the `alloca()` semantics, to me it seems like `__ESBMC_alloca` is expected by our users to behave like `alloca()`. Is that right?

rafaelsamenezes:
> > > leaving the function's scope that called alloca() and still using its return value is a bug.
> > 
> > True, but that should be how we implement `alloca` not `__ESBMC_alloca`. If you think that this might lead to user confusion... we could go for another name ( `__ESBMC_static_alloca` ? ).
> 
> Oh, I see. Sorry, I missed that. Yes, there's a difference. At the moment we have `-D__ESBMC_alloca=__builtin_alloca`. I was just talking with Lucas about introducing a new ESBMC intrinsic, but we might also be able to reuse this one. You said:
> 
> > 2. If we are considering how `__ESBMC_alloca` is used in production... its just a way of allocating a piece of memory for free. This is often used to initialize some byte arrays or to fake registers addresses for the harness.
> 
> Given that we define it as __builtin_alloca which is assigned the `alloca()` semantics, to me it seems like `__ESBMC_alloca` is expected by our users to behave like `alloca()`. Is that right?

Not really. They expect an allocation that will "just work", I don't think we have any users that rely on proper stack alloca. I implemented as the builtin alloca because it was easier at the time.

mikhailramalho:
historically, alloca was implemented to work on the sv-comp benchmarks, so:
1. we don't check for memory allocation
2. decl_dead ignores objects allocated using alloca

__builtin_alloca should work in the same way.

If we want to return the object from get_env, either create a global or a local static object (although I'm sure the latter is available in C).

fbrausse (approved):
I am fine with fixing the C++ OM by delegating back to the C one.

Regarding the alloca() discussion I think we have to separate it from getenv() here because of what Lucas said:
> We have many places in our OM, where we use malloc without calling free. This might be the cause of why we don't check for memory leaks in SV-COMP since we return too many incorrect results.

If that's the case then we might need to separate `alloca()` (which by <alloca.h> is usually defined as `__builtin_alloca()`, by SV-COMP due to their preprocessed sources most certainly) from either a new intrinsic to replace the malloc() calls inside our OMs or we could instead use __ESBMC_alloca and give it a new definition that is more in line with what Rafael suggests users want.

lucasccordeiro:
I can see that @fbrausse's PR is already handling the stdlib issue.

Closed via https://github.com/esbmc/esbmc/pull/1248.

PR #1246, opened 2023-08-02
[C++ verification] Add support for list initializaitons

XLiZHI:
`esbmc main.cpp`

```
#include <cassert>

struct dim3
{
  unsigned int x, y, z;
};

dim3 s1 =  {1,2,3};

int main() 
{
  assert(s1.x == 1); //failed
  return 0;
}
```
This is because the struct is judged to be a POD, which skips initialization:
https://github.com/esbmc/esbmc/blob/5f1c964abff300466a10ee5b049e98044a0cc5c1/src/clang-c-frontend/clang_c_convert.cpp#L557-L564

We need to determine if the aggregation type is explicitly initialized.

Original symbol table:
`s1={ .x=0, .y=0, .z=0 };`

This PR:
`s1={ .x=1, .y=2, .z=3 };`


regression/esbmc-cpp/bug_fixes/1006_aggregate/main.cpp
```diff
@@ -6,12 +6,19 @@ struct uint3
   unsigned int x, y, z;
 };
 
-struct uint3 indexOfThread[1024];
+uint3 indexOfThread[1024];
+
+uint3 s1 = {1,1,1};
```
kunjsong01: Could you restore 1006_* TCs and move this to a new test case please?

regression/esbmc-cpp/bug_fixes/1006_aggregate_fail/main.cpp
```diff
@@ -6,12 +6,19 @@ struct uint3
   unsigned int x, y, z;
 };
 
-struct uint3 indexOfThread[1024];
+uint3 indexOfThread[1024];
+
+uint3 s1 = {1,1,1};
```
kunjsong01: same as above

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -553,14 +553,17 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
     if(r)
       return true;
 
+    bool aggregate_without_init = aggregate_value && val.id() == "sideeffect" &&
```
kunjsong01: Is there another way to check explicit initialisation in the clang AST `vd`?
We usually inspect clang AST to determine how we proceed with constructing the irept. 
Since `val` is our irept, relying on the characteristics of irept may not do good to generality.

kunjsong01 (changes requested):

src/clang-c-frontend/clang_c_convert.cpp
XLiZHI: Maybe we can modify it in this method? Do you have any thoughts? I don't understand exactly how this method is called.

https://github.com/esbmc/esbmc/blob/5f1c964abff300466a10ee5b049e98044a0cc5c1/src/clang-cpp-frontend/clang_cpp_main.cpp#L8-L16

kunjsong01: The problem here boils down to distinguishing zero initialization from value initialization - we should probably do it in the converter using the information from AST. Otherwise we won't be able to access such information in the adjustment step. 
Probably just need to refine the conditions, which should do the job.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -498,11 +498,11 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
   symbol.file_local = (vd.getStorageClass() == clang::SC_Static) ||
                       (!vd.isExternallyVisible() && !vd.hasGlobalStorage());
 
-  bool aggregate_value_init = is_aggregate_type(vd.getType());
+  bool aggregate_value = is_aggregate_type(vd.getType());
```
kunjsong01: Please restore variable name. `dim3 s1 =  {1,2,3};` is copy list initializaiton. For the best software engineering practce, we shouldn't touch any code for aggregate/POD in this patch.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -547,20 +547,26 @@ bool clang_c_convertert::get_var(const clang::VarDecl &vd, exprt &new_expr)
      * initializer can't have side-effects (it's a constant expression). */
     code_blockt *orig = current_block;
     current_block = nullptr;
+    const clang::Stmt *stmt = vd.getInit();
```
kunjsong01, line 550: It looks about right... but one more thing to consider: does it generalize to direct list initialization? 
Would it be possible for you to try the followings? and what's the clang AST like? 
```
dim3 s1 = {1, 2, 3}; // copy list initialization
dim3 s1 {1, 2, 3}; // direct list initialization
```

kunjsong01 (changes requested):
We need to change the title of this PR to "Add support for list initializaitons"

kunjsong01:
A second thought: 

For a quick fix, would it be possible to change the line in operational model to just use direct initialization`dim3 s1(1,2,3);`? so that our progress won't be blocked by it. 

I reckon our frontend has no problems supporting direct initializaiton.

XLiZHI:
> A second thought:
> 
> For a quick fix, would it be possible to change the line in operational model to just use direct initialization`dim3 s1(1,2,3);`? so that our progress won't be blocked by it.
> 
> I reckon our frontend has no problems supporting direct initializaiton.

Thank you very much for your advice.

src/clang-c-frontend/clang_c_convert.cpp
XLiZHI: Hi @kunjsong01 ,

Our compiler doesn't seem to support some C++ 11 features, and I get the following error message when I try to confirm the AST:
```
Parsing main2.cpp
main2.cpp:8:9: error: expected ';' after top level declarator
uint3 s1 {1, 2, 3};
        ^
        ;
ERROR: PARSING ERROR
```
I tested it in the latest version.

lucasccordeiro:
@kunjsong01 : can I ask you whether you still have comments for this PR?

@XLiZHI: can I ask you whether you have addressed all issues raised by @kunjsong01?

XLiZHI:
> @kunjsong01 : can I ask you whether you still have comments for this PR?
> 
> @XLiZHI: can I ask you whether you have addressed all issues raised by @kunjsong01?

Hi Dr @lucasccordeiro, there is one feature of C++ that may need to be discussed, and the rest I think are settled.

https://github.com/esbmc/esbmc/pull/1246#discussion_r1305871902

kunjsong01:
@lucasccordeiro Hi Lucas, I don't have any additional comments for this PR.

kunjsong01 (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #1248, opened 2023-08-02
Remove some C++ headers with definitions duplicated from C

### LINKED ISSUE #1243: [C++ OM]  address of stack memory associated with local variable 'tmp' returned

fbrausse:
This PR:
- removes from cpp/library:
  - <string.h>
  - <unistd.h>
  - <math.h>
  - <stdlib.h>
  - <stdio.h>
  - <float.h>
- replaces in cpp/library (mostly by delegating to the C headers, notable exception: \<cmath>):
  - \<cmath>
  - \<cctype>
  - \<cassert>
  - \<cstdarg>
  - \<cstdlib>
  - \<cstdio>
- and adds in c2goto/headers:
  - <__esbmc/stddefs.h>: some #defines that make life easier working with C++
  - <ctype.h>: a copy of the declarations from the removed \<cctype>
  - <stdlib.h>: a slightly modified copy from the removed \<cstdlib>
  - <string.h>

Related: #965


lucasccordeiro:
This is an excellent initiative, @fbrausse.

mikhailramalho (commented):
nice

fbrausse:
Since this PR is adding definitions potentially incompatible with whatever SV-COMP thinks is right, I'll run it over the benchmarks to be sure nothing breaks.

mikhailramalho:
Do we need to add the definitions?

fbrausse:
> Do we need to add the definitions?

For C++ yes: there are CUDA regression tests that do `using std::sin;` and that doesn't work if `sin` is not defined in `namespace std`.

For C I tend to also say yes though that's just about declarations and type definitions, because, for instance, `#include <stdlib.h>` should provide a declaration of `void *malloc(size_t);` and also a definition of the `div_t` struct in order to provide a declaration of `div_t div(int num, int den);`.

fbrausse:
@mikhailramalho Alternatively we could just choose to not provide these standard headers and let the system's library do that. I don't know what'll happen. Shall we try?

fbrausse:
> I don't know what'll happen.

Found out what will happen: we get c_link errors for types that were defined (by the system's libs) during c2goto. This happens when running ESBMC with a --sysroot (as I need to for CHERI). I'd rather like to keep our OM headers in sync with the models themselves in this PR.

lucasccordeiro:
I agree with you, @fbrausse.

Do you have the SV-COMP results for this PR?

fbrausse:
@lucasccordeiro They should be ready in ~2h: https://github.com/esbmc/esbmc/actions/runs/5744047334

fbrausse:
I think we should build & run the regression suite on MacOS as well since this is (partially) changing ESBMC's view on system libaries for C/C++. @rafaelsamenezes Do you know what's the status of the MacOS CI?

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/5699725875) (8bf4ef883c3cac2b49961bbc38382904ee9780ff):
```
  correct:           14365
    correct true:     8166
    correct false:    6199
  incorrect:            59
    incorrect true:     29
    incorrect false:    30
  unknown:           12035
  Score:             21123 (max: 43043)
```

[This PR](https://github.com/esbmc/esbmc/actions/runs/5744047334) (406a09073c9bc301390bea1671999e4295dbbc2c):
```
  correct:           14366
    correct true:     8139
    correct false:    6227
  incorrect:            59
    incorrect true:     29
    incorrect false:    30
  unknown:           12034
  Score:             21097 (max: 43043)
```
No changes to incorrect. Correct true did go down a bit but correct false did increase.

rafaelsamenezes:
> I think we should build & run the regression suite on MacOS as well since this is (partially) changing ESBMC's view on system libaries for C/C++. @rafaelsamenezes Do you know what's the status of the MacOS CI?

Not even started. To be honest, I am not sure whether we should care that much for macOS (afaik I am the only using it for development). 

Most of the "macOS" issues that we have nowadays are because of M1 (which we can't test). Is there any way for us to test ARM builds?

fbrausse:
2x timeout -> unknown:
- no_overflow c/array-lopstr16/break-1.i: mk_eq() on a btor_smt_ast and a tuple_node_smt_ast; already fails on master
- unreach_call c/loop-invgen/string_concat-noarr.i: master is timeout even for 300s. A debug build of master and also this PR already aborts at https://github.com/esbmc/esbmc/blob/5f1c964abff300466a10ee5b049e98044a0cc5c1/src/goto-programs/loop_unroll.cpp#L94 because the `t->guard` is
  ```
  constant_bool
  * value : true
  * type : bool
  ```

7x unknown -> correct false in valid-memsafety and 1x unknown -> timeout also in valid-memsafety.

kunjsong01 (approved):

lucasccordeiro (approved):

lucasccordeiro:
@fbrausse: thanks for submitting this PR!

PR #1260, opened 2023-08-06
[CUDA] Add CuDNN OM

XLiZHI:
The convolution operation is modeled to implement the same operation as CuDNN.

Please refer to the specific principle: `img2col` and `GEMM`
https://arxiv.org/pdf/1410.0759.pdf Page 4

An accelerated implementation of the activation function sigmoid as well as a derivative implementation using a hash table, while adding an implementation for ReLU and its derivative.


regression/cuda/benchmarks/114_cudnn_conv/main.cu
```diff
@@ -0,0 +1,41 @@
+#include <cudnn.h>
+#include <assert.h>
+
+int main()
+{
+    cudnnHandle_t handle;
```
lucasccordeiro: Can you please add some comments?

What is the purpose of this regression test? What are you testing?

regression/cuda/benchmarks/114_cudnn_conv/main.cu
```diff
@@ -0,0 +1,41 @@
...
+    float img[1 * 1 * 3 * 3] = {1, 2, 3, 4, 5, 6, 7, 8, 9};
+    float f[1 * 1 * 2 * 2] = {1, 2, 3, 4};
+    float out[1 * 1 * 2 * 2] = {};
+    float dev[1 * 1 * 2 * 2] = {37, 47, 67, 77};
+
+    cudnnConvolutionForward(handle, &alpha, xDesc, img, wDesc, f, convDesc, CUDNN_CONVOLUTION_FWD_ALGO_GEMM, 0, 0, &beta, yDesc, out, 1, 3, 2, 1, 0);
+
+    for (int i = 0; i < 4; i++)
+    {
+        assert(out[i] == dev[i]);
```
lucasccordeiro: What is the meaning of this assertion?

regression/cuda/benchmarks/114_cudnn_conv/main.cu
```diff
@@ -0,0 +1,41 @@
...
+    float alpha = 1.0f, beta = 0.0f;
+    float img[1 * 1 * 3 * 3] = {1, 2, 3, 4, 5, 6, 7, 8, 9};
+    float f[1 * 1 * 2 * 2] = {1, 2, 3, 4};
+    float out[1 * 1 * 2 * 2] = {};
+    float dev[1 * 1 * 2 * 2] = {37, 47, 67, 77};
+
+    cudnnConvolutionForward(handle, &alpha, xDesc, img, wDesc, f, convDesc, CUDNN_CONVOLUTION_FWD_ALGO_GEMM, 0, 0, &beta, yDesc, out, 1, 3, 2, 1, 0);
+
+    for (int i = 0; i < 4; i++)
+    {
```
lucasccordeiro: Remove brackets.

regression/cuda/benchmarks/114_cudnn_conv/main.cu
```diff
@@ -0,0 +1,41 @@
...
+
+    float alpha = 1.0f, beta = 0.0f;
+    float img[1 * 1 * 3 * 3] = {1, 2, 3, 4, 5, 6, 7, 8, 9};
+    float f[1 * 1 * 2 * 2] = {1, 2, 3, 4};
+    float out[1 * 1 * 2 * 2] = {};
+    float dev[1 * 1 * 2 * 2] = {37, 47, 67, 77};
+
+    cudnnConvolutionForward(handle, &alpha, xDesc, img, wDesc, f, convDesc, CUDNN_CONVOLUTION_FWD_ALGO_GEMM, 0, 0, &beta, yDesc, out, 1, 3, 2, 1, 0);
+
+    for (int i = 0; i < 4; i++)
```
lucasccordeiro: Why 4?

lucasccordeiro (changes requested):

regression/cuda/benchmarks/114_cudnn_conv/main.cu
XLiZHI: I've added some notes in the comments.

lucasccordeiro (approved):

lucasccordeiro:

emanino:
Hi @XLiZHI ,

I looked through your operational models with interest. In particular, I was curious to see your implementation of the sigmoid activation function. I have three questions:

1) How did you decide that 4000 entries are enough to represent the sigmoid accurately? What was your rationale behind this choice?

2) Many of the entries are repeated, especially around an input u very close to zero. I worry that this is due to some bug in the code that produced the values of the table.

3) The implementation of the functions cudnnActivationForward() and cudnnActivationBackward() assume that the activation is a sigmoid. Are these functions not meant to be generic and take any activation function as inputs? Specifically, how is the argument cudnnActivationDescriptor_t activationDesc meant to be used?

I look forward to reading your thoughts on this.

XLiZHI:
Hi @emanino,

Thank you for your interest in this OM.

> How did you decide that 4000 entries are enough to represent the sigmoid accurately?

For the implementation of the activation function, I refer to previous research https://arxiv.org/pdf/1907.12933.pdf
We cannot guarantee the precision of the activation function in CuDNN because it uses an approximate method to speed up the computation of the function. 4000 is most likely a parameter used to retrieve the lookup table, and the sigmoid function ranges from 0 to 1, so it's correct as long as the lookup table covers that.

> Many of the entries are repeated

Yes, I've noticed that too. It has to do with precision, like we're now accurate to 6 decimal places, but the data actually has values after the sixth decimal place, and we just omitted part of the decimals. I think we can improve it in the future.

> The implementation of the functions cudnnActivationForward() and cudnnActivationBackward() assume that the activation is a sigmoid. Are these functions not meant to be generic and take any activation function as inputs? Specifically, how is the argument cudnnActivationDescriptor_t activationDesc meant to be used?

As for this part, CuDNN does not publish its source code. If we follow CuDNN's logic to pass parameters, our verification efficiency will be greatly reduced (we expect to simplify some operations that are not related to verification). So, this part is still under development, and maybe a better way will be found.

Hope this answers your questions.

emanino:
> For the implementation of the activation function, I refer to previous research https://arxiv.org/pdf/1907.12933.pdf We cannot guarantee the precision of the activation function in CuDNN because it uses an approximate method to speed up the computation of the function. 4000 is most likely a parameter used to retrieve the lookup table, and the sigmoid function ranges from 0 to 1, so it's correct as long as the lookup table covers that.

Thank you for your answer. As a word of advice, be wary of trusting old unmaintained research code. I strongly suggest to refactor that sigmoid lookup table in the near future.

> If we follow CuDNN's logic to pass parameters, our verification efficiency will be greatly reduced (we expect to simplify some operations that are not related to verification).

I understand. However, how do you verify a neural network that does not contain sigmoids but ReLUs or some other activation function? Do you not call cudnnActivationForward() and cudnnActivationBackward() in that case?

XLiZHI:
> As a word of advice, be wary of trusting old unmaintained research code. I strongly suggest to refactor that sigmoid lookup table in the near future.

Thank you very much for your advice!

> However, how do you verify a neural network that does not contain sigmoids but ReLUs or some other activation function? Do you not call cudnnActivationForward() and cudnnActivationBackward() in that case?

Actually our CuDNN model is not complete yet, I have not built a complete neural network model (i.e. the model is missing a fully connected layer), so this part is still a work in progress.

PR #1254, opened 2023-08-04
Two fixes

fbrausse:
This PR fixes #1252, #1253.

The first one was missing a simple check in the loop-unroller. (Is this `get_loop_bounds()` code actually useful? It looks overly specific to me.) Unfortunately, we still can't solve the SV-COMP test case (though it is quite small and clearly crafted) so I've put an even reduced version into regression/disabled.

The second bug was using an unitialized value (`array_id`) in array_conv causing a wrong equality being asserted during collate_array_values(). Besides the SV-COMP error, this was also observable in (at least)
- regression/esbmc-unix2/01_pthread58
- regression/esbmc-unix2/01_pthread_join_2
- regression/esbmc/fam_false_0

with valgrind. The new assertion in the tuple-node-flattener would sometimes trigger in such cases. I've also added a reduced version of the benchmark as a regression test.

*Edit:* Also fixes #632.

PR #1264, opened 2023-08-07
Make irep2 construction a bit more C++y

fbrausse:
This PR is removing a few constructors for irep2 types. The reason is that the current way is highly context-sensitive, e.g. for `not2tc(expr)` and similar things that take just a single or no argument like `bool_type2tc()`. Is it empty? Is it a bool-type?

This means that things like the following are no longer allowed, where `expr` is a C++ variable of type `expr2tc`:
- `constant_int2tc c = expr`
- `constant_int2tc c(expr)`

Also, the conversion from a raw pointer is not supported anymore, because it is error-prone and leads to unnecessary encoding of the internal workings irep2 is doing for construction.
- `expr2tc(new constant_int2t(type, 5))`

We don't need to explicitly allocate memory at every call site. Removing these constructions allows us to call `std::make_shared<Type>` which in theory should save some memory allocations.

However, the main benefit (for me) is consistency and less context-sensitivity.

*Edit:* I'm also removing a few unused or duplicated concepts and rename an expr2 trait:
- variadic_vector
- set_of_traits
- expr2t_traits_notype
- expr2t_traits_always_construct => expr2t_traits_notype
- dummy_type_tag


mikhailramalho (approved):
Ow, that's something I've tried to implement a couple of times and never succeeded!

Also, I think we can make some of the irep2 constructors constexpr now, right?

lucasccordeiro (approved):

lucasccordeiro:
@fbrausse: Thanks for submitting this Pr ;-)

fbrausse:
> Also, I think we can make some of the irep2 constructors constexpr now, right?

Which ones did you have in mind? At the moment I'm still a bit unhappy with the current state because I couldn't make the only remaining something2tc() constructor perfect-forwarding. Also the type traits are still unremovable.

mikhailramalho:
At least the ones you set '= default;' can be made constexpr, I think

fbrausse:
~OK, I've finally been able to make the something2tc's constructor perfect-forwarding. It should now be possible to std::move() things like vectors. Of course none of the concrete expr2t or type2t constructors expect that, yet, so it won't be of much help immediately.~

~I'm happy now, from my point of view this PR is good to go.~

*Edit:* Sorry, strike the above. Having sets of constructors with different semantics reintroduces the ambiguity: `not2tc(not2tc(expr))` would just be a copy of `not2tc(expr)`. I don't think we can avoid this without making `something2tc` a function only, instead of a full-blown C++ type.

fbrausse:
I've removed the ambiguity (and the perfect-forwarding) again and started an SV-COMP run in https://github.com/esbmc/esbmc/actions/runs/5794385437

src/goto-programs/goto_check.cpp
```diff
@@ -248,8 +248,8 @@ void goto_checkt::overflow_check(
 
   // add overflow subgoal
   expr2tc overflow = is_neg2t(expr)
-                       ? expr2tc(new overflow_neg2t(to_neg2t(expr).value))
-                       : expr2tc(new overflow2t(expr));
+                       ? expr2tc(overflow_neg2tc(to_neg2t(expr).value))
+                       : expr2tc(overflow2tc(expr));
```
mikhailramalho, line 252: We don't need the expr2tc calls around overflow2tc and overflow_neg2tc, right?

src/irep2/irep2.h
```diff
@@ -177,60 +204,20 @@ template <class T>
...
+  constexpr irep_container() = default;
+  constexpr irep_container(const irep_container &ref) = default;
+  constexpr irep_container(irep_container &&ref) = default;
 
-  template <class Y>
-  explicit irep_container(const Y *p) : std::shared_ptr<T>(const_cast<Y *>(p))
-  {
-  }
+  irep_container &operator=(irep_container const &ref) = default;
+  irep_container &operator=(irep_container &&ref) = default;
```
mikhailramalho, line 212: constexpr?

src/irep2/irep2.h
```diff
@@ -177,60 +204,20 @@ template <class T>
...
-  {
-    std::shared_ptr<T>::operator=(r);
-    T *p = std::shared_ptr<T>::operator->();
-    return *this;
-  }
-
-  template <class Y>
-  irep_container &operator=(const irep_container<Y> &ref)
+  explicit irep_container(std::shared_ptr<T> &&p)
+    : std::shared_ptr<T>(std::move(p))
```
mikhailramalho, line 219: constexpr too maybe?

src/irep2/irep2.h
```diff
@@ -298,11 +285,15 @@ typedef irep_container<expr2t> expr2tc;
 typedef std::pair<std::string, std::string> member_entryt;
 typedef std::list<member_entryt> list_of_memberst;
 
+class irep2t : public std::enable_shared_from_this<irep2t>
+{
+};
+
```
mikhailramalho, line 291: wouldn't we allow implicit conversion by setting the same base class for type2t and expr2t? If so, we should remove this base class

mikhailramalho (commented):
I just left a couple more comments, but I wanted to know if there is any benefit in doing calls like:
```
    isinf2tc op1_inf { side_2 };
```
instead of:
```
    expr2tc op1_inf = isinf2tc(side_2);
```
If so, can you create another PR for that? This one is big enough already.

Also, can you share the memory results from sv-comp?

src/goto-programs/goto_check.cpp
fbrausse: Unfortunately we do because otherwise C++ complains about both sides not having equal types. I don't indend to put a lot more effort into something2tc (like conversion operators or the like), because in a later PR I would like to remove it.

src/irep2/irep2.h
fbrausse: We can mark it constexpr, but it won't help. It needs to copy/move the underyling std::shared_ptr, which [doesn't have](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%3D) a constexpr operator=().

src/irep2/irep2.h
fbrausse: The purpose is to allow implicit conversion of any type2t or expr2t to irep2t. Can't implicitly cast down from irep2t to the others, though. I don't need this base class in this PR and can remove it, however, for addressing #1208 I would eventually like to be able to build the explicit graph with both type2t and expr2t nodes inside it.

fbrausse:
> I just left a couple more comments, but I wanted to know if there is any benefit in doing calls like:
> 
> ```
>     isinf2tc op1_inf { side_2 };
> ```
> 
> instead of:
> 
> ```
>     expr2tc op1_inf = isinf2tc(side_2);
> ```

I am currently implementing the second thing. The reason is that as long as `not2tc` is a *type*, the constructor of `not2tc` cannot distinguish between copy-constructing a shared instance of the given one or constructing a new instance of `not2t` containing the given one. I don't see a way to resolve this ambiguity as long as the *2tc things are types. If on the other hand it was a function, then there is no more ambiguity. That's why I'm going for the second version.

> If so, can you create another PR for that? This one is big enough already.

I agree, it's already big. Will put those changes in another PR.

> Also, can you share the memory results from sv-comp?

Sure, as soon as it's ready.

src/goto-programs/goto_check.cpp
fbrausse: This could actually go away if we remove something2tc and make the *2tc() things function calls.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/5761763785) (https://github.com/esbmc/esbmc/commit/fda18bbc408e86d3fa577e388190f6e2d908253a) (actually from the last #1254, seems we forgot to update the stats-30s.txt again; I'll add it to this PR):
```
  correct:           14373
    correct true:     8144
    correct false:    6229
  incorrect:            59
    incorrect true:     29
    incorrect false:    30
  unknown:           12027
  Score:             21109 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5794385437) (6ae3bb3e786c6bc278bb7e7dcc1c6d40a899fc02):
```
  correct:           14368
    correct true:     8142
    correct false:    6226
  incorrect:            59
    incorrect true:     29
    incorrect false:    30
  unknown:           12032
  Score:             21102 (max: 43043)
```
There is only fluctuation around the 30s timeout that I can see.

> Also, can you share the memory results from sv-comp?

- no-overflow: image
- termination: image
- unreach-call: image
- valid-memsafety: image

fbrausse:
> seems we forgot to update the stats-30s.txt again; I'll add it to this PR

Or rather not since Lucas has already updated it and it would conflict with master. Branch rebased.

lucasccordeiro:

PR #1265, opened 2023-08-08
[Concurrency] Do not produce interleavings after __ESBMC_main ends

lucasccordeiro:
This approach to stop producing interleavings after __ESBMC_main, when checking memory leaks, keeps our SV-COMP regression:

This PR:
````
   correct:             434
     correct true:      146
    correct false:     288
   incorrect:             9
     incorrect true:      9
     incorrect false:     0
   unknown:             246
   Score:               292 (max: 1036)
````
GitHub actions: https://github.com/esbmc/esbmc/actions/runs/5807542502

Master:

````
   correct:             434
     correct true:      146
     correct false:     288
   incorrect:             9
     incorrect true:      9
     incorrect false:     0
   unknown:             246
   Score:               292 (max: 1036)
````

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/5789501214

### DIFF
--- a/regression/esbmc-unix/01_malloc_17/main.c
+++ b/regression/esbmc-unix/01_malloc_17/main.c
@@ -1,0 +1,37 @@
+#include <pthread.h>
+#include <stdlib.h>
+
+pthread_mutex_t *mutex;
+
+void *thread1(void *arg)
+{
+  pthread_mutex_lock(mutex); 
+  pthread_mutex_unlock(mutex);
+
+  return NULL;
+}
+
+
+void *thread2(void *arg)
+{
+  pthread_mutex_lock(mutex); 
+  pthread_mutex_unlock(mutex);
+
+  return NULL;
+}
+
+
+int main(void)
+{
+  pthread_t id1, id2;
+
+  mutex = (pthread_mutex_t *) malloc(sizeof(pthread_mutex_t));
+  if (mutex==NULL)
+    exit(0);
+  pthread_mutex_init(mutex, NULL);
+
+  pthread_create(&id1, NULL, thread1, NULL);
+  pthread_create(&id2, NULL, thread2, NULL);
+
+  return 0;
+}
--- a/regression/esbmc-unix/01_malloc_17/test.desc
+++ b/regression/esbmc-unix/01_malloc_17/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--memory-leak-check
+^VERIFICATION FAILED$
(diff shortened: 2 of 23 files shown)


Anthonysdu:
So it means we stop interleaving if the main thread has ended? Why it is particularly used in memeory leak checking?

lucasccordeiro:
Hi @Anthonysdu,

In principle, we should stop interleaving after the __ESBMC_main has ended. However, this approach has implications for the deadlock and data races check. We don't maintain these checks for quite some time. It would require further investigation to understand why we miss the deadlock and data race bugs when we stop interleaving after the __ESBMC_main ends.

I have asked @fbrausse to prepare a PR where we modify the semantics of the abort function to follow the SV-COMP rules. I would like to see how this PR performs memory leak checks in multi-threaded programs using the semantics of the abort function defined by SV-COMP.

In any case, we should investigate the deadlock and data race check issue.

fbrausse:
> However, this approach has implications for the deadlock and data races check.

If so, shouldn't the condition to stop interleaving also depend on whether those flags are set?

lucasccordeiro:
> If so, shouldn't the condition to stop interleaving also depend on whether those flags are set?

@fbrausse: It makes sense.

I have started an SV-COMP run with this approach at https://github.com/esbmc/esbmc/actions/runs/5812475835.

lucasccordeiro:
This PR with the condition to continue producing interleavings when `__ESBMC_main` ends for data race and deadlock checks:

````
   correct:             433
     correct true:      145
     correct false:     288
   incorrect:             9
     incorrect true:      9
     incorrect false:     0
   unknown:             247
   Score:               290 (max: 1036)
````

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/5812475835

lucasccordeiro:
This PR over the concurrency memory safety benchmarks without memory leak check enabled in our SV-COMP script:

````
2023-08-10T14:07:28.1158850Z Statistics:            744 Files
2023-08-10T14:07:28.1159935Z   correct:             447
2023-08-10T14:07:28.1160927Z     correct true:      447
2023-08-10T14:07:28.1163484Z     correct false:       0
2023-08-10T14:07:28.1164195Z   incorrect:            13
2023-08-10T14:07:28.1179027Z     incorrect true:      2
2023-08-10T14:07:28.1179975Z     incorrect false:    11
2023-08-10T14:07:28.1180521Z   unknown:             284
2023-08-10T14:07:28.1180992Z   Score:               654 (max: 1480)
````

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/5818881940

PR #1266, opened 2023-08-08
irep2.2: the *2tc() constructs are now functions and no longer types

fbrausse:
This PR removes the possibility to construct irep2 terms like so:
- `constant_int2tc x(type, 5)`

In particular, the `constant_int2tc` or any other *2tc symbol in the code is no longer a type, but a function. This path was chosen because it removes any remaining ambiguity that might have existed for `not2tc(not2tc(expr))`. It is now clear that no copy-construction of the inner `not2tc(expr)` is being performed here but instead a double-negation of `expr` is constructed.

The only types remaining to reference the irep2 expressions are `expr2tc` and `type2tc`. In addition of course the `*2t` (without `c`) variants still exist, but they now have to be obtained through the `to_*()` functions.

I've added a few words about those and about the necessity to avoid const_cast() for correctness in the irep2 comments.

~At the moment this PR is still based on #1264, so I'll mark it a draft while the other one is still open. It might not be worth reviewing while it still contains the changes of the base PR.~

*Edit:* Besides `something2tc` I'm again removing some unused or unnecessary concepts:
- expr2t::num_nodes() and ::depth() also because they were misleading
- esbmct::expr2t_default_traits
- the `container` template param to some of the irep2 type traits
- special-cases for "empty" expr2t-s

To my mind, besides removing the ambiguity, the approach taken in this PR has a few benefits over allowing `constant_int2tc x(type, 5)` to coexist:
- every new irep2 term is written as `expr2tc e = specific2tc(...)` or `type2tc t = specific2tc(...)`
- when writing every irep2 term like that, `e` and `t` can be overwritten by other expr2tc or type2tc subsequently, without causing typing problems as would be the case with `x` above
- expr2tc and type2tc can now be seen as algebraic types where the `specific2tc(...)` is a constructor (and it has others, too), and the `to_*/is_*()` functions are the equivalent of destructors or match expressions; in the future we could consider combining these two functions into one `as_*()` returning the combined information in form of a pointer, for instance
- if we are crazy enough, we could even move the expr_id and type_id fields into the expr2tc/type2tc "pointers" and save a dereference for `is_*/as_*()`


fbrausse:
I've included in this PR modifications of the irep_container. Main ones are:
- the underlying std::shared_ptr is a private base class and no longer accessible. The reasons are detailed in the comments, but boil down to: const-ness guarantees should hold
- it has its own (full) set of comparison operators now, which I tweaked a bit to also compare addresses; this has measurable effects on, for instance, regression/esbmc/17_apache

I'm running it over SV-COMP as well, but I'd also like to know whether this rather invasive change finds approval.

lucasccordeiro (approved):

I'm okay with this approach.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/5772779672) (c53f31cf8315c48d8bc71fe9877c813ccb4a2646):
```
  correct:           14357
    correct true:     8130
    correct false:    6227
  incorrect:            58
    incorrect true:     29
    incorrect false:    29
  unknown:           12044
  Score:             21095 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5811632478) (1f49fd6b36897b9457f21b328c1637c7ee2c239d):
```
  correct:           14370
    correct true:     8142
    correct false:    6228
  incorrect:            59
    incorrect true:     29
    incorrect false:    30
  unknown:           12030
  Score:             21104 (max: 43043)
```
The new timeout -> incorrect was incorrect already on master, just slower:
- goblint-coreutils/instrumented_interval_cut_comb.i

In unreach-call there are 2 unknown -> correct which failed on master due to segfaults (somewhere in the irep2 simplifier for less-/greater-than expressions):
- c/float-newlib/double_req_bl_1231b.c
- c/float-newlib/double_req_bl_1251b.c

As expected, the additional check affects very fast benchmarks a tiny bit wrt. cputime: image
For memory on the other hand, this PR seems to consistently perform equal or better: image

fbrausse:
> For memory on the other hand, this PR seems to consistently perform equal or better:

Though it may be comparing apples and oranges, since the #1264 already did that, but here I'm comparing against a run without that. Comparing to the #1264 run, this PR also has the 2x unknown -> correct and shows no difference wrt. memory.

rafaelsamenezes (approved):

shmarovfedor (approved):

lucasccordeiro:

Issue #1259, opened 2023-08-05
Investigate the memory leak check implemented in ESBMC


lucasccordeiro:
The memory leak check implemented in ESBMC was disabled in the last competition due to many incorrect results.

This recent PR reports some numbers: https://github.com/esbmc/esbmc/pull/1250. However, there are still more issues beyond the `getenv` function that needs to be investigated.

fbrausse:
Looking at `c/weaver/unroll-2.wvr.yml`, ESBMC claims "forgotten memory" (which is true, an array is malloc'ed and not free'd), but to me it seems that they assume `abort()` would not perform memleak checks since they encode assumptions via `assume_abort_if_not()`, which aborts. When modifying it:
```diff
diff --git a/c/weaver/unroll-2.wvr.c b/c/weaver/unroll-2.wvr.c
index 1d1d1f580c..87c94190b2 100644
--- a/c/weaver/unroll-2.wvr.c
+++ b/c/weaver/unroll-2.wvr.c
@@ -77,13 +77,13 @@ int main() {
   f = create_fresh_int_array(size);
   
   assume_abort_if_not(n < 4294967296 / 2);
-
+#if 0 /* let's ignore the finite state manipulation done here */
   // main method
   pthread_create(&t1, 0, thread1, 0);
   pthread_create(&t2, 0, thread2, 0);
   pthread_join(t1, 0);
   pthread_join(t2, 0);
-
+#endif
   assume_abort_if_not(x1 != x2);
   reach_error();
 
@@ -95,8 +95,10 @@ int *create_fresh_int_array(int size) {
   assume_abort_if_not(size <= (((size_t) 4294967295) / sizeof(int)));
 
   int* arr = (int*)malloc(sizeof(int) * (size_t)size);
+#if 0 /* this is what ESBMC does anyway, only takes time */
   for (int i = 0; i < size; i++) {
     arr[i] = __VERIFIER_nondet_int();
   }
+#endif
   return arr;
 }
```
ESBMC still claims `forgotten memory` because abort() calls the __ESBMC_memory_leak_checks(). Removing that from the abort() OM makes the modified one work as SV-COMP intends (VERIFICATION SUCCESSFUL).

However, like so many things in SV-COMP, this is highly underspecified (unless I missed something). For memsafety they claim they look at all finite executions. To me calling `abort()` does terminate (abnormally) the execution, making it definitely finite...

fbrausse:
[...]

PR #1267, opened 2023-08-09
Allow not checking for memory leaks on abnormal termination

fbrausse:
This PR adds a new #define passed by the esbmc-wrapper.py and which is interpreted in c2goto/stdlib.c by our OMs for `abort()` and `_Exit()`. Addresses part of #1259.

### DIFF
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -217,6 +217,11 @@ esbmc_dargs += "-D'__builtin_unreachable()' "
 # <https://github.com/esbmc/esbmc/pull/1190#issuecomment-1637047028>
 esbmc_dargs += "--no-vla-size-check "
 
+memleak_check_args = "--memory-leak-check "
+# It seems SV-COMP doesn't want to check for memleaks on abort(), etc.
+# see also <https://github.com/esbmc/esbmc/issues/1259>
+memleak_check_args += "-D__ESBMC_NO_MEMLEAK_CHECK_ON_ABNORMAL_TERMINATION "
+
 import re
 def check_if_benchmark_contains_pthread(benchmark):
   with open(benchmark, "r") as f:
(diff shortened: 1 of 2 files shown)


mikhailramalho:
I don't think this will work as intended, as the models are converted to goto at build time

You may want to do something like the old __ESBMC_floatbv_mode, see:
c8c762fa10e558c12d3ef4119552fb9f6ff0af6f: the frontend handle
932d559ca1dc5fa46b9e4f69f3ef766ca86a2981: how it's used in the sqrt model

fbrausse:
@mikhailramalho thank you for the pointers. Indeed, it does not work as intended. :)

fbrausse:
The code you quoted has moved into the old frontend and seems to have since been made independent of that option in the c2goto-libm. I was looking for alternative examples and found `_Bool __ESBMC_is_little_endian();` which is handled in goto_symext::run_intrinsic(). To me it seems better to replace such constants earlier than symex. Does anyone know why we aren't doing that anymore?

mikhailramalho:
oh, right, this has to be done in symex if we are changing the models, otherwise, the models will be generated with whatever flag is set when compiling esbmc.

For sqrt it was not a problem because we generate models for both fp and fixedbv.

fbrausse:
Right, that makes sense. But is there no step earlier than symex and aren't we running a kind of LTO-style full-program analysis after linking? I could imagine that the linker could set these constants in esbmc only (not c2goto) as it's running after the models are loaded.

Anyway, these hints already help, thanks!

lucasccordeiro:
Shall we use the same approach as checking for deadlock?

If `--deadlock-check` is used, then we invoke `int pthread_join_switch(pthread_t thread, void **retval)`; otherwise, we invoke `int pthread_join_noswitch(pthread_t thread, void **retval)`.

lucasccordeiro:
````C
  if(config.options.get_bool_option("deadlock-check"))
  {
    record_define("pthread_mutex_lock=pthread_mutex_lock_check");
    record_define("pthread_mutex_unlock=pthread_mutex_unlock_check");
    record_define("pthread_cond_wait=pthread_cond_wait_check");
    record_define("pthread_join=pthread_join_switch");
  }
  else if(config.options.get_bool_option("lock-order-check"))
  {
    record_define("pthread_join=pthread_join_noswitch");
    record_define("pthread_mutex_lock=pthread_mutex_lock_nocheck");
    record_define("pthread_mutex_unlock=pthread_mutex_unlock_nocheck");
    record_define("pthread_cond_wait=pthread_cond_wait_nocheck");
  }
  else
  {
    record_define("pthread_join=pthread_join_noswitch");
    record_define("pthread_mutex_lock=pthread_mutex_lock_noassert");
    record_define("pthread_mutex_unlock=pthread_mutex_unlock_noassert");
    record_define("pthread_cond_wait=pthread_cond_wait_nocheck");
  }
````

fbrausse:
> Shall we use the same approach as checking for deadlock?
> 
> If `--deadlock-check` is used, then we invoke `int pthread_join_switch(pthread_t thread, void **retval)`; otherwise, we invoke `int pthread_join_noswitch(pthread_t thread, void **retval)`.

Yes, that's an option. And for checking whether we fix the example mentioned in #1259 it's good enough. I'll create a PR for that. Looking at the bigger picture though, I think it would help to discuss a bit further whether an LTO-style GOTO pass would be useful and/or whether it makes sense to clearly define what these passes depend on and when they run.

PR #1269, opened 2023-08-10
[Concurrency] Remove unviable interleaving

Anthonysdu:
ESBMC produces incorrect false for this benchmark:

```
#include <pthread.h> 
int x, y;
int b1;

void thr1(){
  if (y){
    b1 = 0;
  }
}

void thr2(){
  x = 1;
  assert(x >= 1)
}

int main() {
  pthread_t t1, t2;
  pthread_create(&t1, NULL, thr1, NULL);
  pthread_create(&t2, NULL, thr2, NULL);
  return 0;
}
```
This is because esbmc has verified an unviable interleaving.

An unviable interleaving is defined as follows:
1. There is a context switch from last state to current state. (last_active_thread != active_thread),
2. and the guard of current active thread is false.

If an interleaving has switched to a thread whose guard is false, this interleaving should not only stop further exploration, but also clear current VCC(s). (Is that right?)

### DIFF
--- a/regression/esbmc-unix2/11_lamport/test.desc
+++ b/regression/esbmc-unix2/11_lamport/test.desc
@@ -1,4 +1,4 @@
 CORE
 lamport.c
---unwind 1 --no-unwinding-assertions
+--unwind 1 --no-unwinding-assertions --no-por
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-unix2/11_lamport/test.desc
+++ b/regression/esbmc-unix2/11_lamport/test.desc
@@ -1,4 +1,4 @@
 CORE
 lamport.c
---unwind 1 --no-unwinding-assertions --no-por
+--unwind 1 --no-unwinding-assertions --context-bound 2
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-unix2/11_lamport_2/lamport.c
+++ b/regression/esbmc-unix2/11_lamport_2/lamport.c
@@ -1,0 +1,21 @@
+#include <pthread.h> 
+int x, y;
+int b1;
+
+void thr1(){
+  if (y){
+    b1 = 0;
+  }
+}
+
+void thr2(){
+  x = 1;
+  assert(x >= 1);
+}
+
+int main() {
+  pthread_t t1, t2;
+  pthread_create(&t1, NULL, thr1, NULL);
+  pthread_create(&t2, NULL, thr2, NULL);
+  return 0;
+}
--- a/regression/esbmc-unix2/11_lamport_2/test.desc
+++ b/regression/esbmc-unix2/11_lamport_2/test.desc
@@ -1,0 +1,4 @@
+CORE
+lamport.c
+--unwind 1 --no-unwinding-assertions --no-por
+^VERIFICATION SUCCESSFUL$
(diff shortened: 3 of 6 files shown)


lucasccordeiro:
> If an interleaving has switched to a thread whose guard is false, this interleaving should not only stop further exploration, but also clear current VCC(s). (Is that right?)

This approach will remove all previously generated VCCs. Is it what you want to achieve if interleaving is not viable?

lucasccordeiro:
@Anthonysdu: It is weird that we perform this check just in the end at `std::shared_ptr<goto_symext::symex_resultt>
reachability_treet::get_next_formula()`.

````C
    if(get_cur_state().interleaving_unviable)
      break;
````

We might perform this check at the beginning of the method to avoid generating the VCC.

lucasccordeiro:
Something like:

````C
    next_thread_id = decide_ileave_direction(get_cur_state());

    if(get_cur_state().interleaving_unviable)
      break;
      
    create_next_state();

    switch_to_next_execution_state();
````

Anthonysdu:
> > If an interleaving has switched to a thread whose guard is false, this interleaving should not only stop further exploration, but also clear current VCC(s). (Is that right?)
> 
> This approach will remove all previously generated VCCs. Is it what you want to achieve if interleaving is not viable?

Yes remove all the VCC(s) in current unviable interleaving, then start a new interleaving.

Anthonysdu:
> Something like:
> 
> ```c
>     next_thread_id = decide_ileave_direction(get_cur_state());
> 
>     if(get_cur_state().interleaving_unviable)
>       break;
>       
>     create_next_state();
> 
>     switch_to_next_execution_state();
> ```

Yes..
I will have a try

Anthonysdu:
> Something like:
> 
> ```c
>     next_thread_id = decide_ileave_direction(get_cur_state());
> 
>     if(get_cur_state().interleaving_unviable)
>       break;
>       
>     create_next_state();
> 
>     switch_to_next_execution_state();
> ```

This apporach works!

lucasccordeiro:
> This apporach works!

I'm glad to hear that ;-)

lucasccordeiro (approved):
@Anthonysdu: can you please run this PR over the SV-COMP benchmarks?

lucasccordeiro:
I'm running this PR over the SV-COMP benchmarks via https://github.com/esbmc/esbmc/actions/runs/5823405050.

lucasccordeiro:
``--no-por, --context-bound 2, 900 seconds:``

Before this PR:
````
   correct:             434
     correct true:      146
     correct false:     288
   incorrect:             9
     incorrect true:      9
     incorrect false:     0
   unknown:             246
   Score:               292 (max: 1036)
````
GitHub actions: https://github.com/esbmc/esbmc/actions/runs/5789501214


This PR:
````
2023-08-11T13:13:34.0485334Z Statistics:            689 Files
2023-08-11T13:13:34.0485872Z   correct:             435
2023-08-11T13:13:34.0486440Z     correct true:      147
2023-08-11T13:13:34.0487241Z     correct false:     288
2023-08-11T13:13:34.0487778Z   incorrect:             9
2023-08-11T13:13:34.0488345Z     incorrect true:      9
2023-08-11T13:13:34.0502804Z     incorrect false:     0
2023-08-11T13:13:34.0504403Z   unknown:             245
2023-08-11T13:13:34.0505419Z   Score:               294 (max: 1036)
````
GitHub actions: https://github.com/esbmc/esbmc/actions/runs/5829337539

PR #1281, opened 2023-08-14
Allow more fine-grained control over debug log messages

### LINKED ISSUE #1280: Debug log-level becoming unwieldy
Passing `--verbosity 9` (or similarly high values) to ESBMC generates a lot of output. Some parts list all the things they're touching and these lists are not only in unrelated parts of the code in most cases, but also of arbitrary length. This makes the Debug log-level hard to use.

It would be good if each `log_debug()` call would contain a reference to a "module" to which this Debug message belongs. If such modules would also have their own log-levels (in contrast to the global one we have in `messaget::state` now), then for instance via something like `--verbosity slicer:9` we could get Debug messages from the slicer but leave everything else on the default log-level.

Such a reference to a module could be a free-form string and the module itself just an entry that maps the string to a log-level. The module could potentially also contain more logging-related information, like color of the messages, etc.

fbrausse:
This PR allows `--verbosity` to be specified multiple times, where later specifications overwrite earlier ones. In addition, to make this meaningful, instead of just a number N its parameter can optionally be a string of the form "module:N" where module is one of the strings specifying which "module" a log_debug() message belongs to. The following "modules" for these debug messages are introduced in this PR, based on the context the log_debug() calls are made in:
- clang
- c++
- multi-property
- interval
- contractor
- goto-loop
- memset
- goto-trace
- rename
- slice
- ssa
- jimple
- solidity
- cvc4
- mathsat
- tuple-node
- smtlib

Closed #1280. As an additional change, the log_*() things are now macros so they allow access to the correct `__FILE__` and `__LINE__`. This information is not printed, yet. The format of the log output also remains unchanged (modulo typos I found).

Example: `--verbosity slice:9` will now only print debug messages from the slicer and keep the normally output messages unchanged. In particular, it does not print the clang invocation params or debug messages from other "modules".

### DIFF
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -435,7 +435,7 @@ const struct group_opt_templ all_cmd_options[] = {
     // Abort if the program contains a recursion
     {"abort-on-recursion", NULL, ""},
     // Verbosity of message, probably does nothing
-    {"verbosity", boost::program_options::value<int>(), ""},
+    {"verbosity", boost::program_options::value<std::vector<std::string>>(), ""},
     // --break-at $insnnum will cause ESBMC to execute a trap
     // instruction when it executes the designated GOTO instruction number.
     {"break-at", boost::program_options::value<std::string>(), ""},
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -434,10 +434,13 @@ const struct group_opt_templ all_cmd_options[] = {
     {"double-assign-check", NULL, ""},
     // Abort if the program contains a recursion
     {"abort-on-recursion", NULL, ""},
-    // Verbosity of message, probably does nothing
+    /* see <https://github.com/esbmc/esbmc/pull/1281> for a list of supported
+     * modules; check "grep -rw 'log_debug(' src" for more up-to-date info. */
     {"verbosity",
      boost::program_options::value<std::vector<std::string>>(),
-     ""},
+     "Verbosity of log output, can be given multiple times. Parameter is "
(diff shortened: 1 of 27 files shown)


lucasccordeiro (approved):

src/esbmc/options.cpp
```diff
@@ -435,7 +435,9 @@ const struct group_opt_templ all_cmd_options[] = {
     // Abort if the program contains a recursion
     {"abort-on-recursion", NULL, ""},
     // Verbosity of message, probably does nothing
```
kunjsong01: Shall we update the comment? "probably does nothing" -> "see PR128 for fine-grained control over debug log message"

src/esbmc/options.cpp
```diff
@@ -435,7 +435,9 @@ const struct group_opt_templ all_cmd_options[] = {
...
-    {"verbosity", boost::program_options::value<int>(), ""},
+    {"verbosity",
+     boost::program_options::value<std::vector<std::string>>(),
+     ""},
```
kunjsong01: we can probably add more help info in line 440, something like "module:N to get the N-level debug messages from a specific module"

kunjsong01 (approved):
Nice feature! LGTM

Just some nitpicks on comments and help info

src/esbmc/options.cpp
fbrausse: Done.

src/esbmc/options.cpp
fbrausse: Done.

lucasccordeiro:

PR #1284, opened 2023-08-15
Log output

fbrausse:
This PR does 3 things:
- update the `fmt` library from 7.1.3 to 9.1.0,
- perform compile-time checks of all the format strings going to log_*() + fixing some bugs related to that, and
- as discussed in <https://github.com/esbmc/esbmc/discussions/678> the output is to `stderr` only.

As particularly the last item might change the user experience if using ESBMC in a scripted context, I'd like to know whether anyone sees problems with printing to `stderr` only by default. This can of course be changed via `--file-output`, see <https://github.com/esbmc/esbmc/pull/655>.


rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #1291, opened 2023-08-18
include filename in output

Anthonysdu:
To make the log more readable if there are many input files,
otherwise it will output as:
```
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
Parsing
```

### DIFF
--- a/src/langapi/language_ui.cpp
+++ b/src/langapi/language_ui.cpp
@@ -28,7 +28,7 @@ bool language_uit::parse(const std::string &filename)
 
   if(mode < 0)
   {
-    log_error("failed to figure out type of file", filename);
+    log_error("failed to figure out type of file {}", filename);
     return true;
   }
 
@@ -48,7 +48,7 @@ bool language_uit::parse(const std::string &filename)
   std::ifstream infile(filename.c_str());
   if(!infile)
   {
-    log_error("failed to open input file", filename);
+    log_error("failed to open input file {}", filename);
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


mikhailramalho (approved):
lgtm.

mikhailramalho:
Can you also check the "unwinding" message? I think it saw it with a similar problem (not showing the loop info).

lucasccordeiro (approved):

lucasccordeiro:
> Can you also check the "unwinding" message? I think it saw it with a similar problem (not showing the loop info).

I fully agree with Mikhail.

@Anthonysdu: can you still do it in this PR?

Anthonysdu:
> Can you also check the "unwinding" message? I think it saw it with a similar problem (not showing the loop info).

Do you mean?
```
State 1 file main.c line 18 column 3 function main thread 0
Violated property:
  file main.c line 18 column 3 function main
  unwinding assertion loop 
```

Anthonysdu:
The loop info in class `locationt` is empty, and it seems we didn't implement a function to set the loop id.

My suggestion is showing the info instead: `unwinding assertion loop at line ***`,

or we implement more to represent the loop id.

What do you think? @lucasccordeiro @mikhailramalho

mikhailramalho:
No, the one printed during symex

Anthonysdu:
is it we want to do?

before:
```
Unwinding loop 2  iteration  1   file main.c line 18 column 3 function main
Not unwinding
```

after:

```
Unwinding loop 2  iteration  1   file main.c line 18 column 3 function main
Not unwinding loop 2  iteration  2   file main.c line 18 column 3 function main
```

mikhailramalho:
Yes!

PR #1295, opened 2023-08-22
Enhance the memsafety analysis

### LINKED ISSUE #1029: esbmc-wrapper.py interprets "forgotten memory" as an "unknown" result for the memory property
From https://github.com/esbmc/esbmc/pull/1019#issuecomment-1551019121
> In `valid-memsafety` there is 1 new correct -> unknown (`c/goblint-regression/09-regions_19-nested_nr-deref.i`) that did find a NULL dereference before, but now finds
`
dereference failure: forgotten memory: dynamic_2_value
`

In addition to a possible NULL-pointer dereference this benchmark does indeed also not free its allocated memory, so I gather the above verdict given by ESBMC is valid. However, as the title says, the `esbmc-wrapper.py` interprets this result as "unknown": https://github.com/esbmc/esbmc/blob/670ecf0ce25abb48c2759f3ac13a26be245da2e2/scripts/competitions/svcomp/esbmc-wrapper.py#L118-L120

Is this behaviour of the wrapper and/or ESBMC in this case intended?

fbrausse:
This PR enhances the memsafety check a bit further. It does:
- add an option `--malloc-zero-is-null` that results in `malloc(0)` and `alloca(0)` returning NULL. The SV-COMP memsafety benchmarks seem to assume this behaviour.
- replaces more uint_type2?() by size_type2() calls where array/memory sizes are meant
- fixes an oversight from the previous memsafety PR about the operator=() of goto_symext (this one seems to do special things, maybe it makes sense to separate the default handling - copying - from the special things by means of a new base-class of goto_symext?)
- most importantly, the memsafety check is enhanced by constructing an expression into structures and (constant-size) arrays based on the suffix computed by the value-set analysis. This helps in tracking the "still reachable" memory, which the memsafety category of SV-COMP does not want to be reported. For arrays of dynamic size I don't have an implementation, yet, since I would need a way to encode the co-reachability problem on graphs. The reachability problem (given a graph and two nodes s and t, is there a path from s to t?) has been studied in computational complexity theory and is NL-complete. Importantly, it also is co-NL-complete, due to the (constructive) proof by Immerman & Szelepcsényi, i.e. there are non-deterministic log-space algorithms solving the complement problem (is there *no* path from s to t?), but these are not straight-forward. It would be easier if we had the graph explicitly; I might add such a structure in the future.

For now, the constant-size arrays cover a few more memsafety benchmarks, which were previously solved incorrectly by ESBMC. An SV-COMP run is under way.


fbrausse:
[Merge base](https://github.com/esbmc/esbmc/actions/runs/5853305071)  (old master) at 88a4be1d5de348300e2b1b604848f6b85bc388d5:
```
  correct:           14763
    correct true:     8466
    correct false:    6297
  incorrect:            59
    incorrect true:     30
    incorrect false:    29
  unknown:           11637
  Score:             21805 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/5952234935) (c9fd291992372451c9ca60b8b25b11b133e59231) where the wrapper correctly reports fail_memtrack instead of unknown:
```
  correct:           15257
    correct true:     8463
    correct false:    6794
  incorrect:            61
    incorrect true:     31
    incorrect false:    30
  unknown:           11141
  Score:             22248 (max: 43043)
```
The incorrect ones in valid-memsafety change quite a bit:
- 11x incorrect (false valid-deref) -> timeout
- 12x unknown -> incorrect:
  - c/pthread/twostage_3-2.i is incorrect true
  - the others are determined to be incorrect false (valid-memtrack). Of those, however, we already checked that at least one (c/termination-recursive-malloc/chunk1-2.i) has a wrong verdict as our counter-exaample is valid: https://github.com/esbmc/esbmc/issues/1259#issuecomment-1676818480 The others with wrong verdict mentioned in that comment are unchanged by this PR.

In addition there is
- 1x timeout -> incorrect in unreach-call, it is c/goblint-coreutils/instrumented_interval_cut_comb.i which is not new and just timeout fluctuation

With respect to benchmarks classified as unknowns, there are:
- 3x timeout -> unknown in unreach-call; as far as I can see these are producing stack overflows for expressions created by our builtin memset():
  - c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--net--ethernet--cavium--liquidio--liquidio.ko-entry_point.cil.out.i
  - c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-43_2a-drivers--net--ethernet--cavium--liquidio--liquidio.ko-entry_point.cil.out.i
  - c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-43_2a-drivers--net--ethernet--chelsio--cxgb--cxgb.ko-entry_point.cil.out.i
- 2x correct (false valid-deref) -> unknown (false valid-memtrack):
  - c/ldv-validator-v0.6/linux-stable-42f9f8d-1-111_1a-sound--oss--opl3.ko-entry_point.cil.out.i
  - c/ldv-validator-v0.8/linux-stable-063f96c-1-144_1a-drivers--mmc--host--vub300.ko.unsigned-entry_point_ldv-val-v0.8.cil.out.i
- 2x timeout -> unknown (false valid-memtrack), and
- 34x unknown -> unknown (false valid-memtrack); all of these are supposed to be false valid-deref; I'll put them on my todo list

The rest is timeout fluctuation.

lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #1300, opened 2023-08-23
Remove goto ptr check

shmarovfedor:
In this PR I:
* Removed the unused (i.e., commented out) and outdated check for the pointers at the GOTO level.
* Introduced two new options for disabling the removal of unreachable and NO-OP code. Those are useful mostly for debugging and for providing some desired functionality during multi-property checking (see #1293).
* Performed some refactoring of the `esbmc_partseoptions` module:
  * merged some duplicated code,
  * renamed some methods to be more descriptive,
  * added description to the remaining methods and summarised the general ESBMC flow.


lucasccordeiro (approved):
Thanks for submitting this PR, @shmarovfedor.

fbrausse (commented):
Nice cleanup! I like the usage of tvt. It's makes the meaning much clearer.

emanino:
Thanks for adding thorough comments on the core BMC steps. I think they make the structure of the code easier to understand.

PR #1303, opened 2023-08-25
Handle more symbolic types surviving thrash_type_symbols()

### LINKED ISSUE #1289: ESBMC 7.3 crashes (Symbolic type id in type_size_t::size_bits_expr)
ESBMC 7.3 crashes for this small C program using the following command: `esbmc DATA2/FormAI_24071.c --overflow`

```c
//FormAI DATASET v1.0 Category: Matrix operations ; Style: complex
#include <stdio.h>

#define MATRIX_SIZE 3

typedef struct {
    float real;
    float imag;
} Complex;

void print_complex(Complex num) {
    printf("%f", num.real);
    if (num.imag >= 0) {
        printf("+%fi", num.imag);
    } else {
        printf("%fi", num.imag);
    }
}

void add_matrices(Complex matrix1[][MATRIX_SIZE], Complex matrix2[][MATRIX_SIZE], Complex result[][MATRIX_SIZE]) {
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            result[i][j].real = matrix1[i][j].real + matrix2[i][j].real;
            result[i][j].imag = matrix1[i][j].imag + matrix2[i][j].imag;
        }
    }
}

void multiply_matrices(Complex matrix1[][MATRIX_SIZE], Complex matrix2[][MATRIX_SIZE], Complex result[][MATRIX_SIZE]) {
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            result[i][j].real = 0;
            result[i][j].imag = 0;
            for (int k = 0; k < MATRIX_SIZE; k++) {
                Complex product;
                product.real = matrix1[i][k].real * matrix2[k][j].real - matrix1[i][k].imag * matrix2[k][j].imag;
                product.imag = matrix1[i][k].real * matrix2[k][j].imag + matrix1[i][k].imag * matrix2[k][j].real;

                result[i][j].real += product.real;
                result[i][j].imag += product.imag;
            }
        }
    }
}

int main() {
    Complex matrix1[MATRIX_SIZE][MATRIX_SIZE] = {
        { {1, 2}, {3, 4}, {5, 6} },
        { {7, 8}, {9, 10}, {11, 12} },
        { {13, 14}, {15, 16}, {17, 18} }
    };

    Complex matrix2[MATRIX_SIZE][MATRIX_SIZE] = {
        { {2, 1}, {4, 3}, {6, 5} },
        { {8, 7}, {10, 9}, {12, 11} },
        { {14, 13}, {16, 15}, {18, 17} }
    };

    Complex matrix_sum[MATRIX_SIZE][MATRIX_SIZE];
    Complex matrix_product[MATRIX_SIZE][MATRIX_SIZE];

    add_matrices(matrix1, matrix2, matrix_sum);
    multiply_matrices(matrix1, matrix2, matrix_product);

    printf("Matrix 1:\n");
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            print_complex(matrix1[i][j]);
            printf("\t");
        }
        printf("\n");
    }
    printf("\n");

    printf("Matrix 2:\n");
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            print_complex(matrix2[i][j]);
            printf("\t");
        }
        printf("\n");
    }
    printf("\n");

    printf("Matrix Sum:\n");
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            print_complex(matrix_sum[i][j]);
            printf("\t");
        }
        printf("\n");
    }
    printf("\n");

    printf("Matrix Product:\n");
    for (int i = 0; i < MATRIX_SIZE; i++) {
        for (int j = 0; j < MATRIX_SIZE; j++) {
            print_complex(matrix_product[i][j]);
            printf("\t");
        }
        printf("\n");
    }
    printf("\n");

    return 0;
}
```

The output is :
```
esbmc DATA2/FormAI_24071.c --overflow
Parsing
Converting
Generating GOTO Program
GOTO program creation time: 0.051s
ERROR: Symbolic type id in type_size_t::size_bits_expr
symbol
* symbol_name : tag-struct __anon_typedef_Complex_at_FormAI_24071_c__6_9
Aborted (core dumped)
```

fbrausse:
Fixes #1289 using method (b) discussed there. ~using method (a) discussed there.~

~This patch introduces something I called a "context type graph". The reason is that I thought it best to go via the strongly connected components to decide which symbolic types are safe to replace under pointers.~

~I actually wanted the whole graph (including expressions, locations, comments, etc.), however, when developing this patch I realized there actually is no generic way of finding out whether an irept object is a type, an expression, a location or something else. At least I didn't find any. So now the graph is just the subgraph of that restricted to types, the context itself, symbols and symbols' value expressions.~

*Edit:* Turns out, the original approach, method (a), produced more locations where symbolic types need handling. So I removed it from this PR and just kept the additional namespace parameters passed to the type_byte_size*() family of functions. I realize that using migrate_namespace_lookup during irep2 simplification is not a long-term solution since we want to eventually get rid of it. But as stated in the original issue, just not simplifying when symbolic types are present is a bad option as well.

Just as a reminder, symbolic types are used to represent so-called non-tree edges in the depth-first exploration of the type graph. Here is an example illustrating how the C type `struct slist { struct slist *next; }` is represented: image
The rounded node is a symbol in the context, the hard lines represent typet fields inside a symbolt or typet (so-called tree-edges) and the dashed line is what the symbol-type refers to. That's a back-edge for this symbol. There could be other non-tree edges when the type does refer to other types and not to itself. thrash_type_symbols() contracts certain ones of these non-tree-edges (and the edge from the symbol to its type). It doesn't touch types under pointers, though, in order to avoid cycles in the internal representation.

This poses problems with pointer arithmetic as witnessed by #1289, because symbol-types may still be present under pointers. So originally, I tried to contract all non-tree edges except for back-edges regardless of whether under a pointer or not. But that led to more symbolic types appearing down the line, fixes for which are now in this PR. The new non-tree edge contraction I removed, though, since it's not needed when there is more ns.follow().


fbrausse:
A quick note: though this fixes the particular issue mentioned initially, it is not comprehensive. Strongly connected components just happen to be what our recursions can't deal with. They could still be part of a user program. To fix that I think there are only two ways (ideally, combined):
1. get rid of irep1, and/or
2. get rid of thrash_type_symbols.

To my mind, both are part of a transition to a more explicit approach regarding representing the context as a graph. This might also save memory.

lucasccordeiro:
@fbrausse: can I ask you whether you would have further updates about this PR?

fbrausse:
@lucasccordeiro I'm a bit unhappy about not using the irept API and have an idea how to fix that. I'll mark it a draft until I've figured it out.

lucasccordeiro:
Excellent!

Thanks, @fbrausse.

fbrausse:
> I'm a bit unhappy about not using the irept API and have an idea how to fix that. I'll mark it a draft until I've figured it out.

While in principle this approach would also support creating cycles in our irep data structures, this PR explicitly does not: that's what the "large SCCs" are for. Plus it's not super intrusive, so after thinking about it I am fine with this rather local change.

The alternative approach #1310 is based on this PR, so merging this one does not break the other one in case there are reasons to use the other approach later. I'll mark this PR as ready and would be interested in feedback whether the graph-based approach is acceptable here.

src/goto-programs/goto_convert_functions.cpp
```diff
@@ -354,9 +606,108 @@ void goto_convert_functionst::wallop_type(
...
+  std::vector<node_id> stack;
+  size_t index = 0;
+
+  std::unordered_set<node_id> large_scc_syms;
+
+  sccst(const context_type_grapht &G) : G(G), data(G)
+  {
+  }
+
+  void tarjan(node_id v) /* Tarjan's 1972 algorithm to compute SCCs */
```
lucasccordeiro: Can you add the paper's title and publication venue here?

src/goto-programs/goto_convert_functions.cpp
```diff
@@ -354,9 +606,108 @@ void goto_convert_functionst::wallop_type(
...
+template <>
+struct hash<node_id> : hash<size_t>
+{
+  static_assert(sizeof(node_id) == sizeof(size_t));
+
+  size_t operator()(const node_id &v) const noexcept
+  {
+    size_t w;
+    memcpy(&w, &v, sizeof(w));
+    return hash<size_t>::operator()(w); // this is basically a no-op
```
lucasccordeiro: Add an explanation of why we need this no-op.

lucasccordeiro (approved):
I have just minor comments here.

src/goto-programs/goto_convert_functions.cpp
fbrausse: done.

src/goto-programs/goto_convert_functions.cpp
fbrausse: It's for consistency with whatever the C++ stdlib is doing when hashing a size_t. I've added a small remark.

fbrausse:
Well, this doesn't look promising, yet.

[Master](https://github.com/esbmc/esbmc/actions/runs/5952234935) (c9fd291992372451c9ca60b8b25b11b133e59231):
```
  correct:           15257
    correct true:     8463
    correct false:    6794
  incorrect:            61
    incorrect true:     31
    incorrect false:    30
  unknown:           11141
  Score:             22248 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/6075469598) (afa119f9085b6ae88b1f1a235db6c439ad574460):
```
  correct:           15191
    correct true:     8413
    correct false:    6778
  incorrect:            49
    incorrect true:     20
    incorrect false:    29
  unknown:           11219
  Score:             22500 (max: 43043)
```
Plenty of new unknowns, due to symbolic types in places they are not expected. Maybe it makes sense to go for option (b) mentioned in <https://github.com/esbmc/esbmc/issues/1289#issuecomment-1691545397>.

fbrausse:
It does fix the incorrect verdict for unreach-call
- c/aws-c-common/aws_array_list_clean_up_harness_negated.i

though. Not with 30s, but locally with a larger timeout. Seems this PR (same for #1310) is not doing everything wrong, though.

lucasccordeiro:
Hi @fbrausse,

These are the latest results for the master, which differ from the one you reported here.

https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/stats-30s.txt

fbrausse:
Hi Lucas, that's true. It is because this PR is not based on current master. The merge-base between master and the tested afa119f9085b6ae88b1f1a235db6c439ad574460 is f2a977285c6d2fa9a7dd088de35677872d65c2b3, which does not include the latest merged PRs.

fbrausse:
I'll run SV-COMP again with the latest changes to this PR and also on the current master. We may see more timeouts since the internal representation is larger now.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/6030250105) (6f56c855b5f2fbc384840e9d53841065004d06e3):
```
  correct:           15249
    correct true:     8453
    correct false:    6796
  incorrect:            52
    incorrect true:     22
    incorrect false:    30
  unknown:           11158
  Score:             22518 (max: 43043)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/6095957206) (989f048f4ef971f5a691d8d57af92be8d89acbb9):
```
  correct:           15246
    correct true:     8445
    correct false:    6801
  incorrect:            49
    incorrect true:     20
    incorrect false:    29
  unknown:           11164
  Score:             22587 (max: 43043)
```
There are still new unknowns due to symbolic types, but not that many anymore. I'll fix those in this PR.

lucasccordeiro:
Great!

Thank you, @fbrausse.

fbrausse:
@rafaelsamenezes Do you know what's up with the Windows build?

rafaelsamenezes:
> @rafaelsamenezes Do you know what's up with the Windows build?

Some dependency link got broken. I am working on it right now. Should be fixed in a few hours.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/6030250105) (6f56c855b5f2fbc384840e9d53841065004d06e3):
```
  correct:           15249
    correct true:     8453
    correct false:    6796
  incorrect:            52
    incorrect true:     22
    incorrect false:    30
  unknown:           11158
  Score:             22518 (max: 43043)
```
This PR with more ns.follow() [and when symbols in small SCCs get replaced](https://github.com/esbmc/esbmc/actions/runs/6108079592) (d069c08dcdde3db0af809e97170bd93a3fba2328):
```
  correct:           15237
    correct true:     8438
    correct false:    6799
  incorrect:            52
    incorrect true:     22
    incorrect false:    30
  unknown:           11170
  Score:             22491 (max: 43043)
```
This PR [with only more ns.follow()](https://github.com/esbmc/esbmc/actions/runs/6121234997) (a0a47927088badb7768a5298c09f723ed4f08e9c):
```
  correct:           15231
    correct true:     8438
    correct false:    6793
  incorrect:            52
    incorrect true:     22
    incorrect false:    30
  unknown:           11176
  Score:             22485 (max: 43043)
```
Compared to the previous run, d069c08dcdde3db0af809e97170bd93a3fba2328 just changed the irep2 simplifier for pointer_offset2t to also follow symbolic types.

There are a few more timeouts, but no changes to incorrect ones. In memsafety there are some unknown -> timeout and vice versa. These memsafety diffs involving unknowns are due to `dereference failure: memset of memory segment of size 40 with 44 bytes` in c/uthash-2.0.2/uthash_\*_test\*.yml

With only more ns.follow() and not touching thrash_type_symbols() (the current state in this PR and the last numbers printed above), there is less fluctuation due to timeouts. Notable ones where performance changed drastically are in unreach-call:
- 3x timeout -> correct in under 1sec:
  - c/loops/for_infinite_loop_2.yml
  - c/loops/while_infinite_loop_[13].yml
- 2x correct in under 1sec -> timeout:
  - c/recursive/Ackermann02.yml
  - c/hardware-verification-bv/btor2c-lazyMod.vcegar_QF_BV_itc99_b13_p21.c

lucasccordeiro:
Thanks for submitting this PR, @fbrausse!

PR #1304, opened 2023-08-30
[New Feature] Set the classs/contract name for function verification

ChenfengWei0:
1. Add options `class` and `contract`

```sh
  --class cname                         set function's class name
  --contract cname                      set contract name
```

**Example**

- 1.cpp

```c++
#include <cassert>

class overflow
{
public:
  void add(int xx, int yy)
  {
    int z = xx + yy;
  }
};

class overflow2
{
public:
  void add()
  {
    assert(1);
  }
};

int main()
{
  return 0;
}
```

If only set `--function add`, we will get a warning

```sh
ERROR: warning: main symbol `add' is ambiguous
```
This "ERROR" will not terminate the program, however. It shows:

```sh
[Counterexample]


State 1 file 1.cpp line 16 column 5 function add thread 0
Violated property:
  file 1.cpp line 16 column 5 function add
  arithmetic overflow on add
  !overflow("+", xx, yy)


VERIFICATION FAILED
```

**With the `class` name set:**

- `esbmc 1.cpp --function add --class overflow --overflow-check`

```sh
[Counterexample]


State 1 file 1.cpp line 16 column 5 function add thread 0
Violated property:
  file 1.cpp line 16 column 5 function add
  arithmetic overflow on add
  !overflow("+", xx, yy)


VERIFICATION FAILED
```

- `esbmc 1.cpp --function add --class overflow2 --overflow-check`

```sh
VERIFICATION SUCCESSFUL
```

This also works for solidity, with the option `contract`

---

2. Based on 1, we changed the old `--contract` to `--sol`.  
 
We used the `--contract` option to read `.sol` files because at that time we only supported verification of one single contract, but now we support multiple contract verification.

```sh
esbmc --contract ./example1.sol ./example1.solast --function test  // old
esbmc --sol ./example1.sol ./example1.solast --contract Base --function test  // new
```


lucasccordeiro (approved):
This is a very nice feature!

rafaelsamenezes:
What do you think about using namespaces instead? So we could use something like `--function overflow2::add`, this could help supporting templates in the future.

lucasccordeiro:
JBMC uses the approach suggested by @rafaelsamenezes. However, I don't think it is very intuitive for the end user.

@fbrausse and @mikhailramalho: what do you think?

src/clang-c-frontend/clang_c_main.cpp
```diff
@@ -67,6 +67,13 @@ bool clang_c_maint::clang_main()
 
   forall_symbol_base_map(it, context.symbol_base_map, main)
   {
+    // if user provided class/contract name
+    if(!config.cname.empty() && !config.main.empty() && main != "main")
+    {
+      const std::string fmt = "@" + config.cname + "@F@" + main;
+      if((it->second).as_string().find(fmt) == std::string::npos)
```
fbrausse: - No need for parenthesis.
- Are you sure this should check whether the symbol's id has a substring like that? Shouldn't it rather be has_prefix()?

src/esbmc/options.cpp
```diff
@@ -152,6 +155,9 @@ const struct group_opt_templ all_cmd_options[] = {
    {{"function",
      boost::program_options::value<std::string>()->value_name("name"),
      "set main function name"},
+    {"class",
+     boost::program_options::value<std::string>()->value_name("cname"),
+     "set function's class name"},
```
fbrausse: Could you explain a bit more what that means?

src/util/config.cpp
```diff
@@ -110,6 +110,12 @@ bool configt::set(const cmdlinet &cmdline)
   if(cmdline.isset("function"))
     main = cmdline.getval("function");
 
+  if(cmdline.isset("class"))
+    cname = cmdline.getval("class");
+
+  if(cmdline.isset("contract"))
+    cname = cmdline.getval("contract");
```
fbrausse, line 117: I think you removed/renamed the "contract" option.

src/util/config.h
```diff
@@ -119,6 +119,7 @@ class configt
   } ansi_c;
 
   std::string main;
+  std::string cname;
```
fbrausse, line 123: Maybe add a comment.

src/clang-c-frontend/clang_c_main.cpp
```diff
@@ -67,6 +67,13 @@ bool clang_c_maint::clang_main()
 
   forall_symbol_base_map(it, context.symbol_base_map, main)
   {
+    // if user provided class/contract name
+    if(!config.cname.empty() && !config.main.empty() && main != "main")
```
fbrausse: Couldn't there be a class/namespace/etc containing a function called "main"?

fbrausse (commented):
I don't know Solidity very well, so I can't comment on the utility of namespace vs. class.

Besides a few comments below, I would like to point out that while this implementation might work for Solidity, for nested classes / namespaces in e.g. C++ it's not intuitive for me how it would work. Is it intended for also work for C++?

ChenfengWei0:
> I don't know Solidity very well, so I can't comment on the utility of namespace vs. class.
> 
> Besides a few comments below, I would like to point out that while this implementation might work for Solidity, for nested classes / namespaces in e.g. C++ it's not intuitive for me how it would work. Is it intended for also work for C++?

It is intended for Solidity but I also want to support C++. I just tested the namespaces and nested classes and it worked.

E.g.
```c++
#include <cassert>

namespace first_space
{
  void func()
  {
     assert(1);
  }
}
 
namespace second_space
{
  void func()
  {
     assert(0);
  }
}
using namespace first_space;
int main ()
{
  func();
  return 0;
}
```

```sh
esbmc 1.cpp --function func --class first_space
VERIFICATION SUCCESSFUL

esbmc 1.cpp --function func --class second_space
VERIFICATION FAILED
```
But it is indeed not very intuitive. Maybe we need to rename the option and add more descriptions?

src/clang-c-frontend/clang_c_main.cpp
ChenfengWei0: My mistake. Will fix.

src/util/config.cpp
ChenfengWei0: Basically, I replaced the old `contract` option with `sol` and added the new `contract` for setting the contract name.

src/clang-c-frontend/clang_c_main.cpp
ChenfengWei0: I think they might not share the same prefix

E.g.
```
// Classes
c:@S@overflow@F@add#I#I#
c:@S@overflow2@F@add#

// Contract
c:@C@Base@F@test#14
c:@C@Derived@F@test#28

// Namespace
c:@N@first_space@F@func#
c:@N@second_space@F@func#

// Nested class
c:@S@Enclosing@F@func#
c:@S@Enclosing@S@Nested@F@func#*$@S@Enclosing#
```

src/esbmc/options.cpp
ChenfengWei0: The idea is to let the user set the name of the class in which the function is inside

fbrausse: I meant in the comment for the help message.

src/util/config.cpp
fbrausse: So this `cmdline.isset("contract")` can never happen, yes?

src/esbmc/options.cpp
ChenfengWei0: Oh yes, I have changed it to "set the class/namespace name where the function is inside".

src/util/config.cpp
ChenfengWei0: I think it will happen. Basically, the "contract" is the "class" in solidity. So in solidity, instead of setting `--class xx`, we set `--contract xx`. This is how it will happen

fbrausse: Sorry, my mistake, didn't read the changes to the options carefully enough.

PR #1305, opened 2023-08-31
Interval singleton propagation

rafaelsamenezes:
This enables computing expressions over the abstract domain of intervals. If the result of an computation is a singleton then we can safely replace it with its contents. This has the benefit of being kinda of a constant propagation and a smarter simplifier. 

There is the interesting bit that this can be "tuned": it can be fast/imprecise or slow/precise by playing with the interval analysis flags. For example:

```c
int main() {
  int a = 1;
  int b = 2;
  int c = a + b;

  for(int i = 0; i < b; i++);
  return 0;  
}
```

With just plain interval analysis will become:

```c
int main() {
  int a = 1;
  int b = 2;
  int c = 3;

  for(int i = 0; i < c; i++);
  return 0;  
}
```

While interval analysis and interval arithmetic will become:

```c
int main() {
  int a = 1;
  int b = 2;
  int c = 3;

  for(int i = 0; i < 3; i++);
  return 0;  
}
```

~Note: This is still a draft because there is a bug in the arithmetic that I need to fix before this can be merged. Right now we need to also activate interval extrapolation to get any results with arithmetic.~

There was a bug in the typecast that its now fixed.


rafaelsamenezes:
Master (30s):

```
2023-09-04T15:39:28.0836440Z Statistics:          10224 Files
2023-09-04T15:39:28.0839864Z   correct:            4355
2023-09-04T15:39:28.0841166Z     correct true:     2719
2023-09-04T15:39:28.0841744Z     correct false:    1636
2023-09-04T15:39:28.0842269Z   incorrect:            23
2023-09-04T15:39:28.0843049Z     incorrect true:     20
2023-09-04T15:39:28.0843615Z     incorrect false:     3
2023-09-04T15:39:28.0844768Z   unknown:            5846
2023-09-04T15:39:28.0845355Z   Score:              6386 (max: 17014)
```

PR (30s):

```
2023-09-02T09:13:29.2744547Z Statistics:          10224 Files
2023-09-02T09:13:29.2745839Z   correct:            4358
2023-09-02T09:13:29.2746476Z     correct true:     2726
2023-09-02T09:13:29.2766500Z     correct false:    1632
2023-09-02T09:13:29.2767753Z   incorrect:            22
2023-09-02T09:13:29.2768280Z     incorrect true:     20
2023-09-02T09:13:29.2768800Z     incorrect false:     2
2023-09-02T09:13:29.2769256Z   unknown:            5844
2023-09-02T09:13:29.2769768Z   Score:              6412 (max: 17014)
```

lucasccordeiro (approved):

src/goto-programs/abstract-interpretation/interval_analysis.cpp
```diff
@@ -39,18 +39,81 @@ static inline void simplify_guard(expr2tc &expr, const interval_domaint &state)
...
 
+static void optimize_expression(expr2tc &expr, const interval_domaint &state)
+{
+  ///////////////////
+  // PRECONDITIONS //
+  ///////////////////
+  if(is_nil_expr(expr))
+    return;
+
+  // We can't simplify addr-of sub-expr
```
fbrausse: Why not?

src/goto-programs/abstract-interpretation/interval_analysis.cpp
```diff
@@ -39,18 +39,81 @@ static inline void simplify_guard(expr2tc &expr, const interval_domaint &state)
...
+  if(is_address_of2t(expr))
+    return;
+
+  // We can't replace the LHS of an assignment
```
fbrausse: Seems somewhat clear, but the comment isn't really helpful in explaining anything.

fbrausse (approved):
> This enables computing expressions over the abstract domain of intervals.

LGTM. But I have two comments (besides the nitpicks below):
- Why are intervals an "abstract domain"? What's abstract about them?
- While I do appreciate commented code, could we maybe reduce the
  ```
  ///////////////
  // SCREAMING //
  ///////////////
  ```
  a bit?

src/goto-programs/abstract-interpretation/interval_analysis.cpp
rafaelsamenezes: I will add a comment. 

But `int x = 3; int *ptr = &x;` would become `int x = 3; int *ptr = &3;`

fbrausse: Yeah, that wouldn't make sense. But `&*ptr` could be simplified. However, I'm not entirely clear in what way this overlaps with what the irep2 simplifier is already doing.

rafaelsamenezes:
> Why are intervals an "abstract domain"? What's abstract about them?

Not sure if I can give a good answer.

Its an abstract domain because its used for abstract interpretation.  I'm not reimplementing all the goto operations for intervals, I am just using the ones from the "abstract domain" implementation.

In a book, what makes it an abstract domain is that:
-  its a lattice (-infinity is the lowest, +infinity is the biggest)
- has the transform operation. An abstract interpreter will invoke the transform function over the interval domain
- has the meet operator: set hull.
- has the weaker operator: is subset.

fbrausse:
Thanks, Rafael.

src/goto-programs/abstract-interpretation/interval_analysis.cpp
rafaelsamenezes: Hm, that could be done with the value set analysis.

src/goto-programs/abstract-interpretation/interval_analysis.cpp
rafaelsamenezes: I have added an example in the comment

src/goto-programs/abstract-interpretation/interval_analysis.cpp
rafaelsamenezes: I have updated the comment.

src/goto-programs/abstract-interpretation/interval_analysis.cpp
fbrausse: Many thanks, that makes it much clearer for readers who are not familiar with this code!

PR #1309, opened 2023-09-04
[Solidity] Verify the contract as a whole

ChenfengWei0:
The idea is to verify the properties that must be held at all times.

e.g. consider a 2D grid:

```solidity
contract Robot {
    int x = 0;
    int y = 0;

    constructor() {}

    function moveLeftUp() public {
        --x;
        ++y;
    }

    function moveLeftDown() public {
        --x;
        --y;
    }

    function moveRightUp() public {
        ++x;
        ++y;
    }

    function moveRightDown() public {
        ++x;
        --y;
    }

    function inv() public view {
    // 1. hold in function mode, where we verify this function only
    // 2. fail in contract mode, as we may call any other functions beforehand
    //   for example, the user calls: 
    //    inv()
    //    => test()   
    //    => moveLeftDown()   
    //    => inv()  fail

        assert(x == 0 && y == 0);  
    }

    function test() public view {
        assert((x + y) % 2 == 0); // this should always hold
    }
}
```

In practice, these assertions(Invariant) stand for the certain conditions that we want to make sure are not violated throughout the blockchain transaction. 

Usage: 
- Previously, the esbmc-solidity front-end forced the `function` option to be enabled. 
- And in this contract verification mode, we only need to set the `contract` to be verification

```sh
esbmc contract.sol contract.solast --contract Robot--color --k-induction # no function set
```
For now, this feature only works for a single contract.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3477,4 +3484,163 @@ void solidity_convertert::convert_type_expr(
...
+  const std::string id = prefix + contractName;
+  if(context.find_symbol(id) == nullptr)
+    return true;
+
+  const symbolt &contract = *context.find_symbol(id);
+
+  // a contract should be a struct
+  assert(contract.type.is_struct());
+
+  // 1. call constructor()
```
lucasccordeiro: Can you please explain why you must perform steps 1, 2 and 3?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3477,4 +3484,163 @@ void solidity_convertert::convert_type_expr(
...
+      continue;
+
+    if(context.find_symbol(func_id) == nullptr)
+      return true;
+
+    const symbolt &func = *context.find_symbol(func_id);
+
+    // construct ifthenelse
+
+    // guard: nondet_uint()
```
lucasccordeiro: why do you need nondet_uint? Why not nondet_ushort?

lucasccordeiro (changes requested):
Can you add further comments to this code?

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: My mistake. Will fix.

ChenfengWei0: My mistake. Will fix.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: Will add more documentations to explain.

ChenfengWei0:
> Can you add further comments to this code?


The idea is to convert  
```solidity
Contract Base
{
    constrcutor(){}
    function A(){}
    function B(){}
}
```

to something like

```c++
void sol_main()
{
    Base()  // constructor call
    while(nondet_int())
     {
         if(nondet_int()) A();
         if(nondet_int()) B();
      }
}

```

lucasccordeiro:
Great! Thanks, @ChenfengWei0.

Please add some comments to the code, e.g., including this example or its generalization  ;-)

lucasccordeiro:
@ChenfengWei0: can you rewrite this code as?

````C
void sol_main()
{
    Base()  // constructor call
    while(nondet_bool())
     {
         if(nondet_bool()) A();
         if(nondet_bool()) B();
      }
}
````

lucasccordeiro:
`nondet_bool()` should be more efficient.

ChenfengWei0:
> `nondet_bool()` should be more efficient.

Thanks for the suggestions!🙂

fbrausse:
> The idea is to verify the properties that must be held at all times.
> [...]
> `assert((x + y) % 2 == 0); // this should always hold`

I am a bit confused by the requirement and why it should hold. Please forgive me if the question does not make sense in the context of Solidity, but for regular C/C++ this condition would not be true in presence of multi-threading due to race conditions. Is that a thing for Solidity?

ChenfengWei0:
> > The idea is to verify the properties that must be held at all times.
> > [...]
> > `assert((x + y) % 2 == 0); // this should always hold`
> 
> I am a bit confused by the requirement and why it should hold. Please forgive me if the question does not make sense in the context of Solidity, but for regular C/C++ this condition would not be true in presence of multi-threading due to race conditions. Is that a thing for Solidity?

As far as I know, Solidity is not a multi-threaded programming language, so there is no interleaving between each function.

fbrausse:
> As far as I know, Solidity is not a multi-threaded programming language, so there is no interleaving between each function.

Thanks for clarifying. Does that mean the solidity framework wouldn't necessarily produce the correct results with the `--data-races-check` option or is it just not possible to produce a data race with the current mono-language ESBMC inputs? Should we maybe include a warning and/or override if given by the user? Sorry for being offtopic, I was just wondering and can open a new issue about it if this makes sense.

ChenfengWei0:
> > As far as I know, Solidity is not a multi-threaded programming language, so there is no interleaving between each function.
> 
> Thanks for clarifying. Does that mean the solidity framework wouldn't necessarily produce the correct results with the `--data-races-check` option or is it just not possible to produce a data race with the current mono-language ESBMC inputs? Should we maybe include a warning and/or override if given by the user? Sorry for being offtopic, I was just wondering and can open a new issue about it if this makes sense.

Thanks for raising this issue. Will check!

ChenfengWei0:
> > As far as I know, Solidity is not a multi-threaded programming language, so there is no interleaving between each function.
> 
> Thanks for clarifying. Does that mean the solidity framework wouldn't necessarily produce the correct results with the `--data-races-check` option or is it just not possible to produce a data race with the current mono-language ESBMC inputs? Should we maybe include a warning and/or override if given by the user? Sorry for being offtopic, I was just wondering and can open a new issue about it if this makes sense.

Here is my investigation:
1. There seems to be no actual race conditions in the solidity.
2. I have run some tests with the `--data-race-check` option and the results seem correct
3. However, some blogs and papers say that solidity does need "mutex"( a concept rather than a keyword) in some situations, to prevent reentrancy attacks. Not sure if it would be affected by ESBMC dead-lock options.

I have set up issue #1311  for further investigation.

fbrausse:
Perfect, many thanks for looking into it.

lucasccordeiro (approved):
Thanks for addressing our comments, @ChenfengWei0.

PR #1310, opened 2023-09-04
Thrash type symbols only for types earlier in topological ordering

fbrausse:
This is an alternative to #1303, which replaces the current algorithm of thrash_type_symbols() instead of just augmenting it.

As the types in the context form a possibly cyclic graph, the order of replacements of symbolic types (the edges in the type graph) is important. In short, the implementation in this PR replaces only those symbolic types under pointers which are *not* part of the current strongly connected component (SCC). It accomplishes that by using the topological ordering of the graph induced by the SCCs of the original type graph.

The effect is that the presence of symbolic types changes in two ways:
- some types under pointers are resolved now (if not cyclic)
- types in some expressions are not resolved now if they are a subtype of a cyclic type

In particular the last point required me to add explicit references to the namespace to some parts in value-set and dereference, since there expressions are being constructed into symbols of possibly cyclic types. I am not sure that I caught all cases, though. One way to be sure is to follow @mikhailramalho's earlier suggestion in https://github.com/esbmc/esbmc/pull/1206#discussion_r1269834049: make the namespace parameter to the type_byte_size*() family of functions non-optional.


mikhailramalho:

This PR has been open since 2023-09-11 (still marked as draft) with no activity. The underlying bug (issue #1289 "Symbolic type id in type_size_t::size_bits_expr") is closed. The type-symbol thrashing algorithm has since evolved differently in master.

PR #1314, opened 2023-09-05
[CUDA verification] Support for more TCS

XLiZHI:
1. Data races checks for Pointers are enabled.

2. Correct the verification result of [00_race15](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-unix/00_race15).

3. More supported CUDA TCs are enabled.

So far, the varification accuracy of CUDA is 130/154 = 84.4%

### DIFF
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -36,7 +36,7 @@ void rw_sett::read_write_rec(
     const symbolt *symbol = ns.lookup(symbol_expr.get_identifier());
     if(symbol)
     {
-      if(!symbol->static_lifetime /*&& expr.type().id()=="pointer"*/)
+      if(!symbol->static_lifetime && !expr.type().is_pointer())
       {
         return; // ignore for now
       }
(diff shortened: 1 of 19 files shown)


src/goto-programs/rw_set.cpp
```diff
@@ -36,7 +36,7 @@ void rw_sett::read_write_rec(
     const symbolt *symbol = ns.lookup(symbol_expr.get_identifier());
     if(symbol)
     {
-      if(!symbol->static_lifetime /*&& expr.type().id()=="pointer"*/)
+      if(!symbol->static_lifetime && expr.type().id() != "pointer")
```
mikhailramalho: you can probably use is_pointer_type() here (or `type().is_pointer()`, I don't remember exactly).

mikhailramalho (commented):
I see the functions are commented on all test cases, is that correct? i.e., `//k <<<GRIDDIM, BLOCKDIM>>>(dev_a);`

XLiZHI:
> I see the functions are commented on all test cases, is that correct? i.e., `//k <<<GRIDDIM, BLOCKDIM>>>(dev_a);`

Yes, this is correct, the comment is CUDA program original invokes the kernel function, we will change it for a custom kernel function call methods in OM, such as `ESBMC_verify_kernel (k, GRIDDIM BLOCKDIM, dev_a);`

We keep this comment here for reference.

src/goto-programs/rw_set.cpp
XLiZHI: done

lucasccordeiro (approved):

lucasccordeiro:

PR #1318, opened 2023-09-05
[SMT backend] Added Bitwutzla 0.2.0 to ESBMC

### LINKED ISSUE #1302: error while loading shared libraries: libbitwuzlabb.so: No such file or directory
Steps to reproduce:
1. install bitwuzla into some local prefix directory
    ```
    git clone -b 0.1.1 git@github.com:bitwuzla/bitwuzla.git bitwuzla-0.1.1 &&
    cd bitwuzla-0.1.1 &&
    mkdir prefix &&
    ./configure.py --shared --prefix $PWD/prefix &&
    ninja -C build install &&
    export BITW=$PWD/prefix
    ```
2. configure ESBMC to use this directory
    ```
    cd esbmc &&
    mkdir build &&
    cd build &&
    cmake .. -DBitwuzla_DIR=${BITW} -DBUILD_STATIC=OFF &&
    ninja
    ```
3. run `src/esbmc/esbmc`:
    ```
    src/esbmc/esbmc: error while loading shared libraries: libbitwuzlabb.so: cannot open shared object file: No such file or directory
    ```

`ldd src/esbmc/esbmc | grep "not found"` reports (for me)
```
	libbitwuzlabb.so => not found
	libbitwuzlabv.so => not found
	libbitwuzlals.so => not found
```

These libraries are in ${BITW}/lib64 for me, but the rpath is not set. This most likely is the meson-related problem https://github.com/mesonbuild/meson/issues/2567 which they claim has been fixed in v0.55 via https://github.com/mesonbuild/meson/pull/7103 though I don't understand how to apply whatever the fix finally might have been.

This is a workaround:
```
patchelf --add-rpath '$ORIGIN/../lib64' ${BITW}/lib64/libbitwuzla.so
```

It looks like this issue is one of Bitwuzla's build system and not of ours. But I've opened it here anyway in case someone has the same problem. We should follow up with the Bitwuzla folks about this.

lucasccordeiro:


BUILDING.md
```diff
@@ -20,7 +20,7 @@ Before starting, note that ESBMC is mainly distributed under the terms of the [A
 | MathSAT   | no       | 5.5.4           |
 | Yices     | no       | 2.6.1           |
 | Z3        | no       | 4.8.9           |
-| Bitwuzla  | no       | 0.1.1           |
+| Bitwuzla  | no       | 0.2.0           |
```
fbrausse, line 23: I didn't check: does the additional parameter to `bitwuzla_print_formula()` below mandate that we bump the min-version?

fbrausse (approved):
LGTM, thanks Lucas. Just one question:

rafaelsamenezes (approved):

lucasccordeiro:
Thanks, @fbrausse.

I have just added a comment to explain the magic number 2 :-)

"Print formulas using binary bit-vector output format"

PR #1319, opened 2023-09-06
Interval Analysis assumptions now only uses local symbols

rafaelsamenezes:
Previously, the Interval Analysis assumptions contained all the symbols in the function. These assumptions were added using some heuristic. This PR:

1. Fixes an issue that assumptions were not instrumented correctly for loops. `1: IF X GOTO Y` used to be instrumented as `ASSUME state; 1:IF X GOTO Y;` while the correct instrumentation is `1: ASSUME state; IF X GOTO Y;` . This was making the invariant useless for the k-induction algorithm.
2. Now the assumption is only added before IFs, Asserts, and Assumes. The main difference is that previosly we also instrumented before/after function calls and after an IF.

### DIFF
--- a/src/goto-programs/abstract-interpretation/interval_analysis.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_analysis.cpp
@@ -92,8 +92,6 @@ void instrument_intervals(
   const ait<interval_domaint> &interval_analysis,
   goto_functiont &goto_function)
 {
-  std::unordered_set<expr2tc, irep2_hash> symbols;
-
   // Inline optimizations
   Forall_goto_program_instructions(i_it, goto_function.body)
   {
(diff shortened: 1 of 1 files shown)


rafaelsamenezes:
Master (30s):

```
2023-09-02T09:13:29.2744547Z Statistics:          10224 Files
2023-09-02T09:13:29.2745839Z   correct:            4358
2023-09-02T09:13:29.2746476Z     correct true:     2726
2023-09-02T09:13:29.2766500Z     correct false:    1632
2023-09-02T09:13:29.2767753Z   incorrect:            22
2023-09-02T09:13:29.2768280Z     incorrect true:     20
2023-09-02T09:13:29.2768800Z     incorrect false:     2
2023-09-02T09:13:29.2769256Z   unknown:            5844
2023-09-02T09:13:29.2769768Z   Score:              6412 (max: 17014)
```

Results of this PR (30s):

```
2023-09-06T14:28:05.5425562Z Statistics:          10224 Files
2023-09-06T14:28:05.5427542Z   correct:            4363
2023-09-06T14:28:05.5428692Z     correct true:     2723
2023-09-06T14:28:05.5429367Z     correct false:    1640
2023-09-06T14:28:05.5429792Z   incorrect:            21
2023-09-06T14:28:05.5430452Z     incorrect true:     18
2023-09-06T14:28:05.5430918Z     incorrect false:     3
2023-09-06T14:28:05.5431344Z   unknown:            5840
2023-09-06T14:28:05.5444534Z   Score:              6462 (max: 17014)
```

mikhailramalho:
btw, you can get the loop variables from the the loopst class, if you use it to find all the loops in a function, like the k-induction algorithm.

rafaelsamenezes:
> btw, you can get the loop variables from the the loopst class, if you use it to find all the loops in a function, like the k-induction algorithm.

Hm, using the loop variables seem to make more sense during the k-induction transformation. We can even be more precise and add the loop termination invariant as an assert. What do you think?

lucasccordeiro (approved):

PR #1321, opened 2023-09-06
[Solidity] Support "require" keywords

ChenfengWei0:
The keyword "require" is virtually identical to "assume"

E.g.

```solidity
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.8.0;

contract Robot {
    int x = 0;
    int y = 0;

    constructor() {}

    function moveLeftUp() public {
        --x;
        ++y;
    }

    function moveLeftDown() public {
        --x;
        --y;
    }

    function moveRightUp() public {
        ++x;
        ++y;
    }

    function moveRightDown() public {
        ++x;
        --y;
    }

    function test2() public {
        require(x == 0);
        require(y == 12);  // if "require(y==2)", SMTCheccker could give correct answer
        assert(x == 0 && y == 0);  // could reach and should fail
    }

    function test() public {
        assert((x + y) % 2 == 0);
    }
}
```

I have tested this case with other SMT verifiers for solidity, including the SMTChecker (built inside the solidity compiler) and Mythril. Only Esbmc-Solidity could give the correct result.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2071,8 +2078,7 @@ bool solidity_convertert::get_decl_ref_builtin(
   code_typet convert_type;
   typet return_type;
   if(
-    decl["name"] == "assert" || decl["name"] == "__ESBMC_assume" ||
-    decl["name"] == "__VERIFIER_assume")
+    name == "assert" || name == "__ESBMC_assume" || name == "__VERIFIER_assume")
```
lucasccordeiro, line 2081: Should ` name == "assert"` be ` name == "assume"`?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2057,11 +2058,17 @@ bool solidity_convertert::get_decl_ref_builtin(
   // -ve ref id means built-in functions or variables.
   // Add more special function names here
   assert(
-    decl["name"] == "assert" || decl["name"] == "__ESBMC_assume" ||
-    decl["name"] == "__VERIFIER_assume");
+    (decl["name"] == "assert" || decl["name"] == "require" ||
```
lucasccordeiro, line 2061: Shall we also add `assume` here?

lucasccordeiro (changes requested):

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: The `name == assert` is for assert keywords in Solidity. Here “assume”(which is actually "require") has already been converted to "__ESBMC_assume", thus we have `name == "__ESBMC_assume"`.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: I don't think so. Because there is no "assume" keyword in Solidity -- they use "require" instead, thus we have `decl["name"] == "require"`

lucasccordeiro:
Thanks for these clarifications, @ChenfengWei0.

lucasccordeiro (approved):

PR #1323, opened 2023-09-07
Turn smtlib build optional

rafaelsamenezes:
They are dependencies for using the smtlib interface which:

- Makes windows builds depends on a third-party implementation of them.
- Should not be mandatory.


lucasccordeiro (approved):

fbrausse:
If the smtlib backend is going to be optional, there should be a switch to enable or disable it. Building it depending on whether flex/bison is found on the host OS alone is never a good idea: Can't be sure of the build configuration on an unknown system + it's near impossible to give reliable instructions how to build ESBMC with the smtlib backend enabled.

fbrausse:
Also, https://github.com/esbmc/esbmc/blob/c9cd2646987a91ad0b96ab287aee0c990a30a3fa/src/solvers/solve.cpp#L21 assumes that the smtlib backend is always available.

lucasccordeiro (changes requested):
Shall we update the building instructions?

fbrausse (approved):
LGTM.

lucasccordeiro (approved):

PR #1326, opened 2023-09-10
Keep benchmark definition file inside esbmc repo

rafaelsamenezes:
I mistakenly pushed some patches into ESBMC master that adds a field to set the git commit/tag of sv-benchmarks to be executed. However, the benchmark definition file might change based on the version of the benchmarks. Here I am adding the default one that we used during the competition.


fbrausse (approved):
That looks good to me. Thank you.

lucasccordeiro (approved):

lucasccordeiro:

PR #1362, opened 2023-09-20
[Concurrency] optimized static analysis for pthread_create

### LINKED ISSUE #1255: [data-races-check] unsupported cases
`esbmc main.c --data-races-check`  same if add `--no-por`

```
#include <pthread.h>

void* t1(void *arg) {
    int* shared = (int*) arg;
    *shared = 4;                 //race here

    return NULL;
}

void* t2(void *arg) {
    int* shared = (int*) arg;
    *shared = 2;                //race here

    return NULL;
}

int main() {
    pthread_t id1, id2;
    int shared_data = 0;

    pthread_join(id1, NULL);
    pthread_join(id2, NULL);

    return 0;
}
```
There is a data race in the above code, I checked our TCs about data race check, I found that this case is not covered.

By the way, CUDA OM is implemented based on this formalism, and I think this is why we failed to detect data races.

### LINKED ISSUE #1324: [concurrency]  dereference failed in data races check
`esbmc main4.c --data-races-check`
```
#include <pthread.h>

void* t1(void *arg) {
    int* shared = (int*) arg;
    *shared = 4;                          <-------dereference
}

int main() {
    pthread_t id1;
    int shared_data = 0;

    return 0;
}
```
When performing a data race check, for a dereferencing operation, we should get the object that the pointer points to, for example, in the code above:

`*shared -----> shared_data` 

We currently handle dereference incorrectly in master：
https://github.com/esbmc/esbmc/blob/c9cd2646987a91ad0b96ab287aee0c990a30a3fa/src/goto-programs/rw_set.cpp#L86-L98

As shown in the above code, after converting `expr` to `expr2tc`, we obtain:

```
symbol
* name : c:main4.c@48@F@t1@shared
* renamelev : Level 0
* level1_num : 0
* level2_num : 0
* thread_num : 0
* node_num : 0
* type : pointer
  * subtype : signedbv
      * width : 32
```
The id of this expr2tc is changed to symbol, which means that the `dereference` cannot continue:
https://github.com/esbmc/esbmc/blob/c9cd2646987a91ad0b96ab287aee0c990a30a3fa/src/pointer-analysis/dereference.cpp#L149-L152

XLiZHI:
This PR deals with `pthread_create` in the static analysis.

Static analysis fails to set the value set correctly when we run the method on the child thread using `pthread_create`. Therefore, when we are dealing with a function call, we need to extract its arguments to get `func` and `arg`, then call `func(arg)` to get value sets.

```
int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
    void *(*start_routine)(void*), void *arg);
```
for this TC:
```
#include <pthread.h>

void* t1(void *arg) {
    int* shared = (int*) arg;
    *shared = 4;

    return NULL;
}

int main() {
    pthread_t id1;
    int shared_data = 1;

    return 0;
}
```

`esbmc main.c --show-goto-value-set`

master 57d538ea93ad58343a1c089a04e5192c1cd2c5e3:
```
c:main4.c@48@F@t1@shared = { } 
```
this patch:
```
c:main4.c@48@F@t1@shared = { <shared_data, 0, signed int> } 
```

**1. Do we need to add data races assertions for all goto functions?**

I don't think so, we fixed dereference at goto level in this patch, now we take care of all dereference in the code (including various libraries), adding unnecessary assertions will increase the verification time.

**2. How do we identify goto functions that need to add data race assertions?**

Since data races occur only between threads, we get the called and related functions from `pthread_create`.


mikhailramalho (changes requested):
This patch is going to need some work before we can land it.

First of all, could you explain what's the motivation for these changes? It helps if you write a summary in the PR. Why are we ignoring everything that's not pthred_create or "c:main6.c@16@F@printValue@p"?
Secondly, why are we changing the arguments and passing the new ones to `do_function_call`?
Thirdly, remove unnecessary comments and debug code from the PR.

XLiZHI:
> This patch is going to need some work before we can land it.
> 
> First of all, could you explain what's the motivation for these changes? It helps if you write a summary in the PR. Why are we ignoring everything that's not pthred_create or "c:main6.c@16@F@printValue@p"? Secondly, why are we changing the arguments and passing the new ones to `do_function_call`? Thirdly, remove unnecessary comments and debug code from the PR.

Hi @mikhailramalho, This PR isn't complete yet, I've updated it with a few comments that I hope will help.

mikhailramalho:
> > This patch is going to need some work before we can land it.
> > First of all, could you explain what's the motivation for these changes? It helps if you write a summary in the PR. Why are we ignoring everything that's not pthred_create or "c:main6.c@16@F@printValue@p"? Secondly, why are we changing the arguments and passing the new ones to `do_function_call`? Thirdly, remove unnecessary comments and debug code from the PR.
> 
> Hi @mikhailramalho, This PR isn't complete yet, I've updated it with a few comments that I hope will help.

Okay, I noticed that you marked it as a draft now. Also, remember not to request reviews in these cases either.

src/goto-programs/rw_set.cpp
```diff
@@ -36,7 +36,9 @@ void rw_sett::read_write_rec(
     const symbolt *symbol = ns.lookup(symbol_expr.get_identifier());
     if(symbol)
     {
-      if(!symbol->static_lifetime && !expr.type().is_pointer())
+      if(
+        !symbol->static_lifetime &&
+        expr.type().subtype().id() != "dereferenced_symbol")
```
lucasccordeiro, line 41: Why should we ingore `` expr.type().subtype().id() != "dereferenced_symbol"''? can you please add some explanation here?

src/goto-programs/rw_set.cpp
```diff
@@ -88,11 +90,12 @@ void rw_sett::read_write_rec(
...
 
-    exprt tmp(expr.op0());
     expr2tc tmp_expr;
-    migrate_expr(tmp, tmp_expr);
+    migrate_expr(expr, tmp_expr);
     dereference(target, tmp_expr, ns, value_sets);
-    tmp = migrate_expr_back(tmp_expr);
+
+    exprt tmp = migrate_expr_back(tmp_expr);
+    tmp.type().subtype() = typet("dereferenced_symbol");
```
lucasccordeiro, line 101: Why must we introduce a new subtype in ESBMC ``dereferenced_symbol''?

src/goto-programs/static_analysis.cpp
```diff
@@ -52,6 +52,75 @@ void static_analysis_baset::operator()(const goto_programt &goto_program)
...
+  {
+    insert_idt(goto_functions, idts, identifier);
+  }
+}
+
+void static_analysis_baset::insert_idt(
+  const goto_functionst &goto_functions,
+  std::set<irep_idt> &idts,
+  const irep_idt &idt)
+{
```
lucasccordeiro, line 72: Can you document these steps?

src/goto-programs/static_analysis.cpp
```diff
@@ -52,6 +52,75 @@ void static_analysis_baset::operator()(const goto_programt &goto_program)
...
+    locationt to_l = *it;
+    // Recursively fetches the function calls inside and adds them to the set,
+    // func1{func2}->func2{func3}->fun3(),
+    // add idts of func1, func2, func3.
+    if(l->is_function_call())
+    {
+      const code_function_call2t &code = to_code_function_call2t(l->code);
+      irep_idt identifier = to_symbol2t(code.function).get_symbol_name();
+
+      // We should ignore all function calls from libraries
```
lucasccordeiro: Explain why we must ignore all function calls from libraries. Why do we consider only `pthread'?

src/goto-programs/static_analysis.cpp
```diff
@@ -300,6 +369,28 @@ void static_analysis_baset::do_function_call_rec(
...
     irep_idt identifier = to_symbol2t(function).get_symbol_name();
 
+    std::vector<expr2tc> new_args;
+
+    // When calling 'pthread_create' method, we need to make adjustments
+    // to properly make VSA understand the func we are going to
+    // call on the child thread
+
+    // TODO: A more robust way to find the method
+    if(id2string(identifier) == "c:@F@pthread_create")
```
lucasccordeiro: Shall we use `has_prefix` here?

lucasccordeiro (changes requested):

src/goto-programs/rw_set.cpp
XLiZHI: We're not ignoring `dereferenced_symbol` here, we're actually adding a data races check assertion to the symbol it's `dereferenced_symbol`.

For example:
```
void* t1(void *arg) {
    int* shared = (int*) arg;
    *shared = 4;            
}
int main() {
    pthread_t id1;
    int shared_data = 0;

    return 0;
}
```
The `shared_data` pointed to here is a local variable, and we might **need a way** to get it past the filter in this code.
So I added a subtype named `dereferenced_symbol` to all dereferenced `expr`.

fbrausse:
Just a quick question: you moved some `cuda/benchmarks` regression tests to `cuda/cuda_data_races`, but I can't seem to find this directory in the `regression/CMakeLists.txt`. Are you disabling these tests?

src/goto-programs/rw_set.cpp
lucasccordeiro: I see.

What would be the difference between `dereference` (already in ESBMC) and `dereferenced_symbol`?

src/goto-programs/static_analysis.cpp
XLiZHI: Here we are using a recursive approach, if there is a function call inside, we find the function and double check if there is a function call inside, In fact, data races do not occur in these libraries, and for some libraries, infinite recursion may occur. There's no good way to identify all of them, so for now we'll only consider the following:
```
pthread_mutex_lock( )
pthread_mutex_unlock( )
pthread_exit(NULL)
```

XLiZHI:
> Just a quick question: you moved some `cuda/benchmarks` regression tests to `cuda/cuda_data_races`, but I can't seem to find this directory in the `regression/CMakeLists.txt`. Are you disabling these tests?

These TCs are involved in another issue, and I will address them later.

lucasccordeiro:
@XLiZHI: BTW, can you add more test cases related to data-race check? You can get some from SV-COMP or even enable some regression tests in our suite if they are marked as `KNOWNBUG`.

src/goto-programs/rw_set.cpp
XLiZHI: Well, here dereference means `*shared`, and dereferenced_symbol means `shared_data`.

XLiZHI:
> @XLiZHI: BTW, can you add more test cases related to data-race check? You can get some from SV-COMP or even enable some regression tests in our suite if they are marked as `KNOWNBUG`.

Yes, sure, I also have an open issue about "index", maybe we can add them in the next PR?

lucasccordeiro:
> Yes, sure, I also have an open issue about "index", maybe we can add them in the next PR?

Fine by me.

src/goto-programs/static_analysis.cpp
lucasccordeiro: Thanks for this explanation. Can you improve the documentation in the code?

src/goto-programs/rw_set.cpp
lucasccordeiro: Thanks for this explanation. Can you improve the documentation in the code?

XLiZHI: Sure! I will improve them, thanks for your review.

mikhailramalho: We shouldn't be introducing new types, as it will probably crash esbmc if it ever reaches the backend. To add a new type, we should add it to both irep1 and irep2 + implement the appropriate migration.

So what you want to do is to get the content of a global variable? You can probably do that by checking if the static/global flags are set on the symbol

src/goto-programs/static_analysis.cpp
```diff
@@ -300,6 +378,28 @@ void static_analysis_baset::do_function_call_rec(
...
+      // along with the arg of that func
+      exprt tmp = migrate_expr_back(arguments[2]);
+      identifier = to_symbol_expr(tmp.op0()).get_identifier();
+
+      new_args.push_back(arguments[3]);
+
+      pthread_set.insert(identifier);
+    }
+    else
+      new_args = arguments;
```
mikhailramalho, line 401: My previous comment still stands: we shouldn't be rewriting the function argument; why is this needed?

mikhailramalho (changes requested):

src/goto-programs/static_analysis.cpp
XLiZHI: I explained it in the comments above.  When we encounter `pthread_create (id, NULL, func, arg)`, the function call we need to execute is `func(arg)`, rather than `pthread_create`. Only then can we get the **value set** of `func` correctly.

That is, the argument is special compared to a normal function call, and we need to distinguish between the two cases. In subsequent functions, the argument is retrieved via `locationt`, which means that we are still getting arguments from `pthread_create`, and we need to change that.

https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/pointer-analysis/value_set_domain.cpp#L32-L39

src/goto-programs/rw_set.cpp
XLiZHI: > We shouldn't be introducing new types, as it will probably crash esbmc if it ever reaches the backend. To add a new type, we should add it to both irep1 and irep2 + implement the appropriate migration.

I thought about this, but none of the exprs are passed back to the backend, which only uses this to decide to add assertions to check for data races.

> So what you want to do is to get the content of a global variable? You can probably do that by checking if the static/global flags are set on the symbol

The tricky part is that these symbols are not global variables, as shown above, `shared_data` is only a local variable, but we need to deal with that.

fbrausse (commented):
> > Just a quick question: you moved some `cuda/benchmarks` regression tests to `cuda/cuda_data_races`, but I can't seem to find this directory in the `regression/CMakeLists.txt`. Are you disabling these tests?
> 
> These TCs are involved in another issue, and I will address them later.

If this change to our regression suite is unrelated to the other changes in this PR, we shouldn't disable the tests here. For reference: What other issue are do you mean?

General comments:
- Regarding adding a new "dereferenced_symbol" type: What is it and what meaning does it have as a subtype of the type of some expression? This should be documented in the code. Appearently you're using it to enable something in the recursion. Why does the type of expressions need to change incompatibly for that? `.subtype()` has a very clear meaning for a type: It is any type that a new type (such as arrays or pointers) is built from. Is the original type of the expression ever restored? Can't you just add a parameter to this recursion and test that?

- pthread_create() is special in that (a) it is calling another function through a pointer and passing along arguments, and (b) for now is the only function that can spawn other threads which are relevant for the data-races check. Since this PR specifically targets pthread_create(), I wonder what happens with this:
  ```c
  void my_trampoline(void *(*f)(void *), void *data) { pthread_t t; pthread_create(&t, NULL, f, data); }
  void* t1(void *arg) { /* as before */ }
  int main()
  {
    int shared = 1;
    my_trampoline(t1, &shared);
  }
  ```

In my opinion this data-races check would better be performed during symex.

lucasccordeiro:
> In my opinion this data-races check would better be performed during symex.

I agree; there are too many workarounds to get this check fixed in the goto level.

XLiZHI:
> What other issue are do you mean?

These TCS all contain reference references, and the array index is a variable, which cannot be handled at this time. It passed because dereferencing didn't work correctly.

> Can't you just add a parameter to this recursion and test that?

Thanks for reminding me. Maybe this is another way to go.

> I wonder what happens with this.

I took that into account. We're getting the identifier, which means that it doesn't matter where the `pthread_create` function is called, we're getting its function identifier.

> In my opinion this data-races check would better be performed during symex.

This is the solution I was considering at first, but I'm not familiar with symex. In symex, do we also add assertions in goto functions? It seems like a different representation is used here. I looked at the code, including static analysis, and the value sets all seem to be serving data races check, and porting seems to be a non-trivial job.

Many of the functions in this PR are not actually intended to solve `pthread_create`. We fix the way to get references, the problem that comes with this is that we have to handle dereferences in all goto functions, which makes some unnecessary assertions. We cannot tell which dereference is the one we need, so most of the functions are for filtering, making sure that we only deal with relevant goto functions as much as possible, and I think even we face this problem in `symex`.

src/goto-programs/static_analysis.cpp
mikhailramalho: That's still not the way to do that in esbmc. I suggest you look into how the pthread_trampoline works in esbmc. 

Furthermore, if you need the argument to perform the analysis, this must happen during symbolic execution.

src/goto-programs/rw_set.cpp
mikhailramalho: Why do you need to handle local variable data in multi-threaded code? These do not introduce new interleavings in the esbmc multi-threaded symbolic execution.

XLiZHI: Since we need it as the object of the assertion, it would be violated if it appeared in a different thread. Dereferencing may have different variable names, but the object they point to remains the same.

fbrausse:
> > What other issue are do you mean?
> 
> These TCS all contain reference references, and the array index is a variable, which cannot be handled at this time. It passed because dereferencing didn't work correctly.

So they are known bugs with this PR, yes? Could you please mark them as KNOWNBUG in their respective test.desc files? That's much better than disabling the test cases entirely.

> > Can't you just add a parameter to this recursion and test that?
> 
> Thanks for reminding me. Maybe this is another way to go.

The internal representation is a somewhat sensitive subject. There should only be structural changes to it (e.g., adding new types or expression constructs) when there is a clear need for it. Usually, any change to the irep (both, irep1 and irep2) require touching many other parts of the code in order to support the new constructs.

> > I wonder what happens with this.
> 
> I took that into account. We're getting the identifier, which means that it doesn't matter where the `pthread_create` function is called, we're getting its function identifier.

If this works as well, nice :+1:. Though I don't fully understand why it does.

> > In my opinion this data-races check would better be performed during symex.
> 
> This is the solution I was considering at first, but I'm not familiar with symex. In symex, do we also add assertions in goto functions? It seems like a different representation is used here. I looked at the code, including static analysis, and the value sets all seem to be serving data races check, and porting seems to be a non-trivial job.

I agree, it would be more invasive.

Symex makes a transition from the GOTO language to SSA. The SSA is a sequence of assignments, assertions and assumptions (and a few other constructs, but they are less relevant), all under some condition called a guard. Basically all jumps are gone. If you only consider the assignments, it is quite similar to a straight-line program. Assertions get added there as part of interpreting a specific construct, for instance `goto_symext::symex_free()` adds one here: https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/goto-symex/builtin_functions.cpp#L326

src/goto-programs/rw_set.cpp
mikhailramalho: I didn't understand, can you give an example?

XLiZHI:
```
#include <pthread.h>

void* t1(void *arg) {
    int* shared1 = (int*) arg;

    return NULL;
}

void* t2(void *arg) {
    int* shared2 = (int*) arg;

    return NULL;
}

int main() {
    pthread_t id1, id2;
    int shared_data = 0;

    pthread_join(id1, NULL);
    pthread_join(id2, NULL);

    return 0;
}
```
Sure, as shown in the example above, we are creating two threads to run the functions t1 and t2. As you can see in the two functions we have variables `shared1` and `shared2`, they have different variable names but they both point to `shared_data`. 

Our current approach to data race detection is to use assert to determine if the same identifier exists in multiple threads (e.g., `assert! shared1`), we would not get the correct result. So we need to dereference it and we get the symbol - `shared_data`, So whether it's `shared1` or `shared2`, we use `shared_data` to add assert (`assert! shared_data`)

mikhailramalho: and why we create an `assert !shared1` on the `*shared1 = 4;`?

It should clearly be `assert !*shared1`. Is esbmc ignoring the dereference?

XLiZHI: In the current master goto functions are as follows:
```
        *shared1=4;
        // 247 file main6.c line 5 column 5 function t1
        ASSERT !tmp_c:main6.c@48@F@t1@shared1 // R/W data race on c:main6.c@48@F@t1@shared1
```

```
        *shared2=2;
        // 257 file main6.c line 12 column 5 function t2
        ASSERT !tmp_c:main6.c@168@F@t2@shared2 // R/W data race on c:main6.c@168@F@t2@shared2
```
esbmc does not ignore dereference, but it only takes the identifer of the symbol to add assertions.
https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/goto-programs/rw_set.cpp#L60-L66

In other words, we can also say that currently dereference is not working correctly.
https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/goto-programs/rw_set.cpp#L86

mikhailramalho:
This is what must be fixed then. A couple of questions that should help you
with this task:

1. Why is esbmc ignoring the dereference?
2. Have you tried to verify this program in older versions of esbmc? I have
a feeling that if this ever worked, it might have been broken with changes
to get_base_object.
3. Can esbmc find the property violation when shared_data is a global
variable passed via the pthread_create parameter? What about a static one?
4. Can esbmc find the property violation when shared_data is a global
variable passed without passing it via the pthread_create parameter? What
about a static one?
5. What do the symex and the formula look like for these cases? What should
it look like?
6. What does it look like in cbmc?

XLiZHI:
In fact, this patch has solved this problem:

1. First we checked that our static analysis was correct using `--goto-value-set`, and then we investigated that we couldn't get the value set of the function called via `pthread_create`.   value set :`c:main4.c@48@F@t1@shared = { }` 
I then changed the static analysis to treat `pthread_create` as a normal function call. `c:main4.c@48@F@t1@shared = { <shared_data, 0, signed int> } ` Now we have the value set.

2. I found the reason why dereference doesn't work, which means we didn't handle dereference at all before. More details here https://github.com/esbmc/esbmc/issues/1324#issue-1886072096. I made a change in this patch and now dereference works properly. Since symbols obtained by dereferenced are local variables, we need some way to get it through the filter.

3.  Fixing dereferenced comes with its own set of issues. For example, we'll handle all dereferencing in our code. This can cause a lot of problems, including adding unnecessary assertions, dereferencing errors, and more. This patch proposes a method to obtain the relevant goto function recursively according to the function call through `pthread_create`, so as to avoid processing all goto functions and reduce the verification time.

> Have you tried to verify this program in older versions of esbmc? I have
> a feeling that if this ever worked, it might have been broken with changes
> to get_base_object

As far as I know, this problem has been around since the beginning.

> Can esbmc find the property violation when shared_data is a global
> variable passed via the pthread_create parameter? What about a static one?

No, all functions called via pthread_create fail to get the value set correctly and thus fail to dereference. That's why I'm changing the pthread_create function call in the static analysis.

> Can esbmc find the property violation when shared_data is a global
> variable passed without passing it via the pthread_create parameter? What
> about a static one?

In this case the result is correct, but as I mentioned, the problem is with pthread_create.

> 5. What do the symex and the formula look like for these cases? What should
> it look like?
> 6. What does it look like in cbmc?

I haven't investigated yet.

fbrausse:
I've looked into this code and while I agree that the current dereference is not working, this patch should fix it:
```diff
diff --git a/src/goto-programs/rw_set.cpp b/src/goto-programs/rw_set.cpp
index 9cbcaf7e7f..c536bd99f5 100644
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -88,11 +88,10 @@ void rw_sett::read_write_rec(
     assert(expr.operands().size() == 1);
     read(expr.op0(), guard);
 
-    exprt tmp(expr.op0());
     expr2tc tmp_expr;
-    migrate_expr(tmp, tmp_expr);
+    migrate_expr(expr, tmp_expr);
     dereference(target, tmp_expr, ns, value_sets);
-    tmp = migrate_expr_back(tmp_expr);
+    exprt tmp = migrate_expr_back(tmp_expr);
 
     read_write_rec(tmp, r, w, suffix, guard);
   }
diff --git a/src/pointer-analysis/dereference.cpp b/src/pointer-analysis/dereference.cpp
index f18594f697..68b375971b 100644
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -563,8 +563,12 @@ expr2tc dereferencet::make_failed_symbol(const type2tc &out_type)
 
   // Due to migration hiccups, migration must occur after the symbol
   // appears in the symbol table.
+  namespacet new_ns(new_context);
+  const namespacet *old_ns = std::exchange(migrate_namespace_lookup, &new_ns);
   expr2tc value;
   migrate_expr(symbol_expr(*s), value);
+  migrate_namespace_lookup = old_ns;
+
   return value;
 }
 
```
Now, the pthread_create() model makes a call to the intrinsic __ESBMC_set_thread_internal_data(threadid, startdata) where startdata's value-set now would correctly contain
```
c:pthread_lib.c@5413@F@pthread_create@startdata.func = { <t1, 0, void * (void *)> } 
c:pthread_lib.c@5413@F@pthread_create@startdata.start_arg = { <shared_data, 0, signed int> }
```
for the initial example.

But the value-set (or rather the static_analysist implementation) does not know about the intrinsic and therefore the value-set for the pointer argument `void *arg` of t1() stays empty. If I understand correctly, messing with the pthread_create() function call here could actually be not such a bad idea since our model uses intrinsics not understood by the value-set analysis. And all that analysis really needs to know is that eventually, `t1(&shared_data)` will be called to track the assignment to the argument.
Still there are better ways than to abuse the `.subtype()`.

An alternative to messing with the call could also be adding an `if (0) start_routine(arg);` statement to the pthread_create() model. This would result in t1()'s value-set correctly containing
```
c:gh-1324.c@48@F@t1@shared = { <shared_data, 0, signed int> } 
c:gh-1324.c@31@F@t1@arg = { <shared_data, 0, signed int> } 
```
though it has downsides as well:
- Once the value-set analysis starts interpreting if-conditions, this will break.
- pthread_create()'s value-set will be merged with t1()'s and result in a bunch of unnecessary symbols being present in it.

fbrausse:
To sum up, here is how I think the concerns by the reviewers can be addressed:
1. Make the regression tests KNOWNBUG instead of disabling them.
2. Do not introduce a new "dereferenced_symbol" type, instead use some flag to communicate whatever is necessary (I didn't fully understand why it's there).
3. In order to address Mikhail's concerns (which I share, to some degree), run this data-races instrumentation not on the original context, but on a copy containing modified pthread_create() symbol, which only forwards the function call. And revert the changes to static_analysis. This way it would be clear what exactly is being done (and with an additional comment, also why).

lucasccordeiro:
Thanks, @fbrausse.

I fully support these suggestions to make this PR mergeable.

XLiZHI:
Thank you all, I agree and understand the first two points, but I don't quite understand the third point.

> run this data-races instrumentation not on the original context, but on a copy containing modified pthread_create() symbol, which only forwards the function call. And revert the changes to static_analysis.

what do you mean by original context and copy, I don't quite understand the operation of this part. Could you give me some details? 

In addition, it seems to me that the value set generated by static analysis does not affect subsequent symbolic execution, it is only used to add data races assertions in the goto function. I don't see static analysis used anywhere else.

https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/esbmc/esbmc_parseoptions.cpp#L1713-L1722

> An alternative to messing with the call could also be adding an if (0) start_routine(arg); statement to the pthread_create() model. This would result in t1()'s value-set correctly containing

This seems like a much simpler approach, would you like me to use it instead of the existing one?

mikhailramalho:
> 2. I found the reason why dereference doesn't work, which means we didn't handle dereference at all before. More details here [[concurrency]  dereference failed in data races check #1324 (comment)](https://github.com/esbmc/esbmc/issues/1324#issue-1886072096). I made a change in this patch and now dereference works properly. Since symbols obtained by dereferenced are local variables, we need some way to get it through the filter.

The change in this patch is not correct for the reasons I've mentioned already. The code in #1324 looks like the correct fix.

> 3. Fixing dereferenced comes with its own set of issues. For example, we'll handle all dereferencing in our code. This can cause a lot of problems, including adding unnecessary assertions, dereferencing errors, and more. 

These are not problems, these are improvements to the verification process. If esbmc is performing an incomplete analysis, it must be fixed.

> As far as I know, this problem has been around since the beginning.
> 

Can you confirm it?

> In this case the result is correct, but as I mentioned, the problem is with pthread_create.
> 

This seems not to be a pthread_create problem, but rather an argument problem. How different is the formula generated when shared_data is a global not passed via the argument?

> > 5. What do the symex and the formula look like for these cases? What should
> >    it look like?

It is important that you understand the symex and the generated formula, so we can identify where things went wrong.

mikhailramalho:
> 3. In order to address Mikhail's concerns (which I share, to some degree), run this data-races instrumentation not on the original context, but on a copy containing modified pthread_create() symbol, which only forwards the function call. And revert the changes to static_analysis. This way it would be clear what exactly is being done (and with an additional comment, also why).

I'm still skeptical that we need a special handle for the pthread_create function; that's why I asked what the data race assertion looks like.

The static analysis code is quite dumb (at least compared to the dereference code), and the current approach is to instrument every function in the code: if the assertion is reachable in a multi-threaded context, we should be able to prove it holds or not.

@XLiZHI we need to understand how the data race is encoded, e.g., do we have auxiliary variables to flag read/write access when checking the interleavings?

fbrausse:
> I'm still skeptical that we need a special handle for the pthread_create function

For the value-set on master it is impossible to make the connection between the parameter `&shared_data` given to pthread_create() and the parameter to `t1(void *arg)`. arg's value-set will just remain empty, that is, the dereference will be a failed symbol. The reason is that the parameter is passed via the `__ESBMC_set_thread_internal_data` intrinsic and received by pthread_trampoline() via the get-version of this intrinsic. This intrinsic is only handled in symex.

Value-set doesn't know about this intrinsic and therefore can't make the connection. So either the general static_analysist or the specific value_set_analysist need to be taught about either the intrinsics, or the higher-level pthread_create/pthread_trampoline functions in order to make this connection.

>  that's why I asked what the data race assertion looks like.

In the t1/t2 example, when the `shared_data` is a global variable, the address of which is taken in t1() and t2(), the instrumented GOTO code looks like this:
```
t1 (c:@F@t1):
        // 385 file gh-1362-1.c line 6 column 5 function t1
        signed int * shared1;
        // 386 file gh-1362-1.c line 6 column 5 function t1
        tmp_c:gh-1362-1.c@66@F@t1@shared1=1;
        // 387 file gh-1362-1.c line 6 column 5 function t1
        shared1=&shared_data;
        // 388 file gh-1362-1.c line 6 column 5 function t1
        tmp_c:gh-1362-1.c@66@F@t1@shared1=0;
        // 389 file gh-1362-1.c line 6 column 5 function t1
        ASSERT !tmp_c:gh-1362-1.c@66@F@t1@shared1 // W/W data race on c:gh-1362-1.c@66@F@t1@shared1
        // 390 file gh-1362-1.c line 7 column 5 function t1
        tmp_c:@shared_data=1;
        // 391 file gh-1362-1.c line 7 column 5 function t1
        *shared1=4;
        // 392 file gh-1362-1.c line 7 column 5 function t1
        tmp_c:@shared_data=0;
        // 393 file gh-1362-1.c line 7 column 5 function t1
        ASSERT !tmp_c:@shared_data // W/W data race on c:@shared_data
        // 394 file gh-1362-1.c line 7 column 5 function t1
        ASSERT !tmp_c:gh-1362-1.c@66@F@t1@shared1 // R/W data race on c:gh-1362-1.c@66@F@t1@shared1
        // 395 file gh-1362-1.c line 9 column 5 function t1
        RETURN: 0
        // 396 file gh-1362-1.c line 10 column 1 function t1
        END_FUNCTION // t1

t2 (c:@F@t2):
        // 397 file gh-1362-1.c line 13 column 5 function t2
        signed int * shared2;
        // 398 file gh-1362-1.c line 13 column 5 function t2
        tmp_c:gh-1362-1.c@194@F@t2@shared2=1;
        // 399 file gh-1362-1.c line 13 column 5 function t2
        shared2=&shared_data;
        // 400 file gh-1362-1.c line 13 column 5 function t2
        tmp_c:gh-1362-1.c@194@F@t2@shared2=0;
        // 401 file gh-1362-1.c line 13 column 5 function t2
        ASSERT !tmp_c:gh-1362-1.c@194@F@t2@shared2 // W/W data race on c:gh-1362-1.c@194@F@t2@shared2
        // 402 file gh-1362-1.c line 14 column 5 function t2
        tmp_c:@shared_data=1;
        // 403 file gh-1362-1.c line 14 column 5 function t2
        *shared2=2;
        // 404 file gh-1362-1.c line 14 column 5 function t2
        tmp_c:@shared_data=0;
        // 405 file gh-1362-1.c line 14 column 5 function t2
        ASSERT !tmp_c:@shared_data // W/W data race on c:@shared_data
        // 406 file gh-1362-1.c line 14 column 5 function t2
        ASSERT !tmp_c:gh-1362-1.c@194@F@t2@shared2 // R/W data race on c:gh-1362-1.c@194@F@t2@shared2
        // 407 file gh-1362-1.c line 16 column 5 function t2
        RETURN: 0
        // 408 file gh-1362-1.c line 17 column 1 function t2
        END_FUNCTION // t2
```

After quite a lot of interleavings (2986), the assert on instruction 393 is found to be violated. So it seems for every accessed symbol a new `tmp_`-prefixed variable is introduced tracking the accesses. This variable is set to 1 before the access, then the access is made, and it is set to 0 after that. As a final step an assertion is encoded that this variable still is 0.

When we just fix the dereference, the GOTO code for the original (non-global) t1/t2 example would look like that:
```
t1 (c:@F@t1):
        // 383 file gh-1362.c line 4 column 5 function t1
        signed int * shared1;
        // 384 file gh-1362.c line 4 column 5 function t1
        tmp_c:gh-1362.c@48@F@t1@shared1=1;
        // 385 file gh-1362.c line 4 column 5 function t1
        shared1=(signed int *)arg;
        // 386 file gh-1362.c line 4 column 5 function t1
        tmp_c:gh-1362.c@48@F@t1@shared1=0;
        // 387 file gh-1362.c line 4 column 5 function t1
        ASSERT !tmp_c:gh-1362.c@48@F@t1@shared1 // W/W data race on c:gh-1362.c@48@F@t1@shared1
        // 388 file gh-1362.c line 4 column 5 function t1
        ASSERT !tmp_c:gh-1362.c@31@F@t1@arg // R/W data race on c:gh-1362.c@31@F@t1@arg
        // 389 file gh-1362.c line 5 column 5 function t1
        tmp_symex::invalid_object21=1;
        // 390 file gh-1362.c line 5 column 5 function t1
        *shared1=4;
        // 391 file gh-1362.c line 5 column 5 function t1
        tmp_symex::invalid_object21=0;
        // 392 file gh-1362.c line 5 column 5 function t1
        ASSERT !tmp_symex::invalid_object21 // W/W data race on symex::invalid_object21
        // 393 file gh-1362.c line 5 column 5 function t1
        ASSERT !tmp_c:gh-1362.c@48@F@t1@shared1 // R/W data race on c:gh-1362.c@48@F@t1@shared1
        // 394 file gh-1362.c line 7 column 5 function t1
        RETURN: 0
        // 395 file gh-1362.c line 8 column 1 function t1
        END_FUNCTION // t1

t2 (c:@F@t2):
        // 396 file gh-1362.c line 11 column 5 function t2
        signed int * shared2;
        // 397 file gh-1362.c line 11 column 5 function t2
        tmp_c:gh-1362.c@168@F@t2@shared2=1;
        // 398 file gh-1362.c line 11 column 5 function t2
        shared2=(signed int *)arg;
        // 399 file gh-1362.c line 11 column 5 function t2
        tmp_c:gh-1362.c@168@F@t2@shared2=0;
        // 400 file gh-1362.c line 11 column 5 function t2
        ASSERT !tmp_c:gh-1362.c@168@F@t2@shared2 // W/W data race on c:gh-1362.c@168@F@t2@shared2
        // 401 file gh-1362.c line 11 column 5 function t2
        ASSERT !tmp_c:gh-1362.c@151@F@t2@arg // R/W data race on c:gh-1362.c@151@F@t2@arg
        // 402 file gh-1362.c line 12 column 5 function t2
        tmp_symex::invalid_object22=1;
        // 403 file gh-1362.c line 12 column 5 function t2
        *shared2=2;
        // 404 file gh-1362.c line 12 column 5 function t2
        tmp_symex::invalid_object22=0;
        // 405 file gh-1362.c line 12 column 5 function t2
        ASSERT !tmp_symex::invalid_object22 // W/W data race on symex::invalid_object22
        // 406 file gh-1362.c line 12 column 5 function t2
        ASSERT !tmp_c:gh-1362.c@168@F@t2@shared2 // R/W data race on c:gh-1362.c@168@F@t2@shared2
        // 407 file gh-1362.c line 14 column 5 function t2
        RETURN: 0
        // 408 file gh-1362.c line 15 column 1 function t2
        END_FUNCTION // t2
```
It can be seen here that `symex::invalid_object21` and 22 are introduced, because the value-sets for `arg` are empty. As mentioned above, the reason is that the intrinsic isn't understood by the value-set/static_analysis.

fbrausse:
> This seems like a much simpler approach, would you like me to use it instead of the existing one?

@XLiZHI It is simpler, but it's not working as intended. :)

Since the if-condition is `0`, the new `tmp_`-prefixed variables would be assigned `0` as well (it is the guard, after all). So still some special handling would be required. This is one of the rare cases we have, where the operational model for value-set needs to be different than that for symex. At least as long as value_set_analysist doesn't understand the intrinsic.

mikhailramalho:
Thanks for the detailed answer @fbrausse.

> 
> After quite a lot of interleavings (2986), the assert on instruction 393 is found to be violated.

Wait, this is the result of this PR, right? IIUC, we can't find the violation with the current master.

> So it seems for every accessed symbol a new `tmp_`-prefixed variable is introduced tracking the accesses. This variable is set to 1 before the access, then the access is made, and it is set to 0 after that. As a final step an assertion is encoded that this variable still is 0.

Yeah, that's what I thought, we need these auxiliary variables to track read/writes.

> For the value-set on master it is impossible to make the connection between the parameter `&shared_data` given to pthread_create() and the parameter to `t1(void *arg)`. arg's value-set will just remain empty, that is, the dereference will be a failed symbol. The reason is that the parameter is passed via the `__ESBMC_set_thread_internal_data` intrinsic and received by pthread_trampoline() via the get-version of this intrinsic. This intrinsic is only handled in symex.
> 
> Value-set doesn't know about this intrinsic and therefore can't make the connection. So either the general static_analysist or the specific value_set_analysist need to be taught about either the intrinsics, or the higher-level pthread_create/pthread_trampoline functions in order to make this connection.

My point is: given that this is a static analysis, i.e., there is no way to know the value of the variable, shouldn't the assertion be improved?

Let me try to summarize my understanding: 
1. We have a problem identifying data races when the shared memory passed via args is accessed because of names
2. We don't handle dereferences when adding the data race assertions
3. By adding code to handle the dereference, value-set doesn't know what *arg means, so it creates the invalid_object assignment and assertion, still with different names

Is this correct?

If so, the problem doesn't seem to be either the missing dereference case or the value-set, but rather the naming.

> 
> When we just fix the dereference, the GOTO code for the original (non-global) t1/t2 example would look like that:
> 
> ```
> t1 (c:@F@t1):
>         // 383 file gh-1362.c line 4 column 5 function t1
>         signed int * shared1;
>         // 384 file gh-1362.c line 4 column 5 function t1
>         tmp_c:gh-1362.c@48@F@t1@shared1=1;
>         // 385 file gh-1362.c line 4 column 5 function t1
>         shared1=(signed int *)arg;
>         // 386 file gh-1362.c line 4 column 5 function t1
>         tmp_c:gh-1362.c@48@F@t1@shared1=0;
>         // 387 file gh-1362.c line 4 column 5 function t1
>         ASSERT !tmp_c:gh-1362.c@48@F@t1@shared1 // W/W data race on c:gh-1362.c@48@F@t1@shared1
>         // 388 file gh-1362.c line 4 column 5 function t1
>         ASSERT !tmp_c:gh-1362.c@31@F@t1@arg // R/W data race on c:gh-1362.c@31@F@t1@arg
>         // 389 file gh-1362.c line 5 column 5 function t1
>         tmp_symex::invalid_object21=1;
>         // 390 file gh-1362.c line 5 column 5 function t1
>         *shared1=4;
>         // 391 file gh-1362.c line 5 column 5 function t1
>         tmp_symex::invalid_object21=0;
>         // 392 file gh-1362.c line 5 column 5 function t1
>         ASSERT !tmp_symex::invalid_object21 // W/W data race on symex::invalid_object21
>         // 393 file gh-1362.c line 5 column 5 function t1
>         ASSERT !tmp_c:gh-1362.c@48@F@t1@shared1 // R/W data race on c:gh-1362.c@48@F@t1@shared1
>         // 394 file gh-1362.c line 7 column 5 function t1
>         RETURN: 0
>         // 395 file gh-1362.c line 8 column 1 function t1
>         END_FUNCTION // t1
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> t2 (c:@F@t2):
>         // 396 file gh-1362.c line 11 column 5 function t2
>         signed int * shared2;
>         // 397 file gh-1362.c line 11 column 5 function t2
>         tmp_c:gh-1362.c@168@F@t2@shared2=1;
>         // 398 file gh-1362.c line 11 column 5 function t2
>         shared2=(signed int *)arg;
>         // 399 file gh-1362.c line 11 column 5 function t2
>         tmp_c:gh-1362.c@168@F@t2@shared2=0;
>         // 400 file gh-1362.c line 11 column 5 function t2
>         ASSERT !tmp_c:gh-1362.c@168@F@t2@shared2 // W/W data race on c:gh-1362.c@168@F@t2@shared2
>         // 401 file gh-1362.c line 11 column 5 function t2
>         ASSERT !tmp_c:gh-1362.c@151@F@t2@arg // R/W data race on c:gh-1362.c@151@F@t2@arg
>         // 402 file gh-1362.c line 12 column 5 function t2
>         tmp_symex::invalid_object22=1;
>         // 403 file gh-1362.c line 12 column 5 function t2
>         *shared2=2;
>         // 404 file gh-1362.c line 12 column 5 function t2
>         tmp_symex::invalid_object22=0;
>         // 405 file gh-1362.c line 12 column 5 function t2
>         ASSERT !tmp_symex::invalid_object22 // W/W data race on symex::invalid_object22
>         // 406 file gh-1362.c line 12 column 5 function t2
>         ASSERT !tmp_c:gh-1362.c@168@F@t2@shared2 // R/W data race on c:gh-1362.c@168@F@t2@shared2
>         // 407 file gh-1362.c line 14 column 5 function t2
>         RETURN: 0
>         // 408 file gh-1362.c line 15 column 1 function t2
>         END_FUNCTION // t2
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> It can be seen here that `symex::invalid_object21` and 22 are introduced, because the value-sets for `arg` are empty. As mentioned above, the reason is that the intrinsic isn't understood by the value-set/static_analysis.

This look wrong... ? 

I guess it makes sense now that we don't handle the dereference: given that we add the auxiliary variables to track read/write, we don't care about who or where we are assigning the value, but rather if the names of the auxiliary variables match? i.e., this example fails because one auxiliary variable is named `tmp_c:main.c@31@F@t1@arg` and the other is `tmp_c:main.c@151@F@t2@arg`.

Maybe we should find a way to not use the variable name, but its value? I can only see this working at symex, or we'll need to special case pthread_create. I still don't like the idea of a special case for pthread_create tho.

What if we create a new intrinsic during static analysis that is handled at symex? Something like: 
```
ASSERT !__ESBMC_check_arg_data_race(arg)
```
At symex, we would dereference arg and create the proper auxiliary flags.

Creating these assertions based on the naming seems very fragile

fbrausse:
> > After quite a lot of interleavings (2986), the assert on instruction 393 is found to be violated.
> 
> Wait, this is the result of this PR, right? IIUC, we can't find the violation with the current master.

No, this was the result when just the dereference in rw_set got fixed *and* `shared_data` is a global variable.

mikhailramalho:
> > > After quite a lot of interleavings (2986), the assert on instruction 393 is found to be violated.
> > 
> > 
> > Wait, this is the result of this PR, right? IIUC, we can't find the violation with the current master.
> 
> No, this was the result when just the dereference in rw_set got fixed _and_ `shared_data` is a global variable.

I see, the global variable as passed via the function argument? I guess this should fix the problem then.

mikhailramalho:
also, 2986 interleavings!? Another thing that must be fixed

fbrausse:
> Let me try to summarize my understanding:
> 
>  1. We have a problem identifying data races when the shared memory passed via args is accessed because of names
>  2. We don't handle dereferences when adding the data race assertions
>  3. By adding code to handle the dereference, value-set doesn't know what *arg means, so it creates the invalid_object assignment and assertion, still with different names
> 
> Is this correct?

Item 1. yes, but I'm not sure it's because of "passed via args", see the array example below.
Item 2. is correct: it's a bug: `dereference()` only does anything when it is given a dereference-expression. This should be fixed regardless of pthread_create.
Item 3. is correct.

> If so, the problem doesn't seem to be either the missing dereference case or the value-set, but rather the naming.

The missing entry in the value-set is a problem: pthread_create() parameters are not passed properly to functions outside of symex (due to intrinsic only being interpreted there) and, as I understand it, the value_set_analysist is a method that's also supposed to work statically.

I fully agree about the naming being fragile. It's just the symbol's name, but what if it is an array `int x[2]` and thread1 only accesses `x[0]` and thread2 only accesses `x[1]`? The current implementation would also report this as a data-race. Is it? What happens to symbols of other composite types like structs? Does anyone know the precise definition of what constitutes a data-race for composite types? All I can find is about "memory location", which, considering typical cache-line sizes of 64 bytes, could span multiple components of a composite type.

> also, 2986 interleavings!? Another thing that must be fixed

Indeed!

@XLiZHI what about splitting this PR into two? Keep the deref fixes you made and move the special-casing of pthread_create into a separate PR so that both changes can be looked at separately. Sorry if this creates more seemingly unnecessary work for you, but I do agree with @mikhailramalho's points about relying on the naming being too fragile. To be honest, I don't understand how the current master's approach could work reliably as a static analysis pass. It seems to produce an over-approximation of all possible data races.

lucasccordeiro:
@XLiZHI: We must recognize that the problem you're trying to solve is not straightforward. So, let's try to break this problem into sub-problems. Please submit a PR that fixes the dereference issue and another that handles the `pthread_create` issue.

How does it sound to you?

XLiZHI:
> I fully agree about the naming being fragile. It's just the symbol's name, but what if it is an array int x[2] and thread1 only accesses x[0] and thread2 only accesses x[1]? The current implementation would also report this as a data-race. Is it?

There are ways to handle arrays by combining the inner index into "[index]", but as the code shows, it also has the problem that index will not work if it is not a constant value but a variable. This works for compound types. As you can see, this is a recursive function that steps through each case.
https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/goto-programs/rw_set.cpp#L74-L85

> what about splitting this PR into two? Keep the deref fixes you made and move the special-casing of pthread_create into a separate PR so that both changes can be looked at separately.

That's fine with me. But in my opinion, it makes no sense to just fix dereference, because it does not implement any functionality. It may degrade performance because we do dereferencing in all goto fucntions.

lucasccordeiro:
@XLiZHI: what about we try to solve this problem of checking for data-races at symex level?

fbrausse:
> But in my opinion, it makes no sense to just fix dereference, because it does not implement any functionality. It may degrade performance because we do dereferencing in all goto fucntions.

Just fixing bugs is what many of my previous PRs did. ;) For me it serves the purpose of establishing a baseline of known-working code. But I agree that performance shouldn't suffer unnecessarily.

> There are ways to handle arrays by combining the inner index into "[index]", but as the code shows, it also has the problem that index will not work if it is not a constant value but a variable.

Thanks, @XLiZHI. So accessing `x[(size_t)pthread_self()]` wouldn't be covered by the check, I see. Do you know what are the users of `--data-races-check` right now and whether they expect this dynamic index access to work?

XLiZHI:
> @XLiZHI: what about we try to solve this problem of checking for data-races at symex level?

Of course, moving to symex would solve the value set problem, symex seems to be a different representation, which I don't quite understand yet.

But I think even moving to symex we would face some problems:

1. Do we handle all dereferences at data races check? If not, how should we judge which dereferences we should deal with.

2. Symbolic execution is a process. Where should we put it? We need to be able to access both the value set and the value of the variable.

3.  Is SSA similar to a goto function? Can I add the corresponding races assertion by looping through the SSA?

> Do you know what are the users of --data-races-check right now and whether they expect this dynamic index access to work?

For me, I would like to solve this problem so that ESBMC can handle data race checking for CUDA programs (all based on this, you can check TCs disabled in this PR for details), and I would also try SV-COMP.

mikhailramalho:
> That's fine with me. But in my opinion, it makes no sense to just fix dereference, because it does not implement any functionality. It may degrade performance because we do dereferencing in all goto fucntions.

I don't understand your claim about performance: it's obvious that the current code is more "performant", we are doing an incomplete analysis.

~

I agree with 2 PRs, but instead of the second one making changes to special case the pthread_create during static analysis, I suggest we go ahead and change symex. We already concluded that we don't have all relevant information during static analysis, so let's not add more technical debt.

fbrausse:
> We already concluded that we don't have all relevant information during static analysis, so let's not add more technical debt.

True for the data-races check, but not necessarily true for the value-set. Is this value-set and static_analysis pass being used statically during goto elsewhere? If so, there should be cases being missed for pthread_create as well.

@rafaelsamenezes you mentioned in https://github.com/esbmc/esbmc/issues/1324#issuecomment-1710704523 that you're using something similar to the method adjusted in this branch based on the static_analysist class. Are you experiencing similar issues with pthread_create?

rafaelsamenezes:
> @rafaelsamenezes you mentioned in https://github.com/esbmc/esbmc/issues/1324#issuecomment-1710704523 that you're using something similar to the method adjusted in this branch based on the static_analysist class. Are you experiencing similar issues with pthread_create?

I have only worked with single thread.

Its a bit hard to understand the discussion at this point. Ideally the VSA should work on both the static analyses and symex to a similar extent (as they even share a lot of code). The main difference is that in goto it works at level 0 and in symex at level 1. Having it available for GOTO programs is essential for the interval analysis (in future) and the GCSE algorithm that you just refered. 

The main issue here is that the "static analysis" frameworks (abstract interpretation and static analysis) are not thread aware. The thread lifetime should be reflected when trying to reach a fix point.

XLiZHI:
Hi everyone, thanks for your analysis.

Just a thought, since we can't handle dereference and  get an index that isn't a constant value  at the goto level. maybe we can add dynamic assertions to solve in symex?(Like: assert ! address of `symbol`), instead of just using a static **symbol name**. 

I think maybe we can borrow ideas from array bounds checking.
https://github.com/esbmc/esbmc/blob/27c6bbbbefd636956d0dd4cd28238e1fa4cfd109/src/pointer-analysis/dereference.cpp#L2255-L2263

PR #1364, opened 2023-09-20
replace pow() recursion with loop

Anthonysdu:
esbmc cannot finish unrolling the recursion in `pow()` function, and also doesn't work with `--unwindset` because it doesn't have a loopid.
```
int main() {
  int res = pow(2, 4);
  assert(res == 16);
}
```
Shall we replace the recursion with the loop?

### DIFF
--- a/src/c2goto/library/libm/pow.c
+++ b/src/c2goto/library/libm/pow.c
@@ -5,18 +5,30 @@
 double pow(double base, double exponent)
 {
 __ESBMC_HIDE:;
-  int result = 1;
+  double result = 1.0;
   if(exponent == 0)
     return result;
 
   if(exponent < 0)
-    return 1 / pow(base, -exponent);
+  {
+    base = 1.0 / base;
+    exponent = -exponent;
+  }
 
-  float temp = pow(base, exponent / 2);
-  if((int)exponent % 2 == 0)
-    return temp * temp;
-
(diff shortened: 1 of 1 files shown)


lucasccordeiro (changes requested):
@Anthonysdu: can you please add regression tests?

fbrausse:
This is the wrong implementation. The `pow` function is supposed to compute a double-precision approximation of x ** y, i.e., exp(log(x) * y). Certain range reductions like log(x) = 2 * log(sqrt(x)) could be done iteratively when working with an exact representation, but y itself is not an integer and cannot be treated as such by a loop.

Anthonysdu:
> This is the wrong implementation. The `pow` function is supposed to compute a double-precision approximation of x ** y, i.e., exp(log(x) * y). Certain range reductions like log(x) = 2 * log(sqrt(x)) could be done iteratively when working with an exact representation, but y itself is not an integer and cannot be treated as such by a loop.

I see, thanks

emanino:
@Anthonysdu if you need a double-precision operational model of pow(), have a look at
https://github.com/esbmc/esbmc/pull/1232

fbrausse:
#1421 has landed, which includes an unrolled version of the exponentiation-by-squaring method patched in this PR. @Anthonysdu Shall we close this PR?

PR #1369, opened 2023-09-25
Fix output of negative integer constants

### LINKED ISSUE #1366: --smtlib does not output negative integers correctly
When running on the following program

```
extern void abort(void);
void reach_error() { __assert_fail("0", "vnew2.c", 3, "reach_error"); }
extern void abort(void);
void assume_abort_if_not(int cond) {
  if(!cond) {abort();}
}
void __VERIFIER_assert(int cond) {
  if (!(cond)) {
  }
  return;
}
int SIZE = 20000001;
int __VERIFIER_nondet_int();
int main() {
   int n = 0;
   int j = 1;
   int k = __VERIFIER_nondet_int();
  __ESBMC_assume(k == j * -1);
  __VERIFIER_assert(k == (2 - 1) * j * -1);
}
```

with options 

`--smtlib --ir --output out`

The integers -1 are not represented correctedly in the SMTLIB output. They are represented as -1 instead of (- 1).

fbrausse:
Fixes #1366. In SMT-LIB v2 format negative integer literals do not exist, but those numbers have to be expressed as, for instance, `(- 1)`. This PR addresses this for the `--smtlib` output.


lucasccordeiro (approved):

Can you please add a regression test here, if possible?

ibnyusuf:
Looks good. Thanks!

ibnyusuf (approved):
Looks good. Thanks!

fbrausse:
> Can you please add a regression test here, if possible?

Done. I've taken the example from the issue and check that an equality with `(- 1)` indeed is generated for it.

lucasccordeiro:

PR #1370, opened 2023-09-26
Use check_linker_flag() instead of compiler id

fbrausse:
The linker used by the compiler is not determined by its name. For instance, on typical Linux distributions the Clang compiler defaults to GNU ld (aka ld.bfd) while on macOS Clang seems to use Apple's ld64, which accepts options similar to ld.lld, but different from ld.bfd or ld.gold.

So instead we test all the variants that are known to us.

### DIFF
--- a/src/solvers/smtlib/CMakeLists.txt
+++ b/src/solvers/smtlib/CMakeLists.txt
@@ -26,20 +26,42 @@ target_include_directories(smtlib
 # The smtlib backend needs a larger-than-default stack size (at least in Debug
 # builds) because the data structures built for the S-exprs in solver responses
 # could be very large and the std::list destructors use a lot of stack space
-set(stacksize 20971520)  # 20 MB ought to be enough for everyone
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:${stacksize}" CACHE STRING "Linker flags" FORCE)
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
-  set(stacksize 0x1400000) # macOS uses hex by default (see Issue #1212)
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-stack_size,${stacksize}" CACHE STRING "Linker flags" FORCE)
-else()
-  # GNU probably
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,stack-size=${stacksize}" CACHE STRING "Linker flags" FORCE)
+
+# 20 MB ought to be enough for everyone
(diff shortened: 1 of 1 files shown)


fbrausse:
I should say why this change is needed: When compiling with Clang on Linux it aborts with
```
/usr/bin/x86_64-pc-linux-gnu-ld.bfd: Warning: grouped short command line options are deprecated: -stack_size
/usr/bin/x86_64-pc-linux-gnu-ld.bfd: Error: unable to disambiguate: -stack_size (did you mean --stack_size ?)
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```
due to assuming the ld64/ld.lld option style.

rafaelsamenezes (approved):
LGTM. Now we should be able to add some clang sanitizer builds.

lucasccordeiro (approved):

fbrausse:
> Now we should be able to add some clang sanitizer builds.

Have you tried the GCC sanitizers?

lucasccordeiro:

src/solvers/smtlib/CMakeLists.txt
```diff
@@ -26,20 +26,42 @@ target_include_directories(smtlib
...
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:${stacksize}" CACHE STRING "Linker flags" FORCE)
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
-  set(stacksize 0x1400000) # macOS uses hex by default (see Issue #1212)
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-stack_size,${stacksize}" CACHE STRING "Linker flags" FORCE)
-else()
-  # GNU probably
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,stack-size=${stacksize}" CACHE STRING "Linker flags" FORCE)
+
+# 20 MB ought to be enough for everyone
```
mikhailramalho, line 30: famous last words

mikhailramalho (commented):
lgtm

rafaelsamenezes:
> Have you tried the GCC sanitizers?

Not really, do they use the same flags as clang? If not, it might be a matter of just adding it in here: https://github.com/esbmc/esbmc/blob/master/scripts/cmake/Sanitizers.cmake

fbrausse:
> Not really, do they use the same flags as clang? If not, it might be a matter of just adding it in here: https://github.com/esbmc/esbmc/blob/master/scripts/cmake/Sanitizers.cmake

I think so, yes. I'll give the `CMAKE_BUILD_TYPE="Sanitizer"` a try. (Though I wonder whether it wouldn't be better to have this as an option independent of the CMAKE_BUILD_TYPE.)

rafaelsamenezes:
> Though I wonder whether it wouldn't be better to have this as an option independent of the CMAKE_BUILD_TYPE

I was wondering the same thing yesterday. When I developed this, I didn't know that Debug/Release could have different bugs.

PR #1372, opened 2023-09-26
builtin-memset: same behaviour in Debug and release builds

### LINKED ISSUE #1171: Crash for memset on float types

fbrausse:
Fixes #1171: If the optimization isn't feasible for release, it shouldn't abort() for debug either. I've made the accompanying comment a TODO: memset should also work on `float` and `double`. E.g., `float f; memset(&f, 0, sizeof(f));` is valid and will produce the value `f == 0.0f`.

### DIFF
--- a/regression/bitwuzla/memset-good/test.desc
+++ b/regression/bitwuzla/memset-good/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 memset-good.c
 --bitwuzla
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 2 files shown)


lucasccordeiro (approved):
That's great! 

PR #1373, opened 2023-09-26
Guard simplification and a few random fixes

fbrausse:
These changes came out of #1297. Unlike that one, this PR doesn't try to modify the implementation of guards because that one still has too much overhead. I will note again, though, that the current implementation of operators -= and |= is undefined and we should look into these.

Changes on guards are:
1. Remove std::move from `merge_state_guards()` because it is doing nothing: `guardt` defines the copy-constructor and therefore no move-constructor is implicitly being generated by the compiler. It just isn't available. This construction in the function uses the copy-ctor. It was backported from CBMC in ff829c0fd324e0f4322505450a56d68444ef203b and there still is unchanged. There guards don't have a move-ctor either.
2. Since now the definition of the copy-ctor of guards is unnecessary, remove it. This enables the implicit move-ctors, which are used elsewhere. To me this simplifies reasoning about the behaviour of guard objects.
3. Fix building the expr2tc: `build_guard_expr()` would only assign `g_expr` if the `guard_list` had strictly more than one symbol. However, `guardt::add()` assumes that `g_expr` would be valid also if `guard_list` had a single symbol.

Minor other changes:
- Support multi-digit MathSAT version numbers (like the current v5.6.10).
- Pointer arithmetic for --32 should be allowed operands of 64 bit size.

The following SV-COMP numbers are for this PR **except** the change in item 2. above. While to me it should only improve speed, C++ sometimes is hard to analyze, so please let me know if you would like me to run SV-COMP again with the change in item 2.

[Master](https://github.com/esbmc/esbmc/actions/runs/6240185504) (ce7742ec1fcd179c4b7f0818abe31e09151fd1ab):
```
  correct:           14161
    correct true:     7814
    correct false:    6347
  incorrect:            22
    incorrect true:      7
    incorrect false:    15
  unknown:            9622
  Score:             21511 (max: 38482)
```
[This PR except item 2.](https://github.com/esbmc/esbmc/actions/runs/6294055509) (dabde69490b95683dc7f5439a0a34ffee6502d02):
```
  correct:           14156
    correct true:     7811
    correct false:    6345
  incorrect:            21
    incorrect true:      6
    incorrect false:    15
  unknown:            9628
  Score:             21535 (max: 38482)
```
The one less incorrect in the stats is not actually solved by this PR, it is just a bit slower: c/pthread/triangular-longest-2.i in unreach-call (27s -> timeout).

### DIFF
--- a/src/goto-symex/symex_goto.cpp
+++ b/src/goto-symex/symex_goto.cpp
@@ -222,22 +222,23 @@ static inline guardt merge_state_guards(
   // impossible. Therefore we only integrate it when to do so simplifies the
   // state guard.
 
-  // This function can trash either state's guards, since goto_state is dying
-  // and state's guard will shortly be overwritten.
+  // In CBMC this function can trash either state's guards, since goto_state is
+  // dying and state's guard will shortly be overwritten. However, we still use
+  // either state's guard, so keep them intact.
   if(
     (!goto_state.guard.is_false() && !state.guard.is_false()) ||
     state.guard.disjunction_may_simplify(goto_state.guard))
   {
     state.guard |= goto_state.guard;
-    return std::move(state.guard);
+    return state.guard;
   }
   else if(state.guard.is_false() && !goto_state.guard.is_false())
   {
-    return std::move(goto_state.guard);
+    return goto_state.guard;
   }
   else
   {
-    return std::move(state.guard);
+    return state.guard;
   }
 }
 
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):
Thanks, @fbrausse.

Is there some chance to consider adding a regression test here?

fbrausse:
A side note: Without change 1. I didn't see any problems unless applying #744 ontop. Although that PR should be strictly an optimization and not introduce semantic changes, it did affect SV-COMP's unreach-call
- c/aws-c-common/aws_array_list_pop_back_harness.i

without the change 1. With it, it works under #744 as well.

fbrausse:
> Is there some chance to consider adding a regression test here?

Tricky. As mentioned in the side-note above, under #744 I would have an example for a visible change. But without that I don't know, yet. We can give this PR some time and I'll think about it.

lucasccordeiro:
That's fine, @fbrausse. Let's consider the regression test in the upcoming PR.

PR #1374, opened 2023-09-26
Fix some Clang warnings

fbrausse:
Just fix some warnings generated by Clang-16 on Linux. It's mostly:
- virtual dtor missing
- some else-cases shouldn't happen
- some override declarations missing

There are still more warnings left, in particular about unused functions, parameters and local variables. I've not touched those since I don't know the context off-hand and whether they could indeed be just removed.


mikhailramalho (approved):
lgtm

lucasccordeiro (approved):

PR #1376, opened 2023-09-27
Simplify irep1: do not resolve index on constants under addr-of

### LINKED ISSUE #1093: ESBMC crash on k-induction :ERROR: Unrecognized address_of operand:
ESBMC 7.2 crashes on the following small C program:

ESBMC command:  `esbmc DATA1/FormAI_11479.c --unwind 1 --k-induction --overflow`

note: the --k-induction and --overflow switches trigger the vulnerability.

```c
//FormAI DATASET v0.1 Category: Building a XML Parser ; Style: cheerful
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

typedef struct {
    char* tag;
    char* value;
} XMLNode;

void parseXML(char* xml) {
    char* ptr = xml;

    while (*ptr != '\0') {
        if (*ptr == '<') {
            XMLNode node;
            char* start_tag = ptr;

            // get the node tag
            while (*ptr != '>' && *ptr != ' ') {
                ptr++;
            }
            node.tag = (char*) malloc((ptr - start_tag) + 1);
            strncpy(node.tag, start_tag+1, (ptr - start_tag) - 1);
            node.tag[ptr - start_tag - 1] = '\0';

            //get the node value
            start_tag = ptr + 1;
            while (*start_tag == ' ') {
                start_tag++;
            }
            if (*start_tag != '<') {
                ptr = start_tag;
                while (*ptr != '<') {
                    ptr++;
                }
                node.value = (char*) malloc((ptr - start_tag) + 1);
                strncpy(node.value, start_tag, ptr - start_tag);
                node.value[ptr - start_tag] = '\0';

                printf("Tag: %s -> Value: %s\n", node.tag, node.value);
            } else {
                printf("Tag: %s -> No Value\n", node.tag);
            }

            free(node.tag);
            free(node.value);
        }
        ptr++;
    }
}

int main(void) {
    char* xml = "<root><name>XML Parser</name><author>John Doe</author><version>1.0.0</version></root>";
    parseXML(xml);
    return EXIT_SUCCESS;
}
```

ESBMC output:

```

Symex completed in: 0.078s (550 assignments)
Slicing time: 0.003s (removed 238 assignments)
Generated 224 VCC(s), 123 remaining after simplification (312 assignments)
ERROR: Unrecognized address_of operand:
address_of
* pointer_obj : constant_int
    * value : 60
    * type : signedbv
      * width : 8
* type : pointer
  * subtype : signedbv
      * width : 8
Aborted (core dumped)

```

### LINKED ISSUE #1097: The irep1 simplifier isn't controlled by `--no-simplify`
Turns out this is the irep1 simplifier, which appearently isn't controlled by `--no-simplify`.

_Originally posted by @fbrausse in https://github.com/esbmc/esbmc/issues/1093#issuecomment-1594423056_

I would expect the cmd-line option `--no-simplify` to also affect the irep1 simplifier living in `src/util/simplify_{expr,expr_class,utils}.*`. It does not, though.

fbrausse:
Fixes #1093.

I've disabled simplifying index-exprs on constant arrays, strings, etc. when the index itself is under an address-of.

In case the CI says "go", to be on the safe side, I'd like to run this over the benchmarks, too. Might be that a more fine-grained solution is required if there are major slow-downs due to disabling this particular simplification.

### DIFF
--- a/regression/esbmc/github_1093/gh-1093.c
+++ b/regression/esbmc/github_1093/gh-1093.c
@@ -1,0 +1,57 @@
+//FormAI DATASET v0.1 Category: Building a XML Parser ; Style: cheerful
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef struct {
+    char* tag;
+    char* value;
+} XMLNode;
+
+void parseXML(char* xml) {
+    char* ptr = xml;
+
+    while (*ptr != '\0') {
+        if (*ptr == '<') {
+            XMLNode node;
+            char* start_tag = ptr;
+
+            // get the node tag
+            while (*ptr != '>' && *ptr != ' ') {
+                ptr++;
+            }
+            node.tag = (char*) malloc((ptr - start_tag) + 1);
+            strncpy(node.tag, start_tag+1, (ptr - start_tag) - 1);
+            node.tag[ptr - start_tag - 1] = '\0';
+
+            //get the node value
+            start_tag = ptr + 1;
+            while (*start_tag == ' ') {
+                start_tag++;
+            }
+            if (*start_tag != '<') {
+                ptr = start_tag;
+                while (*ptr != '<') {
+                    ptr++;
+                }
+                node.value = (char*) malloc((ptr - start_tag) + 1);
+                strncpy(node.value, start_tag, ptr - start_tag);
+                node.value[ptr - start_tag] = '\0';
+
+                printf("Tag: %s -> Value: %s\n", node.tag, node.value);
+            } else {
+                printf("Tag: %s -> No Value\n", node.tag);
+            }
+
+            free(node.tag);
+            free(node.value);
+        }
+        ptr++;
+    }
+}
+
+int main(void) {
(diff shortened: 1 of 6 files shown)


fbrausse:
While at it let's also fix #1097.

fbrausse:
Seems to be fine.

From the last guard PR #1297: [Master](https://github.com/esbmc/esbmc/actions/runs/6294055509):
```
  correct:           14156
    correct true:     7811
    correct false:    6345
  incorrect:            21
    incorrect true:      6
    incorrect false:    15
  unknown:            9628
  Score:             21535 (max: 38482)
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/6329349910) (34a0319ad6253af7d60b1089a52ae2bb197a3236):
```
  correct:           14171
    correct true:     7824
    correct false:    6347
  incorrect:            22
    incorrect true:      7
    incorrect false:    15
  unknown:            9612
  Score:             21531 (max: 38482)
```
The new unknown -> incorrect is the timeout jitter c/pthread/triangular-longest-2.i mentioned in the guard PR. The others look like timeout noise as well.

lucasccordeiro:

PR #1382, opened 2023-09-28
Remove --memlimit 14000000 setting from test.desc for all regression tests

### LINKED ISSUE #1381: Default memlimit on C++ regression tests conflicts with AddressSanitizer
On a CMAKE_BUILD_TYPE=Sanitizer build of current master, I'm getting these crashes:
```
=================================================================
ERROR: Failed to mmap
```
for nearly all C++ regression tests, which have the `--memlimit 14000000` setting. That's 14 mio. MiB, way more than I have. When removing this setting, most of the regression tests work.

Can we remove this memlimit setting from the regression tests? Why is it there and why just for C++ tests?

*Edit:* Turns out, the default unit is `m` if none is given. So 14000000 is 14 TB. I'm sure that's not the right default.

fbrausse:
Closes #1381. I've also added a description of the default unit for the `--memlimit` option to ESBMC's help message. Hint: it's not bytes. The C++ tests still have the `--timeout 900` parameter, which, personally, I would remove as well. In particular since we can now pass such global constraints via `ESBMC_OPTS="--timeout 900" ctest ...`. In case they only apply to C++ for some reason, this PR doesn't touch them, though.

### DIFF
--- a/regression/cuda/COM_sanity_checks/call_kernel/test.desc
+++ b/regression/cuda/COM_sanity_checks/call_kernel/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.cu
-  --memlimit 14000000 --timeout 900
+ --timeout 900
 
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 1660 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #1390, opened 2023-10-07
Initial implementation of Python frontend 

brcfarias:
This pull request contains the initial implementation of a Python frontend for ESBMC. The frontend leverages the [ast2json](https://pypi.org/project/ast2json/) package, specifically through astgen.py, to generate the Abstract Syntax Tree.

In this approach, we've established a process (fork) that invokes the Python interpreter with astgen.py and the target program to be verified. This process generates an ast.json file in the /tmp directory as output. We then parse this JSON data in a similar manner to what occurs in the Jimple and Solidity frontends.

It's important to note that the current assumption is that the code will have type annotations. This assumption is based on the fact that the current target application for this frontend will include explicit type annotations for variables, parameters, and return values. We will address expressions without type annotations in future updates.

The current implementation already supports arithmetical and bitwise operations. This functionality enables the detection of errors like division by zero and bit shift with negative values.


lucasccordeiro (changes requested):

1) Can I ask you to add regression tests with `VERIFICATION SUCCESSFUL'?
2) Can I also ask you to document your implemented methods?

lucasccordeiro:
BTW, can I also ask you to check the CI of this PR?

src/python-frontend/astgen.py
```diff
@@ -0,0 +1,25 @@
...
+    if len(sys.argv) != 2:
+        print("Usage: python astgen.py <file path>")
+        return
+
+    filename = sys.argv[1]
+
+    with open(filename, "r") as source:
+        tree = ast.parse(source.read())
+
+    #print(ast.dump(tree, indent=4))
```
kunjsong01: remove?

src/python-frontend/python_language.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  // Create a child process to execute Python
+  bp::child process(cmd);
+
+  // Wait for execution
+  process.wait();
+
+  if(process.exit_code())
+  {
+    // Python execution failed
```
kunjsong01: perhaps, wrap this comment in log_error?

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,256 @@
...
+  }
+  }
+
+  return expr;
+}
+
+symbolt python_converter::get_var_decl(const nlohmann::json &ast_node)
+{
+  // Get variable type
+  current_element_type =
```
kunjsong01, line 180: get_typet might return an empty type. Do we need to add an additional check after this one?

kunjsong01 (approved):
LGTM. 
Just some nitpicks.

brcfarias:
Thanks for the suggestions.

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,258 @@
...
+  symbol.id = id;
+  return symbol;
+}
+
+locationt get_location_from_decl(const nlohmann::json &ast_node)
+{
+  locationt location;
+  location.set_line(ast_node["target"]["lineno"].get<int>());
+  // TODO: Modify ast.py to include Python filename in the generated json
+  location.set_file("program.py"); // FIXME: This should be read from JSON
```
lucasccordeiro, line 95: Why do we have this filename hardcoded?

CMakeLists.txt
```diff
@@ -108,6 +108,25 @@ if(ENABLE_JIMPLE_FRONTEND)
...
+        COMMAND python3 -c "import ast2json"
+        RESULT_VARIABLE PYTHON_PACKAGE_RESULT
+        OUTPUT_QUIET
+    )
+
+    # Install ast2json if not available
+    if (NOT PYTHON_PACKAGE_RESULT EQUAL 0)
+        message("Installing ast2json")
+        execute_process(
+            COMMAND pip3 install ast2json
```
fbrausse: Can we put this behind the `DOWNLOAD_DEPENDENCIES` CMake flag?

scripts/cmake/InstallFiles.cmake
```diff
@@ -24,4 +24,8 @@ install(FILES scripts/release-notes.txt DESTINATION .)
 # Hack to ship boost dll's if needed
 if(DEFINED BOOST_DLL_FILE)
   install(FILES ${BOOST_DLL_FILE} DESTINATION bin)
+endif()
+
+if(ENABLE_PYTHON_FRONTEND)
+install(FILES src/python-frontend/astgen.py DESTINATION resources)
```
fbrausse: Indent please.

scripts/cmake/Utils.cmake
```diff
@@ -1,7 +1,6 @@
 # Module to utilities used throughout the project
 
-
-set(ESBMC_BIN "${CMAKE_BINARY_DIR}/src/esbmc/esbmc")
+set(ESBMC_BIN "${CMAKE_INSTALL_PREFIX}/bin/esbmc")
```
fbrausse: This breaks my workflow, why is it needed?

src/CMakeLists.txt
```diff
@@ -21,6 +21,10 @@ if(ENABLE_JIMPLE_FRONTEND)
 add_subdirectory (jimple-frontend)
 endif()
 
+if(ENABLE_PYTHON_FRONTEND)
+add_subdirectory(python-frontend)
```
fbrausse: Indentation.

src/langapi/mode.cpp
```diff
@@ -14,6 +14,9 @@ static const char *const extensions_cpp[] =
 
 static const char *const extensions_sol_ast[] = {"solast", nullptr};
 static const char *extensions_jimple[] = {"jimple", nullptr};
+
+const char *extensions_python[] = {"py", nullptr};
```
fbrausse: Please make this variable static and const.

src/langapi/mode.cpp
```diff
@@ -23,6 +26,8 @@ static const language_desct language_desc_Jimple = {
   "Jimple",
   extensions_jimple};
 
+const language_desct language_desc_python = {"Python", extensions_python};
```
fbrausse: static.

src/python-frontend/astgen.py
```diff
@@ -0,0 +1,23 @@
...
+        print("Usage: python astgen.py <file path>")
+        return
+
+    filename = sys.argv[1]
+
+    with open(filename, "r") as source:
+        tree = ast.parse(source.read())
+
+    ast_json = ast2json(tree)
+    json_filename = "/tmp/ast.json"
```
fbrausse: Is there any way to make this a bit more generic? Typically the convention in `/tmp` is that existing files should not be overwritten.

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,258 @@
...
+#include "util/arith_tools.h"
+#include "util/expr_util.h"
+#include "util/message.h"
+
+#include <fstream>
+#include <unordered_map>
+
+using json = nlohmann::json;
+
+const char *json_filename = "/tmp/ast.json";
```
fbrausse: static const.

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+std::unordered_map<std::string, std::string> operator_map = {
```
fbrausse: static const.

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+enum class ExpressionType
+{
+  UNARY_OPERATION,
+  BINARY_OPERATION,
+  LITERAL,
+  UNKNOWN,
+};
+
+StatementType get_statement_type(const json &element)
```
fbrausse: Can we make this static? No need to export and potentially confuse the linker in case there is another function with the same signature somewhere else. Also applies to the non-member functions below.

src/python-frontend/python_language.cpp
```diff
@@ -0,0 +1,81 @@
...
+bool python_languaget::parse(const std::string &path)
+{
+  log_debug("python", "Parsing: {}", path);
+
+  fs::path script(path);
+  if(!fs::exists(script))
+    return true;
+
+  fs::path esbmc_binary_path = fs::read_symlink("/proc/self/exe").parent_path();
+  fs::path python_script_path = esbmc_binary_path / "../resources/astgen.py";
```
fbrausse: This will not work on native Windows. We could either hide it behind a #define, or do what we do with other objects distributed with ESBMC: bundle it in the binary and extract it during runtime. See the `flail.py` tool for bundling, also there is the CMake function `mangle` defined in scripts/cmake/Utils.cmake. For extracting I would point to src/c2goto/cprover_libc_sources.cpp where the standard headers are extracted when translating C sources.

fbrausse (commented):
I have a few comments about the style of the changes in this PR. I haven't deeply looked into the implementation, yet.

src/python-frontend/python_converter.cpp
brcfarias: empty_type represents void, not necessarily an error condition. So, it may not be necessary to handle it at this point.

CMakeLists.txt
brcfarias: "DOWNLOAD_DEPENDENCIES" is set to "On" only when using "script/build.sh". This can be an issue when building directly with CMake (as outlined in BUILDING.md).

fbrausse: This suggestion is independent of build.sh. `DOWNLOAD_DEPENDENCIES` is a user-facing option. I usually have it disabled, for reasons. In case the PYTHON_PACKAGE_RESULT is not 0 and DOWNLOAD_DEPENDENCIES is turned off, there should be no downloading of dependencies. *Edit:* Instead, an error should be raised by CMake.

As a side note: `pip install` installs either
- system-wide (e.g. under `/`) in case it has rights to do so, or
- local to the current user under `$HOME/.local`.

Both are not the case with DOWNLOAD_DEPENDENCIES as we have it now: it only does it in the local cmake build directory. @rafaelsamenezes Do you have a suggestion how we could work around this issue?

scripts/cmake/Utils.cmake
brcfarias: > This breaks my workflow

Could you please clarify which specific workflow? It is passing on CI, where we have a clean build.

> why is it needed?

astgen.py is installed at CMAKE_INSTALL_PREFIX/resources and referenced using a relative path from the esbmc binary folder in python_language.cpp. However, when using CMAKE_BINARY_DIR, the binary referenced during test execution is in the build folder (where CMake artifacts are generated). Consequently, astgen.py cannot be found, causing failures in the Python tests.

src/python-frontend/python_language.cpp
brcfarias: Nice. I will check.

scripts/cmake/Utils.cmake
fbrausse: > > This breaks my workflow
> 
> Could you please clarify which specific workflow? It is passing on CI, where we have a clean build.

My workflow is: `mkdir build && cd build && cmake [some-weird-flags] .. && ninja && ctest`. In essence, there is no `ninja install`, because up to now it was unnecessary.

> astgen.py is installed at CMAKE_INSTALL_PREFIX/resources

Indeed. In case we make the bundling/extraction work (as per the other comment somewhere below), I think this relative path would go away and `astgen.py` could live somewhere in `/tmp` (or whatever it is called on Windows).

CMakeLists.txt
brcfarias: > This suggestion is independent of build.sh

Yes, we can use -DDOWNLOAD_DEPENDENCIES independently of build.sh, however this option is not mentioned in the build documentation (BUILDING.md), which could pose difficulties for newcomers.
Should we update it?

mikhailramalho:
Some comments via email as I don't have access to my github account for
this week:

* Make the strings constexpr,
* I'm not sure you can make the unordered map constexpr as well, but please
try
* don't use 'using' in the code base, in this case for the json lib.


Also, I'm interested in knowing how are you going to handle dynamic
functions, as there is no support starting in the goto, all the way to the
smt backend.

Em seg., 9 de out. de 2023 11:02, Bruno Farias ***@***.***>
escreveu:

src/python-frontend/astgen.py
brcfarias: done

CMakeLists.txt
fbrausse: Yes, but doesn't have to be this PR.

rafaelsamenezes: > Both are not the case with DOWNLOAD_DEPENDENCIES as we have it now: it only does it in the local cmake build directory. @rafaelsamenezes Do you have a suggestion how we could work around this issue?

I would prefer to keep all side effects of `DOWNLOAD_DEPENDENCIES` to inside the build directory. If we need the dependency for building purposes (e.g. "python2goto")...  we could [set up a python virtual environment](https://discourse.cmake.org/t/possible-to-create-a-python-virtual-env-from-cmake-and-then-find-it-with-findpython3/1132) inside the build folder.

As a runtime dependency: I don't think it is the CMake responsability to install it as this will break sharing builds between systems. IMO when trying to parse Python, do a check if the package is available. If not then abort with an error message containing the pip install command. This shuld keep common users happy and advanced users will know how to adapt the command.

brcfarias: Agreed, I will remove this part from the CMakeLists and handle it in Python.

brcfarias: Done

src/python-frontend/python_language.cpp
brcfarias: I opted for bundling astgen.py. Thanks for this information.

scripts/cmake/Utils.cmake
brcfarias: > I think this relative path would go away

Done, so I reverted this modification.

lucasccordeiro (approved):

src/python-frontend/python_converter.cpp
```diff
@@ -0,0 +1,256 @@
...
+    auto value = element["value"];
+    if(element["value"].is_number_integer())
+    {
+      expr = from_integer(value.get<int>(), int_type());
+    }
+    break;
+  }
+  default:
+  {
+    assert(!"Unimplemented type in rule expression");
```
rafaelsamenezes: Shouldn't this be an abort? Remember that asserts are only triggered in Debug builds.

src/python-frontend/astgen.py
rafaelsamenezes: You could use the tmp_path class (in util/filesystem.h)

rafaelsamenezes (changes requested):
Just some nitpicks.

src/python-frontend/python_language.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return true;
+
+  const std::string python_script_path = dump_python_script() + "/astgen.py";
+
+  // Execute python script to generate json file from AST
+  std::string cmd("python3 " + python_script_path + " ");
+  cmd += path;
+
+  // Create a child process to execute Python
+  bp::child process(cmd);
```
fbrausse: Instead of building the single "command-line" style string, it is more robust to use the multi-string form giving each argument explicitly. That way the variables could also contain spaces or other symbols that would otherwise possibly interpreted by the shell.

fbrausse (commented):
Besides the hardcoded "program.py" (which is marked as future work, I'm fine with that) and the hardcoded "/tmp/ast.json", this looks great!

Regarding the latter, you could either make the output file a parameter to the `astgen.py` script and use the tmp_path as @rafaelsamenezes suggested, or the `astgen.py` script could instead print to stdout, see e.g. the example using boost's `ipstream` here: https://www.boost.org/doc/libs/1_83_0/doc/html/boost_process/tutorial.html#boost_process.tutorial.io.

src/python-frontend/python_converter.cpp

src/python-frontend/astgen.py
brcfarias: Done.

src/python-frontend/python_language.cpp
```diff
@@ -78,6 +79,12 @@ bool python_languaget::typecheck(
   contextt &context,
   const std::string & /*module*/)
 {
+  // Add the output directory for ast.json to the context, allowing its retrieval in python_converter
+  symbolt ast_output_path;
+  ast_output_path.id = "python_ast_path";
+  ast_output_path.value.set("path", ast_output_dir);
+  context.add(ast_output_path);
+
   python_converter converter(context);
```
fbrausse: Can't you just give the path to the constructor of the python_converter? The context shouldn't know about the temporary directory.

src/python-frontend/python_language.cpp
brcfarias: Done

brcfarias:
> you could either make the output file a parameter to the `astgen.py` script and use the tmp_path as @rafaelsamenezes suggested

Done.

fbrausse (commented):
Many thanks @brcfarias! One last detail.

src/python-frontend/python_language.cpp
brcfarias: Indeed, much simpler. Done.

fbrausse (approved):
LGTM.

lucasccordeiro:

PR #1391, opened 2023-10-09
Global Common Subexpression Elimination

rafaelsamenezes:
This PR add support for a Global Common Subexpression Elimination optimization inside ESBMC through `--gcse` flag. It is done by computing the Available Expressions of all program statements (using abstract interpretation) and then replacing the subexpressions with an intermediate variable.

The main benefit of this optimization is in codes that have complex repeated dereferences (see `regression/esbmc/gcse/main.c`) , which is hard for both the symbolic execution and to the solver. Specifically, replacing

```c
unsigned foo(table *tbl,
               unsigned EntryIndex)
{
  // ...
  tbl->RegMap[EntryIndex]
    .Auxiliary.Flags |= 1;
  tbl->RegMap[EntryIndex]
    .Auxiliary.WriteCount++;
  tbl->RegMap[EntryIndex]
    .Value = Data64;
  // ...
}
```

with

```c
unsigned foo(table *tbl,
               unsigned EntryIndex)
{
  // ...
  field *pRegmap = 
    &(tbl->RegMap[EntryIndex]);
  
  pRegmap->Auxiliary.Flags |= 1;
  pRegmap->Auxiliary.WriteCount++;
  pRegmap->Value = Data64;

  // ...
}
```

Caused a program that was taking 160s of symex and crashing during solving... to take 10s in total.

~~I am opening as draft for now as there is still some things left to do and some incorrect results that still needs to be solved.~~

The code still is missing a few features and has some limitations. However, I think that the core is stable enough to be merged already. This way I can start working on the features on smaller PRs.


fbrausse:
Since we've recently had changes to the data-race check: Ping @XLiZHI. Without having looked at the changes, yet, I think this might be incompatible with concurrency.

rafaelsamenezes:
@fbrausse yeah it will not work (same reason: the abstract interpreter is not thread aware). I forgot to tell that it's optional though.

mikhailramalho:
Quite interesting!

I suggest you also take a look at the 'gennaro patch' in our commit
history, where we add new variables for the guards: the idea was to
simplify the guard expressions when they become too big, but the patch was
never that smart

Em seg., 9 de out. de 2023 10:48, Rafael Sá Menezes <
***@***.***> escreveu:

lucasccordeiro (approved):
Great work, @rafaelsamenezes!

LGTM.

fbrausse:
Is this PR ready?

lucasccordeiro:
Yes ;-)

rafaelsamenezes:
> Is this PR ready?

For review, yes. I will just take a look at the unit tests that I forgot to enable. 

The entire thing still has lots of corner cases that will not work. However, I think that this PR will become overly complex for reviewing.

src/goto-programs/abstract-interpretation/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
-add_library(abstract-interpretation ai.cpp ai_domain.cpp interval_domain.cpp interval_analysis.cpp)
+add_library(abstract-interpretation ai.cpp ai_domain.cpp interval_domain.cpp interval_analysis.cpp common_subexpression_elimination.cpp)
```
fbrausse: I would prefer shorter file names. There could be a comment at the top of the .h or .cpp file saying "cse stands for [...]". The reason is that these long paths show up in the debugger everywhere and it gets harder to set breakpoints or look at stack traces.

src/util/algorithms.cpp
```diff
@@ -5,6 +5,8 @@
 
 bool goto_functions_algorithm::run(goto_functionst &goto_functions)
 {
+  _goto_functions = goto_functions;
+  runOnProgram(goto_functions);
```
fbrausse, line 8: This copy seems costly. Why is it necessary / where is it used?

src/pointer-analysis/value_set.cpp
```diff
@@ -414,7 +415,7 @@ void value_sett::get_value_set_rec(
 
     default:
       log_error("Unexpected side-effect: {}", *expr);
-      abort();
+      throw std::runtime_error("unsuported side-effect");
```
fbrausse: If this exception is never caught we should continue to abort().

If it is caught and handled somewhere then an own `value_set_error` or similar exception class would make the usage clearer.

src/pointer-analysis/value_set.cpp
```diff
@@ -1302,7 +1308,9 @@ void value_sett::assign_rec(
   }
   else
   {
-    throw std::runtime_error("assign NYI: `{}'" + get_expr_id(lhs));
+    log_error("[VSA] assign NYI: `{}'", get_expr_id(lhs));
+    lhs->dump();
+    throw std::runtime_error("unsupported NYI");
```
fbrausse: Same comment about exceptions as above: either abort() or specific class.

fbrausse (commented):
Haven't looked at the meat of this PR, yet. Just some comments from the "outside CSE" point of view.

src/util/algorithms.cpp
rafaelsamenezes: Hm, at the time it was because the Value-Set Analysis is a dependency for GCSE. I will invert the dependency though.

src/pointer-analysis/value_set.cpp
rafaelsamenezes: Added a class for the exception

fbrausse: Nice, thanks.

src/goto-programs/abstract-interpretation/CMakeLists.txt
fbrausse: Thanks, Rafeal.

src/goto-programs/abstract-interpretation/gcse.h
```diff
@@ -0,0 +1,167 @@
...
+#include <pointer-analysis/value_set_analysis.h>
+/**
+ * @brief Abstract domain to obtain all available expressions (AE)
+ *
+ * The domain is the set of expressions that were already computed up to this point.
+ *
+ * ... // AE: []
+ * int x = a + b + c // AE: [a + b, a + b + c]
+ * int y = a + b + d // AE: [a + b, a + b + c, a + b + d]
+ * c = 42 // AE: [a + b, a + b + d]
```
fbrausse, line 16: Is `a + b + c` missing for this last step?

src/goto-programs/abstract-interpretation/gcse.h
```diff
@@ -0,0 +1,167 @@
...
+    const goto_programt::const_targett &to,
+    std::unordered_set<expr2tc, irep2_hash> &matched_expressions) const;
+
+  symbolt
+  create_cse_symbol(const type2tc &t, const goto_programt::const_targett &to);
+
+private:
+  unsigned symbol_counter = 0;
+  const std::string prefix = "__ESBMC_cse_symbol";
+  std::shared_ptr<value_set_analysist> &vsa;
```
fbrausse: Why is this a shared_ptr if only a reference to it is stored? Can't it be a reference to the value_set_analysist directly?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
+#include "goto-programs/goto_program.h"
+#include "irep2/irep2_expr.h"
+#include "irep2/irep2_type.h"
+#include "pointer-analysis/value_set_analysis.h"
+#include "util/std_code.h"
+#include "util/std_expr.h"
```
fbrausse: How about including all these via the `<...>` form? The `"..."` one can't find them anyway in the search path starting from this file.

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+void cse_domaint::make_expression_available(const expr2tc &E)
+{
+  if(!E)
+    return;
+
+  // Logical expressions can be short-circuited
+  // So, only the first operand will be available
+  if(is_and2t(E) || is_or2t(E))
+  {
+    return;
```
fbrausse, line 138: The comment says that "only the first operand will be available". But the implementation skips the entire expression including all of its operands. Is this deliberate? Couldn't it instead defer to calling make_expression_available() with only the first operand?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  if(is_if2t(E))
```
fbrausse: Similar to above: The condition of the if2t could be made available, no?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  {
+    return;
+  }
+  // Skip sideeffects
+  if(is_sideeffect2t(E))
+    return;
+
+  // TODO: I hate floats
+  if(is_floatbv_type(E))
+    return;
```
fbrausse, line 152: Why? :)

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  // Did we check it already?
+  if(available_expressions.count(E))
+    return;
+
+  // Let's recursively make it available!
+  E->foreach_operand(
+    [this](const expr2tc &e) { make_expression_available(e); });
+
+  // TODO: LHS members should always be recomputed
+  if(is_with2t(E) || is_member2t(E) || is_dereference2t(E))
```
fbrausse: I don't understand the comment: What does "recomputed" mean here?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  if(is_sideeffect2t(E))
+    return;
+
+  // TODO: I hate floats
+  if(is_floatbv_type(E))
+    return;
+
+  // Did we check it already?
+  if(available_expressions.count(E))
+    return;
```
fbrausse, line 157: The pattern of `count()` followed by `insert()` in general is inefficient because the hash and the lookup need to computed twice. Would it be possible to make use of the `bool` member of the return value of `insert()` instead?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  }
+  for(auto x : to_remove)
+    available_expressions.erase(x);
+}
+
+void cse_domaint::havoc_expr(
+  const expr2tc &target,
+  const goto_programt::const_targett &i_it)
+{
+  if(is_dereference2t(target) && vsa != nullptr)
```
fbrausse: Can it ever happen that `vsa == nullptr`?

src/goto-programs/abstract-interpretation/gcse.cpp
```diff
@@ -0,0 +1,468 @@
...
+  {
+  case ASSIGN:
+  {
+    const code_assign2t &code = to_code_assign2t(instruction.code);
+    make_expression_available(code.source);
+    // Expressions that contain the target will need to be recomputed
+    havoc_expr(code.target, to);
+    // Target may be an expression as well
+    // TODO: skip recursive definitions only. For example, x = x + 1 should be skipped does not
+    // makes 'x' available. However, y = x + 1 does.
```
fbrausse: I can't parse this TODO sentence, could you clarify?

Also, IIUC, having an `x = x` assignment will first make `x` available (if it's not already), and then it will remove `x` again from the available expressions, with the result that `x` no longer is available though no state has changed. Is that correct and intentional?

fbrausse (commented):
The comments are really helpful! I'll finish the review once I understood my remaining open questions properly.

src/goto-programs/abstract-interpretation/gcse.h
rafaelsamenezes: It is not, `a+b+c` is not available anymore because `c` changed. So the expression would need to be recomputed again

fbrausse: Right, that makes sense. Thanks.

src/goto-programs/abstract-interpretation/gcse.h
rafaelsamenezes: Good call, this member was just used as a "passthrough" initialization for the domain. I removed it and now the domain vsa is initialized directly

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: The main issue was that I had trouble deciding what should be the spec for the intermediate variable e.g. an addition between a very small and a very big float. As I am not really versed in IEEE-754... I preferred to put it as a TODO.

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: Recomputed is my way of saying that something is technically available but we should recompute it anyway. For example:

```
int *arr = &some_arr[0];
arr[0] 42; // *(arr + 0) = 42
```

If we cache `*(arr + 0)` into some variable. The goto convertion will generate an invalid expression:

```
int *arr = &some_arr[0];
int deref = *(arr + 0);
deref = 42;
```

This is correct for RHS though, unless there is an assignment on it.

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: Yeah, if someone uses the avaliable expression domain without using the gcse class. I could convert it into an assertion... but I am not sure if that should be a hard requirement.

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: Broken english, I will update the comment only with the exception.

Id expressions such as `x=x` should be fine indeed. The issue of `x = x + 1` is that all previous dependencies of `x` are no longer available. If I add it back though, it will cause issues during the merge operation.

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: Oops

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: Done

src/goto-programs/abstract-interpretation/gcse.cpp
rafaelsamenezes: I changed the logic. Now we always add and remove the exceptions afterwards

fbrausse: With this change I don't quite follow the logic anymore. Besides still requiring 2 lookups (could be avoided by using the iterator returned from insert()), the more important change now is that any symbol added to the set isn't retained in it anymore once the recursion completes. Why wouldn't expressions stay available after `make_expression_available()`?

rafaelsamenezes: I don't want to add every expression (see https://github.com/esbmc/esbmc/pull/1391#discussion_r1412785917). And now the lookup will only happen twice in the corner cases.

rafaelsamenezes:
I will be merging this since:

- It is getting out of sync of master ESBMC.
- It is an optional feature
- It is simple to revert

@fbrausse I still want to have a meeting with you to discuss the details and issues of the code.

lucasccordeiro (approved):

src/goto-programs/abstract-interpretation/gcse.cpp
fbrausse: Ignore second part of comment, I was misunderstanding things.

fbrausse (approved):
LGTM.

Issue #1392, opened 2023-10-10
Varying results for out-of-range casts


Fleischmaki:
Hi, consider the following program (test.c)
```
extern void __VERIFIER_error(void);
extern unsigned int __VERIFIER_nondet_uint(void);

int main(){
        unsigned int n = __VERIFIER_nondet_uint();
        if(!(0 <= (int) (n))){
                if(846 <= n){
                        __VERIFIER_error();
                }
        }
        return 0;
}
```
Running `esbmc bug.c` gives `VERIFICATION FAILED`,
but running `esbmc --interval-analysis bug.c` gives `VERIFICATION_SUCCESFUL`
Not sure if this is intented as the counterxample given is `n = 4294967295` so casting it to signed int gives undefined behaviour iirc. (as the value exceeds the value range for signed integers)
Commit version: e5e00743dceed71bf7a48cc3f34887bf0a0955b8
OS: Ubuntu 20.04

rafaelsamenezes:
Just confirmed it on Integer domain (wrapped domain is fine).

Update: the issue is that we get rid of the typecasts when canonizing the path condition into CNF.

lucasccordeiro:
Closed via #1393.

PR #1393, opened 2023-10-11
CNF generation no longer removes typecasts

rafaelsamenezes:
As reported at #1392, the CNF generation was removing typecast from formulas. This PR disables it.

This will result in the following new limitations for an expression such as `0 < (unsigned) n`:

1. The contractor only works on symbols. This means that it will not add any restriction over `n`.
2. For Integer intervals, getting the interval of `(unsigned) n` will result in top (typecast are not supported). 

Results (PR):
```
  correct:            4003
    correct true:     2559
    correct false:    1444
  incorrect:             7
    incorrect true:      7
    incorrect false:     0
  unknown:            5527
  Score:              6338 (max: 15923)
```

I will update once I have the master results.


lucasccordeiro (approved):

lucasccordeiro:

PR #1396, opened 2023-10-17
Python frontend: If/else support

brcfarias:
This PR adds:
1. if/else statement handling
2. Python AST display via `--parse-tree-{only,too}` options


lucasccordeiro (changes requested):

Can you add regression tests that include non-determinism?

Can you also write some documentation about these developed methods?

rafaelsamenezes (approved):
Overall, I just did not like exposing the `nlohmann::json` to the `python_converter` class. However, I am not sure how to avoid it.

src/esbmc/options.cpp
```diff
@@ -59,6 +59,12 @@ const struct group_opt_templ all_cmd_options[] = {
       boost::program_options::value<std::string>()->value_name("cname"),
       "set contract name"},
    }},
+#endif
+#ifdef ENABLE_PYTHON_FRONTEND
+  {"Python frontend",
+   {
+     {"show-python-ast", NULL, "show Python AST"},
```
fbrausse: I wonder whether we can re-use the existing --parse-tree-{only,too} options instead of adding a new but semantically equivalent option.

brcfarias:
> Overall, I just did not like exposing the `nlohmann::json` to the `python_converter` class. However, I am not sure how to avoid it.

Agreed. I intend to refactor it after implementing a basic set of functionalities, aiming to encapsulate JSON access, similar to what you did in `jimple_ast`.

src/esbmc/options.cpp
brcfarias: I think so. I'll take a look.

brcfarias: Done. Thank you.

lucasccordeiro (approved):

fbrausse (approved):

lucasccordeiro:

Issue #1335, opened 2023-09-12
ESBMC explores all switch cases if passing a struct member


Anthonysdu:
```
#include <assert.h>

struct temp{
    long x;
};
    
void loop1(){
    for(int i=0;i<1;i++){
        assert(1);
    }
}


void loop2(){
    for(int i=0;i<1;i++){
        assert(1);
    }
}

void fun(unsigned long x){

    assert(x == 1);

    switch(x){
    case 1:
        loop1();
        return;
    case 2:
        loop2();
        return;
    default:
        return;
    }
}

int main(){
    struct temp temp1;
    temp1.x = 1;
    fun(temp1.x);
    return 0;
}
```
`esbmc main.c:`

```
ESBMC version 7.3.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing main.c
parse() time: 0.011s
Converting
typecheck(): 0.003s
clear_parse(): 0.001s
Generating GOTO Program
goto_convert() time: 0.002s
GOTO program creation time: 0.111s
GOTO program processing time: 0.001s
Starting Bounded Model Checking
Unwinding loop 2 iteration 1   file main.c line 8 column 5 function loop1
Unwinding loop 3 iteration 1   file main.c line 15 column 5 function loop2
Symex completed in: 0.001s (18 assignments)
Slicing time: 0.000s (removed 17 assignments)
Generated 2 VCC(s), 0 remaining after simplification (1 assignments)
BMC program time: 0.001s

VERIFICATION SUCCESSFUL
```

but if I replace `fun(temp1.x)` with `fun(1)` or `int x = 1; fun(x)` it will only go into case1 (unwind loop1).
Is it a bug or as designed? Since this can potencially cause huge wind overhead.

fbrausse:
Happens for unions and arrays, too:
```c
    unsigned long x[1];
    x[0] = 1;
    fun(x[0]);
```
and
```c
    union { unsigned long x; } temp;
    temp.x = 1;
    fun(temp.x);
```
But only if they're assigned a value after initialization. With direct initialization only a single unwinding is produced (for union, array and struct), e.g.
```c
    union { unsigned long x; } temp = { 1 };
    fun(temp.x);
```
However, for `unsigned long x; x = 1; fun(x);` also just a single unwinding is performed. To me it looks like there is some special-casing going on in constant propagation. Don't know where exactly, though. Probably a check whether it is a constant_int2t.

[...]

PR #1398, opened 2023-10-17
enable const propagation for with2t

Anthonysdu:
This PR trys to re-enable constant propagation for with2t, as discussed in #1335 

There is a significant performance down for arrays in my testcases, maybe we could just enable const propagation for with2t structs only, or we need to find a way to refine propagation for arrays?


src/util/simplify_expr2.cpp
```diff
@@ -852,6 +852,17 @@ expr2tc with2t::do_simplify() const
 
 expr2tc member2t::do_simplify() const
 {
+  expr2tc current_source = source_value;
+  while(is_with2t(current_source))
+  {
+    const with2t &with = to_with2t(current_source);
+    if(member == to_constant_string2t(with.update_field).value)
```
fbrausse, line 859: I am surprised that this works. The `with2t::update_field` (unfortunately) can also be an integer-typed expression in case the `source_value` is an array. Could this be a reason for the slow-down you see?

fbrausse (commented):
Are the performance degredations only happening with the two regression tests you added or which ones were you referring to?

As a comment regarding the simplification algorithm: I don't think the loop you added should come sequentially before the other simplifications. It should encompass all the other simplifications.

src/util/simplify_expr2.cpp
fbrausse: How are you measuring "performance down"? What's the `CMAKE_BUILD_TYPE` you're using for that?

lucasccordeiro:
@Anthonysdu: Can I ask you for the status of this PR?

src/util/simplify_expr2.cpp
Anthonysdu: > I am surprised that this works. The `with2t::update_field` (unfortunately) can also be an integer-typed expression in case the `source_value` is an array. Could this be a reason for the slow-down you see?

Do you mean the struct like with this structure?
```
struct stru{
    int arr[2];
};
```
And also we need to const propag for array:
```
int arr[2];
arr[0] = 1;
fun_array(arr);
```

lucasccordeiro:
@Anthonysdu: do you have plans to work on this PR?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2128.

PR #1403, opened 2023-10-18
Add and default to `DebugOpt` build type for faster PRs

fbrausse:
This PR adds the new value `DebugOpt` for the CMAKE_BUILD_TYPE variable. It contains the flags "-O2 -g". The difference to the existing `RelWithDebInfo` is that `-DNDEBUG` is *not* part of `DebugOpt`. It thus keeps all internal assertions, but compiles with -O2. Also switch the default from `Debug` to `DebugOpt` for the CI.

Locally, I go from Debug:
```
time ninja -j16
real	4m22,021s
user	51m27,560s
sys	2m55,290s

time ctest -j15
real	19m36,899s
user	257m17,031s
sys	5m50,710s
```
to DebugOpt:
```
time ninja -j16
real	5m48,021s
user	74m6,785s
sys	3m25,686s

time ctest -j15
real	9m49,003s
user	122m58,941s
sys	5m4,590s
```
My hope is that the CI will be faster. It might also result in some -Werrors since optimization enables more of the compiler's checks and/or in some failed test cases. I would like to ask for help in fixing those in code I'm not too familiar with.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -35,7 +35,7 @@ jobs:
     uses: ./.github/workflows/build.yml
     with:
       operating-system: ${{ matrix.os }}
-      build-flags: '-b Debug -e On'
+      build-flags: '-b DebugOpt -e On'
       testing: true
       
     # Check project with clang-format
@@ -71,4 +71,4 @@ jobs:
       run: pip install cmakelint
     - name: Run CMake Lint
       run: find . -iname *.cmake -o -iname CMakeLists.txt | grep -v build | xargs cmakelint || echo "failure"
-      
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,10 +30,21 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
   set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
       STRING "Choose the type of build." FORCE)
-  # Set the possible values of build type for cmake-gui
-  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
-    "Debug" "Release" "MinSizeRel" "RelWithDebInfo" "Sanitizer")
 endif()
+
+# Set the possible values of build type for cmake-gui
+set(CMAKE_CXX_FLAGS_DEBUGOPT "-O2 -g" CACHE STRING "C++ compiler flags for DebugOpt build" FORCE)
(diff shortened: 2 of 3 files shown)


fbrausse:
@rafaelsamenezes Could you take a look?
```
In file included from /home/runner/work/esbmc/esbmc/unit/goto-programs/interval_template.test.cpp:1:
/home/runner/work/esbmc/esbmc/src/goto-programs/abstract-interpretation/interval_template.h: In function ‘interval_templatet<int> operator*(const interval_templatet<int>&, const interval_templatet<int>&)’:
/home/runner/work/esbmc/esbmc/src/goto-programs/abstract-interpretation/interval_template.h:18:7: error: ‘result.interval_templatet<int>::lower’ may be used uninitialized [-Werror=maybe-uninitialized]
   18 | class interval_templatet
      |       ^~~~~~~~~~~~~~~~~~
/home/runner/work/esbmc/esbmc/src/goto-programs/abstract-interpretation/interval_template.h:343:27: note: ‘result.interval_templatet<int>::lower’ was declared here
  343 |     interval_templatet<T> result;
      |                           ^~~~~~
```
https://github.com/esbmc/esbmc/actions/runs/6564879380/job/17832190303?pr=1403

rafaelsamenezes (commented):
I will check the interval issue.

Once this is merged, it would be good to extend the windows script to use this. I usually had to manually override the build flags to remove NDEBUG

rafaelsamenezes (approved):

lucasccordeiro (approved):

fbrausse:
Thanks, Rafael.

> Once this is merged, it would be good to extend the windows script to use this.

I agree. If it's easier, I can push this branch to this repository so we can all push commits there. Would probably need to open a new PR, though.

fbrausse:
@rafaelsamenezes I think the maybe_uninitialized warning is not a false positive but actually happens when the synthesized copy-ctor copies `lower` even though `lower_set` is false. It should be possible to avoid that one by combining both into a member of type `std::optional<T>` (which internally is supposed to properly handle initialization of the storage associated with T). Though that may be a too intrusive change of the API.

rafaelsamenezes:
> t should be possible to avoid that one by combining both into a member of type std::optional<T> (which internally is supposed to properly handle initialization of the storage associated with T). Though that may be a too intrusive change of the API.

I like this change actually.

rafaelsamenezes:
@fbrausse could you rebase and check whether the Intervals are still giving an error?

fbrausse:
@rafaelsamenezes Works flawlessly, many thanks!

fbrausse:
For the Ubuntu CI we now have ~29m build and ~41m running tests. That's much better than before (#1407): ~21m build and 1h 24m running tests.

PR #1404, opened 2023-10-19
Update BUILDING.md

fbrausse:
- Restore instructions for building statically against an external Clang.
- Mention DOWNLOAD_DEPENDENCIES and what it does.
- Add instruction how to build shared against a system Clang.
- Annotate / Remark on optional parts of these instructions.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -88,7 +88,7 @@ They are optional.
 For shared builds it is recommended to use the system's LLVM/Clang, which on
 Ubuntu can be obtained by:
 ```
-apt-get install libclang-cpp-dev &&
+apt-get install libclang-cpp11-dev &&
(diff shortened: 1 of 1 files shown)


fbrausse:
It might make sense in addition to these changes, to have an interactive install script, which explains what's going on and also asks questions like
- Do you want a static or shared build?
- I found 5 different LLVMs: [...] Which one should I use?
- Do you want contractors/python/etc. support?
- Which of these solvers? [...]

fbrausse:
We cannot reliably build a static version of ESBMC without building all the dependencies statically as well. However, a shared build is possible on most distros against the distro's packaged versions. So I'm wondering why the BUILDING.md file promotes a static build and explains how to build the dependencies as well, given that it does require installing the distro's build tools anyway.

lucasccordeiro (approved):

PR #1406, opened 2023-10-19
Python frontend: while statements

brcfarias:
This PR:

- Add support for **while** statements
- Add support for **compound assignments** (+=, -=, etc)
- Include **Python filename** information in the JSON

### DIFF
--- a/regression/python/compound-assign/main.py
+++ b/regression/python/compound-assign/main.py
@@ -3,4 +3,3 @@ count *= 2
 count /= 2
 count += 1
 count -= 1
-result : int = 1/1
--- a/regression/python/compound-assign/main.py
+++ b/regression/python/compound-assign/main.py
@@ -1,0 +1,6 @@
+count : int = 1
+count *= 2
+count /= 2
+count += 1
+count -= 1
+result : int = 1/1
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -299,35 +299,6 @@ void python_converter::get_compound_assign(
   target_block.copy_to_operands(code_assign);
 }
 
-exprt python_converter::get_block(const nlohmann::json &ast_block)
-{
-  code_blockt block;
-
-  // Iterate through the statements of the block
-  for(auto &element : ast_block)
-  {
-    StatementType type = get_statement_type(element);
-
-    switch(type)
-    {
-    case StatementType::VARIABLE_ASSIGN:
-    {
-      // Add an assignment to the block
-      get_var_assign(element, block);
-      break;
-    }
-    case StatementType::FUNC_DEFINITION:
-    case StatementType::IF_STATEMENT:
-    case StatementType::UNKNOWN:
-    default:
-      log_error("error");
-      abort();
-    }
-  }
-
-  return block;
-}
-
 static codet convert_expression_to_code(exprt &expr)
 {
   if(expr.is_code())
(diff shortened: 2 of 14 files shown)


regression/python/compound-assign/main.py
```diff
@@ -0,0 +1,6 @@
+count : int = 1
+count *= 2
+count /= 2
+count += 1
+count -= 1
+result : int = 1/1
```
fbrausse: Shouldn't this involve `count`?

src/python-frontend/python_converter.cpp
```diff
@@ -1,4 +1,5 @@
 #include "python-frontend/python_converter.h"
+#include "python_frontend_types.h"
 #include "util/std_code.h"
 #include "util/c_types.h"
 #include "util/arith_tools.h"
```
fbrausse: Sorry, I'm seeing this just now: Could you instead of including via the quoted form `"..."` please use the angle brackets form `<...>`? The `src` directory is part of the include paths during building, there is no need to use the quoted form. (applies to all the includes for the python frontend).

src/python-frontend/python_converter.cpp
```diff
@@ -240,73 +204,85 @@ void python_converter::get_var_assign(
...
-    switch(type)
-    {
-    case StatementType::VARIABLE_ASSIGN:
+    context.add(symbol);
+  }
+  else if(ast_node["_type"] == "Assign")
+  {
+    std::string id = ast_node["targets"][0]["id"].get<std::string>();
+    symbolt *symbol = context.find_symbol(std::string("py:@") + id);
+    if(symbol != nullptr)
```
fbrausse: What happens otherwise? Should it even be possible that an assignment to an unknown symbol is being performed and, if so, what's supposed to happen then? I'd suggest to assert this condition.

src/python-frontend/python_converter.cpp
```diff
@@ -337,42 +313,75 @@ void python_converter::get_if_statement(
...
     }
   }
 
-  // add "main"
+  return block;
+}
+
+bool python_converter::convert()
+{
+  current_function_name = "globalscope";
```
fbrausse: I wonder what this string is being used for. Given that `globalscope` could well be a valid Python function's name it seems confusing.

src/python-frontend/python_frontend_types.h
```diff
@@ -0,0 +1,46 @@
...
+enum class ExpressionType
+{
+  UNARY_OPERATION,
+  BINARY_OPERATION,
+  LITERAL,
+  VARIABLE_REF,
+  UNKNOWN,
+};
+
+static const std::unordered_map<std::string, std::string> operator_map = {
```
fbrausse: Type definitions like the enums above are great to have in header files, but symbol definitions like this are bad in the sense that they get duplicated in every C++ translation unit this header is included in. Meaning that the resulting ESBMC binary would carry multiple versions of this symbol. In case this symbol is indeed used by multiple different C++ files, would you consider making this symbol `extern` and defining it only once in some translation unit? If it's used only once, please remove it from the header. Same applies to `statement_map` below.

src/python-frontend/python_converter.cpp
I was considering this: https://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html#SEC6
"" for own headers and <> for system headers.

src/python-frontend/python_converter.cpp
brcfarias: > I'd suggest to assert this condition.

Done

src/python-frontend/python_converter.cpp
brcfarias: It would be used to identify symbols (statements) that are not inside functions. However, since it is not being used at the moment, I have removed it. In fact, it might be better to leave the function name part empty in the symbol ids of "global scope" to avoid this kind of collision.

regression/python/compound-assign/main.py
brcfarias: Just removed this line.

lucasccordeiro (approved):

src/python-frontend/python_frontend_types.h
brcfarias: > If it's used only once, please remove it from the header.

Done.

src/python-frontend/python_converter.cpp
fbrausse: > "" for own headers and <> for system headers.

That's a useful interpretation when the project is organized such that the "own" headers are internal and placed alongside the sources in the same directory (or a sub-directory thereof). This is nearly, but not fully, the case for ESBMC. In fact the "..." variant just adds the current directory as a first search path (and also considers the ones given by `-iquote`, but we don't use that compiler option). But for ESBMC's convention that wouldn't help: `#include "util/std_code.h"` refers to the src/util sub-directory, not a subdir of the current one. The convention is to specify the full path under src/ and use the angle-bracket form.

See items 1.-6. at https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html for how the search paths are built and used.

fbrausse (approved):
Thanks for addressing my comments. LGTM.

src/python-frontend/python_converter.cpp
```diff
@@ -337,42 +337,73 @@ void python_converter::get_if_statement(
...
+    case StatementType::COMPOUND_ASSIGN:
     {
-      get_if_statement(element, main_code);
+      get_compound_assign(element, block);
+      break;
+    }
+    case StatementType::FUNC_DEFINITION:
+    case StatementType::UNKNOWN:
+    default:
+      log_error("error");
```
rafaelsamenezes, line 396: Could you add a more specific error message?

src/python-frontend/python_converter.cpp
```diff
@@ -20,38 +21,26 @@ static const std::unordered_map<std::string, std::string> operator_map = {
...
   {"USub", "unary-"},
-  {"Eq", "="}};
-
-enum class StatementType
-{
-  VARIABLE_ASSIGN,
-  FUNC_DEFINITION,
-  IF_STATEMENT,
-  UNKNOWN,
+  {"Eq", "="},
```
rafaelsamenezes, line 24: Could you create a test case for equality?

rafaelsamenezes (changes requested):

PR #1407, opened 2023-10-19
Interval now uses std::optional to keep the bounds

rafaelsamenezes:
As reported in https://github.com/esbmc/esbmc/pull/1403, the `lower/upper` and `lower_set/upper_set` had an unhealthy relation that could lead to some unitialized uses. This PR removes the `_set` variables and replaces all `lower/upper` uses into an `std::optional`. Besides... during the refactoring I found some issues that are now fixed:

- the `has_changed` function used to check the upper bound incorrectly;
- the interpolation function was not using infinity on the previous interval as a precondition.

I will also start a reachability run.

### DIFF
--- a/src/goto-programs/abstract-interpretation/interval_domain.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_domain.cpp
@@ -75,9 +75,9 @@ void interval_domaint::apply_assume_symbol_truth(
     interval.make_ge_than(1);
   else if(is_signedbv_type(sym.type))
   {
-    if(interval.lower_set && interval.get_lower() == 0)
+    if(interval.lower && interval.get_lower() == 0)
       interval.make_ge_than(1);
-    else if(interval.upper_set && interval.get_upper() == 0)
+    else if(interval.upper && interval.get_upper() == 0)
       interval.make_le_than(-1);
   }
 
@@ -222,13 +222,10 @@ template <class T>
 T interval_domaint::extrapolate_intervals(const T &before, const T &after)
 {
   T result; // Full extrapolation
-
-  bool lower_decreased =
-    !after.lower_set ||
-    (before.lower_set && after.lower_set && after.lower < before.lower);
-  bool upper_increased =
-    !after.upper_set ||
-    (before.upper_set && after.upper_set && after.upper > before.upper);
+  bool lower_decreased = !after.lower || (before.lower && after.lower &&
+                                          *after.lower < *before.lower);
+  bool upper_increased = !after.upper || (before.upper && after.upper &&
+                                          *after.upper > *before.upper);
 
   if((lower_decreased || upper_increased) && !widening_under_approximate_bound)
     return result;
@@ -236,17 +233,20 @@ T interval_domaint::extrapolate_intervals(const T &before, const T &after)
   // Set lower bound: if we didn't decrease then just update the interval
   if(!lower_decreased)
   {
-    result.lower_set = before.lower_set;
-    result.lower = before.lower;
+    if(!before.lower)
+      result.lower.reset();
+    else
+      result.lower = *before.lower;
   }
 
   // Set upper bound:
   if(!upper_increased)
   {
-    result.upper_set = before.upper_set;
-    result.upper = before.upper;
+    if(!before.upper)
+      result.upper.reset();
+    else
+      result.upper = *before.upper;
   }
-
   return result;
 }
 
@@ -278,15 +278,15 @@ T interval_domaint::interpolate_intervals(const T &before, const T &after)
 {
   T result;
 
-  bool lower_increased = !before.lower_set;
-  bool upper_decreased = !before.upper_set;
+  // before: [-infinity, +infinity], after: [a,b] ==> [a,b]
+  bool lower_increased = !before.lower && after.lower;
+  bool upper_decreased = !before.upper && after.upper;
 
-  result.lower_set = lower_increased ? after.lower_set : before.lower_set;
-  result.lower = lower_increased ? after.lower : before.lower;
+  if(lower_increased)
+    result.lower = *after.lower;
 
-  result.upper_set = upper_decreased ? after.upper_set : before.upper_set;
-  result.upper = upper_decreased ? after.upper : before.upper;
-
+  if(upper_decreased)
+    result.upper = *after.upper;
   return result;
 }
 
@@ -582,10 +582,10 @@ void interval_domaint::apply_assume_less(const expr2tc &a, const expr2tc &b)
       lhs.make_sound_le(rhs);
     else
     {
-      if(rhs.upper_set)
+      if(rhs.upper)
         lhs.make_le_than(rhs.get_upper());
 
-      if(lhs.lower_set)
+      if(lhs.lower)
         rhs.make_ge_than(lhs.get_lower());
     }
   }
@@ -648,9 +648,9 @@ template <>
 expr2tc interval_domaint::make_expression_value<integer_intervalt>(
   const integer_intervalt &interval,
   const type2tc &type,
-  bool upper) const
+  bool is_upper) const
 {
-  return from_integer(upper ? interval.upper : interval.lower, type);
+  return from_integer(is_upper ? *interval.upper : *interval.lower, type);
 }
 
 template <>
@@ -662,7 +662,8 @@ expr2tc interval_domaint::make_expression_value<real_intervalt>(
   expr2tc value = gen_zero(type);
   constant_floatbv2t &v = to_constant_floatbv2t(value);
 
-  const auto d = (upper ? interval.upper : interval.lower).convert_to<double>();
+  const auto d =
+    (upper ? *interval.upper : *interval.lower).convert_to<double>();
   v.value.from_double(d);
 
   // 'from_double' changes the original spec. This makes solvers complain that we are comparing
@@ -716,12 +717,12 @@ expr2tc interval_domaint::make_expression_helper<wrapped_interval>(
   }
   else
   {
-    assert(interval.upper_set && interval.lower_set);
+    assert(interval.upper && interval.lower);
     // Interval: [a,b]
     std::vector<expr2tc> disjuncts;
 
     auto convert = [this, &src, &symbol, &disjuncts](wrapped_interval &w) {
-      assert(w.lower <= w.upper);
+      assert(*w.lower <= *w.upper);
 
       std::vector<expr2tc> s_conjuncts;
       expr2tc value1 = make_expression_value(w, src.type, true);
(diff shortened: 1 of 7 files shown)


src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -222,31 +222,31 @@ template <class T>
...
 
   // Set lower bound: if we didn't decrease then just update the interval
   if(!lower_decreased)
   {
-    result.lower_set = before.lower_set;
-    result.lower = before.lower;
+    if(!before.lower)
+      result.lower.reset();
+    else
+      result.lower = *before.lower;
```
fbrausse: This could be simplified to `result.lower = before.lower`. When the RHS is empty, so will the LHS be after this.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -222,31 +222,31 @@ template <class T>
...
 
   // Set upper bound:
   if(!upper_increased)
   {
-    result.upper_set = before.upper_set;
-    result.upper = before.upper;
+    if(!before.upper)
+      result.upper.reset();
+    else
+      result.upper = *before.upper;
```
fbrausse: Same as above.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -278,15 +278,15 @@ T interval_domaint::interpolate_intervals(const T &before, const T &after)
 {
   T result;
 
-  bool lower_increased = !before.lower_set;
-  bool upper_decreased = !before.upper_set;
-
-  result.lower_set = lower_increased ? after.lower_set : before.lower_set;
-  result.lower = lower_increased ? after.lower : before.lower;
+  // before: [-infinity, +infinity], after: [a,b] ==> [a,b]
+  bool lower_increased = !before.lower && after.lower;
```
fbrausse, line 272: Unless I'm missing something, isn't this changing the semantics from what essentially was `before.lower_set || after.lower_set` earlier?

src/goto-programs/abstract-interpretation/interval_template.h
```diff
@@ -20,29 +21,25 @@ class interval_templatet
 public:
   virtual ~interval_templatet() = default;
 
-  interval_templatet() : lower_set(false), upper_set(false)
+  interval_templatet()
```
fbrausse: `= default`?

src/goto-programs/abstract-interpretation/interval_template.h
```diff
@@ -20,29 +21,25 @@ class interval_templatet
...
 
-  explicit interval_templatet(const T &l, const T &u)
-    : lower_set(true), upper_set(true), lower(l), upper(u)
+  explicit interval_templatet(const T &l, const T &u) : lower(l), upper(u)
   {
   }
 
   /* TODO: There is a clear dependency between the lower/upper and
   *        and lower_set/upper_set variable. Shouldn't we convert the
   *        the uses into the constraints functions?
```
fbrausse: This TODO seems to be taken care of now, no?

src/goto-programs/abstract-interpretation/interval_template.h
```diff
@@ -117,7 +113,7 @@ class interval_templatet
  */
   virtual bool empty() const
   {
-    return !is_top() && (upper_set && lower_set && lower > upper);
+    return !is_top() && (lower && upper && *lower > *upper);
```
fbrausse: The `!is_top()` condition here does not seem to be needed (it's implied by either side of the conjunction `lower && upper`).

Additionally, wrt. to going to `double` eventually: are you considering NaNs or other partially ordered types? If so, I suggest changing `*lower > *upper` to `!(*lower <= *upper)` because comparisons on NaNs are always false.

src/goto-programs/abstract-interpretation/interval_template.h
```diff
@@ -169,23 +164,28 @@ class interval_templatet
   // Sound version (considering over approximations)
   void make_sound_le(interval_templatet<T> &v)
   {
-    upper_set = upper_set || v.upper_set;
-    upper = std::min(upper, v.upper);
-    v.lower_set = lower_set && v.lower_set;
-    v.lower = std::max(lower, v.lower);
+    // [lower, upper] <= [v.lower,v.upper] <==> [lower, min(upper,v.upper)] <= [max(lower, v.lower)]
```
fbrausse: The last interval in this comment seems incomplete. `v.upper` missing maybe?

src/goto-programs/abstract-interpretation/interval_template.h
```diff
@@ -643,20 +633,15 @@ class interval_templatet
...
 
-    if(lower_set && upper_set)
-      return !(
-        lower_set_equal && upper_set_equal && lower_equal && upper_equal);
+    if((lower != i.lower) || (upper != i.upper))
+      return true;
 
-    if(lower_set)
-      return !(upper_set_equal && upper_equal);
+    if(lower && *lower != *i.lower)
```
fbrausse: I don't understand how this condition could be true when the above one `lower != i.lower` hasn't already been. Did you maybe mean to compare the `.has_value()` above?

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -50,11 +48,11 @@ class wrapped_interval : public interval_templatet<BigInt>
     if(is_signed)
     {
       auto middle = get_upper_bound() / 2; // [128] 256
-      if(value >= middle)
-        value -= middle * 2;
+      if(*value >= middle)
+        value = *value - middle * 2;
```
fbrausse: You could use `*value -= [...]`. No need to re-init the std::optional.

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -78,10 +76,10 @@ class wrapped_interval : public interval_templatet<BigInt>
     if(!w.is_bottom())
     {
       auto mod = w.get_upper_bound();
-      result.lower = (w.upper + 1) % mod;
-      result.upper = (w.lower - 1) % mod;
+      result.lower = (*w.upper + 1) % mod;
+      result.upper = (*w.lower - 1) % mod;
       if(result.upper < 0)
-        result.upper += mod;
+        result.upper = *result.upper + mod;
```
fbrausse: Same `*result.upper += [...]` as above.

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -607,14 +605,14 @@ class wrapped_interval : public interval_templatet<BigInt>
...
-      if(result.upper < 0)
-        result.upper += mod;
-      assert(result.lower >= 0);
-      assert(result.upper >= 0);
+      result.lower = (*lhs.lower - *rhs.upper) % mod;
+      result.upper = (*lhs.upper - *rhs.lower) % mod;
+      if(*result.lower < 0)
+        result.lower = *result.lower + mod;
+      if(*result.upper < 0)
+        result.upper = *result.upper + mod;
```
fbrausse: Same as above: `*result.something += [...]`.

fbrausse (commented):
Nice! I have a few comments though (and some nitpicks).

src/goto-programs/abstract-interpretation/interval_template.h
rafaelsamenezes: Nice catch, I can just keep the initial condition actually.

src/goto-programs/abstract-interpretation/interval_domain.cpp
rafaelsamenezes: Yeah, the code was wrong (we still need to add more tests for extrapolation/interpolation). We can only interpolate when coming from an infinity.

src/goto-programs/abstract-interpretation/interval_template.h
rafaelsamenezes: Yeah, the implementation was also wrong

src/goto-programs/abstract-interpretation/interval_domain.cpp
fbrausse: Documentation of these methods could help. I'm not clear on what they are actually supposed to do.

rafaelsamenezes:
This PR (30s):

```
  correct:            4042
    correct true:     2563
    correct false:    1479
  incorrect:             7
    incorrect true:      7
    incorrect false:     0
  unknown:            5488
  Score:              6381 (max: 15923)
```

Master (30s):

```
  correct:            3994
    correct true:     2549
    correct false:    1445
  incorrect:            16
    incorrect true:      6
    incorrect false:    10
  unknown:            5527
  Score:              6191 (max: 15923)
```

src/goto-programs/abstract-interpretation/interval_template.h
fbrausse: Is `[max(lower, v.lower)]` meant to describe a singleton interval containing only one point? *Edit:* I'm talking about the comment in this line.

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -463,10 +461,10 @@ class wrapped_interval : public interval_templatet<BigInt>
...
       return t;
 
-    const BigInt a = s.lower;
-    const BigInt b = s.upper;
-    const BigInt c = t.lower;
-    const BigInt d = t.upper;
+    const BigInt a = *s.lower;
+    const BigInt b = *s.upper;
+    const BigInt c = *t.lower;
+    const BigInt d = *t.upper;
```
fbrausse: Could you use const-reference here to avoid copies?

lucasccordeiro (approved):

I agree with @fbrausse's suggestions. So, please ensure that you address them.

src/goto-programs/abstract-interpretation/interval_template.h
rafaelsamenezes: Just updated the comment

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -396,10 +394,10 @@ class wrapped_interval : public interval_templatet<BigInt>
...
     }
 
-    const BigInt a = s.lower;
-    const BigInt b = s.upper;
-    const BigInt c = t.lower;
-    const BigInt d = t.upper;
+    const BigInt a = *s.lower;
+    const BigInt b = *s.upper;
+    const BigInt c = *t.lower;
+    const BigInt d = *t.upper;
```
fbrausse: Here too, please. :)

fbrausse (approved):
One last comment, otherwise LGTM.

src/goto-programs/abstract-interpretation/interval_domain.cpp
rafaelsamenezes: I added a bit more (and an example) on the docs. Let me know what you think

fbrausse: Thank you, that helps a lot!

lucasccordeiro:

Issue #1383, opened 2023-09-28
Failing regression tests for a Sanitizer build


fbrausse:
When running a CMAKE_BUILD_TYPE=Sanitizer build on Linux over the regression tests, I'm getting `573 tests failed`. Summary:
- 19x in esbmc-solidity
- ~39x in esbmc-cpp11~
- ~352x in esbmc-cpp~
- ~138x in cuda~

The remaining ones:
- Interval Analysis - Add Arithmetic (Failed)
- Interval Analysis - Add Range Arithmetic (Failed)
- Interval Analysis - Arithmetic Right Shift (Failed)
- Interval Analysis - Base Sign (Failed)
- Interval Analysis - Base Unsigned (Failed)
- Interval Analysis - Bitand (Failed)
- Interval Analysis - Bitor (Failed)
- Interval Analysis - Div Arithmetic (signed) (Failed)
- Interval Analysis - If/Else Statement LHS (Negative) (Failed)
- Interval Analysis - If/Else Statement LHS (Signed) (Failed)
- Interval Analysis - If/Else Statement LHS (Unsigned) (Failed)
- Interval Analysis - If/Else Statement LHS and RHS (Failed)
- Interval Analysis - If/Else Statement RHS (Failed)
- Interval Analysis - If/Else Statement RHS (Negative) (Failed)
- Interval Analysis - Left Shift (Failed)
- Interval Analysis - Right Shift (Failed)
- Interval Analysis - Sub Arithmetic (Failed)
- Interval Analysis - Ternary (Failed)
- Interval Analysis - Ternary (false) (Failed)
- Interval Analysis - Ternary (true) (Failed)
- Interval Analysis - While Statement (Failed)
- Scenario: the loop unroller detects bounded loops (Failed)
- regression/esbmc-old/github_597_oldfrontend_1 (Failed)
- regression/esbmc-old/github_597_oldfrontend_2 (Failed)
- regression/esbmc-old/github_839 (Failed)

Full list: [ctest.summary.txt](https://github.com/esbmc/esbmc/files/12748475/ctest.summary.txt)

*Edit:* I should say that this is after removing the `--memlimit` from the C++ tests via #1382.

fbrausse:
More statistics:
- 28x LeakSanitizer: detected memory leaks
- 531x AddressSanitizer: stack-use-after-scope
- 15x AddressSanitizer: heap-buffer-overflow
- 4x AddressSanitizer: stack-buffer-overflow

[...]

PR #1409, opened 2023-10-20
Interval Analysis unit tests now deletes migration reference

rafaelsamenezes:
As described in https://github.com/esbmc/esbmc/issues/1383. Interval Analysis unit tests were causing issues on Address Sanitizer.


lucasccordeiro (approved):

fbrausse (approved):
LGTM. The object is allocated in https://github.com/esbmc/esbmc/blob/bce1df6202c75e67ab4fc9b23e5f414a555492ea/unit/testing-utils/goto_factory.cpp#L189 I suppose there is no way to de-allocate it around the same location? A comment what this delete is about might help future readers.

PR #1412, opened 2023-10-21
Empty loops are now converted to assume 0

### LINKED ISSUE #1411: Inductive step can't handle infinite goto loops
The pattern:

```c
int my_assume(int n)
{
_EXIT: goto _EXIT;
}

int main() {
        int c = __VERIFIER_nondet_int();
        if(!c) my_assume(c);

        for(int i = 0; i < 5; i++);

        __ESBMC_assert(0, "");
}
```

Inductive step says that this program is true (` ./src/esbmc/esbmc   --unwind 3  check.c --inductive-step`). This is an issue as it causes wrong true veredicts.

As a sidenote,  replacing the goto loop to a while 1 (or an assume 0) solves the issue.

rafaelsamenezes:
Empty goto loops were not added into the `function_loops`, this creates a whole chain of misinterpretation as the `k_induction` methods needs the loop for the convertion to work. 

These loops are also used historically in TACAS to describe an `assume(0)` in an "automata" way, so I just did the convetion back.

Fixes #1411

### DIFF
--- a/regression/esbmc/github_1411_false/main.c
+++ b/regression/esbmc/github_1411_false/main.c
@@ -1,0 +1,13 @@
+int my_assume(int n)
+{
+_EXIT: goto _EXIT;
+}
+
+int main() {
+        int c = __VERIFIER_nondet_int();
+        if(!c) my_assume(c);
+
+        for(int i = 0; i < 5; i++);
+
+        __ESBMC_assert(0, "");
+}
--- a/regression/esbmc/github_1411_false/test.desc
+++ b/regression/esbmc/github_1411_false/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--k-induction
+^VERIFICATION FAILED$
--- a/src/goto-programs/goto_loops.cpp
+++ b/src/goto-programs/goto_loops.cpp
@@ -1,3 +1,4 @@
+#include "irep2/irep2_utils.h"
 #include <goto-programs/goto_loops.h>
 #include <util/expr_util.h>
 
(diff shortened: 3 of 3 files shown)


lucasccordeiro (approved):

mikhailramalho:
By implementing it inside the funcion_loop does this only run when using
k-induction or inductive-step flags?

The rest of esbmc might benefit from this transformation as a separate
step, probably before any k-induction transformation

lucasccordeiro:
This PR (30s):

````
  correct:           14210
    correct true:     7821
    correct false:    6389
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9565

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/6598561797/
````

Master (30s):

````
  correct:           14192
    correct true:     7810
    correct false:    6382
  incorrect:            32
    incorrect true:      7
    incorrect false:    25
  unknown:            9581
  Score:             21378 (max: 38482)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/6589914881
````

rafaelsamenezes:
> By implementing it inside the funcion_loop does this only run when using
> k-induction or inductive-step flags?
> 

I will double check it. Anyway, I think that the goto loops could be a full preprocessing and Analysis and the k-induction should go through a dependency inversion (not inherit goto_loops).

mikhailramalho:
There are two new incorrect true. Are they related to this new code or are
they noise?

lucasccordeiro:
This PR solves two incorrect true.

mikhailramalho:
You're right. Care to share which ones?

I'm surprised that the infinite loop caused incorrect trues, I was
expecting it to cause unknown results.

lucasccordeiro:
I believe Rafael has added the case to the regression suite. If it is not the identical benchmark, I am sure he will be happy to share it with you.Sent from my iPhoneOn 22 Oct 2023, at 15:55, Mikhail R. Gadelha ***@***.***> wrote:﻿
You're right. Care to share which ones?



I'm surprised that the infinite loop caused incorrect trues, I was

expecting it to cause unknown results.

rafaelsamenezes:
@mikhailramalho this is the original reachability benchmark: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/list-ext2-properties/simple_search_value-2.c

mikhailramalho:
thanks @rafaelsamenezes, I remember this test case. I also remember spending hours trying to fix it and eventually giving up. Thanks for the fix.

PR #1413, opened 2023-10-22
Add models for frexp() and ldexp()

### LINKED ISSUE #1416: Wrong verdict for isfinite() on NAN and INFINITY
This program gives VERIFICATION FAILED:
```c
#include <math.h>
#include <assert.h>

int main()
{
	assert(!isfinite(NAN));
}
```
Same for INFINITY.

fbrausse:
This PR adds the libm models for float, double and long double versions of frexp(), ldexp(), scalbn() and scalbln(). I've used the [man-pages](https://www.kernel.org/doc/man-pages/) for reference.

*Edit:* I've also added a new intrinsic `__ESBMC_bitcast(tgt, src)` getting two pointers and producing an assignment of *tgt from the bitcast-to-tgt-type *src. This is to avoid the memcpy() loop.

The `float` and `double` versions are always part of the c2goto libs, however, for 32-bit x86 (and PowerPC) I can't provide models for `long double`, see the comment close to the top of frexp.c for details. Means, on those platforms the *l() versions will just be missing.

Also fixes ieee_floatt::is_finite() and is_normal(), a bug triggered by the simplifier.

I've moved the macro definitions for the Windows headers from libm to our overridden headers. This is done in order to have a consistent environment between ESBMC's models and user programs.


fbrausse:
Seems Codacy is broken. They can't even tell what their error is. Let's ignore that.

rafaelsamenezes:
> Seems Codacy is broken. They can't even tell what their error is. Let's ignore that.

Apparently it just crashed. The [details of that commit](https://app.codacy.com/gh/esbmc/esbmc/commit?cid=777456484&fileBranchId=38490510) says `Current Status: Not Analyzed `

fbrausse:
> Apparently it just crashed. The [details of that commit](https://app.codacy.com/gh/esbmc/esbmc/commit?cid=777456484&fileBranchId=38490510) says `Current Status: Not Analyzed `

Yeah, it was somewhere during their "preprocessing" phase. Never mind though, seems to be back up.

fbrausse:
Codacy now complains about the "unused" label __ESBMC_hide... Nothing we can do about that.

emanino (commented):
Hi Franz,

thanks for the PR. I have the following question:

1) in regression/floats-regression/double-frexp-ldexp/main.c - lines 13-17 - how do you deal with the case where x is +-infinity? in that case xe may contain anything, including large negative numbers. That would cause a "zero times infinity" issue in the ldexp function which might break your assertion at line 15.

2) the same question in 1) applies to regression/floats-regression/float-frexp-ldexp/main.c

fbrausse:
> in regression/floats-regression/double-frexp-ldexp/main.c - lines 13-17 - how do you deal with the case where x is +-infinity? in that case xe may contain anything, including large negative numbers. That would cause a "zero times infinity" issue in the ldexp function which might break your assertion at line 15.

My frexp(3) man-page says
> If x is positive infinity (negative infinity), positive infinity  (negative  infinity)  is  re‐
turned, and the value of *exp is unspecified.

Similarly for ldexp(3):
> If x is positive infinity (negative infinity), positive infinity  (negative  infinity)  is  re‐
turned.

[cppreference](https://en.cppreference.com/w/c/numeric/math/ldexp) says the same:
> If arg is ±∞, it is returned, unmodified

So in both functions I first test whether it is an infinity, in which case the `x` is returned unchanged and `*exp` is not touched. Did you intend for `frexp()` to store a nondet_int() in *exp in case of infinities?

emanino:
> Hi Franz,
> 
> thanks for the PR. I have the following question:
> 
>     1. in regression/floats-regression/double-frexp-ldexp/main.c - lines 13-17 - how do you deal with the case where x is +-infinity? in that case xe may contain anything, including large negative numbers. That would cause a "zero times infinity" issue in the ldexp function which might break your assertion at line 15.
> 
>     2. the same question in 1) applies to regression/floats-regression/float-frexp-ldexp/main.c

I just want to add this. Do you think it is worth supporting the scalbn() family of functions too? They are almost identical to ldexp(), except that they allow any FLT_RADIX (not necessarily only powers of two). For now, we could just have a wrapper that works only when FLT_RADIX==2.

emanino:
> > in regression/floats-regression/double-frexp-ldexp/main.c - lines 13-17 - how do you deal with the case where x is +-infinity? in that case xe may contain anything, including large negative numbers. That would cause a "zero times infinity" issue in the ldexp function which might break your assertion at line 15.
> 
> My frexp(3) man-page says
> 
> > If x is positive infinity (negative infinity), positive infinity  (negative  infinity)  is  re‐
> > turned, and the value of *exp is unspecified.
> 
> Similarly for ldexp(3):
> 
> > If x is positive infinity (negative infinity), positive infinity  (negative  infinity)  is  re‐
> > turned.
> 
> [cppreference](https://en.cppreference.com/w/c/numeric/math/ldexp) says the same:
> 
> > If arg is ±∞, it is returned, unmodified
> 
> So in both functions I first test whether it is an infinity, in which case the `x` is returned unchanged and `*exp` is not touched. Did you intend for `frexp()` to store a nondet_int() in *exp in case of infinities?

Many thanks for the clarification! I was not intending to store nondet_int(), but now that you mention it, it could be a good idea. It's up to you.

fbrausse:
> I just want to add this. Do you think it is worth supporting the scalbn() family of functions too? They are almost identical to ldexp(), except that they allow any FLT_RADIX (not necessarily only powers of two). For now, we could just have a wrapper that works only when FLT_RADIX==2.

I agree. They're there now, too.

emanino (approved):

mikhailramalho (approved):
lgtm!

I would only add some simple test cases using some random constants, I don't mind adding it to the current test case.

The goal is to check if the simplifier is doing the right thing when simplifying the new code

fbrausse:
@mikhailramalho Good call! `assert(!isfinite(INFINITY))` works only with `--no-simplify`. Investigating...

fbrausse:
Need to investigate the isfinite() issue further in #1416. I've added some tests for finite constants. Now just hoping that the Windows build will eventually work.

fbrausse:
Windows seems to be sorted out now. SV-COMP only has a few calls to `frexp` and `ldexp` in c/goblint-coreutils, but those tests are not enabled in the stable '23 version the CI uses.

emanino:
> Windows seems to be sorted out now. SV-COMP only has a few calls to `frexp` and `ldexp` in c/goblint-coreutils, but those tests are not enabled in the stable '23 version the CI uses.

Hi Franz,

I would like to understand what was the issue with Windows, and how you managed to overcome it. Can you share a few more details?

Many thanks in advance.

fbrausse:
> I would like to understand what was the issue with Windows, and how you managed to overcome it. Can you share a few more details?

Sure. I should say I don't have Windows. We don't provide an own <math.h> header with ESBMC, but rely on that of the system. For Windows it appearently comes from the now-default "UCRT" (Universal C RunTime). I've found this repo, which provides the headers for some versions: https://github.com/tpn/winsdk-10/tree/master/Include/10.0.16299.0/ucrt

Now the UCRT headers by default provide inline definitions of some math functions in <corecrt_math.h>, e.g. [ceilf](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.16299.0/ucrt/corecrt_math.h#L641) or [frexpf](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.16299.0/ucrt/corecrt_math.h#L695). When we define another version of those (as this PR does), we need to somehow disable the definition of these inline functions in UCRT.

That's what took me some time to figure out how to do. Basically one has to `#define _CRT_FUNCTIONS_REQUIRED 0` before the `<corecrt_math.h>` is included. But defining this generally led to problems with other parts of UCRT: [<errno.h>](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/ucrt/errno.h) and others appearently also use this macro to determine whether some things are defined. So what this PR is doing now is overriding the <corecrt_math.h> and selectively, only for that particular header, defining the macro. Afterwards the original value of the macro is restored in order to get definitions of `errno` and similar symbols.

fbrausse:
Maybe I should add that for Glibc/Linux this wasn't a problem, because their <math.h> is more sane than UCRT in the sense that it doesn't contain inline function definitions.

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/6598561797/) (from https://github.com/esbmc/esbmc/pull/1412#issuecomment-1773993317):
```
  correct:           14210
    correct true:     7821
    correct false:    6389
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9565
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/6626333590) (34f9643907517794f599d196acc7889524b8c80a):
```
  correct:           14184
    correct true:     7802
    correct false:    6382
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9591
  Score:             21426 (max: 38482)
```

lucasccordeiro (approved):

lucasccordeiro:

PR #1397, opened 2023-10-17
FreeBSD/CheriBSD compatibility

fbrausse:
This PR aims at successfully building and running ESBMC on CheriBSD-14 (which is based on FreeBSD-14). This includes compilation and running the regression test (except for the CHERI ones, for now).

As FreeBSD is not using GCC's libstdc++ by default, I have included a few C++ fixes as well.

I'll keep this PR a draft while sorting out the remaining errors.

PR #1417, opened 2023-10-24
<fenv.h> only use own definitions on FreeBSD

fbrausse:
PR #1397 added an own version of <fenv.h>. However, that resulted in new incorrect SV-COMP results: https://github.com/esbmc/esbmc/commit/f9a93907c084fd839bdd94b261100c98b8cabb57 

For SV-COMP we still need the system's definitions, because of their preprocessing hard-coding the constants for rounding modes.

So this PR only uses the own definitions on FreeBSD (because there the system's header defines inline functions which ESBMC can't override if they're present) and goes to the system's <fenv.h> otherwise. I've started an SV-COMP run in <https://github.com/esbmc/esbmc/actions/runs/6626419104>.

### DIFF
--- a/src/c2goto/headers/fenv.h
+++ b/src/c2goto/headers/fenv.h
@@ -1,4 +1,7 @@
 
+/* their <fenv.h> contains inline function definitions incompatible with ours */
+#ifdef __FreeBSD__
+
 /* partially taken from FreeBSD-14 */
 
 #pragma once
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

fbrausse:
[Master](https://github.com/esbmc/esbmc/actions/runs/6598561797/) (from https://github.com/esbmc/esbmc/pull/1412#issuecomment-1773993317):
```
  correct:           14210
    correct true:     7821
    correct false:    6389
  incorrect:            30
    incorrect true:      5
    incorrect false:    25
  unknown:            9565
```
[This PR](https://github.com/esbmc/esbmc/actions/runs/6626419104) (ec7897f6dfdb875baad38c32add015a66f0e31ff):
```
  correct:           14193
    correct true:     7812
    correct false:    6381
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:            9593
  Score:             21637 (max: 38482)
```

lucasccordeiro:
Wonderful ;-)

Issue #1415, opened 2023-10-23
Structs are not symbolically initialized correctly


rafaelsamenezes:
While investigating some of the current incorrect trues I reached at this [AWS benchmark](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_array_eq_harness-1.i). The relevant piece of code is the following:

```c
// This is pair <index, byte>
struct store_byte_from_buffer old_byte_from_lhs; 
// Store a random byte from the array lhs into the struct
save_byte_from_array((uint8_t *)lhs, lhs_len, &old_byte_from_lhs); 
```

Where `save_byte_from_array` has a bug that arrays of length 1 do not initialize the struct which will case a later assertion to fail. By manually initializing the struct members to nondet values (or calling `--initialize-nondet-variables`) we can detect the bug and give a correct veredict.

As a side-note: aren't sv-benchmarks supposed to be explicit about nondet?

lucasccordeiro:
> As a side-note: aren't sv-benchmarks supposed to be explicit about nondet?

I think so. You can submit a merge request to fix the benchmark.

PR #1418, opened 2023-10-24
Nondet initialization is no longer optional

rafaelsamenezes:
As shown by #1415, ESBMC does not handle the nondet initialization correctly in some cases. This PR makes that non initialized variables are always considered nondet.


lucasccordeiro (approved):

mikhailramalho:
Care to share in which cases ESBMC does not handle the nondet initialization correctly?

rafaelsamenezes:
> Care to share in which cases ESBMC does not handle the nondet initialization correctly?

Structs were the only ones that I noticed.

rafaelsamenezes:
Results of this PR:

```
  correct:           14164
    correct true:     7793
    correct false:    6371
  incorrect:            28
    incorrect true:      3
    incorrect false:    25
  unknown:            9613
  Score:             21461 (max: 38482)
```
~Not sure why/how this is generating the extra incorrect results and unknowns.~. The latest master is the result from a PR and not  the actual master (so this run is off). So the previous master run is:

```
  correct:           14192
    correct true:     7810
    correct false:    6382
  incorrect:            32
    incorrect true:      7
    incorrect false:    25
  unknown:            9581
  Score:             21378 (max: 38482)
```

I will start another run though.

rafaelsamenezes:
After latest rebase:

```
  correct:           14182
    correct true:     7809
    correct false:    6373
  incorrect:            18
    incorrect true:      3
    incorrect false:    15
  unknown:            9605
  Score:             21655 (max: 38482)
```

fbrausse:
I think these were the last results we had from #1417:
```
  correct:           14193
    correct true:     7812
    correct false:    6381
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:            9593
  Score:             21637 (max: 38482)
```

fbrausse (approved):

fbrausse:
Still should figure out the failing regression tests. Any idea whether they're making unreasonable assumptions?

rafaelsamenezes:
>  Any idea whether they're making unreasonable assumptions?

Some were knownbugs. Others are concurrency related, I am still not sure why this patch would affect them.

mikhailramalho (changes requested):
Please add a test case, you can reduce the aws one.

rafaelsamenezes:
> Please add a test case, you can reduce the aws one.

Reducing it will take ages, I will look at the diff on the sv-benchmarks to see if there is a better regression. Nevertheless, there is the original github_644_bug that was what motivated the original flag.

mikhailramalho:
> > Please add a test case, you can reduce the aws one.
> 
> Reducing it will take ages, I will look at the diff on the sv-benchmarks to see if there is a better regression. Nevertheless, there is the original github_644_bug that was what motivated the original flag.

What do you mean it will take ages? Can't you just run cvise/creduce with parallel jobs and wait for the results?

lucasccordeiro:
@rafaelsamenezes: shall we close this PR?

rafaelsamenezes:
> @rafaelsamenezes: shall we close this PR?

No, it depends on: https://github.com/esbmc/esbmc/pull/1432

This PR results:

```
  correct:           14206
    correct true:     7817
    correct false:    6389
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:            9580
  Score:             21655 (max: 38482)
```

Master
```
  correct:           14200
    correct true:     7816
    correct false:    6384
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:            9586
  Score:             21648 (max: 38482)
```

It solves an incorrect true and adds a new one (which is the same noise from #1432).

lucasccordeiro (approved):

rafaelsamenezes:
The following tests are failing:

- regression/esbmc-unix/github_538_weaver_true. ESBMC is now finding a wrong assertion fail, I am still checking whether this is a pthread bug though. 
- regression/esbmc/github_666-id (Failed). I am having trouble running the slicer-id on WSL

fbrausse:
Regarding regression/esbmc/github_666-id: Your additional assignment makes symex change the renumbering: try with `--no-slice-id c:no_slice.c@21@F@main@foo?1!0&0#3` (i.e., changing `#2` to `#3`).

*Edit:* This "regression" test is very brittle. It's just there to ensure that the `--no-slice-id` still works. In case we have no users of this feature, it might be worth removing it.

fbrausse:
> additional assignment 

I wonder whether we can avoid that. Appearently `int` and other scalar types have already been assigned nondet values before this PR. So now we're assigning them twice, is that correct?

fbrausse:
regression/esbmc-unix/github_538_weaver_true run with `--no-slice` reveals
```
State 126 file main.c line 17 column 3 function thread1 thread 1
  pthread_thread_running = { 0, 1, 0, 0, [... repeating 0] }
```
as last assignment to pthread_thread_running; the second thread has never "officially" started. So `pthread_join(t2, 0)` appears to be a no-op and causes `thread2()` never to be run. I think that `t2` never got assigned and therefore remains nondet (and in the counter-example is assigned the value of `t1`) because `pthread_create(&t2, 0, thread2, 0)` never did run during symex. Maybe comparing which symbols get nondet-init'ed in https://github.com/esbmc/esbmc/pull/1418/files#diff-53c4483e9463ae9f1c2165578ba3a4e1e3c662ffd067524678137c49dad168edR32 between this PR and master will help figure this out?

*Edit:* Sorry, thread 2 seems to have run, please ignore that reasoning above.

fbrausse:
The `--no-slice` counter-example is strange. It has this sequence:
```
State 93 file pthread_lib.c line 169 column 3 function pthread_trampoline thread 2
  pthread_thread_ended[2] = 1

State 94 file pthread_lib.c line 170 column 3 function pthread_trampoline thread 2
  num_threads_running = 2 (00000000 00000010)

State 95 file pthread_lib.c line 145 column 7 function pthread_exec_key_destructors thread 2
  i = 0 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 112 file main.c line 17 column 3 function thread1 thread 1
  pthread_thread_ended = { 0, 0, 0, 0, [... repeating 0] };
```
Note how `pthread_thread_ended[2]` is set to 1 in thread 2 and thread 1 *later* has the `pthread_thread_ended` array contain 0 at [2].

rafaelsamenezes:
> Appearently int and other scalar types have already been assigned nondet values before this PR.

Hm, I am not sure where. It might be during DECL in symex...?

lucasccordeiro (approved):

lucasccordeiro:

PR #1419, opened 2023-10-25
Python frontend: Function definitions/calls

brcfarias:
This PR includes:
- Support for function definitions and calls.
- Regression tests for this feature


src/python-frontend/python_converter.cpp
```diff
@@ -288,15 +324,29 @@ void python_converter::get_var_assign(
...
     lhs = symbol_expr(*symbol);
   }
 
-  code_assignt code_assign(lhs, rhs);
-  code_assign.location() = location_begin;
-  target_block.copy_to_operands(code_assign);
+  /* If the right-hand side (rhs) of the assignment is a function, such as: x : int = func()
+   * we need to adjust the left-hand side (lhs) of the function call to refer to the lhs of the current assignment.
+   */
+  if(rhs.is_code())
```
rafaelsamenezes: This if-else could be converted in an if-return. Also, it is not clear that code implies a function call.

lucasccordeiro (approved):
That's really good, @brcfarias.

Thanks for submitting this PR!

rafaelsamenezes (changes requested):

src/python-frontend/python_converter.cpp
```diff
@@ -345,26 +395,124 @@ void python_converter::get_conditional_stms(
...
+  // Symbol identification
+  current_func_name = function_node["name"].get<std::string>();
+  std::string id = "py:" + python_filename + "@F@" + current_func_name;
+  std::string module_name =
+    python_filename.substr(0, python_filename.find_last_of("."));
+
+  // Handle empty function arguments
+  if(function_node["args"]["args"].empty())
+  {
+    type.make_ellipsis();
```
fbrausse: make_ellipsis() allows a function of that type to be called with an arbitrary number of more arguments than the type specifies. In C it's used to represent e.g.
```c
void f(int, ...)
```
Now f can be called as `f(42, 3, "more", "parameters")`. I don't think this is allowed for Python functions taking no arguments.

brcfarias: Thanks for pointing that out.
Fixed.

src/python-frontend/python_converter.cpp

src/python-frontend/python_converter.cpp
```diff
@@ -345,26 +394,120 @@ void python_converter::get_conditional_stms(
...
+  locationt location = get_location_from_decl(function_node);
+
+  // Symbol identification
+  current_func_name = function_node["name"].get<std::string>();
+  std::string id = "py:" + python_filename + "@F@" + current_func_name;
+  std::string module_name =
+    python_filename.substr(0, python_filename.find_last_of("."));
+
+  // Handle empty function arguments
+  if(!function_node["args"]["args"].empty())
```
fbrausse: No need to test this anymore :)

fbrausse (approved):
LGTM.

PR #1421, opened 2023-10-25
RFC: Low-precision models for log() family, exp() and pow()

fbrausse:
Following the discussion in #818 I thought it best to make a PR with concrete code and see what people think about it.

I'll add some test cases, so this PR is not final, yet.

This is the code for the last case (2x sqrt, Taylor order 5) mentioned in <https://github.com/esbmc/esbmc/issues/818#issuecomment-1776923729>.

> Here are some local running times for successfully verifying
> 
> ```c
> 	double x = nondet_double();
> 	__ESBMC_assume(0 < x && isfinite(x));
> 	int xe;
> 	frexp(x, &xe);
> 	double l = log2(x);
> 	double d = fabs(l - xe);
> 	assert(d <= 1);
> ```
> * 4x sqrt, Taylor order 8: 745.220s
> * 3x sqrt, Taylor order 8: 566.800s
> * 3x sqrt, Taylor order 7: 336.039s
> * 3x sqrt, Taylor order 6: 281.983s
> * 2x sqrt, Taylor order 6: 202.370s
> * 2x sqrt, Taylor order 5: 98.741s
> 
> According to my quick calculation, the last one (2x sqrt, Taylor order 5) should still provide correct results up to relative error 10^-5. That's far from the <1ulp math libraries provide, but I found the verification times to be interesting.
> [...]
> I've also tried verifying
> 
> ```c
> 	double x = nondet_double();
> 	__ESBMC_assume(1 < x && x < 2);
> 	double l = log(x);
> 	double e = exp(l);
> 	double d = fabs(e - x);
> 	assert(d < 2e-2);
> ```
> with the last configuration and it took 5362.173s (~1.5h).


fbrausse:
In my opinition this PR is ready. It includes a version of pow() that now finds bugs in regression tests (esbmc/github_1352*) due to a special case for integral exponents (note: this is independent of the additional special-casing for base 2, I just included it to speed up the regression tests). The numbers involving exp() in https://github.com/esbmc/esbmc/issues/818#issuecomment-1780855080 do not apply anymore: I reduced the order of the Taylor approximation from 8 to 6.

Codacy reports some warnings which I don't think are useful here: 
image

lucasccordeiro (approved):

mikhailramalho (approved):
lgtm. I'm fine with balancing precision and verification time, I was even going to mention having an option to increase precision, but I'm not sure if users will use it.

My only request is to add a brief note in our readme explaining our design choice, the trade-offs, what's the expected error (I think I saw you mentioned 10^-5 somewhere), etc.

fbrausse:
> I'm fine with balancing precision and verification time, I was even going to mention having an option to increase precision, but I'm not sure if users will use it.

@mikhailramalho Thanks, I'll add a note to the README. I've been discussing with Rafael whether it would be a good idea to have "high-precision" and "fast" models for the math library. Your comment goes into the same direction. I think we'll need to make our OMs a bit more flexible for that.

src/c2goto/library/libm/exp.c
```diff
@@ -0,0 +1,135 @@
...
+{
+  /* Compute truncated Taylor series for e^x - 1 around 0:
+   * x + x^2/2! + x^3/3! + x^4/4! + ... + x^n/n! */
+  double acc = x;
+  double smd = x;
+  acc += (smd *= x / 2);
+  acc += (smd *= x / 3);
+  acc += (smd *= x / 4);
+  acc += (smd *= x / 5);
+  acc += (smd *= x / 6);
```
rafaelsamenezes, line 14: The `--goto-unwind` should be smart enough if you convert this into a for loop with a constant bound.

rafaelsamenezes (approved):
The OMs look good. However, the repetitions seems avoidable. I guess we could start using the `goto-unwind` during the `c2goto`.

src/c2goto/library/libm/exp.c
lucasccordeiro: But this would require the users always to invoke the `--goto-unwind` :-)

fbrausse: > But this would require the users always to invoke the `--goto-unwind` :-)

I actually thought about using `--goto-unwind` but it doesn't work when `--unwind` is specified as well, as is done for the github_1352* test cases.

rafaelsamenezes: > But this would require the users always to invoke the --goto-unwind :-)

If we set it to default at `c2goto` that would not be a requirement.

lucasccordeiro:
> If we set it to default at c2goto that would not be a requirement.

I see. Nice!

src/c2goto/library/libm/exp.c
mikhailramalho: does --goto-unwind unwind all loops or is it only the ones with known loop bounds?

rafaelsamenezes: > only the ones with known loop bounds

Only loops of the like  `for(int i =0; i< N; i++)` where `i` is read-only at the loop body, and `N` is a constant (it needs to be a value).

fbrausse: Even if it works with c2goto, there are 2 reasons I wouldn't recommend doing that (here):
1. c2goto is optional, it would be non-intuitive to have the precompiled libs behave different than the non-precompiled ones
2. the *_taylor() functions are coded suboptimally wrt. precision: it's in almost all cases better to use some kind of Horner scheme to evaluate polynomials; I haven't done that in order to be able to easily tweak the order of the polynomials as I was aiming for good verification times

The second point does not apply to pow_by_squaring() and there an unrolled loop would be nice. But the first point still bothers me.

fbrausse:
The CI testing job keeps getting killed, likely due to memory usage (CTest terminates with signal 143). Seems Bitwuzla is using more memory than Boolector as it started happening with this change:
```diff
diff --git a/regression/bitwuzla/github_1352-32bit/test.desc b/regression/bitwuzla/github_1352-32bit/test.desc
index a95b795e8b..2affeeffd5 100644
--- a/regression/bitwuzla/github_1352-32bit/test.desc
+++ b/regression/bitwuzla/github_1352-32bit/test.desc
@@ -1,5 +1,5 @@
 THOROUGH
 gh-1352.c
---unwind 1 --no-unwinding-assertion --force-malloc-success --32
+--unwind 1 --no-unwinding-assertion --force-malloc-success --32 --bitwuzla
 ^VERIFICATION FAILED$
 \bdereference failure: array bounds violated\b
diff --git a/regression/bitwuzla/github_1352/test.desc b/regression/bitwuzla/github_1352/test.desc
index 22eefc398b..ec7bcb8b32 100644
--- a/regression/bitwuzla/github_1352/test.desc
+++ b/regression/bitwuzla/github_1352/test.desc
@@ -1,5 +1,5 @@
 THOROUGH
 gh-1352.c
---unwind 1 --no-unwinding-assertion --force-malloc-success
+--unwind 1 --no-unwinding-assertion --force-malloc-success --bitwuzla
 ^VERIFICATION FAILED$
 \bdereference failure: array bounds violated\b
```
Before that Boolector would time out on these two regression tests. Locally, they use ~5.6G memory with Bitwuzla. They also can't be run with Z3 (not just on Windows) due to memory (more than 40G locally). Yices "only" uses ~3.2G on them but takes a bit longer and is currently not built into the CI via the `build.sh` script. :(

fbrausse:
I've simplified the test case (problem was: calling `pow(2,n)` repeatedly without changing `n`) and now Boolector solves it with 600M memory, while Bitwuzla needs 1.7G. So I've moved the test cases back to Boolector and kept the original ones in regression/disabled. I've also tried @rafaelsamenezes's GCSE but (understandably) it didn't help with reducing the multiple equal calls.

lucasccordeiro:
Many thanks, @fbrausse.

I will merge your PR as soon as CI is green.

fbrausse:
Thanks @lucasccordeiro. I'll update the README with the error bound info once I finished calculating it.

PR #1422, opened 2023-10-25
Added a default table style for benchexec

rafaelsamenezes:
This PR adds support for using custom columns when using table-generator. This will enable us to start looking for patterns and points to improve. With the current table the output will contain:

- The bound where the violation/proof was found
- The time that it took to generate the GOTO program
- The time that it took to preprocess the GOTO program
- The total (sum) time of Symex
- The total (sum) time of Solving

Example:

Screenshot 2023-10-25 170248

It depends on https://github.com/sosy-lab/benchexec/pull/929 being merged though.

### DIFF
--- a/scripts/competitions/svcomp/table.xml
+++ b/scripts/competitions/svcomp/table.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" ?>
+<!DOCTYPE table PUBLIC "+//IDN sosy-lab.org//DTD BenchExec table 1.10//EN" "https://www.sosy-lab.org/benchexec/table-1.10.dtd">
+<table>
+
+  <column title="status"/>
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):
Nice ;-)

fbrausse:
Do you have an example how the diff-tables look like? What's the command line to use this format with table-generator?

rafaelsamenezes:
> Do you have an example how the diff-tables look like? 

It will be similar, the status will be the main factor

diff

> What's the command line to use this format with table-generator?

` table-generator -x table.xml <file.xml.bz2>`

fbrausse (approved):
Thanks, that's great @rafaelsamenezes.

Issue #1426, opened 2023-10-27
ESBMC crashes on symbolic unions


rafaelsamenezes:
By changing the regression `esbmc/github_732-2-2` to have nondet functions:

```c
typedef union {
  int x[12000];
} t1;

typedef struct {
  int x;
  t1 y;
} t2;

t2 nondet_t2();

int main()
{
  t2 a = nondet_t2();
  int *b = &a.x;
  b[1] = a.x;
  assert(*b);
}
```

Causes esbmc to fail with: `esbmc/src/util/c_expr2string.cpp:1200: virtual std::string c_expr2stringt::convert_union(const exprt&, unsigned int&): Assertion !init.empty()' failed. Aborted`

lucasccordeiro:
@fbrausse: Can I ask you whether you could check this issue?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/1427.

PR #1427, opened 2023-10-27
Unions are not required to have an init member anymore

rafaelsamenezes:
As shown in #1426 ESBMC would crash in a symbolic union. The expression itself, was constructed by reading a field of the union. Which means that the operator was known but there are no garantees on the init member.


lucasccordeiro (approved):

fbrausse (changes requested):
The `init.as_string()` will be empty in the code below your change resulting in `.=` which is probably not what we want to print.

fbrausse (approved):
Sorry, I misread your change. You're guarding against that. LGTM.

src/util/c_expr2string.cpp
```diff
@@ -1194,10 +1194,8 @@ c_expr2stringt::convert_union(const exprt &src, unsigned &precedence)
   const exprt::operandst &operands = src.operands();
   const irep_idt &init [[maybe_unused]] = src.component_name();
```
fbrausse: We can remove the `maybe_unused` now.

PR #1428, opened 2023-10-28
[goto-contractor] ibex contractor for interval analysis 

Mohannad-Aldughaim:
- Added ibex contractor option to be used in interval analysis:
Using --interval-analysis-ibex-contractor will enable this option in interval analysis. this option will apply the contractor on each goto instruction in the goto program and feed the contracted intervals back to interval analysis. 

- Added preliminary version of ibex contractor to be used post interval analysis.
Can be used with --goto-contractor-algo2. this option will apply goto contractor post interval analysis and insert the contracted intervals beck into the goto program as assumes. 

- Changed the structure of classes to include an ibex parser.

To follow:
- Regression and documentation.
- fix or disable for goto-contractor method.

Regards,
-Mohannad


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1682,10 +1682,10 @@ bool esbmc_parseoptionst::process_goto_program(
       goto_k_induction(goto_functions);
     }
 
-    if(cmdline.isset("goto-contractor"))
+    if(cmdline.isset("goto-contractor") || cmdline.isset("goto-contractor-algo2"))
```
lucasccordeiro: We should find a better name for "goto-contractor-algo2". What is "algo2"?

src/esbmc/options.cpp
```diff
@@ -338,7 +338,10 @@ const struct group_opt_templ all_cmd_options[] = {
     {"k-induction", NULL, "prove by k-induction "},
     {"goto-contractor",
      NULL,
-     "enable contractor-based interval refinements on goto level"},
+     "enable contractor-based interval refinements on goto level on asserts"},
+    {"goto-contractor-algo2",
```
lucasccordeiro: what about "goto-contractor-condition"?

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -1120,6 +1121,21 @@ void interval_domaint::assume(const expr2tc &cond)
   }
 #endif
 
+  if(enable_ibex_contractor)
+  {
+    interval_analysis_ibex_contractor contractor;
+    if(contractor.parse_guard(new_cond))
+    {
+      contractor.maps_to_domains(int_map, real_map);
+      //contractor.modularize_intervals();
```
lucasccordeiro: Remove this line.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -1120,6 +1121,21 @@ void interval_domaint::assume(const expr2tc &cond)
   }
 #endif
 
+  if(enable_ibex_contractor)
```
lucasccordeiro, line 1127: Add some explanation.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -1219,6 +1235,13 @@ void interval_domaint::assume_rec(const expr2tc &cond, bool negation)
     else if(is_floatbv_type(cond) && enable_real_intervals)
       apply_assume_symbol_truth<real_intervalt>(to_symbol2t(cond), negation);
   }
+  else if(is_constant_bool2t(cond))
```
lucasccordeiro, line 1240: Why do we need this? Add some explanation.

src/goto-programs/goto_contractor.cpp
```diff
@@ -13,27 +18,74 @@ void goto_contractort::get_contractors(goto_functionst goto_functions)
...
-    if(ins.is_assume())
-      parse_intervals(ins.guard);
+  interval_analysis(goto_functions, ns);
+
+  std::ostringstream oss;
+  interval_analysis.output(goto_functions, oss);
+
+  std::ostringstream oss2;
+
+  //Forall_goto_functions(f_it, goto_functions)
```
lucasccordeiro: Remove this line.

src/goto-programs/goto_contractor.cpp
```diff
@@ -13,27 +18,74 @@ void goto_contractort::get_contractors(goto_functionst goto_functions)
...
+        {
+          auto var_name = to_symbol2t(it->second.getSymbol()).get_symbol_name();
+          auto new_interval = interval_analysis[i_it].get_int_map()[var_name];
+          if(!new_interval.is_top())
+          {
+            auto lb = new_interval.get_lower().to_int64();
+            auto ub = new_interval.get_upper().to_int64();
+            auto i = it->second.getInterval();
+
+            std::cout << "at location: " << i_it->location_number
```
lucasccordeiro: We need to use our methods to report the log.

src/goto-programs/goto_contractor.cpp
```diff
@@ -13,27 +18,74 @@ void goto_contractort::get_contractors(goto_functionst goto_functions)
...
+  auto f_it = goto_functions.function_map.find("c:@F@main");
+  {
+    Forall_goto_program_instructions(i_it, f_it->second.body)
+    {
+      if(
+        i_it->is_assert() ||
+        (i_it->is_function_call() &&
+         is_symbol2t(to_code_function_call2t(i_it->code).function) &&
+         to_symbol2t(to_code_function_call2t(i_it->code).function)
+             .get_symbol_name() == "c:@F@__VERIFIER_assert"))
```
lucasccordeiro, line 54: What about other assert statements like __ESBMC_assert?

src/goto-programs/goto_contractor.cpp
```diff
@@ -13,27 +18,74 @@ void goto_contractort::get_contractors(goto_functionst goto_functions)
...
+              map.update_lb_interval(lb, var_name);
+            if(new_interval.upper && (isinf(i.ub()) || i.ub() < ub))
+              map.update_ub_interval(ub, var_name);
+          }
+          it++;
+        }
+      }
+    }
+  }
+  // instrument_intervals(interval_analysis, f_it->second);
```
lucasccordeiro: Remove this line.

src/goto-programs/goto_contractor.cpp
```diff
@@ -167,12 +219,272 @@ void goto_contractort::apply_contractor()
   log_status("{}", oss.str());
 }
 
-ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
+bool goto_contractort::initialize_main_function_loops()
+{
+  auto it = goto_functions.function_map.find("c:@F@main");
```
lucasccordeiro, line 218: What about if we model check C++ programs? We should use something like has_prefix.

src/goto-programs/goto_contractor.cpp
```diff
@@ -167,12 +219,272 @@ void goto_contractort::apply_contractor()
...
+void goto_contractort::insert_assume_at(
+  goto_functiont goto_function,
+  std::_List_iterator<goto_programt::instructiont> instruction)
+{
+  expr2tc cond = gen_true_expr();
+  for(auto const &var : map.var_map)
+  {
+    //testing with updating both bounds after reduction.
+    auto X = var.second.getSymbol();
+    //if(var.second.isIntervalChanged())
```
lucasccordeiro: Remove this line.

src/goto-programs/goto_contractor.cpp
```diff
@@ -167,12 +219,272 @@ void goto_contractort::apply_contractor()
...
+  goto_functions.update();
+  return true;
+}
+
+void goto_contractort::insert_assume_at(
+  goto_functiont goto_function,
+  std::_List_iterator<goto_programt::instructiont> instruction)
+{
+  expr2tc cond = gen_true_expr();
+  for(auto const &var : map.var_map)
```
lucasccordeiro, line 240: We should add some explanations.

src/goto-programs/goto_contractor.h
```diff
@@ -432,39 +510,112 @@ class goto_contractort : public goto_functions_algorithm
...
+  }
+  ~interval_analysis_ibex_contractor()
+  {
+    delete(vars);
+  }
+
+  bool parse_guard(expr2tc &guard)
+  {
+    auto t_0 = std::chrono::steady_clock::now();
+    //ibex::Ctc *c = create_contractor_from_expr2t(guard);
```
lucasccordeiro: Remove this line.

lucasccordeiro (changes requested):
Could you please add a few regression tests here?

src/goto-programs/goto_contractor.cpp
fbrausse: You can use `log_debug("contractor", "at location: {} Var {} this out {} should be: {}", ...)`.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Not sure i understand this. May i ask for an example?

src/esbmc/esbmc_parseoptions.cpp
Mohannad-Aldughaim: changed to goto-contractor-condition.

src/esbmc/options.cpp
Mohannad-Aldughaim: changed to goto-contractor-condition.

src/goto-programs/abstract-interpretation/interval_domain.cpp
Mohannad-Aldughaim: done

src/goto-programs/abstract-interpretation/interval_domain.cpp
Mohannad-Aldughaim: Done. 
Added "/// use ibex contractors to reduce the intervals for interval analysis"

src/goto-programs/abstract-interpretation/interval_domain.cpp
Mohannad-Aldughaim: Done.
Added "// //added in case "cond = false" which happens when the ibex contractor results in empty set. "

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: It seems that __ESBMC_assert is considered an assert (unlike __VERIFIER_assert which is considered a function call) which is covered by this condition.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Removed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Done.
Added:
/// Here we build an assume instruction with a conjunction of multiple conditions.
/// We start with a true expression and add other conditions with and2tc
/// eventually, we will have somthing like: true and x>0 and x<10

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: Removed.

src/goto-programs/goto_contractor.cpp
```diff
@@ -228,39 +603,44 @@ ibex::Ctc *goto_contractort::create_contractor_from_expr2t(const expr2tc &expr)
...
+  g = create_function_from_expr2t(rel->side_2);
   if(f == nullptr || g == nullptr)
     return nullptr;
-  //TODO: implement "not" flag  and process.
-  switch(get_base_object(expr)->expr_id)
+
+  switch(base_object->expr_id)
   {
   case expr2t::expr_ids::greaterthanequal_id:
     c = new ibex::NumConstraint(*vars, (*f)(*vars) >= (*g)(*vars));
```
fbrausse, line 676: Are these dynamic memory allocations ever being cleaned up?

src/goto-programs/goto_contractor.cpp
```diff
@@ -339,21 +772,26 @@ ibex::Function *goto_contractort::create_function_from_expr2t(expr2tc expr)
...
     break;
   }
   case expr2t::expr_ids::symbol_id:
   {
     int index = create_variable_from_expr2t(expr);
-    if(index != CspMap::NOT_FOUND)
+    if(index == CspMap::NOT_FOUND)
     {
-      f = new ibex::Function(*vars, (*vars)[index]);
+      log_error("ERROR: MAX VAR SIZE REACHED");
```
fbrausse: log_error() already prints "ERROR", no need to prefix.

src/goto-programs/goto_contractor.h
```diff
@@ -236,22 +250,21 @@ class Contractors
 class CspMap
 {
 public:
-  static constexpr int MAX_VAR = 26;
+  static constexpr int MAX_VAR = 20;
```
fbrausse, line 319: I still don't understand this limit, why we have it or why it would need to be lowered. Could you please explain it?

src/goto-programs/goto_contractor.h
```diff
@@ -337,34 +401,47 @@ class goto_contractort : public goto_functions_algorithm
    * Fourth, inserting assumes in the program to reflect the contracted intervals.
    * @param _goto_functions
    */
-  goto_contractort(goto_functionst &_goto_functions)
+  goto_contractort(
+    goto_functionst &_goto_functions,
+    const namespacet &ns,
+    optionst options)
```
fbrausse: Please use a const-reference for passing `options` in order to not copy it every time.

src/goto-programs/goto_contractor.h
```diff
@@ -337,34 +401,47 @@ class goto_contractort : public goto_functions_algorithm
...
-        return;
+        vars = new ibex::Variable(CspMap::MAX_VAR);
+        parser = expr_to_ibex_parser(&map, vars);
+        log_debug(
+          "contractor", "1/4 - Parsing asserts to create CSP Constraints.");
+        get_contractors(_goto_functions);
+        if(contractors.is_empty())
+        {
+          log_status(
+            "contractor",
```
fbrausse, line 511: The additional first "module" string is only needed for log_debug(). This call here will just print "contractor", which is unlikely to be intended.

src/goto-programs/abstract-interpretation/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_library(abstract-interpretation ai.cpp ai_domain.cpp interval_domain.cpp interval_analysis.cpp)
 target_link_libraries(abstract-interpretation fmt::fmt)
 target_include_directories(abstract-interpretation
-        PUBLIC ${Boost_INCLUDE_DIRS}
+        PUBLIC ${Boost_INCLUDE_DIRS} ${IBEX_INCLUDE_DIRS} ${IBEX_INCLUDE_DIRS}/ibex ${IBEX_INCLUDE_DIRS}/ibex/3rd
```
fbrausse: These should only be added if IBEX was found & enabled during the build.

fbrausse (commented):
There are a lot of changes here, and I've not fully understood this PR entirely. Below a few comments and questions. I'll take another look once I understood more.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: This represents the variables for a given constraint. I reached this number because it is the maximum number of variables in a single condition in SV-Comp. However, I want to explore the possibility of having it dynamic if you have a suggestion.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: My understanding was that ibex used smart pointers. This doesn't seem to be the case. I will fix this one.

Mohannad-Aldughaim: fixed. I created lists that keep track of each created object and delete it in the deconstructor.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: fixed.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: That's actually meant to be log_debug. fixed.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: fixed.

src/goto-programs/abstract-interpretation/CMakeLists.txt
Mohannad-Aldughaim: fixed.

src/goto-programs/goto_contractor.cpp
fbrausse: Sorry, but this memory "management" code has to be fixed properly:
- You have already been dynamically allocating (with `new`) all these ibex objects, which may or may not be smart pointers; please check! If they indeed are, do not dynamically allocate them.
- Now you're remembering which ones to delete in a **linked list**. This is just not great at all, because each element of a linked list is dynamically allocated **again**. So now you're performing 2 dynamic allocations where I believe none are necessary.

Why do I believe that none are necessary? Because their API, as used by the contractors code, takes ibex (sub-)objects by reference, not by (raw) pointer. That's why you're dereferencing the dynamically allocated ibex objects all the time you're using their API. Please correct me if I'm wrong, but this doesn't seem to be right usage.

I realize that it's been there as it is, and this PR is already big enough, so we can postpone this to another PR, but it should be the next step for the contractors code.

src/goto-programs/goto_contractor.h
fbrausse: > fixed.

Not really. You've added `const` but not a reference, e.g. `const optionst &options`. The difference is that with a reference no copy of the `options` object is being made. Currently this unnecessary copy is being performed.

src/goto-programs/goto_contractor.cpp
fbrausse: I've had a quick look at the ibex implementation. Basically all of its types assume pointer stability for contained objects. That's certainly not an idiomatic choice, but it explains why allocations of the form the contractors code is doing are indeed necessary.

They even have their own type for tracking allocations similar to your change: https://github.com/ibex-team/ibex-lib/blob/master/src/tools/ibex_Memory.h It also uses linked list, and in addition even an dynamically allocated sub-object to store in there. Which is worse. But it has a useful comment in `~Memory()`:
> deleted in reverse order (in case some objects depend on previous objects)

I presume the only way to use ibex' strange API efficiently in a dynamic setting is to write an own arena-style memory allocator ontop of a [std::deque](https://en.cppreference.com/w/cpp/container/deque) or similar. @ all: Am I missing something here?

Anyway, so I agree we need to track the allocated objects. My suggestion as long as we don't have this arena-style memory allocator would be to append to a std::vector instead of prepending to std::list to track the pointers. For deletion the std::vector needs to be traversed in reverse insertion order.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: I see. I will fix it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: Thank you, Franz. Your comments gave me great insight. I have changed the lists to vectors and traversed them in reverse. May I ask you why vectors instead of lists? is it to avoid more dynamic allocations?

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: Should be ok now.

src/goto-programs/goto_contractor.cpp
fbrausse: > May I ask you why vectors instead of lists? is it to avoid more dynamic allocations?

Happy to help. Yes, it is to avoid dynamic allocations. A recent profiling of ESBMC shows that we already spend a lot of time in malloc and grouping allocations together is a good way to keep the pressure down. Especially when we can't exactly know how complex the expressions are going to be. A std::vector only ever keeps one array, regardless how many elements it holds. The array is reallocated roughly for every log(n) added element.

> I have changed the lists to vectors and traversed them in reverse.

The `~Contractor` destructor still seems to go through in forward direction, though. I wonder why both destructors delete the allocated ibex elements...

src/goto-programs/goto_contractor.h
fbrausse: Great, thank you.

fbrausse:
@Mohannad-Aldughaim We recently changed Github's branch protection rule namings, that's why this PR cannot be merged. If you could merge master (or rebase) and resolve the conflict in src/goto-programs/goto_contractor.cpp shown here, then we can go ahead. I'll take a fresh look at it later. All in all it looks quite nice.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: > The ~Contractor destructor still seems to go through in forward direction, though. I wonder why both destructors delete the allocated ibex elements...

Fixed. Sorry, I missed that one.

src/goto-programs/goto_contractor.h
fbrausse: Could you please add your explanation as a comment, together with a TODO saying something like "this should be dynamic"?

lucasccordeiro:
@Mohannad-Aldughaim and @rafaelsamenezes: could we also enable the regression tests for the goto-contractors? otherwise, I'm afraid that it will break again in the future.

src/goto-programs/goto_contractor.cpp
lucasccordeiro: What happens with C++ programs? Do we still use c:@F@main?

lucasccordeiro (changes requested):

src/goto-programs/goto_contractor.cpp
fbrausse: There is `config.main` holding the value of the `--function` parameter, which is the generic way determine the entry point of the program. However, there are a few users of `c:@F@main`, hence this might be a broader issue: 
image

lucasccordeiro (approved):

lucasccordeiro:

I have spoken with @rafaelsamenezes for you to enable the regression test in our CI. However, this work should be part of another PR.

fbrausse:
@rafaelsamenezes @lucasccordeiro should we include this in the svcomp24 branch?

lucasccordeiro:
Yes, definitely 👍

PR #1429, opened 2023-10-29
Support shared builds against different Clang versions on Ubuntu

fbrausse:
This PR generalizes the build.sh script to allow for shared builds (new parameter `-S OFF`) and to select which Ubuntu-22.04-packaged Clang version to build the frontend against (new parameter `-c MAJOR` where I've tested builds with MAJOR being one of 11, 12, 13, 14, 15). Shared builds now also link against the system's Z3.

Reasons for this change: We're slowly moving away from Clang-11 as the default. Binary builds of newer versions are not distributed by upstream anymore. So we should look into building against one provided by the distribution. Additionally, it is worth noting that the C++ ABI is not stable across compiler versions, that means pre-built binaries are not guaranteed to link correctly.

@rafaelsamenezes Do you think it would be worth to extend the CI to additionally run a shared build against one of the newer Clangs?

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   build:
-    name: Build ESBMC (Windows)
+    name: Build ESBMC (${{ inputs.operating-system }})
     runs-on: ${{ inputs.operating-system }}
     steps:
     - name: Set up Python
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -1,0 +1,50 @@
+
+on:
+  workflow_call:
+    inputs:
+      operating-system:
+        description: "Targeted OS (e.g. ubuntu-latest, windows-latest)"
+        required: true
+        type: string
+      build-flags:
+        description: "Flags to be passed to build.sh script"
+        required: true
+        type: string
+      testing:
+        description: "Whether to run tests (regression) for this build"
+        required: true
+        type: boolean
+
+jobs:
+  build:
+    name: Build ESBMC (Windows)
+    runs-on: ${{ inputs.operating-system }}
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: check python
+      run: python --version
+    - name: Update vcpkg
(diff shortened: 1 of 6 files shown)


rafaelsamenezes:
> Do you think it would be worth to extend the CI to additionally run a shared build against one of the newer Clangs

I think it should be fine, as long as we only do up to "core" testing.

fbrausse:
Added a dynamically linked build against Ubuntu's Clang-13 to the CI; it works fine without changes to the regression testing modes.

fbrausse:
I've reduced the visual clutter in the Github interface showing the checks run for a PR a bit.

To reduce clutter further, it would be nice if we could rename "build (ubuntu-latest)" and "build (windows-latest)" into something involving neither "build" nor the OS label, since this information is duplicated and makes it harder to read. Additionally, "Build ESBMC (Windows)" should most likely be "Build ESBMC (windows-latest)". Who can access Github's branch protection rules?

fbrausse:
If this PR gets approval, we'll need to adjust the master branch's protection rules to the new names before merging it.

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,50 @@
+
+on:
+  workflow_call:
+    inputs:
+      operating-system:
```
rafaelsamenezes, line 5: I think we can remove this from the Windows build

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,50 @@
...
+        description: "Targeted OS (e.g. ubuntu-latest, windows-latest)"
+        required: true
+        type: string
+      build-flags:
```
rafaelsamenezes, line 9: Same as above

rafaelsamenezes (approved):
LGTM!

.github/workflows/build-windows.yml
fbrausse: I kept it because Github supports multiple versions of each OS. For Windows they have `windows-2022` and `windows-2019`: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners

lucasccordeiro (approved):

PR #876, opened 2022-12-07
Several random fixes

mikhailramalho:
This is a big PR that includes several fixed I had scattered in several branches.

The biggest one is the rewrite of polymorphic gcc functions handling: now we create model on the fly during adjust for atomic/sync calls (note that I didn't implement all the models only the ones we "supported" before, I left it as an exercise to someone else). This allowed me to remove the old sync_fetch_add implementation (which only handled int), and the weird atomic_load/atomic_store macros.

Other small changes include:
* Renaming of a few intrinsics that were missing the __ESBMC_ prefix
* Fix printing "error: error: " in cprover_library
* Fix warning in pthread_lib.c and the clang c++ frontend
* Added support for imaxabs (from Rafael)
* Added some C++ template support for name generation
* Removed a bunch of compilers defines that we were setting.
* cleanup our predefined includes
* Fix warning messages to include "WARNING: "
* Make --quiet actually mean quiet, i.e., esbmc prints no output whatsoever
* Moved support of builtins to symex, including the implementation of all the builtin arith checks that we used to support.
* Fixed the "not unwinding loop" in symex that didn't say which loop it wasn't unwinding anymore (this one annoyed me for eons).
* Support overflow exprt in cex

PR #1433, opened 2023-10-30
[NFC] Fix warnings and enable Werror on github actions

mikhailramalho:
This PR is part of a series of PRs derived from #876: although that PR was accepted, it's been a nightmare to rebase, so I decided to split it.

This first PR fixes all warnings on esbmc (except for one in smtlib) and enables werror in the build.sh script (for linux only, as I don't have macos to test this PR).

### DIFF
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -1401,6 +1401,7 @@ void clang_cpp_convertert::get_base_map(
       continue;
 
     auto status = map.insert({class_id, base_cxxrd});
+    (void)status;
     assert(status.second);
   }
 }
--- a/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
@@ -621,6 +621,7 @@ void clang_cpp_convertert::get_overriden_methods(
       continue;
 
     auto status = map.insert({method_id, *md_overriden});
+    (void)status;
     assert(status.second);
   }
 }
--- a/regression/esbmc-solidity/constructor_3/test.desc
+++ b/regression/esbmc-solidity/constructor_3/test.desc
@@ -1,4 +1,4 @@
-CORE
+THOROUGH
 contract.solast
 --function Derive --sol contract.sol
 ^VERIFICATION FAILED$
--- a/regression/esbmc-solidity/constructor_3/test.desc
+++ b/regression/esbmc-solidity/constructor_3/test.desc
@@ -1,4 +1,4 @@
-THOROUGH
+KNOWNBUG
 contract.solast
 --function Derive --sol contract.sol
 ^VERIFICATION FAILED$
--- a/regression/esbmc-solidity/constructor_6/test.desc
+++ b/regression/esbmc-solidity/constructor_6/test.desc
@@ -1,4 +1,4 @@
-CORE
+THOROUGH
 contract.solast
 --function DD --sol contract.sol
 ^VERIFICATION FAILED$
--- a/regression/esbmc-solidity/constructor_6/test.desc
+++ b/regression/esbmc-solidity/constructor_6/test.desc
@@ -1,4 +1,4 @@
-THOROUGH
+KNOWNBUG
 contract.solast
 --function DD --sol contract.sol
 ^VERIFICATION FAILED$
--- a/regression/esbmc-solidity/multiple_contracts_2/test.desc
+++ b/regression/esbmc-solidity/multiple_contracts_2/test.desc
@@ -1,4 +1,4 @@
-CORE
+THOROUGH
 mul.solast
 --function empty --sol mul.sol
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-solidity/multiple_contracts_2/test.desc
+++ b/regression/esbmc-solidity/multiple_contracts_2/test.desc
@@ -1,4 +1,4 @@
-THOROUGH
+KNOWNBUG
 mul.solast
 --function empty --sol mul.sol
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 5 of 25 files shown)


lucasccordeiro (approved):

scripts/build.sh
```diff
@@ -22,7 +22,7 @@ ubuntu_setup () {
     echo "Installing Python dependencies" &&
     pip3 install --user meson ast2json &&
     meson --version &&
-    BASE_ARGS="$BASE_ARGS -DENABLE_OLD_FRONTEND=On -DENABLE_PYTHON_FRONTEND=On -DDOWNLOAD_DEPENDENCIES=On -DBUILD_STATIC=On" &&
+    BASE_ARGS="$BASE_ARGS -DENABLE_WERROR=On -DENABLE_OLD_FRONTEND=On -DENABLE_PYTHON_FRONTEND=On -DDOWNLOAD_DEPENDENCIES=On -DBUILD_STATIC=On" &&
```
fbrausse: The build.sh script has option `-e` enabling -Werror. It is not a good idea to do this unconditionally, or even by default, since this would prevent us from recommending users to use build.sh to get some compilation going when they don't know how to do otherwise.

The required Github actions already pass `-e ON` enabling -Werror for PRs.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1401,6 +1401,7 @@ void clang_cpp_convertert::get_base_map(
       continue;
 
     auto status = map.insert({class_id, base_cxxrd});
+    (void)status;
```
fbrausse, line 1404: Won't this result in a -Wunused warning with -DNDEBUG?

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -621,6 +621,7 @@ void clang_cpp_convertert::get_overriden_methods(
       continue;
 
     auto status = map.insert({method_id, *md_overriden});
+    (void)status;
```
fbrausse, line 624: Same as above.

fbrausse (changes requested):
The build.sh changes shouldn't happen. Besides that just questions about -Wunused.

scripts/build.sh
mikhailramalho: Are you sure we are passing this flag and that this flag works as expected? Because esbmc has several warnings and by enabling Werror on build just doesn't work.

mikhailramalho:
So, I think I understood what's going on: the warnings I've fixed only happen in Release Mode because of -DNDEBUG, and we only test the Debug version in PRs.

When doing a release, werror is disabled.

fbrausse (approved):
LGTM. One question: I'm getting
```
In file included from /usr/lib/llvm/11/include/clang/AST/Stmt.h:18,
                 from /usr/lib/llvm/11/include/clang/AST/Expr.h:23,
                 from /usr/lib/llvm/11/include/clang/AST/DeclCXX.h:22,
                 from /home/kane/scorch/esbmc/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp:13:
/usr/lib/llvm/11/include/clang/AST/StmtIterator.h:78:38: warning: 'template<class _Category, class _Tp, class _Distance, class _Pointer, class _Reference> struct std::iterator' is deprecated [-Wdeprecated-declarations]
   78 |                          public std::iterator<std::forward_iterator_tag,
      |                                      ^~~~~~~~
/usr/lib/gcc/x86_64-pc-linux-gnu/12/include/g++-v12/bits/stl_iterator_base_types.h:127:34: note: declared here
  127 |     struct _GLIBCXX17_DEPRECATED iterator
      |                                  ^~~~~~~~
```
since upgrading to GCC-12 for building ESBMC. I am wondering whether we should include `-Wno-deprecated` or `-Wno-deprecated-declarations` or at least `-Wno-error=deprecated-declarations` when building with `-Werror` enabled.

fbrausse:
@mikhailramalho You can mark the test cases KNOWNBUG instead of THOROUGH. The cases marked with the latter are run for the Linux build.

mikhailramalho:
> LGTM. One question: I'm getting
> 
> ```
> In file included from /usr/lib/llvm/11/include/clang/AST/Stmt.h:18,
>                  from /usr/lib/llvm/11/include/clang/AST/Expr.h:23,
>                  from /usr/lib/llvm/11/include/clang/AST/DeclCXX.h:22,
>                  from /home/kane/scorch/esbmc/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp:13:
> /usr/lib/llvm/11/include/clang/AST/StmtIterator.h:78:38: warning: 'template<class _Category, class _Tp, class _Distance, class _Pointer, class _Reference> struct std::iterator' is deprecated [-Wdeprecated-declarations]
>    78 |                          public std::iterator<std::forward_iterator_tag,
>       |                                      ^~~~~~~~
> /usr/lib/gcc/x86_64-pc-linux-gnu/12/include/g++-v12/bits/stl_iterator_base_types.h:127:34: note: declared here
>   127 |     struct _GLIBCXX17_DEPRECATED iterator
>       |                                  ^~~~~~~~
> ```
> 
> since upgrading to GCC-12 for building ESBMC. I am wondering whether we should include `-Wno-deprecated` or `-Wno-deprecated-declarations` or at least `-Wno-error=deprecated-declarations` when building with `-Werror` enabled.

I managed to reproduce this with llvm 11, maybe we can bump the downloaded version of llvm instead?

fbrausse:
> I managed to reproduce this with llvm 11, maybe we can bump the downloaded version of llvm instead?

We might need to provide a pre-compiled version of llvm+clang ourselves. The pre-built versions by https://github.com/llvm/llvm-project lack some continuity wrt. x86_64 Ubuntu builds. E.g. in 15.0.0 the major release doesn't even contain x86_64 Ubuntu builds anymore, only RHEL-8.4. More importantly though for the latest v15 patch release 15.0.7 there aren't any x86_64 Linux builds provided anymore. Not sure what their release policy is regarding pre-compiled versions.

lucasccordeiro:

PR #1437, opened 2023-10-31
Python frontend: Add support for conditional operator and function flag

brcfarias:
This PR:
- Adds support for the ternary operator, enabling us to handle Python programs with expressions like this:
`return a - b if a > b else 0`

- Introduces support for the `--function` option, allowing users to choose a specific function for verification.


src/python-frontend/python_converter.cpp
```diff
@@ -395,40 +405,73 @@ static codet convert_expression_to_code(exprt &expr)
...
   exprt cond = get_expr(ast_node["test"]);
   cond.location() = get_location_from_decl(ast_node["test"]);
 
+  // Recover type
+  current_element_type = t;
   // Extract 'then' block from AST
-  exprt then = get_block(ast_node["body"]);
+  exprt then;
+  if(ast_node["body"].is_array())
+  {
```
rafaelsamenezes: remove brackets for one-line block

src/python-frontend/python_converter.cpp
```diff
@@ -539,7 +584,8 @@ exprt python_converter::get_block(const nlohmann::json &ast_block)
     case StatementType::IF_STATEMENT:
     case StatementType::WHILE_STATEMENT:
     {
-      get_conditional_stms(element, block);
+      exprt cond = get_conditional_stm(element);
```
rafaelsamenezes, line 583: Why this change? It is not clear from the title/description

rafaelsamenezes (changes requested):
Some nitpicks. Also, it would be good if you could update the title/description with more descriptive content

src/python-frontend/python_converter.cpp
```diff
@@ -590,6 +640,29 @@ bool python_converter::convert()
   // Read all statements
   exprt block_expr = get_block(ast_json["body"]);
 
+  const std::string function = config.options.get_option("function");
```
lucasccordeiro, line 642: Add some explanation here.

lucasccordeiro (approved):
Minor comments. 

src/python-frontend/python_converter.cpp
brcfarias: `get_conditional_stms` was previously used solely for converting if/else and while statements. In this context, we passed a target block (function), and the statement was inserted as an operand. However, with ternary operators, we now have if/else as expressions.

src/python-frontend/python_converter.cpp
brcfarias: Added as comment. Thanks

PR #1439, opened 2023-11-01
[python-frontend] Add support for non-determinism

brcfarias:
This PR:
- Adds support for nondet_X() functions
- Adds regression test for non-determinism

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -7,6 +7,7 @@
 #include <util/message.h>
 
 #include <fstream>
+#include <regex>
 #include <unordered_map>
 
 static const std::unordered_map<std::string, std::string> operator_map = {
@@ -252,16 +253,14 @@ exprt python_converter::get_expr(const nlohmann::json &element)
     {
       std::string func_name = element["func"]["id"];
 
-      // Non-determinism
-      auto starts_with = [](const std::string &str, const std::string &prefix) {
-        return str.compare(0, prefix.length(), prefix) == 0;
-      };
+      // Non-determinism functions restricted to basic types supported in Python
+      std::regex pattern(
+        R"(nondet_(int|char|bool|float)|__VERIFIER_nondet_(int|char|bool|float))");
 
-      std::string nondet("nondet");
-      if(starts_with(func_name, nondet))
+      if(std::regex_match(func_name, pattern))
       {
         // Function name pattern: nondet_(type). e.g: nondet_bool(), nondet_int()
-        size_t underscore_pos = func_name.find("_");
+        size_t underscore_pos = func_name.rfind("_");
         if(underscore_pos != std::string::npos)
         {
           std::string type = func_name.substr(underscore_pos + 1);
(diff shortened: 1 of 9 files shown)


lucasccordeiro (changes requested):
Can you please add some regression tests with __VERIFIER_nondet_X()? See the definition here: https://sv-comp.sosy-lab.org/2023/rules.php

In particular, try to add some regression tests that fail too.

src/python-frontend/python_converter.cpp
```diff
@@ -251,6 +251,26 @@ exprt python_converter::get_expr(const nlohmann::json &element)
...
+      auto starts_with = [](const std::string &str, const std::string &prefix) {
+        return str.compare(0, prefix.length(), prefix) == 0;
+      };
+
+      std::string nondet("nondet");
+      if(starts_with(func_name, nondet))
+      {
+        // Function name pattern: nondet_(type). e.g: nondet_bool(), nondet_int()
+        size_t underscore_pos = func_name.find("_");
+        if(underscore_pos != std::string::npos)
```
fbrausse, line 264: These conditions (starts-with "nondet" and constains "_") are too lax: they would also match e.g. `nondeterministic_procedure()` but should only apply to those functions that follow the naming conventions.

lucasccordeiro:
I agree with @fbrausse.

Please check these definitions: 

https://github.com/esbmc/esbmc/blob/master/src/clang-c-frontend/clang_c_language.cpp#L431

https://github.com/esbmc/esbmc/blob/master/src/clang-c-frontend/clang_c_language.cpp#L445

We should match them.

lucasccordeiro (approved):
Excellent!

Many thanks for addressing our comments, @brcfarias.

src/python-frontend/python_converter.cpp
```diff
@@ -251,10 +252,32 @@ exprt python_converter::get_expr(const nlohmann::json &element)
...
+      std::regex pattern(
+        R"(nondet_(int|char|bool|float)|__VERIFIER_nondet_(int|char|bool|float))");
+
+      if(std::regex_match(func_name, pattern))
+      {
+        // Function name pattern: nondet_(type). e.g: nondet_bool(), nondet_int()
+        size_t underscore_pos = func_name.rfind("_");
+        if(underscore_pos != std::string::npos)
+        {
+          std::string type = func_name.substr(underscore_pos + 1);
```
fbrausse, line 266: This won't work for the `__VERIFIER_*` versions because func_name.find() will return the first match.

fbrausse: If you were to write the regex like `(__VERIFIER_)?nondet_(int|char|bool|float)` you could then use the [`std::match_result::operator[]`](https://en.cppreference.com/w/cpp/regex/match_results/operator_at) on element 2, I think.

fbrausse: Oh, sorry, you're using rfind. I missed that. Please ignore.

fbrausse (approved):
LGTM. One nitpick: The test whether the underscore was found could be removed (or become an assertion).

PR #1444, opened 2023-11-02
Python frontend: Add support for __ESBMC_assume and __VERIFIER_assume

brcfarias:
This PR:
- Adds support for `__ESBMC_assume` and `__VERIFIER_assume`
- Adds regression tests

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -202,6 +202,74 @@ python_converter::get_location_from_decl(const nlohmann::json &ast_node)
   return location;
 }
 
+exprt python_converter::get_function_call(const nlohmann::json &element)
+{
+  if(element.contains("func") && element["_type"] == "Call")
+  {
+    std::string func_name = element["func"]["id"];
+
+    // nondet_X() functions restricted to basic types supported in Python
+    std::regex pattern(
+      R"(nondet_(int|char|bool|float)|__VERIFIER_nondet_(int|char|bool|float))");
+
+    if(std::regex_match(func_name, pattern))
+    {
+      // Function name pattern: nondet_(type). e.g: nondet_bool(), nondet_int()
+      size_t underscore_pos = func_name.rfind("_");
+      if(underscore_pos != std::string::npos)
+      {
+        std::string type = func_name.substr(underscore_pos + 1);
+        exprt rhs = exprt("sideeffect", get_typet(type));
+        rhs.statement("nondet");
+        return rhs;
+      }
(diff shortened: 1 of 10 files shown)


src/python-frontend/python_converter.cpp
```diff
@@ -202,6 +202,71 @@ python_converter::get_location_from_decl(const nlohmann::json &ast_node)
...
+      if(context.find_symbol(symbol_id.c_str()) == nullptr)
+      {
+        // Create/init symbol
+        symbolt symbol;
+        symbol.mode = "C";
+        symbol.module = python_filename;
+        symbol.location = location;
+        symbol.type = code_typet();
+        symbol.name = func_name;
+        symbol.id = func_name;
```
fbrausse, line 241: The `.id` should probably be `symbol_id`, no?

brcfarias: The idea is to associate all `(ESBMC|VERIFIER)assume()` calls with the same symbol. Therefore, we only insert it into the context on the first use. I can change this line to `symbol.id = symbol_id`; for the sake of readability, which will maintain the current behavior since symbol_id is updated with symbol_id = func_name; in the same block.

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

src/python-frontend/python_converter.cpp
fbrausse: I see. Still, the `c:@F@` prefix is usually part of the symbol's id. If it's `py:@F@` here, or no prefix, we're going to have to change https://github.com/esbmc/esbmc/blob/81f107552dfc25bde3598b694632738b2d516f45/src/goto-programs/goto_contractor.cpp#L22 and potentially other places, too.

For reference, this is what the symbol looks like for C code using it in a test.c file:
```
Symbol......: c:@F@__VERIFIER_assume
Base name...: __VERIFIER_assume
Module......: esbmc_intrinsics
Mode........: C (C)
Type........: code
      * arguments: 
        0: argument
            * type: signedbv
                * width: 32
                * #cpp_type: signed_int
            * #location: 
              * file: test.c
              * line: 2
              * function: __VERIFIER_assume
              * column: 31
            * #base_name: cond
            * #identifier: c:test.c@50@F@__VERIFIER_assume@cond
      * return_type: empty
          * #cpp_type: void
Flags.......: lvalue
Location....: file esbmc_intrinsics.h line 82 column 1
```

PR #1445, opened 2023-11-02
[Multi-property] Refactor option "multi-fail-fast"

ChenfengWei0:
```
 --multi-fail-fast n                   stops after first n VCC violation found
                                        in multi property mode
```

e.g.
```c++
int main()
{
  int arr[1];
  arr[3] = 10;
  assert(1 == 0);
}
```

- `--multi-property --multi-fail-fast 0`
```
VERIFICATION SUCCESSFUL
```

- `--multi-property --multi-fail-fast 1`
```
[Counterexample]


State 1 file 2.c line 7 column 3 function main thread 0
Violated property:
  file 2.c line 7 column 3 function main
  assertion 1 == 0
  1 == 0


VERIFICATION FAILED
```

- `--multi-property --multi-fail-fast 2` (>=2)

```
[Counterexample]


State 1 file 2.c line 7 column 3 function main thread 0
Violated property:
  file 2.c line 7 column 3 function main
  assertion 1 == 0
  1 == 0

Slicing for Claim array bounds violated: array `arr' upper bound (0.000s)
Slicing time: 0.000s (removed 10 assignments)
Solving claim 'array bounds violated: array `arr' upper bound' with solver Boolector 3.2.2

[Counterexample]


State 1 file 2.c line 6 column 3 function main thread 0
Violated property:
  file 2.c line 6 column 3 function main
  array bounds violated: array `arr' upper bound
  0


VERIFICATION FAILED
```

I believe this debug option can help us to determine which claim's verification process stuck the multi-property.

---

This PR also fixed some remaining issues discussed in #1337.


src/esbmc/options.cpp
```diff
@@ -482,7 +482,8 @@ const struct group_opt_templ all_cmd_options[] = {
     {"goto-coverage",
      NULL,
      "this activates --make-assert-false and --multi-property and "
```
fbrausse: Please replace the last "and" in this line with a comma - for readability. Lots of "and" in this sentence. ;)

src/esbmc/bmc.cpp
```diff
@@ -746,7 +749,20 @@ smt_convt::resultt bmct::multi_property_check(
...
                        &reached_mul_claims,
                        &total_instance,
-                       &is_goto_cov](const size_t &i) {
+                       &is_goto_cov,
+                       &fail_fast,
+                       &counter](const size_t &i) {
+    //"multi-fail-fast n": stop after first n SATs found.
+    if(!fail_fast.empty() && counter >= stoi(fail_fast))
+    {
+      if(stoi(fail_fast) < 0)
```
fbrausse: Two comments:
1. Repeatedly calling `stoi` is not helping performance nor readability.
2. This check whether the `fail_fast` option's value is out of range should not be done repeatedly by the `job_function`, but rather before it. Also: why is this check behind the condition `!fail_fast.empty() && counter >= stoi(fail_fast)`?

src/esbmc/bmc.cpp
```diff
@@ -770,95 +786,82 @@ smt_convt::resultt bmct::multi_property_check(
...
+      {
+        reached_mul_claims.emplace(cmt_loc);
+        is_unverified = true;
+      }
+      else
+      {
+        // the ins is true if the element was actually inserted
+        auto [it, ins] = reached_claims.emplace(cmt_loc);
+        is_unverified = ins;
+      }
```
fbrausse, line 838: I suppose this `if`/`else` could be simplified to only call `.emplace` once.

src/esbmc/bmc.cpp
```diff
@@ -770,95 +786,82 @@ smt_convt::resultt bmct::multi_property_check(
...
-          {
-            std::ofstream out(fmt::format("{}-{}", output_file, ce_counter++));
-            show_goto_trace(out, ns, goto_trace);
-          }
-          std::ostringstream oss;
-          log_fail("\n[Counterexample]\n");
-          show_goto_trace(oss, ns, goto_trace);
-          log_result("{}", oss.str());
-          final_result = result;
+          std::ofstream out(fmt::format("{}-{}", output_file, ce_counter++));
```
fbrausse: Sorry, I'm seeing this just now. The help message for `--cex-output arg` says
> save the counterexample into a file

Appearently that's not what's happening here. Could you please update the help message to account for
1. the fact that with `--multi-property` multiple cex-output files are being created, and
2. that in this case the format is `arg-N` where N is a decimal increasing from zero(?).

I'm not often thinking about Windows, but isn't appending some string to filenames given by Windows users counter-intuitive? If this is just a debug option, this remark is not super important. But imagining a user accustomed to Windows passing `--multi-property --cex-output cex.txt` the result might be pretty confusing for this user, particularly so when displaying filename extensions has been turned off.

fbrausse (commented):
In general looks good, but I have a few comments about some constructs used in this function, which I didn't realize before.

lucasccordeiro:
@ChenfengWei0: Can I ask you to address @fbrausse's comments?

@Anthonysdu: it would be great if you could assess this PR, too. It is related to your internship ;-)

src/esbmc/options.cpp
ChenfengWei0: Will change

src/esbmc/bmc.cpp
ChenfengWei0: will move `stoi` and out of range checking outside the function.

src/esbmc/bmc.cpp
ChenfengWei0: I think we need this `if-else` because we are handling two different data structures.

```c++
  std::unordered_set<std::string> reached_claims;
  std::unordered_multiset<std::string> reached_mul_claims;
 ...
  if(is_goto_cov)
  {
    reached_mul_claims.emplace(cmt_loc);
    is_unverified = true;
  }
  else
  {
    // the ins is true if the element was actually inserted
    auto [it, ins] = reached_claims.emplace(cmt_loc);
    is_unverified = ins;
  }
```
- goto-cov needs `reached_mul_claims` as we need to collect instances.

Or we can always update `reached_mul_claims` and `reached_claims` together. So that we don't need the `if-else`.

fbrausse: Oh, indeed. I didn't realize those were two different ones. Thanks, that's fine then, no change is necessary.

src/esbmc/bmc.cpp
ChenfengWei0: I changed it to the prefix format. So now it will generate files like `0-cex.txt`, `1-cex.txt`.

I also updated the help message:
```
  --cex-output arg                      save the counterexample into a file or,
                                        in multi-property mode, multiple files
                                        with prefix 'N-' where 'N' is a decimal
                                        increasing from zero
```

fbrausse: Thanks!

fbrausse (approved):

Anthonysdu (approved):

lucasccordeiro:

PR #1448, opened 2023-11-03
[interval-analysis] --interval-anaylsis-csv-dump now produce compressed files

Mohannad-Aldughaim:
Using the option --interval-analysis-csv-dump will produce compressed files. This is due to how large the file can become.

### DIFF
--- a/src/goto-programs/abstract-interpretation/interval_analysis.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_analysis.cpp
@@ -5,6 +5,9 @@
 #include <goto-programs/abstract-interpretation/interval_domain.h>
 #include <unordered_set>
 #include <util/prefix.h>
+#include <boost/iostreams/filtering_streambuf.hpp>
+#include <boost/iostreams/copy.hpp>
+#include <boost/iostreams/filter/zlib.hpp>
 
 static inline void get_symbols(
   const expr2tc &expr,
@@ -195,8 +198,15 @@ void interval_analysis(
     Forall_goto_functions(f_it, goto_functions)
       dump_intervals(oss, f_it->second, interval_analysis);
 
-    std::ofstream csv(csv_file);
-    csv << oss.str();
+    // to decompress:
+    // zlib-flate -uncompress intervals.csv.z interval.unz
+    std::stringstream ss{oss.str()};
+    std::ofstream file(
+      csv_file + ".z", std::ios_base::out | std::ios_base::binary);
+    boost::iostreams::filtering_streambuf<boost::iostreams::output> out;
+    out.push(boost::iostreams::zlib_compressor());
+    out.push(file);
+    boost::iostreams::copy(ss, out);
   }
 
   Forall_goto_functions(f_it, goto_functions)


rafaelsamenezes:
Windows build is falling with:

```
023-11-03T14:09:49.1754595Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol deflate referenced in function "protected: int __cdecl boost::iostreams::detail::zlib_base::xdeflate(int)" (?xdeflate@zlib_base@detail@iostreams@boost@@IEAAHH@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1852691Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol deflateEnd referenced in function "protected: void __cdecl boost::iostreams::detail::zlib_base::reset(bool,bool)" (?reset@zlib_base@detail@iostreams@boost@@IEAAX_N0@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1862223Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol inflate referenced in function "protected: int __cdecl boost::iostreams::detail::zlib_base::xinflate(int)" (?xinflate@zlib_base@detail@iostreams@boost@@IEAAHH@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1870472Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol inflateEnd referenced in function "protected: void __cdecl boost::iostreams::detail::zlib_base::reset(bool,bool)" (?reset@zlib_base@detail@iostreams@boost@@IEAAX_N0@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1876331Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol deflateReset referenced in function "protected: void __cdecl boost::iostreams::detail::zlib_base::reset(bool,bool)" (?reset@zlib_base@detail@iostreams@boost@@IEAAX_N0@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1883433Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol inflateReset referenced in function "protected: void __cdecl boost::iostreams::detail::zlib_base::reset(bool,bool)" (?reset@zlib_base@detail@iostreams@boost@@IEAAX_N0@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
2023-11-03T14:09:49.1888941Z boost_iostreams-vc140-mt.lib(zlib.obj) : error LNK2019: unresolved external symbol crc32 referenced in function "protected: void __cdecl boost::iostreams::detail::zlib_base::after(char const * &,char * &,bool)" (?after@zlib_base@detail@iostreams@boost@@IEAAXAEAPEBDAEAPEAD_N@Z) [D:\a\esbmc\esbmc\build\src\esbmc\esbmc.vcxproj]
```

My guess is that we are missing some boost dependency in [here](https://github.com/esbmc/esbmc/blob/master/scripts/build.ps1#L9). I will double check tomorrow.

fbrausse:
> Using the option --interval-analysis-csv-dump will produce compressed files. This is due to how large the file can become.

How large do they get?

Mohannad-Aldughaim:
> > Using the option --interval-analysis-csv-dump will produce compressed files. This is due to how large the file can become.
> 
> How large do they get?

in some cases over 1GB

fbrausse (commented):
Unless you have objections, I would suggest to default to the `gzip_compressor`, because (a) raw zlib is almost never used nowadays (the `zlib-flate` tool you're noting in the comment comes from [qpdf](https://qpdf.sourceforge.net/) which is completely unrelated to ESBMC), and (b) gzip is supported by a lot more programs on all platforms.

As an alternative to linking with zlib, you could also leave it up to the user to compress in whatever format they desire for instance by writing the CSV to stdout (common for that is the "file name" `-`) though this might clash with the `--output` option that can also go to stdout.

rafaelsamenezes:
> Unless you have objections, I would suggest to default to the `gzip_compressor`, because (a) raw zlib is almost never used nowadays (the `zlib-flate` tool you're noting in the comment comes from [qpdf](https://qpdf.sourceforge.net/) which is completely unrelated to ESBMC), and (b) gzip is supported by a lot more programs on all platforms.
> 
> As an alternative to linking with zlib, you could also leave it up to the user to compress in whatever format they desire for instance by writing the CSV to stdout (common for that is the "file name" `-`) though this might clash with the `--output` option that can also go to stdout.

I agree that replacing to gzip seems better. I did not understand what you meant by leaving to the user. Do you mean as compressing all of the input through the message interface?

fbrausse:
> I agree that replacing to gzip seems better. I did not understand what you meant by leaving to the user. Do you mean as compressing all of the input through the message interface?

I meant to just write the CSV data to stdout, raw, no message interface. We moved the message outputs to stderr precisely to have one channel free for writing outputs. I'd imagine the user to run `esbmc [...] --interval-analysis-csv-dump - | gzip > my.csv.gz`.

rafaelsamenezes:
> > I agree that replacing to gzip seems better. I did not understand what you meant by leaving to the user. Do you mean as compressing all of the input through the message interface?
> 
> I meant to just write the CSV data to stdout, raw, no message interface. We moved the message outputs to stderr precisely to have one channel free for writing outputs. I'd imagine the user to run `esbmc [...] --interval-analysis-csv-dump - | gzip > my.csv.gz`.

Ah I see. The issue of that approach is that we were using it in conjunction with benchexec. So it seemed better to extend esbmc than to extend the wrapper.

fbrausse:
> Ah I see. The issue of that approach is that we were using it in conjunction with benchexec.

Could you elaborate? How is this functionality being used with benchexec? Couldn't the wrapper compress the stdout output? Anyways, if that's too complicated I'm fine with linking to zlib, just make sure it's either available everywhere (via instructions and build scripts) or that it's optional and the build can proceed without it.

rafaelsamenezes:
> Couldn't the wrapper compress the stdout output? 

It could, it's just assumes extra software on the runner side. It seems simpler to just have ESBMC doing it.

rafaelsamenezes:
I guess just using python zip librarry (https://docs.python.org/3/library/zipfile.html) inside the wrapper might be even easier. Just:

1. Generate the CSV file
2. Compress it into a zip file
3. Delete the original

fbrausse:
> It could, it's just assumes extra software on the runner side.

Same as now, I guess: `unresolved external symbol deflate`...

> python zip librarry

Or the wrapper could chain together the subprocesses, e.g. something like this:
```python
with open('my.csv.gz', 'w') as f:
  p1 = subprocess.Popen(the_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  p2 = subprocess.Popen('gzip', stdin=p1.stdout, stdout=f)
  (_, stderr) = p1.communicate()
  p2.communicate()
```

mikhailramalho:

This PR has been open since 2023-11-06 with no activity. `--interval-analysis-csv-dump` compressed output is a small UX improvement; if interval-analysis CSV dumps are still big enough to need compression, a fresh PR against current master is easier than rebasing.

PR #1450, opened 2023-11-04
[data-races-check] optimise efficiency

XLiZHI:
The current data race checking strategy processes all assignment instructions for checking, which creates a large number of symbols and increases thread interleaving.

This PR filters the instructions in some libraries to increase the efficiency of verification without affecting the accuracy of verification.

### DIFF
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -86,11 +86,16 @@ void add_race_assertions(
 {
   namespacet ns(context);
 
+  bool is_atomic = false;
+
   Forall_goto_program_instructions(i_it, goto_program)
   {
     goto_programt::instructiont &instruction = *i_it;
 
-    if(instruction.is_assign())
+    if(instruction.is_atomic_begin())
+      is_atomic = true;
+
+    if(instruction.is_assign() && !is_atomic)
     {
       exprt tmp_expr = migrate_expr_back(instruction.code);
       rw_sett rw_set(ns, value_sets, i_it, to_code(tmp_expr));
@@ -155,6 +160,9 @@ void add_race_assertions(
 
       i_it--; // the for loop already counts us up
     }
+
+    if(instruction.is_atomic_end())
+      is_atomic = false;
   }
 
   remove_no_op(goto_program);
@@ -181,7 +189,8 @@ void add_race_assertions(
   w_guardst w_guards(context);
 
   Forall_goto_functions(f_it, goto_functions)
-    add_race_assertions(value_sets, context, f_it->second.body, w_guards);
+    if(f_it->first != goto_functions.main_id())
+      add_race_assertions(value_sets, context, f_it->second.body, w_guards);
 
   // get "main"
   goto_functionst::function_mapt::iterator m_it =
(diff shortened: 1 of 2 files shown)


lucasccordeiro:
@XLiZHI: I think we should avoid using the filename to improve efficiency. 

@rafaelsamenezes has developed some mechanisms to enable and disable the checks within our operational models.

@rafaelsamenezes: could you please provide some insights here?

rafaelsamenezes:
I don't like this approach of using file names as it can generate ambiguity if the user provides their own C files.  Also, IMO only `atomic_` functions should be filtered as they are guaranteed to be atomic. For me the appropriate way would be to manually add `__ESBMC_atomic_begin()` and `__ESBMC_atomic_end()` to all functions that shouldn't be checked for data-race. That being said, I would be fine for now if you just replace the files with the explicit function names that should be skipped.

lucasccordeiro (approved):
That's the correct solution ;-)

Let's see what the CI will report here.

lucasccordeiro:
@XLiZHI: could you please provide some numbers about the performance difference due to this PR?

rafaelsamenezes (approved):
LGTM!

XLiZHI:
> @XLiZHI: could you please provide some numbers about the performance difference due to this PR?

This is a difference running locally, and I think it would be even bigger in CI.
|   | Thread interleavings | Time |
| ------------- | ------------- |-----------|
| 00_race01 |142   | 0m0.836s ->  0m0.598s   |
| 00_race02  | 210   | 0m1.103s ->  0m0.814s  |
| 00_race03  | 247   | 0m1.463s  -> 0m1.050s   |
| cuda/011_value_causing_race  | 400   | 0m40.646s -> 0m36.646s      |

lucasccordeiro:
Master (30s):

````
  correct:           14205
    correct true:     7820
    correct false:    6385
  incorrect:            20
    incorrect true:      5
    incorrect false:    15
  unknown:            9580
  Score:             21625 (max: 38482)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/6710604417
````

This PR (30s):

````
  correct:           14162
    correct true:     7795
    correct false:    6367
  incorrect:            18
    incorrect true:      3
    incorrect false:    15
  unknown:            9625
  Score:             21621 (max: 38482)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/6775741760
````

File: src/goto-programs/add_race_assertions.cpp [hunk 1]
```diff
@@ -181,7 +189,8 @@ void add_race_assertions(
   w_guardst w_guards(context);
 
   Forall_goto_functions(f_it, goto_functions)
-    add_race_assertions(value_sets, context, f_it->second.body, w_guards);
+    if(f_it->first != goto_functions.main_id())
```
mikhailramalho, line 192: Sorry for the late reply to this PR, but what does this line do? Are we ignoring data races in main?

XLiZHI: Yes, I don't think the assignment in `_ESBMC_main` needs to be considered. They are all initialization operations and do not participate in thread interleaving.

PR #1456, opened 2023-11-07
[Concurrency] optimise index races check

XLiZHI:
This PR will improve the case where index is symbol in data races check.

For example TC as follows:
```
#include <pthread.h>

int a[4];

void* t1(void *arg) {
    a[__ESBMC_get_thread_id()] = 1;                 

    return NULL;
}

void* t2(void *arg) {
    a[__ESBMC_get_thread_id()] = 1;

    return NULL;
}

int main() {
    pthread_t id1, id2;

    pthread_create(&id1, NULL, t1, NULL);
    pthread_create(&id2, NULL, t2, NULL);

    return 0;
}
```
goto functions as follow:

8_FW8_3W_R{6QQIPU~))`)5

KI6FKV%`(Z S%Z0AZK_9D6Y

There is a problem with this PR, that is, the number of thread interleaving is increased greatly. I don't quite understand why the `array` has more thread interleaving in the verification. Can we improve this?

### DIFF
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -15,7 +15,8 @@ public:
 
   std::list<irep_idt> w_guards;
 
-  const symbolt &get_guard_symbol(const irep_idt &object)
+  const symbolt &
+  get_guard_symbol(const irep_idt &object, const exprt &original_expr)
   {
     const irep_idt identifier = "tmp_" + id2string(object);
 
@@ -25,10 +26,13 @@ public:
 
     w_guards.push_back(identifier);
 
+    type2tc index = type2tc(array_type2tc(get_bool_type(), expr2tc(), true));
+
     symbolt new_symbol;
     new_symbol.id = identifier;
     new_symbol.name = identifier;
-    new_symbol.type = typet("bool");
+    new_symbol.type =
+      original_expr.is_index() ? migrate_type_back(index) : typet("bool");
     new_symbol.static_lifetime = true;
     new_symbol.value.make_false();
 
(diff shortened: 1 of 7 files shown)


lucasccordeiro:
@XLiZHI: could you please fix the coding style issue?

src/goto-programs/add_race_assertions.cpp
```diff
@@ -37,20 +41,35 @@ class w_guardst
...
   }
 
-  const exprt get_guard_symbol_expr(const irep_idt &object)
+  const exprt
+  get_guard_symbol_expr(const irep_idt &object, const exprt &original_expr)
   {
-    return symbol_expr(get_guard_symbol(object));
+    exprt expr = symbol_expr(get_guard_symbol(object, original_expr));
+
+    if(original_expr.is_index() && expr.type().is_array())
```
lucasccordeiro, line 49: What about if we have a struct instead of an array?

lucasccordeiro (changes requested):
Can you also add the regression tests here from this PR?

You could add one test with the outcome "SUCCESSFUL" and another one with "FAILED".

src/goto-programs/add_race_assertions.cpp
XLiZHI: This type depends on the `tmp` symbol we create, so far we only create two types: `array` and `bool`, we treat sruct/member as "id +'.' + component.as_string()" bool type symbol. 

https://github.com/esbmc/esbmc/blob/0afdaa2400136ef78567f6092e63509e1c8e5c60/src/goto-programs/rw_set.cpp#L68-L73

We'll consider these types later if needed.

XLiZHI:
> @XLiZHI: could you please fix the coding style issue?

The problem comes from the use of "read" as the function name, this happens in `master` as well, and I think changing the name should fix it.

7%QH8NQ853 6RLLS VK%M7R

src/goto-programs/add_race_assertions.cpp
lucasccordeiro: Okay, that's fine by me.

Can you please add the regression tests?

lucasccordeiro (changes requested):

lucasccordeiro (approved):

src/goto-programs/add_race_assertions.cpp
```diff
@@ -25,10 +26,13 @@ class w_guardst
 
     w_guards.push_back(identifier);
 
+    type2tc index = type2tc(array_type2tc(get_bool_type(), expr2tc(), true));
```
fbrausse, line 29: Is the additional `type2tc()` around the array_type2tc needed? If so, it's a bug.

src/goto-programs/add_race_assertions.cpp
```diff
@@ -66,13 +85,16 @@ void w_guardst::add_initialization(goto_programt &goto_program) const
 
   for(const auto &w_guard : w_guards)
   {
-    exprt symbol = symbol_expr(*ns.lookup(w_guard));
+    const symbolt s = *ns.lookup(w_guard);
```
fbrausse, line 88: Can you make this a `const symbolt &s` please? No need to copy the symbol.

fbrausse (approved):
LGTM, just minor nitpicks.

lucasccordeiro:

lucasccordeiro:
I'm sorry that I missed Franz's comments.

@XLiZHI: could you please address them in another (small) PR?

PR #1457, opened 2023-11-08
strerror: fix integer overflow

fbrausse:
This was causing an incorrect result for the no-overflow property of the sv-comp'24 pre-run in (at least)
- c/busybox-1.22.0/logname-1.yml


lucasccordeiro (approved):

Issue #1436, opened 2023-10-31
Data races check strategy


XLiZHI:
```
#include <pthread.h>

void* t1(void *arg) {
    int* shared = (int*) arg;
    *shared = 1;                 

    return NULL;
}

int main() {
    pthread_t id1, id2;
    int shared_data;

    pthread_create(&id1, NULL, t1, &shared_data);

    return 0;
}
```

The following is the current principle of data races check, which can be roughly divided into four parts:

We use the identifier to create a new symbol: `"tmp_" + "c:main6.c@48@F@t1@shared"`

```
t1 (c:@F@t1):
        ...
        tmp_c:main6.c@48@F@t1@shared=1;  // (1). Set

        *shared=1;

        tmp_c:main6.c@48@F@t1@shared=0;  // (2). Reset

        ASSERT !tmp_c:main6.c@48@F@t1@shared // W/W data race on c:main6.c@48@F@t1@shared  // (3). Assertion

        RETURN: 0
        END_FUNCTION // t1
        
__ESBMC_main (__ESBMC_main):
       ...
        tmp_c:main6.c@48@F@t1@shared=0; // (4). Initialise 
```

Thus, a violation of assertion (3) during thread interleaving represents the occurrence of a data race.

The method for adding data race check is at the `goto` level, meaning that it is not thread-aware. Therefore, static analysis fails to dereference at the thread level.

But it would be a bit of a mistake to port all the processes to symex, which involves reachability trees and so on. I think the best way to do this is to implement it in goto functions and then dereferencing it in symex:

1. We add a special prefix (_ESBMC_deref) to symbols that need to be dereferenced.
2. In symex for the special prefix we dereferenced it.

It looks like this:
```
t1 (c:@F@t1):
        ...
        _ESBMC_deref_c:main6.c@48@F@t1@shared=1;  // (1). Split into "c:main6.c@48@F@t1@shared" and dereference
                                             to  get "shared_data", and create new symbol "tmp_c: shared_data"

        *shared=1;

        _ESBMC_deref_c:main6.c@48@F@t1@shared=0;  // (2). same to (1)

        ASSERT !_ESBMC_deref_c:main6.c@48@F@t1@shared // W/W data race on c:main6.c@48@F@t1@shared  // (3). Assertion

        RETURN: 0
        END_FUNCTION // t1
        
__ESBMC_main (__ESBMC_main):
       ...
[...]

PR #1460, opened 2023-11-09
[data-races-check] Fixing dereference

XLiZHI:
This PR adds new methods to symex to support dereferencing in data races check.

The process looks like this:

`goto-programs` -> `race check` -> if `dereference` fail -> add "_ESBMC_deref" prefix

`goto-symex` -> `symex_step` - > `symex_assign` & `symex_assert` ->if has prefix, dereference -> create new symbol ->continue...

More details #1436


src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -279,6 +285,7 @@ cudaDeviceList_t *cudaDeviceList = NULL;
 //Insert a device
 void cudaDeviceInsert(int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: There is a return statement in the middle of this function. We need to run __ESBMC_atomic_end() before it, too.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -313,27 +320,29 @@ void cudaDeviceInsert(int device)
...
     //printf("->Device: %d Active:%d\n",auxDevice->id,auxDevice->active);
     auxDevice = auxDevice->prox;
   }
+  __ESBMC_atomic_end();
 }
 
 //Searching for a device in the devices list
 int searchCudaDevice(int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: return stmt in this function, too.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -347,12 +356,14 @@ int searchCudaDevice(int device)
...
     }
   }
+  __ESBMC_atomic_end();
   return 0;
 }
 
 //Checks whether the device is in use
 int cudaDeviceActive(int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: Also return inside.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -370,11 +381,13 @@ int cudaDeviceActive(int device)
     }
   }
   return 0;
+  __ESBMC_atomic_end();
 }
 
 //Start a device
 int cudaDeviceStart(int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: Same.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -386,12 +399,14 @@ int cudaDeviceStart(int device)
...
     auxDevice = auxDevice->prox;
   }
+  __ESBMC_atomic_end();
   return 0;
 }
 
 // Choose a device to work
 cudaError_t cudaSetDevice(int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: Same.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -491,6 +511,7 @@ cudaGetDeviceProperties(struct cudaDeviceProp *prop, int device);
 extern cudaError_t
 cudaGetDeviceProperties(struct cudaDeviceProp *prop, int device)
 {
+  __ESBMC_atomic_begin();
```
fbrausse: Same.

src/cpp/library/CUDA/cuda_runtime_api.h
```diff
@@ -537,37 +559,48 @@ cudaError_t cudaThreadSynchronize()
...
+  __ESBMC_atomic_end();
   return CUDA_SUCCESS;
 }
 
 extern __device__ __cudart_builtin__ cudaError_t CUDARTAPI
 cudaDeviceGetAttribute(int *value, enum cudaDeviceAttr attr, int device)
 {
+  __ESBMC_atomic_begin();
   lastError = CUDA_SUCCESS;
+  __ESBMC_atomic_end();
```
fbrausse: Does an atomic block around these single assignments change the number of interleavings?

fbrausse (changes requested):
Please don't change the code of the test cases unless absolutely necessary, rather add new ones. If one breaks, mark it KNOWNBUG. Why are these changes to the code of test cases necessary?

XLiZHI:
> Please don't change the code of the test cases unless absolutely necessary, rather add new ones. If one breaks, mark it KNOWNBUG. Why are these changes to the code of test cases necessary?

These test cases are incorrect and there is no data race(You can see the index inside), I corrected them.

src/cpp/library/CUDA/cuda_runtime_api.h
XLiZHI: lt will fail the **Codacy Static Code Analysis**.

fbrausse: My understanding is that if we don't do it then the atomic block will be kept open for ESBMC. Is that correct?

lucasccordeiro: You are correct, @fbrausse.

@XLiZHI: you should call atomic-end before the return statement.

lucasccordeiro (approved):
@fbrausse: could you please run this PR over the sv-comp concurrency benchmarks with the data race check enabled?

src/cpp/library/CUDA/cuda_runtime_api.h
fbrausse: @XLiZHI If adding the atomic-end before return causes Codacy to go wild, that's not a big problem. It's better to make ESBMC faster by avoiding interleavings through atomic blocks than to listen to an inferior static-analysis tool. ;) It's not like Codacy doesn't help, it's just not always right.

fbrausse:
> These test cases are incorrect and there is no data race(You can see the index inside), I corrected them.

This is not entirely clear to me. For instance, you're modifying regression/cuda/benchmarks/016_bad_inter_group/main.cu and keep its verdict unchanged. So a few questions run through my mind:
- why are you modifying it?
- why isn't there a new test case containing the modification while keeping the old one as it was?

This applies to all the changed test cases. I will have a closer look later, but if - in the meantime - you could provide some reasoning as to why you think that the test cases have been incorrect (and what that exactly means - by definition a test case is not incorrect, but maybe its verdict is) and need to be fixed, that would be really helpful.

I can understand changing the options for ESBMC because of changes made to its code, though here as well it would be good if there was an explanation.

Disclaimer: I am not super confident in all this threading code, so please be patient with me.

XLiZHI:
Hi @fbrausse, let's analyse the assignment operations involved in these two TCs.

First, our model lets functions run in child threads through UNIX library, in the following code N represents the number of child threads we create, where N = 2. `dev_a` and `1` are args of functions `inline_test`
```
//inline_test<<<1,N>>>(dev_a, 1);	
ESBMC_verify_kernel_intt(inline_test, 1, N, dev_a, 1);
```
I've listed the values of these variables for ease of understanding:
|id |gridDim|blockDim | blockIdx | threadIdx |
| ---| ---| ---   | --- | --- |
|1 |{1, 1, 1}| {2, 1, 1}| {0, 0, 0} |{0, 0, 0}|
|2 |{1, 1, 1}| {2, 1, 1}| {0, 0, 0} |{1, 0, 0}|

In [015_miscfail4/main.cu](https://github.com/esbmc/esbmc/pull/1460/files#diff-e2995cb212a8f63be3f696bdb982cc3b2be90d7765c6c1a436d4d14fd8cb6d64):

id1(offset = 2, tid = threadIdx.x = 0): 
```
int temp = A[tid + offset];       // A[0] = A[2]          after change-> A[0] = A[1]
A[tid] += temp;
```
id2(offset = 2, tid = threadIdx.x = 1): 
```
int temp = A[tid + offset];       // A[1] = A[3]         after change-> A[1] = A[2]        R/W in A[1], Right?
A[tid] += temp;
```

You can see that there is no data race, as for why it passed before, it was because we couldn't handle the case where `index` is symbol inside, which treats all indices as A[0]. As for why we changed `--content-bound`, I don't really understand how it works, but it seems like there's no way to check for data races when bound is 2.

In [016_bad_inter_group/main.cu](https://github.com/esbmc/esbmc/pull/1460/files#diff-f26e015684b35de6285952ae62bd1f227fcfc2dba27c84f44384b3885b34444e)

id1( blockIdx.x = 0, blockDim.x= 2, threadIdx.x = 0): 
```
A[ blockIdx.x*blockDim.x + threadIdx.x ] += (A[ (blockIdx.x + 1)*blockDim.x + threadIdx.x ]);       
// A[0] = A[2]
```
id2( blockIdx.x = 0, blockDim.x= 2, threadIdx.x = 1): 
```
A[ blockIdx.x*blockDim.x + threadIdx.x ] += (A[ (blockIdx.x + 1)*blockDim.x + threadIdx.x ]);
// A[1] = A[3]
```
Same here! I hope I made that clear.

fbrausse:
Many thanks, @XLiZHI! That indeed made it much clearer for me what the nature of the test case modifications is. Translated to my understanding, you're basically saying that previously these tests did spuriously report data-races because of a missing implementation in ESBMC. Now this PR implements the/some missing parts and that's why the tests wouldn't be reported as a data-race now. But you also tried to fix the intention of these tests in order for them to become racey again. I hope I got this sequence kind-of correctly.

If so, then I would still suggest that you don't actually modify the code of those existing test cases, but instead
1. add a copy of them with a new name and with the modifications/corrections, and
2. replace the verdict (VERIFICATION FAILED) in the original tests with the one they should have according to your reasoning laid out above.

My reason for this suggestion is that it's good to test more different conditions. All test cases are an investment into detecting future bugs earlier. You've already invested energy to understand why the original TCs didn't do what they were supposed to. Still the original TCs are doing something else: testing that there is *no* data-race because of "wrong" (or rather: different) indices.

lucasccordeiro:
I fully support @fbrausse's suggestions ;-)

fbrausse:
It seems that these tests have already been modified to use `N=2` instead of the larger numbers in the comments. I believe if you go to a slightly higher N, say 3, they would become racey again. That's an option to test even more conditions.

lucasccordeiro:
@XLiZHI: could you please address @fbrausse's suggestions?

regression/cuda/benchmarks/015_miscfail4/test.desc
```diff
@@ -2,4 +2,4 @@ CORE
 main.cu
 --unwind 9 --force-malloc-success --state-hashing --context-bound 2 --data-races-check --no-bounds-check
 
-^VERIFICATION FAILED$
+^VERIFICATION SUCCESSFUL$
```
fbrausse, line 5: Do you have a failing version of this one?

src/goto-programs/add_race_assertions.cpp
```diff
@@ -15,18 +15,22 @@ class w_guardst
...
-  const symbolt &
-  get_guard_symbol(const irep_idt &object, const exprt &original_expr)
+  const symbolt &get_guard_symbol(
+    const irep_idt &object,
+    const exprt &original_expr,
+    bool deref)
   {
-    const irep_idt identifier = "tmp_" + id2string(object);
+    const irep_idt identifier =
+      deref ? "_ESBMC_deref_" + id2string(object) : "tmp_" + id2string(object);
```
fbrausse: Just for my own sanity: would it be possible to use 2 underscores, i.e., `__ESBMC_deref_` as prefix? I don't think we're using a single underscore (and we shouldn't, because implementation stuff is supposed to use two).

Also, maybe for a later PR, I think we have a similar technique in place already in the value_set: it uses a suffix, though, something like `[]` for arrays and `.member` for structs/unions. It could be that you're replicating a similar strategy here, in which case it might make sense to unify both of them... Not for this PR, though.

src/goto-programs/add_race_assertions.cpp
```diff
@@ -154,7 +162,8 @@ void add_race_assertions(
       }
 
       // now add assignments for what is written -- reset
-      forall_rw_set_entries(e_it, rw_set) if(e_it->second.w)
+      forall_rw_set_entries(
+        e_it, rw_set) if(e_it->second.w || e_it->second.deref)
```
fbrausse, line 166: The formatting here is awful to read. We should do something about that with the clang-formatter. This is a general comment, has nothing to do with this PR.

src/goto-programs/rw_set.cpp
```diff
@@ -37,7 +38,8 @@ void rw_sett::read_write_rec(
     const symbolt *symbol = ns.lookup(symbol_expr.get_identifier());
     if(symbol)
     {
-      if(!symbol->static_lifetime && !expr.type().is_pointer())
+      if(
+        !symbol->static_lifetime && !dereferenced /*!expr.type().is_pointer()*/)
```
fbrausse: Remove commented code?

src/goto-programs/rw_set.cpp
```diff
@@ -77,20 +80,30 @@ void rw_sett::read_write_rec(
...
     expr2tc tmp_expr;
-    migrate_expr(tmp, tmp_expr);
+    migrate_expr(expr, tmp_expr);
     dereference(target, tmp_expr, ns, value_sets);
-    tmp = migrate_expr_back(tmp_expr);
+    exprt tmp = migrate_expr_back(tmp_expr);
+
+    if(
+      has_prefix(id2string(tmp.identifier()), "symex::invalid_object") ||
+      id2string(tmp.identifier()) == "")
```
fbrausse, line 98: Could you please add a comment here about when this particular condition is true and what the purpose of falling back to `expr.op0()` is?

Just as a note: `dereference()` may actually return a large if-then-else expression that's not just a single symbol, so it might be worth looking out for that.

src/goto-programs/rw_set.h
```diff
@@ -87,7 +87,8 @@ class rw_sett
     bool w,
     const std::string &suffix,
     const guardt &guard,
-    const exprt &original_expr);
+    const exprt &original_expr,
+    bool dereferenced = false);
```
fbrausse, line 91: Would it be possible to make `dereferenced` not a defaulted parameter but to pass it explicitly at every call-site? Or are there too many for this PR?

src/goto-symex/builtin_functions.cpp
```diff
@@ -1506,3 +1506,89 @@ void goto_symext::intrinsic_get_object_size(
...
+  else
+  {
+    tmp_exprt = tmp_exprt.op0();
+    iden = tmp_exprt.is_index() ? id2string(tmp_exprt.op0().identifier())
+                                : id2string(tmp_exprt.identifier());
+  }
+
+  // Only instructions with special prefixes are processed
+  // _ESBMC_deref_a = "_ESBMC_deref_" + "a"
+  if(has_prefix(iden, "_ESBMC_deref"))
```
fbrausse: Same comment about double underscores as above.

fbrausse (commented):
Looks quite good now, thanks! I have a couple of remaining questions and suggestions, though.

regression/cuda/benchmarks/015_miscfail4/test.desc
XLiZHI: No need, it's the same type as 014

src/goto-programs/rw_set.h
XLiZHI: Why do we need to explicitly pass this parameter?

lucasccordeiro (approved):

lucasccordeiro:

fbrausse:
Nice! I ran all the previous incorrect ones and this PR fixed:
- c/goblint-regression/04-mutex_10-ptrmunge_nr.yml
- c/goblint-regression/04-mutex_11-ptr_rc.yml
- c/goblint-regression/04-mutex_22-deref_read.yml
- c/goblint-regression/04-mutex_37-indirect_rc.yml
- c/goblint-regression/06-symbeq_15-list_entry_nr.yml

Impressive!

lucasccordeiro:
Excellent 👏🏻

Keep up the good work, @XLiZHI ;-)

PR #1464, opened 2023-11-09
Add option to build an SV-COMP version of ESBMC

fbrausse:
This PR doesn't change defaults, but adds:
- option "ESBMC_SVCOMP" to cmake, causing a #define of the same name to be available during the build
- option "-C" to scripts/build.sh enabling cmake's ESMBC_SVCOMP variable
- if the ESBMC_SVCOMP #define is set, pass it as __ESBMC_SVCOMP on to the C/C++ frontends; this is to make the c2goto models take advantage of it.
- in case it is set, let our `fopen()` model return `(void *)1` and `fclose()` be a no-op.

These are "internal" options in the sense that a user of ESBMC doesn't need to know. The intention is to enable this option for the SV-COMP build, obviously, in order to switch to different OM implementations, if necessary.

*Edit:* If enabled, fixes sv-comp memsafety
- c/loop-industry-pattern/aiob_4.c.v+cfa-reducer.yml
- c/loop-industry-pattern/aiob_4.c.v+lh-reducer.yml
- c/loop-industry-pattern/aiob_4.c.v+lhb-reducer.yml
- c/loop-industry-pattern/aiob_4.c.v+nlh-reducer.yml

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -118,7 +118,7 @@ usage() {
     echo "  -d         enable debug output for this script and c2goto"
     echo "  -S ON|OFF  enable/disable static build [ON for Ubuntu, OFF for macOS]"
     echo "  -c VERS    use packaged clang-VERS in a shared build on Ubuntu [11]"
-    echo "  -C YEAR    build an SV-COMP'YEAR version (YEAR has two digits) []"
+    echo "  -C         build an SV-COMP version [disabled]"
     echo
     echo "This script prepares the environment, downloads dependencies, configures"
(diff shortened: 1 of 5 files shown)


mikhailramalho (approved):
LGTM.

I would just drop the numbering and only define `ESBMC_SVCOMP` for the competition. We used to have stuff specific for each year of esbmc (like the competition scripts, with the options, etc), but after a while these stop making sense: everyone is only interested in the latest results/scripts/options when trying to reproduce our results.

rafaelsamenezes (approved):

Issue #1110, opened 2023-06-20
Error linking ESBMC w/ old cpp frontend and IBEX: `multiple definition of slow`


fbrausse:
```
[0/2] Re-checking globbed directories...
[1/1] Linking CXX executable src/esbmc/esbmc
FAILED: src/esbmc/esbmc 
[...]

PR #1466, opened 2023-11-10
[goto-contractor] Diabled old frontend for build.sh and enabled regression for interval analysis ibex contractor.

Mohannad-Aldughaim:
In this PR:
-Disabled old frontend and enabled goto contractor in build.sh #1110  
-Enabled regression for interval analysis ibex contractor option.
-Cleaned up Cmake files for goto contractor and abstract interpretation.


scripts/build.sh
```diff
@@ -11,12 +11,14 @@ BASE_ARGS="\
...
 "
+# Must disable old frontend to enable goto contractor.
 SOLVER_FLAGS="\
     -DENABLE_BOOLECTOR=On \
     -DENABLE_YICES=Off \
     -DENABLE_CVC4=OFF \
     -DENABLE_BITWUZLA=On \
-    -DENABLE_GOTO_CONTRACTOR=OFF \
+    -DENABLE_GOTO_CONTRACTOR=On \
+    -DACADEMIC_BUILD=On \
```
fbrausse: I'm sorry for seeing this just now, but what is the purpose of ACADEMIC_BUILD? All I can find is that if cmake downloads dependencies, then it enables Ibex being built against [soplex](https://soplex.zib.de/). That thing is Apache-2.0-licensed as well as ESBMC. (Why) do we need that option?

scripts/build.sh
```diff
@@ -62,8 +64,9 @@ ubuntu_setup () {
     echo "Installing Python dependencies" &&
     pip3 install --user meson ast2json &&
     meson --version &&
+    # Must disable goto contractor to enable enable old front end.
```
fbrausse: Please add a reference to Github's #1110 issue to this comment, noone in the future is going to check past PRs.

scripts/build.sh
Mohannad-Aldughaim: I think its for licencing purposes but I am not sure of the details.
@rafaelsamenezes can explain it better.

scripts/build.sh
Mohannad-Aldughaim: I added #1110 and a link to the issue.

scripts/build.sh
rafaelsamenezes: As stated in the [their docs](https://ibex-lib.readthedocs.io/en/latest/install.html#linux-and-macos): 

_"Soplex is under [ZIB](http://scip.zib.de/academic.txt) academic licence. If you intend to use Ibex with Soplex commercially, you may consider contacting Soplex for a commercial licence."_

We should probably disable this for now as we need to inform the user somehow.

rafaelsamenezes: Although... the latest version of soplex is now apache. So we could try to compile manually and just point it during ibex compilation.

fbrausse (approved):

scripts/build.sh
fbrausse: Do we have any use for an soplex-backed ibex? Does it make any difference? Anyways, I'm fine with keeping it for now, though the user should indeed be informed in case this soplex thing makes any measurable difference for ESBMC. The build.sh script passing this option could print a warning...

Mohannad-Aldughaim: From [ibex manual 2.8](https://ibex-lib.readthedocs.io/en/latest/install.html#standard-install).
> Warning The current release of Ibex is compatible with Soplex 3.1.1

Not sure why it's a warning. Maybe a typo.

Also:
>Warning Soplex is a tierce library under [ZIB](http://scip.zib.de/academic.txt) licence, not LGPL. By configuring Ibex with Soplex, you implicitly accept the terms of this licence.

>If you don’t accept these terms, you can configure Ibex without Soplex:
>~/Ibex/ibex-2.8.9/$ sudo ./waf configure
>However, the performances will be strongly degraded, especially for the global optimizer.

From [soplex license](https://soplex.zib.de/index.php#license):
>Releases up to and including Version 6.0.2 remain under the [ZIB Academic License](https://soplex.zib.de/academic.txt) as indicated by the files contained in the releases. The new license applies from Version 6.0.3 onwards.

The soplex downloaded automatically is 4.0.2 
I will run some benchmarks to see if there is any difference in performance with and without soplex. 
I will also try using ibex with latest soplex and see if it works. 
For now, I will add the warning in the findIbex cmake file and build.sh.

lucasccordeiro (approved):

fbrausse (changes requested):
Wait, sorry, in case we're not linking the latest version but a ZIB-licensed one, we can't enable this by default. By default, the build.sh script should produce an Apache-2.0-licensed build. We're using build.sh for the releases.

Mohannad-Aldughaim:
Update:
This may take longer than I thought. I will update the build.sh to not include soplex, for now, to keep it Apache-2.0-licensed build.
I was able to use the latest soplex although I had to update the wrapper for it in ibex in my own [fork.](https://github.com/Mohannad-Aldughaim/ibex-lib)  
Tomorrow I will test it to see if ibex working properly and also check if we really need soplex.

fbrausse:
> Update: This may take longer than I thought. I will update the build.sh to not include soplex, for now, to keep it Apache-2.0-licensed build. I was able to use the latest soplex although I had to update the wrapper for it in ibex in my own [fork.](https://github.com/Mohannad-Aldughaim/ibex-lib)

Thanks for the update. Note however, that your fork will have to remain LGPL-3.0 licensed, as ibex itself is. It is derived work. Actually, now that I think about this, we can't even distribute a statically linked binary (as we usually do) that contains the ibex library, because LGPL only has the linking exception. As soon as we include it, ESBMC would need to be LGPL-3.0 licensed.

I don't really want to get into this licensing, but it is important.

Can I ask why you want to change the defaults for the build.sh script?

fbrausse:
OK, I've started reading up (again) on LGPL linking, and it seems that basically as long as the source-code is available to recompile ESBMC, we can keep Apache-2.0, also for statically linked builds. That means we can enable contractors by default. And soplex as well, in a compatibly licensed version.

https://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic
https://copyleft.org/guide/comprehensive-gpl-guidech12.html#x15-11000011

scripts/cmake/FindIbex.cmake
```diff
@@ -13,6 +13,9 @@ if(DOWNLOAD_DEPENDENCIES AND (NOT DEFINED IBEX_DIR))
 
    message("[ibex] Setting up ibex")
    if(ACADEMIC_BUILD)
+      message(WARNING "ibex is using soplex. Soplex is a tierce library under ZIB licence, not LGPL. \
+      By configuring Ibex with Soplex, you implicitely accept the terms of this licence. \
```
fbrausse: Please rephrase to something similar as https://github.com/esbmc/esbmc/blob/589b1880df82014034c249ad6a9ede9928173ed6/src/solvers/yices/CMakeLists.txt#L80

Most importantly, this affects redistributability of the binary: it cannot be distributed anymore. Also users cannot implicitly accept license terms.

scripts/build.sh
```diff
@@ -11,13 +11,24 @@ BASE_ARGS="\
     -DENABLE_JIMPLE_FRONTEND=On \
     -DCMAKE_INSTALL_PREFIX:PATH=$PWD/release \
 "
+# Must disable old frontend to enable goto contractor. github issue #1110
+# https://github.com/esbmc/esbmc/issues/1110.
```
fbrausse, line 15: Once should be enough, we can remove this comment here.

scripts/build.sh
```diff
@@ -11,13 +11,24 @@ BASE_ARGS="\
...
     -DENABLE_YICES=Off \
     -DENABLE_CVC4=OFF \
     -DENABLE_BITWUZLA=On \
-    -DENABLE_GOTO_CONTRACTOR=OFF \
+    -DENABLE_GOTO_CONTRACTOR=On \
+    -DACADEMIC_BUILD=Off \
 "
+# Check if academic build and goto contractor options are enabled and warn about soplex.
+if [ "$(echo "$SOLVER_FLAGS" | grep -c '\-DENABLE_GOTO_CONTRACTOR=On')" -gt 0 ] &&
+      [ "$(echo "$SOLVER_FLAGS" | grep -c '\-DACADEMIC_BUILD=On')" -gt 0 ]; then
```
fbrausse: I don't think the build.sh script needs to warn if CMake already does.

fbrausse (changes requested):

Mohannad-Aldughaim:
> Can I ask why you want to change the defaults for the build.sh script?

I wanted to have it enabled by default and fix any issues when building with ibex. But i don't mind if its not possible. 

> OK, I've started reading up (again) on LGPL linking, and it seems that basically as long as the source-code is available to recompile ESBMC, we can keep Apache-2.0, also for statically linked builds. That means we can enable contractors by default. And soplex as well, in a compatibly licensed version.
> 
> https://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic https://copyleft.org/guide/comprehensive-gpl-guidech12.html#x15-11000011

Sorry, not sure i understand 100%. If we build ESBMC with statically linked ibex (LGPL) that's statically linked to soplex (ZIB) the resulting ESBMC can be Apache-2.0? If so, sounds good :)

scripts/build.sh
Mohannad-Aldughaim: Removed.

scripts/build.sh
Mohannad-Aldughaim: Removed.

fbrausse:
> Sorry, not sure i understand 100%. If we build ESBMC with statically linked ibex (LGPL) that's statically linked to soplex (ZIB) the resulting ESBMC can be Apache-2.0? If so, sounds good :)

As I understand no: ZIB overrides everything, it's not compatible with either LGPL-3 or Apache-2.0. So we'll either need a recent enough soplex which has Apache-2.0, or no soplex at all.

fbrausse:
The soplex folks are quite picky. Their webpage https://soplex.zib.de/index.php#license says

> Any publication for which SoPlex is used must include an acknowledgement and a reference to one of the following articles, depending on the version used: [...]

We might need to take this into account for the SV-COMP tool description.

Mohannad-Aldughaim:
So I ran some benchmarks with and without soplex. The difference is little and the one without soplex seems to perform better.  So lets keep it out of our build for now.

fbrausse:
> So I ran some benchmarks with and without soplex. The difference is little and the one without soplex seems to perform better. So lets keep it out of our build for now.

Perfect, many thanks @Mohannad-Aldughaim. This simplifies life.

lucasccordeiro:
Dear all,

Can I ask you for the status of this PR?

fbrausse:
I'd still like to have the warning rephrased such that it addresses the main culprit: https://github.com/esbmc/esbmc/pull/1466#discussion_r1390580499

Other than that, I guess we should run sv-comp over this, maybe with the contractors enabled, but maybe not¹, to see effects.

¹ I'm not entirely clear on whether compile-time enabling this functionality already changes things for sv-comp, as there have been changes to the interval-analysis code that are independent to the contractors themselves.

fbrausse:
Master:
```
  correct:           14518
    correct true:     8109
    correct false:    6409
  incorrect:            23
    incorrect true:     15
    incorrect false:     8
  unknown:            9264
  Score:             22019 (max: 38482)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/6851371634
```
This PR:
```
  correct:           14504
    correct true:     8119
    correct false:    6385
  incorrect:            23
    incorrect true:     15
    incorrect false:     8
  unknown:            9278
  Score:             22015 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/6856032445
```

scripts/cmake/FindIbex.cmake
Mohannad-Aldughaim: Changed the message to: 
message(WARNING "the version of ibex you have is ZIB licensed, distribution is impossible.")

Although, I didn't get the lost in space reference here. 
Should i change it to STATUS?

Mohannad-Aldughaim:
I ran benchmarks in reachability category. The improvement happens in loops sub category.  Is it allowed to check which subcategory we are running in the wrapper?

scripts/cmake/FindIbex.cmake
fbrausse: Warning is fine, many thanks!

If you haven't watched it, I'd recommend the 2018 reboot of the series. ;)

fbrausse (approved):

lucasccordeiro:

fbrausse:
> I ran benchmarks in reachability category. The improvement happens in loops sub category. Is it allowed to check which subcategory we are running in the wrapper?

I don't know. From today's meeting I understood that looking at file-names is not an option. @rafaelsamenezes, @shmarovfedor do you know more?

Issue #1165, opened 2023-07-08
Error handling wide string literals


fbrausse:
Consider:
```c
#include <assert.h>
#include <wchar.h>

int main()
{
	wchar_t s[] = L"012";
	size_t n = 4;
	assert(sizeof(s) == sizeof(*s) * n);
	for (size_t i=0; i<n-1; i++)
		assert(s[i] == (wchar_t)(L'0' + i));
}
```
ESBMC incorrectly claims
```
State 1 file wchar.c line 10 function main thread 0
----------------------------------------------------
Violated property:
  file wchar.c line 10 function main
  assertion s[i] == (wchar_t)(L'0' + i)
  s[(signed long int)i] == (signed int)(48 + i)


VERIFICATION FAILED
```

This also affects
- regression/esbmc-unix/github_1009_success,
- regression/esbmc-unix2/github_615_[123]
- regression/esbmc/github_330_7

It is caused by `string_constantt::set_value()` not being aware of wide-character "string constants". It unconditionally adds 1 byte to the size of the passed value (probably to account for '\0'-termination) without considering its `type` information. E.g., in the above example, the size for `L"012"` is computed as 13 == (3 characters) * (4 bytes per character) + 1, while it should be 16(?). However, the counter-example even suggests that not just the size expression is handled incorrectly but also the values inside the string constant.

fbrausse:
The string2array() and constant_string2t::to_array() functions are broken for non-char subtypes.

PR #1469, opened 2023-11-11
Output huge (un)signedbv constants in a "validator-friendly" way, whatever that means

fbrausse:
The constant expression -9223372036854775808 is actually not supported by contemporary C, because the unary - and the rest are not handled together. That means that the literal 9223372036854775808 will be interpreted as an unsigned long long int and the (implicit) cast to signed long long int will be undefined (because it will touch the sign bit).

The second change done here is to output any constant > LLONG_MAX in hex format. Both of these changes are to please CPAchecker, a tool on a mission to banish large numbers...


fbrausse:
Just for fun:
```
Invalid configuration
Exception in thread "main" org.sosy_lab.cpachecker.cpa.automaton.AutomatonGraphmlParser$WitnessParseException: Cannot parse witness: The witness automaton provided is invalid! Reason: Cannot parse <x = 18446744073709551615;>
[...]
Caused by: org.sosy_lab.cpachecker.cfa.parser.eclipse.c.CFAGenerationRuntimeException: Integer value is too large to be represented by the highest possible type (unsigned long long int): 18446744073709551615.
	at org.sosy_lab.cpachecker.cfa.parser.eclipse.c.ASTLiteralConverter.getLeastRepresentedTypeForValue(ASTLiteralConverter.java:434)
	at org.sosy_lab.cpachecker.cfa.parser.eclipse.c.ASTLiteralConverter.parseIntegerLiteral(ASTLiteralConverter.java:143)
[...]
```

mikhailramalho (approved):
Ouch

Do we need all these cases? Do it work if we always print the hex value?

brcfarias (approved):
I understand that the first case is necessary to handle LLONG_MIN, but could we unify the handling of positive values with hex?

fbrausse:
> could we unify the handling of positive values with hex?

Are you asking whether we should print all positive numbers in hex? If yes, then I would say no, because this change applies generally (not just for svcomp), and most of our users are humans.

Or did I misunderstand?

fbrausse:
> Ouch
> 
> Do we need all these cases? Do it work if we always print the hex value?

It wouldn't reduce the number of cases unfortunately as we'd have to convert it to absolute value for comparison and figure out its sign in order to print `"-"` (or not). The BigInt API isn't really helpful here. But maybe I missed something?

fbrausse:
@brcfarias Could you please check whether this change fixes the errors you did see with CPAchecker? Especially this one I'm not sure about:
```
Error: Invalid configuration (Cannot parse witness: The witness automaton provided is invalid! Reason: Cannot parse <iter->element = { .key=(void *)9223934986799808512, .value=(void *)11259136776077632 };>) (AutomatonGraphmlParser.getAssumptions, SEVERE)
```

brcfarias:
> @brcfarias Could you please check whether this change fixes the errors you did see with CPAchecker? 

I can run the command below with the esbmc binary referenced in svcomp repository (This test case is related to the LLONG_MIN error)

`./esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps -D'__builtin_unreachable()' --no-vla-size-check ../../sv-benchmarks/c/Juliet_Test/CWE190_Integer_Overflow__int64_t_fscanf_square_21_bad_abs.i --64 --witness-output CWE190_Integer_Overflow__int64_t_fscanf_square_21_bad_abs.i.graphml --no-pointer-check --no-bounds-check --overflow-check --no-assertions --k-induction --max-inductive-step 3`

but when running it with a locally compiled debug version, it crashes with:

```
esbmc/src/util/string_constant.cpp:24: void string_constantt::set_value(const irep_idt&): Assertion `bv_width(type().subtype()) == config.ansi_c.char_width' failed.
[1]    129708 IOT instruction (core dumped)  ./esbmc --no-div-by-zero-check --force-malloc-success --state-hashing
```

Do you know which compilation flags are used for svcomp?

fbrausse:
Yes, try -DCMAKE_BUILD_TYPE=Release.

This error is quite unfortunate, it's https://github.com/esbmc/esbmc/issues/1165 I think. Long standing and we haven't fixed it, yet. But for most benchmarks it's not relevant.

brcfarias:
Using a release version, the error did not occur anymore; however, the tool printed the following warning:

`Expression <operand1=[data], operand2=[operand1=[-9223372036854775807L], operand2=[1], operator=[MINUS]], operator=[EQUALS]> contains unknown variables: [data] (AutomatonGraphmlParser.getAssumptions, WARNING)
`

And it logged the following result:
`Verification result: TRUE. No property violation found by the chosen configuration.`
for this test, where the expected result is false.

fbrausse:
> Using a release version, the error did not occur anymore;

Nice, thanks.

> however, the tool printed the following warning:
> 
> `Expression <operand1=[data], operand2=[operand1=[-9223372036854775807L], operand2=[1], operator=[MINUS]], operator=[EQUALS]> contains unknown variables: [data] (AutomatonGraphmlParser.getAssumptions, WARNING) `
> 
> And it logged the following result: `Verification result: TRUE. No property violation found by the chosen configuration.` for this test, where the expected result is false.

Yeah, there are other bugs. When you look at line 1589 where ESBMC says the assignment to `data` should be, there is no `data` on that line. It's one after `fscanf (stdin, "%" "l" "d", &data);`.

fbrausse:
> > 
> 
> Yeah, there are other bugs. When you look at line 1589 where ESBMC says the assignment to `data` should be, there is no `data` on that line. It's one after `fscanf (stdin, "%" "l" "d", &data);`.

Reason is that we increment the PC before interpreting the intrinsic: https://github.com/esbmc/esbmc/blob/9283607139a397129d6415a3fbbe7b090aa02115/src/goto-symex/symex_main.cpp#L271-L272 Not sure why we do that, but the code rightly so calls it "misery": https://github.com/esbmc/esbmc/blob/9283607139a397129d6415a3fbbe7b090aa02115/src/goto-symex/builtin_functions.cpp#L1288-L1291

lucasccordeiro (approved):

Issue #1470, opened 2023-11-12
Overflow witness contains no assignment


fbrausse:
For the SV-COMP task c/Juliet_Test/CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad.i ESBMC finds an integer overflow but the counter-example and witness file do not contain assignments to the variable `data` in question. This is the part where the overflow happens in the source file:
```c
void CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad()
{
    int64_t data;
    data = 0LL;
    if(staticReturnsTrue())
    {
        fscanf (stdin, "%" "l" "d", &data);
    }
    if(staticReturnsTrue())
    {
        {
            int64_t result = data + 1; // overflow here
            printLongLongLine(result);
        }
    }
}
```
The counter-example is
```
[Counterexample]


State 4 file CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad.i line 1568 column 13 function CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad thread 0
----------------------------------------------------
Violated property:
  file CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad.i line 1568 column 13 function CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad
  arithmetic overflow on add
  !overflow("+", data, 1)
```
Consequently, the witness file only contains
```
    <node id="N0">
      <data key="entry">true</data>
    </node>
    <node id="N1"/>
    <edge id="E0" source="N0" target="N1">
      <data key="enterFunction">main</data>
      <data key="createThread">0</data>
    </edge>
    <node id="N2">
      <data key="violation">true</data>
    </node>
    <edge id="E1" source="N1" target="N2">
      <data key="startline">1568</data>
      <data key="threadId">0</data>
    </edge>
```
Neither UAutomizer nor CPAchecker find the overflow using this witness. With a `--no-slice` witness at least UAutomizer is able to validate the result.

lucasccordeiro:
Shall we have a run with the option --no-slice enabled?

lucasccordeiro:
Or --compact-trace.

fbrausse:
> With a `--no-slice` witness at least UAutomizer is able to validate the result.

CPAchecker still doesn't want to find the violation.

> Or --compact-trace.

Good idea, I'll try.

lucasccordeiro:
[...]

PR #1472, opened 2023-11-12
symex_input: use source location of intrinsic for assignments

fbrausse:
In symex_step() the PC is incremented before executing the intrinsic, this leads to assignments made by intrinsics to have the wrong source location. Here we fix the symex_input() family of intrinsics to use the correct source location.

For counter-examples and witnesses to also print these assignments, additionally to the other conditions on goto-steps allow function call steps to be interpreted as assignments - if we do that internally.

This fixes a warning printed by CPAchecker about the assigned symbol not being on the source line indicated in the witness file for (at least) SV-COMP benchmarks no-overflow
1. c/Juliet_Test/CWE190_Integer_Overflow__int64_t_fscanf_square_21_bad_abs.i
2. c/Juliet_Test/CWE190_Integer_Overflow__int64_t_fscanf_add_08_bad.i

Unfortunately, CPAchecker still doesn't want to find the property violation, but with --compact-trace this now enables UAutomizer to validate the second one above.

Related: #1470, though we still can't mark it fixed because of --compact-trace. It's really unclear why --compact-trace makes a difference for item 2. above, I can't see anything in the larger counter-example that would be of help for validating the witness...

### DIFF
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -496,6 +496,8 @@ void goto_symext::symex_input(const code_function_call2t &func_call)
   else
     abort();
 
+  cur_state->source.pc--;
+
   if(func_call.ret)
     symex_assign(code_assign2tc(
       func_call.ret,
@@ -527,6 +529,8 @@ void goto_symext::symex_input(const code_function_call2t &func_call)
       symex_assign(code_assign2tc(item.object, val), false, cur_state->guard);
     }
   }
+
+  cur_state->source.pc++;
 }
 
 void goto_symext::symex_cpp_new(const expr2tc &lhs, const sideeffect2t &code)
--- a/src/goto-symex/goto_trace.cpp
+++ b/src/goto-symex/goto_trace.cpp
@@ -283,7 +283,8 @@ void violation_graphml_goto_trace(
     case goto_trace_stept::ASSIGNMENT:
       if(
         step.pc->is_assign() || step.pc->is_return() ||
-        (step.pc->is_other() && is_nil_expr(step.lhs)))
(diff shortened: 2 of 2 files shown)


lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:

mikhailramalho:
Why do we have to decrease it and increment it later?

mikhailramalho:
the step is incremented in line 271, before calling run_intrinsic, can't we just increment it after calling run_instrinsic instead?

fbrausse:
> the step is incremented in line 271, before calling run_intrinsic, can't we just increment it after calling run_instrinsic instead?

Yes. That is the goal of #1473. But there are many intrinsics, all with different implementations, and I don't have the time right now to do all of them in one PR.

mikhailramalho:
Since we know this is not the right solution, we shouldn't be merging it to master.

It should go into the sv-comp branch since it's specifically related to the competition, otherwise, we are yet again adding tech debt to esbmc that very rarely we come back to fix.

fbrausse:
I disagree. The right solution is to remove the increment in symex_step. But it has a lot of dependencies, all of which can cause breakage. So doing one of them after another now, in particular with preference for those that affect SV-COMP, is a correct way to approach this problem. This is not adding debt, it is removing it: once all functions listed in #1473 are converted, we can remove the unnecessary increment before run_intrinsic() together with all the dec-/increments inside the implementations.

fbrausse:
I like to see this as a divide-and-conquer style approach. One at a time, see what breaks, move forward. Once done, the problem is solved. The thing with the intrinsics is that each of them has the same assumptions. So we could change all of them at once, but if something breaks, it will be hard to fix for such a large change. That's why I thought small steps to be better.

PR #1477, opened 2023-11-13
[Concurrency] support "printf" for data races check

XLiZHI:
This PR improves the strategy for data race detection. |

Now identify "printf" as a `read` operation.

JBSOD_28 CQN}O5Z9)M)MLP

### DIFF
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -20,6 +20,11 @@ void rw_sett::compute(const codet &code)
     Forall_operands(it, expr)
       read_rec(*it);
   }
+  else if(statement == "return")
+  {
+    assert(code.operands().size() == 1);
+    read_rec(code.op0());
+  }
 }
 
 void rw_sett::assign(const exprt &lhs, const exprt &rhs)
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -14,6 +14,12 @@ void rw_sett::compute(const codet &code)
     assert(code.operands().size() == 2);
     assign(code.op0(), code.op1());
   }
+  else if(statement == "printf")
+  {
+    exprt expr = code;
+    Forall_operands(it, expr)
+      read_rec(*it);
+  }
 }
 
 void rw_sett::assign(const exprt &lhs, const exprt &rhs)
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -58,7 +58,8 @@ void rw_sett::read_write_rec(
         symbol->name == "__ESBMC_alloc" ||
         symbol->name == "__ESBMC_alloc_size" || symbol->name == "stdin" ||
         symbol->name == "stdout" || symbol->name == "stderr" ||
-        symbol->name == "sys_nerr")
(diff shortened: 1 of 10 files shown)


src/goto-programs/rw_set.cpp
```diff
@@ -14,6 +14,12 @@ void rw_sett::compute(const codet &code)
     assert(code.operands().size() == 2);
     assign(code.op0(), code.op1());
   }
+  else if(statement == "printf")
```
lucasccordeiro, line 17: Can you also check whether we need to consider the read operation for "return"?

lucasccordeiro (changes requested):

Can you also please add a regression test that succeeds?

We should also check whether we must consider the read operation for "return" statements.

lucasccordeiro (approved):
Excellent!

Thanks for addressing our comments, @XLiZHI.

fbrausse (approved):
LGTM!

lucasccordeiro:
Thank you submitting this PR, @XLiZHI.

fbrausse:
```
  correct:             348
    correct true:      311
    correct false:      37
  incorrect:            14
    incorrect true:     12
    incorrect false:     2
  unknown:             421
  Score:               243 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/6851996102
```

lucasccordeiro:
That's getting better and better!

Previous results:

````
  correct:             335
    correct true:      305
    correct false:      30
  incorrect:            24
    incorrect true:     20
    incorrect false:     4
  unknown:             424
  Score:               -64 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/6833596642
````

lucasccordeiro:
We still fail to detect the data race in the main function:

````C
// This file is part of the SV-Benchmarks collection of verification tasks:
// https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks
//
// SPDX-FileCopyrightText: 2005-2021 University of Tartu & Technische Universität München
//
// SPDX-License-Identifier: MIT

#include<pthread.h>
#include<stdlib.h>
#include<stdio.h>

struct s {
  int datum;
  struct s *next;
} *A, *B;

struct s *new(int x) {
  struct s *p = malloc(sizeof(struct s));
  p->datum = x;
  p->next = NULL;
  return p;
}

pthread_mutex_t A_mutex = PTHREAD_MUTEX_INITIALIZER;
pthread_mutex_t B_mutex = PTHREAD_MUTEX_INITIALIZER;

void *t_fun(void *arg) {
  pthread_mutex_lock(&A_mutex);
  A->datum++; // RACE! <-- this line is also relevant.
  pthread_mutex_unlock(&A_mutex);

  pthread_mutex_lock(&B_mutex);
  B->datum++; // <-- this is not relevant at all.
  pthread_mutex_unlock(&B_mutex);
  return NULL;
}

int main () {
  pthread_t t1;

  A = new(3);
  B = new(5);

  pthread_create(&t1, NULL, t_fun, NULL);

  int *data;
  pthread_mutex_lock(&A_mutex);
  data = &A->datum; // NORACE
  pthread_mutex_unlock(&A_mutex);
  *data = 42; // RACE! <-- this is the real bug!
  return 0;
}
````

I think @mikhailramalho pointed out this issue somewhere.

PR #1481, opened 2023-11-13
Python frontend: Add support for recursion

brcfarias:
This PR:
- Adds support for recursion by inserting the function symbol into the context before converting the function body. This addresses the "symbol not found" issue that may occur if the body contains a recursive call.
- Converts function calls to side effects when handling expressions, as "required" by the goto conversion layer.

### DIFF
--- a/regression/python/recursion-fail/main.py
+++ b/regression/python/recursion-fail/main.py
@@ -1,0 +1,11 @@
+def factorial(n:int) -> int:
+    if n == 0 or n == 1:
+        return 1
+    else:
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):
Wonderful, @brcfarias.

Can you also please add a regression test that gives "VERIFICATION FAILED"?

fbrausse (approved):
LGTM! Thanks for addressing my comments from your previous PR.

lucasccordeiro:

PR #1488, opened 2023-11-14
Put some effort into representing array_of terms to the user

### LINKED ISSUE #1484: Avoid printing `ARRAY_OF(n)` in counter-examples and witnesses
Replace `ARRAY_OF(n)` with `{0}` if `n == 0`, otherwise replace it with `{n, n, ..., n}` for constant-size arrays.

fbrausse:
Fixes #1484. I'll add some test cases later.

### DIFF
--- a/regression/esbmc/output-format-array_of-double/main.c
+++ b/regression/esbmc/output-format-array_of-double/main.c
@@ -1,0 +1,7 @@
+
+double A[2];
+
+int main()
+{
+	assert(!A);
+}
--- a/regression/esbmc/output-format-array_of-double/test.desc
+++ b/regression/esbmc/output-format-array_of-double/test.desc
@@ -1,0 +1,6 @@
+CORE
+main.c
+--witness-output - --compact-trace
+^VERIFICATION FAILED$
+\bA = \{ 0\.000000 \}
+>A\b.*\b0\.0000000000000000\b
--- a/regression/esbmc/output-format-array_of-float/main.c
+++ b/regression/esbmc/output-format-array_of-float/main.c
@@ -1,0 +1,7 @@
+
+float A[2];
+
+int main()
+{
+	assert(!A);
+}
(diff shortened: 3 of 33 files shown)


lucasccordeiro (approved):

I would appreciate it if we could make all assignments "compilable" in the counterexample generation.

Later, we can build a violation witness validator that gets these assignments and makes a test harness to validate the counterexample.

fbrausse:
I've merged the functionality of #1493 into this PR, because it was convenient. But it's more flexible now:
- functions languaget::from_(expr|type) now accept a "presentation mode", which can be HUMAN or WITNESS and based on that they choose a set of printing flags, see next item
- the printers now accept a set of flags, currently defined:
  - one for printing zero-initialized compounds in a shorter way allowed by C - this is only enabled for HUMAN presentation
  - one for printing a decimal representation of floatbvs that is accurate enough to uniquely retrieve the bit pattern, this is now only enabled for WITNESS presentation
- the option `--witness-output` now accepts `-` as parameter, meaning stdout

The HUMAN mode is for the printed counter-example, and WITNESS mode obviously for the --witness-output. This means that floats in the counter-example still have their 6 fractional decimals, but the witness file now has the full width required by the type's precision.

rafaelsamenezes (approved):

fbrausse:
@shmarovfedor I've touched the `goto2c` for this, it also allows these flags. But I'm not sure that was the intention. Maybe `goto2c` should have an own default for these printer flags?

lucasccordeiro (approved):

fbrausse:
As we can now run a validator via our CI, I'll start such a run.

fbrausse:
I think it might make sense to put some of the changes from #1469 also behind a flag. I'll look into it.

lucasccordeiro:
@fbrausse: This PR is already very large; you can create other smaller PRs with the suggested improvements.

We should also track the witness validation results as we do for the verification results. This PR can be the baseline.

Thanks for submitting this PR ;-)

fbrausse:
@lucasccordeiro Indeed, sorry about that. It did get that large, because unfortunately the from_(expr|type) functions are tied to the language API and that is spread all across. The svcomp23 run with validation is: https://github.com/esbmc/esbmc/actions/runs/6988615298

PR #1489, opened 2023-11-15
Added support for Witness Validation in benchexec CI

rafaelsamenezes:
This PR:

1. Removes old benchexec-diff.yml action. This is mainly beccause I was using it for debugging and I don't think anybody else uses it.
2. Adds CPAChecker  violation-witness run after we executed esbmc on sv-comp. I also tried to download the artifacts from a previous run... however the proxy madness that happens at the machines prevented me.
3. I removed the `--ci` from the XML. This is mainly done to keep our approach consistent to what CPAChecker uses in their default XML.

Example: https://github.com/esbmc/esbmc/actions/runs/6874158348/job/18695658778


lucasccordeiro:
@rafaelsamenezes: as discussed, it would be great to show the summary of the witness validation such as:

image

fbrausse:
Very nice, many thanks @rafaelsamenezes. That should help us with the violation witness problems. Could you perhaps add a checkbox (or a list where we can select "None" or "CPAchecker") to control whether to run the additional validation stage? What might be good to have as well, but not necessarily for this PR, are our witness files themselves, and potentially also CPAchecker's witness files, which it usually stores in `output/witness.graphml`.

rafaelsamenezes:
>  Could you perhaps add a checkbox (or a list where we can select "None" or "CPAchecker") to control whether to run the additional validation stage? 

Sure

> What might be good to have as well, but not necessarily for this PR, are our witness files themselves, and potentially also CPAchecker's witness files, which it usually stores in output/witness.graphml.

I am not sure what you mean here. ESBMC witnesses are already available in the artifact. I will add the CPAChecker ones though.

fbrausse:
> > 
> 
> I am not sure what you mean here. ESBMC witnesses are already available in the artifact. I will add the CPAChecker ones though.

Thanks, you're absolutely right about ESBMC's. Just CPAChecker's then.

fbrausse (approved):

rafaelsamenezes:
@lucasccordeiro I added the summary

@fbrausse I added the flag and all result files from cpachecker **should** be in the package. I could not test the extra files as apparently we can't produce good witnesses in Reachsafety-Arrays with 10s.

lucasccordeiro (approved):

fbrausse:
I've started a valid-memcleanup run with 30s, shouldn't take too long. If that succeeds, I'll put this change also into svcomp24 and we can have a full run with validation.

fbrausse:
Hm, the paths to the benchmarks don't match, so table-generator doesn't put them into the same line:
image

rafaelsamenezes:
> Hm, the paths to the benchmarks don't match, so table-generator doesn't put them into the same line

I see, I will try to change the directory of cpachecker to the same level of esbmc

fbrausse:
Would be nice if the `sv-benchmarks/` prefix could be removed for both, but I'm not clear on the working directories of both commands.

fbrausse:
For correctness witnesses the cpa-checker run says
```
Invalid configuration: Validating (violation|correctness) witnesses is not supported if option witness.validation.(violation|correctness).config is not specified.
```

lucasccordeiro:
@fbrausse and @rafaelsamenezes: Can I ask you for the status of this PR?

rafaelsamenezes:
> For correctness witnesses the cpa-checker run says

For correctness we need to create another configuration.

lucasccordeiro:
@fbrausse: Can I merge this PR?

fbrausse:
@lucasccordeiro While the table-generator cannot align the tasks between both runs it's not of too much help - or would require some post-processing. It's better to fix it now, I think.

lucasccordeiro:
Thanks, @fbrausse.

@rafaelsamenezes: could you please address @fbrausse's suggestion?

rafaelsamenezes:
@fbrausse could you give more directions on what you are doing? I get the correct diff by copying all the target xml.bz2 into the  same folder and then running table-generator

image

fbrausse:
> @fbrausse could you give more directions on what you are doing? I get the correct diff by copying all the target xml.bz2 into the same folder and then running table-generator

Many thanks, @rafaelsamenezes, this works for me as well. I was missing the "copy all *.xml.bz2 into same directory" step. @lucasccordeiro Good to go.

PR #1491, opened 2023-11-15
Fix wrong line number in if-then-else

### LINKED ISSUE #1490: Wrong line number for assignment in if-then-else
```c
int main()
{
	int k = __VERIFIER_nondet_int() % 1024;
	int b = __VERIFIER_nondet_int() != 0;
	int r = k;
	if (b) // this is line 6
		r |= 16;
	else
		r |= 64;
	assert(r < 10);
}
```
ESBMC finds the correct result, but the line number in state 4 is wrong, it's the one of the if statement, not the one from the actually taken branch.
```
State 1 file test18.c line 3 column 2 function main thread 0
  k = 0 (00000000 00000000 00000000 00000000)

State 2 file test18.c line 4 column 2 function main thread 0
  b = 0 (00000000 00000000 00000000 00000000)

State 3 file test18.c line 5 column 2 function main thread 0
  r = 0 (00000000 00000000 00000000 00000000)

State 4 file test18.c line 6 column 2 function main thread 0
  r = 64 (00000000 00000000 00000000 01000000)

State 5 file test18.c line 10 column 2 function main thread 0
Violated property:
  file test18.c line 10 column 2 function main
  assertion
  r < 10
```
I've found this when investigating sv-comp's
- c/bitvector/s3_clnt_3.BV.c.cil-2a.c

brcfarias:
This PR fixes #1490

`y = tmp_y.instructions.begin();` already initializes y with the correct location for the else block. Therefore, it is not necessary to overwrite it with `y->location = location;`


fbrausse:
Great! With this, cpa-checker can validate our witness for
- c/bitvector/s3_clnt_3.BV.c.cil-2a.c

Thanks, Bruno. Could you add a regression test?

mikhailramalho (approved):
nice

fbrausse (approved):
LGTM! But I'd recommend to make the regression test a bit more stable, something like
```c
int main()
{
	int k = __VERIFIER_nondet_int() % 10;
	int b = __VERIFIER_nondet_int() != 0;
	int r = k & b; // should be zero, we're testing the 'else' case
	if (b)
		r |= 8;
	else
		r |= 64;
	assert(r < 10);
}
```

fbrausse:
@rafaelsamenezes Do you have any idea why the `Interval Analysis - If/Else Statement LHS (Negative)` unit test fails? I noted there are some line-no comparisons in there, which this PR touches (the first instruction in the 'else' branch in the GOTO).

lucasccordeiro (approved):

lucasccordeiro:
@brcfarias : could you please fix the CI?

unit/goto-programs/interval_analysis.test.cpp
```diff
@@ -392,7 +392,7 @@ TEST_CASE(
     "a = -52;\n"
```
fbrausse, line 392: Actually, the comment here should read `a: [MIN_INT, -50]`. Otherwise, good catch, I just wonder why the line number change did trigger it.

PR #1493, opened 2023-11-15
Increase precision of floatbv constants in cex/witness

fbrausse:
This is to ensure that the bit pattern provided by cex is actually the one C would interpret for the decimal representation of the floating point literals.

It's less nice for users to see these enlarged fractional parts, but the previous 6 decimal digits aren't even enough for `float`.

Related: #1492

### DIFF
--- a/src/util/format_spec.h
+++ b/src/util/format_spec.h
@@ -16,12 +16,12 @@ public:
   stylet style;
 
   format_spect()
-    : min_width(0), precision(6), zero_padding(false), style(AUTOMATIC)
+    : min_width(0), precision(16), zero_padding(false), style(AUTOMATIC)
   {
   }
 
   explicit format_spect(stylet _style)
-    : min_width(0), precision(6), zero_padding(false), style(_style)
+    : min_width(0), precision(16), zero_padding(false), style(_style)
   {
   }
 


src/util/format_spec.h
```diff
@@ -16,12 +16,12 @@ class format_spect
   stylet style;
 
   format_spect()
-    : min_width(0), precision(6), zero_padding(false), style(AUTOMATIC)
+    : min_width(0), precision(16), zero_padding(false), style(AUTOMATIC)
```
lucasccordeiro, line 19: Do we need to implement this change to make the witness validators happy?

lucasccordeiro (changes requested):

src/util/format_spec.h
fbrausse: It's in order to have the bit-pattern match how C would parse the decimal floating point numbers. If C gets a different bit pattern, all bets are off that the counter-example actually is one. So, yes and no: I needed this when testing validators to ensure they actually get what ESBMC computed.

So for witness files this should surely be done, but maybe not with the current approach. I would argue that also for counter-examples we should start using one of the existing algorithms that produce the shortest unique float representation.

fbrausse: For instance the `State 5` in #1492 is currently represented to the user as
```
  y = 2.425029e-1f (00111110 01111000 01010010 10110000)
```
But parsing that via strtod() or strtof() yields `0x 3e 78 52 ae`, so the last 5 bits changed.

fbrausse:
Superseded by #1488.

PR #1496, opened 2023-11-16
wrapper: handle deref failure from memset() intrinsic

fbrausse:
For valid-memsafety c/uthash-2.0.2/uthash_SFH_test9-2.i ESBMC's memset intrinsic produces
```
Violated property:
  file string.c line 282 column 3 function memset
  dereference failure: memset of memory segment of size 40 with 44 bytes
```
But the wrapper doesn't recognize this as FALSE_DEREF, yet. With this PR it does.


rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #1502, opened 2023-11-16
Add __ESBMC_unreachable() and --enable-unreachability-intrinsic

fbrausse:
This intrinsic is to be used for operational models that have no suitable implementation for returning a proper result except for `Unknown` from the esbmc-wrapper.py.

*Edit:* It doesn't ensure that we return `Unknown` if used for acosf() for instance, but chances are pretty good. To increase the likelihood we could as a further step check whether the guard is known to be true.


lucasccordeiro (approved):

rafaelsamenezes (approved):
This seems similar to `__builtin_unreachable`

fbrausse:
> This seems similar to `__builtin_unreachable`

Depends, maybe it's just the naming. #1500 makes __builtin_unreachable() a no-op for (most of) sv-comp, because lots of benchmarks have it and there's no point in us symbolically executing something without semantics. This here is to force the wrapper to return `Unknown` when some code path, like through an unimplemented `acosf()` is reachable.

Any suggestions for better names?

fbrausse:
This should actually be mutually exclusive with all the other claims, but I don't know how to achieve that.

fbrausse:
> This should actually be mutually exclusive with all the other claims, but I don't know how to achieve that.

If we had categories of claims and a piece of code that combines these categories based on options, that could help a lot in controlling what the actual assertions are that we're giving to the solver, in which order and with which assumptions. I'd imagine that the multi-property check could also make use of that.

emanino:
I'm going to test it on the nn benchmarks shortly.

emanino (approved):

PR #1503, opened 2023-11-16
Interpret non-adorned fptrs in function argument list

### LINKED ISSUE #1495: Function pointer in argument list recognized as 'code', not 'pointer'
From c/ldv-regression/callfpointer.i
```c
extern void abort(void);
void reach_error() { __assert_fail("0", "callfpointer.i", 3, "reach_error"); }

void f(void g(int)) { // replace with void (*g)(int) for VERIFICATION SUCCESSFUL
 g(1);
}

void h(int i) {
 if(i==1) {
 } else {

 }
}
int main(void) {
 f(h);

 return 0;
}
```
ESBMC seems to incorrectly parse the type of the function pointer parameter to `f` and subsequently it can't call `f`:
```
ERROR: function call: argument "c:callfpointer.i@272@F@f@g" type mismatch: got pointer, expected code
```

fbrausse:
Fixes #1495.


fbrausse:
```diff
-  else if (param_type.is_code())
+  else if(param_type.is_code())
```
Yeah... all the time. Who came up with the idea that not having spaces was the way to go? ESBMC is the only project I know that has this convention.

lucasccordeiro:
We should change that!

lucasccordeiro (approved):

brcfarias (approved):

mikhailramalho (commented):
Isn't the test case supposed to fail?

fbrausse:
@mikhailramalho you're absolutely right, sorry.

mikhailramalho:
> ```diff
> -  else if (param_type.is_code())
> +  else if(param_type.is_code())
> ```
> 
> Yeah... all the time. Who came up with the idea that not having spaces was the way to go? ESBMC is the only project I know that has this convention.

Probably just the default of clang format

fbrausse:
#823 did also have a fix for this, but a different one: https://github.com/esbmc/esbmc/pull/823/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590L686-L687

Basically switching from `.getOriginalType()` to `.getType()` in order for the decays to be applied properly. I'm wondering whether that would be a better way than manually decaying the original type...

PR #1505, opened 2023-11-17
Remove TACAS nondet declarations for svcomp builds

rafaelsamenezes:
Some benchmarks use some weird signature. Since it is unreasonable for us to fix all... let's just assume the return type is the nondet type.

Results: https://github.com/esbmc/esbmc/actions/runs/6905490581

### DIFF
--- a/src/clang-c-frontend/clang_c_language.cpp
+++ b/src/clang-c-frontend/clang_c_language.cpp
@@ -447,6 +447,9 @@ float nondet_float();
 double nondet_double();
 
 // TACAS definitions,
+#ifndef __ESBMC_SVCOMP
+// Some TACAS benchmarks use the wrong signature for nondet
+// e.g. uthash_SFH_nondet_test4-2.i  contains `extern int __VERIFIER_nondet_uint(void);`
 int __VERIFIER_nondet_int();
 unsigned int __VERIFIER_nondet_uint();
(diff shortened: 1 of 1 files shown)


mikhailramalho (commented):
I think we should conditionally append the definitions to the strings, instead of adding the ifndef to it.

This string you've changed is added to the program during verification, so we would need to define it in the esbmc wrapper for it to work correctly.

fbrausse:
@mikhailramalho I think it would work, but I agree that conditionally adding the strings is easier to understand.

It would work, because during an sv-comp build of esbmc we have `ESBMC_SVCOMP` set to 1 and during c2goto (and normal operation) we pass `-D__ESBMC_SVCOMP` to the clang frontend.

fbrausse:
In any case, are we sure that all sv-comp tasks declare the prototypes they use?

rafaelsamenezes:
Results (30s):
```
 Statistics:          31725 Files
  correct:           18513
    correct true:    11559
    correct false:    6954
  incorrect:           172
    incorrect true:     39
    incorrect false:   133
  unknown:           13040
  Score:             26696 (max: 53142)
```

Guess it may be worth

fbrausse (commented):
I think it's not useful. :) Shall we close this?

lucasccordeiro:
Yes, definitely.

Issue #1508, opened 2023-11-17
Wrong veredict for float memset in k-induction


rafaelsamenezes:
The following program:

```c
float a;
main() {
  for (;;) {
    memset(&a, 0, sizeof(a));
    reach_error();
    for (;;)
      ;
  }
}
```

Gives an incorrect veredict when using k-induction

fbrausse:
Another version of this reduction:
```c
/* original problem with c/neural-networks/softsign_w4_r1_case_0_unsafe.c-amalgamation.i */

main() {
  for (;;) {
    for (int i=0;
         i<2;
         i++)
      ;
    assert(-1 > 0);
  }
}
```
Run with `--k-induction --ssa-trace --symex-ssa-trace --no-simplify`, for the last k=2 inductive step this trace is output:
```
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#1 == !(!1)

Thread 0 file gh-1508-manual.c line 9 column 5 function main
ASSUME
!(!goto_symex::guard?0!0&0#1) => -1 > 0

Thread 0 file gh-1508-manual.c line 5 column 10 function main
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#4 == !(!1)

Thread 0 file gh-1508-manual.c line 9 column 5 function main
ASSERT
execution_statet::\guard_exec?0!0 => (!(!goto_symex::guard?0!0&0#1) && !(!goto_symex::guard?0!0&0#4) => -1 > 0)
assertion

Thread 0 file gh-1508-manual.c line 11 column 1 function main
ASSUME
0
```
The first ASSUME in there is suspicious to me.

mikhailramalho:
Weird that the base case didn't catch it before the inductive step ran

fbrausse:
Base case for this (k=2):
```
Thread 0 file gh-1508-manual.c line 5 column 10 function main
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#1 == !(!1)

Thread 0 file gh-1508-manual.c line 7 column 10 function main
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#2 == !(!(0 < 2))

Thread 0 file gh-1508-manual.c line 7 column 10 function main
ASSIGNMENT (HIDDEN)
goto_symex::guard?0!0&0#3 == !(!(0 + 1 < 2))

Thread 0 file gh-1508-manual.c line 5 column 5 function main
ASSUME
[...]

PR #1509, opened 2023-11-17
Optimize floats memset for zero initialization

rafaelsamenezes:
As shown in #1508 esbmc can't handle float memset properly as it breaks k-induction for some reason. This PR is just a workaround by enabling an extra simplification for the case of `float f; memset(&f, 0, sizeof(float));`

### DIFF
--- a/regression/esbmc/github_1508_non_optimized-2/main.c
+++ b/regression/esbmc/github_1508_non_optimized-2/main.c
@@ -1,0 +1,9 @@
+float a;
+main() {
+  for (;;) {
+    memset(&a, 0, sizeof(a)-1);
+    reach_error();
+    for (;;)
+      ;
+  }
+}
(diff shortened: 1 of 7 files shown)


lucasccordeiro (approved):
@rafaelsamenezes: could you also add a regression test that gives "VERIFICATION SUCCESSFUL"?

rafaelsamenezes:
Results (30s):

```
  correct:           14433
    correct true:     8013
    correct false:    6420
  incorrect:            64
    incorrect true:     14
    incorrect false:    50
  unknown:            9308
  Score:             21198 (max: 38482)
```

It solves a few extra cases.

fbrausse (approved):
LGTM. @rafaelsamenezes are these numbers for svcomp main branch or svcomp23 or which ones? Could you add the link?

lucasccordeiro:

PR #1514, opened 2023-11-21
Fixed crash caused by invalid id in value set

XLiZHI:
ESBMC should not crash for program.

```
typedef struct {
  int a;
} b;

int main()
{
  ((b *)60000)->a = 0;
}
```
For some code that directly manipulated memory, this would cause our static analysis module to crash.

### DIFF
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -1247,7 +1247,7 @@ void value_sett::assign_rec(
     {
       const expr2tc obj = object_numbering[it.first];
 
-      if(!is_unknown2t(obj))
+      if(!is_unknown2t(obj) && !is_invalid2t(obj))
         assign_rec(obj, values_rhs, suffix, add_to_sets);
     }
   }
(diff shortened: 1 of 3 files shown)


src/pointer-analysis/value_set.cpp
```diff
@@ -1288,7 +1288,7 @@ void value_sett::assign_rec(
   else if(
     is_constant_string2t(lhs) || is_null_object2t(lhs) ||
     is_valid_object2t(lhs) || is_deallocated_obj2t(lhs) ||
-    is_dynamic_size2t(lhs) || is_constant_array2t(lhs))
+    is_dynamic_size2t(lhs) || is_constant_array2t(lhs) || is_invalid2t(lhs))
```
fbrausse: This check should be done where the invalid2t expression is obtained originally in the recursion: in the `is_dereference2t(lhs)` branch. The reason is that when possible we should handle the "special" value-set expressions `unknown2t`, `invalid2t`, etc. together, such that it's clear that `assign_rec()` itself only handles the non-special ones; i.e., the special should not escape.

XLiZHI: I see, thank you!

fbrausse (approved):
Thank you @XLiZHI, LGTM.

lucasccordeiro (approved):

PR #1515, opened 2023-11-21
Python frontend: Add type inference for dynamic variables

brcfarias:
This PR introduces support for type inference in assignments.

It envolves iterating through the AST to identify assignments lacking annotation information. For each identified assignment, a corresponding node with type information is added to the AST.

For example:

Consider the Python assignment:
```python
n = 10
```
Original AST representation::
```json
{
    "_type": "Module",
    "body": [
        {
            "_type": "Assign",
            "col_offset": 0,
            "end_col_offset": 6,
            "end_lineno": 1,
            "lineno": 1,
            "targets": [
                {
                    "_type": "Name",
                    "col_offset": 0,
                    "ctx": {
                        "_type": "Store"
                    },
                    "end_col_offset": 1,
                    "end_lineno": 1,
                    "id": "n",
                    "lineno": 1
                }
            ],
            "type_comment": null,
            "value": {
                "_type": "Constant",
                "col_offset": 4,
                "end_col_offset": 6,
                "end_lineno": 1,
                "kind": null,
                "lineno": 1,
                "n": 10,
                "s": 10,
                "value": 10
            }
        }
    ],
    "filename": "test.py",
    "type_ignores": []
}
```

After handling the assignment, we generate an AST corresponding to a Python assignment with annotation:

```python
n:int = 10
```
which corresponds to the following AST:

```json
{
    "_type": "Module",
    "body": [
        {
            "_type": "AnnAssign",
            "annotation": {
                "_type": "Name",
                "col_offset": 2,
                "ctx": {
                    "_type": "Load"
                },
                "end_col_offset": 5,
                "end_lineno": 1,
                "id": "int",
                "lineno": 1
            },
            "col_offset": 0,
            "end_col_offset": 10,
            "end_lineno": 1,
            "lineno": 1,
            "simple": 1,
            "target": {
                "_type": "Name",
                "col_offset": 0,
                "ctx": {
                    "_type": "Store"
                },
                "end_col_offset": 1,
                "end_lineno": 1,
                "id": "n",
                "lineno": 1
            },
            "value": {
                "_type": "Constant",
                "col_offset": 8,
                "end_col_offset": 10,
                "end_lineno": 1,
                "kind": null,
                "lineno": 1,
                "n": 10,
                "s": 10,
                "value": 10
            }
        }
    ],
    "filename": "test.py",
    "type_ignores": []
}
```

This PR adds support for these types of type inferences:
```python
a = 10 # infer from constant
b = a # infer from rhs type

def(c:int) -> None :
  d = c # infer from function arg type
```

Other scenarios will be covered in future PRs.

Moreover, we have unit test to check that the generated json precisely matches what would have been produced if the program had originally included type annotations.


lucasccordeiro (changes requested):
@brcfarias: could you please add a regression test that gives "VERIFICATION FAILED"?

fbrausse (approved):
LGTM. If you want to go further with the type inference, I would suggest to be cautious about built-in operations, because many of them can be overridden: https://docs.python.org/3/reference/datamodel.html

rafaelsamenezes (approved):

lucasccordeiro (approved):

PR #1516, opened 2023-11-22
Add OM for pthread_rwlock_rdlock

Anthonysdu:
The regression testcase is a reduced version from the incorrect false of `read_write_lock-1-pthread` in SVCOMP.

From the pthread document: 

The calling thread acquires the write lock if no other thread (reader or writer) holds the read-write lock rwlock.  
The calling thread acquires the read lock if a writer does not hold the lock and there are no writers blocked on the lock.

### DIFF
--- a/regression/esbmc-unix/00_read_write/test.c
+++ b/regression/esbmc-unix/00_read_write/test.c
@@ -1,0 +1,26 @@
+#include <pthread.h>
+#include <assert.h>
+int x, y;
+pthread_rwlock_t rwlock;
+void *writer(void *arg) {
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

Can I also ask you to add a regression test that gives "VERIFICATION FAILED"?

mikhailramalho (approved):
Lgtm

lucasccordeiro (approved):
Many thanks, @Anthonysdu.

fbrausse (approved):
LGTM, thanks!

lucasccordeiro:

PR #1522, opened 2023-11-24
Python frontend: Support for // operator

brcfarias:
This PR:
- Adds support for the floor division (//) operator
- Fixes asserts with negative numbers
- Limits annotation to a specific function when `--function` is used


lucasccordeiro (approved):

PR #1523, opened 2023-11-26
Solidity bug fix

ChenfengWei0:
Fix a bug related to the `current_functionDecl`


lucasccordeiro (commented):
In the release build, the assertions are disabled.

Shall we handle this case and report an error message to the user?

ChenfengWei0:
> In the release build, the assertions are disabled.
> 
> Shall we handle this case and report an error message to the user?

Changed as suggested.

lucasccordeiro (approved):

PR #1525, opened 2023-11-27
Python frontend: Small adjustments

brcfarias:
This PR:
- Fixes assignments within whiles in function bodies.
- Adds type annotation for floor divisions.
- Limits AST print to a specific function when the `--function` option is used.
- Ensures that functions are invoked with with non-det arguments when the `--function` is used.

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -256,7 +256,7 @@ const nlohmann::json python_converter::find_var_decl(
   const std::string &var_name,
   const nlohmann::json &json)
 {
-  for (auto &element : ast_json["body"])
+  for (auto &element : json["body"])
   {
     if (
       (element["_type"] == "AnnAssign") &&
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -252,11 +252,15 @@ exprt python_converter::get_unary_operator_expr(const nlohmann::json &element)
   return unary_expr;
 }
 
-const nlohmann::json python_converter::find_var_decl(const std::string &id)
+const nlohmann::json python_converter::find_var_decl(
+  const std::string &var_name,
+  const nlohmann::json &json)
 {
   for (auto &element : ast_json["body"])
   {
-    if ((element["_type"] == "AnnAssign") && (element["target"]["id"] == id))
+    if (
+      (element["_type"] == "AnnAssign") &&
+      (element["target"]["id"] == var_name))
       return element;
   }
   return nlohmann::json();
@@ -428,7 +432,19 @@ void python_converter::get_var_assign(
   {
     // Get type from declaration node
     std::string var_name = ast_node["targets"][0]["id"].get<std::string>();
-    nlohmann::json ref = find_var_decl(var_name);
+
+    // Get variable from current function
+    nlohmann::json ref;
+    for (const auto &elem : ast_json["body"])
+    {
+      if (elem["_type"] == "FunctionDef" && elem["name"] == current_func_name)
+        ref = find_var_decl(var_name, elem);
+    }
+
+    // Get variable from global scope
+    if (ref.empty())
+      ref = find_var_decl(var_name, ast_json);
+
     assert(!ref.empty());
     current_element_type =
       get_typet(ref["annotation"]["id"].get<std::string>());
@@ -503,7 +519,7 @@ void python_converter::get_compound_assign(
 {
   // Get type from declaration node
   std::string var_name = ast_node["target"]["id"].get<std::string>();
-  nlohmann::json ref = find_var_decl(var_name);
+  nlohmann::json ref = find_var_decl(var_name, ast_json);
   assert(!ref.empty());
   current_element_type = get_typet(ref["annotation"]["id"].get<std::string>());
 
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):

src/python-frontend/python_converter.cpp
```diff
@@ -66,7 +66,10 @@ static typet get_typet(const std::string &ast_type)
 {
   if (ast_type == "float")
     return float_type();
-  if (ast_type == "int")
+  if (
+    ast_type == "int" ||
+    ast_type ==
+      "uint64") /*ToDo: Obtain uint64 type information from class data*/
     return int_type();
```
fbrausse, line 72: This is looking increasingly strange to me.
1. Python's `int` type is unbounded, i.e., it does not have a width and it can't overflow.
2. If the annotated type is `uint64` then how can that correspond to `int_type()`, which for `--32` and `--64` has 32 bits *and* is signed.

brcfarias: > If the annotated type is uint64 then how can that correspond to int_type(), which for --32 and --64 has 32 bits and is signed.

Currently, I am addressing floor divisions involving an `uint64` type. In this context, `uint64` corresponds to a class that inherits from `uint`, which, in turn, inherits from `int`. Additionally, the `uint` class includes a method, `type_byte_length`(), returning the type's size in bytes.

To accurately map `uint64` to `long_uint_type()`, I understand that we would need support for classes and inheritance to determine the type size. Meanwhile, for verifying isolated API functions, I am temporarily adjusting `//` operations to `int` and mapping `uint64` to `int_type`.

With the introduction of class support, we will be able to correctly infer the right-hand side in `// `operations, as demonstrated below:

```python
a: uint64 = 4
b: uint64 = 2
c = a // b
```

> Python's int type is unbounded, i.e., it does not have a width and it can't overflow.

As you mentioned, we have support for unbounded integers in the backend, but I still need to dedicate time to explore this infrastructure and topic.

If it's deemed undesirable to keep this code in the master temporarily, I can revert this change until we have inheritance support, and then we will be able to infer the types of these custom integers without restrictions.

fbrausse: It's no problem having experimental code in the master branch if it's opt-in and/or user-visibly documented as such. We do it all the time, usually via opt-in. As the Python frontend isn't advertised nor enabled by default, yet, that's all fine.

Please correct me if I'm wrong: From your comment it seems that this is a stop-gap solution to get some tests going and it will have to change later. The `ToDo` comment you added doesn't explain that all because it's not clear what more information is needed (`uint64` is most certainly derived from C's `uint64_t` type, so what else would be needed?), and the only thing Python's `int` and this `uint64` seem to have in common that neither of them can actually be modelled by `int_type()`.

Regarding the class/inheritance structure of the unknown-to-me API you mentioned: It's fine to make assumptions, the class structure isn't necessarily required to be understood mechanically in order to get its properties right. If the `uint64` type in Python could actually be modelled by ESBMC's unsignedbv type of 64 bit width, then I suppose it should. That's at least what a comment should say, because priorities can shift and other people might start working on that code. It's confusing. Please either correct it or add a comment explaining this non-obvious choice.

(Note you'll probably need `long_long_uint_type()`; Windows has its own idea what `long` on 64-bit architectures should be.)

> As you mentioned, we have support for unbounded integers in the backend, but I still need to dedicate time to explore this infrastructure and topic.

If we want a first-class unbounded integer type in ESBMC, it's going to be an interesting task: we'll need a new type and support for it everywhere.

brcfarias: > Please correct me if I'm wrong: From your comment it seems that this is a stop-gap solution to get some tests going and it will have to change later. 

Yes, exactly.

> If the uint64 type in Python could actually be modelled by ESBMC's unsignedbv type of 64 bit width, then I suppose it should. Please either correct it or add a comment explaining this non-obvious choice.

I will close this PR to handle this point and subsequently reopen it.

fbrausse: Great, many thanks!

brcfarias:
as

src/python-frontend/python_converter.cpp
brcfarias: @fbrausse Many thanks for your comments.

>The ToDo comment you added doesn't explain that all because it's not clear what more information is needed (uint64 is most certainly derived from C's uint64_t type, so what else would be needed?)

I did some tests involving the `uint64` type:

`x = uint64(0xFFFFFFFFFFFFFFFF) + 1` causes a `raise ValueError(f"value out of bounds for {cls}")` error
`x = uint64(-5)` causes a `ValueError: unsigned type <class 'remerkleable.basic.uint64'> must not be negative` error

> the only thing Python's int and this uint64 seem to have in common that neither of them can actually be modelled by int_type()

I understand that we can use long_long_uint_type  to represent `uint64`

> unknown-to-me API

This is the library: https://github.com/protolambda/remerkleable/blob/91ed092d08ef0ba5ab076f0a34b0b371623db728/remerkleable/basic.py#L243C4-L243C4

> If the uint64 type in Python could actually be modelled by ESBMC's unsignedbv type of 64 bit width, then I suppose it should

Done

fbrausse: Nice, thank you.

fbrausse (approved):
Looks good to me.

lucasccordeiro:

PR #1530, opened 2023-11-30
Do not overwrite location of first instruction during goto-conversion

fbrausse:
During goto-conversion of switch-case code, the location of the first instruction is being overwritten with that of the label of case. This PR fixes the counter-example produced for no-overflow
- c/Juliet_Test/CWE190_Integer_Overflow__int64_t_fscanf_postinc_15_bad.i

(maybe other tasks as well) such that UAutomizer now can validate ESBMC's witness.

### DIFF
--- a/regression/esbmc/switch-line-no/main.c
+++ b/regression/esbmc/switch-line-no/main.c
@@ -1,0 +1,15 @@
+#include <assert.h>
+
+int main()
+{
+	int data = 0;
+	switch (42) {
+	case 42: { // line 7
+		data = nondet_int(); // line 8
+		break;
+	}
+	default:
+		break;
+	}
+	assert(!data);
+}
--- a/regression/esbmc/switch-line-no/test.desc
+++ b/regression/esbmc/switch-line-no/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+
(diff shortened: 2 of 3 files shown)


lucasccordeiro (approved):

fbrausse:
I'll add some regression test case to check the line numbers...

lucasccordeiro:
Thanks, @fbrausse.

Please let me know once you add the regression tests.

BTW, can we also keep tracking of the witness validation results?

fbrausse:
> Thanks, @fbrausse.
> 
> Please let me know once you add the regression tests.

Will do.

> BTW, can we also keep tracking of the witness validation results?

@rafaelsamenezes Did you have any luck in finding out what went wrong with the validator run in <https://github.com/esbmc/esbmc/actions/runs/6988615298>?

lucasccordeiro:
@rafaelsamenezes is traveling to Brazil as we speak. He might take a few days to respond here :-)

fbrausse:
> Please let me know once you add the regression tests.

Done. Once validators can reliably be run via the CI we can start recording those numbers as well, probably makes sense to keep one file per tool and timeout.

lucasccordeiro:

PR #1531, opened 2023-11-30
Fix overflow in ... erhm... the overflow check

fbrausse:
~Besides the fix in the title, this PR also simplifies the usage of powers of two via BigInt. I've hidden the `setPower2()` function and replaced it with `power2()` and `power2m1()`. Additionally, the unary `operator-` on BigInt now takes its argument by-value, thereby preventing additional copies for things like `-BigInt::power2(n)`.~

*Edit:* I'll put the BigInt simplification into a different PR, it uncovered another issue.


lucasccordeiro (approved):

Could you also please add some regression tests?

fbrausse:
> Could you also please add some regression tests?

Done.

lucasccordeiro:
Many thanks, @fbrausse.

PR #1536, opened 2023-12-07
[OM] add Operational Model for semaphore

### LINKED ISSUE #1517: ESBMC - data races check error - semaphores
Hi all,

I am working with the verification of multithreaded applications using ESBMC, specifically utilizing the deadlock and data races check in-built feature of the tool. I recently found an example in C with the data races check where it detects data race despite handling the data race problem using semaphores as seen in the code attached. I have tried the same example with MUTEX and the data race check works fine in that case. Why is the data race condition being detected in the case of semaphores?

Command executed: esbmc rc_ex_sem.c --unwind 2 --no-unwinding-assertions --data-races-check
[rc_ex_sem.docx](https://github.com/esbmc/esbmc/files/13440285/rc_ex_sem.docx)

XLiZHI:
This PR adds the OM of semaphore to support the new locks.


lucasccordeiro (changes requested):
@XLiZHI: could you please add a regression test that fails?

XLiZHI:
Can I ask if `__ESBMC_HIDE:;` is necessary? I found that Codacy Static Code Analysis complains about this.

fbrausse:
> Can I ask if `__ESBMC_HIDE:;` is necessary? I found that Codacy Static Code Analysis complains about this.

It is necessary to hide the implementation details of the operational models from the counter-examples printed by ESBMC. We can ignore Codacy's warning about this.

src/c2goto/library/semaphore_lib.c
```diff
@@ -0,0 +1,106 @@
...
+  __ESBMC_atomic_begin();
+  __ESBMC_assert(!__pshared, "ESBMC does not check multiprocesses for now");
+  __ESBMC_sem_lock_field(*__sem) = 0;
+  __ESBMC_sem_count_field(*__sem) = __value;
+  __ESBMC_sem_init_field(*__sem) = 1;
+  __ESBMC_atomic_end();
+  return 0;
+}
+
+int sem_init_check(sem_t *__sem)
```
fbrausse: As this function is only used internally, could you mark it `static`?

src/c2goto/library/semaphore_lib.c
```diff
@@ -0,0 +1,106 @@
...
+#define __ESBMC_sem_lock_field(a) ((a).__lock)
+#define __ESBMC_sem_count_field(a) ((a).__count)
+#define __ESBMC_sem_init_field(a) ((a).__init)
+
+/************************* Sem manipulation routines ************************/
+
+int sem_init(sem_t *__sem, int __pshared, unsigned int __value)
+{
+  __ESBMC_atomic_begin();
+  __ESBMC_assert(!__pshared, "ESBMC does not check multiprocesses for now");
```
fbrausse: I think this assertion might be misleading to user code working with semaphores, but I'm not 100% sure. The man-page says
>  A process-shared  semaphore  must  be  placed  in a shared memory region (e.g., a System V
              shared memory segment created using shmget(2), or a POSIX  shared  memory  object  built
              created using shm_open(3)).

As we don't support shared memory, yet, it's impossible to check this condition. But we also don't check certain validity conditions for other OMs (e.g., `fscanf` ignores its first argument). I would suggest to ignore the value of `pshared` until `shmget()` and friends are modelled. Maybe add a comment about this instead?

src/c2goto/library/semaphore_lib.c
```diff
@@ -0,0 +1,106 @@
...
+  return 0;
+}
+
+int sem_wait_check(sem_t *__sem)
+{
+  __ESBMC_atomic_begin();
+  sem_init_check(__sem);
+  __ESBMC_sem_count_field(*__sem) -= 1;
+  __ESBMC_assert(
+    __ESBMC_sem_count_field(*__sem) < 0, "Deadlocked state in sem_wait");
```
fbrausse: Maybe I'm misunderstanding, but this assertion seems strange to me. Initializing a semaphore with value 2 and calling sem_wait() once would trigger this assertion, right?

src/c2goto/library/semaphore_lib.c
```diff
@@ -0,0 +1,106 @@
...
+#define __ESBMC_sem_init_field(a) ((a).__init)
+
+/************************* Sem manipulation routines ************************/
+
+int sem_init(sem_t *__sem, int __pshared, unsigned int __value)
+{
+  __ESBMC_atomic_begin();
+  __ESBMC_assert(!__pshared, "ESBMC does not check multiprocesses for now");
+  __ESBMC_sem_lock_field(*__sem) = 0;
+  __ESBMC_sem_count_field(*__sem) = __value;
```
fbrausse, line 23: This can overflow, in which case there would be undefined behaviour, which the `--overflow-check` should catch, but it's optional. In case we do not check for overflows the semaphore's value would end up negative, which should never happen, if I understand correctly. Could you add an assertion that `value <= INT_MAX` or similar?

fbrausse (changes requested):
Could I ask you to add versions of the regression tests that also test the other 2 of the 3 implementations `check`, `noassert` and `nocheck`, e.g. by asserting that `count == 2` at the end of main()?

Independently, the `sem_post_*()` functions all have the same implementation. Why is it necessary to have 3 identical ones? Also the 2 `sem_wait_(noassert|nocheck)` functions are identical.

src/c2goto/library/semaphore_lib.c
XLiZHI: Hi @fbrausse, the purpose of this assertion is to check for deadlock caused by multiple calls to sem_wait(). According to your example, if value is 2, it will trigger this assertion when we call `sem_wait()` for the **third** time in a row. Because our semaphore is decremented each time the function is called.

It is designed to avoid such situations:
```
void *increment(void *arg)
{
  sem_wait(&sem); // decrement the semaphore
  count++;
  sem_wait(&sem); // decrement the semaphore again and it will trigger this assertion, may cause deadlock
  return NULL;
}
...

sem_init(&sem, 0, 1);
```

Anthonysdu:
Could you try all of these regressions with deadlock check? It appears that it always report a violation on my machine.

src/c2goto/library/semaphore_lib.c
fbrausse: > According to your example, if value is 2, it will trigger this assertion when we call `sem_wait()` for the **third** time in a row. Because our semaphore is decremented each time the function is called.

Are you sure? Initializing to 2, calling it once will have `__ESBMC_sem_count_field(*__sem) == 1`, right? So the assertion `[...] < 0` is violated for the first call.

XLiZHI:
> Could you try all of these regressions with deadlock check? It appears that it always report a violation on my machine.

Hi @Anthonysdu, many thanks, I think I did it the wrong way, thanks for pointing it out.

src/c2goto/library/semaphore_lib.c
XLiZHI: Sorry, I understand this is the wrong way to do it and I will fix it.

src/c2goto/headers/pthread.h
```diff
@@ -27,6 +27,10 @@
 #define NULL 0
 #endif
 
+/* Declaration of external variables */
+extern unsigned short int __ESBMC_num_threads_running;
+extern unsigned short int __ESBMC_blocked_threads_count;
```
fbrausse: I'd suggest to not do that in the header file which is user-facing. You could declare these for instance in `semaphore_lib.c` instead or create a new "internal" header inside `c2goto/library`.

XLiZHI: I see, thanks.

fbrausse: Great, many thanks!

fbrausse (approved):
LGTM.

Anthonysdu (approved):

lucasccordeiro (approved):

PR #1538, opened 2023-12-14
SMT memspace fixes

### LINKED ISSUE #1537: ESBMC missing a warning when dropping an irrelevant assumption
Hi,
I just executed `esbmc-linux/linux-release/bin/esbmc --multi-property test.c` with the following code:

```
#include <assert.h>
#include <stddef.h>
extern void __VERIFIER_assume(int);

int main(){
  __VERIFIER_assume(NULL != "a");
  char *str0 = (char *) 0xFFFFFFFFFFFFFFFF;
  char *str1 = "";
  assert(str0 <= str1);
}
```
ESBMC corretly warns me about the assertion, however if I drop the (completely irrelevant) assumption, it doesn't do so, even tough the assertion would clearly still fail.
I am working with ESBMC version 7.4.0 64-bit x86_64 linux.

fbrausse:
This PR addresses a few issues in the SMT conversion:
- The `start` and `end` symbols representing the start and end address of objects that have their address taken were used inconsistently: for the NULL and INVALID symbols, both endpoints were inclusive, i.e., [0,0] for NULL and [1,ADDR_MAX] for INVALID. However, for all other symbols, the `end` part was exclusive, i.e. the objects span [start, end). ~This is fixed to have `end` inclusive for them, too.~ This behaviour was required and I streamlined the memspace code a bit.
- Ordered comparison of pointers was done in a strange way when the pointers didn't refer to the same object (see commit message for details). Unless `--no-pointer-relation-check` is given, during GOTO we already insert same-object assertions for ordered comparisons of pointers, so the SMT layer now assumes this.
- As the pointer-relation-check is required for a sound translation of ordered pointer comparisons now, I've extended the help message disabling it to say `(unsound)` for this option.
- I've briefly looked into comparing just addresses for pointer equality, but things started breaking. So I've added a few comments about why the current version of pointer equality in the SMT layer is done as it is, even if not strictly conforming, together with a TODO note saying how this could be remedied. I don't see an easy way to do it now, so this PR leaves equalities alone.

The first 2 items above together fix #1537.


mikhailramalho (approved):
If disabling `no-pointer-relation-check` is unsound, we should remove the option altogether.

fbrausse:
> If disabling `no-pointer-relation-check` is unsound, we should remove the option altogether.

For developing this the option was quite useful to me, but I agree that it shouldn't be advertised. There is a "debug" category in the help message. How about moving this option there instead of removing it entirely?

fbrausse:
Hm, the two benchexec runs https://github.com/esbmc/esbmc/actions/runs/7210494798 and https://github.com/esbmc/esbmc/actions/runs/7210497316 I started for this PR have aborted. @rafaelsamenezes Do you have any idea why? The log on vm08 says at the end:
```
[2023-12-14 20:54:37Z ERR  GitHubActionsService] POST request to https://pipelinesghubeus21.actions.githubusercontent.com/f763nDHYdR39Y2GeDQWyzkj0DpEhXpmdLXJVThNnBiJrvE5aIT/00000000-0000-0000-0000-000000000000/_apis/distributedtask/hubs/Actions/plans/46b05286-b612-4f59-ad10-2bda0bf4d91e/events failed. HTTP Status: BadRequest
[2023-12-14 20:54:37Z ERR  JobRunner] TaskOrchestrationPlanTerminatedException received, while attempting to raise JobCompletedEvent for job 59b2525c-c828-5d68-e125-c1a11a282638.
[2023-12-14 20:54:37Z ERR  JobRunner] GitHub.DistributedTask.WebApi.TaskOrchestrationPlanTerminatedException: Orchestration plan 46b05286-b612-4f59-ad10-2bda0bf4d91e is not in a runnable state.
   at GitHub.Services.WebApi.VssHttpClientBase.HandleResponseAsync(HttpResponseMessage response, CancellationToken cancellationToken)
   at GitHub.Services.WebApi.VssHttpClientBase.SendAsync(HttpRequestMessage message, HttpCompletionOption completionOption, Object userState, CancellationToken cancellationToken)
   at GitHub.Services.WebApi.VssHttpClientBase.SendAsync(HttpMethod method, Guid locationId, Object routeValues, ApiResourceVersion version, HttpContent content, IEnumerable`1 queryParameters, Object userState, CancellationToken cancellationToken)
   at GitHub.DistributedTask.WebApi.TaskHttpClient.RaisePlanEventAsync[T](Guid scopeIdentifier, String planType, Guid planId, T eventData, CancellationToken cancellationToken, Object userState)
   at GitHub.Runner.Worker.JobRunner.CompleteJobAsync(IJobServer jobServer, IExecutionContext jobContext, AgentJobRequestMessage message, Nullable`1 taskResult)
```
Maybe we need to update the Github runner software?

fbrausse:
The Windows build doesn't like my regex in the new regression test cases, because the assertion in the counter-example appears to have an encoding issue. From the Windows log:
```
  assertion s
t
r
0
 
<
=
 
s
t
r
1
```

lucasccordeiro (approved):

fbrausse:
Master:
```
  correct:           14435
    correct true:     8012
    correct false:    6423
  incorrect:            25
    incorrect true:     15
    incorrect false:    10
  unknown:            9345
  Score:             21807 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7249259396
```
This PR:
```
  correct:           14435
    correct true:     8012
    correct false:    6423
  incorrect:            31
    incorrect true:     15
    incorrect false:    16
  unknown:            9339
  Score:             21711 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7259665084
```
There are:
- no-data-race: 1x timeout -> incorrect (it's noise, already known)
- unreach-call:
  - 2x timeout -> incorrect
  - 3x correct -> incorrect

Of the unreach-call ones, all are in aws-c-common. I'm looking into them.

fbrausse:
Turns out allocating one more address for an object in the smt-memspace at the end, i.e. [start, end] is indeed necessary for pointer equality and for int-to-ptr conversion to work when handling zero-size objects. I checked locally that the aws-c-common tasks are fixed by this recent change restoring the previous allocation behaviour. To be sure, I'll re-run the benchmarks though.

fbrausse:
Yup, fixes it. Master:
```
  correct:           14435
    correct true:     8012
    correct false:    6423
  incorrect:            25
    incorrect true:     15
    incorrect false:    10
  unknown:            9345
  Score:             21807 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7249259396
```
This PR now:
```
  correct:           14525
    correct true:     8090
    correct false:    6435
  incorrect:            27
    incorrect true:     15
    incorrect false:    12
  unknown:            9253
  Score:             21943 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7273793461
```
The 2 "new" incorrect are noise:
- no-data-race: c/pthread-driver-races/char_pc8736x_gpio_pc8736x_gpio_change_pc8736x_gpio_current.i
- unreach-call: c/ldv-linux-3.14/linux-3.14_linux-kernel-locking-spinlock_drivers-net-ethernet-myricom-myri10ge-myri10ge.cil.i

lucasccordeiro:
Excellent, @fbrausse. The numbers look very good!

lucasccordeiro (approved):

PR #1541, opened 2023-12-17
Fix comments of assertion instructions generated from wide/non-char string literals

### LINKED ISSUE #1540: Encoding issue for counter-example on Windows
> The Windows build doesn't like my regex in the new regression test cases, because the assertion in the counter-example appears to have an encoding issue. From the Windows log:
```
  assertion s
t
r
0
 
<
=
 
s
t
r
1
```

_Originally posted by @fbrausse in https://github.com/esbmc/esbmc/issues/1538#issuecomment-1857508165_

@rafaelsamenezes @lucasccordeiro Since you have Windows machines available, could you please take a look?

fbrausse:
This PR tries to address #1540 and also removes a buggy type computation from the `string_constantt` class (related: #1165). Background: ESBMC prints the contents of string literals from the source code into the counter example, e.g. comments of assertions. On Windows (at least), there is `_wassert()` taking a wide character string, which (usually?) is UTF-16 there. Printing that out resulted in the garbage seen in #1540.

This PR extends the irep1 and irep2 string constants to also include a "kind", that is one of:
- default: for `""` literals
- wide: for `L""` literals
- unicode: for `u8""`, `u""` and `U""` literals.

This information is used when computing the comment to put on a user-assertion. The string-literal (unfortunately) has to be converted to the multi-byte representation ESBMC's message system uses (aka narrow strings). The problem here is that we don't know the source / target encodings. For now, the assumption is that the host encoding (given by the locale `""`) will also be used for the target program. I've added a variable `locale_name` (for now always the empty string) to the config. The idea is that we / the user can set this to the default locale name of the target platform. I'm not doing that, yet, so it is always empty (meaning: host platform's locale).

Besides all this locale-handling, there is another potential problem for cross-platform verification: `wchar_t` on Windows is either UTF-16 or GB18030 (as far as I know) and has 2 bytes, everywhere else it is whatever the locale says (usually UTF-32) and has 4 bytes. Cross-platform verification between both, using the above assumption that the host's locale matches the target's cannot work. I don't have a good solution to this problem, so the code catches this case and falls back to printing the garbage from #1540 then.

I've also touched the `string_type2t` to add the correct subtype.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert_literals.cpp
+++ b/src/clang-c-frontend/clang_c_convert_literals.cpp
@@ -46,14 +46,8 @@ bool clang_c_convertert::convert_string_literal(
   }
   else
   {
-    /* TODO: need to take the encoding into account, for that we should examine
-     * string_literal.getKind() to interpret the strings in the correct source
-     * and execution character sets.
-     *
-     * For now we assume all strings are Unicode, in particular char * and
-     * wchar_t * are also Unicode (which isn't true in China: they use GB18030).
-     */
-    string_constantt string(string_literal.getBytes().str(), type);
+    string_constantt string(
+      string_literal.getBytes().str(), type, string_literal.isWide());
     dest.swap(string);
   }
 
--- a/src/clang-c-frontend/clang_c_convert_literals.cpp
+++ b/src/clang-c-frontend/clang_c_convert_literals.cpp
@@ -46,6 +46,13 @@ bool clang_c_convertert::convert_string_literal(
   }
   else
   {
+    /* TODO: need to take the encoding into account, for that we should examine
+     * string_literal.getKind() to interpret the strings in the correct source
+     * and execution character sets.
+     *
+     * For now we assume all strings are Unicode, in particular char * and
+     * wchar_t * are also Unicode (which isn't true in China: they use GB18030).
+     */
     string_constantt string(string_literal.getBytes().str(), type);
     dest.swap(string);
   }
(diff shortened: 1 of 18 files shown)


lucasccordeiro (approved):

mikhailramalho (approved):
lgtm

fbrausse:
Thanks @mikhailramalho. Do you by any chance know whether we can get the source character set (encoding) from Clang in the frontend? After all, it does create these wide strings somehow...

mikhailramalho:
> Thanks @mikhailramalho. Do you by any chance know whether we can get the source character set (encoding) from Clang in the frontend? After all, it does create these wide strings somehow...

I'm not sure, but maybe from clang/Basic/TargetInfo.h? I remember that it provides a bunch of information about the target system (I even left a TODO in the clang_c_convert.cpp file about using it in the future)

fbrausse:
While there's huge amount of useful information in the TargetInfo and similar headers, I couldn't find anything about the source character set. Looks like Clang assumes Unicode always. From <https://reviews.llvm.org/D81530#2086129>:
> The C++ standard doesn't say anything about the encoding of characters on disk (the "input encoding" of bytes -> source character set) - it starts with the source character set, I think. So what do implementations do?
> - clang: from reading the code, clang only supports UTF-8. It supports gcc's -finput-charset flag, but setting it to anything other than UTF-8 is an error!
> - GCC: supports a variety of input encodings, configured with -finput-charset or locale. As such it's not really reasonable for a header designed to be included to require any particular value, as you can't pass a flag for just that header.

Which is fine with the assumptions made this PR, I guess. Would like to know whether that also holds for platforms using GB18030, though. In any case, I think this PR is good to go.

lucasccordeiro:

PR #1543, opened 2023-12-17
[Solidity] support pattern-base checking in contract mode

ChenfengWei0:
In the contract mode, the target function (`--function`) is not given, so we check all the functions in the contract instead.


lucasccordeiro (approved):

fbrausse (commented):
LGTM, thanks!

PR #1544, opened 2023-12-18
Optimise the efficiency of data race detection

XLiZHI:
I found a paper that proposed a way to make race detection in ESBMC significantly reduce thread interleaving.
https://www.academia.edu/download/43202005/TRANSYT_A_Tool_for_the_Verification_of_A20160229-25351-14feb9w.pdf#page=96

image
We can reduce unnecessary thread interleaving by using atomic. At the same time, we can guarantee the correctness of the verification.
```
#include <pthread.h>

int a[4];

void* t1(void *arg) {
    a[__ESBMC_get_thread_id()] = 1;                 

    return NULL;
}

void* t2(void *arg) {
    a[__ESBMC_get_thread_id()] = 1;

    return NULL;
}

int main() {
    pthread_t id1, id2;

    pthread_create(&id1, NULL, t1, NULL);
    pthread_create(&id2, NULL, t2, NULL);

    return 0;
}
```
For example, the above code: `esbmc main.c --data-races-check`
Master:
image
This PR:
image

6000->1000, That's a big improvement.

### DIFF
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -155,7 +155,10 @@ void add_race_assertions(
         i_it = ++t;
       }
 
-      // Avoid adding too much thread interleaving
+      // Avoid adding too much thread interleaving by using atomic block
+      // tmp_A = 0;
+      // atomic {A = n; Assert tmp_A == 0; tmp_A = 1;}
+      // See https://github.com/esbmc/esbmc/pull/1544
       goto_programt::targett t = goto_program.insert(i_it);
       *t = ATOMIC_BEGIN;
       i_it = ++t;
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -69,7 +69,6 @@ public:
 
   const exprt get_w_guard_expr(const rw_sett::entryt &entry)
   {
-    assert(entry.w || entry.deref);
     return get_guard_symbol_expr(
       entry.object, entry.original_expr, entry.deref);
   }
@@ -143,8 +142,7 @@ void add_race_assertions(
       i_it++;
 
       // now add assignments for what is written -- reset
-      forall_rw_set_entries(
-        e_it, rw_set) if (e_it->second.w || e_it->second.deref)
+      forall_rw_set_entries(e_it, rw_set)
       {
         goto_programt::targett t = goto_program.insert(i_it);
 
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -142,6 +142,36 @@ void add_race_assertions(
       instruction.make_skip();
       i_it++;
 
+      // Avoid adding too much thread interleaving
+      goto_programt::targett t = goto_program.insert(i_it);
+      *t = ATOMIC_BEGIN;
+      i_it = ++t;
+
+      // insert original statement here
+      // We need to keep all instructions before the return,
+      // so when we process the return we need add the
+      // original instruction at the end
+      if (!original_instruction.is_return())
+      {
+        goto_programt::targett t = goto_program.insert(i_it);
+
+        *t = original_instruction;
+        i_it = ++t;
+      }
+
+      // now add assertion for what is read and written
+      forall_rw_set_entries(e_it, rw_set)
+      {
+        goto_programt::targett t = goto_program.insert(i_it);
+
+        expr2tc assert;
+        migrate_expr(w_guards.get_assertion(e_it->second), assert);
+        t->make_assertion(assert);
+        t->location = original_instruction.location;
+        t->location.comment(e_it->second.get_comment());
+        i_it = ++t;
+      }
+
       // now add assignments for what is written -- set
       forall_rw_set_entries(e_it, rw_set) if (e_it->second.w)
       {
@@ -157,16 +187,8 @@ void add_race_assertions(
         i_it = ++t;
       }
 
-      // insert original statement here
-      // We need to keep all instructions before the return,
-      // so when we process the return we need add the
-      // original instruction at the end
-      if (!original_instruction.is_return())
-      {
-        goto_programt::targett t = goto_program.insert(i_it);
-        *t = original_instruction;
-        i_it = ++t;
-      }
+      *t = ATOMIC_END;
+      i_it = ++t;
 
       // now add assignments for what is written -- reset
       forall_rw_set_entries(
@@ -180,19 +202,6 @@ void add_race_assertions(
         migrate_expr(theassign, t->code);
 
         t->location = original_instruction.location;
-        i_it = ++t;
-      }
-
-      // now add assertion for what is read and written
-      forall_rw_set_entries(e_it, rw_set)
-      {
-        goto_programt::targett t = goto_program.insert(i_it);
-
-        expr2tc assert;
-        migrate_expr(w_guards.get_assertion(e_it->second), assert);
-        t->make_assertion(assert);
-        t->location = original_instruction.location;
-        t->location.comment(e_it->second.get_comment());
         i_it = ++t;
       }
 
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -142,6 +142,21 @@ void add_race_assertions(
       instruction.make_skip();
       i_it++;
 
+      // now add assignments for what is written -- reset
+      forall_rw_set_entries(
+        e_it, rw_set) if (e_it->second.w || e_it->second.deref)
+      {
+        goto_programt::targett t = goto_program.insert(i_it);
+
+        t->type = ASSIGN;
+        code_assignt theassign(
+          w_guards.get_w_guard_expr(e_it->second), false_exprt());
+        migrate_expr(theassign, t->code);
+
+        t->location = original_instruction.location;
+        i_it = ++t;
+      }
+
       // Avoid adding too much thread interleaving
       goto_programt::targett t = goto_program.insert(i_it);
       *t = ATOMIC_BEGIN;
@@ -187,21 +202,10 @@ void add_race_assertions(
         i_it = ++t;
       }
 
-      *t = ATOMIC_END;
-      i_it = ++t;
-
-      // now add assignments for what is written -- reset
-      forall_rw_set_entries(
-        e_it, rw_set) if (e_it->second.w || e_it->second.deref)
       {
         goto_programt::targett t = goto_program.insert(i_it);
 
-        t->type = ASSIGN;
-        code_assignt theassign(
-          w_guards.get_w_guard_expr(e_it->second), false_exprt());
-        migrate_expr(theassign, t->code);
-
-        t->location = original_instruction.location;
+        *t = ATOMIC_END;
         i_it = ++t;
       }
 


lucasccordeiro:
Many thanks, @XLiZHI.

Can you also add the mentioned test to our regression suite?

XLiZHI:
> Many thanks, @XLiZHI.
> 
> Can you also add the mentioned test to our regression suite?

Hi Dr @lucasccordeiro,

This is a performance improvement, and we already have the corresponding TC in regression. It is in #1456 .

lucasccordeiro:
Excellent, @XLiZHI. Thank you.

Can I ask you to run this PR over the SV-COMP benchmarks, particularly the data-race track?

XLiZHI:
> Excellent, @XLiZHI. Thank you.
> 
> Can I ask you to run this PR over the SV-COMP benchmarks, particularly the data-race track?

Hi Dr @lucasccordeiro, I just found out there are some issues with this PR, I need to look into it.

XLiZHI:
Hi Dr @lucasccordeiro, Could you please show me how to run the PR over the SV-COMP benchmarks, I found that I can only use the branch in the ESBMC repository in action.

fbrausse:
@XLiZHI Indeed, you'll need to push your branch to this repository and then you can use the Github actions to run the benchmarks. I believe you do have rights to push to this repo.

XLiZHI:
> @XLiZHI Indeed, you'll need to push your branch to this repository and then you can use the Github actions to run the benchmarks. I believe you do have rights to push to this repo.

Hi @fbrausse, Thank you for your help. I have not pushed to this repo before, let me have a look.

lucasccordeiro:
image

fbrausse:
> > @XLiZHI Indeed, you'll need to push your branch to this repository and then you can use the Github actions to run the benchmarks. I believe you do have rights to push to this repo.
> 
> Hi @fbrausse, Thank you for your help. I have not pushed to this repo before, let me have a look.

For me it looks like this:
```
$ git remote -v
fb@github	git@github.com:fbrausse/esbmc.git (fetch)
fb@github	git@github.com:fbrausse/esbmc.git (push)
upstream	git@github.com:esbmc/esbmc.git (fetch)
upstream	git@github.com:esbmc/esbmc.git (push)
```
Then I can push to `upstream` like so:
```
$ git push upstream <branch-name>
```

XLiZHI:
> > > @XLiZHI Indeed, you'll need to push your branch to this repository and then you can use the Github actions to run the benchmarks. I believe you do have rights to push to this repo.
> > 
> > Hi @fbrausse, Thank you for your help. I have not pushed to this repo before, let me have a look.
> 
> For me it looks like this:
> ```
> $ git remote -v
> fb@github	git@github.com:fbrausse/esbmc.git (fetch)
> fb@github	git@github.com:fbrausse/esbmc.git (push)
> upstream	git@github.com:esbmc/esbmc.git (fetch)
> upstream	git@github.com:esbmc/esbmc.git (push)
> ```
> Then I can push to `upstream` like so:
> ```
> $ git push upstream <branch-name>
> ```

Many thanks!

XLiZHI:
Master:
```
  correct:             350
    correct true:      314
    correct false:      36
  incorrect:            15
    incorrect true:     13
    incorrect false:     2
  unknown:             418
  Score:               216 (max: 1488)
  
  https://github.com/esbmc/esbmc/actions/runs/7246666881/job/19739584083
```
This PR:
```
  correct:             356
    correct true:      312
    correct false:      44
  incorrect:            24
    incorrect true:     11
    incorrect false:    13
  unknown:             403
  Score:               108(max: 1488)

https://github.com/esbmc/esbmc/actions/runs/7250942122/job/19752791921
```
This PR re-run:
```
  correct:             360
    correct true:      315
    correct false:      45
  incorrect:            24
    incorrect true:     11
    incorrect false:    13
  unknown:             399
  Score:               115 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/7262405095/job/19786051779
```

lucasccordeiro:
@XLiZHI: why do we get new incorrect results? Did they time out before?

XLiZHI:
> @XLiZHI: why do we get new incorrect results? Did they time out before?

Yes, you can see that the number of unknowns has decreased.

lucasccordeiro:
@fbrausse: what do you think about this PR?

fbrausse:
I think the approach here is very good to reduce the number of unnecessary interleavings, but I haven't looked at the changes, yet. It probably makes sense to go through all operational models and find out whether we can add more atomic blocks.

As for the benchmark results, I see
- 2x incorrect true -> correct false
- 3x correct -> timeout (but they already were close to the 30s limit before)
- 7x timeout -> correct (4 of them are solved in about 1sec now)
- 11x timeout -> incorrect false

For the last ones of this it would be good to know whether they're incorrect on master already, but it takes a really long time to run them.

fbrausse (commented):
Looks good and the code is quite understandable if you know the description of this PR. ;)

I'd suggest to add a description of the transformation as a comment to this code.

XLiZHI:
> Looks good and the code is quite understandable if you know the description of this PR. ;)
> 
> I'd suggest to add a description of the transformation as a comment to this code.

Hi @fbrausse,

Are we going to merge this PR? This may result in a lower score. We can also improve it later.

BTW, can you please tell me how you look at the benchmark results, it has a long log and I don't find a summary of incorrect results.

fbrausse:
> Are we going to merge this PR? This may result in a lower score. We can also improve it later.

To me it looks good, I'd just add some of your initial description of this PR as a comment, so future readers understand what's going on. What's still a bit puzzling to me is where the equivalent of `x_write_flag = 0` after the atomic block is created.

> BTW, can you please tell me how you look at the benchmark results, it has a long log and I don't find a summary of incorrect results.

It's a manual process (that can be helped a bit with scripts):
1. For both benchmark runs, go to the Github Action site, download the `esbmc-result` artifact (a .zip file), unpack it and then unpack the .zip contained therein again.
2. Run `table-generator -f html -d -o $OUT $OLD $NEW` where OLD and NEW refer to the .xml.bz2 files you want to compare and OUT is some temporary directory
3. Look at the resulting .html inside the OUT directory.

Alternatively to steps 2 and 3, there are some instructions on the wiki page: https://github.com/esbmc/esbmc/wiki/SV-COMP-and-Test-COMP-FAQ

*Edit:* The table-generator is part of benchexec: https://github.com/sosy-lab/benchexec

XLiZHI:
> What's still a bit puzzling to me is where the equivalent of x_write_flag = 0 after the atomic block is created?

In the figure, `x_write_flag = 0` is below the atomic block, but in our code, I put it above the atomic block. This involves our variable initialisation, and since during symbolic execution we create variables that are not initialised, it will violate our assertion (see #1436). 

for x (read): 
```
x_write_flag = 0
Assert (x_write_flag == 0)
```
This way, we can initialise the variables created during symbolic execution to 0. (Of course, I'm looking for a better way.)

Thanks for your guidance.

fbrausse:
@XLiZHI We should still understand the "new" incorrect ones which were timeouts before. One option to approach them is to run master and see if they were already incorrect with a longer timeout. Another option to go about this is to check the counter-example of the incorrect-false ones (all of the 11) and to find out whether (a) the counter-example is actually valid (in which case the benchmark is wrongly labelled and we should open an issue with sv-benchmarks), or (b) there is an explanation why the counter-example is wrongly generated by ESBMC. If case (b) turns out to have nothing to do with this PR, I'd say this PR is great as it is.

XLiZHI:
[results.2023-12-19_18-58-45.diff.zip](https://github.com/esbmc/esbmc/files/13714954/results.2023-12-19_18-58-45.diff.zip)

Great, the result is successfully generated, but i still have a doubt, we are using the `.i` file right, how should we choose the flag to use?

For example, this code https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i?ref_type=heads

If just use `--data-races-check`, the process will be killed.

fbrausse:
When you click the result cell in the table 
image
an overlay should pop up that contains the log (you might need to start some python command in order for that to work, instructions will be shown). This log is the one from the run, and it contains the exact invocation that was used. In your example I see 
image

So, the wrapper was invoked with
> ./esbmc-wrapper.py -p /home/rafaelsa/sv-benchmarks/c/properties/no-data-race.prp -s kinduction --arch 32 /home/rafaelsa/sv-benchmarks/c/pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i

and that wrapper in turn did execute ESBMC with these parameters:
> ./esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check /home/rafaelsa/sv-benchmarks/c/pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i --32  --no-por --context-bound 2 --witness-output witness.graphml --no-pointer-check --no-bounds-check --data-races-check --no-assertions --incremental-bmc 

Because the path manipulations can be quite tedious, for me it made sense to have a script that prints (or optionally runs) the final ESBMC command. I'm using this one, you'll probably need to adapt some paths at the start, though. [sv-comp.sh.txt](https://github.com/esbmc/esbmc/files/13715117/sv-comp.sh.txt)

But anyways, once you have the log in the pop-up overlay, it will also show the counter-example ESBMC found.

lucasccordeiro:
@XLiZHI: Can I ask you whether you had a chance to investigate the incorrect results?

XLiZHI:
> @XLiZHI: Can I ask you whether you had a chance to investigate the incorrect results?

Sure, I have started to investigate these results, and I need some time due to the large number.

XLiZHI:
| File Path                                          | Comparison | Type of Issue           |
|----------------------------------------------------|------------|-------------------------|
| goblint-regression/28-race_reach_82-list_racefree.i | same       | deref-struct            |
| goblint-regression/06-symbeq_05-funloop_hard2.i     | same       | deref-struct            |
| goblint-regression/28-race_reach_81-list_racing.i   | same       | deref-struct            |
| goblint-regression/28-race_reach_83-list2_racing1.i | same       | deref-struct            |
| goblint-regression/28-race_reach_84-list2_racing2.i | same       | deref-struct            |
| goblint-regression/28-race_reach_85-list2_racefree.i| same       | deref-struct            |
| weaver/unroll-2.wvr.c                               | same       | deref-dynamic object    |
| weaver/unroll-3.wvr.c                               | same       | deref-dynamic object    |
| unroll-cond-2.wvr.c                                 | same       | deref-dynamic object    |

9/11 obtained results consistent with master.

```
pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i
pthread-driver-races/char_pc8736x_gpio_pc8736x_gpio_change_pc8736x_gpio_current.i 
```
But I couldn't reproduce the above two codes, and after investigating, I found that they ran the wrong code (I'm 100% sure), so I assumed that we were using a **regex** to match the filename. The actual codes that the action runs are:
```
pthread-driver-races/char_generic_nvram_read_nvram_write_nvram-race.i
pthread-driver-races/char_pc8736x_gpio_pc8736x_gpio_change_pc8736x_gpio_current-race.i 
```
@fbrausse What do you think? Does our matching method need to be improved?

fbrausse:
@XLiZHI Your analysis is spot on. The pc8736x things got modified recently in the sv-benchmarks repository because of data races, same as the nvram ones: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1450

As the remaining verdicts appearently are unchanged (just a lot faster now), I think this PR is good to go. Though I'd still recommend putting the reasoning about how the instrumentiation actually changes assignments into a comment in the code.

XLiZHI:
> @XLiZHI Your analysis is spot on. The pc8736x things got modified recently in the sv-benchmarks repository because of data races, same as the nvram ones: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1450
> 
> As the remaining verdicts appearently are unchanged (just a lot faster now), I think this PR is good to go. Though I'd still recommend putting the reasoning about how the instrumentiation actually changes assignments into a comment in the code.

@fbrausse Many thanks, I'll add some comments to help understand.

fbrausse:
> Does our matching method need to be improved?

I think you may have just been using a different version of the sv-benchmarks repo. For the Github action by default we still choose the `svcomp23` branch, which doesn't include those `-race` changes you observed.

XLiZHI:
> > Does our matching method need to be improved?
> 
> I think you may have just been using a different version of the sv-benchmarks repo. For the Github action by default we still choose the `svcomp23` branch, which doesn't include those `-race` changes you observed.

~~I think I used the default setting, and in the results I posted above there was an action link.~~
Oh, I see. Never mind.

lucasccordeiro (approved):

fbrausse (approved):
Nice!

PR #1547, opened 2023-12-19
Remove string_typet and string_type2t for good

fbrausse:
These 2 types are only used for constant-string expressions (string literals in the given source code). In C/C++ they have (const) array type, nothing special. Removing the special-casing of string types simplifies the code and doesn't lead to any obvious regressions.

Master:
```
  correct:           14435
    correct true:     8012
    correct false:    6423
  incorrect:            25
    incorrect true:     15
    incorrect false:    10
  unknown:            9345
  Score:             21807 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7249259396
```
This PR:
```
  correct:           14461
    correct true:     8025
    correct false:    6436
  incorrect:            28
    incorrect true:     16
    incorrect false:    12
  unknown:            9316
  Score:             21782 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7250081086
```
I've checked the 3x new timeout -> incorrect: they're already incorrect on master:
- no-data-race: c/pthread-driver-races/char_pc8736x_gpio_pc8736x_gpio_change_pc8736x_gpio_current.i
- unreach-call:
  - c/ldv-linux-3.14/linux-3.14_linux-kernel-locking-spinlock_drivers-net-ethernet-myricom-myri10ge-myri10ge.cil.i
  - c/pthread/triangular-longest-2.i

The rest is timeout noise.

### DIFF
--- a/regression/esbmc/memset-const-2/memset-const.c
+++ b/regression/esbmc/memset-const-2/memset-const.c
@@ -1,0 +1,10 @@
+#include <string.h>
+
+const char b[] = "abc";
+
+int main(int argc, char **argv)
+{
+	char a[2];
+	char *c = argc == 1 ? a : b;
+	memset(c, 0, 1);
+}
--- a/regression/esbmc/memset-const-2/test.desc
+++ b/regression/esbmc/memset-const-2/test.desc
@@ -1,0 +1,4 @@
+KNOWNBUG
+memset-const.c
+
+^VERIFICATION FAILED$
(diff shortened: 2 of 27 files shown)


fbrausse:
The KNOWNBUG regression test `memset-const` added by this PR is *not* buggy on master, so it's a regression introduced by this PR. But: The reason it fails now is that without the explicit string-type the intrinsic_memset() doesn't have a way of finding out about const-ness easily. The other one `memset-const-2` shows this, as it is already incorrect on master: in general we don't take const-ness of the targets symbol into account correctly.

lucasccordeiro (approved):

lucasccordeiro:

PR #1549, opened 2023-12-20
Fix the general case for string-literal to array conversion

### LINKED ISSUE #1165: Error handling wide string literals

fbrausse:
The conversion of constant string expressions to a corresponding array expression did exist for irep1 and separately for irep2. Both did assume `char` as the type underlying the constant and the array. These assumptions are no longer valid as we now support `wchar_t` and `u8`, `u` and `U` string literals.

This PR:
- Replaces the irep1 version of that conversion by delegating to the irep2 one (via migrate, this works also during parsing because string constants do not refer to other symbols for which the `migrate_namespace_lookup` would be needed).
- Modifies the irep2 version to properly take endianness and non-char subtypes into account.
- Restores the regression tests disabled in #905 due to the missing wide string support.

Fixes #1165. I've added the test case from there.


fbrausse:
I should say that this requires endianness to be little or big, the no-endian setting doesn't work for wide strings since we have to interpret the individual elements. Do we have any uses for no-endian? Personally, I would remove this setting from the config (it's not exposed via the cmdline either). What do you think?

*Edit:* Hm, seems that `--no-arch` does set no-endian. Do we have any users of that feature?

fbrausse:
Master:
```
  correct:           14431
    correct true:     8004
    correct false:    6427
  incorrect:            26
    incorrect true:     15
    incorrect false:    11
  unknown:            9348
  Score:             21779 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7268971774
```
This PR:
```
  correct:           14436
    correct true:     8010
    correct false:    6426
  incorrect:            27
    incorrect true:     15
    incorrect false:    12
  unknown:            9342
  Score:             21774 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7268971103
```
There are
- 2x timeout -> incorrect (both are known and already wrong on master)
- 1x incorrect -> timeout (also incorrect with this PR)

Seems, sv-comp isn't testing wide strings.

lucasccordeiro (approved):

lucasccordeiro:

PR #1551, opened 2023-12-20
Basic unicode string tests + small fixes

fbrausse:
- Fixes an oversight in #1549: shift by bytes, not by bits.
- Adds some basic tests for char16_t and char32_t; ~the simplifier is still failing on those~, not all assumptions on chars have been cleaned up, yet.
- The irep1 simplifier now only simplifies index into narrow strings.
- The irep2 simplifier OTOH fully supports index into constant strings now.
- array2string() now only accepts arrays of 8-bit wide subtypes (it's not clearly definable how array2string should behave wrt. the string encodings, etc.) - it's only used for massaging expressions given to the printf_formattert later. That one has a bit inconsistent string handling and I've left it alone; we should start to (again?) use the parse_format_string() there and keep the new "kind" info of strings around.


lucasccordeiro (approved):

lucasccordeiro:

PR #1552, opened 2023-12-20
Python frontend: Class definition and instantiation

brcfarias:
This pull request adds support for class definition and instantiation.
Some relevant details:

- The __init__() method is treated as class constructor.
- In constructor calls like `obj1 = MyClass(..)`, `obj1` is passed as the first argument to the constructor.
- In Python, "Variables defined in the class definition are **class attributes**; they are shared by instances. **Instance attributes** can be set in a method with self.name = value. Both class and instance attributes are accessible through the notation “self.name”, and an instance attribute hides a class attribute with the same name when accessed in this way."(https://docs.python.org/3/reference/compound_stmts.html#class).
- This pull request specifically addresses __instance attributes__. When an attribute is referred by "self" it is added as a field to the corresponding self type. __Class attributes__ will be addressed in a future (perhaps the next) PR.
- The current implementation does not yet consider default inheritance from "object" class. This feature will also be addressed in a subsequent update.

### DIFF
--- a/src/clang-c-frontend/clang_c_adjust_expr.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_expr.cpp
@@ -24,8 +24,8 @@ bool clang_c_adjust::adjust()
   // warning! hash-table iterators are not stable
 
   symbol_listt symbol_list;
-  context.Foreach_operand_in_order(
-    [&symbol_list](symbolt &s) { symbol_list.push_back(&s); });
+  context.Foreach_operand_in_order([&symbol_list](symbolt &s)
+                                   { symbol_list.push_back(&s); });
 
   // Adjust types first, so that symbolic-type resolution always receives
   // fixed up types.
@@ -236,6 +236,8 @@ void clang_c_adjust::adjust_side_effect(side_effect_exprt &expr)
     }
     else if (statement == "nondet")
     {
+      /* Bypassing side effects with `nondet` as a statement since
+       * the goto layer can handle it. */
     }
     else
     {
--- a/src/clang-c-frontend/clang_c_adjust_expr.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_expr.cpp
@@ -24,8 +24,8 @@ bool clang_c_adjust::adjust()
   // warning! hash-table iterators are not stable
 
   symbol_listt symbol_list;
-  context.Foreach_operand_in_order([&symbol_list](symbolt &s)
-                                   { symbol_list.push_back(&s); });
+  context.Foreach_operand_in_order(
+    [&symbol_list](symbolt &s) { symbol_list.push_back(&s); });
 
   // Adjust types first, so that symbolic-type resolution always receives
   // fixed up types.
--- a/src/clang-c-frontend/clang_c_adjust_expr.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_expr.cpp
@@ -234,6 +234,9 @@ void clang_c_adjust::adjust_side_effect(side_effect_exprt &expr)
     else if (statement == "gcc_conditional_expression")
     {
     }
+    else if (statement == "nondet")
+    {
+    }
     else
     {
       log_error("unknown side effect: {} at {}", statement, expr.location());
--- a/src/python-frontend/python_annotation.h
+++ b/src/python-frontend/python_annotation.h
@@ -132,8 +132,17 @@ private:
 
         // lhs
         auto target = element["targets"][0];
-        int col_offset = target["col_offset"].template get<int>() +
-                         target["id"].template get<std::string>().size() + 1;
+        std::string id;
+        if (target.contains("id"))
+          id = target["id"];
+        else if (target["_type"] == "Attribute")
+          id = target["value"]["id"].template get<std::string>() + "." +
+               target["attr"].template get<std::string>(); // e.g. f.blah
+
+        assert(!id.empty());
+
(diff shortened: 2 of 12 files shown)


src/clang-c-frontend/clang_c_adjust_expr.cpp
```diff
@@ -234,6 +234,9 @@ void clang_c_adjust::adjust_side_effect(side_effect_exprt &expr)
     else if (statement == "gcc_conditional_expression")
     {
     }
+    else if (statement == "nondet")
```
lucasccordeiro, line 237: Why do we have this statement?

src/python-frontend/python_annotation.h
```diff
@@ -125,8 +134,17 @@ class python_annotation
 
         // lhs
         auto target = element["targets"][0];
-        int col_offset = target["col_offset"].template get<int>() +
-                         target["id"].template get<std::string>().size() + 1;
+        std::string id;
```
lucasccordeiro, line 136: can you add some comments here?

src/python-frontend/python_annotation.h
```diff
@@ -156,6 +174,26 @@ class python_annotation
         element["value"]["end_col_offset"] =
           element["value"]["end_col_offset"].template get<int>() + type.size() +
           1;
+
+        if (element["value"].contains("args"))
```
lucasccordeiro, line 181: Can you add some comments here?

src/python-frontend/python_converter.cpp
```diff
@@ -201,6 +202,17 @@ void python_converter::adjust_statement_types(exprt &lhs, exprt &rhs) const
   }
 }
 
+std::string python_converter::create_symbol_id() const
+{
+  std::string symbol_id = "py:" + python_filename;
+  if (!current_class_name.empty())
+    symbol_id += "@C@" + current_class_name;
```
lucasccordeiro: The + operator is usually okay for small string operations. However, it might be more efficiently done using std::stringstream or std::string::append to avoid unnecessary copying.

src/python-frontend/python_converter.cpp
```diff
@@ -433,21 +461,51 @@ exprt python_converter::get_expr(const nlohmann::json &element)
...
+      // Get object type name from symbol. e.g.: tag-MyClass
+      std::string obj_type_name;
+      const typet &symbol_type =
+        (symbol->type.is_pointer()) ? symbol->type.subtype() : symbol->type;
+      for (const auto &it : symbol_type.get_named_sub())
+      {
+        if (it.first == "identifier")
+          obj_type_name = it.second.id_string();
+      }
+      assert(!obj_type_name.empty());
```
lucasccordeiro: Shall we really use assert statements here? What happens with the release mode when we disable these assertions?

lucasccordeiro (changes requested):

src/clang-c-frontend/clang_c_adjust_expr.cpp
brcfarias: A `sideeffect` with `nondet` as statement is generated at:
https://github.com/brcfarias/esbmc/blob/69112185c12d40be1eed5d87ea57c4eb6e9e23ac/src/python-frontend/python_converter.cpp#L352C5-L364C6

Due to the addition of the adjust step (see https://github.com/brcfarias/esbmc/blob/69112185c12d40be1eed5d87ea57c4eb6e9e23ac/src/python-frontend/python_language.cpp#L101) to manage pointer to struct usage, it was necessary to add it to avoid the abort at https://github.com/brcfarias/esbmc/blob/69112185c12d40be1eed5d87ea57c4eb6e9e23ac/src/clang-c-frontend/clang_c_adjust_expr.cpp#L243

src/python-frontend/python_converter.cpp
brcfarias: In release we would have a crash when dereferecing class_symbol at https://github.com/brcfarias/esbmc/blob/69112185c12d40be1eed5d87ea57c4eb6e9e23ac/src/python-frontend/python_converter.cpp#L502C1-L503C1.

I will remove the asserts, check if class_symbol is valid before dereferencing, and call `abort()` if so.

lucasccordeiro: Excellent! Thanks, @brcfarias.

src/clang-c-frontend/clang_c_adjust_expr.cpp
lucasccordeiro: Can you please add some comments here?

src/python-frontend/python_converter.cpp
brcfarias: I changed to stringstream. Thanks.

lucasccordeiro (approved):
Thanks for addressing my comments, @brcfarias.

fbrausse (approved):
In general looks like a good approach to me.

Since supporting the full dynamic nature of the language is probably out of scope, I'd suggest to start documenting the particular Python dialect supported by this frontend. Some things like the following reflect existing practice, but are not required by the language:
- name `self` isn't fixed (could also be `me` or any other valid identifier)
- restriction about `int` values we talked about before
- computed instance attibutes not supported (e.g. non-literal access via get/setattr())
- variable declarations need type annotations

lucasccordeiro:

Can you please submit another PR that addresses @fbrausse's suggestion about the documentation?

PR #1560, opened 2023-12-24
Fix overflow- encoding

### LINKED ISSUE #1559: Wrong results when checking sub unsigned underflow
esbmc reports a wrong bug when verifying this program with `--unsigned-overflow-check`  and any solver other than boolector:
```
int main()
{
  unsigned int x = 0;
  unsigned int y = 2;
  unsigned int z = y - x;
}
```

mikhailramalho:
The overflow checks for a - b is now (a - b) <= a

Fixed #1559

### DIFF
--- a/regression/esbmc/github_1559/main.c
+++ b/regression/esbmc/github_1559/main.c
@@ -1,0 +1,6 @@
+int main()
+{
+  unsigned int x = 0;
+  unsigned int y = 2;
+  unsigned int z = y - x;
+}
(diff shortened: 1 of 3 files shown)


mikhailramalho:
Also, `overflow_25_unsigned/main.c` already checks for a failure results, so no need to add another test case for that.

lucasccordeiro (approved):

Shall we run it over the SV-COMP benchmarks?

mikhailramalho:
I don't think it's necessary, there were no wrong results on the overflow
category when I last run the rework-built-in.

I've found this issue by accident when writing one of the regressions

lucasccordeiro:
Thanks, @mikhailramalho.

lucasccordeiro:
@mikhailramalho: Can I ask you whether you have run this PR over the SV-COMP benchmarks?

PR #1564, opened 2023-12-31
Support for context-aware intervals

rafaelsamenezes:
This PR enables the abstract domain to have more information about the context where it is invoked. Specifically, this enables:

1. Global variables
2. Reasoning about functions calls: parameters and return values.
3. ~~Disables the interval analysis unit tests due to #1567~~ (solved by PR #1568)

### DIFF
--- a/regression/esbmc/interval-analysis-function-1/main.c
+++ b/regression/esbmc/interval-analysis-function-1/main.c
@@ -9,6 +9,5 @@ int foo(int n)
 
 int main()
 {
-  foo(51);
-  foo(60);
+  foo(42);
 }
--- a/regression/esbmc/interval-analysis-function-1/main.c
+++ b/regression/esbmc/interval-analysis-function-1/main.c
@@ -1,0 +1,14 @@
+#include <assert.h>
+
+int foo(int n)
+{
+  // n: [51,60]
+  assert(n == 42); // always holds
+}
+
+
+int main()
+{
+  foo(51);
+  foo(60);
+}
--- a/regression/esbmc/interval-analysis-function-1/main.c
+++ b/regression/esbmc/interval-analysis-function-1/main.c
@@ -2,7 +2,7 @@
 
 int foo(int n)
 {
-  // n: [51,60]
+  // n: [42,42]
   assert(n == 42); // always holds
 }
 
--- a/src/goto-programs/abstract-interpretation/interval_domain.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_domain.cpp
@@ -40,11 +40,6 @@ void interval_domaint::update_symbol_interval(
   const symbol2t &sym,
   const integer_intervalt value)
 {
-  // TODO: we can't handle globals
-  // as the analysis is not context-aware
-  //
-  if (has_prefix(sym.thename, "c:@"))
-    return;
   int_map[sym.thename] = value;
 }
 
--- a/src/goto-programs/abstract-interpretation/interval_domain.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_domain.cpp
@@ -720,8 +720,7 @@ expr2tc interval_domaint::make_expression_helper<wrapped_interval>(
     // Interval: [a,b]
     std::vector<expr2tc> disjuncts;
 
-    auto convert = [this, &src, &symbol, &disjuncts](wrapped_interval &w)
-    {
+    auto convert = [this, &src, &symbol, &disjuncts](wrapped_interval &w) {
       assert(*w.lower <= *w.upper);
 
       std::vector<expr2tc> s_conjuncts;
@@ -770,8 +769,7 @@ expr2tc interval_domaint::make_expression_helper(const expr2tc &symbol) const
     return gen_false_expr();
 
   std::vector<expr2tc> conjuncts;
-  auto typecast = [&symbol](expr2tc v)
-  {
+  auto typecast = [&symbol](expr2tc v) {
     c_implicit_typecast(v, symbol->type, *migrate_namespace_lookup);
     return v;
   };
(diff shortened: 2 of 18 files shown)


rafaelsamenezes:
Master
```
  correct:           15016
    correct true:     8444
    correct false:    6572
  incorrect:            50
    incorrect true:     17
    incorrect false:    33
  unknown:            8739
  Score:             22388 (max: 38482)
```
PR
```
  correct:           15127
    correct true:     8555
    correct false:    6572
  incorrect:            49
    incorrect true:     18
    incorrect false:    31
  unknown:            8629
  Score:             22610 (max: 38482)
```

regression/esbmc/interval-analysis-function-1/main.c
```diff
@@ -0,0 +1,13 @@
+#include <assert.h>
+
+int foo(int n)
+{
+  // n: [51,60]
```
fbrausse: This comment is unclear to me: where do the bounds 51 and 60 come from?

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -891,13 +901,30 @@ void interval_domaint::transform(
...
+    /* The current implementation of the abstract interpreter do not store
+     * the return variable (which would be too tricky anyway). We can deal
+     * with this by constructing a tmp symbol in which we can apply assumptions
+     * later */
+    expr2tc return_var = symbol2tc(
+      function.ret_type, fmt::format("c:{}:ret", instruction.function));
+
+    assign(
+      code_assign2tc(return_var, to_code_return2t(instruction.code).operand));
   }
```
fbrausse, line 935: Missing `break`?

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -910,6 +937,63 @@ void interval_domaint::transform(
 
   default:;
```
fbrausse: Maybe remove this default label, so the compiler can help catching missing cases?

fbrausse (commented):
In general LGTM, just a few comments.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -891,13 +901,30 @@ void interval_domaint::transform(
...
+
+    /* The current implementation of the abstract interpreter do not store
+     * the return variable (which would be too tricky anyway). We can deal
+     * with this by constructing a tmp symbol in which we can apply assumptions
+     * later */
+    expr2tc return_var = symbol2tc(
+      function.ret_type, fmt::format("c:{}:ret", instruction.function));
+
+    assign(
+      code_assign2tc(return_var, to_code_return2t(instruction.code).operand));
```
fbrausse, line 933: Is it guaranteed that the operand and return_var exist? What happens with `void` functions that just `return;`?

rafaelsamenezes: 1. The assignment function will select the correct domain based on the type (i.e. bitvectors or floats). An unsupported domain won't affect the abstract state at all.
2. Based on the flow, the return variable should always be initialized before being used (as it requires a function call and a return to have happened before). Nevertheless, if a var was not initialized in the domain, we just assume it to be a top (i.e., -inf to inf).

src/goto-programs/abstract-interpretation/interval_domain.cpp
rafaelsamenezes: yeah, good catch!

regression/esbmc/interval-analysis-function-1/main.c
rafaelsamenezes: ah, it was from another example and I forgot to remove.

src/goto-programs/abstract-interpretation/interval_domain.cpp
rafaelsamenezes: I added the the cases and removed the default. Maybe the try/catch could be the next step!

src/goto-programs/abstract-interpretation/interval_domain.cpp
fbrausse: Not sure I fully understand. Let's say there is a `return;` in a function returning `void`. What would it mean for the operand and the type of `return_var`? (+/-inf) doesn't really make sense for `void`, does it? Maybe I'm missing something.

rafaelsamenezes: > Let's say there is a return; in a function returning void.

Since this is neither a bitvector (for wrapped and integer domains) and neither a float/fixed (for real domain)... we will just ignore it at the assign function.

> (+/-inf) doesn't really make sense for void, does it? 

It doesn't, I meant for what happens if some crazy flow tries to use a variable that was not initialized. This should be impossible though (maybe if someone does some magic if longjmp?).

fbrausse: Makes sense, thanks!

> (maybe if someone does some magic if longjmp?)

Fortunately not implemented, yet. But that's a valid concern we'll have to address when the time comes.

fbrausse (approved):
LGTM!

lucasccordeiro (approved):

lucasccordeiro:
@rafaelsamenezes: could you please fix the CI issues here?

rafaelsamenezes:
I changed the failing test case. It seems that the issue is related to using a function pointer. The function ptr is making the abstract interpreter confused about where is the next instruction, resulting in a bottom.

I also added the original test case as a knownbug. It seems to be related to having the function ptr call at the end of a block.

Issue #1565, opened 2023-12-31
Interval analysis gives unsound saftey proof


Novak756:
Hi, for the following program
```c
extern void __VERIFIER_error(void);

extern long __VERIFIER_nondet_long(void);
extern _Bool __VERIFIER_nondet_bool(void);

int main(){
        _Bool b1 = __VERIFIER_nondet_bool();
        _Bool b2 = __VERIFIER_nondet_bool();
        long l1 = __VERIFIER_nondet_long();
        long l2 = __VERIFIER_nondet_long();
        if(b2){
                if(!(((b1) || ((l2) == (0))) && (!(b2)))){
                        if(!(((b2) || ((l1) == (0))) && (!(b1)))){
                        __VERIFIER_error();
                        }
                }
        }
    return 0;
}
```
running esbmc with `--interval-analysis` enabled returns `VERIFICATION_SUCCESSFUL`.
Running without flags gives the correct result (`VERIFICATION_FAILED`).

Output with `--interval-analysis --interval-analysis-dump`: [res.txt](https://github.com/esbmc/esbmc/files/13801107/res.txt)
OS: Docker Ubuntu 20.04 on WSL 2 Ubuntu
Version: Commit 9b63b634905094e17495bebf5a1fd4269a376380

lucasccordeiro:
@rafaelsamenezes: could you please check this issue?

rafaelsamenezes:
I'm still investigating, but some notes:

- The intervals seems correct, the only interval computed is `1 <= b2` after the first condition.
- The issue seems to be in the boolean interval propagation.

```
IF !b2 THEN GOTO 1
IF !(!(((_Bool)((signed int)b1) || l2 == 0) && !b2)) THEN GOTO 1
ASSERT ((_Bool)((signed int)b2) || l1 == 0) && !b1
```

Is being converted to 

```
IF !b2 THEN GOTO 1
IF !((signed int)b1 == 0) || l2 == 0 THEN GOTO 1
ASSERT !b1
```

In the assertion, `l1 == 0` is being wrongly removed.

rafaelsamenezes:
Apparently the issue is related to typecasting with Integer intervals. Changing to the wrapped domains (with an extra patch) changes the code to

```
IF !b2 THEN GOTO 1
ASSERT !b1
```

The middle IF was removed.

lucasccordeiro:
Closed via #1625.

lucasccordeiro:
[...]

PR #1566, opened 2024-01-01
Wrapped intervals can now optimize expressions

rafaelsamenezes:
Related to #1565 

This PR enables using the singleton and boolean optimization for wrapped intervals.

### DIFF
--- a/src/goto-programs/abstract-interpretation/wrapped_interval.h
+++ b/src/goto-programs/abstract-interpretation/wrapped_interval.h
@@ -1428,7 +1428,7 @@ private:
   static BigInt compute_upper_bound(const type2tc &t)
   {
     if (is_pointer_type(t))
-      return BigInt::power2(32);
+      return BigInt::power2(config.ansi_c.address_width);
 
     return BigInt::power2(t->get_width());
   }
--- a/src/goto-programs/abstract-interpretation/wrapped_interval.h
+++ b/src/goto-programs/abstract-interpretation/wrapped_interval.h
@@ -1429,7 +1429,7 @@ private:
   {
     if (is_pointer_type(t))
       return BigInt::power2(32);
-    
+
     return BigInt::power2(t->get_width());
   }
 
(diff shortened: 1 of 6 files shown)


lucasccordeiro (changes requested):
@rafaelsamenezes: could you please add some regression tests?

src/goto-programs/abstract-interpretation/wrapped_interval.h
```diff
@@ -1425,6 +1427,9 @@ class wrapped_interval : public interval_templatet<BigInt>
 private:
   static BigInt compute_upper_bound(const type2tc &t)
   {
+    if (is_pointer_type(t))
+      return BigInt::power2(32);
```
fbrausse: The constant 32 here looks fishy. Consider replacing it with `pointer_width()` or `address_width` from config.ansi_c - depending on the meaning of "upper bound" here: what's the order relation considered for wrapped intervals on pointers?

fbrausse (commented):
Besides the comment below, looks fine to me.

I didn't immediately get what the implementation of the new optimize_expr_interval() was doing: It basically just does singleton propagation, but the code is obscured a bit by the early case split based on the expr's type. I'd imagine it to be easier to follow when looking at .singleton() first, but that's just a nitpick.

src/goto-programs/abstract-interpretation/wrapped_interval.h
rafaelsamenezes: > The constant 32 here looks fishy. 

You are right, 32 should be changed.

> what's the order relation considered for wrapped intervals on pointers

At this point it would not be the pointer, just its address value (guess this answer it) so `if(ptr != NULL)` or `((size_t)ptr == 0x42)` could be eventually processed. 

We are still not tracking any pointers though, the change was needed because the wrapped domain always needs an "upper bound".

fbrausse (approved):

lucasccordeiro (changes requested):
shall we enable `interval-analysis-wrapped` by default?

rafaelsamenezes:
> shall we enable `interval-analysis-wrapped` by default?

Not yet. Still have some experiments to do.

lucasccordeiro (approved):

PR #1568, opened 2024-01-02
`migrate_namespace_lookup` now managed by language_uit

### LINKED ISSUE #1567: Goto generation in unit test is not initializing the namespace correctly
This causes migration segmentation fault errors at the unit tests

fbrausse:
This PR moves the initialization and management of the global `migrate_namespace_lookup` from the various entry routines to the `language_uit` class, which also holds the context that ends up being used throughout goto and symex. This change should fix #1567 by keeping the context alive after goto_factory::get_goto_functions() ends.

For this to work it was necessary to introduce move-constructors for `contextt` and `language_uit`. The copy/move ctors weren't being used before, so this change just adds a way to move objects of those two types. Each has different reasons why copy construction isn't supported.

### DIFF
--- a/src/c2goto/c2goto.cpp
+++ b/src/c2goto/c2goto.cpp
@@ -62,7 +62,7 @@ class c2goto_parseopt : public parseoptions_baset, public language_uit
 {
 public:
   c2goto_parseopt(int argc, const char **argv)
-    : parseoptions_baset(c2goto_options, argc, argv), language_uit(cmdline)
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

rafaelsamenezes (approved):
Thanks @fbrausse , I was not sure how to start fixing this!

I also confirmed that this solves #1567.

PR #1572, opened 2024-01-03
Fix gcc-13 warnings

fbrausse:
Even if it's not spring, yet, this PR does a bit of spring-cleaning:

Fixes a bunch of compiler warnings new in GCC-13 (compared to v12):
- unnecessary std::move in return statements
- Jimple frontend: suspicious handling of error cases + fmt-lib usage
- update fmt-lib to v10.2.0 due to use of possibly dangling references (for expr2t parameters) fixed in v10.1.1

Besides that:
- reset c2goto verbosity to Result for a cleaner build log
- setjmp() OM: return nondet value to avoid Clang warning
- change esbmc_version_string to function and move it (and fixed dependencies) out of globals.cpp

### DIFF
--- a/scripts/cmake/ExternalDependencies.cmake
+++ b/scripts/cmake/ExternalDependencies.cmake
@@ -5,7 +5,7 @@ include(FetchContent)
 # FMT
 FetchContent_Declare(fmt
   GIT_REPOSITORY https://github.com/fmtlib/fmt.git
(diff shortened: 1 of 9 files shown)


rafaelsamenezes (approved):
LGTM!

lucasccordeiro (approved):

PR #1574, opened 2024-01-03
Added cstdint include for globals.cpp

rafaelsamenezes:
Building ESBMC in Fedora Linux causes the error:

```
/home/esbmc/repos/esbmc/src/esbmc/globals.cpp:23:18: error: unknown type name 'uint8_t'
   23 | extern "C" const uint8_t buildidstring_buf[];
```

This PR fixes it.

### DIFF
--- a/src/esbmc/globals.cpp
+++ b/src/esbmc/globals.cpp
@@ -1,4 +1,5 @@
 #include <langapi/mode.h>
+#include <cstdint>
 
 const mode_table_et mode_table[] = {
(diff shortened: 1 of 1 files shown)


fbrausse:
This is also fixed by #1572.

rafaelsamenezes:
> This is also fixed by https://github.com/esbmc/esbmc/pull/1572.

Indeed, just checked with your patch!

Close by #1572

PR #1587, opened 2024-01-09
Python frontend: Class attributes

brcfarias:
This PR adds support for **class attributes** as described by Python documentation (https://docs.python.org/3/reference/compound_stmts.html#class-definitions):

"_Variables defined in the class definition are class attributes; they are shared by instances. Instance attributes can be set in a method with self.name = value. Both class and instance attributes are accessible through the notation “self.name”, and an instance attribute hides a class attribute with the same name when accessed in this way._"

Class attributes are modeled as static class members in C++. Thus, when a class attribute is identified, a symbol with an id in the format _filename@C@classname@varname_ is inserted into the context.

### DIFF
--- a/src/python-frontend/python_converter.h
+++ b/src/python-frontend/python_converter.h
@@ -18,7 +18,7 @@ private:
   void
   get_return_statements(const nlohmann::json &ast_node, codet &target_block);
   void get_function_definition(const nlohmann::json &function_node);
-  void get_class_definition(const nlohmann::json &class_node);
+  void get_class_definition(const nlohmann::json &class_node, codet &target_block);
 
   locationt get_location_from_decl(const nlohmann::json &ast_node);
   exprt get_expr(const nlohmann::json &element);
--- a/src/python-frontend/python_converter.h
+++ b/src/python-frontend/python_converter.h
@@ -3,6 +3,8 @@
 #include <util/context.h>
 #include <nlohmann/json.hpp>
 
+#include <map>
+
 class codet;
 class struct_typet;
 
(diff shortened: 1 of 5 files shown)


brcfarias:
@rafaelsamenezes It looks like the CI failed because the source branch is only in my project fork.

fbrausse (changes requested):
Instead of modifying the existing test case, could you please add a new one testing the new support for class attributes?

> @rafaelsamenezes It looks like the CI failed because the source branch is only in my project fork.

While that's true and we should allow PRs from forks (because it's standard procedure on Github), it seems that there are formatting changes. @brcfarias Could you double-check that everything touched here conforms to clang-format-11?

lucasccordeiro:
@brcfarias: I agree with @fbrausse's comments. Could you please address them?

src/python-frontend/python_converter.h
```diff
@@ -47,4 +49,8 @@ class python_converter
   std::string current_func_name;
   std::string current_class_name;
   exprt *ref_instance;
+  bool is_converting_lhs = false;
+
+  // Map object to list of instance attributes
+  std::map<std::string, std::vector<std::string>> instance_attr_map;
```
kunjsong01: perhaps using unordered_map for better time complexity if the order doesn't matter?

kunjsong01 (changes requested):
just a nitpick for minor performance improvement...

brcfarias:
@fbrausse Thanks for the comments.

> Instead of modifying the existing test case, could you please add a new one testing the new support for class attributes?

I just updated the PR with a new regression test for class attributes.

> Could you double-check that everything touched here conforms to clang-format-11?

I was using clang-format-9 :grinning:

src/python-frontend/python_converter.h
brcfarias: Thanks @kunjsong01. Just changed the type to unordered_map.

lucasccordeiro (approved):

PR #1596, opened 2024-01-16
Fix range of rand() and random() to include upper endpoint

### LINKED ISSUE #1592: Wrong ranges for rand() and random()
Prompted by https://github.com/esbmc/esbmc/issues/1591#issuecomment-1890477240 I did take another look at our models for rand() and random().

rand() is supposed to return a value between 0 and RAND_MAX, both inclusively. Our implementation doesn't allow for the value RAND_MAX.

random() is supposed to return a value between 0 and $2^{31}-1$, both inclusively. Our model also doesn't allow for the upper bound to be returned.

fbrausse:
This PR closes #1592.


lucasccordeiro (approved):

lucasccordeiro:

PR #1601, opened 2024-01-18
Add libm musl float OMs used in svcomp24

fbrausse:


lucasccordeiro (approved):

Shall we consider some regression tests to be part of this PR?

fbrausse:
> Shall we consider some regression tests to be part of this PR?

I will add some of the faster NN benchmarks as regression tests.

lucasccordeiro (approved):

PR #1602, opened 2024-01-19
Python frontend: Support for pass statement and var declarations without initialization

brcfarias:
https://docs.python.org/3/tutorial/controlflow.html:

_The [pass](https://docs.python.org/3/reference/simple_stmts.html#pass) statement does nothing. It can be used when a statement is required syntactically but the program requires no action. For example:_

```
while True:
    pass  # Busy-wait for keyboard interrupt (Ctrl+C)
```

_This is commonly used for creating minimal classes:_
```
class MyEmptyClass:
    pass
```


rafaelsamenezes (commented):
Does this support functions that return values? I would think that it should return some form of nondet value if it is the case.

brcfarias:
> Does this support functions that return values? I would think that it should return some form of nondet value if it is the case.

Hi @rafaelsamenezes. Yes, the verification is successfull for this program:

```
def foo() -> float:
    pass

foo()
```

brcfarias:
The output for:

```
def foo() -> float:
    pass

print(foo())
```

is `None`

brcfarias:
However we are crashing for the following valid program:

```
def foo() -> float:
    pass

assert (foo() == None)
```

> I would think that it should return some form of nondet value if it is the case.

From the tests, this wouldn't be the case, since nondet() would return a value of the same type as the function's return type instead of None.

This is a case that I can handle on the next PR.

lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #1607, opened 2024-01-20
VSA is now only computed with GCSE flag

rafaelsamenezes:
This enables VSA only when gcse is enabled. Mainly to prevent issues since VSA is not very stable


fbrausse (approved):
LGTM.

lucasccordeiro (approved):

PR #1608, opened 2024-01-21
Python frontend: Empty class instantiation and built-in functions support

brcfarias:
This PR adds support for instantiating empty classes as below:

```
class EmptyClass:
    pass;

obj = EmptyClass()
```

In addition, it handles calls to the built-in functions [int()](https://docs.python.org/3/library/functions.html#int) and [bool()](https://docs.python.org/3/library/functions.html#bool), addressing cases such as:

```
x = int(1)
y = bool(true)
```

### DIFF
--- a/regression/python/built-in-functions/main.py
+++ b/regression/python/built-in-functions/main.py
@@ -1,0 +1,5 @@
+x = int(1)
+assert(x == 1)
+
+y = bool(True)
+assert(y == True)
--- a/regression/python/built-in-functions/test.desc
+++ b/regression/python/built-in-functions/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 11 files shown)


lucasccordeiro (approved):
LGTM.

lucasccordeiro (commented):
@brcfarias: could you please add this test case to our regression?

class EmptyClass:
    pass;

obj = EmptyClass()

brcfarias:
> @brcfarias: could you please add this test case to our regression?
> 
> class EmptyClass: pass;
> 
> obj = EmptyClass()

@lucasccordeiro I modified regression/python/classes/main.py to include this case.

lucasccordeiro:
Thank you.

fbrausse (commented):
I'd still suggest to keep the regression test cases small and focussed on testing a single thing, where possible. That way it is easier to understand what the bug actually is once the test case fails. It's fine to add a one-line test case.

lucasccordeiro:
@brcfarias: could you please implement @fbrausse's suggestion?

brcfarias:
@fbrausse, @lucasccordeiro I agree. However, in this case, I found it more coherent to move this part to the regression/pass/main.py regression test.

PR #1612, opened 2024-01-21
Remove shared_ptr from being used unnecessarily in API

fbrausse:
As @brcfarias mentioned in private conversation, shared_ptr (and unique_ptr as well, to some extent) as function parameters is quite uncommon and has some risks, in particular when passed by-reference.

The bmct class (and a few related APIs) were doing that for the `runtime_solver`, the `symex` instance, `symex_target_equationt`, and `symex_resultt`. This PR removes shared_ptr as parameter to functions in those APIs by instead passing the objects via reference. It also adds more `const` in order to communicate that an object is just read. The hope is that the code is a bit clearer to follow.

The first two, `runtime_solver` and `symex` are now held by unique-ptrs instead of shared ones, because they are not actually being shared in the sense that a copyable object needs to store them as members.

### DIFF
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -99,9 +99,7 @@ void bmct::successful_trace()
   }
 }
 
-void bmct::error_trace(
-  std::shared_ptr<smt_convt> &smt_conv,
-  std::shared_ptr<symex_target_equationt> &eq)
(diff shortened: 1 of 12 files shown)


rafaelsamenezes (approved):
LGTM.

lucasccordeiro (approved):

PR #935, opened 2023-03-16
Adding support for cvc5

ibnyusuf:
Upgrading CVC4 support to CVC5. Whilst the APIs are not compatible, there are relatively similar, so the process wasn't too difficult. This PR is **not** ready to merge. The following actions are required:

- [ ] Add instructions for bulding cvc5 on Linux
- [ ] Modify CMakeLists.txt in solvers/cvc5/ directory such that path to dynamically linked library `libcvc5.1.dylib` does not need to be provided explicitly when running ESBMC. Currently, I run with command[1]
- [ ] In certain places in cvc_conv.cpp I was unable to update to new API due to a lack of familiarity with what ESBMC functions are supposed to achive. All such places are commented with `// AYB... `

[1]`DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:/Users/user/ESBMC/CVC5/lib/ ../release/bin/esbmc --k-induction --max-k-step 2 ../../esbmc_benchmarks/k_induction_paper_ex.c --cvc`

PR #1613, opened 2024-01-21
SV-Comp builds should include all supported solvers

rafaelsamenezes:
This PR will enable the build of every solver for our SV-Comp testing builds. It also adds support for MathSAT.

I will leave CVC4 disabled as [the build is outdated](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=266721#c3) and we should move to CVC5: https://github.com/esbmc/esbmc/pull/935

Run example: https://github.com/esbmc/esbmc/actions/runs/7602481313

### DIFF
--- a/.github/workflows/benchexec.yml
+++ b/.github/workflows/benchexec.yml
@@ -96,7 +96,7 @@ jobs:
     uses: ./.github/workflows/build.yml
     with:
       operating-system: ubuntu-latest
-      build-flags: '-b Release -e Off'
+      build-flags: '-b Release -e Off -C'
       testing: false     
 
   run-benchexec:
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):
Nice!

fbrausse (approved):
LGTM. I wonder though whether we could drop the check whether $HOME/$solver exists in the future. It's quite uncommon for software builds to inspect the user's home directory this way.

rafaelsamenezes:
> LGTM. I wonder though whether we could drop the check whether $HOME/$solver exists in the future. It's quite uncommon for software builds to inspect the user's home directory this way.

I agree, it's there more for historical reasons when we moved from autoconf

PR #1614, opened 2024-01-21
[clang-cpp-frontend] Fix the copy and move constructor

XLiZHI:
This PR fixes the incorrect function call and adds the rvalue reference test case.

example:
```
struct MyStruct {
    int value;
};

int main() {
    MyStruct a = {10};

    MyStruct c(a);
}
```
`esbmc main.cpp --goto-functions-only`

result:
9_R5(9DG8%D34T{T7HZU2KG


lucasccordeiro (approved):
Nice 😊

lucasccordeiro:

Issue #1556, opened 2023-12-23
[Python] Shall we really invoke the SMT solver for this simple Python program?


lucasccordeiro:
I would expect our constant propagation algorithm to solve this case easily without calling the SMT solver.

````Python
class MyClass:
    def __init__(self, value: int):
        self.data:int = value


obj1 = MyClass(5)
assert(obj1.data == 5)
obj1.data = 10
assert(obj1.data == 10)
obj1.data = 20
assert(obj1.data == 20)

obj2 = MyClass(30)
assert(obj2.data == 30)
obj2.data = 40
assert(obj2.data == 40)
````

````
$ esbmc main.py
ESBMC version 7.4.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing main.py
Converting
Generating GOTO Program
GOTO program creation time: 0.039s
GOTO program processing time: 0.000s
Starting Bounded Model Checking
Symex completed in: 0.000s (14 assignments)
Slicing time: 0.000s (removed 4 assignments)
Generated 9 VCC(s), 5 remaining after simplification (10 assignments)
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
Encoding to solver time: 0.000s
Solving with solver Boolector 3.2.2
Runtime decision procedure: 0.000s
BMC program time: 0.001s

VERIFICATION SUCCESSFUL
````

lucasccordeiro:
It looks like we don't propagate the store operations involving constants:

````
Program constraints:
// 0 file main.py line 3 column 8 function MyClass
/* 1 */ obj1&0#1 == (obj1&0#0 WITH [data:=5])

// 3  (assertion)
/* 2 */ (assert)execution_statet::\guard_exec?0!0 => obj1&0#1.data == 5

// 4 file main.py line 8 column 0
/* 3 */ obj1&0#2 == (obj1&0#1 WITH [data:=10])

// 5  (assertion)
/* 4 */ (assert)execution_statet::\guard_exec?0!0 => obj1&0#2.data == 10

// 6 file main.py line 10 column 0
/* 5 */ obj1&0#3 == (obj1&0#2 WITH [data:=20])

// 7  (assertion)
/* 6 */ (assert)execution_statet::\guard_exec?0!0 => obj1&0#3.data == 20

// 0 file main.py line 3 column 8 function MyClass
[...]

PR #1617, opened 2024-01-22
Slicer support for Arrays/Structs

rafaelsamenezes:
This PR intends to expand our slicer to work with fields and structs by replacing the WITH operations with just the plain symbol. SMT solvers have big trouble dealing with all the updates. For example, the following program:

```c
#define N 50
int main()
{  
  int arr[N];
  for (unsigned long i = 0; i < N; i++)
    arr[i] = (int)i;

  __ESBMC_assert(arr[42] == 42, "");
}
```

Will generate:
```
Thread 0 file test.c line 9 column 3 function main
ASSIGNMENT ()
arr?1!0&0#44 == (arr?1!0&0#43 WITH [42:=42])
... 
Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
arr?1!0&0#51 == (arr?1!0&0#50 WITH [49:=49])

Thread 0 file test.c line 9 column 3 function main
ASSERT
execution_statet::\guard_exec?0!0 => arr?1!0&0#51[42] == 42
assertion
```

The increasing number of WITH takes progressively larger solving time. The slicer could replace the with id operations: 
```
Thread 0 file test.c line 9 column 3 function main
ASSIGNMENT ()
arr?1!0&0#44 == (arr?1!0&0#43 WITH [42:=42])
... 
Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
arr?1!0&0#51 == (arr?1!0&0#50

Thread 0 file test.c line 9 column 3 function main
ASSERT
execution_statet::\guard_exec?0!0 => arr?1!0&0#51[42] == 42
assertion
```

For the program above, changing the N resulted in the following Runtime decision procedure times with Z3 v4.12.4:

| N   | Master | Array Slicing |
| ---- | -----------| ------------- |
| 50 | 0.00s | 0.00s |
| 500 | 0.03s | 0.001s |
| 5 000 | 0.141s | 0.004s |
| 50 000 | 18.807s | 0.051s |
| 500 000 | +300s | 5.506s |

The implementation still requires a bit of testing, specially when an assertion depends on a symbolic index. Also, it would be great if we had some way of recursively replacing the equivalences. I think this could be done in another pass though.

### DIFF
--- a/src/goto-symex/slice.cpp
+++ b/src/goto-symex/slice.cpp
@@ -46,11 +46,10 @@ bool symex_slicet::get_array_symbols(const expr2tc &expr)
   if (is_index2t(expr))
   {
     const index2t &w = to_index2t(expr);
-    if (is_symbol2t(w.source_value) && is_constant_int2t(w.index))
+    if (is_symbol2t(w.source_value))
     {
       const symbol2t &s = to_symbol2t(w.source_value);
-      const constant_int2t &i = to_constant_int2t(w.index);
-      array_depends[s.thename.as_string()].emplace(i.value.to_uint64());
+      array_depends[s.thename.as_string()].emplace(w.index);
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


fbrausse (commented):
I'm not sure this is sound. You're using the unadorned symbol name as key into the map. The array symbol could be touched in other ways (e.g. by non-constant index), in which case this change would overwrite that, wouldn't it?

rafaelsamenezes:
> I'm not sure this is sound. You're using the unardorned symbol name as key into the map. The array symbol could be touched (e.g. by non-constant index), in which case this change would overwrite that, wouldn't it?

Hm, depends. Right now the pre-requisite is that the index operation in the assertion contains a constant for the index. Similarly, for the WITH operation that its going to be replaced, it needs to be a constant index. Which means that

`int n = 1; arr[0] = 1; arr[n] = 2; assert(arr[n] == 2)` would not slice anything at all.

The main limitation is... if we have an assert that mix! I guess having a list of array that depends on a symbol should suffice.

lucasccordeiro (changes requested):
@rafaelsamenezes: can you please add some regression tests with the corner cases discussed in this issue?

lucasccordeiro:
@Anthonysdu: can you please evaluate this PR over the RMM project?

lucasccordeiro:
@rafaelsamenezes: we could also move forward with this PR: https://github.com/esbmc/esbmc/pull/1398. 

@Anthonysdu started working on it, but there was not much progress for some reason.

lucasccordeiro:
https://github.com/esbmc/esbmc/issues/1556 is also related to this PR.

rafaelsamenezes:
> @rafaelsamenezes: we could also move forward with this PR: #1398.
> 
> @Anthonysdu started working on it, but there was not much progress for some reason.

Hm, I think that the main issue is that this could increase the Symex time by a lot (since we need to compute huge arrays every time).  This is fine for reasonably sized arrays, but computing an array of 50k positions everytime might be not worth (specially when just one index is needed).

The slicer right now is trying to take advantage as a "lazy-evaluation". That being said, it would be great to do the WITH propagation on the sliced SSA!

mikhailramalho:

PR #1623, opened 2024-01-24
[witness] fixed overflow in ID node

### LINKED ISSUE #1621: Duplicate node IDs in GraphML violation witness
The violation witnesses in GraphML format generated by ESBMC may contain duplicate node IDs.

Here's an example [witness](https://github.com/esbmc/esbmc/files/14028514/witness.graphml.zip) obtained by verifying [`szymanski.4.prop1-func-interl.c`](https://gitlab.com/sosy-lab/research/data/word-level-hwmc-benchmarks/-/blob/02bb055de5016fac23431567e23b87e1ab646937/bv/c-lazyMod/beem/szymanski.4.prop1-func-interl.c) with property [`unreach-call.prp`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/properties/unreach-call.prp)
using ESBMC 7.4.0 (downloaded from the SV-COMP 2024 [archive](https://doi.org/10.5281/zenodo.10198805)).
It can be reproduced with the following command:

```bash
./esbmc-wrapper.py -p unreach-call.prp -s incr --arch 64 szymanski.4.prop1-func-interl.c
```

The witness contains 3 nodes with ID "N0" at lines 50, 393262, and 786478.
This is likely caused by an integer overflow, as shown by the following snippet (line 196648 of the attached witness).

```xml
    <node id="N32767"/>
    <edge id="E32766" source="N32766" target="N32767">
      <data key="startline">4317</data>
      <data key="assumption">var_1456_arg_0 = 0;</data>
      <data key="threadId">0</data>
    </edge>
    <node id="N-32768"/>
    <edge id="E32767" source="N32767" target="N-32768">
      <data key="startline">4318</data>
      <data key="assumption">var_1456_arg_1 = 1;</data>
      <data key="threadId">0</data>
    </edge>
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1621.

Here is the command line used to produce the witness:

````
time esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check szymanski.4.prop1-func-interl.c --64 --witness-output witness.graphml --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
````

The newly generated witness is here: [witness.zip](https://github.com/esbmc/esbmc/files/14039888/witness.zip)

### DIFF
--- a/src/goto-symex/witnesses.h
+++ b/src/goto-symex/witnesses.h
@@ -5,6 +5,7 @@
 #include <goto-symex/goto_trace.h>
 #include <string>
 #include <regex>
+#include <big-int/bigint.hh>
 
 typedef boost::property_tree::ptree xmlnodet;
 
@@ -13,7 +14,7 @@ typedef boost::property_tree::ptree xmlnodet;
 class nodet
 {
 private:
-  static unsigned int _id;
+  static BigInt _id;
 
 public:
   std::string id;
@@ -25,15 +26,15 @@ public:
   std::string invariant_scope;
   nodet(void)
   {
-    id = "N" + std::to_string(_id);
-    _id++;
+    id = "N" + integer2string(_id);
+    _id += 1;
   }
 };
 
 class edget
 {
 private:
-  static unsigned int _id;
+  static BigInt _id;
 
 public:
   std::string id;
@@ -54,15 +55,15 @@ public:
   nodet *to_node;
   edget(void)
   {
-    id = "E" + std::to_string(_id);
-    _id++;
+    id = "E" + integer2string(_id);
+    _id += 1;
     from_node = NULL;
     to_node = NULL;
   }
   edget(nodet *from_node, nodet *to_node)
   {
-    id = "E" + std::to_string(_id);
-    _id++;
+    id = "E" + integer2string(_id);
+    _id += 1;
     this->from_node = from_node;
     this->to_node = to_node;
   }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by ChenfengWei0 — APPROVED** [review]

File: src/goto-symex/witnesses.h [hunk 1]
```diff
@@ -33,7 +33,7 @@ class nodet
 class edget
 {
 private:
-  static short int _id;
+  static unsigned int _id;
```
mikhailramalho: why is this static?

mikhailramalho (changes requested):
We should use BigInt instead, to prevent this problem from ever happening.

src/goto-symex/witnesses.h
ChenfengWei0: I think it is because we might call `new nodet()` and `new edget()` multiple times in `witnesses.cpp`. We want only one copy of the static member shared by all objects, thereby accumulating the counter `_id += 1` in the constructor.

ChenfengWei0:
> We should use BigInt instead, to prevent this problem from ever happening.

Changed to BigInt as suggested

mikhailramalho (approved):
lgtm

fbrausse:
> We should use BigInt instead, to prevent this problem from ever happening.

I can't think of a way that `size_t` would ever overflow - there cannot be more objects in memory than it can count to. But I'm fine with BigInt in this rather non-critical path as well.

PR #1625, opened 2024-01-26
Interval analysis now inverts boolean correctly

rafaelsamenezes:
Solves #1565

The original issue was that the `invert_bool` function expected an interval to be a singleton i.e., [0,0] or [1,1]. If the interval was not that, it would assume that it was (-inf,+inf). This resulted in invert bool for [1,+inf) return [1, +inf) which is wrong.

### DIFF
--- a/regression/esbmc/github_1565-2/test.desc
+++ b/regression/esbmc/github_1565-2/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --interval-analysis
 ^VERIFICATION FAILED$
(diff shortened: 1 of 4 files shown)


lucasccordeiro (approved):
LGTM.

fbrausse (approved):
Yes. This looks right to me.

lucasccordeiro:

rafaelsamenezes:
Master (120s):

```
  correct:           15105
    correct true:     8545
    correct false:    6560
  incorrect:            45
    incorrect true:     17
    incorrect false:    28
  unknown:            8655
  Score:             22658 (max: 38482
```

This PR (120s):

```
  correct:           15107
    correct true:     8542
    correct false:    6565
  incorrect:            45
    incorrect true:     17
    incorrect false:    28
  unknown:            8653
  Score:             22657 (max: 38482)
```

PR #1627, opened 2024-01-27
Protect make_n_ary() from being called with empty list

### LINKED ISSUE #1626: Assertion `!v.empty()' failed at smt_conv.h:216
Command-line: 

$esbmc file.c

File:

````C
#include <stdio.h>
#include <stdlib.h>

typedef struct {
} Foo;

void Foo_Execute(Foo *self) {
    Foo *ptr = self;
    printf("pointer value: %s\n", *ptr);
}
int main() {
    Foo *foo = (Foo *)malloc(sizeof(Foo));
    if (foo == NULL) {
        fprintf(stderr, "Memory allocation failed\n");
        return -1;
    }
    free(foo);
    Foo_Execute(foo);
    return 0;
}
````

Output:

````
esbmc: /home/lucas/ESBMC_Project/esbmc/src/solvers/smt/smt_conv.h:216: const smt_ast* smt_convt::make_n_ary(Object, Method, const ast_vec&) [with Object = smt_convt*; Method = const smt_ast* (smt_convt::*)(const smt_ast*, const smt_ast*); smt_astt = const smt_ast*; smt_convt::ast_vec = std::vector<const smt_ast*>]: Assertion `!v.empty()' failed.
Aborted
````

fbrausse:
Addresses #1626. Basically, smt-conv's make_n_ary() doesn't support an empty argument list. ~I'll go through the rest of the node-flattener checking that other invocations of that function are properly guarded.~

This PR:
- replaces the public make_n_ary() API with separate functions for conjunction/disjunction that handle an empty list of arguments, and
- adds TCs to test both, the default, and the tuple-sym-flattener (with which master aborted for this example as well).
- adds a KNOWNBUG case, which most likely is due to #1628.


rafaelsamenezes (approved):
LGTM!

lucasccordeiro (approved):

PR #1593, opened 2024-01-14
Pass alignment attribute constraint through to smt

fbrausse:
This PR fixes #1590, ~but it doesn't solve it, fully, yet~. For ~static storage~ variables the `__aligned__` attribute is honoured now, ~but for automatic storage duration it is not, yet. The reason for not supporting automatic storage is most likely the missing de-renumbering that this PR's SMT layer change doesn't do (ignore the double-negation, I'm not a native speaker ;) )~.

PR #1629, opened 2024-01-28
Value-set takes alignment of references to symbols into account

### LINKED ISSUE #1622: ESBMC unable to reason about symbolic indexes properly
In the following program:

```c
#include <stdio.h>
#include <assert.h>
#define UINT8  unsigned char
#define UINT16 unsigned short int
#define UINT32 unsigned long
#define UINT64 unsigned long long

typedef struct {
	UINT32                       Offset;
	UINT32                       Width;
	UINT64                       Value;
	char* Name;
	UINT64                       Reserved;
} REGMAP_ENTRY; 

typedef struct {
    REGMAP_ENTRY* RegMap;
} REGMAP_TABLE; 

REGMAP_ENTRY m_PORT_RegMap[] = {
  {0x0,   32, 0x00000800, "REG0", 0xFFFF0000},
  {0x0,   32, 0x00000800, "REG1", 0xFFFF0000},
  {0x0,   32, 0x00000800, "REG2", 0xFFFF0000},
  {0x0,   32, 0x00000800, "REG3", 0xFFFF0000}
};

REGMAP_TABLE m_PORT_RegMapTable = {
  m_PORT_RegMap
};

typedef struct {
    void *       RegCntxtPortList1[4];
} IP_INST;

int main()
{
    IP_INST* pInst;
    pInst = malloc(sizeof(IP_INST));
    
    // SOMETHING WEIRD HERE
    int PortIdx = 0;    
    pInst->RegCntxtPortList1[PortIdx] = &m_PORT_RegMapTable;
    REGMAP_TABLE* RegMapTable = pInst->RegCntxtPortList1[PortIdx];
    REGMAP_ENTRY* pRegmap = &(RegMapTable->RegMap[3]);
    assert(pRegmap != 0);
    free(pInst);
    return 0;
}

int main_fixed()
{
    IP_INST* pInst;
    pInst = malloc(sizeof(IP_INST));
    
    // No PortIdx
    pInst->RegCntxtPortList1[0] = &m_PORT_RegMapTable;
    REGMAP_TABLE* RegMapTable = pInst->RegCntxtPortList1[0];
    REGMAP_ENTRY* pRegmap = &(RegMapTable->RegMap[3]);
    assert(pRegmap != 0);
    free(pInst);
    return 0;
}
```

In the function `main` the variable `PortIdx` is used to get a position of the array. Running `esbmc --force-malloc-success` will result in the assertion failing (the assertion passes when running with GCC). However, replacing the variable with its value (at `main_fixed`) solves the issue. This may be related to: https://github.com/esbmc/esbmc/issues/1458#issuecomment-1806090365

fbrausse:
When creating references in the VSA, expressions referring to symbols were treated as being aligned to just 1 byte. At least since #1593, but maybe also before, symbols' types can carry alignment info. This PR makes VSA take this alignment info into account.

Fixes #1622 by addressing problem 1 identified in <https://github.com/esbmc/esbmc/issues/1622#issuecomment-1911845863>. The other problem still exists.

With just that change, these tests started failing:
- regression/esbmc-unix2/01_pthread52, 53 and 56
- regression/esbmc/gcse
- regression/esbmc/github_1091

The pthread ones are fixed now in this PR as well:
- make helper functions to TLS variables assume being called in atomic context (which they are in)
- call destructor of TLS variables with value, not address of value, as required by POSIX

The GCSE one to me looks like it is wrongly labelled and should actually be VERIFICATION SUCCESSFUL. @rafaelsamenezes can you confirm this?

And the github_1091 test case now fails because it does `malloc(n+4)` and `malloc(n-4)`. I've marked it KNOWNBUG as this is issue #1631.


rafaelsamenezes:
> The GCSE one to me looks like it is wrongly labelled and should actually be VERIFICATION SUCCESSFUL. @rafaelsamenezes can you confirm this?

Yeah, it was true indeed.

rafaelsamenezes (approved):

lucasccordeiro (approved):

Can I ask you whether you could run it over the SV-COMP benchmarks?

fbrausse:
> Can I ask you whether you could run it over the SV-COMP benchmarks?

It's already running: [master](https://github.com/esbmc/esbmc/actions/runs/7697632732) and [this PR](https://github.com/esbmc/esbmc/actions/runs/7697630189). Should be ready in a few hours.

lucasccordeiro:
Many thanks, @fbrausse.

fbrausse:
Master:
```
  correct:           13872
    correct true:     7656
    correct false:    6216
  incorrect:            26
    incorrect true:     14
    incorrect false:    12
  unknown:            9907
  Score:             20888 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7697632732
```
This PR:
```
  correct:           13813
    correct true:     7603
    correct false:    6210
  incorrect:            27
    incorrect true:     14
    incorrect false:    13
  unknown:            9965
  Score:             20760 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/7697630189
```
I'll check the details tomorrow.

fbrausse:
The diff between the sv-comp results contains lots of timeout noise. The "new" incorrect is just faster, it's incorrect on master as well. Notable differences are in unreach-call:
- c/pthread-divine/tls_basic.i (correct true) performance down from 1s to timeout, I'll look into it
- c/ldv-challenges/linux-3.14_complex_emg_linux-drivers-clk1_drivers-net-ethernet-neterion-vxge-vxge.cil.i and
  c/ldv-challenges/linux-3.14_complex_emg_linux-kernel-locking-mutex_drivers-net-ethernet-neterion-vxge-vxge.cil.i
  both are timeout -> unknown, failing now with `ERROR: GOTO instruction type 0 not handled in goto_symext::symex_step`,
  I'm checking how master did behave and where this error might come from

In valid-memsafety also a false valid-memtrack result flipped to false valid-deref result, which is the correct one.

fbrausse:
> timeout -> unknown, failing now with `ERROR: GOTO instruction type 0 not handled in goto_symext::symex_step`,
> I'm checking how master did behave and where this error might come from

Master was the same, with this PR it's just faster.

PR #1633, opened 2024-01-30
[Concurrency] mpor fix

Anthonysdu:
The mpor algorithm is designed for three and more threads, and the rules are:

1. schedule the transitions in increasing order of thread ids (thread 0 -> thread 1 -> thread 2 ...)
2. for two **independent** transitions t1, t2 and tid(t1) < tid (t2), we only allow t2 schedule before t1 when:
1.) there is a dependency chain ftom t2 to t1.
2.) there exists a transition t3, where tid(t3) < tid(t1), and t2 <x t3 <x t1 along computation x, and there is a dependency chain from t2 to t3.

So I think we shouldn't rule out current thread if it is main thread.


lucasccordeiro:
@Anthonysdu : could you please run the master and this PR over the concurrency benchmarks in SV-COMP?

Anthonysdu:
This PR in 2 mins: https://github.com/esbmc/esbmc/actions/runs/7717433779/job/21037155911
```
  correct:             404
    correct true:      126
    correct false:     278
  incorrect:             3
    incorrect true:      3
    incorrect false:     0
  unknown:             258
  Score:               434 (max: 999)
```

Master (--no-por) in 2 mins: https://github.com/esbmc/esbmc/actions/runs/7717461614/job/21037215774
```
  correct:             392
    correct true:      114
    correct false:     278
  incorrect:             3
    incorrect true:      3
    incorrect false:     0
  unknown:             270
  Score:               410 (max: 999)
```

lucasccordeiro:
@Anthonysdu: you mention master with --no-por. Did you also run this PR with --no-por?

Similarly, did you also run the master with the same options of the PR?

Anthonysdu:
> @Anthonysdu: you mention master with --no-por. Did you also run this PR with --no-por?
> 
> Similarly, did you also run the master with the same options of the PR?

I believe this pr doesn't change the result with --no-por, but I will do other runs:
master with & without --no-por and this pr with & without --no-por in 15 mins.

lucasccordeiro:
Many thanks, @Anthonysdu.

Anthonysdu:
master --no-por https://github.com/esbmc/esbmc/actions/runs/7722408717
```
  correct:             421
    correct true:      138
    correct false:     283
  incorrect:             4
    incorrect true:      4
    incorrect false:     0
  unknown:             240
  Score:               431 (max: 999)

cpu time: 229458.81
wall time: 8128.87
```


master --por https://github.com/esbmc/esbmc/actions/runs/7722448790
```
  correct:             428
    correct true:      145
    correct false:     283
  incorrect:             5
    incorrect true:      5
    incorrect false:     0
  unknown:             232
  Score:               413 (max: 999)
```

this pr --por https://github.com/esbmc/esbmc/actions/runs/7722420270
```
  correct:             422
    correct true:      139
    correct false:     283
  incorrect:             4
    incorrect true:      4
    incorrect false:     0
  unknown:             239
  Score:               433 (max: 999)

cpu time: 223648.73
wall time: 7927.32
```

this pr --no-por https://github.com/esbmc/esbmc/actions/runs/7722478222
```
  correct:             421
    correct true:      138
    correct false:     283
  incorrect:             4
    incorrect true:      4
    incorrect false:     0
  unknown:             240
  Score:               431 (max: 999)
```
Overall, it didn't help imporve the svcomp score much, because all the incorrect trues need more context switches.

lucasccordeiro:
Many thanks, @Anthonysdu.

Can I ask you to fix the CI issues before we merge this PR into the master?

src/goto-symex/execution_state.h
```diff
@@ -413,6 +413,8 @@ class execution_statet : public goto_symext
    *  and is read-only. */
   bool is_transition_blocked_by_mpor() const
   {
+    if (active_thread == 0)
```
lucasccordeiro, line 422: Can you add some comments here to explain this if-statement?

lucasccordeiro (approved):

mikhailramalho (commented):
Do we need to context-bound all these test cases? Are they slower with this change?

Anthonysdu:
> Do we need to context-bound all these test cases? Are they slower with this change?

All of these cases run slower (timeout), because the old implemtation cut too many necessary interleavings, for example, the testcase I added, it only allow one interleaving to verify, but the violation assertion is at the 10th interleaving. 

So these regressions results running fast with mpor maybe incomplete. I think It's better to set context bounds if it doesn't change the veridit.

mikhailramalho:
> > Do we need to context-bound all these test cases? Are they slower with this change?
> 
> All of these cases run slower (timeout), because the old implemtation cut too many necessary interleavings, for example, the testcase I added, it only allow one interleaving to verify, but the violation assertion is at the 10th interleaving.
> 
> So these regressions results running fast with mpor maybe incomplete. I think It's better to set context bounds if it doesn't change the veridit.

I see, maybe we should mark them as THOROUGH and remove the context-bound then, or at least have a copy of the test with THOROUGH and no context bound

Anthonysdu:
> I see, maybe we should mark them as THOROUGH and remove the context-bound then, or at least have a copy of the test with THOROUGH and no context bound

Does "THOROUGH" mean keep running them until finish? I am worried it may cause the CI running a few hours slower.

mikhailramalho:
> > I see, maybe we should mark them as THOROUGH and remove the context-bound then, or at least have a copy of the test with THOROUGH and no context bound
> 
> Does "THOROUGH" mean keep running them until finish? I am worried it may cause the CI running a few hours slower.

AFAIK, these are slower tests that we run when a more comprehensive regression test is needed, day-to-day regression tests only run CORE tests.

@rafaelsamenezes can you confirm this?

rafaelsamenezes:
> > > I see, maybe we should mark them as THOROUGH and remove the context-bound then, or at least have a copy of the test with THOROUGH and no context bound
> > 
> > Does "THOROUGH" mean keep running them until finish? I am worried it may cause the CI running a few hours slower.
> 
> AFAIK, these are slower tests that we run when a more comprehensive regression test is needed, day-to-day regression tests only run CORE tests.
> 
> @rafaelsamenezes can you confirm this?
> 


The individual tests still have a timeout (I am not sure about the actual value). But yeah, thorough is supposed to be tests that need a specific solver or shouldn't executed in every configuration.

lucasccordeiro:
@Anthonysdu: we can have the two versions:

1) the original version marked as THOROUGH
2) the modified version with the context-bound set marked as CORE.

2) is already in this PR. Could you please do 1)?

src/goto-symex/execution_state.h
```diff
@@ -413,6 +413,14 @@ class execution_statet : public goto_symext
    *  and is read-only. */
   bool is_transition_blocked_by_mpor() const
   {
+    /** Mpor defines: schedule the transitions in increasing order of thread ids,
+     *  and for two independent transitions t1, t2 and tid(t1) < tid (t2), we only allow t2 schedule before t1 when:
+     *  there is a dependency chain ftom t2 to t1, or there exists a transition t3, where tid(t3) < tid(t1), and t2 <x t3 <x t1 along computation x,
```
fbrausse, line 418: Typo: "ftom".

Anthonysdu:
There are many timeouts (1500s) for the THOROUGH, I think it is another performance issue that needs to be investigated.

Anthonysdu:
Shall we remove or disable the THOROUGH testcases? @fbrausse @mikhailramalho?

lucasccordeiro:
@Anthonysdu: If I'm correct, we agreed to move these test cases to another folder in our meeting.

@fbrausse: what is the folder name?

fbrausse:
> @Anthonysdu: If I'm correct, we agreed to move these test cases to another folder in our meeting.
> 
> @fbrausse: what is the folder name?

It's `regression/disabled`. We can keep them there to revisit at a later point.

lucasccordeiro:

PR #1636, opened 2024-02-01
[clang-cpp-frontend] fix the cpp new initialisation

XLiZHI:
The current C++ frontend does not initialise some types correctly.

```
#include <cassert>

int main() 
{
    int* p = new int(10);

    assert(*p == 10);

    return 0;
}
```
image

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/new_init_fail/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/new_init_fail/main.cpp
@@ -1,0 +1,14 @@
+#include <cassert>
+
+int main() 
+{
+    int* p = new int(10);
+    double* s = new double(9.8);
+    bool* b = new bool(true);
+
+    assert(*p == 0); //should be 10
+    assert(*s == 0); //should be 9.8
+    assert(*b == true);
+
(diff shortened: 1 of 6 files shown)


regression/esbmc-cpp/bug_fixes/new_init_fail/main.cpp
```diff
@@ -0,0 +1,14 @@
+#include <cassert>
+
+int main() 
+{
+    int* p = new int(10);
+    double* s = new double(9.8);
+    bool* b = new bool(true);
+
+    assert(*p == 0); //should be 10
+    assert(*s == 0); //should be 9.8
```
brcfarias, line 10: Since there are two assertion failures, perhaps you could use --multi-property in the test.desc and verify if both failures are being emitted?

XLiZHI: Thanks for pointing it out！

lucasccordeiro (approved):
LGTM.

brcfarias (approved):

lucasccordeiro:

PR #1638, opened 2024-02-01
[ibex-contractor] bugfixes 

Mohannad-Aldughaim:
Small update with bugfixes:
- Ibex fixpoint was causing crashes, fixedpoint is done manually now.
- Integers (signed and unsigned) and floats are converted from one to another correctly now.
- Removed commented and unused code

### DIFF
--- a/src/goto-programs/goto_contractor.cpp
+++ b/src/goto-programs/goto_contractor.cpp
@@ -988,7 +988,7 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
         isinf(var.second.getInterval().ub()))
         continue;
 
-      BigInt upper_limit,lower_limit(0);
+      BigInt upper_limit, lower_limit(0);
       upper_limit = BigInt::power2(
         X->type->get_width() - (is_unsignedbv_type(X->type) ? 0 : 1));
       upper_limit = upper_limit - 1;
@@ -1005,7 +1005,7 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
 
         if (var.second.getInterval().lb() < lower_limit.to_int64())
           integerValue = lower_limit;
-        else if(var.second.getInterval().lb() > upper_limit.to_uint64())
+        else if (var.second.getInterval().lb() > upper_limit.to_uint64())
           integerValue = upper_limit;
         else
           integerValue = (long)ceil(var.second.getInterval().lb());
--- a/src/goto-programs/goto_contractor.cpp
+++ b/src/goto-programs/goto_contractor.cpp
@@ -462,8 +462,7 @@ vart::vart()
 void vart::dump()
 {
   std::ostringstream oss;
-  oss << to_symbol2t(this->getSymbol()).get_symbol_name() << this->getInterval()
-      << "\n";
+  oss << this->var_name << " " << this->getInterval();
   log_status("{}", oss.str());
 }
 
@@ -670,18 +669,24 @@ expr_to_ibex_parser::create_constraint_from_expr2t(const expr2tc &expr)
   if (f == nullptr || g == nullptr)
     return nullptr;
 
+  //change expression
+  //replace
   switch (base_object->expr_id)
   {
   case expr2t::expr_ids::greaterthanequal_id:
     c = new ibex::NumConstraint(*vars, (*f)(*vars) >= (*g)(*vars));
     break;
   case expr2t::expr_ids::greaterthan_id:
+    //if int do
+    //base_object->foreach_operand(op){};
+    //c = new ibex::NumConstraint(*vars, (*f)(*vars) >= (*g)(*vars)+1);
     c = new ibex::NumConstraint(*vars, (*f)(*vars) > (*g)(*vars));
     break;
   case expr2t::expr_ids::lessthanequal_id:
     c = new ibex::NumConstraint(*vars, (*f)(*vars) <= (*g)(*vars));
     break;
   case expr2t::expr_ids::lessthan_id:
+    //c = new ibex::NumConstraint(*vars, (*f)(*vars) <= (*g)(*vars)-1);
     c = new ibex::NumConstraint(*vars, (*f)(*vars) < (*g)(*vars));
     break;
   case expr2t::expr_ids::equality_id:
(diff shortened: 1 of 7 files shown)


lucasccordeiro (changes requested):
@Mohannad-Aldughaim : could you please add some regression tests here?

src/goto-programs/goto_contractor.cpp
```diff
@@ -933,9 +944,20 @@ void interval_analysis_ibex_contractor::apply_contractor()
   auto t_0 = std::chrono::steady_clock::now();
   auto X = map.create_interval_vector();
   auto c_out = contractor.get_outer();
+  auto Y = X;
+  int i = 50;
 
   c_out->contract(X);
 
+  //Find a fixed point in 5 or fewer iterations
+  while (Y != X && i >= 0)
```
fbrausse, line 953: 1. It looks to me like you're doing 50, not 5, iterations.
2. Where do the 50 come from and why has Ibex been crashing? We should file a bug with them or find out whether it's on our side.

src/goto-programs/goto_contractor.cpp
```diff
@@ -967,47 +989,42 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
...
 
-        auto w = to_symbol2t(X).type->get_width();
-        int is_unsigned = is_unsignedbv_type(X->type) ? 0 : 1;
-        long long type_limit = -pow(2, w - is_unsigned);
-        if (integer < type_limit)
-          integer = type_limit;
+        if (var.second.getInterval().lb() < r.to_int64())
+          integerValue = r;
+        else
+          integerValue = (long)ceil(var.second.getInterval().lb());
```
fbrausse, line 1011: Maybe I'm missing something, but couldn't the `.lb()` also be larger than the maximum value of (u)int64_t?

src/goto-programs/goto_contractor.cpp
```diff
@@ -967,47 +989,42 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
         continue;
 
       //if there is a lower bound,
-      if (!isinf(var.second.getInterval().lb()))
+      if (isnormal(var.second.getInterval().lb()))
```
fbrausse: Would `isfinite()` work here, too?

src/goto-programs/goto_contractor.cpp
```diff
@@ -967,47 +989,42 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
...
+        else
+          integerValue = (long)ceil(var.second.getInterval().lb());
 
-        auto lb = create_value_expr(integer, X->type);
+        auto lb = constant_int2tc(X->type, integerValue);
         auto cond1 = greaterthanequal2tc(X, lb);
         cond = and2tc(cond, cond1);
       }
-      if (!isinf(var.second.getInterval().ub()))
+      if (isnormal(var.second.getInterval().ub()))
```
fbrausse: Same question as above about isfinite().

src/goto-programs/goto_contractor.cpp
```diff
@@ -967,47 +989,42 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
...
-        long integerValue;
-
-        if (
-          (long)var.second.getInterval().ub() < 0 &&
-          var.second.getInterval().ub() > 0)
-          integerValue = (long)var.second.getInterval().ub() - 1;
-        else
-          integerValue = (long)var.second.getInterval().ub();
+        BigInt r(0);
+        r = r.power2(
```
fbrausse: `BigInt::power2` is a static function, no need to init to 0 first and call it as a member.

src/goto-programs/goto_contractor.cpp
```diff
@@ -967,47 +989,42 @@ expr2tc interval_analysis_ibex_contractor::result_of_outer(expr2tc exp)
...
 
-        //intersect with type limits
-        auto w = to_symbol2t(X).type->get_width();
-        int is_unsigned = is_unsignedbv_type(X->type) ? 0 : 1;
-        double type_limit = pow(2, w - is_unsigned) - 1;
+        BigInt integerValue(0);
 
-        if (integerValue > type_limit)
-          integerValue = type_limit;
+        if (var.second.getInterval().ub() > r.to_uint64())
```
fbrausse: Same question as above: Could the .ub() be smaller than the minimum value? Also, why are you comparing the .lb() to the signed int64, and the .ub() to the unsigned uint64?

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: I am not sure but does isfinite() include NaN? if not I'll change it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: 1- It should be 5. I'll change it. It is from Principles of Abstract Interpretation Book by Patrick Cousot. 
2- I think it's from their side. the crash happens when I try to access the function in the CtcFixedPoint to dump it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: You are right, it could be. I'll change it.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: For the lb(), the r is either the smallest int of the given bitvector for signed or 0 for unsigned so if we use signed value of 0 its fine.
For the ub(), we use uint because the calculated r is a positive integer which is the same value if it is signed and unsigned.

src/goto-programs/goto_contractor.h
```diff
@@ -405,6 +405,23 @@ class CspMap
...
+    oss << "This map has : " << this->n;
+    if (is_empty_vector)
+    {
+      oss << "empty vector";
+      log_status("{}", oss.str());
+      return;
+    }
+    log_status("{}", oss.str());
+
+    for (auto var : var_map)
```
fbrausse: This will create a copy of every std::pair in the var_map. Please use `const auto &` or similar.

src/goto-programs/goto_contractor.cpp
fbrausse: According to [documentation](https://linux.die.net/man/3/isfinite), isfinite() will be true when the argument is neither infinite nor NaN.

src/goto-programs/goto_contractor.h
Mohannad-Aldughaim: done.

src/goto-programs/goto_contractor.cpp
Mohannad-Aldughaim: changed.

src/goto-programs/goto_contractor.cpp
fbrausse: I see, thanks.

fbrausse (approved):
LGTM.

lucasccordeiro (approved):

PR #1639, opened 2024-02-03
[python-frontend] Inheritance and overriding

brcfarias:
This PR adds supports for inheritance in Python, addressing the following aspects:

- When a class lacks a definition for \_\_init\_\_, the corresponding method from the base class must be referenced. In such cases, the first class in the inheritance list with a defined \_\_init\_\_ method should be utilized.
- Instance attributes are now added in all instance methods.
- Methods overriding is also addressed.

The regression tests provides a overview of the features that are being handled.

### DIFF
--- a/regression/python/classes/main.py
+++ b/regression/python/classes/main.py
@@ -1,21 +1,16 @@
 class MyClass:
     def __init__(self, value: int):
-        self.x:int = value
-
-    def foo(self):
-        pass
+        self.data:int = value
 
 
 obj1 = MyClass(5)
-assert(obj1.x == 5)
-obj1.x = 10
-assert(obj1.x == 10)
-obj1.x = 20
-assert(obj1.x == 20)
+assert(obj1.data == 5)
+obj1.data = 10
+assert(obj1.data == 10)
+obj1.data = 20
+assert(obj1.data == 20)
 
 obj2 = MyClass(30)
-assert(obj2.x == 30)
-obj2.x = 40
-assert(obj2.x == 40)
-
-obj2.foo() # Calling base class method
+assert(obj2.data == 30)
+obj2.data = 40
+assert(obj2.data == 40)
--- a/regression/python/classes/main.py
+++ b/regression/python/classes/main.py
@@ -18,4 +18,4 @@ assert(obj2.x == 30)
 obj2.x = 40
 assert(obj2.x == 40)
 
-obj2.foo()
+obj2.foo() # Calling base class method
--- a/regression/python/classes/main.py
+++ b/regression/python/classes/main.py
@@ -1,16 +1,21 @@
 class MyClass:
     def __init__(self, value: int):
-        self.data:int = value
+        self.x:int = value
+
+    def foo(self):
+        pass
 
 
 obj1 = MyClass(5)
-assert(obj1.data == 5)
-obj1.data = 10
-assert(obj1.data == 10)
-obj1.data = 20
-assert(obj1.data == 20)
+assert(obj1.x == 5)
+obj1.x = 10
+assert(obj1.x == 10)
+obj1.x = 20
+assert(obj1.x == 20)
 
 obj2 = MyClass(30)
-assert(obj2.data == 30)
-obj2.data = 40
-assert(obj2.data == 40)
+assert(obj2.x == 30)
+obj2.x = 40
+assert(obj2.x == 40)
+
+obj2.foo()
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -368,11 +368,10 @@ symbolt *python_converter::find_function_in_base_classes(
   // Find class node in the AST
   auto class_node = json_utils::find_class(ast_json["body"], class_name);
 
-  std::string current_class = class_name;
-  std::string current_func_name = (is_ctor) ? class_name : method_name;
-
   if (class_node != nlohmann::json())
   {
+    std::string current_class = class_name;
+    std::string current_func_name = (is_ctor) ? class_name : method_name;
     // Search for method in all bases classes
     for (const auto &base_class_node : class_node["bases"])
     {
@@ -417,10 +416,9 @@ std::string python_converter::get_classname_from_symbol_id(
 
 exprt python_converter::get_function_call(const nlohmann::json &element)
 {
-  bool is_member_function_call = false;
-
   if (element.contains("func") && element["_type"] == "Call")
   {
+    bool is_member_function_call = false;
     std::string func_name;
     std::string obj_name;
     if (element["func"]["_type"] == "Name")
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -335,8 +335,8 @@ const nlohmann::json python_converter::find_var_decl(
   for (auto &element : json["body"])
   {
     if (
-      (element["_type"] == "AnnAssign") && (element["target"]).contains("id") &&
-      (element["target"]["id"] == var_name))
+      element["_type"] == "AnnAssign" && element["target"].contains("id") &&
+      element["target"]["id"] == var_name)
       return element;
   }
   return nlohmann::json();
@@ -359,7 +359,7 @@ python_converter::get_location_from_decl(const nlohmann::json &ast_node)
 
 symbolt *python_converter::find_function_in_base_classes(
   const std::string &class_name,
-  std::string symbol_id,
+  const std::string &symbol_id,
   std::string method_name,
   bool is_ctor) const
 {
@@ -372,6 +372,7 @@ symbolt *python_converter::find_function_in_base_classes(
   {
     std::string current_class = class_name;
     std::string current_func_name = (is_ctor) ? class_name : method_name;
+    std::string sym_id = symbol_id;
     // Search for method in all bases classes
     for (const auto &base_class_node : class_node["bases"])
     {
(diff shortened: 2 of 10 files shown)


src/python-frontend/python_converter.cpp
```diff
@@ -641,6 +766,29 @@ bool python_converter::is_constructor_call(const nlohmann::json &json)
...
+  const std::string &func_name,
+  const std::string &obj_symbol_id)
+{
+  std::string self_id =
+    create_symbol_id() + "@C@" + class_name + "@F@" + func_name + "@self";
+
+  auto self_instance = instance_attr_map.find(self_id);
+  if (self_instance != instance_attr_map.end())
+  {
+    std::vector<std::string> &attr_list = instance_attr_map[obj_symbol_id];
```
lucasccordeiro: Is the order of elements in `attr_list` important? If not, you could consider using `std::unordered_map` and `std::unordered_set` for faster lookups.

src/python-frontend/python_converter.cpp
```diff
@@ -1030,6 +1172,22 @@ void python_converter::get_class_definition(
...
+    // Get class definition from symbols table
+    symbolt *class_symbol = context.find_symbol("tag-" + base_class_name);
+    if (!class_symbol)
+    {
+      log_error("Base class not found: {}\n", base_class_name);
+      abort();
+    }
+    struct_typet &class_type = static_cast<struct_typet &>(class_symbol->type);
+    for (auto component : class_type.components())
+      clazz.components().push_back(component);
```
lucasccordeiro: Shall we use `emplace_back` instead of `push_back` for better performance?

regression/python/classes/main.py
```diff
@@ -1,16 +1,21 @@
...
+        pass
 
 
 obj1 = MyClass(5)
-assert(obj1.data == 5)
-obj1.data = 10
-assert(obj1.data == 10)
-obj1.data = 20
-assert(obj1.data == 20)
+assert(obj1.x == 5)
```
lucasccordeiro: I recommend keeping the original regression and creating a new one with these changes.

lucasccordeiro (changes requested):

lucasccordeiro (approved):
LGTM.

src/python-frontend/python_converter.cpp
```diff
@@ -335,7 +335,7 @@ const nlohmann::json python_converter::find_var_decl(
   for (auto &element : json["body"])
   {
     if (
-      (element["_type"] == "AnnAssign") &&
+      (element["_type"] == "AnnAssign") && (element["target"]).contains("id") &&
       (element["target"]["id"] == var_name))
```
fbrausse: The parenthesis around the individual conditions and subexpressions here are confusing. Please consider removing them.

src/python-frontend/python_converter.cpp
```diff
@@ -346,18 +346,91 @@ locationt
...
+      const std::string &base_class = base_class_node["id"].get<std::string>();
+      if (is_ctor)
+        method_name = base_class;
+
+      std::size_t pos = symbol_id.rfind("@C@" + current_class);
+
+      symbol_id.replace(
+        pos,
+        std::string("@C@" + current_class + "@F@" + current_func_name).length(),
+        std::string("@C@" + base_class + "@F@" + method_name));
```
fbrausse: Here, the value given to parameter `symbol_id` is modified in each iteration of the loop. Likely, this should operate on a copy of that value.

src/python-frontend/python_converter.cpp
```diff
@@ -641,6 +766,29 @@ bool python_converter::is_constructor_call(const nlohmann::json &json)
...
+
+  auto self_instance = instance_attr_map.find(self_id);
+  if (self_instance != instance_attr_map.end())
+  {
+    std::set<std::string> &attr_list = instance_attr_map[obj_symbol_id];
+
+    for (const auto &element : self_instance->second)
+    {
+      if (
+        std::find(attr_list.begin(), attr_list.end(), element) ==
```
fbrausse: No need for std::find() on a std::set. As a std::set contains unique elements, just .insert() can be used without even checking whether the element existed in the set before. In fact this entire manual loop can be replaced by a call to attr_list.[merge](https://en.cppreference.com/w/cpp/container/set/merge)().

src/python-frontend/python_converter.cpp
```diff
@@ -1030,6 +1172,22 @@ void python_converter::get_class_definition(
...
+    const std::string &base_class_name = base_class["id"].get<std::string>();
+    // Get class definition from symbols table
+    symbolt *class_symbol = context.find_symbol("tag-" + base_class_name);
+    if (!class_symbol)
+    {
+      log_error("Base class not found: {}\n", base_class_name);
+      abort();
+    }
+    struct_typet &class_type = static_cast<struct_typet &>(class_symbol->type);
+    for (auto component : class_type.components())
```
fbrausse: This creates a copy for each `component`. Please use `const auto &` or similar.

fbrausse (commented):
Does the delegation to the first `__init__` in the list of base classes work recursively, e.g., will instantiating D invoke B's init method here?
```python
class A: pass
class B:
  def __init__(self): pass
class C(B): pass
class D(A, C): pass
```

I am not a big fan of encoding the relationship between methods and classes in the symbol's identifier. Is the C++ frontend using the same mechanism, or could we possibly find a way to make these relationships accessible via some kind of API?

src/python-frontend/python_converter.cpp
brcfarias: Thanks for the comments.
`merge()` removes elements from the source. As they should not be removed, I kept the `insert()` call. In fact, the loop is unnecessary.

brcfarias:
> Does the delegation to the first `__init__` in the list of base classes work recursively?

Not yet, but I'll implement it. Thanks for pointing that out.

> I am not a big fan of encoding the relationship between methods and classes in the symbol's identifier. Is the C++ frontend using the same mechanism, or could we possibly find a way to make these relationships accessible via some kind of API?

The relationship is encoded in the AST/JSON: https://github.com/brcfarias/esbmc/blob/035bcf33fbcbf8b0a74762fc47fec05854ebd71c/src/python-frontend/python_converter.cpp#L377-L379

src/python-frontend/python_converter.cpp
fbrausse: I see, thanks for the clarification and update.

src/python-frontend/python_converter.cpp
fbrausse: Maybe I'm missing something, but with the current change a copy is being created outside of the loop, and the loop still repeatedly modifies this single copy. If the intention is to replace current_class with base_class and current_func_name with method_name, I think this will only work for the first iteration, as after that the replacement will have been stored in `sym_id`.

fbrausse:
> > Does the delegation to the first `__init__` in the list of base classes work recursively?
> 
> Not yet, but I'll implement it. Thanks for pointing that out.

:+1: 

> > I am not a big fan of encoding the relationship between methods and classes in the symbol's identifier. Is the C++ frontend using the same mechanism, or could we possibly find a way to make these relationships accessible via some kind of API?
> 
> The relationship is encoded in the AST/JSON: https://github.com/brcfarias/esbmc/blob/035bcf33fbcbf8b0a74762fc47fec05854ebd71c/src/python-frontend/python_converter.cpp#L377-L379

Right. I meant an API in ESBMC that allows obtaining symbols corresponding to base classes and/or overridden functions without having to fiddle with the identifiers in a non-documented way. Do you know whether or how the C++ frontend is doing this?

src/python-frontend/python_converter.cpp
brcfarias: >I think this will only work for the first iteration.

I guess I got your point now. _method_name_ doesn't need to be updated.
We are searching for the same method across different classes, so only the classname needs to vary.

For example if _main.py_ contains: 
```
class A:
  def foo()

class B(A):
  pass
  
obj = B()
obj.foo()
```
 
After parsing A and B we will have the following symbol for the method foo in the context:  py:main.py@C@A@foo

Given that the obj is of type B, we first look for "py:main.py@C@B@foo" (which doesn't exist), then look for "py:main.py@C@A@foo".

brcfarias:
> Right. I meant an API in ESBMC that allows obtaining symbols corresponding to base classes and/or overridden functions without having to fiddle with the identifiers in a non-documented way. Do you know whether or how the C++ frontend is doing this?

I see. In C++ this part is handled as follows: 

https://github.com/esbmc/esbmc/blob/14e5afd61283327b2defaf22304b06e8a47850ed/src/clang-c-frontend/clang_c_convert.cpp#L1692-L1723

If the method isn't virtual we simply call "new_expr = member_exprt(base, comp.name(), comp.type());", as seen on line 1706 above.

If it is virtual, this part is responsible for managing dynamic binding: https://github.com/esbmc/esbmc/blob/14e5afd61283327b2defaf22304b06e8a47850ed/src/clang-cpp-frontend/clang_cpp_convert_bind.cpp#L153-L174

I'm unsure if it's worth reusing this approach or perhaps extending struct_typet to include "base classes".

src/python-frontend/python_converter.cpp
fbrausse: That wasn't my point but seems to be valid, too. :)

Consider a class `Z` (one character) and one `XY` (2 characters) to be iterated over in the loop, in this order, and let the current class name be `ABC` (3 characters). What will happen is `sym_id = "py:main.py@C@ABC@F@foo"` is getting replaced by `sym_id = "py:main.py@C@Z@F@foo"`. In the second iteration, the substring `@C@ABC` will not be found anymore in `sym_id`, because after the first iteration the current_class `ABC` has already been replaced with `Z`.

That's why I thought that this should work on a new copy of the original string in each iteration.

*Edit:* Never mind, you're updating the current_class. Didn't see that, sorry. From my point of view this method could use a comment and I think a proper way of identifying sub-/base-class relationships and overridden methods in the symbol table is needed.

brcfarias: > In the second iteration, the substring @C@ABC will not be found anymore in sym_id, because after the first iteration the current_class ABC has already been replaced with Z.

current_class is updated at the end of each iteration. So, in the second call to sym_id.rfind("@C@" + current_class);, current_class would hold "Z", and the substring would be found.

fbrausse:
> I'm unsure if it's worth reusing this approach or perhaps extending struct_typet to include "base classes".

I also don't know, yet. To my mind, the best option would be to extend the `symbolt` class with that kind of information, not the types themselves. But we can leave this for later.

fbrausse (approved):

lucasccordeiro:

PR #1647, opened 2024-02-06
[Solidity] Tuple Support

JacobYiu:


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -541,6 +550,192 @@ bool solidity_convertert::get_struct_class_method(
...
+  // 1. populate name and id
+  std::string id, name;
+  struct_typet t = struct_typet();
+
+  // Conversion of tuple to struct so tuple has no default name
+  // But we still need to get unique label
+  // e.g. "sol:@C@BASE@tuple#14"
+  std::string label = std::to_string(struct_def["id"].get<int>());
+  name = "tuple#" + label;
+  id = "sol:@C@" + current_contractName + "@" + name;
```
ChenfengWei0: maybe add an assertion `assert(!current_contractName.empty());`

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -541,6 +550,192 @@ bool solidity_convertert::get_struct_class_method(
...
+  // 5. set symbol attributes for the struct
+  symbolt symbol;
+  get_default_symbol(symbol, debug_modulename, t, name, id, location_begin);
+
+  symbol.is_type = true;
+  symbolt &added_symbol = *move_symbol_to_context(symbol);
+
+  //Components consist of our variables used in our tuple
+  nlohmann::json ast_nodes;
+  ast_nodes = struct_def["components"];
```
ChenfengWei0: maybe check `assert(struct_def.contains("components"));`

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -541,6 +550,192 @@ bool solidity_convertert::get_struct_class_method(
...
+  // 6. Populate the Identifiers 
+
+  // std::string contract_name;
+  // if(get_current_contract_name(struct_def, contract_name))
+  // {
+  //   return true;
+  // }
+
+  for (nlohmann::json::iterator itr = ast_nodes.begin(); itr != ast_nodes.end();
+      ++itr)
```
ChenfengWei0: I feel like you should use `get_expr()` to parse each component. See `CallExprClass` => `structConstructorCall`:
```c++
// popluate components
    for (unsigned int i = 0; i < inits.operands().size() && i < args.size();
         i++)
    {
      exprt init;
      if (get_expr(args.at(i), members.at(i)["typeDescriptions"], init))
        return true;

      const struct_union_typet::componentt *c =
        &to_struct_type(t).components().at(i);
      typet elem_type = c->type();

      solidity_gen_typecast(ns, init, elem_type);
      inits.operands().at(i) = init;
    }
```

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2602,7 +2830,7 @@ bool solidity_convertert::get_func_decl_ref(
   exprt &new_expr)
 {
   // Function to configure new_expr that has a +ve referenced id, referring to a function declaration
-  // This allow to get func symbol before we add it to the symbol table
+  // This allow to get func syfmbol before we add it to the symbol table
```
ChenfengWei0: typo

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -4073,6 +4304,103 @@ nlohmann::json solidity_convertert::make_array_elementary_type(
...
+  // Add the last part
+  token = str.substr(startPos);
+  if (!token.empty()) {
+      output.push_back(token);
+  }
+
+  return output;
+}
+
+std::vector<nlohmann::json> solidity_convertert::make_struct_elementary_types(
```
ChenfengWei0: I think you can obtain the type info in the `typeDescriptions` node in each component, right? 

You can do something like:
```c++
   if (get_expr(args.at(i), members.at(i)["typeDescriptions"], init))
        return true;
```

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1543,10 +1738,43 @@ bool solidity_convertert::get_expr(
...
+      //Do we need to deal with mismatch of types? 
+      
+      //RHS 
+      std::cout << "Populating Struct " << std::endl;
+      
+
+      std::string label = std::to_string(expr["id"].get<int>());
+      std::string name, id;
+      name = "tuple#" + label;
+      id = "sol:@C@" + current_contractName + "@" + name;
```
ChenfengWei0: same as above

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -541,6 +550,192 @@ bool solidity_convertert::get_struct_class_method(
...
+
+  std::cout << "id is " << id << std::endl; 
+
+  // 2. Check if the symbol exists already
+  // If it does not exist, then it should return an error
+  // You cannot declare a variable in a tuple
+  // e.g. (uint age, uint money) = (....);
+  if(context.find_symbol(id) != nullptr)
+  {
+    std::cout << "Could not find symbol" << std::endl;
```
ChenfengWei0: Maybe just log_error

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -40,7 +41,7 @@ solidity_convertert::solidity_convertert(
 
 bool solidity_convertert::convert()
 {
-  // This function consists of two parts:
+  // This function consists of two output:
```
ChenfengWei0: I think we keep the word "parts"

src/solidity-frontend/solidity_convert.cpp
JacobYiu: I have yet to delete this function in case I may use it in the future for an unknown reason. I will delete it shortly

src/solidity-frontend/solidity_convert.cpp
JacobYiu: I think this is more suited towards RHS, since its calling a previous initialized struct and populating the values. But this is definitely useful for my current implementation, which is LHS. Thank you for the help!

ChenfengWei0:
Overall, it looks good. Just some small comments:
1. As said, for each tuple comment, we can obtain the type info from `typeDescriptions` nodes. However, I am also okay with the current implementation. You can stick to the `make_struct_elementary_types` if you are unfamiliar with the `get_expr` solution I mentioned.
2. `tupleQueueIds`. Do you know if it will work when handling interaction between contracts e.g. inheritance? Will there be an ID conflict?
```solidity
contract Base
{
  uint x;
  bool y;
  constructor() { (x,y) = (1,true); }
}

contract Derived is Base
{
  function test() public 
  {
    assert(y == true);
  }
}
```

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1155,6 +1167,124 @@ bool solidity_convertert::get_statement(
...
+        // get rhs
+        exprt rhs;
+        if (get_expr(stmt["expression"], rhs))
+          return true;
+
+        unsigned int ls = to_struct_type(lhs.type()).components().size();
+        unsigned int rs = rhs.operands().size();
+        assert(ls == rs);
+
+        for (unsigned int i = 0; i < ls; i++)
```
lucasccordeiro: Shall we use `size_t` here?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1155,6 +1167,124 @@ bool solidity_convertert::get_statement(
...
+              stmt["expression"]["typeDescriptions"],
+              func_call))
+          return true;
+        get_tuple_function_call(_block, func_call);
+
+        unsigned int ls = to_struct_type(lhs.type()).components().size();
+        unsigned int rs = to_struct_type(rhs.type()).components().size();
+        assert(ls == rs);
+
+        for (unsigned int i = 0; i < ls; i++)
```
lucasccordeiro: Shall we use `size_t` here?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2310,6 +2516,138 @@ bool solidity_convertert::get_binary_operator_expr(
...
+      if (rt.get("#sol_type") == "tuple_instance")
+      {
+        // e.g. (x,y) = (1,2); (x,y) = (func(),x);
+        // =>
+        //  t.mem0 = 1; #1
+        //  t.mem1 = 2; #2
+        //  x = t.mem0; #3
+        //  y = t.mem1; #4
+
+        unsigned int i = 0;
```
lucasccordeiro: Shall we use `size_t`?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2310,6 +2516,138 @@ bool solidity_convertert::get_binary_operator_expr(
...
+          return true;
+
+        // add function call
+        get_tuple_function_call(_block, rhs);
+
+        unsigned int ls = to_struct_type(lhs.type()).components().size();
+        unsigned int rs = to_struct_type(new_rhs.type()).components().size();
+        assert(ls == rs);
+
+        for (unsigned int i = 0; i < ls; i++)
```
lucasccordeiro: Shall we use `size_t`?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3432,6 +3778,261 @@ bool solidity_convertert::get_array_to_pointer_type(
...
+  get_default_symbol(symbol, debug_modulename, t, name, id, location_begin);
+  symbolt &added_symbol = *move_symbol_to_context(symbol);
+
+  auto &args = ast_node.contains("components")
+                 ? ast_node["components"]
+                 : ast_node["returnParameters"]["parameters"];
+
+  // populate params
+  //TODO: flatten the nested tuple (e.g. ((x,y),z) = (func(),1); )
+  unsigned int counter = 0;
```
lucasccordeiro: Shall we use `size_t`?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3432,6 +3778,261 @@ bool solidity_convertert::get_array_to_pointer_type(
...
+    return true;
+  if (c_name.empty())
+    return true;
+
+  name = "tuple_instance" + std::to_string(ast_node["id"].get<int>());
+  id = "sol:@C@" + c_name + "@" + name;
+  return false;
+}
+
+bool solidity_convertert::get_tuple_function_ref(
```
lucasccordeiro, line 3976: Try to document this method.

lucasccordeiro (approved):
I have just minor comments.

ChenfengWei0:
Changed as suggested. Thanks for the advice, @lucasccordeiro !

lucasccordeiro:
Thanks for submitting this PR, @JacobYiu and @ChenfengWei0. Great work!

Issue #1644, opened 2024-02-05
dereference adjustment


XLiZHI:
For the example code:

```
class Foo {
  public:
    Foo() {};
    void Execute();
};

int main()
{
  Foo *foo = new Foo();

  *foo;

  return 0;
}
```
I observed in the frontend that the expr of `*foo` is:
```
dereference
  * type: symbol
      * identifier: tag-Foo
  * operands: 
    0: symbol
        * type: pointer
            * subtype: symbol
                * identifier: tag-Foo
        * name: foo
        * identifier: c:main.cpp@128@F@main#@foo
        * #location: 
          * file: main.cpp
          * line: 17
          * function: main
          * column: 4
        * #lvalue: 1
```
`*foo` during symbolic execution is:
```
code_expression
* operand : dereference
    * pointer : symbol
        * name : c:main.cpp@128@F@main#@foo
        * renamelev : Level 0
        * level1_num : 0
        * level2_num : 0
        * thread_num : 0
        * node_num : 0
        * type : pointer
          * subtype : symbol
              * symbol_name : tag-Foo
    * type : struct
      * members : 
        0: signedbv
          * width : 32

      * member_names : 
        0: a

      * member_pretty_names : 
        0: a

      * typename : Foo
      * packed : false
* type : empty
```
I can't find out where this `type` was adjusted.
```
    * type : struct
      * members : 
        0: signedbv
          * width : 32

      * member_names : 
        0: a

      * member_pretty_names : 
        0: a

      * typename : Foo
      * packed : false
```
1. Can I ask where this type is adjusted?
2. Can we dereference in the absence of this type? Like dereference mode: free.

fbrausse:
Hi @XLiZHI. The frontend's type
```
  * type: symbol
      * identifier: tag-Foo
```
is a "symbolic" type, i.e., it stands for another one that can be obtained by looking it up in a namespace, via namespacet::lookup().

The reason for these symbolic types is to avoid loops in the irep1 (and irep2) representation for e.g. `struct slist` in
```c
struct slist {
  struct slist *next;
};
```
Loops in the representation of types are not supported/expected by many parts of ESBMC. I've described symbolic types and their role in the graph of all types/expressions of the program in #1303.

On the other hand, symbolic types are also not expected by many parts inside ESBMC (anymore), and the reason is that they are being replaced by the function thrash_type_symbols() everywhere, except under pointers (to avoid the loops).

[...]

PR #1648, opened 2024-02-06
[C++ VERIFICATION] Improved cpp new and delete

XLiZHI:
Fixed C++ dangling pointer checking. 

Fixed C++ double delete.

Next PR todo: move the adjustment of cpp new and delete into symex.

still wait for #1644

### DIFF
--- a/regression/esbmc-cpp/cpp/dangling_memleak/main.cpp
+++ b/regression/esbmc-cpp/cpp/dangling_memleak/main.cpp
@@ -1,0 +1,31 @@
+#include <iostream>
+
+class Foo {
+  public:
+    Foo() {value = 0;};
+    void Execute();
+    void Increment();
+  private:
+    int value;
+};
+
+void Foo::Execute() {
+        printf("Executing...");
+}
+
+void Foo::Increment() {
+    value++;
+}
+
+int main()
+{
+  Foo *foo = new Foo();
+
+  foo->Increment();  // Incrementing the value
+
+  foo->Execute(); // Executing
+
+  delete foo;
+
+  return 0;
+}
--- a/regression/esbmc-cpp/cpp/dangling_memleak/test.desc
+++ b/regression/esbmc-cpp/cpp/dangling_memleak/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+--memory-leak-check
+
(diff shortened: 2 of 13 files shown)


regression/esbmc-cpp/cpp/ dangling_ptr/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro: There is a memory leak in this program.

Can you please create another test case that enables --memory-leak-leak?

regression/esbmc-cpp/cpp/ dangling_ptr/main.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+int main()
+{
+  Foo *foo = new Foo();
+
+  foo->Increment();  // Incrementing the value
+
+  foo->Execute(); // Executing
+
```
lucasccordeiro: Can you also create another test case that adds the delete to mark it as VERIFICATION SUCCESSFUL?

lucasccordeiro (changes requested):
I have two comments here.

lucasccordeiro (approved):

fbrausse (approved):
You can unify both branches by casting `*code2` to the common base class `const code_expression_data &`. Otherwise LGTM.

XLiZHI:
> You can unify both branches by casting `*code2` to the common base class `const code_expression_data &`. Otherwise LGTM.

Hi @fbrausse,

Thanks for your advice, I tried this, but cmake complained about it.
```
const code_expression_data &deref_code = is_code_cpp_delete2t(code2)? to_code_cpp_delete2t(code2)
                                                                    : to_code_cpp_del_array2t(code2);
```
different types ‘code_cpp_delete2t’ and ‘code_cpp_del_array2t’. Or do we have a general type cast function?

fbrausse:
`const auto &code_expr = static_cast<const code_expression_data &>(*code2);` should work.

XLiZHI:
> const auto &code_expr = static_cast<const code_expression_data &>(*code2);

Great, thanks

lucasccordeiro:

Issue #1637, opened 2024-02-01
Design for Regression Tests


rafaelsamenezes:
Right now, our regression suite looks like:

```
<MODE>
<MAIN FILE>
<FLAGS>
<regex to match>
```

From my understanding:

- Mode: CORE (less than 30s tests that don't need any specific solver or arch), KNOWNBUG (tests that are failing and were supposed to work), THOROUGH (longer tests or tests that need a specific feature that is not available in every configuration), and FUTURE (which is basically KNOWNBUG)
- Main file: The file that contains the main entrypoint (kinda redundant)
- Flags: The flags and extra files
- Regex: A multiline regex that must be met for the test to be considered accepted.

However, this syle is starting to show some limitations. Mainly in the thorough because of the many: architectures, solvers and operating systems. We usually solved this by creating a directory (as a "test suite"), which is very limitated and hard to combine considering all the configurations available.

Should we consider evolving the format?

mikhailramalho:
>
> Should we consider evolving the format?
>

Definitely; having to duplicate test cases to test them with different
flags is a major drawback IMHO — also perl.

I particularly like that the tests are separated into directories, but I
don't mind changing that.

-- 

Mikhail R. Gadelha, Ph.D.

fbrausse:
Wrt. #1585 I think we need a way to opt-in for --32 and --64 tests, for big and little endian ones, basically everything that changes the default architecture settings. Same for #1118 and <https://github.com/esbmc/esbmc/issues/1240#issuecomment-1664046567>: Opt-in to regression tests that are not generic wrt. solver.

As @rafaelsamenezes is well aware, I'm a Gentoo user, so the following is taken from their USE-flag approach for packages:
1. for each dimension (solver, bit-widths, endian, etc.) define an identifier and assign the list of all used identifiers as a new field in the .desc file; default: empty
2. bonus points for allowing to combine KNOWNBUG with THOROUGH and CORE

Regarding item 1., since we know ESBMC's flags, we can automate the initial migration via a script detecting flags. The testing_tool.py already parses the flags in order to decide which ones of those are file names that need prefixing when running ESBMC outside of the test's directory: https://github.com/esbmc/esbmc/blob/4dc3b4898c2ef754fbf68a736bad38ec6246e4b2/regression/testing_tool.py#L58-L61

brcfarias:
I also think it's a good idea to manage this through lists within each test.desc file.

rafaelsamenezes:
[...]

PR #1659, opened 2024-02-09
Replacement of XML tests

rafaelsamenezes:
This PR is an intermediate step to help us with #1637

This was done by: adding a method to apply a function over all testcases, and by adding a `save_test()` into BaseTest (now named TestCase) that overwrites any tests in our current spec. So it was basically invoking apply_transform_over_tests with a lambda that invokes `save_test()`

In Summary:

1. All tests are now unified in a TestCase class
2. Removed the capacity of running all unit tests directly from testing_tool (this functionality was deprecated in favour of using CMake to find the tests)
3. Added a function to apply a function over all test cases.


regression/testing_tool.py
```diff
@@ -381,3 +287,29 @@ def main():
...
     main()
+
+# Utilities for REPL. This helps us to do batch changes over all tests.
+# Example:
+# 1. Invoke python3 in shell
+# 2. `import testing_tool`
+# 3. `testing_tool.apply_transform_over_tests(testing_tool.print_test)`
+
+# TODO We probably should obtain this from CMake somehow
+TEST_SUITES = [ "bitwuzla", "cvc", "mathsat", "smtlib", "z3", "esbmc-solidity", "Interval-analysis-ibex-contractor",
```
fbrausse: Re comment: yes, we should. People create new sub-directories all the time.

rafaelsamenezes: I guess we could change the transform function to work with a directory and use CMake to invoke a wrapper of the transform.

fbrausse: Sounds complicated. As this is for manual use only, I'm fine leaving this list hardcoded for now. Maybe add a comment to check whether the list is still up to date for manual invocations of the new function.

regression/testing_tool.py
```diff
@@ -381,3 +287,29 @@ def main():
...
+                "esbmc-old", "jimple", "python", "cheri-c", "cheri-128", "esbmc-cpp/cpp", "esbmc-cpp/cbmc", "esbmc-cpp/polymorphism_bringup",
+                "esbmc-cpp/inheritance_bringup", "esbmc-cpp/bug_fixes", "esbmc-cpp/OM_sanity_checks", "esbmc-cpp/string", "esbmc-cpp/stream",
+                "esbmc-cpp/gcc-template-tests", "esbmc-cpp/template", "esbmc-cpp/algorithm", "esbmc-cpp/deque", "esbmc-cpp/list", "esbmc-cpp/map",
+                "esbmc-cpp/multimap", "esbmc-cpp/multiset", "esbmc-cpp/priority_queue", "esbmc-cpp/queue", "esbmc-cpp/set", "esbmc-cpp/stack",
+                "esbmc-cpp/unix", "esbmc-cpp/vector", "esbmc-cpp11/cpp", "esbmc-cpp11/constructors", "esbmc-cpp11/new-delete", "esbmc-cpp11/reference",
+                "esbmc-unix", "esbmc-unix2", "esbmc", "cbmc", "cstd", "llvm", "floats", "floats-regression", "k-induction", "csmith", "k-induction-parallel",
+                "cuda/benchmarks", "cuda/Supported_long_time", "nonz3", "incremental-smt", "extensions", "cuda/COM_sanity_checks", "linux"]
```
fbrausse: It's easier to work with git or other line-based tools when such lists (a) have just a single entry per line and, less importantly, (b) are ordered (alphabetically, for instance). Fewer merge conflicts.

fbrausse (approved):
Thanks Rafael, this is a good initiative.

lucasccordeiro:

PR #1660, opened 2024-02-10
Removed testllvm.desc tests

### LINKED ISSUE #1658: testllvm.desc regression files
Why they exist? Can we just remove it? They are currently not checked and they also differ from the test.desc file

rafaelsamenezes:
These test descriptions exists for historical reasons and are not used.

Closes #1658


lucasccordeiro:

PR #1662, opened 2024-02-11
< and <= fixes for Wrapped Intervals

rafaelsamenezes:
The operators were implemented incorrectly. This PR fixes it by using a new intermediate function to compute the bounds of the interval. Additionally, some small fixes were done for the inver_bool and equality functions.

Master (with --interval-analysis-wrapped):

```
  correct:            4224
    correct true:     3161
    correct false:    1063
  incorrect:           771
    incorrect true:    579
    incorrect false:   192
  unknown:            4542
  Score:            -14215 (max: 15923)
```

This PR (with --interval-analysis-wrapped):

```
  correct:            4282
    correct true:     2667
    correct false:    1615
  incorrect:             6
    incorrect true:      6
    incorrect false:     0
  unknown:            5249
  Score:              6757 (max: 15923)
```

### DIFF
--- a/src/goto-programs/abstract-interpretation/wrapped_interval.h
+++ b/src/goto-programs/abstract-interpretation/wrapped_interval.h
@@ -959,13 +959,13 @@ public:
       Examples (signed char):
 
       - get_interval_bounds([10, 127]) --> <10, 127>
-      - get_interval_bounds([10, 128]) --> <-128, 10>
-      - get_interval_bounds([10, 255]) --> <-128, 10>
+      - get_interval_bounds([10, 128]) --> <-128, 127>
+      - get_interval_bounds([10, 255]) --> <-128, 127>
       - get_interval_bounds([129, 130]) --> <-127, -126>
       - get_interval_bounds([255, 10]) --> <-1, 10>
 
-      @warning This is not to represent a range interval, wrapped can have holes!
-      From the example: [10, 128] contains both <-128, 10> but it does not contain 9 or -127!
+      Note: This is not to represent a range interval, wrapped can have holes!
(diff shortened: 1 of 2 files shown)


fbrausse:
I have to say that my understanding of the terminology used in the interval implementation is limited. For example, I don't know what *north-/south-pole* of a type really means, hence I also have no intuition what the *cuts* are in this context.

After a while I understood the example in the comment on `get_interval_bounds`, but the `@brief` description I did not.

lucasccordeiro (commented):
@rafaelsamenezes: could you please add some comments / examples to explain this code?

rafaelsamenezes:
> I have to say that my understanding of the terminology used in the interval implementation is limited. For example, I don't know what *north-/south-pole* of a type really means, hence I also have no intuition what the *cuts* are in this context.
> 
> After a while I understood the example in the comment on `get_interval_bounds`, but the `@brief` description I did not.

I'm sorry, I will expand the documentation on those functions. In summary, these are the regions where -1 (or max uint) become zero (for South pole) and max int becomes min int (for north pole).
The terms come from the original paper, so that why I used them. I am fine with changing and adding a note if you have a more common name.

lucasccordeiro (approved):

fbrausse (approved):
> In summary, these are the regions where -1 (or max uint) become zero (for South pole) and max int becomes min int (for north pole).
The terms come from the original paper, so that why I used them.

I understand nothing, am going to have to read the paper. The changes here, in particular wrt. invert_bool() and equality() look fine, though.

lucasccordeiro:
@rafaelsamenezes: could you please create another related PR with regression tests?

Issue #825, opened 2022-07-27
[C++ Verification] Standardize all the dynamic alloc call


kunjsong01:
Previously discussed in https://github.com/esbmc/esbmc/pull/815#issuecomment-1196741044

Standardize all the dynamic alloc call to deal with malloc/new and free/delete in symex.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/1670.

PR #1670, opened 2024-02-12
[C++ verification] Standardize the cpp new/delete

XLiZHI:
see #825

### DIFF
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -542,9 +542,9 @@ void goto_symext::symex_input(const code_function_call2t &func_call)
 
 void goto_symext::symex_cpp_new(const expr2tc &lhs, const sideeffect2t &code)
 {
-  bool do_array;
+  expr2tc size = code.size;
 
-  do_array = (code.kind == sideeffect2t::cpp_new_arr);
+  bool do_array = (code.kind == sideeffect2t::cpp_new_arr);
 
   unsigned int &dynamic_counter = get_dynamic_counter();
   dynamic_counter++;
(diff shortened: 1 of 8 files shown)


src/goto-symex/builtin_functions.cpp
```diff
@@ -588,27 +588,33 @@ void goto_symext::symex_cpp_new(const expr2tc &lhs, const sideeffect2t &code)
...
+  expr2tc ptr_obj = pointer_object2tc(pointer_type2(), ptr_rhs);
+  track_new_pointer(ptr_obj, newtype, size);
 
   dynamic_memory.emplace_back(
     rhs_copy, cur_state->guard, false, symbol.name.as_string());
 }
 
-// XXX - implement as a call to free?
-void goto_symext::symex_cpp_delete(const expr2tc &)
+void goto_symext::symex_cpp_delete(const expr2tc &expr)
```
fbrausse: Why not implement it as a call to symex_free()?

XLiZHI: Hi @fbrausse, Do you mean to use `symex_free()` inside this function? 

Maybe we don't need `symex_cpp_delete`, I tried using symex_free instead:

```
  else if (is_code_cpp_del_array2t(code2) || is_code_cpp_delete2t(code2))
  {
    replace_dynamic_allocation(code2);
    replace_nondet(code2);

    symex_free(code2);
  }
```

fbrausse: We should probably check that the allocation has actually been performed with `new[]` for `delete[]`, but other than that I think symex_cpp_delete() can just delegate to symex_free().

lucasccordeiro (approved):
This is a very nice PR!

Thanks for submitting it, @XLiZHI.

rafaelsamenezes (approved):

PR #1671, opened 2024-02-12
[C++ verification] added memory deallocation check

XLiZHI:
see https://github.com/esbmc/esbmc/pull/1670#discussion_r1486085149

### DIFF
--- a/regression/esbmc-cpp/cpp/mismatch_del/main.cpp
+++ b/regression/esbmc-cpp/cpp/mismatch_del/main.cpp
@@ -1,0 +1,8 @@
+int main()
+{
+  int* IntArray = new int[5];
+    
+  delete IntArray; //should be IntArray[]
+
+  return 0;
+}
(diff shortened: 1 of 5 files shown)


fbrausse (commented):
Great, thank you @XLiZHI! Could you add a positive test case as well (just a modification of the one you already added)?

As a next step, not necessarily in this PR, we might want to consider to disallow passing `new` (and `new[]`) allocated pointers to free(). Same for malloc and delete, I guess.

XLiZHI:
> Great, thank you @XLiZHI! Could you add a positive test case as well (just a modification of the one you already added)?

Hi @fbrausse ,

A positive one could be:
```
int main()
{
  int* IntArray = new int[5];

  delete[] IntArray; 

  return 0;
}
```
But since we already have many similar TCs in the regression, I think maybe we don't need to add it?

fbrausse (approved):
> But since we already have many similar TCs in the regression, I think maybe we don't need to add it?

I don't really have an overview of the current C++ regression tests. If you believe that's already there, I'm fine with this PR as it is.

lucasccordeiro (approved):

lucasccordeiro:

Issue #1667, opened 2024-02-11
[C++] Parsing fails with combinations of `--no-abstracted-cpp-includes` and/or `--no-library` for `<iostream>`


intrigus-lgtm:
### Information
ESMB version: `ESBMC version 7.5.0 64-bit x86_64 linux`
Kernel/OS: `Linux debian 6.1.0-17-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.69-1 (2023-12-30) x86_64 GNU/Linux`
GCC version: `gcc (Debian 12.2.0-14) 12.2.0`
Clang version: `Debian clang version 14.0.6 Target: x86_64-pc-linux-gnu`

### Code

```cpp
#include <iostream>
#include <string_view>

int main()
{
    std::string_view message = "Hello, World!";
}

```

If we run `esbmc --cppstd 17 --input-file cpptest.cpp --no-abstracted-cpp-includes --no-library` we get the following errors:
<details>

```shell
$ esbmc --cppstd 17 --input-file cpptest.cpp --no-abstracted-cpp-includes --no-library 
ESBMC version 7.5.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with systemlibc
Parsing cpptest.cpp
In file included from cpptest.cpp:1:
In file included from /../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/iostream:39:
In file included from /../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/ostream:38:
In file included from /../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/ios:40:
/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/char_traits.h:431:33: error: use of undeclared identifier 'memcpy'; did you mean 'wmemcpy'?
        return static_cast<char_type*>(__builtin_memcpy(__s1, __s2, __n));
                                       ^
<command line>:6:26: note: expanded from here
#define __builtin_memcpy memcpy
                         ^
/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/cwchar:201:11: note: 'wmemcpy' declared here
  using ::wmemcpy;
          ^
In file included from cpptest.cpp:1:
In file included from /../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/iostream:39:
[...]

PR #1672, opened 2024-02-12
[C++ Operational Models] Added <string_view> model

lucasccordeiro:
This PR adds the C++ <string_view> operational model to address this issue https://github.com/esbmc/esbmc/issues/1667.

It also fixes the warning produced for the C++ `<iostream>` operational model.

````CPP
#include <iostream>

int main()
{
  std::cout << "Hello World!" << std::endl;
  return 0;
} 
````

````
hello.cpp:7:34: warning: address of function 'std::endl' will always evaluate to 'true' [-Wpointer-bool-conversion]
  std::cout << "Hello World!" << std::endl;
                              ~~ ^~~~~~~~~
hello.cpp:7:34: note: prefix with the address-of operator to silence this warning
  std::cout << "Hello World!" << std::endl;
                                 ^
                                 &
````

### DIFF
--- a/regression/esbmc-cpp/cpp/hello_world/main.cpp
+++ b/regression/esbmc-cpp/cpp/hello_world/main.cpp
@@ -1,7 +1,7 @@
 #include <iostream>
 
-int main() {
+int main()
+{
   std::cout << "Hello World!" << std::endl;
   return 0;
 }
-
--- a/regression/esbmc-cpp/cpp/hello_world/main.cpp
+++ b/regression/esbmc-cpp/cpp/hello_world/main.cpp
@@ -1,0 +1,7 @@
+#include <iostream>
+
+int main() {
+  std::cout << "Hello World!" << std::endl;
+  return 0;
+}
+
--- a/regression/esbmc-cpp/cpp/hello_world/test.desc
+++ b/regression/esbmc-cpp/cpp/hello_world/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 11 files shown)


fbrausse:
Hi Lucas! There are two comments I have to this PR, neither of them is a blocker, it's just to get feedback:
1. The string_view class in the stdlib is templated based on the character type of the string. The model this PR supplies is not. This may lead to complications in the future when users try to use the templated version (e.g., on `wchar_t`).
2. More generally, do we aim at "hiding" the C++ standard library, or do we want to embrace it in the sense that we fully support their implementations as much as possible? In case the answer is yes to the latter, I think this PR might be going into the wrong direction by adding more models (in this case just headers, really) that are possibly incompatible with other parts of the stdlib.

The reason I'm bringing up item 2. is that it is my view that we're not equipped to keep up with the development of the C++ standard (library): It takes a lot of work to support the recent additions (C++20 ranges, for instance, will be resource-intensive) just for the library side of things, not talking about language changes.

lucasccordeiro:
Hi @fbrausse,

Many thanks for your comments.

Regarding 1), can I ask you please to expand this PR to consider the template based approach on the character type as you suggested?

Regarding 2), as we discussed yesterday, I believe the best approach would be to support the system's headers, i.e, to be as close as possible to the standard C++ libraries. What do the other colleagues think?

fbrausse:
> Regarding 1), can I ask you please to expand this PR to consider the template based approach on the character type as you suggested?

As I said, I don't think it's necessary right now for this PR. I just wanted to bring it up so that everyone's aware. We can add the templates in another PR.

lucasccordeiro:
Thanks, @fbrausse. I will merge this PR for now and open an issue to consider the template based version.

PR #1673, opened 2024-02-12
Support for extracting test labels

rafaelsamenezes:
This PR adds support to extract labels from the test.desc file. I am opening this as a draft because I am not sure whether this is a good approach or not. In summary, we use CMake to parse the test file and extract the tags. I also tried to use python (with this
[testing_tool.patch](https://github.com/esbmc/esbmc/files/14251197/testing_tool.patch)) with `execute_process` and storing the stdout.


mikhailramalho:

PR #1675, opened 2024-02-13
[C++ operational models] enable regression tests for the <algorithm> lib

lucasccordeiro:


src/cpp/library/algorithm
```diff
@@ -14,725 +14,868 @@
...
-bool equal(InIt1 first1, InIt1 last1, InIt2 first2) {
-	while (first1 != last1) {
-		if (!(*first1 == *first2))
-			return false;
-		++first1;
-		++first2;
-	}
-	return true;
-}
+typedef bool(funcao)(int arg1, int arg2);
```
rafaelsamenezes: Should we consider removing the Portuguese?

lucasccordeiro: Yes :-) This OM was written by @feliperodri in 2012 :-)

lucasccordeiro: Done.

src/cpp/library/utility
```diff
@@ -56,7 +56,7 @@ pair<T1, T2> make_pair(T1 x, T2 y)
  */
 
 template <class T>
-typename std::remove_reference<T>::type &&move(T &&t) noexcept
+typename std::remove_reference<T>::type &&move(T &&t)
```
XLiZHI: Not specifically for this PR, I think we may need to decide whether or not to convert the function depending on the cpp version.

lucasccordeiro: I consider this:

````CPP
#if __cplusplus >= 201103L // Check if C++11 or later is being used
template <class T>
typename std::remove_reference<T>::type &&move(T &&t) noexcept
#else // Code for older C++ standards
template <class T>
typename std::remove_reference<T>::type &&move(T &&t)
#endif
{
  return static_cast<typename std::remove_reference<T>::type &&>(t);
}
````

Can we do better than this?

XLiZHI: Maybe we can identify and handle it in the frontend? because we have to deal with the standard library, may encounter various situations. For this PR, I think it's good now.

lucasccordeiro: The std::move utility didn't exist in C++98. 

The correct implementation should be:

````CPP
#if __cplusplus >= 201103L // Check if C++11 or later is being used
template <class T>
typename std::remove_reference<T>::type &&move(T &&t) noexcept
{
  return static_cast<typename std::remove_reference<T>::type &&>(t);
}
#endif

````

lucasccordeiro: > Maybe we can identify and handle it in the frontend? because we have to deal with the standard library, may encounter various situations. For this PR, I think it's good now.

I agree with you. 

Let's open an issue to remind us to handle these cases in the C++ frontend.

rafaelsamenezes (approved):

XLiZHI (approved):

PR #1679, opened 2024-02-14
Removed insecure from build.sh

### LINKED ISSUE #1677: [Security] Disabling security of curl _permanently_ when using `scripts/build.sh`
This code here disables the security of curl **permanently** for the current user.
This is **unacceptable**.
https://github.com/esbmc/esbmc/blob/2d6cd6d0bd997bfafa0c28d7ff49904de22c75f5/scripts/build.sh#L91-L94

I also don't see why this is necessary. Boolector appears to only use `curl` twice and not in a way that needs this hack I'd say:
https://github.com/search?q=repo%3ABoolector%2Fboolector%20%22curl%22&type=code

rafaelsamenezes:
It used to cause problems during Boolector builds.

Closes #1677

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -88,10 +88,6 @@ ubuntu_setup () {
     SOLVER_FLAGS="$SOLVER_FLAGS \
         -DENABLE_Z3=ON \
     " &&
-    # Hack: Boolector might fail to download some dependencies using curl (maybe we should patch it?)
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

lucasccordeiro:

fbrausse (approved):

PR #1680, opened 2024-02-14
Add tool regdb.py for regression test management

fbrausse:
The purpose of this tool is to more easily query and list the test cases and some of their specifics in order to get an overview. For instance:

	$ ./regdb.py -v

will list all known (and enabled) test cases including what the tool calls "flags" for each one. Shortened example output:

	esbmc-cpp11/constructors/Constructor14-1                               : cxx
	esbmc-cpp11/constructors/Constructor15-fail                            : bug cxx
	esbmc-unix/00_race06                                                   : c
	esbmc-unix/00_race07                                                   : 32 c
	esbmc-unix/github_1147                                                 : cxx
	esbmc-unix/github_28_successful                                        : 64 c
	Interval-analysis-ibex-contractor/01-arithmetic-operations-fail        : c ia
	jimple/kt-array-false                                                  : jimple
	mathsat/00_Division_false-no-overflow                                  : c mathsat

The list of flags can be obtained with -l.

It supports querying the set of all regression tests based on the flags, using an extremely simple language for now (described in help message). I'm open to suggestions what features that query language should support.

With @rafaelsamenezes we discussed that using this approach might be a way to get CTest labels attached in a generic way, and it should also help to select the tests compatible with the build at build time. This PR isn't attempting that, yet.


regression/regdb.py
```diff
@@ -0,0 +1,174 @@
...
+    'bitwuzla'
+]
+
+EXTENSIONS = {
+    'c'     : {'c', 'i'},
+    'cxx'   : {'cc', 'cpp'},
+    'cuda'  : {'cuda'},
+    'py'    : {'py'},
+    'sol'   : {'solast'},
+    'jimple': {'jimple'},
```
rafaelsamenezes, line 26: We also have .goto

regression/regdb.py
```diff
@@ -0,0 +1,174 @@
...
+    else:
+        for f in sorted(flgs):
+            print(f)
+    # finish
+    sys.exit(0)
+
+def flags(tc : TestCase):
+    r = set()
+    if tc.test_mode in FAIL_MODES:
+        r.add('bug')
```
rafaelsamenezes: Maybe convert this into a global for reference

regression/regdb.py
```diff
@@ -0,0 +1,174 @@
...
+OPT2FLAGS = {
+    '--' + solver: {solver} for solver in SOLVERS
+} | { # Frontend related
+    '--16'           : {'16'},
+    '--32'           : {'32'},
+    '--64'           : {'64'},
+    '--binary'       : {'goto'},
+    '--little-endian': {'le'},
+    '--big-endian'   : {'be'},
+    '--no-arch'      : {'ne', 'no-arch'},
```
rafaelsamenezes, line 38: I am not familiar with low level stuff, so I don't know how common it is. But I don't think "be", "le" and "ne" are intuitive.

rafaelsamenezes (changes requested):
Overall looks good. Just some small comments.

regression/regdb.py
fbrausse: It's under the `--binary` option as we don't have it by file name extension.

regression/regdb.py
fbrausse: done.

regression/regdb.py
fbrausse: "be" and "le" are quite common abbreviations, e.g., for the target triples "armel-unknown-linux-gnu". The point was for these flags to be brief, so they are easy to type and read, in particular when there are many of them. The options `-l -v` give a list of all known flags together with descriptions. The descriptions are still missing, unfortunately.

rafaelsamenezes: It's fine then. Just to be sure that it's a common thing.

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #1682, opened 2024-02-14
Support detecting Clang resource dir for >= v16

### LINKED ISSUE #1661: Failed to determine path to Clang headers on openSUSE
Configuring ESBMC on openSUSE Tumbleweed causes ESBMC to fail with `Failed to determine path to Clang headers: not a directory: /usr/lib64/clang/17.0.6/include`. The correct path is `/usr/lib64/clang/17/include` and using ` -DOVERRIDE_CLANG_HEADER_DIR=<path>` does not help.

I also noticed the same issue in macOS. Currently, I do a symbolic link as a workaround.

fbrausse:
Closes #1661, I hope.


rafaelsamenezes (approved):
Thanks, it fixed the issue.

PR #1690, opened 2024-02-19
regdb: allow complete formulas in queries

fbrausse:
Instead of just conjunctively combined glob patterns, queries can now be formulas. From the help message:
```
QUERY is an infix formula (! or - for negation, & or space for conjunction, |
for disjunction) over glob patterns being matched against flags of test cases.
QUERY has this grammar (with spaces allowed around any of Formula, Product and
Factor):

  Formula   ::= Product ('|' Product)*
  Product   ::= Factor ('&'? Factor)*
  Factor    ::= '(' Formula ')'
              | ('!' | '-') Factor
              | '%' Predicate
              | GlobPattern
  Predicate ::= 'consistent'

The named predicate 'consistent' is true iff the set of flags of a test case are
consistent, that is, for categorized flags of the form CATEGORY=VALUE at most
one VALUE is allowed per CATEGORY.
```
Example for consistency checking (here, there are multiple different solvers specified):
```
$ ./regdb.py -vq '!%consistent'
csmith/csmith03: arch=ppc bitw=32 bug lang=c os=macos solver=boolector solver=z3
```

This should be the final step before we could start using these QUERY-like formulas over TC-flags in the TCs themselves, in order to start specifying compatibility. Additionally, in the future, I'd like to use this tool in the CMake build process to identify those regression tests, that are compatible with the current build options. For instance, disabling certain solvers would exclude those solver-specific TCs from the CTest configuration. But that's outside of this PR.


rafaelsamenezes:
> QUERY is an infix formula (! or - for negation, & or space for conjunction, | for disjunction) 

I would prefer for only one prefix operator per operation. Is there any reason for using more than one?

I am not sure about a use case for factor producing formulas. For example, `(solver=z3 & (strat=kind | arch=ppc) | solver=btor)` seems overkill.

fbrausse:
> I would prefer for only one prefix operator per operation. Is there any reason for using more than one?

The reason was to avoid quoting on the shell. ;)

> I am not sure about a use case for factor producing formulas. For example, `(solver=z3 & (strat=kind | arch=ppc) | solver=btor)` seems overkill.

It's just standard infix notation, as used in arithmetic as well. Without it, there couldn't be parenthesized sub-expressions, it would be required to put queries in DNF. I agree that your example - while valid with this PR - indeed doesn't make a lot of sense.

So would your suggestion be to only allow disjunctions for these categorized flags containing `=`?

I'm currently trying to figure out which kind of queries the CMake build would need to construct and for that I thought it best to have full flexibility with the query language. Didn't want to touch CMake in this PR, though. But if you think it's too much, we can keep it a draft and I'll look into the CMake requirements here.

rafaelsamenezes:
> I'm currently trying to figure out which kind of queries the CMake build would need to construct and for that I thought it best to have full flexibility with the query language.

I am just wondering that for a full fledged querying language, there might be a library that does that for us.

>  But if you think it's too much, we can keep it a draft and I'll look into the CMake requirements here.

I don't know if it's too much. I am not even sure what we want to achieve with the queries. If it's just for filters, the disjunctions at category should be enough.

fbrausse:
> I am not even sure what we want to achieve with the queries. If it's just for filters, the disjunctions at category should be enough.

I see. Well, ideally, I was hoping we could achieve three things with it:
1. CMake builds a query string matching the build config, passes it to regdb.py and retrieves the set of regression tests that are supported by the build. Unsupported TCs could be for instance because solvers or libraries (ibex) missing, or because a 32-bit environment/sysroot is not available.
2. Use the flags of the supported regression tests to attach CTest labels s.t. running ctest on just the subset identified by the CTest labels becomes feasible. The benefit of using regdb.py for that would be consistency with the determination of a test is supported at all.
3. Adding a QUERY to the first line of test.desc after the test's label. This query would restrict how the test can be run (e.g. `-os=win`).

It is my hope that items 1. and 3. together allow us to get rid of the rigid directory structure in the `regression` subdirectory.

*Edit:* It would also be a step towards #1581: the missing sysroot would be identified by CMake and thus the failing regression test would be disabled.

mikhailramalho:

PR #1691, opened 2024-02-19
[C++ operational models] fixed compilation errors and warning messages in the <queue>lib 

lucasccordeiro:
This PR does the following:

(1) added a space between consecutive right angle brackets.
(2) removed in-class initialization of non-static data members.
(3) enabled various regression tests from the `<queue>` library.


src/cpp/library/queue
```diff
@@ -18,9 +18,9 @@ template <
 class priority_queue
 {
   T buf[QUEUE_CAPACITY];
-  int _size = 0;
-  T head = 0;
-  T tail = 0;
+  size_t _size;
+  T head;
+  T tail;
```
mikhailramalho, line 23: any specific reason to remove the init value?

lucasccordeiro: The constructor method already initialises these attributes. We can keep them, but we will need to check C++ version. This type of initialisation is allowed from C++11.

fbrausse:
As a general comment here: I think we should find out whether it makes sense to support C++03 and below. It will be a major source for #ifdefs, because of the new move-constructor and -assignment, and generally everything involving rvalue-references. If we can avoid that by saying "C++11 and above", I think we could save a lot of time. That includes the "space between > >" as well: starting with C++11 it's fine to have ">>" instead (and Clang obviously supports it).

XLiZHI:
According to https://en.cppreference.com/w/cpp/container/queue,  the support of `<queue>` header file started with C++ 98. The way I thought about it was to use the macro `__cplusplus`, but that would be a lot more work. Or can we parse this header independently in C++ 11?

lucasccordeiro:
> can we parse this header independently in C++ 11?

We can parse this header independently in C++11. It just produces warning messages.

The soteria project is currently funding the development of our Clang-C++ frontend. The goal is to formally verify the JDK, which mixes C++98 and C++11.

For now, I believe the best way is to use the `ifdef`. Later, we can remove this `ifdef` and support only the recent version as suggested by @fbrausse, i.e., "C++11 and above".

I hope you guys are okay with this solution.

src/cpp/library/queue
```diff
@@ -93,13 +99,19 @@ public:
...
-  int tail = 0;
+#if __cplusplus >= 201103L // Check if C++11 or later is being used
+  size_t _size = 0;
+  size_t head = 0;
+  size_t tail = 0;
+#else  
+  size_t _size;
+  size_t head;
+  size_t tail;
+#endif
```
rafaelsamenezes: Can't this be unified? The class constructor already initializes those values to 0

src/cpp/library/queue
```diff
@@ -14,13 +14,19 @@ namespace std
...
   T buf[QUEUE_CAPACITY];
-  int _size = 0;
+#if __cplusplus >= 201103L // Check if C++11 or later is being used
+  size_t _size = 0;
   T head = 0;
   T tail = 0;
+#else  
+  size_t _size;
+  T head;
+  T tail;
```
rafaelsamenezes, line 23: Same as below

rafaelsamenezes (changes requested):

src/cpp/library/queue
lucasccordeiro: This is what I mentioned before :-)

See my comments: "The constructor method already initialises these attributes."

However, @mikhailramalho and @fbrausse had another suggestion :-)

rafaelsamenezes (approved):

src/cpp/library/queue
lucasccordeiro: Fixed.

PR #1711, opened 2024-02-23
Fixes for interval bitwise arithmetic

rafaelsamenezes:
This PR fixes the bitwise operations for Integer intervals by:

1. Fixed the computation of whether we are dealing with signed or unsigned type. This is done by adding an optional type parameter into the integer intervals.
2. Removed the incorrect bitnot implementation.

| ID  | Mode | CT  | CF  | IT  | IF  | Notes |
| --- | --- | --- | --- | --- | --- | --- |
| 797 | No Arithmetic | 2657 | 1602 | 6   | 0   | Baseline |
| 802 | Arithmetic | 2194 | 1228 | 3   | 0   | The new unique were most in product-lines. |
| 828 | Arithmetic + Bitwise | 2195 | 1225 | 3   | 0 | Same as 802 |

### DIFF
--- a/src/goto-programs/abstract-interpretation/bitwise_bounds.h
+++ b/src/goto-programs/abstract-interpretation/bitwise_bounds.h
@@ -987,20 +987,12 @@ INT_X unsigned_2_signed_min_extend(UINT_X a, UINT_X b, UINT_X n)
   return a;
 }
 
-#define IS_UINT(LHS, RHS)                                                      \
-  LHS.get_lower().is_uint64() && LHS.get_upper().is_uint64() &&                \
-    RHS.get_lower().is_uint64() && RHS.get_upper().is_uint64()
-
 #define UINT_FUNC(FUNC, LHS, RHS)                                              \
   FUNC(                                                                        \
     LHS.get_lower().to_uint64(),                                               \
     LHS.get_upper().to_uint64(),                                               \
     RHS.get_lower().to_uint64(),                                               \
     RHS.get_upper().to_uint64())
-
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):
LGTM.

fbrausse (commented):
Is there a reason to *not* have the underlying type for intervals derived from other kinds of expressions? To my mind, having it optional complicates reasoning about the semantics. In any case, documentation what it means when the `optional<type2tc>` is not set would be good to have.

fbrausse:
PS: in case you don't need to differentiate between an empty type2tc and an empty optional type2tc, I'd suggest to only use the empty type2tc and remove the optional around it.

rafaelsamenezes:
> Is there a reason to *not* have the underlying type for intervals derived from other kinds of expressions? To my mind, having it optional complicates reasoning about the semantics. In any case, documentation what it means when the `optional<type2tc>` is not set would be good to have.

Not really. The thing is: I am currently trying to figure out whether the integer domain is worth keeping. My initial tests show that the wrapped domain is always better (which has the non optional type that you mention.

And since I am fixing bugs right now. I feel like revamps should wait until after everything is sorted. Besides arithmetic, there are still bugs in the widening.

fbrausse:
Sorry, I think you may have misunderstood. By "empty type2tc" I didn't mean the equivalent to `void`, aka empty_type2t, but just an uninitialized type2tc, if this does make sense.

fbrausse (approved):
Thanks for addressing my comment. LGTM.

lucasccordeiro:

PR #1719, opened 2024-02-27
Update Z3 version

rafaelsamenezes:
Should improve #1698


lucasccordeiro (approved):

fbrausse (commented):
Shall we make the non-Windows build default to the same version as the Windows one?

rafaelsamenezes:
> Shall we make the non-Windows build default to the same version as the Windows one?

Just added it

fbrausse (approved):

PR #1720, opened 2024-02-27
[Coverage] Support Condition Coverage

ChenfengWei0:
This PR aims to support condition coverage in ESBMC. Most importantly, we should be able to support short-circuit evaluation.

example 1
```c++
int main()
{
  int a = 1;
  int b = 2;
  if ((a == 1 && b == 2 || a == b && true))
  {
    assert(1);
  }
  else
    assert(0);
}
```

result
```sh
[Coverage]

  !(b == 2)     file file.c line 14 column 3 function main : SATISFIED
  b == 2        file file.c line 14 column 3 function main : UNSATISFIED
  !(a == 1)     file	 file.c line 14 column 3 function main : SATISFIED
  a == 1        file file.c line 14 column 3 function main : UNSATISFIED
Condition Coverage: 50%
```
As we can see, the guard expressions `a == b` and `true` have been short-circuited

---

example 2
```c++
int main()
{
  int a = 1;
  int b = 2;
  if (a == 1 && (b == 2 || a == b) && 1.1)
  {
    assert(1);
  }
  else
    assert(0);
}
```

result
```sh
[Coverage]

  !(b == 2)     file file.c line 14 column 3 function main : SATISFIED
  b == 2        file file.c line 14 column 3 function main : UNSATISFIED
  !(a == 1)     file file.c line 14 column 3 function main : SATISFIED
  a == 1        file file.c line 14 column 3 function main : UNSATISFIED
  1 != 1        file file.c line 14 column 3 function main : SATISFIED
  !(1 != 1)     file file.c line 14 column 3 function main : UNSATISFIED
Condition Coverage: 75%
```

---	

example 3
```c++
int main()
{
  int a = 1;
  int b = 2;
  while ((a == 1 && b == 3) || (a != b && true))
  {
    assert(1);
  }
}
```

result
```
[Coverage]

  !(b == 3)     file file.c line 30 column 3 function main : SATISFIED
  b == 3        file file.c line 30 column 3 function main : UNSATISFIED
  a != b        file file.c line 30 column 3 function main : SATISFIED
  !(a != b)     file file.c line 30 column 3 function main : UNSATISFIED
  !(a == 1)     file file.c line 30 column 3 function main : SATISFIED
  a == 1        file file.c line 30 column 3 function main : UNSATISFIED
  1 == 1        file file.c line 30 column 3 function main : SATISFIED
  !(1 == 1)     file file.c line 30 column 3 function main : UNSATISFIED
Condition Coverage: 100%
```
no short-circuit in this case

---
Command
```sh
  --condition-coverage                  this activates --multi-property,
                                        deactivates --keep-verified-claims, and
                                        shows the coverage of condition
                                        statements (short-circuit evaluation)
  --condition-coverage-claims           enable condition-coverage and show the
                                        instrumented claims
```


fbrausse:
I am not entirely sure why so much string manipulation is necessary to be performed by goto_coveraget. Couldn't the expressions be manipulated instead? Also note that the from_expr() output is typically meant for human consumption, that is, it format is not necessarily fixed and could change - possibly breaking the implementation in this PR.

Also, there should be no need to run id.as_string() on an irep_idt if you're just comparing it to other strings. In fact, comparing two irep_idts is almost always faster than string comparisons (in case both sides already exist as an irep_idt), because an irep_idt is just a fancy `int`. That's why we have the many named constants in irept, typet and exprt.

In addition, I am unhappy about the static variables goto_coveraget is manipulating. Could these be moved inside the goto_coveraget object? We're generally trying to get rid of modified global state as much as possible, see #755. This is for multiple reasons, most prominently to avoid the static initialization order fiasco of C++, but also because it makes multi-threading and reasoning about this global state very hard. *Edit:* and unit-testing.

ChenfengWei0:
> I am not entirely sure why so much string manipulation is necessary to be performed by goto_coveraget. Couldn't the expressions be manipulated instead? 

The main idea is to add and read the parenthesis. For instance, consider the symbol table of `a == 1 && (b == 1 || a != b)` :

```
and
  0: =
     0: symbol
	        * identifier: c:file.c@457@F@main@a
     1: constant
  1: or
     0: =
        0: symbol
               * identifier: c:file.c@472@F@main@b
        1: constant
     1: notequal
	 0: symbol
		* identifier: c:file.c@457@F@main@a
	 1: symbol
		* identifier: c:file.c@472@F@main@b
```

As we can see, there is no parenthesis in the symbol table. Surely we canto infer somehow. But I just read it from the string for convenience. Additionally, we will also add the implied parenthesis:

```
a && b || c && d => (a && b) || (c && d)
```

This is also done by string manipulation.

> Also note that the from_expr() output is typically meant for human consumption, that is, it format is not necessarily fixed and could change - possibly breaking the implementation in this PR.

The result of `from_expr()` will be used as the `claim_msg` in multi-property checking. If the `from_expr` changes, it might not only break this PR, but may break the normal multi-property checking.
```c++
if (!is_goto_cov)
  // obtain the guard info from the assertions
  claim_msg = from_expr(ns, "", it->source.pc->guard);
else
  // in goto-coverage mode, the assertions are converted to assert(0）
  // the original guards are stored in comment.
  claim_msg = it->comment;
```

> Also, there should be no need to run id.as_string() on an irep_idt if you're just comparing it to other strings. 

Will change. Thanks for the suggestion.

>Could these be moved inside the goto_coveraget object? We're generally trying to get rid of modified global state as much as possible, see #755.

The reason is that we want to store some value in the goto-coverage process and later read the value in another file, i.e. `bmc.cpp`.  E.g. `total_cond_assert = goto_coveraget(ns).get_total_cond_assert();`. That's why we use the static variable here, to share among all instances of that class

fbrausse:
There is a reason why we represent everything as expression trees: ability to manipulate them and extract information. Please do consider manipulating those trees (and afterwards, when everything is ready, dumping them to the string the coverage check expects). I understand that switching from_expr() output from e.g. C to something else might break other parts of ESBMC, but that doesn't mean we should cement this C output. Over the past few years we've been expanding ESBMC to cover more languages. Thus the from_expr() format cannot be relied upon in the future - unlike the expression trees. Those are part of ESBMC's data structures and they are much easier to adapt to changes than processing arbitrary strings.

Regarding the statics: I don't understand why the goto_coveraget object containing `total_cond_assert`, `total_assert_instance` and `total_instrument` as members couldn't be passed on to whatever code is using those values as an object. Why do those values need to be non-constant statics? That's not an extendible design choice and unless there are technical reasons we shouldn't do that. I will repeat: Modifiable static objects incur problems:
- [static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof)
- multi-threading
- reasoning about this global state
- unit-testing

lucasccordeiro:
> I understand that switching from_expr() output from e.g. C to something else might break other parts of ESBMC, but that doesn't mean we should cement this C output. Over the past few years we've been expanding ESBMC to cover more languages. Thus the from_expr() format cannot be relied upon in the future - unlike the expression trees. Those are part of ESBMC's data structures and they are much easier to adapt to changes than processing arbitrary strings.

I agree with @fbrausse.

ChenfengWei0:
I think the `from_expr` issue could be addressed in another PR... The main reason is that the current multi-property is also highly relied on from_expr. So if we do not use the `from_expr` in the middleware, we will eventually use it in the backend.
```c++
// obtain the guard info from the assertions
claim_msg = from_expr(ns, "", it->source.pc->guard);
```
So if we do not want to use `from_expr`, we should clean them all in other PR. However, I do not know how to do that.

ChenfengWei0:
> I think the `from_expr` issue could be addressed in another PR... The main reason is that the current multi-property is also highly relied on from_expr. So if we do not use the `from_expr` in the middleware, we will eventually use it in the backend.
> 
> ```c++
> // obtain the guard info from the assertions
> claim_msg = from_expr(ns, "", it->source.pc->guard);
> ```
> 
> So if we do not want to use `from_expr`, we should clean them all in other PR. However, I do not know how to do that.

Hi @fbrausse, would you be happy with this idea?

fbrausse:
> The main reason is that the current multi-property is also highly relied on from_expr. So if we do not use the `from_expr` in the middleware, we will eventually use it in the backend.

We can use it whereever we like, as long as the consumer of that string is aware of the semantics. See below.

> So if we do not want to use `from_expr`, we should clean them all in other PR. However, I do not know how to do that.

from_expr() by default produces strings meant for human consumption https://github.com/esbmc/esbmc/blob/1813a9cbbe2c0016d79d310935d22a1f48abda8d/src/langapi/language_util.h#L18-L22 in to the input language that the symbol was defined in: https://github.com/esbmc/esbmc/blob/1813a9cbbe2c0016d79d310935d22a1f48abda8d/src/langapi/language_util.cpp#L37-L39

It has a second mode of operation, selected by the `target` parameter, which is for witness files. So we do want to continue using from_expr() where the target consumer of the string matches the `target` parameter. Arguably, the slicer code you quoted already doesn't do this as the `claim_msg`, among other things, is appearently being used by the multi-property check to avoid checking assertions multiple times.

What this PR is doing, is hard-coding the following expectations on strings produced by from_expr():
- always C-like output (which as an implementation detail for identifier `""` is true, but not documented anywhere nor is there a reason to stay that way), and
- a certain way of putting parenthesis (which also cannot be relied upon, in particular for formats meant for humans: compilers for instance warn about missing parenthesis for `a && b || c` even though the language C is perfectly happy with that construct).

As far as I can tell, these two expectations are new with this PR and they're not assumed elsewhere. In essence, this PR is parsing the from_expr() output meant for human consumption, which is new and in my opinion should not be done. Note that the `claim_msg` example is not parsing the string.

> I think the `from_expr` issue could be addressed in another PR... [...] would you be happy with this idea?

Doing that would add technical debt in the sense that with this PR the from_expr() format for `""` and boolean expressions would largely be set in stone, which I'm personally not particularly happy with.

I'll add review remarks.

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+      forall_goto_program_instructions (it, goto_program)
+      {
+        if (
+          it->is_assert() &&
+          it->location.property().as_string() == "instrumented assertion" &&
+          it->location.user_provided() == true)
+        {
+          idf = it->location.comment().as_string() + "\t" +
+                it->location.as_string();
+          total_cond_assert.insert(idf);
```
fbrausse: If this string conceptually is supposed to represent a pair (comment, location), could we make it such? Perhaps a std::pair or an own class that even gives these two fields proper identifiers? Alternatively, could you store `it` directly?

src/esbmc/bmc.cpp
```diff
@@ -853,5 +859,109 @@ smt_convt::resultt bmct::multi_property_check(
...
+          log_status("  {} : SATISFIED", claim);
+        tracked_instance++;
+        reached_claims.erase(claim);
+        total_cond_assert_cpy.erase(claim);
+
+        // reversal
+        std::string delimiter = "\t";
+        auto pos = claim.find(delimiter);
+        std::string claim_msg = claim.substr(0, pos);
+        std::string claim_loc = claim.substr(pos + 1);
```
fbrausse: This code deconstructs the conceptual "pair" from the above remark and makes the assumption that `"\t"` is unique. Both would not be needed if it was a pair (or the iterator into the goto-instructions) in the first place.

src/goto-symex/goto_symex_state.h
```diff
@@ -311,7 +311,8 @@ class goto_symex_statet
    */
   inline void pop_frame()
   {
-    assert(call_stack.back().goto_state_map.size() == 0);
+    //! temporary fix for the condition coverage
+    //! assert(call_stack.back().goto_state_map.size() == 0);
```
fbrausse: Why is disabling this assertion necessary?

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
 // run the algorithm on the copy of the original goto program
-void goto_coveraget::gen_assert_instance(goto_functionst goto_functions)
+int goto_coveraget::get_total_assert_instance() const
 {
   // 1. execute goto uniwnd
+  int total_assert_instance = 0;
   bounded_loop_unroller unwind_loops;
   unwind_loops.run(goto_functions);
   // 2. calculate the number of assertion instance
+  forall_goto_functions (f_it, goto_functions)
```
fbrausse: This entire loop is basically copy of goto_coveraget::get_total_instrument(), right? Could you unify them?

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+    }
+  return total_assert_instance;
+}
+
+std::unordered_set<std::string> goto_coveraget::get_total_cond_assert() const
+{
+  std::unordered_set<std::string> total_cond_assert = {};
+  std::string idf = "";
+
+  forall_goto_functions (f_it, goto_functions)
```
fbrausse, line 125: Same loop again, filtering out particular assertions. Let's avoid duplicating code.

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
       Forall_goto_program_instructions (it, goto_program)
       {
-        if (it->is_assert())
-          total_assert_instance++;
+        const std::string cur_filename =
+          get_filename_from_path(it->location.file().as_string());
+        filename = get_filename_from_path(filename);
+
+        // e.g. IF !(a > 1) THEN GOTO 3
+        if (!is_true(it->guard) && it->is_goto() && filename == cur_filename)
```
fbrausse: What is the purpose of this filename comparison?

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+{
+  const std::string &id = expr.id().as_string();
+
+  if ((id == "and" || id == "or") && flag == false)
+  {
+    bool flg0 = false;
+    collect_operands(expr.op0(), operands, flg0);
+    if (!flg0)
+      operands.emplace_back(expr.op0());
+    // operators.emplace_back(id);
```
fbrausse: Remove commented code, please.

src/goto-programs/goto_coverage.h
```diff
@@ -1,30 +1,71 @@
...
   int get_total_assert_instance() const;
+  std::unordered_set<std::string> get_total_cond_assert() const;
+  std::string get_filename_from_path(std::string path);
 
 protected:
-  static int total_instrument;
-  static int total_assert_instance;
+  // turn a OP b OP c into a list a, b, c
+  static void
+  collect_operands(const exprt &expr, std::list<exprt> &operands, bool &flag);
```
fbrausse: Could you please document what this function is supposed to do? Appearently, from the code, it collects operands of `and` and `or` subexprs. Does it handle negation? Or other logical connectives?

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+  // e.g. if(return_bool()) ==> IF !(_Bool)return_value$_return_bool$1
+  //                        ==> IF !return_value$_return_bool$1
+  // !FIXME: maybe this should be regex like \(\_[A-Za-z]+\) iF there are other typecasts
+  const std::string subStr = "(_Bool)";
+  size_t pos = str.find(subStr);
+  while (pos != std::string::npos)
+  {
+    str.erase(pos, subStr.length());
+    pos = str.find(subStr, pos);
+  }
```
fbrausse: Hard-coded C assumption. Typically, in ESBMC, if we want to not process (certain) typecasts, there would be a recursive function applied to an expression tree, which replaces those typecasts with the result of the recursive function applied to the child node. Depending on what the code below does (I see lots of string manipulation), it might be a good idea to implement such a recursive function modifying an expression tree.

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+
+  std::list<std::string> opt;
+  for (std::size_t i = 0; i < str.length(); i++)
```
fbrausse: What does this loop do?

src/goto-programs/goto_coverage.cpp
```diff
@@ -38,64 +62,666 @@ void goto_coveraget::add_false_asserts(goto_functionst &goto_functions)
...
+    else if (str[i] == ')')
+      opt.emplace_back(")");
+  }
+
+  // add implied parentheses in boolean expression
+  // e.g.
+  //   if(a&&b || c&&d) ==> if((a&&b) || (c&&d))
+  //   (a == 1 || b != 2 && (b == 3 || a == 1)) => (a == 1 || (b != 2 && (b == 3 || a == 1)))
+  // general rule: add parenthesis between || and &&
+  // (&&||&&)
```
fbrausse: 1. For what purpose do you add "implied parentheses"?
2. Does it work on constant_string expressions as well, e.g. `"a &&"[0] == 'a'`?
3. In general, again, hard-coded C assumption. Symbols could contain `(` just fine.

src/goto-programs/goto_coverage.h
```diff
@@ -1,30 +1,71 @@
...
+  static void
+  collect_operators(const exprt &expr, std::list<std::string> &operators);
```
fbrausse: Same, could you please document what this function is supposed to do, what kind of expressions it expects and what its result is? From the code it's not entirely clear.

fbrausse (changes requested):
I've not gone through all of it, yet, because frankly I'm having trouble making sense of the string manipulations and thus would like some more documentation.

src/goto-symex/goto_symex_state.h
ChenfengWei0: In some situations, this assertion would fail and obstruct the output. I found the only way to fix it is to comment this assertion.

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Will change

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Will change to a std::pair

src/goto-programs/goto_coverage.h
ChenfengWei0: Will add documentation. It does not collect operands of negation

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Collect operators. Will add documentation

src/goto-programs/goto_coverage.cpp
ChenfengWei0: I agree. But it might be time-consuming to do that. I am afraid I cannot implement that in this PR.

src/goto-programs/goto_coverage.cpp
ChenfengWei0: 1. Implied parentheses in a boolean expression refer to the implicit order of operations dictated by the precedence and associativity rules of boolean operators. E.g. `&&` > `||`.
2. Not sure. I guess it could. Need to do some experiment

src/goto-symex/goto_symex_state.h
fbrausse: My best guess right now would be because you're actually altering the `symex->goto_functions`: They're passed by-reference to the goto_coveraget constructor, stored in that class also by (modifyable) reference, and subsequently modified, e.g. by the bounded_loop_unroller. Maybe elsewhere, too.

It would be good to understand why this assertion did exist in the first place and why this PR is changing the semantics of the goto_symex_statet.

src/goto-programs/goto_coverage.h
fbrausse: > It does not collect operands of negation

Are you sure? What about `!(a && b)`? My understanding of the code is, that `a` and `b` get collected, but the negation gets dropped. They are part of the operand of the negation.

On that example: what operator(s) would collect_operators() collect here?

lucasccordeiro:
@ChenfengWei0: Can I ask you whether you had a chance to address Franz's comments?

ChenfengWei0:
> @ChenfengWei0: Can I ask you whether you had a chance to address Franz's comments?

I will resolve the issues after finishing the solidity mapping.

lucasccordeiro (approved):

src/goto-symex/goto_symex_state.h
fbrausse: Seems to not be the case anymore.

fbrausse (approved):
Looks much better, many thanks for addressing my concerns.

lucasccordeiro:
Many thanks for submitting this PR, @ChenfengWei0!

That's a great piece of work added to the ESBMC project!

Many congratulations!

intrigus-lgtm:
Could you kindly explain in depth what condition coverage is supposed to be?
For example, consider this code:
```
#include <assert.h>
#include <stdbool.h>

int b;
int main() {
int a = nondet_int();
  if (a == 1 && b == 2) { // |1|
//  if ((a == 1 && b == 2) != 0) { // |2|
    assert(0 && "if");
  } else
    assert(0 && "else");
}
```

If you run this with `--condition-coverage --condition-coverage-claims --no-cov-asserts` and |1| in place, you get this as result:
```
[Coverage]

  !(a == 1)	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 8 column 3 function main : SATISFIED
  a == 1	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 8 column 3 function main : SATISFIED
  !(b == 2)	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 8 column 3 function main : SATISFIED
  b == 2	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 8 column 3 function main : UNSATISFIED
Reached Conditions:  4
Total Conditions:  4

Condition Properties - SATISFIED:  3
Condition Properties - UNSATISFIED:  1

Condition Coverage: 75%
```

But if you comment |1| and uncomment |2|, you get this as result:
```
[Coverage]

  !(((signed int)(a == 1 && b == 2)) != 0)	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 9 column 3 function main : SATISFIED
  ((signed int)(a == 1 && b == 2)) != 0	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 9 column 3 function main : UNSATISFIED
Reached Conditions:  2
Total Conditions:  2

Condition Properties - SATISFIED:  1
Condition Properties - UNSATISFIED:  1

Condition Coverage: 50%
```

What is the expected result? How many conditions are there to cover?

What about conditions like `if(1 == 1 && 2 == 2)`? Should those be considered conditions?

intrigus-lgtm:
@ChenfengWei0  I would greatly appreciate some insight into this as I was trying to improve some code which resulted in regression tests related to goto-coverage to fail.

lucasccordeiro:
@ChenfengWei0 : you could extend the ESBMC documentation to add @intrigus-lgtm's suggestion: https://ssvlab.github.io/esbmc/documentation.html. That's important!

ChenfengWei0:
> But if you comment |1| and uncomment |2|, you get this as result:
> 
> ```
> [Coverage]
> 
>   !(((signed int)(a == 1 && b == 2)) != 0)	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 9 column 3 function main : SATISFIED
>   ((signed int)(a == 1 && b == 2)) != 0	file /home/sg/dev/esbmc-contributing/regression/goto-coverage/github_1720_1/main.c line 9 column 3 function main : UNSATISFIED
> Reached Conditions:  2
> Short Circuited Conditions:  0
> Total Conditions:  2
> 
> Condition Properties - SATISFIED:  1
> Condition Properties - UNSATISFIED:  1
> 
> Condition Coverage: 50%
> ```


@intrigus-lgtm This looks like a bug to me. 

- `cbmc 1.c --cover condition --timestamp wall --unwind 2 --no-standard-checks`

```
** coverage results:
2025-03-20T10:07:32.532714 [main.coverage.1] file 1.c line 9 function main condition '(signed int)(a == 1 && b == 2) != 0' false: FAILED
[main.coverage.2] file 1.c line 9 function main condition '(signed int)(a == 1 && b == 2) != 0' true: SATISFIED
[main.coverage.3] file 1.c line 9 function main condition 'a == 1' false: SATISFIED
[main.coverage.4] file 1.c line 9 function main condition 'a == 1' true: SATISFIED
[main.coverage.5] file 1.c line 9 function main condition 'b == 2' false: FAILED
[main.coverage.6] file 1.c line 9 function main condition 'b == 2' true: SATISFIED
[main.coverage.7] file 1.c line 11 function main condition '(__CPROVER_bool)"if"' false: FAILED
[main.coverage.8] file 1.c line 11 function main condition '(__CPROVER_bool)"if"' true: FAILED
[main.coverage.9] file 1.c line 11 function main condition '0 != 0' false: FAILED
[main.coverage.10] file 1.c line 11 function main condition '0 != 0' true: FAILED
[main.coverage.11] file 1.c line 14 function main condition '(__CPROVER_bool)"else"' false: SATISFIED
[main.coverage.12] file 1.c line 14 function main condition '(__CPROVER_bool)"else"' true: FAILED
[main.coverage.13] file 1.c line 14 function main condition '0 != 0' false: FAILED
[main.coverage.14] file 1.c line 14 function main condition '0 != 0' true: SATISFIED

2025-03-20T10:07:32.532776 ** 6 of 14 covered (42.9%)
```

Since we need to align the CBMC output, I believe the conditions `a == 1` and `b == 2` need to be evaluated.

Issue #1582, opened 2024-01-08
[Concurrency] Incorrect false for sv-benchmarks/c/weaver/chl-poker-hand-symm.wvr.c


lucasccordeiro:
````
./esbmc-wrapper.py -p ../../sv-benchmarks/c/properties/unreach-call.prp -s kinduction --arch 32 ../../sv-benchmarks/c/weaver/chl-poker-hand-symm.wvr.c


--------------------------------------------------------------------------------


Verifying with ESBMC
Command: ./esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check ../../sv-benchmarks/c/weaver/chl-poker-hand-symm.wvr.c --32  --no-por --context-bound 2 --witness-output witness.graphml --enable-unreachability-intrinsic --no-pointer-check --no-bounds-check --incremental-bmc 
ESBMC version 7.4.0 64-bit x86_64 linux
Target: 32-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing ../../sv-benchmarks/c/weaver/chl-poker-hand-symm.wvr.c
Converting
Generating GOTO Program
GOTO program creation time: 0.139s
GOTO program processing time: 0.003s
Checking base case, k = 1
Starting Bounded Model Checking
Not unwinding loop 6 iteration 1   file chl-poker-hand-symm.wvr.c line 245 column 3 function create_fresh_int_array
Symex completed in: 0.002s (52 assignments)
Slicing time: 0.000s (removed 50 assignments)
Generated 0 VCC(s), 0 remaining after simplification (2 assignments)
BMC program time: 0.002s
No bug has been found in the base case
Checking forward condition, k = 1
Starting Bounded Model Checking
Not unwinding loop 6 iteration 1   file chl-poker-hand-symm.wvr.c line 245 column 3 function create_fresh_int_array
Symex completed in: 0.002s (52 assignments)
Slicing time: 0.000s (removed 50 assignments)
Generated 1 VCC(s), 1 remaining after simplification (2 assignments)
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
Encoding to solver time: 0.000s
Solving with solver Boolector 3.2.2
Runtime decision procedure: 0.000s
The forward condition is unable to prove the property
.....
[...]

PR #1721, opened 2024-02-27
[Concurrency] Make interleavings unviable after main thread has ended.

Anthonysdu:
ESBMC does not verify the VCCs after the main thread has ended, however, it is still exploring them, we can stop the further interleavings to reduce states space.

The results of this PR: https://github.com/esbmc/esbmc/actions/runs/8048711271/job/21981301419

```
  correct:             427
    correct true:      142
    correct false:     285
  incorrect:             5
    incorrect true:      4
    incorrect false:     1
  unknown:             233
  Score:               425 (max: 999)
```

master: 

```
  correct:             421
    correct true:      138
    correct false:     283
  incorrect:             5
    incorrect true:      4
    incorrect false:     1
  unknown:             239
  Score:               415 (max: 999)
```

The incorrect false is at issue #1582 . (Need to check whether it's correctly labelled, cbmc also report failure)

### DIFF
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -263,6 +263,7 @@ void execution_statet::symex_step(reachability_treet &art)
       // TODO: we should support verifying memory leaks in multi-threaded C programs.
       assume(gen_false_expr());
       end_thread();
+      interleaving_unviable = true;
     }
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

Issue #1028, opened 2023-05-21
[parallel-solving] segmentation fault


lucasccordeiro:
ESBMC produces a segmentation fault if we run the following command:

 esbmc [main.c](https://github.com/esbmc/esbmc/blob/master/regression/esbmc-unix/05_pfscan-1.0_01/main.c) --incremental-bmc --32 --color --parallel-solving

````
Slicing for Claim dereference failure: Incorrect alignment when accessing data object (0.000s)
Slicing time: 0.007s (removed 444 assignments)
No solver specified; defaulting to Boolector
Slicing time: 0.008s (removed 444 assignments)
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
Segmentation fault
````

**Comment by ChenfengWei0** [comment]
Hmm... I had the #1011 problem when running this test, with or without `--parallel-solving`

```sh
esbmc main.c --incremental-bmc --32

...
ERROR: Looking up index of nonexistant member "__lock" in struct/union "pthread_mutex_t"
[1]    6092 IOT instruction  ../../../../release/bin/esbmc main.c --incremental-bmc --32
```

I think we should also test the other two options (`--multi-property` and `--multi-fail-fast`) , to see if they work properly

fbrausse:
There is a race condition when hashing expr2t's for the `smt_cache` in smt_conv. Most likely the reason is that all parallel jobs refer to the same expr2tc objects as members of `symex_target_equationt::SSA_stept`.

*Edit:* for debugging this particular example for now you can use #823. The race condition however exists independently of whether the program uses pthreads.

ChenfengWei0:
> There is a race condition when hashing expr2t's for the `smt_cache` in smt_conv. Most likely the reason is that all parallel jobs refer to the same expr2tc objects as members of `symex_target_equationt::SSA_stept`.
> 
> _Edit:_ for debugging this particular example for now you can use #823. The race condition however exists independently of whether the program uses pthreads.

Thanks for your help!

ChenfengWei0:
Based on the #832 
- `--multi-property` (the non-parallel-solving) works fine
- `--parallel-sovling` raises assertion failure

```sh
esbmc: /root/workspace/ESBMC_Commit/esbmc/src/util/string_container.h:68: const std::string &string_containert::get_string(size_t) const: Assertion `no < string_vector.size()' failed.
[...]

PR #1725, opened 2024-03-01
Investigating std::shared_ptr overhead

fbrausse:
The std::shared_ptr is quite flexible, allowing for custom deleters, allocators, and for casting between nearly arbitrary base-classes of the types of the objects being tracked while requiring no changes to the types themselves. It is also thread-safe (except for its use_count() method).

This flexibility comes with a cost. This PR is trying to figure out the cost. For this purpose, I've moved irep2 from being based on std::shared_ptr to being based on a custom type, that has less flexibility (and thus requires a bit more care when being used for arbitrary types).

In comparison to libstdc++'s std::shared_ptr implementation, the new ksptr::sptr one only requires storing a single pointer instead of two, and I've also externalized the control block (where the use-count and potentially deleter, etc. are being tracked).

First, for 30s SV-COMP there are not many changes. Master:
```
  correct:           13965
    correct true:     7709
    correct false:    6256
  incorrect:            31
    incorrect true:     13
    incorrect false:    18
  unknown:            9809
  Score:             20970 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/8102075164
```
This PR:
```
  correct:           13995
    correct true:     7717
    correct false:    6278
  incorrect:            32
    incorrect true:     13
    incorrect false:    19
  unknown:            9778
  Score:             20992 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/8110826297
```
(The "new" incorrect is noise.)

However, it seems that ESBMC is using about 10% less memory with this PR in the memory-intensive benchmarks. Running time is nearly unaffected, a reduction between 0 and maybe 3% in local experiments. I did expect more, since the use-count is not atomic anymore. Nevertheless, I could confirm that the 10% reduction in memory consumption is *only* due to the fact, that the ksptr::sptr stores a single pointer, while std::shared_ptr stores two of them internally.

I'm not sure whether this is a way we want to go. Atm. the ksptr::sptr isn't even pretending to be thread-safe, but that could be changed.

Any thoughts? Maybe also about what else to try?

### DIFF
--- a/src/irep2/irep2.h
+++ b/src/irep2/irep2.h
@@ -398,6 +398,8 @@ typedef std::list<member_entryt> list_of_memberst;
 
 class irep2t
 {
+public:
+  virtual ~irep2t() noexcept = default;
 };
 
 /** Base class for all types.
--- a/src/irep2/irep2.h
+++ b/src/irep2/irep2.h
@@ -396,7 +396,7 @@ typedef irep_container<expr2t> expr2tc;
 typedef std::pair<std::string, std::string> member_entryt;
 typedef std::list<member_entryt> list_of_memberst;
 
-class irep2t : public std::enable_shared_from_this<irep2t>
+class irep2t
 {
 };
 
--- a/src/irep2/irep2.h
+++ b/src/irep2/irep2.h
@@ -33,6 +33,8 @@
 #include <util/irep.h>
 #include <vector>
 
+#include <util/ksptr.hh>
+
 // Ahead of time: a list of all expressions and types, in a preprocessing
 // list, for enumerating later. Should avoid manually enumerating anywhere
 // else.
@@ -202,7 +204,7 @@ class constant_vector2t;
  *  this class provides.
  */
 template <class T>
-class irep_container : private std::shared_ptr<T>
+class irep_container : private ksptr::sptr<T>
 {
 public:
   constexpr irep_container() noexcept = default;
@@ -218,8 +220,8 @@ public:
   // Obviously this is fairly unwise because any std::shared_ptr
   // won't be using the detach facility to manipulate things, however it's
   // necessary for std::make_shared.
-  explicit irep_container(std::shared_ptr<T> &&p)
-    : std::shared_ptr<T>(std::move(p))
+  explicit irep_container(ksptr::sptr<T> &&p)
+    : ksptr::sptr<T>(std::move(p))
   {
   }
 
@@ -248,14 +250,14 @@ public:
 
   const T *get() const noexcept
   {
-    return std::shared_ptr<T>::get();
+    return ksptr::sptr<T>::get();
   }
 
   // the non-const versions detach
   T *get() // never throws
   {
     detach();
-    T *tmp = std::shared_ptr<T>::get();
+    T *tmp = ksptr::sptr<T>::get();
     tmp->crc_val = 0;
     return tmp;
   }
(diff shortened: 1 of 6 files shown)


mikhailramalho:
Interesting work!

Any reason why use_count is no longer atomic? I guess is not a problem currently since we fork the process instead of using threads in k-induction parallel

fbrausse:
> Interesting work!
> 
> Any reason why use_count is no longer atomic? I guess is not a problem currently since we fork the process instead of using threads in k-induction parallel

I was trying to get rid of as many features that we don't use. The k-induction-parallel indeed has separate processes, so it's not affected. Some time ago the multi-property check used to be multi-threaded, but that led to lots of race conditions (on the expr2t::crc_val, the irep_idt string pool, etc., #1028 has the known ones), so it was eventually removed. So in essence, we can't multithread ESBMC easily right now.

lucasccordeiro:
@fbrausse: Thanks for submitting this PR and starting this discussion.

Our main goal is to parallelize ESBMC. So, I'm entirely in favor of this initiative to make ESBMC thread-safe. The 10% reduction in memory consumption is also a great gain for this PR.

brcfarias:
Great initiative.
If the ownership of the pointers isn't truly shared among different components/threads, switching from shared_ptr to unique_ptr could be an option.
Perhaps it's worth taking a look at boost::intrusive_ptr and WebKit's WTF::RefPtr<T> as well.

fbrausse:
> Our main goal is to parallelize ESBMC. So, I'm entirely in favor of this initiative to make ESBMC thread-safe.

That has actually not been the goal with this PR. If anything, it's going into the opposite direction: Find out how much overhead there is in the current irep2 setup. Turns out, memory we can save, but appearently the atomicity of the ref-counter doesn't incur much overhead.

> The 10% reduction in memory consumption is also a great gain for this PR.

Thanks! Before we start doing that, I would like to take a moment to reflect though: If storing one additional pointer per irep2 object accounts for 10% of the total memory usage, we might want to investigate other ways to reduce the memory footprint.

We currently have 115 different expr2t and type2t classes. They store between 4 and 14 (with this PR) or between 5 and 17 (on master) pointers worth of data. All of them share
- vtable pointer
- a shared_ptr/ksptr::sptr use_count
- expr_id/type_id
- crc_val

From my point of view it's interesting to note that there is a functional dependency: vtable ptr -> expr_id/type_id. Putting the id behind the vtable ptr via a virtual method should save another 10% at the cost of a runtime dereference.

I might give this a try as well. We'll see. What I did try ontop of this PR is:
- Using free-lists based on the size of the irep2 objects and using typeid() to identify the type-erased objects. They didn't help a lot, neither with memory nor with timing. Some 0-4% for timing. There is lots of reuse, though: a few million reused irep2 objects are seen quite often.
- Ontop of the free-lists, I tried using an arena allocator only for irep2. This does not affect memory usage, but helps a bit with runtime, maybe another 1-5%.

All in all, those changes will gain 10% of memory and between 0 and 10% of runtime. This means, that symex being slow in comparison to CBMC has different reasons.

fbrausse:
> If storing one additional pointer per irep2 object accounts for 10% of the total memory usage, [...]
> Putting the id behind the vtable ptr via a virtual method should save another 10% at the cost of a runtime dereference.

I've looked at that and it's not true: It is the shared pointers which contribute the 10%, not the pointed-to irep2 objects themselves. In my tests for benchmarks consuming 6 GB memory, only about 20-50 MB of actual irep2 objects were alive simultaneously at any point in time. The difference is really just due to the shared pointer having half the size. So there's no point in reducing the size of irep2 objects here.

Turns out, that the interval analysis does store a lot of state in hash tables, among it appearently plenty of expr2tc/type2tc shared pointers. We should investigate whether it's possible to reduce the size of the state stored by the interval analysis. For instance, this benchmark goes from 350 MB to 6 GB when enabling the interval analyis:
- c/hardware-verification-bv/btor2c-lazyMod.telephony.3.prop1-back-serstep.c

In private conversation Rafael confirmed: These hardware benchmarks are indeed very slow with interval analysis.

fbrausse:
This investigation has shown two things, the first of which has largely been fixed:
- expr2tc and type2tc each containing 2 pointers can lead to lots of memory overhead, but the cause has been taken care of in #1755
- the atomicity of the shared counter does not factor into the symex running time measurably.

My conclusion from this: there's no need to change the basis of irep2. Symex runtime is hiding elsewhere. I'll close this.

PR #1726, opened 2024-03-01
Support for symbolic reconstruction of expression

rafaelsamenezes:
This adds a method to "flatten" an expression given the current program state. This should help doing analysis at Symex and SSA level, specifically Interval Analysis.

### DIFF
--- a/src/goto-symex/symex_target_equation.cpp
+++ b/src/goto-symex/symex_target_equation.cpp
@@ -396,8 +396,9 @@ void symex_target_equationt::replace_rec(
       e = step.rhs;
   }
 
-  e->Foreach_operand([&step, &keep_local, this](expr2tc &inner)
-                     { replace_rec(step, inner, keep_local); });
+  e->Foreach_operand([&step, &keep_local, this](expr2tc &inner) {
+    replace_rec(step, inner, keep_local);
+  });
 }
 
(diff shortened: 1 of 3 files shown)


src/goto-symex/symex_target_equation.cpp
```diff
@@ -379,6 +379,42 @@ unsigned int symex_target_equationt::clear_assertions()
...
+  e->Foreach_operand([&step, &keep_local, this](expr2tc &inner) {
+    replace_rec(step, inner, keep_local);
+  });
+}
+
+void symex_target_equationt::reconstruct_symbolic_expression(
+  expr2tc &expr,
+  bool keep_local_variables) const
+{
+  keep_local_variables = true;
```
fbrausse: This looks wrong here.

fbrausse (commented):
So you're now starting to add passes to the SSA. Very nice, I think we should have more of them. I did need quite a bit to understand the algorithm. Maybe it's already too late. One comment about keep_local_variables.

src/goto-symex/symex_target_equation.cpp
rafaelsamenezes: Ooops, nice catch. This was left from some debugging. It is getting late indeed.

fbrausse (approved):

lucasccordeiro (approved):

PR #1731, opened 2024-03-04
[cpp-clang] Support sizeof expression

intrigus-lgtm:
Adds support for sizeof expressions in the context of template packs:
> Represents an expression that computes the length of a parameter pack.
>```
> template<typename ...Types>
> struct count {
>  static const unsigned value = sizeof...(Types);
> };
> ```

Please be careful when reviewing this PR, it's my first one.

Here is a test, but I didn't know what the best location for it would be:

```cpp
#include <assert.h>

template <int... a>
struct b
{
    int c() { return sizeof...(a); }
};

int main()
{
    b<0, 1, 2, 3> d;
    int theSize = d.c();
    assert(theSize == 4);
}
```

```shell
Generating GOTO Program
GOTO program creation time: 0.016s
WARNING: no body for function __ESBMC_pthread_start_main_hook
WARNING: no body for function __ESBMC_atexit_handler
WARNING: no body for function __ESBMC_pthread_end_main_hook
Symex completed in: 0.000s (8 assignments)
Slicing time: 0.000s (removed 8 assignments)
Generated 1 VCC(s), 0 remaining after simplification (0 assignments)
BMC program time: 0.000s

VERIFICATION SUCCESSFUL
```

### DIFF
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -23,6 +23,7 @@ CC_DIAGNOSTIC_POP()
 #include <util/std_expr.h>
 #include <fmt/core.h>
 #include <clang-c-frontend/typecast.h>
+#include <util/c_types.h>
 
 clang_cpp_convertert::clang_cpp_convertert(
   contextt &_context,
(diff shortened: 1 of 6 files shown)


intrigus-lgtm:
Fixed missing import and switched to hopefully correct `size_type()` instead of `uint_type()`.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -23,6 +23,7 @@ CC_DIAGNOSTIC_POP()
 #include <util/std_expr.h>
 #include <fmt/core.h>
 #include <clang-c-frontend/typecast.h>
+#include "util/c_types.h"
```
fbrausse: Let's use \<angle\> brackets for includes. The "quoted" version starts looking in the source file's directory (where the header wouldn't be found).

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -759,6 +760,32 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+      ross.flush();
+      log_error("{}", oss.str());
+      return true;
+    }
+    else
+    {
+      new_expr = constant_exprt(
+        integer2binary(size_of_pack.getPackLength(), bv_width(size_type())),
+        integer2string(size_of_pack.getPackLength()),
+        size_type());
```
fbrausse, line 784: size_type() should be fine here, indeed. Alternatively, you could've used something like
```c++
typet type;
if (get_type(size_of_pack.getType(), type))
  return true;
```

fbrausse (commented):
Thanks, looks good, just a comment about the #include.

fbrausse:
Also, could you please add the C++ program featuring the sizeof... to a new subdirectory in regression/esbmc-cpp11/template, e.g. regression/esbmc-cpp11/template/sizeof? You're going to have to add regression/esbmc-cpp11/template to regression/CMakeLists.txt, to the list `REGRESSIONS_CPP11` of directories.

intrigus-lgtm:
Fixed import to use angle brackets.
Added positive and negative tests.

lucasccordeiro (approved):

XLiZHI (approved):
LGTM

fbrausse (approved):
Thanks for addressing my comments. LGTM.

PR #1732, opened 2024-03-05
[Concurrency] Do not record __ESBMC_rounding_mode

Anthonysdu:
I feel there may be no need to perform context switches for every global variables that not from user's program, for example, disabling context switches if meet `__ESBMC_rounding_mode` can reduce the interleavings size.

This PR: https://github.com/esbmc/esbmc/actions/runs/8155914160/job/22293103022

```
  correct:             428
    correct true:      143
    correct false:     285
  incorrect:             5
    incorrect true:      4
    incorrect false:     1
  unknown:             232
  Score:               427 (max: 999)

Runtime: 9730s
```
master:
```
  correct:             427
    correct true:      142
    correct false:     285
  incorrect:             5
    incorrect true:      4
    incorrect false:     1
  unknown:             233
  Score:               425 (max: 999)

Runtime: 12600s
```
Total runtime reduction is 2870 (s).

And there is a bug in the regression, should be a guard problem, while using `--no-por` can trigger the violation.

Another minimal case:

```c
#include <assert.h>
#include <pthread.h>
_Bool receive = 0;
int i = 0;
void *t1() {
  for (int i = 0; i < 2; i++) {
    if (receive) {
      assert(i < 1);
      receive = 0;
    }
  }
  return NULL;
}
int main() {
  pthread_t id;
  pthread_create(&id, NULL, t1, NULL);
  receive = 1;
  return 0;
}
```
esbmc --no-simplify --unwind 3 main.c: assertion i < 1 failure
esbmc --no-por main.c: success

### DIFF
--- a/regression/esbmc-unix/03_circular_buffer_02/test.desc
+++ b/regression/esbmc-unix/03_circular_buffer_02/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --context-bound 3
 ^VERIFICATION FAILED$
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -837,7 +837,8 @@ void execution_statet::get_expr_globals(
       name == "c:@__ESBMC_alloc" || name == "c:@__ESBMC_alloc_size" ||
       name == "c:@__ESBMC_is_dynamic" ||
       name == "c:@__ESBMC_blocked_threads_count" ||
-      name.find("c:pthread_lib") != std::string::npos)
+      name.find("c:pthread_lib") != std::string::npos ||
+      name == "c:@__ESBMC_rounding_mode")
     {
       return;
     }


lucasccordeiro (approved):

lucasccordeiro:

fbrausse:
> Another minimal case: [...]
> 
> esbmc --no-simplify --unwind 3 main.c: assertion i < 1 failure
> esbmc --no-por main.c: success

I think this minimal case should trigger the assertion violation:
1. main() creates thread t1, but doesn't set `receive = 1`, yet
2. t1 starts running and performs the first iteration of the loop (receive is still 0)
3. main thread sets `receive = 1`
4. t1 performs second iteration of the loop (receive == 1, but i == 1 as well), assertion violation

There actually is not a huge difference to the esbmc-unix/03_circular_buffer_02 test case, which could do the same: perform iterations while either if-condition is not met. This changes the iteration counter `i` and therefore the inserted element. I don't understand how on master the result was wrong before this PR:
```
State 4 file main.c line 45 column 5 function insertLogElement thread 1
  buffer[1] = 4 (00000100)

State 5 file main.c line 83 column 7 function t2 thread 2
Violated property:
  file main.c line 83 column 7 function t2
  assertion removeLogElement()==i
  return_value$_removeLogElement$1 == i
```

fbrausse:
> esbmc --no-simplify --unwind 3 main.c: assertion i < 1 failure
> esbmc --no-por main.c: success

Oh, now I understand what you're saying. This minimal case already fails on master with those settings you provided. Sorry, the context hasn't been clear to me. LGTM. Maybe add your new TC as well.

PR #1733, opened 2024-03-06
[Concurrency] Do not record __ESBMC_pthread_thread

Anthonysdu:
This PR follows #1732 , we skip context switch for ESBMC pthread variables: `__ESBMC_pthread_thread_running`,`__ESBMC_pthread_thread_ended` and `__ESBMC_pthread_thread_detach`

```
  correct:             432
    correct true:      147
    correct false:     285
  incorrect:             7
    incorrect true:      6
    incorrect false:     1
  unknown:             226
  Score:               371 (max: 999)
```

master
```
  correct:             428
    correct true:      143
    correct false:     285
  incorrect:             5
    incorrect true:      4
    incorrect false:     1
  unknown:             232
  Score:               427 (max: 999)
```

The CPU time is reduced from 9730s to 7130s (false 3800s -> 2870s, true 5930s -> 4260s).

There are two more incorrect trues, but it's not because of this PR, since current master also reports incorrect true by --context-bound 2 --no-por.
image

### DIFF
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -256,7 +256,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
                  check_if_benchmark_contains_pthread(benchmark))
 
   if concurrency:
-    command_line += " --no-por --context-bound 2 "
+    command_line += " --smt-during-symex --smt-thread-guard --z3 --no-por --context-bound 2 "
     #command_line += "--no-slice " # TODO: Witness validation is only working without slicing
 
   # Add witness arg
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -256,7 +256,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
                  check_if_benchmark_contains_pthread(benchmark))
 
   if concurrency:
-    command_line += " --smt-during-symex --smt-thread-guard --z3 --no-por --context-bound 2 "
+    command_line += " --no-por --context-bound 2 "
     #command_line += "--no-slice " # TODO: Witness validation is only working without slicing
 
   # Add witness arg
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -838,7 +838,8 @@ void execution_statet::get_expr_globals(
       name == "c:@__ESBMC_is_dynamic" ||
       name == "c:@__ESBMC_blocked_threads_count" ||
       name.find("c:pthread_lib") != std::string::npos ||
-      name == "c:@__ESBMC_rounding_mode")
+      name == "c:@__ESBMC_rounding_mode" ||
+      name.find("c:@__ESBMC_pthread_thread") != std::string::npos)
     {
       return;
     }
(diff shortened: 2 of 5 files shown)


src/goto-symex/execution_state.cpp
```diff
@@ -838,7 +838,8 @@ void execution_statet::get_expr_globals(
       name == "c:@__ESBMC_is_dynamic" ||
       name == "c:@__ESBMC_blocked_threads_count" ||
       name.find("c:pthread_lib") != std::string::npos ||
-      name == "c:@__ESBMC_rounding_mode")
+      name == "c:@__ESBMC_rounding_mode" ||
```
lucasccordeiro, line 841: I wonder whether there is a more efficient way to perform these string comparisons, something like:

````CPP
#include <unordered_set>

// Precompute the set of strings to compare against
static const std::unordered_set<std::string> excludedNames = {
    "c:@__ESBMC_alloc", "c:@__ESBMC_alloc_size",
    "c:@__ESBMC_is_dynamic", "c:@__ESBMC_blocked_threads_count",
    "c:@__ESBMC_rounding_mode"
};

// Check if the given name is excluded
bool isExcluded(const std::string& name) {
    return excludedNames.count(name) > 0 ||
           name.find("c:pthread_lib") != std::string::npos;
}

// Usage example:
if (isExcluded(name)) {
    return;
}
````

**Review by lucasccordeiro — CHANGES_REQUESTED** [review]

File: src/goto-symex/execution_state.cpp (hunk 1 above)
Comment (fbrausse) [reply src/goto-symex/execution_state.cpp:841]: I doubt it. The string equality-based comparisons check the length first.

I guess it's more performant to check the common prefix `c:@__ESBMC_` and then do string comparisons. The latter will usually be quick since the strings have different lengths and/or start with a different symbol.

But benchmarks will tell the truth...

Comment (mikhailramalho) [reply src/goto-symex/execution_state.cpp:841]: It should be faster to compare the dstring instead of the string, i.e., the thename from the symbol, instead of calling as_string()

Comment (Anthonysdu) [reply src/goto-symex/execution_state.cpp:841]: The approaches have little preformance difference in esbmc-unix and esbmc-unix2:

this PR + hashset: 489.3s + 717.5s
check common prefix first + hashset: 487.7s + 713.9s
dstring + hashset: 488.9s + 720.5s.

And there could be deviation because each one is single run on my laptop.
How about further evaulate this performance in other PR?

**Comment by lucasccordeiro** [comment]
> How about further evaulate this performance in other PR?

Fine by me.

File: src/goto-symex/execution_state.cpp (hunk 1 above)
Comment (fbrausse) [reply src/goto-symex/execution_state.cpp:841]: > How about further evaulate this performance in other PR?

Sure, that's fine. As a clarification, my comment was recommending against any sort of hash table for these comparisons, because hash table lookups require hashing the entire string before the first array indexing is performed and then checking equality of the strings obtained in that bucket (if any) - the hashing cost makes sense to pay when the number of strings in the table is large in comparison to the lengths of the strings being looked up.

**Review by fbrausse — APPROVED** [review]

**Comment by lucasccordeiro** [comment]
@Anthonysdu : could you please revert the last commits?

**Review by lucasccordeiro — APPROVED** [review]

File: scripts/competitions/svcomp/esbmc-wrapper.py [hunk 2]
```diff
@@ -256,7 +256,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
                  check_if_benchmark_contains_pthread(benchmark))
 
   if concurrency:
-    command_line += " --no-por --context-bound 2 "
+    command_line += " --smt-during-symex --smt-thread-guard --z3 --no-por --context-bound 2 "
```
fbrausse: Whoa, wait.

Anthonysdu: sorry, by mistake..I thought this was merged

fbrausse: Done now.

PR #1734, opened 2024-03-07
[python-frontend] Add support for float numbers

brcfarias:
This PR adds support for assignments and arithmetic operations involving floating-point numbers.


fbrausse (commented):
LGTM, just one question about the typing.

regression/python/float-fail/main.py
```diff
@@ -0,0 +1,6 @@
+num1 = 1.5
+num2 = 3.0
+
+div = num2 / num1
+assert div == 2  # div is a float
```
fbrausse: This looks strange from a Python perspective. I get from the comment that you're making this fail because of the difference in the types, but there actually are no type annotations here, so on the right-hand side we're dealing with Python's `int` type, which is not `int64_t` or similar (I forgot how it's actually called in the API you're targeting).

So I'm wondering whether regarding these kind of implicit conversions there are rules in the API you're targeting for `int64_t`. Because for Python itself this code is fine: `int` just gets (somehow) converted to `float` (which is `double` in C/C++ lingo) and the comparison is performed there. But is the same being done for `int64_t`?

lucasccordeiro (approved):

regression/python/float-fail/main.py
brcfarias: You're right, thanks for pointing out.
> I'm wondering whether regarding these kind of implicit conversions there are rules in the API you're targeting for int64_t

No, there aren't. In fact, this assertion is valid even in the context of consensus API due to the internal conversion you mentioned.
I'll move the test to 'knownbug' and revisit it later.

lucasccordeiro:
@brcfarias: Could you please add a description of this PR?

fbrausse (approved):

lucasccordeiro:

Issue #1729, opened 2024-03-04
[C++ Verification] ERROR: Can't construct rvalue reference to array type during dereference


XLiZHI:
This error occurs after PR #1696 is merged.

After my initial investigate, the simplified code is as follows:

```
struct B
{
};

struct list
{
  B list[10];
};

int main ()
{
  list mylist;
  return 0;
}
```

There are problems with lhs:
```
member
* source_value : dereference
    * pointer : symbol
        * name : c:@S@list@F@list#this
        * renamelev : Level 1
        * level1_num : 1
        * level2_num : 0
        * thread_num : 0
        * node_num : 0
        * type : pointer
          * subtype : symbol
              * symbol_name : tag-list
    * type : struct
      * members : 
        0: array
          * subtype : struct
              * members : 

              * member_names : 

              * member_pretty_names : 

              * typename : B
              * packed : false
          * array_size : constant_int
              * value : 10
              * type : unsignedbv
                * width : 64
          * size_is_infinite : false

      * member_names : 
        0: list

      * member_pretty_names : 
        0: list

      * typename : list
      * packed : false
* member_name : list
* type : array
  * subtype : struct
      * members : 

      * member_names : 

      * member_pretty_names : 

      * typename : B
      * packed : false
  * array_size : constant_int
      * value : 10
      * type : unsignedbv
        * width : 64
  * size_is_infinite : false
```

fbrausse:
> After my initial investigate, the simplified code is as follows:

What's the original code?

XLiZHI:
commit: a86f3cc537f765031eb53ebc6338bc41fbbfa02c
```
B (c:@S@B@F@B#):
        // 20 file main9.cpp line 1 column 8 function B
        END_FUNCTION // B
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

list (c:@S@list@F@list#):
        // 21 
        FUNCTION_CALL:  B(&this->list[0])
        // 22 file main9.cpp line 5 column 8 function list
        END_FUNCTION // list
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

main (c:@F@main#):
        // 23 no location
        list mylist;
        // 24 
        FUNCTION_CALL:  list(&mylist)
        // 25 file main9.cpp line 13 column 3 function main
        RETURN: 0
        // 26 file main9.cpp line 14 column 1 function main
        END_FUNCTION // main
[...]

PR #1736, opened 2024-03-12
[C++ Verification] avoid the temporary object

XLiZHI:
This PR attempts to avoid some temporary objects generated by member initialization.

Try to resolve the issue:
- https://github.com/esbmc/esbmc/issues/1729

Fixed 4 TCs, ~~caused 1 KOWNBUG:~~

~~`regression/esbmc-cpp/cpp/ch21-7` : dereference failure: invalid pointer, this TC was fixed in the last PR and seems to be caused by temporary obj.~~

It is not clear to me at this time why temporary variables affect constructors, it seems to cause some variables to fail to initialize. It would be nice to explain what temporary objects do.


lucasccordeiro (approved):

fbrausse:
> It is not clear to me at this time why temporary variables affect constructors, it seems to cause some variables to fail to initialize. It would be nice to explain what temporary objects do.

They are handled as goto-sideeffects and introduce a new symbol (with a new name). The tmp-obj expression itself is being replaced with a symbol_exprt referring to this new symbol. This happens in goto_convertt::remove_temporary_object(). These tmp-obj expressions have two optional attributes:
1. If the tmp-obj expression has an operand, it is used as the initialization of the new symbol (via assignment).
2. If the tmp-obj expression has an .initializer() (which is supposed to contain code that can do arbitrary things with the new symbol), then all `new_object` expressions in that initializer will get replaced by the symbol_expr referring to the new symbol, see goto_convertt::replace_new_object().

The initializer and  `new_object` expression being used in get_constructor_call() is basically a reference to the anonymous symbol, which the `else` branch was using to build the `this` pointer, for instance when there is no declaration attached to the constructor call. E.g. something like a temporary here:
```c++
struct A { int x; A(int x) : x(x) {} };

A a(A(42).x);
```

~Now that I look at this, I'm wondering whether the `new_object` expression, which is assigned to the operand of the tmp-obj expression (i.e., case 1) in the `else` branch of get_constructor_call() is actually being replaced by the proper symbol_expr.~

*Edit:* Sorry, my bad, it's using case 2. That should be fine.

XLiZHI:
> They are handled as goto-sideeffects and introduce a new symbol (with a new name). The tmp-obj expression itself is being replaced with a symbol_exprt referring to this new symbol. This happens in goto_convertt::remove_temporary_object(). These tmp-obj expressions have two optional attributes:
> 
> 1. If the tmp-obj expression has an operand, it is used as the initialization of the new symbol (via assignment).
> 2. If the tmp-obj expression has an .initializer() (which is supposed to contain code that can do arbitrary things with the new symbol), then all `new_object` expressions in that initializer will get replaced by the symbol_expr referring to the new symbol, see goto_convertt::replace_new_object().

Thanks @fbrausse, I've read this code before, and what makes it difficult for me is that we do not have enough information to determine whether temporary objects are needed. I've been working on it for a few days, but I haven't made any progress. In addition, we really need some simple TCs to explore this problem. 

> Now that I look at this, I'm wondering whether the new_object expression, which is assigned to the operand of the tmp-obj expression (i.e., case 1) in the else branch of get_constructor_call() is actually being replaced by the proper symbol_expr.

At present, I use this PR to avoid some errors temporarily, but I think we should try to avoid temporary objects, which will make the goto program very large. What else do you think we need to do in this PR?

fbrausse:
> At present, I use this PR to avoid some errors temporarily, but I think we should try to avoid temporary objects, which will make the goto program very large. What else do you think we need to do in this PR?

1. It would of course be good to find a more permanent solution, maybe in this PR.
2. We can't always avoid temporary objects, e.g. in the example above: we need to call the ctor of `A` twice there, once with the declared symbol `a` and once with a temporary. Whether this is to be done via a `temporary_object` expression or by introducing a new symbol in the converter directly (as done for compound literals, for instance), I don't know. Ideally, the tmp-obj expression would do what it's supposed to.

For compound literals, see https://github.com/esbmc/esbmc/blob/2f5a0ee8f81ad27a4b2fa5956213e8dcc2d56eee/src/clang-c-frontend/clang_c_convert.cpp#L1732 and https://github.com/esbmc/esbmc/blob/2f5a0ee8f81ad27a4b2fa5956213e8dcc2d56eee/src/clang-c-frontend/clang_c_convert.cpp#L531

I tried reusing this code in my last PR about C++ references, but it basically inserted the symbol declarations at the wrong position in the code that's why I stayed with the tmp-obj expression there.

@kunjsong01 did call need_new_object() in that place before my last change: https://github.com/esbmc/esbmc/blob/2f5a0ee8f81ad27a4b2fa5956213e8dcc2d56eee/src/clang-cpp-frontend/clang_cpp_convert.cpp#L819-L824

However, as it just examines the direct parent expression, this didn't work: there could be casts and other things in the parent expression(s).

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -826,8 +826,15 @@ bool clang_cpp_convertert::get_constructor_call(
    */
 
   // Calling base constructor from derived constructor
+  bool need_new_obj = false;
+
   if (new_expr.base_ctor_derived())
     gen_typecast_base_ctor_call(callee_decl, call, new_expr);
+  else if (
+    new_expr.get_bool("memberInit") &&
```
fbrausse: Let's stick to the non-capitalized names delimited by underscores. Also, I believe, this should be prefixed by `#` in order to indicate that it's a "comment".

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -826,8 +826,15 @@ bool clang_cpp_convertert::get_constructor_call(
...
+    !callee_decl.type().return_type().get_bool("#copy_cons"))
```
fbrausse: Could you add a comment indicating why copy-constructors are special here? For instance, I'm wondering what move-constructors would do, or actually in what way user-defined constructors are different.

XLiZHI:
```
        A a;
        // 23 file main8.cpp line 7 column 5 function main
        A tmp$1;
        // 24 file main8.cpp line 7 column 5 function main
        tmp$1=NONDET(struct);
        // 25 file main8.cpp line 7 column 7 function main
        A tmp$2;
        // 26 file main8.cpp line 7 column 7 function main
        tmp$2=NONDET(struct);
        // 27 
        FUNCTION_CALL:  A(&tmp$2, 42)
        // 28 
        FUNCTION_CALL:  A(&tmp$1, tmp$2.x)
        // 29 file main8.cpp line 7 column 3 function main
        a=tmp$1;
```

```
`-FunctionDecl 0x82bad58 <line:5:1, line:9:1> line:5:5 main 'int ()'
  `-CompoundStmt 0x82bb2c0 <line:6:1, line:9:1>
    `-DeclStmt 0x82bb2a8 <line:7:3, col:15>
      `-VarDecl 0x82bae40 <col:3, col:14> col:5 a 'A' callinit
        `-CXXConstructExpr 0x82bb278 <col:5, col:14> 'A' 'void (int)'
          `-MemberExpr 0x82bb228 <col:7, col:13> 'int' .x 0x82b4858
            `-CXXFunctionalCastExpr 0x82bb200 <col:7, col:11> 'A' functional cast to struct A <ConstructorConversion>
              `-CXXConstructExpr 0x82bb0b0 <col:7, col:11> 'A' 'void (int)'
                `-IntegerLiteral 0x82baeb8 <col:9> 'int' 42
```

> However, as it just examines the direct parent expression, this didn't work: there could be casts and other things in the parent expression(s).

Actually I have tried, but I don't think we can stably use AST, as you said the parent nodes may exist cast.

src/clang-cpp-frontend/clang_cpp_convert.cpp
XLiZHI: After further investigation, I found that the arguments of the cpy constructor was wrong:
```
#include <cassert>

struct B {
    int data;

    B(int x) : data(x) {}
};

struct A {
    B b;

    A(const B& bCopy) : b(bCopy) {}
};

int main() {
    B bOriginal(10);
    A a(bOriginal);

    assert(a.b.data == 10);

    return 0;
}
```
use tmp obj:

```
A (c:@S@A@F@A#&1$@S@B#):
        // 24 file main10.cpp line 12 column 25 function A
        B tmp$1;
        // 25 file main10.cpp line 12 column 25 function A
        tmp$1=NONDET(struct);
        // 26 
        FUNCTION_CALL:  B(&tmp$1, &(*bCopy))
        // 27 
        this->b=tmp$1;
        // 28 file main10.cpp line 12 column 35 function A
        dead tmp$1;
        // 29 file main10.cpp line 12 column 35 function A
        END_FUNCTION // A
```
original:
```
A (c:@S@A@F@A#&1$@S@B#):
        // 24
        FUNCTION_CALL:  B(&(*bCopy), &this->b)   // should be B( &this->b, &(*bCopy)) I think.
        // 25 file main10.cpp line 14 column 35 function A
        END_FUNCTION // A
```

I haven't tried move constructor yet, but I think it should be the same.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -826,8 +826,17 @@ bool clang_cpp_convertert::get_constructor_call(
...
   // Calling base constructor from derived constructor
+  bool need_new_obj = false;
+
   if (new_expr.base_ctor_derived())
     gen_typecast_base_ctor_call(callee_decl, call, new_expr);
+  else if (new_expr.get_bool("#member_init"))
+  {
+    /*
+    struct A {}; struct B {A a; B(int _a) : a(_a) {}};
+    In this case, use variables to construct the object directly
```
fbrausse: `A` needs a constructor taking an int. Also, I'd replace "variables" by "members" for clarity.

fbrausse (approved):
Nice, good catch with the adjuster call! I missed that. LGTM! One nitpick.

Issue #1737, opened 2024-03-12
[C++ verification] problem with the std::enable_if template


lucasccordeiro:
Bug report from an ESBMC user: " I have a problem with the std::enable_if template. Its is defined in <type_traits> and is used extensively around the code I’m trying to compile."

lucasccordeiro:
@XLiZHI : could you please check this issue?

lucasccordeiro:
Example:

````CPP
#include <iostream>
#include <type_traits>

// Function template that takes two arguments and returns their sum
template<typename T>
typename std::enable_if<std::is_integral<T>::value, T>::type
add(T a, T b) {
    return a + b;
}

int main() {
    // Call the add function with integer arguments
    std::cout << "add(3, 4): " << add(3, 4) << std::endl;

    // Uncommenting this line would result in a compile-time error
    //std::cout << "add(3.5, 4.5): " << add(3.5, 4.5) << std::endl;

    return 0;
}
````

ESBMC command line:

esbmc main.cpp

ESBMC output:

````
ESBMC version 7.5.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing alex.cpp
main.cpp:6:15: error: no type named 'enable_if' in namespace 'std'
typename std::enable_if<std::is_integral<T>::value, T>::type
~~~~~~~~~~~~~~^~~~~~~~~
main.cpp:6:24: error: expected unqualified-id
typename std::enable_if<std::is_integral<T>::value, T>::type
                       ^
ERROR: PARSING ERROR
````

fbrausse:
This issue with enable_if has been addressed by #1739 on the current master branch. In case it's not for the user, let's open another issue then.

PR #1739, opened 2024-03-13
Remove <type_traits> + support C++17 variable template declarations

fbrausse:
Addresses #1737 by removing our \<type_traits\> header overriding the system library's. It should be fine in particular for this header, because it is meant to only contain compile-time constant definitions (and lots of them at that).

Also adds support for variable template declarations as required by the system's header for C++ >= 17 (e.g. string_view needs those).

Sorry @XLiZHI for "stealing" this issue, I just wanted to quickly check whether we really need to override these compile-time constants.


XLiZHI (approved):
Thanks @fbrausse , I am surprised that we can use this standard library, last time I tried it there were a lot of errors.

lucasccordeiro (approved):

fbrausse:
> I am surprised that we can use this standard library, last time I tried it there were a lot of errors.

Initially I also had about 400 regression errors, but nearly all of them were due to \<utility\> including \<type_traits\> unconditionally, though it only exists since C++11. The others were about std::swap having multiple definitions and variable templates (which I'm actually not entirely sure we really need to convert to symbols, but we can address that once we get into C++17 support).

PR #1741, opened 2024-03-14
SMTLIB output validation suite

rafaelsamenezes:
This allow us to check whether the SMTLIB output is producing valid formulas by checking the produced formulas against Z3. I also removed the --array-flattener flag from the default flags as it was causing some error in ESBMC (I will open an issue later for those).

### DIFF
--- a/regression/CMakeLists.txt
+++ b/regression/CMakeLists.txt
@@ -37,10 +37,29 @@ function(add_esbmc_regression_test folder modes test)
       LABELS "regression;${folder}")
 endfunction()
 
+function(add_smtlib_regression_test folder modes test)
+    set(test_name "smtlib/${folder}/${test}")
+    add_test(NAME ${test_name}
+             COMMAND ${Python_EXECUTABLE} ${ESBMC_REGRESSION_TOOL}
+                     --tool=${ESBMC_BIN}
+                     --regression=${CMAKE_CURRENT_SOURCE_DIR}/${folder}
+                     --modes ${modes}
+                     --file=${test}
+                     --smt_test
+                     --smt_solver=z3
+                     ${BENCHBRINGUP_ARGS}) # additional args for BENCHBRINGUP workflow
+    # TODO: Testing Tool should have a command to dump the TAG of a test		   
+    set_tests_properties(${test_name}
+      PROPERTIES SKIP_RETURN_CODE 10
+      LABELS "smtz3;${folder}")
+endfunction()
 function(add_esbmc_regression folder modes)
     SUBDIRLIST(SUBDIRS ${CMAKE_CURRENT_SOURCE_DIR}/${regression})
     foreach(test ${SUBDIRS})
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

regression/CMakeLists.txt
```diff
@@ -37,10 +37,29 @@ function(add_esbmc_regression_test folder modes test)
...
+function(add_smtlib_regression_test folder modes test)
+    set(test_name "smtlib/${folder}/${test}")
+    add_test(NAME ${test_name}
+             COMMAND ${Python_EXECUTABLE} ${ESBMC_REGRESSION_TOOL}
+                     --tool=${ESBMC_BIN}
+                     --regression=${CMAKE_CURRENT_SOURCE_DIR}/${folder}
+                     --modes ${modes}
+                     --file=${test}
+                     --smt_test
+                     --smt_solver=z3
```
fbrausse, line 49: This assumes that the executable `z3` is installed on the `UNIX AND NOT APPLE` system. For my PC that's true, but can we expect that for everyone running `ctest`?

regression/testing_tool.py
```diff
@@ -189,12 +201,23 @@ def test(self):
...
+            try:
+                p = subprocess.run(smt_command, stdout=PIPE, stderr=PIPE, timeout=10)
+                if p.returncode != 0:
+                    self.fail("Unable to parse the generated SMT formula. Flags: " + str(test_case.generate_run_argument_list(*executor.tool)))
+            except subprocess.TimeoutExpired:
+                return
+        # NOTE: We can't check for RC as some CORE test cases results in PARSING ERRORS :) 
         if TestCase.RUN_ONLY:
-            if rc != 0:
-                self.fail(f"Wrong output for process. Bombed out with exit code {rc}")
```
fbrausse, line 197: What's the problem with the PARSING ERROR tests here? They did exist before, no?

rafaelsamenezes: No problem, I am just not sure what we are verifying at this point. Are we verifying for a successful execution? It can be verification successful, SMTLIB, verification failed, parsing error.  We could try to validate every one of them individually, but then we don't know what the original test case wanted.

Thats why I was annoyed that the PARSING ERROR for a limitation is in core.  That being said, we could also consider testing whether the frontend can find issues correctly.

So, I preferred to just disable it. RUN_ONLY should check some byproduct of the run, not the run itself.

regression/CMakeLists.txt
rafaelsamenezes: Oops, I am going to do a proper add_executable call here.

lucasccordeiro:
@rafaelsamenezes: can I ask you for the status of this PR?

mikhailramalho:
Closing as part of stale-branch cleanup. Master has `regression/smtlib/` and related infrastructure now. If this validation suite still adds value over what shipped, a fresh PR is easier than rebasing nearly a year of drift. The branch will be deleted; commits remain reachable via this PR's history.

PR #1744, opened 2024-03-14
Python: Add support for "bytes" type and array indexing

brcfarias:
This PR adds support for assignments and indexing with [bytes](https://docs.python.org/3/library/stdtypes.html#bytes-and-bytearray-operations) type.

Since ast library cannot distinguish strings and bytes, a "esbmc_type_annotation" node is being added in the AST to handle it.


fbrausse (approved):
LGTM.

lucasccordeiro (approved):

PR #1745, opened 2024-03-15
[C++ verification] added support for nullptr_t

### LINKED ISSUE #1727: Unrecognized clang builtin type nullptr_t
cpptest.pre.cpp:
```cpp
void a(decltype(nullptr));
```

Running

```
esbmc --cppstd 17 --input-file cpptest.pre.cpp --no-abstracted-cpp-includes --no-library
```

Fails with:

```
ERROR: Unrecognized clang builtin type nullptr_t
BuiltinType 0x563faf29bce0 'nullptr_t'

ERROR: CONVERSION ERROR
```

### LINKED ISSUE #1742: [C++ Verification] Conversion of unsupported clang expr: "CXXNullPtrLiteralExpr" to expression
$ esbmc main.cpp --cppstd 11

````CPP
#include <iostream>

void myFunction(int* ptr) {
    if (ptr == nullptr) {
        std::cout << "Pointer is nullptr" << std::endl;
    } else {
        std::cout << "Pointer is not nullptr" << std::endl;
    }
}

int main() {
    int* ptr1 = nullptr;
    int* ptr2 = new int(5);

    myFunction(ptr1); // Passing nullptr
    myFunction(ptr2); // Passing valid pointer

    delete ptr2;

    return 0;
}
````
ESBMC output:

````
ESBMC version 7.5.0 64-bit x86_64 linux
Parsing main.cpp
Converting
ERROR: Conversion of unsupported clang expr: "CXXNullPtrLiteralExpr" to expression
CXXNullPtrLiteralExpr 0x7953248 <main.cpp:4:16> 'nullptr_t'

ERROR: CONVERSION ERROR
````

### PR DESCRIPTION
**Pull request by XLiZHI**
Added a new node: `clang: : Stmt: : CXXNullPtrLiteralExprClass`, do the same with `GNUNullExprClass`.

In this PR, `BuiltinType::NullPtr` is modeled as `int`, (Maybe `long` would be more appropriate?)

For the code: 

`int* p = nullptr;`

AST:
```
    | `-VarDecl 0x7af1d18 <col:5, col:14> col:10 used p 'int *' cinit
    |   `-ImplicitCastExpr 0x7af1d90 <col:14> 'int *' <NullToPointer>
    |     `-CXXNullPtrLiteralExpr 0x7af1d80 <col:14> 'nullptr_t'
```

_Edit_: Does anyone know why `git-auto-commit-action` always fails?

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1830,8 +1830,7 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
   case clang::Stmt::CXXNullPtrLiteralExprClass:
   case clang::Stmt::GNUNullExprClass:
   {
-    const clang::Expr &gnun =
-      static_cast<const clang::Expr &>(stmt);
+    const clang::Expr &gnun = static_cast<const clang::Expr &>(stmt);
 
     typet t;
     if (get_type(gnun.getType(), t))
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1397,6 +1397,11 @@ bool clang_c_convertert::get_builtin_type(
     c_type = "__uint128";
     break;
 
+  case clang::BuiltinType::NullPtr:
+    new_type = pointer_type();
+    c_type = "uintptr_t";
+    break;
+
 #ifdef ESBMC_CHERI_CLANG
   case clang::BuiltinType::IntCap:
     new_type = intcap_typet();
@@ -1822,10 +1827,11 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     break;
   }
 
+  case clang::Stmt::CXXNullPtrLiteralExprClass:
   case clang::Stmt::GNUNullExprClass:
   {
-    const clang::GNUNullExpr &gnun =
-      static_cast<const clang::GNUNullExpr &>(stmt);
+    const clang::Expr &gnun =
+      static_cast<const clang::Expr &>(stmt);
 
     typet t;
     if (get_type(gnun.getType(), t))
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by lucasccordeiro — APPROVED** [review]

**Review by fbrausse — CHANGES_REQUESTED** [review]
I don't think we can model it as `int`:
image
See https://en.cppreference.com/w/cpp/types/nullptr_t

It should either be like `uintptr_t` or `void *` (despite the above stating that it's not a pointer type, but the type-checking will already have been performed by Clang). For `uintptr_t`, `unsignedbv_typet(config.ansi_c.pointer_width())` should work.

**Comment by fbrausse** [comment]
> _Edit_: Does anyone know why `git-auto-commit-action` always fails?

It fails, because the CI doesn't have rights to push to the branch in your cloned repository. I don't quite know how to avoid it, but I also don't like the auto-commit-feature very much and thus have disabled it for my branches (in this repo): https://github.com/esbmc/esbmc/blob/cfb6274105c9ac9d51d5abb18b60698f10de1003/.github/workflows/pull_request.yml#L87-L91

Would need to look into the Github CI documentation how to identify PRs whose source branch is outside this repo, in order to disable the auto-commit there as well. @rafaelsamenezes Do you know how we can avoid this?

**Review by fbrausse — COMMENTED** [review]
The title says "support for nullptr_t". Does it?
```c++
#include <cassert>

bool f(std::nullptr_t x)
{
  return x == &x;
}
int main() { assert(!f(nullptr)); }
```

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -796,7 +796,7 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     }
     break;
   }
-
```
mikhailramalho: please remove whitespace

fbrausse (approved):
LGTM. Thanks for addressing my comments.

lucasccordeiro:

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1397,6 +1397,11 @@ bool clang_c_convertert::get_builtin_type(
     c_type = "__uint128";
     break;
 
+  case clang::BuiltinType::NullPtr:
+    new_type = pointer_type();
+    c_type = "uintptr_t";
```
intrigus-lgtm, line 1402: Why is the c_type uintptr_t?

fbrausse: Because it's not supposed to be a pointer type, but have the same size as a pointer.

PR #1755, opened 2024-03-19
Optimization of Interval Analysis

rafaelsamenezes:
This PR changes how the Interval Domain defines its state by:

1. Using more references and variants
2. Not storing TOP variables.
3. Filtering the variables that need to be joined
4. The interval domain is now a reference.

### DIFF
--- a/src/goto-programs/abstract-interpretation/interval_domain.cpp
+++ b/src/goto-programs/abstract-interpretation/interval_domain.cpp
@@ -54,8 +54,9 @@ void interval_domaint::update_symbol_interval(
     intervals.erase(sym.thename);
     return;
   }
-    
-  intervals[sym.thename] = std::make_shared<interval_domaint::integer_intervalt>(value); 
+
+  intervals[sym.thename] =
+    std::make_shared<interval_domaint::integer_intervalt>(value);
 }
 
 template <>
@@ -68,7 +69,8 @@ void interval_domaint::update_symbol_interval(
     intervals.erase(sym.thename);
     return;
   }
-  intervals[sym.thename] = std::make_shared<interval_domaint::real_intervalt>(value);
+  intervals[sym.thename] =
+    std::make_shared<interval_domaint::real_intervalt>(value);
 }
 
 template <>
@@ -113,7 +115,6 @@ void interval_domaint::apply_assume_symbol_truth(
     interval.make_le_than(1);
   }
 
-
   update_symbol_interval(sym, interval);
 }
 
@@ -133,7 +134,8 @@ interval_domaint::get_interval_from_const(const expr2tc &e) const
 #include <cmath>
 
 template <>
-interval_domaint::real_intervalt interval_domaint::get_interval_from_const(const expr2tc &e) const
+interval_domaint::real_intervalt
+interval_domaint::get_interval_from_const(const expr2tc &e) const
 {
   interval_domaint::real_intervalt result; // (-infinity, infinity)
   if (!is_constant_floatbv2t(e))
@@ -197,9 +199,8 @@ interval_domaint::get_interval_from_const(const expr2tc &e) const
 }
 
 template <>
-interval_domaint::integer_intervalt
-interval_domaint::generate_modular_interval<interval_domaint::integer_intervalt>(
-  const symbol2t sym) const
+interval_domaint::integer_intervalt interval_domaint::generate_modular_interval<
+  interval_domaint::integer_intervalt>(const symbol2t sym) const
 {
   auto t = sym.type;
   interval_domaint::integer_intervalt result;
@@ -224,7 +225,8 @@ interval_domaint::generate_modular_interval<interval_domaint::integer_intervalt>
 }
 
 template <>
-interval_domaint::real_intervalt interval_domaint::generate_modular_interval<interval_domaint::real_intervalt>(
+interval_domaint::real_intervalt
+interval_domaint::generate_modular_interval<interval_domaint::real_intervalt>(
   const symbol2t) const
 {
   // TODO: Support this
@@ -508,7 +510,6 @@ T interval_domaint::get_interval(const expr2tc &e) const
 
       else if (is_modulus2t(e))
         result = lhs % rhs;
-
     }
     break;
 
@@ -667,15 +668,16 @@ void interval_domaint::apply_assume_less<wrapped_interval>(
     make_bottom();
 }
 
-
 template <>
-bool interval_domaint::is_mapped<interval_domaint::integer_intervalt>(const symbol2t &sym) const
+bool interval_domaint::is_mapped<interval_domaint::integer_intervalt>(
+  const symbol2t &sym) const
 {
   return intervals.find(sym.thename) != intervals.end();
 }
 
 template <>
-bool interval_domaint::is_mapped<interval_domaint::real_intervalt>(const symbol2t &sym) const
+bool interval_domaint::is_mapped<interval_domaint::real_intervalt>(
+  const symbol2t &sym) const
 {
   return intervals.find(sym.thename) != intervals.end();
 }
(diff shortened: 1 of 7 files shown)


lucasccordeiro:
@rafaelsamenezes : could you please add some numbers here from the SV-COMP benchmarks?

1) **Baseline**: master without interval analysis.
2) **Interval+Master**: master with interval analysis.
3) **Interval+PR**: this PR with interval analysis.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -11,52 +11,79 @@
 #endif
 // Let's start with all templates specializations.
 
+template <size_t Index, class Interval>
+Interval interval_domaint::get_interval_from_variant(const symbol2t &sym) const
+{
+  const auto &it = intervals.find(sym.thename);
+  // TODO: mix floats/integer
+  if (it != intervals.end() && it->second.index() == Index)
```
fbrausse: Can the second condition ever fail here? If not, it might be possible to drop the `class Interval` parameter and use `decltype(std::get<Index>(declval<interval>()))` instead. If yes, I'd suggest to add a comment describing why/how it could fail (some mismatch between Index and the type stored in the variant).

src/goto-programs/abstract-interpretation/interval_domain.h
```diff
@@ -26,6 +23,17 @@ using real_intervalt =
...
+  using integer_intervalt = interval_templatet<BigInt>;
+  using real_intervalt =
+    interval_templatet<boost::multiprecision::cpp_bin_float_100>;
+
+  using interval = std::variant<
+    std::shared_ptr<integer_intervalt>,
+    std::shared_ptr<real_intervalt>,
+    std::shared_ptr<wrapped_interval>>;
+
+  using interval_map = std::unordered_map<irep_idt, interval, irep_id_hash>;
```
fbrausse, line 38: Could you please add a comment here saying what "no entry" in this map means, and also what an entry means, where the associated shared_ptr points to nothing (if that can even happen)? Just so we don't forget.

src/goto-programs/goto_contractor.h
```diff
@@ -18,10 +18,9 @@
 #include <ibex/ibex_Ctc.h>
 #include <irep2/irep2.h>
 #include <util/type_byte_size.h>
-#include <goto-programs/abstract-interpretation/interval_domain.h>
 #include <goto-programs/abstract-interpretation/interval_analysis.h>
 #include <limits>
-
+#include <chrono>
```
fbrausse, line 23: I don't see \<chrono\> being used here.

fbrausse (commented):
In general looks good, just a few comments.

src/goto-programs/abstract-interpretation/interval_domain.cpp
rafaelsamenezes: This might happen due to some wrong logic e.g., calling get<0> on a float interval.

fbrausse: std::get\<..\> will catch this (throws [std::bad_variant_access](https://en.cppreference.com/w/cpp/utility/variant/bad_variant_access)). If this guarding against logic errors, the function shouldn't return a default value in that case but abort.

rafaelsamenezes: Hm, I tried a bit with:
 
 ```cpp
template <size_t Index>
decltype(*(std::get<Index>(std::declval<interval_domaint::interval>())))
interval_domaint::get_interval_from_variant(const symbol2t &sym) const
{
  const auto &it = intervals->find(sym.thename);
  if (it != intervals->end())
  {
    return *std::get<Index>(it->second);
  }
  using IntervalRef = decltype(*std::get<Index>(std::declval<interval_domaint::interval>()));
  return IntervalRef(sym.type);
}
```

But it seems that the bottom bit is giving me a segmentation fault. Mainly because that type is a reference. I might consider just returning a shared ptr though.

fbrausse: Nah, that's different than what I intended, I'll give it a try tomorrow.

src/goto-programs/abstract-interpretation/interval_domain.h
rafaelsamenezes: I added the comment

src/goto-programs/abstract-interpretation/interval_domain.cpp
fbrausse: OK, never mind, I don't have a better solution right now. Thanks for addressing the second condition in the if-stmt.

src/goto-programs/abstract-interpretation/interval_domain.h
fbrausse: Can "key exists, but shared-pointer is null" ever happen or is it supposed to happen?

rafaelsamenezes:
No Intervals:

2024-03-19T19:38:40.6621949Z Statistics:           9537 Files
2024-03-19T19:38:40.6640396Z   correct:            4263
2024-03-19T19:38:40.6643318Z     correct true:     2609
2024-03-19T19:38:40.6645233Z     correct false:    1654
2024-03-19T19:38:40.6647107Z   incorrect:             7
2024-03-19T19:38:40.6650398Z     incorrect true:      7
2024-03-19T19:38:40.6651349Z     incorrect false:     0
2024-03-19T19:38:40.6652535Z   unknown:            5267
2024-03-19T19:38:40.6654529Z   Score:              6648 (max: 15923)

Master:

2024-03-25T13:08:37.8766712Z Statistics:           9537 Files
2024-03-25T13:08:37.8767709Z   correct:            4256
2024-03-25T13:08:37.8769303Z     correct true:     2674
2024-03-25T13:08:37.8770351Z     correct false:    1582
2024-03-25T13:08:37.8788326Z   incorrect:             6
2024-03-25T13:08:37.8789629Z     incorrect true:      6
2024-03-25T13:08:37.8790562Z     incorrect false:     0
2024-03-25T13:08:37.8791481Z   unknown:            5275
2024-03-25T13:08:37.8792467Z   Score:              6738 (max: 15923)

Optimized:

2024-03-25T13:01:10.3131124Z Statistics:           9537 Files
2024-03-25T13:01:10.3162537Z   correct:            4366
2024-03-25T13:01:10.3163486Z     correct true:     2716
2024-03-25T13:01:10.3164478Z     correct false:    1650
2024-03-25T13:01:10.3166164Z   incorrect:             7
2024-03-25T13:01:10.3167074Z     incorrect true:      7
2024-03-25T13:01:10.3167936Z     incorrect false:     0
2024-03-25T13:01:10.3168765Z   unknown:            5164
2024-03-25T13:01:10.3169631Z   Score:              6858 (max: 15923)

src/goto-programs/goto_contractor.h
rafaelsamenezes: removed it

src/goto-programs/abstract-interpretation/interval_domain.h
rafaelsamenezes: Added a new note

src/goto-programs/goto_contractor.h
rafaelsamenezes: Hm, apparently there is: `/esbmc/src/goto-programs/goto_contractor.h:637:21: error: ‘std::chrono’ has not been declared
  637 |     auto t_0 = std::chrono::steady_clock::now();``

I am double checking which change actually caused the chrono to be an issue now

rafaelsamenezes: This PR changed the includes of the goto_contractor.h so it does no longer include the interval_domain. I guess some part of that chain would also include the chrono

rafaelsamenezes:
The windows failure seems to be due to the current issue with xz: https://github.com/microsoft/vcpkg/issues/37906#issuecomment-2030111199

lucasccordeiro (approved):

PR #1756, opened 2024-03-20
[clang-cpp-frontend] Simplify if conditions

XLiZHI:
New functions are used to simplify the if condition.


fbrausse (changes requested):
I also thought about a change like this when looking into the reference handling, and I'm sure fine with it. Except: could we please stick to the lower-case function identifiers?

fbrausse (approved):
Thanks, LGTM.

lucasccordeiro (approved):

lucasccordeiro:

PR #1757, opened 2024-03-23
Python: Add support for len()

brcfarias:
- Add support for [len() ](https://docs.python.org/3/library/functions.html#len)based on the **__ESBMC_get_object_size** intrinsic function, currently restricted to `bytes` type.
- Refactor the generation of func_name/symbol_id in `get_function_call()`
- Handle negative indexes on subscript


lucasccordeiro (approved):
LGTM.

rafaelsamenezes (approved):

PR #1763, opened 2024-03-28
[python-fe] Replace abort() with warning for undefined functions

brcfarias:
Currently ESBMC crashes when encountering a call to an undefined function within Python programs. This PR replaces the crash behavior with printing a warning message containing the names of the undefined functions. This adjustment enables verification to proceed without interruption.


lucasccordeiro (approved):

PR #1764, opened 2024-04-02
[smt_convt] fiexed seg fault in error trace

### LINKED ISSUE #1743: Segmentation fault when handling struct member
esbmc outputs segmentation fault for this program:
```c
#include <assert.h>
typedef struct {
  int element[20];
  int head;
  int tail;
} QType;

QType queue;

int dequeue(QType *q) {
  int x = q->element[q->head];
  return x;
}

void foo() {
  if (dequeue(&queue))
    assert(0);
}

int main(void) {
  queue.head = 0;
  queue.tail = 0;
  queue.element[queue.tail] = 1;
  foo();
  return 0;
}
```

```
Solving with solver Boolector 3.2.1
Building error trace
```
If however, replace the 

```c
  int x = q->element[q->head];
  return x;
```
with 
```c
return q->element[q->head];
``` 
then it works fine.
```
[Counterexample]


State 1 file main1.c line 23 column 3 function main thread 0
  queue.element[(signed long int)queue.tail] = { .element={ 0 } WITH [0:=1], .head=0, .tail=0 }

State 5 file main1.c line 17 column 5 function foo thread 0
Violated property:
  file main1.c line 17 column 5 function foo
  assertion 0
  0
```

XLiZHI:
For the code:

`int x = q->element[q->head];`

seg fault when building error trace, this is because `q->head` has not been converted into `int`.

### DIFF
--- a/regression/esbmc-cpp/map/map_begin_bug/test.desc
+++ b/regression/esbmc-cpp/map/map_begin_bug/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 4 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/map/map_clear_bug/test.desc
+++ b/regression/esbmc-cpp/map/map_clear_bug/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 4 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/map/map_end_bug/test.desc
+++ b/regression/esbmc-cpp/map/map_end_bug/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 4 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/map/map_erase_bug/test.desc
+++ b/regression/esbmc-cpp/map/map_erase_bug/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 7 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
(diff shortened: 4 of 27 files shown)


lucasccordeiro (approved):
This is a very nice PR! Well spotted!

Can I ask you to run this PR over the SV-COMP benchmarks, too?

rafaelsamenezes (approved):

XLiZHI:
This PR:
```
  correct:           14042
    correct true:     7742
    correct false:    6300
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9721
  Score:             20904 (max: 38482)
 https://github.com/esbmc/esbmc/actions/runs/8529211991
```

Master:
```
  correct:           14003
    correct true:     7712
    correct false:    6291
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9760
  Score:             20835 (max: 38482)
  https://github.com/esbmc/esbmc/actions/runs/8529208038
```

lucasccordeiro:

fbrausse:
@XLiZHI I think this PR may have caused the current failure to run this regression test on Windows:
- regression/k-induction/cpp_priority_queue_size_bug

As a fix, in d55d4b53471f8776eed9fd957e8f61380879d2fd I've separated windows from the rest for a different TC. Shall we duplicate the above test into regression/windows as a KNOWNBUG as well?

*Edit:* It got unnoticed, because Windows was already failing during configure due to the `xz` disaster.

XLiZHI:
`regression/k-induction/cpp_priority_queue_size_bug`

The original result of this TC is seg fault and is therefore marked as KNOWNBUG in Linux, I checked that the error on window is a parsing error and I think we can mark it as THOROUGH.

fbrausse:
> `regression/k-induction/cpp_priority_queue_size_bug`
> 
> The original result of this TC is seg fault and is therefore marked as KNOWNBUG in Linux, I checked that the error on window is a parsing error and I think we can mark it as THOROUGH.

regression/k-induction/cpp_priority_queue_size_bug on current master is not KNOWNBUG but CORE (as done by this PR): https://github.com/esbmc/esbmc/blob/1813a9cbbe2c0016d79d310935d22a1f48abda8d/regression/k-induction/cpp_priority_queue_size_bug/test.desc#L1
On Windows it fails with some parsing error that's related to the weird expectancy of Clang-16. I'm asking to duplicate it as KNOWNBUG into regression/windows *edit:* and to mark it THOROUGH, yes.

PR #1768, opened 2024-04-03
[clang-cpp-frontend] Supported CXXThrowExpr node

### LINKED ISSUE #1767: [C++ Frontend] Conversion of unsupported clang expr: "CXXThrowExpr" to expression
````
Converting
ERROR: Conversion of unsupported clang expr: "CXXThrowExpr" to expression
CXXThrowExpr 0x62f86c8 <filename> 'void'
`-CXXConstructExpr 0x62f8698 <col:23, col:38> 'std::exception':'std::exception' 'void (const std::exception &) throw()' elidable
  `-MaterializeTemporaryExpr 0x62f8680 <col:23, col:38> 'const std::exception':'const std::exception' lvalue
    `-ImplicitCastExpr 0x62f8668 <col:23, col:38> 'const std::exception':'const std::exception' <NoOp>
      `-CXXBindTemporaryExpr 0x62f8598 <col:23, col:38> 'std::exception':'std::exception' (CXXTemporary 0x62f8598)
        `-CXXTemporaryObjectExpr 0x62f8560 <col:23, col:38> 'std::exception':'std::exception' 'void () throw()'
ERROR: CONVERSION ERROR
````
 

Here is the code if fails for:
````CPP
            #include <stdexcept>

            inline void UtAssert(std::string assert_info)
            {
                printf("Assert Failure: %s\n", assert_info.c_str());
                throw std::exception();
            }
````

### PR DESCRIPTION
**Pull request by XLiZHI**
This PR supports _try-catch-throw_ nodes and implements exception lists (for _throw_ now).

For the example code:

```
void myfunction()
{
  throw 1;
}

int main()
{
  myfunction();

  return 0;
}
```
Result:
```
[Counterexample]


State 1 file main8.cpp line 2 column 3 function myfunction thread 0
Violated property:
  file main8.cpp line 2 column 3 function myfunction
  Throwing an exception of type signed_int but there is not catch for it.


VERIFICATION FAILED
```


lucasccordeiro (approved):

rafaelsamenezes (commented):
I don't like this workaround. Users will think that the throw is actually being verified. 

I would prefer this to be converted into a throw instruction, it will probably crash at symex if you invoke the instruction though. So:

1. Generate a throw expression.
2. Add a knownbug where it crashes at symex.

This way we will only verify programs that don't reach a throw statement.

lucasccordeiro:
I agree with @rafaelsamenezes's comments.

XLiZHI:
_update:_ Now that I'm done with the try-catch part of the frontend, I went through our old frontend implementation and tried to reproduce the expressions (catch && thorw) it generated.

The reason I did this is that our [implementation](https://github.com/esbmc/esbmc/blob/master/src/goto-symex/symex_catch.cpp) in symex is all based on the old frontend, what still needs to be done:

1. Complete the exception list: Do we follow the old frontend to generate exception ids? https://github.com/esbmc/esbmc/blob/b3ab4e0e53888bef07f2ee96e3767d00567805d8/src/cpp/cpp_exception_id.cpp#L14
2. Check the fitness of the frontend and backend, it may need to be adjusted.

lucasccordeiro:
Many thanks, @XLiZHI.

> Complete the exception list: Do we follow the old frontend to generate exception ids?

I would say so. I think @mikhailramalho and @feliperodri implemented this exception list. 

I would also suggest reading this paper https://onlinelibrary.wiley.com/doi/10.1002/stvr.1793. I remember @feliperodri spent quite some time to write the try-catch rules to generate the exception ids.

XLiZHI:
Thank you @lucasccordeiro, This paper was very useful and I understand these implementation details. Regarding this PR, I do not want to complicate it, I will finish them in the subsequent PRs.

rafaelsamenezes (approved):

rafaelsamenezes:
> Thank you @lucasccordeiro, This paper was very useful and I understand these implementation details. Regarding this PR, I do not want to complicate it, I will finish them in the subsequent PRs.

I would be fine merging the current approach as a incremental solution

lucasccordeiro:
> I would be fine merging the current approach as a incremental solution.

I fully agree with @rafaelsamenezes.

lucasccordeiro:

Issue #1776, opened 2024-04-10
[C++ Frontend] Recursive, extending structs via pointers: `Assertion 'type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.`


intrigus-lgtm:
`a` refers to `b` via a pointer.
So when we try to parse the `a`, we will notice the field `c` with the type `b`.
Then we will try to get the type for `b`.
However, now `b` will try to get the type for its base class: `a`.
And now we will call `to_struct_type(class_a)` which will fail, because `a` is not yet a complete type.

I tried a bit to fix this, but wasn't very successful so far.

I think @fbrausse was working on recursive structs last time, so maybe he has an idea.

```cpp
class a;
class b;
class a {
  b *c;
};
class b : a {};
```

```shell
esbmc: esbmc/src/util/std_types.h:251: struct_typet& to_struct_type(typet&): Assertion `type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.
```

fbrausse:
This looks similar to #637, which ultimately was fixed by using the `type.incomplete()` flag in https://github.com/esbmc/esbmc/blob/7736aca9b39e7449d4409bd12863220eab730d79/src/clang-c-frontend/clang_c_convert.cpp#L357-L362 as a way to stop the recursion. Maybe the C++ frontend could also check this flag? Note that the flag is currently cleared directly after the return, so before the C++ convert functions are invoked. That clearing would need to be moved down to the end, so the flag retains its meaning for the C++ frontend.

----------------------

In the past different approaches to this tricky get_struct_union_class() were discussed, but they've been deemed too invasive. One maybe not so invasive way could be:
- get_struct_union_class() always stops at the above point, only leaving `incomplete_*` in the context until all top-level declarations are performed.
- Additionally, this method would put the new incomplete type (together with the Clang AST node) into a stack called `yet_to_be_defined`.
- At the end of convert_top_level_decl() there would be a loop iterating over `yet_to_be_defined` and pulling in the complete definitions via the second half of get_struct_union_class() that step 1 now skips. This loop needs to be a fixpoint search, because translating the complete type could put more incomplete ones onto the stack.

PR #1777, opened 2024-04-10
Handle recursive conversion in base type

fbrausse:
Turns out, I had a patch for #1776 laying around since a while. It's hard to get a good regression test as there seems to be a problem with accessing public members of a base class (1776-2 and 1776-3).


XLiZHI (approved):
LGTM

regression/esbmc-cpp/bug_fixes/1776-4/gh-1776.cc
```diff
@@ -0,0 +1,17 @@
...
+};
+class b : public a {};
+
+int main()
+{
+	b B;
+	a *A = &B;
+	A->c = A;
+	assert(A->c == A->c->c);
+}
```
XLiZHI, line 17: This TC can also pass in the current master.

fbrausse: Hm, good point, meaning that this TC is pretty useless for this PR. Probably worth investigating the failures of the other ones. For reference, with this PR 1776-2 prints:
```
Generating GOTO Program
ERROR: Looking up index of nonexistant member "c" in struct/union "b"
Aborted
```
and 1776-3 does this:
```
[Counterexample]


State 1 file gh-1776.cc line 15 column 2 function main thread 0
Violated property:
  file gh-1776.cc line 15 column 2 function main
  dereference failure: Access to object out of bounds


VERIFICATION FAILED
```

intrigus-lgtm:
The PR doesn't handle this test case:

```cpp
struct a;
struct b;
struct c
{
  void d(b);
};
struct a : c
{
  virtual ~a();
};
struct b : a
{
};
```

Crashes with (line number is not 100% accurate):
```shell
esbmc/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp:561: void clang_cpp_convertert::add_vtable_variable_symbols(const clang::CXXRecordDecl&, const struct_typet&, const switch_table&): Assertion `vt_symb_type' failed.
```

While this test case works fine (note that `b` does not extend `a` anymore):
```cpp
struct a;
struct b;
struct c
{
  void d(b);
};
struct a : c
{
  virtual ~a();
};
struct b /* : a */
{
};
```

intrigus-lgtm:
Thanks for creating the test case commit, which I've applied to my own PR: #1781.

fbrausse:
Superseded by #1781. Also: doesn't make a lot of sense deriving from an incomplete type. It's better to keep to the C/C++ parsing rules.

Issue #4, opened 2017-12-31
std:bad_cast on csmith


pmatos:
Hi,

Got a `std:bad_cast` on esbmc running on generated test for compilers.
Attached file. Ran command line:
```
$ installs/esbmc-v4.6.0-linux-static-64/bin/esbmc --k-induction prog.i
```
[prog.i.gz](https://github.com/esbmc/esbmc/files/1595211/prog.i.gz)

pmatos:
Happy new year!

pmatos:
Here's a simpler one to look at:

```
$ cat prog.i
union {
  short a
} b, *c = &b;
main() { *c; }
```

jmorse:
Hi Paulo,

Thanks for the fault report -- compiled without NDEBUG, a `is_struct_type(type)` assertion triggers in ` dereferencet::construct_struct_ref_from_const_offset_array`. It looks like this is due to incomplete union work. We recently changed the underlying storage of all unions to be byte arrays, and figured
 * All union field accesses become dereferences of non-union type, which we can handle
 * C99 assignments of unions become memcpy's from the clang frontend (I think)

Which avoids us having to ever construct a union during a dereference.

IIRC the C spec for your example requires `*c` to be "evaluated", but instead ESBMC goes too far by attempting to create a union out of the underlying byte array and croaks. The solution is probably to encode some pointer validity and bounds checks, and skip any reference building. Fitting that into existing abstractions is tricky, it'll probably be a few days until there's a patch.

(Happy new year!).

Issue #3, opened 2017-06-02
Invalid assertion on two dimension array value with non-determinstic index


guhetier:
ESBMC detect an invalid assertion violation when : 

- the assertion condition is about a value in a two dimension array
- an index get a non-deterministic value, that is forced to a specific value by pruning all values but one

ESBMC seems to process correctly assertions about the value of the non-deterministic index alone, or
about the value in the array with constant indexes.

```c
/* esbmc esbmc_array_nondet_err.c */
int main(void)
{
    unsigned int b;
    unsigned int a = 0;
    int table[2][2]  = {{1,0}, {1,0}};

    if (nondet_uint() == 0) {
        a = 1;
    } else {
        __ESBMC_assume(0);
    }
    __ESBMC_assert(table[1][0], "FOO"); /* Assertion ok */
    __ESBMC_assert(a == 1, "FOO"); /* Assertion ok */
    b = table[a][0];
    __ESBMC_assert(b, "ERROR"); /* Assertion violated */
    __ESBMC_assert(table[a][0], "ERROR"); /* arrays bound violated */
    return (0);
}
```

Trace : 

```
/*
ESBMC version 4.2.0 64-bit x86_64 linux
file esbmc_array_nondet_err.c: Parsing
Converting
Type-checking esbmc_array_nondet_err
Generating GOTO Program
GOTO program creation time: 0.074s
GOTO program processing time: 0.000s
Starting Bounded Model Checking
Symex completed in: 0.001s (27 assignments)
Slicing time: 0.000s
Generated 6 VCC(s), 5 remaining after simplification
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector arithmetic
Encoding to solver time: 0.003s
Solving with solver Boolector 2.4.1
Encoding to solver time: 0.003s
Runtime decision procedure: 0.000s
Building error trace

Counterexample:

State 1 file esbmc_array_nondet_err.c line 8 function main thread 0
c::main
----------------------------------------------------
  c::main::$tmp::return_value_nondet_uint$1=0 (0)

State 4 file esbmc_array_nondet_err.c line 17 function main thread 0
c::main
----------------------------------------------------
  esbmc_array_nondet_err::main::1::b=0 (0)

State 5 file esbmc_array_nondet_err.c line 18 function main thread 0
c::main
----------------------------------------------------
Violated property:
  file esbmc_array_nondet_err.c line 18 function main
  ERROR
  (_Bool)b

[...]

Issue #2, opened 2017-06-02
False "array bound violated error" of two dimension arrays


guhetier:
A failing assertion over a value on the second column of a two dimension array
raise the error `array bounds violated` instead of the assertion violation message.

A minimal example follows :

```c
/* esbmc esbmc_array_bound_err.c */
int main(void)
{ 
    int a = 1;
    int b = 0;
    int t[2][1]  = { {0}, {1}};
    __ESBMC_assert(t[a][b] == 0, "TEST1");
    return (0);
}
```

Error trace :
```
ESBMC version 4.2.0 64-bit x86_64 linux
file esbmc_array_bound_err.c: Parsing
Converting
Type-checking esbmc_array_bound_err
Generating GOTO Program
GOTO program creation time: 0.094s
GOTO program processing time: 0.002s
Starting Bounded Model Checking
Symex completed in: 0.003s (17 assignments)
Slicing time: 0.000s
Generated 5 VCC(s), 1 remaining after simplification
No solver specified; defaulting to Boolector
Encoding remaining VCC(s) using bit-vector arithmetic
Encoding to solver time: 0.000s
Solving with solver Boolector 2.4.1
Encoding to solver time: 0.000s
Runtime decision procedure: 0.002s
Building error trace

Counterexample:

State 1 file esbmc_array_bound_err.c line 7 function main thread 0
c::main
----------------------------------------------------
Violated property:
  file esbmc_array_bound_err.c line 7 function main
  array bounds violated: array `t'[] upper bound
  t[a][b] == 0

VERIFICATION FAILED
```

mikhailramalho:
Hum, apparently esbmc finds the property violation and it's printing the wrong message. I'm looking into it now.

mikhailramalho:
Fixed the issue. It will be released as part of ESBMC 4.2.2.

Thank you for the report.

mikhailramalho:
Hi,

The fixes were released as part of ESBMC 4.3.

PR #1781, opened 2024-04-13
Fix recursive structs. Fixes #1776

### LINKED ISSUE #1776: [C++ Frontend] Recursive, extending structs via pointers: `Assertion 'type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.`

intrigus-lgtm:
Previously, esbmc would try to parse fields of forward declarations.
Consider this code:
```cpp
class a; // https://github.com/esbmc/esbmc/pull/1
class b; // https://github.com/esbmc/esbmc/issues/2
class a { // https://github.com/esbmc/esbmc/issues/3
  b *c;
};
class b : a {}; // https://github.com/esbmc/esbmc/issues/4
```
When parsing https://github.com/esbmc/esbmc/pull/1, clang would resolve the actual definition for us.
This means that we would skip ahead to https://github.com/esbmc/esbmc/issues/3 and try to parse the fields
of `a`. When doing this, we will encounter `b` and try to get a type
for it. This will ultimately fail, because we are still defining `a`,
but `b` extends `a` and so `a` must be a _complete_ type which it is
not at that point.

We can avoid this problem by _not_ continuing to parse fields if the
declaration is not a complete one.
When now parsing https://github.com/esbmc/esbmc/pull/1, we would not skip ahead to https://github.com/esbmc/esbmc/issues/3 anymore.
Similarly, when parsing https://github.com/esbmc/esbmc/issues/2, we would not skip ahead to https://github.com/esbmc/esbmc/issues/4 anymore.
There is still a slight problem:
when parsing https://github.com/esbmc/esbmc/issues/3 we would however still have a problem.
We would try to get a type for `b` again, which would then fail
again, because `a` is not yet complete.

This problem is easily solved, by not recording structs in the
context if we already have a symbol there for it.


fbrausse (approved):
Thanks, LGTM. I also looked into it today and came to the same conclusion: follow the declaration/definition style of the Clang AST / the source code. I just didn't want to make the call in get_type() dependent on whether we already had the symbol and was looking for a Clang API that would let me tell the difference. Unsuccessfully. So I guess I'm fine with it, thanks again.

fbrausse:
Actually, maybe it makes sense to add a comment to this case in get_type():

-----------------------
This only happens when the definition of the record type is part of a declaration of a field / variable, as for the anonymous union in, e.g., /usr/include/bits/types/__mbstate_t.h
```c
typedef struct
{
  int __count;
  union
  {
    __WINT_TYPE__ __wch;
    char __wchb[4];
  } __value;		/* Value so far.  */
} __mbstate_t;
```
or in c2goto/library/libm/musl/libm.h
```c
#define asuint(f) ((union{float _f; uint32_t _i;}){f})._i
```
-----------------------

I still think there is a reason why this irregularity happens, which may have to do with the implementation of get_var(), but I can't pinpoint it, yet.

intrigus-lgtm:
Unfortunately this is still not a complete fix :(

```cpp
template <typename> struct a;
struct b;
template <typename> struct a { b *c; };
struct b : a<int> {};
```
(Basically the original testcase but this time with templates)

Fails again with 
```
Assertion `type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.
```

fbrausse:
> (Basically the original testcase but this time with templates)

I think it's worth trying to remove the get_struct_union_class() calls in the C++ frontend, specifically, in get_base_map(). There are two of them and - if we follow the decl/def style of the AST correctly - none should be necessary: you can't derive from an incomplete base class.

lucasccordeiro (approved):

lucasccordeiro:
@intrigus-lgtm: you can consider submitting another PR once you handle the new issue mentioned here.

PR #1789, opened 2024-04-23
Support builtin template

intrigus-lgtm:


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -275,6 +276,17 @@ bool clang_c_convertert::get_decl(const clang::Decl &decl, exprt &new_expr)
   case clang::Decl::Typedef:
     break;
 
+  case clang::Decl::BuiltinTemplate:
+  {
+    // expanded by clang itself
+    const clang::BuiltinTemplateDecl &btd =
+      static_cast<const clang::BuiltinTemplateDecl &>(decl);
+    assert(
```
rafaelsamenezes: An error message followed by an abort would be better here. This assert will not be triggered for release builds.

rafaelsamenezes (changes requested):
LGTM overall. Just a comment to reeplace an assert into an error + abort

src/clang-c-frontend/clang_c_convert.cpp
intrigus-lgtm: Fixed.

intrigus-lgtm:
TCs should now also pass after moving them to the correct directory and using the system c++ headers.

XLiZHI (approved):
LGTM

lucasccordeiro (approved):

PR #1791, opened 2024-04-23
[python] Variable initialization via ** operator and uint64() function

brcfarias:
Adding support for variables initializations from: 
- exponentiation operator (**)
```
  x = 2**4
  assert x == 16
```

- uint64() call
```
x = uint64(10)
assert x == 10
```


regression/python/constants/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: Can you add a test case that produces "VERIFICATION FAILED"?

lucasccordeiro (approved):
@brcfarias: I have just one minor comment!

lucasccordeiro:

PR #1793, opened 2024-04-23
[Help needed] [clang-cpp] Support ArrayInitLoopExpr

### LINKED ISSUE #1792: Support `ArrayInitLoopExprClass`.
esbmc is missing support for the [ArrayInitLoopExpr](https://clang.llvm.org/doxygen/classclang_1_1ArrayInitLoopExpr.html):

> Represents a loop initializing the elements of an array.
>
> The need to initialize the elements of an array occurs in a number of contexts:
>
>    **in the implicit copy/move constructor for a class with an array member**
>    when a lambda-expression captures an array by value
>    when a decomposition declaration decomposes an array
>
> There are two subexpressions: a common expression (the source array) that is evaluated once up-front, and a per-element initializer that runs once for each array element.
>
> Within the per-element initializer, the common expression may be referenced via an [OpaqueValueExpr](https://clang.llvm.org/doxygen/classclang_1_1OpaqueValueExpr.html), and the current index may be obtained via an [ArrayInitIndexExpr](https://clang.llvm.org/doxygen/classclang_1_1ArrayInitIndexExpr.html).

Testcase:

```cpp
#include <cassert>

struct a
{
  int b[3];
};
int main()
{
  a d{};
  assert(d.b[0] == 0);
  assert(d.b[1] == 0);
  assert(d.b[2] == 0);
  d.b[0] = 1;
  assert(d.b[0] == 1);
  assert(d.b[1] == 0);
  assert(d.b[2] == 0);

  int i = 0;
  a e{d};
  assert(e.b[0] == 1);
  assert(e.b[1] == 0);
  assert(e.b[2] == 0);
  e.b[2] = 20;
  assert(e.b[0] == 1);
  assert(e.b[1] == 0);
  assert(e.b[2] == 20);

  return 0;
}
```

intrigus-lgtm:
This PR _tries_ to fix #1792.

Unfortunately I could not get this to really work.
My proof of concept (basically: https://github.com/esbmc/esbmc/commit/6bad2dde73e66fa63ec01589e97bcaf3a04c7dd8#diff-784347884400427356b283749ef36cc6e458d4d4df4f8e3a2a4007244a637b1eR1057-R1176) which I implemented in `clang_cpp_convert.cpp` actually successfully verified the test case, however I would assume that this would be the wrong place for the code.

I was thinking that `clang_cpp_adjust_expr.cpp` would be a better place, so I tried to move my code there.
However it fails with `migrate expr failed`.

**I would really appreciate if someone could take a look at this or guide me because I _feel_ like this should be pretty simple to implement for someone that is more familiar with the code.**

The high level goal of the PR is to support code like this:
```cpp
struct a
{
  int b[3];
};
int main()
{
  a d{};
  a e{d}; // copy constructor
}
```

When copying `d` to `e`, the array `b` also has to be copied.
Clang represents that using an `ArrayInitLoopExpr`.
So basically this should compile to this code:
```cpp
Symbol......: c:@S@a@F@a#&1$@S@a#
Module......: cpptest.pre
Base name...: a
Mode........: C++
Type........: constructor  (struct a *, const struct a &)
Value.......: 
{

array_init_loop_expr_tmp1 = &a::ref->b[0];
this->b[0] = array_init_loop_expr_tmp1[0];
this->b[1] = array_init_loop_expr_tmp1[1];
this->b[2] = array_init_loop_expr_tmp1[2];

}
```

Currently it looks like this (maybe that is the problem?):
```cpp
Symbol......: c:@S@a@F@a#&1$@S@a#
Module......: cpptest.pre
Base name...: a
Mode........: C++
Type........: constructor  (struct a *, const struct a &)
Value.......: 
{

{
array_init_loop_expr_tmp1 = &a::ref->b[0];

this->b[0] = array_init_loop_expr_tmp1[0];

this->b[1] = array_init_loop_expr_tmp1[1];

this->b[2] = array_init_loop_expr_tmp1[2];

};

}
```

(The `dag_match_replace_helper` stuff can be moved elsewhere if needed)

### DIFF
--- a/src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
@@ -4,9 +4,10 @@
 #include <util/c_types.h>
 #include <util/destructor.h>
 #include <util/expr_util.h>
+#include "util/message.h"
 
 clang_cpp_adjust::clang_cpp_adjust(contextt &_context)
-  : clang_c_adjust(_context)
+  : clang_c_adjust(_context), tmp_symbol("cpp_adjust::", 0)
 {
 }
 
@@ -204,6 +205,12 @@ void clang_cpp_adjust::adjust_side_effect_assign(side_effect_exprt &expr)
     }
     adjust_expr(rhs);
   }
+  else if (lhs.type().is_array() && rhs.statement() == "array_init_loop_expr")
+  {
+    // adjust array initialization
+    adjust_array_init_loop_expr(lhs, rhs, expr);
+    adjust_expr(expr);
+  }
   else
     clang_c_adjust::adjust_side_effect(expr);
 }
@@ -281,4 +288,125 @@ void clang_cpp_adjust::align_se_function_call_return_type(
   const typet &return_type = (typet &)f_op.type().return_type();
   if (return_type.id() != "constructor" && return_type.is_not_nil())
     expr.type() = return_type;
+}
+
+static bool dag_match_replace_helper(
+  exprt &to_search,
+  exprt &replacement,
+  const std::function<bool(exprt &)> &is_match)
+{
+  if (is_match(to_search))
+  {
+    to_search = replacement;
+    return true;
+  }
+  else
+  {
+    for (auto &op : to_search.operands())
+    {
+      if (dag_match_replace_helper(op, replacement, is_match))
+      {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+// Assumes that pattern and to_search are both DAGs.
+// Expects that there is no more than one match.
+// Returns true if a match was found and replaced.
+static bool dag_match_replace_one(
+  exprt &to_search,
+  exprt &replacement,
+  const std::function<bool(exprt &)> &is_match)
+{
+  bool result = dag_match_replace_helper(to_search, replacement, is_match);
+  if (result)
+  {
+    assert(!dag_match_replace_helper(to_search, replacement, is_match));
+  }
+  return result;
+}
+
+void clang_cpp_adjust::adjust_array_init_loop_expr(
+  exprt &lhs,
+  exprt &rhs,
+  exprt &new_expr)
+{
+  assert(rhs.statement() == "array_init_loop_expr");
+  assert(lhs.type().is_array());
+
+  exprt source_array_expr = rhs.op0();
+
+  typet array_element_type = source_array_expr.type().subtype();
+  typet array_element_ptr_type = pointer_typet(array_element_type);
+
+  exprt array_init = side_effect_exprt("assign", lhs.type());
+  array_init.copy_to_operands(lhs, rhs);
+  symbolt source_array_symbol = tmp_symbol.new_symbol(
+    context, array_element_ptr_type, "array_init_loop_expr_tmp");
+  context.add(source_array_symbol);
+  source_array_symbol.lvalue = true;
+  source_array_symbol.value = source_array_expr;
+
+  exprt source_array_expr_evaluated = symbol_expr(source_array_symbol);
+
+  // Generate `source_array_expr_evaluated = source_array_expr;`
+  side_effect_exprt evaluate_source_array_stmt("assign");
+  // Cast from e.g. int[3] to int *
+  exprt source_array_typecasted = source_array_expr;
+  gen_typecast(ns, source_array_typecasted, array_element_ptr_type);
+  evaluate_source_array_stmt.type() = array_element_ptr_type;
+  evaluate_source_array_stmt.copy_to_operands(
+    source_array_expr_evaluated, source_array_typecasted);
+
+  code_blockt array_init_body = code_blockt();
+  convert_expression_to_code(evaluate_source_array_stmt);
+  array_init_body.operands().push_back(evaluate_source_array_stmt);
+
+  exprt per_element_initializer = rhs.op1();
+
+  // Replace reference to `source_array_expr` with `source_array_expr_evaluated` such that
+  // it is only evaluated once as per the clang docs.
+  if (!dag_match_replace_one(
+        per_element_initializer,
+        source_array_expr_evaluated,
+        [&](exprt &other) { return other == source_array_expr; }))
+  {
+    log_error("Failed to match expected pattern in {} {}", __func__, __LINE__);
+    abort();
+  }
+
+  exprt array_size_expr = rhs.op2();
+  assert(array_size_expr.is_constant());
+  int64_t array_size =
+    binary2integer(array_size_expr.value().as_string(), true).to_int64();
+  for (int64_t i = 0; i < array_size; i++)
+  {
+    exprt new_rhs;
+    new_rhs = per_element_initializer;
+
+    exprt pos = constant_exprt(i, index_type());
+
+    if (!dag_match_replace_one(
+          new_rhs,
+          pos,
+          [&](exprt &other)
+          { return other.name().operator==("__ARRAY_INIT_INDEX_EXPR__"); }))
+    {
+      log_error(
+        "Failed to match expected pattern in {} {}", __func__, __LINE__);
+      abort();
+    }
+
+    exprt new_lhs = index_exprt(lhs, pos, array_element_type);
+
+    array_init = side_effect_exprt("assign", lhs.type());
+    array_init.copy_to_operands(new_lhs, new_rhs);
+    convert_expression_to_code(array_init);
+    array_init_body.operands().push_back(array_init);
+  }
+  //  convert_expression_to_code(array_init_body);
+  new_expr = array_init_body;
 }
--- a/regression/esbmc/array-init-loop-expr/copy-constructor/main.cpp
+++ b/regression/esbmc/array-init-loop-expr/copy-constructor/main.cpp
@@ -1,0 +1,29 @@
+#include <cassert>
+
+struct a
+{
+  int b[3];
+};
+int main()
+{
+  a d{};
+  assert(d.b[0] == 0);
+  assert(d.b[1] == 0);
+  assert(d.b[2] == 0);
+  d.b[0] = 1;
+  assert(d.b[0] == 1);
+  assert(d.b[1] == 0);
+  assert(d.b[2] == 0);
+
+  int i = 0;
+  a e{d};
+  assert(e.b[0] == 1);
+  assert(e.b[1] == 0);
+  assert(e.b[2] == 0);
+  e.b[2] = 20;
+  assert(e.b[0] == 1);
+  assert(e.b[1] == 0);
+  assert(e.b[2] == 20);
+
+  return 0;
+}
--- a/regression/esbmc/array-init-loop-expr/copy-constructor/test.desc
+++ b/regression/esbmc/array-init-loop-expr/copy-constructor/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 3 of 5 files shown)


fbrausse:
Looks to me like this is generating a `code_blockt` inside an `code_expressiont`, which is unexpected. If you look at goto_convertt::convert_expression() you'll notice it never calls the top-level convert() again.

I also noticed you keep an empty `id` on the array_init_index_expr. That makes the .dump() output (and probably the current one from your symbol table) hard to read.

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
```diff
@@ -282,3 +289,124 @@ void clang_cpp_adjust::align_se_function_call_return_type(
...
+  typet array_element_type = source_array_expr.type().subtype();
+  typet array_element_ptr_type = pointer_typet(array_element_type);
+
+  exprt array_init = side_effect_exprt("assign", lhs.type());
+  array_init.copy_to_operands(lhs, rhs);
+  symbolt source_array_symbol = tmp_symbol.new_symbol(
+    context, array_element_ptr_type, "array_init_loop_expr_tmp");
+  context.add(source_array_symbol);
+  source_array_symbol.lvalue = true;
+  source_array_symbol.value = source_array_expr;
```
fbrausse, line 351: The .new_symbol() call already registers the symbol with the context (hence it returns a reference), no need to add it again. In fact, it should be accessed by reference, otherwise you're modifying a copy outside of the context with the last 2 statements.

mikhailramalho (changes requested):
hey @intrigus-lgtm I think we need to take a step back here and check if this is the correct approach. TBH, the adjust code seems a bit over-engineered.

Whenever we work on the frontend I always suggest we first check if the feature works in the old frontend, then on cbmc. If it does in either of them, half of the work is done.

For this one, cbmc does support the feature, so it's very likely that if we create an irep similar to the one there, the rest of esbmc will process it properly, in this case, the irep is:
```
  0: code
      * type: empty
      * statement: decl_block
      0: code
          * type: empty
          * #source_location: 
            * file: main2.cpp
            * line: 8
            * function: main
          * statement: decl
          0: symbol
              * type: struct_tag
                  * identifier: tag-a
              * identifier: main::1::d
              * #lvalue: 1
          1: struct
              * type: struct_tag
                  * identifier: tag-a
              * #source_location: 
                * file: main2.cpp
                * line: 8
                * function: main
              0: array
                  * type: array
                      * #member_name: tag-a
                      * size: constant
                          * type: signedbv
                              * width: 64
                              * #c_type: signed_long_int
                          * value: 3
                      0: signedbv
                          * width: 32
                          * #c_type: signed_int
                  * #source_location: 
                    * file: main2.cpp
                    * line: 8
                    * function: main
                  0: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 8
                        * function: main
                      * value: 0
                  1: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 8
                        * function: main
                      * value: 0
                  2: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 8
                        * function: main
                      * value: 0
```
Can you try to create a similar irep on the convert file?

mikhailramalho:
Bear in mind that esbmc has some different names for stuff (e.g, location instead of source_location), so you'll probably need to take a look around to see what's used in esbmc.

XLiZHI:
For reference, this is the symbol table of CBMC. 

```cpp
Symbol......: main
Pretty name.: main()
Module......: main8
Base name...: main
Mode........: C
Type........: signed int (void)
Value.......: {
  struct a d={ .a::b={ 0, 0, 0 } };

  struct a e;
  e = d;

}
Flags.......:
Location....: file main8.cpp line 5

Symbol......: main::1::d
Pretty name.: main::1::d
Module......: main8
Base name...: d
Mode........: cpp
Type........: struct a
Value.......: { .b={ 0, 0, 0 } }
Flags.......: lvalue thread_local file_local
Location....: file main8.cpp line 7 function main

Symbol......: main::1::e
Pretty name.: main::1::e
Module......: main8
Base name...: e
Mode........: cpp
Type........: struct a
Value.......: 
Flags.......: lvalue thread_local file_local
Location....: file main8.cpp line 8 function main
```

intrigus-lgtm:
> hey @intrigus-lgtm I think we need to take a step back here and check if this is the correct approach. TBH, the adjust code seems a bit over-engineered.
> 
> Whenever we work on the frontend I always suggest we first check if the feature works in the old frontend, then on cbmc. If it does in either of them, half of the work is done.
> 
> For this one, cbmc does support the feature, so it's very likely that if we create an irep similar to the one there, the rest of esbmc will process it properly, in this case, the irep is:
> 
> ```
> SNIP
>```
> 
> Can you try to create a similar irep on the convert file?

@mikhailramalho 
Can you please tell me how you tested this code in cbmc?
As far as I am aware, there is no C equivalent for the problem I'm trying to solve.
The problem is not with the "THIS" line, but with the "Hypothetical syntax" line that the `ArrayInitLoopExpr` basically represents.
```cpp
  struct a { int b[3]; };
  struct a d={};

  struct a e;
  e = d; // THIS
  e.b = d.b // Hypothetical syntax
```

I could also not get my example to work with the old c++ frontend; it seems that it doesn't like aggregate inits.

Wouldn't it be better to try to represent the clang ast as faithfully as possible instead of assuming that it will always have a certain shape?

mikhailramalho:
hey, so cbmc also verifies cpp files, so just create a foo.cpp file and paste your code there. To print the symbol table, run cbmc with --show-symbol-table. I also use this patch to print the ireps:
```
diff --git a/src/goto-programs/show_symbol_table.cpp b/src/goto-programs/show_symbol_table.cpp
index 4a5b870ab6..936cb9487c 100644
--- a/src/goto-programs/show_symbol_table.cpp
+++ b/src/goto-programs/show_symbol_table.cpp
@@ -83,11 +83,16 @@ void show_symbol_table_plain(
 
     std::string type_str, value_str;
 
-    if(symbol.type.is_not_nil())
+    if(symbol.type.is_not_nil()){
+      out << symbol.type.pretty() << '\n';
       ptr->from_type(symbol.type, type_str, ns);
+    }
 
-    if(symbol.value.is_not_nil())
+
+    if(symbol.value.is_not_nil()) {
+      out << symbol.value.pretty() << '\n';
       ptr->from_expr(symbol.value, value_str, ns);
+    }
 
     out << "Symbol......: " << symbol.name << '\n' << std::flush;
     out << "Pretty name.: " << symbol.pretty_name << '\n';
```
cbmc doesn't handle the cassert header either, so you'll need to use __CPROVER_assert instead.

The symbol table and irep for:
```
int main()
{
  struct a { int b[3]; };
  struct a d={};

  struct a e;
  e = d; // THIS
//  e.b = d.b // Hypothetical syntax
}
```
is
```
code
  * return_type: signedbv
      * width: 32
      * #c_type: signed_int
  * parameters: 
code
  * type: code
      * return_type: signedbv
          * width: 32
          * #c_type: signed_int
      * parameters: 
  * #source_location: 
    * file: main2.cpp
    * line: 4
    * function: main
  * statement: block
  0: code
      * type: empty
      * statement: decl_block
  1: code
      * type: empty
      * statement: decl_block
      0: code
          * type: empty
          * #source_location: 
            * file: main2.cpp
            * line: 6
            * function: main
          * statement: decl
          0: symbol
              * type: struct_tag
                  * identifier: main::1::tag-a
              * identifier: main::1::d
              * #lvalue: 1
          1: struct
              * type: struct_tag
                  * identifier: main::1::tag-a
              * #source_location: 
                * file: main2.cpp
                * line: 6
                * function: main
              0: array
                  * type: array
                      * #member_name: main::1::tag-a
                      * size: constant
                          * type: signedbv
                              * width: 64
                              * #c_type: signed_long_int
                          * value: 3
                      0: signedbv
                          * width: 32
                          * #c_type: signed_int
                  * #source_location: 
                    * file: main2.cpp
                    * line: 6
                    * function: main
                  0: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 6
                        * function: main
                      * value: 0
                  1: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 6
                        * function: main
                      * value: 0
                  2: constant
                      * type: signedbv
                          * width: 32
                          * #c_type: signed_int
                      * #source_location: 
                        * file: main2.cpp
                        * line: 6
                        * function: main
                      * value: 0
  2: code
      * type: empty
      * statement: decl_block
      0: code
          * type: empty
          * #source_location: 
            * file: main2.cpp
            * line: 8
            * function: main
          * statement: decl
          0: symbol
              * type: struct_tag
                  * identifier: main::1::tag-a
              * identifier: main::1::e
              * #lvalue: 1
  3: code
      * type: empty
      * #source_location: 
        * file: main2.cpp
        * line: 9
        * function: main
      * statement: expression
      0: side_effect
          * type: struct_tag
              * identifier: main::1::tag-a
          * #source_location: 
            * file: main2.cpp
            * line: 9
            * function: main
          * statement: assign
          * #lvalue: 1
          0: symbol
              * type: struct_tag
                  * identifier: main::1::tag-a
              * #source_location: 
                * file: main2.cpp
                * line: 9
                * function: main
              * identifier: main::1::e
              * #lvalue: 1
          1: symbol
              * type: struct_tag
                  * identifier: main::1::tag-a
              * #source_location: 
                * file: main2.cpp
                * line: 9
                * function: main
              * identifier: main::1::d
Symbol......: main
Pretty name.: main()
Module......: main2
Base name...: main
Mode........: C
Type........: signed int (void)
Value.......: {

  struct main::1::a d={ .main::1::a::b={ 0, 0, 0 } };

  struct main::1::a e;

  e = d;
}
Flags.......:
Location....: file main2.cpp line 3
```

intrigus-lgtm:
> hey, so cbmc also verifies cpp files, so just create a foo.cpp file and paste your code there.

Wow, I didn't know that. Thank you very much!
The [documentation](https://diffblue.github.io/cbmc/) does a pretty good job at hiding that fact^^
(The GitHub page does say that it also supports C++, but I missed that)

I'll try to look into this a bit more later this week/next week.

lucasccordeiro:
@intrigus-lgtm: Can I ask you whether you had a chance to advance with this PR? Please let us know if you need further support.

intrigus-lgtm:
@lucasccordeiro: sorry, I was super busy the last two weeks with (co-)organizing [GPN CTF](https://play.ctf.kitctf.de/).
Right now I'm trying to understand a different nasty bug related to references, but I do plan to look into this soonish.

lucasccordeiro:
@intrigus-lgtm: Did you have any chance to progress with this PR? ;-)

intrigus-lgtm:
> @intrigus-lgtm: Did you have any chance to progress with this PR? ;-)

I solved my problem, by replacing the offending array in the struct with an int128_t field.
This felt like a better use of my time, so I will likely not work on this anymore.

lucasccordeiro:
@intrigus-lgtm: Given that this PR is merged https://github.com/esbmc/esbmc/pull/2101, can I ask whether we could close your PR?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2101.

intrigus-lgtm:
> @intrigus-lgtm: Given that this PR is merged #2101, can I ask whether we could close your PR?

Ah yes, closing was fine.
Missed this one.

PR #1800, opened 2024-04-26
[python-frontend] Add type annotation to methods

brcfarias:
- Extending type annotation to include methods. Previously, only functions were considered.
- Updated variable lookup to search in the global scope when not found in local functions. Now this scenario can be handled:
```
x = 10

def func():
  y = x
```


lucasccordeiro (approved):
LGTM.

lucasccordeiro:

PR #1802, opened 2024-04-30
[clang-cpp] Fix member init ref

intrigus-lgtm:


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1085,17 +1085,27 @@ bool clang_cpp_convertert::get_function_body(
           {
             exprt lhs;
             // parsing non-static member initializer
-            if (get_decl_ref(*init->getMember(), lhs))
+            if (get_decl_ref(*init->getMember(), lhs, false))
```
XLiZHI: Maybe we can set an `irep_namet` to avoid adding a new variable? such as `lhs.set("#member_init", 1);`

lucasccordeiro:
@intrigus-lgtm: can I ask you to check @XLiZHI's comment before we merge this PR?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1085,17 +1085,27 @@ bool clang_cpp_convertert::get_function_body(
...
 
+            if (lhs.type().reference())
+            {
+              /* Insert a cast such that (psuedo code) `int x = 1; int &y = x` is converted to
+               `int x = 1; int &y = &x`.
+               (for _actual_ variables, this is done in `get_var`) */
+              gen_typecast(ns, rhs, lhs.type());
+              /* Remove the reference from the type, otherwise `convert_reference`
+               in `clang_cpp_adjust` will try to dereference it, which is wrong. */
+              lhs.type().set("#reference", false);
```
fbrausse: I'm not sure it's the best idea to remove the reference info here. Would it be possible to adapt convert_reference() (or rather adjust_reference()) such that for "assign" expressions to references, which are initializations, the dereferencing is not done?

mikhailramalho: I agree, can we also move this handling to cpp_adjust? How does cbmc handles this code?

mikhailramalho (changes requested):

intrigus-lgtm:
ESBMC (with patch):
```
Symbol......: c:@S@a@F@a#
Module......: main
Base name...: a
Mode........: C++
Type........: constructor  (struct a *)
Value.......: 
{
this->c = &this->b;

this->b = 42;

}
```

cbmc:
```
Symbol......: a::a(this)
Pretty name.: 
Module......: test
Base name...: a
Mode........: cpp
Type........: auto (struct a *) -> constructor 
Value.......: {
  this->c = &this->b;
  this->b = 42;
}
```

I am really not familiar with cbmc but this at least appears to be the area that is related to references and initialization:
https://github.com/diffblue/cbmc/blob/2da2c76fa5b061b0be85d6035fbe55d537b60d78/src/cpp/cpp_typecheck_conversions.cpp#L1601-L1618
https://github.com/diffblue/cbmc/blob/2da2c76fa5b061b0be85d6035fbe55d537b60d78/src/cpp/cpp_typecheck_code.cpp#L339-L372

src/clang-cpp-frontend/clang_cpp_convert.cpp
intrigus-lgtm: Is https://github.com/esbmc/esbmc/pull/1802/commits/bd332b490f2947d281e740bcbb746dcf0b50bba0 more in line with what you expected?

intrigus-lgtm: Or should the code rather be moved to https://github.com/esbmc/esbmc/blob/d564ede96e67d1fe073cd0a2dcc074539e5075fc/src/clang-c-frontend/clang_c_adjust_expr.cpp#L226-L230 ?
It is not always clear to me what belongs where if I'm honest.

mikhailramalho: I need some more time to check if this is the correct solution, let me give it a try later today or tomorrow and I'll let you know

lucasccordeiro (approved):

XLiZHI (approved):

PR #1804, opened 2024-05-01
[python-frontend] Converting "for v in range(start, end, step)"

brcfarias:
This PR addresses the conversion of for-range statements such as:
```
for x in range(1, 5, 1):
  print(x)
```

Our approach involves transforming this construct into a corresponding while loop with the following structure:
```
def ESBMC_range_next_(curr: int, step: int) -> int:
  return curr + step

def ESBMC_range_has_next_(curr: int, end: int, step: int) -> bool:
  return curr + step <= end

start = 1  # start value is copied from the first parameter of range call
has_next:bool = ESBMC_range_has_next_(1, 5, 1) # ESBMC_range_has_next_ parameters copied from range call
while has_next == True:
    print(start)
    start = ESBMC_range_next_(start, 1)
    has_next = ESBMC_range_has_next_(start, 5, 1)
```

This process involves the following steps:
1) The definition for **ESBMC_range_next_** and **ESBMC_range_has_next_** are provided as models in "_memory-models/range.py_". This file is bundled in the ESBMC binary and deployed via "flail" functions. These functions are loaded prior to the conversion of the main program, as we do for other languages (C/C++).
2) The conversion from the for-loop to the corresponding while loop is accomplished through AST manipulation in astgen.py.

### DIFF
--- a/regression/python/classes/main.py
+++ b/regression/python/classes/main.py
@@ -5,6 +5,9 @@ class MyClass:
     def foo(self) -> None:
         self.data = 1
 
+    def blah(self) -> None:
+        x:int = self.data
+
 
 obj1 = MyClass(5)
 assert(obj1.data == 5)
--- a/regression/python/classes/main.py
+++ b/regression/python/classes/main.py
@@ -2,6 +2,9 @@ class MyClass:
     def __init__(self, value: int):
         self.data:int = value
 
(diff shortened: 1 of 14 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #1808, opened 2024-05-02
[python-frontend] Support for class methods

brcfarias:
This PR adds support for: 

- Conversion of class methods and their corresponding calls:
```python
class MyClass:
    @classmethod
    def my_method(cls) -> int:
        return 1

assert MyClass.my_method() == 1
```

- [BLSFieldElement](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#custom-types) type, which corresponds to an uint256.

### DIFF
--- a/regression/python/class-methods-fail/main.py
+++ b/regression/python/class-methods-fail/main.py
@@ -1,0 +1,6 @@
+class MyClass:
+    @classmethod
+    def my_method(cls) -> int:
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #1811, opened 2024-05-03
SSA feature extractor

rafaelsamenezes:
As discussed in the local meetings, we could start checking other types of encoding that could be used. But first we should check how many benchmarks could be affected by this. So this PR basically prints features found in the SSA trace, we can then use benchexec to collect statistics about how many benchmarks are going to be affected.


src/esbmc/options.cpp
```diff
@@ -34,13 +34,15 @@ const struct group_opt_templ all_cmd_options[] = {
...
-    {"ssa-smt-trace", NULL, "print generated SMT during SMT encoding"},
+   {
+     {"quiet",
+      NULL,
+      "do not print unwinding information during symbolic execution"},
+     {"compact-trace", NULL, ""},
+     {"symex-trace", NULL, "print instructions during symbolic execution"},
+     {"ssa-trace", NULL, "print SSA during SMT encoding"},
+     {"ssa-smt-trace", NULL, "print generated SMT during SMT encoding"},
+     {"ssa-features-dump", NULL, "print features in the SSA (just before convertion)"},
```
intrigus-lgtm:
```suggestion
     {"ssa-features-dump", NULL, "print features in the SSA (just before conversion)"},
```

src/esbmc/options.cpp
```diff
@@ -34,13 +34,15 @@ const struct group_opt_templ all_cmd_options[] = {
...
+      NULL,
+      "do not print unwinding information during symbolic execution"},
+     {"compact-trace", NULL, ""},
+     {"symex-trace", NULL, "print instructions during symbolic execution"},
+     {"ssa-trace", NULL, "print SSA during SMT encoding"},
+     {"ssa-smt-trace", NULL, "print generated SMT during SMT encoding"},
+     {"ssa-features-dump", NULL, "print features in the SSA (just before conversion)"},
     {"symex-ssa-trace", NULL, "print generated SSA during symbolic execution"},
     {"goto2c", NULL, "translate the GOTO program to C"},
     {"show-goto-value-sets",
```
fbrausse, line 49: This is weird formatting. Not your PR. Just wondering whether clang-format is having a bad time with this options.cpp file and we should just do it in one go...

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+bool ssa_features::run(symex_target_equationt::SSA_stepst &steps)
+{
+
+  features.clear();
+
+  for (const auto &step : steps)
+  {
+    const expr2tc expressions[2] = {step.cond, step.guard};
+    for (const expr2tc &e : expressions)
```
fbrausse: You could use
```c++
for (const expr2tc &e : std::array { step.cond, step.guard })
```
to make it easier to read. If not that's fine, too.

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,96 @@
...
+  case expr2t::constant_floatbv_id:
+    features.insert(SSA_FEATURES::NON_INTEGER_NUMERAL);
+    break;
+    
+  case expr2t::expr_ids::mul_id:
+  case expr2t::expr_ids::div_id:
+  case expr2t::modulus_id:
+    if (!(is_constant_expr(dynamic_cast<const arith_2ops &>(*e).side_1) ||
+          is_constant_expr(dynamic_cast<const arith_2ops &>(*e).side_2)))
+      features.insert(SSA_FEATURES::NON_LINEAR);
```
fbrausse, line 49: Questionable choice: Is `(unsigned short)23 * 42` (i.e. a typecast) really a non-linear expression?

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,96 @@
...
+  case expr2t::ashr_id:
+  case expr2t::lshr_id:
+  case expr2t::bitor_id:
+  case expr2t::bitand_id:
+  case expr2t::bitxor_id:
+  case expr2t::bitnand_id:
+  case expr2t::bitnor_id:
+  case expr2t::bitnxor_id:
+    // TODO: shifts are linear if shifted by a constant size
+    features.insert(SSA_FEATURES::NON_LINEAR);
```
fbrausse: ~The shift operations are only non-linear if the RHS is non-constant.~

*Edit:* I somehow didn't read your TODO. Never mind.

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  e->foreach_operand([this](const expr2tc &next) {
+    check(next);
+});
+}
+
+
+void ssa_features::print_result() const
+{
+  log_status("SSA: Printing features");
```
fbrausse: Using the same prefix `SSA: ` for this "introductory" message here makes grepping the output for these SSA-feature messages more difficult. Is it necessary?

fbrausse (commented):
In general looks like a good idea which I do approve of. But I have a few comments, mostly about being too conservative.

src/goto-symex/features.cpp
rafaelsamenezes: Could you clarify? You commented on modulus. Also, shouldn't be simplified by the constant folding?

src/goto-symex/features.cpp
rafaelsamenezes: Removed it

src/goto-symex/features.cpp
rafaelsamenezes: Just changed

src/goto-symex/features.cpp
fbrausse: Sorry, I was referring to `mul_id`, which is also handled by this switch case, Github just defaults to 4 lines of context. Constant folding should do it, yes. Is it a necessary preprocessing step to this algorithm?

rafaelsamenezes: > Sorry, I was referring to `mul_id`, which is also handled by this switch case, Github just defaults to 4 lines of context. Constant folding should do it, yes. Is it a necessary preprocessing step to this algorithm?

Ow I see, you are right. Well, I could add a `contains_symbolic(const expr2tc &e)` into irep utils and use it then. Guess that is safer.

fbrausse: Well, the question is, what "SSA contains non-linear features" really means. Consider for instance `a := b / 0`. This cannot be simplified because it's undefined. But if `b` was a constant, contains_symbolic() would return false. How would that correspond to "contains non-linear features"? Maybe an easier approach is to call the simplification directly and ask whether the result is "entirely constant" meaning, that it positively only contains constant_*2t expressions and defined functions (such as typecasts and no signed overflows). This predicate could be extended later, no need to get it perfect right now in this PR.

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,130 @@
...
+  if (features.count(SSA_FEATURES::BITWISE_OPERATIONS))
+    log_status("SSA: Contains BITWISE OPERATIONS");
+  if (features.count(SSA_FEATURES::OVERFLOW_ASSERTIONS))
+    log_status("SSA: Contains OVERFLOW ASSERTIONS");
+  if (features.count(SSA_FEATURES::ARRAY))
+    log_status("SSA: Contains ARRAY");
+  if (features.count(SSA_FEATURES::STRUCTS))
+    log_status("SSA: Contains STRUCTS");
+}
+bool ssa_features::is_entirely_constant(const expr2tc &e)
```
fbrausse, line 102: Can I haz a newline before the function? :)

src/goto-symex/features.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+  default:
+    break;
+  }
+
+  bool result = true;
+  e->foreach_operand([this, &result](const expr2tc &next) {
+    result &= is_entirely_constant(next); });
+
+  return result;
```
fbrausse, line 126: The `1 / 0` example from my previous comment isn't handled here, yet, is it?

fbrausse (approved):
Generally LGTM, though I have a question.

lucasccordeiro (approved):

src/goto-symex/features.cpp
rafaelsamenezes: Not really, but I thought this was the sort of thing that we would figure out later.

fbrausse: Fine by me, just wanted to clarify. Maybe it makes sense to add a TODO comment.

rafaelsamenezes: I added a TODO in the division operator

PR #1812, opened 2024-05-03
Fix unnamed parameter

intrigus-lgtm:
Previously, when an identifier was empty, we would skip the loop iteration
by `continue`-ing it. This however had the effect that the iterator `it1`
would not be advanced, because that statement was at the end of the loop
and we just skipped it.
Move the advancement of the iterator such that it is always executed once
per loop iteration regardless of whether we might skip parts of the current
iteration.

The end result of not advancing the iteration was that we would confuse arguments and e.g. dereference the wrong one.
`regression/esbmc-cpp/cpp/unnamed-parameter/main.cpp` would fail, because `*c` in `b(*c)` would not dereference `&f` (the correct argument) but `0` which would fail!

### DIFF
--- a/regression/esbmc-cpp/cpp/unnamed-parameter-fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/unnamed-parameter-fail/main.cpp
@@ -1,0 +1,15 @@
+#include <cassert>
+
+struct a
+{
+  a(int, int *c) : b(*c)
+  {
+  }
+  int b;
+};
+int f = 42;
+int main()
+{
+  a a(0, &f);
+  assert(a.b == 11);
+}
--- a/regression/esbmc-cpp/cpp/unnamed-parameter-fail/test.desc
+++ b/regression/esbmc-cpp/cpp/unnamed-parameter-fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/cpp/unnamed-parameter/main.cpp
+++ b/regression/esbmc-cpp/cpp/unnamed-parameter/main.cpp
@@ -1,0 +1,15 @@
+#include <cassert>
+
+struct a
+{
+  a(int, int *c) : b(*c)
+  {
+  }
+  int b;
+};
+int f = 42;
+int main()
+{
+  a a(0, &f);
+  assert(a.b == 42);
+}
--- a/regression/esbmc-cpp/cpp/unnamed-parameter/test.desc
+++ b/regression/esbmc-cpp/cpp/unnamed-parameter/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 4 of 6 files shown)


mikhailramalho (approved):
ouch! nice catch

ChenfengWei0:
Hi @intrigus-lgtm, thank you very much for submitting this PR. The PR also somehow fixed a KNOWNBUG regression test in the Solidity constructor, which leads to the regression test failure:
```
1103 - regression/esbmc-solidity/constructor_4 (Failed)
```
To fix this, may I ask if you could change the KNOWNBUG to CORE? In `regression/esbmc-solidity/constructor_4/test.desc` first line: ~~KNOWNBUG~~ to CORE. Many thanks.

intrigus-lgtm:
> Hi @intrigus-lgtm, thank you very much for submitting this PR. The PR also somehow fixed a KNOWNBUG regression test in the Solidity constructor, which leads to the regression test failure:
> 
> ```
> 1103 - regression/esbmc-solidity/constructor_4 (Failed)
> ```
> 
> To fix this, may I ask if you could change the KNOWNBUG to CORE? In `regression/esbmc-solidity/constructor_4/test.desc` first line: ~KNOWNBUG~ to CORE. Many thanks.

(Running the tests locally with 16 cores sometimes crashes my pc due to a weird interaction between zfs and the kernel when under memory pressure, so I usually don't run the full test suite before opening a PR)

I had a quick look at the test case and it has IR like this:

```
symbol
 * type: code
     * arguments: 
       0: argument
           * type: pointer
               * subtype: empty
                   * #cpp_type: void
       1: argument
           * type: unsignedbv
               * width: 8
           * #location: 
             * file: contract.sol
             * line: 7
             * function: BB
           * #base_name: x
           * #identifier: sol:@C@BB@F@BB@x#5
       2: argument
           * type: unsignedbv
               * width: 8
           * #location: 
             * file: contract.sol
             * line: 7
             * function: BB
           * #base_name: y
           * #identifier: sol:@C@BB@F@BB@y#7
     * return_type: empty
         * #cpp_type: void
 * name: BB
 * identifier: sol:@C@BB@F@BB#
 * #location: 
 * #lvalue: 1
```

For this function:
```solidity
    function empty(uint8 x, uint256 y) public {
        data = data + x;
        empty1();
    }
```

Notice the 0-th (implicit, I guess) parameter, which doesn't have a name.
Previously, the arguments would be shifted incorrectly.

ChenfengWei0:
> Notice the 0-th (implicit, I guess) parameter, which doesn't have a name.
> Previously, the arguments would be shifted incorrectly.

Yes, you are right. 

This empty pointer is used to "cheat" the esbmc frontend, i.e. bypassing the type-check and format-check in the frontend. If we re-write this function `empty` in C++, the symbol table would be:
```
Symbol......: c:@S@BB@F@empty#I#I#
Module......: 1
Base name...: empty
Mode........: C++
Type........: void (struct BB *, int, int)
...
```
The first argument `struct BB *` stands for the `this` pointer.  Thus the empty pointer should be something like
```
0: argument
    * type: pointer
        * subtype: symbol
            * identifier: tag-BB
    * #base_name: this
    * #identifier: c:@S@BB@F@BB#&1$@S@BB#this
```
instead of void.

lucasccordeiro (approved):

lucasccordeiro:

PR #1815, opened 2024-05-07
[C++ verification] supported more type of expction var

XLiZHI:
This RP supports exception ids of struct and array types:

1. Add a record for the base class to the type expression of the struct/class.
2. Exception variables are no longer initialized. (The initialized expression is incomplete for the struct type)
3. Ajusted the format of THROW_DECL_END goto function.
4. Fixed the empty expression in rethrow.
5. Adjusted C++get body function: `void func() try{ }catch{ }` In this case there is no block, so it is skipped.

esbmc-cpp/try_catch pass rate: 75%
old frontend: 88%

### DIFF
--- a/regression/esbmc-cpp/try_catch/ch13_1/test.desc
+++ b/regression/esbmc-cpp/try_catch/ch13_1/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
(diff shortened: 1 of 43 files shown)


lucasccordeiro (approved):

lucasccordeiro:

Issue #1818, opened 2024-05-08
Release ESBMC v7.6


lucasccordeiro:
Dear all,

I would like to release a new ESBMC version 7.6.

Do you have any PRs to consider for this release?

fbrausse:
I'm good with the current state. No urgent PRs from my side.

ChenfengWei0:
I'm also good with the current state. 
- The Solidity "mapping and built-in member" PR is nearly done. However, since the `realloc` issue is unfixed, I think we should not merge it into 7.6.
- The condition coverage PR still needs a lot of improvement and bug fixes. So it should not be merged into 7.6 either.

brcfarias:
I don't have any pull requests to consider. I'm ok with the release for ESBMC.

fbrausse:
I believe there is nothing left to do here. :)

lucasccordeiro:
Many thanks, @fbrausse.

PR #1819, opened 2024-05-08
[C++ verification] supported noexcept

XLiZHI:
This PR support `noexcept` and fixed bugs:

1. Fixed catch order: `catch(base){} catch(base1){}`,  Assuming that both base and base1 can be caught, only the first catch block is now jumped.
2. Fixed `throw decl end`: `void a {throw_decl(); return; throw_decl_end; }` throw decl end will not execution in this case. Now this PR will be cleared when the function ends.
3. Fixed pointer type catch: catch(int &i), Pointer types don't use rename: `signed int *i = &thrown_obj`.

#1818

### DIFF
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -1177,28 +1177,30 @@ bool clang_cpp_convertert::get_function_body(
   auto *type = fd.getType().getTypePtr();
   if (const auto *fpt = llvm::dyn_cast<const clang::FunctionProtoType>(type))
   {
-    if (fpt->hasDynamicExceptionSpec())
+    if (fpt->hasExceptionSpec())
     {
-      // e.g: void func() throw(int) { throw 1;}
-      // body is converted to
-      // {THROW_DECL(signed_int) throw 1; THROW_DECL_END}
       codet decl = codet("throw_decl");
-      for (unsigned i = 0; i < fpt->getNumExceptions(); i++)
+      if (fpt->hasDynamicExceptionSpec())
+      {
+        // e.g: void func() throw(int) { throw 1;}
+        // body is converted to
+        // {THROW_DECL(signed_int) throw 1; THROW_DECL_END}
+        for (unsigned i = 0; i < fpt->getNumExceptions(); i++)
+        {
+          codet tmp;
+          if (get_type(fpt->getExceptionType(i), tmp.type()))
+            return true;
+
+          decl.move_to_operands(tmp);
+        }
+      }
+      else if (fpt->hasNoexceptExceptionSpec())
       {
         codet tmp;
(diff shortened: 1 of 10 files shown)


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1177,28 +1177,30 @@ bool clang_cpp_convertert::get_function_body(
...
-      for (unsigned i = 0; i < fpt->getNumExceptions(); i++)
+      if (fpt->hasDynamicExceptionSpec())
       {
-        codet tmp;
-        if (get_type(fpt->getExceptionType(i), tmp.type()))
-          return true;
+        // e.g: void func() throw(int) { throw 1;}
+        // body is converted to
+        // {THROW_DECL(signed_int) throw 1; THROW_DECL_END}
+        for (unsigned i = 0; i < fpt->getNumExceptions(); i++)
```
lucasccordeiro, line 1188: shall we use `size_t` instead of `unsigned` here?

src/goto-symex/symex_main.cpp
```diff
@@ -167,6 +167,14 @@ void goto_symext::symex_step(reachability_treet &art)
   case END_FUNCTION:
     symex_end_of_function();
 
+    if (stack_catch.size())
```
lucasccordeiro: Shall we use `empty()` instead of `size()` to check if stack_catch is not empty?

src/goto-symex/symex_main.cpp
```diff
@@ -167,6 +167,14 @@ void goto_symext::symex_step(reachability_treet &art)
...
+    {
+      // Get to the correct try (always the last one)
+      goto_symex_statet::exceptiont *except = &stack_catch.top();
```
lucasccordeiro, line 173: Shall we use `back()` instead of `top()`?

lucasccordeiro (approved):
I have minor comments only.

src/clang-cpp-frontend/clang_cpp_convert.cpp
XLiZHI: According to https://clang.llvm.org/doxygen/classclang_1_1FunctionProtoType.html, `getExceptionType (unsigned i) const`, I think `unsigned` avoids the type cast.

src/goto-symex/symex_main.cpp
XLiZHI: `stack_catch` used a stack container and I think `back()` works for sequence containers.

src/goto-symex/symex_main.cpp
XLiZHI: done

fbrausse (approved):

mikhailramalho:
eventually we should get rid of these instructions and move the exception specification/noexcept to the goto_functiont and framet

lucasccordeiro:

Issue #1824, opened 2024-05-10
Inconsistencies within C++ library


fbrausse:
This file includes all headers for our OMs of the C++ standard library: [stl.h](https://github.com/esbmc/esbmc/files/15272770/stl.txt) The following are commented out, because including them would lead to errors like duplicate or inconsistent definitions:
- [x] bitset
- [x] fstream
- [x] locale
- [x] memory
- [x] sstream
- [x] valarray

This issue here is meant to track the missing support for the above.

PR #1825, opened 2024-05-10
Fix some C++ APIs: memory, ios and istream

fbrausse:
This PR:
- removes duplicate definitions of std::fill() from \<memory\>, fixing usage of that header for #1824
- adds operator>> support for std::istream and types (un)signed char and (un)signed long long (the latter for C++11 only)
- adds std::hexfloat() and std::defaultfloat() to \<ios\> for C++11

I've added a regression test using the std::istream operator>> APIs.

### DIFF
--- a/regression/esbmc-cpp11/cpp/istream-api/test-istream.cc
+++ b/regression/esbmc-cpp11/cpp/istream-api/test-istream.cc
@@ -1,0 +1,63 @@
+
+#include <istream>
+
+std::ios & g(std::ios &i)
+{
+	return i;
+}
+
+void f(std::istream &is)
+{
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):
Nice.

Can you also add a regression test that produces "VERIFICATION FAILED"?

fbrausse:
> Can you also add a regression test that produces "VERIFICATION FAILED"?

Hi @lucasccordeiro. At this point I'm just trying to get the API right to allow inclusion of more C++ headers. There is no new functionality behind it, yet. I'd rather deal with that in another PR.

lucasccordeiro:

PR #1828, opened 2024-05-12
[Solidity] Support Solidity built-in variables and functions

ChenfengWei0:
This PR added support for most of the built-in variables and functions in Solidity, including `tx`, `block`, `msg`, et al. Specifically, we created a Solidity template to implement these functions in C and parsed it as the intrinsic library.

It also contains some unfinished codes for the Solidity mapping, which should be finished and submitted in another PR.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3062,6 +3233,17 @@ bool solidity_convertert::get_type_description(
...
   }
+  case SolidityGrammar::TypeNameT::MappingTypeName:
+  {
+    // e.g.
+    //  "typeIdentifier": "t_mapping$_t_uint256_$_t_string_storage_$",
+    //  "typeString": "mapping(uint256 => string)"
+    // since the key will always be regarded as string, we only need to obtain the value type.
+
+    typet val_t;
+    //!TODO
```
lucasccordeiro: Can you explain here for this TODO? What should be done?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -3415,7 +3597,7 @@ bool solidity_convertert::get_elementary_type_name(
   }
   case SolidityGrammar::ElementaryTypeNameT::BYTES:
   {
-    new_type = uint_type();
+    new_type = unsignedbv_typet(256);
```
lucasccordeiro, line 3603: Why 256?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -156,6 +156,10 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
     {
       return TupleTypeName;
     }
+    if (typeIdentifier.substr(0, 10) == "t_mapping(")
```
lucasccordeiro: Shall we use `std::string::compare` function instead of `substr` for more efficient comparison?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -225,13 +225,16 @@ TypeNameT get_type_name_t(const nlohmann::json &type_name)
...
-    else if (
-      type_name["typeIdentifier"].get<std::string>().find("ArrayToPtr") !=
-      std::string::npos)
+    else if (typeIdentifier.find("ArrayToPtr") != std::string::npos)
     {
       // ArrayToPointer decay in DeclRefExpr when dereferencing an array, e.g. a[0]
       return PointerArrayToPtr;
     }
+    // for Special Variables and Functions
+    else if (typeIdentifier.substr(0, 7) == "t_magic")
```
lucasccordeiro: Shall we use the `std::string::compare` function instead of `substr` for a more efficient comparison?

src/solidity-frontend/solidity_grammar.cpp
```diff
@@ -686,10 +696,15 @@ ExpressionT get_expression_t(const nlohmann::json &expr)
...
     if (type_name == SolidityGrammar::TypeNameT::StructTypeName)
       return StructMemberCall;
-    if (type_name == SolidityGrammar::TypeNameT::EnumTypeName)
+    else if (type_name == SolidityGrammar::TypeNameT::EnumTypeName)
       return EnumMemberCall;
-    if (type_name == SolidityGrammar::TypeNameT::ContractTypeName)
+    else if (type_name == SolidityGrammar::TypeNameT::ContractTypeName)
       return ContractMemberCall;
+    else
+      //!Fixme. Assume it's a builtin member
```
lucasccordeiro: Shall we replace `fixeme` with `TODO`?

src/solidity-frontend/solidity_grammar.h
```diff
@@ -303,8 +309,8 @@ enum ExpressionT
   // rule Tuple
   Tuple,
 
-  // rule revert
-  Revert,
+  // Mapping
```
lucasccordeiro: Why do we need this comment? You could expand it further instead.

src/solidity-frontend/solidity_template.h
```diff
@@ -0,0 +1,386 @@
...
+
+const std::string sol_header = R"(
+#include <stddef.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+)";
+
+/*
+  uint == uint256_t
```
lucasccordeiro, line 23: Do they contain the same number of bits?

  uint == uint256_t
  bytes == uint256_t
  bytes32 == uint256_t

lucasccordeiro (approved):
I have just minor comments.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: That is the unfinished mapping codes. Will add a log_error and abort

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: In solidity, Bytes = Bytes32. And we further convert the bytes32 to uint256 in the front end.

src/solidity-frontend/solidity_grammar.cpp
ChenfengWei0: Will do. Thanks for the suggestion.

src/solidity-frontend/solidity_template.h
ChenfengWei0: Yes, I believe so.

src/solidity-frontend/solidity_grammar.h
ChenfengWei0: My mistake. It should be `rule Mapping`, something like:
```c++
enum ExpressionT{
  // rule identifier
  DeclRefExprClass,

  // rule literal
  Literal,

  // rule Tuple
  Tuple,

  // rule Mapping
  Mapping,
}
```

Basically, they are used to identify the type of the expression.

src/solidity-frontend/solidity_grammar.cpp
ChenfengWei0: Will do

ChenfengWei0:
Changed as suggested. Thanks for the suggestions, @lucasccordeiro .

lucasccordeiro:

fbrausse (commented):
I would suggest making `sol_library` a stand-alone file that will be included like the c2goto source/header or the cpp headers.

ChenfengWei0:
> I would suggest making `sol_library` a stand-alone file that will be included like the c2goto source/header or the cpp headers.

Sure, I will take a look. Thanks for the suggestions.

PR #1832, opened 2024-05-13
Fix some more C++ APIs: bitset, csignal and string_view

fbrausse:
This PR:
- makes \<bitset\> usable with other C++ headers, see #1824
- makes \<csignal\> interoperable with \<vector\> (though it should eventually (a) be moved to \<signal.h\> and (b) ~needs some regression tests to confirm it works at all~ *edit:* it has: esbmc-cpp/cpp/ch20_4, esbmc-cpp/cpp/ch20_8 and esbmc-cpp/OM_sanity_checks/csignal)
- fixes a missing return in std::string_view.


lucasccordeiro (approved):

lucasccordeiro:

PR #1839, opened 2024-05-16
[Solidity] Bug fixing and code refactoring

### LINKED ISSUE #1823: [Solidity] Unexpected ContractMemberCall
```solidity
pragma solidity >=0.5.0;

contract Base {
    function UO_Minus(int x) public {
        assert(x == -10);
    }

    function UO_Not(uint16 x) public {
        assert(~x == 65535);
    }

    function UO_LNot(int x) public {
        assert(!(x == 0));
    }
}

contract Derived {
    function test_op() public {
        Base y = new Base();
        int x = 10;
        uint16 xx = 10;

        y.UO_Minus(-x);
        y.UO_Not(0);
        y.UO_LNot(x);
    }
}
```
command: `esbmc --sol example.sol example.solast --k-induction`

result:
```
...
 @@@ got Expr: SolidityGrammar::ExpressionT::CallExprClass
 @@@ got Expr: SolidityGrammar::ExpressionT::ContractMemberCall
ERROR: Unexpected ContractMemberCall
 @@@ got Expr: SolidityGrammar::ExpressionT::UnaryOperatorClass
 @@@ got Expr: SolidityGrammar::ExpressionT::DeclRefExprClass
 @@@ got Expr: SolidityGrammar::ExpressionT::CallExprClass
 @@@ got Expr: SolidityGrammar::ExpressionT::ContractMemberCall
ERROR: Unexpected ContractMemberCall
 @@@ got Expr: SolidityGrammar::ExpressionT::Literal
 @@@ got Expr: SolidityGrammar::ExpressionT::CallExprClass
 @@@ got Expr: SolidityGrammar::ExpressionT::ContractMemberCall
ERROR: Unexpected ContractMemberCall
 @@@ got Expr: SolidityGrammar::ExpressionT::DeclRefExprClass
Generating GOTO Program
GOTO program creation time: 0.619s
GOTO program processing time: 0.009s
Unwinding loop 40 iteration 1
Not unwinding loop 40 iteration 2
Unwinding loop 41 iteration 1
Not unwinding loop 41 iteration 2
Symex completed in: 0.010s (60 assignments)
Slicing time: 0.000s (removed 39 assignments)
Generated 14 VCC(s), 8 remaining after simplification (21 assignments)
esbmc: /root/workspace/ESBMC_Commit/esbmc/src/solvers/boolector/boolector_conv.cpp:466: virtual const smt_ast* boolector_convt::mk_eq(smt_astt, smt_astt): Assertion `a->sort->get_data_width() == b->sort->get_data_width()' failed.
[1]    31597 IOT instruction  ./esbmc --sol example.sol example.solast --unwind 2
```

ChenfengWei0:
- Bug fix: #1823
- fix some potential bugs related to bytes.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2274,12 +2272,13 @@ bool solidity_convertert::get_binary_operator_expr(
   }
   else if (expr.contains("leftExpression"))
   {
-    nlohmann::json commonType = expr["commonType"];
+    nlohmann::json literalType_l = expr["leftExpression"]["typeDescriptions"];
+    nlohmann::json literalType_r = expr["rightExpression"]["typeDescriptions"];
 
-    if (get_expr(expr["leftExpression"], commonType, lhs))
+    if (get_expr(expr["leftExpression"], literalType_l, lhs))
```
fbrausse, line 2278: I didn't check the differences in the implementations of the two get_expr() versions, one with a `common_type` parameter and one without it. But to me it seems that this change (and also the `"rightExpression"` just below) are in fact not using a "common type" anymore, but the type of whatever expression is already passed to get_expr(). Maybe these should call the 2-parameter version without the common type?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -2469,25 +2465,56 @@ bool solidity_convertert::get_binary_operator_expr(
...
+      // e.g.
+      //    data1 = "test"; data2 = 0x74657374; // "test"
+      //    assert(data1 == data2); // true
+      // Do type conversion before the bswap
+      // the arguement of bswap should only be int/uint type, not string
+      // e.g. data1 == "test", it should not be bswap("test")
+      // instead it should be bswap(0x74657374)
+      if (!is_bytes_type(lhs.type()))
+      {
+        if (get_expr(expr["leftExpression"], expr["commonType"], lhs))
```
fbrausse, line 2487: I wonder why parsing the expression again, but this time with a possibly different "common type" is necessary. It has already been translated (see comment above), no? If so, isn't the "convert_type_expr" happening below already taking care of casting to the common type?

fbrausse (commented):
This `expr.contains("commonType")` handling happens quite often with this PR. I wonder whether that could be simplified/merged into a single implementation. Also I have a few more specific questions below.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: I just tested the two-parameter version and it failed almost every regression test. I believe we still need the type information (`literalType`) when parsing the literal.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: This is mainly for the string, (may also affect hex variable)
- So when not involving bytes, the string literal should be parsed as a char array by function `convert_string_literal()`
- When the string is used to initialise a byte type variable (`bytes1 x = "a";`), the string literal should be parsed as a hex value by function `convert_hex_literal()`, where the corresponding hex value is already in the JSON node.
- The `literalType` is used to decide which function to call. Therefore we would call the get_expr again and overwrite the `lhs/rhs` if the type of the literal is byte.

The `convert_type_expr` does a different job, mainly handling the implicit typecast E.g. the var type is bytes16, and commonType is bytes4 , the `convert_type_expr`  will perform the conversion `bytes16` => `bytes4`

ChenfengWei0:
I have simplified and removed some of the codes related to `"commonType` as suggested.

src/solidity-frontend/solidity_convert.cpp
fbrausse: Alright. Thanks for checking. This might mean, though, that the type parameter, which is currently named `expr_common_type` is not really a "common" type. But never mind, this is independent of this PR.

src/solidity-frontend/solidity_convert.cpp
fbrausse: I see, many thanks for this explanation! Maybe some of it could go into comments (if not already there).

fbrausse (approved):

lucasccordeiro (approved):

lucasccordeiro:

Issue #1836, opened 2024-05-15
Issues including algorithm with cpp 11


rafaelsamenezes:
The following program is failing with `--cppstd 11`

```cpp
#include <algorithm>
 
int main()
{
    return 0;
}
```

Fails with:

```
Parsing main.cpp
In file included from main.cpp:1:
/tmp/esbmc-cpp-headers-88a1-0b31-4ee7/algorithm:1169:12: error: no member named 'swap' in namespace 'std'
      std::swap(*j, *(j - 1));
```

lucasccordeiro:
@rafaelsamenezes : You're using an old version of ESBMC. With the current version, you should get:

````C
$ esbmc alex3.cpp --std c++11
ESBMC version 7.6.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing alex3.cpp
Converting
Generating GOTO Program
GOTO program creation time: 0.730s
GOTO program processing time: 0.002s
Starting Bounded Model Checking
Symex completed in: 0.003s (19 assignments)
Slicing time: 0.000s (removed 19 assignments)
Generated 0 VCC(s), 0 remaining after simplification (0 assignments)
BMC program time: 0.003s

VERIFICATION SUCCESSFUL
````

**Comment by fbrausse** [comment]
There is an #if switch in the \<algorithm\> / \<utility\> header that conditionally defers to \<type_traits\> in order to provide std::swap for C++ >= 11. We don't provide our own \<type_traits\> (anymore). Thus, it might be that the installed header doesn't contain a std::swap definition. I'd need to know the package/version installed in order to fix this. Inquiry has been sent.

**Comment by fbrausse** [comment]
Just got a confirmation that their GCC-12.1 is installed in a non-standard location. We might need to pass `--gcc-install-dir` and/or `--gcc-toolchain` through to Clang in order or it to select the right GCC installation.

I actually would argue for a general pass-through option like `-Xclang $OPT` or `-Wc,$OPTS` s.t. we can easily let users add options to the clang command line related to their environment.

**Comment by fbrausse** [comment]
Indeed, old version being selected, GCC-4.8 (note, this is ESBMC v7.6, see #1837):
```
ESBMC version 7.5.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing main.cpp
[...]

PR #1840, opened 2024-05-16
Interpret option -Wc,OPT1,OPT2,... as passing OPT1 OPT2 ... to the clang frontend directly

fbrausse:
This is meant to be used for non-standard options required by certain setups in order to, for instance in #1836, pass through options like --gcc-install-dir when a custom GCC installation in a non-standard path is being used.


lucasccordeiro (approved):

lucasccordeiro:

PR #1843, opened 2024-05-18
[python-frontend] Operational model for int.from_bytes()

brcfarias:
- Adding support for [int.from_bytes](https://docs.python.org/3/library/stdtypes.html#int.from_bytes)
```
>> int.from_bytes(b'\x00\x10', byteorder='big')
16
```

- Given that the frontend still doesn't support strings, the AST is being manipulated to replace the "byteorder" parameter with a boolean "is_bigendian".
- "bytes" array data is being encoded in base64 because [json.dump()](https://github.com/esbmc/esbmc/blob/ece7849b815b754df83ff833e6089d7d902d0e8c/src/python-frontend/astgen.py#L128) adds escape characters for non-printable ASCII characters. Additionally, in some cases it was not possible to diferentiate text from numerical values (e.g. "AAFF"). Now, "bytes" data is encoded in base64, and we store the corresponding int value after decoding in python_converter.
- Adding new ["consensus"](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types) type mappings.


regression/python/int_from_bytes/main.py
```diff
@@ -0,0 +1,40 @@
+bytes_data = b'\x00\x10'
+x:int = int.from_bytes(bytes_data, 'big', False)
+assert x == 16
+
+# bytes_data_2 = b'\x00\x10'
```
lucasccordeiro: Do we really need to keep these commented statements?

src/python-frontend/models/int.py
```diff
@@ -0,0 +1,71 @@
+class int:
+  @classmethod
+  def from_bytes(cls, bytes_data:bytes, big_endian:bool, signed:bool) -> int:
+    result:int = 0
+    index:int = 0
+    step:int = 1
+    byte:int = 0
+
+    ## If little endian
```
lucasccordeiro, line 9: What about Big Endian?

src/python-frontend/preprocessor.py
```diff
@@ -0,0 +1,45 @@
+import ast
```
lucasccordeiro, line 1: Please add some comments to this code.

src/python-frontend/python_converter.cpp
```diff
@@ -518,26 +519,37 @@ function_id python_converter::build_function_id(const nlohmann::json &element)
...
+  // build symbol_id
   if (func_name == "len")
   {
     func_name = __ESBMC_get_object_size;
     func_symbol_id = "c:@F@" + __ESBMC_get_object_size;
   }
+  else if (is_builtin_type(obj_name))
+  {
+    class_name = obj_name;
+    func_symbol_id =
```
lucasccordeiro: Shall we consider using a `std::stringstream` for building the func_symbol_id string to be more efficient and readable?

src/python-frontend/python_converter.cpp
```diff
@@ -724,7 +737,11 @@ exprt python_converter::get_function_call(const nlohmann::json &element)
         c_typecastt c_typecast(ns);
         c_typecast.implicit_typecast(arg, pointer_typet(empty_typet()));
       }
-      call.arguments().push_back(arg);
+
+      if (arg.type().is_array())
```
lucasccordeiro, line 743: Add some comments here.

src/python-frontend/python_converter.cpp
```diff
@@ -1001,10 +1026,18 @@ void python_converter::get_var_assign(
   {
     // Get type from current annotation node
     size_t type_size = 0;
-    if (
-      ast_node["value"].contains("value") &&
-      ast_node["value"]["value"].is_string())
-      type_size = ast_node["value"]["value"].get<std::string>().size();
+    if (ast_node["value"].contains("value"))
```
lucasccordeiro, line 1031: Shall we consider adding some error handling here?

src/python-frontend/python_converter.cpp
```diff
@@ -1575,18 +1611,29 @@ bool python_converter::convert()
...
 
-  // Load memory models -----
-  std::stringstream model_path;
-  model_path << ast_json["ast_output_dir"].get<std::string>() << "/range.json";
-  std::ifstream model_file(model_path.str());
-  nlohmann::json model_json;
-  model_file >> model_json;
+  if (!config.options.get_bool_option("no-library"))
+  {
+    // Load operational models -----
```
lucasccordeiro, line 1618: We could print a message to the ESBMC user saying that we are loading the operational model with the name of the model. What do you think?

lucasccordeiro (changes requested):

src/python-frontend/models/int.py
brcfarias: Both endiannesses are working. I updated the [regression test](https://github.com/brcfarias/esbmc/blob/e76dff43fcff187194e43b812793e1d03402ca1a/regression/python/int_from_bytes/main.py#L3C1-L7C17) to cover it.

src/python-frontend/python_converter.cpp
brcfarias: I don't think so in this case. The information of type_size being obtained in this if statement is only necessary when dealing with arrays. If it's not an array, type_size can simply remain zero because it's not used in `get_typet`.

src/python-frontend/python_converter.cpp
brcfarias: That sounds good. I'll make this modification.

brcfarias: image

lucasccordeiro (approved):

PR #1848, opened 2024-05-24
TLDR build instructions

rafaelsamenezes:


brcfarias:
Could we suggest simply running **scripts/build.sh**, given that we have something similar there?

lucasccordeiro (approved):

rafaelsamenezes:
> Could we suggest simply running **scripts/build.sh**, given that we have something similar there?

I am not supportive of that because build.sh was designed to be "CI friendly", and it has some issues for common developers/users:

- It tries to build ESBMC in two main configurations: Academic (GPL-compliant and non-comercial) and Comercial. These builds come with every little optional feature that we have support.
- It installs tons of dependencies to get around the old stable clang that we use.

I think that most users that compile ESBMC to not care about these optional features. What they need is:
- CMake flags used (which I can give to my IDE/editor)
- A LLVM installation
- A solver installed

We could try to make build.sh more user friendly, but IMO it is kinda redundant as the effort would be better off in explaining all dependencies and build configurations. I expect that users that care about this to be advanced and to be able to handle their CMake.

brcfarias (approved):
Perfect. Thanks for the explanation.

PR #1851, opened 2024-05-27
<exception>, <stdexcept> remove global std::message + add ctors from std::string + fix std::round in <cmath>

### LINKED ISSUE #1849: std::string not defined via <stdexcept>
This was reported by an industrial partner.
```c++
#include <stdexcept>

int main()
{
  std::string s;
}
```
ESBMC gives
```
stdexcept-string.cc:5:8: error: no type named 'string' in namespace 'std'
  std::string s;
  ~~~~~^
ERROR: PARSING ERROR
```
Should be a simple fix in our <stdexcept> header.

### LINKED ISSUE #1850: std::round not defined in <cmath>
This was reported by an industrial partner.
```c++
#include <cmath>

int main()
{
	return std::round(0.5);
}
```
ESBMC gives
```
round.cc:5:9: error: no member named 'round' in namespace 'std'; did you mean simply 'round'?
        return std::round(0.5);
               ^~~~~~~~~~
               round
```
A simple `using ::round` + the overloads for float and long double should be enough here.

fbrausse:
Fixes #1849 and fixes #1850. Also fixed 3 knownbug, but added 1:
- regression/esbmc-cpp/try_catch/ch13_3

Seems to me that something with access to class member of virtual classes is going wrong. It claims there would be out-of-bounds access (and a legitimate unhandled throw of std::bad_alloc).


lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:

PR #1856, opened 2024-06-03
Add <math.h> to C++ headers to support old glibc

fbrausse:
Older versions of glibc define 'struct exception' for the deprecated matherr(3) System V API - unless __cplusplus is defined. As undefining __cplusplus is only required for LLVM's libc++ (and not the default libstdc++ for glibc), make undefining it conditional on libc++ usage.

As a side note: it might be interesting to run our regression test suite also for libc++ instead of the default libstdc++. With #1840 merged, this could now easily be achieved with `ESBMC_OPTS="-Wc,-stdlib=libc++"`.


fbrausse:
Regarding the failed clang-format CI check: Could we start using [IndentPPDirectives: AfterHash](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#indentppdirectives) to ease readability with nested preprocessor directives such as this?
```diff
@@ -6,16 +6,16 @@
 /* Support LLVM's libc++, which defines the float and long double overloads
  * in the top-level :: namespace in <math.h> when __cplusplus is set. These
  * clash with our definitions due to the exception declaration. */
-# pragma push_macro("__cplusplus")
-# undef __cplusplus
+#pragma push_macro("__cplusplus")
+#undef __cplusplus
 extern "C"
 {
-# include_next <math.h>
+#include_next <math.h>
 }
-# pragma pop_macro("__cplusplus") // restore macro
+#pragma pop_macro("__cplusplus") // restore macro
 
 #else
 
-# include_next <math.h>
+#include_next <math.h>
 
 #endif
```

lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:
@fbrausse: do you have a regression test for this PR?

fbrausse:
> do you have a regression test for this PR?

Unfortunately not. Neither locally nor on our CI is the glibc old "enough" to exhibit the error. Newer versions of glibc do not define the deprecated matherr(3) nor the accompanying `struct exception` anymore.

lucasccordeiro:
Many thanks, @fbrausse.

PR #1858, opened 2024-06-04
Fix include search path order

fbrausse:
This PR corrects the order of the C/C++ include search paths Clang uses. In particular, for C++ they now look like this:
1. any user-specified `-I` from left to right
2. our C++ library headers, corresponds to: src/cpp/library/{CUDA,Qt,Qt/Core}
3. our C++ standard library and system headers, corresponds to: src/cpp/library
4. our C standard library and system headers, corresponds to: src/c2goto/headers
5. the system's C++ standard library, e.g., /usr/include/c++/v1 for libc++ or /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/{.,x86_64-pc-linux-gnu,backward} for the default libstdc++
6. the Clang resource directory's /include, e.g. /usr/lib/llvm/16/bin/../../../../lib/clang/16/include/../include
7. the default system include directories, e.g. /usr/local/include and /usr/include
8. any user-specified `--idirafter` from left to right

Before, the order was 1, 3, 2, 4, 6, 5, 7, 8 leading to the following error for just `#include <cstddef>` when used with LLVM's libc++ via `-Wc,-stdlib=libc++`:
```
In file included from ../debug/test-cmath.cc:1:
/usr/include/c++/v1/cstddef:46:5: error: <cstddef> tried including <stddef.h> but didn't find libc++'s <stddef.h> header.           This usually means that your header search paths are not configured properly.           The header search paths should contain the C++ Standard Library headers before           any C Standard Library, and you are probably using compiler flags that make that           not be the case.
#   error <cstddef> tried including <stddef.h> but didn't find libc++'s <stddef.h> header. \
    ^
/usr/include/c++/v1/cstddef:59:9: error: no member named 'nullptr_t' in the global namespace
using ::nullptr_t;
      ~~^
ERROR: PARSING ERROR
```

This PR fixes it. Note that `build_compiler_args()` now is not called from the clang_c_languaget constructor anymore allowing the C++ version to properly override it. Also, the vector built by it is no longer cached, which shouldn't be a problem since we construct a new *_languaget object for each source file anyways.

Additionally, I've switched from regex-replacing `_Bool` with `bool` for C++ frontend to a pre-processor based replacement, basically to allow `_Bool` to be part of identifiers.

*Edit:* This PR also makes our C++ \<math.h\> header more robust by including \<cstddef\> before checking libc++-specific macros.


fbrausse:
Master:
```
  correct:           14188
    correct true:     7800
    correct false:    6388
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9575
  Score:             21108 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/9366268385
```
This PR:
```
  correct:           14193
    correct true:     7803
    correct false:    6390
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9570
  Score:             21116 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/9366270080
```

lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:

PR #1867, opened 2024-06-09
[Solidity] Support mapping data structure

ChenfengWei0:
This PR adds support for the mapping data structure of mapping and fixes some previously hidden bugs.

---
remain issues:
- All verification results are correct. However, in some cases, it's significantly slower when running the library functions in Solidity than in C, which led to a timeout error. (In C, it's super fast)
  - E.g. it's unwinding the `strcmp` unstoppablely. I might test the inifite_array solution if I cannot resolve this speed issue, but I am not sure if this would help.
- An array out of bounds false positive when dereferencing the string constant.
---
Incomplete part: mapping towards complex data structure e.g. struct/contract. It should not be hard to implement, but I want to solve the speed issue first.

---

About the idea of moving the solidity_template.c to c2goto.  Since the CMakelist uses a wildcard to match .c files, I tried creating a `solidity.c` file under the `src/c2goto/library/`. However, the library did not appear in the symbol table, thus for now I am sticking to the current solution.


fbrausse (approved):
Looks good to me.

As a general note, not for this PR: The contents the strings in solidity_template.h should really be moved to a separate file, similar to what we do with /library and /headers in src/c2goto.

lucasccordeiro (approved):

lucasccordeiro:

ChenfengWei0:
> Looks good to me.
> 
> As a general note, not for this PR: The contents the strings in solidity_template.h should really be moved to a separate file, similar to what we do with /library and /headers in src/c2goto.

Yes, I agree. I tried yet failed. Will try to do it again.

Issue #1866, opened 2024-06-07
[cpp-clang] Broken multiple-inheritance


intrigus-lgtm:
esbmc (32ab7a8680c08f8172ad3a0c5796695f67dc4155) fails for this very simple code:

```
#include <assert.h>

struct B
{   
  char e = 22;
};
struct A
{
  char b = 111;
};
struct f : B, A
{   
};
int main()
{
  f i = f();
  assert(i.b == 111);
  assert(i.e == 22);
}
```
Fails with this log:
<details>

```
Generating GOTO Program
GOTO program creation time: 0.437s
GOTO program processing time: 0.000s
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

__ESBMC_pthread_start_main_hook (c:@F@__ESBMC_pthread_start_main_hook):
        // 0 no location
        FUNCTION_CALL:  yield()
        // 1 file pthread_lib.c line 118 column 3 function __ESBMC_pthread_start_main_hook
        ATOMIC_BEGIN
        // 2 file pthread_lib.c line 119 column 3 function __ESBMC_pthread_start_main_hook
        num_total_threads=num_total_threads + 1;
        // 3 file pthread_lib.c line 120 column 3 function __ESBMC_pthread_start_main_hook
        num_threads_running=num_threads_running + 1;
        // 4 file pthread_lib.c line 121 column 3 function __ESBMC_pthread_start_main_hook
        ATOMIC_END
        // 5 file pthread_lib.c line 122 column 1 function __ESBMC_pthread_start_main_hook
        END_FUNCTION // __ESBMC_pthread_start_main_hook
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

__ESBMC_pthread_end_main_hook (c:@F@__ESBMC_pthread_end_main_hook):
        // 6 no location
        FUNCTION_CALL:  yield()
        // 7 file pthread_lib.c line 130 column 3 function __ESBMC_pthread_end_main_hook
        ATOMIC_BEGIN
        // 8 file pthread_lib.c line 131 column 3 function __ESBMC_pthread_end_main_hook
        num_threads_running=num_threads_running - 1;
        // 9 file pthread_lib.c line 132 column 1 function __ESBMC_pthread_end_main_hook
        END_FUNCTION // __ESBMC_pthread_end_main_hook
[...]

PR #1873, opened 2024-06-14
[clang-cpp] move struct to pointer conversion to adjuster.

intrigus-lgtm:
While investigating #1866 I noticed this code which IMHO shouldn't be a concern of a function that performs typecasting.
While any problematic struct to pointer conversion should be caught by the compiler (besides the conversion of foo.bar() to bar(&foo)), keeping this code would likely complicate my fix for #1866.

### DIFF
--- a/src/clang-cpp-frontend/clang_cpp_adjust.h
+++ b/src/clang-cpp-frontend/clang_cpp_adjust.h
@@ -47,6 +47,7 @@ public:
   void adjust_side_effect_assign(side_effect_exprt &expr);
   void adjust_function_call_arguments(
     side_effect_expr_function_callt &expr) override;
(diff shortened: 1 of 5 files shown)


mikhailramalho:
Can you add a test case fixed by this change?

intrigus-lgtm:
I found https://github.com/esbmc/esbmc/pull/1907 while working on this, but I don't have a test case per se that is fixed.

I was initially under the impression that this would allow a cleaner fix for https://github.com/esbmc/esbmc/issues/1866 but have now come to the conclusion that this is not the case.
Also, I had funny problems where some tests failed and some test passed depending on the order of calling `clang_c_adjust::adjust_function_call_arguments(expr);`.

So feel free to close this PR.

PR #1874, opened 2024-06-15
[python-frontend] Basic support for strings

brcfarias:
Adding basic support for **strings**. Specifically, it is now possible to initialize variables with strings, index string elements and compare them with characters, and check strings length.

In future commits, we will add support for other operations such as concatenation, substrings, and more. A complete list of operations is available here: https://docs.python.org/3/library/stdtypes.html#textseq


lucasccordeiro (approved):

lucasccordeiro:

PR #1877, opened 2024-06-16
[z3-smt] never pass nullptr as name. Breaks `--smt-model`

### LINKED ISSUE #1876: `--smt-model` fails with z3 on `regression/esbmc-cpp/stack/stack_class`
ESBMC crashes with a SIGSEGV when running `regression/esbmc-cpp/stack/stack_class` with the additional `--smt-model` flag and using z3 as a solver.

When giving the smt model, we will at some point land here:
https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/model/model_smt2_pp.cpp#L135

from which we will then get the name of the function.
https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/model/model_smt2_pp.cpp#L142
```cpp
unsigned len  = pp_symbol(out, c->get_name());
```

pp_symbol will then try to do a `strlen` on the name of the function.
https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/model/model_smt2_pp.cpp#L46
```cpp
return static_cast<unsigned>(strlen(s.bare_str()));
```

This will horribly fail if we have passed a nullptr to it as a name.

intrigus-lgtm:
Fixes #1876.

I have no idea, what the original purpose of passing a nullptr could have been.
I basically just did what yices also does:
https://github.com/esbmc/esbmc/blob/749b9673062870fdf86c5ee349ac6948f5cc772a/src/solvers/yices/yices_conv.cpp#L969

### DIFF
--- a/regression/z3/github_1876/gh-1876.cpp
+++ b/regression/z3/github_1876/gh-1876.cpp
@@ -1,0 +1,61 @@
+#include <stack>
+#include <string>
+#include <cassert>
+using namespace std;
+
+template <class T>
+class A
+{
+public:
+  A()
+  {
+    this->a = true;
+  }
+  bool get()
+  {
+    return this->a;
+  }
+  void set(bool x)
+  {
+    this->a = x;
+  }
+
+private:
+  bool a;
+};
+
+bool operator!=(A<bool> x, A<bool> y)
+{
+  if (x.get() != y.get())
+    return true;
+  return false;
+}
+
+bool operator==(A<bool> x, A<bool> y)
+{
+  if (x.get() == y.get())
+    return true;
+  return false;
+}
+
+bool operator<(A<bool> x, A<bool> y)
+{
+  return false;
+}
+
+int main()
+{
+  stack<A<bool>> first;
+  A<bool> x;
+  A<bool> y;
+  A<bool> z;
+  first.push(x);
+  first.push(y);
+  first.push(z);
+  first.pop();
+  assert(!(first.empty()));
+  assert((first.top()).get());
+  assert(first.size() == 2);
+  assert(false);
+  return 0;
+}
--- a/regression/z3/github_1876/test.desc
+++ b/regression/z3/github_1876/test.desc
@@ -1,0 +1,4 @@
+THOROUGH
+gh-1876.cpp
+--unwind 1 --smt-model --no-unwinding-assertions --timeout 900 --z3
+^VERIFICATION FAILED$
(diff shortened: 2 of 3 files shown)


lucasccordeiro (commented):
Can you please also add a regression test?

mikhailramalho:
what does the documentation say about passing a nullptr when creating a constant? what about an empty name?

intrigus-lgtm:
> Can you please also add a regression test?

Done.

> what does the documentation say about passing a nullptr when creating a constant? what about an empty name?

I did not find any documentation that says anything at all.
If we look at the source code, it actually does handle the nullptr case:
https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/util/symbol.h#L98-L112

However, it doesn't do it here where the crash is:
https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/model/model_smt2_pp.cpp#L33-L48

Reporting this to z3 upstream would be possible, but I don't really consider that a good use of my time, considering that yices does it in the exact same way (after this pr is merged).

lucasccordeiro (approved):

mikhailramalho:
> > Can you please also add a regression test?
> 
> Done.
> 
> > what does the documentation say about passing a nullptr when creating a constant? what about an empty name?
> 
> I did not find any documentation that says anything at all. If we look at the source code, it actually does handle the nullptr case: [Z3Prover/z3@`01e47bf`/src/util/symbol.h#L98-L112](https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/util/symbol.h#L98-L112)
> 
> However, it doesn't do it here where the crash is: [Z3Prover/z3@`01e47bf`/src/model/model_smt2_pp.cpp#L33-L48](https://github.com/Z3Prover/z3/blob/01e47bfe263f8eabf2daa2da35c7b248ee33fff9/src/model/model_smt2_pp.cpp#L33-L48)
> 
> Reporting this to z3 upstream would be possible, but I don't really consider that a good use of my time, considering that yices does it in the exact same way (after this pr is merged).

It's always a great use of anyone's time to report these things to the developers, even more so because this is an open-source project and we are using their tool for free. Nothing is more fair than helping them improve their work.

lucasccordeiro:
> It's always a great use of anyone's time to report these things to the developers, even more so because this is an open-source project and we are using their tool for free. Nothing is more fair than helping them improve their work.

@intrigus-lgtm: I agree with @mikhailramalho. Nikolaj Bjorn is a very nice colleague. I'm sure that Nikolaj will be very happy to receive your issue so that he can improve the Z3 codebase.

intrigus-lgtm:
Better late than never :)
https://github.com/Z3Prover/z3/issues/7789

lucasccordeiro:
Nice, @intrigus-lgtm ;-)

PR #1878, opened 2024-06-17
[c++ verification] supported CXXTypeidExpr node and fixed bugs

XLiZHI:
This PR adds new node support: `CXXTypeidExpr`.

Fixed two bugs:

1. Avoid type conversion if the type of the function argument is the same as the destination type.
2. Fixed an error caused by calculating `class` size in adjusting `new` expression, added missing namespace.

For example:
```
class C {
public:
    C(const std::string &a) : a(a) {}
private:
    std::string a;
};

new C("a");
```
we need to compute the size of all members of this struct (e.g. calss: `std::string`), so `ns` is necessary.
https://github.com/esbmc/esbmc/blob/749b9673062870fdf86c5ee349ac6948f5cc772a/src/util/type_byte_size.cpp#L95-L99


lucasccordeiro (approved):

intrigus-lgtm:
> Fixed two bugs:
> 
>     1. ...
> 
>     2. Fixed an error caused by calculating `class` size in adjusting `new` expression, added missing namespace.
> 
> 
> For example:
> 
> ```
> class C {
> public:
>     C(const std::string &a) : a(a) {}
> private:
>     std::string a;
> };
> 
> new C("a");
> ```
> 
> we need to compute the size of all members of this struct (e.g. calss: `std::string`), so `ns` is necessary.
> 
> https://github.com/esbmc/esbmc/blob/749b9673062870fdf86c5ee349ac6948f5cc772a/src/util/type_byte_size.cpp#L95-L99

I don't know what the contract is for `namespace.follow` (because there is sadly no documentation), but I expected a call to `follow` to work recursively?
That is, shouldn't https://github.com/esbmc/esbmc/pull/1878/commits/47e04a64c9cc0a7acc79d6cc8d42f29914523071#diff-7d89a774bdea371c90cfc5f34a359018006171b9c29dca5687c254a01b462429R17 have worked?

XLiZHI:
`I don't know what the contract is for namespace.follow (because there is sadly no documentation), but I expected a call to follow to work recursively?`
Maybe this comment helps: https://github.com/esbmc/esbmc/issues/1644#issuecomment-1929201828

If we didn't pass `ns` (ns = false), it wouldn't be executed:
https://github.com/esbmc/esbmc/blob/7b1e7f2d750f8d579db145a118338a9c52b81347/src/util/type_byte_size.cpp#L96

src/util/c_typecast.cpp
```diff
@@ -616,8 +616,8 @@ void c_typecastt::implicit_typecast_followed(
     // to prevent anything that may surprise us.
 
     if (
-      ns.follow(dest_type.subtype()).id() ==
-      src_type.id()) // TODO: remove this condition
+      ns.follow(dest_type.subtype()).id() == src_type.id() &&
+      src_type != dest_type) // TODO: remove this condition
```
fbrausse, line 620: This is weird to me. If `src_type == dest_type`, (a) why would this even be executed, and (b) what would *not* executing it change? I'm confused, sorry.

XLiZHI: An abnormal type expression is generated in our frontend:
```
pointer
 *subtype: pointer
```
If we typecast it, it would look like this: &&expr, this causes the assertion in the `address_of` method to fail.

PR #1879, opened 2024-06-17
[cpp-references] first adjust references, then assign.

intrigus-lgtm:
Makes this TC pass:
```cpp
#include <assert.h>

float dd = 50.0f;
struct f
{
  float &c = dd;
};
int main()
{
  f i;
  i.c = 20.0f;
  assert(i.c == 20.0f);
}
```

Found while adding decomposition declaration support to tuples.


fbrausse (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #1882, opened 2024-06-19
[Solidity] add support for interface

prototypeC14:
This PR adds support for the `interface` and partial support for the `abstract` in Solidity


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -59,6 +59,77 @@ bool solidity_convertert::convert()
   // We need to convert Solidity AST nodes to the equivalent symbols and add them to the context
   nlohmann::json &nodes = src_ast_json["nodes"];
 
+  // Since the interface itself does not create any vulnerabilities
+  // consider removing it and its inheritance information from the contract
```
ChenfengWei0: I agree that the interface would not affect the verification result. However, deleting nodes from the source JSON may not be a good idea. 

May I ask if your approach supports this example?
```solidity
pragma solidity >=0.5.0;

interface caculator {
    enum E {
        x,
        y
    }
}

contract Test {
    function caculator() external pure{
        assert(caculator.E.x <= caculator.E.y);
    }
    function test() external pure{
        assert(caculator.E.x > caculator.E.y);
    }
}
```

prototypeC14: This example is not supported; I will change my approach.

src/solidity-frontend/pattern_check.cpp
```diff
@@ -51,10 +51,18 @@ bool pattern_checker::do_pattern_check()
 bool pattern_checker::start_pattern_based_check(const nlohmann::json &func)
 {
   // SWC-115: Authorization through tx.origin
-  check_authorization_through_tx_origin(func);
```
ChenfengWei0, line 54: Why do we need to delete this?

ChenfengWei0 (commented):
For the functions that do not have a "body", would it be better to assign them with an empty body instead of changing the verification algorithm?

ChenfengWei0 (approved):

fbrausse (commented):
I don't quite understand why an "empty body" is simulated for the cases EventDefinition, "interface" and "abstract". A high-level explanation would go a long way to make this more maintainable. For starters: Am I right in assuming that changing the JSON to "empty body" actually generates a GOTO function with an empty body? And if so, why is a definition of the function actually needed?

ChenfengWei0:
> I don't quite understand why an "empty body" is simulated for the cases EventDefinition, "interface" and "abstract". A high-level explanation would go a long way to make this more maintainable. For starters: Am I right in assuming that changing the JSON to "empty body" actually generates a GOTO function with an empty body? And if so, why is a definition of the function actually needed?

So the idea is to support examples like this:

```solidity
pragma solidity >=0.5.0;

interface Base{
    function getResult() external view ;
}

contract Test {
    function test() public
    {
        Base x;
        x.getResult();
    }
}
```

So the Interface(Contract) type var `x` is calling the function `getResult`, which means it is expected to have a corresponding function symbol in the symbol table. However, the function `getResult` does not have a body, so we manually make an empty body, which is virtually identical to setting the symbol `sol:@C@Base@F@getResult`'s value to an empty code_blockt.

fbrausse:
> So the Interface(Contract) type var `x` is calling the function `getResult`, which means it is expected to have a corresponding function symbol in the symbol table. However, the function `getResult` does not have a body, so we manually make an empty body, which is virtually identical to setting the symbol `sol:@C@Base@F@getResult`'s value to an empty code_blockt.

I see. For C functions that just have a declaration and no body we print something like "no body for function FOO". Why do we actually need to define a body for Solidity?

ChenfengWei0:
> I see. For C functions that just have a declaration and no body we print something like "no body for function FOO". Why do we actually need to define a body for Solidity?

There are two things here:
1. add an empty "body" node to the Json AST: to utilize the existing function-handling APIs, which always expect a body sub-node in the function definition node.
2. add an empty body to the no-body function( i.e. add an empty code_blockt) to the symbol's value: to avoid the no-body function warning ( like "no body for function FOO").

fbrausse:
Yes. So why do we want to avoid the "no body for function FOO" warning when actually there is no body for function FOO? This warning is useful for us: it allows us to claim that whatever is happening there is not under our control. If we define a body (e.g. an empty one) this warning disappears and we cannot claim that "user provided no implementation, therefore we can assume what we usually do when there's no body". My suggestion would be to not record a value in the symbol table for those functions and to actually print the warnings.

ChenfengWei0:
> we cannot claim that "user provided no implementation

Hmm, I thought they should not have any implementation anyway. According to the documentation,
> "Interface can not have any function with implementation."

---
> My suggestion would be to not record a value in the symbol table for those functions and to actually print the warnings

Will do,

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @prototypeC14 and @ChenfengWei0.

PR #1883, opened 2024-06-19
Updates outdated actions used in CI

intrigus-lgtm:
Fixes all the warnings:
grafik

### DIFF
--- a/.github/workflows/benchexec.yml
+++ b/.github/workflows/benchexec.yml
@@ -145,7 +145,7 @@ jobs:
         run: mv $HOME/output.zip ./output.zip      
       - name: Debug Info
         run: ls
-      - uses: actions/upload-artifact@v1
+      - uses: actions/upload-artifact@v4
         with:
           name: esbmc-result
           path: output.zip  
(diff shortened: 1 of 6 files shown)


.github/workflows/benchexec.yml
```diff
@@ -145,13 +145,13 @@ jobs:
         run: mv $HOME/output.zip ./output.zip      
       - name: Debug Info
         run: ls
-      - uses: actions/upload-artifact@v1
+      - uses: actions/upload-artifact@v4
```
rafaelsamenezes, line 148: Can you do a quick benchexec run? Last time that I tried the local runner script was not smart enough to deal with the proxies.

intrigus-lgtm:
Superseded by #1884.

.github/workflows/benchexec.yml
intrigus-lgtm: This is indeed still the case.
But, let's continue this conversation here:
https://github.com/esbmc/esbmc/pull/1884

PR #1884, opened 2024-06-19
Feat/actions update deps versions

intrigus-lgtm:
TODO/RFC:
1. ~~benchexec currently [fails](https://github.com/esbmc/esbmc/actions/runs/9587363003):
Suggested solution is to look at https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github
I don't have sufficient access to the self-hosted runner so don't know what exactly has to change.~~ Solved.
2. PR jobs currently [fail](https://github.com/esbmc/esbmc/actions/runs/9584512859).
This is because we try to upload multiple artifacts with the name "release-ubuntu-latest":
https://github.com/esbmc/esbmc/blob/3a3d934dc6bd737bef602b26f45eccbedcf57e60/.github/workflows/pull_request.yml#L64-L80
We could consider changing [this](https://github.com/esbmc/esbmc/blob/3a3d934dc6bd737bef602b26f45eccbedcf57e60/.github/workflows/build-unix.yml#L35) to also include the `build-flags` in the filename which would work, but also look a bit ugly.
Alternatively, we could add a new workflow input `prefix` that allows us to distinguish between the artifacts.
NOTE:
The way it currently works is that the last job to finish overwrites the result of the previous job.
E.g. [here](https://github.com/esbmc/esbmc/actions/runs/9583234178) the release is 320 MB, but [here](https://github.com/esbmc/esbmc/actions/runs/9572900556) it's 378 MB.
It's actually possible to replicate the old behavior, but I don't think that is what we really want.

This PR supersedes https://github.com/esbmc/esbmc/pull/1883, because I can not try the `benchexec` workflow from a fork

### DIFF
--- a/.github/workflows/benchbringup.yml
+++ b/.github/workflows/benchbringup.yml
@@ -37,7 +37,7 @@ jobs:
           BENCHMARK_TO_RUN: ${{ inputs.benchmark }}
           LOG_DIR: ${{ inputs.logdirectory }}
         run: ./scripts/build.sh -b Release -r On
-      - uses: actions/upload-artifact@v1
+      - uses: actions/upload-artifact@v4
         with:
           name: release-linux
           path: ./build/src/esbmc/esbmc
@@ -47,7 +47,7 @@ jobs:
           ESBMC_OPTS: ${{ inputs.options }}
           LOG_DIR: ${{ inputs.logdirectory }}
         run: cd build/ && ctest -j4 --output-on-failure --progress .
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           name: benchmark-logs
           path: ./build/regression/${{ inputs.logdirectory }}
--- a/.github/workflows/benchbringup.yml
+++ b/.github/workflows/benchbringup.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
       - name: Confirm the benchmark
         run: echo ${{ github.event.inputs.benchmark }}
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Build ESBMC
         env:
           TEST_TIMEOUT: ${{ inputs.timeout }}
--- a/.github/workflows/benchexec.yml
+++ b/.github/workflows/benchexec.yml
@@ -145,7 +145,7 @@ jobs:
         run: mv $HOME/output.zip ./output.zip      
       - name: Debug Info
         run: ls
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v4
         with:
           name: esbmc-result
           path: output.zip  
@@ -169,7 +169,7 @@ jobs:
       - name: Show summary (CPAChecker)
         if: ${{ inputs.cpachecker == true }}
         run: tail $HOME/witness-output/*results*.txt
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v4
         if: ${{ inputs.cpachecker == true }}
         with:
           name: witness-result
--- a/.github/workflows/benchexec.yml
+++ b/.github/workflows/benchexec.yml
@@ -107,7 +107,7 @@ jobs:
       - name: Setup SV-Benchmarks version
         run: cd $HOME/sv-benchmarks && git fetch --all --tags && git checkout ${{ inputs.svbenchmarks }}
       - name: Download Linux Build
-        uses: actions/download-artifact@master
+        uses: actions/download-artifact@v4
         with:
           name: 'release-ubuntu-latest--b Release -e Off -C'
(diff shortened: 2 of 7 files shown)


fbrausse:
> 1. benchexec currently [fails](https://github.com/esbmc/esbmc/actions/runs/9587363003):

I believe the problem occurs because our self-hosted runners are behind an authenticated SOCKS proxy, which for unclear reasons the old version of the Github action supports while the new one doesn't. @rafaelsamenezes might know more about the setup.

It might be related to this issue of theirs: <https://github.com/actions/download-artifact/issues/200>. The workaround of exporting `NODE_TLS_REJECT_UNAUTHORIZED=0` suggested there doesn't look promising, but could point to some issue with Node.

rafaelsamenezes:
> > 1. benchexec currently [fails](https://github.com/esbmc/esbmc/actions/runs/9587363003):
> 
> I believe the problem occurs because our self-hosted runners are behind an authenticated SOCKS proxy, which for unclear reasons the old version of the Github action supports while the new one doesn't. @rafaelsamenezes might know more about the setup.
> 
> It might be related to this issue of theirs: [actions/download-artifact#200](https://github.com/actions/download-artifact/issues/200). The workaround of exporting `NODE_TLS_REJECT_UNAUTHORIZED=0` suggested there doesn't look promising, but could point to some issue with Node.

Hm, I will take a look later today on how to fix this Node issue. Hopefully is just a matter of adding some certificates.

intrigus-lgtm:
> > 1. benchexec currently [fails](https://github.com/esbmc/esbmc/actions/runs/9587363003):
> 
> I believe the problem occurs because our self-hosted runners are behind an authenticated SOCKS proxy, which for unclear reasons the old version of the Github action supports while the new one doesn't. @rafaelsamenezes might know more about the setup.

The old upload action probably used a different endpoint/domain.

https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes:
> On self hosted runners, additional [firewall rules](https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes) may be required.

fbrausse:
Very strange, https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-a-proxy-server-with-self-hosted-runners does not mention anything about SOCKS proxies. I have access to our proxy machine and could potentially install a different implementation, but authenticated SOCKS seems pretty good for our use case. Does anyone know which HTTP proxy server is any good?

intrigus-lgtm:
> Very strange, https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-a-proxy-server-with-self-hosted-runners does not mention anything about SOCKS proxies. I have access to our proxy machine and could potentially install a different implementation, but authenticated SOCKS seems pretty good for our use case. Does anyone know which HTTP proxy server is any good?

Does the socks proxy limit connections to specific domains?

fbrausse:
> Does the socks proxy limit connections to specific domains?

No, it doesn't. It's [Dante](https://www.inet.no/dante/) with no DNS-specific things in the config.

rafaelsamenezes:
I have replaced the socks proxy. We should be able to upgrade the actions now.

intrigus-lgtm:
@rafaelsamenezes thank you.

So now the only thing remaining is my [question number 2](https://github.com/esbmc/esbmc/pull/1884#issue-2363208964):

rafaelsamenezes (approved):

intrigus-lgtm:
setup-python@v2 did not work correctly on Windows.
Despite claiming to have setup 3.8, cmake was still somehow able to find 3.12 instead.
With setup-python@v5 this changed and now cmake "correctly" found 3.8.
This however causes problems, because we actually need 3.9.

We still need a solution for [bullet point two](https://github.com/esbmc/esbmc/pull/1884#issue-2363208964).

rafaelsamenezes:
> setup-python@v2 did not work correctly on Windows. Despite claiming to have setup 3.8, cmake was still somehow able to find 3.12 instead. With setup-python@v5 this changed and now cmake "correctly" found 3.8. This however causes problems, because we actually need 3.9.

Do we have an open issue for this? Our scripts _should_ work on the latest python version

> 
> We still need a solution for [bullet point two](https://github.com/esbmc/esbmc/pull/1884#issue-2363208964).

According to https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact we might be able to just add a prefix/postfix into the artifact name.

lucasccordeiro (approved):

rafaelsamenezes:
With https://github.com/esbmc/esbmc/pull/2023 merged, I guess item 2 is also solved.

PR #1885, opened 2024-06-20
CFG abstraction for GOTO programs

rafaelsamenezes:
This PR constructs a class to be used as an abstraction of the goto program as a CFG. This can be used for us to apply more classic compiler algorithms directly.

As an example, the following program

```c
int main()
{
  int n = 10;

  loop:
     n--;
     if (n) goto loop;
  end:
  return 0;
}
```

Produces the CFG:

graphviz(1)


src/goto-programs/goto_cfg.h
```diff
@@ -0,0 +1,40 @@
...
+     * @brief Generates a dot file containing the CFG.
+     *
+     * @param filename output file name
+     */
+    void dump_graph() const;
+
+     /**
+     * @brief A basic block is a sequence of instructions that has no branches in it.
+     */
+    struct basic_block
```
fbrausse: Let's add some more information here:
- Last instruction is a (possibly conditional) branch instruction (right?)
- Each basic block contains at least one instruction 

Also, I'm unclear about `successors`: shouldn't this be an ordered list? How is the `true` branch differentiated from the `false` one? Do we handle `switch` statements specially in GOTO and/or the CFG?

rafaelsamenezes: > Last instruction is a (possibly conditional) branch instruction (right?

How come? Last Instruction is supposed to be end function.

> Also, I'm unclear about successors: shouldn't this be an ordered list? How is the true branch differentiated from the false one? Do we handle switch statements specially in GOTO and/or the CFG?

It does not matter for a CFG. It is not supposed to be aware of the branch conditions, algorithms will compute the precondition based on the predecessors.

fbrausse: Well, I was assuming the BBs follow a pattern similar to https://clang.llvm.org/doxygen/classclang_1_1CFGBlock.html#details:

> Represents a single basic block in a source-level [CFG](https://clang.llvm.org/doxygen/classclang_1_1CFG.html).
>
> It consists of:
>
> (1) A set of statements/expressions (which may contain subexpressions). (2) A "terminator" statement (not in the set of statements). (3) A list of successors and predecessors.
>
> Terminator: The terminator represents the type of control-flow that occurs at the end of the basic block. The terminator is a Stmt* referring to an AST node that has control-flow: if-statements, breaks, loops, etc. If the control-flow is conditional, the condition expression will appear within the set of statements in the block (usually the last statement).
>
> Predecessors: the order in the set of predecessors is arbitrary.
>
> Successors: the order in the set of successors is NOT arbitrary. We currently have the following orderings based on the terminator:
> [...]

rafaelsamenezes: Hm, that looks interesting and easy enough to do.

lucasccordeiro (approved):

src/goto-programs/goto_cfg.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+#include <fstream>
+
+void goto_cfg::dump_graph() const
+{
+  log_status("Dumping CFG");
+  for (const auto &[function, bbs] : basic_blocks)
+  {
+    std::ofstream file("cfg_" + function + ".dot");
```
fbrausse, line 119: Should we make the pattern (e.g. `cfg_%s.dot` here) part of the `--dump-goto-cfg` option as a parameter?

rafaelsamenezes: You mean giving an optional pattern?

fbrausse: Yes, giving a pattern (if we really need to create a new file for each function), but I would make it a mandatory parameter. An alternative to having a separate file for each function might be to use `subgraph`, but the resulting rendering could get quite large.

rafaelsamenezes: > Yes, giving a pattern (if we really need to create a new file for each function), but I would make it a mandatory parameter

Sure, I don't expect nobody to use this beside us though.

>  An alternative to having a separate file for each function might be to use subgraph, but the resulting rendering could get quite large.

Yeah, I used the same approach than LLVM to export this.

fbrausse (approved):

Issue #1769, opened 2024-04-04
[Solidity] Support for mapping keyword


HOLLYwyh:
"mapping" is an important data structure in Solidity, and I sincerely hope that in the upcoming versions, ESBMC can support verification related to this data structure.

lucasccordeiro:
Thanks for raising this issue, @HOLLYwyh.

@ChenfengWei0 is currently working on implementing this feature in ESBMC.

By the way, we're looking for relevant Solidity projects in the community. Please let us know if you want to recommend some open-source Solidity projects. We want to apply ESBMC to verify them formally.

lucasccordeiro:
Closing as a duplicate of #1707, which tracks `mapping` support in the Solidity frontend and already carries the specification references and current status (partial support landed, per @ChenfengWei0). Please follow #1707 for progress. Related splits: #1708 (`event`) and #1709 (`balanceof`).

PR #1886, opened 2024-06-20
[Solidity] Remove solidity_template and move to c2goto

ChenfengWei0:
Changed as suggested in #1769


src/c2goto/headers/CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ set(inputs
            ubuntu20.04/kernel_5.15.0-76/include/linux/spinlock.h
            ubuntu20.04/kernel_5.15.0-76/include/linux/slab.h
            ubuntu20.04/kernel_5.15.0-76/include/asm/uaccess.h
+
+           # Solidity
+           solidity.h
```
fbrausse: I would suggest to move this file to the `__esbmc` subdirectory, if possible.

PR #1892, opened 2024-06-21
Removed BUG mark from FindLocalLLVM.cmake

rafaelsamenezes:
We can use the *nix builds on Ubuntu and macOS (homebrew) just fine.


fbrausse (approved):
Shared linking works fine indeed, but static doesn't, at least on Ubuntu because they messed up their CMake files: the Clang libs all depend on "LLVM", but Ubuntu only provides the shared libLLVM.so.

lucasccordeiro (approved):

PR #1908, opened 2024-06-26
[build] make buildidobj cope with missing .git repository & cleanup

intrigus-lgtm:
I've started use https://www.jetbrains.com/help/clion/remote-projects-support.html for running the tests remotely on a server.
However, clion by default does not copy the `.git` folder. So while `git` is installed, it still fails.

I changed the script to just return the information that we would have returned if `git` were not available.


rafaelsamenezes (approved):
Nice, LGTM. My only nitpick is that I would have preferred an optional type instead of a union.

lucasccordeiro (approved):

Issue #1902, opened 2024-06-25
[multi-property] Our incremental verification approach keeps incrementing k for a loop-free program


lucasccordeiro:
Our incremental verification approach keeps incrementing `k` for a loop-free program when using multi-property verification:

$ esbmc file.c --k-induction --overflow --memory-leak-check --unlimited-k-step --multi-property

````
//FormAI DATASET v1.0 Category: Time Travel Simulator ; Style: complex
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

int main() {
  time_t current_time;
  struct tm *current_tm;
  int time_travel_year, time_travel_month, time_travel_day;
  time_t time_travel_seconds;
  struct tm *time_travel_tm;
  int current_year, time_travelled_years, age_current, age_time_travelled;

  // Get current time
  current_time = time(NULL);
  current_tm = localtime(&current_time);

  // Print current time
  printf("Current Time: %02d/%02d/%d\n", current_tm->tm_mday, current_tm->tm_mon + 1, current_tm->tm_year + 1900);

  // Get user inputs for time travel date
  printf("Enter year: ");
  scanf("%d", &time_travel_year);
  printf("Enter month: ");
  scanf("%d", &time_travel_month);
  printf("Enter day: ");
  scanf("%d", &time_travel_day);

  // Convert time travel date to seconds
  time_travel_tm = localtime(&current_time);
  time_travel_tm->tm_year = time_travel_year - 1900;
  time_travel_tm->tm_mon = time_travel_month - 1;
  time_travel_tm->tm_mday = time_travel_day;
  time_travel_seconds = mktime(time_travel_tm);

  // Determine current year and years travelled
  current_year = current_tm->tm_year + 1900;
  time_travelled_years = time_travel_year - current_year;

  // Print time travel date and years travelled
  printf("Time Travel Date: %02d/%02d/%d\n", time_travel_day, time_travel_month, time_travel_year);
  printf("Years Travelled: %d\n", time_travelled_years);

  // Determine current age and age at time travel date
  age_current = current_year - 1990;
  age_time_travelled = age_current + time_travelled_years;

  // Print age at current time and at time travel date
  printf("Age at Current Time: %d\n", age_current);
  printf("Age at Time Travel Date: %d\n", age_time_travelled);

  return 0;
}
````

PR #1910, opened 2024-06-27
[multi-property] Partially bug fixed for multi-property with kind and incr

ChenfengWei0:
Related issue: #1902

So the issue is caused by that, the claims (`dereference failure`) being added during the symex process, which cannot be removed permanently from the goto_functions. Thus, every time we set up a BMC process and run base-case, we will repeatedly output the verified claims, as shown in #1902.

IMO, a potential solution is to add a flag to disable the assertion instrumentation for verified claims during symex, e.g. `--no-pointer-check:instruction_location`. 

This PR, however, does not do so (because I am not sure if it's the right way to do so). So instead I have done a partial fixing, where ESBMC will still keep running until `verification unknown`, but we will at least not report duplicated CEs.


fbrausse:
@ChenfengWei0 I don't think these regression tests should actually return `VERIFICATION UNKNOWN`. Better to mark them `KNOWNBUG` and enter the correct expected result.

ChenfengWei0:
> @ChenfengWei0 I don't think these regression tests should actually return `VERIFICATION UNKNOWN`. Better to mark them `KNOWNBUG` and enter the correct expected result.

Just tried `KNOWNBUG` together with SUCCESSFUL/FAILED/UNKNOWN and none of these could pass the ctest. So I changed it to `FUTURE` instead.

src/esbmc/bmc.h
```diff
@@ -38,6 +43,9 @@ class bmct
 protected:
   const contextt &context;
   namespacet ns;
+  // for multi-kind/incr
+  std::set<std::pair<std::string, std::string>> &to_remove_claims;
+  std::set<std::pair<std::string, std::string>> dump = {};
```
fbrausse: I don't understand what purpose these member serve exactly. Could you please explain in a comment what they contain and where it's used?

Also, since this `std::pair<std::string, std::string>` is used with a particular meaning at multiple locations (first: claim_msg, second: claim_loc), I think it makes sense to introduce a new type that has these two strings as members, in order to differentiate those objects from arbitrary pairs of strings (such as those held by, for instance, a `std::map<std::string, std::string>`).

src/esbmc/esbmc_parseoptions.h
```diff
@@ -103,6 +103,9 @@ class esbmc_parseoptionst : public parseoptions_baset, public language_uit
 
 public:
   goto_functionst goto_functions;
+
+  // for multi-kind/incr
+  std::set<std::pair<std::string, std::string>> to_remove_claims;
```
fbrausse: Same as above: what information is being accumulated here, and does it need to be stored in the `esbmc_parseoptionst` class instance?

src/esbmc/options.cpp
```diff
@@ -200,7 +200,8 @@ const struct group_opt_templ all_cmd_options[] = {
     {"no-slice", NULL, "do not remove unused equations"},
     {"multi-fail-fast",
      boost::program_options::value<int>()->value_name("n"),
-     "stops after first n VCC violation found in multi property mode"},
+     "stops after first n VCC violation found in multi property mode."
+     "(not compatible with incremental bmc strategies)"},
```
fbrausse: What does the plural form "incremental bmc strategies" refer to? Just `--incremental-bmc` or also k-induction or anything else? Please clarify.

rafaelsamenezes: You should move this from the option and check it at the code level. If it is a hard requirement you should crash or report with an warning if its just buggy.

src/esbmc/esbmc_parseoptions.h
rafaelsamenezes: Shouldn't we use unordered here?

src/esbmc/bmc.h
rafaelsamenezes: What is this dump var?

src/esbmc/bmc.cpp
```diff
@@ -917,38 +942,61 @@ smt_convt::resultt bmct::multi_property_check(
...
-          for (auto &instruction : it.second.body.instructions)
+          if (it->is_assert())
           {
-            if (
-              instruction.is_assert() &&
-              from_expr(ns, "", instruction.guard) == claim.claim_msg &&
-              instruction.location.as_string() == claim.claim_loc)
+            claim_msg = from_expr(ns, "", it->guard);
+            claim_loc = it->location.as_string();
+            claim_pair = std::make_pair(claim_msg, claim_loc);
```
rafaelsamenezes: Instead of a pair coldn't you just concatenate the strings? That should give a better interface (std::string) instead of (pair<string,string>)

rafaelsamenezes (changes requested):

fbrausse (changes requested):
I'm trying to understand why we need to remove claims that already have been generated instead of avoiding to generate them in the first place. Also the way they're removed seems a bit too intrusive for my understanding: Are these `std::sets` and references thereto really required to stay alive during the entire lifetime of objects of the classes `bmct` and `esbmc_parseoptionst`?

lucasccordeiro:
@ChenfengWei0: can I ask you whether you had a chance to check @rafaelsamenezes and @fbrausse comments?

src/esbmc/options.cpp
ChenfengWei0: Will do

src/esbmc/bmc.cpp
ChenfengWei0: Okay I will change to string concatenate

src/esbmc/bmc.h
ChenfengWei0: Will drop it

src/esbmc/esbmc_parseoptions.h
ChenfengWei0: Will change

src/esbmc/options.cpp
ChenfengWei0: It Should at least include `incremental-bmc`/`k-induction`/`k-induction-parallel`. 

Will change the description

ChenfengWei0:
> I'm trying to understand why we need to remove claims that already have been generated instead of avoiding to generate them in the first place. Also the way they're removed seems a bit too intrusive for my understanding: Are these `std::sets` and references thereto really required to stay alive during the entire lifetime of objects of the classes `bmct` and `esbmc_parseoptionst`?

I think we are more like converting the verified claims' guard to True. So if there is a claim that has been verified to be `P_SATISFIABLE`, we convert it to a `SKIP` via make_skip, which will set the guard to True.

I think we could also avoid the generation of verified claims. For instance, we could label the verified claims in Goto with instruction property and identify them during the generation. If you are not happy with the current solution I can do this approach.

The idea of `std::sets` is that,  every round of the `base-case`, `forward-condition`, and `inductive-step` will set up a `bmct` object and run BMC individually. So we need to share the information among these BMCs. This information is used to skip the claims that are generated during the SymEx and not inside the Goto (which we cannot label directly).

lucasccordeiro:
@ChenfengWei0: Can I ask you about the status of this PR?

ChenfengWei0:
> @ChenfengWei0: Can I ask you about the status of this PR?

I am currently busy with bug fixing on the solidity frontend. I will get back to this PR once I'm done.

lucasccordeiro:
Thanks for this update, @ChenfengWei0.

PR #1914, opened 2024-06-28
[python-frontend] Add support for mod.cls.method calls

### LINKED ISSUE #1901: [python-frontend] Import the whole module
ESBMC should handle imports like this:

```python
import my_module

n:int = my_module.my_class.my_classmethod()

assert n == 0
```
from a module with a class definition:
```python
class my_class:
    @classmethod
    def my_classmethod(cls) -> int:
        return 0
```

ivomachado:
To allow chained call when there is an "import module" and a call to a class method with "module.class.classmethod()".

### DIFF
--- a/regression/python/import-whole-module/main.py
+++ b/regression/python/import-whole-module/main.py
@@ -1,3 +1,0 @@
-import other
-
-assert other.OtherClass.foo() == 3
--- a/regression/python/import-whole-module/main.py
+++ b/regression/python/import-whole-module/main.py
@@ -1,0 +1,3 @@
+import other
+
+assert other.OtherClass.foo() == 3
--- a/src/python-frontend/json_utils.h
+++ b/src/python-frontend/json_utils.h
@@ -8,8 +8,11 @@ namespace json_utils
 template <typename JsonType>
 JsonType find_class(const JsonType &ast_json, const std::string &class_name)
 {
-  auto it =
-    std::find_if(ast_json.begin(), ast_json.end(), [&](const JsonType &obj) {
+  auto it = std::find_if(
+    ast_json.begin(),
+    ast_json.end(),
+    [&](const JsonType &obj)
+    {
       return obj.contains("_type") && obj["_type"] == "ClassDef" &&
              obj["name"] == class_name;
     });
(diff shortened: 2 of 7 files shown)


src/python-frontend/json_utils.h
```diff
@@ -44,6 +44,25 @@ bool is_class(const std::string &name, const JsonType &ast_json)
...
+        std::ifstream imported_file(module_path.str());
+        if (!imported_file.is_open())
+          return false;
+
+        JsonType imported_module_json;
+        imported_file >> imported_module_json;
+
+        if (is_class(name, imported_module_json))
+          return true;
+      }
```
brcfarias: Please define a lambda for this, and call it in both blocks.

src/python-frontend/python_converter.cpp
brcfarias, line 1: Run `clang-format-11 -i src/python-frontend/python_converter.cpp` to fix the indentation.

src/python-frontend/python_converter.cpp
```diff
@@ -545,7 +545,14 @@ function_id python_converter::build_function_id(const nlohmann::json &element)
   {
     is_member_function_call = true;
     func_name = func_json["attr"];
-    obj_name = func_json["value"]["id"];
+    if (func_json["value"]["_type"] == "Attribute")
+    {
```
brcfarias: We try to avoid using curly braces for single-line if statements. &#128578;

brcfarias (changes requested):

regression/python/import-whole-module/main.py
brcfarias, line 1: `regression/python/import/main.py` contains an import in the same format (whole module).
The diference is that a class method is being invoked here. Could you please merge them by moving the class definition and class method call there?

brcfarias (changes requested):

brcfarias:
Fixes #1901.

src/python-frontend/json_utils.h
ivomachado: Done

src/python-frontend/python_converter.cpp
ivomachado: Done

src/python-frontend/python_converter.cpp
ivomachado: Done

regression/python/import-whole-module/main.py
ivomachado: Done

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @ivomachado.

PR #1918, opened 2024-06-28
[Solidity] add support for Contract type variables

prototypeC14:
Previously, we do not support Contract type due to that we do not support struct type. Since we already support struct now, we should be able to parse Contract type variables.


ChenfengWei0:
@prototypeC14 , You may want to add a `test.desc` file.

prototypeC14:
I'll add it after feature support

PR #1919, opened 2024-06-29
[python-frontend] String comparison

brcfarias:
This PR involves loading our C models, and using **strncmp** to compare strings in Python.


brcfarias:
Oops, I need to adjust the filtering logic during the models loading. Working on it.

brcfarias:
I refactored `goto-programs/read_goto_binary.h` into a `goto_binary_reader` class, and added an attribute that allows filtering the functions to be loaded. When checking a Python file, the list of functions is provided.

lucasccordeiro (approved):

lucasccordeiro:

src/util/type.cpp
```diff
@@ -19,6 +19,11 @@ bool is_array_like(const typet &type)
   return type.is_vector() || type.is_array() || type.is_incomplete_array();
 }
 
+bool is_string(const typet &type)
+{
+  return type.is_array() && type.subtype().is_signedbv();
+}
```
fbrausse, line 25: This is weird. An array of signed bit-vectors (of any size) is certainly not a "string". What is a string? We have string literals, which have an associated (or inferred) encoding. But an array of ints by itself certainly is not a string. I recommend moving this function out of the general util/type support library.

src/c2goto/cprover_library.cpp
```diff
@@ -86,6 +87,9 @@ static const struct buffer
   },
 #endif
 };
+
+static std::vector<std::string> python_c_models = {"strncmp"};
```
fbrausse, line 91: Can this be made `const`?

src/c2goto/cprover_library.cpp
```diff
@@ -192,13 +196,19 @@ void add_cprover_library(contextt &context, const languaget *c_language)
...
+      return add_bundled_library_sources(context, *language);
     log_error("Zero-lengthed internal C library");
     abort();
   }
 
-  if (read_goto_binary_array(clib->start, clib->size, new_ctx, goto_functions))
+  goto_binary_reader goto_reader;
+
+  if (language && language->id() == "python")
+    goto_reader.set_functions_to_read(python_c_models);
```
fbrausse, line 208: I think instead of hard-coding the list of functions here, it might be better to extend the `languaget` class to provide a function that prepares the `goto_reader` according to its own needs.

fbrausse (commented):
I have a couple of comments.

src/util/type.cpp
brcfarias: > But an array of ints by itself certainly is not a string

It was based on this:

image

image

fbrausse: #1547 removed string types, so the string_constantt expression is just a special way to produce a constant array now. The reason for removing the string type was to simplify the code. It has not been used anywhere, but always been converted to the underlying array anyway.

PR #1928, opened 2024-07-04
[Solidity] Add support for import path

prototypeC14:
Added support for import path, allowing multiple .sol files to be imported during verification.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -44,6 +44,9 @@ bool solidity_convertert::convert()
   //  1. First, we perform pattern-based verificaiton
   //  2. Then we populate the context with symbols annotated based on the each AST node, and hence prepare for the GOTO conversion.
 
+  // First, we handle the miltipule JSON files(imported files)
```
fbrausse: Typo: "multiple"

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -199,6 +202,42 @@ bool solidity_convertert::convert()
   return false; // 'false' indicates successful completion.
 }
 
+void solidity_convertert::multi_Json_file(nlohmann::json &src_ast_json)
```
fbrausse: Let's stick to lower-case symbol names delimited by underscores.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -199,6 +202,42 @@ bool solidity_convertert::convert()
...
+{
+  // This function is used to handle multiple JSON files into a single JSON file
+  src_ast_json = src_ast_json_array[0];
+
+  // Find the position to insert after "PragmaDirective"
+  size_t insert_pos = 0;
```
fbrausse, line 244: Am I right in assuming that there ever is at most one PragmaDirective? What does it do?

fbrausse (changes requested):
I'd like to request a couple of minor changes and have a question.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: 1.The result of 'import'is actually multiple original AST structures written into a single AST file.
image
2. But our parsing rules only contend for the structure of a single AST so we need to consolidate multiple ASTs into a single one, i.e. consolidate the import (contract_import.sol).
3. The beginning of the nodes in a single AST consists of an import information description section, a version description section. This is followed by all the information that needs to be verified, i.e. contract, interface, and library nodes, as shown in the following figure.
```json
    "nodes": [
        {
            "absolutePath": "contract_import.sol",
            "file": "./contract_import.sol",
            "id": 1,
            "nameLocation": "-1:-1:-1",
            "nodeType": "ImportDirective",
            "scope": 17,
            "sourceUnit": 37,
            "src": "36:31:0",
            "symbolAliases": [],
            "unitAlias": ""
        },
        {
            "id": 2,
            "literals": [
                "solidity",
                ">=",
                "0.8",
                ".0"
            ],
            "nodeType": "PragmaDirective",
            "src": "68:24:0"
        },
        {
            "abstract": false,
            "baseContracts": [],
            "canonicalName": "Derive",
            "contractDependencies": [
                36
            ],
            "contractKind": "contract",
            "fullyImplemented": true,
            "id": 16,
            "linearizedBaseContracts": [
                16
            ],
            "name": "Derive",
            "nameLocation": "104:6:0",
            "nodeType": "ContractDefinition",
            "nodes": 
]
```
4.So the rest of the key nodes need to be inserted sequentially thereafter("nodeType": "PragmaDirective",Yes, there is only one in each structure).The reason I chose to insert it here instead of at the end is that by splitting a piece of solidity code into multiple files, I found that the (contract,interface...) in the AST file The import order of the (contract, interface...) in the AST file is backwards compared to the unsplit, so it is necessary to insert it from the head node to avoid parsing failures such as inheritance.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: I'll correct it immediately.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: I'll correct it immediately.

src/solidity-frontend/solidity_convert.cpp
fbrausse: Many thanks for the detailed explanation! It's really helpful to understand what's going on. Could I ask you to maybe put some of this information into comments in the code? For instance, it would be good to know that PragmaDirective occurs exactly once per structure, and why you chose to perform the insertion in that particular order (i.e., your item ~3.~ 4.). Many thanks!

prototypeC14: Sure, I can add some this information as comments in the code.

fbrausse (approved):
LGTM!

src/solidity-frontend/solidity_convert.h
```diff
@@ -209,7 +210,9 @@ class solidity_convertert
   contextt &context;
   namespacet ns;
   // json for Solidity AST. Use vector for multiple contracts
-  nlohmann::json &src_ast_json;
+  nlohmann::json src_ast_json_array = nlohmann::json::array();
```
ChenfengWei0, line 216: Do we need this default value? I thought it will be initialized in the constructor initialize list.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -199,6 +202,49 @@ bool solidity_convertert::convert()
   return false; // 'false' indicates successful completion.
 }
 
+void solidity_convertert::multi_json_file()
+{
+  // This function is used to handle multiple JSON files into a single JSON file
+  src_ast_json = src_ast_json_array[0];
```
ChenfengWei0, line 238: What does the `src_ast_json_array[0]` stand for? Could you please add some documentations here?

ChenfengWei0 (approved):
Overall looks good to me. Just some questions:
- This PR seemingly focuses on the import path only. I am wondering if it also supports [grammar](https://docs.soliditylang.org/en/latest/layout-of-source-files.html#importing-other-source-files) like:
```solidity
import * as symbolName from "filename";
import "filename" as symbolName;
import {symbol1 as alias, symbol2} from "filename";
```
Which you might need to read from the `ImportDirective` node from ast_json. Otherwise you might want to change the title to "[Solidity] Add support for import path"

- Also, I am wondering if this PR support cases like
```solidity
contract Base{}
// import afterwards
import "./contract2.sol";
```
And 
```solidity
// example.sol
contract Base{}

// self-import
import "./example.sol";
```
And
```solidity
contract Base{}

// duplicate 
import "./example.sol";
import "./example.sol";
```
It would be better to add these to the regression tests

src/solidity-frontend/solidity_convert.cpp
prototypeC14: Yes, src_ast_json_array[0] means the .sol file that is being verified and not being imported.

src/solidity-frontend/solidity_convert.h
prototypeC14: I've used the same display construct here as in language to distinguish it from the single AST structure below, which I think makes the code easier to read.

prototypeC14:
> Overall looks good to me. Just some questions:
> 
> * This PR seemingly focuses on the import path only. I am wondering if it also supports [grammar](https://docs.soliditylang.org/en/latest/layout-of-source-files.html#importing-other-source-files) like:
> 
> ```solidity
> import * as symbolName from "filename";
> import "filename" as symbolName;
> import {symbol1 as alias, symbol2} from "filename";
> ```
> 
> Which you might need to read from the `ImportDirective` node from ast_json. Otherwise you might want to change the title to "[Solidity] Add support for import path"
> 
> * Also, I am wondering if this PR support cases like
> 
> ```solidity
> contract Base{}
> // import afterwards
> import "./contract2.sol";
> ```
> 
> And
> 
> ```solidity
> // example.sol
> contract Base{}
> 
> // self-import
> import "./example.sol";
> ```
> 
> And
> 
> ```solidity
> contract Base{}
> 
> // duplicate 
> import "./example.sol";
> import "./example.sol";
> ```
> 
> It would be better to add these to the regression tests

Sorry for taking so long to reply, I've been making corrections to the content and now make the following note:
1. Yes, this pr focus on for the import path only. For import * as symbolName from "filename"; import "filename" as symbolName; These two are mainly used for library, which is transformed into contractkind:library even when contract is imported, but it seems that we don't support the library keyword at the moment so we don't consider this aspect.
2. For import {symbol1 as alias, symbol2} from "filename"; Because the parsing for 'nickname' was not implemented, it was not read in. However, the ImportDirective node is indeed necessary. Therefore, I added an algorithm to sort the import relationships and order, making it more rigorous. I believe this will be beneficial for future development.
3. The last three cases are supported and I'll add them to the regression tests.

lucasccordeiro (approved):

lucasccordeiro:
Thank you for submitting this PR, @prototypeC14.

PR #1931, opened 2024-07-10
[c++ verification] Fix the string type in Conditional Operator

### LINKED ISSUE #1930: [C++ Verification] ERROR: compute_pointer_offset, unexpected irep:
ESBMC crashes for this C++ code using the following command:

$ esbmc pol_assert.cpp --incremental-bmc

````CPP
#include <iostream>
#include <sstream>
#include <string>
#include <cassert>
using namespace std;

// Base class
class Animal {
public:
    // Virtual function
    virtual void makeSound() const {
        cout << "Some generic animal sound" << endl;
    }

    // Virtual destructor
    virtual ~Animal() {}
};

// Derived class: Dog
class Dog : public Animal {
public:
    void makeSound() const override {
        cout << "Woof!" << endl;
    }
};

// Derived class: Cat
class Cat : public Animal {
public:
    void makeSound() const override {
        cout << "Meow!" << endl;
    }
};

// Function to check polymorphic behavior using assertions
void checkPolymorphism(Animal* animalPtr, const string& expectedSound) {
    // Redirect cout to a stringstream to capture the output
    stringstream buffer;
    streambuf* old = cout.rdbuf(buffer.rdbuf());

    // Call makeSound and capture the output
    animalPtr->makeSound();
    string actualSound = buffer.str();

    // Restore original cout buffer
    cout.rdbuf(old);

    // Assert that the actual sound matches the expected sound
    assert(actualSound == expectedSound && "Polymorphism check failed");
}

int main() {
    // Creating instances of derived classes
    Dog dog;
    Cat cat;

    // Pointer to base class
    Animal* animalPtr;

    // Pointing to dog
    animalPtr = &dog;
    checkPolymorphism(animalPtr, "Woof!\n");  // Check polymorphism

    // Pointing to cat
    animalPtr = &cat;
    checkPolymorphism(animalPtr, "Meow!\n");  // Check polymorphism

    // Array of base class pointers
    Animal* animals[2];
    animals[0] = &dog;
    animals[1] = &cat;

    // Polymorphic behavior with assertions
    for (int i = 0; i < 2; ++i) {
        string expectedSound = (i == 0) ? "Woof!\n" : "Meow!\n";
        checkPolymorphism(animals[i], expectedSound);
    }

    return 0;
}
````

````
ERROR: compute_pointer_offset, unexpected irep:
if
* cond : equality
    * side_1 : symbol
        * name : c:pol_assert.cpp@1693@F@main#@i
        * renamelev : Level 1
        * level1_num : 1
        * level2_num : 0
        * thread_num : 0
        * node_num : 0
        * type : signedbv
          * width : 32
    * side_2 : constant_int
        * value : 0
        * type : signedbv
          * width : 32
    * type : bool
* true_value : constant_array
    * members :
      0: constant_int
        * value : 87
        * type : signedbv
          * width : 8
      1: constant_int
        * value : 111
        * type : signedbv
          * width : 8
      2: constant_int
        * value : 111
        * type : signedbv
          * width : 8
      3: constant_int
        * value : 102
        * type : signedbv
          * width : 8
      4: constant_int
        * value : 33
        * type : signedbv
          * width : 8
      5: constant_int
        * value : 10
        * type : signedbv
          * width : 8
      6: constant_int
        * value : 0
        * type : signedbv
          * width : 8

    * type : array
      * subtype : signedbv
          * width : 8
      * array_size : constant_int
          * value : 7
          * type : unsignedbv
            * width : 64
      * size_is_infinite : false
* false_value : constant_array
    * members :
      0: constant_int
        * value : 77
        * type : signedbv
          * width : 8
      1: constant_int
        * value : 101
        * type : signedbv
          * width : 8
      2: constant_int
        * value : 111
        * type : signedbv
          * width : 8
      3: constant_int
        * value : 119
        * type : signedbv
          * width : 8
      4: constant_int
        * value : 33
        * type : signedbv
          * width : 8
      5: constant_int
        * value : 10
        * type : signedbv
          * width : 8
      6: constant_int
        * value : 0
        * type : signedbv
          * width : 8

    * type : array
      * subtype : signedbv
          * width : 8
      * array_size : constant_int
          * value : 7
          * type : unsignedbv
            * width : 64
      * size_is_infinite : false
* type : array
  * subtype : signedbv
      * width : 8
  * array_size : constant_int
      * value : 7
      * type : unsignedbv
        * width : 64
  * size_i
````

### PR DESCRIPTION
**Pull request by XLiZHI**
When the `true` and `false` expressions in the conditional operator are `string` type, we need to cast them to pointer types.

This happens in AST when both StringLiterals are of the same size:
`string expectedSound = (i == 0) ? "Woof!" : "Meow!";`
"Woof!": char[6]
"Meow!": char[6]
<details>

```cpp
    | `-CompoundStmt 0x6dcef48 <line:9:3, line:11:3>
    |   `-DeclStmt 0x6dcef30 <line:10:5, col:56>
    |     `-VarDecl 0x6dcebd8 <col:5, col:49> col:12 expectedSound 'std::string' cinit
    |       `-ExprWithCleanups 0x6dcef18 <col:12, col:49> 'std::string'
    |         `-CXXConstructExpr 0x6dceee8 <col:12, col:49> 'std::string' 'void (const std::string &) noexcept' elidable
    |           `-MaterializeTemporaryExpr 0x6dceed0 <col:28, col:49> 'const std::string' lvalue
    |             `-ImplicitCastExpr 0x6dceeb8 <col:28, col:49> 'const std::string' <NoOp>
    |               `-ImplicitCastExpr 0x6dcedb0 <col:28, col:49> 'std::string' <ConstructorConversion>
    |                 `-CXXConstructExpr 0x6dced80 <col:28, col:49> 'std::string' 'void (const char *)'
    |                   `-ImplicitCastExpr 0x6dced48 <col:28, col:49> 'const char *' <ArrayToPointerDecay>
    |                     `-ConditionalOperator 0x6dced18 <col:28, col:49> 'const char [6]' lvalue
    |                       |-ParenExpr 0x6dcecb8 <col:28, col:35> 'bool'
    |                       | `-BinaryOperator 0x6dcec98 <col:29, col:34> 'bool' '=='
    |                       |   |-ImplicitCastExpr 0x6dcec80 <col:29> 'int' <LValueToRValue>
    |                       |   | `-DeclRefExpr 0x6dcec40 <col:29> 'int' lvalue Var 0x6dcea78 'i' 'int'
    |                       |   `-IntegerLiteral 0x6dcec60 <col:34> 'int' 0
    |                       |-StringLiteral 0x6dcecd8 <col:39> 'const char [6]' lvalue "Woof!"
    |                       `-StringLiteral 0x6dcecf8 <col:49> 'const char [6]' lvalue "Meow!"
```
</details>

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/github_1930/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_1930/main.cpp
@@ -1,0 +1,18 @@
+#include <string>
+#include <cassert>
+
+using namespace std;
+
+int main()
+{
+  for (int i = 0; i < 2; ++i)
+  {
+    string expectedSound = (i == 0) ? "Woof!" : "Meow!";
+    if (i == 0)
+      assert(expectedSound == "Woof!");
+    else
+      assert(expectedSound == "Meow!");
+  }
+
+  return 0;
+}
--- a/regression/esbmc-cpp/bug_fixes/github_1930/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_1930/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+--unwind 7
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/bug_fixes/github_1930_fail/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_1930_fail/main.cpp
@@ -1,0 +1,18 @@
+#include <string>
+#include <cassert>
+
+using namespace std;
+
+int main()
+{
+  for (int i = 0; i < 2; ++i)
+  {
+    string expectedSound = (i == 0) ? "Woof!" : "Meow!";
+    if (i == 0)
+      assert(expectedSound == "Woof!");
+    else
+      assert(expectedSound == "Woof!"); // Meow!
+  }
+
+  return 0;
+}
--- a/regression/esbmc-cpp/bug_fixes/github_1930_fail/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_1930_fail/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+--unwind 7
+
+^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/cpp/ch6_1/test.desc
+++ b/regression/esbmc-cpp/cpp/ch6_1/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/ch6_2/test.desc
+++ b/regression/esbmc-cpp/cpp/ch6_2/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/ch6_4/test.desc
+++ b/regression/esbmc-cpp/cpp/ch6_4/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 Time1.cpp --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/ch6_7/test.desc
+++ b/regression/esbmc-cpp/cpp/ch6_7/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 Time2.cpp --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/ch6_9/test.desc
+++ b/regression/esbmc-cpp/cpp/ch6_9/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 Time3.cpp --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/ch7_8/test.desc
+++ b/regression/esbmc-cpp/cpp/ch7_8/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 Time6.cpp --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
--- a/src/clang-c-frontend/clang_c_adjust_expr.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_expr.cpp
@@ -1360,8 +1360,11 @@ void clang_c_adjust::adjust_if(exprt &expr)
 
   // Typecast both the true and false results
   // If the types are inconsistent
-  gen_typecast(ns, expr.op1(), expr.type());
-  gen_typecast(ns, expr.op2(), expr.type());
+  if (expr.type() != expr.op1().type() || expr.type() != expr.op2().type())
+  {
+    gen_typecast(ns, expr.op1(), expr.type());
+    gen_typecast(ns, expr.op2(), expr.type());
+  }
 }
 
 void clang_c_adjust::align_se_function_call_return_type(
--- a/src/util/c_typecast.cpp
+++ b/src/util/c_typecast.cpp
@@ -872,6 +872,17 @@ void c_typecastt::do_typecast(exprt &dest, const typet &type)
 
   if (dest_type.is_array() || dest_type.is_incomplete_array())
   {
+    if (dest.id() == "if")
+    {
+      // Special case: if expression
+      // To typecast the if expression, we need to apply the operations: true and false
+      dest.type() = type;
+
+      do_typecast(dest.op1(), type);
+      do_typecast(dest.op2(), type);
+      return;
+    }
+
     index_exprt index;
     index.array() = dest;
     index.index() = gen_zero(index_type());


fbrausse:
I think this might be better fixed by properly handling this case here:
https://github.com/esbmc/esbmc/blob/47fa8ec8a7f146d57c5c889893cdfaae21d5525c/src/clang-c-frontend/clang_c_convert.cpp#L2902

At the moment this case is ignored.

XLiZHI:
> I think this might be better fixed by properly handling this case here:
> 
> https://github.com/esbmc/esbmc/blob/47fa8ec8a7f146d57c5c889893cdfaae21d5525c/src/clang-c-frontend/clang_c_convert.cpp#L2902
> 
> At the moment this case is ignored.

Adjusting if expressions is tricky: we need to cast the _true_ and _false_ expressions directly, rather than on top of the _if_ expression (`address_of(if_expr)`). I tried changing CK_ArrayToPointerDecay, but it was a bit complicated and we ended up needing the `adjust_if` method to adjust.

<details>

```diff
diff --git a/src/clang-c-frontend/clang_c_convert.cpp b/src/clang-c-frontend/clang_c_convert.cpp
index b1c471114..2e6aab0ae 100644
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -2899,7 +2899,6 @@ bool clang_c_convertert::get_cast_expr(

   switch (cast.getCastKind())
   {
-  case clang::CK_ArrayToPointerDecay:
   case clang::CK_FunctionToPointerDecay:
   case clang::CK_BuiltinFnToFnPtr:
   case clang::CK_UncheckedDerivedToBase:
@@ -2927,6 +2926,7 @@ bool clang_c_convertert::get_cast_expr(
   case clang::CK_LValueToRValue:
   case clang::CK_LValueBitCast:

+  case clang::CK_ArrayToPointerDecay:
   case clang::CK_PointerToBoolean:
   case clang::CK_PointerToIntegral:
     gen_typecast(ns, expr, type);
diff --git a/src/util/c_typecast.cpp b/src/util/c_typecast.cpp
index 8ebafa61d..e6337c62e 100644
--- a/src/util/c_typecast.cpp
+++ b/src/util/c_typecast.cpp
@@ -665,6 +665,12 @@ void c_typecastt::implicit_typecast_followed(
       return; // ok
     }

+    if (expr.id() == "if")
+    {
+      expr.type() = dest_type;
+      return;
+    }
+
     if (
       src_type.id() == "pointer" || src_type.is_array() ||
       src_type.id() == "incomplete_array")
```

mikhailramalho:
I agree with Franz here, what made supporting CK_ArrayToPointerDecay
complicated?

XLiZHI:
I am okay for both, I attached a patch to my last comment, and I'll try it if it's good.

mikhailramalho:
how does the final expr looks like without your changes to c_typecast.cpp?

XLiZHI:
here is expr without changes to c_typecast.cpp

<details>

```
address_of
  * type: pointer
      * subtype: signedbv
          * width: 8
          * #constant: 1
          * #cpp_type: signed_char
  * operands: 
    0: index
        * type: signedbv
            * width: 8
            * #constant: 1
            * #cpp_type: signed_char
        * operands: 
          0: if
              * type: array
                  * size: constant
                      * type: unsignedbv
                          * width: 64
                      * value: 0000000000000000000000000000000000000000000000000000000000000110
                      * #cformat: 6
                  * subtype: signedbv
                      * width: 8
                      * #constant: 1
                      * #cpp_type: signed_char
                  * #constant: 1
              * operands: 
                0: =
                    * type: bool
                        * #cpp_type: bool
                    * operands: 
                      0: symbol
                          * type: signedbv
                              * width: 32
                              * #cpp_type: signed_int
                          * name: i
                          * identifier: c:main8.cpp@80@F@main#@i
                          * #location: 
                            * file: main8.cpp
                            * line: 10
                            * function: main
                            * column: 29
                          * #lvalue: 1
                      1: constant
                          * type: signedbv
                              * width: 32
                              * #cpp_type: signed_int
                          * value: 00000000000000000000000000000000
                          * #location: 
                            * file: main8.cpp
                            * line: 10
                            * function: main
                            * column: 34
                          * #cformat: 0
                    * #location: 
                      * file: main8.cpp
                      * line: 10
                      * function: main
                      * column: 28
                1: string-constant
                    * type: array
                        * size: constant
                            * type: unsignedbv
                                * width: 64
                            * value: 0000000000000000000000000000000000000000000000000000000000000110
                            * #cformat: 6
                        * subtype: signedbv
                            * width: 8
                            * #constant: 1
                            * #cpp_type: signed_char
                        * #constant: 1
                    * value: Woof!
                    * kind: default
                    * #location: 
                      * file: main8.cpp
                      * line: 10
                      * function: main
                      * column: 39
                2: string-constant
                    * type: array
                        * size: constant
                            * type: unsignedbv
                                * width: 64
                            * value: 0000000000000000000000000000000000000000000000000000000000000110
                            * #cformat: 6
                        * subtype: signedbv
                            * width: 8
                            * #constant: 1
                            * #cpp_type: signed_char
                        * #constant: 1
                    * value: Meow!
                    * kind: default
                    * #location: 
                      * file: main8.cpp
                      * line: 10
                      * function: main
                      * column: 49
              * #location: 
                * file: main8.cpp
                * line: 10
                * function: main
                * column: 28
          1: constant
              * type: signedbv
                  * width: 64
              * value: 0000000000000000000000000000000000000000000000000000000000000000
```

mikhailramalho:
so, the typecast adds the address-off + index[0] to the ternary statement, is that right?

That seems wrong, I would expect the address-off + index[0] to be added to operands of the ternary expression.

XLiZHI:
> so, the typecast adds the address-off + index[0] to the ternary statement, is that right?
> 
> That seems wrong, I would expect the address-off + index[0] to be added to operands of the ternary expression.

Yes! That's why I changed `c_typecast.cpp`, we will do type cast for operands in `adjust_if` method.

mikhailramalho:
Actually, I think that's correct, and moving the cast doesn't seem to have any effect, esbmc is already generating:
```
        FUNCTION_CALL:  string(&tmp$1, &(return_value$_nondet_bool$2 ? { 87, 111, 111, 102, 33, 0 } : { 77, 101, 111, 119, 33, 0 })[0])
```
I think the problem here is the string literal not having the adress-of + index[0]:
```
`-FunctionDecl 0x556bc3dda678 <line:6:1, line:11:1> line:6:5 main 'int ()'
  `-CompoundStmt 0x556bc3ddb090 <line:7:1, line:11:1>
    |-DeclStmt 0x556bc3dda9f0 <line:8:3, col:59>
    | `-VarDecl 0x556bc3dda730 <col:3, col:52> col:10 used expectedSound 'string':'std::string' cinit
    |   `-ImplicitCastExpr 0x556bc3dda8f0 <col:26, col:52> 'string':'std::string' <ConstructorConversion>
    |     `-CXXConstructExpr 0x556bc3dda8c0 <col:26, col:52> 'string':'std::string' 'void (const char *)'
    |       `-ImplicitCastExpr 0x556bc3dda8a8 <col:26, col:52> 'const char *' <ArrayToPointerDecay>
    |         `-ConditionalOperator 0x556bc3dda878 <col:26, col:52> 'const char[6]' lvalue
    |           |-CallExpr 0x556bc3dda818 <col:26, col:38> 'bool'
    |           | `-ImplicitCastExpr 0x556bc3dda800 <col:26> 'bool (*)()' <FunctionToPointerDecay>
    |           |   `-DeclRefExpr 0x556bc3dda7e0 <col:26> 'bool ()' lvalue Function 0x556bc3c2d408 'nondet_bool' 'bool ()'
    |           |-StringLiteral 0x556bc3dda838 <col:42> 'const char[6]' lvalue "Woof!"
    |           `-StringLiteral 0x556bc3dda858 <col:52> 'const char[6]' lvalue "Meow!"
```
I think @fbrausse worked on that at some point, wdyt? 

My guess is that we need to handle the string_constantt in the adjust code and add the address-of + index[0]

fbrausse:
This example is failing with this PR for an assertion in the constructor of address_of2t expression, which is being constructed ontop of another address_of2t:
```c++
#include <string>
#include <cassert>

using namespace std;

int main()
{
  for (int i = 0; i < 2 ; ++i)
  {
    const char (&x)[6] = (i == 0) ? "Woof!" : "Meow!";
    string expectedSound = x;
    bool r;
    if (i == 0)
      r = (expectedSound == "Woof!");
    else
      r = (expectedSound == "Meow!");
    assert(r);
  }

  return 0;
}
```
On master, it also fails, later, at a different location:
```
ERROR: compute_pointer_offset, unexpected irep:
if
[...]
* type : array
  * subtype : signedbv
      * width : 8
  * array_size : constant_int
      * value : 6
      * type : unsignedbv
        * width : 64
  * size_is_infinite : false
```
compute_pointer_offset() basically isn't prepared to handle if2t.

With the fix suggested in https://github.com/esbmc/esbmc/pull/1931#issuecomment-2221219887 there is no change to master for this example.

Regarding this PR: I don't think it's a good idea to unconditionally replace the type of if-exprs with a pointer. Also, the `is_string()` function is controversial, in general (<https://github.com/esbmc/esbmc/pull/1919#discussion_r1661127280>), and here: it doesn't work for multi-dimensional arrays or arrays of structs or unions.

XLiZHI:
I looked into the case, I am a little confused that when we are adjusting `if` expr, we are converting the `string-constant` to the array type.

https://github.com/esbmc/esbmc/blob/de10dee30288a464dbbfa2dfa6575ff3522130e6/src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp#L137-L148

True and False expr is converted into this: 
`&(i == 0 ? { 87, 111, 111, 102, 33, 0 } : { 77, 101, 111, 119, 33, 0 })[0]`

Is it equivalent to string-constant? (i == 0 ? &"Woof!"[0] : &"Meow!"[0]) If not, then we have to avoid this typecast.

fbrausse:
> Is it equivalent to string-constant? (i == 0 ? &"Woof!"[0] : &"Meow!"[0])

It is not. In C the arrays would all implicitly decay to pointers, and we wouldn't be talking about this case. But in C++ they don't decay automatically, because the type system can maintain the "string type" even through function invocations:
```c++
#include <string>
#include <cassert>

using namespace std;

const char (& woof())[6] { return "Woof!"; }
const char (& meow())[6] { return "Meow!"; }

int main()
{
  for (int i = 0; i < 2 ; ++i)
  {
    const char (&x)[6] = (i == 0) ? woof() : meow();
    string expectedSound = x;
    bool r;
    if (i == 0)
      r = (expectedSound == "Woof!");
    else
      r = (expectedSound == "Meow!");
    assert(r);
  }

  return 0;
}
```
> If not, then we have to avoid this typecast.

I agree. Clang is giving us the location where this typecast (decay) is supposed to be performed, as
```
    |       `-ImplicitCastExpr 0x556bc3dda8a8 <col:26, col:52> 'const char *' <ArrayToPointerDecay>
```
This is where we should do it, I think.

fbrausse:
We can use this patch to teach compute_pointer_offset() about if2t, but it still doesn't fix the test case :(
```patch
diff --git a/src/util/type_byte_size.cpp b/src/util/type_byte_size.cpp
index 12ecc4ac52..7055070e93 100644
--- a/src/util/type_byte_size.cpp
+++ b/src/util/type_byte_size.cpp
@@ -421,6 +421,15 @@ expr2tc type_sizet::pointer_offset_bits(const expr2tc &expr) const
     return bitsize(0);
   }
 
+  if (is_if2t(expr)) {
+    const if2t &the_if = to_if2t(expr);
+    expr2tc yes = pointer_offset_bits(the_if.true_value);
+    expr2tc no = pointer_offset_bits(the_if.false_value);
+    expr2tc result = if2tc(yes->type, the_if.cond, yes, no);
+    expr2tc simpl = result->do_simplify();
+    return simpl ? simpl : result;
+  }
+
   log_error("compute_pointer_offset, unexpected irep:\n{}", expr->pretty());
   abort();
 }
```

fbrausse:
Let's look at the general approach for the C++ references. My understanding: Translate every such reference (const, lvalue or not) to a pointer, dereference it, and set a flag indicating that it's been a reference. I still think that's a good approach (but maybe we should move the flag to the pointer, see below). If we look at the original example, `string(cond ? "x" : "y")`, it works: `string(&(cond ? "y" : "z")[0])` is fine, where `"y"` and `"z"` are string-constants.

Btw. string_constantt is just a special kind of constant-array. The only information it keeps in addition to the constant array is the encoding given in the source file. This is needed in particular for Windows, where the default type of strings passed via assert() is not a `char []` but a `wchar_t []`. From this we can infer the encoding.

For my first example this approach should also work - if we do a case distinction: `const char (&x)[2] = cond ? "y" : "z";` would be translated to `const char *x = &(cond ? "y" : "z")[0];`. The assignment to the translation of `const char (&x)[2]` would need to make the case distinction: am I assigning a reference to another reference? Then just assign the pointers. Otherwise, either we're assigning a reference to a non-reference (which means the pointer corresponding to the reference has to be dereferenced), or we're assigning a value to a reference (in which case we need to take the address of the value). This I tried to do in the two cases in c_typecastt::implicit_typecast_followed().

Ideally, it would also work for my last example, where the functions' return type would be "dereference of pointer to char, with is_reference flag set". But unfortunately, I think, it does not work. Dereference is not part of **our** type system, i.e., the return type (pointer to char) wouldn't be tagged with "this actually is a reference". That's a problem. Not necessarily for this PR, but I think we may need to move from
- tag dereferenced pointer with "this is a C++ reference"

to 
- tag pointer with "this is the address of a C++ reference".

What do you think @mikhailramalho @XLiZHI?

XLiZHI:
> We can use this patch to teach compute_pointer_offset() about if2t, but it still doesn't fix the test case :(
> 
> ```diff
> diff --git a/src/util/type_byte_size.cpp b/src/util/type_byte_size.cpp
> index 12ecc4ac52..7055070e93 100644
> --- a/src/util/type_byte_size.cpp
> +++ b/src/util/type_byte_size.cpp
> @@ -421,6 +421,15 @@ expr2tc type_sizet::pointer_offset_bits(const expr2tc &expr) const
>      return bitsize(0);
>    }
>  
> +  if (is_if2t(expr)) {
> +    const if2t &the_if = to_if2t(expr);
> +    expr2tc yes = pointer_offset_bits(the_if.true_value);
> +    expr2tc no = pointer_offset_bits(the_if.false_value);
> +    expr2tc result = if2tc(yes->type, the_if.cond, yes, no);
> +    expr2tc simpl = result->do_simplify();
> +    return simpl ? simpl : result;
> +  }
> +
>    log_error("compute_pointer_offset, unexpected irep:\n{}", expr->pretty());
>    abort();
>  }
> ```

This was my initial attempt, but I didn't use it because when we were working with the following code:
```cpp
#include <string>
#include <cassert>

using namespace std;

int main()
{
  for (int i = 0; i < 2; ++i)
  {
    string expectedSound = (i == 0) ? "Woof!" : "Meow";    // has diff size [5] and [4]
  }

  return 0;
}
```
```

-ImplicitCastExpr 0x730ede0 <col:39> 'const char *' <ArrayToPointerDecay>
 | `-StringLiteral 0x730ecd8 <col:39> 'const char [6]' lvalue "Woof!"
-ImplicitCastExpr 0x730edf8 <col:49> 'const char *' <ArrayToPointerDecay>
 |`-StringLiteral 0x730edc0 <col:49> 'const char [5]' lvalue "Meow"
```
This code works, so we probably don't need to add the `if2t` case.

> or my first example this approach should also work - if we do a case distinction: const char (&x)[2] = cond ? "y" : "z"; would be translated to const char *x = &(cond ? "y" : "z")[0];. The assignment to the translation of const char (&x)[2] would need to make the case distinction: am I assigning a reference to another reference? Then just assign the pointers. Otherwise, either we're assigning a reference to a non-reference (which means the pointer corresponding to the reference has to be dereferenced), or we're assigning a value to a reference (in which case we need to take the address of the value). This I tried to do in the two cases in c_typecastt::implicit_typecast_followed().

Maybe we can adjust the if expression to do this?

I think the key point is how to handle address_of(if). Ideally, we do this directly inside the if expression instead of on the outside.

mikhailramalho:
> 
> * tag dereferenced pointer with "this is a C++ reference"
> 
> to
> 
> * tag pointer with "this is the address of a C++ reference".
> 
> What do you think @mikhailramalho @XLiZHI?

I agree we should move at some point, I'm just worried about the implementation and how long it would take change this. @XLiZHI Is this something that can be added to your roadmap?

> ```
> 
> -ImplicitCastExpr 0x730ede0 <col:39> 'const char *' <ArrayToPointerDecay>
>  | `-StringLiteral 0x730ecd8 <col:39> 'const char [6]' lvalue "Woof!"
> -ImplicitCastExpr 0x730edf8 <col:49> 'const char *' <ArrayToPointerDecay>
>  |`-StringLiteral 0x730edc0 <col:49> 'const char [5]' lvalue "Meow"
> ```
> 

But this is correct because of the array-to-pointer decay no?

> 
> > or my first example this approach should also work - if we do a case distinction: const char (&x)[2] = cond ? "y" : "z"; would be translated to const char *x = &(cond ? "y" : "z")[0];. The assignment to the translation of const char (&x)[2] would need to make the case distinction: am I assigning a reference to another reference? Then just assign the pointers. Otherwise, either we're assigning a reference to a non-reference (which means the pointer corresponding to the reference has to be dereferenced), or we're assigning a value to a reference (in which case we need to take the address of the value). This I tried to do in the two cases in c_typecastt::implicit_typecast_followed().
> 
> Maybe we can adjust the if expression to do this?
> 
> I think the key point is how to handle address_of(if). Ideally, we do this directly inside the if expression instead of on the outside.

IIUC, the problem right now is the compute_pointer_offset_bits function, right? This function tries to compute the offset of an expr2tc, why did you decide not to go forward with the changes you mentioned?

XLiZHI:
> But this is correct because of the array-to-pointer decay no?

Yes, I mean, in this case the 'compute_pointer_offset_bits' function doesn't run into the `if2t` case, maybe a simplification was made somewhere and we could probably avoid adding it.


> IIUC, the problem right now is the compute_pointer_offset_bits function, right? This function tries to compute the offset of an expr2tc, why did you decide not to go forward with the changes you mentioned?

This is because there was assertion fail in the subsequent string comparison. If we don't first solve the problem mentioned above, then the assignment here should be incorrect.
```cpp
    if (i == 0)
      assert(expectedSound == "Woof!");
    else
      assert(expectedSound == "Meow!");
```

lucasccordeiro:
@XLiZHI: Can I ask you for the status of this PR?

lucasccordeiro:
@XLiZHI: Can I ask you for the status of this PR?

lucasccordeiro:
@XLiZHI: can you please run this PR over the SV-COMP benchmarks?

XLiZHI:
This PR: 

````
  correct:           18724
    correct true:    11389
    correct false:    7335
  incorrect:            85
    incorrect true:     64
    incorrect false:    21
  unknown:           14760
  Score:             27729 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13429233055
````

Master:
```
  correct:           18686
    correct true:    11354
    correct false:    7332
  incorrect:            84
    incorrect true:     63
    incorrect false:    21
  unknown:           14799
  Score:             27688 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13408677525
```

lucasccordeiro:
@XLiZHI: there is one additional incorrect result. Could you please check whether it is just "noise"?

XLiZHI:
> @XLiZHI: there is one additional incorrect result. Could you please check whether it is just "noise"?

image

lucasccordeiro (approved):

lucasccordeiro:

PR #1933, opened 2024-07-11
[python-frontend] Implement Operational Modelss that are imported

ivomachado:
Start with a very small math model implementation to showcase the feature.

This allows a python file to import `math` and import our OM for it.


lucasccordeiro (approved):

src/python-frontend/parser.py
```diff
@@ -140,11 +148,13 @@ def main():
...
     # Iterate over all .py files in the directory
     for python_file in glob.glob(os.path.join(models_dir, "*.py")):
-        with open(python_file) as model:
-            model_tree = ast.parse(model.read())
-            filename = os.path.basename(python_file)
-            # Generate JSON from AST for the memory models.
-            generate_ast_json(model_tree, filename, None, output_dir)
+        filename = os.path.basename(python_file)
+        module_name = filename[:-3]
+        if not is_imported_model(module_name):
```
brcfarias: Could change to “if is_imported continue”?

src/python-frontend/models/math.py
brcfarias, line 1: Is it possible to add a math function to this initial version?

brcfarias (approved):

src/python-frontend/models/math.py
ivomachado: yes, it is. I'm going to add floor, ok?

lucasccordeiro:
Thanks for submitting this PR, @ivomachado.

PR #1935, opened 2024-07-16
[python-frontend] Handling int.bit_length()

brcfarias:
This PR adds support for [int.bit_length()](https://docs.python.org/3/library/stdtypes.html) by:
- Providing an operational model for this function.
- Converting `obj.bit_length()` calls to `bit_length(obj)` in the python_converter.

Additionally, I'm removing the [is_string](https://github.com/esbmc/esbmc/blob/de10dee30288a464dbbfa2dfa6575ff3522130e6/src/util/type.cpp#L22C1-L25C2) function from util/type as suggested [here](https://github.com/esbmc/esbmc/pull/1919#discussion_r1661127280).

### DIFF
--- a/regression/python/int_bit_length/test.desc
+++ b/regression/python/int_bit_length/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -326,7 +326,18 @@ exprt python_converter::get_binary_operator_expr(const nlohmann::json &element)
 
   assert(!op.empty());
 
-  if (op == "Eq" && is_string(lhs.type()) && is_string(rhs.type()))
+  std::string lhs_type = get_var_type(lhs.name().as_string());
+  std::string rhs_type = get_var_type(rhs.name().as_string());
+
+  if (
+    rhs_type.empty() && element.contains("comparators") &&
+    element["comparators"][0].contains("value") &&
+    element["comparators"][0]["value"].is_string())
+  {
+    rhs_type = "str";
+  }
+
+  if (op == "Eq" && lhs_type == "str" && rhs_type == "str")
   {
     if (rhs.type() != lhs.type())
       return gen_boolean(false);
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -94,7 +94,7 @@ static struct_typet::componentt build_component(
   return comp;
 }
 
-// Convert Python/AST types to irep2 types
+// Convert Python/AST types to irep types
 typet python_converter::get_typet(const std::string &ast_type, size_t type_size)
 {
   if (ast_type == "float")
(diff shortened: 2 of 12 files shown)


src/python-frontend/python_converter.cpp
```diff
@@ -694,19 +706,32 @@ exprt python_converter::get_function_call(const nlohmann::json &element)
...
-      if (func_name == "str")
-        arg_size = arg["value"].get<std::string>().size(); // get string length
+      obj_symbol_id = create_symbol_id() + "@" + caller;
+      obj_symbol = context.find_symbol(obj_symbol_id);
 
-      typet t = get_typet(func_name, arg_size);
-      exprt expr = get_expr(arg);
-      expr.type() = t;
-      return expr;
+      if (
```
lucasccordeiro, line 735: Can you write some documentation here?

src/python-frontend/python_converter.cpp
```diff
@@ -1147,6 +1176,26 @@ size_t get_type_size(const nlohmann::json &ast_node)
...
+  {
+    if (elem["_type"] == "FunctionDef" && elem["name"] == current_func_name)
+      ref = find_var_decl(var_name, elem);
+  }
+
+  // Get variable from global scope
+  if (ref.empty())
+    ref = find_var_decl(var_name, ast_json);
+
+  if (ref.empty())
```
lucasccordeiro, line 1203: Why do we return `std::string()`?

src/python-frontend/python_converter.cpp
```diff
@@ -670,6 +667,21 @@ exprt python_converter::get_function_call(const nlohmann::json &element)
       return rhs;
     }
 
+    if (is_builtin_type(func_name) || is_consensus_type(func_name))
+    {
+      // Replace the function call with a constant value. For example, x = int(1) becomes x = 1
```
lucasccordeiro: Why do we need to do this? Give some rationale in the comments.

src/python-frontend/python_converter.cpp
```diff
@@ -326,7 +326,18 @@ exprt python_converter::get_binary_operator_expr(const nlohmann::json &element)
 
   assert(!op.empty());
 
-  if (op == "Eq" && is_string(lhs.type()) && is_string(rhs.type()))
+  std::string lhs_type = get_var_type(lhs.name().as_string());
+  std::string rhs_type = get_var_type(rhs.name().as_string());
+
+  if (
```
lucasccordeiro, line 334: Add a comment here.

regression/python/int_bit_length/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: Can you also add a test case that gives `VERIFICATION FAILED`?

lucasccordeiro (changes requested):

src/python-frontend/python_converter.cpp
brcfarias: We return an empty string when the variable is not found in the program.

src/python-frontend/python_converter.cpp
brcfarias: I added a comment explaining it.

lucasccordeiro (approved):

lucasccordeiro:

PR #1937, opened 2024-07-17
[python-frontend] Inferring type from function returns

brcfarias:
This PR enables type inference from function returns in scenarios like:

```
def foo() -> int :
  return 1

x = foo()  # Inferred type of x: int
```


lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:

PR #1938, opened 2024-07-17
Fix rvalue-to-array problem for expression statements and C++ ctor init lists

### LINKED ISSUE #1934: [C++ Verification] Can't construct rvalue reference to array type during dereference
```c++
class Base
{
public:
  int ss[128];
  Base() 
  {
    ss;
  }
};

int main()
{
  Base x;
}
```

I got 
```sh
Parsing 1.cpp
1.cpp:13:5: warning: expression result unused [-Wunused-value]
    ss;
    ^~
Converting
Generating GOTO Program
GOTO program creation time: 0.455s

[1]    11807 IOT instruction  ./esbmc 1.cpp
```

If I comment the abort statements in `dereference.cpp`
```c++
  else if (is_array_type(type))
  {
     // log_error(
     //  "Can't construct rvalue reference to array type during dereference\n"
     // "(It isn't allowed by C anyway)\n");
    // abort();
    ;
  }
```
then everything works fine, including the case shown below

```c++
#include <assert.h>
class Base
{
public:
  int ss[128];
  Base() : ss{0} 
  {
  }
};

int main()
{
  Base x;
  assert(x.ss[1] == 0); // hold
  assert(x.ss[0] == 1); // fail
}
```

@XLiZHI , do you have any ideas on this? Since I need this feature for Solidity, would it be okay if I just comment the abort?

fbrausse:
Closes #1934 by making an adjustment in the frontend to the specific type of statements (expression, that is not an assignment, and has array-like type) the dereference module still chokes on. I don't know of any other contexts of array-typed dereferences besides those identified in that issue. There may be more.

*Edit:* I've added printing the type of the statements to the output of --goto-functions-(only|too), because it's much clearer. If it isn't for others, I can remove it from this PR.

*Edit 2:* Thanks to @XLiZHI providing another failing example, there now also is a fix for C++ constructor initializers of arrays. Those aren't initialized by assignment anymore with this PR, but the symbol itself is updated. I've added the example as a regression test.


XLiZHI:
I tried this idea and it won't handle list initialization if we change it like this.

```cpp
class Base
{
public:
  int ss[3];
  Base() : ss{0} 
  {
  }
};
```
it will be `this->ss[0] = {0, 0, 0};` An error will occur with this assignment

fbrausse:
~Well, this is because assignments to arrays are not supported. What we do in C is to "initialize" them. That's not an assignment in the sense of symex_assign(), is it?~

*Edit:* Never mind, you're right. Back to square one.

fbrausse:
@XLiZHI struct and union assignments are fine, right? And C++ classes are not arrays, but basically structs. So why do we assign members of those structs in the constructor when the rest of the language isn't prepared to assign arrays? Would it be possible to assign the entire struct? The initial value (as created by the constructor) could be created by the frontend not as a sequence of ASSIGN instructions, but really as a constant_exprt of struct_typet, no?

XLiZHI:
> @XLiZHI struct and union assignments are fine, right? And C++ classes are not arrays, but basically structs. So why do we assign members of those structs in the constructor when the rest of the language isn't prepared to assign arrays? Would it be possible to assign the entire struct? The initial value (as created by the constructor) could be created by the frontend not as a sequence of ASSIGN instructions, but really as a constant_exprt of struct_typet, no?

That's a good idea. If I understand correctly, we modify the initial value of the member directly in the constructor, instead of assigning to the member to initialize it, right?

fbrausse:
> If I understand correctly, we modify the initial value of the member directly in the constructor, instead of assigning to the member to initialize it, right?

Yes, exactly. Assign only the fully initialized struct-type'd expression. No individual member assignments (except for base classes, which by the same token, are surely not arrays).

XLiZHI:
Here are some of my attempts:

First I want to get struct_typet here `const symbolt *t = ns.lookup(lhs.op0().type().subtype().identifier());`
https://github.com/esbmc/esbmc/blob/4a5e3e77293a36c0510d87b40a6230c632c8c30b/src/clang-cpp-frontend/clang_cpp_convert.cpp#L1144-L1150

I got the following expression:
```cpp
Symbol......: tag-Base
Base name...: Base
Module......: main6
Mode........: C++ (C++)
Type........: struct
      * tag: Base
      * components: 
        0: component
            * type: array
                * size: constant
                    * type: unsignedbv
                        * width: 64
                    * value: 0000000000000000000000000000000000000000000000000000000000000011
                    * #cformat: 3
                * subtype: signedbv
                    * width: 32
                    * #cpp_type: signed_int
                * #member_name: tag-Base
            * name: ss
            * access: public
            * pretty_name: ss
            * #location: 
              * file: main6.cpp
              * line: 4
              * column: 3
      * #location: 
        * file: main6.cpp
        * line: 1
        * column: 1
```
Basically, it's not clear to me what a `constant_exprt` of `struct_typet` with values should look like, The value of a `component` is assigned when the constructor is called.

Also, I think it's worth trying another approach (although it's not efficient):
```
this->ss[0] = 0;
this->ss[1] = 1;
...
```
This seems to be in line with how compilers work.

Finally, I experimented with directly commenting the code where the error occurred, and the assignment was correct. I guess we probably won't have to worry about boundary access, clang will do the type checking.

fbrausse:
Thanks for looking into it!

For instance with C
```c
struct Base {
  int ss[3];
  char *tmp;
} x = { { 1, 2, 3 }, (void *)0 };

int main() {
	x.ss;
}
```
the symbol `c:@x` would look like this in the context:
<details>

```
(gdb) p context.find_symbol("c:@x")->dump()
Symbol......: c:@x
Base name...: x
Module......: gh-1938
Mode........: C (C)
Type........: struct
      * tag: struct Base
      * components: 
        0: component
            * type: array
                * size: constant
                    * type: unsignedbv
                        * width: 64
                    * value: 0000000000000000000000000000000000000000000000000000000000000011
                    * #cformat: 3
                * subtype: signedbv
                    * width: 32
                    * #cpp_type: signed_int
            * name: ss
            * pretty_name: ss
            * #location: 
              * file: gh-1938.c
              * line: 2
              * column: 3
        1: component
            * type: unsignedbv
                * width: 32
            * name: anon_pad$1
            * pretty_name: anon_pad$1
            * is_padding: 1
        2: component
            * type: pointer
                * subtype: signedbv
                    * width: 8
                    * #cpp_type: signed_char
            * name: tmp
            * pretty_name: tmp
            * #location: 
              * file: gh-1938.c
              * line: 3
              * column: 3
      * #location: 
        * file: gh-1938.c
        * line: 1
        * column: 1
Value.......: struct
      * type: struct
          * tag: struct Base
          * components: 
            0: component
                * type: array
                    * size: constant
                        * type: unsignedbv
                            * width: 64
                        * value: 0000000000000000000000000000000000000000000000000000000000000011
                        * #cformat: 3
                    * subtype: signedbv
                        * width: 32
                        * #cpp_type: signed_int
                * name: ss
                * pretty_name: ss
                * #location: 
                  * file: gh-1938.c
                  * line: 2
                  * column: 3
            1: component
                * type: unsignedbv
                    * width: 32
                * name: anon_pad$1
                * pretty_name: anon_pad$1
                * is_padding: 1
            2: component
                * type: pointer
                    * subtype: signedbv
                        * width: 8
                        * #cpp_type: signed_char
                * name: tmp
                * pretty_name: tmp
                * #location: 
                  * file: gh-1938.c
                  * line: 3
                  * column: 3
          * #location: 
            * file: gh-1938.c
            * line: 1
            * column: 1
      * operands: 
        0: constant
            * type: array
                * size: constant
                    * type: unsignedbv
                        * width: 64
                    * value: 0000000000000000000000000000000000000000000000000000000000000011
                    * #cformat: 3
                * subtype: signedbv
                    * width: 32
                    * #cpp_type: signed_int
            * operands: 
              0: constant
                  * type: signedbv
                      * width: 32
                      * #cpp_type: signed_int
                  * #location: 
                    * file: gh-1938.c
                    * line: 4
                    * column: 11
                  * #cformat: 1
              1: constant
                  * type: signedbv
                      * width: 32
                      * #cpp_type: signed_int
                  * value: 00000000000000000000000000000010
                  * #location: 
                    * file: gh-1938.c
                    * line: 4
                    * column: 14
                  * #cformat: 2
              2: constant
                  * type: signedbv
                      * width: 32
                      * #cpp_type: signed_int
                  * value: 00000000000000000000000000000011
                  * #location: 
                    * file: gh-1938.c
                    * line: 4
                    * column: 17
                  * #cformat: 3
            * #location: 
              * file: gh-1938.c
              * line: 4
              * column: 9
        1: constant
            * type: unsignedbv
                * width: 32
            * value: 00000000000000000000000000000000
        2: constant
            * type: pointer
                * subtype: signedbv
                    * width: 8
                    * #cpp_type: signed_char
            * value: NULL
            * #location: 
              * file: gh-1938.c
              * line: 4
              * column: 22
      * #location: 
        * file: gh-1938.c
        * line: 4
        * column: 7
Flags.......: lvalue static_lifetime
Location....: file gh-1938.c line 1 column 1
```

</details>

The `value` is an `exprt("struct")` with the struct's type (which includes a padding member) and the operands of the "struct"-expression are the initializers for the members.

XLiZHI (commented):
Nice! I'm fine with the change. But I do have a little thought about what we can do with the `ss` variable.   
`Base() {ss; }`
I think the ss variable that appears here is useless, isn't it? Why don't we just ignore this kind of expression and not execute it, like CBMC does. I think this will avoid unnecessary adjustments, as this is the edge case.

fbrausse:
> I think the ss variable that appears here is useless, isn't it? Why don't we just ignore this kind of expression and not execute it, like CBMC does. I think this will avoid unnecessary adjustments, as this is the edge case.

Because of [that](https://github.com/esbmc/esbmc/issues/1934#issuecomment-2232584628):
```c
struct Base
{
  int ss[128];
};

int main()
{
  (*(struct Base *volatile *)0)->ss;
}
```
Compiled with Clang and GCC it fails at runtime with a segfault. ESBMC should catch that.

fbrausse:
Master:
```
  correct:           14125
    correct true:     7755
    correct false:    6370
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9638
  Score:             21000 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/9990181610
```
This PR:
```
  correct:           14130
    correct true:     7759
    correct false:    6371
  incorrect:            42
    incorrect true:     13
    incorrect false:    29
  unknown:            9633
  Score:             21009 (max: 38482)

https://github.com/esbmc/esbmc/actions/runs/9990179646
```

XLiZHI (approved):

lucasccordeiro (approved):

src/clang-c-frontend/clang_c_adjust_code.cpp
```diff
@@ -42,6 +42,23 @@ void clang_c_adjust::adjust_code(codet &code)
     adjust_decl_block(code);
   else
   {
+    if (statement == "expression" && is_array_like(code.op0().type()))
+    {
+      /* An array-type'd statement like "y->ss;" where y is a pointer to
+       *
+       *   struct { int ss[128]; }
+       *
+       * is not assumed to not exist by the dereference code. Thus, convert it
```
intrigus-lgtm, line 51: Is this double negation intentional?

fbrausse: It's not. ;) Thanks, I'll fix the comment.

PR #1940, opened 2024-07-18
[concurrency] support data race checking for GOTO instructions

### LINKED ISSUE #1936: [Concurrency] ESBMC misses a simple case of data-race (read/write)
$esbmc data-race.c --data-races-check

````C
#include <pthread.h>
#include <assert.h>

int x;

void* taskA(void* arg)
{
  x++;
  return NULL;
}

void* taskB(void* arg)
{
  if(!x)
    assert(1);
  return NULL;
}

int main() {
  pthread_t idA, idB;

  pthread_create(&idA, NULL, taskA, NULL);
  pthread_create(&idB, NULL, taskB, NULL);

  return 0;
}
````

### PR DESCRIPTION
**Pull request by XLiZHI**
This PR adds a data race check for GOTO.

`if (!x)` Will be treated as reading `x`:

```cpp
        ASSIGN tmp_c:@x=false;
        ATOMIC_BEGIN
        ASSERT !tmp_c:@x // R/W data race on c:@x
        ATOMIC_END
        IF !(!(_Bool)x) THEN GOTO 1
```

_Still to explore:_ The C++ frontend treats Pointers differently than C, but the internal library GOTO should be the same (they are parsed and converted at build). The number of threads interleaved in C++ is mostly increased compared to C, which needs to be explored further.

### DIFF
--- a/regression/esbmc-unix/github_1936/main.c
+++ b/regression/esbmc-unix/github_1936/main.c
@@ -1,0 +1,32 @@
+#include <pthread.h>
+#include <assert.h>
+
+int x;
+
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+void* taskA(void* arg)
+{
+  pthread_mutex_lock(&m);
+  x++;
+  pthread_mutex_unlock(&m);
+  return NULL;
+}
+
+void* taskB(void* arg)
+{
+  pthread_mutex_lock(&m);
+  if(!x)
+    assert(1);
+  pthread_mutex_unlock(&m);
+  return NULL;
+}
+
+int main() {
+  pthread_t idA, idB;
+
+  pthread_create(&idA, NULL, taskA, NULL);
+  pthread_create(&idB, NULL, taskB, NULL);
+
+  return 0;
+}
--- a/regression/esbmc-unix/github_1936/test.desc
+++ b/regression/esbmc-unix/github_1936/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--data-races-check --context-bound 2
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-unix/github_1936_fail/main.c
+++ b/regression/esbmc-unix/github_1936_fail/main.c
@@ -1,0 +1,26 @@
+#include <pthread.h>
+#include <assert.h>
+
+int x;
+
+void* taskA(void* arg)
+{
+  x++;
+  return NULL;
+}
+
+void* taskB(void* arg)
+{
+  if(!x)
+    assert(1);
+  return NULL;
+}
+
+int main() {
+  pthread_t idA, idB;
+
+  pthread_create(&idA, NULL, taskA, NULL);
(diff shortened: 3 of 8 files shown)


lucasccordeiro:

Your PR seems to fix this regression test:

````
	380 - regression/esbmc-unix2/11_podelski.fig3.lics04 (Failed)
````

For this regression test that times out, you should check whether it is possible to optimize our implementation. If this task is huge, we can mark it as THOROUGH and open an issue for further investigation.
 
````
	2612 - regression/esbmc-cpp/cpp/cpp_with_pthread_02 (Timeout)
````

We should also run this PR over the SV-COMP benchmarks, particularly the concurrency category.

XLiZHI:
> For this regression test that times out, you should check whether it is possible to optimize our implementation. If this task is huge, we can mark it as THOROUGH and open an issue for further investigation.

This timeout is strange to me because the changes in this PR should not affect this test case.

lucasccordeiro:
You could rerun the CI to check whether the timeout still occurs.

XLiZHI:
> You could rerun the CI to check whether the timeout still occurs.

It seems that this problem only happens in C++ (The number of thread interleavings is very large), if I change the filename from C++ to C, the verification is very fast. Also, I've run the benchmark for sv-comp, but it seems to keep queuing (https://github.com/esbmc/esbmc/actions/runs/9997597630)

lucasccordeiro:
@XLiZHI: could you please compare the GOTO representation of this program's C and C++ versions?

Since the pthread library heavily relies on pointers, I suspect that our C and C++ frontends handle pointers differently.

lucasccordeiro:
This PR (https://github.com/esbmc/esbmc/actions/runs/10059866749):

````
2024-07-23T17:32:07.7199223Z Statistics:          23805 Files
2024-07-23T17:32:07.7201253Z   correct:           14129
2024-07-23T17:32:07.7202807Z     correct true:     7755
2024-07-23T17:32:07.7203958Z     correct false:    6374
2024-07-23T17:32:07.7205040Z   incorrect:            42
2024-07-23T17:32:07.7229477Z     incorrect true:     13
2024-07-23T17:32:07.7231349Z     incorrect false:    29
2024-07-23T17:32:07.7232456Z   unknown:            9634
2024-07-23T17:32:07.7233726Z   Score:             21004 (max: 38482)
````

Master (https://github.com/esbmc/esbmc/actions/runs/10059869614):

````
2024-07-23T17:21:21.9101601Z Statistics:          23805 Files
2024-07-23T17:21:21.9103138Z   correct:           14145
2024-07-23T17:21:21.9104213Z     correct true:     7766
2024-07-23T17:21:21.9105144Z     correct false:    6379
2024-07-23T17:21:21.9106319Z   incorrect:            42
2024-07-23T17:21:21.9111562Z     incorrect true:     13
2024-07-23T17:21:21.9112777Z     incorrect false:    29
2024-07-23T17:21:21.9113694Z   unknown:            9618
2024-07-23T17:21:21.9114640Z   Score:             21031 (max: 38482)
````

lucasccordeiro (approved):

lucasccordeiro:

PR #1941, opened 2024-07-19
[python-frontend] Strings concatenation

brcfarias:
This PR addresses strings concatenations, such as:

```
str1 = "He"
str2 = "llo"
str3 = str1 + str2
assert str3 == "Hello"
```

In the `python_converter` a resulting string `expr` is generated by combining the values of  each operand `expr`.

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -310,8 +310,7 @@ std::string python_converter::get_operand_type(const nlohmann::json &element)
 
 exprt python_converter::get_binary_operator_expr(const nlohmann::json &element)
 {
-  auto left =
-    (element.contains("left")) ? element["left"] : element["target"];
+  auto left = (element.contains("left")) ? element["left"] : element["target"];
 
   decltype(left) right;
   if (element.contains("right"))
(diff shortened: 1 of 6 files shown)


src/python-frontend/python_converter.cpp
```diff
@@ -1307,7 +1399,17 @@ void python_converter::get_var_assign(
   if (has_value && rhs != exprt("_init_undefined"))
   {
     if (lhs_symbol)
+    {
+      if (lhs_type == "str")
```
lucasccordeiro, line 1400: Is this "str" hardcoded?

lucasccordeiro, line 1400: Can you add some documentation here?

lucasccordeiro (changes requested):

src/python-frontend/python_converter.cpp
brcfarias: Yes. **str** is a built-in class in Python used to represent strings: https://docs.python.org/3/library/stdtypes.html#str

lucasccordeiro (approved):

lucasccordeiro:

PR #1950, opened 2024-07-25
Fix Python process creation for Windows

brcfarias:
- Replacing hardcoded `"/parser.py"` path with boost::filesystem::path to fix issue on Windows where "\\" is the path separator.
- Checking OS to determine Python executable name


lucasccordeiro (approved):

PR #1953, opened 2024-07-26
[Concurrency] Enable race check for __VERIFIER_assert

### LINKED ISSUE #1952: [data-races check] check instrumentation for __VERIFIER_assert and assert
If we replace `__VERIFIER_assert` with `assert`, ESBMC can detect the data-race in this C program extracted from SV-COMP:

$ time esbmc qw2004-1.c --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --32  --no-por --context-bound 2 --no-pointer-check --no-bounds-check --no-assertions --incremental-bmc --data-races-check

````C
#include <pthread.h>
#include <assert.h>

volatile int stoppingFlag;
volatile int pendingIo;
volatile int stoppingEvent;
volatile int stopped;

int BCSP_IoIncrement() {
    __VERIFIER_atomic_begin();
    if (stoppingFlag) {
        __VERIFIER_atomic_end();
        return -1;
    } else {
        pendingIo = pendingIo + 1;
    }
    __VERIFIER_atomic_end();
    return 0;
}

int dec() {
    __VERIFIER_atomic_begin();
    pendingIo--;
    int tmp = pendingIo;
    __VERIFIER_atomic_end();
    return tmp;
}

void BCSP_IoDecrement() {
    int pending;
    pending = dec();
    if (pending == 0) {
        stoppingEvent = 1;
    }
}

void* BCSP_PnpAdd(void* arg) {
    int status;
    status = BCSP_IoIncrement();
    if (status == 0) {
        __VERIFIER_assert(!stopped);
    }
    BCSP_IoDecrement();
    return 0;
}

void* BCSP_PnpStop(void* arg) {
    stoppingFlag = 1;
    BCSP_IoDecrement();
    assume_abort_if_not(stoppingEvent);
    stopped = 1;
    return 0;
}

int main() {
    pthread_t t;
    pendingIo = 1;
    stoppingFlag = 0;
    stoppingEvent = 0;
    stopped = 0;
    pthread_create(&t, 0, BCSP_PnpStop, 0);
    BCSP_PnpAdd(0);
    return 0;
}
````

### PR DESCRIPTION
**Pull request by XLiZHI**
We use the flag of `--no-assertions` in SV-COMP, which causes the assertion not to be converted to GOTO. So I made some changes to `__VERIFIER_assert` to get it converted and race checked.

### DIFF
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -34,39 +34,15 @@ void rw_sett::compute(const exprt &expr)
     {
       assert(code.operands().size());
       read_write_rec(code.op0(), false, true, "", guardt(), exprt());
-      // check if the functions is __VERIFIER_assert(!cond)
-      if (code.op1().identifier() == "c:@F@__VERIFIER_assert")
-      {
-        // check the cond, usually has only one arg
-        assert(code.op2().operands().size() == 1);
-        read_rec(code.op2().op0());
-      }
-    }
-  }
-  else if (instruction.is_goto() || instruction.is_assert())
-  {
-    if (expr.is_not())
-    {
-      assert(expr.operands().size() == 1);
-      compute(expr.op0());
-    }
-    else if (expr.id() == "=" || expr.is_notequal())
-    {
-      assert(expr.operands().size() == 2);
-      read_rec(expr.op0());
-      read_rec(expr.op1());
-    }
-    else if (expr.is_or() || expr.is_and())
-    {
-      assert(expr.operands().size());
-      forall_operands (it, expr)
+      // check args of function call
+      Forall_operands (it, code.op2())
         read_rec(*it);
     }
-    else if (expr.is_typecast())
-    {
-      assert(expr.operands().size() == 1);
-      read_rec(expr.op0());
-    }
+  }
+  else if (
+    instruction.is_goto() || instruction.is_assert() || instruction.is_assume())
+  {
+    read_rec(expr);
   }
 }
 
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -35,8 +35,12 @@ void rw_sett::compute(const exprt &expr)
       assert(code.operands().size());
       read_write_rec(code.op0(), false, true, "", guardt(), exprt());
       // check args of function call
-      Forall_operands (it, code.op2())
-        read_rec(*it);
+      if (
+        !has_prefix(instruction.location.function(), "ESBMC_execute_kernel") &&
+        !has_prefix(instruction.location.function(), "ESBMC_verify_kernel"))
+        Forall_operands (it, code.op2())
+          if (!(*it).type().is_pointer())
+            read_rec(*it);
     }
   }
   else if (
--- a/src/goto-programs/rw_set.cpp
+++ b/src/goto-programs/rw_set.cpp
@@ -30,6 +30,18 @@ void rw_sett::compute(const exprt &expr)
       assert(code.operands().size() == 1);
       read_rec(code.op0());
     }
+    else if (statement == "function_call")
+    {
+      assert(code.operands().size());
+      read_write_rec(code.op0(), false, true, "", guardt(), exprt());
+      // check if the functions is __VERIFIER_assert(!cond)
+      if (code.op1().identifier() == "c:@F@__VERIFIER_assert")
+      {
+        // check the cond, usually has only one arg
+        assert(code.op2().operands().size() == 1);
+        read_rec(code.op2().op0());
+      }
+    }
   }
   else if (instruction.is_goto() || instruction.is_assert())
   {
--- a/regression/cuda/Supported_long_time/005_soundness_issue_pair/test.desc
+++ b/regression/cuda/Supported_long_time/005_soundness_issue_pair/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.cu
---unwind 5 --force-malloc-success --state-hashing --context-bound 2
+--unwind 5 --force-malloc-success --state-hashing --context-bound 2 --incremental-bmc
 
 ^VERIFICATION SUCCESSFUL$
--- a/regression/cuda/Supported_long_time/032_return_value/test.desc
+++ b/regression/cuda/Supported_long_time/032_return_value/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.cu
---context-bound 2 --force-malloc-success --unwind 9 --state-hashing
+--context-bound 2 --force-malloc-success --unwind 9 --state-hashing --incremental-bmc
 
 ^VERIFICATION SUCCESSFUL$
--- a/regression/cuda/Supported_long_time/035_test2/test.desc
+++ b/regression/cuda/Supported_long_time/035_test2/test.desc
@@ -1,6 +1,6 @@
 CORE
 main.cu
---force-malloc-success --unwind 9 --state-hashing
+--force-malloc-success --unwind 9 --state-hashing --context-bound 2
 
 ^VERIFICATION SUCCESSFUL$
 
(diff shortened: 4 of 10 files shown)

### CODE REVIEW

**Comment by XLiZHI** [comment]
This PR:
https://github.com/esbmc/esbmc/actions/runs/10111382931

Master:
https://github.com/esbmc/esbmc/actions/runs/10111390819

**Comment by mikhailramalho** [comment]
Could you explain it better this patch? Maybe our script is incorrect?

The no-assertions flags should do what it says: all assertions in the code are ignored, we only check the properties we introduce ourselves.

**Comment by XLiZHI** [comment]
Hi @mikhailramalho, please see this test case #1953, the variable in `__VERIFIER_assert `needs to be checked in this example.

In this patch, only the assertion added in the race check is executed, `__VIERIFIER_assert` will not be executed.

**Comment by rafaelsamenezes** [comment]
> Hi @mikhailramalho, please see this test case #1953, the variable in `__VERIFIER_assert `needs to be checked in this example.

Why are we passing --no-assertions If the user specified assertions are important for the analysis?

**Comment by lucasccordeiro** [comment]
> Why are we passing --no-assertions If the user specified assertions are important for the analysis?

Our SV-COMP script introduces this `` --no-assertions`.

The real question is: Why do we introduce this flag for SV-COMP?

**Comment by lucasccordeiro** [comment]
A git blame indicates that @fbrausse introduced this flag:

image


Maybe we should remove this https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/esbmc-wrapper.py#L293?

**Comment by XLiZHI** [comment]
> > Why are we passing --no-assertions If the user specified assertions are important for the analysis?
> 
> Our SV-COMP script introduces this `` --no-assertions`.
> 
> The real question is: Why do we introduce this flag for SV-COMP?

I believe the reason for adding this flag is for performance, if there are only two situations in the data race check: there is a race or there is no race, then I think we can keep the flag and we just check for read and write.

**Comment by mikhailramalho** [comment]
There are programs that have assertions but we want to check memory
properties. We should not check assertions in the memory safety category.

**Comment by mikhailramalho** [comment]
This should be handled like the memory properties, no code assertions
should be checked when checking for data races.

**Comment by fbrausse** [comment]
> A git blame indicates that @fbrausse introduced this flag:

Well, to be fair, I "introduced" the entire case for data race checks, not just this flag. But see below.

> There are programs that have assertions but we want to check memory properties. We should not check assertions in the memory safety category.

Yes, exactly this. If there are doubts whether the expressions evaluated in the context of `__VERIFIER_assert` are eligible for data races, we should discuss at the SV-COMP meeting. But the assertions themselves shouldn't lead to VERIFICATION FAILED - at least when `--no-assertions` is passed.

**Comment by XLiZHI** [comment]
This PR:

```
  correct:             364
    correct true:      318
    correct false:      46
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:             393
  Score:               106 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/10132108196/job/28016059913
```

Master:

```
  correct:             364
    correct true:      319
    correct false:      45
  incorrect:            31
    incorrect true:     11
    incorrect false:    20
  unknown:             388
  Score:                11 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/10111390819/job/27964665923
```

File: src/goto-symex/symex_main.cpp [hunk 1]
```diff
@@ -319,6 +319,12 @@ void goto_symext::symex_step(reachability_treet &art)
         if (run_builtin(call, id.as_string()))
           return;
       }
+      else if (
+        has_prefix(id.as_string(), "c:@F@__VERIFIER_assert") && no_assertions)
+      {
+        cur_state->source.pc++;
+        return;
+      }
```
mikhailramalho: Is this needed?

XLiZHI: I think the `-no-assertions` flag means the assertions don't need to be executed, does it?

mikhailramalho: IIRC the no assertion flag is checked when creating the assert ssa, so this should not be needed

src/goto-programs/rw_set.cpp
```diff
@@ -30,6 +30,18 @@ void rw_sett::compute(const exprt &expr)
       assert(code.operands().size() == 1);
       read_rec(code.op0());
     }
+    else if (statement == "function_call")
+    {
+      assert(code.operands().size());
```
mikhailramalho, line 35: We shouldn't restrict ourselves to only check verifier_assert, here's how cbmc handles it: https://github.com/diffblue/cbmc/blob/9dbbada7fa638b49c80242fe22f563933ca57d70/src/goto-instrument/rw_set.cpp#L57

mikhailramalho (changes requested):

lucasccordeiro:
@XLiZHI: I agree with @mikhailramalho's comments.

Additionally, we should refactor this module since our `compute` method adds too many if-conditions to check for particular cases.

Finally, in addition to `goto` and `assert`, we should consider the instrumentation for `assume` statements at https://github.com/esbmc/esbmc/blob/master/src/goto-programs/rw_set.cpp#L34.

XLiZHI:
This PR:
```
  correct:             366
    correct true:      320
    correct false:      46
  incorrect:            23
    incorrect true:     10
    incorrect false:    13
  unknown:             394
  Score:               158 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/10142643100/job/28043045255
```
Master:
```
  correct:             364
    correct true:      319
    correct false:      45
  incorrect:            31
    incorrect true:     11
    incorrect false:    20
  unknown:             388
  Score:                11 (max: 1488)

https://github.com/esbmc/esbmc/actions/runs/10111390819/job/27964665923
```

lucasccordeiro (approved):
Excellent, @XLiZHI.

Thanks for submitting this PR!

lucasccordeiro:
@XLiZHI: BTW, are the remaining incorrect true due to the context-bound?

XLiZHI:
> @XLiZHI: BTW, are the remaining incorrect true due to the context-bound?

Yes, I found an incorrect one, and if we change it to context bound 3, we get the correct result. I will try to use a different flag in the next PR to see if the score improves.

lucasccordeiro:
Many thanks, @XLiZHI.

Once @mikhailramalho is happy with this PR, I'll merge it.

XLiZHI:
CI failed because there were performance issues and many CUDA TCs timed out.  I recommend that we temporarily disable the C++ race check test cases until the performance issues are resolved, as they currently have too much thread interleaving.

lucasccordeiro:
> CI failed because there were performance issues and many CUDA TCs timed out. I recommend that we temporarily disable the C++ race check test cases until the performance issues are resolved, as they currently have too much thread interleaving.

I'm fine with that.

Can you open a ticket to track this performance issue?

lucasccordeiro:
@XLiZHI: Our symbolic execution has some hardcoded strings for C only:

https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L255
https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L836
https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L837
https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L838
https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L840
https://github.com/esbmc/esbmc/blob/master/src/goto-symex/execution_state.cpp#L841

We should fix that.

mikhailramalho (approved):
LGTM

lucasccordeiro:

XLiZHI:
Here is result using --context-bounds 3:
```
  correct:             149
    correct true:       98
    correct false:      51
  incorrect:            18
    incorrect true:     10
    incorrect false:     8
  unknown:             616
  Score:              -201 (max: 1488)
  
  https://github.com/esbmc/esbmc/actions/runs/10160181608/job/28096796250
```

lucasccordeiro:
@XLiZHI: Thanks for providing these results. 

With context-bound 3, we can reduce the number of incorrect results but also the number of correct results :-(.

PR #1956, opened 2024-07-31
[Solidity] code refactoring, bug fixed for inheritance and more

ChenfengWei0:


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -531,7 +621,30 @@ bool solidity_convertert::get_var_decl(
       return true;
   }
 
-  bool is_state_var = ast_node["stateVariable"] == true;
+  if (
+    t.get("#sol_type") == "CONTRACT" &&
+    t.identifier().as_string() == prefix + current_contractName)
+  {
+    // contract Base{
+    //   Base x;
```
intrigus-lgtm, line 568: How is this implemented/why does it work in Solidity?

ChenfengWei0:
```solidity
contract Base
{
    Base x;
    uint public y;
    constructor()
    {
        y = 1;
    }

    function test() public   {
        x = Base(0x9cF86D8D08bC34248210474C4B019befb0fE70fA);
        assert(x.y() == 1); 
    }
}
```
- We first deploy this contract, gather the address 
- We initialize x with the contract address and deploy it again.
- Run function `test` and the assertion holds.

src/pointer-analysis/dereference.cpp
```diff
@@ -908,10 +908,27 @@ void dereferencet::build_reference_rec(
...
+    */
+    std::string sol;
+    options.get_option("sol", sol);
+    if (sol.empty())
+    {
+      log_error(
+        "Can't construct rvalue reference to array type during dereference\n"
+        "(It isn't allowed by C anyway)\n");
+      abort();
+    }
```
fbrausse: This fixes nothing.

fbrausse (changes requested):

src/pointer-analysis/dereference.cpp
ChenfengWei0: It just avoid the abort. The result is correct.

fbrausse: What is the result?

ChenfengWei0: E.g.

```solidity
contract test {
    uint[3] public t = [0, 0]; // t = [0,0,0]

    constructor() {
        (t, t) = ([1], [2]);  // t = [2,0,0]; t=[1,0,0]
    }
}

contract test1 is test {
    function test2() public view {
        assert(t[0] == 1);
        assert(t[1] == 0);
        assert(t[2] == 0);
    }
}
```
- Command: `--unwind 10 --no-unwinding-assertions`
- Result: `VERIFICATION SUCCESSFUL`

I am not sure what's going on in the middleware & backend. 
E.g.
```solidity
 int x[3] = [1, 2, 3];
 constructor(){
     x = [4, 5, 6];
 }
```
I guess the assignment will just overwrite the x's value. The corresponding symbol table is:
```
Symbol......: sol:@C@test@F@test#19
Module......: example
Base name...: test
Mode........: C++
Type........: constructor  (struct test *)
Value.......:
{
&(this->t = { 1, 2, 3 })[0];

&(this->t = { 4, 5, 6 })[0];

}
Flags.......: lvalue
Location....: file example.sol line 4
```

fbrausse: Thanks for the examples, they're helpful. The reason the dereference module doesn't handle array dereferences is because there are no C semantics for it - since C doesn't have that construct: A dereference in C always refers to either a built-in type like `int` or to a non-array struct or union (member). For the C++ frontend we got around that problem in two ways:
1. array-typed statements: https://github.com/esbmc/esbmc/blob/7b1e7f2d750f8d579db145a118338a9c52b81347/src/clang-c-frontend/clang_c_adjust_code.cpp#L45-L60
2. during a constructor call: https://github.com/esbmc/esbmc/blob/7b1e7f2d750f8d579db145a118338a9c52b81347/src/clang-cpp-frontend/clang_cpp_convert.cpp#L1184-L1190

Would that be an option for you with Solidity?

ChenfengWei0: Could they support the dynamic array? Like:
```solidity
unit[] x;

constructor()
{
  x = [1];
}

function test() public
{
  x = [1, 2, 3];
  x = [4, 5];
}
```

---

I can make the fixed array to pointer type in the front end, like the dynamic array. E.g. 
```
Symbol......: sol:@C@Base@x#3
Module......: example
Base name...: x
Mode........: C++
Type........: unsigned _ExtInt(256) *
Value.......: 0
Flags.......: lvalue file_local
Location....: file example.sol line 2

Symbol......: sol:@C@Base@F@test#20
Module......: example
Base name...: test
Mode........: C++
Type........: void (struct Base *)
Value.......:
{
this->x = &{ 1, 2, 3 }[0];

this->x = &{ 4, 5 }[0];

}
Flags.......: lvalue
Location....: file example.sol line 4
```
But I don't feel like changing the code for the fixed array as they already 'work' somehow.

ChenfengWei0:
This PR introduces a major refactoring, shifting from a C-based implementation to a C++-based one. This transition enables better support for complex structures such as inheritance (the previous version was somewhat broken), contract instances (which is virtually identical to class instances), and more. Additionally, it includes several bug fixes and test improvements.

- Major refactoring: Migrated from C-based to C++-based architecture to support advanced features like inheritance and contract instances.
- Bug fixes: Addressed issues in features including constructors, dynamic arrays, tuples, and mappings.
- Test improvements: Fixed or removed some regression tests to align with the new implementation.

lucasccordeiro (approved):

lucasccordeiro:

PR #1957, opened 2024-07-31
Python frontend: Handling single type lists

brcfarias:
This PR adds support for single type lists to be handled as arrays.
The following types of lists are now supported:

```
l1 = [1,2,3]          # Integer list
l2 = [1.2, 2.1]       # Float list
l3 = [True, False]    # Boolean list
l4 = ["abc", "def"]   # String list
```

Multi-type lists are not yet supported.


lucasccordeiro (approved):
Well done, @brcfarias!

PR #1903, opened 2024-06-25
[Concurrency] consider pointer when analyzing reads and writes for global expression

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1889.

When analyzing global expression reads and writes, we should consider pointers. Identifying whether a pointer references only local expressions can help reduce the number of interleavings. However, paying attention to pointers when analyzing reads and writes for global expressions is crucial.

Issue #1889, opened 2024-06-20
[Concurrency] inconsistent results after pointer dereference


Anthonysdu:
```c
#include<pthread.h>
#include<assert.h>

typedef struct SafeStackItem {
  volatile int Value;
  int Next;
} SafeStackItem;
typedef struct SafeStack {
  SafeStackItem array[3];
  int head;
  int count;
} SafeStack;
pthread_t threads[3];
SafeStack stack;

int exchange(int *obj, int v) {
  int t = *obj;
  *obj = v;
  return t;
}

int Pop(void) {
  int head1 = stack.head;
  //int next1 = stack.array[head1].Next;
  //stack.array[head1].Next = -1;
  int next1 = exchange(&stack.array[head1].Next, -1);
  if (next1 >= 0)
    return head1;
  return -1;
}
void *thread(void *arg) {
  int idx = (int)(size_t)arg;
    int elem;
    while (1) {
      elem = Pop();
      if (elem >= 0)
        break;
    }
    stack.array[elem].Value = idx;
    if (!(stack.array[elem].Value == idx)) {
      assert(0);
    }
  return ((void *)0);
}
int main(void) {
  pthread_create(&threads[0], ((void *)0), thread, (void *)0);
  pthread_create(&threads[1], ((void *)0), thread, (void *)1);
  return 0;
}
```
ESBMC reports `VERIFICATION SUCCESSFUL` by `esbmc main.c --unwind 1 --no-unwinding-assertions --no-por`

CBMC:  `cbmc main.c --unwind 1`: `line 41 assertion 0: FAILURE`

Dartagnan: 
```
===== Safety violation found =====
	E49:	@main.c#41: user assertion
=================================================
FAIL
```

It is worth noting that if replace the line 

```c
int next1 = exchange(&stack.array[head1].Next, -1);
```
with the  functionally equivalent code in comments above

```c
  int next1 = stack.array[head1].Next;
  stack.array[head1].Next = -1;
```
Then esbmc
```
Violated property:
  file main.c line 41 column 7 function thread
  assertion 0
  0


VERIFICATION FAILED
```

lucasccordeiro:
@Anthonysdu: This is a multi-threaded C program. We might be missing the interleaving to expose this assertion failure. You're our expert in the concurrency model :-)

Can you please investigate this issue?

lucasccordeiro:
We are missing some context switches. If we modify the `exchange` function as:

````C
int exchange(int *obj, int v) {
[...]

PR #1958, opened 2024-07-31
[Concurrency] only consider pointer to global when analyzing reads and writes

XLiZHI:
Support for pointers was added in the previous PR, and thread interleaving is added when reading or writing pointers, which results in very large thread interleaving when verifying multiple threads.

I think we should add some filtering to reduce thread interleaving, at least not unconditionally checking any pointers.

In this PR, I want to discuss the idea that there might be a better solution.

#1889
#1903 

_Update:_
This PR:
```
  correct:           14147
    correct true:     7761
    correct false:    6386
  incorrect:            29
    incorrect true:     12
    incorrect false:    17
  unknown:            9629
  Score:             21252 (max: 38482)
  
  https://github.com/esbmc/esbmc/actions/runs/10339896288/job/28620094507
```
Master:
```
  correct:           14100
    correct true:     7729
    correct false:    6371
  incorrect:            28
    incorrect true:     12
    incorrect false:    16
  unknown:            9677
  Score:             21189 (max: 38482)

  https://github.com/esbmc/esbmc/actions/runs/10339894034/job/28620090531
```

The incorrect one from no-data-race, TIMEOUT in master

### DIFF
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -843,9 +843,28 @@ void execution_statet::get_expr_globals(
     {
       return;
     }
+
+    bool point_to_global = false;
+    if (symbol->type.is_pointer() && symbol->name != "invalid_object")
+    {
+      expr2tc tmp = expr;
+      cur_state->rename(tmp);
+
+      std::unordered_set<expr2tc, irep2_hash> symbols;
+      get_symbols(tmp, symbols);
+      for (const auto &symbol_expr : symbols)
+      {
+        const std::string &n = to_symbol2t(symbol_expr).thename.as_string();
+        const symbolt *s = ns.lookup(n);
+        if (!s)
+          return;
+        point_to_global = s->static_lifetime;
+      }
+    }
+
     if (
-      (symbol->static_lifetime || symbol->type.is_dynamic_set()) ||
-      symbol->type.is_pointer())
+      symbol->static_lifetime || symbol->type.is_dynamic_set() ||
+      point_to_global)
     {
       std::list<unsigned int> threadId_list;
       auto it_find = art1->vars_map.find(expr);
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -845,20 +845,32 @@ void execution_statet::get_expr_globals(
     }
 
     bool point_to_global = false;
-    if (symbol->type.is_pointer() && symbol->name != "invalid_object")
+    if (
+      symbol->type.is_pointer() && symbol->name != "invalid_object" &&
+      !symbol->static_lifetime)
     {
       expr2tc tmp = expr;
       cur_state->rename(tmp);
+      expr2tc deref = dereference2tc(to_pointer_type(tmp->type).subtype, tmp);
+      value_setst::valuest dest;
+      cur_state->value_set.get_reference_set(deref, dest);
 
-      std::unordered_set<expr2tc, irep2_hash> symbols;
-      get_symbols(tmp, symbols);
-      for (const auto &symbol_expr : symbols)
+      for (const auto &obj : dest)
       {
-        const std::string &n = to_symbol2t(symbol_expr).thename.as_string();
-        const symbolt *s = ns.lookup(n);
-        if (!s)
-          return;
-        point_to_global = s->static_lifetime;
+        if (
+          is_object_descriptor2t(obj) &&
+          is_symbol2t(to_object_descriptor2t(obj).object))
+        {
+          const std::string &n =
+            to_symbol2t(to_object_descriptor2t(obj).object).thename.as_string();
+          const symbolt *s = ns.lookup(n);
+          if (!s)
+            continue;
+          point_to_global = s->static_lifetime || s->type.is_dynamic_set();
+
+          if (point_to_global)
+            break;
+        }
       }
     }
 
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -850,7 +850,9 @@ void execution_statet::get_expr_globals(
       !symbol->static_lifetime)
     {
       expr2tc tmp = expr;
+      /* Rename it so that it can be dereferenced in current state */
       cur_state->rename(tmp);
+      /* Collect all the objects pointed to by the pointer */
       expr2tc deref = dereference2tc(to_pointer_type(tmp->type).subtype, tmp);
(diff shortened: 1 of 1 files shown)


XLiZHI:
```
#include <pthread.h>

void* t1(void *arg) {
    int* shared1 = (int*) arg;

    return NULL;
}

void* t2(void *arg) {
    int* shared2 = (int*) arg;

    return NULL;
}

int main() {
    pthread_t id1, id2;
    int shared_data = 0;

    pthread_join(id1, NULL);
    pthread_join(id2, NULL);

    return 0;
}
```

master: 28199 interleavings
This PR: 221 interleavings

src/goto-symex/execution_state.cpp
```diff
@@ -843,9 +843,28 @@ void execution_statet::get_expr_globals(
...
     }
+
+    bool point_to_global = false;
+    if (symbol->type.is_pointer() && symbol->name != "invalid_object")
+    {
+      expr2tc tmp = expr;
+      cur_state->rename(tmp);
+
+      std::unordered_set<expr2tc, irep2_hash> symbols;
+      get_symbols(tmp, symbols);
```
mikhailramalho: does this return all objects a pointer can point to?

src/goto-symex/execution_state.cpp
```diff
@@ -843,9 +843,28 @@ void execution_statet::get_expr_globals(
...
+      for (const auto &symbol_expr : symbols)
+      {
+        const std::string &n = to_symbol2t(symbol_expr).thename.as_string();
+        const symbolt *s = ns.lookup(n);
+        if (!s)
+          return;
+        point_to_global = s->static_lifetime;
```
mikhailramalho: We can leave the loop once we find at least one global object

mikhailramalho: shouldn't this be `symbol->static_lifetime || symbol->type.is_dynamic_set()` like the condition below?

src/goto-symex/execution_state.cpp
XLiZHI: `rename` replaces a symbol expression with the last known expression
```
int global;
void func(int *a){*a = 1; } 
int main(){ func(&global); }

```
In this example, the expr for `a` will be replaced by the expr for `global`, than we use `get_symbols()` to get the symbol expr in the result.

mikhailramalho:
Sometimes we can determine where a pointer is pointing to, e.g.:

int global;
void func(int *a){*a = 1; }
int main(){
  int local;
  func(nondet()? &local : &global);
}
Does get symbol return all symbols a pointer can point to?

XLiZHI:
```c
#include<pthread.h>
#include<assert.h>

typedef struct SafeStackItem {
  volatile int Value;
  int Next;
} SafeStackItem;
typedef struct SafeStack {
  SafeStackItem array[3];
  int head;
  int count;
} SafeStack;
pthread_t threads[3];
SafeStack stack;

int exchange(int *obj, int v) {
  int t = *obj;
  *obj = v;
  return t;
}

int Pop(void) {
  int next;
  int head1 = stack.head;
  int next1 = exchange(nondet_bool()?&next : &stack.array[head1].Next, -1);
  if (next1 >= 0)
    return head1;
  return -1;
}
void *thread(void *arg) {
  int idx = (int)(size_t)arg;
    int elem;
    while (1) {
      elem = Pop();
      if (elem >= 0)
        break;
    }
    stack.array[elem].Value = idx;
    if (!(stack.array[elem].Value == idx)) {
      assert(0);
    }
  return ((void *)0);
}
int main(void) {
  pthread_create(&threads[0], ((void *)0), thread, (void *)0);
  pthread_create(&threads[1], ((void *)0), thread, (void *)1);
  return 0;
}
```

it works in this case:
```c
State 12 file main4.c line 25 column 3 function Pop thread 2
  next1 = 0 (00000000 00000000 00000000 00000000)

State 13 file main4.c line 38 column 5 function thread thread 2
  stack.array[(signed long int)elem].Value = (signed int)((unsigned long int)((void *)1))

State 15 file main4.c line 40 column 7 function thread thread 1
Violated property:
  file main4.c line 40 column 7 function thread
  assertion 0
  0


VERIFICATION FAILED
```

lucasccordeiro:
@XLiZHI: I agree with @mikhailramalho's comments.

The crucial part is checking the set of objects a given pointer can point to.

lucasccordeiro:
@XLiZHI: this method can provide the set of objects that a pointer might point to: 

https://github.com/esbmc/esbmc/blob/master/src/goto-symex/symex_dereference.cpp#L53

XLiZHI:
```diff
diff --git a/src/goto-symex/execution_state.cpp b/src/goto-symex/execution_state.cpp
index 3f5f828df..fbf7a378d 100644
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -843,9 +843,35 @@ void execution_statet::get_expr_globals(
     {
       return;
     }
+
+    bool point_to_global = false;
+    if (symbol->type.is_pointer() && symbol->name != "invalid_object")
+    {
+      expr2tc deref = dereference2tc(to_pointer_type(expr->type).subtype, expr);
+      value_setst::valuest dest;
+      cur_state->value_set.get_reference_set(deref, dest);
+
+      for (const auto &obj : dest)
+      {
+        if (
+          is_object_descriptor2t(obj) &&
+          is_symbol2t(to_object_descriptor2t(obj).object))
+        {
+          const std::string &n =
+            to_symbol2t(to_object_descriptor2t(obj).object).thename.as_string();
+          const symbolt *s = ns.lookup(n);
+          if (!s)
+            continue;
+          point_to_global = s->static_lifetime || s->type.is_dynamic_set();
+
+          if (point_to_global)
+            break;
+        }
+      }
+    }
+
     if (
-      (symbol->static_lifetime || symbol->type.is_dynamic_set()) ||
-      symbol->type.is_pointer())
+      symbol->static_lifetime || symbol->type.is_dynamic_set())
     {
       std::list<unsigned int> threadId_list;
       auto it_find = art1->vars_map.find(expr);
```

Here's my new solution, but it doesn't work. I looked into the reason, and it was because of failed dereferences.

`FUNCTION_CALL:  return_value$_exchange$1=exchange(&stack.array[(signed long int)head1].Next, -1)`

we assign `&stack.array[(signed long int)head1].Next` to `obj` when call this functions, but it failed when we try to dereference after assignment.

lucasccordeiro (approved):

PR #1963, opened 2024-08-05
[python-frontend] mutated regression test to identify known bugs

haru-02:
While trying to perform mutation test on ESBMC with the regression test sets, a few of the test cases helped identify a few bugs in ESBMC. I will make a few issue posts and link it here to the pull request. shortly. Requesting to add this test to a new branch if possible.

### DIFF
--- a/regression/python/built-in-functions/main.py
+++ b/regression/python/built-in-functions/main.py
@@ -1,5 +1,5 @@
 def foo() -> None:
-    a = int(2)
+    a = int(2.1212121212)
     assert a == 2
 
 foo()
--- a/regression/python/compound-assign/main.py
+++ b/regression/python/compound-assign/main.py
@@ -1,5 +1,13 @@
-count : int = 1
+count : int = 2
 count *= 2
 count /= 2
 count += 1
 count -= 1
+count //= 1
+count **= 1
+count %= 1
+count >>= 1
+count <<= 1
+count &= 1
+count |= 1
+count ^= 1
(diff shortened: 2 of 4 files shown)


haru-02:
Please refer issues #1966 and #1964 for further info and discussion on the specific bug. Thank you.

lucasccordeiro:
@haru-02: can I ask you whether you still have plans to move forward with this PR?

haru-02:
> @haru-02: can I ask you whether you still have plans to move forward with this PR?

Thank you for reaching out @lucasccordeiro . The original goal of this PR was to document and discuss ESBMC’s behaviour for these test cases. During my dissertation, I worked on a few issues related to the mutation tests. I do intend to contribute further to ESBMC in my free time, beyond the scope of the dissertation, where possible.

pmatos:
Thanks for putting these together as part of your mutation-testing work.

Revisiting these mutations against current `master`:

- The primary bug they targeted — `int(float)` truncation (#1964) — has since been fixed, and #1964 is closed. The `built-in-functions` test now covers a superset of these `int(...)` scenarios.
- The `int_bit_length` case turns out to be an unwinding-bound artifact rather than a correctness bug: `int(2**31).bit_length() == 32` verifies successfully once the unwind bound reaches the bit width (e.g. `--unwind 33`).
- The `compound-assign` mutation now trips a type-annotation error rather than demonstrating a verification gap.

Since these tests were intended to document bugs that no longer reproduce, and CORE regression tests can't intentionally fail on `master`, I'm closing this as obsolete.

(Aside: the `int(2**64).bit_length()` case from #1964 does still reproduce a genuine overflow issue on current `master`; that's a separate item worth tracking on its own, and this PR doesn't exercise it.) Thanks again for the contribution.

PR #1968, opened 2024-08-07
[C++ verification] fixed array type init

### LINKED ISSUE #1962: [c++] Can't construct rvalue reference to array type
```cpp
class Base
{
public:
  int ss[1] = {1};
  int sss[1] = {1};
};

int main()
{
  Base x;
}
```

Result
```
Parsing 1.cpp
Converting
Generating GOTO Program
GOTO program creation time: 0.470s
Checking base case, k = 1
(It isn't allowed by C anyway
```

XLiZHI:
Here's the solution I came up with

```cpp
#include <cassert>

class Base
{
public:
  int ss[1];
  int sss[1];
  Base(): ss{1}, sss{1} {}

};

int main()
{
  Base x;
  assert(x.ss[0] == 1);
  assert(x.sss[0] == 1);
}
```

Goto:
```cpp
Base (c:@S@Base@F@Base#):
        // 20 no location
        DECL Base array_init$;
        // 21 no location
        ASSIGN array_init$=NONDET(struct);
        // 22 
        ASSIGN array_init$.ss={ 1 };
        // 23 
        OTHER &array_init$.ss[0];     ---> Where did this come from

        // 24 
        ASSIGN *this=array_init$;
        // 25 
        ASSIGN array_init$.sss={ 1 };
        // 26 
        OTHER &array_init$.sss[0];   ---> Where did this come from

        // 27 
        ASSIGN *this=array_init$;
        // 28 file main1.cpp line 8 column 26 function Base
        DEAD c:@S@Base@F@Base#this_array_init$
        // 29 file main1.cpp line 8 column 26 function Base
        END_FUNCTION // Base
```
But I'm a little confused there are some extra instructions, I'm not sure where they came from, but I'm pretty sure this PR didn't cause that problem.

### DIFF
--- a/src/clang-c-frontend/clang_c_adjust_code.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_code.cpp
@@ -57,7 +57,7 @@ void clang_c_adjust::adjust_code(codet &code)
        * unused.
        */
       exprt &op = code.op0();
-      if (op.op0().type().is_pointer())
+      if (op.has_operands() && op.op0().type().is_pointer())
         op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
     }
     adjust_operands(code);
--- a/src/clang-c-frontend/clang_c_adjust_code.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_code.cpp
@@ -57,7 +57,7 @@ void clang_c_adjust::adjust_code(codet &code)
        * unused.
        */
       exprt &op = code.op0();
-      if (op.has_operands() && op.op0().type().is_pointer())
+      if (op.statement() != "assign")
         op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
     }
     adjust_operands(code);
--- a/src/clang-c-frontend/clang_c_adjust_code.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_code.cpp
@@ -57,7 +57,8 @@ void clang_c_adjust::adjust_code(codet &code)
        * unused.
        */
       exprt &op = code.op0();
-      op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
+      if (op.op0().type().is_pointer())
+        op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
     }
     adjust_operands(code);
   }
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -1257,7 +1257,6 @@ bool clang_cpp_convertert::get_function_body(
         else
         {
           initializer = side_effect_exprt("assign", member.type());
-          initializer.set("#member_init", 1);
           initializer.move_to_operands(member, rhs);
           initializer.location() = new_expr.location();
           initializers.push_back(initializer);
@@ -1269,15 +1268,10 @@ bool clang_cpp_convertert::get_function_body(
         log_error("Unsupported initializer in {}", __func__);
         abort();
       }
-
-      // Convert to code and insert side-effect in the operands list
-      // Essentially we convert an initializer to assignment, e.g:
-      // t1() : i(2){ }
-      // is converted to
-      // t1() { this->i = 2; }
-      for (exprt &initializer : initializers)
-        convert_expression_to_code(initializer);
     }
+
+    for (exprt &initializer : initializers)
+      convert_expression_to_code(initializer);
 
     // Insert initializers at the beginning of the body
     body.operands().insert(
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -1129,8 +1129,13 @@ bool clang_cpp_convertert::get_function_body(
     exprt::operandst initializers;
     initializers.reserve(cxxcd.getNumCtorInitializers());
 
-    // Parse the initializers, if any
+    /* State to track initialization of array member initializers. */
+    symbolt *array_init_sym = nullptr; /* temp symbol for array init */
+    /* track whether the temp symbol is up to date. */
+    bool init_sym_uptodate = true;
 
+    // Parse the initializers, if any
+ 
     // `init` type is clang::CXXCtorInitializer
     for (auto init : cxxcd.inits())
     {
@@ -1153,6 +1158,8 @@ bool clang_cpp_convertert::get_function_body(
         initializer.set("#delegating_ctor", 1);
         if (get_expr(*init->getInit(), initializer))
           return true;
+        initializers.push_back(initializer);
+        init_sym_uptodate = false;
       }
       else if (init->isBaseInitializer())
       {
@@ -1162,6 +1169,8 @@ bool clang_cpp_convertert::get_function_body(
         initializer.base_ctor_derived(true);
         if (get_expr(*init->getInit(), initializer))
           return true;
+        initializers.push_back(initializer);
+        init_sym_uptodate = false;
       }
       else if (init->isMemberInitializer())
       {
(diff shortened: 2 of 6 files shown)


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1191,27 +1192,29 @@ bool clang_cpp_convertert::get_function_body(
...
+
+          symbolt new_symbol;
+
+          new_symbol.name = "array_init$";
+          new_symbol.id = id2string(this_ptr.identifier()) + "_array_init$";
+          new_symbol.type = this_type;
+
+          symbolt &init_sym = *context.move_symbol_to_context(new_symbol);
+
+          init_decl = code_declt(symbol_expr(init_sym));
```
fbrausse: I think these new instructions might come from this introduction of a new symbol for each array member. I don't think we need that: symex will take care of renaming the existing symbol to something new. In essence, I think, we just need one new symbol `x` for each constructor call, representing an object of the constructed type - independent of how many array members it has.

The sequence would then be:
```
DECL x // <-- this is the single symbol for this constructor of Base
x.ss = {1} // <-- reuse symbol
*this = x
x.sss = {1} // <-- reuse symbol
*this = x
```

fbrausse: That's not it, I just checked. Assignments of array members `x.y = z` are side_effect_exprs of array type, so my earlier change in the C-adjuster converts those to `&(x.y = z)[0]`. Somewhere in the preprocessing is that being pulled apart into two statements `x.y = z` followed by another `&(x.y)[0]` statement.

fbrausse:
I've had a bit trouble following the code because you're creating a new symbol for each array. However, it is not new as it always has the same id derived from the id of the `this` pointer, i.e., the constructor's id. While I now believe it should work, I've created the following patch that is easier for me to follow. Feel free to apply it in full or in part.
<details>

```diff
diff --git a/src/clang-cpp-frontend/clang_cpp_convert.cpp b/src/clang-cpp-frontend/clang_cpp_convert.cpp
index e2edf4b8d4..98efcc0988 100644
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -1129,9 +1129,13 @@ bool clang_cpp_convertert::get_function_body(
     exprt::operandst initializers;
     initializers.reserve(cxxcd.getNumCtorInitializers());
 
+    /* State to track initialization of array member initializers. */
+    symbolt *array_init_sym = nullptr; /* temp symbol for array init */
+    /* track whether the temp symbol is up to date. */
+    bool init_sym_uptodate = true;
+
     // Parse the initializers, if any
-    code_declt init_decl;
-    bool need_array_init = false;
+
     // `init` type is clang::CXXCtorInitializer
     for (auto init : cxxcd.inits())
     {
@@ -1154,6 +1158,8 @@ bool clang_cpp_convertert::get_function_body(
         initializer.set("#delegating_ctor", 1);
         if (get_expr(*init->getInit(), initializer))
           return true;
+        initializers.push_back(initializer);
+        init_sym_uptodate = false;
       }
       else if (init->isBaseInitializer())
       {
@@ -1163,6 +1169,8 @@ bool clang_cpp_convertert::get_function_body(
         initializer.base_ctor_derived(true);
         if (get_expr(*init->getInit(), initializer))
           return true;
+        initializers.push_back(initializer);
+        init_sym_uptodate = false;
       }
       else if (init->isMemberInitializer())
       {
@@ -1185,44 +1193,77 @@ bool clang_cpp_convertert::get_function_body(
         /* We can't assign to arrays, dereference() will choke. */
         if (is_array_like(member.type()))
         {
-          /* Instead, create a constant expression of the class's type, where
-           * each component, except for the one referred to in 'lhs', just is a
-           * member expression into (*this). Then assign this new expression
-           * to (*this). */
+          /* Instead, introduce a single new symbol 'array_init$' of the class's
+           * type, and encode the following sequence:
+           *
+           *  0. array_init$ = *this;
+           *  1. array_init$.member = rhs;
+           *  2. *this = array_init$;
+           *
+           * Where possible, we leave out instruction 0, i.e., if the previous
+           * initializer already was an array and at the very beginning. This is
+           * tracked via the boolean 'init_sym_uptodate'.
+           *
+           * Step 2. is necessary after each member, because the next
+           * initializer can legally expect previous members obtained via the
+           * (this) pointer to have already been initialized. */
+
           const exprt &this_ptr = member.op0(); /* the (this) pointer */
           const struct_union_typet &this_type =
             to_struct_union_type(ns.follow(this_ptr.type().subtype()));
 
-          need_array_init = true;
-
-          symbolt new_symbol;
-
-          new_symbol.name = "array_init$";
-          new_symbol.id = id2string(this_ptr.identifier()) + "_array_init$";
-          new_symbol.type = this_type;
-
-          symbolt &init_sym = *context.move_symbol_to_context(new_symbol);
-
-          init_decl = code_declt(symbol_expr(init_sym));
+          if (!array_init_sym)
+          {
+            symbolt new_symbol;
+            new_symbol.name = "array_init$";
+            new_symbol.id = id2string(this_ptr.identifier()) + "_array_init$";
+            new_symbol.type = this_type;
+            if (context.move(new_symbol, array_init_sym))
+            {
+              log_error(
+                "Duplicate temporary array symbol for initializer in {}",
+                __func__);
+              abort();
+            }
+            assert(array_init_sym);
+          }
+
+          exprt this_sym = dereference_exprt(this_ptr, this_ptr.type());
+          exprt init_sym = symbol_expr(*array_init_sym);
+
+          if (!init_sym_uptodate)
+          {
+            /* 0. array_init$ = *this; */
+            side_effect_exprt update("assign", this_type);
+            update.copy_to_operands(init_sym, this_sym);
+            initializers.push_back(update);
+            init_sym_uptodate = true;
+          }
 
           exprt new_member = member;
-          new_member.op0() = symbol_expr(init_sym);
+          new_member.op0() = init_sym;
 
+          /* 1. array_init$.member = rhs; */
           initializer = side_effect_exprt("assign", new_member.type());
           initializer.move_to_operands(new_member, rhs);
-
-          convert_expression_to_code(initializer);
           initializers.push_back(initializer);
 
-          exprt rhs_sym = symbol_expr(init_sym);
-          exprt this_sym = dereference_exprt(this_ptr, this_type);
+          /* 2. *this = array_init$; */
           initializer = side_effect_exprt("assign", this_sym.type());
-          initializer.move_to_operands(this_sym, rhs_sym);
+          initializer.move_to_operands(this_sym, init_sym);
+          initializers.push_back(initializer);
         }
         else
         {
+          // Essentially we convert an initializer to assignment, e.g:
+          // t1() : i(2){ }
+          // is converted to
+          // t1() { this->i = 2; }
           initializer = side_effect_exprt("assign", member.type());
+          initializer.set("#member_init", 1);
           initializer.move_to_operands(member, rhs);
+          initializers.push_back(initializer);
+          init_sym_uptodate = false;
         }
       }
       else
@@ -1230,21 +1271,22 @@ bool clang_cpp_convertert::get_function_body(
         log_error("Unsupported initializer in {}", __func__);
         abort();
       }
+    }
 
-      // Convert to code and insert side-effect in the operands list
-      // Essentially we convert an initializer to assignment, e.g:
-      // t1() : i(2){ }
-      // is converted to
-      // t1() { this->i = 2; }
+    for (exprt &initializer : initializers)
       convert_expression_to_code(initializer);
-      initializers.push_back(initializer);
-    }
 
     // Insert initializers at the beginning of the body
     body.operands().insert(
       body.operands().begin(), initializers.begin(), initializers.end());
-    if (need_array_init)
+
+    if (array_init_sym)
+    {
+      /* Need to declare the temp symbol for array initialization if it has
+       * been used. */
+      code_declt init_decl(symbol_expr(*array_init_sym));
       body.operands().insert(body.operands().begin(), init_decl);
+    }
   }
 
   auto *type = fd.getType().getTypePtr();
```

</details>

fbrausse:
Shall we add the multiple cases from #1962? Also, I would suggest to add one or two cases of the `Base` example where non-array members are between the arrays, and/or at the beginning of the class.

XLiZHI:
Thank you! @fbrausse I found the reason for the extra instructions.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1233,13 +1275,20 @@ bool clang_cpp_convertert::get_function_body(
       // t1() : i(2){ }
       // is converted to
       // t1() { this->i = 2; }
-      convert_expression_to_code(initializer);
-      initializers.push_back(initializer);
+      for (exprt &initializer : initializers)
+        convert_expression_to_code(initializer);
```
fbrausse: This should be one level up, that is, after the loop.

src/clang-c-frontend/clang_c_adjust_code.cpp
```diff
@@ -57,7 +57,8 @@ void clang_c_adjust::adjust_code(codet &code)
        * unused.
        */
       exprt &op = code.op0();
-      op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
+      if (op.has_operands() && op.op0().type().is_pointer())
+        op = address_of_exprt(index_exprt(op, constant_exprt(0, index_type())));
```
fbrausse, line 61: Doing this only for dereferenced pointers is the wrong condition. The point of this transformation is to avoid array-typed top-level expressions in statements. It's better to test whether the statement is actually an assignment. Because in that case we don't need to do the conversion. E.g., when just testing for pointer type, the following variation of #1938 starts failing again (it's fine on current master):
```c
struct S { int ss[128]; };

int main()
{
	struct S s, *t = &s;
	(*t).ss;
}
```

fbrausse (changes requested):

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1233,13 +1275,20 @@ bool clang_cpp_convertert::get_function_body(
       // t1() : i(2){ }
       // is converted to
       // t1() { this->i = 2; }
```
fbrausse: I'd suggest to also move this comment into the `else` branch of the member-init, which it documents. Might be misleading otherwise.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1184,62 +1193,97 @@ bool clang_cpp_convertert::get_function_body(
...
           initializer = side_effect_exprt("assign", this_sym.type());
-          initializer.move_to_operands(this_sym, rhs_sym);
+          initializer.move_to_operands(this_sym, init_sym);
+          initializer.location() = new_expr.location();
+          initializers.push_back(initializer);
         }
         else
         {
           initializer = side_effect_exprt("assign", member.type());
+          initializer.set("#member_init", 1);
```
XLiZHI: Do we need it?

fbrausse: No, I don't think so. I forgot to remove it, sorry.

fbrausse (approved):
Thank you for implementing this! LGTM.

lucasccordeiro (approved):

PR #1972, opened 2024-08-13
intrinsic_memset: gen_byte_expression() uses compatible types for constants

### LINKED ISSUE #1969: [ESBMC-CHERI] Assertion `a->sort->get_data_width() == b->sort->get_data_width()' failed
While building ESBMC-CHERI (Whether hybrid or purecap), an error occurred as follows:
```cpp
/usr/include/c++/11/bits/stl_uninitialized.h:994: undefined reference to `__ubsan_handle_type_mismatch_v1'
/usr/bin/ld: /usr/include/c++/11/bits/stl_uninitialized.h:994: undefined reference to `__ubsan_handle_type_mismatch_v1'
/usr/bin/ld: /usr/include/c++/11/bits/stl_uninitialized.h:994: undefined reference to `__ubsan_handle_type_mismatch_v1'
/usr/bin/ld: /usr/include/c++/11/bits/stl_uninitialized.h:995: undefined reference to `__ubsan_handle_type_mismatch_v1'
/usr/bin/ld: _deps/catch2-build/src/libCatch2d.a(catch_reporter_multi.cpp.o):/usr/include/c++/11/bits/stl_uninitialized.h:995: more undefined references to `__ubsan_handle_type_mismatch_v1' follow
collect2: error: ld returned 1 exit status
```
I tried to add `-DCMAKE_BUILD_TYPE=Sanitizer -DSANITIZER_TYPE=UBSAN` in the past to build successfully, but then we couldn't build the debug type.

~~This error seems to come from `cheribsd`, is there a way we can avoid it?~~

_update:_ I found the reason, catch2 seems to be that we need to build when we use` -DBUILD_TESTING=On`, and cheribsd doesn't seem to have this `ubsan` library (even I used `--use-cheri-ubsan`  and `--use-cheri-ubsan-runtime` to cheribuild), so the current solution is `-DBUILD_TESTING=Off`

fbrausse:
Fixes #1969.

### DIFF
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -1042,8 +1042,8 @@ static inline expr2tc gen_byte_expression(
 
   expr2tc mask = gen_zero(type);
 
-  const auto eight = constant_int2tc(int_type2(), BigInt(8));
-  const auto one = constant_int2tc(int_type2(), BigInt(1));
+  const auto eight = constant_int2tc(type, BigInt(8));
+  const auto one = constant_int2tc(type, BigInt(1));
   for (unsigned i = 0; i < num_of_bytes; i++)
   {
     result = shl2tc(type, result, eight);


rafaelsamenezes (approved):
LGTM

XLiZHI (approved):
Thanks! @fbrausse

intrigus-lgtm:
Instead of this patch, did you try changing uint64_t here to size_type()?
I remember having an issue with a similar error message, but could also be that this is unrelated to your issue ^^
https://github.com/esbmc/esbmc/blob/9a25e56e07f8edf5cc45d4bcf77e7dab7a2ecbc3/src/util/c_sizeof.cpp#L35

fbrausse:
> Instead of this patch, did you try changing uint64_t here to size_type()? I remember having an issue with a similar error message, but could also be that this is unrelated to your issue ^^
> 
> https://github.com/esbmc/esbmc/blob/9a25e56e07f8edf5cc45d4bcf77e7dab7a2ecbc3/src/util/c_sizeof.cpp#L35

That's a good point, we should change that. But indeed, it's unrelated here.

PR #1975, opened 2024-08-13
[clang-cpp] Add proper support for pseudo destructors

intrigus-lgtm:


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -795,7 +795,14 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
   case clang::Stmt::CXXPseudoDestructorExprClass:
   {
-    new_expr = exprt("pseudo_destructor");
+    const clang::CXXPseudoDestructorExpr &cxxpd =
+      static_cast<const clang::CXXPseudoDestructorExpr &>(stmt);
```
mikhailramalho, line 799: can't we just return here?

We don't actually evaluate the base expression in adjust, clang already did it for us.

intrigus-lgtm: I'm afraid I don't get what you mean.

The base expression can have side effects so we must generate IR that evaluates the base.

I added a [test](https://github.com/esbmc/esbmc/pull/1975/files#diff-ae0f70c59c9ea5ccdebfa1ec9383909a767598f4385025a0f5151ee39095143c) for this case.

mikhailramalho: Can you try just:
```
if (get_expr(*cxxpd.getBase(), new_expr))
  Return true;
```
And check if the tests still pass?

mikhailramalho: If it does, you can remove all changes from adjust

intrigus-lgtm: This does not work.
```
esbmc: /home/sg/dev/esbmc/src/util/std_types.h:392: code_typet& to_code_type(typet&): Assertion `type.id() == typet::t_code' failed.
```
This is because we/clang have generated a function_call that has a pseudo destructor as its target. So the adjuster then fails because the target does not have type code.
So we'd have to generate a new function for a pseudo destructor or just remove the whole function_call as I did in this PR.

mikhailramalho: I see. Let's follow the standard then and override adjust_side_effect_function_call in the cpp_adjust + handle the pseudo constructor and fallback to c_adjust otherwise.

mikhailramalho (changes requested):
Let's follow the standard then and override adjust_side_effect_function_call in the cpp_adjust + handle the pseudo constructor and fallback to c_adjust otherwise.

lucasccordeiro:
@intrigus: Are @mikhailramalho's comments clear? If not, please let us know your questions so that we can move forward with this PR.

src/clang-cpp-frontend/clang_cpp_convert.cpp
intrigus-lgtm: Done in https://github.com/esbmc/esbmc/pull/1975/commits/9c52ccaa63be5860525390ad3c51f75529d092a7.

intrigus-lgtm:
PTAL :)

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
```diff
@@ -394,3 +410,15 @@ void clang_cpp_adjust::convert_exception_id(
   if (!cpp_type.empty())
     ids.emplace_back(cpp_type + suffix);
 }
+void clang_cpp_adjust::adjust_side_effect_function_call(
```
mikhailramalho, line 414: add new line between the methods

mikhailramalho (approved):
Looks way better. I suggest only dropping the `adjust_cpp_pseudo_destructor_call` function and moving the code to the new function you created: it's so small and there isn't much else going on in `adjust_side_effect_function_call`  anyway.

lucasccordeiro (approved):

As soon as you address @mikhailramalho's minor comment, I'll merge it into the master.

lucasccordeiro:

PR #1978, opened 2024-08-14
SSA promotion

rafaelsamenezes:
This PR adds the flag `--goto-ssa-promotion` which will introduce phi-nodes into the goto program for all program variables. It consists in:

1. Computing dominator information about the CFG.
2. Computing information about the variables: when it is defined and when it is used. The current precondition of the promotion is that the symbol address can not be used in a store or function call (this is similar to what LLVM mem2reg does).
3. Introducing the phi2t expression. This expression consists in the connection between the 2 previous basic blocks. To index the previous instructions, I am using locations. This is not ideal, and the main issue is that we do not have any stable indexes for instructions.
4. Inserting the phinodes and renaming all uses and definitions.

Although this PR still has some extensions and features that could be implemented, the basic feature is implemented. The next big step is to add support for the phi2t at goto-symex.


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1881,6 +1881,16 @@ bool esbmc_parseoptionst::process_goto_program(
       goto_coveraget tmp(ns, goto_functions);
       tmp.add_false_asserts();
     }
+
+    // From here on all changes should take phi-nodes into account
```
lucasccordeiro: Why should we take phi-nodes into account?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1881,6 +1881,16 @@ bool esbmc_parseoptionst::process_goto_program(
...
+    if (cmdline.isset("goto-ssa-promotion"))
+    {
+      goto_cfg cfg(goto_functions);
+      ssa_promotion ssa(cfg, goto_functions, context);
+      ssa.promote();
```
lucasccordeiro, line 2347: Add some documentation here.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1960,6 +1970,13 @@ bool esbmc_parseoptionst::output_goto_program(
       return true;
     }
 
+    if (cmdline.isset("dump-goto-cfg"))
```
lucasccordeiro, line 2430: Can you explain this option?

src/goto-programs/goto_cfg.cpp
```diff
@@ -74,21 +77,33 @@ goto_cfg::goto_cfg(goto_functionst &goto_functions)
     }
 
     // Third pass - identify all the successors/predecessors
+    int counter = 0;
```
lucasccordeiro: shall we consider it as `size_t`?

src/goto-programs/goto_cfg.cpp
```diff
@@ -74,21 +77,33 @@ goto_cfg::goto_cfg(goto_functionst &goto_functions)
...
     for (auto &bb : bbs)
     {
       assert(bb->begin != bb->end);
       auto last = bb->end;
       last--;
+      bb->uuid = counter++;
 
       for (const auto &bb2 : bbs)
       {
+        const bool is_goto = last->is_goto() || last->is_backwards_goto();
```
lucasccordeiro, line 92: Add some explanation.

src/goto-programs/goto_cfg.cpp
```diff
@@ -74,21 +77,33 @@ goto_cfg::goto_cfg(goto_functionst &goto_functions)
...
+        const bool is_trivial_goto = is_goto && is_true(last->guard);
         if (bb2->begin == bb->end)
         {
-          bb->successors.insert(bb2);
-          bb2->predecessors.insert(bb);
+          if (is_trivial_goto)
```
lucasccordeiro, line 96: What do we define as a trivial goto?

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+
+template <class F>
+void goto_cfg::foreach_bb(
+  const std::shared_ptr<goto_cfg::basic_block> &start,
+  F foo)
+{
+  std::unordered_set<std::shared_ptr<goto_cfg::basic_block>> visited;
+  std::vector<std::shared_ptr<goto_cfg::basic_block>> to_visit({start});
+
+  while (!to_visit.empty())
```
lucasccordeiro, line 222: Add some documentation here.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+    log_status("CFGNode");
+    node->begin->dump();
+
+    log_status("Dominated");
+    for (const auto &edge : edges)
+      edge->begin->dump();
+  }
+}
+
+CFGNode Dominator::idom(const CFGNode &n) const
```
lucasccordeiro, line 307: Add some documentation here.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+    break;
+  }
+
+  return result;
+}
+
+Dominator::DJGraph::DJGraph(const CFGNode &cfg, const Dominator &dom)
+  : tree(dom)
+{
+  // A DJ-Graph is a graph composed by D-Edges and J-Edges
```
lucasccordeiro, line 408: Add a reference to a paper?

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+  log_status("Dumping DJ-Graph");
+  for (const auto &[k, edges] : _graph)
+  {
+    log_status("CFGNode {}", k->uuid);
+    for (const auto &e : edges)
+      log_status("\t->{}", e->uuid);
+  }
+}
+
+std::unordered_map<CFGNode, size_t> Dominator::DomTree::get_levels() const
```
lucasccordeiro, line 432: Add some documentation.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+    {
+      levels[nodes] = level;
+      worklist.insert(nodes);
+    }
+  }
+  return levels;
+}
+
+std::unordered_set<CFGNode>
+Dominator::DomTree::get_subtree(const CFGNode &n) const
```
lucasccordeiro, line 456: Add some documentation.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+                                   const irep_idt &var,
+                                   goto_programt::instructionst::iterator &I,
+                                   const CFGNode &bb) {
+    // Are we dealing with a global?
+    if (!symbol_map.count(var.as_string()))
+      return;
+    ;
+    symbol_information &sym = symbol_map[var.as_string()];
+
+    sym.def_blocks.insert(bb);
```
lucasccordeiro, line 547: Shall we consider using `emplace` when inserting elements into containers such as `std::unordered_set` to avoid unnecessary copying or moving?

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+  std::unordered_set<std::string> unpromotable_symbols;
+
+  const auto insert_definition = [&symbol_map](
+                                   const irep_idt &var,
+                                   goto_programt::instructionst::iterator &I,
+                                   const CFGNode &bb) {
+    // Are we dealing with a global?
+    if (!symbol_map.count(var.as_string()))
+      return;
+    ;
```
lucasccordeiro: You have an unnecessary semicolon after return!

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+{
+  std::unordered_map<std::string, symbol_information> symbol_map;
+  std::unordered_set<std::string> unpromotable_symbols;
+
+  const auto insert_definition = [&symbol_map](
+                                   const irep_idt &var,
+                                   goto_programt::instructionst::iterator &I,
+                                   const CFGNode &bb) {
+    // Are we dealing with a global?
+    if (!symbol_map.count(var.as_string()))
```
lucasccordeiro, line 543: Shall we return early if the condition fails?

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+      return;
+    ;
+    symbol_information &sym = symbol_map[var.as_string()];
+
+    sym.def_blocks.insert(bb);
+    sym.def_instructions.insert(
+      std::make_shared<goto_programt::instructiont>(*I));
+  };
+
+  const auto insert_use = [&symbol_map, &unpromotable_symbols](
```
lucasccordeiro, line 552: In the `insert_use` lambda, you extract symbols twice (first with `get_addr_symbols` and then with `get_symbols`). If the logic permits, consider combining these two loops, reducing code duplication to improve performance!

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+      unpromotable_symbols.insert(symbol.thename.as_string());
+    }
+
+    symbols.clear();
+    get_symbols(e, symbols);
+    for (const expr2tc &sym_expr : symbols)
+    {
+      assert(is_symbol2t(sym_expr));
+      const symbol2t &symbol = to_symbol2t(sym_expr);
+      assert(symbol_map.count(symbol.thename.as_string()));
```
lucasccordeiro, line 571: While assertions are good for debugging, they won't catch issues in production. You can handle cases where the symbol is unexpectedly missing, either by logging an error or throwing an exception.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+
+std::unordered_map<std::string, ssa_promotion::symbol_information>
+ssa_promotion::extract_symbol_information(const CFGNode &start) const
+{
+  std::unordered_map<std::string, symbol_information> symbol_map;
+  std::unordered_set<std::string> unpromotable_symbols;
+
+  const auto insert_definition = [&symbol_map](
+                                   const irep_idt &var,
+                                   goto_programt::instructionst::iterator &I,
```
lucasccordeiro, line 540: Ensure that your variable and function names are consistent and descriptive. For instance, `I` for the iterator can be `instruction_iter` to clarify its purpose. Similarly, `e` can be renamed to expression for clarity.

src/goto-programs/goto_cfg.cpp
```diff
@@ -165,3 +194,646 @@ void goto_cfg::dump_graph() const
...
+          insert_use(it->code, it, bb);
+          insert_use(it->guard, it, bb);
+        }
+      }
+    });
+
+  for (auto symbol : unpromotable_symbols)
+    symbol_map.erase(symbol);
+
+  // TODO: Technically, not all address uses needs to be avoided
```
lucasccordeiro, line 623: The existing TODO comments are useful, but they could be made more specific.

lucasccordeiro (changes requested):

src/esbmc/esbmc_parseoptions.cpp
rafaelsamenezes: I changed the comment. The idea is that  the phi-nodes locations are sensitive, so if the goto program changes the phinodes might become invalid.

src/esbmc/esbmc_parseoptions.cpp
rafaelsamenezes: Done

src/esbmc/esbmc_parseoptions.cpp
rafaelsamenezes: This already came from a previous PR, I just moved it to be before --goto-functions-*

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Just replaced it

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Added a comment

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: A goto with a trivial guard, so it only has one target in practice.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Added a comment

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: The header contains the doc for this method

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: The reference is at the declaration of the class

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Do we really need it? It is to get the levels of a tree.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Same as above

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: In the current implementation we try to tackle all symbols of the function at the same time. So we can't early terminate.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: I guess we could add a method `apply_over_all_expr_symbols`. I would prefer to leave it to another PR though.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: If the assert is skipped it will result in an exception throw by trying to access an element that does not exist in the map.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: The current implementation is consistent with esbmc code base.

src/goto-programs/goto_cfg.cpp
rafaelsamenezes: Not sure how much more specific I can be.

src/irep2/irep2_utils.h
```diff
@@ -1,6 +1,7 @@
 #ifndef UTIL_IREP2_UTILS_H_
 #define UTIL_IREP2_UTILS_H_
 
+#include "irep2/irep2.h"
```
XLiZHI:
```suggestion
#include <rep2/irep2.h>
```

Can we use angle bracket here? so it looks consistent.

src/goto-programs/goto_cfg.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include "context.h"
+#include "location.h"
+#include <optional>
```
XLiZHI, line 5: same here

XLiZHI (commented):
Sorry, I don't know much about CFG, but I have some comments about headers

pmatos:
thanks for resolving the conflicts - i will take a look at this.

pmatos:
@rafaelsamenezes — reviewing this during a stale-PR sweep.

The feature is still absent from `master` (no SSA-promotion pass, no `--goto-ssa-promotion`, no `phi2t`) and not superseded, so it's still wanted. The catch is that the irep2 flattening in #4594 removed the template-registration files this PR uses to register `phi2t`, so the irep2 half now has to be re-implemented against the new `expr_kinds.inc` / `irep2_dispatch.h` dispatch model — the branch no longer merges (GitHub reports conflicts on the deleted template files).

Also still outstanding from the earlier review:

- **Potential soundness issue:** `phi` operands are keyed on a `locationt::hash()` of `{line, column}`, which can collide and link the wrong predecessor.
- `info.mode` is written after the struct is copied into `symbol_map` (so the write is lost); the `assert(symbol_map.count(...))` is over-strict for globals/builtins; and the `phi` migration via `implies_exprt` is ill-typed.
- No end-to-end test exercising `ssa_promotion::promote()` (only the dominator / DJ-graph / dominance-frontier pieces are tested), and `phi2t` isn't yet consumed by goto-symex, so the pass has no verification effect yet.

Would you like to rebase and continue this (port `phi2t` to the new irep2 dispatch, address the review points, add an end-to-end test), or should we close it for now and revisit later?

rafaelsamenezes:
> @rafaelsamenezes — reviewing this during a stale-PR sweep.
> 
> The feature is still absent from `master` (no SSA-promotion pass, no `--goto-ssa-promotion`, no `phi2t`) and not superseded, so it's still wanted. The catch is that the irep2 flattening in #4594 removed the template-registration files this PR uses to register `phi2t`, so the irep2 half now has to be re-implemented against the new `expr_kinds.inc` / `irep2_dispatch.h` dispatch model — the branch no longer merges (GitHub reports conflicts on the deleted template files).
> 
> Also still outstanding from the earlier review:
> 
> * **Potential soundness issue:** `phi` operands are keyed on a `locationt::hash()` of `{line, column}`, which can collide and link the wrong predecessor.
> * `info.mode` is written after the struct is copied into `symbol_map` (so the write is lost); the `assert(symbol_map.count(...))` is over-strict for globals/builtins; and the `phi` migration via `implies_exprt` is ill-typed.
> * No end-to-end test exercising `ssa_promotion::promote()` (only the dominator / DJ-graph / dominance-frontier pieces are tested), and `phi2t` isn't yet consumed by goto-symex, so the pass has no verification effect yet.
> 
> Would you like to rebase and continue this (port `phi2t` to the new irep2 dispatch, address the review points, add an end-to-end test), or should we close it for now and revisit later?

Hi @pmatos, thanks for the review. I will close this and try again.

rafaelsamenezes:
Will retry it.

PR #1463, opened 2023-11-09
Take cur_state->guard into account for creating a new thread

fbrausse:
Fixes SV-COMP memsafety
- c/goblint-regression/04-mutex_46-escape_nr.yml
- c/ldv-races/race-2_1-container_of.yml
- c/ldv-races/race-2_2-container_of.yml
- c/ldv-races/race-2_3-container_of.yml
- c/ldv-races/race-2_4-container_of.yml
- c/ldv-races/race-2_5-container_of.yml
- c/ldv-races/race-3_1-container_of-global.yml
- c/ldv-races/race-3_2-container_of-global.yml

I'm running sv-comp on this.

PR #1980, opened 2024-08-14
[concurrency] inherit the current guard when create a new thread

### LINKED ISSUE #1973: [concurrency] reachability tree error due to nondet
```c
#include <pthread.h>
#include <assert.h>

pthread_t t1, t2;

_Bool flag = 0;

void *thread2(void *arg) {
   // part 1
   assert(!flag);
   return 0;
}

void *thread1(void *arg) {
    pthread_create(&t2, NULL, thread2, NULL);
    return 0;
}

int main(void) {
    if (nondet_int()) 
    {
      pthread_create(&t1, NULL, thread1, NULL);
      return 0;
    }
    // part 2
    flag = 1;

    return 0;
}
```

The result should be `verification successful`.

In this code, part1 and part2 should never be interleaved, because after creating the thread, the subsequent code part2 will not be executed. I think this is probably caused by creating the thread again in the child thread.

XLiZHI:
The current algorithm does not seem to take into account the creation of new threads in child threads.

In this case, the created thread should inherit the child thread's guard instead of using the global guard.


XLiZHI:
This PR:
```
  correct:           14140
    correct true:     7755
    correct false:    6385
  incorrect:            29
    incorrect true:     12
    incorrect false:    17
  unknown:            9636
  Score:             21239 (max: 38482)
https://github.com/esbmc/esbmc/actions/runs/10392154267
```

Master:
```
  correct:           14149
    correct true:     7762
    correct false:    6387
  incorrect:            29
    incorrect true:     12
    incorrect false:    17
  unknown:            9627
  Score:             21255 (max: 38482)
https://github.com/esbmc/esbmc/actions/runs/10392148531
```

src/goto-symex/execution_state.cpp
```diff
@@ -727,6 +727,7 @@ unsigned int execution_statet::add_thread(const goto_programt *prog)
   unsigned int thread_nr = threads_state.size();
   new_state.source.thread_nr = thread_nr;
   new_state.global_guard = cur_state->guard;
+  new_state.guard = cur_state->guard;
```
fbrausse, line 729: I've updated the global_guard one line above your addition in #1463, which had a similar aim to the one your PR describes, and it fixed a couple of SV-COMP benchmarks.

Since with this PR's change the cur_state->guard would be copied to both, the new_state's guard and also its global_guard, I'm wondering whether this duplication is necessary.

XLiZHI: I think the global guard changes might work in this case:
```
int main(void) {
    pthread_create(&t1, NULL, thread1, NULL); // global guard
    pthread_create(&t2, NULL, thread2, NULL); // global guard
    return 0;
}
```
The current guard is reflected in the guards in the current thread:
```
void *thread1(void *arg) {
    pthread_create(&t2, NULL, thread2, NULL);   // inherit the guard from thread1
    return 0;
}

int main(void) {
    pthread_create(&t1, NULL, thread1, NULL);
    return 0;
}
```
I did not understand this part deeply, please correct me if I am wrong.
_Edit:_ Maybe we only need to change the cur guard. 

We added the new thread in the main thread without setting the `cur guard` for the new thread. This means that in the new thread, `cur guard` is empty, so `new_state.global_guard = cur_state->guard;` will not work when we create new thread in this child thread.

fbrausse: > This means that in the new thread, `cur guard` is empty, so `new_state.global_guard = cur_state->guard;` will not work when we create new thread in this child thread.

Can you explain why it wouldn't work?

XLiZHI: When we created this thread, we didn't set a guard on the thread's state. Therefore, the `cur_states->guard` in that thread is empty. `new_state.global_guard` is not assigned a meaningful value. `cur_state->guard` is only created in the main thread.

mikhailramalho: I never really got this code, what's the difference between the global guard and the state guard? Do we need both?

XLiZHI: My understanding is that the current guard is added to `global_guard` when we create a new thread.
https://github.com/esbmc/esbmc/blob/69fbe894f3733bb808d5d57d9d35fb9653a71070/src/goto-symex/builtin_functions.cpp#L683

Both the guard and the global guard of cur states are used when an assertion is executed.
https://github.com/esbmc/esbmc/blob/69fbe894f3733bb808d5d57d9d35fb9653a71070/src/goto-symex/symex_main.cpp#L104-L105

fbrausse (approved):

fbrausse:
Sorry, I forgot to ask: are the svcomp benchmarks working with the global_guard before still working with the current version of this PR?

XLiZHI:
> Sorry, I forgot to ask: are the svcomp benchmarks working with the global_guard before still working with the current version of this PR?

I re-ran sv-comp benchmarks on the current PR and let's wait for the result.

https://github.com/esbmc/esbmc/actions/runs/10426340268

fbrausse:
Many thanks! If they work fine we could open an issue about getting rid of the global_guard. If indeed what you quoted above is the only usage of it, it seems we might just be able to.

lucasccordeiro:
This PR (https://github.com/esbmc/esbmc/actions/runs/10426340268):

````
2024-08-17T01:16:16.7640895Z Statistics:          23805 Files
2024-08-17T01:16:16.7647377Z   correct:           14156
2024-08-17T01:16:16.7647902Z     correct true:     7768
2024-08-17T01:16:16.7648311Z     correct false:    6388
2024-08-17T01:16:16.7683099Z   incorrect:            29
2024-08-17T01:16:16.7685272Z     incorrect true:     12
2024-08-17T01:16:16.7686223Z     incorrect false:    17
2024-08-17T01:16:16.7687454Z   unknown:            9620
2024-08-17T01:16:16.7688674Z   Score:             21268 (max: 38482)
````

Master (https://github.com/esbmc/esbmc/actions/runs/10420430093):

````
  correct:           14146
    correct true:     7762
    correct false:    6384
  incorrect:            29
    incorrect true:     12
    incorrect false:    17
  unknown:            9630
  Score:             21252 (max: 38482)
````

lucasccordeiro (approved):

lucasccordeiro:

PR #1984, opened 2024-08-15
[BUILDING] update CHERI build instructions

XLiZHI:
This PR updated the build instructions for CHERI.

I still had a few questions about CHERI:

1. Regarding `sysroot`, is there a difference between the libraries in `rootfs-riscv64-purecap` and `rootfs-riscv64-hybrid`? If not, do we need to build only one of them? 

2. CHERI TCs are present in regression, but it is not currently enabled. Is it possible to add CHERI TCs to CI? This might require architecture-specific runners? Like morello.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -127,12 +127,6 @@ ESBMC_CLANG=$(echo -D{LLVM,Clang}_DIR=$PWD/clang13) &&
 ESBMC_STATIC=Off
 ```
 
-If you want to build ESBMC-CHERI, please built with the option:
-
-```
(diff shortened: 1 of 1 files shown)


fbrausse:
Hi @XLiZHI.
Re 1. There are only very tiny differences between hybrid and purecap sysroots for the same architecture. They shouldn't matter for our purposes. I'd recommend to choose the hybrid one.

2. Ideally, it shouldn't require specific runners (i.e., host archs), that's that the sysroot is for. But it would require installing/unpacking those sysroots on the runner(s).

However, many of the CHERI regressions tests are not supported, yet, as I lacked the time to finish implementing the full support.

lucasccordeiro (approved):

Issue #1982, opened 2024-08-15
Realloc and memcpy issues while verifying a vector structure


rafaelsamenezes:
I was trying to use esbmc to verify the following code:

```c
typedef struct vector vector;

#define VECTOR_INIT_LENGTH 10
#define VECTOR_FACTOR 2

struct vector {
  char *_buf;
  size_t length;
  size_t _capacity;
  size_t _unit_length;
};

vector vector_init(const size_t unit_length) {
  vector v;
  v._unit_length = unit_length;
  v.length = 0;
  v._capacity = VECTOR_INIT_LENGTH;
  v._buf = malloc(unit_length * VECTOR_INIT_LENGTH);

  return v;
}

void vector_destroy(vector * const v) {
  if (v->_buf != NULL)
    free(v->_buf);
}

void vector_reserve(vector * const v, const size_t quantity) {
  assert(v->_capacity < quantity); // We should be increasing
#ifdef __ESBMC
  char *buf = malloc(quantity * v->_unit_length);
  memcpy(buf, v->_buf, v->_capacity);
  v->_capacity = quantity;
  free(v->_buf);
  v->_buf = buf;
#else
  v->_capacity = quantity;
  v->_buf = realloc(v->_buf, v->_capacity * v->_unit_length);
  #endif
}

void vector_push_back(vector * const v, const void * const elem) {
  if (v->length == v->_capacity)
    vector_reserve(v, v->_capacity * VECTOR_FACTOR);

  void *addr =  &((char*)(v->_buf))[v->length++ * v->_unit_length];
  memcpy(addr, elem, v->_unit_length);
}

char *vector_at(vector *const v, const size_t i) {
  assert(i < v->length);
  return (char*)(v->_buf) + (i * v->_unit_length);
}

void vector_test() {
  size_t length = 25;
  char arr[length];
  for (char i = 0; i < length; i++)
    arr[i] = i;

  printf("\t- Testing vectors\n");
  vector v = vector_init(sizeof(char));

  for (int i = 0; i < length; i++)
    vector_push_back(&v, &arr[i]);

  assert(v.length == length);
  for (int i = 0; i < length; i++)
      assert(*vector_at(&v, i) == i);

  vector_destroy(&v);
}
```
I had the following issues

1. The `_buf` member cannot be a `void*` or our typechecker will complain during symex. (Fix available at #1986);
2. The assertion ` assert(*vector_at(&v, i) == i);` fails when using realloc.
[...]

PR #1986, opened 2024-08-16
Type compat now checks empty types before typed pointers

rafaelsamenezes:
As show in #1982 (item 1), the use of `void*` could lead to some invalid typechecking. This PR changes the order of the checks so we short circuit void cast before pointer checks.


lucasccordeiro (approved):
@rafaelsamenezes: could you please run this PR over the SV-COMP benchmarks?

rafaelsamenezes:
> @rafaelsamenezes: could you please run this PR over the SV-COMP benchmarks?

It shouldn't make a difference as the assertions are removed from release builds.

Issue #1964, opened 2024-08-06
[Python-frontend] known bugs - builtin functions and int_bit_length


haru-02:
### built-in-functions :
Let us take a look at few examples below from #1963 
```
def foo() -> None:
    a = int(2.121212121212)
    assert a == 2

foo()

x = int(1)
assert x == 1

y = bool(True)
assert y == True
```
assert a == 2 fails ! a value will be 2 according to python interpreter.
Solving this also solves the issue in the "constants" regression test.

### int-bit-length :
```
x = int(0)
assert x.bit_length() == 0

y = int(16)
assert y.bit_length() == 5

z = int(2 ** 64)
assert z.bit_length() == 65
```
This assertion  for z should be successful, but fails at the end. this edge case should be analysed. regular python can interpret this number.

brcfarias:
Hi @haru-02,

For the first item, please take a look here:

https://github.com/esbmc/esbmc/blob/f1960cc1cb5041ffa1611270e445a2610bf5a658/src/python-frontend/python_converter.cpp#L799-L814

You could address this by checking if the **int** function is being invoked and truncating the argument (arg).

haru-02:
Thank you @brcfarias , I will check this out

PR #1987, opened 2024-08-16
[python-frontend][WIP] fixed int type casting for constants

haru-02:
fixed the error for the test mentioned previously in #1964 for built-in-functions. It can now accurately truncate constants, but a workaround for evaluating expressions is required.


regression/python/built-in-functions/main.py
```diff
@@ -4,7 +4,7 @@ def foo() -> None:
 
 foo()
 
-x = int(1)
```
brcfarias, line 7: Could you keep this basic usage and add another assignment?

src/python-frontend/python_converter.cpp
```diff
@@ -803,12 +803,20 @@ exprt python_converter::get_function_call(const nlohmann::json &element)
...
       size_t arg_size = 1;
       const auto &arg = element["args"][0];
+	
+      nlohmann::json temp_arg = arg;
 
+      if (func_name == "int" && arg["value"].is_number())
+      {
+        double arg_value = arg["value"].get<double>();
+        temp_arg["value"] = static_cast<int>(arg_value);  // Truncate value in the copy
+      }
```
brcfarias, line 813: Some suggestions:
- Remove `const` from `arg` and manipulate it directly.
- Add it as an `else` block for the `if (func_name == "str")`
- Use `arg["value"].is_number_float()` instead of `arg["value"].is_number()`

brcfarias (changes requested):

PR #1989, opened 2024-08-17
[python-frontend] Get type from function arguments

brcfarias:
This PR enables inferring the type of variables from function arguments. For example:

```python
def foo(x:int, y:int):
  z = x # The type of 'z' is inferred as 'int' from the 'x' argument
```

Additionally, it refactors `python_annotation.h `to enhance readability.


lucasccordeiro (approved):

lucasccordeiro:

PR #1990, opened 2024-08-17
Fix overloading of virtual methods. Fixes #1749

### LINKED ISSUE #1749: [clang-cpp] Overloads in vtable building lead to assertion error.
Running this code

```cpp
struct c
{
  virtual int d();
  virtual int d(int);
};
```
results in:

```
esbmc: esbmc/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp:585: void clang_cpp_convertert::add_vtable_variable_symbols(const clang::CXXRecordDecl&, const struct_typet&, const switch_table&): Assertion `value.type() == compo.type()' failed.
``` 

The reason pretty much is that the current code does not handle overloads.
`method_name` is just `d` in both cases.
https://github.com/esbmc/esbmc/blob/fb01095919c3cbbd2566d0c73fb6bc57b16cda61/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp#L108
https://github.com/esbmc/esbmc/blob/fb01095919c3cbbd2566d0c73fb6bc57b16cda61/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp#L110
(I don't know which of the two lines is the exact culprit but it doesn't really matter)

The end result is that there is only one entry for `d` in the `switch_map` here and so we are guaranteed to find a wrong type:
https://github.com/esbmc/esbmc/blob/fb01095919c3cbbd2566d0c73fb6bc57b16cda61/src/clang-cpp-frontend/clang_cpp_convert_vft.cpp#L582

intrigus-lgtm:


lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:

PR #1994, opened 2024-08-21
Fedora Build Fix + Fedora Build Instructions

Yiannis128:
I've fixed the build process for Fedora, and now it builds like it does with Ubuntu LTS 20.04


src/esbmc/CMakeLists.txt
```diff
@@ -45,6 +45,6 @@ message("Looking at ${CVC_LINK_DIR}")
 target_link_libraries(esbmc ${OLD_FRONTEND_TARGETS} ${SOLIDITY_FRONTEND_TARGETS} ${PYTHON_FRONTEND_TARGETS}
                             ${GOTO_CONTRACTOR_TARGETS} ${JIMPLE_FRONTEND_TARGETS} ${JIMPLE_FRONTEND_TARGETS}
                             clangcfrontend clangcppfrontend filesystem symex pointeranalysis langapi util_esbmc bigint
-                            solvers clibs gotoalgorithms cache ${Boost_LIBRARIES} goto2c)
+                            solvers clibs gotoalgorithms cache ${Boost_LIBRARIES} goto2c m stdc++)
```
rafaelsamenezes: libm should only be added for Unix targets.

rafaelsamenezes (changes requested):
Overall its ok. Just make libm unavailable for windows targets.

Yiannis128:
> Overall its ok. Just make libm unavailable for windows targets.

Is the library unavailable on Windows? Or is it another reason?

CMakeLists.txt
```diff
@@ -57,7 +57,6 @@ include(Utils)
 include(OSConfiguration)
 include(Options)
 include(SendFileHack)
-include(AppleConfiguration)
```
Yiannis128, line 60: @rafaelsamenezes AppleConfiguration is included in OSConfiguration already, so I've removed it here. Is that OK?

src/esbmc/CMakeLists.txt
Yiannis128: OK, I've created a new variable that gets set differently for each OS in OSConfiguration. It's called `OS_INCLUDE_LIBS`, it gets used on targets that did not compile successfully. It's set as a blank value for any OS other than UNIX now.

lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:

PR #1995, opened 2024-08-22
[concurrency] use address of var for the data race check

XLiZHI:
This PR rewrites the code for dereference by the data race checking algorithm. we have now used the address as evidence that a data race has occurred.

1. introduce a new expression: `RACE_CHECK(&x)`, Its op is the address of the variable, which we will replace during symbolic execution.
2. introduce new infinite array: `__ESBMC_races_flag = array_of(0);` during symbolic execution, we take the variable address as index.
3. `__ESBMC_yield()` is added before the atomic block to force the switch, which avoids situations in which a race cannot be checked using context-bound 2

To summarize, we implement the data race check as follows:

WRITE:
```
__ESBMC_yield(); 
Atomic
{
Assert (!__ESBMC_races_flag[&(*x)]);
__ESBMC_races_flag[&(*x)] = 1;  // set
*x = 2;
}
__ESBMC_races_flag[&(*x)] = 0; // reset
```

READ:
```
__ESBMC_yield(); 
Atomic
{
Assert (!__ESBMC_races_flag[&(*x)]);
int y = *x ;
}
```

This PR:
```
  Statistics:          783 Files
  correct:             366
    correct true:      310
    correct false:      56
  incorrect:             9
    incorrect true:      7
    incorrect false:     2
  unknown:             408
  Score:               420 (max: 1488)
  https://github.com/esbmc/esbmc/actions/runs/10509106071
```
Master:
```
  correct:             371
    correct true:      323
    correct false:      48
  incorrect:            18
    incorrect true:     10
    incorrect false:     8
  unknown:             394
  Score:               246 (max: 1488)
https://github.com/esbmc/esbmc/actions/runs/10507450329
```

### DIFF
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -20,13 +20,14 @@ public:
     const exprt &original_expr,
     bool deref)
   {
-    if (deref)
-    {
-      irep_idt name = "c:@__ESBMC_races_flag";
-      const symbolt *symbol = context.find_symbol(name);
+    // if (deref)
+    // {
+    //   irep_idt name = "__ESBMC_races_check";
+    //   const symbolt *symbol = context.find_symbol(name);
+    //   assert(symbol);
 
-      return *symbol;
-    }
+    //   return *symbol;
+    // }
 
     const irep_idt identifier =
       deref ? "__ESBMC_deref_" + id2string(object) : "tmp_" + id2string(object);
@@ -58,18 +59,16 @@ public:
     const exprt &original_expr,
     bool deref)
   {
-    exprt expr = symbol_expr(get_guard_symbol(object, original_expr, deref));
-
     if (deref)
     {
       exprt address = address_of_exprt(original_expr);
-      index_exprt index = index_exprt(expr, address, bool_type());
+      exprt check("races_check", typet("bool"));
+      check.copy_to_operands(address);
 
-      expr.swap(index);
-
-      return expr;
+      return check;
     }
 
+    exprt expr = symbol_expr(get_guard_symbol(object, original_expr, deref));
     if (original_expr.is_index() && expr.type().is_array())
     {
       index_exprt full_expr = to_index_expr(original_expr);
@@ -97,17 +96,28 @@ public:
       get_guard_symbol_expr(entry.object, entry.original_expr, entry.deref));
   }
 
-  void add_initialization(goto_programt &goto_program) const;
+  void add_initialization(goto_programt &goto_program);
 
 protected:
   contextt &context;
 };
 
-void w_guardst::add_initialization(goto_programt &goto_program) const
+void w_guardst::add_initialization(goto_programt &goto_program)
 {
   goto_programt::targett t = goto_program.instructions.begin();
   const namespacet ns(context);
 
+  type2tc arrayt = array_type2tc(get_bool_type(), expr2tc(), true);
+  const irep_idt identifier = "__ESBMC_races_flag";
+  w_guards.push_back(identifier);
+  symbolt new_symbol;
+  new_symbol.id = identifier;
+  new_symbol.name = identifier;
+  new_symbol.type = migrate_type_back(arrayt);
+  new_symbol.static_lifetime = true;
+  new_symbol.value.make_false();
+  context.move_symbol_to_context(new_symbol);
+  
   for (const auto &w_guard : w_guards)
   {
     const symbolt &s = *ns.lookup(w_guard);
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -20,24 +20,13 @@ public:
     const exprt &original_expr,
     bool deref)
   {
-    // if (deref)
-    // {
-    //   irep_idt name = "__ESBMC_races_check";
-    //   const symbolt *symbol = context.find_symbol(name);
-    //   assert(symbol);
-
-    //   return *symbol;
-    // }
-
-    const irep_idt identifier =
-      deref ? "__ESBMC_deref_" + id2string(object) : "tmp_" + id2string(object);
+    const irep_idt identifier = "tmp_" + id2string(object);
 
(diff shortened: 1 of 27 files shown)


XLiZHI:
I think this method works for any variable type, such as arrays, member variables, etc. However, it also puts demands on the solver, which I found Boolector took a lot of time to solve, Z3 is much more efficient int this method. Especially in the following situations:

```
struct s {
  int datum;
  struct s *next;
} *A, *B;

void *t_fun(void *arg) {
  A->next->datum++;
}
```

 ```
        FUNCTION_CALL:  yield()

        ATOMIC_BEGIN

        ASSERT !(RACE_CHECK(&A->next->datum)) // W/W data race on c:@A.next.datum

        ASSIGN RACE_CHECK(&A->next->datum)=1;

        ASSIGN A->next->datum=A->next->datum + 1;

        ATOMIC_END

        ASSIGN RACE_CHECK(&A->next->datum)=0;
```

In my opinion, is it because it takes too much resources to calculate the address. I want to know if my method will work? Is there a better way?

**Comment by lucasccordeiro** [comment]
@XLiZHI: Your solution looks good to me.

Have you confirmed whether the incorrect results are resolved or still result in timeouts?

Regarding solver performance, our current slicer isn't field-sensitive for structs. The next step would be to enhance the slicer by considering struct fields. We also want to share our benchmarks with the SMT solver developers, enabling them to optimize their solver based on our data or provide more effective guidance on encoding the verification conditions.

BTW, our CI for your PR is failing.

**Comment by XLiZHI** [comment]
> Have you confirmed whether the incorrect results are resolved or still result in timeouts?

I have compared the results and I can confirm that these incorrect results are resolved.

> BTW, our CI for your PR is failing.

CI failed because in macos, the new boost version was used, which is not related to this PR.

File: src/goto-symex/builtin_functions.cpp [hunk 1]
```diff
@@ -1704,3 +1618,26 @@ bool goto_symext::run_builtin(
 
   return false;
 }
+
+void goto_symext::replace_races_check(expr2tc &expr)
```
lucasccordeiro, line 1622: Can you write some documentation here? You could reuse the text from this PR.

src/goto-symex/builtin_functions.cpp
```diff
@@ -1704,3 +1618,26 @@ bool goto_symext::run_builtin(
...
+      replace_races_check(e);
+  });
+
+  if (is_races_check2t(expr))
+  {
+    // replace with __ESBMC_races_flag[address_of(var)]
+    const races_check2t &obj = to_races_check2t(expr);
+
+    expr2tc flag;
+    migrate_expr(symbol_expr(*ns.lookup("c:@F@__ESBMC_races_flag")), flag);
```
lucasccordeiro, line 1640: Does it work for C++ programs, too?

src/irep2/irep2_expr.cpp
```diff
@@ -388,7 +389,8 @@ void with2t::assert_consistency() const
 {
   if (is_array_type(source_value))
   {
-    assert(is_bv_type(update_field->type));
+    assert(
```
lucasccordeiro, line 392: Shall we use an assert or a log?

src/util/migrate.cpp
```diff
@@ -2715,6 +2721,15 @@ exprt migrate_expr_back(const expr2tc &ref)
     theexpr.copy_to_operands(op0);
     return theexpr;
   }
+  case expr2t::races_check_id:
+  {
+    const races_check2t &ref2 = to_races_check2t(ref);
```
lucasccordeiro, line 2726: Can you document this code?

src/goto-programs/add_race_assertions.cpp
```diff
@@ -218,6 +231,20 @@ void add_race_assertions(
         i_it = ++t;
       }
 
+      // now add assignments for what is written -- reset
```
lucasccordeiro, line 240: Can you improve this documentation?

lucasccordeiro, line 240: What is written?

src/goto-programs/add_race_assertions.cpp
```diff
@@ -79,17 +83,28 @@ class w_guardst
...
   contextt &context;
 };
 
-void w_guardst::add_initialization(goto_programt &goto_program) const
+void w_guardst::add_initialization(goto_programt &goto_program)
 {
   goto_programt::targett t = goto_program.instructions.begin();
   const namespacet ns(context);
 
+  type2tc arrayt = array_type2tc(get_bool_type(), expr2tc(), true);
```
lucasccordeiro, line 102: Can you write a documentation?

lucasccordeiro (changes requested):

src/goto-symex/builtin_functions.cpp
XLiZHI: Yes, because this symbol is not created by the frontend.

src/goto-symex/builtin_functions.cpp
XLiZHI: done

src/util/migrate.cpp
XLiZHI: done

src/goto-programs/add_race_assertions.cpp
XLiZHI: done

src/irep2/irep2_expr.cpp
XLiZHI: This code will only run in the debug version, and based on the name of the function "assert_consistency()", I think assert is appropriate.

lucasccordeiro (approved):

lucasccordeiro:
@XLiZHI: thanks for submitting this PR.

Can I ask you to update our latex file about the ESBMC documentation?  This can be done in another PR.

lucasccordeiro:
I mean, this documentation: https://github.com/esbmc/esbmc/blob/master/docs/manual.tex

lucasccordeiro:

PR #1997, opened 2024-08-23
update boost digest type

### LINKED ISSUE #1993: MacOS Fails to build on boost@1.86
I started running into this issue today when building esbmc.

> `src/util/crypto_hash.cpp:47:26: error: non-const lvalue reference to type 'digest_type' (aka 'unsigned char[20]') cannot bind to a value of unrelated type 'unsigned int[5]'
>   p_crypto->s.get_digest(hash);

> /opt/homebrew/include/boost/uuid/detail/sha1.hpp:179:43: note: passing argument to parameter 'digest' here
> inline void sha1::get_digest(digest_type& digest)`

On investigating, it seems by boost version was upgraded from 1.85 to 1.86.
The change log has this particular line of code from 1.85 to 1.86
`typedef unsigned int(digest_type)[5]; => typedef unsigned char digest_type[ 20 ];`

reverting to boost 1.85 works.

XLiZHI:
```
  size_t to_size_t() const
  {
    size_t result = hash[0];
    for (int i = 1; i < 20; i++)
      // Do we care about overlaps?
      result ^= hash[i];
    return result;
  }
```
I'm not sure about this. Unsigned int: 32; Unsigned char: 8. Does this method need to be changed?

### DIFF
--- a/src/util/crypto_hash.h
+++ b/src/util/crypto_hash.h
@@ -4,20 +4,28 @@
 #include <memory>
 #include <string>
 
+#if BOOST_VERSION >= 108600
+#  define HASH_SIZE 20
+using HashType = unsigned char;
+#else
+#  define HASH_SIZE 5
+using HashType = unsigned int;
+#endif
+
 class crypto_hash_private;
 
 class crypto_hash
 {
 public:
   std::shared_ptr<crypto_hash_private> p_crypto;
-  unsigned int hash[5];
+  HashType hash[HASH_SIZE];
 
   bool operator<(const crypto_hash &h2) const;
 
   size_t to_size_t() const
   {
     size_t result = hash[0];
-    for (int i = 1; i < 5; i++)
+    for (int i = 1; i < HASH_SIZE; i++)
       // Do we care about overlaps?
       result ^= hash[i];
     return result;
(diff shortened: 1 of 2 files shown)


rafaelsamenezes:
Could you give a bit more context? Why are you increasing the size to 20?

Edit: just saw the issue.

src/util/crypto_hash.h
```diff
@@ -10,10 +10,24 @@ class crypto_hash
 {
 public:
   std::shared_ptr<crypto_hash_private> p_crypto;
+#if BOOST_VERSION >= 108600
```
rafaelsamenezes: Could you just use a define or a static constant for the array length? 

This way you dont need to repeat this ifdef.

rafaelsamenezes (changes requested):

src/util/crypto_hash.h
XLiZHI: done

lucasccordeiro:
@XLiZHI: Does this PR fixes this issue https://github.com/esbmc/esbmc/issues/1993?

XLiZHI:
> @XLiZHI: Does this PR fixes this issue #1993?

Yes

lucasccordeiro (approved):

lucasccordeiro:

PR #1999, opened 2024-08-25
[Condition Coverage] Update Condition Coverage Calculation Strategy

ChenfengWei0:
We now count all the conditions, instead of only the ones inside the guard of  if-exprs. This includes the condition within `ASSERT`, `ASSUME`, `ASSIGN`, `GOTO`, `RETURN`, `FUNCTION_CALL` and `OTHER`.

E.g.
```c++
int main()
{
	int a = 1;
	if (a == 1)
	{
		assert(1);
	}
	else
	{
		assert(0);
	}
}
```
<details>
<summary>Result:</summary>

```sh
# --condition-coverage --no-cov-asserts 
# does not count the assert, which is the same result as previous
Reached Conditions:  2
Total Conditions:  2
Condition Properties - SATISFIED:  1
Condition Properties - UNSATISFIED:  1
Condition Coverage: 50%

# --condition-coverage
Reached Conditions:  4
Short Circuited Conditions:  2
Total Conditions:  6
Condition Properties - SATISFIED:  2
Condition Properties - UNSATISFIED:  2
Condition Coverage: 33.333333333333336%

# For comparison: cbmc
# --cover condition --unwind 2 --timestamp wall
** coverage results:
2024-08-25T22:24:44.549526 [main.coverage.1] file Wtest32-B5.c line 7 function main condition 'a == 1' false: SATISFIED
[main.coverage.2] file Wtest32-B5.c line 7 function main condition 'a == 1' true: FAILED
[main.coverage.3] file Wtest32-B5.c line 9 function main condition '1 != 0' false: SATISFIED
[main.coverage.4] file Wtest32-B5.c line 9 function main condition '1 != 0' true: FAILED
[main.coverage.5] file Wtest32-B5.c line 13 function main condition '0 != 0' false: FAILED
[main.coverage.6] file Wtest32-B5.c line 13 function main condition '0 != 0' true: FAILED

2024-08-25T22:24:44.549573 ** 2 of 6 covered (33.3%)
2024-08-25T22:24:44.549589 ** Used 2 iterations
```

</details>

---

we add options `condition-coverage-vb` and `condition-coverage-claims-vb` to generate the runtime verbose output. This will be helpful when the program is so large that ESBMC will eventually run timeout. 

<details>
<summary>Example:</summary>

```sh
Solving claim '!(1 != 0)' with solver Boolector 3.2.2
Solving with solver Boolector 3.2.2
Current Condition Coverage: 0%

Solving claim '1 != 0' with solver Boolector 3.2.2
Solving with solver Boolector 3.2.2
Current Condition Coverage: 16.666666666666668%

Solving claim 'a == 1' with solver Boolector 3.2.2
Solving with solver Boolector 3.2.2
Current Condition Coverage: 33.333333333333336%
```

</details>

---

Optimisation:
- disable the pointer-check by default to speed up
- re-compute line number/target number after inserting the instrumentation assertion.

--- 

Known Issue
- for `ASSIGN`, `RETURN`, `FUNCTION_CALL` and `OTHER`, if its condition contains `function_call()`, the number of total conditions might be greater than the number in the source code. This is due to the removal of side effects and the insertion of the auxiliary variables and statements.
<details>
<summary>Example:</summary>

```c
bool t2()
{
	bool x;
	return x + 1;
}

bool t1()
{
	return 1 == 2 && t2();
}

int main()
{
	int nS;
	int nE = t1();
	bool x = ((nS != nE) && t1() || (2 == 3));  // ASSIGN
        // assert((nS != nE) && t1() || (2 == 3));  // ASSERT
}
```

Result

```sh
# ASSIGN
[Coverage]

Reached Conditions:  10
Short Circuited Conditions:  2
Total Conditions:  12

Condition Properties - SATISFIED:  6
Condition Properties - UNSATISFIED:  4

Condition Coverage: 50%

# ASSERT
[Coverage]

Reached Conditions:  8
Short Circuited Conditions:  2
Total Conditions:  10

Condition Properties - SATISFIED:  5
Condition Properties - UNSATISFIED:  3

Condition Coverage: 50%
```

</details>


src/esbmc/bmc.cpp
```diff
@@ -910,17 +924,36 @@ smt_convt::resultt bmct::multi_property_check(
...
-      if (output_file != "")
+      // for verbose output of cond coverage
+      if (is_vb)
       {
-        std::ofstream out(fmt::format("{}-{}", ce_counter++, output_file));
-        show_goto_trace(out, ns, goto_trace);
+        auto current_pair = std::make_pair(claim.claim_msg, claim.claim_loc);
+        if (total_cond.count(current_pair))
+        {
+          if (options.get_bool_option("condition-coverage-claims-vb"))
```
lucasccordeiro: What is `vb`?

src/esbmc/bmc.cpp
```diff
@@ -946,6 +979,20 @@ smt_convt::resultt bmct::multi_property_check(
         }
       }
     }
+    else if (is_vb && result == smt_convt::P_UNSATISFIABLE)
```
lucasccordeiro: What is `is_vb`? What is `result`? can you use more intuitive variable names?

src/esbmc/bmc.cpp
```diff
@@ -836,6 +848,8 @@ smt_convt::resultt bmct::multi_property_check(
                        &reached_mul_claims,
                        &is_assert_cov,
                        &is_cond_cov,
+                       &is_vb,
```
lucasccordeiro, line 865: What is `vb`?

src/esbmc/bmc.cpp
```diff
@@ -791,10 +791,15 @@ smt_convt::resultt bmct::multi_property_check(
   std::unordered_multiset<std::string> reached_mul_claims;
   bool is_assert_cov = options.get_bool_option("assertion-coverage") ||
                        options.get_bool_option("assertion-coverage-claims");
+  // this might be extended to assert cond
```
lucasccordeiro: The number of options for condition coverage is increasing a lot. We should create a new category of options when invoking `esbmc --help`.

src/esbmc/bmc.cpp
```diff
@@ -805,6 +810,13 @@ smt_convt::resultt bmct::multi_property_check(
   const bool is_fail_fast = !fail_fast.empty() ? true : false;
   const int fail_fast_limit = is_fail_fast ? stoi(fail_fast) : 0;
   int fail_fast_cnt = 0;
+
+  // for cond-vb
+  goto_coveraget temp(ns, symex->goto_functions);
+  const std::set<std::pair<std::string, std::string>> &total_cond =
```
lucasccordeiro: Why do we need a pair of strings here?

src/esbmc/bmc.cpp
```diff
@@ -910,17 +924,36 @@ smt_convt::resultt bmct::multi_property_check(
...
+          }
+          log_result(
+            "Current Condition Coverage: {}%\n",
+            reached_claims.size() * 100.0 / total_cond.size());
+        }
+      }
+
+      else
+      {
+        // Generate Output
```
lucasccordeiro, line 964: Which type of output?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1839,6 +1841,10 @@ bool esbmc_parseoptionst::process_goto_program(
       options.set_option("base-case", true);
       options.set_option("multi-property", true);
       options.set_option("keep-verified-claims", false);
+      options.set_option("no-pointer-check", true);
+
+      // for re-do remove-sideeffects
```
lucasccordeiro, line 1844: Why?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1848,15 +1854,22 @@ bool esbmc_parseoptionst::process_goto_program(
...
       // for multi-property
       options.set_option("result-only", true);
       options.set_option("base-case", true);
       options.set_option("multi-property", true);
       options.set_option("keep-verified-claims", false);
+      // prevent adding property checking assertions during SymEx
+      options.set_option("no-pointer-check", true);
       // unreachable conditions should be also considered as short-circuited
 
+      // for re-do remove-sideeffects
```
lucasccordeiro, line 1866: why?

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1866,8 +1879,15 @@ bool esbmc_parseoptionst::process_goto_program(
...
       std::string filename = cmdline.args[0];
       goto_coveraget tmp(ns, goto_functions, filename);
-      tmp.replace_all_asserts_to_guard(gen_true_expr());
+
+      // if we do not want to count the guard in the assertions
+      if (cmdline.isset("no-cov-asserts"))
+        tmp.replace_all_asserts_to_guard(gen_true_expr());
       tmp.gen_cond_cov();
+
+      // redo conversion to remove_sideeffect
```
lucasccordeiro, line 1884: why?

src/esbmc/options.cpp
```diff
@@ -545,6 +545,13 @@ const struct group_opt_templ all_cmd_options[] = {
       NULL,
       "use '--condition-coverage-claims' while disable "
       "'--no-remove-unreachable'"},
+     {"condition-coverage-vb",
```
lucasccordeiro: Let's create a new category here for condition-coverage.

src/esbmc/bmc.cpp
```diff
@@ -805,6 +810,13 @@ smt_convt::resultt bmct::multi_property_check(
   const bool is_fail_fast = !fail_fast.empty() ? true : false;
   const int fail_fast_limit = is_fail_fast ? stoi(fail_fast) : 0;
   int fail_fast_cnt = 0;
+
+  // for cond-vb
+  goto_coveraget temp(ns, symex->goto_functions);
```
lucasccordeiro:
````CPP
// Create a goto_coveraget object to handle coverage-related tasks
goto_coveraget coverage_handler(ns, symex->goto_functions);

// Conditionally get the total conditional assertions if both is_vb and is_cond_cov are true
const auto &total_cond = (is_vb && is_cond_cov) 
                          ? coverage_handler.get_total_cond_assert()
                          : std::set<std::pair<std::string, std::string>>();
````

File: src/goto-programs/goto_coverage.cpp [hunk 12]
```diff
@@ -18,21 +18,27 @@ void goto_coveraget::replace_all_asserts_to_guard(
...
+          replace_assert_to_guard(guard, it, is_instrumentation);
       }
     }
 }
 
+void goto_coveraget::replace_assert_to_guard(
+  const expr2tc &guard,
+  goto_programt::instructiont::targett &it,
+  bool is_instrumentation)
+{
```
lucasccordeiro, line 39: Add some comments here.

src/goto-programs/goto_coverage.cpp
```diff
@@ -192,11 +233,64 @@ void goto_coveraget::gen_cond_cov()
           guard = handle_single_guard(guard);
 
           exprt pre_cond = nil_exprt();
+          pre_cond.location() = it->location;
           gen_cond_cov_assert(guard, pre_cond, goto_program, it);
         }
+
+        else if (it->is_assign())
```
lucasccordeiro, line 249: Please document the body of the if-statements.

lucasccordeiro (changes requested):

src/esbmc/bmc.cpp
ChenfengWei0: It's for the runtime **v**erbose **o**utput. By enabling this, we will output the coverage information when handling each instrumentation assertion.  This will be helpful when the program is too large and ESBMC cannot finish, we can still get some info about the coverage.

Will add documentation

src/esbmc/bmc.cpp
ChenfengWei0: Will change

src/esbmc/bmc.cpp
ChenfengWei0: Will do

src/esbmc/bmc.cpp
ChenfengWei0: It's the combination of assertion's guard and location, which is used to identify each assertion in multi-property checking.

Will add documentation

src/esbmc/esbmc_parseoptions.cpp
ChenfengWei0: Because we deliberately skip some of the sideeffects removal process when generating the Goto program. This is to keep the condition/guards format and avoid introducing auxiliary variables, which will affect the coverage calculation.

The documentation can be find here:
https://github.com/esbmc/esbmc/blob/f5057afb207cac2cb61b9be1b3ed2354366de96e/src/goto-programs/goto_coverage_rm.cpp#L3C1-L6C43

If we do not redo the removal of the sideeffects, the line number/target will be incorrect and sometimes lead to segfault

src/esbmc/bmc.cpp
ChenfengWei0: Many thanks! Will change

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Will do

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Will do

regression/goto-coverage/github_1720_13/main.c
```diff
@@ -1,1723 +1,14 @@
-//This Program is Problem4-REACHABILITY-TRAINING-RERS18.c
```
fbrausse, line 1: Why are you deleting most of this test case?

fbrausse (commented):
I don't like the new `--*-vb` options. ESBMC already has `--verbosity` and it accepts a string value. For example, `log_debug("module", "my debug message")` would use this string value for e.g. option `--verbosity module:9`, such that debug messages for `"module"` would be printed. It might be better to use the --verbosity option for the "coverage" results by allowing log_status(), log_result() or similar to also pass such a module parameter. Maybe just log_result() to keep the noise down.

Otherwise the changes look good to me (modulo Lucas' comments).

regression/goto-coverage/github_1720_13/main.c
ChenfengWei0: Might be a mistake... There should be only one test case deleted

ChenfengWei0:
> I don't like the new `--*-vb` options. ESBMC already has `--verbosity` and it accepts a string value. For example, `log_debug("module", "my debug message")` would use this string value for e.g. option `--verbosity module:9`, such that debug messages for `"module"` would be printed. It might be better to use the --verbosity option for the "coverage" results by allowing log_status(), log_result() or similar to also pass such a module parameter. Maybe just log_result() to keep the noise down.
> 
> Otherwise the changes look good to me (modulo Lucas' comments).

Thanks for the advice. Will change

lucasccordeiro (approved):

PR #2002, opened 2024-08-27
[python-frontend] Enhancements to type inference and fix for int.bit_length()

brcfarias:
- Improving the type inference algorithm to support operations like:

```python
f = e.bit_length()      # Get type from method return
g = e.bit_length() - 1  # Get type from binary operations with method call as an operand

```
- Fixing a crash that occurred when invoking `int.bit_length()` inside functions:

```python
def foo(x:int) -> None:
  y = x.bit_length()
```

- Refactor: Added a class to manage symbol ID creation and its conversion to string


lucasccordeiro (approved):

PR #2006, opened 2024-08-28
Added Config File Support

Yiannis128:
This PR adds config file support to ESBMC. Now the arguments can be specified in a ini/toml style syntax in a file, and ESBMC will load those values when running. ESBMC will read the env var `ESBMC_CONFIG_PATH` and load the file pointed to by that environment variable. Flags such as `--interval-analysis` can be enabled by simply being included with any value (I chose to use `true` in the example below as it's quite descriptive, but even `false` will enable them).

There are some things still left to do:

- [x] Expand home variable for `ESBMC_CONFIG_PATH` (it needs to be a full path currently, as `~/git/esbmc/esbmc.ini` does not work)
- [x] Allow relative paths to be specified in `ESBMC_CONFIG_PATH`
- [x] Make ESBMC automatically look at `$HOME/.config/esbmc.toml` or `%userprofile%\esbmc.toml` for default values

##### Command

```sh
ESBMC_CONFIG_FILE="/home/yiannis/git/esbmc/esbmc.ini" src/esbmc/esbmc ~/git/esbmc-ai/samples/threading.c
```

##### Config File

```ini
interval-analysis = true
goto-unwind = true
unlimited-goto-unwind = true
k-induction = true
state-hashing = true
add-symex-value-sets = true
k-step = 2
floatbv = true
unlimited-k-steps = true
memory-leak-check = true
context-bound = 2
```

### DIFF
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -260,11 +260,11 @@ bool cmdlinet::parse(
       // File path provided is invalid.
       if (file)
       {
-        log_status("Reading config file: {}", config_path);
+        log_status("Reading config file: {}", config_path.value());
       }
       else
       {
-        log_error("Could not read config file: {}", config_path);
+        log_error("Could not read config file: {}", config_path.value());
         return false;
       }
       boost::program_options::store(
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -269,8 +269,8 @@ bool cmdlinet::parse(
     }
     catch (std::runtime_error &e)
     {
-      log_error("Error while loading config file: {}", e.what());
-      return false;
+      throw std::runtime_error(
+        "Error while loading config file: " + std::string(e.what()));
     }
 
     // Load config file
@@ -281,10 +281,11 @@ bool cmdlinet::parse(
       // File path provided is invalid.
       if (!file)
       {
-        log_error("Error while reading config file: {}", config_path.value());
-        return false;
+        throw std::runtime_error(
+          "Error while reading config file: " + config_path.value());
       }
 
+      // Load config file
       boost::program_options::store(
         parse_toml_file(file, all_cmdline_options, false), vm);
     }
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -3,6 +3,7 @@
 #include <cstdlib>
 #include <filesystem>
 #include <fstream>
+#include <optional>
 #include <sstream>
 
 #include <util/cmdline.h>
@@ -148,9 +149,24 @@ const char *cmdlinet::getval(const char *option) const
   return value->second.front().c_str();
 }
 
-std::string cmdlinet::expand_path_string(std::string path) const
+std::string cmdlinet::expand_user(std::string const path) const
 {
-  return path;
+#ifdef WIN32
+#  define HOME_ENV_NAME "USERPROFILE"
+#else
+#  define HOME_ENV_NAME "HOME"
+#endif
+
+  std::string result = std::string(path);
+
+  // Case ~
+  const std::optional<std::string> home_path = std::getenv(HOME_ENV_NAME);
+  if (!result.empty() && result[0] == '~' && home_path)
+  {
+    result.replace(0, 1, home_path.value());
+  }
+
+  return result;
 }
 
 // Will return empty string if invalid.
@@ -159,8 +175,7 @@ std::string cmdlinet::get_config_file_location() const
   const auto envloc = std::getenv("ESBMC_CONFIG_FILE");
   if (envloc)
   {
-    const std::string config_path =
-      this->expand_path_string(std::string(envloc));
+    const std::string config_path = this->expand_user(std::string(envloc));
     if (std::filesystem::exists(config_path))
     {
       return config_path;
@@ -174,8 +189,7 @@ std::string cmdlinet::get_config_file_location() const
 #else
 #  define DEFAULT_CONFIG_PATH "~/.config/esbmc.toml"
 #endif
-    const std::string config_path =
-      this->expand_path_string(DEFAULT_CONFIG_PATH);
+    const std::string config_path = this->expand_user(DEFAULT_CONFIG_PATH);
     if (std::filesystem::exists(config_path))
     {
       return config_path;
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -3,7 +3,6 @@
 #include <cstdlib>
 #include <filesystem>
 #include <fstream>
-#include <optional>
 #include <sstream>
 
 #include <util/cmdline.h>
@@ -162,15 +161,13 @@ std::string cmdlinet::expand_user(std::string const path) const
   // Case ~
   const std::optional<std::string> home_path = std::getenv(HOME_ENV_NAME);
   if (!result.empty() && result[0] == '~' && home_path)
-  {
     result.replace(0, 1, home_path.value());
-  }
 
   return result;
 }
 
 // Will return empty string if invalid.
-std::string cmdlinet::get_config_file_location() const
+std::optional<std::string> cmdlinet::get_config_file_location() const
 {
   const auto envloc = std::getenv("ESBMC_CONFIG_FILE");
   if (envloc)
@@ -195,7 +192,7 @@ std::string cmdlinet::get_config_file_location() const
       return config_path;
     }
   }
-  return "";
+  return std::nullopt;
 }
 
 bool cmdlinet::parse(
@@ -257,9 +254,10 @@ bool cmdlinet::parse(
       vm);
     // Load config file
     const auto config_path = this->get_config_file_location();
-    if (!config_path.empty())
+    if (config_path && !config_path->empty())
     {
-      std::ifstream file(config_path);
+      std::ifstream file(config_path.value());
+      // File path provided is invalid.
       if (file)
       {
         log_status("Reading config file: {}", config_path);
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -7,6 +7,7 @@
 #include <sstream>
 
 #include <stdexcept>
+#include <string>
 #include <util/cmdline.h>
 #include <util/message.h>
 #include <util/config_file.h>
@@ -168,7 +169,7 @@ std::string cmdlinet::expand_user(std::string const path) const
   if (!result.empty() && result[0] == '~' && home_path)
     result.replace(0, 1, home_path.value());
 
-  return result;
+  return std::filesystem::absolute(result).string();
 }
 
 // Returns the config file path if it is located, if config files should not be
--- a/src/util/cmdline.cpp
+++ b/src/util/cmdline.cpp
@@ -173,8 +173,8 @@ std::string cmdlinet::expand_user(std::string const path) const
 }
 
 // Returns the config file path if it is located, if config files should not be
-// loaded returns a std::nullopt. If a wrong file location is specified, then a
-// std::runtime_error is thrown.
+// loaded returns a std::nullopt. If a wrong file location is specified, then an
+// empty string is returned
 std::optional<std::string> cmdlinet::get_config_file_location() const
 {
   const auto envloc = std::getenv("ESBMC_CONFIG_FILE");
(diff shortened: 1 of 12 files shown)


Yiannis128:
Turns out relative paths are supported :smile:

src/util/cmdline.cpp
```diff
@@ -145,6 +148,42 @@ const char *cmdlinet::getval(const char *option) const
   return value->second.front().c_str();
 }
 
+std::string cmdlinet::expand_path_string(std::string path) const
```
mikhailramalho: why do you need this function?

rafaelsamenezes:
> (I chose to use true in the example below as it's quite descriptive, but even false will enable them).

I guess this could be expanded later. Looks good so far!

src/util/cmdline.cpp
```diff
@@ -145,6 +148,42 @@ const char *cmdlinet::getval(const char *option) const
...
+{
+  return path;
+}
+
+// Will return empty string if invalid.
+std::string cmdlinet::get_config_file_location() const
+{
+  const auto envloc = std::getenv("ESBMC_CONFIG_FILE");
```
mikhailramalho, line 180: We don't load anything from env, all options are passed as flags to the binary. I don't mind keeping this but please add an option too.

src/util/cmdline.cpp
```diff
@@ -145,6 +148,42 @@ const char *cmdlinet::getval(const char *option) const
...
+    }
+  }
+  else
+  {
+    // Load default config file if it exists.
+#ifdef WIN32
+#  define DEFAULT_CONFIG_PATH "%userprofile%\\esbmc.toml"
+#else
+#  define DEFAULT_CONFIG_PATH "~/.config/esbmc.toml"
+#endif
```
mikhailramalho: better to search for the config file in the current running dir.

src/util/cmdline.cpp
```diff
@@ -145,6 +148,42 @@ const char *cmdlinet::getval(const char *option) const
...
+    const std::string config_path =
+      this->expand_path_string(DEFAULT_CONFIG_PATH);
+    if (std::filesystem::exists(config_path))
+    {
+      return config_path;
+    }
+  }
+  return "";
```
mikhailramalho: you can return an optional instead of an empty string.

src/util/cmdline.cpp
```diff
@@ -195,12 +234,33 @@ bool cmdlinet::parse(
...
+    if (!config_path.empty())
+    {
+      std::ifstream file(config_path);
+      if (file)
+      {
+        log_status("Reading config file: {}", config_path);
+      }
+      else
+      {
+        log_error("Could not read config file: {}", config_path);
```
mikhailramalho: let's not fail if there is no config file.

mikhailramalho (changes requested):
Please don't use brackets in one line if/else

mikhailramalho:
> but even `false` will enable them

You mean false will disable it, right?

Yiannis128:
> > but even `false` will enable them
> 
> You mean false will disable it, right?

No, flags such as `--floatbv` just check for the existence of the setting with any value. This is how the default behaviour in the Boost library config parser works as far as I know.

So for example to disable floatbv in the example given above you have to remove it from the config. I'm not sure if I like this either.

src/util/cmdline.cpp
fbrausse: Well, the `ESBMC_OPTS` envvar does exist for while. When also adding an option the precedence should be documented:
1. ESBMC_OPTS
2. ESBMC_CONFIG_FILE
3. command line options

possibly with the additional note that duplicate/contradicting options in either will lead to ESBMC refusing to work, see also discussion #1916.

fbrausse (commented):
While I do like the simplicity of your solution I feel that maybe a different approach might make more sense, depending on what you actually want to achieve. In case you're after a set of default flags to be applied and ESBMC_OPTS not being usable or desirable for this purpose, there also is the widely supported `@FILE` argument approach, where FILE would contain a whitespace-separated list of options that would be parsed as usual whenever such an `@FILE` argument is encountered. They could even be nested.

That approach would at least avoid the confusing `floatbv = false` setting, which I don't think ESBMC should expose to its users.

src/util/cmdline.cpp
fbrausse: Tools usually have a list of paths to search for config files in. It could include (paths starting from) the current working directory (at the top), like clang-format, but doesn't have to. Also, system-wide settings in `/etc` are typically used whenever a config file in the user's home directory doesn't override it.

mikhailramalho:
> > > but even `false` will enable them
> > 
> > 
> > You mean false will disable it, right?
> 
> No, flags such as `--floatbv` just check for the existence of the setting with any value. This is how the default behaviour in the Boost library config parser works as far as I know.
> 
> So for example to disable floatbv in the example given above you have to remove it from the config. I'm not sure if I like this either.

this is incredibly misleading, let's find an alternative for that.

src/util/cmdline.cpp
Yiannis128: This function is a placeholder (now changed to a better name: `expand_user`) that will expand the `~` into the home directory of the user. This effectively gets replaced by `$HOME` on UNIX and `%userprofile%` on Windows.

src/util/cmdline.cpp
Yiannis128: The reason I included this one is that there might be users that would like to set a global config file from their ENV. Granted, with the inclusion of a config file, we don't have a reason to add any more env vars :)

> We don't load anything from env, all options are passed as flags to the binary. I don't mind keeping this but please add an option too.

So should the command line parameter be called `--config` with a short `-c` if it doesn't conflict with anything?

src/util/cmdline.cpp
Yiannis128: Should I add a list of locations to look for a file:

1. `--config` parameter
2. Environment variable
3. Current directory
4. User config folder? `~/.config/esbmc.toml`

Please let me know what you think about this order of execution.

src/util/cmdline.cpp
Yiannis128: This fails only if the location of the environment variable is set, in which case, it shows that the user had a clear intention to load the config file. If it is unset, it will not fail. Should I change it from this behaviour?

Yiannis128:
> While I do like the simplicity of your solution I feel that maybe a different approach might make more sense, depending on what you actually want to achieve. In case you're after a set of default flags to be applied and ESBMC_OPTS not being usable or desirable for this purpose, there also is the widely supported `@FILE` argument approach, where FILE would contain a whitespace-separated list of options that would be parsed as usual whenever such an `@FILE` argument is encountered. They could even be nested.
> 
> That approach would at least avoid the confusing `floatbv = false` setting, which I don't think ESBMC should expose to its users.

> > > > but even `false` will enable them
> > > 
> > > 
> > > You mean false will disable it, right?
> > 
> > 
> > No, flags such as `--floatbv` just check for the existence of the setting with any value. This is how the default behaviour in the Boost library config parser works as far as I know.
> > So for example to disable floatbv in the example given above you have to remove it from the config. I'm not sure if I like this either.
> 
> this is incredibly misleading, let's find an alternative for that.

I will try to change this to use a more standardized approach: if set `false` it will be deactivated, and if set `true` then it will be activated. I agree that this is not good design as well.

src/util/cmdline.cpp
Yiannis128: Done

Yiannis128:
I have looked at the source code for the boost lib and they allow you to define a custom config source. I will end up using the TOML library in the end and linking it to the config loader we have. I will need some time to figure it out.

lucasccordeiro:
@Yiannis128: Can I ask you for the status of this PR?

Yiannis128:
> @Yiannis128: Can I ask you for the status of this PR?

There has been progress. The majority of the work left now is the glue code for the TOML library and the BOOST cmdline loader.

Once that is done, the PR can be reviewed 😃

Yiannis128:
> > > > but even `false` will enable them
> > > 
> > > 
> > > You mean false will disable it, right?
> > 
> > 
> > No, flags such as `--floatbv` just check for the existence of the setting with any value. This is how the default behaviour in the Boost library config parser works as far as I know.
> > So for example to disable floatbv in the example given above you have to remove it from the config. I'm not sure if I like this either.
> 
> this is incredibly misleading, let's find an alternative for that.

OK that has been fixed now, it's now working as intended. So in this case:

```toml
interval-analysis = true
goto-unwind = true
unlimited-goto-unwind = true
k-induction = true
state-hashing = true
add-symex-value-sets = true
k-step = 2
floatbv = true
unlimited-k-steps = false
memory-leak-check = true
context-bound = 2
```

Only `unlimited-k-steps` will be set as false.

src/util/cmdline.cpp
Yiannis128: I have fixed the bug now! Only if a path is set wrong, it will fail. If a path is not set, it will try loading from the config directory, if no config file is found in that case, it will not load anything.

Yiannis128:
I found one issue and that is, when running tests during building (make). If you have the config file in a default location or the system variable is set, the unit tests will fail. Does anyone know how to fix this? Is there a macro that is set for unit tests?

Ideally, the config loading is disabled during unit tests and regression tests...

src/util/cmdline.cpp
Yiannis128: Currently, the following is done:

1. The `ESBMC_CONFIG` env is checked if set, if it is set to "" this explicitly disables config loading, if it's set to a path to a TOML config file it will use that. If it is set to a wrong location, ESBMC will exit with an error (if a path is specified, it needs to be correct).
2. If `ESBMC_CONFIG` is unset, then the config directory of the user profile is checked as a default location.

This is a nice and simple solution for users to understand and captures the following scenarios:

1. I can have a default config file in the user config directory.
2. If I want to override it I can use the env var to temporarily set it to disabled ("") or to another file.

lucasccordeiro:
@Yiannis128: did you address all comments from @mikhailramalho  and @fbrausse?

Yiannis128:
> @Yiannis128: did you address all comments from @mikhailramalho and @fbrausse?

Yes, should I resolve all the conversations that are unresolved? Everything is good to be reviewed now :) The only issue is that tests fail if you have a  config because it loads from the config file, so I wonder if there's a way to know if a test is running? https://github.com/esbmc/esbmc/pull/2006#issuecomment-2386506694

lucasccordeiro:
Thanks, @Yiannis128.

Can you change the draft status so that @mikhailramalho and @fbrausse can review it again?

src/util/config_file.cpp
```diff
@@ -0,0 +1,144 @@
+#include "config_file.h"
+
+#include <boost/program_options/option.hpp>
+#include <string>
+#include <set>
+#include <fmt/core.h>
+
+#include "lib/toml.hpp"
+
+// Define this because we cant use log_debug due to this being too early in init.
```
mikhailramalho: I don't understand this here, our log interface was not initialized at this point?

Yiannis128:
> Thanks, @Yiannis128.
> 
> Can you change the draft status so that @mikhailramalho and @fbrausse can review it again?

Yes, I will do that! Thank you

mikhailramalho (changes requested):
It looks way better, please fix the following:
1. Don't throw exceptions: in the case of optionals just return empty and if stuff fails in `cmdlint::parse`, I think returning `true` will make esbmc abort.
2. Yes, I think a new option to load config file is ok, and we should abort if we want a config file but it's not there or it's invalid

Nit: don't use curly brackets in single-line ifs

src/util/config_file.cpp
Yiannis128: I think not because `parseoptions_baset::set_verbosity_msg` is called in `esbmc_parseoptionst::doit`, `cmdlinet::parse` is called in `parseoptions_baset::parseoptions_baset` which is the constructor of the class. So using `log_debug` does not print anything even if `--verbosity 8` is set. Please correct me if I misunderstand.

Yiannis128: I've spoken to @rafaelsamenezes, and he suggested an alternative, instead I can write some unit tests for the config loader to ensure that it functions as intended. So only error messages will be kept, log messages have been removed.

Yiannis128: I've written unit tests and removed this :)

Yiannis128:
Hello, I have written unit tests for this feature, please let me know what you think!

mikhailramalho (approved):
LGTM

lucasccordeiro (approved):

Yiannis128:
@lucasccordeiro I have applied the formatting to config_file_test!

lucasccordeiro:

PR #2008, opened 2024-08-29
[C++ verification] fix builtin isnan

### LINKED ISSUE #1998: [C++ verification] New issue with ESBMC for C++: std::isnan
This issue was reported by one of our industrial partners:

````
After I updated the repo, ESBMC started failing while compiling the code. I was able to reproduce the failure on a small test.

 
Here is the content of the cpp file:
````

````CPP
#include <cstdint>
#include <cmath>


    float max(float a, float b)

    {

        assert(!std::isnan(a) && "Max<float> A was NaN!");

        assert(!std::isnan(b) && "Max<float> B was NaN!");

        return ((a > b) ? a : b);

    }


int main()

{

    float n1, n2;

 

    return max(n1, n2);

}
````
 
````
The command is here: esbmc main.cpp -Wno-writable-strings --std c++11 --verbosity clang:9 "-Wc,--gcc-install-dir=/nfs/site/itools/em64t_SLES12SP5/pkgs/gcc/12.1.0/lib/gcc/x86_64-pc-linux-gnu/12.1.0".
````
 

Here is the failure:

````
main.cpp:6:22: error: expected unqualified-id

        assert(!std::isnan(a) && "Max<float> A was NaN!");

                     ^

/usr/include/math.h:268:6: note: expanded from macro 'isnan'

     (sizeof (x) == sizeof (float)                                            \

     ^

main.cpp:7:22: error: expected unqualified-id

        assert(!std::isnan(b) && "Max<float> B was NaN!");

                     ^

/usr/include/math.h:268:6: note: expanded from macro 'isnan'

     (sizeof (x) == sizeof (float)                                            \

     ^

ERROR: PARSING ERROR
````

### PR DESCRIPTION
**Pull request by XLiZHI**
This PR tweaks the namespace for isnan.

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/github_1998/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_1998/main.cpp
@@ -1,0 +1,17 @@
+#include <cmath>
+#include <cassert>
+
+float max(float a, float b)
+
+{
+  assert(std::isnan(a) && "Max<float> A was NaN!");
+
+  assert(std::isnan(b) && "Max<float> B was NaN!");
+
+  return ((a > b) ? a : b);
+}
+int main()
+{
+  float n1 = NAN, n2 = NAN;
+  return max(n1, n2);
+}
--- a/regression/esbmc-cpp/bug_fixes/github_1998/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_1998/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/bug_fixes/github_1998_fail/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_1998_fail/main.cpp
@@ -1,0 +1,17 @@
+#include <cmath>
+#include <cassert>
+
+float max(float a, float b)
+
+{
+  assert(!std::isnan(a) && "Max<float> A was NaN!");
+
+  assert(!std::isnan(b) && "Max<float> B was NaN!");
+
+  return ((a > b) ? a : b);
+}
+int main()
+{
+  float n1, n2;
+  return max(n1, n2);
+}
--- a/regression/esbmc-cpp/bug_fixes/github_1998_fail/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_1998_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION FAILED$
--- a/src/cpp/library/cmath
+++ b/src/cpp/library/cmath
@@ -149,6 +149,8 @@ template <typename Integer>
 double modf(Integer num, double *iptr);
  */
 
+using ::__builtin_isnan;
+
 #undef CIMPORT1
 #undef CIMPORT2
 

### CODE REVIEW

**Review by lucasccordeiro — APPROVED** [review]

**Review by rafaelsamenezes — APPROVED** [review]

**Comment by lucasccordeiro** [comment]

**Comment by mikhailramalho** [comment]
I don't think this is correct, are you sure builtins part of the std namespace?

**Comment by XLiZHI** [comment]
> I don't think this is correct, are you sure builtins part of the std namespace?

In standard library `cmath`,  isnan:
```
  constexpr bool
  isnan(float __x)
  { return __builtin_isnan(__x); }
```
Do you mean such an implementation?

**Comment by mikhailramalho** [comment]
You added `__builtin_isnan` inside the `std` namespace, so this should fail but works on esbmc:
```
#include <cmath>

int main()
{
  float x;
  return std::__builtin_isnan(x);
}
```

**Comment by XLiZHI** [comment]
> You added `__builtin_isnan` inside the `std` namespace, so this should fail but works on esbmc:
> 
> ```
> #include <cmath>
> 
> int main()
> {
>   float x;
>   return std::__builtin_isnan(x);
> }
> ```

Oh, I see. here is my new solution:
```diff
diff --git a/src/cpp/library/cmath b/src/cpp/library/cmath
index b982c021d..3c85f765e 100644
--- a/src/cpp/library/cmath
+++ b/src/cpp/library/cmath
@@ -149,7 +149,22 @@ template <typename Integer>
 double modf(Integer num, double *iptr);
  */

-using ::__builtin_isnan;
+#undef isnan;
+
+bool isnan(float x)
+{
+  return __builtin_isnan(x);
+}
+
+bool isnan(double x)
+{
+  return __builtin_isnan(x);
+}
+
+bool isnan(long double x)
+{
+  return __builtin_isnan(x);
+}

 #undef CIMPORT1
 #undef CIMPORT2
```
Is that OK?

mikhailramalho:
> > You added `__builtin_isnan` inside the `std` namespace, so this should fail but works on esbmc:
> > ```
> > #include <cmath>
> > 
> > int main()
> > {
> >   float x;
> >   return std::__builtin_isnan(x);
> > }
> > ```
> 
> Oh, I see. here is my new solution:
> 
> ```diff
> diff --git a/src/cpp/library/cmath b/src/cpp/library/cmath
> index b982c021d..3c85f765e 100644
> --- a/src/cpp/library/cmath
> +++ b/src/cpp/library/cmath
> @@ -149,7 +149,22 @@ template <typename Integer>
>  double modf(Integer num, double *iptr);
>   */
> 
> -using ::__builtin_isnan;
> +#undef isnan;
> +
> +bool isnan(float x)
> +{
> +  return __builtin_isnan(x);
> +}
> +
> +bool isnan(double x)
> +{
> +  return __builtin_isnan(x);
> +}
> +
> +bool isnan(long double x)
> +{
> +  return __builtin_isnan(x);
> +}
> 
>  #undef CIMPORT1
>  #undef CIMPORT2
> ```
> 
> Is that OK?

it looks like a better solution, but why are you undefing `ìsnan`?

XLiZHI:
> it looks like a better solution, but why are you undefing ìsnan?

we include standard lib `math.h` before, it has a define: `#define isnan(x) __builtin_isnan(x)`
I think this ensures that no redefinition happens.

mikhailramalho:
> > it looks like a better solution, but why are you undefing ìsnan?
> 
> we include standard lib `math.h` before, it has a define: `#define isnan(x) __builtin_isnan(x)` I think this ensures that no redefinition happens.

yeah, that's the main issue, why is this define enabled when we run esbmc but not when compiling the file with clang?

XLiZHI:
This is because the `isnan` method is different in C++ and C.
They have a namespace difference to avoid confusion：
C++: std::isnan(x)
C: isnan(x)

In our OM, The math.h we used actually comes from the C standard library, and the one we used to compile with clang comes from the C++ standard library.

BTW, If we remove undef it still works, I just think it's safer. This is done in the standard library.

mikhailramalho:
> This is because the `isnan` method is different in C++ and C. They have a namespace difference to avoid confusion： C++: std::isnan(x) C: isnan(x)
> 
> In our OM, The math.h we used actually comes from the C standard library, and the one we used to compile with clang comes from the C++ standard library.

Including `math.h` in a cpp program should work when compiling with clang++/g++, so something is wrong on esbmc.

This is the `math.h`:
```
/* Return nonzero value if X is a NaN.  We could use `fpclassify' but
   we already have this functions `__isnan' and it is faster.  */
# if (__GNUC_PREREQ (4,4) && !defined __SUPPORT_SNAN__) \
     || __glibc_clang_prereq (2,8)
#  define isnan(x) __builtin_isnan (x)
# else
#  define isnan(x) __MATH_TG ((x), __isnan, (x))
# endif
```

My question is why when running esbmc, the first define is selected, but not the second, like what happens when we just call clang++. ESBMC should have the same behaviour,

> 
> BTW, If we remove undef it still works, I just think it's safer. This is done in the standard library.

Is it? My system's cmath doesn't have any undef, where is this undef located?

XLiZHI:
> Is it? My system's cmath doesn't have any undef, where is this undef located?

in c++/11/ cmath:


> My question is why when running esbmc, the first define is selected, but not the second, like what happens when we just call clang++. ESBMC should have the same behaviour,

In the common sense, `std::isnan` can't compile with` math.h`, so I think the clang/gcc compiler has some compatibility optimizations. 
image

mikhailramalho:
What do you mean by common sense? This compiles just fine:
```
#include <math.h>

int main()
{
  float x;
  return std::isnan(x);
}
```
esbmc should be able to verify it as well

XLiZHI:
> What do you mean by common sense? This compiles just fine:
> ```
> #include <math.h>
> 
> int main()
> {
>   float x;
>   return std::isnan(x);
> }
> ```
> esbmc should be able to verify it as well

There are two versions of math.h, in the C++ version it contains the header cmath (see screenshot from the previous comment). This is why clang++/g++ compiles successfully.

In ESBMC, we dont have std::isnan in cmath.

PR #2011, opened 2024-08-31
[clang-cpp] fix recursive CXXMemberCalls, fixes #63

### LINKED ISSUE #63: Segmentation fault (core dumped)
ESBMC crashes in this particular C++ example:
```
[lucascordeiro@localhost src]$ esbmc binary-tree.cpp -I ~/esbmc-cpp-linux-64-static/libraries/
ESBMC version 3.0.0 64-bit x86_64 linux
file binary-tree.cpp: Parsing
Converting
Type-checking binary-tree.cpp
# 
# include <iostream>
# include <cassert>

using namespace std;

template <class T>
class nodet
{
public:
    T key;
    nodet *p;
    nodet *left;
    nodet *right;
};

template <class T>
class treet
{
public:
    treet();
    void preorder_tree_walk(nodet<T> *x);
    void inorder_tree_walk(nodet<T> *x);
    void posorder_tree_walk(nodet<T> *x);
    void insert(nodet<T> *z, T key);
    nodet<T> *tree_search(nodet<T> *x, T k);
    nodet<T> *tree_minimum(nodet<T> *x);
    nodet<T> *tree_maximum(nodet<T> *x);
    nodet<T> *get_root();
private:
    nodet<T> *root;
};

template <class T>
treet<T>::treet()
{
    root = NULL;
}

template <class T>
void treet<T>::preorder_tree_walk(nodet<T> *x)
{
  if (x!=NULL)
  {
    cout << "node: " << x->key << endl;
    preorder_tree_walk(x->left);
    preorder_tree_walk(x->right);
  }
}

template <class T>
void treet<T>::inorder_tree_walk(nodet<T> *x)
{
  if (x!=NULL)
  {
    inorder_tree_walk(x->left);
    cout << "node: " << x->key << endl;
    inorder_tree_walk(x->right);
  }
}

template <class T>
void treet<T>::posorder_tree_walk(nodet<T> *x)
{
  if (x!=NULL)
  {
    posorder_tree_walk(x->left);
    posorder_tree_walk(x->right);
    cout << "node: " << x->key << endl;
  }
}

template <class T>
void treet<T>::insert(nodet<T> *z, T key)
{
  nodet<T> *y, *x;
  z = new nodet<T>;
  z->key=key; z->left=NULL; z->right=NULL; y = NULL; x=root;
  while (x != NULL)
  {
    y = x;
    if (z->key < x->key) x = x->left;
    else x = x->right;
  }

  z->p = y;
  if (y==NULL) {root = z; root->left=NULL; root->right=NULL;}
  else if (z->key < y->key) y->left = z;
  else if (z->key > y->key) y->right = z;
}

template <class T>
nodet<T> *treet<T>::tree_search(nodet<T> *x, T k)
{
  if (x == NULL || k == x->key)
    return x;
  if (k < x->key)
    tree_search(x->left, k);
  else
    tree_search(x->right,k);
}

template <class T>
nodet<T> *treet<T>::tree_minimum(nodet<T> *x)
{
  while (x->left != NULL)
    x=x->left;
  return x;
}

template <class T>
nodet<T> *treet<T>::tree_maximum(nodet<T> *x)
{
  while (x->right != NULL)
    x=x->right;
  return x;
}

template <class T>
nodet<T> *treet<T>::get_root()
{
    return root;
}

int main(void)
{
  nodet<char> *t = new nodet<char>;
  treet<char> *n = new treet<char>;
  n->insert(t, 'F');
  n->insert(t, 'C');
  n->insert(t, 'E');
  n->insert(t, 'L');
  n->insert(t, 'G');
  n->insert(t, 'A');
  n->insert(t, 'B');
  n->insert(t, 'I');
  n->insert(t, 'H');
  n->insert(t, 'J');

  cout << "### inorder_tree_walk: " << endl;
  n->inorder_tree_walk(n->get_root());

  cout << "### preorder_tree_walk: " << endl;
  n->preorder_tree_walk(n->get_root());

  cout << "### posorder_tree_walk: " << endl;
  n->posorder_tree_walk(n->get_root());

  t = n->tree_search(n->get_root(),'A');
  assert(t->key == 'A');

  cout << "### tree_search: " << 'A' << "==" << t->key << endl;
  t = n->tree_minimum(n->get_root());
  assert(t->key == 'A');

  cout << "### tree_minimum: " << t->key << endl;
  t = n->tree_maximum(n->get_root());
  assert(t->key == 'L');
  cout << "### tree_maximum: " << t->key << endl;

  return 0;
}
```

intrigus-lgtm:


intrigus-lgtm:
This is the best solution I could come up with.
Please let me know if you think there is a simpler solution.

mikhailramalho:
We have to understand why this recursive call crashes esbmc when it's a member function but not when it's just a function.

intrigus-lgtm:
> We have to understand why this recursive call crashes esbmc when it's a member function but not when it's just a function.



> We have to understand why this recursive call crashes esbmc when it's a member function but not when it's just a function.

When it's just a function, the recursive call is represented as a DeclRefExpr.

Source:
```
void b() {
    b();
}
```

AST
```
`-FunctionDecl <line:1:1, line:3:1> line:1:6 referenced b 'void ()'
  `-CompoundStmt <col:10, line:3:1>
    `-CallExpr <line:2:5, col:7> 'void'
      `-ImplicitCastExpr <col:5> 'void (*)()' <FunctionToPointerDecay>
        `-DeclRefExpr <col:5> 'void ()' lvalue Function 0xb960de0 'b' 'void ()'
```

When it's a member function, it's as described here:
https://github.com/esbmc/esbmc/pull/2011/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590R703-R712

mikhailramalho:
This is not the correct explanation why the bug is happening. 
You have to understand the code path that's causing the infinite loop.

intrigus-lgtm:
> This is not the correct explanation why the bug is happening. You have to understand the code path that's causing the infinite loop.

What is lacking in [this explanation](https://github.com/esbmc/esbmc/pull/2011/files#diff-77a14a166a14947af9a9605e058ed20d935cf5200891234187cf015166925590R703-R712)?

lucasccordeiro:
@mikhailramalho and @intrigus-lgtm: It would be great if we could advance with this PR.

mikhailramalho:
The conclusion of this explanation is wrong, as is this solution. 

Recapping, when we are parsing:
```
CXXMethodDecl 0x555556afa5c0 <foo.cpp:5:3, line:10:3> line:5:7 used b 'int (int)' implicit-inline
|-ParmVarDecl 0x555556afa4c0 <col:9, col:13> col:13 used i 'int'
`-CompoundStmt 0x555556affb88 <line:6:3, line:10:3>
  |-IfStmt 0x555556afa748 <line:7:5, line:8:14>
  | |-BinaryOperator 0x555556afa6e0 <line:7:9, col:14> 'bool' '=='
  | | |-ImplicitCastExpr 0x555556afa6c8 <col:9> 'int' <LValueToRValue>
  | | | `-DeclRefExpr 0x555556afa688 <col:9> 'int' lvalue ParmVar 0x555556afa4c0 'i' 'int'
  | | `-IntegerLiteral 0x555556afa6a8 <col:14> 'int' 0
  | `-ReturnStmt 0x555556afa738 <line:8:7, col:14>
  |   `-ImplicitCastExpr 0x555556afa720 <col:14> 'int' <LValueToRValue>
  |     `-DeclRefExpr 0x555556afa700 <col:14> 'int' lvalue ParmVar 0x555556afa4c0 'i' 'int'
  `-ReturnStmt 0x555556affb78 <line:9:5, col:23>
    `-BinaryOperator 0x555556affb58 <col:12, col:23> 'int' '+'
      |-ImplicitCastExpr 0x555556affb40 <col:12> 'int' <LValueToRValue>
      | `-DeclRefExpr 0x555556afa768 <col:12> 'int' lvalue ParmVar 0x555556afa4c0 'i' 'int'
      `-CXXMemberCallExpr 0x555556afa840 <col:16, col:23> 'int'
        |-MemberExpr 0x555556afa798 <col:16> '<bound member function type>' ->b 0x555556afa5c0
        | `-CXXThisExpr 0x555556afa788 <col:16> 'a *' implicit this
        `-BinaryOperator 0x555556afa820 <col:18, col:22> 'int' '-'
          |-ImplicitCastExpr 0x555556afa808 <col:18> 'int' <LValueToRValue>
          | `-DeclRefExpr 0x555556afa7c8 <col:18> 'int' lvalue ParmVar 0x555556afa4c0 'i' 'int'
          `-IntegerLiteral 0x555556afa7e8 <col:22> 'int' 1
```
In particular:
```
MemberExpr 0x555556afa798 '<bound member function type>' ->b 0x555556afa5c0
`-CXXThisExpr 0x555556afa788 'struct a *' implicit this
```
We try to get the member declaration:
```
      exprt comp;
      if (get_decl(*member.getMemberDecl(), comp))
        return true;
```
which is the recursive function.

Now, look at how we use `comp`:
```
      exprt comp;
      if (get_decl(*member.getMemberDecl(), comp))
        return true;

      if (!is_member_decl_static(member))
      {
        exprt base;
        if (get_expr(*member.getBase(), base))
          return true;

        assert(!comp.name().empty());
        // for MemberExpr referring to struct field (or method in case of C++ class)
        new_expr = member_exprt(base, comp.name(), comp.type());
      }
      else
      {
        // for static members, use the member decl symbol directly
        // without making a member_exprt, e.g.
        // If the member_exprt refers to a class static member, then
        // replace "OBJECT.MyStatic = 1" with "MyStatic = 1;"
        assert(comp.statement() == "decl");
        assert(comp.op0().is_symbol());
        new_expr = comp.op0();
      }
```
We need it's `name()`, `type()` and `op0()`, which from the comments seems to be the member symbol.

To fix the issue, you need to get the type via `get_type`, the name probably via `get_decl_name(member.getMemberDecl(), foo, bar)` (won't trigger the recursion). The symbol I can't remember from the top of my head, I suggest creating another example without the recursion and checking what's `.op0()` exactly.

intrigus-lgtm:
> We need it's `name()`, `type()` and `op0()`, which from the comments seems to be the member symbol.
> 
> To fix the issue, you need to get the type via `get_type`, the name probably via `get_decl_name(member.getMemberDecl(), foo, bar)` (won't trigger the recursion). The symbol I can't remember from the top of my head, I suggest creating another example without the recursion and checking what's `.op0()` exactly.

Thanks for looking into this, but this is also not a perfect solution I'd say:

Compare `get_type(*member.getMemberDecl()->getType(), comp_type)`
```
code
  * arguments: 
    0: argument
        * type: signedbv
            * width: 32
            * #cpp_type: signed_int
  * return_type: signedbv
      * width: 32
      * #cpp_type: signed_int
```
vs.
`comp.type()`
```
code
  * arguments: 
    0: argument
        * type: pointer
            * subtype: symbol
                * identifier: tag-a
        * #base_name: this
        * #identifier: c:@S@a@F@c_wtf_hashtag_I_wtf_hashtag_this
    1: argument
        * type: signedbv
            * width: 32
            * #cpp_type: signed_int
        * #location: 
          * file: main.cpp
          * line: 6
          * function: c
          * column: 10
        * #base_name: i::0
        * #identifier: c:main.cpp@58@S@a@F@c_wtf_hashtag_I_wtf_hashtag_@i::0
  * return_type: signedbv
      * width: 32
      * #cpp_type: signed_int
  * #inlined: 1
  * #member_name: tag-a
```

(Ignore the `_wtf_hashtag_` in the identifier)

mikhailramalho:
I had some time to look into it: the only thing missing is the "this" pointer in the first argument, which we can get from `clang_cpp_convertert::get_function_this_pointer_param`.

Annoyingly this function is only available in the clang_cpp_convert, so we need to do something like `get_struct_union_class_methods_decls` where the c frontend implementation is empty.

I also noticed some code duplication when building the function type in `get_method`/`get_function` and building it via the `clang::Type::FunctionProto` in the `get_type` switch. We should fix that in a separate PR.

lucasccordeiro:
Thanks for your feedback, @mikhailramalho.

@intrigus-lgtm: can I ask you for the status of this PR?

mikhailramalho:
I'll have time next week to work on this, if @intrigus-lgtm is not available.

intrigus-lgtm:
> I'll have time next week to work on this, if @intrigus-lgtm is not available.

Thank you, that would be of great help.

mikhailramalho:
This patch fixes the issue:
```
From 66e323685bee0bce387b0d3ac73a2060486b6620 Mon Sep 17 00:00:00 2001
From: "Mikhail R. Gadelha" <mikhail@igalia.com>
Date: Wed, 11 Dec 2024 13:22:34 -0300
Subject: [PATCH] foo

Signed-off-by: Mikhail R. Gadelha <mikhail@igalia.com>
---
 src/clang-c-frontend/clang_c_convert.cpp   | 35 +++++++++++++++++-----
 src/clang-c-frontend/clang_c_convert.h     |  6 ++++
 src/clang-cpp-frontend/clang_cpp_convert.h |  2 +-
 3 files changed, 35 insertions(+), 8 deletions(-)

index 30e2a130f..52451a9da 100644
@@ -1835,19 +1835,32 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
 
     if (!perform_virtual_dispatch(member))
     {
-      exprt comp;
-      if (get_decl(*member.getMemberDecl(), comp))
+      typet comp_type;
+      if (get_type(*member.getMemberDecl()->getType(), comp_type))
         return true;
 
+      // Add this parameter if this is a c++ method
+      if (
+        const auto *cxxmd =
+          llvm::dyn_cast<clang::CXXMethodDecl>(member.getMemberDecl()))
+      {
+        if (get_function_this_pointer_param(
+              *cxxmd, to_code_type(comp_type).arguments()))
+          return true;
+      }
+
+      std::string id, name;
+      get_decl_name(*member.getMemberDecl(), name, id);
+
       if (!is_member_decl_static(member))
       {
         exprt base;
         if (get_expr(*member.getBase(), base))
           return true;
 
-        assert(!comp.name().empty());
+        assert(!name.empty());
         // for MemberExpr referring to struct field (or method in case of C++ class)
-        new_expr = member_exprt(base, comp.name(), comp.type());
+        new_expr = member_exprt(base, id, comp_type);
       }
       else
       {
@@ -1855,9 +1868,9 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
         // without making a member_exprt, e.g.
         // If the member_exprt refers to a class static member, then
         // replace "OBJECT.MyStatic = 1" with "MyStatic = 1;"
-        assert(comp.statement() == "decl");
-        assert(comp.op0().is_symbol());
-        new_expr = comp.op0();
+        // assert(comp.statement() == "decl");
+        // assert(comp.op0().is_symbol());
+        // new_expr = comp.op0();
       }
     }
     else
@@ -3835,6 +3848,14 @@ bool clang_c_convertert::is_fd_virtual_or_overriding(
   return false;
 }
 
+bool clang_c_convertert::get_function_this_pointer_param(
+  const clang::CXXMethodDecl &,
+  code_typet::argumentst &)
+{
+  // It just can't happen in C
+  return false;
+}
+
 bool clang_c_convertert::get_vft_binding_expr(
   const clang::MemberExpr &,
   exprt &)
diff --git a/src/clang-c-frontend/clang_c_convert.h b/src/clang-c-frontend/clang_c_convert.h
index 1d4c0e0dc..84d97d71c 100644
--- a/src/clang-c-frontend/clang_c_convert.h
+++ b/src/clang-c-frontend/clang_c_convert.h
@@ -42,6 +42,7 @@ class FieldDecl;
 class MemberExpr;
 class EnumConstantDecl;
 class APValue;
+class CXXMethodDecl;
 } // namespace clang
 
 std::string
@@ -122,6 +123,11 @@ protected:
   virtual bool get_function_params(
     const clang::FunctionDecl &fd,
     code_typet::argumentst &params);
+
+  virtual bool get_function_this_pointer_param(
+    const clang::CXXMethodDecl &fd,
+    code_typet::argumentst &params);
+
   /**
    *  Parse each individual parameter of the function
    */
diff --git a/src/clang-cpp-frontend/clang_cpp_convert.h b/src/clang-cpp-frontend/clang_cpp_convert.h
index e3c3c7c26..951f8fe75 100644
--- a/src/clang-cpp-frontend/clang_cpp_convert.h
+++ b/src/clang-cpp-frontend/clang_cpp_convert.h
@@ -78,7 +78,7 @@ protected:
    */
   bool get_function_this_pointer_param(
     const clang::CXXMethodDecl &fd,
-    code_typet::argumentst &params);
+    code_typet::argumentst &params) override;
 
   bool get_struct_union_class(const clang::RecordDecl &rd) override;
 
-- 
2.43.0

mikhailramalho:
however, I still don't like it. Especially handling C++ stuff on the C frontend; maybe a get_member_expr would be better

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2193.

PR #2014, opened 2024-09-02
[regression] Infinite loop (?) on symbolic execution

### LINKED ISSUE #65: Infinite loop (?) on symbolic execution
OBS: the new llvm frontend is required in order to reproduce this error.

The program in [1] causes esbmc to hang during symbolic execution. Maybe an infinite loop.

Symex trace shows:

```
    num_total_threads=0;
    num_threads_running=0;
    __ESBMC_deallocated=ARRAY_OF(FALSE);
    __ESBMC_alloc=ARRAY_OF(FALSE);
    p=&{ 2, 4 }[0];
    __ESBMC_alloc_size=ARRAY_OF(0);
    pc=&{ 1f, 10f, 100f }[0];
    __ESBMC_is_dynamic=ARRAY_OF(FALSE);
    FUNCTION_CALL:  pthread_start_main_hook()
    ATOMIC_BEGIN
    num_total_threads=num_total_threads + 1;
    num_threads_running=num_threads_running + 1;
    ATOMIC_END
    END_FUNCTION
    FUNCTION_CALL:  main()
    signed int n;
    n=2;
    signed int * p;
    p=&n;
    p=&{ *p, 0 }[0];
```

lucasccordeiro:
This PR should close https://github.com/esbmc/esbmc/issues/65.


mikhailramalho (changes requested):
we need an assertion on the test case to make sure the right behaviour is implemented.

lucasccordeiro:
> we need an assertion on the test case to make sure the right behaviour is implemented.

Fair point. Implemented as suggested.

regression/esbmc/github_65/main.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+
+int *p = (int[]){2, 4}; // creates an unnamed static array of type int[2]
+                        // initializes the array to the values {2, 4}
+                        // creates pointer p to point at the first element of the array
+const float *pc = (const float []){1e0, 1e1, 1e2}; // read-only compound literal
+ 
+int main(void)
+{
+    int n = 2, *p = &n;
```
mikhailramalho: Can you change the value of `n`? When p is first created, the first element of the array is also 2.

intrigus-lgtm (approved):

regression/esbmc/github_65/main.c
lucasccordeiro: Sure. Implemented as suggested.

PR #2017, opened 2024-09-05
[python-frontend] Fix list indexing and inference

brcfarias:
- When invoking ESBMC with the `--function` flag, the called functions are now identified and converted.
```
def func1() -> int:
  return 0

def func2() -> int:
  return func1()

def func3() -> int:
  return 1
```
Runnnig `esbmc file.py --function func2` will convert func2 and func1 (since func1() is called by func2()).

- Variables initialized from functions that returns a _**list**_ are now correctly inferred.
```
def make_list: list[int]
  l = [1,2,3]
  return l

l1 = make_list()  # l1 should now be inferred as list[int]
```

- Previously, **_lists_** created from function returns could not be indexed.

### DIFF
--- a/regression/python/function-option-fail/main.py
+++ b/regression/python/function-option-fail/main.py
@@ -1,0 +1,7 @@
+def ok() -> None:
+    assert(1)
+
+def fail() -> None:
+    assert(0)
+
+ok()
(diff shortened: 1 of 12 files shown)


lucasccordeiro (approved):

lucasccordeiro:

Issue #1967, opened 2024-08-06
[Python] ERROR: PARSING ERROR


lucasccordeiro:
Bug report from a ESBMC-Python user:
 
````
I’ve tried on a simple python file as

                esbmc.exe cpufreqs_catcher.py

This is the return

                ESBMC version 7.6.1 64-bit x86_64 windows

Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc

Parsing cpufreqs_catcher.py

  File "....\esbmc-python-astgen-943e-9ce3-9b0b\parser.py", line 26

    print(f"Error: Module '{module_name}' not found.")

                                                    ^

SyntaxError: invalid syntax

ERROR: Python execution failed

ERROR: PARSING ERROR
````
 
_"Per default we need Python 2.7 installed on our PCs but I also have Python 3 if needed. Just let me know how do I ask ESBMC to use the Py3 path and I can test it."_

brcfarias:
Could you please provide the program that caused the issue?

~~The problem occurred when trying to process an import [here](https://github.com/esbmc/esbmc/blob/9a25e56e07f8edf5cc45d4bcf77e7dab7a2ecbc3/src/python-frontend/parser.py#L26)~~

I'm not sure if the Python version could make a difference here, but we just need python3 to be visible to the user (e.g. available in one of folders in $PATH)

lucasccordeiro:
Here is the Python code that causes this error:

````Python
import os

 

while True:

    stream = os.popen('adb devices -l')

    out = stream.read()

    lines = out.split("\n")[1:]

 

    print 'Devices Available:'

    devices = []

    for item in lines:

        items = item.split()

        if len(items) < 2:

            continue

        devices.append(items[0])

        print items[0] + " -> " + items[2].replace("product:", "")

 

    print '\n'

    device_id = raw_input( 'Select one by typing its id (0,1,2,3...): ')

    print '\n'

 

    with open(devices[int(device_id)] + '-cpufreqs.txt', 'w') as cpufile:

        for id in range(10):

            stream = os.popen('adb -s ' + devices[int(device_id)] + ' shell cat /sys/devices/system/cpu/cpufreq/policy'+ str(id) +'/scaling_available_frequencies')

            output = 'CPU ' + str(id) + ': ' + stream.read()

            print output

            cpufile.write(output)

````

brcfarias:
[...]

PR #2020, opened 2024-09-09
[python-frontend] import os

### LINKED ISSUE #2013: [Python Frontend] Support for OS library
We should add a skeleton of the OS library (https://docs.python.org/3/library/os.html) so ESBMC can verify Python programs that rely on this lib. We could have just the method declaration to avoid crashing ESBMC. We could print something like "WARNING: no body for method XXX".

brcfarias:
- Adding stubs for the following `os` functions: `popen`, `listdir`, `makedirs`, `remove`, `path.exists`, `path.basename`
- Implemented logic for loading these modules.

- In Python, it's possible to import a module and access its submodules or functions without explicitly importing them
```
import os
os.path.exists()  #path is an os submodule
```

To address this, I check the AST to detect calls to submodules and implicitly 'import' them when necessary.

- I plan to use the libpython (https://docs.python.org/3/c-api/index.html) to manage module imports, as handling these internally can be complex.

Fixes #2013


lucasccordeiro (changes requested):
Can you add the test from this issue https://github.com/esbmc/esbmc/issues/1967?

lucasccordeiro:

PR #2025, opened 2024-09-11
[Concurrency] Avoid forced switching in built-in library

XLiZHI:
```

#include <pthread.h>

void* t1(void *arg) {
    int* shared1 = (int*) arg;

    return NULL;
}

void* t2(void *arg) {
    int* shared2 = (int*) arg;

    return NULL;
}

int main() {
    pthread_t id1, id2;
    int shared_data = 0;

    pthread_join(id1, NULL);
    pthread_join(id2, NULL);

    return 0;
}
```

Master: Thread interleavings 221

This PR: Thread interleavings 7

### DIFF
--- a/regression/esbmc-unix/01_cond_01/test.desc
+++ b/regression/esbmc-unix/01_cond_01/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --no-unwinding-assertions --unwind 3 --lock-order-check --context-bound 3 --smt-during-symex --smt-thread-guard --z3
 ^VERIFICATION SUCCESSFUL$
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -46,7 +46,7 @@ public:
     const exprt &original_expr,
     bool deref)
   {
-    if (deref)
+    if (deref || original_expr.is_member())
     {
       // introduce a new expression: RACE_CHECK(&x)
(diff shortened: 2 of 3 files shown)


lucasccordeiro:
Many thanks, @XLiZHI. 

Did you try to investigate these improvements by changing our operational models?

XLiZHI:
> Many thanks, @XLiZHI.
> 
> Did you try to investigate these improvements by changing our operational models?
> 
> Can you run this PR over the SV-COMP benchmarks?

__ESBMC_yield() was added during the conversion in the frontend, and I think the way to modify the OM is to use a label (like __ESBMC_hide:) on __ESBMC_begin(). However, I found that we currently have limited use of label (just for hide), and I thought this might require additional extensions.

At the moment our Benchexec seems to be down, but I've run it before and here's what we get:
This PR:
```
  correct:           14207
    correct true:     7810
    correct false:    6397
  incorrect:            19
    incorrect true:      8
    incorrect false:    11
  unknown:            9579
  Score:             21585 (max: 38482)
  
  https://github.com/esbmc/esbmc/actions/runs/10547846850/job/29221237224
```
Master:

```
  correct:           14146
    correct true:     7751
    correct false:    6395
  incorrect:            20
    incorrect true:      9
    incorrect false:    11
  unknown:            9639
  Score:             21433 (max: 38482)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/10610362531
```

lucasccordeiro (approved):

PR #2037, opened 2024-09-20
[C++ verification] Fix operator new

### LINKED ISSUE #2004: `operator new` fails.
```cpp
#include <cstdlib>

struct aaaa {
  int a;
  int b;
};

int main()
{
  void *foo = operator new(8);
  // void *bar = new char;
  // void *aaa = malloc(sizeof(char));
  // void *bbb = new aaaa;
}
```

fails with:
```
terminate called after throwing an instance of 'type2t::symbolic_type_excp'
```

This is because the generated GOTO looks like this:

<details>

```cpp
main (c:@F@main#):
        // 534 file cpptest.pre.cpp line 10 column 3 function main
        void * foo;
        // 535 file cpptest.pre.cpp line 10 column 15 function main
        void * return_value$_operator new$1;
        // 536 file cpptest.pre.cpp line 10 column 15 function main
        return_value$_operator new$1=new void;
        // 537 no location
        *return_value$_operator new$1=nil;
```
</details>

We are trying to dereference a `void` pointer here which can only fail (`*return_value$_operator new$1=nil;`)!

What is the reason for implementing `operator new` in terms of `new T`:
https://github.com/esbmc/esbmc/blob/80c4e3b48bbd7bdf7ceb7b58d2458e68ddc917f7/src/goto-programs/builtin_functions.cpp#L739-L752

I would have expected it to be the other way around:
`new T` would internally call `operator new`.

Anyway, in my locals tests, replacing everything (L739-L752) with just `do_malloc(lhs, function, arguments, dest);` seemed to work.

XLiZHI:
1. No initialization is required once memory is allocated using operator new.
2. ~~According to the handling of void* in malloc, introduce a new statement: `op_new`.~~

~~We can't simply use `do_malloc`, although the operation is similar, because in C verification we need to use `--force-malloc-success` to ensure that the memory is allocated successfully. This is not required in C++ verification.~~

### DIFF
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -289,11 +289,18 @@ void goto_convertt::do_cpp_new(
 
   // grab initializer
   goto_programt tmp_initializer;
-  cpp_new_initializer(lhs, rhs, tmp_initializer);
+  if (!rhs.get_bool("op_new"))
+    cpp_new_initializer(lhs, rhs, tmp_initializer);
 
   exprt alloc_size;
+  exprt alloc_type;
 
-  if (rhs.statement() == "cpp_new[]")
+  if (rhs.get_bool("op_new"))
+  {
+    alloc_type = static_cast<const exprt &>(rhs.find("alloctype"));
+    alloc_size = static_cast<const exprt &>(rhs.size_irep());
+  }
+  else if (rhs.statement() == "cpp_new[]")
   {
     alloc_size = static_cast<const exprt &>(rhs.size_irep());
     if (alloc_size.type() != size_type())
@@ -326,6 +333,7 @@ void goto_convertt::do_cpp_new(
   exprt new_expr("sideeffect", rhs.type());
   new_expr.statement(rhs.statement());
   new_expr.cmt_size(alloc_size);
+  new_expr.cmt_type(alloc_type);
   new_expr.location() = rhs.find_location();
 
   // produce new object
@@ -744,10 +752,16 @@ void goto_convertt::do_function_call_symbol(
   {
     assert(arguments.size() == 1);
 
+    typet alloc_type;
+    exprt alloc_size;
+    get_alloc_type(arguments[0], alloc_type, alloc_size);
+
     // Change it into a cpp_new expression
-    side_effect_exprt new_function("cpp_new");
+    side_effect_exprt new_function("cpp_new[]");
     new_function.add("#location") = function.cmt_location();
-    new_function.add("sizeof") = arguments.front();
+    new_function.add("size") = arguments.front();
+    new_function.add("alloctype") = alloc_type;
+    new_function.set("op_new", true);
 
     // Set return type, a allocated pointer
     // XXX jmorse, const-qual misery
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -733,7 +733,7 @@ void goto_convertt::do_function_call_symbol(
     exprt alloc_size;
     get_alloc_type(arguments[0], alloc_type, alloc_size);
 
-    // Change it into a cpp_new expression
+    // Change it into a cpp_new[] expression
     side_effect_exprt new_function("cpp_new[]");
     new_function.add("#location") = function.cmt_location();
     new_function.add("size") = arguments.front();
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -289,8 +289,7 @@ void goto_convertt::do_cpp_new(
 
   // grab initializer
   goto_programt tmp_initializer;
-  if (!rhs.get_bool("op_new"))
-    cpp_new_initializer(lhs, rhs, tmp_initializer);
+  cpp_new_initializer(lhs, rhs, tmp_initializer);
 
   exprt alloc_size;
   exprt alloc_type;
@@ -352,11 +351,17 @@ void goto_convertt::cpp_new_initializer(
   goto_programt &dest)
 {
   // grab initializer
-  code_expressiont initializer;
+  code_expressiont initializer = (code_expressiont &)rhs.initializer();
 
-  if (rhs.initializer().is_nil())
+  if (initializer.is_nil())
+    return;
+
+  if (!initializer.op0().get_bool("constructor"))
   {
-    // Initialize with default value
+    // for auto *p = Foo(3) and int *p = 3
+    // constructor case:  init is Foo(&(*new_object), 3)
+    // other case: init is 3,
+    // we turn "3" into "*new_object = 3"
     side_effect_exprt assignment("assign");
     assignment.type() = rhs.type().subtype();
 
@@ -365,54 +370,26 @@ void goto_convertt::cpp_new_initializer(
     new_object.set("#lvalue", true);
     new_object.type() = rhs.type().subtype();
 
-    // Default value is zero
-    exprt default_value = gen_zero(rhs.type().subtype());
-
-    assignment.move_to_operands(new_object, default_value);
+    assignment.move_to_operands(new_object, initializer.op0());
     initializer.expression() = assignment;
   }
+
+  // XXX jmorse, const-qual misery
+  const_cast<exprt &>(rhs).remove("initializer");
+
+  if (rhs.statement() == "cpp_new[]")
+  {
+    // build loop
+  }
+  else if (rhs.statement() == "cpp_new")
+  {
+    exprt deref_new("dereference", rhs.type().subtype());
+    deref_new.copy_to_operands(lhs);
+    replace_new_object(deref_new, initializer);
+    convert(to_code(initializer), dest);
+  }
   else
-  {
-    initializer = (code_expressiont &)rhs.initializer();
-
-    if (!initializer.op0().get_bool("constructor"))
-    {
-      // for auto *p = Foo(3) and int *p = 3
-      // constructor case:  init is Foo(&(*new_object), 3)
-      // other case: init is 3,
-      // we turn "3" into "*new_object = 3"
-      side_effect_exprt assignment("assign");
-      assignment.type() = rhs.type().subtype();
-
-      // the new object
-      exprt new_object("new_object");
-      new_object.set("#lvalue", true);
-      new_object.type() = rhs.type().subtype();
-
-      assignment.move_to_operands(new_object, initializer.op0());
-      initializer.expression() = assignment;
-    }
-
-    // XXX jmorse, const-qual misery
-    const_cast<exprt &>(rhs).remove("initializer");
-  }
-
-  if (initializer.is_not_nil())
-  {
-    if (rhs.statement() == "cpp_new[]")
-    {
-      // build loop
-    }
-    else if (rhs.statement() == "cpp_new")
-    {
-      exprt deref_new("dereference", rhs.type().subtype());
-      deref_new.copy_to_operands(lhs);
-      replace_new_object(deref_new, initializer);
-      convert(to_code(initializer), dest);
-    }
-    else
-      assert(0);
-  }
+    assert(0);
 }
 
 void goto_convertt::do_exit(
--- a/regression/esbmc-cpp/bug_fixes/github_2004/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_2004/main.cpp
@@ -1,0 +1,16 @@
+#include <cassert>
+#include <new>
+
+int main()
+{
+  void *foo = operator new(sizeof(int));
+
+  int *intPtr = static_cast<int *>(foo);
+  *intPtr = 42;
+
+  assert(*intPtr == 42);
+
+  operator delete(foo);
+
+  return 0;
+}
--- a/regression/esbmc-cpp/bug_fixes/github_2004/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_2004/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/bug_fixes/github_2004_fail/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_2004_fail/main.cpp
@@ -1,0 +1,18 @@
+#include <cassert>
+#include <new>
+
+int main()
+{
+  void *foo = operator new(sizeof(int));
+
+  // allocated memory is not enough
+  // int 4; double 8
+  double *intPtr = static_cast<double *>(foo);
+  *intPtr = 42;
+
+  assert(*intPtr == 42);
+
+  operator delete(foo);
+
+  return 0;
+}
--- a/regression/esbmc-cpp/bug_fixes/github_2004_fail/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_2004_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/cpp/operator_new/main.cpp
+++ b/regression/esbmc-cpp/cpp/operator_new/main.cpp
@@ -11,7 +11,6 @@ int main()
 {
 	dev_a = (int*) operator new(sizeof(int));
 	assert(dev_a != NULL);
-	assert(*dev_a == 0);
 
 	return 0;
 }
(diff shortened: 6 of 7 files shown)


lucasccordeiro (approved):

mikhailramalho (changes requested):
Why do you need to introduce a new id for the new operator? Can't you just fix the current one?

Also, the new operator can fail, just like malloc, so it should behave like that, including having to use the force-malloc-success flag to make it always succeed.

XLiZHI:
> Why do you need to introduce a new id for the new operator? Can't you just fix the current one?
> 
> Also, the new operator can fail, just like malloc, so it should behave like that, including having to use the force-malloc-success flag to make it always succeed.

Do you agree with using `do_malloc` to handle operator new?

mikhailramalho:
> > Why do you need to introduce a new id for the new operator? Can't you just fix the current one?
> > Also, the new operator can fail, just like malloc, so it should behave like that, including having to use the force-malloc-success flag to make it always succeed.
> 
> Do you agree with using `do_malloc` to handle operator new?

Why is it a problem?

XLiZHI:
> > > Why do you need to introduce a new id for the new operator? Can't you just fix the current one?
> > > Also, the new operator can fail, just like malloc, so it should behave like that, including having to use the force-malloc-success flag to make it always succeed.
> > 
> > 
> > Do you agree with using `do_malloc` to handle operator new?
> 
> Why is it a problem?

My idea is that we can avoid additional processing if we use `do_malloc` to handle op_new because they do the same thing, just as @intrigus-lgtm did before.

If we want to use `do_cpp_new` to implement this, we need to distinguish between these three: `cpp_new`, `cpp_new[]` and  `op_new`, they have some differences (For example, op_new does not need to be initialized), so I introduced op_new to distinguish them better, it makes the judgement easier. Also, we might need to introduce new flags to handle it, like `--force-opnew-success`?

mikhailramalho:
But why do you need op_new? Is there a case, system, arch, etc where new doesn't initialize?

XLiZHI:
During symbolic execution, we determine `do_array` according to whether it is `cpp_new_arr`, in the function executing cpp_new. For op_new I think we should create dynamic arrays. I think a new statement is convenient to judge this situation. Also, because type expressions of void types have an empty subtype, this requires a different adjustment than `cpp_new_arr`.

Of course op_new is not necessary, I can remove it, but the judgment will be complicated.

> Is there a case, system, arch, etc where new doesn't initialize?

Maybe this code fits what you said:  `int* p = new int;` my previous expression is not accurate.

mikhailramalho:
> During symbolic execution, we determine `do_array` according to whether it is `cpp_new_arr`, in the function executing cpp_new. For op_new I think we should create dynamic arrays. I think a new statement is convenient to judge this situation. Also, because type expressions of void types have an empty subtype, this requires a different adjustment than `cpp_new_arr`.
> 
> Of course op_new is not necessary, I can remove it, but the judgment will be complicated.
> 

IIUC, we have two cases:_
- `int* p = new int;` should create cpp_new with the size of 1 element, e.g., equivalent to `malloc(sizeof(int))`
- `int* p = new int[XYZ];` should create cpp_new_arr with size XYZ, e.g., equivalent to `malloc(XYZ * sizeof(int))`, a dyn array.

The only diff in handling both in symex_cpp_new is that we create a dyn_array for the latter.

Why would op_new create a dyn_array if it's a single value in memory? 

I didn't understand:

> Also, because type expressions of void types have an empty subtype, this requires a different adjustment than `cpp_new_arr`.

Can you give an example?

> > Is there a case, system, arch, etc where new doesn't initialize?
> 
> Maybe this code fits what you said: `int* p = new int;` my previous expression is not accurate.

Is there a requirement from somewhere that we don't need to initialize new? If that's the case, we can get rid of the cpp_new_initialize call or condition it based on the requirements.

For instance, in the test case changed by this PR, shouldn't this work:
```
$ cat foo.cpp 
#include <assert.h>

#define NULL 0
#define TYPE int
#define INTR int

TYPE a;
TYPE *dev_a;
	
int main()
{
  dev_a = (int*) operator new(sizeof(int));
  assert(dev_a != NULL);
  assert(*dev_a == 0);

  int* p = new int;
  assert(*p == 0);
  return 0;
}
``` 
?

XLiZHI:
> Why would op_new create a dyn_array if it's a single value in memory?

I investigated similar code in C verification and found that it does this (in `symex_mem`, we create dyn arr):
`void *foo = malloc(8);`

> Can you give an example?

In #2004, The old code creates a dyn symbol that is a subtype of void, which is:
```
pointer
  subtype: empty
  #cpp_type: void
```
The new symbol is of type empty, causing the assertion violation.


```
#include <assert.h>

#define NULL 0
#define TYPE int
#define INTR int

TYPE a;
TYPE *dev_a;
	
int main()
{
  dev_a = (int*) operator new(sizeof(int));
  assert(dev_a != NULL);
  assert(*dev_a == 0);                                                        //should fail

  int* p = new int;
  assert(*p == 0);                                                           //should fail
  return 0;
}
```
At least that's the case in clang++.

mikhailramalho:
Em sex., 20 de set. de 2024 às 13:21, Xianzhiyu ***@***.***>
escreveu:

> Why would op_new create a dyn_array if it's a single value in memory?
>
> I investigated similar code in C verification and found that it does this
> (in symex_mem, we create dyn arr):
> void *foo = malloc(8);
>

Now that seems to be the problem, probably because when it sees a constant
we assume it's an array. Does the malloc code see the rhs of the expr it's
returning the allocated memory to? If that's the case, it should be trivial
to decide if we need an array or an object, and fall back to an array if we
can't decide.

CBMC seems to be doing the right thing here, can you check how it does that
too?


> Can you give an example?
>
> In #2004 <https://github.com/esbmc/esbmc/issues/2004>, The old code
> creates a dyn symbol that is a subtype of void, which is:
>
> pointer
>   subtype: empty
>   #cpp_type: void
>
> The new symbol is of type empty, causing the assertion violation.
>
> #include <assert.h>
>
> #define NULL 0
> #define TYPE int
> #define INTR int
>
> TYPE a;
> TYPE *dev_a;
> 	
> int main()
> {
>   dev_a = (int*) operator new(sizeof(int));
>   assert(dev_a != NULL);
>   assert(*dev_a == 0);                                                        //should fail
>
>   int* p = new int;
>   assert(*p == 0);                                                           //should fail
>   return 0;
> }
>
> At least that's the case in clang++.
>

Weirdly it doesn't for me. Unless I enable a sanitizer, then I only get an
assertion failure, not even a sanitizer warning.


-- 

Mikhail R. Gadelha, Ph.D.

intrigus-lgtm:
In my opinion, it would be best to conceptually split
```c++
foobar* aaa = new foobar; // calls foobar's constructor
```

into two statements:

```c++
void* foobar_memory = operator new (sizeof foobar);
foobar* aaa = new(foobar_memory) foobar;
```
This way, `operator new` can be implemented as just allocating the memory and we don't need special-casing for it.
After the code has been refactored to be split into two, we can then consider simply implementing `operator new` on top of `malloc`.

mikhailramalho:
> In my opinion, it would be best to conceptually split
> 
> ```c++
> foobar* aaa = new foobar; // calls foobar's constructor
> ```
> 
> into two statements:
> 
> ```c++
> void* foobar_memory = operator new (sizeof foobar);
> foobar* aaa = new(foobar_memory) foobar;
> ```
> 
> This way, `operator new` can be implemented as just allocating the memory and we don't need special-casing for it. After the code has been refactored to be split into two, we can then consider simply implementing `operator new` on top of `malloc`.

Do we need it tho? Remember that any change to the types, including adding new ids, needs to be handle all the way from the frontend to the backend.

Is the current operator new handling so broken that we need a new approach here?

mikhailramalho:
Can't we just remove `cpp_new_initializer(lhs, rhs, tmp_initializer);`?

XLiZHI:
> Can't we just remove `cpp_new_initializer(lhs, rhs, tmp_initializer);`?

If we remove the initialization, we cannot handle the following code:
`int* p = new int(1);`

I checked out the CBMC implementation and found the following two:
1. CBMC does not support the verification of operator new.
2. The malloc size determination in CBMC is similar to that in ESBMC in that it is not complicated and uses dynamic arrays in most cases.

I think the operator new case is like malloc, and we only need to think about alloc_size, I don't think the difference between `dyn_arr` and `dyn_value` is that big. dyn_arr can handle any case.

mikhailramalho:
> > Can't we just remove `cpp_new_initializer(lhs, rhs, tmp_initializer);`?
> 
> If we remove the initialization, we cannot handle the following code: `int* p = new int(1);`
> 
> I checked out the CBMC implementation and found the following two:
> 
> 1. CBMC does not support the verification of operator new.
> 2. The malloc size determination in CBMC is similar to that in ESBMC in that it is not complicated and uses dynamic arrays in most cases.
> 
> I think the operator new case is like malloc, and we only need to think about alloc_size, I don't think the difference between `dyn_arr` and `dyn_value` is that big. dyn_arr can handle any case.

Can't you check if it has an initializer then?

mikhailramalho:
@XLiZHI how about this:
```diff
diff --git a/src/goto-programs/builtin_functions.cpp b/src/goto-programs/builtin_functions.cpp
index 77a053a7e..42aeaaf20 100644
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -344,11 +344,16 @@ void goto_convertt::cpp_new_initializer(
   goto_programt &dest)
 {
   // grab initializer
-  code_expressiont initializer;
+  code_expressiont initializer = (code_expressiont &)rhs.initializer();
+  if(initializer.is_nil())
+    return;
 
-  if (rhs.initializer().is_nil())
+  if (!initializer.op0().get_bool("constructor"))
   {
-    // Initialize with default value
+    // for auto *p = Foo(3) and int *p = 3
+    // constructor case:  init is Foo(&(*new_object), 3)
+    // other case: init is 3,
+    // we turn "3" into "*new_object = 3"
     side_effect_exprt assignment("assign");
     assignment.type() = rhs.type().subtype();
 
@@ -357,37 +362,12 @@ void goto_convertt::cpp_new_initializer(
     new_object.set("#lvalue", true);
     new_object.type() = rhs.type().subtype();
 
-    // Default value is zero
-    exprt default_value = gen_zero(rhs.type().subtype());
-
-    assignment.move_to_operands(new_object, default_value);
+    assignment.move_to_operands(new_object, initializer.op0());
     initializer.expression() = assignment;
   }
-  else
-  {
-    initializer = (code_expressiont &)rhs.initializer();
 
-    if (!initializer.op0().get_bool("constructor"))
-    {
-      // for auto *p = Foo(3) and int *p = 3
-      // constructor case:  init is Foo(&(*new_object), 3)
-      // other case: init is 3,
-      // we turn "3" into "*new_object = 3"
-      side_effect_exprt assignment("assign");
-      assignment.type() = rhs.type().subtype();
-
-      // the new object
-      exprt new_object("new_object");
-      new_object.set("#lvalue", true);
-      new_object.type() = rhs.type().subtype();
-
-      assignment.move_to_operands(new_object, initializer.op0());
-      initializer.expression() = assignment;
-    }
-
-    // XXX jmorse, const-qual misery
-    const_cast<exprt &>(rhs).remove("initializer");
-  }
+  // XXX jmorse, const-qual misery
+  const_cast<exprt &>(rhs).remove("initializer");
 
   if (initializer.is_not_nil())
   {
```

It removes the default initialization of new objects, and passes all tests in PR + `int* p = new int(1);`

src/goto-programs/builtin_functions.cpp
```diff
@@ -744,10 +729,16 @@ void goto_convertt::do_function_call_symbol(
   {
     assert(arguments.size() == 1);
 
+    typet alloc_type;
+    exprt alloc_size;
+    get_alloc_type(arguments[0], alloc_type, alloc_size);
+
     // Change it into a cpp_new expression
```
intrigus-lgtm:
```suggestion
    // Change it into a cpp_new[] expression
```

intrigus-lgtm:
The new patch let's my code work, so :+1:

lucasccordeiro:

Thanks very much for your review, @mikhailramalho and @intrigus-lgtm.

mikhailramalho:
This shouldn't have been merged as it introduces a new id for no good reason.

XLiZHI:
> This shouldn't have been merged as it introduces a new id for no good reason.

This PR doesn't introduce a new id, it just sets a flag.

mikhailramalho:
> > This shouldn't have been merged as it introduces a new id for no good reason.
> 
> This PR doesn't introduce a new id, it just sets a flag.

My mistake, it sets a flag for no good reason.

You only need the patch I sent to pass the new tests introduced by this PR, please provide a test case that's only fixed by the new flag or we'll need to revert this change.

XLiZHI:
> > > This shouldn't have been merged as it introduces a new id for no good reason.
> > 
> > 
> > This PR doesn't introduce a new id, it just sets a flag.
> 
> My mistake, it sets a flag for no good reason.
> 
> You only need the patch I sent to pass the new tests introduced by this PR, please provide a test case that's only fixed by the new flag or we'll need to revert this change.


This one?
```
#include <cassert>
#include <new>

int main()
{
  // Allocate memory for 5 characters
  void *foo = operator new(sizeof(char) * 5);

  // Use the allocated memory as an array of characters
  char *charPtr = static_cast<char *>(foo);
  charPtr[0] = 'H';
  charPtr[1] = 'e';
  charPtr[2] = 'l';
  charPtr[3] = 'l';
  charPtr[4] = 'o';

  // Check each character using assertions
  assert(charPtr[0] == 'H');
  assert(charPtr[1] == 'e');
  assert(charPtr[2] == 'l');
  assert(charPtr[3] == 'l');
  assert(charPtr[4] == 'o');

  // Deallocate the memory
  operator delete(foo);

  return 0;
}
```

mikhailramalho:
> > > > This shouldn't have been merged as it introduces a new id for no good reason.
> > > 
> > > 
> > > This PR doesn't introduce a new id, it just sets a flag.
> > 
> > 
> > My mistake, it sets a flag for no good reason.
> > You only need the patch I sent to pass the new tests introduced by this PR, please provide a test case that's only fixed by the new flag or we'll need to revert this change.
> 
> This one?
> 
> ```
> #include <cassert>
> #include <new>
> 
> int main()
> {
>   // Allocate memory for 5 characters
>   void *foo = operator new(sizeof(char) * 5);
> 
>   // Use the allocated memory as an array of characters
>   char *charPtr = static_cast<char *>(foo);
>   charPtr[0] = 'H';
>   charPtr[1] = 'e';
>   charPtr[2] = 'l';
>   charPtr[3] = 'l';
>   charPtr[4] = 'o';
> 
>   // Check each character using assertions
>   assert(charPtr[0] == 'H');
>   assert(charPtr[1] == 'e');
>   assert(charPtr[2] == 'l');
>   assert(charPtr[3] == 'l');
>   assert(charPtr[4] == 'o');
> 
>   // Deallocate the memory
>   operator delete(foo);
> 
>   return 0;
> }
> ```

Why is this test case not part of the PR?

Please create a new PR to fix cpp_new instead of creating a new handler for it as I requested in my review.

XLiZHI:
I think there are three different ways to get the size:

1. new int -----> size = 1  `dyn_value`
2. new int[n] -----> size = sizeof(int) * n  `dyn_arr`
3. operator new(n) -------> size = n

If we treat `operator new` as `new` or `new[]`, size will be the incorrect value. That's why I added special handling to `op_new`. If you insist on not introducing new handler, I will revert this code.

lucasccordeiro:
@XLiZHI: did you add test cases that cover all these three scenarios?

XLiZHI:
> @XLiZHI: did you add test cases that cover all these three scenarios?

Since this PR is for the third case, I've only added two simple TCs, both of which were allocated the correct memory size in my investigation.

mikhailramalho:
There are no 3 scenarios, there is either new and new[].

Fix cpp_new and don't introduce new complexity to the codebase without reason.

lucasccordeiro:
@mikhailramalho: what would be your proposed solution to fix this issue https://github.com/esbmc/esbmc/issues/2004 based on `new`and `new[]`? 

It would be great if we could advance with modeling this feature in our C++ frontend.

mikhailramalho:
yes, the main issue is that our internal representation of new and new[] is weird: instead of having the size as an operand, we save it in a named flag, causing all kinds of weird ways of handling it.

The first step is to fix this and then handle `operator new` and `operator new[]` as the actual types we have defined, instead of forcing `new` to become `new[]` as in this PR.

PR #2045, opened 2024-09-25
[C++ verification] support CXXForRangeStmtClass

XLiZHI:
For example code - For Range:

```cpp
int arr[] = {1, 2, 3, 4, 5};

for (int num : arr)
{
 }
```
Converted it to the following goto program:

```cpp
        DECL signed int [5] arr;

        ASSIGN arr={ 1, 2, 3, 4, 5 };

        DECL signed int [5] * __range1;

        ASSIGN __range1=&arr[0];

        DECL signed int * __begin1;

        ASSIGN __begin1=(signed int *)(&(*__range1)[0]);

        DECL signed int * __end1;

        ASSIGN __end1=&(*__range1)[0] + 5;

     1: IF !(__begin1 != __end1) THEN GOTO 2

        DECL signed int num;

        ASSIGN num=*__begin1;

        DEAD c:main5.cpp@98@F@main#@num

        ASSIGN __begin1=__begin1 + 1;

        GOTO 1

     2: DEAD c:main5.cpp@106@F@main#@__end1
     
        DEAD c:main5.cpp@106@F@main#@__begin1

        DEAD c:main5.cpp@108@F@main#@__range1
```


lucasccordeiro (approved):

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1042,6 +1042,75 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+    if (get_expr(*cxxfor.getBeginStmt(), begin))
+      return true;
+
+    exprt end;
+    if (get_expr(*cxxfor.getEndStmt(), end))
+      return true;
+
+    decls.move_to_operands(range, begin, end);
+    convert_expression_to_code(decls);
+    decls.set("for_range", 1);
```
intrigus-lgtm: Change to `true`.
```suggestion
    decls.set("for_range", true);
```

regression/esbmc-cpp/cpp/for_range/main.cpp
intrigus-lgtm, line 1: Can we move the tests to cpp11 where this feature has been introduced?

intrigus-lgtm (changes requested):
Can you also please "support"/_add tests_ for init statements added with c++20?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1042,6 +1042,75 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+
+    codet::operandst &ops = body.operands();
+    ops.insert(ops.begin(), loop);
+    convert_expression_to_code(body);
+
+    code_fort code_for;
+    code_for.init() = decls;
+    code_for.cond() = cond;
+    code_for.iter() = inc;
+    code_for.body() = body;
```
mikhailramalho, line 1109: These are unused later on adjust. Can't we set the ranges here and don't change adjust instead?

XLiZHI: This is a good point

mikhailramalho:
@XLiZHI this passes the two tests in this PR and no changes are required to the adjust code:
```
index d6430cae0..4b3ad88b0 100644
@@ -1042,6 +1042,63 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
     break;
   }
 
+  case clang::Stmt::CXXForRangeStmtClass:
+  {
+    const clang::CXXForRangeStmt &cxxfor =
+      static_cast<const clang::CXXForRangeStmt &>(stmt);
+
+    exprt cond;
+    if (get_expr(*cxxfor.getCond(), cond))
+      return true;
+
+    codet inc = code_skipt();
+    const clang::Stmt *inc_stmt = cxxfor.getInc();
+    if (inc_stmt)
+      if (get_expr(*inc_stmt, inc))
+        return true;
+    convert_expression_to_code(inc);
+
+    exprt range;
+    if (get_expr(*cxxfor.getRangeStmt(), range))
+      return true;
+
+    exprt begin;
+    if (get_expr(*cxxfor.getBeginStmt(), begin))
+      return true;
+
+    exprt end;
+    if (get_expr(*cxxfor.getEndStmt(), end))
+      return true;
+
+    codet decls("decl-block");
+    decls.move_to_operands(range.op0(), begin.op0(), end.op0());
+
+    codet body = code_skipt();
+    const clang::Stmt *body_stmt = cxxfor.getBody();
+    if (body_stmt)
+      if (get_expr(*body_stmt, body))
+        return true;
+
+    const clang::Stmt *loop_var = cxxfor.getLoopVarStmt();
+    exprt loop;
+    if (loop_var)
+      if (get_expr(*loop_var, loop))
+        return true;
+
+    codet::operandst &ops = body.operands();
+    ops.insert(ops.begin(), loop);
+    convert_expression_to_code(body);
+
+    code_fort code_for;
+    code_for.init() = decls;
+    code_for.cond() = cond;
+    code_for.iter() = inc;
+    code_for.body() = body;
+
+    new_expr = code_for;
+    break;
+  }
+
   default:
     if (clang_c_convertert::get_expr(stmt, new_expr))
       return true;
```

mikhailramalho:
I didn't really like the following line:
```
decls.move_to_operands(range.op0(), begin.op0(), end.op0());
``` 

But unfortunately, I couldn't come up with a better solution. `begin`, `range` and `end` generate `decl-blocks`, so we end up with `decl-blocks` inside `decl-blocks`, which violates our internal assertions. Do you want to try to investigate a better solution? Otherwise, just add a comment on why we are accessing the op0.

Annoyingly, I couldn't fix the need for an `--unwind 6` either. ESBMC should be able to handle so a simple case, with a static array.

intrigus-lgtm (approved):

lucasccordeiro:

PR #2048, opened 2024-09-28
[C++ verification] fix std::trunc and add OMs for std::unique_ptr

XLiZHI:


regression/esbmc-cpp11/cpp/trunc/main.cpp
```diff
@@ -7,25 +7,25 @@ int main()
   float f2 = -2.99f;
   float f3 = 0.0f;
 
-  assert(trunc(f1) == 3.0f);
```
lucasccordeiro, line 10: You could keep the previous test case and add a new one ;-)

lucasccordeiro (approved):

regression/esbmc-cpp11/cpp/trunc/main.cpp
XLiZHI: We already have a similar TC to test trunc (C code)

regression/floats-regression/rounding_functions/main.c:
```c
  __VERIFIER_assert(trunc(2.3) == 2.0);
  __VERIFIER_assert(trunc(3.8) == 3.0);
  __VERIFIER_assert(trunc(5.5) == 5.0);
  __VERIFIER_assert(trunc(-2.3) == -2.0);
  __VERIFIER_assert(trunc(-3.8) == -3.0);
  __VERIFIER_assert(trunc(-5.5) == -5.0);
```

lucasccordeiro:

PR #2050, opened 2024-09-28
[clang-cpp] fix undeduced return types in unused templates

intrigus-lgtm:


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -640,6 +640,22 @@ bool clang_c_convertert::get_function(
       return false;
   }
 
+  // per https://eel.is/c++draft/dcl.spec.auto#general-14 return types of template functions are
+  // only deduced when they are instantiated (i.e. used).
+  // We skip those functions as they are unused anyway and the rest of esbmc can't handle undeduced types.
+  if (fd.getReturnType()->isUndeducedType() && !fd.isUsed(true))
```
mikhailramalho: isUsed is not reliable, e.g., sometimes functions are not marked as used if called through function pointers

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -640,6 +640,22 @@ bool clang_c_convertert::get_function(
...
+  {
+    return false;
+  }
+
+  // per https://eel.is/c++draft/dcl.spec.auto#general-14 return types of template functions are
+  // only deduced when they are instantiated (i.e. used).
+  // We skip those functions as they are unused anyway and the rest of esbmc can't handle undeduced types.
+  if (fd.getReturnType()->isUndeducedType() && !fd.isUsed(true))
```
mikhailramalho: duplicated code.

mikhailramalho (changes requested):
only checking `ìf(fd.getReturnType()->isUndeducedType())` fixes the issue?

Also, don't use brackets on single line `if`s.

src/clang-c-frontend/clang_c_convert.cpp
intrigus-lgtm: This was the result of badly resolving a merge conflict, sorry.

src/clang-c-frontend/clang_c_convert.cpp
intrigus-lgtm: If being called through a function pointer, it should be used, because we have to take the address of it.
I can change this to just `ìf(fd.getReturnType()->isUndeducedType())` if you want.

intrigus-lgtm: In the end, I included the `!fd.isUsed(true)` conjunct just in case there would ever be an undeduced type that **is** used. If we would ever hit such a case, we would crash in https://github.com/esbmc/esbmc/blob/bbd0ed41a74845b0f89b7152c1a6859d260be10a/src/clang-c-frontend/clang_c_convert.cpp#L1253
instead of just skipping the function and having surprising problems later on.

XLiZHI (approved):

lucasccordeiro:
@intrigus, @intrigus-lgtm : could you please fix the conflicting files so we can merge this PR?

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: The assert is fine, let's just remove the `isUsed`.

Also, can params be undeduced too? If so, we shouldn't parse the method in that case either.

mikhailramalho: Can we have undeduced types in c code? Otherwise, we can move this check to `get_method` instead.

intrigus-lgtm: > Can we have undeduced types in c code? Otherwise, we can move this check to `get_method` instead.

Yes-ish:
```cpp
auto foo();
```
generates this AST:
```
TranslationUnitDecl
`-FunctionDecl <line:1:1, col:10> col:6 foo 'auto ()'
```
so we have to continue to handle this in `get_function` I believe.
https://godbolt.org/z/j5PdoofMv

intrigus-lgtm: > The assert is fine, let's just remove the `isUsed`.
> 
> Also, can params be undeduced too? If so, we shouldn't parse the method in that case either.

I didn't find any cases, where params are undeduced. In fact, `auto` in params, almost always generates some kind of implicit template, see: https://en.cppreference.com/w/cpp/language/auto#Parameter_declarations

This was my test, which works fine with this PR:
```cpp
#include <utility>

int foo(auto a);

template<auto n> // C++17 auto parameter declaration
std::pair<decltype(n), decltype(n)> f() // auto can't deduce from brace-init-list
{
    return {n, n};
}

// generic lambda, operator() is a template with two parameters
auto glambda = [](auto a, auto&& b) -> bool { return a < b; };
// bool b = glambda(3, 3.14); // OK
```

intrigus-lgtm: I moved the part that checks whether the function is used to an assert.

mikhailramalho: please remove the isUsed call, as it is not reliable.

intrigus-lgtm: > please remove the isUsed call, as it is not reliable.

In all my testing I did not find any case where `isUsed` was not reliable.
It's also in an assert so only devs will be affected should there ever be a false result.

mikhailramalho: Your tests do not cover all possible C++ programs in existence. 

We release esbmc with RelWithDebInfo, so asserts are enabled for users too.

Remove the isUsed assertion.

lucasccordeiro:
> We release esbmc with RelWithDebInfo, so asserts are enabled for users too.

@rafaelsamenezes can confirm it, but I think we don't enable the assertions in the ESBMC releases.

@intrigus-lgtm : Would you like ESBMC to show the user an error message? We could use `log_error` instead.

rafaelsamenezes:
RelWithDebInfo also defines NDEBUG. So it should be without any assertions.

lucasccordeiro:
Thanks for confirming there are no assertions in the ESBMC release version, @rafaelsamenezes.

@intrigus-lgtm: Thanks for submitting this PR.

@mikhailramalho and @XLiZHI: Thanks for reviewing this PR.

lucasccordeiro (approved):

lucasccordeiro:
@intrigus-lgtm: The GitHub web interface complains about merging this PR. There might still be some issues about the rebase. Could you please check that?

mikhailramalho:
The `isUsed` macro is unreliable; we used it in esbmc before and it caused issues in the past. Remove the assertion.

lucasccordeiro:
> The isUsed macro is unreliable; we used it in esbmc before and it caused issues in the past. Remove the assertion.

@mikhailramalho: Could you please point to the issue so that @intrigus-lgtm can check it and advance with this PR?

mikhailramalho:
Please check esbmc's history: we used it and eventually, we started to get a bunch of wrong results in sv-comp because we were removing/not parsing functions that were needed.

review.llvm.org used to have a bunch of PRs to fix isUsed issues but the site is gone now.

lucasccordeiro:
Many thanks, @mikhailramalho.

@intrigus-lgtm: Could you please remove the macro as @mikhailramalho suggested?

Could you also check the rebase process on your PR?

lucasccordeiro:
@intrigus-lgtm: Thanks for addressing @mikhailramalho's suggestion.

PR #2051, opened 2024-09-29
[C++ verification] support std::forward and std::make_unique

XLiZHI:
There is a bug with `remove_sideeffect` that causes the destructor to generate the wrong goto function, which needs further investigation.


lucasccordeiro (approved):

lucasccordeiro:

> There is a bug with remove_sideeffect that causes the destructor to generate the wrong goto function, which needs further investigation.

Do we have an open issue for this?

Issue #1713, opened 2024-02-23
Assertion `a.is_constant()' failed at value_set.cpp:804


Anthonysdu:
```c
#define PAGE_SIZE ((1) << XLAT_GRANULARITY_SIZE_SHIFT)

struct rmm_core_manifest {
  int a;
};

int main() {
  static unsigned char el3_rmm_shared_buffer[PAGE_SIZE]
      __attribute__((__aligned__(PAGE_SIZE)));
  static struct rmm_core_manifest *boot_manifest =
      (struct rmm_core_manifest *)el3_rmm_shared_buffer;

  return 0;
}
```

`esbmc main.c -DXLAT_GRANULARITY_SIZE_SHIFT=7`:

```
ESBMC version 7.5.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Enabling --no-slice due to presence of --smt-during-symex
Parsing main.c
Converting
Generating GOTO Program
GOTO program creation time: 0.059s
GOTO program processing time: 0.000s
No solver specified; defaulting to Boolector
Starting Bounded Model Checking
esbmc: /home/runner/work/esbmc/esbmc/src/pointer-analysis/value_set.cpp:804: void value_sett::get_reference_set_rec(const expr2tc&, value_sett::object_mapt&) const: Assertion `a.is_constant()' failed.
Aborted (core dumped)
```

fbrausse:
This should be relatively easy to resolve in the clang-c frontend: The alignment needs to be a compile-time constant. Maybe we're not using the right Clang API to obtain it.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2060.

PR #2057, opened 2024-10-07
simplify operator for alignment

Anthonysdu:
This PR fixes issue #1713, we need to simplify the operator for alignment to pass the assertion check.

### DIFF
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -801,8 +801,12 @@ void value_sett::get_reference_set_rec(const expr2tc &expr, object_mapt &dest)
       const irept &a = sym->type.find("alignment");
       if (a.is_not_nil())
       {
-        assert(a.is_constant());
-        irep_idt v = static_cast<const exprt &>(a).value();
+        expr2tc temp;
+        migrate_expr(static_cast<const exprt &>(a), temp);
+        simplify(temp);
+        auto temp2 = migrate_expr_back(temp);
+        assert(temp2.is_constant());
+        irep_idt v = temp2.value();
         BigInt V = binary2integer(v.as_string(), false);
         assert(V.is_positive());
         assert(V <= UINT_MAX);


lucasccordeiro (changes requested):
@Anthonysdu: could you please add some regression tests?

mikhailramalho (changes requested):
Let's follow Franz's suggestion and fix that in the frontend. The AST seems to already provide the constant:
```
    | `-VarDecl 0x555556ad1320 <line:8:3, col:55> col:24 used el3_rmm_shared_buffer 'unsigned char[2]' static
    |   `-AlignedAttr 0x555556ad13a8 <line:9:22, col:43> aligned
    |     `-ConstantExpr 0x555556ad1388 <line:1:19, col:54> 'int'
    |       |-value: Int 2
    |       `-ParenExpr 0x555556ad1278 <col:19, col:54> 'int'
    |         `-BinaryOperator 0x555556ad1258 <col:20, <command line>:6:37> 'int' '<<'
    |           |-ParenExpr 0x555556ad1218 <foo.c:1:20, col:22> 'int'
    |           | `-IntegerLiteral 0x555556ad11f8 <col:21> 'int' 1
    |           `-IntegerLiteral 0x555556ad1238 <<command line>:6:37> 'int' 1
```

lucasccordeiro:
@Anthonysdu: if you don't understand the Clang-C frontend, I would suggest that you work together with @XLiZHI to help you here.

lucasccordeiro:
Oh, sorry!

I pressed the wrong button!

@Anthonysdu: could you please revert it?

mikhailramalho:
A quick look in the API and I see that ConstantExpr has a `hasAPValueResult()`, so let's check if it has an apvalue, and create a constant accordingly (there are a few different kinds of apvalues).

If we can't create the constant or there is no apvalue, we should fall back to building the expr.

src/pointer-analysis/value_set.cpp
```diff
@@ -801,8 +801,12 @@ void value_sett::get_reference_set_rec(const expr2tc &expr, object_mapt &dest)
       const irept &a = sym->type.find("alignment");
       if (a.is_not_nil())
       {
-        assert(a.is_constant());
```
XLiZHI, line 804: I debugged this line of the code and printed the expr for `a`.
```
shl
  * type: signedbv
      * width: 32
      * #cpp_type: signed_int
  * operands: 
    0: constant
        * type: signedbv
            * width: 32
            * #cpp_type: signed_int
        * #location: 
          * file: main3.c
          * line: 9
          * function: main
          * column: 34
        * #cformat: 1
    1: constant
        * type: signedbv
            * width: 32
            * #cpp_type: signed_int
        * value: 00000000000000000000000000000111
        * #location: 
          * file: main3.c
          * line: 9
          * function: main
          * column: 34
        * #cformat: 7
  * #location: 
    * file: main3.c
    * line: 9
    * function: main
    * column: 34
```
This is actually still constant, but it's a computation, maybe because the assertion is too strict?

PR #2062, opened 2024-10-08
Add static llvm-16 build to CI, use llvm-16 as the default

XLiZHI:


XLiZHI:
Hi @rafaelsamenezes, Do you know why arm64 CI fails? I believe we support shared builds for llvm-16. Maybe arm64 doesn't have llvm16 installed?

intrigus-lgtm:
> Hi @rafaelsamenezes, Do you know why arm64 CI fails? I believe we support shared builds for llvm-16. Maybe arm64 doesn't have llvm16 installed?

Well, we never install any packages on aarch64 (notice the return):
https://github.com/esbmc/esbmc/blob/ea5f9adb268b5eb9440400749fd7c2696b57bf70/scripts/build.sh#L68-L75

And we also don't add the llvm repo for aarch64, because the self-hosted runner's name is "ARM64":
https://github.com/esbmc/esbmc/blob/ea5f9adb268b5eb9440400749fd7c2696b57bf70/.github/workflows/build-unix.yml#L24-L30

XLiZHI:
Yes, I was investigating the CI logs and found that arm64 doesn't seem to have downloaded llvm, so I'm guessing it might be pre-installed in the self host.

rafaelsamenezes:
> Yes, I was investigating the CI logs and found that arm64 doesn't seem to have downloaded llvm, so I'm guessing it might be pre-installed in the self host.

That seems to be the case. We need to manually double check this.

mikhailramalho:
also, feel free to remove the `#CLANG_VERSION_MAJOR` checks where applicable.

XLiZHI:
I do not seem to have permissions to manage self host, can someone help me check it?

lucasccordeiro:
> I do not seem to have permissions to manage self host, can someone help me check it?

I have just changed your permission from `maintainer` to `admin`.

XLiZHI:
> > I do not seem to have permissions to manage self host, can someone help me check it?
> 
> I have just changed your permission from `maintainer` to `admin`.

Thank you, but we don't seem to be configuring on Github. In my understanding, it might be to `SSH` into arm64 runner and install llvm-16 on it that we need. What we might need is an `SSH` link and permission token to do this. I have no experience with this part, please correct me if I am wrong.

XLiZHI:
I temporarily used llvm-13 to build ARM64 and it worked. Currently, our windows builds still use LLVM-11. I checked the llvm repository and found that it has few releases for windows, only LLVM-18 has a win release.

I noticed that @Anthonysdu has a windows version of LLVM-11 in repository, this version is not published in LLVM repo, is this a custom build?

Anthonysdu:
I remember last time windows CI failed by llvm13 so we switched back to llvm11, and I built it from source. Can you remember @rafaelsamenezes?

rafaelsamenezes:
> I remember last time windows CI failed by llvm13 so we switched back to llvm11, and I built it from source. Can you remember @rafaelsamenezes?

Yeah, we may need to something similar for CHERI builds

lucasccordeiro:
@rafaelsamenezes: could you please check the failing CI here?

rafaelsamenezes:
Wait there is some mixing up here.

1. Windows builds needs to be custom built because LLVM release of windows is incomplete and non-standard. I played around trying to linking it with a MSVC build but failed miserably. So we either revert the windows to 11 or do a custom build for 16
2. The arm64 linux should be able to use the same configurations from apt that x86 uses. The main difference here is that the arm64 is using Ubuntu 22.04.

XLiZHI:
> The arm64 linux should be able to use the same configurations from apt that x86 uses. The main difference here is that the arm64 is using Ubuntu 22.04.

I tried setting up apt for arm64, but the arm64 runner seems to require a password to use sudo, which seems to be different from the other self-host configurations.

```
Run wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc &&
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
Error: Process completed with exit code 1.
```
Or we can use llvm-13 to build the ARM64 version, which works.

I think we also need to run sv-comp to make sure nothing is broken.

rafaelsamenezes:
> wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc

We prevent sudo access to our self-hosted runners, I will try to setup this later this month. Meanwhile you are correct, arm64 builds should not hold us off.

rafaelsamenezes (approved):
@XLiZHI is this ready for review?

lucasccordeiro (approved):
@XLiZHI: could you please run this PR over the SV-COMP benchmarks before we merge it?

XLiZHI:
> @XLiZHI is this ready for review?

Shell we wait for the results of sv-comp?

https://github.com/esbmc/esbmc/actions/runs/11254680345/job/31293593693

lucasccordeiro:
Recent master results for 30 seconds:

````
  correct:           16534
    correct true:    11085
    correct false:    5449
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13738
  Score:             26883 (max: 51772)
````

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/11219200024

This PR:

````
  correct:           16474
    correct true:    11034
    correct false:    5440
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13798
  Score:             26772 (max: 51772)
````

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/11264291514

lucasccordeiro:

PR #2072, opened 2024-10-14
[Coverage] Enable no-unwinding-assertions in Coverage mode

ChenfengWei0:
The calculation of the coverage relies on the instrumented assertions. This, however, will be affected by other instrumentation like unwinding-assertions. Therefore, we need to enable `no-unwinding-assertions` by default

### DIFF
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -2120,17 +2120,17 @@ void esbmc_parseoptionst::add_property_monitors(
   std::map<std::string, std::pair<std::set<std::string>, expr2tc>> monitors;
 
   context.foreach_operand([this, &monitors](const symbolt &s) {
-      if (
-        !has_prefix(s.name, "__ESBMC_property_") ||
-        s.name.as_string().find("$type") != std::string::npos)
-        return;
+    if (
+      !has_prefix(s.name, "__ESBMC_property_") ||
+      s.name.as_string().find("$type") != std::string::npos)
+      return;
 
-      // strip prefix "__ESBMC_property_"
-      std::string prop_name = s.name.as_string().substr(17);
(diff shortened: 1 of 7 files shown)


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1843,6 +1843,10 @@ bool esbmc_parseoptionst::process_goto_program(
       options.set_option("keep-verified-claims", false);
       options.set_option("no-pointer-check", true);
 
+      // enable '--no-unwinding-assertions' if '--unwind' is enabled
+      if (cmdline.isset("unwind"))
```
lucasccordeiro, line 1847: I think we should enable `--no-unwinding-assertion` only if the user chooses `--branch-coverage`.

@rafaelsamenezes and @mikhailramalho: What are your thoughts here?

lucasccordeiro (changes requested):

src/esbmc/esbmc_parseoptions.cpp
ChenfengWei0: > I think we should enable --no-unwinding-assertion only if the user chooses --branch-coverage.

I believe that is how it works. We also apply this strategy for assertion and condition coverage as well

```
 if (
      cmdline.isset("assertion-coverage") ||
      cmdline.isset("assertion-coverage-claims"))
{
     if (cmdline.isset("unwind"))
        options.set_option("no-unwinding-assertions", true);
}
```

lucasccordeiro (approved):
Thanks for clarifying this, @ChenfengWei0.

lucasccordeiro:
@ChenfengWei0: could you please add regression tests for assertion and condition coverages?

You have just added tests for branch coverages.

lucasccordeiro:

PR #2077, opened 2024-10-15
[CI] Specify the Ubuntu version

XLiZHI:
GitHub action updated Ubuntu to the latest version 24.04. But we don't fully support it yet, so we specify 22.04 in CI.

BTW, the workflow was updated in the previous PR, now we should remove the llvm-11 "require" tag in workflow, and add a "require" tag to llvm-16.

### DIFF
--- a/.github/workflows/build-onpush.yml
+++ b/.github/workflows/build-onpush.yml
@@ -17,6 +17,6 @@ jobs:
   build-unix:
     uses: ./.github/workflows/build-unix.yml
     with:
-      operating-system: ubuntu-latest
+      operating-system: ubuntu-22.04
       build-flags: '-b Debug'
       testing: true
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -32,7 +32,7 @@ jobs:
     uses: ./.github/workflows/build-unix.yml
     name: static llvm-16 DebugOpt
     with:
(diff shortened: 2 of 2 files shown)


lucasccordeiro (approved):

lucasccordeiro:
> BTW, the workflow was updated in the previous PR, now we should remove the llvm-11 "require" tag in workflow, and add a "require" tag to llvm-16.

@XLiZHI: could you please submit a PR for this?

XLiZHI:
> > BTW, the workflow was updated in the previous PR, now we should remove the llvm-11 "require" tag in workflow, and add a "require" tag to llvm-16.
> 
> @XLiZHI: could you please submit a PR for this?

This action needs to be done on the organisation page, in action-general if I remember correctly.

lucasccordeiro:
> This action needs to be done on the organisation page, in action-general if I remember correctly.

@XLiZHI: You're now admin :-)

XLiZHI:
> > This action needs to be done on the organisation page, in action-general if I remember correctly.
> 
> @XLiZHI: You're now admin :-)

https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization

Yes, but I tried to visit the organization page, and I couldn't see the settings. Only the "owner" and “member” can change the settings in the organization.

lucasccordeiro:

Issue #2063, opened 2024-10-08
[Coverage] Assertion Instances Coverage: 1100%


lucasccordeiro:
$ esbmc main.c --assertion-coverage --unwind 10

````C
#include <assert.h>
int main () {
  int a , x , y = 0 , z;
  a = nondet_int();
  x = nondet_int();
  z = nondet_int();
  if (a > 25000 && x == 30000) {
    while (y++ < z) {
      if (y % 3 != 0) {
        a++;
        x--;
      } else {
      a--; x++;
      }
      assert (a != x);
   }
  }
}
````

````
[Coverage]

Total Asserts: 1
Total Assertion Instances: 1
Reached Assertion Instances: 11
Assertion Instances Coverage: 1100%

VERIFICATION FAILED
````

lucasccordeiro:
@ChenfengWei0: Did you have a chance to look at this issue?

ChenfengWei0:
> @ChenfengWei0: Did you have a chance to look at this issue?

Will check after finishing the test-comp stuff

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2084.

PR #2084, opened 2024-10-17
[Coverage] Update assertion coverage

ChenfengWei0:
Issue: #2063 

Result:

```
[Coverage]

Total Asserts: 1
Total Assertion Instances: unknown / non-deterministic
Reached Assertion Instances: 4
Assertion Instances Coverage Unknown

VERIFICATION FAILED
```

If an assertion is inside a nondet loop, then it has nondet assertion instance. Thus we cannot know the total number of assertion instances.


lucasccordeiro (approved):

lucasccordeiro:

PR #2090, opened 2024-10-19
[C++ verification] fix explicit destructor

### LINKED ISSUE #2052: [C++ verification] An explicit destructor causes an incorrect goto program
```cpp
class A
{
  int *ptr;
public:
  A() : ptr(nullptr)
  {
  }
  // dtor cause the problem
  ~A()
  {
  }
};

A func()
{
  return A();
}

int main()
{
  A a = func();
}
```
goto functions:
```c
main (c:@F@main#):
        // 38 file main5.cpp line 21 column 1 function main
        DECL A a;
        // 39 file main5.cpp line 21 column 7 function main
        DECL A tmp$1;
        // 40 file main5.cpp line 21 column 7 function main
        ASSIGN tmp$1=NONDET(struct);
        // 41 file main5.cpp line 21 column 7 function main
        DECL A tmp$2;
        // 42 file main5.cpp line 21 column 7 function main
        ASSIGN tmp$2=NONDET(struct);
        // 43 file main5.cpp line 21 column 7 function main
        FUNCTION_CALL:  func()            <--- where is return value?
        // 44 
        DECL A tmp$3;
        // 45 
        ASSIGN tmp$3=NONDET(struct);
        // 46 
        ASSIGN tmp$3=tmp$2;
        // 47 
        FUNCTION_CALL:  A(( A *)(&tmp$1), &tmp$3)
        // 48 file main5.cpp line 21 column 1 function main
        ASSIGN a=tmp$1
```

XLiZHI:
This PR fixes the GOTO error caused by the explicit destructor.

1. `FUNCTION_CALL:  func()` ---> `FUNCTION_CALL:  tmp$1 = func()`;
2. ~~No longer call destructors for temporary objects.~~

```
class MyClass {
private:
    int* data;
public:
    MyClass() {
        data = new int[10];
    }

    ~MyClass() {
        delete[] data;
    }
};
```
~~When we create a temporary object of type `MyClass`, calling the destructor dereferences the invalid object.~~

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/github_2052/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_2052/main.cpp
@@ -1,0 +1,24 @@
+#include <cassert>
+
+class A
+{
+public:
+  int num;
+  A (int n) : num(n)
+  {
+  }
+  ~A()
+  {
+  }
+};
+
+A func(int n)
+{
+  return A(n);
+}
+
+int main()
+{
+  A a = func(1);
+  assert(a.num == 1);
+}
--- a/regression/esbmc-cpp/bug_fixes/github_2052/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_2052/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/bug_fixes/github_2052_fail/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_2052_fail/main.cpp
@@ -1,0 +1,24 @@
+#include <cassert>
+
+class A
+{
+public:
+  int num;
+  A (int n) : num(n)
+  {
+  }
+  ~A()
+  {
+  }
+};
+
+A func(int n)
+{
+  return A(n);
+}
+
+int main()
+{
+  A a = func(1);
+  assert(a.num == 0);
+}
--- a/regression/esbmc-cpp/bug_fixes/github_2052_fail/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_2052_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION FAILED$
--- a/src/clang-cpp-frontend/clang_cpp_convert.cpp
+++ b/src/clang-cpp-frontend/clang_cpp_convert.cpp
@@ -2245,14 +2245,21 @@ bool clang_cpp_convertert::is_ConstructorOrDestructor(
 
 void clang_cpp_convertert::make_temporary(exprt &expr)
 {
-  if (expr.statement() != "temporary_object")
+  if (expr.statement() == "temporary_object")
+    return;
+
+  // make the temporary
+  side_effect_exprt tmp_obj("temporary_object", expr.type());
+  tmp_obj.location() = expr.location();
+  if (!expr.get_bool("constructor"))
   {
-    // make the temporary
-    side_effect_exprt tmp_obj("temporary_object", expr.type());
-    codet code_expr("expression");
-    code_expr.operands().push_back(expr);
-    tmp_obj.initializer(code_expr);
-    tmp_obj.location() = expr.location();
+    tmp_obj.move_to_operands(expr);
     expr.swap(tmp_obj);
+    return;
   }
+
+  codet code_expr("expression");
+  code_expr.operands().push_back(expr);
+  tmp_obj.initializer(code_expr);
+  expr.swap(tmp_obj);
 }


XLiZHI:
This PR:
```
  correct:           16463
    correct true:    11024
    correct false:    5439
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13809
  Score:             26751 (max: 51772)
  
  https://github.com/esbmc/esbmc/actions/runs/11420162249/job/31788318111
```
Master:
```
  correct:           16456
    correct true:    11017
    correct false:    5439
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13816
  Score:             26737 (max: 51772)
  https://github.com/esbmc/esbmc/actions/runs/11426032693/job/31788670371
```

lucasccordeiro (approved):

mikhailramalho:
why did you disable calling the destructor of temporaries? What happens if it changes a global state?
```
#include <cassert>

int foo = 0;

class A
{
public:
  int num;
  A (int n) : num(n)
  {
  }
  ~A()
  {
    foo = 1;
  }
};

A func(int n)
{
  return A(n);
}

int main()
{
  A a = func(1);
  assert(a.num == 1);
  assert(foo == 1);
}

**Comment by XLiZHI** [comment]
> why did you disable calling the destructor of temporaries? What happens if it changes a global state?

When the constructor of a class contains the creation of a dynamic object, creating a temporary object does not call the constructor, and destroying it results in a double delete, as in this example:

```
MyClass func() { return MyClass();}

MyClass a;
MyClass tmp$1;

tmp$1 = func();
a = tmp$1;

~MyClass(a);
~MyClass(tmp$1);  --> invaild obj
```

By the way, the mentioned example fails the assertion in Clang++.
```
int main(): Assertion `foo == 1' failed.
Aborted
```

**Comment by mikhailramalho** [comment]
maybe the constructor is elided?

something simpler:
```
#include <cassert>

class A
{
public:
  int num;
  A (int n) : num(n)
  {
  }
  ~A()
  {
    assert(0);
  }
};

A func(int n)
{
  return A(n);
}

int main()
{
  A a = func(1);
  assert(a.num == 1);
}

XLiZHI:
This example also works the same for esbmc and clang++. In this PR, instead of not calling the destructor, it is not called for the temporary object. So for objects declared in code, the destructor will still be called normally。

```
MyClass func() { return MyClass();}

MyClass a;
MyClass tmp$1;

tmp$1 = func();
a = tmp$1;

~MyClass(a);
// ~MyClass(tmp$1);  --> do not call dtor for tmp var
```

XLiZHI:
Also, here's an example of why we shouldn't call the destructor for a temporary object.

```
#include <cassert>
int count = 0;

class A
{
public:
  int num;
  A (int n) : num(n)
  {
  }
  ~A()
  {
    count++;
  }
};

A func(int n)
{
  return A(n);
}

void dtor()
{
  A a = func(1);
  assert(a.num == 1);
}

int main()
{
  dtor();
  assert(count == 1);
}
```
It works fine with Clang++.

mikhailramalho:
I still find it weird that we don't call destructors for temporaries. Can you test the following?
```
#include <cassert>

class A
{
public:
  int num;
  A () {}
  ~A() { assert(0); }
};

void func(A n)
{
  n.num = 3;
}

int main()
{
  func({});
}
```
Sorry, I can't test these cases rn, that's why I'm sending them here.

XLiZHI:
This example has incorrect results in esbmc, maybe we should distinguish whether we need to call destructors for temporary objects or not. 

1. only has tmp$ -->call dtor
2. has assignment: a = tmp$ --> do not call

The double delete mentioned above should be avoided, especially if the class member contains dynamic objects.

mikhailramalho:
temporary objects are just like any other object (as stated in the name, they are just temporary), so we should always call their destructors.

I'm not sure about also skipping returns, but maybe the DEAD instruction adds the destructor?

XLiZHI:
```
#include <cassert>
int count = 0;

class A
{
public:
  int num;
  A (int n) : num(n)
  {
  }
  ~A()
  {
    count++;
  }
};

A func(int n)
{
  return A(n);
}

void dtor()
{
  A a = func(1);
  assert(a.num == 1);
}

int main()
{
  dtor();
  assert(count == 1);
}
```
If that were the case, the count would no longer be 1 in this example, and we would call dtor multiple times.

Also, how do we handle the double delete case?

mikhailramalho:
> 
> If that were the case, the count would no longer be 1 in this example, and we would call dtor multiple times.

IIUC this works because the constructor is being elided, not even the copy constructor is being called in `A a = func(1);`. Can you delete the copy constructor in this example and check if it still works?

> 
> Also, how do we handle the double delete case?

I don't know

XLiZHI:
Let's fix the return value problem first, and I'll investigate if we can omit temporary objects.

lucasccordeiro:
@XLiZHI: Can I ask you for the status of this PR?

lucasccordeiro:
@XLiZHI: Can I ask you for the status of this PR?

lucasccordeiro (approved):

lucasccordeiro:

PR #2093, opened 2024-10-23
[C verification] fix value sets for constant union

XLiZHI:
This PR tries to fix an issue with constant union when it gets value sets.

SSA:
`u.__anon_field_1.p = { .__anon_field_1={ .cap=&"hello"[0] }.__anon_field_1 WITH [p:=0xFFFFFFFFFFFFFFFF] }`

https://github.com/esbmc/esbmc/blob/9b7c5bb649adbaee140a892500e83502bd3ae300/src/pointer-analysis/value_set.cpp#L433-L444

In this method, we need to get the value sets of the original union first.

Value sets in this PR:
```
c:main4.c@283@F@main@initial_cursor?1!0 = { *, <"hello", 0, 1, signed char [6]>, # } 
c:main4.c@216@F@main@u?1!0.__anon_field_1.c = { <"hello", 0, 1, signed char [6]> } 
c:main4.c@216@F@main@u?1!0.cap = { #, <"hello", 0, 1, signed char [6]> } 
c:@__ESBMC_num_threads_running = { # } 
c:main4.c@216@F@main@u?1!0.__anon_field_1.p = { # } 
c:stdlib.c@__ESBMC_atexits = { <0, 0, 1,  struct atexit_key> } 
c:main4.c@176@F@main@cap_ptr?1!0 = { <"hello", 0, 1, signed char [6]> } 
```
Verification result:

```
State 4 file main4.c line 25 column 5 function main thread 0
Violated property:
  file main4.c line 25 column 5 function main


VERIFICATION FAILED
```


lucasccordeiro:
@XLiZHI: can you please run this PR over the SV-COMP benchmarks?

XLiZHI:
This PR: 
```
  correct:           16510
    correct true:    11068
    correct false:    5442
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13762
  Score:             26842 (max: 51772)
  
https://github.com/esbmc/esbmc/actions/runs/11478792353
```

Master: 
```
  correct:           16456
    correct true:    11017
    correct false:    5439
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13816
  Score:             26737 (max: 51772)
  
  https://github.com/esbmc/esbmc/actions/runs/11426032693/job/31788670371
```

mikhailramalho:
I don't really know enough about the deref code to do a review, but why are we always accessing the first member of the constant union?

Shouldn't we read the latest written member instead?

XLiZHI:
> I don't really know enough about the deref code to do a review, but why are we always accessing the first member of the constant union?
> 
> Shouldn't we read the latest written member instead?

I am not familiar with value sets. In my understanding, the symbol that is assigned is a new symbol.
```
turn
  a.c=e
into
  a'==a WITH [c:=e]
```
For example, it seems that a' should be assigned the value of a and then updated.

mikhailramalho:
I meant this:
```
get_value_set_rec(cu.datatype_members[0], dest, suffix, original_type);
```
more specifically:
```
cu.datatype_members[0]
```

You are always using the first member of the union, but I expect that we would read the latest written member.

XLiZHI:
> You are always using the first member of the union, but I expect that we would read the latest written member.

This expression has one and only one member. For the constant union expression, it looks like this:
```c
constant_union
* members : 
  0: address_of
    * pointer_obj : index
        * source_value : constant_string
            * value : hello
            * kind : default
            * type : array
              * subtype : signedbv
                  * width : 8
              * array_size : constant_int
                  * value : 6
                  * type : unsignedbv
                    * width : 64
              * size_is_infinite : false
        * index : constant_int
            * value : 0
            * type : signedbv
              * width : 64
        * type : signedbv
          * width : 8
    * type : pointer
      * subtype : signedbv
          * width : 8

* init_field : cap
* type : union
  * members : 
    0: pointer
      * subtype : empty
    1: struct
      * members : 
        0: unsignedbv
          * width : 64
        1: unsignedbv
          * width : 64

      * member_names : 
        0: c
        1: p

      * member_pretty_names : 
        0: c
        1: p

      * typename : struct __anon_struct_at_main4_c__8_5
      * packed : false

  * member_names : 
    0: cap
    1: __anon_field_1

  * member_pretty_names : 
    0: cap
    1: __anon_field_1

  * typename : union cheri
  * packed : false
```
In the case of unions, their members share addresses, so we only need to get the address of the initialized field. Just like we are similarly handling constant struct ---- to get the address.
https://github.com/esbmc/esbmc/blob/b7e5813eedb31c18ea3a2ba4ce173292e6b8151b/src/pointer-analysis/value_set.cpp#L425-L430

As for address updates, in my investigation, esbmc updates this at the solver level, we don't update it in value sets.

lucasccordeiro (approved):

lucasccordeiro:
@XLiZHI: could you also add a regression test that gives "VERIFICATION SUCCESSFUL"?

regression/bitwuzla/github_1352-success/test.desc
```diff
@@ -1,4 +1,4 @@
 THOROUGH
 gh-1352.c
---unwind 1 --no-unwinding-assertion --force-malloc-success --bitwuzla
```
rafaelsamenezes, line 3: This should be part of another PR.

rafaelsamenezes (changes requested):

lucasccordeiro:
Thanks for sending this PR, @XLiZHI.

rafaelsamenezes (approved):

rafaelsamenezes:
@XLiZHI can you do a new PR adding a teat case with a constant union with more than 2 members and the initialized member not being the first (as @mikhailramalho pointed)? 

We are merging this PR due to it fixing the master branch.

XLiZHI:
> @XLiZHI can you do a new PR adding a teat case with a constant union with more than 2 members and the initialized member not being the first (as @mikhailramalho pointed)? 
> 
> We are merging this PR due to it fixing the master branch.

The members of a union variable share a common first address and occupy the same memory space, so only the value of one of the members can be stored at any one time.

I can't think of a situation where there is more than one member, can you provide one?

rafaelsamenezes:
> The members of a union variable share a common first address and occupy the same memory space, so only the value of one of the members can be stored at any one time.

I see, then its ok. 

We probably should change the constant_union to allow only one member. I don't think that there is anything preventing the generation of a constant union with more than one member. To be fair, at smt_conv we don't seem to care as well

```cpp
  case expr2t::constant_union_id:
  {
    // Get size
    const constant_union2t &cu = to_constant_union2t(expr);
    const std::vector<expr2tc> &dt_memb = cu.datatype_members;
    expr2tc src_expr =
      dt_memb.empty() ? gen_zero(get_uint_type(0)) : dt_memb[0];
#ifndef NDEBUG
    if (!cu.init_field.empty())
    {
      const union_type2t &ut = to_union_type(expr->type);
      unsigned c = ut.get_component_number(cu.init_field);
      /* Can only initialize unions by expressions of same type as init_field */
      assert(src_expr->type->type_id == ut.members[c]->type_id);
    }
#endif
    a = convert_ast(typecast2tc(
      get_uint_type(type_byte_size_bits(expr->type).to_uint64()),
      bitcast2tc(
        get_uint_type(type_byte_size_bits(src_expr->type).to_uint64()),
        src_expr)));
    break;
  }
```

PR #2095, opened 2024-10-24
[symex] make alloca handling more robust

intrigus-lgtm:
Don't free pointers returned from `alloca`te functions. Pointers from functions called `alloca$` will still be incorrectly freed, but this is for a future PR.

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/2095_incorrect_freed_objects_alloca$te/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/2095_incorrect_freed_objects_alloca$te/main.cpp
@@ -1,0 +1,23 @@
+#include <stddef.h>
+#include <stdlib.h>
+struct
+{
+  float *alloca$te()
+  {
+    return static_cast<float *>(malloc(sizeof(float)));
+  }
+} a;
+float *alloca$te()
+{
+  return a.alloca$te();
+}
+float *b()
+{
+  return alloca$te();
+}
+int main()
+{
+  float *x = b();
+  free(x);
+  return 0;
+}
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

XLiZHI (approved):

rafaelsamenezes (approved):

rafaelsamenezes:
Maybe we could replace it with a regex? Something like "alloca$\d+".

lucasccordeiro:
> Maybe we could replace it with a regex? Something like "alloca$\d+".

@intrigus: could you please try this approach?

intrigus-lgtm:
> > Maybe we could replace it with a regex? Something like "alloca$\d+".
> 
> @intrigus: could you please try this approach?

I will not try this approach as it's also not complete, i.e. a function called `alloca$111` would still fail.
I don't have the time for a complete solution and the alloca handling in esbmc looks pretty broken anyway.

The main motivation for this PR is to unblock my c++ code that uses std::allocator<T>.

rafaelsamenezes:
> > > Maybe we could replace it with a regex? Something like "alloca$\d+".
> > 
> > 
> > @intrigus: could you please try this approach?
> 
> I will not try this approach as it's also not complete, i.e. a function called `alloca$111` would still fail. I don't have the time for a complete solution and the alloca handling in esbmc looks pretty broken anyway.

That would generate "alloca$111$<digits>", no?

lucasccordeiro:
> The main motivation for this PR is to unblock my c++ code that uses std::allocator.

Thanks, @intrigus.

I have asked Rafael to submit another PR with his proposal.

lucasccordeiro:

intrigus-lgtm:
> > The main motivation for this PR is to unblock my c++ code that uses std::allocator.
> 
> Thanks, @intrigus.
> 
> I have asked Rafael to submit another PR with his proposal.

Sorry, for not being more helpful, just pretty busy and keeping my thesis' deadline in mind.

PR #2096, opened 2024-10-24
add restriction for constant_union

rafaelsamenezes:
In https://github.com/esbmc/esbmc/pull/2093#issuecomment-2435034215 we claim that there is at most 1 member in a constant union. This adds a explicit restriction for it.


XLiZHI (approved):

lucasccordeiro (approved):

rafaelsamenezes:
Maybe we could replace it with a regex? Something like "alloca$\d+".

PR #2099, opened 2024-10-26
[build] fix compilation with Clang 19

ligurio:
```
  185 |       comp.set_is_unnamed_bitfield(fd.isUnnamedBitfield());
      |                                       ^~~~~~~~~~~~~~~~~
      |                                       isUnnamedBitField
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,8 @@ set (ESBMC_VERSION "${ESBMC_VERSION_MAJOR}.${ESBMC_VERSION_MINOR}.${ESBMC_VERSIO
 # The only default solver available is smtlib
 set (ESBMC_AVAILABLE_SOLVERS "smtlib")
 
+set(MAX_SUPPORTED_LLVM_VERSION_MAJOR 16)
+
 #################################
 # Configuration                 #
 #################################
--- a/scripts/cmake/FindLocalLLVM.cmake
+++ b/scripts/cmake/FindLocalLLVM.cmake
@@ -42,6 +42,11 @@ elseif (DEFINED ESBMC_CHERI AND NOT ESBMC_CHERI AND ESBMC_CHERI_CLANG)
   unset(ESBMC_CHERI_CLANG_MORELLO)
 endif()
 
+if (${LLVM_VERSION_MAJOR} GREATER ${MAX_SUPPORTED_LLVM_VERSION_MAJOR})
+  message(WARNING "LLVM version ${LLVM_VERSION_MAJOR} is greater than maximum "
+                  "supported (${MAX_SUPPORTED_LLVM_VERSION_MAJOR})")
+endif()
+
 if (${LLVM_VERSION_MAJOR} LESS 11)
   message(FATAL_ERROR "Could not find LLVM/Clang >= 11.0 at all: please specify with -DLLVM_DIR/-DClang_DIR")
 else()
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -182,7 +182,11 @@ bool clang_c_convertert::get_decl(const clang::Decl &decl, exprt &new_expr)
       comp.type().width(integer2string(result.Val.getInt().getSExtValue()));
       comp.type().set("#bitfield", true);
       comp.type().subtype() = t;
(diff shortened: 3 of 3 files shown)


mikhailramalho:
What version of clang are you using?

ligurio:
> What version of clang are you using?

Clang 19

XLiZHI:
We currently support up to Clang 16. As far as I know this API change was in the release of Clang 17. If you want to use a higher version, Maybe this will work:
```
#if LLVM_VERSION_MAJOR > 16
#  define isUnnamedBitField(fd) fd->isUnnamedBitField()
#else
#  define isUnnamedBitField(fd) fd->isUnnamedBitfield()
#endif
....

#undef isUnnamedBitField(fd)
```

mikhailramalho:
or better still do:
```
#if LLVM_VERSION_MAJOR > 16
      comp.set_is_unnamed_bitfield(fd.isUnnamedBitField());
#else
      comp.set_is_unnamed_bitfield(fd.isUnnamedBitfield());
#endif
```

ligurio:
The patch updated and force-pushed.

Also, I would a check for the latest support Clang version to `CMakeLists.txt` and show a warning if a version is not in the range of supported versions. What do you think?

lucasccordeiro:
Your suggestion to print a warning message looks good, @ligurio.

ligurio:
> Your suggestion to print a warning message looks good, @ligurio.

@lucasccordeiro, added

lucasccordeiro:
Many thanks, @ligurio.

lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:
Thanks for submitting this PR, @ligurio.

mikhailramalho:
The cmake part should be the other way around, the minimum version is 16, and we should abort cmake config when it's lower than that

lucasccordeiro:
> The cmake part should be the other way around, the minimum version is 16, and we should abort cmake config when it's lower than that

I'm unsure here. As far as I'm aware, we have tested until v16. We didn't test the recent versions (e.g., v19). Please correct me if I'm wrong here.

XLiZHI:
> I'm unsure here. As far as I'm aware, we have tested until v16. We didn't test the recent versions (e.g., v19). Please correct me if I'm wrong here.

Our minimum LLVM version should be 13, which is the latest version of Clang-CHERI. But since our windows version is still using clang 11 (we need to build the windows version manually), we can't remove it yet.

mikhailramalho:
> > The cmake part should be the other way around, the minimum version is 16, and we should abort cmake config when it's lower than that
> 
> I'm unsure here. As far as I'm aware, we have tested until v16. We didn't test the recent versions (e.g., v19). Please correct me if I'm wrong here.

I've been using clang 18 for a while now, and this patch breaks the compilation. This change was introduced for clang 19, not 18.

I find it weird that we now decided to support up to a maximum version rather than a minimum. It was never like that before. We should incentivize people to use newer versions of the lib since (more often than not) they improve performance, add new features, and include fixes.

lucasccordeiro:
Dear @mikhailramalho,

> I've been using clang 18 for a while now, and this patch breaks the compilation. This change was introduced for clang 19, not 18.

That's why @ligurio added a warning message to indicate if a version is not in the range of supported versions.

> We should incentivize people to use newer versions of the lib since (more often than not) they improve performance, add new features, and include fixes.

I fully agree that we should motivate the users to use the latest lib versions. However, we should also check for regression to avoid what happened in your environment, i.e., the proposed patch broke your compilation.

rafaelsamenezes:
> > > The cmake part should be the other way around, the minimum version is 16, and we should abort cmake config when it's lower than that
> > 
> > 
> > I'm unsure here. As far as I'm aware, we have tested until v16. We didn't test the recent versions (e.g., v19). Please correct me if I'm wrong here.
> 
> I've been using clang 18 for a while now, and this patch breaks the compilation. This change was introduced for clang 19, not 18.
> 
> I find it weird that we now decided to support up to a maximum version rather than a minimum. It was never like that before. We should incentivize people to use newer versions of the lib since (more often than not) they improve performance, add new features, and include fixes.

Are you reffering to this line: ``#if LLVM_VERSION_MAJOR > 16`

Which should be 18?

PR #2110, opened 2024-10-31
[python-frontend] Fix filename for imported modules

brcfarias:


lucasccordeiro (changes requested):
Can you add the regression test associated with this PR?

regression/python/try-fail/main.py
```diff
@@ -0,0 +1,3 @@
+from jira import JIRA  # this imported module contains try statements
```
lucasccordeiro, line 1: Where is the try-statement in this code? Did I miss something?

lucasccordeiro (changes requested):

regression/python/try-fail/main.py
brcfarias: The imported code from this module, which can be located at $HOME/lib/python3.10/site-packages/jira/\_\_init\_\_.py on Linux, contains:

```
"""The root of JIRA package namespace."""
from __future__ import annotations

try:
    import importlib.metadata

    __version__ = importlib.metadata.version("jira")
except Exception:
    __version__ = "unknown"

from jira.client import (
    JIRA,
    Comment,
    Issue,
    Priority,
    Project,
    Role,
    User,
    Watchers,
    Worklog,
)
from jira.config import get_jira
from jira.exceptions import JIRAError

__all__ = (
    "Comment",
    "__version__",
    "Issue",
    "JIRA",
    "JIRAError",
    "Priority",
    "Project",
    "Role",
    "User",
    "Watchers",
    "Worklog",
    "get_jira",
)
```

brcfarias: The issue is that for every import of "module," we generate a corresponding "module.json." In the case of an "import jira," the interpreter resolves this to the file <path>/jira/\_\_init\_\_.py. Thus, we were generating an \_\_init\_\_.json, while the Python converter was expecting a jira.json. This import scenario exercises that condition, and I added this check to demonstrate that the module was imported; however, we do not support the try block contained in the code.

lucasccordeiro (approved):
Thanks for this clarification, @brcfarias.

lucasccordeiro:

PR #2113, opened 2024-11-01
Introduce new option to convert pointer and bound checks into assumptions

XLiZHI:
According to CHERI's guarantee, for the checking of capability pointers, the assertions of pointer checking and boundary checking are converted into assumptions.

### DIFF
--- a/regression/esbmc/dangling_assume/main.c
+++ b/regression/esbmc/dangling_assume/main.c
@@ -1,0 +1,11 @@
+#include <stdlib.h>
+
+int main() {
+    int *ptr = (int*) malloc(sizeof(int));
+    *ptr = 42;
+    free(ptr);
+    *ptr = 50;
+
+    return 0;
+}
+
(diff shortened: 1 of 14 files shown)


lucasccordeiro (changes requested):
Can you please add a few regression tests to check this new option?

mikhailramalho (approved):
Overall LGTM, just one point about the simplification of the assumptions: I would always add them without ever trying to simplify, even if we can statically assume they are true.

I would assume that the more constraints we add, even if they are simple, the easier it would be for the solver to solve the problem.

lucasccordeiro (approved):

lucasccordeiro:
@XLiZHI: try to run some experiments about @mikhailramalho's suggestion. If you observe that no simplification gives the best results, you can submit a PR to remove this simplification. The important part now is to start running experiments.

XLiZHI:
> I would assume that the more constraints we add, even if they are simple, the easier it would be for the solver to solve the problem.

I agree that the solver easily solves this kind of problem, but we are currently focusing on the performance using assumption, and I want to keep the same behavior as assertion and avoid unnecessary solving.

PR #2115, opened 2024-11-04
[python-frontend] Filter global variables and classes during function verification

brcfarias:
This PR enhances the verification process with the `--function` option by filtering global variables and classes. Previously converting all global elements could cause verification failures due to unsupported features, even if those elements weren't used in the specified function.

Now we track only the classes and variables used within the function being verified. This tracking is nested, meaning that if function A calls function B, we also verify the global variables and classes used in function B.

### **Example**
```python
x = 1
y = blah()  # function blah is undefined (could be an unsupported feature)

def foo() -> None:
    z = x
```
When running ESBMC with `esbmc --function foo`, the verification should not fail as the variable `y` isn't used in the function `foo()`. Only the global variable `x` is needed for verifying this function.

### DIFF
--- a/regression/python/global/main.py
+++ b/regression/python/global/main.py
@@ -1,0 +1,23 @@
+class OtherClass:
+  pass
+
+class MyClass:
+  m_data: OtherClass
+
+a = 1
+d = 2
+e = 1
+z = blah()
+
+def bar() -> None:
+  f = e
+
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

lucasccordeiro:

Issue #2102, opened 2024-10-28
[SMT incremental] Produce segfault when verifying a multi-threaded program


lucasccordeiro:
$ esbmc main.c --no-div-by-zero-check --force-malloc-success --no-align-check --k-step 2 --floatbv --unlimited-k-steps  --unwind 2 --compact-trace --context-bound 3 --smt-during-symex

````C
#include <stdlib.h>
#include <pthread.h>

// This is based on the weaver benchmark

int lA=2, lB=2;
int *A, *B; // will be initialized with NONDET
int iA, iB; // initialized with: 0


// increments iA until A[iA] != B[iA]
void* thread1() {
  //int c = 0;
  while(iA < lA && iA < lB)
    if(A[iA] == B[iA]) iA++;
    else break;

  return 0;
}

// increments iB until A[iB] != B[iB]
void* thread2() {

  while(iB < lB && iB < lA)
    if(A[iB] == B[iB]) iB++;
    else break;

  return 0;
}


int main() {
  // Nondet arrays
  A = malloc(sizeof(int) * lA);
  B = malloc(sizeof(int) * lB);

  pthread_t t1, t2;
  pthread_create(&t1, 0, thread1, 0);
  pthread_create(&t2, 0, thread2, 0);
  pthread_join(t1, 0);
  pthread_join(t2, 0);

  __ESBMC_assert(iA == iB, "iA and iB must be equal");
}
````

This is a previous PR, where @fbrausse solved a segfault in our SMT incremental: https://github.com/esbmc/esbmc/pull/1915.

XLiZHI:
I learned the SMT code. Under my investigation, I found that the problem comes from assertion.

`__ESBMC_assert(iA == iB, "iA and iB must be equal");` It works just fine if we remove this line.

https://github.com/esbmc/esbmc/blob/1af829d440e4f46af3987c7dd606427f95072593/src/goto-symex/symex_target_equation.cpp#L474-L482
An invalid iterator was encountered here. Earlier, we stored some iterators in a list and extracted them.

https://github.com/esbmc/esbmc/blob/1af829d440e4f46af3987c7dd606427f95072593/src/goto-symex/symex_target_equation.cpp#L365-L378
In this method, we seem to have removed some SSA assertions directly, and if we comment out this method, it works just fine.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/issues/2102.

lucasccordeiro:
[...]

PR #2120, opened 2024-11-06
[SMT incremental] Avoid duplicate erase

XLiZHI:
When we use flag --smt-during-symex, instead of copying the target, we share the original target.

https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/execution_state.cpp#L1271-L1275

Detail: https://github.com/esbmc/esbmc/issues/2102#issuecomment-2447859524

https://github.com/esbmc/esbmc/blob/1af829d440e4f46af3987c7dd606427f95072593/src/goto-symex/symex_target_equation.cpp#L365-L378

This method will erase the assertion, resulting in invalid iterators later.

### DIFF
--- a/src/goto-symex/symex_target_equation.cpp
+++ b/src/goto-symex/symex_target_equation.cpp
@@ -368,7 +368,7 @@ unsigned int symex_target_equationt::clear_assertions()
 
   for (SSA_stepst::iterator it = SSA_steps.begin(); it != SSA_steps.end(); it++)
   {
-    if (it->type == goto_trace_stept::ASSERT && !smt_during_symex)
+    if (it->type == goto_trace_stept::ASSERT)
     {
       SSA_stepst::iterator it2 = it;
       it--;
--- a/src/goto-symex/symex_target_equation.cpp
+++ b/src/goto-symex/symex_target_equation.cpp
@@ -368,9 +368,7 @@ unsigned int symex_target_equationt::clear_assertions()
 
   for (SSA_stepst::iterator it = SSA_steps.begin(); it != SSA_steps.end(); it++)
   {
-    bool user_provided =
-      smt_during_symex ? it->source.pc->location.user_provided() : false;
-    if (it->type == goto_trace_stept::ASSERT && !user_provided)
+    if (it->type == goto_trace_stept::ASSERT && !smt_during_symex)
     {
       SSA_stepst::iterator it2 = it;
       it--;
--- a/src/goto-symex/symex_target_equation.cpp
+++ b/src/goto-symex/symex_target_equation.cpp
@@ -368,7 +368,9 @@ unsigned int symex_target_equationt::clear_assertions()
 
   for (SSA_stepst::iterator it = SSA_steps.begin(); it != SSA_steps.end(); it++)
   {
-    if (it->type == goto_trace_stept::ASSERT)
+    bool user_provided =
+      smt_during_symex ? it->source.pc->location.user_provided() : false;
+    if (it->type == goto_trace_stept::ASSERT && !user_provided)
     {
       SSA_stepst::iterator it2 = it;
       it--;
--- a/src/goto-symex/symex_target_equation.h
+++ b/src/goto-symex/symex_target_equation.h
@@ -25,7 +25,6 @@ public:
     debug_print = config.options.get_bool_option("symex-ssa-trace");
     ssa_trace = config.options.get_bool_option("ssa-trace");
     ssa_smt_trace = config.options.get_bool_option("ssa-smt-trace");
-    smt_during_symex = config.options.get_bool_option("smt-during-symex");
   }
 
   // assignment to a variable - must be symbol
@@ -209,7 +208,6 @@ protected:
   bool debug_print;
   bool ssa_trace;
   bool ssa_smt_trace;
-  bool smt_during_symex;
 
 private:
   void debug_print_step(const SSA_stept &step) const;
--- a/src/goto-symex/symex_target_equation.h
+++ b/src/goto-symex/symex_target_equation.h
@@ -25,6 +25,7 @@ public:
     debug_print = config.options.get_bool_option("symex-ssa-trace");
     ssa_trace = config.options.get_bool_option("ssa-trace");
     ssa_smt_trace = config.options.get_bool_option("ssa-smt-trace");
+    smt_during_symex = config.options.get_bool_option("smt-during-symex");
   }
 
   // assignment to a variable - must be symbol
@@ -208,6 +209,7 @@ protected:
   bool debug_print;
   bool ssa_trace;
   bool ssa_smt_trace;
+  bool smt_during_symex;
 
 private:
   void debug_print_step(const SSA_stept &step) const;
--- a/regression/incremental-smt/github_2102/main.c
+++ b/regression/incremental-smt/github_2102/main.c
@@ -1,0 +1,48 @@
+#include <stdlib.h>
+#include <pthread.h>
+
+// This is based on the weaver benchmark
+
+int lA = 2, lB = 2;
+int *A, *B; // will be initialized with NONDET
+int iA, iB; // initialized with: 0
+
+// increments iA until A[iA] != B[iA]
+void *thread1()
+{
+  //int c = 0;
+  while (iA < lA && iA < lB)
+    if (A[iA] == B[iA])
+      iA++;
+    else
+      break;
+
+  return 0;
+}
+
+// increments iB until A[iB] != B[iB]
+void *thread2()
+{
+  while (iB < lB && iB < lA)
+    if (A[iB] == B[iB])
+      iB++;
+    else
+      break;
+
+  return 0;
+}
+
+int main()
+{
+  // Nondet arrays
+  A = malloc(sizeof(int) * lA);
+  B = malloc(sizeof(int) * lB);
+
+  pthread_t t1, t2;
+  pthread_create(&t1, 0, thread1, 0);
+  pthread_create(&t2, 0, thread2, 0);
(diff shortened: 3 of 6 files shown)


src/goto-symex/symex_target_equation.cpp
```diff
@@ -368,7 +368,9 @@ unsigned int symex_target_equationt::clear_assertions()
 
   for (SSA_stepst::iterator it = SSA_steps.begin(); it != SSA_steps.end(); it++)
   {
-    if (it->type == goto_trace_stept::ASSERT)
+    bool user_provided =
```
lucasccordeiro: This is confusing. You claim that if `smt_during_symex` is enabled, the user will never provide the assertion.

I think the idea here is to avoid erasing the assertion when doing incremental SMT, right? If so, why don't you simply use the flag  `smt_during_symex`  in the if-condition?

src/goto-symex/symex_target_equation.h
```diff
@@ -208,6 +209,7 @@ class symex_target_equationt : public symex_targett
   bool debug_print;
   bool ssa_trace;
   bool ssa_smt_trace;
+  bool smt_during_symex;
```
lucasccordeiro: We declare `bool smt_during_symex;` in `execution_state.h`. Do we need to declare it twice?

lucasccordeiro (changes requested):

src/goto-symex/symex_target_equation.h
XLiZHI: There is no inheritance between `symex_target_equation.h` and `execution_state.h`, I guess this decl is necessary.

src/goto-symex/symex_target_equation.cpp
lucasccordeiro: When doing incremental SMT, the assertion is indeed needed for lemma learning.

XLiZHI: Sorry, the variable name is misleading. My question is whether certain assertions, like internally generated checks (e.g., bounds checks), need to be removed during incremental BMC?

lucasccordeiro: No, they don't need to be removed. These assertions can help the SMT solver when performing incremental learning.

lucasccordeiro (approved):
LGTM.

mikhailramalho:
Can I get more details on the problem?

When poping a context, we should get rid of the assertions added after the last push, are we keeping the popped assertions somewhere else?

XLiZHI:
> When poping a context, we should get rid of the assertions added after the last push, are we keeping the popped assertions somewhere else

In my understanding, when `--smt-during-symex` is not used, these assertions are not kept elsewhere, this may be due to the fact that they have been executed before.

lucasccordeiro:
The assertions are removed here: https://github.com/esbmc/esbmc/blob/master/src/goto-symex/reachability_tree.cpp#L257, which might be too early for the incremental SMT checking.

mikhailramalho:
I didn't understand your answer, please clarify.

When we pop a context (i.e., when using --smt-during-symex) these assertions need to be removed. They need to be gone when calling pop(). 

From what I can understand here, the assertions are removed during pop but we have pointers/iterators pointing to them, is that what you are trying to fix? If that's the case, why are we not updating wherever the iterators are coming from?

mikhailramalho:
> The assertions are removed here: [`master`/src/goto-symex/reachability_tree.cpp#L257](https://github.com/esbmc/esbmc/blob/master/src/goto-symex/reachability_tree.cpp?rgh-link-date=2024-11-06T16%3A01%3A48Z#L257), which might be too early for the incremental SMT checking.

What do you mean by too early? We are resetting the state so we can go back to the previous interleaving; it should work just like the incremental smt, we are poping the assertions from this path

lucasccordeiro:
> What do you mean by too early? We are resetting the state so we can go back to the previous interleaving; it should work just like the incremental smt, we are poping the assertions from this path.

Are we really popping from this point? If so, we might still have pointers/iterators pointing to these assertions.

XLiZHI:
1. First I think for `--smt-during-symex`, we push assertions to see if each step violates the assertion. Instead of using a copy of the target, we use the target directly. is that correct?
2. https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/execution_state.cpp#L1261-L1262 When the destructor is called, erase all stored iterators. If we removed assertions earlier, there would be invalid iterators here. Because we're using list to store copies of iterators. https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/symex_target_equation.h#L238

lucasccordeiro:
Should pop be called here https://github.com/esbmc/esbmc/blob/master/src/goto-symex/reachability_tree.cpp?rgh-link-date=2024-11-06T16%3A01%3A48Z#L257?

lucasccordeiro:
This line was added by @jmorse 11 years ago :-)

image

mikhailramalho:
> > What do you mean by too early? We are resetting the state so we can go back to the previous interleaving; it should work just like the incremental smt, we are poping the assertions from this path.
> 
> Are we really popping from this point? If so, we might still have pointers/iterators pointing to these assertions.

We should be doing that, pop a context will remove all assertions added after the last push.

> 1. First I think for `--smt-during-symex`, we push assertions to see if each step violates the assertion. Instead of using a copy of the target, we use the target directly. is that correct?

We should have a single target, i.e., a single equation; we are always adding and removing assertions to it to check if the equations still holds.

Think of it as a DFS in a graph: the equation is the path from the root to a leaf node and each node is a new context (and a set of assertions) added to the equations. When we pop, the assertions are removed but we keep the assertions added in the previous nodes.

> 2. https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/execution_state.cpp#L1261-L1262
>     When the destructor is called, erase all stored iterators. If we removed assertions earlier, there would be invalid iterators here. Because we're using list to store copies of iterators. https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/symex_target_equation.h#L238

IIUC, we remove the assertions when popping, but keep pointers/iterators pointing to them, and then we double-free that in the destructor. This sounds like we should erase all the pointers/iterators from wherever they are stored when popping.

XLiZHI:
> IIUC, we remove the assertions when popping, but keep pointers/iterators pointing to them, and then we double-free that in the destructor. This sounds like we should erase all the pointers/iterators from wherever they are stored when popping.

This was my initial solution, I tried to remove the corresponding iterator, but after removing it we need to push a new iterator to `scoped_end_points`. I don't have a solution yet.

> We should have a single target, i.e., a single equation; we are always adding and removing assertions to it to check if the equations still holds.

This is what I want to say, we only have one target, and that target is not copied, it's main target. What puzzles me about this is that for the main target, if we just remove all the assertions from the SSA, will there be cases where we can't find the assertions in the subsequent states?

I'm still learning this part of the code, can you please walk me through it?

XLiZHI:
In smt-during-symex mode, uses a shared target approach. Instead of creating a new target with each clone, it pushes a context (push_ctx()) on the same target. This shared approach ensures that assertions from each path are retained and verified throughout the entire execution process. Therefore, user-defined assertions are immediately added to the target's SSA, allowing them to be verified at each step of symbolic execution. The purpose of push_ctx() is to save the current state of the context, including the path’s user assertions, so they can be restored during backtracking.

In standard mode, its clone operation creates a new object as the target for path exploration, and it includes an assert(SSA_steps.size() == 0) check to ensure that SSA_steps is empty when cloned. 

The difference between the two is that one converts user-defined assertions to SSA for real-time execution, and the other is unified execution. And I also found that user-defined assertions only appear in SSA_steps when `smt-during-symex` is used.  To sum up, I think there is no need to clear user-defined assertions in `symex_target_equationt::clear_assertions()`, as it will be done automatically in pop:
https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-symex/execution_state.cpp#L1261-L1262

lucasccordeiro:
Hi guys,

I have just checked the push/pop implementation. The `clear_assertions()` method seems to be called only at https://github.com/esbmc/esbmc/blob/master/src/goto-symex/reachability_tree.cpp#L257.

From what I understand (you can correct me here), ESBMC already manages the stack efficiently. It can push a context, add assertions, check satisfiability, and then pop the context if the assertions are no longer needed, enabling temporary use without permanent retention of the assertions.

My recommendation for this PR would be to do something like:

````CPP
  if (execution_states.size() != 0 && !smt_during_symex)
  {
    // When backtracking, erase all the assertions from the equation before
    // continuing forwards. They've all already been checked, in the trace we
    // just backtracked from. Thus there's no point in checking them again.
    symex_target_equationt *eq =
      static_cast<symex_target_equationt *>((*cur_state_it)->target.get());
    unsigned int num_asserts = eq->clear_assertions();

    // Remove them from the count of remaining assertions to check. This allows
    // for more traces to be discarded because they do not contain any
    // unchecked assertions.
    (*cur_state_it)->total_claims -= num_asserts;
    (*cur_state_it)->remaining_claims -= num_asserts;
  }

  return execution_states.size() != 0;
}
````

This patch keeps our regression.

lucasccordeiro:
This PR is a hotfix for this issue https://github.com/esbmc/esbmc/issues/2102.

In the future, we must reevaluate how the SMT incremental is implemented in ESBMC.

rafaelsamenezes (approved):

Issue #2109, opened 2024-10-31
Successful verification with `--claim`, but failure without


intrigus-lgtm:
I'm trying to verify a moderately big c++ program.
When running it normally (i.e. without `--claim`) I get a counter example from let's say claim 72.
When I'm now running esbmc again, but add `--claim 72`, it verifies succesfully.
I find this very weitd.
Am I missing something?

(I can not share the code right now, because I use a pretty custom fork of esbmc and won't have access to the code till next monday.)

intrigus-lgtm:
Anybody got an idea whether this is a bug or a mistake on my side?

lucasccordeiro:
@intrigus-lgtm: what is the type of property for claim 72 (memory safety, overflow, concurrency, user-specified assertion)?

intrigus-lgtm:
It's a user-specified assertion.

lucasccordeiro:
> It's a user-specified assertion.

You have found a bug in ESBMC. If you could send us a reduced version of the program, that would allow us to inspect the ESBMC source code and identify the root cause of this issue.

Does this issue also happen with the option `--multi-property`?

intrigus-lgtm:
~~Okay the boring resolution of my problem is that `--show-claims` is one-based, while `--claim` expects the index to be zero-based...~~
EDIT: That's was actually only a coincidence. The actual problem is described in https://github.com/esbmc/esbmc/pull/2121.

@lucasccordeiro any objections to change `--show-claims` to be zero-based as well?
I'd say that this should break the least code hopefully.

lucasccordeiro:
> any objections to change --show-claims to be zero-based as well?

@intrigus-lgtm: No objections at all ;-)

Can you please submit a PR to fix this issue?

intrigus-lgtm:
I submitted https://github.com/esbmc/esbmc/pull/2121 which kind of fixes the issue, although my previous conclusion was wrong.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2121.

PR #2121, opened 2024-11-07
Revert "Jimple throws are now shown at --show-claims"

intrigus-lgtm:
Reverts esbmc/esbmc#1632

This workaround is the reason for https://github.com/esbmc/esbmc/issues/2109
This is because we also add claims for THROW instructions:
https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-programs/show_claims.cpp#L32-L40
But don't account for them here:
https://github.com/esbmc/esbmc/blob/ea42da3613c6027f0ac81e08f3ac74850e032205/src/goto-programs/set_claims.cpp#L19-L25
This then causes a mismatch.

<details>
<summary>Test</summary>

```cpp
#include <cassert>
void unused()
{
  throw;
}

int main()
{
  assert(0);
}

void unused2()
{
  assert(1);
}
```
</details>

One could "fix" this by adding THROW instructions to `set_claims`, but this would be wrong.
We can't just replace a throw in C++ with a skip.

### DIFF
--- a/src/goto-programs/show_claims.cpp
+++ b/src/goto-programs/show_claims.cpp
@@ -27,20 +27,6 @@ void show_claims(
         description,
         from_expr(ns, identifier, instruction.guard));
     }
-
-#ifdef ENABLE_JIMPLE_FRONTEND
-    // In jimple asserts are modelled as throws (and try/catch is not really supported)
-    if (instruction.is_throw())
-    {
-      count++;
-      log_status(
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

rafaelsamenezes (approved):

lucasccordeiro:

PR #2122, opened 2024-11-07
[C++ verification] relax type assertions for vtable variable symbols

XLiZHI:
```
Class BaseClass {
public:
    virtual BaseClass* foo(bool var = false) = 0;
};

class DerivedClass : public BaseClass {
public:
    DerivedClass* foo(bool var = false);
};
```
This conversion is a covariant return type, which allows the override function of a derived class to return a pointer or reference of the derived class type instead of a pointer or reference of the base class type.

I think the current assertions in debug builds are too restrictive and should be relaxed.


mikhailramalho (approved):
LGTM

lucasccordeiro (approved):

lucasccordeiro:

PR #2123, opened 2024-11-07
Revert "[C++ verification] relax type assertions for vtable variable symbols"

intrigus-lgtm:
Reverts esbmc/esbmc#2122

Please give a bit more time when reviewing a PR.
This assert has actually saved me many times when developing ESBMC and `assert(value.type().id() == compo.type().id());` is actually not very useful as an assert in my opinion.

Instead, I'd propose to replace this assert with checking whether `value.type()` is a subtype of `compo.type()`.


intrigus-lgtm:
I'll add a commit that does this subtype checking shortly.

mikhailramalho:
> I'll add a commit that does this subtype checking shortly.

No need to revert the patch if you are going to submit another PR fixing the issue. Also make sure you don't remove the new test added by the previous PR.

mikhailramalho:
Closing as part of stale-branch cleanup. PR #2122 (which this PR reverts) was merged 2024-11-07. If the vtable type-assertion relaxation is still problematic, a fresh issue + investigation against current master would be more productive than a year-stale revert. The branch will be deleted.

PR #2124, opened 2024-11-08
[coverage] initialize pointer in function mode

ChenfengWei0:
```c++
typedef struct
{
  time_t endTs;
  Precipitation precip;
} WeatherObservation;

static float calculateTotalPrecipitation(WeatherObservation *observations,
                                         int dayOffset,
                                         int timezoneOffset)
{
  if (!observations)
    return 0.0f;

  time_t cutoff = time(NULL) - (dayOffset * 24 * 60 * 60);
  float total = 0.0f;

  for (int i = 0; (observations)[i].endTs != 0; i++)
  {
    if ((observations)[i].endTs >= cutoff)
    {
      total += (observations)[i].precip.total;
    }
  }
  return total;
}
```

---

```
State 1 file 1.c line 15 column 1 thread 0
  observations#temp_bool = 0

State 2 file 1.c line 15 column 1 thread 0
  observations#temp = { .endTs=7, .precip={ .total=nil }, .anon_pad$2=0 }

State 3 file 1.c line 25 column 3 function calculateTotalPrecipitation thread 0
Violated property:
  file 1.c line 25 column 3 function calculateTotalPrecipitation
  !((observations + (signed long int)i)->endTs != 0)
  (observations + (signed long int)i)->endTs != 0

[Counterexample]


State 1 file 1.c line 15 column 1 thread 0
  observations#temp_bool = 0

State 2 file 1.c line 15 column 1 thread 0
  observations#temp = { .endTs=0, .precip={ .total=nil }, .anon_pad$2=0 }

State 3 file 1.c line 19 column 3 function calculateTotalPrecipitation thread 0
Violated property:
  file 1.c line 19 column 3 function calculateTotalPrecipitation
  (_Bool)observations != 0
  !((_Bool)observations != 0)
```

### DIFF
--- a/regression/goto-coverage/github_2124_1/test.desc
+++ b/regression/goto-coverage/github_2124_1/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.c
---condition-coverage --unwind 1 --function handleRunplanEvent --no-standard-check --no-simplify
+--condition-coverage-claims --unwind 1 --function handleRunplanEvent --no-standard-check --no-simplify
 ^Total Conditions:  8
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_1/test.desc
+++ b/regression/goto-coverage/github_2124_1/test.desc
@@ -1,7 +1,5 @@
 CORE
 main.c
 --condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
-^Condition Properties - SATISFIED:  3
-^Condition Properties - UNSATISFIED:  1
-^Condition Coverage: 37.5%
+^Total Conditions:  8
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_1/test.desc
+++ b/regression/goto-coverage/github_2124_1/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
+^Condition Properties - SATISFIED:  3
+^Condition Properties - UNSATISFIED:  1
+^Condition Coverage: 37.5%
+^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_1/test.desc
+++ b/regression/goto-coverage/github_2124_1/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.c
---condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
+--condition-coverage --unwind 1 --function handleRunplanEvent --no-standard-check --no-simplify
 ^Total Conditions:  8
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_2/test.desc
+++ b/regression/goto-coverage/github_2124_2/test.desc
@@ -1,5 +1,5 @@
 CORE
 main.c
---condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
+--condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check --no-simplify
 ^Total Conditions:  8
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_2/test.desc
+++ b/regression/goto-coverage/github_2124_2/test.desc
@@ -1,7 +1,5 @@
 CORE
 main.c
 --condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
-^Condition Properties - SATISFIED:  3
-^Condition Properties - UNSATISFIED:  1
-^Condition Coverage: 37.5%
+^Total Conditions:  8
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_2124_2/test.desc
+++ b/regression/goto-coverage/github_2124_2/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check
+^Condition Properties - SATISFIED:  3
+^Condition Properties - UNSATISFIED:  1
+^Condition Coverage: 37.5%
+^VERIFICATION FAILED$
--- a/src/goto-programs/assign_params_as_non_det.cpp
+++ b/src/goto-programs/assign_params_as_non_det.cpp
@@ -58,29 +58,52 @@ bool assign_params_as_non_det::runOnFunction(
     exprt lhs = symbol_expr(*context.find_symbol(_id));
 
     typet l_t = lhs.type();
+
+    // rhs
+    exprt rhs = exprt("sideeffect", l_t);
+    rhs.statement("nondet");
+    rhs.location() = l;
+
+    // assignment
+    goto_programt tmp;
+    goto_programt::targett assignment = tmp.add_instruction(ASSIGN);
+    assignment->location = l;
+    assignment->function = it->location.get_function();
+
+    code_assignt code_assign(lhs, rhs);
+    code_assign.location() = it->location;
+    expr2tc new_assign_expr;
+    migrate_expr(code_assign, new_assign_expr);
+    assignment->code = new_assign_expr;
+
+    // insert
+    goto_program.insert_swap(it++, *assignment);
+    --it;
+
+    // special handling for non-void pointer
     if (l_t.is_pointer() && l_t.subtype() != empty_typet())
     {
       /*
         e.g. int* lhs;
         to
-        lhs = null;
-        bool temp;
-        temp = nondet();
-        if(temp)
+        bool lhs#temp_bool;
+        lhs#temp_bool = nondet();
+        if(lhs#temp_bool)
         {
-          int temp2;
-          temp2 = nondet();
-          assume(temp2 != 0);
-          lhs = &temp2;
+          int lhs#temp;
+          lhs#temp = nondet();
+          assume(lhs#temp != 0);
+          lhs = &lhs#temp;
         }
         
         During this process we create two auxiliary variable. One for the boolean flag, and one for the object.
         - We need to consider the situation where pointers can be null, thus we put the assignment under an if-statement
-        - If the flag (`temp`) is true, we create an object (`temp2`) and assign its address to the pointer.
+        - If the flag (`lhs#temp_bool`) is true, we create an object (`lhs#temp`) and assign its address to the pointer.
       */
 
       // lhs = null;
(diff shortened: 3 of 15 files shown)


ChenfengWei0:
Update

```c++
typedef struct
{
  time_t endTs;
  Precipitation precip;
} WeatherObservation;

static float calculateTotalPrecipitation(WeatherObservation *observations)
{
  if (!observations)
    return 0.0f;
}
```

---

```
[Counterexample]


State 1 file 1.c line 15 column 1 thread 0
  observations#temp_bool = 1

State 2 file 1.c line 17 column 3 function calculateTotalPrecipitation thread 0
Violated property:
  file 1.c line 17 column 3 function calculateTotalPrecipitation
  !((_Bool)observations != 0)
  (_Bool)observations != 0

Solving claim '(_Bool)observations != 0' with solver Boolector 3.2.2
Encoding to solver time: 0.001s
Solving with solver Boolector 3.2.2
Runtime decision procedure: 0.009s

[Counterexample]


State 1 file 1.c line 15 column 1 thread 0
  observations#temp_bool = 0

State 2 file 1.c line 15 column 1 thread 0
  observations#temp = { .endTs=0, .precip={ .total=7 }, .anon_pad$2=1 }

State 4 file 1.c line 17 column 3 function calculateTotalPrecipitation thread 0
Violated property:
  file 1.c line 17 column 3 function calculateTotalPrecipitation
  (_Bool)observations != 0
  !((_Bool)observations != 0)


[Coverage]

  !((_Bool)observations != 0)   file 1.c line 17 column 3 function calculateTotalPrecipitation : SATISFIED
  (_Bool)observations != 0      file 1.c line 17 column 3 function calculateTotalPrecipitation : SATISFIED
Reached Conditions:  2
Total Conditions:  2

Condition Properties - SATISFIED:  2
Condition Properties - UNSATISFIED:  0

Condition Coverage: 100%

VERIFICATION FAILED
```

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,200 @@ bool assign_params_as_non_det::runOnFunction(
...
+      // bool temp;
+      // temp = nondet();
+      // if(temp)
+      // {
+      //   int temp2;
+      //   temp2 = nondet();
+      //   assume(temp2 != 0);
+      //   lhs = &temp2;
+      // }
+
```
ChenfengWei0, line 105: we need to add a type check here. If it's a void pointer we cannot dereference it

sshivaji: Yes, How to cover this, just put in a warning? Saying that "Pointer details cannot be accessed because a void pointer is in the params, please change method to not have void* if you want to see their contents" maybe?

intrigus-lgtm (commented):
I guess I'm somewhat missing the big picture here, but why is this needed?
What prevents us from adding a small main function that instead calls the target function?
Those four special cases in goto_coverage also don't really look pretty IMHO.

sshivaji:
The idea is we want function coverage. If there are 20 methods in a file,
rewriting the code to call main function for every method is painful. The
idea of this approach is to get the pointer values during function calls so
that we can understand the values used during function coverage and we can
reuse these to generate test cases.

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,201 @@ bool assign_params_as_non_det::runOnFunction(
...
+    {
+      // e.g. int* lhs;
+      // to
+      // lhs = null;
+      // bool temp;
+      // temp = nondet();
+      // if(temp)
+      // {
+      //   int temp2;
+      //   temp2 = nondet();
```
lucasccordeiro: Why do we need a second nondet variable?

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,201 @@ bool assign_params_as_non_det::runOnFunction(
...
+      goto_programt tmp4;
+      goto_programt::targett if_statement = tmp4.add_instruction();
+      if_statement->location = l;
+      if_statement->function = it->location.get_function();
+
+      if_statement->make_goto(it);
+      if_statement->guard = guard;
+
+      // insert
+      --it;
```
lucasccordeiro, line 265: Why do we need to repeat --it four times?

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,201 @@ bool assign_params_as_non_det::runOnFunction(
...
+      --it;
+      --it;
+      --it;
+      goto_program.insert_swap(it++, *if_statement);
+      ++it;
```
lucasccordeiro, line 274: Why do we need to repeat ++it three times?

lucasccordeiro (changes requested):
@ChenfengWei0: could you please answer my questions?

src/goto-programs/assign_params_as_non_det.cpp
ChenfengWei0: One for the boolean flag, and one for the object.
- We need to consider the situation where pointers can be null, thus we put the assignment under an if-statement
- If the flag (`temp`) is true, we create an object (`temp2`) and assign its address to the pointer.

src/goto-programs/assign_params_as_non_det.cpp
ChenfengWei0:
```
Instrument_1
Instrument_2  <-- we want to insert a statement before here
Instrument_3
Instrument_4  
Origin_1        <-- 'it' is currently here
```

src/goto-programs/assign_params_as_non_det.cpp
ChenfengWei0: After insertion, reset the pointer `it`'s position to the `origin_1`

sshivaji:
From my perspective, this output is very useful for test generation and for
customer use. Technical optimization points aside .. :)

Em ter., 12 de nov. de 2024 09:31, Chenfeng Wei ***@***.***>
escreveu:

src/goto-programs/assign_params_as_non_det.cpp
lucasccordeiro: Can you document it?

src/goto-programs/assign_params_as_non_det.cpp
lucasccordeiro: Can you document it?

src/goto-programs/assign_params_as_non_det.cpp
lucasccordeiro: Can you write your explanation as a comment?

lucasccordeiro (approved):

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,217 @@ bool assign_params_as_non_det::runOnFunction(
...
+        bool temp;
+        temp = nondet();
+        if(temp)
+        {
+          int temp2;
+          temp2 = nondet();
+          assume(temp2 != 0);
+          lhs = &temp2;
+        }
+
```
rafaelsamenezes, line 100: Would you mind making your instrumentation comments match the name of these variables?

src/goto-programs/assign_params_as_non_det.cpp
```diff
@@ -36,24 +57,217 @@ bool assign_params_as_non_det::runOnFunction(
...
+      goto_program.insert_swap(it++, *zero_assignment);
+      --it;
+
+      // get subType() => int
+      typet subt = l_t.subtype();
+      // if it's symbol, get the original type
+      if (subt.is_symbol())
+        subt = context.find_symbol(subt.identifier())->type;
+
+      // create obj and move it to the symbol table
```
rafaelsamenezes, line 134: I am not sure if this is your case, but what if the parameter is an array given as a pointer? 
```c
void sort(int *ptr, int n) { 
  ptr[1] = ptr[2];
}
```

I guess this could be sorted out in the future using an infinite array but for now could you please add some KNOWNBUG test cases (if they actually are)?

rafaelsamenezes (changes requested):
Just asking for better comments

lucasccordeiro:
@ChenfengWei0: could you please address @rafaelsamenezes' comments?

ChenfengWei0:
> @ChenfengWei0: could you please address @rafaelsamenezes' comments?

I found some strange cases that would break the coverage. Right now I am investigating.

src/goto-programs/assign_params_as_non_det.cpp
ChenfengWei0: I change the comments instead

src/goto-programs/assign_params_as_non_det.cpp
ChenfengWei0: I did some tests and it seems to work fine. 

So if there is a pointer parameter, we do
```c++
void sort(int *ptr) { 
  // e.g. int x; sort(&x);
  // or int x[10]; sort(x);
  ptr = nondet();  // this should cover the array case?
  bool ptr#temp_bool;
  ptr#temp_bool = nondet();
  if(!ptr#temp_bool)
  {
     int ptr#temp;  //This would make more sense for struct type as we are creating an object
     ptr#temp = nondet();
     assume(ptr#temp);
     ptr = &ptr#temp;
  }
}

```

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

regression/goto-coverage/github_2124_1/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
+main.c
+--condition-coverage-claims --unwind 1 --function handleRunplanEvent --no-standard-check --no-simplify
```
intrigus-lgtm, line 3: Why is this being run with `--no-simplify`?

regression/goto-coverage/github_2124_2/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
+main.c
+--condition-coverage --unwind 2 --function handleRunplanEvent --z3 --no-standard-check --no-simplify
```
intrigus-lgtm, line 3: Same question here.

regression/goto-coverage/github_2124_1/test.desc
ChenfengWei0: I don't remember

regression/goto-coverage/github_2124_2/test.desc
ChenfengWei0: I don't remember

PR #2128, opened 2024-11-09
Constant propagation of with2t expression

XLiZHI:


lucasccordeiro (approved):

lucasccordeiro:
@XLiZHI: can you run this PR over the SV-COMP benchmarks?

src/goto-symex/goto_symex_state.cpp
```diff
@@ -141,7 +149,7 @@ bool goto_symex_statet::constant_propagation(const expr2tc &expr) const
     bool noconst = true;
 
     expr->foreach_operand([this, &noconst](const expr2tc &e) {
-      if (noconst && !constant_propagation(e))
+      if (noconst && (!constant_propagation(e) || is_with2t(e)))
```
mikhailramalho: is this needed?

src/util/simplify_expr2.cpp
```diff
@@ -890,6 +890,26 @@ expr2tc member2t::do_simplify() const
...
+  {
+    auto current_source = source_value;
+
+    // Traverse through nested 'with' expressions
+    while (is_with2t(current_source))
+    {
+      const auto &with = to_with2t(current_source);
+
+      // Check if the member matches the update field
+      if (member == to_constant_string2t(with.update_field).value)
```
mikhailramalho: is `with.update_field` always a constant_string? If it isn't, you need to check it before calling `to_constant_string`

mikhailramalho (commented):
Please check if this doesn't make performance worse, we saw regressions before when enabling constant folding of with expressions

src/goto-symex/goto_symex_state.cpp
XLiZHI: Yes, It is necessary and without it there will be serious performance issues.

src/util/simplify_expr2.cpp
XLiZHI: It is contant_string.

regression/esbmc/github_690_z3_cast_success/test.desc
```diff
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
---ir --z3
+--ir --z3 -Wno-error=implicit-function-declaration
```
mikhailramalho, line 3: This is an unrelated change

regression/esbmc-cpp/cpp/ch16_9/test.desc
```diff
@@ -1,4 +1,4 @@
 KNOWNBUG
 main.cpp
---unwind 10 --no-unwinding-assertions --timeout 900
+--unwind 2 --no-unwinding-assertions --timeout 900
```
mikhailramalho: is this related to this patch?

regression/esbmc/github_2119/main.c
```diff
@@ -0,0 +1,26 @@
...
+typedef struct buffer buffer;
+
+int main () {
+  int counter = 6;
+  buff.size = 6;
+  buff.limit = 3;
+
+  for (int i = 0; i < buff.size; i++) {
+    if(buff.limit == i && !buff.counter[i])
+      counter--;
```
mikhailramalho: whitespace

src/goto-symex/goto_symex_state.cpp
mikhailramalho: Then at least add a comment explaining why this is here. Can you share an example where this is required?

src/util/simplify_expr2.cpp
mikhailramalho: We don't have this constraint written anywhere, `update_field` is declared as an `expr2tc`, so let's add a check here.

mikhailramalho (changes requested):
The new test case you add works just fine for me without your patch: 
```
$ esbmc main.c
ESBMC version 7.7.0 64-bit x86_64 linux
Parsing main.c
Converting
Generating GOTO Program
GOTO program creation time: 0.398s
Unwinding loop 2 iteration 1   file main.c line 17 column 3 function main
Unwinding loop 2 iteration 2   file main.c line 17 column 3 function main
Unwinding loop 2 iteration 3   file main.c line 17 column 3 function main
Unwinding loop 2 iteration 4   file main.c line 17 column 3 function main
Unwinding loop 2 iteration 5   file main.c line 17 column 3 function main
Unwinding loop 2 iteration 6   file main.c line 17 column 3 function main
Symex completed in: 0.007s (22 assignments)
Slicing time: 0.000s (removed 21 assignments)
Generated 13 VCC(s), 0 remaining after simplification (1 assignments)
BMC program time: 0.007s

VERIFICATION SUCCESSFUL
```
Please add a test case where we can see this patch in action

regression/esbmc/github_690_z3_cast_success/test.desc
XLiZHI: This is relevant. It passed in llvm13 in this PR, but llvm16 has stricter checks and therefore needs to add this flag.

intrigus-lgtm: Just curious, shouldn't this have been caught by CI when updating to llvm16?
Why only in this PR?

lucasccordeiro:
@XLiZHI : Our benchexec is now running again. Could you please run this PR over the SV-COMP benchmarks?

src/goto-symex/goto_symex_state.cpp
mikhailramalho: Use the comment to describe why this is needed. It is obvious that we skip `with2t`s but why are we doing it?

regression/esbmc/github_690_z3_cast_success/test.desc
mikhailramalho: IIRC, we ignore the results if it's marked as a KNOWBUG

lucasccordeiro:
This PR (https://github.com/esbmc/esbmc/actions/runs/11776848285):

````
  correct:           16446
    correct true:    11011
    correct false:    5435
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13826
  Score:             26721 (max: 51772)
````

Master ( https://github.com/esbmc/esbmc/actions/runs/11529888639):

````
  correct:           16532
    correct true:    11085
    correct false:    5447
  incorrect:            28
    incorrect true:     18
    incorrect false:    10
  unknown:           13740
  Score:             26881 (max: 51772)
````

@XLiZHI: could you please check the `unknown` cases to confirm whether they are crashes or timeouts?

mikhailramalho (approved):
LGTM, let's just update the comment.

lucasccordeiro:
@mikhailramalho: do you refer to this comment?

image

mikhailramalho:
No, this one:

image

It tells me nothing about why we are only focusing on structs. I would expect a comment saying that enabling others will regress performance and hopefully a test case where this regression is reproducible.

XLiZHI:
> It tells me nothing about why we are only focusing on structs. I would expect a comment saying that enabling others will regress performance and hopefully a test case where this regression is reproducible.

I have removed this change.

lucasccordeiro:

PR #2131, opened 2024-11-11
[bitwuzla] get raw string value in bv instead of smt2-formatted

intrigus-lgtm:
If we use `bitwuzla_term_to_string` we'll get e.g. ``#b010101` as a result which we will pass to binary2integer which does not handle strings that start with `#b` properly.
Instead, we use `bitwuzla_term_value_get_str` because it would just return `010101` instead of the unwanted prefix. **The end result of the old behavior was incorrect values (every value was just zero) in traces!**

### DIFF
--- a/regression/bitwuzla/get_model_values/main.cpp
+++ b/regression/bitwuzla/get_model_values/main.cpp
@@ -1,0 +1,63 @@
+int Contains(float value)
+{
+  for (;;)
+    ;
+}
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

mikhailramalho (approved):
LGTM

lucasccordeiro:

PR #2133, opened 2024-11-11
Fix #1976: improve __builtin_{memcpy|memmove}

### LINKED ISSUE #1976: [clang-cpp] `__builtin_memcpy` is not useable without also including `memcpy` header
```cpp
#include <cassert>

int main() {
  int src = 1;
  int dest = 2;
  __builtin_memcpy(&dest, &src, sizeof(int));
  assert(dest == 1);
}
```
fails with:
```
Parsing foo.cpp
foo.cpp:6:3: error: use of undeclared identifier 'memcpy'
  __builtin_memcpy(&dest, &src, sizeof(int));
  ^
<command line>:6:26: note: expanded from here
#define __builtin_memcpy memcpy
                         ^
ERROR: PARSING ERROR
```

This is because https://github.com/esbmc/esbmc/pull/572 simply defines `__builtin_memcpy` as `memcpy` which fails when `memcpy` is not included.
Manually adding includes is not a solution, because this problem happens in a system header file.
Related: https://github.com/esbmc/esbmc/issues/1669

intrigus-lgtm:
> Implement `__builtin_{memcpy|memmove}` by defining them to be equal to `__esbmc_builtin_{memcpy|memmove}`.
We have to do this, because clang forbids us from directly implementing a function called `__builtin_{memcpy|memmove}`.
`__esbmc_builtin_{memcpy|memmove}` then simply delegates to `{memcpy|memmove}`.
We add the functions to the internal additions header, because they are builtins and should be available without
having to import any headers. (The implementation of string.c is available implicitly).

This fixes #1976 and also improves the solution to #569.
If there is a simpler approach, I'm eager to hear it.


mikhailramalho:
Have a look at clang_c_adjust_expr.cpp and the`do_special_functions` method, the builtins are handled there: there is a function that checks the prefix (`compare_unscore_builtin`) and examples on how to handle them.

I think the easiest solution here is just to rename the identifier

intrigus-lgtm:
> Have a look at clang_c_adjust_expr.cpp and the`do_special_functions` method, the builtins are handled there: there is a function that checks the prefix (`compare_unscore_builtin`) and examples on how to handle them.
> 
> I think the easiest solution here is just to rename the identifier

Just renaming the identifier will not work:
```cpp
    else if (compare_unscore_builtin(identifier, "memcpy"))
    {
      expr.function().name("memcpy");
      expr.function().identifier("c:@F@memcpy");
    }
```

```
Generating GOTO Program
ERROR: Function `c:@F@memcpy' not found
```

It's possible to add a definition to the internal additions:
```
void *memcpy(void *, const void *, __SIZE_TYPE__);
```
but this is also not a good solution, because code like this will then not generate a warning (`warning: implicitly declaring library function 'memcpy' with type 'void *(void *, const void *, unsigned long)' [-Wimplicit-function-declaration]`) anymore:
```c
#include <assert.h>
int main() {
char src[1] = {1};
char dest[1] = {22};
memcpy(dest, src, sizeof(src));
assert(dest[0] == 1);
}
```

mikhailramalho:
> > Have a look at clang_c_adjust_expr.cpp and the`do_special_functions` method, the builtins are handled there: there is a function that checks the prefix (`compare_unscore_builtin`) and examples on how to handle them.
> > I think the easiest solution here is just to rename the identifier
> 
> Just renaming the identifier will not work:
> 
> ```c++
>     else if (compare_unscore_builtin(identifier, "memcpy"))
>     {
>       expr.function().name("memcpy");
>       expr.function().identifier("c:@F@memcpy");
>     }
> ```
> 
> ```
> Generating GOTO Program
> ERROR: Function `c:@F@memcpy' not found
> ```
> 

Yeah, that's because you only update the function call (e.g., the `__builtin_memmove(buffer + 2, buffer, 5)` in `main()`) , not the symbol in the symbol table.

You need to `symbolt *s = context.find_symbol(identifier);` and update the name and id (maybe module too?)

> It's possible to add a definition to the internal additions:
> 
> ```
> void *memcpy(void *, const void *, __SIZE_TYPE__);
> ```
> 
> but this is also not a good solution, because code like this will then not generate a warning (`warning: implicitly declaring library function 'memcpy' with type 'void *(void *, const void *, unsigned long)' [-Wimplicit-function-declaration]`) anymore:
> 
> ```c
> #include <assert.h>
> int main() {
> char src[1] = {1};
> char dest[1] = {22};
> memcpy(dest, src, sizeof(src));
> assert(dest[0] == 1);
> }
> ```

I agree, we should warning about these things like compilers do.

lucasccordeiro:
@intrigus-lgtm: can I ask you for the status of this PR?

intrigus-lgtm:
> @intrigus-lgtm: can I ask you for the status of this PR?

I think there is a better solution than the PR in its current state.
When adding the cprover additions, we're parsing/adding libc and should pull the body from e.g. memcpy and use that for __builtin_memcpy I think.

lucasccordeiro:
> I think there is a better solution than the PR in its current state.

@intrigus-lgtm: Can I ask you whether you could submit this "best" solution?

intrigus-lgtm:
> > I think there is a better solution than the PR in its current state.
> 
> @intrigus-lgtm: Can I ask you whether you could submit this "best" solution?

Sorry, I can't invest time into this right now as I have more important priorities and the current fix is sufficient for me.
I definitely plan to work on this after my thesis deadline (start of January) so please bear with me.

lucasccordeiro:
> I definitely plan to work on this after my thesis deadline (start of January) so please bear with me.

@intrigus-lgtm: We have open positions in our group. If you want to continue working on ESBMC after your MSc dissertation, you could consider applying to our research positions in Manchester ;-)

lucasccordeiro:
@intrigus-lgtm: I hope you have successfully submitted your thesis ;-) Can I ask you for the status of this PR?

intrigus-lgtm:
> @intrigus-lgtm: I hope you have successfully submitted your thesis ;-) Can I ask you for the status of this PR?

There were unfortunately some unexpected delays, so my new submission date is end of January.
I still plan to fix this PR then.

intrigus-lgtm:
@mikhailramalho can I get your comments about my new approach?

Effectively, we rewrite some references to builtin functions with the implementation in the libc/libm.
For example, all references to `__builtin_memcpy` are transparently replaced with `memcpy`.

Because this is done early, we don't have to care about missing symbols later.

(I can investigate whether we should move any of this to the adjuster, but then we'd have to adjust more exprs.
E.g. we'd have to adjust references to functions, the functions itself and potentially all parameters of such functions)

lucasccordeiro:
@intrigus-lgtm: Your PR solves many known bugs in our C++ regression. Could you please update the test description file?

lucasccordeiro (approved):

intrigus-lgtm:
> @intrigus-lgtm: Your PR solves many known bugs in our C++ regression. Could you please update the test description file?

I would not call those bugs solved.
The TCs rely on `--no-unwinding-assertions` so some of the TCs that should fail, do not.
This is because they are never unwound often enough to hit the failing assertion and they also hit no unwinding assertions as they are disabled.
Created https://github.com/esbmc/esbmc/issues/2282 for this.

lucasccordeiro:
Thanks, @intrigus.

These test cases were written over a decade ago (see https://ieeexplore.ieee.org/document/6601583/); the students who wrote these tests already left the group.

Can you please remove the --no-unwinding-assertions and increase the --unwind bound for these failing test cases?

intrigus-lgtm:
> @intrigus-lgtm : Can you please remove the `--no-unwinding-assertions` and increase the `--unwind` bound?

There are hundreds of such test cases outside list and multimap.
That's why I created an issue because we'd likely have to improve memcpy support or else we need huge unwindings everywhere.
And then we need some kind of triage script that would figure out the correct unwinding number automatically.
But I'd prefer not in this PR.

lucasccordeiro:
No problem, @intrigus-lgtm. Indeed, we must improve the memcpy implementation to make it more efficient.

Can you please check the failing test case in the CI?

rafaelsamenezes (approved):

lucasccordeiro:

PR #2134, opened 2024-11-12
more typos and anglicisms

rurban:
this time I went through all comments with hunspell. also fixed some british anglicisms, where we had the same us-english counterpart, for consistency.

behaviour -> behavior
honour -> honor
dependancy -> dependency

there was even a wrong test filename constract.sol


lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @rurban.

PR #2135, opened 2024-11-12
llvm-19 works fine, bump MAX_SUPPORTED_LLVM_VERSION_MAJOR

rurban:
works fine on fedora core 41. 
llvm-18 on ubuntu 24.04 fail's, but ubuntu fails all the time for their own CXX ABI incompetencies.

e.g.
with clang++-18 and the gcc-13 libc++.

/usr/lib/llvm-18/include/cxxabi.h:51:1: error: functions that differ only in their return type cannot be overloaded
   50 | extern _LIBCXXABI_FUNC_VIS __cxa_exception*
      |                            ~~~~~~~~~~~~~~~~
   51 | __cxa_init_primary_exception(void* object, std::type_info* tinfo, void(_LIBCXXABI_DTOR_FUNC* dest)(void*)) throw();
      | ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/cxxabi_init_exception.h:70:7: note: previous declaration is here
   69 |       __cxa_refcounted_exception*

needs a seperate ticket. the workaround on ubuntu is to change /usr/lib/llvm-18/include/cxxabi.h:51 to __cxa_refcounted_exception

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ set (ESBMC_VERSION "${ESBMC_VERSION_MAJOR}.${ESBMC_VERSION_MINOR}.${ESBMC_VERSIO
 # The only default solver available is smtlib
 set (ESBMC_AVAILABLE_SOLVERS "smtlib")
 
-set(MAX_SUPPORTED_LLVM_VERSION_MAJOR 16)
+set(MAX_SUPPORTED_LLVM_VERSION_MAJOR 19)
 
 #################################
 # Configuration                 #


lucasccordeiro (changes requested):
I think @mikhailramalho  has another suggestion here. We should bump the minimum supported version instead of bumping the maximum supported version. Is that correct, @mikhailramalho?

mikhailramalho:
Yeah, can you swap the logic?

rurban:
Sure. What's the minimal version?

rafaelsamenezes:
> Sure. What's the minimal version?

We already do a check for >= 11.

Guess @mikhailramalho just wants the upper restriction to be removed (which I am in favour).

mikhailramalho:
How about the min to be 18? It's the highest version available on the latest ubuntu LTS

rafaelsamenezes:
> How about the min to be 18? It's the highest version available on the latest ubuntu LTS

Windows build still needs to be updated beyond 11. I don't have any preferences otherwise.

PR #2138, opened 2024-11-12
[C++ verification] support lambda capture

### LINKED ISSUE #2136: [C++ Verification] ESBMC does not support lambdas with captures for now
````CPP
#include <iostream>

int main() {
    int x = 42;

    // Lambda function that captures 'x' by value
    auto lambda_with_capture = [x]() {
        std::cout << "Captured value: " << x << std::endl;
    };

    // Call the lambda
    lambda_with_capture();

    return 0;
}
````

````
$ esbmc lambda.cpp
[ESBMC version 7.7.0 64-bit x86_64 linux
Parsing lambda.cpp
Converting
ERROR: ESBMC does not support lambdas with captures for now
ERROR: CONVERSION ERROR
````

### PR DESCRIPTION
**Pull request by XLiZHI**
```c
main (c:@F@main#):
        // 20 file main5.cpp line 5 column 5 function main
        DECL signed int x;
        // 21 file main5.cpp line 5 column 5 function main
        ASSIGN x=42;
        // 22 file main5.cpp line 6 column 5 function main
        DECL signed int y;
        // 23 file main5.cpp line 6 column 5 function main
        ASSIGN y=1;
        // 24 file main5.cpp line 8 column 5 function main
        DECL class __anon_class_at_main5_cpp_main_8_34 lambda_with_capture_1;
        // 25 file main5.cpp line 8 column 5 function main
        ASSIGN lambda_with_capture_1={ .__anon_field_0=x, .__anon_field_1=y };
        // 32 file main5.cpp line 35 column 26 function main
        FUNCTION_CALL:  operator()(&lambda_with_capture_1)
```


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -989,20 +989,33 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
     get_struct_union_class(*lambda_expr.getLambdaClass());
 
     // Construct a new object of the lambda class
     typet lambda_class_type;
     if (get_type(
           *lambda_expr.getLambdaClass()->getTypeForDecl(), lambda_class_type))
       return true;
-    new_expr = gen_zero(pointer_typet(lambda_class_type));
+
+    if (lambda_expr.capture_size())
```
intrigus-lgtm: Can you check whether we can remove the if-else and just always create the struct?

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -989,20 +989,33 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+        exprt init;
+        if (get_expr(**it, init))
+          return true;
+
+        sym.move_to_operands(init);
+      }
+      new_expr = sym;
+    }
+    else
+      new_expr = gen_zero(pointer_typet(lambda_class_type));
```
intrigus-lgtm: I think this and the comment below can be removed.

intrigus-lgtm (approved):
If you have time, could you also implement the C++14 feature "capture initialization":
https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture
E.g.
```cpp
auto func1 = [](int i = 6) { return i + 4; };
```

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -989,20 +989,33 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
           *lambda_expr.getLambdaClass()->getTypeForDecl(), lambda_class_type))
       return true;
-    new_expr = gen_zero(pointer_typet(lambda_class_type));
+
+    if (lambda_expr.capture_size())
+    {
+      const struct_union_typet &this_type =
+        to_struct_union_type(ns.follow(lambda_class_type));
+      exprt sym("struct", this_type);
+      for (auto it = lambda_expr.capture_init_begin();
```
intrigus-lgtm: Can you not use `capture_init` here like so?
```cpp
for (const auto &init : lambda_expr.capture_inits())
      {
        exprt init_expr;
        if (get_expr(*init, init_expr))
          return true;

        sym.move_to_operands(init_expr);
      }
```
 https://clang.llvm.org/doxygen/classclang_1_1LambdaExpr.html#acbad4b6db44728efe65914d34d36a91c here instead?

XLiZHI:
@intrigus-lgtm, I think there is a problem with the existing lambda TCs, I don't think we support `std::function`, can you check your previous PR?

mikhailramalho:
Nice, what happened with the new failing cases?

Also, can you take a look if we can reuse this code for c blocks? https://en.wikipedia.org/wiki/Blocks_(C_language_extension) 

It's like lambdas but for C.

regression/CMakeLists.txt
```diff
@@ -124,6 +124,7 @@ set(REGRESSIONS_CPP11 esbmc-cpp11/cpp
                       esbmc-cpp11/template
                       esbmc-cpp11/tuple
                       esbmc-cpp11/array
+                      esbmc-cpp11/lambda
```
XLiZHI, line 127: > Nice, what happened with the new failing cases?

They failed in the first place.CI didn't check them because it didn't put CMakeLists in.

intrigus-lgtm: Ugh that's embarrassing :rofl: 
I'm looking into this right now, but feel free to merge this PR anyway.

lucasccordeiro:

PR #2139, opened 2024-11-12
Update README.md for Mac, for full compatibility

sshivaji:


README.md
```diff
@@ -37,6 +37,57 @@ cmake .. -DENABLE_Z3=1 -DC2GOTO_SYSROOT=/Library/Developer/CommandLineTools/SDKs
 make -j8
 ````
 
+However, ESBMC is not as stable as the Linux version. Hence, we recommend using AMD64 via docker for optimal performance, until full compatibility is achieved. Sample docker compose and dockerfiles follow:
```
intrigus-lgtm, line 40: What does it mean that ESMBC is not as stable as the Linux version? Are there any open bugs for those stability issues?

sshivaji: Good question. I have many issues I can report, but I did not do so yet, because for example even Float16 does not work to my knowledge

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @sshivaji.

PR #2143, opened 2024-11-14
Update esbmc-wrapper for testcomp2025

ChenfengWei0:


scripts/competitions/testcomp/esbmc-wrapper-cov.py
```diff
@@ -190,7 +190,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, coverage)
 
   # Special case for termination, it runs regardless of the strategy
   if prop == Property.coverage:
-    command_line += "--goto-unwind --base-k-step 5 --no-bounds-check --no-pointer-check --quiet "
+    command_line += "--goto-unwind --base-k-step 5 --no-bounds-check --no-pointer-check --quiet --no-standard-checks"
```
lucasccordeiro: Shall we also use `--compact-trace` to get the value of the variables?

lucasccordeiro (changes requested):

scripts/competitions/testcomp/esbmc-wrapper-cov.py
ChenfengWei0: I guess it's unnecessary since we don't need to generate test cases for pointers, struct... Instead, we only generate cases for certain assignments like `int num = __VERIFIER_nondet_int();`

ChenfengWei0: Btw, may I ask what does the `--compact-trace` actually do? I found out that the description for this option is empty
```
{"compact-trace", NULL, ""}
```

ChenfengWei0: Just did some experiments and it seems the results with and without `--compact-trace` are identical

lucasccordeiro:
Thanks for checking, @ChenfengWei0.

The compact trace enables `--no-slice`, but it does not print the ESBMC's internal symbol.

We should add a description!

lucasccordeiro (approved):

lucasccordeiro:

PR #2159, opened 2024-11-25
[C++ verification] Avoid redundant type operation for lambda

### LINKED ISSUE #2155: [clang-cpp-frontend] Assertion `type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.
```
class classA
{
public:
    template <typename T>
    static bool bar(T var)
    {
        return true;
    }
};

class classB 
{
public:
    classB(){}

    bool foo()
    {
        return classA::bar([this]()
        {
            return true;
        });
    }
};

int main() 
{
    classB obj;
    obj.foo();
    return 0;
}
```

esbmc: /home/lxzy/ESBMC_Project/esbmc/src/util/std_types.h:217: const struct_union_typet& to_struct_union_type(const typet&): Assertion `type.id() == typet::t_struct || type.id() == typet::t_union || type.id() == typet::t_class' failed.
Aborted

XLiZHI:
In this case, there is an incomplete struct, so we can just use symbolic type.


lucasccordeiro (approved):
@XLiZHI: could you please add a test case with `VERIFICATION FAILED`?

lucasccordeiro (approved):

intrigus-lgtm (approved):

lucasccordeiro:

PR #2160, opened 2024-11-26
[Incremental SMT] Verify if --smt-during-symex is enabled before performing incremental assertion checks

### LINKED ISSUE #1367: Issue with --smt-symex-assert option
Consider the following simple program:

```
extern void abort(void);
void reach_error() { __assert_fail("0", "vnew2.c", 3, "reach_error"); }
extern void abort(void);
void assume_abort_if_not(int cond) {
  if(!cond) {abort();}
}
void __VERIFIER_assert(int cond) {
  if (!(cond)) {
  }
  return;
}
int SIZE = 20000001;
int __VERIFIER_nondet_int();
int main() {
   int n = 0;
   int j = 1;
   int k = __VERIFIER_nondet_int();
  __ESBMC_assume(k == j * -1);
  __VERIFIER_assert(k == (2 - 1) * j * -1);
}
```

It is obvious that the assertion on the final line holds.

However, trying to prove it during symbolic execution incorrectly fails as the addition of a print statement here 

https://github.com/esbmc/esbmc/blob/ca1285d661954aa8253de68aea009e401cecc38e/src/goto-symex/symex_target_equation.cpp#L539

will show. The issue, is that the state guard is not being properly added to the assertion, making the assertion unprovable.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1367.

### DIFF
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -345,7 +345,7 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
     options.set_option("no-slice", true);
   }
 
-  if (cmdline.isset("smt-thread-guard") || cmdline.isset("smt-symex-guard"))
+  if (cmdline.isset("smt-thread-guard") || cmdline.isset("smt-symex-guard") || cmdline.isset("smt-symex-assert"))
   {
     if (!cmdline.isset("smt-during-symex"))
     {
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -349,13 +349,8 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
     cmdline.isset("smt-thread-guard") || cmdline.isset("smt-symex-guard") ||
     cmdline.isset("smt-symex-assert"))
   {
-    if (!cmdline.isset("smt-during-symex"))
-    {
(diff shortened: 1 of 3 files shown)


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -345,7 +345,9 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
     options.set_option("no-slice", true);
   }
 
-  if (cmdline.isset("smt-thread-guard") || cmdline.isset("smt-symex-guard"))
+  if (
+    cmdline.isset("smt-thread-guard") || cmdline.isset("smt-symex-guard") ||
+    cmdline.isset("smt-symex-assert"))
```
XLiZHI, line 350: Can we enable `smt-during-symex` by default when we detect the use of these flags?

lucasccordeiro: Yes, this might be a better solution!

Anthonysdu: I agree, it's would be easier to use the incremental flags by enable it:
`options.set_option("smt-during-symex", true);`

lucasccordeiro: Implemented as suggested.

XLiZHI (approved):

Anthonysdu (approved):

PR #2162, opened 2024-11-26
Intrigus/fix nan simplify

intrigus-lgtm:
NaN != NaN is always true, but this defeats the purpose of the check
The assumption is that `src != source_value` means that we were able to simplify
something and return a simplified expression.
In case of NaN, we would continuously return "simplified" expressions, because
NaN != NaN would always be true.

There might be other cases elsewhere, but I only encountered the problem in this specific simplification and honestly don't have time to look at those potential other cases.

### DIFF
--- a/regression/esbmc/simplify-float-nan/test.desc
+++ b/regression/esbmc/simplify-float-nan/test.desc
@@ -1,0 +1,4 @@
+KNOWNBUG
+main.cpp
+--overflow-check
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/simplify-float-nan/test.desc
+++ b/regression/esbmc/simplify-float-nan/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --overflow-check
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/simplify-float-nan/main.cpp
+++ b/regression/esbmc/simplify-float-nan/main.cpp
@@ -1,0 +1,8 @@
+extern "C" void memcpy(void *, void *, int);
+float a = __builtin_nanf(0);
+double b = a;
+double other = 1;
+int main()
+{
+  memcpy(&other, &b, 8);
+}
--- a/src/util/simplify_expr2.cpp
+++ b/src/util/simplify_expr2.cpp
@@ -3071,7 +3071,24 @@ expr2tc byte_extract2t::do_simplify() const
     }
   }
 
-  if (src != source_value || off != source_offset)
+  bool src_unequal = src != source_value;
+  // NaN != NaN is always true, but this defeats the purpose of the check
+  // (the assumption is that `src != source_value` means that we were able to simplify
+  // something)
+  if (is_constant_floatbv2t(src) && is_constant_floatbv2t(source_value))
+  {
+    ieee_floatt src_val = to_constant_floatbv2t(src).value;
+    ieee_floatt source_val = to_constant_floatbv2t(source_value).value;
+    if (
+      src_val.is_NaN() && source_val.is_NaN() &&
+      src_val.get_exponent() == source_val.get_exponent() &&
+      src_val.get_fraction() == source_val.get_fraction())
+    {
+      src_unequal = false;
+    }
+  }
+
+  if (src_unequal || off != source_offset)
     return byte_extract2tc(type, src, off, big_endian);
 
   return {};


lucasccordeiro (approved):
LGTM.

regression/esbmc/simplify-float-nan/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.cpp
+--overflow-check
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: BTW, shall we classify this test case as `VERIFICATION FAILED`?

This is the Valgrind output:

````
==24150== Memcheck, a memory error detector
==24150== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==24150== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==24150== Command: ./int
==24150==
==24150== Invalid read of size 1
==24150==    at 0x49A06C0: __strtof_nan (strtod_nan_main.c:37)
==24150==    by 0x1091C0: __static_initialization_and_destruction_0(int, int) (in /home/lucas/examples/int)
==24150==    by 0x1091F8: _GLOBAL__sub_I_a (in /home/lucas/examples/int)
==24150==    by 0x4979EBA: call_init (libc-start.c:145)
==24150==    by 0x4979EBA: __libc_start_main@@GLIBC_2.34 (libc-start.c:379)
==24150==    by 0x1090A4: (below main) (in /home/lucas/examples/int)
==24150==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==24150==
==24150==
==24150== Process terminating with default action of signal 11 (SIGSEGV)
==24150==  Access not within mapped region at address 0x0
==24150==    at 0x49A06C0: __strtof_nan (strtod_nan_main.c:37)
==24150==    by 0x1091C0: __static_initialization_and_destruction_0(int, int) (in /home/lucas/examples/int)
==24150==    by 0x1091F8: _GLOBAL__sub_I_a (in /home/lucas/examples/int)
==24150==    by 0x4979EBA: call_init (libc-start.c:145)
==24150==    by 0x4979EBA: __libc_start_main@@GLIBC_2.34 (libc-start.c:379)
==24150==    by 0x1090A4: (below main) (in /home/lucas/examples/int)
==24150==  If you believe this happened as a result of a stack
==24150==  overflow in your program's main thread (unlikely but
==24150==  possible), you can try to increase the size of the
==24150==  main thread stack using the --main-stacksize= flag.
==24150==  The main thread stack size used in this run was 8388608.
==24150==
==24150== HEAP SUMMARY:
==24150==     in use at exit: 0 bytes in 0 blocks
==24150==   total heap usage: 0 allocs, 0 frees, 0 bytes allocated
==24150==
==24150== All heap blocks were freed -- no leaks are possible
==24150==
==24150== For lists of detected and suppressed errors, rerun with: -s
==24150== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
Segmentation fault
````

**Review by lucasccordeiro — COMMENTED** [review]
@intrigus-lgtm: could you please check this test.desc?

File: regression/esbmc/simplify-float-nan/test.desc (hunk 1 above)
Comment (XLiZHI) [reply regression/esbmc/simplify-float-nan/test.desc:4]: Valgrind seems to have crashed here. This example doesn't seem to have an overflow, but it can lead to undefined behavior?

**Review by XLiZHI — APPROVED** [review]

**Comment by intrigus-lgtm** [comment]
Before merging, we could also fix this without (potentially) changing every simplification:
we'd have to change the `==` operator of `ieee_floatt` to just check whether the "bits" are equal instead of what IEEE-754 mandates when checking for equality.
Note that originally this was the case for the class and then it has been "fixed":
https://github.com/esbmc/esbmc/commit/894e12a0a6fb440f29b2a92cbfeb4fc9288387b0

**Review by mikhailramalho — CHANGES_REQUESTED** [review]
This PR is incorrect, any comparison operation with NaNs will always return false.

NaN != NaN has a well-defined behavior and it should always return false. That's what my commit fixed back then.

**Comment by intrigus-lgtm** [comment]
> This PR is incorrect, any comparison operation with NaNs will always return false.
> 
> NaN != NaN has a well-defined behavior and it should always return false. That's what my commit fixed back then.

Please tell me how to fix this issue then.

**Comment by intrigus-lgtm** [comment]
Without this fix, we get infinte recursion and die with a segfault in the end due to a stack overflow.

**Comment by intrigus-lgtm** [comment]
Without this fix, we get infinte recursion and die with a segfault in the end due to a stack overflow.

**Comment by mikhailramalho** [comment]
What's causing the infinite recursion?

**Comment by mikhailramalho** [comment]
You can also open an issue with a detailed description of the problem,
hopefully including a test case, and ask for suggestions or propose a plan

**Comment by intrigus-lgtm** [comment]
> What's causing the infinite recursion?
As I said in the PR description:
when we're simplifying, `source_value` is the thing we want to simplify and `src` is the potentially simplified result.
We then proceed to check whether `source_value != src`. If it is unequal, we were able to simplify something and return the simplified expression. This means that for any expression that contains `NaN`, we ALWAYS say that we were able to simplify something which is just not true, despite `source_value != src` obviously.

The overall context is this:
We first try to simplify the expr in `#1` which calls `byte_extract2t::do_simplify` (the code modified in this PR) and it returns a **non-nil** value, due to `NaN` as explained before.
We therefore enter the `if` and try to simplify it again in `#2`. This brings us right back to `expr2t::simplify` causing infinite recursion in the end, because we NEVER reach a fix-point (despite the existence of a fix-point!)
```cpp
expr2tc expr2t::simplify() const
{
  try
  {
    // Corner case! Don't even try to simplify address of's operands, might end up
    // taking the address of some /completely/ arbitary pice of data, by
    // simplifiying an index to its data, discarding the symbol.
    if (expr_id == address_of_id) // unlikely
      return expr2tc();

    // And overflows too. We don't wish an add to distribute itself, for example,
    // when we're trying to work out whether or not it's going to overflow.
    if (expr_id == overflow_id)
      return expr2tc();

    // Try initial simplification
    expr2tc res = do_simplify(); // #### 1
    if (!is_nil_expr(res))
    {
      // Woot, we simplified some of this. It may have _additional_ fields that
      // need to get simplified (member2ts in arrays for example), so invoke the
      // simplifier again, to hit those potential subfields.
      expr2tc res2 = res->simplify();  // #### 2

      // If we simplified even further, return res2; otherwise res.
      if (is_nil_expr(res2))
        return res;
      else
        return res2;
    }
```

> You can also open an issue with a detailed description of the problem, hopefully including a test case, and ask for suggestions or propose a plan

The PR contains a test case and I proposed a plan to fix this.
We can either use the fix proposed in this PR, which special-cases this for `byte_extract2t`, but there might be other cases elsewhere.
Or: we can change back the behavior of `operator==` of `ieee_float` as it was originally and I'd say there was a reason that it was implemented this way (it's explicitly mentioned in the docs!).
Or: we can make every simplification method return a pair of <boolean, expr> where the boolean signifies whether we were able to simplify anything, because we cannot rely on the (dis-)equality of `input == simplify(input)`.

mikhailramalho:
I have questions about your PRs because they are unclear. Please be more
specific in the PR's description, as I have very limited time available for
this project. It doesn't matter if you try to point out that you wrote the
explanation before, if they are not clear, I will still not get them.

Em sex., 29 de nov. de 2024 às 15:37, intrigus-lgtm <
***@***.***> escreveu:

> What's causing the infinite recursion?
> As I said in the PR description:
> when we're simplifying, source_value is the thing we want to simplify and
> src is the potentially simplified result.
> We then proceed to check whether source_value != src. If it is unequal,
> we were able to simplify something and return the simplified expression.
> This means that for any expression that contains NaN, we ALWAYS say that
> we were able to simplify something which is just not true, despite source_value
> != src obviously.
>
> The overall context is this:
> We first try to simplify the expr in #1 which calls
> byte_extract2t::do_simplify (the code modified in this PR) and it returns
> a *non-nil* value, due to NaN as explained before.
>

Now we are getting somewhere: do we fail to extract a byte from a NaN? If
so, this sounds like a proper fix.


-- 

Mikhail R. Gadelha, Ph.D.

lucasccordeiro:
@intrigus-lgtm: did you have a chance to read @mikhailramalho's question?

intrigus-lgtm:
> @intrigus-lgtm: did you have a chance to read @mikhailramalho's question?

Yes, but I still don't really think this is a proper solution. We shouldn't have to add this code to every possibe location in the simplifier.
While I found only a visible bug in this case, other cases might exist as well.

lucasccordeiro:
@mikhailramalho: Can I ask you whether you would have any suggestions on how to move forward with this PR?

lucasccordeiro:
There is no crash from Valgrind, @XLiZHI. I agree that its output is not nice :-)

Valgrind reports:

````
==24150== Invalid read of size 1
==24150==    at 0x49A06C0: __strtof_nan (strtod_nan_main.c:37)
==24150==    by 0x1091C0: __static_initialization_and_destruction_0(int, int) (in /home/lucas/examples/int)
==24150==    by 0x1091F8: _GLOBAL__sub_I_a (in /home/lucas/examples/int)
==24150==    by 0x4979EBA: call_init (libc-start.c:145)
==24150==    by 0x4979EBA: __libc_start_main@@GLIBC_2.34 (libc-start.c:379)
==24150==    by 0x1090A4: (below main) (in /home/lucas/examples/int)
==24150==  Address 0x0 is not stack'd, malloc'd or (recently) free'd

ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)

@intrigus-lgtm : Can I ask you for your thoughts here? You marked the regression as successful, but Valgrind finds an error.

intrigus-lgtm:
Still thinking about the best solution for this.
There is no immediate pressure to solve this, so I'd keep this open.

mikhailramalho:
Closing as part of stale-branch cleanup. `byte_extract2t::do_simplify` in master has been substantially rewritten since this PR was authored (current implementation at `src/util/expr_simplifier.cpp:5568`); the specific `src != source_value` short-circuit path the patch targets no longer exists in that form. If the NaN infinite-recursion bug still reproduces against current master, a fresh issue + PR would be more useful than rebasing. The branch will be deleted.

PR #2165, opened 2024-11-28
[C++ verification] adjust lambda operator

XLiZHI:
Here is AST for lambda expr:
```c
| |       `-LambdaExpr 0x6f82f10 <col:28, line:22:9> '(lambda at main6.cpp:19:28)'
| |         |-CXXRecordDecl 0x6f82a78 <line:19:28> col:28 implicit class definition
| |         | |-DefinitionData lambda pass_in_registers standard_layout trivially_copyable literal can_const_default_init
| |         | | |-DefaultConstructor
| |         | | |-CopyConstructor simple trivial has_const_param needs_implicit implicit_has_const_param
| |         | | |-MoveConstructor exists simple trivial needs_implicit
| |         | | |-CopyAssignment trivial has_const_param needs_implicit implicit_has_const_param
| |         | | |-MoveAssignment
| |         | | `-Destructor simple irrelevant trivial
| |         | |-CXXMethodDecl 0x6f82bc0 <col:35, line:22:9> line:19:28 used constexpr operator() 'void () const' inline
| |         | | `-CompoundStmt 0x6f82d18 <line:20:9, line:22:9>
| |         | |   `-UnaryOperator 0x6f82d00 <line:21:13, col:17> 'int' postfix '++'
| |         | |     `-MemberExpr 0x6f82cd0 <col:13> 'int' lvalue ->data 0x6f828a8
| |         | |       `-CXXThisExpr 0x6f82cc0 <col:13> 'classB *' implicit this
| |         | |-FieldDecl 0x6f82e98 <line:19:29> col:29 implicit 'classB *'
| |         | `-CXXDestructorDecl 0x6f82f40 <col:28> col:28 implicit referenced ~(lambda at main6.cpp:19:28) 'void () noexcept' inline default trivial
```

When parsing the operator, the AST declares the original variable (this->data), however we should use a member variable (__this) from the lambda class, so we try to replace the operator expression in this PR.

Before:
```
operator() (c:@S@classB@F@foo#@Sa@F@operator()#1):
        // 30 file main6.cpp line 21 column 13 function operator()
        ASSIGN this->data=this->data + 1;                      
        // 31 file main6.cpp line 22 column 9 function operator()
        END_FUNCTION // operator()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
failed: "data" is not a member of lambda
```

this PR:
```
operator() (c:@S@classB@F@foo#@Sa@F@operator()#1):
[...]

PR #2168, opened 2024-11-29
Revert "[C++ verification] adjust lambda operator"

intrigus-lgtm:
Reverts esbmc/esbmc#2165

Reason: The code doesn't actually work and could be simplified.
Rather than trying to improve the (now) existing code, it would be best to state afresh IMHO.

I'm creating a new PR that adds the reverted TCs and new ones.

For example, this TC is broken (which it was already before your new PR, **but the approach to fix that would also fix the problems that your PR fixed in a cleaner way**):
```
#include <cassert>
 
int main()
{
    int aaa = 22;
    int result = [aaa, bbb = 256]() { return aaa + bbb; }();
    assert(result == 2762);
    return 0;
}
```

The pretty well hidden method you were probably looking for is [getCaptureFields](https://clang.llvm.org/doxygen/classclang_1_1CXXRecordDecl.html#a5f9bae915113a518675d1f2445bed2f8) and can be used like so:
```
llvm::DenseMap<const ValueDecl*, FieldDecl*> captures{};
FieldDecl* thisCapture{};
cxxRecordDecl.getCaptureFields(captures, thisCapture);
```

It gives a mapping from a capture to the associated field in the closure.


XLiZHI:
Good. This looks like a useful API, it's hidden so well that I didn't find it, I'll take a look at the method.(This is why I added the discussion tag to my last PR)

As for the previous code, I'll remove them in the new PR, no need to revert it.

intrigus-lgtm:
> Good. This looks like a useful API, it's hidden so well that I didn't find it, I'll take a look at the method.(This is why I added the discussion tag to my last PR)

Well, I did plan to review your PR, but I'm in the final phase of my Master's thesis right now and ~1 day is currently not enough time to properly review a PR. So I would have preferred a bit more time before a PR is merged if I'm requested for review.
> 
> As for the previous code, I'll remove them in the new PR, no need to revert it.

I'd _really_ prefer to start afresh, you can of course copy paste stuff, but it makes it easier to review for me from the previous state.

As for why the [existing tests](https://github.com/esbmc/esbmc/blob/master/regression/esbmc-cpp11/lambda/lambda_01/main.cpp) "work":

```
operator() (c:test2.cpp@102@F@main#@Sa@F@operator()#1):
        // 38 file /tmp/test2.cpp line 10 column 9 function operator()
        ASSERT x == 42 // assertion x == 42
        // 39 file /tmp/test2.cpp line 11 column 9 function operator()
        ASSERT y == 1 // assertion y == 1
        // 40 file /tmp/test2.cpp line 12 column 5 function operator()
        END_FUNCTION // operator()
```

this is the code generated for
```
#include <cassert>

int main()
{
    int x = 42; // ## 1
    int y = 1;

    auto lambda_with_capture_1 = [=]()
    {
        assert(x == 42); // ## 2
        assert(y == 1);
    };
```

The reason it "works" is that the `x` inside the lambda (`#1`) is the **same* `x` as the one outside (`#1`).
Due to the way esbmc works, this `x` is still visible inside the lambda, while we should use `__x` inside the lambda instead of `x`.
(`class __anon_class_at_/tmp/test2_cpp_main_8_34 lambda_with_capture_1={ .__x=x, .__y=y };`)

XLiZHI:
> the reason it "works" is that the x inside the lambda (#1) is the *same x as the one outside (#1).
> Due to the way esbmc works, this x is still visible inside the lambda, while we should use __x inside the lambda instead of x.
> (class __anon_class_at_/tmp/test2_cpp_main_8_34 lambda_with_capture_1={ .__x=x, .__y=y };)

Indeed, when I was exploring this method, I realized the reason and I think my method is not robust and would like to discuss it further.

I agree with revert if you insist. Do you want to take over the job or should I do it?

intrigus-lgtm:
> > the reason it "works" is that the x inside the lambda (#1) is the *same x as the one outside (#1).
> > Due to the way esbmc works, this x is still visible inside the lambda, while we should use __x inside the lambda instead of x.
> > (class __anon_class_at_/tmp/test2_cpp_main_8_34 lambda_with_capture_1={ .__x=x, .__y=y };)
> 
> Indeed, when I was exploring this method, I realized the reason and I think my method is not robust and would like to discuss it further.
> 
> I agree with revert if you insist. Do you want to take over the job or should I do it?

Revert would still be best I think.
Yeah, I can take over this, if that's fine with you.

XLiZHI (approved):

lucasccordeiro:
@intrigus-lgtm: Can I ask you to submit a PR that properly fixes this lamba operator? You can remove @XLiZHI's solution from your upcoming PR.

intrigus-lgtm:
Superseded by https://github.com/esbmc/esbmc/pull/2171
Sorry for the noise.

PR #2171, opened 2024-12-01
[C++] Fix lambda captures.

intrigus-lgtm:
Adds support for c++ lambda captures.
Commit-by-commit review is helpful.
Most important commits:
highlights existing deficiencies in the capturing:
https://github.com/esbmc/esbmc/commit/14f42182c6870f4bf0b923bfcd29f48615b25995
highlights an unrelated bug to references and (pre|post)(inc|dec)rement:
https://github.com/esbmc/esbmc/commit/11088b183557b81e7e454ddebccddfa780f8cd82
the actual code change:
https://github.com/esbmc/esbmc/commit/d5f890c9f00543e5b1a72285ecaa56a82b8425ca


XLiZHI (commented):
I agree with the changes to lambda expr.

Now I have some new ideas. After we are able to get `llvm::DenseMap<const clang::ValueDecl *, clang::FieldDecl *>`, we can set a global variable (clang-cpp-frontend) of type bool to true (is_lambda).

In `get_function_body`. Replace the decl with the field decl, if the decl is in map (map.find(decl)) and flag (is_lambda) is true.
https://github.com/esbmc/esbmc/blob/b456d8fad46222f55dc159ae9d9b45567173d50e/src/clang-cpp-frontend/clang_cpp_convert.cpp#L37-L39

Look like this:
```
if (is_lambda && map.find(decl))
{
   get_decl(fielddecl, new_expr)
   return false;
}
```

The same approach can be used for the `this` pointer:
https://github.com/esbmc/esbmc/blob/b456d8fad46222f55dc159ae9d9b45567173d50e/src/clang-cpp-frontend/clang_cpp_convert.cpp#L857-L858

What do you think? I think this method of replacing expressions is somewhat complicated and not very robust.

intrigus-lgtm:
> I agree with the changes to lambda expr.
> 
> Now I have some new ideas. After we are able to get `llvm::DenseMap<const clang::ValueDecl *, clang::FieldDecl *>`, we can set a global variable (clang-cpp-frontend) of type bool to true (is_lambda).
> 
> In `get_function_body`. Replace the decl with the field decl, if the decl is in map (map.find(decl)) and flag (is_lambda) is true.
> 
> https://github.com/esbmc/esbmc/blob/b456d8fad46222f55dc159ae9d9b45567173d50e/src/clang-cpp-frontend/clang_cpp_convert.cpp#L37-L39
> 
> Look like this:
> 
> ```
> if (is_lambda && map.find(decl))
> {
>    get_decl(fielddecl, new_expr)
>    return false;
> }
> ```
> 
> The same approach can be used for the `this` pointer:
> 
> https://github.com/esbmc/esbmc/blob/b456d8fad46222f55dc159ae9d9b45567173d50e/src/clang-cpp-frontend/clang_cpp_convert.cpp#L857-L858
> 
> What do you think? I think this method of replacing expressions is somewhat complicated and not very robust.

That's actually something I was also thinking about, but wasn't sure whether that will be much cleaner.
I would hope that the current approach is quite robust (at least I think it is^^), but if you have time, feel free to try your idea and report back.

I would normally love to investigate this myself, but lack the time right now, so if you can help that would be nice.

lucasccordeiro (approved):
LGTM.

lucasccordeiro:
@intrigus-lgtm: could you please check these two test cases failing with your PR?

````
The following tests FAILED:
	2259 - regression/esbmc-cpp14/deduced-return-type/auto-function (Failed) esbmc-cpp14/deduced-return-type regression
	2264 - regression/esbmc-cpp14/deduced-return-type/auto-template-not-instantiated-fail (Failed) esbmc-cpp14/deduced-return-type regression
````

File: src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp [hunk 1]
```diff
@@ -59,55 +59,49 @@ void clang_cpp_adjust::gen_implicit_union_copy_move_constructor(symbolt &symbol)
...
-  const struct_typet::componentst &components =
-    to_struct_type(class_symb).components();
-
-  for (const auto &c : components)
+  if (op.is_symbol())
   {
-    if (c.get_bool("#capture_this"))
+    irept replacement_symbol = lambda_capture_fields.find(op.identifier());
+    if (replacement_symbol.is_not_nil())
     {
```
XLiZHI: Please remove the bracket

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
```diff
@@ -59,55 +59,49 @@ void clang_cpp_adjust::gen_implicit_union_copy_move_constructor(symbolt &symbol)
...
-      exprt dest = member_exprt(deref, c.get_name(), c.type());
-      exprt memderef =
-        dest.type().is_pointer() ? dereference_exprt(dest, dest.type()) : dest;
-
-      replace_capture_this(memderef, body);
+      op.swap(replacement_symbol);
+    }
+  }
+  else
+  {
```
XLiZHI, line 73: same here

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
```diff
@@ -59,55 +59,49 @@ void clang_cpp_adjust::gen_implicit_union_copy_move_constructor(symbolt &symbol)
...
+    Forall_operands (subop, op)
+    {
```
XLiZHI: same

XLiZHI (changes requested):

intrigus-lgtm:
> @intrigus-lgtm: could you please check these two test cases failing with your PR?
> 
> ```
> The following tests FAILED:
> 	2259 - regression/esbmc-cpp14/deduced-return-type/auto-function (Failed) esbmc-cpp14/deduced-return-type regression
> 	2264 - regression/esbmc-cpp14/deduced-return-type/auto-template-not-instantiated-fail (Failed) esbmc-cpp14/deduced-return-type regression
> ```

Fixed, I accidentally moved some files related to the tests.

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
intrigus-lgtm: Done.

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
intrigus-lgtm: Done.

src/clang-cpp-frontend/clang_cpp_adjust_expr.cpp
intrigus-lgtm: I think it would look odd to have braces for the `if`, but not for the `else`.

XLiZHI (approved):

lucasccordeiro:

PR #2173, opened 2024-12-03
[C++ OM] removed <utility>

lucasccordeiro:


intrigus-lgtm (changes requested):
Why are we removing this?

lucasccordeiro:
Hi @intrigus-lgtm,

When verifying C++ programs using ESBMC, we aim to maximize the use of the system's native libraries and minimize reliance on operational models for C++ libraries.

lucasccordeiro:
@intrigus-lgtm and @XLiZHI: We have enabled 40 test cases by using the utility system's library in this PR. We should also check how mature our C++ frontend is before using other systems' libraries.

XLiZHI:
> @intrigus-lgtm and @XLiZHI: We have enabled 40 test cases by using the utility system's library in this PR. We should also check how mature our C++ frontend is before using other systems' libraries.

That‘s nice . When I was working on the utility OM, there were a lot of conversion errors using the standard library directly.

lucasccordeiro:
> That‘s nice . When I was working on the utility OM, there were a lot of conversion errors using the standard library directly.

We are evolving with our C++ frontend ;-)

intrigus-lgtm:
> Hi @intrigus-lgtm,
> 
> When verifying C++ programs using ESBMC, we aim to maximize the use of the system's native libraries and minimize reliance on operational models for C++ libraries.

In my experience, mixing c++ system libraries (and c++ oms) and the esbmc c library (can) leads to problems, because the program under test can pull in c++ system libraries that then pull in c system headers which later conflict with the esbmc c headers.

It might be as easy as tweaking the include path and guaranteeing to remove the c system headers, but I couldn't find a command switch (in clang) that does that.

Also one problem that we definitely have to consider is that we also have to pull in the implementation of the c++ system library as most of it might be implemented in the header, but there are definitely bits that are not and cause missing functions for esbmc.

For example this `__throw_bad_function_call` is defined in a header while the actual implementation is here:
https://github.com/gcc-mirror/gcc/blob/4114b7fb1cb4cb90b9fafc22213d7d9579bc19e0/libstdc%2B%2B-v3/src/c%2B%2B11/functional.cc#L33
(`__throw_bad_function_call` is used in `std::function`)

XLiZHI (approved):

intrigus-lgtm (approved):

PR #2176, opened 2024-12-03
Keep llvm-13 check only for ARM targets

rafaelsamenezes:
There is currently no need for llvm-13 in x86. We still test it only for CHERI, therefore only ARM64 is needed.


lucasccordeiro (approved):

XLiZHI (approved):

lucasccordeiro:

PR #2177, opened 2024-12-06
Keep sv-benchmarks up to date

Anthonysdu:
We are currently running the sv-benchmarks main branch, but the repos in different machines are behind with origin/master,  which results in different benchmark set:

https://github.com/esbmc/esbmc/actions/runs/12088163144
https://github.com/esbmc/esbmc/actions/runs/12088171663

This PR keeps the main branch up-to-date with origin/master (if we are benchmarking main branch).

PR #2178, opened 2024-12-06
[Concurrency] New thread schedule policy

Anthonysdu:
When context switch happens, current thread schedule policy is trying to start with the lowest thread id (tid = 0) and then increase the thread id, but this is not very efficient for bug detection, because the newly created thread will be the last to switch to, and the interleaving could not reach deeper child threads quickly.

Here introduces a new policy: for a list of threads: [0,1,2, 3, active_thread, 4, 5, n-1], when a context switch occurs at active_thread, it first traverse to the right until n-1, if no threads can be scheduled, then reverse from active_thread to explore the left until reach 0.

 This PR with` --context bound 3 --no-por`
```
  correct:             425
    correct true:      131
    correct false:     294
  incorrect:            10
    incorrect true:      6
    incorrect false:     4
  unknown:             290
```

Master with `--context-bound 2 --no-por`
```
  correct:             426
    correct true:      137
    correct false:     289
  incorrect:            11
    incorrect true:      8
    incorrect false:     3
  unknown:             288
  Score:               259 (max: 1099)
```
There are 2 more incorrect false from new benchmark set `libvsync` (Timeout to incorrect false), should be a guard bug, I will need to take a look today.

BTW, symex-assume didn't improve the result, the previous runs are based on different benchmark set :) (See #2177 )
This PR with `--default-solver bitwuzla --smt-symex-assume --context-bound 3 --no-por`
```
  correct:             409
    correct true:      119
    correct false:     290
  incorrect:            10
    incorrect true:      6
    incorrect false:     4
  unknown:             306
  Score:               272 (max: 1099)
```
It runs faster in a subset of benchmarks though, should take a look this week.

### DIFF
--- a/regression/esbmc-unix2/github_402_fail3/test.desc
+++ b/regression/esbmc-unix2/github_402_fail3/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --no-por --no-goto-merge
-^ERROR: There are goto statements that shouldn't be merged at this point$
+^VERIFICATION FAILED$
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -256,7 +256,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
                  check_if_benchmark_contains_pthread(benchmark))
 
   if concurrency:
-    command_line += " --no-por --context-bound 2 "
+    command_line += " --no-por --context-bound 3 --default-solver bitwuzla --smt-symex-assume "
     #command_line += "--no-slice " # TODO: Witness validation is only working without slicing
 
   # Add witness arg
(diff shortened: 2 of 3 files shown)


Anthonysdu:
We are using --context-bound 3 --no-por in svcomp:

https://sv-comp.sosy-lab.org/2025/results/results-verified/esbmc-kind.2024-12-02_10-54-34.results.SV-COMP25_unreach-call.ConcurrencySafety-Main.xml.bz2.fixed.xml.bz2.table.html#/

lucasccordeiro (approved):

lucasccordeiro:

lucasccordeiro:
BTW, before this PR, our regression test suite used to take:

````
100% tests passed, 0 tests failed out of 377

Label Time Summary:
esbmc-unix     = 947.59 sec*proc (209 tests)
esbmc-unix2    = 879.90 sec*proc (168 tests)
regression     = 1827.49 sec*proc (377 tests)

Total Test time (real) = 467.43 sec

real    7m47.602s
user    30m2.527s
sys     1m32.285s
````

After this PR, our regression test suite takes:

````
100% tests passed, 0 tests failed out of 377

Label Time Summary:
esbmc-unix     = 466.82 sec*proc (209 tests)
esbmc-unix2    = 774.42 sec*proc (168 tests)
regression     = 1241.24 sec*proc (377 tests)

Total Test time (real) = 311.01 sec

real    5m11.274s
user    20m3.453s
sys     1m11.682s
````

Good job done, @Anthonysdu!

XLiZHI (commented):
LGTM！

PR #2179, opened 2024-12-06
[C++ verification] add ctor OMs for tuple and map

XLiZHI:
Known issue: #2180

### DIFF
--- a/src/util/c_typecast.cpp
+++ b/src/util/c_typecast.cpp
@@ -617,7 +617,8 @@ void c_typecastt::implicit_typecast_followed(
 
     if (
       ns.follow(dest_type.subtype()).id() == src_type.id() &&
-      src_type != dest_type && !expr.is_address_of()) // TODO: remove this condition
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

src/util/c_typecast.cpp
```diff
@@ -617,7 +617,8 @@ void c_typecastt::implicit_typecast_followed(
 
     if (
       ns.follow(dest_type.subtype()).id() == src_type.id() &&
-      src_type != dest_type) // TODO: remove this condition
+      src_type != dest_type &&
```
intrigus-lgtm, line 620: Nit: this TODO is now 6 months old

XLiZHI: https://github.com/esbmc/esbmc/pull/1931#issuecomment-2229400786

We discussed how to handle the reference type before, but this can be a big change. This code will be refactored.

lucasccordeiro:

PR #2185, opened 2024-12-08
[Concurrency] Guard and Interleavings fix

Anthonysdu:
This PR https://github.com/esbmc/esbmc/actions/runs/12218538927
The result score is the same as master: 
`--context-bound 3 --no-por`

```
  correct:             425
    correct true:      131
    correct false:     294
  incorrect:            10
    incorrect true:      6
    incorrect false:     4
  unknown:             290
```
But verification time is improved slightly:

image

Master:
Correct true: 6650s
Correct false: 4890s

This PR:
Correct true: 6610s
Correct false: 4340s

### DIFF
--- a/regression/esbmc-unix/SV_COMP_03/test.desc
+++ b/regression/esbmc-unix/SV_COMP_03/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --context-bound 2 --no-assertions --data-races-check -Wno-error=implicit-function-declaration
 ^VERIFICATION FAILED$
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -202,8 +202,9 @@ void execution_statet::symex_step(reachability_treet &art)
 
   merge_gotos();
 
-  // If current state guard is false, it shouldn't perform further context switch. 
-  if (!state.guard.is_false() || !is_cur_state_guard_false(state.guard.as_expr()))
+  // If current state guard is false, it shouldn't perform further context switch.
+  if (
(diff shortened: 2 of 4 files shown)


regression/esbmc-unix/SV_COMP_03/test.desc
Anthonysdu, line 1: This is a bug from partial order reduction,
will fix this in mpor pr.

regression/esbmc-unix/SV_COMP_03/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
Anthonysdu, line 1: This is a bug from partial order reduction,
will fix this in mpor pr.

src/goto-symex/execution_state.cpp
```diff
@@ -715,15 +732,6 @@ void execution_statet::execute_guard()
...
       guardt().as_expr(), parent_guard, get_active_state().source, first_loop);
-  // Check to see whether or not the state guard is false, indicating we've
-  // found an unviable interleaving. However don't do this if we didn't
-  // /actually/ switch between threads, because it's acceptable to have a
-  // context switch point in a branch where the guard is false (it just isn't
-  // acceptable to permit switching).
-  if (
-    last_active_thread != active_thread &&
-    is_cur_state_guard_false(threads_state[active_thread].guard.as_expr()))
-    interleaving_unviable = true;
```
Anthonysdu, line 726: It is too early to set interleaving_unviable because the active_thread guard will be updated later in merge_gotos(), Instead set the value after merge_gotos()

lucasccordeiro (approved):

lucasccordeiro:

PR #2186, opened 2024-12-09
[Concurrency] Remove Global Guard

### LINKED ISSUE #1991: [goto-guard] conflict states interleave
The test cases come from https://github.com/esbmc/esbmc/issues/397, this issue was fixed by using a global guard to prevent assertions from being executed https://github.com/esbmc/esbmc/pull/399.  

But this fix comes with another problem: assertions are not executed in the current state either:
```c
#include <pthread.h>
#include <assert.h>
pthread_t c;
int d;
void *e(void *) { d = 6; }
int main() {
  if (nondet_bool()) {
    pthread_create(&c, NULL, e, NULL);
    return 0;
  }
  d = 3;
  assert(d == 0);  // should be fail
  return 0;
}
```
However, in the current master: `VERIFICATION SUCCESSFUL`

I looked into how the symbolic execution part works, and I think of the above code as three states: 
|        | State 0                                      | State 1                        | State 2            |
|--------|:------------------------------------------:|:-----------------------------:|:-----------------:|
| Thread | Thread 0                                     | Thread 0                       | Thread 1           |
| Code   | pthread_create(&c, NULL, e, NULL);<br>return 0; | d = 3;<br>assert(d == 3);<br>return 0; | d = 6;<br>return 0; |

In this table **state 1** and **state 2**, as well as **state 0** and **state 1** are conflicting because there is `return 0;`

But when I look at the symex trace, there is thread interleaving between **state 1** and **state 2**. 

When code execution reaches return in **state 0**, guard will become false, However, after `merge_goto()`, all state guards in thread 0 become true.

https://github.com/esbmc/esbmc/blob/7e9c1065ee2ae6055376e78afae773e2a21a5c2d/src/goto-symex/execution_state.cpp#L197-L203

This leads to an unconditional add current path to the set of paths to be preserved:
https://github.com/esbmc/esbmc/blob/7e9c1065ee2ae6055376e78afae773e2a21a5c2d/src/goto-symex/execution_state.cpp#L494-L495

Anthonysdu:
This PR fixes #1991.

Full bechexec run with 30s: https://github.com/esbmc/esbmc/actions/runs/12225704810

```
  correct:           18523
    correct true:    11184
    correct false:    7339
  incorrect:            83
    incorrect true:     62
    incorrect false:    21
  unknown:           14961
  Score:             27387 (max: 55885)
```
Master:

```
  correct:           18489
    correct true:    11159
    correct false:    7330
  incorrect:            82
    incorrect true:     61
    incorrect false:    21
  unknown:           14998
  Score:             27360 (max: 55885)
```

Concurrent: https://github.com/esbmc/esbmc/actions/runs/12226524553
```
  correct:             425
    correct true:      131
    correct false:     294
  incorrect:            10
    incorrect true:      6
    incorrect false:     4
  unknown:             290
```

### DIFF
--- a/regression/esbmc-solidity/github_497_5/test.desc
+++ b/regression/esbmc-solidity/github_497_5/test.desc
@@ -1,5 +1,4 @@
 CORE
 MyContract_array.solast
 --function func_array --falsification --sol MyContract_array.sol
-.*line 13 function func_array$
 ^VERIFICATION FAILED$
--- a/regression/esbmc-unix/global_guard/main.c
+++ b/regression/esbmc-unix/global_guard/main.c
@@ -1,0 +1,14 @@
+#include <pthread.h>
+#include <assert.h>
+pthread_t c;
+int d;
+void *e(void *) { d = 6; }
+int main() {
+  if (nondet_bool()) {
+    pthread_create(&c, NULL, e, NULL);
+    return 0;
+  }
+  d = 3;
+  assert(d == 0);  // should be fail
+  return 0;
+}
--- a/regression/esbmc-unix/global_guard/test.desc
+++ b/regression/esbmc-unix/global_guard/test.desc
@@ -1,0 +1,3 @@
+CORE
+main.c
+^VERIFICATION FAILED$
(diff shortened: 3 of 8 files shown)


lucasccordeiro:
@Anthonysdu: Can I ask you for the status of this PR?

Issue #2005, opened 2024-08-28
[C++, OM] `vector` mixes `new` and `realloc`


intrigus-lgtm:
While investigating https://github.com/esbmc/esbmc/pull/2003#issuecomment-2314559356
I reduced the test case to something like this:
```cpp
#include <assert.h>
#include <stdlib.h>
char *a = new char;
int main() {
  0 == realloc(a, 0);
  a[0] = 'a';
  a[1] = 'f';
  assert(a[0] != 'a');
}
```
this exits with an assertion failure in clang (as expected), but verifies correctly in esbmc (wrong).

This is consistent with vector using [new](https://github.com/esbmc/esbmc/blame/80c4e3b48bbd7bdf7ceb7b58d2458e68ddc917f7/src/cpp/library/vector#L129) and [realloc](https://github.com/esbmc/esbmc/blame/80c4e3b48bbd7bdf7ceb7b58d2458e68ddc917f7/src/cpp/library/vector#L208).

fbrausse:
There are problems with the current realloc implementation, see, e.g., <https://github.com/esbmc/esbmc/issues/1982#issuecomment-2290909339>. This issue could be related.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/commit/dff3113571aa57b8027b3ade0e9b1306851e9c98.

Issue #2040, opened 2024-09-24
[Vector OM] error: no matching constructor for initialization 


lucasccordeiro:
````CPP
#include <iostream>
#include <vector>
#include <string>
#include <algorithm>

// Define a struct for User (equivalent to interface in TypeScript)
struct User {
    int id;
    std::string name;
    std::string email;
    bool isAdmin; // Since C++ doesn't support optional values natively, we use a default value in constructors.

    User(int id, const std::string &name, const std::string &email, bool isAdmin = false)
        : id(id), name(name), email(email), isAdmin(isAdmin) {}
};

// A class for User management
class UserManager {
private:
    std::vector<User> users;

public:
    // Add a new user
    void addUser(const User& user) {
        users.push_back(user);
        std::cout << "User " << user.name << " added successfully." << std::endl;
    }

    // Get a user by their ID
    User* getUserById(int id) {
        for (auto &user : users) {
            if (user.id == id) {
                return &user;
            }
        }
        return nullptr; // Return nullptr if the user is not found
    }

    // List all users
    void listUsers() const {
        std::cout << "All Users:" << std::endl;
        for (const auto &user : users) {
            std::cout << "ID: " << user.id << ", Name: " << user.name
                      << ", Email: " << user.email
                      << (user.isAdmin ? ", Admin" : "") << std::endl;
        }
    }

    // Remove a user by their ID
    void removeUser(int id) {
        users.erase(std::remove_if(users.begin(), users.end(),
            [id](const User& user) { return user.id == id; }), users.end());
        std::cout << "User with ID " << id << " removed." << std::endl;
    }
};

int main() {
    // Instantiate UserManager
    UserManager userManager;

    // Create new users
    User user1(1, "Alice", "alice@example.com");
    User user2(2, "Bob", "bob@example.com", true);

    // Add users to the system
    userManager.addUser(user1);
    userManager.addUser(user2);

    // List all users
    userManager.listUsers();

    // Get user by ID
    User* user = userManager.getUserById(1);
    if (user != nullptr) {
        std::cout << "User found: " << user->name << std::endl;
    } else {
        std::cout << "User not found." << std::endl;
    }

    // Remove a user
    userManager.removeUser(1);
    userManager.listUsers();

    return 0;
}
````
````
$ esbmc typescript.cpp
ESBMC version 7.6.1 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing typescript.cpp
[...]

PR #2187, opened 2024-12-09
[C++ verification] Fix cpp new alloc size

XLiZHI:
This issue has been around since we used the wrong type to calculate the allocation size for cpp new in the past, which resulted in the failure to detect some buffer overflows, such as:

```c
char *p = new char[10];
p[11]; -->ver successful in master
```

The TCs failed is mainly due to the `push_back` in vector, which uses `realloc`, which has been problematic:
https://github.com/esbmc/esbmc/issues/2005
https://github.com/esbmc/esbmc/issues/1982#issuecomment-2290909339

I'm thinking about getting OM out of `realloc`.

### DIFF
--- a/regression/esbmc-cpp/algorithm/algorithm32/test.desc
+++ b/regression/esbmc-cpp/algorithm/algorithm32/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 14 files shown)


lucasccordeiro:
@XLiZHI: I got confused about this PR. You're changing the test desc of many regression tests from `CORE` to `UNKNOWNBUG`. Were they incorrectly labeled before?

XLiZHI:
> @XLiZHI: I got confused about this PR. You're changing the test desc of many regression tests from `CORE` to `UNKNOWNBUG`. Were they incorrectly labeled before?

This is due to bugs in realloc, and it exists before this PR. These TCs are passed before, just because they have the wrong allocation size (larger than it really is), allowing realloc to work by coincidence in some assertions.

It has to do with the size of allocation: https://github.com/esbmc/esbmc/issues/2040#issuecomment-2371614393.

lucasccordeiro (approved):

lucasccordeiro:

PR #2190, opened 2024-12-10
Don't use bigint for steps

mikhailramalho:
Every bigint is converted to uint64_t in the first place, so let's get rid of the conversion step and use uint64_t directly


lucasccordeiro (approved):

lucasccordeiro:

PR #2192, opened 2024-12-11
[Concurrency] Read global variable from function call

Anthonysdu:
There are missing context switch points when the arguments in a function call are global variables, leading to incorrect false. (See the test case.)


Anthonysdu:
benchexec: https://github.com/esbmc/esbmc/actions/runs/12270411020

Anthonysdu:
wait.. I wonder why the global var passed to that function didn't cause a context switch, should take a look before this pr.

Anthonysdu:
> wait.. I wonder why the global var passed to that function didn't cause a context switch, should take a look before this pr.

```
thread id: 1, 471, file main.c line 5 column 3 function a
c:main.c@65@F@a@b
```
the symbol is the function parameter name and wasn't considered as global.

lucasccordeiro:
> the symbol is the function parameter name and wasn't considered as global.

Shall we extract the function parameter name?

src/goto-symex/execution_state.cpp
```diff
@@ -301,8 +301,73 @@ void execution_statet::symex_step(reachability_treet &art)
...
+        dereference(operand, dereferencet::READ);
+
+    // Always run intrinsics, whether guard is false or not. This is due to the
+    // unfortunate circumstance where a thread starts with false guard due to
+    // decision taken in another thread in this trace. In that case the
+    // terminate intrinsic _has_ to run, or we explode.
+    if (is_symbol2t(call.function))
+    {
+      const irep_idt &id = to_symbol2t(call.function).thename;
+      if (has_prefix(id.as_string(), "c:@F@__ESBMC"))
```
ChenfengWei1: Does it also work in C++?

Anthonysdu: This function is from it's parent class, I presume it should work.

But this PR is broken with some tests, will need to check it later.

Anthonysdu:
> > the symbol is the function parameter name and wasn't considered as global.
> 
> Shall we extract the function parameter name?

The argument was simplified to a boolean value in symex_goto that's why we lost track of it.

src/goto-symex/execution_state.cpp
ChenfengWei1: In my experience, the function name in C++ might be different compared to C, but I am not sure if it’s the case for the intrinsics function.

src/goto-symex/symex_function.cpp
Anthonysdu, line 1: It's werid that the comment above said read the arguments before renaming but it didn't actually read them.

src/goto-symex/symex_function.cpp
```diff
@@ -251,7 +251,10 @@ void goto_symext::symex_function_call_code(const expr2tc &expr)
   // read the arguments -- before the locality renaming
```
Anthonysdu, line 251: It's werid that the comment above said read the arguments before renaming but it didn't actually read them.

Anthonysdu:
benchexec:https://github.com/esbmc/esbmc/actions/runs/12306951840

lucasccordeiro (approved):

lucasccordeiro:

XLiZHI (approved):

PR #2193, opened 2024-12-11
Prevent recursion when parsing a member function that calls itself

mikhailramalho:
It should replace PR #2011 

We split the handling of members and removed the recursive call that may cause an infinite recursion.

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/63_recursion_cxx_member_call/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/63_recursion_cxx_member_call/main.cpp
@@ -1,0 +1,21 @@
+#include <cassert>
+
+struct a
+{
+  int b(int i)
+  {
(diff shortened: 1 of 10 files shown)


mikhailramalho:
bitfiels are broken with this patch, I'll check it tomorrow.

mikhailramalho:
This now enables two tests that were previously failing for the same reason (infinite recursion).

My only complaint is the new `get_bitfield_type` method: it breaks the standard of other `get_*_type` methods where we pass a clang type and get an esbmc type in return; `get_bitfield_type` takes a FieldDecl and two esbmc types (the second is the return type). I thought of naming it `get_bitfield_type_from_Decl` but it sounded too verbose.

lucasccordeiro (approved):

lucasccordeiro:

PR #2196, opened 2024-12-13
[python-frontend] Handling module aliases

brcfarias:
This PR enables type inference and verification for functions imported and invoked through aliases.

- Add a module manager to extract function return type from imported modules.
- Support function calls from aliased modules

**Example**
**_other.py_**
```python
def sum(a:int, b:int) : int ->
  return a+b
```

_**main.py:**_
```python
import other as ot
x = 1
y = 2
z = ot.sum(x,y) # imported function with alias
```


regression/python/import-as/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: Can you also add a test case that provides `VERIFICATION FAILED`? For example, you could check for overflow.

src/python-frontend/json_utils.h
```diff
@@ -94,6 +96,39 @@ JsonType &find_function(JsonType &json, const std::string &func_name)
...
+    }
+  }
+  throw std::runtime_error("Function " + func_name + " not found\n");
+}
+
+template <typename JsonType>
+const std::string
+get_object_alias(const JsonType &ast, const std::string &obj_name)
+{
+  for (auto &node : ast["body"])
```
lucasccordeiro, line 119: This loop iterates through all nodes and names, which can become expensive for large ASTs. 

Could we consider stopping early once the target is found or preprocessing the AST into a more query-efficient structure?

src/python-frontend/json_utils.h
```diff
@@ -94,6 +96,39 @@ JsonType &find_function(JsonType &json, const std::string &func_name)
   throw std::runtime_error("Function " + func_name + " not found\n");
 }
 
+template <typename JsonType>
+JsonType &find_imported_function(JsonType &ast, const std::string &func_name)
```
lucasccordeiro: Shall we consider `const JsonType &ast`?

lucasccordeiro (changes requested):

src/python-frontend/module.h
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  void add_submodule(const std::shared_ptr<module> mod)
+  {
+    submodules_.insert(mod);
+  }
+
+  function get_function(const std::string &func_name) const
+  {
+    auto it = std::find_if(
```
lucasccordeiro, line 60: Could we use a `std::unordered_map` or `std::map` instead of `std::vector`? This would enable O(1) (unordered_map) or O(log n) (map) lookups.

src/python-frontend/module.h
```diff
@@ -0,0 +1,70 @@
...
+
+    return {};
+  }
+
+  const std::vector<function> &functions() const
+  {
+    return functions_;
+  }
+
+  const std::unordered_set<std::shared_ptr<module>> &submodules() const
```
lucasccordeiro: Shall we consider std::unique_ptr? If shared ownership is required, consider documenting why shared ownership is necessary here.

src/python-frontend/python_annotation.h
```diff
@@ -330,28 +348,76 @@ class python_annotation
...
     std::string type("");
 
-    const std::string &obj = get_object_name(call["func"]);
+    const std::string &obj = get_object_name(call["func"], std::string());
 
-    python_module pm(obj);
-    if (pm.is_standard_module())
-      return pm.get_function_return(call["func"]["attr"]);
+    // Get type from imported module
+    if (mm_)
```
lucasccordeiro: Shall we use a more suggestive name for this `_mm`?

src/python-frontend/json_utils.h
brcfarias: Thanks for the suggestion. The function stops early with a return once the target alias is found.

src/python-frontend/module_manager.cpp
```diff
@@ -0,0 +1,214 @@
...
+      if (!current_module && !module_name.empty())
+      {
+        current_module = std::make_shared<module>(module_name);
+        modules_.insert(current_module);
+      }
+    }
+    else
+    {
+      // If there is a parent module, create or get it as a submodule
+      auto existing_submodule = std::find_if(
```
lucasccordeiro, line 143: Shall we use an associative container (`std::unordered_map`) for faster lookups?

src/python-frontend/module_manager.cpp
```diff
@@ -0,0 +1,214 @@
...
+void module_manager::load_directory(
+  const fs::path &current_path,
+  std::shared_ptr<module> parent_module)
+{
+  for (const auto &entry : fs::directory_iterator(current_path))
+  {
+    if (entry.is_regular_file() && entry.path().extension() == ".json")
+    {
+      // Create a module for the JSON file
+      auto submodule = create_module(entry.path());
```
lucasccordeiro, line 79: Shall we wrap the JSON parsing code in a try-catch block to handle exceptions?

lucasccordeiro (approved):
Thanks for addressing the comments, @brcfarias.

I have reviewed the code again. I added just a few suggestions to consider for improvement.

Apart from that, LGTM.

lucasccordeiro:

PR #2200, opened 2024-12-16
[Concurrency] Reduce context switch with shared variables only

Anthonysdu:
esbmc does context switch at any global variables, this pr reduces the switch points that will only consider switches if a variable is shared between threads. E.g, a variable is global but only accessed in single thread, then different thread interleaves won't make a difference. (nothing shared memory access happened between threads)

This PR: https://github.com/esbmc/esbmc/actions/runs/12348994796/job/34459820661

```
  correct:             431
    correct true:      135
    correct false:     296
  incorrect:            14
    incorrect true:     10
    incorrect false:     4
  unknown:             280
  Score:               182 (max: 1099)
```
Total time: 211000s 

master: 

```
  correct:             425
    correct true:      131
    correct false:     294
  incorrect:            10
    incorrect true:      6
    incorrect false:     4
  unknown:             290
```
Total time : 220000s

It increases the incorrect true though (Timeout to true), because those benchmarks that start with `fib_unsafe` need at least `context bound 7` to detect the violation.

image


Anthonysdu:
All of the failed regressions report:

`esbmc: /home/runner/work/esbmc/esbmc/src/pointer-analysis/value_set.cpp:545: void value_sett::get_value_set_rec(const expr2tc&, value_sett::object_mapt&, const string&, const type2tc&, bool) const: Assertion `sym.rlevel != symbol2t::renaming_level::level2_global' failed.`

I removed this condition because it should perform analysis before all new threads are created via pthread_create,

https://github.com/esbmc/esbmc/blob/9c9839608ae9f45ccaf7a4c2fadb0aada3183f74/src/goto-symex/execution_state.cpp#L308

then it will trigger the assertion when searching global vars here:
https://github.com/esbmc/esbmc/blob/9c9839608ae9f45ccaf7a4c2fadb0aada3183f74/src/goto-symex/execution_state.cpp#L874

Does anyone know why the symbol should not be level2 global in the value set?

lucasccordeiro:
> Does anyone know why the symbol should not be level2 global in the value set?

@rafaelsamenezes and @mikhailramalho could help here (and correct me if I'm wrong).

Level-2 global symbols represent globally-scoped variables uniquely renamed for each program location. Including such symbols in the value set may lead to inconsistencies or incorrect analysis results due to mismatched scopes or states.

lucasccordeiro:
See: https://github.com/esbmc/esbmc/blob/master/docs/manual.tex

"Global program state is also stored here as a \url{value_sett} object to track pointer value sets, and a \url{renaming::level2t} storing data on the latest assignments to variables."

PR #2207, opened 2024-12-26
[C verification] support clang __builtin_va_*

### LINKED ISSUE #2194: WARNING:  no body for function __builtin_va_start
The code simply uses va_start and va_end. Based on looking at the output from goto2c, it appears an extern function declaration is created, and the function is invoked normally.

Does` the lack of a function body make the analysis unsound? I expect that these types of warnings are generally to alert users to provide source for the missing functions. In this case however, I have no idea what esbmc would expect, considering va_start appears to be tightly coupled to clang.

The command line is simply esbmc --incremental-bmc test.c

Furthermore, though this may be unrelated, incremental-bmc terminates with verification succeeded, solution found by the forward conditin, all states are reachable. k=1. That seems incorrect to me, and so I can't help but think it may be related.

ESBMC version: 7.7
Operating system: Ubuntu 24.10
Exact command line resulting in the issue: esbmc --incremental-bmc test.c
What behaviour did you expect: No warnings
What happened instead: Got warnings

XLiZHI:


lucasccordeiro (approved):

lucasccordeiro:

PR #2081, opened 2024-10-16
Simplified Trace

Yiannis128:
This PR adds a simplified trace to ESBMC, using the `simplified-trace` flag option. The main goal of this PR is to present the counterexample trace in a less verbose, more human-readable format. Discussion on said format is welcome.

Closes #2080

Issue #2080, opened 2024-10-16
Simplified counter-example traces


ChenfengWei0:
https://github.com/esbmc/esbmc-ai/issues/149

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2208.

PR #2208, opened 2024-12-27
add cex-only && simplify-trace options

ChenfengWei0:
- `cex-only`: `do not print the state trace`
- `simplify-trace`: `simplify the trace and exclude the assignments whose variables are not from user-input files`

The `simplify-trace` can work with options like `compact-trace` and `cex-only`

#2081 
 #2080


lucasccordeiro (approved):

intrigus-lgtm (commented):
Not a big fan of adding all those ifs everywhere.
Is there no cleaner way or possibility to put all those checks into a single place?

src/goto-symex/goto_trace.cpp
```diff
@@ -360,13 +385,44 @@ void correctness_graphml_goto_trace(
...
 }
 
+void show_simplified_location(std::ostream &out, const locationt &location)
+{
+  std::string dest;
+  const irep_idt &file = location.get_file();
+  const irep_idt &line = location.get_line();
+  const irep_idt &function = location.get_function();
+
+  if (file != "")
```
lucasccordeiro, line 423: You could improve this code. For example:

````CPP
void appendInfo(std::string &dest, const std::string &label, const std::string &value)
{
    if (!value.empty())
    {
        if (!dest.empty())
            dest += " ";
        dest += label + " " + id2string(value);
    }
}

if (!file.empty())
    appendInfo(dest, "file", file);

if (!line.empty())
    appendInfo(dest, "line", line);

if (!function.empty())
    appendInfo(dest, "function", function);
````

**Review by lucasccordeiro — CHANGES_REQUESTED** [review]

**Comment by lucasccordeiro** [comment]
> Is there no cleaner way or possibility to put all those checks into a single place?

I agree with @intrigus-lgtm. I have added one suggestion here.

File: src/goto-symex/goto_trace.cpp (hunk 1 above)
Comment (ChenfengWei0) [reply src/goto-symex/goto_trace.cpp:423]: Changed as requested

**Review by lucasccordeiro — APPROVED** [review]

**Comment by lucasccordeiro** [comment]
@intrigus-lgtm: are you happy with this PR version?

File: src/goto-symex/goto_trace.cpp [hunk 2]
```diff
@@ -411,32 +478,43 @@ void show_goto_trace(
...
-        step.pc->is_assign() || step.pc->is_return() ||
-        (step.pc->is_other() && is_nil_expr(step.lhs)) ||
-        step.pc->is_function_call())
+        !cex_only && (step.pc->is_assign() || step.pc->is_return() ||
+                      (step.pc->is_other() && is_nil_expr(step.lhs)) ||
+                      step.pc->is_function_call()))
       {
+        if (
+          simplify_trace &&
+          !input_file_check(step.pc->location.get_file().as_string()))
```
intrigus-lgtm: I would assume that this can fail with a segfault when a step's location does not have an associated file (or is this actually impossible?).
Especially for C++, where we generate quite some code for vtables and other stuff, we didn't really generate any locations for the code (It's also unclear what location to use for constructs that have no source-code location).

src/goto-symex/goto_trace.cpp
```diff
@@ -225,17 +225,42 @@ void show_goto_trace_gui(
   }
 }
 
+// Return true if the file is from the user-input
```
intrigus-lgtm: What is user-input for you?

I personally would not have considered include files as user-input.
As said in another comment, some instructions don't have any location.

src/goto-symex/goto_trace.cpp
```diff
@@ -375,11 +431,22 @@ void show_goto_trace(
     case goto_trace_stept::ASSERT:
       if (!step.guard)
       {
-        show_state_header(out, step, step.pc->location, step.step_nr);
+        if (!cex_only)
```
intrigus-lgtm: Why do we not show a state header in this case?
If `--show-stacktrace` and `--cex-only` are passed, we still show a stack trace, but no state header?

src/goto-symex/goto_trace.cpp
```diff
@@ -360,13 +385,44 @@ void correctness_graphml_goto_trace(
...
   const namespacet &ns,
   const goto_tracet &goto_trace)
 {
   unsigned prev_step_nr = 0;
   bool first_step = true;
+  bool cex_only = config.options.get_bool_option("cex-only");
+  bool simplify_trace = config.options.get_bool_option("simplify-trace");
 
   for (const auto &step : goto_trace.steps)
   {
```
intrigus-lgtm, line 443: Can't we simplify the code to this?
This should allow us to remove all those `if(!cex_only)` statements.
```suggestion
  for (const auto &step : goto_trace.steps)
  {
    // we only care about the counter example, which is only triggered by assert steps. Ignore all other steps.
    if (cex_only && step.type != goto_trace_stept::ASSERT)
      continue;
```

intrigus-lgtm (changes requested):
Some things that could simplify this even more.

src/goto-symex/goto_trace.cpp
ChenfengWei0: I believe it has a default value `""`.

src/goto-symex/goto_trace.cpp
ChenfengWei0: > What is user-input for you?

```c++
input_file = options.get_option("input-file");
include_files = cmdline.get_values("include-file");
```

> some instructions don't have any location

I am not sure why we need to worry about the instructiont that has no file location. If it has no location, then it will be simplified, since it's internally created rather than user-input

src/goto-symex/goto_trace.cpp
ChenfengWei0: > Why do we not show a state header in this case?

Cex-only: do not print the state trace.

> If --show-stacktrace and --cex-only are passed, we still show a stack trace, but no state header?

show-stacktrace: show function call trace. E.g.
```sh
# --show-stack-trace --cex-only

[Counterexample]

Violated property:
  file 1.c line 3 column 22 function reach_error
Stack trace:
  c:@F@reach_error at file 1.c line 1646 column 11 function ssl3_connect
  c:@F@ssl3_connect at file 1.c line 1076 column 3 function main
  c:@F@main
  assertion 0
  0


VERIFICATION FAILED

Bug found (k = 7)
```

src/goto-symex/goto_trace.cpp
ChenfengWei0: The idea is to only print out the trace step that is related to the files provided by the user

src/goto-symex/goto_trace.cpp
ChenfengWei0: Changed as suggested

src/goto-symex/goto_trace.cpp
intrigus-lgtm: Internally created instructions can exist in user-provided files.

E.g. all code related to vtable initialization is internally created from user-provided files.

ChenfengWei0: Sorry for not having the background knowledge of the vtable initialization.

> ll code related to vtable initialization is internally created from user-provided files.

Are these "code related to vtable initialization" already in the input files before feeding into ESBMC? If so, why does it not having a file location? If not, e.g. inserted in the frontend/GOTO stage, I believe they should be filtered.

ChenfengWei0: Also if the location is missing, its output will confuse our users anyway.

lucasccordeiro:
@intrigus-lgtm: Do you still have comments about this PR?

lucasccordeiro:

PR #2216, opened 2025-01-01
Update README.md for M1 mac

sshivaji:


lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @sshivaji.

intrigus-lgtm:
@sshivaji can you explain what exactly isn't working on arm Macs?
I just keep hearing that it doesn't work, but I haven't seen a concrete issue that one could fix.
While using amd64 is an easy workaround to any potential problems, I think that it would be best to fix those issues and to provide the best experience to users.
(I don't have any arm Macs)

sshivaji:
I think it's fine now. I tried tests on many files, I think it is decent
now, most of the problems were with float_16. I did not know that was a
major thing for CLANG on the M1/ARM Mac.

I appreciate the speed at which we are addressing issues. I will open any
issue I find. We do need to clean up the make build process to make it
easier for new people trying the ARM Mac. Always better to have a native
solution than the complex docker amd64 workaround.

PR #2221, opened 2025-01-06
GCSE now only caches overflow inner operands

rafaelsamenezes:
GCSE was caching the overflow operand, which caused ESBMC to be confused down the line. This PR now only cache each operand operators.


lucasccordeiro (approved):

lucasccordeiro:

PR #2234, opened 2025-01-08
Force update of tags in benchexec action

rafaelsamenezes:
Git is complaining that the remote tags have changed. It should be fine to always force update. Example: https://github.com/esbmc/esbmc/actions/runs/12668133362/job/35303938167


lucasccordeiro (approved):

lucasccordeiro:

Issue #2237, opened 2025-01-09
[Python frontend] Wrong simplification for strings


lucasccordeiro:
Even if we call `esbmc class.py --no-simplify --no-propagation`, ESBMC still produces an incorrect result.

````Python
def f(x:str) -> str:
    return x

k = f("abc")

assert k == "abc"
````

GOTO representation:

````
__ESBMC_main (__ESBMC_main):
        // 17 no location
        FUNCTION_CALL:  k=f(&{ 97, 98, 99 })
        // 18 no location
        ASSERT 0
        // 19 no location
        END_FUNCTION
````

Symbol table:

````
Symbol......: __ESBMC_main
Module......:
Base name...: __ESBMC_main
Mode........:
Type........: void ()
Value.......:
{

{
k=f(&{ 97, 98, 99 })
  assert(0);
  }
}
Flags.......: lvalue
Location....:
````

lucasccordeiro:
This code fragment might be the culprit of this issue:

https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_converter.cpp#L404-L430

lucasccordeiro:
Reduced version that exposes the bug:

````Python
def f(x: str) -> str:
    return x

assert f("abc") == "abc"
````

lucasccordeiro:
I have reduced the program size to understand the root cause of the problem better.

````Python
def f(x: str) -> str:
    return x
assert f("abc") == "abc"
````

The bug is in this method: https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_converter.cpp#L351-L549.

The left-hand side of the assert statement contains a function call, which the `get_binary_operator_expr` method fails to handle correctly for string comparison. This limitation prevents the correct evaluation of the equality expression in the assert statement.

The method needs to be refactored to address this. Over time, we have added numerous `if` statements to handle specific cases. Still, this ad hoc approach does not scale well, and the method must support an increasing variety of statement types.

Refactoring it into a recursive method would enable the code to systematically break down and handle complex expressions, improving maintainability and extending support for additional cases in the future.

sshivaji:
To reproduce:

1. Create a file simple.py with

````
def f(x: str) -> str:
    return x
assert f("abc") == "abc"
````

2. Run esbmc on the file:
````
esbmc simple.py
````

It will return VERIFICATION failed.

3. esbmc simple.py --goto-functions-only

Will show a goto trace:

[...]

PR #2239, opened 2025-01-10
Partial fix for assert

sshivaji:
This is a partial fix, there is still an error in GOTO:

__ESBMC_main (__ESBMC_main):
        // 17 file class.py line 3 column 7
        DECL signed char [3] return_value$_f$1;
        // 18 
        FUNCTION_CALL:  return_value$_f$1=f(&{ 97, 98, 99 }[0])
        // 19 file class.py line 3 column 0
        ASSERT &return_value$_f$1[0] == &{ 97, 98, 99 }[0] (this line should read ASSERT &return_value$_f$1 == &{ 97, 98, 99 }[0]) we dont need the [0] on the LHS expression
        // 20 no location
        DEAD __ESBMC_main::$tmp::return_value$_f$1
        // 21 no location
        END_FUNCTION

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -373,7 +373,6 @@ exprt python_converter::get_binary_operator_expr(const nlohmann::json &element)
     expr = side_effect;
   };
 
-  // Function calls in expressions like "fib(n-1) + fib(n-2)" need to be converted to side effects
   if (lhs.is_function_call())
     to_side_effect_call(lhs);
   if (rhs.is_function_call())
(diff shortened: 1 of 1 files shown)


lucasccordeiro:
Thanks, @sshivaji. As described here (https://github.com/esbmc/esbmc/issues/2237#issuecomment-2581192057), we must handle this function call to perform the string comparison, which is missing in the current implementation of the `get_binary_operator_expr`:

````
sideeffect
  * type: array
      * size: constant
          * type: unsignedbv
              * width: 64
          * value: 0000000000000000000000000000000000000000000000000000000000000011
          * #cformat: 3
      * subtype: signedbv
          * width: 8
  * operands:
    0: symbol
        * type: code
            * arguments:
              0: argument
                  * type: pointer
                      * subtype: signedbv
                          * width: 8
                  * #location:
                    * file: class.py
                    * line: 1
                    * function: f
                    * column: 6
                  * #base_name: x
                  * #identifier: py:class.py@F@f@x
            * return_type: array
                * size: constant
                    * type: unsignedbv
                        * width: 64
                    * value: 0000000000000000000000000000000000000000000000000000000000000011
                    * #cformat: 3
                * subtype: signedbv
                    * width: 8
        * name: f
        * identifier: py:class.py@F@f
    1: arguments
        * operands:
          0: address_of
              * type: pointer
                  * subtype: array
                      * size: constant
                          * type: unsignedbv
                              * width: 64
                          * value: 0000000000000000000000000000000000000000000000000000000000000011
                          * #cformat: 3
                      * subtype: signedbv
                          * width: 8
              * operands:
                0: constant
                    * type: array
                        * size: constant
                            * type: unsignedbv
                                * width: 64
                            * value: 0000000000000000000000000000000000000000000000000000000000000011
                            * #cformat: 3
                        * subtype: signedbv
                            * width: 8
                    * operands:
                      0: constant
                          * type: signedbv
                              * width: 8
                          * value: 01100001
                          * #cformat: 97
                      1: constant
                          * type: signedbv
                              * width: 8
                          * value: 01100010
                          * #cformat: 98
                      2: constant
                          * type: signedbv
                              * width: 8
                          * value: 01100011
                          * #cformat: 99
  * statement: function_call
  * #location:
    * file: class.py
    * line: 4
    * function:
    * column: 7
 ````

I would recommend checking why this `function_call` is not handled in `get_binary_operator_expr`.

As described here (https://github.com/esbmc/esbmc/issues/2237#issuecomment-2581192057), after we fix this issue, we need to refactor this code. It is getting too messy with many "if"-statements.

intrigus-lgtm (changes requested):
Why did you remove all those comments?
Also, I don't understand why we are modelling string literals as arrays of chars?
A string literal is a proper object of type `<class 'str'>` so we should IMHO also model it as such and not as some weird array of chars.

lucasccordeiro:
> A string literal is a proper object of type <class 'str'> so we should IMHO also model it as such and not as some weird array of chars.

I think it comes from @brcfarias' original implementation since he's using `strncmp` to compare the array of chars: https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_converter.cpp#L416-L429

intrigus-lgtm:
> > A string literal is a proper object of type <class 'str'> so we should IMHO also model it as such and not as some weird array of chars.
> 
> I think it comes from @brcfarias' original implementation since he's using `strncmp` to compare the array of chars: https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_converter.cpp#L416-L429

A better long term fix, would be to add a proper class for strings that internally has an array of chars and then compares that in its `__eq__` method using `strncmp`.

lucasccordeiro:
> A better long term fix, would be to add a proper class for strings that internally has an array of chars and then compares that in its __eq__ method using strncmp.

I agree.

sshivaji:
Added Lucas comments he sent me via message:


  
  [1/9/25, 10:20:02 PM] Lucas: lhs.dump();
  printf("lhs_type: %s\n", lhs_type.c_str());
  rhs.dump();
  printf("rhs_type: %s\n", rhs_type.c_str());
  if (lhs_type == "str" && rhs_type == "str")
  {

[1/9/25, 10:20:09 PM] Lucas: you can dump the expressions
[1/9/25, 10:20:25 PM] Lucas: you can see that we have
[1/9/25, 10:20:36 PM] Lucas: lhs_type:
rhs_type: str
[1/9/25, 10:20:47 PM] Lucas: the lhs_type is empty
[1/9/25, 10:21:01 PM] Lucas: so, we don't handle it in exprt python_converter::get_binary_operator_expr(const nlohmann::json &element)
[1/9/25, 10:21:02 PM] Shivkumar Shivaji: wow, thanks!
[1/9/25, 10:21:34 PM] Lucas: you need to fill out the lhs member and then handle the case, which is a function call
[1/9/25, 10:21:36 PM] Shivkumar Shivaji: yes, i could tell that LHS was empty
[1/9/25, 10:21:42 PM] Shivkumar Shivaji: but i did not know how to fix it
[1/9/25, 10:22:05 PM] Lucas: you need to handle the function call inside python_converter::get_binary_operator_expr(const nlohmann::json &element)
[1/9/25, 10:22:17 PM] Lucas: this is why I mention this method needs refactoring
[1/9/25, 10:22:39 PM] Lucas: Bruno has a method to handle function_call, but it is not used in  python_converter::get_binary_operator_expr(const nlohmann::json &element)

sshivaji:
@intrigus-lgtm Good point on comments, will bring them back!

PR #2250, opened 2025-01-19
[python-frontend] Refactoring function call handling

brcfarias:
A refactoring to improve organization and readability consisting of:
- Moving the logic for generating `typet` from the `python_convert` class to the `type_handler` class.
- Moving the logic for constructing `exprt` objects for function calls from the `python_convert` class to the `function_call_expr` class.


src/python-frontend/python_converter.cpp
```diff
@@ -679,7 +560,7 @@ python_converter::find_symbol_in_global_scope(std::string &symbol_id) const
   if (func_start_pos != std::string::npos)
     symbol_id.erase(func_start_pos, func_end_pos - func_start_pos);
 
-  return context.find_symbol(symbol_id);
+  return symbol_table_.find_symbol(symbol_id);
 }
 
 std::string python_converter::get_classname_from_symbol_id(
```
intrigus-lgtm: Remove if unused?

src/python-frontend/function_call_expr.cpp
```diff
@@ -0,0 +1,446 @@
...
+  }
+}
+
+static std::string get_classname_from_symbol_id(const std::string &symbol_id)
+{
+  // This function might return "Base" for a symbol_id as: py:main.py@C@Base@F@foo@self
+
+  std::string class_name;
+  size_t class_pos = symbol_id.find("@C@");
+  size_t func_pos = symbol_id.find("@F@");
```
intrigus-lgtm, line 62: Is it possible to avoid parsing strings?
This will very likely bite someone in the future.

src/python-frontend/function_call_expr.h
```diff
@@ -0,0 +1,46 @@
...
+
+class function_call_expr
+{
+public:
+  function_call_expr(const nlohmann::json &call, python_converter &converter);
+
+  exprt build();
+
+private:
+  void build_function_id();
```
lucasccordeiro, line 31: Can you add some documentation for these methods? You can follow the same patterns as we have on https://github.com/esbmc/esbmc/blob/master/src/goto-symex/goto_symex_state.h.

src/python-frontend/type_handler.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <util/c_types.h>
+#include <nlohmann/json.hpp>
+
+class python_converter;
+
+class type_handler
+{
+public:
+  type_handler(const python_converter &converter);
```
lucasccordeiro, line 11: Can you add some documentation for these methods? You can follow the same patterns as we have on https://github.com/esbmc/esbmc/blob/master/src/goto-symex/goto_symex_state.h.

lucasccordeiro (changes requested):

src/python-frontend/python_converter.cpp
brcfarias: Thanks for pointing this out.

src/python-frontend/function_call_expr.cpp
brcfarias: Yes, I plan to address this. Thanks for the suggestion.

lucasccordeiro (approved):

lucasccordeiro:

PR #2252, opened 2025-01-20
update llvm16 for windows

Anthonysdu:


rafaelsamenezes:
@Anthonysdu did you follow the same build instructions from the Wiki or you had to something different? If so, could you also update it?

Anthonysdu:
> @Anthonysdu did you follow the same build instructions from the Wiki or you had to something different? If so, could you also update it?

I did something different, but yeah will update the wiki once the CI is fine.

PR #2259, opened 2025-01-31
Loop invariants checker

rafaelsamenezes:
This is a rebase of `ahmed-vampire-for-loops`. The main idea of this branch is to use Vampire to check if loop invariants hold and then, if yes, replace the loop with it. I may even split these in several PRs

## Quantifiers

This will introduce `__ESBMC_forall` and `__ESBMC_exists` quantifiers.

- [ ] Add quantifier expressions
- [X] Add smtlib support
- [ ] Add support for solver APIs (specially Z3)
- [ ] Add test cases

The initial proposal was using `__quantifier(void*, _Bool) `, for example:

```c
int x;
if (x > 10)
__ESBMC_forall(&x, x > 0);
```

I think that should hold.

## Incremental SMT with Vampire

The changes are a bit too Vampire heavy, they could be more agnostic.

- [ ] Check smtlib solving through pipes/binary
- [ ] Move Vampire specific mentions into solvers

## Invariants

This PR introduces a new instruction: INVARIANT. They reffer to loop invariants used by the main algorithm. They can probably be incorporated into an ASSUMPTION.

- [ ] Extend assumptions with two boolean fields: `is_loop_invariant`, `is_candidate`
- [ ] Adapt code to make ESBMC ignore candidate assumptions.

The loop invariants are sequentially read and maps into the next loop. This could probably be a custom algorithm

- [ ] Convert the loop invariant class to an algorithm

Finally at symex, in a loop with the candidate... invoke the smt solver.

- [ ] Call smt solver for candidate invariants


selig:
In this

```
int x;
if (x > 10)
__ESBMC_forall(&x, x > 0);
```

I think it would better be

```
int x;
if (x > 10)
assert(__ESBMC_forall(&y, y < 0 ==> x > y));
```

i.e. we need to wrap it in `assert` an we want something that's true. `forall x :: x > 0` is false... x could be -1. The x in the quantified expression is not the same one that just got defined. `forall y :: y < 0 ==> x > y` is true as we know `x > 10` ... any value less than 0 is also less than 10.

selig:
Can `is_candidate` be `is_candidate_invariant` or `is_candidate_loop_invariant` to be more descriptive?

lucasccordeiro:
This PR is no longer valid.

PR #2261, opened 2025-01-31
Support for SMT quantifiers

rafaelsamenezes:
As shown in #2259 some flows need to use quantifiers, and since we are actually using SMT... we should exploit.

Here I introduce two new expressions: `bool forall(symbol, predicate)` and `bool exists(symbol, predicate)`. Where:
- Symbol is going to be replaced with a new symbolic value (as in a closure)
- If the symbol is also in the predicate, then it will be replaced to the same name as the LHS (to avoid ssa renumbering)

```c
extern void __ESBMC_assume(_Bool);                  
extern _Bool __ESBMC_forall(void *, _Bool);
extern _Bool __ESBMC_exists(void *, _Bool);

int main()
{
  unsigned n;
  int arr[n];
  unsigned i;
  __ESBMC_assume(__ESBMC_forall(&i, !(i < n) || arr[i] == 2));
  __ESBMC_assert(!__ESBMC_exists(&i,  (i < n) && arr[i] == 42), "forall init");
 // OK
  arr[n/2] = 42;                                      

  __ESBMC_assert(!__ESBMC_exists(&i,  (i < n) && arr[i] == 42), "forall init");
 // Fails
}

```

Limitations:

- No smtlib (only Z3)
- No support for multiple symbols, eventually we would like to have something like `forall({i,j}. i < j ==> arr[i] < arr[j]`
- No support for recursive calls `forall i. forall j. foo`
- There is a bug that when the bonded symbol has a constant value it may replace the predicate with the value. For example, `i = 0; exists i . i != 0` fails because of that.  Our simplifier is smart enough to replace it with the value, breaking the symbolic chain.


src/goto-programs/builtin_functions.cpp
```diff
@@ -510,6 +511,31 @@ void goto_convertt::do_function_call_symbol(
...
+    if (arguments.size() != 2)
+    {
+      log_error("`{}' expected to have two arguments", id2string(base_name));
+      abort();
+    }
+    // make it a side effect if there is an LHS
+    if (lhs.is_nil())
+      return;
+
+    exprt rhs =
```
lucasccordeiro, line 526: Can you document this part of the code?

src/solvers/smt/smt_conv.cpp
```diff
@@ -177,6 +178,25 @@ void smt_convt::push_ctx()
   ctx_level++;
 }
 
+std::string
+get_last_name_from_body(const std::string original, const expr2tc body)
```
lucasccordeiro: Can you write some documentation here?

src/solvers/z3/z3_conv.cpp
```diff
@@ -1695,3 +1695,37 @@ smt_astt z3_convt::mk_smt_fpbv_abs(smt_astt op)
       z3_ctx, Z3_mk_fpa_abs(z3_ctx, to_solver_smt_ast<z3_smt_ast>(op)->a)),
     op->sort);
 }
+
+smt_astt z3_convt::mk_quantifier(
```
lucasccordeiro: Can you write some documentation?

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -443,6 +443,9 @@ void __VERIFIER_atomic_end();
 /* Causes a verification error when its call is reachable; internal use in math
  * models */
 void __ESBMC_unreachable();
+// Quantifiers
```
lucasccordeiro: Can you test these intrinsic functions with this example?

````C
extern void __ESBMC_assume(_Bool);
extern _Bool __ESBMC_forall(void *, _Bool);
extern _Bool __ESBMC_exists(void *, _Bool);

int main()
{
  unsigned n;
  int arr[n];
  unsigned i;

  // Ensure valid `n`
  __ESBMC_assume(n > 0);

  // Avoid buffer under-/overflow
  __ESBMC_assume(i >= 0 && i < n);

  // Case 1: All elements initialized to a specific value
  __ESBMC_assume(__ESBMC_forall(&i, arr[i] == 2));
  __ESBMC_assert(!__ESBMC_exists(&i, arr[i] == 42), "forall init - all 2s");

  // Modify an element in the middle
  arr[n/2] = 42;
  __ESBMC_assert(!__ESBMC_exists(&i, arr[i] == 42), "forall init - fails expected");

  // Case 2: All elements initialized to 0
  __ESBMC_assume(__ESBMC_forall(&i, arr[i] == 0));
  __ESBMC_assert(!__ESBMC_exists(&i, arr[i] == 42), "forall init - all 0s");

  // Modify last element
  arr[n-1] = 42;
  __ESBMC_assert(!__ESBMC_exists(&i, arr[i] == 42), "forall init - last modified");

  // Case 3: Ensure at least one value is not 2
  __ESBMC_assume(!__ESBMC_forall(&i, arr[i] == 2));
  __ESBMC_assert(__ESBMC_exists(&i, arr[i] != 2), "exists - at least one not 2");

  // Case 4: Randomized values
  __ESBMC_assume(arr[0] == 10);
  __ESBMC_assume(arr[n-1] == 20);
  __ESBMC_assert(__ESBMC_exists(&i, arr[i] == 10), "exists - first element check");
  __ESBMC_assert(__ESBMC_exists(&i, arr[i] == 20), "exists - last element check");

  return 0;
}
````

**Review by lucasccordeiro — CHANGES_REQUESTED** [review]

**Comment by lucasccordeiro** [comment]
@rafaelsamenezes : I have just compiled your branch.

Our current implementation fails in this case:

````C
#include <assert.h>

int main()
{
  unsigned n;
  int arr[n];
  unsigned i;

  __ESBMC_assume(n > 0);

  __ESBMC_assume(i >= 0 && i < n);

  arr[n/2] = 42;
  assert(!__ESBMC_exists(&i, arr[i] == 42));

  return 0;
}
````

**Comment by rafaelsamenezes** [comment]
> @rafaelsamenezes : I have just compiled your branch.
> 
> Our current implementation fails in this case:
> 
> ```c
> #include <assert.h>
> 
> int main()
> {
>   unsigned n;
>   int arr[n];
>   unsigned i;
> 
>   __ESBMC_assume(n > 0);
> 
>   __ESBMC_assume(i >= 0 && i < n);
> 
>   arr[n/2] = 42;
>   assert(!__ESBMC_exists(&i, arr[i] == 42));
> 
>   return 0;
> }
> ```

Its fine. It should fail.

```
n > 0;
i >= 0;
i < n;

arr[n/2] = 42
exists i . arr[i] = 42 ---- True
```

Since you negate the exists then it becomes a false.

**Comment by lucasccordeiro** [comment]
You are correct!

**Comment by lucasccordeiro** [comment]
Does it work if you remove the negation?

**Comment by lucasccordeiro** [comment]
This test case also fails:

````C
#include <assert.h>

extern void __ESBMC_assume(_Bool);
extern _Bool __ESBMC_forall(void *, _Bool);
extern _Bool __ESBMC_exists(void *, _Bool);

int main()
{
  unsigned n;
  int arr[n];
  unsigned i;

  __ESBMC_assume(n > 1 && n < 10);
  __ESBMC_assume(i >= 0 && i < (n - 1));

  // Assume sorted order
  _ESBMC_assume(_ESBMC_forall(&i, arr[i] <= arr[i+1]));

  // Ensure last element is the largest
  __ESBMC_assert(arr[1] >= arr[0], "Array is sorted");

  return 0;
}
````

**Comment by lucasccordeiro** [comment]
This test case also fails with this PR:

````C
#include <stdio.h>
#include <string.h>

extern void __ESBMC_assume(_Bool);
extern _Bool __ESBMC_forall(void *, _Bool);
extern _Bool __ESBMC_exists(void *, _Bool);

typedef struct {
    char name[20];
    int age;
} Person;

int main() {
    Person people[] = {
        {"Alice", 25},
        {"Bob", 30},
        {"Charlie", 22}
    };
    unsigned n = sizeof(people) / sizeof(people[0]);
    unsigned i;

    // Check if a person named 'Alice' exists
    __ESBMC_assume(n > 0);
    __ESBMC_assert(__ESBMC_exists(&i, (i < n && strcmp(people[i].name, "Alice") == 0)), "Alice is missing!");

    return 0;
}
````

File: src/solvers/z3/z3_conv.cpp (hunk 3 above)
Comment (rafaelsamenezes) [reply src/solvers/z3/z3_conv.cpp]: added comment

File: src/clang-c-frontend/clang_c_language.cpp (hunk 4 above)
Comment (rafaelsamenezes) [reply src/clang-c-frontend/clang_c_language.cpp]: i adapted this example into the regression test

**Review by lucasccordeiro — APPROVED** [review]

File: src/solvers/smt/smt_conv.cpp [hunk 5]
```diff
@@ -177,6 +178,27 @@ void smt_convt::push_ctx()
...
 }
 
+/* Iterate over "body" expression looking for symbols with the same name
+   as lhs. Then, replaces it. */
+void replace_name_in_body(
+  const expr2tc &lhs,
+  const expr2tc &replacement,
+  expr2tc &body)
+{
+  // TODO: lhs and replacement should be a list of pairs to deal with multiple symbols
```
intrigus-lgtm, line 188: Has this been resolved?

src/solvers/smt/smt_conv.cpp
```diff
@@ -1231,6 +1253,58 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
     a = convert_ast(cm.side_2);
     break;
   }
+  case expr2t::forall_id:
+  case expr2t::exists_id:
+  {
+    // TODO: We should detect (and forbid) recursive calls to any quantifier (eg., forall i . forall j. i < j).
```
intrigus-lgtm, line 1259: Same here?

src/solvers/smt/smt_conv.h
```diff
@@ -330,6 +330,20 @@ class smt_convt
...
   virtual smt_astt mk_isint(smt_astt a);
+  virtual smt_astt
+  mk_quantifier(bool is_forall, std::vector<smt_astt> lhs, smt_astt rhs)
+  {
+    (void)is_forall;
+    (void)lhs;
+    (void)rhs;
+    abort();
+  }
+  virtual smt_astt mk_exists(smt_astt symbols, smt_astt body)
```
intrigus-lgtm: Did I miss where this method is used?

src/solvers/smt/smt_conv.h
```diff
@@ -330,6 +330,20 @@ class smt_convt
   virtual smt_astt mk_real2int(smt_astt a);
   virtual smt_astt mk_int2real(smt_astt a);
   virtual smt_astt mk_isint(smt_astt a);
+  virtual smt_astt
+  mk_quantifier(bool is_forall, std::vector<smt_astt> lhs, smt_astt rhs)
```
intrigus-lgtm, line 334: Would probably be cleaner to use an enum here, but potentially not worth the effort.

intrigus-lgtm (commented):
How does this interact with generating models for counter-examples?
If I remember correctly, e.g., Z3 does not generate models for variables bound in quantifiers.

Some very minor questions.

src/solvers/smt/smt_conv.cpp
rafaelsamenezes: I will leave this to another PR

src/solvers/smt/smt_conv.cpp
rafaelsamenezes: I will leave this to another PR

src/solvers/smt/smt_conv.h
rafaelsamenezes: Nice catch! Missed it while refactoring.

rafaelsamenezes:
> How does this interact with generating models for counter-examples? If I remember correctly, e.g., Z3 does not generate models for variables bound in quantifiers.
> 
> Some very minor questions.

You are right, it returns an undefined boolean for the quantifier. For now I put a warning message and defaulted the model to false.

src/solvers/smt/smt_conv.cpp
rafaelsamenezes: This code was removed

lucasccordeiro (approved):

lucasccordeiro:
There is now some user documentation about quantifiers at https://ssvlab.github.io/esbmc/documentation.html#quantifiers.

lucasccordeiro:

PR #2281, opened 2025-02-05
[C++ verification] fix to_string

### LINKED ISSUE #2280: [C++ frontend] String comparison
ESBMC incorrectly returns `VERIFICATION FAILED` for this example:

````CPP
#include <cassert>
#include <string>

int main() {
    std::string a = "60";

    // String conversion
    assert(std::to_string(60) == "60");

    return 0;
}
````

````
State 68 file hex.cpp line 8 column 5 function main thread 0
Violated property:
  file hex.cpp line 8 column 5 function main
  assertion std::to_string(60) == "60"
  return_value$_operator==$4


VERIFICATION FAILED

Bug found (k = 3)
````

### PR DESCRIPTION
**Pull request by XLiZHI**

### CODE REVIEW

**Review by lucasccordeiro — APPROVED** [review]

**Comment by lucasccordeiro** [comment]

File: src/cpp/library/string [hunk 1]
```diff
@@ -3263,66 +3264,75 @@ istream &getline(istream &is, string &str);
...
+  itoa(value, buffer, 10);
   return string(buffer);
 }
 
 std::string to_string(unsigned long long value)
 {
   char buffer[50];
-  sprintf(buffer, "%llu", value);
+  //sprintf(buffer, "%llu", value);
+  itoa(value, buffer, 10);
```
intrigus-lgtm, line 3311: That's not correct.
`itoa` only taks an `int` and not an `unsigned long long`.
E.g. something like `assert(std::to_string(9223372036854775807ull) == "9223372036854775807");` should fail.

Can you create an issue/testcase for this problem?
Also was it not possible to "[e]xtend `goto_symext::symex_printf`" as suggested in line 3263?

XLiZHI: It does miss the case.

The comment on `to_string` was created by me. Perhaps it would be more efficient to implement it in symex, but so far I haven't found a way to extend `symex_printf`, so I'm following our model of `stringstream`.  I'll check later.

PR #2285, opened 2025-02-06
[OM] Check for NULL pointers in src and dest in memcpy (DO NOT MERGE)

lucasccordeiro:
This PR (30s):

````
  correct:           18678
    correct true:    11348
    correct false:    7330
  incorrect:           103
    incorrect true:     62
    incorrect false:    41
  unknown:           14788
  Score:             27386 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13215831829
````

Master (30s):

````
  correct:           18686
    correct true:    11356
    correct false:    7330
  incorrect:            84
    incorrect true:     63
    incorrect false:    21
  unknown:           14799
  Score:             27690 (max: 55885)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/13098042643
````

### DIFF
--- a/scripts/competitions/svcomp/stats-30s.txt
+++ b/scripts/competitions/svcomp/stats-30s.txt
@@ -1,11 +1,11 @@
 Statistics:          33569 Files
-  correct:           17662
-    correct true:    10372
-    correct false:    7290
+  correct:           17705
+    correct true:    10400
+    correct false:    7305
   incorrect:            35
     incorrect true:     14
     incorrect false:    21
-  unknown:           15872
-  Score:             27250 (max: 55885)
+  unknown:           15829
+  Score:             27321 (max: 55885)
 
-https://github.com/esbmc/esbmc/actions/runs/15429068267
+https://github.com/esbmc/esbmc/actions/runs/15709379929

### CODE REVIEW

**Review by XLiZHI — APPROVED** [review]

File: regression/esbmc/memcpy2/main.c [hunk 1]
```diff
@@ -0,0 +1,9 @@
+#include <string.h>
+
+int main() {
+  char *src = "Hello";
+  char dest[5];
+  memcpy(dest, src, 5);
+  return 0;
```
intrigus-lgtm:
```suggestion
  char* result = memcpy(dest, src, 5);
  assert(result == dest && dest[4] == 'o');
  return 0;
```

intrigus-lgtm: Just so the TC actually checks a bit more.

intrigus-lgtm (approved):

intrigus-lgtm:
Btw. `memcpy(NULL, NULL, 0)` is also well-defined since a few weeks:
https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined#compiler_builtins

regression/esbmc/memcpy2/main.c
lucasccordeiro: Implemented as suggested.

intrigus-lgtm:
Why did you remove the `dest != NULL` check?

lucasccordeiro:
> Why did you remove the dest != NULL check?

Just for investigation: I want to understand which assertion leads to the incorrect false results from sv-comp.

regression/esbmc/31_cdaudio/test.desc
```diff
@@ -1,4 +1,4 @@
 THOROUGH
 main.c
 --unwind 1 --no-unwinding-assertions --no-bounds-check --no-pointer-check --no-div-by-zero-check -Wno-error=implicit-function-declaration -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION FAILED$
```
intrigus-lgtm: How is this test case even supposed to compile outside of esbmc?
E.g. the linker fails due to missing definitions:
https://godbolt.org/z/nMa56KrjM

lucasccordeiro:
I would be keen to merge this PR, but it will affect our scores in SV-COMP. We must find a way to enable these checks via the command line.

@rafaelsamenezes and @intrigus-lgtm : do you have any suggestions?

intrigus-lgtm:
> I would be keen to merge this PR, but it will affect our scores in SV-COMP. We must find a way to enable these checks via the command line.
> 
> @rafaelsamenezes and @intrigus-lgtm : do you have any suggestions?

Where can one find a list of affected test cases?

lucasccordeiro:
Thanks, @intrigus-lgtm. I have asked @rafaelsamenezes to extract this information from BenchExec. He'll post it here.

regression/esbmc/31_cdaudio/test.desc
XLiZHI: This is probably an error from the Clang linker. Clang was able to generate the AST successfully, which is what ESBMC needs.
[https://godbolt.org/z/c3W9qvaY9]()

PR #2287, opened 2025-02-07
[clang-cpp] don't eagerly parse template specializations

intrigus-lgtm:
Template specialization can be defined after the actual template. If we eagerly parse those specializations, we can get into all kind of trouble, because we might refer to types that have not yet been defined.


XLiZHI:
Does this PR solve the seg fault in the issue? I still get the same error.

intrigus-lgtm:
> Does this PR solve the seg fault in the issue? I still get the same error.

No, this PR as of now just allows us to parse the testcase, that is, we should not hit the `add_vtable_variable_symbols` assert anymore.

intrigus-lgtm:
The latest commit should fix the test errors.
I had some failing template testcases in my thesis, which I solved in another, more complicated way.
Currently verifying whether this PR would also solve them and then adding them as TCs.

intrigus-lgtm:
This is now ready for review.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1539,6 +1539,25 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+    // When parsing the template `C` we do not parse the specializations, so there
+    // is no other code that could parse the type of `C<0>`.
+    // (Yes, clang allows us to get all specializations of a template, but that leads to other problems.
+    // See #1782 or #2284)
+
+    if (const auto nns = decl.getQualifier())
+    {
+      if (const auto type = nns->getAsType())
+      {
+        assert(!nns->isDependent());
```
lucasccordeiro, line 1554: Shall we consider adding a debug log to identify cases where `isDependent()` is triggered?

lucasccordeiro (changes requested):
Minor comments only.

Shall we consider including edge cases, such as (1) template specialization with non-trivial inheritance and (2) multiple levels of template instantiation?

src/clang-c-frontend/clang_c_convert.cpp
intrigus-lgtm: This is intended to be a sanity check, I don't think having a nested name specifier that is dependent is actually possible in this overall context, but you can never be too sure.

lucasccordeiro (approved):

lucasccordeiro:
Thanks for this clarification, @intrigus-lgtm.

Let's wait for @XLiZHI and @mikhailramalho to give feedback about your PR.

intrigus-lgtm:
> Minor comments only.
> 
> Shall we consider including edge cases, such as (1) template specialization with non-trivial inheritance and (2) multiple levels of template instantiation?

I'll try to find some additional TCs for (2).
(1) might also be possible, but we could potentially run into bugs in ESBMC's modelling of inheritance.

XLiZHI (commented):
This PR looks good, just one small question.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1539,6 +1539,25 @@ bool clang_c_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+        typet ignored;
+        if (get_type(*type, ignored))
```
XLiZHI, line 1556: What does this line of code do? I understand this is a robustness check, but getting the type is confusing to me. I tried to comment it out and was still able to pass all the TCs in this PR.

intrigus-lgtm: This is needed so that gcc-templates-test/init8 (search for init8) still works.

Was this not clear from the comments above?
If so, what should I add?
Should I reference that init8 would fail otherwise?

XLiZHI: Yes, it would be nice if you could mention the corresponding test case.

intrigus-lgtm: We have to get the type so that static functions defined in templates still work.
E.g. C<0>::goo().
Because the NestedNameSpecifier (C<0>) is the only reference to C<0>.
Alternatively, we could change the code that handles calls to pull in callef functions.

The current solution is still partially wrong if I'm reading the standard correctly, because a static member function should not cause the instantiation of the whole class template. But getting this completely right is probably tricky and not needed to make progress right now :)

XLiZHI (approved):

lucasccordeiro:
Thanks fr submitting this PR, @intrigus-lgtm.

PR #2289, opened 2025-02-10
[C++ verification] fix circular type dependencies issue [do not merge]

XLiZHI:
```cpp
template <class T>
class A
{
  void foo(T t)
  {
  }
};

template <class d>
class g
{
  A<d> f;
};

class i
{
  g<i> h;
};

int main()
{
}
```

tag-A<i> --> tag-i
tag-i --> tag-g<i>
tag-g<i> --> tag-A<i>

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/template_dep/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/template_dep/main.cpp
@@ -1,0 +1,22 @@
+template <class T>
+class A
+{
+  void foo(T t)
+  {
+  }
+};
+
+template <class d>
+class g
+{
+  A<d> f;
+};
+
+class i
+{
+  g<i> h;
+};
+
+int main()
+{
+}
--- a/regression/esbmc-cpp/bug_fixes/template_dep/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/template_dep/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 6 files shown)


lucasccordeiro (approved):
@XLiZHI: could you also add a regression test that produces "VERIFICATION FAILED"?

lucasccordeiro:
@XLiZHI: BTW, shall we maintain a visited set or track dependency resolution status? For example, could we detect deeper circular dependencies (e.g., A → B → C → A) to (infinite) recursion?

intrigus-lgtm (changes requested):
Please give me some time to test this.

XLiZHI:
> @XLiZHI: BTW, shall we maintain a visited set or track dependency resolution status? For example, could we detect deeper circular dependencies (e.g., A → B → C → A) to (infinite) recursion?

It's a very good proposal. I'll check it out.

XLiZHI:
I found an issue that symbolic type is passed to symex.

```cpp
int main()
{
 i obj;
}
```

It's strange that for some types of circular dependencies, renaming just once doesn't seem to handle all of them (It needs to be renamed again). Does anyone have a similar problem? 

https://github.com/esbmc/esbmc/blob/23d6678c83ebfa942d672fdc090742378454fd68/src/goto-programs/goto_convert_functions.cpp#L402-L404

lucasccordeiro:
This code https://github.com/esbmc/esbmc/blame/23d6678c83ebfa942d672fdc090742378454fd68/src/goto-programs/goto_convert_functions.cpp#L402-L404 was written by @jmorse 10 years ago :-)

PR #2290, opened 2025-02-12
[python] try harder to find a python binary

intrigus-lgtm:
Some computers might only have `python3` installed so also search for it.
Previously, we would crash with a `permission denied` error, because `python` could not be executed (only `python3` exists).

Fix this so that we provide a helpful message to the user.


sshivaji:
Thanks for fixing this, I noticed this error many times before.

brcfarias (approved):
LGTM

lucasccordeiro (approved):

lucasccordeiro:

PR #2293, opened 2025-02-12
added basic support for half floats and complex types

rafaelsamenezes:
This is an going milestone to support immitrin headers.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -1421,7 +1421,6 @@ bool clang_c_convertert::get_builtin_type(
 
   case clang::BuiltinType::Float16:
   case clang::BuiltinType::Half:
-  case clang::BuiltinType::Float16:
   case clang::BuiltinType::BFloat16:
     new_type = half_float_type();
     c_type = "_Float16";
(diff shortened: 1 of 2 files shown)


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1403,6 +1421,7 @@ bool clang_c_convertert::get_builtin_type(
 
   case clang::BuiltinType::Float16:
   case clang::BuiltinType::Half:
+  case clang::BuiltinType::BFloat16:
```
intrigus-lgtm, line 1424: Adding this case here is likely just wrong as far as I know, because afaik the formats of bfloat16 and half differ:
https://en.wikipedia.org/wiki/Bfloat16_floating-point_format#bfloat16_floating-point_format

pmatos:
All four changes in this `builtin-hack` draft are now implemented properly on `master`:

- `_Complex` support via #4133 (a real `complex_typet` plus arithmetic, `__real__`/`__imag__`, `<complex.h>` models, and regression tests) — rather than the 2-element-vector mapping here;
- the missing-return-value handling via #3659 (warn + nondet substitution, instead of the `#if 0`'d `abort()`);
- `__builtin_bit_cast` via #4411 (a real byte-level reinterpret, instead of the `gen_zero` stub).

The BFloat16 → `half_float_type` mapping (the one change the reviewer flagged as likely incorrect) is now gated behind `--dont-care-about-missing-extensions`, with the default path erroring out.

Closing as superseded.

PR #2294, opened 2025-02-13
[python] preprocessor extension for function definitions

Ben-Eichhoefer:
Previously default arguments and keyword arguments would not be correctly translated into the lower level IRs resulting in a failed verification despite being a correct program
Approach currently only works for functions that are defined within the module being verified

During preprocessing explicitly add defaults & keyword arguments to function call as if they positional arguments


intrigus-lgtm:
Does this work with code like this?
Per https://codeql.github.com/codeql-query-help/python/py-modification-of-default-value/:
> **The default value of a parameter is computed once when the function is created, not for every invocation.** The “pre-computed” value is then used for every subsequent call to the function. Consequently, if you modify the default value for a parameter this “modified” default value is used for the parameter in future calls to the function. This means that the function may not behave as expected in future calls and also makes the function more difficult to understand.

```python
def mutate_list(elem:str, elems : list[str] = []):
   elems.append(elem)
   return elems

assert(len(mutate_list("a")) == 1)
assert(len(mutate_list("a")) == 2)
assert(len(mutate_list("a")) == 3)
assert(len(mutate_list("a")) == 4)
```

Ben-Eichhoefer:
> Does this work with code like this? Per https://codeql.github.com/codeql-query-help/python/py-modification-of-default-value/:

Thanks for pointing that out, it currently does not.
I will look to improve on that.

regression/python/function-keyword-argument/main.py
```diff
@@ -0,0 +1,5 @@
+def return_sum(x:int,y:int) -> int:
```
brcfarias: Fix function name, please

src/python-frontend/preprocessor.py
```diff
@@ -126,5 +128,52 @@ def visit_Call(self, node):
                 node.args[1] = ast.NameConstant(value=True)
             else:
                 node.args[1] = ast.NameConstant(value=False)
+
+        # if not a function or preprocessor doesn't have function definition return
+        if isinstance(node.func,ast.Attribute) or node.func.id not in self.functionParams:
```
brcfarias: Could `node.func` be something other than `ast.Attribute` here?
Is there an expected type? If so, would it be better to use `if not isinstance(node.func, <expected type>)`?

src/python-frontend/preprocessor.py
```diff
@@ -126,5 +128,52 @@ def visit_Call(self, node):
...
+            return node
+
+        # append defaults
+        for i in range(len(node.args),len(expectedArgs)):
+            if expectedArgs[i] in keywords:
+                node.args.append(keywords[expectedArgs[i]])
+            elif (functionName, expectedArgs[i]) in self.functionDefaults:
+                node.args.append(ast.Constant(value = self.functionDefaults[(functionName, expectedArgs[i])]))
+            else:
+                # print("File ", self.module_name, "line", node.lineno)
```
brcfarias: Could add a warning message here and remove the commented lines?

brcfarias:
Feel free to address only the comments for now, and then cover the remaining cases in other PRs incrementally.

src/python-frontend/preprocessor.py
Ben-Eichhoefer: Just to clarify, do you mean something along the lines of:
```python
else:
    print(f"Warning - error in file {self.module_name} line {node.lineno}, {functionName}() missing required positional argument: '{expectedArgs[i]}'")
    break
```

brcfarias: Exactly.

lucasccordeiro (approved):

lucasccordeiro:
> Does this work with code like this?
Per https://codeql.github.com/codeql-query-help/python/py-modification-of-default-value/:

@Ben-Eichhoefer: Can you please submit another PR to address @intrigus-lgtm's comments?

brcfarias (approved):

lucasccordeiro:

PR #2297, opened 2025-02-15
[dereference] fix array dereferencing for struct

### LINKED ISSUE #2292: [Pointer Analysis] Improve the handling of function pointers
ESBMC incorrectly produces `VERIFICATION FAILED` for this program.

I suspect that the bug lies in our value set analysis.

$esbmc main.c

````C
#include <assert.h>
#include <stdbool.h>
typedef struct {
  bool (*a)();
  void (*act)();
} j;
j f;
bool b() { return 0; }
void d() { assert(0); }
bool c() { return 1; }
void e() {};
int main() {
  j actions[2];
  actions[0] = (j){b, d};
  actions[1] = (j){c, e};
  j *g = actions;
  {
    j *actions = g;
    for (int i = 0; i < 2; i++)
      if (actions[i].a())
        f = g[i];
  }
  f.act();
  return 0;
}
````

lucasccordeiro:
Fixes #2292.

This patch fixes the array dereferencing mechanism by correctly handling structure-type elements. If an array contains structures, the dereferencing process now correctly accesses the indexed structure and resolves nested fields. Added `assert(is_scalar_type(arr_subtype))` to prevent unexpected behaviors when handling non-structure and non-scalar types.

SV-COMP run (30s): https://github.com/esbmc/esbmc/actions/runs/13348717496


lucasccordeiro:
This PR (30s):

````
  correct:           18695
    correct true:    11365
    correct false:    7330
  incorrect:            83
    incorrect true:     62
    incorrect false:    21
  unknown:           14791
  Score:             27740 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13348717496
````

Master (30s):

````
  correct:           18686
    correct true:    11356
    correct false:    7330
  incorrect:            84
    incorrect true:     63
    incorrect false:    21
  unknown:           14799
  Score:             27690 (max: 55885)

GitHub actions: https://github.com/esbmc/esbmc/actions/runs/13098042643
````

PR #2299, opened 2025-02-16
[python] support for random.random() and random.randint()

lucasccordeiro:
This PR introduces an operational model to support `random.random()` and `random.randint()` in the `random` module.


src/python-frontend/models/random.py
```diff
@@ -0,0 +1,12 @@
+# Stubs for random module
+class random:
+    @classmethod
+    def randint(a:int, b:int) -> int:
+        value:int = nondet_int()
+        __ESBMC_assume(value >= 0 and value <= b)
```
brcfarias: __ESBMC_assume(value >= a and value <= b)

brcfarias (changes requested):

src/python-frontend/models/random.py
```diff
@@ -0,0 +1,12 @@
...
+        return value  # Ensures value is within [a, b]
+    @classmethod
+    def random() -> float:
+        value:float = nondet_float()
+        __ESBMC_assume(value >= 0.0 and value <= 1.0)
```
brcfarias: https://docs.python.org/3/library/random.html#random.random
According to the documentation, the value must be less than 1.0.
`__ESBMC_assume(value >= 0.0 and value < 1.0)`

brcfarias (changes requested):

brcfarias:
Perhaps the description could be adjusted to indicate that support for `random.random()` and `random.randint()` is being added?

src/python-frontend/models/random.py
lucasccordeiro: Implemented as suggested.

src/python-frontend/models/random.py
lucasccordeiro: Implemented as suggested.

lucasccordeiro:
> Perhaps the description could be adjusted to indicate that support for random.random() and random.randint() is being added?

Implemented as suggested.

PR #2304, opened 2025-02-18
Feature/numpy extended support to lib numpy

sergillam:
This PR enhances numpy support by ensuring compatibility with mixed integer and double operations. It also fixes test cases and adds support for negative numbers in create_binary_op.

Key changes:
1. Fixed handling of mixed integer/double operations.
2. Updated test cases for binary functions in the numpy module.
3. Added support for negative numbers in create_binary_op.


src/python-frontend/numpy_call_expr.cpp
```diff
@@ -2,8 +2,14 @@
 #include <python-frontend/symbol_id.h>
 #include <python-frontend/python_converter.h>
 #include <util/expr.h>
+<<<<<<< HEAD
```
lucasccordeiro: Can you fix this?

src/python-frontend/numpy_call_expr.cpp
```diff
@@ -183,8 +233,37 @@ exprt numpy_call_expr::get() const
...
+    auto it = numpy_functions.find(function);
+    if (it != numpy_functions.end())
+    {
+        auto list = create_list(call_["args"][0]["value"].get<int>(), it->second);
+        return converter_.get_expr(list);
+    }
+
+    if (is_math_function())
+    {
+        auto lhs = extract_value(call_["args"][0]);
```
lucasccordeiro: Shall we consider a boundary check here?

````CPP
if (call_["args"].size() < 2) {
    throw std::runtime_error("Insufficient arguments for math function.");
}
````

File: src/python-frontend/numpy_call_expr.cpp [hunk 3]
```diff
@@ -183,8 +233,37 @@ exprt numpy_call_expr::get() const
...
+        auto rhs = extract_value(call_["args"][1]);
+
+        auto result = std::visit(
+            [&](auto l, auto r) -> nlohmann::json {
+                using LType = decltype(l);
+                using RType = decltype(r);
+
+                if constexpr (std::is_same_v<LType, int> && std::is_same_v<RType, int>)
```
lucasccordeiro: Currently, the code assumes only int and double types.

Shall we consider other types (e.g., float, long)?

src/python-frontend/numpy_call_expr.cpp
```diff
@@ -183,8 +233,37 @@ exprt numpy_call_expr::get() const
...
+                {
+                    return create_binary_op(function, l, r);
+                }
+
+                auto left = std::holds_alternative<int>(lhs) ? static_cast<double>(std::get<int>(lhs)) : std::get<double>(lhs);
```
lucasccordeiro: Can you add some comments here?

lucasccordeiro (changes requested):

src/python-frontend/numpy_call_expr.cpp
brcfarias: Since Python only has 'int' to represent integer numbers (without size variations) and we are currently limiting it to 64 bits, you can always read from JSON using `long long.`

src/python-frontend/numpy_call_expr.cpp
sergillam: Fixed in New Commit

src/python-frontend/numpy_call_expr.cpp
sergillam: I considered a verification of size because we currently use the binary operation template

src/python-frontend/numpy_call_expr.cpp
sergillam: ok

src/python-frontend/numpy_call_expr.cpp
sergillam: It is not necessary because this stretch of code has been removed

lucasccordeiro (approved):
Thanks for addressing the comments, @sergillam.

regression/numpy/power/main.py
```diff
@@ -1,7 +1,10 @@
 import numpy as np
 
-x = np.power(2, 3)
-assert x == 8
+result1 = np.power(2, 3)
+assert result1 == 8
 
-y = np.power(2,7, dtype=np.int8)
-assert y == -128
```
brcfarias, line 7: Please keep this case.

regression/numpy/add/main.py
```diff
@@ -1,10 +1,55 @@
...
+result_1 = np.add(1.0, 4.0)
+assert result_1 == 5.0
 
-y = np.add(1, 2)
-assert y == 3
+result_2 = np.add(1, 2)
+assert result_2 == 3
 
-z = np.add(127, 1, dtype=np.int8)
-assert z == -128
```
brcfarias: Please keep this case.

brcfarias (changes requested):

regression/numpy/power/main.py
sergillam: ok, added this case

regression/numpy/add/main.py
sergillam: ok, added this case

src/python-frontend/numpy_call_expr.cpp
```diff
@@ -137,49 +175,60 @@ exprt numpy_call_expr::get() const
...
-        if (value_size > dtype_size)
+        if (dtype.find("int") != std::string::npos)
         {
-          log_warning(
-            "{}:{}: Integer overflow detected in {}() call. Consider using a "
-            "larger integer type.",
-            converter_.current_python_file,
-            call_["end_lineno"].get<int>(),
-            function_id_.get_function());
-        }
```
brcfarias, line 639: I would like to keep this warning, as we are using it in demonstrations—for example, when detecting an overflow, as in the case I asked you to re-enable.

brcfarias (changes requested):

src/python-frontend/numpy_call_expr.cpp
sergillam: ok, added warning message

lucasccordeiro:
@sergillam: Can I ask you for the status of this PR?

PR #2308, opened 2025-02-20
[python-frontend] Add return type annotations and override option

brcfarias:
- Adding return type annotations to functions without them. For example:
```
def foo():
    return 1
```
Becomes:
```
def foo() -> int:
    return 1
```

- It also introduces the `--override-return-annotation` flag, which replaces existing return annotations with the inferred type. For example:

```
def foo() -> bool:
    return 1.2
```
Becomes:
```
def foo() -> float:
    return 1.2

```

### DIFF
--- a/regression/python/func-no-params-types/main.py
+++ b/regression/python/func-no-params-types/main.py
@@ -1,0 +1,2 @@
+def foo(a, b):
+    pass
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2313, opened 2025-02-25
[Need help][C verification] add new builtin function for memcpy

XLiZHI:
This is the feasible way I can come up with.

I used _extract_ and _concat_ to simulate the behavior of _memcpy_ to avoid the for loop. However, there are a few points that I think are important. First, bound checking requires additional implementation, and second, even if we disable bound checking, once out of bound, it would violate the assertion of the solver (In the past is the verification successful). 

I'm not sure if this implementation is necessary, maybe there is a better one?

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -2873,7 +2873,6 @@ void clang_c_convertert::rewrite_builtin_ref(
 {
   static const std::list<std::string> builtins_to_rewrite = {
     "__builtin_malloc",
-    "__builtin_memcpy",
     "__builtin_memmove",
     "__builtin_strcpy",
     "__builtin_free",
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -1608,6 +1608,45 @@ bool goto_symext::run_builtin(
     return true;
   }
 
+  if (has_prefix(symname, "c:@F@__builtin_memcpy"))
+  {
+    expr2tc tgtptr = func_call.operands[0];
+    expr2tc srcptr = func_call.operands[1];
+    expr2tc size = func_call.operands[2];
+    expr2tc deref;
+
+    internal_deref_items.clear();
+    deref = dereference2tc(get_empty_type(), tgtptr);
+    dereference(deref, dereferencet::INTERNAL);
+    assert(internal_deref_items.size() == 1);
+    expr2tc tgt = internal_deref_items.front().object;
+
+    internal_deref_items.clear();
+    deref = dereference2tc(get_empty_type(), srcptr);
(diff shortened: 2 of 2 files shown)


rafaelsamenezes:
You should keep the original memcpy as well, there are cases were optimizing it is impossible. For example, for memset we do not optimize when the length is symbolic. I suggest you:

1. Find a case where is simple to optimize. For example: src, dst and length are all known and not symbolic (or multiple objects). All other cases you just bump into the original implementation (see memset optimization).
2. Reconstruct the dst object with the correct expression. This can get tricky depending on the type.
3. Only then start trying to generalizing more (symbolic srcs, multiple objects). 

My suggestion is to focus on 1 and 2 with some basic use cases such as copying a constant string from a buffer into another.

XLiZHI:
Thanks @rafaelsamenezes , I think it's going to get complicated.

> Find a case where is simple to optimize. For example: src, dst and length are all known and not symbolic (or multiple objects). All other cases you just bump into the original implementation (see memset optimization).‘

This is a good starting point, but I wonder if it is necessary to optimize for these simple examples, as far as I know, a known length does not cause performance issues.

> Reconstruct the dst object with the correct expression. This can get tricky depending on the type.

I'm a little confused about this. What is the purpose of reconstructing dst? Avoid generating complex formulas?

rafaelsamenezes:
> This is a good starting point, but I wonder if it is necessary to optimize for these simple examples, as far as I know, a known length does not cause performance issues.

Depends what you mean by performance, it decreases the bound needed to verify the program. I implemented the memset one due to a use case where a user was trying to initialize an array of a few kilobytes (10³ - 10⁴) while his harness did not have loops.

By optimizing the memset out, we could increase the performance just by finding the proof at a lower bound.

> I'm a little confused about this. What is the purpose of reconstructing dst? Avoid generating complex formulas?

- Lowering the bound required to construct it through BMC
- Make use of our simplifier and propagations.

Avoid generating complex formulas is also nice, it may help the solvers.

intrigus-lgtm:
I think the most important step should be to find testcases where there are performance problems.
Otherwise, for any proposed implementation we have to guess whether it improved anything.

lucasccordeiro:
Here is an example. @rafaelsamenezes' `memset` implementation deals efficiently with the memory set, but `memcpy` takes ages unfolding the loop:

````C
#include <stdio.h>
#include <string.h>
#include <assert.h>

#define SIZE 100000

void test_memcpy()
{
    char src[SIZE];
    char dest[SIZE];  // Ensure destination buffer is large enough
    memset(dest, 1, sizeof(dest));  // Initialize with zeros

    // Use memcpy to copy data
    memcpy(dest, src, SIZE);  // +1 to include null terminator

    // Check if the copied data matches the source
    //assert(strcmp(src, dest) == 1);
}

int main()
{
    test_memcpy();
    return 0;
}
````

Issue #2312, opened 2025-02-25
ESBMC assumption on argv


tihanyin:
ESBMC assumes that the `argv` index can be arbitrarily out of bounds, such as accessing `argv[3823823]`, regardless of the actual number of arguments.

The provided code simply prints the user's name to the output. However, ESBMC reports **VERIFICATION FAILED** due to a **dereference failure: NULL pointer**. 

What is the solution to achieve **VERIFICATION SUCCESS**? It appears that in every scenario, the result is **VERIFICATION FAILED**. 

PoC small program:
```
#include <stdio.h>
#include <string.h>

int main(int argc, char *argv[]) {
    char name[20]="";

    if (argc > 1 && argv[1] != NULL) {  
        strncpy(name, argv[1], 10);
    }

    printf("Hello, %s!\n", name);
    return 0;
}
```

Yiannis128:
Is this similar to what's reported in: https://github.com/esbmc/esbmc/issues/2244#issuecomment-2639922397

rafaelsamenezes:
After chatting a bit of Lucas, maybe we could implement something like this in GOTO.

```c
#define MAX 255

int main(int argc, char *argv[]) {
  // Kinda what esbmc does at clang_c_frontend
  __ESBMC_assume(argc >= 1 && argc <= MAX);
  argv = malloc(sizeof(char*)*(argc+1));
  argv[argc] = NULL;

  // What we could do at GOTO
  for(int i = 0; i < argc; i++ )
    {
      unsigned nondet;
      __ESBMC_assume(nondet > 0);
      argv[i] = malloc(sizeof(char)*nondet);
    }
}
``

ChenfengWei0:
> After chatting a bit of Lucas, maybe we could implement something like this in GOTO.
> 
> #define MAX 255
> 
> int main(int argc, char *argv[]) {
>   // Kinda what esbmc does at clang_c_frontend
>   __ESBMC_assume(argc >= 1 && argc <= MAX);
>   argv = malloc(sizeof(char*)*(argc+1));
>   argv[argc] = NULL;
> 
>   // What we could do at GOTO
>   for(int i = 0; i < argc; i++ )
>     {
>       unsigned nondet;
>       __ESBMC_assume(nondet > 0);
>       argv[i] = malloc(sizeof(char)*nondet);
>     }
> }
> ``

We definitely could. But wouldn’t this involve’s a lot of work, if we directly use Goto the write this algo?

PR #2106, opened 2024-10-30
[coverage] support coverage in function mode

ChenfengWei0:
```c
int foo(int a, int b)
{
  if(a>0)
    printf("a>0\n");
  else if(b == 2)
    printf("b == 2\n");
  return a + b;
}

int main()
{
  int x = nondet_int();
  int y = nondet_int();
  assert(foo(x,y) == 10);
  return 0;
}
```

`--condition-coverage --function foo `

```
[Coverage]

Reached Conditions:  4
Short Circuited Conditions:  0
Total Conditions:  4

Condition Properties - SATISFIED:  4
Condition Properties - UNSATISFIED:  0

Condition Coverage: 100%

VERIFICATION FAILED
```

PR #2314, opened 2025-02-26
Support argc&argv in function mode

ChenfengWei0:


ChenfengWei0:
The CLI failure seems to be due to a conflict related to the usage of `malloc` in `coverage.c`. If I remove it, the issue is resolved. I don't know how to fix it.

lucasccordeiro:
@ChenfengWei0: I don't understand the issue yet.

Let's consider this regression test that is failing with your PR:

````C
#include <assert.h>
#include <errno.h>
#include <pthread.h>
#include <stdlib.h>

void *worker(void *k)
{
  pthread_key_t *key = k;

  long val = (long)pthread_getspecific(*key);
  assert(val == 0);

  int r = pthread_setspecific(*key, (void *)42);
  if(r == ENOMEM)
  {
    exit(1);
  }
  assert(r == 0);

  val = (long)pthread_getspecific(*key);
  assert(val == 42);

  return NULL;
}

int main()
{
  pthread_key_t key;
  int r = pthread_key_create(&key, NULL);

  if(r == ENOMEM)
  {
    exit(1);
  }
  assert(r == 0);

  pthread_t tid;

  long val = (long)pthread_getspecific(key);
  assert(val == 0);

  pthread_create(&tid, NULL, worker, &key);

  val = (long)pthread_getspecific(key);
  assert(val == 0);

  r = pthread_setspecific(key, (void *)16);
  if(r == ENOMEM)
  {
    exit(1);
  }
  assert(r == 0);

  val = (long)pthread_getspecific(key);
  assert(val == 16);

  pthread_join(tid, NULL);
  val = (long)pthread_getspecific(key);
  assert(val == 16);

  return 0;
}
````

Here is the command line described in the test.desc:

````
$ esbmc main.c --force-malloc-success --unwind 3 --context-bound 2
````

What is the issue of your PR in this regression test?

**Comment by ChenfengWei0** [comment]
This PR:
```
...
          * width : 8
    * type : unsignedbv
      * width : 64
* type : unsignedbv
  * width : 64
terminate called after throwing an instance of 'vsa_not_implemented_exception'
[1]    27790 IOT instruction (core dumped)  ./esbmc 1.c --force-malloc-success
```

This PR without the `coverage.c`
```
VERIFICATION SUCCESSFUL
```

This PR with the function body commented  in `coverage.c`, that is
```c
#include <stdio.h>
#include <stdlib.h>

// for argc & argv
__attribute__((annotate("__ESBMC_inf_size"))) char __ESBMC_inf_str[1];
char **ESBMC_malloc_argv(int argc)
{
// __ESBMC_HIDE:;
//   char **argv_copy = (char **)malloc((argc + 1) * sizeof(char *));
//   for (int i = 0; i < argc; i++)
//   {
//     argv_copy[i] = __ESBMC_inf_str;
//   }
//   argv_copy[argc] = NULL;

//   return argv_copy;
}
```
Result:
```
VERIFICATION SUCCESSFUL
```

rafaelsamenezes:
> This PR:
> 
> ```
> ...
>           * width : 8
>     * type : unsignedbv
>       * width : 64
> * type : unsignedbv
>   * width : 64
> terminate called after throwing an instance of 'vsa_not_implemented_exception'
> [1]    27790 IOT instruction (core dumped)  ./esbmc 1.c --force-malloc-success
> ```

Could you dump the full expression and attach the log?

ChenfengWei0:
[1.log](https://github.com/user-attachments/files/18989414/1.log)

intrigus-lgtm (changes requested):
This does not solve #2312.
Afaik the issue is unrelated to function mode so I don't quite understand why function mode is extended instead of the more general case.

Additionally it doesn't work unless unwinding assertions are disabled which can hide other bugs.

E.g.
```cpp
#include <stdio.h>
#include <string.h>

int main(int argc, char *argv[])
{
    char name[20] = "";

    __ESBMC_assume(argc == 10);
    __ESBMC_assert(0, "0");
}
```
"works" flawlessly, but fails (correctly!) if `argc == 9` is used instead.

ChenfengWei0:
> This does not solve https://github.com/esbmc/esbmc/issues/2312.

This is not a direct fix for #2312. As stated in the title, it addresses the function mode, which is a refinement of #2124 and #2106.  

> "works" flawlessly, but fails (correctly!) if argc == 9 is used instead.

E.g. `--force-malloc-success --k-induction --function main`

Result of `==10`
```
[Counterexample]


State 3 file 1.c line 4 column 1 thread 0
  argc = 10 (00000000 00000000 00000000 00001010)

State 29 file 1.c line 9 column 5 function main thread 0
Violated property:
  file 1.c line 9 column 5 function main
  0
  0


VERIFICATION FAILED

Bug found (k = 11)
```
> it doesn't work unless unwinding assertions are disabled which can hide other bugs.

Not sure if this is relevant to this PR. 

I mean, I would be more than happy to use `k-induction` if it would terminate, as it currently seems to be unstoppable for the regression cases.

intrigus-lgtm:
I'm not a fan of special casing `main` in more places and essentially copy-pasting the code.
Why can't a user simply run this without `--function main` if they want the special handling for `main`?

------------

In my humble opinion, it would be best to model argv differently.
Right now, we just model argv as a big dynamic allocation that has enough space for `(argc + 1) * sizeof char*`, but we never assign a pointer to any argv[i] so it just points anywhere.
I don't see a way how to initialize every element of argv[i] properly if we want to keep supporting arbitrarily big `argc` values, because we would afaik always run into unwinding limits.

So, we can either have arbitrarily big `argc` values or properly initialized argv where argv[i] is not NULL for `i < argc` and where argv[i] points to a `char *` that resembles a valid string, i.e., is null-terminated.

I'd be more than happy to see a solution that combines the best of both worlds, but until then I'd think that doing it like cbmc does would be better:
https://github.com/diffblue/cbmc/blob/159af34d61a15dc42dd3d7593d298cbebd7e9e9c/src/goto-instrument/model_argc_argv.cpp#L39

They introduced a flag which specifies how many elements `argc` of `argv` to model.
This would not run into any problems with unwindings, because the flag is a constant value for esbmc.

PR #2316, opened 2025-02-26
[C++ verification] fix abs method

XLiZHI:


src/cpp/library/cmath
```diff
@@ -12,10 +12,15 @@
 /* delegates to math.h in the C++ headers */
 #include <math.h>
 #include <type_traits>
+#include <stdlib.h>
```
lucasccordeiro, line 15: Why do we need to include stdlib.h?

lucasccordeiro (changes requested):

src/cpp/library/cmath
XLiZHI: Because in C, `int abs(int n);` is defined in this header file.

https://en.cppreference.com/w/c/numeric/math/abs

lucasccordeiro (approved):

intrigus-lgtm (approved):

lucasccordeiro:

Issue #1424, opened 2023-10-26
Warn about undefined external symbols just like for functions?


fbrausse:
> I've double-checked an indeed forgot to include some of MUSL's constants in the runs.

_Originally posted by @fbrausse in https://github.com/esbmc/esbmc/issues/818#issuecomment-1780626386_

This was about a symbol defined `extern` and thus silently treated as nondet by ESBMC. When functions are being called that have no body available a message is printed. In contrast for accesses to such external symbols no information is given in the log.

Is this intentional? It would have been easier for me to run the correct command line if this information had been there.

rafaelsamenezes:
> Is this intentional?

I guess, we are just no explicit about it. I see no problem in advise the user that some functions are not defined.

fbrausse:
> I guess, we are just no explicit about it.

Indeed, that's what this question is about: can we be explicit in the log about usage of external symbols? I guess the linker might be able to give an overapproximation here.

PR #2322, opened 2025-02-27
Fix #2320 by ignoring `#pragma comment(linker, ...)` directives DO NOT MERGE.

### LINKED ISSUE #2320: Partially support `#pragma comment` due to usage in Windows headers
The TCs`regression/esbmc/github_1165/wchar.c` and `regression/llvm/char_literal` fail with:
```
ERROR: unrecognized / unimplemented clang declaration PragmaComment
PragmaCommentDecl 0x203adfb0508 <C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\wchar.h:213:13> col:13 linker "/alternatename:_Avx2WmemEnabled=_Avx2WmemEnabledWeakValue"

ERROR: CONVERSION ERROR
```

This is caused by a recent change in the header.

intrigus-lgtm:


lucasccordeiro:
@intrigus-lgtm: Can I ask you for the status of this PR?

lucasccordeiro:
@intrigus-lgtm: Do you still intend to work on this PR?

intrigus-lgtm:
> @intrigus-lgtm: Do you still intend to work on this PR?

Not that much.
I think the problem I hit when trying to workaround the changes is that, we would still pull in the whole header related to avx and vector stuff, which would then break parsing with ESBMC.
Or maybe `defined(__AVX2__)` would always be enabled, which I assumed to be false?
Don't remember anymore, just that fixing it properly would require a lot more time than I was willing to spend.

intrigus-lgtm:
Let's see what CI says, because the logs already expired...

pmatos:
Thanks for digging into this and for the clear write-up.

Since this was marked "DO NOT MERGE" and you've indicated you don't intend to take it further, I'm closing the PR. The underlying problem — the `#pragma comment(linker, ...)` conversion error triggered by current Windows `ucrt` headers — is still real and remains tracked by the open #2320 (with #1424 as a related dependency), so nothing is lost by closing this here. Thanks again for the investigation and the reproducer.

PR #2327, opened 2025-03-02
[Solidity] Bounded Cross-Contract Verification Algorithm

ChenfengWei0:
This PR implements the Bounded Cross-Contract Verification (BCCV), which allows the verifier to model the external calls precisely. The corresponding option is `--bound`.

In contrast, the arbitrary modelling can be enabled via option `--unbound`

---

This PR also includes some bug fix, and adds the option `assign-param-nondet` to control the assign_param_nondet algorithm in the function mode (was enabled by default).


lucasccordeiro:
@ChenfengWei0: Can I ask you for the status of this PR?

ChenfengWei0:
Some bugs remain, but that’s for another PR.

lucasccordeiro:
@ChenfengWei0: could you please check our CI?

lucasccordeiro:
@ChenfengWei0: could you please add a description for this PR?

ChenfengWei0:
> @ChenfengWei0: could you please add a description for this PR?

Done

src/esbmc/options.cpp
```diff
@@ -88,7 +88,14 @@ const struct group_opt_templ all_cmd_options[] = {
     {"no-visibility",
      NULL,
      "force to verify every function, even it's an unreachable "
-     "internal/private function"}}},
+     "internal/private function"},
```
lucasccordeiro, line 91: @can you move these options under the solidity frontend?

src/esbmc/options.cpp
```diff
@@ -548,6 +555,10 @@ const struct group_opt_templ all_cmd_options[] = {
      {"branch-function-coverage-claims",
       NULL,
       "enable branch-coverage-ext and shows all reached claims"},
+     {"assign-param-nondet",
```
lucasccordeiro, line 558: Isn't it already the case?

src/goto-programs/goto_coverage.cpp
```diff
@@ -819,8 +827,39 @@ bool goto_coveraget::is_target_func(const irep_idt &f) const
...
   exprt symbol = symbol_expr(*ns.lookup(f));
-  if (symbol.name().as_string() != target_function)
+  if (symbol.name().as_string() != tgt_name)
     return false;
 
   return true;
 }
+
+// negate the condition inside the assertion
+void goto_coveraget::negating_asserts(const std::string &tgt_fname)
```
lucasccordeiro, line 838: Why do we need this? can you add some comments here?

lucasccordeiro (changes requested):

src/esbmc/options.cpp
ChenfengWei0: I feel like we should have an option to control (currently this algorithm is enabled by default when using `--function`). 

The main reason is that this `assign-param-nondet` would cause some trouble in solidity, particularly for Object-Type parameters. Given that solidity frontend now based on the cpp-frontend, I would assume this might also affect c++.

src/goto-programs/goto_coverage.cpp
ChenfengWei0: Basically it's added only because we used this in the submitted paper. It's more like a debug option. 

The idea is that, if the **claim** is verified safe (unbounded), and its **negated** **claim** is also verified safe, then we say this claim is unreachable.

ChenfengWei0: Added comment as suggested

src/esbmc/options.cpp
ChenfengWei0: Just checked.. It seems to be already under solidity option

image

lucasccordeiro (approved):

lucasccordeiro:

PR #2328, opened 2025-03-02
[k-induction] Improve k-induction assertion->assumption transformation

### LINKED ISSUE #1508: Wrong veredict for float memset in k-induction

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1508.

Our transformation replaces `assert(claim_expr)` with `assume(claim_expr)`, but only if it's not the last loop iteration (unwind < max_unwind - 1).  Although this is a common technique in k-induction to strengthen the induction hypothesis, if the assumption is incorrect or too weak, the proof might be unsound (i.e., miss counterexamples).

In this PR, we do the following: 

1. Assertion Check: First, an assertion (e.g., a property that must hold) is checked.
2. Assumption Addition: If the assertion holds, it is added as an assumption (constraint) for subsequent checks, thus improving our incremental verification via lemma learning.

Master (30s):

````
  correct:           18673
    correct true:    11343
    correct false:    7330
  incorrect:            84
    incorrect true:     63
    incorrect false:    21
  unknown:           14812
  Score:             27664 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13543395439/
````

This PR (30s):

````
  correct:           18074
    correct true:    10747
    correct false:    7327
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15455
  Score:             27973 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13638155792
````


mikhailramalho (commented):
This should probably only be enabled if you are running `--inductive-step`. By using `--k-induction`, the base case already checked that the assertion holds

intrigus-lgtm:
Potentially remove the hack from #1509 ?

lucasccordeiro:
> Potentially remove the hack from https://github.com/esbmc/esbmc/pull/1509 ?

Thanks for pointing this out, @intrigus-lgtm. I have opened this issue: https://github.com/esbmc/esbmc/issues/2331 for @rafaelsamenezes.

PR #2329, opened 2025-03-03
[C verification] add new option to check unknown method arguments

XLiZHI:


intrigus-lgtm:
What's the use case for this?
Why doesn't this handle void* pointers?

Unrelated to this PR, but I think that ESBMC should hard-fail when encountering a unknown function.

XLiZHI:
> What's the use case for this? Why doesn't this handle void* pointers?

This is a new option developed for unknown function verification. For example, in the CHERI verification, some function internals are not known and we only need to consider whether the passed arguments will access illegal memory.

The new method adds support for the void type.

XLiZHI:
It seems that this failed TC has nothing to do with this PR.

lucasccordeiro:
Try to increase `--k-step` to 80.

intrigus-lgtm:
> > What's the use case for this? Why doesn't this handle void* pointers?
> 
> This is a new option developed for unknown function verification. For example, in the CHERI verification, some function internals are not known and we only need to consider whether the passed arguments will access illegal memory.
> 
> The new method adds support for the void type.

But what if you have a pointer to a struct and the struct itself contains pointers?
Then the function could still access illegal memory.

Are there so many functions that simply "implementing" them and dereferencing the arguments yourself doesn't work?

src/esbmc/options.cpp
```diff
@@ -387,7 +387,10 @@ const struct group_opt_templ all_cmd_options[] = {
      "results in a verification failure when its call is reachable"},
     {"conv-assert-to-assume",
      NULL,
-     "convert assertions for bounds and pointer checks into assumptions"}}},
+     "convert assertions for bounds and pointer checks into assumptions"},
+    {"unknown-method-args-check",
+     NULL,
+     "check every argument to an unknown function call"}}},
```
intrigus-lgtm: Without looking at the source code, it's not clear (to me) what this function does.

XLiZHI:
> > > What's the use case for this? Why doesn't this handle void* pointers?
> > 
> > 
> > This is a new option developed for unknown function verification. For example, in the CHERI verification, some function internals are not known and we only need to consider whether the passed arguments will access illegal memory.
> > The new method adds support for the void type.
> 
> But what if you have a pointer to a struct and the struct itself contains pointers? Then the function could still access illegal memory.
> 
> Are there so many functions that simply "implementing" them and dereferencing the arguments yourself doesn't work?

It depends on what we want to verify. The situation you mentioned is not completely covered even in known functions, ESBMC only check read/write pointers.

lucasccordeiro:

intrigus-lgtm:
I still think that following CBMC's approach of generating bodies would have resulted in a cleaner architecture.
E.g. see https://github.com/tautschnig/cbmc/blob/2bef1f0d04c8e1578591563463579680d4c2a718/doc/cprover-manual/goto-instrument.md

lucasccordeiro:
@XLiZHI: could you please check @intrigus-lgtm's suggestion?

PR #2332, opened 2025-03-04
[symex] Property learning / strengthening

lucasccordeiro:
This PR aims to improve property learning/strengthening as follows:

1) Assert the Property: First, we check whether the property holds in a given context.
2) Satisfied, Assume It: If the property is proven correct, it can then be assumed in subsequent verification steps to simplify reasoning and strengthen future proofs.

Master (30s):

````
  correct:           18072
    correct true:    10748
    correct false:    7324
  incorrect:            41
    incorrect true:     14
    incorrect false:    27
  unknown:           15456

````

This PR (30s): 
````
  correct:           18066
    correct true:    10739
    correct false:    7327
  incorrect:            41
    incorrect true:     14
    incorrect false:    27
  unknown:           15462
  Score:             27925 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13736267786
````

### DIFF
--- a/regression/esbmc-cpp/try_catch/nec_ex4-diamond-shared-inheritance/test.desc
+++ b/regression/esbmc-cpp/try_catch/nec_ex4-diamond-shared-inheritance/test.desc
@@ -1,4 +1,5 @@
 KNOWNBUG
 main.cpp
 --unwind 10 --no-unwinding-assertions --error-label ERROR --timeout 900
+
 ^VERIFICATION FAILED$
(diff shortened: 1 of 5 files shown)


intrigus-lgtm:
Could you reduce `nec_ex4-diamond-shared-inheritance/main.cpp` to a version that still triggers the bug
`ERROR: function call: argument "c:@S@streamexception@F@streamexception#&1$@S@streamexception#::ref" type mismatch: got struct, expected pointer`?

I don't think this PR actually solves that problem, but just avoids it by pruning dead code.

Issue #1377, opened 2023-09-28
ESBMC 7.3.0 crashes  - Z3 error Sort mismatch at argument #2 for function 


tihanyin:
ESBMC 7.3.0 crashes for the following C program:

command: `esbmc FormAI_49255.c --unwind 1 --no-unwinding-assertions --k-induction --overflow --z3`

C program : 

```c

//FormAI DATASET v1.0 Category: Pac-Man Game Clone ; Style: Dennis Ritchie
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <time.h>

#define BOARD_WIDTH 20
#define BOARD_HEIGHT 15
#define MAX_PELLETS 100

typedef enum { UP, DOWN, LEFT, RIGHT } Direction;

typedef struct {
    int x;
    int y;
} Point;

typedef struct {
    Point position;
    Direction direction;
} Pacman;

typedef struct {
    Point position;
    int isPowerPellet;
} Pellet;

typedef struct {
    Pellet pellets[MAX_PELLETS];
    int pelletCount;
} PelletList;

typedef struct {
    Point position;
    int isOpen;
} Wall;

typedef struct {
    Wall walls[BOARD_WIDTH][BOARD_HEIGHT];
    PelletList pellets;
    Pacman pacman;
    Direction ghost1Direction;
    Point ghost1Position;
} GameBoard;

/** Initializes a new game board */
void initGameBoard(GameBoard *board) {
    // Initialize pellets
    board->pellets.pelletCount = 0;
    for (int i = 0; i < BOARD_WIDTH; i++) {
        for (int j = 0; j < BOARD_HEIGHT; j++) {
            if ((i == 0 && j == 0) || (i == BOARD_WIDTH - 1 && j == BOARD_HEIGHT - 1) || (i == 0 && j == BOARD_HEIGHT - 1) || (i == BOARD_WIDTH - 1 && j == 0)) {
                continue; // Don't put pellets in the corner squares
            }
            Pellet pellet = { {i, j}, rand() % 10 == 0 };
            board->pellets.pellets[board->pellets.pelletCount++] = pellet;
        }
    }

    // Initialize walls
    memset(board->walls, 0, sizeof(Wall) * BOARD_WIDTH * BOARD_HEIGHT);
    for (int i = 0; i < BOARD_WIDTH; i++) {
        board->walls[i][0].isOpen = 0;
        board->walls[i][BOARD_HEIGHT - 1].isOpen = 0;
    }
[...]

Issue #1379, opened 2023-09-28
ESBMC 7.3.0 crashes (ALL SMT solver)


tihanyin:
The ESBMC 7.3.0 crashes for all the SMT solver for this C program:

command : `esbmc FormAI_54669.c  --overflow --bitwuzla`

Output: 
```
ERROR: Bitwuzla error encountered
bitwuzla: error:  invalid call to 'bitwuzla::Term bitwuzla::mk_term(bitwuzla::Kind, const std::vector<bitwuzla::Term>&, const std::vector<long unsigned int>&)', terms with mismatching sort at indices 0 and 1
```

command : `esbmc FormAI_54669.c  --overflow --z3`

Output: 
```
Encoding remaining VCC(s) using bit-vector/floating-point arithmetic
ERROR: Z3 error Sorts Bool and (_ BitVec 8) are incompatible encountered
Aborted (core dumped)
```
command : `esbmc FormAI_54669.c  --overflow `

Output: 
```
ERROR: Boolector error encountered
[boolector] boolector_eq: nodes must have equal sorts

Aborted (core dumped)

```

C program :
```c
//FormAI DATASET v1.0 Category: Bitwise operations ; Style: protected
#include <stdio.h>
#include <stdlib.h>

// Function to check if a number is power of 2 or not
int isPowerOfTwo(unsigned int n) {
    return (n && !(n & (n - 1)));
}

// Function to swap two numbers using bitwise XOR
void swap(unsigned int* a_ptr, unsigned int* b_ptr) {
    // If both numbers are not same
    if (*a_ptr != *b_ptr) {
        // Perform XOR operation
        *a_ptr ^= *b_ptr;
        // Assign the result to the first number
        *b_ptr ^= *a_ptr;
        // Assign the result to the second number
        *a_ptr ^= *b_ptr;
    }
}

// Function to count the number of set bits in an integer
int countSetBits(unsigned int n) {
    int count = 0;
    while (n) {
        n &= (n - 1);
        count++;
    }
    return count;
}

// Function to get the maximum of two numbers without using any comparison operator
int maximum(int x, int y) {
    return x ^ ((x ^ y) & -(x < y));
}

int main() {
    unsigned int x = 15, y = 20;

    // Check if x is power of 2 or not
    if (isPowerOfTwo(x))
        printf("%u is power of 2\n", x);
    else
        printf("%u is not power of 2\n", x);

    // Swap the values of x and y
[...]

PR #2338, opened 2025-03-06
[SMT backend] Improve Overflow Handling in ESBMC

### LINKED ISSUE #2337: ESBMC boolector crash
For the following code ESBMC 7.8.1 is crashing with

```
ERROR: Boolector error encountered
[boolector] boolector_eq: nodes must have equal sorts

Abort trap: 6
```

ESBMC command:  `esbmc esbmc_crash.c --overflow`

CODE:
```c
#include <stdio.h>
#include <stdlib.h>

int maximum(int x, int y) {
    return x ^ ((x ^ y) & -(x < y));
}
int main() {
   
    unsigned int x = 15, y = 20;
    printf("    [*] The maximum of %d and %d is %d\n", 32, 56, maximum(32, 56));
    return 0;
}
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2337.

This PR improves overflow detection in ESBMC by refining the implementation of `overflow_neg` in `smt_overflow.cpp`. Additionally, it adds regression tests to ensure correct behavior when checking for arithmetic overflow, specifically negation-related overflows.

### DIFF
--- a/src/solvers/smt/smt_overflow.cpp
+++ b/src/solvers/smt/smt_overflow.cpp
@@ -204,20 +204,21 @@ smt_astt smt_convt::overflow_neg(const expr2tc &expr)
   if (int_encoding)
     return mk_smt_bool(false);
 
-  // Single failure mode: MIN_INT can't be neg'd
+  // Extract operand
   const overflow_neg2t &neg = to_overflow_neg2t(expr);
   unsigned int width = neg.operand->type->get_width();
 
-  // Handle boolean negation properly
-   if (is_bool_type(neg.operand->type)) {
-    // If operand is boolean, ensure it's converted before negation
-    expr2tc neg_value = neg2tc(neg.operand->type, neg.operand);
-    return convert_ast(neg_value);
-  }
+  // Ensure operand is an integer (convert boolean to signed integer if needed)
+  expr2tc operand = is_bool_type(neg.operand->type) 
+                        ? typecast2tc(signedbv_type2tc(width), neg.operand) 
+                        : neg.operand;
 
-  // Handle standard integer negation overflow (INT_MIN case)
-  expr2tc min_int =
-    constant_int2tc(neg.operand->type, -BigInt::power2(width - 1));
-  expr2tc val = equality2tc(neg.operand, min_int);
+  // Compute the minimum integer value for the operand's type
+  expr2tc min_int = constant_int2tc(operand->type, -BigInt::power2(width - 1));
+
+  // Check if negation would result in overflow (negating INT_MIN)
+  expr2tc val = equality2tc(operand, min_int);
+
   return convert_ast(val);
 }
+
--- a/src/solvers/smt/smt_overflow.cpp
+++ b/src/solvers/smt/smt_overflow.cpp
@@ -209,9 +209,9 @@ smt_astt smt_convt::overflow_neg(const expr2tc &expr)
   unsigned int width = neg.operand->type->get_width();
 
   // Ensure operand is an integer (convert boolean to signed integer if needed)
-  expr2tc operand = is_bool_type(neg.operand->type) 
-                        ? typecast2tc(signedbv_type2tc(width), neg.operand) 
-                        : neg.operand;
+  expr2tc operand = is_bool_type(neg.operand->type)
+                      ? typecast2tc(signedbv_type2tc(width), neg.operand)
+                      : neg.operand;
 
   // Compute the minimum integer value for the operand's type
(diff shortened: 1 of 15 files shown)


intrigus-lgtm:
Does this also fix https://github.com/esbmc/esbmc/issues/1377 ?

src/solvers/smt/smt_overflow.cpp
```diff
@@ -204,12 +204,20 @@ smt_astt smt_convt::overflow_neg(const expr2tc &expr)
...
 
-  // Single failure mode: MIN_INT can't be neg'd
+  // Extract operand
   const overflow_neg2t &neg = to_overflow_neg2t(expr);
   unsigned int width = neg.operand->type->get_width();
 
-  expr2tc min_int =
-    constant_int2tc(neg.operand->type, -BigInt::power2(width - 1));
-  expr2tc val = equality2tc(neg.operand, min_int);
+  // Ensure operand is an integer (convert boolean to signed integer if needed)
```
intrigus-lgtm: Why convert to a signed integer and not an unsigned integer instead?

lucasccordeiro:
> Does this also fix https://github.com/esbmc/esbmc/issues/1377 ?

@intrigus-lgtm: This PR does not fix this issue. Let me reduce the program and fix this issue in another PR.

intrigus-lgtm (commented):
This still feels a bit like a hack.
E.g. see https://github.com/esbmc/esbmc/issues/1379#issuecomment-1739036916

XLiZHI:
I think `github_2337_2`, `github_2337_3`, `github_2337_4`, and `github_2337_5` are not related to this PR's change. The cause of the error is mainly the bool operation ( a > b ).

src/solvers/smt/smt_overflow.cpp
XLiZHI: I guess you mean the option?
`{"unsigned-overflow-check", NULL, "enable arithmetic over- and underflow check for unsigned integers"}`

lucasccordeiro: The conversion to a signed integer is done to handle the overflow caused by negating the smallest signed integer (INT_MIN). If we use an unsigned integer, there would be no overflow for the negation because of the wrapping behavior.

lucasccordeiro:
> This still feels a bit like a hack.

@intrigus-lgtm: Do you think you could submit an alternative PR? :-)

As @mikhailramalho pointed out:

"I agree that this is unsatisfying, we should do everything following the
C/C++ standard up to before converting to SMT and then convert the data to
the format that SMT expects, but I think this will be a huge project."

lucasccordeiro:
> I think github_2337_2, github_2337_3, github_2337_4, and github_2337_5 are not related to this PR's change. The cause of the error is mainly the bool operation ( a > b ).

@XLiZHI: I agree with you, but I don't think it harms to add these test cases to check the other cases in the overflow method ;-)

src/solvers/smt/smt_overflow.cpp
```diff
@@ -204,12 +204,20 @@ smt_astt smt_convt::overflow_neg(const expr2tc &expr)
...
+  expr2tc operand = is_bool_type(neg.operand->type)
```
rafaelsamenezes: Can't you always do the typecast? Our simplifier should be smart enough to handle double typecast.

rafaelsamenezes (approved):
I would also suggest creating an umbrella issue so we do not lose track of these typecast hacks.

src/solvers/smt/smt_overflow.cpp
lucasccordeiro: Implemented as suggested.

XLiZHI (approved):

lucasccordeiro:
I have just created this umbrella issue: https://github.com/esbmc/esbmc/issues/2339

intrigus-lgtm:
Unrelated to this PR, but I'd say the code should simply ignore unsignedbvs:

`esbmc --overflow-check --unsigned-overflow-check`
```c
int main() {
unsigned int a = 0x80000000;
return -a;
}
```

```
State 3 file foo.c line 3 column 1 function main thread 0
Violated property:
  file foo.c line 3 column 1 function main
  arithmetic overflow on neg
  !overflow("unary-", a)


VERIFICATION FAILED
```

lucasccordeiro:
Thanks, @intrigus-lgtm. 

I have just created this PR to fix this issue: https://github.com/esbmc/esbmc/pull/2340.

PR #2340, opened 2025-03-07
[Goto-Check] Fix Unsigned Negation Overflow Handling

### LINKED ISSUE #1844: Negation of Unsigned is not being computed correctly
Hi again,
for the following program where the error is unreachable:
```c
extern void __VERIFIER_error(void);
unsigned int id(int x){
    return x;
}

int main(){
    if((4294967295U+ 1U) - id(4294967295U) < 0U){
	__VERIFIER_error();
    }
    return 0;
}
```
running `esbmc --interval-analysis` gives `VERIFICATION FAILED`. Running without `--interval-analysis` gives the correct result `VERIFICATION SUCCESSFUL`.

Version: ece7849b815b754df83ff833e6089d7d902d0e8c
OS: Ubuntu 20.04

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1844.

This pull request addresses a regression in ESBMC related to unsigned negation overflow checking (see https://github.com/esbmc/esbmc/pull/2338#issuecomment-2705119397). It introduces new test cases and updates the SMT solver logic to handle negation for unsigned integers properly.

Master (30s):

````
  correct:           18074
    correct true:    10747
    correct false:    7327
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15455
  Score:             27973 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13638155792
````

This PR (30s):

````
  correct:           18072
    correct true:    10748
    correct false:    7324
  incorrect:            41
    incorrect true:     14
    incorrect false:    27
  unknown:           15456

````

### CODE REVIEW

**Review by intrigus-lgtm — CHANGES_REQUESTED** [review]
That change in `simplify_expr2.cpp` doesn't look fully right to me.
I looked into this problem due to https://github.com/esbmc/esbmc/issues/1844 and came up with a patch instead.
(The partial root cause is that `Negator` does not really care whether the result of its computation is outside the range of the type.
But it seems that this is intentionally so, because for binary ops (`simplify_arith_2ops`), truncation is applied inside the function like this:
```cpp
    // Fix rounding when an overflow occurs
    if (!is_nil_expr(simpl_res) && is_constant_int2t(simpl_res))
      migrate_expr(
        from_integer(
          to_constant_int2t(simpl_res).value,
          migrate_type_back(simpl_res->type)),
        simpl_res);
```

Here's the patch that applies the logic for binary ops to unary ops.

```patch
diff --git a/src/util/simplify_expr2.cpp b/src/util/simplify_expr2.cpp
--- a/src/util/simplify_expr2.cpp	(revision 16c77a592e4bff03b6e1d04431ae83d6b8d9c77f)
+++ b/src/util/simplify_expr2.cpp	(date 1741047981372)
@@ -683,6 +683,15 @@
       (constant_int2t & (*)(expr2tc &)) to_constant_int2t;
 
     simpl_res = TFunctor<constant_int2t>::simplify(to_simplify, to_constant);
+
+    // Properly handle truncation when an overflow occurs
+    // or when the result gets out of the bounds of the type
+    if (!is_nil_expr(simpl_res) && is_constant_int2t(simpl_res))
+      migrate_expr(
+        from_integer(
+          to_constant_int2t(simpl_res).value,
+          migrate_type_back(simpl_res->type)),
+        simpl_res);
   }
   else if (is_fixedbv_type(value))
   {
```

src/solvers/smt/smt_overflow.cpp
```diff
@@ -200,25 +200,42 @@ smt_astt smt_convt::overflow_cast(const expr2tc &expr)
...
 
-  // Simplify if typecast was performed
-  simplify(operand);
+  if (is_unsigned)
+  {
+    // **Unsigned Negation Overflow Check**
+    // In unsigned arithmetic, negation (-x) is effectively (UINT_MAX + 1 - x).
+    // Any nonzero value negated will wrap around, which is unexpected behavior.
+    expr2tc zero = constant_int2tc(neg.operand->type, 0);
+    // Overflow occurs if operand is x == 0
```
intrigus-lgtm: It's the other way around, isn't it?
Overflow occurs <=> x != 0
E.g. UBSAN agrees with me:
https://godbolt.org/z/YKhvenec7

lucasccordeiro:
Thanks, @intrigus-lgtm. I'm checking your patch using my mobile phone :-) I can't see the difference between the original code and your patch; they look the same (unless my phone is not working correctly :-) what is the difference?

intrigus-lgtm:
(Whether it's a good idea that `Negator` and the similar classes can produce values that are out of range for the type is a different question. But this is also a problem for e.g. `constant_int2tc` where the value does not have to fit into the type...)

intrigus-lgtm:
> Thanks, @intrigus-lgtm. I'm checking your patch using my mobile phone :-) I can't see the difference between the original code and your patch; they look the same (unless my phone is not working correctly :-) what is the difference?

The first code is the existing code of `simplify_arith_2ops` (**2**) while the patch would apply to `simplify_arith_1op` (**1**) and add this similar code.

src/solvers/smt/smt_overflow.cpp
```diff
@@ -200,25 +200,42 @@ smt_astt smt_convt::overflow_cast(const expr2tc &expr)
...
-  // Simplify if typecast was performed
-  simplify(operand);
+  if (is_unsigned)
+  {
+    // **Unsigned Negation Overflow Check**
+    // In unsigned arithmetic, negation (-x) is effectively (UINT_MAX + 1 - x).
+    // Any nonzero value negated will wrap around, which is unexpected behavior.
+    expr2tc zero = constant_int2tc(neg.operand->type, 0);
+    // Overflow occurs if operand is x != 0
+    expr2tc val = equality2tc(neg.operand, zero);
```
intrigus-lgtm: This should also be updated :)

lucasccordeiro:
Yes, I will do as a soon as I get to the computer :-) I am still on the mobile phone :-)Sent from my iPhoneOn 7 Mar 2025, at 18:14, intrigus-lgtm ***@***.***> wrote:﻿
@intrigus-lgtm commented on this pull request.



In src/solvers/smt/smt_overflow.cpp:
>  
-  // Simplify if typecast was performed
-  simplify(operand);
+  if (is_unsigned)
+  {
+    // **Unsigned Negation Overflow Check**
+    // In unsigned arithmetic, negation (-x) is effectively (UINT_MAX + 1 - x).
+    // Any nonzero value negated will wrap around, which is unexpected behavior.
+    expr2tc zero = constant_int2tc(neg.operand->type, 0);
+    // Overflow occurs if operand is x != 0
+    expr2tc val = equality2tc(neg.operand, zero);

This should also be updated :)

lucasccordeiro:
@intrigus-lgtm: Please feel free to review this PR again. I have added your patch.

src/util/simplify_expr2.cpp
```diff
@@ -731,6 +740,23 @@ expr2tc neg2t::do_simplify() const
...
+    const expr2tc modulus_expr = constant_int2tc(value->type, modulus);
+
+    // Perform modular negation: (modulus - x) % modulus
+    const expr2tc negated_value = sub2tc(value->type, modulus_expr, value);
+    expr2tc wrap = modulus2tc(value->type, negated_value, modulus_expr);
+    wrap->simplify();
+
+    return wrap;
+  }
+
```
intrigus-lgtm, line 759: This should now be unnecessary.

regression/esbmc/github_2338_9/main.c
```diff
@@ -0,0 +1,8 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0x80000000; // Power of two boundary case
+  unsigned int result = -a; 
+  assert(result == 0x80000001); // Incorrect assertion
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_8/main.c
```diff
@@ -0,0 +1,8 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0x80000000; // Negative negation of the halfway point (overflow risk)
+  unsigned int result = -a; 
+  assert(result == 0x7FFFFFFF); // Incorrect expected result for unsigned negation
```
intrigus-lgtm, line 6: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_7/main.c
```diff
@@ -0,0 +1,8 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0x80000001;
+  unsigned int result = -a;
+  assert(result == 0x7FFFFFFF); // Expected modular negation
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_6/main.c
```diff
@@ -0,0 +1,8 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0x40000000;
+  unsigned int result = -a;
+  assert(result == 0xC0000000); // Expected modular negation
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_5/main.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0x00000001;
+  unsigned int result = -a;
+  assert(result == 0xFFFFFFFF); // Expected result due to modular arithmetic
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_4/main.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0xFFFFFFFF;
+  unsigned int result = -a;
+  assert(result == 1);
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_2/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unsigned-overflow-check --no-simplify
```
intrigus-lgtm: The use of `--no-simplify` should now be unnecessary after the fix.

regression/esbmc/github_2338_2/main.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
+
+int main() {
+  unsigned int a = 0xFFFFFFFF;
+  unsigned int result = -a;
+  assert(result == 1);
```
intrigus-lgtm: I think this assert is never actually triggered, because the negation will cause an error beforehand.

regression/esbmc/github_2338_10/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unsigned-overflow-check
```
intrigus-lgtm: IMHO, this test should not have this flag and it should result in successful verification.
Instead, we should run this testcase once with the `--interval-analysis` flag and once without as stated in #1844 
Instead of `--interval-analysis`, we could also just run the TC once with `--no-simplify`.

Also, could you please move this to `github_1844` so it's clear where the TC originates from?

regression/esbmc/github_2338_9/main.c
lucasccordeiro: Assertion removed.

regression/esbmc/github_2338_7/main.c
lucasccordeiro: Assertion removed.

regression/esbmc/github_2338_6/main.c
lucasccordeiro: Assertion removed.

regression/esbmc/github_2338_5/main.c
lucasccordeiro: Assertion removed.

regression/esbmc/github_2338_2/test.desc
lucasccordeiro: removed `--no-simplify`.

regression/esbmc/github_2338_2/main.c
lucasccordeiro: Assertion removed.

src/util/simplify_expr2.cpp
lucasccordeiro: @intrigus-lgtm: @rafaelsamenezes' suggestion was to keep both solutions. Are you okay with that?

regression/esbmc/github_2338_4/main.c
lucasccordeiro: Assertion removed.

regression/esbmc/github_2338_10/test.desc
lucasccordeiro: Implemented as suggested.

lucasccordeiro:
@intrigus-lgtm: Please feel free to review it again. As discussed above, @rafaelsamenezes' suggestion is also to keep his solution.

src/util/simplify_expr2.cpp
intrigus-lgtm: I'm not a big fan of having similar code in two places.
Having the code in `simplify_arith_1op` is more future proof than having the code in the neg2t class.
But I don't feel strongly about this.

It just made more sense for me to copy the existing structure from binary ops also for unary ops and to not specialcase truncation/overflow for specific operators, because this is also not the case elsewhere.

PR #2343, opened 2025-03-08
[SMT backend] Add Overflow Detection for Integer Arithmetic 

lucasccordeiro:
This pull request adds ESBMC's handling of overflow detection for integer arithmetic operations: addition, subtraction, division, and multiplication. Integer operations follow arithmetic rules, including overflow behavior, whereas bitvectors typically use modular arithmetic. Directly reusing bitvector encoding could misinterpret signed numbers and break integer arithmetic semantics for overflow check.

### Tasks
- [x] Addition
- [x] Subtraction
- [x] Multiplication
- [x] Division
- [x] Unary

Master (30s):

````
  correct:           18072
    correct true:    10748
    correct false:    7324
  incorrect:            41
    incorrect true:     14
    incorrect false:    27
  unknown:           15456

````

This PR (30s):

````
  correct:           18063
    correct true:    10740
    correct false:    7323
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15466
  Score:             27955 (max: 55885)

GitHub Actions: https://github.com/esbmc/esbmc/actions/runs/13749597180
````

### DIFF
--- a/regression/esbmc/integer1/main.c
+++ b/regression/esbmc/integer1/main.c
@@ -1,0 +1,10 @@
+#include <assert.h>
+
+int main() {
+  int a = nondet_int();
+  int b = nondet_int();
+
+  __ESBMC_assume(a > 0 && b > 0);
+
+  assert((a + b) > 0);
+}
--- a/regression/esbmc/integer1/test.desc
+++ b/regression/esbmc/integer1/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--z3 --ir
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/integer11_fail/main.c
+++ b/regression/esbmc/integer11_fail/main.c
@@ -1,0 +1,61 @@
+/*
+  hardware integer division program, by Manna
+  returns q==A//B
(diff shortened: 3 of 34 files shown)


intrigus-lgtm (commented):
I think I'm not knowing enough about integer encoding, but why can't we reuse the encoding of the bitvector case?

lucasccordeiro:
Hi @intrigus-lgtm,

This is a very good question.

I have tried to reuse as much code as possible. However, note that integer variables in the mathematical domain vary from -infinity to +infinity, while bitvectors have a limited fixed length. We have some useful discussion about this topic here: https://ssvlab.github.io/lucasccordeiro/papers/tse2012.pdf.

Additionally, we can't directly reuse the bitvector encoding for integers because integers have specific signed and unsigned representations, while bitvectors are fixed-length sequences of bits. Integer operations follow arithmetic rules, including overflow behavior, whereas bitvectors typically use modular arithmetic. Directly reusing bitvector encoding could misinterpret signed numbers and break integer arithmetic semantics for overflow check.

I am happy to discuss it further.

PR #2351, opened 2025-03-17
[OM] reimplement atexit using infinite arrays

lucasccordeiro:
This PR reimplements our `atexit` handling code using infinite arrays (`__ESBMC_inf_size` annotation). This approach replaces dynamic memory allocation with an infinite array to prevent malloc-related verification efforts and thus simplifies the generation of verification conditions.

Master (30s):

````
  correct:           18070
    correct true:    10742
    correct false:    7328
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15459
  Score:             27964 (max: 55885)


https://github.com/esbmc/esbmc/actions/runs/13888616601
````

This PR (30s):

````
  correct:           18075
    correct true:    10743
    correct false:    7332
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15454
  Score:             27970 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/13906715936
````

### DIFF
--- a/regression/esbmc/atexit-5/main.c
+++ b/regression/esbmc/atexit-5/main.c
@@ -1,0 +1,81 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+extern _Bool __VERIFIER_nondet_bool();
+
+/* Enhanced regression test for atexit */
+
+int **g = NULL;
+int *h_ptr = NULL;
+
+void free_g1() {
+    printf("Freeing g\n");
+    free(g);
+    g = NULL;
+}
+
+void free_g2() {
+    if (g != NULL) {
+        printf("Freeing *g\n");
+        free(*g);
+    }
+}
+
+void free_h_ptr() {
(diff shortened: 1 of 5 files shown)


XLiZHI:
Is there an example that shows the improvement of this change?

lucasccordeiro:
> Is there an example that shows the improvement of this change?

Our suite has four regression tests, but they are all very simple to verify (which is fine).

Let me think about a large example to check the performance difference!

lucasccordeiro:
@XLiZHI: I have added some examples to show how fast this solution is compared with the malloc implementation. In particular, you can try out `  regression/esbmc/atexit-6/main.c` :-) 

Please let me know if you have further comments.

XLiZHI (approved):
Thanks for the example, Prof. @lucasccordeiro. LGTM

intrigus-lgtm (approved):

PR #2354, opened 2025-03-19
[Z3] Support bitwise for integer arithmetic

lucasccordeiro:
This PR refactors our bitwise operations handling and enhances our SMT conversion functions to better support integer and boolean types. In particular, it introduces more consistent handling for boolean and integer types and improves compatibility with the Z3 solver's bit-vector operations.
- [X] bitnot_id
- [X] bitxor_id
- [x] bitand_id
- [x] bitor_id
- [x] bitnand_id
- [x] bitnor_id
- [x] bitnxor_id

This PR (30s): 

````
  correct:           18073
    correct true:    10743
    correct false:    7330
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15456
  Score:             27968 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/13956944037
````

Master (30s):

````
  correct:           18075
    correct true:    10743
    correct false:    7332
  incorrect:            40
    incorrect true:     13
    incorrect false:    27
  unknown:           15454
  Score:             27970 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/13906715936
````

### DIFF
--- a/regression/esbmc/bitwise1/main.c
+++ b/regression/esbmc/bitwise1/main.c
@@ -1,0 +1,43 @@
+#include <assert.h>
+#include <stdio.h>
+
+void test_xor_operations() {
+    // Test case 1: Basic XOR operation
+    int a = 5;  // 0101 in binary
+    int b = 3;  // 0011 in binary
+    int result = a ^ b;  // 0101 ^ 0011 = 0110 (should be 6)
+    assert(result == 6);
+    printf("Test case 1 passed\n");
+
+    // Test case 2: XOR with zero (should return the same number)
+    int c = 0;  // 0000 in binary
+    result = a ^ c;  // 0101 ^ 0000 = 0101 (should be 5)
+    assert(result == 5);
+    printf("Test case 2 passed\n");
+
+    // Test case 3: XOR of a number with itself (should return 0)
+    result = a ^ a;  // 0101 ^ 0101 = 0000 (should be 0)
+    assert(result == 0);
+    printf("Test case 3 passed\n");
+
+    // Test case 4: XOR with all bits set (should invert the bits)
+    int d = ~0;  // All bits set (typically -1)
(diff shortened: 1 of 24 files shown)


intrigus-lgtm (changes requested):
1. If we want to support bitwise operations for integer mode, support should be added  on a higher level and not only for z3.
2. `mk_not` should **continue** to be only used for boolean not. There is a one to one correspondence between `mk_foo` and the functions of the smtlib theories, so I don't see why we should break this convention.

I also only somewhat see the point of integer mode when it just uses bitvectors behind the scenes. Together with the unsoundness of it #2357 of it, it's hard to understand what integer mode is useful for.

lucasccordeiro:
Hi @intrigus-lgtm,

Thanks for your feedback.

If we want to fully support bitwise operations in integer mode, I agree that the implementation should be handled at a higher level rather than only in the Z3 backend.

Regarding `mk_not`, I see your point about maintaining a strict one-to-one mapping with SMT-LIB functions. I have just fixed this here.

I also acknowledge your concerns about integer mode's usefulness, particularly given its reliance on bit vectors and the potential unsoundness issues. If integer mode is retained, we should clarify its intended use cases and constraints to ensure it's well understood and consistently applied.

I am happy to talk to you more if you have any specific concerns or suggestions.

intrigus-lgtm:
> Hi @intrigus-lgtm,
> ...
> I also acknowledge your concerns about integer mode's usefulness, particularly given its reliance on bit vectors and the potential unsoundness issues. If integer mode is retained, we should clarify its intended use cases and constraints to ensure it's well understood and consistently applied.

My comment about the usefulness of integer mode was maybe a bit harsh, I was just confused that the PR would add unsoundness, because bitwise operations truncate/constrain the values of the previously unbounded integers, but with the background that it's already unsound, that's more reasonable.

PR #2358, opened 2025-03-19
[C verification] remove nonexistent function

### LINKED ISSUE #2355: [Clang-C-Frontend]  Assertion `type.id() == t_floatbv' failed.
````C
#include <assert.h>

int nand(int a, int b) {
    return !(a & b);  // NAND is the negation of AND
}

int main() {
    // Test NAND operation
    assert(nand(1, 1) == 0); // 1 NAND 1 should be 0
    return 0;
}
````

````
$ esbmc main.c
Parsing main.c
Converting
esbmc: /home/lucas/ESBMC_Project/esbmc/src/util/std_types.h:650: const floatbv_typet& to_floatbv_type(const typet&): Assertion `type.id() == t_floatbv' failed.
Aborted (core dumped)
````

XLiZHI:
Fixes https://github.com/esbmc/esbmc/issues/2355.

nand() doesn't actually exist in the standard C library; the <math.h> header only defines nan(), nanf(), and nanl().

We introduced a new function : -)


lucasccordeiro (approved):

intrigus-lgtm (changes requested):
I didn't try to run this, but what happens if you name your function just `nan` instead?
I would assume that it would still fail.
As far as I know, `nan` is not a reserved name (unlike for example names starting with "str" or "to"), so it's perfectly legal to have such a function when not including `math.h`.

I think a proper solution would be to remove the adjustments for `nan` functions, only leaving those called `__builtin_nan` and then introduces appropriate functions to the `math.h` header that call those builtins.

lucasccordeiro:
@XLiZHI: could you please check @intrigus-lgtm's comment? Does this PR handle the mentioned test case?

XLiZHI:
According to the AST, using nan, nanl, nanf does not call `__builtin_nan()`. This edge case is not worth discussing. I tested to clang 20, so I think I'll stick with the current version.

lucasccordeiro:

Thanks for reviewing this PR, @intrigus-lgtm.

Issue #1290, opened 2023-08-16
ESBMC 7.3 crashes (terminate called after throwing an instance of 'array_type2t::dyn_sized_array_excp')


tihanyin:
ESBMC 7.3 crashes for this small C program using the following command:

`esbmc DATA2/FormAI_24310.c --overflow `

```c
//FormAI DATASET v1.0 Category: Alien Language Translator ; Style: synchronous
#include <stdio.h>
#include <string.h>

void translate(char* str) {
    // Define the rules of the alien language
    int len = strlen(str);
    char translated_str[len+1];
    int vowel_check[len+1];
    memset(vowel_check, 0, sizeof(vowel_check));
    memset(translated_str, 0, sizeof(translated_str));

    for(int i=0; i<len; ++i) {
        if(str[i] == 'a' || str[i] == 'e' || str[i] == 'i' || str[i] == 'o' || str[i] == 'u') {
            vowel_check[i] = 1;
        }
    }

    int pos = 0;
    for(int i=0; i<len; ++i) {
        if(vowel_check[i]) {
            if(i+3 < len && !vowel_check[i+1] && !vowel_check[i+2] && !vowel_check[i+3]) {
                strncpy(translated_str+pos, str+i+1, 2);
                translated_str[pos+2] = str[i];
                pos += 3;
                i += 3;
            } else {
                translated_str[pos++] = str[i];
            }
        } else {
            translated_str[pos++] = str[i];
        }
    }

    printf("Translated Alien Language: %s\n", translated_str);
}

int main() {
    char str[100];
    printf("Enter the Alien's language to translate: ");
    scanf("%[^\n]%*c", str);

    translate(str);
    return 0;
}
```

OUTPUT:
```
esbmc DATA2/FormAI_24310.c --overflow
ESBMC version 7.3.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing
Converting
terminate called after throwing an instance of 'array_type2t::dyn_sized_array_excp'
Aborted (core dumped)
```

fbrausse:
On a Debug build this already fails earlier, during migrate_expr():
```
Parsing gh-1290.c
Converting
esbmc: src/util/migrate.cpp:535: expr2tc sym_name_to_symbol(irep_idt, type2tc): Assertion `endatptr != atstr.c_str()' failed.
```
[...]

PR #2359, opened 2025-03-19
[TC] add TC from #1290

intrigus-lgtm:
It's unclear when this issue has been fixed, but it's not reproducible on latest master. We skip BMC, because the original issue was happening before it.


lucasccordeiro (approved):

lucasccordeiro:

Issue #1175, opened 2023-07-10
Dereferencing from a const should be illegal


rafaelsamenezes:
ESBMC should fail on a program such as:

```c
int main()
{
  int *ptr = 90;
  int v = *ptr;
}
```

The VCC generated was:

```
dereference failure: invalid pointer
{-1} alloc&0#1 == ARRAY_OF(0)
{-2} is_dynamic&0#1 == ARRAY_OF(0)
|--------------------------
{1} execution_statet::\guard_exec?0!0 => !(!alloc&0#1[POINTER_OBJECT((signed int *)90)] && is_dynamic&0#1[POINTER_OBJECT((signed int *)90)] || (signed int *)90 == invalid-object)
```

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/commit/bfe6c3434545be77492dbd5459318bde585ee960.

intrigus-lgtm:
This has not actually been fixed.
E.g. try this:
```
int main()
{
  int *ptr = (int*)90;
  int v = *ptr;
}
```

intrigus-lgtm:
Also, this currently breaks windows tests I think, because there int to pointer conversion is not an error but merely a warning:
See e.g.: https://github.com/esbmc/esbmc/actions/runs/13876300679/job/38829045505#step:9:3265

lucasccordeiro:
Thanks, @intrigus-lgtm. I have just changed the title of this issue to remind us to check in Windows.

intrigus-lgtm:
Opened https://github.com/esbmc/esbmc/pull/2361 to make this issue compatible with Windows as the underlying issue is still unsolved on all platforms.

intrigus-lgtm:
Hmm, this still doesn't look fully fixed to me?
```
int main()
{
    int *ptr = (int *)93;
    int v = *ptr;
}
```
Succeeds with `VERIFICATION SUCCESSFUL`.

PR #2361, opened 2025-03-19
Make TC from #1175 compatible with windows

intrigus-lgtm:
Clang rejects this code because there is an implicit converison from int to pointer.
Windows is more lax than this and accepts this code.
Use an explicit cast instead.


lucasccordeiro (approved):

lucasccordeiro:

PR #2362, opened 2025-03-19
[TC] add variant of github_60 test

intrigus-lgtm:
Add a variant of the test from #60/#1533 that should fail on all platforms (previously it only failed on Windows, because only Windows used Z3)
`--no-slice` is used, because otherwise the bug is not triggered.


lucasccordeiro (approved):

lucasccordeiro:

PR #2389, opened 2025-04-02
[Cov & Sol] Coverage support for Solidity

ChenfengWei0:
**Coverage:**  
- ESBMC's coverage calculation now works for Solidity programs.  

**Solidity:**  
- Allow users to set multiple verified target contracts. For example, `--contract "A B"` will verify both contracts A and B.  
- Added support for the `maximum` and `minimum` keywords.  
- Fixed some bugs and performed refactoring.


src/goto-programs/goto_coverage.h
```diff
@@ -62,6 +62,9 @@ class goto_coveraget
   std::string get_filename_from_path(std::string path);
   void set_target(const std::string &_tgt);
   bool is_target_func(const irep_idt &f, const std::string &tgt_name) const;
+  void set_no_hide();
```
XLiZHI: This function doesn't seem to be used.

src/goto-programs/goto_coverage.cpp
```diff
@@ -30,12 +30,10 @@ void goto_coveraget::replace_all_asserts_to_guard(
   Forall_goto_functions (f_it, goto_functions)
     if (f_it->second.body_available && f_it->first != "__ESBMC_main")
     {
-      // "--function" mode
-      if (
-        target_function != "" && !is_target_func(f_it->first, target_function))
+      goto_programt &goto_program = f_it->second.body;
```
XLiZHI, line 33: https://github.com/esbmc/esbmc/blob/6e787fda5012c1336325baadf33cbf1a77031f31/src/goto-programs/goto_convert_functions.cpp#L39-L51

I believe we already have a function to handle hide label, does it execute before goto_coverage? If so, maybe we can just use `f_it->second.body.hide` to get it.

XLiZHI (commented):
Some comments

lucasccordeiro:
@XLiZHI and @intrigus-lgtm: are you okay with the changes done in the clang c frontend?

XLiZHI (approved):

src/goto-programs/goto_coverage.h
ChenfengWei0: changed as suggested

src/goto-programs/goto_coverage.cpp
ChenfengWei0: changed as suggested. Many thanks

src/clang-c-frontend/clang_c_main.cpp
```diff
@@ -70,17 +70,29 @@ bool clang_c_maint::clang_main()
     // if user provided class/contract name
     if (!config.cname.empty() && !config.main.empty())
     {
-      const std::string fmt = "@" + config.cname + "@F@" + main;
-      if (it->second.as_string().find(fmt) == std::string::npos)
+      // Allow user to provide a range. e.g.
```
lucasccordeiro, line 73: Can you provide a test case in C++ that checks this new "feature"?

lucasccordeiro (changes requested):

src/clang-c-frontend/clang_c_main.cpp
ChenfengWei0: Added as suggested

regression/esbmc-cpp/cpp/github_2389_1/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.cpp
+--class "A B" --function test
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: Shouldn't it give verification failed? There is an assert(0) in the test method of class B.

src/clang-c-frontend/clang_c_main.cpp
```diff
@@ -70,18 +70,37 @@ bool clang_c_maint::clang_main()
     // if user provided class/contract name
     if (!config.cname.empty() && !config.main.empty())
     {
-      const std::string fmt = "@" + config.cname + "@F@" + main;
-      if (it->second.as_string().find(fmt) == std::string::npos)
-        continue;
+      // Allow user to provide a range. e.g.
+      // --class "A B" --function test
+      // --contract "Base Derieve" --function _ESBMC_main
+      // yet only the first matched one will get verified
```
lucasccordeiro, line 76: Why will only the first match be verified? What is the purpose of this range?

lucasccordeiro (changes requested):

regression/esbmc-cpp/cpp/github_2389_1/test.desc
ChenfengWei0: Only the first matched one will be verified

src/clang-c-frontend/clang_c_main.cpp
ChenfengWei0: Currently, ESBMC lacks an option to set multiple verified functions—we can only specify one target function. This class range would make more sense once we can set all ambiguous functions (i.e., functions with the same name) as verification targets.

For now, it’s primarily used in Solidity to define a range of target contracts.

ChenfengWei0: Also, since esbmc now supports multi-property, we might extend the `--function` to allow multiple inputs. Like `--function ''test1 test2"`

lucasccordeiro:

PR #2390, opened 2025-04-02
Update ExternalDependencies.cmake

rafaelsamenezes:
This should fix the old dependencies which where asking for minimum cmake of < 3.5

### DIFF
--- a/src/solvers/boolector/CMakeLists.txt
+++ b/src/solvers/boolector/CMakeLists.txt
@@ -14,7 +14,7 @@ else()
       NAME boolector
       DOWNLOAD_ONLY YES
       GITHUB_REPOSITORY rafaelsamenezes/boolector
-      GIT_TAG esbmc)
+      GIT_TAG esbmc2)
 
     # We can't use ExternalProject because we need it ready
     # at configuration time
--- a/src/solvers/boolector/CMakeLists.txt
+++ b/src/solvers/boolector/CMakeLists.txt
@@ -13,7 +13,8 @@ else()
     CPMAddPackage(
       NAME boolector
       DOWNLOAD_ONLY YES
-      GITHUB_REPOSITORY rafaelsamenezes/boolector)
+      GITHUB_REPOSITORY rafaelsamenezes/boolector
+      GIT_TAG master)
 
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

src/solvers/boolector/CMakeLists.txt
```diff
@@ -13,8 +13,7 @@ else()
     CPMAddPackage(
       NAME boolector
       DOWNLOAD_ONLY YES
-      GITHUB_REPOSITORY Boolector/boolector
-      GIT_TAG 3.2.3)
+      GITHUB_REPOSITORY rafaelsamenezes/boolector)
```
intrigus-lgtm: Why does this refer to your personal repo now?

rafaelsamenezes: Because Boolector cmake is old and the project is archived

rafaelsamenezes:
Still have some issue: https://github.com/hwmcc/btor2tools/issues/29

src/solvers/boolector/CMakeLists.txt
intrigus-lgtm: Why can't we just patch the CMake file after checking out upstream?
Alternatively, could we move this repo under the esbmc organization?

XLiZHI:
Can we upload the zip file and download it at build time? Boolector is no longer updated anyway.

lucasccordeiro:
> Can we upload the zip file and download it at build time? Boolector is no longer updated anyway.

I think we use this approach for Clang Windows.

src/solvers/boolector/CMakeLists.txt
rafaelsamenezes: > Why can't we just patch the CMake file after checking out upstream?

That's definitely a better idea.  I will just try to get this passing and the work on the patches.

rafaelsamenezes:
> Can we upload the zip file and download it at build time? Boolector is no longer updated anyway.

Let us find out everything that is broken. I think that bitwuzla will have the same issue

lucasccordeiro:

Thanks for reviewing it, @intrigus-lgtm and @XLiZHI.

PR #2392, opened 2025-04-03
[Solidity] Upgrade the AST merging algorithm for multiple files

### LINKED ISSUE #2388: [Solidity] incorrect AST-merging algorithm for multi-files
Image

Some nodes are missing after the merge

@prototypeC14

prototypeC14:
This pull request is to fix #2388 by upgrading the algorithm.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -182,42 +188,115 @@ void solidity_convertert::merge_multi_files()
...
+        }
       }
     }
   }
+
+  // The initial part of the nodes in a single AST includes an import information description section
+  // and a version description section.This is followed by all the information that needs to be verified.
+  // Therefore, the rest of the key nodes need to be inserted sequentially thereafter
+  // It also means before the first ContractDefinition node.
+  // size_t insert_pos = 0;
```
ChenfengWei0: remove the comment code

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -182,42 +188,115 @@ void solidity_convertert::merge_multi_files()
...
-          src_ast_json["nodes"].begin() + insert_pos, node);
-        ++insert_pos; // Adjust the insert position for the next node
+      std::string import_path = node["absolutePath"];
+
+      if (visited_paths.count(import_path))
+        continue;
+      visited_paths.insert(import_path);
+
+      const auto &imported_ast = path_to_json.at(import_path);
+      // printf(
```
ChenfengWei0: use `log_debug("solidity", "....")` instead

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -182,42 +188,115 @@ void solidity_convertert::merge_multi_files()
...
+    if (nodes[i]["nodeType"] == "ImportDirective")
     {
-      insert_pos = i;
-      break;
+      last_import_pos = i;
     }
   }
 
-  for (size_t i = 1; i < src_ast_json_array.size(); ++i)
+  // Insert ImportDirective node for main file
```
ChenfengWei0, line 267: why do you need to create your own node? also how can you ensure that id `1` is not conflicted with other nodes.

I would prefer to use the node directly from the original AST

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -182,42 +188,115 @@ void solidity_convertert::merge_multi_files()
...
+
+      const auto &imported_ast = path_to_json.at(import_path);
+      // printf(
+      //   "Merging AST from imported file: %s\n", import_path.c_str());
+      // printf(
+      //   "  AST file is as follows:\n%s\n",
+      //   imported_ast.dump(2).c_str());
+      const auto &imported_nodes = imported_ast["nodes"];
+
+      size_t insert_before = i;
```
ChenfengWei0, line 291: I do not understand this. Why we need to insert before and after

ChenfengWei0 (changes requested):

src/solidity-frontend/solidity_convert.cpp
prototypeC14: To distinguish nodes from different absolute paths, this node is constructed using its source absolute path and (max ID + 1) to avoid conflicts.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: Since each file contains the paths of its imported files, directly inserting them would cause nesting, making it hard to distinguish which nodes belong to which paths.

prototypeC14: For example，assume A.sol import B.sol, B.sol import C.sol.
The structure are supposed be:
```C++
"absolutePath": "C.sol" --> this contains in the AST of B.sol
"nodes": [from C.sol]
"absolutePath": "B.sol" --> this contains in the AST of A.sol
"nodes": [from B.sol]
// This part represents the original content of the file before any insertion, without any import statements.
"absolutePath": "A.sol" --> This part is manually added
"nodes": [from A.sol]

File: src/solidity-frontend/solidity_convert.cpp [hunk 5]
```diff
@@ -182,39 +188,94 @@ void solidity_convertert::merge_multi_files()
...
-  // It also means before the first ContractDefinition node.
-  size_t insert_pos = 0;
-  for (size_t i = 0; i < src_ast_json["nodes"].size(); ++i)
+  // Track visited paths to avoid duplicate insertions
+  std::unordered_set<std::string> visited_paths;
+  visited_paths.insert(main_path); // main file is already there
+
+  // The purpose of the following processing is to ensure that the merged AST is presented in the following format:
+  // Assume A.sol import B.sol, B.sol import C.sol
+  // "absolutePath": "C.sol" --> this contains in the AST of B.sol
```
ChenfengWei0, line 237: why c.sol contains the AST of B? This seems anti-intuitive
I would each to see the format like:

// Assume A.sol import B.sol, B.sol import C.sol
{absolutpath : A
Nodes:
[
{absolutpath : C}
{Nodes: ...}
{absolutpath : B}
{Nodes: ...}
{absolutpath : A}
{Nodes: ...}
]}

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: I still don't think you need to manually construct anything. My question is: if a import b and b import c, in the ast file of a, will there be any node that contains same `ID`?

ChenfengWei0: If not, you can just merge them into the `nodes` of a, without modifying anything (maybe insert the absolut_path of a beforehand )

prototypeC14: Within the compilation of a single file, there will be no duplicate IDs. The Solidity compiler (solc) assigns a unique, incrementally generated ID to each node. Direct merging can lead to nesting in some cases, making it difficult to clearly associate paths with their corresponding nodes. Manual adjustment is necessary.

ChenfengWei0: > Direct merging can lead to nesting in some cases

Such as?

src/solidity-frontend/solidity_convert.cpp
prototypeC14: What I mean is that there is a dedicated statement block that contains the path information of the imported files, as shown below.
"""json
"absolutePath": "contract.sol",
    "nodes": [
        {
            "absolutePath": "contract2.sol",
            "file": "./contract2.sol",
            "id": 1,
            "nameLocation": "-1:-1:-1",
            "nodeType": "ImportDirective",
            "scope": 40,
            "sourceUnit": 63,
            "src": "36:25:0",
            "symbolAliases": [],
            "unitAlias": ""
        }
"""
When A imports B and B imports C, the statement block containing the path information for B is stored within A, and similarly, C's path info is stored within B. These blocks can be used to identify the corresponding absolute paths, reducing the need for manual creation. However, since A does not import any files, such a statement block does not exist for it and must be manually added.

prototypeC14: What I mean is that there is a dedicated statement block that contains the path information of the imported files, as shown below.
```json
"absolutePath": "contract.sol",
"nodes": [
{
"absolutePath": "contract2.sol",
"file": "./contract2.sol",
"id": 1,
"nameLocation": "-1:-1:-1",
"nodeType": "ImportDirective",
"scope": 40,
"sourceUnit": 63,
"src": "36:25:0",
"symbolAliases": [],
"unitAlias": ""
}
```
When A imports B and B imports C, the statement block containing the path information for B is stored within A, and similarly, C's path info is stored within B. These blocks can be used to identify the corresponding absolute paths, reducing the need for manual creation. However, since A does not import any files, such a statement block does not exist for it and must be manually added.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: A import B, C ; B import D, E. so direct merging would be like:
```C++
"nodes": [from E.sol]
"nodes": [from D.sol]
"nodes": [from C.sol]
"absolutePath": "E.sol" 
"absolutePath": "D.sol"
"nodes": [from B.sol]
"absolutePath": "C.sol" 
"absolutePath": "B.sol" 
"nodes": [from A.sol]
```

ChenfengWei0:
could you pleas add some Tcs

ChenfengWei0:
@prototypeC14 , The TCs seem not cover the issue in #2388

ChenfengWei0 (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #2394, opened 2025-04-11
[Solidity] Add support for the library entity

prototypeC14:
The purpose of this pull request is to add support for the "library" in Solidity.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1506,6 +1506,20 @@ bool solidity_convertert::get_noncontract_defition(nlohmann::json &ast_node)
...
+  {
+    if (get_struct_class(ast_node))
+      return true;
+    if (convert_ast_nodes(ast_node))
+      return true;
+    // static instance
+    std::string lib_name = ast_node["name"].get<std::string>();
+    symbolt lib_instance;
+    get_static_contract_instance(lib_name, lib_instance);
+    lib_instance.static_lifetime = true;
```
ChenfengWei0: You don't need to specify it's `static`. they are handled in the `get_static_contract_instance`.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1914,6 +1928,40 @@ bool solidity_convertert::get_unbound_expr(
...
+      const auto &ref_decl = find_decl_ref(ref_id);
+
+      if (ref_decl["nodeType"] == "ContractDefinition" &&
+          ref_decl["contractKind"] == "library")
+      {
+        std::string lib_name = ref_decl["name"];
+        std::string instance_name, instance_id;
+        get_static_contract_instance_name(lib_name, instance_name, instance_id);
+
+        const symbolt *lib_sym = context.find_symbol(instance_id);
```
ChenfengWei0:
```
 std::string lib_name = ref_decl["name"];
std::string instance_name, instance_id;
get_static_contract_instance_name(lib_name, instance_name, instance_id);

const symbolt *lib_sym = context.find_symbol(instance_id);
```

Redundant. This is basically `get_static_contract_instance`.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1914,6 +1928,40 @@ bool solidity_convertert::get_unbound_expr(
...
+
+        exprt lib_expr = symbol_expr(*lib_sym);
+        std::string func_name = member_access["memberName"];
+
+        code_function_callt func_call;
+        locationt l;
+        get_location_from_node(expr, l);
+        func_call.location() = l;
+        func_call.function() = member_exprt(lib_expr, func_name, code_typet{});
```
ChenfengWei0: I am not sure if we need to consider the unbound? maybe check how SMTChecker perform. If in unbound, SMTChecker models it as arbitrary external call, then we need to do the same.

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: I feel like you need to improve the get_function_params,  to support `storage`.

ChenfengWei0 (changes requested):

ChenfengWei0:
A library and contract should be convert to something like:


```C++
class CounterLib
{
public:
  static CounterLib lib;

  void increment(int &counter)
  {
    counter += 1;
  }

  int current(const int &counter)
  {
    return counter;
  }
};

CounterLib CounterLib::lib; // get_contract_static_instance

class CounterApp
{
private:
  int counter = 0;

public:
  void incrementCounter()
  {
    CounterLib::lib.increment(counter); // ignore the `CounterLib::` when building the IR
  }

  int getCounter() const
  {
    return CounterLib::lib.current(counter);
  }
};
```

I am not sure if we already support `reference` type. If not, we can simplify convert it to `reference_exprt`.

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -4020,6 +4036,39 @@ bool solidity_convertert::get_expr(
...
+
+        if (
+          decl["nodeType"] == "ContractDefinition" &&
+          decl["contractKind"] == "library")
+        {
+          std::string lib_name = decl["name"];
+          std::string func_name = callee_expr_json["memberName"];
+
+          const auto &func_def = get_func_decl_ref(lib_name, func_name);
+          if (func_def.is_null())
```
ChenfengWei0: || func_def.empty()

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -4020,6 +4036,39 @@ bool solidity_convertert::get_expr(
...
+          {
+            log_error(
+              "Cannot locate function '{}' in library '{}'",
+              func_name,
+              lib_name);
+            return true;
+          }
+
+          side_effect_expr_function_callt call;
+          if (get_non_library_function_call(func_def, expr, call))
```
ChenfengWei0: What if there are arguments in the function call?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1566,6 +1566,22 @@ bool solidity_convertert::get_noncontract_defition(nlohmann::json &ast_node)
     // for abstract contract
     add_empty_body_node(ast_node);
   }
+  else if (
```
ChenfengWei0, line 1624: could you please add some comments?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -1566,6 +1566,22 @@ bool solidity_convertert::get_noncontract_defition(nlohmann::json &ast_node)
...
+    node_type == "ContractDefinition" && ast_node["contractKind"] == "library")
+  {
+    std::string lib_name = ast_node["name"].get<std::string>();
+
+    if (get_struct_class(ast_node))
+      return true;
+    current_baseContractName = lib_name;
```
ChenfengWei0, line 1633: Please avoid overwriting the current_baseContractName before recording it first. 

Maybe something like:

old = current_baseContractName 
current_baseContractName  = xxx;

...

current_baseContractName  = old;

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -4020,6 +4036,39 @@ bool solidity_convertert::get_expr(
       callee_expr_json.contains("nodeType") &&
       callee_expr_json["nodeType"] == "MemberAccess")
     {
+      const auto &base = callee_expr_json["expression"];
+
+      if (base.contains("referencedDeclaration"))
+      {
+        int base_decl_id = base["referencedDeclaration"];
+        const auto &decl = find_decl_ref(src_ast_json, base_decl_id);
```
ChenfengWei0: maybe find_decl_ref_unique_id since the libary cannot be possibly inherited

ChenfengWei0 (changes requested):

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -4038,11 +4038,19 @@ bool solidity_convertert::get_expr(
     {
       const auto &base = callee_expr_json["expression"];
 
-      if (base.contains("referencedDeclaration"))
+      if (base["referencedDeclaration"] > 0)
```
ChenfengWei0: base.contains("referencedDeclaration") && base["referencedDeclaration"] > 0

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: Oh i see you are using get_non_library_function_call. so there should be no issues then

ChenfengWei0:
@prototypeC14 Could you please add some examples using [SafeMath](https://github.com/ConsenSysMesh/openzeppelin-solidity/blob/master/contracts/math/SafeMath.sol) ?

ChenfengWei0:
Also, I would suggest skipping verifying the library entity, so when we do

for(cname : contractNameList)
{
...
// we need to filter the library
}

The filter should be in the same place as the interface or abstract contract,

regression/esbmc-solidity/reentrance_5/test.desc
```diff
@@ -1,4 +1,4 @@
 CORE
 contract.solast
---sol contract.sol --contract Reproduction --incremental-bmc
+--sol contract.sol --contract Reproduction --incremental-bmc --16
```
ChenfengWei0: Do not use `--16`.  Maybe try` --no-standard-checks --reentry-check` instead.

ChenfengWei0: Alternatively, mark it as `KNOWNBUG`

ChenfengWei0 (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #2395, opened 2025-04-11
[python-frontend] Reuse C libm models for NumPy math functions

brcfarias:
This modification enables the reuse of libm models to handle NumPy math functions.
I started with the `ceil` function. Calls such as `numpy.ceil(x) `are now handled by `libm/ceil.c@ceil`, while calls like `numpy.ceil([1, 2, 3]) `are handled by `libm/ceil.c@ceil_array`. This approach makes it easier to support additional math functions.

Additionally, I fixed some issues related to `list` handling.

### DIFF
--- a/src/c2goto/library/libm/ceil.c
+++ b/src/c2goto/library/libm/ceil.c
@@ -26,7 +26,6 @@ ceil_def(long double, ceill, rintl);
 
 #undef ceil_def
 
-
 // This function is used by the Python frontend
 void ceil_array(const double *v, double *out, int size)
 {
--- a/src/c2goto/library/libm/ceil.c
+++ b/src/c2goto/library/libm/ceil.c
@@ -25,3 +25,14 @@ ceil_def(double, ceil, rint);
 ceil_def(long double, ceill, rintl);
 
 #undef ceil_def
+
+void ceil_array(const double *v, double *out, int size)
+{
+__ESBMC_HIDE:;
+  int i = 0;
+  while (i < size)
+  {
+    out[i] = ceil(v[i]);
+    ++i;
+  }
+}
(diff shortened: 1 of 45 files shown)


lucasccordeiro (approved):

lucasccordeiro:

src/clang-c-frontend/clang_c_adjust_code.cpp
```diff
@@ -37,6 +37,19 @@ void clang_c_adjust::adjust_code(codet &code)
   }
   else if (statement == "function_call")
   {
+    for (auto &op : code.operands())
```
intrigus-lgtm, line 40: Why is this only done for function calls and nothing else?
That is, why isn't this case implicitly handled in `adjust_operands(code)`?

src/c2goto/library/libm/ceil.c
```diff
@@ -25,3 +25,15 @@ ceil_def(double, ceil, rint);
 ceil_def(long double, ceill, rintl);
 
 #undef ceil_def
+
+// This function is used by the Python frontend
+void ceil_array(const double *v, double *out, int size)
```
intrigus-lgtm, line 30: Can we please not add functions to headers that aren't mandated by the C standard?
Or if you absolutely want to keep it here, can we prefix them with two underscores `__` so that the name becomes a reserved name?

brcfarias: https://github.com/esbmc/esbmc/pull/2407/commits/3b5ab917b59e30fbd7f6f9b9a8ae1a158cf59aea

src/clang-c-frontend/clang_c_adjust_code.cpp
brcfarias: Thanks for the comment. Not sure yet what the C frontend does in terms of encoding that causes this difference.
I’ll update the code with a comment once I have more details, or make a change to the Python frontend if needed.

PR #2398, opened 2025-04-16
[Solidity] function call with options

ChenfengWei0:
e.g. 
- low-level: `(bool success, ) = msg.sender.call{value: amount}("");`
- high-level: `target.deposit{value: msg.value}();`

Also fixed some bugs within the inheritance algo


lucasccordeiro (approved):

lucasccordeiro:

PR #2400, opened 2025-04-17
[Solidity] Fix function order bug in AST merge

prototypeC14:
This PR:
1. Fixes incorrect function order during inheritance in some cases, where functions appeared in the wrong order when merging ASTs with inheritance.
2. Adds related regression tests.


src/solidity-frontend/solidity_convert.cpp
```diff
@@ -221,6 +221,34 @@ void solidity_convertert::merge_multi_files()
...
+    }
+    import_graph[path] = imports;
+  }
+
+  // Perform topological sorting
+  std::vector<nlohmann::json> sorted_json_files =
+    topological_sort(import_graph, path_to_json);
+
+  // Update order of src_ast_json_array
+  src_ast_json_array = sorted_json_files;
```
ChenfengWei0: Do we still need to reverse the order of the array?

src/solidity-frontend/solidity_convert.cpp
```diff
@@ -259,6 +287,64 @@ void solidity_convertert::merge_multi_files()
       _nodes.push_back(node); // then add each individual node inside the array
   }
 }
+// topological sort is to make sure the order of contract AST is correct(Avoid some counterinstuitive cases)
+// e.g. when contract A import B : contract A AST should be before contract B AST
+std::vector<nlohmann::json> solidity_convertert::topological_sort(
```
ChenfengWei0: Can we return the reference to improve performance? E.g. const &

src/solidity-frontend/solidity_convert.cpp
prototypeC14: Yes, we still need it, and sorting this way has lower overhead.

src/solidity-frontend/solidity_convert.cpp
prototypeC14: Improved as requested.

ChenfengWei0:
I have no idea why the windows build failed. The code itself looks fine to me

src/solidity-frontend/solidity_convert.cpp
ChenfengWei0: I did not see the changes. Maybe you have roll-backed it

lucasccordeiro (approved):

lucasccordeiro:

PR #2416, opened 2025-05-01
[Python]  add chr() built-in function support

lucasccordeiro:
This PR introduces support for Python’s built-in chr() function in the ESBMC Python frontend. The chr() function takes an integer and returns a one-character string corresponding to the ASCII code of that integer.


regression/python/casting13/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+
+^ERROR: TypeError: chr() argument must be int, not number$
```
brcfarias: should be "not float" instead of "not number"?

brcfarias (approved):
LGTM

regression/python/casting13/test.desc
lucasccordeiro: This comes from the JSON file. We could hardcode this error message to include float.

src/python-frontend/function_call_expr.cpp
```diff
@@ -108,6 +108,70 @@ exprt function_call_expr::build_constant_from_arg() const
     double arg_value = arg["value"].get<double>();
     arg["value"] = static_cast<int>(arg_value);
   }
+  else if (func_name == "chr")
+  {
+    int int_value = 0;
+
+    // Handle unary negative constant (e.g., -1)
+    if (arg.contains("_type") && arg["_type"] == "UnaryOp")
```
intrigus-lgtm, line 116: Why has this to be handled here specifically?

src/python-frontend/function_call_expr.cpp
```diff
@@ -108,6 +108,70 @@ exprt function_call_expr::build_constant_from_arg() const
...
+    else
+    {
+      throw std::runtime_error("TypeError: chr() argument must be int");
+    }
+
+    // Range check
+    if (int_value < 0 || int_value > 255)
+    {
+      throw std::runtime_error(
+        "ValueError: chr() argument out of ASCII range: " +
```
intrigus-lgtm, line 168:
```
chr(25633)
'搡'
```
works perfectly fine.
I assume this is just a simplification?
If so, maybe state that this is an ESBMC change?
See https://docs.python.org/3/library/functions.html#chr

lucasccordeiro: Thanks for pointing this out, @intrigus-lgtm.

I have just pushed a commit to fix this: https://github.com/esbmc/esbmc/commit/799263ccf5e122a255794acbe46b3b30866cebc3

src/python-frontend/function_call_expr.cpp
lucasccordeiro: I'm happy to move this to another place. Would you have any suggestions?

intrigus-lgtm: I don't understand the python frontend well enough, so I cannot suggest another place unfortunately.
It just felt a bit weird to handle this explicitly in this place and not in some more general place. But maybe that's perfectly fine for how the python frontend works.

lucasccordeiro: I'm refactoring the code here: https://github.com/esbmc/esbmc/pull/2417

If you have some suggestions, please let me know.

I'm still learning the Python frontend with @brcfarias :-)

PR #2422, opened 2025-05-06
[solidity] fix reentry attack checks

ChenfengWei0:
This PR
- fix reentry checks algo
- fix inheritance
- fix contract address assignment


src/solidity-frontend/solidity_template.h
```diff
@@ -146,20 +147,21 @@ __ESBMC_HIDE:;
   return address_t(hash);
 }
 
-uint256_t _pow(unsigned int base, unsigned int exp) {
```
lucasccordeiro, line 149: Why don't you use our C math functions? https://github.com/esbmc/esbmc/tree/master/src/c2goto/library

lucasccordeiro (changes requested):

src/solidity-frontend/solidity_template.h
ChenfengWei0: image

Yeah I believe we are using the c math now.

lucasccordeiro (approved):

lucasccordeiro:

PR #2431, opened 2025-05-08
[symex] Support for CPROVER R_OK

### LINKED ISSUE #2430: Support for CPROVER R_OK
CBMC has a predicate for checking whether a reading or writing is valid on a memory object. We could have something similar, that could allow us to expand on the transcoder approach and to create explicit assertions over memory operations. In CBMC this is an irep expression but we could do as an intrinsic in ESBMC. 

## Function

- `char __ESBMC_r_ok(void *addr, size_t length)`: This should return True if reading `length` bytes of `addr` will not extrapolate the object that `addr` is pointing to.

### Should pass

```c
int main() {
  char arr[10];
  assert(__ESBMC_r_ok(arr, 0));
}
```

```c
int main() {
  int N, M;  
  char arr[N];
  __ESBMC_assume(N > 0 && M < N); 
  assert(__ESBMC_r_ok(arr, M));
}
```

```c
int main() {
  int N, M;  
  char *arr = malloc(N);
  __ESBMC_assume(N > 0 && M < N); 
  assert(__ESBMC_r_ok(arr, M));
}
```

```c
int main() {
  int N, M;  
  char *arr = malloc(nondet_int() ? N : N+1);
  __ESBMC_assume(N > 0 && M < N); 
  assert(__ESBMC_r_ok(arr, M));
}
```

### Should fail

```c
int main() {
  char arr[10];
  assert(__ESBMC_r_ok(arr, 11));
}
```

```c
int main() {
  int N, M;  
  char arr[N];
  assert(__ESBMC_r_ok(arr, M));
}
```

```c
int main() {
  int N, M;  
  char *arr = malloc(N);
  assert(__ESBMC_r_ok(arr, M));
}
```

```c
int main() {
  int N, M;  
  char *arr = malloc(nondet_int() ? N : N-1);
  __ESBMC_assume(N > 0 && M < N); 
  assert(__ESBMC_r_ok(arr, M));
}
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2430#event-17576778481.

This PR introduces the `__ESBMC_r_ok` intrinsic to ESBMC's symbolic execution engine. This function enables users to verify that a memory read of a specified length from a given address stays within the bounds of the allocated object, improving ESBMC’s ability to check low-level memory safety.


src/goto-symex/symex_main.cpp
```diff
@@ -798,6 +798,61 @@ void goto_symext::run_intrinsic(
...
+
+    internal_deref_items.clear();
+
+    // Perform an internal dereference to resolve the base object being pointed to
+    expr2tc deref = dereference2tc(get_empty_type(), addr);
+    dereference(deref, dereferencet::INTERNAL);
+
+    assert(!internal_deref_items.empty());
+
+    expr2tc base_obj = internal_deref_items.front().object;
```
rafaelsamenezes: You are only computing the first element here. What if it is a pointer to multiple places?

rafaelsamenezes (changes requested):
The test cases should run with --no-pointer-check

lucasccordeiro:
> The test cases should run with --no-pointer-check

Implemented as suggested.

src/goto-symex/symex_main.cpp
lucasccordeiro: You're correct. I have just fixed it!

rafaelsamenezes (approved):

src/goto-symex/symex_main.cpp
```diff
@@ -798,6 +798,75 @@ void goto_symext::run_intrinsic(
...
+        continue;
+
+      base_size = sub2tc(base_size->type, base_size, item.offset);
+      expr2tc lower_bound = lessthanequal2tc(zero, cast_len_expr);
+      expr2tc upper_bound = lessthanequal2tc(cast_len_expr, base_size);
+      expr2tc check = and2tc(lower_bound, upper_bound);
+      expr2tc result = implies2tc(item.guard, check);
+
+      final_result =
+        is_nil_expr(final_result) ? result : and2tc(final_result, result);
```
intrigus-lgtm, line 861: When is `final_result` ever nil?

intrigus-lgtm:
I don't quite understand, why `regression/esbmc/r_ok14/main.c` is supposed to succeed.
If we translate the testcase and use godbolt with address sanitizer, we get a crash:
https://godbolt.org/z/e95bdx8sn

So why is this acceptable in ESBMC?

intrigus-lgtm:
I guess this has fallen through the cracks after being merged, but @lucasccordeiro could you please have a look at my comments?

lucasccordeiro:
Hi @intrigus-lgtm,

I'm sorry I didn't respond to your comment earlier.

I have checked this test case with @rafaelsamenezes and @feliperodri; they both confirm that this is supposed to succeed.

@rafaelsamenezes: Could you please give further insights here?

rafaelsamenezes:
> I don't quite understand, why `regression/esbmc/r_ok14/main.c` is supposed to succeed.
> 
> If we translate the testcase and use godbolt with address sanitizer, we get a crash:
> 
> https://godbolt.org/z/e95bdx8sn
> 
> 
> 
> So why is this acceptable in ESBMC?

Your test case is trying to access the sixth position of the array, which is an OOB.

The original test was checking whether "is fine to read 5 bytes starting from the base address".

intrigus-lgtm:
> > I don't quite understand, why `regression/esbmc/r_ok14/main.c` is supposed to succeed.
> > If we translate the testcase and use godbolt with address sanitizer, we get a crash:
> > https://godbolt.org/z/e95bdx8sn
> > So why is this acceptable in ESBMC?
> 
> Your test case is trying to access the sixth position of the array, which is an OOB.
> 
> The original test was checking whether "is fine to read 5 bytes starting from the base address".

Ah, that makes sense.
I guess I was confused by the issue description that has a similar example which you wrote should fail:
https://github.com/esbmc/esbmc/issues/2430#issue-3049469155

> Should fail
>
>```
>int main() {
>  char arr[10];
>  assert(__ESBMC_r_ok(arr, 10));
>}
>```

rafaelsamenezes:
> > > I don't quite understand, why `regression/esbmc/r_ok14/main.c` is supposed to succeed.
> > > If we translate the testcase and use godbolt with address sanitizer, we get a crash:
> > > https://godbolt.org/z/e95bdx8sn
> > > So why is this acceptable in ESBMC?
> > 
> > 
> > Your test case is trying to access the sixth position of the array, which is an OOB.
> > The original test was checking whether "is fine to read 5 bytes starting from the base address".
> 
> Ah, that makes sense. I guess I was confused by the issue description that has a similar example which you wrote should fail: [#2430 (comment)](https://github.com/esbmc/esbmc/issues/2430#issue-3049469155)
> 
> > Should fail
> > ```
> > int main() {
> >  char arr[10];
> >  assert(__ESBMC_r_ok(arr, 10));
> > }
> > ```

Ow, thanks for showing. I fixed the comment.

PR #2442, opened 2025-05-13
[libm] fix acos function

### LINKED ISSUE #2439: [libm] investigate acos from libm
These assertions should hold:

````C
#include <math.h>
#include <assert.h>

int main() {
    assert(acos(1.0) >= 0);
    assert(acos(0.0) >= 1.57);
    return 0;
}
````

XLiZHI:
fix #2439

image
https://en.wikipedia.org/wiki/Inverse_trigonometric_functions


emanino:
> fix #2439
> 
> image https://en.wikipedia.org/wiki/Inverse_trigonometric_functions

This approximation of arccos is only valid for positive inputs, i.e. 0 < x < 1, but it fails for negative inputs. For a visual check, I find it very useful to drop the difference equation on Wolfram Alpha as follows:
https://www.wolframalpha.com/input?i=plot+arccos%28x%29+-+arctan%28sqrt%281-x*x%29%2Fx%29+from+-1+to+1

I think a quick fix would be to subtract 3 for negative inputs. However, these sort of approximations will always violate the exact rounding requirements of the IEEE754 standard. For floating-point inputs (32 bit), we made the choice of preferring bit-precise operational models in ESBMC. I would suggest reading [this paper](https://hal.science/hal-04474530/document) for a bit of background on the topic.

lucasccordeiro:
@emanino: Thanks for your feedback.

Let me digest this material. I'll submit a few PRs to ensure our math models are sound, and I'll keep you as a reviewer.

@XLiZHI: Thanks for sending this patch.

XLiZHI:
Hi @emanino, 

Thank you for your feedback, in order to avoid the problem that arctan cannot distinguish quadrants, I used `atan2` instead of `arctan` in this PR, I think it should be enough to handle the case of x ∈ [-1, 1], if I understand correctly.

emanino:
> Hi @emanino,
> 
> Thank you for your feedback, in order to avoid the problem that arctan cannot distinguish quadrants, I used `atan2` instead of `arctan` in this PR, I think it should be enough to handle the case of x ∈ [-1, 1], if I understand correctly.

You are correct, atan2 should work for negative numbers.

PR #2443, opened 2025-05-13
CPROVER migration compatibility

rafaelsamenezes:
In this PR I am introducing ireps that were originated from CBMC into the migration. However, I am focusing on implementing them as intrinsic function calls. As we can then test the behavior using C files.

This is still a draft as it requires:
- [x] ~There is a bug in the transcoder that is making the pop_frame assertion break.~ Applying same hack as coverage.
- [X] ~Overflow-result needs to invoke our intrinsic function.~ Can't apply late goto transformations.
- [X]  ~r_ok intrinsic. ~ Will fix in another PR
- [X] .goto testcases need to be constructed from CBMC and then adapted to ESBMC format (using goto-transcoder). 
- [x] ~Export header of ESBMC. This is a C file including all the intrinsic function signatures, this is meant to be exported as a .goto file.~ Testing utilities will be ported to utils (#2476).


lucasccordeiro (approved):
@rafaelsamenezes: could you please run this PR over the sv-comp benchmarks?

lucasccordeiro:
Master (30s):

````
  correct:           17618
    correct true:    10337
    correct false:    7281
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15916
  Score:             27171 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15107109902
````

This PR (30s):

````
  correct:           17630
    correct true:    10343
    correct false:    7287
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15904
  Score:             27189 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15307201671
````

PR #2447, opened 2025-05-14
[CI] Integrate cheribsd-riscv64 system root

XLiZHI:
The build on GitHub runner using cheribuild took too long and exceeded the runner's hard disk capacity.

Currently I have uploaded the cheribsd SDK built locally and hope it works as expected. ;)

### DIFF
--- a/.github/workflows/build-unix.yml
+++ b/.github/workflows/build-unix.yml
@@ -21,13 +21,6 @@ jobs:
     runs-on: ${{ inputs.operating-system }}
     steps:
     - uses: actions/checkout@v4
-    - name: Add apt.llvm.org repository
-      run: |
-        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc &&
-        source /etc/os-release &&
-        sudo add-apt-repository -y "deb http://apt.llvm.org/${UBUNTU_CODENAME}/ llvm-toolchain-${UBUNTU_CODENAME}-16 main" &&
-        sudo apt-get update
-      if: ${{ startsWith(inputs.operating-system, 'ubuntu-') }}
     - name: Build ESBMC
       run: ./scripts/build.sh ${{ inputs.build-flags }}
     - uses: actions/upload-artifact@v4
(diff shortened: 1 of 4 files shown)


src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -44,6 +44,8 @@ void clang_c_languaget::build_include_args(
   {
     compiler_args.push_back("-isystem");
     compiler_args.push_back(*libc_headers);
+    compiler_args.push_back("-Wno-implicit-function-declaration");
+    compiler_args.push_back("-Wno-invalid-noreturn");
```
intrigus-lgtm: Why is this is necessary?
This also affects user code, doesn't it?

src/c2goto/library/pthread_lib.c
```diff
@@ -141,7 +141,7 @@ __ESBMC_HIDE:;
     __ESBMC_thread_key *l = search_key(i);
     if (__ESBMC_thread_key_destructors[i] && l->value)
     {
-      __ESBMC_thread_key_destructors[i](l->value);
+      __ESBMC_thread_key_destructors[i]((void *)l->value);
```
intrigus-lgtm: Would it make sense to change the definition of `__ESBMC_thread_key_destructors` so that it expects a `const void*`?
https://github.com/esbmc/esbmc/blob/ad37d0263963e8ef60916246de03426de108790e/src/c2goto/library/pthread_lib.c#L43C3-L43C33

src/c2goto/library/builtin_libs.c
```diff
@@ -28,6 +28,8 @@
 
 #    include <cheri/cheric.h>
 
+__SIZE_TYPE__ get_all_permissions(const cc128_cap_t *cap);
```
intrigus-lgtm: Shouldn't this function be declared with two leading underscores?
Alternatively, shouldn't it be inside the cheri ifdef?

src/c2goto/library/pthread_lib.c
lucasccordeiro: Thanks, @intrigus-lgtm. I have just submitted a PR to fix these warning messages: https://github.com/esbmc/esbmc/pull/2448

lucasccordeiro (approved):

lucasccordeiro:

PR #2448, opened 2025-05-15
[OM] Fix warning messages for pthread and setjmp models

lucasccordeiro:


src/c2goto/library/pthread_lib.c
```diff
@@ -141,7 +147,7 @@ __ESBMC_HIDE:;
     __ESBMC_thread_key *l = search_key(i);
     if (__ESBMC_thread_key_destructors[i] && l->value)
     {
-      __ESBMC_thread_key_destructors[i](l->value);
+      __ESBMC_thread_key_destructors[i]((void *)l->value);
```
intrigus-lgtm, line 150: Why is `l->value` `const` in the first place?

src/c2goto/library/pthread_lib.c
```diff
@@ -53,6 +53,12 @@ pthread_t __ESBMC_get_thread_id(void);
 void __ESBMC_really_atomic_begin(void);
 void __ESBMC_really_atomic_end(void);
 
+int pthread_id_to_index(pthread_t t)
+{
+  // Here we assume pthread_t is just a small integer type.
+  return (int)(uintptr_t)t;
```
intrigus-lgtm: Isn't this in contrast to:
https://github.com/esbmc/esbmc/blob/a3f1dcf7f4d4d279599d3bc66924332628213ce4/src/c2goto/headers/pthread.h#L41

lucasccordeiro: Indeed, I have just fixed it.

src/c2goto/library/pthread_lib.c
lucasccordeiro: This is because of this: https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_setspecific.html. While pthread_key_create requires this https://linux.die.net/man/3/pthread_key_create.

PR #2449, opened 2025-05-15
[CI] migrated to Ubuntu 24.04

XLiZHI:
This PR tries to build ESBMC on Ubuntu:

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

`GNU C++: 13.3.0`: Brings stricter type checking -- Some python frontend changes had to be made to remove the warning!

`error: possibly dangling reference to a temporary [-Werror=dangling-reference]`
`error: redundant move in return statement [-Werror=redundant-move]`

`Python 3.12.3`: The build script for IBEX uses the imp module, which was removed in Python 3.12  -- use cmake to build the latest IBEX.

`boots 1.83` : It seems like a new compression algorithm was introduced and we need a new dependency: l`ibbz2-dev liblzma-dev`

Some of the standard libraries have also changed (GCC 14), and some of the OMs need to be modified.

Given that the majority of users are developing in 22.04, is it necessary to adapt to 24.04?


lucasccordeiro (approved):

lucasccordeiro:
I'm happy with this PR. We should always use the latest version of the tools.

ChenfengWei0 (approved):

Anthonysdu (approved):

scripts/cmake/Irep2Optimization.cmake
```diff
@@ -2,10 +2,8 @@
 # This is required by MSVC
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.0.0")
```
rafaelsamenezes, line 5: Is this a problem on newer versions as well?

rafaelsamenezes (changes requested):

scripts/cmake/Irep2Optimization.cmake
XLiZHI: Yes, boost has a lot of warnings that cmake treats as errors. It seemed uncertain to me that the preprocessed headers would improve gcc compilation speed, so I revert the code anyway.

rafaelsamenezes: Yeah, the preprocessed headers are needed more because of MSVC limits  rather than optimisation. It's fine to skip them.

PR #2461, opened 2025-05-21
[goto-check] enable ub shift check by default

XLiZHI:


lucasccordeiro (changes requested):
Thanks, @XLiZHI.

Can you check our sv-comp script and the impact of this change for the sv-comp benchmarks?

XLiZHI:
> Thanks, @XLiZHI.
> 
> Can you check our sv-comp script and the impact of this change for the sv-comp benchmarks?

We didn't use this flag in the script (or we need to disable this check), and I will run sv-comp to confirm.

lucasccordeiro:
> We didn't use this flag in the script (or we need to disable this check), and I will run sv-comp to confirm.

Many thanks, @XLiZHI.

PR #2464, opened 2025-05-21
[CHERI] support the verification of CHERI capability bounds

XLiZHI:
In this PR:

1. A struct type infinite array is introduced to store CHERI capabilities, which will be easier to extend.

2. The `capabilities_base2tc` and `capabilities_top2tc` expressions were introduced to normalize the ability to retrieve from arrays.

3. The pointer type is extended to introduce a new variable to store the `can_carry_provenace` information.

4. Extend the dereference module: have added the judgment of whether the pointer has capabilities, supported CHERI bounds in the bounds check module.

I tried to improve some comments to support this PR, maybe this PR does not need to be broken into several small ones.


lucasccordeiro (changes requested):
@XLiZHI: could you please add some regression tests?

src/pointer-analysis/dereference.cpp
```diff
@@ -747,7 +747,10 @@ expr2tc dereferencet::build_reference_to(
   }
   else if (is_array_type(value)) // Encode some access bounds checks.
   {
-    bounds_check(value, final_offset, type, tmp_guard);
+    bool can_carry = is_pointer_type(deref_expr) &&
+                     to_pointer_type(deref_expr->type).carry_provenance;
+    expr2tc tmp_expr = can_carry ? deref_expr : expr2tc();
+    bounds_check(value, final_offset, type, tmp_guard, tmp_expr);
```
intrigus-lgtm, line 753: Why not move this to a `cheri_capability_bound_check` method, so `bounds_check` doesn't have to be changed and there isn't a mostly unused default parameter?

XLiZHI: `bounds_check` isn't just called in this one place, as we discussed earlier, and we want to be consistent. Bounds check is handled in the same function to help with scaling.

intrigus-lgtm: Hmm, but all the other callers of the function don't care about cheri bounds checking, because only a single call to `bound_check` has been changed in this PR.

lucasccordeiro (approved):

lucasccordeiro:

PR #2477, opened 2025-05-27
[symex] removed temporary fix for the condition coverage

lucasccordeiro:


ChenfengWei0:
I just ran the ctest locally, and it seems the removal of the temporary fix did not break anything. So I guess it's an old fix that forgot to remove. Actually I cannot remember why we need this.

rafaelsamenezes:
> I just ran the ctest locally, and it seems the removal of the temporary fix did not break anything. So I guess it's an old fix that forgot to remove. Actually I cannot remember why we need this.

Can you test it on testcomp benchmarks? It might be triggered by something in there

ChenfengWei0:
> > I just ran the ctest locally, and it seems the removal of the temporary fix did not break anything. So I guess it's an old fix that forgot to remove. Actually I cannot remember why we need this.
> 
> Can you test it on testcomp benchmarks? It might be triggered by something in there

I am not sure how to do that. Is there a benchexec option for GitHub action?

I just re-read the code. The fix was seemingly introduced because, at the time, I didn't know about `goto_functions.update()`. Since we now run that update step, as long as we insert new instrumentation before the update, it should not trigger the internal assertion check failures anymore.

PR #2479, opened 2025-05-28
[C verification] Adjust the behavior of parameter checking for the unknown method

XLiZHI:
please see the code comments.


lucasccordeiro (changes requested):
Can you document better this PR? What is the goal? 

Can you also run this PR over the SV-COMP benchmarks?

XLiZHI:
> Can you document better this PR? What is the goal?
> 
> Can you also run this PR over the SV-COMP benchmarks?

This PR is a modification of the previous implementation (#2329). 

The purpose of this implementation is: we will check for unknown functions, first of all, these unknown functions may have undefined behavior in reality after being called, because the inside of the function is not visible. So one checking strategy is that we no longer trust the pointer type arguments used to call the function and mark them as invalid. We would then violate the assertion when dereferencing these pointers.

This is an option and I believe it will not have any impact on sv-comp.

lucasccordeiro:

PR #2480, opened 2025-05-28
[python] add support for Is and IsNot operators

lucasccordeiro:
This PR extends ESBMC’s Python frontend to support `Is` and `IsNot`.

Documentation: https://docs.python.org/3/reference/expressions.html#is


src/python-frontend/python_converter.cpp
```diff
@@ -871,6 +872,50 @@ exprt python_converter::handle_string_operations(
...
+  else if (lhs.type().is_array() && rhs.type().is_array())
+  {
+    // Compare base addresses of the arrays
+    is_expr.copy_to_operands(
+      get_array_base_address(lhs), get_array_base_address(rhs));
+  }
+  else
+  {
+    // Default identity comparison
+    is_expr.copy_to_operands(lhs, rhs);
```
brcfarias, line 897: We have the same block here and at line 886. Perhaps it could be simplified.

lucasccordeiro: Good catch! Let me submit a small PR.

PR #2483, opened 2025-05-29
[build] fix solvers gmp library link

XLiZHI:
We have multiple solvers that depend on the gmp library at the same time, perhaps putting the gmp lib link at the top level is better.

https://github.com/esbmc/esbmc/actions/runs/15322131655/job/43108043486

PR #2485, opened 2025-05-29
Revert "[build] fix solvers gmp library link"

lucasccordeiro:
Reverts esbmc/esbmc#2483


rafaelsamenezes:
I don't think that it should be reverted. It might be just some extra dependency missing.

PR #2488, opened 2025-05-30
[build] fix solvers build

XLiZHI:


rafaelsamenezes (changes requested):
Can you provide a bit more context?

XLiZHI:
When building locally, we use DIR to specify the location of the solver. So we need to pass the ENABLE_*** parameter to the parent layer to make sure the GMP library gets linked.

https://github.com/esbmc/esbmc/blob/84384f3ae7c67deee8e23c1751d0a750c2c2a1ad/src/solvers/CMakeLists.txt#L36

@lucasccordeiro, Did this PR solve your build error?

PR #2489, opened 2025-05-31
[python] add support for numpy.dot() with 1D and mixed-dimension arrays

lucasccordeiro:
This PR extends our Python frontend to support additional `numpy.dot()` operand combinations, enabling verification of vector and matrix dot products across a wider range of dimensions.

- 1D × 1D → scalar
- 1D × 2D → 1D array
- 2D × 1D → 1D array
- 2D × 2D → 2D array (already supported)


src/python-frontend/numpy_call_expr.cpp
```diff
@@ -396,32 +396,132 @@ exprt numpy_call_expr::create_expr_from_call()
...
       {
-        size_t m = lhs["elts"].size();
-        size_t n = lhs["elts"][0]["elts"].size();
-        size_t n2 = rhs["elts"].size();
-        size_t p = rhs["elts"][0]["elts"].size();
+        // Helper function to check if array is 2D (has nested elts)
+        auto is_2d_array = [](const nlohmann::json &arr) {
+          return arr["_type"] == "List" && !arr["elts"].empty() &&
+                 arr["elts"][0].contains("elts");
+        };
```
brcfarias: Perhaps this could be moved to `type_handler` class?
Also I'm not sure how 3d+ arrays are encoded, but if it contains "elts" at the second level, could we detect them as 2d? If that is the case, I would suggest using the return from type_handler::get_typet() and checking if the `subtype()` is an array, and subtype().subtype() is not.

lucasccordeiro: > Perhaps this could be moved to type_handler class?

Implemented as suggested.

PR #2496, opened 2025-06-01
[regression] Show warning when KNOWNBUG tests stop failing

brcfarias:
**Before:**

image

**Now:**

image

### DIFF
--- a/regression/numpy/det2/test.desc
+++ b/regression/numpy/det2/test.desc
@@ -1,4 +1,3 @@
 CORE
 main.py
---bitwuzla
 ^VERIFICATION FAILED$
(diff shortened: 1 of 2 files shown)


rafaelsamenezes (changes requested):
Hm, when designing this I made it fail to force people converting their known bugs into core.

Is there any reason to avoid forcing this?

brcfarias:
The behaviour is basically the same. It still fails but the new log makes it clearer what happened. This can be especially useful for new devs.

rafaelsamenezes:
> The behaviour is basically the same. It still fails but the new log makes it clearer what happened. This can be especially useful for new devs.

Ah I see. That's ok then.

lucasccordeiro (approved):

lucasccordeiro:

PR #2498, opened 2025-06-02
[smt_conv] replace reserve with resize

rafaelsamenezes:
Some mistake, `reserve` does not preallocate the vector. So I am changing to `resize`.

### DIFF
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -302,7 +302,7 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
   }
 
   std::vector<smt_astt> args;
-  args.reserve(expr->get_num_sub_exprs());
+  args.resize(expr->get_num_sub_exprs());
 
   switch (expr->expr_id)
   {


lucasccordeiro (approved):

intrigus-lgtm:
> void reserve( size_type new_cap );
(constexpr since C++20)
Increase the capacity of the vector (the total number of elements that the vector can hold **without requiring reallocation**) to a value that's greater or equal to new_cap.

I don't understand your reasoning. Reserve _does_ preallocate the necessary memory.
Or what do you mean with preallocation?
While resize now fills the vector with default constructed elements which probably isn't what we want here or is it?

rafaelsamenezes:
> > void reserve( size_type new_cap );
> > (constexpr since C++20)
> > Increase the capacity of the vector (the total number of elements that the vector can hold **without requiring reallocation**) to a value that's greater or equal to new_cap.
> 
> I don't understand your reasoning. Reserve _does_ preallocate the necessary memory. Or what do you mean with preallocation? While resize now fills the vector with default constructed elements which probably isn't what we want here or is it?

Reserve has no guarantees of allocating anything. It is just giving some information about the size to expect. 

For example. On `g++ (GCC) 15.1.1 20250425`

```cpp
#include <vector>
int main() {
  std::vector<int> vec;
  vec.reserve(10);
  vec[1] = 42;
}
```

Running it will result in:

```
/usr/include/c++/15.1.1/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = int; _Alloc = std::allocator<int>; reference = int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted (core dumped)
```

intrigus-lgtm:
> > > void reserve( size_type new_cap );
> > > (constexpr since C++20)
> > > Increase the capacity of the vector (the total number of elements that the vector can hold **without requiring reallocation**) to a value that's greater or equal to new_cap.
> > 
> > 
> > I don't understand your reasoning. Reserve _does_ preallocate the necessary memory. Or what do you mean with preallocation? While resize now fills the vector with default constructed elements which probably isn't what we want here or is it?
> 
> Reserve has no guarantees of allocating anything. It is just giving some information about the size to expect.
> 
> For example. On `g++ (GCC) 15.1.1 20250425`
> 
> ```c++
> #include <vector>
> int main() {
>   std::vector<int> vec;
>   vec.reserve(10);
>   vec[1] = 42;
> }
> ```
> 
> Running it will result in:
> 
> ```
> /usr/include/c++/15.1.1/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = int; _Alloc = std::allocator<int>; reference = int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
> Aborted (core dumped)
> ```

Yes, if we do it like this (`vec[1] = ...`) it will fail.
But `vec.reserve(10)` DOES allocate something.
It allocates space for `int`s, so the capacity of the vector changes to `10` while the _size_ is still 0.
`vec.resize(10)` DOES also allocate something, but it fills the memory with default-constructed elements.
So `capacity` and `size` become 10.

But in our case it's IMHO useless to construct all those default elements, because we will immediately overwrite them here:
https://github.com/esbmc/esbmc/blob/146404a91062ed314f96433ac3d39e795c60450d/src/solvers/smt/smt_conv.cpp#L331

So instead of changing from reserve to resize, we should simply change this loop to use `push_back` instead of explicitly accessing the vector by index.

```cpp
expr->foreach_operand(
      [this, &args](const expr2tc &e) { args.push_back(convert_ast(e)); });
}
```

Here's also a small godbolt example:
https://godbolt.org/z/q46n4sq5e

rafaelsamenezes:
> > > > void reserve( size_type new_cap );
> > > > (constexpr since C++20)
> > > > Increase the capacity of the vector (the total number of elements that the vector can hold **without requiring reallocation**) to a value that's greater or equal to new_cap.
> > > 
> > > 
> > > I don't understand your reasoning. Reserve _does_ preallocate the necessary memory. Or what do you mean with preallocation? While resize now fills the vector with default constructed elements which probably isn't what we want here or is it?
> > 
> > 
> > Reserve has no guarantees of allocating anything. It is just giving some information about the size to expect.
> > For example. On `g++ (GCC) 15.1.1 20250425`
> > ```c++
> > #include <vector>
> > int main() {
> >   std::vector<int> vec;
> >   vec.reserve(10);
> >   vec[1] = 42;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Running it will result in:
> > ```
> > /usr/include/c++/15.1.1/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = int; _Alloc = std::allocator<int>; reference = int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
> > Aborted (core dumped)
> > ```
> 
> Yes, if we do it like this (`vec[1] = ...`) it will fail. But `vec.reserve(10)` DOES allocate something. It allocates space for `int`s, so the capacity of the vector changes to `10` while the _size_ is still 0. `vec.resize(10)` DOES also allocate something, but it fills the memory with default-constructed elements. So `capacity` and `size` become 10.
> 
> But in our case it's IMHO useless to construct all those default elements, because we will immediately overwrite them here:
> 
> https://github.com/esbmc/esbmc/blob/146404a91062ed314f96433ac3d39e795c60450d/src/solvers/smt/smt_conv.cpp#L331
> 
> So instead of changing from reserve to resize, we should simply change this loop to use `push_back` instead of explicitly accessing the vector by index.
> 
> ```c++
> expr->foreach_operand(
>       [this, &args](const expr2tc &e) { args.push_back(convert_ast(e)); });
> }
> ```
> 
> Here's also a small godbolt example: https://godbolt.org/z/q46n4sq5e

The argument loop will only fix some of the default cases. The rest of `smt_conv` still assume that they can access any position of the args.

Now, I could:

1. Rewrite the full function ensuring that an element exists before accessing it.
2. Use std::array.
3. Replace a reserve with a resize.

intrigus-lgtm:
If you don't mind, I would like to think about this a bit more.

lucasccordeiro:
This PR (30s):

````
  correct:           17694
    correct true:    10387
    correct false:    7307
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15840
  Score:             27297 (max: 55885)
  
https://github.com/esbmc/esbmc/actions/runs/15390397258
````

Master (30s):

````
  correct:           17639
    correct true:    10350
    correct false:    7289
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15895
  Score:             27205 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15319973264
````

lucasccordeiro:
I'll close this PR since @intrigus-lgtm has proposed a better approach to handle this issue in https://github.com/esbmc/esbmc/pull/2502.

PR #2500, opened 2025-06-02
[goto-programs] add validation for invalid member accesses on incomplete or incorrectly typed expressions (Fixes #2236)

### LINKED ISSUE #2236: ESBMC 7.8 segmentation fault
The following program lead to a segmentation fault using ESBMC 7.8:

**command with --unwind 1**:  `bin/esbbin/esbmc --no-unwinding-assertions --unwind 1 DATASETv2/gpt35-27570.c ` --> **VERIFICATION SUCCESSFUL**

**command without unwind**:  `bin/esbmc DATASETv2/gpt35-27570.c ` --> **Segmentation fault (core dumped)**
OUTPUT:
```
ESBMC version 7.8.0 64-bit x86_64 linux
Parsing DATASETv2/gpt35-27570.c
Converting
Generating GOTO Program
GOTO program creation time: 0.105s
Unwinding loop 35 iteration 1   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 2   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 3   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 4   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 5   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 6   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 7   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 8   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 9   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
Unwinding loop 35 iteration 10   file DATASETv2/gpt35-27570.c line 35 column 3 function initialize_clients
WARNING: no body for function socket
WARNING: no body for function perror
WARNING: no body for function bind
WARNING: no body for function perror
WARNING: no body for function listen
WARNING: no body for function perror
WARNING: no body for function accept
WARNING: no body for function perror
```

PROGRAM:

```c
//FormAI DATASET v1.0 Category: Socket programming ; Style: protected
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>
#include <string.h>
#include <unistd.h>
#include <pthread.h>
#include <errno.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>

#define PORT 8888
#define MAX_CONNECTIONS 10

// Define buffer size for incoming messages
#define BUFFER_SIZE 1024

// Define a mutex to protect shared variables
pthread_mutex_t mutex;

// Define a struct to hold client information
struct client {
  int socket;
  struct sockaddr_in address;
  bool is_active;
};

// Define an array of client structs for tracking connections
struct client clients[MAX_CONNECTIONS];

// Initialize all client structs
void initialize_clients() {
  int i;
  for (i = 0; i < MAX_CONNECTIONS; i++) {
    clients[i].socket = -1;
    clients[i].is_active = false;
  }
}

// Add a new client to the client array
void add_client(int socket, struct sockaddr_in address) {
  int i;
  for (i = 0; i < MAX_CONNECTIONS; i++) {
    // If the slot is empty, add the client
    if (clients[i].socket == -1) {
      clients[i].socket = socket;
      clients[i].address = address;
      clients[i].is_active = true;
      break;
    }
  }
}

// Remove a client from the client array
void remove_client(int socket) {
  int i;
  for (i = 0; i < MAX_CONNECTIONS; i++) {
    if (clients[i].socket == socket) {
      clients[i].socket = -1;
      clients[i].is_active = false;
      break;
    }
  }
}

// Handle incoming messages from a client
void handle_client(int socket) {
  char buffer[BUFFER_SIZE];
  int bytes_received;

  // Receive messages and broadcast to all clients
  while (true) {
    bytes_received = recv(socket, buffer, BUFFER_SIZE, 0);
    if (bytes_received == -1) {
      perror("Error receiving message");
      break;
    } else if (bytes_received == 0) {
      printf("Client disconnected\n");
      break;
    } else {
      // Lock the mutex to protect shared variables
      pthread_mutex_lock(&mutex);

      // Broadcast message to all active clients
      int i;
      for (i = 0; i < MAX_CONNECTIONS; i++) {
        if (clients[i].is_active && clients[i].socket != socket) {
          send(clients[i].socket, buffer, bytes_received, 0);
        }
      }

      // Unlock the mutex
      pthread_mutex_unlock(&mutex);
    }
  }

  // Close the socket and remove the client
  close(socket);
  remove_client(socket);
}

// Thread function for handling new client connections
void *connection_handler(void *socket_ptr) {
  int new_socket = *(int *) socket_ptr;

  // Add the new client to the client array
  add_client(new_socket, clients->address);

  // Handle messages from the client
  handle_client(new_socket);

  return 0;
}

int main(int argc, char const *argv[]) {
  int server_socket, new_socket;
  struct sockaddr_in address;
  int addr_len = sizeof(address);
  pthread_t thread_id;

  // Initialize clients array
  initialize_clients();

  // Create a server socket
  if ((server_socket = socket(AF_INET, SOCK_STREAM, 0)) == 0) {
    perror("Error creating socket");
    exit(EXIT_FAILURE);
  }

  // Bind the server socket to a port
  address.sin_family = AF_INET;
  address.sin_addr.s_addr = INADDR_ANY;
  address.sin_port = htons(PORT);
  if (bind(server_socket, (struct sockaddr *)&address, sizeof(address)) < 0) {
    perror("Error binding socket");
    exit(EXIT_FAILURE);
  }

  // Listen for incoming connections
  if (listen(server_socket, MAX_CONNECTIONS) < 0) {
    perror("Error listening for connections");
    exit(EXIT_FAILURE);
  }

  // Accept incoming connections and spawn a thread to handle each one
  while (true) {
    if ((new_socket = accept(server_socket, (struct sockaddr *)&address, (socklen_t *)&addr_len)) < 0) {
      perror("Error accepting connection");
      exit(EXIT_FAILURE);
    }

    if (pthread_create(&thread_id, NULL, connection_handler, &new_socket) < 0) {
      perror("Error creating thread");
      exit(EXIT_FAILURE);
    }
  }

  // Close the server socket
  close(server_socket);

  return 0;
}
```

lucasccordeiro:
The best solution might be to fix this in our clang-c frontend. @XLiZHI: could you please check it?

If you don't find a solution, this PR resolves [Issue #2236](https://github.com/esbmc/esbmc/issues/2236), which caused crashes when performing member access (e.g., b->a) on improperly defined or incomplete types such as arrays of anonymous structs. The fix introduces validation of member expressions during the goto_convert phase to ensure only semantically valid member accesses are migrated into the GOTO program.


XLiZHI:
If I understand correctly, this PR will not convert (b->a) but will ignore it, right?

`void c() { b->a;  }`

I think that here, b would degenerate into a pointer to the 0th element, which is equivalent to (&b[0])->a). We should adjust it.

lucasccordeiro:
@XLiZHI: Please feel free to push your suggestions here or open a new pull request to fix this issue properly.

intrigus-lgtm:
We always ignore any array to pointer decay casts that clang generates for us:
https://github.com/esbmc/esbmc/blob/bb01667c984f10ba6b326e4d85a23a004bdfa08d/src/clang-c-frontend/clang_c_convert.cpp#L2910-L2914
and just return the original expression unchanged:
https://github.com/esbmc/esbmc/blob/bb01667c984f10ba6b326e4d85a23a004bdfa08d/src/clang-c-frontend/clang_c_convert.cpp#L2979

This then causes IR like this:
<details>

```
Symbol......: c:@F@c_func
Base name...: c_func
Module......: foo
Mode........: C (C)
Type........: code
      * arguments: 
      * return_type: empty
          * #cpp_type: void
Value.......: code
      * type: code
      * operands: 
        0: code
            * type: code
            * operands: 
              0: member
                  * type: signedbv
                      * width: 32
                      * #cpp_type: signed_int
                  * operands: 
                    0: symbol
                        * type: array
                            * size: constant
                                * type: unsignedbv
                                    * width: 64
                                * value: 0000000000000000000000000000000000000000000000000000000000000001
                                * #cformat: 1
                            * subtype: struct
                                * tag: struct __anon_struct_at_foo_c__1_1
                                * components: 
                                  0: component
                                      * type: signedbv
                                          * width: 32
                                          * #cpp_type: signed_int
                                      * name: a
                                      * pretty_name: a
                                      * #location: 
                                        * file: foo.c
                                        * line: 3
                                        * column: 3
                                * bases: 
                                * #location: 
                                  * file: foo.c
                                  * line: 1
                                  * column: 1
                                * #is_layouted: 1
                        * name: explode
                        * identifier: c:@explode
                        * #location: 
                          * file: foo.c
                          * line: 5
                          * function: c_func
                          * column: 17
                        * #lvalue: 1
                  * component_name: a
                  * #location: 
                    * file: foo.c
                    * line: 5
                    * function: c_func
                    * column: 17
            * statement: expression
            * #location: 
              * file: foo.c
              * line: 5
              * function: c_func
              * column: 17
      * statement: block
      * #location: 
        * file: foo.c
        * line: 5
        * function: c_func
        * column: 15
      * #end_location: 
        * file: foo.c
        * line: 5
        * function: c_func
        * column: 29
Flags.......: lvalue
Location....: file foo.c line 5 column 1
```

</details>

XLiZHI:
> We always ignore any array to pointer decay casts that clang generates for us:
> 
> https://github.com/esbmc/esbmc/blob/bb01667c984f10ba6b326e4d85a23a004bdfa08d/src/clang-c-frontend/clang_c_convert.cpp#L2910-L2914
> 
> 
> and just return the original expression unchanged:
> https://github.com/esbmc/esbmc/blob/bb01667c984f10ba6b326e4d85a23a004bdfa08d/src/clang-c-frontend/clang_c_convert.cpp#L2979
> 
> This then causes IR like this:

I will investigate it in another PR.

lucasccordeiro:
Thanks for submitting this patch, @XLiZHI.

XLiZHI:
Hi @lucasccordeiro and @intrigus-lgtm 

I dug into the problem a bit, Clang AST provides decay for array to pointer. but Most of the time, esbmc has its own typecast method, such as based on the type of the argument to a function.  For instance:

```c
void foo(int m, int n, int x[m][n])
{
  x[1][0] = 1;
}

int main(void)
{
  int A[2][2] = {{0, 0}, {0, 0}};
  foo(2, 2, A);

  return 0;
}
```

with AST array to pointer decay:
`        FUNCTION_CALL:  foo(2, 2, (signed int [n] *)((void *)((signed int [2] *)(&A[0][0]))))`

without:
`         FUNCTION_CALL:  foo(2, 2, (signed int [n] *)((void *)(&A[0][0])))`
         
What I'm trying to say is that this kind of redundant casting often happens, leading to errors. So, I tried to keep the logic of ESBMC itself, which is to typecast as needed, rather than fully following the AST.

intrigus-lgtm:
> Hi @lucasccordeiro and @intrigus-lgtm
> 
> I dug into the problem a bit, Clang AST provides decay for array to pointer. but Most of the time, esbmc has its own typecast method, such as based on the type of the argument to a function. For instance:
> 
> ```c
> void foo(int m, int n, int x[m][n])
> {
>   x[1][0] = 1;
> }
> 
> int main(void)
> {
>   int A[2][2] = {{0, 0}, {0, 0}};
>   foo(2, 2, A);
> 
>   return 0;
> }
> ```
> 
> with AST array to pointer decay: ` FUNCTION_CALL: foo(2, 2, (signed int [n] *)((void *)((signed int [2] *)(&A[0][0]))))`
> 
> without: ` FUNCTION_CALL: foo(2, 2, (signed int [n] *)((void *)(&A[0][0])))`
> 
> What I'm trying to say is that this kind of redundant casting often happens, leading to errors. So, I tried to keep the logic of ESBMC itself, which is to typecast as needed, rather than fully following the AST.

Thanks for looking into this!
Could you elaborate what you mean with "What I'm trying to say is that this kind of redundant casting often happens, leading to errors."?
I would have expected that fully following the clang AST would be better, because it means that we have to do less adjustments.

PR #2502, opened 2025-06-02
[smt] don't access vector out of bounds

intrigus-lgtm:
`reserve` only changes the capacity, but not the size.
After `args.reserve(20)`, `args[10]` would still fail.
The loop that initialized all `args` has been changed to use push_back instead of indices.
The few cases that later tried to write to e.g. `args[0]` where there was previously no value written to `args[0]` have been changed to now do the write.

Alternative to #2498


rafaelsamenezes (approved):

lucasccordeiro (approved):
Thanks for submitting this PR, @intrigus-lgtm!

PR #2511, opened 2025-06-09
[solidity] improve overall performance

ChenfengWei0:
```sh
# tested on reentrance_1
# before
Over 5 mins
# after
esbmc --sol example.sol example.solast --incremental-bmc --no-standard-checks  1.90s user 0.19s system 64% cpu 3.255 total

# tested on reentrance_2
# before
Over 5 mins
# after
esbmc --sol example.sol example.solast --contract Reproduction     4.13s user 0.20s system 80% cpu 5.384 total
```

image

now the whole regression test can be finished in 120 second (-j 8)
---

This PR also contains some bug fixing, including library API (_max/_min), address binding in trusted mode


regression/esbmc-solidity/doftcoin_1/contract.sol
```diff
@@ -0,0 +1,192 @@
...
+    
+    constructor() {
+        owner = msg.sender;
+    }
+
+    function transferOwnership(address newOwner) public {
+        owner = newOwner;
+    }
+}
+
```
lucasccordeiro, line 11: The regression tests are a bit of a mess :-) Can you remove empty lines, comments, etc?

lucasccordeiro (changes requested):

regression/esbmc-solidity/doftcoin_1/contract.sol
ChenfengWei0: Changed as suggested

lucasccordeiro (approved):

lucasccordeiro:

Issue #1182, opened 2023-07-11
Realloc always succeds


rafaelsamenezes:
ESBMC `realloc` does not account for the error corner cases: `If there is not enough memory, the old memory block is not freed and null pointer is returned. `

Running esbmc with the following test should fail:

```c
#include <stdlib.h>
int main()
{
  int *arr = malloc(sizeof(int) * 5);
  if(arr == NULL)
    return 0;

  arr[0] = 42;
  arr = realloc(arr, sizeof(int) * 7);
  arr[6] = 42;
}
```

lucasccordeiro:
You could check the malloc implementation; we already model this non-determinism of memory allocation there.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/commit/cd81c69b32f148c6fb6512a35a89afb10d25e86e.

PR #2514, opened 2025-06-10
Consider the guard when dealing with sideeffects

XLiZHI:
This is one of the known problems:

1. The current guard needs to be taken into account when dealing with memory allocation, especially in if expressions.
2. There is something wrong with our realloc operation that causes out of bound. This is because we only adjust the bound in the track new pointer, while nothing changes in the dynamic object. I tried to modify the symbol directly, but at the SMT level there was a bit width mismatch.

Should we rethink the implementation of realloc, perhaps malloc and memcpy would be a good approach instead?

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -132,25 +132,26 @@ void goto_symext::do_simplify(expr2tc &expr)
 // Handle side effects
 void goto_symext::handle_sideeffect(
   const expr2tc &lhs,
-  const sideeffect2t &effect)
+  const sideeffect2t &effect,
+  const guardt &guard)
 {
   switch (effect.kind)
   {
   case sideeffect2t::cpp_new:
   case sideeffect2t::cpp_new_arr:
-    symex_cpp_new(lhs, effect);
+    symex_cpp_new(lhs, effect, guard);
     break;
   case sideeffect2t::realloc:
-    symex_realloc(lhs, effect);
+    symex_realloc(lhs, effect, guard);
     break;
   case sideeffect2t::malloc:
-    symex_malloc(lhs, effect);
+    symex_malloc(lhs, effect, guard);
     break;
   case sideeffect2t::alloca:
-    symex_alloca(lhs, effect);
+    symex_alloca(lhs, effect, guard);
     break;
   case sideeffect2t::va_arg:
-    symex_va_arg(lhs, effect);
+    symex_va_arg(lhs, effect, guard);
     break;
   case sideeffect2t::printf2:
     // Do nothing for printf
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
@XLiZHI: We had many discussions to implement realloc, e.g.:

https://github.com/esbmc/esbmc/issues/1182
https://github.com/esbmc/esbmc/issues/1807
https://github.com/esbmc/esbmc/issues/1982

lucasccordeiro:
@XLiZHI: what is the status of this issue? Have you run it over the SV-COMP benchmarks?

XLiZHI:
This PR:

```
  correct:           17641
    correct true:    10366
    correct false:    7275
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15893
  Score:             27223 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15693165417
```

Master:
```
  correct:           17644
    correct true:    10356
    correct false:    7288
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15890
  Score:             27216 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15693167368
```

lucasccordeiro (approved):

lucasccordeiro:

src/goto-symex/symex_assign.cpp
```diff
@@ -142,7 +142,7 @@ void goto_symext::handle_sideeffect(
     symex_cpp_new(lhs, effect, guard);
     break;
   case sideeffect2t::realloc:
-    symex_realloc(lhs, effect, guard);
+    symex_realloc(lhs, effect, guardt());
```
intrigus-lgtm, line 145: Why this change?
Why do we ignore `guard` for realloc?

intrigus-lgtm:
The `handle_sideeffect` and `handle_conditional` functions do not have documentation for the new `guard` parameter and I don't feel qualified to add them in a PR.

> The current guard needs to be taken into account when dealing with memory allocation, especially in if expressions.

Why is this? Is there a testcase that shows something that has been fixed by this PR that is not related to `realloc`?
I don't quite trust the `realloc` model.

src/goto-symex/symex_assign.cpp
XLiZHI: If you look at the implementation of renumber in realloc, you will see that it is not symbolic. Therefore, we can't simply use guard to control it, which is a design flaw. The problem with this is that when guard is false, it will also renumber and cause an error.

Guard is therefore set to true to avoid errors in the current implementation. it works fine in both cases: 1. guard is true, normal realloc. 2. If guard is false, only the renumber of the pointer object is applied, the size does not change.

I plan to update the implementation of this realloc.

PR #2515, opened 2025-06-11
[cpp] add support for std::is_sorted and improve vector initialization

### LINKED ISSUE #2273: [C++ frontend] error: no member named 'is_sorted' in namespace 'std'
````CPP
#include <iostream>
#include <vector>
#include <string>
#include <algorithm>
#include <cassert>

class Car {
public:
    std::string brand;
    int speed;

    Car(std::string brand, int speed) : brand(brand), speed(speed) {}
};

bool compare_speed(const Car& c1, const Car& c2) {
    return c1.speed < c2.speed;
}

int main() {
    std::vector<Car> cars = {
        Car("Toyota", 120),
        Car("Honda", 130),
        Car("Ford", 110)
    };

    std::vector<Car> sorted_cars = cars;
    std::sort(sorted_cars.begin(), sorted_cars.end(), compare_speed);

    assert(std::is_sorted(sorted_cars.begin(), sorted_cars.end(), compare_speed) && "List is not sorted correctly!");

    std::cout << "All assertions passed successfully!" << std::endl;
    return 0;
}
````

````
$ esbmc lambda.cpp
Parsing lambda.cpp
lambda.cpp:29:17: error: no member named 'is_sorted' in namespace 'std'
    assert(std::is_sorted(sorted_cars.begin(), sorted_cars.end(), compare_speed) && "List is not sorted correctly!");
           ~~~~~^
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
In file included from lambda.cpp:2:
/tmp/esbmc-cpp-headers-7e76-c904-cc50/vector:202:15: error: no matching constructor for initialization of 'Car[]'
    buf = new T[init.size()];
              ^
lambda.cpp:20:29: note: in instantiation of member function 'std::vector<Car>::vector' requested here
    std::vector<Car> cars = {
                            ^
lambda.cpp:7:7: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
class Car {
      ^
lambda.cpp:7:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
lambda.cpp:12:5: note: candidate constructor not viable: requires 2 arguments, but 0 were provided
    Car(std::string brand, int speed) : brand(brand), speed(speed) {}
    ^
In file included from lambda.cpp:2:
/tmp/esbmc-cpp-headers-7e76-c904-cc50/vector:192:15: error: no matching constructor for initialization of 'Car[]'
    buf = new T[x._size];
              ^
lambda.cpp:26:36: note: in instantiation of member function 'std::vector<Car>::vector' requested here
    std::vector<Car> sorted_cars = cars;
                                   ^
lambda.cpp:7:7: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
class Car {
      ^
lambda.cpp:7:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
lambda.cpp:12:5: note: candidate constructor not viable: requires 2 arguments, but 0 were provided
    Car(std::string brand, int speed) : brand(brand), speed(speed) {}
    ^
ERROR: PARSING ERROR
````

### PR DESCRIPTION
**Pull request by lucasccordeiro**
Fixes https://github.com/esbmc/esbmc/issues/2273.

This PR introduces support for the C++ standard library function `std::is_sorted` and updates ESBMC's C++ vector model to ensure correct object construction. It also adds regression tests to verify the correctness of sorting operations and the behavior of `std::is_sorted`.

### DIFF
--- a/src/cpp/library/vector
+++ b/src/cpp/library/vector
@@ -157,7 +157,8 @@ public:
     for (; tmp != t2; tmp++)
       count++;
 
-    buf = new T[count];
+    buf = static_cast<T*>(malloc(sizeof(T) * count));
+    __ESBMC_assume(buf);
     int n = 0;
     for (; t1 != t2; t1++)
     {
@@ -186,7 +187,8 @@ public:
     for (InputIt tmp = t; tmp != addr; tmp++)
       count++;
 
-    buf = new T[count];
+    buf = static_cast<T*>(malloc(sizeof(T) * count));
+    __ESBMC_assume(buf);
     for (int i = 0; i < count; i++)
       buf[i] = *t++;
 
@@ -206,7 +208,8 @@ public:
 
   vector(const vector &x) : _size(0), _capacity(1)
   {
-    buf = new T[x._size];
+    buf = static_cast<T*>(malloc(sizeof(T) * x._size));
+    __ESBMC_assume(buf);
     _size = x._size;
     for (int i = 0; i < _size; i++)
       buf[i] = x.buf[i];
@@ -227,7 +230,8 @@ public:
 
   vector &operator=(const vector &x)
   {
-    buf = new T[x._size];
+    buf = static_cast<T*>(malloc(sizeof(T) * x._size));
+    __ESBMC_assume(buf);
     _size = x._size;
     for (int i = 0; i < _size; i++)
       buf[i] = x.buf[i];
@@ -252,7 +256,8 @@ public:
     for (; tmp != t2; tmp++)
       count++;
 
-    buf = new T[count];
+    buf = static_cast<T*>(malloc(sizeof(T) * count));
+    __ESBMC_assume(buf);
     int n = 0;
     for (; t1 != t2; t1++)
     {
--- a/src/cpp/library/vector
+++ b/src/cpp/library/vector
@@ -135,6 +135,8 @@ public:
     _size = 0;
     _capacity = 10;
   }
+  // TODO: Support destructors. 
+  // Check discussion at https://github.com/esbmc/esbmc/pull/2515
 #if 0
   // Destructor
   ~vector()
--- a/src/cpp/library/vector
+++ b/src/cpp/library/vector
@@ -155,8 +155,7 @@ public:
     for (; tmp != t2; tmp++)
       count++;
 
-    buf = static_cast<T*>(malloc(sizeof(T) * count));
-    __ESBMC_assume(buf);
+    buf = new T[count];
     int n = 0;
     for (; t1 != t2; t1++)
     {
@@ -206,8 +205,7 @@ public:
 
   vector(const vector &x) : _size(0), _capacity(1)
   {
-    buf = static_cast<T*>(malloc(sizeof(T) * x._size));
-    __ESBMC_assume(buf);
+    buf = new T[x._size];
     _size = x._size;
     for (int i = 0; i < _size; i++)
       buf[i] = x.buf[i];
@@ -228,8 +226,7 @@ public:
 
   vector &operator=(const vector &x)
   {
-    buf = static_cast<T*>(malloc(sizeof(T) * x._size));
-    __ESBMC_assume(buf);
+    buf = new T[x._size];
     _size = x._size;
     for (int i = 0; i < _size; i++)
       buf[i] = x.buf[i];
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: src/cpp/library/vector [hunk 1]
```diff
@@ -189,7 +189,8 @@ public:
 
   vector(const vector &x) : _size(0), _capacity(1)
   {
-    buf = new T[x._size];
+    buf = static_cast<T*>(malloc(sizeof(T) * x._size));
```
rafaelsamenezes, line 211: Aren't you creating a memleak here? Maybe using something like __ESBMC_alloca or add a free in a destructor (is there one available?)

src/cpp/library/vector
```diff
@@ -199,11 +200,12 @@ public:
...
-    buf = new T[init.size()];
+    buf = static_cast<T*>(malloc(sizeof(T) * init.size()));
+    __ESBMC_assume(buf);
     _size = init.size();
-    for (int i = 0; i < _size; i++)
-      buf[i] = *(init.begin() + i);
-    _capacity = _size;
+    auto it = init.begin();
+    for (int i = 0; i < _size; i++, ++it)
+      new(&buf[i]) T(*it);
```
rafaelsamenezes: What's the intention here?

XLiZHI: I think we don't support placement-new, and in my tests this test case will fail:

```
    std::vector<int> v = {1, 2, 3};
    assert(v.size() == 3);
    assert(v[0] == 1);
    assert(v[1] == 2);
    assert(v[2] == 3);
```

XLiZHI (commented):
This is an approach that works! I was going to use operator new for this, but it has some issues.

rafaelsamenezes (changes requested):
Just asking more explanations about the vector changes.

Can you add a test case with memleak check?

src/cpp/library/vector
XLiZHI: I recommend reverting to the original assignment method

lucasccordeiro: Thanks, @XLiZHI. Implemented as suggested.

src/cpp/library/vector
lucasccordeiro: I have tried to replace:

buf = static_cast<T*>(malloc(sizeof(T) * init.size()));

with

buf = __ESBMC_alloc(sizeof(T) * init.size());

We get:

````
/tmp/esbmc-cpp-headers-0f84-efba-1103/vector:205:11: error: called object type 'bool[1]' is not a function or function pointer
    buf = __ESBMC_alloc(sizeof(T) * init.size());
          ^~~~~~~~~~~~~
main.cpp:12:29: note: in instantiation of member function 'std::vector<Car>::vector' requested here
    std::vector<Car> cars = {
                            ^
ERROR: PARSING ERROR
````

Let me see if I can use free in the destructor.

**Comment by lucasccordeiro** [comment]
> Can you add a test case with memleak check?

I have just added a test case as known bug. Our C++ OMs are full of new allocations without a corresponding free.

The simple test case here calls other C++ OMs.

I'll leave for this one for @XLiZHI to fix in the entire C++ OM.

**Comment by XLiZHI** [comment]
Let me clarify the issues I know about C++ models:

Memory leaks do exist in the C++ model, and the reason there is no destructor in the model is because esbmc was designed primarily for C. esbmc uses a shallow copy when assigning values to structs, which in some cases causes the dynamic object to be deallocated when the temporary object is destroyed.

We have to think about what happens if we create temporary objects of C++ structs if there are dynamic objects in the members.

**Review by XLiZHI — APPROVED** [review]

**Comment by lucasccordeiro** [comment]
Thanks, @XLiZHI.

I commented out the destructor!

@rafaelsamenezes : please check @XLiZHI's comment above ;-)

File: src/cpp/library/vector [hunk 3]
```diff
@@ -130,11 +130,24 @@ public:
...
   explicit vector() : _size(0), _capacity(1)
   {
-    buf = new T[10];
+    buf = static_cast<T*>(malloc(sizeof(T) * 10));
+    __ESBMC_assume(buf);
     _size = 0;
     _capacity = 10;
   }
-
+#if 0
```
rafaelsamenezes, line 140: TODO?

rafaelsamenezes (approved):

src/cpp/library/vector
lucasccordeiro: Implemented as suggested.

PR #2516, opened 2025-06-11
[goto-symex] use size directly

XLiZHI:
We keep the calculation of object size consistent with cpp new. The information from Clang AST is enough to calculate the size needed for the bounds check.

_update:_ When size is 1, we need to derive the size based on the type.

`type_byte_size_expr` doesn't seem to handle size expressions with symbols (not all constants) very well.

Master:
```
  correct:           17662
    correct true:    10372
    correct false:    7290
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15872
  Score:             27250 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/15429068267
```

This PR (new):
```
  correct:           17666
    correct true:    10372
    correct false:    7294
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15868
  Score:             27254 (max: 55885)
```


lucasccordeiro:
Thanks, @XLiZHI.

lucasccordeiro (changes requested):
Can you some comments/documentation here?

lucasccordeiro (approved):

lucasccordeiro:

PR #2517, opened 2025-06-11
[cpp] resolve mismatch template deduction errors with pointer types

lucasccordeiro:
Add explicit pointer specializations and fix return type consistency for mismatch function to handle mixed iterator/pointer arguments.

### DIFF
--- a/regression/esbmc-cpp/algorithm/algorithm9/test.desc
+++ b/regression/esbmc-cpp/algorithm/algorithm9/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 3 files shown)


XLiZHI:
Based on the standard library API, I think we only need two mismatch versions (before C++ 17):

```
template<class InIt1, class InIt2>
std::pair<InIt1, InIt2>
mismatch(InIt1 first1, InIt1 last1, InIt2 first2)
{
  while (first1 != last1 && *first1 == *first2)
  {
    ++first1;
    ++first2;
  }
  return std::make_pair(first1, first2);
}
```
and
```
template<class InIt1, class InIt2, class Pred>
std::pair<InIt1, InIt2>
mismatch(InIt1 first1, InIt1 last1, InIt2 first2, Pred pred)
{
  while (first1 != last1 && pred(*first1, *first2))
  {
    ++first1;
    ++first2;
  }
  return std::make_pair(first1, first2);
}
```
Using `InIt2*` is not universal, only bare pointers are supported. Maybe that would be better.

PR #2534, opened 2025-06-17
[C++ verification] Add a new implementation of typeid

XLiZHI:
The type of polymorphism needs to be fetched at runtime, which is not handled by the frontend, so for now we skip it for preliminary support.

The TC in this PR reflects all currently supported methods.


lucasccordeiro:

PR #2536, opened 2025-06-18
[c++] add allocator template parameter to std::vector

lucasccordeiro:
Add second template parameter with default std::allocator<T> to match standard library interface. Fixes "too many template arguments" errors when using custom allocators.


XLiZHI:
Instead of creating new functions, we need to modify the existing functions.

image

Let's say we just add Allocator parameter to the original function and give it a default value.

`explicit vector( const Allocator& alloc = Allocator() );`

PR #2538, opened 2025-06-19
[c++] add deterministic hash specializations for std::hash

lucasccordeiro:
This PR adds a set of deterministic and type-safe `std::hash` specializations to `functional`. These implementations aim to ensure platform-independent, deterministic behavior across supported fundamental types, including integral, floating point, pointer, and string types.

### DIFF
--- a/src/cpp/library/functional
+++ b/src/cpp/library/functional
@@ -2,65 +2,102 @@
 #define STL_FUNCTIONAL
 
 #include <cassert>
-#include <utility>  // std::forward
+#include <utility> // std::forward
 #include <cstddef>
 #include <type_traits>
-#include <cstdint>  // for uint32_t, uint64_t
-#include <cmath>    // for INFINITY
-#include <string>   // for std::string
+#include <cstdint> // for uint32_t, uint64_t
+#include <cmath>   // for INFINITY
+#include <string>  // for std::string
 
-namespace std {
+namespace std
+{
 
 // Base class for callable objects
-class callable_base {
+class callable_base
+{
 public:
   virtual ~callable_base() = default;
-  virtual int invoke(int, int) const { assert(false); return 0; }
-  virtual int invoke(int) const { assert(false); return 0; }
-  virtual bool invoke_bool(int, int) const { assert(false); return false; }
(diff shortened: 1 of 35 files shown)


src/cpp/library/functional
```diff
@@ -116,6 +121,273 @@ public:
...
+    key *= 0xc2b2ae35;
+    key ^= key >> 16;
+    return key;
+}
+
+// Helper function to handle signed integer hashing consistently
+template<typename SignedInt>
+inline size_t __esbmc_hash_signed_int(SignedInt key) {
+    // Convert to size_t safely
+    size_t unsigned_key = static_cast<size_t>(key);
```
XLiZHI: `size_t` is unsigned, and converting directly from a negative signed value causes sign expansion problems.

src/cpp/library/functional
```diff
@@ -116,6 +121,273 @@ public:
...
+            hash_value = hash_value * prime + static_cast<size_t>(static_cast<unsigned char>(str[i]));
+        }
+        
+        return __esbmc_deterministic_hash(hash_value);
+    }
+};
+
+// Utility functions for testing
+template<typename T>
+bool verify_hash_determinism(const T& value) {
```
XLiZHI: Is this a standard library function?

src/cpp/library/functional
```diff
@@ -116,6 +121,273 @@ public:
...
+    size_t hash1 = hasher(value);
+    size_t hash2 = hasher(value);
+    
+    // Should always be equal due to deterministic implementation
+    assert(hash1 == hash2);
+    return hash1 == hash2;
+}
+
+template<typename T>
+void verify_hash_basic_properties(const T& value1, const T& value2) {
```
XLiZHI: Is this a standard library function?

src/cpp/library/functional
```diff
@@ -231,4 +503,21 @@ decltype(auto) invoke(Func&& f, Args&&... args) {
 
 } // namespace std
 
+// Simple verification helper macros
+#define VERIFY_HASH_DETERMINISM(type, value) \
```
XLiZHI: Should these be kept in OM?

PR #2540, opened 2025-06-20
[CI] improve auto commit

XLiZHI:
Improve the code style check CI:

Only branches in the ESBMC repository will trigger the auto commit, because git commit does not have permission to modify other repositories.

Otherwise, we show the corresponding code, throw an error.


lucasccordeiro:

PR #2542, opened 2025-06-23
Parallel Solving

Luke-Sanderson:
This PR aims to re-add the parallel-solving option from https://github.com/esbmc/esbmc/commit/fe766d340e1391f5d9b60e504409f55b49f3bee2 to ESBMC


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -473,6 +473,14 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
   }
 #endif
 
+  // parallel solving activates "--multi-property"
```
intrigus-lgtm, line 476: That comment is only partially honest.

src/esbmc/bmc.cpp
```diff
@@ -1433,7 +1433,34 @@ smt_convt::resultt bmct::multi_property_check(
         clear_verified_claims(claim, is_goto_cov);
   };
 
-  std::for_each(std::begin(jobs), std::end(jobs), job_function);
+  // PARALLEL
+  if(options.get_bool_option("parallel-solving"))
+  {
+    /* NOTE: I would love to use std::for_each here, but it is not giving
```
intrigus-lgtm, line 1539: What results would you have expected here?
Do you have example code in another branch/patch?

ChenfengWei0:
I’m happy to see that we’re bringing back parallel solving. I expect this could significantly improve the performance of branch coverage calculation in the Test-Comp competition. That said, I’m not sure whether this implementation might break k-induction for the current multi-property verification.

src/esbmc/bmc.cpp
Luke-Sanderson: These comments are copied from fe766d340e1391f5d9b60e504409f55b49f3bee2 when readding their implementation of parallisation. I am using it as a starting point.

Here is another [branch with the std::for_each implementation](https://github.com/Luke-Sanderson/esbmc/blob/31b85d2c6d777ff9f00db925087684f9cef55b5f/src/esbmc/bmc.cpp#L1451) but unlike the manual implementation does not seem to create threads when looking at htop.

lucasccordeiro:
@Luke-Sanderson: Two test cases have failed here:

````
Errors while running CTest
	5489 - regression/parallel-solving/pthread_join (Failed) parallel-solving regression
	5490 - regression/parallel-solving/pthread_mutex (Failed) parallel-solving regression
Error: Process completed with exit code 8.
````

Can you please take a look at these tests?

lucasccordeiro (approved):

PR #2546, opened 2025-06-24
[python-frontend] Handling string attributes

brcfarias:
- Handle string attributes as char pointers;
- Append `'\0'` to char arrays to enable use of `strcmp` for string comparisons.


lucasccordeiro (approved):

lucasccordeiro:

PR #2551, opened 2025-06-26
Recognising "global" declarations in ESBMC-Python

yjtew:
The changes made allows for ESBMC-Python to recognise the "global" keyword. 4 new tests were written to ensure that ESBMC-Python produces the correct verification results/errors as Python does.

Notably, all symbols generated had their static_lifetime flag set to false regardless of scope prior to this pull request. I had changed it to only be false if the variable is within a class or a function.

### DIFF
--- a/regression/python/global-decl-fail/main.py
+++ b/regression/python/global-decl-fail/main.py
@@ -1,0 +1,7 @@
+x = 10
+
+def modify():
+    x = x + 1
+    global x
+  
+modify()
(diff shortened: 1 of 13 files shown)


brcfarias:
Hi @yjtew, many thanks for the PR.
I think that global storage management shouldn't be a `python_convert `concern. We're actually planning to distribute its responsibilities among other classes. Given what we currently have, it could be part of the `module_manager `/`module` classes. This way, python_convert could query the global variables from that module. What do you think? I still don't have a full picture yet, but perhaps it's a good starting point.

lucasccordeiro (approved):

lucasccordeiro:
@yjtew: Thanks for submitting this PR.

@brcfarias: Could you please submit a PR with your suggestion?

PR #2555, opened 2025-06-26
Adjust Python tests

brcfarias:
- Fix some tests.
- Rename failing regression test folders to end with "fail", and the successful ones to not contain it.
- Add a CI task to check consistency against the interpreter, meaning we expect the interpreter to return 1 for all "fail" tests and 0 for the successful ones.


lucasccordeiro (approved):

rafaelsamenezes (changes requested):
I have a few questions:

1. Is there any reason why you didn't do this inside the testing tool? 
2. You also changed the code of some tests, can you update your PR description for that?

lucasccordeiro:
@brcfarias: could you please check @rafaelsamenezes' questions?

PR #2545, opened 2025-06-24
Update memcpy function

mihaistate:
Improved the memcpy function to store 8 bytes per iteration and check 8-byte alignment for pointers
If the pointers are aligned, the function safely, it copies 8 bytes of memory using uint64_t pointers for efficiency. Remaining bytes are handled in a trailing byte loop.
If the pointers are not 8-bytes aligned, the function falls back to a safe byte-by-byte copy.
Addressed issue #2282

Issue #2282, opened 2025-02-05
False positives/false negatives in test cases due `--no-unwinding-assertions`. Improve `memcpy`


intrigus-lgtm:
In a recent PR I "fixed" some testcases:
https://github.com/esbmc/esbmc/pull/2133/commits/f64662c8f02ba7763ee113279996b8175d363038

However, the TCs rely on `--no-unwinding-assertions` so some of the TCs that should fail, do not.
This is because they are never unwound often enough to hit the failing assertion and they also hit no unwinding.

Also, some testcases that should succeed, to succeed, but **only**, because unwinding assertions are disabled.

The high number of needed unwindings (81 or 82(?)) appears to be caused by `memcpy` being used to copy very large objects, necessitating a high number of unwindings of the for loop that copies individual bytes.

How to solve this:
1. Remove `--no-unwinding-assertions` from testcases.
2. Increase the number of unwinding assertions if needed.
3. Potentially optimize `memcpy` to copy more than one byte per iteration.
Something like this, where we either copy multiple individual bytes per iteration or directly copy a type bigger than bytes like 64-bit longs (not implemented here)
```c
void *memcpy(void *dst, const void *src, size_t n)
{
__ESBMC_HIDE:;
  char *cdst = dst;
  const char *csrc = src;
  // Copy 16 bytes at a time
  size_t i = 0;
  for (; i + 16 <= n; i += 16) {
    cdst[i] = csrc[i];
    cdst[i + 1] = csrc[i + 1];
    cdst[i + 2] = csrc[i + 2];
    cdst[i + 3] = csrc[i + 3];
    cdst[i + 4] = csrc[i + 4];
    cdst[i + 5] = csrc[i + 5];
    cdst[i + 6] = csrc[i + 6];
    cdst[i + 7] = csrc[i + 7];
    cdst[i + 8] = csrc[i + 8];
    cdst[i + 9] = csrc[i + 9];
    cdst[i + 10] = csrc[i + 10];
    cdst[i + 11] = csrc[i + 11];
    cdst[i + 12] = csrc[i + 12];
    cdst[i + 13] = csrc[i + 13];
    cdst[i + 14] = csrc[i + 14];
    cdst[i + 15] = csrc[i + 15];
  }
  // Copy 8 bytes at a time
  for (; i + 8 <= n; i += 8) {
    cdst[i] = csrc[i];
    cdst[i + 1] = csrc[i + 1];
    cdst[i + 2] = csrc[i + 2];
[...]

PR #2561, opened 2025-07-01
Updated memcpy function

mihaistate:
Improved the memcpy function to store 8 bytes per iteration and check 8-byte alignment for pointers
If the pointers are aligned, the function safely, it copies 8 bytes of memory using uint64_t pointers for efficiency. Remaining bytes are handled in a trailing byte loop.
If the pointers are not 8-bytes aligned, the function falls back to a safe byte-by-byte copy.
Addressed issue https://github.com/esbmc/esbmc/issues/2282
This is the continuation of #2545


mihaistate:
I have an error for ESBMC Build Ubuntu here: https://github.com/esbmc/esbmc/actions/runs/15980661626/job/45074069694
I updated the CMakeList.txt file with the aim of solving the issue: https://github.com/esbmc/esbmc/pull/2561/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a. I set C++ to be 17 as a solution for the errors that appear, it did not work, I also added locally an executable and an compiling target: set (ESBMC_VERSION "${ESBMC_VERSION_MAJOR}.${ESBMC_VERSION_MINOR}.${ESBMC_VERSION_PATCH}")
set (CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)

add_executable(esbmc ...)
target_compile_features(esbmc PUBLIC cxx_std_17) . 
This also gives me another error, so I don't know if I should keep the last two lines or not: CMake Error at src/esbmc/CMakeLists.txt:44 (add_executable):
  add_executable cannot create target "esbmc" because another target with the
  same name already exists.  The existing target is an executable created in
  source directory "/root/ESBMC/esbmc".  See documentation for policy CMP0002
  for more details.
I would like to know if I can somehow import the already created target that the compiler reffers to, but I don't know how to identify it inside /root/ESBMC/esbmc by using grep -R "add_executable *(esbmc" -n . Would it be another way to solve the error?

intrigus-lgtm:
https://github.com/esbmc/esbmc/pull/2561/commits/c06748b889d2d1b017c7b00a5adb7e27ac76ca95 looks pretty broken and I don't know why.
If possible, could you fix this?

src/c2goto/library/string.c
```diff
@@ -269,11 +275,80 @@ __ESBMC_HIDE:;
...
-  for (size_t i = 0; i < n; i++)
-    cdst[i] = csrc[i];
+  // If n is 0, return dst (nothing to copy)
+  if (n == 0)
+    return dst;
+
+  // NULL pointer checks
+  __ESBMC_assert(dst != NULL, "Destination pointer is NULL.");
+  __ESBMC_assert(src != NULL, "Source pointer is NULL.");
+  __ESBMC_assume(n <= 1024);
```
rafaelsamenezes: It seems to be some leftover here

src/c2goto/library/string.c
```diff
@@ -269,11 +275,80 @@ __ESBMC_HIDE:;
...
+  char *cdst = (char *)dst;
+  const char *csrc = (const char *)src;
+
+  if (((uintptr_t)dst % 8 == 0) && ((uintptr_t)src % 8 == 0))
```
rafaelsamenezes: We are probably better off doing some intrinsic for this. ESBMC method does not like pointers being casted into integers

src/c2goto/library/string.c
```diff
@@ -269,11 +275,80 @@ __ESBMC_HIDE:;
...
+  return dst;
+__ESBMC_HIDE:;
+  // If n is 0, return dst (nothing to copy)
+  if (n == 0)
+    return dst;
+
+  // NULL pointer checks
+  __ESBMC_assert(dst != NULL, "Destination pointer is NULL.");
+  __ESBMC_assert(src != NULL, "Source pointer is NULL.");
+  __ESBMC_assume(n <= 1024);
```
rafaelsamenezes: Some leftover?

src/c2goto/library/string.c
```diff
@@ -269,11 +275,80 @@ __ESBMC_HIDE:;
...
+    // Copy the  remaining bytes
+    for (; i < n; ++i)
+    {
+      cdst[i] = csrc[i];
+    }
+  }
+  else
+  {
+
+<<<<<<< HEAD
```
rafaelsamenezes: Some broken git

rafaelsamenezes (changes requested):
Other than the current rebase issue:
- I am not convinced that this will be any better as those pointer operations are very expensive for the decision procedure. Still, I see no problem in benchmarking this.
- There is some potential if this is implemented directly in the symbolic execution as there you can compute the alignment directly.

mihaistate:
@intrigus-lgtm Thanks, I fixed some unnecessary changes from https://github.com/esbmc/esbmc/commit/c06748b889d2d1b017c7b00a5adb7e27ac76ca95 and the error disappeared.

src/c2goto/library/string.c
mihaistate: I deleted it in my last commit

src/c2goto/library/string.c
mihaistate: I removed the  __ESBMC_assume(n <= 1024); line by ammending my last commit.

src/c2goto/library/string.c
mihaistate: Solved it

src/c2goto/library/string.c
mihaistate: @rafaelsamenezes I created a new function called intrinsic_memset inside builtin_functions.cpp, which addresses the byte allignment and handles the bytes computation symbolically, and moved the copy of 8 bytes per iteration from string.c into this new function. Although I may have to handle some CI issues, I am open for a review.

mihaistate:
https://github.com/esbmc/esbmc/actions/runs/16166444615/job/45629392570?pr=2561#logs
I have this error for .github/workflows/build-unix.yml:
Run if [[ "${RUNNER_OS}" == "macOS" ]]; then
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
Error: Process completed with exit code 1.

Any suggestions for this error for ARM64 linux build?

mihaistate:
Dr.@lucasccordeiro I have been working on this PR in the last week, which is a continuation of the PR #2545, opened two weeks ago, with my initial solution. I wish I could have been more quicker, but it took me some days to fix some issues from my rebase and CI errors. After the initial review, I had to implement a new intrinsic function in my last commit and then fix some build errors. I will try finish this task soon, after handling some CI errors, and I will be open for a review.

intrigus-lgtm:
> Dr.@lucasccordeiro I have been working on this PR in the last week, which is a continuation of the PR #2545, opened two weeks ago, with my initial solution. I wish I could have been more quicker, but it took me some days to fix some issues from my rebase and CI errors. After the initial review, I had to implement a new intrinsic function in my last commit and then fix some build errors. I will try finish this task soon, after handling some CI errors, and I will be open for a review.

Quick question, are you developing on Windows?
A lot of the changes in https://github.com/esbmc/esbmc/pull/2561/commits/c06748b889d2d1b017c7b00a5adb7e27ac76ca95 look like line ending changes and should be reverted.

intrigus-lgtm:
I think you should be able to significantly clean up the pr by using `git rebase -i`:
Here's a random tutorial that looked okay (just in case you don't already know how the command works)
https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d

For example, you should be able to issue an `edit` to https://github.com/esbmc/esbmc/pull/2561/commits/c06748b889d2d1b017c7b00a5adb7e27ac76ca95 only and then simply remove all unwanted changes.

Note, after rebasing you'll likely have to force push your PR.
You can do this safer by using `git push --force-with-lease`.

mihaistate:
> > Dr.@lucasccordeiro I have been working on this PR in the last week, which is a continuation of the PR #2545, opened two weeks ago, with my initial solution. I wish I could have been more quicker, but it took me some days to fix some issues from my rebase and CI errors. After the initial review, I had to implement a new intrinsic function in my last commit and then fix some build errors. I will try finish this task soon, after handling some CI errors, and I will be open for a review.
> 
> Quick question, are you developing on Windows? A lot of the changes in [c06748b](https://github.com/esbmc/esbmc/commit/c06748b889d2d1b017c7b00a5adb7e27ac76ca95) look like line ending changes and should be reverted.
 
Well, I have Windows, but I do the development on the latest Ubuntu version on WSL, in VS Code. I did git diff origin/master..HEAD -- regression/esbmc-cpp/unordered_map/emplace_fail/, git checkout origin/master -- regression,  git add regression/, git diff --cached, git commit --amend, git push -f and I reverted in this way changes from https://github.com/esbmc/esbmc/commit/c06748b889d2d1b017c7b00a5adb7e27ac76ca95.

Thanks for the suggestions, @intrigus-lgtm, I already have experience on how to do a PR rebase properly, when I will need to the rebase, I will be more careful,  this one is more manageable.

.github/workflows/build-unix.yml
intrigus-lgtm, line 1: Please revert

CREDITS
intrigus-lgtm, line 1: Please revert

intrigus-lgtm (changes requested):
Please also remove all other unrelated changes.

.github/workflows/build-unix.yml
mihaistate: I did it here: https://github.com/esbmc/esbmc/pull/2561/commits/68f9ec2e77ecc33dac6ed8b6753146d38b338cb4

CREDITS
mihaistate: Done

intrigus-lgtm:
I fixed/cleaned up your PR here:
https://github.com/esbmc/esbmc/compare/master...intrigus/fixup/memcpy

Notably, https://github.com/esbmc/esbmc/commit/3ca2ece24a636db5a0642aab6c15de2035fcbe41 breaks your implementation, because this actually causes an endless loop if your have something like `memcpy(foo, bar, symbolic_value)`, because your intrinsic will then try to bump a call to `memcpy`, which will call the intrinsic again and again forever. See https://github.com/esbmc/esbmc/commit/2bffa9a939f6a3745a223e7fbd2784686af66c2c for test cases.

I would recommend following the implementation of `intrinsic_memset`:
https://github.com/esbmc/esbmc/blob/90f3ca14db430a6d075a7fb563cf8cfd46ea8b58/src/c2goto/library/string.c#L288
https://github.com/esbmc/esbmc/blob/90f3ca14db430a6d075a7fb563cf8cfd46ea8b58/src/goto-symex/symex_main.cpp#L585
https://github.com/esbmc/esbmc/blob/90f3ca14db430a6d075a7fb563cf8cfd46ea8b58/src/clang-c-frontend/clang_c_language.cpp#L390
https://github.com/esbmc/esbmc/blob/90f3ca14db430a6d075a7fb563cf8cfd46ea8b58/src/goto-symex/builtin_functions.cpp#L1358

mihaistate:
@intrigus-lgtm I improved the memcpy implementation, I created a mempcy__impl() function, similarly to the memset one, and added your tests from https://github.com/esbmc/esbmc/commit/2bffa9a939f6a3745a223e7fbd2784686af66c2c

mihaistate:
I encoutered some errors for the regression/esbmc/31_cdaudio test, was it expected to work well, or do I need to edit it to solve the errors?
 root@ASUSVivobookS:~/ESBMC/esbmc/regression/esbmc/31_cdaudio# esbmc main.c 
ESBMC version 7.9.0 64-bit x86_64 linux
Parsing main.c
main.c:11:24: error: typedef redefinition with different types ('unsigned short' vs 'int')
typedef unsigned short wchar_t;
                       ^
/usr/lib/llvm-16/lib/clang/16/include/../include/stddef.h:74:24: note: previous definition is here
typedef __WCHAR_TYPE__ wchar_t;
                       ^
main.c:117:22: error: typedef redefinition with different types ('unsigned int' vs 'unsigned long')
typedef unsigned int size_t;
                     ^
/usr/lib/llvm-16/lib/clang/16/include/../include/stddef.h:46:23: note: previous definition is here
typedef __SIZE_TYPE__ size_t;
                      ^
main.c:2640:9: warning: #pragma once in main file [-Wpragma-once-outside-header]
#pragma once
        ^

*main.c:7454:16: error: incompatible function pointer types assigning to 'NTSTATUS (*)(PDEVICE_OBJECT, PIRP, PVOID)' (aka 'long (*)(struct _DEVICE_OBJECT *, struct _IRP *, void *)') from 'NTSTATUS (*)(PDEVICE_OBJECT, PIRP, PKEVENT)' (aka 'long (*)(struct _DEVICE_OBJECT *, struct _IRP *, struct _KEVENT *)') [-Wincompatible-function-pointer-types]
      compFptr = & CdAudioSignalCompletion;
               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
main.c:7459:32: error: incompatible function pointer types assigning to 'NTSTATUS (*)(PDEVICE_OBJECT, PIRP, PVOID)' (aka 'long (*)(struct _DEVICE_OBJECT *, struct _IRP *, void *)') from 'NTSTATUS (*)(PDEVICE_OBJECT, PIRP, PKEVENT)' (aka 'long (*)(struct _DEVICE_OBJECT *, struct _IRP *, struct _KEVENT *)') [-Wincompatible-function-pointer-types]
  irpSp___0->CompletionRoutine = & CdAudioSignalCompletion;
                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~**

intrigus-lgtm:
I am sorry, but I cannot help you with this PR in this state.
There are simply too many unrelated changes in it which are confusing.
I think what you really want is to rebase this PR on latest master like I did for https://github.com/esbmc/esbmc/tree/intrigus/fixup/memcpy.

mihaistate:
@intrigus-lgtm I did the rebase, now I managed to solve the mentioned test to pass. Now, I have some other tests that get the expected result locally: 
root@ASUSVivobookS:~/ESBMC/esbmc# esbmc regression/k-induction/byte_add_bug/main
.c --k-induction -Wno-error=implicit-function-declaration -Wno-error=implicit-in
t--z3
ESBMC version 7.9.0 64-bit x86_64 linux
Parsing regression/k-induction/byte_add_bug/main.c
warning: unknown warning option '-Werror=implicit-int--z3'; did you mean '-Werror=implicit-int'? [-Wunknown-warning-option]
Converting
Generating GOTO Program
GOTO program creation time: 0.175s
Checking base case, k = 1
Unwinding loop 2 iteration 1   file regression/k-induction/byte_add_bug/main.c line 50 column 5 function mp_add
Not unwinding loop 2 iteration 2   file regression/k-induction/byte_add_bug/main.c line 50 column 5 function mp_add
Unwinding loop 3 iteration 1   file regression/k-induction/byte_add_bug/main.c line 79 column 5 function mp_add
Not unwinding loop 3 iteration 2   file regression/k-induction/byte_add_bug/main.c line 79 column 5 function mp_add
Symex completed in: 0.001s (148 assignments)
Slicing time: 0.001s (removed 78 assignments)
Generated 1 VCC(s), 1 remaining after simplification (70 assignments)
Solving with solver Z3 v4.8.12
.
.
[Counterexample]


State 1 file regression/k-induction/byte_add_bug/main.c line 96 column 5 function main thread 0
  a = 16575 (00000000 00000000 01000000 10111111)

..

State 25 file regression/k-induction/byte_add_bug/main.c line 88 column 5 function mp_add thread 0
  r = 33279 (00000000 00000000 10000001 11111111)

State 26 file regression/k-induction/byte_add_bug/main.c line 100 column 5 function main thread 0
Violated property:
  file regression/k-induction/byte_add_bug/main.c line 100 column 5 function main
  assertion r == a + b
  r == a + b


VERIFICATION FAILED

Bug found (k = 3)
root@ASUSVivobookS:~/ESBMC/esbmc#

but fails in the CI with the same arguments:          

1486/1618 Test #1489: regression/k-induction/byte_add_bug ................................***Failed    0.55 sec
test_byte_add_bug (__main__.RegressionBase) ... FAIL
======================================================================
FAIL: test_byte_add_bug (__main__.RegressionBase)
----------------------------------------------------------------------
  File "D:\a\esbmc\esbmc\regression\testing_tool.py", line 251, in test
    self.fail(error_message_prefix + error_message)
AssertionError: 
TEST: D:/a/esbmc/esbmc/regression/k-induction\byte_add_bug
EXPECTED TO FIND: ['^VERIFICATION FAILED$']
PROGRAM OUTPUT
ESBMC version 7.9.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing D:/a/esbmc/esbmc/regression/k-induction\byte_add_bug\main.c
Checking base case, k = 1
Not unwinding loop 2 iteration 1   file D:/a/esbmc/esbmc/regression/k-induction\byte_add_bug\main.c line 50 column 5 function mp_add
Not unwinding loop 3 iteration 1   file D:/a/esbmc/esbmc/regression/k-induction\byte_add_bug\main.c line 79 column 5 function mp_add
ARGUMENTS: ['C:/deps/esbmc/bin/esbmc.exe', '--k-induction', 'D:/a/esbmc/esbmc/regression/k-induction\\byte_add_bug\\main.c']
----------------------------------------------------------------------
Ran 1 test in 0.475s
FAILED (failures=1)
0.469
          Start 1490: regression/k-induction/check_if
1487/1618 Test #1490: regression/k-induction/check_if ....................................   Passed    0.51 sec

  File "D:\a\esbmc\esbmc\regression\testing_tool.py", line 251, in test
    self.fail(error_message_prefix + error_message)
AssertionError: 
TEST: D:/a/esbmc/esbmc/regression/k-induction\check_if_bug
EXPECTED TO FIND: ['^VERIFICATION FAILED$']
PROGRAM OUTPUT
ESBMC version 7.9.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing D:/a/esbmc/esbmc/regression/k-induction\check_if_bug\main.c
Not unwinding loop 2 iteration 1   file D:/a/esbmc/esbmc/regression/k-induction\check_if_bug\main.c line 12 column 3 function main

ARGUMENTS: ['C:/deps/esbmc/bin/esbmc.exe', '--k-induction', '-Wno-error=implicit-function-declaration', 'D:/a/esbmc/esbmc/regression/k-induction\\check_if_bug\\main.c']
----------------------------------------------------------------------
Ran 1 test in 0.410s
FAILED (failures=1) It is the same case for regression/k-induction/array_bug/main.c

mihaistate:
At the last CI checks, I have this error, not all the test have been done, but there is no failed test yet, maybe the error is caused by a timeout set in  benchbringup.yml and build-linux.yml: 1224/5526 Test #1227: regression/esbmc/memcpy2 .....................................................................   Passed    0.32 sec
          Start 1228: regression/esbmc/memcpy2_sym
Error: Process completed with exit code 143.   
Should I try in these files to lower the parralelism and increase timeout like this: 
- name: Run user-specified benchmark
  continue-on-error: true
  env:
    ESBMC_OPTS: ${{ inputs.options }}
    LOG_DIR: ${{ inputs.logdirectory }}
  run: |
    cd build/
    ctest -j4 --timeout 300 --output-on-failure --progress . ? @rafaelsamenezes @XLiZHI

XLiZHI:
Hi @mihaistate, have you tested it locally? does it time out?

BTW, you should not modify cmake, CI, etc in this PR. In my opinion they are not relevant to the implementation of memcpy. Please open a separate PR and explain your reasons for changing them, if you think they are necessary.

mihaistate:
Hi, @XLiZHI! 

I have added to the esbmc/CMakeLists.txt file this paragraph:
if(NOT WIN32)
  find_package(PkgConfig REQUIRED)
  pkg_check_modules(GMP	REQUIRED IMPORTED_TARGET GLOBAL gmp)
  pkg_check_modules(GMPXX REQUIRED IMPORTED_TARGET GLOBAL gmpxx)

  message(STATUS "GMP LIB DIRS: ${GMP_LIBRARY_DIRS}")
  message(STATUS "GMPXX LIB DIRS: ${GMPXX_LIBRARY_DIRS}")

  target_include_directories(esbmc PRIVATE ${GMP_INCLUDE_DIRS})
  target_link_directories(esbmc PRIVATE ${GMP_LIBRARY_DIRS}) 
  #target_link_libraries(esbmc PRIVATE ${GMP_LIBRARIES} ${GMPXX_LIBRARIES})
  #target_link_libraries(esbmc PRIVATE ${GMP_LIBRARY_DIRS})
  target_link_libraries(esbmc PRIVATE ${GMP_LIBRARIES} ${GMPXX_LIBRARIES})
endif()

The idea is that this issue occured only for this branch, for the memcpy implementation, and was related to some compile errors with the gmp package (that I needed to install locally, version 6.2.1, outside the project) and its correlation with bitwuzla. I just tried to comment out all these lines in the cmake file and now I get errors like this again: 

bzla::BitVector const&, bool)':
bitvector.cpp:(.text+0x8f56): undefined reference to `__gmpz_add_ui'
/usr/bin/ld: bitvector.cpp:(.text+0x8f6e): undefined reference to `__gmpz_urandomm'
/usr/bin/ld: bitvector.cpp:(.text+0x8f83): undefined reference to `__gmpz_add'
/usr/bin/ld: bitvector.cpp:(.text+0x8fa1): undefined reference to `__gmpz_add'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::BitVector(unsigned long, bzla::RNG&)':
bitvector.cpp:(.text+0x912b): undefined reference to `__gmpz_urandomb'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::BitVector(unsigned long)':
bitvector.cpp:(.text+0x55): undefined reference to `__gmpz_init'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::BitVector(bzla::BitVector const&)':
bitvector.cpp:(.text+0x1b1): undefined reference to `__gmpz_init_set'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::iset(bzla::BitVector const&)':
bitvector.cpp:(.text+0x551): undefined reference to `__gmpz_set'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::compare(bzla::BitVector const&) const':
bitvector.cpp:(.text+0x5b9): undefined reference to `__gmpz_cmp'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::set_bit(unsigned long, bool)':
bitvector.cpp:(.text+0x693): undefined reference to `__gmpz_setbit'
/usr/bin/ld: bitvector.cpp:(.text+0x6d1): undefined reference to `__gmpz_clrbit'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlabv.a(bv_bitvector.cpp.o): in function `bzla::BitVector::flip_bit(unsigned long)':
bitvector.cpp:(.text+0x771): undefined reference to `__gmpz_combit'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_get_ull(__mpz_struct const*)':
gmp_utils.cpp:(.text+0x17): undefined reference to `__gmp_bits_per_limb'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_hash(__mpz_struct const*)':
gmp_utils.cpp:(.text+0x9c): undefined reference to `__gmp_bits_per_limb'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_set_ull(__mpz_struct*, unsigned long)':
gmp_utils.cpp:(.text+0x5): undefined reference to `__gmpz_set_ui'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_init_set_ull(__mpz_struct*, unsigned long)':
gmp_utils.cpp:(.text+0x65): undefined reference to `__gmpz_init_set_ui'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_init_set_sll(__mpz_struct*, long)':
gmp_utils.cpp:(.text+0x75): undefined reference to `__gmpz_init_set_si'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_fdiv_q_2exp_ull(__mpz_struct*, __mpz_struct const*, unsigned long)':
gmp_utils.cpp:(.text+0x255): undefined reference to `__gmpz_fdiv_q_2exp'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_fdiv_r_2exp_ull(__mpz_struct*, __mpz_struct const*, unsigned long)':
gmp_utils.cpp:(.text+0x265): undefined reference to `__gmpz_fdiv_r_2exp'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzla.a(util_gmp_utils.cpp.o): in function `bzla::util::mpz_mul_2exp_ull(__mpz_struct*, __mpz_struct const*, unsigned long)':
gmp_utils.cpp:(.text+0x275): undefined reference to `__gmpz_mul_2exp'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlals.a(rng_rng.cpp.o): in function `bzla::RNG::~RNG()':
rng.cpp:(.text+0x40): undefined reference to `__gmp_randclear'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlals.a(rng_rng.cpp.o): in function `bzla::RNG::RNG(unsigned int)':
rng.cpp:(.text+0x361): undefined reference to `__gmp_randinit_mt'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlals.a(rng_rng.cpp.o): in function `bzla::RNG::RNG(bzla::RNG const&)':
rng.cpp:(.text+0x29): undefined reference to `__gmp_randinit_set'
/usr/bin/ld: /usr/local/lib/x86_64-linux-gnu/libbitwuzlals.a(rng_rng.cpp.o): in function `bzla::RNG::RNG(unsigned int)':
rng.cpp:(.text+0x381): undefined reference to `__gmp_randseed_ui'
collect2: error: ld returned 1 exit status
make[2]: *** [src/esbmc/CMakeFiles/esbmc.dir/build.make:265: src/esbmc/esbmc] Error 1
make[1]: *** [CMakeFiles/Makefile2:2610: src/esbmc/CMakeFiles/esbmc.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

So, modifying the cmake to include and find the gmpxx packages needs to be addressed in this PR, if there is not another way to tackle the compile errors, there are only a few changes. I would be open to any other suggestion of handling these errors. The build command that I use is: cmake .. -DENABLE_Z3=1   -DOVERRIDE_CLANG_HEADER_DIR=/usr/lib/llvm-14/lib/clang/14.0.6/include   -DCMAKE_PREFIX_PATH=/root/bitwuzla/builddir; make -j4. I also tried with cmake .. -DENABLE_Z3=1, and the same errors occur on my machine on the latest version of master, which I was expecting to compile perfectly.

mihaistate:
@XLiZHI I am trying with act 0.2.79:
root@ASUSVivobookS:~/ESBMC/esbmc# act -W .github/workflows/build-unix.yml
INFO[0000] Using docker host 'unix:///var/run/docker.sock', and daemon socket 'unix:///var/run/docker.sock' 
[build-unix.yml/Build ESBMC ()] 🚧  Skipping unsupported platform -- Try running with `-P =...`
root@ASUSVivobookS:~# nano ~/. actrc
root@ASUSVivobookS:~# cd ESBMC
root@ASUSVivobookS:~/ESBMC# cd esbmc/
root@ASUSVivobookS:~/ESBMC/esbmc# act --dryrun -W .github/workflows/build-unix.yml -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest
INFO[0000] Using docker host 'unix:///var/run/docker.sock', and daemon socket 'unix:///var/run/docker.sock' 
*DRYRUN* [build-unix.yml/Build ESBMC ()] 🚧  Skipping unsupported platform -- Try running with `-P =...`
root@ASUSVivobookS:~/ESBMC/esbmc# 
I have some persisting errors, maybe the docker file cannot be established properly. Maybe the command is not the best one. I have Docker configured. 
This is my actrc file: 
-P ubuntu-latest=catthehacker/ubuntu:full-latest
-P ubuntu-22.04=catthehacker/ubuntu:full-22.04
-P ubuntu-20.04=catthehacker/ubuntu:full-20.04
-P ubuntu-18.04=catthehacker/ubuntu:full-18.04
Should I stick with act or try another tool?

XLiZHI:
Hi @mihaistate,

> The idea is that this issue occured only for this branch, for the memcpy implementation, and was related to some compile errors with the gmp package (that I needed to install locally, version 6.2.1, outside the project) and its correlation with bitwuzla. I just tried to comment out all these lines in the cmake file and now I get errors like this again:

You can try to pass this flag when you build: -DENABLE_BITWUZLA=On, you can also skip this solver, there is no need to use all of them.

For CI, here is ctest for regression testing:

https://github.com/esbmc/esbmc/blob/master/regression/README.md

mihaistate:
> Hi @mihaistate,
> 
> > The idea is that this issue occured only for this branch, for the memcpy implementation, and was related to some compile errors with the gmp package (that I needed to install locally, version 6.2.1, outside the project) and its correlation with bitwuzla. I just tried to comment out all these lines in the cmake file and now I get errors like this again:
> 
> You can try to pass this flag when you build: -DENABLE_BITWUZLA=On, you can also skip this solver, there is no need to use all of them.
> 
> For CI, here is ctest for regression testing:
> 
> https://github.com/esbmc/esbmc/blob/master/regression/README.md

mihaistate:
Thanks, I will revert the cmake file to look identical to the master one, as now build is successfull with the Bitwuzla flag on.

XLiZHI:
> Thanks, I will revert the cmake file to look identical to the master one, as now build is successfull with the Bitwuzla flag on.

You can submit another PR for the Bitwuzla build error, I mean this change doesn't have to be here.

mihaistate:
@XLiZHI ok, for now I managed to solve that error with Bitwuzla on, now I am working with the CI.
For [static llvm-16 DebugOpt / Build ESBMC (ubuntu-22.04](https://github.com/esbmc/esbmc/actions/runs/16518127041/job/46713475295?pr=2561#logs) runner, I get these errors: 1174/1631 Test #1177: regression/cbmc/01_cbmc_String2 ....................................***Failed    0.87 sec
test_01_cbmc_String2 (__main__.RegressionBase) ... FAIL

======================================================================
FAIL: test_01_cbmc_String2 (__main__.RegressionBase)
----------------------------------------------------------------------
  File "D:\a\esbmc\esbmc\regression\testing_tool.py", line 251, in test
    self.fail(error_message_prefix + error_message)
AssertionError: 
TEST: D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String2
EXPECTED TO FIND: ['^VERIFICATION FAILED$']

PROGRAM OUTPUT
ESBMC version 7.10.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String2\main.c
D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String2\main.c:10:10: warning: expression result unused [-Wunused-value]
  p[input];
  ~ ~~~~~^
Converting
ERROR: Out of memory
and 
          Start 1178: regression/cbmc/01_cbmc_String3
1175/1631 Test #1178: regression/cbmc/01_cbmc_String3 ....................................***Failed    0.87 sec
test_01_cbmc_String3 (__main__.RegressionBase) ... FAIL

======================================================================
FAIL: test_01_cbmc_String3 (__main__.RegressionBase)
----------------------------------------------------------------------
  File "D:\a\esbmc\esbmc\regression\testing_tool.py", line 251, in test
    self.fail(error_message_prefix + error_message)
AssertionError: 
TEST: D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String3
EXPECTED TO FIND: ['^VERIFICATION FAILED$']

PROGRAM OUTPUT
ESBMC version 7.10.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String3\main.c
Converting
ERROR: Out of memory


ARGUMENTS: ['C:/deps/esbmc/bin/esbmc.exe', 'D:/a/esbmc/esbmc/regression/cbmc\\01_cbmc_String3\\main.c']

----------------------------------------------------------------------
Ran 1 test in 0.805s

FAILED (failures=1), but on my local machine, WSL Ubuntu, where I think I am working with ESBMC 7.9, I get the expected result on the feature branch: oot@ASUSVivobookS:~/ESBMC/esbmc/regression/cbmc/01_cbmc_String2# esbmc main.c 
ESBMC version 7.9.0 64-bit x86_64 linux
Parsing main.c
main.c:10:10: warning: expression result unused [-Wunused-value]
  p[input];
  ~ ~~~~~^
Converting
Generating GOTO Program
GOTO program creation time: 0.334s
Symex completed in: 0.005s (14 assignments)
Generated 1 VCC(s), 1 remaining after simplification (2 assignments)
Solving with solver Z3 v4.8.12
Runtime decision procedure: 0.010s
Building error trace

[Counterexample]


State 1 file main.c line 10 column 3 function main thread 0
Violated property:
  file main.c line 10 column 3 function main

         Start 1178: regression/cbmc/01_cbmc_String3
1175/1631 Test #1178: regression/cbmc/01_cbmc_String3 ....................................***Failed    0.87 sec
test_01_cbmc_String3 (__main__.RegressionBase) ... FAIL

======================================================================
FAIL: test_01_cbmc_String3 (__main__.RegressionBase)
----------------------------------------------------------------------
  File "D:\a\esbmc\esbmc\regression\testing_tool.py", line 251, in test
    self.fail(error_message_prefix + error_message)
AssertionError: 
TEST: D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String3
EXPECTED TO FIND: ['^VERIFICATION FAILED$']

PROGRAM OUTPUT
ESBMC version 7.10.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing D:/a/esbmc/esbmc/regression/cbmc\01_cbmc_String3\main.c
Converting
ERROR: Out of memory

VERIFICATION FAILED   
#802: regression/esbmc/memcpy1_sym .......................................***Failed  254.76 sec
test_memcpy1_sym (__main__.RegressionBase) ... FAIL  How could I make these tests pass also in the CI, as they pass locally?

src/c2goto/library/string.c
```diff
@@ -266,16 +268,35 @@ __ESBMC_HIDE:;
...
 }
 
+void *memcpy(void *dst, const void *src, size_t n)
+{
+__ESBMC_HIDE:;
+  if (n == 0 || src == NULL || dst == NULL)
+    return dst;
+
+  __ESBMC_assume(dst != NULL);
+  __ESBMC_assume(src != NULL);
```
XLiZHI: Why is it needed? 

we already have this right?
```
  if (n == 0 || src == NULL || dst == NULL)
    return dst;
```

src/c2goto/library/string.c
```diff
@@ -266,16 +268,35 @@ __ESBMC_HIDE:;
...
+
+  void *hax = &__memcpy_impl;
+  (void)hax;
```
XLiZHI, line 293: remove this?

Anthonysdu (changes requested):
You don't need to change benchexec file, if you want to benchexec custom testcases there is already an option 'custom'.

src/c2goto/library/string.c
mihaistate: Their purposes are distinct, the if statement ensures runtime safety, it avoids calling memcpy when no copying is needed or pointers are null. The assume statements help symbolic exectuion and solver by ensuring that the remaining execution path only continues when pointers are not null.

src/c2goto/library/string.c
mihaistate: The last two lines represent a method that keep the compiler or the symbolic engine aware of the existence of __memcpy_impl. Some build systems strip unused symbols—even if referenced indirectly in the engine—and leaving this reference avoids dead‑code elimination or linker mis‑optimization.
This ensures the implementation is retained and available at runtime or symbolic execution time—even if the compiler thinks it is unused otherwise.
Deleting this might result in missing symbol errors at compile time or during symbolic execution binding.

mihaistate:
@Anthonysdu I reverted my changes, as they are no longer needed. The point was to increase the CI running time limit to avoid a 143 error, at which tests were stopped by running, without any failed test.

mihaistate:
@XLiZHI Thank you for your support!
The Codacy error in the CI is related to   __ESBMC_assume(dst != NULL);  __ESBMC_assume(src != NULL); assertions, which are needed to tell the model checker that only non-null pointers matter for the symbolic execution of the copy path. Can we somehow keep them for the merge state?

mihaistate:
Dr. @lucasccordeiro this PR is almost ready to be merged.

Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jul 24 21:50 EEST
```
lucasccordeiro: Can you please remove this log?

regression/Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
lucasccordeiro: Remove?

regression/Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jul 24 20:53 EEST
+----------------------------------------------------------
+End testing: Jul 24 20:53 EEST
```
lucasccordeiro: Remove?

regression/esbmc/31_cdaudio/main.c
```diff
@@ -1,6 +1,7 @@
 /* Generated by CIL v. 1.3.6 */
 /* print_CIL_Input is true */
 
+
```
lucasccordeiro: Remove?

lucasccordeiro (changes requested):

lucasccordeiro:
@mihaistate: could you please address my comments?

I'm running this PR over the SV-COMP benchmarks here: https://github.com/esbmc/esbmc/actions/runs/16540322562

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
lucasccordeiro: Remove?

src/goto-symex/builtin_functions.cpp
```diff
@@ -1550,6 +1550,155 @@ void goto_symext::intrinsic_memset(
...
+        constant_int2tc(get_uint64_type(), BigInt(src_offset + i)));
+      expr2tc value = src_idx;
+      dereference(value, dereferencet::READ);
+      symex_assign(code_assign2tc(dst_idx, value), false, guard);
+    }
+  }
+  else
+  {
+    //Copy byte-by-byte
+    for (unsigned i = 0; i < num_bytes; ++i)
```
lucasccordeiro: Shall we use size_t instead of unsigned?

src/goto-symex/builtin_functions.cpp
```diff
@@ -1550,6 +1550,155 @@ void goto_symext::intrinsic_memset(
...
+        constant_int2tc(get_uint64_type(), BigInt(dst_offset + i)));
+      expr2tc src_idx = index2tc(
+        get_uint64_type(),
+        src_item.object,
+        constant_int2tc(get_uint64_type(), BigInt(src_offset + i)));
+      expr2tc value = src_idx;
+      dereference(value, dereferencet::READ);
+      symex_assign(code_assign2tc(dst_idx, value), false, guard);
+    }
+    for (unsigned i = (num_bytes / 8) * 8; i < num_bytes; ++i)
```
lucasccordeiro: shall we use size_t instead of unsigned?

src/goto-symex/builtin_functions.cpp
```diff
@@ -1550,6 +1550,155 @@ void goto_symext::intrinsic_memset(
...
+  {
+    log_debug("memcpy", "Symbolic offsets not supported");
+    bump_call(func_call, "c:@F@__memcpy_impl");
+    return;
+  }
+  uint64_t dst_offset = to_constant_int2t(dst_item.offset).value.to_uint64();
+  uint64_t src_offset = to_constant_int2t(src_item.offset).value.to_uint64();
+
+  //Compute alingnment
+  bool aligned = (dst_offset % 8 == 0) && (src_offset % 8 == 0);
```
lucasccordeiro, line 1647: Why do we use the magic number 8? Shall we consider the char size here? ESBMC has a datatype for it.

src/goto-symex/builtin_functions.cpp
```diff
@@ -1550,6 +1550,155 @@ void goto_symext::intrinsic_memset(
...
+
+  if (aligned)
+  {
+    for (unsigned i = 0; i + 8 <= num_bytes; i += 8)
```
lucasccordeiro: Why do we use the magic number 8? Shall we consider the char size here? ESBMC has a datatype for it.

src/goto-symex/builtin_functions.cpp
```diff
@@ -1550,6 +1550,155 @@ void goto_symext::intrinsic_memset(
...
+
+  // Only handle constant-size copies for now
+  simplify(n);
+  if (!is_constant_int2t(n))
+  {
+    log_debug("memcpy", "Symbolic size not supported, falling back");
+    bump_call(func_call, "c:@F@__memcpy_impl");
+    return;
+  }
+
```
lucasccordeiro, line 1599: Can you add a comment on which situation we handle here? For example, do we just execute the code below if both src and dst point to a valid object?

lucasccordeiro (changes requested):

Testing/Temporary/CTestCostData.txt
mihaistate: It will be removed in my future amended commit

Testing/Temporary/LastTest.log
mihaistate: Yes, I will leave the file empty

regression/Testing/Temporary/CTestCostData.txt
mihaistate: I will

regression/Testing/Temporary/LastTest.log
mihaistate: The file will be empty.

regression/esbmc/31_cdaudio/main.c
mihaistate: I will

src/goto-symex/builtin_functions.cpp
mihaistate: The idea is to find out if we can copy 8 bytes per iteration in the next lines of code to make the symbolic memcpy function memory efficient by reducing the number of iterations. The idea is also taken from the issue itself: https://github.com/esbmc/esbmc/issues/2282. Char is only 1 byte size, there is already a byte-by-byte copy on the else branch of if (aligned)

src/goto-symex/builtin_functions.cpp
mihaistate: The idea is to copy 8 bytes per iterations, so we reduce their number consistenltly, especially when there are thousands of them, and it is also a suggestion of the issue: #2282. Since char is 1 byte, we are already addressing a byte-by-byte copy here, in case the there is no 8-bytes allignment: 
  else
  {
    //Copy byte-by-byte
    for (size_t i = 0; i < num_bytes; ++i)
    {
      expr2tc offset = constant_int2tc(get_uint64_type(), BigInt(i));
      expr2tc dst_idx = index2tc(
        get_uint8_type(),
        dst_item.object,
        constant_int2tc(get_uint64_type(), BigInt(src_offset + i)));
      expr2tc src_idx = index2tc(
        get_uint8_type(),
        src_item.object,
        constant_int2tc(get_uint64_type(), BigInt(src_offset + i)));
      expr2tc value = src_idx;
      dereference(value, dereferencet::READ);
      symex_assign(code_assign2tc(dst_idx, value), false, guard);
    }

regression/Testing/Temporary/LastTest.log
XLiZHI: Please remove the file, It still exists.

Testing/Temporary/LastTest.log
XLiZHI, line 1: Please remove the file, It still exists.

src/c2goto/library/string.c
XLiZHI: I still don't get this "Some build systems strip unused symbols"

Are you using LLM to generate code? What happens if we remove these two lines? can you please explain?

Testing/Temporary/LastTest.log
mihaistate: Ok, should I also remove CTestCostData.txt?

src/c2goto/library/string.c
XLiZHI:
```
if (n == 0 || src == NULL || dst == NULL)
return dst;
```

But you already dealt with the case when src and dst are null. 
It seems to me that the assume will be:
  __ESBMC_assume(true);
  __ESBMC_assume(true);

src/goto-symex/builtin_functions.cpp
mihaistate: Yes, I put size_t

src/goto-symex/builtin_functions.cpp
mihaistate: Yes, I also put size_t here recently.

Testing/Temporary/LastTest.log
XLiZHI: > Ok, should I also remove CTestCostData.txt?

All the empty files you added in this PR should be deleted.

src/goto-symex/builtin_functions.cpp
mihaistate: //Only proceed with this case when when the size `n` is a compile-time constant.
// Otherwise, fall back to byte-by-byte copying in __memcpy_impl

src/c2goto/library/string.c
mihaistate: I will remove them

src/c2goto/library/string.c
mihaistate: I did the function myself, the idea was to call the memcpy_impl to make sure it is taken into consideration by the linker or compiler (I looked at this thread: https://www.reddit.com/r/C_Programming/comments/18fm6fz/what_is_a_good_dummy_statement_to_prevent/?utm_source=chatgpt.com), instead of being at the risk of getting stripped, it is just an optional feature, if we remove it, I think that there wouldn't be any warnings, and the function would still be taken into consideration when called symbolically.

mihaistate: I removed the reference, now the function is more compile time efficient

mihaistate: After removing these lines and the ESBMC_assert, I started to have some failed tests for llvm debug output and build unix runners: 
The following tests FAILED:
	100 - regression/esbmc/00_memcpy_01 (Failed)            esbmc regression
	101 - regression/esbmc/00_memcpy_02 (Failed)            esbmc regression
	102 - regression/esbmc/00_memcpy_04 (Failed)            esbmc regression
	230 - regression/esbmc/bitwise11 (Failed)               esbmc regression
	257 - regression/esbmc/builtin_memcpy (Failed)          esbmc regression
	258 - regression/esbmc/builtin_memcpy-fail (Failed)     esbmc regression
	268 - regression/esbmc/cast_02 (Failed)                 esbmc regression
	283 - regression/esbmc/endianness (Failed)              esbmc regression
	284 - regression/esbmc/endianness_01 (Failed)           esbmc regression
	285 - regression/esbmc/endianness_02 (Failed)           esbmc regression
	330 - regression/esbmc/github_1210-1-array (Failed)     esbmc regression
	331 - regression/esbmc/github_1210-1-struct (Failed)    esbmc regression
	332 - regression/esbmc/github_1210-1-union (Failed)     esbmc regression
	334 - regression/esbmc/github_1210-2-struct (Failed)    esbmc regression
	335 - regression/esbmc/github_1210-2-union (Failed)     esbmc regression
	399 - regression/esbmc/github_1590-generic (Failed)     esbmc regression
	400 - regression/esbmc/github_1590-local (Failed)       esbmc regression
	432 - regression/esbmc/github_1807 (Failed)             esbmc regression
	446 - regression/esbmc/github_1976 (Failed)             esbmc regression
	649 - regression/esbmc/github_569 (Failed)              esbmc regression
	650 - regression/esbmc/github_569_fail (Failed)         esbmc regression
	666 - regression/esbmc/github_591-2 (Failed)            esbmc regression
	667 - regression/esbmc/github_591-3 (Failed)            esbmc regression
	748 - regression/esbmc/github_939 (Failed)              esbmc regression
	804 - regression/esbmc/memcpy1 (Failed)                 esbmc regression
	805 - regression/esbmc/memcpy1_sym (Failed)             esbmc regression
	806 - regression/esbmc/memcpy2 (Failed)                 esbmc regression
	807 - regression/esbmc/memcpy2_sym (Failed)             esbmc regression
	808 - regression/esbmc/memcpy3 (Failed)                 esbmc regression
	809 - regression/esbmc/memcpy3_sym (Failed)             esbmc regression
	928 - regression/esbmc/struct_bitfields_02 (Failed)     esbmc regression
	932 - regression/esbmc/struct_bitfields_06 (Failed)     esbmc regression
	933 - regression/esbmc/struct_bitfields_07 (Failed)     esbmc regression
	934 - regression/esbmc/struct_bitfields_08 (Failed)     esbmc regression
	935 - regression/esbmc/struct_bitfields_09 (Failed)     esbmc regression
	936 - regression/esbmc/struct_bitfields_10 (Failed)     esbmc regression
	937 - regression/esbmc/struct_bitfields_11 (Failed)     esbmc regression
	938 - regression/esbmc/struct_bitfields_12 (Failed)     esbmc regression
	939 - regression/esbmc/struct_bitfields_13 (Failed)     esbmc regression
Errors while running CTest
	940 - regression/esbmc/struct_bitfields_14 (Failed)     esbmc regression
	1296 - regression/cstd/memcpy (Failed)                   cstd regression   So, at least some of the lines were useful, since before removing them, the CI unix build and llvm debug output were passing.

lucasccordeiro:
This PR (30s):

````
  correct:           17614
    correct true:    10334
    correct false:    7280
  incorrect:            34
    incorrect true:     13
    incorrect false:    21
  unknown:           15921
  Score:             27196 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16540322562/
````


Master (30s):

````
  correct:           17620
    correct true:    10341
    correct false:    7279
  incorrect:            34
    incorrect true:     13
    incorrect false:    21
  unknown:           15915
  Score:             27209 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16404623531
````

lucasccordeiro:
I have started another SV-COMP run: https://github.com/esbmc/esbmc/actions/runs/16548449240

lucasccordeiro:
@mihaistate: there are many regression tests failing. Could you **check**?

mihaistate:
@lucasccordeiro I commited a change that could make the regression tests pass again, as they did at the previous commit.  
Then, if they pass, I would probably need to run the SV-COMP tests again. Thanks
So, the results of this PR are: 
  correct:           17614
    correct true:    10334
    correct false:    7280
and the results of master are: 
  correct:           17620
    correct true:    10341
    correct false:    7279

I guess that this PR's results based on the https://github.com/esbmc/esbmc/commit/dde4bd64ad2527ef2b50c0794dd089a300c8f99b commit are good, they are almost as good as the master one, would I need need to make an extra test pass?

lucasccordeiro:
No.

We just need to keep our regression and the SV-COMP run.

src/c2goto/library/string.c
mihaistate: As we can see the hax method is also in the memset function, already implemented from master:
void *memset(void *s, int c, size_t n)
{
__ESBMC_HIDE:;
  void *hax = &__memset_impl;
  (void)hax;
  return __ESBMC_memset(s, c, n);
}  so I need to have these two lines as well in my memcpy implementation, without it, the memcpy_impl can be ignored with this file is compiled and even removed somehow from compilation stage if not called in the memcpy function

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @mihaistate.

I have just started another SV-COMP run at https://github.com/esbmc/esbmc/actions/runs/16552122501.

Once this run is completed, I'll merge this PR.

Anthonysdu:
There are still two files should be removed!

mihaistate:
> There are still two files should be removed!

Which ones?

Anthonysdu:
> > There are still two files should be removed!
> 
> Which ones?

See my review on top, what are the two bin/act files?

mihaistate:
I am going to do something about them here: #2642

src/c2goto/library/string.c
intrigus-lgtm: > remove this?

This is necessary so that whenever the `memcpy` implementation is called, we also include `__memcpy_impl` so that we can bump failing calls to the intrinsic (missing preconditions for intrinsic) to `__memcpy_impl`.

mihaistate: @XLiZHI please look at the results, and also take intrigus's argument into account

PR #2562, opened 2025-07-01
[GOTO guard]  constant propagation for assume and add warning for GOTO guard

### LINKED ISSUE #2552: Handling of extern variables
Could you please clarify the way ESBMC handles extern variables?
For the following program ESBMC returns `VERIFICATION SUCCESSFULL`:
```c
#include <assert.h>
extern void __VERIFIER_assume(int cond);
extern int x;
int main(int argc, char **argv) {
__VERIFIER_assume(argc == 1);
if(argc > 5) {x = 42;}
assert(x == 42);
}
```
intuitively I would have expected ESBMC to report the assertion as possibly failing, considering the value of `x` as nondeterministic (since the `if` is never reached). Or is this not the case?
Stragely using the option `--smt-symex-guard` I do get the result I was expecting (`VERIFICATION FAILED`).

Version: ESBMC version 7.9.0 64-bit x86_64 linux
Built from 28e4e89c777b1e74dd8ed7f80892d6d1fd7b64bb (latest)

XLiZHI:


src/goto-symex/symex_goto.cpp
```diff
@@ -26,28 +26,33 @@ void goto_symext::symex_goto(const expr2tc &old_guard)
...
+      catch (runtime_encoded_equationt::dual_unsat_exception &e)
+      {
+        // If reach here it means both guard G and !G are unsatisfiable,
+        // basically means we can't prove the guard must be true or must be false.
+        new_guard_false = false;
+      }
     }
+    else
+      log_warning(
+        "The GOTO guard cannot be determined during symbolic execution, "
```
lucasccordeiro: "Unable to determine the GOTO guard during symbolic execution. To proceed, please use the option: --smt-symex-guard."

lucasccordeiro (changes requested):

Anthonysdu:
BTW is it necessary to add this warning? I think it's should be normal if the goto guard can't be determined in symex.

src/goto-symex/symex_goto.cpp
XLiZHI: Done

XLiZHI:
> BTW is it necessary to add this warning? I think it's should be normal if the goto guard can't be determined in symex.

Yes, I think this is the right approach, but we should experiment first, not sure how much performance will be affected.

Anthonysdu:
> > BTW is it necessary to add this warning? I think it's should be normal if the goto guard can't be determined in symex.
> 
> Yes, I think this is the right approach, but we should experiment first, not sure how much performance will be affected.

I mean this will bascially suggest the developer to use incremental-smt solving which doesn't guarantee to achieve verification speed up and in lots of cases are even slower, otherwise there would be huge number of warnings in the log.

For example in this case:
`esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check ../../sv-benchmarks/c/hardness-nfm22/hardness_loopvsstraightlinecode_50-while_file-73.i --32 --witness-output witness.graphml --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --unwind 3`

Interval Analysis time: 0.033s
GOTO program processing time: 0.036s
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
Unwinding loop 2 iteration 1   file ../../sv-benchmarks/c/hardness-nfm22/hardness_loopvsstraightlinecode_50-while_file-73.i line 137 column 2 function main
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard
WARNING: The GOTO guard cannot be determined during symbolic execution, please using: --smt-symex-guard

XLiZHI:
Are incremental-SMT and smt-symex-guard bindings necessary? Is it possible to use smt-symex-guard alone? It is also useful in single-threaded programs. Currently, if we can't determine goto guard during symex, esbmc will treat guard as true, which causes a lot of errors.

Also, maybe I can find a way to only have the warning appear once.

Anthonysdu:
> Are incremental-SMT and smt-symex-guard bindings necessary? 

smt-symex-guard is an Incremental SMT approach. Check here: https://github.com/esbmc/esbmc/blob/1792a59dd08739eafd6042e1bc2f92baa1c741a6/src/goto-symex/symex_target_equation.cpp#L519



> Currently, if you can't determine goto guard, esbmc will treat guard as true, which causes a lot of errors.

What do you mean the errors? Currently if esbmc can't determine the goto guard it will fork the current path into two separate branches to simulate conditional execution based on the guard expression.

XLiZHI:
Please see the example in PR:

When handling an assignment, if cur_state guard is not false, the assignment/function call is executed and is treated as a true guard by default. 

This is the problem, if we can't identify guard, it could be executed under multiple paths and do constant propagation (recording constant values). Constant propagation does not take guards into account.

```
VERIFICATION CONDITIONS:

file main5.c line 7 column 1 function main
assertion x == 42
{-1} argc'&0#0 >= 1
{-2} argc'&0#0 <= 268435454
{-3} argc?1!0&0#1 == argc'&0#0
{-4} (_Bool)((signed int)(argc?1!0&0#1 == 1))
|--------------------------
{1} execution_statet::\guard_exec?0!0 => 42 == 42         ---> rename here, should be x == 42
```

Anthonysdu:
I see there is problem when doing constant propagation and we should find a fix to it, but for the warnings, we can either keep one or just remove them, what do you think? @lucasccordeiro

lucasccordeiro:
> I see there is problem when doing constant propagation and we should find a fix to it, but for the warnings, we can either keep one or just remove them, what do you think? @lucasccordeiro

We should first find a fix to this issue.

For the warning, we could print it just once. It would be nice to warn the user about this issue.

lucasccordeiro:
@XLiZHI: After you fix the warning message to print it once, can you please run this PR over the sv-comp benchmarks?

XLiZHI:
This PR:
```
  correct:           17703
    correct true:    10403
    correct false:    7300
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15831
  Score:             27322 (max: 55885)
 https://github.com/esbmc/esbmc/actions/runs/16009026858/job/45163289682
```

Master:
```
  correct:           17650
    correct true:    10355
    correct false:    7295
  incorrect:            35
    incorrect true:     14
    incorrect false:    21
  unknown:           15884
  Score:             27221 (max: 55885)
https://github.com/esbmc/esbmc/actions/runs/16009023775/job/45163326876
```

lucasccordeiro (approved):

Anthonysdu (approved):

lucasccordeiro:

intrigus-lgtm:
I'm a bit late to the party, but why don't we just follow what CBMC is doing and initialize the extern variable to nondet if the user didn't provide a value?

`cbmc` successfully fails and is also more robust than our current solution which e.g. fails when using `--no-simplify`.
```
cbmc wtf.c --show-goto-functions
CBMC version 5.95.1 (cbmc-5.95.1) 64-bit x86_64 linux
Parsing wtf.c
Converting
Type-checking wtf
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation

__CPROVER__start /* __CPROVER__start */
        // 13 no location
        // Labels: __CPROVER_HIDE
        CALL __CPROVER_initialize()
        // 14 no location
        ASSUME argc' ≥ 0
        // 15 no location
        INPUT address_of("argc"[0]) argc'
        // 16 
        ASSIGN argv'[cast(argc', signedbv[64])] := NULL
        // 17 file wtf.c line 4
        CALL return' := main(argc', address_of(argv'[0]))
        // 18 file wtf.c line 4
        OUTPUT address_of("return'"[0])
        // 19 no location
        END_FUNCTION

__CPROVER_initialize /* __CPROVER_initialize */
        // 6 file <built-in-additions> line 8
        // Labels: __CPROVER_HIDE
        ASSIGN __CPROVER_dead_object := NULL
        // 7 file <built-in-additions> line 7
        ASSIGN __CPROVER_deallocated := NULL
        // 8 file <built-in-additions> line 12
        ASSIGN __CPROVER_max_malloc_size := 36028797018963968
        // 9 file <built-in-additions> line 9
        ASSIGN __CPROVER_memory_leak := NULL
        // 10 file <built-in-additions> line 16
        ASSIGN __CPROVER_rounding_mode := 0
        // 11 file wtf.c line 3
        ASSIGN x := side_effect statement="nondet" is_nondet_nullable="1"
        // 12 no location
        END_FUNCTION

main /* main */
        // 0 file wtf.c line 5 function main
        ASSUME cast(cast(main::argc = 1, signedbv[32]), 𝔹)
        // 1 file wtf.c line 6 function main
        IF ¬(main::argc > 5) THEN GOTO 1
        // 2 file wtf.c line 6 function main
        ASSIGN x := 42
        // 3 file wtf.c line 7 function main
     1: ASSERT x = 42 // assertion x == 42
        // 4 file wtf.c line 8 function main
        SET RETURN VALUE side_effect statement="nondet" is_nondet_nullable="1"
        // 5 file wtf.c line 8 function main
        END_FUNCTION
```

lucasccordeiro:
Thanks, @intrigus-lgtm.

This PR is not intended to fix the extern variable issue.

The goal here is to perform constant propagation for assume. 

@XLiZHI should have written a more detailed description and not linked the commit to the extern issue.

XLiZHI:
Thanks for explanation, Prof @lucasccordeiro .

Let me fill in some details here:

> but why don't we just follow what CBMC is doing and initialize the extern variable to nondet if the user didn't provide a value?

The external variable x is indeed nondet in esbmc, there isn't any initialization for x in the front end, so it's nondet in the SMT formula. In my understanding, if a variable is explicitly initialized as nondet, it is also removed during the symex, so there is no difference here.

https://github.com/esbmc/esbmc/blob/31607d7fc29eb31a51684f070bc04911b23df6e8/src/goto-symex/symex_assign.cpp#L850-L866

The main cause of this issue is goto guard. Constant propagation is used incorrectly due to the inability to decide on guard at symex. As a result, the error VCC is generated after rename.

Like this example:
```
int x = nondet_int();
bool guard = nondet_bool()
__ESBMC_assume(guard);
if (guard)
 x = 42;           --> assignment is executed here

assert(x == 42);   --> after rename, it become 42 == 42 -> true
```

intrigus-lgtm:
> Like this example:
> 
> ```
> int x = nondet_int();
> bool guard = nondet_bool()
> __ESBMC_assume(guard);
> if (guard)
>  x = 42;           --> assignment is executed here
> 
> assert(x == 42);   --> after rename, it become 42 == 42 -> true
> ```

What's wrong with this example?
If we assume `guard` to be `true` than we will of course assign `42` to `x` and so the assertion is always correct.

If we change the example to this:
```c
#include <stdbool.h>

int main()
{
    int x = nondet_int();
    int guard = nondet_int();
    __ESBMC_assume(guard == 5);
    if (guard == 4)
        x = 42;

    assert(x == 42);
}
```
Then we'll get a counter-example as expected?
```
[Counterexample]


State 1 file /tmp/wtf.c line 5 column 5 function main thread 0
  x = 43 (00000000 00000000 00000000 00101011)

State 2 file /tmp/wtf.c line 6 column 5 function main thread 0
  guard = 5 (00000000 00000000 00000000 00000101)

State 4 file /tmp/wtf.c line 11 column 5 function main thread 0
Violated property:
  file /tmp/wtf.c line 11 column 5 function main
  assertion
  x == 42


VERIFICATION FAILED
```

PR #2568, opened 2025-07-02
[Solidity] support selfdestruct

ChenfengWei0:
`selfdestruct` is kind of being deprecated starting from solc version 0.8.24 (the latest version is 0.8.31). Considering how widely it has been used, I think we should add support for it — it's not difficult to support anyway.


lucasccordeiro (approved):

lucasccordeiro:

PR #2582, opened 2025-07-04
[smt] add rational value extraction for floating-point counterexamples in int_encoding mode

lucasccordeiro:
In integer/real arithmetic mode, floating point values are represented as rational numbers in the SMT solver, but `get_by_ast()` was returning empty expressions, causing counterexamples to show "(assignment removed)" instead of actual values.

Changes:
- Add virtual `get_rational()` method to `smt_convt` interface.
- Implement rational-to-floating-point conversion in get_by_ast() for floatbv types.
- Provide default implementation that return false (to be overridden by solvers).

This enables the proper display of floating-point variable values in counterexamples when using integer/real arithmetic mode. Solver-specific implementations of the new virtual method is required for each SMT backend. However, I have implemented it for Z3 only here.

### DIFF
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -2620,30 +2620,37 @@ expr2tc smt_convt::get_by_ast(const type2tc &type, smt_astt a)
   }
 }
 
-double smt_convt::convert_rational_to_double(const BigInt &numerator, const BigInt &denominator)
+double smt_convt::convert_rational_to_double(
+  const BigInt &numerator,
+  const BigInt &denominator)
 {
   size_t buffer_size = 1024;
-  
+
   std::vector<char> num_buffer(buffer_size);
   std::vector<char> den_buffer(buffer_size);
-  
+
   numerator.as_string(num_buffer.data(), buffer_size, 10);
   denominator.as_string(den_buffer.data(), buffer_size, 10);
-  
+
   // Check for buffer overflow
   size_t num_len = strlen(num_buffer.data());
   size_t den_len = strlen(den_buffer.data());
-  
-  if (num_len >= buffer_size - 1 || den_len >= buffer_size - 1) {
-    log_warning("BigInt to string conversion may have been truncated - buffer too small");
+
+  if (num_len >= buffer_size - 1 || den_len >= buffer_size - 1)
+  {
+    log_warning(
+      "BigInt to string conversion may have been truncated - buffer too small");
     return 0.0; // Safer to return 0 than use truncated data
   }
-  
-  try {
+
+  try
+  {
     double num_val = std::stod(num_buffer.data());
     double den_val = std::stod(den_buffer.data());
     return (den_val != 0.0) ? num_val / den_val : 0.0;
-  } catch (const std::exception &) {
+  }
+  catch (const std::exception &)
+  {
     return 0.0;
   }
 }
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -2541,8 +2541,25 @@ expr2tc smt_convt::get_by_ast(const type2tc &type, smt_astt a)
   case type2t::floatbv_id:
     if (int_encoding)
     {
-      /* TODO: how to retrieve an floatbv from a rational or algebraic real
-       * number in a meaningful way? */
+      BigInt numerator, denominator;
+      if (get_rational(a, numerator, denominator)) 
+      {
+        double value = convert_rational_to_double(numerator, denominator);
+        unsigned int type_width = type->get_width();
+      
+        if (type_width == ieee_float_spect::single_precision().width()) 
+        {
+          ieee_floatt ieee_val(ieee_float_spect::single_precision());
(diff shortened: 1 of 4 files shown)


lucasccordeiro:
Consider this C program:

````C
#include <assert.h>
int main()
{
  float f;
  double d;
  unsigned char x;

  d = f;

  if (f == x)
    assert(d != x);
}
````

Before this PR:

````
[Counterexample]


State 1 file main.c line 4 column 3 function main thread 0
  f(assignment removed)
State 2 file main.c line 6 column 3 function main thread 0
  x = 0

State 3 file main.c line 8 column 3 function main thread 0
  d = (double)f

State 4 file main.c line 11 column 5 function main thread 0
Violated property:
  file main.c line 11 column 5 function main
  assertion d != x
  d != (double)((signed int)x)


VERIFICATION FAILED
````

After this PR:

````
[Counterexample]


State 1 file main.c line 4 column 3 function main thread 0
  f = 0.000000f

State 2 file main.c line 6 column 3 function main thread 0
  x = 0

State 3 file main.c line 8 column 3 function main thread 0
  d = 0.000000

State 4 file main.c line 11 column 5 function main thread 0
Violated property:
  file main.c line 11 column 5 function main
  assertion d != x
  d != (double)((signed int)x)


VERIFICATION FAILED
````

File: src/solvers/z3/z3_conv.cpp [hunk 1]
```diff
@@ -1318,6 +1318,70 @@ BigInt z3_convt::get_bv(smt_astt a, bool is_signed)
...
+      Z3_ast evaluated;
+      bool eval_success =
+        Z3_model_eval(z3_ctx, current_model, za->a, true, &evaluated);
+
+      if (eval_success && evaluated != nullptr)
+      {
+        // Check if the evaluated result is a numeral
+        if (Z3_get_ast_kind(z3_ctx, evaluated) == Z3_NUMERAL_AST)
+        {
+          Z3_string str = Z3_get_numeral_string(z3_ctx, evaluated);
```
XLiZHI: I don't know the Z3 API very well, so I'm not sure how robust this approach is. Can we use following method?

```
    /**
       \brief Similar to #Z3_get_numeral_string, but only succeeds if
       the value can fit as a rational number as machine \c int64_t int. Return \c true if the call succeeded.

       \pre Z3_get_ast_kind(c, v) == Z3_NUMERAL_AST

       \sa Z3_get_numeral_string

       def_API('Z3_get_numeral_rational_int64', BOOL, (_in(CONTEXT), _in(AST), _out(INT64), _out(INT64)))
    */
    bool Z3_API Z3_get_numeral_rational_int64(Z3_context c, Z3_ast v, int64_t* num, int64_t* den);
```

lucasccordeiro: Implemented as suggested.

src/solvers/smt/smt_conv.cpp
```diff
@@ -2595,6 +2620,45 @@ expr2tc smt_convt::get_by_ast(const type2tc &type, smt_astt a)
...
+
+  try
+  {
+    double num_val = std::stod(num_buffer.data());
+    double den_val = std::stod(den_buffer.data());
+    return (den_val != 0.0) ? num_val / den_val : 0.0;
+  }
+  catch (const std::exception &)
+  {
+    return 0.0;
```
intrigus-lgtm, line 2658: When do we expect an exception to be thrown here?
Shouldn't we log a warning in that case?

lucasccordeiro: Thanks, @intrigus-lgtm.

Can I ask you to submit a PR with your suggestion?

intrigus-lgtm: See https://github.com/esbmc/esbmc/pull/2595

lucasccordeiro: Thank you.

lucasccordeiro:
@intrigus-lgtm: Here is an example that triggers the warning:

````C
#include <assert.h>
#include <float.h>
#include <math.h>

int main() {
    double x = nondet_double();
    double y = nondet_double();

    __ESBMC_assume(x > 0.0 && x < 1.0e-154);
    __ESBMC_assume(y > 0.0 && y < 1.0e-154);

    double result = x * y;

    // Underflow check
    assert(result > 0.0);

    return 0;
}
````

$esbmc main.c --ir

````
Building error trace
WARNING: Failed to convert BigInt to double: numerator=, denominator=
WARNING: Failed to convert BigInt to double: numerator=, denominator=
WARNING: Failed to convert BigInt to double: numerator=, denominator=
WARNING: Failed to convert BigInt to double: numerator=, denominator=

[Counterexample]


State 1 file main.c line 6 column 5 function main thread 0
  x = 0.000000

State 2 file main.c line 7 column 5 function main thread 0
  y = 0.000000

State 5 file main.c line 12 column 5 function main thread 0
  result = 0.000000

State 6 file main.c line 15 column 5 function main thread 0
Violated property:
  file main.c line 15 column 5 function main
  assertion result > 0.0
  result > 0.000000


VERIFICATION FAILED
````

PR #2583, opened 2025-07-04
[Concurrency] Rename expr as level1 in globals_list

Anthonysdu:


Anthonysdu:
This fixes a bug in mpor, where there should be variable dependency while the expression in different rename levels makes it independent and interleavings was wrongly cut by mpor.
For example:
```c
#include<pthread.h>
#include<assert.h>
#include<stdlib.h>
void assume_abort_if_not(int b) {
  if (!b)
    abort();
}
int e, f, g = 0;
void *thread1() {
  e = 1;   //ASSIGN rename_level(e) stored == 0
  g = 3;
  return NULL;
}
void *thread2() {
  assume_abort_if_not(e == 0); // FUNCRION_CALL rename_level(e) == 1, mismatching shared information and wrongly cut by mpor
  f++;
  if (g)
    assert(0);
  return NULL;
}
int main() {
  pthread_t t,t1;
  pthread_create(&t, 0, thread1, 0);
  pthread_create(&t1, 0, thread2, 0);
  return 0;
}
```

lucasccordeiro (approved):

lucasccordeiro:

PR #2586, opened 2025-07-05
[bmc] add colored summary and results for multi-property verification

lucasccordeiro:
This PR:
- Shows colored ✓/✗ indicators for individual claim results
- Displays summary with pass/fail counts, timing, and solver info
- Uses green for passed claims, red for failures

Here is the output for VERIFICATION FAILED:

image

image

VERIFICATION SUCCESSFUL:

image

### DIFF
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -1511,10 +1511,10 @@ void bmct::report_simple_summary(const SimpleSummary &summary) const
 
   std::ostringstream timing_oss;
   timing_oss << "Solver: " << summary.solver_name
-             << " • Decision procedure total time: " << time2string(summary.total_time_s)
-             << "s"
-             << " • Avg: " << std::fixed << std::setprecision(1) << time2string(avg_time)
-             << "s/property";
+             << " • Decision procedure total time: "
+             << time2string(summary.total_time_s) << "s"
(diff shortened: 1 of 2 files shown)


ChenfengWei0 (commented):
I think we can utilize the rules defined in the `--color` mode. Like
image

src/esbmc/bmc.cpp
```diff
@@ -1443,3 +1481,42 @@ smt_convt::resultt bmct::multi_property_check(
...
+  // Build the properties summary string with colors
+  std::ostringstream properties_oss;
+  properties_oss << "Properties: " << summary.total_properties << " verified";
+
+  if (summary.passed_properties > 0)
+    properties_oss << " " << GREEN << "✓ " << summary.passed_properties
+                   << " passed" << RESET;
+
+  if (summary.failed_properties > 0)
+    properties_oss << ", " << RED << "✗ " << summary.failed_properties
```
ChenfengWei0, line 1511: Like for example, we can use `log_error` here to make the output in red.

Or we can keep this implementation, but only enable it when enabling `--color`, if we don't want the `[Error]` prefix

lucasccordeiro: Thanks, @ChenfengWei0. Could you please submit a small PR with the --color suggestion?

PR #2587, opened 2025-07-06
[multi-property] fix clear verifired claims

ChenfengWei0:
```
int main()
{
    int x = 0;

    assert(x < 0);
    while(++x<10)
        assert(x<5);
}
```

Bug fix for things like:
c763974dd391c17a9c261774f6520c0e

where `1` should be true, but still got verified as failed.

```sh
# before
Properties: 6 verified, √3 skipped, X3 failed
# after
Properties: 6 verified, √4 skipped, X2 failed
```


lucasccordeiro:

PR #2592, opened 2025-07-07
[C++ verification] improve C++ string, memory and vector

XLiZHI:
There are significant differences between OM and standard library implementations. It is generally difficult to use them together. This PR tries to adjust the OM to be as close as possible to the standard library implementation and API.


lucasccordeiro (approved):

lucasccordeiro:

PR #2596, opened 2025-07-09
[smt] add IEEE 754 compliance for floating-point operations in real encoding

### LINKED ISSUE #2590: --ir  fails to detect float underflow
The assertion incorrectly passes with the (--ir)  but correctly fails with (--floatbv).

```
#include <assert.h>
#include <float.h>
#include <math.h>
 
int main() {
    double x = nondet_double();
    double y = nondet_double();
 
    __ESBMC_assume(x > 0.0 && x < 1.0e-154);
    __ESBMC_assume(y > 0.0 && y < 1.0e-154);
 
    double result = x * y;
 
    // Underflow check
    assert(result > 0.0);  
 
    return 0;
}
```

Run with ./src/esbmc/esbmc float_underflow.c --ir --z3

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2590.

This PR implements IEEE 754 underflow, overflow, and subnormal number handling for floating-point operations (add, sub, mul, div, fma) when using integer/real encoding mode. Previously, these operations were treated as exact real arithmetic,
missing IEEE 754 behaviors such as underflow to zero.

### DIFF
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -254,6 +254,48 @@ smt_astt smt_convt::convert_assign(const expr2tc &expr)
   return side2;
 }
 
+smt_astt smt_convt::get_zero_real()
+{
+  // Returns SMT representation of zero (0.0)
+  return mk_smt_real("0");
+}
+
+smt_astt smt_convt::get_double_min_normal()
+{
+  // IEEE 754 double precision minimum normal positive value (2^-1022)
+  return mk_smt_real("2.2250738585072014e-308");
+}
+
+smt_astt smt_convt::get_double_min_subnormal()
+{
+  // IEEE 754 double precision minimum positive subnormal value (2^-1074)
+  return mk_smt_real("4.9406564584124654e-324");
+}
+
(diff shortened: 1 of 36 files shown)


src/solvers/smt/smt_conv.cpp
```diff
@@ -254,6 +254,102 @@ smt_astt smt_convt::convert_assign(const expr2tc &expr)
...
+  auto single_spec = ieee_float_spect::single_precision();
+
+  smt_astt min_normal, min_subnormal, max_normal;
+
+  // IEEE 754 double precision (64-bit): 11 exponent bits, 52 fraction bits
+  if (exponent_bits == double_spec.e && fraction_bits == double_spec.f)
+  {
+    min_normal = mk_smt_real("2.2250738585072014e-308");    // 2^(-1022)
+    min_subnormal = mk_smt_real("4.9406564584124654e-324"); // 2^(-1074)
+    max_normal =
```
XLiZHI: Some of these are used multiple times in different places, I propose to define them uniformly in some places. Does this make the code more readable?

PR #2603, opened 2025-07-11
[C++ verification] add OM for random

XLiZHI:


lucasccordeiro:

src/cpp/library/random
```diff
@@ -0,0 +1,179 @@
...
+public:
+  using result_type = unsigned int;
+
+  random_device()
+  {
+  }
+
+  result_type operator()()
+  {
+    result_type val = nondet_uint();
```
intrigus-lgtm, line 71: Why did you not use `nondet_util::nondet_helper<result_type>::get();` here?

src/cpp/library/random
```diff
@@ -0,0 +1,179 @@
...
+
+  explicit mersenne_twister_engine(result_type seed = 5489u)
+  {
+    // Ignoring seeds
+  }
+
+  result_type operator()()
+  {
+    result_type val = nondet_util::nondet_helper<result_type>::get();
+    __CPROVER_assume(val >= min() && val <= max());
```
intrigus-lgtm, line 115: Why the sudden use of `__CPROVER_assume` instead of `__ESBMC_assume`?

XLiZHI: They are same.

src/cpp/library/random
XLiZHI: Because it's a fixed type.

src/cpp/library/random
intrigus-lgtm: I know, I was just surprised to see the sudden switch.
But doesn't really matter :)

src/cpp/library/random
intrigus-lgtm: Yeah, okay, I see where you're coming from.

src/cpp/library/random
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  static constexpr result_type min()
+  {
+    return 0;
+  }
+
+  static constexpr result_type max()
+  {
+    return 0x7FFFFFFF;
```
intrigus-lgtm, line 83: Isn't this wrong?
https://en.cppreference.com/w/cpp/numeric/random/random_device/max.html
https://godbolt.org/z/7aa4fv86d

intrigus-lgtm: @XLiZHI

lucasccordeiro: @intrigus-lgtm : You could submit a small PR to fix this wrong assumption.

XLiZHI: It should be 0xFFFFFFFFU for unsigned int. You are right. #2610

PR #2609, opened 2025-07-14
[Concurrency] Fix pthread_rwlock opreational model

Anthonysdu:


Anthonysdu:
So far ESBMC treats rw_lock similar to mutex, but as from the document: https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_rwlock_rdlock.html:

> A thread may hold multiple concurrent read locks on rwlock (that is, successfully call the pthread_rwlock_rdlock() function n times).

lucasccordeiro (changes requested):
Can you add regression tests to check for the different scenarios?

lucasccordeiro (approved):

lucasccordeiro:

PR #2612, opened 2025-07-15
[C verification] redesign of realloc

XLiZHI:
This PR splits realloc into malloc and internally implemented array copy.

There might be a new implementation of memcpy, and I think it might be possible to reuse the same code.


lucasccordeiro:
Does your realloc PR handle this case?

https://github.com/esbmc/esbmc/issues/1982

PR #2624, opened 2025-07-17
[python] Handle super() call

brcfarias:
Covering features from [regression/python/class10/main.py](https://github.com/esbmc/esbmc/blob/master/regression/python/class10/main.py)


lucasccordeiro (approved):

lucasccordeiro:

PR #2625, opened 2025-07-17
[python] allow the ord function to accept variables as inputs

Ben-Eichhoefer:
Previously the ord function could only be verified if it had a constant as input: `a = ord("a")`

This pr extends that to allow code such as the snippet below to be verified by looking up the variable and replacing the 'Name' node with a 'Constant' node
```
a = "a"
b = ord("a")
assert a == 97
```

This PR also includes some other unrelated regression tests that currently fail.


lucasccordeiro (approved):

lucasccordeiro:

Issue #2618, opened 2025-07-16
[parallel-solving] `mk_select` failed in boolector/z3


ChenfengWei0:
```c
#include <assert.h>
#include <string.h>
#include <stdlib.h>
struct InnerStruct {
    char innerValue[10];
};

int main() {
    struct InnerStruct array[2];

    strcpy(array[0].innerValue, "20");
    strcpy(array[1].innerValue, "40");

    for (int i = 0; i < 2; i++) {
        int innerValue = atoi(array[i].innerValue);
        assert(innerValue == 40);
    }

    return 0;
}
```

- result (occasionally):
```
esbmc: /root/workspace/ESBMC_Commit/esbmc/src/solvers/boolector/boolector_conv.cpp:508: virtual const smt_ast* boolector_convt::mk_select(smt_astt, smt_astt): Assertion `a->sort->get_domain_width() == b->sort->get_data_width()' failed.
[New Thread 0x7fffc6ffd640 (LWP 535732)]
```

- gdb dump:

```sh
#7  0x0000555556201625 in boolector_convt::mk_select (this=0x7fffd401dcc0, a=0x7fffd40cf7b0,
    b=0x7fffd40cfa80)
    at /root/workspace/ESBMC_Commit/esbmc/src/solvers/boolector/boolector_conv.cpp:508
#8  0x00005555561a56cf in smt_ast::select (this=0x7fffd40cf7b0, ctx=0x7fffd401dcc0, idx=...)
    at /root/workspace/ESBMC_Commit/esbmc/src/solvers/smt/smt_conv.cpp:3320
#9  0x00005555561a00f6 in smt_convt::convert_array_index (this=0x7fffd401dcc0, expr=...)
    at /root/workspace/ESBMC_Commit/esbmc/src/solvers/smt/smt_conv.cpp:2368
#10 0x000055555619752f in smt_convt::convert_ast (this=0x7fffd401dcc0, expr=...)
    at /root/workspace/ESBMC_Commit/esbmc/src/solvers/smt/smt_conv.cpp:858
#11 0x00005555561939b4 in smt_convt::convert_assign (this=0x7fffd401dcc0, expr=...)
--Type <RET> for more, q to quit, c to continue without paging--
    at /root/workspace/ESBMC_Commit/esbmc/src/solvers/smt/smt_conv.cpp:245
#12 0x0000555555c6e134 in symex_target_equationt::convert_internal_step (this=0x7fffe67fbae0,
[...]

PR #2632, opened 2025-07-21
[multi-property] fixed assertion message

### LINKED ISSUE #2629: [multi-property] No failure dereference
For this code 
```c
#include <stdlib.h>

int main() {
    int *ptr = malloc(sizeof(int));
    int choice;
    
    if (choice == 1) {
        free(ptr);        
    } else {
        *ptr = 42;        // potential NULL deref
    }
    
    return 0;
}

```
Here is the command and the result:
```c 
esbmc 2.c 
```
```c
[Counterexample]


State 1 file 2.c line 4 column 5 function main thread 0
  ptr = (signed int *)0

State 2 file 2.c line 5 column 5 function main thread 0
  choice = 0 (00000000 00000000 00000000 00000000)

State 3 file 2.c line 10 column 9 function main thread 0
Violated property:
  file 2.c line 10 column 9 function main
  dereference failure: NULL pointer


VERIFICATION FAILED

```
Here is the command and the result:
```c
esbmc 2.c --color --parallel-solving
```
It is hard to tell what the fail '1' is pointing to.
```c
[PROGRESS] Solving with solver Boolector 3.2.3
✓ PASSED: '1'
✓ PASSED: '1'
✓ PASSED: '1'
Runtime decision procedure: 0.002s
✗ FAILED: '1'

VERIFICATION FAILED

```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2629.

### DIFF
--- a/regression/esbmc/github_2629/main.c
+++ b/regression/esbmc/github_2629/main.c
@@ -1,0 +1,14 @@
+#include <stdlib.h>
+
+int main() {
+    int *ptr = malloc(sizeof(int));
+    int choice;
+
+    if (choice == 1) {
+        free(ptr);
+    } else {
+        *ptr = 42;        // potential NULL deref
+    }
+
+    return 0;
+}
--- a/regression/esbmc/github_2629/test.desc
+++ b/regression/esbmc/github_2629/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--multi-property
+^\✓ PASSED: \'dereference failure: invalidated dynamic object freed\'$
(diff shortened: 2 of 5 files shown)


intrigus-lgtm (changes requested):
I don't understand how this PR could fix e.g. #2618
All it basically does is switching the then and else branches?
How would that fix a concurrency problem?

lucasccordeiro:
You're correct. I have just removed it from the comment above.

ChenfengWei0:
I'm not sure if this is the correct fix, but one thing I noticed is that `is_goto_cov` may not be correct or at least misleading. It seems it should be something like `is_assert_cov`, since we only convert assertions to assert(0) in assertion coverage mode to check reachability.

ChenfengWei0 (approved):
LGTM

lucasccordeiro:
Here is the new output:

image

image

PR #2637, opened 2025-07-24
[c++] add template sort(Compare) overload to list implementation

lucasccordeiro:
This PR adds `template<class Compare> void sort(Compare comp)` method that uses the same selection sort algorithm as the existing parameterless `sort()`, but applies the custom comparison function.


XLiZHI (approved):

lucasccordeiro:
I have just added a small test case to test the sort method for list: https://github.com/esbmc/esbmc/commit/5533e3ff1c8f2440dd7b18137b6fcb5416df7dd3

PR #2638, opened 2025-07-25
[symex] assume assertions after checking to strengthen path conditions (EXPERIMENTAL!!!)

lucasccordeiro:
This PR adds `assume(new_expr)` after `assertion(new_expr, msg)` to strengthen path conditions with proven properties. This appears to facilitate improved SMT solver constraint propagation and path pruning. The approach seems particularly effective for k-induction, where each proven assertion serves as an assumption for subsequent verification steps, enabling the SMT solver to eliminate infeasible paths earlier in the search. In particular:
- Base case: Proven assertions become assumptions for the next step
- Inductive step: Each assertion immediately strengthens the induction hypothesis
- Solver can eliminate infeasible paths much faster

Performance improvements on k-induction regression test suites:
- K-induction: 18.5% faster (2344→1909 sec*proc)
- Total regression: 12.3% faster (3823→3353 sec*proc)
- Real time: 10.7% faster (956→853 sec)

### DIFF
--- a/regression/esbmc-unix2/20_flex/test.desc
+++ b/regression/esbmc-unix2/20_flex/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --unwind 1 --ir --z3
 ^VERIFICATION FAILED$
--- a/regression/esbmc-unix2/20_flex/test.desc
+++ b/regression/esbmc-unix2/20_flex/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --unwind 1 --ir --z3
 ^VERIFICATION FAILED$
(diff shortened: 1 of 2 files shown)


lucasccordeiro:
Master (30s):

````
  correct:           17620
    correct true:    10341
    correct false:    7279
  incorrect:            34
    incorrect true:     13
    incorrect false:    21
  unknown:           15915
  Score:             27209 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16404623531
````

This PR (30s):
````
  correct:           17473
    correct true:    10212
    correct false:    7261
  incorrect:            35
    incorrect true:     13
    incorrect false:    22
  unknown:           16061
  Score:             26917 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16518833748/
````

PR #2644, opened 2025-07-28
[python] Extend chr() function to process multibyte characters

Ben-Eichhoefer:
Introduce a function to convert integer values to a 1 character utf-8 string rather than casting integers > 0x7F to 1 byte characters.
Perform a symbol lookup for Name nodes to more accurately perform casting.


src/python-frontend/function_call_expr.cpp
```diff
@@ -179,7 +179,36 @@ void function_call_expr::handle_int_to_float(nlohmann::json &arg) const
   arg["value"] = static_cast<double>(value);
 }
 
-void function_call_expr::handle_chr(nlohmann::json &arg) const
+std::string utf8_encode(unsigned int int_value)
```
lucasccordeiro, line 182: Shall we try to validate this input?

Currently, it accepts values ranging from 0 to ~4,294,967,295 (the full unsigned integer range).

I can see that we have a valid Unicode range: 0x000000 to 0x10FFFF (1,114,111) only. However, I believe that we have invalid UTF-8 sequences for inputs > 0x10FFFF.

src/python-frontend/function_call_expr.cpp
```diff
@@ -221,15 +250,46 @@ void function_call_expr::handle_chr(nlohmann::json &arg) const
...
+    if (!sym)
+    {
+      // if symbol not found revert to a variable assignment
+      arg["value"] = std::string(1, static_cast<char>(int_value));
+      typet t = type_handler_.get_typet("chr", 1);
+      exprt expr = converter_.get_expr(arg);
+      expr.type() = t;
+      return expr;
+    }
+    const auto &const_expr = to_constant_expr(sym->value);
```
lucasccordeiro: Is it correct to assume that `sym->value` is always a constant?

src/python-frontend/function_call_expr.cpp
```diff
@@ -221,15 +250,46 @@ void function_call_expr::handle_chr(nlohmann::json &arg) const
...
+    std::string binary_str = id2string(const_expr.get_value());
+    int_value = std::stoul(binary_str, nullptr, 2);
```
lucasccordeiro: `std::stoul()` can `throw std::invalid_argument` or `std::out_of_range`. Shall we handle this exception for malformed binary strings?

lucasccordeiro (changes requested):

src/python-frontend/function_call_expr.cpp
Ben-Eichhoefer: Yes that's a better idea, currently the validation is performed before calling the function. I will improve this.

Ben-Eichhoefer: I have added validation within the function, but as the parameter is an unsigned int, negative values underflow and the error message may display the wrong value. Do you think it is ok to allow the user to infer the error? Alternatively I could  update the function to accept signed int instead.

lucasccordeiro: I would suggest keeping the `unsigned int` parameter type, as it's semantically correct for Unicode code points. 

I believe the underflow behavior for negative inputs helps by making them large values that get caught by your range check. If you prefer, you could add validation for surrogate pairs (0xD800-0xDFFF), as these are invalid in UTF-8. something like:

````CPP
    // Check for surrogate pairs (invalid in UTF-8)
    if (int_value >= 0xD800 && int_value <= 0xDFFF) {
        std::ostringstream oss;
        oss << "Code point 0x" << std::hex << std::uppercase << int_value 
            << " is a surrogate pair, invalid in UTF-8";
        throw std::invalid_argument(oss.str());
    }
````

BTW, you could also consider using hex formatting in your error messages. Your current approach with internal validation is preferable to external validation, and displaying error messages for underflowed values on negative inputs should be acceptable here.

Ben-Eichhoefer: Thanks for the suggestions, hex is definitely more readable. I also hadn't realised that the surrogate pair range was reserved and after some experimentation strings containing the reserved characters crash the solvers.

Currently the behavior of the function doesn't perfectly mimic chr as python does produce an output for the reserved characters like chr(0xd800) -> '\ud800'

lucasccordeiro (approved):
Thanks, @Ben-Eichhoefer.

The coding style CI has rejected your commit. Could you please check it?

Apart from that, I don't have further comments.

PR #2650, opened 2025-07-31
Add NoneType as a type in get_type_from_constant

Ben-Eichhoefer:
Allows return5 regression test to pass


regression/python/return12/main.py
```diff
@@ -0,0 +1,5 @@
+def return_none() -> None:
```
lucasccordeiro: Can you also add a variant of this test case to fail?

lucasccordeiro (changes requested):

lucasccordeiro (approved):

lucasccordeiro:
@Ben-Eichhoefer: How do we deal with these functions?

````Python
def maybe_int() -> int | None:
    return None  # This should definitely be OK

def conditional_return(x: bool) -> int:
    if x:
        return 42
    return None 
````

**Comment by Ben-Eichhoefer** [comment]
> @Ben-Eichhoefer: How do we deal with these functions?
> 
> ```python
> def maybe_int() -> int | None:
>     return None  # This should definitely be OK
> 
> def conditional_return(x: bool) -> int:
>     if x:
>         return 42
>     return None 
> ```

That is actually what I'm looking at right now, currently this doesn't cause an issue with the typing as such.
But as None is eventually evaluated as the signed long '0', something like the below passes when it shouldn't.

```python
a:int = None

if False:
    a = 0

assert a == 0
```

lucasccordeiro:

lucasccordeiro:
> But as None is eventually evaluated as the signed long '0', something like the below passes when it shouldn't.

@Ben-Eichhoefer: can you please open an issue for this one?

PR #2651, opened 2025-08-01
Add Interactive Loop Invariant Support to ESBMC (Experimental)

LukeW1999:
# Loop Invariant Support for ESBMC (Updated after bug fixing)

We've added k-induction based loop invariant support to ESBMC. Basically, instead of unwinding loops a bunch of times (which gets expensive fast), you can now provide loop invariants that help the verifier understand what's happening without all that unwinding.

## Why we built this

Loop unwinding for big loops is slow and sometimes you hit iteration limits. With invariants, we can handle loops that might run thousands of times or even unbounded loops. Plus, after implementation of the invariant inference, the checker would be reuseble to check if it is correct and then we can use CEGAR to refine it. 

## How it works

Added a new builtin `__ESBMC_loop_invariant()` that you stick in your loops. When you run ESBMC with `--loop-invariant`, it transforms the loop using standard k-induction:

1. Check that your invariant holds at the start
2. Assume the invariant, run one iteration, make sure it still holds  
3. After the loop, use the invariant to prove whatever you're trying to prove

Here's a simple example:

```c
int main() {
    int i = 0;
    int sum = 0;
    
    __ESBMC_loop_invariant(i >= 0 && i <= 1000 && sum == i * 10);
    while (i < 1000) {
        sum += 10;
        i++;
    }
    
    assert(sum == 10000);  // This works now!
    return 0;
}
```

## What we tested

Ran over 11 basic example of invariant usage in `loop_invariants/basic_case`, we have passed majority of them. 

Ran it on 15 examples from the k-induction benchmark suite. Several that used to give up now verify successfully:
- `check_if`: GIVE UP → VERIFICATION SUCCESSFUL  
- `count_down`: GIVE UP → VERIFICATION SUCCESSFUL
- `count_up_down`: GIVE UP → VERIFICATION SUCCESSFUL
- failed on nested loops now.
Ran over `loop_invariants/nested_loop_invaraints` and it failed.

Works fine for basic while/for/do-while loops.

### Integer overflow false positives

This doesn't work right yet:
```c
unsigned int x = 0, y = 4;
__ESBMC_loop_invariant(y % 4 == 0 && x % 4 == 0);
while (x < 300) {
    x = x + y;
    y = y + 4;
    assert(x != 3);
}
```
--loop-invariants known x < 300 but y would be NONDETINT, so would result in an overflow.

```
a = NONDETINT;
b = NONDETINT; 
assume (a % 4 ==0 && b % 4 ==0)
If ! a < 300 jump to 1

 a = a + b;
 b = b + 4;
assert (x != 3)
assert (a % 4 ==0 && b % 4 ==0)
assume 0

1: 
```
We're being too aggressive with the havoc step and generating values that cause fake overflows. Need to add some bounds checking.

### Nested loops are broken

This fails:
```c
int i = 0, sum = 0, j = 0;

__ESBMC_loop_invariant(i >= 0 && i <= 10 && sum == i * 10 + j);
while (i < 10) {
    sum += 10;
    i++;
    
    // This inner loop messes things up
    __ESBMC_loop_invariant(j >= 0 && j <= 10 && sum == i * 10 + j);
    while (j < 10) {
        sum++;
        j++;
    }
}
```

The problem is we branch too early and lose track of state between the inner and outer loops. Haven't figured out a clean way to fix this yet.

### You have to write the invariants yourself

Obviously the tool can't guess what your loop is supposed to do. You need to figure out the invariant yourself, which can be tricky. If you get it wrong, verification fails.

## Implementation notes

Main changes:
- New `LOOP_INVARIANT` instruction type
- `goto_loop_invariant.cpp` does the heavy lifting
- Added the builtin function support
- Hooked it into the main symbolic execution loop

The algorithm is pretty straightforward - we find loops with invariant calls, replace them with the k-induction structure, then run normal verification on the result.

## What's next

1. Fix the overflow issue by being smarter about havoc ranges
2. Get nested loops working (probably need to rethink how we handle the state)
3. Maybe add some basic invariant inference for simple patterns
4. Better error messages when invariants are wrong

## Usage

Just add `--loop-invariant` when you run ESBMC:
```bash
esbmc --loop-invariant your_file.c
```

Existing code still works exactly the same - this is completely opt-in.


LukeW1999:
Need to pass the Codacy, but it looks fine for me

regression/loops/loop_invariants/0-loop_invariants.c
```diff
@@ -0,0 +1,38 @@
...
+ 
+     /*@ loop invariant i >= 0 && i <= 5000000 && sum == i * 10;
+     @*/
+     
+     //first check the base case
+     while (i < 5000000)//replace the condition to i <5000000
+     {   
+        //should havoc the variables: change the invovlved variables to non deterministic values
+        //use assumption to set i >= 0 && i <= 5000000 && sum == i * 10;
+        __ESBMC_loop_invariant(i >= 0 && i <= 5000000 && sum == i * 10);
```
rafaelsamenezes: I think I would prefer this outside the loop. @lucasccordeiro any opinion?

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -472,6 +472,92 @@ _Noreturn void __ESBMC_unreachable();
...
+ *   
+ * // Branch 2: 
+ *   else {
+ *     // invariant + !condition (need to add theloop exit condition)
+ *     assume (!(lb <= ub));
+ *     assume (lb - 1 <= ub && ub < size && mid == ((unsigned int)lb + (unsigned int)ub) >> 1);
+ *     // we assume the invariant is preserved after the loop, then we can continue on this branch to check the following assertions.
+ *   }
+ * }
+ *
```
rafaelsamenezes: This documentation should be closer to the goto_loop_invariant.h file

src/cpp/cpp_language.cpp
```diff
@@ -63,6 +63,7 @@ void cpp_languaget::internal_additions(std::ostream &out)
          "bool assertion, const char *description);"
       << std::endl;
   out << "bool __ESBMC_same_object(const void *, const void *);" << std::endl;
+  out << "void __ESBMC_loop_invariant(void*, bool);" << std::endl;
```
rafaelsamenezes: No need to change the old cpp frontend

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -339,6 +340,9 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
   if (cmdline.isset("compact-trace"))
     options.set_option("no-slice", true);
 
+  // Set default value for loop-invariant option
+  options.set_option("loop-invariant", false);
```
rafaelsamenezes: It is already false by default

src/goto-programs/builtin_functions.cpp
```diff
@@ -546,32 +546,67 @@ void goto_convertt::do_function_call_symbol(
...
+    if (options.get_bool_option("no-assertions") && !is_assume && !is_loop_invariant)
       return;
 
-    goto_programt::targett t =
-      dest.add_instruction(is_assume ? ASSUME : ASSERT);
-    migrate_expr(arguments.front(), t->guard);
+    // Rafael's invariant merging: combine consecutive __invariant() calls 
+    // into a single LOOP_INVARIANT instruction for efficiency
+    // not tested yet, but should be correct
+      goto_programt::targett t;
```
rafaelsamenezes: We should  test this and confirm if its working.

src/goto-programs/goto_loop_invariant.cpp
```diff
@@ -0,0 +1,205 @@
...
+  for (const auto &invariant : invariants)
+  {
+    // Create assert instruction for each invariant
+    goto_programt::targett t = dest.add_instruction(ASSERT);
+    t->guard = invariant;
+    t->location = loop_exit->location;
+    t->location.comment("loop invariant inductive step");
+  }
+  
+  // 2. Insert ASSUME(FALSE) to terminate the loop
```
rafaelsamenezes, line 287: I am not sure if this works for nested loops

rafaelsamenezes (changes requested):
I just read the code. I will still try to play with it a bit.

regression/loops/loop_invariants/0-loop_invariants.c
LukeW1999: Sure, I think that should be decide by Lucas.

src/clang-c-frontend/clang_c_language.cpp
LukeW1999: Let me just move it to the goto_loop_invariant.h file

src/cpp/cpp_language.cpp
LukeW1999: Understand, thank you.

src/goto-programs/builtin_functions.cpp
LukeW1999: I would write some extra test cases tomorrow to test it. 
This PR need a lot more test to do.

src/goto-programs/goto_loop_invariant.cpp
LukeW1999: If implemented well and cooperate with multi-properties correctly, It could work. But it should be doubtful now. I aim to pass some simple nested loops test cases next week.

regression/loops/loop_invariants/0-loop_invariants.c
lucasccordeiro: I agre with @rafaelsamenezes.

Shall we do something like:

````C
  while (i < 5000000)//replace the condition to i <5000000
  __ESBMC_loop_invariant(i >= 0 && i <= 5000000 && sum == i * 10);
     {
      .....
     }   
````

File: src/clang-c-frontend/clang_c_language.cpp (hunk 2 above)
Comment (lucasccordeiro) [reply src/clang-c-frontend/clang_c_language.cpp]: I agree with @rafaelsamenezes.

File: src/cpp/cpp_language.cpp (hunk 3 above)
Comment (lucasccordeiro) [reply src/cpp/cpp_language.cpp]: I agree with @rafaelsamenezes.

File: src/goto-programs/goto_loop_invariant.cpp [hunk 7]
```diff
@@ -0,0 +1,205 @@
...
+  const loopst &loop,
+  const std::vector<expr2tc> &invariants)
+{
+  // Find the end of loop body (before GOTO back to loop head)
+  goto_programt::targett loop_exit = loop.get_original_loop_exit();
+  goto_programt::targett insert_point = loop_exit;
+  
+  // Move back to find the last assignment in loop body
+  // Add safety check to prevent going beyond beginning
+  int safety_counter = 0;
```
lucasccordeiro: Shall we use size_t?

src/goto-programs/goto_loop_invariant.cpp
```diff
@@ -0,0 +1,205 @@
...
+  const int max_iterations = 1000; // Prevent infinite loop
```
lucasccordeiro: This is a magic number. Can we set this as an option in the command-line? We can keep a default value if the user does not change it.

src/esbmc/esbmc_parseoptions.cpp
LukeW1999: Removed

src/goto-symex/symex_main.cpp
```diff
@@ -1381,3 +1389,32 @@ void goto_symext::add_memory_leak_checks()
...
+  // this aims to use esbmc to use a single step to prove the loop invariant
+  // Basic guard check - skip if guard is false
+  if (cur_state->guard.is_false())
+    return;
+
+  // Get the loop invariant
+  const goto_programt::instructiont &instruction = *cur_state->source.pc;
+
+  log_status("Processing {} loop invariant", instruction.get_loop_invariants().size());
+  for (const auto &invariant : instruction.get_loop_invariants())
```
lucasccordeiro: Do we check whether the invariants are well-formed or valid expressions before processing?

src/cpp/cpp_language.cpp
LukeW1999: Removed

src/goto-symex/symex_main.cpp
```diff
@@ -1381,3 +1389,32 @@ void goto_symext::add_memory_leak_checks()
...
+  { 
+    expr2tc rename_invariant = invariant;
```
lucasccordeiro: I think we have a potential memory issue here. You're creating copies of expressions (expr2tc rename_invariant = invariant) without considering the size of invariants.

lucasccordeiro (changes requested):
Can you create a regression suite called "loop-invariant"? You could get our regression tests from the `k-induction` suite and add invariants.

lucasccordeiro:
@LukeW1999: I have just invited you to join our esbmc organisation: https://github.com/esbmc/esbmc/invitations

regression/loops/loop_invariants/0-loop_invariants.c
rafaelsamenezes: I would say to keep the standard (or use an annotation)

```c
   __ESBMC_loop_invariant(i >= 0 && i <= 5000000 && sum == i * 10);
  while (i < 5000000)//replace the condition to i <5000000

     {
      .....
     }   
 ```

Comment (lucasccordeiro) [reply regression/loops/loop_invariants/0-loop_invariants.c]: Thanks, @rafaelsamenezes.

I'm fine with this suggestion.

**Comment by LukeW1999** [comment]
> @LukeW1999: I have just invited you to join our esbmc organisation: https://github.com/esbmc/esbmc/invitations

Thank you, Sir. And thank you for the review.

File: src/goto-programs/builtin_functions.cpp (hunk 5 above)
Comment (LukeW1999) [reply src/goto-programs/builtin_functions.cpp]: No, it can not work properly. Despite it combined the invariant together but is not working with my file.

For this example:
```
  __ESBMC_loop_invariant(i >= 0 && i <= 5000000);
  __ESBMC_loop_invariant(sum == i * 10);
  while (i < 5000000)
```
It shows 
```        
// 19 file ../regression/loop_invariants/basic_case/8-multiple_invariant/8-multiple_invariant.c line 8 column 3 function main
        LOOP_INVARIANT i >= 0 && i <= 5000000 sum == i * 10
        // 20 file ../regression/loop_invariants/basic_case/8-multiple_invariant/8-multiple_invariant.c line 10 column 3 function main
     1: ASSERT i >= 0 && i <= 5000000 // loop invariant base case
        // 21 file ../regression/loop_invariants/basic_case/8-multiple_invariant/8-multiple_invariant.c line 10 column 3 function main
        ASSERT sum == i * 10 // loop invariant base case
        // 22 file ../regression/loop_invariants/basic_case/8-multiple_invariant/8-multiple_invariant.c line 10 column 3 function main
```

Still checked in two assertion. I would try to fix it today.

File: regression/loops/loop_invariants/0-loop_invariants.c (hunk 1 above)
Comment (LukeW1999) [reply regression/loops/loop_invariants/0-loop_invariants.c]: Have change to this format
```
   __ESBMC_loop_invariant(i >= 0 && i <= 5000000 && sum == i * 10);
  while (i < 5000000)//replace the condition to i <5000000

     {
      .....
     }   
```

File: src/clang-c-frontend/clang_c_language.cpp (hunk 2 above)
Comment (LukeW1999) [reply src/clang-c-frontend/clang_c_language.cpp]: Resolved

File: src/goto-programs/builtin_functions.cpp (hunk 5 above)
Comment (LukeW1999) [reply src/goto-programs/builtin_functions.cpp]: Have fixed this function

File: src/goto-programs/goto_loop_invariant.cpp (hunk 6 above)
Comment (LukeW1999) [reply src/goto-programs/goto_loop_invariant.cpp:287]: Not working at all. I think I would fix it in the following weeks.

File: src/goto-programs/goto_loop_invariant.cpp (hunk 7 above)
Comment (LukeW1999) [reply src/goto-programs/goto_loop_invariant.cpp]: Changed to size_t now

File: src/goto-programs/goto_loop_invariant.cpp (hunk 8 above)
Comment (LukeW1999) [reply src/goto-programs/goto_loop_invariant.cpp]: I do not know if we have to provide option for that. If a loop has no any assignment in last 1000 statement, it would be hard to read. But I would provide a option next week

**Comment by LukeW1999** [comment]
> Can you create a regression suite called "loop-invariant"? You could get our regression tests from the `k-induction` suite and add invariants.

I am not very sure about how to make a regression suite in ESBMC. I would ask Tong next week.



> Can you create a regression suite called "loop-invariant"? You could get our regression tests from the `k-induction` suite and add invariants.

I have tested 3 k-induction test cases. Result is pretty satisfying that all these three cannot verified by k-induction. But with correct invariants, they all can be verified and if provided wrong invariants, wrong invariants would be report.

Some of the review suggestion, I would try to fix next week. 
Many K-induction cases have nested loop. I do not think the nested loops need to be solved in this PR. It would need more effort to implement the whole system and I want to split to some other PRs.

**Comment by lucasccordeiro** [comment]
I agree. You can tackle nested loops on another PR!

File: src/goto-programs/goto_loop_invariant.cpp [hunk 11]
```diff
@@ -0,0 +1,309 @@
...
+  if (condition_it == goto_function.body.instructions.end())
+    return; // No loop condition found
+  
+  // Insert BEFORE the loop condition (after the base case assert)
+  goto_programt::targett insert_point = condition_it;
+  
+  goto_programt dest;
+  
+  // 1. Insert HAVOC (nondet assignments) before the loop condition
+  auto const &loop_vars = loop.get_modified_loop_vars();
```
XLiZHI, line 233: Hi @LukeW1999, I took a quick look at the PR but didn't test this code. Shouldn't we get `unmodified_vars` here?

src/goto-programs/goto_loop_invariant.cpp
```diff
@@ -0,0 +1,309 @@
...
+  const std::vector<expr2tc> &invariants)
+{
+  // Find the end of loop body (before GOTO back to loop head)
+  goto_programt::targett loop_exit = loop.get_original_loop_exit();
+  goto_programt::targett insert_point = loop_exit;
+  
+  // Move back to find the last assignment in loop body
+  // Add safety check to prevent going beyond beginning
+  size_t safety_counter = 0;
+  const size_t max_iterations = 10000; // Prevent infinite loop of searching. I did not provide a option now.
```
XLiZHI: Can you please give some examples of what this method does?

src/goto-programs/goto_loop_invariant.cpp
LukeW1999: My idea is to change the related variables. But It could be wrong. Let me get some time to test it. Or could you help me implement the PR together recently? Do you have any time to work together? I think at least some parts would be able to reuse if you want to implement a invariant inference latter.

src/goto-programs/goto_loop_invariant.cpp
LukeW1999: The aim is to find the last ASSIGN in the loop and then add the assertion (loop_invariants) after the ASSIGN. It is used to avoid checking the k+1 state before the loop is passed through. It is counting from the end to beginning. I take advantage of the idea how you implement on C++ sample.

src/goto-programs/goto_loop_invariant.cpp
XLiZHI: I think here we need to assign the `nondet` object to the var related to the guard.

Here is the "add" function for "unmodified var":
https://github.com/esbmc/esbmc/blob/c44a15bff182aff9cad55a458439f86d385a44d4/src/goto-programs/goto_loops.cpp#L161-L166

For example:

```
while (i > n)
{
 if (i = 0)
   a = b;
}
```

i is guard and "unmodified var", we should convert it into:

```
i = nondet;
if (i > n)
 if (i = 0)
   a = b;
```

a is modified var, right?

The part I want to achieve may have nothing to do with this, but I'm happy to assist this PR.

src/goto-programs/goto_loop_invariant.cpp
XLiZHI: Can we simply add it at the end of the "loop block"? What's the difference like this?

src/goto-programs/goto_loop_invariant.cpp
XLiZHI: `auto const &loop_vars = loop.get_modified_loop_vars();` to `auto const &loop_vars = loop.get_unmodified_loop_vars();`?

LukeW1999: Sorry for changing my comment. But I tried to use unmodified_loop_vars and it is wrong.

The idea of havoc is:
find related variable [] of loop invariant:
HAVOC is change the number to non-determined. 

For an example

```
__ESBMC_loop_invariant(i >= n && a == b);
while (i > n)
{
 if (i = 0)
   a = b;
i++;
}
```

We cannot only havoc the i. We also need havoc a. So we can check like this:


```
i=NONDET;
a=NONDET;
assume (i >= n && a == b)
if (i > n)
{
 if (i = 0)
   a = b;
i++;
assert (i >= n && a == b)
}
```

If change to unmodified, it only havoc i instead of [i,a].

But the implementation faces some other problem, it can only works on the while and cannot work on the for loop. Do you have any idea about that?

src/goto-programs/goto_loop_invariant.cpp
LukeW1999: I aim to avoid nested loop, but rethink it is actually useless. Let me have a try. You should be correct.

LukeW1999:
@lucasccordeiro 
Sir, here is the result for this PR. It is not affecting the SV-Comp since the option is activate only when using --loop-invariant.

  correct:           17618
    correct true:    10339
    correct false:    7279
  incorrect:            32
    incorrect true:     11
    incorrect false:    21
  unknown:           15919
  Score:             27269 (max: 55885)

XLiZHI (commented):
LGTM, just minor comments.

src/goto-programs/goto_loop_invariant.cpp
```diff
@@ -0,0 +1,295 @@
...
+ *     // 6. Terminate the loop
+ *     assume (false);
+ *   }
+ *   return -1;
+ *   
+ * // Branch 2: 
+ *   else {
+ *     // invariant + !condition (need to add theloop exit condition)
+ *     assume (!(lb <= ub));
+ *     assume (lb - 1 <= ub && ub < size && mid == ((unsigned int)lb + (unsigned int)ub) >> 1);
```
XLiZHI, line 75: I'm curious about whether the "assume" added here is necessary, because the 6th step (assume(0)) above has already concluded the branch 1.

In addition, the third step has already assumed the loop invariant. What would happen if we removed these two assumptions?

XLiZHI (commented):
LGTM, just minor comments.

lucasccordeiro:
@LukeW1999: Thanks for submitting this PR. I'll merge it for now.

Could you please address @XLiZHI and @rafaelsamenezes comments on another PR?

lucasccordeiro (approved):

src/goto-programs/goto_loop_invariant.cpp
LukeW1999: It makes sences. I would test to change the assumption instruction insertion next week.

It does look like a redundance. I just resolve the conversation here and reply it in the next PR.

PR #2652, opened 2025-08-01
[Parallel Solving] GitHub #2619 output fix

### LINKED ISSUE #2619: [parallel-solving] incomplete results (bug-finding and CE)
```c
#include <assert.h>

int main()
{
  int arr[1];
  arr[3] = 10;
  for (int i = 0; i < 10; i++)
  {
    assert(1 == 0);
  }
}
```

## multi-property

` esbmc 3.c --color  --k-induction --multi-property`

```sh
✗ FAILED: '1 == 0'

[Counterexample]


State 1 file 3.c line 9 column 5 function main thread 0
Violated property:
  file 3.c line 9 column 5 function main
  assertion 1 == 0
  1 == 0

✗ FAILED: '0'

[Counterexample]


State 1 file 3.c line 6 column 3 function main thread 0
Violated property:
  file 3.c line 6 column 3 function main
  array bounds violated: array `arr' upper bound
  0

Properties: 2 verified, ✗ 2 failed
```

## parallel-solving (1 assertion failure missed)

`esbmc 3.c --color  --k-induction --parallel-solving`

```sh
✗ FAILED: '0'

VERIFICATION FAILED

Bug found (k = 1)
```

## multi-property && parallel-solving

`esbmc 3.c --color  --k-induction --parallel-solving --multi-property`

```sh
✗ FAILED: '1 == 0'
✗ FAILED: '0'

VERIFICATION FAILED
```

Luke-Sanderson:
#2619 is fixed by the change to esbmc_parseoptions.cpp. Parallel-solving enables multi-property on the options object but is_mul only checks the command line arguments.

The rest of the changes fix a race condition introduced by removing result-only from parallel-solving by default to be in line with multi-property.

### DIFF
--- a/src/goto-programs/goto_program.h
+++ b/src/goto-programs/goto_program.h
@@ -359,45 +359,36 @@ public:
       guard = gen_true_expr();
     }
 
-    instructiont(const instructiont& other)
-    : code(other.code),
-      function(other.function),
-      location(other.location),
-      type(other.type),
-      guard(other.guard),
-      // `targets` list is intentionally not copied. The `copy_from` function
-      // is responsible for creating new instructions and then fixing up the
-      // target iterators to point within the new list.
-      labels(other.labels),
-      inductive_step_instruction(other.inductive_step_instruction),
-      inductive_assertion(other.inductive_assertion),
-      location_number(other.location_number),
-      loop_number(other.loop_number),
-      target_number(other.target_number),
-      scope_id(other.scope_id),
-      parent_scope_id(other.parent_scope_id)
+    instructiont(const instructiont &other)
+      : code(other.code),
+        function(other.function),
+        location(other.location),
+        type(other.type),
+        guard(other.guard),
+        labels(other.labels),
+        inductive_step_instruction(other.inductive_step_instruction),
+        inductive_assertion(other.inductive_assertion),
+        location_number(other.location_number),
+        loop_number(other.loop_number),
+        target_number(other.target_number),
+        scope_id(other.scope_id),
+        parent_scope_id(other.parent_scope_id)
     {
-        // The `instruction_mutex` is NOT copied. The new object's mutex
-        // is default-initialized to a fresh, unlocked state.
+      // instruction_mutex is not copied
     }
 
-    // 3. Custom Copy Assignment Operator (NEW)
-    instructiont& operator=(const instructiont& other)
+    instructiont &operator=(const instructiont &other)
     {
       if (this == &other)
         return *this;
 
-      // Lock our own mutex to make the assignment operation safe.
       std::lock_guard<std::mutex> lock(instruction_mutex);
 
-      // Copy data members. Do not touch the mutex itself.
       code = other.code;
       function = other.function;
       location = other.location;
       type = other.type;
       guard = other.guard;
-      // As with the copy constructor, `copy_from` will fix the targets.
-      // Clearing them here is the safest approach.
       targets.clear();
       labels = other.labels;
       inductive_step_instruction = other.inductive_step_instruction;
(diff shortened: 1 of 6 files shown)


src/goto-programs/goto_program.h
```diff
@@ -356,6 +359,95 @@ class goto_programt
       guard = gen_true_expr();
     }
 
+    instructiont(const instructiont &other)
```
ChenfengWei0, line 361: May I ask what is this for?

Luke-Sanderson: It's a copy constructor. The instruction type is used in dynamic lists which need a copy constructor to work. Previously the copy constructor was implicit but when I added the muted (which can't be copied) it deleted the default copy constructor.

This copy constructor should act the same as the default and implicitly creates a new mutex on copy.

lucasccordeiro:
@Luke-Sanderson: Can you please check this regression test?

````
The following tests FAILED:
	340 - regression/esbmc/github_1253 (Failed)             esbmc regression
Errors while running CTest
````

File: src/goto-programs/goto_program.h (hunk 1 above)
Comment (lucasccordeiro) [reply src/goto-programs/goto_program.h:361]: Individual instruction mutexes seem like over-engineering here...  can you measure the overhead of per-instruction mutexes?

File: src/goto-programs/goto_program.h [hunk 2]
```diff
@@ -356,6 +359,95 @@ class goto_programt
...
+    {
+      // instruction_mutex is not copied
+    }
+
+    instructiont &operator=(const instructiont &other)
+    {
+      if (this == &other)
+        return *this;
+
+      std::lock_guard<std::mutex> lock(instruction_mutex);
```
lucasccordeiro: Is this correct?

It seems that this line only locks the destination object's mutex (this->instruction_mutex) but doesn't lock the source object's mutex (other.instruction_mutex).

src/goto-programs/goto_program.h
```diff
@@ -87,6 +88,8 @@ class goto_programt
   class instructiont
   {
   public:
+    std::mutex instruction_mutex;
```
lucasccordeiro: Is this the right granularity for thread safety?

lucasccordeiro: Shall we use std::shared_mutex (reader-writer lock) if instruction objects are read more frequently than they are modified?

lucasccordeiro (changes requested):

src/goto-programs/goto_program.h
Luke-Sanderson: I might have done this wrong here but my understanding is that this being a copy assignment rather than a move assignment means 'other' is const and therefore doesn't need a lock.

The move assignment on L423 has both locks because the 'other' is not const

src/goto-programs/goto_program.h
Luke-Sanderson: This is a fair critique and I don't have a good answer. The more granular the mutex the more mutexes there will be but each mutex will be held on average for less time (more interleaving). I will create two implementations with mutexes at different levels on Monday and try to measure a difference in performance

Luke-Sanderson: It seems to race condition only occurs in clear_verified_claims so having a single lock in bmc.cpp and none in the instruction functions is sufficient. This is more effecient than a shared_mutex. I also don't see a better location for the mutex.

src/goto-programs/goto_program.h
Luke-Sanderson: I have removed all locks other than the one is bmc.cpp so a shared_lock no longer makes sense

lucasccordeiro (approved):

PR #2658, opened 2025-08-04
[C verification] check empty function body

XLiZHI:
```c
void reach_error(){}
extern unsigned char __VERIFIER_nondet_uchar(void);
int main() {
  unsigned char n = __VERIFIER_nondet_uchar();
  if (n == 0) {
    return 0;
  }
  unsigned char v = 0;
  unsigned char s = 0;
  unsigned int  i = 0;
  while (i < n) {
    v = __VERIFIER_nondet_uchar();
    s += v;
    ++i;
  }
  if (s < v) {
    reach_error();
    return 1;
  }
  if (s > 65025) {
    reach_error();
    return 1;
  }
  return 0;
}
```

In this example, since `reach_error` defines an empty body, it causes esbmc to fail to correctly convert it into an assert false.

In this PR, we should at least check whether the body is empty.


XLiZHI:
This PR:
```
  correct:           17648
    correct true:    10366
    correct false:    7282
  incorrect:            32
    incorrect true:     11
    incorrect false:    21
  unknown:           15889
  Score:             27326 (max: 55885)
https://github.com/esbmc/esbmc/actions/runs/16723328029
```

Master: 
```
  correct:           17627
    correct true:    10348
    correct false:    7279
  incorrect:            34
    incorrect true:     13
    incorrect false:    21
  unknown:           15908
  Score:             27223 (max: 55885)
https://github.com/esbmc/esbmc/actions/runs/16723308389/job/47333774646
```

Hi Prof @lucasccordeiro, I think we need this PR. At the least, we should enable it in SV-COMP. In my investigation, several benchmarks exhibited this situation.

XLiZHI:
```
void assert(_Bool a) {}

int main()
{
  assert(0);
  return 0;
}
```

How should we deal with this TC? Can we treat it is as assertion if the body is empty?

In sv-comp, it seams that we do so.

lucasccordeiro (approved):

lucasccordeiro:

PR #2659, opened 2025-08-04
[Fuzzing] Update big-int fuzzing test

DevM-uk:
Adds a number of improvements for the big-int fuzzing test:

- Ensures fuzzing inputs are always null-terminated before being passed to the big-int constructor
- Ensures that invalid fuzzing inputs are discarded properly
- Performs some operations on the fuzzed big-ints to increase coverage
- Provides an optional seed corpus for libFuzzer to potentially improve fuzzing performance


DevM-uk:
When running the big-int fuzzing test, ASAN currently produces a failure as the test does not ensure that the data passed to the big-int constructor has a null terminator. As I don't think it's a bug in the big-int implementation, I ensure that the data generated by libfuzzer always ends in a null terminator and that invalid inputs are not added to the fuzzing corpus.

I still want to look further into the big-int fuzzer in case I want to make further changes for this PR.

rafaelsamenezes:
> As I don't think it's a bug in the big-int implementation,

Hm,  it seem error prone as an API anyway. I would say that we probably should kill that constructor and only allow strings.

lucasccordeiro:
@DevM-uk: Can I ask you whether you still have plans to work on this PR?

DevM-uk:
> @DevM-uk: Can I ask you whether you still have plans to work on this PR?

Yes, I am currently looking into some segfaults I sometimes get when running the fuzzer. After that I will add some tests for the big-int operators before moving on.

> > As I don't think it's a bug in the big-int implementation,
> 
> Hm, it seem error prone as an API anyway. I would say that we probably should kill that constructor and only allow strings.

I can look at moving to a different constructor here although this one is still quite convenient for fuzzed inputs.

lucasccordeiro:
Thanks, @DevM-uk.

DevM-uk:
I believe the segfault was due to an outdated Clang when compiling with ASAN enabled. Using Clang 18 has resulted in no more segfaults.

lucasccordeiro:
@DevM-uk: what is the status of this PR?

DevM-uk:
> @DevM-uk: what is the status of this PR?

Unless anyone has any requests or concerns, I've added all I wanted to the big-int fuzzing for now. The last thing I am going to was to was run the fuzzer for a few hours (overnight or tomorrow) to hopefully expose any more segfaults or bugs in the big-int implementation. Other than that I will update the PR description now and rebase onto main ready for review.
My next plan is to look into fuzzing the generation of goto programs from C/C++ and possibly Python.

unit/big-int/bigint.fuzz.cpp
```diff
@@ -7,49 +7,64 @@
...
    - Constructors
+   - Subset of operators
  \*******************************************************************/
 
 #include <cctype>
 #include <cassert>
 #include <big-int/bigint.hh>
 #include <vector>
 #include <stddef.h>
+#include <iostream>
```
rafaelsamenezes: No need for iostream here

rafaelsamenezes (changes requested):
Just asking for some clean up. Also there are files in fuzz-corpus that should be removed.

DevM-uk:
> Also there are files in fuzz-corpus that should be removed.

I wasn't sure whether I wanted to include the fuzz-corpus directory in PR. It contains three valid inputs for the big-int implementation which acts as a seed corpus for libfuzzer, allowing it to be more efficient. That being said, as the input structure isn't too complex (the biggest problem being the null terminator), libfuzzer does work just fine without the initial seed corpus so there should be no problem removing it.

rafaelsamenezes:
> > Also there are files in fuzz-corpus that should be removed.
> 
> 
> 
> I wasn't sure whether I wanted to include the fuzz-corpus directory in PR. It contains three valid inputs for the big-int implementation which acts as a seed corpus for libfuzzer, allowing it to be more efficient. That being said, as the input structure isn't too complex (the biggest problem being the null terminator), libfuzzer does work just fine without the initial seed corpus so there should be no problem removing it.

Ow I see. It might be worth to this on a separate PR. One that we could consider a design to these corpus considering the project structure.

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @DevM-uk.

PR #2663, opened 2025-08-05
[goto-contractor] handle missing interval data in goto_contractor

### LINKED ISSUE #2662: [goto-contractor] terminate called after throwing an instance of 'std::out_of_range'
````C
#include<assert.h>

int main(){
 int x=1, y=0;

 while(y <1000 &&__VERIFIER_nondet_int())
 {
   x=x+y;
   y=y+1;
 }

 assert(x >= y);

 return 0;
}
````

````
$ esbmc goto-contractor.c --k-induction --goto-contractor
Parsing goto-contractor.c
Converting
Generating GOTO Program
GOTO program creation time: 0.517s
Interval Analysis time: 0.003s
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
Aborted (core dumped)
````

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2662.

Prevent `std::out_of_range` crash by checking if variables exist in the interval analysis map before accessing them. Add a fallback for missing interval data at specific program locations.

### DIFF
--- a/regression/goto-contractor/github_2662/main.c
+++ b/regression/goto-contractor/github_2662/main.c
@@ -1,0 +1,16 @@
+#include<assert.h>
+
+int main(){
+ int x=1, y=0;
+
+ while(y <1000 &&__VERIFIER_nondet_int())
+ {
+   x=x+y;
+   y=y+1;
+ }
+
+ assert(x >= y);
+
+ return 0;
+}
+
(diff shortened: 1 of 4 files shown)


lucasccordeiro:
@rafaelsamenezes: could you please check why we get an empty interval here?

PR #2666, opened 2025-08-07
[python] Support subtyping in isinstance() checks

brcfarias:
Store "bases" in class definitions and fix base class lookup.


lucasccordeiro (approved):

lucasccordeiro:

PR #2667, opened 2025-08-08
[C verification] add OM for nan function

### LINKED ISSUE #2368: [Clang-C-Frontend] Assertion `type.id() == t_floatbv' failed - II
```c
#include <assert.h>

static int nan(int a, int b) {
    return !(a & b);  // NAND is the negation of AND
}

int main() {
    // Test NAND operation
    assert(nan(1, 1) == 0); // 1 NAND 1 should be 0
    return 0;
}
```

Defining a *static* function called `nan` is perfectly legal even if a bit weird.

But it fails with:
```
Parsing foo.c
Converting
esbmc: /home/runner/work/esbmc/esbmc/src/util/std_types.h:650: const floatbv_typet& to_floatbv_type(const typet&): Assertion `type.id() == t_floatbv' failed.
```

As I said in https://github.com/esbmc/esbmc/pull/2358#pullrequestreview-2700785973, I think it would be far better to provide definitions for the functions in question or to at least match them more precisely.

XLiZHI:


lucasccordeiro (approved):

lucasccordeiro:

PR #2670, opened 2025-08-09
[Parallel Solving] Fix data races in irep

Luke-Sanderson:
Adds two new regression tests. uthash-1 caused a data race in bmc.cpp and uthash-2 caused a data race in irep.

Changes in bmc.cpp adds new locks using existing mutexes for verified claims sets.

Changes in irep.h add a mutex to the irep.data dt class. This protects the data during the detach when the irep is copied on write.


lucasccordeiro (approved):

lucasccordeiro:

PR #2674, opened 2025-08-11
[python] Handling hasttr()

brcfarias:
We have a mapping from objects to a list of attributes in the `python_converter`. I am handling the `hasattr(obj, 'attrname')` function by using this map to check whether an attribute belongs to an object.


lucasccordeiro (approved):

lucasccordeiro:

PR #2675, opened 2025-08-11
[Parallel Solving] Data Race in Symex 

Luke-Sanderson:
Test caused a TSAN data race warning every 1 in 2 runs. Caused by multiple threads attempting to increment output_count concurrently


lucasccordeiro:
@Luke-Sanderson: Let's see if this PR fixes the issue https://github.com/esbmc/esbmc/pull/2676.

Luke-Sanderson:
> @Luke-Sanderson: Let's see if this PR fixes the issue #2676.

Yes that PR does fix the data race issue. Closing this PR now

PR #2676, opened 2025-08-11
[goto-symex] Fix static output_count scoping in symex_target_equationt

lucasccordeiro:
This PR replace static `unsigned output_count` with an instance member variable in `symex_target_equationt` class. This fixes thread safety issues and ensures each equation instance maintains its own counter state. In particular, this PR:

- Adds `output_count` as protected member variable
- Initializes in constructor  
- Resets counter in `clear()` method
- Removes global state sharing between instances


Luke-Sanderson (approved):
LGTM

PR #2481, opened 2025-05-28
Disabled broken solvers for benchexec CI

rafaelsamenezes:
Yices and Bitwuzla are not building on benchexec CI runs. The crashes seems to be related to GMP.

Lets avoid it for now and enable again when it is working.

PR #2677, opened 2025-08-11
[smt] updated to Bitwuzla 0.8.2

lucasccordeiro:


scripts/build.sh
```diff
@@ -70,19 +70,47 @@ ubuntu_setup () {
...
     fi
 
     sudo apt-get update &&
     sudo apt-get install -y $PKGS &&
 
+    # Install GMP 6.3.0 from source (needed for bitwuzla)
+    echo "Installing GMP 6.3.0 from source..." &&
+    ORIGINAL_DIR="$PWD" &&
+    cd /tmp &&
+    # Try multiple mirrors for reliability
```
intrigus-lgtm: Isn't there an official mirror domain that will dynamically resolve to the closest mirror?

scripts/build.sh
```diff
@@ -70,19 +70,47 @@ ubuntu_setup () {
...
+    (wget https://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz || \
+     wget https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz || \
+     wget https://mirror.koddos.net/gnu/gmp/gmp-6.3.0.tar.xz) &&
+    tar -xf gmp-6.3.0.tar.xz &&
+    cd gmp-6.3.0 &&
+    ./configure --prefix=/usr/local --enable-cxx --enable-static &&
+    make -j$(nproc) &&
+    sudo make install &&
+    sudo ldconfig &&
+    echo "GMP 6.3.0 installed successfully" &&
```
intrigus-lgtm, line 92: Why are we doing this here instead of cmake?
Shouldn't everyone use gmp 6.3.0 even if they don't use build.sh?

scripts/build.sh
Luke-Sanderson: https://ftpmirror.gnu.org/ is the official mirror domain. 
https://ftpmirror.gnu.org/gmp/gmp-6.3.0.tar.xz will download from the closest mirror as you said

scripts/build.sh
rafaelsamenezes: Hm, not sure. We should allow users to use their own local GMP if they want.

scripts/build.sh
rafaelsamenezes: Good idea. I updated the mirror

scripts/build.sh
intrigus-lgtm: Bitwuzla itself specifies GMP 6.3.0:
https://github.com/bitwuzla/bitwuzla/blob/9fca27e2b0b4ad43269880da9e6a6c8bed29f962/subprojects/gmp-6.3.0.wrap#L2
I think all the weird issues https://github.com/esbmc/esbmc/pull/2483 https://github.com/esbmc/esbmc/pull/2481 are related to the fact that Bitwuzla wanted GMP 6.3.0 which it (I think) successfully installed (using the meson build tool) but then when linking we didn't properly link to the GMP 6.3.0 Bitwuzla wanted, but the old version from the system.

PR #2678, opened 2025-08-12
[Parallel Solving] Data race in goto trace

Luke-Sanderson:
Previously step type could be changed from ASSERT to SKIP by a different thread during the goto trace. This PR reuses a mutex for the same problem in bmc and renames the mutex.


Luke-Sanderson:
I invesitgated changing type into an atomic enum and using compare_and_exchange_strong instead of make_skip but because of the side effects of make_skip this was not possible and the mutex is required

lucasccordeiro (approved):

lucasccordeiro:

PR #2679, opened 2025-08-12
Ask solver for LHS values when ESBMC fails to generate a constant

rafaelsamenezes:
In some cases, we might be trying to do some pointer arithmetics:

```c
#include <stdint.h>
int array[10];

int main()
{
  __ESBMC_assume(((uint64_t)&array) >= 0x20 && ((uint64_t)&array) <= 0x60);
  uint64_t myvalue = (uint64_t)&array;
  __ESBMC_assert(myvalue + 2 < 0x50, "fails");
}
```

By default ESBMC will return 

```
State 13 file ../test.c line 7 column 3 function main thread 0
  myvalue = (unsigned long int)(&array[0])
```

Which is pretty useless to track the issue. We could try to track down some glue for the expression. But for this case we might as well just ask the solver for what the value is. Now the CE shows:

```
myvalue = 78 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 01001110)
```

I am still not sure about this design though. Specially because it only works if you have an intermediate variable.


lucasccordeiro:
@rafaelsamenezes: Could extra `smt_conv.get()` calls slow things down for large traces? You could make it optional or only run it for complex expressions (casts, arithmetic).

Your solution works with intermediate variables, which might confuse users. You can either document this clearly or extend the feature to cover more cases.

Note that symbolic forms can reveal relationships. Consider showing both symbolic and concrete values, or adding a verbose mode for extra context.

Test what happens if `smt_conv.get()` fails or times out. You could check behavior with arrays, structs, pointer chains, and different pointer sizes (32-bit vs 64-bit).

Instead of always calling the solver, you could cache results to avoid duplicates. I think I mentioned that the assert cache solution in ESBMC is buggy... You could also use heuristics to decide when concrete values are helpful.

lucasccordeiro:
@rafaelsamenezes: what should we do with this PR?

rafaelsamenezes:
> @rafaelsamenezes: what should we do with this PR?

It still requires a bit more of playing around to check when it breaks. And I still need to validate if the CE generated makes sense.

lucasccordeiro:
@XLiZHI: could you please check this PR?

PR #2680, opened 2025-08-12
[Parallel Solving] Data race in irep2

Luke-Sanderson:
crc_mutex in the irep2 class did not cover the copy constructor. If do_crc is called on object A and object B copy constructed from object A at the same time a data race occurs on the crc_val.

Adding the crc_mutex to the copy constructor avoid this.


lucasccordeiro (approved):

lucasccordeiro:

PR #2682, opened 2025-08-12
[OM] Optimize strncpy for BMC performance

### LINKED ISSUE #1630: [Multi-property] [OM] Performance of strncpy()
We have to manually set the unwind uper bound for the strncpy() call, otherwise the loop unrolling will be infinite at line:https://github.com/esbmc/esbmc/blob/9de1ff86233ed78729b7897e4db0d9d3db9be0d8/src/c2goto/library/string.c#L44
 

```
Unwinding loop 4 iteration 804   file string.c line 44 column 5 function strncpy
Unwinding loop 4 iteration 805   file string.c line 44 column 5 function strncpy
Unwinding loop 4 iteration 806   file string.c line 44 column 5 function strncpy
...
```
For example, the minimal upper bound for SIZE - len("Hello") = 5.
```
#include <string.h>
#define SIZE 10

int main() {
  const char *source = "Hello";
  char destination[SIZE];
  strncpy(destination, source, SIZE);
  return 0;
}
```
esbmc main.c --unwindset 4:5 : success
esbmc main.c --unwindset 4:4 : unwinding assertion loop

This can cause a performance differerence, for example, on rmm:
```
unwind                                          time
17 		unwind assertion 	        2m41s
18 		success 		        3m29s
20 		success 		        4m36s
26 		success 		        7m32s

```
26 is the default bound (max array length) I set, and 18 is the minimal bound I found that can output the same result, and the runtime can be improved from 7m32s to 3m29s!

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1630.

This PR replaces the null-padding while loop with `memset` to eliminate "infinite" loop unrolling during BMC analysis. The original implementation required manual unwind bounds and caused significant performance overhead.

Before: Two loops - copy loop + null-padding loop (requires unrolling)
After:  Copy loop + memset call (no unrolling needed)

Performance improvement:
- Eliminates up to (n - strlen(src)) loop iterations
- For simple cases, no longer requires manual --unwindset parameter tuning


lucasccordeiro:
Master (30s):

````
  correct:           17587
    correct true:    10307
    correct false:    7280
  incorrect:            32
    incorrect true:     11
    incorrect false:    21
  unknown:           15950
  Score:             27206 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16907144497/
````

This PR (30s):

````
  correct:           17580
    correct true:    10301
    correct false:    7279
  incorrect:            32
    incorrect true:     11
    incorrect false:    21
  unknown:           15957
  Score:             27193 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/16922627367/
````

PR #2684, opened 2025-08-13
[pointer-analysis] add alignment checks for direct pointer dereferences

### LINKED ISSUE #1175: Dereferencing from a const should be illegal

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1175.

ESBMC was missing alignment validation for direct pointer dereferences, causing misaligned accesses such as `*(int*)90` to pass verification when they should fail with alignment violations.

This PR:

- Moves alignment checks before validity checks in `build_reference_to()`
- Reuses the existing `check_alignment()` method for consistency
- Supports both constant pointers and symbolic pointers
- Uses proper type handling (signed types for `pointer_offset2tc`)
- Respects existing options such as `--no-align-check`


regression/esbmc/github_1175_1/main.c
```diff
@@ -0,0 +1,7 @@
+int main()
+{
+  const int addr = 100;
+  int *ptr = (int*)addr;  // Cast const int to pointer
```
intrigus-lgtm, line 4: I don't understand, why this should fail with:
```
Violated property:
  file /tmp/main4.c line 5 column 3 function main
  dereference failure: Incorrect alignment when accessing data object
```
100 is divisible by 4.
If we change 100 to 101, we get `VERIFICATION SUCCESSFUL`?

Is the polarity of the check wrong?

regression/esbmc/github_1175_11/test.desc
```diff
@@ -0,0 +1,4 @@
+KNOWNBUG
```
intrigus-lgtm, line 1: Why is (only!) this test failing?

regression/esbmc/github_1175_9/test.desc
```diff
@@ -0,0 +1,4 @@
+KNOWNBUG
```
intrigus-lgtm, line 1: Same for here.

src/pointer-analysis/dereference.cpp
```diff
@@ -612,6 +612,64 @@ bool dereferencet::dereference_type_compare(
...
+  }
+
+  // Handle typecast expressions with constant integer sources
+  if (
+    is_typecast2t(deref_expr) &&
+    is_constant_int2t(to_typecast2t(deref_expr).from))
+  {
+    return constant_int2tc(
+      bitsize_type2(),
+      to_constant_int2t(to_typecast2t(deref_expr).from).value * 8);
```
intrigus-lgtm, line 659: This smells a bit unsound-ish to me.
Wouldn't `create_pointer_offset_bits` return 8000 for `(int8_t) (256*4 - 24)` (256*4 - 24 = 1000) while instead it should return `-192`?

src/pointer-analysis/dereference.cpp
```diff
@@ -612,6 +612,64 @@ bool dereferencet::dereference_type_compare(
...
 }
 
+void dereferencet::check_pointer_alignment(
+  modet mode,
+  const type2tc &type,
+  const expr2tc &deref_expr,
+  const guardt &guard)
+{
+  if (options.get_bool_option("no-align-check"))
+    return;
```
intrigus-lgtm, line 622: Technically that check is more or less redundant, now that check_alignment also has that function call.

regression/esbmc-cpp/map/map_equal_range/test.desc
```diff
@@ -1,4 +1,4 @@
 CORE
 main.cpp
---unwind 5 --no-unwinding-assertions --timeout 900
+--unwind 5 --no-unwinding-assertions --timeout 900 --no-align-check
```
intrigus-lgtm, line 3: Why all those changes in the `map` sub folder?

src/pointer-analysis/dereference.cpp
lucasccordeiro: Thanks, @intrigus-lgtm. I agree with you. I have just removed it here https://github.com/esbmc/esbmc/commit/d48ca77b2a53f80909a767c165d19da596494f13.

I'll address the other comments via small PRs.

src/pointer-analysis/dereference.cpp
lucasccordeiro: Well spotted, @intrigus-lgtm.

I have just encoded the pointer offset calculation symbolically and let the `check_alignment` (https://github.com/esbmc/esbmc/blob/master/src/pointer-analysis/dereference.cpp#L2415-L2460) do the computation: https://github.com/esbmc/esbmc/commit/ff23dd763608f13b9575ca18a2cfbdf94910677e. 

This has found a regression that was incorrectly classified in our suite: https://github.com/esbmc/esbmc/commit/0daf88c840a162e97d163e2c04fd6585345037fa.

intrigus-lgtm: > This has found a regression that was incorrectly classified in our suite: [0daf88c](https://github.com/esbmc/esbmc/commit/0daf88c840a162e97d163e2c04fd6585345037fa).

I'm still not 100% sold on what we are doing here.
That is, `((b *)60000)->a = 0;` rightfully crashes in most environments such as the "normal" linux environment.
But if you consider memory-mapped IO instead, this code might be perfectly valid, isn't it?

lucasccordeiro: That's a good question.

@rafaelsamenezes has worked on this with one of our industrial partners.

@rafaelsamenezes: Do you have any thoughts here?

lucasccordeiro: I created this issue to keep the discussion going on: https://github.com/esbmc/esbmc/issues/2693

PR #2685, opened 2025-08-14
[Parallel Solving] Data Race in irep2 cont

Luke-Sanderson:
Issue fixed in https://github.com/esbmc/esbmc/pull/2680 still occurs about 1 in 20 times due to missing a lock for crc_val. Searched all the files to crc_val references and this should be all.


lucasccordeiro (approved):

lucasccordeiro:

PR #2688, opened 2025-08-15
[Witness] support yaml-based violation witness

XLiZHI:
```c
void reach_error(){}
extern unsigned char __VERIFIER_nondet_uchar(void);
int main() {
  unsigned char n = __VERIFIER_nondet_uchar();
  if (n == 0) {
    return 0;
  }
  unsigned char v = 0;
  unsigned char s = 0;
  unsigned int  i = 0;
  while (i < n) {
    v = __VERIFIER_nondet_uchar();
    s += v;
    ++i;
  }
  if (s < v) {
    reach_error();
    return 1;
  }
  if (s > 65025) {
    reach_error();
    return 1;
  }
  return 0;
}
```

violation witness:
```yml
- entry_type: "violation_sequence"
  metadata:
    format_version: "2.0"
    uuid: "018cdad1-1dd5-4f78-8d1a-3881aab1e25b"
    creation_time: "2025-08-18T17:47:57+08:00"
    producer:
      name: "ESBMC kind"
      version: "7.10.0"
    task:
      input_files:
        - "main5.c"
      input_file_hashes:
        "main5.c": "72b03ba33e19d6ceb5991bb647260e50c2802287d89f83660de1cd2bdd11303c"
      specification: "G ! call(reach_error())"
      data_model: "LP64"
      language: "C"
  content:
    - segment:
        - waypoint:
            type: "assumption"
            action: "follow"
            constraint:
              value: "n == 2"
              format: "c_expression"
            location:
              file_name: "main5.c"
              line: 4
              column: 3
              function: "main"
    - segment:
        - waypoint:
            type: "assumption"
            action: "follow"
            constraint:
              value: "v == 129"
              format: "c_expression"
            location:
              file_name: "main5.c"
              line: 12
              column: 9
              function: "main"
    - segment:
        - waypoint:
            type: "assumption"
            action: "follow"
            constraint:
              value: "s == 129"
              format: "c_expression"
            location:
              file_name: "main5.c"
              line: 13
              column: 5
              function: "main"
    - segment:
        - waypoint:
            type: "assumption"
            action: "follow"
            constraint:
              value: "v == 251"
              format: "c_expression"
            location:
              file_name: "main5.c"
              line: 12
              column: 9
              function: "main"
    - segment:
        - waypoint:
            type: "assumption"
            action: "follow"
            constraint:
              value: "s == 124"
              format: "c_expression"
            location:
              file_name: "main5.c"
              line: 13
              column: 5
              function: "main"
    - segment:
        - waypoint:
            type: "target"
            action: "follow"
            location:
              file_name: "main5.c"
              line: 17
              column: 5
              function: "main"

```

### DIFF
--- a/regression/witnesses/examples/example_yaml1.c
+++ b/regression/witnesses/examples/example_yaml1.c
@@ -1,0 +1,25 @@
+void reach_error(){}
+extern unsigned char __VERIFIER_nondet_uchar(void);
+int main() {
+  unsigned char n = __VERIFIER_nondet_uchar();
+  if (n == 0) {
+    return 0;
+  }
+  unsigned char v = 0;
+  unsigned char s = 0;
+  unsigned int  i = 0;
+  while (i < n) {
+    v = __VERIFIER_nondet_uchar();
+    s += v;
+    ++i;
+  }
+  if (s < v) {
+    reach_error();
+    return 1;
+  }
+  if (s > 65025) {
+    reach_error();
+    return 1;
+  }
+  return 0;
+}
--- a/src/goto-symex/goto_trace.cpp
+++ b/src/goto-symex/goto_trace.cpp
@@ -362,7 +362,6 @@ void violation_yaml_goto_trace(
   yml.verified_file = options.get_option("input-file");
   log_progress("Generating Violation Witness for: {}", yml.verified_file);
 
-#if 0
   for (const auto &step : goto_trace.steps)
   {
     switch (step.type)
@@ -370,24 +369,20 @@ void violation_yaml_goto_trace(
     case goto_trace_stept::ASSERT:
       if (!step.guard)
       {
-        graph.check_create_new_thread(step.thread_nr, prev_node);
-        prev_node = graph.edges.back().to_node;
-
-        nodet *violation_node = new nodet();
-        violation_node->violation = true;
-
-        edget violation_edge(prev_node, violation_node);
-        violation_edge.thread_id = std::to_string(step.thread_nr);
-        violation_edge.start_line = get_line_number(
-          graph.verified_file,
+        waypoint wp;
+        wp.type = waypoint::target;
+        wp.file = yml.verified_file;
+        wp.line = get_line_number(
+          yml.verified_file,
           std::atoi(step.pc->location.get_line().c_str()),
           options);
-
-        graph.edges.push_back(violation_edge);
+        wp.column = step.pc->location.get_column().c_str();
+        wp.function = step.pc->location.function().c_str();
+        yml.segments.push_back(wp);
 
         /* having printed a property violation, don't print more steps. */
 
-        graph.generate_graphml(options);
+        yml.generate_yaml(options);
         return;
       }
       break;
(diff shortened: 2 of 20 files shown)


XLiZHI:
This PR supports the new YAML format for violation witness:

1. During the pre-processing stage, the program will check whether the file extension of the input is correct. If the extension is empty, it will automatically complete it.

2. Based on SSA step, the violation witness has been filled. In theory, if the witness based on XML can be successfully generated, then the new one based on YAML can also be achieved.

src/goto-symex/goto_trace.cpp
```diff
@@ -353,6 +353,70 @@ void violation_graphml_goto_trace(
   }
 }
 
+void violation_yaml_goto_trace(
```
lucasccordeiro, line 356: violation_graphml_goto_trace and violation_yaml_goto_trace share a lot of logic. Perhaps you could refactor common logic into reusable helpers (e.g., trace → waypoints).

src/goto-symex/witnesses.cpp
```diff
@@ -861,18 +927,21 @@ void check_replace_invalid_assignment(std::string &assignment)
     assignment.clear();
 }
 
-std::string
-get_formated_assignment(const namespacet &ns, const goto_trace_stept &step)
+std::string get_formated_assignment(
```
lucasccordeiro, line 930: get_formated_assignment replaces = with == for YAML but keeps ; optional.

Current implementation risks producing slightly inconsistent YAML constraints (A == {...};).

Should we ensure YAML assignments never include trailing ;?

regression/witnesses/examples/example_yaml1.c
```diff
@@ -0,0 +1,25 @@
+void reach_error(){}
```
lucasccordeiro, line 1: Can you also add regression tests for the following scenarios?

- Nested structures
- Floating-point values with precision
- Branching traces (multiple paths)
- Function calls/returns

src/goto-symex/witnesses.h
```diff
@@ -70,6 +70,27 @@ class edget
   }
 };
 
+class waypoint
```
lucasccordeiro, line 73: Your waypoint enum includes placeholders (function_enter, function_return, branching), but these aren’t used in YAML generation yet. Is there a particular reason?

lucasccordeiro (changes requested):

src/goto-symex/witnesses.h
XLiZHI: I implemented it based on the schemas in sv-witness, and it is expected to be scalable. Although there are no examples using it currently, I think this implementation is appropriate.

https://gitlab.com/sosy-lab/benchmarking/sv-witnesses/-/blob/main/format/schemas/violation-witness-2.0-schema.yml?ref_type=heads

image

regression/witnesses/examples/example_yaml1.c
XLiZHI: Sure, will do.

src/goto-symex/witnesses.cpp
XLiZHI: The current implementation logic is that the "==" operator should be applied to all YAML and XML (I have checked the schemas of XML).

However, in YAML, we should not use trailing for assignment and have already **removed** it. please refer to the example above.

> Current implementation risks producing slightly inconsistent YAML constraints (A == {...};).

` (A == {...};)` This assignment comes from the test case of XML-based witness.

regression/witnesses/examples/example_yaml1.c
lucasccordeiro: Thank you.

src/goto-symex/witnesses.h
lucasccordeiro: That's fine then!

src/goto-symex/goto_trace.cpp
XLiZHI: I attempted to reuse some code, but the only common logic shared by these two functions is the loop processing of `goto_trace.steps`. The classes and variables they are used are all different. 

Could you please give me some hints?

lucasccordeiro: That's fine ;-) Indeed, I don't think we can further simplify this.

lucasccordeiro (approved):

PR #2699, opened 2025-08-19
[python] Class convertion adjustments

brcfarias:
This PR adjusts:
- Reading class attributes from functions;
- The `range` operational model to use functions instead of class methods;
- Base class constructor calls using `super().__init__()`.


lucasccordeiro (approved):

lucasccordeiro:

PR #2701, opened 2025-08-19
[python] add math.comb() support for ESBMC-Python

lucasccordeiro:
This PR implements binomial coefficient calculation with while loops and assert statements for ESBMC Python verification compatibility.


regression/python/combo4/main.py
```diff
@@ -0,0 +1,5 @@
+import math
+
+for n in range(0, 20):
+    for k in range(0, n + 1):
```
feliperodri, line 4: You don't need to specify a bound in cases like this?

src/python-frontend/models/math.py
```diff
@@ -1 +1,46 @@
...
+    
+    # Calculate C(n, k) using the multiplicative formula
+    # C(n, k) = n * (n-1) * ... * (n-k+1) / (k * (k-1) * ... * 1)
+    result: int = 1
+    i: int = 0
+    while i < k:
+        result = result * (n - i) // (i + 1)
+        i = i + 1
+    
+    return result
```
feliperodri, line 44: new line

regression/python/combo4/main.py
lucasccordeiro: We infer these bounds automatically ;-)

src/python-frontend/models/math.py
lucasccordeiro: Removed.

PR #2702, opened 2025-08-19
CDash support

rafaelsamenezes:
This PR syncs PRs and master results into [CDash](https://my.cdash.org/index.php?project=ESBMC+test). For now everything is synced into Experimental, in the future we could make that master is always on continuous while branchs go to experimental.


.github/workflows/build-unix.yml
```diff
@@ -37,4 +37,4 @@ jobs:
 
     - name: Run tests
       if: ${{ inputs.testing == true }}
-      run: cd build/ && ctest -j4 --output-on-failure --progress .
+      run: cd build/ && ctest -j1 -D Experimental --output-on-failure --progress .
```
intrigus-lgtm: Why are we running single-threaded now?

rafaelsamenezes: Currently there a few test cases that are crashing on the CI. These are tests that spawns sub-processes: k-induction-parallel and python.

My first idea is that it might be due to a data-race. By my understanding, gcov is not atomic. So I tested it again with -j1 just to double check if it would solve.

The next step is to actually try using Clang coverage instead.

PR #2713, opened 2025-08-22
[Invariants checker] Fix the loop invariant affect scope, it should work for nested loop now

LukeW1999:
This is a small fix for loop invariant functionality that enables support for nested loop invariants. 

However, I strongly recommend avoiding using loop invariants in nested loops for the following reasons:
1. It is extremely difficult to write loop invariants that are correct and sufficiently strong for nested structures.
2. Value changes in nested loops typically depend on the outer loop state, creating recursive-like dependencies. When using loop invariants, these historical dependencies are often broken, making it very difficult to write practically useful examples.

@XLiZHI - I've addressed your comment about the assumption issue. As you pointed out, the assumption does not need to be provided outside the loop scope. It is not providing but I have changed in the beginning code comment to avoid misunderstanding.


XLiZHI (approved):

regression/loop-invariants/4-1-assertion_inside_loop/main.c
```diff
@@ -0,0 +1,18 @@
+
+int main()
+{
+  unsigned int x, y;
+  x = 0U;
+  y = 4U;
+  //__ESBMC_loop_invariant(x % 4 == 0);
```
lucasccordeiro, line 7: Why do we keep this line?

regression/loop-invariants/4-2-assertion_inside_loop/main.c
```diff
@@ -0,0 +1,18 @@
+
+int main()
+{
+  unsigned int x, y;
+  x = 0U;
+  y = 4U;
+  __ESBMC_loop_invariant(x % 4 == 0);
+  //__ESBMC_loop_invariant(y % 4 == 0);
```
lucasccordeiro, line 8: Why do we keep this line?

regression/loop-invariants/nested-0-correct_sum/main.c
```diff
@@ -8,17 +8,20 @@ int main()
   int sum = 0;
   int j = 0;
 
-  __ESBMC_loop_invariant(i >= 0 && i <= 10 && sum == i * 10);
+  __ESBMC_loop_invariant(i >= 0 && i <= 10 && sum == i * 11);
   while (i < 10)
   {
-    sum += 10;
```
lucasccordeiro, line 14: why did you change this test case?

lucasccordeiro (changes requested):

regression/loop-invariants/4-1-assertion_inside_loop/main.c
LukeW1999: They are two variants from [4-assertion_inside_loop/main.c](https://github.com/esbmc/esbmc/pull/2713/files#diff-4096979e4996de36aca3d9922c8c7e0aeb77a560b63fa762d1f4c5818386e6a9). I want to show if the loop invariants are not strong enough. Even though the single invariant is correct, it would still fail the proof of invariants checking.

regression/loop-invariants/4-2-assertion_inside_loop/main.c
lucasccordeiro: I see!

regression/loop-invariants/nested-0-correct_sum/main.c
LukeW1999: 1. And I was trying to write a code it would work like (adding 11 times (10 times +1 in the second loop)). But it was implemented to add 10 then add 1 each times. I made a mistake in the code. The inside loop and outside loops were not having any connection. 

2. The formal test case is provided with wrong loop invairants. Before, I can not find that out since the nested loop is not supported at that time.

3. I removed sum from outside loop so it would be less dependency now and can use loop invariants to help proof.

LukeW1999: But the old test could also be valuable. I am sorry for doing that. I would avoid to do that.

lucasccordeiro (approved):

PR #2714, opened 2025-08-22
[python] Handle lists with variable-length strings

lucasccordeiro:
This PR fixes crashes when processing Python lists containing strings of different lengths. It improves type compatibility checking to treat char arrays as compatible string types.


src/python-frontend/type_handler.cpp
```diff
@@ -106,6 +106,46 @@ std::string type_handler::get_var_type(const std::string &var_name) const
...
+    return true;
+
+  // Both are character arrays (strings) - consider them compatible
+  if (t1.is_array() && t2.is_array())
+  {
+    const array_typet &arr1 = to_array_type(t1);
+    const array_typet &arr2 = to_array_type(t2);
+
+    // If both have char subtype, they're both strings
+    if (arr1.subtype() == char_type() && arr2.subtype() == char_type())
```
brcfarias: Should we compare only the subtypes instead of checking for char_type(), so it works for lists of other types?

src/python-frontend/type_handler.cpp
```diff
@@ -106,6 +106,46 @@ std::string type_handler::get_var_type(const std::string &var_name) const
...
+      return true;
+  }
+
+  return false;
+}
+
+/// Get a normalized/canonical type for list element type inference
+/// This ensures all strings use the same representative type regardless of length
+typet type_handler::get_canonical_type(const typet &t) const
```
brcfarias: Could rename to get_canonical_string_type?

brcfarias (approved):

src/python-frontend/type_handler.cpp
lucasccordeiro: I've implemented this as suggested.

src/python-frontend/type_handler.cpp
lucasccordeiro: I've made this as suggested.

PR #2719, opened 2025-08-25
[Python verification] support exception handling

XLiZHI:
Add three statement: RAISE, TRY and EXCEPTHANDLER:

convert
```python
try:
  raise ValueError("message")
except ValueError as e:
  print(e)
```
into:

```c
        CATCH tag-ValueError->1

        DECL ValueError return_value$_ValueError$1;

        FUNCTION_CALL:  ValueError(&return_value$_ValueError$1, &"message")

        THROW tag-ValueError, tag-BaseException: return_value$_ValueError$1

        DEAD __ESBMC_main::$tmp::return_value$_ValueError$1

        CATCH 

        GOTO 2

     1: DECL ValueError e;

        ASSIGN e=NONDET( ValueError { signed char * message; });

        DEAD py:main5.py@e
```

~~Now we add the annotation to the AST of the model.~~


lucasccordeiro (approved):
That's really nice 😊 

@brcfarias might provide further feedback.

src/python-frontend/python_annotation.h
```diff
@@ -124,6 +124,17 @@ class python_annotation
...
+  {
+    // Add type annotation to all class methods
+    for (Json &element : ast_["body"])
+    {
+      // Process classes and their methods
+      if (element["_type"] == "ClassDef")
+        annotate_class(element);
+    }
+  }
+
```
brcfarias: Thank you for the PR! Do we really need to annotate here since all models (including the new one) are already annotated? I’d prefer to keep all models pre-annotated to reduce runtime.

brcfarias (changes requested):

src/python-frontend/python_annotation.h
brcfarias: It should work this way:
`self.message: str = message`

XLiZHI: Many thanks! @brcfarias

brcfarias (approved):

PR #2721, opened 2025-08-27
[python] ternary operator support

lucasccordeiro:
This PR adds support for Python ternary (conditional) expressions in our Python frontend. The implementation handles the value_if_true if condition else value_if_false syntax with proper type resolution, casting, and verification support.

### DIFF
--- a/src/python-frontend/python_converter.cpp
+++ b/src/python-frontend/python_converter.cpp
@@ -2872,6 +2872,56 @@ void python_converter::get_compound_assign(
   target_block.copy_to_operands(code_assign);
 }
 
+typet resolve_ternary_type(
+  const typet &then_type,
+  const typet &else_type,
+  const typet &default_type)
+{
+  if (then_type == else_type)
+    return then_type;
+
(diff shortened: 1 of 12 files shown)


src/python-frontend/python_converter.cpp
```diff
@@ -2872,6 +2872,56 @@ void python_converter::get_compound_assign(
...
+  if (then_type.is_unsignedbv() && else_type.is_unsignedbv())
+    return then_type;
+
+  // Both arrays (strings)
+  if (then_type.is_array() && else_type.is_array())
+    return then_type;
+
+  // Both booleans
+  if (then_type.is_bool() && else_type.is_bool())
+    return then_type;
```
brcfarias: Is this block already covered at line 2880?

brcfarias (approved):

src/python-frontend/python_converter.cpp
lucasccordeiro: Indeed. It would make sense to keep these checks if we wanted to return the largest bit-vector, which is not the case here.

PR #2722, opened 2025-08-27
Slice unused array updates

rafaelsamenezes:
Solvers such as Z3 can struggle when we have a huge amount of needless array updates (see https://github.com/Z3Prover/z3/issues/7134#issuecomment-1957258357).  For example, this program:

```c
#define N 10000
int main()
{
  int myarray[N];

  for (int i = 0; i < N; i++)
    myarray[i] = i;

  __ESBMC_assert(myarray[5] == 5, "should pass");
  
}
```

Even though we only care about the index 5, we generate a huge update chain in the SSA:

```
...

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#10000 == (myarray?1!0&0#9999 WITH [9998:=9998])

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#10001 == (myarray?1!0&0#10000 WITH [9999:=9999])

Thread 0 file test.c line 9 column 3 function main
ASSERT
execution_statet::\guard_exec?0!0 => myarray?1!0&0#10001[5] == 5
should pass
```

Even though Z3 updated a fix for this, it is still not on pair with Boolector. We can, however, do this preprocessing locally. By updating the slicer with two new rules:

1. Add a new dependency track for arrays and indexes.
2. Only keep WITH operations that change the tracked index. Otherwise, replace it with the updated array.

We still keep tons of identities now, which most solvers can deal very easily. For example, for the previous program. We now have:

```
...

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#6 == myarray?1!0&0#5

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#7 == (myarray?1!0&0#6 WITH [5:=5])

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#8 == myarray?1!0&0#7

...

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#10000 == myarray?1!0&0#9999

Thread 0 file test.c line 7 column 5 function main
ASSIGNMENT ()
myarray?1!0&0#10001 == myarray?1!0&0#10000

Thread 0 file test.c line 9 column 3 function main
ASSERT
execution_statet::\guard_exec?0!0 => myarray?1!0&0#10001[5] == 5
should pass
```

Before this patch, Z3 would take 0.38s for this program.
With this patch, Z3 now takes 0.08s.

### DIFF
--- a/src/goto-symex/slice.cpp
+++ b/src/goto-symex/slice.cpp
@@ -103,19 +103,20 @@ void symex_slicet::run_on_assignment(
           return;
       }
     }
+
+    const symbol2t &lhs = to_symbol2t(SSA_step.lhs);
+    auto it = indexes.find(lhs.get_symbol_name());
     // WITH(symbol[constant_index] := constant_value)
     if (
       is_with2t(SSA_step.rhs) &&
       is_symbol2t(to_with2t(SSA_step.rhs).source_value) &&
       is_constant_int2t(to_with2t(SSA_step.rhs).update_field))
     {
-      const symbol2t &lhs = to_symbol2t(SSA_step.lhs);
       const with2t &with = to_with2t(SSA_step.rhs);
       const symbol2t &rhs_symbol = to_symbol2t(with.source_value);
       const constant_int2t &rhs_index = to_constant_int2t(with.update_field);
 
       // Is lhs in the watch list?
-      auto it = indexes.find(lhs.get_symbol_name());
       if (it != indexes.end())
       {
         // Found an array in the dependency list! Its guard should be added to the dependency list
@@ -144,6 +145,14 @@ void symex_slicet::run_on_assignment(
         }
         return;
       }
+    }
+
+    // We might be trying to initialize an array in a weird way
+    if (it != indexes.end())
+    {
+      get_symbols<true>(SSA_step.guard);
+      get_symbols<true>(SSA_step.rhs);
+      return;
     }
 
     // we don't really need it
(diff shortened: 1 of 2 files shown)


lucasccordeiro (approved):

This is a great simplification for our VCCs!

Once you run this PR over the SV-COMP benchmarks, we can merge it ;-)

lucasccordeiro:
This PR (300s):

````
  correct:           20670
    correct true:    12770
    correct false:    7900
  incorrect:            49
    incorrect true:     17
    incorrect false:    32
  unknown:           12848
  Score:             32384 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17264879234/job/48995786324
````

Master (300s):
````
  correct:           20664
    correct true:    12768
    correct false:    7896
  incorrect:            48
    incorrect true:     17
    incorrect false:    31
  unknown:           12855
  Score:             32392 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17009501745/
````

rafaelsamenezes:
It seems that the difference came from some drivers. The extra incorrect one is shown as a crash in the the previous run but I always replicate it locally. So it is pretty much  the same result with boolector.

lucasccordeiro:

Issue #2172, opened 2024-12-01
[C++] references and (pre|post)(inc|dec)rement in structs


intrigus-lgtm:
References in structs and (pre|post)(inc|dec)rement don't play together nicely.
Probably, because we don't desugar those operators, that is, `this->foo++;`
is not turned into `this->foo = this->foo + 1;` and then the adjuster doesn't see any assignments and therefore doesn't handle the references.
(The adjuster does not for those operators and only adjust the operands. But I guess since the operand is only `this->foo` and not "assigned" we don't turn this into `*this->foo` (assuming `foo` is a reference)

TCs:
https://github.com/esbmc/esbmc/commit/11088b183557b81e7e454ddebccddfa780f8cd82

PR #2728, opened 2025-08-28
Improved (pre/post) and (inc/dec) operations handling

mihaistate:
References in structs and (pre|post)(inc|dec)rement operations were not working well.
This is happening, because the operators are not desugared, that is, this->foo++;
is not turned into this->foo = this->foo + 1; and then the adjuster was not seeing any assignments and therefore it was not handling the references.
This PR addresses these issues by creating an UnaryOperatorClass, in which pre-fixed operations are solved, and updating the CXXMemberCallExprClass to handle the post increment/decrement operations.
Addressed issue: https://github.com/esbmc/esbmc/issues/2172
Added 3 tests that pass


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -592,8 +593,56 @@ bool clang_cpp_convertert::get_expr(const clang::Stmt &stmt, exprt &new_expr)
...
+    // Check if we have reference bindings for this method
+    auto it = method_reference_bindings.find(method);
+    if (it != method_reference_bindings.end())
+    {
+      // This method uses reference members
+      exprt &ref_target = it->second;
 
+      // For increment/decrement methods, create the operation directly
+      std::string method_name = method->getNameAsString();
+      if (method_name == "increment")
```
XLiZHI, line 608: This method is very interesting. Are you modifying a member call based on function name: "increment"?

what about `void increment() { --(*ptr); }`?

regression/esbmc/structs_desugaring/main.cpp
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
```
XLiZHI, line 1: Works well in master.

regression/esbmc/structs_fields_check/main.cpp
```diff
@@ -0,0 +1,12 @@
+#include <assert.h>
```
XLiZHI, line 1: Works well in master.

regression/esbmc/structs_pre_post_ops/main.cpp
```diff
@@ -0,0 +1,31 @@
+#include <assert.h>
```
XLiZHI, line 1: works well in master

src/clang-cpp-frontend/clang_cpp_convert.cpp
mihaistate: root@ASUSVivobookS:~/ESBMC_project/esbmc/regression/esbmc/structs_pre_post_ops# esbmc main.cpp 
ESBMC version 7.9.0 64-bit x86_64 linux
Parsing main.cpp
Converting
Generating GOTO Program
GOTO program creation time: 0.200s
Symex completed in: 0.005s (27 assignments)
Slicing time: 0.000s (removed 26 assignments)
Generated 47 VCC(s), 0 remaining after simplification (1 assignments)
BMC program time: 0.005s

VERIFICATION SUCCESSFUL for 

#include <assert.h>

struct Counter {
    int *ptr;
    
    Counter(int &r) : ptr(&r) {}
    void increment() { (*ptr)++; }
    void decrement() { (*ptr)--; }
    void decrement2() { (*ptr -= 1);}
    void decrement3() { --(*ptr);}
    void increment2() { --(*ptr);}
};

int main() {
    int x = 5;
    Counter c(x);
    
    // Test single increment
    c.increment();
    assert(x == 6);  // x should be 6 after increment
    
    // Test decrement
    c.decrement();
    assert(x == 5);  // x should be 5 after decrement
    
    c.decrement2();
    assert(x == 4);

    c.decrement3();
    assert(x == 3);
   
    c.increment2();
    assert(x == 2);

    return 0;
}
I am handling a function call based on an internal method, it has nothing to do with the external function names of the tests.

regression/esbmc/structs_desugaring/main.cpp
mihaistate: When I created the future branch last month from master, it was not working well, the test was failing, so that's why I needed to develop special cases for operations desugaring. It was not working until I developed the UnaryOperator case, I think I should have synchronized my branch with master at some point, maybe by opening a PR sooner. I did git fetch origin at some point, about a week ago, maybe the master update that handles the desugaring issue is quite recent

regression/esbmc/structs_fields_check/main.cpp
mihaistate: It was the same situation as for the structs_desugaring test in the above conversation

regression/esbmc/structs_pre_post_ops/main.cpp
mihaistate: Same as in the above tests

PR #2729, opened 2025-08-29
[C ++ verification] add reference adjustments to pre/post - inc/dec 

### LINKED ISSUE #2172: [C++] references and (pre|post)(inc|dec)rement in structs

XLiZHI:
Fix #2172
```cpp
#include <assert.h>

int main()
{
  int other = 0;
  struct nested
  {
    int &y;
  };
  nested n{other};
  assert(++(n.y) == 1);
}
```

goto before:
```c
        // 19 file main6.cpp line 11 column 10 function main
        ASSIGN n.y=n.y + 1;
        // 20 file main6.cpp line 11 column 3 function main
        ASSERT n.y == 1 // assertion ++(n.y) == 1
```

This PR:
```c
        // 19 file main6.cpp line 11 column 10 function main
        ASSIGN *n.y=*n.y + 1;
        // 20 file main6.cpp line 11 column 3 function main
        ASSERT *n.y == 1 // assertion ++(n.y) == 1
```

### DIFF
--- a/regression/esbmc-cpp11/reference/Reference7-preincrement-struct-fail/test.desc
+++ b/regression/esbmc-cpp11/reference/Reference7-preincrement-struct-fail/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --std c++11 
 
--- a/regression/esbmc-cpp11/reference/Reference7-preincrement-struct/test.desc
+++ b/regression/esbmc-cpp11/reference/Reference7-preincrement-struct/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --std c++11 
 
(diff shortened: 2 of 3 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2738, opened 2025-09-02
[python] Fix forward reference handling in function calls

lucasccordeiro:
Previously, function calls to forward-declared functions returned empty expressions, causing those calls to be ignored during verification. This fix distinguishes between forward references (functions defined in the current Python source) and external functions by searching the AST.

Forward references now generate proper function call expressions that can be resolved in later compilation phases, ensuring all user-defined function calls are properly verified.

Fixes the issue where assertions in forward-referenced functions were not being checked, leading to false verification success.

### DIFF
--- a/src/python-frontend/function_call_expr.cpp
+++ b/src/python-frontend/function_call_expr.cpp
@@ -1258,41 +1258,8 @@ exprt function_call_expr::get()
         // by searching the AST directly
         bool is_forward_reference = false;
 
-        // Lambda to recursively search for FunctionDef nodes
-        std::function<bool(const nlohmann::json &)> search_ast =
-          [&](const nlohmann::json &node) -> bool {
-          if (!node.is_object())
-            return false;
-
-          // Check if this is a function definition with matching name
-          if (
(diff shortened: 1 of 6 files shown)


src/python-frontend/function_call_expr.cpp
```diff
@@ -1254,8 +1254,88 @@ exprt function_call_expr::get()
...
       if (!func_symbol)
       {
-        log_warning("Undefined function: {}", func_name.c_str());
-        return exprt();
+        // Check if this function is defined anywhere in the current Python source
+        // by searching the AST directly
+        bool is_forward_reference = false;
+
+        // Lambda to recursively search for FunctionDef nodes
+        std::function<bool(const nlohmann::json &)> search_ast =
```
brcfarias: I think this lambda could be a function in json_utils.h

brcfarias (approved):

src/python-frontend/function_call_expr.cpp
lucasccordeiro: I've implemented this as suggested.

Issue #2167, opened 2024-11-29
Inconsistent behavior in `dump_smt` across solvers


intrigus-lgtm:
The behavior in `::dump_smt` is not consistent across solvers:
 
Z3 dumps the smt formula to the file defined in `output`. If there is no file defined, it just swallows everything.
https://github.com/esbmc/esbmc/blob/9a2bd1847c5a8c68f255981abb5471bec123a1f4/src/solvers/z3/z3_conv.cpp#L1306-L1317

Note that this wasn't always the case; previously it would output to stdout when no `output` was defined:

https://github.com/esbmc/esbmc/pull/1342/commits/a2ff1a8dcc0ab1ec98d047e91655caee1565a169#diff-0489271635db802271e51ff12caa04e85c9fe16e8a38361d69256b0afbc802dfL1305-L1321

Bitwuzla always prints to stdout

https://github.com/esbmc/esbmc/blob/9a2bd1847c5a8c68f255981abb5471bec123a1f4/src/solvers/bitwuzla/bitwuzla_conv.cpp#L828-L832

Smtlib observes `output`

https://github.com/esbmc/esbmc/blob/9a2bd1847c5a8c68f255981abb5471bec123a1f4/src/solvers/smtlib/smtlib_conv.cpp#L163-L175

while Boolector again just prints to stdout:

https://github.com/esbmc/esbmc/blob/9a2bd1847c5a8c68f255981abb5471bec123a1f4/src/solvers/boolector/boolector_conv.cpp#L790-L793

This is annoying when you're wondering why `--z3 --smt-formula-only` doesn't print anything on stdout.
The `output` parameter is also overused:
it is used in smt-formula printing, vcc printing, goto2c printing and maybe somewhere else.

lucasccordeiro:
> The behavior in ::dump_smt is not consistent across solvers:

@intrigus-lgtm: A PR would be welcome to maintain consistency with our SMT solvers.

mikhailramalho:
How about we return a stringstream from dump_smt and handle the output in
the outer layer of the smt solver?

That way each solver only have to return the text and it would be easier to
unify the behaviour: if we want to output it we can just call log_whatever
or write it to a file

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/2524.

PR #2743, opened 2025-09-05
[smt] Standardize dump_smt() interface across solver backends

### LINKED ISSUE #929: [SMT backend] ESBMC is not printing the SMT formula with Z3
For some reason, ESBMC is not printing the SMT formula with Z3 when we use the options: `--smt-formula-only` and `--smt-formula-too`.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/929.

All `dump_smt()` methods now consistently return `std::string` with SMT formula content instead of mixed printing/logging behavior. Centralize output handling in `bmc.cpp` with safe dynamic allocation for file reads.

Fixes inconsistent --smt-formula-only/--smt-formula-too behavior.


src/esbmc/bmc.cpp
```diff
@@ -233,7 +233,14 @@ smt_convt::resultt bmct::run_decision_procedure(
     options.get_bool_option("smt-formula-too") ||
     options.get_bool_option("smt-formula-only"))
   {
-    smt_conv.dump_smt();
+    std::string smt_formula = smt_conv.dump_smt();
+
+    // Print the SMT formula to stdout
+    if (!smt_formula.empty())
+    {
+      fprintf(stdout, "%s\n", smt_formula.c_str());
```
rafaelsamenezes: You should check for the `--output` option for the destination: https://github.com/esbmc/esbmc/blob/master/src/esbmc/options.cpp#L320

rafaelsamenezes (changes requested):

src/esbmc/bmc.cpp
lucasccordeiro: I've implemented this as suggested.

lucasccordeiro:
This PR also properly addresses @intrigus-lgtm's issue: https://github.com/esbmc/esbmc/issues/2167

PR #2744, opened 2025-09-05
[Concurrency] Don't terminate thread in pthread_exit

### LINKED ISSUE #2737: [C] ESBMC missing an assertion after possible call to pthread_exit()
In the following program ESBMC does not mark the assertion as failing (on default options): 
```c
#include<assert.h>
int main() {
        int exit = __VERIFIER_nondet_int();
        //__VERIFIER_assume(exit == 0);
        if(exit) pthread_exit(0);
        assert(0);
}
``` 
However, for `exit = 0` it should be reachable (regardless of the semantics of `pthread_exit`).
Indeed, when assuming this, ESBMC correctly marks the assertion as failing.

ESBMC built from  10c70eb9b602ae612f770c5a8ec5d299aca31413

Anthonysdu:
This PR fixes #2737 

We only mark a thread has ended at the end of the thread. In `pthread_exit() ` we can just use `assume(0)`  at the end to cut further execution in this thread.

This PR:
https://github.com/esbmc/esbmc/actions/runs/17470662298/job/49619645540
```
Statistics:           3179 Files
  correct:            1783
    correct true:     1327
    correct false:     456
  incorrect:            35
    incorrect true:     17
    incorrect false:    18
  unknown:            1361
  Score:              2278 (max: 5725)
```

Master:
https://github.com/esbmc/esbmc/actions/runs/17466249770/job/49604469493
```
Statistics:           3179 Files
  correct:            1783
    correct true:     1327
    correct false:     456
  incorrect:            35
    incorrect true:     17
    incorrect false:    18
  unknown:            1361
  Score:              2278 (max: 5725)
```

### DIFF
--- a/regression/esbmc-cpp/unordered_map/basic_insert_fail/test.desc
+++ b/regression/esbmc-cpp/unordered_map/basic_insert_fail/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.cpp
---incremental-bmc
+--incremental-bmc --z3
 ^VERIFICATION FAILED$
--- a/regression/esbmc-unix/03_use_if/test.desc
+++ b/regression/esbmc-unix/03_use_if/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
---no-por --context-bound 4 -Wno-error=implicit-function-declaration
+--no-por --context-bound 6 -Wno-error=implicit-function-declaration
 ^VERIFICATION FAILED$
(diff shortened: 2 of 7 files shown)


lucasccordeiro (approved):

Anthonysdu:
@lucasccordeiro  Don't merge it yet, there was one suspectus timeout testcase on a simple c++ program without pthread_exit() call, I'll see if is still exits after the regression finshes.

PR #2751, opened 2025-09-08
[Python verification] raise exception to match Python semantics

XLiZHI:
This PR will replace the original function calls with the use of "raise" for exceptions:

```python
user_input = input()
number1 = float(user_input)
```

ESBMC will raise a `ValueError` for it.
```c
Parsing main6.py
Converting
Generating GOTO Program
GOTO program creation time: 0.436s
GOTO program processing time: 0.007s
ERROR: Exception thrown of type tag-ValueError at file main6.py line 2
Symex completed in: 0.002s (13 assignments)
Generated 3 VCC(s), 1 remaining after simplification (1 assignments)
Building error trace

[Counterexample]


State 1 file main6.py line 2 column 0 thread 0
Violated property:
  file main6.py line 2 column 0
  Throwing an exception of type tag-ValueError but there is not catch for it.


VERIFICATION FAILED
```


src/python-frontend/function_call_expr.cpp
```diff
@@ -839,12 +837,11 @@ exprt function_call_expr::build_constant_from_arg() const
...
       if (expr.type().is_array()) // strings are arrays
       {
-        // TODO: raise an exception to match Python semantics
         std::string var_name = arg["id"].get<std::string>();
-        log_warning(
-          "ValueError: int() conversion may fail - variable '{}' may contain "
-          "non-integer string",
-          var_name);
+        std::string m = "int() conversion may fail - variable" + var_name +
+                        "may contain non-float string";
```
lucasccordeiro: Should it be `non-integer string` instead of `non-float string`?

lucasccordeiro (changes requested):

brcfarias:
I think "tag-" could be removed from " Throwing an exception of type tag-ValueError but there is not catch for it.
"

brcfarias (approved):

lucasccordeiro (approved):

lucasccordeiro:

PR #2759, opened 2025-09-10
[C verification] fix assume constant propagation

XLiZHI:
In the IEEE-754 floating-point number semantics, there can be situations where the numerical comparison is equal, but the internal bit patterns are different: +0.0 and -0.0.


lucasccordeiro (approved):

lucasccordeiro:

PR #2761, opened 2025-09-10
[Python] Fix Python single string concatenation operations

LukeW1999:
## Problem
Python string comparisons are failing after concatenation operations. For example:
```python
str = "fine"
str = str + "!"
assert str == "fine!"  # This fails
```

**Root cause**: Does not follow the python implementation. In python, "a" should be string not a char.

## Impact
- ✅ Fixes failing test cases (`concat9`, `concat10`, `concat11`), this PR added. two concat ops equality checks and concat 11 is strlen for "a" type.
- ✅ No side effects detected (passed `scripts/check_python_tests.sh`)
- ⚠️ I could not test completely. But I do used this scripts

## Note
Add no code but remove the single char converter.


lucasccordeiro:
@LukeW1999: Does your PR handle this case?

````python
s: str = "a"
assert len(s) == 1
````

**Comment by lucasccordeiro** [comment]
We have these two places, where we handle characters with size one:

type_handler.cpp:
````C++
  if (
    ast_type == "str" || ast_type == "chr" || ast_type == "hex" ||
    ast_type == "oct" || ast_type == "ord" || ast_type == "abs")
  {
    if (type_size == 1)
    {
      typet type = char_type();      // 8-bit char
      type. Set("#cpp_type", "char"); // For C backend compatibility
      return type;
    }
    return build_array(char_type(), type_size); // Array of characters
  }

````

python_converter.cpp:
````C++
  // Handle single-character string as char literal
  if (str_val.size() == 1 && !is_bytes_literal(element))
  {
    typet t = type_handler_.get_typet("str", str_val.size());
    return from_integer(static_cast<unsigned char>(str_val[0]), t);
  }
````

We should treat these chars as a string, having the character+null terminator.

**Comment by LukeW1999** [comment]
@lucasccordeiro This PR is quite bad, only works for str += "a", assert (str == "str+a"). I would change a solution. I suppose to include the problems like:
```python
str = "a"
str = str + "a"
assert str == "aa"
str += "a"
assert str == "aaa"

```

```python
str = "a"
assert(len(str) == 1)
```

LukeW1999:
@lucasccordeiro  Hello, sir. I am really sorry. Just realize there is because of before someone made some mistakes. I realize it could be fixed with removing code instead adding extra code.

LukeW1999:
This PR brokes the ability of python string casting. I would repair it ASAP.

LukeW1999:
This PR is closed by multiple commits by Lucas and superseded by #2768.

The implementation in this PR was overly simplified and caused multiple issues:
- Failed to handle edge cases properly
- Did not pass regression tests
- Created new verification failures

The new PR #2768 provides a complete rewrite for this task.

PR #2767, opened 2025-09-11
[python] Support heterogeneous lists and append operations

brcfarias:
This PR extends the Python frontend to support heterogeneous (multi-type) lists.
It also introduces the infrastructure for the **append** function, which serves as the basis for other list operations.


sulaychaudhry:
Hi @brcfarias, I've been using this branch while debugging elsewhere in the Python code as it fixes some type symbol issues with structs - thank you :)
I am still new to the codebase and do not fully understand the new python_converter code for structs, but I've noted that Python still forgets to create type symbols for structs if they're not used as pointers anywhere.

image image image

However if I choose to use it as a pointer, it will create the type symbol just fine:

image image image

In particular, I've found you can't just define the pointer, you have to assign to it as well or the type symbol isn't created.
Do you think this is something that could be relatively easy to fix? Thank you again for all of these fixes to the python converter.

sulaychaudhry:
After a closer look, I can see that the issue should be contained entirely within generate_symbol_deps, so will be a separate issue (struct is not being included in list of dependency IDs generated). Will look at this for a different PR. Sorry to bother Bruno!

brcfarias:
Hi @sulaychaudhry,
The loading of dependencies in cprover_library has been improved in this PR, but handling symbols from external (included) dependencies is still pending. This is something we could explore further in this part of the code, as you’ve already identified.

sulaychaudhry:
Got you @brcfarias - yep am working on a separate PR now for this!

sulaychaudhry:
Hi @brcfarias - I've made the PR at #2782
It should fix all type symbol issues on the Python frontend; if you pull it into this PR you shouldn't need your changes to pulling in symbols I believe. I have placed skipped symbols into an "ignored" context handled at the end of the converter to avoid double dependency handling & in the case that a dependency appears (and is cleared) before the relevant symbol shows up. Let me know if you see any issues :)

brcfarias:
Hi @sulaychaudhry 
I applied your modifications locally on this branch, and everything is still working fine. Thanks for that!

sulaychaudhry:
No problem @brcfarias!

PR #2769, opened 2025-09-12
[python] Fix a string conversion when char is used with ord/chr functions

LukeW1999:
## Problem
String variables declared with `str` type annotation were being automatically converted to character arrays even when used with `ord()`/`chr()` functions, causing type compatibility segment fault.

## Solution
- Add pre-scan phase to identify variables used as arguments to `ord()`/`chr()` functions with single character
- Skip automatic string array conversion for these variables to maintain single character representation

## Changes
- Added `prescan_ord_chr_usage()` method to scan AST for ord/chr function calls since the adding ord_type is after the comparison. I think there could be some better solution with changing the ord and chr operations assignment. 
- The current scan AST to check the format char + ord/chr could be not efficient.
- Added `ord_chr_used_variables` set to track single char used with ord/chr
- Modified string literal assignment logic in `get_var_assign()` to skip array conversion when variable is used with ord/chr

## Known Issues
This previous fix #2768  introduces a regression break in ternary_string case. The `ternary_string` fails in this minimal format:

```python
whitespace: str = " "
space_check: str = "space" if whitespace == " " else "no_space"
assert space_check == "space"
```

**Root cause**: The single-character representation compai propagates to `space_check` during modifying on single character comparsion, causing:
- Missing `\0` terminator leads to infinite loops in string operations (guessing)
- `len(space_check)` function would also fail

This PR did not fix this problem yet, should use some way to put the null character back to the end of the assgined array or refine the modifying on single character comparsion. 

## Testing
- ✅ Fixes: `regression/python/ord-var/` 
- ❌ Breaks: `ternary_string` test case

### DIFF
--- a/regression/python/string23/main.py
+++ b/regression/python/string23/main.py
@@ -1,0 +1,5 @@
+def is_foo(a: str) -> bool:
+    return a == "foo"
+
+e = is_foo("foo")
+assert e
--- a/regression/python/string23/test.desc
+++ b/regression/python/string23/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/string23_fail/main.py
+++ b/regression/python/string23_fail/main.py
@@ -1,0 +1,5 @@
+def is_foo(a: str) -> bool:
+    return a != "foo"
+
+e = is_foo("foo")
+assert e
--- a/regression/python/string23_fail/test.desc
+++ b/regression/python/string23_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION FAILED$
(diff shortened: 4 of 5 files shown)


LukeW1999:
### Update

For this case
```python
whitespace: str = " "
space_check: str = "space" if whitespace == " " else "no_space"
assert space_check == "space"
```
in goto
```goto
        ASSIGN whitespace={ 32, 0 };
        // 773 file ./regression/python/string23/main.py line 2 column 0
        ASSIGN space_check=0 ? { 115, 112, 97, 99, 101, 0 } : (signed char [6])(&{ 110, 111, 95, 115, 112, 97, 99, 101, 0 }[0]);
        // 774 file ./regression/python/string23/main.py line 3 column 7
        DECL signed int return_value$_strcmp$1;
        // 775
        FUNCTION_CALL:  return_value$_strcmp$1=strcmp(&space_check[0], &{ 115, 112, 97, 99, 101, 0 }[0])
        // 776 file ./regression/python/string23/main.py line 3 column 0
        ASSERT return_value$_strcmp$1 == 0
```

It seems not losing the `\0` but (lhs char comparison to rhs modification) is not working at the right time.

PR #2773, opened 2025-09-12
Replace libm pow operational model with MUSL implementation

sulaychaudhry:
The libm operational model for math.h's pow was inaccurate for floating point exponents, causing assertions that otherwise hold in the base language to fail verification (e.g. pow(4, 0.5) == 2 and pow(16, 0.25) == 2; and equivalents in Python).

This PR removes the libm's rough approximation of math.h pow and replaces it with the MUSL implementation, which has worst-case error of 0.54 ULP, runs quickly, and has no loops to unwind. From my testing so far this fixes the issue; the ESBMC correctly verifies pow assumptions.

As pow_by_squaring is no longer needed, it has been removed from the Python symbol table.

An alternative OpenLibm port was also successfully implemented to similar results, but MUSL has been used for consistency.


sulaychaudhry:
Running into issues with Python compatibility; again it works perfectly for the C frontend but Python doesn't seem to interpret libm into GOTO in the same way that C does. Structs don't seem to be supported? I get type symbol errors every time I try to run the Python frontend on code invoking pow. Will look again in the morning

image
image

If there is a way to tell the Python goto reader to create symbols for these structs, then it should be fixable, if not will have to port further to identical behaviour without structs

sulaychaudhry:
Python regression tests should perfectly work once #2782 is merged and with another PR to fix Python not initialising static & const variables correctly in imported C models.

src/c2goto/cprover_library.cpp
```diff
@@ -144,8 +144,20 @@ const static std::vector<std::string> python_c_models = {
   "det",
   "matmul",
   "pow",
+  "top12",
```
lucasccordeiro, line 147: `top12`, `checkint`, `specialcase`, `eval_as_double` and `fp_barrier` are generic names in a global whitelist and can collide with user functions. Prefix them.

regression/esbmc/pow1/main.c
```diff
@@ -0,0 +1,16 @@
+#include <math.h>
+#include <assert.h>
+
+/* Test basic integer powers */
+
+int main()
+{
+    assert(pow(2.0, 3.0) == 8.0);
```
lucasccordeiro, line 8: Every new test passes literal arguments, which constant-fold. Add a nondet-argument case so the musl implementation is actually symbolically executed.

regression/esbmc/pow1/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--z3
```
lucasccordeiro, line 3: Pinning `--z3` makes this test fail on builds without Z3, and the other pow tests do not pin it. Drop the flag.

lucasccordeiro (changes requested):

src/c2goto/library/libm/musl/pow.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/* Returns 1 if input is the bit representation of 0, infinity or nan.  */
+static inline int zeroinfnan(uint64_t i)
+{
+__ESBMC_HIDE:;
+	return 2 * i - 1 >= 2 * asuint64(INFINITY) - 1;
+}
+
+double pow(double x, double y)
```
lucasccordeiro, line 261: This model is entirely IEEE bit-punning (`asuint64`, `top12`, table indexing), which is wrong under `--fixedbv`; the arithmetic-only `libm/pow.c` it replaces was not. Add an `__ESBMC_FIXEDBV` variant as `libm/sqrt.c` does.

src/c2goto/library/libm/musl/pow.c
```diff
@@ -0,0 +1,357 @@
...
+	double_t ylo = y - yhi;
+	double_t lhi = asdouble(asuint64(hi) & -1ULL << 27);
+	double_t llo = hi - lhi + lo;
+	ehi = yhi * lhi;
+	elo = ylo * lhi + y * llo; /* |elo| < |ehi| * 2^-25.  */
+#endif
+	return exp_inline(ehi, elo, sign_bias);
+}
+
+// Keep old __pow for compatibility if needed
```
lucasccordeiro, line 352: `__pow` has no caller anywhere in src/ or regression/, and the comment narrates the change rather than a reason. Drop it, or state the ABI requirement that keeps it.

regression/esbmc/pow7/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--boolector
```
lucasccordeiro, line 3: This is the only pow test pinned to `--boolector`, an optional backend. If these NaN assertions only hold there, that is an encoding defect to fix, not a solver to pin.

regression/esbmc/pow-fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone also passes if the musl model fails on an unrelated property. Add a regex naming the expected failing assertion.

regression/esbmc/pow-fail2/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: Same as pow-fail: pin the expected failing property, otherwise any failure satisfies this test.

regression/esbmc/pow-fail/main.c
```diff
@@ -0,0 +1,9 @@
+#include <math.h>
+#include <assert.h>
+
+int main()
+{
+    int a = 2;
+    int b = 3;
+    assert(pow(a, b) == 9); // pow(2,3) == 8 not 9; should fail
+}
```
lucasccordeiro, line 9: All 13 new regression files lack a trailing newline. Add one to each.

PR #2779, opened 2025-09-16
[sv-comp] removed --add-symex-value-sets option

lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/17981309326

Master (30s):
````
  correct:           17513
    correct true:    10248
    correct false:    7265
  incorrect:            26
    incorrect true:     11
    incorrect false:    15
  unknown:           16030
  Score:             27169 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17895649321/
````

### CODE REVIEW

**Comment by rafaelsamenezes** [comment]
Minimized example:

```c
extern int __VERIFIER_nondet_int();
extern void abort(void);
#include <assert.h>
void reach_error() { assert(0); }

#include <stdlib.h>

typedef struct node {
    int val;
    struct node *next;
} Node;

int main() {
    Node *p, *list = malloc(sizeof(*list));
    list->next = NULL;
    list->val = 10;

    while (1) {
      for (p = list; p != NULL; p = p->next)
          __ESBMC_assert(p->val <= 20 && p->val >= 10, "check");         
    }
}
```

Something weird is happening with the while+for loop. Without the `--add-symex-value-sets option` it generates

```
        ASSIGN p=NONDET( struct node *);
        // 344 file test.c line 19 column 12 function main
     1: ASSIGN p=list;
        // 345 file test.c line 19 column 7 function main
        ASSIGN p=NONDET( struct node *);
        // 346 file test.c line 19 column 7 function main
     2: IF p == ( struct node *)0 THEN GOTO 3
        // 347 file test.c line 20 column 11 function main
        ASSERT p->val <= 20 && p->val >= 10 // check
        // 348 file test.c line 19 column 33 function main
        ASSIGN p=p->next;
        // 349 file test.c line 19 column 7 function main
        GOTO 2
        // 350 file test.c line 18 column 5 function main
     3: GOTO 1
```

Note that the jump in label 2 to 3 for the `while(1)` loop. The same program with the option on:

```
        ASSIGN p=list;
        // 332 file test.c line 19 column 7 function main
     1: IF p == ( struct node *)0 THEN GOTO 3
        // 333 file test.c line 20 column 11 function main
     2: ASSERT p->val <= 20 && p->val >= 10 // check
        // 334 file test.c line 19 column 33 function main
        ASSIGN p=p->next;
        // 335 file test.c line 19 column 7 function main
        GOTO 2
        // 336 file test.c line 18 column 5 function main
     3: GOTO 1
```

Note that label 2 no longer has a jump to label 3. So the for loop has no stop condition, leading to a spurious CE.

rafaelsamenezes:
#2833 solves the same issue.

PR #2781, opened 2025-09-17
[symex] add comprehensive realloc regression tests and fix realloc symbolic model

lucasccordeiro:
This PR introduces a significant number of new regression tests for the `realloc` function to improve the robustness of ESBMC's symbolic execution. The tests cover various `realloc` use cases, including:

- `realloc(NULL, size)` behavior as malloc.
- `realloc(ptr, 0)` to free memory and return NULL.
- Growing and shrinking memory blocks while preserving data.
- Modeling `realloc` failure and ensuring the old pointer remains valid.
- Complex scenarios such as dynamic vector growth and alternating resize operations.

This PR also refactors the realloc symbolic execution logic in `src/goto-symex/builtin_functions.cpp`. The updated model more accurately handles the data copying and failure semantics of `realloc`, addressing a known limitation with symbolic-sized reallocations. 

A new option, `--max-symbolic-realloc-copy`, has been added to provide a configurable limit for symbolic copying, balancing accuracy with performance.


lucasccordeiro:
This PR (30s):

````
  correct:           17521
    correct true:    10245
    correct false:    7276
  incorrect:            26
    incorrect true:     11
    incorrect false:    15
  unknown:           16022
  Score:             27174 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17791566481/
````

Master (30s):

````
  correct:           17534
    correct true:    10259
    correct false:    7275
  incorrect:            27
    incorrect true:     11
    incorrect false:    16
  unknown:           16008
  Score:             27185 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17733058520
````

XLiZHI:
Minor changes:
1. The description of the option has been updated. The default value should be 128.

2. I think there is no need to use exception handling when parsing the value of the option. Boost should have already checked the user's input value. 

3. We may need to consider some situations, such as the case where the input value is a negative number.

PR #2782, opened 2025-09-17
Fix symbol dependencies system for Python frontend

sulaychaudhry:
The Python frontend generates its symbol dependencies slightly differently to other frontends, as it has a function whitelist that it selectively pulls in symbols based on.

The old implementation was broken for a few reasons:
- Not traversing the irep tree far enough to find type symbols (specifically for C structs found in the OM libraries)
- Non-function symbols were discarded by the Python function filter, even if they could be potential dependencies later on
- Dependency symbols were looked for in new_ctx when Python already pulls in all of new_ctx

This PR changes the symbol dependencies system to fix these issues, while preserving functionality for the other frontends.
This allows C operational models containing structs to work on the Python frontend without type symbol issues.

### DIFF
--- a/src/c2goto/cprover_library.cpp
+++ b/src/c2goto/cprover_library.cpp
@@ -357,7 +357,8 @@ void add_cprover_library(contextt &context, const languaget *language)
        * have these dependencies already available in symbol_deps.
        * Therefore add dependencies that result from this new symbol
        */
-      if (language && language->id() == "python") {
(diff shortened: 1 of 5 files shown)


brcfarias (approved):

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @sulaychaudhry.

PR #2784, opened 2025-09-18
[simplifier] Add simplification rules for self-comparisons and idempotent expressions

lucasccordeiro:
This PR extends the simplification logic in `simplify_expr2.cpp` by adding new rules for logical, bitwise, and relational operators. The goal is to reduce redundant expressions early in the pipeline, thus improving the performance of generated SMT formulas.


lucasccordeiro:
This PR (30s):

````
  correct:           17542
    correct true:    10271
    correct false:    7271
  incorrect:            26
    incorrect true:     11
    incorrect false:    15
  unknown:           16001
  Score:             27221 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17842407204/job/50736057436
````

Master (30s):

````
  correct:           17523
    correct true:    10245
    correct false:    7278
  incorrect:            26
    incorrect true:     11
    incorrect false:    15
  unknown:           16020
  Score:             27176 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/17842409735/job/50736958777
````

PR #2785, opened 2025-09-19
[simplifier] add arithmetic and bitwise expression simplifications

lucasccordeiro:
This PR extends our expression simplifier with additional algebraic simplification rules for arithmetic and bitwise operations.


rafaelsamenezes (approved):
Looks ok overall. Would say to try the subtraction of MIN_INT on clang before merging.

lucasccordeiro:
> Looks ok overall. Would say to try the subtraction of MIN_INT on clang before merging.

Thanks! I have added it as a test case to our suite.

PR #2787, opened 2025-09-20
[python] Fix NoneType encoding

### LINKED ISSUE #2657: [python] NoneType is evaluated as unsigned long int 0
After conversion to symbol table the constant None is represented as the unsigned long int 0, resulting in code like the below incorrectly passing verification.

```python
x = None

if False:
    x = 0

assert x == 0
```

brcfarias:
Fixes #2657

### DIFF
--- a/regression/python/none-fail/main.py
+++ b/regression/python/none-fail/main.py
@@ -1,0 +1,2 @@
+x = None
+assert x == 0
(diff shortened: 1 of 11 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2803, opened 2025-09-27
[simplifier] improve pointer offset simplification in VCC generation

lucasccordeiro:
This PR enhances our expression simplifier to handle pointer offset expressions more effectively, reducing verification condition (VCC) complexity.

**Two key fixes to `pointer_offset2t::do_simplify()`:**
1. **NULL symbol handling**: Added recognition of NULL symbols to return zero offset
   - Improves expressions such as `POINTER_OFFSET((unsigned char *)0)` → `0`
2. **Typecast-wrapped array indices**: Added logic to look through typecasts when checking for constant array indices  
   - Improves expressions such as `POINTER_OFFSET(&arr[(signed long int)7])` → `28`

**Before:**

````C
4 == (unsigned long int)(POINTER_OFFSET((unsigned char *)0) + 4)
(unsigned long int)(POINTER_OFFSET((void *)(&arr[7]))) == 28
````

**After:**

````C
4 == 4 → true (simplified away)
28 == 28 → true (simplified away)
````


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/18063187049

Master (30s): https://github.com/esbmc/esbmc/actions/runs/18062339467

PR #2805, opened 2025-09-28
[python] enable --force-malloc-success

brcfarias:


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1787,6 +1787,11 @@ bool esbmc_parseoptionst::parse_goto_program(
         exit(0);
     }
 
+    if (cmdline.isset("force-malloc-success"))
+    {
+      options.set_option("force-malloc-success", true);
+    }
```
rafaelsamenezes, line 1793: The if can be removed:
```c
options.set_option("force-malloc-success", cmdline.isset("force-malloc-success"));
```

Also, shouldn't this be done at the frontend? If we try to mix C and Python code, how will this interact?

I am also in favour to revert the force-malloc-success flag to malloc-can-fail. I don't think people rely on this.

rafaelsamenezes (changes requested):
This PR is doing 3 things:

1. Enabling force-malloc-success for python
2. Extend the cmdlinet class to allow setting flags manually. But then also reuse the options object to re-read  some of the flags. That seems a bit confusing and I am not so sure about it.
3. Enable testing that relies on the OMs from c2goto.

Could you improve the PR description (or split in 3 individual PRs)? 

Also, wouldn't be better to use `bool parse(const std::string &filename, cmdlinet &cmdline);` with an `optiont` instead? That way you could avoid 2.

brcfarias:
Nice. The cmdline is populated via `parse()` call in `esbmc_parseoptionst() `-> `parseoptions_baset()`, and then we read the flags from cmdline to populate the optionst at `esbmc_parseoptionst::get_command_line_options()`. After that `optionst` is forwarded for the goto generation. 
Since it is now necessary to set a flag after parsing the user options, it was necessary to add this new method. But if we could inject the optionst into `python_languaget`, then we could simply change it there.

> Also, wouldn't be better to use bool parse(const std::string &filename, cmdlinet &cmdline); with an optiont instead? That way you could avoid 2.

Yes, I totally agree. I just didn’t want to change the interfaces initially.

lucasccordeiro:
This PR is not valid anymore.

PR #2809, opened 2025-09-29
[clang-c-frontend] Implement __atomic_exchange_n builtin

### LINKED ISSUE #2808: [sv-comp] incorrect false for libvsync/cnalock.i?
$ esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check  cnalock.i --64 --memory-leak-check --no-reachable-memory-leak --no-assertions --no-abnormal-memory-leak --malloc-zero-is-null --incremental-bmc

````
[Counterexample]


State 9 file cnalock.i line 3898 column 5 function vatomicptr_xchg thread 1
  tmp = invalid-object

State 10 file cnalock.i line 5391 column 5 function cnalock_acquire thread 1
  tail = ( struct cna_node_s *)(invalid-object)

State 12 file cnalock.i line 3846 column 5 function __atomic_store_n thread 1
Violated property:
  file cnalock.i line 3846 column 5 function __atomic_store_n
  dereference failure: invalid pointer


VERIFICATION FAILED

Bug found (k = 1)
````

### PR DESCRIPTION
**Pull request by lucasccordeiro**
Fixes https://github.com/esbmc/esbmc/issues/2808.

This PR implements the missing `__atomic_exchange_n` builtin to model atomic exchange operations properly. Implementation follows the existing pattern for atomic builtins:

1. Read and store the old value from the pointer location
2. Write the new value to the pointer location
3. Return the old value

All operations are wrapped in an atomic begin/end scope.

This atomic builtin follows GCC's __atomic built-in functions specification (https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html). Other atomic implementations, such as C11 stdatomic.h and Clang's _c11_atomic* extensions exist but are not addressed in this commit.

### DIFF
--- a/src/clang-c-frontend/clang_c_adjust_polymorphic_functions.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_polymorphic_functions.cpp
@@ -427,6 +427,8 @@ code_blockt clang_c_adjust::instantiate_gcc_polymorphic_builtin(
   }
   else if (has_prefix(identifier.as_string(), "c:@F@__atomic_exchange_n"))
   {
+    // This atomic builtin follows GCC's __atomic built-in functions specification.
+    // See https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html.
     const typet &type = code_type.return_type();
 
     const exprt &result =
--- a/src/clang-c-frontend/clang_c_adjust_polymorphic_functions.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_polymorphic_functions.cpp
@@ -427,7 +427,42 @@ code_blockt clang_c_adjust::instantiate_gcc_polymorphic_builtin(
   }
   else if (has_prefix(identifier.as_string(), "c:@F@__atomic_exchange_n"))
   {
-    // TODO
+    const typet &type = code_type.return_type();
+
+    const exprt &result =
+      symbol_expr(result_symbol(identifier_with_type, type, context));
+
+    code_declt decl(result);
+    block.operands().push_back(decl);
+
+    code_typet::argumentt arg0 = code_type.arguments()[0];
+    code_typet::argumentt arg1 = code_type.arguments()[1];
+
+    dereference_exprt arg0_deref(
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/clang-c-frontend/clang_c_adjust_polymorphic_functions.cpp [hunk 1]
```diff
@@ -427,7 +427,42 @@ code_blockt clang_c_adjust::instantiate_gcc_polymorphic_builtin(
...
+    block.operands().push_back(assign_old);
+
+    // Store new value at pointer location
+    code_assignt assign_new(
+      arg0_deref, symbol_exprt(arg1.cmt_identifier(), arg1.type()));
+    assign_new.location() = new_loc;
+    block.operands().push_back(assign_new);
+
+    // atomic scope end
+    side_effect_expr_function_callt atomic_end;
```
XLiZHI, line 458: ~~Should `return` be inside an atomic block?~~
Never mind, return should be at the end.
```c
__ESBMC_atomic_begin();
result = *ptr;
*ptr = val;
return result;
__ESBMC_atomic_end();
```

XLiZHI (approved):

lucasccordeiro:
Master (30s):

````
  correct:           17560
    correct true:    10281
    correct false:    7279
  incorrect:            26
    incorrect true:     11
    incorrect false:    15
  unknown:           15983
  Score:             27249 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/18079581235/job/51441368755
````

This PR (30s): 

````
  correct:           17562
    correct true:    10285
    correct false:    7277
  incorrect:            23
    incorrect true:     11
    incorrect false:    12
  unknown:           15984
  Score:             27303 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/18097175216
````

intrigus-lgtm:
Is there any reason why we don't replace all those TODOs with an intrinsic that fails when reachable?
`__ESBMC_unreachable` cannot be used, as it's not enabled by default.
So either something custom, or we switch `--enable-unreachability-intrinsic` on by default.

PR #2816, opened 2025-09-30
[python] Pass temporary variable as self for constructors

### LINKED ISSUE #2812: Exception when calling a function that returns an object
For this Python program:
```python
class Foo:
    def __init__(self, x: int):
        self.x = x

def foo(x: int) -> Foo:
    return Foo(x)

# Works fine:
#f = Foo(4)
# Throws an exception:
f = foo(4)
```
ESBMC throws an exception:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.470s
terminate called after throwing an instance of 'type2t::symbolic_type_excp'
Aborted (core dumped)
```
Creating the object through `f = Foo(4)` works fine.

brcfarias:
Fixes #2812

### DIFF
--- a/regression/python/gb-2812/main.py
+++ b/regression/python/gb-2812/main.py
@@ -1,0 +1,8 @@
+class Foo:
+    def __init__(self, x: int):
+        self.x = x
+
+def foo(x: int) -> Foo:
+    return Foo(x)
+
+f = foo(4)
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

lucasccordeiro:
@brcfarias and @XLiZHI: Thanks for submitting this PR!

PR #2845, opened 2025-10-09
Unified constant propagation of with expressions

XLiZHI:
```
int main()
{
  typedef int *intptr;
  intptr a[10];
  int x, y;
  
  unsigned i;
  __ESBMC_assume(i<10);

  a[i]=&x;
  a[5]=&y;
  
  assert(*(a[i])==x);  
}
```

We should not simplify when the expressions are the same. If this is the case, the VCC of this program is incorrectly simplified. Because i can be equal to 5.


XLiZHI:
This PR:
https://github.com/esbmc/esbmc/actions/runs/18426879208

Master: 
https://github.com/esbmc/esbmc/actions/runs/18426880581

lucasccordeiro:
This PR (30s):

````
  correct:           17723
    correct true:    10413
    correct false:    7310
  incorrect:            23
    incorrect true:     12
    incorrect false:    11
  unknown:           15823
  Score:             27576 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/18491386716/job/52687058936
````

Master (30s):

````
  correct:           17735
    correct true:    10419
    correct false:    7316
  incorrect:            22
    incorrect true:     12
    incorrect false:    10
  unknown:           15812
  Score:             27610 (max: 55885)

https://github.com/esbmc/esbmc/actions/runs/18471944758/job/52629135521
````

lucasccordeiro (approved):

PR #2868, opened 2025-10-16
[python] fix boost::process compilation on macOS

lucasccordeiro:
Boost >= 1.87 defaults to process v2, which lacks `search_path`. Conditionally use v1 on macOS and newer Boost versions while maintaining compatibility with older Linux systems.


rafaelsamenezes (approved):
Seems to be compiling fine on my Mac

PR #2894, opened 2025-10-20
Memcpy optimization

rafaelsamenezes:
This pull request introduces an intrinsic implementation for the `memcpy` function in the symbolic execution engine, enabling more precise modeling and verification of memory copy operations in C programs.  Currently it only works for primitives, and it has an extra preprocessing cost (which makes it slower for small memcpy and faster for big memcpy). 

30s run:
```
  correct:           17695
    correct true:    10385
    correct false:    7310
  incorrect:            22
    incorrect true:     11
    incorrect false:    11
  unknown:           15852
  Score:             27552 (max: 55885)
```

This is a draft as I am doing another 30s run with this branch and master just to be extra sure.


src/goto-symex/builtin_functions.cpp
```diff
@@ -1673,6 +1673,367 @@ static inline expr2tc gen_value_by_byte(
...
+  if (
+    dst->type == src->type && !dst_offset && !src_offset &&
+    type_byte_size(dst->type).to_uint64() == num_of_bytes)
+    return src;
+
+  if (
+    is_array_type(src->type) || is_array_type(dst->type) ||
+    is_struct_type(dst->type) || is_union_type(dst->type) ||
+    is_struct_type(src->type) || is_union_type(src->type))
+    return expr2tc();
```
lucasccordeiro, line 1757: Can you document that this is a limitation? It will be addressed in future PRs!

src/goto-symex/builtin_functions.cpp
```diff
@@ -1673,6 +1673,367 @@ static inline expr2tc gen_value_by_byte(
   return gen_byte_expression(type, src, value, num_of_bytes, offset);
 }
 
+// Computes the equivalent object value when considering a memcpy operation on it
+static inline expr2tc gen_byte_memcpy(
```
lucasccordeiro: Shall we create unit tests for this method? The logic itself is intricate and might be prone to off-by-one or bit-shifting errors :-)

lucasccordeiro (approved):
This is a great PR ;-) Well done!

I have just minor comments.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Added a log_debug message with a todo.

regression/cstd/intrinsic_memcpy_bug4/main.c
```diff
@@ -0,0 +1,14 @@
...
+int main()
+{
+  int a;
+  int b;
+
+  char *ptr1 = &a;
+  char *ptr2 = &b;
+  memcpy(ptr1+1, ptr2+1, sizeof(char));
+
+  assert(a == b);
```
intrigus-lgtm, line 12: What is this testing?

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -392,7 +392,8 @@ _Bool __ESBMC_is_little_endian();
 int __ESBMC_rounding_mode = 0;
 
 void *__ESBMC_memset(void *, int, unsigned int);
```
intrigus-lgtm:
```suggestion
void *__ESBMC_memset(void *, int, __SIZE_TYPE__);
```

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -392,7 +392,8 @@ _Bool __ESBMC_is_little_endian();
...
-
+      void *__ESBMC_memcpy(void *, const void *, unsigned long long);
```
intrigus-lgtm:
```suggestion
void *__ESBMC_memcpy(void *, const void *, __SIZE_TYPE__);
```

src/c2goto/library/string.c
```diff
@@ -269,7 +269,7 @@ __ESBMC_HIDE:;
   return cpy;
 }
 
-void *memcpy(void *dst, const void *src, size_t n)
+void *__memcpy_impl(void *dst, const void *src, unsigned long long n)
```
intrigus-lgtm:
```suggestion
void *__memcpy_impl(void *dst, const void *src, size_t n)
```

intrigus-lgtm (changes requested):
Initial review.

src/goto-symex/goto_symex.h
```diff
@@ -466,6 +466,19 @@ class goto_symext
     reachability_treet &art,
     const code_function_call2t &func_call);
 
+  /**
+   * @brief Intrinsic call for C memset function call
```
intrigus-lgtm: Refers to memset instead of memcpy.

regression/cstd/intrinsic_memcpy_bug4/main.c
rafaelsamenezes: Checking whether offsets are being handled.

src/goto-symex/builtin_functions.cpp
rafaelsamenezes: Working on it

src/clang-c-frontend/clang_c_language.cpp
rafaelsamenezes: Implemented as suggested

src/c2goto/library/string.c
rafaelsamenezes: Implemented as suggested

src/clang-c-frontend/clang_c_language.cpp
rafaelsamenezes: Implemented as suggested

PR #2913, opened 2025-10-21
[python verification] fix exception with formatted string literal

### LINKED ISSUE #2912: [Python] Exception with formatted string literal
For the following program:
```python
def foo(s: str) -> None:
    if s != "foo":
        raise ValueError(f"Unexpected value '{s}' passed to foo()")

foo("foo")
```
ESBMC throws an exception:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
  what():  [json.exception.type_error.302] type must be string, but is number
Aborted (core dumped)
```
ESBMC version: 9ed399899

XLiZHI:
Fixes https://github.com/esbmc/esbmc/issues/2912.

### DIFF
--- a/regression/python/github_2912/main.py
+++ b/regression/python/github_2912/main.py
@@ -1,0 +1,5 @@
+def foo(s: str) -> None:
+    if s != "foo":
+        raise ValueError("Unexpected value '{s}' passed to foo()")
+
+foo("foo")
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2918, opened 2025-10-22
[python] Fix relative imports

### LINKED ISSUE #2915: [Python] Module not found for nested module
Given a project with the following structure:
```bash
.
├── ex.py
└── l
    ├── __init__.py
    └── ks
        ├── __init__.py
        └── foo.py

2 directories, 4 files
```
and the following file contents:
`ex.py`:
```python 
import l
from l.ks import foo

f: foo.Foo = l.create_foo("foo")
```
`l/__init__.py`:
```python 
from .ks.foo import Foo

def create_foo(s: str) -> Foo:
    return Foo(s)
```
`l/ks/__init__.py`:
```python 
"""KS"""
```
`l/ks/foo.py`:
```python 
class Foo:
    def __init__(self, s: str) -> None:
      self.s = s
```
ESBMC errors out as follows:
```bash
$ esbmc ex.py 
Parsing ex.py
ERROR: Module 'ks.foo' not found.
Please install it with: pip3 install ks.foo
```
Is there a different directory structure that ESBMC would accept?

ESBMC version: 92da1947b

brcfarias:
Fixes #2915


lucasccordeiro (approved):

lucasccordeiro:

PR #2940, opened 2025-10-24
[python verification] support isinstance with tuples

### LINKED ISSUE #2933: [Python] Feature request: support `isinstance` with tuples
Python's built-in `isinstance` supports passing tuples as an argument in which case it returns `True` if the object is an instance of any of the types: https://docs.python.org/3/library/functions.html#isinstance

Here's an example program:
```python
s: int = 5
assert isinstance(s, (int, str))
```
which is equivalent to:
```python3
assert isinstance(s, int) or isinstance(s, str)
```
ESBMC currently errors out with the tuple:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
ERROR: Unsupported type format in isinstance()
```

XLiZHI:
fixes #2933

### DIFF
--- a/regression/python/github_2933/main.py
+++ b/regression/python/github_2933/main.py
@@ -1,0 +1,3 @@
+s: int = 5
+
+assert isinstance(s, (int, str))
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2941, opened 2025-10-24
 [python] Fix: Handle empty width string in get_type_width on macOS

### LINKED ISSUE #2937: stoi crash for MacOS
This example crashes on MacOS:

```python
from typing import Optional


class EchoMessageInput:
    def __init__(self, message: Optional[str]):
        self.message = message

    def _verify(self) -> None:
        if self.message is not None:
            assert 1 <= len(self.message)
            assert len(self.message) <= 10


class EchoService:
    def echo_message(self, input: EchoMessageInput) -> None:
        input._verify()


# Create client
client = EchoService()

# Create input
input_data = EchoMessageInput(message="Hello")

# Call the service
client.echo_message(input_data)
```

with:

```
ESBMC version 7.11.0 64-bit aarch64 macos
Parsing echo_example.py
Converting
libc++abi: terminating due to uncaught exception of type std::invalid_argument: stoi: no conversion
[1]    21855 abort      esbmc echo_example.py
```

LukeW1999:
Fixes #2937.

### Problem
ESBMC-Python was crashing on macOS due to differences in how macOS and Linux handle symbol processing. When `type.width().c_str()` returns an empty string, calling `std::stoi()` directly causes an exception.

### Solution
Added a check to verify the width string is not empty before attempting to parse it. If empty or parsing fails, the function now falls back to inferring the type width from the type name.

### Testing
- Tested on Ubuntu 22.04 to ensure no regression in existing ESBMC-Python functionality
- Resolves the macOS crash while maintaining backward compatibility

### Changes
- Added empty string validation before `std::stoi()` call
- Improved error handling flow in `get_type_width()`

### TODO
Currently, ESBMC-Python fails multiple regression tests on macOS, mostly due to minor differences in how macOS and Linux handle certain operations. These issues are expected to be addressed incrementally in the near future.

This PR should have fixed 15 related regression failing cases on the Mac.
#### master:
- **passed**: 723 
- **failed**:  49 
- **skipped**: 133 (THOROUGH)
#### this:
- **passed**: 738 
- **failed**: 34 
- **skipped**: 133

### DIFF
--- a/src/python-frontend/type_handler.cpp
+++ b/src/python-frontend/type_handler.cpp
@@ -756,7 +756,7 @@ int type_handler::get_array_dimensions(const nlohmann::json &arr) const
 
 size_t type_handler::get_type_width(const typet &type) const
 {
-  // First try to parse width directly  
+  // First try to parse width directly
   std::string width_str = type.width().c_str();
   if (!width_str.empty())
   {
@@ -769,12 +769,11 @@ size_t type_handler::get_type_width(const typet &type) const
       // Fall through to type name inference
     }
   }
-  
+
   // If width is empty or parsing failed, try to infer from type name
   std::string type_str = type.width().as_string();
   if (!type_str.empty())
   {
-
     // Handle common Python/ESBMC type mappings
     if (type_str == "int32")
       return 32;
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

PR #2949, opened 2025-10-25
Use __ESBMC_alloca instead of malloc in list.c

rafaelsamenezes:
Stack allocations:

- Always works
- Will not trigger memleaks (useful if we eventually mix C/Python code)


lucasccordeiro:
Excellent, @rafaelsamenezes!

lucasccordeiro (approved):

lucasccordeiro:
@rafaelsamenezes: could you also remove all --force-malloc-success from our regression suite?

PR #2953, opened 2025-10-26
memcpy optimization

rafaelsamenezes:
This pull request  of `memcpy` operations within  arrays

## Notes for python

This *could* improve the python lists. For example

```
a = [1,2,3,4,5]
``` 

Could potentially be decreased from 21s to 8s if some extra features are implemented.

1. Move list hashing to the frontend. Its only role is to convert type information into a numeric representation.
2. Investigate why list push operations still perform dereferencing/address-of computations even when list size is constant. This prevents this programs constructs from being verified without incremental solving.
3. Implement direct initialization for static Python list literals. A memcpy call here is unnecessary. 

Moreover, item 2 is the only one needed to fixed before this patch affects python.

```diff
static inline bool list_push_object(List *l, Object *o)
 {
   assert(l != NULL);
   assert(o != NULL);
-  return list_push(l, o->value, o->type_id, o->size);
+  return list_push(l, o->value, o->type_id, 8);
 }
```

```diff
 static inline size_t list_hash_string(const char *str)
 {
+#if 0
   size_t hash = 5381;
   int c;
   while ((c = *str++))
   {
     hash = ((hash << 5) + hash) + c;
   }
-  return hash;
+#endif
+  return (size_t)str;
 }
```

Which are hacks to demonstrate optimizations that should be done.


lucasccordeiro (approved):

rafaelsamenezes:
I tested this in benchexec, it seems to have the same overall performance for 120s. I got the same number of benchmarks if I restricted to 100s.

PR #2969, opened 2025-10-28
[python] Fix macOS different exception catch

LukeW1999:
#### reason: 
Bug should come from the different version of Libc++ of Mac and Linux.

#### Result: 
Now the MacOS ESBMC python could pass the python regression test. 
[result](https://github.com/LukeW1999/esbmc_dev_CI/actions/runs/18851938015/job/53790412936)


lucasccordeiro (approved):

lucasccordeiro:

PR #2970, opened 2025-10-28
[python] fix #2966: Problem of isinstance handles ternary operators 

### LINKED ISSUE #2966: [Python] isinstance assertion unexpectedly fails
Program:
```python3
from typing import Any

i = 42
x: Any = "foo" if i > 10 else i + 10
assert isinstance(x, str)
```
ESBMC output:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
WARNING: Ternary branches have incompatible types: array vs signedbv, using default pointer
Generating GOTO Program
GOTO program creation time: 0.401s
Symex completed in: 0.000s (4 assignments)
Slicing time: 0.000s (removed 3 assignments)
Building error trace

[Counterexample]


State 1 file ex.py line 5 column 0 thread 0
Violated property:
  file ex.py line 5 column 0
  assertion
  0


VERIFICATION FAILED
```
ESBMC version: 97828b786

LukeW1999:
Fixes https://github.com/esbmc/esbmc/issues/2966.

## Description

Fixes `isinstance()` incorrectly failing when checking variables assigned from ternary operators with mixed types.

**Example:**
```python
from typing import Any
i = 42
x: Any = "foo" if i > 10 else i + 10
assert isinstance(x, str)  # Was: ASSERT 0, Now: runtime check
```

## Solution

Modified `handle_isinstance()` to detect IfExp assignments and generate runtime type checks instead of static `ASSERT 0`:
- Both branches match → `ASSERT 1` 
- Neither matches → `ASSERT 0`
- Mixed → `ASSERT (cond ? true : false)` 

## Testing

Added regression tests:
- `regression/python/github_2966/` (pass case)
- `regression/python/github_2966_fail/` (fail case)

### DIFF
--- a/regression/python/github_2966/main.py
+++ b/regression/python/github_2966/main.py
@@ -1,0 +1,8 @@
+# Test case for GitHub issue #2966
+# isinstance should work correctly with ternary operators
+from typing import Any
+
+i = 42
+x: Any = "foo" if i > 10 else i + 10
+assert isinstance(x, str)
+
--- a/regression/python/github_2966/test.desc
+++ b/regression/python/github_2966/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
+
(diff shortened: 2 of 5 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #2971, opened 2025-10-28
Python c2goto fix

rafaelsamenezes:
Add cprover library had the wrong check, which was breaking adding intrinsic functions to the Python frontend. This had the workaround of manually adding the symbol, but now we don't need it anymore so I moved the definitions out of src/python-frontend/function_call_builder.cpp

### DIFF
--- a/src/c2goto/cprover_library.cpp
+++ b/src/c2goto/cprover_library.cpp
@@ -93,6 +93,11 @@ static const struct buffer
 // The goto reader will only pick up symbols for these functions and their dependencies
 // This is a Python-specific whitelist invoked when you use set_functions_to_read
 const static std::vector<std::string> python_c_models = {
+  "__ESBMC_get_object_size",
+  "__ESBMC_assume",
+  "__VERIFIER_assume",
+  "__ESBMC_loop_invariant",
+  "strlen",
   "list_create",
   "list_in_bounds",
   "list_at",
@@ -334,7 +339,7 @@ void add_cprover_library(contextt &context, const languaget *language)
     const symbolt *symbol = context.find_symbol(s.id);
     if (
       (language && language->id() == "python") ||
-      (symbol != nullptr && symbol->value.is_nil()))
+      (symbol != nullptr && !symbol->value.is_nil()))
     {
       store_ctx.add(s);
 
--- a/src/python-frontend/function_call_builder.cpp
+++ b/src/python-frontend/function_call_builder.cpp
@@ -9,13 +9,6 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 
-const std::string kGetObjectSize = "__ESBMC_get_object_size";
-const std::string kStrlen = "strlen";
-const std::string kEsbmcAssume = "__ESBMC_assume";
-const std::string kVerifierAssume = "__VERIFIER_assume";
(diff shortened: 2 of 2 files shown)


lucasccordeiro (approved):
Nice! Thanks for submitting this PR, @rafaelsamenezes.

src/c2goto/cprover_library.cpp
```diff
@@ -334,7 +339,7 @@ void add_cprover_library(contextt &context, const languaget *language)
     const symbolt *symbol = context.find_symbol(s.id);
     if (
       (language && language->id() == "python") ||
-      (symbol != nullptr && symbol->value.is_nil()))
+      (symbol != nullptr && !symbol->value.is_nil()))
```
lucasccordeiro, line 342: Inverting the test supplies library bodies only for symbols the program already defines, and drops them for declaration-only functions. Keep `is_nil()`.

lucasccordeiro (changes requested):
Branch is stale: master has since rewritten this condition to `is_solidity || uses_whitelist`, and the deleted `len`/loop-invariant symbol creation now has no fallback outside the whitelist. Rebase before further review.

src/c2goto/cprover_library.cpp
```diff
@@ -93,6 +93,11 @@ static const struct buffer
 // The goto reader will only pick up symbols for these functions and their dependencies
 // This is a Python-specific whitelist invoked when you use set_functions_to_read
 const static std::vector<std::string> python_c_models = {
+  "__ESBMC_get_object_size",
+  "__ESBMC_assume",
+  "__VERIFIER_assume",
+  "__ESBMC_loop_invariant",
+  "strlen",
```
lucasccordeiro, line 100: Whitelisting `strlen` links the real `size_t strlen(const char *)` body, replacing the synthesized `long long int (void *)` declaration Python `len()` used. Add a regression test pinning `len()` on `str`.

src/python-frontend/function_call_builder.cpp
```diff
@@ -37,13 +30,13 @@ bool function_call_builder::is_numpy_call(const symbol_id &function_id) const
...
   const std::string &func_name = function_id.get_function();
-  return (func_name == kEsbmcAssume || func_name == kVerifierAssume);
+  return (func_name == "__ESBMC_assume" || func_name == "__VERIFIER_assume");
 }
 
 bool function_call_builder::is_len_call(const symbol_id &function_id) const
 {
   const std::string &func_name = function_id.get_function();
-  return func_name == kGetObjectSize || func_name == kStrlen;
+  return func_name == "__ESBMC_get_object_size" || func_name == "strlen";
```
lucasccordeiro, line 39: `is_len_call` loses its only caller when the symbol-table block below is deleted, leaving it dead. Remove it and its declaration in `function_call_builder.h`.

src/python-frontend/function_call_builder.cpp
```diff
@@ -9,13 +9,6 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 
-const std::string kGetObjectSize = "__ESBMC_get_object_size";
```
lucasccordeiro, line 12: These names must now match `python_c_models` in cprover_library.cpp exactly, or the symbol silently disappears. Keep them as shared constants rather than inlined literals.

PR #2972, opened 2025-10-28
[python] Crash when __init__ calls a method defined later

### LINKED ISSUE #2962: [Python] Crash when `__init__` calls another method
Program:
```python3
class Foo:
    def __init__(self) -> None:
        self._init_members()

    def _init_members(self) -> None:
        pass

f = Foo()
```
ESBMC output:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
```
ESBMC version: 97828b786

LukeW1999:
Fixes https://github.com/esbmc/esbmc/issues/2962.

### Problem
ESBMC crashes with segmentation fault when a Python class's `__init__` method calls an instance method defined later in the class:
```python
class Foo:
    def __init__(self):
        self._init_members()  # ← Crash: _init_members not yet in symbol table
    
    def _init_members(self):
        pass
```

### Root Cause
The Python frontend processes class methods **sequentially in definition order**. When processing the `__init__` method body, subsequent methods like `_init_members` have not yet been added to the symbol table. This causes `func_symbol` to be `nullptr`, triggering either an `assert` failure or null pointer dereference.

**Why C++ doesn't have this problem:** The C++ frontend uses Clang, which performs complete semantic analysis before ESBMC conversion. Clang scans all method declarations in a class first (building the complete symbol table), then processes method bodies. This two-phase approach naturally handles forward references.

### Solution
When an instance method symbol is not found, create a **forward reference**:
- Use the symbol ID to create `symbol_exprt` (doesn't require the symbol to exist yet)
- Defer symbol resolution to the linking phase
- Return immediately to avoid null pointer dereference

This approach mimics how compilers (including Clang) handle forward references.

### Future Improvements
A more fundamental solution would be **two-phase class processing**:
Register all method signatures (build complete symbol table). Then, process method bodies (all symbols now available)

This matches the standard compiler design but requires a refactoring development on the python frontend I guess.

### DIFF
--- a/regression/python/github_2962/main.py
+++ b/regression/python/github_2962/main.py
@@ -1,0 +1,8 @@
+class Foo:
+    def __init__(self) -> None:
+        self._init_members()
+
+    def _init_members(self) -> None:
+        pass
+
+f = Foo()
--- a/regression/python/github_2962/test.desc
+++ b/regression/python/github_2962/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_2962_fail/main.py
+++ b/regression/python/github_2962_fail/main.py
@@ -1,0 +1,10 @@
+# Test for Issue #2962: Method is properly executed and can fail
+# Verifies that _init_members() called from __init__ actually runs
+class Foo:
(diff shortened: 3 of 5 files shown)


lucasccordeiro (approved):

PR #2977, opened 2025-10-28
[python] Fix nested attribute chain method calls in Python frontend

### LINKED ISSUE #2974: [Python] Variable not found error
Code:
```python3
class Foo:
    def __init__(self) -> None:
        pass

    def foo(self) -> int:
        return 42

class Bar:
    def __init__(self) -> None:
        self.f = Foo()

    def bar(self) -> int:
        return self.f.foo()
```
ESBMC output:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
ERROR: Variable f not found
```

LukeW1999:
## Problem
When calling methods through nested attribute chains (e.g., `self.f.foo()` or `a.b.c.method()`), ESBMC fails with "Variable f not found" error.

**Example:**
```python
class Foo:
    def foo(self) -> int:
        return 42

class Bar:
    def __init__(self) -> None:
        self.f: Foo = Foo()
    
    def bar(self) -> int:
        return self.f.foo()  # ERROR: Variable f not found
```

**Root cause:** The old code extracts the direct attribute name `"f"` and tries to find it as a local variable, instead of resolving the type of `self.f` to determine it's a `Foo` instance.

## Solution
Implemented **recursive type resolution** that walks the attribute chain from left to right, resolving each component's type through the symbol table.

For `self.f.foo()`:
1. Resolve `self` → type `Bar` (from symbol table)
2. Resolve `self.f` → type `Foo` (lookup member `f` in `Bar` struct)
3. Extract class name `Foo` and use it to find method `foo()`

**Implementation:**
- Added recursive type resolver in `function_call_builder.cpp`
- Modified `get_object_name()` in `function_call_expr.cpp` to reuse resolved types
- Handles pointer dereferencing and symbol type following
- Falls back to old behavior if type resolution fails

### Modified files
- `src/python-frontend/function_call_builder.cpp` - Added recursive type resolver
- `src/python-frontend/function_call_expr.cpp` - Reuse resolved class names

### New test cases
- `github_2972` / `github_2972_fail` - Single-level chains: `self.f.foo()`
- `github_2972_1` / `github_2972_1_fail` - Multi-level chains: `self.b.a.method()`

**Note:** This fix also handles **multi-level attribute chains** (e.g., `a.b.c.method()`) beyond the single-level case reported in the original issue.

## Related Issue
Fixes #2974

##

Thanks to @zhassan-aws for discovering this bug and for your continued contributions and patience. We truly appreciate your support!


zhassan-aws:
Thanks for the quick fix!

> Fixes https://github.com/esbmc/esbmc/pull/2972

Should this be marked as fixing https://github.com/esbmc/esbmc/pull/2974 instead?

LukeW1999:
> Thanks for the quick fix!
> 
> > Fixes #2972
> 
> Should this be marked as fixing #2974 instead?

Oops, Sir. My brain was still on this morning's PR. 

Fixed, thanks!

LukeW1999:
@zhassan-aws , Sir, I sincerely apologize for tagging you before the implementation was fully tested. I was only testing in RelWithDebInfo mode locally, while the CI also runs DebugOpt mode which has stricter assertions enabled.

**Current Status:**
- ✅ Basic functionality (single-level calls like `self.f.foo()`) works perfectly in all build modes
- ⚠️ Multi-level nested calls (`self.b.a.method_a()`) have a pointer offset calculation issue in DebugOpt mode

I will work on fixing the nested calls today. I've learned from this mistake and will ensure my local testing matches the CI build configurations before submitting a PR and notify you. This won't happen again.

Thank you for your patience and understanding.

LukeW1999:
## Issue Found in DebugOpt Mode

While testing this PR with ESBMC's comprehensive build configurations, I discovered that the implementation only works in **RelWithDebInfo** mode but fails in **DebugOpt** mode. This is a good example of how ESBMC's rigorous testing standards help catch potential issues early.

### Root Cause

The issue stems from ESBMC's well-designed type safety guarantees. The `member2t` constructor has an assertion that enforces a critical invariant: in the irep2_expr

```cpp
assert(source->type->type_id == type2t::struct_id || 
       source->type->type_id == type2t::union_id);
```

This assertion is **active in DebugOpt** but removed by `-DNDEBUG` in RelWithDebInfo. In my previous commit, the Python frontend models members (e.g., `self`) as pointers, which violates this invariant.

### Proposed Fix

I've implemented a fix inspired by the Clang frontend's `adjust_member` approach. The key changes are:

**1. Insert dereference when creating member expressions** 

```cpp
// Apply adjust_member logic (from Clang frontend): insert dereference if source is pointer
exprt &base = member_expr.struct_op();
if (base.type().is_pointer())
{
  exprt deref("dereference");
  deref.type() = base.type().subtype();
  deref.move_to_operands(base);
  base.swap(deref);
}
```

**2. Correctly classify nested instance attributes**
- Distinguish between `self.b.a.method()` (instance method) and `module.Class.method()` (class method)
- Check the root of the attribute chain to determine the correct function type

**3. Recursively handle nested `Attribute` nodes**
- Build expressions dynamically for multi-level chains
- Apply pointer adjustment at each level

### Current Status

**✅ What works:**
- Single-level chains (`self.f.foo()`) pass all tests in DebugOpt now
- The `python` and 'numpy' regression test suite passes on my device

**⚠️ Known limitation:**
Multi-level chains (`self.b.a.method_a()`) generate correct GOTO programs (`method_a(&self->b.a)`) but trigger a runtime "illegal offset" error during pointer verification.

With `--no-pointer-check`, the logic is correct and produces expected results, suggesting the issue is in address calculation rather than logic errors.

### Request for Review

I believe this is related to how nested member offsets are computed, but I'd appreciate any feedback from contributors familiar with:

1. The Clang frontend's `clang_c_adjust` implementation
2. ESBMC's memory model and offset calculation for nested structures
3. Whether the Python frontend should have a unified "adjust" phase like Clang

I've marked the multi-level test as `KNOWNBUG` and added a test case with `--no-pointer-check` to document the current behavior.

lucasccordeiro (approved):

lucasccordeiro:

lucasccordeiro:
@LukeW1999: could you please work together with @XLiZHI to tackle the other parts of this PR?

PR #2984, opened 2025-10-29
[SV-COMP] support thread local for data races check

XLiZHI:
A thread-local variable is stored per thread, so each thread accesses its own private value instead of sharing it with other threads.

Although this PR does not actually support thread local, it does not suffer from a data race in a data race check.


XLiZHI:
This PR:
```
Statistics:           1029 Files
  correct:             419
    correct true:      307
    correct false:     112
  incorrect:             9
    incorrect true:      7
    incorrect false:     2
  unknown:             601
  Score:               470 (max: 1823)
https://github.com/esbmc/esbmc/actions/runs/18909956710
```

Master:
```
Statistics:           1029 Files
  correct:             420
    correct true:      307
    correct false:     113
  incorrect:            12
    incorrect true:      7
    incorrect false:     5
  unknown:             597
  Score:               423 (max: 1823)
https://github.com/esbmc/esbmc/actions/runs/18909962579
```

lucasccordeiro (approved):

lucasccordeiro:

PR #3004, opened 2025-10-31
[python] Fix Array Type Dereference Error in Python List Iteration and Class Member Assignment

### LINKED ISSUE #2999: [Python] rvalue reference error with string field
For this program:
```python
class Foo:
    def __init__(self) -> None:
        self.x = "foo"

    def foo(self) -> int:
        return 42


f: Foo = Foo()
f.foo()
```
ESBMC errors out:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.547s

Aborted (core dumped)
```
ESBMC version: 521c862b0

LukeW1999:
Fixes https://github.com/esbmc/esbmc/issues/2999.

## Problem
String iteration like `for s in ["foo", "bar"]` was trying to dereference `char[4]*` directly, causing crashes. Similar issue when assigning strings to class members.

## Changes
Fixed array-to-pointer decay in two places:
- `python_list.cpp`: Convert `char[N]*` → `char*` when accessing list elements
- `python_converter.cpp`: Keep array types for class members, apply pointer decay for local variables

This matches Python's string semantics where strings are immutable references.

## Tests
Added `github_2999`, `github_2999_2`, `github_2999_3` with corresponding failure cases.

### DIFF
--- a/regression/python/github_2999/main.py
+++ b/regression/python/github_2999/main.py
@@ -1,0 +1,3 @@
+l: list[str] = ["foo", "bar", "baz"]
+for s in l:
+    pass
--- a/regression/python/github_2999/test.desc
+++ b/regression/python/github_2999/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--unwind 4
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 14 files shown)


lucasccordeiro (approved):

PR #3016, opened 2025-10-31
[python] implement str.join() method with string concatenation

### LINKED ISSUE #3000: [Python] Crash with join operation
Program:
```python
l: list[str] = ["foo", "bar", "baz"]
s = " ".join(l)
assert s == "foo bar baz"
```
ESBMC output:
```
$ esbmc ex.py 
Parsing ex.py
  what():  [json.exception.type_error.302] type must be string, but is number
Aborted (core dumped)
```

LukeW1999:
Fixes https://github.com/esbmc/esbmc/issues/3000.

## In short
Implements support for Python's `str.join()` method in the ESBMC Python frontend.

## Implementation
- Added `handle_str_join()` method in `python_converter.cpp`
- Fixed crash when calling methods on string literals in `python_annotation.h`
- Supports variable references, empty lists, single elements, and multi-character separators

## Test Coverage
Added 17 regression tests covering:
- Basic join operations 
- Edge cases (empty list, single element, empty separator)
- Type error test (`github_3000_8_fail`)

## Known Issues

**Limited support for direct list literals:**
`github_3000_1` is marked as `KNOWNBUG` - currently optimized for variable references. It should be able to set string as default and do a check in future since join's nature of string elements.

**Counterexample readability:**
When verification fails, the output shows low-level C symbols instead of Python-level information. For example, `github_3000_7_fail` outputs:
```
Violated property: return_value$_strcmp$1 == 0
```
Instead of showing the actual vs expected string values.

### DIFF
--- a/src/python-frontend/python_annotation.h
+++ b/src/python-frontend/python_annotation.h
@@ -1139,6 +1139,25 @@ private:
     return result;
   }
 
+  // Helper: Get type name from a constant value (e.g., "str" for string literal)
+  std::string get_type_from_constant(const Json &constant_node) const
+  {
+    if (!constant_node.contains("value"))
+      return "";
+    
+    const auto &value = constant_node["value"];
+    if (value.is_string())
+      return "str";
+    else if (value.is_number_integer())
+      return "int";
+    else if (value.is_number_float())
+      return "float";
+    else if (value.is_boolean())
+      return "bool";
+    
+    return "";
+  }
+
   std::string get_object_name(const Json &call, const std::string &prefix)
   {
     if (call["value"]["_type"] == "Attribute")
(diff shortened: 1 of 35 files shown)


lucasccordeiro (approved):

src/python-frontend/python_annotation.h
```diff
@@ -1139,6 +1139,25 @@ class python_annotation
...
+    else if (value.is_number_integer())
+      return "int";
+    else if (value.is_number_float())
+      return "float";
+    else if (value.is_boolean())
+      return "bool";
+
+    return "";
+  }
+
```
brcfarias, line 1160: @LukeW1999,
The `get_type_from_constant` function already existed in this file. It's duplicated now:
https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_annotation.h#L589-L607
https://github.com/esbmc/esbmc/blob/master/src/python-frontend/python_annotation.h#L1143-L1159
Could you please remove one of them?

LukeW1999: Hi, @brcfarias 
Sure. I would fix it when I went back to home. 
I am really sorry for my mistake. Thanks a lot for telling me.

brcfarias: That’s fine. I don’t think it’s urgent :-)

PR #3047, opened 2025-11-04
[python] Add Support for Python Class Forward References

### LINKED ISSUE #2997: [Python] Feature request: support forward references
Python allows referring to types that are defined later through wrapping them in single quotes, e.g.:
```python
class Foo:
    def __init__(self, x: int) -> None:
        self.x = x

    def bar(self) -> 'Bar':
        return Bar(self)

class Bar:
    def __init__(self, f: Foo) -> None:
        self.x = f.x

f: Foo = Foo(5)
b: Bar = f.bar()
assert b.x == 5
```
In the above program, `Bar` is defined after `Foo`, so it needs to be wrapped in single quotes in the return type of the `bar` method. This is typically required when there is a cyclic dependency between classes.

LukeW1999:
## Summary
Implements support for Python class forward references in type annotations, fixing issue #2997. Enables handling of both string-based forward references (`'Bar'`) and direct name references in method return types and parameter annotations.

## Problem
ESBMC couldn't handle cases where a class references another class defined later:

```python
class Foo:
    def bar(self) -> 'Bar':  # Forward reference 
        return Bar(self)

class Bar:  
    def __init__(self, f: Foo) -> None:
        self.x = f.x
```

This format is still not supported, Please use in this format:
```python
class Foo:
    def bar(self) -> 'Bar':  
        return Bar(self)

class Bar:  
    def __init__(self, f: Foo) -> None:
        self.x : int = f.x # <- please give the type like this 
```

## Solution
Implemented a three-phase approach:

1. **Pre-scan attributes**: Collect class attributes from method bodies before processing references
2. **Resolve forward references**: Scan return types and parameter annotations, process referenced classes immediately
3. **Enhanced type compatibility**: Allow pointer types to match struct parameters (Python object semantics)
4. **Auto-generate constructors**: Create default `__init__` when missing (respects inheritance)

## Changes
- `python_converter.cpp/h`: Forward reference resolution logic
- `function_call_expr.cpp`: Enhanced argument type checking
- `symex_function.cpp`: Pointer-to-struct type matching
- `type_handler.cpp`: Forward reference detection

## Testing
Added 18 test cases covering:
- Basic forward references with constructors
- Multi-level class chains (A → B → C)
- Classes without explicit `__init__`
- Attribute copying patterns
- Both success and failure cases

All tests verify correct value propagation and type safety.

---

**Closes #2997**


src/python-frontend/python_converter.cpp
```diff
@@ -4793,6 +4794,55 @@ void python_converter::get_attributes_from_self(
...
+
+  // Find and process the referenced class definition
+  const auto &ref_class_node =
+    find_class((*ast_json)["body"], referenced_class);
+
+  if (!ref_class_node.empty())
+  {
+    // Save and restore current class context
+    std::string saved_class = current_class_name_;
+    get_class_definition(ref_class_node, target_block);
```
XLiZHI, line 4841: A proper way to handle forward references doesn't need to get the class definition in advance, we just need a type (which can be symbolic).

```
symbol
  type: tag-A
```

When we come across a class/struct that has not been resolved yet, we can create a symbol for it to mark it as incomplete. Finally, the symbol type is returned.

Because when transforming the AST sequentially, we will come across this class/struct later on, at which point we can complete the symbol.

Something like this:

```cpp
get_class(...)
{
    symbol s = find(id);
     if (!s)
     {
       s = create_imcomplete_sym();
       move to context;
       return s;
     }

     if (sym.is_imcomlete)
          complete it;

     return s;
}
```

This saves us from repeatedly parsing a class definition and also handles circular dependencies.

XLiZHI (changes requested):

PR #3049, opened 2025-11-04
[python] add __init__ to class construction

LukeW1999:
See #3047.

This is PR suggested to split the the implementation of the #3047 to make it more clear.

TODO: 
Should not get the class definition in advance, we just need a type (which can be symbolic).
```
symbol
  type: tag-A
```


LukeW1999:
This PR adds automatic default constructor generation for Python classes using C++-style incomplete type handling.

**Key improvement:** Classes without explicit `__init__` can now be instantiated correctly.

**Example now supported:**
```python
class Bar:
    pass

class Foo:
    def bar(self) -> Bar:  
        return Bar()  # Works now - default __init__ auto-generated

f = Foo()
b = f.bar()
```

Previously, `Bar()` would fail because master doesn't generate default constructors.
## TODO
[ ] Need to suppor the forward reference for -> 'Bar' format

PR #3051, opened 2025-11-04
[SVCOMP] update script for SVCOMP26 && fix correctness witness

XLiZHI:
To further fix witness, we still need self host runner to update JAVA to 16 or above, to support CPAchecker.

Do we need the pass --CI flag? So far I can't find witness files in `esbmc-output`, but I remember there used to be (two years ago?). Am I missing something?

### DIFF
--- a/.github/workflows/benchexec.yml
+++ b/.github/workflows/benchexec.yml
@@ -29,13 +29,13 @@ on:
         type: choice
         description: Run-set to use (be sure to set mode to "RunSet")
         options:
-        - SV-COMP25_unreach-call
-        - SV-COMP25_no-data-race
-        - SV-COMP25_valid-memcleanup
-        - SV-COMP25_valid-memsafety
-        - SV-COMP25_termination
-        - SV-COMP25_no-overflow
-        - SV-COMP25_ConcurrencySafety
+        - SV-COMP26_unreach-call
+        - SV-COMP26_no-data-race
+        - SV-COMP26_valid-memcleanup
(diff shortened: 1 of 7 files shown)


rafaelsamenezes:
> To further fix witness, we still need self host runner to update JAVA to 16 or above, to support CPAchecker.
> 
> Do we need the pass --CI flag? So far I can't find witness files in `esbmc-output`, but I remember there used to be (two years ago?). Am I missing something?

It's been a while since I configured this. It should be fine to install a more recent java.  The current hosts are a bit bugged though:

1. In the past we relied on Ubuntu 20.04 and 22.04 which used cgroups v1.
2. Recently (as in last month), I updated to Ubuntu 24.04, which uses cgroups v2. Benchexec does have support to it, which required this change: https://github.com/esbmc/esbmc/commit/c1c19cc135a99169e51a05aae6aad30c62c0efa7
3. I still had some issues with it, so I changed the machine back to cgroups v1. And did not revert the change.

It might be due to this... although the best solution is probably to learn cgroups and fix the issues.

rafaelsamenezes:
Ah, extra. It might be due to this: https://github.com/esbmc/esbmc/commit/d3cc51415469a2aa4e169733cbcc12027aa90d1a

The result-files is not for witnesses.

XLiZHI:
Many thanks! @rafaelsamenezes

rafaelsamenezes (approved):

lucasccordeiro (approved):

PR #3069, opened 2025-11-05
[python] Add Class forward reference

LukeW1999:
This PR adds syntax support for Python class forward references, allowing ESBMC to parse type hints like `def foo() -> 'Bar':`.

This is the second PR derived from #3047, following #3049 which added `__init__` generation.

## Changes

- Added `process_forward_reference()` to handle string-form forward references in type annotations
- Implemented two-phase class definition (incomplete → complete) to prevent infinite recursion
- Added `find_class()` helper to locate class definitions in AST

## Limitations

⚠️ This PR provides **syntax support only**. Full semantic verification is not yet implemented. Test cases marked as `KNOWNBUG` require future work.

## Related

- Part of #2997
- Depends on #3049
- Derived from #3047


src/python-frontend/python_converter.cpp
```diff
@@ -4419,6 +4419,74 @@ void python_converter::get_attributes_from_self(
   }
 }
 
+// Find class definition
+const nlohmann::json &python_converter::find_class(
```
XLiZHI: https://github.com/esbmc/esbmc/blob/262b19567712cb436d571cef7e5bbea5c9a2262b/src/python-frontend/json_utils.h#L46-L56

src/python-frontend/python_converter.cpp
```diff
@@ -4521,7 +4643,63 @@ void python_converter::get_class_definition(
...
+  }
+
+  // Only generate default constructor like __init__(self) -> None if:
+  // 1. Class has no __init__ method AND
+  // 2. Class has no user-defined base class (to avoid overriding inherited constructor)
+  if (!has_init && !has_user_defined_base)
+  {
+    // Generate a default constructor: __init__(self) -> None
+    code_typet function_type;
+    function_type.return_type() = none_type();
```
XLiZHI: The constructor's return type should contain "constructor" label.

src/python-frontend/python_converter.cpp
```diff
@@ -4475,8 +4564,41 @@ void python_converter::get_class_definition(
     if (class_member["_type"] == "FunctionDef")
     {
       get_attributes_from_self(class_member["body"], clazz);
-      added_symbol->type = clazz;
+    }
+  }
+  added_symbol->type = clazz;
 
+  // Pre-scan for forward references
+  for (auto &class_member : class_node["body"])
```
XLiZHI, line 4553: We don't need any pre-scan, try to use the symbolic type directly because it doesn't need any definition.

As far as I know, Python doesn't parse type annotations ahead of time to check for constructor calls inside function bodies.

@brcfarias Do you have any thoughts?

XLiZHI (changes requested):

src/python-frontend/python_converter.cpp
brcfarias: Hi @XLiZHI, I’ll take a look at the change and get back to you, likely tomorrow.

LukeW1999: Many thanks to @XLiZHI for feedback and @brcfarias for your kind help.

I really agree with your point - the pre-scan does look somewhat redundant and could potentially break the existing method passing and annotation handling mechanisms.

However, I'm having some difficulty understanding how to implement this with symbolic types. The current issue I'm facing is that the delayed resolution mechanism doesn't include classes. In the example:
```python
class Foo:
    def bar(self) -> 'Bar':
        return Bar(self)  # <-- at this point

class Bar:
    def __init__(self, f: Foo) -> None:
        self.x: int = f.x
```
When we reach the Bar(self) call inside Foo.bar(), the code tries to access the symbol 'tag-Bar' to determine if it's a constructor call, but Bar hasn't been constructed in the symbol table yet. I maybe need some help to fix the implementation on the incomplete class annotation.


Additionally, I'd like to mention that the pre-scan approach might actually provide better functionality for handling transitive forward references. The current PR successfully handles chained object passing scenarios like: in #3070 
```python
class A:
    def __init__(self, val: int) -> None:
        self.val: int = val

    def create_b(self) -> 'B':
        return B(self)

class B:
    def __init__(self, a: A) -> None:
        self.val: int = a.val

    def create_c(self) -> 'C':
        return C(self)

class C:
    def __init__(self, b: B) -> None:
        self.val: int = b.val

a = A(7)
b = a.create_b()
c = b.create_c()
assert c.val == 7
```
I'm not entirely sure if the previous annotation approach could handle these multi-level dependencies correctly. Would switching to symbolic types risk breaking this functionality, or am I misunderstanding how symbolic types would work in this context?

Would appreciate your guidance on the best path forward here!

src/python-frontend/python_converter.cpp
LukeW1999: Many thanks.

src/python-frontend/python_converter.cpp
LukeW1999: Hi, @XLiZHI 
I am sorry to say I cannot remove prescan before. Bruno teached me how to implement it. I really apperciate your guidance and kind help.
I have directly changed on this #3070 I think now the logic is more clear in that PR.

PR #3070, opened 2025-11-05
[python] Add support for passing objects with forward-referenced types

LukeW1999:
This PR implements pointer-to-struct semantics for passing Python objects as function parameters, enabling ESBMC to verify code like `Bar(foo: Foo)` where objects are passed by reference.

This is the third PR derived from #3047, following #3049 (incomplete types) and #3069 (forward references).

## Changes

- **AST Layer**: Allow parameter objects to access instance members (`python_converter.cpp`)
  - Added nested attribute chain handling (e.g., `self.b.a`)
  - Auto-dereference pointers in `create_member_expression()`

- **Type Checking**: Accept `Foo*` for `Foo` parameter type (`function_call_expr.cpp`)
  - Resolve symbol types for proper compatibility checking

- **Symex Layer**: Auto-dereference pointers during parameter assignment (`symex_function.cpp`)
  - Verify type compatibility at IR level

- **Tests**: Added comprehensive test with 18 scenarios (460 lines)
  - Boundary values (zero, negative numbers)
  - Deep nesting (5-level chains)
  - Same-type object passing
  - Diamond-shaped dependencies
  - State accumulation across transformations

## Test Results

- Enabled: `github_2997_5`, `github_2997_6`, `github_2997_7`
- New: `object_passing_comprehensive` 

## Related

- Part of #2997
- Depends on #3049, #3069
- Derived from #3047


src/python-frontend/python_converter.cpp
```diff
@@ -4496,8 +4546,41 @@ void python_converter::get_class_definition(
...
+  added_symbol->type = clazz;
+
+  // Pre-scan for forward references
+  for (auto &class_member : class_node["body"])
+  {
+    if (class_member["_type"] != "FunctionDef")
+      continue;
+
+    // Scan return type annotations for forward references
+    if (!class_member["returns"].is_null())
```
brcfarias: At this point, you can check whether it is a constructor call for a class that is not yet in the symbol table, and use that as the condition for loading it.

brcfarias (changes requested):

src/python-frontend/python_converter.cpp
LukeW1999: Hi @brcfarias,
I moved the call point out of the get_class to the function_call and get_var_assign. It should be more clear now. And, this change should be able to make the current forward reference function be called more lazy.

I really apperciate your kind help and guidance.

lucasccordeiro (approved):

lucasccordeiro:

PR #3100, opened 2025-11-09
[python] fix type inference for @overload functions with Literal arguments

### LINKED ISSUE #3091: [Python] Incorrect `AttributeError` with function overload
For the following program:
```python
from typing import Any, Literal, overload

class Foo:
    def __init__(self) -> None:
        pass

    def foo(self) -> None:
        pass
    

class Bar:
    def __init__(self) -> None:
        pass

    def bar(self) -> None:
        pass

@overload
def create(s: Literal["foo"], t: str) -> Foo:
    ...

@overload
def create(s: Literal["bar"], t: str) -> Bar:
    ...
    
def create(s: str, t: str) -> Foo | Bar:
    if s == 'foo':
        return Foo()
    elif s == 'bar':
        return Bar()
    else:
        raise NotImplementedError("Unknown class")

b = create("bar", t='t')
b.bar()
```
ESBMC reports an attribute error on the call to `bar`, even though the `bar` method exists in class `Bar`:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
WARNING: Union with no recognized types, defaulting to pointer
WARNING: AttributeError: 'Foo' object has no attribute 'bar'
Generating GOTO Program
GOTO program creation time: 0.578s
Unwinding loop 12 iteration 1   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 12 iteration 2   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 12 iteration 3   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Symex completed in: 0.002s (59 assignments)
Slicing time: 0.000s (removed 58 assignments)
Generated 21 VCC(s), 1 remaining after simplification (1 assignments)
Building error trace

[Counterexample]


State 1 file ex.py line 38 column 0 thread 0
Violated property:
  file ex.py line 38 column 0
  AttributeError: 'Foo' object has no attribute 'bar'
  0


VERIFICATION FAILED
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3091.

This PR adds support for resolving `@overload` decorated functions by matching literal string arguments to their corresponding Literal type hints. This allows ESBMC-Python to infer the correct return type from overloaded function calls instead of defaulting to the first type in a union, preventing false AttributeError warnings. For example, `create("bar")` now correctly infers `Bar` instead of `Foo|Bar`.

Thanks to [zhassan-aws](https://github.com/zhassan-aws) for reporting this issue.

### DIFF
--- a/regression/python/github_3091/main.py
+++ b/regression/python/github_3091/main.py
@@ -1,0 +1,36 @@
+from typing import Any, Literal, overload
+
+class Foo:
+    def __init__(self) -> None:
+        pass
+
+    def foo(self) -> None:
+        pass
+    
+
+class Bar:
+    def __init__(self) -> None:
+        pass
+
+    def bar(self) -> None:
+        pass
+
+@overload
+def create(s: Literal["foo"], t: str) -> Foo:
+    ...
+
+@overload
+def create(s: Literal["bar"], t: str) -> Bar:
+    ...
+    
+def create(s: str, t: str) -> Foo | Bar:
+    if s == 'foo':
+        return Foo()
+    elif s == 'bar':
+        return Bar()
+    else:
+        raise NotImplementedError("Unknown class")
+
+b = create("bar", t='t')
+b.bar()
+assert isinstance(b, Bar)
--- a/regression/python/github_3091/test.desc
+++ b/regression/python/github_3091/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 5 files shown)


zhassan-aws:
Thanks for adding this @lucasccordeiro!

lucasccordeiro:
You're welcome, @zhassan-aws  ;-)

PR #3104, opened 2025-11-10
[python] enforce strict type checking for keyword arguments

### LINKED ISSUE #3093: [Python] Incorrect argument type not reported across different modules
Thanks for adding the `--strict-types` flag in https://github.com/esbmc/esbmc/pull/3025.

I observed that the flag doesn't catch an incorrect argument type when the function is in a different file than the call, e.g:

`ex.py`:
```python
import ll

ll.foo(a=5)
```
`ll.py`:
```python
def foo(a: str) -> None:
    pass
```
ESBMC output:
```bash
$ esbmc ex.py --strict-types
Parsing ex.py

Type checking warning:
ex.py:3: error: Argument "a" to "foo" has incompatible type "int"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 1 source file)

Converting
Generating GOTO Program
GOTO program creation time: 0.579s
Symex completed in: 0.000s (2 assignments)
Slicing time: 0.000s (removed 2 assignments)
Generated 0 VCC(s), 0 remaining after simplification (0 assignments)
BMC program time: 0.000s

VERIFICATION SUCCESSFUL
```
Is it possible to extend the strict type checking across modules?

ESBMC version: 58e643b24

LukeW1999:
Fixes https://github.com/esbmc/esbmc/issues/3093.

**Summary**

- **Problem**: With `--strict-types` enabled, ESBMC only validates positional arguments. Keyword arguments that violate annotations merely trigger a warning, so cases like `github_3093` pass unexpectedly.

- **Solution**: Extend `check_argument_types()` to receive the keyword list and reuse the same type-matching logic for both positional and keyword arguments. The helper now emits a `TypeError` whenever a keyword argument violates the expected type.

**Tasks**
- Added three test for this bug.
- Reuse the idea of the exception capture and return.


lucasccordeiro (approved):

lucasccordeiro:

PR #3107, opened 2025-11-11
[python] fix string literal type annotation for str returns

LukeW1999:
### Problem
- `-> "str"` (with quotes) returns **array type** ❌
- `-> str` (no quotes) returns **pointer type** ✅
- Type mismatch causes verification to fail

```python
def str_annotation() -> "str":  # ← String literal annotation
    return "hello"

result = str_annotation()
assert result == "hello"
```

### Fix
add a handling:

```cpp
if (type_string == "str")
  type.return_type() = gen_pointer_type(char_type());
else
  type.return_type() = type_handler_.get_typet(type_string);
```

### Result
Both annotation styles now return pointer type consistently.

**Test**: `return10_2` changes from `KNOWNBUG` → `CORE` ✅

### DIFF
--- a/regression/python/return10_2/test.desc
+++ b/regression/python/return10_2/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --incremental-bmc
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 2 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #3115, opened 2025-11-11
[python] Add required argument validation for cross-module function calls

### LINKED ISSUE #3010: [Python] Missing arguments not flagged when using keyword arguments
For this program:
```python
def foo(w: int, x: int, y: int = 3) -> int:
    return x + y

foo(3)
```
ESBMC correctly flags the missing required arguments:
```bash
$ esbmc ex.py 
Parsing ex.py

Type checking warning:
ex.py:4: error: Missing positional argument "x" in call to "foo"  [call-arg]
Found 1 error in 1 file (checked 1 source file)

WARNING: foo() missing required positional argument: 'x'

* file: ex.py
* line 4
* function: foo
* column: 0 
Converting
Generating GOTO Program
GOTO program creation time: 0.546s
Symex completed in: 0.000s (3 assignments)
Slicing time: 0.000s (removed 2 assignments)
Building error trace

[Counterexample]


State 1  thread 0
Violated property:


VERIFICATION FAILED
```
However, if we use keyword arguments instead:
```python
def foo(w: int, x: int, y: int = 3) -> int:
    return x + y

foo(w=3)
```
verification unexpectedly succeeds:
```bash
$ esbmc ex.py 
Parsing ex.py

Type checking warning:
ex.py:4: error: Missing positional argument "x" in call to "foo"  [call-arg]
Found 1 error in 1 file (checked 1 source file)

WARNING: foo() missing required positional argument: 'x'

* file: ex.py
* line 4
* function: foo
* column: 0 
Converting
Generating GOTO Program
GOTO program creation time: 0.541s
Symex completed in: 0.000s (4 assignments)
Slicing time: 0.000s (removed 4 assignments)
Generated 0 VCC(s), 0 remaining after simplification (0 assignments)
BMC program time: 0.000s

VERIFICATION SUCCESSFUL
```

LukeW1999:
Fixes #3010.

## Description

This PR fixes missing required argument detection for cross-module function calls in the Python frontend.

### Problem

When a function is defined in an imported module (e.g., `ll.py`) and called from the main file, ESBMC was not detecting missing required arguments. The function signature was present in the symbol table, but the `handle_keywords` lambda was unconditionally filling missing arguments with `None` without checking if they had default values.

### Example

**ex.py:**
```python
import ll

f = ll.Foo()
f.foo(b="b")  # Missing required argument 'a'
```

**ll.py:**
```python
class Foo:
    def __init__(self) -> None:
        pass

    def foo(self, a: str, b: str) -> None:
        pass
```

Previously, ESBMC would not detect that the required argument `a` is missing from the call to `f.foo()`.

---

## Changes
Refactored argument processing to properly track and store default values:

- Extracted parameter registration logic into `register_function_argument`, which returns the parameter's index in the argument list
- Modified `process_function_arguments` to collect these indices and associate default values from the AST's defaults and kw_defaults arrays with the correct parameters
- Added validation in `get_function_call` to check if missing arguments have default values or `Optional` types, throwing TypeError for missing required arguments


lucasccordeiro:

PR #3124, opened 2025-11-12
[python] Prioritize local modules over system modules

### LINKED ISSUE #3116: [Python] Feature request: load local module
The Python interpreter seems to prioritize local modules over system ones by default.

For instance, given the following program which uses the [toml library](https://docs.python.org/3/library/tomllib.html):
```python
import toml

# a dummy toml string
toml_string = """
[servers]

[servers.alpha]
ip = "10.0.0.1"
role = "frontend"
"""

# parse the toml string
parsed_toml = toml.loads(toml_string)
print(parsed_toml)
assert parsed_toml == "foo"
```
the assertion fails as expected:
```bash
$ python ex.py 
{'servers': {'alpha': {'ip': '10.0.0.1', 'role': 'frontend'}}}
  File "/home/ubuntu/examples/python/pygo/exv10/ex.py", line 15, in <module>
    assert parsed_toml == "foo"
AssertionError
```
 However, if I place a file called `toml.py` in the same directory with the following contents:
```python
def loads(s: str) -> str:
    return "foo"
```
the assertion passes:
```bash
$ python ex.py 
foo
```
It would be nice if ESBMC matches this behavior of loading the local module if present. Currently, it would error out:
```bash
$ esbmc ex.py 
Parsing ex.py
ERROR: Object "toml" not found.
```
but if I uninstall the system toml package and try again, verification succeeds:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.430s
Unwinding loop 17 iteration 1   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 17 iteration 2   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 17 iteration 3   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Symex completed in: 0.001s (43 assignments)
Slicing time: 0.001s (removed 43 assignments)
Generated 17 VCC(s), 0 remaining after simplification (0 assignments)
BMC program time: 0.002s

VERIFICATION SUCCESSFUL
```
This feature would allow us to add models of Python libraries as needed.

LukeW1999:
## Description  
Implements local module priority for Python frontend to match standard Python interpreter behavior.  
  
## Fixes  
Closes #3116  
  
## Testing  
- Verified local `toml.py` is loaded before system `toml` package  
- Still would fallback to system modules when local module doesn't exist  
- All existing Python regression tests pass
- Would not affect math, os, numpy import

### DIFF
--- a/regression/python/github_3116/main.py
+++ b/regression/python/github_3116/main.py
@@ -1,0 +1,16 @@
+import toml  # type: ignore[import]
+
+
+def main() -> None:
+    toml_string = """
+[servers]
+[servers.alpha]
+ip = "10.0.0.1"
+role = "frontend"
+"""
+
+    parsed_toml = toml.loads(toml_string)
+    assert parsed_toml == "foo"
+
+
+main()
--- a/regression/python/github_3116/test.desc
+++ b/regression/python/github_3116/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_3116/toml.py
+++ b/regression/python/github_3116/toml.py
@@ -1,0 +1,2 @@
+def loads(_: str) -> str:
+    return "foo"
--- a/regression/python/github_3116_1/main.py
+++ b/regression/python/github_3116_1/main.py
@@ -1,0 +1,16 @@
+import toml as local_toml 
+
+
+def main() -> None:
+    data = """
+[servers]
+[servers.alpha]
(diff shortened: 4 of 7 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #3164, opened 2025-11-19
[python] Fix type inference for nested attribute access in Python frontend

### LINKED ISSUE #3151: [Python] Object not found error when referring to attribute
For the following program:
```python
class Foo:
    def __init__(self, name: str) -> None:
        self.name = name

    def foo(self) -> str:
        return self.name

class Bar:
    def __init__(self) -> None:
        self.f: Foo = Foo("bar")

    def bar(self) -> str:
        s = self.f.foo()
        return s

b = Bar()
assert b.bar() == "bar"
```
ESBMC errors out:
```bash
$ esbmc ex.py 
Parsing ex.py
ERROR: Object "self.f" not found.
```
ESBMC version: 71b5cce0f

LukeW1999:
Fixes #3151.

## Description

This PR addresses a limitation in the Python frontend where type inference failed for nested attribute access (e.g., `self.b.a.get_value()`), causing valid code to fail verification.

Additionally, it fixes a scoping issue with Annotated Assignments (`AnnAssign`) where local variables were not explicitly declared, causing the pointer analysis to incorrectly classify them as globals.

### 1. Type Inference Logic (`python_annotation.h`)
- Implemented a recursive `extract_attr_chain` function to resolve chains of attributes (e.g., `self.b.a`).
- The logic now identifies the base object (currently restricted to `self`), resolves the type of each attribute via `__init__`, and traverses the class definitions to determine the final method's return type.

### 2. Local Scope Declaration (`python_converter.cpp`)
- Added `code_declt` generation for `AnnAssign` nodes.
- **Reason:** Without this explicit declaration, the symbol table treated these local variables as implicit globals (level 2), leading to false positives in pointer analysis checks.

## Test Coverage

A new regression suite `regression/python/nested-attr-*` has been added:
- **Multi-level Nesting:** Verified 2, 3, and 4 levels of nesting (e.g., `nested-attr-1`, `nested-attr-6`).
- **Negative Testing:** Added `_fail` cases (e.g., `github_3151_fail`) to ensure the solver correctly reports `VERIFICATION FAILED` when nested values do not match assertions.
- **Inheritance & Shadowing:** Verified cases where attributes are inherited or shadowed by local variables.

## Known Limitations (Marked as KNOWNBUG)

Several test cases are marked as `KNOWNBUG` in `test.desc`. These highlight the boundaries of the current fix:

1.  **Global/Module Scope (`nested-attr-14`, `nested-attr-15`):**
    * **Issue:** The current fix explicitly checks for `base_obj == "self"` and requires a valid `current_func` context.
    * **Symptom:** Nested attributes accessed at the module level (e.g., `l1.l2.val` or `b.f.foo()`) are ignored by this logic.

2.  **Intermediate Variables / Aliasing (`nested-attr-13`):**
    * **Issue:** The fix relies on explicit attribute chains in the AST (e.g., `self.f.foo()`). It does not perform data flow analysis to track aliases.
    * **Symptom:** Breaking the chain into variables (e.g., `temp = self.f; temp.foo()`) fails because the base object is `temp`, not `self`.

3.  **Method Chaining (`nested-attr-7`):**
    * **Issue:** `extract_attr_chain` halts on `Call` nodes.
    * **Symptom:** Chains with intermediate method calls (e.g., `self.builder.build().val`) result in an `Object "" not found` error.

4.  **Complex Expressions (`nested-attr-8`, `nested-attr-9`):**
    * Mixed-type attributes and nested calls within complex arithmetic expressions are marked as known bugs.


lucasccordeiro (approved):

lucasccordeiro:

PR #3169, opened 2025-11-19
Always use c parser to process bundles c libaries (solves #3088)

### LINKED ISSUE #3088: Python frontend parsing failure
Weird issue in the Python frontend where it appears to be trying to parse a c file using the Python parser, which goes about as well as you would expect. Any idea what might be causing this or where I should look to try and debug this?

```
Parsing test.py

Type checking warning:
test.py:11: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
test.py:12: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
Found 2 errors in 1 file (checked 1 source file)

<unknown>:33: SyntaxWarning: invalid escape sequence '\d'
Converting
file /tmp/esbmc.4fd6-b6a5-ff4e/library/builtin_libs.c: Parsing

Type checking warning:
/tmp/esbmc.4fd6-b6a5-ff4e/library/builtin_libs.c:10: error: Unterminated string literal (detected at line 10)  [syntax]
Found 1 error in 1 file (errors prevented further checking)

  File "/tmp/esbmc-python-astgen-7bac-d588-8992/parser.py", line 504, in <module>
    main()
  File "/tmp/esbmc-python-astgen-7bac-d588-8992/parser.py", line 460, in main
    tree = ast.parse(source.read())
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sgtpl/.pyenv/versions/3.12.11/lib/python3.12/ast.py", line 52, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<unknown>", line 10
    /* Disable cheri-compressed-cap's debug assertions since they assert that
                                   ^
SyntaxError: unterminated string literal (detected at line 10)
```

Test.py is just the example in the README.md:
```
import random as rand


def div1(cond: int, x: int) -> int:
    if not cond:
        return 42 // x
    else:
        return x // 10


cond: int = rand.random()
x: int = rand.random()

assert div1(cond, x) != 1
```

sgpthomas:
Fixes #3088 by always using the C parser when processing bundled C libraries.

### DIFF
--- a/src/c2goto/cprover_library.cpp
+++ b/src/c2goto/cprover_library.cpp
@@ -282,7 +282,11 @@ void add_cprover_library(contextt &context, const languaget *language)
   if (clib->size == 0)
   {
     if (language)
-      return add_bundled_library_sources(context, *language);
+    {
+      // C library sources must be parsed with C frontend, not the current language
+      std::unique_ptr<languaget> c_lang(new_language(language_idt::C));
+      return add_bundled_library_sources(context, *c_lang);
+    }
     log_error("Zero-lengthed internal C library");
     abort();
   }


lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @sgpthomas.

PR #3190, opened 2025-11-24
[Python verfication] fix crash with a custom exception

### LINKED ISSUE #3181: [Python] Crash with a custom exception
For this program:
```python
class MyException(BaseException):
    def __init__(self, msg: str) -> None:
        self.msg = msg

def foo(x: int) -> None:
    if x == 42:
        raise MyException("bad number!")

import random

r = random.randint(41, 43)
try:
    foo(r)
except MyException as e:
    print("caught MyException exception")
```
ESBMC crashes:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.630s
ERROR: Exception thrown of type MyException at file ex.py line 7
```
ESBMC version: 339aa022a

XLiZHI:
fixes #3181

### DIFF
--- a/regression/python/github_3181/main.py
+++ b/regression/python/github_3181/main.py
@@ -1,0 +1,15 @@
+class MyException(BaseException):
+    def __init__(self, msg: str) -> None:
+        self.msg = msg
+
+def foo(x: int) -> None:
+    if x == 42:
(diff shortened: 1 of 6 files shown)


lucasccordeiro:

PR #3192, opened 2025-11-24
[feat] Add C11 Atomicity

### LINKED ISSUE #3191: [Bug] Cannot handle C11 _Atomic types
## Summary

ESBMC 7.11.0 (latest build) fails to convert C11 atomic types with error:

```
ERROR: Conversion of unsupported clang cast operator: "NonAtomicToAtomic"
```

This prevents verification of real-world multi-threaded code that uses standard C11 atomics.

## Minimal Reproducible Example

```c
#include <stdatomic.h>
#include <stdint.h>
#include <assert.h>

void test_minimal(void) {
    // Simplest possible case - just initialization
    _Atomic(int) x = 42;  // ERROR: NonAtomicToAtomic
    assert(x == 42);
}

int main(void) {
    test_minimal();
    return 0;
}
```

## Full Output

```
./esbmc test.c
[PROGRESS] Parsing test.c
[PROGRESS] Converting
[ERROR] Conversion of unsupported clang cast operator: "NonAtomicToAtomic" to expression
ImplicitCastExpr 0x97f8000 <test.c:7:22> '_Atomic(int)' <NonAtomicToAtomic>
`-IntegerLiteral 0x97f7fe0 <col:22> 'int' 42

[ERROR] CONVERSION ERROR
```

## Expected Behavior

ESBMC should support C11 atomic types (`_Atomic` qualifier) since they are part of the C11 standard and widely used in multi-threaded code.

The conversion should handle the implicit cast from non-atomic to atomic types, which is valid in C11:

> An assignment operator shall have a modifiable lvalue as its left operand.
> For atomic types, this is a read-modify-write operation with memory_order_seq_cst memory order semantics.

## Actual Behavior

ESBMC fails during the conversion phase with an unsupported cast operator error.

## Related

This appears to be a limitation in ESBMC's Clang frontend integration, specifically in handling `ImplicitCastExpr` with `<NonAtomicToAtomic>` cast kind.

## Request

Could ESBMC add support for C11 atomic types? This would be a significant improvement for verifying concurrent systems code.

Alternative: If full atomic semantics are too complex, could ESBMC at least parse atomics and treat them as non-atomic for verification purposes (with a warning)? This would allow verification of the non-concurrent aspects of the code.

Yiannis128:
This PR addresses issue #3191 where we allow for implicit conversion of _Atomic() to it's base type:

```c
_Atomic(int) a = 1;
int b;
b = a; // Used to fail
```


Yiannis128:
The question now is how to treat atomic functions like `__c11_atomic_load`, should we fail or let them silently pass? Look at the test: regression/esbmc/github_3191_atomic_fail/main.c

This is meant to fail because we technically don't fully support those functions but it passes. What should I do?

@rafaelsamenezes 

This is the GCC reference of how atomics should be handled: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

regression/esbmc/github_3191_atomic_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
```
LukeW1999, line 1: You can change to KNOWBUG here. If you think it should fail.

rafaelsamenezes:
> The question now is how to treat atomic functions like `__c11_atomic_load`, should we fail or let them silently pass? Look at the test: regression/esbmc/github_3191_atomic_fail/main.c
> 
> 
> 
> This is meant to fail because we technically don't fully support those functions but it passes. What should I do?
> 
> 
> 
> @rafaelsamenezes 
> 
> 
> 
> This is the GCC reference of how atomics should be handled: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

Hm, in this case there is the I don't care about missing extensions flag which would be a fall through. I would suggest removing it only if you are going to do a proper support for atomicity.

It shouldn't be tricky but it will require lots of work. You would need to add this property into symbols and then each operation needs to be wrapped in an atomic block.

Yiannis128:
> > The question now is how to treat atomic functions like `__c11_atomic_load`, should we fail or let them silently pass? Look at the test: regression/esbmc/github_3191_atomic_fail/main.c
> > This is meant to fail because we technically don't fully support those functions but it passes. What should I do?
> > @rafaelsamenezes
> > This is the GCC reference of how atomics should be handled: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
> 
> Hm, in this case there is the I don't care about missing extensions flag which would be a fall through. I would suggest removing it only if you are going to do a proper support for atomicity.
> 
> It shouldn't be tricky but it will require lots of work. You would need to add this property into symbols and then each operation needs to be wrapped in an atomic block.

I'll look into this further and keep in touch when I need help. I'll try expanding this PR to handle atomicity.

Issue #3132, opened 2025-11-13
[python] extend the automated test case generation feature for Python programs


lucasccordeiro:
Currently, ESBMC cannot generate test cases for branch, decision, and assertion coverage for Python programs because the GOTO representation does not conform to the expected format.

This paper describes our automated test case generation algorithm: https://ssvlab.github.io/lucasccordeiro/papers/fase2025.pdf.

The task here is to examine how GOTO programs are structured in C and adjust the Python GOTO representation to match the same format.

PR #3199, opened 2025-11-25
[Python] Fix Python coverage support

prototypeC14:
## Summary
This PR is to fix Python coverage support by aligning GOTO representation with C/Solidity

## Issues Fixed
- see #3132 
## Problem Description
Python frontend could not generate test cases for branch, decision, and assertion coverage because the GOTO representation in coverage statistics. This caused:
Incorrect branch counts (e.g., 30 branches instead of 2)
Inconsistent behavior compared to C programs

## Solution
Implemented a three-function architecture similar to Solidity:
- python_init 
- Contains only user code from the main module
- Initializes static lifetime variables

## Testing
Verified that Python coverage behavior now matches C programs:
Test case(Python):
```python
def f(n: int) -> int:
    if n > 0:
        return 1
    else:
        return 0
f(5)
```
Result:
```C++
[Coverage]

Branches : 2
Reached : 1
Branch Coverage: 50%
```
Test case(Equivalent C):
```C
int f(int n) {
    if (n > 0)
        return 1;
    else
        return 0;
}

int main() {
    f(5);
    return 0;
}
```
```C++
[Coverage]

Branches : 2
Reached : 1
Branch Coverage: 50%
```


lucasccordeiro (approved):

PR #3202, opened 2025-11-26
[python-frontend] Implement runtime type checking via isinstance injection

### LINKED ISSUE #3178: [Python] Feature request: automatically inject `isinstance` assertions
Whenever there is an assignment to a type-annotated variable, whether in a regular assignment statement, e.g.:
```python
s: str = ...
```
or in a function call, e.g.:
```python
def foo(s: str, i: int, ...):
    ...

foo(...)
```
it would be useful if ESBMC automatically injects `isinstance` assertions to ensure the value is of the correct type. This would ensure that any type errors are caught and that the subsequent code can "safely" assume that the correct type is used.

sergillam:
Fixes https://github.com/esbmc/esbmc/issues/3178.


regression/python/type-annotation-assign-check/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+--overflow-check
```
lucasccordeiro: Why do we need to enable --overflow-check? I suggest removing this dependency. The isinstance assertion should detect the type mismatch.

src/python-frontend/python_converter.cpp
```diff
@@ -34,6 +35,7 @@
 #include <unordered_map>
 
 #include <boost/filesystem.hpp>
+#include <iostream>
```
lucasccordeiro: Can you remove this include?

src/python-frontend/python_converter.cpp
```diff
@@ -1424,6 +1426,188 @@ void python_converter::resolve_dict_subscript_types(
   }
 }
 
+std::vector<typet> python_converter::collect_annotation_types(
```
lucasccordeiro: Can you move all the new methods to a new class called python_typechecking?

lucasccordeiro (changes requested):
@sergillam: Can I ask you to:

1) Remove the dependency on --overflow-check. The isinstance assertion should detect the type mismatch.
2) Can you move the new methods to a new class called python_typechecking?
3) Can you create test cases that explore the incremental SMT solving (i.e., --smt-during-symex and --smt-symex-assert).
4) Can you fix the regression tests that are failing?

regression/python/type-annotation-assign-check/test.desc
sergillam: ok, the line is removed in all test

src/python-frontend/python_converter.cpp
sergillam: include removed, now use the log error macro instead of using the std::error function

regression/python/type-annotation-assign-check/test.desc
lucasccordeiro: Thank you.

src/python-frontend/python_converter.cpp
lucasccordeiro: Thank you.

src/python-frontend/python_converter.cpp
sergillam: The files moved in commit https://github.com/esbmc/esbmc/commit/f1f123f14b12bbcd0acd018cedaff745a7de007c

brcfarias (changes requested):
Hi @sergillam, I think that this is not a python_converter concern, and that it might be more appropriate to implement it either at the GOTO level (e.g., inside goto_check) or, preferably, right after annotating the program in python_languaget::parse.

lucasccordeiro:
Indeed, the `goto_check` would be the best option. Thanks for this suggestion, @brcfarias.

@sergillam: could you please improve your PR with this suggestion?

Many thanks,
Lucas

lucasccordeiro (approved):
Thanks for submitting this PR and addressing our comments, @sergillam.

src/python-frontend/python_typechecking.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+python_typechecking::python_typechecking(python_converter &converter)
+  : converter_(converter)
+{
+}
+
+std::vector<typet> python_typechecking::collect_annotation_types(
+  const nlohmann::json &annotation) const
+{
+  std::vector<typet> collected;
```
brcfarias, line 19: I think you could reuse the type_handler class here to obtain the types from the JSON.

src/python-frontend/python_typechecking.cpp
```diff
@@ -0,0 +1,438 @@
...
+  if (func_node["_type"] == "Attribute" && func_node.contains("attr"))
+    return func_node["attr"].get<std::string>();
+
+  return {};
+}
+
+bool python_typechecking::class_derives_from(
+  const std::string &class_name,
+  const std::string &expected_base) const
+{
```
brcfarias, line 395: Also think this logic should be moved into type_handler and then reused here. We plan to move this inheritance handling to a dedicated class/component, so it's better to keep everything within python_converter and type_handler for now.

brcfarias (changes requested):

src/goto-programs/goto_check.cpp
```diff
@@ -27,7 +30,8 @@ class goto_checkt
       enable_unsigned_overflow_check(
         options.get_bool_option("unsigned-overflow-check")),
       enable_ub_shift_check(options.get_bool_option("ub-shift-check")),
-      enable_nan_check(options.get_bool_option("nan-check"))
+      enable_nan_check(options.get_bool_option("nan-check")),
+      enable_python_type_check(options.get_bool_option("python-type-check"))
```
brcfarias: One last simple suggestion: could you rename it (including the function names) from python-type-check to is-instance-check?

brcfarias (changes requested):

src/python-frontend/python_typechecking.cpp
sergillam: I’ve updated python_typechecking::collect_annotation_types to reuse the existing type resolution pipeline as much as possible.

src/python-frontend/python_typechecking.cpp
sergillam: I’ve refactored the Python type checking to reuse type_handler for both annotation resolution and inheritance checks, keeping the logic centralized and ready to be moved into a dedicated component later.

src/goto-programs/goto_check.cpp
sergillam: Done

lucasccordeiro:

PR #3210, opened 2025-11-27
[python] Fix list comparison when function call returns a list

brcfarias:


lucasccordeiro (approved):

Could you please check regression/python-intensive/problem3_1_fail (Timeout)?

You should set the unwinding bound here.

brcfarias:
Partial fix for #3213

PR #3227, opened 2025-12-01
[python] Fix nested return inference

### LINKED ISSUE #3225: [python] Failure on regression/quixbugs/gcd
main.py:
```python
def gcd(a, b):
    if b == 0:
        return a
    else:
        return gcd(b, a % b)


assert gcd(17, 0) == 17
assert gcd(13, 13) == 13
assert gcd(37, 600) == 1
assert gcd(20, 100) == 20
assert gcd(624129, 2061517) == 18913
assert gcd(3, 12) == 3
```

ESBMC (version: c4ac093174) output:
```bash
Parsing regression/quixbugs/gcd/main.py

Type checking warning:

<unknown>:45: SyntaxWarning: invalid escape sequence '\d'
Converting
Generating GOTO Program
WARNING: function should not return value

* file: regression/quixbugs/gcd/main.py
* line: 3
* function: gcd
* column: 8
WARNING: function should not return value

* file: regression/quixbugs/gcd/main.py
* line: 5
* function: gcd
* column: 8
GOTO program creation time: 0.679s
GOTO program processing time: 0.011s
[1]    25033 segmentation fault (core dumped)  esbmc regression/quixbugs/gcd/main.py
```

brcfarias:
Fixes type inference for functions with returns inside control flow blocks (if/else/while/for). Previously returns inside if/else were not found resulting in void return type. Now uses infer_from_return_statements to recursively search nested blocks.
Fixes #3225


lucasccordeiro (approved):

PR #3247, opened 2025-12-04
[python] Add nested function support with closure semantics

### LINKED ISSUE #3233: [python] Nested functions cannot access outer scope variables
ESBMC fails to resolve variables referenced by nested functions.

main.py:
```python
def foo(x):
    def bar():
        return x
    return bar()

assert foo(1) == 1
```

ESBMC (a04f7c43a8) output:
```bash
Parsing main.py

Type checking warning:

<unknown>:45: SyntaxWarning: invalid escape sequence '\d'
Converting
ERROR: Symbol not found py:main.py@x
[1]    19873 IOT instruction (core dumped)  esbmc main.py
```

brcfarias:
## Summary
Adds support for nested functions with closure semantics in the Python frontend.

## Changes
- Implement hierarchical function naming (`parent@F@child`)
- Add closure variable resolution across parent scopes
- Refactor `json_utils::find_var_decl()` with modular helper functions
- Support nested function calls in `function_call_builder`
- Add `scope_stack_` for runtime symbol resolution

## Example
```python
def foo(x):
    def bar():
        return x  # bar() can access x from foo()
    return bar()

assert foo(1) == 1   # Now works
```

Fixes #3233


lucasccordeiro (approved):

PR #3253, opened 2025-12-04
update date for PPA readme

LukeW1999:
Just update the date.

BTW, I really recommend people to check PPA build, if they want to use it or recommend it to other users. It is only provided with a minial usable version with Z3 only.

If you think some feature need to be supported, just let me know.


lucasccordeiro (approved):

PR #3254, opened 2025-12-04
[Build] Change the fmt usage to explicitly format

LukeW1999:
### Description

This Pull Request addresses a potential namespace ambiguity/conflict by explicitly qualifying all calls to the formatting function `format_to` with the `fmt::` namespace prefix.

### Changes

The core change across all custom `fmt::formatter` specializations in the `src/util/message/formats` directory is:

* Changing `return format_to(ctx.out(), ...);`
* To the explicit call: `return fmt::format_to(ctx.out(), ...);`

### Rationale

In C++ projects, especially those integrating multiple libraries, an unqualified call to `format_to` can lead to ambiguity issues (due to Argument-Dependent Lookup or other name resolution complexities) if another library or the global namespace defines a function with the same name.

### DIFF
--- a/src/util/message/formats/bigint.h
+++ b/src/util/message/formats/bigint.h
@@ -40,6 +40,6 @@ struct fmt::formatter<BigInt>
     char tmp[128];
     char *number;
     number = p.as_string(tmp, 128, base);
-    return format_to(ctx.out(), "{}", number);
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

PR #3257, opened 2025-12-05
[Python] Add support for generating pytest test files

prototypeC14:
This PR adds support for generating pytest test files from ESBMC counterexamples. When a property is violated, the corresponding counterexample is now translated into a pytest test case, so that the generated tests can be executed and managed
using pytest.


lucasccordeiro:

Can you please add some regression tests and documentation explaining how to use this new feature?

lucasccordeiro:
@prototypeC14: what do you think if we create a new class "pytest" derived from witness?

After this PR, it would be great to create executable test cases for C using ctest.

src/goto-symex/witnesses.h
```diff
@@ -237,3 +240,76 @@ void generate_testcase(
   const std::string &file_name,
   const symex_target_equationt &target,
   smt_convt &smt_conv);
+
+/// This generates pytest test-cases for Python programs
```
lucasccordeiro: Shall we create a pytest.cpp and pytest.h?

lucasccordeiro (changes requested):

lucasccordeiro (approved):

lucasccordeiro:

PR #3260, opened 2025-12-05
[python] Add all() and any() built-in functions

### LINKED ISSUE #3240: [python] Add support for the all built-in function
Implement a model for the  [all](https://docs.python.org/3/library/functions.html#all) built-in function.

brcfarias:
Implements `all()` and `any()` built-in functions as Python models.

- `all(iterable)` - Returns `True` if all elements are truthy (or if empty)
- `any(iterable)` - Returns `True` if any element is truthy

Fixes #3240


lucasccordeiro (approved):

PR #3262, opened 2025-12-06
feat: web install script for ESBMC

Yiannis128:
An easy and convenient script that downloads the latest release of ESBMC and installs it on a user's system using the following command (when merged to master):

```sh
curl 'https://raw.githubusercontent.com/esbmc/esbmc/refs/heads/master/scripts/web-install.sh' | bash
```

If you add that command to the ESBMC website, then people can install it to `~/.local/bin/esbmc` very conveniently and easily.

I've also adjusted the README.md as there were various formatting inconsistencies. Now it should look much nicer.

Example Run:
```sh
curl 'https://raw.githubusercontent.com/esbmc/esbmc/refs/heads/web-install/scripts/web-install.sh' | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2511  100  2511    0     0   9393      0 --:--:-- --:--:-- --:--:--  9404
ESBMC Linux Install Script

✓ wget is installed
✓ unzip is installed
✓ uuidgen is installed

🚧 Working Directory: /tmp/esbmc-web-install-ae779269-3f38-4d3a-b6d1-5caf661c7705
⏳️ Downloading latest version from https://github.com/esbmc/esbmc/releases/latest/download/esbmc-linux.zip
⏳️ Extracting esbmc-linux.zip

🌟 ESBMC version 7.11.0 64-bit x86_64 linux is successfully installed at ~/.local/bin/esbmc

🧹☁️ Cleaning up /tmp/esbmc-web-install-ae779269-3f38-4d3a-b6d1-5caf661c7705...

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,25 +20,37 @@ We provide some background material/publications to help you understand what ESB
 
 Our main website is [esbmc.org](http://esbmc.org). 
 
-### How to build/install ESBMC
+## Installing ESBMC
 
-#### Installing via PPA (Ubuntu - Recommended)
+### Ubuntu (PPA)
 
 The easiest way to install ESBMC on Ubuntu is through our official [PPA](https://launchpad.net/~esbmc/+archive/ubuntu/esbmc), which provides stable releases for automatic installation:
 
-````
+```
 sudo add-apt-repository ppa:esbmc/esbmc
 sudo apt update
 sudo apt install esbmc
-````
+```
 
 This method is recommended for general users and supports Ubuntu 22.04 (Jammy) and 24.04 (Noble).
 
-#### Ubuntu 24.04
+### Install Script (Linux)
+
+Run the following command to install ESBMC on Linux systems:
+
+```sh
+curl 'https://raw.githubusercontent.com/esbmc/esbmc/refs/heads/master/scripts/web-install.sh' | bash
+```
+
+>Please review the script before running it: https://github.com/esbmc/esbmc/blob/master/scripts/web-install.sh
+
+## Building ESBMC
+
+### Ubuntu 24.04
 
 To compile ESBMC on Ubuntu 24.04 with LLVM 14 and the SMT solver Z3:
 
-````
+```
 sudo apt update
 sudo apt-get install -y clang-14 llvm-14 clang-tidy-14 python-is-python3 python3 git ccache unzip wget curl bison flex g++-multilib linux-libc-dev libboost-all-dev libz3-dev libclang-14-dev libclang-cpp-dev cmake
 git clone https://github.com/esbmc/esbmc.git
@@ -46,18 +58,18 @@ cd esbmc
 mkdir build && cd build
 cmake .. -DENABLE_Z3=1
 make -j4
-````
+```
 
 To enable Python frontend support, add the `-DENABLE_PYTHON_FRONTEND=1` flag to the cmake command and ensure you have Python 3 with the `ast2json` module installed:
 
-````
+```
 pip install ast2json
 cmake .. -DENABLE_Z3=1 -DENABLE_PYTHON_FRONTEND=1
 make -j4
-````
+```
 
 
-#### Fedora 40
+### Fedora 40
 
 To compile ESBMC on Fedora 40 with the latest version of LLVM and the SMT solver Z3:
 
@@ -75,7 +87,7 @@ To build ESBMC with other operating systems and SMT solvers, please see the [BUI
 
 The user can also download the latest ESBMC binary for Ubuntu and Windows from the [releases page](https://github.com/esbmc/esbmc/releases).
 
-#### FreeBSD
+### FreeBSD
 
 ESBMC should compile just fine in FreeBSD as long as the 32-bit libraries are enabled
 
@@ -87,7 +99,7 @@ cmake .. -DLLVM_DIR=../clang16 -DClang_DIR=../clang16
 make -j4
 ```
 
-#### Mac OS X
+### Mac OS X
 
 M1/M2/M3/M4 Macs are now supported.
 
@@ -106,7 +118,7 @@ make -j8
 make install
 ```
 
-### Using the SMT Boolector Solver
+## Using the SMT Boolector Solver
 
 Boolector is a fast solver and is recommended. To install Boolector, use the following one-line command:
 
@@ -149,10 +161,10 @@ RUN apt-get update && apt-get install -y \
 
 # Keep the container running with tail -f /dev/null
 CMD ["bash", "-c", "tail -f /dev/null"]
-````
+```
 
 Docker compose file:
-````
+```
 version: '3.8'
 services:
   esbmc:
@@ -162,16 +174,17 @@ services:
       dockerfile: Dockerfile  # Assuming your Dockerfile is named `Dockerfile`
     tty: true
     stdin_open: true
-````
+```
+
 The Linux/Amd64 line is very important for virtualizing Amd64. Now do docker-compose up --build. You can then follow the Linux instructions. Make -j16 works well on M2 mac's and beyond.
 
-### How to use ESBMC
+## Usage
 
-#### Verifying C Programs
+### Verifying C Programs
 
 As an illustrative example to show some of the ESBMC features, consider the following C code:
 
-````C
+```C
 #include <stdlib.h>
 int *a, *b;
 int n;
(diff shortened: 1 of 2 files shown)


lucasccordeiro (approved):

rafaelsamenezes (changes requested):
I think this feels a bit too over specific and creating problems that we are not trying to solve.  We already ship the binaries, we can even make an installer with cpack if needed.

I already somewhat regret adding the option to let CMake set our dependencies.

Yiannis128:
> I think this feels a bit too over specific and creating problems that we are not trying to solve. We already ship the binaries, we can even make an installer with cpack if needed.
> 
> I already somewhat regret adding the option to let CMake set our dependencies.

>I think this feels a bit too over specific and creating problems that we are not trying to solve.

What are these problems it's creating?

It has some benefits like automatically updating ESBMC to the latest version everytime there's a release if you alias the command and run esbmc-update:

```sh
alias esbmc-update="..."
esbmc-update
```

Also, it's meant as an alternative "automatic" install method for other Linux distros, since Ubuntu is the only one with a package currently... (This doesn't mean I'm in favor of creating a package for every distro since maintaining that would be a nightmare, I think Ubuntu only is just fine).

Also, we can put a download command on the github page if we use this:

image

Please let me know what you think.

rafaelsamenezes:
> What are these problems it's creating?

We shouldn't be trying to be a package manager or a dependency manager, this leads to users with obscure configurations seeking help. What I mean here is that we are making a promise to users that this will just work; this is will generate unexpected problems.

How about just using the installer from cpack? It already generates a shell script that can be used to install ESBMC whenever they want. I also expect the CMake developers to have faced every weird configuration.

Yiannis128:
> > What are these problems it's creating?
> 
> We shouldn't be trying to be a package manager or a dependency manager, this leads to users with obscure configurations seeking help. What I mean here is that we are making a promise to users that this will just work; this is will generate unexpected problems.
> 
> How about just using the installer from cpack? It already generates a shell script that can be used to install ESBMC whenever they want. I also expect the CMake developers to have faced every weird configuration.

So you're saying we can use CMake to provide the install script, but would that fetch the latest version from the internet? How would users run it?

The issue I'm trying to solve with this PR is to easily update ESBMC. Which puts it in ~/.local/bin which complies with the [XDG Specification](https://specifications.freedesktop.org/basedir/latest/), which should cover the majority of Linux configurations. This is what it says:

> User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.
>
>Since $HOME might be shared between systems of different architectures, installing compiled binaries to $HOME/.local/bin could cause problems when used on systems of differing architectures. This is often not a problem, but the fact that $HOME becomes partially architecture-specific if compiled binaries are placed in it should be kept in mind.

Would you prefer to use cpack? Can we use it the same way as the web-install script?

Also, do you think this will cause maintainability issues in the future?

rafaelsamenezes:
> > > What are these problems it's creating?
> > 
> > 
> > We shouldn't be trying to be a package manager or a dependency manager, this leads to users with obscure configurations seeking help. What I mean here is that we are making a promise to users that this will just work; this is will generate unexpected problems.
> > How about just using the installer from cpack? It already generates a shell script that can be used to install ESBMC whenever they want. I also expect the CMake developers to have faced every weird configuration.
> 
> So you're saying we can use CMake to provide the install script, but would that fetch the latest version from the internet? How would users run it?
> 
> The issue I'm trying to solve with this PR is to easily update ESBMC. Which puts it in ~/.local/bin which complies with the [XDG Specification](https://specifications.freedesktop.org/basedir/latest/), which should cover the majority of Linux configurations. This is what it says:
> 
> > User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.
> > Since $HOME might be shared between systems of different architectures, installing compiled binaries to $HOME/.local/bin could cause problems when used on systems of differing architectures. This is often not a problem, but the fact that $HOME becomes partially architecture-specific if compiled binaries are placed in it should be kept in mind.
> 
> Would you prefer to use cpack? Can we use it the same way as the web-install script?
> 


CPack just generates an .sh that can be used in the same way pretty much. And it comes for free if we ever decide to add libraries, manuals or other installed files.

I would just add it into the releases automatically and then use same command that you provided but with it instead. So it would be the same as the web-install. Althought it would ask the user to accept our license (which can be skipped with flags).

> Also, do you think this will cause maintainability issues in the future?

You mean with your script or cpack? CPack main issue is if we ever move out from CMake. Your PR main issue is if users starts coming up with weird configurations and add more and more corner cases.

Yiannis128:
> > > > What are these problems it's creating?
> > > 
> > > 
> > > We shouldn't be trying to be a package manager or a dependency manager, this leads to users with obscure configurations seeking help. What I mean here is that we are making a promise to users that this will just work; this is will generate unexpected problems.
> > > How about just using the installer from cpack? It already generates a shell script that can be used to install ESBMC whenever they want. I also expect the CMake developers to have faced every weird configuration.
> > 
> > 
> > So you're saying we can use CMake to provide the install script, but would that fetch the latest version from the internet? How would users run it?
> > The issue I'm trying to solve with this PR is to easily update ESBMC. Which puts it in ~/.local/bin which complies with the [XDG Specification](https://specifications.freedesktop.org/basedir/latest/), which should cover the majority of Linux configurations. This is what it says:
> > > User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.
> > > Since $HOME might be shared between systems of different architectures, installing compiled binaries to $HOME/.local/bin could cause problems when used on systems of differing architectures. This is often not a problem, but the fact that $HOME becomes partially architecture-specific if compiled binaries are placed in it should be kept in mind.
> > 
> > 
> > Would you prefer to use cpack? Can we use it the same way as the web-install script?
> 
> CPack just generates an .sh that can be used in the same way pretty much. And it comes for free if we ever decide to add libraries, manuals or other installed files.
> 
> I would just add it into the releases automatically and then use same command that you provided but with it instead. So it would be the same as the web-install. Althought it would ask the user to accept our license (which can be skipped with flags).
> 
> > Also, do you think this will cause maintainability issues in the future?
> 
> You mean with your script or cpack? CPack main issue is if we ever move out from CMake. Your PR main issue is if users starts coming up with weird configurations and add more and more corner cases.

Hi, I will look into this cpack and adjust this PR to see if it's worth implementing that over this. I will remake as draft.

LukeW1999:
@Yiannis128 I think this is a nice idea for improving user convenience. If it works well, it could be a good addition. At least the downloading link provided in the scripts don't need to change.

Yiannis128:
@rafaelsamenezes I've looked at cpack and here are my main concerns:

1. ESBMC build process is already extremely bloated and slow: takes minimum **1h but some runs take 2h 30m**, if we add cpack, we would increase it further with this. Also, it requires additional packages to be installed to run cpack for each type of distribution you want to build for.
2. cpack is _too_ complicated imo, the web script relies on existing GitHub release infrastructure but doesn't touch the build process, it's independent. I would rather not add additional bloat to the build process for something as simple as putting the ESBMC executable in ~/.local/bin.
3. The web install script downloads the latest version of esbmc and puts it in ~/.local/bin which is compliant with linux desktop standards, so we shouldn't be obligated to support non-standard Linux configs anyway.
4. You can store the web installer as well into ~/.local/bin and run it when there's a new ESBMC release and automatically update the ESBMC version, which cpack can't do.
5. Most weird linux configs usually are done by people who wouldn't use the install script, or read it before running it as they're probably advanced users, the install script is meant for beginners. Also, there's been cases where cpack can fail as well if not configured properly, which is quite hard to do for ESBMC.
6. There's other open source projects that use install scripts, from the top of my head https://ollama.com/ uses it.

Here is why I would choose cpack:

* We start providing installation mediums for different OS' and Linux distros. Like if we want to start creating Windows installers and Linux installers, but at that point we are adding more work for us and I think is out of scope for what we're doing.

Could we trial run this and see if we get any issues, perhaps (then remove it)?

lucasccordeiro:
@rafaelsamenezes: do you have any thoughts here?

rafaelsamenezes:
> @rafaelsamenezes I've looked at cpack and here are my main concerns:
> 

I am sorry, I did not like any of these arguments.

- The advantage of cpack is that any changes on CMake are reflected on it. If we ever have the need of installing extra plugins or libraries we will have to update the web script. It also adds the same time complexity of zipping a folder.
- I can give a few projects that have/had something like this. Ollama, docker, and even the verify rust std initiative. This makes sense for those projects because to run the tool you need to detect hardware, install dependencies, create services, add PPAs. These rely on very specific/sketchy environments. ESBMC depends only on libc headers, which the web install script is not preparing at all.

At least it should behave more like an installer (allow users to select destination) or install the dependencies needed. I would even prefer just a command block teaching a wget for the latest release + unzip command.

Yiannis128:
Outdated, already applied README.md formatting changes, and not planning to implement `web-install.sh` anymore

PR #3266, opened 2025-12-06
[Code Contract] Add Function Contract Support

LukeW1999:
## Summary

Implements **function contract verification** for ESBMC, enabling modular verification through contract-based abstraction. This is Phase 1 of a two-phase implementation, focusing on function-level verification.

### Format problem
Despite it would be better to make the format like 
```
int incremental (int x) __ESBMC_require {
}
```
But this format would rely on Clang attribute annotation and can only pass in the string, which increase too much development effort.

I changed the design to this format.
```
int incremental (int x) {
__ESBMC_require 
}
```

### New Contract Syntax
- `__ESBMC_requires(condition)` - Precondition clauses
- `__ESBMC_ensures(condition)` - Postcondition clauses  
- `__ESBMC_return_value` - Special variable for return values in ensures clauses
- `__ESBMC_old(variables)` - Use to save the last frame result of the variables and prepare for comparison  

### Command-Line Option
```bash
esbmc program.c --enforce-contract "*"          # All functions
esbmc program.c --enforce-contract func_name    # Specific function
```

### How It Works
For a function with contracts, ESBMC:
1. Renames original function to `__ESBMC_contracts_original_F`
2. Generates wrapper that:
   - **Assumes** requires (precondition)
   - Calls original function
   - **Asserts** ensures (postcondition)

## Example

```c
int increment(int x)
{
    __ESBMC_requires(x > 0);
    __ESBMC_ensures(__ESBMC_return_value > x); // Will pass
    return x + 1;
}

int main()
{
    int a = 5;
    int result = increment(a);
    assert(result > a);  
    return 0;
}
```

**Verification:**
```bash
$ esbmc main.c --enforce-contract "*"
VERIFICATION SUCCESSFUL
```

```c
int increment(int x)
{
    __ESBMC_requires(x > 0);
    __ESBMC_ensures(__ESBMC_return_value == x); // Will fail
    return x + 1;
}

int main()
{
    int a = 5;
    int result = increment(a);
    assert(result > a);  
    return 0;
}
```

**Verification:**
```bash
$ esbmc main.c --enforce-contract "*"
VERIFICATION FAILED
```
### Known Limitation: Counterexample Traces

**Note**: Contract-based verification may produce less detailed counterexample traces, similar to loop invariant verification. This is due to the over-approximation inherent in modular reasoning. While detailed trace reconstruction is theoretically possible using CEGAR (Counterexample-Guided Abstraction Refinement), it is deferred to future work. This trade-off is standard in contract-based verification systems.

Another need to be aware, as an user, you might realize it is not understandable some counterexample when using --function with this --enforce-contract. This is because, the old dependency would be throw away if you use it with --function. The reason is just same with the former one.

## Testing

Added **26 regression tests** covering:
- Basic functionality (requires/ensures)
- All comparison operators (`>`, `<`, `>=`, `<=`, `==`, `!=`)
- Complex logical expressions (`&&`, `||`)
- Multiple parameters and return types (`int`, `double`, `float`)
- Boundary values (zero, negative, large numbers)
- Failure scenarios (ensures/requires violations)

All tests pass ✅

## Key Technical Features

1. **Semi-Dynamic Type System**: `__ESBMC_return_value` automatically matches function return type
2. **Symbol Table Management**: Proper handling to avoid pollution
3. **Integration with `--function` option**: Works seamlessly for isolated testing

## What's NOT Included (Future PR)

The following are **intentionally left for Phase 2**:
- `--replace-call-with-contract` mode (system-level verification)
- `assigns` clause support  
- Havoc mechanism for frame conditions

These require additional infrastructure (havoc support) and will be completed in the next PR.

## Files Changed

- `src/goto-programs/contracts/` - New contract processing module (750 lines)
- `src/clang-c-frontend/` - Semi-dynamic type support for `__ESBMC_return_value`
- `src/esbmc/` - Command-line option integration
- `regression/function_contract/` - 26 new test cases

## Backwards Compatibility

✅ No breaking changes - contract features only activate with `--enforce-contract` flag

## Related Issues

This implements contract-based verification similar to Frama-C/ACSL, SPARK Ada, and Dafny.

---

**Phase 1 Complete**: Function-level verification ready for production use  
**Phase 2 Coming**: System-level verification with havoc mechanism


lucasccordeiro (approved):

lucasccordeiro:

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2628,6 +2633,111 @@ void esbmc_parseoptionst::print_ileave_points(
...
+  // Process enforce-contract option
+  if (has_enforce)
+  {
+    std::set<std::string> to_enforce;
+    const std::list<std::string> &enforce_list =
+      cmdline.get_values("enforce-contract");
+
+    for (const auto &func : enforce_list)
+    {
+      if (func == "*")
```
rafaelsamenezes, line 2687: This could be turned into a lambda and reused in the next block. I also am not very happy with this logic here... maybe add a TODO to have a field in the goto functions for this.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2026,6 +2027,10 @@ bool esbmc_parseoptionst::process_goto_program(
 
     goto_check(ns, options, goto_functions);
 
+    // Process function contracts if enabled
+    if (process_function_contracts(options, goto_functions))
```
rafaelsamenezes: I would prefer to move the checks from the function into here.

```c++
bool has_enforce = cmdline.isset("enforce-contract");
bool has_replace = cmdline.isset("replace-call-with-contract");
if(has_enforce || has_replace)
  process_function_contracts(goto_functions, has_replace, has_enforce); // also changing the method signature.
```

src/esbmc/esbmc_parseoptions.h
```diff
@@ -47,6 +47,14 @@ class esbmc_parseoptionst : public parseoptions_baset, public language_uit
   virtual bool
   output_goto_program(optionst &options, goto_functionst &goto_functions);
 
+  /// \brief Process function contracts if enabled
+  /// \param options Command line options
+  /// \param goto_functions GOTO functions
+  /// \return True if processing should stop, false otherwise
+  bool process_function_contracts(
```
rafaelsamenezes: I suggest moving this signature to `void process_function_contracts(goto_functionst, bool, bool)`.

The function itself does not throw any errors; and even if did... I would prefer an abort.

src/goto-programs/contracts/contracts.h
```diff
@@ -0,0 +1,133 @@
...
+#define ESBMC_CONTRACTS_H
+
+#include <goto-programs/goto_functions.h>
+#include <util/context.h>
+#include <util/namespace.h>
+#include <set>
+#include <string>
+
+/// \brief Basic contract handling class
+/// Provides contract checking and replacement functionality at goto function level
```
rafaelsamenezes, line 48: Add a brief explanation for what we understand as contracts.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
+/*
```
rafaelsamenezes: Probably the wrong place for this comment. Also a bit redundant as the class docs also contains the same information (except complexity) .

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+void code_contractst::rename_function(
+  const irep_idt &old_id,
+  const irep_idt &new_id)
+{
+  auto it = goto_functions.function_map.find(old_id);
+  if (it == goto_functions.function_map.end())
+  {
+    log_error("Function {} not found for renaming", old_id);
+    return;
```
rafaelsamenezes: Shouldn't this be an abort?

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+  }
+
+  // Copy function to new name
+  goto_functiont &old_func = it->second;
+  goto_functions.function_map[new_id] = old_func;
+  goto_functions.function_map[new_id].update_instructions_function(new_id);
+
+  // Update symbol table
+  symbolt *old_sym = context.find_symbol(old_id);
+  if (old_sym != nullptr)
```
rafaelsamenezes: Shouldn't this be ensured?

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+expr2tc code_contractst::extract_ensures_clause(const symbolt &contract_symbol)
+{
+  // Extract from contract symbol's value field
+  if (contract_symbol.value.is_nil())
+    return gen_true_expr();
+
+  // For now, we store both requires and ensures in the value
+  // TODO: Implement proper separation if needed
+  return gen_true_expr();
```
rafaelsamenezes, line 184: This seems incorrrectt

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+}
+
+expr2tc
+code_contractst::extract_assigns_clause(const symbolt & /* contract_symbol */)
+{
+  // TODO: Extract assigns clause from contract symbol
+  return expr2tc();
```
rafaelsamenezes, line 192: add log_warning or abort. So users of this function knows that is still not implemented

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+  if (targets.empty())
+    return;
+
+  for (const auto &target : targets)
+  {
+    expr2tc rhs = gen_nondet(target->type);
+    goto_programt::targett t = dest.add_instruction(ASSIGN);
+    t->code = code_assign2tc(target, rhs);
+    t->location = location;
+    t->location.comment("contract assigns havoc");
```
rafaelsamenezes: We can probably get a better comment here. Users will be very confused with this.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -0,0 +1,627 @@
...
+  }
+}
+
+void code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
+{
+  for (const auto &function_name : to_enforce)
+  {
+    // Skip compiler-generated functions (destructors, etc.)
```
rafaelsamenezes: Is there any specific reason for this?

regression/function_contract/basic_replace_call/main.c
```diff
@@ -0,0 +1,27 @@
+/* Test: Requires violation with --replace-call-with-contract
```
rafaelsamenezes, line 1: Test description is wrong

rafaelsamenezes (changes requested):
This is a great PR @LukeW1999 , introducing contract-based verification into ESBMC. I am not very familiar with contracts, so feel free to correct me regarding my comments.

The ideia is to make use of new intrinsic functions to:

- Mark predicates as precondition: REQUIRE.
- Mark predicates as postcondition: ENSURE.
- Explicit return value: RETURN_SYMBOL.

It then sets two strategies:

- Enforce: It works by wrapping the function call into a: ASSUME(REQUIRE) ==> ASSERT(ENSURE).
- Replace: It works by converting the function into: ASSERT(REQUIRE)  ==> ASSUME(ENSURE)

Regarding partial correctness, both strategies seems unsound and incomplete. This is due to "enforce" discarding the base case and the "replace" not ensuring the property holds. That being said, by havoc + enforce to prove that a contract holds and then replacing the uses is sound. I think that this is similar to what you are trying to achieve?

Another issue that I am having here is that your contract of `x > 0 ==> x + 1 > x` is not correct when considering bitvectors as `x = MAX_INT ==> x + 1 = MIN_INT`.

LukeW1999:
Hi @rafaelsamenezes , 

Thank you for the detailed review - your feedback is incredibly helpful. You're the expert on abstract interpretation in our group, and I really value your guidance.

Regarding code quality: I acknowledge it needs improvement. This document focuses on explaining the soundness argument first, and I'll refactor the implementation accordingly.

### Two-Phase Modular Verification Strategy

You mentioned that "havoc + enforce to prove that a contract holds and then replacing the uses is sound." I believe our approach follows this same principle, though the implementation details may differ slightly. Let me explain our strategy and the soundness argument:

**Phase 1 (this PR):** Prove implementation satisfies contract
```
ASSUME(requires) → [actual code] → ASSERT(ensures)
```

**Phase 2 (planned):** Use proven contracts at call sites
```
ASSERT(requires) → havoc(return + side_effects) → ASSUME(ensures)
```

### Why This Is Sound

**Formal argument:**

Let `V_post(s) = {r | Post(f, s, r)}` = all values satisfying the postcondition for input state `s`.

**What Phase 1 proves:**
```
∀s. Pre(f)(s) ⟹ ⟦f⟧(s) ∈ V_post(s)
```
Meaning: For any state satisfying the precondition, the actual execution result is in `V_post`.

**What Phase 2 verifies at call site `r = f(x)`:**

At call site with concrete input state `s₀`:
```
1. assert(Pre(f)(s₀))           // Caller must satisfy precondition
2. havoc(r)                      // r becomes arbitrary value
3. assume(r ∈ V_post(s₀))       // Restrict to postcondition
4. assert(φ(r))                  // Prove property holds
```

If Phase 2 passes, we've proven: `Pre(f)(s₀) ⟹ (∀r ∈ V_post(s₀). φ(r))`

**Complete soundness proof:**

Let `r_actual = ⟦f⟧(s₀)` be the actual execution result.

```
From Phase 1: Pre(f)(s₀) ⟹ r_actual ∈ V_post(s₀)     ... (1)
From Phase 2 step 1: Pre(f)(s₀) holds                  ... (2)
From (1) and (2): r_actual ∈ V_post(s₀)                ... (3)

Phase 2 verification goal: ∀r ∈ V_post(s₀). φ(r)      ... (4)
If Phase 2 passes, (4) is proven                        ... (5)
From (3) and (5): φ(r_actual)                           ... (6)

∴ If both phases pass, the actual execution satisfies φ. ∎
```

**Key insight:** 
- Phase 1 proves: `r_actual ∈ V_post` (implementation ⊆ contract)
- Phase 2 proves: `∀r ∈ V_post. φ(r)` (property holds for all contract-allowed values)
- Combined: `φ(r_actual)` (actual execution satisfies property)

Havoc is essential: it creates the universal quantification, then assume restricts to `V_post`, ensuring we only assume what was proven.

### Example: Weak Contract (Still Sound)

```c
int add_one(int x) {
    __ESBMC_requires(x > 0 && x < INT_MAX);
    __ESBMC_ensures(__ESBMC_return_value > x);  // Weak: doesn't say == x+1
    return x + 1;
}

int y = add_one(5);
assert(y > 5);   // ✓ Passes
assert(y == 6);  // ✗ Fails - contract doesn't promise this
```

**Phase 1 verification:**
```
V_post(5) = {r | r > 5} = {6, 7, 8, 9, ...}  // Infinite set
r_actual = ⟦add_one⟧(5) = 6

Phase 1 proves: 6 ∈ {6, 7, 8, 9, ...} ✓
```

**Phase 2 verification:**
```
assert(5 > 0 && 5 < INT_MAX)  ✓ Precondition holds
havoc(y)                       // y ∈ ℤ (all integers)
assume(y > 5)                  // y ∈ {6, 7, 8, ...}

assert(y > 5):  Prove ∀y ∈ {6,7,8,...}. y > 5
                True by construction ✓ PASS

assert(y == 6): Prove ∀y ∈ {6,7,8,...}. y == 6
                False: counterexample y = 7
                ✗ FAIL (SMT solver should find: y = 7 ∈ V_post but 7 ≠ 6)
```

**Why this is sound:**
- Contract promises `result > x`, not `result == x + 1`
- `V_post(5)` contains infinitely many values: {6, 7, 8, 9, ...}
- Even though `r_actual = 6`, the weak contract can't rule out other values
- We correctly reject `assert(y == 6)` because it's not guaranteed by the contract

**Without havoc (UNSOUND):**
```
y = 5 + 1              // Assumes implementation details!
assume(y > 5)          
assert(y == 6)         ✓ INCORRECTLY PASSES
```
This would make verification implementation-dependent rather than contract-dependent.

---

### Questions for You
**1. "Enforce discards base case"**

I'm not sure I fully understand this concern. Could you clarify which scenario you mean?

Are you referring to recursive functions like factorial? If so, I agree this is genuinely challenging. I'm struggling with how to write precise contracts for such functions - the recursive nature makes it unclear what the postcondition should even express without circular dependencies.

Should we explicitly scope our approach to exclude recursive functions, or is there a way to handle them that I'm missing?

**2. Integer overflow**

I did not notice that before. You are right about the bitvector semantics issue. I'm struggling with the best approach:

**What would you recommend?** Should contracts operate on mathematical integers or match C's bitvector semantics?

**3. Side-effect analysis for Phase 2**

For havoc completeness (return value + all modified variables), My plan is using a hybird approach:
- **Auto-inference from GOTO:** Analyze function body to find all assignments (zero annotation burden, but may miss aliasing)
- **Explicit `assigns` clauses (Frama-C style):** `__ESBMC_assigns(*p, global);` (precise but verbose)

---

Thanks again for your time and generous help!

rafaelsamenezes:
Thanks for the explanation, I feel like I can understand a bit more about what you are trying to achieve.

> Questions for You

> 1. "Enforce discards base case". Are you referring to recursive functions like factorial?

No, it was a bad choice of words from my side. Right now Phase 1 is supposed to prove that the contract itself holds. For that... you should do:

`HAVOC(F) ==> ASSUME(requires) ==> [actual code]  ==> ASSERT(ensures)`

This way you are proving that if an input meets the requirement, then the contract will hold. I do have some concern if this is not a pure function... but it is ok for now.

Phase 2 is about replacing the contract itself with the property. So you do:

`[program code] ==> ASSERT(requires) ==> ASSUME(ensures)`

Is my understanding correct? I feel like this way we can still verify programs in which the contract vacuously hold:

```c
int my_contract(int x) {
    __ESBMC_requires(0);
   __ESBMC_ensures(0);
    return 0;
}
```

Right now, Phase 1 will tell that this contract is valid for example.


> 2. Integer overflow

I am not sure why it is not working already. Give a look at ESBMC VCC and see if the property is being translated correctly. I suggest just fixing the current test cases (by adding the extra `&& x < MAX_INT`) for now, and maybe adding as a KNOWNBUG. This shouldn't be a blocker.

> 3.  Side-effect analysis for Phase 2

I would argue that there shouldn't be any havoc in Phase 2 (unless we are doing some kind verification with it).

LukeW1999:
Hi @rafaelsamenezes,

Thank you for the patient explanation! 

## Phase 1: I agree completely

You're absolutely right that Phase 1 needs:
```
HAVOC(F) ==> ASSUME(requires) ==> [actual code] ==> ASSERT(ensures)
```

Currently, when using `--function` mode, ESBMC automatically havocs uninitialized parameters, so the implementation implicitly follows this pattern:
```
--function foo ==> [auto HAVOC parameters] ==> ASSUME(requires) ==> [actual code] ==> ASSERT(ensures)
```

However, I didn't make this explicit in the code or documentation, which creates a problem: **if someone verifies a contract from `main()` with side effects, the contract might incorrectly pass** because we're not explicitly ensuring the havoc parameters step happens.

### What I would do:
I would ensure to make havoc before the function be called to make the verification here be more sound.


## Phase 2: Critical question about havoc necessity (Correct me if I am wrong)

Regarding Phase 2, I have a technical concern about not using havoc. Let me illustrate with a concrete example:

```c
int global = 10;

int foo(int x) {
    __ESBMC_requires(x >= 0);
    __ESBMC_ensures(__ESBMC_return_value == x + 1);
    __ESBMC_assigns(global);
    __ESBMC_ensures(global == \old(global) + x);
    
    global += x;
    return x + 1;
}

int main() {
    int result = foo(5);
    assert(global == 15);
    assert(result == 6);
}
```

**Phase 2 without havoc:**
```c
int main() {
    int global = 10;  // global already has concrete value 10
    
    // Replace foo(5) call
    assert(5 >= 0);
    
    int result;            // Uninitialized = symbolic ✓
    assume(result == 6);   // ✓ Can constrain result
    
    assume(global == 15);  // ✗ Contradiction! global is already 10
    // Path becomes unreachable (like assume(false))
    
    assert(global == 15);  // "Passes" because unreachable
    assert(result == 6);   // "Passes" because unreachable
    assert(false);         // Would also "pass"!
}
```

**My concern:** If we don't havoc variables that already have concrete values, the `assume` statements will create contradictions, making the entire path unreachable. This means:
1. All subsequent assertions will vacuously pass (unsound!)
2. Even `assert(false)` would "pass" because the path is unreachable

**Am I missing something here?** Without havoc, how can `assume` modify variables that already have concrete values without causing the path to become unreachable?

**Phase 2 with havoc:**
```c
int main() {
    int global = 10;
    
    assert(5 >= 0);
    
    HAVOC(result);         // result becomes symbolic
    HAVOC(global);         // global "forgets" it was 10, becomes symbolic
    
    assume(result == 6);   // ✓ Constrains symbolic result
    assume(global == 15);  // ✓ Constrains symbolic global
    
    assert(global == 15);  // ✓ Proves correctly
    assert(result == 6);   // ✓ Proves correctly
}
```

This is why I think havoc is essential - it makes variables symbolic so `assume` can constrain them without creating contradictions.

## Integer Overflow

You're right about this issue. I think my PR would cause some effect on the existing assertions. I was only think about user designed assertion when designing.

I still need some time to figure out and improve my code. I would change it to draft and ask your help again. I think for this Integer Problem I maybe need to fix it after submitting a MVP.

Thanks again for your guidance!

rafaelsamenezes:
You need to consider that both phases are needed in your proof obligation. For example, in your code:

```c
int global = 10;

int foo(int x) {
    __ESBMC_requires(x >= 0);
    __ESBMC_ensures(__ESBMC_return_value == x + 1);
    __ESBMC_assigns(global);
    __ESBMC_ensures(global == \old(global) + x);
    
    global += x;
    return x + 1;
}

int main() {
    int result = foo(5);
    assert(global == 15);
    assert(result == 6);
}
```

Indeed, as you pointed out Phase 2 will result in:

```
ASSERT (x >= 0); // 5 >= 0 OK
ASSUME( global = global + x)  // ASSUME(false)
```

However, Phase 1 will result in:

```
global = *;
x = *;
ASSUME(x >= 0); 
ASSERT(global = global + x)
```

Phase 1 should catch that the contract is broken.

LukeW1999:
Hi, @rafaelsamenezes 

Thank you for the detailed explanation! You're incredible.
I acknowledge that my current design has a flaw - the responsibilities of Phase 1 and Phase 2 are mixed.
You correctly identified that:

Phase 1 should be more eager about verifying contract correctness (using havoc to ensure the contract holds for all possible states)
Phase 2 should focus cleanly on state transformation 

I wasn't clearly separating these concerns, which caused confusion about when and why havoc is needed.
My Plan
I'll refactor to make this separation clearer:

Phase 1: Aggressively verify contract correctness with havoc
Phase 2: Properly handle state transformation for call-site verification

I'll update my PR to reflect this improved design. Thanks again for your guidance - this discussion really clarified the fundamental principles! I mean I really apperciate you can help me and share your insight with me.

src/goto-programs/contracts/contracts.h
LukeW1999: added at the top of the contract.h. I think I need to imporve it though.

src/esbmc/esbmc_parseoptions.cpp
LukeW1999: I have changed to use a lambda here. I added a TODO to clearify the next plan.

src/goto-programs/contracts/contracts.cpp
LukeW1999: I added the log_warning, should be quick to implement the system level replacement of the contract. But I cannot guarantee since I am a little bit struggle with the sideeffect and havoc right now.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Changed

src/goto-programs/contracts/contracts.cpp
LukeW1999: I changed to use a log_error and abort for nullptr

src/goto-programs/contracts/contracts.cpp
LukeW1999: now it is 
`    t->location.comment("contract assigns: assign non-deterministic value");`

regression/function_contract/basic_replace_call/main.c
LukeW1999: fixed, I added a rejection to --replace-call-with-contract. it would be like this right now.

```
ERROR: ERROR: --replace-call-with-contract mode is not yet fully implemented.

Current status:
  ✓ --enforce-contract mode: FULLY SUPPORTED
  ✗ --replace-call-with-contract mode: NOT IMPLEMENTED

The replace mode requires:
  - Contract symbol management
  - __ESBMC_old() snapshot handling at call sites
  - __ESBMC_return_value replacement
  - Proper assigns clause havoc logic

Please use --enforce-contract mode instead.

TODO: Complete implementation of replace_calls() and generate_replacement_at_call()
Aborted (core dumped)
```

LukeW1999:
Hello @rafaelsamenezes Sir, 
Do you have time to review this PR again today so it can be merged.

rafaelsamenezes (approved):
LGTM

lucasccordeiro:

PR #3283, opened 2025-12-09
[python-frontend] - Add Python str.strip  builtins and regression tests

sergillam:
Add suport  str.strip


src/python-frontend/string_handler.cpp
```diff
@@ -909,6 +909,47 @@ exprt string_handler::handle_string_lstrip(
...
+    str_ptr.copy_to_operands(str_expr);
+  }
+
+  side_effect_expr_function_callt call;
+  call.function() = symbol_exprt(func_symbol_id, code_typet());
+  call.arguments().push_back(str_ptr);
+  call.type() = pointer_typet(char_type());
+  call.location() = location;
+
+  return call;
```
brcfarias, line 950: I think that most of the changes in the other points were probably necessary because you're returning a `side_effect_expr_function_callt` here instead of a `code_function_callt` call.
Take a look at what we do here: https://github.com/esbmc/esbmc/blob/9381b01a5108d292c3daa2e22f6cca071ddd5acc/src/python-frontend/python_converter.cpp#L1656-L1667
I don’t remember if our API already has a helper for this. Could you please check?

brcfarias (changes requested):

src/python-frontend/string_handler.h
```diff
@@ -252,6 +252,14 @@ class string_handler
    */
   exprt handle_string_lstrip(const exprt &str_expr, const locationt &location);
 
+  /**
+   * @brief Handle str.strip() method (leading/trailing whitespace)
```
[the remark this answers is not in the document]
sergillam: Done

src/python-frontend/string_handler.cpp
sergillam: Done

src/pointer-analysis/value_set.cpp
```diff
@@ -1100,6 +1100,15 @@ void value_sett::assign(
       auto *rhs_data = static_cast<const struct_union_data *>(rhs->type.get());
       const std::vector<type2tc> &members = rhs_data->members;
       const std::vector<irep_idt> &member_names = rhs_data->member_names;
+      const bool rhs_is_unknown = is_unknown2t(rhs);
```
lucasccordeiro: We shouldn't change our value set analysis!

src/python-frontend/python_converter.cpp
```diff
@@ -2630,13 +2671,30 @@ symbolt &python_converter::create_tmp_symbol(
   const exprt &symbol_value)
 {
   locationt location = get_location_from_decl(element);
-  std::string path = location.file().as_string();
-  std::string name_prefix =
-    path + ":" + location.get_line().as_string() + var_name;
+  return create_tmp_symbol(var_name, symbol_type, symbol_value, location);
+}
+
+symbolt &python_converter::create_tmp_symbol(
```
lucasccordeiro: Should we reuse our existing methods here?

src/python-frontend/python_converter.cpp
```diff
@@ -3927,17 +3999,17 @@ void python_converter::handle_function_call_rhs(
...
-    target_block.copy_to_operands(rhs);
+    append_statement(target_block, rhs);
 
     code_assignt code_assign(lhs, symbol_expr(tmp_var_symbol));
     code_assign.location() = location;
     rhs = code_assign;
   }
 
-  target_block.copy_to_operands(rhs);
+  append_statement(target_block, rhs);
```
lucasccordeiro: Why is this necessary?

src/python-frontend/python_converter.cpp
```diff
@@ -5589,7 +5710,53 @@ void python_converter::validate_return_paths(
     "Missing return statement detected in function '" + current_func_name_ +
     "'");
 
-  function_body.copy_to_operands(missing_return_assert);
+  append_statement(function_body, missing_return_assert);
+}
+
+void python_converter::validate_block_structure(
```
lucasccordeiro: why do we need this method?

lucasccordeiro (changes requested):
This PR is very complex; it does many more things than its title suggests.

Overall, we should not change our pointer analysis because of the Python frontend unless we really need to add a new feature.

Could you please split it into small PRs and justify most of these changes here?

src/pointer-analysis/value_set.cpp
sergillam: I returned with the file value.cpp from master
Add the processing on the python-front-end

src/python-frontend/python_converter.cpp
sergillam: Thanks for the observarion. I updated the create_tmp_symbol overload that takes only a locationt so that it now reuses our existing helper,  it creates a dummy JSON node, calls the original create_tmp_symbol that takes an AST node, and then reuses the symbol. This keeps the temporary-symbol logic in one place and avoids duplicating the behavior.

Let me know if you'd like this handled differently

src/python-frontend/python_converter.cpp
sergillam: I switched target_block.copy_to_operands(rhs) to append_statement(target_block, rhs) because at that point rhs can still be a plain expression (side_effect_expr_function_callt). copy_to_operands blindly pushes whatever you give it, so we were ending up with non‑code operands under the block. Later goto_convert asserts that every block operand is a codet, so it would crash.

append_statement addresses that by:
1 Checking for a nil node, and
2 Running convert_expression_to_code(rhs), which wraps an expression in a proper code_expression/code_function_callt.


So the change ensures we only enqueue statements in the block and avoids the assertion failure.

src/python-frontend/python_converter.cpp
sergillam: The validate_block_structure came about because there was no other point that caught these errors before the goto_convert

brcfarias: Hi @sergillam, after applying my suggestion, this type of validation shouldn’t be necessary here. I also think it’s better to keep the abort/crash in the goto-conversion process if the frontend is not adding the elements in the block as `codet`.

brcfarias (changes requested):

src/python-frontend/python_converter.cpp
sergillam: Hi @brcfarias, I applied your suggestion;
Are there any other observations?

src/python-frontend/reproduce_issue.py
brcfarias, line 1: Could you remove this file?

src/python-frontend/python_converter.cpp
brcfarias, line 1: I'm not sure all these changes are necessary. Could we try reverting them to see how things behave?

src/python-frontend/reproduce_issue.py
sergillam: done

src/python-frontend/python_converter.cpp
sergillam: remove append_statement and move_statement wrappers

brcfarias:
I simplified the changes a bit. I kept the test `regression/python/type-annotation-reassign-check` as `KNOWNBUG`, so we didn’t need the type-annotation adjustments for now. In the end, the core issue was that the model for `__python_str_strip` had a `for`.  :-)
The goto emitted by c2goto is leaving the first operand as nil in the `for`, as shown below:
```
code
  * type: code
  * operands: 
    0: nil
        * type: 
    1: code
        * type: code
            * arguments: 
        * operands: 
          0: nil
          ...
          ...
        * statement: for
        * #location: 
          * file: esbmc/src/c2goto/library/python/string.c
          * line: 144
          * function: __python_str_strip
          * column: 3
  * statement: block
  * #location: 
    * file: esbmc/src/c2goto/library/python/string.c
    * line: 144
    * function: __python_str_strip
    * column: 3
  * #end_location: 
    * file: esbmc/src/c2goto/library/python/string.c
    * line: 146
    * function: __python_str_strip
    * column: 3
```

With this new check:
```cpp
if (!it.is_code())
{
   log_error(
      "goto_convert: non-code operand in this block:\n{}\n", code.pretty());
   abort();
}
```
it’s easier to spot if anyone else has the same problem.

brcfarias (approved):

lucasccordeiro (approved):

Thanks for simplifying it, @brcfarias.

PR #3289, opened 2025-12-10
[python] Refactor hasattr() handling into symex

brcfarias:
This PR refactors the implementation of `hasattr()` to handle attribute verification into symex:

- The frontend now builds a new irep `hasattr` instead of evaluating the check immediately.
- At the symex level `simplify_python_builtins() ` evaluates hasattr by checking if an object (struct type) has a given attribute using `get_structure_member_names()`.
- For counterexamples, we now print HASATTR(...)

Master:
```
[Counterexample]


State 1 file /tmp/main.py line 8 column 0 thread 0
Violated property:
  file /tmp/main.py line 8 column 0
  assertion
  0


VERIFICATION FAILED
```

This PR:
```
[Counterexample]


State 1 file /tmp/main.py line 8 column 0 thread 0
Violated property:
  file /tmp/main.py line 8 column 0
  assertion
  HASATTR(f, "y")


VERIFICATION FAILED
```

### DIFF
--- a/regression/python/has-attr-fail/main.py
+++ b/regression/python/has-attr-fail/main.py
@@ -1,0 +1,17 @@
+class Foo:
+    def __init__(self):
+        self.x = 10
+
+
+# Foo only has attribute "x".
+f = Foo()
+assert hasattr(f, "y")  # should be false, so this must fail
+
+# TODO: Fix this case
+# try:
+#     _ = f.y
+# except AttributeError:
+#     pass
+
+# Another missing attribute check
+assert hasattr(f, "z")
--- a/regression/python/has-attr-fail/test.desc
+++ b/regression/python/has-attr-fail/test.desc
@@ -1,0 +1,3 @@
+CORE
+main.py
+^VERIFICATION FAILED$
--- a/regression/python/has-attr/main.py
+++ b/regression/python/has-attr/main.py
@@ -1,7 +1,21 @@
 class Foo:
-    pass
+    def __init__(self):
+        self.x = 10
+
 
 f = Foo()
-f.x = 10
-assert hasattr(f, 'x') == True
-assert hasattr(f, 'y') == False
+
+# Existing attribute is detected.
+assert hasattr(f, "x")
+
+# Missing attribute returns False.
+assert not hasattr(f, "y") # This should pass
+
+# After adding dynamically, hasattr turns true.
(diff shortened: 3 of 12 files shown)


src/solvers/smt/smt_conv.cpp
```diff
@@ -1478,6 +1478,48 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
     a = convert_bitcast(expr);
     break;
   }
+  case expr2t::hasattr_id:
```
lucasccordeiro: Why do we need to add this to our SMT backend? 

We can handle this at the symex level; most of this code already exists there.

lucasccordeiro (changes requested):

src/solvers/smt/smt_conv.cpp
brcfarias: Without this case, I was getting:
```
ERROR: Couldn't convert expression in unrecognised format
hasattr
* value : with
    * source_value : symbol
        * name : py:regression/python/has-attr/main.py@f
        * renamelev : Level 2 (global)
        * level1_num : 0
        * level2_num : 0
        * thread_num : 0
        * node_num : 0
        * type : struct
          * members : 
            0: signedbv
              * width : 64

          * member_names : 
            0: x

          * member_pretty_names : 
            0: x

          * typename : Foo
          * packed : false
    * update_field : constant_string
        * value : x
        * kind : default
        * type : array
          * subtype : unsignedbv
              * width : 8
          * array_size : constant_int
              * value : 2
              * type : unsignedbv
                * width : 64
          * size_is_infinite : false
    * update_value : constant_int
        * value : 10
        * type : signedbv
          * width : 64
    * type : struct
      * members : 
        0: signedbv
          * width : 64

      * member_names : 
        0: x

      * member_pretty_names : 
        0: x

      * typename : Foo
      * packed : false
* attr : constant_string
    * value : x
    * kind : default
    * type : array
      * subtype : unsignedbv
          * width : 8
      * array_size : constant_int
          * value : 2
          * type : unsignedbv
            * width : 64
      * size_is_infinite : false
* type : bool
[1]    1672831 IOT instruction (core dumped)  esbmc regression/python/has-attr/main.py
```

But I managed to handle it only with the change in goto_symext::simplify_python_builtins

PR #3306, opened 2025-12-11
[simplifier] improve bitor2t simplifier with additional logical identities

lucasccordeiro:
This PR extends our bitwise simplifier (`bitor2t::do_simplify`) with a set of well-known boolean/bitvector identities that eliminate redundant OR/AND/XOR combinations. These rewrites reduce expression size, improve canonicalization, and increase solver friendliness.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/20149805270

Master (30s): https://github.com/esbmc/esbmc/actions/runs/20149799463

PR #3312, opened 2025-12-12
[Python verification] support str find method

### LINKED ISSUE #3296: [Python] Feature request: add support for str find
Example program:
```python
s: str = "foo:bar:baz"
i = s.find(':')
assert i == 3
```
Currently, ESBMC complains about unsupported function:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
WARNING: Undefined function 'find' - replacing with assert(false)
Generating GOTO Program
GOTO program creation time: 0.685s
Symex completed in: 0.001s (41 assignments)
Slicing time: 0.000s (removed 40 assignments)
Building error trace

[Counterexample]


State 1 file ex.py line 2 column 8 thread 0
Violated property:
  file ex.py line 2 column 8
  Unsupported function 'find' is reached
  0


VERIFICATION FAILED
```
ESBMC version: 12ca0eb1c

XLiZHI:
Fixes https://github.com/esbmc/esbmc/issues/3296.

Some issues with the simplified module have led us to have to use `--incremental-bmc`.

Cannot be simplified:
```
s: str = "foo:bar:baz"
i = s.find(':')
```

Can be:
```
s: str = "foo:bar:baz"
a: str = ':'
i = s.find(a)
```

Additionally, in Python's c2goto, it seems that such code cannot be used. Has anyone encountered this issue?
```
for (int i = 0; i < 1; ++i)
{
}
```

### DIFF
--- a/regression/python/str-find/main.py
+++ b/regression/python/str-find/main.py
@@ -1,0 +1,9 @@
+s: str = "foo:bar:baz"
+i = s.find(':')
+assert i == 3
+
+i1 = s.find('')
+assert i1 == 0
+
+i2 = s.find('d')
+assert i2 == -1
--- a/regression/python/str-find/test.desc
+++ b/regression/python/str-find/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.py
+--incremental-bmc
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/str-find_fail/main.py
+++ b/regression/python/str-find_fail/main.py
@@ -1,0 +1,9 @@
+s: str = "foo:bar:baz"
+i = s.find(':')
+assert i != 3
+
+i1 = s.find('')
+assert i1 == 0
+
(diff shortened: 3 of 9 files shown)


lucasccordeiro (approved):

lucasccordeiro:
> Cannot be simplified:

I'm working quite hard on this; you can check the number of PRs I have already submitted to improve our simplifier.

@rafaelsamenezes is also working on the simplifier now to see whether we can improve performance.

> Additionally, in Python's c2goto, it seems that such code cannot be used. Has anyone encountered this issue?

@brcfarias already told me to avoid this, but I don't understand why we have this limitation :-)

@brcfarias: Do you have any thoughts to share here?

brcfarias:
I’m also getting this error and don’t know the reason yet. For now you can replace it with a corresponding `while`, and it works correctly.

PR #3314, opened 2025-12-12
[simplifier] add boolean and arithmetic simplifier rewrites

lucasccordeiro:
This PR extends our expression simplifier with two new classes of rewrite rules and adds a comprehensive regression suite to ensure correctness across boolean, integer, floating-point, overflow, mixed-width, and casted forms of arithmetic expressions.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/20181333238

Master (30s): https://github.com/esbmc/esbmc/actions/runs/20181347552

PR #3315, opened 2025-12-13
[python] Add string module constants model

brcfarias:
See https://docs.python.org/3/library/string.html
Needed by #3316


lucasccordeiro (approved):

lucasccordeiro:

PR #3317, opened 2025-12-13
[simplifier] add equality-cancellation simplifications with regression coverage for overflow semantics

lucasccordeiro:
This PR extends `equality2t::do_simplify()` with algebraic cancellation rules for bitvector expressions, improving equality simplification while preserving correctness under C integer semantics.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/20194414372
Master (30s): https://github.com/esbmc/esbmc/actions/runs/20194414967

PR #3324, opened 2025-12-15
[CI] use cache to accelerate the build process

XLiZHI:
This PR has set up a cache for External Dependencies to avoid rebuilding them from scratch every time.

https://github.com/actions/cache

We will use the SHA values of all the CMakeLists in the project as identifiers. When there are any modifications to them, we will update the cache.


lucasccordeiro (approved):

lucasccordeiro:

PR #3328, opened 2025-12-16
[CI] optimize cache hits && switch to the appropriate runner

XLiZHI:
1. This PR has switched some GitHub host runners to the slim version. 
2. Remove the unnecessary packages to save disk space.

For some simple tasks with low load, we can use some runners with lower costs: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners

Use ccache to significantly accelerate the build process, following these rules:
1. Check if there are any code modifications (in *.cpp and *.h files) in the branch. If so, update the cache.
2. Only update the cache in the onpush workflow.

### DIFF
--- a/.github/workflows/benchexec-diff.yml
+++ b/.github/workflows/benchexec-diff.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   create-diff:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #3329, opened 2025-12-16
Fix handling for --is-instance-check

### LINKED ISSUE #3301: `--is-instance-check` for Python not handling optional keyword args correctly
For this program:
```python
def foo(*, a: int, b: str | None = None):
    pass

foo(a=0)
```
when I call `esbmc` with the `--is-instance-check` option, I get a verification failure:
```
State 1 file test.py line 3 column 19 function foo thread 0
Violated property:
  file test.py line 3 column 19 function foo
  Type annotation check for 'b'
  ISINSTANCE(b, {  })


VERIFICATION FAILED
```
However, I expected this program to pass because `b` should default to `None` when not provided which is of type `str | None`.

sergillam:
Fix Optional handling for --is-instance-check by registering kw-only param annotations before asserts, teaching both the converter and goto_check to treat None/Optional correctly, and adding pass/fail github_3301 regressions that cover optional keyword arguments

Fixes https://github.com/esbmc/esbmc/issues/3301.


lucasccordeiro (approved):

lucasccordeiro:

PR #3330, opened 2025-12-17
Builtin object size no longer relies on simplifier

rafaelsamenezes:
This PR:

- Make the object size builtin more consistent with the same behaviour without simplifications enabled.
- Duplicated a few tests with --no-simplify.
- Added the simplification needed as a unit test.


lucasccordeiro (approved):

lucasccordeiro:

PR #3331, opened 2025-12-18
Fix crash when processing Python dictionary len() operations

LukeW1999:
## Problem

ESBMC crashes when processing Python dictionaries with `dict[str, list[str]]` types when calling `len()` on them. The crash occurs during symbolic execution:

```
Assertion `sym.rlevel != symbol2t::renaming_level::level2_global' failed.
```

Root causes:
1. Python frontend generates `__ESBMC_get_object_size` calls for `len()` on dictionaries
2. Function names have Python-style prefix (`py:1.py@F@__ESBMC_get_object_size`) instead of C-style (`c:@F@__ESBMC_get_object_size`)
3. Symbolic execution engine doesn't recognize Python-style intrinsic function names
4. `value_set` analysis hits an overly strict assertion when processing dictionary member expressions with `level2_global` symbols

### 1. Function Name Conversion (`src/goto-symex/symex_main.cpp`)

Detect Python-style intrinsic function names (`py:*@F@__ESBMC_*`) and convert them to C-style format (`c:@F@__ESBMC_*`) before dispatching to `run_intrinsic`. Uses a whitelist to only convert known intrinsics, avoiding errors with non-intrinsic `__ESBMC_` functions like `__ESBMC_copy_value`.

### 2. Dictionary Size Calculation (`src/goto-symex/builtin_functions.cpp`, `src/goto-symex/goto_symex.h`)

Modified `intrinsic_get_object_size` to handle Python dictionary types. Dictionaries are `__python_dict__` structs with `keys` and `values` members. For `__ESBMC_get_object_size(dict)`, we extract the `keys` member and call `__ESBMC_list_size` to get the dictionary size. Extracted common logic into `handle_dict_size_for_get_object_size` helper function.

### 3. Value Set Assertion Fix (`src/pointer-analysis/value_set.cpp`)

Disabled the overly strict assertion that fails when processing Python dictionary operations. Added debug logging for `level2_global` symbols. This assertion was already known to fail in pthread scenarios; Python dictionary operations are another valid case.

## Testing

Added regression tests:
- `regression/python/github_3326/main.py`: Pass test - verifies `len()` on dictionaries works
- `regression/python/github_3326_fail/main.py`: Fail test - should catch incorrect assertions (marked as `KNOWNBUG`)

### Known Issue: Fail Test Incorrectly Passes

The `github_3326_fail` test incorrectly reports `VERIFICATION SUCCESSFUL` instead of `VERIFICATION FAILED`. This is a **separate semantic issue** unrelated to the crash fix.

**Test case:**
```python
l = f["fruit"]  # l = ["apple", "banana"]
for s in l:
    assert s == "wrong"  # Should fail, but passes
```

**Why it fails incorrectly:**

1. **Type information loss**: When accessing `l` from the dictionary, it's converted to `(void *)` pointer, losing Python list type information. ESBMC treats `l` as a generic pointer instead of a Python list.

2. **Missing constraints**: Array indexing `s = l[ESBMC_index_0]` doesn't generate constraints restricting `s` to actual list values ("apple" or "banana"). Without list type info, ESBMC can't track what's actually in the list.

3. **SMT solver finds invalid model**: The `strcmp(s, "wrong") == 0` assertion can be satisfied because `s` isn't constrained to list elements.

**GOTO program shows:**
```
ASSIGN l=(void *)((signed char *)dict_val_obj$39->value);
ASSIGN s=l[ESBMC_index_0];
ASSERT return_value$_strcmp$1 == 0  // strcmp(s, "wrong") == 0
```

The `void *` conversion loses type information, so the solver can assign any value to `s`.

**Impact:**
- Pre-existing issue in ESBMC's Python list handling, not introduced by this PR
- Crash fix is independent of this semantic issue
- Test marked as `KNOWNBUG` to document the limitation
- Should be fixed in a separate PR focusing on preserving list type information

## Files Changed

- `src/goto-symex/symex_main.cpp`: Python function name conversion
- `src/goto-symex/builtin_functions.cpp`: Dictionary size calculation
- `src/goto-symex/goto_symex.h`: Helper function declaration
- `src/pointer-analysis/value_set.cpp`: Fixed `level2_global` assertion
- `regression/python/github_3326/`: Pass test case
- `regression/python/github_3326_fail/`: Fail test case (KNOWNBUG)

## Do not Merge yet
 I think it should not be merged right now since it would secretly incorrectly pass failing case. I would try to fix it and update in this PR.

### DIFF
--- a/src/goto-symex/builtin_functions.cpp
+++ b/src/goto-symex/builtin_functions.cpp
@@ -2627,26 +2627,97 @@ void goto_symext::intrinsic_builtin_object_size(
     cur_state->guard);
 }
 
+bool goto_symext::handle_dict_size_for_get_object_size(
+  const code_function_call2t &func_call,
+  const expr2tc &dict_expr,
+  const struct_type2t &struct_type)
+{
+  if (struct_type.name != "__python_dict__")
+    return false;
+
+  // Find the "keys" member in the dictionary struct
+  for (size_t i = 0; i < struct_type.members.size(); ++i)
+  {
+    if (struct_type.member_names[i] == "keys")
+    {
+      // Create member expression for dict.keys and call __ESBMC_list_size
+      expr2tc keys_member = member2tc(struct_type.members[i], dict_expr, "keys");
+      expr2tc newcall = func_call.clone();
+      code_function_call2t &list_size_call = to_code_function_call2t(newcall);
+      list_size_call.function = symbol2tc(get_empty_type(), "c:@F@__ESBMC_list_size");
+      list_size_call.operands.clear();
+      list_size_call.operands.push_back(keys_member);
+      bump_call(list_size_call, "c:@F@__ESBMC_list_size");
+      return true;
+    }
+  }
+  return false;
+}
+
 void goto_symext::intrinsic_get_object_size(
   const code_function_call2t &func_call,
   reachability_treet &)
 {
   assert(func_call.operands.size() == 1 && "Wrong get_object_size signature");
   expr2tc ptr = func_call.operands[0];
+  cur_state->rename(ptr);
 
-  // Work out what the ptr points at.
+  // Handle Python dictionary types: dictionaries are represented as structs
+  // with "keys" and "values" members. For __ESBMC_get_object_size(dict), we
+  // return the size of the keys list.
+  if (is_struct_type(ptr->type))
+  {
+    const struct_type2t &struct_type = to_struct_type(ptr->type);
+    if (handle_dict_size_for_get_object_size(func_call, ptr, struct_type))
+      return;
+  }
+
+  // For pointer types, dereference to find the object
   internal_deref_items.clear();
   expr2tc deref = dereference2tc(get_empty_type(), ptr);
   dereference(deref, dereferencet::INTERNAL);
 
-  assert(is_array_type(internal_deref_items.front().object->type));
-  expr2tc obj_size =
-    to_array_type(internal_deref_items.front().object->type).array_size;
+  if (internal_deref_items.empty())
+  {
+    // No object found, return 0
+    expr2tc ret_ref = func_call.ret;
+    dereference(ret_ref, dereferencet::READ);
+    symex_assign(
+      code_assign2tc(ret_ref, gen_zero(ret_ref->type)),
+      false,
+      cur_state->guard);
+    return;
+  }
 
+  const auto &item = internal_deref_items.front();
+  expr2tc obj = item.object;
+
+  // Handle pointer to dictionary
+  if (is_struct_type(obj->type))
+  {
+    const struct_type2t &struct_type = to_struct_type(obj->type);
+    if (handle_dict_size_for_get_object_size(func_call, obj, struct_type))
+      return;
+  }
+
+  // Handle array types (original behavior)
+  if (is_array_type(obj->type))
+  {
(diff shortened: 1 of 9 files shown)


src/goto-symex/builtin_functions.cpp
```diff
@@ -2627,28 +2627,99 @@ void goto_symext::intrinsic_builtin_object_size(
...
+
+  // For pointer types, dereference to find the object
   internal_deref_items.clear();
   expr2tc deref = dereference2tc(get_empty_type(), ptr);
   dereference(deref, dereferencet::INTERNAL);
 
-  assert(is_array_type(internal_deref_items.front().object->type));
-  expr2tc obj_size =
-    to_array_type(internal_deref_items.front().object->type).array_size;
+  if (internal_deref_items.empty())
```
lucasccordeiro, line 2682: Can you produce a warning message here?

src/goto-symex/symex_main.cpp
```diff
@@ -332,12 +332,75 @@ void goto_symext::symex_step(reachability_treet &art)
     if (is_symbol2t(call.function))
     {
       const irep_idt &id = to_symbol2t(call.function).thename;
-      if (has_prefix(id.as_string(), "c:@F@__ESBMC"))
+      std::string id_str = id.as_string();
+
+      // Handle __ESBMC intrinsic functions with both C and Python prefixes.
+      // Python frontend generates function names like "py:1.py@F@__ESBMC_get_object_size"
+      // which need to be converted to "c:@F@__ESBMC_get_object_size" format.
+      if (has_prefix(id_str, "c:@F@__ESBMC"))
```
lucasccordeiro: This solution might slow down our symex.

I suspect this might be the cause: https://github.com/esbmc/esbmc/blob/master/src/python-frontend/preprocessor.py#L903.

Can you try to intercept this call at https://github.com/esbmc/esbmc/blob/master/src/python-frontend/function_call_builder.cpp#L56?

You can check some examples:

strings:
https://github.com/esbmc/esbmc/blob/master/src/python-frontend/function_call_builder.cpp#L228

lists:
https://github.com/esbmc/esbmc/blob/master/src/python-frontend/function_call_builder.cpp#L249

dictionaries:
https://github.com/esbmc/esbmc/blob/master/src/python-frontend/function_call_builder.cpp#L278

src/pointer-analysis/value_set.cpp
```diff
@@ -577,14 +577,30 @@ void value_sett::get_value_set_rec(
...
+     *    - regression/esbmc-unix/02_account_symbolic_06
+     *    - regression/esbmc/10_bicycle_01
+     *    - regression/nonz3/10_bicycle_02
+     * 2. When handling Python dictionary operations, member expressions
+     *    may contain level2_global symbols that are valid in the context
+     *    of nested structure access.
+     * We disable the assertion and continue processing, as level2_global
+     * symbols are handled correctly by the value set analysis.
      */
+    if (sym.rlevel == symbol2t::renaming_level::level2_global)
```
lucasccordeiro, line 593: This can be removed once you intercept the call in the function builder.

lucasccordeiro (changes requested):

src/goto-symex/symex_main.cpp
LukeW1999: Thank you, Sir @lucasccordeiro. I would avoid affecting the C part next time.

PR #3334, opened 2025-12-18
[Python verification] fix overload with different fields

### LINKED ISSUE #3286: [Python] Error when using union with different fields
For the following two-file program:

`ex.py`:
```python
import ll

b = ll.create("bar")
```
`ll.py`:
```python
from typing import Literal, overload

class Foo:
    def __init__(self) -> None:
        self.x = 5

class Bar:
    def __init__(self) -> None:
        pass

@overload
def create(s: Literal["foo"]) -> Foo:
    ...

@overload
def create(s: Literal["bar"]) -> Bar:
    ...
    
def create(s: str) -> Foo | Bar:
    if s == 'foo':
        return Foo()
    elif s == 'bar':
        return Bar()
    else:
        raise NotImplementedError("Unknown class")
```

ESBMC errors out:
```bash
$ esbmc ex.py 
Parsing ex.py
Converting
WARNING: Union with no recognized types, defaulting to pointer
Generating GOTO Program
GOTO program creation time: 0.719s
Unwinding loop 11 iteration 1   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 11 iteration 2   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
Unwinding loop 11 iteration 3   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 110 column 12 function strcmp
ERROR: Looking up index of nonexistant member "x" in struct/union "Bar"
Aborted (core dumped)
```
ESBMC version: 9381b01a5

XLiZHI:
fixes #3286

### DIFF
--- a/regression/python/github_3286/ll.py
+++ b/regression/python/github_3286/ll.py
@@ -1,0 +1,25 @@
+from typing import Literal, overload
+
+class Foo:
+    def __init__(self) -> None:
+        self.x = 5
+
+class Bar:
+    def __init__(self) -> None:
+        pass
+
+@overload
+def create(s: Literal["foo"]) -> Foo:
+    ...
+
+@overload
+def create(s: Literal["bar"]) -> Bar:
+    ...
+    
+def create(s: str) -> Foo | Bar:
+    if s == 'foo':
+        return Foo()
+    elif s == 'bar':
+        return Bar()
+    else:
+        raise NotImplementedError("Unknown class")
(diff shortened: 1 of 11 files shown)


brcfarias (approved):

lucasccordeiro (approved):

PR #3340, opened 2025-12-19
 [python-frontend] Support class imports from modules

### LINKED ISSUE #3338: [Python] Object not found error on a two-file program
For this two-file program:
`ex.py`:
```python
import ll

s = "foo"
c = ll.Foo(s)
```
`ll.py`:
```python
def f(b: bool) -> None:
    pass

class Foo:
    def __init__(self, s: str) -> None:
        self.s = s
```
ESBMC errors out:
```bash
$ esbmc ex.py 
Parsing ex.py
ERROR: Object "ll" not found.
```
ESBMC version: 99810b6c7

LukeW1999:
Fixes #3338 

## Summary
Add support for importing and using Python classes from modules in ESBMC's Python frontend.

## Changes
- Added `class_definition` structure to store class metadata (name, base classes, methods)
- Extended `module` class with `add_class()`, `get_class()`, and `add_classes()` methods
- Updated `create_module()` to parse `ClassDef` nodes from Python AST
- Modified type inference in `python_annotation.h` to handle class constructor calls
- Added `get_string_safe()` helper for robust JSON parsing

## How it works
When importing a module:
1. Parser extracts class definitions from AST JSON (name, bases, methods)
2. Classes are stored in an `unordered_set` for O(1) lookup
3. During type inference, `get_class()` is called for constructor expressions
4. Constructor calls return the class name as the inferred type

## Testing
Added 5 regression tests covering:
- Basic class import and constructor calls
- Class inheritance
- Class methods
- Submodule classes
- Null return type handling

### DIFF
--- a/regression/python/github_3338/ex.py
+++ b/regression/python/github_3338/ex.py
@@ -1,5 +1,4 @@
-d: dict[str, list[int]] = {'a': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
-l: list[int] = d['a']
-assert len(l) == 10
-for i in d['a']:
-    assert i <= 10
+import ll
+
+s = "foo"
+c = ll.Foo(s)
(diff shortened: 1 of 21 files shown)


lucasccordeiro (approved):

lucasccordeiro:

Issue #3234, opened 2025-12-02
[python] Failure on regression/quixbugs/bucketsort


brcfarias:
main.py:
```python
def bucketsort(arr, k):
    counts = [0] * k
    for x in arr:
        counts[x] += 1

    sorted_arr = []
    for i, count in enumerate(counts):
        sorted_arr.extend([i] * count)

    return sorted_arr

assert bucketsort([3, 11, 2, 9, 1, 5], 12) == [1, 2, 3, 5, 9, 11]
assert bucketsort([3, 2, 4, 2, 3, 5], 6) == [2, 2, 3, 3, 4, 5]
```

ESBMC (a04f7c43a8) output:
```bash
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing bucketsort/main.py

Type checking warning:

/tmp/esbmc-python-astgen-2e24-75db-a93c/parser.py:79: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  if isinstance(value_node, ast.Str):
/tmp/esbmc-python-astgen-2e24-75db-a93c/parser.py:81: DeprecationWarning: ast.Bytes is deprecated and will be removed in Python 3.14; use ast.Constant instead
  elif isinstance(value_node, ast.Bytes):
<unknown>:45: SyntaxWarning: invalid escape sequence '\d'
Converting
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
[1]    34033 IOT instruction (core dumped)  esbmc bucketsort/main.py
```

lucasccordeiro:
@brcfarias: I can't reproduce this crash. Shall we close this issue?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/tree/master/regression/quixbugs/bucketsort.

PR #3342, opened 2025-12-19
[python] Handle augmented assignments

brcfarias:
This PR fixes the handling of augmented assignments on subscript targets by lowering them to plain assignments in the preprocessor (e.g., counts[0] += 1 becomes counts[0] = counts[0] + 1).

Needed by #3234


lucasccordeiro (approved):

PR #3347, opened 2025-12-23
[symex] track and report trivially verified assertions

### LINKED ISSUE #3341: Feature request: option to keep simplified assertions
When an assertion expression can be simplified to true, ESBMC seems to drop the assertion altogether, e.g.
```python
x = 1
y = x
assert x == y
y += 1
assert x < y
y -= 1
assert x != y
```
ESBMC's output shows one property only:
```
$ esbmc ex.py --multi-property
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.692s
Symex completed in: 0.001s (44 assignments)
Slicing time: 0.000s (removed 43 assignments)
Generated 3 VCC(s), 1 remaining after simplification (1 assignments)
Solving claim 'assertion at file ex.py line 7 column 0' with solver Boolector 3.2.3
✗ FAILED: 'assertion at file ex.py line 7 column 0'

[Counterexample]


State 1 file ex.py line 7 column 0 thread 0
Violated property:
  file ex.py line 7 column 0
  assertion
  x != y

Properties: 1 verified, ✗ 1 failed
Solver: Boolector 3.2.3 • Decision procedure total time: 0.001s • Avg: 0.001s/property

VERIFICATION FAILED
```

Is it possible to add an option to keep the assertions to be able to determine from the output which assertions were checked?

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3341.

This PR adds tracking for assertions that are simplified to true during symbolic execution. These "simplified" assertions are now:
- Logged immediately when detected during symex.
- Counted separately in verification statistics.
- Displayed in the final properties summary.

This provides better visibility into which properties were proven via simplification rather than by SMT solver reasoning.

Changes:
- Add `simplified_claims` counter to `goto_symext` and `execution_statet`.
- Update `symex_resultt` to include simplified claim count.
- Extend SimpleSummary with `simplified_properties` tracking.
- Update `schedule_execution_statet` for simplified claims in `--schedule mode`.
- Add "✓ PASSED" log messages when assertions simplify to true.
- Include simplified count in symex completion and final summary output.

This is the new output:

image

Thanks to [zhassan-aws](https://github.com/zhassan-aws) for requesting this feature.

### DIFF
--- a/regression/esbmc-unix/simplifier/test.desc
+++ b/regression/esbmc-unix/simplifier/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --multi-property
-^Properties: 10 verified \✓ 1 passed, \✗ 9 failed$
+^VERIFICATION FAILED$
--- a/regression/python/any-example_fail/test.desc
+++ b/regression/python/any-example_fail/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.py
 --multi-property
-^Properties\: 2 verified, \✗ 2 failed$
+^VERIFICATION FAILED$
--- a/regression/python/cover5/test.desc
+++ b/regression/python/cover5/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.py
 --multi-property --unwind 10
-^Properties: 43 verified \✓ 38 passed, \✗ 5 failed$
+^VERIFICATION FAILED$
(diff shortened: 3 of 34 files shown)


lucasccordeiro:
I expect various regression tests to fail here because we have changed the multi-property output. Once the regression suite is run, I'll fix them before merging this PR.

PR #3356, opened 2025-12-26
[python-frontend] - add minimal Python str.split support 

sergillam:


src/python-frontend/function_call_builder.cpp
```diff
@@ -2,13 +2,104 @@
...
 #include <python-frontend/python_converter.h>
 #include <python-frontend/symbol_id.h>
 #include <python-frontend/type_utils.h>
 #include <util/arith_tools.h>
 
 #include <boost/algorithm/string/predicate.hpp>
 
+namespace
+{
+void copy_location_fields(const nlohmann::json &from, nlohmann::json &to)
```
lucasccordeiro: Shall we move this to json_utils?

src/python-frontend/function_call_builder.cpp
```diff
@@ -2,13 +2,104 @@
...
+    to["lineno"] = from["lineno"];
+  if (from.contains("col_offset"))
+    to["col_offset"] = from["col_offset"];
+  if (from.contains("end_lineno"))
+    to["end_lineno"] = from["end_lineno"];
+  if (from.contains("end_col_offset"))
+    to["end_col_offset"] = from["end_col_offset"];
+}
+
+bool extract_constant_string(
```
lucasccordeiro: Shall we move this to string_handler?

src/python-frontend/function_call_builder.cpp
```diff
@@ -2,13 +2,104 @@
...
+    {
+      out = var_value["value"]["value"].get<std::string>();
+      return true;
+    }
+  }
+
+  return false;
+}
+
+exprt build_split_list(
```
lucasccordeiro: Shall we move this to string_handler?

lucasccordeiro (changes requested):

I have just a few comments about the code organisation (i.e., move some methods to their respective handlers).

Also, can you add a test case with a symbolic string? You can use `nondet_str()` to obtain this symbolic string. Even if we don't handle it yet, we should at least raise an error message via log_error to inform the user that symbolic strings are unsupported.

src/python-frontend/function_call_builder.cpp
sergillam: Done

src/python-frontend/function_call_builder.cpp
sergillam: done

src/python-frontend/function_call_builder.cpp
sergillam: Done

src/python-frontend/function_call_builder.cpp
lucasccordeiro: Thank you.

src/python-frontend/function_call_builder.cpp
lucasccordeiro: Thank you.

src/python-frontend/function_call_builder.cpp
lucasccordeiro: Thank you.

lucasccordeiro (approved):

PR #3357, opened 2025-12-26
[python-frontend] - add Python str.replace support and regression tests

sergillam:
add Python str.replace support

### DIFF
--- a/regression/python/string-replace_fail/main.py
+++ b/regression/python/string-replace_fail/main.py
@@ -1,0 +1,6 @@
+def main() -> None:
+    s = "12.34".replace(".", "")
+    assert s == "12.34"
+
+
+main()
--- a/src/c2goto/library/python/string.c
+++ b/src/c2goto/library/python/string.c
@@ -280,14 +280,14 @@ __ESBMC_HIDE:;
     return (char *)s;
 
   // Get string lengths
-  int old_len = __python_strnlen_bounded(old_sub, 256);
-  int new_len = __python_strnlen_bounded(new_sub, 256);
-  int len_s = __python_strnlen_bounded(s, 1024);
+  size_t old_len = __python_strnlen_bounded(old_sub, 256);
+  size_t new_len = __python_strnlen_bounded(new_sub, 256);
+  size_t len_s = __python_strnlen_bounded(s, 1024);
 
   // Bound assumptions for ESBMC - limit string sizes to reasonable values
-  __ESBMC_assume(len_s >= 0 && len_s <= 1024);
-  __ESBMC_assume(old_len >= 0 && old_len <= 256);
-  __ESBMC_assume(new_len >= 0 && new_len <= 256);
+  __ESBMC_assert(len_s >= 0 && len_s <= 1024, "len_s bounds");
+  __ESBMC_assert(old_len >= 0 && old_len <= 256, "old_len bounds");
+  __ESBMC_assert(new_len >= 0 && new_len <= 256, "new_len bounds");
 
   if (old_len == 0)
   {
--- a/src/c2goto/library/python/string.c
+++ b/src/c2goto/library/python/string.c
@@ -249,6 +249,77 @@ __ESBMC_HIDE:;
   return -1;
 }
 
+char *__python_str_replace(
+  const char *s,
+  const char *old_sub,
+  const char *new_sub,
+  int count)
+{
+__ESBMC_HIDE:;
+  if (!s)
+    return (char *)s;
+
+  if (!old_sub || !new_sub)
+    return (char *)s;
+
+  if (count == 0)
+    return (char *)s;
+
+  if (old_sub[0] == '\0')
+  {
+    __ESBMC_assert(
+      0, "replace() with empty pattern not supported in minimal support");
+    return (char *)s;
+  }
+
+  char *buffer = __ESBMC_alloca(512);
+  int pos = 0;
+  int i = 0;
+  int remaining = count;
+  int old_len = strlen(old_sub);
+  int new_len = strlen(new_sub);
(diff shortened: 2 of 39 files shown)


src/python-frontend/function_call_builder.cpp
```diff
@@ -3,12 +3,113 @@
...
+    {
+      out = var_value["value"]["value"].get<long long>();
+      return true;
+    }
+  }
+
+  return false;
+}
+
+exprt build_replace_string(
```
lucasccordeiro: Shall we move this to string_handler?

src/python-frontend/function_call_builder.cpp
```diff
@@ -3,12 +3,113 @@
...
+    {
+      out = var_value["value"]["value"].get<std::string>();
+      return true;
+    }
+  }
+
+  return false;
+}
+
+bool extract_constant_integer(
```
lucasccordeiro: Shall we move this to json_utils?

src/python-frontend/function_call_builder.cpp
```diff
@@ -3,12 +3,113 @@
...
+#include <python-frontend/string_builder.h>
 #include <python-frontend/symbol_id.h>
 #include <python-frontend/type_utils.h>
 #include <util/arith_tools.h>
 
 #include <boost/algorithm/string/predicate.hpp>
 
+namespace
+{
+bool extract_constant_string(
```
lucasccordeiro: Shall we move this to string_handler?

regression/python/string-replace_fail/main.py
```diff
@@ -0,0 +1,6 @@
+def main() -> None:
```
lucasccordeiro, line 1: Can you add a test case with symbolic strings via `nondet_str()`? Even if we don't handle it yet, we should raise an error message via `log_error()` to inform the user that we don't support it yet.

lucasccordeiro (changes requested):

src/python-frontend/function_call_builder.cpp
sergillam: I think it’s better to keep this in json_utils, as before. These helpers resolve AST/variables via python_converter and get_var_value, so they belong to the JSON/AST utility layer. string_handler should stay focused on string operations over already-converted exprt, without coupling to AST resolution.

regression/python/string-replace_fail/main.py
sergillam: Done

src/python-frontend/function_call_builder.cpp
sergillam: done

src/python-frontend/function_call_builder.cpp
sergillam: done

src/python-frontend/function_call_builder.cpp
sergillam: done

src/c2goto/library/python/string.c
```diff
@@ -249,6 +250,146 @@ __ESBMC_HIDE:;
...
+  if (count == 0)
+    return (char *)s;
+
+  // Get string lengths
+  int old_len = strlen(old_sub);
+  int new_len = strlen(new_sub);
+  int len_s = strlen(s);
+
+  // Bound assumptions for ESBMC - limit string sizes to reasonable values
+  __ESBMC_assume(len_s >= 0 && len_s <= 1024);
```
lucasccordeiro: Can you replace __ESBMC_assume with __ESBMC_assert?

src/c2goto/library/python/string.c
```diff
@@ -249,6 +250,146 @@ __ESBMC_HIDE:;
...
+{
+__ESBMC_HIDE:;
+  if (!s || !old_sub || !new_sub)
+    return (char *)s;
+
+  if (count == 0)
+    return (char *)s;
+
+  // Get string lengths
+  int old_len = strlen(old_sub);
```
lucasccordeiro: Shall we replace all `int` with `size_t`?

lucasccordeiro (changes requested):

Can you also please check the warnings from codacy?

src/c2goto/library/python/string.c
sergillam: Done

src/c2goto/library/python/string.c
sergillam: __ESBMC_assume to __ESBMC_assert changes the meaning from “restrict the search space for verification” to “require this bound in all executions.” That can introduce false negatives and regressions if the tool cannot prove that len_s <= 1024 for all inputs, or if real inputs can exceed that limit. It only makes sense as an assert if the bound is a deliberate functional requirement of the model, not just a verification simplification.

sergillam: Done (altered)

src/c2goto/library/python/string.c
```diff
@@ -9,6 +11,21 @@ __ESBMC_HIDE:;
   return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
 }
 
+static int __python_strnlen_bounded(const char *s, int max_len)
```
lucasccordeiro, line 14: Do we really need this?

lucasccordeiro (changes requested):

src/c2goto/library/python/string.c
sergillam: Yes, we need it. __python_strnlen_bounded avoids unbounded strlen on symbolic strings and keeps the BMC search finite, while also enforcing NUL‑termination within the bound via __ESBMC_assert

PR #3361, opened 2025-12-28
[python] Fix list repetition with symbolic size

brcfarias:
Before this change we were crashing with:
```
Converting
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
```

This fix is needed by #3234


lucasccordeiro (approved):

lucasccordeiro:

PR #3365, opened 2025-12-28
Frontend Python function split

sergillam:
function string  split


src/python-frontend/json_utils.h
```diff
@@ -44,6 +46,19 @@ bool search_function_in_ast(const JsonType &node, const std::string &func_name)
   return false;
 }
 
+template <typename JsonType>
+void copy_location_fields(const JsonType &from, JsonType &to)
```
brcfarias, line 48: Consider reusing `python_converter::get_location_from_decl`.

brcfarias (changes requested):

sergillam:
Reviewed the failure: it’s an abstraction-level mismatch.

- Python treats strings as abstract objects, so accessing the null terminator (\0) beyond the logical length raises IndexError.

- ESBMC lowers strings to C byte arrays, where accessing \0 is a valid memory operation.

Because the string-split-count tests are white-box (meant to validate ESBMC’s internal memory model), they aren’t valid for a standard Python interpreter. I added try/except handling in string-split-count and string-split-count_2 to address this, since there are similar tests with the same issue.

src/python-frontend/json_utils.h
lucasccordeiro: @sergillam: I agree with @brcfarias here.

Can you please reuse `python_converter::get_location_from_decl`?

scripts/check_python_tests.sh
```diff
@@ -9,6 +9,7 @@ if [ -f "../esbmc-venv/bin/activate" ]; then
 fi
 
 all_passed=true
+failed_tests=()
```
lucasccordeiro, line 12: Can you document why these changes to check_python_tests.sh are needed?

When @brcfarias created this script, I didn't like it much. However, I later realised that this script is useful, since we can incorrectly label test cases. 

So, the goal of this script is to run the Python interpreter over our test cases to check whether we have incorrectly labelled our regression tests.

lucasccordeiro (changes requested):

scripts/check_python_tests.sh
sergillam: The idea is just to flag what failed to improve visualization

src/python-frontend/json_utils.h
sergillam: Done

lucasccordeiro:
Thanks for addressing our comments, @sergillam.

PR #3367, opened 2025-12-29
[build] enable ccache for CI

brcfarias:


brcfarias:
@XLiZHI I think we missed the step "In order to use ccache in your other steps, point the compiler to it" from https://github.com/hendrikmuhs/ccache-action

XLiZHI:
> @XLiZHI I think we missed the step "In order to use ccache in your other steps, point the compiler to it" from https://github.com/hendrikmuhs/ccache-action

We have already done it in cmake ;-)
https://github.com/esbmc/esbmc/blob/master/scripts/cmake/OptimizationCCache.cmake

The current issue is that in Onpush and PR, the ESBMC is built using different flags (debug and debugopt). This resulted in the situation where PRs seemed unable to benefit from caching (because they were different).The cache must be in the master branch in order to be applied to the PR.

It would be great if you have any optimization methods.

intrigus-lgtm:
How big is the speedup this provides?
Because in my experience compiling is not the bottleneck, but running the tests is.

scripts/build.sh
```diff
@@ -275,6 +275,14 @@ case $OS in
   *) echo "Unsupported OS $OSTYPE" ; exit 1; ;;
 esac || exit $?
 
+# Enable ccache if available.
+if command -v ccache >/dev/null 2>&1; then
```
lucasccordeiro, line 279: No workflow saves or restores a ccache directory, so this is a no-op in CI. Add a cache step, or the PR delivers nothing.

scripts/build.sh
```diff
@@ -275,6 +275,14 @@ case $OS in
   *) echo "Unsupported OS $OSTYPE" ; exit 1; ;;
 esac || exit $?
 
+# Enable ccache if available.
```
lucasccordeiro, line 278: This silently changes every local build with ccache installed, with no way to opt out. Gate it behind a build.sh flag.

lucasccordeiro (changes requested):

PR #3368, opened 2025-12-29
Add regression tests for ctest and pytest cases generation

prototypeC14:
This pr is for adding ctest and pytest cases generation.


lucasccordeiro (approved):

lucasccordeiro:

PR #3382, opened 2026-01-06
[python-frontend] Missing self Parameter in Standalone Constructor Calls

LukeW1999:
## Problem

Standalone constructor calls (e.g., `Positive(2)`) cause ESBMC verification failure:
```
VERIFICATION FAILED
```

Root cause: When no LHS variable exists, frontend doesn't add the required `self` parameter.

### 1. Create Temporary `self` for Standalone Calls

**File:** `function_call_expr.cpp`

When `!current_lhs` and argument count equals user-provided count:
- Create temporary symbol `$ctor_self$`
- Add declaration statement
- Insert as first argument

Argument count check prevents duplicate additions.

### 2. Strip Temporary `self` from Constructor Arguments

**File:** `function_call_expr.cpp`

When constructor is used as argument (e.g., `foo(datetime(2,3,4))`):
- Inner `datetime(2,3,4)` adds temporary `$ctor_self$`
- This `self` shouldn't be passed to outer function `foo`
- Extract common function `strip_ctor_self_parameters()` to remove all `$ctor_self$` parameters

### 3. Handle Return Statements

**File:** `python_converter.cpp`

For constructors in return statements (e.g., `return Positive(2)`):
- Strip any existing temporary `$ctor_self$`
- Add correct `self` parameter (the return temp variable)

## Testing

```bash
./src/esbmc/esbmc regression/python/class15/main.py
# Test case: foo(datetime(2, 3, 4))
```

I added the dataframe to class15 since the generation of the `self` might break its default return. It would be easier to find relative sample with class14.


lucasccordeiro (approved):

lucasccordeiro:

Issue #3236, opened 2025-12-02
[python] Failure on regression/quixbugs/levenshtein


brcfarias:
main.py
```python
def levenshtein(source, target):
    if source == '' or target == '':
        return len(source) or len(target)

    elif source[0] == target[0]:
        return levenshtein(source[1:], target[1:])

    else:
        return 1 + min(
            levenshtein(source,     target[1:]),
            levenshtein(source[1:], target[1:]),
            levenshtein(source[1:], target)
        )


assert levenshtein("electron", "neutron") == 3
assert levenshtein("kitten", "sitting") == 3
assert levenshtein("rosettacode", "raisethysword") == 8
assert levenshtein("abcdefg", "gabcdef") == 2
assert levenshtein("", "") == 0
assert levenshtein("hello", "olleh") == 4
```

ESBMC (a04f7c43a8) output:
```bash
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing levenshtein/main.py

Type checking warning:

<unknown>:45: SyntaxWarning: invalid escape sequence '\d'
Converting
ERROR: min() with more than 2 arguments not yet supported
```

lucasccordeiro:
@brcfarias: I can't reproduce this issue anymore.

Could you please check it on your end?

brcfarias:
I can't reproduce it either.

lucasccordeiro:
This one might go to the time-out category of your list.

brcfarias:
Closing this issue as it's passing now.

PR #3391, opened 2026-01-08
[python] Infer builtin call types in logical expressions

brcfarias:
Improve type inference for builtin calls (e.g., len) in logical expressions, ensuring statements like `len(a) or len(b)` are handled as int in the Python frontend.

Fix needed by #3236


lucasccordeiro (approved):

PR #3393, opened 2026-01-08
Feature: python string upper 

sergillam:
new PR to function string upper


regression/python/dict20/test.desc
```diff
@@ -1,4 +1,4 @@
 THOROUGH
 main.py
---unwind 6 --no-standard-checks
+--unwind 1 --no-standard-checks --no-unwinding-assertions
```
LukeW1999: Why this needs to be changed? But I believe you do it for some reason. Just clarify here.

LukeW1999:
Hi, @sergillam I really apperciate your code quality and consistency. LGTM.

But, could you add two of single char string examples? one passing and one failing. It should be also working but I just want to have a double check.

sergillam:
> Hi, @sergillam I really apperciate your code quality and consistency. LGTM.
> 
> But, could you add two of single char string examples? one passing and one failing. It should be also working but I just want to have a double check.

> Add Test

lucasccordeiro (approved):

lucasccordeiro:

PR #3403, opened 2026-01-12
Fix union type handling in pointer dereference and symbolic assignment

dan-eicher:
Another Claude found issue.

                                                                                                                        
  Summary:                                                                                                              
                                                                                                                        
  This patch adds proper support for union types in two areas where they were previously not handled correctly:         
                                                                                                                        
  1. Symbolic assignment (symex_assign.cpp): Added a case for constant_union2t in symex_assign_rec(). When assigning to a union, the code now correctly assigns through the active member by extracting the corresponding member from the RHS and recursing on the member expression.                                                                               
  2. Pointer dereference (dereference.cpp): Extended construct_struct_ref_from_dyn_offs_rec() to handle union types:    
    - Added explicit handling for unions where all members are at offset 0. When the target type matches the union, it  
  guards that the offset is zero. For composite union members, it recursively searches for matching types.              
    - Extended byte array reconstruction to support unions (previously only structs were handled). For unions read from 
  byte arrays, the first member is read at the given offset and wrapped in a constant_union2tc.                         
                                                                                                                        
  Problem:                                                                                                              
                                                                                                                        
  When dereferencing pointers to unions with dynamic offsets, ESBMC would fall through to incorrect code paths because  
  union types weren't explicitly handled. This could cause verification failures or incorrect results when analyzing    
  code that uses unions with pointer arithmetic or type punning.                                                        
                                                                                                                        
  Technical details:                                                                                                    
                                                                                                                        
  - Unions differ from structs in that all members share offset 0, rather than having sequential offsets                
  - The fix ensures union member access respects this by not incrementing offsets when iterating through union members  
  - A minor control flow fix changes else if (is_array_type(...)) to if (is_array_type(...)) to ensure array bounds     
  checking isn't skipped after code type checks                                                                         

```
 * Minimal reproducer for ESBMC union dereference bug.
 *
 * Before the fix in dereference.cpp and symex_assign.cpp, ESBMC could not
 * handle direct union assignment through pointers with symbolic/dynamic offsets.
 *
 * This test copies a union element at a symbolic index - the operation that
 * failed before the patch.
 */

#include <stdint.h>
#include <assert.h>

typedef union slot {
    int16_t s;   /* signed short */
    uint16_t r;  /* unsigned short (reference) */
} slot_t;

#define STACK_SIZE 8

slot_t stack[STACK_SIZE];

/* ESBMC nondet functions */
int16_t nondet_s2(void);
uint8_t nondet_u1(void);

int main(void) {
    /* Symbolic stack pointer */
    uint8_t sp = nondet_u1();
    __ESBMC_assume(sp >= 1);
    __ESBMC_assume(sp < STACK_SIZE);

    /* Symbolic value to store */
    int16_t val = nondet_s2();
    stack[sp - 1].s = val;

    /*
     * Direct union copy at symbolic offset.
     * This is the operation that failed before the patch - ESBMC's
     * construct_struct_ref_from_dyn_offs_rec() didn't handle union types,
     * causing verification to fail or produce incorrect results.
     *
     * Workaround was: stack[sp].r = stack[sp - 1].r;
     * But direct assignment should work:
     */
    stack[sp] = stack[sp - 1];

    /* Verify the copy worked */
    assert(stack[sp].s == val);
    assert(stack[sp].s == stack[sp - 1].s);

    return 0;
}
```

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -389,8 +389,7 @@ void goto_symext::symex_assign_rec(
     if (!the_union.datatype_members.empty())
     {
       const expr2tc &lhs_memb = the_union.datatype_members[0];
-      expr2tc rhs_memb = member2tc(
-        lhs_memb->type, rhs, the_union.init_field);
+      expr2tc rhs_memb = member2tc(lhs_memb->type, rhs, the_union.init_field);
       symex_assign_rec(lhs_memb, full_lhs, rhs_memb, full_rhs, guard, hidden);
     }
   }
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -382,6 +382,18 @@ void goto_symext::symex_assign_rec(
   {
     symex_assign_structure(lhs, full_lhs, rhs, full_rhs, guard, hidden);
   }
+  else if (is_constant_union2t(lhs))
+  {
+    // For unions, assign through the active member
+    const constant_union2t &the_union = to_constant_union2t(lhs);
+    if (!the_union.datatype_members.empty())
+    {
+      const expr2tc &lhs_memb = the_union.datatype_members[0];
+      expr2tc rhs_memb = member2tc(
+        lhs_memb->type, rhs, the_union.init_field);
+      symex_assign_rec(lhs_memb, full_lhs, rhs_memb, full_rhs, guard, hidden);
+    }
+  }
   else if (is_extract2t(lhs))
   {
     symex_assign_extract(lhs, full_lhs, rhs, full_rhs, guard, hidden);
--- a/src/pointer-analysis/dereference.cpp
+++ b/src/pointer-analysis/dereference.cpp
@@ -790,7 +790,8 @@ expr2tc dereferencet::build_reference_to(
       return expr2tc();
     /* here, both of them are code */
   }
-  else if (is_array_type(value)) // Encode some access bounds checks.
+
+  if (is_array_type(value)) // Encode some access bounds checks.
   {
     bool can_carry = is_pointer_type(deref_expr) &&
                      to_pointer_type(deref_expr->type).carry_provenance;
@@ -944,9 +945,7 @@ void dereferencet::build_reference_rec(
 
   // All accesses to code need no further construction
   if (is_code_type(value) || is_code_type(type))
-  {
     return;
-  }
 
   if (is_struct_type(type))
     flags |= flag_dst_struct;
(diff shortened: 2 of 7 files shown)


lucasccordeiro:
@dan-eicher: can you please add the test case that exposes the issue?

dan-eicher:
Think I did the pushing the regression test thing correctly, I'm just kind of the middleman between the daffy robots and this repo so...

src/pointer-analysis/dereference.cpp
```diff
@@ -1894,33 +1927,51 @@ void dereferencet::construct_struct_ref_from_dyn_offs_rec(
...
-        gen_long(array_offset->type, type_byte_size_bits(target_type)));
+      // We now have a vector of fields reconstructed from the byte array
+      expr2tc the_struct = constant_struct2tc(type, std::move(fields));
+      output.emplace_back(accuml_guard, the_struct);
+    }
+    else if (is_union_type(type))
+    {
+      // For unions from byte arrays, read the first member at offset
+      const union_type2t &uniontype = to_union_type(type);
+      if (!uniontype.members.empty())
```
XLiZHI, line 1961: As far as I know, the members of the union cannot be empty (the compiler will complain). I think we can remove this.

src/pointer-analysis/dereference.cpp
```diff
@@ -1878,6 +1877,40 @@ void dereferencet::construct_struct_ref_from_dyn_offs_rec(
     return;
   }
 
+  if (is_union_type(value->type))
+  {
+    // Handle union types. All members of a union are at offset 0.
+    // If target type matches, guard that offset is zero and return.
+    expr2tc tmp = value;
```
XLiZHI, line 1883: It seems that some operations are similar to structs. Can we compress/reuse some of the code?

src/goto-symex/symex_assign.cpp
```diff
@@ -367,6 +367,18 @@ void goto_symext::symex_assign_rec(
   {
     symex_assign_structure(lhs, full_lhs, rhs, full_rhs, guard, hidden);
   }
+  else if (is_constant_union2t(lhs))
+  {
+    // For unions, assign through the active member
+    const constant_union2t &the_union = to_constant_union2t(lhs);
+    if (!the_union.datatype_members.empty())
```
XLiZHI, line 389: same here

XLiZHI (commented):
Some minor comments.

lucasccordeiro:
@dan-eicher: can you also please check the CI? There is one regression test failing.

	389 - regression/esbmc/bitwise5 (Failed)                esbmc regression

lucasccordeiro:
@dan-eicher: The current master (796deb0bba13920a3ca8a0adc141b872bd471acd) correctly handles the program you reported in this issue.

````C
#include <stdint.h>
#include <assert.h>

typedef union slot {
    int16_t s;   /* signed short */
    uint16_t r;  /* unsigned short (reference) */
} slot_t;

#define STACK_SIZE 8

slot_t stack[STACK_SIZE];

/* ESBMC nondet functions */
int16_t nondet_s2(void);
uint8_t nondet_u1(void);

int main(void) {
    /* Symbolic stack pointer */
    uint8_t sp = nondet_u1();
    __VERIFIER_assume(sp >= 1);
    __VERIFIER_assume(sp < STACK_SIZE);

    /* Symbolic value to store */
    int16_t val = nondet_s2();
    stack[sp - 1].s = val;

    /*
     * Direct union copy at symbolic offset.
     * This is the operation that failed before the patch - ESBMC's
     * construct_struct_ref_from_dyn_offs_rec() didn't handle union types,
     * causing verification to fail or produce incorrect results.
     *
     * Workaround was: stack[sp].r = stack[sp - 1].r;
     * But direct assignment should work:
     */
    stack[sp] = stack[sp - 1];

    /* Verify the copy worked */
    assert(stack[sp].s == val);
    assert(stack[sp].s == stack[sp - 1].s);

    return 0;
}
````

**Comment by XLiZHI** [comment]
This PR:
```
  correct:           19372
    correct true:    11949
    correct false:    7423
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17202
  Score:             30745 (max: 61270)
https://github.com/esbmc/esbmc/actions/runs/20957912360
```

Master:
````
  correct:           19335
    correct true:    11917
    correct false:    7418
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17239
  Score:             30676 (max: 61270)
https://github.com/esbmc/esbmc/actions/runs/20957915435
````

lucasccordeiro:
Thanks for submitting this PR, @dan-eicher.

@XLiZHI: Thanks for reviewing it.

PR #3409, opened 2026-01-14
symbolic string variable

sergillam:
add symbolic string support in the Python frontend and fixes symbol/linking issues that caused “no body for function” warnings and regression failures.


LukeW1999:
@sergillam Thank you for propose this PR. But, this is quite large. Could you provide a some explanation over the problem and idea so it can save me sometime :) .

src/python-frontend/parser.py
```diff
@@ -40,6 +40,14 @@ def is_testing_framework(module_name):
     return module_name in testing_frameworks
 
 
+def is_frontend_builtin_module(module_name):
```
LukeW1999: I do not know a lot about this. Do we have the usage of 'import esbmc' in Python?

src/python-frontend/function_call_builder.cpp
```diff
@@ -884,9 +882,8 @@ exprt function_call_builder::build() const
         else if (!string_handler::extract_constant_string(
                    *sep_kw, converter_, separator))
         {
-          throw std::runtime_error(
-            "split() only supports constant string separators in minimal "
-            "support");
+          // If separator is not constant, use empty string (whitespace split)
+          separator = "";
```
LukeW1999, line 886: It looks very tricky. Would that cause false negative? I think I would ask Bruno to have a look here.

src/python-frontend/function_call_builder.cpp
```diff
@@ -898,9 +895,8 @@ exprt function_call_builder::build() const
         if (!string_handler::extract_constant_string(
               call_["args"][0], converter_, separator))
         {
-          throw std::runtime_error(
-            "split() only supports constant string separators in minimal "
-            "support");
+          // If separator is not constant, use empty string (whitespace split)
+          separator = "";
```
LukeW1999, line 899: same

src/python-frontend/function_call_builder.cpp
LukeW1999: Hi @brcfarias, Could you have a look arount his runtime spilit logic? Would this cause false negative?

src/python-frontend/string_handler.cpp
```diff
@@ -882,51 +882,248 @@ exprt string_handler::handle_char_isspace(
...
+
+      while (!chars.empty() && is_whitespace(chars.back()))
+        chars.pop_back();
+
+      return string_builder_->build_null_terminated_string(chars);
+    }
+  }
+
+  // If chars_arg is provided, use __python_str_strip_chars
+  if (!chars_arg.is_nil() && chars_arg.is_not_nil())
```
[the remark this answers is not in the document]
LukeW1999: Is that the same condition? Do we need to remove one?

src/c2goto/library/python/string.c
```diff
@@ -594,6 +759,27 @@ __ESBMC_HIDE:;
   return buffer;
 }
 
+// Python string split - splits a string by separator
+// Returns a Python list (represented as PyObject*)
+// For ESBMC, we'll return a simple structure representing the split result
+void *__python_str_split(const char *str, const char *sep, long long maxsplit)
+{
+__ESBMC_HIDE:;
+  // This is a placeholder implementation for symbolic execution
```
LukeW1999, line 768: Could you exaplain how the backend would handle this placeholder function?

brcfarias:
> @sergillam Thank you for propose this PR. But, this is quite large. Could you provide a some explanation over the problem and idea so it can save me sometime :) .

I agree on starting with some description of the problem and solution, preferably with an example to understand the context.

brcfarias:
Maybe it would make sense to split it into smaller PRs that fix smaller issues, and then reference an umbrella issue.

lucasccordeiro:
> Maybe it would make sense to split it into smaller PRs that fix smaller issues, and then reference an umbrella issue.

I agree with @LukeW1999 and @brcfarias.

@sergillam: The first PR could be the one with the regression tests. After this PR, you start submitting small PRs to fix the regression tests (passing from KNOWNBUG to CORE).

What do you think?

src/python-frontend/string_handler.cpp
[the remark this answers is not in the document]
sergillam: Removed

src/python-frontend/string_handler.cpp
```diff
@@ -882,51 +882,248 @@ exprt string_handler::handle_char_isspace(
...
+
+      return string_builder_->build_null_terminated_string(chars);
+    }
   }
-  else if (!str_expr.type().is_pointer())
+
+  // If chars_arg is empty, strip whitespace (default behavior)
+  std::vector<typet> arg_types = { pointer_typet(char_type()) };
+
+  if (!chars_arg.is_nil() && chars_arg.is_not_nil())
```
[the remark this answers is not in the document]
sergillam: Removed

lucasccordeiro (approved):

lucasccordeiro:

PR #3419, opened 2026-01-20
Add support for symbolic string split handling

sergillam:
1. Update github_3033/github_3034 regression tests to use symbolic strings for split cases.
2. Model symbolic split results so list indexing and empty-string checks behave correctly.
3. Fix type inference for dynamic list elements to avoid JSON crashes during conversion.


lucasccordeiro (approved):

There is some repetition in list.c and string.c. However, I'll submit another PR to improve this.

PR #3425, opened 2026-01-20
[C++ verification] pointer-to-member

XLiZHI:
This PR introduces two new ireps to handle pointer to member. 

We rely on constant propagation to perform the assignment of pointer members.

Support format:

Assignment:
`int S::*pm = &S::x;`
Access:
`s.*pm`

Assignment:
`int S::*pm = &S::x;`
`S *r = &s;`
Access:
`r->*pm`

### DIFF
--- a/src/irep2/irep2_expr.h
+++ b/src/irep2/irep2_expr.h
@@ -834,6 +834,27 @@ public:
   typedef esbmct::expr2t_traits<source_value_field, member_field> traits;
 };
 
+class member_ref_data : public datatype_ops
+{
+public:
+  member_ref_data(
+    const type2tc &t,
+    datatype_ops::expr_ids id,
+    const irep_idt &m)
+    : datatype_ops(t, id), member(m)
+  {
+  }
+  member_ref_data(const member_ref_data &ref) = default;
+
+  irep_idt member;
+
+  // Type mangling:
+  typedef esbmct::
+    field_traits<irep_idt, member_ref_data, &member_ref_data::member>
+      member_field;
+  typedef esbmct::expr2t_traits<member_field> traits;
+};
+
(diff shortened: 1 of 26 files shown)


lucasccordeiro (approved):
LGTM.

Anthonysdu (approved):

src/irep2/irep2_expr.h
```diff
@@ -2924,6 +2974,40 @@ class member2t : public member_expr_methods
...
+};
+
+/** Member pointer
+ *  @extends ptr_mem_data */
+class ptr_mem2t : public ptr_mem_expr_methods
+{
+public:
+  /** Primary constructor.
+   *  @param type Type of extracted member.
+   *  @param memb Name of member to extract.  */
```
intrigus-lgtm: Wrong parameter and missing parameter.

src/util/c_typecast.cpp
```diff
@@ -666,6 +666,17 @@ void c_typecastt::implicit_typecast_followed(
       return; // ok
     }
 
+    if (dest_type.find("to-member").is_not_nil())
+    {
+      if (src_type != dest_type)
+      {
+        // pointe-to-member
+        // TODO: take inheritance into account
```
intrigus-lgtm, line 674: Add a KNOWNBUG testcase for inheritance?

intrigus-lgtm (changes requested):
If you don't mind, I would like to understand this in more depth until the weekend.
Also, I think there could be a few more interesting testcases e.g. reassigning member pointers and stuff like that.

src/util/c_typecast.cpp
XLiZHI: It will be added after the feature is implemented.

PR #3437, opened 2026-01-24
Fix len(list) resolution for untyped variables in Python frontend

sergillam:
Fixes len() handling for untyped list variables in the Python frontend by resolving their actual symbol type so list lengths use __ESBMC_list_size instead of strlen


lucasccordeiro (approved):

It is very nice that you've fixed two benchmarks from the OpenAI dataset ;-)

PR #3445, opened 2026-01-26
Add Python string method support with regression coverage p2

sergillam:
This PR adds frontend support for several Python string methods and wires them into the call dispatcher. It also introduces regression tests covering common cases for these methods and updates the documentation to reflect the supported string operations


lucasccordeiro (approved):

lucasccordeiro:

PR #3455, opened 2026-01-28
[build] Clean up temp dirs on signal-based termination

pmatos:
SIGALRM (timeout) calls _exit() which skips static destructors, leaving orphaned esbmc-headers-* dirs in /tmp. Register temp directory paths globally and clean them in signal handlers before exiting.

I ran of /tmp space quite a few times while running tests with timeout. :)


lucasccordeiro (approved):

PR #3458, opened 2026-01-29
[Pytest] Add support for nondet_list and nondet_dict in pytest case generation

prototypeC14:
This PR adds support for `nondet_list` and `nondet_dict` in pytest case generation.


lucasccordeiro (approved):
@prototypeC14: Thanks for submitting this PR.

Could you please submit another PR to update the documentation with one running example of the newly supported feature?

PR #3462, opened 2026-01-30
Fix humaneval_6 in Python frontend

sergillam:
This PR fixes the Python frontend handling behind humaneval_6 by properly materializing literal split results, improving split typing in list comprehensions, and folding constant parse_nested_parens calls and literal list comparisons to avoid solver crashes/loops.


src/python-frontend/json_utils.h
```diff
@@ -196,6 +238,41 @@ const JsonType get_var_node(const std::string &var_name, const JsonType &block)
...
+    {
+      if (is_main_guard(element))
+      {
+        JsonType found = search_if_body(element);
+        if (!found.empty())
+          return found;
+      }
+      else
+      {
+        JsonType found = search_if_body(element);
```
LukeW1999, line 271: Need to remove the duplication here

LukeW1999 (approved):
LGTM, one duplication needs to be removed

PR #3463, opened 2026-01-30
[python] add listcomp for function call

LukeW1999:
Before the listcomp can only used static elements. now It also supports to use function inside.


src/python-frontend/function_call_expr.cpp
```diff
@@ -1771,6 +1702,60 @@ exprt function_call_expr::handle_list_append() const
...
+      // side_effect_expr_function_callt
+      const side_effect_expr_function_callt &call =
+        to_side_effect_expr_function_call(value_to_append);
+      func_expr = call.function();
+      func_args = call.arguments();
+      ret_type = call.type();
+    }
+
+    if (ret_type.is_nil() || ret_type.is_empty())
+      ret_type = int_type(); // fallback for untyped functions
```
lucasccordeiro: Can you please add a log_warning here? We need to warn the user if we fallback to int_type().

lucasccordeiro (approved):
Overall, LGTM.

I have just a very small comment.

src/python-frontend/function_call_expr.cpp
LukeW1999: log_warning added here.

lucasccordeiro: Thank you!

lucasccordeiro (approved):

PR #3466, opened 2026-01-30
feat: attribute `__ESBMC_EXTERN_NOVAL`

Yiannis128:
Adds the `__ESBMC_EXTERN_NOVAL` annotation to prevent ESBMC from assigning nondeterministic values to extern variables when their definitions exist elsewhere.

## Reason

By default, ESBMC assigns nondeterministic values to extern variables since it cannot see across translation units. However, when operational models define global variables (e.g., `timezone`, `daylight` in `time.h`), this behavior causes verification to fail because the extern declaration gets a nondet value instead of the actual defined value.

## Example use-case

```c
// header.h - operational model header
__ESBMC_EXTERN_NOVAL extern int counter;

// library.c - operational model implementation
int counter = 42;

// user_code.c
#include "header.h"
int main() {
    __ESBMC_assert(counter == 42, "Should pass");
    return 0;
}
```

Without the annotation, `counter` would be nondeterministic and the assertion could fail. With `__ESBMC_EXTERN_NOVAL`, ESBMC uses the actual defined value.


intrigus-lgtm (changes requested):
Please provide a complete testcase that actually shows a problem.
Your existing testcase works fine on the latest master I have on my machine.

Yiannis128:
> Please provide a complete testcase that actually shows a problem. Your existing testcase works fine on the latest master I have on my machine.

You want to see a situation where this feature is needed?

Yiannis128:
> Please provide a complete testcase that actually shows a problem. Your existing testcase works fine on the latest master I have on my machine.

Hi, I've pushed an additional test that demonstrates this. I found that this behaviour (automatic nondet) happens only in operational models since header global variables marked as extern are assumed as nondet. So `__ESBMC_EXTERN_NOVAL` is only valid for operational model use. In non-OMs it doesn't do anything.

lucasccordeiro (approved):

PR #3471, opened 2026-02-01
[python] optimize list compare

LukeW1999:
**Description:**

Optimize `__ESBMC_list_eq` and value comparison to reduce state explosion and speed up list-eq.

**Changes:**
- Make `__ESBMC_list_eq` iterative with bounded depth (`__ESBMC_LIST_MAX_DEPTH=4`) instead of recursive.
- Add 1/2/4-byte fast paths in `__ESBMC_values_equal`.
- Bump unwind: dict46 3→6, list18 1→3 (needed for the new loop, the number might be able to be smaller, I did not try to optimize it).

**list-eq regression (ctest -R list-eq, on my device):**

|        | This PR | master |
|--------|---------|--------|
| Total  | ~435 s  | ~707 s |
| list-eq3 | 227 s | 414 s |
| list-eq3_fail | 154 s | 220 s |
| list-eq1 | 8.6 s | 14.0 s |
| list-eq2 | 7.4 s | 14.1 s |
| list-eq9 | 7.3 s | 14.2 s |

Some particularly examples, which originally took less than 1 second, became slower (0.5s -> 1s around).

---

## Todo
__ESBMC_LIST_MAX_DEPTH might result in wrong result for depth >4 example. I want to introduce a list comparison depth configurable option with one unwinding assertion to its end.


src/c2goto/library/python/list.c
```diff
@@ -8,6 +8,28 @@
...
+__ESBMC_values_equal(const void *a, const void *b, size_t size)
+{
+  if (a == b)
+    return true;
+  // Direct comparison for common sizes - no loop needed
+  if (size == 8)
+    return *(const uint64_t *)a == *(const uint64_t *)b;
+  if (size == 4)
+    return *(const uint32_t *)a == *(const uint32_t *)b;
+  if (size == 2)
```
lucasccordeiro: Do we have cases that we need size 2 and 4?

src/c2goto/library/python/list.c
```diff
@@ -8,6 +8,28 @@
...
+    return *(const uint16_t *)a == *(const uint16_t *)b;
+  if (size == 1)
+    return *(const uint8_t *)a == *(const uint8_t *)b;
+  // Fallback for larger/unusual sizes
+  return memcmp(a, b, size) == 0;
+}
+
+// Maximum nesting depth to prevent state explosion
+#define __ESBMC_LIST_MAX_DEPTH 4
```
lucasccordeiro: Can you add an assertion that checks whether we need this depth greater than 4?

lucasccordeiro (changes requested):

src/c2goto/library/python/list.c
LukeW1999: No, we do not. We map to size 8. I would remove it. Thank a lot, @lucasccordeiro

src/c2goto/library/python/list.c
LukeW1999: Sure, I am doing and would request review when I implement it.

LukeW1999:
Follow-up changes:
1. Added --python-list-compare-depth option to configure max nesting depth (default: 4)
2. Added depth limit assertion with erro message when exceeded (` list comparison depth limit exceeded (use --python-list-compare-depth to increase)`)
3. Removed size == 2 and size == 4 fast paths (not used by Python frontend)
4. Added regression tests: list_depth_test (pass,~3 sec) and list_depth_test_fail (THOROUGH, ~2min)
Note: When depth limit is exceeded, verification takes longer (~2min) due to BMC needing to find a counterexample.

lucasccordeiro (approved):

PR #3474, opened 2026-02-02
[util] avoid compiler limit: blocks nested too deeply

XLiZHI:
In "migrate" method, the following structure was used:

```cpp
if (id == 1) 
{
  ...
}
else if (id == 2) 
{
  ...
} 
else if (id == 3) 
{
  ...
} 
```
As we introduce more ireps, this structure reaches the limit of the compiler's capabilities:  `error C1061: compiler limit: blocks nested too deeply`

Therefore, in this PR, the structure is modified as follows:

```cpp
if (id == 1) 
{
  ...
  return;
}

if (id == 2) 
{
  ...
  return;
} 
```

### DIFF
--- a/src/util/migrate.cpp
+++ b/src/util/migrate.cpp
@@ -606,18 +606,24 @@ void migrate_expr(const exprt &expr, expr2tc &new_expr_ref)
   if (expr.id() == "nil")
   {
     new_expr_ref = expr2tc();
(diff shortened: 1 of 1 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #3475, opened 2026-02-02
Fix humaneval_6-1 by folding parse_nested_parens on literals

sergillam:
This PR resolves the humaneval_6-1 failure caused by string concatenation overflow inside the loop. It evaluates parse_nested_parens at compile time when the input is a string literal, producing a literal list directly in the fronten


lucasccordeiro (approved):

PR #2575, opened 2025-07-03
[C verification] fix external var init

XLiZHI:
In this PR:
1. Assign nondet to the uninitialized external variable.
2. I think constant propagation is still very important for assume.

` ASSIGN x=NONDET(signed int);`

PR #3484, opened 2026-02-04
[C/CPP] Linking and extern cleanup

intrigus-lgtm:
Commit by commit review encouraged.

One foot gun that this includes is that __ESBMC_deallocated and friends won't be properly initialized when running in `--no-library` mode.
This should be solvable by splitting the libc-like cprover binary into two parts, one part that is the existing libc-like part and the other part is the one that just initializes __ESBMC_deallocated, __ESBMC_rounding_mode and friends without which the core functionality wouldn't work.
Left as future work or until someone encounters this problem in practice which should be quite rare :)

Also more or less reverts https://github.com/esbmc/esbmc/pull/2575 and moves this part to the stage where we "link to main".
This allows us to properly log warnings for `extern` symbols for which we didn't get a definition.
We initialize those symbols then with nondet even when an actual compiler would reject such files.

Currently, one limitation of our approach of using AstImporter to merge multiple source files is that the importer is quite liberal and doesn't complain about e.g. duplicate global symbols.
It would be more C like to actually produce one "object" file for every input file and then link those together like an actual compiler would do.

### DIFF
--- a/src/c2goto/cprover_library.cpp
+++ b/src/c2goto/cprover_library.cpp
@@ -1,4 +1,7 @@
 #include <c2goto/cprover_library.h>
+
+#include "util/symbolic_types.h"
+
 #include <ac_config.h>
 #include <boost/filesystem.hpp>
 #include <cstdlib>
@@ -440,4 +443,33 @@ void add_cprover_library(contextt &context, const languaget *language)
     log_error("Failed to merge C library");
     abort();
   }
+  // We basically need a place where we know that ESBMC produces the "main" executable that will be run.
+  // This is the best place that I've found and mimics how a real compiler would work:
+  // First compile all source files to objects files, then link them together and then link with the libc
+  // library. Only when linking to the libc library, we know that all unresolved extern symbols (those whose
+  // value is nil) will stay unresolved. A normal linker would reject such files, but we provide some compatibility with
+  // those and initialize the extern variables to nondet.
+  context.Foreach_operand(
+    [&context](symbolt &s)
+    {
+      if (s.is_extern && !s.type.is_code())
+      {
+        if (s.value.is_nil())
+        {
+          log_warning(
+            "extern variable with id {} not found, initializing value to nondet! This code would not compile with an actual compiler.",
+            s.id);
+          exprt value =
+            exprt("sideeffect", get_complete_type(s.type, namespacet{context}));
+          value.statement("nondet");
+          s.value = value;
+        }
+        else
+        {
+          log_error("extern variable with id {} is not nil.", s.id);
+          s.dump();
+          abort();
+        }
+      }
+    });
 }
--- a/src/c2goto/cprover_library.cpp
+++ b/src/c2goto/cprover_library.cpp
@@ -449,27 +449,26 @@ void add_cprover_library(contextt &context, const languaget *language)
   // library. Only when linking to the libc library, we know that all unresolved extern symbols (those whose
   // value is nil) will stay unresolved. A normal linker would reject such files, but we provide some compatibility with
   // those and initialize the extern variables to nondet.
-  context.Foreach_operand(
-    [&context](symbolt &s)
+  context.Foreach_operand([&context](symbolt &s) {
+    if (s.is_extern && !s.type.is_code())
     {
-      if (s.is_extern && !s.type.is_code())
+      if (s.value.is_nil())
       {
-        if (s.value.is_nil())
-        {
-          log_warning(
-            "extern variable with id {} not found, initializing value to nondet! This code would not compile with an actual compiler.",
-            s.id);
-          exprt value =
-            exprt("sideeffect", get_complete_type(s.type, namespacet{context}));
-          value.statement("nondet");
-          s.value = value;
-        }
-        else
-        {
-          log_error("extern variable with id {} is not nil.", s.id);
-          s.dump();
-          abort();
-        }
+        log_warning(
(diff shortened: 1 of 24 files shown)


Yiannis128:
Great job, looking forward to this getting this merged 😀

src/c2goto/library/builtin_libs.c
```diff
@@ -243,3 +243,10 @@ __SIZE_TYPE__ __esbmc_cheri_length_get(void *__capability cap)
 }
 
 #endif
+
+__attribute__((annotate("__ESBMC_inf_size"))) _Bool __ESBMC_alloc[1];
```
XLiZHI, line 247: https://github.com/esbmc/esbmc/blob/3558e59f83940cd16fb3282f0f2a0a05a6a94192/src/clang-c-frontend/clang_c_language.cpp#L380-L388

We have already declared these variables. What reason is there for doing so?

Is it just to make it easier for other front-end to declare it?

src/c2goto/cprover_library.cpp
```diff
@@ -440,4 +443,32 @@ void add_cprover_library(contextt &context, const languaget *language)
...
+          s.id);
+        exprt value =
+          exprt("sideeffect", get_complete_type(s.type, namespacet{context}));
+        value.statement("nondet");
+        s.value = value;
+      }
+      else
+      {
+        log_error("extern variable with id {} is not nil.", s.id);
+        s.dump();
```
XLiZHI, line 469: For the users, I think it's not appropriate to dump here.

src/util/c_link.cpp
```diff
@@ -53,18 +53,6 @@ class c_linkt : public typecheckt
```
XLiZHI, line 53: I'm not sure which other places we still used c_link. We replaced it with astimporter.

src/c2goto/library/builtin_libs.c
intrigus-lgtm: The declarations in `esbmc_intrinsics.h` are just `extern` and we have to provide an actual definition/storage for them. Otherwise they would be undefined and initialized to nondet.

The reason for the split is that otherwise you would run into problems when merging the cprover-libc and the program you're trying to verify:
both (previously) defined the same global symbol which is not allowed.

src/util/c_link.cpp
intrigus-lgtm: AstImporter is not really good.
It happily merges duplicate symbols which should fail.

I also quite like the architecture that old CBMC used:
you can drop in a goto-cc in the normal compilation process and get a foo.gb (goto binary) as a result which you can then perform BMC on.
It would hugely simplify using ESBMC if you could use it as a drop-in replacement as well.
For that, you'd need to merge multiple invocations that produced a goto binary and then merge them using c_link.

src/c2goto/cprover_library.cpp
intrigus-lgtm: I don't feel very strongly about this and didn't check what exactly the other error messages do and I kinda assume it's a wild mix, but at least for bug reports, this is helpful IMHO.
Your decision.

intrigus-lgtm:
Moved the manual initialization of ESBMC_alloc and friends back to python and jimple due to the way those two frontends work. They do not automatically pick up those symbols and cleaning up the python frontend turned out to be more work/restructuring than anticipated.

intrigus-lgtm:
Interestingly, https://github.com/esbmc/esbmc/pull/3484/commits/c5ece145a17cf9a0f9139469b096919efe760c91 works when running on my Debian 13 machine, but not in CI 🙃 
Reverted the changes for now.

src/util/c_link.cpp
rafaelsamenezes: > AstImporter is not really good.

In some edge cases it help us. IIRC with flexible array members and some C++ cases.

> It happily merges duplicate symbols which should fail.

Isn't that an upstream bug?

> I also quite like the architecture that old CBMC used:

Yeah, I also have some plans on making ESBMC behave a bit more like a compiler. A goto-cc and goto-linker would improve the flow a lot.

intrigus-lgtm:
~~@lucasccordeiro what's the best way to triage the results of a benchexec run?~~
I just realized that my benchexec run is better than the existing run, so ignore.

lucasccordeiro:
Super!

What are the links for your run? I can extract the data.

intrigus-lgtm:
> Super!
> 
> What are the links for your run? I can extract the data.

https://github.com/esbmc/esbmc/actions/runs/21725676173

lucasccordeiro:
Your PR:

````
  correct:           19358
    correct true:    11961
    correct false:    7397
  incorrect:            27
    incorrect true:     11
    incorrect false:    16
  unknown:           17215
  Score:             30711 (max: 61270)
````

Master (30s): https://github.com/esbmc/esbmc/actions/runs/21738493620/job/62709718922

````
  correct:           19309
    correct true:    11922
    correct false:    7387
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17265
  Score:             30655 (max: 61270)
````

lucasccordeiro:
@intrigus-lgtm: I've asked @rafaelsamenezes to review your PR. 

@rafaelsamenezes: Did you have a chance to check it?

rafaelsamenezes (approved):

lucasccordeiro (approved):

lucasccordeiro:

Issue #3479, opened 2026-02-03
`nondet_dict` and `nondet_list` fail to generate sets with multiple elements.


prototypeC14:
`nondet_dict(n)` and `nondet_list(n)` generate sets with at most one unique element, no matter how large n is set.
For example:
```python
def test_dict_size():
    d = nondet_dict(3)  # Expected: a dictionary with up to 3 entries
    # Assertion: the dictionary should be able to have 2 or more entries
    assert len(d) < 2  

def test_list_elements():
    lst = nondet_list(3)  # Expected: a list with up to 3 elements

    if len(lst) >= 2:
        # Assertion: the first two elements of the list should be able to differ
        assert lst[0] == lst[1]  # Expect this assertion to be violated
```
Expected behavior:
- test_dict_size: ESBMC should find a counterexample, e.g. d = {1: 10, 2: 20}, which makes count >= 2.
- test_list_elements: ESBMC should find a counterexample, e.g. lst = [5, 3], which makes lst[0] != lst[1].

Actual behavior
- tests report VERIFICATION SUCCESSFUL, because:
- The dictionary always has only one entry (no matter how large size is).
- All elements in the list are always identical.

PR #3485, opened 2026-02-05
[Python] Fix the issue where nondet_dict and nondet_list fail to generate sets with multiple elements.

prototypeC14:
see #3479


src/python-frontend/models/nondet.py
```diff
@@ -88,18 +94,41 @@ def nondet_dict(max_size: int = _DEFAULT_NONDET_SIZE,
...
+        if key_type is None:
+            k: int = nondet_int()
+        else:
+            k1: Any = key_type
+        # TODO here we should do like
+        # elif isinstance(key_type, bool):
+        #   k_bool: Any = nondet_bool()
+        # elif isinstance(key_type, float):
+        #   k_float: Any = nondet_float()
+        # but for now we dont support isinstance and int/bool... keys,
```
lucasccordeiro: Can you open an issue for this?

lucasccordeiro (changes requested):

src/python-frontend/models/nondet.py
prototypeC14: Sure.
Just to clarify: isinstance(...) is already supported now, but identity-style checks like x is int (or x is bool) are not supported at the moment.

lucasccordeiro: No worries.

Can you please open an issue with an example that exposes the problem?

prototypeC14: Yes, here is the issue #3520

lucasccordeiro: Thank you.

lucasccordeiro (approved):

PR #3489, opened 2026-02-06
BuiltinTemplate should not trigger errors

rafaelsamenezes:
BuiltinTemplates are intrinsics that are expanded by clang at compilation. The latest libc++ standard from Ubuntu 24.04 relies heavily on them, so ESBMC is crashing on tons of CPP cases for no reason.

So I removed the abort and left a warning, this is just to remind us that this is broken if we want to use thoose intrinsics with nondet values.


lucasccordeiro (approved):

lucasccordeiro:

PR #3491, opened 2026-02-06
Update CPP signature for tuples and arrays

rafaelsamenezes:
This fixes and improve compatibility with libstdc++ from Ubuntu 24.04.


lucasccordeiro (approved):

PR #3492, opened 2026-02-06
feat: safer config file loading

Yiannis128:
This PR adds type checking for config fields. This means this is now not possible anymore (which would have silently failed):

```toml
memory-leak-check = "true"
```

It will fail with a message:

```
config: key 'memory-leak-check' is a boolean flag but got string value "true"
```

Additionally, this PR adds verbose logging of the form:

```
[CONFIG] loaded memory-leak-check = true
[CONFIG] loaded k-step = 2
```

When loading values from a config file. So `ESBMC_CONFIG_FILE=config.toml esbmc ....`. This is an important feature because the users should be made aware what configuration their running. If running from the commandline they know because they typed in the flags they want to use, but config file is different. There's scenarios where a config file might be in use and not obvious (set in the ~/.profile etc).

This will also help bug reports since users may report the command they used, but not supply the contents of their config. This will make it more clear.

If no config is used, then this won't change the output at all.

### DIFF
--- a/src/util/config_file.cpp
+++ b/src/util/config_file.cpp
@@ -1,7 +1,7 @@
 #include "config_file.h"
 
 #include <boost/program_options/option.hpp>
-#include <stdexcept>
+#include <cstdlib>
 #include <string>
 #include <set>
 #include <fmt/core.h>
@@ -46,8 +46,8 @@ boost::program_options::basic_parsed_options<char> parse_toml_file(
   }
   catch (const toml::parse_error &err)
   {
-    throw std::runtime_error(
-      fmt::format("config: error parsing TOML file: {}", err.what()));
+    log_error("config: error parsing TOML file: {}", err.what());
+    abort();
   }
 
   // Get all the long option names and use those as allowed options.
@@ -94,13 +94,16 @@ boost::program_options::basic_parsed_options<char> parse_toml_file(
       case toml::node_type::string:
       {
         if (bool_flag_options.count(key_name))
-          throw std::runtime_error(fmt::format(
+        {
+          log_error(
             "config: key '{}' is a boolean flag but got string value \"{}\""
             " (use {} = true instead of {} = \"true\")",
             key_name,
             value_node->as_string()->get(),
             key_name,
-            key_name));
+            key_name);
+          abort();
+        }
 
         add_option(key_name, value_node->as_string()->get(), true);
         break;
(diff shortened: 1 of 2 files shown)


src/util/config_file.cpp
```diff
@@ -73,48 +78,57 @@ boost::program_options::basic_parsed_options<char> parse_toml_file(
     if (tbl.contains(key_name))
     {
       auto value_node = tbl.get(key_name);
+
+      auto add_option = [&](const std::string &key, const std::string &val,
+                            bool quote = false) {
+        if (quote)
+          log_status("[CONFIG] loaded {} = \"{}\"", key, val);
```
rafaelsamenezes: How about `log_debug`?

src/util/config_file.cpp
```diff
@@ -73,48 +78,57 @@ boost::program_options::basic_parsed_options<char> parse_toml_file(
...
       {
       case toml::node_type::string:
       {
-        const auto value = value_node->as_string()->get();
-        // For some reason takes it in as an array of values.
-        const auto option = boost::program_options::option(
-          key_name, std::vector<std::string>(1, value));
-        result.options.push_back(option);
+        if (bool_flag_options.count(key_name))
+          throw std::runtime_error(fmt::format(
```
rafaelsamenezes: How about `log_error` + abort?

rafaelsamenezes (changes requested):
Just minor things:

- The messages should go for debug mode. 
- Avoid exceptions that are not going to be treated.

src/util/config_file.cpp
Yiannis128: Hey so I looked into this (even though I think verbose messaging would be better) and I found out it's literally impossible to use `log_debug` because:

1. `cmdline.parse()` loads config file via `parse_toml_file()`
2. `set_verbosity_msg()` sets verbosity level from `--verbosity flag`

So by the time verbosity is configured, the config loading has already happened. The default verbosity is `Status`, meaning `log_debug` messages will never be visible for config loading not even with `--verbosity 8`.

src/util/config_file.cpp
Yiannis128: Let me know if you prefer this, it makes the test for this kinda hacky. I'm pushing the changes now.

Edit: Or would it be better to switch to regression tests... I might do that actually...

Yiannis128: I had to remove the regression tests because they don't support setting env variables.

lucasccordeiro (approved):

lucasccordeiro:

PR #3493, opened 2026-02-06
[simplifier] fix member access simplification through bitcast expressions

### LINKED ISSUE #3487: Missing member(byte_update) simplification causes incorrect verification of pointer dereference
Yet another Claude found bug with, maybe, the source of said bug:

ESBMC incorrectly reports VERIFICATION FAILED on a correct program. When a struct is zero-initialized with `= {0}` and then `memcpy` writes into its byte array member, a subsequent dereference through the struct's pointer member produces an incorrect result.

## Minimal reproducer

```c
#include <assert.h>
#include <stdint.h>
#include <string.h>

typedef struct {
    uint8_t mem[2];
    uint16_t* ptr;
} box_t;

int main(void) {
    box_t b = {0};
    uint16_t val = 42;
    b.ptr = &val;

    int16_t v = 1;
    memcpy(b.mem, &v, 2);

    assert(*b.ptr == 42);
    return 0;
}
```

## Expected

`VERIFICATION SUCCESSFUL` — `b.ptr` points to `val` which is 42, and neither the `= {0}` nor the `memcpy` into `b.mem` can affect `val` or the pointer itself.

## Actual

`VERIFICATION FAILED` — ESBMC claims `*b.ptr == 42` does not hold.

## Key observations

- **Removing `= {0}`** (e.g. declaring `box_t b;` then `memset(&b, 0, sizeof(b));`) → passes
- **Removing the `memcpy`** → passes
- **Moving the `memcpy` target to a separate buffer** → passes
- Both conditions (`= {0}` init AND `memcpy` into the byte array member) are required to trigger the bug
- Reproduces with no flags: `esbmc test.c`

## Root cause

The bug is in `member2t::do_simplify()` in `src/util/simplify_expr2.cpp` (around line 973).

This function handles member extraction from structs. It has cases for:
1. `constant_struct2t` / `constant_union2t` — extracts field from a constant struct
2. `with2t` — extracts field from a struct-update expression

But there is **no case for `byte_update2t`**. The execution flow:

1. `box_t b = {0}` creates a `constant_struct2t`
2. `memcpy(b.mem, &v, 2)` wraps it in a `byte_update2t` (updating bytes at the `mem` field's offset)
3. Dereferencing `*b.ptr` creates `member2t(byte_update2t(...), "ptr")`
4. `do_simplify()` checks for `constant_struct2t` → no; checks for `with2t` → no; returns empty
5. The unsimplified `member(byte_update)` expression is passed to the SMT solver, which produces wrong results

The `with2t` case already has the right logic pattern: it checks whether the updated member matches the extracted member, and if not, recurses into the source value. The same approach should work for `byte_update2t`: if the byte update's offset range doesn't overlap the member being extracted, simplify to `member(source_value)`.

Possibly related to #3418.

## Version

ESBMC 8.0.0 (commit 3558e59f83940cd16fb3282f0f2a0a05a6a94192 on master)

[esbmc_minimal7.c](https://github.com/user-attachments/files/25114857/esbmc_minimal7.c)

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3487.

This PR adds a `bitcast` case that traverses `bitcast(byte_update(...))` to extract members of the original struct before the byte update.

Thanks to [dan-eicher](https://github.com/dan-eicher) for reporting this issue.


lucasccordeiro:
Master (30s): https://github.com/esbmc/esbmc/actions/runs/21760966082
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/21760971632

intrigus-lgtm:
Why is this simplification necessary?
That is, shouldn't ESBMC still work without this simplification?
Do the TCs work when being run with `--no-simplify`?

lucasccordeiro:
Hi @intrigus-lgtm,

I’ll do my best to address your questions below. However, if you find a better solution, I would be happy to accept your pull request.

> Why is this simplification necessary?

The simplifier was generating incorrect results!

When `member2tc(bitcast(...), "ptr")` calls `simplify()`, `member2t::do_simplify()` returns empty for a bitcast. Meanwhile, the parent continues to simplify its sub-expressions. This results in the member access being partially unsimplified, producing a malformed expression that reaches the solver as `symex::invalid_object0&0#0`. 

Yes, we could trace back a few steps to identify where this malformed expression was created. As I said before, I'm happy to accept a PR, where you fix it at earlier stages at our simplification pipeline.

> Shouldn't ESBMC still function without this simplification?

When using `--no-simplify`, the entire expression tree remains unchanged and consistent! 

This allows our SMT backend to translate something like `member(bitcast(byte_update(bitcast(constant_struct))), "ptr")` into appropriate SMT constraints.

> Do the test cases work when run with --no-simplify?

Yes, ESBMC can verify the original program when running with `--no-simplify`.

Best,  
Lucas

dan-eicher:
> Yes, we could trace back a few steps to identify where this malformed expression was created

Is it even a malformed expression? We're doing some non-typical C stuff which is why we're hitting all these different code paths and finding these issues.

PR #3494, opened 2026-02-07
Allow untyped params in nested functions (HumanEval/6 P1)

sergillam:
This PR  the parameter annotation requirement for nested functions by defaulting missing annotations to Any, while keeping top‑level functions strict. It unblocks HumanEval/6 P1 without modifying the original test.


lucasccordeiro (approved):

lucasccordeiro:

PR #3495, opened 2026-02-07
[python/humaneval] Fix enumerate transformation and scope/type inference

brcfarias:
This PR enables the `humaneval_0 ` and ` humaneval_6` tests. To fix that, we needed adjustments in the `enumerate ` pre‑processing, name resolution, and type inference.

### 1) enumerate in nested loops

We were not expanding enumerate correctly in nested loops.

The pre‑processor converts:

```python
for idx, elem in enumerate(iterable, start):
      BODY
```

  into:

```python
  ESBMC_iter = iterable
  ESBMC_index = start     # or 0 if not provided (enumeration index)
  ESBMC_array_index = 0   # always starts at 0 (array access index)
  ESBMC_length = len(ESBMC_iter)

  while ESBMC_array_index < ESBMC_length:
      idx = ESBMC_index
      elem = ESBMC_iter[ESBMC_array_index]
      ESBMC_index += 1
      ESBMC_array_index += 1
      BODY
```

  The issue was that the variable names were reused in nested loops, so the inner loop overwrote the outer loop’s variables. Example:
```python
  for idx, elem in enumerate(numbers):
      for idx2, elem2 in enumerate(numbers):
          ...

```
This generated:
```python
  ESBMC_iter = numbers
  ESBMC_index = 0
  ESBMC_array_index = 0
  ESBMC_length = len(ESBMC_iter)

  while ESBMC_array_index < ESBMC_length:
      idx  = ESBMC_index
      elem = ESBMC_iter[ESBMC_array_index]

      # inner enumerate reusing the same names
      ESBMC_iter = numbers
      ESBMC_index = 0
      ESBMC_array_index = 0
      ESBMC_length = len(ESBMC_iter)
      ...

```
Fix: use unique names per nesting level (as already done for range):
```python
  ESBMC_iter_1 = numbers
  ESBMC_index_1 = 0
  ESBMC_array_index_1 = 0
  ESBMC_length_1 = len(ESBMC_iter_1)

  while ESBMC_array_index_1 < ESBMC_length_1:
      idx  = ESBMC_index_1
      elem = ESBMC_iter_1[ESBMC_array_index_1]

      ESBMC_iter_0 = numbers
      ESBMC_index_0 = 0
      ESBMC_array_index_0 = 0
      ESBMC_length_0 = len(ESBMC_iter_0)
      ...

```

### 2) Name resolution inside blocks

With unique names, the declarations ended up inside while blocks, but `get_var_node()` only searched the “flat” function body.
That ignored the fact that variables declared inside if/else, while, try, and match are visible in the current scope.

Fix: `get_var_node()` now searches recursively in those blocks, without entering new scopes (FunctionDef, ClassDef, Lambda).

### 3) Call‑site inference

The change in (2) exposed a regression in infer-loop-var: inference was resolving argument names in the callee scope, but it should resolve them in the call‑site scope.

Example:
``` python
  def f(s):
      for x in s: pass   # x local to f

  x = "test"            # x global
  f(x)
```
If inference resolves x inside f, it picks the local x (no type) and s becomes Any. The correct behavior is to resolve x at the call site (global), inferring s: str.

Fix: when inferring parameters from calls, name resolution now uses the call‑site context (global or parent function).

### DIFF
--- a/regression/humaneval/humaneval_0/test.desc
+++ b/regression/humaneval/humaneval_0/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --unwind 9 --no-bounds-check --no-pointer-check --no-align-check
 ^VERIFICATION SUCCESSFUL$
--- a/regression/python/enumerate17/main.py
+++ b/regression/python/enumerate17/main.py
@@ -1,0 +1,8 @@
+numbers = [1, 2]
+total = 0
+
+for i, x in enumerate(numbers):
+    for j, y in enumerate(numbers):
+        total += (i + 1) * (j + 1) * y
+
+assert total == 15
--- a/regression/python/enumerate17/test.desc
+++ b/regression/python/enumerate17/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+ --unwind 9
+^VERIFICATION SUCCESSFUL$
--- a/regression/quixbugs/next_permutation/main.py
+++ b/regression/quixbugs/next_permutation/main.py
@@ -23,10 +23,10 @@ def next_permutation(perm):
 
 assert next_permutation([3, 2, 4, 1]) == [3, 4, 1, 2]
 assert next_permutation([3, 5, 6, 2, 1]) == [3, 6, 1, 2, 5]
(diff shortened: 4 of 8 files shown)


lucasccordeiro (approved):

PR #3486, opened 2026-02-05
[clang-c-frontend] revert the changes in #2347.

lucasccordeiro:
This PR reverts the changes in #2347.

PR #3499, opened 2026-02-08
 [clang-c] Fix argv modeling: add shared string storage for valid pointer tracking

### LINKED ISSUE #2312: ESBMC assumption on argv

LukeW1999:
## Summary

  Fix false positive "invalid pointer" errors when `argv[i]` is passed to string
  functions (e.g., `strncpy`, `strcmp`). The root cause was that argv elements were
  purely symbolic with no backing memory objects, causing dereference checks to report invalid pointers.

  Fixes #2312.

  ## Problem

  The old argv model initialized all elements to `NULL` via `array_of`. This meant
  value-set analysis had no memory object backing `argv[i]`, so any dereference —
  even behind a proper `if (argc > 1)` guard — was flagged as an invalid pointer.

  ## Solution

  Create a shared symbolic `char[128]` array (`argv_str_storage`) in the symbol
  table and initialize all argv elements to `&argv_str_storage[0]` via `array_of`,
  then set `argv[argc] = NULL` per the C standard.

  ### Design trade-offs and known limitations

  **Why a single shared storage instead of per-element allocation:**
  Ideally each `argv[i]` should point to its own independent symbolic string buffer.
  However, I have not found an efficient way to do per-element memory allocation in
  this context. Allocating individual buffers for each argv element (similar to how
  linked list nodes are modeled) causes significant slowdown during symbolic
  execution, as the solver must reason about N separate memory objects. Using a
  single shared buffer is a pragmatic compromise — it gives value-set analysis a
  valid memory object to track, which is sufficient to eliminate the false positives,
  while keeping the performance overhead minimal.

  **The `argc <= 256` upper bound:**
  This is a magic number and a known limitation. Without an upper bound, the solver
  treats argc as unbounded, leading to extremely large (and unrealistic)
  counterexamples. A larger bound provides more coverage but directly increases
  solver overhead — symbolic execution must explore paths for each possible argc
  value. 

  I do not think we need to provide a user-configurable for it. 256 was chosen as a
  reasonable default that covers real-world usage without severe performance impact.
  --beside the github2312_9 now from core to timeout (over 500s with no result on my device)

  ### Code changes

  In `src/clang-c-frontend/clang_c_main.cpp`:
  1. Add `assume(argc <= 256)` to bound argc realistically
  2. Restore `dynamic_size` assignment for argv array (removed by #3486)
  3. Create `argv_str_storage` symbol (`char[128]`) in symbol table
  4. Use `array_of(&argv_str_storage[0])` to initialize all argv elements
  5. Assign `argv[argc] = NULL` per C standard

  ## Test changes

  **KNOWNBUG → CORE** (false positives eliminated):
  - `esbmc/github_2312`, `github_2312_1`, `github_2312_5`, `github_2312_6`
  - `esbmc/github_60`, `esbmc/github_1533`
  - `esbmc-unix/01_broadcast_01`, `esbmc-unix2/23_picosat-846_01`
  - `nonz3/29_exStbHwAcc`

  **THOROUGH → CORE**: `esbmc-cpp/cpp/ch20_2`, `esbmc-unix2/27_exStbFb`

  **New**: `esbmc/github_2312_10` — buffer overflow via `atoi(argv[1])` without bounds check

  **CORE → KNOWNBUG**:
  - `esbmc/github_2312_9`: times out , I just use `--unwind 10` and labelled as knownbug

  **THOROUGH → KNOWNBUG**:
  - `csmith/csmith01`: Previously reported `VERIFICATION FAILED`, but the error
    trace was suspicious — it accessed argv-related memory that had no backing
    storage and immediately reported a failure. With correct argv modeling (elements
    now point to valid symbolic memory), the verification result changed. This
    indicates the previous failure was likely a false positive caused by the
    incomplete argv model, not a genuine bug in the program. Marked as KNOWNBUG
    pending further investigation. 

`The forward condition is unable to prove the property
Unable to prove or falsify the program, giving up.
VERIFICATION UNKNOWN` 

This is what I get right now for csmith01.

  ## Test plan

  - [x] `github_2312` family tests pass
  - [x] `github_60`, `github_1533` pass
  - [x] Multi-threaded tests (`01_broadcast_01`, `23_picosat-846_01`) pass
  - [ ] `github_2312_9` timeout needs investigation (performance limitation)
  - [ ] `csmith01` result change needs investigation


LukeW1999:
Update: 
- it fails the windows github 1533 seems since windows do not have __stdio_common_vfprintf. I changed test.desc to THOROUGH, but it is jut for passing the CI the problem of the different behavior still exsit.
- the new argv model is slow resulting in 
This PR

Statistics:          36600 Files https://github.com/esbmc/esbmc/actions/runs/21797552878/job/62887990978
  correct:           19221
    correct true:    11869
    correct false:    7352
  incorrect:            24
    incorrect true:      8
    incorrect false:    16
  unknown:           17355
  Score:             30578 (max: 61270)

Master https://github.com/esbmc/esbmc/actions/runs/21789581512/job/62868281829

  correct:           19353
    correct true:    11958
    correct false:    7395
  incorrect:            27
    incorrect true:     11
    incorrect false:    16
  unknown:           17220
  Score:             30703 (max: 61270)

I am double checking if it can provide better result for 300s. Otherwise the model has to be simpilied.

LukeW1999:
For 300s:
This PR https://github.com/esbmc/esbmc/actions/runs/21806694882/job/62914755833

  correct:           23194
    correct true:    15179
    correct false:    8015
  incorrect:            45
    incorrect true:     13
    incorrect false:    32
  unknown:           13361
  Score:             37445 (max: 61270)



Master https://github.com/esbmc/esbmc/actions/runs/21332513969/job/61399901891

  correct:           23973
    correct true:    16089
    correct false:    7884
  incorrect:            54
    incorrect true:     18
    incorrect false:    36
  unknown:           12573

  Score:             38910 (max: 61270)

Performance gets worse with this change.

PR #3513, opened 2026-02-11
Fix join() on dynamic lists and add split/join regressions

sergillam:
Add regressions covering join with dynamically built lists and split+join behavior.
Fix join to handle dynamic lists correctly in the frontend.
Correct list element size computation for array types.
Implement a basic runtime split model for 1‑char separators and improve join length handling.
Restore zip(str, str) typing in builtins model.

### DIFF
--- a/src/c2goto/library/python/list.c
+++ b/src/c2goto/library/python/list.c
@@ -60,7 +60,10 @@ size_t __ESBMC_list_size(const PyListObject *l)
   return l ? l->size : 0;
 }
 
-static inline void *__ESBMC_copy_value(const void *value, size_t size)
+static inline void *__ESBMC_copy_value(
+  const void *value,
+  size_t size,
+  size_t type_id)
 {
   // None type (NULL pointer with size 0)
   // Don't allocate: return NULL to preserve None semantics
@@ -69,22 +72,10 @@ static inline void *__ESBMC_copy_value(const void *value, size_t size)
 
   void *copied = __ESBMC_alloca(size);
 
-  if (size == 1)
-    *(uint8_t *)copied = *(const uint8_t *)value;
-  else if (size == 2)
-    *(uint16_t *)copied = *(const uint16_t *)value;
-  else if (size == 4)
-    *(uint32_t *)copied = *(const uint32_t *)value;
-  else if (size == 8)
-    *(uint64_t *)copied = *(const uint64_t *)value;
-  else if (size == 16)
-  {
-    // Handle 16-byte structs (such as dictionaries) explicitly
-    *(uint64_t *)copied = *(const uint64_t *)value;
-    *((uint64_t *)copied + 1) = *((const uint64_t *)value + 1);
-  }
-  else
-    memcpy(copied, value, size);
+  // Always copy byte-wise to preserve object semantics and avoid unsafe
+  // type-punning on non-scalar or unaligned values.
+  (void)type_id;
+  memcpy(copied, value, size);
 
   return copied;
 }
@@ -96,7 +87,7 @@ bool __ESBMC_list_push(
   size_t type_size)
 {
   // TODO: __ESBMC_obj_cpy
-  void *copied_value = __ESBMC_copy_value(value, type_size);
+  void *copied_value = __ESBMC_copy_value(value, type_size, type_id);
 
   // Use a pointer to avoid repeated indexing
   PyObject *item = &l->items[l->size];
--- a/src/c2goto/library/python/list.c
+++ b/src/c2goto/library/python/list.c
@@ -60,7 +60,13 @@ static inline void *__ESBMC_copy_value(const void *value, size_t size)
 
   void *copied = __ESBMC_alloca(size);
 
-  if (size == 8)
+  if (size == 1)
+    *(uint8_t *)copied = *(const uint8_t *)value;
+  else if (size == 2)
+    *(uint16_t *)copied = *(const uint16_t *)value;
+  else if (size == 4)
+    *(uint32_t *)copied = *(const uint32_t *)value;
+  else if (size == 8)
     *(uint64_t *)copied = *(const uint64_t *)value;
   else if (size == 16)
   {
--- a/src/c2goto/library/python/list.c
+++ b/src/c2goto/library/python/list.c
@@ -14,12 +14,21 @@ __ESBMC_values_equal(const void *a, const void *b, size_t size)
 {
   if (a == b)
     return true;
+  if (size == 0)
+    return true;
   // Direct comparison for common sizes - no loop needed
   // Python frontend maps: int/float -> 8 bytes, bool -> 1 byte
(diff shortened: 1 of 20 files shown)


src/c2goto/library/python/string.c
```diff
@@ -1,9 +1,22 @@
 #include <ctype.h>
+#include <stdbool.h>
 #include <limits.h>
 #include <stddef.h>
 #include <string.h>
 #include "python_types.h"
 
+#define ESBMC_PY_STR_TYPE_ID ((size_t)0x826e83195d0d60f0ULL)
```
lucasccordeiro, line 8: Why do we need this?

src/c2goto/library/python/string.c
```diff
@@ -1,9 +1,22 @@
...
+
+// List helpers (defined in list.c)
+extern PyListObject *__ESBMC_list_create(void);
```
lucasccordeiro, line 11: Currently, we have python_types.h.

Shall we create a list.h to expose these functions?

src/python-frontend/python_list.cpp
```diff
@@ -18,6 +18,14 @@
 #include <string>
 #include <functional>
 
+static size_t esbmc_type_name_hash(const std::string &name)
```
lucasccordeiro, line 23: Why do we keep this hard-coded?

src/python-frontend/python_list.cpp
```diff
@@ -18,6 +18,14 @@
...
+{
+  // Stable hash for string type to match runtime models.
+  if (name == "str")
+    return 0x826e83195d0d60f0ULL;
```
lucasccordeiro: hard-coded!

lucasccordeiro (changes requested):

src/c2goto/library/python/string.c
sergillam: Moved ESBMC_PY_STR_TYPE_ID to [python_types.h](https://file+.vscode-resource.vscode-cdn.net/home/sergillam/.vscode-insiders/extensions/openai.chatgpt-0.5.73-linux-x64/webview/#) and documented it there.
We need a stable type_id for string elements produced by the runtime (split/join). The frontend uses a stable hash for "str" when pushing list elements; if the runtime uses a different id, list operations like contains, equality, and join can silently fail due to type mismatch.
Keeping a shared constant ensures runtime/frontend interoperability for string lists.

src/c2goto/library/python/string.c
sergillam: I added `src/c2goto/library/python/list.h` to expose the list helpers and kept `python_types.h` only for shared type defs/constants. `string.c` now includes `list.h` instead of redeclaring the externs.

src/python-frontend/python_list.cpp
sergillam: We keep a deterministic hash here to ensure stable type_ids across runs/compilers and to match the runtime models (e.g., list/string). Using std::hash would be implementation‑defined and can vary between builds. If you prefer, I can move this helper to a shared header (e.g., python_types.h) to avoid duplication.

src/python-frontend/python_list.cpp
sergillam: I moved the hard-coded string type_id into python_types.h as ESBMC_PY_STR_TYPE_ID and now python_list.cpp uses that shared constant. This keeps a stable, deterministic type_id aligned with the runtime models without duplicating the literal.

src/c2goto/library/python/string.c
```diff
@@ -774,59 +787,168 @@ __ESBMC_HIDE:;
...
+      len = item->size;
+      if (s[len - 1] == '\0')
+        len = len - 1;
+      else
+        len = __python_strnlen_bounded(s, 1024);
+    }
+    else
+    {
+      len = __python_strnlen_bounded(s, 1024);
+    }
```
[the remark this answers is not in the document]
sergillam: Done

src/c2goto/library/python/string.c
```diff
@@ -774,59 +787,168 @@ __ESBMC_HIDE:;
...
       }
+      i++;
     }
   }
+  else
+  {
+    char *buf = __ESBMC_alloca(len_str + 1);
+    memcpy(buf, str, len_str);
+    buf[len_str] = '\0';
+    __ESBMC_list_push(out, buf, ESBMC_PY_STR_TYPE_ID, len_str + 1);
```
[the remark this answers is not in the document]
sergillam: Done

src/python-frontend/string_handler.cpp
```diff
@@ -3042,71 +3042,259 @@ exprt string_handler::handle_str_join(const nlohmann::json &call_json)
...
+        all_chars.insert(all_chars.end(), elem_chars.begin(), elem_chars.end());
+      }
 
-    // Start with the first element
-    std::vector<exprt> first_chars =
-      string_builder_->extract_string_chars(elem_exprs[0]);
-    all_chars.insert(all_chars.end(), first_chars.begin(), first_chars.end());
+      return string_builder_->build_null_terminated_string(all_chars);
+    }
+  }
```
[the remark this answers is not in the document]
sergillam: Done

src/python-frontend/string_handler.cpp
```diff
@@ -3042,71 +3042,259 @@ exprt string_handler::handle_str_join(const nlohmann::json &call_json)
...
+          string_builder_->extract_string_chars(elem_exprs[i]);
+        all_chars.insert(all_chars.end(), elem_chars.begin(), elem_chars.end());
+      }
 
-    // Start with the first element
-    std::vector<exprt> first_chars =
-      string_builder_->extract_string_chars(elem_exprs[0]);
-    all_chars.insert(all_chars.end(), first_chars.begin(), first_chars.end());
+      return string_builder_->build_null_terminated_string(all_chars);
+    }
```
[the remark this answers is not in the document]
sergillam: Done

src/c2goto/library/python/list.c
```diff
@@ -60,7 +60,13 @@ static inline void *__ESBMC_copy_value(const void *value, size_t size)
 
   void *copied = __ESBMC_alloca(size);
 
-  if (size == 8)
+  if (size == 1)
```
lucasccordeiro: Can you please revert these changes? We've already run experiments that show these additional cases do not occur in our benchmark set.

lucasccordeiro (changes requested):

src/c2goto/library/python/list.c
sergillam: I’d prefer not to drop it as it fixes the join regressions (humaneval_common_join_*). If those tests are not part of the target suite, I can revert, but then we need an alternative to avoid the small-size memcpy/unwind issue. Do you want the revert anyway?

PR #3523, opened 2026-02-13
[tools] Update plugin docs with marketplace installation commands

pmatos:


pmatos:
Given the plugin is now upstream, you don't need to use the --plugin-dir anymore. Claude can clone esbmc and install the plugin by adding the marketplace and then the plugin as mentioned in the docs.

lucasccordeiro (approved):

PR #3527, opened 2026-02-14
Fix pluralization of 'loop invariant' in status message

### LINKED ISSUE #3526: Ungrammatical pluralization in 'Processing N loop invariant' log message
When ESBMC processes loop invariants, the status message always uses the singular form regardless of count:

```
Processing 0 loop invariant
Processing 2 loop invariant
```

The message should read "loop invariants" when the count is not 1.

Location: `src/goto-symex/symex_main.cpp:1467`

jessealama:
Correctly pluralize "loop invariant" when the count is not 1 (closes #3526).


lucasccordeiro (approved):
Thanks for submitting this PR, @jessealama.

PR #3531, opened 2026-02-15
[build] Disable backtrace functionality on non-GLIBC platforms.

Keno:
<execinfo.h> and the backtrace()/backtrace_symbols_fd() functions are glibc extensions not available on musl libc or FreeBSD. Guard them with #ifdef __GLIBC__ so the signal handler still compiles on non-glibc platforms. The handler continues to function without backtrace support — it resets to SIG_DFL, dumps /proc/self/maps if available, and re-raises the signal.

This PR fixes the build failure. To restore functionality, we would need to add a dependency on a third party execinfo implementation or on libunwind (with a manual implementation for the printing).

AI Disclosure: This change was created by Claude as part of packaging ESMBC for Yggdrasil, the Julia project's binary packaging project (https://github.com/JuliaPackaging/Yggdrasil/pull/13150).

### DIFF
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -59,8 +59,10 @@ extern "C"
 
 #ifndef _WIN32
 #  include <sys/wait.h>
-#  include <execinfo.h>
 #  include <fcntl.h>
+#  ifdef __GLIBC__
+#    include <execinfo.h>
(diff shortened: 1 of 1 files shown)


rafaelsamenezes (approved):

lucasccordeiro (approved):
Thanks for submitting this PR, @Keno.

PR #3532, opened 2026-02-16
[C++ verification] fix the crash caused by the default alignment

### LINKED ISSUE #3522: esbmc crash on single c++ line
`test.cpp`:
```c++
struct __attribute__((__aligned__)) __aligned_storage_max_align_t {};
```

reproduce:
```
esbmc test.cpp
ESBMC version 8.0.0 64-bit x86_64 linux
Parsing test.cpp
Converting
[1]    1944879 segmentation fault (core dumped)  ../../build/install/bin/esbmc test.cpp
```

`__aligned__` without an argument is valid - https://gcc.gnu.org/onlinedocs/gcc/Common-Type-Attributes.html
> Whenever you leave out the alignment factor in an aligned attribute specification, the compiler automatically sets the alignment for the type to the largest alignment that is ever used for any data type on the target machine you are compiling for.

XLiZHI:
fix #3522

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/github_3522/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_3522/main.cpp
@@ -1,0 +1,7 @@
+struct __attribute__((__aligned__)) __aligned_storage_max_align_t
+{
+};
+
+int main()
+{
+}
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

lucasccordeiro:

PR #3539, opened 2026-02-17
Revert "[simplifier] simplify A | (A || B) to A || B in bitwise OR expressions"

intrigus-lgtm:
I'm probably simply missing some additional context, but I don't see how this simplification is correct?
https://godbolt.org/z/5bPq1TonT

Also, `strip_typecasts` could have been used in `conditions_equivalent`.


lucasccordeiro:

We're looking at the simplifications performed by LLVM and determining which are sound to integrate into ESBMC.

For this particular simplification, see https://github.com/llvm/llvm-project/blob/6b58449b2c813588f9c251d4f0963762d6746881/llvm/lib/Analysis/InstructionSimplify.cpp#L2447.

I'm happy to revert this PR if you can explain why this specific simplification is unsound.

lucasccordeiro:
Okay, I see the point. We invoked this simplification even for integer variables, which is unsound. Good catch! 

Thanks for reviewing it and for submitting this PR.

lucasccordeiro (approved):

intrigus-lgtm:
> Okay, I see the point. We invoked this simplification even for integer variables, which is unsound. Good catch!
> 
> Thanks for reviewing it and for submitting this PR.

Sorry for just a revert PR, didn't have time to fix this. Should be possible to check whether the outer context is boolean or sth. like that.

lucasccordeiro:
No worries, @intrigus-lgtm. In any case, the decision to revert is the best one.

We can easily get trapped in these simplifications. 

We should invest effort in extending our simplifier, since it pays off to produce shorter SMT formulas for the solvers.

PR #3542, opened 2026-02-17
Add  regression tests for Python math functions

sergillam:
Adds one regression test per math function (success + fail) covering edge and corner cases.
Expands coverage for trigonometric, hyperbolic, exp/log, integer helpers, float helpers, and misc math utilities.
add test imports in math


lucasccordeiro (approved):

PR #3546, opened 2026-02-18
[C++ verification] support pointer to member function

XLiZHI:
This PR implements the feature "pointer to member function":

Unlike member var, the member function exists in the symbol table. Therefore, we simplify the assignment to a function reference.

Code:

`    int (S::*pmf)() = &S::getX;`

Goto:
```
       DECL signed int (*)();

       ASSIGN pmf=(signed int (*)())(&getX);
```


lucasccordeiro (approved):

PR #3549, opened 2026-02-19
Revert "[C++ verification] fix the crash caused by the default alignment"

intrigus-lgtm:
The PR didn't actually fix the problem, but just hid it.
See this failing testcase which [works fine on clang](https://godbolt.org/z/sr83jrGjq):
```cpp
#include <assert.h>
#include <string.h>

struct __attribute__((__aligned__)) __aligned_storage_max_align_t {};
struct __aligned_storage_max_align_t2 {};

long align1 = alignof(__aligned_storage_max_align_t);
long sizeof1 = sizeof(__aligned_storage_max_align_t);
long align2 = alignof(__aligned_storage_max_align_t2);
long sizeof2 = sizeof(__aligned_storage_max_align_t2);

int main() {
    struct __aligned_storage_max_align_t test = {};
    char zeroes[16] = {};
    assert(align1 == 16L);   // LIES: only works because this is resolved by
                             // clang and not esbmc
    assert(sizeof1 == 16L);  // LIES: only works because this is resolved by
                             // clang and not esbmc
    assert(align2 == 1L);
    assert(sizeof2 == 1L);
    assert(!memcmp(&test, &zeroes, 16)); // FAILS because the object isn't actually 16 bytes big!
}
```

### DIFF
--- a/regression/esbmc-cpp/bug_fixes/github_3522/main.cpp
+++ b/regression/esbmc-cpp/bug_fixes/github_3522/main.cpp
@@ -1,7 +1,0 @@
-struct __attribute__((__aligned__)) __aligned_storage_max_align_t
-{
-};
-
-int main()
-{
-}
--- a/regression/esbmc-cpp/bug_fixes/github_3522/test.desc
+++ b/regression/esbmc-cpp/bug_fixes/github_3522/test.desc
@@ -1,4 +1,0 @@
-CORE
-main.cpp
-
-^VERIFICATION SUCCESSFUL$
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -388,14 +388,11 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
         const clang::AlignedAttr &aattr =
           static_cast<const clang::AlignedAttr &>(*attr);
 
-        if (aattr.getAlignmentExpr())
-        {
-          exprt alignment;
-          if (get_expr(*(aattr.getAlignmentExpr()), alignment))
-            return true;
+        exprt alignment;
+        if (get_expr(*(aattr.getAlignmentExpr()), alignment))
+          return true;
 
-          t.set("alignment", alignment);
-        }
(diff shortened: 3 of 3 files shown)


lucasccordeiro:
Thanks, @intrigus-lgtm.

@XLiZHI: Could you please check this PR?

XLiZHI:
What you mean by "hiding problems"?

When dealing with AST and encountering a null node, how should it be handled? I think it is appropriate to set it as the default value.
```c
| |-AlignedAttr 0x86f39d8 <col:23> aligned
| | `-<<<NULL>>>
```

Can you please provide a solution that you consider appropriate instead of submitting a revert PR? It didn't solve any problems.

Additionally, I don't think we should revert this PR. It has resolved the UB issue (the aligned expression can be null).

intrigus-lgtm:
> What you mean by "hiding problems"?
> 
> When dealing with AST and encountering a null node, how should it be handled? I think it is appropriate to set it as the default value.
> 
> ```c
> | |-AlignedAttr 0x86f39d8 <col:23> aligned
> | | `-<<<NULL>>>
> ```
> 
> Can you please provide a solution that you consider appropriate instead of submitting a revert PR? It didn't solve any problems.
> 
> Additionally, I don't think we should revert this PR. It has resolved the UB issue (the aligned expression can be null).

> Whenever you leave out the alignment factor in an aligned attribute specification, the compiler automatically sets the alignment for the type to the largest alignment that is ever used for any data type on the target machine you are compiling for.

We are not doing this.
ASTContext has methods to still get the correct alignment.
But the current fix just hides the problem instead of properly fixing the problem.
IMHO there's nothing more dangerous in a verification tool than silent bugs so I usually strongly prefer crashing when something unexpected happens.

lucasccordeiro:
I agree with @intrigus-lgtm that it is better to crash than produce an incorrect result.

However, can I suggest that we fix this issue properly?

@XLiZHI: as discussed, please submit a PR to handle @intrigus-lgtm's test case.

intrigus-lgtm:
Superseded by https://github.com/esbmc/esbmc/pull/3569

PR #3557, opened 2026-02-19
add: zulip sponsorship acknowledgement

Yiannis128:
We got sponsored by Zulip, if we decide to move from Slack, we should consider adding this as they have requested it.


lucasccordeiro (approved):

PR #3569, opened 2026-02-21
[CPP] actually fix default alignment #3522

intrigus-lgtm:


XLiZHI:
I'm okay with the changes made to the front end.

But these test cases cannot confirm your fix.

https://github.com/esbmc/esbmc/blob/e356d4e29c27e1b89558e5db382c3afa373cbdc4/src/clang-c-frontend/clang_c_convert.cpp#L1798-L1806

For sizeof/alignof, we directly obtained the results through clang.

intrigus-lgtm:
> But these test cases cannot confirm your fix.
> 
> https://github.com/esbmc/esbmc/blob/e356d4e29c27e1b89558e5db382c3afa373cbdc4/src/clang-c-frontend/clang_c_convert.cpp#L1798-L1806
> 
> For sizeof/alignof, we directly obtained the results through clang.

I mean, without my changes the TCs fail and I tried some indirect testing in e.g. https://github.com/esbmc/esbmc/blob/d6082c167a74312b8084f44878c24c8337d81ae9/regression/esbmc-cpp/bug_fixes/github_3522_2/main.cpp
Both the alignof values as well as the values for the alignment of structs are from clang. I could use offsetof to check whether the offsets of e.g. aligned fields in structs make sense. I could also have like an `struct foo[2]` where I use pointer arithmetic to check whether `foo[1]` is at the right distance from `foo[0]` which would mean that alignment has been honored.

What approach would you recommend instead for testing this?

lucasccordeiro (approved):

@XLiZHI: Can you please submit a small PR with the tests you have in mind?

Issue #3567, opened 2026-02-21
[python] ERROR: join() argument must be a list of strings


lucasccordeiro:
$esbmc main.py

```Python
assert ",".join(["a", "b", "c"]) == "a,b,c"
```

```
ERROR: join() argument must be a list of strings
```

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/3575.

PR #3575, opened 2026-02-23
Support list literals in str.join() (#3567)

sergillam:
str.join()` previously only accepted variable references (`Name` AST nodes) as the iterable argument. Calling it with a direct list literal — e.g. `",".join(["a", "b", "c"])`


lucasccordeiro (approved):

PR #3576, opened 2026-02-23
Fix/GitHub 3566 raise IndexError on out-of-bounds string indexing

sergillam:
Replace the safe-null-fallback `if_exprt` in `handle_index_access()` with a bounds check that emits `cpp-throw IndexError` when the index >= logical string length (array size minus null terminator).


lucasccordeiro (approved):
Thanks for sending this PR, @sergillam.

PR #3577, opened 2026-02-23
[python] add Decimal construction support

pmatos:
Add an operational model for decimal.Decimal with constant-folding construction.

The preprocessor rewrites Decimal(), Decimal(int), Decimal("str"), and Decimal(float) calls by evaluating them at preprocess time with CPython and emitting the model's 4-field internal constructor (sign, int_val, exp, is_special). Aliased imports (from decimal import Decimal as D, import decimal as dec) are handled. Non-constant and multi-argument calls raise NotImplementedError.

Includes regression tests and Hypothesis property tests validating the model against CPython's decimal.Decimal.

This is the first PR of 5, split to ease review.


pmatos:
Interesting - there's some codacy warnings - I will take a look.

pmatos:
> Interesting - there's some codacy warnings - I will take a look.

OK - I think they were false positives and not applicable to the code model.

lucasccordeiro:
Thanks for sending this PR, @pmatos.

Our operational models for the Python libs are located at https://github.com/esbmc/esbmc/tree/master/src/python-frontend/models.

Do you think you could follow this approach?

lucasccordeiro (approved):

pmatos:
@lucasccordeiro added the tests for now to the script ignore list.

The check_python_tests.sh script validates that regression test main.py files produce the same results when run under real CPython as they would under ESBMC. 

However, the decimal and decimal_fail tests access internal model fields (_sign, _int, _exp, _is_special) that only exist in ESBMC's simplified Decimal struct. Real Python's C-accelerated decimal.Decimal doesn't expose these attributes. 

Since ESBMC's Decimal model currently only supports construction (no public methods like ==, str(), or as_tuple()), there's no way to rewrite the tests using public API that works in both environments. Model-vs-CPython conformance is already validated by the Hypothesis property tests, which run under real Python and compare preprocessor output against Decimal.as_tuple(). Once ESBMC models public Decimal operations, these tests can be rewritten to use standard API and removed from the ignore list.

In the followup PRs I will improve this situation and add some documentation to how the Hypothesis testing works, if that's ok with you.

lucasccordeiro:
That's fine, @pmatos.

For now, you can ignore these test cases by updating the https://github.com/esbmc/esbmc/blob/master/scripts/check_python_tests.sh file.

pmatos:
> That's fine, @pmatos.
> 
> For now, you can ignore these test cases by updating the https://github.com/esbmc/esbmc/blob/master/scripts/check_python_tests.sh file.

That's what I already did.

scripts/check_python_tests.sh
```diff
@@ -24,6 +24,8 @@ ignored_dirs=(
   "cover5"
   "convert-byte-update2"
   "constants"
+  "decimal"
+  "decimal_fail"
```
pmatos, line 28: @lucasccordeiro ^

Issue #3578, opened 2026-02-23
[python] incorrect verification result for string slicing


lucasccordeiro:
$ esbmc main.py --incremental-bmc

```python
def test_slicing():
    s = "abcdef"
    assert s[10:] == ""

test_slicing()
```

```
[Counterexample]


State 7  thread 0
----------------------------------------------------
Violated property:
  array bounds violated: array `main.py:8$array_slice$84' upper bound
  (signed long int)main.py:8$i$85 < (signed long int)tmp$7


VERIFICATION FAILED
```

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/3582.

Issue #3295, opened 2025-12-10
[Python] Crash with string slice operation inside a function


zhassan-aws:
For this program:
```python
def f(s: str) -> None:
    ss: str = s[1:]
    assert ss == "oo"

s: str = "foo"
f(s)
```
ESBMC crashes:
```bash
$ esbmc ex.py 
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing ex.py
Converting
Generating GOTO Program
ERROR: 
* type: 
migrate expr failed
Aborted (core dumped)
```
ESBMC version: 12ca0eb1c

zhassan-aws:
It works if `s[1:]` is replaced with `s[1:len(s)]`.

lucasccordeiro:
Thanks for reporting this issue, @zhassan-aws.

Issue #3082, opened 2025-11-07
[Python] Verification unexpectedly fails with string slice


zhassan-aws:
For the following program:
```python
def foo(s: str) -> None:
    ss = s[1:3]
    #assert ss == "oo"

foo("foobar")
```
ESBMC reports verification failure:
```bash
$ esbmc ex.py --unwind 7
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing ex.py
Converting
Generating GOTO Program
GOTO program creation time: 0.405s
GOTO program processing time: 0.003s
Starting Bounded Model Checking
Unwinding loop 9 iteration 1   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 2   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 3   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 4   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 5   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 6   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Not unwinding loop 9 iteration 7   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 38 iteration 1   
Unwinding loop 9 iteration 1   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 2   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 3   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
Unwinding loop 9 iteration 4   file /home/ubuntu/git/esbmc/src/c2goto/library/string.c line 278 column 3 function __memcpy_impl
[...]

Issue #3043, opened 2025-11-03
[Python] Exception with string slice inside function


zhassan-aws:
This is similar to #2976 but with the string slicing statement within the body of a function:
```python
def foo(s: str) -> None:
    l = 2
    ss: str = s[1:l]
```
ESBMC output:
```bash
$ esbmc ex.py 
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing ex.py
Converting
terminate called after throwing an instance of 'nlohmann::json_abi_v3_11_3::detail::type_error'
  what():  [json.exception.type_error.302] type must be number, but is number
Aborted (core dumped)
```
ESBMC version: ce676e3a0

zhassan-aws:
This appears to only happen when the slice operation involves a variable. Using constant indices, e.g. `s[1:2]` works fine.

The original program referred to `len`:
```python
ss: str = s[1:len(s)]
```

lucasccordeiro:
@rafaelsamenezes: Could you please check this issue?

lucasccordeiro:
Thanks for reporting this issue, @zhassan-aws.

PR #3582, opened 2026-02-23
Fix/GitHub 3578 Fix string slicing with out-of-bounds indices and pointer parameters

sergillam:
Fix two string slicing bugs in the Python frontend:

1. **Out-of-bounds slice indices (#3578):** `s[10:]` on `"abcdef"` caused
   "array bounds violated" instead of returning `""` as Python semantics require.
2. **Crash on pointer string slicing (#3043, #3082, #3295):** Slicing a string
   passed as a function parameter (`def foo(s: str): ss = s[1:3]`) caused
   `migrate expr failed` abort during GOTO program generation.

### DIFF
--- a/regression/python/github_3578/main.py
+++ b/regression/python/github_3578/main.py
@@ -1,0 +1,5 @@
+def test_slicing():
+    s = "abcdef"
+    assert s[10:] == ""
+
+test_slicing()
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

PR #3584, opened 2026-02-23
Fix/GitHub 3579 Fix `join()` failing with `"join() argument must be a list of strings"` when the argument comes from a `split()` call.

sergillam:
The `join()` handler only accepted two argument forms:
1. A direct `List` literal (e.g. `",".join(["a", "b", "c"])`)
2. A `Name` reference whose AST declaration value is a `List` node

When the argument is the result of `split()` (e.g. `parts = s.split(",")` then `",".join(parts)`), the variable's value in the AST is a `Call` node — not a `List` — causing the handler to reject it.

### DIFF
--- a/regression/python/github_3579/main.py
+++ b/regression/python/github_3579/main.py
@@ -1,0 +1,6 @@
+def test_split_and_join():
+    s = "a,b,c"
+    parts = s.split(",")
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

Issue #3533, opened 2026-02-16
[Python Frontend] Assigning different types to the same variable causes internal failure


prototypeC14:
When a variable is declared multiple times with different types in the same function, ESBMC fails with an assertion error, even when annotated as Any.
Failing example:
```python
from typing import Any

def test(flag: bool) -> None:
    if flag:
        x: Any = 1
    else:
        x: Any = '1'
    y: Any = x

test(True)
```
Error:
```bash
pointer-analysis/value_set.cpp:1179: void value_sett::assign(const expr2tc &, const expr2tc &, bool):
Assertion `base_type_eq(rhs->type, lhs_type, ns)' failed.
```
Workaround: use different variable names
```python
from typing import Any

def test(flag: bool) -> None:
    if flag:
        x: Any = 1
        y: Any = x
    else:
        x1: Any = '1'
        y: Any = x1

test(True)

```
This works without error.

lucasccordeiro:
@prototypeC14: This PR seemed to fix this issue: https://github.com/esbmc/esbmc/pull/3555.

I've added your test cases here: https://github.com/esbmc/esbmc/commit/29d5704cd2d72bc6b1c3fb4d883ec99378843c60.

Please feel free to reopen this issue if you still face problems assigning different types to the same variable.

PR #3586, opened 2026-02-24
[Python] Fix internal failure when assigning different types to the same variable

prototypeC14:
See issue #3533


lucasccordeiro (approved):

PR #3591, opened 2026-02-24
Fix/GitHub 3589 handle address_of on constant_struct in SMT encoding

sergillam:
When passing a tuple literal to a function that expects a pointer parameter (e.g. `math.dist((0, 0), (3, 4))`), the SMT encoder crashes with: ERROR: Unrecognized address_of operand: address_of pointer_obj : constant_struct

### DIFF
--- a/regression/python/github_3589/main.py
+++ b/regression/python/github_3589/main.py
@@ -1,0 +1,3 @@
+import math
+
+assert math.isclose(math.dist((0, 0), (3, 4)), 5.0)
--- a/regression/python/github_3589/test.desc
+++ b/regression/python/github_3589/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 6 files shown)


intrigus-lgtm:
Can you provide the generated GOTO code for the crash?

intrigus-lgtm (changes requested):
I really don't think this is the right approach.
We're trying to take the address of something which doesn't have an address.
Instead we should store those tuples in the caller as a local variable and *then* pass the address of that local variable to the function.
Just because there is an existing edge-case or maybe hack for arrays doesn't mean that we should add more hacky code.
GOTO:
```
python_user_main (python_user_main):
        // 4004 
        DECL double return_value$_dist$1;
        // 4005 
        FUNCTION_CALL:  return_value$_dist$1=dist(&{ .element_0 = 0 }, &{ .element_0=3, .element_1=4 })
        // 4006 file foo.py line 3 column 0
        FUNCTION_CALL:  py:foo.py@__assert_temp=isclose(return_value$_dist$1, 5.000000, 1.000000e-9, 0.000000)
        // 4007 file foo.py line 3 column 0
        ASSERT (signed int)py:foo.py@__assert_temp == 1
        // 4008 no location
        DEAD python_user_main::$tmp::return_value$_dist$1
        // 4009 no location
        END_FUNCTION
```

lucasccordeiro:
@sergillam: I agree with @intrigus-lgtm here.

> Instead we should store those tuples in the caller as a local variable and then pass the address of that local variable to the function.

This is the way to go; we should not change our SMT backend.

lucasccordeiro (approved):

Issue #3553, opened 2026-02-19
[python] incorrect verification result for humaneval


lucasccordeiro:
$ esbmc main.py --unwind 15 --no-bounds-check --no-pointer-check --no-align-check --compact-trace

```Python
def is_palindrome(string: str) -> bool:
    return string == string[::-1]


def make_palindrome(string: str) -> str:
    if not string:
        return ''

    beginning_of_suffix = 0

    while not is_palindrome(string[beginning_of_suffix:]):
        beginning_of_suffix += 1

    return string + string[:beginning_of_suffix][::-1]

if __name__ == "__main__":
    assert make_palindrome('') == '' ## passed
    assert make_palindrome('x') == 'x' ## passed
    assert make_palindrome('xyz') == 'xyzyx'
````

```
State 21 file main.py line 9 column 4 function make_palindrome thread 0
----------------------------------------------------
  beginning_of_suffix = 0 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 25 file main.py line 9 column 4 function make_palindrome thread 0
----------------------------------------------------
  beginning_of_suffix = 0 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 29 file main.py line 9 column 4 function make_palindrome thread 0
----------------------------------------------------
  beginning_of_suffix = 0 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 34 file main.py line 9 column 4 function make_palindrome thread 0
----------------------------------------------------
  beginning_of_suffix = 0 (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 38 file main.py line 19 column 4 thread 0
----------------------------------------------------
Violated property:
  file main.py line 19 column 4
  assertion return_value$_strcmp$7 == 0
  return_value$_strcmp$7 == 0


VERIFICATION FAILED
```

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/tree/master/regression/python/github_3553

PR #3598, opened 2026-02-24
Fix/GitHub 3553Fix string slicing with step support (e.g. [::-1]) for issue #3553

sergillam:
Adds __python_str_slice C model using __ESBMC_alloca to handle Python string slicing with arbitrary step values, fixes pointer-type slicing in functions by delegating to the new model, and adds inline step support for local array slicing, enabling correct verification of palindrome and string reversal patterns.


lucasccordeiro (approved):

PR #3601, opened 2026-02-25
Fix/GitHub 3592

sergillam:
The native acos handler bypassed the Python model's ValueError raise for out-of-domain arguments. Added a guard that checks if the operand is outside [-1.0, 1.0] and raises ValueError before calling the C acos, matching the existing sqrt domain check pattern.

### DIFF
--- a/regression/python/github_3592/main.py
+++ b/regression/python/github_3592/main.py
@@ -1,0 +1,10 @@
+import math
+
+def test_trigonometry():
+    try:
+        math.acos(2)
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

lucasccordeiro:
@sergillam: I've pushed a small refactoring for your PR: https://github.com/esbmc/esbmc/commit/2ee9b98eea570957c09dfe4cbc5a6048a93a1ebd.

PR #3602, opened 2026-02-25
Fix/GitHub 3593

sergillam:
The native log handler bypassed the Python model's ValueError raise for non-positive arguments. Added a guard that checks if the operand is <= 0 and raises ValueError before calling the C log, matching the existing sqrt domain check pattern.


lucasccordeiro (approved):

PR #3603, opened 2026-02-25
[build] prioritize build-dir includes over source-dir for gen-headers

pmatos:
Stale copies of generated headers (ac_config.h, solver_config.h) in the source tree would shadow the correctly configured ones from the build tree, causing solver backends to silently not register. Put build-directory include paths before source-directory paths so configure_file() outputs always take precedence.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,6 +2,7 @@
 # Modules                       #
 #################################
 
+include_directories(${CMAKE_BINARY_DIR}/src)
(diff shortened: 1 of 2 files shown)


pmatos:
This solves an issue that I was facing when build esbmc without solvers. Then I ran cmake to add z3 and bitwuzla, but running ninja didn't force the linking of the esbmc executable. 
By changing the order of the header inclusion, the right headers will be seen and the relinking will happen automatically.

lucasccordeiro (approved):

PR #3604, opened 2026-02-25
[C verification] add support for volatile checks

XLiZHI:
Type "volatile" indicates that the memory location may be rewritten by the environment at any time, and thus the value read is unpredictable.

In this PR, the `--volatile-check` option will be used to model this type:

```
volatile load → nondet

volatile store → normal assignment
```
An example: 

```
int main()
{
  volatile int flag = 0;

  while (flag != 0)
  {
    __ESBMC_assert(0, "");
  }
}
```
Ideally, the loop is not reachable, but due to factors of the hardware environment, the value may be changed at any time:

`$esbmc main.c --volatile-check`

```
Violated property:
  file main.c line 7 column 5 function main
  assertion 0
  0
```

### DIFF
--- a/regression/esbmc/volatile_01/main.c
+++ b/regression/esbmc/volatile_01/main.c
@@ -1,0 +1,9 @@
+int main()
+{
+  volatile int flag = 0;
+
+  while (flag != 0)
+  {
+    __ESBMC_assert(0, "");
+  }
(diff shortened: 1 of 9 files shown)


lucasccordeiro (approved):

PR #3612, opened 2026-02-26
[python] fix negative indexing on mutated lists (#3610)

### LINKED ISSUE #3610: [python] incorrect verification result for list.insert()
$esbmc main.c --unwind 9

```python
def test_insert():
    l = [1, 3]
    l.insert(100, 4)
    assert l[-1] == 4

test_insert()
```

```
State 438 file main.py line 8 column 4 function test_insert thread 0
Violated property:
  file main.py line 8 column 4 function test_insert
  assertion (signed int)main.py:8eq_tmp111 == 1
  (signed int)main.py:8eq_tmp111 == 1


VERIFICATION FAILED
```

pmatos:
The Python frontend pre-resolved negative list indices at compile time using the list literal's initial size. After mutation (append, insert, extend), the compile-time size was stale, causing l[-1] to access the wrong element. Now negative indices on lists are resolved at runtime via __ESBMC_list_size in build_list_at_call.

Also fix dead code in build_list_at_call where to_integer's return value was checked incorrectly (returns false on success, not true).

Fixes #3610


lucasccordeiro (approved):
Wonderful! Nice contribution ;-)

Many thanks for submitting this PR, @pmatos!

PR #3613, opened 2026-02-26
fix/github_3553_p2

sergillam:
Introduce python_consteval, a lightweight Python AST interpreter that evaluates user-defined functions at conversion time when all arguments are compile-time constants. This eliminates deeply nested loops from the GOTO program, allowing bounded model checking to verify cases like make_palindrome('xyz') without --no-unwinding-assertions.

- Add python_consteval.h/.cpp with support for constants, variables,   arithmetic, string ops, comparisons, slicing, control flow, and   recursive user-defined function calls
- Hook consteval into python_converter::get_function_call() before the normal call path
- Update github_3553 test: remove --no-unwinding-assertions, reduce  unwind bound from 20 to 15"

### DIFF
--- a/regression/humaneval/humaneval_10/test.desc
+++ b/regression/humaneval/humaneval_10/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --unwind 15 --no-bounds-check --no-pointer-check --no-align-check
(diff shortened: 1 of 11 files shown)


lucasccordeiro (approved):

PR #3615, opened 2026-02-26
Fix/GitHub 3609 Fix constant out-of-bounds list indexing to raise IndexError in Python frontend 

sergillam:
This PR fixes constant out-of-bounds list indexing in the Python frontend by emitting IndexError (instead of a frontend conversion error), allowing Python try/except IndexError code to verify correctly, and adds a regression test for github_3609.


lucasccordeiro (approved):

Issue #3622, opened 2026-02-26
[python] Unexpected list out-of-bounds error for symbolic negative index (lst[-1])


lucasccordeiro:
$esbmc main.py

```python
lst = []
n = nondet_int()
__ESBMC_assume(0 <= n <= 3)

i = 0
while i < n:
    lst.append(nondet_int())
    i += 1

v = nondet_int()
assert lst[-1] == v
```

```
Converting
ERROR: List out of bounds at main.py line: 11
```

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/commit/926af40304006f876b65b7c164bf24145b95bee5.

PR #3630, opened 2026-02-27
[python-frontend] Add regression tests for symbolic negative list indexing (lst[-1])

sergillam:
Add regression tests for issue #3622: verify that symbolic negative list indexing (lst[-1]) on dynamically-built lists does not cause a spurious conversion error, and correctly detects out-of-bounds when the list may be empty.


lucasccordeiro (changes requested):
@sergillam: could you please rename the test cases to include "nondet"? This way, you don't need to add the exceptions in `check_python_tests.sh`.

lucasccordeiro (approved):

PR #3644, opened 2026-02-28
fix: prevent crash on LoopHintAttr with unsafe value pointer

Yiannis128:
## Summary

- Guard the `LoopHintAttr` value read in pragma unroll handling with a safe `dyn_cast<IntegerLiteral>` check instead of calling `EvaluateAsInt` directly on a potentially dangling pointer
- Strengthen 9 pragma unroll regression tests to also assert the `Applying #pragma unroll` output line

## Problem

When processing `#pragma unroll` directives, the `LoopHintAttr::getValue()` pointer can be dangling or point to an unexpected expression node on some Clang versions. Calling `EvaluateAsInt` on such a pointer causes a segfault.

## Fix

In the `Numeric` state branch, validate the value pointer using `dyn_cast<clang::IntegerLiteral>` before reading it. The `dyn_cast` checks the `StmtClass` discriminator, which safely rejects corrupt or unexpected expression pointers. If the value is a proper `IntegerLiteral`, its value is read directly. For non-literal constant expressions, the existing `EvaluateAsInt` path is used as a fallback.

### DIFF
--- a/regression/esbmc/pragma_unroll_dowhile_true/test.desc
+++ b/regression/esbmc/pragma_unroll_dowhile_true/test.desc
@@ -1,4 +1,5 @@
 CORE
 main.c
 --no-unwinding-assertions
+^Applying #pragma unroll 5 to loop
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/pragma_unroll_for_true/test.desc
+++ b/regression/esbmc/pragma_unroll_for_true/test.desc
@@ -1,4 +1,5 @@
 CORE
 main.c
 --no-unwinding-assertions
+^Applying #pragma unroll 5 to loop
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/pragma_unroll_macro_true/test.desc
+++ b/regression/esbmc/pragma_unroll_macro_true/test.desc
@@ -1,4 +1,5 @@
 CORE
 main.c
 --no-unwinding-assertions
+^Applying #pragma unroll 6 to loop
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/pragma_unroll_nested_complete_true/test.desc
+++ b/regression/esbmc/pragma_unroll_nested_complete_true/test.desc
@@ -1,4 +1,6 @@
 CORE
 main.c
 --no-unwinding-assertions
+^Applying #pragma unroll 3 to loop
+^Applying #pragma unroll 5 to loop
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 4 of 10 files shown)


lucasccordeiro (approved):

intrigus-lgtm:
> ## Summary
> 
>     * Guard the `LoopHintAttr` value read in pragma unroll handling with a safe `dyn_cast<IntegerLiteral>` check instead of calling `EvaluateAsInt` directly on a potentially dangling pointer
> 
>     * Strengthen 9 pragma unroll regression tests to also assert the `Applying #pragma unroll` output line
> 
> 
> ## Problem
> 
> When processing `#pragma unroll` directives, the `LoopHintAttr::getValue()` pointer can be dangling or point to an unexpected expression node on some Clang versions. Calling `EvaluateAsInt` on such a pointer causes a segfault.
> ## Fix
> 
> In the `Numeric` state branch, validate the value pointer using `dyn_cast<clang::IntegerLiteral>` before reading it. The `dyn_cast` checks the `StmtClass` discriminator, which safely rejects corrupt or unexpected expression pointers. If the value is a proper `IntegerLiteral`, its value is read directly. For non-literal constant expressions, the existing `EvaluateAsInt` path is used as a fallback.

Late to the party, but do you have a testcase/environment that didn't work before?

Yiannis128:
> > ## Summary
> > ```
> > * Guard the `LoopHintAttr` value read in pragma unroll handling with a safe `dyn_cast<IntegerLiteral>` check instead of calling `EvaluateAsInt` directly on a potentially dangling pointer
> > 
> > * Strengthen 9 pragma unroll regression tests to also assert the `Applying #pragma unroll` output line
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ## Problem
> > When processing `#pragma unroll` directives, the `LoopHintAttr::getValue()` pointer can be dangling or point to an unexpected expression node on some Clang versions. Calling `EvaluateAsInt` on such a pointer causes a segfault.
> > ## Fix
> > In the `Numeric` state branch, validate the value pointer using `dyn_cast<clang::IntegerLiteral>` before reading it. The `dyn_cast` checks the `StmtClass` discriminator, which safely rejects corrupt or unexpected expression pointers. If the value is a proper `IntegerLiteral`, its value is read directly. For non-literal constant expressions, the existing `EvaluateAsInt` path is used as a fallback.
> 
> Late to the party, but do you have a testcase/environment that didn't work before?

Nope, this was to solve an issue from industry that didn't even solve it in the end... The reason his program was crashing was not this, just had difficulty FL because we didn't have access to his code. In theory it should make the code more robust. What do you think?

PR #3659, opened 2026-03-01
[C verification] allow unsupported extensions as the return value

### LINKED ISSUE #3640: Function must return value error where there seems to be a return statement.
I ran esbmc on this file: https://github.com/rstudio/libclang-builtin-headers/blob/master/builtin-headers/3.4/mmintrin.h   
with the `--function _mm_adds_pi8` option. 

This should look into this function: https://github.com/rstudio/libclang-builtin-headers/blob/2799a8436b50c4d51f0fa6973bf57b093b7c2e3f/builtin-headers/3.4/mmintrin.h#L140

However, ESBMC gives me this error:
```
ERROR: function must return value
[1]    39648 IOT instruction (core dumped)  ../esbmc/build/src/esbmc/esbmc test.c --function _mm_adds_pi8
```
Is this expected behavior? If so, why?

XLiZHI:
Fixes #3640.

ESBMC will crash when the return value is an unknown function call.

Although we have `--dont-care-about-missing-extensions,` its coverage is not comprehensive. Perhaps we should follow the `add_return` method and treat the return values of these unknown functions as nondet.

### DIFF
--- a/regression/esbmc-unix/unsupported_extensions/main.c
+++ b/regression/esbmc-unix/unsupported_extensions/main.c
@@ -1,0 +1,5 @@
+#include <mmintrin.h>
+
+int main()
+{
+}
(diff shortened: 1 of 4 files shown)


lucasccordeiro (approved):

PR #3676, opened 2026-03-03
[python-frontend] harden list/dict reassignment handling and add github_3623_1 (fail) regression

sergillam:
Refactors duplicated PyList type-detection logic, adds a fail-fast frontend diagnostic for unsupported dict -> list reassignment with safer dict-literal access handling, and introduces regression/python/github_3623 to lock the error path with a dedicated regression test


lucasccordeiro (approved):

PR #3682, opened 2026-03-03
[esbmc] Fix misleading "process crashed" warning in k-induction parellel mode

pmatos:
When using --k-induction-parallel, the parent process intentionally kills child processes via SIGKILL once a solution is found. The waitpid() monitoring code was unconditionally logging "X process crashed" for any terminated child, even those killed on purpose. 

Track which child processes were intentionally killed and only warn when a process is terminated by an unexpected signal.

This really confused me into thinking something went wrong.


lucasccordeiro (changes requested):
Thanks, @pmatos.

Can you please add a regression test?

pmatos:
> Thanks, @pmatos.
> 
> Can you please add a regression test?

This is not fixing a functional change. This is fixing an error message in --k-induction-parallel. What kind of regression test are you looking for here?

lucasccordeiro:
That's fine then.

Can you please check the CI?

```
The following tests FAILED:
	2939 - regression/k-induction-parallel/z_sum_array (Timeout)
Errors while running CTest
```

pmatos:
> That's fine then.
> 
> Can you please check the CI?
> 
> ```
> The following tests FAILED:
> 	2939 - regression/k-induction-parallel/z_sum_array (Timeout)
> Errors while running CTest
> ```

Sorry - right.  If should be fine now.

lucasccordeiro (approved):

PR #3686, opened 2026-03-03
Fixed template matching for side1 symbol

rafaelsamenezes:
Added case for when no typecast is needed.


lucasccordeiro (approved):

PR #3689, opened 2026-03-03
Fix crash on nested Unicode encode/decode method chains (#3635)

### LINKED ISSUE #3635: [python] crash when verifying Python escape sequences and Unicode operations (`splitlines`, UTF-8 encode/decode)
ESBMC crashes with an internal assertion failure while verifying a Python program that uses escape sequences and Unicode string operations.

```python
def test_escape_and_unicode():
    s = "line1\nline2"
    assert s.splitlines() == ["line1", "line2"]

    u = "café"
    assert "é" in u
    assert u.encode("utf-8").decode("utf-8") == "café"

test_escape_and_unicode()
```

```
Converting
Assertion failed: (it != m_data.m_value.object->end()), function operator[], file json.hpp, line 2147.
Abort trap: 6
```

sergillam:
Fixes https://github.com/esbmc/esbmc/issues/3635.

This PR fixes a Python frontend crash caused by unsafe AST access in nested call receivers (e.g.,u.encode("utf-8").decode("utf-8")) by adding guarded receiver resolution and a safe UTF-8 round-trip path, and adds regression test github_3635 to cover splitlines and Unicode encode/decode behavior


lucasccordeiro (approved):

PR #3699, opened 2026-03-04
[Python] Fix/3696 Crash: "unsupported comparison between pointer-backed and non-pointer values” when returning list from function

### LINKED ISSUE #3696: [python] Crash: “Unsupported comparison between pointer-backed and non-pointer values” when returning list from function
ESBMC crashes when verifying a simple Python program that passes a list to a function and returns it unchanged.

The following program:

```python
def return_list(myList: list[int]) -> list[int]:
    tempList = myList
    return tempList

list1 = [1, 2, 3]
list2 = return_list(list1)

assert list2[0] == 1
```

results in:

```
Parsing main.py
Converting
ERROR: Unsupported comparison between pointer-backed and non-pointer values at main.py:8
```

This program is valid Python.

Semantically:
- myList refers to the same list object as `list1`
- `tempList` is an alias
- `list2` refers to the same list object
- `list2[0] == 1` should evaluate to True

The program should verify successfully.

prototypeC14:
Fixes #3696.


lucasccordeiro (approved):

PR #3702, opened 2026-03-04
Function Contract Op Frame: Assigns Compliance & Loop Frame Rule

LukeW1999:
## Summary

This PR introduces the **Operational Frame Rule** into ESBMC's function contract and loop invariant systems, enabling formal verification that functions and loops only modify the memory locations they explicitly declare. It also fixes several soundness issues in `--replace-call-with-contract` and extends `__ESBMC_forall/exists` support to `requires` clauses.

The core pattern implemented is:

```
Snapshot(pre-state) → Havoc/Call → Assert(unchanged == snapshot)
```

---

### Phase 1 — Loop Frame Rule (`--loop-frame-rule`)

A new `frame_enforcer` module (ASSUME mode) strengthens k-induction by preserving relationships between modified and unmodified variables across the loop havoc step.

```c
__ESBMC_loop_invariant(i >= 0 && i <= n);
__ESBMC_loop_assigns(i, sum);
while (i < n) { sum += i++; }
```

### Phase 2A — Global Variable Assigns Compliance

`--enforce-contract` now asserts that the function body only modifies the globals declared in `__ESBMC_assigns`. Previously the clause was recorded but never enforced — silently unsound.

### Phase 2B — Array Element Assigns Compliance (`__ESBMC_assigns(arr[idx])`)

Uses a nondet-witness approach: creates `j = NONDET ∈ [0, ARRAY_ALLOC_ELEMS)`, snapshots `arr[j]` before the call, asserts `j == idx || arr[j] == snap` after. Requires `--assume-nonnull-valid`.

### Phase 2C — Pointer-Parameter Dereference Assigns Compliance (`__ESBMC_assigns(*p)`)

Snapshots `*p` for each pointer parameter not in the assigns clause before the call, asserts `*p == snap` after. Requires `--assume-nonnull-valid`.

---

### `--replace-call-with-contract` pointer havoc

Previously, without an assigns clause, pointer parameters were never havoc'd before `ASSUME(ensures)`. This made all pointer-based assertions vacuously pass (unsound). **Fix:** emit `*actual_arg = NONDET(T)` for each pointer param when no assigns clause is present.

### `__ESBMC_forall/exists` in `requires` + `--replace-call-with-contract`

`extract_requires_from_body` was not calling `inline_temporary_variables`, so `return_value$___ESBMC_forall$N` symbols were transplanted to the caller context as dangling references. **Fix:** inline quantifier return values in requires extraction (same as ensures already did).

### Struct/pointer return field `ensures`

`((T*)__ESBMC_return_value)->field` in ensures now correctly handles pointer-returning functions. **Fix** in `replace_return_value_in_expr`: added handling for the direct-symbol pattern (pointer return) alongside the existing address-of pattern (by-value struct return).

---

## New Command-Line Options

| Option | Description |
|---|---|
| `--loop-frame-rule` | Enable snapshot/frame-assume around the loop havoc step |
| `--assume-nonnull-valid` | Allocate memory for pointer/array params in enforce mode (required for Phase 2B/2C) |

> **Note:** Assigns compliance checking is always-on for `--enforce-contract`. No separate flag needed.

---

### New Module: `frame_enforcer` (`src/goto-programs/frame_enforcer.{h,cpp}`)

| Method | Description |
|---|---|
| `materialize_snapshots()` | Emits `DECL` + `ASSIGN` to capture pre-state |
| `enforce_frame_rule()` | Emits `ASSUME` (loop) or `ASSERT` (contract) for each unassigned variable |
| `collect_global_variables()` | Scans symbol table for all static-lifetime lvalues |
| `classify_assigns_targets()` | Categorises assigns targets into scalars, array elements, pointer dereferences |

### Modified: `contracts.{h,cpp}`

- `generate_checking_wrapper()`: snapshots globals before call, asserts frame compliance after
- `generate_replacement_at_call()`: havocs pointer params when no assigns clause; handles `forall`/`exists` in requires
- `extract_requires_from_body()`: now calls `inline_temporary_variables` (fixes forall bug)
- `inline_temporary_variables()`: extended to also inline `return_value$___ESBMC_forall$N` and `return_value$___ESBMC_exists$N`

### New Builtin: `__ESBMC_loop_assigns`

Variadic macro (up to 5 arguments) for declaring loop-modified variables, stored on `LOOP_INVARIANT` instructions.

### GOTO Program Data Structure

`instructiont` gains a `loop_assigns_targets` field participating in `swap()`, `reset()`, and instruction output.

---

## Tests

**283/283 pass, zero regressions** (224 function contract + 57 loop invariant + 2 python).

Notable new test groups: `assigns_compliance_*`, `ptr_deref_assigns_*`, `arr_assigns_*`, `assume_nonnull_valid_*`, `ptr_sound_*`, `replace_forall_requires_*`, `struct_ptr_return_field_*`.

---

## Known Limitations

- Array assigns compliance is bounded to `ARRAY_ALLOC_ELEMS = 100` elements
- Multi-level pointer assigns (`ptr->sub->field`) not yet supported
- `__ESBMC_forall/exists` in contracts requires `--z3` (Boolector does not support quantifiers)


LukeW1999:
Update: use the special handling to avoid the short circus from clang for contract. It helps a lot. Now, Function contract ensure and requre should be more reliable.

LukeW1999:
@copilot help me to check if the code have any dead code, logic error or code or logic dulipcation can use Lambda or some way to refactor.

LukeW1999:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/24050493727/job/70151622474 
  correct:           19364
    correct true:    11970
    correct false:    7394
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17210
  Score:             30758 (max: 61270)


Master (30s): https://github.com/esbmc/esbmc/actions/runs/23869442048/job/69607591822
  correct:           19292
    correct true:    11902
    correct false:    7390
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17283
  Score:             30634 (max: 61270)

LukeW1999:
**Update — KNOWNBUG Tests for Known Assigns Limitations**

Added three KNOWNBUG regression tests to explicitly document current capability boundaries in assigns compliance checking, rather than leaving them as silent unsound cases:

- **`arr_assigns_large_index_knownbug`**: Phase 2B's nondet witness only ranges over `[0, ARRAY_ALLOC_ELEMS=100)`. Any `assigns(arr[idx])` with `idx >= 100` produces a false positive. Known limitation: unsound for large indices.

- **`assigns_global_arr_elem_knownbug`**: Phase 2B does not apply to global arrays. The frame enforcer snapshots the whole array as a scalar and always fails on any element write. Known limitation: global array element assigns are unsupported.

- **`assigns_multilevel_ptr_knownbug`**: Phase 2C only handles single-level dereferences (`*ptr`, `ptr->field`). Multi-level patterns (`ptr->sub->field`) are not classified; violations on non-assigned fields go undetected. Known limitation: unsound (false negative).

Also replaced `string::find() == 0` with `starts_with()` in `frame_enforcer.cpp` for the internal-symbol filter.

lucasccordeiro (approved):

src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -510,6 +527,28 @@ void __ESBMC_loop_invariant(_Bool);
...
 
 #define __ESBMC_contract __attribute__((annotate("__ESBMC_contract")))
+
+/* Function contract annotations: always available so annotated files compile
+ * with any ESBMC mode (direct BMC, --enforce-contract, --replace-call-with-contract).
+ * In direct BMC: requires/ensures/assigns are dropped (no-ops) by goto_sideeffects.cpp.
+ * Active contract enforcement only kicks in with --enforce-contract etc. */
+void __ESBMC_requires(_Bool);
+void __ESBMC_ensures(_Bool);
+extern int __ESBMC_return_value;
```
XLiZHI, line 537: Is there any particular reason for declaring `__ESBMC_return_value` as extern?

Now ESBMC complains in every program: `[WARNING] extern variable with id c:@__ESBMC_return_value not found, initializing value to nondet! This code would not compile with an actual compiler.`

lucasccordeiro:
@XLiZHI: In addition, shall we replace this `log_warning` with `log_debug`? We have users complaining about these messages.

src/clang-c-frontend/clang_c_language.cpp
LukeW1999: Sorry, I would fix it.

XLiZHI:
> @XLiZHI: In addition, shall we replace this `log_warning` with `log_debug`? We have users complaining about these messages.

I agree, this message seems a bit lengthy.

LukeW1999:
> > @XLiZHI: In addition, shall we replace this `log_warning` with `log_debug`? We have users complaining about these messages.
> 
> I agree, this message seems a bit lengthy.

https://github.com/esbmc/esbmc/pull/4025/changes/58ed9add3288cd7d734f183d2cc23f877dd88094

I changed it to log_debug, it now need to use `--verbosity 8` to print.

lucasccordeiro:
Great!

Many thanks, @LukeW1999.

PR #3726, opened 2026-03-06
Fix #3711: extract side-effects when loop invariant contains function call in the middle

LukeW1999:
First of all, I would like to express my sincere gratitude to @MarkAA-uw for discovering and contributing to this issue!

## Problem

With `--loop-invariant`, verification could **fail** when a loop invariant contained a **function call in the middle** of the invariant expression, even though the same invariant written with the call at the beginning or end would **pass**. The behaviour depended on how the C front-end emitted short-circuit evaluation in GOTO form, not on the actual semantics of the invariant.

Example (from issue #3711, contributed by MarkAA-uw):

```c
qn = mk(0);
qm = mk(1);
__ESBMC_loop_invariant(
  0 <= qn && qn < 256 &&
  in_range(qm) &&    // function call in the middle → FAILED
  (qn != qm));
for (int count = 0; count < 10; count++) { ... }
```

Semantically equivalent variants passed:

- `in_range(qm)` at the **beginning** or **end** of the invariant → PASS.

## Root cause

- `extract_loop_invariants` only sees the guard expression of the LOOP_INVARIANT instruction. With short-circuit evaluation, the guard uses a **compiler temporary** (e.g. `tmp$2`) that holds the result of `(0 <= qn && qn < 256 && in_range(qm))`; the invariant symbols were therefore `{ tmp$2, qn, qm }`, and did **not** include the function-call return symbol (e.g. `return_value$_in_range$1`).
- The previous `extract_and_remove_side_effects` only looked for FUNCTION_CALL/DECL whose return/variable was in **inv_symbols**. So it never collected the `in_range(qm)` call or its DECL, and they were not re-executed before the base case / inductive step assertions.
- In the inductive step, after HAVOC, the temporary and the function result were no longer tied to `qm`, so the solver could find spurious counterexamples and the loop invariant check failed.

## Solution

- **Backward dependency slicing** from the LOOP_INVARIANT: maintain a set `dep_symbols` (initially the symbols in the invariant). When walking backwards, treat ASSIGNs as defining dependencies: if the LHS is in `dep_symbols` and the RHS is non-trivial (not 0/1/NONDET), add that ASSIGN to the side-effects and extend `dep_symbols` with symbols from the RHS. This way we trace from `tmp$2` back to `return_value$_in_range$1` and then to the FUNCTION_CALL.
- **FUNCTION_CALL**: if the return symbol is in `dep_symbols`, add the call to side-effects and record it in `fc_return_syms`.
- **DECL**: only collect DECLs for symbols that are in `fc_return_syms` or in `dep_symbols \ inv_symbols`, and **do not** collect DECLs for variables that are **modified in the loop** (e.g. loop counters, user variables like `i`, `sum`, `y`), so that existing tests (e.g. count_up_down) still pass. Use `loop.get_modified_loop_vars()` and a name-matching predicate (with suffix matching only for compiler-style names containing `$`) to avoid moving user variable DECLs.

Additional helpers:

- `is_trivial_rhs`: skip collecting ASSIGNs whose RHS is constant 0/1, bool, or NONDET, so we only pull in the “real” definition (e.g. from a function call).
- `is_likely_compiler_temp`: names containing `$` are treated as compiler temporaries; user names are not.
- `symbol_name_matches`: exact match or, when allowed, suffix match for `::symbol_id` so that DECLs with short names still match `dep_symbols`/`fc_return_syms` that may be scope-qualified.

## Testing

- **regression/loop-invariants/github_3711**: invariant with function call in the middle, at the end, and at the beginning; plus a variant with no function call. All expect VERIFICATION SUCCESSFUL.
- **regression/loop-invariants/github_3711_fail1**: wrong invariant (e.g. `qn < 0`) → VERIFICATION FAILED.
- **regression/loop-invariants/github_3711_fail2**: wrong invariant (e.g. `qn == qm`) → VERIFICATION FAILED.
- Existing loop-invariant tests (e.g. count_up_down) remain passing.

Should have closed Discussion #3711.


lucasccordeiro (approved):

PR #3727, opened 2026-03-06
[Phase 3] Add Python complex type infrastructure, parsing/type inference, and mixed-operation promotion P2

sergillam:
This PR of Python complex support in the ESBMC frontend.
 introduces the complex type infrastructure as a dedicated struct-based representation with helper utilities for detection and promotion.
 extends parsing and preprocessing so complex literals are annotated and inferred correctly across assignment and non-assignment contexts.
implements literal conversion and binary-operation interception with numeric promotion, enabling correct mixed arithmetic and equality semantics for complex values.
The complex() constructor handling was expanded to cover Python-compatible argument forms, keyword handling, and exception propagation behavior.
Regression coverage was added and extended for constructor behavior and phase-3 mixed binary promotion scenarios, and the new tests pass in both Python and ESBMC for the supported scope.


lucasccordeiro (approved):

PR #3730, opened 2026-03-07
fix - use pre-computed symbolic length for len(input()) #3713

### LINKED ISSUE #3713: [ERROR: SMT solver failed] - Verificação com múltiplas validações em Python
O código completo está no arquivo  teste.py 
Para reproduzir o código com a saída obtida, basta executar:  esbmc teste.py --unwind 30

```
MAX_PRICE_LEN: int = 6

def main() -> None:
    price: str = input() 
    length_price: int = len(price)
    assert 1 <= length_price <= 5
    # chamadas das validações omitidas, para simplificação

main()
```

Saída Obtida:

```
ESBMC version 8.0.0 64-bit x86_64 windows
Target: 64-bit little-endian x86_64-unknown-windows with esbmclibc
Parsing teste.py
Converting
ERROR: Unhandled symbol format in string extraction.
ERROR: Unhandled symbol format in string extraction.
Generating GOTO Program
GOTO program creation time: 1.456s
GOTO program processing time: 0.005s
Starting Bounded Model CheckingSymex completed in: 0.218s (1652 assignments)
Caching time: 0.004s (removed 0 assertions)
Slicing time: 0.013s (removed 484 assignments)
Generated 282 VCC(s), 96 remaining after simplification (1168 assignments)
Encoding to solver time: 0.028s
Solving with solver Z3 v4.13.3
Runtime decision procedure: 2112.142s
ERROR: SMT solver failed
```

Resultado Esperado : VERIFICATION FAILED

sergillam:
Fixes #3713

When `len()` is called on a string returned by `input()`, the Python frontend was falling through to C `strlen()`, which gets loop-unrolled during symbolic execution. This generated O(N) verification conditions per `len()` call and caused Z3 to timeout on programs with multiple `len()` invocations and large unwind bounds.

The `handle_input()` function already creates a bounded symbolic `$input_len$` variable alongside the `$input_str$` array, but this companion symbol was never reused when `len()` was later called on the string.

This PR tracks the `$input_str$ → $input_len$` mapping in `python_converter` and propagates it through variable assignments, so that `len()` on any alias of an `input()` string returns the symbolic length directly, skipping `strlen()` entirely.

### DIFF
--- a/regression/python/github_3713/main.py
+++ b/regression/python/github_3713/main.py
@@ -1,0 +1,8 @@
+MAX_PRICE_LEN: int = 6
+
+def main() -> None:
+    price: str = input()
+    length_price: int = len(price)
+    assert 1 <= length_price <= 5
+
+main()
--- a/regression/python/github_3713/test.desc
+++ b/regression/python/github_3713/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--unwind 30
+^VERIFICATION FAILED$
--- a/scripts/check_python_tests.sh
+++ b/scripts/check_python_tests.sh
@@ -69,6 +69,8 @@ ignored_dirs=(
   "input3"
   "input5"
   "input6"
+  "github_3712"
+  "github_3713"
   "insertion_fail"
   "insertion3_fail"
   "jpl"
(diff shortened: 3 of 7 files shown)


lucasccordeiro (approved):

PR #3738, opened 2026-03-08
 clear stale input-len mapping on variable reassignment #3713

sergillam:
Follow-up to #3713. When a variable that previously aliased an `input()` string
is reassigned to a literal or a non-input symbol, the old `$input_str$ → $input_len$`
mapping was not cleared. This caused `len()` to return the stale symbolic length
instead of computing the correct value.


lucasccordeiro (approved):

PR #3741, opened 2026-03-08
[loop invariants] fix loop invariant with function calls in && chain

LukeW1999:
## Description

Fixes #3735

Loop invariants of the form `__ESBMC_loop_invariant(A && f(x) && B)` (function call in the middle of `&&`) were mis-handled: the generic `remove_sideeffects` path turned the expression into short-circuit GOTO and dropped guard `A` from the stored invariant, leading to wrong verification results.

**Fix:**

1. **goto_sideeffects.cpp** — Special-case `__ESBMC_loop_invariant`: flatten the top-level `&&`, run `remove_sideeffects` per conjunct, rebuild the `and` chain. The full expression is preserved in the LOOP_INVARIANT instruction.
2. **goto_loop_invariant.cpp** — Extract DECL/FUNCTION_CALL instructions for invariant temporaries into a side-effects block and re-insert it before each ASSERT/ASSUME so calls are re-evaluated after HAVOC.
3. Loops with no modified variables are no longer skipped; the base-case ASSERT is still emitted to catch invalid invariants.

**Tests:** `github_3735` (all_cases.c: fun first/middle/last in `&&`), `github_3735_fail` (expect FAILED), `github_3735_multi_fc` (two FCs in `&&`).


lucasccordeiro (approved):

PR #3745, opened 2026-03-08
python: resolve subscript via __getitem__/__setitem__

brcfarias:
This PR adds proper indexing resolution through [\_\_getitem\_\_ ](https://docs.python.org/3/reference/datamodel.html#object.__getitem__)and [\_\_setitem\_\_ ](https://docs.python.org/3/reference/datamodel.html#object.__setitem__) when those methods are defined on the object type, matching expected Python behavior for user-defined classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ src/CMakeCache.txt
 src/CMakeFiles/
 src/CPackConfig.cmake
 src/CPackSourceConfig.cmake
+/.mypy_cache/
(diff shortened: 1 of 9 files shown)


lucasccordeiro:

Can you please check the conflict? I believe we can accept both changes.

brcfarias:
Yes, it is possible to keep both changes. I'm waiting for the CI results to check whether any tests fail, then I'll push a commit to fix the conflict.

lucasccordeiro (approved):

PR #3768, opened 2026-03-10
[python] fix crash on nested attribute method call returning string (#3767)

### LINKED ISSUE #3767: [python] crash with nested attribute method call returning str (get_value_set_rec assertion failure)
ESBMC crashes when verifying a Python program that makes a method call through nested object attributes and returns a str. The crash occurs during symex due to an assertion failure in `value_set.cpp`.

Minimal Reproducer:

```python
class A:
    def f(self, s: str) -> str:
        return "HELLO" if s == "hello" else s

class B:
    def __init__(self) -> None:
        self.a: A = A()

class C:
    def __init__(self) -> None:
        self.b: B = B()

    def g(self, s: str) -> str:
        return self.b.a.f(s)

c = C()
r: str = c.g("hello")
assert r == "HELLO"
```

Expected Behavior: ESBMC should verify the program successfully since the assertion holds.

Actual Behavior: ESBMC aborts with an internal assertion failure:

```
ESBMC version 8.1.0 64-bit aarch64 macos

Parsing main.py
Converting
Generating GOTO Program
GOTO program creation time: 1.235s
GOTO program processing time: 0.026s

Checking base case, k = 1

Not unwinding loop 88 iteration 1
file src/c2goto/library/string.c line 110 column 12
function strcmp

Assertion failed: (sym.rlevel != symbol2t::renaming_level::level2_global),
function get_value_set_rec, file value_set.cpp, line 594.

Abort trap: 6
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3767.

Two bugs caused ESBMC to crash on `self.b.a.f(s)` patterns:

1. **Incorrect method classification (function_call_expr.cpp)**: `is_nested_instance_attr` only checked one level of attributes, so calls such as `self.b.a.f()` were incorrectly treated as ClassMethod instead of InstanceMethod. As a result, self was not passed, leading to an invalid symbol reaching `value_set.cpp`. Fixed by traversing the full attribute chain to the root Name.

2. **level2_global assertion (value_set.cpp)**: `get_value_set_rec()` assumed `level2_global` symbols never occur. However, they appear when string methods are called on global objects during symbolic execution. Fixed by resolving `level2_global` symbols using their base name, consistent with the earlier removal of the similar level2 assertion.

### DIFF
--- a/regression/python/nested-attr-13/test.desc
+++ b/regression/python/nested-attr-13/test.desc
@@ -1,5 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --incremental-bmc
 ^VERIFICATION SUCCESSFUL$
-
--- a/regression/python/nested-attr-8/test.desc
+++ b/regression/python/nested-attr-8/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --incremental-bmc
 ^VERIFICATION SUCCESSFUL$
--- a/regression/python/nested-attr-9/test.desc
+++ b/regression/python/nested-attr-9/test.desc
@@ -1,5 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --incremental-bmc
 ^VERIFICATION SUCCESSFUL$
-
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -587,17 +587,23 @@ void value_sett::get_value_set_rec(
 
     // Look up this symbol, with the given suffix to distinguish any arrays or
     // members we've picked out of it at a higher level.
-    valuest::const_iterator v_it = values.find(sym.get_symbol_name() + suffix);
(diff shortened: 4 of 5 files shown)


lucasccordeiro:
Master (30s): https://github.com/esbmc/esbmc/actions/runs/22822939562

```
  correct:           19312
    correct true:    11913
    correct false:    7399
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17263
  Score:             30665 (max: 61270)
```

This PR (30s): https://github.com/esbmc/esbmc/actions/runs/22890009201

```
  correct:           19385
    correct true:    11990
    correct false:    7395
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17190
  Score:             30815 (max: 61270)
```

rafaelsamenezes (approved):
I think its fine. Although I am not very confident in the assertion removal. Regardless, if there is a bug... then it is being triggered by something else.

PR #3777, opened 2026-03-11
[loop invariant] Enhanced Loop Invariant Verification via Combined K-Induction

### LINKED ISSUE #3760: [Loop invariant] verification strategy of loop invariant
The current loop invariant method of ESBMC is too abstract. A weak loop invariant can cause assertion failure, even though it is correct, this is because we have messed up the variables were used:

esbmc main.c --loop-invariant
```
int main(void) {
  unsigned int x = __VERIFIER_nondet_uint();
  unsigned int y = x;
  
  __ESBMC_loop_invariant(x >=0 && y>=0);
  while (x < 1024) {
    x++;
    y++;
  }

  __VERIFIER_assert(x == y);
}
```
```
[Counterexample]


State 1 file main3.c line 18 column 3 function main thread 0
  y = 4294967295 (11111111 11111111 11111111 11111111)

State 2 file main3.c line 18 column 3 function main thread 0
  x = 3758096383 (11011111 11111111 11111111 11111111)

State 3 file main3.c line 3 column 22 function reach_error thread 0
Violated property:
  file main3.c line 3 column 22 function reach_error
  assertion 0
  0


VERIFICATION FAILED
```

We should apply abstract methods to verify whether the loop invariant is correct, and then use k-induction to accelerate the verification process, the specific idea is as follows:

esbmc --k-induction
```
int main(void) {
  unsigned int x = __VERIFIER_nondet_uint();
  unsigned int y = x;

//branch 1: verify that the invariant holds true under any circumstances
  if (nondet_bool())
  {
    __ESBMC_assert(x == y, "");
    x = nondet_unint();
    y = nondet_unint();
    __ESBMC_assume(x == y);

    while (x < 1024)
    {
      x++;
      y++;
      __ESBMC_assert(x == y, "");

     // end the loop
      __ESBMC_assume(0);
    }

    // end this branch
    __ESBMC_assume(0);
  }

// branch 2: If the invariant is hold true, we will add assumption to the loop and combine it with k-induction to accelerate
  while (x < 1024) {
    x++;
    y++;
    __ESBMC_assume(x == y);
  }

  __VERIFIER_assert(x == y);
}
```

LukeW1999:
## Summary
This update addresses issue #3760 by improving the --loop-invariant mode. The previous implementation often produced spurious counterexamples when provided with "correct but weak" invariants. The new solution introduces a combined verification mode that integrates invariant checking with k-induction for more robust analysis. All the idea and guidance of implementation is from @XLiZHI .

---

## The Problem
The legacy --loop-invariant mode replaces annotated loops with a three-part havoc abstraction:
1. Base-case assertion
2. Havoc + Assume (abstracting the loop body)
3. Inductive-step assertion

The Issue: If an invariant is correct but weak (e.g., x >= 0 && y >= 0), the HAVOC step assigns arbitrary values to loop-modified variables. This can lead to a state where post-loop assertions (e.g., assert(x == y)) fail falsely, even if the program logic is sound.

Example: In a havoc step, y could be set to UINT_MAX, causing a spurious failure of assert(x == y).

---

## Solution: New Verification Modes

We now provide two distinct modes for loop verification:

### 1. --loop-invariant (New Combined Mode)
This is now the default combined pass. It inserts a non-deterministic Branch 1 to verify inductivity while using Branch 2 for k-induction. This default uses k-induction since loop invariant cannot be helpful without k-induction. 

Logic Transformation:
--------------------
```
IF !nondet_bool() GOTO loop_head       // Skip to Branch 2 (k-induction)

// --- Branch 1: Inductivity Check ---
ASSERT(INV)                            // Base case
HAVOC(loop_vars); 
ASSUME(INV)
ASSUME(loop_entry_cond)

ASSERT(INV)                            // Inductive step
ASSUME(false)                          // Terminate Branch 1 execution path

loop_head:
// --- Branch 2: k-induction ---
ASSUME(INV)                            // Hint for k-induction
GOTO loop_head
```
--------------------

Expected Outcomes:
- Wrong (Not inductive): Branch 1 ASSERT fails; clear "invariant not inductive" error.
- Correct but Weak: Branch 1 passes; k-induction proves the property via forward condition.
- Correct and Strong: Branch 1 passes; k-induction closes at the inductive step.

*Note: --loop-invariant now implicitly enables k-induction; no extra flags are required.*

### 2. --loop-invariant-check (Legacy Mode)
The original three-part havoc abstraction is preserved under this new name for backward compatibility. It is actually loop free goto so the k-induction would not affect anything when option is on.

#### Why keep it:
Despite potentially producing false positives, this mode avoids loop unrolling entirely. This makes it faster when the goal is purely to check invariant inductivity, without the overhead of k-induction's forward exploration.

---

### Core Logic
* goto_loop_invariant.{h,cpp}: 
    - Implemented goto_loop_invariant_combinedt pass.
    - Added goto_loop_invariant_combined() entry point.
    - Created extract_invariants_near shared helper to unify invariant extraction.
* esbmc_parseoptions.cpp: 
    - Updated --loop-invariant to route to the combined mode.
    - Updated --loop-invariant-check to maintain legacy behavior.
* options.cpp: Refined CLI descriptions for clarity.

### Regression Testing
* Updated all existing loop-invariant tests to use the renamed --loop-invariant-check flag.
* New Tests Added:
    - kinduction_combined_strong_inv: Verifies k=2, inductive step closure.
    - kinduction_combined_weak_inv: Verifies k=11, forward condition proof.
    - kinduction_combined_wrong_inv: Verifies base-case violation detection.

---

## Test Results
* Pass Rate: 100%  But should test more cases for the new implementation in next PR.
* Fixes: Closes #3760.


lucasccordeiro (approved):
Nice contribution to the ESBMC project!

XLiZHI:
The suggestions provided by Copilot seem reasonable. 

I haven't thoroughly checked this implementation yet, but will the copy goto method correctly handle the goto label?

LukeW1999:
> The suggestions provided by Copilot seem reasonable.
> 
> I haven't thoroughly checked this implementation yet, but will the copy goto method correctly handle the goto label?

Copilot’s comments were spot on here – it correctly pointed out two issues (GOTO target remapping in the copied body and missing side‑effect extraction for invariants). I added three regression tests and confirmed two problems, and then updated the combined --loop-invariant pass to (1) remap all GOTO targets inside the copied slice and (2) reuse the legacy side‑effect extraction logic. I think we still need to further test it though.

LukeW1999:
@XLiZHI It should be fixed now.

XLiZHI (approved):
LGTM

PR #3787, opened 2026-03-11
Optimize complex lowering and import symbol resolution caches

sergillam:
This PR improves Python frontend performance in three critical areas: imported symbol resolution, repeated AST scans, and the complex-number hot path.
It introduces a dedicated cmath lowering policy module, plus import lookup caching and module import-name caching to reduce conversion overhead.
For complex operations, it applies earlier folding/simplifications in complex_div and complex_log while preserving edge-case semantics (including exception behavior in critical boundary cases).


lucasccordeiro (commented):
This PR introduces significant changes to the converter!

Could we be more conservative with these changes?

Also, could you please provide a table showing the performance gains between the master version and this PR?

lucasccordeiro:
Another thing to consider.

Our converter has reached 10K lines of code. I think it is already doing more than it should.

We should collectively review the current implementation and refactor it into smaller classes to manage size/complexity.

brcfarias:
I think you could split it into three PRs (one for each feature).
Regarding the caching, it would be better to have a dedicated class for it instead of keeping it in `function_call_expr`, which should (ideally) only contain the logic related to building function call expressions.

lucasccordeiro (approved):

brcfarias:
Did you notice any performance improvement with this caching in your tests? It would be great if you could share or commit an example.

src/python-frontend/json_utils.h
```diff
@@ -118,11 +118,22 @@ bool is_class(const std::string &name, const JsonType &ast_json)
 template <typename JsonType>
 bool is_module(const std::string &module_name, const JsonType &ast)
 {
-  std::stringstream file_path;
-  file_path << ast["ast_output_dir"].template get<std::string>() << "/"
-            << module_name << ".json";
-  std::ifstream file(file_path.str());
-  return file.is_open();
+  static std::unordered_map<std::string, bool> is_module_cache;
```
brcfarias, line 121: Maybe we can move this caching logic into cmath_lowering_policy and keep this part focused on JSON.

PR #3790, opened 2026-03-12
[python] embed ast2json library and run mypy as subprocess

Yiannis128:
Vendor ast2json 0.4 into src/python-frontend/libs/ so that it is embedded into the ESBMC binary at build time, removing the need to have it installed on the system.

Switch mypy from in-process API call to subprocess invocation so it no longer needs to be importable from the parser's Python environment.

Fix flail.py symbol name generation to use the full relative path for files in subdirectories, avoiding C symbol collisions when multiple files share the same name (e.g. two __init__.py files).

All in all this should allow ESBMC Python to just work without any deps (aside from the optional mypy which can be installed using any env (i.e. pipx))


brcfarias:
@Yiannis128, the idea is interesting for the user, but I think it’s better to keep Python dependency management through pip. If we need to update ast2json in the future or add new dependencies, we will have to manipulate source code instead of simply using pip.

Yiannis128:
> @Yiannis128, the idea is interesting for the user, but I think it’s better to keep Python dependency management through pip. If we need to update ast2json in the future or add new dependencies, we will have to manipulate source code instead of simply using pip.

This PR would mean that the API is more stable. What happens if an update to AST2JSON breaks some API in ESBMC Python parser? This would mean forcing the user to downgrade. Additionally, there may be bugs that are existing on some other version of AST2JSON that the user is using, which we wouldn't know because we always target the latest version.

Essentially, this couples AST2JSON version to ESBMC. Which makes it return more expected results across different systems.

If you want, I can make it so we specify a target version in the cmake (for AST2JSON) and automatically set that up, but it's too much work for essentially a single file library. AST2JSON is so small I think it's justified.

Also, this PR makes ESBMC able to run the Python frontend without any additional setup steps.

brcfarias:
> This PR would mean that the API is more stable. What happens if an update to AST2JSON breaks some API in ESBMC Python parser? This would mean forcing the user to downgrade. Additionally, there may be bugs that are existing on some other version of AST2JSON that the user is using, which we wouldn't know because we always target the latest version.

Good point. How about fixing it to a specific version and check it in the `check_dependencies()`?
I think it is better than embedding third-party code into our codebase. Also think that managing via pip is more adequate if we need to add new dependencies in the future.

scripts/flail.py
```diff
@@ -120,14 +120,26 @@ def cat_cli_command(self):
...
     def obtain_var_name(self):
         obj = Path(self.filepath)
-        return self.prefix + (obj.name.replace('.hs', '_buf')
-                                      .replace('.h', '_buf')
-                                      .replace('.c', '_buf')
-                                      .replace('.py', '_buf')
-                                      .replace('.goto', '_buf')
-                                      .replace('.txt', '_buf')
-                                      .replace('buildidobj', 'buildidstring')
-                                      .replace('-', '_'))
```
Yiannis128, line 130: Restore this

pmatos:
> @Yiannis128, the idea is interesting for the user, but I think it’s better to keep Python dependency management through pip. If we need to update ast2json in the future or add new dependencies, we will have to manipulate source code instead of simply using pip.

`pip` management is a huge pain - at least on linux. Either you install it as a user and you need to configure where the user local conf is or as sudo systemwide. This is a huge amount of friction for a user that wants to verify python. I think we should vendor ast2json inside esbmc tag a version that we update when we want. This has the further advantage that we control what's shipped. If we ship esbmc and tell the user to install ast2json we don't know which version will be installed, meaning a malicious user infiltrating ast2json could break esbmc functionality - this is unlikely but possible mainly because ast2json is not a popular package.

pmatos:
> > This PR would mean that the API is more stable. What happens if an update to AST2JSON breaks some API in ESBMC Python parser? This would mean forcing the user to downgrade. Additionally, there may be bugs that are existing on some other version of AST2JSON that the user is using, which we wouldn't know because we always target the latest version.
> 
> Good point. How about fixing it to a specific version and check it in the `check_dependencies()`? I think it is better than embedding third-party code into our codebase. Also think that managing via pip is more adequate if we need to add new dependencies in the future.

Vendoring happens in all large projects. See https://github.com/FEX-Emu/FEX for example  - their External/ folder vendors a bunch of dependencies. This is an important project for dynamic binary translation used by the Valve Steam Frame and it cannot rely on moving ground underneath. I think we should do something similar.

brcfarias:
Thanks for the comments, @pmatos.
I think using git submodules could be a good way to manage it.
We could have a `src/python_frontend/deps` folder containing a submodule pointing to [https://github.com/YoloSwagTeam/ast2json](https://github.com/YoloSwagTeam/ast2json), from which we can generate/reference the ast2json module.

LukeW1999:
Hi guys @Yiannis128 @brcfarias @pmatos,

Backing Yiannis on this one — I have some skin in the game here.

I currently maintain a PPA and a Homebrew tap for ESBMC, and both rely on `pip` to install `ast2json`. That's fine for a personal tap/PPA, but it's a hard blocker for getting ESBMC into Homebrew core or the Ubuntu archive: neither permits pulling from PyPI during package build/install, and `ast2json` isn't in either ecosystem's package set. PEP 668 closes the door even further on recent Ubuntu.

If this PR is merged, that blocker goes away and the Python frontend becomes packageable through official channels. For 76 lines, vendoring once upstream is much cleaner than every downstream packager carrying the same patch — and pinning the version actually helps with pmatos's supply-chain concern rather than hurting it.

brcfarias:
Ok, I think we could package it into our binary, but keep it in our repository as a submodule. What do you think?

Yiannis128:
> Ok, I think we could package it into our binary, but keep it in our repository as a submodule. What do you think?

I could do that. It makes it simpler since we can target a specific git tag. The only downside to inlining everything in the repo is if the repo gets taken down, but that's a very extreme and unlikely case. I'll start building this.

rafaelsamenezes:
Hm, I would suggest against both pip and submodules. I'd suggest:

1. No pip dependencies. IMO, the project should be transparent to package managers or building the dependency ourselves in CMake. Besides that there is the casual supply chain attack.
2. Considering the project, I don't see the point of using git submodules. Let CMake handle this. 
3. I would even suggest forking the package into ESBMC org, again to avoid supply chain attacks as I have no clue who maintains these scripts.

brcfarias:
> Hm, I would suggest against both pip and submodules. I'd suggest:
> 
> 1. No pip dependencies. IMO, the project should be transparent to package managers or building the dependency ourselves in CMake. Besides that there is the casual supply chain attack.
> 2. Considering the project, I don't see the point of using git submodules. Let CMake handle this.
> 3. I would even suggest forking the package into ESBMC org, again to avoid supply chain attacks as I have no clue who maintains these scripts.

Nice, I agree that this solution is simpler to maintain. I’m fine with doing it via CMake.

Yiannis128:
> Hm, I would suggest against both pip and submodules. I'd suggest:
> 
> 1. No pip dependencies. IMO, the project should be transparent to package managers or building the dependency ourselves in CMake. Besides that there is the casual supply chain attack.
> 2. Considering the project, I don't see the point of using git submodules. Let CMake handle this.
> 3. I would even suggest forking the package into ESBMC org, again to avoid supply chain attacks as I have no clue who maintains these scripts.

I agree with no pip deps for `ast2json` which is required for Python verification.

The issue is that `mypy` is quite complex to build as it has C code that builds for each system. We would be setting up infra for that. It's easier to just look for the "exe" that's installed by pipx as it's not really essential for Python verification, it's more like DLC. On the other hand ast2json rarely gets updated so we could do the fork suggestion, but doesn't git submodules stay with a specific tag? So we could manually audit each version before we change the tag. Let me know what you think.

rafaelsamenezes:
> The issue is that `mypy` is quite complex to build as it has C code that builds for each system. We would be setting up infra for that.

How much different is it from what we already do for most solvers?

>  It's easier to just look for the "exe" that's installed by pipx as it's not really essential for Python verification, it's more like DLC.

I am not even sure wether mypyc is a dependency of mypy.

> On the other hand ast2json rarely gets updated so we could do the fork suggestion, but doesn't git submodules stay with a specific tag?

CMake also let us use tags.

>  So we could manually audit each version before we change the tag.

Tags are not immutable. We don't need to do the fork... but we should be sure that the project doesn't tend to overwrite those.

LukeW1999:
Thanks @rafaelsamenezes — forking ast2json into the ESBMC org is a nice answer to the supply-chain worry, and I'm on board with CMake handling the integration rather than a submodule.

One packaging constraint I should flag before we lock in the mechanism: Homebrew core and the Debian/Ubuntu archive both build in **offline environments**, so anything that fetches sources at configure/build time (e.g. `FetchContent` pulling from a Git tag) won't work for official packaging.

What does work is having the sources physically present in the release tarball — either vendored into the repo, or via a release workflow that bundles them in. CMake can absolutely still drive the build from those local sources; it just can't be the thing downloading them.

Would a setup along those lines work for you?

Yiannis128:
> Thanks @rafaelsamenezes — forking ast2json into the ESBMC org is a nice answer to the supply-chain worry, and I'm on board with CMake handling the integration rather than a submodule.
> 
> One packaging constraint I should flag before we lock in the mechanism: Homebrew core and the Debian/Ubuntu archive both build in **offline environments**, so anything that fetches sources at configure/build time (e.g. `FetchContent` pulling from a Git tag) won't work for official packaging.
> 
> What does work is having the sources physically present in the release tarball — either vendored into the repo, or via a release workflow that bundles them in. CMake can absolutely still drive the build from those local sources; it just can't be the thing downloading them.
> 
> Would a setup along those lines work for you?

So we'll have to discuss this in person tomorrow or something because I need to understand the limitations that those build envs have more closely before locking in a plan.

LukeW1999:
Hi @Yiannis128,

I won't be in the office tomorrow. @Yiannis128 would you mind picking this up with @rafaelsamenezes and @brcfarias in person? Quick summary of the packaging constraints so you've got it:

**Debian / Ubuntu archive:** builds run in `sbuild`/`pbuilder` chroots with no network. Only the source tarball + declared `Build-Depends` from the archive are visible. `ast2json` isn't in the archive, and pip-from-PyPI during build is forbidden by policy (PEP 668 also blocks it on recent Ubuntu).

**Homebrew core:** formula builds run in a sandbox with network disabled during build. Network is only available earlier in the resource-fetch phase, and only for URLs declared in the formula with checksums. No ad-hoc `pip install`.

**Net effect:** by the time CMake runs, no network. So:
- Vendored sources ✅
- Submodules ✅ *if* the release tarball ships with submodule contents populated (default `git archive` doesn't do this — needs an explicit step)
- CMake `FetchContent` / `ExternalProject` pulling at configure or build time ❌
- `pip install` during build ❌

Yiannis128:
> Hi @Yiannis128,
> 
> I won't be in the office tomorrow. @Yiannis128 would you mind picking this up with @rafaelsamenezes and @brcfarias in person? Quick summary of the packaging constraints so you've got it:
> 
> **Debian / Ubuntu archive:** builds run in `sbuild`/`pbuilder` chroots with no network. Only the source tarball + declared `Build-Depends` from the archive are visible. `ast2json` isn't in the archive, and pip-from-PyPI during build is forbidden by policy (PEP 668 also blocks it on recent Ubuntu).
> 
> **Homebrew core:** formula builds run in a sandbox with network disabled during build. Network is only available earlier in the resource-fetch phase, and only for URLs declared in the formula with checksums. No ad-hoc `pip install`.
> 
> **Net effect:** by the time CMake runs, no network. So:
> 
> * Vendored sources ✅
> * Submodules ✅ _if_ the release tarball ships with submodule contents populated (default `git archive` doesn't do this — needs an explicit step)
> * CMake `FetchContent` / `ExternalProject` pulling at configure or build time ❌
> * `pip install` during build ❌

Hi no worries and thanks for the extra details. Ill follow up tomorrow with the rest of the team.

rafaelsamenezes:
Since the scripts are so small. We can:

1. Keep this PR as it is. Embedding it directly in the repo.
2. Add a flag in CMake to let user specifies custom versions of the ast2json

Since it is in Python it should not have any issues in embedding for any OS.

LukeW1999:
Hi @brcfarias, sorry about that — I didn't mean to remove you as a reviewer. 

It looks like there's an issue with your permissions on the repo: your role is showing as Contributor rather than Collaborator, which seems to be why GitHub dropped the review request after the force-push and why I'm unable to re-add you as a reviewer.

scripts/flail.py
```diff
@@ -304,3 +316,13 @@ def test_variable_name_4_prefix(self):
         obj = Flail("buildidobj", 'prefix_')
         expected = "prefix_buildidstring"
         self.assertEqual(obj.obtain_var_name(), expected)
+
+    def test_variable_name_subdir(self):
+        obj = Flail("libs/__init__.py", 'esbmc_')
+        expected = "esbmc_libs___init___buf"
+        self.assertEqual(obj.obtain_var_name(), expected)
+
+    def test_variable_name_nested_subdir(self):
```
brcfarias, line 325: Not sure whether these test functions are needed.

src/python-frontend/libs/__init__.py
brcfarias, line 1: This could be removed.

brcfarias (approved):

scripts/flail.py
LukeW1999: - Kept. These tests cover the subdirectory path-handling logic that fixes the C symbol collision issue, which is the core of the flail.py change.
- Removing them would leave that fix untested and easy to regress.

src/python-frontend/libs/__init__.py
LukeW1999: Thanks, removed

LukeW1999:
Hi everyone, Thanks a lot for your kind help and work @Yiannis128 @brcfarias @rafaelsamenezes @brcfarias 

For clarity, the full scope of this PR is:

* `ast2json` 0.4 is now vendored into `src/python-frontend/libs/` and embedded into the binary at build time via `flail.py`

  * No runtime `pip` dependency.

* `mypy` is now invoked via subprocess (`shutil.which("mypy")`)

  * Gracefully skipped if unavailable.
  * No import-time dependency.

* `flail.py` symbol generation now uses the full relative path for files inside subdirectories

  * Fixes C symbol collisions between same-named files (e.g. multiple `__init__.py` files).

**Why this matters for downstream packaging:**

I verified with Claude that both Homebrew Core and Debian/Ubuntu archive builds run in fully offline build sandboxes. By build time, tools such as `wget`, `curl`, `pip`, and CMake `FetchContent` are unavailable/blocked.

That means vendoring sources into the release tarball is effectively the only packaging-compatible approach.

With this PR, the Homebrew formula becomes a straightforward:

`depends_on "mypy"`

with no `pip`, virtualenv, or fetch logic required. This removes the primary blocker for an official Homebrew Core submission. I will submit a PR once I've figured out how to make the installation comply with Homebrew's standard layout.

Hi @rafaelsamenezes — would appreciate a final sign-off once CI passes.

One more thing: I’ll likely need your help later with the packaging side. We probably need to create a new repository named homebrew-esbmc under the esbmc organization to host the tap version. Let me know what you think!

LukeW1999 (approved):

PR #3793, opened 2026-03-12
[test] Fix/3781 regression test scripts lack timeout enforcement and fix/3646 --multi-property prevents witness generation

prototypeC14:
This pr:
- Add default 1000s timeout at both Python and CTest layers to prevent hanging tests from blocking CI indefinitely
- Handle TimeoutExpired exception properly so timed-out tests report a failure message instead of an unhandled error
- Add per-test memory limiting with a default 8 GB cap, change via `--memory-limit`
- Enable witness generation with `--multi-property`


lucasccordeiro (changes requested):
@prototypeC14: could you please add some regression tests?

lucasccordeiro:
Regression tests for this "Enable witness generation with --multi-property".

prototypeC14:
> Regression tests for this "Enable witness generation with --multi-property".

Got it, I’ve added a few regression test cases along with their expected generated files.

lucasccordeiro (approved):

Issue #3781, opened 2026-03-11
[testing] regression test scripts lack timeout enforcement, causing silent CI runner failures


lucasccordeiro:
**Problem:**
The regression test scripts located in [regression/](https://github.com/esbmc/esbmc/tree/master/regression) do not enforce any per-test timeout or resource limit. When a test case hangs due to excessive execution time or memory consumption, there is no signal to identify it as the culprit. Instead, the CI runner eventually exhausts its resources and terminates ungracefully, producing no actionable diagnostics.

The hosted GitHub Actions runner loses communication with the server mid-job (after hanging for several hours), reporting only:

"The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error."

This makes it impossible to determine which specific test case caused the hang. See [this example CI run](https://github.com/esbmc/esbmc/actions/runs/22925214413/job/66533765843?pr=3761) for reference (job failed after 5h 9m 30s).

**Expected behaviour:**

Individual test cases that exceed a defined time or memory threshold should be killed and reported as failed with a clear timeout message.
The test suite should continue executing remaining tests rather than stalling indefinitely. A summary of which tests started but did not finish should be surfaced in the CI output.

**Proposed steps:**

Audit the CI logs using grep (or equivalent) to identify tests that were started but never completed; these are the likely hang candidates.
Introduce per-test timeouts in the regression test runner (e.g., via timeout in shell scripts, or equivalent flags in the test harness). Optionally, add per-test memory limits to catch memory exhaustion separately from time-outs.

PR #3797, opened 2026-03-12
Enforce per-test timeouts in regression runners and improve timeout diagnostics #3781

sergillam:
This PR hardens regression execution by enforcing per-test timeout behavior and improving timeout diagnostics across both regression runners.

It introduces a configurable CMake timeout (ESBMC_TEST_TIMEOUT) and passes it to the Python regression runner.
The Python runner now applies an effective timeout per test, preserves test-level --timeout semantics, and performs stronger process-tree cleanup on timeout for both Unix-like systems and Windows.
The timeout cleanup path now includes a defensive wait timeout to avoid rare teardown hangs.

The Perl regression runner (test.pl) now supports timeout watchdog execution via timeout/gtimeout, reports clear fallback warnings when watchdog binaries are unavailable, and writes diagnostic markers into output files.
To avoid duplicate fallback markers in parallel mode, it uses an atomic sentinel file (sysopen with O_CREAT|O_EXCL) scoped per run, with robust cleanup including abrupt-exit handling via END.

Test coverage was extended substantially in testing_tool_test.py, including:

timeout behavior with test-defined timeout precedence,
process-tree termination verification,
Windows timeout branch validation with taskkill,
Windows fallback validation when taskkill does not fully terminate the process,
integration coverage for test.pl fallback behavior, including single-emission of TIMEOUT_UNAVAILABLE=1.


intrigus-lgtm (commented):
I don't think we should only test CORE in CI.
What's the point of having THOROUGH tests if we don't even run them in CI?

lucasccordeiro (approved):
Thanks for submitting this PR, @sergillam/

lucasccordeiro:
@intrigus-lgtm: The thorough tests are also executed.

@rafaelsamenezes: Could you please review this PR since you wrote these testing scripts?

intrigus-lgtm:
> @intrigus-lgtm: The thorough tests are also executed.

Looking at e.g. this action run, I don't think this is the case?
https://github.com/esbmc/esbmc/actions/runs/23075098426/job/67033695399

lucasccordeiro:
> Looking at e.g. this action run, I don't think this is the case?

You're correct, @intrigus-lgtm.

@sergillam: We still need to run the test cases marked with THOROUGH. Shall we increase the timeout here?

rafaelsamenezes (changes requested):
Not sure why this PR is necessary:

1. The timeout flag is already set and available in CMake from a previous PR.
2. We don't use perl scripts anymore.
3. Why those tests are marked as THOROUGH? Is it just because they take too long?
4. Why are setting only CORE tests?

This PR is doing too many things with not enough justification.

sergillam:
> Not sure why this PR is necessary:
> 
>     1. The timeout flag is already set and available in CMake from a previous PR.
> 
>     2. We don't use perl scripts anymore.
> 
>     3. Why those tests are marked as THOROUGH? Is it just because they take too long?
> 
>     4. Why are setting only CORE tests?
> 
> 
> This PR is doing too many things with not enough justification.

The main goal of this PR is to reduce PR CI turnaround time. Right now, timeout-heavy regression tests are significantly slowing feedback loops and blocking day-to-day development.

What this PR is trying to do:
1. Keep PR gating focused on CORE tests (fast, high-signal checks).
2. Move timeout-prone cases out of the PR fast path (THOROUGH), so they can still run in extended/nightly coverage.
3. Keep development moving while we open follow-up optimization tasks for the slow modules/tests.

This is not intended to lower quality permanently; it is a CI staging strategy:
- CORE remains mandatory for PR merge.
- THOROUGH remains covered outside the fast gate.
- We plan to optimize the slow suites and then reconsider reclassification.

rafaelsamenezes:
> > Not sure why this PR is necessary:
> > ```
> > 1. The timeout flag is already set and available in CMake from a previous PR.
> > 
> > 2. We don't use perl scripts anymore.
> > 
> > 3. Why those tests are marked as THOROUGH? Is it just because they take too long?
> > 
> > 4. Why are setting only CORE tests?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This PR is doing too many things with not enough justification.
> 
> The main goal of this PR is to reduce PR CI turnaround time. Right now, timeout-heavy regression tests are significantly slowing feedback loops and blocking day-to-day development.

If you wanna optimize, remove test duplication, minimize the test suite I would be OK. Remove throrough tests is not the way to go. Our priority is to avoid regressions, or at least understand them.

> What this PR is trying to do:
> 
> 1. Keep PR gating focused on CORE tests (fast, high-signal checks).
> 2. Move timeout-prone cases out of the PR fast path (THOROUGH), so they can still run in extended/nightly coverage.
> 3. Keep development moving while we open follow-up optimization tasks for the slow modules/tests.
> 
> This is not intended to lower quality permanently; it is a CI staging strategy:
> 
> * CORE remains mandatory for PR merge.
> * THOROUGH remains covered outside the fast gate.
> * We plan to optimize the slow suites and then reconsider reclassification.

That being said, there is a few things interesting from this PR.: 

- This CORE only change could be put inside [on branch checks](https://github.com/esbmc/esbmc/blob/master/.github/workflows/build-onpush.yml).
- Adding a TIMEOUT into the regression but still reporting as an error. I think this is already supported by [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-timeout)

sergillam:
> > > Not sure why this PR is necessary:
> > > ```
> > > 1. The timeout flag is already set and available in CMake from a previous PR.
> > > 
> > > 2. We don't use perl scripts anymore.
> > > 
> > > 3. Why those tests are marked as THOROUGH? Is it just because they take too long?
> > > 
> > > 4. Why are setting only CORE tests?
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This PR is doing too many things with not enough justification.
> > 
> > 
> > The main goal of this PR is to reduce PR CI turnaround time. Right now, timeout-heavy regression tests are significantly slowing feedback loops and blocking day-to-day development.
> 
> If you wanna optimize, remove test duplication, minimize the test suite I would be OK. Remove throrough tests is not the way to go. Our priority is to avoid regressions, or at least understand them.
> 
> > What this PR is trying to do:
> > 
> > 1. Keep PR gating focused on CORE tests (fast, high-signal checks).
> > 2. Move timeout-prone cases out of the PR fast path (THOROUGH), so they can still run in extended/nightly coverage.
> > 3. Keep development moving while we open follow-up optimization tasks for the slow modules/tests.
> > 
> > This is not intended to lower quality permanently; it is a CI staging strategy:
> > 
> > * CORE remains mandatory for PR merge.
> > * THOROUGH remains covered outside the fast gate.
> > * We plan to optimize the slow suites and then reconsider reclassification.
> 
> That being said, there is a few things interesting from this PR.:
> 
>     * This CORE only change could be put inside [on branch checks](https://github.com/esbmc/esbmc/blob/master/.github/workflows/build-onpush.yml).
> 
>     * Adding a TIMEOUT into the regression but still reporting as an error. I think this is already supported by [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-timeout)

Thanks, Rafael i agree with your concerns and with the priority of preserving regression signal.

My suggestion is to split this work as follows:

1) PR A (this one): CI branch-check scope only
- keep PR fast-gate behavior,
- keep timeout attribution/reporting,
- no broad regression reclassification as a generic workaround.

2) PR B: regression suite cleanup/optimization
- dedup/minimization work,
- targeted test reclassification only when justified with data,
- explicit metrics and rationale.

If you prefer a different split, i can adapt. 
This is just my proposed structure to keep review scope clear.

.github/workflows/pull_request.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     name: static llvm-21 DebugOpt
     with:
       operating-system: ubuntu-24.04
-      build-flags: '-b DebugOpt -e ON'
+      build-flags: '-b DebugOpt -e ON -R ON'
```
rafaelsamenezes: This should be reverted.

rafaelsamenezes (changes requested):

.github/workflows/pull_request.yml
sergillam: Done

PR #3802, opened 2026-03-12
 Move frontend Python tests from tests/ to unit/python-frontend

sergillam:
This PR moves frontend Python tests to the unit test hierarchy for proper test-layer organization.
test_complex_phase2.py and test_decimal_model.py were relocated from tests/python-frontend/ to unit/python-frontend/ without content changes.
This aligns the repository structure with the convention that frontend internal behavior checks live under unit/, while end-to-end verification scenarios stay in regression/python/.


lucasccordeiro (approved):

brcfarias:
@sergillam, currently we keep only API tests in this folder. Why do you need to add these files?

lucasccordeiro:
Thanks, @brcfarias.

I've also considered removing these files.

@sergillam: Can you also justify why you need these?

PR #3831, opened 2026-03-14
[TC] remove jira dependency and test, replace with new test

intrigus-lgtm:
We should not "--break-system-packages" install dependencies for tests.
Instead, we should always try to keep tests self-contained so our tests don't
randomly break if dependencies were to change.
The current test didn't even test what it was supposed to test anymore.
What it currently tested will be added as a new test in the next commit.

Currently the preprocessor fails with:
````
List comprehension assignment requires a simple target name
```


lucasccordeiro (approved):

PR #3832, opened 2026-03-15
python: support imports inside function bodies

brcfarias:
This PR fixes imports declared inside function bodies.
Example:
```python
def f():
  from other import inc
  return inc(1)
```

The frontend only handled module-level imports, which resulted in an undefined-function warning followed by a conversion error:
```bash
WARNING: Undefined function 'inc' - replacing with assert(false)
ERROR: Unsupported comparison with unresolved operand type at /home/bruno/main.py:7
```

This change keeps the imported names during type inference, resolves imports found in FunctionDef.body, and adds regression tests for this case.

Needed by #3813

### DIFF
--- a/regression/humaneval/humaneval_133/test.desc
+++ b/regression/humaneval/humaneval_133/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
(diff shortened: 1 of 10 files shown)


lucasccordeiro (approved):

PR #3847, opened 2026-03-16
[python] support __bool__ in conditionals

brcfarias:
Example:
```python
class Foo:
    def __init__(self, value: bool):
        self.value = value

    def __bool__(self) -> bool:
        return self.value


x = Foo(True)
y = 0

if x:
    y = 1
else:
    y = 2

assert y == 1
```

Conditions like `if x:` now call `x.__bool__()` when available. Regression tests were added for both passing and failing cases.

Needed by #3813


lucasccordeiro (approved):

PR #3857, opened 2026-03-16
python: support __bool__ in ternary expressions

brcfarias:
Example:
```python
  class Foo:
    def __init__(self, value: bool):
        self.value = value

    def __bool__(self) -> bool:
        return self.value


def error():
    assert False
    return 1


x = 0 if Foo(True) else error()
assert x == 0
```

Conditions in expressions like `a if x else b` are now converted into `x.__bool__()` when available. Before this change, cases like this would fail with:
```bash
ERROR: first argument of `if' must be boolean, but got
```

Needed by #3813.


lucasccordeiro (approved):

Issue #3853, opened 2026-03-16
[python] false positive `AttributeError` due to loss of parameter type information


lucasccordeiro:
### Description

ESBMC reports an incorrect `AttributeError` when verifying a simple Python program that calls a method on an object passed as a function argument. ESBMC loses the argument's type information and explores infeasible executions in which the parameter is treated as an object without the required attribute.

### Minimal Reproducer

```python
class A:
    def f(self):
        pass

def g(x):
    x.f()

g(A())
```

### Expected Behaviour

The program should verify successfully.

The only execution path is:

```
A() -> g(A()) -> A.f()
```

Since the only argument passed to `g` is an instance of `A`, the call `x.f()` is always valid and cannot raise an exception.

### Actual Behaviour

ESBMC reports the following counterexample:

```
[Counterexample]

State 1 file main.py line 6 column 4 function g thread 0
----------------------------------------------------
Violated property:
  file main.py line 6 column 4 function g
  AttributeError: '' object has no attribute 'f'
  0

VERIFICATION FAILED
```

This indicates that ESBMC explored a path where `x` is treated as an object without the attribute `f` (e.g., a string), even though the only call site passes `A()`.

### Analysis

The issue appears to stem from the Python frontend losing the dynamic type information of the function parameter `x`. Instead of restricting `x` to the type passed at the call site (`A`), the verifier over-approximates `x` as an arbitrary Python object and explores infeasible states where `x` does not have method `f`.

This leads to a spurious `AttributeError`.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/3859.

PR #3859, opened 2026-03-17
[Python] Fix/3853 False positive AttributeError due to loss of parameter type information

prototypeC14:
See issue #3853 
- When a function parameter is passed a class instance at the call site, ESBMC now infers the parameter's type from the class constructor. 
- This enables correct type resolution for method calls and attribute accesses on such parameters.


lucasccordeiro (approved):

PR #3861, opened 2026-03-17
python: resolve tuple subscripts via __getitem__/__setitem__

brcfarias:
This PR adds support for handling tuple subscripts with `__getitem__()` and `__setitem__()`.

Example:
 ```python
  class MyClass:
    def __init__(self) -> None:
        self.value = 0

    def __setitem__(self, key:Tuple[int,int], value: int) -> None:
        i, j = key
        assert i == 2
        assert j == 3
        assert value == 11
        self.value = value


  obj = MyClass()
  obj[2, 3] = 11
  v = obj[2, 3]
  assert v == 11
```

This allows objects that implement these dunder methods to work with tuple subscripts in the same way they already work with list subscripts. Before this change, cases like this would fail with:
```text
ERROR: Cannot unpack pointer - only tuples and arrays can be unpacked
```

Needed by #3813

### DIFF
--- a/regression/python/dunder-getsetitem-tuple-fail/main.py
+++ b/regression/python/dunder-getsetitem-tuple-fail/main.py
@@ -1,0 +1,16 @@
+from typing import Tuple
+
+class MyClass:
+    def __init__(self) -> None:
+        self.value = 0
+
+    def __setitem__(self, key:Tuple[int,int], value: int) -> None:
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):
@brcfarias: For the next PRs, you could use nondet functions to make the test cases harder to verify.

PR #3868, opened 2026-03-17
[clang-cpp] fix spurious destructor warning for trivial implicit dtors (#3854)

lucasccordeiro:
Fixes #3854.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_3854_fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_3854_fail/main.cpp
@@ -1,0 +1,28 @@
+// Regression test for GitHub issue #3854 (expected failure):
+// Verify ESBMC correctly finds counterexamples (not masked by destructor
+// warning false negatives) for classes with implicit trivial destructors.
+class Num
+{
+private:
+  int val;
+
+public:
+  explicit constexpr Num(int num) : val(num) {}
+
+  Num() : val(nondet_int()) {}
+
+  Num inc() const { return Num(val + 1); }
+
+  bool operator==(const Num &rhs) { return this->val == rhs.val; }
+};
+
+int main()
+{
+  Num n0;
+  Num n1 = n0.inc();
+
+  // n0.val + 1 == n1.val, so n0 == n1 is always false: violation expected
+  __ESBMC_assert(n0 == n1, "n0 equals n1 after inc");
+
+  // No explicit return — tests implicit destructor call path
+}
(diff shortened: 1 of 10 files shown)


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -360,6 +360,21 @@ bool clang_cpp_convertert::get_method(
   if (clang_c_convertert::get_function(md, new_expr))
     return true;
 
+  // For implicit or explicitly-defaulted trivial destructors, Clang does not
+  // synthesize a body (hasBody() returns false), leaving symbol.value as nil.
+  // This causes a "no body for function ~T#" warning when destructor calls are
+  // emitted at scope exit. A trivial destructor is a no-op: generate an empty body.
```
intrigus-lgtm: Why is it not enough to only check for implicit dtors?

intrigus-lgtm: I kind of think that this method should probably go into the get_function_body method of the cpp frontend where we already have some ctor initialization logic even if it means that we have to relax some of the `fd.hasBody()` checks around the frontend.

regression/esbmc-cpp/cpp/github_3854_fail/main.cpp
```diff
@@ -0,0 +1,28 @@
+// Regression test for GitHub issue #3854 (expected failure):
+// Verify ESBMC correctly finds counterexamples (not masked by destructor
+// warning false negatives) for classes with implicit trivial destructors.
```
intrigus-lgtm: What is this supposed to mean?
I don't see a difference between the patch applied and without aside from the warnings.

intrigus-lgtm (commented):
The tests don't actually test anything automatically?
Maybe [this](https://regex101.com/r/JFo7Cp/1) regex would work?

lucasccordeiro:
> Maybe [this](https://regex101.com/r/JFo7Cp/1) regex would work?

Thanks, @intrigus-lgtm. This regex seems to work! Let's see what our CI says.

src/clang-cpp-frontend/clang_cpp_convert.cpp
lucasccordeiro: > Why is it not enough to only check for implicit dtors?

Does default also produce a bodyless decl in Clang?

lucasccordeiro: > I kind of think that this method should probably go into the get_function_body method of the cpp frontend where we already have some ctor initialization logic even if it means that we have to relax some of the fd.hasBody() checks around the frontend.

Implemented as suggested! Let's see what our CI complains here :-)

regression/esbmc-cpp/cpp/github_3854_fail/main.cpp
lucasccordeiro: I agree! This comment is somewhat misleading. Fixed!

PR #3872, opened 2026-03-18
python: handle dict instance attributes

brcfarias:
This PR fixes dict instance attributes used through subscripts.

  Example:
  ```python
  class Counter:
      def __init__(self) -> None:
          self.data: dict[int, int]
          self.data = {}

      def __getitem__(self, key: int) -> int:
          try:
              return self.data[key]
          except KeyError:
              return 0

      def __setitem__(self, key: int, value: int) -> None:
          self.data[key] = value
```
This change recovers the value type from the attribute annotation in `__init__` when handling `self.data[key]`.

Needed by #3813


lucasccordeiro (approved):

PR #3877, opened 2026-03-19
Fix CTest test case generation for C and C++

prototypeC14:
This pr:
- Extended esbmc_verifier.h generation to C and C++. The header is now force-included via -include in both C and C++ CMakeLists.txt, so source files compile correctly without explicit __VERIFIER_* declarations.
- Added deduplication of collected test cases before generation.


lucasccordeiro (approved):

PR #3879, opened 2026-03-19
Added new intrinsic to specify unwind numbers

rafaelsamenezes:
This PR introduces the intrinsic `__ESBMC_unroll(int N)` which has the behavior of being a programatically way to do undindset. It is an alternative to the prefered pragma unroll because:

1. ASTImporter can be a bit glitchy
2. Other frontends can also have the same intrinsic

The idea is that users should be able to

```c
int main()
{
  int sum = 0;

  __ESBMC_unroll(5);
  for(int i = 0; i < 5; i++)
    sum += i;

  return 0;
}
```

And then even with k=1 for incremental it will be solved. Note that we implicitly add one to the number of iterations (i.e. 6) to make it more intuitive for users.


lucasccordeiro (approved):
Nice!

lucasccordeiro:
@rafaelsamenezes: Could you please update our documentation with this new feature?

rafaelsamenezes:
> @rafaelsamenezes: Could you please update our documentation with this new feature?

Done, added as an example under the pragma unroll. I prefered to keep it a bit low profile so that users go for the pragma unroll first.

Yiannis128 (approved):
LGTM (after the reg/unit tests finish)

Issue #3386, opened 2026-01-07
[python] failure on regression/quixbugs/lis


lucasccordeiro:
main.py:

````Python
def lis(arr):
    ends = {}
    longest = 0

    for i, val in enumerate(arr):

        prefix_lengths = [j for j in range(1, longest + 1) if arr[ends[j]] < val]

        length = max(prefix_lengths) if prefix_lengths else 0

        if length == longest or val < arr[ends[length + 1]]:
            ends[length + 1] = i
            longest = max(longest, length + 1)

    return longest

assert lis([]) == 0
assert lis([3]) == 1
assert lis([10, 20, 11, 32, 22, 48, 43]) == 4
assert lis([4, 2, 1]) == 1
assert lis([5, 1, 3, 4, 7]) == 4
assert lis([4, 1]) == 1
assert lis([-1, 0, 2]) == 3
assert lis([0, 2]) == 2
assert lis([4, 1, 5, 3, 7, 6, 2]) == 3
assert lis([10, 22, 9, 33, 21, 50, 41, 60, 80]) == 6
assert lis([7, 10, 9, 2, 3, 8, 1]) == 3
assert lis([9, 11, 2, 13, 7, 15]) == 4
````

ESBMC output (https://github.com/esbmc/esbmc/commit/54ea11f6c357c20b2e19081ddf77abf426d9f984):

````
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing main.py

Type checking warning:
main.py:2: error: Function is missing a type annotation
main.py:3: error: Need type annotation for "ends" (hint: "ends: Dict[<type>, <type>] = ...")
main.py:36: error: Call to untyped function "lis" in typed context
main.py:37: error: Call to untyped function "lis" in typed context
main.py:38: error: Call to untyped function "lis" in typed context
main.py:39: error: Call to untyped function "lis" in typed context
main.py:40: error: Call to untyped function "lis" in typed context
main.py:41: error: Call to untyped function "lis" in typed context
main.py:42: error: Call to untyped function "lis" in typed context
main.py:43: error: Call to untyped function "lis" in typed context
main.py:44: error: Call to untyped function "lis" in typed context
main.py:45: error: Call to untyped function "lis" in typed context
main.py:46: error: Call to untyped function "lis" in typed context
main.py:47: error: Call to untyped function "lis" in typed context
[...]

PR #3883, opened 2026-03-20
python: handle list conditions in conditional expressions

brcfarias:
This PR handles lists used as conditions.

  Example:
  ```python
  xs = []
  value = 1 if xs else 0
  assert value == 0
```

This makes list conditions follow Python semantics by evaluating them through their size.
Before this change ESBMC was failing with:
```text
Violated property:
  file main.py line 3 column 0
  assertion value == 0
  value == 0
```

Needed by #3386


lucasccordeiro (approved):

PR #3888, opened 2026-03-22
python: handle short-circuit in boolean conditions

### LINKED ISSUE #3386: [python] failure on regression/quixbugs/lis

brcfarias:
This PR adds short-circuit handling for boolean conditions.

Example:
```python
ends = {}
longest = 0
length = 0
val = 1
arr = [1]
x = 0

if length == longest or val < arr[ends[length + 1]]:
    x = 1

assert x == 1
```

Before this change, ESBMC was failing with:
```text
ERROR: Exception thrown of type KeyError at file regression/python/boolop-short-circuit/main.py line 7
```
This makes and and or in conditions follow evaluation order, so later operands are only evaluated when needed.

Fixes #3386


src/python-frontend/python_converter.h
```diff
@@ -339,6 +339,12 @@ class python_converter
 
   exprt get_conditional_stm(const nlohmann::json &ast_node);
 
+  bool is_coverage_mode() const;
```
lucasccordeiro, line 342: Why do we need these three new methods here? Does it mean that when we enable the coverage mode, we won't use your patch?

lucasccordeiro (changes requested):

src/python-frontend/python_converter.h
brcfarias: They were needed because the patch rewrites boolean conditions to implement short-circuit evaluation using temporaries and explicit branches. Without these guards coverage and pytest-generation tests failed. I am not sure whether it would be better to adjust the tests or to make these modes handle the rewritten form. I'll take a look at this.

PR #3889, opened 2026-03-22
[c++] add/fix C++ OMs and suppress system headers (-nostdinc++)

### LINKED ISSUE #3887: [c++] Ambiguous std symbols on macOS when including <memory> / <iostream>
ESBMC fails to parse even minimal C++ programs on macOS due to ambiguous references between ESBMC’s internal C++ headers and the system libc++ (std::__1).

This results in multiple errors, such as:
- reference to 'char_traits' is ambiguous
- reference to 'istream' is ambiguous
- reference to 'ostream' is ambiguous

Minimal Reproducer:

```cpp
#include <memory>
#include <iostream>

int main() {
  std::unique_ptr<int> x = std::make_unique<int>(5);
}
```

Expected Behavior: The program should parse and proceed to verification without errors.

Root Cause Analysis

The issue appears to be caused by conflicts between:
- ESBMC’s internal C++ headers (esbmc-cpp-headers)
- macOS system libc++ headers (std::__1 inline namespace)

Both define overlapping symbols such as:

- std::char_traits
- std::istream
- std::ostream
- std::basic_string

This leads to ambiguous name lookup during parsing.

Likely Cause

ESBMC is unintentionally mixing:

- its own standard library implementation
- with the system’s libc++ headers
- instead of fully isolating one or the other.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3887.

**Summary**:                                                                                                                         
  - Suppress system C++ headers via -nostdinc++ when ESBMC OMs are enabled, preventing ambiguity between ESBMC stubs and host libc++/libstdc++.                                                                                                                         
  - Add missing limits OM (std::numeric_limits<T> specialisations for all arithmetic types).
  - Add missing std::uncaught_exception(), set_terminate(), set_unexpected() and related types to the exception OM.                         
  - Fix string OM: null-terminate buffers in basic_string(char,n) constructor and all append() overloads; use strcpy in operator=; pass operator==/!= by const-ref to avoid spurious implicit copy.                                                                               
  - Fix utility OM: replace = default copy/move assignment and add explicit copy constructor for pair (ESBMC does not synthesise = default for template members).                                                                                                                    
  - Fix list OM: add missing #include "type_traits" for enable_if/is_integral.
  - Fix tuple OM: add minimal reference_wrapper<T> needed by make_tuple.                                                                    
  - Fix cstddef OM: guard nullptr_t/decltype behind #if __cplusplus >= 201103L for C++03 compatibility.                                     
  - Fix exception OM: guard noexcept behind _ESBMC_NOEXCEPT macro for C++03 compatibility.                                                  
                                                                                                                                            
**Test plan**                                                                                                                                                                                                                                                  
  - Existing regression suite passes (esbmc-cpp/string, esbmc-cpp/unordered_map, esbmc-cpp/stack, esbmc-cpp11/array, esbmc-cpp/OM_sanity_checks, esbmc-cpp/stream, esbmc-cpp/multimap, esbmc-cpp/try_catch)
  - No new regressions introduced by -nostdinc++


intrigus-lgtm (changes requested):
This is not the correct approach IMHO.

The problem is that we include `<utility>` but don't provide an implementation in our OM for it.

This means that we get this weird mix, where we have some of our OMs that define e.g. char_traits but then we also have libc++'s <utility> header pull in libc++'s defines of char_traits which inevitably crashes.

To fix the problem, we should implement `<utility>`.
Alternatively, a user would have to only use the host's c++ stdlib by fully disabling the OMs.

lucasccordeiro:
Hi @intrigus-lgtm,

I also don't like this PR :-(

However, I'm currently blocked from using ESBMC for C++ on MacOS.

The ideal solution is to handle the system's headers, but we're still far from supporting them in ESBMC.

What do you mean by "To fix the problem, we should implement ."?

intrigus-lgtm:
> Hi @intrigus-lgtm,
> 
> I also don't like this PR :-(
> 
> However, I'm currently blocked from using ESBMC for C++ on MacOS.
> 
> The ideal solution is to handle the system's headers, but we're still far from supporting them in ESBMC.
> 
> What do you mean by "To fix the problem, we should implement ."?

Sorry, GitHub's markdown parser interpreted `<utility>` as HTML and stripped it.
I think we (sh/c)ould something like this here:
1. implement `<utility>`
2. disable including the C++ host headers on EVERY platform (not just macOS)
3. Optionally allow the user to mix OMs and C++ host headers

lucasccordeiro:
Thanks, @intrigus-lgtm.

> implement \<utility\>

Implemented as suggested. I had to add also other OMs that I removed some time ago.

> disable including the C++ host headers on EVERY platform (not just macOS)

Implemented as suggested.

> Optionally allow the user to mix OMs and C++ host headers

This could be a future PR! I've created this issue https://github.com/esbmc/esbmc/issues/3897.

intrigus-lgtm (commented):
There are still some outdated references to e.g. `_LIBCPP_EXCEPTION` in a few headers that should be cleaned up.

src/cpp/library/exception
```diff
@@ -0,0 +1,84 @@
...
+
+// noexcept is C++11; use a compat macro so this header parses in C++03 mode.
+#if __cplusplus >= 201103L
+#define _ESBMC_NOEXCEPT noexcept
+#else
+#define _ESBMC_NOEXCEPT
+#endif
+
+// ESBMC intrinsic — no header needed.
+bool nondet_bool();
```
intrigus-lgtm: Why is this needed?
Shouldn't this always be available?

lucasccordeiro: Fixed.

lucasccordeiro:
> There are still some outdated references to e.g. _LIBCPP_EXCEPTION in a few headers that should be cleaned up.

Fixed.

lucasccordeiro:
Many thanks for all your comments, @intrigus-lgtm!

Issue #3876, opened 2026-03-19
[c++] incorrect virtual dispatch with qualified base call under multiple inheritance


lucasccordeiro:
ESBMC incorrectly handles virtual dispatch when calling a base class method via a qualified name (Base::method()) in the presence of multiple inheritance. This results in a spurious counterexample where a trivially true assertion is reported as failing.

Minimal Reproducer:

```cpp
#include <cassert>

struct B1 { virtual void dummy() = 0; };

struct B8 {
    virtual bool guard() = 0;
    bool eval() { return guard(); }
};

struct Derived : B1, B8 {
    void dummy() override {}

    bool guard() override { return true; }

    bool run() { return B8::eval(); }
};

int main() {
    Derived d;
    assert(d.run());
}
```

The assertion should always hold.

Reason:

```
run() calls B8::eval()

eval() calls guard()

guard() is virtual and must dispatch to Derived::guard()

Derived::guard() returns true
```

So:

```
VERIFICATION SUCCESSFUL
```

However, ESBMC reports a failure:

```
Violated property:
  assertion main
  !((_Bool)((signed long int)(!return_value$_run$2)))

VERIFICATION FAILED
```

This implies `run()` is evaluated as false, which is impossible under correct C++ semantics.

This appears to be caused by incorrect handling of this and/or virtual dispatch when:
- calling a base class method via a qualified call (B8::eval())
- in a multiple inheritance hierarchy

Likely issues:
- Missing or incorrect this pointer adjustment to the B8 subobject
- Loss of dynamic type information
- Incorrect lowering of the virtual call guard() (treated as non-virtual or mis-dispatched)

Key Observation

- If multiple inheritance is removed, the issue disappears:

```
struct Derived : B8 { ... };
```

This strongly suggests the bug is related to base subobject layout/pointer adjustment.

According to C++ semantics:

- A qualified call to a base class method (e.g., `B8::eval()`) does not disable virtual dispatch inside that method.
- So `guard()` must still dispatch to the most-derived override.

intrigus-lgtm:
All of C++ polymorphism is very much broken with more exotic virtual/multiple inheritance cases.

Might be able to try whether it works with this branch: https://github.com/esbmc/esbmc/compare/master...intrigus/cpp-layout-restructuring but it won't be possible to integrate this anytime soon and I'm also not confident that at the end my approach is the right/best one.

PR #3891, opened 2026-03-23
Revert "[clang-cpp] fix virtual dispatch through non-first base in multiple inheritance (#3876)"

intrigus-lgtm:
Reverts esbmc/esbmc#3880

I think this PR has the fundamental problem that it mixes clang's `ASTRecordLayout` despite us not using this class anywhere else.

This is super brittle as it means that we will base this-ptr adjustments on a source that is completely unrelated to anything else that ESBMC computes as we compute the struct layout ourselves and perform padding decisions etc.
So this-ptr would use ASTRecordLayout while the rest is based on our layouting.

We should either consider removing our struct layouting and basing it fully on `ASTRecordLayout` (and then apply a cleaner version of this PR) [but this would also mean that we must implement our vtables exactly in the way clang expects probably and would depend on clang then [which we already do for sizeof and alignof, but I disgress]] or re-implementing what clang does using the primitives of ESBMC itself, but I'd personally rather not knowingly add PRs where there'll definitely be funny bugs in the future.


lucasccordeiro:
Thanks, @intrigus-lgtm.

I agree with your comments above.

As we did with @XLiZHI's PR before, where you noted the issue wasn't resolved correctly, could you please submit a PR that fixes the original issue in this PR instead of reverting it?

We'll be happy to review and accept your PR.

rafaelsamenezes:
> We should either consider removing our struct layouting and basing it fully on ASTRecordLayout

I agree. 

Is this affecting you directly right now? The fix was merged quickly because it was blocking some users.

PR #3895, opened 2026-03-23
Unroll fixes

rafaelsamenezes:
Two small fixes for pragma unroll:

1. Now the messages is only show with a increased verbosity level.
2. Added support for constant folding for the unwind parameter.


lucasccordeiro (approved):

Issue #3237, opened 2025-12-02
[python] Failure on regression/quixbugs/longest_common_subsequence


brcfarias:
main.py
```python
def longest_common_subsequence(a, b):
    if not a or not b:
        return ''

    elif a[0] == b[0]:
        return a[0] + longest_common_subsequence(a[1:], b[1:])

    else:
        return max(
            longest_common_subsequence(a, b[1:]),
            longest_common_subsequence(a[1:], b),
            key=len
        )

assert longest_common_subsequence("headache", "pentadactyl") == "eadac"
assert longest_common_subsequence("daenarys", "targaryen") == "aary"
assert longest_common_subsequence("XMJYAUZ", "MZJAWXU") == "MJAU"
assert longest_common_subsequence("thisisatest", "testing123testing") == "tsitest"
assert longest_common_subsequence("1234", "1224533324") == "1234"
assert longest_common_subsequence("abcbdab", "bdcaba") == "bcba"
assert longest_common_subsequence("TATAGC", "TAGCAG") == "TAAG"
assert longest_common_subsequence("ABCBDAB", "BDCABA") == "BCBA"
assert longest_common_subsequence("ABCD", "XBCYDQ") == "BCD"
assert longest_common_subsequence("acbdegcedbg", "begcfeubk") == "begceb"
```

ESBMC (a04f7c43a8) output:
```bash
ESBMC version 7.11.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing longest_common_subsequence/main.py

Type checking warning:

<unknown>:45: SyntaxWarning: invalid escape sequence '\d'
Converting
Generating GOTO Program
ERROR: 
* type: 
migrate expr failed
[1]    37702 IOT instruction (core dumped)  esbmc longest_common_subsequence/main.py
```

PR #3899, opened 2026-03-24
python: return one-character strings from string indexing

brcfarias:
This PR fixes string indexing in `python_list`.

Example:
```python
def f(a: str) -> str:
    if not a:
        return ''
    return a[0]

assert f("ab") == "a"
```

When indexing a string, the frontend now returns a one-character Python string instead of a scalar character value.
Before this change ESBMC the example above was failing with:
```text
[Counterexample]


State 2 file /tmp/main.py line 6 column 0 thread 0
Violated property:
  file /tmp/main.py line 6 column 0
  assertion tmp$3
  tmp$3


VERIFICATION FAILED
```
And regression/python/string-index-concat-char/main.py (added in this PR) was failing with: 

```text
Unwinding recursion f iteration 1 (1 max)
esbmc: /home/bruno/esbmc/src/pointer-analysis/dereference.cpp:443: virtual expr2tc dereferencet::dereference_expr_nonscalar(expr2tc&, guardt&, dereferencet::modet, const expr2tc&): Assertion `!has_dereference(expr)' failed.
[2]    583413 IOT instruction (core dumped)  esbmc /tmp/main.py --incremental-bmc
```

Needed by #3237

### DIFF
--- a/regression/python/string-index-concat-char/main.py
+++ b/regression/python/string-index-concat-char/main.py
@@ -1,0 +1,6 @@
+def f(a):
+    if not a:
+        return ''
+    return a[0] + f(a[1:])
+
+assert f("ab") == "ab"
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/3917.

PR #3900, opened 2026-03-25
[Witness] improved the method for loop invariants && added more TCs

XLiZHI:
1. The way of loop copying has been improved, drawing on the methods in `unroll_loops.cpp`, and the strict fallback has been removed.
2. The position of the loop invariant assumption was adjusted, moving from before the loop exit to after the loop head. This enables the utilization of k induction in the first round of iteration.

```
  while (x > y)
  {
    __ESBMC_assume(x + y == n); // move to here
    x--;
    y++;
    if (x + y != n)
      reach_error();
  }
```


lucasccordeiro (approved):

lucasccordeiro:

PR #3902, opened 2026-03-25
Optimize math dispatch/power paths and strengthen cache/typing robustness

sergillam:
This PR improves robustness and performance in the Python frontend by combining:
- typed fallback fixes for attribute-error paths,
- cache abstraction/integration improvements,
- class-inference and method-lookup hardening,
- targeted optimizations in `python_math`,
- expanded unit coverage.


brcfarias:
Could you please explain how this improve performance, perhaps with an example? Also, would it be possible to split this into smaller PRs?

sergillam:
> Could you please explain how this improve performance, perhaps with an example? Also, would it be possible to split this into smaller PRs?

Before:
- Repeated resolutions were O(n) each.
- Total: O(m*n).

Now (with cache):
- First resolution: O(n).
- Repeated resolutions: average `O(1)`.
- Total: O(n + m).

This translates to measurable runtime improvement in our runs:

- Branch `chore/python-frontend-math-optimizations`:
  - 1107 tests 
  - Total real time: 916.79s

- Branch `master`:
  - 1107 tests 
  - Total real time: 987.34s

Net gain of this branch vs master:
- 70.55s faster
- ~7.15% total real-time improvement

By group (sec*proc):
- `python/`: 12798.24 vs 13841.09 → -7.53%
- `python-intensive/`: 838.64 vs 902.52 → -7.08%
- `regression` (total): 13656.71 vs 14763.88 → -7.50%

src/python-frontend/function_call_expr.cpp
```diff
@@ -44,6 +45,125 @@ constexpr unsigned int SURROGATE_END = 0xDFFF;
...
+      return math_binary_dispatch_kindt::hypot;
+    break;
+  case 8:
+    if (fn == "copysign")
+      return math_binary_dispatch_kindt::copysign;
+    break;
+  default:
+    break;
+  }
+  return math_binary_dispatch_kindt::none;
```
brcfarias: Could we just check fn content without a switch-case here?

brcfarias (changes requested):

src/python-frontend/function_call_expr.cpp
```diff
@@ -44,6 +45,125 @@ constexpr unsigned int SURROGATE_END = 0xDFFF;
...
+  atan2,
+  pow,
+  fmod,
+  copysign,
+  hypot
+};
+
+math_unary_dispatch_kindt classify_math_unary_dispatch(std::string_view fn)
+{
+  switch (fn.size())
```
brcfarias: Could we just check fn content without a switch-case?

brcfarias (changes requested):

src/python-frontend/function_call_expr.cpp
```diff
@@ -3799,31 +3973,119 @@ function_call_expr::get_dispatch_table()
...
+         case math_unary_dispatch_kindt::asinh:
+           return converter_.get_math_handler().handle_asinh(arg_expr, call_);
+         case math_unary_dispatch_kindt::acosh:
+           return converter_.get_math_handler().handle_acosh(arg_expr, call_);
+         case math_unary_dispatch_kindt::atanh:
+           return converter_.get_math_handler().handle_atanh(arg_expr, call_);
+         case math_unary_dispatch_kindt::fabs:
+           return converter_.get_math_handler().handle_fabs(arg_expr, call_);
+         case math_unary_dispatch_kindt::trunc:
+           return converter_.get_math_handler().handle_trunc(arg_expr, call_);
```
brcfarias: Could you refactor this into a single math_handler::handle() method that returns different values based on the expression? Then you can simply call it here.

src/python-frontend/function_call_expr.cpp
```diff
@@ -3799,31 +3973,119 @@ function_call_expr::get_dispatch_table()
...
+             lhs_expr, rhs_expr, call_);
+         case math_binary_dispatch_kindt::fmod:
+           return converter_.get_math_handler().handle_fmod(
+             lhs_expr, rhs_expr, call_);
+         case math_binary_dispatch_kindt::copysign:
+           return converter_.get_math_handler().handle_copysign(
+             lhs_expr, rhs_expr, call_);
+         case math_binary_dispatch_kindt::hypot:
+           return converter_.get_math_handler().handle_hypot(
+             lhs_expr, rhs_expr, call_);
```
brcfarias: The same idea here.

src/python-frontend/function_call_expr.cpp
```diff
@@ -44,6 +45,125 @@ constexpr unsigned int SURROGATE_END = 0xDFFF;
...
+{
+  none,
+  atan2,
+  pow,
+  fmod,
+  copysign,
+  hypot
+};
+
+math_unary_dispatch_kindt classify_math_unary_dispatch(std::string_view fn)
```
brcfarias: Could this function be moved to math_handler? Then the enums above could be moved there as well.

src/python-frontend/function_call_expr.cpp
```diff
@@ -44,6 +45,125 @@ constexpr unsigned int SURROGATE_END = 0xDFFF;
...
+    if (fn == "trunc")
+      return math_unary_dispatch_kindt::trunc;
+    break;
+  default:
+    break;
+  }
+  return math_unary_dispatch_kindt::none;
+}
+
+math_binary_dispatch_kindt classify_math_binary_dispatch(std::string_view fn)
```
brcfarias: Could this function be moved to math_handler? Then the enums above could be moved there as well.

src/python-frontend/function_call_expr.cpp
sergillam: Done

src/python-frontend/function_call_expr.cpp
sergillam: Done

src/python-frontend/function_call_expr.cpp
sergillam: Done

src/python-frontend/function_call_expr.cpp
sergillam: Done

brcfarias:
Can we keep this PR focused on caching and remove unrelated changes?

src/python-frontend/function_call_expr.cpp
```diff
@@ -227,63 +228,90 @@ int function_call_expr::decode_utf8_codepoint(const std::string &utf8_str) const
 
 void function_call_expr::get_function_type()
 {
+  const auto node_type_of = [](const nlohmann::json &node) -> std::string {
+    if (node.is_object() && node.contains("_type") && node["_type"].is_string())
+      return node["_type"].get<std::string>();
+    return "";
+  };
+
+  if (!call_.contains("func") || !call_["func"].is_object())
```
brcfarias: You can remove this check. At this point, we only need to initialize `function_type_`, as this validation is already handled externally.

src/python-frontend/function_call_expr.cpp
```diff
@@ -227,63 +228,90 @@ int function_call_expr::decode_utf8_codepoint(const std::string &utf8_str) const
 
 void function_call_expr::get_function_type()
 {
+  const auto node_type_of = [](const nlohmann::json &node) -> std::string {
+    if (node.is_object() && node.contains("_type") && node["_type"].is_string())
+      return node["_type"].get<std::string>();
+    return "";
```
brcfarias: `_type` is a mandatory field here, so you could trigger an exception instead of returning an empty string. Something like:
```
const auto node_type_of = [](const nlohmann::json &node) -> std::string {
  if (!node.is_object() || !node.contains("_type") || !node["_type"].is_string())
    throw std::runtime_error("Missing or invalid _type field");
  return node["_type"].get<std::string>();
};
```

src/python-frontend/function_call_expr.cpp
```diff
@@ -2302,10 +2330,23 @@ exprt function_call_expr::build_constant_from_arg() const
 
 std::string function_call_expr::get_object_name() const
 {
+  if (
+    !call_.contains("func") || !call_["func"].is_object() ||
+    !call_["func"].contains("value") || !call_["func"]["value"].is_object())
```
brcfarias: I think this check is unnecessary as well.

src/python-frontend/function_call_expr.cpp
```diff
@@ -2302,10 +2330,23 @@ exprt function_call_expr::build_constant_from_arg() const
...
+  {
+    return "";
+  }
+
   const auto &subelement = call_["func"]["value"];
+  const auto node_type_of = [](const nlohmann::json &node) -> std::string {
+    if (node.is_object() && node.contains("_type") && node["_type"].is_string())
+      return node["_type"].get<std::string>();
+    return "";
```
brcfarias: We have the same lambda here as in `get_function_type`, so the previous comment applies here as well. You could define a function and reuse it in both places.

src/python-frontend/function_call_expr.cpp
```diff
@@ -3799,31 +3854,82 @@ function_call_expr::get_dispatch_table()
...
+         if (!dispatched.is_nil())
+           return dispatched;
+       }
+
+       // Enforce canonical arity/error behavior for handled names even when
+       // args count is wrong, without duplicating per-function dispatch here.
+       if (converter_.get_math_handler().is_unary_dispatch_function(func_name))
+       {
+         exprt arg_expr = require_one_arg();
+         if (std::optional<exprt> guarded = guard_one_real_arg(arg_expr);
```
brcfarias: I'm not sure what "guarded" means in this context. Could we use more intuitive names for methods and variables to make the code easier to read?

src/python-frontend/function_call_expr.cpp
```diff
@@ -5309,10 +5238,10 @@ function_call_expr::find_possible_class_types(const symbolt *obj_symbol) const
...
 
   // Check cache first.
-  auto cached =
-    converter_.get_function_call_cache().get_possible_class_types(cache_key);
-  if (cached.has_value())
-    return cached.value();
+  if (
+    const auto *cached =
+      converter_.get_function_call_cache().get_possible_class_types(cache_key))
+    return *cached;
```
brcfarias, line 5253: Please avoid raw pointers when possible. Could we revert to the previous implementation?

src/python-frontend/json_utils.h
```diff
@@ -178,14 +178,40 @@ template <typename JsonType>
 const std::string
 get_object_alias(const JsonType &ast, const std::string &obj_name)
 {
-  for (auto &node : ast["body"])
```
brcfarias, line 181: I think the changes in this file could be removed.

src/python-frontend/python_converter.cpp
```diff
@@ -6299,11 +6313,22 @@ void python_converter::get_var_assign(
   {
     is_converting_rhs = true;
 
+    // For plain, unannotated reassignments (Assign), do not force RHS typing
+    // through the current LHS type hint. Python semantics are dynamic and
+    // expressions like `x = 2` after `x = 1.0` must retag `x` as integer.
+    const bool plain_assign = ast_node["_type"] == "Assign";
+    exprt *saved_lhs = current_lhs;
+    if (plain_assign)
+      current_lhs = nullptr;
```
brcfarias: I also think this block isn’t necessary. Could you remove it, please?

src/python-frontend/python_converter.cpp
```diff
@@ -6299,11 +6313,22 @@ void python_converter::get_var_assign(
...
+
     if (lhs_symbol)
       rhs = get_rhs_with_dict_resolution(ast_node, lhs_symbol->type);
     else
       rhs = get_expr(ast_node["value"]);
 
+    if (plain_assign)
+      current_lhs = saved_lhs;
```
brcfarias: I also think this block isn’t necessary. Could you remove it, please?

sergillam: Done

src/python-frontend/python_converter.cpp
sergillam: DOne

src/python-frontend/json_utils.h
sergillam: done

src/python-frontend/function_call_expr.cpp
sergillam: done

src/python-frontend/function_call_expr.cpp
sergillam: Done

src/python-frontend/function_call_expr.cpp
sergillam: Done

src/python-frontend/function_call_expr.cpp
```diff
@@ -44,6 +45,16 @@ constexpr unsigned int SURROGATE_END = 0xDFFF;
 constexpr const char *CLASS_MARKER = "@C@";
 constexpr const char *FUNCTION_MARKER = "@F@";
 
+/// Extract the mandatory "_type" field from a Python AST JSON node.
+/// Throws if the node is not an object or lacks a valid "_type" string.
+std::string node_type_of(const nlohmann::json &node)
+{
+  if (
+    !node.is_object() || !node.contains("_type") || !node["_type"].is_string())
+    return "";
```
brcfarias: We're still returning an empty string.

sergillam: Done

lucasccordeiro (approved):

PR #3908, opened 2026-03-26
[ir-ra] add theorem-driven support for round-to-away

ShemaAlosaimi:
## Summary

This PR adds theorem-driven IR-RA support for `ROUND_TO_AWAY`.

It extends the existing single-step mode-aware SMT enclosure by adding a dedicated tight path for round-to-nearest, ties-away-from-zero, while preserving the current behaviour for the already supported modes.

## Main idea

`ROUND_TO_AWAY` is a nearest-rounding mode, so it uses the same symmetric nearest-mode enclosure as `ROUND_TO_EVEN`:

- `ra_lo = r - B_near(r)`
- `ra_hi = r + B_near(r)`

where:
- `B_near(r) = eps_rel * |r| + eps_abs`
- `eps_rel = 2^-53` for binary64
- `eps_rel = 2^-24` for binary32

The difference from `ROUND_TO_EVEN` is in tie-breaking only; the linear enclosure and constants remain the same.

## Main changes

- add `ROUND_TO_AWAY` detection in `apply_ieee754_semantics`
- add a dedicated symmetric tight path for `ROUND_TO_AWAY`
- reuse the existing nearest-mode constants for double and single precision
- add double and single precision regressions for the new path
- repurpose the weak-fallback regression to use a symbolic rounding mode

## Regression coverage

This PR includes:

- `ra-rounding-away-tight-bounds`
  - verifies the double-precision `ROUND_TO_AWAY` tight path
  - checks for:
    - `ra_lo_aw`
    - `ra_hi_aw`
    - `ite`
    - the expected double nearest-mode epsilon numerator
    - `VERIFICATION FAILED`

- `ra-rounding-away-tight-bounds-single`
  - verifies the single-precision `ROUND_TO_AWAY` tight path
  - checks for:
    - `ra_lo_aw`
    - `ra_hi_aw`
    - `ite`
    - the expected single nearest-mode epsilon numerator
    - `VERIFICATION FAILED`

- `ra-rounding-directed-weak-fallback`
  - updated to use a symbolic rounding mode
  - verifies that non-constant rounding modes still use weak fallback

## Scope

This PR implements only:
- `ROUND_TO_AWAY`

With this change, all five concrete IEEE-754 rounding modes now have dedicated single-step theorem-driven SMT enclosures:
- `ROUND_TO_EVEN`
- `ROUND_TO_AWAY`
- `ROUND_TO_PLUS_INF`
- `ROUND_TO_MINUS_INF`
- `ROUND_TO_ZERO`

The remaining weak fallback is now for symbolic or otherwise unrecognized rounding modes.

### DIFF
--- a/regression/ir-ra/ra-rounding-away-tight-bounds-single/main.c
+++ b/regression/ir-ra/ra-rounding-away-tight-bounds-single/main.c
@@ -1,0 +1,50 @@
+/* Regression test: ROUND_TO_AWAY tight enclosure under --ir-ra, single precision.
+ *
+ * PURPOSE
+ * -------
+ * Mirrors ra-rounding-away-tight-bounds/ but exercises the IEEE single-
+ * precision (float) path. Verifies that the get_single_eps_rel() /
+ * get_single_min_subnormal() branch in apply_ieee754_semantics is reached and
+ * emits ra_lo_aw:: / ra_hi_aw:: symbols, not the weak fallback.
+ *
+ * PROOF SHAPE (B_near, single precision)
+ * --------------------------------------
+ * ROUND_TO_AWAY uses the same symmetric nearest-mode enclosure as the double
+ * case:
+ *   ra_lo = r - B(r)
+ *   ra_hi = r + B(r)
+ * where B(r) = eps_rel * |r| + eps_abs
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):

PR #3910, opened 2026-03-27
[goto-check] enable shift UB check by default

### LINKED ISSUE #2789: [goto-check] extend goto-check to verify when shift distance is negative
Example:

````C
int a = nondet_int();
assert((0 >> a) == 0);
````

### PR DESCRIPTION
**Pull request by NripeshN**
## Summary

- Enables the undefined behavior check for shift operations (`--ub-shift-check`) **by default**, so users no longer need to pass the flag explicitly.
- Adds `--no-ub-shift-check` to opt out, consistent with other standard checks (`--no-bounds-check`, `--no-div-by-zero-check`, etc.).
- Includes `--no-ub-shift-check` in the `--no-standard-checks` umbrella flag.
- Keeps `--ub-shift-check` accepted for backward compatibility.

Fixes #2789

### CODE REVIEW

File: src/goto-programs/goto_check.cpp [hunk 1]
```diff
@@ -29,7 +29,7 @@ class goto_checkt
       enable_overflow_check(options.get_bool_option("overflow-check")),
       enable_unsigned_overflow_check(
         options.get_bool_option("unsigned-overflow-check")),
-      enable_ub_shift_check(options.get_bool_option("ub-shift-check")),
+      disable_ub_shift_check(options.get_bool_option("no-ub-shift-check")),
```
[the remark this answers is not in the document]
NripeshN: This would be inconsistent with the rest of the codebase. No other default check has a re-enable flag:

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -435,6 +435,9 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
   if (cmdline.isset("ub-shift-check"))
     options.set_option("ub-shift-check", true);
 
+  if (cmdline.isset("no-ub-shift-check"))
+    options.set_option("no-ub-shift-check", true);
```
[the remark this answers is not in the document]
NripeshN: This would be inconsistent with the rest of the codebase. No other default check has a re-enable flag:

lucasccordeiro:
@NripeshN: We should actually expand the goto-check to instrument the code so we can detect the negative number.

For this C program:

```C
#include <assert.h>

int main()
{
  int a = -2147483648;
  assert((0 >> a) == 0);
  return 0;
}
```

ESBMC should detect two violations:

- Shift distance is negative in (0 >> a).
- Shift distance too large in (0 >> a).

However, ESBMC's current encoding just detects:

- Shift distance too large in (0 >> a).

NripeshN:
Thanks @lucasccordeiro. I've split the shift UB check into separate claims so each violation is reported independently:

- `shift distance is negative` (right_op >= 0)
- `shift distance too large` (right_op < bit_width)
- `shift operand is negative` (left_op >= 0, for signed left shift)

With `--multi-property` on your test case, ESBMC now reports:

```
✗ FAILED: 'shift distance is negative'
✓ PASSED: 'shift distance too large'
```

For `a = -2147483648`, the "too large" check passes because `-2147483648 < 32` is true in signed comparison. Both violations are detected when using a nondet input (the solver finds `a = -1` for negative and `a = 32` for too-large independently).

intrigus-lgtm (changes requested):
We should remove "ub-shift-check" if it's the default.
This should be consistent with how the other flags are used.

Are there any docs where we mention those flags?

Also, nice change to add better error messages 🎉

NripeshN:
@intrigus-lgtm Done -- removed `--ub-shift-check` entirely. Only `--no-ub-shift-check` remains, consistent with the other default checks. Updated all 27 regression tests that referenced it. No docs mention the flag.

regression/esbmc/ub_09_shift_default_check_fail/main.c
```diff
@@ -1,7 +1,7 @@
 /*
  * Regression test for GitHub issue #2789.
  * The shift UB check should be enabled by default so that a nondet
- * shift distance is flagged without requiring --ub-shift-check.
+ * shift distance is flagged without requiring.
```
intrigus-lgtm: Broken comment

lucasccordeiro:
@NripeshN: Could you please check the conflicts?

regression/esbmc/ub_09_shift_default_check_fail/main.c
NripeshN: Fixed -- the sed that removed `--ub-shift-check` from test files also mangled the comment. Rebased and corrected.

NripeshN:
@lucasccordeiro Rebased on latest master, conflicts resolved.

lucasccordeiro (changes requested):
@NripeshN: as part of your PR, can you also update the sv-comp script to disable this check around line 244?

https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/esbmc-wrapper.py

If I merge your PR as is, I'm sure we'll find many bugs in the sv-comp benchmarks, but we won't have the capacity to report and fix them for the sv-comp community.

NripeshN:
Done ✅

lucasccordeiro (approved):

lucasccordeiro:
Thanks for submitting this PR, @NripeshN.

src/goto-programs/goto_check.cpp
```diff
@@ -282,7 +282,7 @@ void goto_checkt::overflow_check(
 {
   if (
     !enable_overflow_check && !enable_unsigned_overflow_check &&
-    !enable_ub_shift_check)
+    disable_ub_shift_check)
```
lucasccordeiro, line 285: `disable_ub_shift_check` is false by default, so this early return never fires and add/sub/mul/div overflow checks now run without `--overflow-check`. Keep the shift term scoped to shifts.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1998,6 +1998,10 @@ bool esbmc_parseoptionst::process_goto_program(
       options.set_option("no-vla-size-check", true);
       options.set_option("no-align-check", true);
       options.set_option("no-bounds-check", true);
+      options.set_option("no-ub-shift-check", true);
+      //?
```
lucasccordeiro, line 2002: `//?` and two commented-out `set_option` lines left in. Remove them.

src/esbmc/options.cpp
```diff
@@ -535,9 +535,9 @@ const struct group_opt_templ all_cmd_options[] = {
     {"unsigned-overflow-check",
      NULL,
      "Enable arithmetic over- and underflow check for unsigned integers"},
-    {"ub-shift-check",
+    {"no-ub-shift-check",
```
lucasccordeiro, line 538: Deleting `--ub-shift-check` breaks every existing command line that passes it. Keep it as a deprecated no-op.

regression/esbmc/ub_01_shift_left_negative_op2/test.desc
```diff
@@ -1,4 +1,4 @@
 CORE
 main.c
---ub-shift-check
+
```
lucasccordeiro, line 3: No test exercises the new `--no-ub-shift-check` flag. Add one asserting the shift UB is suppressed when it is passed.

lucasccordeiro (changes requested):

src/goto-programs/goto_check.cpp
```diff
@@ -581,21 +581,31 @@ void goto_checkt::shift_check(
 
   expr2tc right_op_size_check = lessthan2tc(right_op, left_op_type_size);
 
-  expr2tc ub_check = and2tc(right_op_non_negative, right_op_size_check);
+  add_guarded_claim(
+    right_op_non_negative,
+    "shift distance is negative",
+    "undef-behavior",
```
lucasccordeiro, line 587: `src/util/base/cwe_mapping.cpp` keys CWE-1335 on the literal "undefined behavior on shift operation"; these renamed claims no longer match. Add the new messages to that table.

src/goto-programs/goto_check.cpp
```diff
@@ -581,21 +581,31 @@ void goto_checkt::shift_check(
...
+    loc,
+    guard);
+
+  add_guarded_claim(
+    right_op_size_check,
+    "shift distance too large",
+    "undef-behavior",
```
lucasccordeiro, line 594: `regression/esbmc/shift_assign_count_type_fail/test.desc` pins `^  undefined behavior on shift operation lshr$` and is not updated here. Update it for the renamed claims.

src/goto-programs/goto_check.cpp
```diff
@@ -581,21 +581,31 @@ void goto_checkt::shift_check(
...
+    loc,
+    guard);
 
   if (is_shl2t(expr) && is_signedbv_type(left_op))
   {
```
lucasccordeiro, line 599: Master gates this claim on `config.language.cpp_std < cxx_stdt::cpp20` (#4241). Keep that gate when rebasing, or C++20 negative left shifts are wrongly flagged by default.

regression/esbmc/ub_09_shift_default_check_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone matches any failing property. Add a regex pinning `shift distance is negative` so the test proves the default check fired.

regression/esbmc/ub_10_shift_default_check_pass/main.c
```diff
@@ -0,0 +1,21 @@
...
+ * Regression test for GitHub issue #2789.
+ * Safe shift operations with constant distances should pass
+ * verification even with the default UB shift check enabled.
+ */
+
+#include <assert.h>
+
+int main()
+{
+  int a = 256 << 1;
```
lucasccordeiro, line 11: These constant shift claims simplify to true and are dropped, so this test also passes with `--no-ub-shift-check`. Use a nondet, assumed-in-range shift distance.

scripts/competitions/svcomp/esbmc-wrapper.py
```diff
@@ -264,6 +264,9 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
   witness_name = os.path.basename(benchmark) if esbmc_ci else "witness"
   command_line += "--witness-output " + witness_name + " "
 
+  # Disable the default shift UB check for SV-COMP
+  command_line += "--no-ub-shift-check "
```
lucasccordeiro, line 268: This disables the new default for every property, so SV-COMP never exercises it. Scope it to the properties where shift UB is out of scope.

PR #3911, opened 2026-03-27
[clang-c] handle C11 atomic cast operators in frontend

### LINKED ISSUE #3191: [Bug] Cannot handle C11 _Atomic types

NripeshN:
Handle `CK_NonAtomicToAtomic` and `CK_AtomicToNonAtomic` implicit casts so that `_Atomic` qualified variables can be parsed and verified. ESBMC already strips `_Atomic(T)` to `T` in the type converter, so these casts just need `gen_typecast`.

Before this patch, any use of `_Atomic int x = 0;` would fail with:
```
ERROR: Conversion of unsupported clang cast operator: "NonAtomicToAtomic"
```

Fixes #3191


lucasccordeiro:
@NripeshN: Can you please create multi-threaded programs using pthreads and use the `--data-races-check` option to verify whether your implementation works?

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3211,9 +3211,9 @@ bool clang_c_convertert::get_cast_expr(
 #endif
 
   case clang::CK_NonAtomicToAtomic:
-    if (config.options.get_bool_option("dont-care-about-missing-extensions"))
-      break;
-    [[fallthrough]];
+  case clang::CK_AtomicToNonAtomic:
+    gen_typecast(ns, expr, type);
+    break;
```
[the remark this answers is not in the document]
NripeshN: The type converter already silently strips `_Atomic(T)` to `T` with a FIXME at line 1250 noting that irep2 has no atomic representation yet. Adding a warning at the cast level would fire on every implicit atomic conversion (potentially many times per TU) for something the codebase already acknowledges and intentionally ignores. I think a warning belongs with the type-level FIXME if the maintainers want one, not at each cast site.

NripeshN:
@lucasccordeiro Added two pthread regression tests:

- `c11_atomic_03_pthread_race` -- two threads doing `counter = counter + 1` on an unprotected `_Atomic int`, detected as a W/W data race with `--data-races-check`
- `c11_atomic_04_pthread_safe` -- same but with mutex protection, passes verification

Both parse and verify correctly with the patch.

regression/esbmc/c11_atomic_03_pthread_race/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--data-races-check
+^VERIFICATION FAILED$
```
intrigus-lgtm: According to [TSAN](https://clang.llvm.org/docs/ThreadSanitizer.html), there is no data race in this code:
https://www.godbolt.org/z/xfKsW7qG9
(Remove the _Atomic and it complains)

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3211,9 +3211,9 @@ bool clang_c_convertert::get_cast_expr(
 #endif
 
   case clang::CK_NonAtomicToAtomic:
-    if (config.options.get_bool_option("dont-care-about-missing-extensions"))
-      break;
-    [[fallthrough]];
+  case clang::CK_AtomicToNonAtomic:
+    gen_typecast(ns, expr, type);
```
intrigus-lgtm, line 3215: I haven't thought about this deeply, but this looks too simple.
_Somewhere_ we have to do an atomic load/store when interacting with atomic types; maybe not here, but somewhere else.

intrigus-lgtm (changes requested):
This doesn't fully fix https://github.com/esbmc/esbmc/issues/3191 (if it's intended to do that)
We need to lower _Atomic types to atomic loads and stores where appropriate.
I think we already support certain atomics via our support for gcc builtins?

src/clang-c-frontend/clang_c_convert.cpp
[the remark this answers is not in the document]
NripeshN: You're right -- since ESBMC currently desugars `_Atomic(T)` to `T` at the type level (line 1250), the implicit casts here just propagate that. Proper atomic semantics would need the type system to preserve `_Atomic` and then lower reads/writes to the appropriate `__c11_atomic_load`/`__c11_atomic_store` builtins (which ESBMC already handles via `get_atomic_expr`). That's a bigger change though -- this PR just unblocks parsing so `_Atomic` programs don't crash.

regression/esbmc/c11_atomic_03_pthread_race/test.desc
NripeshN: Good catch. You're right -- per C11, `_Atomic` accesses are race-free by definition. Since ESBMC currently strips the qualifier and doesn't model the implicit synchronization, it false-positives here. I'll remove this test -- it's testing a known limitation, not the cast fix.

NripeshN:
Closing -- the fix was merged via #3949. Thanks for the reviews.

PR #3912, opened 2026-03-27
[goto-programs] detect pointer-based loop bounds for range-for unrolling

### LINKED ISSUE #2771: [CPP] Couldn't determine uper bound for range-based for loop
```c++
#include <iostream>
#include <vector>

int main() {
    std::vector<int> numbers = {10, 20, 30, 40};

    std::cout << "Vector elements: ";
    for (int num : numbers) {
        std::cout << num << " ";
    }
    return 0;
}


```

tong@tong:~/esbmccpp$ esbmc main.cpp
Parsing main.cpp
Converting
Generating GOTO Program
GOTO program creation time: 0.647s
GOTO program processing time: 0.017s
WARNING: no body for function piecewise_construct_t#
WARNING: no body for function allocator#
Unwinding loop 40 iteration 1   file /tmp/esbmc-cpp-headers-5c72-09fb-217d/vector line 358 column 5 function vector
Unwinding loop 40 iteration 2   file /tmp/esbmc-cpp-headers-5c72-09fb-217d/vector line 358 column 5 function vector
Unwinding loop 40 iteration 3   file /tmp/esbmc-cpp-headers-5c72-09fb-217d/vector line 358 column 5 function vector
Unwinding loop 40 iteration 4   file /tmp/esbmc-cpp-headers-5c72-09fb-217d/vector line 358 column 5 function vector
Unwinding loop 39 iteration 1   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 2   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 3   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 4   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 5   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 6   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 7   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 8   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 9   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 10   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 11   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 12   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 13   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 14   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 15   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 16   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 17   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 18   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 19   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 20   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 21   file main.cpp line 8 column 5 function main
Unwinding loop 39 iteration 22   file main.cpp line 8 column 5 function main

NripeshN:
Extend `bounded_loop_unroller` to recognize pointer iteration patterns (`ptr != end; ptr++`) produced by C++ range-for over vectors/arrays. When both init and end share the same base symbol with constant offsets, the bound is resolved statically.

Before this, `for (int x : vec)` with a 4-element vector would unwind indefinitely since the pattern matcher only handled integer-indexed `i < K` loops. Now it terminates in ~2s.

Fixes #2771


XLiZHI:
Hi @NripeshN,

In the regression test, I noticed that you used --incremental-bmc. However, in the current master branch, they can all be verified very quickly. 

If you wish to test for static loop bounds, then any strategies such as incremental or induction should not be used (as they will ignore the known bounds).

Additionally, IIUC, the `unroll_loop` function will only be enabled when `goto-unwind` is used. By default, it is not activated.

NripeshN:
Thanks for the review @XLiZHI, you're right. I verified that `--incremental-bmc` already handles this correctly on master (the forward condition resolves at k=5). My regression tests were using `--incremental-bmc` which masked the fact that the static unroller pattern wasn't actually matching the post-inlining GOTO IR for vector range-for loops.

After further investigation, the pointer iteration pattern after full inlining is more complex than the simple `ptr != end; ptr++` form I was matching. The `bounded_loop_unroller` would need a deeper rewrite to handle the actual IR shape.

Since `--incremental-bmc` already solves the original reported problem, I'll close this PR. The issue may still be relevant for plain BMC mode with `--goto-unwind`, but that would need more investigation into what the GOTO IR actually looks like after inlining.

PR #3914, opened 2026-03-27
python: support dict.update(other_dict)

brcfarias:
This PR adds support for `dict.update(other_dict)`.

Example:
```python
d = {0: 1}
other = {1: 2}
d.update(other)
assert d[0] == 1
assert d[1] == 2
```
Previously this only worked when update() received a dict literal directly as shown below:
```text
ERROR: update() argument must be a dict literal in ESBMC model
```
This change makes it also work with another dictionary.

Needed by #3915


lucasccordeiro (approved):

Issue #3915, opened 2026-03-27
Error on regression/quixbugs/shortest_path_lengths


brcfarias:
main.py:
```python
from collections import defaultdict

def shortest_path_lengths(n, length_by_edge):
    length_by_path = defaultdict(lambda: float('inf'))
    length_by_path.update({(i, i): 0 for i in range(n)})
    length_by_path.update(length_by_edge)

    for k in range(n):
        for i in range(n):
            for j in range(n):
                length_by_path[i, j] = min(
                    length_by_path[i, j],
                    length_by_path[i, k] + length_by_path[k, j]
                )

    return length_by_path

def test1():
    """Case 1: Basic graph input."""

    graph = {
        (0, 2): 3,
        (0, 5): 5,
        (2, 1): -2,
        (2, 3): 7,
        (2, 4): 4,
        (3, 4): -5,
        (4, 5): -1,
    }
    result = shortest_path_lengths(6, graph)

    expected = {
        (0, 0): 0,
        (1, 1): 0,
        (2, 2): 0,
        (3, 3): 0,
        (4, 4): 0,
        (5, 5): 0,
        (0, 2): 3,
        (0, 5): 4,
        (2, 1): -2,
        (2, 3): 7,
        (2, 4): 2,
        (3, 4): -5,
        (4, 5): -1,
        (0, 1): 1,
        (0, 3): 10,
        (0, 4): 5,
        (1, 0): float("inf"),
        (1, 2): float("inf"),
        (1, 3): float("inf"),
        (1, 4): float("inf"),
        (1, 5): float("inf"),
        (2, 0): float("inf"),
        (2, 5): 1,
        (3, 0): float("inf"),
        (3, 1): float("inf"),
        (3, 2): float("inf"),
        (3, 5): -6,
        (4, 0): float("inf"),
        (4, 1): float("inf"),
        (4, 2): float("inf"),
        (4, 3): float("inf"),
        (5, 0): float("inf"),
        (5, 1): float("inf"),
        (5, 2): float("inf"),
        (5, 3): float("inf"),
        (5, 4): float("inf"),
    }

    assert result == expected


def test2():
    """Case 2: Linear graph input."""

    graph = {
        (0, 1): 3,
        (1, 2): 5,
        (2, 3): -2,
        (3, 4): 7,
    }
    result = shortest_path_lengths(5, graph)

    expected = {
        (0, 0): 0,
        (1, 1): 0,
        (2, 2): 0,
        (3, 3): 0,
[...]

PR #3918, opened 2026-03-28
python: support dictionary comprehensions

brcfarias:
This PR adds support for dictionary comprehensions.

Example:
```python
d = {i: i for i in range(1)}
assert d[0] == 0
```

Previously, ESBMC failed with:
```text
ERROR: Unsupported expression DictComp at line 1
```
Needed by #3915


lucasccordeiro (approved):

PR #3920, opened 2026-03-28
[deref] fix spurious failure for scalar-to-struct/union type-punning

### LINKED ISSUE #3919: [pointer-analysis] spurious "Union pointer pointed at scalar" dereference failure on type-punning via pointer cast
ESBMC reports a dereference failure when casting a pointer to a different (but layout-compatible) type and dereferencing it. The reported error is spurious and caused by overly strict memory modeling rather than an actual invalid memory access.

Minimal Reproducer:
```c
typedef union {
	struct { int z; } x;
} t1;

typedef struct {
	int y[2];
} t2;

t1 v1;
t2 v2;

int main()
{
	v1 = *(t1 *)&v2;
}
```

---

ESBMC Output:

```
Violated property:
  dereference failure: Union pointer pointed at scalar

VERIFICATION FAILED
```

This program should not trigger a dereference failure. While the code may violate strict aliasing rules and thus be technically undefined behavior under the C standard, it does not perform an invalid memory access:

* `v2` is properly allocated
* The cast does not produce an invalid pointer
* The dereference reads valid memory within bounds

At most, this case should be treated as:

* undefined behavior (if ESBMC models strict aliasing), or
* allowed reinterpretation (if ESBMC permits low-level memory reasoning)

But it should **not** be classified as a dereference failure.

This suggests that ESBMC rejects reinterpretation of memory across incompatible types, even when:

* the pointer is valid
* the memory is in-bounds
* no actual invalid dereference occurs

### Analysis

The issue seems to stem from ESBMC’s memory model enforcing strict type-based separation of objects. In this case:

* `v2` is modeled as a `struct`
* It is cast to a `t1*` (union type)
* ESBMC rejects the dereference due to type mismatch rather than memory safety concerns

This leads to a **spurious counterexample**.

### Suggested Improvements

* Distinguish between:

  * invalid pointer dereferences (true errors)
  * type-punning / aliasing violations (undefined but not memory-unsafe)
* Optionally introduce:
  * a warning or UB flag for strict aliasing violations
  * a relaxed mode allowing byte-level reinterpretation (similar to `memcpy` semantics)

lucasccordeiro:
Fixes #3919.

In `build_reference_rec()`, the `flag_src_scalar | flag_dst_struct and flag_src_scalar | flag_dst_union` cases with a constant offset unconditionally emitted a dereference failure. 

This caused false positives when valid C code reinterpreted a scalar as a struct or union via a pointer cast (e.g., v1 = *(t1 *)&v2), because the dereference layer drills through array members to a scalar element and then needs to reconstruct the composite type from it.

In this PR, when access starts at offset 0, and the scalar and destination types have the same bit width, emit a bitcast instead of failing. This models valid C-style type-punning without relaxing any actual memory-safety checks.


lucasccordeiro:
Master (30s): https://github.com/esbmc/esbmc/actions/runs/23692011062.
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/23692014313.

PR #3922, opened 2026-03-29
[coverage] fix multi-file input ignoring extra source files

ChenfengWei0:
Coverage instrumentation improvements and reporting

  - Fix multi-file coverage: extra source files (cmdline.args[1:]) are now included in the instrumentation location
  pool, fixing ignored coverage for secondary input files
  - --cov-assume-asserts: convert user assertions to assumptions instead of assert(true) to preserve path constraints,
  preventing false coverage reports
  - Suppress non-coverage output: in coverage mode, only instrumented coverage claims are reported to the terminal;
  converted assert(true) noise is silenced
  - --cov-report-json: output structured JSON coverage report (cov-report.json) with per-claim source locations and
  coverage status
  - HTML coverage report: add scripts/cov-report.py to generate lcov-style HTML reports from the JSON output, with
  source annotation and per-file coverage bars

Assisted-by: Claude-Opus4.6

图片


lucasccordeiro (approved):
Nice PR, @ChenfengWei0.

Can you also submit another PR to update our documentation at https://esbmc.github.io/docs/coverage/?

PR #2926, opened 2025-10-23
[SV-COMP] refactor the data race check

XLiZHI:
`races_check[][] = array_of(array_of(0));`
It is not feasible to create an unbounded 2D array. 

Therefore, in this new method, we try to flatten the 2D array to keep track of the variable.

PR #3924, opened 2026-03-30
[smt] fix int-to-ptr fallback for byte-updated pointer patterns

### LINKED ISSUE #2153: [firmware] incorrect true
Command-line: $ esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check tdg_servtd_wr__cover__success_havoc_memory.i --64 --witness-output witness.graphml --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3

Program: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/intel-tdx-module/tdg_servtd_wr__cover__success_havoc_memory.i?ref_type=heads

`c/intel-tdx-module/tdg_servtd_wr__cover__success_havoc_memory.i`

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2153.

When a pointer is nondeterministically byte-manipulated through a char* cast (void*, struct field, or whole-struct cast), converting the result back to a pointer via `convert_typecast_to_ptr` should not pin the fallback to the INVALID object.  Pinning to INVALID (a constant) conflicted with the negated assertion, preventing the solver from finding the counterexample.

This PR replaces the fallback with `not_matched → addr_space[output.obj].start + output.offset = target`, which gives the solver freedom to pick any object whose address equals the target integer.  It also extends pattern detection to include struct-field extraction (via extract2t, wrapping byte_update2t).  Lastly, it adds regression tests github_2153_2 and github_2153_3.

### DIFF
--- a/regression/esbmc/github_2153/test.desc
+++ b/regression/esbmc/github_2153/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
-
+--incremental-bmc
 ^VERIFICATION FAILED$
--- a/regression/esbmc/github_2153_2/main.c
+++ b/regression/esbmc/github_2153_2/main.c
@@ -1,0 +1,18 @@
+// Variant of github_2153: byte-manipulation via struct-field pointer cast
+// at a non-zero byte offset.  After ptr[1] = nondet_char(), bug.bar can
+// represent any pointer value, so the assertion must fail.
+
+struct
+{
+  void *bar;
+} bug;
+
+int some_var;
+
+int main()
+{
+  char *ptr = (char *)&bug.bar;
+  ptr[1] = nondet_char();
+
+  __ESBMC_assert(bug.bar != &some_var, "");
+}
--- a/regression/esbmc/github_2153_2/test.desc
+++ b/regression/esbmc/github_2153_2/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_2153_2/test.desc
+++ b/regression/esbmc/github_2153_2/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
-
+--incremental-bmc
 ^VERIFICATION FAILED$
--- a/regression/esbmc/github_2153_3/main.c
+++ b/regression/esbmc/github_2153_3/main.c
@@ -1,0 +1,19 @@
+// Variant of github_2153: byte-manipulation through a char* cast to the whole
+// struct (not to a specific field).  The goto-symex produces a struct-level
+// byte_update, so bug.bar is extracted from byte_update<struct>(...).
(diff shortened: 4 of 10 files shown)


lucasccordeiro:
Master (30s): https://github.com/esbmc/esbmc/actions/runs/23729386049.

```
  correct:           19382
    correct true:    11984
    correct false:    7398
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17193
  Score:             30806 (max: 61270)
```

This PR (30s): https://github.com/esbmc/esbmc/actions/runs/23729393558.

``` 
  correct:           19283
    correct true:    11898
    correct false:    7385
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17292
  Score:             30621 (max: 61270)
```

XLiZHI:
Hi Prof @lucasccordeiro,

Thanks for your PR.

https://github.com/esbmc/esbmc/commit/dd19ab5cfad6f75f2c8bd0100e76def2781d7aaa

My solution approach two years ago was similar to this one. However, during my investigation, I discovered that there were some issues with this constraint. In the address space of ESBMC, the address (address_start + offes) does not provide sufficient constraints.

At that time, I was looking for a complete solution, such as constraints like "object -> target_object && offset -> target_offset".

I compared the results of CBMC on [test_case](https://github.com/esbmc/esbmc/pull/3924/changes#diff-6b86cc09f2a17debdfd8a4a5866a7ba694419bca71906a47551a20a534a14cc2).

```

State 64 file main2.c function init line 10 thread 0
  byte_extract_little_endian(local_data.ptr, 7l, char)=(char)(irep("(\"bv\" \"width\" (\"64\"))"))(const void *)&some_var[56l + 7l, 56l] (?)
```

In the final trace, the ptr of local_data was properly pointed to some_var.

In this PR, ESBMC output:

```
State 8 file main2.c line 10 column 9 function init thread 0
  (signed char *)base[i] = ( struct local_s { void * ptr; })(byte_update_little_endian((unsigned long int)({ .ptr=INVALID67108865 }), 7, 0))
```

`INVALID67108865` indicates that the constraints are insufficient.

I tried my best to avoid using "address" directly. That's why I made this change for the data races check #2926.

_Update_: Before we find the correct implementation method, I suggest we keep this PR. **Incorrect true** is something we should avoid at all costs.

lucasccordeiro:
Thanks, @XLiZHI.

I agree with you; these constraints we introduced with this PR are not enough.

> Update: Before we find the correct implementation method, I suggest we keep this PR. Incorrect true is something we should avoid at all costs.

Thank you!

I'm currently running SV-COMP with a longer timeout to check whether this PR reports any false alarms.

PR #3926, opened 2026-03-30
ci: restore Codecov integration with dedicated coverage workflow

lucasccordeiro:
Add coverage.yml workflow that builds ESBMC with GCC coverage instrumentation, runs ctest, filters lcov data, and uploads to Codecov on every push to master. Add -k ON|OFF flag to build.sh to enable ENABLE_COVERAGE and install lcov. Fix Coverage.cmake to accept GCC and Clang (not GCC-only), simplify to --coverage (dropping the redundant -fprofile-arcs -ftest-coverage), and add a codecov.yml for server-side path filtering.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -8,7 +8,6 @@ jobs:
   coverage:
     name: Build and upload coverage to Codecov
     runs-on: ubuntu-24.04
-    timeout-minutes: 90
     env:
       CC: gcc
       CXX: g++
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  coverage:
+    name: Build and upload coverage to Codecov
+    runs-on: ubuntu-24.04
+    timeout-minutes: 90
```
rafaelsamenezes: Not sure about this timeout

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,49 @@
...
+    env:
+      CC: gcc
+      CXX: g++
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies and build with coverage
+        run: ./scripts/build.sh -b Debug -k ON deps build
```
rafaelsamenezes: Not sure about this

rafaelsamenezes (approved):

PR #3931, opened 2026-03-31
[ir-ra] rename --ir-ra option to --ir-ieee

ShemaAlosaimi:
## Summary

This PR renames the user-facing option `--ir-ra` to `--ir-ieee`.

The change is limited to:
- option registration/parsing
- solver call sites using the option
- regression `.desc` files invoking the flag
- user-facing documentation/comments that mention the CLI flag by name

No directory names or internal identifiers are changed.

## Validation

I checked the rename end-to-end with:
- an existing regression that still produces `VERIFICATION SUCCESSFUL` under `--ir-ieee`
- a nearest-mode theorem-driven regression that still reaches the expected SMT enclosure path under `--ir-ieee`

### DIFF
--- a/regression/esbmc/github_2572/test.desc
+++ b/regression/esbmc/github_2572/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
---z3 --ir-ra
+--z3 --ir-ieee
 ^VERIFICATION SUCCESSFUL$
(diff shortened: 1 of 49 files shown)


lucasccordeiro (approved):

PR #3941, opened 2026-03-31
[bug fix] --no-standard-checks not suppressing VLA size check

ChenfengWei0:
`--no-standard-checks` was not suppressing the "VLA array size in bytes overflows address space size" assertion. The root cause is a
  timing issue: `--no-standard-checks` was expanded into individual `no-*` options only in `process_goto_program()`, which runs **after**
  `goto_convert()`. However, VLA size check assertions are generated during `goto_convert()` by `generate_dynamic_size_vla()`, which reads
  the `no-vla-size-check` option — an option that wasn't set yet at that point.

  Using `--no-vla-size-check` directly always worked, because it is parsed from the command line before `goto_convert()`.

  **Reproducer** (from bug report):
  ```c
  int main()
  {
    int SIZE = 16;
    int vec[ SIZE ];
    int max = vec[0];
    int idx_v0;
    int idx;
    for ( idx = 1; idx < SIZE; ++idx )
    { if ( vec[idx] >= max )
        max = vec[ idx ];
    }
    __ESBMC_assert(
      __ESBMC_forall(
        &idx_v0,
        !( 0 <= idx_v0 ) ||
        !( idx_v0 < SIZE ) ||
        ( vec[ idx_v0 ] <= max )
      ),
      "end of loop assertion"
    );
    return 0;
  }
```
  esbmc --show-claims --no-standard-checks max-1.c incorrectly showed:
  Claim 1: VLA array size in bytes overflows address space size
  Claim 2: end of loop assertion

  After fix, only the user assertion remains:
  Claim 1: end of loop assertion

  Fix

  - Expand --no-standard-checks into individual options before goto_convert() in parse_goto_program()
  - Keep the expansion in process_goto_program() (idempotent) to cover the read_goto_binary path before goto_check()

  Test plan

  - Added regression test regression/esbmc/no_standard_checks_vla/ using original bug report program
  - Verified --show-claims --no-standard-checks shows only user assertion (no VLA claim)
  - Verified VLA claim still appears without --no-standard-checks

  Assisted-by: Claude Opus 4.6

图片

### DIFF
--- a/regression/esbmc/no_standard_checks_vla/main.c
+++ b/regression/esbmc/no_standard_checks_vla/main.c
@@ -1,0 +1,8 @@
+int main()
+{
+  int SIZE = 16;
+  int vec[SIZE];
+  int x = vec[0];
+  __ESBMC_assert(SIZE == 16, "size is 16");
+  return 0;
+}
--- a/regression/esbmc/no_standard_checks_vla/main.c
+++ b/regression/esbmc/no_standard_checks_vla/main.c
@@ -1,8 +1,26 @@
 int main()
 {
   int SIZE = 16;
-  int vec[SIZE];
-  int x = vec[0];
-  __ESBMC_assert(SIZE == 16, "size is 16");
+  int vec[ SIZE ];
+  int max = vec[0];
+  int idx_v0;
+  int idx;
+  //------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lucasccordeiro (approved):

PR #3942, opened 2026-04-01
[goto] fix quantifier body structure lost by remove_sideeffects for || / && chains

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/discussions/3886.

**Problem**: When an `__ESBMC_forall/__ESBMC_exists` body was a `||` or `&&` expression containing a nested quantifier call, `remove_sideeffects()` converted it into a short-circuit ITE GOTO chain, reducing the whole body to a single `temp` symbol.  The outer forall then held `forall(&b, t)` where `b` no longer appeared in `t`, so `replace_name_in_body()` in `smt_conv.cpp` could not substitute the bound variable, producing a vacuously true formula.

**Fix**: This PR mirrors the existing `__ESBMC_loop_invariant` special case: flatten the top-level `||` or `&&` chain, process each operand independently, and rebuild the plain `or`/`and` expression so the logical structure is preserved.


LukeW1999:
LGTM, I would check if it works for nested `||` expression for quantifier.

LukeW1999 (commented):
Here is the minimal case I used to test my review:

**Case 1** — `forall b: (b>0) || forall(a, a==b)` (counterexample: b=0, a=1), expected: `VERIFICATION SUCCESSFUL`
```c
int main() {
    int a1, b1;
    __ESBMC_assert(
        __ESBMC_forall(&b1, (int)((b1 > 0) || __ESBMC_forall(&a1, a1 == b1))) == 0,
        "forall b: (b>0)||forall(a,a==b) should be false");
    return 0;
}
```

**Case 2** — `exists b: (b==5) && ((b!=5) || forall(a, a==b))` (at b=5: b!=5 false; forall a: a==5 false), expected: `VERIFICATION SUCCESSFUL`
```c
int main() {
    int a1, b1;
    __ESBMC_assert(
        __ESBMC_exists(
            &b1, (b1 == 5) && ((b1 != 5) || __ESBMC_forall(&a1, a1 == b1))) == 0,
        "exists b: (b==5)&&((b!=5)||forall(a,a==b)) should be false");
    return 0;
}
```

src/goto-programs/goto_sideeffects.cpp
```diff
@@ -365,6 +387,55 @@ void goto_convertt::remove_sideeffects(
...
+        exprt::operandst &args = expr.op1().operands();
+        if (args.size() == 2 && has_sideeffect(args[1]))
+        {
+          exprt *body_expr = &args[1];
+          // Look through any implicit typecast that Clang may insert
+          if (
+            body_expr->id() == "typecast" && body_expr->operands().size() == 1)
+            body_expr = &body_expr->op0();
+
+          if (body_expr->is_or() || body_expr->is_and())
```
LukeW1999, line 426: This only checks whether the top-level body expression is || or &&.

A sub-expression that is itself a || or && containing a nested quantifier call would still fall through to the short-circuit path.

For example:
```
  __ESBMC_forall(&b, (A && B) || __ESBMC_forall(&a, a == b))
```
We need to apply the flattening recursively to sub-expressions as well.

lucasccordeiro: Fixed!

LukeW1999:
LGTM

LukeW1999 (approved):

PR #3944, opened 2026-04-01
[loop invariant] fix side-effect extraction for #3939

LukeW1999:
Fix #3939


lucasccordeiro (approved):

PR #3948, opened 2026-04-01
python: improve function name resolution

brcfarias:
This PR improves type resolution in the Python frontend.

Example:
```python
  from typing import List

  def foo():
      def bar(x: int) -> int:
          return x
      return bar

  f = foo()
  assert f(1) == 1
```

ESBMC was failing to resolve names and type information correctly in cases involving nested functions as shown below:
```text
ERROR: Variable 'bar' is not defined in function 'foo' at line 5.
[1]    355757 IOT instruction (core dumped)  esbmc regression/python/nested-function-return/main.py
```
This change improves symbol resolution and preserves the information needed to handle these cases during conversion.


lucasccordeiro (approved):

PR #3953, opened 2026-04-02
fix for reverse_linked_list_fail and promove test to CORE

sergillam:
This PR applies targeted fixes in the Python frontend (preprocessor, annotation inference, and converter) to correctly handle the reverse_linked_list_fail scenario, and reclassifies reverse_linked_list_fail from KNOWNBUG to CORE after validation.


brcfarias (changes requested):
Could you please provide a more detailed explanation of this PR?
Also, could you add a regression test that would fail without the change for each fix?
It might be better to split these changes into separate PRs, each with its corresponding tests, as this one seems to mix different things.

lucasccordeiro (approved):

PR #3959, opened 2026-04-02
fix: change std::format to fmt::format as it was too new for some

Yiannis128:


LukeW1999 (approved):
tested. It is working

lucasccordeiro (approved):

PR #3968, opened 2026-04-03
[goto] model C11 _Atomic assignments as separate atomic ops (DO NOT MERGE)

### LINKED ISSUE #3955: [Concurrency] support for atomic types
The assignment of atomic types has the following scenarios:

```c
int main() {
    _Atomic int a = 1;
    _Atomic int b = 2;
    int x = 0;
    int y = 0;

    // 1. Atomic on the left-hand side: atomic write (store)
    a = 10;   // Equivalent to: atomic_store(&a, 10);

    // 2. Atomic on the right-hand side: atomic read (load)
    x = a;    // Equivalent to: x = atomic_load(&a);

    // 3. Both sides are atomic:
    //    - Read from b is atomic
    //    - Write to a is atomic
    //    - But the whole assignment is NOT a single atomic operation
    a = b;    // Equivalent to: tmp = atomic_load(&b); atomic_store(&a, tmp);

    // 4. Right-hand side is atomic:
    //    - Read from b is atomic
    //    - Assignment to y is non-atomic
    y = b;

    // 5. Left-hand side is atomic:
    //    - Write to b is atomic
    //    - Read from x is non-atomic
    b = x;
}
```

One model that I believe in is:
```c
    // 1. a = 10;
    // atomic part: store to a
    __ESBMC_atomic_begin();
    a = 10;
    __ESBMC_atomic_end();

    // 2. x = a;
    // atomic part: load from a
    __ESBMC_atomic_begin();
    tmp = a;
    __ESBMC_atomic_end();
    x = tmp;

    // 3. a = b;
    // atomic parts: load from b, then store to a
    __ESBMC_atomic_begin();
    tmp = b;
    __ESBMC_atomic_end();

    __ESBMC_atomic_begin();
    a = tmp;
    __ESBMC_atomic_end();

    // 4. y = b;
    // atomic part: load from b
    __ESBMC_atomic_begin();
    tmp = b;
    __ESBMC_atomic_end();
    y = tmp;

    // 5. b = x;
    // atomic part: store to b
    tmp = x;
    __ESBMC_atomic_begin();
    b = tmp;
    __ESBMC_atomic_end();
```

Or we could control the context switching based on reading/writing. Maybe this can be achieved here:

https://github.com/esbmc/esbmc/blob/56b507d434aa79047d8f43c4460c69bf9e05432a/src/goto-symex/execution_state.cpp#L781

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/3955.

Each access to a C11 _Atomic-qualified variable is now wrapped in ATOMIC_BEGIN/END in the GOTO program, making it indivisible in the concurrency model.  For assignments where both sides are atomic (e.g., counter = counter + 1), the load and store are emitted as two separate atomic sections with a context-switch window between them, correctly modeling the C11 §5.1.2.4 requirement that they are distinct sequentially-consistent operations rather than one transaction.

This PR also adds five regression tests (github_3955_1 through _5) covering atomic writes, atomic reads, the lost-update scenario, the non-atomic contrast case, and both-sides-atomic assignment with deduplication.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/23954471891.

lucasccordeiro:
We have an SV-COMP regression with this PR (https://github.com/esbmc/esbmc/actions/runs/23954471891).

@XLiZHI: Could you please check this regression?

XLiZHI:
image

XLiZHI:
I conducted some quick checks and found that there was one situation we failed to address: RMW (Read-Modify-Write):

```
_Atomic int x;
x++;
```

should be modeled as:

```
atomic
{
 x = x + 1;
}
```

If in an assignment statement, the write operation on variable x depends on its previous value (where x appears on the right-hand side and participates in the calculation), then this statement must be executed atomically.

XLiZHI:
This PR (30s):

```
  correct:           19388
    correct true:    11989
    correct false:    7399
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17187
  Score:             30817 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24085934995
```

PR #3970, opened 2026-04-04
Extract complex utilities into complex_handler_utils

sergillam:
This PR creates complex_handler_utils as a dedicated utility module for complex number handling, following the same pattern as string_handler_utils.

The main changes are:

- Extract parse_complex_string from a local lambda into a standalone free function, removing the helper lambda trim, parse_double_strict that only existed to support it.
- Move raise_math_real_type_error_expr and raise_math_int_type_error_expr into the complex_utils namespace so they can be reused outside function_call_expr.cpp.
- Replace the hand-written keyword loop in handle_complex() with the existing collect_keyword_values utilities from the string handler.


lucasccordeiro (approved):

Issue #1336, opened 2023-09-12
ESBMC 7.3.0 is crashing with "Can't construct rvalue reference to array type during dereference"


tihanyin:
ESBMC 7.3.0 crass in the folllowing small C program:

command : ` esbmc DATASET/FormAI_75935.c --unwind 1`

This isssue is vey similar to the already closed issue: https://github.com/esbmc/esbmc/issues/1087

```c
//FormAI DATASET v1.0 Category: Data mining ; Style: Dennis Ritchie
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#define MAX_LINE_LENGTH 256
#define MAX_ENTRIES 1000
#define MAX_CHARS_PER_ENTRY 50
#define MAX_VALUES_PER_ENTRY 5

typedef struct {
    char values[MAX_VALUES_PER_ENTRY][MAX_CHARS_PER_ENTRY];
} Entry;

typedef struct {
    char labels[MAX_VALUES_PER_ENTRY][MAX_CHARS_PER_ENTRY];
    int num_entries;
    Entry entries[MAX_ENTRIES];
} Dataset;

// function to parse a CSV file and populate a Dataset struct
void parse_csv_file(char *filename, Dataset *dataset) {
    FILE *fp;
    char line[MAX_LINE_LENGTH];
    char *token;

    fp = fopen(filename, "r");
    if (fp == NULL) {
        printf("Could not open file %s.", filename);
        exit(EXIT_FAILURE);
    }

    // read the first row as labels
    fgets(line, MAX_LINE_LENGTH, fp);
    token = strtok(line, ",");
    int i = 0;
    while (token != NULL) {
        strcpy(dataset->labels[i], token);
        token = strtok(NULL, ",");
        i++;
    }

    // read the rest of the file as entries
    i = 0;
    while (fgets(line, MAX_LINE_LENGTH, fp) != NULL) {
        Entry entry;
        token = strtok(line, ",");
        int j = 0;
        while (token != NULL) {
            strcpy(entry.values[j], token);
            token = strtok(NULL, ",");
            j++;
        }
        dataset->entries[i] = entry;
        i++;
    }
    dataset->num_entries = i;

    fclose(fp);
}

// function to print the dataset
void print_dataset(Dataset dataset) {
    printf("Num Entries: %d\n", dataset.num_entries);
    printf("Labels: ");
    for (int i = 0; i < MAX_VALUES_PER_ENTRY; i++) {
[...]

PR #3971, opened 2026-04-04
[deref] fix crash on struct-with-array-field assignment

lucasccordeiro:
Arrays of struct elements (e.g. `Entry[N]`) were misidentified as byte arrays when the struct happened to be 1 byte wide, because the guard only checked `get_width() == 8`. 

This PR replaces all three byte-array heuristics with `is_byte_type()` (`is_bv_type && width == 8`), which correctly excludes struct, float, and pointer element types regardless of their width.

It fixes the crash: "Can't construct rvalue reference to array type during dereference" on `dataset->entries[i] = entry`, where `Entry` has `char[][]`.


lucasccordeiro:
This PR (30s): 

```
  correct:           19441
    correct true:    12038
    correct false:    7403
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17133
  Score:             30903 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/23973657471
```

Master (30s):

```
  correct:           19434
    correct true:    12030
    correct false:    7404
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17140
  Score:             30888 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/23959577493/job/69887743778
```

intrigus-lgtm:
Nice, this fixes https://github.com/esbmc/esbmc/issues/1336#issuecomment-2738260044 :tada:

PR #3974, opened 2026-04-04
[cprover_library] properly define extern C/C++ IO globals instead of suppressing warnings 

lucasccordeiro:
Fixes #3973.


intrigus-lgtm (changes requested):
This is fundamentally the wrong solution.
We should properly fix those warnings instead of suppressing them.
Fixing this isn't even hard.

E.g. `stdin` is defined to be extern here:
https://pubs.opengroup.org/onlinepubs/007904975/functions/stdin.html
and then e.g. glibc adds non-extern variables here:
https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/libio/stdio.c#L33

So for esbmc we'd just have to add a stdio.c file here:
https://github.com/esbmc/esbmc/tree/1163c47f55f86689e8c1dfa18fc7c256e6d8c4b9/src/c2goto/library

Similar for errno and the C++ version of the io stuff.

lucasccordeiro:
Thanks, @intrigus-lgtm. I agree with you.

Do you want to give it a try and submit a PR that properly fixes the issue?

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/4025.

PR #3975, opened 2026-04-04
[deref] fix crash on empty union field access from byte array

lucasccordeiro:
ESBMC crashes for this SV-COMP benchmark: 

```
$Command: ./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check ../../sv-benchmarks/c/intel-tdx-module/tdh_mng_create__invalid_state_kot_entry__cover_proof_havoc_memory.i --64 --witness-output witness --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3 
```

The main issue is that `construct_struct_ref_from_const_offset_array` was called for both struct and union destination types, but asserted `is_struct_type` only, crashing on any union. 

This PR adds a union branch that mirrors the dynamic-offset counterpart and guards both branches against empty struct/union types (a non-standard GCC extension) by emitting a `dereference_failure` and a `make_failed_symbol`.

### DIFF
--- a/regression/esbmc/github_3976/main.c
+++ b/regression/esbmc/github_3976/main.c
@@ -1,0 +1,18 @@
+#include <stdlib.h>
+
+long a;
+
+typedef union {
+} b;
+
+typedef struct {
+  b c;
+} d;
+
+d *e;
+
+int main() {
+  void *f = malloc(a);
+  e = f;
+  e->c;
+}
--- a/regression/esbmc/github_3976/test.desc
+++ b/regression/esbmc/github_3976/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--no-pointer-check
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_3976/test.desc
+++ b/regression/esbmc/github_3976/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
---no-pointer-check
+--force-malloc-success
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_3976_2/main.c
+++ b/regression/esbmc/github_3976_2/main.c
@@ -1,0 +1,17 @@
+#include <stdlib.h>
+
+typedef union {
+} b;
+
+typedef struct {
+  b c;
+  int a;
+} d;
+
+d *e;
+
+int main() {
+  void *f = malloc(sizeof(d));
+  e = f;
+  e->c;
+}
--- a/regression/esbmc/github_3976_2/test.desc
+++ b/regression/esbmc/github_3976_2/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--force-malloc-success
+^VERIFICATION SUCCESSFUL$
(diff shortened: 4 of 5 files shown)


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/23982071832

intrigus-lgtm:
This TC still fails (`--force-malloc-success`):
```c
#include <stdlib.h>

typedef union {
} b;

typedef struct {
  b c;
  int a;
} d;

d *e;

int main() {
  void *f = malloc(sizeof(d));
  e = f;
  e->c;
}
```
```
State 2 file /tmp/main.c line 16 column 3 function main thread 0
Violated property:
  file /tmp/main.c line 16 column 3 function main
  dereference failure: Object accessed with illegal offset


VERIFICATION FAILED
```

XLiZHI (approved):

intrigus-lgtm (commented):
How is regression/esbmc/github_3976/main.c fixed?
If you remove `--no-pointer-check`, it fails with:
```
State 3 file /tmp/main2.c line 17 column 3 function main thread 0
Violated property:
  file /tmp/main2.c line 17 column 3 function main
  dereference failure: Access to empty union type


VERIFICATION FAILED
```
which doesn't make sense to me?
**Neither clang nor gcc complains about it and also ASAN and UBSAN think this is perfectly fine (after initializing `a` to zero).**
So ESBMC should not say that this code failed to verify unless those checks are SOLELY intended to prevent a crash?
If so, IMHO the error message should clearly state that this is an internal error/unimplemented part.

There also appears to be existing code which handles zero-sized structs in construct_struct_ref_from_const_offset which might be salvageable here.

lucasccordeiro:
Thanks, @intrigus-lgtm.

Initially, I just wanted to fix the crash. 

I'll take a closer look to see if I can properly fix this issue in this PR.

lucasccordeiro:
@intrigus-lgtm: Can you please have another look at this PR? I've added your TC as a regression test.

New SV-COMP run after @intrigus-lgtm's review: https://github.com/esbmc/esbmc/actions/runs/23985704649.

lucasccordeiro:
This PR (30s):

```
  correct:           19397
    correct true:    12004
    correct false:    7393
  incorrect:            23
    incorrect true:      7
    incorrect false:    16
  unknown:           17180
  Score:             30921 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/23985704649/job/69958729272
```

Master (30s):

```
  correct:           19441
    correct true:    12038
    correct false:    7403
  incorrect:            26
    incorrect true:     10
    incorrect false:    16
  unknown:           17133
  Score:             30903 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/23973657471
```

PR #3976, opened 2026-04-04
python: Support int.to_bytes()

brcfarias:
This PR adds support for `int.to_bytes()`.

Example:
```python
def foo(x: int):
    return int.to_bytes(x, 2, "big")

y = foo(255)
assert y[0] == 0
assert y[1] == 255
```

This handles both `int.to_bytes(x, ...)` and `x.to_bytes(...)`, extracts each output byte according to the endianness parameter, and returns a bytes value. On master it was failing with:
```text
Unsupported comparison with unresolved operand type at /tmp/main.py:9
```

### DIFF
--- a/regression/python/int_to_bytes/main.py
+++ b/regression/python/int_to_bytes/main.py
@@ -1,0 +1,14 @@
+def foo(x: int):
+    return int.to_bytes(x, 2, "big")
+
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

PR #3979, opened 2026-04-05
[ir-ra] strengthen tracked-sub interval regression expectations

ShemaAlosaimi:
## Summary

Re-strengthen the tracked-sub regression expectations now that both ieee_sub interval lifting (#3950) and interval propagation across assignments (#3967) are in master.

These expectations were temporarily removed in #3950 because they depended on the propagation fix (#3967), which had not yet landed at the time.

## Changes

Adds the stronger tracked-sub SMT pattern checks back to:

- `ra-interval-lift-sub-both-tracked` — asserts `(- ra_lo::0 ra_hi::0)`
- `ra-interval-lift-sub-both-tracked-single` — same, single-precision
- `ra-interval-lift-sub-one-fresh` — asserts `(- ra_lo::0` prefix (open match)
- `ra-interval-lift-sub-one-fresh-single` — same, single-precision

No source code changes.

## Validation

Ran all 4 regressions locally against the current master build — all pass.


lucasccordeiro (approved):

PR #3984, opened 2026-04-05
Extract dedicated complex_handler class

sergillam:
Extract all complex-number logic from python_converter.cpp and function_call_expr.cpp into a dedicated complex_handler class. This removes ~620 lines of inline lambdas spread across both files and consolidates binary/unary operations, conjugate, abs, pow, cmath.log/log10, type normalization, and symbol caching into a single cohesive handler. All type allocations use cached_double_type() to avoid redundant construction.

Add 24 regression tests covering core operations (arithmetic, unary, abs, conjugate, augmented assignment, bool context, normalization, type errors, equality with NaN/Inf, cmath.log/log10, zero-division) and edge cases (negative/complex exponents, isolated division/subtraction, floordiv/mod TypeError, special pow cases).


lucasccordeiro (approved):

PR #3986, opened 2026-04-06
Centralize complex attribute access (.real, .imag, .conjugate()) in dedicated handler

sergillam:
- Moves .real and .imag attribute resolution from the generic struct fallback in python_converter to complex_handler::handle_attribute_access(). This ensures complex attribute access goes through the same dedicated handler that already manages arithmetic, unary ops, and .conjugate().
- Add handle_attribute_access(obj, attr) to complex_handler  returns member_exprt for "real"/"imag", nil_exprt otherwise
- Hook into two dispatch sites in python_converter.cpp (nested attribute path and direct attribute path), both before the generic struct fallback
- 20 new regression tests: 16 for .real/.imag access (basic, edge values, after operations, expressions, type promotion, identities, mixed arithmetic, chaining, division/power, cross-variable, abs, augmented assignment, reassignment, 2 fail cases) and 4 for .conjugate() (extended, identities, after operations, 1 fail case)


lucasccordeiro (approved):

PR #3990, opened 2026-04-06
Add all() builtin function support

pmatos:
## Summary

- Implement Python's `all()` built-in function for verification, mirroring the existing `any()` handler
- Correctly model empty container elements (`[]`, `()`, `""`, `{}`) as falsy via AST-level detection, since ESBMC's IR represents containers as non-NULL pointers
- Add generator expression lowering (`all(x > 0 for x in lst)`) in the preprocessor, mirroring the existing `_lower_any_genexp()` pattern

## Test plan

- [ ] `regression/python/builtin_all` — 110-line test covering bools, ints (positive/negative/zero/large), floats (0.0/inf/-inf), None, mixed types, empty containers, nested non-empty containers, position-of-falsy ordering
- [ ] `regression/python/builtin_all_fail` — intentionally wrong assertion triggers VERIFICATION FAILED
- [ ] `regression/python/builtin_all_genexp` — generator expressions with and without `if` clauses
- [ ] `regression/python/builtin_all_genexp_fail` — wrong genexp assertion triggers VERIFICATION FAILED
- [ ] Existing `any()` tests pass without regressions


lucasccordeiro (approved):

lucasccordeiro (approved):

pmatos:
@lucasccordeiro I think this is ready.

lucasccordeiro:

PR #3994, opened 2026-04-07
feat: update to bitwuzla 0.9.0

intrigus-lgtm:
EDIT: This has been cleaned up now and is ready.
The current implementation was Codex-generated and is honestly **not good**, but I didn't have time to clean it up yet.
So focus on the general idea, not the concrete implementation.

## Problem
Basically, what was the problem with updating to bitwuzla 0.9.0:
1. Bitwuzla started to statically link to cadical.
2. CVC5 already links statically to cadical.
3. Esbmc links to bitwuzla and CVC5 which fails as there are duplicate symbols for cadical.

## Explored solutions
1. Disable cadical - not really a solution :P
2. Let ESBMC build a specific cadical version and then make cvc5 and bitwuzla link against that => no more duplicate symbols, but then we have to figure out a version that is accepted by both cvc5 and bitwuzla which is a bit annoying.
3. ~~Convert *global* symbols to *local* symbols: we convert all the (global) cadical symbols in bitwuzla and cvc5 to local symbols. When we then link esbmc to bitwuzla and cvc5, there are no duplicate symbols for cadical as duplicate local symbols are fine => cvc5 and bitwuzla can continue to link against whatever version of cadical they want.~~
4. Rename cadical symbols inside bitwuzla, this way they won't clash with the cadical symbols in cvc5.

~~To me, solution 3 felt like the cleanest solution and that's why I implemented it.~~

EDIT: While solution 3 worked locally, it apparently didn't work in CI.
I cleaned up the code and switched to renaming the symbols.

### DIFF
--- a/regression/quixbugs/hanoi/test.desc
+++ b/regression/quixbugs/hanoi/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.py
---unwind 12 --no-standard-checks --bitwuzla
+--unwind 9 --no-standard-checks --bitwuzla
 ^VERIFICATION SUCCESSFUL$
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -248,6 +248,7 @@ collect_ubuntu_packages() {
     libboost-iostreams-dev
     libboost-system-dev
     libboost-filesystem-dev
+    libmpfr-dev
     ninja-build
     python3-setuptools
     libncurses-dev
@@ -285,6 +286,7 @@ collect_macos_formulae() {
     cmake
     z3
     gmp
+    mpfr
     csmith
     boost
     ninja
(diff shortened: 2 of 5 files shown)


pmatos:
I haven't looked deeply into this PR but I really think there's a simpler way to do this. It worries me the amount of cmake code being added for something at, at first glance, is a simple problem that happens elsewhere.

intrigus-lgtm:
> I haven't looked deeply into this PR but I really think there's a simpler way to do this. It worries me the amount of cmake code being added for something at, at first glance, is a simple problem that happens elsewhere.

Very happy to implement a simpler solution if you have one.

intrigus-lgtm:
I switched away from converting to local symbols, but now simply rename cadical's symbols in bitwuzla.

lucasccordeiro (approved):

PR #4001, opened 2026-04-07
Evaluate print() arguments in GOTO to preserve checks

sergillam:
- Materialize each print argument as code in handle_print so side effects and overflow checks are emitted.
- Reclassify print1-fail from KNOWNBUG to CORE
- Add print overflow/safe regression variants.


lucasccordeiro (approved):

Python treats the integer variables as unbounded; so, there is no overflow. However, since we're modeling integer variables using bitvectors, overflow can occur in our encoding.

This PR is fine, since I've seen Python code where developers annotate each integer variable with its bit width. So, this PR is useful here.

However, at some point, we need to treat integer variables as unbounded to adhere to Python semantics.

PR #4005, opened 2026-04-08
[symex] add interval-based guard pruning (--interval-symex-guard)

lucasccordeiro:
This PR maintains a local interval domain during symbolic execution to track variable ranges via `ASSUME`, `ASSIGN`, and `DECL` instructions. At each GOTO, evaluate the guard against the accumulated intervals to detect when it becomes provably true or false, allowing early loop termination without relying solely on the unwind bound.

For example, `while (idx < 5) { ++idx; }` with `__ESBMC_assume(0 <= idx)` now terminates after exactly 5 unrollings instead of hitting the unwind limit.

This PR also fixes `assume_rec` to handle typecast-wrapped conditions by unwrapping them, matching the existing pattern in `havoc_rec`.


src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -1493,6 +1493,10 @@ void interval_domaint::assume_rec(const expr2tc &cond, bool negation)
       apply_assume_symbol_truth<interval_domaint::real_intervalt>(
         to_symbol2t(cond), negation);
   }
+  else if (is_typecast2t(cond) && is_bool_type(cond))
+  {
+    assume_rec(to_typecast2t(cond).from, negation);
+  }
```
rafaelsamenezes, line 1499: Nice. This might impact SV-COMP.

src/goto-programs/abstract-interpretation/interval_domain.cpp
```diff
@@ -1583,6 +1588,26 @@ void interval_domaint::set_options(const optionst &options)
...
+    havoc_rec(instruction.code);
+    break;
+  case ASSIGN:
+    assign(instruction.code);
+    break;
+  case ASSUME:
+    assume(instruction.guard);
+    break;
+  default:
+    break;
```
rafaelsamenezes, line 1612: I suggest adding a warning/debug msg here with the unsupported instruction. Conditional jumps, and throws might generate weird cases.

rafaelsamenezes (changes requested):

src/goto-programs/abstract-interpretation/interval_domain.cpp
lucasccordeiro: Good catch! Implemented as suggested.

PR #4008, opened 2026-04-08
Disabled slow tests

rafaelsamenezes:
These tests take around 30% of CI time. Disabling for now but we should optimize them somehow.


lucasccordeiro (approved):

brcfarias:
Could we keep the humaneval folder as it is and just exclude it from CI? Since it’s a benchmark, it might be better to keep everything in a single folder and avoid adding overhead to CI.

rafaelsamenezes:
> Could we keep the humaneval folder as it is and just exclude it from CI? Since it’s a benchmark, it might be better to keep everything in a single folder and avoid adding overhead to CI.

Maybe we can just mark the full suite as THOROUGH. This way it won't affect coverage test-suite.

brcfarias:
> > Could we keep the humaneval folder as it is and just exclude it from CI? Since it’s a benchmark, it might be better to keep everything in a single folder and avoid adding overhead to CI.
> 
> Maybe we can just mark the full suite as THOROUGH. This way it won't affect coverage test-suite.

Perhaps we could mark only the ones from this PR as THOROUGH.

intrigus-lgtm:
https://github.com/esbmc/esbmc/pull/4039 might be an alternative to moving the tests around, but labeling them as slow instead.

src/goto-symex/symex_goto.cpp
```diff
@@ -53,8 +53,8 @@ void goto_symext::symex_goto(const expr2tc &old_guard)
   // Interval-based guard check (--interval-symex-guard)
   if (!new_guard_false && !new_guard_true && interval_domain_state)
   {
-    tvt res =
-      interval_domaint::eval_boolean_expression(old_guard, *interval_domain_state);
+    tvt res = interval_domaint::eval_boolean_expression(
```
lucasccordeiro, line 56: Formatting-only change unrelated to disabling tests. Drop it from this PR.

lucasccordeiro (changes requested):
Moving the whole `function_contract` suite into `regression/disabled/` drops 140 tests from CI with no record of why or when they return; mark the slow ones THOROUGH instead, which the harness already skips on Windows and macOS.

PR #4012, opened 2026-04-08
[witness] improve location invariant insertion

XLiZHI:
- A forward jump from above → points to assertion (check invariant)
- A backward goto → still points to the original loop head (loop structure remains unchanged)


lucasccordeiro (approved):

PR #4025, opened 2026-04-10
remove the extern definition

LukeW1999:
This pull request makes a minor change to the declaration of `__ESBMC_return_value` in `clang_c_language.cpp` to improve linkage and visibility.

* Changed the declaration of `__ESBMC_return_value` from `extern int` to `int`, making it a definition rather than just a declaration.


lucasccordeiro:
@LukeW1999: could you also convert the log_warning into log_debug?

LukeW1999:
> @LukeW1999: could you also convert the log_warning into log_debug?

Sure. I would do it now.

lucasccordeiro (approved):

PR #4031, opened 2026-04-11
[k-induction] diagnose unresolved properties on VERIFICATION UNKNOWN

LukeW1999:
Fixes #4022.

## Summary

When `--multi-property --k-induction` exhausts all k-steps without a definitive result, ESBMC now runs a targeted diagnostic pass to identify which specific properties remain unresolved, instead of just reporting `VERIFICATION UNKNOWN`. Each property is individually reported as **PASSED** or **UNKNOWN**, so users can identify exactly which assertion or invariant is blocking the proof.

---

### New diagnostic function
- Added `diagnose_unknown_properties()` in `esbmc_parseoptions.cpp` and its declaration in `esbmc_parseoptions.h`
- Runs a per-VCC inductive-step check at the final k-step when k-induction is inconclusive
- Triggered only at the `VERIFICATION UNKNOWN` exit point — the main k-induction loop is unchanged and SV-COMP performance is not affected

### Multi-property check improvements (`bmc.cpp` / `bmc.h`)
- Removed the `base-case`-only restriction in `multi_property_check()` so it can also run during the diagnostic inductive-step pass
- Distinguished between **failed** properties (real counterexample found, shown in red `✗`) and **unknown** properties (inductive step unprovable, shown in yellow `?`)
- Added `unknown_properties` counter to `SimpleSummary` and updated the summary output to display it

### Regression test
- Added `regression/loop-invariants/multi_property_unknown_diagnosis/` for #4022
- The test has intentionally incomplete loop invariants (relational invariant omitted), making the post-loop assertion unprovable in 2 k-steps
- Verifies that the output contains `VERIFICATION UNKNOWN` and `? UNKNOWN: 'post loop modular equality'`

---

## Testing

```
--no-standard-checks --loop-invariant --multi-property --max-k-step 2
```

Expected output:
```
VERIFICATION UNKNOWN
? UNKNOWN: 'post loop modular equality'
```

### DIFF
--- a/regression/loop-invariants/multi_property_unknown_diagnosis/main.c
+++ b/regression/loop-invariants/multi_property_unknown_diagnosis/main.c
@@ -1,0 +1,38 @@
+/* --multi-property --loop-invariant with an incomplete invariant that leaves
+ * the post-loop assertion unprovable by k-induction.
+ *
+ * The loop invariants cover the ranges of sum, bound2, and idx2, but omit
+ * the relational invariant (sum % bound2 == idx2 % bound2).  Without it the
+ * inductive step cannot discharge the post-loop assertion, so the result is
+ * VERIFICATION UNKNOWN.
+ *
+ * When UNKNOWN is reached, ESBMC should run a per-VCC diagnostic pass and
+ * identify the specific unresolved property by name.
+ *
+ * Regression test for GitHub issue #4022.
+ */
+
+int main()
+{
+  int bound2, idx2, sum;
+  idx2 = 0;
+  __ESBMC_assume(0 <= sum && sum < 256);
+  __ESBMC_assume(4 <= bound2 && bound2 < 256);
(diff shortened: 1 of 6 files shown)


LukeW1999:
This PR: [here](https://github.com/esbmc/esbmc/actions/runs/24281368363/job/70904455007
) (30s)

  correct:           19375
    correct true:    11979
    correct false:    7396
  incorrect:            25
    incorrect true:     10
    incorrect false:    15
  unknown:           17200
  Score:             30794 (max: 61270)

Master: [here](https://github.com/esbmc/esbmc/actions/runs/24246199843/job/70813434306) (30s)
  correct:           19387
    correct true:    11994
    correct false:    7393
  incorrect:            25
    incorrect true:      9
    incorrect false:    16
  unknown:           17188
  Score:             30837 (max: 61270)

lucasccordeiro (approved):

PR #4032, opened 2026-04-11
[solver] implement quantifier support in Bitwuzla backend

lucasccordeiro:
Replace the abort()-stub in bitwuzla_convt::mk_quantifier with a real implementation using the Bitwuzla C API: create fresh bound variables with bitwuzla_mk_var, substitute the SSA terms in the body with bitwuzla_substitute_term, then build the FORALL/EXISTS term directly with bitwuzla_mk_term (no VARIABLE_LIST wrapper, unlike CVC5).

Add four regression tests covering forall, exists, a counterexample case, and nested quantifiers.

### DIFF
--- a/regression/bitwuzla/quantifiers-exists/main.c
+++ b/regression/bitwuzla/quantifiers-exists/main.c
@@ -1,0 +1,8 @@
+int main()
+{
+  int value;
+  __ESBMC_assert(__ESBMC_exists(&value, value == 42), "exists 42");
+  __ESBMC_assert(
+    !__ESBMC_exists(&value, value == 0 && value == 1), "contradiction");
+  return 0;
+}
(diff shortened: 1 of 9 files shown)


LukeW1999 (approved):
LGTM 

One thing worth noting though: Bitwuzla struggles with alternating ∀∃ quantifiers. Take this example:

```c
  int main()
    {
      int a, b;
      /* FORALL a. EXISTS b. b == a+1 — trivially true, b=a+1 witnesses it */
      __ESBMC_assert(__ESBMC_forall(&a, __ESBMC_exists(&b, b == a + 1)),
"for every a there exists b = a+1"); 
      return 0;
    }
```

Z3 handles this in <1ms; Bitwuzla times out. This is a known upstream limitation, not something introduced here, so I'd leave it out of scope for this PR. That said, it'd be worth opening a separate issue to document the behavior — users mixing `forall`/`exists` will hit this eventually and wonder why.

lucasccordeiro:
Many thanks, @LukeW1999.

I've just created this follow-up issue: https://github.com/esbmc/esbmc/issues/4033.

PR #4041, opened 2026-04-12
fix guard restore across catch frame pop

sergillam:
- Promote regression/python/import-os2_fail to CORE.
- Fix guard restore in symex pop_frame during exception handling.
- Keep catch path guard active by skipping entry_guard restore when stack_catch is not empty.


lucasccordeiro:
SV-COMP run (100s): 

```
  correct:           21949
    correct true:    14243
    correct false:    7706
  incorrect:            32
    incorrect true:     12
    incorrect false:    20
  unknown:           14619
  Score:             35488 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24302817855
```

PR #4048, opened 2026-04-13
[goto] extract atomicity-check into post-conversion pass

### LINKED ISSUE #123: atomicity check transformations need to be rewritten as a separate module
Similar to what was done with the k-induction module.

Atomicity check should be removed from the goto_convert code, and executed as a separate step after the goto functions are created.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/123.

This PR moves the atomicity instrumentation logic out of `goto_convertt` (where it ran during AST→GOTO conversion on old-style exprt) into a new standalone pass `goto_atomicity_check` that runs after `goto_convert` on fully-built `goto_functionst` with irep2 types.

The new pass uses a two-pass design: first, collect iterators for the instructions to instrument, then process them, so that newly inserted `snapshot/assert/atomic` instructions are never reprocessed. Removes ~340 lines of dead code from `goto_convert.cpp` and `goto_function.cpp`.

This PR also adds regression tests `00_atomicity09` (FAIL: non-atomic RMW on shared global) and `00_atomicity10` (PASS: threads operate only on locals).


src/goto-programs/goto_atomicity_check.cpp
```diff
@@ -0,0 +1,327 @@
+#include <goto-programs/goto_atomicity_check.h>
+#include <irep2/irep2_expr.h>
+#include <irep2/irep2_utils.h>
+#include <util/migrate.h>
+#include <util/symbol_generator.h>
+
+static const irep_idt esbmc_alloc_id("__ESBMC_alloc");
+static const irep_idt esbmc_alloc_size_id("__ESBMC_alloc_size");
```
rafaelsamenezes: This creates a weird dependency over irep_idt string interner which might lead to some weird initialization glitches.

src/goto-programs/goto_atomicity_check.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+class goto_atomicity_checkt
+{
+public:
+  goto_atomicity_checkt(const namespacet &_ns, contextt &_context)
+    : ns(_ns), context(_context), tmp_sym("goto_atomicity::")
+  {
+  }
+
+  void check(goto_programt &program);
```
rafaelsamenezes, line 18: Document public methods

rafaelsamenezes (changes requested):

src/goto-programs/goto_atomicity_check.cpp
lucasccordeiro: Fixed.

src/goto-programs/goto_atomicity_check.cpp
lucasccordeiro: Fixed.

rafaelsamenezes (approved):

unit/goto-programs/atomicity_check.test.cpp
```diff
@@ -0,0 +1,435 @@
...
+        if (g) break;
+      }
+      return 0;
+    }
+  )";
+
+  // Only the if(g) GOTO generates an assertion; the unconditional back-edge
+  // does NOT (it has a true guard). At least one assertion from the if(g) GOTO
+  // confirms the true-guard branch was correctly bypassed.
+  auto P = run_atomicity_check(code);
```
rafaelsamenezes: This test is relying on the simplifier and trivial guards. This shouldn't be tested here, you should assume that there will be no simplification at this point.

rafaelsamenezes (changes requested):

unit/goto-programs/atomicity_check.test.cpp
lucasccordeiro: Fixed.

unit/goto-programs/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 new_unit_test(loop-unroll-algorithms-test "loop_unroll.test.cpp" "test_goto_factory;gotoprograms;gotoalgorithms;filesystem;langapi")
+# new_unit_test(atomicity-check-test "atomicity_check.test.cpp" "test_goto_factory;gotoprograms;clang-cpp-frontend;c2goto;util_esbmc;filesystem;langapi")
```
intrigus-lgtm, line 2: Why is this not enabled?

lucasccordeiro: It was enabled in another PR: https://github.com/esbmc/esbmc/pull/4059.

PR #4051, opened 2026-04-13
[Python] support typing.NewType and simple type aliases

prototypeC14:
Fixes the type_alias `KNOWNBUG`. ESBMC did not recognize typing.NewType, causing Function "TaskId" not found when encountering code like:
```python
  TaskId = NewType('TaskId', int)
  task_id = TaskId(1)
```


lucasccordeiro (approved):

PR #4059, opened 2026-04-14
 Enable atomicity-check and interval-symex-guard goto-programs tests

sergillam:
Re-enable two unit tests in `unit/goto-programs/CMakeLists.txt`:
- `atomicity-check-test`
- `interval-symex-guard-test`
- Uncommented both test targets.
- Updated `atomicity-check-test` link dependencies to valid CMake targets and a stable link order.


rafaelsamenezes:
The main issue is the dynamic vs static builds. Some symbols can be indirectly linked because of this. I've added some changes into linking visibility in this PR.

lucasccordeiro (approved):

PR #4060, opened 2026-04-14
Fix multi-property missing assertion results in inductive step 

LukeW1999:
Fix #4052 

### Problem

When using `--multi-property` together with `--loop-invariant`, post-loop assertions were silently dropped during the inductive step. Only the base case triggered `multi_property_check`, so each assertion was never individually solved or printed in the inductive step, causing results to be missing from the output entirely.

### Root Cause

`run_thread` only called `multi_property_check` for the base case (`--base-case`) or `--diagnose-unknown-properties`, but not for `--inductive-step`. As a result, when k-induction reached the inductive step, per-property solving was bypassed.

Additionally, the `verified_claims` set and its associated mutex were used to skip re-checking already-verified claims across steps. This logic incorrectly suppressed post-loop assertions in the inductive step before they had a chance to be reported.


src/esbmc/bmc.cpp
```diff
@@ -1613,13 +1608,6 @@ smt_convt::resultt bmct::multi_property_check(
       reached_mul_claims.emplace(claim_sig);
     }
 
-    if (verified_claims.count(claim.claim_cstr))
```
ChenfengWei0, line 1616: Why do we need to remove this completely rather than disable it when enabling `loop-invariant`?

lucasccordeiro (approved):

ChenfengWei0:
@LukeW1999 If we don't explicitly clear up the `goto` and SSA for the verified claims, will they be repeatedly reported in the log?

ChenfengWei0:
> @LukeW1999 If we don't explicitly clear up the `goto` and SSA for the verified claims, will they be repeatedly reported in the log?

I just checked. It won't affect anything

LukeW1999:
Hi, @ChenfengWei0 ,

I am sorry. I missed the message. I just come to say sorry. I am glad that it do not affect anything.

PR #4061, opened 2026-04-14
[smt] fix crash in tuple-sym-flattener with non-constant array size

### LINKED ISSUE #2639: [C] Crash when combining symex-guard + tuple-sym-flattener
For the following program calling ESBMC with `--tuple-sym-flattener --smt-symex-guard` results in a crash.
```c
    if(argc > 1)
        ;
}
```
The crash disappears when dropping any of the options.
Version: Built from 2b7d8c382b817fc01b22c736b7c4459a1985b608
Full output:
```
Enabling --smt-during-symex to use features that involve encoding SMT during symex
Parsing test.c
Converting
Generating GOTO Program
GOTO program creation time: 0.152s
GOTO program processing time: 0.000s
ERROR: Non-constant sized array of type constant_array_of2t
Aborted (core dumped)
```

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2639.

`smt_tuple_sym_flattener::tuple_array_create` aborted when the array size was non-constant (e.g., `argv` typed as `char*[argc+1]`), which triggered a crash with `--tuple-sym-flattener --smt-symex-guard`. 

This PR returns the unconstrained fresh symbol instead, matching the existing handling for infinite-sized arrays. It also applies the same defensive fix to `smt_tuple_node_flattener` for the `const_array=false` path.

### DIFF
--- a/regression/esbmc/github_2639/main.c
+++ b/regression/esbmc/github_2639/main.c
@@ -1,0 +1,5 @@
+int main(int argc, char *argv[])
+{
+  if(argc > 1)
+    ;
+}
--- a/regression/esbmc/github_2639/test.desc
+++ b/regression/esbmc/github_2639/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--tuple-sym-flattener --smt-symex-guard --z3
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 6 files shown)


lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/4067.

PR #4066, opened 2026-04-14
[witness] enable assumption in YAML based witness waypoint

XLiZHI:


XLiZHI:
no-overflow-juliet
This PR:
```

==> cpachecker-violation-v2 <==

Statistics:           6232 Files
  correct:            2379
    correct true:        0
    correct false:    2379
  incorrect:           664
    incorrect true:    664
    incorrect false:     0
  unknown:            3189
  Score:            -18869 (max: 9310)
  https://github.com/esbmc/esbmc/actions/runs/24413002411/job/71317638840
```

Master:
```
==> cpachecker-violation-v2 <==

Statistics:           6232 Files
  correct:            2379
    correct true:        0
    correct false:    2379
  incorrect:           664
    incorrect true:    664
    incorrect false:     0
  unknown:            3189
  Score:            -18869 (max: 9310)
  https://github.com/esbmc/esbmc/actions/runs/24246199843/job/70813434306
```

src/goto-symex/witnesses.cpp
```diff
@@ -1110,6 +1110,40 @@ void generate_testcase_metadata()
...
+  // (arithmetic, bitwise, casts, etc.) are handled without enumerating expr_ids.
+  // Unlike get_nondet_symbol, non-nondet symbols do not cause early return.
+  if (!expr)
+    return false;
+
+  if (is_symbol2t(expr))
+    return is_nondet_symbol_name(to_symbol2t(expr).thename.as_string());
+
+  unsigned int n = expr->get_num_sub_exprs();
+  for (unsigned int i = 0; i < n; ++i)
```
lucasccordeiro: Shall we use our `BigInt` here to avoid overflow?

If that's not needed, we could use `size_t` instead of `unsigned int` to keep consistency with the existing code.

lucasccordeiro (changes requested):

src/goto-symex/witnesses.cpp
XLiZHI: https://github.com/esbmc/esbmc/blob/e77b64ae31481171d3fe4f8cec73d1479eb6f49a/src/irep2/irep2.h#L759

https://github.com/esbmc/esbmc/blob/e77b64ae31481171d3fe4f8cec73d1479eb6f49a/src/solvers/smt/tuple/smt_tuple_node.cpp#L21

https://github.com/esbmc/esbmc/blob/e77b64ae31481171d3fe4f8cec73d1479eb6f49a/src/solvers/smt/tuple/smt_tuple_sym.cpp#L21

Thank you for your review, Prof @lucasccordeiro 

I checked the usage of this function. They all use unint as the type. 

Theoretically, it may have risks, but the number of direct sub expressions of any expression is always a fixed few (determined by the IRep2 field). They cannot exceed UNINT_MAX.

Of course, we might also change all of them to size_t. What do you think?

lucasccordeiro: Thanks for checking this overflow issue, @XLiZHI.

Yes, please: change `unsigned int` with `size_t`.

PR #4068, opened 2026-04-14
support str(obj).__str__ dispatch and f-string member interpolation

sergillam:
Fixes the `strings2` KNOWNBUG by adding support for `str(obj)` dispatching to `__str__` on custom classes and resolving f-string interpolation of struct member expressions typed as pointers, `str(obj)` on a class with `__str__` defined was not dispatching to the dunder method. Additionally, `convert_to_string` in f-string handling fell through to a `"<expr>"` placeholder for member expressions typed as `any_type` (void*), producing `"ExampleClass(name=<expr>)"` instead of the actual value.

- Register `{"str", "__str__"}` in the unary dunder dispatch map
- Try `__str__` dispatch in `str()` call handling before falling back to
  `handle_str`
- Handle pointer types in `convert_to_string` so that void*/char* member
  expressions pass through to `concatenate_strings_via_c_function`
- Promote `strings2` from KNOWNBUG to CORE
- Add 3 variant regression tests (plain return, multi-field f-string,
  f-string with member access outside `__str__`)

### DIFF
--- a/regression/ir-ra/ra-interval-lift-mul-rup-both-tracked/test.desc
+++ b/regression/ir-ra/ra-interval-lift-mul-rup-both-tracked/test.desc
@@ -6,5 +6,5 @@ main.c
 \(\* \|smt_conv::ra_lo_up::0\| \|smt_conv::ra_lo_up::0\|
 \(ite
 22204460492503131
-^VERIFICATION FAILED$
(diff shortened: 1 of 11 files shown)


lucasccordeiro (approved):

PR #4072, opened 2026-04-15
Support GCC nested functions

### LINKED ISSUE #4070: Support GCC nested functions
See https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

Nested functions is a GCC extension that allows functions to be declared within a function. The nested functions can access and modify variables in the enclosing scope. Functions can be nested multiple times. Nested functions can jump to a label in the enclosing scope. Nested functions are not closures however, and cannot be returned or used outside of once their parent function has returned. Nested functions can be passed as functions pointers such as in:
```C
void c(void (*f)(void)) {
    f();
}

void a(void) {
    int x = 0;
    void b(void) {
        x = 1;
    }

    c(b);
}
```

(`x` should be 1 at the end of `a`.)

See also https://github.com/diffblue/cbmc/pull/8935

### LINKED ISSUE #4081: [Testsuite Audit]: GCC nested functions extension not supported
## Description

ESBMC fails to parse **41 C programs** that use GCC's nested function extension (function definitions inside other functions). The error is: `function definition is not allowed here`.

Nested functions are a widely-used GCC extension that allows defining helper functions with access to the enclosing scope's variables.

## Reproducers

- `gcc/testsuite/gcc.c-torture/execute/20000822-1.c`
- `gcc/testsuite/gcc.c-torture/execute/20010605-1.c`
- `gcc/testsuite/gcc.c-torture/execute/20010209-1.c`

```c
// Simplified example
int main() {
    int x = 42;
    int helper(int y) { return x + y; } // GCC nested function
    return helper(1) == 43 ? 0 : 1;
}
```

```bash
esbmc test.c --unwind 5 --z3
# Error: function definition is not allowed here
```

## Difficulty

**Medium.** Nested functions capture variables from the enclosing scope, which requires closure-like semantics. A thorough investigation is required.

## Likely Fix Path

A thorough investigation is required, but the likely approach is:

1. Enable the GCC nested functions extension in the Clang frontend (Clang supports `-fnested-functions` partially)
2. Alternatively, implement a source-level or IR-level transformation that lifts nested functions to top-level, passing captured variables as extra parameters (lambda lifting)
3. The transformation in `src/clang-c-frontend/` or `src/goto-programs/` would convert each nested function to a regular function with an explicit environment struct parameter

pmatos:
Fixes #4070
Fixes #4081

## Summary

- Adds a source-to-source transformation pass in the Clang C frontend that lifts GCC nested function definitions to file scope before Clang parses the source
- Captured variables from enclosing scopes become explicit pointer parameters (direct calls) or file-scope static pointers (function pointer use)
- Multi-level nesting is handled iteratively, each pass lifting the outermost level
- No changes to GOTO programs, symbolic execution, or solvers — the transformation is entirely in the frontend

## Test plan

- [x] 6 new regression tests covering basic nesting, capture read/write, nested functions with parameters, function pointer passing, multi-level nesting, and a failing assertion case — all pass
- [x] Existing `esbmc/00_*` regression tests (40 tests) — 100% pass, no regressions
- [x] Normal C files without nested functions — fast no-op path verified


pmatos:
Also fixes #4081

bbb23exposed:
Thank you so much for this!!

lucasccordeiro (approved):

PR #4073, opened 2026-04-15
[Python] preserve container type on subscript assignment

prototypeC14:
This pr:
- fix container variable being retyped after `dp[i] = v`, which broke subsequent reads.
- reversed1 from KNOWNBUG to CORE.


lucasccordeiro (approved):

PR #4075, opened 2026-04-15
[Function Contract] Fix --enforce-contract '*' --function skipping pointer-param allocation

LukeW1999:
## Summary

Fixes #4046: `--enforce-contract '*' --function <name>` was producing spurious `dereference failure: Incorrect alignment` when the target function has pointer parameters.

## Root Cause

Three separate bugs combined to cause the failure:

1. **`is_compiler_generated` false positive on C++ functions** 
   The old check used `func_name.find('#') != std::string::npos` to skip compiler-generated symbols. However, C++ USR-style IDs (e.g. `c:@F@fst#*1I#`) legitimately contain `#` as parameter encodings, causing all parameterised C++ free functions to be silently skipped.

2. **`find_function_symbol` short-name fallback missing** 
   The wildcard expansion inserts full IDs (e.g. `c:@F@fst#*1I#`) into `to_enforce`, while `--function` provides only the short name (`fst`). The lookup only tried exact match and the no-parameter C++ suffix `#`, so parametered C++ functions were never found.

3. **`alloc_ptr_params` comparison used full ID vs. short name** 
   In `enforce_contracts`, `function_name == entry_function` compared the full USR ID against the short name supplied by `--function`, which always evaluated to `false` — skipping pointer-parameter backing-storage allocation entirely.

## Fix

- **`is_compiler_generated`**: Extract the short name from a Clang USR ID before applying heuristics (strip from first `#`, then take the segment after the last `@`). Check for `~` (destructor) and `__cxa_*` on the short name only.
- **`find_function_symbol`**: Add a fallback that iterates `goto_functions` and matches on `candidate->name` (always the short name) when all prefix-based lookups fail.
- **`enforce_contracts`**: Change `function_name == entry_function` to `id2string(func_sym->name) == entry_function` so the comparison always uses the short name.

## Tests Added

| Test | Language | Scenario | Expected |
|---|---|---|---|
| `wildcard_enforce_ptr_pass` | C | single pointer param | VERIFICATION SUCCESSFUL |
| `wildcard_enforce_ptr_fail` | C | wrong ensures clause | VERIFICATION FAILED |
| `wildcard_enforce_two_ptrs_pass` | C | two pointer params | VERIFICATION SUCCESSFUL |
| `wildcard_enforce_two_ptrs_fail` | C | wrong ensures clause | VERIFICATION FAILED |
| `cpp_wildcard_enforce_ptr_pass` | C++ | pointer param, USR ID | VERIFICATION SUCCESSFUL |
| `cpp_wildcard_enforce_ptr_fail` | C++ | wrong ensures clause | VERIFICATION FAILED |
| `cpp_wildcard_enforce_struct_ptr_pass` | C++ | pointer-to-struct param | VERIFICATION SUCCESSFUL |
| `cpp_wildcard_enforce_struct_ptr_fail` | C++ | wrong ensures clause | VERIFICATION FAILED |

## Closes

Closes #4046


LukeW1999:
This Codacy early warning is wrong. This is useful for the code with loop.

PR #4095, opened 2026-04-15
[util] tolerate nil sub-operands in base_type(expr2tc&)

ChenfengWei0:
sideeffect2t of kind `nondet` is the canonical IR shape produced by gen_nondet() (src/irep2/irep2_utils.h): its `operand` and `size` sub-expressions are intentionally nil. Any expression walker must therefore tolerate nil children, the same way check_rec in goto_check and base_type_eqt::base_type_eq_rec already do.

base_type(expr2tc&) unconditionally dereferenced its operands inside the recursive Foreach_operand walk, so any caller reaching it with a subtree containing a nondet sideeffect crashed with a SIGSEGV. In practice this fired from goto_check's div_by_zero_check, whose notequal(divisor, 0) subtree carried the nondet divisor.

Add an early `is_nil_expr(expr)` guard to align with the rest of the IR-walking code, and a unit test that constructs a gen_nondet() and feeds it (both directly and nested inside a notequal) to base_type. The same test also covers a nil top-level expression.

Before the fix, base_type(expr2tc&) dereferenced these nil children in its recursive Foreach_operand walk, so a div-by-zero check whose divisor contained a nondet sideeffect crashed the frontend during goto_check.

Assisted-by: Claude-Opus4.6


XLiZHI (approved):
LGTM

PR #4097, opened 2026-04-15
Fix list string handling and promote list34 coverage

sergillam:
- Promotes list34 to CORE and adds list34_1 variant tests.
- Refines str call handling and fixes list element string typing in the Python frontend.

### DIFF
--- a/src/python-frontend/python_list.cpp
+++ b/src/python-frontend/python_list.cpp
@@ -2900,7 +2900,11 @@ exprt python_list::list_repetition(
   // True when the list operand is a variable (not a literal).
   bool is_variable_list = false;
 
-  auto parse_size_from_symbol = [&](symbolt *size_var, BigInt &out) -> bool {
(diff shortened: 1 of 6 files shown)


src/python-frontend/python_list.cpp
```diff
@@ -2720,15 +2912,133 @@ exprt python_list::list_repetition(
     return true;
   };
 
-  // Get first element's symbol;
-  // return nil when elements are literals (no backing symbol)
-  // so the caller falls back to a runtime list copy.
+  auto parse_repeat_expr = [&](const exprt &expr, BigInt &out) -> bool {
+    if (expr.is_constant())
+    {
```
brcfarias, line 2922: You could add an assert to ensure that expr is an integer.

brcfarias (approved):
LGTM

src/python-frontend/python_list.cpp
sergillam: done

PR #4099, opened 2026-04-16
 promote python/jpl to CORE with --no-unwinding-assertions

sergillam:
- The python/jpl test was marked as KNOWNBUG due to an unwinding assertion failure on the while true reactive loop.
- The frontend correctly handles all constructs in this test (classes, inheritance, lambda, random.randint as nondet). The failure was caused by missing --no-unwinding-assertions flag — standard practice for bounded verification of intentionally infinite loops.


lucasccordeiro (approved):
Thanks dor submitting this PR, @sergillam.

PR #4098, opened 2026-04-16
Revert "Support GCC nested functions"

intrigus-lgtm:
This is absolutely ridiculous.
We re-implemented a full lexer and parser in 1500 lines of code that clearly no-one has properly reviewed in depth.
(HINT: clang already has a lexer that we should have been able to re-use)

This new source-transformation is now applied to all processed source files and is supposed to not change anything when the nested gcc function is not supported, but as this is 99% vibe-coded, who guarantees this???
Do we not care about soundness anymore?
Should we start including statements about our trusted-computing base that include claude/codex/gemini/whatever?

I don't need support for features for the sake of being able to claim that we support something if it isn't **sound** and properly **reviewed**.
With the current architecture and clang, it's impossible to support gcc's nested function extension as clang WILL NOT implement this extension itself.

**If you really cannot live without this extension, then hide it behind a flag so I don't have to care about it and at least use the clang lexer for it.** 

And with minimal effort there are already two edge-cases that this code doesn't support:
```c
#include <assert.h>

void one_fun() {
    int x = 0;
    void local(void) { x++; }
    local();
    local();

    if (x <= 2) local();
    assert(x == 3);
}

void two_fun__() {
    int x = 0;
    void __local(void) { x++; }
    __local();
    __local();

    if (x <= 2) __local();
    assert(x == 3);
}

void two_fun____() {
    int x = 0;
    void local(void) { x++; }
    local();
    local();

    if (x <= 2) local();
    assert(x == 3);
}

int main() {
    one_fun();
    two_fun__();
}
```

```
/home/sg/wtf3.c:27:13: error: redefinition of '__esbmc_nested_two_fun______local'
   27 | static void __esbmc_nested_two_fun______local(int *__capture_x) { (*__capture_x)++; }
      |             ^
/home/sg/wtf3.c:15:13: note: previous definition is here
   15 | static void __esbmc_nested_two_fun______local(int *__capture_x) { (*__capture_x)++; }
      |             ^
[ERROR] PARSING ERROR
```

```c
#include <assert.h>

int main() {
    __label__ lol;
    int x = 0;
    void local(void) {
        goto lol;
    }
    local();
    local();
lol:
    x++;

    if (x <= 2) local();
    assert(x == 3);
}
```

```
/home/sg/wtf.c:3:40: error: expected parameter declarator
    3 | static void __esbmc_nested_main__local(__label__ *__capture_lol) {
      |                                        ^
/home/sg/wtf.c:3:40: error: expected ')'
[...]

PR #4100, opened 2026-04-16
Fix nested function transform: gate behind flag, fix name collision and __label__

pmatos:
## Summary

Addresses issues raised in #4098:

- Gate the nested function source-to-source transformation behind a new `--gcc-nested-functions` CLI flag. The transform no longer runs unconditionally on every C file.
- Fix name collision where underscore-heavy identifiers (`two_fun__` + `__local` vs `two_fun____` + `local`) produced identical lifted names. Uses length-prefixed encoding to make the mapping injective.
- Fix `__label__` declarations being misclassified as typedef-based variable declarations and incorrectly captured by nested functions.
- Add regression tests for both #4098 reproducers.

## Test plan

- [x] Without `--gcc-nested-functions`: nested function files fail at parse (transform not applied)
- [x] With `--gcc-nested-functions`: all 6 existing regression tests pass
- [x] Name collision reproducer from #4098: VERIFICATION SUCCESSFUL
- [x] `__label__` reproducer: no longer misclassifies labels as variables (KNOWNBUG: `__label__` not yet supported at GOTO level)
- [x] Build clean with `-Werror`


pmatos:
Do not merge yet pls.

Yiannis128:
> Do not merge yet pls.

I'll wait until it's ready for review before doing the review as the code may change.

pmatos:
> > Do not merge yet pls.
> 
> I'll wait until it's ready for review before doing the review as the code may change.

ready - I split it into 4 commits for ease of review.

pmatos:
If we merge this, we can close #4098 .

lucasccordeiro (approved):

Yiannis128:
Hi @pmatos , it's been already merged but I wanted to ask if this supports double nested functions? For example global function A() contains nested function b() and b() contains c(). Does this work? Otherwise looks good.

PR #4102, opened 2026-04-16
[witness] add special treatment for branching waypoints

XLiZHI:
This PR has fixed the following branching waypoints:

```c
  do
  {
    if (some_var)
      break;
  } while (0);
```

GOTO:

```
  if (!guard) goto 1;
    goto 2;
 1: 
```
It will be simplified to:
```
  if (guard) goto 2;
```

Therefore, we need to handle it in the opposite way here.

### DIFF
--- a/src/goto-programs/goto_convert.cpp
+++ b/src/goto-programs/goto_convert.cpp
@@ -108,6 +108,7 @@ void goto_convertt::optimize_guarded_gotos(goto_programt &dest)
     {
       it->set_target(it_goto_y->get_target());
       make_not(it->guard);
+      it->flipped_guard = true;
       it_goto_y->make_skip();
     }
   }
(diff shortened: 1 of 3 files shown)


XLiZHI:
no-overflow_juliet

This PR:
```
==> cpachecker-violation-v2 <==

Statistics:           6232 Files
  correct:            2379
    correct true:        0
    correct false:    2379
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            3853
  Score:              2379 (max: 9310)
  https://github.com/esbmc/esbmc/actions/runs/24477376700/job/71535776443
```

Master:
```
==> cpachecker-violation-v2 <==

Statistics:           6232 Files
  correct:            2377
    correct true:        0
    correct false:    2377
  incorrect:           664
    incorrect true:    664
    incorrect false:     0
  unknown:            3191
  Score:            -18871 (max: 9310)
  https://github.com/esbmc/esbmc/actions/runs/24459940268/job/71491878913
```

664x are currently not supported by cpachecker.

lucasccordeiro (approved):

PR #4110, opened 2026-04-17
add: script to extract sv-comp scores every year

Yiannis128:
Two helper scripts under `scripts/sv-comp-graph/` for visualising ESBMC's SV-COMP performance over time.

- `sv_comp_graph.py` — scrapes past SV-COMP result pages on sv-comp.sosy-lab.org and produces one bar chart per ESBMC variant (esbmc, esbmc-kind, esbmc-incr, …) plus a best-per-year chart. HTML scraping because SV-COMP doesn't publish a structured API.
- `sv_comp_history.py` — walks the git history of `scripts/competitions/svcomp/stats-{30s,100s,300s,600s}`, picks the commit nearest mid-month for each month, and plots monthly percentage scores as line graphs (one per budget + a combined overlay). Log y-axis by default with `--scale linear` to switch.

Both share a Pipfile. Run with `pipenv install && pipenv run python <script>.py`.

## Example output

sv-comp-best

sv-comp-esbmc-kind

sv-comp-hist-combined


Yiannis128:
I'm pretty sure I have nothing to do with that failing regression test...

lucasccordeiro (approved):

Nice graphs!

PR #4113, opened 2026-04-18
[pointer-analysis] fix zero-sized type-pun through struct member (#723)

### LINKED ISSUE #723: Error constructing incompatible structure from one with array member
While #716 and #719 are about empty or zero-width unions and are addressed by some PRs, these non-standard types supported as compiler extensions continue to make more problems, in particular related to dereferencing:
```c
typedef union {
	struct {} x;
} t1;

typedef struct {
	int y[2];
} t2;

t1 v1;
t2 v2;

int main()
{
	v1 = *(t1 *)&v2;
}
```
leads to
```
### ERROR: 
Unexpectedly array type'd argument to construct_struct_ref
```
in one of the deferencet::construct_from_* methods.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/723.

`build_reference_rec` now short-circuits when the destination type has size 0. 

Before this PR, recursing through struct members/array elements reached a scalar base case that failed with a spurious "Union pointer pointed at scalar" error. 

Reading 0 bytes is a no-op, so we now emit a trivial value of the target type. 

This PR also reclassifies github_723-empty from KNOWNBUG to CORE and adds a fail companion to guard against the silent suppression of unrelated assertions.


lucasccordeiro:
SV-COMP (30s): 

```
  correct:           19099
    correct true:    11883
    correct false:    7216
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:           17482
  Score:             30614 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24596739941
```

Master (30s): 

```
  correct:           19179
    correct true:    11947
    correct false:    7232
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:           17402
  Score:             30758 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24580085652
```

PR #4114, opened 2026-04-18
[symex] fix malloc with negative size causing vacuous verification

### LINKED ISSUE #1631: Memory space assumptions are unsatisfiable
As a variation of <https://github.com/esbmc/esbmc/issues/775#issuecomment-1172190902> and #1091, the following program is given the wrong result (success) by ESBMC with `--force-malloc-success --compact-trace`:
```c
#include <stdlib.h>
#include <assert.h>

int main()
{
	void *a = (void *)malloc(4);
	void *b = (void *)malloc(-4);
	assert(0);
}
```
The reason is that for each object whose address has been taken (such as the dynamic ones created here), an assumption is encoded by the smt_memspace, that those objects do not overlap. `-4` is cast to size_t, so the malloc is called with 2^n-4 where n is the width of addresses. Thus, the memspace assumptions are already UNSAT and the result is a successful verification.

This also gets the wrong result when using a nondet size != 0.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/1631.

### DIFF
--- a/regression/esbmc/github_1631/main.c
+++ b/regression/esbmc/github_1631/main.c
@@ -1,0 +1,9 @@
+#include <stdlib.h>
+#include <assert.h>
+
+int main()
+{
+  void *a = (void *)malloc(4);
+  void *b = (void *)malloc(-4);
+  assert(0);
+}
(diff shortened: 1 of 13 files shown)


lucasccordeiro:
SV-COMP (30s): https://github.com/esbmc/esbmc/actions/runs/24608990421

PR #4124, opened 2026-04-20
[Python] support dict.fromkeys() class method

prototypeC14:
This pr add support for `dict.fromkeys` method.
- Add `handle_dict_fromkeys` in the Python frontend so calls like
  `dict.fromkeys([1, 2, 3])` and `dict.fromkeys([1, 2], "v")` build a new
  dict with the given keys sharing one default value.
- Keys are deduped to match Python semantics.
-  Limitation: iterable must be a list literal for now. tuple, set,
  range, generators throw (left for follow-ups).


lucasccordeiro (approved):

PR #4128, opened 2026-04-20
Fix Zulip link in README.md

Yiannis128:
Updated Zulip link in README to a sign up link instead of read only link.


lucasccordeiro (approved):

PR #4130, opened 2026-04-21
[Python] support inline literals and smarter return-type inference

prototypeC14:
This pr
-  Fix dict method calls crashed like `{"a":1}.get("x", 3)` .
- Fix `dict.{setdefault, get, pop}` returning the wrong type when the dict was untyped. Prevent type-check failures like `a.setdefault(k, []).append(x)`

### DIFF
--- a/src/python-frontend/python_annotation.h
+++ b/src/python-frontend/python_annotation.h
@@ -304,6 +304,27 @@ public:
     }
   }
 
+public:
+  // Infer the return type of dict.{setdefault, get, pop} from args[1]'s AST shape:
+  // "list"/"dict"/"set"/"tuple" for literal defaults,
+  // "Any" for anything else,
+  // "" when there is no default arg.
+  static std::string infer_type_from_default_arg_shape(const Json &args_node)
+  {
+    if (!args_node.is_array() || args_node.size() < 2)
+      return std::string();
+    const std::string def_type = args_node[1].value("_type", std::string());
+    if (def_type == "List")
+      return "list";
+    if (def_type == "Dict")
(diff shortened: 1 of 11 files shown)


src/python-frontend/python_annotation.h
```diff
@@ -304,6 +304,43 @@ class python_annotation
     }
   }
 
+  // Return the builtin type of a dict.{get, setdefault, pop} default arg.
+  // Recognises literal shapes (list/dict/set/tuple) and scalar Constants
+  // (int/float/str/bool/None). Returns "Any" for non-literals (variables,
+  // calls), "" when the call has no default argument.
+  static std::string infer_type_from_default_arg_shape(const Json &args_node)
```
lucasccordeiro, line 309: @prototypeC14: Could you please check whether we already have a similar "helper" for this type of inference? If that exists, we have an opportunity to refactor some of the code to reuse existing methods.

lucasccordeiro (changes requested):
@prototypeC14: I have just a small comment for this PR.

src/python-frontend/python_annotation.h
prototypeC14: Yes, we already have a similar "helper". We just need to expand it a bit.

lucasccordeiro: Super!

This is an opportunity to improve the existing methods/codebase ;-)

Later, we should also check our coverage report: https://app.codecov.io/gh/esbmc/esbmc.

There are some modules in our Python frontend that we must improve the coverage of the code.

PR #4131, opened 2026-04-21
[solver] make Bitwuzla the default SMT solver

lucasccordeiro:
This PR replaces Boolector with Bitwuzla as the preferred default solver in `pick_default_solver()`. When both are compiled in, Bitwuzla is selected first; Boolector remains the fallback if only it is present. The `all_solvers[]` fallback order and CLI help text are updated to match.

This PR also updates website docs and solvers README to reflect the new default.

### DIFF
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -429,7 +429,7 @@ const struct group_opt_templ all_cmd_options[] = {
      "Check all interleavings, even if a bug was already found"}}},
   {"Solver",
    {{"list-solvers", NULL, "List available solvers and exit"},
-    {"boolector", NULL, "Use Boolector (default)"},
+    {"boolector", NULL, "Use Boolector"},
     {"z3", NULL, "Use Z3"},
     {"z3-debug", NULL, "Extracts Z3 dump and SMT2 formula"},
     {"z3-debug-dump-file",
@@ -443,7 +443,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"cvc4", NULL, "Use CVC4"},
     {"cvc5", NULL, "Use CVC5"},
     {"yices", NULL, "Use Yices"},
-    {"bitwuzla", NULL, "Use Bitwuzla"},
+    {"bitwuzla", NULL, "Use Bitwuzla (default)"},
     {"bv", NULL, "Use solver with bit-vector arithmetic"},
     {"ir",
      NULL,
@@ -460,11 +460,7 @@ const struct group_opt_templ all_cmd_options[] = {
     {"smtlib", NULL, "Use SMT lib format"},
     {"default-solver",
      boost::program_options::value<std::string>()->value_name("<solver>"),
-     "Override default solver used if no concrete one is specified"
-#ifdef BOOLECTOR
-     " (Boolector)"
-#endif
-    },
+     "Override default solver used if no concrete one is specified"},
     {"non-supported-models-as-zero",
      NULL,
      "If ESBMC can't extract a type/expression from the solver, then the value "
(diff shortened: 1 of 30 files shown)


src/solvers/solve.cpp
```diff
@@ -48,24 +48,29 @@ static const std::unordered_map<std::string, solver_creator *> esbmc_solvers = {
...
-  "bitwuzla"};
+  "yices"};
 
 static std::string pick_default_solver()
 {
-#ifdef BOOLECTOR
+#ifdef BITWUZLA
+  log_status("No solver specified; defaulting to Bitwuzla");
+  return "bitwuzla";
+#elif defined(BOOLECTOR)
```
rafaelsamenezes: Don't think this is needed. Just put boolector in the top (after smtlib) in all_solvers.

src/esbmc/options.cpp
```diff
@@ -461,7 +461,9 @@ const struct group_opt_templ all_cmd_options[] = {
     {"default-solver",
      boost::program_options::value<std::string>()->value_name("<solver>"),
      "Override default solver used if no concrete one is specified"
-#ifdef BOOLECTOR
+#ifdef BITWUZLA
+     " (Bitwuzla)"
+#elif defined(BOOLECTOR)
```
rafaelsamenezes: Same as other comment

rafaelsamenezes (changes requested):
Only issue is that boolector shouldn't be specific anymore. Just use the all_solvers to pick the next solver is bitwuzla is not available.

rafaelsamenezes (approved):

src/solvers/solve.cpp
lucasccordeiro: Fixed.

src/esbmc/options.cpp
lucasccordeiro: Fixed.

lucasccordeiro:
This PR (30s):

```
  correct:           20089
    correct true:    13844
    correct false:    6245
  incorrect:            22
    incorrect true:      5
    incorrect false:    17
  unknown:           16489

```

Master (30s):

```
  correct:           19117
    correct true:    11901
    correct false:    7216
  incorrect:            19
    incorrect true:      4
    incorrect false:    15
  unknown:           17464
  Score:             30650 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24636613495/job/72034082444
```

LukeW1999:
## Why `ctest_gen_3` generates 6 test cases with Bitwuzla but 5 with Boolector

The program (`main.c`) has 3 nondet inputs — `b` (bool), `c` (char), `uc` (uchar) — and 6 branch directions to cover:

| Branch | Constrained | Free |
|--------|-------------|------|
| `b=T, c>='A'` | `b`, `c` | `uc` |
| `b=T, c<'A'` | `b`, `c` | `uc` |
| `b=F, uc>128` | `b`, `uc` | `c` |
| `b=F, uc<=128` | `b`, `uc` | `c` |
| *(+ 2 more across the top-level `b` branches)* | | |

### Root cause

For the two `b=true` branches, `uc` is unconstrained (free).

- **Boolector** tends to pick the same default value (e.g. `uc=0`) for free variables across branches. This means two distinct VCCs can produce identical `(b, c, uc)` tuples, which get collapsed by deduplication → **5 test cases**.
- **Bitwuzla** picks more diverse values during model enumeration, so all 6 VCCs yield distinct `(b, c, uc)` tuples → no dedup → **6 test cases**.

Neither solver is wrong — they simply explore the free variable space differently. The `.desc` expectation has been updated to **6**, reflecting what Bitwuzla consistently produces.

Issue #4134, opened 2026-04-21
[CPP] Dynamic cast is reported as PARSE ERROR instead of CONVERSION ERROR


zhihan:
Test case:
```
regression/esbmc-cpp/inheritance/dynamic_cast_simple_bug
```
The test expects a typechecking error (CONVERSION ERROR)


What ESBMC actually does                                                                                                               
                                                                                                                                         
  Clang rejects the assignment during AST construction; ESBMC prints ERROR: PARSING ERROR and exits. 

 The expected output is stale. 
                               
 Originally ESBMC owned its parser and a separate type-checking pass, so type errors surfaced as CONVERSION ERROR. The expectation predates the clang frontend.  
  
 Now the error is reported by clang. The error label PARSING ERROR is itself a misnomer.                                                                                 
  In src/clang-c-frontend/clang_c_language.cpp:288-324, ESBMC's parse() runs clang's full pipeline including Sema (type checking), then
  reports PARSING ERROR for any clang diagnostic. ESBMC's typecheck() (lines 326-337) only does clang-AST → IRep2 conversion. The labels 
  reflect the old hand-rolled-parser architecture, not today's reality. A type error from clang Sema gets labeled "parsing error" — which
   is what tripped the original test author's expectations.   

  The stale test expectations (^CONVERSION ERROR$) date to before 2016, when ESBMC's own parser would happily accept   
  the invalid dynamic_cast and only its separate type-check pass would reject it.

zhihan:
I think we should just update the test cases. There is no bugs to fix. I would also think detecting the error earlier is a good result.

PR #4135, opened 2026-04-21
[CPP] Fix stale knownbug tests by updating the test cases

zhihan:
Summary of the issue is https://github.com/esbmc/esbmc/issues/4134

TL;DR: these tests are stale. Update the test cases because the error code is different now.


lucasccordeiro (approved):
Well spotted, @zhihan!

PR #4138, opened 2026-04-21
[concurrency] add support for pthread_cancel and cancellation API

### LINKED ISSUE #2451: [concurrency] add support for pthread_cancel
See documentation: https://linux.die.net/man/3/pthread_cancel

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/2451.

This PR implements `pthread_cancel`, `pthread_testcancel`, `pthread_setcancelstate`, and `pthread_setcanceltype` in the operational model (`pthread_lib.c`). Per-thread cancellation state is tracked via `__ESBMC_inf_size arrays` (cancel_requested, cancelstate, canceltype), which default to 0 and match the POSIX defaults of ENABLE+DEFERRED without explicit init.

Cancellation points added to `pthread_join`, `pthread_mutex_lock_check`, and `pthread_cond_wait`; `cond_wait` atomically releases the mutex before delivering cancellation to avoid false deadlocks. 

This PR also promotes the `pthread_cleanup11` from `KNOWNBUG` to `CORE` and adds five new regression tests.


lucasccordeiro:
SV-Comp (30s): 

```
  correct:           20036
    correct true:    13796
    correct false:    6240
  incorrect:            21
    incorrect true:      5
    incorrect false:    16
  unknown:           16543
  Score:             33416 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24762302738
```

Master (30s):

```
  correct:           20089
    correct true:    13844
    correct false:    6245
  incorrect:            22
    incorrect true:      5
    incorrect false:    17
  unknown:           16489

```

Anthonysdu:
LGTM, I removed the pthread_testcancel() in pthread_mutex_lock() as none of the mutex functions is a cancelation point. Let's wait if the regression can pass.

Btw, since we don't do constant propagation for inf arrays, there is infinite loop unrolling because of pthread_cleanup_push/pop() in pthread_exit(), unless we specify unwind bound or do incremental-bmc,  we might need to optimise it in the future.

PR #4141, opened 2026-04-22
fix join(genexp) lowering in recursive scenario

sergillam:
- Promote recursive to CORE.
- Add a variant test for join(genexp) in recursive flow.
- Lower sep.join(generator) to sep.join(listcomp) in the Python frontend.
- Rewrite str(x) to x.__str__() in this path to keep object stringification behavior consistent.


lucasccordeiro (approved):

PR #4147, opened 2026-04-22
[cpp] route dynamic_cast through C++-frontend dispatch

zhihan:
Move dynamic_cast lowering out of the C frontend's structural-cast fall-through. The CXXDynamicCastExprClass arm now calls a new clang_cpp_convertert::build_dynamic_cast helper, and get_cast_expr fails loudly if any CK_Dynamic still reaches it. The helper currently mirrors the legacy gen_typecast behaviour so existing tests keep their status; follow-up patches will replace its body with the runtime vptr-comparison ITE described in docs/dynamic_cast_rtti_plan.md (steps 5.3-5.6).


src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +703,330 @@ void clang_cpp_convertert::get_overriden_methods(
...
+
+// Compute the byte offset of subobject S inside D, walking via non-virtual
+// direct bases. Returns std::nullopt if S isn't reachable that way (virtual
+// inheritance somewhere on the path or S not actually a base of D).
+static std::optional<uint64_t> offset_of_subobject(
+  const clang::ASTContext &ctx,
+  const clang::CXXRecordDecl *D,
+  const clang::CXXRecordDecl *S)
+{
+  if (D == S)
```
XLiZHI, line 723: Using `clang::CXXBasePaths` might simplify this process. 
```
clang::CXXBasePaths Paths(false, true, true);
D->isDerivedFrom(S, Paths)

Paths.getDetectedVirtual()
```

https://clang.llvm.org/doxygen/classclang_1_1CXXBasePaths.html

https://clang.llvm.org/doxygen/classclang_1_1CXXRecordDecl.html#aa241550380a15a05e18aab0838d00b92

zhihan:
I am not quite sure about the code coverage comment. @lucasccordeiro is ESBMC measuring unit test coverage, or regression test coverage?

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +703,330 @@ void clang_cpp_convertert::get_overriden_methods(
...
+  // requires this; the catch (bad_cast&) arm of any try/catch then
+  // fires). Falls back if std::bad_cast hasn't been declared (the
+  // program didn't include <typeinfo>).
+  if (is_reference)
+  {
+    const symbolt *bad_cast_sym = ns.lookup("tag-std::bad_cast");
+    if (!bad_cast_sym)
+      return fallback();
+
+    exprt src_deref = dereference_exprt(src_pointer, src_pointer.type());
```
XLiZHI: The code appears three times in this program. Maybe we can reuse it?

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +703,330 @@ void clang_cpp_convertert::get_overriden_methods(
...
+  exprt cast_or_null;
+  if (arms.empty())
+  {
+    // No reachable D can satisfy the cast — emit a typed null directly.
+    cast_or_null = typed_null;
+  }
+  else if (to_void)
+  {
+    // Per-D results differ; chain ITEs so each D selects its own offset.
+    exprt src_deref = dereference_exprt(src_pointer, src_pointer.type());
```
XLiZHI: same here

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +703,330 @@ void clang_cpp_convertert::get_overriden_methods(
...
+    cast_or_null = typed_null;
+    for (auto it = arms.rbegin(); it != arms.rend(); ++it)
+      cast_or_null = if_exprt(
+        equality_exprt(vptr_read, it->first), it->second, cast_or_null);
+  }
+  else
+  {
+    // T* form: every matching D yields the same `(T*) src`, so collapse
+    // the per-D vptr equalities into one OR-chain with a single result.
+    exprt src_deref = dereference_exprt(src_pointer, src_pointer.type());
```
XLiZHI: same here

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +703,330 @@ void clang_cpp_convertert::get_overriden_methods(
...
+  // Each entry is (vtable-variable address, per-D result expression).
+  pointer_typet vptr_type(symbol_typet(vtable_type_prefix + vptr_class_id));
+  std::vector<std::pair<exprt, exprt>> arms;
+  exprt typed_null = gen_zero(target_type);
+
+  for (const clang::CXXRecordDecl *D : translation_unit_record_decls())
+  {
+    if (!D->hasDefinition())
+      continue;
+    if (D->isAbstract())
```
XLiZHI: They should be filtered by `translation_unit_record_decls`.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3231,6 +3230,17 @@ bool clang_c_convertert::get_cast_expr(
     // identical in ESBMC's IR. Both cast directions are no-ops.
     break;
 
+  case clang::CK_Dynamic:
+    // dynamic_cast has no meaning in C and the C frontend cannot lower it
+    // soundly: it requires the C++-only vtable / class-hierarchy machinery
+    // in clang_cpp_convertert. Reaching this point means the C++ frontend
+    // forwarded a CXXDynamicCastExpr to the structural cast path instead
+    // of intercepting it in CXXDynamicCastExprClass.
+    log_error(
```
XLiZHI: I think it's acceptable to handle dynamic casting here (even though this feature comes from C++, we still have some C++-related code above), and this helps maintain consistency.

Or perhaps this particular case should be removed, considering that it is impossible to reach it.

XLiZHI (changes requested):
This is a nice feature!

But I think there are some structural issues with this method: 

The `build_dynamic_cast` function only checks this relationship when the cast occurs: it traverses all the classes of the entire TU, filters each one by calling isDerivedFrom(S), and then uses n`s.lookup` to verify if the symbol exists. These three steps are all repetitive tasks, and their complexity increases linearly with the size of the TU.

According to the current design of vtable in ESBMC, when registering symbols in the `add_vtable_*` function, we should simultaneously maintain a side table: for each processed class, we should append an item to the corresponding `vptr_class_id` list. This is an O(1) insertion and occurs within the existing registration process, without adding any additional traversals.  "build_dynamic_cast" should directly look up the table.

XLiZHI:
> I am not quite sure about the code coverage comment. @lucasccordeiro is ESBMC measuring unit test coverage, or regression test coverage?

Both of them ;)

lucasccordeiro:
> Both of them ;)

@XLiZHI is correct.

src/clang-c-frontend/clang_c_convert.cpp
zhihan: I think this code change is to catch this and error out (defensive for unreachable code). Let me change this to an assertion so it is more obvious.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
zhihan: Done

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
zhihan: Same filter moved into add_vtable_variable_symbols()

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
zhihan: Done

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +710,314 @@ void clang_cpp_convertert::get_overriden_methods(
...
+    // structural typecast is exact and no runtime check is needed.
+    if (S == T || S->isDerivedFrom(T))
+      return fallback();
+  }
+
+  // Walk up S's primary inheritance chain to find the class V that owns
+  // the vptr (the most-derived class on the chain with a
+  // virtual_table::tag-V type symbol).
+  //
+  // TODO: a virtual base on the path is treated as "give up and fall
```
XLiZHI: When there is a virtual base class on the dynamic_cast path, the code falls back.

Given that this is a TODO/unsafe operation, should we output log_error and abort? it might lead to false positives.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +710,314 @@ void clang_cpp_convertert::get_overriden_methods(
...
+      }
+      else
+      {
+        // TODO: under multiple inheritance, S may sit at a non-zero offset
+        // inside D and the adjustment is D-dependent. The correct lowering
+        // gives each D its own arm with `(T*)((char*)src + off)`. Today we
+        // emit one flat `(T*) src` for every match — correct only when
+        // every matching D has S at offset 0 (single inheritance, the
+        // most common case in practice).
+        exprt adj = src_pointer;
```
XLiZHI, line 948: I reckon this is also an unfinished part, possibly pointing to incorrect memory. We can directly terminate the verification.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -698,3 +710,314 @@ void clang_cpp_convertert::get_overriden_methods(
...
+  };
+
+  // Reference form: throw std::bad_cast on no-match (the standard
+  // requires this; the catch (bad_cast&) arm of any try/catch then
+  // fires). Falls back if std::bad_cast hasn't been declared (the
+  // program didn't include <typeinfo>).
+  if (is_reference)
+  {
+    const symbolt *bad_cast_sym = ns.lookup("tag-std::bad_cast");
+    if (!bad_cast_sym)
```
XLiZHI, line 983: If the program does not include <typeinfo>, the symbol "bad_cast" does not exist, and "dynamic_cast" becomes "typecast". The C++ standard requires that failed reference conversions must throw std::bad_cast.

XLiZHI (commented):
Thank you for addressing my comment! looks good to me.

I have only a few minor comments regarding fallback.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
zhihan: So should we get rid of the fallback and just abort here?

XLiZHI: Yes, please

zhihan:
I've updated the three places to log error and abort. PTAL.

XLiZHI:
> I've updated the three places to log error and abort. PTAL.

Hi @zhihan, I’m not seeing any new commits on my end yet. Could you please double-check if the push went through?

zhihan:
Sorry, apparently forgot to push last commit. It is there now.

XLiZHI (approved):
LGTM! 

Could you please solve the issue with the Clang format? 

GitHub Actions can correct the formatting, but it does not have the permission to modify branches outside of this repository.

Some thing like:
`find src -iname *.h -o -iname *.c -o -iname *.cpp -o -iname *.hpp -iname *.hh | grep -v src/ansi-c/cpp | grep -v src/clang-c-frontend/headers/ | grep -v src/c2goto/library/libm/musl | xargs ./clang/bin/clang-format -style=file -i -fallback-style=none`

lucasccordeiro:
@zhihan: I've granted you write access to the esbmc repo; if you create a branch from it, our GitHub Actions can automatically apply the coding style.

lucasccordeiro (approved):
Thanks for submitting this PR, @zhihan.

PR #4157, opened 2026-04-23
[build] bump pytest to >=9.0.3 (CVE-2025-71176)

lucasccordeiro:
pytest < 9.0.3 creates /tmp/pytest-of-{user} directories with insecure permissions, allowing local privilege escalation or DoS (CVSS 6.8). Update the dev-group lower bound and regenerate uv.lock to resolve to the patched 9.0.3 release.


pmatos (approved):
lgtm

PR #4162, opened 2026-04-23
[python] infer concrete types for None-initialised class attributes via usage-site scanning

### LINKED ISSUE #4117: [python] Simulated Heap via Objects
$esbmc main.py

```python
class Node:
    def __init__(self, v):
        self.value = v
        self.next = None

n1 = Node(1)
n2 = Node(2)

n1.next = n2

assert n1.next.value == 2
```

```
ESBMC version 8.2.0 64-bit aarch64 macos
[PROGRESS] Parsing main.py
[PROGRESS] Converting
[ERROR] Cannot resolve nested attribute: value
Abort trap: 6
```

LukeW1999:
## Problem

Class attributes initialised to `None` in `__init__` (e.g. linked-list `self.next = None`, binary-tree `self.left = None`) were given `any_type()` by the Python frontend. This made chained attribute accesses like `n1.next.value` unverifiable — ESBMC could not resolve the struct layout of the intermediate pointer.

Fixes #4117.

## Solution

Adds `infer_attr_type_from_usage()` in `python_converter`, which recovers the concrete type of a `None`-initialised attribute by scanning non-`None` assignments to that attribute:

1. **Module-scope priority** — scans top-level statements for `<var>.<attr> = <rhs>` where `<var>` is a known instance of the target class.
2. **Method-scope fallback** — scans `self.<attr> = <rhs>` inside the class's own methods, ignoring `None` literals.
3. **Alias propagation** — single-hop aliases (`alias = original`) are resolved so `alias.attr = ...` is attributed to the correct class.
4. **Conflict safety** — if the same attribute is assigned values of different classes, the inference aborts and falls back to `any_type()` rather than silently adopting the wrong type.

The inferred type is a lazy `pointer_to(symbol_typet("tag-Cls"))`, so the struct layout is resolved at use time via `ns.follow()` rather than snapshotted during the incomplete class build.

## Tests

| Test | Result | Covers |
|---|---|---|
| `github_4117_none_attr_type_infer` | CORE PASS | Basic linked-list pattern |
| `github_4117_none_attr_type_infer_fail` | CORE FAIL (expected) | Wrong assert correctly rejected |
| `github_4117_deep_chain` | CORE PASS | Multi-level chaining `a.next.next.value` |
| `github_4117_tree_attrs` | CORE PASS | Multiple None attrs (`.left`, `.right`) |
| `github_4117_alias_chain` | CORE PASS | Alias propagation |
| `github_4117_attr_conflict` | KNOWNBUG | Conflicting types → conservative fallback |
| `github_4117_attr_shadow` | KNOWNBUG | Variable reassigned to different class |
| `github_4117_function_internal` | KNOWNBUG | Assignments inside function bodies |

## Known Limitations (tracked as KNOWNBUG)

Three cases are deferred to follow-up PRs:

- **Function-internal tracking** — instance variables created inside function bodies are not scanned; requires scope-aware local variable tracking.
- **Variable shadowing** — a variable reassigned to a different class causes the entry to be dropped; requires flow-sensitive (per-program-point) class tracking.
- **Type conflicts** — when the same attribute receives values of two different classes, inference conservatively gives up; requires union-type modelling or a frontend diagnostic.

## Follow-up

- [ ]  function-scope instance variable tracking
- [ ]  flow-sensitive variable-to-class mapping
- [ ]  union-type / diagnostic for conflicting attribute assignments

Assisted-by: Claude-Opus4.7


lucasccordeiro (approved):

PR #4164, opened 2026-04-23
[Python] Fix dict subscript attribute access and bare-dict value type inference

prototypeC14:
This pr:
- Support `d[key].attr`
- Support class-struct dict values
- Fix bare `dict` annotation (no K/V params): peek at the literal's first value to infer list/dict value type instead of defaulting to char*
- Add a guard to the test case deduplication algorithm


lucasccordeiro (approved):

PR #4168, opened 2026-04-24
  [symex] remove dead code in builtin_functions/ (codecov cleanup)

LukeW1999:
## Summary

Cleans up the dead-code subset of the codecov coverage gap in the new `src/goto-symex/builtin_functions/` split from #4166. Covers 4 of the 8 flagged files; the other 4 were analysed and intentionally left alone (see below).

**Net change**: 68 lines removed across 4 files, 0 behavioural changes.

## Per-file breakdown

| File | Δ | What was removed |
|---|---|---|
| `memory_ops.cpp` | -53 | `offset_simplifier` div(mul,c)/c branch; `guard.is_false()` early returns in `intrinsic_memcpy`/`intrinsic_memset`; symbolic-offset and inline div-simplifier blocks in `intrinsic_memset` (redundant with the `!is_constant_int2t` check below) |
| `io.cpp` | -6 | `symex_printf` `guard.is_false()` early-continue in the pointer-argument loop; collapsed an `if (spec != 'n') / else` pair with identical bodies into a single `actual_format_count++` |
| `object_size.cpp` | -4 | Collapsed `if (param_val < 0 \|\| param_val > 3) { type_value = 0; } else ...` into a positive-guard `if`; the `type_value = 0` assignment was redundant (variable zero-initialised at declaration) |
| `run_builtin.cpp` | -5 | Unreachable `else { log_error("Unknown overflow intrinsics"); abort(); }` in the overflow-intrinsic dispatcher — the outer `has_prefix` chain matches exactly the six prefixes that `is_mult`/`is_add`/`is_sub` exhaustively cover |

## Intentionally kept

The majority of codecov-flagged lines are **defensive fallbacks or rare live paths, not dead code**. Deleting them would introduce:

- **Infinite loops** — e.g. removing `i++` from format-string parser `while`-loops, or `alloc_obj = ...` in the `while (is_if2t(alloc_obj))` unwrap loops in `threads.cpp` / `python_builtins.cpp`
- **Crashes / UB** — removing `bump_call + return` pairs lets `to_constant_int2t` assert on non-constant input; removing exception handlers leaves `type_size` uninitialised for VLAs
- **Soundness bugs** — removing the constant-size `realloc` copy loop silently returns new memory without copying; removing empty-deref early-return makes `memcpy` a no-op instead of falling back to `__memcpy_impl`
- **Feature removal** — `threads.cpp` / `python_builtins.cpp` entries ARE the implementations of the thread and Python-operator intrinsics

Files skipped entirely: `memory_alloc.cpp`, `threads.cpp`, `python_builtins.cpp`, `va_arg.cpp`.

## Why codecov flagged them

These paths are uncovered because the regression suite doesn't exercise certain input shapes — fast-path fallbacks triggered by symbolic sizes and offsets, rare type combinations (pointer memset, bitfield structs, Optional-through-void-pointer), option-gated branches (`max-symbolic-realloc-copy`, `malloc-zero-is-null`, etc.), and thread/monitor intrinsics rarely called directly by concurrency tests. None of that is dead code.

## Follow-up

I will open a separate PR shortly to **add regression tests** targeting the live-but-untested paths inventoried while doing this cleanup (constant-size `realloc`, `memset` on bitfield/union/pointer, `__builtin_object_size(&arr[i])`, `hasattr`/`isinstance` corner cases, thread intrinsic direct invocation, etc.).

Assisted-by: Claude-Opus4.7

### DIFF
--- a/src/goto-symex/builtin_functions/run_builtin.cpp
+++ b/src/goto-symex/builtin_functions/run_builtin.cpp
@@ -74,6 +74,11 @@ bool goto_symext::run_builtin(
     else if (is_sub)
       op = sub2tc(
         func_type.arguments[0], func_call.operands[0], func_call.operands[1]);
+    else
+    {
+      log_error("Unknown overflow intrinsics");
+      abort();
+    }
 
     // Perform overflow check and assign it to the return object
     if (!is_nil_expr(func_call.ret))
--- a/src/goto-symex/builtin_functions/run_builtin.cpp
+++ b/src/goto-symex/builtin_functions/run_builtin.cpp
@@ -74,11 +74,6 @@ bool goto_symext::run_builtin(
     else if (is_sub)
       op = sub2tc(
         func_type.arguments[0], func_call.operands[0], func_call.operands[1]);
-    else
-    {
-      log_error("Unknown overflow intrinsics");
-      abort();
-    }
 
     // Perform overflow check and assign it to the return object
     if (!is_nil_expr(func_call.ret))
--- a/src/goto-symex/builtin_functions/io.cpp
+++ b/src/goto-symex/builtin_functions/io.cpp
@@ -188,9 +188,6 @@ void goto_symext::symex_printf(const expr2tc &lhs, expr2tc &rhs)
       if (!is_pointer_type(arg->type))
         continue;
 
-      if (cur_state->guard.is_false())
-        continue;
-
       // Check the entire expression tree for L2 symbols, not just top-level
       bool has_l2_symbols = false;
       arg->foreach_operand([&has_l2_symbols](const expr2tc &e) {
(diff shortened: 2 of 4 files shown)


src/goto-symex/builtin_functions/run_builtin.cpp
```diff
@@ -74,11 +74,6 @@ bool goto_symext::run_builtin(
     else if (is_sub)
       op = sub2tc(
         func_type.arguments[0], func_call.operands[0], func_call.operands[1]);
-    else
```
lucasccordeiro, line 77: It is fine to keep this one.

lucasccordeiro (changes requested):

src/goto-symex/builtin_functions/run_builtin.cpp
LukeW1999: Added back.

lucasccordeiro (approved):

PR #4169, opened 2026-04-24
[symex] fix copy_count simplify + add regression tests for builtin_functions

LukeW1999:
## Summary

Follow-up to #4166 targeting the uncovered paths that codecov flagged but which were intentionally kept in #4168 (the cleanup PR) as live defensive code. Adds direct regression coverage for four of those paths and, while writing the realloc test, also fixes a one-line bug that blocked the constant-size fast path from ever firing.

## Fix

`copy_memory_content`: the `if2tc(lessthan(old, new), old, new)` expression was constructed but never simplified, so the constant-size fast path (checked by `is_constant_int2t(copy_count)` immediately below) never fired even when both element counts were compile-time constants. Adding `do_simplify` lets the fast path kick in: a 4-element realloc now emits 16 SSA assignments to the dynamic array symbol instead of the previous 128 (`max_symbolic_copy` default).

## Regression tests

| Test | Target (file:line) | Scenario |
|---|---|---|
| `realloc22_const_copy` | `memory_alloc.cpp:86` | constant-size realloc — fast path |
| `memcpy_zero_bytes` | `memory_ops.cpp:418` | zero-byte memcpy early return |
| `memset_pointer` | `memory_ops.cpp:122` | pointer-type memset (byte_update) |
| `memcpy_pointer_primitive` | `memory_ops.cpp:365` | pointer memcpy primitive fallback |

Each test was locally verified (via temporary `log_status`) to exercise the targeted branch before committing.

## Follow-up

A second batch of regression tests (bitfield/struct memset, member/index walk, `__builtin_object_size`, `__builtin_constant_p`, concurrency intrinsics) will be in a subsequent PR.

Assisted-by: Claude-Opus4.7


LukeW1999:
This PR (30s): 

```
  correct:           20005
    correct true:    13764
    correct false:    6241
  incorrect:            21
    incorrect true:      5
    incorrect false:    16
  unknown:           16574
  Score:             33353 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24890153478
```

Master (30s):

```
  correct:           20089
    correct true:    13844
    correct false:    6245
  incorrect:            22
    incorrect true:      5
    incorrect false:    17
  unknown:           16489

```

LukeW1999:
Follow-up on coverage after c85764e:

- `memory_alloc.cpp`: 94.84% (+2.97%)
- `memory_ops.cpp`: 85.50% (+4.98%)
- `cmdline.cpp`: 52.94% (+4.90%)
- `solve.cpp`: 75.38% (+3.08%)

Overall patch coverage is at 69.78%. There are still some uncovered paths remaining — continuing to work on additional regression tests in the next batch

lucasccordeiro (approved):

PR #4173, opened 2026-04-24
[smt] fix zero-width struct members in bitcast/byte ops

lucasccordeiro:
When a struct contained a zero-width member (e.g., an empty C++ class field), flatten_to_bitvector and convert_bitcast's BV->struct path emitted get_uint_type(0) constants and width-0 extracts, which both Bitwuzla and Z3 reject. Filter zero-width members out of the struct concat, substitute gen_zero for them on reconstruction, and guard byte_update/byte_extract against wholly zero-sized aggregates.


lucasccordeiro:
This PR (30s): 

```
  correct:           19989
    correct true:    13753
    correct false:    6236
  incorrect:            21
    incorrect true:      5
    incorrect false:    16
  unknown:           16590
  Score:             33326 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/24910469776
```

Master (30s):

```
  correct:           20089
    correct true:    13844
    correct false:    6245
  incorrect:            22
    incorrect true:      5
    incorrect false:    17
  unknown:           16489

```

PR #4179, opened 2026-04-25
Initial support for camada

mikhailramalho:
Route ESBMC's SMT backends through [camada](https://github.com/mikhailramalho/camada) instead of maintaining a separate adapter per solver. The per-solver adapter directories under `src/solvers/` are deleted and replaced by a single `src/solvers/camada_conv.cpp` that dispatches to the requested backend at runtime through camada's uniform API. Net effect is ~8.5k fewer lines across 883 files.

## What changes

- **New `src/solvers/camada_conv.cpp`.** Implements the solver interface on top of camada (z3, cvc5, mathsat, yices, bitwuzla, and SMT-LIB output), picked at runtime. Pinned to camada **v0.18**.
- **Deleted backends.** `src/solvers/{z3,cvc5,mathsat,yices,bitwuzla,smt,smtlib,prop,sat,minisat,boolector,cvc4,bitwuzllob,neurosym}/` are all gone, along with their `try_*` probes. The remaining `src/solvers/` is flat.
- **Solver split.** `smt_convt` is separated into a converter (`smt_conv.cpp`) and a solver shim (`smt_solver.cpp`, `smt_solver_baset`), with the theory-specific conversion split across `smt_{casts,bitcast,byteops,fp_conv,memspace,overflow}.cpp`.

## CLI changes (breaking)

These flags are **removed**, because the backends or encodings behind them no longer exist:

| removed | replacement |
|---|---|
| `--cvc`, `--cvc4` | `--cvc5` |
| `--boolector` | `--bitwuzla` (now the default) |
| `--bitwuzllob`, `--bitwuzllob-prog`, `--bitwuzllob-model-prog` | `--bitwuzla` |
| `--neurosym`, `--neurosym-prog`, `--neurosym-model-prog` | — |
| `--floatbv` | none needed: the SMT FP theory is the default |
| `--array-flattener`, `--tuple-node-flattener`, `--tuple-sym-flattener` | none: camada picks the encoding |

`--fixedbv` is removed with **no replacement**. It encoded floating-point as *fixed*-point bit-vectors, a lossy approximation that camada has no equivalent for. `--fp2bv` is not a rename of it: that flag bit-blasts IEEE floating-point exactly, and already existed alongside `--fixedbv`. Anything relying on `--fixedbv` semantics will now get different results, not merely a different spelling.

Backends still exposed: `z3`, `cvc5`, `mathsat`, `yices`, `bitwuzla`, `smtlib`.

## CMake plumbing

- `cmake_minimum_required` bumped to 3.24 (camada uses `cmake_path`).
- `Options.cmake` adds `ESBMC_CAMADA_URL`/`ESBMC_CAMADA_NAME` plus `ESBMC_CAMADA_HASH`, which pins the archive's SHA256 so a moved or tampered artefact is a hard error. The prebuilt-solver URLs for the platforms camada builds from source are dropped; the Windows ones stay, since that path still uses prebuilt Z3 and MathSAT.
- `camada::camada` is mapped across all four standard configs so the Visual Studio multi-config generator does not error on an unset `IMPORTED_LOCATION_MINSIZEREL`.
- On MSVC, `CMAKE_MSVC_RUNTIME_LIBRARY` is forwarded into the inner camada build along with `CMP0091 NEW`, keeping the /MT vs /MD CRT setting consistent and avoiding LNK2038.
- Windows-incompatible camada backends (cvc5, mathsat, yices, bitwuzla) are force-disabled at configure time on Windows regardless of `-DENABLE_*`. Only Z3 builds via camada on Windows for now.
- `libz3.dll` is installed alongside `esbmc.exe` on Windows by searching `CAMADA_DEPS_INSTALL_DIR/bin`, restoring the rule that used to live in `src/solvers/z3/CMakeLists.txt`. Without it the Windows runner exits silently at startup when the loader cannot resolve the import (thanks @XLiZHI).

## Known issues

- **Windows builds Z3 only**, as above. Re-enabling the others depends on camada gaining MSVC-friendly builds for them.


src/clang-c-frontend/clang_c_adjust_expr.cpp
```diff
@@ -1099,6 +1099,54 @@ void clang_c_adjust::do_special_functions(side_effect_expr_function_callt &expr)
       bswap_expr.operands() = expr.arguments();
       expr.swap(bswap_expr);
     }
+    else if(
```
mikhailramalho: these are unrelated changes, I'll remove later

src/esbmc/CMakeLists.txt
```diff
@@ -44,13 +44,9 @@ target_include_directories(esbmc
     PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
     PRIVATE ${Boost_INCLUDE_DIRS}
 )
-cmake_path(SET MY_MAGIC_PATH NORMALIZE "${cvc5_DIR}/../../")
-cmake_path(ABSOLUTE_PATH MY_MAGIC_PATH OUTPUT_VARIABLE CVC_LINK_DIR)
-message("Looking at ${CVC_LINK_DIR}")
 target_link_libraries(esbmc ${OLD_FRONTEND_TARGETS} ${SOLIDITY_FRONTEND_TARGETS} ${PYTHON_FRONTEND_TARGETS}
                             ${GOTO_CONTRACTOR_TARGETS} ${JIMPLE_FRONTEND_TARGETS}
-                            clangcfrontend clangcppfrontend symex langapi
+                            clangcfrontend clangcppfrontend filesystem symex pointeranalysis langapi util_esbmc bigint
```
mikhailramalho: this looks like a merge conflict going wrong

src/python-frontend/python_math.cpp
```diff
@@ -581,7 +581,7 @@ exprt python_math::handle_floor_division(
   pos_remainder.copy_to_operands(remainder, gen_zero(div_type));
 
   // diff_signals = is_num_neg ^ is_den_neg;
-  exprt diff_signals("bitxor", bool_type());
+  exprt diff_signals("xor", bool_type());
```
mikhailramalho: This is required for one of the python tests

XLiZHI:
The Camada integration removed src/solvers/z3/ which previously contained an explicit install rule for libz3.dll:

```cmake
  # Install dll for Windows
  find_file(Z3_DLL libz3.dll HINTS ${Z3_DIR} $ENV{HOME}/z3 PATH_SUFFIXES lib bin)
  if(NOT Z3_DLL STREQUAL "Z3_DLL-NOTFOUND")
    install(FILES ${Z3_DLL} DESTINATION bin)
```

Camada now builds Z3 from source and places libz3.dll under `build/camada/build/deps/install/bin/`, but nothing installs it alongside esbmc.exe. As a result, the Windows CI is failing — esbmc.exe cannot find libz3.dll at runtime, which explains why the failing tests show a completely empty PROGRAM OUTPUT: the process exits before producing any output.

Suggest adding the following to the end of `src/solvers/camada/CMakeLists.txt` to restore the old behaviour:

```cmake
  if(WIN32 AND DEFINED CAMADA_DEPS_INSTALL_DIR)
    foreach(_dll libz3.dll)
      find_file(_solver_dll ${_dll}
        HINTS "${CAMADA_DEPS_INSTALL_DIR}"
        PATH_SUFFIXES bin
        NO_DEFAULT_PATH)
      if(_solver_dll)
        message(STATUS "Windows will install ${_solver_dll}")
        install(FILES "${_solver_dll}" DESTINATION bin)
      endif()
      unset(_solver_dll CACHE)
    endforeach()
  endif()
```

mikhailramalho:
> The Camada integration removed src/solvers/z3/ which previously contained an explicit install rule for libz3.dll:
> 
> ```cmake
>   # Install dll for Windows
>   find_file(Z3_DLL libz3.dll HINTS ${Z3_DIR} $ENV{HOME}/z3 PATH_SUFFIXES lib bin)
>   if(NOT Z3_DLL STREQUAL "Z3_DLL-NOTFOUND")
>     install(FILES ${Z3_DLL} DESTINATION bin)
> ```
> 
> Camada now builds Z3 from source and places libz3.dll under `build/camada/build/deps/install/bin/`, but nothing installs it alongside esbmc.exe. As a result, the Windows CI is failing — esbmc.exe cannot find libz3.dll at runtime, which explains why the failing tests show a completely empty PROGRAM OUTPUT: the process exits before producing any output.
> 
> Suggest adding the following to the end of `src/solvers/camada/CMakeLists.txt` to restore the old behaviour:
> 
> ```cmake
>   if(WIN32 AND DEFINED CAMADA_DEPS_INSTALL_DIR)
>     foreach(_dll libz3.dll)
>       find_file(_solver_dll ${_dll}
>         HINTS "${CAMADA_DEPS_INSTALL_DIR}"
>         PATH_SUFFIXES bin
>         NO_DEFAULT_PATH)
>       if(_solver_dll)
>         message(STATUS "Windows will install ${_solver_dll}")
>         install(FILES "${_solver_dll}" DESTINATION bin)
>       endif()
>       unset(_solver_dll CACHE)
>     endforeach()
>   endif()
> ```

Thanks, I'll try that!

src/solvers/CMakeLists.txt
```diff
@@ -1,50 +1,406 @@
...
+  # Download the archive if absent or empty (a previous interrupted run).
+  if(EXISTS "${CAMADA_ARCHIVE}")
+    file(SIZE "${CAMADA_ARCHIVE}" _camada_archive_size)
+    if(_camada_archive_size EQUAL 0)
+      file(REMOVE "${CAMADA_ARCHIVE}")
+    endif()
+  endif()
+  if(NOT EXISTS "${CAMADA_ARCHIVE}")
+    message(STATUS "Downloading Camada from ${ESBMC_CAMADA_URL}")
+    file(DOWNLOAD "${ESBMC_CAMADA_URL}" "${CAMADA_ARCHIVE}"
```
lucasccordeiro: The archive that provides every bit-vector backend is downloaded with no `EXPECTED_HASH`. Pin the hash so a tampered or moved artefact fails the build.

src/solvers/CMakeLists.txt
```diff
@@ -1,50 +1,406 @@
...
+#   2. Looks for an existing camada install (CAMADA_DIR or system).
+#   3. Falls back to downloading, building, and installing camada into the
+#      build tree when DOWNLOAD_DEPENDENCIES is on.
+#   4. Wires camada into the solvercamada library and the global solver list.
 
+if(NOT (ENABLE_Z3 OR ENABLE_CVC5 OR ENABLE_MATHSAT OR ENABLE_YICES OR ENABLE_BITWUZLA))
+  return()
+endif()
+
+set(CAMADA_DIR "$ENV{HOME}/camada" CACHE PATH "Camada installation prefix")
```
lucasccordeiro: Defaulting the search path to `$ENV{HOME}/camada` lets a stale developer install silently replace the pinned download. Default to empty.

lucasccordeiro (changes requested):

src/solvers/camada_conv.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  vars.reserve(lhs.size());
+  for (const auto &var : lhs)
+    vars.push_back(var);
+
+  auto q =
+    is_forall ? solver->mkForall(vars, rhs) : solver->mkExists(vars, rhs);
+  return q;
+}
+
+std::unique_ptr<camada::SMTSolver> solver;
```
lucasccordeiro: These duplicate `smt_solver_baset`'s private members at namespace scope, giving `solver`, `oneshot_prog` and `formula_path` external linkage and static destructors. Delete lines 1285-1293.

lucasccordeiro (commented):
The description no longer matches the diff: it says `solvers/cvc4/` stays and pins camada v0.9, but CVC4, Boolector, MiniSat, bitwuzllob and neurosym backends and the `--fixedbv`/`--floatbv` flags are all deleted and Options.cmake pins v0.18. The branch is also 110 commits behind master. Refresh both before review.

src/solvers/camada_conv.cpp
```diff
@@ -0,0 +1,1468 @@
...
+    smtlib_config(options, logic));
+}
+#endif /* CAMADA_HAVE_SMTLIB */
+
+} // namespace
+
+/* Declared in solve.h: the selection table in solve.cpp points at these, so
+ * they need external linkage. Everything above stays file-local. */
+
+camada::SMTSolverRef create_esbmc_cvc5_solver(const optionst &options)
```
lucasccordeiro, line 409: Unlike every other factory here, this has no `#if CAMADA_HAVE_CVC5` guard, so a camada built without CVC5 fails to link. Add the guard and the `unsupported()` fallback.

src/solvers/camada_conv.cpp
```diff
@@ -0,0 +1,1468 @@
...
+}
+
+expr2tc smt_solver_baset::get_array_elem(
+  smt_astt array,
+  uint64_t index,
+  const type2tc &subtype)
+{
+  auto idx = make_index_expr(array->Sort->getIndexSort(), index);
+  auto elem = solver->getArrayElement(array, idx);
+  auto elem_sort = convert_sort(subtype);
```
lucasccordeiro: `elem_sort` is computed and never used. Remove it.

src/solvers/camada_conv.cpp
```diff
@@ -0,0 +1,1468 @@
...
+}
+
+smt_astt smt_solver_baset::mk_smt_fpbv_inf(bool sgn, unsigned ew, unsigned sw)
+{
+  return solver->mkInf(sgn, ew, sw, fp_encoding());
+}
+
+smt_astt smt_solver_baset::mk_smt_fpbv_rm(ieee_floatt::rounding_modet rm)
+{
+  auto sort = mk_fpbv_rm_sort();
```
lucasccordeiro: `sort` is computed and never used; `mkRM` builds the rounding-mode term itself. Remove it.

src/solvers/camada_conv.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  return fp_sign_test(op, false);
+}
+smt_astt smt_solver_baset::mk_smt_fpbv_abs(smt_astt op)
+{
+  return solver->mkFPAbs(op);
+}
+
+smt_astt smt_solver_baset::mk_from_fp_to_bv(smt_astt op)
+{
+  auto to = mk_bvfp_sort(sort_fp_ew(op->Sort), sort_fp_sw(op->Sort));
```
lucasccordeiro: `to` is computed and never used; `mkIEEEFPToBV` derives the target sort itself. Remove it.

docs/roadmap/fixedbv-tr18037-plan.md
```diff
@@ -0,0 +1,182 @@
+# Plan: native fixed-point (_Fract/_Accum/_Sat) support in ESBMC
```
lucasccordeiro, line 1: A development planning note does not belong in the repository root. Drop it from the PR or move it under `docs/`.

PR #4181, opened 2026-04-25
[goto-programs] port goto_coverage condition helpers to native irep2

lucasccordeiro:
Migrate handle_single_guard, handle_operands_guard, gen_cond_cov_assert,
add_cond_cov_assert, gen_not_eq_expr, gen_and_expr, and gen_not_expr
from legacy exprt to native expr2tc. Walk operands generically via
Foreach_operand and replace string-id checks with is_*_2t predicates.
Removes the seven migrate_expr round-trips at the helper boundaries.
Behaviour preserved against master on the goto-coverage suite.


src/goto-programs/goto_coverage.cpp
```diff
@@ -570,88 +540,94 @@ void goto_coveraget::condition_coverage()
   => assert(!(!b==0 && c>90))
   => assert(!(!(b==0) && !(c>90)))
 */
+/// Returns true iff @p e is one of the relational operators that
+/// gen_cond_cov_assert treats as an instrumented condition leaf.
+static bool is_comparison_op(const expr2tc &e)
```
mikhailramalho: This should probably be moved to the irep2 utils header

lucasccordeiro: Implemented as suggested.

ChenfengWei0 (approved):
LGTM

PR #4184, opened 2026-04-25
[cpp] guard NestedNameSpecifier::getAsType on Clang 22 (#4180)

### LINKED ISSUE #4180: C++ frontend: two converter assertions on standard C++20 code (std::array; namespace-qualified constexpr)
## Summary

While verifying [NVIDIA OpenSMA](https://github.com/NVIDIA/OpenSMA) (an open MCU firmware codebase, C++20) with ESBMC 8.2.0, I hit two distinct C++ frontend assertion failures, each reducible to a tiny standalone source file. Both kill the converter before it produces a GOTO program, so neither the codebase nor any non-trivial template-using C++20 program can currently be analysed on this build.

## Environment

- ESBMC: `8.2.0 64-bit aarch64 macos` (built from source at the time of writing)
- Host: macOS (Darwin 25.4.0), Apple Silicon (aarch64)
- Invocation: `esbmc --std c++20 <file>`

The version string from `--version`:

```
ESBMC version 8.2.0 64-bit aarch64 macos
```

Both bugs reproduce regardless of `--std` (`c++14`, `c++17`, `c++20`, default).

---

### Reproducer (12 lines)

```cpp
#include <array>
#include <cstdint>

int main()
{
    std::array<uint8_t, 4> a{};
    return a[0];
}
```

### Output

```
Parsing bug1_array_crash.cpp
Converting
Assertion failed: (ctor_class_symb), function gen_vptr_initializations, file clang_cpp_adjust_code_gen.cpp, line 48.
```

### Notes

- A plain `[[gnu::packed]]` struct with bit-fields verifies cleanly, so packed-struct handling is not the trigger; the crash is in instantiation of the bundled `<array>` template.
- Workaround: shadow `<array>` with a minimal POD shim (struct wrapping a C array). This unblocks downstream code that consumes `std::array<uint8_t, N>` for fixed-size byte buffers (very common in protocol parsers).

---

### Reproducer (24 lines, no STL)

```cpp
#include <cstdint>

namespace outer {
namespace inner {
constexpr unsigned long Value = 64;
}  // namespace inner
}  // namespace outer

namespace outer {
struct Header
{
    uint8_t a;
};
// Trigger: qualified `inner::Value` used in a sibling-namespace constexpr
// initialiser involving sizeof.
constexpr unsigned long DataLen = inner::Value + sizeof(Header);
}  // namespace outer

int main()
{
    return (int)outer::DataLen;
}
```

### Output

```
Parsing bug2_qualified_constexpr.cpp
Converting
Assertion failed: (Kind == StoredKind::Type), function getAsType, file NestedNameSpecifierBase.h, line 159.
```

### Notes

- Adding `using inner::Value;` before the `constexpr` line avoids the crash, suggesting the converter's `NestedNameSpecifier` handler trips when the qualifier resolves to a *namespace* rather than a *type* in a constant-expression context.
- The pattern is idiomatic and appears verbatim in NVIDIA OpenSMA at `corepdk/modules/mctp-cpp/src/app/pdk-mctp-app-packet.h:27`:
  ```cpp
  constexpr auto PktBufDataLen = platforms::TransmitUnit + sizeof(TransportHeader);
  ```
  where `platforms` is a sibling namespace and `TransmitUnit` is `constexpr size_t`.

---

## Why this matters

Either bug alone blocks verification of any non-trivial modern C++ codebase that uses `std::array` or namespace-qualified `constexpr`. Both showed up on the very first verification target in OpenSMA. I'm reporting both together since they were discovered in the same triage; happy to split into two issues if that's the maintainers' preference.

I have local workarounds (shim `<array>` and a `using` declaration) and have proceeded with verification, but a proper fix in the C++ frontend would remove the need for either workaround.

lucasccordeiro:
Fixes https://github.com/esbmc/esbmc/issues/4180.

Clang 22's `NestedNameSpecifier::getAsType()` asserts when the qualifier is not a Type kind (e.g., a namespace), so namespace-qualified constexpr initializers such as `inner::Value + sizeof(Header)` aborted the converter. 

This PR checks the kind before calling `getAsType()` in the C frontend's `DeclRefExpr` path, and adds a matching precondition assert on the member-pointer qualifier in the C++ frontend (always a class type).


mikhailramalho (approved):
lgtm

PR #4186, opened 2026-04-25
[ci] cap regression timeout and move long-running tests to disabled/

lucasccordeiro:
Cap the per-test regression timeout at 120s on the build-unix Linux
runner so individual outliers fail fast instead of stranding the whole
job under a runner shutdown signal. The cap is applied via
ESBMC_REGRESS_TIMEOUT (the cache variable that feeds the per-test
TIMEOUT property) because `ctest --timeout` is ignored once tests set
TIMEOUT via set_tests_properties. Twenty-five tests that exceeded the
cap are moved under regression/disabled/ (not registered with CTest);
THOROUGH-mode tests get a `_thorough` suffix, CORE-mode tests get
`_disabled`, matching the existing disabled/esbmc-unix/ convention.


LukeW1999:
Hi @lucasccordeiro — small thing before approving. 

Locally, `--timeout 60` doesn't seem to bite on regression tests because they all set `TIMEOUT 1200` per-test (https://github.com/esbmc/esbmc/blob/d2e4dc6d6de3c867e3d9c6ea58c10b6d128ab739/regression/CMakeLists.txt#L19 and https://github.com/esbmc/esbmc/blob/d2e4dc6d6de3c867e3d9c6ea58c10b6d128ab739/regression/CMakeLists.txt#L53), which CTest treats as overriding the cmd-line flag. e.g. `ctest -R basic_insert_fail --timeout 5` still passes after 14s. 

Would `-DESBMC_REGRESS_TIMEOUT=120` at cmake configure get the cap you want? Might be missing context, just flagging.

lucasccordeiro:
@LukeW1999,

Good catch! You're right. 

I've just pushed a follow-up that overrides ESBMC_REGRESS_TIMEOUT at configure time (which is what actually feeds the property) and drops the no-op --timeout from ctest. 

Thanks!

LukeW1999 (approved):

LukeW1999:
LGTM

PR #4197, opened 2026-04-26
support advanced dataclass flags, derived methods and inheritance

sergillam:
Extends the Python frontend's dataclass preprocessor with class-level decorator flags, per-field flags, structural inheritance, and synthesis of comparison dunders. 

- Decorator kwargs: parses init, repr, eq, order, frozen, unsafe_hash, kw_only, slots, match_args from @dataclass(...) with strict boolean validation.
- Compatibility matrix: rejects order=True without eq=True, rejects unsafe_hash=True combined with frozen=True and eq=True, rejects unknown options and non-default fields after defaults — all as deterministic SyntaxErrors.
- Per-field flags: parses init, repr, compare, hash, kw_only from field(...). init=False removes the field from __init__ parameters while still assigning its default in the body. kw_only moves the field to kwonlyargs.
- Structural inheritance: merges fields from dataclass bases ahead of derived fields; derived overrides keep the base position and apply the new default. Works across multi-level hierarchies.
- Derived methods: synthesizes __eq__ (default eq=True) and __lt__ / __le__ / __gt__ / __ge__ (when order=True) over compare-eligible fields. User-defined methods are preserved.

### DIFF
--- a/regression/esbmc-unix/01_malloc_08/test.desc
+++ b/regression/esbmc-unix/01_malloc_08/test.desc
@@ -1,4 +1,4 @@
 THOROUGH
 main.c
---unwind 1 --context-bound 1
+--unwind 1 --context-bound 1 --z3
 ^VERIFICATION FAILED$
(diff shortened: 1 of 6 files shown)


lucasccordeiro (approved):

PR #4202, opened 2026-04-27
[ir-ieee] fix isinf/isfinite semantics under --ir-ieee encoding

ShemaAlosaimi:
## Summary

Fix unsound `isinf` / `isfinite` semantics for `floatbv` values under `--ir-ieee` integer/real encoding.

## What changed

* `convert_is_inf()` no longer returns `false` unconditionally in `int_encoding`
* `convert_is_finite()` no longer returns `true` unconditionally in `int_encoding`
* both predicates are now encoded using exact max-normal IEEE thresholds
* `get_single_max_normal()` and `get_double_max_normal()` now use exact values instead of rounded decimal approximations.


lucasccordeiro:
@claude: could you please review this PR?

lucasccordeiro (approved):

PR #4209, opened 2026-04-28
[bmc] add vacuity probe for loop-invariant verification (adds flags `--check-vacuity` and `--no-check-vacuity`)

Yiannis128:
A loop invariant that implies the loop guard makes the post-loop continuation unreachable, so every downstream claim (including synthesized `__ESBMC_ensures`) discharges as vacuously true. ESBMC would report `VERIFICATION SUCCESSFUL` for clearly-bogus contracts on buggy non-terminating bodies, which is a real soundness gap for LLM-driven contract synthesis (issue [#2](https://github.com/Yiannis128/esbmc/issues/2)).

After `UNSAT` discharge, re-solve the path assumptions alone in a fresh solver. `UNSAT` means the path was unreachable -> map the verdict to `VERIFICATION UNKNOWN` (matching the existing semantic for `k-induction` inductive-step `UNKNOWN`: "couldn't soundly conclude", not "found a bug").

The probe is gated on `--check-vacuity`, default-on under `--loop-invariant-check`, opt-out via `--no-vacuity-check`. Synthetic invariant-base-case / invariant-inductive-step claims are skipped: they sit under the loop pass's `ASSUME(false)` terminator by design.

https://link.springer.com/chapter/10.1007/3-540-63166-6_28


LukeW1999:
Hi @Yiannis128

Nice work overall — the per-claim slicing and the gating around k-induction modes are exactly right. I ran 15 hand-written scenarios (lying invariant, honest invariant with a buggy ensures, nested loops, multi-property mode, zero-iter loops, claim-before-loop, opt-out flag) plus all 248 function_contract regression tests, and everything passes. Soundness and precision look solid.

A few observations worth flagging:

**Per-claim probe overhead.** Each kept claim triggers a `symex_target_equationt vac_eq = *eq` copy plus a fresh solver call. On a 3-loop / 7-claim toy program I measured roughly +7% time and +17% memory, and it scales linearly with claim count. If the overhead becomes a concern on larger programs, one option worth exploring would be using push/pop on a single solver instance instead — the equation gets encoded once, and each claim's probe becomes an incremental `push → assert path assumption → check → pop`, which would also let the solver reuse learned clauses across probes. Caveat: this push/pop suggestion came from Claude, and I haven't verified how it would actually slot into ESBMC's `smt_convt` — directionally it looks like it should ease the memory pressure, but treat it as a pointer rather than a concrete plan.

**`runtime_encoded_equationt::convert(smt_convt&, bool)` silently ignores the vacuity flag.** If a vacuity probe ever gets routed through the incremental encoding path, it'll produce normal-mode results with no indication. Probably fine today since nothing calls it that way, but an `assert(!vacuity_mode)` or a note on the virtual would prevent a future silent-failure footgun.

**Warning text can mislead when the path is unreachable for a non-invariant reason.** For example, with `__ESBMC_requires(n >= 1 && n <= 0)` and a perfectly sane invariant, the probe still fires and reports "loop invariant likely implies the loop guard". The verdict (UNKNOWN) is correct, but the message points at the wrong place. Something like "path assumptions are unsatisfiable; possible causes include an over-constrained invariant, requires clause, or upstream assume" would generalise correctly. A small regression test for the requires-induced case would also be nice to have.

Yiannis128:
> Hi @Yiannis128
> 
> Nice work overall — the per-claim slicing and the gating around k-induction modes are exactly right. I ran 15 hand-written scenarios (lying invariant, honest invariant with a buggy ensures, nested loops, multi-property mode, zero-iter loops, claim-before-loop, opt-out flag) plus all 248 function_contract regression tests, and everything passes. Soundness and precision look solid.
> 
> A few observations worth flagging:
> 
> **Per-claim probe overhead.** Each kept claim triggers a `symex_target_equationt vac_eq = *eq` copy plus a fresh solver call. On a 3-loop / 7-claim toy program I measured roughly +7% time and +17% memory, and it scales linearly with claim count. If the overhead becomes a concern on larger programs, one option worth exploring would be using push/pop on a single solver instance instead — the equation gets encoded once, and each claim's probe becomes an incremental `push → assert path assumption → check → pop`, which would also let the solver reuse learned clauses across probes. Caveat: this push/pop suggestion came from Claude, and I haven't verified how it would actually slot into ESBMC's `smt_convt` — directionally it looks like it should ease the memory pressure, but treat it as a pointer rather than a concrete plan.
> 
> **`runtime_encoded_equationt::convert(smt_convt&, bool)` silently ignores the vacuity flag.** If a vacuity probe ever gets routed through the incremental encoding path, it'll produce normal-mode results with no indication. Probably fine today since nothing calls it that way, but an `assert(!vacuity_mode)` or a note on the virtual would prevent a future silent-failure footgun.
> 
> **Warning text can mislead when the path is unreachable for a non-invariant reason.** For example, with `__ESBMC_requires(n >= 1 && n <= 0)` and a perfectly sane invariant, the probe still fires and reports "loop invariant likely implies the loop guard". The verdict (UNKNOWN) is correct, but the message points at the wrong place. Something like "path assumptions are unsatisfiable; possible causes include an over-constrained invariant, requires clause, or upstream assume" would generalise correctly. A small regression test for the requires-induced case would also be nice to have.

Hi Weiqi, thanks for taking the time to review this. I'll address your concerns and resubmit for review 🚀

LukeW1999:
Hi @Yiannis128, hope you are doing well.

Sorry — I didn't think this through clearly the first time. One follow-up:

ESBMC already splits the two concerns via flags:
- `--enforce-contract` — function-level: assume `requires`, check body
- `--replace-call-with-contract` — system-level: check `requires` at each call site

Under `--enforce-contract`, a contradictory `requires` (e.g. `n >= 1 && n <= 0`) makes the body check vacuous — but that's arguably fine at this layer, since the contract is "assuming requires, body holds". The real failure naturally surfaces under `--replace-call-with-contract` as a concrete call-site FAIL.

So I'm wondering if the requires-induced case should be a **WARNING with verdict SUCCESSFUL** rather than UNKNOWN — let the system-level mode produce the actionable error. The lying-invariant case (the original motivation) is genuinely different: `requires` is satisfiable, and unreachability comes from the invariant itself. If the probe could separate those two, UNKNOWN stays meaningful for the lying-invariant case, and we avoid overloading it for spec-suspicion.

Curious what you think — happy to be told I'm missing something.

LukeW1999 (approved):
Hi @Yiannis128, quick correction first — I misread the scope of the earlier discussion. The issue is about loop invariants, not function contracts as I thought. Thanks for taking the time to walk me through it earlier today. The three regression tests look correct.

One small thing worth flagging: with `--loop-invariant-check`, the probe now fires on any dead path, including programs with no loop invariant at all — e.g. just a pair of contradictory __ESBMC_assume calls. 

```
int main() {
  __ESBMC_assume(x == 1);
  __ESBMC_assume(x == 2);
  assert(x == 999);
  return 0;
}

```

Master returns VERIFICATION SUCCESSFUL for those; this branch returns VERIFICATION UNKNOWN. Without `--loop-invariant-check` the behavior is unchanged, so it's not a wide regression, but it might surprise users who add the flag to a program that happens to have dead paths for unrelated reasons. Might be worth a note in the docs or a scoped check. Happy to be told this is intentional though.

Yiannis128:
> Hi @Yiannis128, quick correction first — I misread the scope of the earlier discussion. The issue is about loop invariants, not function contracts as I thought. Thanks for taking the time to walk me through it earlier today. The three regression tests look correct.
> 
> One small thing worth flagging: with `--loop-invariant-check`, the probe now fires on any dead path, including programs with no loop invariant at all — e.g. just a pair of contradictory __ESBMC_assume calls.
> 
> ```
> extern int __VERIFIER_nondet_int(void);
> int main() {
>   int x = __VERIFIER_nondet_int();
>   __ESBMC_assume(x == 1);
>   __ESBMC_assume(x == 2);
>   assert(x == 999);
>   return 0;
> }
> ```
> 
> Master returns VERIFICATION SUCCESSFUL for those; this branch returns VERIFICATION UNKNOWN. Without `--loop-invariant-check` the behavior is unchanged, so it's not a wide regression, but it might surprise users who add the flag to a program that happens to have dead paths for unrelated reasons. Might be worth a note in the docs or a scoped check. Happy to be told this is intentional though.

This is another correctness issue imo, shouldn't this be VERIFICATION UNKNOWN because the assert contradicts the __ESBMC_assume?

LukeW1999 (approved):
LGTM

Yiannis128:
@rafaelsamenezes we have resolved our discussion, but I would like to ask if you think it's better to have `--check-vacuity` turned on by default when `--loop-invariant-check` is used (current behaviour) and can be turned off with `--no-vacuity-check`?

rafaelsamenezes:
> @rafaelsamenezes we have resolved our discussion, but I would like to ask if you think it's better to have `--check-vacuity` turned on by default when `--loop-invariant-check` is used (current behaviour) and can be turned off with `--no-vacuity-check`?

I think it should be on by default.

PR #4210, opened 2026-04-28
[Abstract-interpretation] add linear equality analysis

XLiZHI:
```c
int main(void) {
  unsigned int x = __VERIFIER_nondet_uint();
  unsigned int y = x;

  while (x < 1024) {
    x++;
    y++;
  }

  assert(x == y);
}
```
  Step 1: State Before the Loop:

```
  y = x → Case A (RHS does not mention y), record y == 1·x + 0:
  state: { y == x }
```

  Step 2: Merge at the Loop Head

```
  The loop head has two incoming edges — take their intersection.

  Edge 1 (first entry into loop): { y == x }

  Edge 2 (back-edge): simulate one loop body first:

  x = x + 1 → Case B (x appears on RHS with coefficient 1)

  Invert: x_old = x_new - 1, substitute into other equations:

  y == x_old  →  y == x_new - 1

  state: { y == x - 1 }

  y = y + 1 → Case B again

  Invert: y_old = y_new - 1. y has an old equation old_eq = x - 1, update it:

  y_new == old_eq + 1 == (x - 1) + 1 == x

  state: { y == x }   ← identical to the entry state

  Intersection:

  edge 1: { y == x }
  edge 2: { y == x }
  merge:  { y == x }   ← both agree, fixed point reached after one iteration
```

  Step 3: Inject the Invariant

  `esbmc main.c --linear-equality-analysis --k-induction`

### DIFF
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -288,7 +288,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
     if concurrency:
       command_line += "--no-pointer-check --no-bounds-check "
     else:
-      command_line += "--no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind "
+      command_line += "--no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --linear-equality-analysis "
   elif prop == Property.datarace:
     # TODO: can we do better in case 'concurrency == False'?
     command_line += "--no-pointer-check --no-bounds-check --data-races-check-only --no-assertions "
--- a/src/goto-programs/abstract-interpretation/linear_equality_analysis.cpp
+++ b/src/goto-programs/abstract-interpretation/linear_equality_analysis.cpp
@@ -1,0 +1,132 @@
+/// \file
+/// Linear Equality Analysis
+///
+/// Algorithm:
+///   1. Run ait<linear_equality_domaint> to compute a fixedpoint of linear
+///      equality states over the full GOTO control-flow graph.
+///   2. For each loop, collect the variables modified or read inside it.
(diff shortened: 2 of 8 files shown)


scripts/competitions/svcomp/esbmc-wrapper.py
```diff
@@ -288,7 +288,7 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci)
     if concurrency:
       command_line += "--no-pointer-check --no-bounds-check "
     else:
-      command_line += "--no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind "
+      command_line += "--no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --linear-equality-analysis "
```
lucasccordeiro, line 291: Enabling an untested analysis in the SV-COMP wrapper turns any unsound invariant into a wrong competition verdict. Land the analysis with tests first.

src/goto-programs/abstract-interpretation/linear_equality_analysis.cpp
```diff
@@ -0,0 +1,132 @@
...
+  expr2tc invariant = state.to_predicate(loop_vars);
+  if (is_true(invariant))
+    return;
+
+  // head_it points to the loop-condition IF; insert after it so the ASSUME
+  // lands at the body entry and is assumed on every iteration.
+  auto body_it = std::next(head_it);
+  goto_programt::instructiont assume_insn;
+  assume_insn.make_assumption(invariant);
+  assume_insn.inductive_step_instruction = config.options.is_kind();
```
lucasccordeiro, line 64: The ASSUME is inserted on the plain-BMC path too, not just k-induction, so an unsound invariant prunes real counterexamples. Restrict insertion to `is_kind()`.

lucasccordeiro (changes requested):
No regression tests: this injects ASSUMEs derived from a brand-new abstract domain, where a single wrong equality turns into a false VERIFICATION SUCCESSFUL. At minimum add passing and failing tests per supported statement form.

PR #4212, opened 2026-04-28
[contracts] fix __ESBMC_is_fresh codegen for bare-pointer form

Yiannis128:
The requires-side wrapper had a fallback else branch that treated the bare-pointer form __ESBMC_is_fresh(p, n) as if the malloc result should be assigned through *p, dereferencing an uninitialised pointer before the body executed. The canonical surface form per the docs now flows through the same simple path as the &p form: the malloc result is assigned directly to the pointer parameter.

Also extend is_fresh_allocated_params tracking to recognise the bare symbol form so add_pointer_validity_assumptions does not overwrite the fresh allocation with a single-element malloc.

Adds four regression tests covering the canonical form: an int* in-bounds dereference, an int* OOB read, a char* (alignment-1) case that previously tripped the broken else branch, and a structural twin of init_controller_pass.


LukeW1999 (approved):
LGTM, Clear and well tested.

PR #4226, opened 2026-04-30
Fix conditions that are always true or false (Cppcheck)

lucasccordeiro:
Remove redundant and dead conditions flagged by Cppcheck across seven
files: always-true guards in string.c, irep2_type.cpp, dereference.cpp,
and smt_casts.cpp; always-false guards in loop_unroll.cpp and
pthread_lib.c; and a dead has_unknown variable in symex_main.cpp whose
only setter was commented out.


src/goto-programs/loop_unroll.cpp
```diff
@@ -159,7 +159,7 @@ int bounded_loop_unroller::get_loop_bounds(loopst &loop)
     }
 
   int bound = k - k0;
-  if (bound <= 0 || (size_t)bound > unroll_limit)
+  if ((size_t)bound > unroll_limit)
```
rafaelsamenezes, line 163: `assert(k >= k0)`

src/irep2/irep2_type.cpp
```diff
@@ -287,14 +287,12 @@ unsigned int struct_union_data::get_component_number(const irep_idt &comp) const
       name);
     abort();
   }
-  else if (count > 1)
+  else
```
rafaelsamenezes: Might as well remove the else

rafaelsamenezes (changes requested):

src/goto-programs/loop_unroll.cpp
lucasccordeiro: Fixed.

src/irep2/irep2_type.cpp
lucasccordeiro: Fixed.

src/c2goto/library/pthread_lib.c
```diff
@@ -776,29 +776,12 @@ __ESBMC_HIDE:;
     key != NULL,
     "In pthread_key_create, key parameter must be different than NULL.");
   // the value NULL shall be associated with the new key in all active threads
-  int result = insert_key_value(__ESBMC_next_thread_key, NULL);
+  insert_key_value(__ESBMC_next_thread_key, NULL);
```
intrigus-lgtm, line 778: Change insert_key_value's return type to void as well?

lucasccordeiro: Fixed.

PR #4233, opened 2026-04-30
[clang-c-frontend] fix aggregate-init crash for derived struct with bitfield base (#4232)

### LINKED ISSUE #4232: [solver] mk_eq / to_solver_smt_ast crash persists after #4217: bitfield-base struct + switch-case + member read
## Summary

PR #4217 closed #4216 but two related crashes persist on the latest `master`. Both are triggered by a struct that inherits from a bitfield base, combined with a `switch` on `static_cast<enum>(field)` and a member read in the case body.

## Environment

```
esbmc --version   # latest master post-#4217
Platform: macOS aarch64
Solver: bitwuzla (default), z3
```

## Crash A — regression introduced by #4217

The original repro from #4216 (`(void)req.data[1]` in the case body) now crashes a different assertion instead of `mk_eq`:

```
Assertion failed: (to_solver_smt_ast != nullptr),
  function to_solver_smt_ast, file smt_ast.h:111
```

**Minimal reproducer** (same `mkeq_width_mismatch.cpp` from the #4216 repro set):

```cpp
#include <cstdint>
enum class E : uint8_t { A = 0 };
struct Base { uint8_t hi : 4; uint8_t lo : 4; };
struct Req : Base { uint8_t data[2]; };
extern "C" uint8_t nondet_u8();
int main() {
    uint8_t idx = nondet_u8();
    __ESBMC_assume(idx >= 1);
    Req req{};
    req.data[1] = nondet_u8();
    switch (static_cast<E>(req.data[0])) {
        case E::A:
            __ESBMC_assert(idx < 1, "OOB");
            (void)req.data[1];   // removing this line => no crash
            break;
        default: break;
    }
    return 0;
}
```

**Command:**
```
esbmc --std c++20 mkeq_width_mismatch.cpp
```

**Expected:** VERIFICATION FAILED ("OOB")
**Actual:** crash at `to_solver_smt_ast, smt_ast.h:111`

## Crash B — original mk_eq crash still present

When the read is consumed by a function call (rather than `(void)`) the original `mk_eq` assertion still fires:

```
Assertion failed: (a->sort->get_data_width() == b->sort->get_data_width()),
  function mk_eq, file bitwuzla_conv.cpp:512
```

**Minimal reproducer:**

```cpp
#include <cstdint>
enum class E : uint8_t { A = 0 };
struct Base { uint8_t hi : 4; uint8_t lo : 4; };
struct Req : Base { uint8_t data[2]; };
extern "C" uint8_t nondet_u8();
static void sink(uint8_t) {}
int main() {
    uint8_t idx = nondet_u8();
    __ESBMC_assume(idx >= 1);
    Req req{};
    req.data[1] = nondet_u8();
    switch (static_cast<E>(req.data[0])) {
        case E::A:
            __ESBMC_assert(idx < 1, "OOB");
            sink(req.data[1]);   // function-call variant triggers mk_eq crash
            break;
        default: break;
    }
    return 0;
}
```

**Command:**
```
esbmc --std c++20 mkeq_width_mismatch_b.cpp
```

**Expected:** VERIFICATION FAILED ("OOB")
**Actual:** crash at `mk_eq, bitwuzla_conv.cpp:512`

## Impact

These crashes block upgrading the NVIDIA/OpenSMA verification harness for `Control::on_set_endpoint_id()`. The production method uses exactly this pattern: `Packet` inherits from a bitfield header struct, the dispatch is `switch(static_cast<Cmd>(msg[2]))`, and the matching case body reads `msg[3]` and `msg[4]`.

## Workaround

None known. The harness currently calls `set_cur_eid()` directly after `validate()`, bypassing the dispatch layer, until this is resolved.

lucasccordeiro:
Fixes #4232.

In C++17/20 aggregate-init of `struct Req : Base`, Clang emits a nested `InitListExpr` for the `Base` sub-object, but ESBMC's IR flattens all base-class components into the derived struct's component list. The one-to-one matching loop assigned the entire `Base` struct literal to the first bitfield component, producing a sort mismatch that crashed at `to_solver_smt_ast` (Crash A) and `mk_eq` (Crash B).

The fix adds `get_base_flattened_inits()` to recursively expand base-class nested `InitListExpr` nodes into a flat initializer list before component matching. It is a no-op for non-derived structs and plain arrays/vectors. Adds two regression tests for #4232.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_4232/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_4232/main.cpp
@@ -1,0 +1,29 @@
+// Reproducer for https://github.com/esbmc/esbmc/issues/4232
+// Derived struct with bitfield base + switch on static_cast<enum>(field).
+// Was crashing: assert(r) in to_solver_smt_ast due to wrong aggregate init.
+#include <cstdint>
+
+enum class E : uint8_t { A = 0 };
+
+struct Base { uint8_t hi : 4; uint8_t lo : 4; };
+struct Req : Base { uint8_t data[2]; };
+
+extern "C" uint8_t nondet_u8();
+
+int main()
+{
+    uint8_t idx = nondet_u8();
+    __ESBMC_assume(idx >= 1);
+
+    Req req{};
+    req.data[1] = nondet_u8();
+
+    switch (static_cast<E>(req.data[0])) {
+        case E::A:
+            __ESBMC_assert(idx < 1, "OOB");
+            (void)req.data[1];
+            break;
+        default: break;
+    }
+    return 0;
+}
--- a/regression/esbmc-cpp/cpp/github_4232/test.desc
+++ b/regression/esbmc-cpp/cpp/github_4232/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.cpp
+--std c++20 --incremental-bmc
+^VERIFICATION FAILED$
+^.*OOB.*$
--- a/regression/esbmc-cpp/cpp/github_4232_pass/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_4232_pass/main.cpp
@@ -1,0 +1,24 @@
+// Passing variant for https://github.com/esbmc/esbmc/issues/4232
+// Same struct/enum pattern, but the assertion holds.
+#include <cstdint>
+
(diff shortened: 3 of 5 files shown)


XLiZHI (approved):
LGTM

PR #4236, opened 2026-05-01
[Solidity] comprehensive refactoring and implementation

ChenfengWei0:
Incremental update on top of the existing Solidity frontend:

- Add 12 C operational models in src/c2goto/library/solidity/ (mappings, dynamic arrays, bytes, abi.*, keccak256/sha256/ecrecover, EOA balance, blockchain/tx/msg context, ether/time units, string, address, builtins, misc) compiled into a separate sol64 goto binary loaded at runtime.
- Drop the inline solidity_template.h (-1442 lines) in favour of the dedicated model library above.
- Rework the type, expression, statement, tuple, ref, mapping, and inheritance converters in src/solidity-frontend/.
- Add --solc-bin and --focus-function CLI options.
- Add 173 regression tests covering the new models and converters.
- Add user-facing documentation under website/content/docs/solidity/.
- Miscellaneous compatibility tweaks to z3, smt, irep2, langapi, goto-programs.

Assisted-by: Claude-Opus4.6

---

Notably, the following non-solidity-frontend files have been modified:
- src/solvers/smt/smt_conv.cpp
- src/util/expr_util.cpp
- src/solvers/z3/z3_conv.cpp
- src/big-int/bigint.cpp


ChenfengWei0:
The `PR / Check C/C++ code-style` job fails because the workflow does not filter out deleted files when collecting the change set. This PR removes `src/solidity-frontend/solidity_template.h`.

Assisted-by: Claude-Opus4.6

PR #4242, opened 2026-05-01
[goto-check] flag negative shift distance under --overflow-check

### LINKED ISSUE #2789: [goto-check] extend goto-check to verify when shift distance is negative

lucasccordeiro:
Per C11 §6.5.7p3 and C++ [expr.shift]/1, a shift by a negative or out-of-range distance is undefined behavior. The shift-distance UB assertions in `shift_check()` were previously gated only on `--ub-shift-check`; extend the gate so they also fire under `--overflow-check` / `--unsigned-overflow-check`, since shift-distance UB is a form of shift-related overflow.

Fixes #2789.

### DIFF
--- a/regression/esbmc/github_2789/main.c
+++ b/regression/esbmc/github_2789/main.c
@@ -1,0 +1,12 @@
+// https://github.com/esbmc/esbmc/issues/2789
+// C11 §6.5.7p3: shift by a negative amount is undefined behavior.
+#include <assert.h>
+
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/25226104778.

PR #4290, opened 2026-05-04
[om] add C11 <threads.h> operational model

### LINKED ISSUE #3449: [Bug] ESBMC fails to parse C11 `<threads.h>` due to header shadowing conflict
## Summary

ESBMC cannot parse files that `#include <threads.h>` (C11 threads). The error is `unknown type name '__tss_t'`.

## Root Cause

ESBMC provides `bits/thread-shared-types.h` as a forwarding header that redirects to its `pthread.h` operational model. This works for C++ (libc++ `__threading_support`), but breaks C11 `<threads.h>`:

1. ESBMC has no `threads.h` operational model, so the real system `/usr/include/threads.h` is used
2. System `threads.h` includes `<bits/thread-shared-types.h>`, expecting glibc internal types (`__tss_t`, `__thrd_t`, etc.)
3. ESBMC's `-isystem` path shadows the real glibc `bits/thread-shared-types.h` with its own version, which only does `#include <pthread.h>`
4. The glibc internal types are never defined, causing parse errors

## Reproduction

```c
// test.c
#include <threads.h>
int main() { return 0; }
```

```
esbmc test.c
```

Fails with:
```
/usr/include/threads.h:37: error: unknown type name '__tss_t'
```

## Possible Fixes

- Add a `threads.h` operational model to ESBMC (wrapping the existing pthread model, mirroring how glibc implements C11 threads over pthreads)
- Or update `bits/thread-shared-types.h` to also define the glibc internal types that system `threads.h` expects

Yiannis128:
## Summary

ESBMC could not parse files that `#include <threads.h>` because the system header references glibc-internal types (`__tss_t`, `__thrd_t`, ...) that the bundled `bits/thread-shared-types.h` shim never satisfied.

This PR ships a simplified `threads.h` that shadows the system header and layers the C11 thread/mutex/cond/tss/once API directly onto ESBMC's existing pthread operational model — mirroring how glibc itself implements C11 threads over pthreads. Each C11 type is `typedef`'d straight onto its pthread counterpart and each call forwards with a return-code remap. The only non-trivial bit is `thrd_create`: the C11 start routine returns `int` while pthread's returns `void *`, so a small per-thread trampoline bridges the signature.

Three regression tests are added under `regression/esbmc-unix/`:
- `c11_threads_parse` — minimal reproducer from the issue
- `c11_threads_basic` — passing: mutex-protected shared counter
- `c11_threads_data_race` — failing: unsynchronized writer

Fixes #3449.


Yiannis128:
## Benchexec

* master https://github.com/esbmc/esbmc/actions/runs/25369863853
* feature https://github.com/esbmc/esbmc/actions/runs/25369832582



  correct:           19995
    correct true:    13751
    correct false:    6244
  incorrect:            22
    incorrect true:      5
    incorrect false:    17
  unknown:           16583
  Score:             33314 (max: 61270)


  correct:           19989
    correct true:    13751
    correct false:    6238
  incorrect:            21
    incorrect true:      5
    incorrect false:    16
  unknown:           16590
  Score:             33324 (max: 61270)

Yiannis128:
## Benchexec

* master https://github.com/esbmc/esbmc/actions/runs/25450329581
* threads https://github.com/esbmc/esbmc/actions/runs/25450374724

Yiannis128:
* master https://github.com/esbmc/esbmc/actions/runs/25575144746
* threads.h https://github.com/esbmc/esbmc/actions/runs/25575156176

Yiannis128:
## Benchexec Results

* master https://github.com/esbmc/esbmc/actions/runs/26456819468
* https://github.com/esbmc/esbmc/actions/runs/26456859080

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+
+int thrd_create(thrd_t *thr, thrd_start_t func, void *arg)
+{
+__ESBMC_HIDE:;
+  struct __c11_thrd_thunk_data *t = malloc(sizeof *t);
+  if (!t)
+    return thrd_nomem;
+  t->func = func;
+  t->arg = arg;
+  pthread_create(thr, NULL, __c11_thrd_thunk, t);
```
lucasccordeiro, line 35: `pthread_create`'s result is discarded, so a failed spawn still returns `thrd_success` and leaks `t`. Check it and return `thrd_error`.

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+  pthread_mutex_destroy(m);
+}
+
+void call_once(once_flag *flag, void (*func)(void))
+{
+__ESBMC_HIDE:;
+  __ESBMC_atomic_begin();
+  if (flag->__data == 0)
+  {
+    flag->__data = 1;
```
lucasccordeiro, line 140: The flag is set and the atomic block left before `func()` runs, so a concurrent caller returns while initialisation is incomplete. C11 7.26.2.1 requires it to block.

regression/esbmc-unix/c11_threads_data_race/main.c
```diff
@@ -0,0 +1,23 @@
+/* C11 threads without synchronization: ESBMC must catch the data race
```
lucasccordeiro, line 1: There is no data race here — the thread is joined before the assertion, so `shared` is deterministically 5. Rename the test and fix the comment.

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
+/* ESBMC operational model for ISO C11 <threads.h>: thin wrappers over the
```
lucasccordeiro, line 1: No test exercises `cnd_*`, `tss_*` or `mtx_trylock`; those wrappers ship unverified. Add regression coverage for them.

lucasccordeiro (changes requested):

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+{
+__ESBMC_HIDE:;
+  /* ESBMC explores all interleavings; yield is a no-op. */
+}
+
+int mtx_init(mtx_t *m, int type)
+{
+__ESBMC_HIDE:;
+  /* OM: mtx_recursive / mtx_timed not distinguished. */
+  (void)type;
```
rafaelsamenezes, line 97: The pthread OM does distinguish recursive mutexes (`pthread_mutexattr_settype`, `__ESBMC_mutex_relock`), so `mtx_recursive` here deadlocks a valid program. Map the type through an attr.

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+int mtx_lock(mtx_t *m)
+{
+__ESBMC_HIDE:;
+  return pthread_mutex_lock(m) == 0 ? thrd_success : thrd_error;
+}
+
+int mtx_timedlock(mtx_t *m, const struct timespec *time_point)
+{
+__ESBMC_HIDE:;
+  /* OM: timeout dropped, models the blocking branch only — sound
```
rafaelsamenezes, line 110: Dropping the timeout removes a behaviour, so this under-approximates rather than over-approximates: a path needing `thrd_timedout` is never explored. Return it nondeterministically.

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+__ESBMC_HIDE:;
+  /* OM: mtx_recursive / mtx_timed not distinguished. */
+  (void)type;
+  return pthread_mutex_init(m, NULL) == 0 ? thrd_success : thrd_error;
+}
+
+int mtx_lock(mtx_t *m)
+{
+__ESBMC_HIDE:;
+  return pthread_mutex_lock(m) == 0 ? thrd_success : thrd_error;
```
lucasccordeiro, line 104: The `_check`/`_noassert` remap in `clang_c_language.cpp` is applied to user TUs only, so the prebuilt clib always binds `_noassert` here and `--deadlock-check` is inert for `mtx_lock`, `thrd_join` and `cnd_wait`. Extend the remap to the library build.

src/c2goto/library/threads_lib.c
```diff
@@ -0,0 +1,209 @@
...
+{
+__ESBMC_HIDE:;
+  return pthread_setspecific(key, val) == 0 ? thrd_success : thrd_error;
+}
+
+void tss_delete(tss_t key)
+{
+__ESBMC_HIDE:;
+  /* ESBMC's pthread OM has no pthread_key_delete; teardown is a no-op. */
+  (void)key;
```
lucasccordeiro, line 208: `pthread_exec_key_destructors` still invokes the destructor `tss_create` registered, so a deleted key's destructor fires at thread exit. Clear `__ESBMC_thread_key_destructors[key]` here.

regression/esbmc-unix/c11_threads_data_race/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: Only `^VERIFICATION FAILED$` is pinned, so the test passes on any violation, not the intended one. Add a regex naming the expected assertion/property.

lucasccordeiro (commented):
Superseded: master already ships `src/c2goto/headers/threads.h`, `src/c2goto/library/threads_lib.c`, the CMakeLists entry and `regression/esbmc-unix/github_3449{,_fail,_join_fail}` from merged #6525. Rebase onto master and reduce this to the delta, or close.

PR #4292, opened 2026-05-05
[goto-cover] skip ASSUME in condition coverage (#4291)

### LINKED ISSUE #4291: [code-coverage] effect of assumes
Bug: ESBMC 8.2.0 instruments the comparison expressions inside __VERIFIER_assume() calls as condition-coverage claims. Since assumes are path constraints (not program logic), they should be excluded from coverage instrumentation. This inflates the condition count.

```C++
void __VERIFIER_assume(int cond);
  int  __VERIFIER_nondet_int(void);
 
  bool foo(int x) {
      if (x > 0 && x < 100) {
          return true;
      }
      return false;
  }
 
  int main(void) {
      __VERIFIER_assume(x >= -1000);
      __VERIFIER_assume(x <= 1000);
      foo(x);
      return 0;
  }
 ```

### PR DESCRIPTION
**Pull request by lucasccordeiro**
__VERIFIER_assume / __ESBMC_assume lower to ASSUME instructions and express path constraints, not program logic. Treating their guards as condition-coverage claims inflated Total Conditions (12 instead of 4 on the issue's reproducer). Skip ASSUME in condition_coverage(); ASSERT handling is unchanged.

Fixes #4291.

### DIFF
--- a/regression/goto-coverage/github_2071_1/test.desc
+++ b/regression/goto-coverage/github_2071_1/test.desc
@@ -1,7 +1,7 @@
 THOROUGH
 main.c
 --condition-coverage --k-induction --base-k-step 2 --no-standard-checks
-^Reached Conditions:  22
+^Reached Conditions:  16
 ^Short Circuited Conditions:  2
-^Total Conditions:  24
+^Total Conditions:  18
 ^VERIFICATION FAILED$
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by ChenfengWei0 — APPROVED** [review]
In Solidity, we have assume-like constructs, for example:
 ```solidity
 require(x > 0 && x < 100, "invalid x");
```

This is translated into an `assume` in the frontend. Therefore, this PR will affect coverage calculation for Solidity. However, that can be addressed in a separate PR.

PR #4300, opened 2026-05-05
OM Verification of `ctype.c`, `strings.c` and `string.c`

Yiannis128:
The following PR is heavily WIP but aims to add contracts for each OM in order to ensure that they're sound and complete. The idea is the following:

* Use the GNU `libc` prose to create contracts for each OM function
* Verify and find bugs
* Fix bugs 😉

At the very least, we are going to catch some edge cases.

Additionally, this PR will implement infrastructure for verifying ESBMC using ESBMC. A lot of the auto-generated files are from Byte Repair Contractor tool, which at some point will become public (before this is ready for review) and integrated into the CI. Currently it's in private beta but reach out if you have questions.

---

Status tracker issue: https://github.com/Yiannis128/esbmc/issues/1


src/c2goto/library/string.c
```diff
@@ -23,13 +23,23 @@
...
+  /* glibc forbids overlap between [dst, dst+strlen(src)+1) and src;
+       different objects covers the common case. */
+  __ESBMC_requires(__ESBMC_POINTER_OBJECT(dst) != __ESBMC_POINTER_OBJECT(src));
+  __ESBMC_ensures(__ESBMC_return_value == dst);
 __ESBMC_HIDE:;
   // Ensure src pointer is non-null
-  __ESBMC_assert(src != NULL, "Source pointer is null");
+  __ESBMC_assert(src != ((char *)0), "Source pointer is null");
 
+  __ESBMC_unroll(9);
```
rafaelsamenezes: This should not be a predefined number. The post condition is probably something on the line of `i >= 0 && i<len ==> src[i] == dst[i]`

rafaelsamenezes (changes requested):

lucasccordeiro:
@Yiannis128: Wonderful work! It was always my dream to use ESBMC to verify ESBMC :-)

lucasccordeiro:
@claude: can you please review this PR?

Yiannis128:
> @Yiannis128: Wonderful work! It was always my dream to use ESBMC to verify ESBMC :-)

Thank you, it's all possible thanks to Contractor and Scribe, both are going to be used in Byte Repair, and of course ESBMC underneath it all. Of course, due to the fact that the LLM is generating these contracts, they are correct but not complete, so we are going to go through an extra review step and ensure they're complete as well. The goal is to verify all the OM.

Yiannis128:
> **Claude finished @lucasccordeiro's task in 3m 29s** —— [View job](https://github.com/esbmc/esbmc/actions/runs/25445025437)
> 
> ................................................................................................................................
> 
> ### Summary
> Severity	Count	Key items
> Critical	3	`memmove` UB, missing copy-content postconditions, hardcoded unroll bounds
> High	4	Trivial `strcmp` ensures, missing ensures for 4 functions, weak `strncpy` precondition, no CI integration
> Medium	5	Weak harness postconditions for several functions
> Low	4	Style, redundant checks, no-op requires
> The `ctype.c` and `strings.c` contracts are in good shape. The main effort needed is in `string.c`: strengthening the copy-content postconditions, fixing the `memmove` direction check, and wiring the harnesses into the test suite.

Some were already noticed by @rafaelsamenezes , the Contractor implementation can find bugs, but unless the properties describing the contracts are complete, cannot prove the lack of bugs. Using an LLM to generate these from GNU libc prose is definitely a limitation

harness/isalnum/main.c
```diff
@@ -0,0 +1,20 @@
+// Auto-generated harness for: isalnum
```
lucasccordeiro, line 1: Auto-generated harnesses land in a new top-level `harness/` tree that no CMake or CTest target runs, and keep a commented-out placeholder assert. Move them under `regression/` with `test.desc`.

src/c2goto/library/ctype.c
```diff
@@ -15,58 +15,93 @@
 #undef tolower
 #undef toupper
 
-int isalnum(int c)
+__ESBMC_contract int isalnum(int c)
 {
+  __ESBMC_requires((c >= -1) && (c <= 255));
```
lucasccordeiro, line 20: `c >= -1` hardcodes `EOF == -1`, which C17 7.4p1 does not guarantee. Use `EOF`.

src/c2goto/library/ctype.c
```diff
@@ -15,58 +15,93 @@
...
+  __ESBMC_ensures(
```
lucasccordeiro, line 21: This postcondition uses numeric literals while other contracts in the same file use char constants. Use char constants throughout.

lucasccordeiro (changes requested):

src/c2goto/library/string.c
```diff
@@ -23,32 +23,55 @@
 #undef memmove
 #undef memchr
 
-char *strcpy(char *dst, const char *src)
+__ESBMC_contract char *strcpy(char *dst, const char *src)
 {
+  __ESBMC_requires(dst != ((void *)0));
+  __ESBMC_requires(src != ((void *)0));
+  __ESBMC_requires(__ESBMC_get_object_size(dst) >= 1);
```
rafaelsamenezes, line 30: `>= 1` does not bound the write: strcpy writes `strlen(src)+1` bytes, so a replaced call with an undersized dst passes. Require `strlen(src)+1`; same at 91 and 104.

src/c2goto/library/string.c
```diff
@@ -61,62 +84,126 @@ __ESBMC_HIDE:;
...
 
-int strcmp(const char *p1, const char *p2)
+__ESBMC_contract int strcmp(const char *p1, const char *p2)
 {
+  __ESBMC_requires(p1 != ((void *)0));
+  __ESBMC_requires(p2 != ((void *)0));
+  __ESBMC_requires(__ESBMC_get_object_size(p1) >= 1);
+  __ESBMC_requires(__ESBMC_get_object_size(p2) >= 1);
+  __ESBMC_ensures(
+    __ESBMC_return_value == 0 || __ESBMC_return_value < 0 ||
```
rafaelsamenezes, line 165: This postcondition holds for every int, so it constrains nothing. State the sign against the first differing byte.

src/c2goto/library/string.c
```diff
@@ -61,62 +84,126 @@ __ESBMC_HIDE:;
...
 
   return c1 - c2;
 }
 
-int strncmp(const char *s1, const char *s2, size_t n)
+__ESBMC_contract int strncmp(const char *s1, const char *s2, size_t n)
 {
+  __ESBMC_requires(s1 != ((void *)0));
+  __ESBMC_requires(s2 != ((void *)0));
+  __ESBMC_requires(__ESBMC_get_object_size(s1) >= n);
```
rafaelsamenezes, line 193: strncmp stops at the first NUL, so `strncmp("ab", "ab", 100)` on 3-byte arrays is legal C but fails this precondition. Bound the read by the strings, not n.

PR #4304, opened 2026-05-05
[cpp] throw std::bad_cast on dynamic_cast<T&> failure

zhihan:
## Summary

Restore the reference-form `dynamic_cast<T&>` throw that was reverted in #4229 due to typeinfo include-order issues.

The new `clang_cpp_convertert::ensure_bad_cast_symbol()` resolves `std::bad_cast` through Clang's hashed name-lookup tables (`DeclContext::lookup` with `IdentifierInfo`), force-translates it via `get_struct_union_class`, and caches the symbol-table entry for the rest of the TU. This makes the symbol available regardless of when (or whether) `<typeinfo>` was included relative to the `dynamic_cast<T&>` site — the original bug behind #4229.

If `<typeinfo>` wasn't included at all, the helper returns `nullptr` and the reference-form arm degrades to `assert(match)` so a mismatched cast still fails loudly with a message asking the user to include `<typeinfo>`, rather than silently succeeding.

This is the §5.7 follow-up referenced in `docs/dynamic_cast_rtti_plan.md` (the rest of the plan landed in #4147).

## Test plan

- [x] `regression/esbmc-cpp/inheritance/dynamic_cast3_bug` flipped `KNOWNBUG` → `CORE` (catches `bad_cast&`, asserts(0) inside → `VERIFICATION FAILED`)
- [x] `regression/esbmc-cpp/try_catch/nec_ex9-dynamic-cast` flipped `KNOWNBUG` → `CORE` (catches `bad_cast&`, gotos `--error-label ERROR` → `VERIFICATION FAILED`)
- [x] New `regression/esbmc-cpp/inheritance/dynamic_cast_ref_pass` covers the matching-type success path through the new throw machinery (`VERIFICATION SUCCESSFUL`)
- [x] Full `esbmc-cpp/inheritance/` suite (67/67) green
- [x] No regressions in a 110-test sweep across `inheritance`, `polymorphism_bringup`, `polymorphism_bringup_overload`, `try_catch` (two pre-existing `--std c++03` cstddef parse failures unrelated to dynamic_cast)

### DIFF
--- a/regression/esbmc-cpp/inheritance/dynamic_cast3_bug/test.desc
+++ b/regression/esbmc-cpp/inheritance/dynamic_cast3_bug/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
--- a/regression/esbmc-cpp/inheritance/dynamic_cast_ref_pass/main.cpp
+++ b/regression/esbmc-cpp/inheritance/dynamic_cast_ref_pass/main.cpp
@@ -1,0 +1,29 @@
+// Reference-form dynamic_cast on a matching type must succeed (no throw),
+// preserving the field value through the cast. Companion to dynamic_cast3_bug
+// which exercises the bad_cast throw direction.
+#include <cassert>
+#include <typeinfo>
+
+class Base
+{
+public:
+  virtual ~Base()
+  {
+  }
+};
+
+class Derived : public Base
+{
+public:
+  int x;
+};
+
+int main()
+{
+  Derived d;
+  d.x = 42;
+  Base &b = d;
+  Derived &rd = dynamic_cast<Derived &>(b);
+  assert(rd.x == 42);
+  return 0;
+}
--- a/regression/esbmc-cpp/inheritance/dynamic_cast_ref_pass/test.desc
+++ b/regression/esbmc-cpp/inheritance/dynamic_cast_ref_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 1 --no-unwinding-assertions --timeout 60
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/try_catch/nec_ex9-dynamic-cast/test.desc
+++ b/regression/esbmc-cpp/try_catch/nec_ex9-dynamic-cast/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 1 --no-unwinding-assertions --error-label ERROR --timeout 900
 
--- a/src/clang-c-frontend/clang_c_convert.h
+++ b/src/clang-c-frontend/clang_c_convert.h
@@ -99,7 +99,7 @@ protected:
   const clang::FunctionDecl *current_functionDecl;
 
   bool convert_builtin_types();
-  bool convert_top_level_decl();
+  virtual bool convert_top_level_decl();
 
   /**
    *  Since this class is inherited by clang-cpp-frontend,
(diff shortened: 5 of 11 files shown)


zhihan:
@XLiZHI  do you know what is the CI failure? I am on a Macbook, I am not sure what the build failure is.

XLiZHI:
> @XLiZHI do you know what is the CI failure? I am on a Macbook, I am not sure what the build failure is.

CI failed due to the following regression test:

```
The following tests FAILED:
	3716 - regression/esbmc-cpp17/cpp/github_3881_8 (Failed)
Errors while running CTest
```

XLiZHI:
Thank you for your PR, @zhihan.

Here are some of my thoughts regarding this implementation:

- First, regarding the throwing of the bad_cast exception. The previous implementation would fail due to the order of including <typeinfo>. I noticed that this PR has already attempted to address this issue using AST. This can solve the problem, but I don't really think it's an efficient solution. I tend to try to find this symbol during the symbolic execution stage, because by then we have already obtained all the symbols. https://github.com/esbmc/esbmc/blob/master/src/goto-symex/symex_catch.cpp
 
- The second point is that through testing, I discovered some issues with the `vptr_match_any()` function. I believe its logic is correct, but it seems not to be functioning properly in some TCs ([github_3881_8](https://github.com/XLiZHI/esbmc/blob/master/regression/esbmc-cpp17/cpp/github_3881_8/main.cpp)).

In conclusion, this is the reason why I temporarily disabled it in my PR https://github.com/esbmc/esbmc/pull/4229. In my understanding, "dynamic cast" doesn't seem to be something that can be resolved "statically". Symbolic execution might be a good place to assist in the implementation.

zhihan:
@XLiZHI Thanks for the review. Pushed 22d4624 to address both points:

**1. Move `std::bad_cast` resolution out of the per-cast hot path**

You're right that doing it per `dynamic_cast<T&>` site (even with caching) is not the right design. New approach: a single `pretranslate_bad_cast()` runs once per TU from an overridden `convert_top_level_decl`, *before* any user code is converted. It does one Clang hashed name-lookup (`std` → `bad_cast`) — not an AST walk — and force-translates the class via `get_struct_union_class`. By the time any `dynamic_cast<T&>` site is visited, `tag-std::bad_cast` is in the symbol table, regardless of include order. `build_dynamic_cast` then just does a plain `ns.lookup` (no AST work, no per-cast cost).

This also fixes a real include-order bug in the previous revision: a TU where `<typeinfo>` appeared in source order *after* an inline `dynamic_cast<T&>` site would silently fall back to `assert(match)` at the cast site even though `bad_cast` was eventually translated. With pretranslate running first, that case now correctly emits the throw.

If `<typeinfo>` isn't included at all, pretranslate is a no-op and `build_dynamic_cast` falls back to `assert(match)` with a comment pointing the user at the fix — so a mismatched cast still fails verification loudly instead of silently succeeding.

**2. `vptr_match_any()` failure on github_3881_8**

Root cause: `add_vtable_variable_symbols` populates `vtable_classes_per_vptr_` at the tail of `get_struct_class_virtual_methods`, *after* method bodies are converted. So an inline body that does `dynamic_cast<self&>(base_ref)` (like `IntCoord::assign`) couldn't find itself in the candidate-D set, and the OR-chain came up empty.

Fix: `pre_register_inherited_vtables` runs at the head of `get_struct_class_virtual_methods` and walks the non-virtual base chain transitively, registering the class against every ancestor whose vtable type symbol already exists (bases are processed before derived, so they always do). Dedup added in the tail registration since both paths now write the map.

`github_3881_8` passes, plus the full `inheritance` (82), `polymorphism_bringup_overload` (49), `try_catch` (81 excluding pre-existing c++03 cstddef failures), and `esbmc-cpp17/cpp` (17) suites all green.

XLiZHI:
Thank you for addressing my comment, @zhihan.

Based on this PR, I have made some improvements:

1. The reference branch of build_dynamic_cast no longer inlines a throw. Instead it emits a `code_function_callt` to the synthetic intrinsic `__ESBMC_throw_bad_cast`, wrapped in an `if (!match) { call }`.
2.  `symex_throw()` is refactored to handle rethrow and last_throw, then delegates to the extracted `symex_throw_dispatch(throw_code)` for the actual catch-stack matching logic.
3. New `symex_throw_bad_cast()`: on first call, resolves `tag-std::bad_cast` via `ns.lookup,` constructs a nondet operand and a `code_cpp_throw2tc`, and caches for reuse.

I’ve made some adjustments, so I need to request a third-party review.

lucasccordeiro (approved):
LGTM.

lucasccordeiro (approved):
LGTM.

Issue #4325, opened 2026-05-06
Path-coverage support: k-path and prime-path coverage on top of multi-property BMC


lucasccordeiro:
## Summary

ESBMC currently supports `--branch-coverage`, `--branch-function-coverage`, `--condition-coverage`, and `--assertion-coverage` (`src/goto-programs/goto_coverage.{h,cpp}`, `src/esbmc/options.cpp:672-709`), with results aggregated in `multi_property_check` and reported via `--cov-report-json` (`src/esbmc/bmc.cpp:920-978`). This issue proposes extending the same machinery with **path coverage**, in two phases.

The design has been informed by a literature survey; references at the end. **Edited after initial review** to address path-explosion memory overhead, witness-depth blow-up, and to clarify infeasibility scoring and function-scope semantics.

## Why

Branch coverage misses bugs that only manifest after a specific *combination* of branch outcomes, and it considers only loop bodies executed once or not at all. A formal treatment of k-bounded unrolling and its relation to coverage strength is given in Huang, Meyer & Weber (ICTSS 2025); empirical evidence that prime-path coverage strictly dominates edge-pair and edge coverage on mutant kill-rates is given in Durelli, Delamaro & Offutt (Sci. Comput. Program. 2018).

There is also direct architectural precedent for layering coverage on top of a BMC engine: FShell (Holzer, Schallhart, Tautschnig & Veith, CAV 2008) built exactly this on top of CBMC, and Angeletti et al. (J. Autom. Reason. 2010) used it in industrial practice to obtain the 100% code coverage required by EN 50128 / SIL 4 for ETCS railway software.

### Phase 1 — k-path coverage (`--k-path-coverage[=N]`)

The k-path criterion (Williams, Marre, Mouy & Roger, EDCC 2005, *PathCrawler*) requires every path with at most k consecutive loop iterations to be exercised. This maps naturally onto BMC's `--unwind k` — same intuition, established name.

- New method `goto_coveraget::k_path_coverage()` in `src/goto-programs/goto_coverage.cpp`, peer to `branch_coverage()` (`goto_coverage.cpp:223-285`).
- At each branching `IF guard GOTO L`, emit `assert(!prefix_witness)` where `prefix_witness` encodes the conjunction of the last ≤ k branch guards along the prefix.

#### Witness encoding (hybrid, not pure structural)

Pure structural AND-chains at every program point would balloon `expr2tc` trees and overwhelm the SMT solver at modest k. The encoding is therefore hybrid:

[...]

PR #4330, opened 2026-05-07
[goto-cover] --k-path-coverage Phase 1 (issue #4325)

lucasccordeiro:
Phase 1 of #4325. PathCrawler-style k-path coverage as a peer to `--branch-coverage`. Reuses `goto_coveraget`, `multi_property_check`, and the JSON report unchanged; adds four CLI knobs and one new instrumentation method.

## What

At each conditional branch `IF g GOTO L`, emit `assert(!witness)` goals for every combination of the last (n-1) prior branch directions × current direction. SAT marks the path reachable, mirroring `branch_coverage`'s falsification convention.

```
--k-path-coverage[=N]      N defaults to --unwind, falling back to 4
--k-path-coverage-claims   show reached claims
--k-path-witness-depth=D   post-simplification depth cap (default 8)
--k-path-max-goals=M       per-function goal cap (default 10000)
```

On goal-cap overflow, instrumentation aborts with an actionable error rather than truncating silently — decision per the issue's "Decided defaults" section.

## Phase-1 limitations (tracked in #4325 for follow-up PRs)

- Linear-walk prefix is an over-approximation when branches join (witnesses may reference variables mutated between branches). Phase 2 will move to a proper CFG analysis.
- No ghost-flag fallback when post-simplification depth > D — Phase 1 just drops the witness.
- No spanning-set scoring; infeasible witnesses count toward the denominator. Phase 2.

## Tests

Four regression tests under `regression/goto-coverage/k_path_cov_{1..4}/`:

| # | Scenario | Expected |
|---|---|---|
| 1 | Two independent `if`s | 6 witnesses, 6 reached, 100% |
| 2 | Loop body with one branch | 6 witnesses, 4 reached (loop-guard correlation) |
| 3 | Function call (per-function scope) | 4 witnesses, 4 reached, 100% |
| 4 | Correlated branches (infeasible prefixes) | 6 witnesses, 4 reached |

Existing branch / condition / assertion coverage tests still pass (sanity-checked locally).

## References

- Williams et al., *PathCrawler: Automatic Generation of Path Tests by Combining Static and Dynamic Analysis*, EDCC 2005, [doi:10.1007/11408901_21](https://doi.org/10.1007/11408901_21).
- Holzer et al., *FShell*, CAV 2008, [doi:10.1007/978-3-540-70545-1_20](https://doi.org/10.1007/978-3-540-70545-1_20).
- Huang, Meyer & Weber, *Loop Unrolling: Formal Definition and Application to Testing*, ICTSS 2025, [doi:10.1007/978-3-032-05188-2_2](https://doi.org/10.1007/978-3-032-05188-2_2).

Refs #4325.

### DIFF
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -714,8 +714,12 @@ void report_coverage(
   bool is_branch_func_cov =
     options.get_bool_option("branch-function-coverage") ||
     options.get_bool_option("branch-function-coverage-claims");
-  bool is_k_path_cov = options.get_bool_option("k-path-coverage") ||
-                       options.get_bool_option("k-path-coverage-claims");
+  // `k-path-coverage` itself stores the CLI integer N; the dedicated
+  // boolean enable flag is set by parseoptions when either CLI flag is
+  // present. This avoids `get_bool_option("k-path-coverage")` returning 0
+  // (false) for valid invocations like `--k-path-coverage` (no value) or
+  // `--k-path-coverage=0` (rejected at parse time, but defensive here).
+  bool is_k_path_cov = options.get_bool_option("k-path-coverage-enabled");
 
   if (is_assert_cov)
   {
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -714,6 +714,8 @@ void report_coverage(
   bool is_branch_func_cov =
     options.get_bool_option("branch-function-coverage") ||
     options.get_bool_option("branch-function-coverage-claims");
+  bool is_k_path_cov = options.get_bool_option("k-path-coverage") ||
+                       options.get_bool_option("k-path-coverage-claims");
 
   if (is_assert_cov)
   {
@@ -916,6 +918,24 @@ void report_coverage(
       log_result("Branch Coverage: N/A (no branches)");
   }
 
+  else if (is_k_path_cov)
+  {
+    const size_t total = goto_coveraget::total_kpath;
+    const size_t tracked_instance = reached_claims.size();
+    log_success("\n[Coverage]\n");
+    log_result("k-Path Witnesses : {}", total);
+    log_result("Reached : {}", tracked_instance);
+
+    if (options.get_bool_option("k-path-coverage-claims"))
+      for (const auto &claim : reached_claims)
+        log_status("  {}", prettify_solidity_expr(claim));
+
+    if (total != 0)
+      log_result("k-Path Coverage: {}%", tracked_instance * 100.0 / total);
+    else
+      log_result("k-Path Coverage: N/A (no k-path goals)");
+  }
+
   // Generate JSON coverage report
   if (options.get_bool_option("cov-report-json"))
   {
@@ -926,6 +946,8 @@ void report_coverage(
       cov_type = "branch";
     else if (is_branch_func_cov)
       cov_type = "branch-function";
+    else if (is_k_path_cov)
+      cov_type = "k-path";
     else if (is_cond_cov)
       cov_type = "condition";
     else if (is_assert_cov)
@@ -1703,7 +1725,8 @@ smt_convt::resultt bmct::multi_property_check(
                        &is,
                        &is_color,
                        &YELLOW,
-                       &runtime_solver](const size_t &i) {
+                       &runtime_solver](const size_t &i)
+  {
     //"multi-fail-fast n": stop after first n SATs found.
     if (is_fail_fast && fail_fast_cnt >= fail_fast_limit)
       return;
@@ -1771,9 +1794,9 @@ smt_convt::resultt bmct::multi_property_check(
     }
 
     // Store solver name initially but not again
-    std::call_once(summary.solver_name_flag, [&]() {
-      summary.solver_name = solver_ptr->solver_text();
-    });
+    std::call_once(
+      summary.solver_name_flag,
+      [&]() { summary.solver_name = solver_ptr->solver_text(); });
     // In coverage mode, only report instrumented coverage claims
     bool is_cov_silent =
       is_goto_cov && claim.claim_property != "instrumented assertion";
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -1662,6 +1662,11 @@ smt_convt::resultt bmct::multi_property_check(
   bool is_branch_func_cov =
     options.get_bool_option("branch-function-coverage") ||
     options.get_bool_option("branch-function-coverage-claims");
+  // "k-Path Cov" — keyed off the dedicated boolean (see line ~717
+  // comment); needed in the is_goto_cov disjunction so the
+  // claim_slicer reads the witness comment, matching the form stored
+  // in goto_coveraget::all_claims.
+  bool is_k_path_cov = options.get_bool_option("k-path-coverage-enabled");
 
   // is_vb: enable verbose output coverage info if the option "--verbosity coverage:N" is set, where N should larger than 0
   // By enabling this, we will output the coverage information when handling each instrumentation assertion.
@@ -1720,6 +1725,7 @@ smt_convt::resultt bmct::multi_property_check(
                        &is_vb,
                        &is_branch_cov,
                        &is_branch_func_cov,
+                       &is_k_path_cov,
                        &is_keep_verified,
                        &is_fail_fast,
                        &fail_fast_limit,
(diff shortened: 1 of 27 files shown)


lucasccordeiro:
Addressed all four Copilot review comments. Each fix has a regression test that would have caught the bug before merge.

**Comment 1 — silent report drop (CRITICAL).** Confirmed: `--k-path-coverage` (no `=N`) stored the boost `implicit_value` of `0`, which `optionst::get_bool_option` (== `atoi(value)`) treated as the feature being disabled, so `report_coverage` skipped the k-path block while instrumentation had already run. Fixed by storing a separate `k-path-coverage-enabled` boolean in the option_map and keying the report off that. Locked in by `regression/goto-coverage/k_path_cov_5_implicit_n`.

**Comment 2 — explicit non-positive values silently fell through.** `--k-path-coverage=0` and `--k-path-coverage=-1` previously bypassed validation and dropped to the `--unwind`/4 fallback. Switched the implicit_value sentinel to `INT_MIN` (unambiguous against any sane user input) and reject any other non-positive value at parse time. Locked in by `k_path_cov_6_zero_n`.

**Comment 3 — silent overflow at large N (CRITICAL).** Confirmed: `pdepth >= 64 ? 0 : (1 << pdepth)` skipped goal emission entirely for very large prefix depths, and the `branch_goals = 2 * pcombos` plus `function_goals + branch_goals > k_path_max_goals` arithmetic could overflow `size_t` in the `pdepth ≈ 62..63` band. Fixed by capping N at 30 in parseoptions (defense-in-depth re-check inside `k_path_coverage()`) and rewriting the goal-budget check in saturating form (`branch_goals > cap` first, then `function_goals > cap - branch_goals`). Locked in by `k_path_cov_7_n_too_large`.

**Comment 4 — misleading `--prime-path-coverage (Phase 2)` reference.** Removed; the error message now points only at currently-available mitigations.

Pushed as commit c2b48dd. All 7 regression tests pass; existing branch / condition / assertion coverage tests unaffected.

LukeW1999 (approved):
LGTM — Phase-1 scoping is clear and the goal-budget hardening from the Copilot round looks solid. 

Local-built a couple of examples and hit one semantic issue I don't think is covered by the limitations section; flagging it before merge.

**Metric inflation from chained contradictions in witness guards**

ESBMC's `simplify` recognises single-level 2-term contradictions (`p ∧ ¬p`) but doesn't fold chained forms like `p ∧ ¬p ∧ q` to `FALSE`. These witnesses pass the `is_false(full)` check at `goto_coverage.cpp:453` and get instrumented as assertions, but the emitted guard `¬(p ∧ ¬p ∧ X)` is a tautology — never falsified, never reached, permanently in the denominator.

This is distinct from the textual-order over-approximation already noted in the PR (which causes real reachable witnesses to get missed — a numerator-side effect). What I'm seeing is the **denominator inflating**: the reported coverage % comes out *lower* than the true path coverage.

**Minimal repro (self-correlated guards):**

```c
int main() {
  int a;
  if (a > 0); else;
  if (a > 0); else;
  if (a > 0); else;
}
```

```
$ esbmc repro.c --k-path-coverage=3 --unwind 2 --no-unwinding-assertions
k-Path Witnesses : 7
Reached          : 6
k-Path Coverage  : 85.71428571428571%
```

Every reachable path *is* covered — expected 100%. Dumping GOTO on the third branch shows:

```
ASSERT a <= 0 || a > 0    // comment: a > 0 && a <= 0
```

The conjunction `a > 0 && a <= 0` is self-contradictory and would ideally be dropped at instrumentation time.

**Switch impact (more user-visible):**

```c
int main() {
  int x, r;
  switch (x) {
  case 1: r = 10; break;
  case 2: r = 20; break;
  case 3: r = 30; break;
  default: r = 0;
  }
  return r;
}
```

```
$ esbmc switch.c --k-path-coverage=2 --unwind 2 --no-unwinding-assertions
k-Path Witnesses : 10
Reached          : 6
k-Path Coverage  : 60%
```

Switches get lowered to `IF x == c THEN GOTO L_case` chains in GOTO. Of the 10 witnesses, 2 are simplifier-missed `(x==1) ∧ (x==2)`-style contradictions (this issue), and 2 are the textual-order limitation already documented. Since switch is a common pattern, users may hit this systematically and see the reported metric come in lower than the true coverage.

Happy to defer to your judgement on whether this is in-scope for Phase 1 or belongs in the limitations section #4325 follow-up.

lucasccordeiro:
@LukeW1999 thanks for the careful read — addressed the structural case (your first repro) in commit f221b92.

**What changed.** Inside the mask loop, I now build the witness alongside a `(stored-guard, polarity)` accumulator. If a new prefix slot would assign opposing polarities to a guard already in the accumulator, the mask is skipped before any expression is built; the same check fires for the current branch direction against the accumulated prefix. Sound: only drops witnesses we can prove unsat from the syntactic structure, never adds phantom coverage.

**Verified on your repro.**

```
$ esbmc repro_chain.c --k-path-coverage=3 --unwind 2 --no-unwinding-assertions
k-Path Witnesses : 6
Reached          : 6
k-Path Coverage  : 100%
```

Down from 6/7 = 85.71% to 6/6 = 100% — the missing reachable witness was never missing; the denominator was inflated by one chain-contradictory tautology.

Locked in by `regression/goto-coverage/k_path_cov_8_chain_contradiction`.

**What's deferred.** The switch-case repro `(x == 1) ∧ (x == 2)` involves *semantically* contradictory comparisons over different stored expressions (`x == 1` and `x == 2` are not syntactic negations of each other), and folding them needs comparison-domain reasoning. That stays a Phase-1 limitation, now explicitly documented in the code comment, and is appropriate scope for the Phase-2 follow-up I tracked in #4325. Confirmed unchanged on your switch repro: still 6/10 = 60% — the precision regression you flagged is preserved as the Phase-1 known-bound, not silently masked.

All 8 `k_path_cov_*` regression tests and the existing branch / condition / assertion coverage tests pass on the rebased branch.

src/esbmc/bmc.cpp
```diff
@@ -714,6 +714,12 @@ void report_coverage(
   bool is_branch_func_cov =
     options.get_bool_option("branch-function-coverage") ||
     options.get_bool_option("branch-function-coverage-claims");
+  // `k-path-coverage` itself stores the CLI integer N; the dedicated
+  // boolean enable flag is set by parseoptions when either CLI flag is
+  // present. This avoids `get_bool_option("k-path-coverage")` returning 0
+  // (false) for valid invocations like `--k-path-coverage` (no value) or
+  // `--k-path-coverage=0` (rejected at parse time, but defensive here).
+  bool is_k_path_cov = options.get_bool_option("k-path-coverage-enabled");
```
prototypeC14, line 722: Shall we extend is_goto_cov at line 1582 to include is_k_path_cov? Otherwise, line 1745 stores claim.claim_cstr ("<witness> at <loc>") instead of claim_sig ("<witness>\t<loc>"), so the JSON lookup at line 820 always misses. As a result, every k-path claim ends up marked as uncovered, and summary.percentage = 0, even though cov_type = "k-path" is set.

lucasccordeiro:
@prototypeC14 thanks — confirmed the bug end-to-end and fixed it in commit e7a963b. You're right on every count.

**Verification of the bug.** Running `k_path_cov_1` with `--cov-report-json`:

| | textual | JSON (`summary`) |
|---|---|---|
| **Before fix** | `k-Path Coverage: 100%` | `{covered: 0, percentage: 0.0, total: 6, uncovered: 6}` |
| **After fix** | `k-Path Coverage: 100%` | `{covered: 6, percentage: 100.0, total: 6, uncovered: 0}` |

`claim_slicer` flips `claim_msg`'s source on `is_goto_cov`: in goto-cov modes it reads the comment (the witness text we stored at `insert_assert` time, matching the form in `goto_coveraget::all_claims`); otherwise it reads `from_expr(it->source.pc->guard)` — the negated assertion expression. Without `is_k_path_cov` in the disjunction, every k-path `claim_sig` recorded into `reached_claims` was the negated form, while the JSON emitter looked up the comment-based form from `all_claims`. They never matched, so every entry was marked uncovered.

**Fix** is a 3-line change in two places: the `is_goto_cov` disjunction in `print_coverage_summary`, and the same disjunction inside the `multi_property_check` `job_function` lambda (with `is_k_path_cov` added to the capture list). All 8 `k_path_cov_*` regression tests and the existing branch / condition / assertion / json coverage tests pass.

**Regression test caveat.** `testing_tool.py` only matches stdout regexes — there's no JSON-content matcher. The textual report is unchanged in both bugged and fixed states (it just counts `reached_claims.size()`), so a strict regression test for the JSON-only divergence isn't directly expressible in the existing framework. The 8 k-path tests + `cov_report_json_1` continue to be the smoke net. If you'd prefer, I can add JSON-content validation in a follow-up by extending `testing_tool.py` (out of scope for this PR).

lucasccordeiro:
Follow-up to address the test-framework gap that let the JSON-vs-textual divergence slip past the first review round: #4331 (extend testing_tool.py to validate cov-report.json and similar file outputs). Out of scope for this PR; tracked separately.

PR #4333, opened 2026-05-07
[clang-c-frontend] handle index on non-array struct base

mikhailramalho:
## Summary

When two TUs declare an external symbol with conflicting types — e.g. `extern int JJ[]` in one TU and `struct complete JJ` in another — clang's `ASTImporter` merges them into the more-complete decl (the struct), but the indexing reference `JJ[i]` from the first TU's body keeps its declared element type from the local `extern int JJ[]` view.

After typecheck, the `index_exprt` has a struct-typed `array_expr`, which `migrate_expr` then tries to lower into `index2tc(struct, i)` and aborts on `is_array_type(source) || is_vector_type(source)`. \`DebugOpt\` builds catch this; \`RelWithDebInfo\` silently falls through.

\`clang_c_adjust::adjust_index\` already rewrites the pointer case (\`p[i]\` → \`*(p+i)\`); this PR extends it to cover the struct/non-array case by emitting \`*((T*)&base + i)\` where \`T\` is the index expression's declared element type. The byte-level access goes through \`dereferencet\`'s well-tested path, which already handles arbitrary-typed storage.

Promotes \`regression/esbmc/github_1210-2-array\` from \`KNOWNBUG\` to \`CORE\` — the test now verifies successfully under \`DebugOpt\` as well.

## Test plan

- [x] \`regression/esbmc/github_1210-2-array\` passes under \`DebugOpt\`
- [ ] CI green on full regression suite

### DIFF
--- a/regression/esbmc/github_1210-2-array/test.desc
+++ b/regression/esbmc/github_1210-2-array/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 incomplete-array.c
 incomplete-lib.c
 ^VERIFICATION SUCCESSFUL$
--- a/src/clang-c-frontend/clang_c_adjust_expr.cpp
+++ b/src/clang-c-frontend/clang_c_adjust_expr.cpp
@@ -479,8 +479,7 @@ void clang_c_adjust::adjust_index(index_exprt &index)
     index.move_to_operands(addition);
     index.id("dereference");
   }
-  else if (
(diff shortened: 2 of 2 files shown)


mikhailramalho:
Code style CI seems broken

lucasccordeiro (approved):
I've just started another run for the CI.

lucasccordeiro:
PR / Check C/C++ code-style (pull_request) Successful in 1m

PR #4334, opened 2026-05-07
[C++ verification]  fix crash on range-based for over 2D array of struct

### LINKED ISSUE #4317: [C++] terminate called after throwing an instance of 'type2t::symbolic_type_excp'
```
#include <string>
#include <cassert>

int main()
{
    const std::string cases[][2] = {
        {"", ""},
    };

    for (const auto &t : cases)
    {
    }

    return 0;
}
```

XLiZHI:
fixes #4317 

fixup_renamed_type already guards against pointer subtypes that are symbol types, but did not handle the case where the subtype is an array type containing a symbol type (e.g. S[][1] where S is a struct). get_width() on such a nested type propagates symbolic_type_excp, which was uncaught and terminated the process.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_4317/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_4317/main.cpp
@@ -1,0 +1,12 @@
+#include <cassert>
+struct S
+{
+  int x;
+};
+int main()
+{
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

Issue #4377, opened 2026-05-08
[c++] Audit: tracking issue for C++17/20/23 frontend gaps with reproducers


lucasccordeiro:
## Summary

Tracking issue for ESBMC's gaps against modern C++ (C++17 / C++20 / C++23). Each item below is a confirmed missing feature with a minimal reproducer, suitable for splitting into individual issues as work is picked up.

This is the result of an audit on master @ `346082e0bf` (ESBMC 8.2.0). Every snippet below was run with:

```sh
esbmc <file.cpp> --std=<c++17|c++20|c++23> --incremental-bmc
```

against the bundled C++ operational models (no `--no-abstracted-cpp-includes`), which is the default mode users hit. Items where the failure mode is a *silent wrong result* are flagged inline; the rest are converter/parser rejections or missing-header failures.

---

### Structured bindings (decomposition declarations)

```cpp
#include <cassert>
#include <utility>
int main() {
  std::pair<int, int> p{1, 2};
  auto [a, b] = p;
  assert(a == 1 && b == 2);
  return 0;
}
```
→ `ERROR: unrecognized / unimplemented clang declaration Decomposition` / `CONVERSION ERROR`.

### `if` / `switch` with init-statement — silent wrong result

```cpp
#include <cassert>
int main() {
  if (int x = 7; true) {
    assert(x == 7);
  }
  return 0;
}
```
→ `VERIFICATION FAILED`. The variable declared in the init-statement of `if` (and equivalently `switch`) loses its initialiser; the counterexample shows `x` taking a non-deterministic value rather than `7`. This is a **silent wrong result** category bug — the program is well-formed C++17 but ESBMC reports a spurious assertion violation.

### Class template argument deduction (CTAD)

```cpp
template<typename T> struct Box { T t; };
int main() { Box b{42}; return b.t - 42; }
```
→ `error: no viable constructor or deduction guide for deduction of template arguments of 'Box'` / `PARSING ERROR`. ESBMC does not synthesise the implicit deduction guide that clang generates for aggregate templates in C++17 mode.

### Missing operational models: `<variant>`, `<any>`, `<filesystem>`

```cpp
#include <variant>
int main() { std::variant<int, double> v = 7; return 0; }
```
→ `fatal error: 'variant' file not found` / `PARSING ERROR`. Same for `#include <any>` and `#include <filesystem>`. None of these standard-library headers are bundled under `src/cpp/library/`.

---

### Concepts

```cpp
#include <type_traits>
[...]

PR #4388, opened 2026-05-09
[om] add C++23 <expected>

lucasccordeiro:
Add a C++23 `<expected>` operational model and a no-op converter case for CTAD deduction guides:

```cpp
#include <expected>
std::expected<int, int> f(bool b) {
  if (b) return 42;
  return std::unexpected<int>(-1);   // explicit template args; see caveat
}
```

* **`<expected>`** under `src/cpp/library/` — `std::expected<T,E>`, `std::unexpected<E>`, `in_place_t` / `unexpect_t`, plus `has_value` / `value` / `error` / `operator*` / `operator->` / `value_or`. Storage uses a discriminator + two members rather than a tagged union, which is far easier to lower and observably equivalent for verification (no real destructors or non-trivial member layouts). Out of scope: monadic `and_then` / `or_else` / `transform` and reference-typed `expected<T&,E>`.

* **`Decl::CXXDeductionGuide`** is now a no-op in `clang_c_convertert::get_decl()`. CTAD deduction guides have no runtime form; clang materialises specialisations through them at template-argument-deduction time. Without this, any header that ships a guide (the new `<expected>`, future `<variant>`/`<any>`, user code, …) failed conversion with `unrecognized / unimplemented clang declaration CXXDeductionGuide`.

**Caveat:** Calling `std::unexpected(-1)` without explicit template arguments still fails — clang's CTAD result lowers to a `DeducedTemplateSpecialization` type that the converter doesn't yet handle. That is a separate, larger item in #4377; this PR works around it by recommending `std::unexpected<E>(...)` (the form most stdlib code already uses internally).

Adds passing + failing tests under `regression/esbmc-cpp20/cpp/github_4377_expected{,_fail}/` covering struct error type, `value_or`, `operator->`. `esbmc-cpp17`/`esbmc-cpp20` suites all pass; `esbmc-cpp/cpp` baseline failures unchanged.

Picks off the `<expected>` item from the C++17/20/23 audit umbrella in #4377.

### DIFF
--- a/regression/esbmc-cpp20/cpp/github_4377_expected/main.cpp
+++ b/regression/esbmc-cpp20/cpp/github_4377_expected/main.cpp
@@ -1,0 +1,38 @@
+#include <cassert>
+#include <expected>
+
+struct Err
+{
+  int code;
+};
+
+std::expected<int, Err> compute(bool ok)
+{
+  if (ok)
+    return 100;
+  return std::unexpected<Err>(Err{-1});
+}
+
+int main()
+{
+  auto a = compute(true);
+  assert(a.has_value());
+  assert(*a == 100);
+  assert(a.value() == 100);
+
+  auto e = compute(false);
+  assert(!e.has_value());
+  assert(e.error().code == -1);
+
+  // value_or fallback.
+  std::expected<int, int> bad = std::unexpected<int>(7);
+  assert(bad.value_or(42) == 42);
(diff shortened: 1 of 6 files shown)


lucasccordeiro:
Heads-up: the `Decl::CXXDeductionGuide` no-op case here is also in #4389 (broader CTAD support). Whichever lands first gives the other a one-line trivial conflict. Otherwise independent.

PR #4389, opened 2026-05-09
[clang-c-frontend] support C++17/20 class template argument deduction

lucasccordeiro:
Add C++17/20 class template argument deduction (CTAD) support to the C frontend:

```cpp
template<typename T> struct Box { T t; };
Box b{42};                        // C++20 aggregate CTAD — was: PARSING ERROR
std::pair p{1, 2.5};              // ctor-based CTAD — was: CONVERSION ERROR
template<typename T> Holder(T) -> Holder<T>;
Holder h{7};                      // explicit guide — was: CONVERSION ERROR
```

Two pieces:

* `Decl::CXXDeductionGuide` handled as a no-op. CTAD deduction guides exist purely for template-argument deduction — clang materialises a concrete specialisation through them and they have no runtime form. The C frontend previously rejected them with `unrecognized / unimplemented clang declaration CXXDeductionGuide`, which broke any header carrying CTAD guides (libstdc++/libc++'s `pair`, the new `<expected>` (#4388), user code, …).

* `Type::DeducedTemplateSpecialization` forwarded to `DeducedType::getDeducedType()`. When a class-template name is written without explicit args, clang resolves it through deduction and stores the result as a `DeducedTemplateSpecialization` whose `getDeducedType()` is the concrete `RecordType`. Forward to that and the rest of `get_type()` handles it normally.

Together these cover aggregate CTAD (C++20 P1816), ctor-based CTAD (C++17), and explicit user-written guides ([over.match.class.deduct] in N4861).

Adds passing + failing tests under `regression/esbmc-cpp20/cpp/github_4377_ctad{,_fail}/`. `esbmc-cpp17`/`esbmc-cpp20` suites all pass; `esbmc-cpp/cpp` baseline failures unchanged.

Picks off the CTAD item from the C++17/20/23 audit umbrella in #4377.

### DIFF
--- a/regression/esbmc-cpp20/cpp/github_4377_ctad/main.cpp
+++ b/regression/esbmc-cpp20/cpp/github_4377_ctad/main.cpp
@@ -1,0 +1,44 @@
+#include <cassert>
+
+// Aggregate template (C++20 P1816 implicit aggregate guides).
+template <typename T>
+struct Box
+{
+  T t;
+};
+
+// Template with user-provided constructor.
+template <typename T>
+struct Wrap
+{
+  T t;
+  Wrap(T x) : t(x)
+  {
+  }
+};
+
+// Template with explicit deduction guide.
+template <typename T>
+struct Holder
+{
+  T value;
+};
+template <typename T>
+Holder(T) -> Holder<T>;
+
+int main()
+{
+  // Aggregate CTAD.
+  Box b{42};
+  assert(b.t == 42);
+
+  // Constructor-based CTAD.
+  Wrap w{99};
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
Heads-up for the reviewer: the `Decl::CXXDeductionGuide` no-op case here also appears in #4388 (`<expected>`). Whichever lands first will give the other a trivial conflict on those lines (delete the dup case). Otherwise the two PRs are independent — #4389 adds the type case and the wider CTAD support; #4388 adds the OM and exercises the same guide handling.

PR #4405, opened 2026-05-09
[clang-c-frontend] lower BinaryOperator::BO_Cmp (C++20 spaceship)

lucasccordeiro:
Lower the C++20 three-way comparison `a <=> b` for integer / pointer operands. Until now, any direct use of `<=>` failed conversion with `Conversion of unsupported clang binary operator: "<=>" to expression`.

```cpp
namespace std {
struct strong_ordering {
  signed char __value_;
  static const strong_ordering less, equal, greater;
};
inline constexpr strong_ordering strong_ordering::less{-1};
inline constexpr strong_ordering strong_ordering::equal{0};
inline constexpr strong_ordering strong_ordering::greater{1};
}

std::strong_ordering o = 5 <=> 3;   // .__value_ == 1 (greater)
std::strong_ordering p = 1 <=> 1;   // .__value_ == 0 (equal)
std::strong_ordering q = 1 <=> 5;   // .__value_ == -1 (less)
```

Lowering follows [expr.spaceship] in N4861: emit a nested conditional that produces the comparison-category struct value with the right signed-char tag (`-1`/`0`/`1`). `std::strong_ordering` / `weak_ordering` / `partial_ordering` all share a single signed-char first field, so the same construction works for all three; `t` is already resolved to the struct by `get_type()` by the time `BO_Cmp` is hit.

**Out of scope:**

- Floating-point spaceship. IEEE-754 NaN requires `partial_ordering::unordered`, which this lowering does not model yet. Direct float spaceship will miss the unordered case but otherwise behave correctly for ordered values.
- Defaulted `auto operator<=>(const T&) const = default;` on a user struct. Clang synthesises a body that compares members via literal-zero overloads (`CXXRewrittenBinaryOperator` + a `_CmpUnspecifiedParam` helper), which today trips a deeper symex assertion separate from this lowering. Pairs with #4386 (`<compare>` OM) — together they unblock the language feature; the symex follow-up is a separate piece of work.

Adds passing + failing regression tests under `regression/esbmc-cpp20/cpp/github_4377_spaceship{,_fail}/` covering less / equal / greater on integers and pointer ordering. `esbmc-cpp17` (27) and `esbmc-cpp20` (35+2) suites pass; `esbmc-cpp/cpp` baseline failures unchanged.

Picks off the `BO_Cmp` converter prerequisite from the C++17/20/23 audit umbrella in #4377.

### DIFF
--- a/regression/esbmc-cpp20/cpp/github_4377_spaceship/main.cpp
+++ b/regression/esbmc-cpp20/cpp/github_4377_spaceship/main.cpp
@@ -1,0 +1,38 @@
+#include <cassert>
+
+// Minimal in-source <compare> equivalent — exercises BinaryOperator's
+// BO_Cmp lowering without depending on a separate <compare> OM.
+namespace std
+{
+struct strong_ordering
+{
+  signed char __value_;
+  static const strong_ordering less;
+  static const strong_ordering equal;
+  static const strong_ordering greater;
+};
+inline constexpr strong_ordering strong_ordering::less{-1};
+inline constexpr strong_ordering strong_ordering::equal{0};
+inline constexpr strong_ordering strong_ordering::greater{1};
+} // namespace std
+
+int main()
+{
+  std::strong_ordering lt = 1 <=> 2;
+  assert(lt.__value_ < 0);
+
+  std::strong_ordering eq = 5 <=> 5;
+  assert(eq.__value_ == 0);
+
+  std::strong_ordering gt = 9 <=> 3;
+  assert(gt.__value_ > 0);
+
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
Superseded — closing in favour of an IREP2 `cmp_three_way` node + SMT lowering instead of the converter-side if/else shim. The shim works for the typical synthesised body (`lhs.x <=> rhs.x` is pure) but double-evaluates operands, loses the semantic anchor symex/interval/value_set could otherwise hook, and would have to be revisited anyway when we add the float / NaN→`partial_ordering::unordered` path. Will reopen as a fresh PR with the IR change; the regression tests carry over verbatim.

PR #4449, opened 2026-05-11
[cpp] fix explicit cast receiver in multiple inheritance

### LINKED ISSUE #3876: [c++] incorrect virtual dispatch with qualified base call under multiple inheritance

zhihan:
## Summary

Fixes pointer adjustment for explicit casts like `static_cast<B8*>(ptr)->eval()` in multiple inheritance scenarios.

## Problem

ESBMC failed to handle explicit casts wrapping the receiver of a virtual call through a non-first base class:

```cpp
Derived d;
Derived *ptr = &d;
assert(static_cast<B8 *>(ptr)->eval()); // VERIFICATION FAILED (incorrect)
```

The existing code only detected method receivers when the cast was directly the child of `MemberExpr`. However, explicit casts create wrapper nodes (`CXXStaticCastExpr`, etc.) in the AST, breaking the detection.

## Solution

Modified the `is_method_receiver` detection in `get_cast_expr` to walk through transparent wrapper expressions before checking for `MemberExpr -> CXXMemberCallExpr` pattern.

**Wrappers handled:**
- `ParenExpr`
- `ImplicitCastExpr`  
- `CXXStaticCastExpr`
- `CStyleCastExpr`
- `CXXFunctionalCastExpr`
- `CXXConstCastExpr`

## Tests Fixed

1. **`github_3876_static_cast`** (KNOWNBUG → CORE)
   - Now correctly verifies with `VERIFICATION SUCCESSFUL`
   - GOTO now shows proper pointer adjustment: `eval(( B8 *)((signed char *)ptr + 8))`

2. **`llbmc_multiple_inheritance-wrong`** (KNOWNBUG → CORE)
   - Was misclassified; actually produces `VERIFICATION SUCCESSFUL`
   - Updated expected output from `VERIFICATION FAILED` to `VERIFICATION SUCCESSFUL`

## Verification

All 82 inheritance tests pass:
```
100% tests passed, 0 tests failed out of 82
```

## Changes

- `src/clang-c-frontend/clang_c_convert.cpp`: Added wrapper-aware receiver detection
- `regression/esbmc-cpp/inheritance/github_3876_static_cast/test.desc`: KNOWNBUG → CORE
- `regression/esbmc-cpp/inheritance/llbmc_multiple_inheritance-wrong/test.desc`: KNOWNBUG → CORE, updated expectation

Fixes #3876 (static_cast variant)

### DIFF
--- a/regression/esbmc-cpp/inheritance/github_3876_static_cast/test.desc
+++ b/regression/esbmc-cpp/inheritance/github_3876_static_cast/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 10 --timeout 60
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/inheritance/llbmc_multiple_inheritance-wrong/test.desc
+++ b/regression/esbmc-cpp/inheritance/llbmc_multiple_inheritance-wrong/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 2 --no-unwinding-assertions --timeout 900
-^VERIFICATION FAILED$
+^VERIFICATION SUCCESSFUL$
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -3457,22 +3457,46 @@ bool clang_c_convertert::get_cast_expr(
     //     byte offset here would corrupt the symbolic model.
     //   - 'Base2 *o = new Derived()' — parent is not MemberExpr at all.
     //     Must remain unadjusted for ESBMC's delete model.
-    // NOTE: 'static_cast<B8*>(ptr)->eval()' is not yet handled correctly
-    // (separate issue tracked by the github_3876_static_cast KNOWNBUG test).
+    //
+    // The check walks through transparent wrapper expressions (casts, parens)
+    // to handle cases like 'static_cast<B8*>(ptr)->eval()'.
     bool is_method_receiver = false;
     {
-      auto parents = ASTContext->getParents(cast);
-      if (!parents.empty())
+      const clang::Stmt *node = &cast;
+      while (true)
       {
-        const clang::MemberExpr *me = parents.begin()->get<clang::MemberExpr>();
-        if (me)
+        auto parents = ASTContext->getParents(*node);
+        if (parents.empty())
+          break;
+
+        if (const auto *me = parents.begin()->get<clang::MemberExpr>())
         {
           auto grandparents = ASTContext->getParents(*me);
           if (
             !grandparents.empty() &&
             grandparents.begin()->get<clang::CXXMemberCallExpr>())
             is_method_receiver = true;
+          break;
         }
+
+        const clang::Stmt *parent_stmt = parents.begin()->get<clang::Stmt>();
+        if (!parent_stmt)
+          break;
+
+        // Walk through transparent wrappers
+        if (
+          llvm::isa<clang::ParenExpr>(parent_stmt) ||
+          llvm::isa<clang::ImplicitCastExpr>(parent_stmt) ||
+          llvm::isa<clang::CXXStaticCastExpr>(parent_stmt) ||
+          llvm::isa<clang::CStyleCastExpr>(parent_stmt) ||
+          llvm::isa<clang::CXXFunctionalCastExpr>(parent_stmt) ||
+          llvm::isa<clang::CXXConstCastExpr>(parent_stmt))
+        {
+          node = parent_stmt;
+          continue;
+        }
+
+        break;
       }
     }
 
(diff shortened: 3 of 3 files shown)


XLiZHI:
Hi @zhihan,

Thanks for your PR!

I'd suggest folding in a small cleanup before merge: replacing the six-way isa chain with `isa<CastExpr> || isa<ParenExpr>` (which also closes the latent `dynamic_cast` gap), collapsing the while(true) into a for loop, and merging did_adjust into do_typecast.

```diff
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -3461,46 +3461,28 @@ bool clang_c_convertert::get_cast_expr(
       // The check walks through transparent wrapper expressions (casts, parens)
       // to handle cases like 'static_cast<B8*>(ptr)->eval()'.
       bool is_method_receiver = false;
+      for (const clang::Stmt *node = &cast;;)
       {
-        const clang::Stmt *node = &cast;
-        while (true)
+        auto parents = ASTContext->getParents(*node);
+        if (parents.empty())
+          break;
+        const auto &parent = *parents.begin();
+        if (const auto *me = parent.get<clang::MemberExpr>())
         {
-          auto parents = ASTContext->getParents(*node);
-          if (parents.empty())
-            break;
-
-          if (const auto *me = parents.begin()->get<clang::MemberExpr>())
-          {
-            auto grandparents = ASTContext->getParents(*me);
-            if (
-              !grandparents.empty() &&
-              grandparents.begin()->get<clang::CXXMemberCallExpr>())
-              is_method_receiver = true;
-            break;
-          }
-
-          const clang::Stmt *parent_stmt = parents.begin()->get<clang::Stmt>();
-          if (!parent_stmt)
-            break;
-
-          // Walk through transparent wrappers
+          auto grandparents = ASTContext->getParents(*me);
           if (
-            llvm::isa<clang::ParenExpr>(parent_stmt) ||
-            llvm::isa<clang::ImplicitCastExpr>(parent_stmt) ||
-            llvm::isa<clang::CXXStaticCastExpr>(parent_stmt) ||
-            llvm::isa<clang::CStyleCastExpr>(parent_stmt) ||
-            llvm::isa<clang::CXXFunctionalCastExpr>(parent_stmt) ||
-            llvm::isa<clang::CXXConstCastExpr>(parent_stmt))
-          {
-            node = parent_stmt;
-            continue;
-          }
-
+            !grandparents.empty() &&
+            grandparents.begin()->get<clang::CXXMemberCallExpr>())
+            is_method_receiver = true;
           break;
         }
+        const clang::Stmt *ps = parent.get<clang::Stmt>();
+        if (!ps || !(llvm::isa<clang::CastExpr>(ps) || llvm::isa<clang::ParenExpr>(ps)))
+          break;
+        node = ps;
       }
 
-      bool did_adjust = false;
+      bool do_typecast = (cast.getCastKind() == clang::CK_DerivedToBase);
       if (
         adjust && total_offset > 0 && is_method_receiver &&
         expr.type().is_pointer())
@@ -3513,13 +3495,13 @@ bool clang_c_convertert::get_cast_expr(
         plus_exprt adjusted(expr, from_integer(total_offset, index_type()));
         adjusted.type() = char_ptr;
         expr = adjusted;
-        did_adjust = true;
+        do_typecast = true;
       }
 
       // Preserve original behaviour: CK_DerivedToBase always called gen_typecast;
       // CK_UncheckedDerivedToBase was a no-op (break) and should only typecast
       // when we actually applied a byte-offset adjustment above.
-      if (cast.getCastKind() == clang::CK_DerivedToBase || did_adjust)
+      if (do_typecast)
         gen_typecast(ns, expr, type);
 
       break;
```

XLiZHI (approved):
LGTM

Issue #4455, opened 2026-05-11
Reduce redundancy in CI


rafaelsamenezes:
Our current CI has some issues:

1. Optimize the flags. There are plenty of `incremental-bmc` or other incremental strategies that are there for no reason (i.e., the test is not about an incremental strategy). We could replace them with --unwind nr; we just need to find out the nr for the given regression.
2. We can probably make a smaller suite for macOS and Windows. For example, we don't need to retest everything that the Linux CI has already tested.
3. We really should refactor the full regression and try to remove duplicate regression tests. Please check the coverage information to decide which tests to reduce.
4. Coverage and Linux build do the same tests twice.

We can submit one PR for each item.

PR #4461, opened 2026-05-12
[build] slim macOS and Windows PR suites

lucasccordeiro:
Addresses #4455 item 2 — "We can probably make a smaller suite for macOS and Windows. For example, we don't need to retest everything that the Linux CI has already tested."

The full regression suite is exercised on every PR by `build-ubuntu`; re-running platform-portable tests on macOS and Windows mostly costs runner time without adding signal. Each non-Linux PR job now runs only the bits whose toolchain or platform behaviour materially differs.

**macOS PR (`build-unix.yml`)**: was `ctest -L python` which is a regex substring match and pulled in `python/`, `python-intensive/`, and `python-coverage/` (3298 tests). Now `ctest -L '^python/$'` — anchored to the `python/` suite only (3195 tests). The Python frontend is the one part of ESBMC whose toolchain differs across OSes (Homebrew Python, stdlib paths, `python3` resolution); the rest is platform-portable.

**Windows PR (`build-windows.yml`)**: was `ctest .` running the full suite (~7000 tests including C++ frontends, Python, Solidity, …). Now `ctest -L 'esbmc/|floats/'` (1325 tests) — the core C model checker plus the floating-point semantics suite, where IEEE-754 details and Clang-on-Windows behaviour can diverge. Genuinely Windows-specific bits (path handling, line endings, Clang frontend) live inside `esbmc/`.

push events still run the full suite on both runners — no observability regression on master. actionlint clean (only pre-existing SC2155/SC1001 warnings on unchanged lines).

Refs #4455.


LukeW1999 (approved):
LGTM

PR #4463, opened 2026-05-12
[irep2] remove dead code_init2t expression kind

lucasccordeiro:
Removes the `code_init2t` expression kind. The class was commented as
"a specialization of code_assign2t" but had zero producers anywhere in
the tree (no frontend, no `migrate.cpp` path, no constructor call), and
its single consumer in `value_set::apply_code` was a dead `else if`
that forwarded to `assign()` — identical to the `code_assign2t` arm
above it.

No functional change. Confirmed locally that the three `esbmc-cpp/`
test failures during my run reproduce on clean master at the same SHA,
so they are pre-existing and unrelated to this change.

Surfaced by the per-node README audit on PR #4458.


mikhailramalho (approved):
Lgtm

PR #4465, opened 2026-05-12
[solver] remove orphaned mk_bvnand / mk_bvnor across all backends

lucasccordeiro:
Removes the orphaned `mk_bvnand` and `mk_bvnor` virtual methods on
`smt_convt` and their per-backend overrides. After PR #4464 (which
removed the `bitnand2t` / `bitnor2t` IR kinds) these had zero callers
anywhere in the source tree.

Deleted:
- `smt_conv.{h,cpp}` base class declarations and `abort()` stubs.
- Override + implementation in: `bitwuzla`, `boolector`, `cvc4`,
  `cvc5`, `yices`, `z3`.
- Override + implementation in `smtlib`, plus the matching
  `SMT_FUNC_BVNAND` / `SMT_FUNC_BVNOR` enum entries and the
  positionally-aligned string-table rows in `smtlib_conv.cpp`
  (the latter contained a long-standing `"vnand"` typo).

No functional change: zero callers in the source tree before, zero
callers after. Local `ctest -L esbmc`: 92% passed, 350 failed — same
as the master baseline at the same SHA.

Follow-up to #4464.


mikhailramalho (commented):
lgtm

mikhailramalho (approved):

PR #4467, opened 2026-05-12
[goto-programs] Add goto_loop_simplify pass: erase / step-recognize / assume-rewrite empty-body loops

mikhailramalho:
## Summary

Adds a new goto-level pass `goto_loop_simplify` that simplifies loops with side-effect-free bodies *before* symex runs. Three rewrite paths fire on different shapes:

- **Path 0 (single-instruction self-loop)**: `A: IF cond GOTO A` / `A: GOTO A` rewritten to `ASSUME !cond` / `ASSUME false` in place. Mirrors `goto_loopst::find_function_loops`'s side-effect rewrite but runs unconditionally for plain BMC.
- **Path 1 (erase dead loops)**: empty body with only ASSIGN-to-local where every modified var dies via DEAD immediately after the loop — rewrites the loop head to `assume(exit_guard)` so multi-threaded busy-wait spinlocks (e.g. `while (shared_flag) {};`) keep their synchronization semantics instead of being silently erased.
- **Path 2 (step recognition)**: counter loops `for(i=INIT; REL(i,BOUND); i+=STEP)` with constant init/bound/step where `i` escapes — replaces the loop with `i = post_value` derived from the relation and step direction. Strictly precise: preserves `assert(i == N)` after `for(i=0;i<N;i++);` (havoc + assume(!cond) would only give `i >= N`).

The pass is gated on `--no-unwinding-assertions` or `--termination` — outside those modes the user-visible non-termination signal (unwinding assertion) needs the loop intact.

Additional symex changes (`db5e4a5623`) make `goto_symext::assume` populate the DPOR scheduler's `goto_state_map` / `pre_goto_guard` machinery the same way `symex_goto` does, so the assume-rewrite doesn't blow up the interleaving count on multi-threaded busy-waits (cf. `regression/esbmc-unix2/11_bakery.simple` — 345 vs ~891 interleavings without the fix).


mikhailramalho:
This PR changes how ESBMC models empty busy-wait loops.

Example:
```
  int ready = 0;
  int count = 0;

  void t1(void) {
    count = 1;
    ready = 1;
  }

  void t2(void) {
    while (!ready) {}
    assert(count == 1);
  }
```
Before this PR, the loop stayed as a bounded GOTO loop. With flags like --unwind 1 --no-unwinding-assertions, ESBMC could unwind the loop once and then continue after it without proving ready == 1. That admitted an artificial path:

ready == 0, count == 0
t2 enters while (!ready)
unwind limit is reached
t2 continues after the loop
assert(count == 1) fails

That path is not a real execution of the program, and was only reached because of the `--no-unwinding-assertion` flag.

After this PR, the empty loop is rewritten to its exit condition:
``` 
void t2(void) {
  __ESBMC_assume(ready);
  assert(count == 1);
}
```
Now any path where ready == 0 is pruned at the assume. The assertion is checked only on paths where ready == 1. Since t1 writes count = 1 before ready = 1, the surviving paths satisfy count == 1, so the previous false failure disappears.

This is the verdict change in 11_lu-fig4-complete.fixed.

mikhailramalho:
Tested locally:
```
=== top 30 branch worse (higher interleaves than master) ===
  +88                    0          88  esbmc-unix2_11_lu-fig4-complete.fixed
  +34                    0          34  esbmc-unix2_11_lu-fig4-complete

=== top 30 branch better (lower interleaves than master) ===
  -20886             21308         422  esbmc-unix2_github_548_fail2
  -4859               5005         146  esbmc-unix_00_race12
  -4859               5005         146  esbmc-unix_00_race11
  -3526               6019        2493  esbmc-unix_00_rwlock5
  -1705               2254         549  esbmc-unix2_github_586_1
  -1329               3091        1762  esbmc-unix2_01_pthread13
  -1152               3393        2241  esbmc-unix_02_phase_02
  -1024               1199         175  esbmc-unix2_11_spin2003-tid
  -702                 792          90  esbmc-unix2_01_pthread7
  -648                1425         777  esbmc-unix_10_bicycle_04
  -555                3644        3089  esbmc-unix2_11_szymanski
  -441                 942         501  esbmc-unix_03_circular_buffer_01
  -401                 462          61  esbmc-unix2_01_pthread36
  -304                 516         212  esbmc-unix2_11_read_write_lock
  -234                1119         885  esbmc-unix_03_atom002
  -197                 484         287  esbmc-unix2_01_pthread40
  -189                 372         183  esbmc-unix2_01_pthread20
  -169                 216          47  esbmc-unix2_11_simple-det
  -148                 344         196  esbmc-unix_00_race10
  -141                 581         440  esbmc-unix_github_1517_01
  -125                 598         473  esbmc-unix_github_1517
  -115                3566        3451  esbmc-unix_github_397_3
  -108                 617         509  esbmc-unix_02_stateful06_02
  -108                 240         132  esbmc-unix_00_rwlock6
  -94                  585         491  esbmc_19_time_var_mutex_true-unreach-call
  -90                  348         258  esbmc-unix_github_1517_04
  -75                  322         247  esbmc-unix_01_malloc_02
  -72                  424         352  esbmc-unix_02_stateful01_01
  -72                 1179        1107  esbmc-unix2_11_lu-fig2.fixed
  -72                  345         273  esbmc-unix2_01_pthread64

=== aggregate ===
  master total interleaves: 190863
  branch total interleaves: 145788
  delta:               -45075 (-23.62%)
  tests compared:      349
    regressed:         2
    improved:          66
    unchanged:         281
```
This branch reduces the number of interleavings considerably in our regression suite, and only increases in the cases where --no-unwinding-assertions were giving the wrong result.

lucasccordeiro (approved):
nice!

PR #4470, opened 2026-05-12
[build] respect ENABLE_<solver>=OFF when $HOME/<solver> exists

### LINKED ISSUE #121: CMake's ENABLE_<solver> OFF is ignored if ESBMC found the solver
If ESBMC found the solver (in my case they are located in $HOME), the backend are still compiled even if ENABLE_<solver> is set to OFF.

lucasccordeiro:
Per-solver CMakeLists previously force-enabled the backend whenever `$HOME/<solver>` was present, silently overriding an explicit `-DENABLE_<SOLVER>=OFF` on the command line. Drop the implicit HOME auto-enable across z3, boolector, cvc4, cvc5, mathsat, yices, and bitwuzla; the `HINTS` in `find_library`/`find_path` still pick up `$HOME/<solver>` when the user opts in.

Fixes #121


mikhailramalho (approved):
lgtm, but we should eventually get rid of these once camada integration lands

PR #4481, opened 2026-05-12
[ci] add clang sanitizer CI workflow (ASan + UBSan)

### LINKED ISSUE #223: Sanitizer errors
Currently, all clang sanitizers give errors when using ESBMC. They should be fixed:

- UndefinedBehaviourSanitizer
- AddressSanitizer
- ThreadSanitizer
- LeakSanitizer
- MemorySanitizer

lucasccordeiro:
Add a GitHub Actions workflow that builds ESBMC under clang's ASan and
UBSan, runs the CORE regression suite under each, deduplicates the
diagnostics into a job-summary table, and uploads the raw logs as a
30-day artifact. Bring-up uses continue-on-error while the suppression
files are populated.

Fixes #223

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,80 @@
+name: Sanitizers
+
+# Builds ESBMC under clang's AddressSanitizer and UndefinedBehaviorSanitizer
+# and runs the CORE regression suite, capturing every sanitizer diagnostic
+# to per-PID log files. The deduplicated findings are written to the job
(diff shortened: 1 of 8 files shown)


pmatos:
I think this is really useful for esbmc. We should run as many tools as we can to improve code quality and ensure no regressions. I would as a follow up do two things:
1. open a new issue to actually run these and fix the issues found
2. when they are green, turn this into a pr gate.

With more PRs coming in via agentic coding, the more gates we have at the PR level, the better the agentic loops we can create.

So, once merged I would complete the two above items.

PR #4484, opened 2026-05-13
[k-induction] deep-dive batch: interval bounds in IS, pointer handling, perf, cache fix, thread_local prep

mikhailramalho:
## Summary

Batch of improvements to ESBMC's k-induction pipeline, from a deep-dive of the code. Each commit is independent and reviewable on its own.

### 1. `[interval] thread loop bounds into k-induction inductive step`

Adds `instrument_loop_bounds_after_kind`: a second interval-analysis instrumentation pass that runs **after** `goto_k_induction`. Recomputes the interval fixpoint with `inductive_step_instruction`-marked instructions treated as transparent (so the havoc doesn't widen modified vars to their full type range), then inserts one `ASSUME(bounds)` between k-induction's preamble and each loop's exit-test. The inserted ASSUME is `inductive_step_instruction` so BC/FC skip it. Also wires `INTERVAL_INSTRUMENTATION_MODE::LOOP_MODE` for plain `--interval-analysis`; previously dead code.

Why it matters: interval analysis runs before `goto_k_induction`, so its bounds at the loop head are not asserted *after* the nondet havoc. IS therefore misses the tightening. With this change the bounds become a sound inductive hypothesis for IS.

### 2. `[k-induction] warn + disable IS for pointer-only loops`

When `--add-symex-value-sets` is enabled and a loop's only modified variables are pointers, `goto_k_induction` silently skipped the entire loop transformation (make_nondet_assign suppressed pointer havoc). Match the precedent for unsupported constructs (recursion, threads): log a warning identifying the loop, then flip `--disable-inductive-step`. (Superseded for pointer-only loops by commit #6 below, but the warning/disable mechanism stays for any future unsupported-construct case.)

### 3. `[k-induction] early-exit in contains_rec`

`contains_rec` used `res = recurse() || res`, walking all operands even after a match. Switched to check-this-node-first then short-circuit. Equivalent result (operand walk is read-only), but skips work on every match. Hot during `remove_unrelated_loop_cond`.

### 4. `[k-induction] memoize callee summaries in goto_loopst`

`get_modified_variables` descended into every callee body on every loop. Two loops in the same outer function calling the same helper re-walked it. Add a per-instance summary cache keyed by callee identifier holding (modified, unmodified) leaf-symbol sets. Mutual recursion is detected via an `in_progress` stack and falls back to legacy "do nothing on cycle"; partial summaries along a cycle-cut path are call-site dependent and are not cached.

### 5. `[k-induction] fix marked_branch cache: store guards and correct value`

Two bugs in `get_entry_cond_rec`'s per-branch cache:

a. The stored value was `false_branch ^ true_branch` — the XOR — but the recursion's actual return at that branch is `false_branch && true_branch`. A cache hit returned a different boolean than a fresh walk would.

b. Only the boolean was cached. The guards collected on the first visit were lost on every later cache hit, silently weakening the entry-condition assume.

Cache value changed to `(reaches, guards_to_merge)`; cache-hit reinjects the cached guards.

### 6. `[k-induction] sound pointer handling under --add-symex-value-sets`

Three coupled changes that let IS soundly cover loops whose modified variables include pointers:

a. `symex_dereference.cpp`: switch the value-set-derived constraint `SAME-OBJECT(p, &obj1+off1) || ...` from `claim()` to `assume()`. It only constrained the counterexample search before; assuming tightens the IS encoding so the havoc'd pointer is bounded to the value-set objects. The value-set analysis is sound, so the assume is safe.

b. `symex_dereference.cpp`: when one value-set element has unknown offset, emit the weaker `SAME-OBJECT(p, &obj)` disjunct (any offset) instead of dropping the entire disjunction. Keeps the assume non-trivial and stays sound.

c. `goto_k_induction.cpp`: drop the pointer-only-loop skip and the pointer-LHS skip in `make_nondet_assign`. The havoc + the new value-set assume now suffice. The warn/disable-IS path added in commit #2 becomes obsolete for pointer-only loops and is removed along with the `optionst&` plumbing through `goto_k_induction()` / `goto_termination()`. The unused `loopst::contains_only_pointers()` helper is dropped.

### 7. `[symex] make process-wide statics thread_local`

Preparation for a future thread-based `doit_k_induction_parallel`. Several process-wide statics (which raced when prototyping the threaded model under TSan) were made `thread_local`:

- `config.ssa_caching_db` (now `get_ssa_caching_db()` returning a `thread_local` reference) — even in sequential mode, removing it from `configt` is the right home for this cache.
- `value_sett::object_numbering` and `obj_numbering_refset`.
- `execution_statet::node_count` and `dynamic_counter`.
- `migrate_namespace_lookup`.

The threading itself is not in this PR — the goto-program `contextt` is shared between phases and would need a `clone()`-style refactor to be thread-safe. See follow-up #4520.

## Test plan

- [x] Builds on the default config (Ninja + Z3 + Bitwuzla).
- [x] New regression test `regression/k-induction/interval_loop_bounds/` covers the `--k-induction --interval-analysis` path.
- [x] All 177 `-L k-induction` tests pass after the full stack.
- [x] All 143 `-R interval` tests pass after the full stack.
- [x] Code-reviewed via the `code-reviewer` agent on commits #1 and #4; review fixes applied.
- [ ] CI runs the full suite.

## Related

- Follow-up issue: #4520 (thread-based parallel k-induction needs a thread-safe goto context).


mikhailramalho:
This PR:
```
tatistics:          36600 Files
  correct:           20196
    correct true:    13942
    correct false:    6254
  incorrect:            19
    incorrect true:      6
    incorrect false:    13
  unknown:           16385
  Score:             33738 (max: 61270)
``` 
master:
```
  correct:           20099
    correct true:    13846
    correct false:    6253
  incorrect:            19
    incorrect true:      5
    incorrect false:    14
  unknown:           16482
  Score:             33561 (max: 61270)
```

PR #4488, opened 2026-05-13
[esbmc] map property violations to CWE ids and add SARIF output

### LINKED ISSUE #1854: Map property violations to CWEs in ESBMC
Hi guys,

Do you see some benefit from mapping the property violations found by ESBMC to the Common Weaknesses Enumeration?

As an example, we have two papers that we perform this mapping:

(1) https://dl.acm.org/doi/10.1145/3617555.3617874
(2) https://arxiv.org/abs/2311.05281

Best,
Lucas

lucasccordeiro:
## Summary
Annotates every reported violation with the matching MITRE CWE identifier(s), pinned to CWE 4.20 and filtered to `ALLOWED` / `ALLOWED-WITH-REVIEW` mapping status. CWE ids appear in the textual counterexample, the JSON trace, the GraphML witness, and a new SARIF 2.1.0 backend selected with `--sarif-output {path|-}`. The freeform violation strings are unchanged, so the SV-COMP wrapper still picks the right `FALSE_*` category.

Mapping table and rationale (including ids dropped vs. the source paper) are documented in `docs/cwe-mapping.md`. Starting point: Table 4 of arxiv:2311.05281; ids whose Vulnerability Mapping Usage is `Discouraged` (119, 788, 690, 20, 682, 755, 664) or `Prohibited` (391) in CWE 4.20 are dropped.

Fixes #1854

## Test plan
- [x] `unit/util/cwe_mapping.test.cpp` — 52 Catch2 assertions over the substring → CWE table, `format_cwe_list`, `cwe_name`, and an id-coverage cross-check (passes)
- [x] `regression/esbmc/cwe_null_deref` — NULL deref → matches `CWE-476` (passes)
- [x] `regression/esbmc/cwe_overflow` — signed-int overflow → matches `CWE-190, CWE-191` (passes)
- [x] `regression/esbmc/cwe_sarif` — verifies the SARIF document carries `version: 2.1.0`, the `null-pointer-dereference` ruleId, and `taxonomies[].name = "CWE"`, `version = "4.20"`, `id = "476"` (passes)
- [x] Focused subset of 20 dereference / overflow / div / bounds regressions still pass
- [x] SV-COMP wrapper smoke: `esbmc-wrapper.py` still returns `FALSE_DEREF` on a NULL-deref benchmark


lucasccordeiro:
Addressed Copilot's five review comments in d54f5e1:

1. **SARIF `reportingDescriptor.name` violates `simpleName` (§3.49.7).** Dropped `name` entirely from both `rules[]` and `taxonomies[].taxa[]`. The human-readable text now lives only in `shortDescription`; the stable identifier is `id`. SARIF validators (including Microsoft's reference one) will no longer reject the document.
2. **`std::stoul` could throw on malformed line strings.** Replaced with a non-throwing `std::from_chars` parse in a small `parse_line` helper that returns `0` on empty / non-numeric input. SARIF emission is now exception-safe.
3. **Duplicate substring → rule table across `cwe_mapping.cpp` and `sarif.cpp`.** Consolidated into a single source of truth in `util/cwe_mapping`. The header now exposes `struct cwe_rule_t { sarif_id, short_description, cwes }` and `cwe_rule_for(comment)`. `sarif.cpp` reads from that — drift between the two tables is now impossible by construction.
4. **`names()` binary-search assumed a hand-sorted vector.** Switched to `std::map<unsigned, std::string_view>`. Order-of-declaration mistakes can no longer cause silent lookup failures.
5. **Rule substring matching assumed manual longest-first ordering.** The table is now sorted by descending `strlen(substring)` at static-init time via `std::stable_sort`. Adding a new entry anywhere in the source is safe.

New unit test (`every SARIF rule id is a valid simpleName`) walks every documented violation comment and verifies the emitted `sarif_id` matches the SARIF `simpleName` character class. All 14 unit cases (538 assertions) and 16 CWE-related ctests pass; a focused 20-test subset of dereference / overflow / bounds regressions remains green.

PR #4501, opened 2026-05-13
[esbmc] detect uninitialised local scalar reads as CWE-457

### LINKED ISSUE #4491: [esbmc] add uninitialised-variable checker mapped to CWE-457
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). The current pointer-validity checker maps a "dereference failure: invalid pointer" to CWEs {416, 822, 824, 908} — covering the *resource* angle (CWE-908 *Use of Uninitialized Resource*) when the pointer's target object is uninitialised, but not the more specific CWE-457 *Use of Uninitialized Variable* for scalar local reads.

### CWE-457 — Use of Uninitialized Variable

[CWE-457](https://cwe.mitre.org/data/definitions/457.html) (`ALLOWED` for vulnerability mapping in CWE 4.20) is the C semantics violation of reading an automatic-storage variable before any write to it. Examples:

```c
int x;
if (cond)
  x = 1;
return x; // CWE-457 when cond is false
```

ESBMC already tracks SSA-level value definedness and could in principle prove or refute this property; what's missing is a checker that surfaces the finding under a stable comment string.

### Proposal

1. Add a checker pass (likely under `src/goto-programs/`) that inserts an `__ESBMC_assert(<defined>, "use of uninitialized variable: <name>")` immediately before every read of a stack scalar that has no preceding write on the current path. Gate behind a new `--uninitialised-vars-check` flag in `src/esbmc/options.cpp`.
2. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping `"use of uninitialized variable"` to `{ "uninitialised-variable", "Use of uninitialized variable", {457} }`.
3. Add the CWE-457 name to `cwe_name` in `src/util/cwe_mapping.cpp` ("Use of Uninitialized Variable").
4. Add a unit test in `unit/util/cwe_mapping.test.cpp` for the substring → 457 mapping.
5. Add a regression test `regression/esbmc/cwe_uninit_var/` with the canonical conditional-init example, expecting `VERIFICATION FAILED` and `CWE-457`. Add a `_pass/` companion that initialises on every path.
6. Update `docs/cwe-mapping.md` (table + count).

### Acceptance criteria

- The new checker produces a sound verdict on the regression set (SAT for the failing case, UNSAT for the passing case).
- The CWE-457 line appears in textual / JSON / GraphML / SARIF output as for the other violations.
- Existing checkers and SV-COMP wrapper output are unaffected when the flag is off (flag defaults to off).

lucasccordeiro:
## Summary
Adds a new GOTO instrumentation pass `goto_check_uninit_vars`, gated behind `--uninitialised-vars-check`, that detects reads of uninitialised automatic-storage local scalars and surfaces them as **CWE-457** *Use of Uninitialized Variable* (`ALLOWED` in CWE 4.20).

For every tracked scalar local `x`, the pass introduces a fresh shadow boolean `__ESBMC_defined$<func>$<n>` that:
- starts `false` at the `DECL` site,
- flips to `true` after every direct write,
- is asserted `true` before every direct read,
- is conservatively flipped to `true` when `&x` is taken (we cannot soundly track writes through pointer aliases without a full alias analysis).

The pass runs **before** `mark_decl_as_non_det`, which would otherwise rewrite every uninitialised `DECL` into a `DECL; ASSIGN nondet` and erase the property.

Fixes #4491

## Test plan
- [x] New unit test cases (`cwe_for matches uninitialised variable`, `cwe_name(457)`, expanded validity matrices) — 15 cases / 564 assertions pass.
- [x] New regression `regression/esbmc/cwe_uninit_var` — conditional init → FAIL with `CWE-457`.
- [x] New regression `regression/esbmc/cwe_uninit_var_pass` — both branches init → SUCCESSFUL.
- [x] New regression `regression/esbmc/cwe_uninit_var_funcall` — `int x; x = foo(); return x;` → SUCCESSFUL (pins the function-call-init false-positive fix from review).
- [x] Existing 31-test regression subset still passes with flag off (no behaviour change by default).
- [x] Manual soundness sample with flag on:
  - `int x; x = x + 1; return x;` → FAIL (self-assign RHS read).
  - `int i; int a[10]; return a[i];` → FAIL on `i` (read inside `&a[i]`-style indexing).
  - `int x; int *p = &x; *p = 5; return x;` → SUCCESSFUL (address-of-then-write through alias).
  - `int x; for (int i = 0; i < 3; i++) x = i; return x;` → SUCCESSFUL.
- [x] code-reviewer flagged two real bugs (FUNCTION_CALL return treated as read, missed reads inside `&a[i]` index); both fixed in this commit, with pinned regression tests.

## Out of scope
- Aggregate types (struct/union/array elements).
- Function parameters (frontend models them as already initialised).
- Pointers — already covered under CWE-908.
- Aliased writes via tracked address-of (would need a Steensgaard/Andersen pass).

### DIFF
--- a/src/esbmc/options.cpp
+++ b/src/esbmc/options.cpp
@@ -581,6 +581,9 @@ const struct group_opt_templ all_cmd_options[] = {
      "thread interleaving"},
     {"lock-order-check", NULL, "Enable for lock acquisition ordering check"},
     {"atomicity-check", NULL, "Enable atomicity check at visible assignments"},
+    {"uninitialised-vars-check",
+     NULL,
+     "Enable check for reads of uninitialised local variables (CWE-457)"},
     {"volatile-check", NULL, "Enable check for volatile variable"},
     {"stack-limit",
      boost::program_options::value<int>()->default_value(-1)->value_name(
--- a/regression/esbmc/cwe_uninit_var/main.c
+++ b/regression/esbmc/cwe_uninit_var/main.c
@@ -1,0 +1,9 @@
+extern int __VERIFIER_nondet_int(void);
+
+int main(void)
+{
+  int x;
+  if (__VERIFIER_nondet_int())
+    x = 1;
+  return x;
+}
--- a/regression/esbmc/cwe_uninit_var/test.desc
+++ b/regression/esbmc/cwe_uninit_var/test.desc
@@ -1,0 +1,6 @@
+CORE
+main.c
+--uninitialised-vars-check --incremental-bmc
+^VERIFICATION FAILED$
+^  use of uninitialized variable: x$
+^  CWE: CWE-457$
--- a/regression/esbmc/cwe_uninit_var_funcall/main.c
+++ b/regression/esbmc/cwe_uninit_var_funcall/main.c
@@ -1,0 +1,8 @@
+extern int foo(void);
+
+int main(void)
+{
+  int x;
+  x = foo();
+  return x;
+}
--- a/regression/esbmc/cwe_uninit_var_funcall/test.desc
+++ b/regression/esbmc/cwe_uninit_var_funcall/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--uninitialised-vars-check --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/cwe_uninit_var_pass/main.c
+++ b/regression/esbmc/cwe_uninit_var_pass/main.c
@@ -1,0 +1,11 @@
+extern int __VERIFIER_nondet_int(void);
+
+int main(void)
+{
+  int x;
+  if (__VERIFIER_nondet_int())
+    x = 1;
+  else
+    x = 2;
+  return x;
+}
--- a/regression/esbmc/cwe_uninit_var_pass/test.desc
+++ b/regression/esbmc/cwe_uninit_var_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--uninitialised-vars-check --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -44,6 +44,7 @@ extern "C"
 #include <goto-programs/set_claims.h>
 #include <goto-programs/show_claims.h>
 #include <goto-programs/loop_unroll.h>
+#include <goto-programs/goto_check_uninit_vars.h>
 #include <goto-programs/mark_decl_as_non_det.h>
 #include <goto-programs/assign_params_as_non_det.h>
 #include <goto2c/goto2c.h>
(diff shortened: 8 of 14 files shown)


lucasccordeiro:
Rebased onto master (resolves the cwe_mapping conflicts with #4499 / #4500) and addressed Copilot's three review comments in `eecf89f`:

1. **`collect()` dropped `lvalue_ctx` on default recursion → false-positive risk on conditional lvalues.** Fixed: default recursion now preserves `lvalue_ctx`, and `dereference2t` in l-value mode is handled explicitly (the pointer operand is an r-value read of the pointer, never the l-value itself). This prevents the base symbol of a wrapped lvalue (e.g. `&(c ? a : b)`) from being treated as a value-read. Soundness comes first: for a formal verifier, suppressing a true read is acceptable; producing a false positive is not.

2. **O(#reads × #locals) display-name lookup.** Replaced the linear scan over `shadows` with a `std::map<irep_idt, std::string> shadow_display` populated alongside the shadow at DECL time; lookup is now O(log #locals) per read.

3. **Header docstring naming mismatch.** Updated the docstring in `goto_check_uninit_vars.h` to use the actual on-wire shadow-symbol pattern `__ESBMC_defined$<function>$<n>`.

Verification on the original soundness sample (all behaviours preserved):
- `int x; x = foo(); return x;` → SUCCESSFUL.
- `int i; int a[10]; return a[i];` → FAIL on `i` (`CWE-457`).
- `int x; x = x + 1; return x;` → FAIL on `x`.
- `int x; int *p = &x; *p = 5; return x;` → SUCCESSFUL.
- `int x; for (int i = 0; i < 3; i++) x = i; return x;` → SUCCESSFUL.

Tests: 16 unit cases / 577 assertions pass; 6 CWE regression tests pass.

Yiannis128:
Hi, thanks for this PR. I am reviewing it now.

Yiannis128 (approved):
LGTM

src/esbmc/options.cpp
```diff
@@ -581,6 +581,11 @@ const struct group_opt_templ all_cmd_options[] = {
      "thread interleaving"},
     {"lock-order-check", NULL, "Enable for lock acquisition ordering check"},
     {"atomicity-check", NULL, "Enable atomicity check at visible assignments"},
+    {"uninitialised-vars-check",
+     NULL,
+     "Detect reads of uninitialised automatic-storage local scalars (CWE-457). "
```
Yiannis128: My only nit-pick. This description should be simplified. Users won't understand it as it describes _how_ something is done but not _what_ it does.

Yiannis128:
@lucasccordeiro for the description of the flag in options.cpp, maybe something like _"Detect reads of uninitialized scalar local variables and report them (as CWE-457, pointers are reported as CWE-908). Struct, union, and array members are not checked."_. It's more readable imo.

src/esbmc/options.cpp
```diff
@@ -581,6 +581,9 @@ const struct group_opt_templ all_cmd_options[] = {
      "thread interleaving"},
     {"lock-order-check", NULL, "Enable for lock acquisition ordering check"},
     {"atomicity-check", NULL, "Enable atomicity check at visible assignments"},
+    {"uninitialised-vars-check",
+     NULL,
+     "Enable check for reads of uninitialised local variables (CWE-457)"},
```
Yiannis128, line 586: I think this is a better description. Please let me know if you disagree: _"Detect reads of uninitialized scalar local variables (and report them as CWE-457, pointers are reported as CWE-908). Struct, union, and array members are not checked."_

Yiannis128 (changes requested):

PR #4505, opened 2026-05-13
[witness] add violation witness validation

XLiZHI:
(--validate-violation-witness) 

Implements witness 2.0 violation witness validation. When --validate-violation-witness --witness <file> is passed, ESBMC parses the YAML witness and uses its waypoints to guide symbolic execution: 

- Branching waypoints constrain the guard at matching if/while branch points, forcing execution down the witnessed path. 

- Assumption waypoints inject the given constraint as an `assume()` at the specified source location. 

- Target waypoint identifies the expected error location; assertions at all other locations are neutralised (replaced with true) so only the witnessed target can fire. The parser extracts waypoints from the content field into a per-segment structure. During symbolic execution the engine advances a cursor through the witness segments, and in claim() non-target assertions are suppressed unless their source line matches the target waypoint.


lucasccordeiro (approved):

PR #4508, opened 2026-05-13
[k-induction] perf: contains_rec early-exit + memoize callee summaries

mikhailramalho:
## Summary

Two small, independent perf fixes in the k-induction preprocessing pass:

1. **`contains_rec` early-exit** (`src/goto-programs/goto_k_induction.cpp`): the recursion previously walked every operand even after a match (`res = recurse() || res`). Now: check the node first (immediate return if it's a tracked symbol), then short-circuit the operand walk as soon as a sibling matches. Equivalent result (`loop_varst` is unordered, operands are read-only) but skips work on every match. Hot during `remove_unrelated_loop_cond`.

2. **Memoize callee summaries** (`src/goto-programs/goto_loops.cpp` + `.h`): `get_modified_variables` previously descended into every callee body on every loop. Two loops in the same outer function calling the same helper re-walked the helper from scratch. Add a per-instance summary cache keyed by callee identifier, holding `(modified_loop_vars, unmodified_loop_vars)`. Each callee body is walked at most once per `goto_loopst` instance; subsequent calls fold the cached set in. Mutual recursion is detected via an `in_progress` stack and falls back to the legacy "do nothing on cycle" behaviour — partial summaries along a cycle-cut path are call-site dependent and are not cached.

This is item #2 from the broader k-induction deep-dive.

## Test plan

- [x] All 176 `-L k-induction` tests pass.
- [x] All 142 `-R interval` tests pass.
- [x] Reviewed by code-reviewer agent (two minor follow-ups applied: preserved the legacy `abort()` for unknown-callee lookups, documented the per-instance cache scope).
- [ ] CI runs the full suite.

### DIFF
--- a/src/goto-programs/goto_k_induction.cpp
+++ b/src/goto-programs/goto_k_induction.cpp
@@ -260,17 +260,18 @@ void goto_k_inductiont::make_nondet_assign(
 
 static bool contains_rec(const expr2tc &expr, const loopst::loop_varst &vars)
 {
+  // Check this node first: if it's a tracked symbol, we're done.
+  if (is_symbol2t(expr) && vars.find(expr) != vars.end())
+    return true;
+
+  // Otherwise recurse into operands and stop at the first match.
   bool res = false;
   expr->foreach_operand([&vars, &res](const expr2tc &e) {
-    if (!is_nil_expr(e))
-      res = contains_rec(e, vars) || res;
-    return res;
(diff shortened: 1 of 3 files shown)


mikhailramalho:
Closing — these commits will be added on top of feat/interval-analysis-instrument-before-loop instead of starting a new PR.

PR #4550, opened 2026-05-15
[python] use bool xor (not bitxor) in floor-div sign correction

### LINKED ISSUE #4548: [bitwuzla] sort-mismatch encoding floor-div of nondet_int() in function-call argument
## Summary

Bitwuzla raises an internal sort-mismatch error during SMT encoding when a floor-division expression on a `nondet_int()` is passed as a function-call argument. The same program verifies SUCCESSFUL under `--z3`. 6-line minimal reproducer.

Surfaced while adding symbolic-shape harnesses to the NKI / ESBMC PoC ([lucasccordeiro/AWS-Neuron](https://github.com/lucasccordeiro/AWS-Neuron)) — multiple symbolic targets reduce to this shape (kernel computes `width // chunk` from a nondet input dimension and passes the result into a stub-library function).

## Environment

- ESBMC `main` at commit `088d2b424d`, built locally.
- macOS aarch64, default Bitwuzla 0.8.2.

## Minimal reproducer (6 lines)

```python
def check(c: int) -> None:
    assert c >= 0

a: int = nondet_int()
__ESBMC_assume(a >= 0)
check(a // 2)
```

Run with default solver:

```
$ esbmc bw.py
...
ERROR: Bitwuzla error encountered
bitwuzla: error:  invalid call to 'Term bitwuzla::TermManager::mk_term(Kind,
          const std::vector<Term> &, const std::vector<uint64_t> &)',
          term with unexpected sort at index 0
```

Run with Z3:

```
$ esbmc --z3 bw.py
...
VERIFICATION SUCCESSFUL
```

## Trigger boundary

I narrowed the trigger empirically against the same shell:

| Variant | Verdict |
|---|---|
| `check(a // 2)` | **ERROR** (Bitwuzla sort mismatch) |
| `check(a)` (no division) | SUCCESSFUL |
| `check(a * 2)` (mult instead of div) | FAILED (correct: `a * 2` can overflow; assertion fails on Bitwuzla — no encoding error) |
| Inline `c: int = a // 2; assert c >= 0` (no fn call) | SUCCESSFUL |
| Same as repro but `--z3` | SUCCESSFUL |

So the trigger is the conjunction of three properties:

1. The expression is **floor division** (`//`) of a `nondet_int()`.
2. The expression appears as a **function-call argument** (not a local assignment).
3. The solver is **Bitwuzla** (Z3 handles the same SMT term).

Drop any one and the bug disappears.

## Diagnosis

When the floor-div result is materialised at the call site, ESBMC's BV-encoder seems to hand Bitwuzla a term whose sort doesn't match what the function's parameter slot expects — possibly because the division-result type isn't reconciled with the parameter's declared `int` before address-of / temp materialisation. The inline path (`c = a // 2; assert ...`) uses the local-symbol storage and avoids whatever sort mismatch the call-arg path introduces.

Same end-stage diagnostic as #4539 / #4541's `make_member` family (rvalue → coerced parameter), but here the issue is in the BV-encoding for Bitwuzla specifically, after the GOTO/symex stages succeed.

## Impact

In the NKI PoC, the `avgpool_symbolic` target trips this:

```python
# harness/avgpool_symbolic.py
H: int = 6 + 2 * k_h           # symbolic
W: int = 6 + 2 * k_w           # symbolic
in_tensor: Tile3D = nl_ndarray_3d(C, H, W, ...)
out: Tile3D = tensor_avgpool_kernel(in_tensor, POOL)
```

The kernel internally computes `sz_hin // sz_pool` and passes the result through a chain of function calls into `tile3d_ap_5d(...)`, which assert-checks the resulting view. Bitwuzla errors at encoding; Z3 verifies in ~5s. The PoC's `verify.py` manifest now pins `--z3` for this one target as a workaround.

`attn_fwd_v3_symbolic` (a different kernel) trips the same Bitwuzla error from a different floor-div site, suggesting the pattern is general.

## Acceptance test (suggested)

The 6-line reproducer above. Once `esbmc bw.py` (default solver) produces `VERIFICATION SUCCESSFUL`, both NKI PoC symbolic harnesses revert from `--z3` to the default Bitwuzla.

## Related

- Not a follow-on to any existing issue; it's a fresh sort-mismatch in the BV-encoder when consuming a floor-div ESBMC IR term inside a function-call argument expression. Z3 isn't affected, which means the IR is well-formed and the lowering for Bitwuzla is missing a sort cast.

lucasccordeiro:
The Python floor-division lowering in `python_math::handle_floor_division` built the sign-difference correction `(a<0) ^ (b<0)` as `bitxor` over `bool_type()` operands. `bitxor` migrates to `bitxor2tc(type, side1, side2)` which carries the bool type into a BV-shaped term — Bitwuzla's encoder rejects it ("term with unexpected sort"). Z3 happens to be lenient and masked the bug, so the issue only surfaced when the floor-div result was materialised at a function-call argument site under default Bitwuzla. Switching the operator to `xor` (the canonical boolean op, asserted bool in `migrate.cpp`) matches the operand types and encodes cleanly on both solvers without changing semantics.

Fixes #4548

### DIFF
--- a/regression/python/github_4548_floordiv_call_arg-fail/main.py
+++ b/regression/python/github_4548_floordiv_call_arg-fail/main.py
@@ -1,0 +1,6 @@
+def check(c: int) -> None:
+    assert c >= 0
+
+a: int = nondet_int()
+__ESBMC_assume(a >= -3)
+check(a // 2)
--- a/regression/python/github_4548_floordiv_call_arg-fail/test.desc
+++ b/regression/python/github_4548_floordiv_call_arg-fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION FAILED$
--- a/regression/python/github_4548_floordiv_call_arg/main.py
+++ b/regression/python/github_4548_floordiv_call_arg/main.py
@@ -1,0 +1,6 @@
+def check(c: int) -> None:
+    assert c >= 0
+
+a: int = nondet_int()
+__ESBMC_assume(a >= 0)
+check(a // 2)
--- a/regression/python/github_4548_floordiv_call_arg/test.desc
+++ b/regression/python/github_4548_floordiv_call_arg/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_4548_floordiv_negative/main.py
+++ b/regression/python/github_4548_floordiv_negative/main.py
@@ -1,0 +1,6 @@
+def check(c: int) -> None:
+    assert c == -3
+
+a: int = nondet_int()
+__ESBMC_assume(a == -5)
+check(a // 2)
--- a/regression/python/github_4548_floordiv_negative/test.desc
+++ b/regression/python/github_4548_floordiv_negative/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/src/python-frontend/python_math.cpp
+++ b/src/python-frontend/python_math.cpp
@@ -581,7 +581,9 @@ exprt python_math::handle_floor_division(
   pos_remainder.copy_to_operands(remainder, gen_zero(div_type));
 
   // diff_signals = is_num_neg ^ is_den_neg;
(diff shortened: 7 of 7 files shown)


mikhailramalho:
I saw this issue when adding support for camada, I have the same fix in the PR https://github.com/esbmc/esbmc/pull/4179 hehe

Issue #4568, opened 2026-05-16
[python] support threading.Thread / start / join with args capture (follow-up to #4565)


lucasccordeiro:
## Summary

Follow-up to #4565 (which lands `threading.Lock` only). This issue tracks the work required to support `threading.Thread`, `Thread.start`, `Thread.join`, and the `args=(...)` capture pattern, so the full reproducer from the original threading issue can be verified end-to-end — including data-race and deadlock detection.

The blocker is the trampoline: `pthread_create` takes a `void *(*)(void *)` function, but Python's `Thread(target=f, args=(...))` is a runtime callable bound to runtime arguments. We need a sound, statically-resolvable bridge between the two.

## Goal

The full example from the original threading issue must verify correctly:

```python
import threading

class SharedResource:
    def __init__(self):
        self.mutex = threading.Lock()
        self.lock = threading.Lock()
        self.A_count = 0
        self.B_count = 0

def thread_A(resource):
    resource.mutex.acquire()
    resource.A_count += 1
    if resource.A_count == 1:
        resource.lock.acquire()
    resource.mutex.release()
    # ... second section ...

def main():
    resource = SharedResource()
    t1 = threading.Thread(target=thread_A, args=(resource,))
    t2 = threading.Thread(target=thread_B, args=(resource,))
    t1.start()
    t2.start()
    t1.join()
    t2.join()
```

Today this is refused at parse time by #4565's guard with the "not yet supported" error.

## Design — per-call-site trampoline (recommended)

For each `Thread(target=f, args=(a, b, c))` construction site:

1. **Statically resolve `target`** to a function symbol via the AST. The keyword arg must be a `Name` (or attribute chain) the frontend can resolve at conversion time. Refuse loudly otherwise — `target=` bound to a runtime variable yields imprecise function-pointer resolution and erodes soundness.
2. **Capture args** into a synthesized struct allocated at construction time. The struct's field types come from the `args` tuple expressions evaluated at the construction site. Field 0 holds `a`, field 1 holds `b`, etc.
3. **Synthesize a trampoline** `void *__pythread_trampoline_<N>(void *self_ptr)` whose body is `f(((ArgStruct*)self_ptr)->arg_0, ..., ->arg_n); return NULL;`. The trampoline is unique per construction site so the call inside it is a direct call edge in the GOTO program (no function-pointer indirection in symex).
[...]

PR #4569, opened 2026-05-16
[python] parse-level threading.Thread / start / join lowering

lucasccordeiro:
Phase 1 of #4568. Adds parse-time validation of `threading.Thread(target=, args=)`
plus an AST desugarer that lowers accepted constructions to a per-site
synthesised trampoline + `__ESBMC_spawn_thread` call. `.start()` lowers
to the spawn intrinsic, `.join()` to `__ESBMC_assume` on a per-site
ended flag. Subclassing, lambda targets, non-tuple-literal `args=`,
unsupported kwargs (daemon/name/kwargs/group), positional arguments,
construction inside loops, and reassignment are refused at parse time
so the model cannot silently weaken concurrency semantics.

Out of scope here: race / deadlock witnesses do not reliably surface
under the current assume-based join. Replacing it with proper
pthread_create / pthread_join wiring is the next step.


lucasccordeiro:
Thanks for the careful review. Pushed `06b45cb` addressing five of the seven items:

- **`__pythread_ended_<N>` missing `global`** — this was the root cause of the vacuous-join behaviour I'd noted in the PR description. With `global` added, cross-thread writes are now correctly observed by `main` after `join`; the `must_fail2` smoke test I'd been struggling with now correctly reports `VERIFICATION FAILED`.
- **Residual Thread/start/join in nested blocks** — added a post-desugar sweep that rejects any `Thread(target=, args=)` / `t.start()` / `t.join()` that survived because it sat in `if` / `try` / `with`. Bare `Thread()` (the desugar output) is allowed through.
- **Bare-constructor alias choice** — now reuses `call_node.func` directly via `copy.deepcopy` instead of picking from a set.
- **`start(...)` / `join(timeout=...)` silently accepted** — now rejected at parse time with a clear message.
- **Stale `visit_Expr` docstring** — corrected (join is one `__ESBMC_assume`, not a three-statement atomic block).

New regression tests:
- `threading_thread_args_tuple_safe` — positive `Thread(target=f, args=(h, 42))` exercise that asserts `h.value == 42` and passes.
- `threading_thread_nested_block_unsupported` — rejection of `if True: t = Thread(...)`.
- `threading_thread_start_with_args_unsupported` — rejection of `t.join(1.0)`.

**Deferred** (called out for a follow-up):
- **Args evaluated at thread start, not at construction time** (your #3) — the current design has the trampoline reference each arg expression directly so multiple threads on the same shared instance see the same memory. Capturing into per-site temps would re-introduce the struct-copy aliasing problem I hit in earlier iterations (each Thread would get its own copy of e.g. `SharedResource`). Phase 2 will switch the lowering to `pthread_create` with the args struct passed via the `void *` parameter, which gives both correct evaluation timing and shared aliasing.

PR #4571, opened 2026-05-16
[python] refuse threading.Thread at parse until sound wiring lands

lucasccordeiro:
Reverts the Thread-accepting parts of #4569 (parser desugarer, trampolines, `__ESBMC_spawn_thread`/`__ESBMC_terminate_thread` paths) and restores parse-time refusal of `threading.Thread`.


lucasccordeiro:
## Why "patch the crash and keep the desugarer" is unsafe

The segfault that triggered this work was the loud failure mode of a model that was already unsound for the property the user was asking about. Posting the analysis so a future attempt at #4568 has the bar to clear.

### The reported crash

`python_converter::get_var_assign` (`src/python-frontend/converter/converter_stmt.cpp:1565`) called `create_lhs_expression(target, lhs_symbol, ...)` with `lhs_symbol == nullptr`, which dereferences the null inside `symbol_expr()` (`src/util/expr_util.cpp:251`). The guard at 1562 only threw when `!is_global`, so a `global`-declared name with no module-level symbol fell through to the null deref.

The trigger was the parser-level threading desugarer added in #4569: it synthesised, into the **enclosing function body** (e.g. `main`),

```python
__pythread_ended_<N> = 0                       # plain Assign  →  local of main
def __pythread_trampoline_<N>():
    global __pythread_ended_<N>                # binds to module-level (doesn't exist)
    f(...)
    __ESBMC_atomic_begin()
    __pythread_ended_<N> = 1                   # ← Assign with is_global=true, lhs_symbol=null → SEGV
    __ESBMC_atomic_end()
    __ESBMC_terminate_thread()
```

`preregister_global_variables` only handles module-top-level `AnnAssign`, so no `__pythread_ended_<N>` symbol existed; `is_global_variable()` returned true (the trampoline pushed the name to `global_declarations`); lookup returned null; segfault. Minimal reproducer: any program with ≥2 `threading.Thread(target=, args=)` sites.

### Why patching just the crash is unsafe

Three independent soundness defects in the desugarer; the crash hid all three.

**1. `__ESBMC_assume`-based join does not synchronise.** `.join()` was lowered to `__ESBMC_assume(__pythread_ended_<N> == 1)`. The intrinsic only prunes infeasible paths in the scheduler's search; it imposes no happens-before edge between the spawned thread's body and the join. A real `pthread_join` is a barrier — this isn't. #4569's commit message admits this: *"race / deadlock witnesses do not reliably surface under the current assume-based join."*

**2. The ended-flag bound to the wrong scope.** The init was emitted as a function-local of the enclosing function; the trampoline declared it `global`; the join read the enclosing function's local. Two different `__pythread_ended_<N>` variables. After patching the crash by hoisting the init to module level, the bindings would line up — but only at the cost of compounding (1) with a now-correctly-readable flag that still doesn't synchronise.

**3. Trampoline argument capture relies on closures the frontend doesn't implement.** `_make_trampoline` references the construction-site arg variables (e.g. `resource`) by name from inside a nested `def` defined in the same function body. Python semantics make that a closure capture of the enclosing function's local. ESBMC's Python frontend does not implement closures over enclosing function locals — the spawned thread would run against an uninitialised `resource`, and the verifier would report SUCCESSFUL on a model that bears no resemblance to the user's program.

For a verifier, a crash is loud; a wrong SUCCESSFUL is silent. The second is the worse failure mode. Hence the revert rather than the patch.

### The bar for the next attempt (#4568)

To restore `threading.Thread` soundly, the wiring needs all of:

1. Real `pthread_create` / `pthread_join` lowering — not `__ESBMC_assume`. The intrinsic must establish the synchronisation edge symex's interleaving search depends on for race/deadlock witnesses to surface.
2. Cross-thread argument passing that does **not** rely on Python closure capture. Options: marshal args through module-level globals the desugarer creates; pass through the trampoline's parameter list; require `args=` to name only module-globally-visible objects. Pick one and verify the symbol-table view of the spawned thread's frame matches user expectations.
3. **Positive and negative coverage**: at minimum, one program with a known race that must produce a counterexample, one with a known deadlock that must produce a counterexample, and the sound-by-construction case (e.g. a properly-locked counter) that must verify SUCCESSFUL. The current revert deliberately drops the eight `_safe` regression tests #4569 added because they asserted SUCCESSFUL on the unsound model — leaving them in would have CI assert that the unsound behaviour was correct, which is the failure mode we cannot tolerate.
4. The `get_var_assign` guard introduced in this PR stays. Any future synthesised AST that references an unregistered global now produces a diagnostic instead of a SIGSEGV, regardless of which feature added it.

PR #4615, opened 2026-05-18
[termination] reduce non-termination to assert(false) reachability

mikhailramalho:
## Summary

Rework `--termination` to use the standard SV-COMP-style reduction:

- `goto_termination` (new function in `goto_k_induction.cpp`) applies
  the k-induction loop transformation (havoc loop vars + assume entry
  condition) to every function except `__ESBMC_main`, then inserts an
  `assert(false)` marker right before `__ESBMC_main`'s `END_FUNCTION`.
- The strategy loop reads the property as
  `CHECK(init(main()), LTL(F end))`:
  - **FC UNSAT at k** — all paths terminated within k iters →
    SUCCESSFUL.
  - **IS UNSAT at k** — a non-terminating execution exists →
    FAILED.
  - Otherwise — try larger k, eventually UNKNOWN at the cap.
- IS UNSAT is only sound when k-induction actually havoc'd the loop
  variables. Under `--add-symex-value-sets`, pointer-only loops are
  skipped by the havoc transform, so the IS non-termination signal is
  disabled in that mode and FC alone proves termination.
- `goto_loop_simplify`'s `--termination` block is disabled: the
  previous narrow rewrite (`while (1) {}` → `assume(false)`) fed FC a
  false termination signal that masked real non-termination as
  SUCCESSFUL.
- Marker insertion in `goto_termination` is shape-agnostic: anchored
  to `__ESBMC_main`'s `END_FUNCTION` instead of a specific entry-call
  symbol, so it works for C++ mangled names, Python frontend
  (`python_user_main`), and `--function-renamed` entries.

## Tests

- `termination13`: `while (1) assert(0);` — `assert` aborts, every
  execution finite → SUCCESSFUL (reclassified from KNOWNBUG/FAILED).
- `termination14`: `while (1) { break; } assert(0);` — break exits,
  assert aborts → SUCCESSFUL (reclassified from KNOWNBUG/FAILED).
- `termination15`: `while (1) {}` — no `end` event, infinite →
  FAILED (new test, pins the soundness of the goto_loop_simplify
  disable).
- `github_2091`: `while(y>=0 && y<=10) { y = (2y+1)/2; }` — genuinely
  non-terminating (y=0 produces y=0 forever); verdict updated from
  UNKNOWN to FAILED.

## Test plan

- [x] `ctest -L termination -j16` — 16/16 pass


mikhailramalho:
Termination only:
PR:
```
Statistics:           2395 Files
  correct:            1132
    correct true:      762
    correct false:     370
  incorrect:             2
    incorrect true:      1
    incorrect false:     1
  unknown:            1261
  Score:              1846 (max: 3816)
```
master:
```
Statistics:           2395 Files
  correct:             764
    correct true:      764
    correct false:       0
  incorrect:             3
    incorrect true:      3
    incorrect false:     0
  unknown:            1628
  Score:              1432 (max: 3816)
```

lucasccordeiro:
Well done, @mikhailramalho!

PR #4633, opened 2026-05-19
[python] enable WITH constant propagation under incremental-bmc/k-induction

### LINKED ISSUE #4629: [symex] --incremental-bmc accepts a false Python dict-keys property that --unwind refutes
## Summary

`--incremental-bmc` returns `VERIFICATION SUCCESSFUL` on a property that `--unwind` correctly refutes with a concrete counterexample. Two configurations of ESBMC return opposite verdicts on the same input — a soundness divergence in the incremental-BMC / k-induction path.

## Reproducer

`regression/python/esbmc-module-nondet3/main.py` (current, before the test-content fix in PR for #4624 follow-up):

```python
from esbmc import nondet_dict, __ESBMC_assume, __ESBMC_assert

def test_symbolic_dict() -> None:
    d = nondet_dict()                       # default expansion: int -> int
    __ESBMC_assume(len(d) <= 3)

    for k in d.keys():
        __ESBMC_assert(isinstance(k, str), "All keys are strings")  # genuinely false

test_symbolic_dict()
```

This property is **logically false**: `nondet_dict()` with default arguments expands to an `int -> int` dict, so iterating `d.keys()` yields ints, not strings.

### `--unwind 9 --bitwuzla --no-standard-checks`
```
VERIFICATION FAILED
Violated property: All keys are strings — ISINSTANCE(k, {  })
```
Correct verdict — concrete counterexample with `size=1` and an int key.

### `--incremental-bmc --bitwuzla --no-standard-checks`
```
VERIFICATION SUCCESSFUL
Solution found by the forward condition; all states are reachable (k = 8)
```
**Wrong verdict** — a false property is being accepted as proved.

## Why it was hidden until now

The `__ESBMC_assert` Python intrinsic was a no-op stub in `models/esbmc.py` until PR #4597 (`205b403779`) added the real `code_assertt` lowering. Before #4597, both verdicts said SUCCESSFUL by vacuous truth — the assertion was silently dropped. After #4597 the assertion is actually checked, and the divergence is visible.

Bisect across the 24 commits between `a8a5059ada` (verified GOOD via PR #4502 Linux CI) and `f654b61063` (BAD) pinpointed `205b403779` as the first commit to expose the divergence. The underlying `--incremental-bmc` false negative predates that commit.

## Scope

- Bug class: false negative in `--incremental-bmc` (and likely `--k-induction` forward-condition convergence).
- Risk: any property that depends on values flowing through Python dict `.keys()` / `.values()` iterators may be vacuously "proved" by incremental-BMC even when it does not hold.

## Suggested investigation

1. Reduce to a minimal C / GOTO repro independent of the Python frontend to isolate whether the bug lives in the Python list/dict operational model, in the k-induction forward-condition encoding, or in the interaction.
2. Bisect the divergence backward (separate from #4597) through the symex / k-induction history to find the introducing change.
3. Add a regression that runs `regression/python/esbmc-module-nondet3` under both `--unwind 9` and `--incremental-bmc` and fails CI if the verdicts disagree.

## Cross-references

- Test was disabled (`THOROUGH` → `KNOWNBUG`) in `21672959e9` to silence CI noise; that demotion should be reverted once this issue is fixed.
- Related but distinct: `597ede0892` disabled `dict_fromkeys_int` (#4624). Worth checking whether that test exposed the same underlying class of bug.

lucasccordeiro:
Fix #4629.

WITH constant propagation is sound -- it only fires when the whole
with-chain is built from constant updates over a propagatable root.
The early-return for `--incremental-bmc` / `--k-induction` was a
performance heuristic (see the pre-existing FIXME), not a soundness
guard.

For the Python dict operational model the heuristic was too aggressive:
suppressing WITH propagation kept dict field updates out of the symex
constant table, so under `--incremental-bmc` the forward-condition
machinery declared the program proved before the per-iteration
`__ESBMC_assert(isinstance(k, str), ...)` was ever reached. `--unwind`
refuted the same property correctly because it does not gate proof on
the forward condition.

This patch enables WITH propagation under incremental-bmc/k-induction
for the Python frontend only; C/C++ retains the legacy perf shortcut.
Regressions `regression/python/github_4629{,_unwind}` pin the verdict
under both drivers.


mikhailramalho:
I'm confused with the description, is the constant propagation of with unsound?

lucasccordeiro:
Thanks -- that matches the patch as it stands: the new condition gates the early-return on `config.language.lid != language_idt::PYTHON`, so C/C++ keeps the legacy perf shortcut and only the Python frontend opts into WITH propagation under `--incremental-bmc` / `--k-induction`. Improving the WITH expression itself to remove the perf concern is left as a follow-up.

PR #4599, opened 2026-05-18
[regression] enable increment-race assertion via --data-races-check

lucasccordeiro:
Flips the threading.Thread increment-race test from KNOWNBUG to CORE by
adding --data-races-check, which yields at shared-global accesses and
bypasses main_thread_ended pruning. The unflagged scheduler-side path is
tracked in #4634 and #4635.

Partial fix for #4584.

PR #4636, opened 2026-05-19
[symex] gate end-of-program leak check on all-threads-terminal

### LINKED ISSUE #4634: [symex] Relaxed main_thread_ended pruning introduces 6 spurious counterexamples on sv-comp
## Summary

While evaluating PR #4599 against sv-comp (300 s budget per benchmark), the
relaxation of `execution_statet::check_if_ileaves_blocked` produced **six
benchmarks that newly report `VERIFICATION FAILED` (incorrect-false / spurious
counterexample) on programs that are in fact safe**. These are not timeouts —
ESBMC actively reports a falsified assertion with a trace.

We do not yet know the per-benchmark root cause; this issue tracks the
investigation.

## Runs

- PR #4599 sv-comp run (300 s): https://github.com/esbmc/esbmc/actions/runs/26048727047
- Master sv-comp run (300 s):    https://github.com/esbmc/esbmc/actions/runs/26048709263

| Category | Master | PR #4599 | Δ |
|---|---|---|---|
| correct true        |  948 |  628 |  −320 |
| correct false       |  253 |  254 |    +1 |
| **incorrect false** | **12** | **18** | **+6** |
| incorrect true      |    4 |    3 |    −1 |
| unknown             | 1959 | 2273 |  +314 |
| **Score**           | 1829 | 1126 | **−703** |

The +314 unknowns are a separate (performance) regression from the looser
exploration; the **+6 incorrect-false is the soundness regression** tracked
here.

## Hypothesised root cause

Master's `main_thread_ended` is a tree-global flag (see #4635), which
prevents `check_if_ileaves_blocked` from generating further context switches
once the first DFS schedule has run `__ESBMC_main` to its `END_FUNCTION`.
PR #4599 narrows the gate to also require that the current schedule's thread 0
has ended, which lets the DFS continue generating interleavings in schedules
where main is still alive.

The extra interleavings reached under the new gate appear to include schedules
that the C semantics rules out — likely candidates:

1. Context switches into a thread whose stack frame is destroyed (post-
   `pthread_exit`, post-`return` from the thread function). The thread is
   marked ended but the scheduler may still pick it as a switch target on
   some paths, observing stale renamed values.
2. Context switches after `main()` returns into threads that should have been
   `pthread_join`-ed — exposing reads/writes against frames that, in the C
   abstract machine, no longer exist.
3. MPOR/DPOR partial-order constraints whose invariants relied on
   `main_thread_ended` as an implicit "no more interleavings past here"
   marker, and now generate enabled-set elements that wouldn't have been
   reachable in the source semantics.

(2) is the strongest prior — the dedicated `c:@F@main` branch in
`execution_state.cpp` that calls `interleaving_unviable = true` would have
been masked by `main_thread_ended` on second-and-later schedules in master.

## Plan

- [ ] Extract the names of the six newly incorrect-false benchmarks from the
      run artefacts above (the workflow's per-benchmark log / witness
      output). They are sv-comp tasks, so each has a stable name.
- [ ] Reproduce one locally on PR #4599's tip plus the patch that does
      *only* the symex change (no test-side `--context-bound 2` bumps).
- [ ] Inspect the falsified property and the counterexample trace —
      specifically the thread id and stack state at the offending step.
- [ ] If the trace is post-termination access (hypothesis 1 or 2), the fix
      is in the scheduler's thread-liveness model, not the
      `main_thread_ended` flag.
- [ ] `creduce-reducer` on one benchmark to produce a 20–50-line
      reproducer for the regression test suite.
- [ ] Add a regression test pinning the correct verdict so any future
      relaxation cannot re-introduce the regression silently.

## Related

- PR #4599 — the change that exposed this; the scheduler-gate hunk was
  reverted out of the PR pending this investigation.
- Companion issue #4635 — `main_thread_ended` is unsound by construction
  (this issue is the prerequisite work to landing the relaxation).
- #4584 — original Python race-detection bug that motivated PR #4599.
- PR #4587 — `get_expr_globals` Python module-global race eligibility
  (sound; not implicated here).

## Acceptance criteria

The six benchmarks are root-caused, the underlying scheduler issue is fixed
soundly (without re-introducing master's unsound throttle), and a regression
test pins each affected benchmark.

lucasccordeiro:
`add_memory_leak_checks()` roots its globals-only reachability constraint
at `value_set_analysist::get_globals()`, so a thread still mid-execution
that holds the only live pointer to a dynamic object via its stack frame
is misread as a forgotten-memory violation. Fire the check only when every
thread in the schedule has reached a terminal state.

Recovers the seven `false(valid-memtrack)` sv-comp regressions surfaced by
#4599 (now: 2 correct-true + 5 unknown, zero incorrect-false). Regression
test `github_4634` reproduces the spurious failure on master under
`--data-races-check` and now passes; `github_4634_fail` confirms genuine
leaks still fire.

Fixes #4634.

### DIFF
--- a/regression/esbmc/github_4634/main.c
+++ b/regression/esbmc/github_4634/main.c
@@ -1,0 +1,68 @@
+// Regression for #4634.
+//
+// Two-element linked list `A -> p`. A spawned thread inserts a new
+// node between them by saving the old `A->next` in a stack local `t`,
+// then overwriting `A->next`. Until the thread completes, the global
+// head `A` no longer reaches the original `p` directly, but `t` does —
+// at every schedule the program is memory-safe (the join pulls the
+// thread to completion before main returns).
+//
+// Before #4634, `add_memory_leak_checks()` rooted its reachability
+// constraint at globals only, so the intermediate schedule where the
+// thread is suspended between `A->next = new_node` and the splice
+// back-link was reported as a forgotten-memory violation whenever the
+// `main_thread_ended` throttle was bypassed (e.g. `--data-races-check`).
+
+#include <pthread.h>
+#include <stdlib.h>
+
+struct s
+{
+  int datum;
+  struct s *next;
+};
+
+static struct s *A;
+static pthread_mutex_t A_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+static void init(struct s *p, int x)
+{
+  p->datum = x;
+  p->next = NULL;
+}
+
+static void *t_fun(void *arg)
+{
+  (void)arg;
+  struct s *p = malloc(sizeof(struct s));
+  if (!p)
+    return NULL;
+  struct s *t;
+  init(p, 7);
+
+  pthread_mutex_lock(&A_mutex);
+  t = A->next;
+  A->next = p;
+  p->next = t;
+  pthread_mutex_unlock(&A_mutex);
+  return NULL;
+}
+
+int main(void)
+{
+  pthread_t t1;
+  struct s *p = malloc(sizeof(struct s));
+  if (!p)
+    return 0;
+  init(p, 9);
+
+  A = malloc(sizeof(struct s));
+  if (!A)
+    return 0;
+  init(A, 3);
+  A->next = p;
+
+  pthread_create(&t1, NULL, t_fun, NULL);
+  pthread_join(t1, NULL);
+  return 0;
+}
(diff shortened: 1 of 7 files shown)


Anthonysdu (approved):

lucasccordeiro:
This PR (300s): 

```
  correct:            1201
    correct true:      948
    correct false:     253
  incorrect:            15
    incorrect true:      4
    incorrect false:    11
  unknown:            1960
  Score:              1845 (max: 5722)

https://github.com/esbmc/esbmc/actions/runs/26102446642.
```

Master (300s): 

```
  correct:            1201
    correct true:      948
    correct false:     253
  incorrect:            16
    incorrect true:      4
    incorrect false:    12
  unknown:            1959
  Score:              1829 (max: 5722)

https://github.com/esbmc/esbmc/actions/runs/26048709263/job/76594971026
```

Issue #4642, opened 2026-05-19
[python] Support arbitrary-precision (bignum) Python int


lucasccordeiro:
## Motivation

Python ints are arbitrary-precision, but ESBMC's Python frontend approximates them with `long_long_int_type()` (int64) — see the `FIXME` at `src/python-frontend/type_handler.cpp:427`. Any program that legitimately uses values outside `[-2^63, 2^63 - 1]` currently either:

1. Silently wraps (wrong verification result), or
2. After #4641, aborts with `ERROR: Python int overflow: …` when the offending value is reached by constant folding of `**`.

The diagnostic added in #4641 only covers the constant-folded `**` path; the underlying limitation is much broader.

## Known residual gaps

- `**` with exponent > `kMaxConstantFoldExponent` (1024) — falls through to `build_power_expression`, which builds a symbolic multiplication tree at the same int64 type and wraps silently (see TODO at `src/python-frontend/python_math.cpp` in #4641).
- Direct big-int literals (e.g. `x = 18446744073709551616`) — overflow at frontend-level constant ingest, not via `handle_power`.
- Arithmetic that produces overflowing results from in-range operands (e.g. `(2 ** 62) * 8`, `factorial(21)`).
- `int.bit_length()`, `int.from_bytes()`, `int.to_bytes()` etc. operating on values that don't fit int64.
- The `(2 ** 64).bit_length() == 65` case from #1964, currently parked as `KNOWNBUG` in `regression/python/github_1964_bit_length_bignum/`.

## Scope

Replace the fixed-width approximation with arbitrary-precision integers for Python `int`. Likely touches:

- `src/python-frontend/type_handler.cpp` — `get_typet("int", …)` lowering.
- `src/python-frontend/python_math.cpp` — all arithmetic operators (`+`, `-`, `*`, `/`, `//`, `%`, `**`, shifts, bitwise).
- `src/python-frontend/converter/converter_binop.cpp` and `converter_compare.cpp` — operand promotion.
- Operational models that consume `int`: `src/python-frontend/models/int.py`, `src/c2goto/library/python/list.c`, etc.
- SMT encoding — depending on representation choice (BigInt-typed struct vs. dynamic-width bv vs. SMT `Int` sort).

## Acceptance criteria

- `regression/python/github_1964_bit_length_bignum/test.desc` flips from `KNOWNBUG` to `CORE`.
[...]

PR #4648, opened 2026-05-20
[irep2] simplify arithmetic and relations over bigint without truncation

lucasccordeiro:
Phase 2B for issue #4642. Stacks on PR #4646 (bigint_type2t kind) — base will fast-forward to master once #4646 lands.

`constant_int2t` stores `BigInt` so the IR can already represent arbitrary precision, but the simplifier wasn't routing bigint operands through the fold paths. Constructing `add2tc(bigint, _, _)` in a debug build threw a width-parity assert (`t->get_width()` on bigint throws `symbolic_type_excp`); in release builds the fold path was unreachable because the dispatch gated on `is_bv_type`, which intentionally excludes bigint. The structural arithmetic node survived to symex without truncation — but the constant fold never happened, leaving sub-expressions like `add(2^200, 2^200)` un-simplified through symex and into SMT lowering.

This PR closes the gap so PR 2C (SMT lowering) and PR 3 (Python frontend cutover) can rely on simplifier fold-through for bigint operands.

Changes:

- `irep2_expr.cpp` — `assert_arith_2ops_consistency` gains an explicit `is_bigint_type(t)` branch that requires both operands to be bigint and skips the bv width parity check.
- `expr_simplifier.cpp` — `simplify_arith_2ops`, `simplify_arith_1op`, and `simplify_relations` now also dispatch when either operand is bigint. TFunctor's `BigInt` arithmetic folds both bv and bigint uniformly; the trailing `from_integer` round-trip width-clips bv and is a no-op for bigint thanks to the PR #4646 fixup.
- `equality2t::do_simplify` / `notequal2t::do_simplify` — the `(x + c1) == c2 -> x == (c2 - c1)` rewrite and its (x - c1) sibling call `fits_in_width(_, side_2->type->get_width(), ...)`. Bigint has no width — the rewrites now skip bigint operands. The structural comparison survives for SMT lowering.

17 new SCENARIO GIVENs / 49 new assertions cover add, sub, mul, div, modulus, neg, identity short-circuit, depth-2 chain, mixed-sign mul, div(0,_), equality/inequality fold, and lessthan fold — every test uses operands at magnitude 2⁸ or higher with peaks at 2⁴⁰⁰, proving no fixed-width truncation along any path. Full unit suite 352/352 pass; Python and C regression samples zero real failures.

Refs #4642


lucasccordeiro:
This PR should solve the problem for now: https://github.com/esbmc/esbmc/pull/4656.

The changes are also centralized in the Python frontend.

PR #4649, opened 2026-05-20
[solver] route bigint_type2t to SMT_SORT_INT on Int-capable backends

lucasccordeiro:
Phase 2C for issue #4642. Stacked on PR #4646 (`bigint_type2t` kind) and PR #4648 (PR 2B: simplifier folds bigint without truncation). PR 2D — separate, future — will route bigint to a wide signed bitvector with an overflow trap on bitvector-only backends (bitwuzla, boolector); this PR addresses the Int-capable backends only (z3, mathsat, smtlib, cvc4, cvc5; yices works but with Real-as-Int caveats for division semantics tracked for PR 3).

Routes `bigint_type2t` to `SMT_SORT_INT` end-to-end:

- `smt_conv.cpp::convert_sort` — `bigint_id` case returns `mk_int_sort()`. On Int-capable backends the override fires; on bv-only backends the base `mk_int_sort` aborts with the existing "Chosen solver doesn't support integer sorts" diagnostic.
- `convert_terminal::constant_int_id` — when the operand type is bigint, route to `mk_smt_int` BEFORE the `expr->type->get_width()` call (which would otherwise throw `symbolic_type_excp`). Independent of the global `int_encoding` option.
- `convert_ast` for `add/sub/mul/div/modulus/neg` and the four order comparisons (`lt/le/gt/ge`): each `else if (int_encoding)` arm is widened to also fire on bigint operand/result types so the Int-sort builders dispatch correctly. `equality`/`notequal` already use sort-agnostic `args[0]->eq(...)` and need no change.

New `unit/solvers/z3_bigint_smt_test` (gated by `ENABLE_Z3`) instantiates the full Z3 backend via `create_new_z3_solver` and pins:
- `convert_sort(bigint_type2tc())` returns `SMT_SORT_INT`.
- `convert_ast` on `constant_int2tc(bigint, _)` produces an Int-sort term with the correct decimal numeral.
- `convert_ast` on `add/mul/neg` over bigint operands lands on Int-sort terms with the correct values.
- A sort-only assertion at 2^200 magnitude (numeral correctness past int64 is the orthogonal PR #4647 (Phase 2A.5) fix to `mk_smt_int`; sort dispatch holds at any magnitude).

Verification: full unit suite 355/355 pass (3 new SCENARIOs, 13 assertions). Python and C regression samples zero real failures.

Refs #4642


lucasccordeiro:
This PR should solve the problem for now: https://github.com/esbmc/esbmc/pull/4656.

The changes are also centralized in the Python frontend.

PR #4650, opened 2026-05-20
[CI] add witness validation workflow and improve compare-runs script

XLiZHI:
Add a new GitHub Actions workflow (`witness-validation.yml`) that validates SV-COMP witnesses produced by ESBMC or CPAChecker against a given benchexec run. The workflow supports three validation modes (Full / RunSet / Task) and generates an HTML comparison table via `compare-runs.sh`.


lucasccordeiro (approved):

PR #4693, opened 2026-05-21
[ci] add macOS to nightly release matrix

pmatos:
## Summary

- Build ESBMC on `macos-latest` alongside Ubuntu and Windows in the nightly release workflow.
- Bundle the resulting `./release` tree as `esbmc-macos.zip` and attach it to the nightly GitHub release.
- The macOS matrix entry uses `-b RelWithDebInfo -e OFF -c 21`, matching the clang version already used by the PR-time `build-macos-arm` job (`.github/workflows/pull_request.yml`).

## Rationale

`scripts/build.sh` already has full Darwin support and the PR workflow has been building successfully on `macos-latest` for some time, but the release workflow only ships Linux and Windows zips. This bridges that gap so nightly releases include a macOS arm64 binary as well.

The change is isolated to `release.yml` — no changes to `build-unix.yml`, `build.sh`, or any other workflow.

## Test plan

- [ ] Trigger the workflow via `workflow_dispatch` (or wait for the Monday cron) and confirm all three OS builds succeed.
- [ ] Confirm `esbmc-linux.zip`, `esbmc-windows.zip`, `esbmc-macos.zip`, and `esbmc.info` are attached to the resulting nightly release.
- [ ] Download and run `esbmc --version` from the macOS zip on a real arm64 macOS host.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,11 +10,17 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-24.04, windows-latest]
+        include:
+          - os: ubuntu-24.04
+            flags: '-b RelWithDebInfo -e OFF'
+          - os: windows-latest
+            flags: '-b RelWithDebInfo -e OFF'
+          - os: macos-latest
+            flags: '-b RelWithDebInfo -e OFF -c 21'
     uses: ./.github/workflows/build.yml
     with:
       operating-system: ${{ matrix.os }}
-      build-flags: '-b RelWithDebInfo -e OFF'
+      build-flags: ${{ matrix.flags }}
       testing: false
 
   build-info:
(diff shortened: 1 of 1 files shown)


rafaelsamenezes:
Is this a static build?

pmatos:
> Is this a static build?

I thought static doesn't work on macos:

https://github.com/esbmc/esbmc/blob/e10dfdf161137f56840c68acb7d25e239b27f461/scripts/build.sh#L200-L207

rafaelsamenezes:
> > Is this a static build?
> 
> I thought static doesn't work on macos:
> 
> https://github.com/esbmc/esbmc/blob/e10dfdf161137f56840c68acb7d25e239b27f461/scripts/build.sh#L200-L207

Then we need to be sure that we are shipping all the libraries with the package. Otherwise users might be confused by downloading it and having missing a library error.

pmatos:
Many other tools simply rely on libraries being available in the host as
shared libraries. What's the policy for esbmc?

rafaelsamenezes:
> Many other tools simply rely on libraries being available in the host as shared libraries. What's the policy for esbmc?

For linux we do static and for Windows we ship the DLLs.

To be honest I am fine following your lead here. I do think that we should be able to move towards dynamic builds and using users own libraries.

lucasccordeiro (approved):

Issue #4715, opened 2026-05-22
Migrate goto-program pipeline from legacy IREP to IREP2 (tracking)


lucasccordeiro:
## Summary

Migrate ESBMC's goto-program pipeline from the legacy stringy IREP (`irept`/`exprt`/`typet`/`codet`) to typed IREP2 (`expr2tc`/`type2tc`), preserving behaviour **exactly**. Tracking/umbrella issue; each phase below is an independently reviewable, revertible PR set.

Behavioural equivalence is the hard constraint: ESBMC is a verifier, so determinism and semantic correctness outrank performance or cleanup. No big-bang rewrites.

## Key finding: the core is already migrated

The goto **instruction representation is already on IREP2** — `goto_programt::instructiont::code` and `::guard` are `expr2tc` (`src/goto-programs/goto_program.h:93,105`), as are loop contracts and the loop / k-induction / abstract-interpretation passes. This is therefore **not a core rewrite** but a job of draining six well-defined legacy *boundaries*.

## The six surviving legacy boundaries

| # | Boundary | Location | Risk |
|---|----------|----------|------|
| B1 | Frontend → goto input (`goto_convert*` consume `codet`) | `goto_convert.cpp:117,136` | Medium |
| B2 | **Symbol table** (`symbolt::value`/`type` are `exprt`/`typet`) | `src/util/symbol.h:15-16` | **High — foundational** |
| B3 | `goto_functiont::type` is `code_typet` | `src/goto-programs/goto_functions.h:24` | Medium |
| B4 | **Goto-binary serialization** still `irept` | `goto_program_irep.cpp`, `util/irep_serialization.*` | **High — format + hashing** |
| B5 | `rw_set` data-race pass works in `exprt` | `rw_set.h:17-18` | Medium |
| B6 | Passes that build `exprt` then `migrate` (round-trips) | `goto_check.cpp`, `add_race_assertions.cpp`, `contracts/contracts.cpp:1730,1896,…` | Medium |

**Ordering constraint:** the `contracts.cpp` `migrate_*_back` round-trips exist only because the symbol table (B2) stores legacy types — so B6 cannot be fully removed until B2 is done.

## High-risk subsystems

[...]

PR #4718, opened 2026-05-22
[goto-programs] migrate rw_set data-race analysis to IREP2

lucasccordeiro:
Phase 2.1 of the goto-program legacy-IREP → IREP2 migration (#4715), and the first code migration validated against the Phase 0 differential harness.

Retypes `rw_sett`/`entryt` to `expr2tc`/`guard2tc` and rewrites the read/write recursion over typed IREP2 nodes (`code_assign2t`, `code_function_call2t`, `symbol2t`, `member2t`, `index2t`, `dereference2t`, `if2t`, …), replacing the legacy `exprt` string dispatch. `add_race_assertions` now passes `instruction.code`/`guard` directly — no `migrate_expr_back` round-trip on the per-instruction hot path. `w_guardst` stays on legacy `exprt` for now (its symbol creation is deferred to Phase 4).

Two IREP2-vs-legacy representation differences are handled so the instrumented goto program stays byte-identical to the pre-migration output:
- the NULL pointer is a namespace-less `symbol2t` in IREP2 (legacy saw a `constant_exprt`) — skip symbols absent from the namespace;
- FP nodes (`typecast`/`nearbyint`/`ieee_*`) carry `__ESBMC_rounding_mode` as an operand the legacy `exprt` did not expose positionally — recurse only into value operands. Genuine reads/writes of the rounding-mode global still register via the symbol path.

**Validation:** zero goto-program diff across the 85-test data-race regression corpus (3 Python tests differ only in the `__file__` astgen-temp-path byte array — proven build-independent pre-existing non-determinism). The `race`/CUDA/`pthread`/`threading` ctest suites pass.

### DIFF
--- a/src/goto-programs/add_race_assertions.cpp
+++ b/src/goto-programs/add_race_assertions.cpp
@@ -55,12 +55,13 @@ public:
 
   const exprt get_w_guard_expr(const rw_sett::entryt &entry)
   {
-    return get_guard_symbol_expr(entry.original_expr);
+    return get_guard_symbol_expr(migrate_expr_back(entry.original_expr));
   }
 
   const exprt get_assertion(const rw_sett::entryt &entry)
   {
-    return gen_not(get_guard_symbol_expr(entry.original_expr));
+    return gen_not(
+      get_guard_symbol_expr(migrate_expr_back(entry.original_expr)));
   }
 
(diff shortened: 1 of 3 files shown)


XLiZHI (approved):
LGTM

PR #4721, opened 2026-05-22
[goto-programs] migrate goto_functiont::type to IREP2 (Phase 3)

lucasccordeiro:
Phase 3 of the goto-program legacy-IREP → IREP2 migration (#4715). Independent of the rw_set PRs (#4718/#4719); branches off master.

Changes `goto_functiont::type` from the legacy `code_typet` to `type2tc` (a `code_type2t`) across 10 sites in 8 files:
- writes use `migrate_type(symbol.type)` (goto_convert, read_bin, contracts);
- symex drops its now-redundant `migrate_type` of the field (one round-trip removed);
- `has_return_value` becomes `ret_type->type_id != empty_id` — `migrate_type` collapses `constructor`/`destructor` return types to `empty_type` (migrate.cpp:331-341), so the legacy three-way id check is exactly equivalent;
- `get_local_identifiers` reads `code_type2t::argument_names`;
- `add_return`, goto_inline `parameter_assignments`, and goto2c keep their legacy `code_typet` logic via a boundary `migrate_type_back` (those consumers migrate later).

`goto_functiont::type` is **never serialized** (rebuilt from the symbol table on read), so there is no goto-binary format coupling.

**Validation:** zero goto-program diff *attributable to this change* across 499 `esbmc-cpp/cpp` tests + the 85-test data-race corpus. The only flagged diffs were a `__TIME__` timestamp (`ch19_1`) and a build-nondeterministic global zero-init reorder — both pre-existing and provably upstream of this change (the init order is fixed in `clang_c_maint::static_lifetime_init` before goto-convert runs). All affected tests pass verdicts; the only ctest failures in the sweep were pre-existing macOS SDK header parse errors. `code-reviewer` verified each mapping against `migrate.cpp` and found no material correctness defects.

### DIFF
--- a/src/goto-programs/contracts/contracts.cpp
+++ b/src/goto-programs/contracts/contracts.cpp
@@ -916,7 +916,7 @@ void code_contractst::enforce_contracts(
     goto_functiont new_func;
     new_func.body = wrapper;
     if (func_sym->type.is_code())
-      new_func.type = to_code_type(func_sym->type);
+      new_func.type = migrate_type(func_sym->type);
     new_func.body_available = true;
     new_func.update_instructions_function(original_id);
 
--- a/src/goto-programs/goto_convert_functions.cpp
+++ b/src/goto-programs/goto_convert_functions.cpp
@@ -67,9 +67,8 @@ void goto_convert_functionst::add_return(
   t->make_return();
   t->location = location;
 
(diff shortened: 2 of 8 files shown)


LukeW1999 (approved):
LGTM

PR #4755, opened 2026-05-24
[symex] guard width queries against empty/code types

lucasccordeiro:
`(lambda: ...)()` calls produced by the defaultdict-lambda lowering in the Python frontend flow into the symex layer with empty or code result types. Three sites then called `type2t::get_width()` unconditionally on those types and crashed with an uncaught `symbolic_type_excp`:

- `fixup_renamed_type` scalar-vs-scalar width-mismatch check (`is_scalar_type` includes `empty_type` and `code_type`).
- `do_memcpy_expression` primitive path (debug assert on `get_width` in `gen_byte_memcpy`).
- `intrinsic_memcpy` `type_byte_size` call (the catch already covered `dyn_sized_array_excp` and `inf_sized_array_excp` but not `symbolic_type_excp`).

Bail out in `fixup_renamed_type` and defer to the C `bump_call` impl in the memcpy path. Restores `defaultdict(lambda: 0); d[k] = v; assert d[k] == v` — the smallest reproducer from #3915. The full `quixbugs/shortest_path_lengths` test still fails on a separate `min()` mixed-type comparison.

Refs #3915

### DIFF
--- a/regression/python/github_3915_defaultdict_lambda/main.py
+++ b/regression/python/github_3915_defaultdict_lambda/main.py
@@ -1,0 +1,10 @@
+from collections import defaultdict
+
+
+def main():
+    d = defaultdict(lambda: 0)
+    d[5] = 7
+    assert d[5] == 7
+
+
+main()
--- a/regression/python/github_3915_defaultdict_lambda/test.desc
+++ b/regression/python/github_3915_defaultdict_lambda/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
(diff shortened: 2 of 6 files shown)


lucasccordeiro:
Copilot is right — verified that `type_byte_size` → `size_bits_expr` handles `empty_id` (returns 1 bit) and `code_id` (returns 0 bit) explicitly without calling `get_width()`, so that catch path is unreachable. The actual crash from #3915 is fully covered by the `do_memcpy_expression` empty/code guard. Removed in 188b8522ca.

PR #4765, opened 2026-05-24
[python] infer defaultdict/empty-dict value type from factory and literal RHS (#3915)

### LINKED ISSUE #3915: Error on regression/quixbugs/shortest_path_lengths

lucasccordeiro:
`resolve_expected_type_for_dict_subscript()` only recognised class-typed RHS when scanning subscript assignments on an unannotated dict, so `defaultdict(int)` (lowered by the preprocessor to `d[k] = int()`) and unannotated `d = {}` followed by `d[k] = 5` both fell through to the `char*` default. Downstream, `min` / `max` over `d[k]` then aborted conversion with:

```
min() arguments must be of comparable types: got pointer * subtype: signedbv * width: 8 and signedbv * width: 64
```

This patch extends the empty-dict candidate scan in `python_dict_handler.cpp::resolve_expected_type_for_dict_subscript` to also accept:

- `Constant` literal RHS — covers `d[k] = 5`, `d[k] = 0.0`, `d[k] = True`, `d[k] = "x"`.
- `Call` to built-in type constructors `int` / `float` / `bool` / `str` — covers `defaultdict(<builtin>)`, which the preprocessor lowers to `d[k] = <builtin>()`.
- Trivial `Lambda` factories whose body is a `Constant` or a builtin constructor call — covers the `defaultdict(lambda: float('inf'))` shape used in `regression/quixbugs/shortest_path_lengths`.

Adds `regression/python/github_3915` (SUCCESSFUL) and `github_3915_fail` (FAILED). The 204-test dict-related regression subset passes.

Quixbugs `shortest_path_lengths` still hits a separate unrelated error (`update()` from a dict-comprehension is not yet modelled), so it remains `KNOWNBUG` for now.

Fixes #3915


lucasccordeiro:
Follow-up commit: handle three more #3915 manifestations that the first commit did not cover.

1. `defaultdict(lambda: <expr>)` was lowered to `d[k] = (<lambda>)()`, but `function_call_builder::build_function_id` only handles `Name` and `Attribute` func types — for a `Lambda` it returned the enclosing function id, so the lambda invocation was emitted as a recursive call to the surrounding function (visible in GOTO as a literal `f()`). Sidestepped in the preprocessor: inline the nullary lambda body directly so `d[k] = <body>` routes through the same value-type-scan that handles `defaultdict(int)`.
2. `visit_Assign` only lowered defaultdict reads when the RHS was a bare `Subscript`. For `d[k] = min(d[k1], d[k2])` the missing-key checks were never inserted, so the value-type scan saw no typed assignment to anchor on. Run the lowerer on every non-Subscript Assign RHS, mirroring `visit_Return` / `visit_Assert`.
3. The empty-dict candidate scan only iterated top-level function-body statements. With (2) fixed, the inserted `d[k] = factory()` checks land inside `for` / `if` / `while` / `with` / `try` bodies. Walk the body recursively into all statement-containing nodes; skip nested `FunctionDef` / `ClassDef`.

Combined, these unblock the conversion-time `char*`-vs-numeric mismatch for the Floyd-Warshall-style nested-loop shape in `regression/quixbugs/shortest_path_lengths`. The quixbugs test itself still hits independent issues (dict-comprehension `.update()`, SMT cubic dict cost over symbolic `n`) and remains `KNOWNBUG`.

Regression: `github_3915` extended with `f_lambda_inf` (Lambda-float-inf factory) and `f_loop_min_rhs` (defaultdict read inside `min()` call inside a `for` body). 202 dict-related and 186 lambda/min/max/loop python tests pass locally.

PR #4838, opened 2026-05-25
[k-induction] symex-side pointer invariants for inductive-step havocs

mikhailramalho:
## Summary

Adapts the pointer-invariant idea from 2LS (template-based shape domain over the loop-back variable) to ESBMC's k-induction inductive step. When `make_nondet_assign` emits `p = nondet()` for a pointer-typed loop var in IS, symex normally wipes p's value-set to `{*}`, which makes the dereference-time `--add-symex-value-sets` assume only pin the resolved address (not the source pointer). The IS encoding then sees a completely unconstrained p and the solver finds spurious counterexamples on programs that are actually true.

At the IS-havoc of a pointer-typed loop var:

1. Snapshot the pre-havoc value-set entry on the L1 key.
2. Rewrite the nondet RHS itself to an ITE chain over the pre-havoc candidates, using the L1-renamed pre-havoc lhs as the SAME-OBJECT discriminator — the pre-havoc lhs is already constrained by the existing SSA chain, so the new p inherits that constraint instead of becoming a free variable pinned via a SAME-OBJECT assume the solver has to discover.
3. After the assignment lands, restore the value-set entry (filtering out unknown/invalid sinks) so deref-time encoding picks up the concrete candidates instead of falling back to `invalid_object`.

The symex-side `--add-symex-value-sets` deref-assume also no longer bails out on the first non-object_descriptor entry; it skips and continues so the deref-time pin survives mixed value-sets.

## Result on SV-COMP lost-proof tests

Of the 627 proofs lost on this branch vs master before this PR's commits, 559 are recovered.

| | before | after |
|---|---|---|
| proofs | 528 | 559 |
| timeouts | 86 | 59 |
| unknowns | 12 | 8 |

list-properties/list-1.c, forester-heap/dll-rb-cnstr_1-1.i, and ~30 other linked-list traversals now prove at IS k=3 in under a second. The remaining 59 timeouts + 8 unknowns are mostly Linux driver code where master's heuristic happened to work but our principled approach doesn't yet cover.

The bug-finding case (sll_of_sll_nondet_append) still correctly reports `Bug found at BC k=5`; the k-induction regression suite (179 tests) passes.


mikhailramalho:
This PR:
```
    correct:           20229
      correct true:    13985
      correct false:    6244
    incorrect:            19
      incorrect true:      6
      incorrect false:    13
    unknown:           16352
    Score:             33814 (max: 61270)
```
Master:
```
  correct:           20133
    correct true:    13889
    correct false:    6244
  incorrect:            20
    incorrect true:      7
    incorrect false:    13
  unknown:           16447
  Score:             33590 (max: 61270)
```

PR #4843, opened 2026-05-25
[ir-ieee] regression: accept Boolector ir-ieee incompatibility in nn float tests

ShemaAlosaimi:
## Summary

Two neural-network floating-point regression tests (`nn-logistic_5_unsafe` and `nn-tanh_5_unsafe`) hard-code `--boolector`.

When `--ir-ieee` is injected globally through `ESBMC_OPTS`, this creates an unsupported solver/encoding combination. The IR/IEEE path uses integer encoding mode, which is not supported by Boolector in this context.

ESBMC already detects this invalid combination and exits with: `ERROR: Boolector does not support integer encoding mode`.

However, both tests currently expect only: `^VERIFICATION FAILED$`.

As a result, they fail under global `--ir-ieee` injection even though ESBMC is behaving correctly by rejecting the unsupported configuration.

This PR extends the expected-output regex in both `test.desc` files to also accept the existing guard message: `^VERIFICATION FAILED$|ERROR: Boolector does not support integer encoding mode`.

Normal Boolector runs are unaffected because `VERIFICATION FAILED` still matches. Under `--ir-ieee` evaluation, the incompatibility error now also matches, so both tests remain green.

## Changed files

- `regression/floats-regression/nn-logistic_5_unsafe/test.desc`
- `regression/floats-regression/nn-tanh_5_unsafe/test.desc`

## Testing

Ran the affected tests on the normal Boolector path:

`ctest -j2 -R "regression/floats-regression/nn-(logistic|tanh)_5_unsafe"`

Result: both tests pass.

### DIFF
--- a/regression/floats-regression/nn-logistic_5_unsafe/test.desc
+++ b/regression/floats-regression/nn-logistic_5_unsafe/test.desc
@@ -1,4 +1,4 @@
 THOROUGH
 logistic_5_unsafe.c-amalgamation.c
 -I . --boolector
-^VERIFICATION FAILED$
+^VERIFICATION FAILED$|ERROR: Boolector does not support integer encoding mode
--- a/regression/floats-regression/nn-tanh_5_unsafe/test.desc
+++ b/regression/floats-regression/nn-tanh_5_unsafe/test.desc
@@ -1,4 +1,4 @@
 THOROUGH
 tanh_5_unsafe.c-amalgamation.c
 -I . --boolector
-^VERIFICATION FAILED$
+^VERIFICATION FAILED$|ERROR: Boolector does not support integer encoding mode


mikhailramalho:
Can't we just drop the boolector flag?

ShemaAlosaimi:
> Can't we just drop the boolector flag?

I considered that, but these tests seem to rely on Boolector because Z3 did not produce a result within the timeout in my testing. Dropping `--boolector` would change the purpose of the tests and may turn this into a solver-performance issue.

This PR keeps the normal Boolector behaviour unchanged (`VERIFICATION FAILED` still matches) and only accepts the existing guard error when `--ir-ieee` is injected globally, since `--boolector` + integer encoding is an unsupported combination.

PR #4847, opened 2026-05-26
[k-induction] iterate guards map in entry-cond assume (fixes #4846)

### LINKED ISSUE #4846: [k-induction] PR #4484 commit 26dd5b8353 breaks post-loop assertions in --loop-invariant flow
## Summary

PR #4484 introduced a regression in 9 regression tests across `loop-invariants`, `incremental-smt`, `witnesses_validate`, and `esbmc-solidity`. `git bisect` against `regression/loop-invariants/github_4052` (one of the failing tests) on the original PR branch `feat/interval-analysis-instrument-before-loop` identified the breaking commit:

> **`26dd5b835300b1d8c10969abbf029de684fd1c8a`**
> `[k-induction] insert entry-cond ASSUME after the IF, combine guards`

## Failing tests on master (post-#4484)

| Test |
|---|
| `regression/esbmc-solidity/github_while_loop` (uses `--k-induction`) |
| `regression/incremental-smt/incremental-18` |
| `regression/incremental-smt/incremental-24` |
| `regression/loop-invariants/github_4052` |
| `regression/loop-invariants/github_4052_1` |
| `regression/loop-invariants/kinduction_combined_strong_inv` |
| `regression/witnesses_validate/multivar_1-1` |
| `regression/witnesses_validate/multivar_1-1_global` |
| `regression/witnesses_validate/safe-program-example` |

Each is currently visible in any PR's CI on master tip — for example
PR #4823 (https://github.com/esbmc/esbmc/actions/runs/26416263076/job/77761317927)
and PR #4840 (the docs PR that merged at 19:46 BST on 2026-05-25). PRs
whose CI ran before #4484 merged (e.g. #4818, #4821, #4822, all at
~13:25 UTC) showed the same tests passing.

## Bisect log (against `regression/loop-invariants/github_4052`)

```
good: 4ea1bc4191  [python] suppress McCabe MC0001 on re.py operational model (#4820)
                  (pre-#4484 master)
bad:  741d4985d8  Merge origin/master into feat/interval-analysis-instrument-before-loop
                  (tip of #4484's branch)

bad:  b9e3e3c533  [goto-programs] don't treat dereferenced pointers as modified vars
good: 4c471b0c3e  [symex] make process-wide statics thread_local
good: 6adea00f27  Apply automatic changes
bad:  26dd5b8353  [k-induction] insert entry-cond ASSUME after the IF, combine guards
good: feba47253e  Merge origin/master ...

first bad commit: 26dd5b8353
```

Local reproduction (clean rebuild at each step):

```sh
# Pre-#4484:
git switch 4ea1bc4191 --detach && cmake . && make esbmc -j
./build/src/esbmc/esbmc regression/loop-invariants/github_4052/main.c \
    --multi-property --loop-invariant --max-k-step 2
# -> VERIFICATION SUCCESSFUL

# At parent of 26dd5b8353 (feba47253e):
# -> VERIFICATION SUCCESSFUL

# At 26dd5b8353 itself:
# -> VERIFICATION UNKNOWN  (`post loop: qa mod` becomes UNKNOWN)
```

## Root-cause analysis

The author's commit message describes the intent: fix a pre-existing soundness bug where the legacy `assume_loop_entry_cond_before_loop` inserted ASSUMEs *at* each matching branch instruction via `insert_swap`. For the IF that is the loop's exit test, that placed the ASSUME *before* the IF; after the back-edge was retargeted to the IF, the ASSUME fired on every iteration including the natural-exit one — producing vacuous UNSAT proofs (e.g. on SV-COMP `sll_of_sll_nondet_append-2`).

The fix combines all per-branch guards into one conjunction and inserts **a single** `ASSUME(entry_cond)` **after** the loop_head IF, so the back-edge keeps targeting the IF.

The new IS layout is:

```
HAVOC vars                          [inductive_step]
loop_head: IF !cond GOTO exit
ASSUME(entry_cond)                  [inductive_step]   <- new placement
body
GOTO loop_head
exit:
... post-loop assertions
```

`regression/loop-invariants/github_4052` exercises the
`--loop-invariant --multi-property` path, which depends on the loop
invariants being visible to the **forward-condition (FC) phase** for
the post-loop assertions. With the new placement, the ASSUME fires
only between the IF and the body — it never fires on the exit path
that leads to the post-loop assertions, so the FC can't use the
invariant `qa % 2 == 1` to prove `__ESBMC_assert(qa % 2 == 1)` after
the loop. Result: post-loop properties report UNKNOWN.

Source: `regression/loop-invariants/github_4052/main.c`:

```c
int main() {
  int idx;
  int qa;
  qa %= 256;
  if (qa < 0) qa *= -1;
  if (qa % 2 == 0) ++qa;
  __ESBMC_loop_invariant(0 <= qa && qa < 256);
  __ESBMC_loop_invariant(qa % 2 == 1);
  for (idx = 0; idx < 10; ++idx) { qa = (qa + 2) % 256; }
  __ESBMC_assert(qa % 2 == 1, "post loop: qa mod");   // line 27 -> UNKNOWN
  __ESBMC_assert(0 <= qa, "post loop: qa ge");
}
```

I haven't read the other 8 failing tests in detail, but `kinduction_combined_strong_inv`, `loop-invariants/github_4052_1`, and the `witnesses_validate/*` set all use `--loop-invariant` or post-loop assertions — same shape.

## Suggested fix direction

The new ASSUME needs to be reachable on the exit path too, so the FC / post-loop phase sees the entry condition. Options:

1. Insert a second `ASSUME(entry_cond)` after the `exit:` label (so post-loop reasoning can use it).
2. Insert `ASSUME(entry_cond && !cond)` on the exit edge (more precise — the exit condition gives the loop-invariant strengthening at the right point).
3. Revert this commit; address the original `sll_of_sll_nondet_append-2` vacuous-UNSAT bug differently (e.g. retarget back-edge with care so the in-body ASSUME doesn't fire on natural exit).

Likely the cleanest is (2) — augment the loop's exit edge with the entry condition.

CC the author of #4484 (Mikhail).

lucasccordeiro:
After `make_nondet_assign`'s `insert_swap`, `loop_head` ends up pointing at the original IF by iterator advance — but `instructiont::swap` exchanges `code/type/guard/targets/...` and NOT `location_number` (see `src/goto-programs/goto_program.h:444-460`). The IF arrives at its new slot carrying the slot's default `location_number=0`. `assume_loop_entry_cond_before_loop`'s `guards.find(tmp_head->location_number)` then missed every guard, the combined entry condition came out empty, and no `ASSUME` was inserted before the IF — losing the inductive-hypothesis pin that post-loop assertions need.

Iterate the `guards` map directly. Conjunction is commutative so iteration order is irrelevant. `loop_exit` becomes unused; drop from the signature.

Fixes #4846. Restores `loop-invariants/github_4052{,_1}`, `kinduction_combined_strong_inv`, and the three `witnesses_validate/*` tests listed in the issue. The two `incremental-smt/incremental-1{8,24}` failures in the same issue have a different root cause (pointer-havoc with `--add-symex-value-sets` from commit 6840d111a5) and are out of scope here.

Local: 63/63 loop-invariants, 179/179 k-induction, 15/15 termination, 21/21 witnesses pass on macOS.


mikhailramalho (approved):
LGTM

Issue #4807, opened 2026-05-24
[python] Umbrella: humaneval regression tests marked KNOWNBUG


lucasccordeiro:
Umbrella tracking issue for regression tests under `regression/humaneval/` originally marked `KNOWNBUG` — ESBMC did not produce the expected verdict on them, and they were excluded from the default regression run.

There were **71 KNOWNBUG tests** at issue creation. All expected `VERIFICATION SUCCESSFUL` and were imported from the [external HumanEval benchmark](https://github.com/openai/human-eval) (Chen et al., 2021).

By choice, this is a **single tracking issue with the full checklist inline** rather than 71 individual sub-issues. Tick each test as it is fixed, and either remove `KNOWNBUG` from its `test.desc` or delete the test if obsolete.

## Status (last reconciled against `master` on 2026-06-07)

- **20 still `KNOWNBUG`** (unresolved — listed unchecked below).
- **51 resolved**, split into:
  - **31 genuinely fixed → now `CORE`** (pass in the default run).
  - **17 re-tagged `FUTURE`** (acknowledged BMC-scalability limits, not code fixes).
  - **3 deleted as obsolete** (invalid under CPython).

## Resolved → fixed (`CORE`)

- [x] `humaneval_2-1` — `numpy.fmod` scalar support in #5213; now CORE
- [x] `humaneval_11` — strlen-bound fix; now CORE
- [x] `humaneval_12` — Optional[T] return + inline-call `is None` in #4878
- [x] `humaneval_13` — parallel tuple-unpack via RHS staging in #4873
- [x] `humaneval_17` — now CORE (was FUTURE in #4891)
- [x] `humaneval_27` — now CORE
- [x] `humaneval_28` — runtime `__python_str_join` OM in #4861
- [x] `humaneval_29` — now CORE
- [x] `humaneval_45` — Div+float-literal RHS infers float return in #4867
- [x] `humaneval_59` — now CORE
- [x] `humaneval_60` — `sum(range(EXPR))` → Gauss form in #4869
- [x] `humaneval_62` — list-return type inference in #5210; now CORE
- [x] `humaneval_69` — now CORE (was FUTURE in #4910)
- [x] `humaneval_78` — now CORE
- [x] `humaneval_79` — `__python_int_to_bin` runtime OM in #4880
- [x] `humaneval_81` — int/float list-literal promotion in #5176 (was regressed from FUTURE #4892)
[...]

PR #4859, opened 2026-05-27
[regression] make FUTURE behave as expected-fail and align fixtures

lucasccordeiro:
`testing_tool.py` listed `FUTURE` in `SUPPORTED_TEST_MODES` with the docstring _"tests known to fail due to missing implementation"_, but `FAIL_MODES` held only `KNOWNBUG`. FUTURE tests therefore ran under normal pass/fail semantics — and authors locked in their current (broken) output as the "expected" regex to make the test pass anyway. Add `FUTURE` to `FAIL_MODES` so its regex now describes the desired post-fix output and matching trips the exit-77 reclassify-to-CORE warning, mirroring KNOWNBUG. Also generalise the "passed but is marked as KNOWNBUG" warning to interpolate `test_case.test_mode`.

Audit and realign the 5 existing FUTURE fixtures against the new semantics:
- `esbmc/struct_bitfields_03` already verifies `SUCCESSFUL` and the bitfield support has caught up — **FUTURE → CORE**.
- `cstd/strxfrm_bug` and `loop-invariants/nested-1-nested_loop_invariants` produce `VERIFICATION FAILED` today and would continue to do so post-fix (strxfrm_bug: under correctly-modelled `strxfrm` Czech `ch > h` makes the assertion `strcmp(out1, out2) >= 0` provably fail; nested-1: the loop body modifies `secret` but no invariant mentions it, so `assert(secret == 50042)` is unprovable under any working nested-loop-invariant implementation). FAILED is the genuine verdict either way — **FUTURE → CORE**, regex restored to `^VERIFICATION FAILED$`.
- `esbmc-solidity/github_497_2` and `esbmc-solidity/import_2` already had correct regexes that fail to match today — **stay FUTURE unchanged**.

Re-tag `humaneval_75` `KNOWNBUG → FUTURE`: the 3-deep prime sieve to 101 is a BMC scalability dead-end, not a bug. Part of draining #4807.

### DIFF
--- a/regression/cstd/strxfrm_bug/test.desc
+++ b/regression/cstd/strxfrm_bug/test.desc
@@ -1,4 +1,4 @@
 FUTURE
 main.c
 --unwind 50 -Wno-error=implicit-function-declaration
-^VERIFICATION FAILED$
+^VERIFICATION SUCCESSFUL$
--- a/regression/cstd/strxfrm_bug/test.desc
+++ b/regression/cstd/strxfrm_bug/test.desc
@@ -1,4 +1,4 @@
-FUTURE
+CORE
 main.c
 --unwind 50 -Wno-error=implicit-function-declaration
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION FAILED$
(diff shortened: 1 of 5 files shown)


LukeW1999 (approved):
Looks good! 

I think it is worth thinking about a gradual audit to migrate some `KNOWNBUG` entries to `FUTURE` where it's really an implementation gap rather than a tracked bug.

PR #4862, opened 2026-05-27
[ci] check clang-format on changed lines only; stop auto-committing

mikhailramalho:
## Summary

- The `code-style` CI job no longer reformats whole files and pushes an "Apply automatic changes" commit. That churned untouched lines (often formatted by a different clang-format version than the pinned v11). It now only **checks** that the lines changed in the PR are clang-format-11 clean (via `git-clang-format`), and on failure posts a PR comment instead of pushing.
- New `scripts/install-format-hook.sh`:
  - default mode installs a git **pre-commit hook** that formats only changed lines on every future commit, using the same clang-format 11 as CI (downloaded and cached under `.cache/`).
  - `--fix [BASE]` mode reformats the lines this branch changed vs `BASE` (default `origin/master`) in the working tree, so a contributor can fix lines **already committed/pushed**, then commit & push.
- On a style failure the PR comment gives both remedies: `--fix` for the already-committed lines, and hook install to prevent it next time. No comment is posted when there is nothing to fix.

## Notes

- Permissions on the `code-style` job drop from `contents: write` to `contents: read` + `pull-requests: write`.
- macOS: uses the x86_64 LLVM 11.0.0 build (runs under Rosetta on Apple Silicon — there is no native arm64 prebuilt for LLVM 11).


rafaelsamenezes (approved):
Nice. I was wondering whether we should also replace (in another PR) clang-11 to whatever latest debian/ubuntu LTS clang.

lucasccordeiro (approved):

mikhailramalho:
> Nice. I was wondering whether we should also replace (in another PR) clang-11 to whatever latest debian/ubuntu LTS clang.

Yeah, this should make it easier to upgrade, as git-clang-format only runs on the changed lines. Do you know which clang-format version is shipped with the latest LTS (26.04)? I'm still on 25.10.

mikhailramalho:
Merging now to kill the bot. We can update the clang-format version later

PR #4868, opened 2026-05-27
[termination] enable interval-analysis for SV-COMP; complete the post-kind bound pass

mikhailramalho:
## Summary

Two changes to the termination / k-induction interval pipeline.

### 1. SV-COMP wrapper: enable `--interval-analysis` for termination (the actual win)

The termination command line in `esbmc-wrapper.py` did not pass
`--interval-analysis`, so the post-k-induction interval bound injection
never ran in competition (the reachability category already passed it).
Enabling it strengthens the inductive step with per-instruction interval
bounds.

**Measured on the full SV-COMP termination set (30s/benchmark):**

| | without `--interval-analysis` | with |
|---|---|---|
| raw score | 1813 | 1859 |
| correct-false | 371 | 423 |
| wrong-false | 1 | 1 |
| wrong-true | 1 | 1 |

**+52 correct-false / +46 raw, no new wrong results** (both remaining
wrongs are pre-existing). This is entirely attributable to the existing
`--interval-analysis` machinery; the wrapper simply wasn't using it for
this category.

### 2. Complete `instrument_loop_bounds_after_kind` for transitively-modified vars (sound, currently verdict-neutral)

The post-kind bound pass collected the symbols to bound only from
instructions appearing **textually** in the loop body, missing loop
variables modified only **through a call** (e.g. a state-machine global
mutated inside a callee, never named in the caller's loop body). It now
also walks k-induction's havoc block (the `x = NONDET()` inductive-step
assigns emitted before the back-edge target) and bounds those targets
too.

**Soundness:** the injected `ASSUME` is flagged
`inductive_step_instruction`, so base case and forward condition skip it
(`execution_state.cpp:235`) — only the inductive hypothesis is
tightened, never concrete BC/FC reasoning. The bound comes from the same
transparency-aware interval re-analysis the existing textual-scan bounds
use, applied to additional symbols at the same program point.

**Honest scope:** measured marginal impact on the current termination set
is **zero** — removing this change while keeping `--interval-analysis`
gives the identical 1859 / 423 result. The added bounds are real and
sound but no current benchmark's verdict depends on them: the eca cluster
needs a *relational* recurrent-set invariant (an interval box still admits
error-reaching states), and elsewhere `--interval-analysis`'s existing
per-instruction guard bounds already suffice. This commit lands it as a
sound completion of a latent incompleteness in the bound pass (it
strengthens the inductive hypothesis and becomes load-bearing for future
relational/ranking work); it is **not** the source of the +46. Reviewers
who prefer to defer verdict-neutral changes can take commit 2
(`b67f66312b`, the wrapper) alone.

## Testing

`regression/k-induction/interval_transitive_bound` pins that the bound on
a transitively-modified global appears in the transformed GOTO (a
transformation pin, not a verdict pin — there is no verdict change to
assert, by the scope note above). 182/182 k-induction, 26/26 termination,
143/143 interval-analysis regression tests pass.

### DIFF
--- a/regression/k-induction/interval_transitive_bound/main.c
+++ b/regression/k-induction/interval_transitive_bound/main.c
@@ -1,0 +1,55 @@
+/* Post-k-induction interval bound for a TRANSITIVELY-modified loop
+ * variable.
+ *
+ * `state` is a global that main's loop modifies only indirectly, via
+ * the call to step(). It never appears textually in main's loop body
+ * (the body just reads `in` and calls step(in)). The interval domain
+ * proves state stays in [0,2] (step assigns it only the constants
+ * 0/1/2 under guards).
+ *
+ * k-induction's make_nondet_assign havocs state (it's in the loop's
+ * transitively-modified set), emitting `state = NONDET()` in the havoc
+ * block before the loop head. instrument_loop_bounds_after_kind must
+ * then pin state to its interval at the inductive-step loop head.
+ * Before the fix, that pass collected only symbols appearing textually
+ * in the loop body, so the callee-modified `state` was missed and the
+ * inductive step ran from an arbitrary state value. The fix also
+ * collects the targets of the havoc block, so state gets bounded.
+ *
+ * The test inspects the transformed GOTO program: an
+ * ASSUME(0 <= state <= 2) must be present (the post-havoc bound on the
+ * transitively-modified variable). */
+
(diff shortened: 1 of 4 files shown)


mikhailramalho:
Annoyingly, it's the same result as master, so it should be safe to merge. This only improves our invariants.

PR #4876, opened 2026-05-27
refactor smt conversion split from smt_conv

sergillam:
- extract IEEE754 conversion logic from smt_conv.cpp into smt_fp_conv.cpp
- add shared FP rounding helpers, and wire the new file in CMake.


mikhailramalho (approved):
It's just moving code around, right? If so, LGTM

sergillam:
> It's just moving code around, right? If so, LGTM

yes

PR #4896, opened 2026-05-28
refine IEEE754 enclosure helpers and rounding-mode conversion in smt_fp_conv

sergillam:
- refines IEEE754 handling in smt_fp_conv.cpp by centralizing weak-enclosure/epsilon helper logic and tightening directed-rounding paths;
-  conversion with explicit supported-mode handling and conservative fallback behavior, while keeping scope limited to solver FP internals.

### DIFF
--- a/src/solvers/smt/smt_fp_conv.cpp
+++ b/src/solvers/smt/smt_fp_conv.cpp
@@ -376,6 +376,53 @@ smt_astt smt_convt::apply_ieee754_semantics(
 {
   if (this->options.get_bool_option("ir-ieee"))
   {
+    auto weak_enclosure_return =
+      [this, &real_result, &fbv_type](const char *reason) -> smt_astt
+    {
+      log_warning(
+        "using weak IEEE754 enclosure ({}), exp={}, frac={}",
+        reason,
+        fbv_type.exponent,
+        fbv_type.fraction);
+      smt_sortt rs = mk_real_sort();
+      smt_astt ra_lo = mk_fresh(rs, "ra_lo_weak::", nullptr);
+      smt_astt ra_hi = mk_fresh(rs, "ra_hi_weak::", nullptr);
+      assert_ast(mk_le(ra_lo, real_result));
+      assert_ast(mk_le(real_result, ra_hi));
+      assert_ast(mk_le(ra_lo, ra_hi));
+      return real_result;
+    };
+
+    auto select_nearest_eps =
+      [this, &fbv_type](smt_astt &eps_rel, smt_astt &eps_abs) -> bool
+    {
+      const auto double_spec = ieee_float_spect::double_precision();
+      const auto single_spec = ieee_float_spect::single_precision();
+      if (
+        fbv_type.exponent == double_spec.e &&
+        fbv_type.fraction == double_spec.f)
+      {
+        eps_rel = get_double_eps_rel();       // 2^-53
+        eps_abs = get_double_min_subnormal(); // 2^-1074
+        return true;
+      }
+      if (
+        fbv_type.exponent == single_spec.e &&
+        fbv_type.fraction == single_spec.f)
+      {
+        eps_rel = get_single_eps_rel();       // 2^-24
+        eps_abs = get_single_min_subnormal(); // 2^-149
+        return true;
+      }
+      return false;
+    };
+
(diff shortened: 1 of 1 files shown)


mikhailramalho (commented):
LGTM.

I did a bunch of similar improvements on the camada version of our fp bitblast, you can take a look at: https://github.com/mikhailramalho/camada/blob/master/src/camadafp.cpp

the tl;dr is that it was very hard to improve performance, and the main remaining hotspot is dominated by the fallback algorithms themselves,
  especially `round(...)`, `unpack(...)`, and `mkLeadingZeros(...)`, not the operations.

mikhailramalho:
Summary of my notes:

###  What worked (kept, in §5 "First FP Fallback Churn Reduction Pass")

  The only FP-bitblast improvement that benchmarked well enough to land:

  - Solver-generation-cached FP special values — NaN, ±0, ±inf, ±1 are constructed once per solver generation and reused instead of rebuilt on each call.
  - Solver-generation-cached RM literals (the rounding-mode constants).
  - Helper-local singleton use of tiny BV constants in the FP path.
  - Targeted helper simplifications that benchmarked well:
    - improved mkLeadingZeros(...) (binary-search-like form)
    - reduced round(...) constant churn
    - mkFPIsInfinite(...) rewritten as equality against cached ±inf
    - small mkFPSqrtImpl() cleanups — reusing sgn_eq_1 instead of rebuilding mkIsNeg(...), and reusing the already-built zero-significand constant for rem_eq_0

  Benchmark status the doc records: non-fp.rem paths got materially better than earlier branches on the Bitwuzla+encoded-FP synthetic benchmark. fp.rem is slower than v0.3 because the current branch uses a more
  complete IEEE-style encoding there — that's a correctness-vs-cost tradeoff accepted intentionally.

###  What tried and reverted

  Five reverted attempt clusters, all called out as "not stable wins":

  §6 FP special-value cache variants beyond the kept design — alternative cache shapes and more eager per-call initialization for the NaN/±0/±inf/±1 constants. The kept solver-generation cache + local singleton
  was already the sweet spot; extra layers didn't pay back.

  §7 mkRoundingDecision(...) refactors — RM identity fast-path dispatch, passing cached RM literals into more sites, and algebraic rewrites of the round-to-even increment predicates (e.g., De Morgan → direct and
   forms). All consistently regressed fp_construct — the extra branches or changed bool/BV structure were worse than the baseline.

  §8 mkLeadingZeros(...) structural rewrites after the kept version — direct bit-concat encodings of the count, small-constant micro-opts inside the loop, array-based zero-literal shortcuts for tiny widths. All
  regressed against the current binary-search-like implementation.

  §9 unpack(...) call-site reuse / early expansion — moving more extraction/classification work into callers, local unpack-reuse in sqrt, fma, etc. Verdict: these increased unconditional AST construction more
  than they saved on repeated helper calls. Camada eagerly constructs formulas, so there's no runtime short-circuit benefit — that's the structural reason these don't work.

  §10 FP predicate rewrites that weren't stable wins — mkFPIsZero(...) as equality to cached ±0, and local mkFPIsNaN(...) / mkFPIsInfinite(...) call-site rewrites using extracted pieces. The mkFPIsInfinite
  global rewrite was kept; the mkFPIsZero equality and the call-site isNaN/isInfinite variants were reverted because some increased local AST size despite looking simpler in source form.

###  Remaining hotspots the doc names (§4)

  - The fallback FP layer is structurally expensive — many intermediate wrapper nodes, heavy recursive term construction.
  - After the churn-reduction pass, the remaining hotspot is dominated by the algorithms themselves: round(...), unpack(...), mkLeadingZeros(...).
  - The dominant remaining benchmark hotspot is fp.rem, which is intentional (correctness over speed).

###  What the doc recommends next (near-term §3)

  - Reduce helper-level node count in round, unpack, mkLeadingZeros without front-loading more eager work.
  - Keep tiny FP/RM constants on dedicated singleton paths.
  - Avoid broad hash-based caches unless a benchmark shows a stable win.
  - Treat fp.rem separately from the rest of the fallback layer.
  - "Lazy lowering only as a later exploratory step, not as an initial optimization target."
  - Expected payoff: medium, concentrated in the heavy paths. Risk: medium to high.

mikhailramalho:
I also rewrote fp.rem, esbmc's version is not quite correct

lucasccordeiro (approved):

PR #4916, opened 2026-05-28
[contracts] free pointer-param backing on enforce-contract exit

### LINKED ISSUE #4908: `--enforce-contract` wrapper leaks auto-allocated pointer parameters under `--memory-leak-check`
When `--enforce-contract <fn>` is used on a function with a pointer parameter, ESBMC's synthesized wrapper allocates a buffer for the parameter (so the body can dereference it) but never frees it before returning. Enabling `--memory-leak-check` then reports a CWE-401 *forgotten memory* violation attributed to the user's function — even though the user's body cannot cause or fix the leak. The two features should compose; today they cannot.

## Minimal example

`read_thru.c`:

```c
__ESBMC_contract
int read_thru(const int *p) {
    __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(*p)));
    __ESBMC_ensures(__ESBMC_return_value == *p);
    __ESBMC_assigns();
    return *p;
}
```

Run:

```sh
esbmc --function read_thru --enforce-contract read_thru --memory-leak-check read_thru.c
```

## Expected

`VERIFICATION SUCCESSFUL`. The contract declares `__ESBMC_assigns()` (empty frame), so the function provably cannot store or escape `p`; the buffer the wrapper allocates is wrapper-local and should be freed on wrapper exit.

## Actual

```
Violated property:
  file read_thru.c line 7 column 1 function read_thru
  dereference failure: forgotten memory: dynamic_1_array
  CWE: CWE-401

VERIFICATION FAILED
```

The violation is reported at the closing brace of `read_thru`, attributing the leak to the user's function. Nothing in the body or contract can fix it — the leaked allocation is the one `--enforce-contract`'s wrapper made itself.

Dropping just `--memory-leak-check` (keeping `--enforce-contract`) produces `VERIFICATION SUCCESSFUL`, confirming the failure is solely the wrapper/leak-check interaction.

## Impact

`--memory-leak-check` becomes unusable on every pointer-taking function verified via `--enforce-contract`, which is the main signature class the feature targets. Tools building on contracts (e.g. contract-inference pipelines) cannot enforce leak-freedom of the function under test without first filtering out every pointer-param target.

## Suggested fix

In the wrapper-emission code under `src/goto-programs/contracts/`, emit a matching `free()` for each auto-allocated pointer parameter at every wrapper exit point (or use a stack allocation when `sizeof(*p)` is a compile-time constant, which is the common case for primitive pointers). The fix is local to the synthesized wrapper; no user-side change or contract-language change is needed.

## Environment

- ESBMC 8.3.0 (64-bit x86_64 Linux)
- No solver flag passed (defaults to z3)

Yiannis128:
The synthesised `--enforce-contract` wrapper mallocs backing storage for each pointer parameter (so the body can dereference it) but never freed it; `--memory-leak-check` then reported CWE-401 forgotten-memory against the user's function. Track every wrapper-internal heap allocation (is_fresh requires-side mallocs and `add_pointer_validity_assumptions` mallocs) in `wrapper_heap_ptrs`, and emit `code_free2tc(p)` for each just before the wrapper's RETURN — after the ensures ASSERT, which may still need to dereference the buffer.

Fixes #4908

### DIFF
--- a/regression/function_contract/memleak_check_enforce_addrof_pass/main.c
+++ b/regression/function_contract/memleak_check_enforce_addrof_pass/main.c
@@ -1,0 +1,16 @@
+/* memleak_check_enforce_addrof_pass: covers the address-of surface form of
+ * __ESBMC_is_fresh combined with --memory-leak-check. The bare-pointer form
+ * is covered by memleak_check_enforce_pass; this case exercises the
+ * address-of peel branch (stripped = &p -> ptr_var = p) at the same time as
+ * the wrapper-free fix for issue #4908.
+ */
+__ESBMC_contract
+int read_thru_addrof(const int *p)
+{
+    __ESBMC_requires(__ESBMC_is_fresh(&p, sizeof(*p)));
+    __ESBMC_ensures(__ESBMC_return_value == *p);
+    __ESBMC_assigns();
+    return *p;
+}
+
+int main(void) { return 0; }
--- a/regression/function_contract/memleak_check_enforce_addrof_pass/test.desc
+++ b/regression/function_contract/memleak_check_enforce_addrof_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function read_thru_addrof --enforce-contract read_thru_addrof --memory-leak-check
+^VERIFICATION SUCCESSFUL$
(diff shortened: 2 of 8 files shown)


LukeW1999 (approved):
Looks good. If it is not rush, I think we can wait the CI becomes all green before merging it.

lucasccordeiro (approved):

PR #4919, opened 2026-05-28
[termination] ranking-function checker with supporting-invariant synthesis

mikhailramalho:
## Summary

A ranking-function termination prover that runs before k-induction's havoc. Proves termination on most "obvious" SV-COMP termination benchmarks by:

- Recognizing several loop body shapes (straight-line, single if/else, N sequential if/else blocks).
- Building a difference measure from the loop guard, including from `&&` conjunctions of relational atoms.
- Synthesizing a supporting invariant from constant pre-loop seeds AND from dominating path-conditions.
- Substituting loop-invariant pointer dereferences with fresh scalars (with cross-pointer disjointness check for soundness).
- Discharging bounded/decrease obligations under a faithful simultaneous body transition + path conditions.

Plus a fix for two longstanding bugs in `--full-inlining` (orthogonal to the ranking work, kept as its own commit).

## Pipeline

1. **Loop recognition** (`recognize_loop`). Accepts:
   - Straight-line scalar bodies (Shape A).
   - Sequential `if`/`else` blocks of straight-line arms, capped at 16 enumerated paths (Shape B + multi-if extension).
   - Guards that are a single relational atom or a top-level `&&` conjunction.
   - Per-path conditions captured by applying prior-block assignments through the IF guard (so post-assignment guards conjoin correctly with pre-state obligations).

2. **Measure derivation** (`measure_candidates_from_guard` / `measure_from_relational`). For a single relational atom: one (m, L) candidate computed in int64 (overflow-safe for <=32-bit operands; wide operands accepted only if they are constants whose value fits in int32). For an `&&` conjunction: one candidate per conjunct; success on any conjunct proves termination.

3. **Faithful body transition** (`apply_body` / `subst_parallel`). Parallel substitution over the loop body: each assignment's RHS is resolved through the post-state computed so far, then a single simultaneous substitution into the measure. Replaces the earlier sequential substitution which mishandled bodies like `{a = a - b; b = a}`.

4. **Supporting invariant** (`synthesize_invariant`). Seeds bound atoms (v >= c / v <= c) from:
   - Constant pre-loop assignments (`d1 = 73`).
   - Dominating `IF guard GOTO L` instructions (a bounded forward reachability check picks the branch that leads to the loop head).
   Atoms are widened to int64 via `widen_arith` (which returns nil for opaque NONDET sideeffects, conservatively dropping unsafe atoms). A monotone fixpoint keeps only atoms inductive under every body path (with each path's condition conjoined into its preservation obligation).

5. **Pointer-invariant deref substitution** (`compute_safe_derefs`). When `*p` appears in a loop and `p` is a loop-invariant pointer with distinct-allocation provenance (malloc/realloc/alloca/`&local`, reached through a symbol/typecast chain in the function prefix), substitute each `*p` with a fresh scalar symbol named `$rank_deref$<addr>$<sym>` (the leading `$` is not a legal C identifier and the hex address makes the synthetic symbol unique). A cross-pointer disjointness check rejects two pointers reaching the same memory cell. The prefix scan is dominance-aware (uses the same `reaches_head` arbitration as the seed walk) and rejects FUNCTION_CALL/OTHER/back-edge prefix instructions.

6. **Self-recursion** (`prove_self_recursion_terminates`). For functions calling themselves directly, prove a parameter-based ranking argument. Mutual recursion and function-pointer calls force UNKNOWN.

7. **Discharge** (`prove_loop_terminates`). For each ranking candidate, check bounded `inv ∧ guard ∧ (m < L)` UNSAT and decrease-on-every-feasible-path `inv ∧ guard ∧ path_cond ∧ ¬(m' < m)` UNSAT. Success on any candidate proves termination.

## `--full-inlining` correctness fix

Independent of the ranking work, the `goto_inline` overload called by `--full-inlining` had two latent bugs that have been in mainline ESBMC for years:

1. **Malformed parameter DECLs.** `parameter_assignments` emitted synthetic parameter declarations with instruction type `OTHER` but code `code_decl2t`. Symex's `OTHER` dispatch doesn't handle `code_decl2t` and throws "unexpected statement: code_decl". Fixed by using `make_decl()` instead of `make_other()` at both emission sites.

2. **Wrong cleanup target.** The post-inlining cleanup was `if (it.first != "main")`, but the user's `main` is keyed by its mangled symbol identifier `c:@F@main` while the synthetic entry point that symex actually runs is keyed by the bare string `__ESBMC_main`. The comparison matched nothing in the map, silently clearing every body including the entry, leaving symex with an empty program. Fixed to compare against `"__ESBMC_main"`, matching the sibling `dest`-taking overload's intent.

`--full-inlining` is now functionally correct on simple programs. Note: enabling `--full-inlining` for the termination wrapper would still need additional work to handle the inlined library scaffolding (the atexit handler dispatch loop, etc.); the flag is left opt-in.

## SV-COMP termination measurement

Full SV-COMP termination set (2413 benchmarks, 30s timeout, 32 workers), comparing against master:

| Metric | Master | This PR | Δ |
|---|---|---|---|
| Raw score | 1893 | **2021** | **+128** |
| Meta score | 1386.27 | **1698.75** | **+312.48** |
| Correct-true | 785 | **849** | +64 |
| Wrong-true | 1 | 1 | 0 |
| Wrong-false | 1 | 1 | 0 |
| ControlFlow-Term CT | 41 | **74** | +33 (sn_i 0.38 → 0.61) |
| Heap-Term CT | 9 | **25** | +16 (sn_i 0.09 → 0.25) |
| Other-Term CT | 722 | **735** | +13 |
| BitVectors-Term sn_i | 0.91 | **1.03** | +0.12 |

**64 cumulative wins, 0 regressions, 0 new wrongs.** The 2 pre-existing wrong-true (lirc_dev) and wrong-false (turbografx) are unchanged.

The remaining still-UT (expected-true UNKNOWN) count across all SV-COMP categories is **4** — the ranking-checker side of termination is essentially complete. Remaining headroom is on the *non-termination* side (595 expected-false UNKNOWN, dominated by eca-rers2012), which is a separate technique (recurrent-set detection) and out of scope here.

## Soundness review

Codex reviewed three times across the work: after the supporting-invariant design, after the multi-if/`&&` work, and after the cumulative branch. Two real soundness gaps were caught and fixed:

- **Q3 widen_arith landmine**: NONDET sideeffects entering an int64 atom comparison would stall the solver. Fixed by returning nil for non-arithmetic nodes and skipping affected atoms.
- **Path-condition substitution against prior-block assigns**: a later IF guard must be evaluated after the prior block's assignments, not at body entry — otherwise a feasible path appears infeasible and its decrease obligation is vacuously discharged (potential wrong-true).

The final state was signed off as **good to ship** with no remaining blockers, after one extra regression test pinning a cross-feature interaction (`ranking_multi_if_cast_path_cond`).

## Commits (12)

```
[goto-programs] fix two longstanding --full-inlining defects
[termination] regression for multi-if path-cond through widening cast
[termination] peel typecasts and accept wide-int32-constant operands
[termination] multi-if bodies, && guards, and path-condition refinement
[termination] substitute loop-invariant pointer derefs with scalars
[termination] accept loops with a single in-body if/else of straight-line arms
[termination] path-condition seeds and NONDET-safe atom widening
[termination] supporting-invariant synthesis for the ranking checker
[termination] build faithful simultaneous loop-body transition
[termination] reject >32-bit guard operands in ranking measure
[termination] prove self-recursive termination by a ranking argument
[termination] linear ranking-function check for scalar counter loops
```

## Regression tests (18)

Each new feature pins both a positive case (it should certify) and, where applicable, a negative soundness case (it must NOT certify under a slight non-terminating variant). The full ranking regression family runs in ~3s.


mikhailramalho:
This PR:
```
Statistics:           2395 Files
  correct:            1456
    correct true:      913
    correct false:     543
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:             938
  Score:              2353 (max: 3816)
```
Master:
```
Statistics:           2395 Files
  correct:             764
    correct true:      764
    correct false:       0
  incorrect:             3
    incorrect true:      3
    incorrect false:     0
  unknown:            1628
  Score:              1432 (max: 3816)
  ```

lucasccordeiro (approved):
Nice!

PR #4875, opened 2026-05-27
[clang-c-frontend] implement __atomic_compare_exchange{,_n} (#4436)

lucasccordeiro:
Model `__atomic_compare_exchange` and `__atomic_compare_exchange_n` under `__ESBMC_atomic_begin/_end` with the GCC strong-CAS contract; both were previously NOPs, so any CAS-based primitive silently left `*ptr` unchanged and the return value unconstrained — the direct cause of the SV-COMP libvsync `hclhlock` false alarm.

Also drops a `value_sett::assign` assert that fired when two TUs declare structs with the same tag but different members (e.g. glibc `struct tm` vs. ESBMC's operational `time.h`), since the libvsync repro cannot even reach the CAS bug otherwise, and adds a focused CORE regression pinning the success/failure contract for both variants.

Fixes #4436.

PR #4922, opened 2026-05-28
[clang-c-frontend] implement __atomic_fetch_{add,sub,or,and,xor,nand}

lucasccordeiro:
Implements `__atomic_fetch_{add,sub,or,and,xor,nand}` in
`clang_c_adjust_polymorphic_functions.cpp`. These were silently
unimplemented: the outer dispatch matched only `__sync_fetch_and_*`,
so the symex bound every `__atomic_fetch_*` call's return value to a
`nondet_symbol` and left `*ptr` unchanged — the same silent-NOP shape
that #4875 fixed for the strong-CAS family. With the fix, the SSA
trace shows the producer's `g_buf[curr] = v` write going to the
expected slot instead of a nondet index.

Also fixes a latent bitwise-vs-logical mistake in the AND/OR cases:
`exprt("and"/"or")` constructs a logical operator, which trips the
non-`bool` operand assertion in `migrate.cpp` once the
`__atomic_fetch_*` path actually executes the inner dispatch. Both
operations are bitwise per the GCC spec.

Adds a focused CORE regression at
`regression/esbmc-unix/github_4435_atomic_fetch_add/` pinning the
contract for add/sub/or/and/xor (each: `*ptr` updated, OLD value
returned).

Refs #4435 (libvsync `bounded_mpmc_check_full`). Note: the
SV-COMP libvsync benchmark still fails after this patch because its
deq path doesn't use `fetch_add` — the remaining false alarm comes
from a separate value-set tracking issue when the per-slot write
happens through a function-call indirection (producer thread 2's
SSA collapses the `q->buf[curr % q->size] = v` write to a no-op
copy). That investigation is in flight; the `KNOWNBUG` regression
from #4921 stays in place until it is fixed.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/26617452345.

PR #4930, opened 2026-05-29
[python] fix json crash on aliased/expr list element type inference

### LINKED ISSUE #4909: [python] JSON assert in GOTO converter: derived list-init element + indexed mutation in for-loop
**ESBMC version:** 8.3.0 64-bit aarch64 macos
**Frontend:** Python
**Platform:** macOS 25.5.0 (Darwin / aarch64)

## Symptom

ESBMC aborts during GOTO conversion with a `nlohmann::json::operator[]` C-assert when a list is initialised with an element coming from a *derived* module-level binding (i.e. assigned from another name or from a non-literal expression) **and** that list is then mutated by indexed assignment inside a `for` loop.

```
$ esbmc --unwind 2 repro.py
Parsing repro.py
Converting
Assertion failed: (it != m_data.m_value.object->end()), function operator[], file json.hpp, line 2147.
```

The crash is at GOTO-IR conversion time, before any solver is invoked, so it is a frontend bug rather than a solver one.

## Minimal reproducer (5 lines, no imports)

```python
Y = 1
X = Y
a = [X]
for i in range(1):
    a[i] = -1
```

This crashes deterministically with `esbmc --unwind 2 repro.py` on 8.3.0. The 4-line variant where the indirection is an expression instead of an alias also reproduces:

```python
X = 1 + 0
a = [X]
for i in range(1):
    a[i] = -1
```

## Variants that do **not** crash (delta-bisected)

Each of the following is verified `VERIFICATION SUCCESSFUL`, isolating the trigger to the three-way combination above:

1. **Bare literal in the list-init**, no derivation chain:

    ```python
    X = 1                       # bare literal, no derivation
    a = [X]
    for i in range(1):
        a[i] = -1               # OK
    ```

2. **Single indexed mutation, no `for` loop**:

    ```python
    X = 1 + 0
    a = [X]
    a[0] = -1                   # OK
    ```

3. **`for` loop without indexed mutation**:

    ```python
    X = 1 + 0
    x = X
    for i in range(1):
        x = -1                  # OK
    ```

All three triggers must be present:
- list-element drawn from a *derived* binding (alias or expression — not a bare literal),
- subscript-assignment (`a[i] = ...`) inside a `for` loop.

## Context

Hit while building a doubly-linked-list stub for a vLLM kv-cache target. Original harness used module-level `K = 4; HEAD = K; TAIL = K + 1; NIL = -1` for sentinel readability, and an `__init__`-style `prev = [HEAD, 0, 1, 2]` + `for i in range(K): if i < n: prev[i] = NIL` pattern. Same crash. Workaround: hard-code the sentinels as literals in the list-init, which is what the harness currently ships with.

## Suggested triage

Looks like an AST-to-JSON converter table lookup that fails to find a key when the list element's RHS is a name resolved late, only manifesting once the same list slot is subsequently subscript-mutated inside a loop (which appears to trigger a different key path in the converter). Reproduces reliably; happy to share the bisect transcript or the full harness on request.

lucasccordeiro:
Fixes #4909.

## Problem

ESBMC aborted during GOTO conversion with a `nlohmann::json::operator[]` assertion (`json.hpp:2147`) when a list was initialised with an element drawn from a *derived* binding (an alias or an expression) and then mutated by indexed assignment inside a `for` loop:

```python
Y = 1
X = Y
a = [X]
for i in range(1):
    a[i] = -1
```

## Root cause

`type_handler::get_typet`'s `Name` branch resolved the type via `get_typet(var["value"]["value"])`, assuming the variable's RHS was a `Constant` (which carries a nested `"value"`). For an alias (`X = Y`, RHS is a `Name`) or an expression (`X = 1 + 0`, RHS is a `BinOp`), `var["value"]` has no `"value"` key, so `operator[]` asserted. A bare literal (`X = 1`) has a `Constant` RHS, which is why only the derived-binding variants crashed.

## Fix

- Resolve the type from the RHS node itself (`get_typet(var["value"])`) instead of `var["value"]["value"]`.
- Walk alias chains (`x = y = … = literal`) iteratively, with a visited-set guard so cyclic aliases (`a = b; b = a`) cannot recurse without bound (they now report a clean "Variable not defined" error instead of overflowing).
- Add a `BinOp` case to `get_typet` (previously unhandled), returning the operand type and promoting to `float` when either side is float.

## Validation

- Both issue reproducers (alias and expression) now reach `VERIFICATION SUCCESSFUL`; the bare-literal variant still works; alias chains `X = Y = Z` resolve; cyclic aliases no longer hang.
- Verified under both Bitwuzla and Z3.
- Adds `regression/python/github_4909` (SUCCESSFUL) and `regression/python/github_4909_fail` (FAILED).
- Ran the `get_typet`-affected regression subset (~500 list/type-inference/range/for tests): no regressions (the only non-pass is `list_pop11_nondet`, which requires the Boolector solver not built in this environment — fails identically on master).

### DIFF
--- a/regression/python/github_4909/main.py
+++ b/regression/python/github_4909/main.py
@@ -1,0 +1,15 @@
+# A list initialised with an element from a *derived* binding (alias or
+# expression) and then mutated by indexed assignment inside a for loop
+# crashed the GOTO converter with a json operator[] assertion: get_typet
+# assumed the aliased RHS was a Constant carrying a nested "value" (#4909).
+Y = 1
+X = Y
+a = [X]
+for i in range(1):
+    a[i] = -1
+assert a[0] == -1
+
+# Expression (BinOp) RHS variant from the same report.
+b0 = 1 + 0
+b = [b0]
+assert b[0] == 1
--- a/regression/python/github_4909/test.desc
+++ b/regression/python/github_4909/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--unwind 2
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_4909_fail/main.py
+++ b/regression/python/github_4909_fail/main.py
@@ -1,0 +1,8 @@
+# Companion to github_4909: the mutated value is resolved correctly, so a
+# wrong expected value must be caught.
+Y = 1
+X = Y
+a = [X]
+for i in range(1):
+    a[i] = -1
+assert a[0] == 999
--- a/regression/python/github_4909_fail/test.desc
+++ b/regression/python/github_4909_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--unwind 2
+^VERIFICATION FAILED$
--- a/src/python-frontend/type_handler.cpp
+++ b/src/python-frontend/type_handler.cpp
@@ -740,6 +740,19 @@ typet type_handler::get_typet(const nlohmann::json &elem) const
       return get_typet(elem["operand"]);
     }
 
+    // Handle Python AST BinOp node (e.g., 1 + 0): the result type follows the
+    // operands, promoting to float when either side is float (#4909).
+    if (
+      elem["_type"] == "BinOp" && elem.contains("left") &&
+      elem.contains("right"))
+    {
+      typet lhs_type = get_typet(elem["left"]);
+      typet rhs_type = get_typet(elem["right"]);
+      if (lhs_type.is_floatbv() || rhs_type.is_floatbv())
+        return double_type();
+      return lhs_type;
+    }
+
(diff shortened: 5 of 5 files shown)


lucasccordeiro:
Superseded by #4923 ("Fix json abort in get_typet on derived list elements"), which was merged and closed #4909 with the same core fix (`get_typet(var["value"])` + BinOp handling). Closing this duplicate.

Minor note for a possible follow-up: #4923's `Name` branch recurses via `get_typet(var["value"])` directly, so a cyclic alias (`a = b; b = a`) could recurse without bound. This branch added an iterative visited-set guard for that edge case — happy to send it as a small hardening PR if you think it's worth it.

PR #4940, opened 2026-05-29
[CI] add WitnessMap pre-validation step for hand-crafted witness tasks

XLiZHI:
Added support for hand-crafted witness task sets (ReachSafety-CorrectnessWitnesses, ReachSafety-ViolationWitnesses, Termination-ViolationWitnesses) to the ESBMC and CPAChecker witness validation configurations. ~~Introduced WitnessMap as a pre-validation step that runs in an isolated venv and writes witness files directly into the BenchExec results directory for downstream validators to consume.a~~


lucasccordeiro (commented):
## Review

Thanks for wiring up the ValidationCrafted witness tasks. I verified the referenced set files: `ViolationWitnesses.set` and `CorrectnessWitnesses.set` **do exist** on the sv-benchmarks `main` branch (the per-subcategory names like `ViolationWitnesses-ControlFlow.set` return 404). Their contents glob `*.witness-validation.yml`, where `ViolationWitnesses.set` mixes `validation-crafted/` (reach) and `termination-restricted-15/` (termination) tasks — which is exactly why the ESBMC side splits them into two rundefinitions by property file. That part is correct.

### What's done well

- **ESBMC rundefinitions are structured correctly.** The new `SV-COMP26_unreach-call_witnesses` / `SV-COMP26_termination_witnesses` blocks deliberately omit `<requiredfiles>` and the `--witness` option, so `handle_witness_of_task()` pulls the witness from the task definition. The global `--validate-violation-witness` / `--validate-correctness-witness` and `-s kinduction` are correctly inherited.
- **Property-file split is right.** Using the same `ViolationWitnesses.set` with `unreach-call.prp` vs `termination.prp` relies on BenchExec only generating runs for tasks that declare the given property — the natural way to demux that mixed set.
- **`PYTHONPATH` is scoped to the ESBMC branch** of `witness.sh` and uses an absolute path, so it survives the later `cd $HOME/output-action`.

### Findings

**🔴 High — CPAChecker additions are inconsistent and likely non-functional**

The new tasks were inserted **inside the existing `SV-COMP26_unreach-call` rundefinition** in `cpachecker-{correctness,violation}-v2.xml`, which carries:

```xml
<requiredfiles>$HOME/witness-files/${rundefinition_name}/${taskdef_name}/witness.yml</requiredfiles>
<option name="-w">$HOME/witness-files/${rundefinition_name}/${taskdef_name}/witness.yml</option>
```

For hand-crafted tasks, no such file exists under `$HOME/witness-files/` (that directory only holds witnesses ESBMC produced in the prior run, moved in by `setup()`). So BenchExec will **skip these tasks** (requiredfiles unsatisfied), and even if it didn't, `-w` would point at the wrong witness instead of the crafted one carried by the task. This contradicts the ESBMC side, which correctly uses *separate* rundefinitions. CPAChecker should get its own `*_witnesses` rundefinitions (no `requiredfiles`/`-w` override) mirroring the ESBMC structure. Is validating crafted witnesses with CPAChecker in scope? If so, this is a bug.

**🟡 Medium — bundled `tools/esbmc.py` appears to be a verbatim copy of upstream `benchexec.tools.esbmc`**

`cmdline()` and `determine_result()` match the current upstream module (same `handle_witness_of_task`, same result mapping, same SPDX header), so as written it provides no customization and the rationale for forking + `tool="tools.esbmc"` is unclear/undocumented:

- If the goal is **version-pinning** against the runner's system BenchExec, note the copy still does `from benchexec.tools.sv_benchmarks_util import handle_witness_of_task, TaskFilesConsidered` — so it only helps if the runner's BenchExec already provides those helpers (added relatively recently). Please confirm the deployed `/usr/bin/benchexec` exposes them, or the import will crash.
- Cleaner alternative: keep `tool="esbmc"` and pass `--tool-directory <dir>` to BenchExec instead of the `PYTHONPATH` export + module rename. That avoids forking the module and the portability issue below.
- At minimum, add a one-line comment in the XML / a short README in `tools/` explaining why it's bundled, so it doesn't silently drift from upstream.

**🟡 Medium — `tool="tools.esbmc"` makes the XMLs non-portable**

The XMLs now only work when `PYTHONPATH` includes the `witness/` dir (set solely in `witness.sh`). Running these benchmark defs manually/locally will fail with `ModuleNotFoundError: tools.esbmc`. The `--tool-directory` approach above also fixes this. Also note the generic package name `tools` on `PYTHONPATH` could shadow another `tools` module.

**🟢 Low**

- **Workflow dropdown gap:** the new task names were added to the `task:` choice list, but the new ESBMC runset names (`SV-COMP26_unreach-call_witnesses`, `SV-COMP26_termination_witnesses`) were **not** added to the `runset:` choice list — so they're only reachable via `mode: Task` or `Full`, not `mode: RunSet`.
- **PR description:** the body is almost entirely struck-through and ends with a stray "...consume.a", so it doesn't document the final approach. Worth rewriting before merge for the record.
- **Labels:** please add at least one (e.g. `CI`/`build`) per repo convention.

### Recommendation

Request changes. The ESBMC-side wiring is sound, but the **CPAChecker additions look ineffective** (High) and the **bundled tool module needs justification + a portability fix** (Medium). I'd like the CPAChecker question answered before merge.

Two things to confirm on the runner that I couldn't check from the diff: (1) that the deployed BenchExec exposes `handle_witness_of_task`/`TaskFilesConsidered`, and (2) that ESBMC actually validates the YAML 2.0 witnesses these crafted tasks carry (otherwise expect mass `unknown`/`error`, which may be the intended signal anyway).

XLiZHI:
Move ReachSafety-CorrectnessWitnesses and ReachSafety-ViolationWitnesses out of the main SV-COMP26_unreach-call rundefinition in the cpachecker witness XMLs into separate SV-COMP26_unreach-call_witnesses rundefinitions, matching the structure already used in the esbmc witness XMLs.

The remaining issues mentioned do not exist.

lucasccordeiro (approved):

PR #4944, opened 2026-05-29
[util] irep2 migration Phase 2.4: eliminate core migrate-seam round-trips

lucasccordeiro:
## What

Implements the achievable scope of **Phase 2.4 — core migrate-seam
round-trip elimination** (`docs/irep2-migration.md` Part II §6). Removes
`migrate_*`/`migrate_*_back` round-trips from the verifier core
(pointer-analysis / solver / IREP2), reading the IREP2 source of truth
directly. Strictly behaviour-preserving.

## Commits

| # | Commit | What |
|---|--------|------|
| 15 | value_set off the seam | `value_sett::do_function_call` reads `to_code_type(symbol.get_type2())` directly instead of `migrate_type(to_code_type(symbol.get_type()))` + a `dynamic_cast`; also drops a dead `migrate_type_back` in the NULL-pointer value-set path (the local was computed but unused). value_set.cpp now has zero `migrate_*` sites. |
| 17 | smt_conv native model builders | `smt_convt::get_by_value` builds model `fixedbv`/`floatbv` constants from the IREP2 type directly (`fixedbv_spect(to_fixedbv_type(type))` + `set_value`, `ieee_float_spect(to_floatbv_type(type))` + `unpack`) instead of staging a legacy `constant_exprt(binary, value, migrate_type_back(type))` parsed via `from_expr`. |
| 19 | prune stale legacy includes | Remove 4 stale `std_*` includes from `irep2/{irep2_expr,irep2_type,irep2_guard,irep2_dispatch}` (they use the `type2tc` `to_*_type` overloads, not the legacy ones) and the now-unused `<util/migrate.h>` from value_set.cpp. |

## Behaviour equivalence

- **value_set:** `migrate_type(symbol.get_type()) == symbol.get_type2()`
  by the symbol's lazy-cache round-trip invariant; `get_type2()` returns a
  reference whose lifetime covers the use (the symbol is a const&
  parameter, no setter runs in the call). The checked `to_code_type`
  replaces the manual `dynamic_cast`.
- **smt_conv:** the native builders mirror `fixedbvt::from_expr` /
  `ieee_floatt::from_expr` bit-for-bit — same spec (the `*_spect` `type2tc`
  ctors read the same fields the migrated-back legacy type carried, cf.
  Phase 2.3) and the same raw `v` / unpacked bit pattern; spec is set
  before `unpack` (which depends on it).
- **includes:** the full build links cleanly with all five removed
  (proof of staleness, incl. no transitive-includer breakage from the
  `irep2_dispatch.h` header).

## Intentionally omitted (blocked — resolved gates, not deferred)

- **Commit 16 (memory_alloc):** would build `array_type2t` directly, but
  that drops the legacy array's `dynamic(true)` flag, which
  `symex_valid_object.cpp:52,93` reads via `symbol.get_type().dynamic()`
  for dynamic-memory valid-object checks. `array_type2t` has no such
  field, so this would be a behaviour change, not a refactor. Left as-is.
- **Commit 18 (symex_catch):** the C++ base-class `bases` sub-irep is
  legacy-only — `struct_type2t` has no `bases` field and `migrate_type`
  never copies it (open question Q3, now resolved) — so the base-class
  walks must keep reading the legacy `get_type().find("bases")`. Left
  as-is, per the plan's "if it does not [survive], this site stays."

## Scope / metric note

Phase 2.4 touches the verifier core (pointer-analysis / goto-symex /
solvers / irep2), not `src/util`, so it does not move the `src/util`
migrate-census scoreboard; the value is removing 4 core-seam round-trips,
two of them hot (per function call; per fixedbv/float model readback).

## Verification

The goto-baseline harness is unreliable here (ESBMC re-bakes its
operational models non-deterministically per build — `docs/irep2-migration.md`
§8.1), so each commit was gated on **deterministic regression-verdict
equivalence vs clean `master`** over a stratified
`regression/{esbmc,floats,esbmc-cpp,python}` subset, plus the full floats
corpus (144/145, only pre-existing `isinf_ir_ieee`) and the affected unit
suites (`simplify2ttest` 236, `arith_tools_test`, `base_type`, `migrate`,
`symbol`). Every gate is identical to clean master. `code-reviewer` agent:
clean, no findings.

### DIFF
--- a/src/irep2/irep2_dispatch.h
+++ b/src/irep2/irep2_dispatch.h
@@ -21,7 +21,6 @@
 #include <util/i2string.h>
 #include <util/ieee_float.h>
 #include <util/migrate.h>
-#include <util/std_types.h>
 #include <irep2/irep2_type.h>
 #include <irep2/irep2_expr.h>
 #include <irep2/irep2_utils.h>
--- a/src/irep2/irep2_expr.cpp
+++ b/src/irep2/irep2_expr.cpp
@@ -8,7 +8,6 @@
 #include <irep2/irep2_dispatch.h>
 #include <util/message/format.h>
 #include <util/migrate.h>
-#include <util/std_types.h>
 
 // Pretty names indexed by expr2t::expr_ids. Driven by expr_kinds.inc;
 // adding a new expression kind there automatically populates this
--- a/src/irep2/irep2_guard.cpp
+++ b/src/irep2/irep2_guard.cpp
@@ -2,7 +2,6 @@
 #include <unordered_set>
 #include <irep2/irep2_guard.h>
 #include <irep2/irep2_utils.h>
-#include <util/std_expr.h>
 
 expr2tc guard2tc::as_expr() const
 {
--- a/src/irep2/irep2_type.cpp
+++ b/src/irep2/irep2_type.cpp
@@ -9,7 +9,6 @@
 #include <util/message.h>
 #include <util/message/format.h>
 #include <util/migrate.h>
-#include <util/std_types.h>
 
 /*************************** Base type2t definitions **************************/
 
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -9,7 +9,6 @@
 #include <util/expr_util.h>
 #include <util/i2string.h>
 #include <irep2/irep2.h>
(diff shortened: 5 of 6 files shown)


mikhailramalho (approved):
LGTM

PR #4957, opened 2026-05-30
[goto] fix scanf overflow check skipped on &arr[i] (#4956)

### LINKED ISSUE #4956: Flaky/environment-dependent verdict on cstd/scanf_bigint_2 and scanf_malloc_bug_3 (--overflow-check)
## Summary

Two CORE regression tests intermittently report `VERIFICATION SUCCESSFUL` when they expect `VERIFICATION FAILED`, depending on platform / toolchain / solver-dependency snapshot rather than on the source under test:

- `regression/cstd/scanf_bigint_2`
- `regression/cstd/scanf_malloc_bug_3`

Both use `--overflow-check` and expect a bug to be flagged (`^VERIFICATION FAILED$`). On some build/run environments ESBMC instead finds no violation and reports `VERIFICATION SUCCESSFUL`, so the tests fail. Because the verdict flips for *identical source*, this is a flaky / environment-dependent verification result, and it intermittently red-lights the CI of unrelated PRs.

## Affected tests

```c
// regression/cstd/scanf_bigint_2/main.c        (test.desc: CORE, --overflow-check, expects ^VERIFICATION FAILED$)
#include <stdlib.h>
int main(int argc, char *argv[]) {
  int *arr = (int *)malloc(100000000000 * sizeof(int));
  for (int i = 0; i < 3; i++) scanf("%100000000000d", &arr[i]);
  for (int i = 0; i < 3; i++) printf("%d", &arr[i]);
}
```

```c
// regression/cstd/scanf_malloc_bug_3/main.c    (test.desc: CORE, --overflow-check, expects ^VERIFICATION FAILED$)
#include <stdlib.h>
int main(int argc, char *argv[]) {
  int *arr = (int *)malloc(3 * sizeof(int));
  for (int i = 0; i < 3; i++) scanf("%13d", &arr[i]);
  for (int i = 0; i < 3; i++) printf("%d", &arr[i]);
}
```

Both were added in 2023 and last touched in 2024 (`01a8a50a21`, the `-Wno-error=` mass change); the sources have not changed since.

## Evidence that the verdict is environment-dependent, not source-dependent

The same commit produces different verdicts across runs/platforms:

- ❌ **`ubuntu-24.04 + llvm-22` (`static llvm-22 DebugOpt`)**: reports `VERIFICATION SUCCESSFUL` → tests fail. Observed on a recent run (e.g. the `ubuntu-24.04` leg of run `26660860780`); the job log shows the verdict, not a timeout or build error.
- ❌ **local macOS arm64**, ESBMC 8.3.0, `RelWithDebInfo`, Z3 + Bitwuzla: deterministically `VERIFICATION SUCCESSFUL` under **both** `--z3` and the default Bitwuzla. Reproduces on a clean `master` checkout (no unrelated changes).
- ✅ **macOS / Windows CI legs**, and earlier `ubuntu-24.04 + llvm-22` runs on other PRs (e.g. #4913, #4925): report `VERIFICATION FAILED` → tests pass.

So identical source verifies differently depending on the environment. The likely driver is the solver-dependency snapshot (`-DDOWNLOAD_DEPENDENCIES=On` pulls the latest Bitwuzla/Z3/Boolector/CVC5 builds, which drift over time) and/or the LLVM-22 C front-end's lowering of the pathological inputs (huge `malloc` size, huge `scanf` field widths).

## Reproduction

```sh
esbmc regression/cstd/scanf_bigint_2/main.c    --overflow-check -Wno-error=implicit-function-declaration
esbmc regression/cstd/scanf_malloc_bug_3/main.c --overflow-check -Wno-error=implicit-function-declaration
# Expected per test.desc: VERIFICATION FAILED
# Observed (macOS arm64 / ubuntu-24.04+llvm-22, current solver deps): VERIFICATION SUCCESSFUL
```

## Likely root-cause area

The expected bug lives in the interaction of `--overflow-check` with the `scanf`/`malloc` operational models on these pathological inputs:

- `scanf_bigint_2`: `malloc(100000000000 * sizeof(int))` and a `%100000000000d` field width — whether the size-multiplication overflow / huge-width handling raises a property depends on the model + solver.
- `scanf_malloc_bug_3`: `%13d` writing into a 3-`int` buffer via the `scanf` model — whether an out-of-bounds/overflow property is generated and proven violable is the question.

This points at `src/c2goto/library/` `scanf`/`malloc` models and/or the overflow instrumentation, with the verdict sensitive to the active solver build. (It is **not** in any language frontend other than C — the Solidity/Python/etc. frontends are not involved.)

## Impact

Intermittently fails CI on unrelated PRs (surfaced on #4951, a Solidity-frontend-only change that has no connection to the C `scanf` path). A flaky CORE test blocks merges and erodes trust in the regression signal.

## Suggested next steps

1. Determine whether the verdict is **solver-version-dependent** (bisect Bitwuzla/Z3/Boolector/CVC5 versions) or **LLVM-version-dependent** (llvm-22 vs older) for these two inputs.
2. Fix the underlying `scanf`/`malloc`/overflow modelling so the expected violation is detected deterministically, **or** — if the inputs are genuinely under-specified for a deterministic verdict — tighten the tests (pin a solver in `test.desc`, constrain the inputs) so they assert a stable property.
3. Until resolved, consider whether these two should be temporarily reclassified so they don't intermittently block unrelated CI.

## Environment

- ESBMC 8.3.0
- Failing CI config: `ubuntu-24.04`, LLVM 22, `DebugOpt`, `BUILD_STATIC=ON`, Boolector + Bitwuzla + Z3 + CVC5, `DOWNLOAD_DEPENDENCIES=On`
- Local repro: macOS arm64, `RelWithDebInfo`, Z3 + Bitwuzla

lucasccordeiro:
`get_base_object()` strips `address_of` and `&arr[i]` lowers to pointer arithmetic (an `add2t`), so `input_overflow_check`'s `is_address_of2t` branch (added in #4946) was dead and the scanf buffer-overflow check was silently skipped — flipping `scanf_bigint_2` / `scanf_malloc_bug_3` to SUCCESSFUL. Peel the pointer-typed operand of the add/sub to recover the buffer symbol instead.

Restores deterministic detection under both Bitwuzla and Z3 (the verdict was never solver/LLVM drift). Adds an in-bounds companion test `scanf_malloc_inbounds` (`%9d` → SUCCESSFUL) to guard against over-triggering.

Fixes #4956

### DIFF
--- a/regression/cstd/scanf_malloc_inbounds/main.c
+++ b/regression/cstd/scanf_malloc_inbounds/main.c
@@ -1,0 +1,14 @@
+#include <stdlib.h>
+// Companion to scanf_malloc_bug_3: same malloc'd int buffer, but the scanf
+// field width (%9d) fits a 32-bit int, so the input-overflow check must NOT
+// fire. Exercises the &arr[i] pointer-arithmetic path in
+// goto_checkt::input_overflow_check and guards against over-triggering.
+int main(int argc, char *argv[])
+{
+  int *arr = (int *)malloc(3 * sizeof(int));
+  for(int i = 0; i < 3; i++)
+  {
+    scanf("%9d", &arr[i]);
+  }
+  return 0;
+}
--- a/regression/cstd/scanf_malloc_inbounds/test.desc
+++ b/regression/cstd/scanf_malloc_inbounds/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--overflow-check -Wno-error=implicit-function-declaration
+^VERIFICATION SUCCESSFUL$
--- a/src/goto-programs/goto_check.cpp
+++ b/src/goto-programs/goto_check.cpp
@@ -454,30 +454,27 @@ void goto_checkt::input_overflow_check(
   for (long unsigned int i = fmt_idx + 1; i <= number_of_format_args + fmt_idx;
        i++)
   {
-    const expr2tc &base_expr = get_base_object(func_call.operands[i]);
+    expr2tc base_expr = get_base_object(func_call.operands[i]);
     irep_idt arg_name;
-    if (is_symbol2t(base_expr))
-      arg_name = to_symbol2t(base_expr).thename;
 
     // e.g
     // int *arr = (int*) malloc(10 * sizeof(int));
     // scanf("%13d",&arr[0]);  --> overflow
-    if (arg_name.empty())
+    //
+    // `&arr[i]` is lowered to pointer arithmetic `arr + i` (an add2t),
+    // which get_base_object does not peel.  Descend into the pointer-typed
+    // operand to recover the underlying buffer symbol.
+    if ((is_add2t(base_expr) || is_sub2t(base_expr)) &&
+        is_pointer_type(base_expr))
     {
-      expr2tc deref = get_base_object(func_call.operands[i]);
-
-      // not the format we expected
-      if (!is_pointer_type(deref))
-        return;
-
-      // we expect an address_of(symbol) underneath
-      if (!is_address_of2t(deref))
-        return;
-      const expr2tc &target = to_address_of2t(deref).ptr_obj;
-      if (!is_symbol2t(target))
-        return;
-      arg_name = to_symbol2t(target).thename;
+      const expr2tc &lhs = *base_expr->get_sub_expr(0);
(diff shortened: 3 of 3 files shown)


mikhailramalho (approved):
Lgtm

PR #1038, opened 2023-05-23
Fix ieee_floatt to_double()

rafaelsamenezes:
This fixes the union used to generate doubles (thanks @fbrausse)

Closes #1037 

SV-Comp run: https://github.com/esbmc/esbmc/actions/runs/5059731127

PR #4981, opened 2026-05-30
[goto-programs] don't warn on float zero in interval analysis

### LINKED ISSUE #1037: ESBMC is not converting ieee_float to double correctly
The definition `double one_modf = 1.0;` makes ESBMC generate the following expression:

```
constant_floatbv
* value : 1.000000
* type : floatbv
  * fraction : 52
  * exponent : 11
```

Which seems correct... however invoking `to_double` causes some weird behavior: the number is not normal anymore. Also, this seems to be triggered by the `from_integer`

```cpp
int one = 1;
assert(isnormal(one)); // Holds

double d_one = 1;
assert(isnormal(d_one)); // Holds

ieee_floatt ieee_one(ieee_float_spect(52,11));

ieee_one.from_double(d_one);
assert(isnormal(ieee_one.to_double())); // Holds

ieee_one.from_double(one);
assert(isnormal(ieee_one.to_double())); // Holds

ieee_one.from_integer(one);
assert(isnormal(ieee_one.to_double()));  // Fails
```

@mikhailramalho any suggestions?

lucasccordeiro:
The interval domain treated every zero double constant as a conversion failure (isnormal(0.0) is false), flooding output with a bogus "ESBMC fails to convert constant_floatbv ... into double" warning. The underlying to_double() bug was already fixed by PR #1038; zero converts to exactly 0.0. Handle zero as the tight interval [0,0], drop the misleading warning, and keep the conservative top result for the remaining non-normal cases (NaN/infinity/subnormal). Adds a unit test and github_1037{,_fail} regression tests.

Fixes #1037

### DIFF
--- a/regression/esbmc/github_1037/main.c
+++ b/regression/esbmc/github_1037/main.c
@@ -1,0 +1,11 @@
+// Regression for #1037: the interval domain used to print a bogus
+// "ESBMC fails to convert constant_floatbv ... value 0.000000 ... into double"
+// warning for every zero double constant, because isnormal(0.0) is false.
+// Zero converts to 0.0 correctly, so no warning must be emitted.
+int main()
+{
+  double x = 0.0;
+  double y = x + 1.0;
+  __ESBMC_assert(y > 0.0, "y must be positive");
+  return 0;
+}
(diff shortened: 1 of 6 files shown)


mikhailramalho (approved):
Lgtm

Issue #5012, opened 2026-05-31
[SV-COMP] Sound return-length model for printf/(v)asprintf (tracks #4976-#4979)


lucasccordeiro:
Tracking issue for a sound operational model of the printf / `(v)asprintf` return value, to close the busybox `no-overflow` false positives #4976 #4977 #4978 #4979 (umbrella #2513) without introducing false negatives.

## Why the naive fix is unsound
Routing `vasprintf` through the current `symex_printf` hides a real overflow (measured in PR #5009). `symex_printf`/`printf_formattert` have two latent **under-approximations**: a non-constant format returns `0` (`src/goto-symex/builtin_functions/io.cpp:50-61,245-249`) and a non-literal `%s` contributes `0` (`src/goto-symex/printf_formatter.cpp:302-312`). And variadic args passed via a `va_list` to a bodyless callee like `vasprintf` are erased (`src/goto-symex/symex_function.cpp:214-251`).

## Soundness invariant
The modelled return `R` must satisfy `R >= true output length` on every path — over-approximation (at worst a false positive) is allowed; under-approximation (a false negative) is not.

## Proposal (summary)
- **G-A** non-constant format ⇒ return unconstrained nondet `>= 0`, never `0`.
- **G-B** non-literal `%s` ⇒ object-size upper bound (or unconstrained), never `0`.
- **G-D** wire `asprintf`/`vasprintf`/`vsprintf`/`vsnprintf` into `printf_kindt`/`symex_printf` (format-arg index + `char**` dest + return = bounded length).
- **G-C** (optional, precision-only) recover `va_arg<N>` symbols of the nearest inlined ancestor to tighten `%s`.

Numeric/char conversions are type-bounded without the arg value, and `sync-1`'s only call is the literal `bb_error_msg("ignoring all arguments")` (no specifiers/varargs) which inlines to a constant format — so a sound model needs no va_list recovery there.

## Honest scope
Closes the **constant-format** cases (e.g. `sync-1`, length 22); genuinely symbolic formats stay sound-but-unbounded (still a precision limitation). The overflow could also shift to `strlen(applet_name)`/`strlen(msg_eol)` if those are unbounded — Phase 0 must confirm the benchmark actually flips to SUCCESSFUL.

## Plan & gates
[...]

Issue #4979, opened 2026-05-30
[SV-COMP] false alarm on no-overflow: whoami-incomplete-2


lucasccordeiro:
## Wrong result in SV-COMP 26 run

**False alarm**: ESBMC reports `VERIFICATION FAILED` (FALSE) on a benchmark whose ground truth is TRUE — a spurious counterexample on a correct program (precision/incompleteness bug).

| Field | Value |
|---|---|
| Property | `no-overflow` |
| Expected verdict | `true` |
| ESBMC verdict | `false(no-overflow)` |
| Benchmark | [`c/busybox-1.22.0/whoami-incomplete-2.yml`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/busybox-1.22.0/whoami-incomplete-2.yml) ([source](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/busybox-1.22.0/whoami-incomplete-2.i)) |
| Run | [SV-COMP 26 benchexec run](https://github.com/esbmc/esbmc/actions/runs/26680912297) |
| ESBMC version | 8.3.0 (commit `911d1790c6ead18fc070e1c14a9062b363fe869a`) |
| Strategy | `kinduction` |
| Related umbrella issue(s) | #2513 |

## Reproducer

Wrapper invocation (matches the SV-COMP run):

```
./esbmc-wrapper.py -p sv-benchmarks/c/properties/no-overflow.prp -s kinduction --arch 64 sv-benchmarks/c/busybox-1.22.0/whoami-incomplete-2.i
```

Underlying ESBMC command:

```
./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check sv-benchmarks/c/busybox-1.22.0/whoami-incomplete-2.i --64 --witness-output witness --no-pointer-check --no-bounds-check --overflow-check --no-assertions --incremental-bmc
```

To reproduce locally, fetch the benchmark from the [sv-benchmarks repo](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/busybox-1.22.0/whoami-incomplete-2.i) and run the wrapper command above against it.

## Log tail

```
[...]

PR #5017, opened 2026-05-31
[symex] sound printf return-length: don't under-approximate a non-constant format

lucasccordeiro:
## What

`symex_printf` modelled the printf-family return value (the character count) as a **constant 0**
when the format string was not a compile-time constant. That is an *under-approximation*: the true
return is then unknown and can be arbitrarily large, so any overflow/size check on the returned
value was silently discharged — a **false negative** (the worst outcome for a verifier).

A genuine signed overflow on `sprintf(buf, fmt, x) + base` with a non-constant `fmt` reported
`VERIFICATION SUCCESSFUL` on master; with this change it correctly reports `VERIFICATION FAILED`.

## How

`printf_formattert` gains a `bounded` flag, cleared whenever a conversion's contribution to the
output length cannot be soundly upper-bounded. `symex_printf` now:

- pins the return to an exact constant only when the format is a compile-time constant **and** every
  conversion was bounded;
- bounds it to `[min,max]` when those hold but some argument is nondet;
- otherwise falls back to an **unconstrained non-negative nondet** — the same sound
  over-approximation ESBMC already uses for a function with no operational model.

Integer/`%c` conversions stay type/width-bounded, so the precise common case is unchanged. The
`%s` / `%e%f%g` / missing-argument clears are conservative soundness hardening (they can only ever
add false positives, never false negatives).

## Validation

Built master and this branch from source and compared:

| case | master | this PR |
|---|---|---|
| non-constant-format `sprintf` return + near-INT_MAX (real overflow) | `SUCCESSFUL` (false negative) | **`FAILED`** |
| constant-format exact-length case | `SUCCESSFUL` | `SUCCESSFUL` |
| constant `%d` real overflow | `FAILED` | `FAILED` |
| `cstd` printf regression subset | 22/22 | **22/22** |

No master-`FAILED` flips to `SUCCESSFUL` (monotone in the safe direction). Two regression tests
added: `printf_nonconst_format_overflow` (→ FAILED) and `printf_const_format_exact` (→ SUCCESSFUL).

## Scope / merge gate

This is the standalone soundness prerequisite from the printf return-length design
(`docs/design-printf-return-length-om.md`, #5012). It does **not** by itself close the busybox
`no-overflow` false positives #4976–#4979 — those call `vasprintf`, which still has no model;
wiring the `(v)asprintf` family in is the next phase (tracked in #5012).

**Merge gate (not run in PR CI):** per the design, before merge this needs a full SV-COMP
`no-overflow` benchexec run confirming **`wrong-false` (false negatives) stays 0**, plus a
precision-delta report. Requesting that run on this branch.

Refs #4976 #4977 #4978 #4979 #5012

### DIFF
--- a/regression/esbmc/printf_const_format_exact/main.c
+++ b/regression/esbmc/printf_const_format_exact/main.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+
+// Precision-preservation control for the printf soundness fix. The format is a
+// compile-time constant with no conversion specifiers, so the sprintf return is
+// pinned to the exact length (5). The subsequent addition cannot overflow, so
+// verification must still succeed — the fix must not over-approximate the
+// fully-determined case.
+int main(void)
+{
+  char buf[64];
+  int n = sprintf(buf, "hello"); // exact length 5
+  int base = 10;
+  int t = base + n; // 15, no overflow
+  return t;
+}
--- a/regression/esbmc/printf_const_format_exact/test.desc
+++ b/regression/esbmc/printf_const_format_exact/test.desc
@@ -2,4 +2,3 @@ CORE
 main.c
 --overflow-check --incremental-bmc
 ^VERIFICATION SUCCESSFUL$
-^EXIT=0$
--- a/regression/esbmc/printf_const_format_exact/test.desc
+++ b/regression/esbmc/printf_const_format_exact/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--overflow-check --incremental-bmc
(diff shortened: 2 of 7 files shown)


lucasccordeiro:
This PR (30s): 

```
  correct:           20360
    correct true:    13681
    correct false:    6679
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           16233
  Score:             33929 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/26737093736.
```

PR #5024, opened 2026-06-01
Revert #4838 (k-induction pointer-invariant havocs): SV-COMP regression

lucasccordeiro:
Reverts #4838. Constraining havoc'd pointer loop variables to their
pre-havoc candidates weakens the k-induction inductive step, letting it
close spuriously and report buggy programs SAFE. Artifact-diffing two
Benchexec runs traced an SV-COMP score drop (38440 to 37395) to this, with
51 Intel-TDX unreach-call tasks flipping from TIMEOUT to unsound SAFE.


lucasccordeiro:
This PR (300s): https://github.com/esbmc/esbmc/actions/runs/26752678376.

Master (300s): 

```
  correct:           23783
    correct true:    15628
    correct false:    8155
  incorrect:            73
    incorrect true:     53
    incorrect false:    20
  unknown:           12744
  Score:             37395 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/26695000028/job/78679032124
```

lucasccordeiro:
Tracking the underlying unsoundness and the correct re-landing path in #5025.

lucasccordeiro:
## Recommend closing — superseded by the merged #5228

This PR reverts #4838 wholesale because the inductive-step pointer-invariant change weakened the inductive step and let k-induction close spuriously on the Intel-TDX-Module `unreach-call` havoc harnesses (the 38440 → 37395 SV-COMP score drop, incorrect-true 11 → 53).

That same soundness hole was fixed on master by **#5228** (`bef6149cad`, *"[k-induction] Disable inductive step on pointer-array writes"*, **Fixes #5224**, merged 2026-06-09). #5228 keeps #4838's precision but detects writes through a pointer to an array element (e.g. `(*dest)[i]`, as in the Intel-TDX nondet-array builders) and disables the inductive step only for those programs — base case and forward condition still run, and existing inductive proofs (stack arrays, single pointer/field writes, `is_pointer_invariant_list`) are preserved.

Net effect: master now has the targeted soundness guard **without** losing the precision that a full revert of #4838 would discard. Reverting #4838 on top of #5228 would be a net precision regression with no soundness benefit.

Verified on master `bd2e4a7d79`: both `8955e6fce1` (#4838) and `bef6149cad` (#5228) are present, and #5224 is closed as completed by #5228.

Recommend closing this revert as superseded.

PR #5027, opened 2026-06-01
[k-induction] sound IS pointer-invariant strengthening via L1 value-set restore

### LINKED ISSUE #5025: Unsound k-induction: IS pointer-invariant havocs (#4838) report buggy programs SAFE
## Summary

PR #4838 (`[k-induction] symex-side pointer invariants for inductive-step havocs`, commit `8955e6fce1`) is **unsound**. It causes the k-induction inductive step to close spuriously and report buggy programs SAFE. Reverted in #5024 to recover the SV-COMP regression; this issue tracks re-landing the idea correctly.

## The unsoundness

When `make_nondet_assign` havocs a pointer-typed loop variable `p` in the inductive step (IS), #4838 rewrites the nondet RHS into an ITE chain over `p`'s *pre-havoc* value-set candidates, so the havoc'd `p` "inherits that constraint instead of becoming a free variable." It also changed the `--add-symex-value-sets` deref-assume in `symex_dereference.cpp` from `return` to `continue` on non-`object_descriptor` entries, and restores the pre-havoc value-set entry after the assignment.

The inductive step must over-approximate an **arbitrary** loop iteration: the havoc'd pointer has to be free to point at any valid object. Constraining it to the pre-havoc candidate set means the IS no longer covers iterations where `p` points elsewhere, so it cannot reach the real counterexample — the induction "succeeds" and ESBMC reports SAFE on programs that are actually unsafe. #4838's own `symex_dereference.cpp` comment makes the assumption explicit ("we are choosing to assume that the pointer points at one of the concrete candidates we *can* name"), which is sound for a deref guard but not as an inductive hypothesis over a havoc.

## Evidence (SV-COMP 300s)

Diffing two "Run Benchexec" artifacts — good `dcdbc68` (run 26187062780) → bad `9b55f23` (run 26695000028):

- Score 38440 → 37395; **incorrect-true 11 → 53**.
- **51 of the 52** newly-introduced incorrect-true tasks are `sv-benchmarks/c/intel-tdx-module/` ReachSafety `unreach-call` harnesses (`tdh_mr_finalize__*`, `tdh_phymem_page_reclaim__*`, the `__cover_{entry,exit,proof}_havoc_{memory,object}` variants).
- All 51 were **TIMEOUT** in the good run and now **terminate with a wrong SAFE** — the signature of an inductive step that went from "unconstrained pointer → spurious CEX → k keeps growing → timeout" to "over-constrained pointer → converges (unsoundly)."
- The runset uses `-s kinduction`, which the SV-COMP wrapper (`scripts/competitions/svcomp/esbmc-wrapper.py:233,316`) expands to `--k-induction --max-inductive-step 3 --add-symex-value-sets` — exactly the gate for #4838's code.

(One unrelated 52nd task, `c/pthread-ext/23_lu-fig2.fixed` [no-data-race], is a separate concurrency flip, not part of this issue.)

## How to re-land correctly

The pre-havoc candidate set may be used **additively** but must never replace the free havoc:

- Keep `p = nondet()` genuinely unconstrained in the IS.
- If a SAME-OBJECT relationship is desired as a heuristic to cut spurious CEXs, add it as an *assumption that is itself discharged* (e.g. a verified pointer invariant proved in a base case / strengthening step), not as a hard constraint baked into the havoc RHS. An inductive hypothesis is only sound if it holds for every reachable iteration; the pre-havoc value-set does not establish that.
- Add a regression pair that fails (reports SAFE) under the unsound encoding and passes (UNKNOWN/CEX) under the corrected one — ideally a minimal pointer-traversing loop with a reachable bug, mirroring the TDX havoc shape.

## Refs
- Reverted by #5024.
- Original PR #4838.

lucasccordeiro:
## Summary

Re-lands the IS pointer-invariant idea from #4838 correctly, without the unsoundness that caused #5025. Targets #5024 (the revert branch) as base.

The original defect: #4838 replaced `p = nondet()` with a deterministic ITE chain over pre-havoc candidates, over-constraining the inductive step and producing spurious SAFE on 51 Intel-TDX `unreach-call` benchmarks.

This implementation keeps `p = nondet()` genuinely free and uses IS strengthening instead:

1. Snapshot the pre-havoc value-set via its **L1 key** (the previous code used `get_value_set(L2-renamed-expr)` which misses L1-keyed entries and falls back to `{unknown}`, causing the assume to always abort).
2. If all entries are concrete (no `unknown`/`invalid`): emit `assume(SAME-OBJECT(p_new, &obj))` disjunction and **restore the value-set entry** to the pre-havoc candidates. Without the restore, the deref-time `--add-symex-value-sets` assume sees `{unknown}` post-havoc and emits nothing, so dereferences fall back to `invalid_object` — spurious CEXs.
3. If any `unknown`/`invalid` present: leave `p` unconstrained (TDX case — no change from post-revert behaviour).

Additional changes:
- `symex_dereference.cpp`: `return` → `continue` so the deref-time assume is emitted even when the value-set has mixed concrete/unknown entries.
- `esbmc_parseoptions.cpp`: auto-enable `--add-symex-value-sets` for all k-induction paths, including the implicit `--loop-invariant` and `--validate-correctness-witness` paths (previous code only checked `cmdline.isset`, missing the implicit paths).
- Read-only pre-snapshot lookup via `values.find()` instead of `get_entry()` to avoid silently inserting empty entries.

## Tests

- `regression/k-induction/is_pointer_invariant_list`: linked-list traversal that IS k=3 proves SAFE (was spurious SAT before this fix).
- `regression/k-induction/is_pointer_invariant_list_bug`: same shape with a genuine bug (h=2 node) — soundness pin that the strengthening does not hide real bugs; expects FAILED.
- Full k-induction suite (108 tests): 100% pass.
- Loop-invariants suite (63 tests): 100% pass.

Fixes #5025.

### DIFF
--- a/regression/incremental-smt/incremental-18/test.desc
+++ b/regression/incremental-smt/incremental-18/test.desc
@@ -1,4 +1,4 @@
 THOROUGH
 main.c
 --smt-during-symex --smt-symex-assert --smt-symex-guard --k-induction --add-symex-value-sets
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION UNKNOWN$
--- a/regression/incremental-smt/incremental-24/test.desc
+++ b/regression/incremental-smt/incremental-24/test.desc
@@ -1,4 +1,4 @@
 THOROUGH
 main.c
 --smt-during-symex --smt-symex-assert --smt-symex-guard --k-induction --add-symex-value-sets
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION UNKNOWN$
--- a/regression/k-induction/is_deep_pointer_bug/main.c
+++ b/regression/k-induction/is_deep_pointer_bug/main.c
@@ -1,0 +1,57 @@
+// Regression for issue #5025: the k-induction inductive-step pointer
+// invariant in #4838 (and the all-concrete gate in #5027 alone) was unsound
+// because it read the symex-prefix value-set rather than a goto-program-level
+// fixpoint. This test exercises the failure mode #4838 lacked a test for:
+//
+//   - a linked list of 5 nodes, last one (h=2) violates the traversal check;
+//   - the violation depth is 5, deeper than --max-inductive-step 3 reaches in
+//     the base case, so the verdict at this k must come from IS, not BC;
+//   - with #4838's unsound rewrite, IS converged over the restricted candidate
+//     set and reported VERIFICATION SUCCESSFUL on this bug;
+//   - with the goto-program-level VSA fixpoint as the candidate source, the
+//     reachable iteration that points p at the 5th node is in the over-
+//     approximation, so the assume does not exclude the bug.
+//
+// The desc accepts FAILED or UNKNOWN as passing (precise verdict is solver-
+// and strategy-dependent). The unsound case is SUCCESSFUL, which the regex
+// must never match.
+#include <stdlib.h>
+
+typedef struct node
+{
+  int h;
+  struct node *n;
+} *List;
+
+int main()
+{
+  List a = (List)malloc(sizeof(struct node));
+  if (!a)
+    return 0;
+  a->h = 1;
+  a->n = 0;
+
+  // Build 4 more nodes; the 4th (5th in the chain) has h == 2.
+  List cur = a;
+  for (int i = 1; i <= 4; i++)
+  {
+    List t = (List)malloc(sizeof(struct node));
+    if (!t)
+      return 0;
+    t->h = (i == 4) ? 2 : 1;
+    t->n = 0;
+    cur->n = t;
+    cur = t;
+  }
+
+  // Traverse: the assertion is violated at the 5th node.
+  List p = a;
+  while (p)
+  {
+    if (p->h != 1)
+ERROR: { return 1; }
+    p = p->n;
+  }
+
+  return 0;
+}
--- a/regression/k-induction/is_deep_pointer_bug/test.desc
+++ b/regression/k-induction/is_deep_pointer_bug/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
+^VERIFICATION (FAILED|UNKNOWN)$
--- a/regression/k-induction/is_pointer_invariant_list/main.c
+++ b/regression/k-induction/is_pointer_invariant_list/main.c
@@ -1,0 +1,48 @@
+// Linked-list traversal where the inductive step can only prove the
+// assertion if the IS pointer-invariant strengthening is active.
+// A build loop extends a list with nodes whose h == 1; a check loop
+// walks it asserting h == 1.  Without value-set restore + SAME-OBJECT
+// assume at the IS havoc, the deref encoding falls back to invalid_object
+// and the solver finds a spurious counterexample.
+#include <stdlib.h>
+extern int __VERIFIER_nondet_int(void);
+
+typedef struct node
+{
+  int h;
+  struct node *n;
+} *List;
+
+int main()
+{
+  List a = (List)malloc(sizeof(struct node));
+  if (!a)
+    return 0;
+  a->h = 1;
+  a->n = 0;
+
+  List end = a;
+  while (__VERIFIER_nondet_int())
+  {
+    List t = (List)malloc(sizeof(struct node));
+    if (!t)
+      return 0;
+    t->h = 1;
+    t->n = 0;
+    end->n = t;
+    end = t;
+  }
+
+  List p = a;
+  while (p)
+  {
+    if (p->h != 1)
+      ERROR:
+      {
+        return 1;
+      }
+    p = p->n;
+  }
+
+  return 0;
+}
--- a/regression/k-induction/is_pointer_invariant_list/test.desc
+++ b/regression/k-induction/is_pointer_invariant_list/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION UNKNOWN$
--- a/regression/k-induction/is_pointer_invariant_list/test.desc
+++ b/regression/k-induction/is_pointer_invariant_list/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
+^VERIFICATION SUCCESSFUL$
--- a/regression/k-induction/is_pointer_invariant_list_bug/main.c
+++ b/regression/k-induction/is_pointer_invariant_list_bug/main.c
@@ -1,0 +1,45 @@
+// Soundness pin: same shape as is_pointer_invariant_list but the build
+// loop appends a node with h == 2.  The IS pointer-invariant strengthening
+// must NOT hide this genuine bug — VERIFICATION FAILED is required.
+#include <stdlib.h>
+extern int __VERIFIER_nondet_int(void);
+
+typedef struct node
+{
+  int h;
+  struct node *n;
+} *List;
+
+int main()
+{
+  List a = (List)malloc(sizeof(struct node));
+  if (!a)
+    return 0;
+  a->h = 1;
+  a->n = 0;
+
+  List end = a;
+  while (__VERIFIER_nondet_int())
+  {
+    List t = (List)malloc(sizeof(struct node));
+    if (!t)
+      return 0;
+    t->h = 2;
+    t->n = 0;
+    end->n = t;
+    end = t;
+  }
+
+  List p = a;
+  while (p)
+  {
+    if (p->h != 1)
+      ERROR:
+      {
+        return 1;
+      }
+    p = p->n;
+  }
+
+  return 0;
+}
--- a/regression/k-induction/is_pointer_invariant_list_bug/test.desc
+++ b/regression/k-induction/is_pointer_invariant_list_bug/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
+^VERIFICATION FAILED$
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -518,8 +518,7 @@ void esbmc_parseoptionst::get_command_line_options(optionst &options)
   // --validate-correctness-witness — are also covered.
   if (
     options.get_bool_option("k-induction") ||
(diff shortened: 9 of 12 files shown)


lucasccordeiro:
This PR (300s): https://github.com/esbmc/esbmc/actions/runs/26770207377.

mikhailramalho (approved):
I think this is the correct solution; let's wait for the sv-comp results!

mikhailramalho:
I'll ask claude to review

mikhailramalho:
Concerned the all-concrete gate may not be sound. Posting the soundness
argument and a concrete counterexample shape before merge — happy to be
wrong; I haven't built this branch locally yet, so the counterexample
behaviour against this PR specifically is a prediction, not measured.

### Why the all-concrete gate is not sufficient

`cur_state->value_set` at the IS havoc point is the **symex-prefix
accumulation**, not a goto-time fixpoint over the loop:

- `src/goto-symex/goto_symex_state.cpp:314` — `value_set.assign(l1_lhs,
  l1_rhs)` updates the value-set inside per-assignment symex.
- `src/goto-symex/symex_goto.cpp:405` — value-set merged across branches
  via `make_union` *during* symex, not via a pre-symex closure.
- `src/goto-symex/execution_state.cpp` — the value-set lives per-symex-state
  and is cloned per branch.

So "all entries are concrete (no `unknown`/`invalid`)" only tells us the
symex prefix that reached the loop head did not have to widen to `unknown`
along *that prefix*. It does **not** establish that every reachable
loop-head state has `p` in the snapshot set — that would require a closure
over the loop body.

### Counterexample shape

```c
// build loop allocates O1..On with nondet length, last node h=2
List a = malloc(...); a->h = 1; a->n = NULL;
List end = a;
while (nondet()) {
  List t = malloc(...);
  t->h = nondet_choice(1, 2);
  t->n = NULL;
  end->n = t; end = t;
}

List p = a;
while (p) {                  // IS havoc on p here
  if (p->h != 1) ERROR;
  p = p->n;
}
```

Under `--max-inductive-step 3`, the symex prefix at the second loop's IS
havoc has unrolled the build loop 3 times — value-set for `p` is `{O1, O2,
O3}`, all-concrete. The gate fires: we assume `SAME-OBJECT(p, &O1) ∨ &O2 ∨
&O3` and restore the value-set. But the real loop-head state space includes
runs that allocated O4..On — objects this symex prefix never created. IS
now excludes any iteration where `p` points to one of them; if a bug lives
in O100, IS misses it. Same defect mechanism as #4838, behind a different
(still incorrect) predicate.

This matches the TDX failure shape: deep allocator chains where the symex
prefix sees a bounded subset, IS converges, and the bug at depth > prefix
is wrongly reported SAFE. Worth re-running the SV-COMP TDX subset against
*this* branch specifically before merge — if the 51 incorrect-true tasks
are still wrong-SAFE here, that confirms the gate is not closing them.

### Suggestion: query `value_set_analysist` instead

`src/pointer-analysis/value_set_analysis.{h,cpp}` already implements a
goto-program-level value-set with a worklist fixpoint over the CFG
(`src/goto-programs/static_analysis.cpp:141`,
`static_analysis_baset::fixedpoint`). It exposes:

```cpp
void get_values(locationt l, const expr2tc &expr,
                value_setst::valuest &dest) override;
```

at `src/pointer-analysis/value_set_analysis.h:57`, implemented as
`(*this)[l].value_set->get_value_set(expr, dest)`. The fixpoint result *is*
a sound over-approximation of all reachable states at that location — the
closure over back-edges is exactly the thing the symex-prefix set is
missing. It's already used for GCSE, `--show-goto-value-sets`, and the
GOTO-program dereference pass; just not wired into the IS strengthening.

Minimal concrete change: at the IS pointer-havoc detection in
`src/goto-symex/symex_assign.cpp`, replace the
`cur_state->value_set.get_entry(...).object_map` snapshot with a
`value_set_analysist::get_values(cur_state->source.pc, ptr_expr, dest)`
query on a precomputed analysis. The rest of the assume + restore can stay
the same — only the *source* of the candidate set changes.

If that plumbing is too much for this PR, the sound intermediate is the
pure revert (#5024), with the strengthening re-landed on top of the
fixpoint query in a follow-up.

### Missing regression test

#4838 went in without a test exercising the failure shape — the existing
`is_pointer_invariant_list_bug` is caught by BC at depth 5 before IS even
runs, so it doesn't guard against the IS-unsoundness class. The minimal
test that does is a list whose bug node is deeper than `--max-inductive-step`
so the verdict must come from IS:

```c
// regression/k-induction/is_deep_pointer_bug/main.c
#include <stdlib.h>
typedef struct node { int h; struct node *n; } *List;

int main()
{
  List a = (List)malloc(sizeof(struct node));
  if (!a) return 0;
  a->h = 1; a->n = 0;

  // 4 more nodes; the 5th (i==4) has h == 2
  List cur = a;
  for (int i = 1; i <= 4; i++) {
    List t = (List)malloc(sizeof(struct node));
    if (!t) return 0;
    t->h = (i == 4) ? 2 : 1;
    t->n = 0;
    cur->n = t; cur = t;
  }

  List p = a;
  while (p) {
    if (p->h != 1)
ERROR: { return 1; }
    p = p->n;
  }
  return 0;
}
```

```
CORE
main.c
--enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
^VERIFICATION (FAILED|UNKNOWN)$
```

Either FAILED or UNKNOWN passes; SUCCESSFUL is the unsound case and must
not match. The expected verdict under the unsound encoding is SUCCESSFUL;
under a sound encoding it should be FAILED or UNKNOWN. Worth running
against this branch to check whether the all-concrete gate actually
catches this case — if it still reports SUCCESSFUL, the soundness story
needs revisiting.

mikhailramalho:
Pushed `bf8670ed44` to this branch with the goto-time VSA approach we discussed. It closes #5025's soundness bug but has an empirical tradeoff you should look at before deciding to merge as-is.

### What changed

`symex_assign.cpp` now reads the IS pointer-invariant candidate set from a **goto-program-level value-set fixpoint** (built per-function, lazy, cached) instead of `cur_state->value_set`. The fixpoint closes over loop back-edges and is therefore a sound over-approximation of every reachable loop-head state — the closure the symex-prefix value-set lacks (the soundness argument from my earlier comment).

Two limitations forced a narrower scope than I'd hoped:

1. **Whole-program VSA is prohibitive.** Building VSA over the operational-model library on a 5-node list test didn't finish in 3 minutes. So the cache is per-function: each function's fixpoint is built on first IS pointer-havoc in that function and reused thereafter. Per-function builds are sub-second on the tests I ran.

2. **The goto-time VSA represents heap allocations as abstract `dynamic_object2t` markers, but ESBMC's SMT encoder has no realization of `address_of(dynamic_object)`.** A goto site id corresponds to potentially many symex-time `dynamic_N_value` SSA symbols (per-state allocation counter), and there's no translation map. So the gate **skips the entire strengthening on any heap candidate** — sound (silent no-op), but loses the precision win on heap-traversing loops.

The post-assignment value-set restore is also removed: VSA's `dynamic_object` entries would otherwise flow into symex's assignment-LHS resolution and trip `assignment to dynamic_object not handled`.

### Tests on `bf8670ed44`

| Test | Verdict | Notes |
|---|---|---|
| New `regression/k-induction/is_deep_pointer_bug` (5-node list, bug at node 5, BC reaches it) | `VERIFICATION FAILED at k=5` | The #5025 failure shape. On the previous tip this reported SUCCESSFUL (unsound). |
| `regression/k-induction/is_pointer_invariant_list` | **TIMEOUT** | On `658e3d3406` this verified SUCCESSFUL at IS k=2 in 0.5s. The all-heap candidate set is now skipped, so the strengthening is silently dropped and this falls back to plain k-induction, which doesn't close on this benchmark. |
| `regression/k-induction/is_pointer_invariant_list_bug` | `VERIFICATION FAILED at k=2` | Unchanged — BC catches the shallow bug before IS runs. |

So you're trading the heap-list precision win for soundness on the deep-heap-bug class (the 51 TDX failures).

### What I'd recommend you decide

You have three options I can see:

a. **Land `bf8670ed44` as the #5025 fix and accept the heap-list regression.** File a follow-up issue for the architectural fix that re-establishes the heap-list precision (use goto VSA as a soundness gate over already-materialized symex `dynamic_N_value` symbols; requires plumbing a site-id → SSA-symbol map through `memory_alloc.cpp` / `cpp_memory.cpp`).

b. **Revert `bf8670ed44`, land `#5024` (pure revert) as the #5025 fix**, and file both the soundness fix *and* the precision restoration as follow-ups. Same outcome from CI's perspective; less code change.

c. **Keep the current strengthening in `#5027` (without my changes) and just accept the SV-COMP regression.** Not recommended — that's what #5025 is filed against.

I'm leaning (a) because it closes #5025 with a regression test that explicitly guards against the failure class, and the heap-list precision can be brought back without any soundness compromise once C4 lands. But this is your call — happy to revert my commit if you'd rather go (b).

### Followups I'd file alongside (a)

- "Goto VSA as soundness gate over symex-time dynamic-object symbols" — restores heap-list precision; requires the site-id → SSA-symbol map mentioned above.
- "Symex-time value-set closure over loop back-edges" — Codex's option C1 in our design discussion; would obviate the goto-time / symex-time mismatch entirely. Larger.

Worth running the full SV-COMP TDX subset against `bf8670ed44` to confirm those 51 incorrect-true cases are closed before merging.

mikhailramalho:
Pushed `0698c3de25` follow-up. **All 51 TDX wrong-SAFE benchmarks from #5025 are now sound** under `bf8670ed44` + `0698c3de25`.

### Verification methodology

Downloaded benchexec artifacts from both runs cited in #5025:
- Good: master pre-#4838 (`dcdbc68`, run 26187062780)
- Bad: with #4838 unsoundness (`9b55f23`, run 26695000028)

Diffed run verdicts to extract exactly the regressions:

| Direction | Count |
|---|---|
| Regressions (good=ok → bad=wrong-SAFE) | **51 incorrect-true**, all `c/intel-tdx-module/`, all `tdh_{mr_finalize,phymem_page_reclaim}__*__cover_*` |
| Regressions (good=ok → bad=wrong-CEX) | 0 |
| Wins (good=wrong → bad=ok) | 4 incorrect-true + 3 incorrect-false (precision gains from #4838) |

### Result on the 51

Ran all 51 against this branch with `-s kinduction` flags (`--k-induction --max-inductive-step 3 --add-symex-value-sets` + the rest of the wrapper's reach config), 900s timeout, 26 concurrent.

| Stage | wrong-SAFE | TIMEOUT |
|---|---|---|
| Pre-`bf8670ed44` (Lucas's earlier tip) | 51 / 51 | 0 |
| After `bf8670ed44` (goto-time VSA gate) alone | 2 / 51 | 49 / 51 |
| After `bf8670ed44` + `0698c3de25` (deref-time gate) | **0 / 51** | **51 / 51** |

The two `cover_proof_*` stragglers slipped through `bf8670ed44` because their unsoundness was emitted from `symex_dereference.cpp`'s per-deref assume (which reads `cur_state->value_set`, the symex-prefix), not from the havoc-time path that `bf8670ed44` repaired. `0698c3de25` gates that path the same way — drop the whole assume on any non-`object_descriptor` or unknown-offset entry.

### Tradeoff still on the table

`is_pointer_invariant_list` (the linked-list traversal #4838 was motivated by) **still times out** under this combined fix. The strengthening's precision win on heap traversals is what `bf8670ed44`'s `dynamic_object` gate sacrificed, and `0698c3de25` tightens that further. Recovery is Codex's follow-up Option C4 (use goto VSA as a soundness gate over already-materialized symex `dynamic_N_value` symbols) — out of scope for this PR.

### Recommendation

If you want the SV-COMP TDX class closed with no other regression I can detect, this branch as it stands is the right thing to merge. If you need the heap-list precision back, the option remains to land #5024 (pure revert) instead and file the C4 follow-up.

Issue #1471, opened 2023-11-12
Struct constant in witness not parseable


fbrausse:
Full error from CPAchecker:
```
Cannot parse witness: The witness automaton provided is invalid! Reason: Cannot parse <dummy_data.AlternativeArchitecture = { .TickCountLow=0, .TickCountMultiplier=0, .InterruptTime={ .LowPart=0, .High1Time=0, .High2Time=0 },
    .SystemTime={ .LowPart=0, .High1Time=0, .High2Time=0 },
    .TimeZoneBias={ .LowPart=0, .High1Time=0, .High2Time=0 },
    .ImageNumberLow=0,
    .ImageNumberHigh=0, .NtSystemRoot=ARRAY_OF(0),
    .MaxStackTraceDepth=0, .CryptoExponent=0,
    .TimeZoneId=0, .Reserved2=ARRAY_OF(0), .NtProductType=0,
    .ProductTypeIsValid=0, .anon_pad_14=0,
    .NtMajorVersion=0, .NtMinorVersion=0, .ProcessorFeatures=ARRAY_OF(0),
    .Reserved1=0, .Reserved3=0,
    .TimeSlip=0, .AlternativeArchitecture=nondet_symbol(symex::nondet4), .anon_pad_22=0,
    .SystemExpirationDate={ .__annonCompField1={ .LowPart=0, .HighPart=0 } }, .SuiteMask=0,
    .KdDebuggerEnabled=0, .anon_pad_26=0 };>
```
This is for sv-comp's unreach-call: c/ntdrivers/floppy.i.cil-1.c

Potentially problematic here:
- nondet_symbol(symex::nondet4),
- ARRAY_OF(0), that's #1484
- the assignments of the `.anon_pad*` members ESBMC artificially introduces.

lucasccordeiro:
Triaged: struct-constant witness values (`{ .field=... }`, `ARRAY_OF(0)`, `nondet_symbol(...)`) are not parseable by CPAchecker. Verifying any fix here requires running the SV-COMP witness validator (CPAchecker / cpa-witness2test) on the regenerated witness, which is not available in this environment — so a fix cannot be confirmed sound without it (an unvalidated witness-format change risks silently breaking currently-passing validation). Recommend addressing as a batch with validator-in-the-loop testing. Note the GraphML assignment-value cases (#1471, #1520) share a root cause: ESBMC emits non-C-parseable aggregate/symbolic values in witness assignments.

[...]

PR #5038, opened 2026-06-02
[witness] drop aggregate-initialiser assumptions from GraphML witnesses

### LINKED ISSUE #1520: Witness not parseable by CPAchecker
This happened in the 4th pre-run for SV-COMP. For `c/seq-pthread/cs_lazy.i` CPA's output is
```Cannot parse witness: The witness automaton provided is invalid! Reason: Cannot parse <__CS_thread_status = { { 0 } };>```
This is likely related to #1488 because `__CS_thread_status` has the following definition as a global variable in the file:
```c
unsigned char __CS_thread_status[2][3 +1];
```

lucasccordeiro:
## Root cause
Violation witnesses emitted array/struct brace initialisers inside `<data key="assumption">` edges — e.g. `A == { 0, 0 };` and nested `status == { { 0, 0, 0 }, ... };`. CPAchecker / cpa-witness2test cannot parse a brace initialiser as a C scalar expression and **reject the entire witness automaton** (issue #1520; the array/struct-value half of #1471).

`get_formated_assignment` already excluded `constant_array2t`/`constant_struct2t`/`constant_union2t`, but `constant_array_of` values still rendered as braces and slipped through. A latent helper (`reformat_assignment_array`/`structs`) was meant to rewrite these into parseable element-wise form, but it was **dead**: the assumption string is built with `==`, while its trigger regexes match a single `=`, so it never fired — arrays leaked out as raw braces.

## Fix
- Clear any witness assumption whose rendered value contains a brace (`[{}]`) in `check_replace_invalid_assignment`. Witnesses are advisory **output** artifacts that never feed back into the verdict, so dropping an assumption cannot introduce a false positive/negative; scalar assumptions still survive for replay.
- Remove the dead `reformat_assignment_array`/`reformat_assignment_structs` functions, their regexes, and the two unreachable call sites.

## Soundness
No verification-logic change. A violation witness with the unparseable aggregate removed goes from *rejected outright* to *parseable but slightly more general* — strictly an improvement for SV-COMP witness validation. Scalar branch assumptions (the ones that actually drive replay to the error) are untouched (verified `x == 5;`, `x == 0;` still emitted).

## Tests
- `github_1520_witness_no_aggregate` — reproducer: a zero-init nested array read into a scalar; the witness drops the unparseable `status` aggregate and keeps the surviving `x == 0` (non-vacuous: positive + negative-lookahead assertions together).
- `github_1520_witness_scalar_kept` — anti-over-prune guard: a plain scalar input assumption must still appear.
- `output-format-array_of-{double,float,struct-int}` — these had locked in the unparseable brace witness; updated to assert the witness is now brace-free while preserving each test's human-readable counterexample-trace assertion.

All 75 witness/trace-touching regression tests pass locally.

## Limitations
End-to-end confirmation with CPAchecker / cpa-witness2test was not run (validator unavailable in the dev environment); the change is validated structurally (no brace ever reaches an assumption edge) and against the regression harness. The full struct-rendering case of #1471 (`nondet_symbol`, `ARRAY_OF` tokens) only reaches a top-level assumption inside an aggregate, which the existing struct guard plus this brace filter already drop; bare-token scalar forms are unreachable here because the path is gated by `is_constant_expr`.

Fixes #1520

### DIFF
--- a/regression/esbmc/github_1520_witness_no_aggregate/main.c
+++ b/regression/esbmc/github_1520_witness_no_aggregate/main.c
@@ -1,0 +1,16 @@
+/* Regression for #1520 (and #1471): an aggregate value (array / nested array /
+ * struct, or any ESBMC-internal placeholder such as nondet_symbol / ARRAY_OF)
+ * must never appear inside a GraphML witness assumption, because the SV-COMP
+ * witness validators (CPAchecker, cpa-witness2test) cannot parse a brace
+ * initialiser and reject the whole automaton.
+ *
+ * Here `status` is a zero-initialised global nested array. The violating step
+ * reads a single scalar element, so the witness should keep the scalar
+ * assumption (x == 0) and drop the unparseable `status == { { 0, ... } }`. */
+unsigned char status[2][3];
+
+int main()
+{
+  unsigned char x = status[1][2];
+  assert(x);
+}
--- a/regression/esbmc/github_1520_witness_no_aggregate/test.desc
+++ b/regression/esbmc/github_1520_witness_no_aggregate/test.desc
@@ -1,0 +1,6 @@
+CORE
+main.c
+--witness-output-graphml - -Wno-error=implicit-function-declaration
+^VERIFICATION FAILED$
+key="assumption">x == 0;
+(?s)\A(?!.*key="assumption">[^<]*[{}])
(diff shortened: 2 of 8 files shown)


XLiZHI (approved):
LGTM

PR #2798, opened 2025-09-25
[om] improved libc models for I/O and error handling

lucasccordeiro:
This PR tries to improve our libc modeling to provide more accurate verification of common UNIX and GNU extensions.

PR #5039, opened 2026-06-02
[om] add operational model for fseeko

### LINKED ISSUE #2797: [sv-comp] no body for function....
/sv-benchmarks/c/coreutils-v8.31

$ esbmc --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check comm_3args_ok.c --64 --witness-output witness.graphml --memory-leak-check --no-reachable-memory-leak --no-assertions --no-abnormal-memory-leak --malloc-zero-is-null --incremental-bmc

WARNING: no body for function fputs_unlocked
WARNING: no body for function __fpending
WARNING: no body for function ferror_unlocked
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function fseeko
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function error
WARNING: no body for function _exit
WARNING: no body for function __fpending
WARNING: no body for function ferror_unlocked
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function fseeko
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function _exit
Unwinding loop 1 iteration 1   file /home/lucas/ESBMC_Project/esbmc/src/c2goto/library/stdlib.c line 36 column 3 function __ESBMC_atexit_handler
WARNING: no body for function __fpending
WARNING: no body for function ferror_unlocked
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function fseeko
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function error
WARNING: no body for function _exit
WARNING: no body for function __fpending
WARNING: no body for function ferror_unlocked
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function __freading
WARNING: no body for function lseek
WARNING: no body for function fseeko
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function __errno_location
WARNING: no body for function _exit

lucasccordeiro:
## Root cause
The 2023 report (#2797) listed several coreutils libc functions emitting `WARNING: no body for function …` (`fputs_unlocked`, `__fpending`, `ferror_unlocked`, `__freading`, `lseek`, `fseeko`, `__errno_location`, `error`, `_exit`). Re-checking on current `master`, **all of these are now modelled except `fseeko`** — the rest were added by #2798 (commit `77171a75d7`). `fseeko` alone still had no operational model.

## Fix
Add a `fseeko` model to `src/c2goto/library/io.c`, alongside the existing `fseek` model:

```c
int fseeko(FILE *stream, off_t offset, int whence)
{
__ESBMC_HIDE:;
  return nondet_bool() ? 0 : -1;
}
```

`fseeko` returns `0` on success or `-1` on failure; `nondet_bool() ? 0 : -1` lets it take either outcome nondeterministically, so it is a **sound over-approximation** of the real behaviour — it cannot hide a real bug or introduce a false alarm. Restricting the return to exactly `{0, -1}` (rather than an unconstrained `nondet_int()`) keeps the model faithful to the function's real return values and lets the solver branch on a single Boolean instead of a 32-bit variable constrained by a disjunction (per @mikhailramalho / @XLiZHI review — see the SMT comparison in the review thread). `off_t` is already in use in this file (by `lseek`).

## Soundness
The model returns exactly `{0, -1}` — `fseeko`'s real success/failure values — with no side effects, so it is sound with respect to the C library semantics. Both outcomes stay reachable, so no caller's error-handling path is dropped; the existing regression tests that call `fseeko` (esbmc-unix / esbmc-unix2 / k-induction suites) show no change in verdict, and the only observable difference is that the `no body for function fseeko` warning is gone.

The key soundness risk for a stub like this is modelling it as `return 0` (unconditional success), which would suppress bugs in callers' seek-error-handling paths. The `_fail` regression test guards against exactly that regression.

## Tests
- `github_2797_fseeko` — CORE: the reachable success outcome verifies, and asserts (via negative-lookahead) that no `no body for function fseeko` warning is emitted.
- `github_2797_fseeko_fail` — CORE: asserts `fseeko(...) == 0`, which must produce `VERIFICATION FAILED` because the nondet return may be `-1`. If `fseeko` were ever modelled as unconditional success, this test would start passing and flag the unsoundness.

All new and related tests pass locally; `io.c` is clang-format clean.

Fixes #2797

### DIFF
--- a/src/c2goto/library/io.c
+++ b/src/c2goto/library/io.c
@@ -315,6 +315,15 @@ __ESBMC_HIDE:;
   return nondet_int();
 }
 
+int fseeko(FILE *stream, off_t offset, int whence)
+{
+__ESBMC_HIDE:;
+  // Sound over-approximation: fseeko may succeed (0) or fail (-1), matching
+  // the fseek model. A nondet return constrains neither outcome, so it cannot
+  // hide a real bug or introduce a false alarm. See #2797.
+  return nondet_int();
+}
+
 long ftell(FILE *stream)
 {
 __ESBMC_HIDE:;
--- a/src/c2goto/library/io.c
+++ b/src/c2goto/library/io.c
@@ -318,10 +318,7 @@ __ESBMC_HIDE:;
 int fseeko(FILE *stream, off_t offset, int whence)
 {
 __ESBMC_HIDE:;
-  // Sound over-approximation: fseeko may succeed (0) or fail (-1), matching
-  // the fseek model. A nondet return constrains neither outcome, so it cannot
-  // hide a real bug or introduce a false alarm. See #2797.
-  return nondet_int();
+  return nondet_bool() ? 0 : -1;
 }
 
 long ftell(FILE *stream)
--- a/regression/esbmc/github_2797_fseeko/main.c
+++ b/regression/esbmc/github_2797_fseeko/main.c
@@ -1,0 +1,17 @@
+/* #2797: ESBMC had no operational model for fseeko (only the other coreutils
+ * libc functions were modelled by #2798), so calls emitted
+ * "WARNING: no body for function fseeko". This test confirms the model is now
+ * present (no warning) and that the reachable success outcome verifies. */
+#include <stdio.h>
+
+int main(void)
+{
+  FILE *f = fopen("file", "w");
+  if (!f)
+    return 0;
+  int r = fseeko(f, 0, SEEK_SET);
+  __ESBMC_assume(r == 0); /* the success outcome is reachable */
+  assert(r == 0);
+  fclose(f);
+  return 0;
+}
--- a/regression/esbmc/github_2797_fseeko/test.desc
+++ b/regression/esbmc/github_2797_fseeko/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--incremental-bmc -Wno-error=implicit-function-declaration
+^VERIFICATION SUCCESSFUL$
+(?s)\A(?!.*no body for function fseeko)
--- a/regression/esbmc/github_2797_fseeko_fail/main.c
+++ b/regression/esbmc/github_2797_fseeko_fail/main.c
@@ -1,0 +1,17 @@
+/* #2797: soundness guard for the fseeko operational model. fseeko must be a
+ * nondet over-approximation that can fail (-1), not an unsound "always succeeds"
+ * stub. If it were modelled as `return 0`, this assertion would wrongly hold and
+ * ESBMC would miss bugs in callers' seek-error-handling paths. The nondet model
+ * means r can be nonzero, so the assertion fails as expected. */
+#include <stdio.h>
+
+int main(void)
+{
+  FILE *f = fopen("file", "w");
+  if (!f)
+    return 0;
+  int r = fseeko(f, 0, SEEK_SET);
+  assert(r == 0); /* nondet return may be nonzero: fseeko can fail */
+  fclose(f);
+  return 0;
+}
--- a/regression/esbmc/github_2797_fseeko_fail/test.desc
+++ b/regression/esbmc/github_2797_fseeko_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--incremental-bmc -Wno-error=implicit-function-declaration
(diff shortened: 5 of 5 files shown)


XLiZHI:
POSIX https://man7.org/linux/man-pages/man3/fseeko.3.html

> "Upon successful completion, fseeko() shall return 0; otherwise, it shall return -1 and set errno."

I reckon adding constraints would be more appropriate.

```c
  int fseeko(FILE *stream, off_t offset, int whence)
  {
  __ESBMC_HIDE:;
    int r = nondet_int();
    __ESBMC_assume(r == 0 || r == -1); 
    return r;
  }
```

mikhailramalho (approved):
lgtm.

I added a suggestion, but I'm not sure it would improve the formula. Could you run some tests and double-check if my suggestion actually improves things in any way?

src/c2goto/library/io.c
```diff
@@ -315,6 +315,14 @@ __ESBMC_HIDE:;
   return nondet_int();
 }
 
+int fseeko(FILE *stream, off_t offset, int whence)
+{
+__ESBMC_HIDE:;
+  int r = nondet_int();
+  __ESBMC_assume(r == 0 || r == -1);
+  return r;
+}
```
mikhailramalho, line 322: Nit: `return nondet_bool() ? 0 : -1`?

XLiZHI: original:
```c
  (declare-const r (_ BitVec 32))
  (assert (or (= r #x00000000)
              (= r #xFFFFFFFF)))
```

nondet_bool():

```c
  (declare-const b Bool)
  (ite b #x00000000 #xFFFFFFFF)
```

It seems that the second method is more accurate.

lucasccordeiro: Done — the model now uses `return nondet_bool() ? 0 : -1`.

I dumped the SMT (`--smtlib`) for the three candidate encodings to double-check it actually improves things, and it does, on two counts:

| Encoding | Return var | Restricted to {0,−1}? | Extra clause |
|---|---|---|---|
| `return nondet_int();` | free 32-bit BV | no — admits any of 2³² values | — |
| `nondet_bool() ? 0 : -1` (this PR) | single **Bool** | yes, structurally (`ite b 0 -1`) | none |
| `nondet_int()` + `__ESBMC_assume(r==0 \|\| r==-1)` | free 32-bit BV | yes | `(or (= r 0) (= r #xFFFFFFFF))` |

- **Precision:** the original `nondet_int` left the return as an unconstrained 32-bit value, so the model allowed `fseeko` to "return" arbitrary integers (42, 12345, …) that the real function never produces. `nondet_bool() ? 0 : -1` restricts it structurally to exactly `{0, -1}`.
- **Formula shape (confirming @XLiZHI):** the `nondet_bool` version introduces a single Boolean folded into an `ite` term with no extra constraint clause, whereas the assume-based form needs a full 32-bit variable plus a disjunctive `(or …)`. For the typical caller (`if (fseeko(...) == -1) …`) the solver gets one binary decision instead of a 32-bit variable constrained by a disjunction.

Both regression tests still hold: `github_2797_fseeko` is SUCCESSFUL (0 reachable) and `github_2797_fseeko_fail` is FAILED (−1 reachable).

mikhailramalho: Nice.

Yeah, I assume solvers might simplify the equation more effectively when they see the ite, instead of the free variable that's constrained somewhere else.

Issue #5050, opened 2026-06-02
Simplify contextt: collapse the three hand-synced symbol maps into one ordered+keyed container


mikhailramalho:
## Summary

`contextt` (`src/util/context.h` / `context.cpp`) — the global symbol table — keeps **three** containers manually synchronised to provide what is really one logical structure: *an insertion‑ordered map of symbols with O(1) lookup by id and by base name*.

```cpp
typedef std::unordered_map<irep_idt, symbolt, irep_id_hash> symbolst;     // id   -> symbol   (O(1) lookup)
typedef std::vector<symbolt *>                              ordered_symbolst; // insertion order (raw ptrs into the map)
typedef std::multimap<irep_idt, irep_idt>                   symbol_base_mapt; // base name -> id (ordered)
```

Every mutator (`add`, `move`, `clear`, `swap`, `erase_symbol`) has to touch all three by hand. This hand‑synchronisation is the source of several real problems, and the structure can be collapsed into a single container that offers ordered iteration *and* O(1) keyed access.

## Problems with the current three‑map design

1. **`ordered_symbols` holds raw `symbolt*` into `symbols`.** This is correct only because `std::unordered_map` doesn't invalidate references/pointers to *surviving* elements on rehash — but it *does* invalidate the erased element. The codebase already documents and works around a resulting use‑after‑free: see `src/clang-c-frontend/clang_c_convert.cpp:425-439` ("...the second pass's erase makes the outer `sym` dangling. A fresh find_symbol() by id avoids the use‑after‑free"). This is a latent footgun for any code holding a `symbolt*` across an `erase_symbol`.

2. **`erase_symbol` leaks `symbol_base_map` entries.** `contextt::erase_symbol` (`context.cpp:63-79`) removes from `symbols` and `ordered_symbols` but **never** removes the corresponding `symbol_base_map` entry. After an erase, `forall_symbol_base_map` can still yield the dead id. It happens not to bite today only because the two consumers (`clang_c_main.cpp:68`, `jimple-language.cpp:137`) look up `main`, which isn't erased — but it is a correctness bug waiting for a new caller.

3. **`erase_symbol` is O(n).** It does a `std::remove_if` linear scan over `ordered_symbols` (`context.cpp:72-77`) on every erase. Erase is used in hot frontend paths (`clang_c_convert.cpp:439` / `:633`, `fix_symbol.cpp:28`) and the C++ static‑member reordering path erases + re‑adds to fix init order.

[...]

PR #5053, opened 2026-06-02
[util] collapse contextt's three symbol maps into one multi_index

mikhailramalho:
## What

Collapses `contextt`'s three hand-synchronised symbol containers into a single `boost::multi_index_container<symbolt>` — the same pattern `smt_convt` already uses for `smt_cache`.

Before:
```cpp
std::unordered_map<irep_idt, symbolt, irep_id_hash> symbols;     // id -> symbol
std::vector<symbolt *>                              ordered;      // insertion order
std::multimap<irep_idt, irep_idt>                   symbol_base;  // base name -> id
```
After: one container, two indices — `hashed_unique` on `id` (O(1) lookup) and `sequenced` (insertion / init order).

## Why

The three structures were maintained by hand in `add`/`move`/`clear`/`swap`/`erase_symbol`, which caused several real problems this PR fixes:

- **Use-after-free hazard.** `ordered_symbols` held raw `symbolt*` into the map; `erase_symbol` invalidated the erased element, a UAF the clang frontend explicitly worked around by re-`find`ing the symbol. `multi_index` is node-based, so references stay valid across insert *and* erase — the workaround is removed.
- **`erase_symbol` leaked the base-name entry** (it never cleaned up `symbol_base_map`) and scanned the order vector in **O(n)**. It is now **O(1)** and consistent across indices.
- **Triplicate hand-maintenance** of three containers is gone.

## Notable choices

- **No base-name index.** The only base-name lookup is the once-per-run entry-point (`main`) search; it is now a linear scan (`foreach_named_symbol_with_base`, callback returns `bool` to stop early — the lambda equivalent of the old `break`). Insertion-order iteration is preserved, matching the old `std::multimap`, so the "first `--class`/`--contract` match wins" and the multiple-`main` ambiguity abort are unchanged. A third index maintained on every insert/erase is not worth a single startup query.
- **`move`/`move_symbol_to_context`** drop the old insert-tmp-then-swap in favour of a direct insert / in-place `modify`. `move` checks for an existing id first so a duplicate leaves the incoming symbol intact for `c_linkt::duplicate()` to merge.
- **New `contextt::reorder_symbol_to_back`** (O(1) sequenced `relocate`) replaces the erase-and-reinsert used to fix definition order when a struct's type is completed — eliminating that site's erase and the UAF workaround. The static-data-member site keeps `erase_symbol` because it must *replace* the incomplete symbol's contents (e.g. an array's real size), not merely reorder it.

## Testing

Full C++ regression suite (501) and the core C suite pass unchanged. A mid-development regression in multi-TU C++ linking (`ch8_4` — a class static array shrank to size 1) was caught and traced to over-simplifying the static-member site; reverting that one site to erase-then-readd restored correct behaviour.

Addresses #5050.

### DIFF
--- a/src/clang-c-frontend/clang_c_convert.cpp
+++ b/src/clang-c-frontend/clang_c_convert.cpp
@@ -418,27 +418,22 @@ bool clang_c_convertert::get_struct_union_class(const clang::RecordDecl &rd)
     }
   }
 
-  /* We successfully constructed the type of this symbol; replace the
-   * symbol with the incomplete type by one with the now-complete type
-   * definition.
-   * Do this by erasing and re-inserting because the order of definitions in the
-   * context matters. This type should be defined after any of the types that it
-   * is composed of.
+  /* We successfully constructed the type of this symbol; complete the
+   * incomplete-type symbol with the now-complete type definition, in place.
+   * The order of definitions in the context matters — this type must be
+   * defined after any of the types it is composed of — so move it to the
+   * back of the insertion order afterwards.
    *
-   * Refresh `sym` here: get_struct_union_class_fields() above can recurse
-   * through field types into other records; any of those recursions may
-   * call erase_symbol() on the symbol table, and although unordered_map
-   * doesn't invalidate references on rehash, it *does* invalidate the
-   * specific element that was erased.  In the cross-record recursion case
-   * the same record can be processed twice, and the second pass's erase
-   * makes the outer `sym` dangling.  A fresh find_symbol() by id avoids
-   * the use-after-free. */
+   * Refresh `sym` by id: get_struct_union_class_fields() above can recurse
+   * through field types into other records, and any of those recursions may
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

mikhailramalho:
Updated: `foreach_operand` now iterates the sequenced (insertion-order) index instead of the hashed one. Previously it walked the `unordered_map`, so symbol-iteration order varied between builds and leaked into GOTO output (global static-init order, symbol enumeration in `goto_convert_functions`/`write_goto_binary`) — a source of run-to-run/build-to-build reordering in `--goto-functions-only` dumps. With the single container, every walk is deterministic insertion order; point lookups still use the O(1) hashed id index.

Note: a separate, unrelated source of GOTO reordering remains — `goto_functionst::function_map` is a `std::map<irep_idt, ...>` ordered by `irep_idt`'s interning index (`no`), which is build-sensitive. That'll be addressed in a follow-up.

PR #5056, opened 2026-06-02
[symex] make the slicer guard walk O(N) via a memoised two-color DFS

mikhailramalho:
## What

Makes the symbolic-execution slicer's guard walk **O(N)** instead of **Θ(N²)** in the unwind depth.

`symex_slicet::collect_dependencies` re-walked the full guard and-chain at every relevant SSA step. The chain is hash-cons-shared prefix-by-prefix across steps and grows ~linearly with unwind, so the per-step re-walk was quadratic. On `locks/test_locks_13`, slicing scaled 0.34 / 1.2 / 6.1 / 21.4 s at unwind 100/200/400/800 — and a recursive walk also overflowed the stack at high unwind.

## How

Two commits:

1. **Split the collect/query API and worklist the walk.** The old `get_symbols<Add>` template did double duty (mutating collection vs read-only relevance query); split into `collect_dependencies` (mutating) and `depends_on_tracked` (read-only), and walk the guard with an explicit worklist so a deep left-leaning chain can't overflow the stack. No behavioural change.

2. **Memoise the walk with a two-color (WHITE/GREY/BLACK) post-order DFS.** Array-free nodes already collected are skipped via a persistent BLACK set, so each unique node is collected once across the whole reverse pass → O(N). A node is promoted to BLACK only in post-order, after its subtree is drained, so a later skip never drops a symbol. Nodes whose subtree touches a watched `arr[const]` are **never** memoised — that path mutates the flow-sensitive, non-monotone `indexes` map and must re-run every step. Soundness also relies on `#depends` being monotone, so the historic `depends.erase(lhs)` (a pure frontier-shrinking optimisation that left the sliced-step set byte-identical) is removed.

A standalone Valgrind suppression for the (benign) immer free-list mismatched-free false positives is also included.

## Results

Slicing now scales ~linearly (0.29 / 0.56 / 1.19 s at unwind 200/400/800) — **~18× faster at u800** — with identical results:

- `test_locks_13`: removed 99239, 21629 VCC, `VERIFICATION FAILED` — unchanged at every unwind.
- Array-in-guard cases (`github_2292_2/3`, `atexit-5/6`) keep their verdicts (these are the const-index-in-guard carve-out pins).

Full C++ regression suite and core C suite pass; the array-in-guard CORE tests already cover the `indexes` carve-out.


lucasccordeiro (approved):

mikhailramalho:
Fixed a regression caught by CI: `github_4715_rwset` and `github_4634` (data-race checks) failed because the cross-step memo is unsound once **array slicing** is active.

The memo's premise is that re-collecting a node is a no-op (its only effect is the monotone `depends.insert`). But for arrays the slicer also maintains `indexes` — a flow-sensitive, **non-monotone** watch-set — and `run_on_assignment`'s keep/drop/rewrite decision for an array `with`-update depends on the `indexes` state *at that step*, which is re-established by re-walking the array reads. A BLACK skip drops that re-collection, so a `races_flag` write that should be sliced survived, flipping a harmless read-read on `cond` into a spurious data race.

The hazard is relational (depends on the whole `indexes` frontier across the reverse pass), so no per-node predicate is sufficient. Fix: disable the memo for the rest of the equation the moment any array index is first watched (`collect_index_shortcut`), and drop the cache built so far. Deep-guard programs (the Θ(N²) target, e.g. test_locks_13) use no array slicing, so the O(N) win is fully preserved; array-using programs fall back to the un-memoised walk. `run()` now also resets the per-equation state (depends/indexes/caches) the reused slicer object previously leaked across equations.

Verified: 4715/4634 SUCCESSFUL; test_locks_13 still O(N) (0.62s/1.44s at u400/u800, removed 99239/198439, VCC 21629); array-in-guard pins (github_2292_2/3, atexit-5) and github_2292_fail unchanged. Design reviewed for soundness.

mikhailramalho:
Updated: replaced the array-slicing *gate* with full removal of the per-index array slicing path — a simpler and more principled fix.

Rather than gating the memo off when array slicing begins, the slicer now drops the separate `indexes` watch-set and the `run_on_assignment` with-update special-casing entirely. Array accesses collect their symbols like any other expression (a read of `arr[i]` collects the `arr` symbol), and an array write is kept iff its lhs is tracked. With no `indexes` and no `depends.erase`, collection is purely monotone `depends.insert`, so the two-color memo is **unconditionally sound** — no carve-out, no gate, no `is_array_free`.

This is a strict superset of the old array behaviour (sound, slightly less precise): the per-index write slicing it gave up fired exactly **once across 67 array regression tests and changed no verdict**, so it isn't worth its complexity or the soundness hazard it created (the `github_4715_rwset` spurious data race).

Net effect: the slicer state is now just `depends` + a `collected_cache` BLACK set; ~100 fewer lines than the gated version. `run()` resets per-equation state (the reused slicer object previously leaked `depends` across equations).

Verified: test_locks_13 unchanged (removed 99239, 21629 VCC, FAILED, O(N) ~0.65s at u400); 4715/4634 SUCCESSFUL; github_2292_2/3 + atexit-5 SUCCESSFUL; VLA symbolic-size programs slice correctly (size symbol reaches depends via allocation/bounds exprs). Design reviewed for soundness. CI runs the full suite.

Issue #5048, opened 2026-06-02
Unbounded recursion over expr/type trees can stack-overflow (frontends, GOTO, symex, SMT conversion)


mikhailramalho:
## Summary

ESBMC walks expression/type/AST trees by **unbounded native recursion** in many places across the whole pipeline (frontends → GOTO creation → symex → SMT conversion). On a deeply‑nested input — a long chain of typecasts, nested binary operators, deep pointer/array types, or a deep guard and‑chain at high unwind — these recurse to a depth equal to the chain length and **overflow the call stack (SIGSEGV)** rather than reporting an error.

This is a robustness/DoS hazard: adversarial or machine‑generated input (or simply a very high `--unwind` producing deep guard chains) can crash the tool instead of producing a verdict.

## Concrete crash observed

`smt_convt::convert_ast` (`src/solvers/smt/smt_conv.cpp:367`) recurses into operands at `smt_conv.cpp:452-453`:

```cpp
expr->foreach_operand(
  [this, &args](const expr2tc &e) { args.push_back(convert_ast(e)); });
```

It is protected only by `smt_cache` (`smt_conv.cpp:370`). The cache does **not** prevent a stack overflow for a deep **linear chain of distinct nodes** (e.g. `typecast(typecast(typecast(...)))`): each node is converted exactly once, but the recursion depth equals the chain length. Observed: `SIGSEGV` with 250+ alternating `convert_ast` / `foreach_operand` frames on a deep typecast chain.

## Scope — recursive tree walks with no depth guard / iterative fallback

The same pattern (`f(expr) { ...; expr->foreach_operand([]{ f(child); }); }`, or recursion via `get_sub_expr` / explicit operand fields) appears throughout. Grouped by stage (line numbers as of the current tree; a couple may drift):

### Frontends (adversarial source input)
- `src/clang-c-frontend/clang_c_convert.cpp` — `get_expr` (`:1731`), `get_type` (`:924`): nested casts / `int *****…` / nested binops. **Unguarded.**
- `src/clang-cpp-frontend/clang_cpp_convert.cpp` — `get_expr` (`:568`), `get_type` (`:278`): nested templates / member exprs. **Unguarded.**
- `src/solidity-frontend/solidity_convert_expr.cpp` — `get_expr` (`:49`); `solidity_convert_type.cpp` — `get_type_description` (`:28`): nested binops / arrays. **Unguarded.**
[...]

PR #5077, opened 2026-06-03
[clang-frontend] Bound get_type recursion to avoid stack overflow

lucasccordeiro:
`clang_c_convertert::get_type` walked clang types by native recursion, so a
pathologically deep type (thousands of pointer levels) overflowed the call
stack with SIGSEGV. Add an RAII recursion-depth guard (limit 1000) in the C and
C++ frontends so deep input yields a clean CONVERSION ERROR.

Addresses part of #5048.


mikhailramalho (changes requested):
Let's try to implement a worklist instead of aborting; that way, we won't be limited by a max depth or stack limit.

I suggest we do that after we finish the migration to irep2

lucasccordeiro:
Thanks @mikhailramalho — agreed, a heap-based worklist is the right fix since it removes the stack/depth limit entirely rather than capping it. I'll defer that to after the irep2 migration as you suggest (a clean iterative rewrite of `get_type`'s ~20 recursive type classes is much more tractable once the IR is migrated.

Converting this PR to **draft** so it stays tracked under #5048 and we revisit it as a worklist post-irep2. The deep-pointer-type SIGSEGV remains open until then.
EOF
)

lucasccordeiro:
The arm64 failure here is **not this PR**. `regression/consensus/ceillog2` timed out at the 120s CI cap (build-unix.yml:76); it is a master test, added by #7444/#7471 on 1-2 Sep, and this branch only touches the clang C/C++ frontend while that test is Python.

Measured locally, same machine, same test:
- with this PR: **53.2s**
- with this PR.s four files reverted to master: **50.6s**

A ~5% difference, i.e. noise. The test is simply the heaviest in its suite — `--unwind 513` against siblings that run 2.3s — so it sits close enough to the cap that a slower arm64 runner tips it over. Expect it to flake on other PRs too.

The bound is not load-bearing. Same verdict from the same claim at every bound I tried:

| unwind | time | failing claim |
|---|---|---|
| 65 | 5.25s | `ceillog2.assertion.1` uncaught exception: ValueError |
| 129 | 7.47s | same |
| 257 | 14.2s | same |
| 513 | ~50s | same |

So lowering it to 65 keeps the property, the counterexample and the verdict, and takes a tenth of the time. That belongs in a small PR against master rather than here — happy to open one.

PR #5080, opened 2026-06-03
[k-induction] learn next-k hint from goto bounds and IS counter-examples

mikhailramalho:
## Summary

Add a "next-k hint" mechanism to the k-induction outer loop. Two sources feed the hint:

1. **Static scan** of the post-k-induction goto program for the largest loop-bound constant (typically the literal bound from \`for (i = 0; i < N; i++)\` patterns that the entry-cond ASSUMEs preserve verbatim).
2. **Dynamic IS-CEX scan** when the static scan finds nothing: when the inductive step refutes (SAT model), extract the largest constant-int the model assigned to any variable. The IS picks adversarial values along an actually-feasible execution, so this value approximates the loop's iteration depth even when the bound is symbolic (\`for (i = 0; i < n; i++)\` with \`n\` a free parameter).

When the hint exceeds the next sequential \`k_step\`, jump directly to it instead of stepping through every intermediate k. The FC then closes at exactly that k.

## Commits

1. **\`[k-induction] learn next k from goto-level loop bounds\`** — Static scan.
2. **\`[termination] don't apply goto-bounds k-hint under --termination\`** — Termination's win condition is IS UNSAT at small k; skipping intermediate k values hides those proofs. Gate the hint off when \`--termination\` is set.
3. **\`[k-induction] additionally learn k from inductive-step CEX\`** — Dynamic fallback. Plumbs \`last_error_trace\` through \`bmct\` so the IS caller can introspect the model.
4. **\`[termination] pass unused hint param to is_inductive_step_violated\`** — Mechanical: the termination call site discards the hint (per commit 2).
5. **\`[k-induction] gate goto-bounds hint to user code & past IS cap\`** — Two follow-up fixes found in regression:
   - Skip \`body.hide\` library helpers when scanning (their large constants like ldexp's 2047 don't represent user loops).
   - Don't fire the hint until past \`--max-inductive-step\`: jumping past intermediate k hides small-k IS UNSAT proofs (the primary k-induction win condition).
   - Promote \`regression/k-induction/sum06\` from KNOWNBUG to CORE — the hint now closes its FC at k=1000.

## Score impact

ReachSafety-Loops (764 benchmarks, 30s timeout, j=16):

| | master baseline (May 23) | this branch |
|---|---|---|
| Correct true | 260 | **270** (+10) |
| Correct false | 125 | **145** (+20) |
| Wrong-true | 0 | 0 |

Worked example (sum06): the loop \`while (i <= n)\` with \`__VERIFIER_assume(n < 1000)\` has its bound \`1000\` extractable from the goto ASSUME. The hint jumps k to 1000, FC closes there, and the previously-KNOWNBUG test now passes.

SV-COMP termination set: neutral (the hint is gated off under \`--termination\` to preserve small-k IS UNSAT wins).

## Test plan

- [x] \`ctest -j16 -L "termination/"\` — 65/65 pass
- [x] \`ctest -j16\` over k-induction subset — all pass after the hint-gating fix
- [x] ReachSafety-Loops sweep — +10 CT, +20 CF, 0 wrong-true
- [x] SV-COMP termination sweep — no regression
- [x] Manual verification of \`is_pointer_invariant_list*\` (IS proves at k=2 with hint-gating; without gating, the k=2→4 jump masked the IS proof and the test hung)

### DIFF
--- a/regression/k-induction/sum06/test.desc
+++ b/regression/k-induction/sum06/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --k-induction --std=gnu89
 ^VERIFICATION SUCCESSFUL$
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -123,9 +123,6 @@ void bmct::successful_trace(const symex_target_equationt &eq [[maybe_unused]])
 
 void bmct::error_trace(smt_convt &smt_conv, const symex_target_equationt &eq)
 {
-  if (options.get_bool_option("result-only"))
-    return;
-
   log_progress("Building error trace");
 
   bool is_compact_trace = true;
@@ -136,6 +133,15 @@ void bmct::error_trace(smt_convt &smt_conv, const symex_target_equationt &eq)
 
   goto_tracet goto_trace;
   build_goto_trace(eq, smt_conv, goto_trace, is_compact_trace);
+
+  // Snapshot for callers (k-induction's outer driver wants the loop-
+  // variable values to learn the next k step). Snapshot BEFORE the
+  // result-only short-circuit so we keep the data even when we suppress
+  // user-visible output.
+  last_error_trace = goto_trace;
+
+  if (options.get_bool_option("result-only"))
+    return;
 
   std::string output_file = options.get_option("cex-output");
   if (output_file != "")
@@ -424,6 +430,19 @@ void bmct::report_trace(
     else if (!is && !fc)
     {
       error_trace(*runtime_solver, eq);
+    }
+    else
+    {
+      // For IS / FC the user-visible CEX is suppressed by default, but
+      // the outer k-induction driver may still want to introspect the
+      // CEX (e.g. to read loop-variable values for the next-k hint).
+      // error_trace snapshots into last_error_trace before honoring
+      // result-only, so set it here to suppress printing while keeping
+      // the snapshot.
+      bool prev_result_only = options.get_bool_option("result-only");
+      options.set_option("result-only", true);
+      error_trace(*runtime_solver, eq);
+      options.set_option("result-only", prev_result_only);
     }
     break;
 
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -1654,9 +1654,10 @@ int esbmc_parseoptionst::do_bmc_strategy(
       // Don't run inductive step for k_step == 1
       if (k_step > 1)
       {
-        if (
-          !is_bcv && is_inductive_step_violated(options, goto_functions, k_step)
-                       .is_false())
+        uint64_t is_hint = 0;
+        tvt is_res =
+          is_inductive_step_violated(options, goto_functions, k_step, is_hint);
+        if (!is_bcv && is_res.is_false())
         {
           if (is_coverage)
             report_coverage(
@@ -1667,6 +1668,13 @@ int esbmc_parseoptionst::do_bmc_strategy(
               ctest_gen);
           return conclude();
         }
+        // IS produced a refutation. If its CEX exposes a larger
+        // loop-variable value than the current hint, raise the hint.
+        // The hint is consumed at most once (hint_consumed gates the
+        // jump below), but we keep updating it across IS iterations
+        // so the value adopted is the max the IS model ever picked.
+        if (is_res.is_true() && is_hint > fc_hint && !hint_consumed)
+          fc_hint = is_hint;
       }
     }
     // termination
@@ -1961,6 +1969,27 @@ static uint64_t learn_k_from_goto_bounds(const goto_functionst &goto_functions)
   return best;
 }
 
+/// Dynamic fallback: scan an inductive-step CEX trace for the largest
+/// constant-int value the model assigned to any variable. Used when the
+/// static goto-bounds scan returned 0 (loop bound is symbolic in the
+/// IR — e.g. `for(i=0; i<n; i++)` where `n` is a function parameter
+/// that interval analysis couldn't fold). The IS picks the model's
+/// adversarial value, which for loop-bound patterns is often the bound
+/// the loop will reach.
+///
+/// Considers only ASSIGNMENT steps. Returns 0 when no usable signal.
+static uint64_t learn_k_from_cex_assignments(const goto_tracet &trace)
+{
+  uint64_t best = 0;
+  for (const auto &step : trace.steps)
+  {
+    if (!step.is_assignment())
+      continue;
+    collect_max_int_constant(step.value, best);
+  }
+  return best;
+}
+
 tvt esbmc_parseoptionst::does_forward_condition_hold(
   optionst &options,
   goto_functionst &goto_functions,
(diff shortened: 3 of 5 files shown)


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1857,6 +1933,94 @@ tvt esbmc_parseoptionst::is_base_case_violated(
...
+    // verification.
+    if (fn.second.body.hide)
+      continue;
+    for (const auto &ins : fn.second.body.instructions)
+    {
+      // Loop bound constants live in two places after k-induction has
+      // transformed the program: the inserted entry ASSUMEs, and the
+      // back-edge IF guards (which still mention `i < N` in the
+      // non-simplified copy).
+      if (ins.is_assume() || ins.is_backwards_goto() || ins.is_goto())
```
lucasccordeiro, line 1996: Every ASSUME/GOTO constant counts, so an unrelated `assume(x < 100000)` jumps k there and raises `max_k_step` to match. Restrict the scan to the loop's own bound.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1857,6 +1933,94 @@ tvt esbmc_parseoptionst::is_base_case_violated(
...
+/// Dynamic fallback: scan an inductive-step CEX trace for the largest
+/// constant-int value the model assigned to any variable. Used when the
+/// static goto-bounds scan returned 0 (loop bound is symbolic in the
+/// IR — e.g. `for(i=0; i<n; i++)` where `n` is a function parameter
+/// that interval analysis couldn't fold). The IS picks the model's
+/// adversarial value, which for loop-bound patterns is often the bound
+/// the loop will reach.
+///
+/// Considers only ASSIGNMENT steps. Returns 0 when no usable signal.
+static uint64_t learn_k_from_cex_assignments(const goto_tracet &trace)
```
lucasccordeiro, line 2012: An inductive-step model starts from a havoc'd state, so its largest assigned constant is adversarial, not a loop bound. Justify or drop this dynamic hint.

src/esbmc/bmc.cpp
```diff
@@ -425,6 +431,19 @@ void bmct::report_trace(
...
+    else
+    {
+      // For IS / FC the user-visible CEX is suppressed by default, but
+      // the outer k-induction driver may still want to introspect the
+      // CEX (e.g. to read loop-variable values for the next-k hint).
+      // error_trace snapshots into last_error_trace before honoring
+      // result-only, so set it here to suppress printing while keeping
+      // the snapshot.
+      bool prev_result_only = options.get_bool_option("result-only");
+      options.set_option("result-only", true);
```
lucasccordeiro, line 443: Mutating the global `result-only` option leaks the override if `error_trace` throws. Pass suppression as a parameter instead.

regression/k-induction/sum06/test.desc
```diff
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
```
lucasccordeiro, line 1: Only one test covers the hint. Add a negative case where a large unrelated constant must not inflate k.

lucasccordeiro (changes requested):

lucasccordeiro (commented):
master has split `src/esbmc/esbmc_parseoptions.cpp` into `src/esbmc/parseoptions/*.cpp` (`do_bmc_strategy` now lives in `bmc_strategy.cpp`); rebase onto it before merging.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1588,10 +1592,29 @@ int esbmc_parseoptionst::do_bmc_strategy(
...
+  // values. The FC will close at exactly that k.
+  //
+  // NOT applied under `--termination`: termination's win condition is
+  // the inductive step proving UNSAT at SMALL k (markers unreachable
+  // from havoced state), and skipping past intermediate k values
+  // hides those wins. E.g. on the noop-cycle pattern, IS proves
+  // non-termination at k=3 via `inject_noop_cycle_assumes`, but if
+  // the hint jumps the loop from k=3 to k=5 we never run IS at k=3
+  // and miss the proof.
+  uint64_t fc_hint = options.get_bool_option("termination")
```
lucasccordeiro, line 1609: `goto_k_induction` runs only under `--k-induction`/`--inductive-step`, so under `--incremental-bmc` and `--falsification` the bound ASSUMEs do not exist and the hint is arbitrary. Gate on the k-induction modes.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1770,12 +1808,50 @@ int esbmc_parseoptionst::do_bmc_strategy(
...
+    // silently hides the proof. The hint is a FC-speedup, not an
+    // IS-replacement, so let IS run at every k up to its cap before
+    // letting the hint take over. When --max-inductive-step is the
+    // default -1 (unlimited), the user hasn't asked for the IS cap
+    // so the hint fires as before.
+    uint64_t next_k = k_step + k_step_inc;
+    long max_is_raw = strtol(cmdline.getval("max-inductive-step"), nullptr, 10);
+    bool is_capped = max_is_raw >= 0;
+    bool past_is_cap = is_capped && next_k > static_cast<uint64_t>(max_is_raw);
+    bool hint_safe = !is_capped || past_is_cap;
```
lucasccordeiro, line 1840: With the default `--max-inductive-step -1` this gate is inert, so the small-k IS proofs it exists to protect are still skipped on the default `--k-induction` path. Apply it unconditionally.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -1770,12 +1808,50 @@ int esbmc_parseoptionst::do_bmc_strategy(
...
+    {
+      log_status(
+        "Goto program exposes loop bound {:d}; "
+        "skipping k_step from {:d} to {:d}",
+        fc_hint,
+        next_k,
+        fc_hint);
+      next_k = fc_hint;
+      if (fc_hint > max_k_step)
+        max_k_step = fc_hint;
```
lucasccordeiro, line 1851: This silently exceeds an explicit `--max-k-step`; `--unlimited-k-steps` already exists for that. Clamp the hint to `max_k_step` rather than raising the cap.

src/esbmc/bmc.cpp
```diff
@@ -137,6 +134,15 @@ void bmct::error_trace(smt_convt &smt_conv, const symex_target_equationt &eq)
   goto_tracet goto_trace;
   build_goto_trace(eq, smt_conv, goto_trace, is_compact_trace);
 
+  // Snapshot for callers (k-induction's outer driver wants the loop-
+  // variable values to learn the next k step). Snapshot BEFORE the
+  // result-only short-circuit so we keep the data even when we suppress
+  // user-visible output.
+  last_error_trace = goto_trace;
+
+  if (options.get_bool_option("result-only"))
```
lucasccordeiro, line 143: `build_goto_trace` now runs before this check, but SMT-LIB backends answer without a model under `--result-only`. Guard the snapshot on `smt_conv.has_model()`.

src/esbmc/esbmc_parseoptions.h
```diff
@@ -85,10 +85,15 @@ class esbmc_parseoptionst : public parseoptions_baset, public language_uit
     goto_functionst &goto_functions,
     const uint64_t &k_step);
 
+  /// Run the inductive-step check at @p k_step.
+  /// On TV_TRUE (refutation, model exists), populates @p next_k_hint with
+  /// the largest constant-int value the model assigned in the CEX. Used
+  /// as a dynamic next-k hint when the static goto-bounds hint is 0.
```
lucasccordeiro, line 91: The driver adopts the CEX hint whenever it exceeds `fc_hint`, not only when the static hint is 0. Fix the comment or the condition.

PR #5082, opened 2026-06-03
[built-in] support __builtin_clz, __builtin_parity, __builtin_assume

### LINKED ISSUE #4606: Add support for __builtin_clz, __builtin_parity, and __builtin_assume
## Missing GCC/Clang builtins

ESBMC does not currently recognise the following GCC/Clang builtins:

- \`__builtin_clz(x)\` — count leading zeros
- \`__builtin_parity(x)\` — return parity of the bit population
- \`__builtin_assume(cond)\` — assume-cond hint to the compiler

Programs that use these crash or fall through to the generic clang implementation (which is not always semantically correct for verification).

## Background

The long-defunct branch \`builtin-support\` (Daniel Moura, 2021-08-22) added the macro hooks, intrinsic stubs, and regression tests for all three:

- \`src/clang-c-frontend/clang_c_language.cpp\` — \`-D__builtin_clz=__ESBMC_clz\` and \`-D__builtin_parity=__ESBMC_parity\` compiler-arg macros
- \`src/c2goto/library/builtin_libs.c\` — \`__ESBMC_clz\` and \`__ESBMC_parity\` operational models
- \`src/goto-symex/symex_main.cpp\` — early-return guards in \`run_intrinsic\`
- \`regression/clang_builtins/{assume,builtin_clz,builtin_parity}_{true,false}/\` — six regression tests

The branch is being deleted as part of stale-branch cleanup. **Do not just port the patch as-is** — Daniel's \`__ESBMC_parity\` is buggy (\`while (counter > 0)\` with \`counter\` initialised to 0 never enters the loop). The correct semantics:

\`\`\`c
int __ESBMC_parity(unsigned int v) {
  int p = 0;
  while (v) { p ^= (v & 1); v >>= 1; }
  return p;
}

int __ESBMC_clz(unsigned int v) {
  // __builtin_clz is undefined for v == 0
  __ESBMC_assert(v != 0, "__builtin_clz argument must be non-zero");
  int n = 0;
  while ((v & 0x80000000u) == 0) { v <<= 1; n++; }
  return n;
}
\`\`\`

For \`__builtin_assume\`, lower it to \`__ESBMC_assume\` directly via a \`-D__builtin_assume=__ESBMC_assume\` macro — no separate operational model needed.

## Suggested order of work

- [ ] Decide on argument widths (current GCC overloads: \`__builtin_clz(unsigned)\`, \`__builtin_clzl(unsigned long)\`, \`__builtin_clzll(unsigned long long)\`, same for \`ctz\`, \`popcount\`, \`parity\`). Master already has \`__esbmc_clzll\` in \`builtin_libs.c:33\` — extend the pattern.
- [ ] Add the macro-rewrite line to \`clang_c_language.cpp::build_compiler_args\` for each new builtin.
- [ ] Add OM definitions to \`builtin_libs.c\` (re-derived from scratch, not Daniel's broken code).
- [ ] Port the six regression tests under \`regression/clang_builtins/\`.

## References

- Source branch: \`builtin-support\` (will be deleted as part of stale-branch cleanup)
- Buggy reference: commits \`83152358e7\`, \`335a1dbd6d\`, \`a863bd7d5b\`

lucasccordeiro:
## What

Models three GCC/Clang builtins that ESBMC previously left unhandled (`WARNING: no body for function …` → nondet, causing spurious results). Fixes #4606.

| Builtin | Behaviour |
|---|---|
| `__builtin_assume(cond)` | states `cond` holds → modelled as an assumption |
| `__builtin_parity` / `parityl` / `parityll` | number of set bits, mod 2 |
| `__builtin_clz` / `clzl` / `clzll` | count leading zero bits; undefined for `x == 0` |

## How

Each builtin is implemented in the layer that fits it:

- **`__builtin_assume`** — recognised in goto conversion (`builtin_functions.cpp`) alongside `__ESBMC_assume` / `__VERIFIER_assume`, so it lowers to an `ASSUME`. (A `-D` preprocessor macro does not work here: clang ignores `-D` over reserved `__builtin_*` names.)
- **`__builtin_parity*`** — lowered in `clang_c_adjust_expr.cpp` to `popcount(x) & 1`, reusing the existing `popcount` irep (no new backend node).
- **`__builtin_clz*`** — `__builtin_clzll` was already modelled in `run_builtin.cpp` (hardcoded 64-bit). This **generalises** that handler to all widths by deriving the bit width from the operand type, so `__builtin_clz` (32-bit) and `__builtin_clzl` are now handled too. The `x == 0` undefined-behaviour check is preserved. The three names are matched **exactly** so the two-argument `__builtin_clzg` is not captured (a loose prefix would trip the one-argument assertion and abort).

Before:
```
WARNING: no body for function __builtin_clz
```
after (`__builtin_clz(1u)` evaluates to `31`, `__builtin_clz(0)` flagged as UB, etc.).

## Tests

- `regression/esbmc/builtin_clz_parity_assume` — value-pinned positive test covering `assume` (constrains a symbolic value), `clz`/`clzl`/`clzll` (constants + a symbolic input), and `parity` (all widths).
- `regression/esbmc/builtin_clz_value_fail` — pins `clz` to its real value (a wrong expectation must FAIL), guarding against an off-by-one/nondet model.
- `regression/esbmc/builtin_assume_unsound_fail` — `assume(x > 5)` must not imply `x > 10`, guarding against an over-strong (unsound) assume model.

The pre-existing `regression/esbmc-unix/builtin_clzll_0{1,2,3}` (including the `clzll(0)` UB case) still pass, confirming the `clzll` generalisation preserves behaviour.


src/goto-symex/builtin_functions/run_builtin.cpp
```diff
@@ -105,51 +105,52 @@ bool goto_symext::run_builtin(
...
 
-  if (has_prefix(symname, "c:@F@__builtin_clzll"))
+  // __builtin_clz / __builtin_clzl / __builtin_clzll: count leading zero bits.
+  // One handler covers all widths — the operand type fixes the bit width. The
+  // result is undefined for a zero argument (matching GCC), reported as UB.
+  // Match the three names exactly: a loose "__builtin_clz" prefix would also
+  // capture __builtin_clzs (16-bit) and the two-argument __builtin_clzg, the
+  // latter tripping the one-argument assertion below. See #4606.
+  if (
+    symname == "c:@F@__builtin_clz" || symname == "c:@F@__builtin_clzl" ||
```
mikhailramalho, line 114: `if (has_prefix(symname, "c:@F@__builtin_clz"))` ?

lucasccordeiro: Good point on the matching. Rather than tweak the prefix here, I moved `__builtin_clz/clzl/clzll` handling out of `run_builtin.cpp` entirely and into `clang_c_adjust_expr.cpp` next to `__builtin_parity`, lowering it to `width - popcount(smear_right(x))`. As a frontend irep the simplifier folds constant arguments (no nondet/symex assumptions), and the exact three-name match there keeps the two-argument `__builtin_clzg` out. So this `run_builtin.cpp` branch is gone in the latest push (b5fe3eb).

One consequence worth flagging: the frontend value lowering can't emit the old symex-level "clz(0) is UB" claim, so `clz(0)` is now modelled as the bit width (defined). I updated `builtin_clzll_03` to match. Happy to restore the UB check (it needs the symex layer) if you'd prefer to keep flagging clz(0).

mikhailramalho (approved):
lgtm

src/goto-symex/builtin_functions/run_builtin.cpp
```diff
@@ -105,57 +105,9 @@ bool goto_symext::run_builtin(
...
-    expr2tc above_zero = equality2tc(shift2, zero);
-    assume(above_zero);
-
-    if (!is_nil_expr(ret))
-      symex_assign(code_assign2tc(ret, typecast2tc(ret->type, clz_sym)));
-
-    return true;
-  }
+  // __builtin_clz / __builtin_clzl / __builtin_clzll are lowered to a
+  // popcount-based expression in the frontend (clang_c_adjust_expr.cpp), so
```
XLiZHI: Shall we remove this comment?

XLiZHI (changes requested):

src/goto-symex/builtin_functions/run_builtin.cpp
mikhailramalho: That's annoying, can our goto-check UB level catch it? Maybe we could skip the rewrite if the argument is zero + UB goto-check adds asserts when enabled?

src/goto-symex/builtin_functions/run_builtin.cpp
mikhailramalho: I agree

src/goto-symex/builtin_functions/run_builtin.cpp
mikhailramalho: Actually, let's always convert clz(0) and add the check for clz UB in goto-check. Feel free to submit another PR if you think it's out of scope for this one.

mikhailramalho: the problem is that if we rewrite the clz in the frontend, goto-check can't see the original clz call :/

src/goto-symex/builtin_functions/run_builtin.cpp
lucasccordeiro: Done — removed the stale comment in 95cab09.

src/goto-symex/builtin_functions/run_builtin.cpp
lucasccordeiro: Agreed it's worth keeping. After looking into it: once clz is fully lowered to a value in the frontend there's no `clz` node left for the goto UB-check to recognise, and the frontend layer can't emit the assert itself — so the UB claim really needs the symex layer. So I kept `__builtin_clz` in `run_builtin.cpp` (latest push, 860439f) rather than moving it to the frontend: the popcount-based encoding still folds a constant argument to a constant (0 remaining VCCs on the test), and `claim(arg != 0)` flags clz(0) as UB. Matched the three names exactly so the two-argument `__builtin_clzg` isn't captured. `builtin_clzll_03`'s UB expectation is restored.

mikhailramalho: can you move the UB check to be added with the other ub checks we do in goto-check? By default, we don't check for UB, and when we do, they are all inserted at the goto level. The current design of this PR deviates from that.

lucasccordeiro: Done in 0937518 — moved the clz(0) UB check to goto-check, alongside the other UB checks:

- Removed the unconditional `claim(arg != 0)` from `run_builtin` (the clz value lowering, which folds constant args, stays there).
- Added a new `--clz-zero-check` option (off by default, like the other UB checks). When set, `goto_check` inserts `ASSERT(arg != 0)` before each `__builtin_clz/clzl/clzll` call (it's still a FUNCTION_CALL in the GOTO, so goto-check sees it).
- By default `clz(0)` is the defined value (bit width); with `--clz-zero-check` it's flagged as UB (verified on both a constant zero and a symbolic argument that could be zero).

`builtin_clzll_03` now opts into `--clz-zero-check`; `builtin_clz_zero_default` covers the unchecked default.

mikhailramalho (approved):
Lgtm

Issue #5075, opened 2026-06-03
[symex] Rewrite exception support to be fully symbolic


lucasccordeiro:
## Summary

Follow-up from review of #5070. As @mikhailramalho noted:

> At some point, we need to rewrite the exception support to be fully symbolic

This issue tracks that rewrite.

## Background

Exception dispatch currently lives in `goto_symext` (`src/goto-symex/symex_catch.cpp`)
and is resolved **structurally at symbolic-execution time** rather than encoded
into the SMT formula:

- `symex_catch()` maintains a `stack_catch` of active `try` regions, each holding
  a `catch_map`/`catch_order` that maps a catch type to a handler target.
- On a `THROW`, `update_throw_target` walks that map and **picks a single handler
  by matching type names as strings** (`is_python_exception_subtype`, exact
  string compare for C++), then redirects control to one concrete target.

Because the match is performed imperatively over a stack of targets instead of
being expressed as guards in the formula, the model is fragile:

- It dereferenced a nil `code` field on catch-all handlers whose first
  instruction is an `ASSERT`/`ASSUME`/`GOTO` — the segfault fixed in #5070.
- Type matching is done by string comparison rather than over the symbolic
  type/value of the thrown object.
- Handler selection commits to one target rather than letting the solver
  explore the dispatch as part of the path condition, which makes it hard to
  reason soundly about throws on conditional/symbolic paths, rethrows,
  unwinding interactions, and cross-frontend (C++ vs Python) semantics.

## Proposed direction

Rewrite exception support so that throw/catch dispatch is **fully symbolic**:
encode "which handler catches this throw" as guarded control flow in the SSA
/ SMT encoding (driven by the symbolic type and value of the thrown object),
rather than resolving it imperatively against `stack_catch` during symex.

This is a design-level change; the goal of this issue is to scope and track it,
not to prescribe an implementation. Points to work out:

- A symbolic representation of the in-flight exception (type id + value).
- Guard-based handler selection (including catch-all / ellipsis and
  base-class / subclass matching) expressed in the formula.
- Correct interaction with loop unwinding, function-call inlining, rethrow,
  and stack unwinding of destructors (C++).
- Shared semantics across the C++ and Python frontends.

## References

- Review comment: https://github.com/esbmc/esbmc/pull/5070#issuecomment-4612017824
- Most recent band-aid: #5070 (segfault on assertion in a catch-all handler)
- Code: `src/goto-symex/symex_catch.cpp`

PR #5108, opened 2026-06-04
[symex] Symbolic exception lowering (--lower-exceptions), opt-in

lucasccordeiro:
## Summary

Follow-up to #5075. Adds an experimental, opt-in lowering of C++/Python
`throw`/`catch` into **guarded control flow over a global exception state**,
performed as a goto-functions transform *before* symbolic execution. Handler
dispatch then becomes a guard the SMT solver resolves over the thrown object's
symbolic type-id, rather than the imperative string-matching currently done in
`goto-symex/symex_catch.cpp`.

Gated on `--lower-exceptions`, **default OFF** — the imperative path is
unchanged and remains the default. This lands the foundation incrementally;
it does not yet replace the imperative dispatch.

## What's included

- **`exception_typeid`** — a closed-world type-id registry + reflexive-transitive
  subtype closure, built from `bases` metadata and from THROW `exception_list`
  chains (so it serves both the C++ and Python frontends).
- **`exception_globals`** — the in-flight exception state
  (`thrown` / `typeid` / `value`).
- **`remove_exceptions`** — the lowering pass (whole-program, all-or-nothing:
  anything outside the supported subset falls back to the imperative path).
- **OM fix** (`src/cpp/library/exception`) — `std::set_terminate` /
  `std::set_unexpected` now actually store and dispatch the installed handler
  (previously stubbed → handlers were silently ignored on all paths).

## Supported subset (lowered today)

C++ class and primitive throws; reference / value (with single-inheritance
slicing) / pointer / `void*` catches and catch-all; nested and inter-procedural
propagation; rethrow; uncaught detection at `main` and `__ESBMC_main` (the latter
covers exceptions from global constructors during static initialisation). Python
`try`/`except`/`raise` lowers through the same machinery.

Not yet lowered (these fall back to the imperative path): parts of the `std`
exception surface; dynamic exception specifications / `noexcept` *violation*
routing; destructor unwinding; `bad_cast` from `dynamic_cast<T&>`.

## Results

Differential `--lower-exceptions` ON vs OFF, **0 divergences**:
- **39 / 67** CORE `regression/esbmc-cpp/try_catch` tests lower.
- **26 / 29** `regression/python/exception*` / `try_finally*` / `try_else*` lower.

The lowered path is also strictly more correct than the imperative dispatch in
three cases it now fixes: nested inner→outer propagation, function-try-block /
static-init exceptions, and the silently-ignored `set_terminate` handler.

## Verification

- Dedicated `regression/esbmc-cpp/try_catch/lower-exceptions_*` (13) and
  `regression/python/lower_exceptions_raise` tests, plus a type-id unit test
  (`unit/goto-programs/exception_typeid.test.cpp`).
- The OM `terminate`/`unexpected` handler-dispatch branches were proven
  **reachable** (C-Live) under both Bitwuzla and Z3; ships with
  `set_terminate_handler` / `set_terminate_default_fail` contract tests.

See `docs/design-symbolic-exceptions.md` for the architecture and the roadmap to
making this the default (and eventually removing the imperative path).

Refs #5075


src/cpp/library/exception
```diff
@@ -49,27 +49,45 @@ inline bool uncaught_exception() _ESBMC_NOEXCEPT
 }
 
 typedef void (*terminate_handler)();
+typedef void (*unexpected_handler)();
+
+// Currently installed handlers (null == library default). Plain globals — the
+// name avoids the __ESBMC_ prefix so cross-function tracking works.
+terminate_handler esbmc_cur_terminate_handler = 0;
+unexpected_handler esbmc_cur_unexpected_handler = 0;
```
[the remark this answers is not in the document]
lucasccordeiro: Good catch — fixed in 920aa368a8: both handler globals are now `static` (internal linkage), matching `<new>`'s `current_handler`. Verified terminate/unexpected dispatch and the two-TU include still work.

src/goto-programs/remove_exceptions.h
```diff
@@ -0,0 +1,25 @@
...
+///
+/// PoC scope: intra-procedural, single-level try regions, reference catches,
+/// no destructor unwinding. A `throw` is rewritten to arm the
+/// __ESBMC_exc_{thrown,typeid,value} globals (see exception_globals.h) and
+/// branch, in catch-declaration order, to the first handler whose static type
+/// is a supertype of the thrown dynamic type — the match expressed as a guard
+/// `typeid in concrete_subtype_ids(C)` (see exception_typeidt) over a symbolic
+/// type-id, with `catch (...)` as an unconditional branch. Regions outside the
+/// PoC subset (nested, value-catch, throws crossing calls) are left untouched
+/// for the existing symex path to handle.
```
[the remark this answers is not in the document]
lucasccordeiro: Fixed in 272a7c16fa: rewrote the banner to the real contract — class+primitive throws; reference/value/pointer/`void*` catches; nested + inter-procedural propagation; rethrow; Python — and the genuine remaining fallbacks (also dropped the stale `__ESBMC_exc_*` names).

docs/design-symbolic-exceptions.md
```diff
@@ -0,0 +1,109 @@
...
+/ `unexpected()` dispatch to it (previously stubbed — handlers were silently
+ignored on all paths).
+
+## Roadmap to default-on
+
+The imperative path can only be removed once the lowered path reaches parity.
+Remaining work, roughly ordered: primitive throws + `std` interop; pointer/`void*`
+catches; Python handler shapes; destructor unwinding; `bad_cast`; throw-spec /
+`noexcept`. Then: two green full-suite differential runs (`--lower-exceptions`
+ON vs OFF) before flipping the default and deleting `symex_catch.cpp`.
```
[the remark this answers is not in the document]
lucasccordeiro: Fixed in 272a7c16fa: dropped the already-implemented items (primitives, pointer/`void*`, Python) from the roadmap, leaving only the true remaining work (throw-spec/noexcept, std surface, destructors, bad_cast).

XLiZHI (commented):
This is a nice implementation! I examined the the goto functions and I reckon this is a suitable solution.

However, I have some concerns regarding the performance/scalability aspects:

1. The exception_typeidt constructor scans all type symbols in the entire symbol table, not limited to exception types. In large C++ programs, hundreds of types may be registered, most of which are unrelated to exceptions. A suitable solution might be to cache and only register the types that appear in the THROW exception_list or CATCH exception_list, rather than all the type symbols.
2. The "run()" function performs four independent full scans for all GOTO instructions: `compute_may_throw, register_chain, program_supported` and `collect()`. In particular, the functions collect() and program_supported access exactly the same set of instructions, and the tasks they perform are highly overlapping.

src/cpp/library/exception
```diff
@@ -49,27 +49,48 @@ inline bool uncaught_exception() _ESBMC_NOEXCEPT
 }
```
mikhailramalho, line 50: Do we even need this handler? The standard library already has the same global pointer and we could get the set_terminate/get_terminate (and the unexpected variations) implementations for free. On my system:
```
  std::terminate_handler __gnu_cxx::__terminate_handler =
      __gnu_cxx::__verbose_terminate_handler;

  std::terminate_handler
  std::set_terminate(std::terminate_handler func) throw()
  {
    return __atomic_exchange_n(&__gnu_cxx::__terminate_handler,
                               func, __ATOMIC_ACQ_REL);
  }

  std::terminate_handler
  std::get_terminate() noexcept
  {
    return __atomic_load_n(&__gnu_cxx::__terminate_handler, __ATOMIC_ACQUIRE);
  }

File: src/goto-programs/exception_globals.h [hunk 5]
```diff
@@ -0,0 +1,34 @@
...
+///   $esbmc_exc_value   : void*   — pointer to the most-derived thrown
+///                                   object, so handlers can bind/slice it.
+///
+/// The typeid is a size_t to line up with the Python frontend's existing
+/// PyObject.type_id field and with exception_typeidt's unsigned ids; the
+/// value is a void* the handler static_casts to the caught type.
+namespace exception_globals
+{
+// NB: avoid the __ESBMC_ prefix — that is reserved for intrinsics and triggers
+// special symex handling that breaks ordinary cross-function global tracking.
```
mikhailramalho: I would expect a __ESBMC_ prefix here; we can silently ignore them at symex if needed.

mikhailramalho:
Codex found two soundness regressions in the opt-in lowered path, both confirmed by differential tests against the default exception path:

1. **Exception state is shared across threads.** `exception_globals` creates one global `thrown/typeid/value` tuple, and `wire_throw`/`wire_call` read and write it without thread isolation. C++ exception propagation is thread-local, so one thread can observe, catch, or clear another thread’s in-flight exception. Repro: thread 1 throws/catches `E`; thread 2 calls a function that cannot throw inside `try { ... } catch (E&) { assert(false); }`. Default path with `--context-bound 2`: `VERIFICATION SUCCESSFUL`; adding `--lower-exceptions`: `VERIFICATION FAILED` at `assert(false)`. The pass should reject concurrent programs until exception state is modeled per thread.

2. **`register_chain()` creates false subtype relationships for multiple inheritance.** The C++ frontend emits a flattened throw list such as `THROW D, A, B` for `struct D : A, B`. `exception_typeidt::register_chain()` treats every later item as a base of every earlier item, consequently recording `A <: B`. After any syntactic `throw D()`, a later `throw A()` can therefore incorrectly match `catch(B&)`. Repro: include an otherwise-unused function containing `throw D()`, then in `main` throw `A()` with `catch(B&) { assert(false); } catch(A&) {}`. Default path: `VERIFICATION SUCCESSFUL`; `--lower-exceptions`: `VERIFICATION FAILED` at `assert(false)`. The flattened C++ throw list must not be interpreted as a linear ancestry chain; the symbol-table hierarchy already represents `D -> {A, B}` correctly.

These should block enabling or merging the lowered path even while it remains experimental, because they produce incorrect verification results.

src/cpp/library/exception
lucasccordeiro: The operational model is standalone — it replaces libstdc++, so there's no real `__gnu_cxx::__terminate_handler` to reuse; the model has to hold its own state. Made it `static` to match `<new>`'s `current_handler`. Happy to model the libstdc++ symbol directly instead if you'd prefer.

src/goto-programs/exception_globals.h
lucasccordeiro: Done in b19b76c206 — renamed to `__ESBMC_exc_*`. The earlier cross-function-tracking breakage turned out to be the entry-function id (`__ESBMC_main`), not the prefix, so no symex change was needed.

lucasccordeiro:
@mikhailramalho thanks — both confirmed and fixed in b19b76c206, with regression tests:

1. **MI false subtypes.** `register_chain` no longer reads the flattened throw list `[D, A, B]` as a linear ancestry — only the dynamic type (front) is a subtype of the rest, so `A <: B` is never invented. Each dynamic type still registers its own ancestry from its own throw, so matching stays complete. Test: `lower-exceptions_multiple_inheritance`.
2. **Thread-shared state.** The pass now rejects programs that call `pthread_create` (falls back to the imperative path) until the exception state is modelled per thread. Test: `lower-exceptions_concurrent`.

Re-ran the ON-vs-OFF differential on CORE `try_catch`: 0 divergences.

lucasccordeiro:
@XLiZHI thanks for the perf review:

1. The `exception_typeidt` constructor scan is O(N) once at GOTO build and only reads each type's `bases` field; queries touch just the relevant subgraph, so the cost is negligible next to symex. Restricting registration to exception types is a fair future optimization.
2. Folded the concurrency check into the `register_chain` scan (one fewer pass). `program_supported` (whole-program gate) and `collect` (per-function, during lowering) run at different granularities, so I kept them separate for now, but consolidating is a reasonable cleanup to revisit.

src/goto-programs/remove_exceptions.cpp
```diff
@@ -0,0 +1,591 @@
...
+    }
+
+    body.update();
+  }
+
+  goto_programt::targett make_epilogue(goto_programt &body)
+  {
+    auto end = body.instructions.end();
+    auto last = end;
+    for (auto it = body.instructions.begin(); it != end; ++it)
```
XLiZHI: Use `auto last = std::prev(body.instructions.end())` might avoid scanning the entire function body.

src/goto-programs/remove_exceptions.cpp
```diff
@@ -0,0 +1,591 @@
...
+      for (const handlert &h : r.handlers)
+      {
+        auto g = add();
+        if (h.type == ellipsis_id)
+        {
+          g->make_goto(h.landing);
+          catch_all = true;
+          break;
+        }
+        g->make_goto(h.landing, and2tc(thrown, match_guard(h.type)));
```
XLiZHI: 1. wire_call inserts IF thrown GOTO dispatch - jumps only if the condition is true,` thrown == true`
2. wire_throw writes thrown = true, then unconditional GOTO dispatch, `thrown == true`
3. The inner dispatch propagates to the outer unconditional GOTO, `thrown == true`

Use `g->make_goto(h.landing, match_guard(h.type));`?

XLiZHI (changes requested):

src/goto-programs/remove_exceptions.cpp
lucasccordeiro: Good observation — `thrown` is always true at a dispatch block (reached only via an armed throw, an `IF thrown` post-call jump, or an uncleared propagate), so the conjunct is redundant. Dropped it in a407dcec7d; differential unchanged (0 divergences).

src/goto-programs/remove_exceptions.cpp
lucasccordeiro: Done in a407dcec7d — END_FUNCTION is always the last instruction, so make_epilogue now uses `std::prev(body.instructions.end())` instead of scanning.

src/cpp/library/exception
mikhailramalho: If we just replicate the standard libraries behavior there is no need for a model. Let's no reinvent the wheel

mikhailramalho:
Do we still need the symex code if we enable this code by default?

src/cpp/library/exception
lucasccordeiro: Done in 2d4509e59c — rewrote the model the libstdc++ way: `__gnu_cxx::__terminate_handler` / `__unexpected_handler` initialised to real default handlers, with `set_`/`get_`/`terminate` operating on them. The default is non-null so `terminate()` is just `(*handler)()` (no special-case branch), and `get_terminate`/`get_unexpected` come for free. Used plain access instead of `__atomic_*` since BMC is single-threaded and the lowering rejects concurrent programs. All 101 try_catch tests + the set_terminate contract tests pass.

lucasccordeiro:
@mikhailramalho not yet — the goal is to remove `symex_catch.cpp` once the lowered path reaches full parity, but it isn't there: it still falls back to the imperative path for parts of the `std` exception surface, throw-spec/noexcept violations, destructor unwinding, `bad_cast`, and concurrent programs. While the flag is opt-in (default OFF) both paths coexist; the imperative path can only be deleted after a green full-suite ON-vs-OFF differential across all suites. I've kept the remaining gaps tracked in `docs/design-symbolic-exceptions.md`.

src/cpp/library/exception
```diff
@@ -49,28 +49,82 @@ inline bool uncaught_exception() _ESBMC_NOEXCEPT
...
+// variants fall out directly. libstdc++ uses __atomic_{exchange,load}_n for
+// thread safety; a plain access is observationally identical here (BMC is
+// single-threaded and the lowering rejects concurrent programs).
+namespace __gnu_cxx
+{
+// Default terminate handler (models __verbose_terminate_handler): aborts.
+inline void __verbose_terminate_handler()
+{
+  __ESBMC_assert(0, "terminate called after throwing an exception");
+  while (1)
```
mikhailramalho, line 71: why  do we need an infinite loop here?

src/cpp/library/exception
mikhailramalho: Renaming the model variables to the libstdc++ names while keeping them static in the header still reimplements the mechanism and changes its semantics. __gnu_cxx::__terminate_handler and __unexpected_handler must represent one program-wide state. With static, every translation unit gets a separate handler, so set_terminate() in one TU is not observed by terminate() in another.

Also, the claim that plain accesses are sufficient because exception lowering rejects concurrent programs does not hold: this <exception> change affects the default path and programs that do not use --lower-exceptions. Please keep declarations in the header and provide one shared operational-model definition/implementation, matching the standard-library structure and atomic behavior.

mikhailramalho:
Codex found three additional soundness issues in the opt-in lowered path:

1. **Nested handled exceptions break bare rethrow.** The handler landing clears only `__ESBMC_exc_thrown`; `typeid` and `value` remain in the single global tuple. If an outer handler catches `A`, then throws and catches `B`, a later `throw;` rethrows `B` instead of the currently handled `A`:

```cpp
try { throw A{}; }
catch (A &) {
  try { throw B{}; } catch (B &) {}
  throw; // must rethrow A; lowered path rethrows B
}
```

The lowering needs a stack of currently handled exceptions or must reject this pattern.

2. **Multiple-inheritance catch binding lacks base-subobject adjustment.** After dispatch matches a derived object against a base handler, `make_landing()` casts the most-derived object's `void *` directly to the caught base pointer. For `struct D : A, B`, binding `catch (B &b)` requires adjustment to the `B` subobject; the current cast points at the start of `D`, producing incorrect field reads. The added multiple-inheritance regression checks dispatch only, not handler binding.

3. **The concurrency fallback is bypassed by indirect `pthread_create` calls.** The pass rejects only direct symbol calls whose name contains `pthread_create`. Calling `pthread_create` through a function pointer leaves lowering enabled despite the shared exception globals. Indirect calls are otherwise explicitly supported by this pass, so concurrency detection must conservatively handle this case or reject indirect calls.

mikhailramalho:
> @mikhailramalho not yet — the goal is to remove `symex_catch.cpp` once the lowered path reaches full parity, but it isn't there: it still falls back to the imperative path for parts of the `std` exception surface, throw-spec/noexcept violations, destructor unwinding, `bad_cast`, and concurrent programs. While the flag is opt-in (default OFF) both paths coexist; the imperative path can only be deleted after a green full-suite ON-vs-OFF differential across all suites. I've kept the remaining gaps tracked in `docs/design-symbolic-exceptions.md`.

Understood. Yeah, I think this will be a better approach compared to the current symex.

mikhailramalho:
Summary of the `terminate` / `unexpected` modelling discussion:

The standard-library model should treat `std::terminate` and `std::unexpected` as process-wide installed handlers, not as per-header or per-translation-unit state. The default terminate handler should not be represented as an infinite loop: for ESBMC it is better modelled as a reported failure followed by path termination, e.g. `__ESBMC_assert(0, "terminate called ..."); __ESBMC_assume(0);`. A bare `assume(0)` would silently discard the path and could turn a real terminate reachability bug into `VERIFICATION SUCCESSFUL`; a `while(1)` risks killing verification progress.

`std::terminate()` is `noexcept` and calls the current terminate handler. A user-installed terminate handler has type `void (*)()`, so the type system does not prevent it from throwing, but the standard requires a terminate handler to terminate execution without returning. A useful ESBMC model should therefore catch both invalid behaviours:

```cpp
inline void terminate() _ESBMC_NOEXCEPT
{
  try
  {
    (*std::get_terminate())();
  }
  catch (...)
  {
    __ESBMC_assert(0, "terminate handler threw an exception");
    __ESBMC_assume(0);
  }

  __ESBMC_assert(0, "terminate handler returned");
  __ESBMC_assume(0);
}
```

`std::unexpected()` is different. Reaching it is not necessarily unrecoverable for pre-C++17 dynamic exception specifications. For `noexcept`, an escaping exception goes directly to `terminate()` and cannot recover through `unexpected`. But for legacy `throw(T...)`, `unexpected()` may recover if the installed unexpected handler throws a replacement exception that satisfies the violated dynamic exception specification. There is also the legacy `std::bad_exception` rule: if the unexpected handler throws a disallowed exception and the violated dynamic specification includes `std::bad_exception`, the implementation may replace it with a `bad_exception` object and continue handler search.

A faithful pre-C++17 recovery model would be:

1. When handler search exits a function with a violated dynamic exception specification, record the violated function's allowed exception list on the current exception/spec frame.
2. Call the current unexpected handler.
3. If the unexpected handler returns, call `std::terminate()`.
4. If it throws/rethrows an exception whose dynamic type is allowed by the violated specification, resume normal handler search at the call of the function whose specification was violated, using that replacement exception.
5. If it throws/rethrows a disallowed exception and `std::bad_exception` is allowed by the violated specification, replace the thrown object/type with `std::bad_exception` and resume handler search at that call.
6. Otherwise call `std::terminate()`.

So a minimal current model can conservatively route unexpected-handler throws to `terminate()`, but that is incomplete: it will miss valid recovery paths in C++98/03/11/14 programs using dynamic exception specifications. If we want parity with the legacy semantics, the exception-spec work should make `unexpected` depend on the violated function's frame/specification rather than model it as a standalone library call.

mikhailramalho (approved):
I left two comments, but I now think that the terminate/unexpected changes should go in a separate PR, likely after I land the rework in #5126 (since they are somewhat related).

I'm also happy with deleting the broken symex code and marking the failing tests as known bugs until the goto lowering pass catches up.

src/cpp/library/exception
lucasccordeiro: You're right — addressed in #5151. The handler globals now have external linkage (not `static`), so they're one program-wide object like the real `__gnu_cxx::__terminate_handler`; ESBMC merges the same-named model global across TUs (no multiple-definition issue). New regression `set_terminate_cross_tu` installs the handler in one TU and calls `std::terminate()` in another.

src/cpp/library/exception
lucasccordeiro: Addressed in #5151 — `terminate()` is no-return, so after the assert control must not fall back to the caller. Replaced the `while (1) {}` with `__ESBMC_assume(0)`, which states the unreachability directly instead of an opaque spin.

PR #5126, opened 2026-06-05
[c++] Rework C++ exception handling: GOTO lowering as the sole path, function-level exception specs, and an operational model

### LINKED ISSUE #5125: [C++] Replace THROW_DECL instructions with function-level exception specifications
## Problem

ESBMC currently represents C++ function exception specifications as executable GOTO instructions, `THROW_DECL` and `THROW_DECL_END`. Exception specifications are function-boundary contracts, not executable body regions, and the current representation causes their enforcement to depend on unrelated control-flow state.

The Clang C++ frontend inserts a `throw_decl` node at the beginning of a function body. GOTO conversion turns it into `THROW_DECL`, and `goto_symext::symex_throw_decl()` copies the specification into `stack_catch.top()`. However, `stack_catch.top()` represents the nearest active `try`/`catch` region, not the current function frame.

This creates several correctness and maintenance problems:

- If no catch region is active, `symex_throw_decl()` does nothing, so the function specification can be silently ignored.
- The specification is attached to the caller/callee catch stack rather than to the function whose boundary it constrains.
- At `END_FUNCTION`, ESBMC clears the specification from whichever catch region happens to be on top.
- `THROW_DECL_END` remains part of the GOTO instruction set even though the current frontend no longer emits it.
- Inlining copies the marker into the caller body, losing the semantic function boundary.
- The instruction-region representation complicates recursion, indirect calls, function pointers, threads, serialization, and future exception lowering.

The C++ semantic rule is boundary-based: a non-throwing exception specification is violated when handler search exits the function body. Older dynamic exception specifications similarly invoke `std::unexpected()` when handler search reaches the function's outermost block with an exception the specification does not allow. A function may throw such an exception and catch it internally without violating its specification.

References:

- Current C++ `[except.handle]`: https://eel.is/c++draft/except.handle
- C++14 dynamic exception specifications `[except.spec]`: https://timsong-cpp.github.io/cppwp/n3337/except.spec

## Proposed design

Remove `THROW_DECL` and `THROW_DECL_END` and introduce a structured exception-specification representation, for example:

```cpp
struct exception_specificationt
{
  enum class kindt
  {
    potentially_throwing,
    non_throwing,
    dynamic
  };

  kindt kind;
  std::vector<irep_idt> allowed_types;
};
```

Store the canonical specification as function/type metadata so it:

- survives GOTO binary serialization;
- is available for declarations without bodies;
- can represent function-pointer and virtual-call types;
- remains associated with the function boundary rather than body instructions.

`goto_functiont` may cache the specification for convenient executable access, but the canonical representation should remain available from the function symbol/type.

When symex creates a function frame, copy or reference the selected function's specification from the function metadata into `goto_symex_statet::framet`. Exception dispatch should then:

1. Search handlers belonging to the current frame.
2. If none matches, check the frame's exception specification before leaving the frame.
3. Invoke `std::terminate()` for a violated non-throwing specification.
4. Invoke `std::unexpected()` for a violated legacy dynamic exception specification.
5. Otherwise continue handler search in the caller frame.

Ideally, active catch handlers should also be owned by each function frame rather than by the global `stack_catch`. That would make function-boundary traversal explicit and naturally handle recursion and threads.

## Inlining

Moving the specification only into runtime frames is insufficient because an inlined callee no longer creates a frame. The initial safe implementation should prevent inlining functions with restrictive exception specifications. A later explicit exception-control-flow representation could preserve exception boundaries while allowing those functions to be inlined.

## Acceptance criteria

- [ ] Replace `THROW_DECL` and `THROW_DECL_END` with function/type exception-specification metadata.
- [ ] Store or reference the active specification on each symex function frame.
- [ ] Enforce a specification only when exception handler search exits that function.
- [ ] Correctly distinguish non-throwing specifications from legacy dynamic exception specifications.
- [ ] Preserve specifications through GOTO binary serialization.
- [ ] Handle direct calls, indirect calls, recursion, and threads correctly.
- [ ] Prevent unsafe inlining of functions with restrictive exception specifications, or preserve their exception boundaries explicitly.
- [ ] Add regression tests where a restricted function catches a disallowed exception internally.
- [ ] Add regression tests where an exception escapes a `noexcept` function and invokes `terminate`.
- [ ] Add regression tests for violated dynamic exception specifications and `unexpected`.
- [ ] Add regression tests covering recursion, indirect calls, inlining, and calls with no surrounding `try` region.

## Relevant code

- `src/clang-cpp-frontend/clang_cpp_convert.cpp`
- `src/clang-cpp-frontend/clang_cpp_adjust_code.cpp`
- `src/goto-programs/goto_convert.cpp`
- `src/goto-programs/goto_program.h`
- `src/goto-symex/symex_catch.cpp`
- `src/goto-symex/symex_main.cpp`
- `src/goto-symex/goto_symex_state.h`
- `src/goto-programs/goto_inline.cpp`

mikhailramalho:
## Summary

Reworks ESBMC's C++ (and Python) exception handling end to end. What started as
#5125 — replacing the executable `THROW_DECL` markers with function-level
exception-specification metadata — grew to subsume the full exception rework:

- the legacy **imperative symex exception engine is removed**, and **GOTO-level
  lowering (`remove_exceptions`) is now the sole exception path**;
- **exception specifications are function-boundary metadata** enforced at the
  function frame (not the nearest catch region);
- a **C++ exception operational model** backs the standard facilities
  (`std::exception_ptr`, nested exceptions, `uncaught_exception(s)`,
  `terminate`/`unexpected`, `bad_cast`);
- supporting **multiple-inheritance** correctness fixes (subobject layout,
  `dynamic_cast` across non-zero base offsets) come along because catch-by-base
  and `dynamic_cast`-driven `bad_cast` depend on them.

> `master` independently merged "enable symbolic exception lowering by default"
> (#5192) and "lower empty catch-all handlers" (#5241). `master` is merged into
> this branch; this PR **supersedes** that work by removing the
> `--lower-exceptions` / `--no-lower-exceptions` flags entirely — lowering is
> unconditional and the only path.

## 1. Exception specifications (original #5125)

Exception specifications are function-boundary contracts, not executable body
regions. They were modelled as `THROW_DECL`/`THROW_DECL_END` instructions whose
enforcement copied the spec into the nearest active `try`/`catch` region, **not**
the current function frame — so a `noexcept`/`throw()` was silently ignored with
no surrounding `try`, the spec attached to the wrong stack, and inlining lost the
boundary.

- **`exception_specificationt`** (`src/util/`): `kind {potentially_throwing,
  non_throwing, dynamic}` + `allowed_types`, stored as named-sub attributes on
  the function symbol's `typet` so it survives goto-binary serialization and is
  available for bodyless declarations. `noexcept(false)` is correctly treated as
  potentially-throwing.
- **Enforcement**: each `framet` carries its spec; on a throw, every frame the
  exception exits before being caught is checked innermost-first — `terminate`
  for a violated non-throwing spec, `unexpected` (with `std::bad_exception`
  substitution) for a violated dynamic spec.
- Functions with a restrictive spec are not inlined (preserved under
  `--full-inlining`) so the boundary survives.
- `THROW_DECL`/`THROW_DECL_END` and the `code_cpp_throw_decl[_end]` IREP2
  expressions are deleted.

## 2. Lowering is the only exception path

- The imperative symex exception engine and its dead THROW/CATCH dispatch are
  deleted; `remove_exceptions` runs unconditionally; the `--lower-exceptions` /
  `--no-lower-exceptions` flags and the differential tooling/design doc are
  removed.
- A bare `throw` with no in-flight exception lowers to `std::terminate`.
- Per-thread exception state → concurrent exceptions lower correctly.
- `terminate`/`unexpected` points route through the OM; a failing
  `dynamic_cast<T&>` lowers `std::bad_cast` in-line (no fallback); empty trailing
  handlers (`catch(...){}`) are normalized and lowered.
- Computed (function-pointer) pthread start routines are over-approximated
  instead of rejected.

## 3. C++ exception operational model

Header-only C++ exception OM providing:

- `std::exception_ptr`, `current_exception` / `rethrow_exception` /
  `make_exception_ptr`, and nested exceptions (`throw_with_nested` /
  `rethrow_if_nested`), with **pay-per-use** lowering (the per-thread handled
  stack / uncaught counter are only instrumented when the program uses them);
- `std::uncaught_exception(s)` via a lowered per-thread count;
- `vector::at` throwing `std::out_of_range`;
- `std::bad_cast` / `std::bad_typeid` deriving from `std::exception` with a
  virtual `what()`;
- exception-model internals hidden from counterexample traces.

## 4. Supporting C++ correctness fixes

- Multiple-inheritance base subobjects are laid out at their real (clang)
  offsets; derived↔base pointer conversions, base ctor/dtor calls, and
  catch-by-base are adjusted accordingly.
- `dynamic_cast` works across non-zero MI base offsets.
- **Known limitations** (reclassified `KNOWNBUG`, tracked for a follow-up): the
  MI offset model does not yet handle polymorphic delete-through-base,
  virtual/diamond inheritance layout, or same-named members in distinct bases;
  `throw_with_nested` fidelity is blocked on the per-class vtable model.

## Tests

New `regression/esbmc-cpp/try_catch/` coverage: `exception_spec_*` (internal
catch, `noexcept` escape → terminate, dynamic violation → unexpected,
`noexcept(false)`, recursion, indirect call, `--full-inlining`, throw with no
`try`); `lower-exceptions_*` (uncaught count, exception_ptr, make_exception_ptr,
nested rethrow, empty catch-all, concurrent, computed pthread routine,
`bad_cast`, `vector::at`); MI subobject layout/catch/`dynamic_cast`; and Python
`raise` cases. Several previously-`KNOWNBUG` tests now detected correctly are
reclassified to `CORE`; genuinely-unsupported MI cases are marked `KNOWNBUG`.

## Validation

`master` merged in and the reconciliation validated: `try_catch` 153/153,
cpp + inheritance + polymorphism 681/681, core-C `esbmc/` 1312/1312, clean build.

Fixes #5125.


lucasccordeiro (approved):

regression/esbmc-cpp/inheritance/llbmc_multiple_inheritance/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: This expects SUCCESSFUL and is downgraded while `-wrong` (expects FAILED) is upgraded — the correct program now false-alarms. Fix the layout change, not the test mode.

regression/esbmc-cpp/inheritance/llbmc_virtual_inheritance/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: Same inversion for virtual inheritance: the correct program is disabled while its `-wrong` twin starts passing. Both must hold.

regression/esbmc-cpp/inheritance/inheritance11/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: a passing test is disabled by this PR. Investigate the new false alarm instead of downgrading.

lucasccordeiro (changes requested):
This PR also re-implements the multiple-inheritance subobject layout and `mi_*` tests already under review in #5552; split the MI work out so the exception rework can be reviewed and landed independently.

src/c2goto/library/exception.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+struct __ESBMC_handled_exception
+{
+  __SIZE_TYPE__ typeid_value;
+  void *object;
+};
+
+static __ESBMC_exception_slot __ESBMC_exception_slots[256];
+static __SIZE_TYPE__ __ESBMC_exception_slot_count = 0;
+static __ESBMC_handled_exception __ESBMC_handled_exceptions[256];
```
lucasccordeiro, line 43: These four slot/stack statics are process-global while the exception state is `__thread`, so concurrent threads corrupt each other's handled stack. Mark them `__thread` and add a threaded `current_exception` test.

src/goto-programs/remove_exceptions.cpp
```diff
@@ -768,16 +1050,52 @@ class exception_loweringt
...
+  void
+  build_handled_stack(goto_programt &body, const std::vector<regiont> &regions)
+  {
+    expr2tc pop = make_c_helper_call(exception_globals::pop_handled_id);
+    if (is_nil_expr(pop))
+      return;
+
+    for (const regiont &r : regions)
+    {
+      auto p = body.insert(r.after_try);
```
lucasccordeiro, line 1092: `after_try` is also the normal-completion target, so this pop runs without a matching `push_handled`: a nested try completing normally inside a handler discards the enclosing entry. Guard the pop.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2429,8 +2456,24 @@ void clang_cpp_convertert::gen_typecast_base_ctor_call(
...
+    !get_non_virtual_base_offset(
+      *derived_ctor->getParent(), base_record, base_offset))
+  {
+    gen_typecast(ns, implicit_this_symb, base_ctor_this_type);
+  }
+  else if (
+    adjust_base_subobject_pointer(
+      implicit_this_symb, base_ctor_this_type, base_offset, false))
+  {
+    abort();
```
lucasccordeiro, line 2474: `abort()` kills ESBMC with no diagnostic. Make `gen_typecast_base_ctor_call` return an error and propagate it like every other conversion failure.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -1857,47 +1881,49 @@ bool clang_cpp_convertert::get_function_body(
...
+    // Legacy dynamic spec throw(T...) (incl. empty throw()): violation calls
+    // std::unexpected.
+    type.set(exception_specificationt::kind_attribute(), "dynamic");
+    irept &decl = type.add("exception_spec_decl");
+    decl.get_sub().clear();
+    for (unsigned i = 0; i < fpt->getNumExceptions(); i++)
+    {
+      typet exc_type;
+      if (get_type(fpt->getExceptionType(i), exc_type))
+        continue;
```
lucasccordeiro, line 1923: Silently dropping an unconvertible type narrows the dynamic spec, so a permitted escape reports a false "exception specification violated". Return an error, or fall back to potentially-throwing.

src/util/exception_specification.cpp
```diff
@@ -0,0 +1,66 @@
+#include <util/exception_specification.h>
+#include <util/type.h>
+
+bool exception_specificationt::allows(const irep_idt &exception_type_id) const
```
lucasccordeiro, line 4: `allows()` has no caller in src/ or unit/, and its exact name comparison would miss subtypes if one appeared. Remove it, or implement it over the subtype relation.

src/goto-symex/symex_main.cpp
```diff
@@ -478,42 +431,11 @@ void goto_symext::symex_step(reachability_treet &art)
...
-
-      except->has_throw_decl = false;
-      except->throw_list_set.clear();
-    }
-
-    cur_state->source.pc++;
-    break;
-
   default:
+    // THROW/CATCH land here: remove_exceptions lowers every throw/catch to
```
lucasccordeiro, line 435: A declined program only gets a `log_warning` from `report_unsupported`, then dies here on `abort()` printing a numeric opcode. Reject it in `remove_exceptions` with a real error.

regression/esbmc-cpp/try_catch/lower-exceptions_rethrow_no_active_fail/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
+main.cpp
+--unwind 4
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 5: `^VERIFICATION FAILED$` alone passes on any failing property. Add the `throw with no active exception` line; same for the other new `_fail` descs with a bare FAILED regex.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -49,6 +53,130 @@ clang_c_convertert::clang_c_convertert(
...
+
+  bases.clear();
+  bases.reserve(ordered.size());
+  for (const ordered_baset &base : ordered)
+    bases.push_back(base.spec);
+
+  return false;
+}
+
+bool clang_c_convertert::get_non_virtual_base_offset(
```
lucasccordeiro, line 97: `get_non_virtual_base_offset` returns true on success, while `get_ordered_direct_bases` above it returns false on success. Align both with the file's true-means-error convention.

docs/plan-remove-symex-exception-engine.md
```diff
@@ -0,0 +1,456 @@
+# Plan: C++ exception lowering — finish the work the engine removal started
```
lucasccordeiro, line 1: This is a 456-line in-progress work plan with open TODO sections, not durable documentation. Drop it from the PR and track the remaining gaps in an issue.

PR #5127, opened 2026-06-05
[symex] Lower noexcept/throw() specifications in --lower-exceptions

lucasccordeiro:
## Summary

Follow-up to #5108: extends the opt-in `--lower-exceptions` pass to lower
`noexcept` / `throw()` specifications, so a no-throw function from which an
exception escapes is modelled as calling `std::terminate` ([except.spec])
instead of forcing the whole program back onto the imperative path.

Still gated on `--lower-exceptions` (default OFF); the imperative path is the
unchanged default.

## What it does

- **Enforcement at the epilogue.** For a no-throw function, the pass asserts at
  the function epilogue that no exception is in flight (`__ESBMC_exc_thrown ==
  false`), reusing the existing uncaught-at-`main` mechanism. A locally-caught
  throw clears the flag, so the assert fires only on a genuine escape — matching
  the standard (the spec is violated when an exception *escapes*, not when one
  is thrown and caught internally).
- `THROW_DECL` / `THROW_DECL_END` markers are rewritten to `SKIP`. Only dynamic
  specifications with real types (`throw(T...)`, a C++14-only form) still force
  fallback.

## Frontend fix (separate commit)

`hasNoexceptExceptionSpec()` is also true for `noexcept(false)`, which *permits*
exceptions to escape. The frontend emitted the same `noexcept` `THROW_DECL`
marker for it as for `noexcept(true)`, which would make the new enforcement
raise a spurious `terminate` on legal code. The marker is now gated on
`canThrow() == CT_Cannot`, so only genuinely non-throwing specifications are
recorded. (Found in review.)

## Results

- Differential `--lower-exceptions` ON vs OFF across CORE `regression/esbmc-cpp/try_catch`:
  **69 same, 0 divergences**, with **47 of 67** now lowering (up from 42).
- New regression tests: `noexcept_escape_fail`, `noexcept_call_fail` (escape →
  FAILED); `noexcept_local_catch`, `noexcept_false` (no violation → SUCCESSFUL).
- Full `esbmc-cpp/try_catch` suite (104) green; type-id unit test green;
  `clang-format-11` clean. The 3 unrelated broad-suite failures
  (`utility2_fail`, `builtin-template[-fail]`) were confirmed pre-existing by
  building without this change.

Like the earlier nested-propagation and static-init fixes, the lowered path is
standard-correct where it diverges from the imperative path (a throw escaping a
`noexcept` function should terminate; the imperative path currently lets it be
caught by a caller).

Refs #5075

### DIFF
--- a/src/goto-programs/remove_exceptions.cpp
+++ b/src/goto-programs/remove_exceptions.cpp
@@ -13,10 +13,32 @@
 namespace
 {
 const irep_idt ellipsis_id = "ellipsis";
+const irep_idt noexcept_id = "noexcept";
 
 bool is_pointer_catch(const irep_idt &type)
 {
   return type.as_string().find("_ptr") != std::string::npos;
+}
+
+/// A THROW_DECL marking a *no-throw* specification (`noexcept` or `throw()`):
+/// its list is empty or names only the "noexcept" marker. A list naming real
+/// types is a dynamic exception specification (throw(T...), C++14-only) which
+/// the lowering does not model.
+bool is_no_throw_decl(const expr2tc &code)
+{
+  for (const irep_idt &t : to_code_cpp_throw_decl2t(code).exception_list)
+    if (t != noexcept_id)
+      return false;
+  return true;
+}
+
+/// True iff the function body carries a no-throw specification.
+bool function_is_noexcept(const goto_programt &body)
+{
+  for (const auto &ins : body.instructions)
+    if (ins.type == THROW_DECL && is_no_throw_decl(ins.code))
+      return true;
+  return false;
 }
 
 /// One catch clause: its static type (or "ellipsis"), the instruction that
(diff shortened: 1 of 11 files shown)


src/goto-programs/remove_exceptions.cpp
```diff
@@ -397,16 +422,21 @@ class exception_loweringt
...
+    //  - a noexcept function: the exception escapes the no-throw boundary →
+    //    terminate ([except.spec]). A locally-caught throw clears `thrown`, so
+    //    this fires only on a genuine escape.
+    const bool is_entry =
+      id2string(fn_id).rfind("c:@F@main#", 0) == 0 || fn_id == "__ESBMC_main";
+    if (is_entry || noexcept_fn)
     {
       auto a = body.insert(std::next(epilogue));
       a->make_assertion(equality2tc(thrown, gen_false_expr()));
       a->location.property("exception");
```
XLiZHI, line 437:
```c
  a->location = epilogue->location; 
  a->function = epilogue->function;
```

We need to add the location information to this assertion

src/goto-programs/remove_exceptions.cpp
```diff
@@ -373,6 +390,14 @@ class exception_loweringt
     std::vector<regiont> regions;
     std::vector<sitet> throws, calls;
     collect(body, regions, throws, calls);
+
+    // Throw-spec markers are consumed here; once the throws are lowered the
+    // imperative throw-decl machinery has nothing to act on.
+    const bool noexcept_fn = function_is_noexcept(body);
```
XLiZHI, line 396: function_is_noexcept(body)  ← Scan: Search for THROW_DECL
THROW_DECL → SKIP Loop      ← Scan: Check for conversion flag 

Scanning can be combined into one:

```c
  bool noexcept_fn = false;
  for (auto &ins : body.instructions)
  {
    if (ins.type == THROW_DECL)
    {
      if (is_no_throw_decl(ins.code))
        noexcept_fn = true;
      ins.make_skip();
    }
    else if (ins.type == THROW_DECL_END)
      ins.make_skip();
  }
```

src/goto-programs/remove_exceptions.cpp
lucasccordeiro: Addressed in #5149 — the epilogue assertion now copies `location`/`function` from the epilogue, so the counterexample shows the function instead of "no location".

src/goto-programs/remove_exceptions.cpp
lucasccordeiro: Done in #5149 — folded the no-throw detection into the THROW_DECL→SKIP pass (one traversal) and removed `function_is_noexcept`.

PR #5164, opened 2026-06-05
[CI] add concurrency limit to push workflow

XLiZHI:
Cancel in-progress coverage runs when a new push to master arrives, so only the latest coverage is uploaded to Codecov.

same as #4734


lucasccordeiro (approved):
Good catch!

PR #5169, opened 2026-06-05
[goto-symex] collapse merged path guard into a single hidden variable

mikhailramalho:
## What

After each `merge_gotos` join, snapshot the merged path condition into a fresh **hidden SSA guard variable** and continue symex with that single symbol, pushing the accumulated conjunction to the solver. A new helper `collapse_guard()` does the snapshot; it is called once, from `merge_gotos()`, after the guard merge. This is the modern equivalent of the historic "gennaro patch" (18d3fe9).

The symex-side guard stays **O(1)** across merges/unwindings instead of growing and being re-simplified (`guard2tc::operator|=`) on every loop-back, and the encoder no longer re-walks an ever-deepening guard and-chain.

This is **behaviour-preserving** — same verification verdicts; it just moves cost out of ESBMC's (super-linear) encoding into the solver.

## Why / measurements

Benchmarked on the SV-COMP `test_locks_13.c` with deep unwinding (`--unwind N`, k-induction flags), default Bitwuzla, 3-pass averages:

| unwind | total (baseline) | total (this PR) | encode (base→PR) | symex (base→PR) | solve (base→PR) |
|-------:|-----------------:|----------------:|------------------|-----------------|-----------------|
| 40  | 0.44 s | 0.43 s | 0.021→0.017 | 0.12→0.10 | 0.034→0.035 |
| 160 | 1.42 s | 1.33 s | 0.140→0.091 | 0.52→0.48 | 0.151→0.183 |
| 320 | 3.06 s | 2.70 s | 0.418→0.207 | 1.13→0.91 | 0.363→0.443 |
| 640 | 7.14 s | **5.72 s** | 1.49→**0.50** | 2.64→**1.92** | 0.86→1.08 |

- **Encoding** is the big win — up to **~2.9× faster at unwind 640** — because the per-step guard becomes a single boolean symbol instead of a replicated deep and-chain.
- **Symex** also speeds up (~1.4× at 640): merges no longer simplify deep guards.
- **Solving** is ~15–18% slower in the mid-range (the conjunction is now the solver's job) — the intended trade.
- Net **end-to-end ~1.25× at 640**, break-even around unwind 80, widening with depth.
- At unwind 1280 the **baseline segfaults** (stack overflow in the recursive `convert_ast` over the giant guard chain); this PR completes. Both eventually hit the recursion ceiling (~2560 here) — making that traversal iterative is a separate follow-up.

A branch-point collapse variant was also measured and **rejected**: minting a guard var per `if` (26/iteration here) bloated slicing ~3× and solving ~2×, ending up worse than baseline. Only the merge-point collapse is kept.

## Status

Draft — relying on CI to run the full regression suite. Regression tests TBD before un-drafting.


mikhailramalho:
This PR:
```
  correct:           23171
    correct true:    15332
    correct false:    7839
  incorrect:            21
    incorrect true:      0
    incorrect false:    21
  unknown:           13408
  Score:             38167 (max: 61270)
```
Master:
```
  correct:           23454
    correct true:    15393
    correct false:    8061
  incorrect:            20
    incorrect true:      0
    incorrect false:    20
  unknown:           13126
  Score:             38527 (max: 61270)
```
I'll close the PR.

PR #5202, opened 2026-06-06
chore: reorganize and cleanup the docs and deps build process

Yiannis128:
Cleans up ESBMC's build documentation, the docs-site structure, and the dependency-fetching parts of the build. No change to what gets built. The docs are now in a much cleaner state, a lot of duplication has been removed, and a lot of topics were moved to more clean locations. The front-page Download section is now also more visible. See the picture below for the Docs main page, before it was a mess of random topics, now it's clean and beautiful.

## Build guide

- Reorganised `docs/development/building` by **platform** — a Tabs → Steps → Details layout (Ubuntu/Debian, Fedora, macOS, FreeBSD, Windows), so contributors follow one tab end to end.
- Fixed the **Fedora** instructions: the old command omitted `-DDOWNLOAD_DEPENDENCIES`, so it failed looking for `fmt`/`nlohmann-json`/`yaml-cpp`/`immer`. The package list is now minimal (build-verified), and `immer` — which Fedora doesn't package — is downloaded by the flag.
- Removed two stale duplicate pages (`linux-build.md`, `windows-build.md`); the Windows content was folded into the Windows tab.

## Documentation restructure

The docs landing page (`docs/_index.md`) was an 1149-line dump migrated from the old site — it duplicated dedicated pages and served no directory role. Split it into focused pages and made the index a concise directory:

- Non-determinism modelling and the verification algorithms → `theory/`.
- Loop invariants → their own page (they expose constructs, like function contracts).
- Quantifiers → `constructs.md`.
- Multiple-file verification, multi-property/all-witnesses and the SMT-backend table → `usage.md`, which links to `theory/` for the in-depth algorithm explanation.
- Python/Solidity/coverage duplicates dropped in favour of the existing dedicated pages.

The home-page Download section was also reorganised into per-OS entries that link to the build guide.

## Website tooling

- Upgraded the **Hextra** theme module `0.9.5 → 0.12.3`, fixing a `hugo server` render error (`.Site.Author` was removed in recent Hugo), and migrated the deprecated `tabs items=` syntax to per-tab `name=`. The site builds warning-free.

## ast2json no longer installed (refs #3790)

`ast2json` is vendored in the source tree, so the build no longer needs it from pip. Removed the obsolete install steps from the docs, `scripts/build.sh`, the differential CI workflow, and the `CLAUDE.md`/`AGENTS.md` contributor notes.

## Dependency fetching

- **Reliability:** `download_zip_and_extract` didn't check the download status, so a blocked/failed download silently left a 0-byte archive that "extracted" to an empty directory (the cause of a confusing `Z3 headers not found` error). It now fails loudly, removes the partial file, and verifies TLS.
- **Reproducibility:** pinned `yaml-cpp` (was tracking its default branch) and added a SHA-256 hash to the `nlohmann-json` download.

## Notes

- The Linux (Ubuntu/Fedora) build paths are build-verified locally. The macOS/FreeBSD/Windows tabs carry the existing instructions forward into the new layout; they were not compile-tested here and would benefit from a maintainer/CI check.
- Out of scope (follow-ups): routing the prebuilt-binary downloads (LLVM/Z3/CVC5/MathSAT) through FetchContent/CPM with per-file hashes (CPM is already vendored and used by ibex), and commit-SHA pinning of the already-tag-pinned source deps.

image


lucasccordeiro (approved):

PR #5208, opened 2026-06-06
[solver] z3: fix SIGABRT building error trace for negative bv numerals

### LINKED ISSUE #5188: --k-induction + --overflow-check: SIGABRT (Z3 invalid argument) while building the error trace
## Introduction

Under `--k-induction --overflow-check`, ESBMC aborts with SIGABRT — `ERROR: Z3 error invalid argument encountered` — while building the error trace for a signed-overflow counterexample, instead of printing the trace and reporting `VERIFICATION FAILED`. The overflow itself is real (signed integer overflow is undefined behavior, C11 §6.5p5), so ESBMC is right to flag it; the defect is the crash during trace construction.

It is specific to k-induction: the same program with a bounded unwind (`--unwind`, no `--k-induction`) completes cleanly. The counterexample arises in the inductive step, where the accumulator and the added value are both unconstrained, so their sum overflows. Reproducing it needs a loop with a *symbolic* bound plus an *independent* symbolic value added in the body — a constant loop bound, or adding the loop counter itself, builds the trace without crashing. No contracts are involved.

## Example

```c
int g(int a, int b) {
    int x = a, prod = 0;
    while (x >= 0) {
        prod = prod + b;   /* signed overflow reachable in the inductive step */
        x--;
    }
    return prod;
}
```

```
$ esbmc repro.c --function g --k-induction --overflow-check
...
Building error trace
ERROR: Z3 error invalid argument encountered
[exit 134 — SIGABRT, core dumped]
```

The same program verifies cleanly when the loop is bounded instead of k-induced:

```
$ esbmc repro.c --function g --overflow-check --unwind 4 --no-unwinding-assertions
VERIFICATION SUCCESSFUL
```

Tested on ESBMC 8.3.0 (x86_64 linux), Z3 v4.16.0.

Yiannis128:
## Introduction

Under `--k-induction --overflow-check`, ESBMC aborts with SIGABRT — `ERROR: Z3 error invalid argument encountered` — while building the error trace for a signed-overflow counterexample, instead of printing the trace and reporting `VERIFICATION FAILED`. The overflow is real (signed integer overflow is undefined behavior, C11 §6.5p5), so flagging it is correct; the defect is the crash during trace construction.

The crash is specific to k-induction. The counterexample comes from the inductive step, where the accumulator and the added value are both unconstrained, so their sum overflows. The same program with a bounded unwind completes cleanly. Fixes #5188.

## Bug reproduction

```c
int g(int a, int b) {
    int x = a, prod = 0;
    while (x >= 0) {
        prod = prod + b;   /* signed overflow reachable in the inductive step */
        x--;
    }
    return prod;
}
```

```
$ esbmc repro.c --function g --k-induction --overflow-check
...
Building error trace
ERROR: Z3 error invalid argument encountered
[exit 134 — SIGABRT, core dumped]
```

Bounded instead of k-induced, the same program verifies:

```
$ esbmc repro.c --function g --overflow-check --unwind 4 --no-unwinding-assertions
VERIFICATION SUCCESSFUL
```

A gdb backtrace puts the abort in `z3_convt::get_bv` → `z3::expr::as_binary` → `Z3_get_numeral_binary_string`. Instrumenting `get_bv` shows the value Z3's model evaluation returns for the trace:

```
ast=(bvneg #x80000400)  ast_kind=NUMERAL  is_numeral=1
Z3_get_numeral_string=-2147484672
```

## Fix

The model evaluation returns the bit-vector value as `(bvneg #x80000400)` — a numeral that Z3 stores with a **negative** internal rational (`Z3_get_numeral_string` returns `-2147484672`; `simplify()` does not fold it). `z3::expr::as_binary` guards only on `is_numeral()`, which is true here, then calls `Z3_get_numeral_binary_string`. That API rejects negative numerals with `Z3_INVALID_ARG`, which trips ESBMC's Z3 error handler and aborts the process.

`get_bv` now reads the value with `Z3_get_numeral_string`, which handles the sign, and reinterprets it within the bit-vector's width via `integer2binary` / `binary2integer`. This covers both the normal non-negative numeral and the negative-internal-value case with one path:

```cpp
BigInt val = string2integer(Z3_get_numeral_string(z3_ctx, e));
return binary2integer(integer2binary(val, e.get_sort().bv_size()), is_signed);
```

The result is bit-exact with the old `as_binary` path for every normal value (verified across 1-bit, non-byte-aligned bitfields, all standard widths, `INT_MIN`/`INT64_MIN`, `__int128`, and fixedbv), so no counterexample value changes. For the case that used to abort, ESBMC now builds the trace and reports `VERIFICATION FAILED` with the genuine overflow.

Two regression tests are added under `regression/k-induction/`:

- `github_5188` — the repro under `--k-induction --overflow-check`, expecting `VERIFICATION FAILED` with `arithmetic overflow on add`. The pre-fix binary SIGABRTs here, so the test guards the crash.
- `github_5188_bounded` — the same program under a bounded unwind, expecting `VERIFICATION SUCCESSFUL`.

### DIFF
--- a/regression/k-induction/github_5188/main.c
+++ b/regression/k-induction/github_5188/main.c
@@ -1,0 +1,14 @@
+/* Regression for issue #5188: under --k-induction --overflow-check ESBMC used
+ * to SIGABRT ("Z3 error invalid argument encountered") while building the error
+ * trace for this signed-overflow counterexample, instead of reporting the
+ * (genuine) overflow. */
+int g(int a, int b)
+{
+  int x = a, prod = 0;
+  while (x >= 0)
+  {
+    prod = prod + b; /* signed overflow reachable in the inductive step */
+    x--;
+  }
+  return prod;
+}
--- a/regression/k-induction/github_5188/test.desc
+++ b/regression/k-induction/github_5188/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function g --k-induction --overflow-check
+^VERIFICATION FAILED$
+arithmetic overflow on add
--- a/regression/k-induction/github_5188_bounded/main.c
+++ b/regression/k-induction/github_5188_bounded/main.c
@@ -1,0 +1,13 @@
+/* Companion to issue #5188: the same loop verifies cleanly under a bounded
+ * unwind (no k-induction), confirming the overflow report in github_5188 is
+ * specific to k-induction's unconstrained inductive step. */
+int g(int a, int b)
+{
+  int x = a, prod = 0;
+  while (x >= 0)
+  {
+    prod = prod + b;
+    x--;
+  }
+  return prod;
+}
--- a/regression/k-induction/github_5188_bounded/main.c
+++ b/regression/k-induction/github_5188_bounded/main.c
@@ -1,9 +1,12 @@
-/* Companion to issue #5188: the same loop verifies cleanly under a bounded
- * unwind (no k-induction), confirming the overflow report in github_5188 is
- * specific to k-induction's unconstrained inductive step. */
+/* Companion to issue #5188: with the added value masked to a small range the
+ * accumulation cannot overflow within a bounded unwind, so ESBMC verifies the
+ * loop cleanly. Contrast with github_5188, where k-induction's unconstrained
+ * inductive step leaves the accumulator and the added value unbounded, so the
+ * same addition overflows. */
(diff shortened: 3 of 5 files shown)


lucasccordeiro:
@claude: Can you please review this PR?

rafaelsamenezes (approved):

PR #5238, opened 2026-06-08
[symex] constant-propagate structs that contain pointer fields

mikhailramalho:
## Problem

A struct `WITH`-chain was only constant-propagated when **every** field update was `is_constant_expr`. A pointer-typed field update (`NULL`, `&object`, a typecast thereof) is never `is_constant_expr`, so a struct containing *any* pointer field was never propagated. Its scalar fields then never constant-folded, leaving data-dependent recursion guards and loop bounds (e.g. `node->n`) symbolic — so symex explored the infeasible recursive/loop branches a concrete pointer would prune, an exponential path blow-up.

Concretely, `regression/esbmc-cpp/try_catch/nec_ex5-recursive` needed an explicit `--depth` bound to terminate at all; without it symex ran for minutes and the solver exhausted memory, even though the program is concrete and runs instantly.

### Root cause (verified)
With `--show-symex-value-sets` / `--show-vcc`: the value-set of the recursive formal is *precise* at feasible depths, but the guard `(struct WITH [n:=v] WITH [ptr:=p]).n` reads a struct **symbol** whose `WITH`-value is never inlined, because `constant_propagation` returns `false` for structs with a pointer field. An all-scalar struct (`{int n; int p;}`) folds; the same struct with a pointer field (`{int n; T* p;}`) does not — that single difference reproduces the blow-up.

## Fix

Use `constant_propagation` for the update values (as the array-`WITH` case directly below already does), but **restrict to scalar- and pointer-typed field updates**: inlining an aggregate update value (array / struct / union / nested `with`) can produce an expression the SMT backend cannot lower — a `char`-array member of a `std::map` node triggered `Typecast for unexpected type`. `nec_ex5-recursive` only needs int/pointer fields to fold.

## Result

- `nec_ex5-recursive` terminates on its own (~0.005s symex, previously OOM at solve); its `--depth` workaround is removed.
- New C regression tests: `struct_ptr_field_recursion_constprop` (guard folds, recursion stops at real depth → SUCCESSFUL) and `_fail` (chain deeper than `--unwind` → FAILED).

## Validation

- Core C `esbmc/` suite: **same 11 pre-existing baseline failures, zero new** (diffed against the unpatched binary).
- Container C++ suites (map/set/list/deque/unordered/string/bitset/…): **640/640 pass** — this is where the struct-with-array risk concentrates; an earlier broader version regressed `map_operator_brackets_2/3`, which the scalar/pointer restriction fixes.
- All three target tests pass via ctest.

### DIFF
--- a/regression/esbmc-cpp/try_catch/nec_ex5-recursive/test.desc
+++ b/regression/esbmc-cpp/try_catch/nec_ex5-recursive/test.desc
@@ -1,5 +1,5 @@
 THOROUGH
 main.cpp
---unwind 2 --no-unwinding-assertions --error-label ERROR --depth 131 --timeout 900
+--unwind 2 --no-unwinding-assertions --error-label ERROR --timeout 900
 
 ^VERIFICATION FAILED$
--- a/regression/esbmc/struct_ptr_field_recursion_constprop/main.c
+++ b/regression/esbmc/struct_ptr_field_recursion_constprop/main.c
@@ -1,0 +1,36 @@
+// Regression for constant propagation of struct values that contain a pointer
+// field.
+//
+// `f` recurses only while `p->n >= 1`. The concrete structure built in main
+// terminates the recursion at its real depth (2): root has n == 1 and points
+// at leaf, leaf has n == 0 so the guard is false and recursion stops.
+//
+// Before the fix, ESBMC refused to constant-propagate a struct value whose
+// WITH-chain updated a pointer-typed field (here `next`), because the pointer
+// update was not classified as a plain constant. As a result `p->n` never
+// folded, the recursion guard stayed symbolic, and symex explored the
+// recursive branch on every path up to the --unwind bound -- tripping the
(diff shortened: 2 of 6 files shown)


lucasccordeiro (approved):
Nice simplification!

mikhailramalho:
This PR 30s:
```
  correct:           20741
    correct true:    13877
    correct false:    6864
  incorrect:             8
    incorrect true:      0
    incorrect false:     8
  unknown:           15851
  Score:             34490 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27139279248/job/80148887650
```
Master 30s:
```
  correct:           20436
    correct true:    13714
    correct false:    6722
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           16157
  Score:             34038 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/26819351870/job/79073128102
```

PR #5244, opened 2026-06-08
[goto] remove the legacy imperative exception path (#5075)

### LINKED ISSUE #5075: [symex] Rewrite exception support to be fully symbolic

lucasccordeiro:
## What

"P4" of issue #5075: **deletes ESBMC's legacy imperative C++/Python exception handling**. Symbolic exception lowering (`remove_exceptions.cpp`) now covers the whole exception corpus (0 fall-backs after #5241 closed the empty-catch-all residual), so it becomes the **only** exception path.

## Behaviour change

Lowering now runs **unconditionally**. A program the pass cannot lower is reported as a **hard error** instead of silently falling back to the imperative path:

```
ERROR: exception lowering: cannot lower a concurrent program (pthread_create)
```

The unsupported set (now errors, accepted per the roadmap): concurrent programs that **use exceptions** (the global exception state is unsound across threads — per-thread state is a separate feature), `--function` isolated exception verification, and `dynamic_cast<T&>` whose `std::bad_cast` is unavailable (no `<typeinfo>`). An exception-**free** concurrent program is unaffected (the pass is a silent no-op).

## Deleted

- `src/goto-symex/symex_catch.cpp` (560 lines) and the `THROW` / `CATCH` / `THROW_DECL` / `THROW_DECL_END` symex dispatch (the `default:` `log_error`+`abort` is now the backstop for any instruction that escapes lowering — proven impossible, see below).
- All imperative exception state: `stack_catch`, `thrown_obj_map`, `last_throw`, `bad_cast_throw`, `inside_unexpected`, `unexpected_end`, `is_python_exception_subtype`, the `exceptiont` class, and the `symex_throw::thrown_obj` symbol. **`merge_state_map` is kept** (shared with normal call/return).
- The `--no-lower-exceptions` and `--extended-try-analysis` options (`--lower-exceptions` is kept as an accepted no-op for the 36 tests that pass it).
- The ON-vs-OFF differential gate (`scripts/lower_exceptions_differential.py` + its CI workflow) — meaningless with no OFF path.

`report_fallback` → `report_unsupported`, which throws ESBMC's fatal-error idiom (a `std::string` caught by `process_goto_program`, logged + returns failure — no `SIGABRT`).

## Soundness (Mode C / C-Dead, discharged)

The removed symex branches are proven **unreachable** post-lowering: `--goto-functions-only` dumps of representative programs (simple throw, rethrow, nested, Python `raise`) show **zero** surviving `THROW`/`CATCH`/`THROW_DECL` instructions — all rewritten to ASSIGN/GOTO/ASSERT. Confirmed under **both Bitwuzla and Z3**. The `default:` abort backstop would fire **visibly** (never silently) if any escaped.

## Simplifications (behaviour-equivalent, reviewed)

- `case ASSIGN:` drops the `thrown_obj_map` copy round-trip (the map had no remaining writer) → `symex_assign(instruction.code)`.
- `pop_frame` guard `if (!guard.is_false() && stack_catch.empty())` → `if (!guard.is_false())` (`stack_catch` was an invariant-empty conjunct).

## Tests

- `lower-exceptions_concurrent` and `lower-exceptions_bad_cast_unsupported` — repurposed to assert the new hard ERROR (two distinct unsupported categories).
- `lower-exceptions_concurrent_no_exc` — positive control: an exception-free concurrent program still verifies SUCCESSFUL (validates the `program_uses_exceptions` guard).
- Full `try_catch` suite 123/123, Python exception subset 31/31, broad C/C++ samples green. Verified under both solvers.

Fixes #5075


src/esbmc/options.cpp
```diff
@@ -381,16 +381,9 @@ const struct group_opt_templ all_cmd_options[] = {
      "Do not unroll bounded loops at goto level (need to enable "
      "--goto-unwind)"},
     {"slice-assumes", NULL, "Remove unused assume statements"},
-    {"extended-try-analysis",
-     NULL,
-     "Skip backward stack search for C++ throw targets"},
     {"lower-exceptions",
```
XLiZHI: Do we still need this option?

lucasccordeiro: Good call — it was a vestigial no-op (lowering is unconditional now). Removed the `--lower-exceptions` option entirely in 1b73a0faf6 and stripped it from the 30 regression `test.desc` flag lines that still passed it.

mikhailramalho:
hey, this is already covered by #5126

mikhailramalho (approved):
Ignore my previous comment; I'll work on top of this PR once it lands.

XLiZHI (approved):

PR #5249, opened 2026-06-08
[Concurrency] Skip context switches on globals that are never written

Anthonysdu:
Add a flag ` --cswitch-skip-readonly-globals ` to disable context switch points for global variables that are never written after initialization.

This PR: --context-bound 3 --cswitch-skip-readonly-globals  600s

https://github.com/esbmc/esbmc/actions/runs/28297708931

```
  correct:            1560
    correct true:     1093
    correct false:     467
  incorrect:            10
    incorrect true:      4
    incorrect false:     6
  unknown:            1606
  Score:              2429 (max: 5722)

```

Master: --context-bound 3

https://github.com/esbmc/esbmc/actions/runs/27905113453

```
  correct:            1557
    correct true:     1090
    correct false:     467
  incorrect:            10
    incorrect true:      4
    incorrect false:     6
  unknown:            1609
  Score:              2423 (max: 5722)

```


lucasccordeiro (changes requested):
Shall we also update the esbmc-wrapper.py script with the new flag?

Anthonysdu:
> Shall we also update the esbmc-wrapper.py script with the new flag?

Let's see if the flag improves current svcomp setting:

Two runs:

Master:  --smt-during symex

https://github.com/esbmc/esbmc/actions/runs/28036621802

```
  correct:            1183
    correct true:      810
    correct false:     373
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:            1986
  Score:              1881 (max: 5722)
```

This PR: --smt-during symex --cswitch-skip-readonly-globals 

https://github.com/esbmc/esbmc/actions/runs/28297701055

```
  correct:            1186
    correct true:      815
    correct false:     371
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:            1983
  Score:              1889 (max: 5722)
```

lucasccordeiro:
@Anthonysdu: There is a lot of dead code in this PR. Are you sure you need all that code for your solution?

lucasccordeiro:
Reviewed this with a focus on the soundness of the read-only-global filter, building the branch locally (aarch64, esbmc 8.3.0) and running adversarial + regression concurrency tests.

**Verdict: looks sound, and I couldn't break it.** I initially suspected the read filter could drop a real race on a shared *heap* object (writes are indirect, so the object is absent from both `ever_written_globals` and `address_taken_globals` → `may_be_written` returns false). It doesn't, for two reasons:
1. The `is_symbol2t(orig)` guard in `get_expr_globals` means pointer/heap-resolved objects are never filtered — only plain named globals are.
2. More fundamentally, the filter only ever drops a **read**, never a **write**. Any global written somewhere keeps its write as a context-switch point, which alone preserves every interleaving with the filtered read. The comment *"a write on its own establishes the may-be-written fact for the other side"* is exactly right.

Tested with the flag on: shared heap object (via global pointer and via thread arg), static global written only through `&g`, direct thread write, post-spawn main write, and a function-pointer-spawn case — all stay `VERIFICATION FAILED`. Real race tests `github_4715_rwset_fail`, `github_4634_fail`, `github_5137_fail` all stay FAILED; safe cases stay SUCCESSFUL (one even went from timeout → SUCCESSFUL — nice perf win).

**One thing before merge: the PR had no regression tests**, and the flag is enabled by default in `esbmc-wrapper.py`, so the soundness contract should be pinned. I pushed two that pass against this branch (`1bafefab68`):

- `regression/esbmc/cswitch_skip_readonly_globals/` (CORE, expect SUCCESSFUL) — read-only global stays safe under `--data-races-check`.
- `regression/esbmc/cswitch_skip_readonly_globals_fail/` (CORE, expect FAILED) — a written global must still race with the flag on; the read filter must not suppress it.

**Two minor nits:**
- `reachability_tree.h`: `bool readonly_global_opt = true;` — the default is dead (the constructor overwrites it from the option, which is off by default), and the comment refers to a non-existent flag `--no-cswitch-on-readonly-globals` (the real one is `--cswitch-skip-readonly-globals`). Suggest defaulting to `false` and fixing the comment.
- The internal-symbol whitelist is duplicated in `reachability_tree.cpp` (`is_internal_global_name`) and `execution_state.cpp` (`is_internal_name`); worth hoisting to one shared helper so they can't drift.

Nice optimization — with the tests added I'd be happy to see it merged.

PR #5252, opened 2026-06-08
Fix Linux ARM64 build: SVE types, 32-bit libc, Solidity stub

LukeW1999:
## Summary

Fixes several issues that prevent ESBMC from building on Linux ARM64 (e.g. Homebrew CI runners, AWS Graviton):

- **SVE builtin types**: `c2goto` crashed with `Unrecognized clang builtin type __SVFloat32_t` when building the libc model on ARM64 Linux. Add cases for all ARM SVE types in `get_builtin_type`, mapping each to the nearest scalar base type. Newer types (`SveMFloat8`, `SveCount`, `SveBoolx2/x4`) are guarded with `LLVM_VERSION_MAJOR` checks for compatibility with older LLVM versions.
- **32-bit libc**: `build.sh` tried to build `clib32.goto` on ARM64 Linux where 32-bit ARM headers are unavailable. Add `-DENABLE_BUNDLE_LIBC_32BIT=OFF` for `aarch64` in `build.sh`.
- **Solidity model**: ARM64 Clang backend does not support `_BitInt > 128` (same limitation as Apple Clang on macOS). Extend the existing macOS stub to also cover Linux `aarch64` in `c2goto/CMakeLists.txt`.
- **Release CI**: Add a weekly Linux ARM64 build job (LLVM 22, system install) to `release.yml` as a Homebrew compatibility check. Does not block the existing release flow.

These issues were discovered when submitting an ESBMC Homebrew formula (Homebrew/homebrew-core#282982), where Homebrew's Linux ARM64 CI runner exposed all three failures after a Homebrew LLVM update (22.1.5 → 22.1.7).


src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -1585,6 +1585,119 @@ bool clang_c_convertert::get_builtin_type(
     }
     // fallthrough
 
+  // ARM SVE (Scalable Vector Extension) types.
```
mikhailramalho: These must be mapped to our vector types

mikhailramalho (changes requested):
We need to create the correct vector types for sve

src/clang-c-frontend/clang_c_convert.cpp
LukeW1999: Sure. I would change to vector now.

LukeW1999: Hi @mikhailramalho,

I turn on a CI to check the ctest here: https://github.com/LukeW1999/esbmc_CI_mac/actions/runs/27141505066/job/80107650216

Should I also turn on the ctest for the PR or release? But I think intrigus turns the ARM check off since the CI of PR is relative slow already.

Or maybe I just leave it as it is. I can propose a PR to turn the ctest on or add a trigger later.

mikhailramalho: It looks like everything is failing?

LukeW1999: Sure, I would address it right now.

LukeW1999: Right now, all the failures I have seen detailly are all 32-bit tests (--32 flag). ARM64 does not have 32-bit libc headers available, so those tests fail with a parse error. I changed the regression option and here is the new CI https://github.com/LukeW1999/esbmc_CI_mac/actions/runs/27149356800/job/80135892108.

Hope it works properly.

LukeW1999:
Right now, all the failures I have seen in detail are from three known ARM64 limitations:

1. **32-bit tests** (`--32` flag): ARM64 has no 32-bit libc headers available, so any test that invokes ESBMC with `--32` fails with a parse error. These are skipped in the fork CI.
2. **Solidity tests**: The ARM64 Clang backend does not support `_BitInt > 128`, so the Solidity model is replaced with an empty stub on ARM64 (same restriction as macOS). All `esbmc-solidity` regression tests are therefore skipped on ARM64.
3. **Interval analysis unit test** (`Interval Analysis - Mult Arithmetic (signed)`): This appears to be a pre-existing ARM64 compatibility issue unrelated to this PR — the wrapped interval arithmetic produces a different result on ARM64. This is worth a separate investigation.

src/clang-c-frontend/clang_c_convert.cpp
mikhailramalho: maybe we should move the 32 bit tests to a separate dir and only run them when the 32-bit is enabled? An idea for a separate PR

LukeW1999: I would create an issue to record and propose a PR later.

lucasccordeiro:
@LukeW1999: Can you please create follow-up issues for this PR?

Issue #4796, opened 2026-05-24
[python] regression/quixbugs/reverse_linked_list: KNOWNBUG


lucasccordeiro:
`regression/quixbugs/reverse_linked_list` is marked `KNOWNBUG` — ESBMC does not produce the expected verdict.

Correct variant of the QuixBugs `reverse_linked_list` algorithm — verification should return `VERIFICATION SUCCESSFUL`.

- **Source**: [`regression/quixbugs/reverse_linked_list/main.py`](https://github.com/esbmc/esbmc/blob/master/regression/quixbugs/reverse_linked_list/main.py)
- **Flags**: `--incremental-bmc`
- **Expected output regexes**:
```
  ^VERIFICATION SUCCESSFUL$
```

## Reproduce

```sh
cd build
ctest -R "^regression/quixbugs/reverse_linked_list$" --output-on-failure
# or directly
esbmc regression/quixbugs/reverse_linked_list/main.py --incremental-bmc
```

## Definition of done

Either fix the underlying limitation and remove `KNOWNBUG` from `regression/quixbugs/reverse_linked_list/test.desc`, or delete the test if obsolete and document why.

Part of #4778.

Issue #4653, opened 2026-05-20
[python] OM rework needed to unblock --ir-gated bignum (toward #4642)


lucasccordeiro:
## Context

#4642 asks for arbitrary-precision Python `int`. The natural lever — widen the IR type under `--ir` so the SMT layer's unbounded `Int` sort gives bignum semantics — was attempted in #4652 (the PR landing the `python_int_typet()` helper) and reverted out of that PR after integration testing.

This issue tracks the operational-model work that has to land before the helper can actually be wired up to `get_typet("int")` and the constant-fold paths.

## What the PR's integration test surfaced

When the helper is wired up so `get_typet("int")` returns `signedbv(512)` under `--ir`, 19 existing `--ir` Python regression tests fail. The failures decompose into two categories:

1. **Inflated `__memcpy_impl` unwinds.** The C OM for lists / dicts / sets sizes per-element storage from the IR type's `width()` (`src/python-frontend/python_list.cpp` `get_list_element_info`, `else` branch at line ~386). A 512-bit element type produces `elem_size_bytes = 64`. Every list push, slice, copy, sort, extend then issues `memcpy(dst, src, 64)`, whose byte-by-byte loop in `src/c2goto/library/string.c:278` blows past the `--unwind 9` that the existing tests use.

2. **Storage/view width mismatch when the push-side size is capped.** A naive "cap the push-side `elem_size_bytes` at 8" fix gets us past the memcpy unwind blow-up, but the read paths (e.g. the `*(signed _ExtInt(512) *)return_value->value == 2` shape that ESBMC generates for `lst[0] == 2`) then cast 8-byte backing storage through a 64-byte view and read 56 bytes of uninitialised memory. This is a real soundness violation (`VERIFICATION FAILED` for a program that should succeed), not a tunable budget issue.

So the OM is currently structurally tied to a single fixed-width element representation, and bignum scalars cannot coexist with it without an OM-side rework.

## What the OM needs

A list/dict/set element representation that decouples the abstract Python int value from the byte width of the backing slot. Concretely, at least one of:

- **Boxed slots.** Each `PyObject` slot stores a pointer to the value, plus a tag for its abstract type. Reads/writes go through the pointer; the slot itself is fixed size regardless of the value's bit width. This matches CPython's actual representation and removes the byte-width coupling completely.
[...]

PR #5262, opened 2026-06-09
[solver] diagnose mismatched-width (dis)equality instead of crashing

lucasccordeiro:
Equality over operands of different bit-width (#4796: a Python object pointer vs a None modelled at width 0) was guarded only by an `assert` that NDEBUG elides, so release builds fed mismatched terms to the solver and crashed with a SIGSEGV.

Route every backend's `mk_eq`/`mk_neq` through a new `smt_convt::check_eq_operand_widths` that aborts with a clear diagnostic on mismatch — the controlled hard-stop debug builds already produced. No verdict changes and the bug is surfaced, not masked; the underlying #4796 typing defect stays open (#4653/#3067). Refs #4796.

### DIFF
--- a/regression/python/github_4796_mk_eq_width/main.py
+++ b/regression/python/github_4796_mk_eq_width/main.py
@@ -1,0 +1,39 @@
+# Regression for #4796: reversing a linked list and comparing the result to a
+# node with `==` made the frontend emit an equality between a Node object
+# pointer (64-bit) and a None modelled at width 0. The per-solver mk_eq asserted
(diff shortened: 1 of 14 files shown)


mikhailramalho:
Doesn't the solver already reject these types? And why doesn't our CI fail in these cases?

I would suggest we go ahead and fix the underlying issue instead of this PR.

lucasccordeiro:
Closed via https://github.com/esbmc/esbmc/pull/5274.

PR #5228, opened 2026-06-07
[k-induction] Disable inductive step on pointer-array writes

lucasccordeiro:
The k-induction inductive step havocs only named symbols, so a loop that writes an array element through a pointer (e.g. `(*dest)[i]`, as in SV-COMP's Intel-TDX-Module nondet-array builders) is left under-generalised and can report an unsound VERIFICATION SUCCESSFUL on an unsafe program — the cause of 54 incorrect "true" verdicts.

Detect such writes and disable the inductive step for the program; the base case and forward condition still run. Stack arrays, single pointer/field writes, and library helpers are left untouched, preserving existing inductive proofs (e.g. is_pointer_invariant_list).

Fixes #5224.

PR #5268, opened 2026-06-09
[k-induction] Phase 2: havoc resolved pointee objects for pointer-array loops

### LINKED ISSUE #5230: [k-induction] Phase 2: soundly havoc pointee objects for pointer-array loops (follow-up to #5224)
## Background

#5224 fixed an unsoundness where the k-induction inductive step proved unsafe programs "true" because it havocs only **named symbols** (`make_nondet_assign` in `goto_k_induction.cpp`), and so could not generalise an **array element written through a pointer** (`(*dest)[i] = …`, the SV-COMP Intel-TDX-Module nondet-array-builder shape).

The fix landed in #5228 (Phase 1) is deliberately **conservative**: when a user loop writes a pointer-reached array element, the inductive step is *disabled* for the whole program (base case and forward condition still run). This restores soundness but **loses the inductive proof** for programs that genuinely need it — any safe program whose proof requires induction over a heap/pointer array now falls back to UNKNOWN/timeout.

## Goal (Phase 2)

Make the inductive step **sound *and* precise** for pointer-array-writing loops, so they can be proven again instead of gated off. Concretely: when a loop writes through a pointer, **havoc the pointee object(s)** at loop entry rather than abandoning the inductive step.

## Approach sketch

- In `make_nondet_assign` (`goto_k_induction.cpp`), for a loop flagged `loopst::modifies_pointer_array`, emit a havoc of the *whole object(s)* the written pointer may reference, using the value-set / points-to analysis already enabled under `--add-symex-value-sets` (cf. the value-set assume in `src/goto-symex/symex_dereference.cpp`).
- For a value-set that is a single concrete object, assign nondet to that object; for multiple candidates, havoc each (or fall back to the Phase 1 gate when the set is `unknown`/unbounded, keeping soundness).
- Once a loop's pointer-reached writes are soundly havoc'd, drop it from the Phase 1 `modifies_pointer_array` gate so the inductive step runs again.

## Acceptance criteria

- The Intel-TDX-Module `unreach-call` tasks that #5224/#5228 turned from incorrect-`true` into UNKNOWN are either proven correctly or remain UNKNOWN — **never** incorrect-`true` again.
- Safe programs that need induction over a heap/pointer array (a new regression, plus ideally some recovered SV-COMP `correct true`) are proven SUCCESSFUL.
- `regression/k-induction/is_pointer_invariant_list` and the `github_5224_*` tests still pass.
- Dual-solver (Bitwuzla + Z3) agreement on the new proofs; a `needs-svcomp-run` to confirm net score recovery vs. Phase 1.

## Notes

- The Phase 1 helper `indexes_through_pointer` (`goto_loops.cpp`) and the `modifies_pointer_array` flag already identify exactly the loops Phase 2 must handle, so this builds directly on #5228.
- Soundness is the hard constraint: an incomplete havoc (missing a reachable pointee) reintroduces the #5224 unsoundness, so any case the value-set analysis cannot fully resolve must keep the Phase 1 gate.

Follow-up to #5224 / #5228.

lucasccordeiro:
Phase 1 (#5228) disabled the inductive step program-wide whenever a loop wrote an array element through a pointer. Phase 2 instead resolves the written pointer against the goto-level value-set fixpoint and, when it resolves to concrete named objects, havocs those objects as whole symbols and keeps the inductive step enabled and sound; it abstains (falling back to Phase 1) on unknown/heap/via-callee pointees. The fixpoint is built once on the pristine program, since value_set_analysist asserts on transformed CFGs.

Fixes #5230.


lucasccordeiro:
This PR (300s): https://github.com/esbmc/esbmc/actions/runs/27260663239

Master (300s):

```
  correct:           24161
    correct true:    15765
    correct false:    8396
  incorrect:            54
    incorrect true:     28
    incorrect false:    26
  unknown:           12385
  Score:             38614 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27098958843/job/79977402115
```

lucasccordeiro:
This PR (30s):

```
  correct:           20723
    correct true:    13858
    correct false:    6865
  incorrect:             3
    incorrect true:      0
    incorrect false:     3
  unknown:           15874
  Score:             34533 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27343167105/job/80788395390
```

Master (30s):

```
  correct:           20728
    correct true:    13853
    correct false:    6875
  incorrect:             3
    incorrect true:      0
    incorrect false:     3
  unknown:           15869
  Score:             34533 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27235445216/job/80429911459
```

PR #5272, opened 2026-06-09
[docs] Doxygen API reference pipeline

Yiannis128:
Sets up an auto-generated Doxygen API reference, published at https://esbmc.github.io/docs/api.

## How it fits together
- `.doxygen` — the single Doxyfile (generates HTML from `src/`; inheritance diagrams via dot; collaboration/call graphs off to bound build time; no project configure needed).
- `scripts/gen-docs.sh` — one wrapper that stamps `PROJECT_NUMBER` from `CMakeLists.txt` and runs `.doxygen`, optionally overriding the output location. The CMake `docs` target, the CI check, and the website deploy all call it, so the version and invocation live in exactly one place.
- `scripts/cmake/Docs.cmake` — `-DBUILD_DOC=On` drives `gen-docs.sh` (was `doxygen_add_docs`, which generated a separate config that could drift).
- `.github/workflows/doxygen.yml` — validates the docs build on master push, on `workflow_dispatch`, and on PRs that change the docs config/script/workflow; uploads the HTML artifact.

## Docs
- `website/content/docs/development/api-reference.md` — local build instructions; linked from the dev section index and the Build Guide.
- `CONTRIBUTIONS.md`, `docs/manual.tex` — now point to the published reference and that guide instead of carrying commands inline.
- `website/hugo.yaml` — "API Reference" navbar link to `/docs/api`.

The public deploy lives in the companion PR esbmc/esbmc.github.io#2, which runs `scripts/gen-docs.sh` against this repo. Merge both together so the navbar link, `/docs/api`, and the docs page resolve.


rafaelsamenezes (approved):

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,44 @@
...
+      - '.github/workflows/doxygen.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  doxygen:
+    runs-on: ubuntu-latest
```
XLiZHI: Can we use `ubuntu-slim` instead of `ubuntu-latest`? This script doesn't seem to require a lot of resources.

Yiannis128: Done

XLiZHI (approved):

PR #5283, opened 2026-06-09
[k-induction] constrain __CPROVER_uninterpreted_* calls to nondet non-zero

### LINKED ISSUE #5145: [SV-COMP] false alarm on unreach-call: aws_hash_table_create_harness
## Wrong result in SV-COMP 26 run

**False alarm**: ESBMC reports `VERIFICATION FAILED` (FALSE) on a benchmark whose ground truth is TRUE — a spurious counterexample on a correct program (precision/incompleteness bug).

| Field | Value |
|---|---|
| Property | `unreach-call` |
| Expected verdict | `true` |
| ESBMC verdict | `false(unreach-call)` |
| Benchmark | [`c/aws-c-common/aws_hash_table_create_harness.yml`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_create_harness.yml) ([source](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_create_harness.i)) |
| Run | [SV-COMP 26 benchexec run](https://github.com/esbmc/esbmc/actions/runs/26947362369) |
| ESBMC version | 8.3.0 (commit `0e3612575faf49c0851eb305dc00b27fbf4b4ebd`) |
| Strategy | `kinduction` |
| Related umbrella issue(s) | #2513 |

## Reproducer

Wrapper invocation (matches the SV-COMP run):

```
./esbmc-wrapper.py -p sv-benchmarks/c/properties/unreach-call.prp -s kinduction --arch 64 sv-benchmarks/c/aws-c-common/aws_hash_table_create_harness.i
```

Underlying ESBMC command:

```
./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check sv-benchmarks/c/aws-c-common/aws_hash_table_create_harness.i --64 --witness-output witness --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
```

To reproduce locally, fetch the benchmark from the [sv-benchmarks repo](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_create_harness.i).

## Spurious counterexample

```
Violated property:
  file /home/benchexec/sv-benchmarks/c/aws-c-common/aws_hash_table_create_harness.i line 208 column 22 function reach_error
  assertion 0
  0
```

```
VERIFICATION FAILED

Bug found (k = 1)

FALSE_REACH
```

lucasccordeiro:
SV-COMP harnesses written for CBMC declare `__CPROVER_uninterpreted_*` functions with concrete C bodies (e.g. `return (uint64_t)a`) but rely on CBMC treating them as SMT uninterpreted functions that never return 0. ESBMC inlines the concrete body, so `key=NULL` maps to `hash=0`; hash tables that reserve 0 as the empty-slot sentinel then immediately reject the entry and trip an assertion, producing a false alarm.

`constrain_cprover_uninterpreted_calls` is added to `process_goto_program` and inserts `ASSUME(result != 0)` after every call site matching the `__CPROVER_uninterpreted_` name pattern, matching CBMC uninterpreted-function semantics. Two regression tests are included: one that verifies the fix (github_5145) and one that confirms genuine violations are still reported (github_5145_fail).

Fixes #5145

### DIFF
--- a/regression/k-induction/github_5145/test.desc
+++ b/regression/k-induction/github_5145/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--k-induction --max-inductive-step 3 --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/k-induction/github_5145_fail/main.c
+++ b/regression/k-induction/github_5145_fail/main.c
@@ -1,0 +1,21 @@
+// Negative variant for issue #5145: ensures the __CPROVER_uninterpreted_*
+// fix does not suppress genuine violations.  After ASSUME(hash != 0) is
+// applied, hash is nondet in [1, UINT64_MAX]; the assertion below can be
+// violated (e.g. hash == 0x42), so ESBMC must still report FAILED.
+#include <assert.h>
+#include <stdint.h>
+
+uint64_t __CPROVER_uninterpreted_hasher(const void *a)
+{
+  return (uint64_t)a;
+}
+
+int main(void)
+{
+  void *key;
+  uint64_t hash = __CPROVER_uninterpreted_hasher(key);
+  // hash is nondet non-zero after the fix; 0x42 is a valid non-zero value,
+  // so this assertion is genuinely reachable.
+  assert(hash != 0x42ULL);
+  return 0;
+}
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -2295,6 +2295,55 @@ bool esbmc_parseoptionst::parse_goto_program(
   return false;
 }
 
+/// CBMC-compatible modelling of __CPROVER_uninterpreted_* functions.
+///
+/// SV-COMP harnesses use these names to declare hash or comparator functions
+/// whose semantics should be truly uninterpreted (any-output, excluding the
+/// zero/NULL sentinel).  CBMC's SMT backend treats them as UF axioms and
+/// never maps NULL → 0.  ESBMC inlines the concrete C body
+/// (e.g. `(uint64_t)a`), which maps NULL → 0 and produces false alarms
+/// whenever a hash table rejects the zero sentinel (issue #5145).
+///
+/// For each call whose bitvector return value is stored, insert
+/// ASSUME(result != 0) immediately after the call.  This is a sound
+/// over-approximation: on real 64-bit targets, non-NULL pointers cast to
+/// uint64_t are always non-zero, and NULL as a hash key is disallowed by
+/// every hash table that reserves 0 as an empty-slot marker.
+static void constrain_cprover_uninterpreted_calls(
+  goto_functionst &goto_functions)
+{
+  Forall_goto_functions(func_it, goto_functions)
+  {
+    if (!func_it->second.body_available)
+      continue;
+    goto_programt::instructionst &insns =
+      func_it->second.body.instructions;
+
+    for (auto it = insns.begin(); it != insns.end(); ++it)
+    {
+      if (!it->is_function_call())
+        continue;
+      const code_function_call2t &call =
+        to_code_function_call2t(it->code);
+      if (!is_symbol2t(call.function))
+        continue;
+      const std::string &fname =
+        id2string(to_symbol2t(call.function).thename);
+      if (fname.find("__CPROVER_uninterpreted_") == std::string::npos)
+        continue;
+      if (is_nil_expr(call.ret) || !is_bv_type(call.ret->type))
+        continue;
+
+      goto_programt::instructiont assume_instr;
+      assume_instr.type = ASSUME;
+      assume_instr.guard =
+        notequal2tc(call.ret, gen_zero(call.ret->type));
+      assume_instr.location = it->location;
+      insns.insert(std::next(it), std::move(assume_instr));
+    }
+  }
+}
+
 // This method performs various analyses and transformations
 // on the given GOTO program. They involve all the techniques that we class
 // as "static analyses" - performed on the given GOTO program before it is
@@ -2467,6 +2516,17 @@ bool esbmc_parseoptionst::process_goto_program(
         cse.run(goto_functions);
       }
     }
+
+    // Model __CPROVER_uninterpreted_* functions as nondet non-zero (issue
+    // #5145).  SV-COMP harnesses built for CBMC use these functions as hash
+    // functions with the convention that 0 is the empty-slot sentinel and
+    // is never a valid hash.  CBMC handles this via SMT UF axioms; ESBMC
+    // inlines the concrete C body, which maps NULL → 0.  Runs after partial
+    // inlining so FUNCTION_CALL sites are still present (partial inlining
+    // with smallfunc_limit = 0 does not inline non-trivial functions).
+    // NOTE: --full-inlining erases all CALL instructions before this point;
+    // use without --full-inlining for the fix to take effect.
+    constrain_cprover_uninterpreted_calls(goto_functions);
 
     bool is_k_induction = cmdline.isset("inductive-step") ||
                           cmdline.isset("k-induction") ||
--- a/src/esbmc/esbmc_parseoptions.cpp
+++ b/src/esbmc/esbmc_parseoptions.cpp
@@ -2309,26 +2309,23 @@ bool esbmc_parseoptionst::parse_goto_program(
 /// over-approximation: on real 64-bit targets, non-NULL pointers cast to
 /// uint64_t are always non-zero, and NULL as a hash key is disallowed by
 /// every hash table that reserves 0 as an empty-slot marker.
-static void constrain_cprover_uninterpreted_calls(
-  goto_functionst &goto_functions)
+static void
+constrain_cprover_uninterpreted_calls(goto_functionst &goto_functions)
 {
-  Forall_goto_functions(func_it, goto_functions)
+  Forall_goto_functions (func_it, goto_functions)
   {
     if (!func_it->second.body_available)
       continue;
-    goto_programt::instructionst &insns =
-      func_it->second.body.instructions;
+    goto_programt::instructionst &insns = func_it->second.body.instructions;
 
     for (auto it = insns.begin(); it != insns.end(); ++it)
     {
       if (!it->is_function_call())
         continue;
-      const code_function_call2t &call =
-        to_code_function_call2t(it->code);
+      const code_function_call2t &call = to_code_function_call2t(it->code);
       if (!is_symbol2t(call.function))
         continue;
-      const std::string &fname =
-        id2string(to_symbol2t(call.function).thename);
+      const std::string &fname = id2string(to_symbol2t(call.function).thename);
       if (fname.find("__CPROVER_uninterpreted_") == std::string::npos)
         continue;
       if (is_nil_expr(call.ret) || !is_bv_type(call.ret->type))
@@ -2336,8 +2333,7 @@ static void constrain_cprover_uninterpreted_calls(
 
       goto_programt::instructiont assume_instr;
       assume_instr.type = ASSUME;
-      assume_instr.guard =
-        notequal2tc(call.ret, gen_zero(call.ret->type));
+      assume_instr.guard = notequal2tc(call.ret, gen_zero(call.ret->type));
       assume_instr.location = it->location;
       insns.insert(std::next(it), std::move(assume_instr));
     }
--- a/regression/k-induction/github_5145/main.c
+++ b/regression/k-induction/github_5145/main.c
@@ -1,0 +1,30 @@
+// Issue #5145: false alarm with k-induction on aws_hash_table_create_harness.
+// __CPROVER_uninterpreted_hasher is defined concretely as (uint64_t)a, which
+// maps NULL to 0. The hash table uses 0 as the empty-slot sentinel and
+// rejects entries with hash_code == 0.  When key is a nondet void pointer,
+// ESBMC explores key = NULL -> hash_code = 0 -> assertion fires, producing a
+// false alarm.
+//
+// CBMC treats __CPROVER_uninterpreted_* functions as truly uninterpreted
+// (black-box SMT functions) and does not explore the hash_code == 0 path.
+// ESBMC matches this by adding ASSUME(result != 0) after calls to such
+// functions, so the spurious path is pruned.
+#include <assert.h>
+#include <stdint.h>
+
+// SV-COMP harness pattern: concrete body, but CBMC-uninterpreted semantics.
+uint64_t __CPROVER_uninterpreted_hasher(const void *a)
+{
+  return (uint64_t)a;
+}
+
+int main(void)
+{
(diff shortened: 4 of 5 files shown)


src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2295,6 +2295,51 @@ bool esbmc_parseoptionst::parse_goto_program(
...
+    {
+      if (!it->is_function_call())
+        continue;
+      const code_function_call2t &call = to_code_function_call2t(it->code);
+      if (!is_symbol2t(call.function))
+        continue;
+      const std::string &fname = id2string(to_symbol2t(call.function).thename);
+      if (fname.find("__CPROVER_uninterpreted_") == std::string::npos)
+        continue;
+      if (is_nil_expr(call.ret) || !is_bv_type(call.ret->type))
```
lucasccordeiro, line 2331: **Soundness: this blanket assumption can mask genuine violations.**

`is_bv_type` excludes `_Bool` (so `bool __CPROVER_uninterpreted_equals` escapes), but **not `int`** — and the name match catches every `__CPROVER_uninterpreted_*` function, including comparator-style ones where 0 is a legitimate, necessary return value. The SV-COMP aws-c-common harnesses this PR targets declare exactly that: e.g. [`aws_priority_queue_init_dynamic_harness-1.i`](https://github.com/sosy-lab/sv-benchmarks/blob/master/c/aws-c-common/aws_priority_queue_init_dynamic_harness-1.i) has `int __CPROVER_uninterpreted_compare(...)` returning −1/0/1 and relies on the 0 ("equal") case. With this pass, the equal-elements path is assumed away, so any violation reachable only through it is silently suppressed — `VERIFICATION FAILED` flips to `VERIFICATION SUCCESSFUL`. Minimal repro that the PR would wrongly verify:

```c
int __CPROVER_uninterpreted_equals(const void *a, const void *b) { return a == b; }
int main() {
  void *x, *y;
  int e = __CPROVER_uninterpreted_equals(x, y);
  assert(e != 0); // genuinely violable; the inserted ASSUME(e != 0) makes it vacuous
}
```

(Verified `VERIFICATION FAILED` on master; the inserted ASSUME makes it pass.) Since the pass also runs unconditionally — no command-line option gates it — this applies to any program whose function names contain the pattern, not just SV-COMP harnesses. Suggest restricting the non-zero assumption to the hasher convention (exact base-name match) and/or gating the pass behind an option.

lucasccordeiro (commented):
The false-alarm class is real and having both a passing and a failing regression test is the right shape. However, I found a soundness problem with the blanket non-zero assumption, plus a factual issue with the CBMC-semantics premise the fix is built on, so I think this needs rework before merging. Detailed findings inline; summary:

1. **Soundness (blocking):** `ASSUME(ret != 0)` is applied to *every* `__CPROVER_uninterpreted_*` function with a bitvector return, but the SV-COMP aws-c-common harnesses also declare `int __CPROVER_uninterpreted_compare(...)` returning −1/0/1, where 0 ("equal") is a legitimate, necessary value. Pruning it can flip a genuine `VERIFICATION FAILED` into `SUCCESSFUL`. The pass is also unconditional, with no flag to disable it.
2. **Premise:** the CProver manual only guarantees functional congruence for `__CPROVER_uninterpreted_*` — there is no non-zero guarantee, so "CBMC … never maps NULL → 0" is not accurate, and "sound over-approximation" is inverted (an injected ASSUME prunes paths; it can only hide violations).
3. **Implementation vs description:** nothing becomes nondet — the concrete body still executes; the assume is effectively `ASSUME(key != NULL)` at each call site.
4. **Mechanics:** unanchored substring match on the symbol id, and the pass lives in the CLI driver rather than `src/goto-programs/`.
5. **Tests:** `--incremental-bmc` is dead weight next to `--k-induction` (verified: byte-identical traces with/without).

My suggestion: either restrict the assumption to an exact hasher base name and gate it behind an option, or (better, and it also removes the `--full-inlining` caveat) model these functions as true uninterpreted functions — ignore the body and return a fresh nondet value with functional congruence.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2295,6 +2295,51 @@ bool esbmc_parseoptionst::parse_goto_program(
   return false;
 }
 
+/// CBMC-compatible modelling of __CPROVER_uninterpreted_* functions.
+///
+/// SV-COMP harnesses use these names to declare hash or comparator functions
+/// whose semantics should be truly uninterpreted (any-output, excluding the
+/// zero/NULL sentinel).  CBMC's SMT backend treats them as UF axioms and
+/// never maps NULL → 0.  ESBMC inlines the concrete C body
```
lucasccordeiro, line 2303: **The CBMC premise stated here is not accurate.** The CProver manual ([modeling-nondeterminism.md](https://github.com/diffblue/cbmc/blob/develop/doc/cprover-manual/modeling-nondeterminism.md)) defines `__CPROVER_uninterpreted_*` as *bodyless* functions whose value is "non-deterministic but different invocations will return the same value if their inputs are the same" — functional congruence only. There is **no non-zero guarantee**; the solver is free to choose `f(NULL) = 0`. (Upstream aws-c-common doesn't need one either: `s_hash_for()` in `source/hash_table.c` remaps zero hashes with `if (!hash_code) { hash_code = 1; }`, and the harness wrapper only asserts `a != NULL` — worth checking what the sv-benchmarks snapshot in #5145 actually trips on.)

Also, "sound over-approximation" (lines 2308–2309) is inverted: injecting an ASSUME *prunes* execution paths, i.e. it under-approximates the behaviours checked — it can never introduce a false alarm, but it can hide real ones. That is exactly the failure mode of the comparator case flagged in the other comment.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2468,6 +2513,17 @@ bool esbmc_parseoptionst::process_goto_program(
       }
     }
 
+    // Model __CPROVER_uninterpreted_* functions as nondet non-zero (issue
```
lucasccordeiro, line 2516: **This doesn't model the call as "nondet non-zero"** — the concrete body still executes, so for the hasher the result remains the deterministic `(uint64_t)key`, and the inserted `ASSUME(result != 0)` is operationally `ASSUME(key != NULL)` injected at every call site. The PR title and the test comments ("hash is nondet in [1, UINT64_MAX]") describe semantics the code doesn't implement.

If CBMC-compatible UF semantics is the goal, the deeper fix is to *ignore/replace the body* for these functions: havoc the return value (fresh nondet per call, plus functional congruence if you want full UF semantics). That would also eliminate the `--full-inlining` caveat below — the hole exists only because the fix is keyed on FUNCTION_CALL instructions surviving until this point in the pipeline.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2295,6 +2295,51 @@ bool esbmc_parseoptionst::parse_goto_program(
...
+
+    for (auto it = insns.begin(); it != insns.end(); ++it)
+    {
+      if (!it->is_function_call())
+        continue;
+      const code_function_call2t &call = to_code_function_call2t(it->code);
+      if (!is_symbol2t(call.function))
+        continue;
+      const std::string &fname = id2string(to_symbol2t(call.function).thename);
+      if (fname.find("__CPROVER_uninterpreted_") == std::string::npos)
```
lucasccordeiro, line 2329: Two mechanical points:

1. **Unanchored substring match** — this fires on any symbol id that embeds the pattern anywhere (e.g. a user function `my__CPROVER_uninterpreted_helper_wrapper` would silently get the ASSUME injected). The in-tree idiom for matching intrinsic-prefixed functions is an anchored prefix, e.g. `has_prefix(id2string(f_it->first), "c:@F@__VERIFIER_atomic_")` in `src/goto-programs/add_race_assertions.cpp:290`. Since static functions get file-qualified USRs (`c:file.c@F@name`), the robust form is to extract the base name after the last `'@'` and require `has_prefix(base_name, "__CPROVER_uninterpreted_")`.

2. **Placement** — sibling GOTO transformations invoked from `process_goto_program` live as their own passes under `src/goto-programs/` (e.g. `assign_params_as_non_det`, `mark_decl_as_non_det` via `goto_functions_algorithm`). A static helper in the CLI driver is harder to test and reuse; consider moving it there.

regression/k-induction/github_5145/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--k-induction --max-inductive-step 3 --incremental-bmc
```
lucasccordeiro, line 3: `--incremental-bmc` is dead weight here (and in `github_5145_fail/test.desc`): with both flags set, the k-induction block in `do_bmc_strategy` runs first and returns on any conclusive result, so the incremental block is never reached — I verified the strategy traces are byte-identical with and without it on this test. No other test in `regression/k-induction/` combines the two flags. Suggest dropping it to keep the flag set minimal and faithful to the issue repro:

```suggestion
--k-induction --max-inductive-step 3
```

regression/k-induction/github_5145_fail/main.c
```diff
@@ -0,0 +1,21 @@
+// Negative variant for issue #5145: ensures the __CPROVER_uninterpreted_*
+// fix does not suppress genuine violations.  After ASSUME(hash != 0) is
+// applied, hash is nondet in [1, UINT64_MAX]; the assertion below can be
```
lucasccordeiro, line 3: Nit: "hash is nondet in [1, UINT64_MAX]" isn't what the implementation does — `hash` is still the concrete `(uint64_t)key`; only `key` is nondet, and the inserted ASSUME excludes `key == NULL`. The test still fails for the right reason (a non-NULL `key` with `(uint64_t)key == 0x42` exists), but the comment should describe the actual mechanism to avoid misleading future readers. Same wording appears in the PR title and the positive test's comments.

src/esbmc/esbmc_parseoptions.cpp
lucasccordeiro: Confirmed against the freshly-built branch binary: the reviewer's exact repro (`int __CPROVER_uninterpreted_equals` + `assert(e != 0)`) now returns `VERIFICATION SUCCESSFUL` — a genuine violation masked. Blocking and correct; the blanket non-zero ASSUME is being dropped.

src/esbmc/esbmc_parseoptions.cpp
lucasccordeiro: Agreed. Verified on the actual benchmark: `s_hash_for` already remaps `NULL→42` and `0→1` and asserts `hash_code != 0` itself, so the zero sentinel is handled in-harness — no non-zero guarantee is needed or implied. The "sound over-approximation" wording is inverted (an injected ASSUME prunes paths → can only hide violations). Doc comment and premise are wrong and going away.

src/esbmc/esbmc_parseoptions.cpp
lucasccordeiro: Correct — the concrete body still executes, so the insert is operationally `ASSUME(key != NULL)`, not "nondet non-zero". I hand-modeled the proper CBMC-style semantics (ignore body, fresh nondet + functional congruence + the harness's symmetry assumes) on the real benchmark to test your suggestion; see the summary comment — it still reports FAILED, so the gap is deeper than the call modelling.

src/esbmc/esbmc_parseoptions.cpp
lucasccordeiro: Both valid: the unanchored `find()` would fire on `my__CPROVER_uninterpreted_helper`, and the right idiom is the anchored base-name prefix (`has_prefix` after the last `@`, per `add_race_assertions.cpp`). A live GOTO transform also belongs in `src/goto-programs/`, not the CLI driver. Noted for whatever replaces this pass.

regression/k-induction/github_5145/test.desc
lucasccordeiro: Agreed — with `--k-induction` set, `do_bmc_strategy` returns on the first conclusive result and the incremental block is never reached, so `--incremental-bmc` is dead weight. Will drop it from both `test.desc` files.

regression/k-induction/github_5145_fail/main.c
lucasccordeiro: Right — `hash` stays the concrete `(uint64_t)key`; only `key` is nondet and the ASSUME excludes `key == NULL`. The wording is misleading and matches the (incorrect) mental model behind the whole pass. Since the pass is being reworked/removed, these synthetic tests go with it.

lucasccordeiro:
## Summary of review — converting to draft

I verified each point against a freshly-built branch binary, and the headline is that this approach should not merge:

**1. Unsound (confirmed).** The comparator repro returns `VERIFICATION SUCCESSFUL` where master gives `FAILED`:
```c
int __CPROVER_uninterpreted_equals(const void *a, const void *b) { return a == b; }
int main(){ void *x,*y; int e = __CPROVER_uninterpreted_equals(x,y); assert(e != 0); }
```
The blanket `ASSUME(ret != 0)` masks a genuine violation.

**2. Premise inaccurate (confirmed).** In the real `aws_hash_table_create_harness.i`, `s_hash_for` already maps `NULL→42` and remaps `0→1`, then asserts `hash_code != 0` itself. The zero sentinel is handled in-harness; CProver guarantees only functional congruence, never non-zero.

**3. The fix does not resolve #5145.** Running the actual benchmark with this pass still gives `VERIFICATION FAILED`. The real false alarm is the post-create assertion `__VERIFIER_assert(uninterpreted_equals(p_elem->key, key))`, unrelated to `hash == 0`.

**4. The suggested "true UF" rework also does not resolve it.** I hand-modelled all three (`hasher`/`equals`/`compare`) as fresh nondet + functional congruence, plus the harness's own symmetry/antisymmetry assumes (memo capacity up to 512) — still `FAILED`. The deeper cause appears to be that the harness allocates the table with `bounded_malloc` and leaves slots havoc'd, so a slot's `hash_code` need not equal `hasher(slot.key)`; ESBMC then explores a stored-table state that CBMC's encoding does not. That's a symex/pointer-model question, not a call-site one.

**Action:** converting this PR to **draft**. #5145 stays open; the real culprit is the havoc'd hash-table slots, which needs a separate root-cause investigation. The unsound non-zero pass and the synthetic tests built around it will be removed rather than reshaped.

lucasccordeiro:
Closing: the non-zero ASSUME approach is unsound (masks comparator `== 0` violations) and does not resolve #5145 — the real benchmark still reports FAILED, and the congruent-UF rework doesn't help either. Root cause is the havoc'd hash-table slots, tracked separately in #5287. #5145 stays open.

PR #5291, opened 2026-06-10
CBMC goto frontend

rafaelsamenezes:
Prrety much just asked CLAUDE to convert goto-transcoder into ESBMC.

I still need to test it a bit and cleanup.


src/goto-programs/cbmc_adapter.cpp
```diff
@@ -0,0 +1,418 @@
...
+void expand_anon_struct(const irept &self)
+{
+  if (has_sub(self, "components"))
+    return;
+  // ESBMC has no parser for CBMC's anonymous naming convention.
+  const std::string ident = self.find("identifier").id_string();
+  if (ident.size() < 11 || ident.compare(0, 10, "tag-#anon#") != 0)
+    return;
+  log_error("CBMC adapter: unsupported anonymous struct {}", ident);
+  abort();
```
lucasccordeiro, line 173: `abort()` here (and at the type-19 / unresolved-target / struct_tag / c_bool sites) crashes ESBMC on malformed *or merely unsupported* input. Prefer a recoverable error: give `adapt_cbmc_to_esbmc` a `bool` return (or out-param) and propagate up to `read_cbmc_goto_object`, which already returns `true` on error.

src/goto-programs/read_cbmc_goto_object.cpp
```diff
@@ -0,0 +1,274 @@
...
+  while (in.good())
+  {
+    if (shift_distance >= 32)
+    {
+      log_error("CBMC goto-binary: input number is too large");
+      abort();
+    }
+
+    unsigned byte = static_cast<unsigned char>(in.get());
+    res |= (byte & 0x7f) << shift_distance;
```
lucasccordeiro: `(byte & 0x7f) << shift_distance` on a 32-bit `unsigned` silently drops bits 32-34 at `shift_distance==28` — the `>=32` guard only fires on the *next* byte, so an over-wide varint truncates instead of erroring. Accumulate into `uint64_t` (and range-check) to make over-width input detectable.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -2214,6 +2223,56 @@ bool esbmc_parseoptionst::read_goto_binary(goto_functionst &goto_functions)
...
+// CBMC binary so goto_convert only ever sees the boilerplate's symbols.
+bool esbmc_parseoptionst::synthesize_cprover_additions(
+  optionst &options,
+  goto_functionst &goto_functions)
+{
+  file_operations::tmp_file tf =
+    file_operations::create_tmp_file("esbmc-cprover-%%%%-%%%%-%%%%.c");
+  static const char boilerplate[] =
+    "/* Auto-generated: bundle all ESBMC additions for CBMC gotos. */\n"
+    "int main(void) { return 0; }\n";
```
lucasccordeiro, line 2278: The synthesised `__ESBMC_main` calls this boilerplate `main`, not the CBMC binary's entry (`__CPROVER__start`/`--function` harness) — so a real binary verifies an empty program and can report a false `SUCCESSFUL`. Per roadmap §4.2 this is the entry-bridging gap; until it's wired, emit a prominent warning (or refuse) so users don't trust the verdict.

src/goto-programs/goto_binary_reader.cpp
```diff
@@ -24,5 +25,19 @@ bool goto_binary_reader::read_goto_binary(
...
+  // Auto-detect the format from the magic header: a CBMC goto-binary starts
+  // with 0x7f 'G' 'B' 'F', whereas ESBMC's own format starts with 'G' 'B' 'F'.
+  char hdr[4] = {0, 0, 0, 0};
+  in.read(hdr, 4);
+  std::streamsize got = in.gcount();
+  in.clear();
+  in.seekg(0, std::ios::beg);
+
+  if (
+    got >= 4 && static_cast<unsigned char>(hdr[0]) == 0x7f && hdr[1] == 'G' &&
```
lucasccordeiro: The `0x7f 'G' 'B' 'F'` magic check is duplicated here, in `has_cbmc_binary_input()`, and in `parse_cbmc_goto()`. Factor a single `is_cbmc_goto_magic(const unsigned char[4])` helper to keep them from drifting.

src/esbmc/options.cpp
```diff
@@ -214,7 +214,9 @@ const struct group_opt_templ all_cmd_options[] = {
     {"no-library", NULL, "Disable built-in abstract C library"},
     {"no-string-literal", NULL, "Ignore string literals (replace with NULL)"},
     {"binary", NULL, "Read goto program instead of source code"},
-    {"cprover", NULL, "Add compatibility layer for CPROVER gotos"},
+    {"no-cprover-additions",
```
lucasccordeiro, line 221: Dropping `--cprover` outright breaks any existing script that passes it (now "unknown option"). Consider keeping it as a deprecated no-op for one release with a warning, rather than removing it.

src/goto-programs/cbmc_adapter.cpp
```diff
@@ -0,0 +1,418 @@
...
+irept symbol_to_esbmc_irep(const cbmc_symbolt &sym)
+{
+  irept result;
+  result.add("type") = sym.stype;
+  result.add("symvalue") = sym.value;
+  result.add("location") = sym.location;
+  result.add("module") = mk(sym.module);
+  result.add("mode") = mk(sym.mode);
+
+  // fix_name is the identity in adapter.rs.
```
lucasccordeiro, line 266: `is_weak` / `is_volatile` / `is_thread_local` / `is_property` are parsed but never mapped into the ESBMC symbol (roadmap §4.5). Silently dropping `volatile`/`thread_local` is a soundness risk — at minimum `log_warning` when a dropped flag is actually set.

lucasccordeiro (commented):
Solid foundation — clean reader/adapter split, good unit coverage, and the roadmap doc is excellent for scoping. Comments below are mostly robustness/soundness; several overlap with gaps you already flag in the roadmap, but they're worth pinning at the code since this is on a path to ingesting untrusted binaries.

Top concerns:
- **`abort()` on malformed/unsupported input.** A bad or merely unsupported CBMC binary takes down the whole process. ESBMC should reject it with a recoverable error. The reader already has a `bool` error channel; the adapter has none — give `adapt_cbmc_to_esbmc` one.
- **Wrong entry point (correctness blocker).** The synthesised `__ESBMC_main` wraps the boilerplate `main`, not the CBMC entry — so real harnesses verify an empty program. Until §4.2 lands, this risks a false `SUCCESSFUL`; please warn loudly / refuse non-trivial binaries rather than silently verify the wrong thing.
- Removing `--cprover` is a breaking CLI change — prefer accept-and-ignore for a release.

Since it's marked WIP, treat these as direction-setting rather than blocking.

src/goto-programs/cbmc_adapter.cpp
rafaelsamenezes: This is an unrecoverable error. Better crash and burn than to pretend to accept an input.

PR #5303, opened 2026-06-11
[om] Allow C++ operational models in the embedded clib (#5298)

### LINKED ISSUE #5298: [goto] Compiling any C++ file into the embedded clib breaks goto_convert (blocks C++ operational models)
## Summary

On current `master`, compiling **any** C++ source file (`library/*.cpp`) into the embedded `clib*.goto` corrupts the library: every subsequently-verified program — including a trivial C `for` loop with no exception or C++ content — fails goto conversion with:

```
ERROR: goto_convert: non-code operand in this block:
```

The malformed block is a `for`-loop whose body became an untyped `code` block with a `nil` first operand (an unlowered `code_for`).

This blocks landing the C++ exception **operational model** (the next stage of the #5126 split), because the OM is delivered as a compiled `src/c2goto/library/exception.cpp` that must be compiled into the clib. But the bug is **not exception-specific** — it reproduces with any `.cpp` in the clib.

## Reproduction

1. Enable the C++ library glob in `src/c2goto/CMakeLists.txt` (`inputs_c ${c2goto_library_files} ${c2goto_cpp_library_files}`) and `LANGAPI_MODE_CLANG_CPP` in `c2goto.cpp`.
2. Add a trivial, non-exception C++ file to the library:
   ```cpp
   // src/c2goto/library/__probe.cpp
   extern "C" int __esbmc_probe_fn(int x) { return x + 1; }
   ```
3. Reconfigure + rebuild, then verify any program:
   ```cpp
   int main() { int s = 0; for (int i = 0; i < 3; i++) s += i; return s; }
   ```
   ```
   esbmc nofexc.cpp --std c++17   →   ERROR: goto_convert: non-code operand in this block
   ```
4. Remove `__probe.cpp` (or the `${c2goto_cpp_library_files}` from `inputs_c`), rebuild → the same program reports `VERIFICATION SUCCESSFUL`.

The malformed block (from the failing run):
```
code
  * operands:
    0: nil
    1: code
        * operands:
          0: nil
          1: <        (i < size)   <- from __VERIFIER_nondet_memory's for-loop in verifier.c
```

## Isolation (what's ruled out)

- **Not the exception OM.** A trivial non-exception `.cpp` reproduces it.
- **Not clib generation.** The combined C+C++ clib generates without error and contains the expected symbols.
- **Not the frontend/c2goto source of the OM work.** Building the full original exception-rework branch (which embeds the same `exception.cpp` into the clib) and running the identical `nofexc.cpp` test → **`VERIFICATION SUCCESSFUL`**. Swapping that branch's clib into a `master`-based `esbmc` still **fails** → the defect is on the **`esbmc` (goto-conversion) side**, surfacing only with a C++-origin body in the loaded library.

## Likely cause

The original exception-rework branch predates the **irep2 V.4.x structured-control-flow work** (`code_ifthenelse` / `code_while` / `code_for` / `code_decl`-with-init, #5266 / #5277 / #5282 / #5284). That branch lowered C++ structured control flow before serialization; on current `master`, a C++ library body appears to retain `code_for` in the serialized clib, and `goto_convert` does not re-lower it when the library body is loaded into a verification run — producing the nil-operand block.

The fix is most likely in the goto-conversion / clib-load pipeline: ensure structured-CF code kinds from loaded C++ library bodies are lowered (or are never serialized un-lowered into the clib).

## Why it matters

This is the sole blocker for compiling the C++ exception operational model into the clib, and more generally it means **no C++ operational model can be added to the embedded library** on current `master` without breaking all verification. Worth fixing before the OM stage of the #5126 split proceeds.

Related: #5126 (exception rework, being split), #5276 (PR-A, landed), the irep2 V.4.x series (#5266, #5277, #5282, #5284).

lucasccordeiro:
## Problem

Compiling any C++ source into ESBMC's embedded clib corrupted `goto_convert` with `ERROR: goto_convert: non-code operand in this block`, breaking verification of *every* program — even plain C with no C++ content. This blocked adding C++ operational models (e.g. a C++ exception OM) to the clib. Fixes #5298.

## Root cause

Registering both the C and C++ frontends in `c2goto` causes `language_uit::typecheck` to run `clang_c_adjust::adjust_for` twice over the shared library context. `adjust_for` wraps a `for` loop in an implicit block to hoist the init declaration:

```
for(a;b;c) d;   ->   { a; for(;b;c) d; }
```

After the first pass the loop's init operand (`op0`) is left nil. The second pass re-ran the wrap, moving the now-nil init into a fresh block as a stray non-code operand — which `goto_convert` rejects.

## Fix

- **`clang_c_adjust_code.cpp`** — make `adjust_for` idempotent: skip the implicit-block wrap when the loop init is nil (nothing to hoist). A freshly converted loop never has a nil init, so this fires only on the re-adjustment path and skips no real hoisting.
- **`c2goto.cpp`** — register the C++ frontend (`LANGAPI_MODE_CLANG_CPP`) in c2goto's mode table.
- **`c2goto/CMakeLists.txt`** — glob and compile `library/cpp/*.cpp` into the clib, with `CONFIGURE_DEPENDS` and a build-time `DEPENDS` entry.
- **`library/cpp/probe.cpp`** — minimal C++ OM probe (loop + range-for) to exercise the pipeline.

## Tests

- `regression/esbmc/github_5298` — a plain C for-loop still lowers and verifies with a C++ OM present.
- `regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om` — a C++ OM body in the clib is usable and verified.
- `regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail` — negative variant asserting a false property over the OM result, proving the body is actually verified (not silently skipped).

All three pass. The broad C++ and core C regression suites show no regressions attributable to this change (the only C++ failures require `--bitwuzla`, which isn't built into the local ESBMC).

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/main.cpp
@@ -1,0 +1,13 @@
+// Regression for esbmc/esbmc#5298: a C++ operational model compiled into the
+// embedded clib must be usable from a verified program. Before the fix, merely
+// having a C++ source in the clib corrupted goto_convert ("non-code operand in
+// this block") because the C++ frontend re-adjusted the C library functions.
+extern "C" int __esbmc_probe_sum(int n);
+extern "C" int __esbmc_probe_range();
+
+int main()
+{
+  __ESBMC_assert(__esbmc_probe_sum(3) == 3, "0+1+2 == 3");
+  __ESBMC_assert(__esbmc_probe_range() == 6, "1+2+3 == 6");
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 5
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/main.cpp
@@ -1,0 +1,10 @@
+// Negative variant of cpp_clib_om (esbmc/esbmc#5298): proves the C++ clib OM
+// body is actually verified (not silently skipped) by asserting a false
+// property over its result.
+extern "C" int __esbmc_probe_sum(int n);
+
+int main()
+{
+  __ESBMC_assert(__esbmc_probe_sum(3) == 999, "intentionally false");
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 5
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_5298/main.c
+++ b/regression/esbmc/github_5298/main.c
@@ -1,0 +1,13 @@
+// Regression for esbmc/esbmc#5298: with a C++ operational model present in the
+// embedded clib, lowering a C for-loop must still succeed. The bug made
+// goto_convert abort with "non-code operand in this block" because the C++
+// frontend re-adjusted the C library functions, and clang_c_adjust::adjust_for
+// was not idempotent (it re-wrapped the for-loop, leaving a stray nil operand).
+int main()
+{
+  int s = 0;
+  for (int i = 0; i < 3; i++)
+    s += i;
+  __ESBMC_assert(s == 3, "0+1+2 == 3");
+  return 0;
+}
--- a/regression/esbmc/github_5298/test.desc
+++ b/regression/esbmc/github_5298/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--unwind 5
+^VERIFICATION SUCCESSFUL$
--- a/src/c2goto/CMakeLists.txt
+++ b/src/c2goto/CMakeLists.txt
@@ -14,7 +14,7 @@ target_include_directories(c2goto
     PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr
     )
 target_link_libraries(c2goto
-  clangcfrontend langapi c2gotoheaders
+  clangcfrontend clangcppfrontend langapi c2gotoheaders
   filesystem ${Boost_LIBRARIES} ${OS_INCLUDE_LIBS})
 set_target_properties(c2goto PROPERTIES
   PRIVATE_HEADER "${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/libc.h;${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/libm.h;${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/headers/libc_hdr.h")
@@ -149,6 +149,11 @@ function(mangle_clib output)
     set(c2goto_solidity_files "")
   endif()
 
+  # Extract C++ operational models (esbmc/esbmc#5298). These are compiled by
+  # the C++ frontend inside c2goto and serialized into the embedded clib.
+  file(GLOB_RECURSE c2goto_cpp_files CONFIGURE_DEPENDS
+    "${CMAKE_CURRENT_SOURCE_DIR}/library/cpp/*.cpp")
+
   # For each config, generate clib*.c and clib*.goto file in the current binary directory
   foreach(in_f ${ARGN})
     set(in_f "${${in_f}}")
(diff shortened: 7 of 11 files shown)


mikhailramalho:
Weird, why do we run clang_c_adjust::adjust_for twice?

mikhailramalho:
I verified this end-to-end (built the branch, reproduced #5298, ran the new tests, and confirmed it unblocks compiling a real C++ exception OM into the clib — all green, including the `_fail` negative). The diagnosis is correct and the fix works. A couple of observations on the root cause and a possible deeper fix worth considering.

### The "runs adjust twice" mechanism checks out

Confirmed against the code:
- `language_uit::typecheck()` loops over **every** registered frontend in `langmap` and calls each one's `typecheck(context, "")` (`language_ui.cpp:95`).
- `langmap` is populated **per input-file extension** (`language_ui.cpp:65-68`), so this only bites when both C and C++ frontends are registered — i.e. exactly when `c2goto` compiles a `.cpp` alongside the C library (this PR's change). Normal single-language verification is unaffected, which is why this was latent until now.
- Each `clang_*_languaget::typecheck` runs `adjuster.adjust()`, and `adjust()` walks the **entire shared context** via `context.Foreach_operand_in_order` (`clang_c_adjust_expr.cpp:23`) — not just the symbols that frontend converted. `clang_cpp_adjust` inherits `clang_c_adjust::adjust_for`, so a `.c` library's `for`-loop gets `adjust_for` applied twice.
- `adjust_for` is the **only** structurally-mutating `adjust_*` method (it hoists the init: `for(a;b;c)d` → `{a; for(;b;c)d}`, nilling `op0`). `adjust_while`/`adjust_switch`/etc. only do idempotent things like bool-casting the condition. So `adjust_for` is uniquely non-idempotent, and the nil-init guard fully closes the crash. Good, minimal fix.

Suggestion: the comment could note that `adjust_for` is the *only* structurally-mutating adjust method (hence the only one needing the guard) — it preempts the natural "do the siblings need this too?" review question.

### Deeper root cause: `adjust()` re-walks the whole shared context

The idempotency guard is correct, but it treats the symptom. The underlying issue is that each frontend's `adjust()` iterates **all** symbols in the shared context, including ones a previous frontend already converted *and adjusted*. Today only `adjust_for` crashes on re-application; every other transform is silently re-applied (benign because they're idempotent — but it's a latent hazard the moment any future `adjust_*` becomes non-idempotent).

A more root-cause fix would be to convert + adjust into a **fresh** context, then merge into the shared one, so `adjust()` only ever sees this frontend's own freshly-converted symbols:

```cpp
bool clang_c_languaget::typecheck(contextt &context, const std::string &module)
{
  set_language_version();
  contextt new_context;
  clang_c_convertert converter(new_context, AST, "C");
  if (converter.convert())
    return true;

  clang_c_adjust adjuster(new_context);
  if (adjuster.adjust())
    return true;

  return c_link(context, new_context, module);   // merge only this frontend's adjusted symbols
}
```

`c_link` already exists as the context-merge primitive (and `add_cprover_library` uses the new-context-then-merge pattern), so this is idiomatic. The catch is blast radius and the merge semantics: `convert()` does ~6 context lookups, and a C++ OM body references C symbols (`__VERIFIER_nondet_*`, `malloc`, …) that, under a fresh context, would be unresolved until the `c_link` merge reconciles them. That changes link/conflict behavior across the shared all-frontend pipeline (every language, every verification run), so it deserves its own PR with full C/C++/Python regression coverage rather than riding in this clib fix.

### Recommendation

Land this as-is for #5298 — it's correct, minimal, and unblocks C++ OMs (verified it lets the exception OM compile into the clib cleanly). Track the fresh-context/scoped-`adjust()` refactor separately as the architectural follow-up.

Minor, likely unrelated to the core fix (possibly from the rebase): the diff also carries an `ENABLE_LD_FRONTEND` `add_subdirectory`, the ARM64 Solidity-stub change, and a Python C-model allowlist generator. Worth a quick confirm those are intended here.

lucasccordeiro:
Thanks for the thorough end-to-end verification, @mikhailramalho — much appreciated. Addressed your points:

**1. Comment on `adjust_for` being the only structural transform.** Done in a348620cc9. The guard now explicitly notes that `adjust_for` is the sole structurally-mutating `adjust_*` method (the siblings `adjust_while`/`adjust_switch`/`adjust_ifthenelse`/`adjust_assign` only insert idempotent typecasts), so the nil-init fence is needed only here — preempting the "do the siblings need this too?" question.

**2. Fresh-context / scoped-`adjust()` refactor.** Agreed this is the real root cause and worth a dedicated PR with full C/C++/Python regression coverage, given it changes link/conflict semantics across the shared all-frontend pipeline. Tracked as #5309 with your diagnosis and proposed `c_link`-merge sketch credited.

**3. The `ENABLE_LD_FRONTEND` `add_subdirectory`, ARM64 Solidity-stub, and Python C-model allowlist changes.** Those were not part of this PR — they were master commits showing through a stale merge-base at the time you reviewed. The branch has since been rebased onto current master, and the PR diff now contains only the clib/`adjust_for` changes (plus the `clang_c_convert` re-declaration fix below). Confirmed clean: `git diff $(git merge-base origin/master HEAD)..HEAD` lists only the 11 intended files.

One addition since your review: commit 028e46ae38 (`[clang-c-frontend] Keep defined function's type on re-declaration`). Once both frontends share the context, the C++ frontend re-converts the nameless `__VERIFIER_nondet_memory` prototype after the named C definition exists, stripping the parameter identifiers symex binds call arguments to — which havoc'd the clib `__VERIFIER_nondet_memory` and diverged k-induction on `nondet_verifier1-3`. The fix skips the type overwrite for a bodyless re-declaration when the existing symbol already has a body. It's part of the same #5298 root cause (multiple frontends sharing one context).

PR #5305, opened 2026-06-11
[witness 2.1] add C expression lexer for witness constraints

XLiZHI:
Introduces a lightweight clang C expression lexer (clang_c_lexert) that parses witness waypoint constraints at the clang_c_language level, before symbolic execution begins. This allows function_return waypoints with ext_c_expression or acsl_expression format (e.g. \result < 0) to be converted to IRep2 expressions and injected as assume constraints during symex.


src/clang-c-frontend/clang_c_language.cpp
```diff
@@ -338,7 +339,22 @@ bool clang_c_languaget::parse(const std::string &path)
...
+        continue;
+      expr2tc e = lexer.parse_expr(wp.value);
+      if (e)
+      {
+        wp.parsed_cond.expr = e;
+        wp.parsed_cond.valid = true;
+      }
+      else
+        log_warning(
+          "witness: could not parse constraint '{}', skipping", wp.value);
```
lucasccordeiro: Silent skip is a soundness risk for violation-witness validation: dropping the function_return constraint relaxes the path condition and can let the target be reached spuriously. Prefer surfacing this as an unconfirmed/failed validation rather than `log_warning` + continue.

lucasccordeiro (commented):
Nice cleanup replacing the regex/source-injection path with a real lexer. A few points for improvement (mostly soundness-adjacent):

- **Silent skip weakens validation.** Constraints that don't parse (or use unsupported shapes like `0 == \result`, arithmetic, `\at`) are dropped with a warning, so the function_return waypoint becomes a no-op. For a *violation*-witness validator that makes it easier to reach the target → risk of confirming a witness that shouldn't validate. Consider failing the validation (or marking it unconfirmed) rather than silently continuing.
- **Cross-phase coupling via static cache.** `parsed_cond` is written into the static `wp_cache` in the frontend and read back at symex time only because both call `get_waypoints(same_path)`. Worth a comment documenting this invariant; ideally assert the cache identity.
- Only one regression test added; repo convention asks for a passing+failing pair. A malformed-constraint test exercising the null-return path would lock down the lexer.
- Unrelated whitespace churn in `irep2_expr.h` — please drop from this PR.

src/goto-symex/symex_assign.cpp
```diff
@@ -528,6 +528,20 @@ void goto_symext::symex_assign(
   guard2tc g(guard); // NOT the state guard!
   symex_assign_rec(lhs, original_lhs, rhs, expr2tc(), g, hidden_ssa);
 
+  if (validate_witness && is_symbol2t(original_lhs))
+  {
+    const std::string nm = to_symbol2t(original_lhs).thename.as_string();
+    if (nm.find("return_value$_") != std::string::npos)
```
lucasccordeiro: `find("return_value$_") != npos` is a substring match anywhere in the name — can misfire on a user symbol containing that text. Prefer a stricter check (prefix / `has_prefix`). Also note this runs on every assignment when `validate_witness` is on.

src/goto-symex/symex_main.cpp
```diff
@@ -225,6 +226,60 @@ goto_symext::symex_resultt goto_symext::get_symex_result()
...
+static void substitute_result(expr2tc &e, const expr2tc &ret_val)
+{
+  if (is_symbol2t(e) && to_symbol2t(e).thename == "\\result")
+  {
+    e = ret_val;
+    return;
+  }
+  if (is_constant_int2t(e))
+  {
+    e = from_integer(to_constant_int2t(e).value, ret_val->type);
```
lucasccordeiro, line 245: Retyping every integer constant to `ret_val->type` silently changes signedness: e.g. `\result < 0` against an unsigned return makes `0` unsigned, so the comparison is trivially false. Worth handling (or at least warning on) signed/unsigned mismatch.

src/util/yaml_parser.cpp
```diff
@@ -106,9 +106,10 @@ struct
   waypoint target;
   bool has_target = false;
 } wp_cache;
+
 } // namespace
 
-std::vector<waypoint> yaml_parser::get_waypoints(const std::string &path)
+std::vector<waypoint> &yaml_parser::get_waypoints(const std::string &path)
```
lucasccordeiro, line 112: Returning a mutable reference to the static cache is what makes the frontend-parsed `parsed_cond` survive to symex. Please add a comment documenting that the returned vector is the shared cache and is mutated in place — it's a non-obvious lifetime contract.

src/goto-symex/symex_assign.cpp
XLiZHI: symbol2t::thename stores the complete USR (Unified Symbol Resolution) of the Clang AST, in the format of: `c:@F@main::$tmp::return_value$_compute$2`. The string starts with "c:", and "return_value$_" appears in the middle, not as a prefix.

To solve this problem, it was changed to `$tmp::return_value$_`, which is more restrictive than the original return_value$_, and $tmp is an exclusive scope for internal synthetic temporary variables of ESBMC.

src/clang-c-frontend/clang_c_language.cpp
XLiZHI: Improved as suggested.

src/goto-symex/symex_main.cpp
XLiZHI: Added a warning as suggested.

src/util/yaml_parser.cpp
XLiZHI: Done.

lucasccordeiro (approved):

PR #5308, opened 2026-06-11
[ir-ieee] add sqrt encoding under --ir-ieee

ShemaAlosaimi:
## Summary

This PR adds `--ir-ieee` support for `ieee_sqrt` in the integer/real SMT encoding path.

Previously, `ieee_sqrt` did not have a dedicated integer/real encoding implementation. This PR adds a guarded real-arithmetic encoding for square root when `--ir-ieee` is enabled, applies the existing IEEE-754 rounding-mode enclosure helpers, and adds interval lifting so sqrt expressions can participate in compositional real-interval propagation.

For negative operands, IEEE-754 returns NaN. Since full NaN semantics are not yet modelled in the `--ir-ieee` integer/real encoding, this PR uses an unconstrained real fallback for the negative-operand case. This avoids making the SMT formula UNSAT while keeping full NaN modelling as future work.

## Motivation

The `--ir-ieee` encoding already has rounding-aware real/interval support for arithmetic operations such as addition, subtraction, multiplication, and division. However, `ieee_sqrt` was still missing a dedicated integer/real encoding path.

A direct real-arithmetic encoding of sqrt would normally use:

`result >= 0` and `result * result == operand`

This is valid only when `operand >= 0`. If asserted unconditionally, the formula becomes UNSAT for paths where `operand < 0`, because no real number satisfies `result * result == operand` for a negative operand. That can incorrectly eliminate feasible program paths and lead to wrong verification results.

This PR guards the quadratic sqrt constraints with `operand >= 0` and keeps the negative-operand path satisfiable.

## Changes

This PR:

* Adds an `int_encoding` implementation for `expr2t::ieee_sqrt_id`.
* Introduces a fresh real symbol for the non-negative sqrt case.
* Guards the exact real sqrt constraints with `operand >= 0`:

  * `operand >= 0 -> sqrt_pos >= 0`
  * `operand >= 0 -> sqrt_pos * sqrt_pos == operand`
* Applies the existing mode-specific IEEE-754 enclosure helpers for:

  * RNE
  * RNA
  * RUP
  * RDN
  * RTZ
* Adds interval lifting for `ieee_sqrt` under `--ir-ieee` for single and double precision.
* Stores sqrt interval bounds in `ir_ra_interval_map` so chained expressions can propagate interval information through sqrt.
* Handles `operand < 0` with an unconstrained real fallback to avoid unsound path elimination.
* Leaves the existing non-integer-encoding floating-point API path unchanged.

For non-negative operands, sqrt is encoded using the guarded quadratic real constraint and rounded using the selected rounding-mode enclosure.

For negative operands, the result is represented by an unconstrained real fallback. This is intentionally not full IEEE NaN modelling.

## Regression Test

Four new regression tests were added under `regression/ir-ra/`:

* `ra-sqrt-rne-basic`

  * Checks the basic `sqrt(4.0) == 2.0` case under `--ir-ieee`.
  * Expected result: `VERIFICATION SUCCESSFUL`.

* `ra-sqrt-interval-lift`

  * Checks that interval lifting for sqrt introduces the expected real symbols.
  * The test uses `x` in `[1, 4]` and checks a clearly falsifiable assertion on `sqrt(x)`.
  * Expected result: `VERIFICATION FAILED`.

* `ra-sqrt-negative-operand-sound`

  * Checks that a definitely negative sqrt operand does not make the SMT formula UNSAT.
  * Expected result: `VERIFICATION FAILED`.

* `ra-sqrt-nan-comparison`

  * Checks that negative-operand sqrt comparisons are not incorrectly verified due to the result being pinned near zero.
  * Expected result: `VERIFICATION FAILED`.

## Notes

This PR does not implement full IEEE-754 NaN semantics.

In IEEE-754, `sqrt(x < 0)` produces NaN, and NaN has special comparison behaviour. The current `--ir-ieee` integer/real encoding does not yet model NaN non-reflexivity, NaN ordering rules, or full NaN propagation.

For this reason, the negative-operand case is represented using an unconstrained real fallback. This keeps the SMT formula satisfiable and avoids eliminating feasible paths, but full NaN support remains future work.

This PR is focused on adding `ieee_sqrt` support to the `--ir-ieee` integer/real encoding path without changing the existing non-`--ir-ieee` floating-point API path.

### DIFF
--- a/src/solvers/smt/smt_conv.cpp
+++ b/src/solvers/smt/smt_conv.cpp
@@ -1098,10 +1098,134 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
   case expr2t::ieee_sqrt_id:
   {
     assert(is_floatbv_type(expr));
-    // TODO: no integer mode implementation
-    a = fp_api->mk_smt_fpbv_sqrt(
-      convert_ast(to_ieee_sqrt2t(expr).value),
-      convert_rounding_mode(to_ieee_sqrt2t(expr).rounding_mode));
+    if (int_encoding)
+    {
+      smt_astt operand = convert_ast(to_ieee_sqrt2t(expr).value);
+      const floatbv_type2t &fbv_type = to_floatbv_type(expr->type);
+      const expr2tc &rounding_mode = to_ieee_sqrt2t(expr).rounding_mode;
+
+      // Two fresh reals are introduced:
+      //
+      //   sqrt_pos  (tagged "ra_sqrt::")   — for operand >= 0.
+      //     Pinned by the guarded quadratic:
+      //       operand >= 0  →  sqrt_pos >= 0  ∧  sqrt_pos² = operand
+      //     The enclosure is applied to this symbol only.
+      //
+      //   sqrt_nan  (tagged "ra_sqrt_nan::")  — for operand < 0.
+      //     Completely unconstrained.  Models the IEEE 754 NaN result
+      //     without making the formula inconsistent.
+      //
+      // The publicly returned result is  ite(operand >= 0, sqrt_pos, sqrt_nan).
+      // This ensures that for negative operands the enclosure constraints on
+      // sqrt_pos do not propagate to the observable result.
+      //
+      // Limitation: NaN non-reflexivity (NaN != NaN in IEEE 754) is not
+      // modelled.  Under --ir-ieee, sqrt(x < 0) returns an unconstrained real,
+      // not a NaN sentinel.
+      smt_sortt rs = mk_real_sort();
+      smt_astt zero = mk_smt_real("0.0");
+      smt_astt op_nonneg = mk_le(zero, operand);
+
+      smt_astt sqrt_pos = mk_fresh(rs, "ra_sqrt::", nullptr);
+      // operand >= 0 → sqrt_pos >= 0
+      assert_ast(mk_or(mk_not(op_nonneg), mk_le(zero, sqrt_pos)));
+      // operand >= 0 → sqrt_pos² = operand
+      assert_ast(
+        mk_or(mk_not(op_nonneg), mk_eq(mk_mul(sqrt_pos, sqrt_pos), operand)));
+
+      // Unconstrained result used when operand < 0.
+      smt_astt sqrt_nan = mk_fresh(rs, "ra_sqrt_nan::", nullptr);
+
+      // Interval-lifted enclosure for ieee_sqrt (--ir-ieee only).
+      // sqrt is monotone increasing on [0, ∞), so the exact hull is:
+      //   lo_r = sqrt(iv.lo),  hi_r = sqrt(iv.hi)
+      // Each bound is a fresh real pinned by the same quadratic axiom.
+      // Mode-dispatch follows the same five-case pattern as add/sub/mul/div.
+      bool interval_lifted = false;
+      if (
+        options.get_bool_option("ir-ieee") &&
+        (smt_fp_rounding_utils::is_nearest_rounding_mode(rounding_mode) ||
+         smt_fp_rounding_utils::is_round_to_away(rounding_mode) ||
+         smt_fp_rounding_utils::is_round_to_plus_inf(rounding_mode) ||
+         smt_fp_rounding_utils::is_round_to_minus_inf(rounding_mode) ||
+         smt_fp_rounding_utils::is_round_to_zero(rounding_mode)))
+      {
+        const auto double_spec = ieee_float_spect::double_precision();
+        const auto single_spec = ieee_float_spect::single_precision();
+        if (
+          (fbv_type.exponent == double_spec.e &&
+           fbv_type.fraction == double_spec.f) ||
+          (fbv_type.exponent == single_spec.e &&
+           fbv_type.fraction == single_spec.f))
+        {
+          auto get_iv = [this](smt_astt t) -> ra_interval_t {
+            auto it = ir_ra_interval_map.find(t);
+            return it != ir_ra_interval_map.end() ? it->second
+                                                  : ra_interval_t{t, t};
+          };
+          ra_interval_t iv = get_iv(operand);
+
+          // Fresh reals for sqrt of interval bounds.
+          // Clamp bounds to zero: the RNE/RNA enclosures from prior operations
+          // can make iv.lo slightly negative (by eps_abs) even when the float
+          // operand is always non-negative.  sqrt is only defined for x >= 0,
+          // so we tighten to max(0, iv.lo) and max(0, iv.hi).
+          smt_astt iv_lo_pos = mk_ite(mk_lt(iv.lo, zero), zero, iv.lo);
+          smt_astt iv_hi_pos = mk_ite(mk_lt(iv.hi, zero), zero, iv.hi);
+
+          smt_astt lo_r = mk_fresh(rs, "ra_sqrt_lo::", nullptr);
+          smt_astt hi_r = mk_fresh(rs, "ra_sqrt_hi::", nullptr);
+          assert_ast(mk_le(zero, lo_r));
+          assert_ast(mk_eq(mk_mul(lo_r, lo_r), iv_lo_pos));
+          assert_ast(mk_le(zero, hi_r));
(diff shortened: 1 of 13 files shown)


src/solvers/smt/smt_conv.cpp
```diff
@@ -1098,10 +1098,134 @@ smt_astt smt_convt::convert_ast(const expr2tc &expr)
   case expr2t::ieee_sqrt_id:
   {
     assert(is_floatbv_type(expr));
-    // TODO: no integer mode implementation
-    a = fp_api->mk_smt_fpbv_sqrt(
-      convert_ast(to_ieee_sqrt2t(expr).value),
-      convert_rounding_mode(to_ieee_sqrt2t(expr).rounding_mode));
+    if (int_encoding)
+    {
+      smt_astt operand = convert_ast(to_ieee_sqrt2t(expr).value);
```
lucasccordeiro, line 1103: Shall we move this implementation to https://github.com/esbmc/esbmc/blob/master/src/solvers/smt/fp/fp_conv.cpp?

lucasccordeiro (changes requested):
## Review findings

The core `--ir-ieee` interval encoding and the ITE NaN split for negative operands are correct. The quadratic axioms are properly guarded by `op_nonneg`. No soundness bug found for the standard single/double path. Four issues below, two of which need addressing before merge.

---

### 1. `--ir` (non-ieee int-encoding) path is untested

`int_encoding` is `true` for both `--ir` and `--ir-ieee`. When `--ir` is used without `--ir-ieee`, the `interval_lifted` block is skipped and the fallback runs:

```cpp
smt_astt pos_result =
  apply_ieee754_semantics(sqrt_pos, fbv_type, nullptr, rounding_mode);
a = mk_ite(op_nonneg, pos_result, sqrt_nan);
```

`apply_ieee754_semantics` in non-`ir-ieee` mode returns a composite ITE expression (overflow/underflow/subnormal cases). **None of the four new tests cover `--ir sqrt`** — all use `--ir-ieee`. If this fallback path regresses, it goes undetected.

**Fix:** Add a `CORE` test with `--ir --z3` (e.g. `sqrt(4.0) == 2.0` → `VERIFICATION SUCCESSFUL`).

---

### 2. `ra-sqrt-interval-lift` is `THOROUGH` (Linux-only)

```
regression/ir-ra/ra-sqrt-interval-lift/test.desc:1: THOROUGH
```

The only test that exercises the interval-lifted path and verifies that `ra_sqrt_lo::` / `ra_sqrt_hi::` symbols are emitted will not run on macOS CI. The main proof that the interval machinery actually fires is skipped on non-Linux platforms.

**Fix:** Add a `CORE` test that exercises the interval-lifted path without relying on `--smt-formula-too` for symbol presence, so it runs everywhere.

---

### 3. (Minor) Unconditional `lo_r`/`hi_r` constraints for definitely-negative operands

When the operand is definitely negative (e.g. `__ESBMC_assume(x < -1.0)`), the clamping gives `iv_lo_pos = iv_hi_pos = 0`, so `lo_r = hi_r = 0`. The enclosure helper then asserts `ra_lo <= sqrt_pos <= ra_hi` with both bounds near zero — these nonlinear constraints are always added even though `sqrt_pos` does not appear in the result (the ITE picks `sqrt_nan`). For programs with complex definitely-negative sqrt paths this adds unnecessary polynomial constraints to the SMT formula.

---

### 4. (Minor) Overlapping negative-operand tests

`ra-sqrt-nan-comparison` and `ra-sqrt-negative-operand-sound` both check that a definitely-negative operand yields an unconstrained (satisfiable) result — one tests `s < 100.0`, the other `s > 0.0`. They exercise the same ITE path and one is redundant.

ShemaAlosaimi:
> ## Review findings
> 
> The core `--ir-ieee` interval encoding and the ITE NaN split for negative operands are correct. The quadratic axioms are properly guarded by `op_nonneg`. No soundness bug found for the standard single/double path. Four issues below, two of which need addressing before merge.
> ### 1. `--ir` (non-ieee int-encoding) path is untested
> 
> `int_encoding` is `true` for both `--ir` and `--ir-ieee`. When `--ir` is used without `--ir-ieee`, the `interval_lifted` block is skipped and the fallback runs:
> 
> ```c++
> smt_astt pos_result =
>   apply_ieee754_semantics(sqrt_pos, fbv_type, nullptr, rounding_mode);
> a = mk_ite(op_nonneg, pos_result, sqrt_nan);
> ```
> 
> `apply_ieee754_semantics` in non-`ir-ieee` mode returns a composite ITE expression (overflow/underflow/subnormal cases). **None of the four new tests cover `--ir sqrt`** — all use `--ir-ieee`. If this fallback path regresses, it goes undetected.
> 
> **Fix:** Add a `CORE` test with `--ir --z3` (e.g. `sqrt(4.0) == 2.0` → `VERIFICATION SUCCESSFUL`).
> ### 2. `ra-sqrt-interval-lift` is `THOROUGH` (Linux-only)
> 
> ```
> regression/ir-ra/ra-sqrt-interval-lift/test.desc:1: THOROUGH
> ```
> 
> The only test that exercises the interval-lifted path and verifies that `ra_sqrt_lo::` / `ra_sqrt_hi::` symbols are emitted will not run on macOS CI. The main proof that the interval machinery actually fires is skipped on non-Linux platforms.
> 
> **Fix:** Add a `CORE` test that exercises the interval-lifted path without relying on `--smt-formula-too` for symbol presence, so it runs everywhere.
> ### 3. (Minor) Unconditional `lo_r`/`hi_r` constraints for definitely-negative operands
> 
> When the operand is definitely negative (e.g. `__ESBMC_assume(x < -1.0)`), the clamping gives `iv_lo_pos = iv_hi_pos = 0`, so `lo_r = hi_r = 0`. The enclosure helper then asserts `ra_lo <= sqrt_pos <= ra_hi` with both bounds near zero — these nonlinear constraints are always added even though `sqrt_pos` does not appear in the result (the ITE picks `sqrt_nan`). For programs with complex definitely-negative sqrt paths this adds unnecessary polynomial constraints to the SMT formula.
> ### 4. (Minor) Overlapping negative-operand tests
> 
> `ra-sqrt-nan-comparison` and `ra-sqrt-negative-operand-sound` both check that a definitely-negative operand yields an unconstrained (satisfiable) result — one tests `s < 100.0`, the other `s > 0.0`. They exercise the same ITE path and one is redundant.

Thanks, @lucasccordeiro , for the review.

**Finding 1 — Fixed.**
Added `regression/ir-ra/ra-sqrt-ir-basic/`:

* `CORE`
* `--ir --z3`
* checks `sqrt(4.0) == 2.0`
* expected result: `VERIFICATION SUCCESSFUL`

This directly exercises the non-`--ir-ieee` `!interval_lifted` fallback path that was changed by this patch.

**Finding 2 — Fixed.**
Added `regression/ir-ra/ra-sqrt-interval-lift-core/`:

* `CORE`
* `--ir-ieee --z3`
* no `--smt-formula-too`
* checks `x in [1,4]` with `assert(sqrt(x) > 100.0)`
* expected result: `VERIFICATION FAILED`

The existing `THOROUGH` interval-lift test is kept for symbol-emission verification on Linux.

The updated `ir-ra` suite passes:

```text
ir-ra: 142/142 passed
```

**Finding 3 — Acknowledged, deferred.**
I agree that `lo_r`/`hi_r` constraints are still asserted even when `sqrt_pos` is discarded by the ITE for definitely-negative operands. This is a formula-size inefficiency rather than a soundness issue. I would prefer to leave it as follow-up work instead of adding more branching complexity in this PR.

**Finding 4 — Kept both tests intentionally.**
I kept both negative-operand tests because they cover two different regression modes:

* `ra-sqrt-negative-operand-sound` catches the UNSAT-masking case. If the quadratic constraint `s * s == operand` is asserted unconditionally, the formula becomes UNSAT for `operand < 0`, the path is discarded, and this test can incorrectly become `VERIFICATION SUCCESSFUL`.

* `ra-sqrt-nan-comparison` catches the enclosure-pinning case. Even with the guarded quadratic, if the observable result is still constrained by the enclosure instead of using the ITE split, the negative result can be pinned near zero, making the assertion incorrectly verify.

So I kept both tests to preserve coverage for both failure modes.

PR #5310, opened 2026-06-11
[clang-c-frontend] Scope adjust() to fresh per-TU context, merge via c_link

### LINKED ISSUE #5309: Scope clang_*_languaget::adjust() to freshly-converted symbols instead of re-walking the shared context
## Background

Follow-up from review of #5303 (fixes #5298).

When `c2goto` registers both the C and C++ frontends so that C++ operational models can be compiled into the embedded clib, `language_uit::typecheck()` runs **every** registered frontend's `adjust()` over the **shared** context. Because `clang_c_adjust::adjust()` walks the entire context via `context.Foreach_operand_in_order` (`clang_c_adjust_expr.cpp`), symbols a previous frontend already converted *and adjusted* get adjusted again.

#5303 fixed the only transform that actually crashed on re-application — `adjust_for`, the sole structurally-mutating `adjust_*` method (it hoists the loop init `for(a;b;c)d` → `{a; for(;b;c)d}`, nilling `op0`; a second pass moved the now-nil init into a fresh block as a stray non-code operand, aborting `goto_convert`). The fix makes `adjust_for` idempotent via a nil-init guard.

## Problem

That guard treats the symptom. The root cause is that `adjust()` re-walks the whole shared context rather than only this frontend's own freshly-converted symbols. Every other `adjust_*` is currently idempotent, so re-application is benign today — but it is a latent hazard the moment any future `adjust_*` becomes non-idempotent.

## Proposed direction

Convert + adjust into a **fresh** context, then merge into the shared one via the existing `c_link` primitive (the same new-context-then-merge pattern `add_cprover_library` already uses), so each frontend's `adjust()` only ever sees its own symbols:

```cpp
bool clang_c_languaget::typecheck(contextt &context, const std::string &module)
{
  set_language_version();
  contextt new_context;
  clang_c_convertert converter(new_context, AST, "C");
  if (converter.convert())
    return true;

  clang_c_adjust adjuster(new_context);
  if (adjuster.adjust())
    return true;

  return c_link(context, new_context, module); // merge only this frontend's adjusted symbols
}
```

## Risk / why it is a separate PR

Blast radius is large: `convert()` does context lookups, and a C++ OM body references C symbols (`__VERIFIER_nondet_*`, `malloc`, …) that, under a fresh context, are unresolved until the `c_link` merge reconciles them. This changes link/conflict semantics across the shared all-frontend pipeline (every language, every verification run), so it needs full C/C++/Python regression coverage on its own rather than riding in the clib fix.

Credit: diagnosis and proposed fix by @mikhailramalho in the #5303 review.

lucasccordeiro:
Fixes #5309. Stacked on #5303 (base branch `fix/5298-cpp-clib-operational-models`) — review/merge #5303 first.

## Problem
Once `c2goto` registers both the C and C++ frontends (#5303), `language_uit::typecheck()` calls every registered frontend's `typecheck()` on a **shared** context. `clang_c_languaget::typecheck` / `clang_cpp_languaget::typecheck` convert and adjust directly into that shared context, and `clang_c_adjust::adjust()` walks the **entire** context — so symbols already converted and adjusted by an earlier pass get re-adjusted, corrupting them. #5303 patched the symptom (made `adjust_for` idempotent); this is the structural fix.

## Change
Each `typecheck` now converts + adjusts the translation unit in an isolated `contextt new_context`, then merges the finished symbols into the shared context with `c_link(context, new_context, module)` — the same convert-into-fresh-context-then-link pattern `add_cprover_library` / `add_bundled_library_sources` already use. `adjust()` only ever sees freshly-converted symbols.

### Follow-on fix in `c_link`
Isolating per TU routes the libc/libm clib build through `c_link` once per file. Each TU pulls the intrinsics header, so a tentative definition (`_Bool __ESBMC_alloc[1];` in `builtin_libs.c`, `int __ESBMC_return_value;` in the intrinsics header) now meets its `extern` re-declaration at link time. The old `c_linkt::duplicate_symbol` variable path aborted here ("never seen in tests") — **nondeterministically**, depending on unstable hash-table symbol ordering. It is aligned with `contextt::move_symbol_to_context`: adopt the incoming definition only when `in_context` is `extern` and the incoming symbol is not, otherwise keep the symbol already in context; never abort on variables.

## Verification
- Embedded clib build: previously aborted nondeterministically; now builds cleanly and deterministically (25/25 repeated `clib64.goto` generations), full `esbmc` builds.
- Regression (local, Bitwuzla-only build): core C `esbmc/`, C++ `esbmc-cpp/cpp/`, and a Python sample — **zero new failures** vs a rebuilt baseline. The non-solver failures (`--ir` IEEE encoding, host C++ headers via `--no-abstracted-cpp-includes`, `--multi-property`) fail identically without this change and are environmental; CI (with Z3/Boolector) covers the solver-gated cases.
- The removed abort branch was demonstrably reachable (the clib build hit it), which discharges the dead-code obligation for the deletion.

### DIFF
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/main.cpp
@@ -1,0 +1,13 @@
+// Regression for esbmc/esbmc#5298: a C++ operational model compiled into the
+// embedded clib must be usable from a verified program. Before the fix, merely
+// having a C++ source in the clib corrupted goto_convert ("non-code operand in
+// this block") because the C++ frontend re-adjusted the C library functions.
+extern "C" int __esbmc_probe_sum(int n);
+extern "C" int __esbmc_probe_range();
+
+int main()
+{
+  __ESBMC_assert(__esbmc_probe_sum(3) == 3, "0+1+2 == 3");
+  __ESBMC_assert(__esbmc_probe_range() == 6, "1+2+3 == 6");
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 5
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/main.cpp
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/main.cpp
@@ -1,0 +1,10 @@
+// Negative variant of cpp_clib_om (esbmc/esbmc#5298): proves the C++ clib OM
+// body is actually verified (not silently skipped) by asserting a false
+// property over its result.
+extern "C" int __esbmc_probe_sum(int n);
+
+int main()
+{
+  __ESBMC_assert(__esbmc_probe_sum(3) == 999, "intentionally false");
+  return 0;
+}
--- a/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/test.desc
+++ b/regression/esbmc-cpp/OM_sanity_checks/cpp_clib_om_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 5
+^VERIFICATION FAILED$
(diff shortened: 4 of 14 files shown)


mikhailramalho:
We'll probably need the same changes for all backends, including clang cpp. Big changes; it needs some thorough testing.

XLiZHI (approved):
LGTM

lucasccordeiro:
This PR (100s): 

```
  correct:           23229
    correct true:    15322
    correct false:    7907
  incorrect:             5
    incorrect true:      0
    incorrect false:     5
  unknown:           13366
  Score:             38471 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27370454565.
```

Master (100s):

```
  correct:           23229
    correct true:    15319
    correct false:    7910
  incorrect:             6
    incorrect true:      0
    incorrect false:     6
  unknown:           13365
  Score:             38452 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27269035320/job/80536864374
```

mikhailramalho (commented):
lgtm

PR #5317, opened 2026-06-12
[symex] Filter function-pointer call targets by signature

### LINKED ISSUE #5296: [symex] Function-pointer call dispatches to wrong-arity targets and drops arguments → spurious VERIFICATION FAILED
## Summary

When a function pointer is read back from heap-allocated memory (e.g. a field of a `malloc`'d struct, assigned across function boundaries), ESBMC's value-set for that pointer is over-approximated to include **other address-taken functions, regardless of signature** — including functions with a **different arity**. The indirect-call dispatch (`goto_symext::get_function_list` / `symex_function_call_deref`) then lists those incompatible functions as call targets. When a 1-argument call is dispatched to a 2-argument target, the missing argument is silently **modelled as nondet** (`argument_assignments`, `src/goto-symex/symex_function.cpp:95`). The solver is then free to pick `NULL`/garbage for that nondet argument, which can trip an assertion inside the wrongly-dispatched callee and report a **spurious `VERIFICATION FAILED`**.

This was found while investigating the `aws_hash_table_create_harness` false alarm (#5287). It is **one of two independent bugs** in that benchmark; this issue tracks the function-pointer dispatch bug specifically. (The other is a pointer-equality inconsistency, tracked separately under #5287.)

## Smoking gun

ESBMC emits, for a **1-argument** hash-function-pointer call, a warning that it is calling the **2-argument** equals function and dropping the second parameter:

```
WARNING: function call to 'c:@F@eqnn': missing argument for parameter '...@F@eqnn@b'; modelled as nondet
```

## Minimal reproducer (~30 lines) — shows the wrong-arity dispatch

```c
#include <stdint.h>
#include <stdlib.h>
extern void abort(void);
extern unsigned long __VERIFIER_nondet_ulong(void);

uint64_t uhash(const void *a) { return (uint64_t)a; }                       // 1-arg
_Bool    eqnn(const void *a, const void *b) {                              // 2-arg
  __ESBMC_assert(b != NULL, "b dropped?");
  return a == b;
}

typedef uint64_t (*hfn)(const void *);
typedef _Bool    (*efn)(const void *, const void *);
struct state { hfn hf; efn ef; size_t size, mask; };
struct table { struct state *p; };

static void alloc_table(struct table *t) {        // assign p in a separate function
  size_t n = __VERIFIER_nondet_ulong();
  if (!(n && n <= 4 && !(n & (n - 1)))) abort();
  struct state *s = malloc(sizeof(struct state) + n * 32);
  s->size = n; s->mask = n - 1; t->p = s;
}

int main(void) {
  struct table t;
  alloc_table(&t);
  t.p->hf = uhash;                                // distinct functions, distinct arities
  t.p->ef = eqnn;
  void *k = (void *)(uintptr_t)__VERIFIER_nondet_ulong();
  if (!k) abort();
  uint64_t h = t.p->hf(k);                        // 1-ARG indirect call through hf
  __ESBMC_assert(h == (uint64_t)k, "hf must be uhash, not eqnn");
  return 0;
}
```

Run:
```
esbmc bugA2.c --64 --k-induction --max-inductive-step 3 \
  --no-pointer-check --no-bounds-check --interval-analysis --force-malloc-success --no-div-by-zero-check
```
Output includes `missing argument for parameter '...@F@eqnn@b'; modelled as nondet` — ESBMC lists the **2-argument** `eqnn` as a candidate for the **1-argument** `hf` call. (Here the wrong-target branch happens to be guarded out, so the run is SUCCESSFUL; the warning is the symptom.)

**Control:** if `hf` and `ef` are given the **same** signature, the warning disappears (`missing-arg warnings: 0`) — confirming arity is the trigger, not the heap read alone.

## Full reproducer that turns the over-approximation into a false alarm

`reduce_5287_min.c` (110 lines, creduce-minimised from `aws_hash_table_create_harness.i`) reports `VERIFICATION FAILED` + `Bug found (k = 1)` under the same flags. There the 1-arg `t->p->hf(k)` call (in `insert`) is dispatched to the 2-arg `uninterpreted_equals_assert_inputs_nonnull`, drops `b`, the solver picks `NULL`, and that callee's own `check(b != NULL)` fires `reach_error`. <details>
<summary><code>reduce_5287_min.c</code> — 110-line failure reproducer (click to expand)</summary>

```c
/* Ultra-minimal reproducer for ESBMC issue #5287 attempt */
#include <stdint.h>
#include <stddef.h>
#include <stdlib.h>

extern unsigned long __VERIFIER_nondet_ulong(void);
extern void abort(void);
extern void __assert_fail(const char *, const char *, unsigned int, const char *)
    __attribute__((__nothrow__, __leaf__)) __attribute__((__noreturn__));

void reach_error(void) { __assert_fail("0", "f.c", 9, "reach_error"); }
void assume_nonnull(_Bool c) { if (!c) abort(); }
void check(_Bool c) { if (!c) { reach_error(); abort(); } }

extern _Bool __CPROVER_uninterpreted_equals(const void *, const void *);
extern uint64_t __CPROVER_uninterpreted_hasher(const void *);

_Bool ueq(const void *a, const void *b) {
    return __CPROVER_uninterpreted_equals(a, b);
}

_Bool uninterpreted_equals_assert_inputs_nonnull(const void *a, const void *b) {
    check(a != NULL);
    check(b != NULL);
    return ueq(a, b);
}

uint64_t uhash(const void *a) {
    check(a != NULL);
    return __CPROVER_uninterpreted_hasher(a);
}

typedef _Bool (*eq_fn)(const void *, const void *);
typedef uint64_t (*hash_fn)(const void *);

struct entry { const void *key; uint64_t hc; };
struct state {
    hash_fn hf;
    eq_fn ef;
    size_t size, mask, count, maxload;
    /* struct entry slots[] */
};

static struct entry *sl(struct state *s) { return (struct entry *)(s+1); }

static int lookup(struct state *s, uint64_t hc, const void *k,
                  struct entry **out, size_t *pi) {
    struct entry *e = &sl(s)[hc & s->mask];
    if (!e->hc) { if (pi) *pi=0; *out=e; return 1; }
    if (e->hc==hc && s->ef(k, e->key)) { if (pi) *pi=0; *out=e; return 0; }
    size_t i=1;
    while(1) {
        size_t idx=(hc+i)&s->mask; e=&sl(s)[idx];
        if(!e->hc){if(pi)*pi=i;*out=e;return 1;}
        if(e->hc==hc&&s->ef(k,e->key)){if(pi)*pi=i;*out=e;return 0;}
        if(((idx-e->hc)&s->mask)<i){if(pi)*pi=i;*out=e;return 1;}
        i++;
    }
}

struct table { struct state *p; };

int insert(struct table *t, const void *k, const void **out) {
    uint64_t hc = t->p->hf(k); if(!hc) hc=1;
    struct entry *e; size_t pi;
    if (!lookup(t->p, hc, k, &e, &pi)) { if(out)*out=e->key; return 0; }
    e->key=k; e->hc=hc; t->p->count++;
    if(out)*out=e->key;
    return 0;
}

static uint64_t nd64(void) {
    return ((uint64_t)__VERIFIER_nondet_ulong()<<32)|
           ((uint64_t)__VERIFIER_nondet_ulong()&0xfffffffful);
}

static void alloc_table(struct table *t, size_t maxn) {
    size_t n=nd64();
    assume_nonnull(n<=maxn && n && !(n&(n-1)));
    assume_nonnull(n<=(size_t)-1/sizeof(struct entry));
    size_t b=n*sizeof(struct entry);
    assume_nonnull(b<=(size_t)-1-sizeof(struct state));
    struct state *s=malloc(sizeof(struct state)+b);
    s->size=n; s->mask=n-1; t->p=s;
}

static _Bool has_empty(const struct table *t) {
    const struct state *s=t->p;
    assume_nonnull(s->count>0);
    size_t i=nd64(); assume_nonnull(i<s->size);
    return sl((struct state*)s)[i].hc==0;
}

int main(void) {
    struct table t;
    alloc_table(&t, 4);
    assume_nonnull(t.p && t.p->size && t.p->mask==t.p->size-1);
    t.p->ef = uninterpreted_equals_assert_inputs_nonnull;
    t.p->hf = uhash;

    assume_nonnull(has_empty(&t));

    void *k = (void*)(uintptr_t)__VERIFIER_nondet_ulong();
    assume_nonnull(k!=NULL);

    const void *fk=NULL;
    insert(&t, k, &fk);
    check(ueq(fk, k));
    return 0;
}

```
</details>

## Root cause

`goto_symext::get_function_list` (`src/goto-symex/symex_function.cpp:411-457`) builds the candidate target list from the dereferenced function pointer (an if-then-else tree of symbols produced by the value-set during `dereference()`), **without any signature/arity filter**. `argument_assignments` (`src/goto-symex/symex_function.cpp:74-118`) then matches the call's actual arguments against the *target's* formal parameters and, when the target has more parameters than the call provides, models the surplus as nondet (`:95`). For function pointers read from imprecisely-modelled heap objects, the value-set returns unrelated address-taken functions, so wrong-arity targets reach `argument_assignments` and arguments are silently dropped.

CBMC's `remove_function_pointers` pass filters candidate functions to those whose **type matches** the function-pointer type; ESBMC's dispatch does not apply an equivalent filter.

## Suggested fix

Filter call-target candidates by signature compatibility before dispatch — either:
- in the value-set / `dereference()` of a function-pointer read (drop functions whose `code_type` is incompatible with the pointer's pointed-to type), or
- in `get_function_list` / `symex_function_call_deref`, skip any candidate whose parameter count / type does not match the call signature (and, defensively, make `argument_assignments`' "missing argument" path a hard error for non-variadic targets rather than silently nondet-filling).

## Environment
- ESBMC 8.3.0, aarch64 macOS, Bitwuzla 0.9.0.

Refs #5287.

lucasccordeiro:
## Problem

An indirect call through a function pointer read from imprecisely-modelled heap memory gets an over-approximated value-set that lists address-taken functions of **unrelated arity**. `symex_function_call_deref` built its candidate target list from that value-set with **no signature filter** (unlike CBMC's `remove_function_pointers`), so a 1-argument call could be dispatched to a 2-argument target. The surplus parameter is silently modelled as nondet (`argument_assignments`), the solver picks NULL/garbage, an assertion inside the wrongly-dispatched callee fires, and ESBMC reports a **spurious `VERIFICATION FAILED`** (`Fixes #5296`).

## Fix

Add a static `function_is_type_compatible(declared_type, candidate, ns)` predicate and filter the candidate list in `symex_function_call_deref`:

- **Prototyped, fixed-arity pointers** drop candidates of mismatched arity or incompatible parameter type (reusing the same number↔pointer interchange the call site already performs).
- **Unprototyped (`void (*)()`) and variadic pointers** still match any target — ESBMC cannot distinguish `void (*)()` from `void (*)(void)`, so these keep the existing permissive K&R dispatch (`function_deref_{2,3,4}`).
- If filtering would remove **every** candidate, the full list is kept, so a frontend type quirk on the real target can never silently skip a call.

Return types are intentionally not compared, to avoid pruning common `void*`/typed-pointer casts. The filter only tightens fully-prototyped pointers; unprototyped indirect calls are unchanged.

## Tests

- `regression/esbmc/github_5296` — heap-read 1-arg pointer whose value-set over-approximates to a 2-arg target; asserts SUCCESSFUL **and** that the `missing argument for parameter` warning is absent (negative-lookahead). Fails on the unfixed binary, passes with the fix.
- `regression/esbmc/github_5296_fail` — same setup but the correctly-dispatched 1-arg target asserts on a possibly-null key; FAILED is reachable only if the filter keeps the real target — guards against over-pruning.

Validated: both new tests, `function_deref_{0..4}`, 180 C++ virtual-dispatch tests, and ~30 other function-pointer tests all pass; no regressions.

### DIFF
--- a/regression/esbmc/github_5296/main.c
+++ b/regression/esbmc/github_5296/main.c
@@ -1,0 +1,68 @@
+// #5296: an indirect call through a function pointer read from a
+// symbolic-size heap object gets an over-approximated value-set that lists
+// address-taken functions of unrelated arity. Dispatching the 1-argument hash
+// call to the 2-argument equals function silently models the missing argument
+// as nondet ("missing argument for parameter ...; modelled as nondet").
+//
+// The pointer provably holds uhash, so the program is SUCCESSFUL either way;
+// the bug is the spurious wrong-arity candidate. This test passes only when
+// that candidate is filtered out (no missing-argument warning).
+#include <stdint.h>
+#include <stdlib.h>
+extern void abort(void);
+extern unsigned long __VERIFIER_nondet_ulong(void);
+
+uint64_t uhash(const void *a) // 1-argument
+{
+  return (uint64_t)a ? (uint64_t)a : 1ull;
+}
+
+_Bool eqnn(const void *a, const void *b) // 2-argument
+{
+  __ESBMC_assert(b != NULL, "b must not be a dropped (nondet) argument");
+  return a == b;
+}
+
+typedef uint64_t (*hfn)(const void *);
+typedef _Bool (*efn)(const void *, const void *);
+struct state
+{
+  hfn hf;
+  efn ef;
+  size_t n;
+};
+struct table
+{
+  struct state *p;
+};
+
+static void alloc_table(struct table *t)
+{
+  size_t n = __VERIFIER_nondet_ulong();
+  if (!(n && n <= 4))
+    abort();
+  struct state *s = malloc(sizeof(struct state) + n * 16); // symbolic-size heap
+  if (!s)
+    abort();
+  s->hf = uhash;
+  s->ef = eqnn;
+  s->n = n;
+  t->p = s;
+}
+
+static uint64_t do_hash(struct table *t, const void *k)
+{
+  return t->p->hf(k); // 1-argument indirect call through a heap-read pointer
+}
+
+int main(void)
+{
+  struct table t;
+  alloc_table(&t);
+  void *k = (void *)(uintptr_t)__VERIFIER_nondet_ulong();
+  if (!k)
+    abort();
+  uint64_t h = do_hash(&t, k);
+  __ESBMC_assert(h != 0, "uhash never returns 0");
+  return 0;
+}
--- a/regression/esbmc/github_5296/test.desc
+++ b/regression/esbmc/github_5296/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--64 --incremental-bmc --no-pointer-check --no-bounds-check --force-malloc-success
+\A(?![\s\S]*missing argument for parameter)[\s\S]*VERIFICATION SUCCESSFUL[\s\S]*\Z
--- a/regression/esbmc/github_5296_fail/main.c
+++ b/regression/esbmc/github_5296_fail/main.c
@@ -1,0 +1,62 @@
+// #5296 companion: the signature filter must keep the *correct* target.
+// Same over-approximated heap function pointer as github_5296, but the
+// legitimately-dispatched 1-argument uhash asserts its argument is non-null and
+// the key may be null. The assertion is reachable only if uhash is still a call
+// target after filtering, so this FAILED result guards against the filter
+// over-pruning and silently skipping the real call.
+#include <stdint.h>
+#include <stdlib.h>
+extern void abort(void);
+extern unsigned long __VERIFIER_nondet_ulong(void);
+
+uint64_t uhash(const void *a) // 1-argument
+{
+  __ESBMC_assert(a != NULL, "uhash argument must be non-null");
+  return (uint64_t)a ? (uint64_t)a : 1ull;
+}
+
+_Bool eqnn(const void *a, const void *b) // 2-argument
+{
+  return a == b;
+}
+
+typedef uint64_t (*hfn)(const void *);
+typedef _Bool (*efn)(const void *, const void *);
+struct state
+{
+  hfn hf;
+  efn ef;
+  size_t n;
+};
+struct table
+{
+  struct state *p;
+};
+
+static void alloc_table(struct table *t)
+{
+  size_t n = __VERIFIER_nondet_ulong();
+  if (!(n && n <= 4))
+    abort();
+  struct state *s = malloc(sizeof(struct state) + n * 16); // symbolic-size heap
+  if (!s)
+    abort();
+  s->hf = uhash;
+  s->ef = eqnn;
+  s->n = n;
+  t->p = s;
+}
+
+static uint64_t do_hash(struct table *t, const void *k)
+{
+  return t->p->hf(k); // 1-argument indirect call through a heap-read pointer
+}
+
+int main(void)
+{
+  struct table t;
+  alloc_table(&t);
+  void *k = (void *)(uintptr_t)__VERIFIER_nondet_ulong(); // may be null
+  uint64_t h = do_hash(&t, k);
+  return 0;
+}
--- a/regression/esbmc/github_5296_fail/test.desc
+++ b/regression/esbmc/github_5296_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--64 --incremental-bmc --no-pointer-check --no-bounds-check --force-malloc-success
+^VERIFICATION FAILED$
--- a/src/goto-symex/symex_function.cpp
+++ b/src/goto-symex/symex_function.cpp
@@ -412,6 +412,49 @@ void goto_symext::symex_function_call_code(const expr2tc &expr)
   cur_state->source.prog = &goto_function.body;
 }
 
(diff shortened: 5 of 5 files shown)


lucasccordeiro:
This PR (30s): 

```
  correct:           20713
    correct true:    13860
    correct false:    6853
  incorrect:             2
    incorrect true:      0
    incorrect false:     2
  unknown:           15885
  Score:             34541 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27408266020.
```

Master (30s):

```
  correct:           20728
    correct true:    13853
    correct false:    6875
  incorrect:             3
    incorrect true:      0
    incorrect false:     3
  unknown:           15869
  Score:             34533 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27235445216/job/80429911459
```

PR #5330, opened 2026-06-12
[irep2] Preserve sizeof(T) type on constants under --irep2-bodies

lucasccordeiro:
## Problem

Under `--irep2-bodies`, `malloc(sizeof(T))` allocated a `char` blob instead of a `struct T`. clang folds `sizeof(T)` to a bare `size_t` constant whose only record of the element type `T` is the `#c_sizeof_type` attribute. `migrate_expr` dropped that attribute on the IREP2 body round-trip, so `get_alloc_type` fell back to `char_type()` and the dynamic object degraded from `struct T` to `char[N]`.

The subsequent struct copy through the mistyped object then either read the wrong fields or tripped a spurious array-bounds violation — a false `VERIFICATION FAILED` that only appears with the flag on. GOTO proof (`--goto-functions-only | grep MALLOC`):

```
OFF: MALLOC(struct pthread_id_value { int id; ... void *value; }, 16)
ON:  MALLOC(signed char, 16)        # before this PR
```

This is a V.4.4 verdict-parity blocker (the misc-C cluster of esbmc/esbmc#4715).

## Fix

`constant_int2t` now carries a non-reflected `type2tc sizeof_type` field (with `excluded_field_bytes`), mirroring the V.4.1/V.4.5 non-reflected `location` pattern. The `fields` tuple is unchanged, so the field is excluded from `cmp`/`crc`/`hash` — plain integer constants stay byte-identical and constant sharing is unaffected. The migrate forward arm reads `#c_sizeof_type` into it; the back arm restores the attribute. A stale comment in `c_typecast.cpp` (whose manual save/restore is still load-bearing for the simplifier path) is clarified.

Off-flag this is a no-op: `migrate_expr` is only on the body path under `--irep2-bodies`, and the new field defaults to nil.

## Result

Fixes 3 misc-C parity divergences (`SUCCESSFUL`→`FAILED` under the flag):
- `regression/esbmc/fam_true_3` — flexible-array-member bounds relaxation lost
- `regression/esbmc-unix2/github_154_success` — struct field layout lost in the byte-array model
- `regression/esbmc/github_4634` — malloc'd-object reachability for the memory-leak walker

## Validation

- migrate unit tests 30/30 (new `sizeof_type`-survival case added)
- `irep2_bodies` regression family 31/31 (C/C++/Solidity/Jimple/Python/coverage)
- cbmc + floats + cstd flag-off 597/597 (no regression from the core irep2 node change)
- `code-reviewer`: 0 critical, 0 major
- New regression tests `github_4715_irep2_bodies_malloc_01{,_fail}` (positive + a negative that guards against vacuous passing)

Independent of #5323 (cluster 1, exceptions) and #5325 (cluster 2, coverage). Refs #4715.

### DIFF
--- a/regression/esbmc/github_4715_irep2_bodies_malloc_01/main.c
+++ b/regression/esbmc/github_4715_irep2_bodies_malloc_01/main.c
@@ -1,0 +1,30 @@
+// esbmc/esbmc#4715 (V.4.4 parity): under --irep2-bodies the body round-trip
+// dropped the folded sizeof(T) element type (#c_sizeof_type) from the malloc
+// size constant, so malloc(sizeof(pair)) allocated a `char` blob instead of a
+// `pair`. The subsequent struct copy `q = *p` then lost the field layout and
+// the value assertions failed spuriously. constant_int2t now carries the
+// sizeof type across the round-trip, so the allocated object is typed `pair`.
+#include <assert.h>
+#include <stdlib.h>
+
+typedef struct
+{
+  int id;
+  void *value;
+} pair;
+
+int main()
+{
+  pair *p = (pair *)malloc(sizeof(pair));
+  if (!p)
+    return 0;
+  p->id = 10;
+  p->value = (void *)20;
+
+  pair q = *p; // struct copy through the malloc'd object
+  assert(q.id == 10);
+  assert(q.value == (void *)20);
+
+  free(p);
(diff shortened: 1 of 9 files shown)


mikhailramalho:
At some point we need to kill this #c_sizeof_type field

PR #5334, opened 2026-06-12
[solver] Hide smt_astt/smt_sortt/ast_vec behind smt_convt

mikhailramalho:
## What

Confines the SMT solver-handle types — `smt_astt`, `smt_sortt`, and
`smt_convt::ast_vec` — and the AST/sort construction machinery to the solver
layer (`src/solvers/`). Nothing outside that directory can name them anymore.

The rest of ESBMC (the SSA equation, `bmc`, trace building, ctest/pytest
generation, the termination/ranking probes) now talks only `expr2tc` to a slim
`smt_convt` interface, which owns the full solver implementation
(`smt_solver_baset`). `smt_conv.h` is sealed: it forward-declares the
implementation and exposes only an expr2tc-level API, so including it does not
drag solver internals into the rest of the tree.

## Why

`smt_astt`/`smt_sortt`/`ast_vec` are solver-internal handles whose lifetime is
tied to a solver context. Leaking them into goto-symex (the SSA equation stored
`guard_ast`/`cond_ast` per step) coupled the equation to the solver and made
context push/pop reasoning fragile. Keeping conversion strictly inside the
solver layer removes that coupling.

## Key changes

- **Equation works in `expr2tc` only.** `SSA_stept` drops `guard_ast`/`cond_ast`
  for an `expr2tc cond_expr`; discharge is built as
  `not2tc(implies2tc(assumpt, cond))` and asserted via `disjunction(...)`,
  relying on `convert_ast`'s `expr2tc` memoisation rather than AST handles.
- **Slim `smt_convt`.** Owns a `smt_solver_baset`; forwards an expr2tc-level API
  (`convert_ast`/`convert_assign`/`assert_expr`/`get`/`get_by_type`/`get_by_ast`/
  `l_get`/`dump_expr`/`renumber_symbol_address`/push/pop/solve). New
  `get_by_ast(expr2tc)` resolves aggregate model values for non-symbol roots.
- **`smt_conv.h` sealed** from `smt_astt`/`smt_sortt`/`ast_vec`; file names
  realigned so `smt_solver.{h,cpp}` holds the implementation and `smt_conv.h`
  the slim interface.
- **`l_get` model cache internalised** (cleared on `pre_solve`/`push_ctx`/
  `pop_ctx`), removing the explicit `model_cache_scopet`.
- **`smt_resultt` folded** into a single header; per-class `resultt`/`P_*`
  shims removed.

## Testing

Pure internal refactor — no behavioural change, so no new regression tests.
Existing suites green on this branch: unit `422/422`, `esbmc-unix2` `186/186`,
`k-induction` `190/190`, `termination` `65/65`; previously context-sensitive
incremental tests (`01_pthread62`/`01_pthread60`) run without hanging.


lucasccordeiro (approved):
Nice!

PR #5364, opened 2026-06-13
[symex] Model __ESBMC_uninterpreted_* / __CPROVER_uninterpreted_* as uninterpreted functions

### LINKED ISSUE #5287: [SV-COMP] aws_hash_table_create false alarm: havoc'd hash-table slots, not __CPROVER_uninterpreted_ zero-hash
Root-cause follow-up to #5145, splitting it from the (incorrect) zero-hash theory pursued in #5283 (now closed).

## Summary

The false alarm on `c/aws-c-common/aws_hash_table_create_harness.i` (#5145) is **not** caused by the hasher mapping `NULL -> 0`. Verified against a current build:

- The harness's own `s_hash_for` already maps `NULL -> 42` and remaps `0 -> 1`, then asserts `hash_code != 0`. The zero sentinel is handled in-harness.
- The actual failing property is the post-create assertion in `aws_hash_table_create_harness`:
  ```c
  if (get_p_elem) {
      __VERIFIER_assert(uninterpreted_equals(p_elem->key, key));
  }
  ```
  i.e. after `aws_hash_table_create`, the returned element's key is not UF-equal to the requested `key`.

## What was ruled out

- **`ASSUME(ret != 0)` on `__CPROVER_uninterpreted_*` (the #5283 approach):** unsound (masks comparator `== 0` violations) **and** ineffective — the benchmark still reports `VERIFICATION FAILED` with the pass applied.
- **True uninterpreted-function modelling (ignore body, fresh nondet + functional congruence):** all three (`hasher`/`equals`/`compare`) were hand-modelled with congruence plus the harness's own symmetry/antisymmetry `assume_abort_if_not`s (memo capacity up to 512). **Still `FAILED`.** So the gap is not in how the three functions are modelled.

## Likely root cause

`ensure_allocated_hash_table` allocates the table with `bounded_malloc(required_bytes)` and leaves the slot array **havoc'd**. Nothing constrains a slot's `hash_code` to equal `hasher(slot.element.key)`. `s_find_entry`/`s_find_entry1` then matches a slot on `entry->hash_code == hash_code && s_hash_keys_eq(state, key, entry->element.key)`, and ESBMC can pick a havoc'd stored state in which the matched slot's key is not UF-equal to `key`. CBMC's encoding apparently does not admit this state.

This points at a symex / pointer-model / havoc-consistency question (slot `hash_code` <-> `hasher(key)` consistency, or the pointer-to-`uint64_t` cast model), not at call-site modelling of `__CPROVER_uninterpreted_*`.

## Reproducer

```
esbmc aws_hash_table_create_harness.i --64 --k-induction --max-inductive-step 3 \
  --no-pointer-check --no-bounds-check --interval-analysis \
  --force-malloc-success --no-div-by-zero-check
```
Benchmark: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_create_harness.i

## Next steps

- Confirm whether constraining each occupied slot's `hash_code == hasher(slot.key)` (a real hash-table invariant the harness does not assert) removes the false alarm.
- If so, decide whether to model that as a harness-level assumption or fix the underlying havoc/pointer-cast handling.

Refs #5145, #5283.

lucasccordeiro:
Model bodyless calls to functions named `__ESBMC_uninterpreted_*` (ESBMC's native namespace) or `__CPROVER_uninterpreted_*` (CBMC's convention) as genuine uninterpreted functions: the concrete body is discarded, the result is a fresh nondeterministic value, and functional congruence is enforced by Ackermannisation (for each earlier call to the same symbol with the same arity, `assume(args_equal => results_equal)`).

This is sound — it adds only a valid property of any function and prunes no behaviour — unlike the closed, unsound PR #5283, which forced a non-zero result.

### Two prefixes, two policies

- **`__ESBMC_uninterpreted_*` — always-on.** Like every other `__ESBMC_*` intrinsic, the native prefix needs no flag. Because all `c:@F@__ESBMC*` calls are routed to `run_intrinsic()` before body inlining (and otherwise hit the fatal "non-intrinsic prefixed with __ESBMC" guard), the native prefix is intercepted there; dead branches are skipped so they don't pollute the congruence history.
- **`__CPROVER_uninterpreted_*` — opt-in** via `--cprover-uninterpreted-functions` (default-off). Existing code may ship a real body for such a function, so it keeps using that body unless the flag is set. This alias is intercepted in `symex_function_call_code()`.

A single shared helper (`symex_uninterpreted_function`) handles both prefixes. (Note: the prefix-family mapping cannot be done with a `-D` macro à la `__ESBMC_alloca=__builtin_alloca`, since the preprocessor only rewrites whole identifier tokens, not the arbitrary suffix after `..._uninterpreted_`.)

### Tests

Six regression tests — three per prefix — cover congruence (equal args give an equal result), non-over-constraint (distinct args may differ), and the #5283 unsoundness probe (asserting the result is non-zero still FAILs). The native-prefix tests run without any flag, pinning down the always-on behaviour.

### Scope

This gives ESBMC real uninterpreted-function support but does **not** by itself resolve #5287 — that benchmark's residual false alarm is an upstream memory-model issue (a byte-encoded pointer field reconstructs to an unstable pointer identity across reads), independent of uninterpreted-function semantics. Refs #5287, #5145.


mikhailramalho:
This implementation feels out of place in symex. We should move it down to smt and only when the solver doesn't support (and I think all solvers we support do unimplemented functions).

I'm just no sure what's the best way to implement this tho. Maybe a boolean is_uf In a sideeffect2tc to flag that it is an unimplemented function?

lucasccordeiro:
Done in 464c74d5 — the modelling is moved out of symex into the SMT layer. Symex now just emits a new `uninterpreted_func2t` expr; each backend declares the function symbol once per name and applies it, so the solver enforces functional congruence natively (z3, bitwuzla, boolector, cvc4, cvc5, yices, mathsat, smtlib). `smt_convt` keeps an Ackermannisation fallback for any backend lacking native UF support. No `is_uf` flag on `sideeffect2tc` was needed.

PR #5368, opened 2026-06-14
[witness 2.1] support branching waypoints for the switch && Improve function_enter/return

XLiZHI:


XLiZHI:
TOTAL=100，PASS=92，FAIL=8
```

  - functions.3-1A0 — expected=true，got=FALSE_REACH
  - functions.3-1B0 — expected=true，got=FALSE_REACH
  - ternary.1A1 — expected=false，got=TRUE
  - ternary.1B0 — expected=true，got=FALSE_REACH
  - ternary.1C1 — expected=false，got=TRUE
  - ternary.2A1 — expected=false，got=TRUE
  - ternary.3A1 — expected=false，got=TRUE
  - ternary.4A1 — expected=false，got=TRUE
```

The round-trip consists of two steps: IRep1 → IRep2 (migrate_expr) followed by IRep2 → IRep1 (migrate_expr_back). In IRep2, only CODE-level nodes (code_decl2t, code_block2t, code_assign2t, etc.) carry a locationt location field, and that field is explicitly passed through both migration directions — so statement-level locations survive the round-trip. 

VALUE-level nodes (expression nodes such as if2t, plus2t, etc.) have no location field. Their locations are silently dropped during forward migration and are absent when the nodes are reconstructed during backward migration. After the round-trip, restore_value_locations traverses the reconstructed tree and calls stamp_value_locations on every value expression. When a value node's file is empty, it is overwritten with the location of the nearest ancestor CODE node. 

As a result, every value expression ends up carrying the location of its enclosing statement rather than the precise source position the frontend recorded for it.

Problems arose after #5366. In Witness, we need to precisely match the information. This is currently our limitation.

XLiZHI:
update:

```
==> esbmc-violation-v2 <==
  correct:              97
    correct true:       56
    correct false:      41
  incorrect:             3
    incorrect true:      0
    incorrect false:     3
  unknown:               0
  Score:               105 (max: 159)
  
  https://github.com/esbmc/esbmc/actions/runs/27695312471/job/81921771518
```

PR #5375, opened 2026-06-15
[Concurrency] remove hash state if mpor cuts the interleaving

Anthonysdu:
ESBMC only executes a transition before knowing it can be scheduled by mpor, we remove its hash record if the last transition shouldn't happen.


Anthonysdu:
This PR: --context-bound 3 --30s

https://github.com/esbmc/esbmc/actions/runs/27575038933

```
  correct:            1120
    correct true:      706
    correct false:     414
  incorrect:             5
    incorrect true:      2
    incorrect false:     3
  unknown:            2051
  Score:              1714 (max: 5722)

```

Master:

https://github.com/esbmc/esbmc/actions/runs/27490499289

```
  correct:            1097
    correct true:      692
    correct false:     405
  incorrect:             6
    incorrect true:      3
    incorrect false:     3
  unknown:            2073
  Score:              1645 (max: 5722)
```

lucasccordeiro (approved):

PR #5379, opened 2026-06-16
Refactor IR-IEEE encoding into dedicated converter

ShemaAlosaimi:
## Summary

  - Extracts all IR-IEEE integer-encoding logic from `smt_solver_baset` into a new `ir_ieee_convt` class
  (`src/solvers/smt/fp/ir_ieee_conv.{h,cpp}`)
  - Moves the five rounding-mode enclosure helpers (RNE/RNA/RUP/RDN/RTZ), the interval map (`ir_ra_interval_map`),
  the ranged-symbol set, and the per-operation encode methods (`encode_ieee_{add,sub,mul,div,fma}`) into the new
  class
  - `smt_solver_baset` now holds an `ir_ieee_api` pointer and delegates the `int_encoding` branch of each
  `ieee_*_id` switch case to it; the sqrt case and `apply_ieee754_semantics` remain in `smt_solver_baset` and call
  back into `ir_ieee_api` for interval/enclosure infrastructure
  - Pure structural refactor — no behaviour change; all 153 IR-RA/IR-IEEE regression tests pass


lucasccordeiro:
**Review — clean extraction, one real issue.**

The move is faithful: each `int_encoding` branch now delegates to `ir_ieee_api`, the sqrt case and `apply_ieee754_semantics` stay put and call back, and the `apply_enclosure` dispatcher + the static `is_known_rounding_mode`/`is_single_or_double` helpers nicely collapse the five repeated `||` chains. No behaviour change spotted.

**Should fix**
- **Ownership (`smt_solver.cpp`):** `ir_ieee_api` is raw-`new`'d in the ctor and `delete`'d in a hand-written `~smt_solver_baset()`, but the class has no deleted/defined copy or move. If it's ever copied → double-free. Use `std::unique_ptr<ir_ieee_convt>`; that also lets you drop the custom destructor (back to `= default`) and is exception-safe. Note the other `*_api` pointers are *not* owned here, so this one is the odd one out.

**Minor**
- `ir_ieee_conv.h` includes `<util/message.h>` but the header has no logging — appears unused, drop it.
- `class ir_ieee_convt;` is declared above the `#include`s in `smt_solver.h`; move it down with the other forward decls for consistency.
- Since this is a refactor pass: the per-format eps lookup (`single`/`double` → `eps_rel`/`eps_abs`) is still duplicated verbatim across all five `apply_ieee754_*_enclosure` helpers. Factoring it into one small helper would be the natural next step (optional, out of scope for a pure move).

Nothing blocking once the ownership point is addressed.

ShemaAlosaimi:
> **Review — clean extraction, one real issue.**
> 
> The move is faithful: each `int_encoding` branch now delegates to `ir_ieee_api`, the sqrt case and `apply_ieee754_semantics` stay put and call back, and the `apply_enclosure` dispatcher + the static `is_known_rounding_mode`/`is_single_or_double` helpers nicely collapse the five repeated `||` chains. No behaviour change spotted.
> 
> **Should fix**
> 
>     * **Ownership (`smt_solver.cpp`):** `ir_ieee_api` is raw-`new`'d in the ctor and `delete`'d in a hand-written `~smt_solver_baset()`, but the class has no deleted/defined copy or move. If it's ever copied → double-free. Use `std::unique_ptr<ir_ieee_convt>`; that also lets you drop the custom destructor (back to `= default`) and is exception-safe. Note the other `*_api` pointers are _not_ owned here, so this one is the odd one out.
> 
> 
> **Minor**
> 
>     * `ir_ieee_conv.h` includes `<util/message.h>` but the header has no logging — appears unused, drop it.
> 
>     * `class ir_ieee_convt;` is declared above the `#include`s in `smt_solver.h`; move it down with the other forward decls for consistency.
> 
>     * Since this is a refactor pass: the per-format eps lookup (`single`/`double` → `eps_rel`/`eps_abs`) is still duplicated verbatim across all five `apply_ieee754_*_enclosure` helpers. Factoring it into one small helper would be the natural next step (optional, out of scope for a pure move).
> 
> 
> Nothing blocking once the ownership point is addressed.

Thanks, @lucasccordeiro , for the review!

I addressed the ownership issue by switching `ir_ieee_api` to `std::unique_ptr<ir_ieee_convt>` and constructing it with `std::make_unique`. Since `ir_ieee_convt` is forward-declared in the header, the destructor is declared in `smt_solver.h` and defined as `= default` in `smt_solver.cpp`, where the full type is visible.

I also dropped the unused `<util/message.h>` include from `ir_ieee_conv.h`. Removing it exposed that `smt_ast.h` was relying on transitive inclusion even though it calls `log_status` from an inline method, so I added the direct include there.

Finally, I moved the `ir_ieee_convt` forward declaration into the existing forward-declaration block for consistency.

The IR-RA/IR-IEEE regression suite still passes: 153/153.

lucasccordeiro (approved):

PR #5381, opened 2026-06-16
[python] multi-element list repetition with runtime count

prototypeC14:
This PR fixes a Python frontend bug where `[a, b] * n` only repeated the first element (produce a list of length 3 instead of 6).


lucasccordeiro (approved):

PR #5388, opened 2026-06-17
[om] add POSIX socket/select/poll operational models and headers

Gugarauj07:
Add operational models for POSIX network APIs to enable formal verification of network software with ESBMC.

New headers (src/c2goto/headers/):
- sys/socket.h  : socket types, sockaddr, socklen_t, AF_*/SOCK_* constants
- sys/select.h  : fd_set, select(), poll(), struct pollfd, POLL* flags
- netinet/in.h  : sockaddr_in, sockaddr_in6, in_addr, IPPROTO_* constants
- arpa/inet.h   : inet_addr(), inet_pton(), htons/htonl/ntohs/ntohl

New operational model (src/c2goto/library/socket_lib.c):
- socket()     : returns non-deterministic fd in [3, 1024)
- bind()       : non-deterministic success/failure
- listen()     : non-deterministic success/failure
- accept()     : returns non-deterministic client fd; fills addr non-det
- connect()    : non-deterministic success/failure
- recv()       : fills buf with non-deterministic bytes (attacker input)
- send()       : returns non-deterministic count in [0, len]
- recvfrom()   : like recv(), also fills src_addr non-deterministically
- sendto()     : like send(), with destination address
- setsockopt() / getsockopt() : non-deterministic success/failure
- getsockname() / getpeername(): fills addr non-deterministically
- shutdown()   : non-deterministic success/failure
- select()     : returns non-deterministic readiness count
- poll()       : fills revents non-deterministically per requested events
- htons/htonl/ntohs/ntohl : modeled as identity (byte order irrelevant)
- inet_addr() / inet_pton() : non-deterministic address values

Infrastructure:
- CMakeLists.txt: register new headers for mangling into libc_hdr.h
- cprover_libc_sources.cpp: create netinet/ and arpa/ dirs at runtime

Regression tests:
- socket_recv_overflow_false: recv() return used as index without bounds check triggers array-bounds violation (VERIFICATION FAILED)
- socket_recv_safe_true: same pattern with proper guard is safe (VERIFICATION SUCCESSFUL)


src/c2goto/library/socket_lib.c
```diff
@@ -0,0 +1,474 @@
...
+
+    return result;
+}
+
+/* ============================================================
+ *  Byte-order conversion functions (htons, htonl, ntohs, ntohl)
+ *  These are modeled as identity functions for verification
+ *  purposes, since ESBMC does not distinguish byte order.
+ * ============================================================ */
+uint16_t htons(uint16_t hostshort)
```
lucasccordeiro: **Blocker — duplicate definition + soundness regression.** `library/inet.c` already defines `htons/htonl/ntohs/ntohl` with correct endianness (`__ESBMC_is_little_endian()` + `__builtin_bswap*`). These identity versions collide with those symbols and silently drop byte-order semantics, masking real endianness bugs. Drop all four from this file and keep the `inet.c` models.

src/c2goto/library/socket_lib.c
```diff
@@ -0,0 +1,474 @@
...
+    }
+    return 0;
+}
+
+/* ============================================================
+ *  select()
+ *  Returns the number of ready fds (>= 0), or -1 on error.
+ *  Modifies readfds/writefds/exceptfds non-deterministically.
+ * ============================================================ */
+int select(int nfds, fd_set *readfds, fd_set *writefds,
```
lucasccordeiro: `select()` never modifies the fd_sets, so after a `>0` return the caller's `FD_ISSET` reads the bits it set itself — `select` looks like it always reports every fd ready. Either nondet-clear the sets or, at minimum, fix the comment on L359 which claims it modifies them.

src/c2goto/library/socket_lib.c
```diff
@@ -0,0 +1,474 @@
...
+        errno = EINTR;
+        return -1;
+    }
+
+    __ESBMC_assume(result >= 0 && (unsigned long)result <= nfds);
+
+    /* Fill revents with non-deterministic values constrained to
+     * the requested events plus error flags */
+    for (unsigned long i = 0; i < nfds; i++) {
+        fds[i].revents = __VERIFIER_nondet_short();
```
lucasccordeiro: Comment (L398) says `revents` is constrained to requested events + error flags, but this assigns a raw nondet short — `revents` can carry bits the caller never put in `events`. Either mask with `fds[i].events | POLLERR | POLLHUP | POLLNVAL`, or drop the comment.

src/c2goto/library/socket_lib.c
```diff
@@ -0,0 +1,474 @@
...
+        "send: buf must not be NULL");
+
+    ssize_t result = (ssize_t)__VERIFIER_nondet_int();
+
+    if (result < 0) {
+        errno = EPIPE;
+        return -1;
+    }
+
+    __ESBMC_assume(result >= 0 && (size_t)result <= len);
```
lucasccordeiro: `result >= 0` is dead — the `if (result < 0) return -1` above already guarantees it. Same in `sendto` (L257), `select` (L373), `poll` (L396). Trim to the upper bound.

src/c2goto/headers/sys/socket.h
```diff
@@ -0,0 +1,108 @@
...
+
+int    setsockopt(int sockfd, int level, int optname,
+                  const void *optval, socklen_t optlen);
+int    getsockopt(int sockfd, int level, int optname,
+                  void *optval, socklen_t *optlen);
+int    getsockname(int sockfd, struct sockaddr *addr, socklen_t *addrlen);
+int    getpeername(int sockfd, struct sockaddr *addr, socklen_t *addrlen);
+
+int    shutdown(int sockfd, int how);
+int    close(int fd);
```
lucasccordeiro: `close()` belongs to `<unistd.h>`, not `<sys/socket.h>`; there's no model for it here and this risks a conflicting redeclaration. Drop it.

src/c2goto/headers/sys/socket.h
```diff
@@ -0,0 +1,108 @@
...
+
+#include <__esbmc/stddefs.h>
+#include <stdint.h>
+#include <stddef.h>
+
+__ESBMC_C_CPP_BEGIN
+
+#ifndef _SSIZE_T_DEFINED
+#define _SSIZE_T_DEFINED
+typedef long ssize_t;
```
lucasccordeiro, line 15: Locally-guarded `ssize_t` typedef can clash with the one the C library headers expose under a different guard macro. Worth confirming a program that includes `<unistd.h>`/`<sys/types.h>` alongside this doesn't hit a redefinition.

src/c2goto/library/socket_lib.c
```diff
@@ -0,0 +1,474 @@
...
+        "inet_pton: src must not be NULL");
+    __ESBMC_assert(dst != (void *)0,
+        "inet_pton: dst must not be NULL");
+
+    int result = __VERIFIER_nondet_int();
+    __ESBMC_assume(result >= -1 && result <= 1);
+
+    if (result == 1) {
+        /* Fill dst with non-deterministic address bytes */
+        int size = (af == 2 /* AF_INET */) ? 4 : 16;
```
lucasccordeiro: Use the `AF_INET` macro instead of the magic `2` (include `<sys/socket.h>` already gives it) — clearer and survives any future value change.

lucasccordeiro (changes requested):
Nice addition — useful coverage of the POSIX socket surface. A few things to tighten before merge; the byte-order one is a blocker.

PR #5407, opened 2026-06-17
[Python] Fix KNIONBUG: np.copysign/fmax/fmin crash on scalar constants

prototypeC14:
This PR fixes a Python frontend crash where `np.copysign(...)`, `np.fmax(...)`, and `np.fmin(...)` on scalar constants triggered "Unknown operator: <fn>" and crashed ESBMC during GOTO migration. These three are libm-style binary ufuncs with no entry in `operator_map()` and no dedicated handler (unlike `fmod` and `power`), so the scalar-pair BinOp dispatch path constructed an exprt with an empty operator id that `migrate_expr` could not lower.

### DIFF
--- a/regression/numpy/copysign/test.desc
+++ b/regression/numpy/copysign/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
 --incremental-bmc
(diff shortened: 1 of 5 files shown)


lucasccordeiro (approved):

PR #5409, opened 2026-06-18
[contracts] Bound array-elem assigns witness by the real is_fresh size

### LINKED ISSUE #5314: --enforce-contract: __ESBMC_assigns(a[k]) raises false 'array bounds violated' (only bare *a works)
## Introduction

Under `--enforce-contract`, naming an array element as an assigns target — `__ESBMC_assigns(a[k])`, or equivalently `__ESBMC_assigns(*(a+k))` — makes ESBMC report `VERIFICATION FAILED` with `dereference failure: array bounds violated`, even when the element is trivially inside the allocated region (e.g. `a[0]` in a 10-element `__ESBMC_is_fresh` buffer). The bare dereference `__ESBMC_assigns(*a)` on the same buffer verifies cleanly.

The false positive is attributed to the contract (line 1), not the body: removing the assigns clause entirely lets the body's own `a[0] = 0;` verify, so the spurious bounds check is on the assigns *target*, not the write. Because `*a` only names the first element, there is currently no way to write a frame clause for `a[k]` with `k > 0`, nor for a buffer range, without tripping this.

## Minimal Repro

```c
__ESBMC_contract
void f(int* a) {
    __ESBMC_requires(__ESBMC_is_fresh(a, 10 * sizeof(int)));
    __ESBMC_assigns(a[0]);
    a[0] = 0;
}
```

```
$ esbmc repro.c --function f --enforce-contract f
Violated property:
VERIFICATION FAILED
```

- `__ESBMC_assigns(*a)` (same program) → `VERIFICATION SUCCESSFUL`.
- `__ESBMC_assigns(a[2])` and `__ESBMC_assigns(*(a+2))` (10-int buffer) → same false `array bounds violated`.
- Dropping the assigns clause → `VERIFICATION SUCCESSFUL` (the body write is in-bounds).

Tested on ESBMC 8.3.0 (x86_64 linux), Z3 v4.16.0.

lucasccordeiro:
Fixes #5314 — under `--enforce-contract`, naming an array element as an assigns target (`__ESBMC_assigns(a[k])` / `__ESBMC_assigns(*(a+k))`) reported a spurious `VERIFICATION FAILED: dereference failure: array bounds violated`, even for an in-bounds element. Bare `__ESBMC_assigns(*a)` worked.

**Root cause.** The Phase 2B array-element frame rule (`materialize_arr_elem_snapshots`) builds a nondet witness index `j` to assert that no non-assigned element changed, and constrained `j` to `[0, ARRAY_ALLOC_ELEMS)` with `ARRAY_ALLOC_ELEMS = 100`. That bound matches the allocation made by `add_pointer_validity_assumptions`, but `__ESBMC_is_fresh(a, n)` allocates exactly `n` bytes (`n/sizeof(elem)` elements) and is *skipped* by that pass. So for an is_fresh buffer of 10 ints, `a[j]` with `j` up to 99 reads past the real allocation → spurious bounds violation.

**Fix.** Thread a `pointer-symbol → byte-size` map built in `generate_checking_wrapper`'s is_fresh allocation loop into `materialize_arr_elem_snapshots`, and bound the witness index by the real element count `size/sizeof(elem)` when the array pointer was allocated by `__ESBMC_is_fresh`; otherwise keep `ARRAY_ALLOC_ELEMS`.

**Soundness.** The smaller bound never masks a real frame violation: a write to an in-bounds element outside the assigns clause is still in `[0, real_count)` and caught, while an out-of-bounds write is caught by the ordinary bounds checker. A negative regression test that writes `a[9]` (the last valid element, not in the assigns clause) still reports `VERIFICATION FAILED`.

**Tests.** `regression/function_contract/github_5314_is_fresh_arr_elem_pass` (a[k] target ⇒ SUCCESSFUL) and `github_5314_is_fresh_arr_elem_fail` (out-of-clause write ⇒ FAILED). Validated via the binary (the function_contract suite is ctest-skipped on macOS); no regressions vs. the pre-fix binary.

### DIFF
--- a/regression/function_contract/github_5314_is_fresh_arr_elem_fail/main.c
+++ b/regression/function_contract/github_5314_is_fresh_arr_elem_fail/main.c
@@ -1,0 +1,11 @@
+/* github #5314 soundness guard: the witness index still ranges over the whole
+ * is_fresh buffer, so a write to an element NOT in the assigns clause (here a[9],
+ * the last valid element) must still be reported as a frame violation. */
+__ESBMC_contract
+void f(int *a)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(a, 10 * sizeof(int)));
+  __ESBMC_assigns(a[0]);
+  a[0] = 0;
+  a[9] = 5; /* not in assigns clause */
+}
--- a/regression/function_contract/github_5314_is_fresh_arr_elem_fail/test.desc
+++ b/regression/function_contract/github_5314_is_fresh_arr_elem_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f
+^VERIFICATION FAILED$
--- a/regression/function_contract/github_5314_is_fresh_arr_elem_pass/main.c
+++ b/regression/function_contract/github_5314_is_fresh_arr_elem_pass/main.c
@@ -1,0 +1,11 @@
+/* github #5314: under --enforce-contract, naming an array element a[k] as an
+ * assigns target tripped a spurious "array bounds violated" because the Phase-2B
+ * witness index was bounded by ARRAY_ALLOC_ELEMS (100) instead of the real
+ * __ESBMC_is_fresh allocation (10 elements). a[0] / a[2] / *(a+2) all in bounds. */
+__ESBMC_contract
+void f(int *a)
+{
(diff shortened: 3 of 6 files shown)


LukeW1999 (approved):
LGTM

PR #5416, opened 2026-06-19
[ir-ieee] propagate NaN predicates through arithmetic

ShemaAlosaimi:
## Summary

This PR propagates already-known NaN predicates through `--ir-ieee` arithmetic operations.

The previous NaN-related PR introduced minimal NaN predicate tracking for negative `sqrt` results and used those predicates when encoding floating-point comparisons. However, the predicate was lost once the value passed through arithmetic operations.

This PR propagates that NaN information through arithmetic operations such as:

```c
double s = sqrt(-1.0);
double t = s + 1.0;
```

After this change, `t` keeps the known NaN predicate from `s`, so later comparisons involving `t` can still apply IEEE NaN comparison behaviour.

## Motivation

The existing NaN-aware comparison handling only works when the compared operand still has a known NaN predicate.

For example, this was handled:

```c
double s = sqrt(-1.0);
assert(s != s);
```

But after an arithmetic operation, the predicate was no longer attached to the result:

```c
double s = sqrt(-1.0);
double t = s + 1.0;
assert(t != t);
```

Since IEEE arithmetic propagates NaN through arithmetic operations, the `--ir-ieee` metadata should preserve already-known NaN predicates through these operations.

## Changes

This PR:

* Propagates known NaN predicates through `ieee_add`.
* Propagates known NaN predicates through `ieee_sub`.
* Propagates known NaN predicates through `ieee_mul`.
* Propagates known NaN predicates through `ieee_div`.
* Propagates known NaN predicates through `ieee_fma`.
* Updates the `sqrt` NaN predicate so `sqrt(NaN)` is also tracked as NaN-aware.

The NaN predicate combination is implemented as a local helper in `ir_ieee_conv.cpp`; no new public API is added.

For binary arithmetic operations, if either operand has a known NaN predicate, the result stores the combined predicate.

For FMA, known NaN predicates from all three operands are combined and stored on the result.

For `sqrt`, the result NaN predicate now includes both:

* the operand's known NaN predicate, if any
* the existing negative-operand predicate `!(operand >= 0)`

This allows cases such as `sqrt(sqrt(-1.0))` to remain NaN-aware.

## Regression Test

Six new CORE regression tests were added under `regression/ir-ra/`:

* `ra-nan-prop-add`
* `ra-nan-prop-sub`
* `ra-nan-prop-mul`
* `ra-nan-prop-div`
* `ra-nan-prop-fma`
* `ra-sqrt-of-nan`

Each test uses `sqrt(-1.0)` as the existing known NaN source, applies the named operation, and checks that the result still follows IEEE NaN comparison behaviour.

Expected result for all tests: `VERIFICATION SUCCESSFUL`.

## Notes

This PR does not introduce new NaN sources.

NaN propagation only applies when an operand already has a known NaN predicate. Currently, those predicates come from the existing negative `sqrt` NaN tracking.

This PR does not add division-by-zero NaN cases, infinities, signed zero, cast rounding, nearbyint, or `isnan` support.

### DIFF
--- a/regression/ir-ra/ra-sqrt-of-nan/main.c
+++ b/regression/ir-ra/ra-sqrt-of-nan/main.c
@@ -1,0 +1,10 @@
+#include <math.h>
+
+int main(void)
+{
+  double s = sqrt(-1.0);
+  double t = sqrt(s);
+  __ESBMC_assert(!(t == t), "sqrt(NaN) == itself is false");
+  __ESBMC_assert(t != t, "sqrt(NaN) != itself");
+  return 0;
+}
--- a/regression/ir-ra/ra-sqrt-of-nan/main.c
+++ b/regression/ir-ra/ra-sqrt-of-nan/main.c
@@ -4,7 +4,7 @@ int main(void)
 {
   double s = sqrt(-1.0);
   double t = sqrt(s);
-  __ESBMC_assert(!(t == t), "sqrt(NaN) == itself is false");
+  __ESBMC_assert(!(t < 100.0), "sqrt(NaN) < 100.0 is false");
   __ESBMC_assert(t != t, "sqrt(NaN) != itself");
   return 0;
 }
--- a/src/solvers/smt/fp/ir_ieee_conv.cpp
+++ b/src/solvers/smt/fp/ir_ieee_conv.cpp
@@ -83,7 +83,9 @@ smt_astt ir_ieee_convt::combine_nan_preds(smt_astt a, smt_astt b) const
 }
 
 void ir_ieee_convt::store_combined_nan_pred(
-  smt_astt result, smt_astt s1, smt_astt s2)
+  smt_astt result,
+  smt_astt s1,
+  smt_astt s2)
 {
   smt_astt nan_p = combine_nan_preds(get_nan_pred(s1), get_nan_pred(s2));
   if (nan_p)
--- a/src/solvers/smt/fp/ir_ieee_conv.cpp
+++ b/src/solvers/smt/fp/ir_ieee_conv.cpp
@@ -82,6 +82,17 @@ ir_ieee_convt::apply_nan_cmp(smt_astt cmp, smt_astt a, smt_astt b, bool is_neq)
   return ctx->mk_ite(either_nan, ctx->mk_smt_bool(is_neq), cmp);
 }
 
+static smt_astt combine_nan_preds(smt_solver_baset *ctx, smt_astt a, smt_astt b)
+{
+  if (!a && !b)
+    return nullptr;
+  if (!a)
+    return b;
+  if (!b)
+    return a;
+  return ctx->mk_or(a, b);
+}
+
 std::pair<smt_astt, smt_astt> ir_ieee_convt::apply_ieee754_rne_enclosure(
   smt_astt real_result,
   smt_astt lo_r,
@@ -418,6 +429,10 @@ smt_astt ir_ieee_convt::encode_ieee_add(const expr2tc &expr)
     auto bounds =
       apply_enclosure(real_result, lo_r, hi_r, fbv_type, rounding_mode);
     store_interval(real_result, bounds.first, bounds.second);
+    smt_astt nan_p =
+      combine_nan_preds(ctx, get_nan_pred(side1), get_nan_pred(side2));
+    if (nan_p)
+      store_nan_pred(real_result, nan_p);
     return real_result;
   }
   return ctx->apply_ieee754_semantics(
@@ -443,6 +458,10 @@ smt_astt ir_ieee_convt::encode_ieee_sub(const expr2tc &expr)
     auto bounds =
       apply_enclosure(real_result, lo_r, hi_r, fbv_type, rounding_mode);
     store_interval(real_result, bounds.first, bounds.second);
+    smt_astt nan_p =
+      combine_nan_preds(ctx, get_nan_pred(side1), get_nan_pred(side2));
+    if (nan_p)
+      store_nan_pred(real_result, nan_p);
     return real_result;
   }
   return ctx->apply_ieee754_semantics(
@@ -493,6 +512,10 @@ smt_astt ir_ieee_convt::encode_ieee_mul(const expr2tc &expr)
     auto bounds =
       apply_enclosure(real_result, lo_r, hi_r, fbv_type, rounding_mode);
     store_interval(real_result, bounds.first, bounds.second);
+    smt_astt nan_p =
+      combine_nan_preds(ctx, get_nan_pred(side1), get_nan_pred(side2));
+    if (nan_p)
+      store_nan_pred(real_result, nan_p);
     return real_result;
   }
   return ctx->apply_ieee754_semantics(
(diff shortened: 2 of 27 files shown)


lucasccordeiro (commented):
Nice, focused follow-up — propagating an existing NaN predicate by OR-ing the operand predicates is sound for `add`/`sub`/`mul`/`div`/`fma` (any NaN operand ⇒ NaN result), and the predicates stay *sufficient* conditions so comparisons can't be falsely forced. A few small cleanups below, plus one test-coverage gap.

**Missing negative test.** All six tests assert `VERIFICATION SUCCESSFUL`, so they pass even if the NaN path never fires (e.g. an over-wide interval would also satisfy `!(t<100.0)`). Per the repo convention (one passing + one failing per PR), please add a companion that asserts `t == t` and expects `VERIFICATION FAILED` — that proves the propagated NaN semantics actually drive the result rather than passing vacuously.

src/solvers/smt/fp/ir_ieee_conv.cpp
```diff
@@ -82,6 +82,17 @@ ir_ieee_convt::apply_nan_cmp(smt_astt cmp, smt_astt a, smt_astt b, bool is_neq)
   return ctx->mk_ite(either_nan, ctx->mk_smt_bool(is_neq), cmp);
 }
 
+static smt_astt combine_nan_preds(smt_solver_baset *ctx, smt_astt a, smt_astt b)
```
lucasccordeiro: This is a third copy of the same "OR two optional predicates" logic — `apply_nan_cmp` already does it inline just above (`either_nan = (na && nb) ? mk_or(na,nb) : ...`), and `smt_solver.cpp` does it again for sqrt. Suggest making this a member (drop the `ctx` param — the class already holds `ctx`), then reuse it in `apply_nan_cmp` so there's a single combiner.

src/solvers/smt/fp/ir_ieee_conv.cpp
```diff
@@ -418,6 +429,10 @@ smt_astt ir_ieee_convt::encode_ieee_add(const expr2tc &expr)
     auto bounds =
       apply_enclosure(real_result, lo_r, hi_r, fbv_type, rounding_mode);
     store_interval(real_result, bounds.first, bounds.second);
+    smt_astt nan_p =
+      combine_nan_preds(ctx, get_nan_pred(side1), get_nan_pred(side2));
+    if (nan_p)
+      store_nan_pred(real_result, nan_p);
```
lucasccordeiro: This 4-line block is repeated verbatim in `sub`/`mul`/`div` (and a 3-operand variant in `fma`). Consider a one-line helper, e.g. `store_combined_nan_pred(result, side1, side2)`, to collapse the four copies.

src/solvers/smt/smt_solver.cpp
```diff
@@ -954,7 +954,12 @@ smt_astt smt_solver_baset::convert_ast_node(const expr2tc &expr)
         a = mk_ite(op_nonneg, pos_result, sqrt_nan);
       }
       if (ir_ieee)
-        ir_ieee_api->store_nan_pred(a, mk_not(op_nonneg));
+      {
+        smt_astt op_nan = ir_ieee_api->get_nan_pred(operand);
+        smt_astt neg_pred = mk_not(op_nonneg);
+        ir_ieee_api->store_nan_pred(
+          a, op_nan ? mk_or(op_nan, neg_pred) : neg_pred);
```
lucasccordeiro: Same OR-combine pattern again. If `combine_nan_preds` were a public method on `ir_ieee_api` it could be called here too, unifying all three sites.

regression/ir-ra/ra-sqrt-of-nan/main.c
```diff
@@ -0,0 +1,10 @@
+#include <math.h>
+
+int main(void)
+{
+  double s = sqrt(-1.0);
+  double t = sqrt(s);
+  __ESBMC_assert(!(t == t), "sqrt(NaN) == itself is false");
```
lucasccordeiro: `!(t == t)` and `t != t` are the same predicate, so the two asserts are redundant. The other tests pair a distinct `!(t < 100.0)` check with `t != t`; mirror that here for an independent check.

ShemaAlosaimi:
> Nice, focused follow-up — propagating an existing NaN predicate by OR-ing the operand predicates is sound for `add`/`sub`/`mul`/`div`/`fma` (any NaN operand ⇒ NaN result), and the predicates stay _sufficient_ conditions so comparisons can't be falsely forced. A few small cleanups below, plus one test-coverage gap.
> 
> **Missing negative test.** All six tests assert `VERIFICATION SUCCESSFUL`, so they pass even if the NaN path never fires (e.g. an over-wide interval would also satisfy `!(t<100.0)`). Per the repo convention (one passing + one failing per PR), please add a companion that asserts `t == t` and expects `VERIFICATION FAILED` — that proves the propagated NaN semantics actually drive the result rather than passing vacuously.





Done. I added failing companion tests for the propagated-NaN cases:



* `ra-nan-prop-add-fail`

* `ra-nan-prop-sub-fail`

* `ra-nan-prop-mul-fail`

* `ra-nan-prop-div-fail`

* `ra-nan-prop-fma-fail`

* `ra-sqrt-of-nan-fail`



Each one asserts `t == t` and expects `VERIFICATION FAILED`, so the tests now check that the propagated NaN predicate actually drives the comparison result rather than only passing due to an over-wide interval.

src/solvers/smt/fp/ir_ieee_conv.cpp
ShemaAlosaimi: Done. I promoted `combine_nan_preds` into `ir_ieee_convt` as a `const` member and reused it in `apply_nan_cmp`, replacing the duplicated inline logic there.

It is exposed publicly because the `sqrt` path in `smt_solver.cpp` also needs to use the same combiner through `ir_ieee_api`. The helper no longer takes `ctx` as a parameter since `ir_ieee_convt` already stores it.

src/solvers/smt/fp/ir_ieee_conv.cpp
ShemaAlosaimi: Done. I added a private `store_combined_nan_pred(result, s1, s2)` helper that combines the operand NaN predicates and stores the result predicate when present.

The repeated blocks in `add`, `sub`, `mul`, and `div` are now collapsed to one call each. For `fma`, I kept the explicit nested `combine_nan_preds` calls because it combines three operands.

src/solvers/smt/smt_solver.cpp
ShemaAlosaimi: Done. The `sqrt` path now uses the same shared combiner:

```cpp

ir_ieee_api->combine_nan_preds(

  ir_ieee_api->get_nan_pred(operand), mk_not(op_nonneg))

```
Because `mk_not(op_nonneg)` is always non-null, `combine_nan_preds` is guaranteed to return a non-null predicate, so the previous explicit null-check and `mk_or` call are no longer needed at the call site.

This keeps the `sqrt` NaN predicate consistent with the arithmetic and comparison paths.

regression/ir-ra/ra-sqrt-of-nan/main.c
ShemaAlosaimi: Fixed. I replaced the redundant `!(t == t)` assertion with an ordered-comparison check:

```c

__ESBMC_assert(!(t < 100.0), "sqrt(NaN) < 100.0 is false");

__ESBMC_assert(t != t, "sqrt(NaN) != itself");

```

This mirrors the pattern used in the other passing tests: one ordered-comparison check and one self-inequality check.

PR #5421, opened 2026-06-19
[witness] Fix ternary branching waypoint column resolution

XLiZHI:
Violation witness validation failed on ternary-operator branching waypoints because the column number was not being filled in and constant guards were incorrectly skipped during segment matching.

 Changes:
- Wire fill_columns into the clang C frontend parse path so ternary waypoints without explicit columns get their ? column resolved before validation 

- Skip witness segment matching when the branch guard is a compile-time constant (!is_constant(old_guard)), preventing spurious mismatches on branches that have no runtime decision point 

- Add regression test svcomp_ternary_1B0 covering the fixed case - Deduplicate identical main.c, ternary.c, and test.desc files in witnesses_validate/ with relative symlinks


XLiZHI:
```
==> esbmc-violation-v2 <==

  correct:              98
    correct true:       57
    correct false:      41
  incorrect:             2
    incorrect true:      0
    incorrect false:     2
  unknown:               0
  Score:               123 (max: 159)
  
  https://github.com/esbmc/esbmc/actions/runs/27830675948/job/82371557771
```

lucasccordeiro (approved):

PR #5424, opened 2026-06-19
[python] Give Python class instances GC lifetime to fix escape UAF

### LINKED ISSUE #4773: [python] regression/python/github_4117_function_internal: instance-var assignments inside function bodies not tracked
`regression/python/github_4117_function_internal` is marked `KNOWNBUG` — ESBMC does not produce the expected `VERIFICATION SUCCESSFUL`.

Source ([`regression/python/github_4117_function_internal/main.py`](https://github.com/esbmc/esbmc/blob/master/regression/python/github_4117_function_internal/main.py)):

```python
class Node:
    def __init__(self, v):
        self.value = v
        self.next = None

def setup() -> Node:
    n1 = Node(1)
    n2 = Node(2)
    n1.next = n2
    return n1

n = setup()
assert n.next.value == 2
```

- **Expected verdict**: `VERIFICATION SUCCESSFUL`.
- **Flags**: (none — default driver)
- **Limitation** (documented in the test): the usage-site scanner does not track instance variables created inside function bodies, so `<var>.attr = <other>` assignments happening inside a function don't refine the attribute's inferred type. The fix needs scope-aware tracking of local instance variables.

Related to closed issue #4117.

## Reproduce

```sh
cd build
ctest -R "^regression/python/github_4117_function_internal$" --output-on-failure
```

## Definition of done

Extend the usage-site scanner to track per-function-scope instance variables and propagate attribute assignments, then flip the test to `CORE`.

Part of #4769.

lucasccordeiro:
## What

Fixes #4773. Python class instances whose address escapes their defining function (e.g. `setup()` returns `n1` whose `.next` points to the function-local `n2`) were reported as a false use-after-free (`accessed expired variable pointer`), because ESBMC stack-allocates Python objects.

```python
class Node:
    def __init__(self, v):
        self.value = v
        self.next = None
def setup() -> Node:
    n1 = Node(1); n2 = Node(2)
    n1.next = n2
    return n1            # n1.next dangled to the popped frame's n2
n = setup()
assert n.next.value == 2   # was: VERIFICATION FAILED (CWE-416)
```

## Why this approach

The root cause is purely a **lifetime** mismatch. Python is garbage-collected and has no manual `free`, so a reference that outlives its defining frame is valid at runtime — the "expired variable pointer" check is always a false positive for Python objects. A frontend "heap-allocate instances as pointers" prototype was evaluated and rejected: it cascades into ~57 regressions (dynamic attribute growth vs. fixed malloc size; pervasive struct-value assumptions across aliasing/return/identity/`isinstance`/list-of-objects; the return path copies the heap object back to the stack anyway).

## How

Give **user-defined Python class instances** whole-program (GC) lifetime in symbolic execution:

- `goto_symext::is_python_gc_object()` — true for `mode == "Python"` struct instances, excluding internal model aggregates (tuples, dicts, `Optional`), mirroring the frontend's `is_excluded_struct_tag`.
- `is_live_variable()` — reports such objects live, so dereferencing an escaped pointer is not flagged expired.
- `pop_frame()` — skips the L2-renaming / value-set teardown for them, so their SSA field-value bindings persist past the frame and reads through the escaped pointer observe the **real** values (not havoc).

It only *skips teardown* — it does **not** switch the objects to `static_lifetime` storage, so distinct call activations keep distinct level1 numbers and recursive/repeated instances never alias.

## Tests

- `regression/python/github_4117_function_internal`: `KNOWNBUG` → `CORE` (now `VERIFICATION SUCCESSFUL`).
- New `regression/python/github_4773_escape_identity` (SUCCESSFUL): mutates the linked object *after* linking and asserts the alias observes it — proves lifetime **and** identity (a copy-on-escape would get this wrong).
- New `regression/python/github_4773_escape_identity_fail` (FAILED): asserts a wrong value, guarding against vacuous success.
- Zero new regressions across all 451 class-using Python regression tests vs a clean baseline; clang-format clean.

## Known follow-up

Escaping **lists** still dangle — a list local is a pointer whose pointee is `__ESBMC_alloca`'d inside the C operational-model `list_create` (mode "C"), so it is not covered by this object-lifetime fix. Narrower in scope; to be tracked separately.


lucasccordeiro:
Pushed two follow-up commits that replace the struct-tag substring matching with an explicit type attribute.

**`6c911a2349` — stamp internal aggregates with a kind attribute**
Tuple/dict/Optional struct types are now stamped with `#python_aggregate` ("tuple"/"dict"/"optional") at their creation sites, and the GC-lifetime, dunder-dispatch, list-object-ref, and binop-membership consumers read it back via helpers in `util/python_types.h`. This removes the brittle `tag.find("dict_")`-style checks (which would misclassify a user class named e.g. `dict_node`) and the goto-symex copy of the frontend tag spellings behind the "keep in sync" comment — single source of truth for those consumers.

**`b47286d321` — remove dead `type_handler::get_tuple_type`**
The json-arg overload had no callers and built an untagged/unstamped tuple struct.

**Scope is intentionally limited.** `is_tuple_type` / `is_dict_type` / the Optional checks / `python_builtins.cpp` keep exact/prefix tag matching: `migrate_type_back` (`util/migrate.cpp:2734-2757`) rebuilds struct types from components + tag + packed only, so the `#`-comment attribute is dropped across an irep2 round-trip while the tag survives. Those predicates can run on post-migration types, so the tag remains the robust key there; the migrated consumers all read pre-migration / symbol-table types where the stamp is intact.

Verified: builds clean; the PR's own tests pass; 129/129 affected-area tests (tuple/optional/divmod/dunder/object/escape/identity) pass. Note: full local suite needs Z3, which isn't in my build — the `--z3`/`--ir` tests should be confirmed in CI.

PR #5427, opened 2026-06-19
  feat(ld): graphical LD frontend, industrial benchmarks, and NuXmv comparison

pierredantas:
## Summary
  
  Three groups of changes for IEC 61131-3 LD verification (ESBMC-PLC+ paper, arXiv 2026):
  
  ### 1. Graphical PLCopen XML frontend
  - New graphical LD (tc6_0201) frontend — parses PLCopen XML and lowers to the textual LD AST
  - **Fix:** XPath now also searches `//pou/actions/action/body/LD` — Beremiz places rungs in SFC action bodies; the previous XPath
  silently dropped them causing vacuous verification
  - **Fix:** heuristic I/O inference for graphical LD without hardware addresses
  - **Feat:** `show_parse()` now prints each rung's contact expression and coil assignment
  
  ### 2. Industrial benchmarks
  tank_level_control, bottle_filling, elevator, traffic_light, water_control, stairs_light, dimmer_light_control (CONTROLLINO
  graphical), beremiz_bacnet / beremiz_traffic_light (Beremiz graphical). Regression tests for water_control and stairs_light.
  
  ### 3. NuXmv BDD/IC3 comparison experiment (`experiments/nuxmv_comparison/`)
  - LD→SMV transpiler, experiment runner, LaTeX table generator, pre-computed results
  - **Key result:** ESBMC k-induction is 400–2,000× faster than NuXmv BDD on programs with integer-valued timer variables; NuXmv times
  out at 120 s on programs with 8 INT timer state variables (4 TON timers)
  
  ## Test plan
  - [ ] `regression/ld/water_control_safe` and `stairs_light_safe` pass with `--k-induction --z3`
  - [ ] Graphical benchmarks parse without dropped rungs
  - [ ] clang-format-11 clean on modified C++ files


src/ld-frontend/parser/plcopen_xml_parser.cpp
```diff
@@ -228,6 +244,210 @@ RungNode PlcopenXmlParser::parse_rung(const void *node_ptr)
...
+        // We append target here so the coil is included.
+        RungNode rung;
+        rung.id = "g" + std::to_string(rung_counter++);
+        rung.loc = {source_file, 0, 0};
+
+        // Emit contacts from the path
+        for (int nid : path)
+        {
+          const GNode &g = nodes.at(nid);
+          if (g.tag != "contact")
```
lucasccordeiro, line 378: **Soundness:** the DFS walks *through* `<block>` nodes (TON/TOF, FBs) but only contacts are emitted, so a rung gated by a timer's `Q` collapses to ignore the timer — a weaker/vacuous model. Same failure class as the XPath fix in this PR. `water_control_original.xml` is contact-only so it isn't exercised, but any graphical program with timers will be silently under-modelled. Suggest emitting a diagnostic / refusing when a non-contact element lies on a rail→coil path.

lucasccordeiro (commented):
Reviewed the C++ frontend changes (rest is benchmark/experiment data). Two PR-level notes:

- **Test plan vs. diff mismatch:** the body says "Regression tests for water_control and stairs_light" and the test plan checks `regression/ld/water_control_safe` / `stairs_light_safe`, but the diff contains no `regression/` files — only `benchmarks/`. CI won't gate these. Add the regression tests or correct the claim.
- Nit: PR title has leading whitespace (`  feat(ld): …`).

Inline comments below.

src/ld-frontend/parser/plcopen_xml_parser.cpp
```diff
@@ -228,6 +244,210 @@ RungNode PlcopenXmlParser::parse_rung(const void *node_ptr)
...
+      if (src >= 0 && nodes.count(src))
+        nodes[src].feeds.push_back(lid);
+    }
+  }
+
+  // Step 4: DFS from every leftPowerRail to every coil.
+  // Each path becomes one RungNode with one contact per step + one coil.
+  int rung_counter = 0;
+
+  std::function<void(int, int, std::vector<int> &, std::set<int> &)> dfs =
```
lucasccordeiro, line 346: Enumerating *all simple paths* rail→coil is exponential in the number of parallel branches. Fine for these small benchmarks, but worth a path/size guard (or a comment noting the bound) before this meets larger vendor exports.

src/ld-frontend/ld_language.cpp
```diff
@@ -159,7 +159,31 @@ void ld_languaget::show_parse(std::ostream &out)
...
+          if (!first)
+            out << " && ";
+          first = false;
+          if (elem.contact.kind == ContactKind::NormallyClosed)
+            out << "!";
+          out << elem.contact.variable;
+        }
+        else if (elem.kind == RungElementKind::Coil)
+        {
+          coil_var = elem.coil.variable;
```
lucasccordeiro: Minor (display only): with multiple coils in one rung, `coil_var`/`coil_kind` are overwritten so only the last prints; a coil-less rung prints a bare `->`.

src/ld-frontend/parser/plcopen_xml_parser.cpp
pierredantas: The continue at line 360 silently drops any non-contact node — so a path leftRail → contact → TON(Q) → coil gets modeled as contact → coil, removing the timer constraint. That is a weaker, vacuous model. All evaluated graphical programs (B1–B3, C1) are contact-only on their rail→coil paths, so this didn't surface in testing. I'll fix this by refusing the path rather than silently skipping the unsupported element — returning early with a diagnostic when a non-contact, non-leftRail node appears on the path. Full FB/timer output support is left as future work and will be noted as a limitation in the paper.

src/ld-frontend/parser/plcopen_xml_parser.cpp
pierredantas: The all-simple-paths DFS is exponential in the number of parallel branches and is only tractable for the small benchmarks evaluated here. I'll add a comment at the DFS call site, noting the complexity bound and flagging it as a known limitation for larger vendor exports. A path/node-count guard can be added as a follow-up if needed.

src/ld-frontend/ld_language.cpp
pierredantas: Confirmed. With multiple coils in a rung, coil_var and coil_kind are overwritten on each iteration, so only the last one prints. I'll fix the display loop to accumulate all coils and print each one separately on its own → line. This is display-only and doesn't affect verification.

PR #5497, opened 2026-06-20
[python] Fold str % formatting instead of crashing the SMT backend

### LINKED ISSUE #5495: [python] str % formatting mis-lowered as pointer modulo, crashes SMT backend (SIGSEGV)
## Summary

`str % args` (printf-style string formatting) is mis-lowered as **numeric modulo on the format string's pointer**, which crashes ESBMC with a **SIGSEGV** in the SMT backend (`bitwuzla_mk_term2`) when the formatted string is consumed (e.g. by `len()` or `==`).

## Reproduce

```python
def main() -> None:
    s = "%d" % 5
    assert len(s) == 1


main()
```

```sh
esbmc main.py --incremental-bmc
```

Result: `Segmentation fault` (exit 139). CPython: `"%d" % 5 == "5"`, `len` is 1 → expected `VERIFICATION SUCCESSFUL`.

## Root cause

The Python frontend's binary-operator converter treats `%` (AST `Mod`) purely as arithmetic. For `"%d" % 5` it builds `&"%d"[0] % 5` — pointer-modulo-int — and types the result variable as `char[0]`. The GOTO is:

```
DECL signed char [0] s;
ASSIGN s = &{ 37, 100, 0 }[0] % 5;     // &"%d"[0] % 5
FUNCTION_CALL: return_value$_strlen$1 = strlen(&s[0])
```

`strlen` over the bogus pointer arithmetic feeds a malformed term to bitwuzla, which dereferences a pointer-sized value that actually contains ASCII bytes (`0x5f656c7075743a42` = `"B:tuple_…"`) and crashes.

Creating the string alone does not crash; consuming it (`len(s)`, `s == "..."`) does. `print(s)` does not. Affects single-arg (`"%d" % 5`), tuple-arg (`"%d-%d" % (3,4)`), and `%s`/`%x`/etc.

## Expected

`str % args` should be modelled as string formatting (at minimum for constant format + constant args), or rejected with a clean diagnostic — never crash.

## Environment

ESBMC 8.3.0 (master), Bitwuzla backend, aarch64 macOS.

lucasccordeiro:
## What

`str % args` (printf-style string formatting) was mis-lowered as **numeric modulo on the format string's pointer** — `&"%d"[0] % 5` — typing the result `char[0]`. Consuming the formatted string with `len()` or `==` then fed a malformed term to the SMT backend and **crashed ESBMC with a SIGSEGV** (`bitwuzla_mk_term2`).

```python
def main() -> None:
    s = "%d" % 5
    assert len(s) == 1   # Segmentation fault before this PR


main()
```

## Fix

In `handle_array_operations` (`converter/converter_binop.cpp`), intercept `Mod` with a **string** LHS and constant-fold the common printf-style conversions into a string literal, re-entering conversion through a synthesised `Constant` node so the result flows through the identical path as a string literal:

- Supported: `%d %i %s %x %X %o %c %%`, negative literals (`UnaryOp(USub)`), and `%s` of a bool (`True`/`False`, matching CPython `str()`).
- Non-constant format/args, floats, widths/precision/flags → a clean `runtime_error` diagnostic instead of a crash.

The gate is `type_utils::is_string_type(lhs)` (char-element array), **not** the structural 8-bit-array test — so `np.int8`/`np.uint8` array `% scalar` keeps flowing through the numeric path untouched.

## Testing

- New regression tests: `regression/python/github_5495` (CORE, SUCCESSFUL — exercises every supported conversion, bool `%s`, `%%`, negative, and `len()` via a bound variable) and `regression/python/github_5495_fail` (CORE, FAILED — wrong-content assertion).
- All folded results verified against CPython; wrong assertions correctly report `VERIFICATION FAILED`; non-constant args produce a graceful error rather than a crash.
- Broad python regression sweep: no non-environmental regressions (numeric `%` and numpy array `%` paths unchanged).

## Known limitations (out of scope; pre-existing)

- `str` **variable** `% args` (non-literal format string) is not folded.
- Inline `len("%d" % 5)` (no intermediate variable) is a separate pre-existing len-handler limitation for inline non-symbol char-array args.

Fixes #5495.

### DIFF
--- a/regression/python/github_5495/main.py
+++ b/regression/python/github_5495/main.py
@@ -1,0 +1,23 @@
+# Regression for #5495: `str % args` (printf-style formatting) was mis-lowered
+# as numeric modulo on the format string's pointer, crashing the SMT backend
+# (SIGSEGV) when the result was consumed by len()/==. Constant format + constant
+# args are now folded to the correct string literal. The `==` checks validate
+# exact content (and therefore length); len() is exercised via a bound variable.
+def main() -> None:
+    assert "%d" % 5 == "5"
+    assert "%i" % 7 == "7"
+    assert "%d-%d" % (3, 4) == "3-4"
+    assert "%s" % "hi" == "hi"
+    assert "val=%s" % 42 == "val=42"
+    assert "%s" % True == "True"
+    assert "%x" % 255 == "ff"
+    assert "%X" % 255 == "FF"
+    assert "%o" % 8 == "10"
+    assert "%c" % 65 == "A"
+    assert "%d%%" % 50 == "50%"
+    assert "%d" % (-7) == "-7"
+    s = "%d" % 5
+    assert len(s) == 1
+
+
+main()
--- a/regression/python/github_5495/test.desc
+++ b/regression/python/github_5495/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_5495_fail/main.py
+++ b/regression/python/github_5495_fail/main.py
@@ -1,0 +1,8 @@
+# Negative variant of github_5495: the formatted string is "3-4", so asserting
+# it equals "3-5" must fail.
+def main() -> None:
+    s = "%d-%d" % (3, 4)
+    assert s == "3-5"
+
+
+main()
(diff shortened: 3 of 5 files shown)


lucasccordeiro:
Superseded by #5502, which was stacked on this branch and has now been merged — its py_percent_format helper and the str-format handling are already in master. Closing as redundant.

PR #5524, opened 2026-06-22
[Witness] fix function_enter waypoint handling and add reachsafety violation-witness regression

XLiZHI:
Replace the incorrect pc++ skip with assume(false) for function_enter AVOID waypoints on both real FUNCTION_CALLs and inlined nondet calls. Unify the two handlers into a single symex_witness_function_enter, and extract all witness-steering functions into builtin_functions/witness.cpp. 

Add 100 SV-COMP reachability violation-witness fixtures under witnesses_validate/reachsafety-violationwitness, covering for/while/if/ switch/ternary/functions benchmarks.


XLiZHI:
- ReachSafety-ViolationWitnesses

```
==> esbmc-violation-v2 <==

  correct:             100
    correct true:       59
    correct false:      41
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:               0
  Score:               159 (max: 159)

https://github.com/esbmc/esbmc/actions/runs/27946221641
```

Issue #5142, opened 2026-06-05
[SV-COMP] false alarm on no-overflow: m0_drivers-media-rc-imon--32_7a--a9e7fb5


lucasccordeiro:
## Wrong result in SV-COMP 26 run

**False alarm**: ESBMC reports `VERIFICATION FAILED` (FALSE) on a benchmark whose ground truth is TRUE — a spurious counterexample on a correct program (precision/incompleteness bug).

| Field | Value |
|---|---|
| Property | `no-overflow` |
| Expected verdict | `true` |
| ESBMC verdict | `false(no-overflow)` |
| Benchmark | [`c/ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5.yml`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5.yml) ([source](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5.i)) |
| Run | [SV-COMP 26 benchexec run](https://github.com/esbmc/esbmc/actions/runs/26947362369) |
| ESBMC version | 8.3.0 (commit `0e3612575faf49c0851eb305dc00b27fbf4b4ebd`) |
| Strategy | `kinduction` |
| Related umbrella issue(s) | #2513 |

## Reproducer

Wrapper invocation (matches the SV-COMP run):

```
./esbmc-wrapper.py -p sv-benchmarks/c/properties/no-overflow.prp -s kinduction --arch 64 sv-benchmarks/c/ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5.i
```

Underlying ESBMC command:

```
./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check sv-benchmarks/c/ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5.i --64 --witness-output witness --no-pointer-check --no-bounds-check --overflow-check --no-assertions --incremental-bmc
```

[...]

PR #5530, opened 2026-06-22
[svcomp] Replace ESBMC_SVCOMP build macro with runtime --sv-comp flag

lucasccordeiro:
Replace the compile-time `ESBMC_SVCOMP` CMake macro with a runtime `--sv-comp` option, so a single normal binary behaves as SV-COMP when the flag is passed instead of needing a separate build. The three frontend sites become `config.options` checks; the `io.c` fopen/fclose fork branches on a new `__ESBMC_sv_comp()` symex intrinsic since the model is compiled at build time. Both competition wrappers now pass `--sv-comp`. Diagnostic-only and sound — verification semantics are unchanged. Refs #5142.


mikhailramalho (approved):
lgtm.

The fopen/fclose is weird tho, do tools like valgrind/asan report memory leaks when a file handler is not closed?

lucasccordeiro:
Checked it: ASan/LSan does **not** flag an unclosed \`fopen\` — glibc links every open \`FILE\` into the global \`_IO_list_all\` chain, so it stays reachable (a deliberate \`malloc\` leak still gets reported, so the checker is on). 

That's why SV-COMP mode returns the \`(void *)1\` sentinel and skips \`free\`: a leaked handle is not a valid-memtrack violation there, so \`malloc\`/\`free\` would raise a false alarm. The default model keeps \`malloc\`/\`free\` to track the resource.

Issue #4237, opened 2026-05-01
[SMT encoding] value-init of struct inheriting class (Derived d{}) crashes to_solver_smt_ast


lucasccordeiro:
## Summary

Value-initialising a `struct` that inherits from a `class` via `{}` crashes ESBMC's SMT encoding with `Assertion failed: (r), function to_solver_smt_ast, file smt_ast.h, line 111`.

Replacing `Derived d{}` with `Derived d;` (default-initialisation) avoids the crash and produces correct output.

## Minimal reproducer

```cpp
#include <cstdint>
extern "C" { uint8_t nondet_u8(); }

class Base {
    uint8_t _x{};
protected:
    void check(uint8_t i) { __ESBMC_assert(i < _x, "OOB"); }
public:
    Base() = default;
};

struct Derived : Base { using Base::check; };

int main() {
    Derived d{};   // crash; change to `Derived d;` → VERIFICATION FAILED correctly
    uint8_t i = nondet_u8();
    __ESBMC_assume(i >= 4);
    d.check(i);
    return 0;
}
```

```
esbmc --std c++20 repro.cpp
```

**Expected**: `VERIFICATION FAILED` ("OOB")  
**Actual**: `Assertion failed: (r), function to_solver_smt_ast, file smt_ast.h, line 111`

## Notes

- Replacing `Derived d{}` with `Derived d;` produces `VERIFICATION FAILED` correctly.
- The crash is in the SMT encoding phase (bitwuzla backend), after symex completes successfully.
- Found while verifying NVIDIA/OpenSMA: the production harness used `VerifControl ctrl{}` where `struct VerifControl : platforms::Control` and `class platforms::Control`. Changing to `VerifControl ctrl;` unblocks the verification.
- Related to (but distinct from): #4232 (fixed by #4233), #4234 (fixed by #4235).

PR #5552, opened 2026-06-22
[cpp] Multiple-inheritance & polymorphic layout: nested base subobjects

mikhailramalho:
## Summary

Fixes a class of soundness and crash bugs in the C++ frontend's handling of
multiple inheritance (MI) and polymorphic class layout. Base classes are now
laid out as **nested base-subobjects at their correct Itanium offsets** instead
of flattened by name, which lets ESBMC reason correctly about:

- reads/writes through a reference or pointer to a **non-first base** (non-zero
  offset) in MI,
- a non-virtual **diamond** holding two distinct subobjects of a shared base,
- two bases with a **same-named field**,
- `dynamic_cast` **down-casts and sibling cross-casts** across non-zero base
  offsets (result re-offset to the correct subobject, not a plain reinterpret),
- virtual **dispatch through a non-primary base** under Itanium primary-base
  vtable sharing.

## What changed

- **Layout**: each direct base becomes one `<base>::@base` nested subobject
  component, placed at its Clang-reported offset via padding. Derived-to-base
  conversions lower to a nested member-access chain rather than an offset-0
  reinterpret.
- **Vtable / dispatch**: Itanium primary-base vtable sharing (the polymorphic
  first base's vptr is reused at offset 0); a merged vtable so the shared slot
  resolves both inherited and the derived's own virtuals; dispatch and
  `dynamic_cast` read the merged table only through the primary (offset-0) vptr,
  while non-primary bases keep their own thunk table.
- **Base ctor calls**: a base constructor receives the address of the
  `<base>::@base` subobject, so a non-zero-offset base is constructed at the
  right bytes.
- **Aggregate init**: a base initializer is now placed as one struct-typed value
  into the `@base` component instead of having its fields flattened into the
  derived's component list — this fixes an SMT sort mismatch crash on
  value-initializing a struct that inherits from a class (#4232, #4237).

## Tests

New/flipped regression tests under `regression/esbmc-cpp/inheritance/`:
`mi_base_subobject_layout`, `mi_dynamic_cast`, `mi_same_named_base_fields`,
`mi_nonvirtual_diamond_field` (+ `_fail` negative companions). The
`github_4232`/`github_4237` aggregate-init tests and their `_pass` variants now
pass.

`inheritance` + `polymorphism_bringup` suites are green (173/173). Full
`esbmc-cpp/cpp` is green except two solver-environment timeouts (`ch9_7`,
`ch21_7`) that pass under Bitwuzla but time out on a Z3-only build (not code
regressions — confirmed by A/B against the Bitwuzla baseline).

## Known limitations (marked KNOWNBUG, tracked as follow-ups)

These pass on master but regressed under the new layout and are each a separate
sub-project, out of scope here:

- `cast_base_to_derived_fail` — invalid-downcast-then-virtual-call UB detection
  is lost because a single-inheritance derived shares the base's offset-0 vptr;
  restoring it needs receiver-frame RTTI checking.
- `github_2544` — **virtual inheritance** layout (needs `getVBaseClassOffset` /
  shared virtual-base subobjects / VTT construction).
- `github_4191_nonptr` — `__builtin_bit_cast`, unrelated to inheritance.


lucasccordeiro:
Heads up: while triaging C++ KNOWNBUGs I found a soundness gap in `symex_free()` in this same file/region (`memory_alloc.cpp`) that this PR's diff also touches — filed as #6263. The base-subobject offset relaxation (added by #6181 for #1866/#3894) doesn't check whether a virtual destructor actually exists in the hierarchy, so `delete` through a non-virtual-destructor base pointer with a mismatched dynamic type is now silently accepted (confirmed as a real bad-free via ASan). This PR's own version of the check looks even more permissive (skips the offset claim entirely for any C++ delete). Flagging here rather than opening a competing PR against the same code path, since you're actively working this area — let me know if you'd like a hand once this PR's direction settles.

regression/esbmc-cpp/cpp/cast_base_to_derived_fail/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: this PR breaks a passing test. Fix the layout change rather than downgrading the test.

regression/esbmc-cpp/cpp/github_2544/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: a previously passing regression is disabled by this PR. It must keep passing.

regression/esbmc-cpp/cpp/github_4191_nonptr/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: CORE to KNOWNBUG: another passing test disabled here. Downgrading hides the regression instead of fixing it.

src/goto-symex/builtin_functions/memory_alloc.cpp
```diff
@@ -732,6 +732,17 @@ void goto_symext::symex_free(const expr2tc &expr)
...
 
+    // C++ `delete` may be applied to a pointer to a base-class subobject that
+    // sits at a non-zero offset within the complete (most-derived) object —
+    // e.g. `Base2 *o = new Derived(); delete o;` where the Base2 subobject is
+    // not at offset 0. This is well-defined (the deleting destructor adjusts
+    // to the complete object via offset-to-top); the whole allocation is
+    // freed regardless of which subobject pointer is held. So the strict
+    // zero-offset requirement only applies to raw free(): a non-zero offset
+    // into a malloc'd block (free(p + k)) is still a real bug.
+    const bool is_cpp_delete =
```
lucasccordeiro, line 743: `delete` through a non-zero-offset base pointer is only defined when the destructor is virtual; unconditionally dropping the check hides real UB. Gate on a polymorphic static type.

src/goto-symex/builtin_functions/memory_alloc.cpp
```diff
@@ -741,7 +752,11 @@ void goto_symext::symex_free(const expr2tc &expr)
       expr2tc offset = item.offset;
       expr2tc eq = equality2tc(offset, gen_ulong(0));
       g.guard_expr(eq);
-      if (options.get_bool_option("conv-assert-to-assume"))
+      if (is_cpp_delete)
```
lucasccordeiro, line 755: Empty branch holding only a comment. Invert the condition so the check is skipped without a no-op block.

lucasccordeiro (changes requested):

lucasccordeiro (commented):
Master already implements nested base subobjects (`@base@<id>` components, `is_base_subobject`, virtual-base fallback, `docs/design/cpp-multiple-inheritance-subobjects.md`, landed via #6822/#7241). This branch is 661 commits behind and re-implements the same layout with different component names and flag keys. Rebase and re-scope onto master's layout before further review.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2492,61 +2532,106 @@ bool clang_cpp_convertert::get_base_map(
...
+  {
+    uint64_t offset;
+    std::string class_id;
+    const clang::CXXRecordDecl *decl;
+  };
+  std::vector<base_entryt> ordered;
   for (const auto &base : map)
   {
-    std::string class_id = base.first;
+    uint64_t off = layout.getBaseClassOffset(base.second).getQuantity();
```
lucasccordeiro, line 2578: `getBaseClassOffset` asserts for virtual bases (they live in `VBaseOffsets`) and silently returns 0 in release builds. Filter `base.isVirtual()` out in `get_base_map`.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2492,61 +2532,106 @@ bool clang_cpp_convertert::get_base_map(
...
+    ordered.push_back({off, base.first, base.second});
+  }
+  std::sort(
```
lucasccordeiro, line 2581: `std::sort` is unstable, so two bases sharing an offset (empty-base optimisation) get an unspecified layout order. Use `std::stable_sort`.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2492,61 +2532,106 @@ bool clang_cpp_convertert::get_base_map(
...
+      base_subobject_name(be.class_id),
+      symbol_typet(be.class_id));
+    base_comp.set("from_base", true);
+    base_comp.set("#is_base_subobject", true);
+    base_comp.set("#base_class_id", be.class_id);
+    stype.components().push_back(base_comp);
+
+    // Advance the running offset past this subobject (Itanium dsize/size).
+    cur_offset +=
+      ASTContext->getASTRecordLayout(be.decl).getSize().getQuantity();
```
lucasccordeiro, line 2625: `getSize()` includes tail padding, so the next base's Clang offset can fall below `cur_offset` and padding is then silently skipped. Use `getDataSize()` and reject overlap explicitly.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -210,6 +210,168 @@ symbolt *clang_cpp_convertert::add_vtable_type_symbol(
...
+      merged_comps.push_back(d_ts.components()[it->second]);
+      merged_vals.push_back(d_vals.operands()[it->second]);
+    }
+    else if (i < prim_vals.operands().size())
+    {
+      merged_comps.push_back(c);
+      merged_vals.push_back(prim_vals.operands()[i]);
+    }
+    else
+      return; // source view desynced; bail safely
```
lucasccordeiro, line 319: Returning silently leaves the merged vtable unbuilt, so dispatch through the shared vptr falls back to the base view and can miss overrides. Log an error here.

src/clang-cpp-frontend/clang_cpp_adjust_code_gen.cpp
```diff
@@ -139,21 +179,70 @@ exprt clang_cpp_adjust::gen_vptr_init_rhs(
...
+  }
 
-  // get the corresponding vtable variable symbol
-  std::string vtable_var_id = comp.type().subtype().identifier().as_string() +
-                              "@" + ctor_type.get("#member_name").as_string();
   const symbolt *vtable_var_symb = namespacet(context).lookup(vtable_var_id);
   assert(vtable_var_symb);
 
-  // get the operand for address_of expr as in `&<vtable_struct_variable>`
+  (void)retargeted;
```
lucasccordeiro, line 210: `retargeted` is assigned but never read. Delete the variable and this `(void)` cast.

src/clang-c-frontend/clang_c_convert.cpp
```diff
@@ -3723,6 +3705,59 @@ void clang_c_convertert::rewrite_builtin_ref(
...
+    }
+
+    std::string base_name, base_id;
+    get_decl_name(*base_decl, base_name, base_id);
+
+    // Step into the nested base-subobject member of this hop.
+    typet base_struct_type = symbol_typet(base_id);
+    lval = member_exprt(lval, base_id + "::@base", base_struct_type);
+
+    cur_qt = spec->getType();
```
lucasccordeiro, line 3746: `cur_qt` is never read after the loop, and `base_name` on line 3739 is unused. Remove both.

regression/esbmc-cpp/inheritance/mi_same_named_base_fields/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
```
lucasccordeiro, line 1: No `_fail` companion. Per CLAUDE.md regression tests come in pairs; add `mi_same_named_base_fields_fail` pinning `^VERIFICATION FAILED$` over the same construct.

regression/esbmc-cpp/inheritance/mi_nonvirtual_diamond_field/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
```
lucasccordeiro, line 1: No `_fail` counterpart. Add `mi_nonvirtual_diamond_field_fail` pinning `^VERIFICATION FAILED$` over the same diamond.

Issue #5138, opened 2026-06-05
[SV-COMP] false alarm on valid-memsafety: comm_3args_ok


lucasccordeiro:
## Wrong result in SV-COMP 26 run

**False alarm**: ESBMC reports `VERIFICATION FAILED` (FALSE) on a benchmark whose ground truth is TRUE — a spurious counterexample on a correct program (precision/incompleteness bug).

| Field | Value |
|---|---|
| Property | `valid-memsafety` |
| Expected verdict | `true` |
| ESBMC verdict | `false(valid-memsafety)` |
| Benchmark | [`c/coreutils-v8.31/comm_3args_ok.yml`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/coreutils-v8.31/comm_3args_ok.yml) ([source](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/coreutils-v8.31/comm_3args_ok.c)) |
| Run | [SV-COMP 26 benchexec run](https://github.com/esbmc/esbmc/actions/runs/26947362369) |
| ESBMC version | 8.3.0 (commit `0e3612575faf49c0851eb305dc00b27fbf4b4ebd`) |
| Strategy | `kinduction` |
| Related umbrella issue(s) | #2513 |

## Reproducer

Wrapper invocation (matches the SV-COMP run):

```
./esbmc-wrapper.py -p sv-benchmarks/c/properties/valid-memsafety.prp -s kinduction --arch 64 sv-benchmarks/c/coreutils-v8.31/comm_3args_ok.c
```

Underlying ESBMC command:

```
./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check sv-benchmarks/c/coreutils-v8.31/comm_3args_ok.c --64 --witness-output witness --memory-leak-check --no-reachable-memory-leak --no-assertions --no-abnormal-memory-leak --malloc-zero-is-null --incremental-bmc
```

To reproduce locally, fetch the benchmark from the [sv-benchmarks repo](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/coreutils-v8.31/comm_3args_ok.c).

## Spurious counterexample

```
Violated property:
[...]

PR #5564, opened 2026-06-22
[symex] treat stack-reachable heap as live in memory-leak check

lucasccordeiro:
Roots the `--no-reachable-memory-leak` reachability scan at the pointer locals of every active call frame (not just static globals) and chases pointers stored in constant-size heap buffers by byte-stitching, so heap still reachable from a live stack local at `exit()`/`abort()` is no longer reported as a forgotten-memory leak. Sound: each pointer is L2-renamed to its current value before `same_object`, and compiler temporaries are excluded. Two regression tests included. Refs #5138 — fixes the memory-leak false-alarm layer; `comm_3args_ok` also needs the unrelated `fclose`/`strcmp` model fixes tracked in #5565.

### DIFF
--- a/regression/esbmc/github_5138/main.c
+++ b/regression/esbmc/github_5138/main.c
@@ -1,0 +1,40 @@
+#include <stdlib.h>
+
+extern char nondet_char(void);
+
+static void cleanup(void)
+{
+}
+
+static void terminate(char **a, int n)
+{
+  (void)a;
+  (void)n;
+  exit(0);
+}
+
+/* Models the SV-COMP entry harness of coreutils/comm_3args_ok: an argv-style
+ * array of char* is stored into a flat malloc'd byte buffer, and the program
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
This PR (100s): 

```
  correct:           22882
    correct true:    15324
    correct false:    7558
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           13711
  Score:             38094 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/28051490182.
```

Master (100s):

```
  correct:           23196
    correct true:    15321
    correct false:    7875
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           13397
  Score:             38405 (max: 61270)

https://github.com/esbmc/esbmc/actions/runs/27852831114/job/82436434901

```

PR #5566, opened 2026-06-23
[witness] add termination violation-witness validation

XLiZHI:
Implements SV-COMP 2.1 termination violation-witness validation. 

`branching(cycle)` waypoints are handled via a `cycle_branch_map` (line → direction) populated on first hit and queried in O(1) on subsequent iterations, making the constraint persistent across all loop unwinds. 

`assumption(cycle)` is handled through the existing `__ESBMC_witness_assume` injection. Termination per-loop and abort-call markers are suppressed in `symex_witness_assert` so k-induction's inductive step runs freely while the witness forces the cycle path. 

Adds 9 regression tests from SV-COMP `validation-crafted` and `termination-restricted-15`, and 100 reachability violation-witness fixtures under `witnesses_validate/reachsafety-violationwitness/`.


XLiZHI:
termination-violationwitness

```
==> esbmc-violation-v2 <==

Statistics:             17 Files
  correct:               9
    correct true:        4
    correct false:       5
  incorrect:             2
    incorrect true:      0
    incorrect false:     2
  unknown:               6
  Score:               -19 (max: 26)
```

PR #5616, opened 2026-06-25
[python] V.4 B.1: resolve symbol_type member/index sources in the adjuster

lucasccordeiro:
Phase **B.1** of the V.1k (b) IREP2-native Python adjuster (`docs/irep2-migration.md`, "V.1k (b)-adjuster"). **Stacked on #5615 (B.0)** — review/merge that first; this PR's base auto-retargets to master once B.0 lands.

Gives `python_adjust::adjust_expr` its first real behaviour: when a `member2t`/`index2t` source carries a transient `symbol_type2t` (the state the V.1k construction-assert relaxation permits), follow it to the resolved `struct_type2t`/`array_type2t` via `namespacet::follow` and retype the source in place (`with_type` keeps `expr2t::type` immutable), rebuilding the immutable member/index node. Operands are visited first, so nested sources (`self.b.a`) resolve inner-to-outer.

**Design note (refined during implementation).** A `member2t`/`index2t` cannot be constructed over a raw pointer source — the construction assert rejects `pointer_id` — so the transient form is always a `symbol_type2t`, carried either on the `symbol2t` instance or on a `dereference2t` of a `pointer→tag-Cls` instance pointer. The single symbol_type branch handles both, which is the IREP2-native equivalent of clang_c_adjust's symbol-type completion + pointer auto-deref.

**Still dead-but-tested:** not wired into `python_language.cpp` (B.2 adds the `--python-irep2-adjust` flag), so ESBMC behaviour is unchanged by construction. `python_adjust_test` grows four resolution cases (direct symbol_type source, dereference source, nested chain inner-to-outer, already-resolved idempotence) — **15 assertions across 7 cases pass**; esbmc links and the nested-attr/github_4117/dataclass fixture smoke set is unaffected.


lucasccordeiro:
Superseded by #5654, which landed this branch's B.1 adjuster resolution to master (`d087f8ab13`, cherry-picked + cleaned up). Merging/rebasing onto current master now yields an empty diff — there is nothing left to merge (the apparent #5651 reverts come back in via the merge). Closing as fully redundant.

PR #5625, opened 2026-06-26
[python] V.1k(b) B.0: IREP2-native Python adjuster skeleton

lucasccordeiro:
## What

First commit of **Part V Phase V.1k design (b)** (`docs/irep2-migration.md`) — the
IREP2-native Python adjuster that will eventually replace the legacy
`clang_cpp_adjust` round-trip on Python output.

This PR ships sub-phase **B.0 — pass skeleton, dead-but-tested**:

- New `src/python-frontend/python_adjust.{h,cpp}`. It mirrors
  `clang_c_adjust::adjust()`: snapshots the symbol list, skips `is_type`
  symbols, and walks each remaining `symbol.get_value2()` via a `const`
  `adjust_expr(const expr2tc&)` that descends every operand through
  `foreach_operand` and **mutates nothing**. The `const` signature makes the
  no-op a compile-time guarantee.
- Unit test `unit/python-frontend/python_adjust_test.cpp` pinning the two B.0
  contracts: *visits-every-node* (count over `add(mul(1,2),3)` == 5) and the
  *`is_type`-skip no-op* (both a function symbol and a type symbol carrying a
  value are left byte-identical; `adjust()` returns `false`).

## Why

V.1k is the keystone that unblocks the Python converter's expression-construction
migration (F-P11 general-operand wall) and W3 attribute carriage. B.0 lands the
traversal scaffold so B.1 (member/index `symbol_type2t` → resolved struct/array
source following) extends a verified walk. Shipping it dead first keeps the risky
resolve-then-build logic off every execution path until it is wired in behind a
flag at B.2.

## Verification

- Not wired into `python_language.cpp` — on **no execution path**, so Python and
  `esbmc-cpp` verdicts are unchanged by construction.
- Full `esbmc` binary builds clean; Python smoke test verifies SUCCESSFUL.
- `python_adjust_test` passes (both contracts). Pre-existing local `--z3`
  regression failures are environmental (Bitwuzla-only local build), unrelated
  to this change.

Refs #4715.


lucasccordeiro:
Superseded by the V.4 lineage — the IREP2-native Python adjuster skeleton landed on master via #5615. Closing to avoid a duplicate skeleton (this B.0 re-adds the same `src/python-frontend/python_adjust.{h,cpp}` that #5615 already merged). The canonical B.1–B.2 continuation is #5654 (off `feat/v4-b1-python-adjust-resolve`).

PR #5627, opened 2026-06-26
[python] V.1k(b) B.2: wire python_adjust behind --python-irep2-adjust

lucasccordeiro:
## What

Part V Phase V.1k (b) sub-phase **B.2 — wire in behind a flag**
(`docs/irep2-migration.md`). Stacked on #5626 (B.1) → #5625 (B.0).

- **New `--python-irep2-adjust` option** (`options.cpp`, IREP2-migration group,
  default **off**).
- **`python_languaget::typecheck`** runs `python_adjust` *after* the legacy
  `clang_cpp_adjust` when the flag is set.
- **`python_adjust::adjust()` is scoped to Python-mode symbols** — the C
  operational-model bodies stay on the legacy path (V.1k RV-adj4); only the Python
  converter emits the pre-adjust `symbol_type2t` member/index sources this pass
  resolves.

## Why

This lands the toggle that B.3 will exercise (the converter starts emitting
member/index nodes over `symbol_type2t` sources pre-adjust, and this pass resolves
them). Wiring it off-by-default keeps every execution path byte-identical until
B.3 flips the converter, while making the new path runnable end-to-end now.

## Verification

- **Flag off ⇒ byte-identical:** the conditional is the only new runtime behavior
  and it is not taken by default.
- **Flag on ⇒ byte-identical on today's corpus:** the converter does not yet emit
  pre-adjust `symbol_type2t` members, so the pass resolves nothing. Verified
  manually: `class Point` attribute access gives `VERIFICATION SUCCESSFUL` with the
  flag both on and off, and a real violation gives `VERIFICATION FAILED` with the
  flag on (not silently passing).
- **Regression tests** `regression/python/python_irep2_adjust{,_fail}` pin both
  verdicts under the flag (CPython sanity via `check_python_tests.sh`: pass exits 0,
  fail exits 1).
- **Unit test** pins the Python-mode scoping (a `mode == "C"` symbol's member source
  is left unresolved). `python_adjust_test`: 8 cases / 30 assertions.
- Existing class-based Python tests pass unchanged (flag-off path).
- Code-reviewer agent: "Good — ready for commit"; no blocking findings.

Refs #4715. Stacked on #5626.

### DIFF
--- a/docs/irep2-migration.md
+++ b/docs/irep2-migration.md
@@ -3400,7 +3400,7 @@ build-IREP2-then-back-migrate pattern file 1 used for non-member expressions
 
 #### V.1k (b)-adjuster — execution scoping (post-V.4.4b: the precondition is now met)
 
-> **Status: B.0 + B.1 landed (2026-06-26); B.2 next.** The V.1k breakthrough above deferred
+> **Status: B.0 + B.1 + B.2 landed (2026-06-26); B.3 next.** The V.1k breakthrough above deferred
 > the (b) IREP2-native adjuster to "V.4+" on the grounds that *a separate adjuster
 > has no IREP2 to operate on until function bodies are IREP2*. **That precondition is
 > now satisfied:** V.4.4b landed and the IREP2 body round-trip is the only body path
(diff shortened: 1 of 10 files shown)


lucasccordeiro:
Superseded by the V.4 lineage, which is master's plan-of-record. The canonical B.0–B.2 implementation is #5615 (skeleton, merged) + #5654 (B.1–B.2 resolution + `--python-irep2-adjust` flag, off `feat/v4-b1-python-adjust-resolve`). This V.1k(b) attempt diverges from it — different regression tests, and a service-based B.3 conclusion (#5629) that the V.4 before-placement finding rejects (running the pass while `clang_cpp_adjust` still resolves sources double-resolves; B.3 is not separable from B.5). Closing as not-to-be-merged. The docs were reconciled to the canonical state in #5653.

PR #5660, opened 2026-06-28
[clang-frontend] apply -Wno-int-conversion universally

lucasccordeiro:
clang 15+ promotes -Wint-conversion to a hard error by default, rejecting GCC-acceptable implicit int-to-pointer conversions, but ESBMC carried -Wno-int-conversion only inside the --sv-comp block of build_compiler_args, so a plain esbmc run still hard-errored at parse time. Move -Wno-int-conversion to the universal section beside -Wno-incompatible-pointer-types; this is a diagnostic-level change only, matching GCC, with conversion semantics unchanged.

Refs #5142.

### DIFF
--- a/docs/svcomp-issue-triage.md
+++ b/docs/svcomp-issue-triage.md
@@ -1385,3 +1385,71 @@ no-overflow precision layer:
 8. **#5565 residual** — `ESBMC_SVCOMP`-gated `fclose(stdout)` static-stream free + `strcmp` argv-string
    pointer-validity; reproduce under a `-DESBMC_SVCOMP=On` build.
 9. **Close-outs pending CI:** #1470, #4427; witness end-to-end validation for #1471/#1492/#4611.
+
+---
+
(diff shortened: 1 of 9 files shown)


lucasccordeiro:
Closing without merging. On reflection this trades away the wrong compatibility target: ESBMC's C frontend *is* clang, so a plain `esbmc foo.c` run should surface clang's own default diagnostics (including clang 15+'s hard error on implicit int-pointer conversion) rather than silently downgrading it to match GCC's warn-only behaviour. The `--sv-comp` gate exists precisely to accommodate CIL-preprocessed kernel benchmarks that rely on this GCC-ism -- that is a SV-COMP-specific accommodation, not a general default we want for all users.

Keeping master's current behaviour (relaxation scoped to `--sv-comp` only). A follow-up doc-only PR records this decision in the SV-COMP triage doc so it is not re-proposed in a later pass.

Issue #4611, opened 2026-05-18
Witness GraphML: review returnFromFunction key + duplicate-node generation


mikhailramalho:
## Background

The long-defunct branch \`adjust_type\` (Mikhail R. Gadelha, 2022-11-02) carried four GraphML witness-generation tweaks that haven't landed:

1. **\`returnFromFunction\` → \`returnFrom\`** (\`src/goto-symex/witnesses.cpp:318,469\`) — the witness key name. SV-COMP's witness format has used both spellings at different times; depending on which validator we're targeting, one is required and the other rejected.
2. **Don't repeat witness nodes** — \`grapht::create_initial_edge\` was rewritten so the initial node is shared rather than newly allocated each time, eliminating spurious duplicate edges.
3. **Only generate witness nodes from the file under verification** — skip trace steps from system headers (\`<assert.h>\`, etc.), which currently pollute the output.
4. **Don't print location from outside assignments** — branch comment: \`otherwise we can erroneously insert a return from main\`. Sounds like a real witness-graph correctness issue.

Plus a sibling fix:
- **Incomplete struct/union types** — \`src/clang-c-frontend/clang_c_adjust_expr.cpp\` (41 lines) handles types that aren't fully declared at the point they're referenced.

## Why this matters

Issue [#1520](https://github.com/esbmc/esbmc/issues/1520) (\`Witness not parseable by CPAchecker\`) has been open since 2023. That report is about a *different* parse failure (constant struct/array literal), but witness-key mismatches are a separate plausible cause of validator rejection that hasn't been investigated.

## Action

- [ ] Audit current GraphML output against the SV-COMP common-witness-format spec (https://github.com/sosy-lab/sv-witnesses). Determine whether \`returnFromFunction\` or \`returnFrom\` is the spec-conformant key. (My read: the 2020-onwards format uses \`returnFromFunction\`; older 1.0 used \`returnFrom\`. Check current ruleset.)
- [ ] Run a witness through Ultimate / CPAchecker / wit2test validators with the current key vs the branch's renamed key; see which is accepted.
- [ ] Pick the duplicate-node and out-of-file-trace-step fixes from the branch (these look unambiguously correct).
- [ ] Investigate the \`return-from-main\` insertion bug Mikhail describes in commit \`48a23ca8f5\`.

## References

- Source branch: \`adjust_type\` (will be deleted as part of stale-branch cleanup)
[...]

PR #5667, opened 2026-06-28
[docs] SV-COMP triage Pass 13–14: scanf fixes + witness issue closures

lucasccordeiro:
**Stacked on #5664** (Pass-12 doc chain); base retargets to master as the chain lands.

Recurring SV-COMP triage reconciliation (§17 Pass 13):

- **§16.2 scanf numeric-specifier fix shipped** as **#5666** — `input_overflow_check` no longer flags width-less numeric/char scanf (`%d`, `%ld`, `%f`, `%c`) as a buffer overflow; the `INF` rule is restricted to genuine unbounded string conversions (`%s`, `%[...]`, `%S`/`%ls`).
- **#4611 concrete items verified already-resolved** — audited the current GraphML emitter: it already uses the modern-spec `enterFunction`/`returnFromFunction` keys, emits no duplicate nodes, and includes no system-header trace steps (only user-file `startline`s). #4611 reduces to the external-validator audit (CPAchecker/Ultimate), which isn't reproducible here.
- **In-flight PRs reconciled:** #5660 (#5142 parse layer), #5664 (#1470 witness-assignment tests), #5666 (scanf).

No code change in this PR. The remaining backlog is research-grade (value-set, reachability, va_list, k-induction, atomics, device-API) or validation-only (#1471/#1492/#4611/#4427) — none isolation-verifiable in this environment, so none patched with a heuristic.

### DIFF
--- a/docs/svcomp-issue-triage.md
+++ b/docs/svcomp-issue-triage.md
@@ -1453,3 +1453,86 @@ not re-touched — in flight as #5660.
 8. **#5565 residual** — `strcmp` argv-string pointer-validity only (the `fclose` half is discharged,
    §15.2); `ESBMC_SVCOMP`/`--sv-comp`-gated.
 9. **Close-outs pending CI:** #1470, #4427; witness end-to-end validation for #1471/#1492/#4611.
+
+---
+
+## 16. Pass 12 — #1470 overflow-witness assignment verified emitted + pinned (2026-06-28)
+
+This pass picks the most concretely-verifiable close-out from §15's item 9 — **#1470** (*overflow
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
Extended with **§18 Pass 14**: records the three witness issues closed after the validity re-audit (#1470, #4611, #1471 — open SV-COMP count 20→17), and that #5666 now bundles both scanf overflow-check parser fixes (width-less-numeric false positive + `%%`/`%*` false negative). Kept on this branch rather than opening a fourth stacked doc PR.

PR #5705, opened 2026-06-30
[goto-symex] Hide internal phi-merge steps from --no-slice trace

lucasccordeiro:
Under bare `--no-slice`, the human-readable counterexample surfaced hidden SSA phi-merge assignments at real source lines with a synthesised merged value, producing the contradictory back-to-back states reported in discussion #5701.

Following @XLiZHI's review, this filters hidden steps unconditionally in `build_goto_trace` (`if (SSA_step.hidden) continue;`) rather than only under compact tracing, instead of propagating a new `hidden` flag onto `goto_trace_stept`. Because the filtering now happens at trace-build time, every trace consumer — plain text, XML, JSON, HTML, SARIF and witnesses — stays consistent, and the now-redundant `is_compact_trace` plumbing is removed.

Adds two regression tests under `regression/esbmc/github_5701*` (a deterministic guard plus the faithful reproducer). Verdicts are unchanged.

### DIFF
--- a/regression/CMakeLists.txt
+++ b/regression/CMakeLists.txt
@@ -502,6 +502,15 @@ set(_slow_regression_tests
     # does not touch its code path). Give it the 2x budget so transient
     # runner-speed differences don't flap it.
     "regression/python/str_rsplit"
+    # set_variadic_methods verifies SUCCESSFUL but is heavy: under
+    # --incremental-bmc it climbs k while re-running symex over the set/list
+    # operational models (variadic union/intersection/difference fold to
+    # __ESBMC_list_extend + __memcpy_impl loops), re-encoding ~1400 VCCs and a
+    # fresh Bitwuzla query each step. It lands ~137s locally and timed out at
+    # 120.71s on the DebugOpt Linux runner in #5705 CI -- a run that does not
+    # touch its code path. Give it the 2x budget so transient CI slowness does
+    # not flap it.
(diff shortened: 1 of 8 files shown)


XLiZHI:
https://github.com/esbmc/esbmc/blob/51604ba2701fdd50ea74f33b6060f480d487f205/src/goto-symex/build_goto_trace.cpp#L69-L70
`--no-slice` might be a raw SSA dump mode for developers, so it was not filtered out here.

I reckon this is the main cause of this issue. The least modification would be to remove the "is_compact_trace" flag.

This way, there is no need to introduce a new member variable named "hidden".

lucasccordeiro:
Thanks @XLiZHI — adopted your suggestion. `build_goto_trace` now filters hidden SSA steps unconditionally (`if (SSA_step.hidden) continue;`), so the new `goto_trace_stept.hidden` member, its propagation, and the `show_goto_trace` change are all gone, and the redundant `is_compact_trace` flag (plus its two computation blocks in `bmc.cpp`) is removed too.

Filtering at trace-build time has the extra benefit of keeping every trace consumer consistent (text, XML, JSON, HTML, SARIF, witnesses), not just the human-readable printer. Both `github_5701*` regression tests and the existing `--no-slice` / `--compact-trace` tests still pass.

XLiZHI (approved):
LGTM

PR #5729, opened 2026-07-01
[python] Wire in and generalize the CPython differential model tests

pmatos:
## What

Wires the dormant pytest/hypothesis suite under `unit/python-frontend/` into
ctest so it runs in CI, and generalises the "load the operational model and
compare it against CPython" differential pattern (previously only
`test_decimal_model.py`) to the pure-Python models.

## Why

`unit/python-frontend/*.py` had no `add_test` and no pytest invocation in any
workflow, so the one real differential-vs-CPython asset provided zero
regression protection. Because they never ran, several had silently bit-rotted:
the parser/preprocessor packages were refactored (`preprocessor.py` ->
`preprocessor/`, `parser.py` -> `parser/parser.py`) and the tests' file-based
module loading pointed at paths that no longer exist.

## Changes

- **Wire into ctest** (`unit/python-frontend/CMakeLists.txt`): register a
  `python_model_diff` test that runs `pytest` over the directory, gated on
  pytest+hypothesis being importable so a minimal build still configures and
  skips cleanly. The suite is pure Python and needs no ESBMC binary, so it runs
  inside the existing ctest step. Labelled `python`.
- **CI deps** (`scripts/build.sh`): install `pytest hypothesis` on Linux and
  macOS.
- **Repair bit-rot**: the 6 preprocessor/complex tests now import the
  refactored packages directly (`import preprocessor`, `from parser import
  parser`); `test_dataclasses_model_api.py` is marked `xfail(strict=False)`
  because `models/dataclasses.py` is an intentional minimal stub that does not
  implement the full dataclasses semantics those tests assert.
- **New differential tests**: `test_{range,float,string,int,math}_model.py`
  (+42 tests) compare each model against the CPython builtin/stdlib with
  hypothesis. `int.py` needs the ESBMC-internal `IntWide` annotation bound
  before exec; the math test covers only the pure-Python subset (binding
  `isinf`/`isnan` to CPython so `floor`/`ceil` are runnable) and skips the
  intrinsic-backed transcendentals.

## Testing

- Full suite on the CI import surface (no `ast2json` installed):
  **204 passed, 5 xfailed, 0 failures**.
- ctest registration validated end-to-end: registers and runs green when
  pytest+hypothesis are present; skips with a STATUS message otherwise.
- Negative check: perturbing a model value makes the corresponding differential
  test fail; reverting restores green.

## Follow-up

The high-traffic **C** operational models (`list.c`, `string.c`, `slice.c`)
cannot be imported into CPython and need a different mechanism (ESBMC-driven
concrete/nondet harnesses). That work is tracked separately in
pmatos/esbmc#7.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -414,7 +414,7 @@ install_gmp_linux() {
 
 install_python_deps_linux() {
   log "Installing Python dependencies"
-  python3 -m pip install --user meson mypy pyparsing toml tomli
+  python3 -m pip install --user meson mypy pyparsing toml tomli pytest hypothesis
   meson --version
 }
 
@@ -436,9 +436,9 @@ install_python_deps_macos() {
   export PATH="$py312_bin:$HOME/Library/Python/3.12/bin:$PATH"
 
   if [[ -n "${VIRTUAL_ENV:-}" ]]; then
-    pip install meson mypy pyparsing toml tomli
+    pip install meson mypy pyparsing toml tomli pytest hypothesis
   else
-    "$py312" -m pip install --user --break-system-packages meson mypy pyparsing toml tomli
+    "$py312" -m pip install --user --break-system-packages meson mypy pyparsing toml tomli pytest hypothesis
   fi
 
   meson --version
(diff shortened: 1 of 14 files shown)


pmatos:
It's great to the see the amount of python dev going into esbmc but I was worried that we didn't have enough testing to ensure things remain sane. Hopefully this PR is a step in the right direction.

lucasccordeiro (approved):
Nice!

PR #5784, opened 2026-07-03
[esbmc] distinguish heap OOB accesses and map to CWE-122

### LINKED ISSUE #4492: [esbmc] distinguish heap OOB writes and map to CWE-122
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). The current `array bounds violated` and `Access to object out of bounds` comments map to a generic set of bounds CWEs ({121, 125, 129, 131, 193, 787}); CWE-122 *Heap-based Buffer Overflow* is **not** emitted, because ESBMC does not currently distinguish heap-allocated from stack-allocated objects in its OOB diagnostic.

### CWE-122 — Heap-based Buffer Overflow

[CWE-122](https://cwe.mitre.org/data/definitions/122.html) (`ALLOWED` for vulnerability mapping in CWE 4.20) is the heap-specific variant of CWE-787 *Out-of-bounds Write*. Distinguishing it is valuable for downstream consumers (e.g. SARIF triage, CVE-style classification, security dashboards) because heap-based overflows have different exploitation profiles than stack-based ones (CWE-121, which we already emit alongside the generic set).

### Proposal

ESBMC's pointer model already knows whether a referenced object came from `malloc`/`calloc`/`realloc` (`dynamic_object2t`) versus the stack. The fix is to consult that information at the point of emitting the `array bounds violated` / `Access to object out of bounds` assertion comment in `src/pointer-analysis/dereference.cpp` and `src/goto-programs/goto_check.cpp`, and to tag the comment when the underlying object is dynamic:

```
array bounds violated: heap object
```

vs the existing

```
array bounds violated
```

Then:

1. Add two new rules in `src/util/cwe_mapping.cpp`:
   - `"array bounds violated: heap object"` → `{ "heap-array-bounds-violated", "Array bounds violated on heap object", {122, 125, 129, 131, 193, 787} }`.
   - `"Access to object out of bounds: heap object"` → `{ "heap-object-out-of-bounds", "Access to heap object out of bounds", {122, 125, 787, 823} }`.
2. Update `cwe_name` with "Heap-based Buffer Overflow" for CWE-122.
3. Add a unit test for the substring → 122 path.
4. Add a regression test `regression/esbmc/cwe_heap_overflow/` allocating with `malloc(N)` and writing at index `N+k`, asserting `CWE-122` in the trace and SARIF.
5. Update `docs/cwe-mapping.md` (two new table rows + count).

Care must be taken: when the OOB write hits a heap object, both the heap-specific rule (122) and the existing generic rule (121 etc.) reasonably apply. Suggested behaviour is to emit **only** the heap-specific rule when the underlying object is heap-allocated; both CWE-122 (variant) and CWE-787 (the immediate parent) are still emitted via the new rule's `cwes` list.

### Acceptance criteria

- For a `malloc`-allocated buffer with an OOB write, `esbmc heap_oob.c` prints `CWE: CWE-122, CWE-125, CWE-787, …`.
- For a stack array with an OOB write, the output is unchanged from today.
- SARIF reports CWE-122 in `taxa` for the heap case only.
- `docs/cwe-mapping.md` updated.

pmatos:
## Summary

- Out-of-bounds accesses against a `malloc`/`calloc`/`realloc` allocation are now tagged `: heap object` in the violation comment, so they map to **CWE-122 (Heap-based Buffer Overflow)** instead of the stack-oriented CWE-121. This closes the gap noted in #4488: ESBMC previously mapped every bounds violation to the generic `{121, 125, 129, 131, 193, 787}` set and never emitted CWE-122.
- The symex dereference code already knows an object's origin — heap allocations carry the `symex_dynamic::` symbol prefix, while `alloca` (stack) carries `symex_dynamic::alloca::`. A small `is_heap_object()` helper in `src/pointer-analysis/dereference.cpp` consults this at the two OOB assertion sites (`bounds_check` for arrays, `check_data_obj_access` for size-1 objects).
- Stack/global arrays and the compile-time `goto_check.cpp` bounds path are **unchanged** — that path early-returns for pointer types and dereferences and runs before symex, so it never sees dynamic objects (heap OOB always reaches symex as a pointer dereference).
- The two heap rules are strict superstrings of the generic comments, so the mapping table's existing longest-substring-first sort makes them win the match for heap accesses; the generic (stack) entries are untouched.

### Acceptance criteria (from #4492)

- ✅ `malloc`-allocated buffer OOB write prints `CWE: CWE-122, CWE-125, CWE-129, CWE-131, CWE-193, CWE-787`.
- ✅ Stack array OOB is unchanged (`CWE-121, …`, no CWE-122).
- ✅ SARIF `taxa` carries CWE-122 (rule `heap-array-bounds-violated`) for the heap case only.
- ✅ `cwe-mapping.md` updated (two new rows, CWE-122 in the id list, counts 31→32 ids / 26→28 kinds).

Note: as with the pre-existing bounds entries, the heap variants do not split reads from writes — the CWE list keeps both CWE-125 (read) and CWE-787 (write), so a heap OOB read is also annotated CWE-122. This matches ESBMC's existing coarse mapping and is documented.

Fixes #4492

## Test plan

- [x] `ctest -R cwe_mapping` — unit suite green (673 assertions, 18 cases)
- [x] `regression/esbmc/cwe_heap_overflow` — heap array OOB → `array bounds violated: heap object` + CWE-122 (FAILED)
- [x] `regression/esbmc/cwe_heap_scalar_oob` — size-1 heap object OOB → `Access to object out of bounds: heap object` + CWE-122 (FAILED)
- [x] `regression/esbmc/cwe_heap_overflow_sarif` — SARIF taxa id 122 via rule `heap-array-bounds-violated`
- [x] `regression/esbmc/cwe_stack_overflow` — stack array OOB stays on CWE-121, no CWE-122
- [x] Existing `cwe_*` (23) and bounds/malloc/dynamic/deref (43) regression tests still pass
- [x] clang-format clean on changed C++


src/pointer-analysis/dereference.cpp
```diff
@@ -2366,6 +2366,24 @@ void dereferencet::valid_check(
...
+  if (!is_symbol2t(base))
+    return false;
+  const symbolt *sym = ns.lookup(to_symbol2t(base).thename);
+  if (!sym)
+    return false;
+  const std::string id = sym->id.as_string();
+  return has_prefix(id, "symex_dynamic::") &&
+         !has_prefix(id, "symex_dynamic::alloca::");
+}
+
```
lucasccordeiro, line 2399: The `has_prefix(id, "symex_dynamic::")` definition of "dynamic" is now duplicated with `valid_check` (line ~2316). Consider a shared helper so the notion of what counts as a dynamic object lives in one place.

src/pointer-analysis/dereference.cpp
```diff
@@ -2582,7 +2604,10 @@ void dereferencet::check_data_obj_access(
     guard2tc tmp_guard = guard;
     tmp_guard.add(gt);
     dereference_failure(
-      "pointer dereference", "Access to object out of bounds", tmp_guard);
+      "pointer dereference",
+      is_heap_object(value, ns) ? "Access to object out of bounds: heap object"
+                                : "Access to object out of bounds",
```
lucasccordeiro, line 2627: This tags a size-1 heap object (e.g. `malloc(sizeof(int))`) OOB as CWE-122, whose canonical definition implies a buffer/array. Defensible, but worth a one-line note in the PR that CWE-122 on scalar heap objects is intentional, alongside the existing read/write caveat.

lucasccordeiro (commented):
Overall sound and clean. I verified the key claims: the `symex_dynamic::`/`alloca::` naming (`memory_alloc.cpp:565`), the longest-substring-first sort so the `: heap object` superstrings win, and that the compile-time `goto_check.cpp:1019` path keeps stack arrays on CWE-121. Doc counters add up. A few minor, non-blocking points below.

3. **Non-symbol base falls back to stack.** When the overflowed base isn't a `symbol2t` (unknown/nondet object), `is_heap_object` returns false and the access is labelled CWE-121. In practice symex dereferences per-object so heap objects resolve to a symbol, but a one-line note that the fallback direction is intentional would make the intent explicit.

4. **Duplicate fixture.** `cwe_heap_overflow/main.c` and `cwe_heap_overflow_sarif/main.c` are byte-identical — fine given each test dir is self-contained, just flagging so it doesn't read as accidental.

lucasccordeiro:
Pushed a follow-up addressing the review points:
- Extracted a shared `is_symex_dynamic_object()` so the `symex_dynamic::` definition of "dynamic object" lives in one place (used by both `valid_check` and `is_heap_object`).
- Added the intent notes requested: non-symbol/nondet base intentionally falls back to the stack classification (CWE-121), and tagging a size-one heap object CWE-122 is deliberate (mirrors the existing read/write coarsening).

The byte-identical fixtures were left as-is — each regression dir is self-contained. Heap/stack/bounds regressions (135) green.

pmatos:
@lucasccordeiro thanks for the review and addressing them in one go. Feel free to merge then if you're happy with it.

PR #5785, opened 2026-07-03
[symex] map no-base-case recursion to CWE-674

### LINKED ISSUE #4493: [esbmc] map k-induction-refuted recursion to CWE-674
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). ESBMC currently emits a `recursion unwinding assertion` / `unwinding assertion loop` comment when the k-bound is exhausted, but this is intentionally **unmapped** in the CWE table because k-bound exhaustion signals insufficient unwinding, not a weakness.

### CWE-674 — Uncontrolled Recursion

[CWE-674](https://cwe.mitre.org/data/definitions/674.html) (`ALLOWED-WITH-REVIEW` for vulnerability mapping in CWE 4.20) is the genuine *weakness* case: the program contains a recursive call whose termination is not bounded by any input or loop invariant. A k-induction *failure* (not just an unwinding-bound failure) on a recursive function is exactly this property.

This is distinct from CWE-835 (loop infinite-loop, separately tracked) — CWE-674 covers recursive calls.

### Proposal

1. Distinguish two different failure modes in `src/goto-symex/symex_function.cpp`:
   - **k-bound exhausted** (current behaviour, `recursion unwinding assertion`) — remains unmapped.
   - **k-induction refutes termination** of the recursive call — emit a new comment, e.g. `uncontrolled recursion in <function>`.
2. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping `"uncontrolled recursion"` to `{ "uncontrolled-recursion", "Uncontrolled recursion", {674} }`.
3. Add the CWE-674 short name to `cwe_name` ("Uncontrolled Recursion").
4. Add a unit test in `unit/util/cwe_mapping.test.cpp`.
5. Add a regression test `regression/esbmc/cwe_uncontrolled_recursion/` with an Ackermann-style or unbounded-recursion function and `--k-induction --termination`, expecting `CWE-674` in the trace.
6. Update `docs/cwe-mapping.md` (table + count + note that this is separate from the unwinding-bound case).

The discriminator is essentially: does k-induction's inductive step refute termination, or did we merely fail to reach a fixpoint within `--max-inductive-step`? The first is CWE-674; the second is a verifier-bound issue and stays unmapped.

### Acceptance criteria

- For a benchmark with a recursive call that genuinely never terminates, ESBMC with `--k-induction` prints `CWE: CWE-674`.
- For a benchmark with a recursive call that terminates but exhausts the default k-bound, the verdict and comment are unchanged from today (no CWE).
- SARIF / JSON / GraphML / textual outputs all reflect the new mapping.
- Unit and regression tests pass.

pmatos:
## Summary

Distinguishes **genuinely uncontrolled recursion** (a recursive function with no reachable base case) from mere unwinding-bound exhaustion, and maps the former to **CWE-674 Uncontrolled Recursion**. The pre-existing `recursion unwinding assertion` (k-bound exceeded) stays intentionally unmapped.

- `src/goto-symex/symex_function.cpp`: new `recursion_has_no_base_case()` — a structural CFG walk from the function entry that treats a *direct recursive self-call* as an impassable sink. If `END_FUNCTION` is unreachable that way, every return path passes through a self-call, so the recursion can never bottom out. The recursion-unwinding claim is then relabeled to `uncontrolled recursion in <fn>`; otherwise it keeps `recursion unwinding assertion`.
- `src/util/cwe_mapping.cpp`: `"uncontrolled recursion"` → `{uncontrolled-recursion, CWE-674}`, plus the `674 → "Uncontrolled Recursion"` short name.
- Docs, unit tests, and five regression tests.

The CWE tag flows through the existing PR #4488 plumbing (textual counterexample, JSON, GraphML, SARIF) with no changes to those backends.

### Soundness

The detector is **sound** — it never relabels a recursion that has *any* non-recursive return path, so terminating recursion is never mislabelled as a weakness. It is intentionally **incomplete**: only direct self-recursion is analysed (mutual and input-dependent non-termination are not flagged), and it only ever relabels a recursion-unwinding claim that already fires, so the combination is a sound non-termination witness. Verdicts and SV-COMP/Test-COMP classification are unchanged (the wrappers key only on `unwinding assertion loop`).

### Note on the discriminator vs. the issue's proposal

The issue proposed keying off *k-induction's inductive step refuting recursion termination*. That path does not exist in the current engine: symex **disables the inductive step on any recursion** (`symex_function.cpp`, "k-induction does not support recursion yet"), `--k-induction`+`--termination` are mutually exclusive, and the termination strategy treats the IS as inconclusive whenever `disable-inductive-step` is set. A faithful implementation of that mechanism would be research-scale and touch core soundness. This PR instead uses a sound structural discriminator that meets the acceptance criteria without relying on the disabled machinery.

Consequently the CWE-674 output appears where a recursion-unwinding **counterexample** is produced — i.e. plain `--unwind` BMC. Under `--k-induction`/`--incremental-bmc`, non-terminating recursion yields `VERIFICATION UNKNOWN` (the forward condition can't prove termination and the inductive step is disabled), so no counterexample/CWE line is emitted — this is inherent, not a regression.

Fixes #4493

## Test plan

- [x] `unit/util/cwe_mapping_test` — 607 assertions / 18 cases, incl. `uncontrolled recursion → CWE-674`, `cwe_name(674)`, and `recursion unwinding assertion` staying unmapped
- [x] `regression/esbmc/cwe_uncontrolled_recursion` — no base case → `uncontrolled recursion in f` + `CWE: CWE-674`, `VERIFICATION FAILED`
- [x] `regression/esbmc/cwe_recursion_unwinding_unmapped` — terminating-but-deep → `recursion unwinding assertion`, no CWE
- [x] `regression/esbmc/cwe_recursion_bounded_ok` — terminates within bound → `VERIFICATION SUCCESSFUL`, no CWE (no false positive)
- [x] `regression/esbmc/cwe_recursion_fibonacci_unmapped` — terminating doubly-recursive → not mislabelled
- [x] `regression/esbmc/cwe_recursion_mutual_incomplete` — mutual recursion → not flagged (documents intentional incompleteness)
- [x] 611-test regression smoke sweep: no new failures (only pre-existing `boolector not built` env cases)
- [x] Existing `cwe_null_deref` / `cwe_overflow` / `cwe_sarif` / unchecked-return / uninit tests still pass

### DIFF
--- a/regression/esbmc/cwe_recursion_mutual_incomplete/test.desc
+++ b/regression/esbmc/cwe_recursion_mutual_incomplete/test.desc
@@ -1,0 +1,6 @@
+CORE
+main.c
+--unwind 8
+^VERIFICATION FAILED$
+^  recursion unwinding assertion$
+\A(?![\s\S]*CWE)[\s\S]*\Z
--- a/src/goto-symex/symex_function.cpp
+++ b/src/goto-symex/symex_function.cpp
@@ -1,5 +1,7 @@
 #include <algorithm>
 #include <cassert>
+#include <map>
+#include <unordered_set>
 #include <goto-symex/execution_state.h>
 #include <goto-symex/goto_symex.h>
 #include <langapi/language_util.h>
@@ -28,6 +30,79 @@ bool goto_symex_utils::is_alloca_return_value_name(const std::string &name)
       return false;
   return true;
 }
+
+namespace
+{
+// A recursive function has a *reachable base case* iff some path from its
+// entry reaches END_FUNCTION without going through a direct recursive
+// self-call. If no such path exists, the function can never return without
+// first recursing again — the recursion is genuinely unbounded (CWE-674
+// "Uncontrolled Recursion") rather than merely deep.
+//
+// The analysis is deliberately structural: it walks the CFG treating a
+// direct self-call as an impassable edge and otherwise following
+// get_successors(). It is sound but incomplete — it never reports a
+// recursion that has *any* non-recursive return path, so terminating
+// recursion that only exhausts the unwinding bound keeps the existing
+// "recursion unwinding assertion" comment. Callers consult this solely to
+// relabel a recursion-unwinding claim that has already fired (the recursion
+// really did reach the bound on a feasible path), so the combination is a
+// sound witness of non-termination.
+bool recursion_has_no_base_case(
+  const goto_functiont &goto_function,
+  const irep_idt &identifier)
+{
+  // Keyed on the mangled identifier alone: sound because symex is
+  // single-threaded and there is exactly one body per identifier per process
+  // (k-induction phases reuse the same function_map; --k-induction-parallel
+  // forks). The structural base-case property is invariant to the unwind
+  // bound and phase, so a cached result never goes stale.
+  static std::map<irep_idt, bool> cache;
+  auto cached = cache.find(identifier);
+  if (cached != cache.end())
+    return cached->second;
+
+  const goto_programt &body = goto_function.body;
+  bool no_base_case = false;
+  if (!body.instructions.empty())
+  {
+    std::unordered_set<const goto_programt::instructiont *> visited;
+    std::vector<goto_programt::const_targett> work{body.instructions.begin()};
+    bool reached_end = false;
+    while (!work.empty())
+    {
+      goto_programt::const_targett cur = work.back();
+      work.pop_back();
+      if (!visited.insert(&*cur).second)
+        continue;
+      if (cur->is_end_function())
+      {
+        reached_end = true;
+        break;
+      }
+      // A direct recursive self-call is an impassable sink: reaching
+      // END_FUNCTION through it requires the recursion to bottom out first.
+      if (cur->is_function_call())
+      {
+        const code_function_call2t &c = to_code_function_call2t(cur->code);
+        if (
+          is_symbol2t(c.function) &&
+          to_symbol2t(c.function).thename == identifier)
+          continue;
+      }
+      goto_programt::const_targetst succs;
+      body.get_successors(cur, succs);
+      for (const auto &s : succs)
+        if (s != body.instructions.end())
+          work.push_back(s);
+    }
+    no_base_case = !reached_end;
+  }
+
+  cache.emplace(identifier, no_base_case);
+  return no_base_case;
+}
+} // namespace
 
 bool goto_symext::get_unwind_recursion(
   const irep_idt &identifier,
@@ -407,7 +482,17 @@ void goto_symext::symex_function_call_code(const expr2tc &expr)
   {
     if (!no_unwinding_assertions)
     {
-      claim(gen_false_expr(), "recursion unwinding assertion");
+      // Distinguish a genuine weakness (a recursive function with no
+      // reachable base case — CWE-674) from a merely-exhausted k-bound. Only
+      // the former gets the "uncontrolled recursion" comment that the CWE
+      // mapping recognises; the latter stays intentionally unmapped.
+      if (recursion_has_no_base_case(goto_function, identifier))
+        claim(
+          gen_false_expr(),
+          "uncontrolled recursion in " +
+            get_pretty_name(identifier.as_string()));
+      else
+        claim(gen_false_expr(), "recursion unwinding assertion");
     }
     else
     {
--- a/src/goto-symex/symex_function.cpp
+++ b/src/goto-symex/symex_function.cpp
@@ -1,6 +1,5 @@
 #include <algorithm>
 #include <cassert>
-#include <map>
 #include <unordered_set>
 #include <goto-symex/execution_state.h>
 #include <goto-symex/goto_symex.h>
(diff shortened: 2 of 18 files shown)


src/goto-symex/symex_function.cpp
```diff
@@ -29,6 +31,79 @@ bool goto_symex_utils::is_alloca_return_value_name(const std::string &name)
...
+bool recursion_has_no_base_case(
+  const goto_functiont &goto_function,
+  const irep_idt &identifier)
+{
+  // Keyed on the mangled identifier alone: sound because symex is
+  // single-threaded and there is exactly one body per identifier per process
+  // (k-induction phases reuse the same function_map; --k-induction-parallel
+  // forks). The structural base-case property is invariant to the unwind
+  // bound and phase, so a cached result never goes stale.
+  static std::map<irep_idt, bool> cache;
```
lucasccordeiro: This file-static cache is process-lifetime global state keyed on the mangled identifier *alone*, not on the body. The "one body per identifier per process" invariant is asserted in the comment but not enforced: any in-process driver that verifies more than one goto model (unit-test harnesses, future multi-file/library use) gets a stale result for a same-named function with a different body. The walk is O(V+E) and runs at most once per recursive function per bound-exhaustion — either drop the cache, or make it a member on `goto_symext` so its lifetime is tied to the verification run. Recommend removing it.

src/goto-symex/symex_function.cpp
```diff
@@ -407,7 +482,17 @@ void goto_symext::symex_function_call_code(const expr2tc &expr)
   {
     if (!no_unwinding_assertions)
     {
-      claim(gen_false_expr(), "recursion unwinding assertion");
+      // Distinguish a genuine weakness (a recursive function with no
+      // reachable base case — CWE-674) from a merely-exhausted k-bound. Only
+      // the former gets the "uncontrolled recursion" comment that the CWE
+      // mapping recognises; the latter stays intentionally unmapped.
+      if (recursion_has_no_base_case(goto_function, identifier))
```
lucasccordeiro, line 512: The walk treats any structural path to `END_FUNCTION` as a base case, ignoring feasibility, so `int f(int n){ if (0) return 0; return f(n+1); }` stays `recursion unwinding assertion` (unmapped) though it never terminates. Within the documented sound-but-incomplete contract, but it depends on goto constant-folding *not* pruning the dead branch — otherwise the same source flips label with optimization level. Add one line to the doc/comment stating feasibility is intentionally not considered, so the label is stable by design rather than by accident.

regression/esbmc/cwe_recursion_mutual_incomplete/test.desc
```diff
@@ -0,0 +1,6 @@
+CORE
+main.c
+--unwind 8
+^VERIFICATION FAILED$
+^  recursion unwinding assertion$
```
lucasccordeiro, line 5: `^  recursion unwinding assertion$` assumes a deterministic firing site. With mutual `f<->g` and `--unwind 8`, please confirm which function trips the bound first is stable across solvers/platforms; if it can vary, this exact anchor is fragile. (The four single-function `--unwind` tests are unaffected.)

lucasccordeiro (commented):
A few actionable points. Non-issues I checked and cleared: the verdict is unchanged (both paths `claim(gen_false_expr(), ...)`); SV-COMP/Test-COMP wrappers key only on `"unwinding assertion loop"` (`esbmc-wrapper.py:125`), not the recursion comment; `witness.cpp:199` filters `"unwinding assertion loop"` only, so witness behavior is identical for the old and new comments; doc counts (26→27 kinds, 31→32 ids) are consistent.

**Missing test for the documented `--k-induction`/`--incremental-bmc` behavior.** The PR states non-terminating recursion yields `VERIFICATION UNKNOWN` / no CWE-674 under those modes and calls it "inherent, not a regression", but nothing pins it — all five regression tests use plain `--unwind`. Add one test (e.g. `cwe_uncontrolled_recursion` under `--incremental-bmc`) asserting no `CWE-674` line, so a future change to the inductive-step handling can't silently alter this contract.

lucasccordeiro:
Follow-up pushed. The main one is a real soundness fix flagged by Codex: a recursion whose base case terminates via `exit()`/`abort()` (`if(n<=0) exit(0); return f(n-1);`) was mislabelled CWE-674. Those paths bottom out through `__ESBMC_assume(0)` in the callee body, invisible to the caller-local walk, so `recursion_has_no_base_case` now treats a call to a program-terminating function as a reachable base case. Verified: that shape is back to `recursion unwinding assertion`, while genuine `f(){return f();}` still gets CWE-674.

Also: dropped the process-lifetime static cache (stale for in-process multi-model drivers; the O(V+E) walk is cheap), documented that feasibility/constant-folding is intentionally not considered, and added regression tests for the noreturn base case and the `--incremental-bmc` (UNKNOWN, no CWE-674) contract. The mutual-recursion anchor is stable — the message carries no function name.

lucasccordeiro (approved):

pmatos:
@lucasccordeiro Thanks for getting this to completion.

PR #5786, opened 2026-07-03
[esbmc] Detect excessive allocation sizes (CWE-789)

### LINKED ISSUE #4497: [esbmc] detect excessive allocation sizes and map to CWE-789
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). ESBMC's existing dynamic-allocation modelling caps `malloc`'s symbolic size only loosely (see `src/goto-symex/builtin_functions/memory_alloc.cpp`); the *root-cause* CWE for "attacker-controlled size argument can be arbitrarily large" — CWE-789 — is not directly detected.

### CWE-789 — Memory Allocation with Excessive Size Value

[CWE-789](https://cwe.mitre.org/data/definitions/789.html) (`ALLOWED` for vulnerability mapping in CWE 4.20) covers `malloc(n)` / `calloc(m, n)` / `new[n]` where `n` is unbounded from the program's perspective (typically attacker-controlled via I/O). The classic failure mode is a DoS / availability vulnerability.

This is naturally expressible in BMC: insert an assertion `__ESBMC_assert(n <= K, "excessive allocation size: <fn>")` immediately before every allocation, where `K` is a configurable bound (`--max-alloc-size`). If symex finds a model with `n > K`, that is a CWE-789 witness.

### Proposal

1. Add a new flag `--excessive-alloc-check[=K]` in `src/esbmc/options.cpp` (default K = 1 MiB, configurable). Wires through to a new instrumentation pass in `src/goto-programs/` that injects an assertion before each `malloc`/`calloc`/`realloc`/operator `new[]`.
2. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping `"excessive allocation size"` to `{ "excessive-allocation", "Memory allocation with excessive size", {789} }`.
3. Add the CWE-789 short name to `cwe_name` ("Memory Allocation with Excessive Size Value").
4. Unit test in `unit/util/cwe_mapping.test.cpp`.
5. Regression test `regression/esbmc/cwe_excessive_alloc/`: a program that reads `n` from `nondet_uint()` and calls `malloc(n)` without a bound check (FAIL), plus a paired `_pass/` with a `n <= 1024` guard.
6. Update `docs/cwe-mapping.md` (table + count).

### Care points

- The bound `K` is a policy choice, not a soundness property. Document this explicitly: ESBMC proves "no path can reach `malloc(n)` with `n > K`", *not* "no path can exhaust memory" (the latter is undecidable in general).
- Should interact cleanly with `--force-malloc-success`: when forcing success, an excessive-size allocation should still raise CWE-789.
- May overlap with future CWE-770 *Allocation of Resources Without Limits or Throttling*; CWE-770 is the class entry — keep CWE-789 (the variant) as the primary mapping.

### Acceptance criteria

- `esbmc unbounded_malloc.c --excessive-alloc-check` emits `CWE: CWE-789` when an unbounded `malloc` size is reachable.
- The paired `_pass/` test passes silently.
- SARIF / JSON / GraphML / textual outputs all reflect the new mapping.
- Existing regressions and SV-COMP wrapper output are unaffected when the flag is off.

pmatos:
## Summary

- Adds an opt-in `--excessive-alloc-check[=K]` goto preprocessing pass that inserts `ASSERT(size <= K)` before every `malloc`/`calloc`/`realloc`/`operator new[]` (default `K` = 1 MiB), mapping violations to **CWE-789** (Memory Allocation with Excessive Size Value).
- The pass computes the byte request per allocation kind (`malloc`/`realloc` sizes are already bytes; `operator new[]` scales the element count by `sizeof(element)`) and threads branch guards so the frontend's `realloc` lowering `p==0 ? malloc(n) : realloc(p,n)` reports the allocator actually reached. `calloc` is covered transitively via its operational model.
- The bound `K` is a **policy** choice, not a soundness property, and the assertion precedes the allocation, so an excessive size is still reported under `--force-malloc-success`.
- CWE-789 flows through all existing annotation paths: textual (`CWE: CWE-789`), SARIF (`ruleId: excessive-allocation` + taxon 789), JSON (`assertion.cwe: [789]`), and GraphML (`<data key="cwe">CWE-789</data>`).
- Mirrors the existing `goto_check_unchecked_return` (CWE-252) pass. Off by default, so existing regressions and SV-COMP wrapper output are unaffected.

Implements the extension proposed in the issue (tracks #4488 / #1854).

Fixes #4497

## Test plan

- [x] `ctest` unit suite — `util/cwe_mapping` (18 cases, 609 assertions) passing, incl. new CWE-789 case + `cwe_name(789)`
- [x] New regression tests passing: `cwe_excessive_alloc` (malloc, FAIL), `cwe_excessive_alloc_pass` (guarded, SUCCESSFUL), `cwe_excessive_alloc_realloc`, `cwe_excessive_alloc_sarif`, `cwe_excessive_alloc_new` (C++ `new[]`)
- [x] 77 allocation/memory/CWE regression tests + 23 existing `cwe_*` tests pass — no regressions
- [x] clang-format 11 clean on all changed C++ files
- [x] Verified `--excessive-alloc-check=K` bound is exact; invalid `K<=0` rejected; SARIF/JSON/GraphML/textual all reflect CWE-789

### DIFF
--- a/src/goto-programs/goto_check_excessive_alloc.cpp
+++ b/src/goto-programs/goto_check_excessive_alloc.cpp
@@ -47,7 +47,7 @@ expr2tc alloc_byte_size(const sideeffect2t &se, const namespacet &ns)
   if (size->type != size_type2())
     size = typecast2tc(size_type2(), size);
 
-  // A char/unknown element type is already a byte count: no scaling needed.
+  // An unknown (nil/empty) element type is already a byte count: no scaling.
   if (is_nil_type(se.alloctype) || is_empty_type(se.alloctype))
     return size;
 
@@ -57,35 +57,14 @@ expr2tc alloc_byte_size(const sideeffect2t &se, const namespacet &ns)
     // uses to model the allocation. A count near 2^64 can wrap the product to
     // a small value and slip under K — a false negative, never a false
     // positive; it mirrors the model rather than contradicting it.
-    BigInt elem_bytes = type_byte_size(se.alloctype, &ns);
-    if (elem_bytes == 1)
-      return size; // sizeof(char)-equivalent: avoid a redundant `* 1`
-    return mul2tc(
-      size_type2(), size, constant_int2tc(size_type2(), elem_bytes));
-  }
-  catch (const array_type2t::dyn_sized_array_excp &)
-  {
-    // A dynamically-sized (VLA) element type, e.g. malloc(sizeof(int[n])):
-    // the frontend records size == 1 with a VLA alloctype, so the byte size
-    // is symbolic but still computable. type_byte_size_expr handles dyn-sized
-    // arrays (unlike the BigInt type_byte_size above, which throws), so scale
-    // by the symbolic element size rather than skipping — otherwise a large
-    // runtime n slips under the bound unchecked.
-    try
-    {
-      expr2tc elem_bytes = type_byte_size_expr(se.alloctype, &ns);
-      if (elem_bytes->type != size_type2())
-        elem_bytes = typecast2tc(size_type2(), elem_bytes);
-      return mul2tc(size_type2(), size, elem_bytes);
-    }
-    catch (const array_type2t::array_size_excp &)
-    {
-      return expr2tc(); // nested element size still unknown — skip
-    }
+    // type_byte_size_expr folds a fixed element type to a constant and returns
+    // a symbolic size for a VLA element (e.g. malloc(sizeof(int[n]))), so this
+    // one path covers both.
+    return mul2tc(size_type2(), size, type_byte_size_expr(se.alloctype, &ns));
   }
   catch (const array_type2t::array_size_excp &)
   {
-    // Element type has no size at all (flexible/incomplete array member).
+    // Element type has no finite size (flexible/incomplete array member).
     return expr2tc();
   }
 }
@@ -130,22 +109,15 @@ void collect_allocs(
   if (is_if2t(e))
   {
     const if2t &i = to_if2t(e);
-    const expr2tc not_cond = not2tc(i.cond);
-    collect_allocs(
-      i.true_value,
-      is_nil_expr(guard) ? i.cond : and2tc(guard, i.cond),
-      ns,
-      out);
-    collect_allocs(
-      i.false_value,
-      is_nil_expr(guard) ? not_cond : and2tc(guard, not_cond),
-      ns,
-      out);
+    auto extend = [&](const expr2tc &c)
+    { return is_nil_expr(guard) ? c : and2tc(guard, c); };
+    collect_allocs(i.true_value, extend(i.cond), ns, out);
+    collect_allocs(i.false_value, extend(not2tc(i.cond)), ns, out);
     return;
   }
 
-  e->foreach_operand(
-    [&](const expr2tc &op) { collect_allocs(op, guard, ns, out); });
+  e->foreach_operand([&](const expr2tc &op)
+                     { collect_allocs(op, guard, ns, out); });
 }
 } // namespace
 
@@ -189,8 +161,10 @@ bool goto_check_excessive_alloc::runOnFunction(
       work.emplace_back(it, std::move(sites));
   }
 
+  if (work.empty())
+    return false;
+
   const expr2tc bound = constant_int2tc(size_type2(), max_alloc_bytes);
-  bool changed = false;
   for (auto &[it, sites] : work)
   {
     // Snapshot before inserting: each insert_swap splices a new ASSERT into
@@ -204,7 +178,7 @@ bool goto_check_excessive_alloc::runOnFunction(
       // conditional so a not-taken branch cannot raise a false witness.
       expr2tc cond = lessthanequal2tc(site.byte_size, bound);
       if (!is_nil_expr(site.guard))
-        cond = or2tc(not2tc(site.guard), cond);
+        cond = implies2tc(site.guard, cond);
 
       goto_programt new_code;
       goto_programt::targett t = new_code.add_instruction(ASSERT);
@@ -215,8 +189,7 @@ bool goto_check_excessive_alloc::runOnFunction(
       t->function = fn;
       body.insert_swap(it, new_code.instructions.front());
     }
-    changed = true;
   }
 
-  return changed;
+  return true;
 }
--- a/src/goto-programs/goto_check_excessive_alloc.cpp
+++ b/src/goto-programs/goto_check_excessive_alloc.cpp
@@ -26,7 +26,8 @@ const char *alloc_fn_name(sideeffect2t::allockind kind)
 }
 
 /// Total requested allocation size in bytes for `se`, or a nil expression
-/// when it cannot be computed (an element type with no static size).
+/// when it cannot be computed (an element type with no size at all, i.e. a
+/// flexible/incomplete array member).
 ///
 /// The allocation model (`goto_symext::symex_mem`, src/goto-symex/
 /// builtin_functions/memory_alloc.cpp) uniformly allocates `se.size` elements
@@ -62,9 +63,29 @@ expr2tc alloc_byte_size(const sideeffect2t &se, const namespacet &ns)
     return mul2tc(
       size_type2(), size, constant_int2tc(size_type2(), elem_bytes));
   }
+  catch (const array_type2t::dyn_sized_array_excp &)
+  {
+    // A dynamically-sized (VLA) element type, e.g. malloc(sizeof(int[n])):
+    // the frontend records size == 1 with a VLA alloctype, so the byte size
+    // is symbolic but still computable. type_byte_size_expr handles dyn-sized
+    // arrays (unlike the BigInt type_byte_size above, which throws), so scale
+    // by the symbolic element size rather than skipping — otherwise a large
+    // runtime n slips under the bound unchecked.
+    try
+    {
+      expr2tc elem_bytes = type_byte_size_expr(se.alloctype, &ns);
+      if (elem_bytes->type != size_type2())
+        elem_bytes = typecast2tc(size_type2(), elem_bytes);
+      return mul2tc(size_type2(), size, elem_bytes);
+    }
+    catch (const array_type2t::array_size_excp &)
+    {
+      return expr2tc(); // nested element size still unknown — skip
+    }
+  }
   catch (const array_type2t::array_size_excp &)
   {
-    // Element type has no static byte size (flexible/infinite array member).
+    // Element type has no size at all (flexible/incomplete array member).
     return expr2tc();
   }
 }
--- a/src/goto-programs/goto_check_excessive_alloc.cpp
+++ b/src/goto-programs/goto_check_excessive_alloc.cpp
@@ -26,10 +26,17 @@ const char *alloc_fn_name(sideeffect2t::allockind kind)
 }
 
 /// Total requested allocation size in bytes for `se`, or a nil expression
-/// when it cannot be computed (a size-one allocation, or an element type with
-/// no static size). `malloc`/`realloc` sizes are already in bytes;
-/// `operator new[]` carries an element count that is scaled by
-/// `sizeof(element)`.
+/// when it cannot be computed (an element type with no static size).
+///
+/// The allocation model (`goto_symext::symex_mem`, src/goto-symex/
+/// builtin_functions/memory_alloc.cpp) uniformly allocates `se.size` elements
+/// of `se.alloctype`, so the requested byte count is `size * sizeof(alloctype)`
+/// for every kind we check. This matters for typed requests: the frontend
+/// lowers a bare `malloc(sizeof(T))` to `size == 1, alloctype == T` (never a
+/// raw byte count), so without the scaling a large `T` would be checked as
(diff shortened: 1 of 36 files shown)


src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,179 @@
...
+  namespacet ns(context);
+  goto_programt &body = F.second.body;
+
+  // Collect the target assignments (with their allocation sites) first:
+  // inserting an ASSERT before an allocation shifts that allocation to the
+  // next slot, so a single forward scan that inserted in place would
+  // re-encounter and re-instrument it.
+  std::vector<std::pair<goto_programt::targett, std::vector<alloc_site_t>>>
+    work;
+  for (auto it = body.instructions.begin(); it != body.instructions.end(); ++it)
```
lucasccordeiro, line 155: This scans *every* function, including linked C library operational models. `calloc` (`stdlib.c:108`), `strdup` (`string.c:271` `malloc(bufsz*sizeof(char))`), `kmalloc` (`kernel.c`), `ldv_malloc`, and `realloc`'s own body all contain `malloc`/`realloc` side-effects that get instrumented here. Two consequences: (a) a `calloc(n, sz)` violation is reported as `excessive allocation size: malloc` at the *library* location, not the user's `calloc` call site — and there is **no calloc regression test** despite calloc being a headline-supported allocator; (b) any program calling a library routine that internally allocates a nondet/input-sized buffer (e.g. `strdup` of a symbolic-length string, `getline`) can raise a spurious CWE-789 *inside the library* even when user code has no excessive allocation. Recommend restricting instrumentation to non-OM/user functions (or documenting this), and adding a calloc test.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,179 @@
...
+    // Snapshot before inserting: each insert_swap splices a new ASSERT into
+    // the slot `it` names and shifts the allocation one step right, so the
+    // location/function are read from the allocation exactly once.
+    const locationt loc = it->location;
+    const irep_idt fn = it->function;
+    for (const alloc_site_t &site : sites)
+    {
+      // `size <= K`, weakened to `guard => size <= K` when the allocation is
+      // conditional so a not-taken branch cannot raise a false witness.
+      expr2tc cond = lessthanequal2tc(site.byte_size, bound);
```
lucasccordeiro, line 180: The `<=` bound is exact but untested at the boundary: every test uses nondet vs the 1 MiB default, none pins `--excessive-alloc-check=K` with an allocation of exactly `K` (must pass) vs `K+1` (must fail). Add a small-`K` boundary test so an off-by-one here can't regress silently — the PR claims exactness was verified manually; encode it.

src/esbmc/esbmc_parseoptions.cpp
```diff
@@ -844,6 +845,21 @@ int esbmc_parseoptionst::doit()
       goto_preprocess_algorithms.emplace_back(
         std::make_unique<goto_check_unchecked_return>(context));
 
+    // Excessive-allocation-size check (CWE-789). The bound K is the byte
+    // limit above which an allocation size is flagged; a bare flag uses the
+    // implicit 1 MiB default (see options.cpp).
+    if (cmdline.isset("excessive-alloc-check"))
+    {
+      int k = atoi(cmdline.getval("excessive-alloc-check"));
```
lucasccordeiro: `atoi(cmdline.getval(...))` re-parses a value boost has already validated as `int` (`value<int>` rejects out-of-range/garbage at parse time and cmdlinet stringifies the int back). `atoi` is redundant and silently yields `0` on any unexpected non-numeric content, routing to the "requires K>=1" rejection. Read the value once, or note that validity comes from boost, not atoi. Low priority.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  // Collect the target assignments (with their allocation sites) first:
+  // inserting an ASSERT before an allocation shifts that allocation to the
+  // next slot, so a single forward scan that inserted in place would
+  // re-encounter and re-instrument it.
+  std::vector<std::pair<goto_programt::targett, std::vector<alloc_site_t>>>
+    work;
+  for (auto it = body.instructions.begin(); it != body.instructions.end(); ++it)
+  {
+    if (!it->is_assign())
```
lucasccordeiro, line 157: The scan only covers allocations reachable from an `is_assign()` RHS. Confirm the frontend always lowers a discarded allocation (`malloc(n);` as a statement, or an allocation surviving in a `CODE_EXPRESSION`/`FUNCTION_CALL` context) to an assignment; if a bare allocation-expression can reach here as a non-assign instruction, it is silently skipped. Likely fine, but worth confirming/documenting given the narrow scan.

lucasccordeiro (commented):
Actionable points below. Non-issues I verified (no action needed): `insert_swap` multi-site ordering is correct — both asserts land before the allocation and jumps are preserved (`goto_program.h:583`); `implicit_value(1048576)` does populate `getval`, so the bare-flag 1 MiB default works; `catch (array_size_excp&)` correctly covers the `inf_sized`/`dyn_sized` subclasses (`irep2_type.h:322,334`); the modular overflow in `count*sizeof` is false-negative-only as documented; doc counts (26→27 kinds, 31→32 ids) are consistent.

lucasccordeiro:
Follow-up pushed, fixing the false-negative Codex caught: `malloc(sizeof(T))` lowers to `size=1, alloctype=T`, so the check was comparing `1 <= K` and never firing for large types. `alloc_byte_size` now computes `size * sizeof(alloctype)` uniformly across malloc/realloc/`new[]`, matching `symex_mem` (which allocates `size` elements of `alloctype`); byte-count requests keep `alloctype==char`, so scaling is the identity there. Verified a 2 MiB struct now trips CWE-789.

Added the missing tests (typed malloc, exact-K/K+1 boundary, calloc) and documented that model allocators (calloc/strdup) are instrumented in their OM bodies (finding located there) and that a discarded allocation is elided by the frontend before the pass. Left the `atoi` as-is — it matches the codebase idiom and boost pre-validates `value<int>`.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,222 @@
...
+      continue;
+    std::vector<alloc_site_t> sites;
+    collect_allocs(to_code_assign2t(it->code).source, expr2tc(), ns, sites);
+    if (!sites.empty())
+      work.emplace_back(it, std::move(sites));
+  }
+
+  const expr2tc bound = constant_int2tc(size_type2(), max_alloc_bytes);
```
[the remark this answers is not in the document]
pmatos: I don't think this can occur, so I've left the encoding as-is. ESBMC has no 16-bit `size_t` target: `--16` selects the **LP32** data model (`config.cpp`, `dm = LP32`), where `int` is 16-bit but `long`/pointer/`size_t` are 32-bit. The default 1,048,576-byte (2^20) bound is therefore comfortably representable and does not wrap.

Empirically, `--16 --excessive-alloc-check` on a bounded symbolic allocation emits `ASSERT (unsigned long)n * 1 <= 1048576` and verifies SUCCESSFUL — not `<= 0`. And since the option is `value<int>` (max `INT_MAX` = 2^31-1 < 2^32), even the largest user-supplied `K` fits in the 32-bit `size_type` on every target (LP32/ILP32/LP64), so the bound can never be encoded modulo the width. Happy to revisit if there's a target where `size_type` is narrower than 21 bits.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+  if (is_nil_expr(se.size))
+    return expr2tc();
+
+  expr2tc size = se.size;
+  if (size->type != size_type2())
+    size = typecast2tc(size_type2(), size);
+
+  if (se.kind != sideeffect2t::allockind::cpp_new_arr)
+    return size; // already a byte count
```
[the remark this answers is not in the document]
pmatos: Addressed. `alloc_byte_size` now scales typed requests by `sizeof(alloctype)` (via `type_byte_size_expr`), so `malloc(sizeof(struct Huge))` is checked as `sizeof(struct Huge) <= K` rather than `1 <= K`; the same scaling covers the `malloc` branch of `realloc(NULL, sizeof(T))`. Pinned by the `cwe_excessive_alloc_typed` (FAILED) and `cwe_excessive_alloc_typed_pass` (SUCCESSFUL) regression tests.

src/goto-programs/goto_check_excessive_alloc.cpp
pmatos: Addressed both points. (1) Instrumenting the OM bodies is deliberate — it is what gives model-based allocators (`calloc`, `strdup`, `kmalloc`, `ldv_malloc`, …) their CWE-789 coverage; the trade-off (a finding reported at the model's `malloc` site rather than the user's call site, and a library routine that allocates an input-sized buffer surfacing a genuine but library-located CWE-789) is now documented in the pass header comment in `goto_check_excessive_alloc.cpp` and in `website/content/docs/cwe-mapping.md`. (2) Added the `cwe_excessive_alloc_calloc` regression test.

src/goto-programs/goto_check_excessive_alloc.cpp
pmatos: Addressed. Added `cwe_excessive_alloc_boundary_ok` (`malloc(100)` with `--excessive-alloc-check=100` -> SUCCESSFUL) and `cwe_excessive_alloc_boundary_fail` (`malloc(100)` with `--excessive-alloc-check=99` -> FAILED), which pin the exact `size <= K` boundary against a silent off-by-one. A `cwe_excessive_alloc_invalid` test also pins the `K <= 0` rejection.

src/esbmc/esbmc_parseoptions.cpp
pmatos: Addressed in 342efc711d with a note. `atoi(cmdline.getval(...))` follows the existing convention (`--unwind` and `--k-path-coverage` read their values the same way), and boost's `value<int>` has already validated the option as an int at parse time — so `atoi` only reads it back and never sees the non-numeric input that would make it silently yield 0. I kept the pattern for consistency with the surrounding options and documented the invariant.

src/goto-programs/goto_check_excessive_alloc.cpp
pmatos: Confirmed and documented in `runOnFunction`. Every allocation that survives into the GOTO program lands on an assignment RHS: a discarded allocation such as `malloc(n);` is elided by the frontend (`goto_convertt::do_mem` returns for a nil LHS) before this pass runs, so there is no bare allocation-expression instruction for the assignment-only scan to skip.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,201 @@
...
+    BigInt elem_bytes = type_byte_size(se.alloctype, &ns);
+    if (elem_bytes == 1)
+      return size; // sizeof(char)-equivalent: avoid a redundant `* 1`
+    return mul2tc(
+      size_type2(), size, constant_int2tc(size_type2(), elem_bytes));
+  }
+  catch (const array_type2t::array_size_excp &)
+  {
+    // Element type has no static byte size (flexible/infinite array member).
+    return expr2tc();
```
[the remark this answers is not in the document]
pmatos: Addressed. `alloc_byte_size` scales via `type_byte_size_expr`, which returns a symbolic byte size for a dynamically-sized (VLA) element type instead of throwing — so `malloc(sizeof(int[n]))` is checked as `n * sizeof(int) <= K` with symbolic `n`. Pinned by `cwe_excessive_alloc_vla` (FAILED) and `cwe_excessive_alloc_vla_pass` (SUCCESSFUL). The pass was since simplified to a single `type_byte_size_expr` path that covers both fixed and VLA element types.

src/goto-programs/goto_check_excessive_alloc.cpp
```diff
@@ -0,0 +1,222 @@
...
+  // sufficient: every allocation that survives to the GOTO program lands on an
+  // assignment (an allocation whose result is discarded, `malloc(n);`, is
+  // elided by the frontend before this pass runs, so there is nothing to
+  // check).
+  std::vector<std::pair<goto_programt::targett, std::vector<alloc_site_t>>>
+    work;
+  for (auto it = body.instructions.begin(); it != body.instructions.end(); ++it)
+  {
+    if (!it->is_assign())
+      continue;
```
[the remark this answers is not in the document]
pmatos: This is a frontend modeling limitation rather than something this pass can act on. `goto_convertt::do_mem` returns early for a nil LHS, so a discarded `malloc(n);` is dropped entirely during goto conversion — there is no allocation side-effect left in the GOTO program for the pass to instrument. Flagging it would require the frontend to first model discarded allocations (they currently have no effect in ESBMC's memory model), which is out of scope for this CWE-789 check. The limitation is documented in `runOnFunction`.

PR #5788, opened 2026-07-03
[esbmc] surface dead-store assignments as advisory CWE-563

### LINKED ISSUE #4496: [esbmc] surface dead-store assignments as advisory CWE-563
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). ESBMC's slicing pass (`src/goto-symex/slice.cpp`) already drops assignments whose LHS is not used downstream — that information is currently used to shrink the SSA before encoding, but it is *not* surfaced as an advisory finding.

### CWE-563 — Assignment to Variable without Use

[CWE-563](https://cwe.mitre.org/data/definitions/563.html) (`ALLOWED` for vulnerability mapping in CWE 4.20, "Assignment to Variable without Use" / "Unused Variable") is the property: an assignment whose value is never observed on any subsequent path. Dead stores are not bugs in themselves but are a strong code-quality signal and frequently mask intended logic that has been silently broken.

This is information ESBMC already computes; it is not currently exposed.

### Proposal

1. Surface slicing's dead-assignment set as an advisory finding (gated behind `--dead-store-check`). When the slicer drops `lhs = rhs` at location L because `lhs` is never read on any reachable path, emit an advisory with comment `dead store: assignment to <lhs> never read`.
2. As with CWE-561, advisory findings emit as SARIF `result.level = "note"` and do not flip the textual verdict; they appear in a separate `advisories[]` block in the JSON trace.
3. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping `"dead store"` to `{ "dead-store", "Dead store", {563} }`.
4. Add the CWE-563 short name to `cwe_name`.
5. Unit + regression tests, paralleling CWE-561.
6. Update `docs/cwe-mapping.md` (table + count + advisory note).

### Care points

- Slicing also removes assignments that are *uninteresting for the property* (e.g. an assignment to an irrelevant ghost variable). Only assignments removed because their LHS is unread on all paths qualify as CWE-563. The slicer would need to expose that distinction, or the advisory pass would need to recompute it from the goto program independently.
- This is purely intra-procedural for now; inter-procedural dead-store detection is a follow-up.

### Acceptance criteria

- For a program with a clearly dead store (`int x = 5; x = 6; return x;` → the `x = 5` is dead), `esbmc dead_store.c --dead-store-check` emits a note-level finding with `CWE: CWE-563`.
- SV-COMP wrapper and existing regressions are unaffected when the flag is off.
- SARIF includes CWE-563 in `taxa` with `level = "note"`.

pmatos:
## Summary

- Add an opt-in `--dead-store-check` that surfaces **dead-store assignments** (an assignment whose value is never read on any subsequent path) as advisory **CWE-563** ("Assignment to Variable without Use") findings.
- Introduce a new intra-procedural **backward live-variable analysis** over each function's GOTO CFG (`goto_check_dead_store`, a non-mutating `goto_functions_algorithm`). It runs **before** `mark_decl_as_non_det` so synthetic `x = nondet` stores are not mistaken for dead stores.
- Advisories are **note-level** and never change the verdict: they appear textually (`dead store: assignment to <lhs> never read` + a `CWE: CWE-563` line) and in SARIF as `result.level = "note"` with CWE-563 in `taxa`. SARIF is emitted regardless of verdict (advisory-only doc on SUCCESS; merged with violations on FAILED).
- Map `"dead store"` → `{ "dead-store", "Dead store", {563} }` and add the CWE-563 short name in `src/util/cwe_mapping.cpp` (single source of truth reused by SARIF/textual output).

This is ESBMC's first *advisory* CWE finding (distinct from the property-violation CWEs from #4488). Scope is intentionally focused: `--dead-store-check` + analysis + textual + SARIF note. The separate `advisories[]` JSON-trace block and any general advisory framework are deferred as follow-ups. Inter-procedural detection is also a follow-up.

### Soundness

Only automatic-storage, non-`extern`, **non-address-taken**, **non-`volatile`** scalar locals in user source are tracked — excluding address-taken variables keeps the analysis sound without an alias analysis, and volatile writes are observable side effects (C11 §5.1.2.3), not dead stores. Partial/indirect writes (`a[i]`, `*p`, `s.f`) are treated conservatively as uses, never as kills; a store that is live on *any* path is not reported. Operational-model/library sources are excluded from reporting.

Fixes #4496

## Test plan

- [x] `ctest -R cwe_dead_store` — 3 new regression tests pass (textual note; SARIF note-level; FAILED+advisory SARIF merge)
- [x] `cwe_mapping_test` unit test — extended for CWE-563; 18 cases / 620 assertions pass
- [x] Existing `cwe`-labelled regressions (incl. `cwe_sarif`) — 42/42 pass (no SARIF regression)
- [x] Broader core suite sample — 4000+ tests pass, no regressions
- [x] clang-format 11 clean
- [x] Manual acceptance: `int x=5; x=6; return x;` → note + `CWE: CWE-563` + `VERIFICATION SUCCESSFUL`; SARIF `taxa` includes 563 at `level:"note"`; flag-off unaffected
- [x] Soundness spot-checks: address-taken, volatile, loop back-edge, branch-live, and library-model cases all correctly produce no false positive


src/esbmc/bmc.cpp
```diff
@@ -121,6 +121,14 @@ void bmct::successful_trace(const symex_target_equationt &eq [[maybe_unused]])
 
   if (witness_yaml_output != "")
     correctness_yaml_goto_trace(options, ns, goto_trace);
+
+  // On a successful verification there is no error trace, but dead-store
+  // advisories (CWE-563) are still valid and must reach SARIF. Emit an
+  // advisory-only document; guarded so flag-off runs write nothing new.
+  if (
+    !dead_store_advisories.empty() &&
```
lucasccordeiro, line 130: The advisory SARIF is coupled to the trace reporters: it's written only from `successful_trace` (reached on `P_UNSATISFIABLE` under `!bs`) and `error_trace` (`!bs && show_cex`). So a FAILED run with cex display suppressed (`--result-only`/`--no-cex`), or a success path that doesn't route through `successful_trace`, silently drops the SARIF advisory — while the textual advisory still prints (it's emitted unconditionally in `process_goto_program`). Emit the advisory SARIF once from a verdict-independent point instead. Related: `--dead-store-check` rejects no strategy combos and there's no `--k-induction`/`--incremental-bmc`-success test, so this matrix is unverified (the one `--incremental-bmc` test is a FAILED case).

src/goto-programs/dead_store_analysis.cpp
```diff
@@ -0,0 +1,290 @@
...
+/// before inlining, so every linked OM/library function with a body is still
+/// present — those live under the extracted-headers temp dir ("-headers-") at
+/// runtime and under c2goto/library/ for in-tree builds (mirrors
+/// remove_exceptions.cpp::is_library_function).
+bool is_reportable_location(const locationt &loc)
+{
+  const std::string file = loc.get_file().as_string();
+  if (file.empty() || file == "<built-in>" || file == "<builtin>")
+    return false;
+  if (has_prefix(file, "/usr/"))
```
lucasccordeiro: `is_reportable_location` filters `/usr/` only, which isn't portable. macOS SDK headers (`/Library/Developer/...`, `/Applications/Xcode...`) and Windows paths won't match, so dead stores in `static inline` header functions get reported at a system-header location there; conversely legitimate user code under `/usr/local/...` is wrongly suppressed. The `-headers-` catch only covers ESBMC's bundled headers. Prefer keying off the symbol's/location's system-header flag rather than hard-coded path prefixes.

src/goto-programs/dead_store_analysis.cpp
```diff
@@ -0,0 +1,290 @@
...
+    irep_idt name;
+  };
+  std::vector<sitet> sites;
+
+  for (size_t i = 0; i < n; ++i)
+  {
+    const auto &it = insns[i];
+
+    goto_programt::const_targetst succs;
+    body.get_successors(it, succs);
```
lucasccordeiro, line 205: `get_successors` here sees only normal edges: this pass runs before the throw/catch -> guarded-control-flow lowering (#5075). A value written in a `try` body and read only in a `catch` handler has no successor edge to the handler, so its store is misreported as dead — a false positive on C++ inputs. Either run after exception lowering, exclude functions containing `THROW`/`CATCH`, or document C++ as out of scope.

regression/esbmc/cwe_dead_store/test.desc
```diff
@@ -0,0 +1,6 @@
+CORE
+main.c
+--dead-store-check
+^VERIFICATION SUCCESSFUL$
+dead store: assignment to x never read
```
lucasccordeiro, line 5: All three tests use the same "overwritten-then-read" shape (`x=5; x=6; return x`), which only exercises the non-final store. Missing: (a) a no-false-positive negative test for an address-taken and a volatile local — the core soundness claim, currently only spot-checked; (b) a final dead store `int x=5; x=6; /* x never read */`, the textbook CWE-563 shape and the one that exercises the `DEAD`-instruction handling. Add at least these two.

lucasccordeiro (commented):
Four actionable points inline. Minor notes: the dense `std::vector<std::vector<bool>>` liveness with a round-robin fixpoint is O(n^2 * bits) worst case — a per-node bitset + worklist would scale far better on large functions (opt-in mitigates); and `dead_store_advisoryt::function` is populated (`loc.get_function()`) but never read by any consumer — drop it or wire it into the output.

Verified as non-issues (no action): `goto_convert` always splits initialized local decls into `DECL`+`ASSIGN` (`goto_convert.cpp:753`), so the `is_decl` branch ignoring `code_decl2t.init` is safe (init is nil at goto level); `code_dead2t` carries its symbol as an `irep_idt`, not an operand, so `DEAD x` correctly doesn't count as a use; the address-taken scan is per-function which suffices for locals; `x = x + 1` self-reference and loop back-edges are handled correctly by the meet.

lucasccordeiro:
Follow-up pushed. Heads-up beyond the review: **`--dead-store-check` segfaulted on any C++ input.** The linked C++ exception model has a void `terminate()` call whose return operand is a null expr, and the pass called `is_symbol2t(fc.ret)` without the null-safe `is_nil_expr` guard. Fixed, with a C++ crash regression added (there was no C++ test, so CI never hit it).

Also addressed the review points: record call-return stores (`x = f();` with `x` unused); skip functions containing `throw`/`catch` (pass runs before exception lowering, so a catch-only read would be a false positive); emit the advisory SARIF from a verdict-independent point (works under `--result-only`); make `is_reportable_location` portable (keep `/usr/local`, add macOS SDK roots); drop the unused `function` field. Added final-dead-store, call-return, address-taken/volatile-negative, and result-only-SARIF tests.

PR #5797, opened 2026-07-03
[esbmc] map non-termination verdicts to CWE-835 (infinite loop)

### LINKED ISSUE #4490: [esbmc] map non-termination verdicts to CWE-835 (infinite loop)
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). ESBMC's `--termination` strategy (`src/esbmc/options.cpp:395`) already proves or refutes loop termination via k-induction, but a non-terminating loop verdict is not tagged with any CWE.

### CWE-835 — Loop with Unreachable Exit Condition (Infinite Loop)

[CWE-835](https://cwe.mitre.org/data/definitions/835.html) (`ALLOWED` for vulnerability mapping in CWE 4.20) is exactly the property `--termination` refutes: the loop's exit condition is unreachable from the loop body. ESBMC has the soundness to claim this finding via k-induction convergence.

### Proposal

1. Identify the assertion comment string emitted when `--termination` reports non-termination (probably under `src/goto-programs/goto_loops.h` or `src/esbmc/bmc.cpp`'s termination strategy — needs short investigation).
2. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping that substring (e.g., `"non-termination"` / `"loop does not terminate"`) to `{ "infinite-loop", "Loop with unreachable exit condition", {835} }`.
3. Update `cwe_name` in `src/util/cwe_mapping.cpp` with the CWE-835 short name "Loop with Unreachable Exit Condition ('Infinite Loop')".
4. Add a unit test in `unit/util/cwe_mapping.test.cpp`.
5. Add a regression test `regression/esbmc/cwe_infinite_loop/` with a minimal `while(1) {}`-style benchmark and `--termination`, asserting `CWE-835` in the trace.
6. Add the row to `docs/cwe-mapping.md` and bump the "28 unique CWEs across 23 violation kinds" count.

### Acceptance criteria

- `esbmc infinite.c --termination` prints `CWE: CWE-835` after the violated-property comment when a loop is proven non-terminating.
- SARIF output includes CWE-835 in `taxa` and the `taxonomies` block.
- The new regression test passes; SV-COMP wrapper classification is unchanged.
- `docs/cwe-mapping.md` count updated.

Note: This issue is for **non-termination findings** (CWE-835), which is distinct from the existing `unwinding assertion` situation. Unwinding-assertion failures signal "k-bound was insufficient", which is not a weakness per se and remains intentionally unmapped.

pmatos:
## Summary

- `--termination` refutes the termination property by proving a loop's exit condition unreachable — exactly [CWE-835](https://cwe.mitre.org/data/definitions/835.html) ("Loop with Unreachable Exit Condition / Infinite Loop"). This wires that verdict into ESBMC's CWE annotation.
- A non-termination verdict is proven by **UNSAT** and therefore has **no counterexample trace**, unlike every other CWE-annotated violation. To keep a single source of truth, `report_non_termination_cwe()` synthesises a one-step `goto_tracet` anchored to the loop's exit-marker ASSERT and feeds it to the existing SARIF / JSON / GraphML emitters; the text output gains a `CWE: CWE-835` line right after the `... non-terminating execution` verdict.
- The substring key `non-terminating execution` (already present verbatim in both verdict messages) maps to `{"infinite-loop", "Loop with unreachable exit condition", {835}}` in `util/cwe_mapping`; verdict text lines and `VERIFICATION FAILED` stay on their own lines, so the SV-COMP wrapper's substring classification is unchanged.
- A per-loop marker is preferred as the anchor (abort-call marker is a fallback); the YAML witness format has no CWE field, so it is intentionally left untouched.
- Docs updated: 31/26 → 32/27 unique CWEs / violation kinds, new mapping-table row, and a note on the traceless non-termination case.

Fixes #4490

## Test plan

- [x] `unit/util/cwe_mapping_test` — 623 assertions pass (adds CWE-835 rule/name/verdict-string cases)
- [x] New regression tests under `regression/termination/`:
  - `cwe_infinite_loop` — text `CWE: CWE-835` after the verdict
  - `cwe_infinite_loop_sarif` — SARIF `taxa` + `taxonomies` carry 835 (`--sarif-output -`)
  - `cwe_infinite_loop_json` — `report.json` `assertion.cwe[0] == 835` (`CHECK_JSON`)
  - `cwe_terminating_no_835` — terminating loop → `VERIFICATION SUCCESSFUL`, no CWE (passing/negative case)
- [x] Existing `regression/termination/` suite — 65/65 pass (no regression)
- [x] Existing `regression/esbmc*/cwe_*` suite — 20/20 pass (shared mapping table unaffected)
- [x] All four channels verified end-to-end (text, SARIF, JSON, GraphML)


pmatos:
### Code review

Found 2 issues:

1. The SARIF/JSON/GraphML annotation anchors to the wrong location. `report_non_termination_cwe()` picks the first `"termination per-loop marker"` in `function_map` order, but `function_map` is ordered by mangled id, so the hidden library helper `__ESBMC_atexit_handler` (which has its own loop marker and sorts before `main`) wins. For `while(1){}` I confirmed the SARIF `artifactLocation.uri` is `.../src/c2goto/library/stdlib.c` at `startLine: 43` — ESBMC's own install path, not the user's loop — and the JSON report even dumps all of `stdlib.c` into `source_files`. The docstring's "best-effort loop" caveat doesn't cover pointing at an internal library file. Suggest skipping `fn.second.body.hide` functions (and/or preferring the entry function) in the scan.

https://github.com/esbmc/esbmc/blob/f8bb90f71fb86b8e20fcd8cf073faabd6e2bf7b4/src/esbmc/esbmc_parseoptions.cpp#L1564-L1598

2. The doc count is stale. `upstream/master` already documents **34 unique CWEs across 28 violation kinds** (563, 674, 122, … merged in the ~155 commits since this branch forked), so `32`/`27` here would regress the count on merge; after rebasing this should be 35/29 with the new ids added to the list.

https://github.com/esbmc/esbmc/blob/f8bb90f71fb86b8e20fcd8cf073faabd6e2bf7b4/website/content/docs/cwe-mapping.md#L11-L16

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lucasccordeiro:
Thanks — both confirmed and addressed.

**1. Wrong anchor location.** Reproduced exactly as you described: for `while(1){}` the SARIF `artifactLocation.uri` was `.../src/c2goto/library/stdlib.c` at `startLine: 43`, and the JSON report's `steps[0].function` was `__ESBMC_atexit_handler`. Root cause is as you diagnosed — `goto_termination` deliberately runs the marker pass on hidden helpers (`goto_termination.cpp:765-767`, so a program whose only loop is a library call still gets a claim), and `function_map` is a `std::map` with `id_string_order`, so `c:@F@__ESBMC_atexit_handler` is visited before `c:@F@main`.

Fixed in b0f5219 by ranking the candidates rather than taking the first hit:

```
USER_LOOP < USER_ABORT < LIB_LOOP < LIB_ABORT
```

keyed on `fn.second.body.hide` — the same field `goto_termination.cpp` itself uses to classify helpers. Note this also revives the abort-call fallback, which was unreachable in practice: `__ESBMC_atexit_handler`'s `while (atexit_count > 0)` is linked into *every* program, so its per-loop marker always beat any abort-call marker under the old "per-loop always wins" rule. After the fix the anchor is `main.c` at the loop exit, and `source_files` no longer drags in `stdlib.c`.

Regression coverage was the real gap here — none of the four existing tests asserted the *location*, which is why this slipped through:

- `regression/termination/cwe_infinite_loop_anchor/` (new) pins the SARIF uri.
- `cwe_infinite_loop_json/test.desc` gains `CHECK_JSON report.json $[0].steps[0].function == "main"`.

Both were verified to fail on the unpatched binary and pass on the patched one. I used `function` rather than `file` for the JSON check because ctest passes an absolute source path, so `file` is path-dependent. Full `regression/termination` suite: no new failures (`ranking_nested_inner_in_if_branch`, `ranking_nested_inner_recognize_through_outer` and `termination13` fail on plain `master` too, unrelated to this PR).

The docstring's "best-effort loop" caveat and the `cwe-mapping.md` section are updated to state that library-helper markers rank below user markers, so the anchor never lands in ESBMC's installed sources.

**2. Stale doc count.** Right — `master` had moved on. The branch is now merged up (1dc6186); the summary reads **35 unique CWEs across 29 violation kinds**, taking the union of #5784's CWE-122 and this PR's CWE-835, and both the "Heap vs. stack out-of-bounds" and "Non-termination (CWE-835)" sections are kept.

PR #5854, opened 2026-07-05
[python] Model the try/except/else clause

lucasccordeiro:
## What

Python's `try/except/else` — the `else` clause runs only when the try body completes without an exception, and its own exceptions are not caught by the try's handlers. ESBMC **silently dropped** the `else` clause (a soundness gap: its side effects were lost, so e.g. `else: x = 2` never took effect).

## Fix

Model it with a guard flag in `get_try_statement`:
1. Create a bool `$try_else_guard$`, declared and set `false` before the cpp-catch.
2. Append `guard = true` as the last statement of the try body — reached only on the no-exception path (any raise diverts to a handler first).
3. After the catch construct, emit `if (guard): <else body>`, so the else runs outside the handler scope — its own exceptions propagate rather than being caught.

`finally + else` remains refused with a clean diagnostic (the finally-duplication lowering would interleave incorrectly with the guard).

## Tests

- `try_except_else` (CORE) — else runs on the no-exception path, is skipped when the body raises and a handler catches it, and works with multiple handlers; also verified in a loop (guard resets each iteration) and nested.
- `try_except_else_fail` (CORE) — an exception raised in the `else` clause is not caught by the try's handlers and propagates uncaught.

Both CPython-sane; full exception regression clean (100%). Plain try/except (no else) is unchanged.

### DIFF
--- a/regression/python/try_except_else/main.py
+++ b/regression/python/try_except_else/main.py
@@ -1,0 +1,69 @@
+# The try `else` clause runs only when the try body completes without an
+# exception (previously it was silently dropped). It must be skipped when an
+# exception is raised, and its own exceptions must NOT be caught by this try's
+# handlers.
+x = 0
+try:
+    pass
+except ValueError:
+    x = 1
+else:
+    x = 2
+assert x == 2                 # no exception -> else runs
+
+y = 0
+try:
+    raise ValueError()
+except ValueError:
+    y = 1
+else:
+    y = 2
+assert y == 1                 # exception -> else skipped
+
+r = []
+try:
+    r.append(1)
+except ValueError:
+    r.append(9)
+else:
+    r.append(2)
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
Superseded by #5843 ("[python] Run the else clause of a try statement"), which is already merged to master and provides `try/except/else` including the `else`+`finally` combination.

One behavioral difference worth noting for a possible follow-up: this PR ran the `else` body *outside* the try's handler scope (via a `$try_else_guard$` flag), so an exception raised inside `else` propagates rather than being caught by the same try — closer to CPython. The merged #5843 appends `else` into the try body, where an `else`-body exception is caught by this try's handlers (a documented minor deviation). If that fidelity matters, we can revisit the guard-flag approach as an enhancement on top of #5843. Closing as duplicate.

PR #5875, opened 2026-07-05
fix: replace unused esbmc_rc with _ (ShellCheck SC2034)

pierredantas:
## Problem

ShellCheck SC2034 flags `esbmc_rc` as an unused variable in `experiments/nuxmv_comparison/run_experiments.sh`:

```bash
read -r esbmc_time esbmc_rc <<< "$(run_timed "$esbmc_log" ..."
```

The exit code captured in `esbmc_rc` is never used — verdict detection is handled by `esbmc_verdict()` via log parsing, which is the correct approach (ESBMC may exit non-zero even on a valid VIOLATION result when using `|| true`).

## Fix

Replace `esbmc_rc` with the conventional `_` dummy variable to silence the warning without changing behaviour:

```bash
read -r esbmc_time _ <<< "$(run_timed "$esbmc_log" ..."
```

No logic changes.


pierredantas:
Closing: this branch was cut from an older state of the fork and, diffed against current master, re-adds the entire experiments/nuxmv_comparison/ tree (plus benchmarks, a CI workflow, and tools/) that has since been relocated to https://github.com/pierredantas/esbmc-plc-nuxmv-experiments. It also conflicts with master and no longer matches its stated scope. The intended ShellCheck SC2034 fix, if still needed, will be reopened as a fresh single-purpose branch off current master.

PR #5885, opened 2026-07-06
[symex] Flag va_arg on a va_list never initialised by va_start

### LINKED ISSUE #5873: va_arg without a preceding va_start is silently accepted (unsound, false VERIFICATION SUCCESSFUL)
**Status**: Triage
**Environment**: `origin/master` @ `48e32054174c5514776ef1eea68eac2b654b1655`, z3 backend; reference CBMC 6.8.0.

### Summary
`va_arg`'s GOTO lowering never references the `va_list` variable itself, so a real, observable ABI-level bug class (reading through an uninitialized `va_list` because `va_start` was never called) is structurally invisible to ESBMC under any flag combination, producing a false `VERIFICATION SUCCESSFUL`.

### Root cause (confirmed via `--goto-functions-only`, not a hypothesis)
```
# ESBMC: va_arg lowers to a positional builtin, no reference to `ap`
ASSIGN return_value$___ESBMC_va_arg$1 = va_arg(0);

# CBMC: `ap` is a real object; va_start assigns it, va_arg dereferences it
DECL sum::1::ap : empty**
ASSIGN sum::1::ap := side_effect statement="va_start"(cast(address_of(sum::count), empty**))
ASSERT ¬(pointer_object(NULL) = pointer_object(sum::1::ap))  // dereference failure: pointer NULL in *ap
... (10 total automatic dereference-safety asserts on `ap` / `*ap`)
ASSIGN sum::$tmp::return_value_gcc_builtin_va_arg := *(cast(*sum::1::ap, signedbv[32]*))
ASSIGN sum::1::ap := sum::1::ap + 1
```
Whether `va_start` is called or not, ESBMC's GOTO program for `sum()` is identical. This is a frontend lowering design choice, not a solver-stage effect — confirmed directly from the GOTO dump, not inferred from behavior.

### Reproducer
```c
#include <stdarg.h>

int sum(int count, ...) {
    va_list ap;
    /* va_start(ap, count); -- deliberately omitted */
    int total = 0;
    for (int i = 0; i < count; i++)
        total += va_arg(ap, int);
    return total;
}

int main() {
    int r = sum(2, 3, 4);
    __ESBMC_assert(r == 7, "reading va_arg without va_start is undefined behavior and should be caught");
    return 0;
}
```
```
$ esbmc va_start_missing.c --function main --z3 --multi-property
```

### Observed (ESBMC)
```
✓ PASSED: 'reading va_arg without va_start is undefined behavior and should be caught'
Properties: 1 verified, 0 failed
VERIFICATION SUCCESSFUL
```
### Reference (CBMC 6.8.0)
```
$ cbmc va_start_missing_cbmc.c --function main
...
[main.assertion.1] reading va_arg without va_start is undefined behavior and should be caught: FAILURE
[sum.pointer_dereference.1] dereference failure: pointer NULL in *ap: FAILURE
[sum.pointer_dereference.7] dereference failure: pointer NULL in *((signed int *)*ap): FAILURE
[sum.pointer_dereference.2..6,8..12] (10 more dereference-safety checks on *ap / *((int*)*ap)): FAILURE
[sum.overflow.1] arithmetic overflow on signed + in total + return_value_gcc_builtin_va_arg: FAILURE
** 14 of 15 failed
VERIFICATION FAILED
```
CBMC models `ap` as a real pointer object, so the missing initialization is caught as an invalid-pointer dereference (12 dereference-safety properties + the assertion + 1 overflow check, 14 total).

### Classification
Soundness gap, not a completeness/nice-to-have gap — this produces a false-negative `VERIFICATION SUCCESSFUL` on code with a genuine memory-safety bug, for a tool whose stated purpose includes memory-safety/UB verification. The abstraction choice (positional `va_arg` builtin, no modeled `va_list` object) is presumably intentional for scalability, but its soundness cost is currently undocumented.

### Next steps
1. Add to the documented list of known-unsound constructs: `va_arg` does not detect a missing/incorrect `va_start`.
2. Evaluate an opt-in "real `va_list` object" mode (mirroring CBMC's approach: a real pointer object + automatic dereference checks), gated behind a flag (e.g. alongside `--uninitialised-vars-check`) to avoid perf regression on the default path.
3. Short term (if 2 is deferred): add the reproducer above to the regression suite under known-limitations, so it isn't later mistaken for a new regression.

LukeW1999:
Fixes #5873

## Problem

`va_arg`'s GOTO lowering never referenced the `va_list` variable: the side effect was emitted with a dummy zero operand and `va_start`/`va_copy` calls were erased during GOTO conversion. Symex resolves varargs purely positionally via the frame's `va_cursor`, so a `va_arg` on a `va_list` that was never initialised by `va_start` — undefined behaviour (C17 7.16.1) — was silently resolved to the caller's arguments, producing a false `VERIFICATION SUCCESSFUL` on the reproducer in #5873 (CBMC reports 14 property failures on the same program).

## Approach

- **GOTO conversion** (`builtin_functions.cpp`): the `va_arg` side effect now carries the `va_list` lvalue as its operand; `__builtin_va_start`/`__builtin_va_copy` calls are kept in the GOTO program as marker `FUNCTION_CALL`s (no body is ever looked up).
- **Symex** (`va_arg.cpp`, `run_builtin.cpp`): tracks which `va_list` objects have been started, keyed on their level-1 identity (base name, activation, thread), so the same object is recognised across frames, recursion, and loop re-declaration. `run_builtin` intercepts the marker calls; `symex_va_arg` raises `missing va_start: va_arg on an uninitialised va_list` when a local `va_list` is consumed unstarted. The claim is only emitted when violated, so correct code gains no extra VCC.
- **Precision guards**: `va_list`s that cannot be resolved to a local variable (parameters, pointer indirection) are conservatively assumed started; a `va_copy` into a `va_list` reached through a pointer marks every may-point-to target as started (widening only — never erasing on a may-alias basis); marking ignores the path guard. All over-approximation is towards "started", so no correct program can be flagged.
- **Inliner** (`goto_inline.cpp`): skips the marker calls so `--full-inlining` does not erase them.

## Testing

- New regression tests: `github_5873_va_start_missing_fail`, `github_5873_va_copy_unstarted_fail` (both must FAIL with the new claim), `github_5873_va_start_present_pass` (va_start/va_copy chains, `va_list` as parameter, `va_arg`/`va_copy` behind a `va_list*` — must stay SUCCESSFUL).
- Adversarial review covered: conditional `va_start`, recursion, struct-member `va_list`, `va_list*` indirection both directions, `--full-inlining`, `--k-induction`, `--incremental-bmc`, vsnprintf/vasprintf operational models.
- Full `esbmc`-label regression suite: 4308 tests, 0 failures. Unit tests: 455, 0 failures.

## SV-COMP impact (benchexec, 30s per task)

- **This branch**: https://github.com/esbmc/esbmc/actions/runs/28786127764 

  correct:           20686
    correct true:    13841
    correct false:    6845
  incorrect:             5
    incorrect true:      0
    incorrect false:     5
  unknown:           15909
  Score:             34447 (max: 61270)

- **Master baseline (30s, 2026-06-26)**: https://github.com/esbmc/esbmc/actions/runs/28250197549

  correct:           20693
    correct true:    13845
    correct false:    6848
  incorrect:             5
    incorrect true:      0
    incorrect false:     5
  unknown:           15902
  Score:             34458 (max: 61270)

Result comparison (via `benchexec-diff`) to follow once the branch run finishes; the expected impact surface is limited to tasks exercising user-written variadic functions.


lucasccordeiro (approved):

PR #5890, opened 2026-07-06
[solver] Make pointer relational comparison a consistent total order

### LINKED ISSUE #5872: Pointer relational comparison (<=) is not antisymmetric with --no-pointer-relation-check
**Status**: Triage
**Environment**: `origin/master` @ `48e32054174c5514776ef1eea68eac2b654b1655`, z3 backend; reference CBMC 6.8.0.

### Summary
For two unrelated, non-array local variables `a`, `b`, ESBMC's `<=` on their addresses satisfies `(&a<=&b) && (&b<=&a)` simultaneously. This violates antisymmetry, a basic property any (partial or total) order must satisfy (`x<=y ∧ y<=x ⟹ x=y`), independent of any discussion about what value a relational comparison between unrelated pointers "should" produce. Not a design tradeoff; classified below as a soundness bug in the `--no-pointer-relation-check` fallback path.

### Reproducer
```c
#include <stddef.h>

int main() {
    char a;
    char b;
    char *p1 = &a;
    char *p2 = &b;
    int le = (p1 <= p2);
    int ge = (p2 <= p1);
    /* if p1 != p2, at most one of (p1<=p2), (p2<=p1) can hold under any order */
    __ESBMC_assert(!(le && ge), "p1<=p2 and p2<=p1 cannot both hold for two distinct objects");
    return 0;
}
```
```
$ esbmc antisymmetry.c --function main --z3 --no-pointer-relation-check
```

### Observed (ESBMC)
```
[Counterexample]

State 1 file antisymmetry.c line 8 column 5 function main thread 0
  le = (signed int)(&a <= &b)

State 2 file antisymmetry.c line 9 column 5 function main thread 0
  ge = (signed int)(&b <= &a)

State 3 file antisymmetry.c line 11 column 5 function main thread 0
Violated property:
  file antisymmetry.c line 11 column 5 function main
  p1<=p2 and p2<=p1 cannot both hold for two distinct objects
  !((_Bool)le && (_Bool)ge)

VERIFICATION FAILED
```
`le` and `ge` are both satisfied as true in the same model.

### Reference (CBMC 6.8.0)
```
$ cbmc antisymmetry_cbmc.c --function main --no-pointer-check
...
[main.assertion.1] p1<=p2 and p2<=p1 cannot both hold for two distinct objects: SUCCESS
** 0 of 1 failed
VERIFICATION SUCCESSFUL
```
(`--no-pointer-check` is CBMC's analogous flag for bypassing its own same-object advisory check.) CBMC's flat/global bitvector address encoding produces a consistent total order, so antisymmetry holds.

### Scoped out: the default "Same object violation" check
Without `--no-pointer-relation-check`, ESBMC does flag this comparison (`Same object violation`, `SAME-OBJECT(...)`). CBMC has an equivalent check and also flags it. Verified this fires for any cross-object relational comparison, not specifically against `NULL` (confirmed with the `a`/`b` example above, no `NULL` involved). Per C11 6.5.8p5, relational comparison between pointers to different objects is undefined behavior, so both tools are correct here — **this check is not the bug and is excluded from this report.**

### Root cause (hypothesis, not confirmed against source)
Once the same-object guard is bypassed, the relational operator likely lowers to a comparison of only the *offset* component of ESBMC's (object, offset) pointer representation. `a` and `b` are both single, non-array locals, so both addresses have offset 0 relative to their own object, and `0<=0` is trivially true in both directions. Not yet located in source; suspect `src/goto-symex/` around `SAME-OBJECT`/`POINTER_OFFSET` handling for relational ops.

### Classification
Soundness bug in the `--no-pointer-relation-check` fallback semantics, not an acceptable "different but internally consistent" design choice — the property broken (antisymmetry) is a basic algebraic invariant of the operator itself, not a hardware/ABI-dependent value.

### Next steps
1. Locate the offset-only fallback in `src/goto-symex/` (`SAME-OBJECT` / `POINTER_OFFSET` handling for relational ops).
2. Evaluate fix direction: with the same-object check disabled, the comparison result should arguably become nondet (unconstrained) rather than a deterministic offset comparison, so it cannot be simultaneously satisfiable in both directions.
3. Until fixed, add a note to `--no-pointer-relation-check`'s help text that the resulting comparison value is not guaranteed to be antisymmetric.
4. Add the reproducer above as a regression test.

LukeW1999:
Fixes #5872

## Problem

With `--no-pointer-relation-check`, pointer relational operators (`<`, `<=`, `>`, `>=`) lowered to a comparison of **only the offset** component of ESBMC's `(object, offset)` pointer representation (`convert_ptr_cmp` in `src/solvers/smt/smt_memspace.cpp`). Two distinct, non-array locals both have offset 0 relative to their own object, so `(&a <= &b) && (&b <= &a)` was satisfiable simultaneously — a violation of antisymmetry, which any partial or total order must satisfy (`x<=y ∧ y<=x ⟹ x=y`). This produced a false `VERIFICATION FAILED` on the antisymmetry assertion in the reproducer; CBMC's flat address encoding gives a consistent total order and verifies it.

## Approach

Compare the `(object, offset)` pairs **lexicographically** instead of offset-only:

```
lt:  obj1 <  obj2 || (obj1 == obj2 && off1 <  off2)
lte: obj1 <  obj2 || (obj1 == obj2 && off1 <= off2)
gt:  obj1 >  obj2 || (obj1 == obj2 && off1 >  off2)
gte: obj1 >  obj2 || (obj1 == obj2 && off1 >= off2)
```

- **Same object** (the C-defined case, C11 6.5.8p5): reduces to the offset order used before — behaviour unchanged.
- **Across objects** (reachable only when the same-object assertion is disabled): an arbitrary but consistent total order over object ids, so antisymmetry, totality and transitivity of the operators all hold.

Offsets are still typecast to unsigned for the comparison (objects can exceed half the address space). The default same-object relation check is unaffected — this only changes the fallback semantics the user opted into with `--no-pointer-relation-check`. Pointer equality is handled on a separate path (direct tuple comparison in `smt_solver.cpp`) and is not routed through `convert_ptr_cmp`.

## Testing

New regression tests under `regression/esbmc/github_5872_ptr_order_*`:
- `_pass` (`--no-pointer-relation-check`): antisymmetry, totality, strict-order exclusivity, trichotomy for distinct objects, and unchanged in-object offset order — all `VERIFICATION SUCCESSFUL`.
- `_fail` (`--no-pointer-relation-check`): asserting both directions of `<=` hold for distinct objects must `VERIFICATION FAILED`.

Additional adversarial checks (transitivity across 3 objects, reflexivity/irreflexivity, `<`/`>` and `<=`/`>=` mirroring, one-past-the-end vs start ordering, NULL-pointer comparisons, symbolic/conditional pointers) pass in both bitvector and int-encoding (`--ir`) modes. Full `esbmc`-label regression suite: 4307 tests, no failures attributable to this change (one THOROUGH `--incremental-bmc` C++ test flaked on a parallel-load timeout and passes cleanly in isolation).

### DIFF
--- a/regression/esbmc/github_5872_ptr_order_fail/main.c
+++ b/regression/esbmc/github_5872_ptr_order_fail/main.c
@@ -1,0 +1,13 @@
+#include <stddef.h>
+
+/* Both directions of <= holding simultaneously for pointers to two distinct
+ * objects would violate antisymmetry; asserting that they do must fail. */
+int main()
+{
+  char a, b;
+  char *p1 = &a, *p2 = &b;
+  int le = (p1 <= p2);
+  int ge = (p2 <= p1);
+  __ESBMC_assert(le && ge, "both directions cannot hold for distinct objects");
+  return 0;
+}
--- a/regression/esbmc/github_5872_ptr_order_fail/test.desc
+++ b/regression/esbmc/github_5872_ptr_order_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--no-pointer-relation-check
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_5872_ptr_order_pass/main.c
+++ b/regression/esbmc/github_5872_ptr_order_pass/main.c
@@ -1,0 +1,26 @@
+#include <stddef.h>
+
+/* With the same-object check disabled, relational comparison of pointers to
+ * distinct objects must still behave as a consistent total order: within one
+ * object it is the C-defined offset order, across objects the direction is
+ * arbitrary but the algebraic properties of the operator must hold. */
+int main()
+{
+  char a, b;
+  char *p1 = &a, *p2 = &b;
+  int le = (p1 <= p2);
+  int ge = (p2 <= p1);
+  /* antisymmetry: for distinct objects at most one direction holds */
+  __ESBMC_assert(!(le && ge), "antisymmetry");
+  /* totality: at least one direction must hold */
+  __ESBMC_assert(le || ge, "totality");
+  /* strict order consistency */
+  int lt = (p1 < p2), gt = (p1 > p2);
+  __ESBMC_assert(!(lt && gt), "lt/gt exclusive");
+  __ESBMC_assert(lt || gt, "trichotomy for distinct objects");
+  /* same-object comparisons keep the C-defined offset order */
+  char arr[4];
+  __ESBMC_assert(&arr[1] < &arr[3], "in-object order");
+  __ESBMC_assert(!(&arr[3] <= &arr[1]), "in-object strictness");
+  return 0;
+}
--- a/regression/esbmc/github_5872_ptr_order_pass/main.c
+++ b/regression/esbmc/github_5872_ptr_order_pass/main.c
@@ -18,9 +18,12 @@ int main()
   int lt = (p1 < p2), gt = (p1 > p2);
   __ESBMC_assert(!(lt && gt), "lt/gt exclusive");
   __ESBMC_assert(lt || gt, "trichotomy for distinct objects");
+  /* >= and <= describe the same total order in mirror directions */
+  __ESBMC_assert((p1 >= p2) == (p2 <= p1), ">= mirrors <=");
   /* same-object comparisons keep the C-defined offset order */
   char arr[4];
   __ESBMC_assert(&arr[1] < &arr[3], "in-object order");
+  __ESBMC_assert(&arr[3] >= &arr[1], "in-object order (>=)");
   __ESBMC_assert(!(&arr[3] <= &arr[1]), "in-object strictness");
   return 0;
 }
--- a/regression/esbmc/github_5872_ptr_order_pass/test.desc
+++ b/regression/esbmc/github_5872_ptr_order_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--no-pointer-relation-check
+^VERIFICATION SUCCESSFUL$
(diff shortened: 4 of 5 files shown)


LukeW1999:
Follow-up commit `d9753f250d`: the codecov diff flagged one uncovered line — the `greaterthanequal_id` branch of `convert_ptr_cmp`. The pass regression exercised `<`, `>`, and `<=` but never `>=`, so that switch arm was never hit.

Added two assertions to `github_5872_ptr_order_pass`:
- cross-object: `(p1 >= p2) == (p2 <= p1)` — `>=` mirrors `<=` under the same total order
- in-object: `&arr[3] >= &arr[1]`

All four relational operators are now covered; both regression tests still pass (pass → SUCCESSFUL, fail → FAILED). No production-code change.

lucasccordeiro:
@LukeW1999: could you please run this PR over the sv-comp benchmarks?

LukeW1999:
Sure.

LukeW1999:
## SV-COMP benchexec (30s/task)

| | Run | Branch |
|---|---|---|
| This PR | [29019129497](https://github.com/esbmc/esbmc/actions/runs/29019129497) | `fix/pointer-relation-antisymmetry` |
| Master baseline | [28571328376](https://github.com/esbmc/esbmc/actions/runs/28571328376) | `master` (2026-07-02) |

Same config: `timeout=30`, `svbenchmarks=main`, `mode=Full`.

No regression expected: for same-object comparisons (the only C-defined case, and the bulk of SV-COMP) the new encoding is logically identical to the old offset-only one; only UB cross-object comparisons change. Score diff to follow once the run finishes.

lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/29092875285
Master (30s): https://github.com/esbmc/esbmc/actions/runs/29092897193

PR #5907, opened 2026-07-07
[goto-programs] Type-aware uncaught-exception verdicts

### LINKED ISSUE #5903: [python] Uncaught-exception verdict is type-less: different exceptions collapse to one "uncaught exception" property
### What the SV-COMP Python track needs

The track scores each task against **one exception family at a time**. Each property file targets a single exception type, e.g.:

```
data-lookup.prp:   CHECK( init(main()), ! uncaught(KeyError) )
                   CHECK( init(main()), ! uncaught(IndexError) )
arithmetic.prp:    CHECK( init(main()), ! uncaught(ZeroDivisionError) )
```

A verifier must answer, per family: *does an exception of exactly this type escape `main()` uncaught?* A task whose expected verdict is `false(KeyError)` must be distinguishable from one whose verdict is `false(IndexError)`.

### What we lack

ESBMC reports a single, **type-less** verdict when any exception escapes to the top level — `uncaught exception` / `__ESBMC_exc_thrown == 0` — regardless of which exception it was. Two different uncaught exceptions produce **byte-identical** output, so we cannot say which family was violated, and cannot map a result to a specific `.prp`.

### Example

`ierr.py` — should violate `! uncaught(IndexError)`:
```python
def main():
    a = [1, 2, 3]
    z = a[5]
main()
```

`kerr.py` — should violate `! uncaught(KeyError)`:
```python
def main():
    d = {"a": 1}
    y = d["b"]
main()
```

Both, with no flags:
```
$ esbmc ierr.py
Violated property:
  uncaught exception
  c:@__ESBMC_exc_thrown == 0
VERIFICATION FAILED

$ esbmc kerr.py
Violated property:
  uncaught exception
  c:@__ESBMC_exc_thrown == 0
VERIFICATION FAILED
```

The two verdicts are indistinguishable — nothing separates the IndexError from the KeyError.

### Current situation — the type information is present, just not consulted

`esbmc --goto-functions-only` shows each throw already sets a **distinct** typeid and message before propagating:

| source   | `__ESBMC_exc_typeid` | `__ESBMC_exc_obj.message`   |
|----------|----------------------|------------------------------|
| `a[5]`   | `11`                 | `"list index out of range"`  |
| `d["b"]` | `12`                 | (KeyError message)           |

Both globals are still live at the top-level assertion. But the uncaught check branches only on *whether* an exception is in flight, never on its type:

- `src/goto-programs/remove_exceptions.cpp:904-911` — the entry epilogue emits the uncaught check with condition `equality2tc(thrown, gen_false_expr())` (i.e. `assert(__ESBMC_exc_thrown == 0)`).
- `src/goto-programs/remove_exceptions.cpp:1143-1144` — the `uncaught` case hard-codes the comment string `"uncaught exception"`.

The machinery to make it type-aware already lives in the same file:
- `exception_typeidt registry` (`remove_exceptions.cpp:364`) — exception name ↔ typeid.
- `match_guard(catch_type)` (`remove_exceptions.cpp:1049-1052`) — already builds `__ESBMC_exc_typeid ∈ concrete_subtype_ids(catch_type)` for `except` dispatch.

**This family of cases is growing.** Making `ZeroDivisionError` a first-class, catchable Python exception (in-review work) routes it through this same throw/`__ESBMC_exc_thrown` path, so `1 // 0` now also produces the generic type-less `uncaught exception` instead of a distinct label. The fix below covers `KeyError`, `IndexError`, and `ZeroDivisionError` in one place.

### What we should do

Let the top-level uncaught check optionally target a specific family, so it maps to one `.prp`:

```
assert( !(__ESBMC_exc_thrown && match_guard(TargetException)) )   // FAIL only for the targeted family
```

and surface the concrete type in the diagnostic (e.g. `uncaught exception: IndexError`) instead of the bare string. The target family comes from the `.prp` via the SV-COMP wrapper (`scripts/competitions/svcomp/esbmc-wrapper.py`), which needs a Python `! uncaught(...)` branch anyway.

### Scope

Covers the exception-model families that flow through the throw/`__ESBMC_exc_thrown` path: `KeyError`, `IndexError`, `ZeroDivisionError`, and user `raise`d exceptions. `AssertionError` is unaffected (already reported as a native assertion with its own location). Dynamic-typing families (`TypeError`, `AttributeError`) are a separate gap — they are not yet raised as runtime exceptions at all (tracked separately).

lucasccordeiro:
Partition the program-entry uncaught-exception property by the escaping
exception's dynamic type, so the verdict names it (`uncaught exception:
IndexError` vs `uncaught exception: KeyError`) instead of collapsing every
family to a single generic `uncaught exception`. This backs the SV-COMP Python
track's per-exception-family checks.

`remove_exceptions` emits one `assert(!(thrown && typeid == id(T)))` per
throwable type plus a residual catch-all; their skip-conditions conjoin to
exactly `thrown == false`, so the split is a sound, complete refinement of the
original single property — no new false-SUCCESSFUL path. The typed label is
best-effort (reads `__ESBMC_exc_typeid`); the residual guarantees a verdict
regardless of `type_id` fidelity. The comment is a superset of the old string,
so existing substring-matching tests still pass; C++ throws are named too
(`uncaught exception: struct X`).

## Before / after

| Program | Before | After |
|---|---|---|
| `z = a[5]` | `uncaught exception` | `uncaught exception: IndexError` |
| `y = d["b"]` | `uncaught exception` | `uncaught exception: KeyError` |
| `raise ZeroDivisionError(...)` | `uncaught exception` | `uncaught exception: ZeroDivisionError` |
| `raise MyError()` | `uncaught exception` | `uncaught exception: MyError` |

## Tests

- New: `regression/python/github_5903_{indexerror,keyerror,zerodiv,userexc}_fail` (typed verdicts) and `github_5903_caught` (caught → no property fires).
- No regressions: exception-related Python suite, C++ `try_catch` suite, and all pre-existing `uncaught exception` tests pass.

Fixes #5903

### DIFF
--- a/regression/python/github_5903_caught/main.py
+++ b/regression/python/github_5903_caught/main.py
@@ -1,0 +1,11 @@
+# Regression for issue #5903: when the exception is caught, the type-partitioned
+# uncaught-exception property must not fire.
+def main():
+    try:
+        a = [1, 2, 3]
+        z = a[5]
+    except IndexError:
+        pass
+
+
+main()
--- a/regression/python/github_5903_caught/test.desc
+++ b/regression/python/github_5903_caught/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_5903_indexerror_fail/main.py
+++ b/regression/python/github_5903_indexerror_fail/main.py
@@ -1,0 +1,9 @@
+# Regression for issue #5903: an uncaught IndexError escaping main must produce
+# a type-named verdict ("uncaught exception: IndexError"), distinct from other
+# exception families.
+def main():
+    a = [1, 2, 3]
+    z = a[5]
+
+
+main()
--- a/regression/python/github_5903_indexerror_fail/test.desc
+++ b/regression/python/github_5903_indexerror_fail/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.py
+
+^VERIFICATION FAILED$
+uncaught exception: IndexError
--- a/regression/python/github_5903_keyerror_fail/main.py
+++ b/regression/python/github_5903_keyerror_fail/main.py
@@ -1,0 +1,8 @@
+# Regression for issue #5903: an uncaught KeyError escaping main must produce a
+# type-named verdict ("uncaught exception: KeyError"), distinct from IndexError.
+def main():
+    d = {"a": 1}
+    y = d["b"]
+
+
+main()
--- a/regression/python/github_5903_keyerror_fail/test.desc
+++ b/regression/python/github_5903_keyerror_fail/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.py
+
+^VERIFICATION FAILED$
+uncaught exception: KeyError
--- a/regression/python/github_5903_userexc_fail/main.py
+++ b/regression/python/github_5903_userexc_fail/main.py
@@ -1,0 +1,11 @@
+# Regression for issue #5903: a user-raised exception escaping main must be
+# named in the verdict ("uncaught exception: MyError").
+class MyError(Exception):
+    pass
+
+
+def main():
+    raise MyError()
+
+
+main()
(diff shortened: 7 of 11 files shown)


XLiZHI (approved):

LukeW1999 (approved):
LGTM

PR #6019, opened 2026-07-13
[docs] Add irep2 formal verification plan

lucasccordeiro:
## What

Adds `docs/irep2-verification-plan.md`, a comprehensive, code-grounded plan for
formally verifying the `src/irep2` subsystem (ESBMC's `expr2t`/`type2t` internal
representation) using ESBMC itself, supported by Catch2 property tests where the
template-heavy code is outside ESBMC's practical reach.

## Contents

- **Component analysis** — the container/refcount/COW machinery, generic
  dispatch, CRC traversal, `get_width`, utilities, and guard set-algebra;
  10 representation invariants (I1–I10) and a P0–P3 safety-criticality ranking.
- **Verification strategy** — a three-tier decomposition: standalone reduced
  ESBMC harnesses for algorithmic kernels (Tier A, H-A1…H-A10), Catch2
  property/differential tests on the real classes (Tier B, H-B1…H-B5), and
  TSan + hand-proof for the atomic/concurrency parts ESBMC cannot cover
  (Tier C). Each harness has preconditions, postconditions, an ESBMC
  invocation, and an anti-vacuity failing variant.
- **Property checking** — memory safety, over/underflow, division by zero,
  invalid deref, bounds, uninitialised data, assertions, and UB, each mapped to
  real ESBMC flags and concrete targets.
- **Data-structure integrity** — ownership, lifetime, aliasing, sharing, and
  copy/move/compare/hash/serialise/traverse obligations.
- **Coverage matrix, risk assessment, recommended verification order, and
  deliverables.**

## Notable findings surfaced (hypotheses to confirm via the harnesses)

The risk assessment flags three concrete missing invariants in the current
source, each with a `file:line` reference:

- **R1** — `array/vector/struct::get_width` multiply/accumulate widths into
  `unsigned int` with no overflow check (`irep2_type.cpp:146,159,188-191`).
- **R2** — `constant_int2t::as_ulong()` asserts only non-negativity; a
  `BigInt > UINT64_MAX` is silently truncated (`irep2_expr.cpp:101-106`; the
  code carries the matching `XXXjmorse` TODO).
- **R3** — `vector_type2t::get_width` lacks the symbolic-size guard the array
  path has, so a symbolic-sized vector null-derefs in release builds
  (`irep2_type.cpp:149-160`).

This PR adds only the plan document; the harnesses and any fixes are follow-ups
scoped by the milestones (M0–M6) in the plan.

### DIFF
--- a/docs/irep2-verification-plan.md
+++ b/docs/irep2-verification-plan.md
@@ -1,0 +1,770 @@
+# IREP2 Formal Verification Plan
+
+**Subsystem:** `src/irep2` — ESBMC's typed, intrusively reference-counted,
+copy-on-write internal representation for expressions (`expr2t`) and types
+(`type2t`).
+**Verifier:** ESBMC itself (bounded model checking + k-induction), supported by
+Catch2 property/differential unit tests for the parts that are outside ESBMC's
+practical reach.
+**Status:** Plan / not yet executed. Every harness below is a *proposal*; nothing
+here asserts a proof has been discharged.
+**Audience:** An engineer who will implement the harnesses and run the
+verification tasks directly from this document.
+
+> **Framing.** ESBMC is a verification tool, so soundness of *its own* IR is
+> load-bearing: a latent bug in `irep2` (a wrong `crc`, a non-total ordering, a
+> mis-computed width, a dropped guard conjunct) can silently corrupt every
+> downstream verdict ESBMC produces. This plan treats `irep2` as safety-critical
+> and adopts a conservative stance: **assume subtle bugs exist until formally
+> disproven.**
+
+---
+
+## 1. Verification objectives
+
+1. **Memory safety of the container machinery.** The hand-rolled intrusive
+   refcount (`irep_container`, `irep2t::refcount`) and copy-on-write `detach()`
+   must never double-free, use-after-free, leak, or free a still-referenced
+   node — including under the self-aliasing assignment the code explicitly
+   guards against (`irep2.h:190-225`).
+2. **Algorithmic correctness of the value-semantics operations.** `crc`
+   (hashing), `cmp`/`operator==` (structural equality), `lt`/`operator<`
+   (ordering), `clone`, `with_type`, and operand/subtype traversal must be
+   mutually consistent and satisfy their mathematical contracts (equality is an
+   equivalence relation; `<` is a strict weak order; `a == b ⇒ crc(a) == crc(b)`).
+3. **Absence of ESBMC-checkable undefined behaviour** in every arithmetic,
+   indexing, and bit-manipulation kernel: no overflow producing a wrong result,
+   no division by zero, no out-of-bounds read/write, no use of uninitialised
+   data, no invalid pointer dereference.
+4. **Representation-invariant preservation.** Every constructor establishes, and
+   every mutator preserves, the per-kind invariants (`fields` covers storage,
+   array-size normalisation, guard canonical-chain ↔ `guard_list`
+   correspondence, `crc_val == 0 ⇔ not-yet-computed`).
+5. **Soundness of the guard set-algebra.** `operator-=` / `operator|=` on
+   `guard2tc` must compute path conditions that are *logically* correct: the
+   optimised O(Δ) prefix walk must agree with the naive semantics in every case,
+   because a wrong result is an unsound path condition (a whole class of missed
+   or spurious counterexamples).
+6. **Regression durability.** Each discharged property is pinned by a
+   check-in-able test (Catch2 case and/or an ESBMC `regression/esbmc/…` harness)
+   that would fail if the property were re-broken.
+
+---
+
+## 2. Scope and the reach of ESBMC on this codebase
+
+`src/irep2` is ~10.2k LOC of modern C++ (C++20/23: `std::apply`,
+`std::unreachable`, fold expressions, `if constexpr`, concepts-style traits),
+linked against `immer`, `bigint`, `fmt`, Boost, and `std::atomic`. ESBMC's C++
+frontend cannot practically ingest the whole subsystem end-to-end (template
+metaprogramming over `K::fields`, `immer`'s persistent trie, `fmt` internals).
+Pretending otherwise would produce vacuous "proofs." The plan therefore splits
+targets into three honest tiers.
+
+| Tier | What it covers | Verification technique | Why |
+|------|----------------|------------------------|-----|
+| **A — ESBMC-verifiable kernels** | Self-contained algorithmic cores: byte/CRC loops, width arithmetic, endianness reconstruction, index math, refcount lifecycle, guard prefix walk, `as_ulong`/`as_long` bounds | **Standalone reduced C/C++ harnesses** verified by ESBMC (`--overflow-check --unsigned-overflow-check --memory-leak-check --ub-shift-check …`) with nondeterministic inputs. Kernel logic is transcribed faithfully (or `#include`d where it compiles standalone) and driven by `nondet_*()`. | These are ordinary imperative code once lifted out of the template layer; ESBMC excels here and gives a deterministic witness on failure. |
+| **B — Property / differential harnesses (in-tree)** | Algebraic laws that need the *real* classes (equality/ordering/CRC consistency, COW value semantics, guard-algebra logical equivalence) | **Catch2 tests in `unit/irep2/`** driving the actual `expr2tc`/`type2tc`/`guard2tc` API, bounded-exhaustive or randomised; the *logical-equivalence* checks additionally shell out to ESBMC/an SMT check on small terms. | The template dispatch cannot be reduced without losing fidelity, but the observable contracts can be exercised against the genuine implementation. |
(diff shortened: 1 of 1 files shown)


docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+assert (i) the other handle's *value* is unchanged, (ii) pointers diverge, (iii)
+no double-free (run the case under ASan in CI). Add a move-assignment aliasing
+case mirroring H-A3 on the genuine `expr2tc`.
+
+#### H-B4 — Guard algebra logical equivalence (P0-soundness)
+
+For small nondet conjunct sets, build `g1`, `g2`; compute `g1 -= g2` and `g1 |=
+g2`; then check with an SMT query that `as_expr()` of the result is logically
+equal to the reference semantics (`g1 ∧ ¬(shared) ` for `-=` in the
+symex-intended sense; `g1 ∨ g2` for `|=`). Because the operators are *heavily*
```
lucasccordeiro: **H-B4 oracle is mis-specified for `-=`.** `operator-=` (`irep2_guard.cpp:334`) computes conjunct set difference: the result is `⋀(conjuncts(g1) \ conjuncts(g2))`, which is *not* logically equivalent to `g1 ∧ ¬(shared)`. As written, the M4 differential check verifies the wrong law. Specify the oracle as the naive O(N) set-difference reference. The `|=` half (`g1 ∨ g2`) is correct.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+int deleted = 0;            // free counter
+_Bool slot_live[N];         // which slots hold the node
+
+void adopt(int s){ if(!slot_live[s]){ slot_live[s]=1; refcount++; } }      // copy-in
+void drop(int s){                                                         // release()
+  if(!slot_live[s]) return;
+  slot_live[s]=0;
+  unsigned prev = refcount--;            // fetch_sub
+  assert(prev >= 1);                     // no underflow  (I1)
+  if(prev==1){ deleted++; assert(deleted==1); }   // exactly one free
```
lucasccordeiro, line 322: **H-A1 sketch reports spurious counterexamples on correct code.** After `deleted==1` (refcount 0), `adopt(s)` can still fire, resurrecting the freed node; the next drop hits `prev==1` again and both `assert(deleted==1)` and the `deleted>0 ⇒ refcount==0` invariant fail. Real containers can't copy from a freed node — gate adoption on a live source (`__ESBMC_assume(!deleted)` or copy-from-live-slot).

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+    unsigned sh = 8*(le? j : w-1-j); assert(sh < 32);
+    c |= (uint32_t)s[idx] << sh;
+  }
+  return c;
+}
+```
+**Invocation:** `--overflow-check --ub-shift-check` (array-bounds is on by
+default). Add a differential check: big-endian reconstruction of 2 bytes equals
+`(s[0]<<8)|s[1]`.
+**Failing variant:** drop the `w<=4` assumption → shift ≥ 32 UB flagged.
```
lucasccordeiro: Dropping the `w<=4` assumption also admits `w==0`, so `slen % w` / `slen/w` in the harness's own preconditions divide by zero — the failing variant fails for the wrong reason. Perturb to e.g. `w ∈ {1..8}` instead.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+---
+
+## 3. Architecture overview
+
+```
+                       irep2t  (base: atomic refcount, atomic crc_val, dbg writer stamp)
+                      /      \
+                 type2t      expr2t   (type_id/expr_id enum + switch dispatch)
+                    |            |
+        <kind>_type2t...   <kind>2t...   (124 expr kinds, 15 type kinds; each: fields tuple + field_names)
```
lucasccordeiro: Expr kind count is **122**, not 124 (`grep -c '^IREP2_EXPR(' src/irep2/expr_kinds.inc`). Repeated in §4.1 C3 and H-B5. Since H-B5 uses this as a completeness criterion, fix the number — or have the sweep derive the count from `expr_kinds.inc` so it never goes stale.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+- **Dual-solver agreement.** Soundness-critical runs (guard algebra, width fixes)
+  require Bitwuzla **and** Z3 to agree, matching the project's Mode-C requirement.
+
+## Appendix B — Standard `test.desc` template for Tier-A harnesses
+
+```
+CORE
+irep2_<kernel>_<ok|fail>.c
+--overflow-check --unsigned-overflow-check --memory-leak-check --k-induction
+^VERIFICATION SUCCESSFUL$          # (or ^VERIFICATION FAILED$ for the fail variant)
```
lucasccordeiro: `test.desc` has no comment syntax — line 4+ is consumed verbatim as a regex by `testing_tool.py`, so a copied template yields a never-matching regex. Move the "(or `^VERIFICATION FAILED$` …)" remark out of the code block. Also, line 3 bakes `--k-induction` into the universal template while H-A5/A6/A7 are plain BMC — show flags as per-harness.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+
+**Harness locations.**
+- Tier A standalone harnesses: `regression/esbmc/irep2_<kernel>_<n>/` (one
+  passing + one failing variant per §CLAUDE.md "two regression tests per PR"),
+  each with a `test.desc`.
+- Tier B property tests: `unit/irep2/*.test.cpp` (new files, wired via
+  `unit/irep2/CMakeLists.txt` `new_unit_test(...)`).
+- Tier C: notes in this document + a `MALLOC_PERTURB_`/TSan CI recipe.
+
+**ESBMC binary for Tier A:** `build/src/esbmc/esbmc` (v8.4.0 present in this
```
lucasccordeiro: "v8.4.0 present in this tree" references the author's local build state and rots immediately. Say "a current master build" and record the commit hash in the verdict log when harnesses run.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+---
+
+## 5. Target properties (ESBMC-supported)
+
+Every Tier-A harness is run with the full property set enabled. Mapping of the
+task's required checks to ESBMC flags:
+
+| Property | ESBMC flag(s) | Primary targets in `irep2` |
+|----------|---------------|----------------------------|
+| Memory safety (bounds, invalid deref, UAF, double-free, leak) | `--memory-leak-check --no-slice` (pointer/bounds/deref checks are **on by default**) | C1 refcount/`release`/`detach`; `do_get_sub_expr<vector>` index math (`irep2_utils.cpp:338-355`); `constant_string_access::operator[]` (`irep2_expr.cpp:255-265`) |
```
lucasccordeiro: `--no-slice` is not a checker flag — either justify why slicing must be disabled for these harnesses or drop it, or it gets cargo-culted into every `test.desc`.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+  unsigned int w=0;
+  for(int i=0;i<n;i++){ assert((unsigned long)w + ws[i] <= UINT_MAX); w+=ws[i]; }
+  return w;
+}
+```
+Drive with `num_elems`, `sub_width`, member widths nondet within realistic
+caps and *also* at adversarial extremes.
+**Invocation:** `--overflow-check --unsigned-overflow-check`.
+**Expected outcome:** **FAILS on the current code** (no guard exists) — this is a
+finding (R1), not a passing proof. The passing harness is the *post-fix* version
```
lucasccordeiro, line 443: Sequencing: H-A5/H-A6 are expected to FAIL on current code, but §12/Appendix B deliver them as `CORE` asserting `^VERIFICATION SUCCESSFUL$`. State explicitly that the R1/R2 harnesses land in the same PR as their fixes (or as `KNOWNBUG` first), or CI is red in between.

lucasccordeiro (commented):
Inline comments below on the items that should change before merge. Minor extra: several `irep2.h` line refs are already drifting by ±1 — keep the symbol-name-plus-line discipline so the anchors stay resolvable.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+#### H-A5 — Width arithmetic overflow (P0-soundness) — **exposes R1**
+
+**Target:** `array_type2t::get_width` `num_elems * sub_width`
+(`irep2_type.cpp:146`), `vector_type2t::get_width` (`:159`), `struct` sum
+(`:188-191`), `union` max (`:199-201`).
+
+```c
+unsigned int width_array(unsigned long num_elems, unsigned int sub_width){
+  // FAITHFUL: return (unsigned int)(num_elems * sub_width);
+  unsigned long prod = (unsigned long)num_elems * sub_width;
```
lucasccordeiro: `unsigned long` is only wider than `unsigned int` on LP64 — use `uint64_t` for the checked product so the harness isn't ILP32-fragile. Note the production code shares this: on ILP32, R1 is an overflow rather than a truncation; the harness should cover both.

docs/irep2-verification-plan.md
```diff
@@ -0,0 +1,770 @@
...
+   or spurious counterexamples).
+6. **Regression durability.** Each discharged property is pinned by a
+   check-in-able test (Catch2 case and/or an ESBMC `regression/esbmc/…` harness)
+   that would fail if the property were re-broken.
+
+---
+
+## 2. Scope and the reach of ESBMC on this codebase
+
+`src/irep2` is ~10.2k LOC of modern C++ (C++20/23: `std::apply`,
```
lucasccordeiro: `src/irep2` is 9,684 lines *including* the 832-line README and CMakeLists; code is ~8.8k. Say "~9k" or drop the figure.

PR #6024, opened 2026-07-13
[irep2] Verify real irep_container refcount lifecycle (H-A1, Tier B)

lucasccordeiro:
## What

First execution step of the [irep2 formal verification plan](https://github.com/esbmc/esbmc/blob/master/docs/irep2-verification-plan.md) (#6019): **H-A1 — refcount conservation, single-free & self-alias UAF-safety**.

This verifies the **actual `irep2` C++ implementation**, not a model. Per the plan's **Tier B** strategy (§2), the templated / `immer` / `fmt` / `std::atomic` layering of `irep2` cannot be ingested by ESBMC end-to-end, so the ownership contract is exercised directly against the genuine `irep_container<T>` / `expr2t` classes — reading the real `irep2t::refcount` atomic after each operation (the same way `unit/irep2/irep2.bench.cpp` drives real nodes).

> Note: an earlier revision of this PR used a hand-written C *model* verified by ESBMC. Following review feedback, it was pivoted to drive the real classes instead.

## The harness — `unit/irep2/refcount.test.cpp`

Catch2 property tests over the real container, asserting:

| Case | Verifies |
|------|----------|
| refcount tracks live handles (I1) | copy-ctor `fetch_add`, move-ctor steal (no bump), scoped destruction decrement — exact counts on the real atomic |
| copy-assign releases old + bumps new | the previous pointee is freed exactly once; surviving value intact |
| self-aliasing assignment is UAF-safe (H-A3) | `x = <member of x's own node>` — the documented `irep2.h:190-225` snapshot-before-release hazard |
| COW `detach` conserves & isolates | non-const `get()` clones iff shared; both handles end sole-owners; structural value preserved |

Measurement uses the **const `get()`** overload so reading `refcount` never triggers `detach()` and corrupts the count.

Result: `./build/unit/irep2/irep2refcounttest` → **All tests passed (23 assertions in 4 test cases)**. Built under the `Sanitizer` build type (`./scripts/build.sh -b Sanitizer -s ASAN`) the same cases additionally witness double-free / use-after-free freedom.

## Notes

- No `src/**` changes — a new unit test + CMake wiring + the plan's execution-log entry. No Mode C dead-code obligation.
- Code-reviewed; the ASan-coverage wording was corrected to reflect that unit tests are witnessed under a local `Sanitizer` build (the `sanitizers.yml` CI job runs the regression suite, not unit tests).

Refs #6019


unit/irep2/CMakeLists.txt
```diff
@@ -1,4 +1,11 @@
 new_unit_test(irep2test "irep2.test.cpp" "util_esbmc")
+new_unit_test(irep2refcounttest "refcount.test.cpp" "util_esbmc")
+
+# libFuzzer harness driving the real irep_container refcount lifecycle with
```
rafaelsamenezes: These comments are needless.

unit/irep2/refcount.fuzz.cpp
```diff
@@ -0,0 +1,27 @@
+// libFuzzer entry point: drive the REAL irep_container<T> refcount lifecycle
+// with nondeterministic input. libFuzzer supplies the byte stream; run_ops
+// decodes it into a sequence of container operations over real expr2tc slots
+// and checks refcount conservation (I1) after every step, while ASan
```
rafaelsamenezes: Again, we all know what ASAN does.

unit/irep2/refcount.fuzz.cpp
```diff
@@ -0,0 +1,27 @@
...
+// (-fsanitize=fuzzer,address) catches any use-after-free / double-free.
+//
+// Built only when -DENABLE_FUZZER=On (see unit/irep2/CMakeLists.txt). Run with:
+//   ./unit/irep2/irep2refcountfuzz -runs=1000000
```
rafaelsamenezes: We can see this by looking at the CMake.

unit/irep2/refcount.fuzz.cpp
```diff
@@ -0,0 +1,27 @@
...
+// The deterministic Catch2 replay in refcount.test.cpp exercises the same
```
rafaelsamenezes: We don't care

unit/irep2/refcount.fuzz.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <cstdlib>
+
+#include "refcount_ops.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
+{
+  // Explicit abort (not assert): fuzz builds usually define NDEBUG, which
+  // would compile out an assert and silently disable the conservation
+  // oracle. ASan (-fsanitize=fuzzer,address) independently flags any
+  // use-after-free / double-free in the real container operations.
```
rafaelsamenezes: Needless comment

unit/irep2/refcount.test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  REQUIRE(node_of(a) != node_of(b)); // distinct nodes now
+  REQUIRE(live_refcount(a) == 1);    // a sole-owns its clone
+  REQUIRE(live_refcount(b) == 1);    // b sole-owns the original
+  REQUIRE(*a == *b);                 // structural value unchanged
+}
+
+// Nondeterministic-input replay: the same byte-stream operation driver that
+// libFuzzer exercises (refcount_ops.h / refcount.fuzz.cpp) is run here on
+// deterministic inputs so the property is pinned in normal CI even with
+// fuzzing off. Each input decodes to a sequence of make/copy/move/detach/
```
rafaelsamenezes: We don't care about fuzzing here.

rafaelsamenezes (changes requested):
Just asking for better comments in general

rafaelsamenezes (approved):

PR #6034, opened 2026-07-13
[k-induction] Make --termination take priority over k-induction

### LINKED ISSUE #6031: Non-deterministic results and false negative for --termination --k-induction-parallel
Since upgrading to v8.4, I have encountered the following non-deterministic and sometimes unsound behavior when combining the options --k-induction-parallel and  --termination:

For the attached [minimized.c](https://github.com/user-attachments/files/29970001/minimized.c) code (which I tried to simplify as much as possible), the call 

`esbmc --termination --k-induction-parallel minimized.c`

sometimes results in 'VERIFICATION FAILED' as expected (see [minimized_expected.log](https://github.com/user-attachments/files/29970051/minimized_expected.log)), but occasionally also unexpectedly results in 'VERIFICATION SUCCESSFUL' (see [minimized_faulty.log](https://github.com/user-attachments/files/29970059/minimized_faulty.log)). I was not able to reproduce this behavior on v8.3, and if I remove either one of the two options I consistently get 'VERIFICATION FAILED' again.

Exact version: 'ESBMC version 8.4.0 64-bit x86_64 linux'

lucasccordeiro:
The parallel k-induction driver interprets its results as assertion reachability and has no termination handling, so combining it with --termination produced non-deterministic and sometimes unsound VERIFICATION SUCCESSFUL for a failing program (#6031).

Make --termination take priority over both k-induction variants (inverting the previous k-induction-wins rule) so the run routes to the sequential termination strategy and the verdict is deterministic. Adds two regression tests.

Fixes #6031

### DIFF
--- a/regression/termination/github_6031-terminating/main.c
+++ b/regression/termination/github_6031-terminating/main.c
@@ -1,0 +1,15 @@
+extern int __VERIFIER_nondet_int(void);
+
+// Companion to github_6031: a terminating program under the same flag
+// combination (--termination --k-induction-parallel). The loop always
+// terminates, so the sequential termination strategy (which --termination
+// now routes to instead of the parallel driver) must report VERIFICATION
+// SUCCESSFUL. This pins that the priority routing is not a blanket "always
+// FAILED".
+int main()
+{
+  int n = __VERIFIER_nondet_int();
+  while (n < 10)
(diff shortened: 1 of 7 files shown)


mikhailramalho (approved):
lgtm

PR #6056, opened 2026-07-14
[python] Group loose python-frontend files into per-concern subdirs

lucasccordeiro:
Relocate the loose top-level files under `src/python-frontend/` into cohesive
per-concern subdirectories — `math/`, `module/`, `exception/`, `type/`, `class/`,
`set/`, `tuple/`, `consteval/`, `numpy/` — and collocate the orphan headers
`python_annotation.h`, `python_dict_handler.h` and `python_list.h` into their
existing sibling dirs. This extends the layout convention already used by
`converter/`, `string/`, `function_call/`, `python-list/` and `python-dict/`.

Pure relocation, split into 8 small per-concern commits: no logic, API or
behaviour change. All three include forms (`<python-frontend/X.h>`, bare
`<X.h>`, quoted `"X.h"`) and the `CMakeLists.txt` source list are updated in
lockstep across `src/python-frontend/` and the `unit/python-frontend/`
consumers. Frontend entry points and cross-cutting orchestration
(`python_language`, `python_converter`, `python_adjust`, `python_expr_builder`,
`python_lambda`, `param_annotations`, `symbol_id`, `json_utils.h`) stay at root.

Verified: each group builds and links cleanly; a stale-include scan across all
moved headers is clean; the diff is renames + include-path edits + CMake path
edits with zero logic hunks; all nine `unit/python-frontend/` test binaries pass;
and a 41-test regression subset spanning every moved module (class, tuple, dict,
complex, cmath, numpy, exception, list, math) passes. `.clang-format` has
`SortIncludes: false`, so path-only include edits do not affect formatting.


lucasccordeiro (commented):
Reviewed by verifying the relocation claims mechanically on a local checkout of the branch, rather than by eye:

- **Zero logic hunks**: filtering every changed line leaves only `#include` edits and CMake source-path edits (the one oddity is a pre-existing UTF-8 BOM on the first line of `python_typechecking.cpp`, carried through unchanged).
- **No stale references**: grepped the whole tree (code, CMake, scripts, workflows, docs) for all 46 old paths in all three include forms — the only hits are unrelated `regression/cbmc` fixtures, a dated triage report in `docs/`, and one comment in `models/int.py` (below).
- **Every include resolves**: each `#include <python-frontend/...>` maps to an existing file, and every source listed in `src/python-frontend/CMakeLists.txt` exists (`pythonastgen.c` is build-time generated, pre-existing).
- **Per-commit bisect safety**: the include-resolution / stale-include / CMake-existence checks pass on each of the 8 commits individually, so moves and consumers land in lockstep and intermediate commits stay buildable.
- **No base drift**: none of the master commits since the branch point touch `src/python-frontend/` or `unit/python-frontend/`. Quoted `"X.h"` includes are sibling-relative in untouched dirs. No `.py` model moves, so no FLAIL/OM rebuild concern.

**Findings:**

1. **Stale soundness cross-reference in `src/python-frontend/models/int.py` (~line 62)** — worth fixing in this PR since it is the only in-tree reference the relocation actually invalidates. The comment `"the literal 512 must be >= kPythonBitLengthCap in src/python-frontend/type_handler.cpp, which a static_assert ties to kPythonBignumWidth"` should now point at `src/python-frontend/type/type_handler.cpp`. It is the maintainer breadcrumb coupling the OM literal to the C++ `static_assert` guarding bignum-width soundness, so keeping the path accurate matters.

2. *(non-blocking, pre-existing)* The tree now has three coexisting directory-naming styles: bare nouns (`class/`, `type/`, `set/`, `tuple/`, `math/`, ...), hyphen-prefixed (`python-list/`, `python-dict/`), and underscored (`python_annotation/`). Extending the existing convention here is reasonable; a tracking issue for eventual normalization may be worthwhile, but not in this PR.

3. *(non-blocking, pre-existing)* `class/python_class_builder.cpp:1-3` still uses bare `<python_converter.h>` / `<symbol_id.h>` / `<json_utils.h>` while the lines below use full `<python-frontend/...>` paths. They resolve via the frontend-root include dir, so this is fine — normalizing the remaining bare forms in moved files would be a cheap follow-up.

Overall: pure relocation as advertised — approve once CI is green, ideally with the one-line `models/int.py` path fix included.

PR #6102, opened 2026-07-16
[k-induction] Reject --base-k-step 0 (unwind 0 = unbounded unwinding)

LukeW1999:
Fixes #6093

## Problem

`esbmc --loop-invariant --base-k-step 0 --max-k-step 2 while-inc1-3.cpp`
unwinds the loop unboundedly, ignoring `--max-k-step 2`, instead of running
a bounded k-induction sweep.

## Root Cause

`--base-k-step 0` seeds the base-case loop at `k = 0`, and every k-induction
base-case driver sets the unwind bound from the current k:
`set_option("unwind", integer2string(k_step))`. ESBMC treats `--unwind 0` as
**unlimited** — loop termination is gated on `max_unwind != 0 && unwind >=
max_unwind` (`symex_goto.cpp`, `symex_function.cpp`), so `max_unwind == 0`
never stops a loop. The base case at `k = 0` therefore unwinds forever. The
existing validation only checks `base-k-step < max-k-step`, so `0` slips
through.

## Approach

Reject `--base-k-step 0` up front in `get_command_line_options` (which runs
for every invocation, before any strategy) with a clear error, matching the
existing `--k-step`/`--max-k-step` validation. k-induction's base case must
start at `k >= 1`; `k = 0` cannot be represented as an unwind bound because
`0` is the "unlimited" sentinel. The option help now documents `n >= 1`.

## Testing

- `regression/k-induction/github_6093` (Mark's exact reproducer): now aborts
  with `ERROR: Please specify --base-k-step >= 1 ...` instead of unwinding
  unboundedly.
- `regression/k-induction/github_6093_pass`: a valid `--base-k-step 2` run
  still verifies SUCCESSFUL, confirming the guard only rejects `k = 0`.
- Full `k-induction` suite (194 tests, incl. parallel) passes; defaults and
  valid `--base-k-step >= 1` are unaffected.

### DIFF
--- a/regression/k-induction/github_6093/main.c
+++ b/regression/k-induction/github_6093/main.c
@@ -1,0 +1,15 @@
+int main()
+{
+  int idx, max, res;
+
+  __ESBMC_assume(0 <= max);
+  res = 0;
+  idx = 0;
+
+  while (idx < max)
+  {
+    ++idx;
+    ++res;
+  }
+  __ESBMC_assert(idx == max, "A.1");
+}
--- a/regression/k-induction/github_6093/test.desc
+++ b/regression/k-induction/github_6093/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--loop-invariant --base-k-step 0 --max-k-step 2
+^ERROR: Please specify --base-k-step >= 1
--- a/regression/k-induction/github_6093_pass/main.c
+++ b/regression/k-induction/github_6093_pass/main.c
@@ -1,0 +1,7 @@
+int main()
+{
+  int i, s = 0;
+  for (i = 0; i < 3; i++)
+    s++;
+  __ESBMC_assert(s == 3, "s_eq_3");
+}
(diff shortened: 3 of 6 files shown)


LukeW1999:
Heads-up on overlap with #6103 (@lucasccordeiro): both PRs touch the same `--base-k-step`/`--max-k-step` validation in `command_line_options.cpp`, `k_induction.cpp`, and `bmc_strategy.cpp`, so whichever lands second will need a quick rebase.

They're complementary, not duplicates:
- This PR (#6102) rejects `--base-k-step 0`, which sets `--unwind 0` (ESBMC's "unlimited" sentinel) and causes unbounded unwinding (discussion #6093).
- #6103 reworks the wording of the existing `base-k-step >= max-k-step` / `k-step >= max-k-step` messages (discussion #6094).

Since `0 >= max-k-step` is false, #6103 does not catch the `base-k-step 0` case, so this guard is still needed on top of it. No behavioral conflict — just adjacent edits to the same blocks. Happy to rebase this onto #6103 (or vice-versa) once merge order is decided.

PR #6104, opened 2026-07-16
[goto-programs] Fix function calls inside __ESBMC_forall/__ESBMC_exists bodies

LukeW1999:
Fixes #6100

## Problem

A function call inside an `__ESBMC_forall`/`__ESBMC_exists` body produced a
wrong result. In the reporter's program `__ESBMC_exists(&var, eq(var, 6))`
(true at `var = 6`) wrongly reported `VERIFICATION FAILED`.

## Root Cause

`__ESBMC_forall`/`__ESBMC_exists` are lowered from function calls, and their
`_Bool` body argument is evaluated by the frontend. A function-call body is
hoisted into a temp *before* the quantifier:

```
FUNCTION_CALL: return_value$_eq = eq(var, 6)      // evaluated once, outer var
ASSIGN        exists = exists(&var, return_value$_eq)
```

So the quantifier binds `var`, but its body is a precomputed boolean that
depends on the *outer* `var`, not the bound one: `∃var. (outer_var == 6)` is
false for a nondet `outer_var`. `remove_sideeffects_for_quantifier_body` was
only reached for `||`/`&&` bodies; a bare call fell through to the generic
hoisting path. Tautological bodies (the `forall`/`exists-1` in the reproducer)
happened to pass regardless.

## Approach

- Route any side-effecting quantifier body (not just `||`/`&&`) through
  `remove_sideeffects_for_quantifier_body`.
- Add `try_inline_predicate_call`: when the body leaf calls a pure predicate
  (a function whose body is a single side-effect-free `return <expr>;`),
  inline it by substituting the actual arguments for the parameters via
  `replace_symbolt`, so the bound variable flows into the predicate.
- Soundness fallbacks — inlining is skipped (with a warning, keeping the
  previous behavior) when: the callee is not a single-`return` pure predicate,
  an actual argument has a side effect (e.g. `eq(g++, 6)`), or any parameter
  symbol survives substitution. Reading globals in the predicate is inlined
  (semantically identical to the call).

## Testing

- `regression/z3/github_6100`: the reproducer — `forall`/`exists` over `eq()`,
  including `exists(&var, eq(var, 6))` — now verifies SUCCESSFUL.
- `regression/z3/github_6100_fail`: `exists(&var, eq(var, var + 1))` is
  genuinely false and must FAIL, guarding against the hoisting regression.
- Verified soundness by hand: `exists(var==var+1)` and `forall(var==6)` still
  FAIL; nested `forall i. exists j. le(i,j)` verifies; a side-effecting arg and
  a non-pure callee both warn and fall back.
- Full `z3`, `bitwuzla`, and `cbmc` suites plus all quantifier tests pass
  (no regressions).

### DIFF
--- a/regression/z3/github_6100/main.c
+++ b/regression/z3/github_6100/main.c
@@ -1,0 +1,12 @@
+_Bool eq(int a, int b)
+{
+  return a == b;
+}
+
+int main()
+{
+  int var;
+  __ESBMC_assert(__ESBMC_forall(&var, eq(var + 1, 1 + var)), "forall");
+  __ESBMC_assert(__ESBMC_exists(&var, eq(var + 1, 1 + var)), "exists-1");
+  __ESBMC_assert(__ESBMC_exists(&var, eq(var, 6)), "exists-2");
+}
--- a/regression/z3/github_6100/test.desc
+++ b/regression/z3/github_6100/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--z3
+^VERIFICATION SUCCESSFUL$
--- a/regression/z3/github_6100_fail/main.c
+++ b/regression/z3/github_6100_fail/main.c
@@ -1,0 +1,13 @@
+_Bool eq(int a, int b)
+{
+  return a == b;
+}
+
+int main()
+{
+  int var;
+  // No value satisfies var == var + 1, so the existential is false and the
+  // assertion must fail. Before the fix, eq() was hoisted out of the
+  // quantifier and evaluated once, masking such cases (discussion #6100).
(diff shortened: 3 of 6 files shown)


LukeW1999:
Superseded by #6105 (@lucasccordeiro), which fixes the same discussion #6100 with a more complete approach: recursive inlining, nested calls, single-use handling for side-effecting arguments, an address-of guard, and a hard diagnostic for non-inlinable bound-variable calls instead of silently hoisting. I verified #6105 is sound (inlining confirmed via --goto-functions-only; valid props SUCCESSFUL, forall(var==6) FAILED, impure callee gives a clean diagnostic). Closing in favour of #6105.

PR #6105, opened 2026-07-16
[goto-programs] Inline function calls in quantifier bodies

lucasccordeiro:
Side-effect removal hoisted function calls out of `__ESBMC_forall`/
`__ESBMC_exists` bodies, freezing the bound variable at its pre-quantifier
value and making the quantifier vacuous (wrong verdicts both ways: the
`exists-2` property in discussion #6100 failed though satisfiable, and
`forall(&var, eq(var, 6))` passed though invalid).

Inline pure single-return callees under the binder instead (recursively,
with typecast-adjusted argument substitution), and reject any remaining
side effect on a bound variable with a clear diagnostic rather than a
silent vacuous quantifier. Side-effecting arguments are only substituted
when the parameter is used exactly once, and parameters under address-of
are never substituted, so calls are neither dropped nor duplicated.

Regression tests: `regression/z3/github_6100` (pass, incl. nested and
duplicated-parameter inlining), `github_6100_fail` (forall must fail),
`github_6100_noninline` (clean diagnostic for a non-inlinable callee).
Verified under both Bitwuzla and Z3.

Fixes the problem reported in discussion #6100.


LukeW1999 (approved):
Reviewed and verified — sound, and a clear superset of my #6104 (which I had opened for the same #6100 before seeing this). Closing #6104 in favor of this one.

I checked the lowering with `--goto-functions-only`: the quantifier bodies are inlined with the bound variable flowing in (0 hoisted `eq`/`inc`/`dbl` FUNCTION_CALLs in `main`), so the binder is no longer detached from the body. Behavioural checks:

- `github_6100` (incl. nested `eq(inc(var), 7)` and duplicated-parameter `dbl(var) == 12`) → VERIFICATION SUCCESSFUL
- `github_6100_fail` (`forall(&var, eq(var, 6))`) → VERIFICATION FAILED
- `github_6100_noninline` (`impure` with a loop) → clean `cannot model call ... on a quantified variable` diagnostic
- extra soundness spot-check: `exists(&var, eq(var, var + 1))` → FAILED

The recursive inlining, the single-use rule for side-effecting arguments, and the address-of guard all look correct, and rejecting a residual side effect *on a bound variable* (rather than silently hoisting it, which detaches it from the binder) is the right call.

Non-blocking nit: the hard `abort()` on a non-inlinable bound-variable call ends the whole run; emitting the error and letting the driver report ERROR/UNKNOWN would be a touch gentler for multi-property runs, but matching the existing "can't model" convention is fine.

PR #6138, opened 2026-07-17
[solver] Constrain object base addresses to their natural alignment

rafaelsamenezes:
## What

`init_pointer_obj` only constrained an object's base address when its type carried an explicit `"alignment"` irep attribute — which only the clang frontend sets, for aggregates (via `padding.cpp`) or an explicit `alignas`. Every other object had a completely free base address, so `(uintptr_t)&x % alignof(T) == 0` was satisfiably false and produced a spurious counterexample.

This affects two groups:

- **Any scalar, on the native path.** `int x; assert((uintptr_t)&x % 4 == 0);` reported `VERIFICATION FAILED` on master. Scalars never get the attribute.
- **Every object on the CBMC `--binary` path**, since `padding.cpp` never runs there. This is the gap `docs/cprover-support-roadmap.md` calls *"the single highest-leverage gap in the corpus"* for the `verify-rust-std` / Kani front.

The fix constrains the base address to the type's alignment, which reduces to the explicit attribute when present and to natural alignment otherwise. Packed types keep alignment 1 and stay unconstrained, exactly as before.

This required moving `alignment()` out of `clang-c-frontend/padding.cpp` into `util/`, beside `type_byte_size()` — it is a general type-layout query, and the solver must not depend on a frontend. No behaviour change to `alignment()` itself; its one external caller (`clang_c_adjust_expr.cpp`) already includes the new header.

## Validation

**The three corpus-wide verdict mismatches are resolved, and now match CBMC per-property, not just at the verdict.** Against CBMC 6.10.0 on the identical Kani binary, with `esbmc --binary --multi-property`:

| Harness | Before | ESBMC now | CBMC 6.10.0 |
|---|---|---|---|
| `non_null_check_cast` | FAILED on `misaligned pointer to reference cast` | 1 failed — `core_0` marker | 1 of 85 failed — same marker |
| `non_null_check_as_ref` | same false alarm | 4 failed — `core_0..3` markers | 4 of 123 failed — same markers |
| `non_null_check_as_mut` | same false alarm | 4 failed — `core_0..3` markers | 4 of 123 failed — same markers |

In each case the substantive alignment assertion (`|result: &NonNull<U>| result.as_ptr().addr() == self.as_ptr().addr()`, `non_null.rs:531`/`464`) now **PASSES** in ESBMC as it does in CBMC, and the only remaining failures are Kani `reachability_check` coverage markers — identical to CBMC's set.

Targeted behaviour checks:

- scalar / array / pointer / plain-struct natural alignment → now `SUCCESSFUL` (was `FAILED`)
- explicit `__attribute__((aligned(16)))` struct, global and local → still `SUCCESSFUL`
- packed struct and `char[16]` → correctly still `FAILED` (alignment 1 must not be over-constrained)

Unit tests: **467/467 pass**. Two regression tests added (`regression/esbmc/natural_alignment{,_fail}`), both verified.

## Not yet validated

- Full regression suite has **not** been run on this branch. @rafaelsamenezes is running the SV-COMP suite.
- The constraint now applies to every object with alignment > 1, so it is worth watching for solver-time regressions on pointer-heavy benchmarks.

## Two corrections to the roadmap (for a follow-up, not fixed here)

1. **The roadmap's CBMC ground truth for these three harnesses is wrong.** It records CBMC proving them `SUCCESSFUL` (`0 of 2826 failed`); CBMC 6.10.0 actually reports `VERIFICATION FAILED` (`1 of 85`, `4 of 123`) on reachability markers. The real defect was narrower than described: ESBMC failed an alignment check that CBMC passes.
2. **Alignment does *not* underlie the 41-crash `align_offset` / `align_to` cluster**, contrary to the roadmap's hypothesis. `check_align_offset_u32` still aborts with `ERROR: Can't construct rvalue reference to array type during dereference` on this branch — an independent dereference limitation that needs its own fix.


lucasccordeiro:
This PR (100s):

```
  correct:           23350
    correct true:    15385
    correct false:    7965
  incorrect:             9
    incorrect true:      0
    incorrect false:     9
  unknown:           13244
  Score:             38591 (max: 61276)

https://github.com/esbmc/esbmc/actions/runs/29573621625/job/87866941636
```

Master (100s):

```
  correct:           22819
    correct true:    15321
    correct false:    7498
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           13777
  Score:             38028 (max: 61276)

https://github.com/esbmc/esbmc/actions/runs/29587164519/job/87913576991
```

PR #6204, opened 2026-07-20
[pointer-analysis] Build references to whole-array values

rafaelsamenezes:
## What

`build_reference_rec` aborted whenever the dereference *target* type was an array:

```
Can't construct rvalue reference to array type during dereference
```

The comment is C-centric: in C, arrays decay to pointers, so an array-typed rvalue never arises. But languages with first-class array values do produce one — Rust's `let a: [u8; N] = *ptr;` reaches ESBMC as a whole-array load through a CBMC goto-binary (`ASSIGN var = *((unsigned char [8] *)…)`). Every destination-array case fell straight through to `abort()`.

This was **the largest single crash cluster** in the `verify-rust-std` / Kani corpus per `docs/cprover-support-roadmap.md`: 41 harnesses (`slice` 32, `ptr` 7, `num` 2), the `align_offset` / `align_to` / `align_to_mut` family.

## Fix

Add `construct_array_ref`, dispatched from the array-destination case instead of aborting:

- **Fast path** — a whole-object load at offset 0 with matching layout (the common case, e.g. a `[u8; N]` value load) returns the source unchanged, or bitcasts a same-width scalar. Avoids an N-element unroll.
- **General path** — read each element from the source at its own offset and assemble a `constant_array`, mirroring the element-wise construction the struct/scalar destination cases already use. Each per-element recursion re-dispatches on the source shape, so array, struct, scalar-reinterpret and union sources — and nested (multidimensional) arrays — are handled uniformly. A non-constant array size (flexible/infinite member) is a `dereference_failure` rather than a crash.

## Validation

Against the local Kani corpus (`esbmc --binary --function <harness>`, cross-checked on the identical binary with CBMC 6.10.0):

- **Whole `align_offset` / `align_to` cluster — 139 harnesses — zero `rvalue reference to array` aborts.** Each now reaches a verdict or a solver timeout at the 45s cap (the timeouts lower to GOTO cleanly — axis-3 solver performance, not a crash).
- **Verdict parity** where both tools decide: e.g. `check_align_offset_u32` → `VERIFICATION FAILED` in both, ESBMC's first-failing property (`core_0`, `mod.rs:2240`) is exactly one of CBMC's reachability markers.
- **The element-wise reinterpret path matches CBMC too**: `align_to_from_u32…align_to_u16` (transmutes `[u8]`→`[u16]`, forcing the general path) → FAILED in both.
- `467/467` unit tests pass; targeted native array/deref/union/cast regressions clean.

The C and Python frontends decay array rvalues before this point (verified by instrumentation: no native C/C++/goto-cc construct reaches `construct_array_ref`), so the path is exercised only via CBMC goto-binaries.

## Regression test

`regression/goto-transcoder/cbmc_deref_array/` pins the crash with the `check_align_offset_u32` Kani binary (184KB, CBMC v6 — consistent with the existing `checked_unchecked_i8` Kani fixture in that suite). Confirmed faithful: it **aborts without the fix** and reaches `VERIFICATION FAILED` with it.

Two notes on the test:
- It uses the `_u32` width deliberately. `_u8` was already crash-free before this change (it avoids the array-typed target — verified it does not reach the new code), so a `_u8` fixture would guard nothing.
- There is **no `SUCCESSFUL` companion**: every Kani harness that reaches this path fires `reachability_check` coverage markers, so all reachable harnesses are `FAILED`. The soundness/passing direction is instead covered by CBMC per-property parity, where the substantive checks pass in both tools. The obsolete `library.goto` (pre-dating in-process additions synthesis) is omitted.

## Follow-ups (not in this PR)

The other two residual `ptr`/`slice` crash clusters from the global sweep are untouched: the 12 `Bitwuzla error` aborts on `non_null_check_write_unaligned_*`, and the 23 silent segfaults (`nonzero_check_cmp_for_*`, `check_vecdeque_swap`). Axis-3 solver-performance timeouts in `slice` remain the largest non-crash bucket.


src/pointer-analysis/dereference.cpp
```diff
@@ -1297,6 +1299,67 @@ void dereferencet::construct_from_array(
...
+  // assemble them. Each recursive call re-dispatches on the source shape, so
+  // this transparently handles array, struct, scalar-reinterpret and union
+  // sources, and nests for multidimensional arrays.
+  const expr2tc source = value;
+  const BigInt count = to_constant_int2t(arr_type.array_size).value;
+  const BigInt elem_bits = type_byte_size_bits(arr_type.subtype, &ns);
+
+  std::vector<expr2tc> elements;
+  elements.reserve(count.to_uint64());
+  for (BigInt i = 0; i < count; i += 1)
```
lucasccordeiro, line 1348: Unbounded unroll: a large constant array size builds N element reads and OOMs where the code previously aborted. Cap `count` and fall back to `dereference_failure`.

src/pointer-analysis/dereference.cpp
```diff
@@ -982,10 +982,12 @@ void dereferencet::build_reference_rec(
...
   {
-    log_error(
-      "Can't construct rvalue reference to array type during dereference\n"
-      "(It isn't allowed by C anyway)\n");
-    abort();
+    // Reading a whole array value through a pointer. C forbids array-typed
+    // rvalues (arrays decay to pointers), but languages where arrays are
+    // first-class values (e.g. Rust `let a: [u8; N] = *ptr;`) reach here via a
+    // CBMC goto-binary. Build the array element by element from the source.
+    construct_array_ref(value, offset, type, guard, mode, alignment);
```
lucasccordeiro, line 989: In WRITE mode this yields a `constant_array2tc`, which cannot be an assignment target, so the store is silently dropped. Reject non-READ modes explicitly.

src/pointer-analysis/dereference.cpp
```diff
@@ -1297,6 +1299,67 @@ void dereferencet::construct_from_array(
...
+    }
+  }
+
+  // General path: read each element from the source at its own offset and
+  // assemble them. Each recursive call re-dispatches on the source shape, so
+  // this transparently handles array, struct, scalar-reinterpret and union
+  // sources, and nests for multidimensional arrays.
+  const expr2tc source = value;
+  const BigInt count = to_constant_int2t(arr_type.array_size).value;
+  const BigInt elem_bits = type_byte_size_bits(arr_type.subtype, &ns);
```
lucasccordeiro, line 1344: `type_byte_size_bits` throws `array_size_excp` on a dynamically sized subtype, which nothing here catches. Route it to `dereference_failure`.

regression/goto-transcoder/cbmc_deref_array/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
```
lucasccordeiro, line 1: Only one test for a new dereference path; the repo requires a failing counterpart. Add a test whose array load makes an assertion fail.

lucasccordeiro (changes requested):

src/pointer-analysis/dereference.cpp
```diff
@@ -1102,10 +1102,12 @@ void dereferencet::build_reference_rec(
     flags |= flag_dst_scalar;
   else if (is_array_type(type))
```
lucasccordeiro, line 1103: The dispatch legend above (line 1034) still lists `* | A | * | <unsupported>: "Can't construct rvalue ref..."`. Update that row to `construct_array_ref`.

src/pointer-analysis/dereference.cpp
```diff
@@ -1428,6 +1430,67 @@ void dereferencet::construct_from_array(
...
+  elements.reserve(count.to_uint64());
+  for (BigInt i = 0; i < count; i += 1)
+  {
+    expr2tc elem_offset = add2tc(
+      offset->type, offset, constant_int2tc(offset->type, i * elem_bits));
+    simplify(elem_offset);
+
+    expr2tc element = source;
+    build_reference_rec(
+      element, elem_offset, arr_type.subtype, guard, mode, alignment);
```
lucasccordeiro, line 1487: Element `i` sits at `offset + i*elem_bits`, so its guaranteed alignment is weaker than the array's. Passing `alignment` unchanged over-approximates it and suppresses `check_alignment` on dynamic offsets; narrow it per element.

src/pointer-analysis/dereference.cpp
```diff
@@ -1428,6 +1430,67 @@ void dereferencet::construct_from_array(
...
+    value = make_failed_symbol(type);
+    return;
+  }
+
+  // Fast path: loading the whole object at offset 0 with a matching layout,
+  // which is the common case (e.g. a Rust `[u8; N]` value load). Avoids
+  // unrolling into N element reads.
+  if (is_constant_int2t(offset) && to_constant_int2t(offset).value == 0)
+  {
+    if (base_type_eq(value->type, type, ns))
```
lucasccordeiro, line 1460: Codecov on this head shows 1461 and 1464-1465 never hit: the fast path is taken by no committed test despite being described as the common case. Add a fixture that exercises it.

regression/goto-transcoder/cbmc_deref_array/test.desc
```diff
@@ -0,0 +1,4 @@
...
+main.goto
+--binary --function _RNvNtNtCsfemxtvIyyHd_4core3ptr6verify22check_align_offset_u32
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone still passes if the constructed array is wrong, since the reachability marker fires regardless. Add the expected property/location line, as the sibling `*_fail` tests do.

Issue #6214, opened 2026-07-20
--generate-ctest-testcase silently overwrites CMakeLists.txt in the working directory


lucasccordeiro:
## Summary

`--generate-ctest-testcase` writes `CMakeLists.txt` into the current working directory unconditionally — no existence check, no warning, no opt-out. If the CWD already contains a `CMakeLists.txt`, it is destroyed with no indication that anything was lost.

## Reproducer

```
$ cd some-project-with-a-cmakelists
$ cat CMakeLists.txt
PROJECT FILE - DO NOT DELETE

$ esbmc main.cpp --generate-ctest-testcase
VERIFICATION FAILED

$ head -2 CMakeLists.txt
# Auto-generated by ESBMC 8.4.0
cmake_minimum_required(VERSION 3.10)
```

The original content is gone. Nothing in the output mentions the overwrite.

I hit this for real while working on #6207: running the generator from a directory inside an ESBMC checkout clobbered `regression/CMakeLists.txt`, a 600-line tracked file. It was only noticed because `git status` flagged it afterwards — outside a version-controlled directory the loss would have been silent and permanent.

## Cause

`write_cmake` (`src/goto-symex/ctest.cpp`) opens the path directly:

```cpp
std::ofstream cmake("CMakeLists.txt");
```

`esbmc_verifier.h` and `test_case*.c` are written the same way and have the same exposure, though a collision there is less likely.

Note also that `ctest_generator::generate_single` takes an `output_dir` parameter and immediately discards it (`(void)output_dir;`), so there is currently no way to redirect the output anywhere else.

## Suggested fix

Any of, roughly in order of preference:

1. Honour the existing `output_dir` parameter and default it to a dedicated subdirectory rather than the CWD.
2. Refuse to overwrite an existing `CMakeLists.txt` that ESBMC did not generate — the emitted files carry an `# Auto-generated by ESBMC` first line, which is a reliable marker to check for.
3. At minimum, log a warning naming each file being overwritten.

## Version

ESBMC 8.4.0, master `cf6a8d56f9`, aarch64 macOS.

PR #6216, opened 2026-07-20
[test] classify .C/.ipp/.cu sources as C++ in generated CTest projects

### LINKED ISSUE #6213: --generate-ctest-testcase: .C/.ipp/.cu inputs get a C-only project and fail to link
## Summary

`is_cpp_source` in `src/goto-symex/ctest.cpp:16-23` recognises only `.cpp`, `.cc` and `.cxx`, while ESBMC's authoritative extension table at `src/langapi/mode.cpp:19-24` also lists `.C`, `.ipp` and `.cu`. A C++ source with one of the missing extensions is verified as C++ but described to CMake as C, so CMake drops it from the target and the generated project fails to link — the same end result as #6208, reached by a different route.

## Reproducer

```cpp
// prog.C
#include <cassert>
extern "C" int __VERIFIER_nondet_int(void);
int main() { int x = __VERIFIER_nondet_int(); assert(x != 42); return 0; }
```

```
$ esbmc prog.C --generate-ctest-testcase
VERIFICATION FAILED
```

ESBMC parses it as C++ (the `<cassert>` include and `extern "C"` both resolve), but emits:

```cmake
project(ESBMCGeneratedTest C)
add_executable(test_case /abs/path/prog.C test_case.c)
```

`test_case.c` is also written with the C body rather than the `extern "C"`-wrapped C++ one. Building:

```
$ mkdir build && cd build && cmake .. && cmake --build .
Undefined symbols for architecture arm64:
  "_main", referenced from: <initial-undefines>
clang: error: linker command failed with exit code 1
```

CMake, told the project is C-only, does not compile `prog.C` at all, so `main` is never defined.

## Cause

```cpp
static bool is_cpp_source(const std::string &filename)
{
  size_t dot = filename.rfind('.');
  if (dot == std::string::npos)
    return false;
  std::string ext = filename.substr(dot);
  return ext == ".cpp" || ext == ".cc" || ext == ".cxx";
}
```

versus `src/langapi/mode.cpp:19-24`, which additionally has `cu`, `ipp` and `C`.

## Suggested fix

Have `is_cpp_source` consult the `langapi` table rather than maintaining a private copy, so the two cannot drift again. Two caveats for whoever picks this up:

- `.C` is case-sensitive and is excluded under `_WIN32` in `mode.cpp`; a shared helper must preserve that guard.
- `.cu` needs more than reclassification. Calling it C++ is closer to correct than calling it C, but a CUDA source in a generated project really wants `enable_language(CUDA)`, so it is worth handling deliberately rather than folding into the same one-line change.

## Version

ESBMC 8.4.0, master `cf6a8d56f9`, aarch64 macOS.

lucasccordeiro:
`is_cpp_source` kept a private list of C++ extensions that had drifted from the authoritative table in `langapi/mode.cpp`, so a `.C`/`.ipp`/`.cu` input was verified as C++ but described to CMake as C, dropping it from the target and leaving the generated project failing to link.

It now delegates to `language_id_by_path`, which also inherits the `_WIN32` exclusion of `.C`. CUDA additionally needs `enable_language(CUDA)`, so `.cu` emits `project(... CXX CUDA)` with coverage flags restricted to the C++ sources, as nvcc rejects `--coverage`.

Three regression tests cover the new extensions, each confirmed failing before the fix; `ctest_gen_multifile` gains a `project(... C)` assertion to pin the negative direction.

Fixes #6213


lucasccordeiro:
Closing as already-merged rather than superseded — the change shipped intact.

#6219 was stacked on this branch and squash-merged first as `4c62c8f597`, which swept both commits here (`a26fafb61`, `d240e940`) into `master`. The squash left no ancestry link, so GitHub couldn't see this PR as already-contained and flagged it `CONFLICTING` instead.

The #6213 fix is live on master: `is_cpp_source` delegates to `language_id_by_path` (`src/goto-symex/ctest.cpp:20-23`), `is_cuda_source` drives `project(... CXX CUDA)`, and the three `ctest_gen_{cu,dot_C,ipp}` CORE tests are present and registered.

Merging `master` in confirmed it: every conflict resolved toward master (the branch side was uniformly pre-#6219 code, and taking it would have reverted the no-clobber fix for #6214), leaving a tree byte-identical to master — 0 changed files. Nothing left to merge.

PR #6222, opened 2026-07-21
[C verificaiton] fix spurious loop pruning from stale phi-assignment intervals

XLiZHI:
The online interval domain (--interval-symex-guard) was producing spurious VERIFICATION FAILED on loops containing if-else blocks. After one loop iteration the guard evaluator incorrectly concluded the loop exit condition was always true and pruned the loop continuation path.

Root cause: the domain maps variables by base name (symbol2t::thename), so both SSA versions of a variable in an if-else share the same key. The execution order is:

if-branch:   nxt_idx = (cur+prv)/2  →  domain["nxt_idx"] = [0,4]
else-branch: nxt_idx = 0            →  domain["nxt_idx"] = [0,0]   ← overwrites
phi-merge:   phi_function queries domain["nxt_idx"] for both sides
             → JOIN([0,0], [0,0]) = [0,0]  instead of JOIN([0,4], [0,0]) = [0,4]

With nxt_idx stuck at [0,0], the domain deduced cur_idx == prv_idx was always true and set new_guard_true, dropping the loop-exit condition from the guard and triggering a false counterexample.

Fix

At the end of each if-branch (when symex_goto stashes the merge_stateT), take a deep copy of the domain and store it in a new merge_stateT::interval_snapshot field. In phi_function, instead of querying the (already-overwritten) domain for both sides, JOIN the if-branch snapshot value with the current else-branch domain value via phi_join_with_snapshot.

The deep copy is necessary because the domain uses copy-on-write semantics: a shared_ptr alias would be mutated by the else-branch's first write.

Storing the snapshot directly in merge_stateT (rather than an external map keyed by pointer) ties the snapshot lifetime to the merge-state object, eliminating the need for explicit cleanup and any dangling-pointer risk.

### DIFF
--- a/regression/esbmc/github_interval_guard_phi/test.desc
+++ b/regression/esbmc/github_interval_guard_phi/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--unwind 5 --no-unwinding-assertions
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_interval_guard_phi/main.c
+++ b/regression/esbmc/github_interval_guard_phi/main.c
@@ -1,0 +1,22 @@
+#define SIZE 10
+typedef int idx_t;
+typedef char data_t;
+int main()
+{
+  data_t vec[SIZE];
+  idx_t prv_idx = SIZE - 1;
+  idx_t cur_idx = 0;
+  idx_t nxt_idx;
+  int loop_idx_vfy = 0;
+  while (cur_idx != prv_idx)
+  {
+    if (vec[cur_idx] == 3)
+      nxt_idx = (cur_idx + prv_idx) / 2;
+    else
+      nxt_idx = 0;
+    prv_idx = cur_idx;
+    cur_idx = nxt_idx;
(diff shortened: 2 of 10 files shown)


regression/esbmc/github_interval_guard_phi/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unwind 5 --no-unwinding-assertions
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: @XLiZHI: Shall we also add a regular expression to ensure we can show the variable assignment/value in the counterexample?

lucasccordeiro (changes requested):
BTW, the C/C++ coding style CI is failing.

regression/esbmc/github_interval_guard_phi/test.desc
XLiZHI: Sure, added a fail variant that checks counterexample variable assignments are correctly shown in the trace.

PR #6233, opened 2026-07-22
[util] Smart cache for extracted internals

### LINKED ISSUE #6223: [feature] ESBMC internals smart cache
ESBMC extracts headers to `/tmp` at every invocation. This is inefficient and also bloats `/tmp` with a bunch of ESBMC folders. Although I don't think it causes any memory errors. My suggestion is the following:

1. At compile time we walk the OM tree and calculate a checksum signature.
2. We put it in the source code using a macro or something dynamic so that that version of ESBMC knows what checksum it's OM files own.
3. When ESBMC starts we check `/tmp` and if a collision is found, we use that. If not, we extract. This way, no more bloat. The trick is, use `/tmp/esbmc-slug` and slug is the checksum, so then we don't need to even iterate over all the `/tmp/esbmc-...` folders.

Yiannis128:
Fixes #6223.

ESBMC writes its bundled clang headers, operational models and helper scripts to disk on every run, creating fresh `/tmp/esbmc-headers-*`, `esbmc.*` and similar directories and deleting them at exit. Over a regression run that is thousands of extractions and a lot of churn through `/tmp`.

This caches each payload. The first run that needs one extracts it into `/tmp/esbmc-cache-<uid>/<name>-<digest>` (mode 0700), and every later run reuses it while the content is unchanged. A trivial run drops from about 158 ms to 141 ms and the suite stops filling `/tmp`.

How it works:

- The cache key is a content digest computed at build time by `scripts/flail.py`, carried as a field of the `ESBMC_FLAIL` row for each embedded file, so it cannot drift from the shipped bytes. Runtime hashing was rejected: hashing the 7.4 MB of headers costs about 75 ms, more than the roughly 7 ms extraction it would guard.
- `cached_extract_dir` extracts into a staging dir and publishes with an atomic `rename`, so parallel runs never observe a partial entry and a hit is complete by construction. A failed extraction (full disk, unusable root) falls back to a per-run temp dir instead of publishing a truncated tree, which for a verifier is the difference between a clean failure and a wrong verdict from truncated headers.
- The cache root is created 0700 and its ownership is confirmed before use, so on a shared machine a leftover or wrong-mode directory of the predictable name is declined and the run falls back rather than reusing something that is not ours.
- The Python parser writes its AST JSON into its output directory, so that stays a per-process directory; only the read-only parser scripts and model sources are cached and shared, exposed to the parser via a symlink.

Configuration: `ESBMC_CACHE_DIR` relocates the cache (for CI pointing at a writable volume, declined with a warning if it is not a 0700 dir we own); `-DESBMC_CACHE_IN_HOME=On` uses `$XDG_CACHE_HOME/esbmc` instead of the temp dir.

New Catch2 cases in `unit/util/filesystem.test.cpp` cover reuse across calls, distinct content keys, entry persistence, atomic publish with no staging residue, private-root ownership, and the fall-back when extraction throws.


lucasccordeiro (commented):
Review of the extracted-internals cache. The design (content-keyed entries, staging + atomic rename, decline-and-fall-back on a root we do not own) is the right shape and the python per-process output-dir fix is correct. Findings below, roughly in severity order: the short-write guard in `create_path_and_write` does not actually cover buffered writes, the fallback path can now terminate the process uncaught, the cache key omits the extraction layout, and `getuid()`/`geteuid()` are used inconsistently for the same check. Remaining items are eviction/staging leaks, error handling around the python `models` symlink, and test coverage.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
   if (!boost::filesystem::exists(p.parent_path()))
     boost::filesystem::create_directories(p.parent_path());
 
-  std::ofstream(path).write(s, n);
+  /* Report a short write (ENOSPC/EIO/EDQUOT) rather than silently truncating.
+   * cached_extract_dir() relies on this to avoid publishing a partial entry. */
+  std::ofstream f(path);
+  f.exceptions(std::ios::failbit | std::ios::badbit);
+  f.write(s, n);
```
lucasccordeiro, line 219: The short-write guard does not cover the case it was added for. `f.write()` only reports an error if the write overflows the stream buffer inline; for any payload smaller than the buffer (which is most bundled headers and every `.py` model) the bytes sit in the filebuf and are flushed by `~basic_ofstream`, whose destructor cannot throw and does not set `badbit`. An ENOSPC on that final flush is silently lost — i.e. exactly the truncated-tree-gets-published scenario `cached_extract_dir()` relies on this to prevent.

Flush explicitly while exceptions are still armed:

```cpp
std::ofstream f(path);
f.exceptions(std::ios::failbit | std::ios::badbit);
f.write(s, n);
f.flush();   // or f.close();
```

Separately, `f.exceptions(...)` on an already-failed open throws via `clear(rdstate())` — that is the intended behaviour here but it is entirely implicit; an explicit `if (!f) throw` would document it.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
...
+       * a guard: a transient failure (a full disk that was since freed, a lost
+       * race) recovers there, while a deterministic one re-throws and
+       * propagates rather than being silently hidden. */
+    }
+  }
+
+  /* No usable cache root, or extraction/publish failed: extract into a private
+   * temp dir removed on exit — the behaviour from before the cache existed. */
+  tmp_path dir = create_tmp_dir(name + "-%%%%-%%%%-%%%%");
+  extract(dir.path());
```
lucasccordeiro, line 375: The fallback re-runs `extract` outside the `try`, so a deterministic failure now propagates out of `cached_extract_dir()`. `main()` has no top-level handler (`src/esbmc/main.cpp`), and these calls happen inside function-local static initialisers (`clang_resource_dir()`, `abstract_cpp_includes()`), so a full disk turns into `std::terminate`/abort with no diagnostic. Before this PR the same condition produced a truncated-but-running verifier. Neither is great, but an uncaught throw is a regression in failure quality — catch here and surface it through `log_error()` with a clean exit, or at minimum document the new failure mode in the header doc.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
...
+#ifndef _WIN32
+  if (mkdir(p.c_str(), 0700) && errno != EEXIST)
+    return false;
+
+  struct stat st;
+  if (lstat(p.c_str(), &st))
+    return false;
+
+  /* mkdir() grants ownership to the effective uid, so compare against that. */
+  return S_ISDIR(st.st_mode) && st.st_uid == geteuid() && !(st.st_mode & 077);
```
lucasccordeiro, line 243: `geteuid()` here vs `getuid()` at line 298 when building the root name. Under a setuid/setgid ESBMC these disagree and the check can never succeed, so the cache silently degrades to per-run extraction forever. Pick one consistently — `geteuid()` is the correct choice since that is the uid `mkdir()` assigns. Note the new unit test asserts `getuid()`, encoding the same inconsistency.

Also: `mkdir(p, 0700)` is masked by umask. A umask that strips owner bits yields mode 0000, which passes `!(st.st_mode & 077)` and `S_ISDIR` but is untraversable, so every subsequent write throws and the run silently falls back on each invocation. Add `(st.st_mode & 0700) == 0700` to the accept condition.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
...
+
+  fs::create_directories(root.parent_path(), ec);
+  if (ensure_private_dir(root))
+    return root;
+
+  /* A user who set ESBMC_CACHE_DIR expects it to be used; if it is not a
+   * directory we own at 0700 we cannot, so say so rather than silently
+   * reverting to per-run extraction. The default root falls back quietly. */
+  if (explicit_dir && *explicit_dir)
+    fprintf(
```
lucasccordeiro, line 313: Two issues:

1. Raw `fprintf(stderr, ...)` bypasses ESBMC's output routing. Use `log_warning()` from `util/message/message.h` like the rest of the codebase.
2. The accept rule is stricter than users will expect. A CI script doing `mkdir -p "$ESBMC_CACHE_DIR"` gets 0755, which is declined — so the documented "set `ESBMC_CACHE_DIR` for CI" workflow warns and falls back on *every* invocation unless the operator happens to have chmod'ed it. Either `chmod(p, 0700)` when the directory is already owned by us, or state the 0700 requirement in the warning text and in CLAUDE.md.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
...
+  namespace fs = boost::filesystem;
+
+  static const fs::path root = cache_root();
+  if (!root.empty())
+  {
+    /* 64 bits of SHA-256 keeps the path short; the key only has to separate
+     * the payloads of distinct builds, not resist collisions. */
+    fs::path target =
+      root /
+      (name + "-" + picosha2::hash256_hex_string(content_key).substr(0, 16));
```
lucasccordeiro, line 335: The key covers the embedded bytes but not the extraction *layout*. `extract` decides where files land (`root + "/include"` here, `/headers`, `/library`, `/library/libm` in `cprover_libc_sources.cpp`). Change any of those without touching a single bundled file and every pre-existing entry is a stale-layout hit that is silently reused — and since `mangle()` invokes flail.py with relative paths under `WORKING_DIRECTORY ${dir}`, keys are build-location *and* build-version independent, so entries are shared across separately installed ESBMC builds. Fold a layout/schema version constant (or `ESBMC_VERSION`) into `content_key` before hashing.

src/util/base/filesystem.cpp
```diff
@@ -196,5 +212,166 @@ void file_operations::create_path_and_write(
...
+    try
+    {
+      /* The tmp_path guard removes the staging dir at every exit from this
+       * scope; on the publish path that is a no-op, the rename having moved
+       * it away already. */
+      tmp_path staging(with_unique_path(
+        root, "." + name + ".tmp-%%%%-%%%%-%%%%", [](const fs::path &p) {
+          return fs::create_directory(p);
+        }));
+      register_tmp_for_cleanup(staging.path());
```
lucasccordeiro, line 353: Staging directories are only reaped by the process that created them. A `SIGKILL`, OOM kill or power loss mid-extraction leaves `.<name>.tmp-*` in the cache root permanently, and nothing here ever prunes them.

More broadly there is no eviction at all: every rebuild that touches any bundled file mints a new ~7.4 MB entry. The "systemd-tmpfiles ages it out" argument holds only for the Linux `/tmp` default — it does not hold for `ESBMC_CACHE_IN_HOME`, for a persistent CI `ESBMC_CACHE_DIR`, or for macOS `$TMPDIR`, which are precisely the two configurations this PR adds. Consider sweeping stale `.<name>.tmp-*` (and optionally entries older than N days) when the root is opened.

src/python-frontend/python_language.cpp
```diff
@@ -50,31 +50,53 @@ extern "C"
...
+{
+  static const auto dir = [&]() {
+    auto d = file_operations::create_tmp_dir("esbmc-python-out-%%%%-%%%%-%%%%");
+    const fs::path src = fs::path(python_scripts_dir()) / "models";
+    const fs::path dst = fs::path(d.path()) / "models";
+    boost::system::error_code ec;
+    if (deadlock_check)
+      fs::copy(src, dst, fs::copy_options::recursive, ec);
+    else
+      fs::create_directory_symlink(src, dst, ec);
```
lucasccordeiro, line 94: `ec` is discarded. If the symlink cannot be created — Windows without developer mode/admin, a filesystem without symlink support, or `dst` already existing — `<out>/models` simply does not exist. `_emit_model_jsons()` then globs nothing, and the failure only surfaces much later in `python_converter.cpp` as

    Python frontend: missing operational-model AST '...'. This usually means parser.py exited before generating it

which is a misleading diagnostic for a symlink failure. Check `ec` and fall back to the recursive `fs::copy` already used on the deadlock path, or report it explicitly. The same applies to the `fs::copy` call above.

src/python-frontend/python_language.cpp
```diff
@@ -50,31 +50,53 @@ extern "C"
...
 
+// Per-process directory the parser writes its AST JSON into. It cannot be the
+// shared script cache: the parser emits <name>.json here and, under
+// --deadlock-check, rewrites models/threading.py, so concurrent ESBMC runs
+// would clobber each other. The parser reads model sources only from
+// <dir>/models, so expose the cached models there — a symlink in the common
+// case, or a private copy when --deadlock-check must mutate them.
+static const std::string &python_output_dir(bool deadlock_check)
+{
+  static const auto dir = [&]() {
```
lucasccordeiro, line 86: The function-local static captures whichever `deadlock_check` value it is first called with. A second call with the opposite value returns a directory whose `models` is a symlink into the shared cache, and `select_threading_model()` would then rewrite `threading.py` *inside the shared entry* — reintroducing exactly the cross-process clobbering commit 92e10df set out to remove. Today `parse()` always passes the same value so this is latent rather than live, but the signature invites the bug. Drop the parameter and read `config.options` inside the initialiser instead.

src/c2goto/cprover_libc_sources.cpp
```diff
@@ -17,79 +14,84 @@ extern "C"
...
-  void foreach_libc_libm_(F &&f [[maybe_unused]])
+  /* Both trees are written together: unlike the previous per-tree laziness
+   * this runs at most once per build, since the result is cached. */
+  static void extract(const std::string &root)
   {
-#define ESBMC_FLAIL(body, size, ...) f(body, size, libc + "/" #__VA_ARGS__);
+    /* The path variables are only read inside the ESBMC_FLAIL expansions of
+     * the generated includes, which cppcheck cannot resolve. */
+    // cppcheck-suppress unreadVariable
+    const std::string headers = root + "/headers";
```
lucasccordeiro, line 45: `headers` shadows the enclosing class's `std::string headers` member, and there are now three `// cppcheck-suppress unreadVariable` annotations for what is a tool artifact. Renaming the locals (`headers_dir`, `libc_dir`, `libm_dir`) removes the shadowing and reads better; it may also drop some of the suppressions since the names then appear in a wider expression context.

unit/util/filesystem.test.cpp
```diff
@@ -86,3 +123,144 @@ TEST_CASE(
...
+  // entry is published, whether we won the rename or lost it.
+  unsigned staging = 0;
+  for (const auto &e : boost::filesystem::directory_iterator(expected_root()))
+    if (e.path().filename().string().rfind(".unit-test.tmp-", 0) == 0)
+      ++staging;
+  REQUIRE(staging == 0);
+}
+
+TEST_CASE(
+  "a throwing extract is not published and falls back",
```
lucasccordeiro, line 207: The design's load-bearing invariants are not the ones covered here. Missing:

- **`create_path_and_write` throwing on a short write.** This test uses a hand-rolled throwing lambda, so the real writer — the thing the "never publish a partial tree" argument depends on — is never exercised. (See the flush comment on `filesystem.cpp`.)
- **`ESBMC_CACHE_DIR` rejection**: a 0755 or foreign-owned directory must warn and fall back. This is the path CI users will hit first.
- **The lost-publish race**: `target` already present when `rename` runs, i.e. `!ec || fs::is_directory(target)`. It is asserted in comments but never executed.

Also every new case is `#ifndef _WIN32`, so the Windows branch of `ensure_private_dir()` and the Windows behaviour of the python `models` symlink are untested. Per the repo convention the PR carries no regression tests either; a `regression/python` case exercising `--deadlock-check` twice in the same tree would pin the models-copy path end to end.

CLAUDE.md
```diff
@@ -64,7 +64,7 @@ ctest -R "regression/esbmc/00_big_endian_01" --output-on-failure
...
+**Extracted internals cache.** ESBMC materialises its bundled clang headers, operational models and helper scripts on disk before clang can read them. These are cached in `/tmp/esbmc-cache-<uid>/` (mode 0700, one entry per payload keyed by a content digest), so a given build extracts each payload once rather than once per run — the suite no longer fills `/tmp` with thousands of `esbmc-headers-*` directories. Entries total a few MB and are safe to delete at any time; they are re-extracted on demand, and `systemd-tmpfiles` ages out stale ones from old builds on its own. Set `ESBMC_CACHE_DIR` to relocate the cache (useful for CI), or configure `-DESBMC_CACHE_IN_HOME=On` to use `$XDG_CACHE_HOME/esbmc` (default `~/.cache/esbmc`) instead, which survives reboots but grows until cleared by hand. On a shared machine the per-uid root is created 0700 and its ownership is confirmed before use, so a leftover or wrong-mode directory of that name is declined and ESBMC falls back to per-run extraction rather than reusing it.
```
lucasccordeiro, line 67: Two claims here are not accurate:

- *"Entries ... are safe to delete at any time"* — not while an ESBMC process is running. `clang_resource_dir()` hands the entry path to clang, and the python `models` symlink points into the entry for the lifetime of the run; deleting it mid-run breaks both. Previously the directory was private to the run, so this genuinely was safe. It is also unsafe to delete *within* an entry: `cached_extract_dir()` treats any existing `target` directory as complete and never revalidates its contents, so a partially-emptied entry is trusted forever. Suggest "delete whole entries when no ESBMC process is running".
- *"`systemd-tmpfiles` ages out stale ones"* — Linux-specific, and does not apply to the two configurations this PR introduces (`ESBMC_CACHE_IN_HOME`, a CI-set `ESBMC_CACHE_DIR`) nor to macOS `$TMPDIR`. Worth qualifying, since there is no eviction in the code.

Yiannis128:
Cancelling this in favor of https://github.com/esbmc/esbmc/issues/6509

PR #6244, opened 2026-07-22
[python] Fix core dump on method call on attribute assigned from self.method() (#6242)

### LINKED ISSUE #6242: [python] crash (SIGABRT/core dump): calling a method on an attribute assigned from a method-call return
### Summary

ESBMC's Python frontend **aborts (SIGABRT, core dump)** when a method is invoked on an instance attribute that was assigned from a *method-call return value*. Instead of emitting a clean error (or handling the call), GOTO conversion hits a missing function symbol and calls `abort()`.

This pattern is common in real Python — an attribute assigned from a method-call return, then a method invoked on that attribute — so it blocks verification of a large class of programs.

### Minimal reproducer

```python
class Talker:
    def make(self) -> int:
        return 5

    def __init__(self):
        self.pub = self.make()

    def go(self):
        self.pub.publish(1)

Talker().go()
```

### Command

```sh
esbmc repro.py --incremental-bmc
```

### Actual output

```
WARNING: migrate_expr: symbol 'py:repro.py@C@Talker@F@publish' missing renaming delimiters, ...
ERROR: Function `py:repro.py@C@Talker@F@publish' not found
<core dumped> (SIGABRT, exit 134)
```

### Expected

ESBMC should not abort on parseable input. It should either report a clean frontend error (the attribute's type does not have a `publish` method) or model the call. A core dump is never an acceptable outcome for a verifier.

### Root cause (initial analysis)

The receiver `self.pub`'s type is mis-resolved. When an attribute is assigned from a call return (`self.pub = self.make()`), the member-call resolver in `src/python-frontend/function_call/builder.cpp` falls back to the **enclosing class** (`Talker`) as the receiver type, so `self.pub.publish(...)` is emitted as a call to the symbol `@C@Talker@F@publish`, which does not exist. GOTO conversion then aborts at `src/goto-programs/builtin_functions.cpp:633`:

```cpp
const symbolt *symbol = ns.lookup(identifier);
if (!symbol)
{
  log_error("Function `{}' not found", id2string(identifier));
  abort();   // <-- core dump on parseable input
}
```

Two defects compound here: (1) the frontend picks the wrong receiver type, and (2) `do_function_call_symbol` aborts rather than surfacing a recoverable diagnostic.

### Notes

- A literal assignment (`self.pub = 5; self.pub.publish(1)`) does **not** crash — it reports a clean `VERIFICATION FAILED`. The crash is specific to attributes assigned from a **call** return.

LukeW1999:
## Summary

Fixes #6242 — a core dump (SIGABRT) when a method is called on an instance
attribute that was assigned from a `self.<method>()` call.

An unannotated `self.attr = self.method()` was left untyped, because the
annotator's `get_type_from_method` did not resolve a `self` receiver (it threw
`Object "self" not found`). With no inferred annotation, the attribute fell
through to `get_attributes_from_self`'s plain-`Assign` branch, which types the
member as `any_type()` and it then defaults to the **enclosing class**. A later
`self.attr.other()` therefore resolved to a nonexistent method on the enclosing
class (`@C@<Class>@F@other`), and GOTO conversion aborted with a core dump.

## Fix

Resolve `self.method()` in `get_type_from_method` against the current class,
walking the base chain (mirroring the existing `super().method()` handling), and
adopt the method's declared (`-> T`) or inferred (`return T()`) return type. The
attribute is now typed by the callee's return class.

## Effect

- **Valid pattern verifies** — `self.pub = self.make(); self.pub.publish(...)`
  where `make()` returns an object (the common ROS2-node shape) now resolves the
  call and verifies instead of crashing.
- **Invalid pattern reports cleanly** — calling a method that does not exist on
  the attribute's real type reports a normal `AttributeError` / `VERIFICATION
  FAILED` instead of a core dump.

## Tests

- `regression/python/github_6242` — valid pattern → `VERIFICATION SUCCESSFUL`.
- `regression/python/github_6242_fail` — same pattern with a wrong assertion
  (the call now yields a concrete value that is checked) → `VERIFICATION FAILED`.

Targeted `class|method|attr|inherit|polymorph|super` python regression subset
(256 tests) passes with no regressions.

## Known related (separate) case

`self.attr = param.method()` where `param` is an **unannotated parameter** still
mis-resolves the receiver to the enclosing class — a distinct root cause
(unannotated-parameter typing), not addressed here.

### DIFF
--- a/regression/python/github_6242/main.py
+++ b/regression/python/github_6242/main.py
@@ -1,0 +1,21 @@
+# github.com/esbmc/esbmc/issues/6242
+# An attribute assigned from a self-method returning an object must be typed by
+# the method's return class, so a later method call on the attribute resolves
+# instead of aborting GOTO conversion.
+class Publisher:
+    def publish(self, msg):
+        return msg + 1
+
+
+class Talker:
+    def make(self):
+        return Publisher()
+
+    def __init__(self):
+        self.pub = self.make()
+
+    def go(self):
+        return self.pub.publish(41)
+
+
+assert Talker().go() == 42
--- a/regression/python/github_6242_fail/test.desc
+++ b/regression/python/github_6242_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--no-unwinding-assertions --unwind 1
+^VERIFICATION FAILED$
--- a/src/python-frontend/python_annotation/annotation_conversion.inl
+++ b/src/python-frontend/python_annotation/annotation_conversion.inl
@@ -1985,6 +1985,46 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
     return "Any";
   }
 
+  // Handle self.method() calls: resolve the method in the current class (or,
+  // failing that, along its base chain) and adopt its return type. Without
+  // this the `self` receiver is unresolved, so an unannotated
+  // `self.attr = self.method()` is left untyped and the attribute defaults to
+  // its enclosing class — which later emits a call to a nonexistent method on
+  // that attribute and aborts GOTO conversion (#6242).
+  if (obj == "self" && !attr_name.empty() && !current_class_name_.empty())
+  {
+    for (std::string cls = current_class_name_; !cls.empty();)
+    {
+      Json class_node = json_utils::find_class(ast_["body"], cls);
+      if (class_node.empty() || !class_node.contains("body"))
+        break;
+      for (const Json &member : class_node["body"])
+      {
+        if (member["_type"] != "FunctionDef" || member["name"] != attr_name)
+          continue;
+        if (member.contains("returns") && !member["returns"].is_null())
+        {
+          const Json &ret = member["returns"];
+          if (ret.contains("id"))
+            return ret["id"].template get<std::string>();
+          if (
+            ret.contains("_type") && ret["_type"] == "Subscript" &&
+            ret.contains("value") && ret["value"].contains("id"))
+            return ret["value"]["id"].template get<std::string>();
+        }
+        std::string inferred =
+          infer_from_return_statements(member["body"], attr_name);
+        return inferred.empty() ? "Any" : inferred;
+      }
+      // Not defined in this class — continue up the first base, as super() does.
+      cls.clear();
+      if (
+        class_node.contains("bases") && !class_node["bases"].empty() &&
+        class_node["bases"][0].contains("id"))
+        cls = class_node["bases"][0]["id"].template get<std::string>();
+    }
+  }
+
   // Handle dict.keys() and dict.values()
   if (attr_name == "keys" || attr_name == "values")
   {
--- a/src/python-frontend/python_annotation/annotation_conversion.inl
+++ b/src/python-frontend/python_annotation/annotation_conversion.inl
@@ -1810,6 +1810,26 @@ std::string python_annotation<Json>::get_type_from_call(const Json &element)
 }
 
 template <class Json>
+std::string python_annotation<Json>::method_return_type(
+  const Json &member,
+  const std::string &method_name)
+{
+  if (member.contains("returns") && !member["returns"].is_null())
+  {
+    const Json &ret = member["returns"];
+    if (ret.contains("id"))
+      return ret["id"].template get<std::string>();
+    if (
+      ret.contains("_type") && ret["_type"] == "Subscript" &&
+      ret.contains("value") && ret["value"].contains("id"))
+      return ret["value"]["id"].template get<std::string>();
+  }
+  std::string inferred =
+    infer_from_return_statements(member["body"], method_name);
+  return inferred.empty() ? "Any" : inferred;
+}
+
+template <class Json>
 std::string python_annotation<Json>::get_type_from_method(const Json &call)
 {
   std::string type("");
@@ -1965,20 +1985,7 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
         {
           if (member["_type"] != "FunctionDef" || member["name"] != attr_name)
             continue;
-          if (member.contains("returns") && !member["returns"].is_null())
-          {
-            const Json &ret = member["returns"];
-            if (ret.contains("id"))
-              return ret["id"].template get<std::string>();
-            if (
-              ret.contains("_type") && ret["_type"] == "Subscript" &&
-              ret.contains("value") && ret["value"].contains("id"))
-              return ret["value"]["id"].template get<std::string>();
-          }
-          // attr_name == member["name"] by the loop invariant above
-          std::string inferred =
-            infer_from_return_statements(member["body"], attr_name);
-          return inferred.empty() ? "Any" : inferred;
+          return method_return_type(member, attr_name);
         }
       }
     }
@@ -1993,7 +2000,12 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
   // that attribute and aborts GOTO conversion (#6242).
   if (obj == "self" && !attr_name.empty() && !current_class_name_.empty())
   {
-    for (std::string cls = current_class_name_; !cls.empty();)
+    // `visited` guards against cyclic base declarations (e.g. class A(B) /
+    // class B(A)): these do not run in Python but ast.parse still accepts
+    // them, so the walk must not loop forever.
+    std::set<std::string> visited;
+    for (std::string cls = current_class_name_;
+         !cls.empty() && visited.insert(cls).second;)
     {
       Json class_node = json_utils::find_class(ast_["body"], cls);
       if (class_node.empty() || !class_node.contains("body"))
@@ -2002,19 +2014,7 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
       {
         if (member["_type"] != "FunctionDef" || member["name"] != attr_name)
           continue;
-        if (member.contains("returns") && !member["returns"].is_null())
-        {
-          const Json &ret = member["returns"];
-          if (ret.contains("id"))
-            return ret["id"].template get<std::string>();
-          if (
-            ret.contains("_type") && ret["_type"] == "Subscript" &&
-            ret.contains("value") && ret["value"].contains("id"))
-            return ret["value"]["id"].template get<std::string>();
-        }
-        std::string inferred =
-          infer_from_return_statements(member["body"], attr_name);
-        return inferred.empty() ? "Any" : inferred;
+        return method_return_type(member, attr_name);
       }
       // Not defined in this class — continue up the first base, as super() does.
       cls.clear();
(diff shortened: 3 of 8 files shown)


lucasccordeiro (changes requested):
Reviewed by building and running the patch locally. Findings below were executed against the patched binary, not inferred.

## Blocking

**1. `annotation_conversion.inl` — unbounded base-chain walk hangs on cyclic bases**

`for (std::string cls = current_class_name_; !cls.empty();)` has no visited set, so `find_class` re-finds the same node forever when the base chain cycles:

```python
class A(A):                      # syntactically valid; ast2json never executes
    def __init__(self):
        self.x = self.make()
```

Hangs (`timeout 25` → exit 124). Mutual `class B(A)` / `class A(B)` with `self.make()` hangs identically. Isolated to this patch: the same `class A(A)` **without** the `self.method()` call completes in 0s.

This regresses the exact property #6242 asks for ("ESBMC should not abort on parseable input") — a hang is worse than the abort it replaces. Add a `std::set<std::string> visited` and `break` on re-entry.

**2. Trades the abort for a false positive on overridden methods**

Resolution starts at `current_class_name_`, the class where the *assignment* appears — not the runtime type of `self`. When a subclass overrides the method, the attribute is typed by the base's return class:

```python
class P1:
    def val(self): return 1
class P2:
    def val(self): return 2
class Base:
    def make(self): return P1()
    def __init__(self): self.o = self.make()
    def go(self): return self.o.val()
class Derived(Base):
    def make(self): return P2()
assert Derived().go() == 2       # CPython: holds
```

CPython: assertion holds. Patched ESBMC: `VERIFICATION FAILED` (master: abort/134). A silent false alarm on a valid program is a worse failure mode than a crash. Either fall back to `"Any"` when a subclass of `current_class_name_` overrides `attr_name`, or land as-is with an explicit `KNOWNBUG` test and documented limitation — but not silently.

**3. Two unrelated commits duplicate already-merged #6218**

`2ae5262cf` and `0f7d91859` (docs, `options.cpp`, 5 `regression/loop-invariants/` dirs) are byte-identical to master's `bcc6b3e5b3` (#6218), which is not an ancestor of this branch:

```
git diff bcc6b3e5b3:website/content/docs/loop-invariants.md 0f7d918593:...  → IDENTICAL
```

The PR body describes only the Python fix and never mentions them. Please rebase onto master and drop both — the diff should be `annotation_conversion.inl` plus the two Python tests.

## Tests

**4. Neither test reproduces the crash in #6242.** The issue's reproducer is `def make(self) -> int` followed by `self.pub.publish(1)` — an *invalid* program that must degrade cleanly. `github_6242` and `github_6242_fail` are both the *valid* pattern, differing only in the asserted constant (`42`/`99`), so nothing pins the reported abort. I confirmed the issue reproducer goes 134 → `VERIFICATION FAILED` under the patch; please add it as a test with a regex on the actual diagnostic, not bare `^VERIFICATION FAILED$`.

**5. New code paths have no coverage.** The base-chain walk and the `returns` annotation branch are untouched by both tests — `github_6242` only exercises the `infer_from_return_statements` fallback on a same-class method. I verified both work manually (inherited `make()` → SUCCESSFUL; `-> Publisher` → SUCCESSFUL), so they are cheap to pin. The `Subscript` branch (`-> list[T]`) is also untested.

**6. `github_6242_fail/test.desc` — unexplained flag asymmetry.** Pass test uses `--unwind 1`; fail test adds `--no-unwinding-assertions`. The programs are identical apart from one constant and contain no loops, so the negative is not a controlled variant of the positive. Drop the flag and pin which property fails.

## Maintainability

**7. The new block duplicates the `super()` handler verbatim** — the `returns`/`id`/`Subscript` extraction plus the `infer_from_return_statements` fallback is copy-pasted. Extract `std::string return_type_of_method(const Json &class_node, const std::string &method)` and call it from both; the `super()` branch then becomes the one-base case of the same walk.

**8. `member["_type"] != "FunctionDef"` misses `AsyncFunctionDef`.** `async def make(self)` silently falls through to the untyped path — straight back into the #6242 crash route. Pre-existing in the `super()` branch, but worth fixing while this code is being factored out.

## Not blocking

**9. `Fixes #6242` overclaims.** The PR body itself notes `self.attr = param.method()` still mis-resolves, and the `abort()` at `src/goto-programs/builtin_functions.cpp:632` — defect (2) in the issue — is untouched. The crash remains reachable. Either fix the abort into a recoverable diagnostic or downgrade to `Refs #6242`.

## Verification performed

Targeted `class|attr|self|inherit|polymorph|super|method` Python subset: **256/256 passed** with the patched binary — the no-regression claim checks out. Full `-L python` did not finish inside the 5-minute cap, so wider regressions are unverified.

lucasccordeiro (commented):
Line-anchored version of my review findings.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -1985,6 +1985,46 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
...
 
+  // Handle self.method() calls: resolve the method in the current class (or,
+  // failing that, along its base chain) and adopt its return type. Without
+  // this the `self` receiver is unresolved, so an unannotated
+  // `self.attr = self.method()` is left untyped and the attribute defaults to
+  // its enclosing class — which later emits a call to a nonexistent method on
+  // that attribute and aborts GOTO conversion (#6242).
+  if (obj == "self" && !attr_name.empty() && !current_class_name_.empty())
+  {
+    for (std::string cls = current_class_name_; !cls.empty();)
```
lucasccordeiro: No visited set, so cyclic bases loop forever. Verified: `class A(A)` with `self.make()` hangs (timeout, exit 124); mutual `A(B)`/`B(A)` too. Add a visited set and break.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -1985,6 +1985,46 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
     return "Any";
   }
 
+  // Handle self.method() calls: resolve the method in the current class (or,
+  // failing that, along its base chain) and adopt its return type. Without
+  // this the `self` receiver is unresolved, so an unannotated
+  // `self.attr = self.method()` is left untyped and the attribute defaults to
+  // its enclosing class — which later emits a call to a nonexistent method on
+  // that attribute and aborts GOTO conversion (#6242).
+  if (obj == "self" && !attr_name.empty() && !current_class_name_.empty())
```
lucasccordeiro, line 2001: Resolves against the defining class, not runtime `self`. A subclass overriding `make()` gives a false `VERIFICATION FAILED` on a valid program (verified vs CPython). Fall back to `Any`.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -1985,6 +1985,46 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
...
+  {
+    for (std::string cls = current_class_name_; !cls.empty();)
+    {
+      Json class_node = json_utils::find_class(ast_["body"], cls);
+      if (class_node.empty() || !class_node.contains("body"))
+        break;
+      for (const Json &member : class_node["body"])
+      {
+        if (member["_type"] != "FunctionDef" || member["name"] != attr_name)
```
lucasccordeiro, line 2015: `AsyncFunctionDef` is not matched, so `async def make(self)` falls through to the untyped path and back into the #6242 crash route.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -1985,6 +1985,46 @@ std::string python_annotation<Json>::get_type_from_method(const Json &call)
...
+        {
+          const Json &ret = member["returns"];
+          if (ret.contains("id"))
+            return ret["id"].template get<std::string>();
+          if (
+            ret.contains("_type") && ret["_type"] == "Subscript" &&
+            ret.contains("value") && ret["value"].contains("id"))
+            return ret["value"]["id"].template get<std::string>();
+        }
+        std::string inferred =
```
lucasccordeiro: This duplicates the `super()` handler verbatim. Extract `return_type_of_method(class_node, method)` and call it from both.

regression/python/github_6242_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+--no-unwinding-assertions --unwind 1
```
lucasccordeiro, line 3: Why `--no-unwinding-assertions` here but not in the passing test? Neither program has a loop. Also pin which property fails, not bare `VERIFICATION FAILED`.

regression/python/github_6242/main.py
```diff
@@ -0,0 +1,21 @@
...
+        return Publisher()
+
+    def __init__(self):
+        self.pub = self.make()
+
+    def go(self):
+        return self.pub.publish(41)
+
+
+assert Talker().go() == 42
```
lucasccordeiro, line 21: Neither test reproduces #6242's actual crash (`make() -> int` then `.publish()`). The base-chain walk and `-> T` annotation paths are also uncovered.

src/esbmc/options.cpp
```diff
@@ -485,10 +485,13 @@ const struct group_opt_templ all_cmd_options[] = {
      "Set max k value for the inductive step"},
     {"loop-invariant",
      NULL,
-     "Verify using loop invariant + k-induction (combined mode)"},
+     "Verify using loop invariant + k-induction (combined mode). Still unwinds "
```
lucasccordeiro: This commit is byte-identical to already-merged #6218 (`bcc6b3e5b3`), along with the docs and loop-invariant tests. Rebase onto master and drop both commits.

LukeW1999:
Thanks @lucasccordeiro — thorough review. Point-by-point (follow-up in #6276):

- **#1 (cyclic-base hang), #7 (super()/self dedup), #5 (inherited coverage), #3 (unrelated loop-invariant commits):** already addressed by the polish commit that merged with this PR — `visited`-set guard, extracted `method_return_type()`, `github_6242_inherited`, and the two duplicate commits were rebased out (they are not in master).
- **#4 (no crash-repro test):** added `github_6242_attribute_error` in #6276 — the issue's original invalid pattern, pinned on the actual `Unsupported function 'publish' is reached` diagnostic.
- **#6 (flag asymmetry):** fixed in #6276 (`github_6242_fail` now uses plain `--incremental-bmc`).
- **#2 (overridden methods):** the suggested "fall back to `Any`" makes it *worse* — an `Any`-typed attribute re-enters the method-call path and hits the still-present `abort()` (your #9), turning the false positive into a crash. Documented as a KNOWNBUG (`github_6242_override`) in #6276; a real fix needs virtual dispatch or the abort change.
- **#8 (`async def`):** recognizing `AsyncFunctionDef` in the annotator alone still aborts on the call (coroutines unmodelled); left as a documented limitation rather than a crashing KNOWNBUG.
- **#9 (`Fixes` overclaim / reachable `abort()`):** agreed — the `abort()` at `builtin_functions.cpp:632` is the shared root for the remaining crash surface (#2/#8) and is the right place for a dedicated follow-up.

PR #6248, opened 2026-07-22
[om] Fix std::string operator>/operator< const char* overloads

### LINKED ISSUE #6229: [om] std::string const char* operator>/operator< are nondeterministic and ambiguous
## Summary

In the C++ operational model (`src/cpp/library/string`), the `const char*`
overloads of `std::string`'s `operator>` and `operator<` are broken in two
independent ways:

1. **`"lit" > s` (literal on the left) is nondeterministic.** The overload is
   declared inside the class as a non-template `friend`
   (`friend bool operator>(const char*, basic_string<CharT>&)`), but the
   out-of-line definition is a function *template*. The declared friend is
   therefore never defined, and ESBMC models the call as returning an
   unconstrained (nondet) value.

2. **`s > "lit"` (literal on the right) is an overload ambiguity.** A member
   `operator>(const char*)` and a free `operator>(basic_string&, const char*)`
   both match, so the program does not even compile:
   `use of overloaded operator '>' is ambiguous`.

The same applies to `operator<`.

## Reproducers

```cpp
#include <string>
int main() {
  std::string a = "AAAA", b = "AA";
  __ESBMC_assert(!("AAAA" > b), "literal on left"); // nondet -> VERIFICATION SUCCESSFUL (wrong)
}
```

```cpp
#include <string>
int main() {
  std::string a = "AAAA";
  return a > "AA";        // error: use of overloaded operator '>' is ambiguous
}
```

## Root cause / fix

`operator==` / `operator!=` avoid both problems by being pure non-member free
templates — no member overloads, no `friend` mismatch. `operator>` / `operator<`
should follow the same pattern: drop the member `const char*` overloads and the
`friend` declarations, leaving the four `const char*` overloads as free templates
that access the (public) buffer directly.

## Relationship to #6227

#6227 fixes the *length-comparison* logic in these same operators. This issue is
about their *declaration/overload-resolution*, which is orthogonal but touches the
same functions, so the fix builds on that one.

Found while fixing #6199 / #6227.

lucasccordeiro:
## Root cause

`std::string`'s `operator>`/`operator<` had `const char*` overloads declared two
ways in `src/cpp/library/string`:

1. **Literal on the left** (`"lit" > s`): declared as an in-class `friend`
   (`friend bool operator>(const char *lhs, basic_string&rhs);`), a
   **non-template** function. The out-of-line definition, however, is a
   separate function **template**. The friend is therefore never actually
   defined, so ESBMC models the call as an unconstrained (nondet) function —
   an unsoundness bug: a genuinely-true assertion depending on the result can
   get a spurious `VERIFICATION FAILED`.
2. **Literal on the right** (`s > "lit"`): a member `operator>(const char*)`
   and the free `operator>(basic_string&, const char*)` template both match,
   so the call doesn't even compile (`use of overloaded operator '>' is
   ambiguous`).

Both issues equally affect `operator<`. #6228 fixed the *length-comparison*
logic in the same operators but explicitly left this declaration/overload
problem for a follow-up (this issue).

## Solution

Drop the member `operator>(const char*)`/`operator<(const char*)` declarations
and all four `friend` declarations, along with their now-orphaned member
definitions. This leaves only the free function templates, which already
access the class's public `str`/`_size` members directly — the same pattern
`operator==`/`operator!=` already use with no friendship required.

## Validation

- Reproduced both bugs pre-fix: the literal-on-left case gives `WARNING: no
  body for function allocator>#C#` and a spurious `VERIFICATION FAILED` on a
  consistency check (`("AAAA" > b) == ("AAAA" > b)` gave a counterexample with
  the two calls disagreeing); the literal-on-right case gives `ERROR: PARSING
  ERROR` / `use of overloaded operator '>' is ambiguous`, matching the exact
  clang diagnostic in the issue.
- New regression tests, both verified to fail before this change and pass
  after:
  - `regression/esbmc-cpp/cpp/github_6229` — literal-on-either-side
    comparisons are deterministic and give the correct result
    (`VERIFICATION SUCCESSFUL`).
  - `regression/esbmc-cpp/cpp/github_6229_fail` — `s > "literal"` now compiles
    and a genuinely-false comparison is correctly reported
    (`VERIFICATION FAILED`); previously this hit the parsing ambiguity error.
- Both new tests agree between Bitwuzla and Z3.
- No regressions: full `esbmc-cpp/string` suite (236 tests) and
  `esbmc-cpp/cpp` suite (582 tests) pass, aside from failures already present
  on master with this change reverted (macOS-local environment quirks:
  `github_2242_1/2`, `github_3897_collision`, `utility2_fail`, and an
  assertion-pretty-print formatting mismatch shared by `github_6200_fail`/
  `github_6227_fail`) — all confirmed unrelated to this change.
- `esbmc-cpp11/14/17/20` suites (228 tests) pass, aside from the two
  pre-existing macOS-local `builtin-template{,_fail}` failures.

Fixes #6229

### DIFF
--- a/regression/esbmc-cpp/cpp/github_6229/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6229/main.cpp
@@ -1,0 +1,28 @@
+#include <cassert>
+#include <string>
+
+int main()
+{
+  std::string a = "AAAA", b = "AA";
+
+  // Literal-on-the-left overloads must be deterministic: the in-class friend
+  // declared a non-template function that the out-of-line template definition
+  // never matched, so ESBMC modelled the call as an undefined (nondet) function.
+  bool lt1 = ("AAAA" > b);
+  bool lt2 = ("AAAA" > b);
+  assert(lt1 == lt2);
+  assert("AAAA" > b);
+  assert(!("AA" > a));
+  assert(!("AAAA" < b));
+  assert("AA" < a);
+
+  // Literal-on-the-right overloads must even compile: a member
+  // operator>(const char*) and a free operator>(basic_string&, const char*)
+  // both matched, making every such call ambiguous.
+  assert(a > "AA");
+  assert(!(b > "AAAA"));
+  assert(!(a < "AA"));
+  assert(b < "AAAA");
+
+  return 0;
+}
--- a/regression/esbmc-cpp/cpp/github_6229/test.desc
+++ b/regression/esbmc-cpp/cpp/github_6229/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--unwind 12
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/github_6229_fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6229_fail/main.cpp
@@ -1,0 +1,15 @@
+#include <cassert>
+#include <string>
+
+int main()
+{
+  std::string a = "AA";
+
+  // Before the fix, "s > literal" did not even compile (ambiguous between the
+  // member operator>(const char*) and the free operator>(basic_string&, const
+  // char*)). With the fix only the free template survives, and since "AA" is
+  // not greater than "AAAA" this genuinely-false assertion is correctly FAILED.
+  assert(a > "AAAA");
+
+  return 0;
+}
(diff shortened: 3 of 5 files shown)


lucasccordeiro:
Closing as a duplicate — #6230 fixes the same root cause (member/friend const char* overloads colliding with the free templates) and additionally adds the missing literal-on-left `>=`/`<=` overloads that this PR didn't cover. Thanks for the more complete fix.

PR #6253, opened 2026-07-22
[om] Wire std::unexpected into dynamic exception spec checks

### LINKED ISSUE #6022: [C++] Dynamic-exception-spec / std::unexpected / std::terminate recovery is incomplete
## Summary

The `--lower-exceptions` pass cannot yet soundly model the `std::unexpected` -> replacement-throw / `std::bad_exception` substitution / `std::terminate` recovery chain for dynamic exception specifications.

## Concrete latent defect

`src/goto-programs/remove_exceptions.cpp:make_unexpected_call()` looks up `c:@F@__ESBMC_unexpected`, but the operational-model function is named `c:@F@__ESBMC_run_unexpected` (name drift between #5276, which added the lookup, and #5326, which named the OM helper). The lookup therefore always returns nil, and the unexpected-handler dispatch block in `build_dynamic_spec_check` is **dead code** — a disallowed escape from a `throw(T...)` spec goes straight to a spec-violation assertion without ever running the installed handler.

## Why fixing the symbol name alone is unsound (verified)

Correcting the symbol flips `lower-exceptions_uncaught_count_dynspec` to SUCCESSFUL (the handler throws a permitted type and recovery succeeds), but it **regresses two CORE tests**:

- `try-catch_terminate_01_bug` (want FAILED -> gets SUCCESSFUL): the handler does a bare `throw;`. Because the program contains bare throws, the handled-stack OM is linked, so the bare throw routes through `__ESBMC_rethrow_current` -> empty handled stack -> `std::terminate` -> the custom `set_terminate` handler -> `abort()` (modeled as `assume(0)`) -> the path is **silently pruned** and the violation vanishes. This is the exact false-negative that `emit_terminate`'s own comment warns about.
- `try-catch_unexpected_04` (want FAILED -> gets SUCCESSFUL).

## What a sound fix needs

1. Wire `make_unexpected_call` to `__ESBMC_run_unexpected` (and construct the call as `symbol_exprt(id, type)`, not `h->get_value()`).
2. Model the `bad_exception` substitution, gated on "a custom handler threw a disallowed type" (so it does not over-fire when there is no custom handler).
3. Ensure lowering-synthesized terminate points remain reported as FAILED regardless of a custom `set_terminate` handler (do not let a handler that ends the path swallow the violation).

## KNOWNBUG tests

- `regression/esbmc-cpp/try_catch/{lower-exceptions_uncaught_count_dynspec, lower-exceptions_bad_exception_subst, try-catch_unexpected_01}`

(`try-catch_unexpected_01` additionally parse-fails under the default `-std c++17`, which dropped `set_unexpected`/dynamic specs; it would need `--std c++14` first.)

lucasccordeiro:
## Root cause

`src/goto-programs/remove_exceptions.cpp`'s `make_unexpected_call()` looked up the
wrong OM symbol for the installed `std::unexpected` handler
(`c:@F@__ESBMC_unexpected` instead of the real `c:@F@__ESBMC_run_unexpected`), so
the lookup always failed and the handler dispatch in `build_dynamic_spec_check`
was dead code: every disallowed escape from a `throw(T...)` spec went straight
to an unconditional "exception specification violated" assertion, whether or
not a `set_unexpected` handler was installed and would have recovered.

Fixing just the symbol regresses `try-catch_terminate_01_bug` and
`try-catch_unexpected_04`, because a sound fix needed two more pieces:

1. **A bare `throw;` inside a custom unexpected handler must re-raise the
   exception whose spec it is resolving** ([except.unexpected]) — but
   `__ESBMC_rethrow_current_exception()` always called `std::terminate()`
   when the handled-exception stack was empty, which it always is here (no
   catch has entered a handler for this exception yet — it's still
   propagating past the spec boundary). It now falls back to the still-set
   `typeid`/`value` globals (deliberately preserved across the handler call
   for exactly this case, per an existing comment), mirroring the same
   fallback `__ESBMC_current_exception_raw` already uses for the identical
   "in flight but not yet on the handled stack" situation.
2. **`std::bad_exception` substitution.** When the handler itself throws a
   type the spec disallows, [except.unexpected] says: if the spec includes
   `std::bad_exception`, the escaping exception is *replaced* by one and
   propagation continues from the violated function's call site; otherwise
   the spec is genuinely violated. New helper `bad_exception_throw()` mirrors
   the existing `build_bad_cast_throw()` pattern (lazy, cached, resolves
   `tag-std::bad_exception`/`tag-class std::bad_exception`).

## Validation

- Ground-truthed every changed-verdict scenario against real
  `clang++ -std=c++03` before touching test.desc files — not just standard
  reading. Confirmed `try-catch_unexpected_04`'s existing `test.desc`
  (`CORE`, expected `VERIFICATION FAILED`) was itself wrong: the program is
  well-defined (`myfunction` throws `char`, violates `throw(int)`,
  `myunexpected()` substitutes `throw 5` which *is* permitted, caught
  normally, prints `unexpected called`/`caught int`, exits 0) — corrected to
  `VERIFICATION SUCCESSFUL`.
- Promoted 3 `KNOWNBUG` tests to `CORE` now that they pass:
  `lower-exceptions_uncaught_count_dynspec`, `lower-exceptions_bad_exception_subst`
  (the exact bad_exception-substitution scenario), and
  `try-catch_unexpected_01` (also needed `--std c++03` added — it previously
  failed to even parse under the default C++17, unrelated to this fix).
- `try-catch_terminate_01_bug` still correctly reports `VERIFICATION FAILED`,
  now for the *right* reason: its first try block (spec includes
  `bad_exception`) correctly recovers via substitution; its second try block
  (spec does not include `bad_exception`) correctly hits the
  "exception specification violated" property.
- Dual-solver agreement (Bitwuzla + Z3) confirmed on every changed-verdict
  test.
- **Mode C dead-code proof** (via the `esbmc-verifier` agent, using
  `__ESBMC_unreachable()` + `--enable-unreachability-intrinsic`): all 3
  new/restructured branches confirmed reachable — the
  `__ESBMC_rethrow_current_exception` fallback, and both nesting levels of
  the bad_exception-substitution check in `build_dynamic_spec_check`.
- No regressions: `esbmc-cpp/try_catch` (165/165 passing),
  `esbmc-cpp/cpp`, `esbmc-cpp11/14/17/20`, and
  `destructors`/`inheritance`/`polymorphism_bringup`/`bug_fixes`/`cbmc`
  suites all pass, aside from failures already present on master
  (macOS-local environment quirks unrelated to this change).

Fixes #6022

### DIFF
--- a/regression/esbmc-cpp/try_catch/lower-exceptions_bad_exception_subst/test.desc
+++ b/regression/esbmc-cpp/try_catch/lower-exceptions_bad_exception_subst/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 4 --no-unwinding-assertions --std c++03
 
--- a/regression/esbmc-cpp/try_catch/lower-exceptions_uncaught_count_dynspec/test.desc
+++ b/regression/esbmc-cpp/try_catch/lower-exceptions_uncaught_count_dynspec/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.cpp
 --unwind 4 --no-unwinding-assertions --std c++11
 
--- a/regression/esbmc-cpp/try_catch/try-catch_unexpected_01/test.desc
+++ b/regression/esbmc-cpp/try_catch/try-catch_unexpected_01/test.desc
@@ -1,5 +1,5 @@
-KNOWNBUG
+CORE
 main.cpp
---unwind 10 --no-unwinding-assertions --timeout 900
+--unwind 10 --no-unwinding-assertions --timeout 900 --std c++03
 
 ^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/try_catch/try-catch_unexpected_04/main.cpp
+++ b/regression/esbmc-cpp/try_catch/try-catch_unexpected_04/main.cpp
@@ -1,3 +1,8 @@
+// myfunction's throw(int) spec is violated by `throw 'x'` (char), which runs
+// myunexpected(); it throws 5, an int, which the spec permits, so it
+// propagates and is caught by catch(int). Ground-truthed against real
+// clang++ -std=c++03: prints "unexpected called" / "caught int", exits 0 --
+// a well-defined recovery, not a violation (github #6022).
 #include <iostream>
 #include <exception>
 using namespace std;
--- a/regression/esbmc-cpp/try_catch/try-catch_unexpected_04/test.desc
+++ b/regression/esbmc-cpp/try_catch/try-catch_unexpected_04/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.cpp
 --unwind 10 --no-unwinding-assertions --timeout 900 --std c++03
-^VERIFICATION FAILED$
+^VERIFICATION SUCCESSFUL$
--- a/src/c2goto/library/cpp/exception.cpp
+++ b/src/c2goto/library/cpp/exception.cpp
@@ -188,9 +188,30 @@ extern "C" void __ESBMC_rethrow_current_exception()
 __ESBMC_HIDE:;
   if (__ESBMC_exception_detail::__ESBMC_handled_exception_depth == 0)
   {
-    __ESBMC_exc_terminate_reason =
-      __ESBMC_exception_detail::__ESBMC_terminate_reason_no_active;
-    std::terminate();
+    // No catch has entered a handler for this exception yet, but one may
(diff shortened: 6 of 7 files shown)


lucasccordeiro:
Closing as a duplicate — #6240 (merged as 1efee1ea8b) fixes the same root cause with a more complete solution: it also makes the default `std::unexpected` handler a no-op so a spec violation with no installed handler can't be silently swallowed by a custom `std::set_terminate` that ends the path (e.g. `abort()`), which this PR didn't address. Thanks for beating me to it.

Issue #4397, opened 2026-05-09
[esbmc-cpp] C++ language-feature regressions marked KNOWNBUG


lucasccordeiro:
Tracking issue for **C++ language-feature** regression tests in `regression/esbmc-cpp/cpp/`, `regression/esbmc-cpp/cbmc/`, and `regression/esbmc-cpp/unix/` that are currently marked `KNOWNBUG` — ESBMC does not produce the expected verdict on them. Each item below is a separate failure; click through to the source on master for the full reproducer.

## Failing tests

- [ ] [`cbmc/Templates6`](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-cpp/cbmc/Templates6) — int g; — expects `VERIFICATION SUCCESSFUL` — flags: `--unwind 5 --no-unwinding-assertions --timeout 900`
- [ ] [`cpp/bicycle`](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-cpp/cpp/bicycle) — Name        : lista5c.cpp — expects `VERIFICATION FAILED` — flags: `Bicycle.cpp Display.cpp EmbeddedPC.cpp Thread.cpp --unwind 3 --no-unwinding-assertions --timeout 900`
- [ ] [`cpp/ch10_4`](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-cpp/cpp/ch10_4) — Fig. 10.10: fig10_10.cpp — expects `VERIFICATION SUCCESSFUL` — flags: `point.cpp circle.cpp --unwind 10 --no-unwinding-assertions --timeout 900`
- [ ] [`cpp/ch13_6`](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-cpp/cpp/ch13_6) — Fig. 13.6: fig13_06.cpp — expects `VERIFICATION FAILED` — flags: `--unwind 5 --no-unwinding-assertions --timeout 900`
- [ ] [`cpp/ch16_3`](https://github.com/esbmc/esbmc/tree/master/regression/esbmc-cpp/cpp/ch16_3) — Fig. 16.9: querystring.cpp — expects `VERIFICATION FAILED` — flags: `--unwind 10 --no-unwinding-assertions --timeout 900`
[...]

PR #6268, opened 2026-07-22
[test] Un-KNOWNBUG the two list::sort regressions

lucasccordeiro:
Both `list/list_sort_bug-2` and `template/list_sort_bug` ran at `--unwind 5` with `--no-unwinding-assertions`, one iteration short of the 6-iteration `strcpy` of the literal `"Three"` in the string OM. `loop_bound_exceeded` then assumes the negated loop guard, which is false for a concrete loop, so the path died and every assertion after `sort()` was vacuously true — reported as `VERIFICATION SUCCESSFUL`, indistinguishable from a proof.

At `--unwind 6` both detect their intended assertion (`main.cpp:33` and `main.cpp:44`), so the expected violated-property location is pinned alongside the verdict; a future vacuous run can no longer pass on the verdict line alone. They run ~135s, so they join the THOROUGH tier alongside the sibling `list_sort*` tests rather than risking the CORE time cap.

Refs #4397

### DIFF
--- a/regression/esbmc-cpp/list/list_sort_bug-2/test.desc
+++ b/regression/esbmc-cpp/list/list_sort_bug-2/test.desc
@@ -1,4 +1,5 @@
-KNOWNBUG
+THOROUGH
 main.cpp
---unwind 5 --no-unwinding-assertions --timeout 900
+--unwind 6 --no-unwinding-assertions --timeout 900
 ^VERIFICATION FAILED$
+^  file .*main\.cpp line 44 column 3 function main$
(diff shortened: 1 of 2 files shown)


lucasccordeiro:
Withdrawing this. The two tests exercise a non-bug (the `list<string>::sort` model is correct — the original KNOWNBUG was a vacuity artifact: at `--unwind 5 --no-unwinding-assertions` the 6-iteration `strcpy` of the literal "Three" truncated the path, so every post-sort assertion held vacuously). At the correct `--unwind 6` the tests detect the intended assertion but take ~126-135s, and the DebugOpt CI hard-caps every test at 120s (`ESBMC_REGRESS_TIMEOUT=120`, overriding per-test timeouts) — even with all safety checks disabled it stays at 126s, so no config fits. As KNOWNBUG the timeout is tolerated; promoting to CORE/THOROUGH turned it into a hard CI failure for no real-bug coverage. Leaving both as KNOWNBUG.

Issue #5955, opened 2026-07-10
[python] consteval seed_globals is mutation-blind: stale container seeds fold asserts to false proofs


lucasccordeiro:
## Symptom

The module-level assert folder (`try_eval_global_expr` → `seed_globals`, used from `converter_stmt.cpp`) seeds a global with its assigned literal when the name has exactly one *assignment statement* — but `seed_globals` counts only `Assign`/`AnnAssign`/`AugAssign` targets, so in-place mutations (`append`, `insert`, `remove`, `l[i] = x`) are invisible and the stale pre-mutation value is folded. Both directions confirmed against CPython 3; the false-proof direction:

**Witness 1 — stale receiver** (CPython: AssertionError; ESBMC: VERIFICATION SUCCESSFUL):
```python
l = [1, 2]
l.append(3)
assert l.count(3) == 0
```

**Witness 2 — stale operand of an otherwise-literal fold** (CPython: AssertionError; ESBMC: VERIFICATION SUCCESSFUL):
```python
x = [1]
x.append(2)
assert [[1], [1, 2]].index(x) == 0
```

Both at `--unwind 6`, ESBMC master + PR #5954 (aarch64 macOS, Bitwuzla default). The environment is also scope-blind: a function-local list shadowing a module global resolves to the global's value.

## Root cause

`python_consteval::seed_globals` (src/python-frontend/python_consteval.cpp) uses a write-once heuristic over assignment statements only. That is sound for immutable scalars but unsound for mutable containers, whose value can change without any assignment to the name.

## Suggested hardening

Do not seed LIST/TUPLE-valued globals whose name appears anywhere in the module as an attribute-call receiver or subscript-store target — or, simpler, never seed container values. Note PR #5954 already gates the list-literal `count()/index()` fold on an empty environment, so that path is unaffected; this issue covers the remaining `try_eval_global_expr` users (the whole-assert folder in particular).

PR #6273, opened 2026-07-23
[python] Fix core dump when calling a list method on a str (#6264)

### LINKED ISSUE #6264: [python] Calling a list method (.append) on a str crashes (core dump) instead of AttributeError
### Summary

Calling a list method such as `.append()` on a **str** crashes ESBMC with a core dump. The frontend routes `str.append(...)` to the list operational model (`__ESBMC_list_push`), producing a type mismatch (array vs pointer) that aborts, instead of raising the `AttributeError` that Python would.

### Reproducer (self-contained)
```python
def main() -> None:
    s = "abc"
    s.append("d")
main()
```

### Command
```sh
esbmc repro.py --incremental-bmc
```

### Actual
```
Converting
ERROR: function call: argument "c:list.c@...@F@__ESBMC_list_push@l" type mismatch: got array, expected pointer
<aborted / core dumped> (exit 134)
```

### Ground truth (CPython)
`"abc".append("d")` → `AttributeError: 'str' object has no attribute 'append'`.

### Expected
A verifier must not core-dump on parseable input. It should raise a catchable `AttributeError` (uncaught → `VERIFICATION FAILED`), matching the exception-safety (`dynamic-typing`) property family.

### Root cause (initial)
The method-call handler routes `.append()` to the list model without first checking that the receiver is actually a list; a `str` receiver (a char array) is passed where `__ESBMC_list_push` expects a list pointer.

### Notes

LukeW1999:
## Summary
Fixes #6264 — calling a list method (e.g. `.append`) on a **str** crashed ESBMC with a core dump.

`is_list_method_call()` guards `count`/`index`/`pop`/`copy`/`clear` with a receiver-is-list check, but `append`/`insert`/`remove`/`extend`/`sort`/`reverse`/`appendleft`/`popleft` fell through to `return true` with **no** receiver check. A `str` receiver was therefore routed into the list model (`__ESBMC_list_push`), passing a char array where a list pointer is expected → `migrate expr` type mismatch → `abort()`.

## Fix
Add a **targeted** str-receiver exclusion just before the final `return true`: if the receiver is a string literal or a Name whose symbol is string-typed, return false so dispatch falls through and the general handler raises the correct `AttributeError`. Non-str receivers are unaffected, so the hot `list.append` path (local / `self.attr` / nested / parameter receivers) is untouched.

## Effect
- `"abc".append("d")` → clean `AttributeError` (uncaught → `VERIFICATION FAILED`) instead of a core dump. Matches CPython (`'str' object has no attribute 'append'`) and the exception-safety (`dynamic-typing`) property family.
- `list.append` in all forms continues to verify.

## Tests
- `regression/python/github_6264_fail` — `str.append` → `VERIFICATION FAILED` (AttributeError, no crash).
- `regression/python/github_6264` — regression guard: `list.append` on local, `self.attribute`, and nested-list receivers → `VERIFICATION SUCCESSFUL`.

The python list/string/dict/set/tuple/comprehension regression subset passes with no regressions.


lucasccordeiro (commented):
Actionable review comments (findings reproduced on the pre-PR binary). int/dict receivers already return a clean `VERIFICATION FAILED`; the crash is specific to char-array types (str/bytes), which sharpens where the fix needs to be complete.

**`src/python-frontend/function_call/expr.cpp` — incomplete: str via Attribute/Subscript receivers still crash.** The new exclusion only handles `_type == "Constant"` and `"Name"`. A str reached through an **Attribute** or **Subscript** receiver still falls through to `return true` and routes into the list model. Reproduced:
- `self.s.append("d")` (str attribute) → `ERROR: List variable not found: self.s`
- `xs[0].append("d")` (list-of-str) → `List variable not found: str`

Both are the #6264 failure mode and are not fixed. Note that because Attribute receivers are never inspected, `self.items.append` (list) and `self.s.append` (str) both reach `return true` by falling through — the block cannot tell them apart, which is the bug. Add these shapes plus a `_fail` test for each.

**`src/python-frontend/function_call/expr.cpp` — `bytes` receiver crashes identically and is uncovered.** `b = b"abc"; b.append(1)` reproduces the exact target crash:
```
ERROR: function call: argument "...__ESBMC_list_push@l" type mismatch: got array, expected pointer
```
`type_utils::is_string_type` does not match `bytes`, so a bytes `Name` receiver isn't excluded. Broaden the exclusion (or use a positive list check) and add a bytes `_fail` test.

**`src/python-frontend/function_call/expr.cpp` — architectural asymmetry; the catch-all `return true` is the real defect.** Every other shared-name guard here (`count`/`index`/`pop`/`copy`/`clear`) claims *positively* — true only when the receiver resolves to `list_type` (plus the `List`/`BinOp` literal shortcuts). This block instead rejects one *non*-list type (str) and lets everything else through. Prefer mirroring the `count`/`index` positive check so all non-list receivers uniformly fall through to the correct `AttributeError`. If a negative check was deliberate to keep *unresolved* list receivers working (e.g. `get_list().append()`, comprehension results, chained calls) — which a positive `get_object_list_symbol` check would drop to `false` — state that in the comment and add a guarding test, since it's the only thing justifying the asymmetry.

**`src/python-frontend/function_call/expr.cpp` — inconsistent symbol lookup.** The Name branch uses `lookup_python_symbol(recv["id"])`, whereas every other branch resolves the receiver via `get_object_list_symbol(...)`. Using a different resolver risks divergent scope/shadowing resolution for the same receiver. Reuse `get_object_list_symbol` (it also already handles attribute receivers via its `$attr_list$` temp, which would cover the Attribute case above).

**`regression/python/github_6264_fail/test.desc` — failure reason not pinned.** The test asserts only `^VERIFICATION FAILED$`. A regressed crash prints no `VERIFICATION FAILED` line, so a crash-regression is caught — but a plain `FAILED` also passes for any unrelated reason. Pin the AttributeError / `dynamic-typing` property (regex the emitted message) so the test asserts *why* it fails.

**Missing coverage.** The negative suite should add: `self.<str>.append`, `<list_of_str>[i].append`, and a `bytes` receiver — each still fails today.

**Minor — anonymous `{ ... }` scope** around the two checks is stylistically inconsistent with the surrounding guard-clause / early-return pattern; fold it in (or into the positive-check refactor).

LukeW1999:
Thanks @lucasccordeiro — all correct. Reworked to the positive check you recommended (pushed):

- **Positive claim (findings #3, #1, #2):** `append`/`insert`/`remove`/`extend`/`sort`/`reverse`/`appendleft`/`popleft` are now claimed only when the receiver resolves to a list, mirroring `count`/`index`/`pop`/`copy`/`clear`. Non-list receivers fall through to the general handler's `AttributeError`. This fixes the previously-uncovered shapes:
  - `str` via **Name** → `uncaught exception: AttributeError`
  - `str` via **Subscript** (`list_of_str[i].append`) → `VERIFICATION FAILED` (`Unsupported function 'append' is reached`)
  - **`bytes`** receiver → `uncaught exception: AttributeError`
- **Finding #4 (consistent resolver):** now uses `get_object_list_symbol` (not `lookup_python_symbol`), so Name/Subscript/Attribute/Call receivers all resolve through the same path; the `List`/`BinOp`/`ListComp` literal shortcut is kept for symbol-less list temporaries.
- **Finding #5 (pin the reason):** `github_6264_fail` now regexes `uncaught exception: AttributeError`; added `github_6264_bytes_fail` and `github_6264_subscript_str_fail` (each pins its diagnostic). `github_6264` still guards the valid `list.append` forms (local / `self.attr` / nested).
- **Validation:** the valid `list.append` forms verify; the 1082-test `list|string|str|append|dict|set|tuple|comprehension|deque|collection|extend|insert|sort|reverse|pop` subset passes 100% (no regression from switching to the positive check).

**One case is out of scope here:** `self.s.append(...)` where `s` is a **str attribute**. With the positive check it no longer routes into the list model, but it then hits the *enclosing-class-fallback* `abort()` (`Function @C@<Class>@F@append not found`) — i.e. finding **#9** from the #6242 review, the shared root, not the list-model crash this PR targets. It crashes identically before and after this PR (no regression). I've left it for the dedicated #9 follow-up (turning that `abort()` into a recoverable diagnostic) rather than widening this PR into the attribute-chain resolver.

lucasccordeiro:
@LukeW1999: Could you please check the CI of this PR?

LukeW1999:
Thanks — this all landed. I pushed a rework (4933aac9b) right after your review that adopts exactly the architecture you recommended, so the negative str-exclusion is gone:

**Positive check (findings #3 + #4).** For `append`/`insert`/`remove`/`extend`/`sort`/`reverse`/`appendleft`/`popleft` the block now claims the call *only* when the receiver resolves to a list — mirroring the `count`/`index`/`pop`/`copy`/`clear` guards — via `get_object_list_symbol` (plus the `List`/`BinOp`/`ListComp` literal shortcut). No more catch-all `return true`; every non-list receiver uniformly falls through to the `AttributeError` path. This also removes the inconsistent `lookup_python_symbol` resolver — the whole block now uses `get_object_list_symbol`, which handles Name, Subscript, Attribute (`$attr_list$` temp) and Call (`$call_list$` temp) receivers and returns non-null only for `list_type`.

**Coverage (findings #1, #2, #5, #6):**
- `github_6264_fail` — str receiver → now pins `uncaught exception: AttributeError` (failure reason asserted, not just `FAILED`).
- `github_6264_bytes_fail` — `b"abc".append(1)` → AttributeError, `FAILED`.
- `github_6264_subscript_str_fail` — list-of-str `xs[0].append(...)` → `Unsupported function 'append' is reached`, `FAILED`.

**One residual, out of scope here:** the str-**attribute** receiver `self.s.append(...)` no longer routes into the list model (the `List variable not found: self.s` you cited is gone), but it now trips a *different*, pre-existing abort in the method-call path: GOTO generation looks up a method `append` on the class, fails, and aborts with `ERROR: Function '...@C@F@append' not found` (core dump) instead of raising `AttributeError`. That's a separate defect in unresolved-attribute-method dispatch, independent of the #6264 char-array crash this PR targets. I'd rather fix it in a dedicated follow-up (it needs the method-not-found path to raise AttributeError generally, not just for str) than widen scope here — I'll file it separately.

LukeW1999:
Correction + fix pushed.

**I have to correct an earlier claim.** In my previous reply I wrote that the positive-check rework passed a validation subset "100% (no regression)". That was wrong: my subset was filtered by the keywords `list|string|str|append|...`, which did **not** match `github_5955_alias_builtin_arg_fail`, and CI caught exactly that test flipping `VERIFICATION FAILED` → `SUCCESSFUL`. I've reproduced it locally on the branch binary. Apologies for the misleading "no regression".

**Root cause — a pure positive check is unsound for the mutators.** `github_5955` is `m = min([l]); m.append(3)`. `min()` is annotated `Any` and lowers `m` to a plain `int` (`signedbv`), even though it returns the list `l` itself. A `resolves-to-list_type` check therefore drops `m.append`, the append is never modelled, and `l.count(3) == 0` wrongly holds. This is exactly the "unresolved/mis-typed list receiver" case you flagged — a pure positive check can't keep it.

**Fix (pushed).** For the list-only mutators I now claim the call *unless the receiver resolves to a character-array type*. The #6264 crash is specifically a str/bytes char array reaching `__ESBMC_list_push` where a `PyListObject*` is expected ("got array, expected pointer"); a real list is a pointer, and a loosely-typed receiver like `m` above is `int` — neither is an array, so both route correctly, while str/bytes are excluded. Because a str/bytes reached through a Name, an attribute, or a subscript *all* resolve to an array type, this uniformly covers your three cases without a shape-by-shape special case:

- `github_6264` (local / `self.attr` / nested list.append) → `SUCCESSFUL`
- `github_6264_fail` (`s = "abc"; s.append`) → `AttributeError`, `FAILED`
- `github_6264_bytes_fail` (`b"abc".append`) → `AttributeError`, `FAILED`
- `github_6264_subscript_str_fail` (`xs[0].append`, list-of-str) → `Unsupported function 'append'`, `FAILED`
- `github_5955_alias_builtin_arg_fail` (`min([l]).append`) → `FAILED` (regression fixed)
- new `github_6264_dynamic_recv` → guards the mis-typed-list receiver so a future pure-positive refactor can't silently re-drop it.

So the direction you asked for — don't blanket-`return true`, exclude the receiver that actually crashes — stands; the exclusion is on the *crashing type* (char array) rather than a pure list-only positive, which is the only form that keeps `#5955` working. I documented this trade-off in the block comment and validated the full 1120-test python regression this time (not a keyword subset).

**Residual (unchanged, out of scope):** `self.s.append` (str *attribute*) still trips the pre-existing method-not-found abort in attribute-method dispatch rather than a clean `AttributeError`; that's the separate #6242-review #9 follow-up I'll file, independent of this char-array fix.

PR #6276, opened 2026-07-23
[python] Follow-up tests + documented limitations for #6242 (review response)

LukeW1999:
## Summary
Follow-up to @lucasccordeiro's review on #6244. Findings #1 (cyclic-base `visited` guard), #7 (dedup into `method_return_type`), #5 (inherited test), and #3 (dropped the two unrelated loop-invariant commits) were already addressed by the polish commit that merged with #6244. This PR closes the remaining **test/documentation** gaps and records the limitations honestly.

## Changes (tests only)
- **`github_6242_attribute_error`** (finding #4): the issue's *original* reproducer — an attribute assigned from a method whose return type is not the enclosing class, then a method call on it. This used to core-dump; it now degrades to `Unsupported function 'publish' is reached` / `VERIFICATION FAILED`. The `test.desc` pins the actual diagnostic, not a bare verdict.
- **`github_6242_override`** (finding #2, KNOWNBUG): documents that `self.attr = self.method()` is typed by the enclosing class and ignores subclass overrides (no virtual dispatch), so a valid polymorphic program is reported as a false `VERIFICATION FAILED`. Marked KNOWNBUG so the limitation is explicit, not silent.
- **`github_6242_fail`** (finding #6): drops the `--no-unwinding-assertions`/`--unwind 1` asymmetry — the loop-free program only needs `--incremental-bmc`, matching `github_6242`.

## On the remaining findings
- **#2 (overridden methods):** I tried the suggested "fall back to `Any`", but an `Any`-typed attribute re-enters the `self.attr.method()` path and hits the still-present `abort()` at `builtin_functions.cpp` (finding #9) — i.e. it turns the false positive into a *crash*, which is worse. So the honest fix today is the KNOWNBUG above; a real fix needs virtual dispatch or the abort→recoverable-diagnostic change.
- **#8 (`async def`):** recognizing `AsyncFunctionDef` in the annotator is not sufficient — the async method call itself still aborts (coroutine handling is unmodelled). Left as a separate limitation rather than a crashing KNOWNBUG test.
- **#9 (`Fixes` overclaim / reachable `abort()`):** acknowledged; the abort at `builtin_functions.cpp:632` is the shared root and remains the right place for a follow-up.

Thanks for the thorough, executed-not-inferred review.

### DIFF
--- a/regression/python/github_6242_fail/test.desc
+++ b/regression/python/github_6242_fail/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.py
---incremental-bmc
+--unwind 1
 ^VERIFICATION FAILED$
--- a/regression/python/github_6242_override/test.desc
+++ b/regression/python/github_6242_override/test.desc
@@ -1,0 +1,4 @@
+KNOWNBUG
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_6242_override/test.desc
+++ b/regression/python/github_6242_override/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.py
---incremental-bmc
-^VERIFICATION SUCCESSFUL$
+--unwind 1
+^VERIFICATION FAILED$
--- a/regression/python/github_6242_unsupported_method_fail/main.py
+++ b/regression/python/github_6242_unsupported_method_fail/main.py
@@ -1,7 +1,9 @@
 # github.com/esbmc/esbmc/issues/6242 (review finding #4)
 # The issue's original reproducer: an attribute assigned from a method whose
 # return type is NOT the enclosing class, then a method call on that attribute.
-# This used to core-dump (SIGABRT); it must now degrade to a clean verdict.
+# This used to core-dump (SIGABRT); it must now degrade to a clean
+# VERIFICATION FAILED (currently the generic "Unsupported function" diagnostic,
+# not a Python AttributeError — hence the dir name, not "attribute_error").
 class Talker:
     def make(self) -> int:
         return 5
--- a/regression/python/github_6242_attribute_error/main.py
+++ b/regression/python/github_6242_attribute_error/main.py
@@ -1,0 +1,16 @@
+# github.com/esbmc/esbmc/issues/6242 (review finding #4)
+# The issue's original reproducer: an attribute assigned from a method whose
+# return type is NOT the enclosing class, then a method call on that attribute.
+# This used to core-dump (SIGABRT); it must now degrade to a clean verdict.
+class Talker:
+    def make(self) -> int:
+        return 5
+
+    def __init__(self):
+        self.pub = self.make()
+
+    def go(self):
+        self.pub.publish(1)
+
+
+Talker().go()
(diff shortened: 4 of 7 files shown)


regression/python/github_6242_fail/test.desc
```diff
@@ -1,4 +1,4 @@
 CORE
 main.py
---no-unwinding-assertions --unwind 1
+--incremental-bmc
```
lucasccordeiro: Siblings use `--unwind 1`; this uses `--incremental-bmc`. Unify the family or fix the rationale.

lucasccordeiro (commented):
Verified all three tests against a freshly rebuilt master binary (harness OK); tests-only, no regressions. A few refinements:

regression/python/github_6242_unsupported_method_fail/main.py
```diff
@@ -0,0 +1,16 @@
...
+# This used to core-dump (SIGABRT); it must now degrade to a clean verdict.
+class Talker:
+    def make(self) -> int:
+        return 5
+
+    def __init__(self):
+        self.pub = self.make()
+
+    def go(self):
+        self.pub.publish(1)
```
lucasccordeiro, line 15: Pinned diagnostic is the generic 'Unsupported function', not AttributeError; the dir name misleads—rename.

regression/python/github_6242_override/test.desc
```diff
@@ -0,0 +1,4 @@
+KNOWNBUG
```
lucasccordeiro: KNOWNBUG passes on any non-SUCCESSFUL output, so a crash regression here stays undetected.

regression/python/github_6242_override/test.desc
```diff
@@ -0,0 +1,4 @@
...
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro: Unmodelled virtual dispatch is a missing capability—FUTURE may fit better than KNOWNBUG.

LukeW1999:
Thanks — all four addressed in the latest commit:

- **`github_6242_fail` flag** — unified the family on `--unwind 1` (the three tests I added were on `--incremental-bmc` while the passing siblings use `--unwind 1`). All still produce the expected verdicts.
- **`github_6242_attribute_error` rename** — renamed to `github_6242_unsupported_method_fail`; the pinned diagnostic really is the generic `Unsupported function 'publish' is reached`, not a Python `AttributeError`, so the old name misled. Comment updated to say so.
- **`github_6242_override` crash-blindness (#3) + FUTURE-vs-KNOWNBUG (#4)** — I dug into `testing_tool.py`: `FAIL_MODES = ["KNOWNBUG", "FUTURE"]` behave *identically* at runtime — both pass whenever the expected regex is absent, so **FUTURE is equally blind to a crash regression** (it would only change the documentary label). To actually fix #3 I converted it to a **CORE characterization test** pinning the current `VERIFICATION FAILED`. That way a crash regression is caught (the `FAILED` line must appear) and the test breaks loudly the day virtual dispatch lands (output flips to `SUCCESSFUL`). The comment documents that the *correct* result is `SUCCESSFUL` (CPython: `Derived().go() == 2`) and that this is a missing-capability xfail. If you'd rather keep it as an aspirational FUTURE xfail and accept the crash-blindness, I'm happy to switch — but CORE-FAILED seemed to better satisfy your #3.

PR #6279, opened 2026-07-23
CXL support

rafaelsamenezes:
This draft is the result of asking Qwen to generate this support


src/solvers/smt/smt_overflow.cpp
```diff
@@ -9,7 +9,13 @@ smt_astt smt_solver_baset::overflow_arith(const expr2tc &expr)
   expr2tc zero = gen_zero(side1->type);
 
   // Guess whether we're performing a signed or unsigned comparison.
-  bool is_signed = (is_signedbv_type(side1) || is_signedbv_type(side2));
+  // For a shift only the shifted operand's signedness matters: the count is a
+  // separate operand whose type says nothing about the range of the result.
+  // Including it sends `(u64)x << 32` -- whose count is a plain signed int --
+  // down the signed path, which reports overflow for any result >= 2^63.
+  bool is_signed = is_shl2t(overflow.operand)
```
lucasccordeiro, line 16: This is a verbatim copy of the fix in #6684. Drop it here and rebase once that lands, so the change has one owner and one test.

src/pointer-analysis/value_set.cpp
```diff
@@ -697,6 +697,10 @@ void value_sett::get_value_set_rec(
         if (is_empty_type(subtype))
         {
           total_offs = to_constant_int2t(non_ptr_op).value;
+          /* The sized path above negates for sub2t after scaling; this
+           * recovery path scales by one, so it must negate too. */
+          if (is_sub2t(expr))
```
lucasccordeiro, line 737: A real pointer-arithmetic fix with no test outside the opt-in CXL suite. Add a `regression/esbmc` test that fails without it.

regression/CMakeLists.txt
```diff
@@ -35,6 +35,15 @@ set(ESBMC_REGRESS_MEMORY_LIMIT 8192 CACHE STRING
 option(ENABLE_SANITIZER_CI
     "Restrict regression to CORE mode for sanitizer-instrumented CI runs" OFF)
 
+# The cxl suite is opt-in and off by default. Its tests are all THOROUGH:
```
lucasccordeiro, line 38: Registering 250 test files off by default puts them in the tree with no CI coverage to keep them honest. Land only the ones that run.

regression/cxl/cxl_aer_01/main.c
```diff
@@ -0,0 +1,129 @@
...
+  uint64_t aer_correctable_count;
+  uint64_t aer_non_fatal_count;
+  uint64_t aer_fatal_count;
+  int aer_enabled;
+};
+
+static struct pci_dev test_pci;
+
+/* Enable AER on the device */
+int pci_enable_aer(struct pci_dev *dev)
```
lucasccordeiro, line 31: The test defines the driver it verifies inside the same file, so it exercises no ESBMC or CXL-model behaviour. Verify the operational model instead.

lucasccordeiro (changes requested):

PR #6282, opened 2026-07-23
[python] Handle relative import with no module name (from . import X) (#6281)

### LINKED ISSUE #6281: [python] 'from . import X' (relative import with no module name) crashes with nlohmann type_error
### Summary

`from . import X` and `from .. import X` — a relative import with **no module name** — crash ESBMC with an uncaught `nlohmann::json` `type_error` (SIGABRT), during parsing/AST processing. For these forms the AST's `module` field is `None` (null), and the frontend reads it as a string.

Found by sampling a large corpus of real ROS2 Python packages: this pattern is ubiquitous in `__init__.py` / package code, and was the single most common non-import crash signature.

### Reproducer (self-contained, one line)
```python
from . import foo
```
(also `from .. import foo`)

### Command
```sh
esbmc repro.py --goto-functions-only
```

### Actual
```
  what():  [json.exception.type_error.302] type must be string, but is null
<aborted / core dumped> (exit 134)
```

### Ground truth (CPython)
`from . import foo` is a valid relative import (`ImportFrom(module=None, names=[alias('foo')], level=1)`).

### Expected
No core dump on parseable input. The relative import should resolve (or degrade to a clean "module not found") like the named form `from ..pkg import X`, which does **not** crash.

### Root cause (initial)
For `from . import X` the AST node has `module = None`; the import-resolution path accesses `module` as a string (`.get<std::string>()` / `operator[]`) without a null guard, so nlohmann throws `type_error.302`. Contrast: `from ..pkg import X` (module = "pkg") is handled cleanly.

### Notes
- Distinct from the `@C@...@F@... not found` abort family; this one aborts during AST processing before conversion.

LukeW1999:
## Summary
Fixes #6281 — `from . import X` / `from .. import X` (relative import with no module name) crashed ESBMC with an uncaught `nlohmann::json` `type_error` (SIGABRT) during AST processing.

Found by sampling a large corpus of real ROS2 Python packages: this pattern (ubiquitous in package `__init__.py` / module code) was the single most common non-import crash signature.

## Root cause
For `from . import X`, `ImportFrom.module` is `null`. Several sites read it as a `std::string`, so nlohmann threw `type_error.302` ("type must be string, but is null").

## Fix
- `python_converter.cpp`: replace the 5 duplicated `(_type=="ImportFrom") ? node["module"] : ...` ternaries with a null-safe `import_module_name()` helper (returns `""` for a null module). At the two import-open failure points, skip an empty-module (relative) import instead of throwing `Cannot open file`.
- `json_utils.h` `is_class()`: null-guard the `ImportFrom` module read.

A relative import with no module name is now treated as **unresolved** and the module still converts and verifies; a *named* relative import (`from ..pkg import X`) is unchanged (already resolved cleanly).

## Tests
- `regression/python/github_6281` — `from . import X` + verifiable body → `VERIFICATION SUCCESSFUL`.
- `regression/python/github_6281_fail` — `from .. import X` + a violated assertion → `VERIFICATION FAILED` (proves the module converts and its assertions are checked, rather than crashing at the import).

Validated against the python `import|module|class|from|package` regression subset (80 tests, 100% pass).

### DIFF
--- a/regression/python/github_6281_fail/test.desc
+++ b/regression/python/github_6281_fail/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.py
---no-unwinding-assertions --incremental-bmc
+--incremental-bmc
 ^VERIFICATION FAILED$
--- a/regression/python/github_6281_fail/test.desc
+++ b/regression/python/github_6281_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--no-unwinding-assertions --incremental-bmc
+^VERIFICATION FAILED$
--- a/regression/python/github_6281/main.py
+++ b/regression/python/github_6281/main.py
@@ -1,0 +1,15 @@
+# github.com/esbmc/esbmc/issues/6281
+# `from . import X` (relative import, no module name) used to crash ESBMC with an
+# uncaught nlohmann type_error. It must degrade to an unresolved import and let
+# the rest of the module verify.
+from . import helper
+
+
+def main() -> None:
+    total = 0
+    for i in range(5):
+        total += i
+    assert total == 10
+
+
+main()
--- a/regression/python/github_6281/test.desc
+++ b/regression/python/github_6281/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/python/github_6281_fail/main.py
+++ b/regression/python/github_6281_fail/main.py
@@ -1,0 +1,14 @@
+# github.com/esbmc/esbmc/issues/6281
+# Same relative import; the module must still convert and its assertions be
+# checked (here a wrong one), rather than crashing at the import.
+from .. import helper
+
+
+def main() -> None:
+    total = 0
+    for i in range(5):
+        total += i
+    assert total == 11
+
+
+main()
--- a/regression/python/github_6281_used/main.py
+++ b/regression/python/github_6281_used/main.py
@@ -1,0 +1,16 @@
+# github.com/esbmc/esbmc/issues/6281
+# Relative import whose name is *used* (called). This exercises the annotator's
+# get_function_return_type / wildcard-import paths, which read the ImportFrom
+# "module" field. For `from . import X` that field is null; the earlier fix only
+# guarded the converter, so these annotator sites still core-dumped with an
+# uncatchable nlohmann type_error. The name stays unresolved, so ESBMC must
+# degrade to a NameError instead of aborting.
+from . import helper
+
+
+def main() -> None:
+    x = helper()
+    assert x == 42
+
+
+main()
--- a/regression/python/github_6281_used/test.desc
+++ b/regression/python/github_6281_used/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.py
+--incremental-bmc
+^ERROR: NameError: name 'helper' is not defined$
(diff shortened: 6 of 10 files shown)


regression/python/github_6281_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+--no-unwinding-assertions --incremental-bmc
```
lucasccordeiro: Drop `--no-unwinding-assertions`: redundant (FAILED without it), and reintroduces the asymmetry you removed in #6276.

lucasccordeiro (changes requested):
Verified on a patched build: **the fix is incomplete** — two more null `["module"]` reads still SIGABRT with the same `type_error.302` whenever the relative-imported name is actually *used* (the common ROS2 case):

- `annotation_conversion.inl:1411` (`get_function_return_type`) — `from . import helper` + `helper()` aborts. The enclosing `catch (std::runtime_error&)` at 1443 does **not** catch `type_error` (it derives from `std::exception`, not `runtime_error`).
- `annotation_conversion.inl:3408` (`resolve_wildcard_import_func`) — `from . import *` aborts.

Both fire during `add_type_annotation` — the exact AST phase #6281 reports. Please null-guard both sites too. The two regression tests here only import-but-never-use the name, so they pass while the common pattern still core-dumps; add a test where the relative-imported name is used.

LukeW1999:
Thanks for the patched-build verification — you're right, the fix was incomplete. Both annotator sites are now null-guarded at the root:

- **`annotation_conversion.inl:1411` (`get_function_return_type`)** — added a `contains("module") && !is_null()` guard before `get_module`. When the relative import carries a null module name it now `throw`s a `std::runtime_error`, which the enclosing `catch (std::runtime_error&)` *does* catch, so control falls through to the wildcard/builtin probes instead of feeding null into `get_module` (the source of the uncatchable `type_error.302`).
- **`annotation_conversion.inl:3408` (`resolve_wildcard_import_func`)** — folded `node["module"].is_null()` into the existing skip condition, so relative star imports are skipped rather than dereferenced.

Verified on `from . import helper` + `helper()` (name actually used):
- pre-fix: `terminate ... type_error.302 ... dumped core` (rc 134)
- post-fix: clean `NameError: name 'helper' is not defined` (rc 254) — degrades to an unresolved import as intended, no abort.

Added `regression/python/github_6281_used` covering the name-*used* case (calls the relative-imported symbol), which is the pattern that hit both sites and that the original two tests missed. Also dropped the redundant `--no-unwinding-assertions` from `github_6281_fail/test.desc` per your inline comment (confirmed it still `VERIFICATION FAILED` without it).

lucasccordeiro:
@LukeW1999: can you please fix the CI?

LukeW1999:
CI fixed in 58f8055.

The red job was the `check_python_tests.sh` CPython smoke check, not the ESBMC build or regression run. That script runs each test as a bare `python3 main.py`, but a top-level relative import (`from . import helper`) is rejected by CPython itself — `ImportError: attempted relative import with no known parent package` — regardless of file layout (adding a sibling `helper.py` or an `__init__.py` does not help; only `python3 -m pkg.main` from a parent package works, which the smoke check does not do).

Since the whole point of these tests is an *unresolved relative import* (the null `module` field is what #6281 crashes on), they can only be validated through the ESBMC regression harness — not by executing them under CPython. I added `github_6281{,_used,_fail}` to the ignore list in that script, exactly as the existing ESBMC-only / intrinsic-using tests are handled. This is a niche relative-import case, so a targeted ignore-list entry (rather than new script logic) keeps it low-maintenance.

The ESBMC regression assertions themselves are unchanged and pass locally.

PR #6306, opened 2026-07-23
[python] Handle string/dotted parameter annotation (fix crash) (#6284)

### LINKED ISSUE #6284: [python] String/forward-ref (or dotted) parameter annotation crashes with nlohmann type_error.305
### Summary

A parameter annotated with a **string forward-reference** (`node: 'Foo'`) — or a **dotted** type (`node: a.b.Robot`) — crashes ESBMC with an uncaught `nlohmann` `type_error.305` (SIGABRT) when the parameter is used in the body. Found by sampling real ROS2 Python (e.g. `robomaster_ros` constructors typed `node: 'RoboMasterROS'`).

### Reproducer (self-contained)
```python
class C:
    def __init__(self, node: 'Foo'):
        x = node
```

### Command
```sh
esbmc repro.py --goto-functions-only
```

### Actual
```
  what():  [json.exception.type_error.305] cannot use operator[] with a string argument with string
<aborted> (exit 134)
```

### Ground truth (CPython)
`node: 'Foo'` is a valid forward-reference annotation (string; not evaluated). The program is valid Python.

### Root cause (initial)
When the frontend reads the parameter's annotation to infer types, it accesses the annotation as a structured node (`annotation["id"]` / `operator[]`) — but a forward-reference annotation is a JSON *string* (and a dotted annotation is an `Attribute` node), so nlohmann throws `type_error.305`. A plain `Name` annotation (`node: Foo`) or an unused string-annotated parameter does not crash.

### Notes
- Distinct from #6281; this aborts on the parameter-annotation read.

LukeW1999:
## Summary
Fixes #6284 — a parameter with a **string forward-reference annotation** (`node: 'Foo'`) or a **dotted annotation** (`node: a.b.Robot`), used in the body, crashed ESBMC with an uncaught `nlohmann` `type_error` (SIGABRT) during type inference.

Found by sweeping self-contained files from a large real ROS2 Python corpus (constructors typed e.g. `node: 'RoboMasterROS'`).

## Root cause
In `infer_type`'s `arg` branch, the annotation was read as `annotation["value"]["id"]` whenever it had a `value` key. That assumes a `Subscript` base (an object), but:
- a `Constant` (forward-ref string) annotation's `value` is a **string** → `operator[]("id")` on a string → `type_error.305`;
- an `Attribute` (dotted) annotation's `value` is an `Attribute` with no `id` → null → `type_error.302`.

## Fix
Dispatch on the annotation node shape instead of blindly indexing:
- `Constant` string → the type-name string itself (forward reference);
- `Subscript` (`list[int]`) → base `value["id"]` (guarded by `is_object()`/`contains("id")`);
- plain `Name` → `["id"]`;
- `Attribute` (dotted) → last component `["attr"]`;
- `BinOp` union → unchanged.

A resolvable forward-ref/dotted annotation now converts and verifies; an unresolvable one degrades to a clean error instead of a crash.

## Tests
- `regression/python/github_6284` — string-annotated parameter used in the body (`node: "Foo"`) → `VERIFICATION SUCCESSFUL`.
- `regression/python/github_6284_fail` — same, wrong assertion → `VERIFICATION FAILED`.

Validated against the annotation-inference regression subset (type-annotation*, param_*, forward-declaration*, github_6211/6242/6255/6258/6284 families) — 67/67 pass, no regressions.

### DIFF
--- a/regression/python/github_6284_fail/main.py
+++ b/regression/python/github_6284_fail/main.py
@@ -1,0 +1,18 @@
+# github.com/esbmc/esbmc/issues/6284
+# Same string-annotated parameter; the module must convert and its assertion be
+# checked (here a wrong one) instead of crashing at the annotation.
+class Foo:
+    def __init__(self, v: int):
+        self.v = v
+
+
+def use(node: "Foo") -> int:
+    x = node
+    return x.v
+
+
+def main() -> None:
+    assert use(Foo(42)) == 99
+
+
+main()
--- a/src/python-frontend/python_annotation/annotation_conversion.inl
+++ b/src/python-frontend/python_annotation/annotation_conversion.inl
@@ -2566,16 +2566,25 @@ InferResult python_annotation<Json>::infer_type(
       return InferResult::UNKNOWN;
 
     // A forward-reference (string) annotation `node: 'Foo'` is a Constant whose
-    // value is the type-name string; a dotted annotation `node: a.b.Robot` is an
-    // Attribute whose last component is the type name. Reading
-    // annotation["value"]["id"] blindly does operator[] on a JSON string / a
-    // null and aborts with nlohmann type_error (#6284), so dispatch on the node
-    // shape instead.
+    // value is the type-name string; a dotted annotation `node: mod.Robot` /
+    // `node: a.b.Robot` is an Attribute whose last component (attr) is the type
+    // name. Reading annotation["value"]["id"] blindly does operator[] on a JSON
+    // string / a null and aborts with nlohmann type_error (#6284), so dispatch
+    // on the node shape instead. The Attribute branch must precede the
+    // value.id branch: for a single-dot annotation the Attribute's value is a
+    // Name holding the module prefix (`mod`), so value.id would otherwise pick
+    // the prefix rather than the type name (`Robot`).
     if (
       stmt["annotation"].contains("_type") &&
       stmt["annotation"]["_type"] == "Constant" &&
+      stmt["annotation"].contains("value") &&
       stmt["annotation"]["value"].is_string())
       inferred_type = stmt["annotation"]["value"].template get<std::string>();
+    else if (
+      stmt["annotation"].contains("_type") &&
+      stmt["annotation"]["_type"] == "Attribute" &&
+      stmt["annotation"].contains("attr"))
+      inferred_type = stmt["annotation"]["attr"].template get<std::string>();
     else if (
       stmt["annotation"].contains("value") &&
       stmt["annotation"]["value"].is_object() &&
@@ -2584,11 +2593,6 @@ InferResult python_annotation<Json>::infer_type(
         stmt["annotation"]["value"]["id"].template get<std::string>();
     else if (stmt["annotation"].contains("id"))
       inferred_type = stmt["annotation"]["id"].template get<std::string>();
-    else if (
-      stmt["annotation"].contains("_type") &&
-      stmt["annotation"]["_type"] == "Attribute" &&
-      stmt["annotation"].contains("attr"))
-      inferred_type = stmt["annotation"]["attr"].template get<std::string>();
     else if (
       stmt["annotation"].contains("_type") &&
       stmt["annotation"]["_type"] == "BinOp")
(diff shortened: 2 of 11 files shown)


src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -2565,11 +2565,30 @@ InferResult python_annotation<Json>::infer_type(
...
+    // shape instead.
+    if (
+      stmt["annotation"].contains("_type") &&
+      stmt["annotation"]["_type"] == "Constant" &&
+      stmt["annotation"]["value"].is_string())
+      inferred_type = stmt["annotation"]["value"].template get<std::string>();
+    else if (
+      stmt["annotation"].contains("value") &&
+      stmt["annotation"]["value"].is_object() &&
+      stmt["annotation"]["value"].contains("id"))
```
lucasccordeiro, line 2591: Single-dot `mod.Robot` matches here → "mod" not "Robot". Move the Attribute branch above this one.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -2565,11 +2565,30 @@ InferResult python_annotation<Json>::infer_type(
...
       inferred_type =
         stmt["annotation"]["value"]["id"].template get<std::string>();
     else if (stmt["annotation"].contains("id"))
       inferred_type = stmt["annotation"]["id"].template get<std::string>();
+    else if (
```
lucasccordeiro, line 2596: Unreachable for single-dot annotations (`np.ndarray` → "np"); only 2+ dots reach here.

src/python-frontend/python_annotation/annotation_conversion.inl
```diff
@@ -2565,11 +2565,30 @@ InferResult python_annotation<Json>::infer_type(
...
+    // A forward-reference (string) annotation `node: 'Foo'` is a Constant whose
+    // value is the type-name string; a dotted annotation `node: a.b.Robot` is an
+    // Attribute whose last component is the type name. Reading
+    // annotation["value"]["id"] blindly does operator[] on a JSON string / a
+    // null and aborts with nlohmann type_error (#6284), so dispatch on the node
+    // shape instead.
+    if (
+      stmt["annotation"].contains("_type") &&
+      stmt["annotation"]["_type"] == "Constant" &&
+      stmt["annotation"]["value"].is_string())
```
lucasccordeiro, line 2581: Unguarded `["value"]` access; add a `contains("value")` guard like the sibling branches.

regression/python/github_6284_fail/main.py
```diff
@@ -0,0 +1,18 @@
+# github.com/esbmc/esbmc/issues/6284
+# Same string-annotated parameter; the module must convert and its assertion be
+# checked (here a wrong one) instead of crashing at the annotation.
+class Foo:
+    def __init__(self, v: int):
+        self.v = v
+
+
+def use(node: "Foo") -> int:
```
lucasccordeiro, line 9: No dotted-annotation test; add a `node: mod.Foo` case to cover the new path.

lucasccordeiro (commented):
Dotted-annotation handling is only half-correct and untested — see inline notes.

LukeW1999:
Addressed in 249c868 — all four inline notes:

- **Attribute branch ordering (2582 / 2587).** Moved the `Attribute → attr` branch ahead of the `value.id` branch. For a single-dot annotation the Attribute's value is a Name holding the module prefix, so `value.id` was matching first and returning the prefix (`mod`) instead of the type name (`Robot`); only 2+ dot annotations reached the Attribute branch. It now handles both single- and multi-dot dotted annotations. I confirmed directly that the reorder makes a single-dot annotation resolve to the final component (a probe with `mod.Foo` extracts `Foo`, not `mod`).
- **Unguarded `["value"]` (2577).** Added `contains("value")` before `["value"].is_string()`, matching the sibling branches.
- **Dotted-annotation test.** Added `github_6284_dotted` / `github_6284_dotted_fail` (`import mod; def use(node: mod.Foo)`), covering the dotted (Attribute) path end to end — both verify as expected under the ESBMC harness and run cleanly under the CPython smoke check (`mod.py` sits beside `main.py`, absolute import).

One note in the spirit of accuracy: these end-to-end tests exercise the dotted dispatch and verify cleanly, but at a concrete call site (`use(mod.Foo(42))`) ESBMC also resolves the parameter from the argument's runtime type, so the reorder is what corrects the *annotator's* returned type for a single-dot annotation rather than something a call-site assertion alone would flip. The reorder is verified by the attr-extraction probe above.

PR #6326, opened 2026-07-23
[contracts] preserve leading conjuncts in disjunctive multi-term ensures (#6298)

### LINKED ISSUE #6298: [contracts] --replace-call-with-contract silently drops leading conjuncts of a multi-term AND nested in a disjunctive ensures
## Summary

`--replace-call-with-contract` silently drops the leading conjuncts of a multi-term `&&` chain when that chain is nested inside one branch of a disjunctive (`||`) `ensures` clause and the chain contains 2+ comparisons involving `__ESBMC_old(...)`. The generated `ASSUME` at the call site ends up far weaker than the source `ensures` — in the reproducer below, it drops from a 3-term conjunct down to a single term — which lets `--replace-call-with-contract` accept states the real `ensures` clause explicitly forbids.

Confirmed via a `--goto-functions-only` dump comparing the *intended* `ensures` (from source) against the *actual* generated `ASSUME` instruction, and traced to a specific code path in `src/goto-programs/contracts/contracts.cpp`.

Reproduced on ESBMC 8.4.0 (release build) and independently confirmed against the `v8.4` tag source.

## Minimal reproducer

```c
typedef struct { int a; int b; int trigger; } S;

void f(S *s) {
    __ESBMC_requires(s != ((void*)0));
    __ESBMC_assigns(s->a, s->b);
    __ESBMC_ensures((s->trigger == 1 && s->a == 99 && s->b == 99) ||
                    (s->trigger != 1 && s->a == __ESBMC_old(s->a) && s->b == __ESBMC_old(s->b)));
    if (s->trigger == 1) {
      s->a = 99;
      s->b = 99;
    }
}

int main(void) {
  S s;
  s.trigger = 0;
  s.a = 5;
  s.b = 7;
  f(&s);
  __ESBMC_assert(s.a == 5 && s.b == 7, "both fields should be unchanged when trigger != 1");
  return 0;
}
```

Run as:
```
esbmc repro.c --replace-call-with-contract "*" --unwind 5
```

**Expected**: `VERIFICATION SUCCESSFUL` — when `trigger != 1`, the `ensures` clause explicitly states both `a` and `b` keep their old values, so the assertion must hold.

**Actual**: `VERIFICATION FAILED`, with a counterexample showing `s.a == -1` (or any other value) even though `trigger == 0` throughout — i.e. `a` is left completely unconstrained instead of pinned to its old value.

## Root cause

`--goto-functions-only` on the reproducer shows the exact generated call-site replacement:

```
DECL signed int old_snapshot_f_call_0;
ASSIGN old_snapshot_f_call_0=(&s)->a;
DECL signed int old_snapshot_f_call_1;
ASSIGN old_snapshot_f_call_1=(&s)->b;
ASSERT &s != (struct S *)0                                    // contract requires
ASSIGN (&s)->a=NONDET(signed int);
ASSIGN (&s)->b=NONDET(signed int);
ASSUME (&s)->trigger == 1 && (&s)->a == 99 && (&s)->b == 99
       || ((&s)->b == old_snapshot_f_call_1 ? 1 : 0)          // contract ensures
```

The second disjunct of the generated `ASSUME` is `(b == old(b) ? 1 : 0)` — the source's `trigger != 1 && a == old(a) && b == old(b)` has been reduced to just `b == old(b)`. The `trigger != 1` and `a == old(a)` conjuncts are gone entirely, so once the first disjunct is false, `a` is left fully unconstrained by the havoc.

This traces to `code_contractst::extract_ensures_from_body()` (`src/goto-programs/contracts/contracts.cpp:465`), which calls the static helper `inline_temporary_variables()` (`contracts.cpp:320`) to reconstruct the original boolean expression from Clang's short-circuit-evaluation GOTO lowering (a chain of conditional assignments to a `$tmp` variable).

The bug is in the "multiple assignments" branch of that helper (`contracts.cpp:380-437`). The comment at lines 389–392 documents the expected lowering shape for `(A && B) || (C && D)`:
```
tmp$7 = NONDET      (initialization)
tmp$7 = 1           (when first branch succeeds - short circuit)
tmp$7 = tmp$6 ? 1:0 (when first branch fails, use second branch result)
```
The code then does (lines 401–437): skip `NONDET`/constant-`0`/constant-`1` assignments, and on finding the *first* assignment whose RHS is non-trivial, immediately `break` (line 436) and recursively inline just that one value:

```cpp
// Found a meaningful assignment (references another temp or is a complex expr)
best_value = assign.value;
best_location = assign.location;
break;
```

This is sufficient for a 2-term conjunct (`A && B`), where Clang's lowering produces exactly one meaningful intermediate `$tmp` assignment to recover. It is **not** sufficient for a 3+-term conjunct such as `trigger != 1 && a == old(a) && b == old(b)`: Clang lowers this into a *chain* of nested short-circuit temporaries, and picking only the first meaningful one and recursing into it single-mindedly loses the outer conjuncts in the chain — exactly the "trigger != 1 && a == old(a)" that goes missing above.

## Confirmed not solver/simplification-level: it's specifically about conjunct count

A control case with an otherwise-identical structure but only **one** assigned field (so the untaken branch's frame-preservation is a single-conjunct chain, `trigger != 1 && a == old(a)`) verifies correctly:

```c
typedef struct { int a; int trigger; } S;

void f(S *s) {
    __ESBMC_requires(s != ((void*)0));
    __ESBMC_assigns(s->a);
    __ESBMC_ensures((s->trigger == 1 && s->a == 99) ||
                    (s->trigger != 1 && s->a == __ESBMC_old(s->a)));
    if (s->trigger == 1) { s->a = 99; }
}

int main(void) {
  S s;
  s.trigger = 0;
  s.a = 5;
  f(&s);
  __ESBMC_assert(s.a == 5, "field unchanged");
  return 0;
}
```
`esbmc repro_control.c --replace-call-with-contract "*" --unwind 5` → `VERIFICATION SUCCESSFUL`, as expected.

Swapping in the 3-term version above (adding one more field to preserve) is the only change that reproduces the bug — pointing squarely at the `inline_temporary_variables` reconstruction, not at solver behavior, `__ESBMC_old` handling in general (a single `old()` reference per disjunct works fine — see the control case), or disjunction handling in general (a disjunction of two *simple*, non-`old()` conditions also works fine).

## Impact

This silently weakens (or, depending on which conjunct survives, could differently misrepresent) the `ensures` clause actually enforced by `--replace-call-with-contract` for any function whose postcondition is a disjunction with a 3+-term conjunct in at least one branch — a natural and common shape for "either the guarded case fires and these fields update, or nothing changes" postconditions on functions that mutate 2+ fields. In our case this produced spurious `VERIFICATION FAILED` results (a genuinely correct program was rejected because the untaken branch's frame-preservation was under-constrained); the same mechanism could equally under-constrain in the other direction and let an actually-incorrect program pass, depending on the specific expression shape.

## Environment

- ESBMC version: 8.4.0 (release), reproduced; confirmed same code path present at `v8.4` tag in the `esbmc/esbmc` source
- OS: macOS, x86_64
- Default solver (bitwuzla); also reproduces with `--z3` (not solver-specific)

LukeW1999:
Fixes #6298.

## Problem

`--replace-call-with-contract` silently drops the leading conjuncts of a 3+-term AND chain nested inside a disjunctive `__ESBMC_ensures`. For

```c
__ESBMC_ensures(
  (s->trigger == 1 && s->a == 99 && s->b == 99) ||
  (s->trigger != 1 && s->a == __ESBMC_old(s->a) && s->b == __ESBMC_old(s->b)));
```

Clang lowers each 3-term `&&` chain into nested short-circuit temporaries. `inline_temporary_variables()` in `src/goto-programs/contracts/contracts.cpp` reconstructed only the innermost temporary's *value* assignment and then `break`-ed out, discarding the branch **guard** that carries the earlier conjuncts. The untaken disjunct

```
s->trigger != 1 && s->a == old(s->a) && s->b == old(s->b)
```

collapsed to just `s->b == old(s->b)`, leaving `s->a` unconstrained — a weaker assumption than the source contract, which spuriously rejects correct programs (and could let incorrect ones pass).

## Fix

Rebuild the full short-circuit conditional `guard ? const : value` instead of keeping the value arm alone:

- add `find_short_circuit_guard()` to locate the conditional `GOTO` selecting the temporary,
- recover and inline the guard, and reassemble the expression as an `if`-expr,
- handle either branch direction (guard jumping to the constant fallback or to the value),
- fall back to the value alone when the control-flow shape doesn't match (no constant fallback, no guarding branch, or an ill-typed conditional).

## Tests

Two new regression tests under `regression/function_contract/`:

- `replace_disjunctive_multiterm_and_pass` — the previously-rejected correct program now verifies (`VERIFICATION SUCCESSFUL`).
- `replace_disjunctive_multiterm_and_fail` — negative control asserting `a == 99` when `trigger != 1`, which must still `FAIL`. This guards against the fix over-correcting into a vacuously-true `ASSUME` that would accept any post-state.

Full `function_contract` regression suite passes locally: **273/273**, including the two new tests, no regressions.


src/goto-programs/contracts/contracts.cpp
```diff
@@ -446,9 +484,39 @@ static expr2tc inline_temporary_variables(
...
+      goto_programt::const_targett guard_loc =
+        is_nil_expr(fallback)
+          ? function_body.instructions.end()
+          : find_short_circuit_guard(
+              function_body, fallback_loc, value_loc, assume_location);
+      if (
+        guard_loc == function_body.instructions.end() ||
+        fallback->type->type_id != expr->type->type_id ||
+        inlined_value->type->type_id != expr->type->type_id)
+        return inlined_value;
```
lucasccordeiro: Silent one-arm fallback. Ternary `ensures` now flips FAILED→SUCCESSFUL — repro in review body.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -446,9 +484,39 @@ static expr2tc inline_temporary_variables(
...
+      // Recover the branch guard so the leading conjuncts are preserved. Fall
+      // back to the value alone when the control-flow shape doesn't match (no
+      // constant fallback, no guarding branch, or an ill-typed conditional).
+      goto_programt::const_targett guard_loc =
+        is_nil_expr(fallback)
+          ? function_body.instructions.end()
+          : find_short_circuit_guard(
+              function_body, fallback_loc, value_loc, assume_location);
+      if (
+        guard_loc == function_body.instructions.end() ||
```
lucasccordeiro: All three bail-outs return silently; `log_warning` here like the other failure paths.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -446,9 +484,39 @@ static expr2tc inline_temporary_variables(
...
+        fallback->type->type_id != expr->type->type_id ||
```
lucasccordeiro: `type_id` ignores width: `int` and `long` both pass. Compare full types.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -309,6 +309,38 @@ static std::vector<var_assignment_info> find_all_assignments(
...
+  const goto_programt::const_targett &assume_location)
+{
+  forall_goto_program_instructions (it, function_body)
+  {
+    if (it == assume_location)
+      break;
+    if (it->is_goto() && !is_true(it->guard))
+    {
+      for (const auto &t : it->targets)
+        if (t == fallback_loc || t == value_loc)
```
lucasccordeiro: Returns first match from function start, not nearest branch dominating `value_loc`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -309,6 +309,38 @@ static std::vector<var_assignment_info> find_all_assignments(
...
+// target is the fallback or the value assignment so the caller can rebuild the
+// full conditional regardless of which arm the branch jumps to. Returns
+// instructions.end() when no such branch exists.
+static goto_programt::const_targett find_short_circuit_guard(
+  const goto_programt &function_body,
+  const goto_programt::const_targett &fallback_loc,
+  const goto_programt::const_targett &value_loc,
+  const goto_programt::const_targett &assume_location)
+{
+  forall_goto_program_instructions (it, function_body)
```
lucasccordeiro: O(n) scan per temporary; contract extraction becomes quadratic in body size.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -408,17 +439,22 @@ static expr2tc inline_temporary_variables(
           continue;
         }
 
-        // Skip trivial constant assignments (0, 1, false, true)
-        // These are short-circuit markers, not the actual expression
+        // The short-circuit fallback: a trivial constant 0/1/false/true.
         if (is_constant_bool2t(assign.value))
         {
+          fallback = assign.value;
+          fallback_loc = assign.location;
```
lucasccordeiro: Every constant overwrites fallback; reverse order means earliest wins, unpaired with guard.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -430,13 +466,15 @@ static expr2tc inline_temporary_variables(
           return expr;
```
lucasccordeiro, line 430: Loop no longer breaks — this can now discard an already-found `value`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -378,24 +410,23 @@ static expr2tc inline_temporary_variables(
...
+      //   T = <value>              (guard held / short circuit not taken)
+      //   GOTO L_end
+      //   L_short: T = <const>     (const = false/0 for &&, true/1 for ||)
+      //   L_end:
+      // which is exactly `T == (<guard> ? <const> : <value>)`. The leading
+      // conjuncts of a 3+-term chain live in <guard> (as a reference to a nested
+      // temporary), so we must rebuild the whole conditional — picking <value>
+      // alone silently drops them (issue #6298).
+      expr2tc value;
+      goto_programt::const_targett value_loc =
```
lucasccordeiro: Not clang-format-11 clean; fits one line. CI will reject.

regression/function_contract/replace_disjunctive_multiterm_and_fail/main.c
```diff
@@ -0,0 +1,42 @@
+/* replace_disjunctive_multiterm_and_fail (issue #6298):
+ * Negative counterpart of replace_disjunctive_multiterm_and_pass. The ensures
+ * pins a to its old value (5) when trigger != 1, so asserting a == 99 must
+ * fail. This guards against the fix over-correcting into a vacuously-true
+ * ASSUME that would accept any post-state.
```
lucasccordeiro: Also FAILED on master; discriminates nothing. Assert against the other disjunct instead.

regression/function_contract/replace_disjunctive_multiterm_and_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract f --unwind 5
```
lucasccordeiro: `--unwind 5` is inert; neither test has a loop.

lucasccordeiro (commented):
**Blocking: this introduces an unsoundness regression.**

```c
__ESBMC_ensures(s->t == 1 ? (s->a == 99 && s->b == 99)
                          : (s->a == __ESBMC_old(s->a) && s->b == __ESBMC_old(s->b)));
```

Caller with `s.t = 1; s.a = 5;` then `__ESBMC_assert(s.a == 5)` — false, the contract says `a == 99`:

| | generated `ASSUME` | verdict |
|---|---|---|
| master | `b == old_b` | **FAILED** (correct) |
| this PR | `!(a == old_a) ? 0 : b == old_b` | **SUCCESSFUL** (unsound) |

The nil-`fallback` bail-out keeps one arm of a non-constant select and drops the condition, so an *incomplete* assumption becomes a *wrong* one and a false assertion is proved.

**Suggestion: prevent the lossy lowering instead of reconstructing it.** `goto_sideeffects.cpp:1504` already flattens `&&`/`||` chains for `__ESBMC_requires`/`__ESBMC_ensures` — it just doesn't recurse into nested chains, which is exactly #6298. Making that flatten recursive (~10 lines, no CFG pattern-matching) fixes the issue reproducer, a 4-term chain, an `||`-chain, mixed `(A||B) && (C||D) && old()`, and a disjunctive `requires`; `function_contract` stays 271/271. It also keeps the assumption readable, which matters for counterexamples:

```
this PR:  ASSUME t==1 && a==99 && b==99 || (!(!(t!=1) ? 0 : a==old_0 ? 1 : 0) ? 0 : b==old_1 ? 1 : 0)
flatten:  ASSUME t==1 && a==99 && b==99 || t!=1 && a==old_0 && b==old_1
```

**Missing coverage:** 4-term chain, `||`-chain (fallback constant `1`), the `--enforce-contract` direction (`extract_ensures_from_body` feeds both call sites), and a `KNOWNBUG` for the ternary above.

Verified by building this PR and master locally and running the `function_contract` suite against both.

LukeW1999:
Reworked along the lines you recommended — thank you, this is a much better fix. aa11be7.

**Dropped the reconstruction; fixed it upstream instead.** `contracts.cpp` is reverted entirely. The requires/ensures lowering in `goto_sideeffects.cpp` already flattened a *top-level* `&&`/`||` chain but called `remove_sideeffects()` on each whole part; for a disjunctive clause whose disjuncts are themselves `&&` chains with `__ESBMC_old()`, that lowered a nested `&&` at once and dropped its leading conjuncts (#6298). New `flatten_contract_clause()` recurses the full `&&`/`||` tree so `remove_sideeffects()` only ever sees a leaf — every conjunct is kept and `__ESBMC_old()` is snapshotted unconditionally at entry. The assumption also stays readable, e.g.

```
ASSUME t==1 && a==99 && b==99 || t!=1 && a==old_0 && b==old_1
```

**The unsoundness is gone.** Your ternary repro (`t==1; a=5;` then `assert a==5`) now returns **VERIFICATION FAILED** (correct), not the SUCCESSFUL the reconstruction produced. A ternary `?:` is not an `&&`/`||` chain, so it is lowered *soundly* but *imprecisely* (the else-branch `old()` constraints are not recovered) — pinned as `replace_ternary_ensures_knownbug` per your suggestion.

Because the reconstruction is gone, the inline notes on it (silent one-arm fallback, `type_id` width, first-match scan, quadratic scan, fallback overwrite, non-breaking loop, clang-format) no longer apply.

**Coverage added** (all the directions you listed):
- `replace_disjunctive_4term_pass` — 4-term conjunct
- `replace_disjunctive_orchain_pass` — pure `||`-chain
- `enforce_disjunctive_multiterm_pass` — the `--enforce-contract` / `extract_ensures_from_body` direction
- `replace_ternary_ensures_knownbug` — the ternary
- `replace_disjunctive_multiterm_and_fail` reworked to assert the **taken** disjunct (the old one failed on master too); inert `--unwind` dropped from both descs.

`function_contract` is **277/277**; the two edited C++ files are clang-format-11 clean.

PR #6423, opened 2026-07-26
[esbmc] Fix multi_property_check SIGSEGV under --schedule

### LINKED ISSUE #6422: [multi-property] SIGSEGV in multi_property_check when a goto-coverage mode is combined with --schedule
Combining any of the goto-coverage modes with `--schedule` segfaults in `bmct::multi_property_check`. No concurrency is required — a six-line sequential program is enough.

Reproduced on `master` at 4043035b4f, `ESBMC version 8.4.0 64-bit x86_64 linux`, built with Z3 v4.13.3 (`CMAKE_BUILD_TYPE=Release`).

## Reproducer

```c
// sched_min.c
int main(void)
{
  if (x > 0)
    return 1;
  return 0;
}
```

```console
$ esbmc sched_min.c --branch-coverage --schedule
...
Generated 2 VCC(s), 2 remaining after simplification (5 assignments)
$ echo $?
139
```

## Affected combinations

| flags (with `--schedule`) | result |
| --- | --- |
| `--branch-coverage` | **SIGSEGV** |
| `--condition-coverage` | **SIGSEGV** |
| `--branch-function-coverage` | **SIGSEGV** |
| `--assertion-coverage` | exit 0, coverage printed |
| *(no coverage flag)* | exit 0 |

Each of these is fine without `--schedule`. `--assertion-coverage` surviving is suggestive: it is the one mode that routes solved claims into `reached_mul_claims` rather than through the `is_goto_cov` claim path (`bmc.cpp:1969`).

## It needs at least one remaining claim

This is the sharpest discriminator I found — the crash is in the per-claim work, not in setup:

```console
$ esbmc -c 'int main(void){return 0;}' ...        # 0 VCC(s)
Generated 0 VCC(s), 0 remaining after simplification (1 assignments)
[Coverage]
Branch Coverage: N/A (no branches)
                                                  # exit 0, no crash

$ # the reproducer above:  2 VCC(s)  -> SIGSEGV
```

So a program with no coverage claims completes normally; one claim is enough to crash.

## Backtrace

```
Program received signal SIGSEGV, Segmentation fault.
#0  bmct::multi_property_check(symex_target_equationt const&, unsigned long, smt_convt&) ()
#1  bmct::run_thread(std::shared_ptr<symex_target_equationt>&) ()
#2  bmct::start_bmc() ()
#3  esbmc_parseoptionst::do_bmc(bmct&) ()
#4  esbmc_parseoptionst::doit() ()
#5  main ()
```

(Release build, so no line numbers — a `Debug` build would localise it inside the frame.)

## Where the two paths diverge

`bmc.cpp:1608-1614` picks a different formula generator for `--schedule`, then `dynamic_pointer_cast`s the result:

```cpp
    goto_symext::symex_resultt solver_result =
      options.get_bool_option("schedule") ? symex->generate_schedule_formula()
                                          : symex->get_next_formula();
    ...
    eq =
      std::dynamic_pointer_cast<symex_target_equationt>(solver_result.target);
```

and `multi_property_check` then iterates `for (size_t i = 1; i <= remaining_claims; i++)` (`bmc.cpp:1911`) constructing a `claim_slicer` per claim.

I checked the obvious candidate and it does **not** appear to be the cause: `schedule_target` is a real `target_template->clone()` when `schedule` is set (`reachability_tree.cpp:313-315`), only `nullptr` on the non-schedule path (line 332), so this is probably not a null `eq`. I did not localise it further — the per-claim slicing against a schedule-generated equation is where I would look next.

## Why I am filing rather than fixing

I hit this while testing flag-combination edge cases for #5787 (`--dead-code-check`, which reuses the branch-coverage instrumentation and therefore inherits the crash: `--dead-code-check --schedule` segfaults identically). It is not caused by that PR — the branch reproduces it only because `--branch-coverage --schedule` already does on `master`, and I confirmed the matrix above against a `master` binary built at 4043035b4f with no `--dead-code-check` support compiled in.

Since the fault is in the shared multi-property/coverage path rather than anything dead-code specific, it seemed wrong to fold a fix into that PR.

Possibly related but distinct: #6413 (multi-property reporting a property both PASSED and FAILED) touches the same function; #86 is a different `--schedule` limitation.

pmatos:
## Summary

- `bmct::multi_property_check`'s first statement read `symex->get_cur_state().simplified_claims`. Under `--schedule`, `reachability_treet::generate_schedule_formula()` loops until `exploration_frames` is fully drained before returning (unlike `get_next_formula()`, which returns after exactly one path while `cur_frame_it` is still valid), so `cur_frame_it` is left dangling and the dereference inside `get_cur_state()` segfaults.
- Fix: thread the already-computed `simplified_claims` through from the caller's `symex_resultt` (`solver_result.simplified_claims`) as a new parameter to `multi_property_check`, instead of re-deriving it from live exploration state that may no longer be valid.
- The issue's flag matrix shows `--assertion-coverage --schedule` surviving where `--branch-coverage`/`--condition-coverage`/`--branch-function-coverage` crash. That's not a difference in how assertion-coverage is handled — on the reported reproducer it yields 0 remaining claims, so `run_thread` takes the early return at `bmc.cpp:1595` before `multi_property_check` is ever reached. `github_6422_5` adds an assert-bearing variant that does exercise `--assertion-coverage --schedule` through `multi_property_check` and crashes on unfixed code; `github_6422_4` documents the zero-claims early-return path (it passes on both fixed and unfixed code, since it never reaches the code path this PR touches).

Fixes #6422

## Test plan

- [x] `regression/goto-coverage/github_6422_{1,2,3,5}` — verified SIGSEGV (exit 139) against the unfixed binary, verified `VERIFICATION FAILED` with correct coverage percentages against the fixed binary
- [x] `regression/goto-coverage/github_6422_4` — control case documenting the zero-remaining-claims early-return path under `--schedule`
- [x] Full `regression/goto-coverage` ctest suite (124 tests) — all passing
- [x] `regression/esbmc` ctest sample (1394 tests) — 99% passing; 11 failures are pre-existing, caused by boolector not being compiled into this local build (confirmed unrelated: same 11 tests fail identically on unfixed `master`)
- [x] clang-format — diff introduces no new violations (verified against clang-format-15; 5 pre-existing violations on unrelated lines are unchanged in count/content, only shifted by line offset)

### DIFF
--- a/regression/goto-coverage/github_6422_2/main.c
+++ b/regression/goto-coverage/github_6422_2/main.c
@@ -1,0 +1,7 @@
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    return 1;
+  return 0;
+}
--- a/regression/goto-coverage/github_6422_4/test.desc
+++ b/regression/goto-coverage/github_6422_4/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--assertion-coverage --schedule
+^Total Asserts: 0
+^VERIFICATION SUCCESSFUL$
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -1643,7 +1643,10 @@ smt_resultt bmct::run_thread(std::shared_ptr<symex_target_equationt> &eq)
        (options.get_bool_option("inductive-step") &&
         options.get_bool_option("loop-invariant"))))
       return multi_property_check(
-        *eq, solver_result.remaining_claims, *runtime_solver);
+        *eq,
+        solver_result.remaining_claims,
+        solver_result.simplified_claims,
+        *runtime_solver);
 
     smt_resultt result = run_decision_procedure(*runtime_solver, *eq);
 
@@ -1773,6 +1776,7 @@ int bmct::ltl_run_thread(symex_target_equationt &equation) const
 smt_resultt bmct::multi_property_check(
   const symex_target_equationt &eq,
   size_t remaining_claims,
+  size_t simplified_claims,
   smt_convt &runtime_solver)
 {
   // Initial values
@@ -1783,7 +1787,10 @@ smt_resultt bmct::multi_property_check(
 
   // Add summary tracking
   SimpleSummary summary;
-  summary.simplified_properties = symex->get_cur_state().simplified_claims;
+  // Taken from the caller's symex_resultt rather than symex->get_cur_state():
+  // under --schedule, generate_schedule_formula() drains exploration_frames
+  // completely before returning, leaving cur_frame_it invalid.
+  summary.simplified_properties = simplified_claims;
   summary.total_properties = remaining_claims + summary.simplified_properties;
   summary.passed_properties =
     summary.passed_properties + summary.simplified_properties;
--- a/src/esbmc/bmc.h
+++ b/src/esbmc/bmc.h
@@ -94,6 +94,7 @@ protected:
   smt_resultt multi_property_check(
     const symex_target_equationt &eq,
     size_t remaining_claims,
+    size_t simplified_claims,
     smt_convt &runtime_solver);
 
   std::vector<std::unique_ptr<ssa_step_algorithm>> algorithms;
--- a/regression/esbmc/github_6422_6/main.c
+++ b/regression/esbmc/github_6422_6/main.c
@@ -1,0 +1,9 @@
+#include <assert.h>
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    assert(x > 0);
+  assert(x != 42);
+  return 0;
+}
--- a/regression/esbmc/github_6422_6/test.desc
+++ b/regression/esbmc/github_6422_6/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--multi-property --schedule
+^Properties: 2 verified ✓ 1 passed, ✗ 1 failed$
+^VERIFICATION FAILED$
--- a/regression/esbmc/github_6422_6/test.desc
+++ b/regression/esbmc/github_6422_6/test.desc
@@ -1,5 +1,4 @@
 CORE
 main.c
 --multi-property --schedule
-^Properties: 2 verified ✓ 1 passed, ✗ 1 failed$
 ^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_6422_1/main.c
+++ b/regression/goto-coverage/github_6422_1/main.c
@@ -1,0 +1,7 @@
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    return 1;
+  return 0;
+}
--- a/regression/goto-coverage/github_6422_1/test.desc
+++ b/regression/goto-coverage/github_6422_1/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+--branch-coverage --schedule
+^Properties: 2 verified, ✗ 2 failed$
+^Branches : 2
+^Reached : 2
+^Branch Coverage: 100%$
+^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_6422_2/test.desc
+++ b/regression/goto-coverage/github_6422_2/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--condition-coverage --schedule
+^Reached Conditions:  2
+^Total Conditions:  2
+^Condition Coverage: 100%$
+^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_6422_3/main.c
+++ b/regression/goto-coverage/github_6422_3/main.c
@@ -1,0 +1,7 @@
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    return 1;
+  return 0;
+}
--- a/regression/goto-coverage/github_6422_3/test.desc
+++ b/regression/goto-coverage/github_6422_3/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--branch-function-coverage --schedule
+^Function Entry Points & Branches : 3
+^Reached : 3
+^Branch Coverage: 100%$
+^VERIFICATION FAILED$
--- a/regression/goto-coverage/github_6422_4/main.c
+++ b/regression/goto-coverage/github_6422_4/main.c
@@ -1,3 +1,7 @@
+// No asserts to instrument, so this exercises the remaining_claims == 0
+// early return in bmct::run_thread, never reaching multi_property_check.
+// Control case for the --schedule fix, not a regression pin for it
+// (see github_6422_5 for that).
 int main(void)
 {
   int x = __VERIFIER_nondet_int();
--- a/regression/goto-coverage/github_6422_4/main.c
+++ b/regression/goto-coverage/github_6422_4/main.c
@@ -1,0 +1,7 @@
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    return 1;
+  return 0;
+}
--- a/regression/goto-coverage/github_6422_5/main.c
+++ b/regression/goto-coverage/github_6422_5/main.c
@@ -1,0 +1,8 @@
+#include <assert.h>
+int main(void)
+{
+  int x = __VERIFIER_nondet_int();
+  if (x > 0)
+    assert(x > 0);
+  return 0;
+}
--- a/regression/goto-coverage/github_6422_5/test.desc
+++ b/regression/goto-coverage/github_6422_5/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+--assertion-coverage --schedule
+^Total Asserts: 1
+^Total Assertion Instances: 1
+^Reached Assertion Instances: 1
+^Assertion Instances Coverage: 100%$
+^VERIFICATION FAILED$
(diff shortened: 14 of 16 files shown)


src/esbmc/bmc.cpp
```diff
@@ -1783,7 +1787,10 @@ smt_resultt bmct::multi_property_check(
 
   // Add summary tracking
   SimpleSummary summary;
-  summary.simplified_properties = symex->get_cur_state().simplified_claims;
+  // Taken from the caller's symex_resultt rather than symex->get_cur_state():
```
lucasccordeiro, line 1790: The crash is not goto-coverage-specific, but every test added here passes a coverage flag, so the general case stays unprotected.

Verified against unfixed `master` (066827889e), no coverage flag:

```c
#include <assert.h>
int main(void) {
  if (x > 0) assert(x > 0);
  assert(x != 42);
  return 0;
}
```

`--multi-property --schedule` → exit 139, and `--multi-property --schedule --base-case` → exit 139.

The coverage flags only matter because they imply `--multi-property` and guarantee a non-zero claim count; the trivial reproducer in `github_6422_1..4` needs them solely to avoid the `remaining_claims == 0` early return. Please add a reproducer under `regression/esbmc/` using plain `--multi-property --schedule` with an assertion, so a future refactor of coverage handling cannot silently drop coverage of the actual defect. The PR title and commit subject should also drop "+ goto-coverage", which understates the blast radius.

src/esbmc/bmc.cpp
```diff
@@ -1783,7 +1787,10 @@ smt_resultt bmct::multi_property_check(
...
+  // under --schedule, generate_schedule_formula() drains exploration_frames
+  // completely before returning, leaving cur_frame_it invalid.
+  summary.simplified_properties = simplified_claims;
```
lucasccordeiro, line 1793: This removes the only current caller that dereferences the dangling iterator, but leaves the trap itself:

```cpp
// reachability_tree.cpp:343
execution_statet &reachability_treet::get_cur_state()
{
  return *cur_frame_it->state;
}
```

After `generate_schedule_formula()` drains `exploration_frames`, `cur_frame_it` is invalid and any post-symex caller re-introduces this exact SIGSEGV with no diagnostic. Since the comment here already documents the invariant, please enforce it where it is established rather than only at this one consumer — e.g. reset `cur_frame_it` to a sentinel at the end of `generate_schedule_formula()` and assert validity in `get_cur_state()`, or document the invalidation in `reachability_treet`'s header next to `get_cur_state()`. Otherwise the next caller pays the same debugging cost.

regression/goto-coverage/github_6422_4/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
+main.c
+--assertion-coverage --schedule
```
lucasccordeiro, line 3: By your own test plan this case "passes on both fixed and unfixed code", so it has no regression value — it documents the zero-claims early return. Nothing in `main.c` or this `test.desc` says so, so a future maintainer seeing `--assertion-coverage --schedule` here will reasonably assume it guards the fix and may delete `github_6422_5` as redundant. Add a one-line comment to `main.c` stating that this exercises the `remaining_claims == 0` early return at `bmc.cpp` and never reaches `multi_property_check`.

regression/goto-coverage/github_6422_2/main.c
```diff
@@ -0,0 +1,7 @@
+int main(void)
```
lucasccordeiro, line 1: `github_6422_1`, `_2`, `_3` and `_4` are byte-identical (all four blobs are `100405e9179`); only the flag line in each `test.desc` differs. Four copies of the same input is avoidable churn. Either fold the flag variants into fewer directories, or — if the suite requires one directory per flag set — note in each `main.c` that it is intentionally shared so a later edit to one does not silently diverge from the others.

src/esbmc/bmc.h
```diff
@@ -94,6 +94,7 @@ class bmct
   smt_resultt multi_property_check(
     const symex_target_equationt &eq,
     size_t remaining_claims,
+    size_t simplified_claims,
```
lucasccordeiro, line 97: `symex_resultt::simplified_claims` is `unsigned int` (goto_symex.h:116) but the parameter is `size_t`, so every call site widens implicitly. `remaining_claims` above has the same mismatch, so this is consistent with what is already there — but since you are touching the signature, consider making both `unsigned int` to match the source of truth, or add a brief note that the widening is deliberate.

lucasccordeiro (changes requested):
The dangling-`cur_frame_it` analysis is right and the fix is the correct shape. Two substantive concerns before this lands: the bug is not goto-coverage-specific (verified locally), and the underlying landmine in `get_cur_state()` is left in place. Details inline.

src/esbmc/bmc.cpp
pmatos: Fixed in `949a4a629e`.

Confirmed your diagnosis — plain `--multi-property` unconditionally sets `base-case` (`esbmc_parseoptions.cpp`), so the same `multi_property_check` path is reachable without any coverage flag; reproduced the crash on unfixed `6186bac84d` with `--multi-property --schedule` (exit 139) before writing the test. Added `regression/esbmc/github_6422_6` using your reproducer, with expected-output regexes taken from an actual run against the fixed binary rather than hand-derived.

Also updated the PR title to drop "+ goto-coverage". Leaving the pushed commit subject (`3088721010`) as published rather than amending it — this repo's convention is to never rewrite pushed commit history, and amending would need a force-push. Happy to do that too if you'd rather have it changed.

Validation: `regression/esbmc/github_6422_6` passes via `testing_tool.py`; confirmed exit 139 on unfixed code first.

src/esbmc/bmc.cpp
pmatos: Fixed in `bab64a0048`.

Added `assert(!exploration_frames.empty() && cur_frame_it != exploration_frames.end())` to both `get_cur_state()` overloads and the sibling `get_cur_scheduler_frame()` overloads (same unguarded dereference), plus documented the invariant next to `get_cur_state()`'s declaration in `reachability_tree.h`.

Checked every existing caller (grepped across `src/`) to confirm the assert can't misfire: they all run from within the active symex loop (`generate_schedule_formula`/`get_next_formula`'s `while` bodies, or builtin-function handlers invoked from `symex_step`), where `exploration_frames` is guaranteed non-empty. The only post-drain caller was `multi_property_check`, which the previous commit already stopped calling `get_cur_state()` from.

Validation: full `goto-coverage` suite (128 tests) plus the `--schedule` concurrency tests (`esbmc-unix/00_bbuf_02`, `00_bbuf_02_new`) all pass unaffected.

regression/goto-coverage/github_6422_4/test.desc
pmatos: Fixed in `b25c40e29c`.

Added a comment to `github_6422_4/main.c` stating it exercises the `remaining_claims == 0` early return and never reaches `multi_property_check` — worded as what the test *does* pin (the early-return path) rather than "no regression value", and naming the code path instead of a line number so it doesn't rot.

Validation: `github_6422_4` still passes via `testing_tool.py` (comment-only change to the C source).

regression/goto-coverage/github_6422_2/main.c
pmatos: **Not relevant (matches existing repo convention)** — Folding the flag variants into fewer directories isn't available given the current harness: `regression/testing_tool.py` reads exactly one args line per `test.desc`, and `CMakeLists.txt`'s `add_esbmc_regression_test` registers one ctest test per directory, so one directory per flag set is the only option without a harness change well outside this PR's scope. The fallback ("note that `main.c` is intentionally shared") isn't accurate either: `github_6422_4` and `_5` share the same flags but use *different* `main.c` with different expectations, so divergence between directories is the normal, supported case here, not a hazard to guard against. Same-suite precedent already does exactly this pattern: `github_1618_1` through `_5` are five byte-identical `main.c` files varying only `--unwind N`, with no shared-fixture note anywhere in that group.

_This assessment was made by two independent AI reviewers (Claude Opus and Codex GPT-5.6). If you disagree, please reply and we'll re-evaluate._

src/esbmc/bmc.h
pmatos: **Style preference (no change)** — `size_t` is the deliberate, already-established choice: `SimpleSummary`'s counters are `std::atomic<size_t>` and the per-claim loop bound is `for (size_t i = 1; i <= remaining_claims; i++)`, so narrowing the new parameter to `unsigned int` would just push the same widening into the function body while adding mixed-width arithmetic. `remaining_claims` (unchanged by this PR) already has the identical `unsigned int` → `size_t` widening at the call site; `simplified_claims` only follows its sibling's existing pattern. The comment already at the assignment (pointing to the caller's `symex_resultt` as the source of truth) covers where the `unsigned int` value comes from, so an additional note would document the language rather than the code.

_This assessment was made by two independent AI reviewers (Claude Opus and Codex GPT-5.6). If you disagree, please reply and we'll re-evaluate._

pmatos:
@lucasccordeiro Regarding your [CHANGES_REQUESTED review](https://github.com/esbmc/esbmc/pull/6423#pullrequestreview-4781981162): both substantive concerns are addressed, with full detail on the respective inline threads:

- **Not coverage-specific**: confirmed and reproduced — added `regression/esbmc/github_6422_6` (plain `--multi-property --schedule`, no coverage flag) in `949a4a629e`, and retitled the PR to drop "+ goto-coverage".
- **`get_cur_state()` landmine left in place**: hardened both `get_cur_state()` overloads and the sibling `get_cur_scheduler_frame()` overloads with an assert plus a documented invariant, in `bab64a0048`.

Also fixed a test-documentation nit on `github_6422_4` (`b25c40e29c`), and replied with no-code-change rationale on two further nits (duplicate `main.c` across the coverage-flag directories, and `size_t` vs `unsigned int`) — both match existing repo convention, detail on their threads.

Let me know if anything still needs another look.

lucasccordeiro (approved):

Issue #4495, opened 2026-05-13
[esbmc] surface provably-dead code as advisory CWE-561


lucasccordeiro:
### Background

PR #4488 introduced CWE annotation for property violations (see [`docs/cwe-mapping.md`](../blob/master/docs/cwe-mapping.md)). ESBMC's reachability machinery (`goto-coverage`, `--enable-unreachability-intrinsic`) can already prove that specific statements are unreachable from the program's entry point under all inputs — this is the soundness backbone of the existing CWE-617 reachability check. The dual property — proving a statement *is dead* — is therefore equally within reach but is not currently reported as a CWE finding.

### CWE-561 — Dead Code

[CWE-561](https://cwe.mitre.org/data/definitions/561.html) (`ALLOWED` for vulnerability mapping in CWE 4.20) covers code that is unreachable from any input. Unlike compiler `-Wunreachable-code`, BMC-based detection is **sound under non-trivial guards** (the typical compiler analysis is not).

### Proposal

This is an advisory / informational CWE rather than a violation, so the integration differs slightly from the existing scheme:

1. Add a new analysis pass (gated behind `--dead-code-check`) under `src/goto-programs/` that iterates over every statement in the goto program and submits `__ESBMC_assert(false, "dead code at <loc>")` *under the path condition that reaches the statement's entry*. If ESBMC proves UNSAT (the path is unreachable), the assertion is vacuously true → the statement is dead.
2. When the pass detects a dead statement, emit a non-fatal *advisory* entry in the trace (separate from the violation stream) with comment `dead code reached: never`. Surface it in JSON as a new `advisories[]` array on the run.
3. Extend SARIF output to emit dead-code findings as `result.level = "note"` (rather than `"error"`) — SARIF's existing levels cover advisory findings cleanly.
4. Add a `rules_table()` entry in `src/util/cwe_mapping.cpp` mapping the new comment to `{ "dead-code", "Dead code", {561} }`.
5. Unit + regression tests as for other CWEs.
6. Update `docs/cwe-mapping.md` and explicitly note the *advisory* nature (does not flip the verdict to FAILED).

### Care points

- This is potentially slow (one SMT call per statement). Default-off is essential.
- Must not interact with the existing CWE-617 reachability checker; "dead code" and "reachable error" are inverse properties of the same SMT machinery.
- SV-COMP compatibility: advisory findings must not change the `FALSE_*` / `TRUE` verdict.

### Acceptance criteria

[...]

PR #6431, opened 2026-07-26
[Coverage] Update ESBMC's coverage output format

ChenfengWei0:
Issue:  #6387

code:
```c
int main()
{
  int a, b, c, d, e;
  //------------------------------------------------------------
  __ESBMC_assert( a * b == b * a, "assertion 1");
  __ESBMC_assert( a * c == c * a, "assertion 2");
}
```

## Previous Result:

```
ESBMC --multi-property --assertion-coverage tiny-1.cpp
[Counterexample]


State 1 file tiny-1.cpp line 6 column 3 function main thread 0
Violated property:
  file tiny-1.cpp line 6 column 3 function main
  assertion 2
  0

Solving claim 'assertion 1 at file tiny-1.cpp line 5 column 3 function main' with solver Z3 v4.13.3
[PROGRESS] Solving with solver Z3 v4.13.3
✗ FAILED: 'assertion 1 at file tiny-1.cpp line 5 column 3 function main'

[Counterexample]


State 1 file tiny-1.cpp line 5 column 3 function main thread 0
Violated property:
  file tiny-1.cpp line 5 column 3 function main
  assertion 1
  0

Properties: 2 verified, ✗ 2 failed
Solver: Z3 v4.13.3 • Decision procedure total time: 0.001s • Avg: 0.000s/property

[Coverage]

Total Asserts: 2
Unreached Asserts: 0
Total Assertion Instances: 2
Reached Assertion Instances: 2
Assertion Instances Coverage: 100%

VERIFICATION FAILED
```

The `VERIFICATION FAILED` is kind of misleading.

## Patched

```
✓ REACHED: 'assertion 2 at file tiny-1.cpp line 6 column 3 function main'

[Reachability trace]


State 1 file tiny-1.cpp line 6 column 3 function main thread 0
Reached:
  file tiny-1.cpp line 6 column 3 function main
  assertion 2

Solving claim 'assertion 1 at file tiny-1.cpp line 5 column 3 function main' with solver Z3 v4.13.3
[PROGRESS] Solving with solver Z3 v4.13.3
✓ REACHED: 'assertion 1 at file tiny-1.cpp line 5 column 3 function main'

[Reachability trace]


State 1 file tiny-1.cpp line 5 column 3 function main thread 0
Reached:
  file tiny-1.cpp line 5 column 3 function main
  assertion 1

Coverage goals: 2 ✓ 2 reached
Solver: Z3 v4.13.3 • Decision procedure total time: 0.001s • Avg: 0.000s/property

[Coverage]

Total Asserts: 2
Unreached Asserts: 0
Total Assertion Instances: 2
Reached Assertion Instances: 2
Assertion Instances Coverage: 100%
BMC program time: 0.023s

COVERAGE ANALYSIS COMPLETE
```

---

We now have two final signals
- `COVERAGE ANALYSIS COMPLETE`: every coverage goal received a definitive reachable/unreachable answer, so the percentages above are exact.
- `COVERAGE ANALYSIS INCOMPLETE`: some goals were never decided or the program was not fully explored, so the percentages are lower bounds. The reason is printed underneath — e.g. `--multi-fail-fast` cut the run short, the solver returned no result, `--smt-formula-only` solved nothing, the assertion-instance total was non-deterministic, or the run hit its maximum k step.

---

Test added: `github_6387_1 ` ... `github_6387_11`. Eleven new tests in total


src/esbmc/bmc.cpp
```diff
@@ -63,6 +63,61 @@ std::unordered_multiset<std::string> goto_functionst::reached_mul_claims;
...
+std::mutex goto_functionst::cov_incomplete_mutex;
+
+/* True when this run measures goto coverage rather than verifying the
+ * program. Every coverage mode instruments the program with `assert` probes
+ * whose falsifiability *is* the measurement: a SAT probe means "this location
+ * is reachable", not "a property was violated" (github issue #6387). The
+ * original assertions no longer carry their source condition, so a coverage
+ * run proves nothing about the program and must not emit a verification
+ * verdict, a counterexample, or a violation witness. */
+static bool is_coverage_run(const optionst &options)
```
lucasccordeiro: This is a second, independently maintained definition of "coverage run". `esbmc_parseoptionst::is_coverage` (`src/esbmc/parseoptions/process_goto_program.cpp:102`) has its own list, and the two disagree: that one also includes `dead-code-check`, `k-path-coverage` and `k-path-coverage-claims`.

Any future coverage mode has to be added in both places or the behaviour splits — the driver takes the `is_coverage` path (exit 0, prints the completeness line) while `report_result` still emits a verdict, or vice versa. Please set a single option in `process_goto_program` (e.g. `options.set_option("coverage-run", true)`, alongside the `multi-property` / `base-case` sets already there) and have both sites read it.

src/esbmc/bmc.cpp
```diff
@@ -1368,6 +1485,11 @@ void bmct::report_result(smt_resultt &res)
   // multi_property_check; suppress any global verdict from this level.
   if (options.get_bool_option("diagnose-unknown-properties"))
     return;
+  // A coverage run replaced the program's assertions with reachability
+  // probes, so it neither proved nor refuted anything about the program.
+  // Its result is the [Coverage] block, not a verification verdict.
+  if (is_coverage_run(options))
```
lucasccordeiro: This return is placed *before* the `dead-code-check` block below, which deliberately keeps `log_error("SMT solver failed")` on P_ERROR and stays silent only on P_SMTLIB. As written, a coverage run whose solver failed prints no error at all — the only trace is a generic `reason: the solver did not return a result for every goal` line much later.

Suggest mirroring the dead-code handling: return early only for a completed run, and still surface P_ERROR.

src/esbmc/bmc.cpp
```diff
@@ -2090,7 +2213,13 @@ smt_resultt bmct::multi_property_check(
                        &runtime_solver](const size_t &i) {
     //"multi-fail-fast n": stop after first n SATs found.
     if (is_fail_fast && fail_fast_cnt >= fail_fast_limit)
+    {
+      // The skipped goals were never solved. Counting them as unreached
+      // would report a percentage that looks measured but is not.
+      if (is_cov_run)
+        note_undecided_cov_goal("--multi-fail-fast limit reached");
```
lucasccordeiro: Two problems here.

1. This fires before `claim_slicer claim(i, ...)` is built, so it counts *every* remaining job, not just instrumented probes. In a program that also has non-coverage claims the reported "N goal(s) undecided" is too high.
2. It bumps only the global `undecided_cov_goals`, not `summary.undecided_goals`. The two counters then contradict each other in the same output — `github_6387_11` is exactly this case: `Coverage goals: 1 ✓ 1 reached` (no `undecided` shown, because `summary.undecided_goals == 0`) followed by `COVERAGE ANALYSIS INCOMPLETE: 2 goal(s) undecided`.

Either derive the closing line from `summary` or increment both consistently; having two sources of truth for the same number will drift again.

src/esbmc/bmc.cpp
```diff
@@ -2203,7 +2336,39 @@ smt_resultt bmct::multi_property_check(
...
+          RESET,
+          prettify_solidity_expr(claim.claim_cstr));
+      else if (solver_result == P_UNSATISFIABLE)
+        log_status(
+          "{}- UNREACHED{}: '{}'",
+          YELLOW,
+          RESET,
+          prettify_solidity_expr(claim.claim_cstr));
+      else
+        note_undecided_cov_goal(
```
lucasccordeiro: The undecided branch records the reason but prints nothing for the goal itself. Every other outcome gets a `✓ REACHED` / `- UNREACHED` line, so a goal the solver could not decide silently disappears from the per-goal listing and the printed lines no longer sum to the goal count. Add a `? UNDECIDED: '<claim>'` line here.

src/esbmc/bmc.cpp
```diff
@@ -2203,7 +2336,39 @@ smt_resultt bmct::multi_property_check(
...
+    }
+    else if (is_goto_cov && solver_result == P_SATISFIABLE)
+    {
+      // A violated claim the coverage pass did not instrument: another
+      // function under --function, or a check symex injects afterwards. A
+      // coverage run reports no violations, so without this it would vanish
+      // entirely. It is not a completeness problem — whether it truncates
+      // exploration depends on the claim — so it is listed separately from
+      // the reasons the percentages may be lower bounds.
+      note_cov_suppressed_violation(claim.claim_msg);
```
lucasccordeiro: `cov_suppressed_violations` is a `std::set` keyed on `claim.claim_msg` alone, so (a) two genuinely different violations that share a comment collapse into one entry, and (b) the warning gives the user no `file:line` to act on — for `assert(x)` lowered without a comment the message is just the stringified guard.

Use the same signature the reached-claim bookkeeping already uses, `claim.claim_msg + "\t" + claim.claim_loc` (see the `claim_sig` construction earlier in this function), and print the location in the warning.

Also note this branch is reached for `--dead-code-check` too (`is_goto_cov` includes it), where `report_coverage_completeness()` is never called — so the set just grows and is never printed. Gate it on `is_cov_run`.

src/esbmc/bmc.cpp
```diff
@@ -2554,27 +2749,45 @@ void bmct::report_simple_summary(const SimpleSummary &summary) const
...
   std::ostringstream properties_oss;
-  properties_oss << "Properties: " << summary.total_properties << " verified";
+  if (is_coverage_run(options))
+  {
+    // Coverage probes are decided reachable / unreachable, never passed or
+    // failed. A skipped claim is one already recorded as reached earlier in
+    // the run (issue #6387).
+    const size_t reached = summary.reached_goals + summary.skipped_properties;
+    const size_t goals =
+      reached + summary.unreached_goals + summary.undecided_goals;
```
lucasccordeiro: `summary.simplified_properties` is dropped from this total. Probes whose guard the simplifier folded away are counted in `total_properties`/`passed_properties` on the non-coverage path but appear in none of `reached_goals` / `unreached_goals` / `undecided_goals`, so `Coverage goals: N` can be smaller than `Total Asserts` in the `[Coverage]` block a few lines later. A probe simplified away is a goal that was decided (unreachable) — it should be counted, most plausibly as unreached.

Separately, folding `skipped_properties` into `reached` is only correct because a skip means "already in reached_claims"; that is worth stating as an invariant here rather than left implicit, since `is_keep_verified` can change it.

src/esbmc/parseoptions/driver.cpp
```diff
@@ -391,29 +391,47 @@ int esbmc_parseoptionst::doit()
...
+    return 0;
+  }
+
+  // A coverage run has no verdict, so it has no failure exit code either: its
+  // result is the [Coverage] block closed by the completeness line. Printed
+  // here, once, so it lands after the last block (k-induction prints one per
+  // phase).
+  if (is_coverage)
+  {
+    report_coverage_completeness();
```
lucasccordeiro, line 431: This prints `COVERAGE ANALYSIS COMPLETE` even when nothing was measured. `--show-vcc` and `--program-only` return `P_SMTLIB` from `run_thread` (`bmc.cpp:1778`, `bmc.cpp:1790`) *before* `multi_property_check` runs, so no goal is ever recorded undecided and `cov_incomplete_reasons` stays empty. `esbmc file.c --assertion-coverage --show-vcc` therefore emits a bare `COVERAGE ANALYSIS COMPLETE` with no `[Coverage]` block at all — precisely the misleading final signal this PR sets out to remove.

Gate on "at least one goal was decided", or have `run_thread`'s early P_SMTLIB returns call `note_cov_incomplete`.

src/esbmc/parseoptions/driver.cpp
```diff
@@ -391,29 +391,47 @@ int esbmc_parseoptionst::doit()
...
     return 0;
```
lucasccordeiro, line 432: Unconditional 0 drops the `P_ERROR` / `P_SMTLIB` propagation that the `dead-code-check` branch immediately above keeps on purpose (issue #4495, and the comment there explains why). A coverage run whose solver crashed is now indistinguishable from a clean one for any script or CI job that checks the exit status. Suggest the same shape: `if (res == P_ERROR || res == P_SMTLIB) return res; return 0;`.

Minor, related: `res` now holds an exit code on the strategy path and an `smt_resultt` on the `do_bmc` path, and the dead-code branch compares it against `P_ERROR`/`P_SMTLIB` regardless. That only works today because `--dead-code-check` is rejected with every strategy flag. Worth splitting into two variables so the coupling is not load-bearing.

src/esbmc/parseoptions/bmc_strategy.cpp
```diff
@@ -458,7 +462,11 @@ int esbmc_parseoptionst::do_bmc_strategy(
     // falsification
     if (options.get_bool_option("falsification"))
     {
-      if (is_base_case_violated(options, goto_functions, k_step).is_true())
+      const bool violated =
+        is_base_case_violated(options, goto_functions, k_step).is_true();
+      // A coverage run has no verdict to falsify: keep unwinding so later k
+      // steps can still reach goals, and leave through the common exit.
+      if (violated && !is_coverage)
```
lucasccordeiro, line 467: This changes `--falsification` + coverage from "stop at the first k where the base case is violated" to "always run all 50 k steps" (`max-k-step` default is 50). Under coverage every probe is SAT, so the base case is violated at k=1 essentially always — the run now re-symexes and re-solves 50 times, and because `multi_property_check` calls `report_coverage` on the base-case path it prints 50 `[Coverage]` blocks.

It also means `--falsification` + coverage can *never* report `COMPLETE`: the loop always falls through to the `note_cov_incomplete("the run ended at the maximum k step ...")` below. `github_6387_2` runs exactly this combination but only asserts the absence of `VERIFICATION FAILED`, so neither effect is caught.

If the intent is "coverage has nothing to falsify", the cheaper and more honest fix is to stop the loop after the first step (the goals are already recorded) rather than exhausting the bound.

src/goto-programs/goto_functions.h
```diff
@@ -80,6 +82,19 @@ class goto_functionst
   static std::mutex reached_claims_mutex;
   static std::mutex reached_mul_claims_mutex;
 
+  // Coverage goals that never received a definitive SAT/UNSAT answer, and
+  // why. A percentage computed while these are outstanding understates the
+  // real coverage, so report_coverage qualifies the number instead of
+  // printing one that looks measured.
+  static std::atomic<size_t> undecided_cov_goals;
```
lucasccordeiro: These are never reset. Under `--k-induction`/`--incremental-bmc` a fresh `bmct` runs per k step, so a goal that stays undecided across steps is counted once per step and `COVERAGE ANALYSIS INCOMPLETE: N goal(s) undecided` can exceed the total number of goals. Please reset (or de-duplicate by claim signature rather than counting) at the start of each `multi_property_check`.

Layering: nothing under `goto-programs/` reads these — they are BMC reporting state used only by `bmc.cpp` and `bmc_strategy.cpp`. Putting them here also pulls `<atomic>` and `<set>` into a widely included header. File-scope statics in `bmc.cpp` behind the accessors already declared in `bmc.h` would be a better home.

regression/goto-coverage/branch_cov_11/test.desc
```diff
@@ -1,4 +1,5 @@
-KNOWNBUG
+CORE
 main.cpp
 --branch-coverage-claims --unwind 3 --no-unwinding-assertions
```
lucasccordeiro, line 3: This is the core soundness gap in the new signal. Coverage modes auto-enable `no-unwinding-assertions` whenever `--unwind` is given (`process_goto_program.cpp:502`, `:528`, `:583`, `:604`, `:639`), so a truncated run cannot tell a genuinely unreachable goal from one that lies beyond the bound — yet this test now expects `Branch Coverage: 83.33` followed by `COVERAGE ANALYSIS COMPLETE`. By the PR's own definition ("every coverage goal received a definitive reachable/unreachable answer, so the percentages are exact") this run is `INCOMPLETE`: `83.33%` is a lower bound.

`note_cov_incomplete` should fire when unwinding assertions are disabled (or the loops were otherwise not proven fully unwound) and at least one goal came back UNREACHED. That covers the common `--unwind N` case, which is most bounded coverage runs.

Separately, flipping `KNOWNBUG` → `CORE` here changes what the test pins: the original recorded a wrong *verdict*. If the underlying defect is now considered fixed, say so in the commit message; if it is merely no longer observable because verdicts are gone, the KNOWNBUG should be re-pointed at whatever still misbehaves rather than repurposed.

regression/goto-coverage/github_6387_7/test.desc
```diff
@@ -0,0 +1,5 @@
+KNOWNBUG
```
lucasccordeiro: This test has no coverage flag (`--multi-property --sarif-output`) and expects `VERIFICATION FAILED` — it is unrelated to #6387 and to anything this PR changes. Adding a `KNOWNBUG` with no reference to an issue describing the SARIF defect leaves the next reader with no way to tell what is expected to change. Please move it to its own PR with a filed issue, or drop it.

lucasccordeiro (commented):
Review as maintainer. The direction is right — a coverage run should not emit a verification verdict — but the new `COVERAGE ANALYSIS COMPLETE` / `INCOMPLETE` signal is not yet sound, and the accounting behind it disagrees with itself in the same output. Details inline.

Missing tests I would want before merge:

- `--assertion-coverage --show-vcc` and `--assertion-coverage --program-only`: currently print `COVERAGE ANALYSIS COMPLETE` with no `[Coverage]` block at all.
- A bounded coverage run with unreached goals (`--branch-coverage --unwind N`): must print `INCOMPLETE`, not `COMPLETE` (see comment on `branch_cov_11`).
- A coverage run under a solver error (P_ERROR): exit code and diagnostic.
- `--assertion-coverage --k-induction` over ≥2 k steps, asserting the `N goal(s) undecided` count does not exceed the number of goals.
- A regression pinning that `--dead-code-check` output/verdict/exit code is unchanged by this PR (the `is_goto_cov` refactor touches its paths too).

regression/goto-coverage/github_6387_5/test.desc
```diff
@@ -0,0 +1,5 @@
+CORE
```
lucasccordeiro: Seven of the eleven new tests use a byte-identical `main.cpp`, and this one adds only `\A(?![\s\S]*[0-9]+ failed)` over `github_6387_1`'s `VERIFICATION FAILED` check — both are subsumed by a single test asserting the full new summary line. Consider collapsing `_1`, `_2`, `_5` and `_10` into one or two tests with the union of the regexes; each duplicate directory is a full `esbmc` invocation on CI.

Also: none of the new tests assert the `Coverage goals: N ✓ M reached` summary line itself, which is the main user-visible output this PR introduces.

ChenfengWei0:
● Update:

  - Bounded runs no longer report COMPLETE — a run truncated at the unwinding bound is now INCOMPLETE.
  - `--show-vcc` / `--program-only` no longer print a coverage verdict.
  - Solver failures are reported again.
  - Fixed contradictory goal counts under `--multi-fail-fast`; undecided goals now get their own line.
  - Completeness state is reset per pass instead of accumulating across k steps.
  - Coverage + `--falsification` runs one pass again instead of 51.
  - Unified the two definitions of "coverage run".
  - Added `github_6387_13`–`_17` for `--show-vcc`, `--program-only`, k-induction, bounded coverage, and `--dead-code-check`.
  - Updated 12 existing tests that were claiming COMPLETE over truncated runs.

PR #6437, opened 2026-07-27
[solver] widen array index domain to fit one-past-end index (#6399)

### LINKED ISSUE #6399: [contracts] is_fresh + --loop-invariant-check + --replace-call-with-contract fails at power-of-two trip counts
## Summary

A wrapper function that loops `K` times calling an `__ESBMC_is_fresh`-guarded
callee (replaced via `--replace-call-with-contract`), with the wrapper's own
loop cut via `__ESBMC_loop_invariant` + `--loop-invariant-check`, fails
verification (`VERIFICATION FAILED`) when `K` is a power of two (1, 2, 4, 8,
16 all tested and fail), but succeeds for every other trip count tested (3,
5, 6, 7, 9, 10, 11, 12 all succeed). Both the "fails" and "succeeds" cases
are for a contract that is genuinely true -- confirmed by re-running the
identical `K` values with the loop fully unrolled instead of cut (drop
`--loop-invariant-check`, raise `--unwind` to cover `K`), which succeeds at
every `K` tested (1, 2, 4, 8), including the ones that fail under
`--loop-invariant-check`. So this is a false `VERIFICATION FAILED`, not a
real property violation.

Also confirmed specific to `is_fresh`: replacing `__ESBMC_is_fresh(e, sizeof(Elem))`
with `e != ((void*)0)` in the identical structure (same loop, same
`--loop-invariant-check` + `--replace-call-with-contract`) succeeds at every
`K` tested (1, 2, 3, 4, 8) -- no power-of-two pattern at all. So the bug
needs all three ingredients together: `is_fresh` on the replaced callee, the
caller's loop cut via `--loop-invariant-check`, and a power-of-two trip count.

## Minimal reproducer

```c
typedef struct { int x; } Elem;
typedef struct { Elem vec[K_SWEEP]; } Vec;

void touch(Elem *e)
{
  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(Elem)));
  __ESBMC_ensures(e->x == 0);
  __ESBMC_assigns(e->x);
  e->x = 0;
}

void touch_all(Vec *v)
{
  unsigned k, j;
  __ESBMC_requires(__ESBMC_is_fresh(v, sizeof(Vec)));
  __ESBMC_assigns(v->vec);
  __ESBMC_ensures(__ESBMC_forall(&j, !(j < K_SWEEP) || v->vec[j].x == 0));

  __ESBMC_loop_invariant(k <= K_SWEEP && __ESBMC_forall(&j, !(j < k) || v->vec[j].x == 0));
  for (k = 0; k < K_SWEEP; k++)
  {
    touch(&v->vec[k]);
  }
}

int main(void) { return 0; }
```

`K_SWEEP` substituted with a literal constant (1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
11, 12, 16 all tested).

Run as:
```
esbmc repro.c --function touch_all --enforce-contract touch_all \
  --no-pointer-check --unwind 8 --loop-invariant-check --z3 \
  --replace-call-with-contract touch
```
(`--unwind` raised to 40 for `K_SWEEP=16`; result unaffected by unwind bound
once above `K_SWEEP`, since the loop is meant to be cut, not unrolled.)

## Results table

| K | Result (is_fresh + loop-invariant-check) | Result (is_fresh, full unwind, no loop-invariant-check) | Result (`!= NULL` + loop-invariant-check) |
|---|---|---|---|
| 1 | FAILED | SUCCESSFUL | SUCCESSFUL |
| 2 | FAILED | SUCCESSFUL | SUCCESSFUL |
| 3 | SUCCESSFUL | SUCCESSFUL | SUCCESSFUL |
| 4 | FAILED | SUCCESSFUL | SUCCESSFUL |
| 5 | SUCCESSFUL | (not re-tested) | (not re-tested) |
| 6 | SUCCESSFUL | (not re-tested) | (not re-tested) |
| 7 | SUCCESSFUL | (not re-tested) | (not re-tested) |
| 8 | FAILED | SUCCESSFUL | SUCCESSFUL |
| 9-12 | SUCCESSFUL | (not re-tested) | (not re-tested) |
| 16 | FAILED | (not re-tested) | (not re-tested) |

Every power of two tested (1, 2, 4, 8, 16) fails under `is_fresh` +
`--loop-invariant-check`; every non-power-of-two tested (3, 5, 6, 7, 9-12)
succeeds. `K=32` timed out under `--loop-invariant-check` at a 2-minute cap
(not conclusive either way, not included above).

## Why this matters

This surfaced while building a real-world benchmark: a fresh-vector wrapper
(`mlk_polyvec_reduce`, looping over `MLKEM_K` calling `mlk_poly_reduce_c`
per element, `MLKEM_K` in {2, 3, 4} for the real ML-KEM parameter sets) with
the faithful mlkem-native contract shape (`is_fresh` + `forall` +
`loop_invariant`, mechanically translated from `memory_no_alias` +
`array_bound` + CBMC's `loop_entry`-based invariants). Two of the three real
parameter values (`MLKEM_K` = 2 and 4) hit this bug; only `MLKEM_K` = 3
happens to avoid it.

## Environment

- ESBMC built from `origin/master` at `ebcaed11bf` (includes #6326/#6358/
  #6365/#6385, i.e. this is not a re-occurrence of any of those).
- OS: macOS, x86_64. Solver: Z3 4.16.0 (`--z3`, required for
  `--loop-invariant-check` per existing docs/#6217).

LukeW1999:
Fixes #6399.

### Symptom

`__ESBMC_is_fresh` + `--loop-invariant-check` + `--replace-call-with-contract`
reported a false `VERIFICATION FAILED` whenever the fresh object's byte size
was a power of two, and verified at every other size. The report framed this as
a trip-count parity; decoupling element size from element count shows it is a
parity on the object's **byte size**.

### Root cause

`size_to_bit_width()` (`src/solvers/smt/smt_solver.cpp`) sizes an array's index
domain:

```cpp
while (domwidth != 0x8000000000000000ULL && domwidth < sz) { domwidth <<= 1; dombits++; }
```

An `n`-element array gets `ceil(log2(n))` bits. For `n` a power of two that is
*exactly* `log2(n)`: the domain represents `0..n-1` and cannot represent `n`
itself. But `n` is the one-past-the-end index — a valid pointer value in C, and
one that contract replacement plus a cut loop does construct. It wraps to `0`
and aliases element 0, so the re-established invariant stops connecting to the
`ensures`.

Every non-power-of-two size takes one more iteration of the shift loop and ends
up with a spare bit. That is the entire parity.

### Fix

`<` becomes `<=`, so the domain is wide enough to hold `n`. It widens
power-of-two sizes only; every other size already had the extra bit, so nothing
else changes.

### Evidence

A/B on a single binary, one character apart, `contracts.cpp` untouched:

| shape | `sizeof` | `< sz` | `<= sz` |
|---|---|---|---|
| aggregate, K=1 | 4 (2²) | FAILED | SUCCESSFUL |
| aggregate, K=2 | 8 (2³) | FAILED | SUCCESSFUL |
| aggregate, K=3 | 12 | SUCCESSFUL | SUCCESSFUL |
| aggregate, K=4 | 16 (2⁴) | FAILED | SUCCESSFUL |
| aggregate, K=8 | 32 (2⁵) | FAILED | SUCCESSFUL |
| aggregate, K=16 | 64 (2⁶) | FAILED | SUCCESSFUL |
| **scalar `int*`** | 16 (2⁴) | FAILED | SUCCESSFUL |
| scalar `int*` | 20 | SUCCESSFUL | SUCCESSFUL |
| negative control | 16 (2⁴) | FAILED | **FAILED** |

The negative controls still fail, so the power-of-two cases are not becoming
vacuous.

Solve time is flat across sizes afterwards (0.23–0.55s for 128–512 byte
objects); the previous run showed no such regularity.

### Tests

- `is_fresh_loopinv_poweroftwo_{pass,fail}` — aggregate shape, the reporter's.
- `is_fresh_loopinv_scalar_poweroftwo_{pass,fail}` — plain `int` buffer.

The scalar pair is the point of the second commit on this PR's history: the
first approach taken here was a contract-side workaround that retyped aggregate
`is_fresh` allocations, and as @lucasccordeiro showed in review, that cannot
reach a plain `int` buffer. Chasing why led to the domain width. The contract
change is gone; `contracts.cpp` is untouched.

### Scope and risk

The change is global — it affects the index domain of every constant-size
array, not just contract code. Two risks seemed worth closing before asking
anyone to merge it.

**Does a wider domain miss out-of-bounds accesses?** No. The worry was that if
the bounds assertion were emitted *after* the index is cast to the domain
width, then under the old code an index of `16` into `int a[16]` would wrap to
`0` and the overflow would have been silently missed — a false negative, and a
far worse bug than the one being fixed. Measured on both builds, out-of-bounds
writes are caught identically and in-bounds writes still verify:

| | `< sz` | `<= sz` |
|---|---|---|
| `int a[16]` (2⁴), out-of-bounds write | FAILED | FAILED |
| `int a[20]`, out-of-bounds write | FAILED | FAILED |
| `int a[16]`, in-bounds write | SUCCESSFUL | SUCCESSFUL |

So the old behaviour was wrong only in the direction of false positives, and
this PR is a false-positive fix, not a soundness fix. (The likely reason is
that the assertion precedes the cast, but that is an inference from the
behaviour above — I did not confirm it in the code.)

**What does the extra bit cost?** An array filled by a loop, then read at a
symbolic index, `--unwind N+1`:

| N | 2ⁿ? | `< sz` | `<= sz` | Δ |
|---|---|---|---|---|
| 128 | yes | 0.095s | 0.117s | +23% |
| 200 | no | 0.260s | 0.291s | +12% |
| 256 | yes | 0.665s | 0.723s | +8.7% |
| 500 | no | 5.183s | 5.949s | +15% |
| 512 | yes | 5.904s | 6.325s | +7.1% |
| 1000 | no | 53.394s | 58.757s | +10% |
| 1024 | yes | 56.474s | 61.916s | +9.6% |

Read this carefully: the non-power-of-two rows are sizes whose domain width is
*unchanged* by this patch, and they move by the same 10–15%. The deltas are
therefore dominated by run-to-run noise, and the real cost of the extra bit
sits below that noise floor. Single runs, not medians — the honest summary is
"no measurable regression, bounded above by ~10%", and the scaling behaviour is
unchanged (both builds grow with the unwind count, not the index width).

**Regression runs on this branch**, 3314 tests, all green: `esbmc/` 1524/1524
(full label, not sampled), the array-heavy labels `cbmc/` + `cstd/` + `floats/`
+ `k-induction/` + `llvm/` + `csmith/` + `bitwuzla/` + `esbmc-cpp/cpp/` 1395,
`function_contract` 291/291, memory suites 104/104. The `bitwuzla/` label
passing matters here: it shows the parity was not a Z3 artefact.

**What I do not claim.** Plain `malloc` plus a cut loop, with no contracts
involved, does **not** reproduce the original failure. So the narrow domain is
the enabling condition rather than the whole causal chain — the contract path
additionally has to construct the one-past-the-end index. The domain arithmetic
and the A/B above are solid; I have not traced which step builds that index,
and I would rather say so than imply a complete story.

### DIFF
--- a/regression/function_contract/is_fresh_loopinv_poweroftwo_fail/test.desc
+++ b/regression/function_contract/is_fresh_loopinv_poweroftwo_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
+^VERIFICATION FAILED$
--- a/regression/function_contract/is_fresh_loopinv_poweroftwo_pass/test.desc
+++ b/regression/function_contract/is_fresh_loopinv_poweroftwo_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
+^VERIFICATION SUCCESSFUL$
--- a/regression/function_contract/is_fresh_loopinv_poweroftwo_fail/main.c
+++ b/regression/function_contract/is_fresh_loopinv_poweroftwo_fail/main.c
@@ -1,0 +1,30 @@
+/* is_fresh_loopinv_poweroftwo_fail (issue #6399):
+ * Negative control for is_fresh_loopinv_poweroftwo_pass. The loop only touches
+ * vec[0..2], so vec[3] is left havoc'd (nondet) and the ensures over all four
+ * elements must FAIL. Confirms the #6399 fix (a wider array index domain) does
+ * not make the power-of-two case vacuously verify.
+ */
+typedef struct { int x; } Elem;
+typedef struct { Elem vec[4]; } Vec;
+
+void touch(Elem *e)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(Elem)));
+  __ESBMC_ensures(e->x == 0);
+  __ESBMC_assigns(e->x);
+  e->x = 0;
+}
+
+void touch_all(Vec *v)
+{
+  unsigned k, j;
+  __ESBMC_requires(__ESBMC_is_fresh(v, sizeof(Vec)));
+  __ESBMC_assigns(v->vec);
+  __ESBMC_ensures(__ESBMC_forall(&j, !(j < 4) || v->vec[j].x == 0));
+
+  __ESBMC_loop_invariant(k <= 3 && __ESBMC_forall(&j, !(j < k) || v->vec[j].x == 0));
+  for (k = 0; k < 3; k++)
+    touch(&v->vec[k]);
+}
+
+int main(void) { return 0; }
--- a/regression/function_contract/is_fresh_loopinv_poweroftwo_pass/main.c
+++ b/regression/function_contract/is_fresh_loopinv_poweroftwo_pass/main.c
@@ -1,0 +1,37 @@
+/* is_fresh_loopinv_poweroftwo_pass (issue #6399):
+ * A wrapper loops K times over an __ESBMC_is_fresh struct, calling an
+ * is_fresh-guarded callee (replaced via --replace-call-with-contract) per
+ * element, with its own loop cut by __ESBMC_loop_invariant. This used to
+ * report a false VERIFICATION FAILED whenever sizeof(Vec) was a power of two
+ * (K = 1,2,4,8,16 -> 4,8,16,32,64 bytes). size_to_bit_width() gave an
+ * n-element array ceil(log2(n)) index bits, which for a power-of-two n is
+ * exactly log2(n): the one-past-the-end index n was then unrepresentable and
+ * wrapped to 0, aliasing element 0. Every other size had a spare bit, hence
+ * the parity. The domain is now wide enough to hold n itself.
+ * sizeof(Vec) = 4*4 = 16 (power of two) here, the historically-failing shape.
+ * The contract is genuinely true.
+ */
+typedef struct { int x; } Elem;
+typedef struct { Elem vec[4]; } Vec;
+
+void touch(Elem *e)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(Elem)));
+  __ESBMC_ensures(e->x == 0);
+  __ESBMC_assigns(e->x);
+  e->x = 0;
+}
+
+void touch_all(Vec *v)
+{
+  unsigned k, j;
+  __ESBMC_requires(__ESBMC_is_fresh(v, sizeof(Vec)));
+  __ESBMC_assigns(v->vec);
+  __ESBMC_ensures(__ESBMC_forall(&j, !(j < 4) || v->vec[j].x == 0));
+
+  __ESBMC_loop_invariant(k <= 4 && __ESBMC_forall(&j, !(j < k) || v->vec[j].x == 0));
+  for (k = 0; k < 4; k++)
+    touch(&v->vec[k]);
+}
+
+int main(void) { return 0; }
--- a/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_fail/main.c
+++ b/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_fail/main.c
@@ -1,0 +1,27 @@
+/* is_fresh_loopinv_scalar_poweroftwo_fail (issue #6399):
+ * Negative control for is_fresh_loopinv_scalar_poweroftwo_pass. The loop only
+ * touches v[0..2], so v[3] stays havoc'd and the ensures over all four
+ * elements must FAIL -- the widened index domain must not make the
+ * power-of-two case vacuously verify.
+ */
+void touch(int *e)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(int)));
+  __ESBMC_ensures(*e == 0);
+  __ESBMC_assigns(*e);
+  *e = 0;
+}
+
+void touch_all(int *v)
+{
+  unsigned k, j;
+  __ESBMC_requires(__ESBMC_is_fresh(v, 4 * sizeof(int)));
+  __ESBMC_assigns(*v);
+  __ESBMC_ensures(__ESBMC_forall(&j, !(j < 4) || v[j] == 0));
+
+  __ESBMC_loop_invariant(k <= 3 && __ESBMC_forall(&j, !(j < k) || v[j] == 0));
+  for (k = 0; k < 3; k++)
+    touch(&v[k]);
+}
+
+int main(void) { return 0; }
--- a/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_fail/test.desc
+++ b/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
+^VERIFICATION FAILED$
--- a/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_pass/main.c
+++ b/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_pass/main.c
@@ -1,0 +1,30 @@
+/* is_fresh_loopinv_scalar_poweroftwo_pass (issue #6399):
+ * The scalar-pointee counterpart of is_fresh_loopinv_poweroftwo_pass. The
+ * fresh object is a plain int buffer, so nothing here is a struct: 16 bytes
+ * (2^4) used to report a false VERIFICATION FAILED while 20 bytes verified,
+ * the same power-of-two parity coming from the array index domain being one
+ * bit too narrow to hold the one-past-the-end index. Kept as a separate test
+ * because a contract-side fix that retypes aggregate is_fresh objects cannot
+ * reach this shape -- only widening the domain does.
+ */
+void touch(int *e)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(int)));
+  __ESBMC_ensures(*e == 0);
+  __ESBMC_assigns(*e);
+  *e = 0;
+}
+
+void touch_all(int *v)
+{
+  unsigned k, j;
+  __ESBMC_requires(__ESBMC_is_fresh(v, 4 * sizeof(int)));
+  __ESBMC_assigns(*v);
+  __ESBMC_ensures(__ESBMC_forall(&j, !(j < 4) || v[j] == 0));
+
+  __ESBMC_loop_invariant(k <= 4 && __ESBMC_forall(&j, !(j < k) || v[j] == 0));
+  for (k = 0; k < 4; k++)
+    touch(&v[k]);
+}
+
+int main(void) { return 0; }
--- a/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_pass/test.desc
+++ b/regression/function_contract/is_fresh_loopinv_scalar_poweroftwo_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
+^VERIFICATION SUCCESSFUL$
--- a/src/solvers/smt/smt_solver.cpp
+++ b/src/solvers/smt/smt_solver.cpp
@@ -2521,9 +2521,11 @@ static unsigned long size_to_bit_width(unsigned long sz)
   uint64_t domwidth = 2;
   unsigned int dombits = 1;
 
-  // Shift domwidth up until it's either larger or equal to sz, or we risk
-  // overflowing.
-  while (domwidth != 0x8000000000000000ULL && domwidth < sz)
+  // Shift domwidth up until it is strictly larger than sz, or we risk
+  // overflowing. Strictly larger, not just equal: sz itself is the
+  // one-past-the-end index, a valid pointer value in C, and a domain that
+  // cannot represent it wraps it to 0 and aliases element 0 (#6399).
+  while (domwidth != 0x8000000000000000ULL && domwidth <= sz)
   {
     domwidth <<= 1;
     dombits++;


src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
...
+      {
+        type2tc pointee = ns.follow(to_pointer_type(ptr_var->type).subtype);
+        // Only aggregate (struct/union) pointees are retyped. The #6399
+        // mis-encoding is triggered by a power-of-two-sized dynamic *byte*
+        // buffer that holds an array the loop-invariant havoc rewrites;
+        // modelling the object with its real aggregate type replaces the flat
+        // byte array with structured fields and sidesteps the bad path. Scalar
+        // pointees keep the u8 buffer, so their bounds diagnostics (e.g.
+        // "array bounds violated" vs "object out of bounds") are unchanged.
+        if (is_struct_type(pointee) || is_union_type(pointee))
```
lucasccordeiro: Restricting the retype to struct/union leaves #6399 reproducible. I ran the scalar analogue of the new `_pass` test against master:

```c
void touch(int *e)
{
  __ESBMC_requires(__ESBMC_is_fresh(e, sizeof(int)));
  __ESBMC_ensures(*e == 0);
  __ESBMC_assigns(*e);
  *e = 0;
}
void touch_all(int *v)
{
  unsigned k, j;
  __ESBMC_requires(__ESBMC_is_fresh(v, 4 * sizeof(int)));   /* 16 bytes */
  __ESBMC_assigns(*v);
  __ESBMC_ensures(__ESBMC_forall(&j, !(j < 4) || v[j] == 0));
  __ESBMC_loop_invariant(k <= 4 && __ESBMC_forall(&j, !(j < k) || v[j] == 0));
  for (k = 0; k < 4; k++)
    touch(&v[k]);
}
```

- `4 * sizeof(int)` (16 bytes, 2^4) → `VERIFICATION FAILED`
- `5 * sizeof(int)` (20 bytes), loop and ensures unchanged → `VERIFICATION SUCCESSFUL`

Same parity, same false negative, and `int` is not struct/union so this patch cannot change it. `void *` pointees and function pointees are in the same position.

So either the retype should cover any pointee whose size is known (the `u8` special case only needs to survive for `void`/opaque), or #6399 stays open and this lands as a partial mitigation with a `KNOWNBUG` test pinning the scalar shape. Please don't close the issue on this patch.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
...
+        {
+          expr2tc sz = size_expr;
+          simplify(sz);
+          if (is_constant_int2t(sz))
```
lucasccordeiro: Requiring a constant excludes the most common array-contract idiom, `__ESBMC_is_fresh(a, n * sizeof(Elem))` with `n` a parameter: `simplify` leaves it non-constant, so it silently takes the `u8` fallback and stays on the mis-encoded path.

A constant is not actually needed — `elem_count` can be `div2tc(size_expr, constant(pbytes))` and left for symex to fold when it can. That covers the symbolic case at no cost to the constant one. If truncation on a non-multiple size is the worry, round up rather than falling back.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
...
+          {
+            BigInt pbytes = type_byte_size(pointee, &ns);
+            if (pbytes > 0 && (to_constant_int2t(sz).value % pbytes) == 0)
```
lucasccordeiro: Two gaps in this guard.

1. Exact-multiple only: `__ESBMC_is_fresh(h, sizeof(Hdr) + n)` (header + trailing payload, a normal C shape) falls back to `u8` and stays broken at power-of-two totals.
2. Sign: `to_constant_int2t(sz).value` is only unsigned by convention. A negative constant satisfies `% pbytes == 0` (e.g. `-16 % 4`), giving a negative `elem_count` fed straight into `sideeffect2t::malloc`. Add `value > 0` to the condition.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
...
+            {
+              elem_type = pointee;
+              elem_count =
+                constant_int2tc(sz->type, to_constant_int2t(sz).value / pbytes);
```
lucasccordeiro: The description says the element-count division has no observable effect and therefore no test. It does have one — allocation bounds are observable through the access path. Verified against master:

```c
typedef struct { int a[3]; } T;   /* 12 bytes */
void f(T *p)
{
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(T)));
  __ESBMC_assigns(*p);
  char *c = (char *)p;
  c[11] = 1;   /* last valid byte */
  c[12] = 1;   /* one past the end */
}
```

`esbmc bounds.c --function f --enforce-contract f --z3` → `dereference failure: array bounds violated: heap object` / `VERIFICATION FAILED`; with the `c[12]` line removed → `VERIFICATION SUCCESSFUL`.

That pair pins the allocation size exactly, and is precisely the boundary probe (byte 191 vs 192) described in the follow-up section. Please add it as a `{pass,fail}` pair — without it, a future edit that drops the `/ pbytes` silently re-introduces the `sizeof(T)^2` over-reservation and nothing in the suite notices.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
...
+      expr2tc elem_count = size_expr;
+      if (is_pointer_type(ptr_var->type))
+      {
+        type2tc pointee = ns.follow(to_pointer_type(ptr_var->type).subtype);
+        // Only aggregate (struct/union) pointees are retyped. The #6399
+        // mis-encoding is triggered by a power-of-two-sized dynamic *byte*
+        // buffer that holds an array the loop-invariant havoc rewrites;
+        // modelling the object with its real aggregate type replaces the flat
+        // byte array with structured fields and sidesteps the bad path. Scalar
+        // pointees keep the u8 buffer, so their bounds diagnostics (e.g.
```
lucasccordeiro: This concedes the aggregate path's bounds diagnostic changes, but the only test pinning that wording, `is_fresh_requires_oob_fail`, uses `int *p` — i.e. the scalar path this patch does not touch. So the message that actually changes has no coverage in either direction.

Add the aggregate counterpart of `is_fresh_requires_oob_fail` (struct pointee, out-of-range element access) pinning the post-patch wording, so the change is recorded rather than discovered by a downstream user.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1067,15 +1067,56 @@ goto_programt code_contractst::generate_checking_wrapper(
   // Emit the malloc + non-null assume for one resolved is_fresh pointer lvalue.
   auto emit_is_fresh_alloc =
     [&](const expr2tc &ptr_var, const expr2tc &size_expr) {
-      // Use u8 as the alloc element type so that size_expr (in bytes) equals
-      // the element count.  symex_mem casts the u8* result to lhs->type.
-      type2tc char_type = get_uint8_type();
+      // Allocate an object of the pointer's *pointee* type when the requested
```
lucasccordeiro: 16 lines of comment before the first statement, and the #6399 rationale is then restated at lines 1091-1097. Per `CLAUDE.md` ("write few comments", one line beats a paragraph), condense to the two facts the code cannot convey — the malloc `size` field is an element count of `alloctype`, and why only aggregates are retyped — and cite #6399 once.

Related: `emit_is_fresh_alloc` is now ~50 lines with the type/count selection nested three deep inside it. Lifting that decision into a named helper (`select_is_fresh_alloc(ptr_type, size_expr, ns) -> {type, count}`) would make both the lambda and the rules readable, and would give the shape-coverage gaps above an obvious single place to be fixed.

regression/function_contract/is_fresh_loopinv_poweroftwo_pass/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
```
lucasccordeiro, line 3: Two flag concerns.

`--z3`: only 2 of the 285 existing `function_contract` tests pin a solver, and the description reports the same behaviour under Z3 and Bitwuzla — so this ties the test to a Z3-enabled build for no stated reason and it will not run on a Bitwuzla-only configuration. Drop it and use the default.

`--no-pointer-check`: this suppresses exactly the checks that would catch a bad allocation shape or size, in a PR whose entire subject is the allocation shape and size. If the test does not verify with pointer checks on, that is worth understanding before merge rather than masking; if it does, drop the flag.

regression/function_contract/is_fresh_loopinv_poweroftwo_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone does not make this a negative control. On the unpatched tree this program also reports FAILED — for the #6399 mis-encoding, not for the intended reason (`vec[3]` left havoc'd) — so the test passes identically before and after the fix and cannot detect the power-of-two case becoming vacuously true.

Pin the property, e.g. add a regex on the violated ensures/postcondition line, as `is_fresh_requires_oob_fail` does with its `dereference failure: array bounds violated` line.

lucasccordeiro (commented):
The root-cause analysis is convincing and the retyping is a reasonable mitigation, but as written this does not fix #6399 — it fixes one shape of it. I reproduced the same power-of-two parity with a **scalar** pointee, which this patch deliberately leaves on the `u8` path (details inline). The `--replace-call-with-contract` + `--loop-invariant-check` byte-array mis-encoding is still there; only aggregate, constant, exact-multiple sizes route around it.

Two things I would want before merge:

1. Either extend the retyping to the remaining shapes, or keep #6399 open and land this as a partial mitigation with a `KNOWNBUG` test pinning the scalar case. Closing #6399 on this patch would lose the bug.
2. Add the allocation-size regression the description says is untestable — it is testable (inline, with a verified reproducer).

Commands used for the reproductions below, against master (`0ac82211d9`), aarch64-macos, ESBMC 8.4.0:
`esbmc <file> --function touch_all --enforce-contract touch_all --no-pointer-check --unwind 8 --loop-invariant-check --z3 --replace-call-with-contract touch`

LukeW1999:
Thank you — the objection was right and following it up found the actual cause.
The PR is now a different change: `contracts.cpp` is untouched and the fix is
one character in the solver.

**Root cause.** `size_to_bit_width()` sizes an array's index domain with
`domwidth < sz`, giving an `n`-element array `ceil(log2(n))` bits. For `n` a
power of two that is exactly `log2(n)`: the domain holds `0..n-1` and cannot
represent `n` itself. `n` is the one-past-the-end index — a valid pointer value
in C, and one the contract-replacement plus cut-loop path constructs — so it
wraps to 0 and aliases element 0, and the re-established invariant stops
connecting to the `ensures`. Every non-power-of-two size takes one more
iteration of the shift loop and gets a spare bit. That is the whole parity, and
it is arithmetic rather than anything about byte arrays or the havoc.

`<` becomes `<=`. Only power-of-two sizes widen.

**Your point 1.** Confirmed and now fixed rather than pinned. Your scalar
reproducer at 16 bytes goes to `SUCCESSFUL`, 20 bytes stays `SUCCESSFUL`, and I
added it as a CORE pair (`is_fresh_loopinv_scalar_poweroftwo_{pass,fail}`)
rather than a `KNOWNBUG` — a plain `int` buffer is exactly the shape the
retyping could never reach, so it is the test that distinguishes the two
approaches. A/B on one binary, contracts untouched, across your table and the
reporter's: every power of two flips FAILED → SUCCESSFUL, 12 and 20 bytes stay
SUCCESSFUL, and both negative controls still FAIL.

**Your point 2.** I have to push back on `div2tc`, though it is moot now that
the retyping is gone. The malloc side effect's `size` field has two consumers
that read it in different units: `symex_mem` builds the object as
`alloctype[size]` (element count) while `track_new_pointer` stores the same
value into `__ESBMC_alloc_size` (bytes). They agree only while `alloctype` is
one byte wide. Passing an element count therefore makes the recorded allocation
size too small by `sizeof(T)` — `__ESBMC_is_fresh(p, 3 * sizeof(Elem))` then
reports `p[0]` out of bounds. I hit this by adding exactly the division you
suggested and measuring it against master. Worth knowing independently of this
PR; that field's semantics are not documented anywhere.

**Your point 3.** Moot with the retyping gone.

**Your point 4.** You were right that the allocation size is observable, and
right about the underlying worry. But the probe as written would not have
caught the regression: `sizeof(T)` gives an element count of 1, which
`symex_mem` special-cases into a bare `T` where both readings coincide. The
shape that catches it is `n * sizeof(T)` with `n > 1`. In the end the honest
answer was not to add a test protecting the division but to delete the
division.

**What I do not claim.** Plain `malloc` plus a cut loop, with no contracts, does
not reproduce the failure. So the narrow domain is the enabling condition, not
the whole causal chain — the contract path additionally has to construct the
one-past-the-end index, and I have not traced which step does. The domain
arithmetic and the A/B are solid; the last link is not.

The change is global, so CI is the real gate here. Local: `function_contract`
291/291 with the new tests, memory suites 104/104, two independent 300-test
samples of `esbmc/` clean, full run going now.

LukeW1999:
The two risks I flagged in the description are now closed; details are in the
updated description, summary here.

**A wider domain does not miss out-of-bounds accesses.** The concern was that
the bounds assertion might be emitted *after* the index is cast to the domain
width — in which case the old code would have silently missed an index of `16`
into `int a[16]`, a false negative rather than the false positive we are
fixing. It does not: the assertion precedes the cast, and out-of-bounds writes
are caught identically under both builds. This PR fixes false positives only.

**The extra bit costs nothing measurable.** On a loop-filled array read at a
symbolic index, the deltas are +7% to +23% — but the non-power-of-two sizes,
whose domain width this patch does not touch, move by the same amount. The
measurement is noise-dominated and the true cost sits below it. Scaling
behaviour is unchanged.

**Regression runs**, 3314 tests, all green: `esbmc/` 1524/1524 (full label),
the array-heavy labels 1395, `function_contract` 291/291, memory suites
104/104. `bitwuzla/` passing rules out a Z3-specific artefact.

One thing stays open and I would rather flag it than let it pass: plain
`malloc` plus a cut loop, no contracts, does not reproduce the original
failure. The narrow domain is the enabling condition, not the entire causal
chain — something in the contract path still has to construct the
one-past-the-end index, and I have not traced which step. The domain
arithmetic and the A/B are solid; that last link is not.

PR #6442, opened 2026-07-27
[ir-ieee] Preserve sign metadata for negative zero constants

ShemaAlosaimi:
## Summary

This PR preserves the IEEE-754 sign of literal negative-zero floating-point constants in the `--ir-ieee` encoding.

In the integer/real encoding, both `+0.0` and `-0.0` have the same numerical SMT real value, `0`. Previously, converting a literal `-0.0` discarded its sign in the SMT representation, so sign-sensitive operations reaching the `--ir-ieee` encoding could not distinguish it from `+0.0`.

This PR extends the existing negative-zero metadata mechanism to literal `-0.0` constants while keeping the metadata attached to an identity-distinct SMT AST.

## Motivation

The previous negative-zero work introduced side metadata for negative zeros produced by subnormal flush-to-zero. Consumers such as division by zero and `signbit` can use that metadata to recover the sign that is absent from the underlying SMT real value.

Literal `-0.0` constants were still converted directly to an ordinary real zero:

```cpp
if (thereal.value.is_zero())
  return mk_smt_real("0");
```

As a result, operations that reached the SMT encoding could not distinguish a literal `-0.0` from `+0.0`.

For example, IEEE-754 requires:

```c
positive_value / -0.0 == -infinity
positive_value / +0.0 == +infinity
```

The negative-zero metadata map is keyed by SMT AST identity, so attaching metadata directly to `mk_smt_real("0")` would also rely on an unsafe assumption that repeated real constants always receive distinct wrapper ASTs. This PR avoids that dependency.

## Changes

- When converting a literal `-0.0`:
  - create a fresh real AST;
  - constrain that AST equal to real zero;
  - attach the existing negative-zero metadata to the fresh AST rather than to the ordinary `mk_smt_real("0")` constant.

- Reuse the zero-equality constraint as the negative-zero predicate rather than allocating a separate constant-true boolean AST.

- Leave ordinary `+0.0` constants untagged.

- Reuse the existing negative-zero metadata infrastructure without introducing a new tracking mechanism.

- Update the related documentation to clarify:
  - the current sources of negative-zero metadata;
  - that callers must not rely on `mk_smt_real` returning pointer-distinct ASTs for repeated constants;
  - the raw-AST-pointer lifetime limitation across `pop_ctx()`;
  - the one-way behaviour of assignment metadata propagation.

## Regression Tests

The existing literal division regression checks both zero signs using a symbolic positive numerator:

```c
double neg = x / (-0.0);
double pos = x / 0.0;

__ESBMC_assert(neg < 0.0, ...);
__ESBMC_assert(pos > 0.0, ...);
```

This verifies that a literal `-0.0` produces negative infinity while an ordinary `+0.0` does not accidentally acquire negative-zero metadata.

Additional coverage includes:

- `ra-neg-zero-literal-div-sign-fail`
  - failing companion for the literal negative-zero division case.

- `ra-neg-zero-literal-signbit`
  - covers the `__builtin_signbit` consumer for literal negative zero.

Two residual merge cases are also pinned as KNOWNBUG regressions:

- `ra-neg-zero-if-merge-lost-sign`
- `ra-neg-zero-ternary-order-lost-sign`

These document cases where negative-zero metadata is not yet propagated soundly through control-flow or conditional merges.

## Testing

The literal division regression and its failing companion pass with their expected verdicts.

The implementation was also tested with `z3_convt::mk_smt_real` temporarily memoised by string, reproducing the AST-sharing scenario identified during review. Under that setup, the `-0.0` and `+0.0` division results remained correctly distinguished.

The temporary memoisation change was reverted after testing.

The literal division regression was additionally checked with:

```text
--smt-symex-assert
--incremental-bmc
--multi-property
```

All three modes completed with `VERIFICATION SUCCESSFUL`.

Full `ir-ra` suite: 255/255 passed.

`regression/esbmc/github_2572`: passed.

`clang-format-11` and `git diff --check`: clean.

## Scope

This PR only adds direct literal negative-zero constants as another producer of the existing metadata.

It does not implement general IEEE-754 signed-zero semantics or sound propagation of negative-zero metadata through control-flow or conditional merges.

In particular, both statement-level `if` merges and conditional `?:` expressions still have cases where negative-zero sign metadata is lost or becomes branch-order dependent. These residual cases are documented by KNOWNBUG regressions and remain separate follow-up work.

The existing raw-AST-pointer metadata lifetime issue across `pop_ctx()` is also not addressed by this PR; it predates this change and affects other side metadata using the same representation.

### DIFF
--- a/regression/ir-ra/ra-neg-zero-literal-div-sign/main.c
+++ b/regression/ir-ra/ra-neg-zero-literal-div-sign/main.c
@@ -1,0 +1,26 @@
+extern double __VERIFIER_nondet_double(void);
+
+int main(void)
+{
+  /* x is genuinely nondet (not pinned to a specific value), so neither
+   * division below can be resolved by ESBMC's own constant folder --
+   * each is only decidable via the actual SMT encoding, which exercises
+   * the literal zero constants' tracked sign directly (no branching/
+   * merge is needed to reach them, so this stays isolated to the
+   * constant-conversion path). */
+  double x = __VERIFIER_nondet_double();
+  __ESBMC_assume(x > 0.0);
+
+  double neg = x / (-0.0);
+  double pos = x / 0.0;
+
+  /* IEEE 754: positive / -0.0 == -infinity, positive / +0.0 == +infinity.
+   * Checking both in the same program confirms that recording
+   * negative-zero metadata for the -0.0 literal does not also mark the
+   * ordinary +0.0 literal as negative. */
+  __ESBMC_assert(
+    neg < 0.0, "dividing a positive value by literal -0.0 must give -infinity");
+  __ESBMC_assert(
+    pos > 0.0, "dividing a positive value by literal +0.0 must give +infinity");
+  return 0;
+}
--- a/regression/ir-ra/ra-neg-zero-literal-div-sign/test.desc
+++ b/regression/ir-ra/ra-neg-zero-literal-div-sign/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--ir-ieee --z3
+^VERIFICATION SUCCESSFUL$
--- a/src/solvers/smt/fp/ir_ieee_conv.h
+++ b/src/solvers/smt/fp/ir_ieee_conv.h
@@ -96,20 +96,28 @@ public:
    *  mk_or(a, b) if both are set. */
   smt_astt combine_nan_preds(smt_astt a, smt_astt b) const;
 
-  /** Record that the SMT AST t is a real zero produced by flushing a
-   *  negative subnormal-range result (smt_solver_baset::mk_subnormal_flush).
-   *  neg_zero_pred is a boolean SMT term that is true iff t stands for
-   *  IEEE 754 -0.0. This tracks only the negative-zero case arising from
-   *  subnormal flushing, not general signed-zero semantics. */
+  /** Record a predicate that is true iff the SMT AST t represents
+   *  IEEE 754 -0.0. Such predicates originate when a negative
+   *  subnormal-range result is flushed to zero and when a literal
+   *  negative-zero float constant is converted, and may subsequently
+   *  be propagated through assignments or wrapper terms. This does not
+   *  implement general signed-zero semantics. */
   void store_neg_zero_pred(smt_astt t, smt_astt neg_zero_pred);
 
-  /** Return the stored negative-zero predicate for t, or nullptr if t is
-   *  not known to be a flushed negative zero. */
+  /** Return the stored negative-zero predicate for t, or nullptr if no
+   *  negative-zero metadata is recorded for t. */
   smt_astt get_neg_zero_pred(smt_astt t) const;
 
   /** Propagate a negative-zero predicate from rhs to lhs after an SSA
    *  assignment. Called from smt_solver_baset::convert_assign alongside
-   *  propagate_interval and propagate_nan_pred. */
+   *  propagate_interval and propagate_nan_pred.
+   *
+   *  One-way: if rhs carries no predicate, any existing entry already
+   *  recorded for lhs is left in place rather than cleared. This is
+   *  harmless under SSA, where each assignment gives lhs a fresh AST
+   *  with no prior entry of its own, but callers reusing an AST across
+   *  more than one assignment should not assume this clears stale
+   *  metadata. */
   void propagate_neg_zero_pred(smt_astt lhs, smt_astt rhs);
 
   /** Re-attach inner's negative-zero predicate (if any) to outer, guarded
@@ -175,10 +183,24 @@ private:
   std::unordered_map<const smt_ast *, smt_astt> ir_ieee_nan_map;
 
   /** Map from AST pointer to its negative-zero predicate (a boolean SMT
-   *  term that is true iff the value is a real zero produced by flushing
-   *  a negative subnormal-range result). Populated only by
-   *  smt_solver_baset::mk_subnormal_flush; an absent entry means that no
-   *  flushed-negative-zero metadata is recorded for the value. */
+   *  term that is true iff the value stands for IEEE 754 -0.0). Predicates
+   *  originate when a negative subnormal-range result is flushed to zero
+   *  (smt_solver_baset::mk_subnormal_flush) and when a literal
+   *  negative-zero float constant is converted
+   *  (smt_solver_baset::convert_terminal's constant_floatbv_id case), and
+   *  may subsequently be propagated through assignments or wrapper terms;
+   *  an absent entry means that no negative-zero metadata is recorded for
+   *  the value.
+   *
+   *  Lifetime: keyed on raw AST pointers and never pruned, but
+   *  smt_solver_baset::pop_ctx() deletes every AST created since the
+   *  matching push_ctx(). A pop can therefore leave both a key and its
+   *  stored predicate value dangling; a later lookup that hits such an
+   *  entry would hand a freed AST to a consumer. This exposure predates
+   *  this map (ir_ieee_nan_map above has the same shape and is subject
+   *  to the same limitation) and is not specific to negative-zero
+   *  tracking, but is noted here since this map's population grows with
+   *  every literal negative-zero constant converted. */
   std::unordered_map<const smt_ast *, smt_astt> ir_ieee_neg_zero_map;
 
   /** Set of symbol names that have already received integer range assertions,
--- a/src/solvers/smt/smt_fp_conv.cpp
+++ b/src/solvers/smt/smt_fp_conv.cpp
@@ -828,9 +828,10 @@ smt_astt smt_solver_baset::convert_signbit(const expr2tc &expr)
     // We can't extract bits, so check the sign mathematically
     is_neg = mk_lt(value, mk_smt_real("0"));
 
-    // A value flushed to zero from a negative subnormal-range result is
-    // IEEE 754 -0.0, but the real zero it collapsed to carries no sign of
-    // its own; consult the tracked negative-zero predicate to recover it.
+    // A value flushed to zero from a negative subnormal-range result, or
+    // a literal -0.0 constant, is IEEE 754 -0.0, but the real zero used
+    // to represent it carries no sign of its own; consult the tracked
+    // negative-zero predicate to recover it.
     smt_astt neg_zero_pred = ir_ieee_api->get_neg_zero_pred(value);
     if (neg_zero_pred)
       is_neg = mk_or(is_neg, neg_zero_pred);
--- a/src/solvers/smt/smt_solver.cpp
+++ b/src/solvers/smt/smt_solver.cpp
@@ -2016,7 +2016,29 @@ smt_astt smt_solver_baset::convert_terminal(const expr2tc &expr)
     if (int_encoding)
     {
       if (thereal.value.is_zero())
+      {
+        // A literal -0.0 constant is a second source of IEEE 754 negative
+        // zero, alongside the subnormal-flush case handled by
+        // mk_subnormal_flush. Reuse the same neg_zero_pred side-channel
+        // rather than adding new tracking machinery -- but tag a fresh
+        // symbol constrained equal to zero, not the shared mk_smt_real("0")
+        // AST directly: nothing guarantees mk_smt_real returns a distinct
+        // pointer per call (see its declaration), so tagging the literal
+        // itself could let an ordinary +0.0 silently inherit this
+        // predicate if any backend ever memoises real constants by value.
+        // Mirrors the NaN branch below, which mints mk_fresh(...) for the
+        // same reason.
+        if (ir_ieee && thereal.value.get_sign())
+        {
+          smt_astt neg_zero_ast =
+            mk_fresh(mk_real_sort(), "ir_ieee::neg_zero_const::", nullptr);
+          smt_astt is_zero = mk_eq(neg_zero_ast, mk_smt_real("0"));
+          assert_ast(is_zero);
+          ir_ieee_api->store_neg_zero_pred(neg_zero_ast, is_zero);
+          return neg_zero_ast;
+        }
         return mk_smt_real("0");
+      }
       if (thereal.value.is_NaN())
       {
         if (ir_ieee)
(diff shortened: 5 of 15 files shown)


src/solvers/smt/smt_solver.cpp
```diff
@@ -2016,7 +2016,18 @@ smt_astt smt_solver_baset::convert_terminal(const expr2tc &expr)
...
+      {
+        smt_astt zero_ast = mk_smt_real("0");
+        // A literal -0.0 constant is a second source of IEEE 754 negative
+        // zero, alongside the subnormal-flush case handled by
+        // mk_subnormal_flush. Reuse the same neg_zero_pred side-channel
+        // rather than adding new tracking machinery: the real value
+        // returned is the ordinary zero AST, with a constant true
+        // predicate recording that it stands for -0.0.
+        if (ir_ieee && thereal.value.get_sign())
+          ir_ieee_api->store_neg_zero_pred(zero_ast, mk_smt_bool(true));
```
lucasccordeiro: This tags an AST that is a *shared-by-convention* constant. Nothing guarantees `mk_smt_real("0")` returns a distinct `smt_ast *` per call — it happens to, because every backend's `mk_smt_real` ends in `new_ast(...)`. The metadata map is keyed on that raw pointer, so the moment any backend memoises real constants (a natural optimisation; Z3's `real_val("0")` already returns the same Z3 term, only the ESBMC wrapper differs), the `+0.0` literal inherits the `-0.0` predicate.

Demonstrated on this branch by adding a `std::map<std::string, smt_astt>` memo to `z3_convt::mk_smt_real`:

```
Violated property:
  dividing a positive value by literal +0.0 must give +infinity
VERIFICATION FAILED
```

i.e. every real zero in the formula becomes negative zero, silently.

The NaN branch immediately below already solves this — it does not tag the shared zero, it mints `mk_fresh(mk_real_sort(), "ir_ieee::nan_const::", nullptr)` and tags that. Please do the equivalent here (tag a value whose identity you own), or key the metadata on something identity-stable rather than an AST pointer. At the very least, state the invariant at this call site and at `mk_smt_real`, because it is currently invisible to anyone optimising the backends.

Separately, `mk_smt_bool(true)` allocates a new boolean AST for every negative-zero constant in the program; a single shared constant-true would do.

src/solvers/smt/fp/ir_ieee_conv.h
```diff
@@ -175,10 +176,14 @@ class ir_ieee_convt
...
-   *  term that is true iff the value is a real zero produced by flushing
-   *  a negative subnormal-range result). Populated only by
-   *  smt_solver_baset::mk_subnormal_flush; an absent entry means that no
-   *  flushed-negative-zero metadata is recorded for the value. */
+   *  term that is true iff the value stands for IEEE 754 -0.0). Predicates
+   *  originate when a negative subnormal-range result is flushed to zero
+   *  (smt_solver_baset::mk_subnormal_flush) and when a literal
+   *  negative-zero float constant is converted
+   *  (smt_solver_baset::convert_terminal's constant_floatbv_id case), and
+   *  may subsequently be propagated through assignments or wrapper terms;
```
lucasccordeiro, line 191: Two things this newly-written doc should state, since it is now the reference for the metadata's contract:

1. **Lifetime.** The map is keyed on raw `const smt_ast *` and is never pruned, but `pop_ctx()` (smt_solver.cpp:278-283) `delete`s every AST created since the matching push, and prunes `smt_cache` accordingly. After a pop, both the keys *and* the stored predicate values dangle; a later `get_neg_zero_pred()` hit would hand `convert_signbit`/`encode_ieee_div` a freed AST to build terms from. `--ir-ieee` does reach the push/pop modes — I ran the new test under `--smt-symex-assert`, `--incremental-bmc` and `--multi-property` and all three work. I could not provoke a wrong verdict, and this predates the PR (the flush and NaN producers have the same exposure), but this change adds a producer that fires on every literal `-0.0`, which multiplies the map's population. Either prune both maps in `pop_ctx()` or record the constraint here.

2. **"may subsequently be propagated through assignments"** — `propagate_neg_zero_pred` (ir_ieee_conv.cpp:85-90) only ever *adds*: when `rhs` carries no predicate it leaves any existing entry for `lhs` in place. That is harmless today because SSA gives each assignment a fresh `lhs` AST (I checked `double z = -0.0; z = 0.0; x / z` — correctly `+inf`), but the doc now advertises propagation as part of the contract, so the one-way behaviour is worth naming.

src/solvers/smt/smt_fp_conv.cpp
```diff
@@ -828,9 +828,10 @@ smt_astt smt_solver_baset::convert_signbit(const expr2tc &expr)
     // We can't extract bits, so check the sign mathematically
     is_neg = mk_lt(value, mk_smt_real("0"));
 
-    // A value flushed to zero from a negative subnormal-range result is
-    // IEEE 754 -0.0, but the real zero it collapsed to carries no sign of
-    // its own; consult the tracked negative-zero predicate to recover it.
+    // A value flushed to zero from a negative subnormal-range result, or
+    // a literal -0.0 constant, is IEEE 754 -0.0, but the real zero used
```
lucasccordeiro, line 832: This comment now claims literal `-0.0` as a source for the `signbit` consumer, and that is true — but no test covers it. It is a genuine second fix and it discriminates:

```c
double x = __VERIFIER_nondet_double();
__ESBMC_assume(x > 0.0);
double z = -0.0;
double q = x / z;                       /* keeps z off the constant folder */
__ESBMC_assert(__builtin_signbit(z) != 0, "signbit(-0.0)");
```

master: `VERIFICATION FAILED`; this branch: `VERIFICATION SUCCESSFUL`. Please add it as `regression/ir-ra/ra-neg-zero-literal-signbit`. Use `__builtin_signbit` rather than `<math.h>`'s `signbit` — on macOS the latter expands to `__inline_signbitd`, which does its own bit-twiddling and never reaches `convert_signbit`, so a test written with `signbit` would silently not exercise this path on that platform.

regression/ir-ra/ra-neg-zero-literal-div-sign/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--ir-ieee --z3
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: No failing companion. 32 of the 255 tests in this suite come in `X` / `X-fail` pairs (`ra-const-nan-pred{,-fail}`, `ra-div-zero-zero-nan{,-fail}`, …), and AGENTS.md asks for at least one passing and one failing test per PR. Add `ra-neg-zero-literal-div-sign-fail` with the assertion inverted (`neg > 0.0`) and `^VERIFICATION FAILED$`, so a future change that drops the metadata entirely is caught by a test flipping in *both* directions rather than only by this one going red.

regression/ir-ra/ra-neg-zero-literal-div-sign/main.c
```diff
@@ -0,0 +1,26 @@
...
+  /* x is genuinely nondet (not pinned to a specific value), so neither
+   * division below can be resolved by ESBMC's own constant folder --
+   * each is only decidable via the actual SMT encoding, which exercises
+   * the literal zero constants' tracked sign directly (no branching/
+   * merge is needed to reach them, so this stays isolated to the
+   * constant-conversion path). */
+  double x = __VERIFIER_nondet_double();
+  __ESBMC_assume(x > 0.0);
+
+  double neg = x / (-0.0);
```
lucasccordeiro, line 14: The PR body says branch/conditional merges are out of scope, but the boundary is not where that puts it — `?:` is fixed by this change while `if` is not, and the `if` form returns a wrong SUCCESSFUL:

```c
double z = c ? -0.0 : 0.0;   /* master: SUCCESSFUL (wrong); this branch: FAILED, r = -INFINITY */
```
```c
double z = 0.0;
if (c) z = -0.0;             /* master and this branch: SUCCESSFUL — but r is -inf when c */
double r = x / z;
__ESBMC_assert(r > 0.0, "...");
```

Both programs are the same program. The second is a missed violation, not just missing precision. Please add the `if` form as a `KNOWNBUG` test next to this one so the residual gap is pinned and shows up when merge propagation lands, and correct the scope paragraph — as written it reads as though neither form works.

lucasccordeiro (changes requested):
Reviewed by building the branch and comparing against a master build of the same tree (macOS/arm64, Z3 4.16.0). The change does what it says: `x / -0.0` goes FAILED → SUCCESSFUL, and it also fixes `__builtin_signbit(-0.0)` and `double z = -0.0; x / z`, which master gets wrong. The new test discriminates (FAILED on master).

Main concern: correctness rests on an invariant that is neither stated nor enforced — that `mk_smt_real("0")` hands back an unshared `smt_ast *`. I made `z3_convt::mk_smt_real` memoise by string and the new test immediately reports *`dividing a positive value by literal +0.0 must give +infinity`* as violated: every real zero in the formula becomes `-0.0`. The NaN branch eight lines below already avoids this by tagging a `mk_fresh` symbol instead of a constant.

Also missing: a `-fail` companion (32 of the 255 ir-ra tests have one; AGENTS.md asks for one passing and one failing test per PR), and any test for the `signbit` consumer whose comment this PR updates.

Unrelated to this PR: `regression/ir-ra/ra-fmod-inf-nan` fails on my machine on master too, so it is not a regression from this change.

Details inline.

src/solvers/smt/smt_solver.cpp
ShemaAlosaimi: Fixed.

Literal `-0.0` no longer attaches metadata directly to the AST returned by `mk_smt_real("0")`.

The conversion now creates a fresh real symbol with `mk_fresh(...)`, constrains it equal to real zero, and attaches the negative-zero metadata to that fresh AST. The same equality term is also reused as the stored predicate, so there is no separate `mk_smt_bool(true)` allocation.

I also updated the `mk_smt_real` documentation to make it explicit that callers must not assume pointer uniqueness across repeated constants.

I re-ran the exact memoisation stress test from the review by temporarily memoising `z3_convt::mk_smt_real` by string. Both the `-0.0` and `+0.0` division checks remained correct, and the full `ir-ra` suite still passed under that setup. The temporary memoisation change was then reverted.

src/solvers/smt/fp/ir_ieee_conv.h
ShemaAlosaimi: Agreed. I documented both constraints explicitly.

`ir_ieee_neg_zero_map` now notes that it is keyed by raw AST pointers, is not currently pruned on `pop_ctx()`, and therefore has the same potential dangling-pointer exposure as the existing NaN metadata map.

I also documented that `propagate_neg_zero_pred` is one-way: it copies metadata when the RHS has an entry, but does not clear an existing LHS entry when the RHS has none. That is safe under the current SSA assumption because each assignment gets a fresh LHS AST, but the comment no longer implies stronger behaviour.

I left `pop_ctx()` pruning out of this PR because the issue predates this change and should be fixed consistently across the affected metadata maps rather than only for negative-zero tracking.

src/solvers/smt/smt_fp_conv.cpp
ShemaAlosaimi: Added `regression/ir-ra/ra-neg-zero-literal-signbit` using `__builtin_signbit`, as suggested.

One platform-specific detail: on my Linux/Z3 4.13.3 build I could not reproduce the fail-before/pass-after behaviour you observed. With the default pipeline, the assertion is discharged before SMT even with the `x / z` use, whereas your macOS/arm64 setup showed that the same construction discriminates.

I kept the regression because it directly expresses the required IEEE-754 behaviour and provides coverage for the `__builtin_signbit` consumer. I also made the test comment platform-neutral and do not claim that it necessarily reaches SMT on every build.

regression/ir-ra/ra-neg-zero-literal-div-sign/test.desc
ShemaAlosaimi: Added `ra-neg-zero-literal-div-sign-fail`.

It uses the same setup as the passing regression, but inverts the negative-zero division assertion and expects `VERIFICATION FAILED`.

I also checked that it discriminates against the unfixed implementation: with the fix present the passing/failing pair behaves as expected, while reverting the literal-negative-zero fix causes the companion to flip in the opposite direction.

regression/ir-ra/ra-neg-zero-literal-div-sign/main.c
ShemaAlosaimi: Thanks — I investigated both forms in more detail, and there is an additional wrinkle in the ternary case.

You are right that the original scope paragraph drew the boundary incorrectly. However, after tracing the generated SMT, I found that the `?:` case is not actually propagating negative-zero metadata correctly either.

For:

```c
double z = c ? -0.0 : 0.0;
```

the result happens to be correct on this branch, but the generated SMT does not encode the sign as depending on `c`. The negative-zero predicate effectively becomes constant, so this is not genuine merge propagation.

I confirmed that by reversing the branches:

```c
double z = c ? 0.0 : -0.0;
```

With that ordering, the verifier returns the wrong result.

So the actual boundary is not “`?:` works while `if` does not.” Neither form currently handles negative-zero metadata soundly across the merge. One ternary ordering only happens to produce the expected result because simplification treats `+0.0` and `-0.0` as numerically interchangeable and retains one branch's AST identity.

The statement-level form still exhibits the issue you identified:

```c
double z = 0.0;
if (c)
  z = -0.0;

double r = x / z;
__ESBMC_assert(r > 0.0, "...");
```

When the negative-zero branch is taken, `r` should be `-INFINITY`, but the negative-zero metadata is lost across the SSA/control-flow merge and the verifier can incorrectly report `VERIFICATION SUCCESSFUL`.

I added two KNOWNBUG regressions to pin both residual cases:

- `ra-neg-zero-if-merge-lost-sign`
- `ra-neg-zero-ternary-order-lost-sign`

The ternary KNOWNBUG uses the branch ordering that currently produces the wrong verdict, so the branch-order dependence is explicitly tracked rather than relying on the ordering that happens to work.

I also corrected the PR scope accordingly. The PR now only claims to add literal negative-zero constants as a producer of the existing metadata. It does **not** claim sound propagation of negative-zero metadata through general control-flow or conditional merges; both the `if` case and the ternary branch-order case remain follow-up work.

lucasccordeiro (approved):

Issue #6443, opened 2026-07-27
k-induction: assert-to-assume conversion never takes effect, masking a latent soundness bug


pierredantas:
# k-induction: assert-to-assume conversion never takes effect, masking a latent soundness bug

**Version:** 8.4.0 (also reproduced on `e73b1f1f`), aarch64 macOS, Bitwuzla 0.9.1 / Z3 4.16.

## Summary

`goto_symext::claim()` documents converting in-loop assertions into assumptions on
all but the last unwinding of the inductive step — the standard way to give
k-induction its induction hypothesis. The conversion is written but never takes
effect, because the claim is recorded *before* the conversion branch and the branch
only adds an assumption on top.

Consequence: the inductive step has no induction hypothesis and degenerates into BMC
from an arbitrary state. Properties that are k-inductive but not 1-inductive are
reported `VERIFICATION UNKNOWN` at every k. Raising k adds obligations to discharge
rather than hypotheses to use, so runtime is flat in k and the verdict never changes.

Separately, the conversion guard reads the iteration counter of the **first loop
entered** rather than the assertion's own enclosing loop. Defect 1 currently masks
this. Fixing Defect 1 alone makes it a **soundness bug** — verified below.

## Defect 1 — conversion is written, not performed

`src/goto-symex/symex_main.cpp`, `goto_symext::claim()`:

```cpp
// add assertion to the target equation
assertion(new_expr, msg);                       // <- claim recorded unconditionally

// Convert asserts in assumes, if it's not the last loop iteration
// This is a common technique in k-induction to strengthen the induction hypothesis.
if (inductive_step && first_loop && !cur_state->source.pc->inductive_assertion)
{
  BigInt unwind = cur_state->loop_iterations[first_loop];
  if (unwind < max_unwind - 1)
  {
    assume(claim_expr);                         // <- and then also assumed
    return;
  }
}
```

The comment says *convert*; the code *adds both*. The claim stays a live proof
obligation on every unwinding.

### Reproducer

```c
int main(void) {
  _Bool x = 0;
  while (1) { __ESBMC_assert(!x, "P"); x = 0; }
}
```

`P` is 2-inductive: from any state, one loop body execution establishes `!x`.

```
$ esbmc disc.c --k-induction --max-k-step 5
Checking inductive step, k = 2 ... The inductive step is unable to prove the property
Checking inductive step, k = 3 ... The inductive step is unable to prove the property
[...]

PR #6451, opened 2026-07-27
k-induction: perform the assert-to-assume conversion (#6443)

pierredantas:
Fixes the first half of #6443.

## The defect

`goto_symext::claim()` documents converting in-loop assertions into the induction hypothesis on all but the last unwinding of the inductive step. It recorded the claim *before* the conversion branch and then only added an assumption on top:

```cpp
assertion(new_expr, msg);                       // claim recorded unconditionally
if (inductive_step && first_loop && ...)
{
  if (unwind < max_unwind - 1)
  {
    assume(claim_expr);                         // ... and then also assumed
    return;
  }
}
```

The claim stayed a live proof obligation on every unwinding, so the inductive step had no induction hypothesis and degenerated into BMC from an arbitrary state. Properties that are k-inductive but not 1-inductive were reported `VERIFICATION UNKNOWN` at every k, and raising k added obligations to discharge rather than hypotheses to use.

## The change

Convert rather than duplicate, with two guards so it only fires where it is sound.

**Not when the claim has simplified to `false`.** Assuming it is `assume(false)`, which kills the path; later claims then simplify away and the step discharges vacuously without ever reaching the solver, so an `assert(0)` inside a loop would be reported SUCCESSFUL.

**Not once more than one loop has been in play on the path.** The guard selects assertions using `first_loop`'s iteration counter, which only describes the assertion's own loop while that naming is stable. `symex_goto` resets `first_loop` when the loop naming it exits, so in a multi-loop program an assertion in a later loop can be assumed on the strength of an unrelated loop's depth — assuming away real violations. The new `multiple_loops_seen` flag tracks this.

Correct loop attribution — which would lift the second restriction and let the conversion apply to multi-loop programs too — is left open in #6443. That needs the assertion's own enclosing loop, which `compute_loop_numbers` does not currently record for instructions inside a loop body.

## Effect

Single-loop programs, including PLC scan-cycle encodings where the safety property sits inside a `while(1)` loop and must serve as its own inductive hypothesis, are now provable. On a 40-line storm-surge-barrier controller with a one-shot timer, `--k-induction` goes from UNKNOWN at every k to SUCCESSFUL at exactly `K = T_DONE - T_ARM + 2` (7, 17, 37, 77, 157 at T_DONE = 10, 20, 40, 80, 160), independently matching an explicit k-induction harness over the same program.

## Testing

| suite | before | after |
|---|---|---|
| `k-induction` | 122/122 | 122/122 |
| `k-induction-parallel` | 74/74 | 74/74 |
| full regression (6462) | 46 failures | **same 46 failures** |

The 46 are pre-existing environment failures on this machine (macOS, LLVM 22 — floats, C++ templates, char literals); the failing set is byte-identical before and after, so this introduces no regressions and none of them are in k-induction.

Two new tests:

- `github_6443_single_loop` — UNKNOWN before this change, SUCCESSFUL after. Guards the fix.
- `github_6443_multi_loop` — FAILED before and after. Pins the soundness guard; if the multi-loop restriction is ever relaxed without correct loop attribution, this catches it.

Both verified to discriminate: the single-loop test reports UNKNOWN on a stock build of the same commit.

https://claude.ai/code/session_01WWW5oL2UQZF5s7U9uyFRDR


src/goto-symex/symex_assign.cpp
```diff
@@ -46,7 +46,8 @@ goto_symext::goto_symext(
     base_case(options.get_bool_option("base-case")),
     forward_condition(options.get_bool_option("forward-condition")),
     inductive_step(options.get_bool_option("inductive-step")),
-    validate_witness(options.get_bool_option("validate-violation-witness"))
+    validate_witness(options.get_bool_option("validate-violation-witness")),
+    multiple_loops_seen(false)
```
lucasccordeiro, line 50: `multiple_loops_seen` is declared immediately after `first_loop` (goto_symex.h:1276) but initialised last, so the mem-init list is out of declaration order. `-Wall -Wextra` are project defaults (CMakeLists.txt:55-56), so this warns and fails any `-DENABLE_WERROR=On` build:

```
src/goto-symex/symex_assign.cpp:49:5: warning: field 'validate_witness' will be initialized after field 'multiple_loops_seen' [-Wreorder-ctor]
```

(reproduced with `clang++ -Wall -Wextra -fsyntax-only` on this branch).

Move it next to `first_loop(0),`. Same for line 192 in `operator=` — put it beside `first_loop = sym.first_loop;` so the two fields that must stay in sync are adjacent.

src/goto-symex/goto_symex.h
```diff
@@ -1263,6 +1263,17 @@ class goto_symext
...
+   *  path -- either a second loop was entered, or the loop that `first_loop`
+   *  named was exited and `first_loop` reassigned.
+   *
+   *  The inductive step's assert-to-assume conversion selects assertions by
+   *  `first_loop`'s iteration counter, which only describes the assertion's
+   *  own loop while that naming stays stable. Once it does not, an assertion
+   *  in one loop can be assumed on the strength of another loop's depth,
+   *  which assumes away real violations (esbmc/esbmc#6443). The conversion is
+   *  therefore restricted to paths where this flag is still false. */
+  bool multiple_loops_seen;
```
lucasccordeiro, line 1276: The doc says the flag is true "either a second loop was entered, or the loop that `first_loop` named was exited". Only the second disjunct is implemented — nothing sets the flag at the loop-entry site (`symex_main.cpp:269`, `if (inductive_step && instruction.loop_number && !first_loop)`). Entering a second or nested loop while `first_loop` is still held by an earlier loop leaves the flag false, and that is precisely the mis-attribution case described here.

Also, "on this path" overstates it: this is a member of `goto_symext`, monotone for the lifetime of the symex object (copied through `operator=` per execution state), not a per-path property. And the name is misleading — it becomes true for a program with exactly one loop, as soon as that loop exits. `first_loop_naming_stale` would describe what is actually tracked.

src/goto-symex/symex_goto.cpp
```diff
@@ -124,7 +124,12 @@ void goto_symext::symex_goto(const expr2tc &old_guard)
 
       // Reset loop counter
       if (instruction.loop_number == first_loop)
+      {
+        // The loop naming `first_loop` has exited; whatever loop is named
+        // next is a different one. See goto_symext::multiple_loops_seen.
         first_loop = 0;
+        multiple_loops_seen = true;
```
lucasccordeiro, line 131: This is the wrong edge to set the flag on. Reaching here means `first_loop = 0`, so the *next* loop entered re-assigns `first_loop` to itself and the attribution is correct — the safe case. The unsafe case is a loop left without hitting this code (`break`/`goto`/`return` out of a `while (1)`, whose backwards goto is unconditional and so never takes the `new_guard_false` path): `first_loop` stays stale, the flag stays false, and the conversion still fires for an assertion in a later loop using the stale counter.

So the guard suppresses the conversion where it was sound and permits it where it may not be. If a guard is wanted here, set it where `first_loop` is *retained* across a new loop entry, i.e. at `symex_main.cpp:269`.

src/goto-symex/symex_main.cpp
```diff
@@ -128,16 +128,37 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
...
+  // describes the assertion's own loop only when there is no other loop to
+  // confuse it with. `loop_iterations` gains an entry on every loop entry,
+  // so a size of one means any enclosing loop must be `first_loop` itself.
+  // In a multi-loop program an assertion in a later loop would otherwise be
+  // assumed on the strength of an unrelated loop's depth, assuming away real
+  // violations (esbmc/esbmc#6443).
   if (
-    inductive_step && first_loop && !cur_state->source.pc->inductive_assertion)
+    inductive_step && first_loop && !is_false(new_expr) &&
+    !multiple_loops_seen && cur_state->loop_iterations.size() <= 1 &&
```
lucasccordeiro, line 160: Two overlapping predicates, and neither is documented as the load-bearing one. Please pick one.

The justification given above for the second one is not correct: "`loop_iterations` gains an entry on every loop entry, so a size of one means any enclosing loop must be `first_loop` itself". Entries are created at the *backwards goto* (symex_goto.cpp:123, :174), not at loop entry. On the first body execution of a second loop the map still holds only the previous, already-exited loop's entry, so `size() <= 1` holds and the conversion fires with the wrong loop's counter. Conversely on the first body execution of the *first* loop the size is 0, so the `<= 1` bound is doing nothing there either.

Combined effect, measured on this branch: the conversion is disabled by any second loop, even an irrelevant one.

```c
int g;
int main(void) {
  for (int a = 0; a < 2; a++) g = g + 1;   /* delete this loop and it proves */
  _Bool x = 0;
  while (1) { __ESBMC_assert(!x, "P"); x = 0; }
}
```
`--k-induction --max-k-step 6` → `VERIFICATION UNKNOWN`; without the first loop → SUCCESSFUL at k = 2. Nesting the `while (1)` inside a `for` is also UNKNOWN. Note the empty-bodied `for (int a = 0; a < 2; a++);` does *not* disable it, because that loop is optimised out of the GOTO program — so whether the feature applies depends on goto-level optimisation, which is a fragile contract for users.

That rules out the stated PLC scan-cycle motivation for any program with an initialisation loop, or a loop reached through a called function or an operational model (`memset`, `strlen`, …). Worth deciding whether this lands as-is with the limitation documented, or waits for the loop attribution in #6443.

src/goto-symex/symex_main.cpp
```diff
@@ -128,16 +128,37 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
...
+  // selects assertions using `first_loop`'s iteration counter, which
+  // describes the assertion's own loop only when there is no other loop to
+  // confuse it with. `loop_iterations` gains an entry on every loop entry,
+  // so a size of one means any enclosing loop must be `first_loop` itself.
+  // In a multi-loop program an assertion in a later loop would otherwise be
+  // assumed on the strength of an unrelated loop's depth, assuming away real
+  // violations (esbmc/esbmc#6443).
   if (
-    inductive_step && first_loop && !cur_state->source.pc->inductive_assertion)
+    inductive_step && first_loop && !is_false(new_expr) &&
```
lucasccordeiro, line 159: `is_false(new_expr)` is syntactic, but the vacuity it is protecting against is semantic: assuming *any* claim that is infeasible in the current context kills the path just as effectively (e.g. `assert(p)` where an earlier `assume` already forces `p` false, or a conjunction the simplifier does not fold). If the failure mode described above is real, this does not close it; if the base case is expected to catch `assert(0)` in a loop, this guard is unnecessary.

The commit message says an `assert(0)` inside a loop would otherwise be reported SUCCESSFUL. Please add that program as a regression test — nothing in this PR exercises the path (`github_6443_multi_loop` is caught by the base case at k = 11 regardless of any of the guards, see the comment on that file). If it *is* real, `is_assume_false()` (already in this file) is the semantic predicate; a syntactic check will leave the hole open.

src/goto-symex/symex_main.cpp
```diff
@@ -147,6 +168,11 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
       return;
     }
   }
+
+  symex_witness_assert(new_expr, msg);
```
lucasccordeiro, line 172: Two side effects of moving these below the conversion, both untested:

- `symex_witness_assert()` is no longer called for converted claims. `--witness-output*` combined with `--k-induction` now emits a different set of witness asserts from the inductive step. If that is intended, say so; either way there is no test.
- `++total_claims` still happens at the top of `claim()` while `remaining_claims` no longer does, so the "Generated N VCC(s), M remaining" line and `--multi-property` accounting change for the inductive step. On `github_6443_single_loop` the step reports `Generated 2 VCC(s), 0 remaining after simplification`. Converted claims are now counted as generated but never as remaining — worth confirming nothing downstream (multi-property bookkeeping, coverage) depends on the old accounting.

src/goto-symex/symex_main.cpp
```diff
@@ -128,16 +128,37 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
...
 
-  symex_witness_assert(new_expr, msg);
-
-  // add assertion to the target equation
-  assertion(new_expr, msg);
-
-  // Convert asserts in assumes, if it's not the last loop iteration
-  // This is a common technique in k-induction to strengthen the induction hypothesis.
-  // also, don't convert assertions added by the bidirectional search
+  // Convert asserts into assumes, if it's not the last loop iteration.
```
lucasccordeiro, line 131: ~25 lines of comment for a five-term condition, plus 11 more on the member declaration. The repo convention (AGENTS.md / CLAUDE.md, *Code Comments*) is few, high-value comments. Most of this is the commit message and the #6443 rationale. Suggest keeping the two non-obvious *why*s — record after converting, and why each guard exists — in about four lines and moving the rest to the commit message.

regression/k-induction/github_6443_multi_loop/main.c
```diff
@@ -0,0 +1,17 @@
...
+ * describe it. Converting it to the induction hypothesis here would assume
+ * the violation away; the bug must still be found. */
+#include <assert.h>
+int main(void)
+{
+  for (int i = 0; i < 10; i++)
+    assert(1);
+
+  for (int j = 0; j < 2; j++)
+    assert(0);
```
lucasccordeiro, line 14: This test does not pin the multi-loop guard. `assert(0)` folds to `false` in `do_simplify`, so `!is_false(new_expr)` blocks the conversion before `multiple_loops_seen` is ever consulted, and `assert(1)` in the first loop returns early on the `is_true` path — neither loop reaches the guard under test. On top of that, the base case finds `assert(0)` unconditionally.

Verified: rebuilt this branch with `!multiple_loops_seen && cur_state->loop_iterations.size() <= 1 &&` deleted from the condition, and the test still reports `VERIFICATION FAILED`, `Bug found (k = 11)` — identical to the guarded build.

To actually pin it the test needs (a) an assertion that is not constant-false, in the second loop, and (b) a violation the base case cannot reach within `--max-k-step`, so that a spurious inductive-step success would be observable as SUCCESSFUL. Please also confirm the replacement fails with the guard removed, the way `github_6443_single_loop` was confirmed against a stock build.

regression/k-induction/github_6443_single_loop/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--k-induction --max-k-step 5
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: This test discriminates (UNKNOWN before, SUCCESSFUL after), but it succeeds without the solver ever seeing a proof obligation:

```
Checking inductive step, k = 2
Generated 2 VCC(s), 0 remaining after simplification (3 assignments)
```

Both in-loop claims are converted, and the surviving one folds to true because `x = 0` is a constant assignment. So it demonstrates that in-loop claims are no longer emitted, not that the induction hypothesis is usable. Please add a case where the step's remaining VCC is discharged *using* the assumption (a non-constant loop variable, so the step reaches the solver and returns unsat).

Also missing: a positive-control test that a genuine in-loop violation in a single-loop program is still reported. I checked `while (1) { assert(i != 5); i++; }` on this branch and it is still FAILED at k = 6, but since the change removes proof obligations from the inductive step that safety net belongs in the suite.

lucasccordeiro (changes requested):
Reviewed by building the branch locally (macOS/arm64, RelWithDebInfo, Bitwuzla) and running experiments against it. `ctest -L k-induction` is 198/198 here, and the core change (recording the claim after, not before, the conversion) is the right shape.

Blocking items:

1. `-Wreorder-ctor` on the new member initialiser — breaks `-DENABLE_WERROR=On` builds.
2. `github_6443_multi_loop` does not exercise the guard it is described as pinning. I rebuilt the branch with `!multiple_loops_seen && cur_state->loop_iterations.size() <= 1 &&` deleted and the test still reports `VERIFICATION FAILED` (bug found at k = 11).
3. The `multiple_loops_seen` flag is set on the wrong edge: it fires exactly where `first_loop` is *correctly* re-assigned, and stays false in the stale-`first_loop` case the flag is meant to cover.
4. As a result the conversion is disabled by any second loop, including an irrelevant one. Prepending `for (int a = 0; a < 2; a++) g = g + 1;` to `github_6443_single_loop/main.c` turns SUCCESSFUL (k = 2) into `VERIFICATION UNKNOWN`; wrapping the `while (1)` in an outer loop does the same. The PLC scan-cycle motivation will not survive a single initialisation loop or a loop inside a called function/OM. This limitation should be stated in the PR body and commit message.

Details inline.

mikhailramalho:
We should run this branch on sv-comp as it looks like a major redesign of k-induction.

lucasccordeiro:
> We should run this branch on sv-comp as it looks like a major redesign of k-induction.

I fully agree.

As soon as @pierredantas addresses the comments and fixes the CI, I can start a run over the SV-COMP benchmarks.

src/goto-symex/goto_symex.h
lucasccordeiro, line 1276: Documented as per-path, but this lives on `goto_symext` and is never reset, so one path exiting a loop disables the conversion for every later path. Move it to `cur_state` or fix the doc.

src/goto-symex/goto_symex.h
```diff
@@ -1263,6 +1263,17 @@ class goto_symext
   irep_idt guard_identifier_s;
   /** Loop numbers. */
   unsigned first_loop;
+  /** True once symbolic execution has had more than one loop in play on this
+   *  path -- either a second loop was entered, or the loop that `first_loop`
```
lucasccordeiro, line 1267: "a second loop was entered" is never implemented — the flag is only set on loop exit. Drop that clause or set it on entry too.

regression/k-induction/github_6443_multi_loop/main.c
```diff
@@ -0,0 +1,17 @@
...
+ * loop, so the inductive step's `first_loop` iteration counter does not
+ * describe it. Converting it to the induction hypothesis here would assume
+ * the violation away; the bug must still be found. */
+#include <assert.h>
+int main(void)
+{
+  for (int i = 0; i < 10; i++)
+    assert(1);
+
+  for (int j = 0; j < 2; j++)
```
lucasccordeiro, line 13: The new `!is_false(new_expr)` guard is untested: this `assert(0)` is excluded by the multi-loop guard instead. Add a single-loop `assert(0)` case.

src/goto-symex/symex_main.cpp
lucasccordeiro, line 131: A 25-line rationale block exceeds the surrounding comment density (CLAUDE.md). Keep one line per guard and move the argument to the commit message.

lucasccordeiro (changes requested):

src/goto-symex/symex_main.cpp
```diff
@@ -121,16 +121,18 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
...
-  // also, don't convert assertions added by the bidirectional search
+  // Convert asserts into assumes on all but the last unwinding of the inductive
+  // step, which is what supplies the induction hypothesis. This must run before
+  // the claim is recorded, or the claim stays a live proof obligation and the
+  // step degenerates into BMC from an arbitrary state (esbmc/esbmc#6443).
   if (
-    inductive_step && first_loop && !cur_state->source.pc->inductive_assertion)
+    inductive_step && first_loop &&
+    // a claim simplified to false would become assume(false), killing the path
+    !is_false(new_expr) &&
```
lucasccordeiro: `is_false` reads the post-`do_simplify` expression, but `--no-simplify` makes that a no-op: `assert(0)` stays `!((_Bool)(!(_Bool)0))` and is assumed away. Make the guard simplification-independent.

src/goto-symex/symex_main.cpp
```diff
@@ -121,16 +121,18 @@ void goto_symext::claim(const expr2tc &claim_expr, const std::string &msg)
...
-  // Convert asserts in assumes, if it's not the last loop iteration
-  // This is a common technique in k-induction to strengthen the induction hypothesis.
-  // also, don't convert assertions added by the bidirectional search
+  // Convert asserts into assumes on all but the last unwinding of the inductive
+  // step, which is what supplies the induction hypothesis. This must run before
+  // the claim is recorded, or the claim stays a live proof obligation and the
+  // step degenerates into BMC from an arbitrary state (esbmc/esbmc#6443).
   if (
-    inductive_step && first_loop && !cur_state->source.pc->inductive_assertion)
+    inductive_step && first_loop &&
```
lucasccordeiro: `claim()` now trips the complexity gate (core CCN 15 → 18). Extract the five-term conversion condition into a named predicate.

unit/goto-symex/overapproximation.test.cpp
```diff
@@ -351,7 +351,7 @@ TEST_CASE(
   // carry their checks. Deleting the guard costs exactly one of each kind --
```
lucasccordeiro: Comment still says all four unwindings must carry their checks, contradicting the new `== 1`. Rewrite it for the assert-to-assume conversion.

unit/goto-symex/overapproximation.test.cpp
```diff
@@ -351,7 +351,7 @@ TEST_CASE(
...
   // a dereference nobody verifies, which is the missed-bug direction, and the
   // same asymmetry the function-pointer filter above is checked for.
-  REQUIRE(asserts_commented(eq, "dereference failure: invalid pointer") == 4);
+  REQUIRE(asserts_commented(eq, "dereference failure: invalid pointer") == 1);
```
lucasccordeiro, line 357: The `!filtered.empty()` mutant drops the post-havoc dereference, which is now converted to an assumption — re-confirm this test still fails with the guard deleted.

regression/k-induction/kinductor_01/test.desc
```diff
@@ -1,4 +1,4 @@
 THOROUGH
 main.c
 --k-induction -Wno-error=implicit-function-declaration
-^VERIFICATION UNKNOWN$
+^VERIFICATION SUCCESSFUL$
```
lucasccordeiro, line 4: THOROUGH is skipped on APPLE/WIN32 (regression/CMakeLists.txt:449-452), so the PR's macOS run never executed this flip. Please confirm the five THOROUGH fixture changes on Linux.

src/goto-symex/symex_main.cpp
pierredantas: Fixed: the guard now folds the expression through `branch_decision_guard` before testing `is_false`, the same idiom `symex_goto`'s `new_guard_false` already uses for #6778, so it's simplification-independent regardless of `--no-simplify`. Extracted into `should_convert_inductive_claim()` alongside the CCN fix below. Verified: unit tests pass (17/17 assertions, overapproximation-test), and `ctest -L k-induction` is 100% pass on everything the macOS build's mode filter actually runs (THOROUGH is skipped at the CTest-modes level on this platform, not just in CI; see the reply on kinductor_01/test.desc for how those five were checked instead).

src/goto-symex/symex_main.cpp
pierredantas: Extracted the five-term condition into `should_convert_inductive_claim()` (declared in goto_symex.h next to `is_assume_false`). `claim()` now calls it as a single predicate, which should bring core CCN back under 15 -- I don't have `lizard` locally to confirm the exact number the CI report uses, so please treat that as expected-but-unverified rather than measured.

unit/goto-symex/overapproximation.test.cpp
pierredantas: Rewrote the comment: the `== 1` counts reflect the assert-to-assume conversion (#6443) collapsing the loop's four unwound dereference checks down to just the last unwinding's surviving as a real assertion, not four checks as the old comment claimed.

unit/goto-symex/overapproximation.test.cpp
pierredantas: Confirmed -- rebuilt with the `!filtered.empty()` guard deleted (object map unconditionally replaced by the empty `filtered` map), re-ran this test case alone: it still passes (3/3 assertions). `asserts_commented` is blind to this mutation.

Traced it as far as I could without expanding scope on this PR: the emptied `object_map` and the pre-filter all-sink map both appear to read back through `value_sett`'s own value-set query path (`src/pointer-analysis/value_set.h`) as effectively "unknown," so `known_exhaustive` in `dereferencet::dereference` ends up false either way and the same fallback `invalid_object` symbol gets synthesized -- meaning the test's assertion counts never actually depend on whether the guard ran. That's pre-existing behavior in the value-set subsystem, not something #6451 introduces, so I haven't touched it here. Filing a separate issue for it rather than fixing it in this PR.

regression/k-induction/kinductor_01/test.desc
pierredantas: Confirmed the five flagged THOROUGH fixtures (github_1735_5, kinductor_01 and sum03 under both k-induction/ and k-induction-parallel/, plus their k-induction-parallel twins -- 5 total) by running each `main.c` directly against a fresh build with today's fix, using the exact flags from its `test.desc`. All five report `VERIFICATION SUCCESSFUL` ("Solution found by the inductive step"), matching the `UNKNOWN -> SUCCESSFUL` flip the diff claims.

Ran directly rather than via `ctest`: this build's `regression/CMakeLists.txt` only puts THOROUGH in `MODES` on non-Apple/non-Windows configs, and `ctest -L k-induction` here reports these five (and the rest of THOROUGH) as Skipped for that reason, on both macOS CI and any local macOS ctest run -- not just the CI matrix. Everything ctest did run passed: 100% (201/201, THOROUGH excluded) on `ctest -L k-induction`.

I don't have a Linux box in this session to run the THOROUGH tier through ctest itself; the direct-invocation results above are the closest I could get to confirming these five without one.

Issue #6464, opened 2026-07-28
[cpp] placement-new reports a spurious heap out-of-bounds under --incremental-bmc / --k-induction


lucasccordeiro:
## Summary

A placement-new into `malloc`'d storage is reported as an out-of-bounds heap write under `--incremental-bmc` and `--k-induction`, while plain BMC (default, and every explicit `--unwind N`) proves the same program safe. The program is clean under `clang++ -std=c++17 -fsanitize=address,undefined`.

```cpp
#include <new>
#include <cstdlib>

struct Elem { char a[8]; };

struct Holder
{
  Elem *buf; int n; int cap;

  Holder()
  {
    buf = (Elem *)malloc(sizeof(Elem) * 10);
    if (!buf) abort();
    n = 0; cap = 10;
  }

  // Never executed below (n is 0, cap is 10), but its presence is what makes
  // the placement-new in add() report a spurious out-of-bounds heap write.
  void grow()
  {
    Elem *nb = (Elem *)malloc(sizeof(Elem) * 20);
    if (!nb) abort();
    buf = nb; cap = 20;
  }

  void add(const Elem &x)
  {
    if (n == cap) grow();
    new (buf + n) Elem(x);
    n++;
  }
};

int main()
{
  Holder h;
  Elem e; e.a[0] = 1;
  h.add(e);
  return 0;
}
```

```
$ esbmc final.cpp --std c++17                                  -> VERIFICATION SUCCESSFUL
$ esbmc final.cpp --std c++17 --unwind 4 --no-unwinding-assertions -> VERIFICATION SUCCESSFUL
$ esbmc final.cpp --std c++17 --incremental-bmc                 -> VERIFICATION FAILED
$ esbmc final.cpp --std c++17 --k-induction                     -> VERIFICATION FAILED

Violated property:
  file final.cpp line 4 column 8 function Elem
  dereference failure: array bounds violated: heap object
```

Line 4 is `struct Elem`, i.e. the implicitly-generated copy constructor; the violated write is its `*this = ...`.

## Ingredients (each one bisected)

Both of these are required; removing either makes the verdict `SUCCESSFUL`:

[...]

Issue #6368, opened 2026-07-24
[cpp] std::vector of a non-trivial type (e.g. vector<vector<int>>) fails: assignment into raw malloc'd storage


lucasccordeiro:
## Summary

A `std::vector` whose element type is itself a non-trivial container (e.g.
`std::vector<std::vector<int>>`) fails verification with a dereference/invalid-free
property violation as soon as an element is stored:

```cpp
#include <vector>
#include <cassert>
int main() {
  std::vector<std::vector<int>> v;
  std::vector<int> row; row.push_back(5);
  v.push_back(row);            // VERIFICATION FAILED (dereference)
  assert(v.size() == 1);
  return 0;
}
```

clang compiles and runs this correctly. Even `v.push_back(row); assert(v.size()==1)`
fails, so the fault is in storing the element, not in the later read.

## Root cause

The `<vector>` model (`src/cpp/library/vector`) allocates its backing store with
`malloc` (raw, uninitialised memory) and writes elements by **assignment** —
`buf[_size] = x;` in `push_back`, and similar in the constructors / `reserve` /
`insert` / `resize` / `assign`. For a trivial element type this is a byte copy and
is fine. For a non-trivial element type it invokes the element's
`operator=`, and `vector::operator=` begins with:

```cpp
free(buf);                       // <-- buf here is the *uninitialised* slot's
buf = static_cast<T *>(malloc(...));  //     garbage pointer
```

So assigning a `vector<int>` into an uninitialised outer-buffer slot calls
`free()` on a garbage pointer → the dereference/invalid-free violation.
`std::vector<std::string>` does *not* hit this because the string model stores a
fixed `char[]` (no heap pointer to free), so a garbage slot has no bad pointer.

## Fix options (both investigated)

1. **Zero the storage** (`calloc` instead of `malloc`, and calloc+copy instead of
   `realloc`, at all ~18 allocation sites): a fresh slot then has a null `buf`
   pointer, so `free(null)` is a safe no-op and assignment works. This is
   **correct** (verified: `vector<vector<int>>` push/grow/resize/assign all pass,
   `vector<int>`/`vector<string>` unaffected) but **prohibitively slow** — the
   default capacity is 10, so every fresh outer vector symbolically zeroes 10
   inner-vector objects, and the basic nested case times out (>150s vs ~3s for a
   plain vector). Not shippable as-is.

2. **Placement-new construction** (`new (&buf[_size]) T(x)` for slots being
[...]

PR #6465, opened 2026-07-28
[test] Pin the placement-new false positive under incremental BMC

lucasccordeiro:
Refs #6464. Blocks #6368.

## Root cause

`src/clang-cpp-frontend/clang_cpp_convert.cpp:836-899` lowers placement-new to `comma(<ctor call retargeted at *(T *)place>, (T *)place)`, so the constructor's `this` is the placement address:

```
```

The dereference check raised inside that constructor loses the precise value-set of the address, so it degenerates to "must be in bounds for every dynamic object in scope":

```
```

Introducing a second `malloc` — even on a guarded path that is never taken — is then enough to make it unprovable, and `--incremental-bmc` / `--k-induction` report `dereference failure: array bounds violated: heap object`. Plain BMC (default, and every explicit `--unwind N`) proves the same program safe.

Both ingredients were bisected:

- **placement-new is required.** `buf[n] = x`, `*(buf + n) = x`, `*(Elem *)((void *)(buf + n)) = x`, and a free function `f(Elem *p, const Elem &x) { *p = x; }` called either as `f(buf + n, x)` or as `f((Elem *)((void *)(buf + n)), x)` all verify. A `void *` round-trip across a function boundary is therefore not the trigger — the generated constructor call is.
- **A second `malloc` on a guarded path is required.** Emptying `grow()` or replacing its body with `cap = m;` makes the verdict `SUCCESSFUL`. Removing only `free(buf)` from it does not.

Not involved: element size (reproduces from `char a[8]` to `char a[132]`), the second allocation's size (a fixed `sizeof(Elem) * 1000` still fails), and `__ESBMC_assume`-ing either size.

## What this PR does

Tests only — no behaviour change. The defect is in the dereference/value-set handling of the constructor call that placement-new generates, which is a separate fix from the reproducer.

## Tests

- `esbmc-cpp/cpp/github_6464_placement_new_incremental` — `KNOWNBUG`, `--std c++17 --incremental-bmc`, expects `VERIFICATION SUCCESSFUL`. Flips to a pass once the defect is fixed.
- `esbmc-cpp/cpp/github_6464_placement_new_bmc` — `CORE`, `--std c++17 --unwind 4 --no-unwinding-assertions`, expects `VERIFICATION SUCCESSFUL`. Keeps the mode that currently gets it right from regressing while the KNOWNBUG is open.

Both sources are self-contained, use no ESBMC intrinsics, and exit 0 under `clang++ -std=c++17 -fsanitize=address,undefined`.

`ctest -R github_6464` passes 2/2. Full `esbmc-cpp/cpp` is unchanged at 657/664, with the same 7 pre-existing macOS failures as `master`.

## Why it matters

This blocks #6368. Constructing `std::vector` elements with placement-new instead of assigning into raw storage is the correct and fast fix there, but with this defect it would make every growable `std::vector<T>` — including `std::vector<std::string>` — report a spurious `VERIFICATION FAILED` in these two modes. The alternative of zeroing raw slots before assigning avoids the imprecision but is prohibitively slow. Details recorded on #6368.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_6464_placement_new_bmc/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6464_placement_new_bmc/main.cpp
@@ -1,0 +1,64 @@
+// github #6464 (control): the same program under plain BMC, where ESBMC gets
+// it right. The --incremental-bmc / --k-induction variant is pinned as a
+// KNOWNBUG next door; this test keeps the working mode from regressing while
+// that one is fixed. Clean under clang++ -std=c++17 -fsanitize=address,undefined.
+//
+// Two ingredients are needed, both bisected on the issue: the placement-new
+// itself (any plain assignment through the same address is fine, including via
+// a void* round-trip across a function boundary), and a second malloc on a
+// guarded path -- here grow(), which is never executed.
+//
+// This blocks #6368: constructing vector elements with placement-new is the
+// right fix there, but it would make every growable std::vector<T> report a
+// spurious failure in these two modes.
+#include <new>
+#include <cstdlib>
+
+struct Elem
+{
+  char a[8];
+};
+
+struct Holder
+{
+  Elem *buf;
+  int n;
+  int cap;
+
+  Holder()
+  {
+    buf = (Elem *)malloc(sizeof(Elem) * 10);
+    if (!buf)
+      abort();
+    n = 0;
+    cap = 10;
+  }
+
+  // Never executed below (n is 0, cap is 10), but its presence is what makes
+  // the placement-new in add() report a spurious out-of-bounds heap write.
+  void grow()
+  {
+    Elem *nb = (Elem *)malloc(sizeof(Elem) * 20);
+    if (!nb)
+      abort();
+    buf = nb;
+    cap = 20;
+  }
+
+  void add(const Elem &x)
+  {
+    if (n == cap)
+      grow();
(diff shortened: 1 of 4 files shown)


lucasccordeiro:
Superseded by #6470, which fixes the defect rather than pinning it.

This PR was opened when the cause looked like a value-set limitation in placement-new's lowering — deep enough that capturing the reproducer as a KNOWNBUG seemed the useful deliverable. It turned out to be narrower: `goto_symex_statet::constant_propagation` bails on `with2t` under `--incremental-bmc` and `--k-induction` for performance, and the bail was unconditional, so struct chains were dropped too and a struct's pointer field never resolved to the object it points at. Exempting structs fixes it, with no verdict changes and an 11.5% runtime drop across the 253 regression tests using those two modes.

Both tests here are carried over into #6470, with the KNOWNBUG flipped to `CORE` and a third added for `--k-induction`, the other affected mode:

- `esbmc-cpp/cpp/github_6464_placement_new_incremental` (`KNOWNBUG` → `CORE`)
- `esbmc-cpp/cpp/github_6464_placement_new_kinduction` (new)
- `esbmc-cpp/cpp/github_6464_placement_new_bmc` (unchanged)

Nothing is lost by closing this. Branch `test/6464-placement-new-incremental-knownbug` is left in place; say the word if you want it deleted.

PR #6470, opened 2026-07-28
[goto-symex] Keep propagating struct WITH chains under incremental BMC

### LINKED ISSUE #6464: [cpp] placement-new reports a spurious heap out-of-bounds under --incremental-bmc / --k-induction

lucasccordeiro:
Fixes #6464. Unblocks #6368. Supersedes #6465, which carried the same reproducer as a KNOWNBUG — its two tests are included here with the KNOWNBUG flipped to CORE.

## Root cause

`goto_symex_statet::constant_propagation` refuses to propagate `with2t` expressions under `--incremental-bmc` and `--k-induction` (`src/goto-symex/goto_symex_state.cpp:166`). The rationale is performance: propagating WITH chains through the repeated symex runs those strategies perform is expensive, and neither relies on constants for loop bounds since both unwind from min k regardless.

The bail is unconditional, so it also drops **struct** chains, short-circuiting the carefully-guarded chain walk immediately below it. A struct built field by field arrives at symex as

```
h#3 == (h#0 WITH [n:=0] WITH [cap:=1] WITH [buf := (Elem *)malloc_result])
```

and without propagation `h.buf` never resolves to the object it points at. The value-set of the address stays ambiguous, which matters as soon as something takes the address's *object* rather than reading through it. Placement-new does exactly that: the frontend lowers it to a constructor call whose `this` is the placement address, so the bounds check raised inside the constructor degenerates to

```
```

— it must hold for every dynamic object in scope. One allocation is fine; a second one anywhere on a guarded path, **even one that is never executed**, makes it unprovable and ESBMC reports `dereference failure: array bounds violated: heap object`. Plain BMC, at any `--unwind`, proves the same program safe.

That explains each ingredient bisected on the issue: only these two modes fail (they are exactly the two options in the condition); only placement-new fails (a plain assignment through the same address is dereferenced normally and stays in bounds for each element of the value set, including via a `void *` round-trip across a function boundary); and a second `malloc` is required (with one object the check still discharges). Element size and the second allocation's size are irrelevant, which is why no assumption added to the model could discharge it locally.

## Solution

Exempt struct types from the bail. Struct chains now fall through to the existing walk, which already bails on anything but fixed-size scalar/pointer/aggregate updates and carries its own rationale for pointer fields. Array and union chains — where the cost the heuristic was written for actually lives — keep the fast bail unchanged, as does the pre-existing Python carve-out.

The narrow form matters: a struct's pointer fields are precisely what the value-set analysis needs, and they are cheap to propagate.

## Tests

- `esbmc-cpp/cpp/github_6464_placement_new_incremental` — `--std c++17 --incremental-bmc`, expects `VERIFICATION SUCCESSFUL` (was the KNOWNBUG in #6465).
- `esbmc-cpp/cpp/github_6464_placement_new_kinduction` — new, same program under `--k-induction`, the other affected mode.
- `esbmc-cpp/cpp/github_6464_placement_new_bmc` — `--unwind 4 --no-unwinding-assertions`, the mode that was always correct, so a future fix cannot trade one mode for the other.

All three are self-contained, use no ESBMC intrinsics, and exit 0 under `clang++ -std=c++17 -fsanitize=address,undefined`.

## Results

Everything below is compared against a binary built from `master` on the same machine, so pre-existing macOS failures are distinguished from regressions.

**No verdict changes, and faster.** Across all 253 regression tests whose `test.desc` uses `--k-induction` or `--incremental-bmc` (the `k-induction/` and `esbmc/` suites), every verdict is unchanged and total runtime drops from 125.4s to 111.0s (**-11.5%**); the worst single regression is +0.28s. Restoring struct propagation folds constants earlier and pays for itself in solver time, so the performance concern behind the original heuristic does not apply to structs.

**Suites:** unit tests 564/564; `k-induction` 122/122; `k-induction-parallel` 74/74; `cbmc` 307/307; `esbmc-unix2` 186/186; `esbmc` 1543/1551; `esbmc-unix` 324/328; `floats` 90/92; `esbmc-cpp/cpp` 663/666; and `esbmc-cpp/{vector,string,try_catch,map,list,algorithm,deque,set,stream,inheritance,bug_fixes,OM_sanity_checks}` plus `esbmc-cpp11/17/20` all 100%.

Every failure above was run under both binaries and is byte-identical to `master` — all of them are macOS `PARSING ERROR`s or environment issues (`floats/underflow8` is an SMT solver failure on both).

**Suites unaffected by construction, checked anyway.** `mopsa` is the Python frontend suite and Python was already exempt from this bail, so the change is a no-op there; a 51-test sample shows no verdict changes (`mopsa/list_in_loop` times out on both binaries, which is why the full suite exceeds a short ctest budget). `esbmc-solidity` cannot be validated on macOS (the sol64 models are empty on darwin), so a 40-test random sample was run under both binaries — no verdict differences.

## Follow-up

This unblocks #6368. The placement-new element-lifetime rework for `std::vector` was implemented and measured while diagnosing this (correct, ~4s on the nested case against >150s for the calloc approach) but could not ship while `vector<std::string>` reported a spurious failure in these two modes. With this merged it can be re-applied; details are recorded on that issue.


lucasccordeiro:
This PR (100s): 

```
  correct:           22860
    correct true:    15323
    correct false:    7537
  incorrect:             9
    incorrect true:      0
    incorrect false:     9
  unknown:           13734
  Score:             38039 (max: 61276)

https://github.com/esbmc/esbmc/actions/runs/30372486084
```

```
  correct:           22819
    correct true:    15321
    correct false:    7498
  incorrect:             7
    incorrect true:      0
    incorrect false:     7
  unknown:           13777
  Score:             38028 (max: 61276)

https://github.com/esbmc/esbmc/actions/runs/29587164519/job/87913576991
```

mikhailramalho:
I'm confused about the fix:

> without propagation h.buf never resolves to the object it points at.

Constant propagation shouldn't be required for correctness. ESBMC should work the same even if we disable constant propagation (e.g., pass `--no-propagation`)

lucasccordeiro:
Fair point, @mikhailramalho. It should work even with `--no-propagation'. I'll investigate this issue further.

Let me also add @XLiZHI here since he has touched this code at symex before.

lucasccordeiro:
@mikhailramalho you are right, and it is worse than the PR description implies. I ran the experiment you suggested and it does not say what it appears to say, for two reasons.

## 1. `--no-propagation` is currently a no-op

Since 148700569a (2019) the option only sets `goto_symext::constant_propagation` (`src/goto-symex/symex_assign.cpp:29`). That bool is never read anywhere — it is only copied in `operator=` (`symex_assign.cpp:183`). The function that actually decides propagation, `goto_symex_statet::constant_propagation` (`src/goto-symex/goto_symex_state.cpp:98`), never consults it.

So `esbmc file.c --no-propagation` today produces byte-identical SSA to a run without the flag, and the experiment silently passes.

## 2. With the flag genuinely wired up, correctness *does* depend on propagation

I added an early `return false` in `goto_symex_statet::constant_propagation` gated on the option, rebuilt, and re-ran on current master (`e912c6ae00`):

| program | propagation on | `--no-propagation` |
|---|---|---|
| this PR's C++ placement-new reproducer, **plain BMC** | SUCCESSFUL | **FAILED** — same `array bounds violated: heap object` |
| same defect in **pure C**, no placement-new, no C++ | SUCCESSFUL | **FAILED** |
| **no struct at all**, plain local pointer | SUCCESSFUL | **FAILED** |

The trigger is neither structs nor placement-new. The minimal shape is a pointer merged over a **never-taken** branch that reassigns it from a second `malloc`:

```c
#include <stdlib.h>
struct Elem { char a[8]; };
int main() {
  struct Elem *buf; int n = 0, cap = 10;
  buf = (struct Elem *)malloc(sizeof(struct Elem) * 10);
  if (!buf) abort();
  if (n == cap) {                                  /* never taken */
    struct Elem *nb = (struct Elem *)malloc(sizeof(struct Elem) * 20);
    if (!nb) abort();
    buf = nb;
  }
  struct Elem e; e.a[0] = 1;
  *(buf + n) = e;
  return 0;
}
```

`esbmc v2.c --unwind 4 --no-unwinding-assertions` → SUCCESSFUL; with a working `--no-propagation` → `dereference failure: array bounds violated: heap object`. No struct, no C++, no incremental BMC.

`--show-vcc` on that program is suggestive. Of the three claims emitted at the write, two are guarded as expected:

```
{1} guard_exec => (... => !(SAME-OBJECT(buf#4 + n#1, &dynamic_1_array)) || <in bounds of dynamic_1>)
```

but one asserts the bounds obligation with no `SAME-OBJECT` disjunct at all:

```
{1} guard_exec => (... => POINTER_OFFSET(buf#4 + n#1) <= alloc_size#4[POINTER_OBJECT(&dynamic_1_array)] && ...)
```

That looks like guard elision on the final value-set alternative in `src/pointer-analysis/dereference.cpp`, but I have not confirmed it yet.

## What this means for the PR

There are two nested defects:

- **D1** — the propagation-independent spurious out-of-bounds above.
- **D2** — the `with2t` bail under `--incremental-bmc`/`--k-induction` disables struct propagation, which exposes D1 for pointers held in struct fields. That is what #6464 reports, and its minimal reproducer is also pure C (~20 lines), not C++ placement-new.

This PR fixes D2's *symptom* by restoring propagation. It does not touch D1. So the fix is a workaround, and the three C++ regression tests here mis-attribute the bug to placement-new.

## The SV-COMP delta

I pulled both `esbmc-result` artifacts and diffed the per-run `category` column. Exactly two tasks are newly incorrect:

| task | property (expected) | master | this PR |
|---|---|---|---|
| `ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5` | no-overflow (true) | `TIMEOUT` | `false(no-overflow)` |
| `goblint-regression/09-regions_13-arraycollapse_nr` | valid-memsafety (true) | `TIMEOUT (false(valid-memtrack))` | `false(valid-memtrack)` |

Nothing went correct → wrong, and no previously-wrong task was fixed. Both flipped **from a timeout**, and master's status for the goblint task is literally `TIMEOUT (false(valid-memtrack))` — master had already computed the same wrong verdict and merely ran out of time. So these are pre-existing false alarms unmasked by the speedup rather than new unsound reasoning. That is defensible on its own, but they are two more false alarms of the same family as #6464, which argues for fixing D1 rather than shipping this.

## Proposed plan

1. Make `--no-propagation` real, as its own small PR with regression tests. Without it there is no oracle for "correctness must not depend on propagation".
2. Root-cause D1 against the struct-free reproducer above.
3. Fix D1. Acceptance gate: every reproducer verifies SUCCESSFUL under plain BMC, `--incremental-bmc` and `--k-induction`, each **with and without** `--no-propagation`.
4. Re-scope this PR. If (3) closes #6464, drop the struct-WITH exemption entirely and keep only the tests, replacing the three near-identical C++ cases with the pure-C reproducer in `regression/esbmc/` plus one C++ placement-new case. If the −11.5% runtime win is still wanted, it goes in a separate PR justified on performance alone, with its own SV-COMP run.
5. Reproduce the two SV-COMP tasks locally and file them separately if they are indeed pre-existing; check whether either is another instance of D1.

I will hold this PR until at least step 3. @XLiZHI, input welcome on the dereference side.

mikhailramalho:
I think we have an issue open from a long time ago to add a test run with --no-propagation (and possibly --incremental-bmc for cases that need it) that would've caught this. 

We could also hide this flag and leave it as debug-only. I don't think any user ever wants to not propagate the constant and make the verification slower.

lucasccordeiro:
Closing in favour of #6506, which fixes the real cause of #6464.

@mikhailramalho's objection was right. The defect is in `dereference.cpp`, not in constant propagation: `construct_struct_ref_from_dyn_offset` seeded its recursion with a true guard and the byte-array arm built a fresh empty one, so the bounds check lost both the path guard and the same-object guard that `build_reference_to` had already applied to the sibling check at `dereference.cpp:810`. Each value-set target's claim then had to hold for every dynamic object at once, and a second `malloc` on a never-taken branch made it unprovable.

With that fixed, the struct-WITH exemption in this PR is unnecessary — the C++ placement-new test from this branch passes on #6506 with no propagation change at all, under plain BMC, `--incremental-bmc` and `--k-induction`, with and without `--no-propagation`. That test is carried over to #6506, alongside smaller pure-C reproducers, since the bug was never specific to placement-new or to C++.

`--no-propagation` turned out to have been inert since 148700569a (2019) — it set a bool nothing read — which is why the experiment suggested in this thread appeared to pass. #6505 fixes that and is what made the underlying defect visible.

Two things from this PR are not superseded and are worth pursuing separately:

- The **-11.5% runtime win** from restoring struct WITH propagation under `--incremental-bmc`/`--k-induction` still stands on its own as a performance change, and should be re-measured and proposed independently of any correctness claim.
- The two new `incorrect false` SV-COMP results reported above (`ldv-commit-tester/m0_drivers-media-rc-imon--32_7a--a9e7fb5`, `goblint-regression/09-regions_13-arraycollapse_nr`) both flipped *from a timeout*, and master's status for the second is literally `TIMEOUT (false(valid-memtrack))` — master had already computed the same wrong verdict and merely ran out of time. They look like pre-existing false alarms unmasked by the speedup rather than new unsound reasoning, but that should be confirmed by a benchexec run rather than assumed; neither reproduces on macOS.

PR #6501, opened 2026-07-29
[contracts] reject __ESBMC_is_fresh on a non-pointer object

### LINKED ISSUE #6469: [contracts] is_fresh crashes on the address of a stack-local variable (works on parameters/malloc)
## Summary

`__ESBMC_is_fresh(&local, sizeof(...))` crashes ESBMC (`irep2_cast_error`)
whenever the argument is the address of a **plain stack-local variable**
declared inside the function itself, regardless of struct complexity (a
single flat `int16_t[256]` struct is enough) and regardless of whether the
contract is later replaced at any call site -- the crash happens on a bare
`--enforce-contract` of the function alone, with **zero** callees. The
identical pattern works fine when the same object is a **function
parameter**, or when the local is heap-allocated (`malloc`) instead of a
plain stack declaration.

## Minimal reproducer

Crashes:
```c
#include <stdint.h>
#define MLKEM_N 256
typedef struct { int16_t coeffs[MLKEM_N]; } mlk_poly;

void test_fn(void)
{
  mlk_poly v;
  __ESBMC_requires(__ESBMC_is_fresh(&v, sizeof(mlk_poly)));
}

int main(void) { return 0; }
```
Run as:
```
esbmc repro.c --function test_fn --enforce-contract test_fn --no-pointer-check --unwind 8 --z3
```
Output:
```
WARNING: Could not find definition for temporary variable: c:@F@test_fn::$tmp::return_value$___ESBMC_is_fresh$1
Enforced contract for function test_fn
libc++abi: terminating due to uncaught exception of type irep2_cast_error: irep2: struct_union_members() called on incompatible type (type_id = pointer)
```

Does **not** crash -- identical struct, `is_fresh` applied to a parameter instead:
```c
#include <stdint.h>
#define MLKEM_N 256
typedef struct { int16_t coeffs[MLKEM_N]; } mlk_poly;

void test_fn(mlk_poly *v)
{
  __ESBMC_requires(__ESBMC_is_fresh(v, sizeof(mlk_poly)));
}

int main(void) { return 0; }
```
`VERIFICATION SUCCESSFUL` (trivially, 0 VCCs, empty body).

Does **not** crash -- same local, heap-allocated instead of stack-declared:
```c
#include <stdint.h>
#include <stdlib.h>
#define MLKEM_N 256
typedef struct { int16_t coeffs[MLKEM_N]; } mlk_poly;

void test_fn(void)
{
  mlk_poly *v = malloc(sizeof(mlk_poly));
  __ESBMC_assume(v != ((void*)0));
  __ESBMC_requires(__ESBMC_is_fresh(v, sizeof(mlk_poly)));
}

int main(void) { return 0; }
```
`VERIFICATION SUCCESSFUL`.

## Confirmed not about struct nesting depth or number of objects

The crash reproduces identically with a single flat struct (above) and with
a 3-level nested struct (`mlk_polymat{ mlk_polyvec{ mlk_poly{ int16_t[256] } } }`),
and with anywhere from 1 to 9 distinct stack-local objects declared in the
same function, each getting its own `is_fresh` requires -- adding more
objects doesn't change the outcome, one is already enough. Confirmed not
about replacement, either: the crash happens with **zero**
`--replace-call-with-contract` flags present, on the bare `--enforce-contract`
of a function whose only content is the stack declarations and the
`is_fresh` requires themselves (no callees, no loop).

## Why this matters

This blocks a very common real-world shape: a top-level API function that
allocates its own intermediate working storage as local structs and passes
their addresses to helper functions, rather than receiving them as
parameters -- e.g. `mlk_indcpa_enc` (`mlkem/src/indcpa.c`), which declares
nine such locals (`mlk_polymat at`, several `mlk_polyvec`, several
`mlk_poly`, a `mlk_polyvec_mulcache`) and is exactly the "one function, deep
call chain" case we wanted to demonstrate compositional verification on.
Every subject built earlier in this benchmark track (`mlk_poly_reduce_c`,
`mlk_poly_add`, `mlk_matvec_mul`, etc.) sidestepped this by always making
the `is_fresh`-guarded object a function **parameter** at every level, never
a local the function declares for itself -- so this specific pattern hadn't
been exercised until now. The workaround (heap-allocate the local, assume
non-null, `is_fresh` the resulting pointer) is straightforward and lets the
faithful contract-composition test proceed, but it's a change to how the
subject is written, not something a real caller of `mlk_indcpa_enc` should
have to do.

## Environment

- ESBMC built from `origin/master` at `a865226d7e` (includes #6326/#6358/
  #6365/#6385/#6437 -- i.e. not a re-occurrence of any of those).
- OS: macOS, x86_64. Solver: Z3 4.16.0 (`--z3`); not solver-specific in any
  way tested (crash occurs before solving starts, during Bounded Model
  Checking setup).

LukeW1999:
Fixes #6469, reported by @adilanwar2399.

`__ESBMC_is_fresh(EXPR, n)` lowers to "allocate n bytes and assign the result into EXPR". Recovering the assignment target peeled an `address_of` unconditionally:

```cpp
if (is_address_of2t(stripped))
  ptr_var = to_address_of2t(stripped).ptr_obj;   // &var -> var
```

That is only coherent when the peeled object is itself a pointer. When it is not, the malloc result is assigned into the object, and the resulting instruction is type-incoherent:

```
ASSIGN v = MALLOC(unsigned char, ...)     // v is a struct, not a pointer
ASSUME v != (struct mlk_poly)NULL
```

Two symptoms follow from the one defect:

- **struct object — abort.** `value_sett::assign` decomposes the struct-typed lhs into components, `make_member` pushes through the typecast `symex_mem` inserted at `memory_alloc.cpp:705`, lands on the `u8*` operand and throws:

  ```
  irep2: struct_union_members() called on incompatible type (type_id = pointer)
  ```

  Backtrace: `symex_mem` → `symex_assign_symbol` → `value_sett::assign` → `value_sett::make_member` (`value_set.cpp:1546`).

- **scalar object — silent.** Not reported in the issue, found while reducing it. `__ESBMC_is_fresh(&x, sizeof(int))` with `int x` reports `VERIFICATION SUCCESSFUL`, because value-set never decomposes a scalar. The lowering is the same nonsense (`ASSIGN x = MALLOC(...)`), it just clobbers `x` and makes the contract vacuous instead of crashing. This is the more dangerous half.

Behaviour across the argument forms, before and after:

| form | before | after |
|---|---|---|
| `is_fresh(p, n)`, `p` a `T *` param | SUCCESSFUL | unchanged |
| `is_fresh(p, n)`, `p` a `T **` param | SUCCESSFUL | unchanged |
| `is_fresh(&p, n)`, `p` a `T *` local | SUCCESSFUL | unchanged |
| `is_fresh(&v, n)`, `v` a struct local | **abort** | diagnosed |
| `is_fresh(&x, n)`, `x` an `int` local | **silently vacuous** | diagnosed |

## The fix

`address_of` is itself pointer-typed, so testing `is_pointer_type(stripped->type)` does not distinguish the two cases; the `address_of` shape has to be matched first and its operand checked. Both shapes were nominally guarded by an `assert`, but `NDEBUG` compiles those out, so release builds diagnosed neither. Replaced with a real check using this file's existing `log_error` + `abort()` idiom (`contracts.cpp:182`, `:205`, `:1992`).

The message points at the two supported ways to say what the reporter wanted:

```
ERROR: __ESBMC_is_fresh needs a pointer it can point at fresh storage, but the
contract of 'f' gives it the address of a non-pointer object. Take that object
as a pointer parameter, or allocate it with malloc and pass the pointer.
```

Both are already confirmed to work in the issue.

## Tests

- `is_fresh_addr_struct_local_fail` — the reporter's reduction, was an abort
- `is_fresh_addr_scalar_local_fail` — the silent variant, was a false SUCCESSFUL
- `is_fresh_addr_ptr_local_pass` — guard: `&p` with `p` a pointer local must keep working, since that peel is the legitimate one

Red before the patch (abort / false pass / pass), green after. `function_contract` and `loop-invariants` are 365/365.

## Not addressed

`value_sett::make_member` asserts `is_struct_type(type) || is_union_type(type)` and then pushes through typecasts without re-checking the operand, so any other producer of such a cast crashes the same way in a release build. This patch removes the only producer I know of rather than hardening `make_member`; hardening it is a reasonable follow-up but is a change to shared analysis code and does not belong in a contracts fix.

### DIFF
--- a/regression/function_contract/is_fresh_addr_array_local_fail/main.c
+++ b/regression/function_contract/is_fresh_addr_array_local_fail/main.c
@@ -1,0 +1,10 @@
+void f(void)
+{
+  int a[10];
+  __ESBMC_requires(__ESBMC_is_fresh(&a, sizeof(a)));
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/function_contract/is_fresh_addr_array_local_fail/test.desc
+++ b/regression/function_contract/is_fresh_addr_array_local_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --no-pointer-check --unwind 4
+^ERROR: __ESBMC_is_fresh needs a pointer it can point at fresh storage
--- a/regression/function_contract/is_fresh_addr_ptr_local_pass/main.c
+++ b/regression/function_contract/is_fresh_addr_ptr_local_pass/main.c
@@ -1,0 +1,15 @@
+typedef struct
+{
+  int coeffs[4];
+} P;
+
+void f(void)
+{
+  P *v;
+  __ESBMC_requires(__ESBMC_is_fresh(&v, sizeof(P)));
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/function_contract/is_fresh_addr_ptr_local_pass/test.desc
+++ b/regression/function_contract/is_fresh_addr_ptr_local_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --no-pointer-check --unwind 4
+^VERIFICATION SUCCESSFUL$
--- a/regression/function_contract/is_fresh_addr_scalar_local_fail/main.c
+++ b/regression/function_contract/is_fresh_addr_scalar_local_fail/main.c
@@ -1,0 +1,10 @@
+void f(void)
+{
+  int x;
+  __ESBMC_requires(__ESBMC_is_fresh(&x, sizeof(int)));
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/function_contract/is_fresh_addr_scalar_local_fail/test.desc
+++ b/regression/function_contract/is_fresh_addr_scalar_local_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --no-pointer-check --unwind 4
+^ERROR: __ESBMC_is_fresh needs a pointer it can point at fresh storage
--- a/regression/function_contract/is_fresh_addr_struct_local_fail/main.c
+++ b/regression/function_contract/is_fresh_addr_struct_local_fail/main.c
@@ -1,0 +1,15 @@
+typedef struct
+{
+  int coeffs[4];
+} P;
+
+void f(void)
+{
+  P v;
+  __ESBMC_requires(__ESBMC_is_fresh(&v, sizeof(P)));
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/function_contract/is_fresh_addr_struct_local_fail/test.desc
+++ b/regression/function_contract/is_fresh_addr_struct_local_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --no-pointer-check --unwind 4
+^ERROR: __ESBMC_is_fresh needs a pointer it can point at fresh storage
--- a/regression/function_contract/is_fresh_array_param_pass/main.c
+++ b/regression/function_contract/is_fresh_array_param_pass/main.c
@@ -1,0 +1,9 @@
+void f(int a[10])
+{
+  __ESBMC_requires(__ESBMC_is_fresh(a, 10 * sizeof(int)));
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/function_contract/is_fresh_array_param_pass/test.desc
+++ b/regression/function_contract/is_fresh_array_param_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --no-pointer-check --unwind 4
+^VERIFICATION SUCCESSFUL$
--- a/src/goto-programs/contracts/contracts.cpp
+++ b/src/goto-programs/contracts/contracts.cpp
@@ -1138,19 +1138,36 @@ goto_programt code_contractst::generate_checking_wrapper(
     // to *p.  Earlier revisions had a "fallback" else branch that emitted
     // *p = malloc(...), which dereferenced an uninitialised p and tripped
     // alignment / pointer-validity checks before the body even ran.
+    // Peeling the address-of is only sound when what it wraps is itself a
+    // pointer.  For &obj with obj a struct or scalar, the peel would assign
+    // the malloc result into obj: value-set analysis then walks a pointer as
+    // if it were a struct and aborts (#6469), and for a scalar obj the
+    // assignment silently clobbers it and the contract means nothing.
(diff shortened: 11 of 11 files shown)


LukeW1999:
Self-review update, measured against master rather than reasoned about.

**The defect is wider than the issue title says.** The report is about a crash on a struct local. Building both binaries and sweeping the argument forms shows the crash is the loud instance of a general problem: the address-of peel is unsound for *any* non-pointer object, and the other instances are silent rather than loud.

| form | master | this branch |
|---|---|---|
| `is_fresh(p, n)`, `p` a `T *` param | SUCCESSFUL | unchanged |
| `is_fresh(p, n)`, `p` a `T **` param | SUCCESSFUL | unchanged |
| `is_fresh(&p, n)`, `p` a `T *` local | SUCCESSFUL | unchanged |
| `is_fresh(a, n)`, `a` an array **param** | SUCCESSFUL | unchanged |
| `is_fresh(s->p, n)`, pointer member | SUCCESSFUL | unchanged |
| `is_fresh(&s->p, n)`, pointer member | SUCCESSFUL | unchanged |
| `is_fresh((void *)p, n)` | SUCCESSFUL | unchanged |
| `is_fresh(&v, n)`, struct local | **abort** | diagnosed |
| `is_fresh(&x, n)`, `int` local | **silently vacuous** | diagnosed |
| `is_fresh(&a, n)`, array local | **silently vacuous** | diagnosed |
| `is_fresh(a, n)`, array local | **silently vacuous** | diagnosed |
| `is_fresh(&s->buf, n)`, array member | **silently vacuous** | diagnosed |

For an array local master emits

```
ASSIGN a[0] = MALLOC(unsigned char, ...)
```

assigning the allocation into the array's first element and reporting `VERIFICATION SUCCESSFUL`. Same shape as the scalar case: value-set never decomposes an `int`, so nothing crashes and the contract quietly means nothing.

None of the five newly-diagnosed forms produced a correct result before, so this rejects nothing that worked. The array *parameter* case is the one that could plausibly have been caught by mistake, since it looks like the array local; it decays to a pointer, is unaffected, and now has a guard test.

**Two things I would flag against my own patch.**

`abort()` is the wrong exit for a user error, and it is inconsistent with the sibling fix in #6502, which routes the same class of mistake through `process_goto_program`'s `return true` and exits 6. Here that would mean giving `generate_checking_wrapper` an error channel it does not have, which is a larger change than this fix warrants; I followed the three existing `log_error` + `abort()` sites in this file instead (`contracts.cpp:182`, `:205`, `:1992`). Worth unifying, separately.

The guards this replaces were `assert`s, so under `NDEBUG` none of these shapes was diagnosed in a release build. That pattern is not unique to this site: `value_sett::make_member` asserts its operand is a struct or union and then pushes through typecasts without re-checking, which is the mechanism the struct case crashes through. This patch removes the only producer I know of rather than hardening shared analysis code.

Issue #6499, opened 2026-07-29
[contracts] ternary ensures checked against the wrong branch when a callee inside the guard is replaced


LukeW1999:
Split out from #6498, which is about the CLI accepting an unresolvable function name. Once names are validated, two `CORE` tests stop being no-ops and both fail. This issue is about *why* they fail; #6498 is about the fact that nothing said they were inert.

Both tests pass `--replace-call-with-contract` a comma-separated string, so today they replace nothing. Expressed with repeated flags, so that the replacement actually happens:

## 1. `regression/function_contract/ring_enforce_process_messages`

```
esbmc main.c --enforce-contract process_messages_at_time \
  --replace-call-with-contract source_reaction_in \
  --replace-call-with-contract node_reaction --no-align-check
```

Isolating which callee is responsible:

| replaced | verdict |
|---|---|
| *(none)* | SUCCESSFUL |
| `node_reaction` | SUCCESSFUL |
| **`source_reaction_in`** | **FAILED** |
| both | FAILED |

`process_messages_at_time` has a single ternary postcondition (`main.c:94`):

```c
__ESBMC_ensures(
    __ESBMC_old(msg_n5_to_s.valid) && __ESBMC_old(msg_n5_to_s.timestamp) == time
    ? (source.reaction_2_fired == true && source.reaction_2_fire_time == time)
    : (source.reaction_2_fired == __ESBMC_old(source.reaction_2_fired) &&
       source.reaction_2_fire_time == __ESBMC_old(source.reaction_2_fire_time) &&
       source.received == __ESBMC_old(source.received)));
```

The counterexample takes the **then** branch. States 27-29 show the replaced `source_reaction_in` contract applied exactly as written, with `reaction_2_fired = 1` and `reaction_2_fire_time = 7 = time`, so the then-arm is satisfied:

```
State 27 file main.c line 104 ... source.received          = {.received=-1, .reaction_2_fired=0, .reaction_2_fire_time=0}
State 28 file main.c line 104 ... source.reaction_2_fired  = {.received=-1, .reaction_2_fired=1, .reaction_2_fire_time=0}
State 29 file main.c line 104 ... source.reaction_2_fire_time = {.received=-1, .reaction_2_fired=1, .reaction_2_fire_time=7}
```

But the property that is checked has lost the guard and the then-arm, keeping one conjunct of the **else** arm:

```
Violated property:
  contract ensures
[...]

PR #6502, opened 2026-07-29
[contracts] reject unknown function names in contract flags

### LINKED ISSUE #6498: [contracts] --replace-call-with-contract silently accepts unknown function names and reports VERIFICATION SUCCESSFUL
## Summary

`--replace-call-with-contract` accepts a function name that resolves to nothing. The run then proceeds as if the flag had never been given, prints `VERIFICATION SUCCESSFUL`, and exits 0. The status line even reports a non-zero count of functions replaced.

The practical consequence is that a typo, or the natural but unsupported comma-separated form `--replace-call-with-contract f,g`, silently degrades a modular verification run into a plain one. Nothing in the output distinguishes the two, so a benchmark or a CI script can report a green modular result for a run in which no contract was ever substituted.

`--enforce-contract` has the same hole, with a warning that is easy to miss and does not affect the verdict or the exit code.

## Reproducer

```c
/* silent.c */
int f(int x)
{
  __ESBMC_ensures(__ESBMC_return_value >= 0);
  return x > 0 ? x : 0;
}

int g(int x)
{
  __ESBMC_ensures(__ESBMC_return_value >= 0);
  return x > 0 ? x : 0;
}

int main(void)
{
  int a = f(3);
  int b = g(3);
  /* True of the bodies (a == 3), not of the contracts (a >= 0 only),
     so this must FAIL whenever f and g are actually replaced. */
  __ESBMC_assert(a == 3 && b == 3, "true of the bodies, not of the contracts");
  return 0;
}
```

| invocation | reported count | verdict |
|---|---|---|
| *(no flag)* | 0 | SUCCESSFUL |
| `--replace-call-with-contract f` | 1 | FAILED |
| `--replace-call-with-contract f --replace-call-with-contract g` | 2 | FAILED |
| `--replace-call-with-contract "*"` | 2 | FAILED |
| **`--replace-call-with-contract f,g`** | **1** | **SUCCESSFUL** |
| **`--replace-call-with-contract typo_that_does_not_exist`** | **1** | **SUCCESSFUL** |

The last two rows are the bug. Both report `Replacing calls with contracts for 1 function(s)` while replacing zero, and both exit 0.

## Root cause

`process_function_list` in `src/esbmc/parseoptions/function_contracts.cpp:63-80` inserts every non-`*` argument into the set verbatim and never resolves it:

```cpp
      if (func == "*")
      {
        result = collect_functions_with_contracts();
        break;
      }
      else
      {
        result.insert(func);      // <-- no validation
      }
```

`code_contractst::find_function_symbol` is the predicate the downstream consumers use, and it is never applied at this point. Downstream:

- `enforce_contracts` does resolve the name, logs `Function {} not found` and then `continue`s (`src/goto-programs/contracts/contracts.cpp:752`), so the run completes and reports a verdict.
- `replace_calls` does not check at all: an unmatched name simply never matches any key of its `function_map` (`src/goto-programs/contracts/contracts.cpp:3520-3552`), so it is a silent no-op.

## Inconsistency within the same feature

| flag | unknown name | exit | verdict |
|---|---|---|---|
| `--function` | `ERROR: main symbol 'typo_xyz' not found` | 0 | no verification runs |
| `--enforce-contract` | `WARNING: Function typo_xyz not found` | 0 | `VERIFICATION SUCCESSFUL` |
| `--replace-call-with-contract` | *(silent)* | 0 | `VERIFICATION SUCCESSFUL` |

`--function` already treats an unresolvable name as an error. The two contract flags should agree with it: all three change *what* is verified, so a name that does not resolve means the verification that ran is not the one requested, and a green verdict for it is misleading.

## Suggested fix

Validate once in the shared entry point, so both flags are covered in one place:

```cpp
  auto process_function_list = [&](const std::list<std::string> &func_list,
                                   const char *opt) {
    std::set<std::string> result;
    for (const auto &func : func_list)
    {
      if (func == "*")
        return collect_functions_with_contracts();

      if (contracts.find_function_symbol(func) == nullptr)
      {
        log_error(
          "--{}: no function named '{}'{}", opt, func,
          func.find(',') != std::string::npos
            ? " (names are not comma-separated: repeat the flag once per "
              "function, or use \"*\")"
            : "");
        abort();
      }
      result.insert(func);
    }
    return result;
  };
```

The comma-specific hint is what turns the reported failure mode into a self-explaining one; without it the error is still correct but the user has to guess why `f,g` is not a function.

Two notes on scope:

- Whether comma lists should additionally be *supported* is a separate design question. Making them a hard error is enough to close this: `*` and repeated flags already cover the use case, and `--function` takes a single name, so accepting a second syntax here would make the three flags disagree again.
- The status line is emitted twice (`function_contracts.cpp:111` and `contracts.cpp:3522`) and counts the requested set rather than the matched set. Once names are validated the count is accurate by construction, but the duplicate print is worth removing while touching this code.

Happy to put up the patch and regression tests (one that errors on an unknown name, one that errors on the comma form, one that confirms the repeated-flag form still replaces both functions).

LukeW1999:
## The problem

Every contract flag silently accepted a name it could not act on, did nothing, and still printed a verdict. `esbmc x.c --replace-call-with-contract typo` reported `VERIFICATION SUCCESSFUL` for a run that verified the program with no contract applied at all. The same held for a name with no body, a name with no contract, a comma-separated list, a C++ id with parameters, `"*"` on a file whose contract macros did not survive preprocessing, and both `--*-all-contracts` flags.

Fixes #6498.

## The change

One eligibility rule, `code_contractst::diagnose_contract_target`, asked once for every flag **before any pass runs**.

It is asked per option, because the two passes' rules genuinely differ and always have: `enforce_contracts` resolves through `find_function_symbol`, `replace_calls` selects through `matches_replace_pattern`. A C++ free function `int fst(int)` has the id `c:@F@fst#I#`, whose tail after the last `@` with one trailing `#` stripped is `fst#I`, not `fst` — so it satisfies enforce and not replace. That divergence is why #6498 stayed open for C++ after the first attempt here, and asking the right rule per option is what closes it.

Checking before any pass runs is not a style preference. `enforce_contracts` rewrites the functions it acts on into wrappers that carry no contract, so a check made afterwards calls a name that *was* used unusable. It rejects `--enforce-contract f --replace-call-with-contract "*"` — the shape four existing tests use — and rejects two valid spellings of one function.

Also here:

- `replace_calls` reports the count it acted on rather than the count requested, and emits that line once instead of twice. `contract_fn_name_repeated` can then assert it and mean something.
- The diagnostics no longer contradict each other on an ambiguous C++ name: the cause is reported by the code that knows it, and the caller's line does not restate a different one.
- `find_function_symbol` stays private.

## Behaviour

```
$ esbmc x.c --replace-call-with-contract typo
ERROR: --replace-call-with-contract: cannot use 'typo': no function of that name has a body here

$ esbmc x.c --replace-call-with-contract f,g
ERROR: --replace-call-with-contract: cannot use 'f,g': no function of that name has a body here (names are not comma-separated; repeat the flag once per function, or use "*")

$ esbmc x.cpp --replace-call-with-contract fst        # int fst(int)
ERROR: --replace-call-with-contract: cannot use 'fst': no function of that name has a body here

$ esbmc nc.c --replace-call-with-contract "*"         # no contracts in the file
ERROR: --replace-call-with-contract: no function with a contract to act on

$ esbmc x.c --enforce-all-contracts                   # contracts, but none annotated
ERROR: --enforce-all-contracts: no function carries the __ESBMC_contract annotation
```

No verdict is printed in any of these, and the exit status is 6.

## Tests

Nine under `regression/function_contract/`: `contract_fn_name_unknown`, `_comma`, `_enforce_unknown`, `_no_body`, `_no_contract`, `_repeated`, `_star_no_contract`, `_all_no_annotation`, and `_cpp_paramspp`, the last pinning the C++ case above. Each error test asserts the exact message **and** `\A(?!(.|\n)*VERIFICATION)`, so a run that prints the error and then goes on to a verdict fails the test — the runner does not consult the exit status, so the absence of a verdict has to be asserted positively.

`ctest -L function_contract` 314/314, `ctest -L loop-invariants` 73/73.

## Known and deliberate

- **A misspelling next to `"*"` is not diagnosed.** `--replace-call-with-contract typo --replace-call-with-contract "*"` runs without complaint, in either order. The wildcard resolves to full symbol ids while users spell short names, so the two cannot be compared without rejecting every legitimate spelling.
- **A name whose only contract is `__ESBMC_assigns()` is now rejected** rather than silently ignored. `has_contracts` does not recognise `contract::assigns_empty`, although `contracts.cpp` handles that comment elsewhere, so an empty frame condition has never actually worked. Surfacing it is an improvement over silence; making it work is a separate change and I will file it.
- **A name that resolves to a contracted function with no call site is accepted.** Replacing calls where there are none is not a usage error, and erroring would break `"*"` over a file with unreachable code.

### DIFF
--- a/regression/function_contract/contract_fn_name_no_body/test.desc
+++ b/regression/function_contract/contract_fn_name_no_body/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract extern_fn
+^ERROR: --replace-call-with-contract: cannot use 'extern_fn': that function is declared but has no body here$
--- a/regression/function_contract/contract_fn_name_no_body/test.desc
+++ b/regression/function_contract/contract_fn_name_no_body/test.desc
@@ -1,4 +1,5 @@
 CORE
 main.c
 --replace-call-with-contract extern_fn
-^ERROR: --replace-call-with-contract: cannot use 'extern_fn': that function is declared but has no body here$
+^ERROR: --replace-call-with-contract: cannot use 'extern_fn': no function of that name has a body here$
+\A(?!(.|\n)*VERIFICATION)
--- a/regression/function_contract/contract_fn_name_repeated/test.desc
+++ b/regression/function_contract/contract_fn_name_repeated/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract f --replace-call-with-contract g
+^VERIFICATION FAILED$
--- a/regression/function_contract/contract_fn_name_repeated/test.desc
+++ b/regression/function_contract/contract_fn_name_repeated/test.desc
@@ -2,3 +2,4 @@ CORE
 main.c
 --replace-call-with-contract f --replace-call-with-contract g
 ^VERIFICATION FAILED$
+^Replacing calls with contracts for 2 function\(s\)$
--- a/regression/function_contract/contract_fn_name_unknown/test.desc
+++ b/regression/function_contract/contract_fn_name_unknown/test.desc
@@ -1,4 +1,5 @@
 CORE
 main.c
 --replace-call-with-contract no_such_function
-^ERROR: --replace-call-with-contract: cannot use 'no_such_function': no function of that name$
+^ERROR: --replace-call-with-contract: cannot use 'no_such_function': no function of that name has a body here$
+\A(?!(.|\n)*VERIFICATION)
--- a/regression/function_contract/contract_fn_name_unknown/test.desc
+++ b/regression/function_contract/contract_fn_name_unknown/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --replace-call-with-contract no_such_function
-^ERROR: --replace-call-with-contract: no function named 'no_such_function'$
+^ERROR: --replace-call-with-contract: cannot use 'no_such_function': no function of that name$
--- a/regression/function_contract/contract_fn_name_unknown/test.desc
+++ b/regression/function_contract/contract_fn_name_unknown/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract no_such_function
+^ERROR: --replace-call-with-contract: no function named 'no_such_function'$
--- a/regression/function_contract/ring_enforce_process_messages/test.desc
+++ b/regression/function_contract/ring_enforce_process_messages/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --enforce-contract process_messages_at_time --replace-call-with-contract source_reaction_in --replace-call-with-contract node_reaction --no-align-check
 ^VERIFICATION SUCCESSFUL$
--- a/regression/function_contract/ring_enforce_process_messages/test.desc
+++ b/regression/function_contract/ring_enforce_process_messages/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --enforce-contract process_messages_at_time --replace-call-with-contract source_reaction_in --replace-call-with-contract node_reaction --no-align-check
 ^VERIFICATION SUCCESSFUL$
--- a/regression/loop-invariants/ring_loop_with_contract/test.desc
+++ b/regression/loop-invariants/ring_loop_with_contract/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --loop-invariant --replace-call-with-contract source_startup --replace-call-with-contract source_reaction_start --replace-call-with-contract source_reaction_in --replace-call-with-contract node_reaction --replace-call-with-contract process_messages_at_time --no-align-check
 ^VERIFICATION SUCCESSFUL$
--- a/regression/loop-invariants/ring_loop_with_contract/test.desc
+++ b/regression/loop-invariants/ring_loop_with_contract/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 --loop-invariant --replace-call-with-contract source_startup --replace-call-with-contract source_reaction_start --replace-call-with-contract source_reaction_in --replace-call-with-contract node_reaction --replace-call-with-contract process_messages_at_time --no-align-check
 ^VERIFICATION SUCCESSFUL$
--- a/src/esbmc/parseoptions/function_contracts.cpp
+++ b/src/esbmc/parseoptions/function_contracts.cpp
@@ -8,8 +8,8 @@
 #include <set>
 #include <string>
 
-// Process function contracts if enabled
-void esbmc_parseoptionst::process_function_contracts(
+// Process function contracts if enabled. Returns true on a usage error.
+bool esbmc_parseoptionst::process_function_contracts(
   goto_functionst &goto_functions,
   bool has_replace,
   bool has_enforce,
@@ -59,32 +59,51 @@ void esbmc_parseoptionst::process_function_contracts(
       return result;
     };
 
-  // Lambda function to process function list (handles "*" wildcard)
-  auto process_function_list = [&collect_functions_with_contracts](
-                                 const std::list<std::string> &func_list) {
-    std::set<std::string> result;
-    for (const auto &func : func_list)
-    {
-      if (func == "*")
+  // Lambda function to process function list (handles "*" wildcard).
+  // An unresolvable name is a usage error rather than a warning: these options
+  // decide *what* gets verified, so silently skipping one leaves ESBMC
+  // verifying something other than what was asked and still reporting a
+  // verdict for it. Sets ok=false on the first bad name.
+  bool ok = true;
+  auto process_function_list =
+    [&collect_functions_with_contracts, &contracts, &ok](
+      const std::list<std::string> &func_list, const char *opt) {
+      std::set<std::string> result;
+      for (const auto &func : func_list)
       {
-        // "*" means all functions with contracts
-        result = collect_functions_with_contracts();
-        break; // "*" means all, so we can break after collecting
-      }
-      else
-      {
+        if (func == "*")
+        {
+          // "*" means all functions with contracts
+          return collect_functions_with_contracts();
+        }
+
+        if (contracts.find_function_symbol(func) == nullptr)
+        {
+          log_error(
+            "--{}: no function named '{}'{}",
+            opt,
+            func,
+            func.find(',') != std::string::npos
+              ? " (names are not comma-separated; repeat the flag once per "
+                "function, or use \"*\")"
+              : "");
+          ok = false;
+          return std::set<std::string>();
+        }
         result.insert(func);
       }
-    }
-    return result;
-  };
+      return result;
+    };
 
   // Process enforce-contract option
   if (has_enforce)
   {
     const std::list<std::string> &enforce_list =
       cmdline.get_values("enforce-contract");
-    std::set<std::string> to_enforce = process_function_list(enforce_list);
+    std::set<std::string> to_enforce =
+      process_function_list(enforce_list, "enforce-contract");
+    if (!ok)
+      return true;
 
     if (!to_enforce.empty())
     {
@@ -104,14 +123,13 @@ void esbmc_parseoptionst::process_function_contracts(
   {
     const std::list<std::string> &replace_list =
       cmdline.get_values("replace-call-with-contract");
-    std::set<std::string> to_replace = process_function_list(replace_list);
+    std::set<std::string> to_replace =
+      process_function_list(replace_list, "replace-call-with-contract");
+    if (!ok)
+      return true;
 
     if (!to_replace.empty())
-    {
-      log_status(
-        "Replacing calls with contracts for {} function(s)", to_replace.size());
       contracts.replace_calls(to_replace);
-    }
   }
 
   // Lambda to collect ONLY functions with __ESBMC_contract annotation
@@ -157,4 +175,6 @@ void esbmc_parseoptionst::process_function_contracts(
       contracts.replace_calls(to_replace);
     }
   }
+
+  return false;
 }
--- a/src/esbmc/parseoptions/function_contracts.cpp
+++ b/src/esbmc/parseoptions/function_contracts.cpp
@@ -59,14 +59,32 @@ bool esbmc_parseoptionst::process_function_contracts(
       return result;
     };
 
-  // Lambda function to process function list (handles "*" wildcard).
-  // An unresolvable name is a usage error rather than a warning: these options
-  // decide *what* gets verified, so silently skipping one leaves ESBMC
-  // verifying something other than what was asked and still reporting a
-  // verdict for it. Sets ok=false on the first bad name.
+  // Reject a name these options cannot act on, and say which way it failed.
+  // All three ways -- unresolvable, no body, no contract -- reach the same
+  // place: nothing is enforced or replaced, yet a verdict is still printed for
+  // a run that is not the one that was asked for. Returns the reason, or an
+  // empty string when the name is usable.
+  auto unusable_because = [&contracts,
+                           &goto_functions](const std::string &func) {
+    const symbolt *sym = contracts.find_function_symbol(func);
+    if (sym == nullptr)
+      return std::string("no function of that name");
+
+    auto it = goto_functions.function_map.find(sym->id);
+    if (it == goto_functions.function_map.end() || !it->second.body_available)
+      return std::string("that function is declared but has no body here");
+
+    if (
+      !contracts.has_contracts(it->second.body) &&
+      !contracts.is_annotated_contract_function(*sym))
+      return std::string("that function declares no contract clauses");
+
+    return std::string();
+  };
+
   bool ok = true;
   auto process_function_list =
-    [&collect_functions_with_contracts, &contracts, &ok](
+    [&collect_functions_with_contracts, &unusable_because, &ok](
       const std::list<std::string> &func_list, const char *opt) {
       std::set<std::string> result;
       for (const auto &func : func_list)
@@ -77,12 +95,14 @@ bool esbmc_parseoptionst::process_function_contracts(
           return collect_functions_with_contracts();
         }
 
-        if (contracts.find_function_symbol(func) == nullptr)
+        std::string reason = unusable_because(func);
+        if (!reason.empty())
         {
           log_error(
-            "--{}: no function named '{}'{}",
+            "--{}: cannot use '{}': {}{}",
             opt,
             func,
+            reason,
             func.find(',') != std::string::npos
               ? " (names are not comma-separated; repeat the flag once per "
                 "function, or use \"*\")"
--- a/src/esbmc/parseoptions/function_contracts.cpp
+++ b/src/esbmc/parseoptions/function_contracts.cpp
@@ -8,7 +8,7 @@
 #include <set>
 #include <string>
 
-// Process function contracts if enabled. Returns true on a usage error.
+// Process function contracts if enabled
 bool esbmc_parseoptionst::process_function_contracts(
   goto_functionst &goto_functions,
   bool has_replace,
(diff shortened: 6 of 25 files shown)


LukeW1999:
Self-review update. Applying the same standard I applied elsewhere today (build it, probe it, do not trust the reasoning) found that the first commit closed only one of three ways to reach the failure this PR is about.

| name passed to `--replace-call-with-contract` | before this update |
|---|---|
| resolves, has a body, has a contract | replaced, FAILED — correct |
| **resolves to a declaration with no body** | "replaced 1", replaced 0, **SUCCESSFUL** |
| **resolves, has a body, has no contract clauses** | "replaced 1", replaced 0, **SUCCESSFUL** |
| does not resolve | correctly rejected |

The two middle rows are the same silent no-op the issue is about, reached differently, and they print the same misleading "Replacing calls with contracts for 1 function(s)" while replacing none. Having argued in this PR that a check which looks like it ran is the worst state, leaving those in place would have been the same fault.

Validation now reports which of the three ways the name failed, so the message says what to fix:

```
ERROR: --replace-call-with-contract: cannot use 'extern_fn': that function is declared but has no body here
ERROR: --replace-call-with-contract: cannot use 'plain': that function declares no contract clauses
ERROR: --replace-call-with-contract: cannot use 'nosuch': no function of that name
```

`"*"` and the working path are unaffected. Two tests added for the new cases; `function_contract` + `loop-invariants` 368/368.

src/esbmc/parseoptions/function_contracts.cpp
```diff
@@ -59,32 +59,71 @@ void esbmc_parseoptionst::process_function_contracts(
...
-        result = collect_functions_with_contracts();
-        break; // "*" means all, so we can break after collecting
-      }
-      else
+  // Reject a name these options cannot act on, and say which way it failed.
+  // All three ways -- unresolvable, no body, no contract -- reach the same
+  // place: nothing is enforced or replaced, yet a verdict is still printed for
+  // a run that is not the one that was asked for. Returns the reason, or an
+  // empty string when the name is usable.
+  auto unusable_because = [&contracts,
```
lucasccordeiro: This validates with a different predicate than the consumer acts on, so #6498 stays open for C++.

`unusable_because` resolves via `find_function_symbol`, which falls back to matching the symbol's *short* name (`contracts.cpp:143-170`). `replace_calls` selects via `matches_replace_pattern` (`contracts.cpp:3485-3510`), which takes the ID tail after the last `@` and strips **one** trailing `#`. For a C++ free function `int fst(int)` the ID is `c:@F@fst#I#`, so the tail is `fst#I` != `fst`.

Verified on a local build (8.4.0, pre-patch, so validation is not the thing filtering here):

```
$ cat t.cpp
int fst(int x) { __ESBMC_ensures(__ESBMC_return_value >= 0); return x > 0 ? x : 0; }
int main() { int b = fst(3); __ESBMC_assert(b == 3, "true of body, not of contract"); return 0; }

$ esbmc t.cpp --replace-call-with-contract fst
Replacing calls with contracts for 1 function(s)
VERIFICATION SUCCESSFUL          # b == 3 held -> the body ran, nothing was replaced
```

With this patch `fst` passes `unusable_because` (symbol found, body available, `has_contracts` true) and the run still reports replacing 1 function while replacing none. That is the same failure the PR is about, reached a fourth way.

The robust fix is a post-condition rather than a pre-condition: have `replace_calls`/`enforce_contracts` return how many functions they actually acted on and error when a caller-supplied name contributed zero. That closes the C++ case, the `*` case and the all-contracts flags in one place, instead of adding a third copy of the eligibility predicate that can drift from the two that already exist.

src/esbmc/parseoptions/function_contracts.cpp
```diff
@@ -59,32 +59,71 @@ void esbmc_parseoptionst::process_function_contracts(
...
+                           &goto_functions](const std::string &func) {
+    const symbolt *sym = contracts.find_function_symbol(func);
```
lucasccordeiro: `find_function_symbol` returns `nullptr` for two distinct reasons, and one of them already logged an error of its own (`contracts.cpp:155-166`). On an overloaded C++ name the user now gets two contradicting ERROR lines:

```
ERROR: Ambiguous function name 'fst'; use a full symbol ID to disambiguate. Candidates: c:@F@fst#I#, c:@F@fst#d#
ERROR: --enforce-contract: cannot use 'fst': no function of that name
```

(first line reproduced on a local build). Ambiguity needs its own reason string, or `find_function_symbol` needs to report which failure it hit rather than collapsing both into `nullptr`.

src/esbmc/parseoptions/function_contracts.cpp
```diff
@@ -104,14 +143,13 @@ void esbmc_parseoptionst::process_function_contracts(
   {
     const std::list<std::string> &replace_list =
       cmdline.get_values("replace-call-with-contract");
-    std::set<std::string> to_replace = process_function_list(replace_list);
+    std::set<std::string> to_replace =
+      process_function_list(replace_list, "replace-call-with-contract");
+    if (!ok)
+      return true;
```
lucasccordeiro: The sibling flags keep the exact bug this PR fixes. `--replace-call-with-contract "*"` skips `unusable_because` entirely, and `--enforce-all-contracts` / `--replace-all-contracts` (lines 174-192) are untouched, so all three still no-op silently when nothing matches:

```
$ cat nc.c
int plain(int x) { return x + 1; }
int main(void) { return plain(2) - 3; }

$ esbmc nc.c --replace-call-with-contract "*"; echo exit=$?
VERIFICATION SUCCESSFUL
exit=0
```

A user who writes `"*"` because the flag told them to, on a file where the contract macros did not survive preprocessing, gets the same misleading verdict. Checking that the resulting set is non-empty (or that the action did something) covers all five entry points; validating only the two explicit-name paths fixes three of five.

src/esbmc/parseoptions/function_contracts.cpp
```diff
@@ -59,32 +59,71 @@ void esbmc_parseoptionst::process_function_contracts(
...
+    };
 
   // Process enforce-contract option
   if (has_enforce)
   {
     const std::list<std::string> &enforce_list =
       cmdline.get_values("enforce-contract");
-    std::set<std::string> to_enforce = process_function_list(enforce_list);
+    std::set<std::string> to_enforce =
+      process_function_list(enforce_list, "enforce-contract");
```
lucasccordeiro: Both lists should be validated before either action runs. As ordered, a bad `--replace-call-with-contract` name aborts only after `enforce_contracts` has already rewritten `goto_functions` and printed `Enforcing contracts for N function(s)` — so `--enforce-contract f --replace-call-with-contract typo` emits a status line for work that is then thrown away. Hoisting both `process_function_list` calls above the two `if` bodies keeps the failure side-effect-free and makes the `ok` checks collapse into one.

src/esbmc/parseoptions/function_contracts.cpp
```diff
@@ -59,32 +59,71 @@ void esbmc_parseoptionst::process_function_contracts(
...
+
+    if (
+      !contracts.has_contracts(it->second.body) &&
+      !contracts.is_annotated_contract_function(*sym))
+      return std::string("that function declares no contract clauses");
+
+    return std::string();
+  };
+
+  bool ok = true;
```
lucasccordeiro: `ok` is a by-reference out-parameter on a lambda that also has a return value, is never reset between the two calls, and is checked by a `if (!ok) return true;` that must be remembered at every future call site. `std::optional<std::set<std::string>>` carries the same information in the return value and makes the failure impossible to skip.

src/goto-programs/contracts/contracts.h
```diff
@@ -159,6 +159,11 @@ class code_contractst
   /// \return True if the function should be skipped (destructor, __cxa_*, etc.)
   bool is_compiler_generated(const std::string &function_name) const;
 
+  /// \brief Find function symbol
+  /// \param function_name Function name (can be full ID or simple name)
+  /// \return Pointer to function symbol, or nullptr if not found
+  symbolt *find_function_symbol(const std::string &function_name);
```
lucasccordeiro: Widening this to public hands callers a mutable `symbolt *` for what is a read-only lookup. More to the point, the eligibility test now exists in three places that must stay in agreement: `enforce_contracts` (`contracts.cpp:749-776`), `replace_calls` (`contracts.cpp:3561-3564`) and the new `unusable_because`. Better to put the predicate where the knowledge lives — e.g. a `std::string diagnose_contract_target(const std::string &) const` on `code_contractst` returning the reason or empty — and leave `find_function_symbol` private. Consider `const symbolt *` either way.

src/esbmc/parseoptions/process_goto_program.cpp
```diff
@@ -439,13 +439,15 @@ bool esbmc_parseoptionst::process_goto_program(
     bool has_replace = cmdline.isset("replace-call-with-contract");
     bool has_enforce_all = cmdline.isset("enforce-all-contracts");
     bool has_replace_all = cmdline.isset("replace-all-contracts");
-    if (has_enforce || has_replace || has_enforce_all || has_replace_all)
+    if (
```
lucasccordeiro: Folding the guard and the call into one `&&` makes a six-line call expression read as a condition; the `return true` at the end is easy to misattribute to the `isset` checks. A nested `if` is clearer here and keeps the diff to the inner statement.

regression/function_contract/contract_fn_name_unknown/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract no_such_function
+^ERROR: --replace-call-with-contract: cannot use 'no_such_function': no function of that name$
```
lucasccordeiro: These tests do not pin the thing the PR is fixing. The runner only supports positive regexes and does not consult the exit status (`regression/README.md:78`), so this passes as long as the ERROR line appears anywhere — including in a run that prints the error and then goes on to `VERIFICATION SUCCESSFUL` with exit 0, which is precisely the pre-patch behaviour you are removing.

Add a negative-lookahead regex so "no verdict is produced" is actually asserted, e.g.

```
\A(?!(.|\n)*VERIFICATION)
```

Same applies to `contract_fn_name_comma`, `contract_fn_name_enforce_unknown`, `contract_fn_name_no_body` and `contract_fn_name_no_contract`.

regression/function_contract/contract_fn_name_repeated/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract f --replace-call-with-contract g
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: This guard is weaker than the PR describes. `VERIFICATION FAILED` is reached if *either* name takes effect: replacing only `g` also makes `b == 3` fail. So it does not distinguish "both replaced" from "only the second replaced". Add

```
^Replacing calls with contracts for 2 function\(s\)$
```

which is the line `replace_calls` already emits and is the property being guarded.

regression/function_contract/ring_enforce_process_messages/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro: `KNOWNBUG` is in `FAIL_MODES` (`regression/testing_tool.py:48`) and is not part of the default CORE run, so this stops executing in CI. #6499 is already pinned by `replace_ternary_ensures_knownbug`, so demoting this one buys no new coverage of the bug and costs the only integration-scale coverage of multi-function replacement on the ring example.

Suggest keeping a `CORE` variant with the replacement list narrowed to the callees that do not hit #6499, alongside the `KNOWNBUG` full-list version. Also worth renaming the demoted directory to the local `*_knownbug` convention (`arr_assigns_large_index_knownbug`, `replace_ternary_ensures_knownbug`, ...) — as it stands nothing in the tree connects either test to #6499.

regression/loop-invariants/ring_loop_with_contract/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
lucasccordeiro: Same concern as `ring_enforce_process_messages`: this is the only test exercising five-way replacement under `--loop-invariant`, and moving it to `KNOWNBUG` takes it out of the CORE run. Please keep a reduced `CORE` variant.

regression/function_contract/contract_fn_name_no_body/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--replace-call-with-contract extern_fn
+^ERROR: --replace-call-with-contract: cannot use 'extern_fn': that function is declared but has no body here$
```
lucasccordeiro: Rejecting a declared-but-undefined function is correct for ESBMC today (contracts are extracted from the body and `replace_calls` skips `!body_available`), but this test freezes it as the intended contract of the flag. Abstracting a call to a function whose body is not in the translation unit is the natural use for `--replace-call-with-contract`; if that is on the roadmap, this test will have to be inverted. Worth a note in #6498 so the decision is recorded rather than only encoded in a regex.

lucasccordeiro (commented):
Review of the validation change. The direction is right, but as written the new check does not close #6498 for C++ (verified locally, see the inline comment on `unusable_because`), and it leaves the `*` / `--enforce-all-contracts` / `--replace-all-contracts` paths with exactly the same hole. Two further points that have no diff line to hang off:

- `src/esbmc/options.cpp:391-400` still describes both flags as taking a `fun` and says nothing about repetition. The error message now teaches the repeated-flag form; `--help` should too, otherwise the only place the rule is written down is a failure path.
- `*` changed from `break` to an early `return`, so `--replace-call-with-contract typo --replace-call-with-contract "*"` used to run over every contract function and now errors out. That looks intended, but it is a behaviour change that is not in the PR description.

LukeW1999:
Reworked and force-pushed (`b441afc`). Thank you for the review — two of your points turned out to be load-bearing in ways I did not see until I had built the thing you asked for and tried to break it.

**Where it landed.** Your first comment asked for a post-condition: have the passes report what they acted on, and error on a name that contributed zero. I built exactly that first. It closes the C++ case, `"*"` and the all-contracts flags as you predicted, but testing it found that it also breaks legitimate command lines, because `enforce_contracts` rewrites what it acts on into a wrapper carrying no contract. Measured on the post-condition build:

```
$ esbmc basic.c --function inc --enforce-contract inc --replace-call-with-contract "*"
Enforced contract for function inc
ERROR: --replace-call-with-contract: no function with a contract to act on
```

That is the shape `enforce_tlv_mid_pass`, `enforce_tlv_parser_hierarchy_pass`, `enforce_election_process_message_pass` and `enforce_ring_process_messages_pass` all use; they survive only because their files contain other contracted functions. `--enforce-contract inc --enforce-contract c:@F@inc`, two valid spellings of one function, failed the same way.

So your fourth comment — validate both lists before either action runs — was not a sequencing preference, it is what makes the check sound at all, and the post-condition design forecloses it. The version I pushed is your first and sixth comments combined: a single `code_contractst::diagnose_contract_target`, asked once per flag before any pass runs. It is asked **per option** because the two rules genuinely differ, which is the substance of your first comment: `find_function_symbol` matches short names, `matches_replace_pattern` takes the id tail and strips one `#`, so `c:@F@fst#I#` gives `fst#I` and never equals `fst`.

**Per comment**, all at `b441afc`:

| your comment | what changed |
|---|---|
| C++ predicate drift | `diagnose_contract_target(name, for_replace)`; `contract_fn_name_cpp_paramspp` pins `int fst(int)` |
| two contradicting ERRORs on ambiguity | the cause is reported by the code that knows it; the caller's line no longer restates a different one |
| `"*"` and the all-contracts flags | all five entry points checked; `_star_no_contract` and `_all_no_annotation` added |
| validate both before either acts | done, and it is now the reason the design changed |
| `ok` out-parameter | gone |
| `find_function_symbol` public / third copy | reverted to private; the rule lives on `code_contractst`, as you suggested |
| folded guard | back to a nested `if` |
| tests do not pin the fix | all six error tests carry `\A(?!(.|\n)*VERIFICATION)` |
| `contract_fn_name_repeated` guard | added, and `replace_calls` now reports the count it acted on rather than the count requested — the line asserted nothing before, since it printed the requested count whether or not anything was replaced |
| both ring `test.desc` | dropped; net change to those files is zero after rebasing on master |

`ctest -L function_contract` 314/314, `ctest -L loop-invariants` 73/73.

**Three things I would like your call on.**

1. `__ESBMC_assigns()` as a function's only contract is now rejected with a reason, where before it was silently ignored. `has_contracts` does not recognise `contract::assigns_empty` even though `contracts.cpp:700` handles that comment, so an empty frame condition has never worked. I left the behaviour alone and will file it separately rather than widen this PR — say if you would rather it went in here.
2. A misspelling next to `"*"` is not diagnosed, in either order. The wildcard resolves to full symbol ids while users spell short names, so they cannot be compared without rejecting every legitimate spelling. Documented in the code; happy to make it an error if you prefer, at the cost of requiring full ids alongside `"*"`.
3. Your last comment asked that the reject-declared-but-undefined decision be recorded in #6498 rather than only in a regex. I will add that note; `contract_fn_name_no_body` currently also never calls `extern_fn`, so it conflates "no body" with "never called" — I will add the call so it pins the intended cause.

Happy to reply per thread instead if that is easier to resolve against.

PR #6506, opened 2026-07-29
[pointer-analysis] Add the same-object guard to byte-array bounds

### LINKED ISSUE #6464: [cpp] placement-new reports a spurious heap out-of-bounds under --incremental-bmc / --k-induction

lucasccordeiro:
`construct_struct_ref_from_dyn_offset` seeded its recursion with a true guard and the byte-array arm built a fresh empty one, so the bounds check lost both the path guard and the same-object guard that `build_reference_to` had already put on the sibling check at `dereference.cpp:810`; each value-set target's claim then had to hold for every dynamic object at once, and a second `malloc` on a never-taken branch was enough to report a spurious heap out-of-bounds.
This is the real cause of #6464 and supersedes the constant-propagation workaround in #6470 — that PR's own C++ placement-new test passes here without any propagation change.
Verified across plain BMC / `--incremental-bmc` / `--k-induction`, each with and without `--no-propagation` (18/18), plus six negative programs agreeing with ASan; every regression suite matches a master build on the same machine.
Stacked on #6505, which the propagation-independence test needs. Fixes #6464.

### DIFF
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -55,7 +55,11 @@ execution_statet::execution_statet(
   const goto_programt *goto_program = &(it->second.body);
 
   // Initialize initial thread state
-  goto_symex_statet state(*state_level2, global_value_set, ns);
+  goto_symex_statet state(
+    *state_level2,
+    global_value_set,
+    ns,
+    options.get_bool_option("no-propagation"));
   state.initialize(
     (*goto_program).instructions.begin(),
     (*goto_program).instructions.end(),
@@ -665,7 +669,11 @@ void execution_statet::execute_guard()
 
 unsigned int execution_statet::add_thread(const goto_programt *prog)
 {
-  goto_symex_statet new_state(*state_level2, global_value_set, ns);
+  goto_symex_statet new_state(
+    *state_level2,
+    global_value_set,
+    ns,
+    options.get_bool_option("no-propagation"));
   new_state.initialize(
     prog->instructions.begin(),
     prog->instructions.end(),
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -43,6 +43,7 @@ execution_statet::execution_statet(
   symex_trace = options.get_bool_option("symex-trace");
   smt_during_symex = options.get_bool_option("smt-during-symex");
   smt_thread_guard = options.get_bool_option("smt-thread-guard");
+  no_propagation = options.get_bool_option("no-propagation");
 
   goto_functionst::function_mapt::const_iterator it =
     goto_functions.function_map.find("__ESBMC_main");
@@ -55,11 +56,7 @@ execution_statet::execution_statet(
   const goto_programt *goto_program = &(it->second.body);
 
   // Initialize initial thread state
-  goto_symex_statet state(
-    *state_level2,
-    global_value_set,
-    ns,
-    options.get_bool_option("no-propagation"));
+  goto_symex_statet state(*state_level2, global_value_set, ns, no_propagation);
   state.initialize(
     (*goto_program).instructions.begin(),
     (*goto_program).instructions.end(),
@@ -158,6 +155,7 @@ void execution_statet::copy_derived_from(const execution_statet &ex)
   symex_trace = ex.symex_trace;
   smt_during_symex = ex.smt_during_symex;
   smt_thread_guard = ex.smt_thread_guard;
+  no_propagation = ex.no_propagation;
 
   CS_number = ex.CS_number;
 
@@ -670,10 +668,7 @@ void execution_statet::execute_guard()
 unsigned int execution_statet::add_thread(const goto_programt *prog)
 {
   goto_symex_statet new_state(
-    *state_level2,
-    global_value_set,
-    ns,
-    options.get_bool_option("no-propagation"));
+    *state_level2, global_value_set, ns, no_propagation);
   new_state.initialize(
     prog->instructions.begin(),
     prog->instructions.end(),
--- a/regression/esbmc-cpp/cpp/github_6464_placement_new_incremental/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6464_placement_new_incremental/main.cpp
@@ -1,0 +1,55 @@
+// github #6464: placement-new into malloc'd storage was reported as an
+// out-of-bounds heap write under --incremental-bmc and --k-induction, but only
+// when a second malloc sits on a guarded path. Clean under clang++ -std=c++17
+// -fsanitize=address,undefined.
+#include <new>
+#include <cstdlib>
+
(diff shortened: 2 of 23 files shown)


XLiZHI (approved):
LGTM

src/goto-symex/execution_state.cpp
```diff
@@ -665,7 +669,11 @@ void execution_statet::execute_guard()
 
 unsigned int execution_statet::add_thread(const goto_programt *prog)
 {
-  goto_symex_statet new_state(*state_level2, global_value_set, ns);
+  goto_symex_statet new_state(
+    *state_level2,
+    global_value_set,
+    ns,
+    options.get_bool_option("no-propagation"));
```
XLiZHI: Is there a better way to initialize no-propagation? The current method has no effect on single-threading, but in multi-threading, it seems that each thread has to obtain the options.

```
// current
goto_symex_statet new_state(
    *state_level2,
    global_value_set,
    ns,
    options.get_bool_option("no-propagation")); // re-queries options


goto_symex_statet new_state(
    *state_level2,
    global_value_set,
    ns,
    cur_state->no_propagation); // inherit the cached value
```

lucasccordeiro: Good catch, thanks — fixed in 0433aafedd.

I went one step further than inheriting from `cur_state`, because `execution_statet` already caches exactly this kind of flag three lines above where the option was being read:

```cpp
symex_trace      = options.get_bool_option("symex-trace");
smt_during_symex = options.get_bool_option("smt-during-symex");
smt_thread_guard = options.get_bool_option("smt-thread-guard");
no_propagation   = options.get_bool_option("no-propagation");   // new
```

Both construction sites — the initial thread and `add_thread` — now pass that member, so the option is looked up once per run rather than once per thread creation, and it is copied in `copy_derived_from` alongside its siblings so cloned execution states keep it.

`cur_state->no_propagation` would also have worked (`add_thread` is only reached from `symex_thread_create` during symex, so `cur_state` is live there), but it only covers the `add_thread` site and makes new threads depend on whichever thread happens to be current. The cached member covers both sites uniformly and matches the file's existing idiom.

One thing that fell out of testing your suggestion, which is worth recording: **`--no-propagation` cannot be used on concurrent programs at all.** `symex_set_thread_internal_data` (`builtin_functions/threads.cpp:84`) requires a constant thread id and `abort()`s otherwise, and that constant only survives while propagation is on:

```
$ esbmc thr.c --no-propagation
ERROR: __ESBMC_set_thread_internal_data received nonconstant thread id
```

That is pre-existing behaviour of the pthread model, not something this PR changes — it simply was not reachable before, since the option had been inert since 148700569a. I have noted it on the option's help text rather than trying to fix the model here. It does not affect the flag's use as a correctness oracle on sequential code, which is what #6506 needs it for.

Concurrency suites re-run clean after the change: `k-induction-parallel` 74/74 and the 29 `pthread_create` tests in `regression/esbmc` all pass.

Issue #5868, opened 2026-07-05
Using ESBMC to verify ESBMC: C++ operational-model gaps block self-verification (frontend is crash-robust)


pmatos:
# Using ESBMC to verify ESBMC: C++ operational-model gaps block self-verification (frontend is crash-robust)

## TL;DR

I ran the pre-installed **`esbmc` 8.3.0** over **all 301 of ESBMC's own `src/` C++ translation units** to answer two questions:

- **(a)** What gaps in ESBMC's C++ support block it from verifying itself?
- **(b)** Are there ESBMC bugs (crashes / hangs / unsoundness) surfaced along the way?

**Results:**
- **(a)** ESBMC's C++ frontend forces `-nostdinc++` and resolves every `#include <...>` **exclusively** against its own operational-model header tree in `src/cpp/library/`. That tree is missing / incomplete in several places, so **every gap is a hard frontend parse error** — Tier-1 (as-is) blocked **all 301 TUs at parse; 0 reached symbolic execution**. Nine gaps are root-caused below (8 actionable, plus expected external-dep limits), ranked by TUs blocked. The single biggest blocker is that **no concurrency STL headers are modeled at all** (`atomic`/`mutex`/`thread`/…), which alone blocks ~228 TUs.
- **(b)** **No ESBMC bug was found.** Across 602 frontend runs plus full BMC on real ESBMC code (`big-int`) and `--function` symex sweeps, ESBMC produced **zero crashes, aborts, hangs, uncaught exceptions, internal errors, or unsoundness**. This is a positive robustness result.

> This is a single umbrella issue capturing the full study; the individual actionable gaps in §(a) are intended to be split into separate issues later (suggested split at the bottom).

---

## Methodology (reproducible)

- **Target selection & flags.** Per-TU compile flags were derived from an existing build's `compile_commands.json` (kept `-D`/`-I`/`-isystem`→`-I`/`-std=`→`--std`/`-include`→`--include-file`; dropped `-o -c -O -W -f`). Each TU was then run as `esbmc <flags> <file.cpp> --goto-functions-only`, which exercises the full **clang parse → AST→irep2 conversion → GOTO conversion** pipeline (no solver needed to surface frontend gaps and crashes).
[...]

PR #6517, opened 2026-07-29
[clang-frontend] Serve bundled headers to clang from memory instead of /tmp

### LINKED ISSUE #6509: [feat] Implement Virtual IO Interface
ESBMC internals use a temporary directory for I/O operations. The current filesystem abstraction is `src/util/base/filesystem.h`. It should be expanded. There should be a new default mode of operation where each file is not extract to a temporary directory and instead is fetched directly in program RAM.

This can be done by encoding each file into a tree struct and making the current I/O interface work over that. This is a big engineering effort but would cut down time/resource usage.

Yiannis128:
Every ESBMC run extracted its bundled clang headers, C++ operational models and internal libc into a temporary directory, purely so clang's FileManager could find them on disk. They are now registered with `file_operations::filesystemt` and served to clang through an `llvm::vfs::InMemoryFileSystem` overlaid on the real one, so nothing is written to disk: 320 files / 7.1 MB of `/tmp` traffic per C run and 488 files / 8.0 MB per C++ run become zero, with wall clock unchanged. Part of #6509.

Two supporting changes come along. flail now appends a NUL sentinel past each generated array, because clang's Lexer requires null-terminated buffers, and excludes it from `_size` so no existing consumer changes. And `esbmc_action` is deleted: the intrinsics go in through `PreprocessorOptions` as a remapped buffer placed first in the forced-include list, which preserves the ordering constraint from #5868 and lets the ASTUnit be built via `LoadFromCompilerInvocation`, the only overload that accepts our FileManager.

Python and Solidity keep extracting to `/tmp`, since a forked `python3` or `solc` has its own address space and can only read real files. No new tests: the change is infrastructural and every existing test exercises it.


scripts/flail.py
```diff
@@ -156,8 +162,12 @@ def _step_6(self, content, output, header, macro : str):
         name = self.obtain_var_name()
         output.write('const char %s[] = {\n' % name)
         output.writelines(content)
-        output.write('};\n')
-        output.write('const unsigned int %s_size = sizeof(%s);\n' % (name, name))
+        # The trailing NUL is a sentinel past the end, excluded from _size:
+        # clang's Lexer requires null-terminated buffers, and bundled files are
+        # handed to it straight from .rodata rather than read back from disk.
```
rafaelsamenezes: Repeating same information from the function doc

src/util/base/filesystem.h
```diff
@@ -128,4 +186,72 @@ void cleanup_registered_tmps();
 void register_pgroup_for_cleanup(long pgid);
 void unregister_pgroup(long pgid);
 void kill_registered_pgroups();
+
+/**
+ * @brief ESBMC's filesystem: files bundled into the binary, overlaid on the
+ *        real one.
+ *
+ * Every file scripts/flail.py bundles is registered here at startup under a
+ * path below ESBMC_VFS_ROOT, costing nothing but a map entry -- the bytes stay
```
rafaelsamenezes: Useless comment

src/util/base/filesystem.h
```diff
@@ -128,4 +186,72 @@ void cleanup_registered_tmps();
...
+ * in .rodata. Lookups resolve by path: the registry first, the real filesystem
+ * on a miss. Callers therefore need not know, and cannot tell, which of the
+ * two answered.
+ *
+ * Bundled files still have to be written out for anything ESBMC cannot reach
+ * into: a forked python3 or solc has its own address space and can only read
+ * real files. materialize() serves those.
+ *
+ * Registration is unsynchronised and must complete before any concurrent
+ * read; calling the modules' register_bundled() from main() satisfies that.
```
rafaelsamenezes: Decide either by keeping register_ prefix as unsyc or not. Adding it to also a sync of registers is confusing.

src/clang-c-frontend/AST/build_ast.cpp
```diff
@@ -154,25 +149,43 @@ std::unique_ptr<clang::ASTUnit> buildASTs(
...
+  /* Inject ESBMC's intrinsics as a forced include ahead of any the user asked
+   * for. They must land after the builtin #defines they rely on but before the
+   * first forced include, which can transitively reach ESBMC's own models and
+   * would otherwise see nondet_* / __ESBMC_* undeclared (github #5868).
+   * Forced includes are emitted into the predefines buffer in order, after all
+   * -D macros, so being first satisfies both constraints.
+   *
+   * This is what lets the ASTUnit be built via LoadFromCompilerInvocation,
+   * which -- unlike the ...Action() overload -- accepts our FileManager and so
+   * reads bundled headers through esbmc_clang_vfs(). */
```
rafaelsamenezes: A bunch of blah-blah-blah

rafaelsamenezes (changes requested):
Just asking for some trimming and better naming

src/util/base/filesystem.h
Yiannis128: Would adding clarifying comments in each register function fix that?

src/util/base/filesystem.cpp
```diff
@@ -50,6 +50,106 @@ void file_operations::kill_registered_pgroups()
...
+{
+  auto it = _bundled.find(path);
+  if (it != _bundled.end())
+    return file_data::bundled(it->second.data(), it->second.size());
+
+  std::ifstream in(path, std::ios::binary | std::ios::ate);
+  if (!in)
+    return {};
+
+  std::string contents(static_cast<size_t>(in.tellg()), '\0');
```
lucasccordeiro: `in.tellg()` returns -1 for non-seekable files, and `static_cast<size_t>(-1)` then makes this allocation throw. Check for a negative `tellg()` and return `{}`.

src/util/base/filesystem.h
```diff
@@ -128,4 +199,71 @@ void cleanup_registered_tmps();
...
+ */
+class filesystemt
+{
+  std::map<std::string, std::string_view> _bundled;
+  std::map<std::string, tmp_path> _materialized;
+
+  template <typename F>
+  void for_each_under(const std::string &prefix, F &&f) const
+  {
+    for (auto it = _bundled.lower_bound(prefix);
```
lucasccordeiro, line 233: Raw prefix match has no path-separator boundary: `list("/esbmc-vfs/libc")` would also match a future `/esbmc-vfs/libcxx` sibling. Require the prefix to be followed by `/` or end-of-string.

src/util/base/filesystem.h
```diff
@@ -128,4 +199,71 @@ void cleanup_registered_tmps();
...
+         it->first.compare(0, prefix.size(), prefix) == 0;
+         ++it)
+      f(it->first, it->second);
+  }
+
+public:
+  static filesystemt &get();
+
+  /** @brief Registers static file contents; see file_data::bundled(). */
+  void add_bundled(const std::string &path, const char *data, size_t size);
```
lucasccordeiro, line 247: `add_bundled()` silently overwrites an existing path without changing `bundled_count()`, so the count-based cache in `esbmc_clang_vfs()` would go stale. Assert on duplicate registration.

src/c2goto/cprover_libc_sources.cpp
```diff
@@ -111,27 +65,30 @@ static std::string get_filename_from_path(const std::string &path)
...
-      {
-        log_error("error processing internal libc source {}", path);
-        abort();
-      }
-      delete l;
-    });
+  // Parsed straight out of .rodata: clang resolves these through
+  // esbmc_clang_vfs(), so no libc source is ever written to disk.
+  for (const std::string &path :
+       file_operations::filesystemt::get().list(vfs_library))
```
lucasccordeiro: `list()` iterates lexicographically, so libm sources are now parsed interleaved with libc rather than after it. Confirm symbol merging is order-insensitive, or preserve the original order.

scripts/flail.py
```diff
@@ -13,110 +14,114 @@
...
+        72,69,65,68,69,82,83,95,83,84,68,65,82,71,95,72,
+        95,13,10,35,100,101,102,105,110,101,32,95,95,69,83,66,
+
+    Step 6: Create a C header containing the output of Step 5 as a char array
+    """
 
-    REGEX_REMOVE_ADDR = re.compile(r'^[0-9]+( +)?')
-    REGEX_MULTI_SPACE = re.compile(r' +')
-    REGEX_ADD_COMMA_END = re.compile(r'.*[0-9]$')
+    REGEX_REMOVE_ADDR = re.compile(r"^[0-9]+( +)?")
```
lucasccordeiro: This file was Black-reformatted (requoting, trailing commas), but the repo mandates YAPF (`.style.yapf`) and `yapf --diff` now fails on it. Drop the cosmetic reformat or re-run YAPF.

CLAUDE.md
```diff
@@ -64,7 +64,7 @@ ctest -R "regression/esbmc/00_big_endian_01" --output-on-failure
 
 **Important: the Python frontend needs `python3` on `PATH`.** ESBMC's Python frontend invokes `python3` to run `parser.py`. `ast2json` is vendored in the source tree (`src/python-frontend/libs/ast2json`), so it no longer needs to be installed separately for Python regression tests. (`mypy` is an optional extra for type checking.)
 
-**Important: /tmp disk space.** Each regression test creates an `esbmc-headers-*` temp directory (~7.4MB) in `/tmp`. Running the full suite generates thousands of these (~70GB total). Clean them after test runs: `rm -rf /tmp/esbmc-headers-*`
+**Note: /tmp disk space.** C and C++ runs write nothing to `/tmp`: bundled clang headers, the C++ operational models and the internal libc are registered with `file_operations::filesystemt` and served to clang out of `.rodata` via `esbmc_clang_vfs()` (`src/clang-c-frontend/AST/vfs_adapter.h`). The Python and Solidity frontends extract to `/tmp`, because they fork `python3`/`solc` and a separate process cannot read ESBMC's memory. Clean up after large runs of those suites: `rm -rf /tmp/esbmc*`
```
lucasccordeiro, line 67: AGENTS.md line 67 still has the old `esbmc-headers-*` /tmp paragraph; this repo requires CLAUDE.md and AGENTS.md to be updated together. Apply the same edit to AGENTS.md.

scripts/flail.py
Yiannis128: Dropped. The class docstring already covers it.

src/util/base/filesystem.h
Yiannis128: Trimmed to four lines: what is registered, that read() falls back to disk, and what materialize() is for.

src/util/base/filesystem.h
Yiannis128: Done in 6627dee: the paragraph is gone and each register function says who reads it and when.

src/clang-c-frontend/AST/build_ast.cpp
Yiannis128: Cut to two lines, just the #5868 ordering constraint.

src/util/base/filesystem.cpp
Yiannis128: Fixed in 2ae97ec: `read()` bails out when `tellg()` is negative.

src/util/base/filesystem.h
Yiannis128: Fixed in 5b42193: `for_each_under()` now requires the prefix to end at a `/` or at the end of the path. Covered by the new "list matches whole path components" unit test, which also registers a `lib-old/` sibling since `-` sorts below `/` and so is scanned first.

src/util/base/filesystem.h
Yiannis128: Fixed in 2ae97ec: `add_bundled()` asserts the path is not already registered.

src/c2goto/cprover_libc_sources.cpp
Yiannis128: Good catch. Rather than reason about order sensitivity I restored the original order in 4bc661f: `add_bundled_library_sources()` is driven by the generated `libc.h`/`libm.h` headers again, not by `list()`, so libc is parsed before libm and in the same order as before.

scripts/flail.py
Yiannis128: Reverted in 4217523. Only the NUL sentinel change remains, and `yapf --diff` reports nothing on `scripts/flail.py` that it did not already report on master.

CLAUDE.md
Yiannis128: AGENTS.md is a symlink to CLAUDE.md in this repo, so it already has the edit.

Yiannis128:
So had to fix some more tests that were checking for the existence of OM in `/tmp`. Now there's an in memory check function for the existence of a file: `file_operations::is_bundled_source()` instead. _Should_ pass now OK.

src/util/base/filesystem.cpp
```diff
@@ -92,6 +93,122 @@ void file_operations::kill_registered_pgroups()
...
+}
+
+void filesystemt::add_bundled(
+  const std::string &path,
+  const char *data,
+  size_t size)
+{
+  /* bundled_count() keys the overlay cache in esbmc_clang_vfs(), so silently
+   * replacing an entry would leave that cache stale. */
+  assert(!_bundled.count(path));
```
lucasccordeiro: `assert` is compiled out in release, so a duplicate registration silently replaces the entry without changing `bundled_count()` — the stale-cache case this comment warns about. Make it a hard failure.

src/util/base/filesystem.cpp
```diff
@@ -92,6 +93,122 @@ void file_operations::kill_registered_pgroups()
...
+{
+  return _bundled;
+}
+
+bool file_operations::is_bundled_source(std::string_view file)
+{
+  // Matched without its leading separator: on Windows the root is spelled
+  // "C:/esbmc-vfs" and clang may hand the rest of the path back native.
+  std::string_view root = std::string_view(ESBMC_VFS_ROOT).substr(1);
+  return file.find(root) != std::string_view::npos ||
```
lucasccordeiro: Unanchored substring match: a user path containing `esbmc-vfs` or `c2goto/library/` is misclassified as bundled. Match the prefix instead.

CLAUDE.md
lucasccordeiro, line 67: No test backs the "C and C++ runs write nothing to /tmp" claim. Add a regression test asserting no temp directory appears for a C run.

lucasccordeiro (changes requested):

src/util/base/filesystem.cpp
Yiannis128: Fixed in e3f23096ea. `emplace()` plus an explicit `fprintf`/`abort()`, so it fires under NDEBUG too. Confirmed the build line carries `-DNDEBUG` under RelWithDebInfo, so the check really was inert exactly where it mattered. Used `fprintf` rather than `log_error` because the `filesystem` CMake target links Boost only and pulling in `util/message` fails to compile for want of fmt; `tmp_file::~tmp_file` in the same file already uses `fprintf`. Since this now aborts where it previously did nothing, I checked every generated flail list for collisions (`cheaders.h` 226 entries, `libc.h` 21, `libm.h` 36, plus cross-prefix between the three libc roots) and there are none.

src/util/base/filesystem.cpp
Yiannis128: Fixed in 7ba7564679. The VFS root is now anchored at position 0 after skipping a Windows drive letter, and matching is component-wise so `/home/u/esbmc-vfs/main.c` and `/esbmc-vfs-notes/main.c` are user code again. The `c2goto/library` spelling cannot be prefix-anchored: that code is compiled into the goto binary at build time, so its symbols carry whatever absolute path the build tree had (`/Users/runner/work/esbmc/esbmc/src/c2goto/library/string.c` in CI). I tightened it to the component sequence `src/c2goto/library`, required to start and end on a boundary. I did consider baking the source dir in from CMake for a true prefix, but if a build ever passed c2goto relative paths the match would silently fail and resurrect the bug cd9c7d9cfc fixed, with no test able to catch it. Four previously-misclassified paths added to the unit test.

CLAUDE.md
Yiannis128: Added in be5383669a, though not as a directory-emptiness check: that would have been vacuous. `tmp_path`'s destructor removes what it created, so a run that extracts and cleans up is indistinguishable from one that never extracted. I confirmed that empirically with inotify on a run pointed at a fresh `TMPDIR`: the existing `gcc_nested_func_01` test creates `esbmc-nested.d683-f2c3.c` there, yet the directory is empty afterwards and its mtime is unchanged.

So the tests pin the paths instead, which is where extraction would actually show up:

- `regression/esbmc/bundled_headers_from_vfs` asserts clang is invoked with `-isystem /esbmc-vfs/libc/headers` and `-resource-dir /esbmc-vfs/clang` (VERIFICATION SUCCESSFUL).
- `regression/esbmc-cpp/cpp/om_source_from_vfs` asserts an OM source location in the counterexample reads `/esbmc-vfs/cpp/vector` (VERIFICATION FAILED).

Reintroducing extraction turns those into a temp directory and both fail. Non-vacuity checked by running the first with `--no-library`, which drops the `-isystem` line and breaks the match. Both regexes tolerate a `C:` prefix since the Windows leg runs the `esbmc/` label. The note in CLAUDE.md now points at both tests and records why the emptiness check does not work.

rafaelsamenezes (approved):

lucasccordeiro (approved):

rafaelsamenezes (approved):

PR #6519, opened 2026-07-29
Python type inference module

mihaistate:
Created a Pyton inference module that uses a lattice domain and a CFG to go over the AST nodes and infer types by using specific python operations and methods in the dataflow solver, which aim to compute the final type of every statement of a function, and, after a fixpoint iteration of the basic blocks of the CFG, return the final type of a function that is then inferred to a node in the module_lifecycle_mixin file of the preprocessor


lucasccordeiro (commented):
Review from a maintainer's perspective. I built this branch and ran it against the existing suite; findings below are reproduced, not inferred.

**Blocking:**

1. `infer_type_from_expr` returns `None` for any method call (`dataflow_solver.py:320`), which crashes `parser.py` with `AttributeError: 'NoneType' object has no attribute 'join'`. **31 of the first 420 `regression/python` tests fail on this branch**, all with that traceback. ESBMC returns no verdict at all on affected inputs.
2. Debug `print()`s are left in the default execution path — `module_lifecycle_mixin.py` unparses the whole module to stdout twice per module, plus per-block AST dumps in `dataflow_solver.py` and `annotate_ast.py`.
3. The new pass runs unconditionally with no opt-out flag and no exception guard, so any internal error takes the frontend down with it.
4. `regression/python/flow-sensitive-narrowing` is marked `CORE`/`VERIFICATION SUCCESSFUL` but **fails** (`assertion 0` at `main.py:23`). `regression/python/flow_sensitive_narrowing` passes vacuously — `Generated 0 VCC(s)`, the function is never called.
5. `pylint` reports 10 × `E0602 undefined-variable` in `resolve_method_call` (parameter is `onj_type`, body reads `obj_type`); CI lint fails.

**The feature does not currently fire.** Two independent bugs make narrowing a no-op, which I confirmed by instrumenting the fixpoint: `analyze_function` overwrites `in_envs[i]` from `out_envs[preds]` (line 64), discarding the narrowed environments written at lines 164-179; and `Type.narrow_with_isinstance` returns `Unknown()`, so `isinstance(x, T)` on a parameter (which starts as `Unknown`) refines nothing. On top of that, `annotate_ast` only rewrites top-level statements of a function body, so assignments inside `if`/`while` — the flow-sensitive cases — are never annotated.

**Soundness.** Emitted annotations drive ESBMC's Python type assignment, so a wrong annotation changes the verdict rather than just the precision. Three concrete issues: parameters are annotated from the *out* environment (post-reassignment types, not what callers may pass); annotations are emitted as `ast.Name` with composite ids like `Union[int,str]` / `Callable[int->str]`, which are not well-formed AST; and the solver silently returns a non-fixpoint after `max_iters` (the `widen()` methods implemented across `lattice.py` are never called).

**Scope.** `cfg_builder` models only top-level `ast.If`; `While`/`For`/`Try`/`With`/`Return`/`break`/`continue` and nested/`elif` `If`s are appended to a block and then silently ignored by the solver. The analysis should fail closed (skip annotation) on unsupported constructs rather than annotate from a partial view of the body.

**Packaging.** `src/python-frontend/CMakeLists.txt` mangles `*.py` recursively, so `pytype_infer/tests/example_src.py` and `test_integration.py` are compiled into the `esbmc` binary and extracted to `/tmp` on every run (I confirmed both under `/tmp/esbmc-python-astgen-*/pytype_infer/tests/`). Move them to `unit/python-frontend/`, where the project's Python tests live, and make them real pytest tests — `test_integration.py` is a module-level script with no assertions and no `__init__.py`.

Inline comments have the specifics.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+                return IntType()
+            if f_name == "bool":
+                return BoolType()
+
+        if isinstance(func, ast.Attribute):
+            obj_type = infer_type_from_expr(func.value, env)
+            method = func.attr
+            arg_types = [infer_type_from_expr(a, env) for a in expr.args]
+
+            return
```
lucasccordeiro: **Blocker — this crashes ESBMC on most Python inputs.**

The `ast.Attribute` branch ends in a bare `return`, so `infer_type_from_expr` yields `None` for every method call. That `None` is stored into `env` at line 83 and then joined at line 130:

```
AttributeError: 'NoneType' object has no attribute 'join'
```

Reproduced end-to-end with a build of this branch:

```python
def f(s):
    a = s.upper()
    return a
f("hi")
```
```
File ".../preprocessor/module_lifecycle_mixin.py", line 20, in finalize_module
File ".../pytype_infer/dataflow_solver.py", line 130, in analyze_function
AttributeError: 'NoneType' object has no attribute 'join'
```

`parser.py` dies, so ESBMC produces no result at all. I ran the first 420 `regression/python` tests against this branch: **31 fail**, all with this traceback (`bytes_hex`, `bytes_fromhex`, `bytes_encode_decode`, the whole `cmath_polar_rect_semantics_*` family, …). Any assignment from `str.upper()`, `list.pop()`, `dict.get()` … is affected.

Fix: call `resolve_method_call(obj_type, method, arg_types)` here and make every path of this function return a `Type` (never `None`). Please also add an assertion/`or Unknown()` guard at the call site so a future missing return degrades instead of crashing.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+        if method == "split":
+            return ListType(StrType())
+
+    if isinstance(obj_type, DictType):
+        if method in ["get", "pop"]:
+            return obj_type.v
+        if method == "keys":
+            return ListType(obj_type.k)
+
+    return Unknown()
```
lucasccordeiro, line 582: `resolve_method_call` is dead and does not run — pylint reports 10 × `E0602 undefined-variable`:

- line 209: the parameter is `onj_type`, the body reads `obj_type`;
- lines 222-230 and 232-240 are byte-identical duplicated `StrType` blocks;
- lines 244/246 use `obj_type.v` / `obj_type.k`, but `DictType` (lattice.py:232-234) defines `val_t` / `key_t` — `AttributeError` once reached.

CI lint fails on this file. Fix the parameter name, delete the duplicate block, use `val_t`/`key_t`, and wire the function into `infer_type_from_expr` (see the comment on line 320) so it is actually covered.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+    if isinstance(expr, ast.Tuple):
+        elems = [infer_type_from_expr(e, env) for e in expr.elts]
+        return TupleType(elems)
+    if isinstance(expr, ast.Dict):
+        k = Unknown()
+        v = Unknown()
+        for key, val in zip(expr.keys, expr.values):
+            k = k.join(infer_type_from_expr(key, env))
+            v = v.join(infer_type_from_expr(val, env))
+        return DictType(k, v)
```
lucasccordeiro, line 692: Unreachable: the `return Unknown()` on line 327 precedes these branches (pylint `W0101`). Tuple and dict literals are therefore never inferred, so `TupleType` and `DictType` are effectively unused. Move the fallback `return Unknown()` after these two branches.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+                for p in block.pred:
+                    pred = out_envs[p]
+                    for k, v in pred.items():
+                        if k in pred_env:
+                            pred_env[k] = pred_env[k].join(v)
+                        else:
+                            pred_env[k] = v
+            else:
+                pred_env = in_envs[i] if in_envs[i] else {}
+            in_envs[i] = pred_env.copy()
```
lucasccordeiro: **The narrowing computed at lines 137-179 has no effect.**

Lines 164-179 merge the narrowed `t_env` / `f_env` into `in_envs[succ_true]` / `in_envs[succ_false]`, but on the next visit of those blocks line 53-64 recomputes `pred_env` purely from `out_envs[p]` and then *overwrites* `in_envs[i]` with it. The narrowed values are discarded before any statement sees them.

Verified on `regression/python/flow_sensitive_narrowing/main.py`: after the fixpoint, `x` is still `Unknown` in every block, so `isinstance(x, int)` never refines it.

The narrowed environment has to be attached to the *edge*, and the block's entry environment computed as `join(narrow(out_envs[p], edge_p→i))`, not as `join(out_envs[p])`.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,440 @@
...
+class Type:
+
+    def join(self, other: 'Type') -> 'Type':
+        raise NotImplementedError
+
+    def widen(self, other: 'Type') -> 'Type':
+        return self.join(other)
+
+    def narrow_with_isinstance(self, cls_name: str) -> 'Type':
+        return Unknown()
```
lucasccordeiro, line 14: `Type.narrow_with_isinstance` returning `Unknown()` is the second reason narrowing is a no-op. `Unknown` (line 52) does not override it, and every parameter starts as `Unknown` (dataflow_solver.py:44), so `isinstance(x, int)` on a parameter refines `Unknown` → `Unknown`.

For the ⊥/no-information case the result should be the asserted type, i.e. `mk_type_from_name(cls_name)`. Without this the headline feature of the PR cannot fire on function parameters — which is exactly what both new regression tests are meant to exercise.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,440 @@
...
+class Unknown(Type):
+
+    def join(self, other: 'Type') -> 'Type':
+        return other
+
+    def is_subtype_of_name(self, cls_name: str) -> bool:
+        return False
+
+    def to_ann_name(self) -> str:
+        return 'unknown'
```
lucasccordeiro, line 61: `Unknown` is doing double duty: it is the join identity / entry value (⊥, "no information yet") *and* the result of a failed narrow ("this branch is impossible") throughout `narrow_with_isinstance` / `remove_isinstance`. Because `X.join(Unknown) == X`, an impossible branch silently contributes nothing instead of being recognised as unreachable, and a genuinely-unanalysed variable is indistinguishable from a contradiction.

Please introduce a distinct `Bottom`/`Never` element, or document precisely which of the two meanings `Unknown` carries at each use site.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+                    print(ast.dump(stmt, indent=2))
+                    targets = stmt.targets if hasattr(stmt, 'targets') else [stmt.target]
+                    if len(targets) >= 1 and isinstance(targets[0], ast.Name):
+                        name = targets[0].id
+                        val = stmt.value
+                        t = infer_type_from_expr(val, env)
+                        if name in env:
+                            env[name] = env[name].join(t)
+                        else:
+                            env[name] = t
```
lucasccordeiro: Assignment performs a *weak* update (`env[name].join(t)`) instead of a strong update. Python assignment rebinds, so `x = str(x)` after `x: int` must give `str`, not `Union[int, str]`.

This directly defeats the PR's purpose: in `regression/python/flow-sensitive-narrowing/main.py` every rebinding of `x` accumulates, and the merge into `out_envs` at lines 126-132 joins with the previous iterate too, so nothing is ever killed. Use `env[name] = t` and let the *CFG merge* (not the assignment) be the only place a join happens.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,341 @@
...
+    in_envs = [dict() for _ in range(n)]
+    out_envs = [dict() for _ in range(n)]
+    entry_env = {}
+    for arg in func_node.args.args:
+        entry_env[arg.arg] = Unknown()
+    if n > 0:
+        in_envs[0] = entry_env.copy()
+    changed = True
+    iteration = 0
+    while changed and iteration < max_iters:
```
lucasccordeiro, line 60: Bailing out at `max_iters` silently returns a **non-fixpoint** result which `annotate_ast` then trusts to emit annotations. Since those annotations feed ESBMC's type assignment, a premature result is a verification-soundness problem, not a precision one.

Note also that `widen()` is implemented on every class in `lattice.py` but is never called from this solver — the lattice has infinite height (`ListType` nesting, `UnionType` growth), so there is nothing guaranteeing termination. Either apply widening after N iterations, or raise/skip annotation for the function when the loop does not converge; do not silently annotate from a partial iterate.

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,51 @@
...
+            cur.succ.extend([idx_then, idx_else])
+            cfg.blocks[idx_then].pred.append(cfg.blocks.index(cur))
+            cfg.blocks[idx_else].pred.append(cfg.blocks.index(cur))
+            cfg.blocks[idx_then].succ.append(idx_join)
+            cfg.blocks[idx_else].succ.append(idx_join)
+            cfg.blocks[idx_join].pred.extend([idx_then, idx_else])
+            cur = join_block
+        else:
+            cur.stmts.append(stmt)
+    return cfg
```
lucasccordeiro: The CFG only models `ast.If` at the *top level* of the function body. Everything else — `While`, `For`, `Try`, `With`, `Return`, `Break`, `Continue`, and **nested / `elif` `If`s** — is appended verbatim to `cur.stmts`, and `analyze_function` only handles `Assign` / `AnnAssign` / `Expr`, so those statements are silently ignored.

Consequence for this PR's own test: `regression/python/flow-sensitive-narrowing/main.py` puts the whole `if/elif/else` chain inside a `while`, so the entire chain lands in one block as a single `ast.While` node and is never analysed. The test does not exercise the feature it is named for.

Please either handle loops/nested statements, or make `build_cfg_for_function` explicitly refuse (and skip annotation for) functions containing unsupported constructs, so the analysis fails closed rather than producing types derived from a partial view of the body.

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,51 @@
...
+            # join block
+            join_block = Block()
+            idx_then = len(cfg.blocks)
+            cfg.blocks.append(then_block)
+            idx_else = len(cfg.blocks)
+            cfg.blocks.append(else_block)
+            idx_join = len(cfg.blocks)
+            cfg.blocks.append(join_block)
+            cur.succ.extend([idx_then, idx_else])
+            cfg.blocks[idx_then].pred.append(cfg.blocks.index(cur))
```
lucasccordeiro: `cfg.blocks.index(cur)` is an O(n) identity scan on every `If`, and it is fragile: `Block` has no `__eq__`, so this happens to work only by object identity. Track the current block's index alongside `cur`.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,101 @@
...
+                    ann_name = t.to_ann_name()
+                    ann = ast.AnnAssign(target=ast.Name(id=name, ctx=ast.Store()),
+                                        annotation=ast.Name(id=ann_name, ctx=ast.Load()),
+                                        value=stmt.value,
+                                        simple=1)
+                    ast.copy_location(ann, stmt)
+                    new_body.append(ann)
+                    continue
+        new_body.append(stmt)
+    func_node.body = new_body
```
lucasccordeiro, line 119: Only *top-level* statements of `func_node.body` are rewritten, so assignments inside `if` / `while` bodies — i.e. exactly the flow-sensitive cases this PR targets — are never annotated.

Confirmed on `regression/python/flow_sensitive_narrowing/main.py`: the solver infers `y: Union[int, str]` but the emitted AST is byte-identical to the input, no annotation at all. Recurse with an `ast.NodeTransformer` over the whole body.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,101 @@
...
+        if isinstance(stmt, ast.Assign):
+            if len(stmt.targets) >= 1 and isinstance(stmt.targets[0], ast.Name):
+                name = stmt.targets[0].id
+                if name in merged and not isinstance(merged[name], Unknown):
+                    t = merged[name]
+                    ann_name = t.to_ann_name()
+                    ann = ast.AnnAssign(target=ast.Name(id=name, ctx=ast.Store()),
+                                        annotation=ast.Name(id=ann_name, ctx=ast.Load()),
+                                        value=stmt.value,
+                                        simple=1)
```
lucasccordeiro: **Invalid annotation nodes.** Annotations are built as `ast.Name(id=t.to_ann_name())`, but `to_ann_name()` returns composite strings: `Union[int,str]`, `list[int]`, `dict[int,str]`, `Callable[int->str]`, `None`, `unknown`, `object`. A `Name` whose `id` contains `[`/`,`/`->` is not a well-formed AST node — it survives `ast.unparse` only by accident and is not what the C++ frontend receives (it consumes the AST via ast2json).

Observed output on this branch:
```python
w: Union[int,str] = v   # ast.Name(id='Union[int,str]'); `Union` is never imported either
```
`Callable[int->str]` is not even parseable Python.

Generics must be emitted as `ast.Subscript(value=Name('list'), slice=...)`, `None` as `ast.Constant(None)`, and `unknown` should mean "emit no annotation" rather than a `Name` with id `unknown`. Please also add a round-trip guard: `ast.parse(ast.unparse(tree))` on the annotated module before returning it.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,101 @@
...
+def annotate_function_with_env_and_signatures(func_node: ast.FunctionDef, out_envs: list):
+    param_types, ret_type = derive_param_and_return_types(func_node, out_envs)
+    # annotate params
+    for arg in func_node.args.args:
+        t = param_types.get(arg.arg, Unknown())
+        if not isinstance(t, Unknown):
+            arg.annotation = ast.Name(id=t.to_ann_name(), ctx=ast.Load())
+    # annotate return
+    if not isinstance(ret_type, Unknown):
+        func_node.returns = ast.Name(id=ret_type.to_ann_name(), ctx=ast.Load())
```
lucasccordeiro: **Unsound signatures.** Parameters are annotated from `merged` — the join of the *out* environments — i.e. the types the parameter holds *after* being reassigned inside the body, not the types a caller may pass.

`regression/python/flow-sensitive-narrowing/main.py` rebinds `x` to `str`, `list` and `int`, so `normalize` would be annotated as accepting all of them although the only call site passes `5`. Since ESBMC uses these annotations to pick the parameter's type, a wrong annotation changes the verification verdict. Parameter types must come from the *entry* environment (or from call sites), never from `out_envs`.

Line 37 has the same problem in reverse: `derive_param_and_return_types` infers the return type with an empty env (line 12), so `return y` is always `Unknown`, and when it does resolve it emits the invalid composite `Name` described above.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,101 @@
...
+    new_body = []
+    for stmt in func_node.body:
+        # convert name = lambda ... to AnnAssign(name: callable = lambda ...)
+        print(type(stmt))
+        if isinstance(stmt, ast.Assign) and len(stmt.targets) >= 1 and isinstance(
+                stmt.targets[0], ast.Name) and isinstance(stmt.value, ast.Lambda):
+            name = stmt.targets[0].id
+            lam = stmt.value
+            ret_t = infer_type_from_expr(lam.body, {})
+            ann_name = 'callable' if isinstance(ret_t, Unknown) else 'callable'
```
lucasccordeiro: `ann_name = 'callable' if isinstance(ret_t, Unknown) else 'callable'` — both arms are identical, and `ret_t` (line 55) is computed only to be discarded. Either use the inferred `CallableType` or drop lines 55-56.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,440 @@
...
+            return other.join(self)
+        return UnionType([self, other])
+
+    def widen(self, other: 'Type') -> 'Type':
+        if isinstance(other, TupleType):
+            if len(self.elems) == len(other.elems):
+                return TupleType([a.widen(b) for a, b in zip(self.elems, other.elems)])
+            if isinstance(other, UnionType):
+                return other.widen(self)
+            return other
```
lucasccordeiro, line 290: `TupleType.widen` returns `None` when `other` is not a `TupleType` (control falls off the end of the outer `if`). Also the `isinstance(other, UnionType)` test on line 288 sits *inside* the `TupleType` branch, so it can never be true — the union case is unreachable. Compare with `ListType.widen` (line 205) / `DictType.widen` (line 245), which handle the fallthrough.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def join(self, other: 'Type') -> 'Type':
+        if isinstance(other, Unknown):
+            return self
+        if isinstance(other, UnionType):
+            return UnionType(self.members + other.members)
+        if any(isinstance(m, FloatType) for m in self.members) and isinstance(other, IntType):
+            new_members = [FloatType() if isinstance(x, IntType) else x for x in self.members]
+            return UnionType(new_members)
+        return UnionType(self.members + [other])
```
lucasccordeiro, line 449: `join` here is not a lattice join: `IntType.join(FloatType)` → `Float` (line 122) and this branch rewrites every `IntType` member of a union to `FloatType`. That bakes the numeric tower into the merge operator and loses int-ness, so a variable that is only ever an `int` on some path can end up annotated `float`. Given the annotation drives ESBMC's type assignment, this is verdict-affecting — keep `join` a plain set union and apply numeric promotion where the promotion actually happens (`BinOp`), not in the lattice.

src/python-frontend/preprocessor/module_lifecycle_mixin.py
```diff
@@ -7,12 +8,18 @@ def finalize_module(self, node):
...
         self._assignment_call_origins.clear()
         try:
             node = self.generic_visit(node)
+            # try:
+            ast.fix_missing_locations(node)
+            print("Before pytype module annotations:")
+            print(ast.unparse(node))
+            node = pytype_infer.annotate_tree(node)
+            print("After pytype annotation module:")
+            print(ast.unparse(node))
```
lucasccordeiro, line 22: **Blocker — debug output on every run.** These four `print()`s dump the entire unparsed module to stdout twice for every module ESBMC processes. `annotate_ast.py` (lines 46, 50, 96, 97) and `dataflow_solver.py` (lines 32-37, 74) add per-block AST dumps on top.

Line 16 is also a leftover commented-out `# try:`.

Please remove all of them, or gate them behind an existing verbosity flag.

Separately: `pytype_infer.annotate_tree` is invoked unconditionally with no way to opt out. A brand-new whole-module type-inference pass in the default path needs a command-line switch (off by default until it is stable) and a `try/except` so an internal failure degrades to "no annotations" rather than killing `parser.py`.

src/python-frontend/preprocessor/module_lifecycle_mixin.py
```diff
@@ -7,12 +8,18 @@ def finalize_module(self, node):
         """Run generic_visit and inject helper nodes requested during traversal."""
         # Per-module scope for the eq-only set and call-origin map.
         saved_eq_only = set(self._eq_only_items_view_targets)
-        self._eq_only_items_view_targets = (
-            self._scan_eq_only_items_view_targets(node.body))
+        self._eq_only_items_view_targets = (self._scan_eq_only_items_view_targets(node.body))
```
lucasccordeiro, line 11: Unrelated re-wrap. Please revert this and the blank line added at `core_visitors_mixin.py:1323` so the diff only contains functional changes — the "apply YAPF auto-formatting" commit is churn against a file this PR does not otherwise touch.

src/python-frontend/preprocessor/core_visitors_mixin.py
```diff
@@ -1,5 +1,6 @@
 import ast
 import copy
+import pytype_infer
```
lucasccordeiro: `import pytype_infer` is unused in this file (pylint `W0611`). Drop it, along with the unrelated blank line added at line 1323.

src/python-frontend/pytype_infer/infer_and_annotate.py
```diff
@@ -0,0 +1,34 @@
...
+    tree = ast.parse(src)
+    annotated = annotate_tree(tree, filename=args.infile)
+    try:
+        s = ast.unparse(annotated)
+    except Exception:
+        try:
+            import astor
+            s = astor.to_source(annotated)
+        except Exception:
+            s = '# could not unparse annotated AST\n' + ast.dump(annotated)
```
lucasccordeiro, line 28: `import astor` — `astor` is not a dependency of this project and is not installed on CI (pylint `E0401`). `ast.unparse` is available on every Python version the frontend supports (3.10+), so drop the fallback. `sys` on line 1 is unused, and `filename` on line 5 is accepted and ignored (`W0613`).

src/python-frontend/pytype_infer/parser.py
```diff
@@ -0,0 +1,6 @@
+import ast, json
+
+
+def read_source_file(path):
+    with open(path, 'r', encoding='utf-8') as f:
+        return f.read()
```
lucasccordeiro: This whole file is dead: `read_source_file` has no callers and both `ast` and `json` are unused. It also shadows the name of the frontend's real entry point package `src/python-frontend/parser/`, which is actively confusing. Please delete it.

regression/python/flow_sensitive_narrowing/main.py
```diff
@@ -0,0 +1,10 @@
+def test(x):
+
+   if isinstance(x, int):
+      y = x + 1
+      assert isinstance(y, int)
+   else:
+      y = str(x)
+      assert isinstance(y, str)
+
+   return y
```
lucasccordeiro, line 10: **Vacuous test.** `test()` is never called, so nothing is verified:

```
Slicing time: 0.000s (removed 74 assignments)
Generated 0 VCC(s), 0 remaining after simplification
VERIFICATION SUCCESSFUL
```

`Generated 0 VCC(s)` — this passes identically with `assert False` in the body. Add a call (`test(5)` / `test("s")`) at module level.

It is also nearly indistinguishable from `regression/python/flow-sensitive-narrowing/` (hyphen vs. underscore) added in the same PR. Keep one directory with a distinguishing name.

regression/python/flow_sensitive_narrowing/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+
+main.py
+^VERIFICATION SUCCESSFUL
```
lucasccordeiro, line 3: `test.desc` format is: line 1 mode, line 2 source file, line 3 flags, line 4+ regexes. The blank line 2 shifts everything: the harness (`regression/testing_tool.py:274-279`) reads `test_file=""` and `test_args="main.py"`, i.e. `main.py` is passed as an *option* string, not as the test's input file. Delete the blank line.

regression/python/flow-sensitive-narrowing/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+--no-unwinding-assertions --unwind 5
+^VERIFICATION SUCCESSFUL
```
lucasccordeiro: **This test does not pass.** Marked `CORE` with `^VERIFICATION SUCCESSFUL`, but on a build of this branch:

```
Violated property:
  file main.py line 23 column 4 function normalize
  assertion 0
VERIFICATION FAILED
```

(line 23 = `assert isinstance(result[2], int)`.) It should not be marked `CORE` until it passes.

Two more points on the flags:
- `--no-unwinding-assertions` is not needed — `while i < 4` with `--unwind 5` unwinds fully — and it silently hides unwinding failures if the bound is later changed. Drop it.
- Per `CLAUDE.md`, every PR should carry at least two regression tests, one passing and one *failing*. There is no negative test here pinning the analysis against over-narrowing (e.g. a program where a wrong `isinstance` annotation would make ESBMC report SUCCESSFUL for a real bug) — that is the test that would actually protect soundness.

regression/python/flow-sensitive-narrowing/main.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+         result.append(x)
+         i += 1
+
+    assert len(result) == 4
+
+
+    assert isinstance(result[0], str)
+    assert isinstance(result[1], list)
```
lucasccordeiro: This assertion is the one that fails (`assertion 0` at line 23). Note that the `if/elif/else` chain is inside a `while`, which `cfg_builder.build_cfg_for_function` does not model at all — the whole loop lands in a single block as one `ast.While` node and is skipped. So even once the crash and the dead-narrowing bugs are fixed, this file does not exercise the new analysis. Please add a test whose narrowing happens in a construct the CFG builder actually supports, and a separate one for the loop case once loops are handled.

src/python-frontend/pytype_infer/dataflow_solver.py
mihaistate: Ok, I improved resolve_method_call and will call it there

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,51 @@
...
+            cfg.blocks[idx_else].pred.append(cfg.blocks.index(cur))
+            cfg.blocks[idx_then].succ.append(idx_join)
+            cfg.blocks[idx_else].succ.append(idx_join)
+            cfg.blocks[idx_join].pred.extend([idx_then, idx_else])
+            cur = join_block
+        else:
+            cur.stmts.append(stmt)
```
lucasccordeiro: `While`/`For` land here, so loops get no back edge and no fixpoint — yet `flow-sensitive-narrowing/main.py` depends on a `while`. Build loop blocks.

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,51 @@
...
+    entry = Block()
+    cfg.blocks.append(entry)
+    cur = entry
+    for stmt in func_node.body:
+        if isinstance(stmt, ast.If):
+            # current block gets condition entry marker (store test)
+            cur.stmts.append(('cond', stmt.test))
+            # then block
+            then_block = Block()
+            then_block.stmts.extend(stmt.body)
```
lucasccordeiro: A nested `If` inside `stmt.body` is copied verbatim and never split into blocks, so narrowing is wrong past depth one. Recurse when building branch blocks.

src/python-frontend/preprocessor/module_lifecycle_mixin.py
```diff
@@ -7,12 +8,18 @@ def finalize_module(self, node):
...
         saved_call_origins = dict(self._assignment_call_origins)
         self._assignment_call_origins.clear()
         try:
             node = self.generic_visit(node)
+            # try:
+            ast.fix_missing_locations(node)
+            print("Before pytype module annotations:")
```
lucasccordeiro, line 18: Debug `print()` on the production preprocessor path pollutes stdout on every run. Remove these and the ones in `annotate_ast.py`/`dataflow_solver.py`.

src/python-frontend/preprocessor/module_lifecycle_mixin.py
```diff
@@ -7,12 +8,18 @@ def finalize_module(self, node):
...
         # Per-module scope for the eq-only set and call-origin map.
         saved_eq_only = set(self._eq_only_items_view_targets)
-        self._eq_only_items_view_targets = (
-            self._scan_eq_only_items_view_targets(node.body))
+        self._eq_only_items_view_targets = (self._scan_eq_only_items_view_targets(node.body))
         saved_call_origins = dict(self._assignment_call_origins)
         self._assignment_call_origins.clear()
         try:
             node = self.generic_visit(node)
+            # try:
```
lucasccordeiro, line 16: Commented-out `# try:` left behind. Delete it.

regression/python/flow_sensitive_narrowing/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+
```
lucasccordeiro: Line 2 must be the source file; the blank line shifts `main.py` into the flags slot (testing_tool.py:346). Remove the blank line.

regression/python/flow-sensitive-narrowing/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
```
lucasccordeiro: Directory name duplicates `flow_sensitive_narrowing` and uses hyphens against the suite convention. Rename to a distinct underscore name.

regression/python/flow-sensitive-narrowing/test.desc
```diff
@@ -0,0 +1,4 @@
...
+main.py
+--no-unwinding-assertions --unwind 5
```
lucasccordeiro: `--no-unwinding-assertions` makes SUCCESSFUL vacuous if 5 unwinds do not cover the loop. Drop the flag and keep `--unwind 5`.

src/python-frontend/pytype_infer/tests/test_integration.py
```diff
@@ -0,0 +1,12 @@
+import ast, os
```
lucasccordeiro, line 1: The recursive `*.py` mangle in src/python-frontend/CMakeLists.txt will FLAIL these test files into the binary, and this one asserts nothing. Move it under `unit/`.

lucasccordeiro (changes requested):

src/python-frontend/pytype_infer/dataflow_solver.py
mihaistate: Added a new resolve_method_call function and solved these bugs

src/python-frontend/pytype_infer/cfg_builder.py
mihaistate: I created separate functions for statements and the for the if block, now the narrowing works past depth one

src/python-frontend/pytype_infer/cfg_builder.py
mihaistate: I created separate methods for while/for loops, continue and break statements

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,213 @@
...
+        cfg,
+        body_idx,
+        loop_ctx,
+    )
+
+    if body_tail is not None:
+        add_edge(cfg, body_tail, header_idx)
+
+    if else_idx is not None:
+        else_tail = build_statements(cfg, stmt.orelse, else_idx, outer_loop)
```
lucasccordeiro: First two arguments are swapped, so any `for ... else` raises `TypeError: 'CFG' object is not iterable` and kills `parser.py`. Reproduced. Use `(stmt.orelse, cfg, else_idx, outer_loop)`.

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,213 @@
...
+
+    return exit_idx
+
+
+def build_statements(statements, cfg, current_idx: int, loop_ctx: LoopContext | None = None):
+    current = current_idx
+
+    for stmt in statements:
+        if isinstance(stmt, ast.If):
+            current = build_if(stmt, cfg, current)
```
lucasccordeiro: `build_if` receives no `loop_ctx`, so a `break`/`continue` inside an `if` inside a loop adds no edge. Thread `loop_ctx` through `build_if` into its `build_statements` calls.

src/python-frontend/pytype_infer/cfg_builder.py
```diff
@@ -0,0 +1,213 @@
...
+
+        elif isinstance(stmt, ast.While):
+            current = build_while(cfg, stmt, current, loop_ctx)
+        elif isinstance(stmt, ast.For):
+            current = build_for(cfg, stmt, current, loop_ctx)
+        elif isinstance(stmt, ast.Break):
+            if loop_ctx is not None:
+                add_edge(cfg, current_idx, loop_ctx.break_target)
```
lucasccordeiro: Edge originates at `current_idx` (the block on entry) rather than `current`, so a `break` following an `if` links the wrong block. Use `current` here and at line 175.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+
+
+def handle_call_effect(call, env, context):
+    if not isinstance(call.func, ast.Attribute):
+        return
+
+    if call.func.attr == "append":
+        handle_append(call, env, context)
+
+    elif call.func.attr == "setDefault":
```
lucasccordeiro, line 408: `"setDefault"` never matches a real call, and `handle_setdfault` requires `"setdefault"` anyway, so the handler is unreachable. Fix the spelling.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+def propagate_to_successors(successors, env, in_envs):
+    for successor in successors:
+        merge_successor_environment(
+            successor,
+            env,
+            in_envs,
+        )
+
+
+def resolve_method_call(obj_type, method, arg_types):
```
lucasccordeiro, line 591: `resolve_method_call` is unreachable for str/list/dict/tuple: `infer_attribute_type` returns a `CallableType` at line 979 first. Confirmed by instrumenting it. Delete this duplicate.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+    if isinstance(expr.op, ast.Add):
+        result = infer_add_type(left, right)
+
+        if result is not None:
+            return result
+
+    if isinstance(left, FloatType) or isinstance(right, FloatType):
+        return FloatType()
+
+    if isinstance(left, IntType) and isinstance(right, IntType):
```
lucasccordeiro, line 932: `ast.Div` falls through to this int/int rule, so `1 / 2` infers `int`; Python true division yields `float`. Handle `Div` separately from `FloorDiv`.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+    if isinstance(expr, ast.Name):
+        return env.get(expr.id, Unknown())
+
+    if isinstance(expr, ast.Attribute):
+        obj_type = infer_type_from_expr(expr.value, env, context)
+        if isinstance(obj_type, ComplexType):
+            if expr.attr == "real":
+                return FloatType()
+
+        if expr.attr == "imag":
```
lucasccordeiro, line 1611: This test sits outside the `ComplexType` guard, so `s.imag` on a `str` infers `float`. Nest it inside the `isinstance(obj_type, ComplexType)` block.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+
+    if isinstance(container_type, SetType):
+        return Unknown()
+
+    return Unknown()
+
+
+def is_type_compatible(actual, expected):
+    if isinstance(actual, Unknown) or isinstance(expected, Unknown):
+        return Type
```
lucasccordeiro, line 1389: Returns the class object `Type` instead of `True`; it only works because a class is truthy. Return `True`.

src/python-frontend/pytype_infer/dataflow_solver.py
```diff
@@ -0,0 +1,1780 @@
...
+            context,
+        )
+
+    if isinstance(expr, ast.Lambda):
+        return infer_lambda_type(expr, env, context)
+
+    return Unknown()
+
+
+""" def infer_type_from_expr(expr, env):
```
lucasccordeiro: 127 lines of a superseded `infer_type_from_expr` kept in a string literal. Delete it, along with the dead blocks at 230-242, 733-768 and 1452-1469.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+    for node in module_node.body:
+        if not isinstance(node, ast.FunctionDef):
+            continue
+        if node.name not in out_envs_by_func:
+            continue
+
+        out_envs = out_envs_by_func[node.name]
+        return_types = return_types_by_func.get(node.name, {})
+        #print(ast.dump(module_node, indent=4))
+        print(ast.unparse(module_node))
```
lucasccordeiro, line 437: Debug output still runs on every module; this line unparses the whole module once per function. Remove these and `dataflow_solver.py` lines 912, 1091, 1439.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+        return ast.Name(id="Any", ctx=ast.Load())
+
+    if isinstance(t, ListType):
+        elem = type_to_ast_annotation(t.elem)
+
+        if elem is None:
+            elem = ast.Name(id="Any", ctx=ast.Load())
+
+        return ast.Subscript(
+            value=ast.Name(id="List", ctx=ast.Load()),
```
lucasccordeiro, line 309: Emits `List[...]`, but `type_utils::is_builtin_type` accepts only lowercase `list` and `List` is never imported. Emit `ast.Name(id="list")`, as the `DictType` branch already does.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+        return ast.Name(id="float", ctx=ast.Load())
+
+    if isinstance(t, StrType):
+        return ast.Name(id="str", ctx=ast.Load())
+
+    if isinstance(t, ComplexType):
+        return ast.Name(id="complex", ctx=ast.Load())
+
+    if isinstance(t, NoneType):
+        return ast.Name(id="None", ctx=ast.Load())
```
lucasccordeiro: `None` is a keyword, so `ast.Name(id="None")` is not a well-formed annotation node. Emit `ast.Constant(value=None)`.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+
+    if isinstance(t, SetType):
+        elem = type_to_ast_annotation(t.elem)
+
+        if elem is None:
+            elem = ast.Name(id="Any", ctx=ast.Load())
+
+        return ast.Subscript(
+            value=ast.Name(id="Set", ctx=ast.Load()),
+            slice=type_to_ast_annotation(t.elem),
```
lucasccordeiro: Recomputes `type_to_ast_annotation(t.elem)`, discarding the fallback assigned at lines 334-337, so `slice` can be `None`. Use `elem`.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+            ctx=ast.Load(),
+        )
+
+    if isinstance(t, InstanceType):
+        return ast.Name(
+            id=t.class_name,
+            ctx=ast.Load(),
+        )
+
+    raise TypeError(f"Cannot convert lattice type to AST annotation: "
```
lucasccordeiro: Raising here aborts `parser.py`, since `module_lifecycle_mixin` calls `annotate_tree` with no guard. Return `None` so the caller skips the annotation instead.

src/python-frontend/pytype_infer/annotate_ast.py
```diff
@@ -0,0 +1,493 @@
...
+    return module_node
+
+
+def annotate_ast(ast_node, opts=None):
+    known_classes = collect_known_classes(ast_node)
+    context = InferenceContext(known_classes=known_classes, )
+    out_envs_by_func = {}
+    return_types_by_func = {}
+    for node in ast_node.body:
+        if isinstance(node, ast.FunctionDef):
```
lucasccordeiro, line 458: Only module-level `FunctionDef`s are analysed, so methods, nested functions and `AsyncFunctionDef` are never inferred despite `collect_known_classes` collecting class attributes.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,631 @@
...
+        if isinstance(other, Unknown):
+            return self
+        if isinstance(other, TupleType):
+            if len(self.elems) == len(other.elems):
+                return TupleType([a.widen(b) for a, b in zip(self.elems, other.elems)])
+
+            return UnionType([self, other])
+        if isinstance(other, UnionType):
+            return other.widen(self)
+        return other
```
lucasccordeiro, line 349: `TupleType.join` returns `other` for unrelated types, so `Tuple[int].join(Str)` is `Str` — not an upper bound. Return `UnionType([self, other])`.

src/python-frontend/pytype_infer/lattice.py
```diff
@@ -0,0 +1,631 @@
...
+    def narrow_with_isinstance(self, cls_name: str) -> 'Type':
+        return self if self.is_subtype_of_name(cls_name) else Unknown()
+
+
+class InstanceType(Type):
+
+    def __init__(self, class_name: str):
+        self.class_name = class_name
+
+    def join(self, other):
```
lucasccordeiro, line 470: No `Unknown` case, so `InstanceType('Foo').join(Unknown())` gives `Union[Foo, Unknown]` and pollutes every merge. Return `self` when `other` is `Unknown`.

regression/python/flow_sensitive_narrowing/test.desc
```diff
@@ -0,0 +1,3 @@
+CORE
+main.py
+^VERIFICATION SUCCESSFUL
```
lucasccordeiro, line 3: Regex is unanchored, and there is no `VERIFICATION FAILED` counterpart. Add `$` and the failing half of the test pair CLAUDE.md requires.

src/python-frontend/preprocessor/core_visitors_mixin.py
```diff
@@ -1296,6 +1296,11 @@ def visit_Assign(self, node):
         """
         Handle assignment nodes, including multiple assignments and tuple unpacking.
         """
+        import ast
```
lucasccordeiro, line 1299: `ast` is already imported at module level (line 1), and the commented-out prints below are unrelated churn. Revert this hunk and the one at line 1327.

src/python-frontend/pytype_infer/lattice.py
mihaistate: I will introduce a Bottom type in the lattice file, as a return annotation type in type_to_ast_annotation, where it returns the Any type, and in the derive_param_and_return_types, where ret_type variable is initialised with a Bottom() value before adding more concrete types.

src/python-frontend/pytype_infer/cfg_builder.py
mihaistate: Done

src/python-frontend/pytype_infer/dataflow_solver.py
mihaistate: I removed resolve_method_call

src/python-frontend/pytype_infer/dataflow_solver.py
mihaistate: I added Div and FloorDiv cases before the  FloatType and IntType cases: 
    if isinstance(expr.op, (ast.Div, ast.FloorDiv)):
        numeric_types = (BoolType, IntType, FloatType)

        if not (
            isinstance(left, numeric_types)
            and isinstance(right, numeric_types)
        ):
            return Unknown()

        if isinstance(expr.op, ast.Div):
            return FloatType()

        # Floor division returns float if either operand is float.
        if isinstance(left, FloatType) or isinstance(right, FloatType):
            return FloatType()

        return IntType()

src/python-frontend/pytype_infer/annotate_ast.py
mihaistate: Done

src/python-frontend/pytype_infer/annotate_ast.py
mihaistate: I will put return None instead of raise TypeError

src/python-frontend/pytype_infer/cfg_builder.py
mihaistate: I restructured the cfg_builder with multiple helper functions and cases for while, try, with, break, continue.

PR #6523, opened 2026-07-29
[clang-c-frontend] Model the C11 __c11_atomic_* builtins

### LINKED ISSUE #2174: [concurrency] support Atomic type
```
#include <pthread.h>
#include <stdatomic.h>

atomic_int counter = 0;

void* increment(void* arg) {
    counter++;
    return NULL;
}

int main() {
    pthread_t t1, t2;

    pthread_create(&t1, NULL, increment, NULL);
    pthread_create(&t2, NULL, increment, NULL);

    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    return 0;
}
```

lucasccordeiro:
Clang lowers the `<stdatomic.h>` generic functions to `__c11_atomic_*` builtins,
which had no synthesised bodies: stores were dropped and loads returned nondet,
so correct programs reported FAILED and value-dependent properties were
discharged vacuously, while `atomic_init` aborted with a CONVERSION ERROR.
Register the family with the existing polymorphic-builtin machinery — which
already wraps each body in `__ESBMC_atomic_begin`/`_end` — and add the missing
`AO__c11_atomic_init` name mapping.

Fixes #2174


XLiZHI (approved):
LGTM

Issue #6480, opened 2026-07-28
[concurrency] --deadlock-check does not scale: 0/30 classical benchmark instances solved in 900s


lucasccordeiro:
### Summary

`--deadlock-check` does not scale to textbook concurrent benchmarks. In *Deadlock Verification via Ordering-Constrained Mutex Modeling* (Wang, Han, Sun, He; CAV 2026, doi:10.1007/978-3-032-32519-8_3, Table 1) ESBMC produced **zero correct verdicts on all 30 classical benchmark instances**:

* 15 deadlocking instances (dining philosophers, bank transfer, traffic intersection, tasks-printers, chefs-appliances; 3/5/7 threads): **0 detected** within the 900 s SV-COMP limit;
* 15 deadlock-free instances: only *bounded* deadlock-free, never a completed proof;
* average cost **≈900 s and ≈26 GB** per task, versus 3.7 s / 49 MB for their tool.

On the 288 mutex-using SV-Benchmarks concurrency tasks (Table 2), ESBMC additionally **missed 4 real deadlocks** purely because it could not find a counterexample inside the time limit, and averaged 593 s / 25.3 GB.

Filed as the scalability half of the paper's findings; the soundness half is #6474 and #6479.

### Local reproduction

Dining philosophers, 3 seats, naive (deadlocking) version, as written in the usual tutorial form with `for` loops over the philosophers:

```c
#include <pthread.h>
#define N 3
pthread_mutex_t fork_[N];

void *phil(void *arg)
{
  long i = (long)arg;
  pthread_mutex_lock(&fork_[i]);
  pthread_mutex_lock(&fork_[(i + 1) % N]);
  pthread_mutex_unlock(&fork_[(i + 1) % N]);
  pthread_mutex_unlock(&fork_[i]);
  return 0;
}

int main(void)
{
  pthread_t t[N];
  for (long i = 0; i < N; i++) pthread_mutex_init(&fork_[i], 0);
  for (long i = 0; i < N; i++) pthread_create(&t[i], 0, phil, (void *)i);
  for (int i = 0; i < N; i++) pthread_join(t[i], 0);
  return 0;
}
```

```
$ timeout 300 esbmc phil3.c --deadlock-check --unwind 4
  300.01 real   293.91 user   3.16 sys
  1084915712  maximum resident set size
(no verdict)
```

No verdict after 300 s, ~1 GB RSS and still climbing. `N = 2` also fails to complete in 120 s.

[...]

PR #6534, opened 2026-07-30
[concurrency] Add iterative deepening on the context bound

lucasccordeiro:
Unbounded DFS explores one schedule to full depth before backtracking, so a counterexample needing few context switches can sit far away in DFS order: the #6480 dining-philosophers deadlock needs 4 switches and is found in ~52s at `--context-bound 4`, yet unbounded exploration gives no verdict in 300s.

`--incremental-context-bound` re-explores with the bound raised by one each round, visiting schedules in order of switch count. A violation at any bound is genuine, so the search stops there; success is reported only once a round completes without the bound cutting an available switch, tracked by the new `reachability_treet::cs_bound_pruned`.

Opt-in, so the default path is unchanged. Addresses search order only, not the per-acquisition context-switch cost, so N=3 philosophers is still out of reach.

Towards #6480


lucasccordeiro:
This PR (900s): https://github.com/esbmc/esbmc/actions/runs/30530795931
Master (900s): https://github.com/esbmc/esbmc/actions/runs/30530746600

src/esbmc/parseoptions/driver.cpp
Anthonysdu, line 1: Are we planing to disable incremental-bmc when running incremental-context-bound? These two won't work together.

lucasccordeiro: Good catch — these can't compose, so the combination is now rejected with an error (dc9933f981): `--incremental-context-bound` now refuses `--termination`, `--incremental-bmc`, `--falsification`, `--k-induction`, `--k-induction-parallel` and `--loop-invariant`, with a regression test pinning the message.

PR #6535, opened 2026-07-30
[python] Extend tagged-object representation to scalar branch joins

brcfarias:
A variable's runtime type can get lost across a branch join. ESBMC only remembers whichever type it saw first. This PR lets a variable carry its real type past the join, reusing the tagged-object representation (PyObject) already used for list elements.


src/python-frontend/converter/converter_binop.cpp
```diff
@@ -868,6 +868,86 @@ exprt python_converter::handle_membership_operator(
...
+exprt python_converter::build_tagged_scalar_eq_literal(
+  const exprt &tagged,
+  const exprt &literal)
+{
+  exprt tagged_addr = python_expr::build_address_of(tagged);
+
+  const std::string lit_type_name =
+    type_handler_.type_to_string(literal.type());
+  constant_exprt lit_type_id(size_type());
+  lit_type_id.set_value(integer2binary(
```
lucasccordeiro: `type_id` distinguishes `bool` from `int`, so a tagged `True` compares unequal to literal `1`, contradicting Python (`True == 1`). Hash bool as `"int"` here and in `get_tagged_scalar_assign`.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2365,10 +2438,133 @@ std::string python_converter::call_return_class(const nlohmann::json &rhs) const
...
+    "tagged scalar symbol must already be declared before its branches are "
+    "converted");
+
+  exprt rhs = get_expr(ast_node["value"]);
+  locationt location = get_location_from_decl(ast_node);
+
+  exprt value_to_store = rhs;
+  if (rhs.type() == bool_type())
+    value_to_store = python_expr::build_typecast(
+      rhs, signedbv_typet(config.ansi_c.long_int_width));
```
lucasccordeiro: Bool is stored as `long_int_width`, but `__python_scalar_eq_num` reads `*(const long long *)`; on 32-bit targets this over-reads the 4-byte backing. Use `long_long_int_width` here and at converter_binop.cpp:913.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2365,10 +2438,133 @@ std::string python_converter::call_return_class(const nlohmann::json &rhs) const
...
+      stmt_type == "Assign" && ast_node.contains("targets") &&
+      ast_node["targets"].size() == 1)
+      tag_target = ast_node["targets"][0];
+    else if (stmt_type == "AnnAssign" && ast_node.contains("target"))
+      tag_target = ast_node["target"];
+
+    if (tag_target.is_object() && tag_target.value("_type", "") == "Name")
+    {
+      const std::string name = tag_target["id"].get<std::string>();
+      if (tagged_scalar_names_.count(name))
```
lucasccordeiro: Post-join reassignments (`x = 2` after the `if`) bypass this gate — the scope guard has already popped the name — and assign a raw scalar into the PyObject symbol. Also gate on the symbol's tagged type.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2365,10 +2438,133 @@ std::string python_converter::call_return_class(const nlohmann::json &rhs) const
...
   const nlohmann::json &ast_node,
   codet &target_block)
 {
+  if (!tagged_scalar_names_.empty())
+  {
+    const std::string stmt_type = ast_node.value("_type", "");
+    nlohmann::json tag_target;
+    if (
+      stmt_type == "Assign" && ast_node.contains("targets") &&
+      ast_node["targets"].size() == 1)
```
lucasccordeiro, line 2530: A multi-target assign (`x = y = 1`) to a tagged name fails the `size() == 1` check and silently falls through to the normal path. Reject it like non-literal values.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -84,6 +84,57 @@ bool is_incompatible_scalar_string_retype(const typet &lhs, const typet &rhs)
...
+    else
+      continue;
+
+    if (target.value("_type", "") != "Name" || !target.contains("id"))
+      continue;
+
+    if (!stmt.contains("value") || stmt["value"].is_null())
+      continue;
+    const auto &value = stmt["value"];
+    if (value.value("_type", "") != "Constant" || !value.contains("value"))
```
lucasccordeiro, line 130: A later non-literal assign leaves a stale literal kind (`x = 1; x = f()` stays `"num"`), tagging a variable the assign path then rejects. Erase the name here instead of `continue`.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -4189,6 +4425,42 @@ exprt python_converter::get_conditional_stm(const nlohmann::json &ast_node)
...
+    locationt tag_location = get_location_from_decl(ast_node);
+    std::string tag_module_name = tag_location.get_file().as_string();
+    symbolt tag_symbol = create_symbol(
+      tag_module_name,
+      name,
+      tag_sid.to_string(),
+      tag_location,
+      type_handler_.get_tagged_object_type());
+    tag_symbol.lvalue = true;
+    tag_symbol.file_local = true;
```
lucasccordeiro: Unconditional `file_local = true` breaks the module-scope convention (`file_local = !current_func_name_.empty()`, see line 2932) that rw_set relies on. Match the convention.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2365,10 +2438,133 @@ std::string python_converter::call_return_class(const nlohmann::json &rhs) const
...
+  target_block.copy_to_operands(backing_decl);
+
+  // Byte size of the backing value: array length * element width for a
+  // string literal, plain width for a numeric scalar.
+  exprt elem_size;
+  if (value_to_store.type().is_array())
+  {
+    const array_typet &array_type = to_array_type(value_to_store.type());
+    const size_t array_length =
+      std::stoull(array_type.size().value().as_string(), nullptr, 2);
```
lucasccordeiro: The byte-size and type-id-hash computations duplicate converter_binop.cpp:890-897 and 878-881 verbatim. Extract shared helpers.

src/c2goto/library/python/scalar.c
```diff
@@ -0,0 +1,66 @@
...
+  return memcmp(a, b, size) == 0;
+}
+
+// Copies `size` bytes from `value` into non-expiring storage, so a
+// PyObject's `.value` pointer stays valid after the assigning branch's
+// locals go DEAD at the join.
+void *__python_scalar_tag_copy(const void *value, size_t size)
+{
+__ESBMC_HIDE:;
+  if (size == 0)
```
lucasccordeiro: The `size == 0` branch is unreachable — both generated call sites pass at least 1 byte (`char[1]` minimum, scalar widths ≥ 8 bits). Remove it.

regression/python/dynamic_type_scalar_tag_join/main.py
```diff
@@ -0,0 +1,7 @@
+cond = nondet_bool()
+if cond:
+    x = 1
+else:
+    x = "a"
+if not cond:
+    assert x == "a"
```
lucasccordeiro, line 7: No test exercises `!=` (NotEq) on a tagged scalar, bool/str divergence, or reassignment after the join. Add CORE tests for these paths.

PR #6543, opened 2026-07-30
[docs] Document the current state of LTL verification

lucasccordeiro:
Rewrites the LTL theory page against ESBMC 8.4.0, in the layout used by the
C/C++ and Python doc sections.

`--ltl` still works and its three regression tests pass, but investigating it
turned up three defects, now filed and documented here rather than fixed:
the monitor never observes updates to the globals its propositions read, so a
definitively violated property reports as *presumably true* (#6546); absent
prefix assertions are read as `LTL_GOOD` (#6547); and the verdict never reaches
the top-level result or exit code (#6548).

Also records the five supported temporal operators, the never-claim polarity,
where the implementation diverges from the SoSyM paper, and replaces the page's
LaTeX with Unicode — the site has no `math: true`, so it was rendering as
literal `$\bot$`.

### DIFF
--- a/website/content/docs/theory/LTL.md
+++ b/website/content/docs/theory/LTL.md
@@ -14,13 +14,16 @@ rather than true/false.
 This page documents what the `--ltl` path does in ESBMC 8.4.0, including where
 the implementation diverges from [1].
 
-> **Note**: LTL is a legacy, lightly maintained path. It works — the shipped
-> regression tests pass and the four-valued verdict is computed correctly — but
-> it is wired into ESBMC differently from every other property class. Most
-> importantly, **`VERIFICATION SUCCESSFUL` and the exit code do not reflect the
-> LTL verdict**; you must read the `Final lowest outcome:` line. See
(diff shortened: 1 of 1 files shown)


lucasccordeiro:
Follow-up commit adds the operator reference and a soundness finding that turned up while building reproducers for the issues.

While generating monitors with a locally built libltl2ba I found that the monitor never observes updates to the globals its propositions read, so a *definitively* violated property is reported as `LTL_SUCCEEDING` (⊤ᵖ). That is a missed violation, filed as #6546. The other two documented gaps are #6547 (absent prefix assertions reported as `LTL_GOOD`) and #6548 (verdict absent from the top-level result and exit code).

The page now leads with a warning naming all three, and records: all five temporal operators (`G` `F` `X` `U` `V`) with the `X` caveat, the never-claim polarity (feed `!(φ)`; verified dual in both directions on six formulas), and libltl2ba's `-H` flag for the extern header.

Issue #6545, opened 2026-07-30
Multiplying an address-derived integer loses object identity


lucasccordeiro:
An address round-tripped through `uintptr_t` arithmetic that recovers the
original address exactly is rejected when the arithmetic includes a
multiplication.

```c
#include <stdint.h>
#include <assert.h>

int main(void)
{
  struct S { int x; int y; int z; } s = {.x = 1};
  uintptr_t u = (uintptr_t)&s;
  u *= 2;
  u -= (uintptr_t)&s;   /* u == (uintptr_t)&s again */
  int *p = (int *)u;
  *p = 3;
  assert(s.x == 3);
}
```

```
$ esbmc main.c
VERIFICATION FAILED     <-- expected SUCCESSFUL
```

This is the noisy direction -- a spurious counterexample, not a missed bug.

### The boundary is one operation

| address-derived integer expression | verdict | |
|---|---|---|
| `u = (uintptr_t)&s` | SUCCESSFUL | correct |
| `u += 4; u -= 4` | SUCCESSFUL | additive round-trip tracked |
| `u = u * 1` | SUCCESSFUL | folds away |
| `o = 4; o *= 2; p = (int *)(o + v)` | SUCCESSFUL | multiplying a *pure integer* is fine |
| **`u *= 2; u -= (uintptr_t)&s`** | **FAILED** | **wrong** |

So only a genuine multiplication of an *address-derived* term defeats object
recovery.

### `offsetof` counts as address-derived

`offsetof(struct S, y)` expands to `(size_t)&((struct S *)0)->y`, so multiplying
it takes the same path as multiplying an address:

```c
uintptr_t u = offsetof(struct S, y); u *= 2;  /* -> FAILED */
uintptr_t u = 4;                     u *= 2;  /* -> SUCCESSFUL */
```

The two programs are otherwise identical. This is why the existing
`github_426_2` fails even though it only ever multiplies an offset.

### Already-known symptoms

Three existing KNOWNBUGs share this single cause:

- `regression/esbmc/github_426_2` — multiplies an `offsetof`
- `regression/esbmc/github_426_3` — multiplies an address
- `regression/esbmc/github_426_4` — multiplies an address

### On disposition

Recovering object identity through a multiplicative term may not be worth the
model complexity. If it is not, the useful outcome is to state it as a known
limitation, so those three KNOWNBUGs stop reading as open defects. Filing it so
that decision is recorded either way.

### Regression tests

[...]

PR #6560, opened 2026-07-30
[test] Pin the address-derived-integer round-trip boundary

lucasccordeiro:
Adds the three cases that bound #6545: an additive round-trip through
`uintptr_t` (CORE, works), a multiplicative one (KNOWNBUG), and
`ptr + offsetof()` (KNOWNBUG).

The third corrects the record on `github_426_2`. The multiplication there is
incidental: `ptr + offsetof()` fails with no multiplication at all, while
multiplying a genuinely nondeterministic integer works. Folding `offsetof` to a
constant fixes both, but turns six `github_2512_*` `container_of` tests into
false alarms, so the modelling is left alone.

No behavioural change. Refs #6545.


lucasccordeiro:
Superseded by #6541, which landed both `ptr_int_additive_roundtrip` and `ptr_int_multiply_roundtrip`. Both `test.desc` files are byte-identical to this branch's, and the `main.c` files differ only in that master carries the explanatory comments and this branch a trailing `return 0;`, so there is nothing left to merge.

PR #6562, opened 2026-07-30
[ltl] Map the LTL verdict onto the verdict and exit code

### LINKED ISSUE #6548: [ltl] Verdict is absent from the top-level result and exit code
## Summary

The `--ltl` verdict reaches only a log line. `bmct::ltl_run_thread`'s result is
counted into `ltl_results_seen` and the LTL branch then returns
`P_UNSATISFIABLE` unconditionally (`src/esbmc/bmc.cpp:1995`), so every run —
including a definitive violation — ends in `VERIFICATION SUCCESSFUL` with exit
code `0`. No counterexample is emitted either.

There is consequently no way to consume an LTL result from a script, CI job or
the regression harness other than by grepping stdout for
`Final lowest outcome:`. The three shipped `regression/ltl/` tests all assert
`^VERIFICATION SUCCESSFUL$`, which holds regardless of the verdict.

## Reproducer

Any failing LTL run shows the mismatch; the sharpest case is a ⊥ verdict:

```
Final lowest outcome: LTL_BAD

VERIFICATION SUCCESSFUL
```

```sh
$ esbmc program.c --ltl notphi.c -DLTL_PREFIX_BOUND=10 >/dev/null 2>&1; echo $?
0
```

On the shipped tests, `regression/ltl/basic` reports `LTL_FAILING` and
`regression/ltl/basic-success` reports `LTL_SUCCEEDING`; both exit `0` with
`VERIFICATION SUCCESSFUL`.

No `[Counterexample]` block is produced for any of the four outcomes, so a
failing run gives no trace to inspect.

## Workaround

The instrumentation is driven by the `__ESBMC_property_*` markers in the monitor
file rather than by `--ltl`
(`src/esbmc/parseoptions/process_goto_program.cpp:458`), so passing the monitor
*without* `--ltl` checks the three prefix assertions as ordinary claims and
restores a non-zero exit code, per-claim results and counterexamples:

```sh
esbmc program.c notphi.c -DLTL_PREFIX_BOUND=10 --multi-property
```

```
✓ PASSED: 'LTL_BAD ...'
✗ FAILED: 'LTL_FAILING ...'
✗ FAILED: 'LTL_SUCCEEDING ...'

VERIFICATION FAILED
```

The lowest failing claim in the order `LTL_BAD` → `LTL_FAILING` →
`LTL_SUCCEEDING` is the lattice verdict; all three passing means ⊤. This
reproduces the `--ltl` verdict on each of the three shipped tests.

## Suggested direction

Map the outcome onto the process result — ⊥ and ⊥ᵖ as a failure, ⊤ᵖ and ⊤ as a
success — or add an explicit flag for it, and emit the counterexample for the
satisfying trace `ltl_run_thread` already obtained.

## Environment

ESBMC 8.4.0 (`31aee3387a`), Bitwuzla 0.9.0, Linux x86_64.

lucasccordeiro:
Stacked on #6556 — review that first; this PR's diff is the second commit.

The LTL branch returned `P_UNSATISFIABLE` unconditionally, so every run —
including a definitive violation — printed `VERIFICATION SUCCESSFUL` and exited
`0`.

`report_ltl_outcome` now maps the lattice onto the process result: ⊥/⊥ᵖ are
violations (exit 1), ⊤ᵖ/⊤ hold (exit 0), and the off-lattice `LTL_UNKNOWN`
reports `VERIFICATION UNKNOWN` (exit 2). `report_trace` bails out early for
`--ltl`, which otherwise dereferences a null `runtime_solver`.

The counterexample half of the issue is not addressed; the doc records why.

Fixes #6548.

### DIFF
--- a/regression/ltl/basic-func/test.desc
+++ b/regression/ltl/basic-func/test.desc
@@ -1,5 +1,5 @@
 CORE
 test.c
 --ltl test.ba-2.c -DLTL_PREFIX_BOUND=10
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION FAILED$
 ^Final lowest outcome: LTL_FAILING$
--- a/regression/ltl/basic-no-monitor/test.desc
+++ b/regression/ltl/basic-no-monitor/test.desc
@@ -1,5 +1,5 @@
 CORE
 test.c
 --ltl
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION UNKNOWN$
 ^Final lowest outcome: LTL_UNKNOWN$
--- a/regression/ltl/basic-truncated-unwind/test.desc
+++ b/regression/ltl/basic-truncated-unwind/test.desc
@@ -1,6 +1,6 @@
 CORE
 test.c
 --ltl test.ba-2.c -DLTL_PREFIX_BOUND=10 --unwind 1
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION UNKNOWN$
 ^Final lowest outcome: LTL_UNKNOWN$
 ^WARNING: LTL monitor preconditions violated
--- a/regression/ltl/basic/test.desc
+++ b/regression/ltl/basic/test.desc
@@ -1,5 +1,5 @@
 CORE
 test.c
 --ltl test.ba-2.c -DLTL_PREFIX_BOUND=10
-^VERIFICATION SUCCESSFUL$
+^VERIFICATION FAILED$
 ^Final lowest outcome: LTL_FAILING$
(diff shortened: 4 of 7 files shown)


lucasccordeiro:
Superseded by #6553, which landed the LTL verdict mapping on 30 Jul. master already sets `res = P_SATISFIABLE` for the two violating lattice values, routes `ltl_uninstrumented` to `report_unknown()`, and keeps the satisfying solver in `runtime_solver` so an LTL violation still prints a counterexample. This branch's five-value lattice would replace that sentinel design and early-return from `report_trace` for LTL, discarding the counterexample master now preserves. The remaining absent-prefix/truncated-monitor work continues in #6556.

PR #6587, opened 2026-07-31
[c++] Construct and destroy the elements of new T[n]

### LINKED ISSUE #6584: [c++] new T[n] runs no element constructors and delete[] no destructors
## Summary

`new T[n]` runs **no** element constructors, and `delete[]` runs no destructors. Members initialised by `T`'s constructor read back as nondeterministic, so well-formed code gets a spurious violation.

```cpp
#include <cassert>
struct T { int v; T() : v(7) {} };
int main() {
  T *p = new T[3];
  assert(p[0].v == 7);   // VERIFICATION FAILED
  delete[] p;
  return 0;
}
```

```
$ esbmc main.cpp --std c++20 --incremental-bmc
Violated property:
  file main.cpp line 5 column 3 function main
  assertion main
VERIFICATION FAILED

$ clang++ -std=c++20 -fsanitize=address,undefined main.cpp && ./a.out   # clean, exit 0
```

The scalar form is a working control — `T *p = new T; assert(p->v == 7);` verifies, so this is specific to the array form.

## Exactly zero elements are constructed

Not "only element 0" — none:

```cpp
int ctor = 0;
struct T { T() { ctor++; } ~T() {} };
int main() { T *p = new T[4]; assert(ctor == 0); delete[] p; return 0; }  // SUCCESSFUL
```

`assert(ctor == 1)`, `== 2` and `== 4` all report FAILED, so the count really is 0. The destructor side is the same: with `~T() { dtor++; }`, `delete[] p` leaves `dtor == 0`.

## Root cause

`goto_symext::symex_cpp_new` (`src/goto-symex/builtin_functions/cpp_memory.cpp`) allocates the dynamic object, assigns the pointer and registers it with `track_new_pointer` / `dynamic_memory`. It never emits a constructor call, and it does not look at the initializer that `clang_cpp_convertert` attaches to the side effect (`clang_cpp_convert.cpp:980-989`, `new_expr.initializer(init)`). For the scalar case the constructor is lowered elsewhere, which is why `new T` works.

So the array form allocates correctly-sized, correctly-typed storage whose elements are never constructed.

## Severity

Because construction and destruction are *both* absent, this is symmetric — it does not produce the invalid-free class of false positive that an asymmetric version would. The damage is that constructor side effects are lost:

- members initialised in the constructor read as nondet, giving spurious violations (above);
- constructors with side effects (counters, resource acquisition, registration) do not happen, which can equally *hide* a real bug;
- `delete[]` skipping destructors loses the release side.

## Relationship to #6574

Same family — construction of array elements not being emitted — but a different site and a different shape. #6574 is class-typed *member* arrays, where element 0 was constructed and the rest were not, leaving destruction unbalanced. Here it is `new T[n]`, where nothing is constructed and nothing destroyed. The fixes are independent; this one needs the initializer to be lowered into per-element construction over the allocated object, with the matching teardown on `delete[]`.

## Suggested regression coverage

A positive/negative pair on constructor-initialised members through `new T[n]`, plus a constructor/destructor counting case so the balance is pinned rather than just the member value.

lucasccordeiro:
Fixes #6584.

## Root cause

Both arms of the array form were empty stubs. `goto_convertt::cpp_new_initializer` (`src/goto-programs/builtin_functions.cpp`) and `goto_convertt::convert_cpp_delete` (`src/goto-programs/goto_convert.cpp`) each contained only

```cpp
    if (rhs.statement() == "cpp_new[]")
    {
      // build loop
    }
```

so the initializer was dropped at goto-convert time and no constructor or destructor ever reached the GOTO program. Members initialised by `T`'s constructor read back nondeterministically, giving a spurious violation on correct code:

```cpp
struct T { int v; T() : v(7) {} };
T *p = new T[3];
assert(p[0].v == 7);   // VERIFICATION FAILED; clean under clang -fsanitize=address,undefined
```

Scalar `new T` was unaffected — it has its own arm — which is what made the array form look like a modelling choice rather than a gap.

## Solution

**Construction.** The initializer the frontend supplies for the array form is shaped for the whole array: a `temporary_object` of type `T[n]` wrapping the element constructor, whose `this` is `&new_object[0]`. Two things follow:

- The else-branch wrap that turns an initializer into `new_object = <expr>` uses `rhs.type().subtype()` and is scalar-shaped; applied to the array form it assigns a `T*` into a `T`. It is skipped for `cpp_new[]`.
- Nothing in that initializer carries the `constructor` flag that the decl path's helper (`find_constructor_call` in `clang_cpp_adjust_code.cpp`) keys on. The constructor is located **by shape** instead — a `function_call` side effect with a function operand and a non-empty argument list — then lifted out and its `this` re-pointed at each element inside a loop bounded by the element count.

`this` is built as pointer arithmetic `lhs + i`, not `&lhs[i]`: `dereference_expr_nonscalar` recurses into an index's `source_value`, and a pointer source is scalar, so the index form trips its "no sudden transition back to scalars" assertion. User-written `p[i]` only survives because the adjuster rewrites it, and this runs after adjust.

A loop rather than unrolling, because the element count need not be a compile-time constant.

**Destruction.** `convert_cpp_delete` emits the matching teardown, recovering the count from `dynamic_size` (the allocation's byte size) divided by the element size, and reusing the scalar arm's null guard per [expr.delete]/7.

The two halves must land together. Constructing without destroying is worse than leaving both stubs empty: `new R[3]; delete[] a;` where `R` acquires memory in its constructor then reports a spurious `forgotten memory` under `--memory-leak-check`, which master (symmetric, both empty) does not.

**No constructor, no loop.** `new int[n]` has no constructor to run, and the array form has always left such elements nondeterministic. Emitting the scalar arm's zero-fill per element would both change that behaviour and, for a symbolic `n`, cost far more than it buys — it regressed `cwe_excessive_alloc_new_pass` to `UNKNOWN` and timed out two other tests before being removed.

## Tests

- `github_6584` — constructor and destructor counts plus element values through `new T[4]` / `delete[]`. `VERIFICATION FAILED` on master, `SUCCESSFUL` here.
- `github_6584_fail` — negative twin; a claim that a later element is unset must be reported.
- `github_6584_dtor` — `new R[3]` / `delete[]` under `--memory-leak-check`. This one passes on master too, since master runs neither constructor nor destructor; it is a pairing guard against the asymmetry above rather than a reproducer of the reported bug.

All three agree with host `clang++ -std=c++20`.

## Validation

`esbmc-cpp` tests are `THOROUGH`, which ctest skips on macOS, so suites were run directly against separately-built master and patched binaries.

**816 tests, zero differences**: `cpp` 688, `inheritance` 85, `template` 29, `destructors` 14.

## Limitation

A class with no data members models as zero bytes here, while C++ gives it a nonzero size ([class]/4). Its allocation is therefore zero bytes and the recovered element count is zero, so destructors do not run for `new Empty[n]`. Constructors do, which is already an improvement on master — and since such a class owns nothing, this cannot produce a leak false positive. Fixing it properly means modelling `sizeof` of an empty class as at least 1, which changes allocation sizes and belongs in its own change.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_6584/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6584/main.cpp
@@ -1,0 +1,31 @@
+// new T[n] must run one constructor per element, and delete[] one destructor
+// per element. Both arms of the lowering were empty stubs, so neither ran and
+// members initialised by T's constructor read back nondeterministically.
+#include <cassert>
+
+int ctor = 0, dtor = 0;
+
+struct T
+{
+  int v;
+  T()
+  {
+    ctor++;
+    v = 7;
+  }
+  ~T()
+  {
+    dtor++;
+  }
+};
+
+int main()
+{
+  T *p = new T[4];
+  assert(ctor == 4);
+  assert(p[0].v == 7);
+  assert(p[3].v == 7);
+  delete[] p;
+  assert(dtor == 4);
+  return 0;
+}
--- a/regression/esbmc-cpp/cpp/github_6584/test.desc
+++ b/regression/esbmc-cpp/cpp/github_6584/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--std=c++20 --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/github_6584_dtor/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6584_dtor/main.cpp
@@ -1,0 +1,23 @@
+// Construction without matching destruction would report a spurious leak, so
+// pin that delete[] releases what the element constructors acquired.
+#include <cstdlib>
+
+struct R
+{
+  int *p;
+  R()
+  {
+    p = (int *)malloc(sizeof(int));
+  }
+  ~R()
+  {
+    free(p);
+  }
+};
+
+int main()
+{
+  R *a = new R[3];
+  delete[] a;
+  return 0;
+}
--- a/regression/esbmc-cpp/cpp/github_6584_dtor/test.desc
+++ b/regression/esbmc-cpp/cpp/github_6584_dtor/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--std=c++20 --memory-leak-check --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc-cpp/cpp/github_6584_fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_6584_fail/main.cpp
@@ -1,0 +1,19 @@
+// Negative twin: the constructor really runs for every element, so a claim
+// that a later element is unset must be reported rather than passing.
+#include <cassert>
+
+struct T
+{
+  int v;
+  T() : v(7)
+  {
+  }
+};
+
+int main()
+{
+  T *p = new T[3];
+  assert(p[2].v != 7);
+  delete[] p;
+  return 0;
+}
--- a/regression/esbmc-cpp/cpp/github_6584_fail/test.desc
+++ b/regression/esbmc-cpp/cpp/github_6584_fail/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.cpp
+--std=c++20 --incremental-bmc
+^VERIFICATION FAILED$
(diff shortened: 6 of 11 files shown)


lucasccordeiro:
Clarifying one line in the description, so it is not read as a stronger claim than it is.

Where I wrote that "the array form has always left such elements nondeterministic", that is correct for `new int[n]`, which default-initialises — a scalar element is genuinely indeterminate per [dcl.init]/7, and this PR keeps it that way deliberately.

It is *not* correct for the parenthesised `new int[n]()`, which value-initialises and must zero. ESBMC does not do that today. That gap is pre-existing and untouched here — verified `VERIFICATION FAILED` on a pre-#6584 build as well — so this PR neither causes nor fixes it. Filed separately as #6588 rather than widened into this change, since it needs to key on `CXXNewExpr::hasInitializer()` to tell the two forms apart, and the cost caveat about per-element loops over a symbolic count applies there too.

PR #6605, opened 2026-07-31
[goto-symex] Solve global reachability as a fixpoint, not a BFS

### LINKED ISSUE #6594: [symex] spurious forgotten-memory on a concurrently extended list
## Summary

With `--memory-leak-check --no-reachable-memory-leak`, a heap node that is still reachable from a global array of list heads is reported as `forgotten memory` when

1. it is inserted at a **nondeterministically chosen** head (`slot[j]` with `j` nondet), **and**
2. more than one such insertion happens, **and**
3. another thread also inserts into the same array.

Remove any one of the three and the leak disappears. The memory is reachable from a global in every execution, so `--no-reachable-memory-leak` should suppress it.

## Reproducer

```c
#include <pthread.h>
#include <stdlib.h>
extern void abort(void);
static void assume(int c){ if(!c) abort(); }

struct s { struct s *next; };
struct s *slot[2];

static struct s *new_node(void){ struct s *p = malloc(sizeof(struct s)); p->next = 0; return p; }
static void list_add(struct s *n, struct s *l){ struct s *t = l->next; l->next = n; n->next = t; }

static void *t_fun(void *a){ (void)a; int i = 0; list_add(new_node(), slot[i]); return 0; }

int main(void){
  for (int q = 0; q < 2; q++) slot[q] = new_node();
  int j = __VERIFIER_nondet_int(), k = __VERIFIER_nondet_int();
  assume(0 <= j && j < 2); assume(0 <= k && k < 2);
  list_add(new_node(), slot[j]);
  list_add(new_node(), slot[k]);
  pthread_t t; pthread_create(&t, 0, t_fun, 0);
  return 0;
}
```

```
esbmc s_ndjk.c --32 --sv-comp --no-div-by-zero-check --force-malloc-success \
  --force-realloc-success --state-hashing --add-symex-value-sets \
  --smt-symex-guard --bitwuzla --cswitch-skip-readonly-globals \
  --memory-leak-check --no-reachable-memory-leak --no-assertions \
  --no-abnormal-memory-leak --malloc-zero-is-null --incremental-bmc
```

```
Violated property:
  dereference failure: forgotten memory: dynamic_9_value
  CWE: CWE-401
VERIFICATION FAILED
```

Every node is on some `slot[q]` list at exit, so nothing is unreachable.

## What changes the verdict

| Variant | Verdict |
|---|---|
| as above (`j`, `k` nondet; thread inserts) | **FAILED** — `forgotten memory` |
| `j`, `k` nondet; thread's index also nondet | **FAILED** — `forgotten memory` |
| `int j = 0, k = 1;` (constant heads), thread index nondet | SUCCESSFUL |
| `int j = 0, k = 1;` and thread index constant | SUCCESSFUL |
| only one insert (`list_add(..., slot[k])` removed) | SUCCESSFUL |
| no `pthread_create` (sequential) | SUCCESSFUL |

So neither the nondeterministic index alone nor the concurrency alone is sufficient — the report needs two nondeterministically-headed insertions plus a concurrent inserter.

## Where this likely goes wrong

`src/goto-symex/symex_main.cpp`, the globals-reachability closure used by `--no-reachable-memory-leak` (~lines 1515-1560). Each reachable symbol is expanded exactly once, guarded by a `visited` set:

```cpp
    std::unordered_set<std::string> visited;
    for (int i = 0; !globals.empty(); i++)
    {
      ...
      /* A symbol can be reached at this frontier level through several pointers
       * with different preconditions (e.g. a node inserted into a list chosen
       * by a non-deterministic index is, in the value-set, reachable from every
       * list head). Merge those entries by symbol, OR-ing their path conditions,
       * before expanding each symbol once. ... */
```

The merge that this comment describes (added for #2335) ORs the path conditions of entries found **at the same frontier level**. The `visited` cull, however, is global across levels: a node that is reachable at depth 1 on one path and at depth 2 on another is expanded only once, under the guard of whichever level reached it first. Its outgoing edges then carry a guard that is false on the executions that reach it via the other depth, which under-approximates reachability in exactly the direction that produces spurious `forgotten memory`.

Two nondeterministically-headed insertions are what make one node's depth path-dependent; the concurrent inserter adds the interleaving in which the depths differ.

## Impact

This is the root cause of one of the nine incorrect-false results in the [SV-COMP 26 benchexec run 30530746600](https://github.com/esbmc/esbmc/actions/runs/30530746600):

- `c/goblint-regression/09-regions_13-arraycollapse_nr.c` (`valid-memsafety`, expected `true`, ESBMC reports `false(valid-memtrack)`)

That benchmark is the same shape: `struct s *slot[10]`, lists built through `list_add`, two insertions at nondeterministic indices `j` and `k`, and a thread inserting at its own nondeterministic index.

Reproduced on master `68d2af5e49`, ESBMC 8.4.0.

lucasccordeiro:
The closure behind `--no-reachable-memory-leak` expanded each symbol once,
under the path condition of whichever frontier level reached it first, so a
symbol reachable at two depths gave its outgoing edges a condition that is
false on the executions taking the longer route. Phase 1 now records only
local edge conditions and phase 2 solves them as a least fixpoint.

Fixes #6594

### DIFF
--- a/regression/esbmc/github_6594/main.c
+++ b/regression/esbmc/github_6594/main.c
@@ -1,0 +1,59 @@
+#include <pthread.h>
+#include <stdlib.h>
+
+extern int __VERIFIER_nondet_int(void);
+extern void abort(void);
+
+static void assume(int c)
+{
+  if (!c)
+    abort();
+}
+
+struct s
+{
+  struct s *next;
+};
+
+struct s *slot[2];
+
+static struct s *new_node(void)
+{
+  struct s *p = malloc(sizeof(struct s));
+  p->next = 0;
+  return p;
+}
+
+static void list_add(struct s *n, struct s *l)
+{
+  struct s *t = l->next;
+  l->next = n;
+  n->next = t;
+}
+
+static void *t_fun(void *a)
+{
+  (void)a;
+  list_add(new_node(), slot[0]);
+  return 0;
+}
+
+/* Every node ends up on one of the slot[] lists, so nothing is forgotten. The
+ * concurrent insertion can push a node two links away from its head, which is
+ * what used to defeat the globals-reachability closure. */
+int main(void)
+{
+  for (int q = 0; q < 2; q++)
+    slot[q] = new_node();
+
+  int j = __VERIFIER_nondet_int(), k = __VERIFIER_nondet_int();
+  assume(0 <= j && j < 2);
+  assume(0 <= k && k < 2);
+
+  list_add(new_node(), slot[j]);
+  list_add(new_node(), slot[k]);
+
+  pthread_t t;
+  pthread_create(&t, 0, t_fun, 0);
+  return 0;
+}
--- a/regression/esbmc/github_6594/test.desc
+++ b/regression/esbmc/github_6594/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--32 --no-div-by-zero-check --force-malloc-success --state-hashing --add-symex-value-sets --memory-leak-check --no-reachable-memory-leak --no-assertions --no-abnormal-memory-leak --malloc-zero-is-null --incremental-bmc
+^VERIFICATION SUCCESSFUL$
--- a/regression/esbmc/github_6594_fail/main.c
+++ b/regression/esbmc/github_6594_fail/main.c
@@ -1,0 +1,63 @@
+#include <pthread.h>
+#include <stdlib.h>
+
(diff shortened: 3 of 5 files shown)


lucasccordeiro:
This PR (100s): https://github.com/esbmc/esbmc/actions/runs/30680929877
Master (100s): https://github.com/esbmc/esbmc/actions/runs/30680908491

PR #6616, opened 2026-08-01
[build] re-derive the version ID on every build

### LINKED ISSUE #6615: [build] --git-hash reports the configure-time commit, not the compiled one
## Summary

`--git-hash` reports the commit the build tree was **configured** at, not the commit that was **compiled**. The value is written once and then never refreshed, so it silently goes stale as soon as you pull and rebuild in an existing build directory.

## Reproduction

On a build directory that has been in use for a while:

```console
$ ls -la build/src/esbmc/esbmc
-rwxr-xr-x 1 user user 490885912 Jul 30 15:58 build/src/esbmc/esbmc

$ git log -1 --format='%H %ad' --date=short
2b84f2a8ec08047093d10659e1d1caf969b4f6e5 2026-07-30

$ ./build/src/esbmc/esbmc --git-hash
ESBMC built from  ef23805e29009084af1496c3f805d665bb4077b8 2026-04-20 14:10:37 by ...
```

The binary was linked on 2026-07-30 from `2b84f2a8ec`, but reports `ef23805e29` from 2026-04-17. That commit is not merely older, it is **not an ancestor of the commit that was built**:

```console
$ git merge-base --is-ancestor ef23805e29 HEAD && echo ancestor || echo "not an ancestor"
not an ancestor
```

Running the generator by hand produces the correct value, so the script is fine:

```console
$ python3 scripts/buildidobj.py /tmp/id.txt && cat /tmp/id.txt
ESBMC built from  2b84f2a8ec08047093d10659e1d1caf969b4f6e5 2026-08-01 ...
```

## Cause

`src/esbmc/CMakeLists.txt`:

```cmake
add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
  COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/scripts/buildidobj.py ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
  ...
)
```

The rule declares no `DEPENDS` and nothing marks it always-out-of-date, so once `buildidobj.txt` exists ninja considers it up to date forever. On the affected tree the file's mtime was 2026-04-20 while the binary's was 2026-07-30.

## Impact

Neither `--version` nor `--git-hash` currently identifies a build:

- `--version` prints `8.4.0` for the `v8.4` tag and for any master build alike.
- `--git-hash` prints whatever commit the build directory was configured at.

That makes both unusable as provenance, which matters for benchmarking and for reproducing reported results. It has already caused a concrete problem: a benchmark suite recorded `--git-hash` in its results files on the documented understanding that it "embeds the exact commit", and the recorded sha predates a contract fix by one day, so the records cannot establish whether the binary that produced the numbers contained the fix under investigation. `--git-hash` is arguably worse than `--version` here, because a 40-character sha and a microsecond timestamp read as authoritative.

## Suggested fix

Give the rule a second output that is never created, so it stays out of date and re-runs on every build. Done naively this would rewrite `buildidobj.txt` every time and relink the binary on every build, so pair it with a guard in `scripts/buildidobj.py` that leaves the file untouched when only the timestamp would differ.

PR to follow.

LukeW1999:
Fixes #6615.

## What

`--git-hash` reported the commit the build tree was **configured** at rather than the one compiled in, because the rule generating the version ID declared no `DEPENDS` and its output already existed, so ninja ran it once and never again.

On the tree this was found on, a binary linked 2026-07-30 from `2b84f2a8ec` reported `ef23805e29` dated 2026-04-20, a commit that is not even an ancestor of the one built.

## How

- `buildidobj-rerun` is added as a second output and is never created, so the rule stays permanently out of date and re-runs every build.
- Rewriting the file every build would relink the whole binary every build, since `buildidobj.c` depends on it. So `buildidobj.py` now compares against the existing file with the timestamp removed, and returns without writing when only the timestamp would differ. The timestamp's meaning becomes "when the version ID last changed", which is also more useful than "when it was last relinked".
- The rule now also declares `DEPENDS` on `buildidobj.py` itself.

The emitted string format is unchanged.

## Testing

Verified in an isolated CMake project with the same generator that the sentinel output does force a re-run on every build, and that a downstream rule consuming the file does **not** re-run while the content is unchanged (this is what keeps the relink cost at zero).

Against this repository:

| check | result |
| --- | --- |
| generated ninja edge | `buildidobj.txt` and `buildidobj-rerun` both declared as outputs, `DEPENDS` on the script |
| output format vs `master` | byte-identical once the timestamp is removed |
| refreshing the stale file in-place | `ef23805e29 2026-04-20` becomes `f0979a428a 2026-08-01 (dirty tree)` |
| repeated runs, nothing changed | file untouched, mtime unchanged, so no relink |
| ID changed under it | rewritten with the correct sha |
| `yapf --style .style.yapf` | clean |

No test is added: this is build-system behaviour, `regression/`'s `test.desc` format does not apply to it, and `unit/` currently hosts only the Python frontend's model tests. Happy to add coverage if you would like it somewhere.

### DIFF
--- a/scripts/buildidobj.py
+++ b/scripts/buildidobj.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import sys, os
+import re  # For comparing build IDs
 import shutil  # For git
 import subprocess  # For git
 from datetime import datetime  # For date
@@ -74,17 +75,41 @@ class BuildObj:
         return True
 
     @staticmethod
+    def format_id() -> str:
+        line = (f'{BuildObj.STR_ESBMC_BUILT_FROM} {BuildObj.get_last_hash()} '
+                f'{BuildObj.get_datetime()} {BuildObj.STR_BY} '
+                f'{BuildObj.get_username()}{BuildObj.STR_AT}'
+                f'{BuildObj.get_hostname()}')
+        if BuildObj.is_dirty_tree():
+            line += f' {BuildObj.STR_DIRTY}'
+        return line
+
+    @staticmethod
+    def without_datetime(line: str) -> str:
+        return re.sub(r'\d{4}-\d{2}-\d{2} [\d:.]+ ', '', line)
+
+    @staticmethod
+    def describes_same_build(output, line) -> bool:
+        """The build system re-runs this on every build so the ID cannot go
+        stale. Rewriting unconditionally would then relink the whole binary
+        every time, so report when the only difference is the timestamp."""
+        # errors='replace' rather than a narrower except: a file written by an
+        # older revision carries the locale encoding, and a decode failure here
+        # would abort the build. Mangled text simply reads as a different ID.
+        try:
+            with open(output, encoding='utf-8', errors='replace') as f:
+                old = f.read()
+        except OSError:
+            return False
+        return BuildObj.without_datetime(old) == BuildObj.without_datetime(line)
+
+    @staticmethod
     def run(output):
-        with open(output, 'w') as f:
-            f.write(f'{BuildObj.STR_ESBMC_BUILT_FROM} ')
-            f.write(f'{BuildObj.get_last_hash()} ')
-            f.write(f'{BuildObj.get_datetime()} ')
-            f.write(f'{BuildObj.STR_BY} ')
-            f.write(f'{BuildObj.get_username()}')
-            f.write(f'{BuildObj.STR_AT}')
-            f.write(f'{BuildObj.get_hostname()}')
-            if BuildObj.is_dirty_tree():
-                f.write(f' {BuildObj.STR_DIRTY}')
+        line = BuildObj.format_id()
+        if BuildObj.describes_same_build(output, line):
+            return
+        with open(output, 'w', encoding='utf-8') as f:
+            f.write(line)
 
 
 def main():
--- a/src/esbmc/CMakeLists.txt
+++ b/src/esbmc/CMakeLists.txt
@@ -27,8 +27,15 @@ if(ENABLE_JIMPLE_FRONTEND)
   set(JIMPLE_FRONTEND_TARGETS jimple)
 endif()
 
-add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
+# A custom target always runs, so the version ID is re-derived on every build
+# rather than once when the build tree is configured, which is what made
+# --git-hash report the configure-time commit instead of the compiled one.
+# buildidobj.py leaves the file untouched when only its timestamp would change,
+# so the flail rule below, and the relink after it, still only fire when the ID
+# actually changes.
+add_custom_target(esbmc_buildid
   COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/scripts/buildidobj.py ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
+  BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   COMMENT "Generating ESBMC version ID"
   VERBATIM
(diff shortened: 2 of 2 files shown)


src/esbmc/CMakeLists.txt
```diff
@@ -27,8 +27,16 @@ if(ENABLE_JIMPLE_FRONTEND)
...
 endif()
 
+# buildidobj-rerun is never created, which keeps this rule permanently out of
+# date so the version ID is re-derived on every build. Without it the rule runs
+# once when the build tree is first configured and never again, so --git-hash
+# reports the commit the tree was configured at rather than the one compiled.
+# buildidobj.py leaves the file untouched when only its timestamp would change,
+# so this does not relink the binary on every build.
 add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/buildidobj.txt
+                          ${CMAKE_CURRENT_BINARY_DIR}/buildidobj-rerun
```
lucasccordeiro: Under the Unix Makefiles generator this sentinel re-runs the flail rule and relinks on every build (verified locally; Ninja is fine). Restrict it to Ninja or use an alternative.

scripts/buildidobj.py
```diff
@@ -73,18 +74,39 @@ def is_dirty_tree() -> bool:
...
+
+    @staticmethod
+    def describes_same_build(output, line) -> bool:
+        """The build system re-runs this on every build so the ID cannot go
+        stale. Rewriting unconditionally would then relink the whole binary
+        every time, so report when the only difference is the timestamp."""
+        try:
+            with open(output, encoding='utf-8') as f:
+                old = f.read()
+        except OSError:
```
lucasccordeiro, line 102: `OSError` misses `UnicodeDecodeError` from a pre-existing non-UTF-8 `buildidobj.txt` (old code wrote with the locale encoding), aborting the build. Catch it too, or pass `errors='replace'`.

LukeW1999:
Thanks both — the Makefiles one was a real hole in my testing, so I have reworked it rather than patched it.

## @lucasccordeiro, Makefiles: confirmed, and worse than reported

Reproduced in a minimal project, and the sentinel does not merely relink, it defeats the guard outright:

```
Ninja:          build 2: GEN: unchanged        build 3: GEN: unchanged
Unix Makefiles: build 2: GEN: wrote            build 3: GEN: wrote
```

CMake's Makefile generator removes the outputs of a multi-output custom command before running it, so the script never sees an existing file to compare against and rewrites unconditionally. My original testing only covered Ninja, which is how this got through.

Rather than restricting to Ninja, which would leave Makefiles builds reporting a stale ID, this now uses a custom target, which always runs under both generators, with the file as a `BYPRODUCTS`. Verified on both, including the case that actually matters for correctness:

| | Ninja | Unix Makefiles |
| --- | --- | --- |
| regenerates on every build | yes | yes |
| no relink while the ID is unchanged | yes | yes |
| relinks **in the same build** when the ID changes | yes | yes |

The third row is the one worth pinning: if the target ran after the rule that consumes the file, a changed ID would only take effect on the following build, which fails silently.

In this repository, on Ninja:

```
build 1 (a source file changed)   [454/477] Generating ESBMC version ID ... Linking
build 2 (nothing changed)         [1/4] Generating ESBMC version ID          <- nothing else
build 3 (nothing changed)         [1/4] Generating ESBMC version ID          <- nothing else
```

and the binary now reports the commit it was actually built from.

## @lucasccordeiro, `UnicodeDecodeError`: right, and it was self-inflicted

`UnicodeDecodeError` derives from `ValueError`, not `OSError`, so it would have escaped and aborted the build. Confirmed against the version you reviewed:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf6 in position 46: invalid start byte
```

Taken your second suggestion, `errors='replace'`, in preference to widening the `except`: it cannot raise at all, and mangled text simply compares unequal, which regenerates the file. That is the correct outcome anyway, since a file in an unknown encoding should be replaced rather than trusted. A latin-1 file from an older revision now yields exit 0 and a correctly rewritten ID.

Worth noting the explicit `encoding=` was itself added to satisfy Codacy on this PR, so the fix for one review comment created the bug found by another. Both are in the same push.

## @copilot-pull-request-reviewer, the missing output does not break Ninja

> Generators like Ninja typically treat a missing declared output as a build error (or at least an always-failing rule)

That is not what happens. In a minimal project with the same rule shape, three consecutive builds each re-ran the command with no error and no failing edge, which is the documented behaviour of this idiom: Ninja re-runs an edge whose declared output is absent.

The suggestion to use an always-running target instead is the right one though, for the different reason above, and that is what this now does.

PR #6618, opened 2026-08-01
[esbmc] run on a thread with a large stack

### LINKED ISSUE #6617: [clang-frontend] Stack overflow converting long binary-operator chains
## Summary

Converting a long chain of binary operators overflows the stack and ESBMC dies with `SIGSEGV`. The clang frontend's expression converter recurses once per operand, so a chain of N terms costs N stack frames, and the process dies somewhere above 8,000 terms on a default 8 MB stack.

## Reproduction

13 lines, 266 bytes, no flags, no solver:

```c
#define A1 x && x && x && x
#define A2 A1 && A1 && A1 && A1
#define A3 A2 && A2 && A2 && A2
#define A4 A3 && A3 && A3 && A3
#define A5 A4 && A4 && A4 && A4
#define A6 A5 && A5 && A5 && A5
#define A7 A6 && A6 && A6 && A6

int main(void)
{
  int x = 1;
  return A7;
}
```

```console
$ esbmc macro_and.c
Parsing macro_and.c
Converting
$ echo $?
139
```

## It is the stack depth, not the input size

Same binary, same file, only `ulimit -s` differs:

| input | 8 MB stack (default) | 64 MB stack |
| --- | --- | --- |
| `A6`, 4,096 terms | exit 0 | exit 0 |
| `A7`, 16,384 terms | **SIGSEGV** | exit 0 |

Bisecting with a flat generated chain puts the boundary between 8,000 and 10,000 terms on this machine. Since it depends on `ulimit -s`, the compiler and inlining, the threshold is not a fixed property of ESBMC and moves between machines.

## It is not specific to `&&`

Substituting `+`, `|` or `,` for `&&` in the reproducer crashes identically, so this is the generic binary-operator path rather than anything about short-circuit lowering.

## It is ESBMC's converter, not the parser

```console
$ clang -fsyntax-only macro_and.c ; echo $?
0
$ gcc -fsyntax-only macro_and.c ; echo $?
0
```

ESBMC also prints `Parsing` before dying in `Converting`, and every backtrace frame is in `clang_c_convertert`.

## Backtrace

```
#0  std::__cxx11::basic_string<...>::basic_string
#1  clang_c_convertert::get_start_location_from_stmt   clang_c_convert.cpp:4779
#2  clang_c_convertert::get_expr                       clang_c_convert.cpp:2049
#3  clang_c_convertert::get_binary_operator_expr       clang_c_convert.cpp:4124
#4  clang_c_convertert::get_expr                       clang_c_convert.cpp:2593
#5  clang_c_convertert::get_binary_operator_expr       clang_c_convert.cpp:4124
#6  clang_c_convertert::get_expr                       clang_c_convert.cpp:2593
...  repeating to the bottom of the stack
```

`get_binary_operator_expr` calls `get_expr` on its LHS, which for a left-associative chain is another `BinaryOperator`, so the two recurse into each other once per term. Frame 0/1 is just where the stack happened to run out.

## Environment

ESBMC 8.4.0, local build from `2b84f2a8ec` (2026-07-30, on top of master), Linux x86_64, glibc 2.35, `RelWithDebInfo`. Independently observed on macOS with a 2026-07-24 build, so it is neither platform-specific nor recently introduced.

## Where this came up

A benchmark generator emitted a contract whose `requires`/`ensures` was a flat conjunction over K×N terms. At K=50, N=256 that is 12,800 terms and ESBMC crashed, which was initially and wrongly recorded as a tool scale limit. Deeply nested expressions are unusual in hand-written code but ordinary in generated code, which is exactly what a verifier tends to be fed.

## Possible fixes

1. **Raise the stack.** Cheapest; `src/esbmc/parseoptions/command_line_options.cpp` already has an unconditional `#ifndef _WIN32` block setting `RLIMIT_CORE` that this would sit next to. The caveat is that on macOS the main thread's stack cannot be grown this way, so it may only fix Linux.
2. **Run the conversion on a thread with an explicitly large stack** (`pthread_attr_setstacksize`), which is what clang and LLVM do for the same reason. Behaves the same on Linux and macOS, at the cost of moving the frontend call onto a thread.
3. **Make the conversion iterative** for same-operator chains. Correct in principle, but `get_expr` is a large dispatcher and this would mean an explicit worklist carrying partial state through it, so the risk looks disproportionate.
4. **A depth guard** that reports cleanly instead of crashing. Worth having regardless of which of the above is chosen, since any bound we pick will eventually be exceeded, and a diagnostic beats `SIGSEGV`.

I am picking this up and will start with 1, measured on both platforms, falling back to 2 if the macOS side does not hold.

LukeW1999:
Fixes #6617.

## What

`clang_c_convertert::get_expr` and `get_binary_operator_expr` recurse into each other once per operand, so a left-associative chain of N terms needs N stack frames. On the default 8 MB stack that dies somewhere above 9,000 terms:

```console
$ esbmc macro_and.c      # 13 lines, 12288 operands via nested macros
Parsing macro_and.c
Converting
```

The input is ordinary as far as compilers are concerned: `gcc -fsyntax-only` and `clang -fsyntax-only` both accept it, and ESBMC gets past `Parsing` before dying in `Converting`. It is not specific to `&&` either, since `+`, `|` and `,` reproduce identically.

Hand-written code rarely looks like this, but generated code does. This surfaced on a benchmark whose contract stated `requires`/`ensures` as a flat conjunction over K×N terms, which at K=50, N=256 is 12,800 operands.

## How

Run ESBMC on a thread created with a 512 MB stack. If the thread cannot be created the previous behaviour is used unchanged, so a restricted environment degrades rather than fails.

Two things are worth flagging, because both were found the hard way.

**Raising `RLIMIT_STACK` does not work.** That was this PR's first implementation. It fixes Linux, and on macOS `setrlimit` returns 0 and a subsequent `getrlimit` reports the larger soft limit, but the main thread's stack still does not grow and the process segfaults exactly as before. Verified on a `macos-26-arm64` runner:

```
soft=7MB hard=finite
setrlimit(63MB) -> 0 errno=0
soft now 63MB
Segmentation fault: 11
```

The same probe with a `pthread_attr_setstacksize(512MB)` thread survives 80 MB of recursion, so the thread is the portable mechanism. This is also what clang and LLVM do for the same reason.

**`esbmc_parseoptionst` has to be constructed on that thread**, not constructed by `main` and passed in. `language_uit`'s constructor initialises `migrate_namespace_lookup`, which is `thread_local`, so building it on the main thread leaves it null on the worker and the first `migrate_expr` dereferences null. That is not deep recursion, it crashes on a one-line input, so it is caught immediately by any test. The other `thread_local`s in the tree are `--k-induction-parallel` counters that legitimately start fresh.

An iterative conversion would be the narrower fix, but `get_expr` is a large dispatcher and threading an explicit worklist through it looks disproportionate. The thread also covers recursion elsewhere in ESBMC rather than this one path.

Windows keeps its current behaviour. Its default stack is smaller still, so it likely wants the same treatment, but that needs `_beginthreadex` and belongs in its own change.

## Testing

`regression/esbmc/deep_binary_chain_pass` and `..._fail`, both 12288 operands built from nested macros so the sources stay short.

The pair is deliberately two-sided. A crash-only test still passes if the conversion survives but builds the wrong expression, so `..._fail` asserts the chain over a nondet `x` and requires `VERIFICATION FAILED`, which only holds if the conjunction was converted with its real semantics.

| | verdict | time | on unfixed code |
| --- | --- | --- | --- |
| `deep_binary_chain_pass` | SUCCESSFUL | 17 s | SIGSEGV |
| `deep_binary_chain_fail` | FAILED | 22 s | SIGSEGV |

Both still pass with the process stack hard-pinned to 8 MB (`ulimit -Hs 8192`), which the `RLIMIT_STACK` version did not, confirming the thread's own stack is what carries them.

Regression runs, all on Linux x86_64:

| suite | result |
| --- | --- |
| `-L function_contract` | 314/314 |
| `-R k_induction` (spawns threads, so the highest-risk area here) | 199/199 |
| `-L esbmc/`, every sixth test | 267/267 |

The full 1600-test `esbmc/` suite does not fit the time budget I was working to, hence the stride; CI covers the rest. `clang-format-11` clean.

Sizing note: the crash boundary sits between 8,000 and 10,000 operands here and moves with `ulimit -s`, the compiler and inlining. 12288 keeps a margin without paying much for it, since conversion cost grows roughly quadratically with chain length (4,096 operands take ~3 s, 16,384 take ~50 s).

## Related

The macOS leg of CI runs only the `python/` suite, so neither these tests nor the original `RLIMIT_STACK` attempt would have been exercised there. Raised separately rather than bundled here.


src/esbmc/main.cpp
```diff
@@ -5,8 +5,60 @@
...
+#ifndef _WIN32
+#  include <pthread.h>
+
+/* Conversion and symbolic execution both recurse over expression structure, so
+ * generated inputs with deeply nested expressions exhaust the default 8MB main
+ * thread stack (#6617). A secondary thread can be given a much larger one,
+ * which is what clang and LLVM do for the same reason. Raising RLIMIT_STACK
+ * instead only works on Linux: macOS accepts the call and reports the new soft
+ * limit, but still does not let the main thread's stack grow. */
+static const size_t esbmc_stack_size = 512UL * 1024 * 1024;
```
lucasccordeiro, line 67: `--memlimit` sets `RLIMIT_DATA`, which on Linux also charges this 512MB stack mapping (verified: `malloc` fails under a 600MB limit). Add the stack size to the limit.

regression/esbmc/deep_binary_chain_pass/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
```
lucasccordeiro, line 1: The Windows PR job runs `ctest -L 'esbmc/|floats/'`, and the fix is `#ifndef _WIN32`, so both new CORE tests will SIGSEGV there. Gate them or cover Windows.

regression/esbmc/deep_binary_chain_pass/main.c
```diff
@@ -0,0 +1,19 @@
+/* 3 * 4^6 = 12288 operands. Conversion recurses once per operand, which used
+ * to overflow the stack before reaching symbolic execution, see #6617. */
+#define A1 x && x && x && x
+#define A2 A1 && A1 && A1 && A1
+#define A3 A2 && A2 && A2 && A2
+#define A4 A3 && A3 && A3 && A3
+#define A5 A4 && A4 && A4 && A4
+#define A6 A5 && A5 && A5 && A5
+#define DEEP A6 && A6 && A6
```
lucasccordeiro, line 9: This takes 34s here on an idle box, not 17s; CI runs `-j2`/`-j4` against a 120s per-test cap. Shorten the chain to keep margin.

LukeW1999:
@copilot-pull-request-reviewer this review is against `47312d9f2`; the branch has since been reworked and your point was the reason.

You were right that `setrlimit` would not carry macOS. I checked it directly on a `macos-26-arm64` runner rather than reasoning about it, and the failure is quieter than "it does not work": the call **succeeds**, and a following `getrlimit` confirms the larger soft limit, yet the stack still does not grow.

```
soft=7MB hard=finite
setrlimit(63MB) -> 0 errno=0
soft now 63MB
Segmentation fault: 11
```

A second probe confirmed the alternative you suggested: a thread created with `pthread_attr_setstacksize(512MB)` survives 80 MB of recursion on the same runner. The PR now does that, for the whole run rather than only the conversion, since the recursion is not unique to that phase.

One thing that is worth recording for anyone doing this again: `esbmc_parseoptionst` has to be **constructed** on the worker thread, not constructed by `main` and passed to it. `language_uit`'s constructor initialises `migrate_namespace_lookup`, which is `thread_local`, so building it on the main thread leaves it null on the worker and the first `migrate_expr` dereferences null. That is not a deep-recursion path, it crashes on a one-line input, so it shows up immediately in any test.

The macOS gap you implicitly flagged, that this would not have been "covered by CI/regression", is real and general: the macOS leg runs only the `python/` suite, so no `esbmc/` test has ever run there on any event. Raised as #6619 rather than bundled here.

LukeW1999:
CI on `69c48a0bd` turned up two failures. They are unrelated to each other, and only one was mine.

## Windows: mine, fixed

`deep_binary_chain_pass` and `deep_binary_chain_fail` both failed. The fix is `#ifndef _WIN32`, so Windows kept its old stack while the tests ran there unconditionally — and Windows defaults to 1MB, so it overflows sooner than anything else.

Rather than skipping the tests there, Windows now gets the same headroom at link time:

```cmake
if(MSVC)
  target_link_options(esbmc PRIVATE /STACK:536870912)
endif()
```

which is the reserve, committed on demand, so it costs address space rather than memory. I cannot test this locally, so I am relying on this run to confirm it.

## Ubuntu: `github_2335_1`, not mine

It failed on a 120s timeout rather than a wrong verdict. I A/B'd it locally against the same tree with only `main.cpp` reverted to master:

| | time | verdict |
| --- | --- | --- |
| with this PR | 165s | SUCCESSFUL |
| `main.cpp` from master | 174s | SUCCESSFUL |

So the change is if anything marginally faster, within noise, and the test is simply a slow one sitting above the 120s cap. It tipped over on a loaded runner. Not touching it here, but it is worth someone raising separately: `--unlimited-k-steps` with `--state-hashing` on that input is around 170s on my machine, so it will keep flapping.

Both new tests passed on the Ubuntu leg (~25s each).

LukeW1999:
Rebased onto master. Worth a note because the conflict with #6637 was not only textual.

CI had stopped running here entirely — one Codacy check and nothing else. The cause was that #6637 landed in `main()` while this branch was rewriting the same function, so the PR went `CONFLICTING`, GitHub could not build `refs/pull/6618/merge`, and the workflow never triggered. Not a queue delay, no run was ever created.

The semantic half matters more. #6637 put `main()`'s body in a try/catch so an escaping exception is reported rather than reaching `std::terminate`. This PR moves that body onto a worker thread, and **an exception escaping a worker thread does not pass through `main`'s handlers** — it goes straight to `std::terminate`. Merging the two naively would have silently undone #6637 for every run.

Resolved by factoring the work plus its handlers into `run_esbmc()`, called from the thread body and from the no-thread fallback, so the handlers travel with the work rather than staying behind on the main thread.

Confirmed after the rebase:

```
--unwind abc          exit 64 (EX_USAGE), diagnostic printed, not 134/SIGABRT
deep_binary_chain_pass  VERIFICATION SUCCESSFUL
deep_binary_chain_fail  VERIFICATION FAILED
```

@lucasccordeiro flagging this since it is your change on the other side of it, in case you would rather the handler sat somewhere else.

PR #6623, opened 2026-08-01
Replace VSA static analysis with Andersen points-to

rafaelsamenezes:
Replaces the GOTO-level points-to analysis with an inclusion-based (Andersen)
one and moves the two whole-program consumers — GCSE and k-induction's
pointer-array-write resolver — onto it. The analysis takes a fraction of a
second and is available to any other consumer that wants points-to data at
GOTO level.

## What changed

`andersent` is a flow- and context-insensitive whole-program points-to
analysis. It derives from `value_setst`, so it is a drop-in for
`value_set_analysist` at the two call sites that consume it. Dense node
interning, inclusion-constraint storage, and a worklist dynamic-transitive-
closure solver; calls through function pointers are resolved to a fixpoint,
since their targets are exactly what is being computed.

Deliberately imprecise: the analysis only has to authorise safe GOTO
rewrites, and symex supplies the real precision. A reserved `TOP` node ("may
point anywhere") is the soundness escape hatch — anything the frontend cannot
model routes there, and the query layer surfaces it as an unknown target so
consumers abstain.

`value_setst` grows `get_reference_set` / `has_location` so `andersent` and
`value_set_analysist` are interchangeable, and both consumers are retyped to
the interface rather than the concrete class.

This removes the bail-out paths both consumers needed. `value_set_analysist`
threw on incomplete cases (`vsa_not_implemented`, symbolic types, bare
strings) and asserted on k-induction-transformed CFGs, so each site caught
broadly and disabled its optimisation wholesale. `andersent` abstains per
query instead.

## Soundness fixes

Making the consumers actually run surfaced four latent bugs. Each has
regression tests, and each was verified by rebuilding without the fix to
confirm the test fails.

**`cse_domaint` bottom tracking, merge and havoc** (`b070395663`).
`is_bottom()` inferred bottom from an empty available-expression set, so a
reachable entry state was indistinguishable from an unvisited one — a merge
could adopt the other side's expressions wholesale, and `ai_baset` treated a
function whose exit had nothing available as unreachable. The intersection
loop also incremented an iterator after `erase()` returned the successor,
skipping every element after a removed one.

**Havoc through nested dereferences** (`cc7d62e5d3`). `havoc_expr` only
queried points-to when the assignment target was *itself* a `dereference2t`.
A store through `p->f` or `(*p)[i]` nests the dereference inside the lvalue,
so no query ran and only a syntactic kill applied — CSE then reused a value
the store had just invalidated. `--gcse` turned three correct programs into
`VERIFICATION FAILED`.

**Unknown, not empty, for an unresolvable reference set** (`bcc2ffcfd9`).
`get_values` / `get_reference_set` left `dest` empty when the expression had
no interned node, and every consumer reads an empty `valuest` as "refers to
no object" rather than "nothing is known". A doubly-indirect store such as
`(*q)->f = 100` reduced to a pointer expression that was never a node.

**Stale location numbers corrupting the GOTO program** (`74ddc8478f`) — the
serious one, and **not GCSE-specific**. `goto_functions_algorithm::run` built
a `goto_loopst` per function right after `runOnFunction`, but only called
`goto_functions.update()` once every function was done. Any pass that inserts
instructions therefore handed loop analysis a body with stale location
numbers: `insert_swap` leaves fresh nodes at the default `0`, and
`instructiont::swap` does not swap that field. `is_backwards_goto()`
classifies an edge by comparing those numbers, so a forward branch whose
target had also been displaced (both at `0`) looked like a back edge, and
`find_function_loops` read the self-loop test
`loop_head->location_number == loop_exit->location_number` as true and
rewrote the branch to `assume(!guard)`.

Under `--gcse` this silently turned failing programs into proofs. On
`sv-benchmarks/c/eca-rers2012/Problem17_label04.c` the first guard of
`calculate_output` became `ASSUME !1`, symex dropped from 602 to 31
assignments with no VCC at all, and ESBMC reported the reachable
`reach_error` as `VERIFICATION SUCCESSFUL`.

**Nine passes derive from `goto_functions_algorithm`** — `loop_unroll`,
`dead_store_analysis`, `goto_contractor`, `mark_decl_as_non_det`,
`assign_params_as_non_det`, the three `goto_check_*` passes and `gcse` — so
any of them that inserts an instruction could hit this. GCSE is simply the
one a benchmark run exercised. The fix renumbers before analysing, and tests
self-loop identity on the iterator rather than a mutable numeric field.

## SV-COMP

Benchexec, 100 s / 6 GB / 1 core.

Without `--gcse`, ReachSafety-ECA (1263 tasks) is byte-identical to master —
65 correct, 0 wrong, no status differs on any task. Across the full set
(30411 tasks) master scores 20027 correct / 5 wrong against 19876 / 6 here;
the delta is within the noise of a run without NUMA pinning.

With `--gcse` enabled, ReachSafety-ECA before and after the loop-renumber
fix:

| | correct | wrong |
|---|---|---|
| before | 137 | **48** |
| after | 65 | **0** |

All 48 wrong answers were tasks that had timed out on master, so nothing
regressed — they were results GCSE gained and got wrong. The 72 "extra
correct" that also disappear were the same artefact landing on the lucky side:
all 72 answered `true`, all 72 had timed out on master, and their CPU-time
distribution is indistinguishable from the 48 wrong ones (median 17.5 s vs
17.3 s). They were not proofs.

## Tests

- 6 new regression tests: four for the nested-dereference and unknown-target
  fixes (one a negative variant asserting the stale value must still fail),
  and two pinning "`--gcse` must not change the verdict" for the
  loop-renumber fix.
- 16 new unit tests: 8 driving the solver from synthetic constraints with no
  GOTO program needed, 8 covering the GOTO frontend end to end.
- Local: 623/623 unit, 196/196 k-induction, 4914/4921 `-L esbmc`. The seven
  failures are pre-existing and unrelated — all C++, and they reproduce with
  these changes reverted.

## Note for reviewers

Master recently fixed `value_sett::assign_rec` so VSA stops aborting on
Python programs, with `regression/python/gcse_vsa_ite_target` guarding it.
That test passes here, but its guard is vacuous on this branch: it asserts
the output does not contain `Unable to compute VSA`, a message this PR
removes along with the try/catch that produced it. The two changes are
independent fixes for the same underlying problem — that GCSE was silently
losing its points-to data — and the Andersen switch stands on the rest of its
case rather than on VSA being broken.


src/pointer-analysis/andersen.cpp
```diff
@@ -0,0 +1,655 @@
...
+  // A nil source is an unconstrained value, so route it through the node that
+  // already points anywhere rather than dropping the assignment.
+  if (is_nil_expr(rhs))
+  {
+    top_source();
+    handle_assign(lhs, top_source_expr(), loc);
+    return;
+  }
+
+  if (!may_carry_pointer(lhs->type) && !may_carry_pointer(rhs->type))
```
lucasccordeiro: `x = (long)p` is dropped here, so a later `q = (void *)x` copies an empty set instead of TOP — an under-approximation. Route pointer-to-integer assignments to TOP.

src/pointer-analysis/andersen.cpp
```diff
@@ -0,0 +1,655 @@
...
+        constraint_kindt::ADDRESS_OF,
+        t,
+        get_node(dynamic_object2tc(objtype, gen_ulong(loc), false, false)));
+      return t;
+    }
+
+    case sideeffect2t::allockind::nondet:
+      // A nondet value is not an address of anything we can name; it also
+      // cannot alias an existing object, so an empty set stays sound.
+      return fresh_node();
```
lucasccordeiro: A nondet *pointer* can be constrained to equal an existing object's address, so the empty set under-approximates. Return a TOP-pointing node for pointer-typed nondet.

src/pointer-analysis/andersen.cpp
```diff
@@ -0,0 +1,655 @@
...
+  const code_type2t &ftype = to_code_type(callee.type);
+  const std::size_t argc = std::min(
+    {ftype.arguments.size(), ftype.argument_names.size(), arguments.size()});
+  for (std::size_t i = 0; i < argc; ++i)
+  {
+    if (!may_carry_pointer(ftype.arguments[i]))
+      continue;
+
+    const node_id formal =
+      get_node(symbol2tc(ftype.arguments[i], ftype.argument_names[i]));
```
lucasccordeiro, line 511: Formals are interned by (type, name); if the body's symbol carries a different type (e.g. symbol_type vs followed struct) the binding targets another node. Key formals by identifier.

src/pointer-analysis/andersen.cpp
```diff
@@ -0,0 +1,655 @@
...
+      changed = true;
+    }
+  }
+
+  return changed;
+}
+
+void andersent::collect_constraints(const goto_functionst &goto_functions)
+{
+  // Known gap: inline asm can write memory through operands this frontend
```
lucasccordeiro: Inline asm writing through a pointer operand is exactly the under-approximation this design forbids. Route unmodelled OTHER instructions' operands to TOP instead of skipping them.

lucasccordeiro (changes requested):

PR #6626, opened 2026-08-01
[svcomp] Handle memsafety sub-properties individually

### LINKED ISSUE #1440: Wrapper script should begin to handle the CHECK() statements individually
In yesterday's SV-COMP meeting a proposal was made to introduce a new property file named `valid-memsafety-ub.prp` (for now) for SV-COMP containing the 2 sub-properties
```
CHECK( init(main()), LTL(G valid-free) )
CHECK( init(main()), LTL(G valid-deref) )
```
It is an adaptation of the existing memsafety one containing the three
```
CHECK( init(main()), LTL(G valid-free) )
CHECK( init(main()), LTL(G valid-deref) )
CHECK( init(main()), LTL(G valid-memtrack) )
```
I tried to explain to the group that this poses problems as the wrapper script decides the category assuming that these 3 sub-properties appear together uniquely in the properties file.

There was little success in explaining this on a formal level, because it appears that everyone on the call assumed that each sub-property was specified individually by SV-COMP's rules and therefore should be supported individually by the tool as well. Including arbitrary(?) boolean combinations of these sub-properties.

According to my understanding that is not what the [SV-COMP rules](https://sv-comp.sosy-lab.org/2024/rules.php) currently specify. Instead it defines the properties as "units", giving definitions for the combined set of checks.¹ But that can change. So we should adapt the wrapper (and possibly ESBMC's options) to account for individual sub-properties in order to be safe in the future.

For reference, this concerns the Juliet benchmarks, in particular these should be adapted according to the author of <https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1285>:
```
#CWE415_Double_Free
Juliet_Test/CWE415*.yml

#CWE476_NULL_Pointer_Dereference
Juliet_Test/CWE476*_int*.yml
Juliet_Test/CWE476*_long*.yml
Juliet_Test/CWE476*_struct*.yml

#CWE121_Stack_Based_Buffer_Overflow
Juliet_Test/CWE121*rand*.yml
Juliet_Test/CWE121*fgets*.yml
Juliet_Test/CWE121*large*.yml
```

¹ The rules page also lacks other specifications, like for instance what `CHECK` actually means, and what several `CHECK` terms in the file should mean. It also only uses the LTL `G` operator and could potentially use others, i.e., the logic isn't clear and only string equality can currently be used on the contents of the property files. But this issue is not about that.

lucasccordeiro:
The wrapper keyed the whole memsafety category off the presence of
`valid-free`, assuming SV-COMP's three sub-properties always appear together.
The proposed `valid-memsafety-ub.prp` has `valid-free` and `valid-deref` only,
and would still have been given `--memory-leak-check` and could have reported a
memtrack violation for a property it was never asked about.

Record which sub-properties the file actually contains, and gate both the leak
check and the `fail_memtrack` verdict on `valid-memtrack`.

Verified against master with `--dry-run`:

| property file | before | after |
|---|---|---|
| `memsafety.prp` (3 checks) | `--memory-leak-check` | unchanged |
| `valid-memsafety-ub.prp` (2 checks) | `--memory-leak-check` | no leak check |

`reach`, `overflow`, `memcleanup`, `termination` and `datarace` classify
unchanged.

Fixes #1440

### DIFF
--- a/scripts/competitions/svcomp/esbmc-wrapper.py
+++ b/scripts/competitions/svcomp/esbmc-wrapper.py
@@ -133,7 +133,7 @@ def parse_result(the_output, prop):
       return Result.fail_termination
 
     if prop == Property.memory:
-      if memory_leak in the_output:
+      if memory_leak in the_output and "valid-memtrack" in memsafety_subprops:
         return Result.fail_memtrack
 
       if invalid_pointer_free in the_output:
@@ -286,7 +286,9 @@ def get_command_line(strat, prop, arch, benchmark, concurrency, dargs, esbmc_ci,
   if prop == Property.overflow:
     command_line += "--no-pointer-check --no-bounds-check --overflow-check --no-assertions "
   elif prop == Property.memory:
-    command_line += "--memory-leak-check --no-reachable-memory-leak --no-assertions "
+    if "valid-memtrack" in memsafety_subprops:
+      command_line += "--memory-leak-check --no-reachable-memory-leak "
+    command_line += "--no-assertions "
     # It seems SV-COMP doesn't want to check for memleaks on abort()
     # see also <https://github.com/esbmc/esbmc/issues/1259>
     command_line += "--no-abnormal-memory-leak "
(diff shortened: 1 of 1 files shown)


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/30716264992.

lucasccordeiro:
Closing: the premise never materialised.

SV-COMP did not adopt the split property. The 2026 rules still define memory
safety as a unit — "All programs in category 'MemorySafety' either satisfy all
(partial) properties or violate exactly one (partial) property" — and
`sv-benchmarks` `c/properties/` contains only `valid-memsafety.prp`, still with
all three `CHECK()` lines. There is no `valid-memsafety-ub.prp`.

The change is behaviour-identical on the property file that actually exists, so
it fixes nothing live. Reopen if SV-COMP ever splits the property.

PR #6628, opened 2026-08-01
[goto-symex] Test MPOR's un-run case on the diagonal, as the recurrence does

lucasccordeiro:
A6.4 of the goto-symex verification plan asks whether the `dependency_chain`
update in `calculate_mpor_constraints` can lose a relation. It can, though not
where the plan suspected: MPOR guards its un-run case on `DCjj` — "Tj has not
run yet" — while we guard it on `DCji`. The two disagree for a thread created
after `Tj` last ran, because thread creation pads every existing row with 0 in
the new column, so the dependency between `Tj`'s last transition and the new
thread's first one is skipped instead of computed.

Evidence and its limits, so the change is not oversold:

- an invariant asserting the two guards coincide fires on an ordinary
  two-thread program, which is what identified the divergence;
- no behavioural difference has been observed. The probe program explores 49
  interleavings either way, and the concurrency regression subset (`race`,
  `mpor`, `pthread`, `cswitch`, `deadlock`, `circular`) is green apart from
  `01_pthread60`, which is the known 120 s-cap timeout — 105.2 s standalone
  here, matching its pre-patch profile;
- the direction is the safe one: the guard only ever adds chains, and a chain
  makes `can_run` more permissive, so no interleaving can be lost by it.

The unit test pins the chain's shape after every interleaving. It does not
discriminate the guard — reverting the guard leaves it green — and its header
says so rather than implying coverage it lacks.

### DIFF
--- a/src/goto-symex/execution_state.cpp
+++ b/src/goto-symex/execution_state.cpp
@@ -1225,7 +1225,11 @@ void execution_statet::calculate_mpor_constraints()
     if (j == active_thread)
       continue;
 
-    if (dependency_chain[j][active_thread] == 0)
+    // MPOR's rule is DCjj(k) == 0 -- "Tj has not run" -- not DCji(k) == 0. The
+    // two differ for a thread created after Tj last ran: its column is padded
+    // with 0, so testing it would skip the dependency below and drop the chain
+    // onto the new thread's first transition (A6.4).
+    if (dependency_chain[j][j] == 0)
     {
       // This thread hasn't been run; continue not having been run.
       new_dep_chain[j][active_thread] = 0;
--- a/src/goto-symex/execution_state.h
+++ b/src/goto-symex/execution_state.h
@@ -489,6 +489,12 @@ public:
     return mpor_says_no;
   }
 
+  /** Read-only accessor for the MPOR dependency chain, for the A6.4 harness. */
+  const std::vector<std::vector<int>> &get_dependency_chain() const
+  {
+    return dependency_chain;
+  }
+
   /** Accessor method for cswitch_forced. Sets it to true. */
   void force_cswitch()
   {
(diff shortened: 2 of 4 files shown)


lucasccordeiro:
Follow-up measurement, since the PR body said no behavioural difference had been
observed and that deserved a sharper answer.

**The dropped chain is live, not hypothetical.** An instrumented build that
reports every time the beef branch computes `res == 1` where the old guard would
have skipped it counts real drops on ordinary programs: 3 on a two-writer,
5 and 7 on staggered-creation variants, and 96 on a three-thread pipeline. So
the old guard was not merely unfaithful on paper — it was skipping dependency
computations that do produce a chain.

**It still changes nothing observable.** Interleaving counts are identical
between the two builds on every probe (69/69, 304/304, 537/537), and a sweep of
`regression/esbmc-unix` comparing verdicts between the patched and unpatched
binaries is at 171 tests with zero verdict divergences. One apparent count
divergence (`01_malloc_06`, 462 vs 458) was a 12 s-timeout artefact: run to
completion both give 479.

That is consistent with the argument in the PR body — the entry ends up `0`
rather than `1`, and `can_run` treats both as permissive, so the loss only bites
if a later two-hop step needed that `1` to avoid a stale `-1`. No program found
so far reaches that.

lucasccordeiro:
The `regression/esbmc-unix` A/B sweep has finished: **376 tests, zero verdict
divergences** between the patched and unpatched binaries.

One row looked like a divergence and is not. `github_6475_safe` reported
SUCCESSFUL on the unpatched binary and nothing on the patched one, because the
sweep used a 12 s cutoff and the test straddles it: run to completion both
report SUCCESSFUL, at 11.70 s unpatched and 12.27 s patched. That ~5 % is the
only cost signal the sweep produced, and it is the expected direction — the
guard records chains that were previously skipped, and a recorded chain makes
`can_run` more permissive. Nine other rows differ only in how far each binary
got inside the cutoff, all with both verdicts absent.

PR #6651, opened 2026-08-02
close remaining numpy view escape and method-form gaps

sergillam:
- view used inline in a list/tuple literal without a named binding (`holder = [x[0]]`) escaped detection entirely; only the already-named form (`row = x[0]; holder = [row]`) was caught.
- Fixes a real process crash: storing a view in a `dict` literal (named or inline) aborted the SMT solver (`z3_conv.cpp` `mk_eq` width-mismatch) instead of producing a diagnostic — `dict_handler_`'s literal-assignment paths never reached the container-escape check that already protected `List`/`Tuple`.
- Adds a regression test confirming escape via `global` from a function-parameter-sourced view (as opposed to a locally created one) is already correctly rejected.
- Fixes `a.reshape(d1, d2, ...)` (dimensions as separate positional arguments): it silently dropped every dimension past the first; only the tuple-shape form worked.
- Adds the real NumPy method forms `a.flatten()`, `a.sum()`, `a.mean()`, `a.min()`, `a.max()`, `a.std()`, `a.var()` — previously unhandled `AttributeError`/`TypeError` — by extending the existing method-to-function dispatch rewrite already used for `a.transpose()`/`a.reshape()`/`a.ravel()`.
- Gives `a.flat[i] = x` a specific diagnostic (`TypeError: mutation through .flat is not supported`) instead of the generic `Unsupported assignment target type: Call`; the write itself remains unsupported

### DIFF
--- a/regression/numpy/reducer_method_success/main.py
+++ b/regression/numpy/reducer_method_success/main.py
@@ -1,0 +1,17 @@
+import numpy as np
+
+a = np.array([1.0, 2.0, 3.0])
+
+s = a.sum()
+m = a.mean()
+mn = a.min()
+mx = a.max()
+sd = a.std()
+vr = a.var()
+
+assert s == 6.0
+assert m == 2.0
+assert mn == 1.0
+assert mx == 3.0
+assert sd >= 0.0
+assert vr >= 0.0
--- a/regression/numpy/reducer_method_success/main.py
+++ b/regression/numpy/reducer_method_success/main.py
@@ -13,5 +13,7 @@ assert s == 6.0
 assert m == 2.0
 assert mn == 1.0
 assert mx == 3.0
-assert sd >= 0.0
-assert vr >= 0.0
+# var = mean((x - mean)^2) = ((1-2)^2 + (2-2)^2 + (3-2)^2) / 3 = 2/3
+assert vr > 0.6666 and vr < 0.6667
+# std = sqrt(var) = sqrt(2/3)
+assert sd > 0.8164 and sd < 0.8165
--- a/src/python-frontend/converter/converter_stmt.cpp
+++ b/src/python-frontend/converter/converter_stmt.cpp
@@ -1316,7 +1316,21 @@ bool python_converter::contains_copied_numpy_view_name(
 
   if (node.is_object())
   {
-    if (node.value("_type", "") == "Name" && node.contains("id"))
+    const std::string node_type = node.value("_type", "");
+
+    // A comprehension/generator always builds a brand-new list/set/dict, so
+    // it cannot itself be a numpy view; and its element/key/value
+    // expressions reference the comprehension's own loop variable(s), which
+    // are not registered as real symbols outside of the comprehension's own
+    // conversion (handle_comprehension/_lower_listcomp) — probing a
+    // Subscript inside one here (e.g. `x[j]` for `for j in ...`) would look
+    // up `j` before it exists and abort the conversion.
+    if (
+      node_type == "GeneratorExp" || node_type == "ListComp" ||
+      node_type == "SetComp" || node_type == "DictComp")
+      return false;
+
+    if (node_type == "Name" && node.contains("id"))
     {
       const std::string id =
         resolve_name_symbol_id(node["id"].get<std::string>());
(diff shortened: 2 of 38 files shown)


src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -1316,15 +1316,63 @@ bool python_converter::contains_copied_numpy_view_name(
...
+    // get_conditional_stm's own short-circuit guard is built), so those
+    // instructions must not leak into the real block. Redirect them into a
+    // throwaway block for the duration of the probe.
+    if (
+      is_basic_numpy_view_subscript(node) &&
+      !root_name_from_subscript(node["value"]).empty())
+    {
+      code_blockt scratch_block;
+      code_blockt *saved_block = current_block;
+      current_block = &scratch_block;
```
lucasccordeiro, line 1376: `get_expr` can throw, leaving `current_block` pointing at the destroyed `scratch_block`. Restore it with a scope guard or try/catch.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3584,22 +3671,41 @@ void python_converter::get_var_assign(
...
-      !base_is_imported_module && method_name == "copy" &&
+    // A method name like sum()/max()/min() is not exclusive to numpy (e.g.
+    // Decimal.max(), a plain module-level function called through an
+    // aliased import); only rewrite when the receiver is actually a
+    // tracked numpy array, matching the check already used for copy()
+    // below.
+    const bool method_base_is_numpy_array =
       !method_base_id.empty() &&
       numpy_array_symbols_.count(method_base_id) != 0;
+    const bool supported_view_method =
```
lucasccordeiro, line 3734: Regression: `numpy_array_symbols_` never records `np.eye`/`identity`/`linspace` results, so `a.transpose()` on those silently stops being rewritten. Add them to `is_numpy_array_constructor_expr`.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3435,6 +3511,17 @@ void python_converter::get_var_assign(
 
     bool is_global = is_global_variable(sid);
 
+    // Same reasoning as the annotated-assignment dict check above: this
+    // path also hands off to dict_handler_ before the generic container
+    // escape check further down ever runs.
+    if (
```
lucasccordeiro: Byte-identical to the check added ~40 lines above, and to the generic one below. Extract one helper and call it from both sites.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3405,6 +3466,21 @@ void python_converter::get_var_assign(
...
     lhs = create_lhs_expression(target, lhs_symbol, location_begin);
 
+    // A dict literal containing a copied numpy view (named or inline, e.g.
+    // {"row": x[0]}) is handled entirely by dict_handler_ below, which
+    // never routes through the generic List/Tuple/Dict escape check further
+    // down this function (dict_handler_ returns before reaching it) — so it
+    // has to be caught here too, or the view ends up embedded in the dict's
+    // runtime representation in a way that crashes SMT encoding instead of
+    // producing a diagnostic (mismatched sort widths in z3_convt::mk_eq).
+    if (
```
lucasccordeiro: No test reaches this annotated-assignment path; `view_dict_literal_escape_fail` only exercises the unannotated copy. Add a `holder: dict = {"row": x[0]}` case.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2560,6 +2608,19 @@ python_converter::extract_target_name(const nlohmann::json &target) const
...
+    target["func"].value("_type", "") == "Attribute" &&
+    target["func"].value("attr", "") == "ravel")
+    // a.flat[i] = x: the preprocessor rewrites every .flat read, including
+    // the one implicit in this assignment's target, to np.ravel(a) — so the
+    // target here is really Subscript(value=Call(ravel(a))), which has no
+    // symbol id to extract. np.ravel(a)[i] = x written directly hits the
+    // same shape and is equally unsupported for the same reason (ravel's
+    // result is a copy, not a writable view of a).
+    throw std::runtime_error(
+      "TypeError: mutation through .flat is not supported");
```
lucasccordeiro, line 2704: This fires for any `obj.ravel()[i] = x`, including a user class, reporting a `.flat` diagnostic that does not apply. Gate on a numpy-sourced receiver.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5930,6 +5930,22 @@ exprt numpy_call_expr::get()
         raw_shape.push_back(d);
       }
     }
+    else if (call_["args"].size() > 2)
```
lucasccordeiro, line 5933: `np.reshape(a, 2, 3)` also lands here, but numpy rejects it (third positional is `order`). Mark the method-form rewrite so only it accepts split dimensions.

regression/numpy/reducer_method_success/main.py
```diff
@@ -0,0 +1,17 @@
...
+mn = a.min()
+mx = a.max()
+sd = a.std()
+vr = a.var()
+
+assert s == 6.0
+assert m == 2.0
+assert mn == 1.0
+assert mx == 3.0
+assert sd >= 0.0
```
lucasccordeiro: `>= 0.0` passes for any non-negative value, so std/var are effectively untested. Assert the actual results with tight bounds (var 0.666..., std 0.816...).

lucasccordeiro (changes requested):

regression/numpy/reducer_method_success/main.py
sergillam: done

src/python-frontend/numpy/numpy_call_expr.cpp
sergillam: done

src/python-frontend/converter/converter_stmt.cpp
sergillam: done

src/python-frontend/converter/converter_stmt.cpp
sergillam: done

src/python-frontend/converter/converter_stmt.cpp
sergillam: done

src/python-frontend/converter/converter_stmt.cpp
sergillam: done

src/python-frontend/converter/converter_stmt.cpp
sergillam: done

PR #6681, opened 2026-08-03
[goto-programs] Delegate loops whose body does not convert natively

lucasccordeiro:
Phase 1 of `docs/roadmap/frontends-to-irep2.md`. Master-based.

**Draft: one gate is outstanding — see below.**

`convert_native_rec` failed the whole-function walk when a `while`/`for` body
did not convert natively or leaked a scope-exit entry. It now delegates the
loop to `convert_while` / `convert_for`, rolling back the abandoned attempt's
`tmp_symbol` counter, `context` and `targets` (the same rollback #6679 adds for
`if` branches).

**Snapshot placement is load-bearing.** It must be taken at the *top* of the
handler, not just before the body: a side-effecting loop condition already
allocates temps through `generate_conditional_branch`, so a later snapshot
leaves those in place and the delegated conversion numbers its own temps one
past them. Snapshotting after the condition left 17 of 51 sampled tests
diverging on `tmp$N`; at the top, 0 diverge.

Gates discharged: `--goto-functions-only` A/B against `--no-irep2-native-body`,
**0 divergences over 51 tests** — byte-identical GOTO, which is what feeds
symex. The regression test is shown to reach the delegated site (7 hits vs 0
on a control).

Gate outstanding: the full `esbmc-cpp` failure-set comparison against a control
build. The dev machine is currently contended by other users (load ~12, three
tests in nine minutes), so the suite could not complete. Marked draft until
that runs — CI will also cover it.


lucasccordeiro:
Outstanding gate now discharged. \`esbmc-cpp\` **9 failures / 721, diff-identical to a control build** — the same set that fails on master here (`github_2242_1/2`, `github_3897_collision`, `github_6200_fail`, `github_6227_fail`, `github_6229_fail`, `github_6494_member_new_fail`, `github_6494_pool_new_fail`, `utility2_fail`).

That was the only reason this was draft; the A/B (0 divergences over 51 tests, byte-identical GOTO) was already recorded in the description. Branch merged with current master before the run. Marking ready.

PR #6684, opened 2026-08-03
[solver] Do not take shift count signedness into the shl overflow check

rafaelsamenezes:
## Problem

`overflow_arith()` derived `is_signed` from **either** operand:

```cpp
bool is_signed = (is_signedbv_type(side1) || is_signedbv_type(side2));
```

For a shift that is wrong. The count is a separate operand whose type says
nothing about the range of the result. `(u64)x << 32` has a plain signed `int`
count, so it took the signed path — which requires the top `sz+1` bits of the
widened result to be all-zero or all-one, and therefore reports overflow for
every result `>= 2^63`, even though all of them are representable in `u64`.

Minimal reproducer:

```c
unsigned int nondet_uint();
int main() {
  unsigned int hi = nondet_uint();
  unsigned long long v = ((unsigned long long)hi << 32);
  assert(v >= 0);
  return 0;
}
```

```
$ esbmc shl.c --unsigned-overflow-check
Violated property:
  arithmetic overflow on shl
  !overflow("shl", (unsigned long int)hi, 32)
VERIFICATION FAILED
```

with the counterexample `hi = 2147483648` (2^31), whose result is `2^63` —
comfortably inside `u64`.

The idiom this breaks is the standard one for assembling a 64-bit register
value from two halves, `((u64)hi << 32) | lo`, which is everywhere in kernel
and driver code.

## Fix

Consult only the shifted operand's signedness for `shl`. `mul` (which shares
the case) keeps the existing behaviour, where both operands genuinely matter.

## Tests

- `regression/esbmc/unsigned_shl_widened` — the widened idiom, now
  `VERIFICATION SUCCESSFUL`.
- `regression/esbmc/unsigned_shl_widened_fail` — `u32 << 8` genuinely
  discarding set bits, still `VERIFICATION FAILED`.

Signed `shl` overflow is also still reported (checked manually with
`int x; x << 30`).

## One test reclassified

`regression/esbmc-unix/instrumented_nohup_comb_overflow` becomes `KNOWNBUG`.
Its expected `VERIFICATION FAILED` was **produced by this false positive**,
firing on the SHA-512 table constants at `main.i:3443+`
(`(N_UL << 32) + M_UL`, flagged for any `N >= 2^31`).

With the false positive gone, symbolic execution reaches further into the file
and hits a separate, pre-existing defect:

```
terminate called after throwing an instance of 'array_type2t::dyn_sized_array_excp'
  what():  Sizeof nondeterministically sized array encountered
```

That crash is **independent of this change** — it reproduces with
`--unsigned-overflow-check` removed entirely. The `test.desc` records the
diagnosis and what to fix before restoring the test to `CORE`.

I'd rather flag this than quietly leave a test whose green status depended on a
wrong answer.


rafaelsamenezes:
Follow-up on the `instrumented_nohup_comb_overflow` reclassification, since my original note was wrong.

I said the test could return to `CORE` once the `dyn_sized_array_excp` crash was fixed. I've now fixed that crash separately (#6687) and re-run this test with **both** changes: it no longer aborts, but it does not terminate either — `--incremental-bmc --unlimited-k-steps` over this file was still unwinding after **40 minutes and 10 GB**.

So the false positive had been supplying an early exit, and without it there is no evidence this test ever reached a verdict on its own merits. Restoring it needs a bound that makes it terminate — most likely a `--k-step` ceiling in place of `--unlimited-k-steps`, chosen so the expected `VERIFICATION FAILED` is actually reachable. That is a decision about what the test is meant to check, so I have left it to whoever owns it rather than guessing.

The `test.desc` note is corrected accordingly in f39598d. #6687 is independent of this PR and does not touch this test.

regression/esbmc-unix/instrumented_nohup_comb_overflow/test.desc
```diff
@@ -1,4 +1,35 @@
-CORE
+KNOWNBUG
 main.i
---no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions --incremental-bmc 
+--no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions --incremental-bmc
 ^VERIFICATION FAILED$
+--
```
lucasccordeiro, line 5: `test.desc` has no `--` separator: every line from 4 on becomes a required regex (testing_tool.py:390). Move this prose to the commit message.

regression/esbmc-unix/instrumented_nohup_comb_overflow/test.desc
```diff
@@ -1,4 +1,35 @@
-CORE
+KNOWNBUG
```
lucasccordeiro, line 1: Downgrading a CORE test to KNOWNBUG drops coverage silently. Replace `--unlimited-k-steps` with a `--k-step` bound that terminates, keeping it CORE.

lucasccordeiro (changes requested):

regression/esbmc-unix/instrumented_nohup_comb_overflow/test.desc
```diff
@@ -1,4 +1,35 @@
...
 main.i
---no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions --incremental-bmc 
+--no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions --incremental-bmc
 ^VERIFICATION FAILED$
+--
+This test's VERIFICATION FAILED was an artefact of a false positive in the
```
lucasccordeiro, line 6: These prose lines are required regexes that can never match, so `all_checks_pass` is always false and the exit-77 reclassification signal can never fire once the bug is fixed.

regression/esbmc-unix/instrumented_nohup_comb_overflow/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.i
---no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions --incremental-bmc 
-^VERIFICATION FAILED$
+--no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --unwind 10 --floatbv --no-vla-size-check --64  --no-por --context-bound 3 --no-pointer-check --no-bounds-check --unsigned-overflow-check --no-assertions
```
lucasccordeiro, line 3: With `--unwind 10` the verdict now comes from an unrelated libc `strchr` unwinding assertion, not the overflow (measured). Choose a bound under which the overflow is the failing property.

regression/esbmc-unix/instrumented_nohup_comb_overflow/test.desc
```diff
@@ -1,4 +1,4 @@
...
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: KNOWNBUG requires this regex to NOT match, but the run still prints VERIFICATION FAILED, so the test exits 77. Restore CORE or expect the actual output.

regression/esbmc/unsigned_shl_widened_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unsigned-overflow-check
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone matches any failure. Add the `arithmetic overflow on shl` violated-property line so the test pins the shift check.

regression/esbmc/unsigned_shl_widened_fail/main.c
```diff
@@ -0,0 +1,15 @@
...
+
+unsigned int nondet_uint();
+
+/* No widening here, so the shift genuinely discards set bits and the
+   unsigned overflow check must still fire. */
+int main()
+{
+  u32 hi = nondet_uint();
+  u32 v = hi << 8;
+  assert(v >= 0);
```
lucasccordeiro, line 13: `v >= 0` is tautological for `u32` and reports PASSED, so it checks nothing. Drop it or assert a property the shift can violate.

PR #6685, opened 2026-08-03
[docs] Open the relational/float reconciliation scope

lucasccordeiro:
Docs only. The second of the two mechanisms blocking the `python_adjust` flip,
and the last one without an owner document — with #6683 this closes the residue
`scope-coupled-arith-assign-conversion.md` §13.1 left open.

All four witnesses (`chained-comparison2_fail`, `lambda15`, `precedence2`,
`sum_tuple`) share one property: an integer and a floating-point operand
meeting at a comparison, a chained comparison, or a bitwise assignment. That is
why per-case triage kept re-homing them — they share a mechanism, not a
syntactic construct.

Read from the tree: `python_adjust`'s relational arm admits only `bv`/`bv`
pairs (`python_adjust.cpp:406-409`), so `0 <= x` with `x` floatbv is never
reconciled, while its arithmetic arm admits floatbv and fixedbv too (`:449`).
There is also no `equality2t`/`notequal2t` arm at all.

Recorded as a **hypothesis gated on a Phase 0 census**, not a conclusion — this
scope's own §13→§14 reversal is the precedent. It is also gated hard on gap-2:
widening the relational guard generally was already tried and rejected for
diverging corpus-wide from clang's promotions over the OM bodies, so the change
needed is a narrowed admission rule, not dropping the guard.

The second commit cross-references both owner documents from the coupled scope.


lucasccordeiro:
Phase 0 has been run and is recorded as §11. It **partially refutes** the hypothesis this scope opened with, which changes the scope:

- **§4 holds for three of four.** `chained-comparison2_fail`, `lambda15` and `precedence2` each reach a float/integer comparison that no arm admits. `precedence2`'s is an **equality**, so the missing `equality2t`/`notequal2t` arm is load-bearing on its own, not a tidiness gap — this scope must cover both halves.
- **`sum_tuple` is refuted.** It contains no heterogeneous comparison at all, so it belongs to a third mechanism and is removed from the witness list rather than carried as a straggler.
- **The existing guard never fires** on any of this traffic (`admitted=0` everywhere), so widening it cannot regress these tests by changing existing behaviour; the risk is entirely in what the new admission rule lets through. gap-2 (G2) remains the gate that matters.

§11.4 also records what Phase 0 did *not* answer: only the four witnesses were censused, not a corpus slice, so the traffic-volume question §6 asks is still open and G2 cannot be discharged without it.

lucasccordeiro:
Phase 0/1 are now complete enough to say this scope's premise is wrong, and I'd rather record that than keep widening.

**§14 measurement:** instrumenting the comparison dispatch for width-mismatched operands returns **zero across all four witnesses**, while that same dispatch sees ~1300 comparison nodes per run — so the walk is reaching the traffic, and the equality that aborts bitwuzla's `mk_eq` simply is not in it. No admission rule at `python_adjust` — narrow, wide or unconditional — can reconcile a node it never sees. Two Phase 1 attempts confirmed this the expensive way before the measurement confirmed it cheaply.

**Where that leaves the four witnesses:**
- `sum_tuple` — refuted in §11, no heterogeneous comparison at all
- `lambda15` — aborts in `mk_eq`, provably not reachable from this dispatch
- `chained-comparison2_fail`, `precedence2` — do contain float/int comparisons the dispatch sees, but whether reconciling them clears the tests is still untested; both attempts aborted on `lambda15`'s shape first

So this may be **two mechanisms**, not one: an adjuster-visible float/int gap for two witnesses, and a downstream width gap for `lambda15`. The abort surfaces after symex, and the census used `--goto-functions-only`, so it cannot distinguish "in the GOTO but outside the walk" from "synthesised during symex" — that is the next probe, and it belongs at the symex/SMT boundary.

I'd suggest treating this PR as the record of that investigation rather than a plan to implement as written. Happy to re-scope it once the layer is known.

lucasccordeiro:
The three probes §15.4 required have been run, and they close this scope.

| probe | answer |
|---|---|
| Does the arm fire on `floatbv == bool`? | **yes**, 6× in `lambda15` |
| Does `c_implicit_typecast_arithmetic` convert it? | **no — and that was a real bug.** The IREP2 `get_c_type` overload never classified `floatbv`, so it returned `OTHER`, which outranks every arithmetic kind; the helper converted neither operand. Fixed in #6688 |
| With the helper fixed, does the arm reconcile? | **yes** — `before=floatbv/bool` → `after=floatbv/floatbv` |

**And `lambda15` aborts identically.** Arm fires, helper converts, operands reconciled, abort unchanged — so the equality reaching the solver is a *different instance*, rebuilt after `python_adjust`. That is what §14 claimed for the wrong reason and §15 withdrew; it is now measured rather than inferred.

I've closed the scope rather than leave a plan that four attempts have refuted. The equality arm was implemented and works, but clears no test, so I did not ship it — a behaviour change on a flag-gated path with no demonstrated benefit isn't worth the risk.

The investigation's shippable output is **#6688**: `c_implicit_typecast_arithmetic` has been a silent no-op for any `expr2tc` pair with a float operand, affecting five Python call sites, four on the default path. The original hypothesis was wrong; chasing it surfaced a real latent bug nothing else had.

§16.4 tells whoever takes the downstream mechanism where to start (`smt_solver.cpp:1320`) and, more usefully, where not to.

Issue #5783, opened 2026-07-03
[ci] Promote ASan/UBSan sanitizer CI to a required PR gate


pmatos:
## Background

PR #4481 added an ASan/UBSan sanitizer CI workflow that currently runs with
`continue-on-error: true` so its findings do not block merges during bring-up.

This issue tracks the second of the two follow-ups from
https://github.com/esbmc/esbmc/pull/4481#issuecomment-4874772445: once the
sanitizer suite is green, **promote it to a required PR gate** so regressions
are caught at review time.

With more PRs arriving via agentic coding, every additional gate at the PR
level tightens the feedback loop and keeps the tree clean.

## Blocked by

- Blocked by #5782 — the sanitizers must report zero unsuppressed findings
  before this workflow can be made blocking.

## Scope

- Remove `continue-on-error` from the sanitizer jobs in
  `.github/workflows/sanitizers.yml` so a finding fails the job.
- Mark the ASan and UBSan jobs as required status checks on `master` (branch
  protection).
- Confirm the gate behaves correctly: a PR that introduces a sanitizer finding
  is blocked, and a clean PR passes.

## Done when

- The sanitizer workflow is a required, blocking check on PRs to `master`.

Issue #5782, opened 2026-07-03
[ci] Run ASan/UBSan sanitizer suite and fix all findings


pmatos:
## Background

PR #4481 added a GitHub Actions workflow that builds ESBMC under clang's
AddressSanitizer and UndefinedBehaviorSanitizer, runs the `CORE` regression
suite under each, and deduplicates the diagnostics into a job-summary table.
During bring-up the workflow runs with `continue-on-error: true` so that the
sanitizer findings do not block merges while the suppression files are being
populated.

This issue tracks the first of the two follow-ups from
https://github.com/esbmc/esbmc/pull/4481#issuecomment-4874772445: actually
running the sanitizers and **fixing the real issues they surface** so the suite
comes back clean.

## Scope

- Triage every ASan/UBSan finding reported by the new workflow.
- Fix genuine memory-safety and undefined-behaviour bugs in ESBMC's own code.
- For findings that are false positives or third-party/unactionable, add a
  justified entry to the appropriate suppression file
  (`scripts/sanitizers/asan-suppressions.txt`,
  `scripts/sanitizers/lsan-suppressions.txt`,
  `scripts/sanitizers/ubsan-suppressions.txt`) with a comment explaining why.
- Goal: the ASan and UBSan jobs report **zero unsuppressed findings** on the
  `CORE` regression suite.

## Done when

- The sanitizer workflow runs green (no unsuppressed findings) on `master`.
- Each suppression entry has a rationale.

## Follow-up

Once this is green, #5783 turns the workflow into a required PR gate.

PR #6726, opened 2026-08-05
[ci] Fold heap addresses out of the sanitizer dedup key

lucasccordeiro:
The UBSan job has been failing, but not on anything it found: the report outgrew GitHub's 1 MB job-summary limit.

```
28044 unique finding(s); 28044 total occurrence(s).
##[error]$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of 1024k, got 5788k
```

UBSan embeds the faulting address mid-message — `member access within address 0x5555... with insufficient space for ...` — and those messages carry no colon for the `split(":", 1)` kind-cut to trim at, so the address survived into the dedup key. Every occurrence of one recurring defect counted as a separate finding.

Folding addresses to a placeholder collapses **that same CI log from 28044 occurrences to 19 findings**, which I verified by replaying the real log from run 30960023559 through the parser.

Worth recording for #5782: all 19 are in vendored `immer` (`build/_deps/immer-src/immer/detail/hamts/node.hpp`) — **none in ESBMC's own code**. So "fix all findings" is a much smaller job than the headline number suggested, and it is mostly a question of whether to suppress a third-party dependency. This also unblocks #5783, since a job that always fails on summary size cannot become a required gate.

Two regression tests added: one asserting three address-varying occurrences dedup to a single finding, one asserting the colon-split still trims `signed integer overflow: ...` as before. 30/30 tests pass.

### DIFF
--- a/scripts/sanitizers/collect_findings.py
+++ b/scripts/sanitizers/collect_findings.py
@@ -40,6 +40,14 @@ from typing import Iterable, Iterator, List
 _UBSAN_RE = re.compile(
     r"(?P<loc>[^:\s]+:\d+(?::\d+)?): runtime error: (?P<msg>.+)$")
 
+# Heap addresses vary run to run and even between occurrences of the same
+# defect. UBSan embeds them mid-message ("member access within address 0x55..
+# with insufficient space for ..."), and those messages carry no colon for the
+# split below to cut at, so the address survived into the dedup key: one
+# recurring finding became tens of thousands of "unique" ones, and the report
+# outgrew the 1 MB job-summary limit. Fold them to a placeholder.
+_ADDR_RE = re.compile(r"0x[0-9a-fA-F]+")
+
 # ASan/LSan reports begin with a header line containing the tool and a
 # kind identifier. The kind is the first token after the colon; ASan
 # sometimes appends parenthesised qualifiers (e.g.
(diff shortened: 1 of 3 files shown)


lucasccordeiro:
Added a second commit populating `ubsan-suppressions.txt`, which the file's own placeholder and `sanitizers.yml:17` describe as the next step.

I pulled the raw artifact from run 30960023559 rather than infer the check name, and the picture is simpler than the counts suggest:

- **All 28044 reports are a single check** — `insufficient-object-size` — in a **single vendored header**, `build/_deps/immer-src/immer/detail/hamts/node.hpp`.
- immer's HAMT node tail-allocates its children past the declared struct, sized from the node's bitmap. UBSan's object-size check does not model tail allocation, so this is a false positive of the idiom rather than a defect, in code we do not own.
- **Nothing in ESBMC's own code fired at all.**

So #5782's "fix all findings" reduces to suppressing one third-party idiom.

**One caveat I cannot settle from here.** UBSan's runtime-suppression support varies by check, and I have no sanitizer build locally to confirm `insufficient-object-size` is suppressible by name. If the next CI run still reports it, the alternative is to stop instrumenting `_deps` (a `-fsanitize-ignorelist` entry or excluding vendored sources), which is a build change rather than a suppressions one.

That uncertainty is also why I have **not** touched `continue-on-error` here. #5783 should follow only once a run confirms this lands at zero findings — removing the guard on an unverified suppression would just move the failure.

PR #6766, opened 2026-08-05
[c2goto] Compile the operational models out of the VFS

Yiannis128:
c2goto was handed `${CMAKE_CURRENT_SOURCE_DIR}/library/*.c`, so every operational-model symbol in `clib*.goto` carried the absolute path of whatever machine built it. That made the shipped goto binaries non-reproducible, put the builder's home directory in front of users in counterexamples, and forced `is_bundled_source()` to recognise ESBMC's own models by matching a build-tree path. c2goto now reads those sources from the same `/esbmc-vfs` namespace ESBMC serves them from, so the recorded locations are build-independent and the predicate collapses to a single VFS prefix test.

Getting there needed the model sources bundled into c2goto (data only, generated by `mangle()` without c2goto's involvement, so no dependency cycle), `register_bundled_libc()` called before parsing rather than from inside the frontend, and `language_uit::parse` answering from `filesystemt` for bundled paths, which never exist on disk, while still opening anything else. musl, Python and Solidity keep their private headers in separate flail lists because `flail.py` keys symbol names on the stem, so `exp2f_data.{c,h}` would collide. Solidity is amalgamated into one TU, so only its generated `#include` lines needed rewriting.

Verified on clean builds with the Python frontend both on and off: no builder path remains in any `clib*.goto`. `regression/esbmc/clib_source_from_vfs` pins the same thing at runtime, by requiring the model location a counterexample reports to be a `/esbmc-vfs` path.

Depends on #6517; based on `6509-vio` and will retarget to `master` when that merges.

### DIFF
--- a/regression/esbmc/clib_source_from_vfs/main.c
+++ b/regression/esbmc/clib_source_from_vfs/main.c
@@ -1,0 +1,8 @@
+#include <string.h>
+
+int main(void)
+{
+  char b[2];
+  strcpy(b, "abcd");
+  return 0;
+}
--- a/regression/esbmc/clib_source_from_vfs/main.c
+++ b/regression/esbmc/clib_source_from_vfs/main.c
@@ -1,8 +1,0 @@
-#include <string.h>
-
-int main(void)
-{
-  char b[2];
-  strcpy(b, "abcd");
-  return 0;
-}
--- a/src/c2goto/CMakeLists.txt
+++ b/src/c2goto/CMakeLists.txt
@@ -1,11 +1,21 @@
 add_subdirectory(headers)
 
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr)
-file(TOUCH ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/libc.h
-           ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/libm.h)
+# c2goto reads the model sources it compiles out of the same VFS that ESBMC
+# serves them from, so the goto binary records /esbmc-vfs paths rather than this
+# machine's build tree. That needs the real bundled lists, not the blank ones a
+# tool that must not *link* a C library would otherwise get.
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/headers)
-file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/headers/libc_hdr.h
-           "#include \"${CMAKE_CURRENT_BINARY_DIR}/headers/libc_hdr.h\"")
+set(c2goto_bundled_lists libc libm libmusl libmuslhdr libcpp headers/libc_hdr)
+if(ENABLE_PYTHON_FRONTEND)
+  list(APPEND c2goto_bundled_lists libpython libpythonhdr)
+endif()
+if(ENABLE_SOLIDITY_FRONTEND)
+  list(APPEND c2goto_bundled_lists libsolidity libsolidityhdr)
+endif()
+foreach(lst ${c2goto_bundled_lists})
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/${lst}.h
+             "#include \"${CMAKE_CURRENT_BINARY_DIR}/${lst}.h\"")
+endforeach()
 
 add_executable(c2goto c2goto.cpp cprover_blank_library.cpp cprover_libc_sources.cpp)
 target_include_directories(c2goto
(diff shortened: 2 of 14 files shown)


src/langapi/language_ui.cpp
```diff
@@ -52,9 +52,9 @@ bool language_uit::parse(const std::string &filename)
 
   config.language.lid = lang;
 
-  // Check that it opens
-  std::ifstream infile(filename.c_str());
-  if (!infile)
+  // Bundled sources answer here too: c2goto compiles the operational models
+  // straight out of the VFS, so they never exist on disk.
+  if (!file_operations::filesystemt::get().exists(filename))
```
lucasccordeiro: `exists()` is true for files that cannot be opened (mode 000), which the old `ifstream` check rejected. Fall back to opening when the path is not bundled.

src/c2goto/CMakeLists.txt
```diff
@@ -225,6 +236,32 @@ function(mangle_clib output)
...
+  mangle(${CMAKE_CURRENT_BINARY_DIR}/libmuslhdr.c
+         ${CMAKE_CURRENT_SOURCE_DIR}/library/libm/musl
+         WILDCARD *.h
+         SINGLE
+         PREFIX esbmc_libmuslhdr_
+         MACRO ESBMC_FLAIL
+         ACC_HEADERS_INTO ${CMAKE_CURRENT_BINARY_DIR}/libmuslhdr.h)
+  mangle(${CMAKE_CURRENT_BINARY_DIR}/libcpp.c
+         ${CMAKE_CURRENT_SOURCE_DIR}/library/cpp
+         WILDCARD *.cpp
```
lucasccordeiro, line 247: `c2goto_cpp_files` uses `GLOB_RECURSE` but this globs non-recursively, so a `library/cpp` subdirectory would yield an unregistered VFS path. Add `RECURSIVE 1`; same for the python and solidity mangles.

src/c2goto/CMakeLists.txt
```diff
@@ -1,11 +1,21 @@
...
-           "#include \"${CMAKE_CURRENT_BINARY_DIR}/headers/libc_hdr.h\"")
+set(c2goto_bundled_lists libc libm libmusl libmuslhdr libcpp headers/libc_hdr)
+if(ENABLE_PYTHON_FRONTEND)
+  list(APPEND c2goto_bundled_lists libpython libpythonhdr)
+endif()
+if(ENABLE_SOLIDITY_FRONTEND)
+  list(APPEND c2goto_bundled_lists libsolidity libsolidityhdr)
+endif()
+foreach(lst ${c2goto_bundled_lists})
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/c2goto-hdr/${lst}.h
```
lucasccordeiro: These reuse the names master `file(TOUCH)`es, so a shared build directory keeps stale content and ninja reports a cycle. Write the forwarders to a distinct directory instead.

lucasccordeiro (commented):
No test covers `language_uit::parse()` accepting a bundled path that is absent from disk; add one so the clib build is not the only thing exercising it.

src/langapi/language_ui.cpp
Yiannis128: Fixed in d7155db. Worth flagging that the frontend segfaults on an unopenable path, so `exists()` alone was more than a worse message; bundled paths now answer from the registry and everything else opens.

src/c2goto/CMakeLists.txt
Yiannis128: Done in eb3feb6. libc and libm stay non-recursive on purpose: a recursive libm would register the musl sources a second time under the same VFS path, which `add_bundled()` rejects.

src/c2goto/CMakeLists.txt
Yiannis128: Removed them instead (eb3feb6): c2goto's own binary dir is on its include path, so nothing is written into a directory master also writes. Caveat dropped from the description.

Yiannis128:
Added `unit/langapi/language_ui.test.cpp` in d7155db: parses a bundled source that is not on disk and checks its symbol reaches the context, plus rejection cases for an unregistered VFS path and an unopenable file. `regression/esbmc/clib_source_from_vfs` (9ff8460) covers the runtime side.

regression/esbmc/clib_source_from_vfs/main.c
```diff
@@ -0,0 +1,8 @@
+#include <string.h>
```
rafaelsamenezes: This test is useless, every other test in the CI already checks for this

rafaelsamenezes (changes requested):

regression/esbmc/clib_source_from_vfs/main.c
Yiannis128: Done

lucasccordeiro (approved):

src/c2goto/CMakeLists.txt
```diff
@@ -303,18 +355,62 @@ function(mangle_clib output)
...
+  # so a bare "/esbmc-vfs/..." would miss every lookup.
+  set(vfs_root "C:/esbmc-vfs")
+else()
+  set(vfs_root "/esbmc-vfs")
+endif()
+
+function(model_vfs_paths output)
+  set(result "")
+  foreach(f ${ARGN})
+    file(RELATIVE_PATH rel "${CMAKE_CURRENT_SOURCE_DIR}/library" "${f}")
```
rafaelsamenezes, line 386: An input outside `library/` yields a `../`-escaping path no mangle registered, surfacing only as c2goto's "failed to open input file". Fail here on a relative path starting `..`.

Yiannis128: Done in 05de9e6e68: `model_vfs_paths()` now `FATAL_ERROR`s at configure time on a relative path starting `..`, so the escape is reported where the input is named rather than as c2goto failing to open a path.

rafaelsamenezes (approved):

PR #6795, opened 2026-08-07
[python] Specialize a *args call to a nested function

### LINKED ISSUE #6741: [python] Arity validator ignores *args: spurious TypeError on any call
## Summary

`_validate_positional_call_arity` counts only named positional parameters. A function declared with `*args` is therefore treated as accepting zero positional arguments, and any call to it raises an uncaught `TypeError` traceback.

## Reproducer

```python
def byteslike(*pos, **kw):
    return len(pos)

assert byteslike(1) == 1
```

## Observed

```
  File ".../preprocessor/core_visitors_mixin.py", line 720, in _validate_positional_call_arity
    raise TypeError(
TypeError: byteslike() takes 0 positional arguments but 1 was given
```

Exit code `1`, full Python traceback. ESBMC 8.4.0.

CPython 3.12.3 runs the file to completion, exit 0.

## Cause

```python
def _validate_positional_call_arity(self, node, function_name, expected_args):
    display_name = self._display_name(function_name)
    if len(node.args) > len(expected_args):
        ...
        raise TypeError(...)
```

`expected_args` holds only the declared positional names. When the signature carries a `*args` vararg, any number of positional arguments is legal and this check must be skipped entirely (and an upper bound must not be enforced).

## Category

**ESBMC defect.** The rejected program is well-formed Python; the arity model is wrong, and the failure surfaces as an uncaught traceback rather than a diagnostic.

## Related

Same validator family as the arity-table clobbering reported separately. Prior arity bug: #3255 (closed).

## Affected CPython tests

44 of the 405 CPython 3.12.3 `Lib/test` files terminate inside the arity/default-filling validators (`_validate_positional_call_arity`, `_fill_missing_args_with_defaults`, `_validate_kwonly_args`, `_validate_duplicate_positional_keyword`). `test_io` is the direct `*args` witness above. The full set:

`test__xxinterpchannels`, `test_argparse`, `test_ast`, `test_asyncgen`, `test_binop`, `test_buffer`, `test_builtin`, `test_bytes`, `test_class`, `test_collections`, `test_context`, `test_copy`, `test_coroutines`, `test_decorators`, `test_descr`, `test_dict`, `test_dtrace`, `test_functools`, `test_http_cookiejar`, `test_httplib`, `test_interpreters`, `test_io`, `test_iter`, `test_itertools`, `test_keywordonlyarg`, `test_logging`, `test_msilib`, `test_ordered_dict`, `test_perf_profiler`, `test_repl`, `test_set`, `test_shelve`, `test_smtplib`, `test_socket`, `test_ssl`, `test_tempfile`, `test_threading_local`, `test_tracemalloc`, `test_ttk_textonly`, `test_typing`, `test_urllib2_localnet`, `test_urllib2net`, `test_weakref`, `test_with`

These 44 share the validator but not necessarily this single root cause; the `*args` case is the one minimised here.

rafaelsamenezes:
#6752 lowered `*args` by specializing each call site, but registers only defs in the module body, so a nested one still reports `Variable 'rest' is not defined`. A nested def that reads nothing its enclosing functions bind loses no scope by being hoisted, so this admits those and drops them once every call to them is specialized.

One that genuinely captures still falls back to the previous diagnostic rather than being hoisted into a wrong program.

Follow-up to #6752, which closed #6741.


lucasccordeiro:
Resolving the conflict turned out not to be mechanical, so flagging what I changed and why — this rewrote a good part of the branch, and it should be your call whether to keep it.

**What the conflict was.** `*args` landed on master while this was open, via #6752, by a different route: a `VarargMixin` that specializes each call site into a fixed-arity copy, rather than packing at the call. Both branches also added `regression/python/github_6741{,_fail}`, hence the add/add conflicts. Merging the two textually would have left two `*args` mechanisms rewriting the same AST, so I took master's wholesale and kept only what this branch adds on top.

**What this branch still contributes.** #6752 registers only defs in the module body, so `*args` in a nested function is untouched by it. Measured against master at `7a2ff5c0b3`:

```python
def outer():
    def inner(*rest):
        return len(rest)
    return inner(1, 2, 3)

assert outer() == 3
```

```
ERROR: Variable 'rest' is not defined in function 'outer@F@inner' at line 3.
```

Your `github_6741_shadowed` is exactly this case, and it fails on master today. It is not the shadowing that breaks it — a module-level `*args` def alongside a same-named-arity sibling verifies fine; it is the nesting.

**What I implemented.** #6752's docstring gives a real reason for the restriction — a specialization is injected at module level, so a nested def "would lose the scope it closes over." That holds only when it *does* close over something. The commit admits a nested def when it reads nothing its enclosing functions bind, and drops it once every call to it has been specialized, which the module-body pass cannot reach. A nested def that genuinely captures still falls back to the previous error rather than being hoisted into a wrong program; `github_6741_nested_closure` pins that boundary.

**What I dropped**, so it is not lost silently: the packing implementation in `core_visitors_mixin`, `functionVarargs`/`functionPackedVarargs`, `_scan_value_referenced_names`, and the `is_entry_module` plumbing through `parser.py` / `json_emitter.py` / `range_alias_pipeline.py`. All of it served the call-site-packing design that master's approach replaces. Your `github_6741_value_reference` is kept — master already produces exactly the `ERROR:` it expects, so it now documents a limitation both designs share.

**Checks.** 2000 of the `python/` suite pass; the three failures I saw (`github_3719_4-nondet`, `github_4149`, `github_4149_fail`) reproduce on unmodified master on the same machine and pass on the Linux CI leg, so they are pre-existing and macOS-local, not from this change. `scripts/check_python_tests.sh github_6741` is green across all five tests, and YAPF reports no diff.

Happy to revert to your original approach if you would rather take that route instead — you know the design intent here better than I do.

rafaelsamenezes:
Superseded by #6752, which landed on master and fixes #6741 by specializing each call site into a fixed-arity copy binding the vararg to a packed tuple. This PR solves the same issue with a different approach (rewriting the collector into a trailing list parameter), so the two conflict throughout `preprocessor/` and in the shared `github_6741` / `github_6741_fail` regressions.

Resolving the conflict in favour of master resets every source file this branch touches — the `is_entry_module` plumbing and `_scan_value_referenced_names` only exist to support the trailing-list lowering — leaving a merge commit made almost entirely of deletions. Closing is cleaner.

One case here is not covered by #6752: a `*args` def nested inside another function. Verified against master at 1f50436:

```python
def outer():
    def inner(*rest):
        return len(rest)
    return inner(1, 2, 3)

assert outer() == 3
```
```
ERROR: Variable 'rest' is not defined in function 'outer@F@inner'
```

The same def at module scope verifies fine, so nesting alone is enough — shadowing is not required. This is a deliberate limitation of the specialization design, documented in `vararg_mixin.py`: "Specializations are injected at module level, so only a def at that level can be copied: a nested one would lose the scope it closes over." Closing the gap means injecting the specialization into the enclosing function rather than the module body, which shares no code with this branch. Tracking separately.

The other test added here, `github_6741_value_reference`, already behaves identically on master (same converter diagnostic), so nothing is lost by closing.

Issue #6542, opened 2026-07-30
[contracts] __ESBMC_is_fresh cannot be satisfied by automatic storage: valid_object is unconstrained for stack pointers


LukeW1999:
> Narrowed on 11 August. This issue originally covered three defects in the replace-side lowering of `__ESBMC_is_fresh`. Two are fixed — aliased arguments by #6796, the discarded extent by #6919 — and the [comment recording that](https://github.com/esbmc/esbmc/issues/6542#issuecomment-5252462596) keeps the original text. What follows is the remaining one.

## Summary

A caller cannot satisfy `__ESBMC_requires(__ESBMC_is_fresh(p, n))` by passing the address of an object with automatic storage, even though that is a perfectly ordinary thing for a caller to do and the contract has no business refusing it.

The requires clause is lowered to `valid_object(p)`, which becomes `__ESBMC_alloc[POINTER_OBJECT(p)]`. That array is only ever written for heap objects. For a stack pointer the entry is unconstrained, so a solver is free to pick it false and the precondition fails.

This is not a contracts defect that happens to touch symex. It is a property of `__ESBMC_alloc` that predates contracts, and the tree has said so for years — `src/goto-symex/dynamic_allocation.cpp`:

```
// XXXjmorse - currently we don't correctly track the fact that stack
// objects change validity as the program progresses, and the solver is
// free to guess that a stack ptr is invalid, as we never update
// __ESBMC_alloc for stack ptrs.
// So, add the precondition that invalid_ptr only ever applies to dynamic
// objects.
```

`invalid_pointer2t` works around it a few lines below by conjoining `__ESBMC_is_dynamic`. The `is_fresh` lowering has no equivalent.

## Reproducer

```c
typedef struct { int coeffs[4]; } P;

void callee(P *p) {
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(P)));
  __ESBMC_assigns(p->coeffs);
  __ESBMC_ensures(p->coeffs[0] == 1);
  p->coeffs[0] = 1;
}

int main(void) {
  P v;
  callee(&v);
  return 0;
}
```

`esbmc main.c --replace-call-with-contract callee --unwind 8`:

```
Violated property:
  contract requires
  VALID_OBJECT(&v) && (!is_dynamic[POINTER_OBJECT(&v)] || ...)
```

[...]

PR #6484, opened 2026-07-28
[contracts] give pointer parameters a nondet extent in enforce harnesses

Yiannis128:
## This is a semantic change to the contract dialect, please agree to it deliberately

This is not only a bug fix. Any existing contract that dereferences a pointer parameter without `__ESBMC_is_fresh` will start reporting `VERIFICATION FAILED` where it previously passed, because the extent it was silently relying on is gone. That is the intended outcome, since the previous pass verdict did not mean what it appeared to mean, but it will affect user contracts outside this repository and not only the 14 tests migrated here.

The affected shape is a `requires` clause that constrains a pointer only with a null check, as in `__ESBMC_requires(p != 0)`, followed by a body that reads or writes through `p`. The fix is to state the extent, as in `__ESBMC_requires(__ESBMC_is_fresh(p, n * sizeof(T)))`. ESBMC names the parameter and suggests that clause in a warning, so the failure points at the fix rather than looking like a bug in the user's code, but the verdict does change.

The alternative considered was capping the assumed extent at one element instead of removing it. That was rejected because it does not close the class of bug, it only moves the boundary from index 25 to index 1: a contract saying nothing about `p` would still admit `*p`. The struct-parameter case below is that design still in place, and it is why struct parameters remain an open item rather than a solved one.

## What changed

The `--enforce-contract` harness backed every primitive pointer parameter with a fixed 100-byte heap object, so the body was checked against a buffer size the contract never stated. A contract of just `__ESBMC_requires(p != 0)` accepted a write to `p[20]` and reported `VERIFICATION SUCCESSFUL`. Bisecting the boundary showed the window was exactly 25 `int`s, which is 100 bytes divided by the element width, confirming #6215's byte-versus-element reading at the same time.

The extent is now nondeterministic, so a parameter is dereferenceable only as far as `__ESBMC_is_fresh` justifies and the solver is free to pick an extent that falsifies anything else. The object stays a single consistent SMT allocation, which is the property the fixed size was there to protect. Each extent is recorded per parameter so the Phase 2B array-element witness index is bounded by the real allocation instead of a constant.

Allocating in bytes rather than in pointed-to elements makes #6215 disappear rather than needing the multiply-by-`sizeof` fix, and it lifts the 100-element cap on `__ESBMC_assigns(arr[i])`. `arr_assigns_large_index` documented that cap as a KNOWNBUG and now passes, so it is promoted to CORE and renamed.

## Struct parameters and the #6483 blocker

[...]

PR #6796, opened 2026-08-07
[contracts] stop assuming pointer parameters do not alias

### LINKED ISSUE #6551: [contracts] the enforce harness assumes pointer parameters do not alias, so enforce+replace proves false properties
## Summary

The `--enforce-contract` harness gives each pointer parameter its own allocation, so a callee's proof may silently rely on its pointer parameters not overlapping — a hypothesis no contract clause states. `--replace-call-with-contract` never checks it. Enforcing a function and then replacing a call where two pointer arguments alias therefore proves properties that are false in the real program: a false `VERIFICATION SUCCESSFUL`.

No `__ESBMC_is_fresh` is involved. This is the general form of the aliasing symptom reported in #6542 §1, which turns out not to be an `is_fresh` defect at all.

## Reproducer

```c
#include <stdlib.h>

typedef struct
{
  int coeffs[4];
} P;

void callee(P *a, P *b)
{
  __ESBMC_requires(a != 0);
  __ESBMC_requires(b != 0);
  __ESBMC_assigns(a->coeffs);
  __ESBMC_ensures(a->coeffs[0] == 1);
  __ESBMC_ensures(b->coeffs[0] == __ESBMC_old(b->coeffs[0]));
  a->coeffs[0] = 1;
}

int main(void)
{
  P *o = (P *)malloc(sizeof(P));
  __ESBMC_assume(o != 0);
  o->coeffs[0] = 5;
  callee(o, o);            /* the same object for both parameters */
  __ESBMC_assert(o->coeffs[0] == 5, "implied by the second ensures");
  return 0;
}
```

| configuration | result |
|---|---|
| `esbmc repro.c --unwind 8` | `VERIFICATION FAILED` |
| `esbmc repro.c --function callee --enforce-contract callee --unwind 8` | `VERIFICATION SUCCESSFUL` |
| `esbmc repro.c --replace-call-with-contract callee --unwind 8` | **`VERIFICATION SUCCESSFUL`** |

The direct run is right: after `callee(o, o)` the object holds 1, not 5. The contract holds in isolation. Composing the two proves the assertion anyway.

The mechanism is the assumed `ensures`: with `a == b == o` and `__ESBMC_old(b->coeffs[0]) == 5`, assuming `a->coeffs[0] == 1` together with `b->coeffs[0] == 5` is assuming `false`, so everything after the call is discharged vacuously.

## Why the callee verifies in isolation

`add_pointer_validity_assumptions` (`src/goto-programs/contracts/contracts.cpp`) iterates the parameter list and emits an independent allocation for each pointer parameter, skipping only those already allocated by `__ESBMC_is_fresh`:

```cpp
  for (const auto &param : params)
  {
    ...
    // Skip params already allocated by __ESBMC_is_fresh to avoid overwriting
    // the is_fresh allocation with a single-element malloc.
    if (skip_params.count(param.get_identifier()))
      continue;
    ...
  }
```

Two parameters therefore always address two distinct objects in the harness, whatever the contract says. `b->coeffs[0] == __ESBMC_old(b->coeffs[0])` is provable there only because of that. The contract states nothing that would let a caller reproduce the hypothesis, and nothing asserts it at a replace site.

This is a departure from the convention the dialect otherwise follows. Neither ACSL nor CBMC assumes parameters do not alias; separation has to be stated, with `\separated` or `__CPROVER_is_fresh`.

## Direction

Two ways to close it, and I think the first is right.

**Weaken the harness.** Let pointer parameters alias unless the contract says otherwise, so the enforce proof stops consuming a hypothesis the contract has not stated. Nothing is then required of callers beyond what they already state, and the composition is sound by construction. The cost is that contracts relying on implicit separation start failing `--enforce-contract`, which is a dialect change of the same kind as #6484 and should be announced as one.

**Strengthen the call site.** Assert pairwise separation of pointer arguments when replacing. This keeps existing enforce proofs green but demands of every caller something the contract never promised, so it converts the unsoundness into false positives instead. It is also the weaker fix: it cannot express partial overlap policies that a contract might legitimately want.

Either way the check belongs at the level of *pointer parameters*, not `__ESBMC_is_fresh`. `regression/function_contract` has no test where a call site passes the same object to two pointer parameters, which is why this has gone unnoticed.

## Relationship to #6542

#6542 §1 reports the same false `VERIFICATION SUCCESSFUL` with `is_fresh` on both parameters and attributes it to the replace-side `is_fresh` lowering. That attribution is wrong; the reproducer above has no `is_fresh` at all. #6542's other two symptoms — the discarded extent and `valid_object` being unconditionally false for automatic storage — are genuinely specific to the `is_fresh` lowering and stand.

## Environment

- ESBMC 8.4.0, x86_64 Linux, `master` at `e912c6ae00`, default solver (Bitwuzla 0.9.0)
- The three configurations above were run on this build; the allocation loop was read in this checkout

LukeW1999:
Closes #6551. Addresses the first case in #6542, which stays open for the discarded extent and the automatic-storage rejection.

`--enforce-contract` in harness mode gave every pointer parameter its own allocation, so a callee's proof could rest on its parameters addressing distinct objects. No contract clause states that, and nothing checked it at a replace site, so enforcing a function and then replacing a call that passes one object twice proved properties false in the real program.

```c
void callee(P *a, P *b)
{
  __ESBMC_requires(a != 0);
  __ESBMC_requires(b != 0);
  __ESBMC_assigns(a->coeffs);
  __ESBMC_ensures(a->coeffs[0] == 1);
  __ESBMC_ensures(b->coeffs[0] == __ESBMC_old(b->coeffs[0]));
  a->coeffs[0] = 1;
}
...
  o->coeffs[0] = 5;
  callee(o, o);
  __ESBMC_assert(o->coeffs[0] == 5, "implied by the second ensures");
```

| | before | after |
| --- | --- | --- |
| direct | FAILED | FAILED |
| `--function callee --enforce-contract callee` | SUCCESSFUL | **FAILED** |

The direct run is right. With `a == b` and `__ESBMC_old(b->coeffs[0]) == 5`, assuming `a->coeffs[0] == 1` alongside `b->coeffs[0] == 5` is assuming `false`, so everything after the call was discharged vacuously.

## Weaken the harness

Taken in the direction #6551 argued for and #6484 set the precedent for: an assumption the contract does not state should not be silently available to the proof. Pointer parameters backed by the harness may now alias one another, chained so every partition of the parameters into aliasing groups is reachable. The flags are named symbols rather than inline nondets, so a counterexample says which two parameters the failing trace aliased:

```
  b = (struct P *)(&harness_ptr_callee_a)
```

A `log_warning` names the parameters that gained an aliasing choice, in the style of the existing `warn_unstated_extents`, so a contract that starts failing after this change explains itself rather than leaving a flag to be found in a trace.

Parameters covered by `__ESBMC_is_fresh` keep their separate allocations, because `is_fresh` does state separation.

## Which makes is_fresh load-bearing, so check it

That leaves `is_fresh` as the way to state separation, and the replace side never checked it — so on its own the harness change would move the hole rather than close it. A contract rejected by the new harness gets `is_fresh` added, enforce goes green, and the replace site still admits an aliasing caller.

`__ESBMC_is_fresh(p, n)` says `p` addresses a *fresh* object, so the obligation is separation from everything else the caller can reach, not merely from the other `is_fresh` parameters. Replace sites now assert it against every other pointer argument:

```
Violated property:
  contract requires: __ESBMC_is_fresh argument must not alias another pointer argument
  !(SAME-OBJECT(o, o))
```

The obligation is owed only where the requires clause asserts the `is_fresh` unconditionally. An ensures-side `is_fresh` describes what the callee produces and asks nothing of the caller, and `__ESBMC_requires(n <= 0 || __ESBMC_is_fresh(p, n))` claims nothing on the other branch; neither may become an unconditional caller obligation.

## Frame check had to learn about aliasing

Letting parameters alias made the assigns frame check report violations that are not violations. `mlk_poly_add(poly *r, const poly *b)` with `__ESBMC_assigns(r->coeffs)`, called as `add(p, p)`, writes only `r->coeffs`, which is exactly what it declares — but the check snapshots `b->coeffs` and asserts it unchanged, and under `r == b` that is the same memory.

The snapshot assertion now carries an exemption for the case where the snapshotted parameter aliases an assigns target's base. Three things keep it honest:

- It is matched on the field, so a sibling field stays protected. `__ESBMC_assigns(r->coeffs)` with a body writing `r->other` still reports `assigns compliance: b->other not in assigns clause`.
- `__ESBMC_assigns(*r)` covers the whole pointee, so it exempts every field. Whole-object assigns is the natural way to write this and would otherwise have produced immediate false positives.
- The base is read in the **pre-state**. The exemption is asserted after the call but guards a value snapshotted before it, so a callee free to assign a global base could otherwise point it at the checked object on the way out and launder an out-of-frame write.

## Freeing what was allocated

Aliasing breaks the assumption that the lvalue which received an allocation still identifies it at wrapper exit. Both allocation paths now free a snapshot taken at allocation time: the pointer-parameter malloc, and the `__ESBMC_is_fresh` allocations, where two parameters aliasing makes `a->p` and `b->p` one lvalue, the second malloc overwrite the first, and freeing the lvalue a double free of one object while the other leaks.

## Tests

Eight added under `regression/function_contract`, and `assume_nonnull_valid_two_ptrs_pass` renamed to `is_fresh_two_ptrs_pass`. Its contract was unsound as written: `p->x = 10; q->x = 20;` with `p == q` leaves `p->x == 20`, so the first ensures fails. Non-nullness was never enough to make it provable, and its comment claimed the separation the harness was silently providing. It now states separation with `is_fresh`, which is what its name says.

The new tests pin the message as well as the verdict, so a harness-internal failure cannot keep them green:

| test | pins |
| --- | --- |
| `github_6551_param_alias_enforce_fail` | the reproducer's enforce side now fails |
| `github_6542_is_fresh_alias_replace_fail` | `is_fresh` pair, aliasing caller, rejected |
| `github_6542_is_fresh_vs_plain_ptr_fail` | `is_fresh` against a *plain* pointer argument |
| `github_6542_is_fresh_distinct_replace_pass` | positive control, distinct objects compose |
| `github_6551_alias_sibling_field_fail` | how narrow the frame exemption is |
| `github_6551_whole_object_assigns_pass` | `assigns(*r)` exempts every field |
| `github_6551_alias_base_launder_fail` | the base is read in the pre-state |
| `github_6551_cond_is_fresh_pass` | a conditional `is_fresh` binds no caller |

398/398 pass across `function_contract` and `loop-invariants`. Of the nine `--function` tests with two or more uncovered pointer parameters that #6551 predicted this could reach, only the one above needed changing.

## This is a dialect change and needs announcing

Same kind as #6484, and it should be announced the same way rather than landed quietly. Contracts that rely on parameters not overlapping start failing `--enforce-contract`, and the fix for each is to state the separation with `__ESBMC_is_fresh` or `__ESBMC_requires(p != q)`. The `log_warning` names the parameters so a contract that starts failing explains itself, but that is a mitigation, not a substitute for the announcement.

I have not written that announcement here: where it goes and how it is worded is a maintainer call, and I would rather it were decided than assumed.

## Not in this PR

- **The hole is narrowed, not closed.** Aliasing is offered only between parameters of identical type, so `void *` and byte-buffer parameters — pervasive in the PQC code this targets — still carry an unstated, unchecked separation hypothesis. So do interior aliases (`f(&s->a, &s->b)`, `f(buf, buf + 1)`), and parameter-to-global aliasing. `const` is not a barrier: `migrate_type` drops it, so the `mlk_poly_add(poly *r, const poly *a, const poly *b)` shape is covered.
- `__ESBMC_requires(p != q)` is a sound, checked escape hatch in both directions and carries none of the above.
- The other two cases in #6542, the discarded extent and `valid_object` being unconditionally false for automatic storage, are untouched.
- Found while testing, unrelated and reproducing on 8.2.0: `__ESBMC_assigns(r->coeffs[0])` does not catch a write to `r->coeffs[1]`. It reproduces with a single pointer parameter, where none of this PR's machinery applies. Filing separately.


LukeW1999:
@lucasccordeiro requesting your review. CI is green across all ten checks.

Three things worth knowing before reading the diff, since this is a soundness fix with a behaviour change attached.

**It rejects more than it used to, and that is the point.** The harness was granting callees a separation hypothesis no contract states, which `--replace-call-with-contract` then never checked, so enforce plus replace proved properties that are false in the real program. Closing that means contracts relying on the unstated assumption now fail `--enforce-contract`. In `regression/function_contract` exactly one test needed changing, and its contract was unsound as written.

**It is a dialect change and I have deliberately not written the announcement.** Where that goes and how it is worded is your call, not something I should decide unilaterally. The reference doc is updated in the same commit, since it currently describes the old behaviour.

**The hole is narrowed, not closed, and the description says so.** Aliasing is offered only between parameters of identical pointer type, so `void *`, byte buffers, interior aliases and parameter-to-global remain assumed separate. I would rather land the narrowing with the residual documented than hold it for a complete fix.

Two related issues are open and not addressed here: #6542's discarded extent and automatic-storage rejection, and #6798, a `goto-symex` join defect where a heap object written on one side of a branch takes that value on both. #6798 reaches contracts through `__ESBMC_is_fresh` but is not a contracts bug; the four-line reproducer uses plain `malloc` and no contract syntax at all.

LukeW1999:
Rebased onto `9a3d7e8a6c`. It had gone `dirty` against the contract work that landed since, and three of the conflicts were with your changes, so here is what I did with each rather than leaving you to read it out of the diff.

**#6787, the unstated-extent warning.** You gated `nondet_extent.push_back(name)` behind `param_extent_is_observable`. I had replaced the line above it, `allocated_ptrs.push_back(p)`, with a snapshot of the allocation. Kept both: the snapshot for the free, your gate for the warning.

```cpp
    allocated_ptrs.push_back(
      retain_allocation_for_free(wrapper, p, name, func, location));
    // The storage is allocated either way; only the advice is withheld, and
    // only when nothing here can observe the extent (#6511).
    if (param_extent_is_observable(func, param.get_identifier()))
      nondet_extent.push_back(name);
    aliasable_params.emplace_back(p, name);
```

**#6788, `materialize_ptr_deref_array_field`.** Purely positional — your new function and my `assigns_target_base` were extracted to the same place. Both kept.

**Header declarations.** Same, both kept.

Re-verified after the rebase: 407/407 across `function_contract` and `loop-invariants`, complexity gate clean, `git-clang-format` clean on the changed lines, and the twelve behaviours this PR asserts all unchanged.

Worth noting one interaction, since #6805 landed in the middle of this. That fix removes the reason a contract needed a constant extent to be usable, so `__ESBMC_is_fresh(p, n)` with a computed `n` now works. This PR makes `is_fresh` the way to state separation, so the two compose better than either does alone: a contract can now state both the extent and the separation symbolically.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2670,6 +2736,53 @@ void code_contractst::materialize_ptr_deref_array_field(
...
+/// Should per-index checking of the declared base ever be tightened, this
+/// stripping has to be revisited with it.
+static expr2tc assigns_target_base(const expr2tc &target, const irep_idt &field)
+{
+  expr2tc e = target;
+  while (is_typecast2t(e))
+    e = to_typecast2t(e).from;
+
+  if (is_symbol2t(e) && is_pointer_type(e->type))
+    return e;
```
lucasccordeiro: `__ESBMC_assigns(p)` names the pointer, not the pointee, so exempting every field of `*p` waives frame checks the clause never granted. Require the dereference form.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -72,6 +72,37 @@ static bool is_fresh_function(const std::string &funcname)
...
+           asserted_unconditionally(to_and2t(e).side_2, name);
+
+  // A conjunct asserts the temp only by being it. Merely mentioning it, say
+  // as an operand of a comparison, says nothing about whether the clause
+  // demands it, so recursing into other node kinds would over-approximate.
+  // The frontend may wrap the temp in a cast or compare it against zero.
+  expr2tc leaf = e;
+  while (is_typecast2t(leaf))
+    leaf = to_typecast2t(leaf).from;
+  if (is_notequal2t(leaf) && is_constant_number(to_notequal2t(leaf).side_2))
```
lucasccordeiro: Only `temp != 0` is recognised; `0 != temp` falls through and the is_fresh silently stops stating separation. Match the constant on either side.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2888,9 +3001,77 @@ code_contractst::materialize_ptr_deref_snapshots(
...
+      expr2tc base = assigns_target_base(target, snap.field_name);
+      if (is_nil_expr(base) || base == snap.ptr_sym)
+        continue;
+
+      // The exemption is asserted after the call but guards a value
+      // snapshotted before it, so the base has to be read in the pre-state
+      // too. A callee free to assign a global base could otherwise point it
+      // at the checked object on the way out and launder a write that was
+      // outside the frame when it happened.
+      std::string base_name = "__ESBMC_frame_aliasbase_" + func_name + "_" +
```
lucasccordeiro: A base snapshot symbol plus DECL/ASSIGN is emitted per (snapshot, target) pair, duplicating them for every repeated base. Hoist one snapshot per distinct base.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3969,6 +4153,156 @@ void code_contractst::replace_calls(const std::set<std::string> &to_replace)
...
+  std::vector<expr2tc> obligations;
+
+  for (size_t i : fresh_params)
+  {
+    if (i >= actual_args.size())
+      continue;
+
+    for (size_t j = 0; j < params.size() && j < actual_args.size(); ++j)
+    {
+      if (j == i || !is_pointer_type(migrate_type(params[j].type())))
```
lucasccordeiro, line 4289: Replace demands separation from every pointer parameter, but enforce only aliases identical types, so the two sides disagree on differently-typed parameters. Align the type test.

lucasccordeiro (changes requested):
No test covers `__ESBMC_requires(p != q)`, the separation remedy the new docs recommend. Add one.

LukeW1999:
All five addressed. Taking them in the order you raised them.

**`__ESBMC_assigns(p)` naming the pointer.** Changed: only `*p` and `p->field` now match, a bare `p` grants no exemption. Worth recording that I had evidence pointing the other way and it did not settle the question — `assigns(r)` with a body writing `r->b` verifies today on a single pointer parameter, so the current behaviour does treat it as the pointee. That shows what the code does, not what the clause means, and yours is the conservative reading: fewer exemptions, more frame checks kept.

**`0 != temp`.** Real defect, and worse than untidy. It did not merely miss a form, it made the separation obligation disappear:

```
__ESBMC_requires(0 != __ESBMC_is_fresh(a, sizeof(P)));   →  SUCCESSFUL   (wrong)
__ESBMC_requires(__ESBMC_is_fresh(a, sizeof(P)));        →  FAILED
```

Now matched on either side; the first form reports FAILED.

**Duplicated base snapshots.** Hoisted to one per distinct base, keyed on the symbol. The names went from a sequence number to the parameter's pretty name while I was in there, since the old ones were unreadable in a trace.

**Aligning the type test.** Took the direction that closes the gap rather than the one that hides it: pointee type is no longer a barrier, so any two pointer parameters may alias, with a typecast on the assignment. Two parameters of different pointer type can address the same storage, and a contract needing them apart has to say so like any other. This also removes the residual the description previously listed for `void *` and byte buffers, which are exactly the shapes the PQC code this targets is full of.

I expected that to cost something and it did not: **409/409**, nothing changed. The case it closes:

```c
void callee(P *a, void *b) {          /* was: enforce granted separation free */
  __ESBMC_requires(a != 0);
  __ESBMC_requires(b != 0);
  __ESBMC_assigns(a->coeffs);
  __ESBMC_ensures(a->coeffs[0] == 1);
  __ESBMC_ensures(((P*)b)->coeffs[0] == __ESBMC_old(((P*)b)->coeffs[0]));
  a->coeffs[0] = 1;
}
```
`direct` FAILED, `--enforce-contract` now FAILED too; it was SUCCESSFUL before.

**A test for `requires(p != q)`.** You are right that recommending it in the docs while testing neither side was not good enough. Added both: `github_6551_requires_ptr_neq_pass` for enforcement accepting it, and `github_6551_requires_ptr_neq_replace_fail` for a replace site making the caller discharge it.

409/409 across `function_contract` and `loop-invariants`, complexity gate clean, `git-clang-format` clean. Only interior aliases (`f(&s->a, &s->b)`, `f(buf, buf + 1)`) and parameter-to-global remain assumed apart; the description says so and no longer claims the type restriction.

LukeW1999:
One more, unprompted, because your fourth point taught me something about my own reasoning.

I had listed two residuals as unavoidable. You asked about one of them, the identical-type restriction, and closing it cost nothing. So I went back and tested the other, which I had also declared without measuring: `__ESBMC_is_fresh` on anything that is not a bare parameter — a global, `s->p`, `*out`. The enforce harness allocates those on their own exactly as it does a parameter, but the replace side keyed its obligation on a parameter position and so emitted nothing for them.

```c
P *g;
void callee(P *a) {
  __ESBMC_requires(__ESBMC_is_fresh(a, sizeof(P)));
  __ESBMC_requires(__ESBMC_is_fresh(g, sizeof(P)));
  __ESBMC_assigns(a->coeffs);
  __ESBMC_ensures(a->coeffs[0] == 1);
  __ESBMC_ensures(g->coeffs[0] == __ESBMC_old(g->coeffs[0]));
  a->coeffs[0] = 1;
}
/* caller: g = o; callee(o); */
```

`direct` FAILED, `--replace-call-with-contract` was SUCCESSFUL and is now FAILED. Obligations for these are keyed on the rebound expression instead, against every pointer argument and against each other. Again no cost: 409/409.

That leaves only interior aliases and parameter-to-global as residuals, and both are genuine, not untested assumptions.

The pattern is worth naming since it cost you a review round: twice I wrote "this cannot be closed" where I meant "I did not try". Both turned out free. I have gone through the rest of the diff for the same shape — restrictions I added without measuring, one-sided matches on commutative operators, and per-pair emission where per-distinct-item would do — and these were the instances.

LukeW1999:
Went through the two remaining residuals properly rather than asserting them, and one more closed. Recording the method as well as the result, since the first attempt at this measured the wrong thing.

**Parameter versus global was real, and is now closed.** `__ESBMC_is_fresh(p, n)` says `p` addresses a fresh object, so it is separate from everything the caller can reach; the obligation ranged only over pointer arguments, and a global is not one.

```c
P g;
void callee(P *p) {
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(P)));
  __ESBMC_assigns(p->x);
  __ESBMC_ensures(p->x == 1);
  __ESBMC_ensures(g.x == __ESBMC_old(g.x));
  p->x = 1;
}
/* caller: g.x = 5; callee(&g); */
```

`direct` FAILED, `enforce` SUCCESSFUL, `replace` was SUCCESSFUL and is now FAILED. Statics named in the requires or ensures clauses are collected and each fresh pointer is asserted separate from them. Only clause-mentioned globals, since those are the ones the contract can be relying on. 410/410, again no cost.

**Interior aliasing needed a better test than the one I first wrote.** My initial case reported `enforce` FAILED and I nearly recorded it as covered. It was failing on `array bounds violated`, the #6484 unstated-extent path, and `master` failed it identically — the aliasing never entered into it. With the extents stated:

```c
void callee(S *p, int *q) {
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(S)));
  __ESBMC_requires(__ESBMC_is_fresh(q, sizeof(int)));
  __ESBMC_assigns(p->b);
  __ESBMC_ensures(p->b == 1);
  __ESBMC_ensures(*q == __ESBMC_old(*q));
  p->b = 1;
}
/* caller: callee(&s, &s.b); */
```

`replace` FAILED, correctly: the separation assertion is `same_object`, which is object-level, so it catches `&s` against `&s.b` even though the offsets differ and the harness only ever offers exact equality. So overlap is covered wherever `is_fresh` is used, and since #6484 a contract cannot dereference a parameter without stating an extent. What remains is two non-fresh parameters that partially overlap, which the description now says instead of the broader claim it made before.

That is the last of them. Everything previously listed as a residual was either closable at no cost or was not a residual.

Issue #6797, opened 2026-08-07
[contracts] a requires clause constraining an __ESBMC_is_fresh object's contents is lost across a conditional write


LukeW1999:
## Summary

`__ESBMC_is_fresh` allocates the object but never writes an initial value through the pointer. A conditional write to that object then creates an ITE phi node with no initial SSA version to join against, and the not-taken arm does not carry the requires clause's assumption. A precondition constraining the object's contents is silently lost, and an `ensures` resting on it fails.

The same contract passes when the pointer's validity is stated as `p != 0` instead, because that path is backed by `emit_struct_stack_backing`, which does emit the initial write.

## Reproducer

```c
typedef struct { int m; int p; } T;

void f(T *t, int v)
{
  __ESBMC_requires(__ESBMC_is_fresh(t, sizeof(T)));
  __ESBMC_requires(t->p == 0);
  __ESBMC_requires(v == t->m);
  __ESBMC_assigns(t->p);
  __ESBMC_ensures(t->p == 0);
  if (v != t->m) { t->p = 1; }   /* unreachable under the requires */
}

int main(void) { return 0; }
```

`--function f --enforce-contract f --unwind 6`:

| how the pointer's validity is stated | result |
| --- | --- |
| `__ESBMC_requires(__ESBMC_is_fresh(t, sizeof(T)))` | **FAILED**, `contract ensures / t->p == 0` |
| `__ESBMC_requires(t != 0)`, nothing else changed | SUCCESSFUL |

The branch cannot be taken under `v == t->m`, so `t->p` keeps the value the requires clause gave it and the ensures holds.

**Control.** Dropping `__ESBMC_requires(t->p == 0)` makes both forms fail, correctly: nothing then establishes the postcondition. So this is not the harness rejecting everything.

## What isolates it

Same contract throughout, only the body varying:

| body | phi? | result |
| --- | --- | --- |
| empty | no | SUCCESSFUL |
| `if (0) { t->p = 1; }` | no, statically dead | SUCCESSFUL |
| `t->p = 0;` unconditional | no | SUCCESSFUL |
| `if (v != t->m) { t->p = 1; }` | **yes** | **FAILED** |
| `t->p = t->p; if (v != t->m) { t->p = 1; }` | yes, but an initial version exists | **SUCCESSFUL** |

The last row is the diagnostic one: hand-writing the initial version makes the failure go away.

## Root cause

[...]

PR #6811, opened 2026-08-08
[contracts] give a fresh object an initial version

### LINKED ISSUE #6798: [goto-symex] a conditionally-written heap object is not merged at the join, proving false properties
## Summary

A heap object that is written on only one side of a branch takes the written value on **both** sides. Reads after the join see the branch's value unconditionally, so a false property is proved. No contracts, no intrinsics, four lines.

## Reproducer

```c
#include <stdlib.h>

typedef struct { int p; } T;

int main(void)
{
  T *t = malloc(sizeof(T));
  if (!t) return 0;

  int v = nondet_int();
  if (v != 0) { t->p = 1; }

  __ESBMC_assert(t->p == 1, "false on the v == 0 path");
  return 0;
}
```

`esbmc repro.c --unwind 4` reports **`VERIFICATION SUCCESSFUL`**. On the `v == 0` path nothing writes `t->p`, so the assertion is false and the property should fail.

Replacing the allocation with a stack object, changing nothing else, reports `VERIFICATION FAILED` correctly:

```c
  T s; T *t = &s;
```

| storage | result |
| --- | --- |
| `malloc` | **SUCCESSFUL** (wrong) |
| stack | FAILED (right) |

Not a regression. Reproduces identically on 7.11.0, 8.0.0, 8.2.0 and current master.

## Where it comes from

`phi_function` in `src/goto-symex/symex_goto.cpp:435`:

```cpp
// If the variable was deleted in this branch, don't create an assignment for it
if (merge_variables.find(variable) == merge_variables.end())
  continue;
```

`merge_variables` is a clone of level2 taken at the branch point. A dynamic object that has never been assigned has no level2 record, so it is absent, the phi is skipped entirely, and the version assigned inside the branch stays current unconditionally. In the failing SSA there is no second version of the object at all.

The comment's premise looks stale. `renaming::level2t::remove()` (`src/goto-symex/renaming.h:243`) has no callers in goto-symex, and scope exit goes through `retire()` (`renaming.h:256`), which keeps the record and bumps the count. So the guard rarely fires on "deleted" and mostly fires on "never assigned", which is exactly when the phi is needed.

A stack object is unaffected because `DECL` creates the level2 record before the branch.

## Confirming the mechanism

Touching the object once anywhere before the branch is enough, and it need not be the location that is conditionally written:

```c
  *(unsigned char *)t = *(unsigned char *)t;   /* one byte */
  if (v != 0) { t->q = 1; }                    /* a different field */
  __ESBMC_assert(t->q == 1, "...");            /* now correctly FAILED */
```

So versions are per object symbol, and one version is all that is missing.

## Suggested fix

`goto_symext::symex_decl` (`src/goto-symex/symex_other.cpp:96-104`) already has the idiom: a dummy `cur_state->level2.make_assignment(l1_sym, expr2tc(), expr2tc())` bumps the version without emitting an assignment. Doing the same in `goto_symext::symex_mem` (`src/goto-symex/builtin_functions/memory_alloc.cpp`, after the dynamic object symbol is added) would give every fresh dynamic object a level2 record at creation. Reads would still yield a free variable, so semantics are unchanged, but `phi_function` would stop skipping.

The alternative is to revisit the early `continue` in `phi_function` itself, which would also cover objects allocated inside a branch, but that is a hotter path.

## Please expect regression fallout

Any existing test that proves something through a conditionally-written heap object is currently passing unsoundly, so a fix will turn some of them red. Those failures would be correct, but they need triage rather than reverting.

## How this was found

Chasing a contracts symptom, `__ESBMC_is_fresh` losing a `requires` clause's constraint across a conditional write (#6797). That turned out to be this bug reached through `__ESBMC_is_fresh`'s allocation rather than anything contract-specific. #6797 should probably be closed as a duplicate of this once that is confirmed.

LukeW1999:
Works around #6797. Independent of #6796 — this applies to `master` on its own, and the two can land in either order.

`__ESBMC_is_fresh(p, n)` allocates the object and never writes through the pointer, so it has no level-2 version. `phi_function` skips such an object at a join, so a write on one side of a branch stays current on both: the write escapes its guard, and a precondition constraining the object's contents is lost.

```c
typedef struct { int m; int p; } T;

void f(T *t, int v)
{
  __ESBMC_requires(__ESBMC_is_fresh(t, sizeof(T)));
  __ESBMC_requires(t->p == 0);
  __ESBMC_requires(v == t->m);
  __ESBMC_assigns(t->p);
  __ESBMC_ensures(t->p == 0);
  if (v != t->m) { t->p = 1; }   /* unreachable under the requires */
}
```

| how the pointer's validity is stated | before | after |
| --- | --- | --- |
| `__ESBMC_is_fresh(t, sizeof(T))` | **FAILED** | SUCCESSFUL |
| `t != 0`, nothing else changed | SUCCESSFUL | SUCCESSFUL |

The branch cannot be taken under `v == t->m`, so `t->p` keeps the value the requires clause gave it. The `t != 0` path was never affected, because it is backed by a `DECL` and therefore has a version.

## The change

One self-assigned byte, emitted after the non-null assume:

```c
expr2tc byte = dereference2tc(
  get_uint8_type(), typecast2tc(pointer_type2tc(get_uint8_type()), ptr_var));
init_inst->code = code_assign2tc(byte, byte);
```

Three properties worth stating, because I got each of them wrong in an earlier attempt:

- **One byte is enough.** Versions are per object symbol, not per location, so touching any byte gives the whole object a version. Verified: a conditional write to `t->q` is fixed by touching `t->m`, and in an array case a conditional write to `a[5]` is fixed by touching `a[0]`.
- **It is therefore O(1) in the extent.** A 2048-byte pointee costs the same as a 4-byte one. `regression/loop-invariants/quantified_array_invariant`, whose pointee is `struct { int16_t e[1024]; }` and which exists to show that cost stays independent of N, runs in 0.28s. Writing the whole pointee instead made that test take 446s, which is what made me look for the cheaper form.
- **A self-assignment, not a havoc.** It supplies a version without changing a value, so it cannot mask anything. It has to follow the non-null assume, or it writes through a pointer symex may still believe is NULL.

Skipped when the extent is symbolic, since it may be zero and the write would then be the out-of-bounds access the harness exists to avoid. `enforce_tlv_mid_pass` (`is_fresh(&buf, buf_len)`) and `cpp_is_fresh_member_pass` are the cases in the tree; both keep the old behaviour, so #6798 survives for them.

## This is a workaround, and the real defect is worse

The underlying bug is `phi_function` in `src/goto-symex/symex_goto.cpp` skipping objects with no prior level-2 record. It is not contract-specific. Filed as #6798 with a four-line reproducer using plain `malloc` and no contract syntax at all, which reports `VERIFICATION SUCCESSFUL` for a property that is false. It reproduces on 7.11.0, 8.0.0, 8.2.0 and master.

I tried fixing it there. It works, and `regression/function_contract` plus `regression/loop-invariants` stay green, but `humaneval_152` goes from 763 MB to 7.75 GB peak and dies on the harness's 8 GB limit. ESBMC has no field-sensitive merging, so every dynamic object gets a whole-byte-array ITE at every join. That needs design work rather than a two-line change, and the measurements are recorded on #6798 so whoever takes it does not repeat them.

So this PR does not claim to fix #6798. It makes `__ESBMC_is_fresh` usable in the meantime, at the cost of one instruction per allocation.

## Tests

`github_6797_is_fresh_initial_version_pass` is the reproducer. `github_6797_is_fresh_no_precondition_fail` is its control: with no requires establishing `t->p`, nothing makes the ensures hold and it must still be rejected — the fix has to restore a lost precondition, not invent one.

397/397 pass across `function_contract` and `loop-invariants`.

### DIFF
--- a/regression/function_contract/github_6797_is_fresh_initial_version_pass/main.c
+++ b/regression/function_contract/github_6797_is_fresh_initial_version_pass/main.c
@@ -1,0 +1,19 @@
+/* github_6797_is_fresh_initial_version_pass:
+ * __ESBMC_is_fresh allocated the object without ever writing through the
+ * pointer, so it had no level-2 version. symex skips the phi at a join for
+ * such an object, leaving the branch's version current on both sides, so the
+ * conditional write below escaped its guard and the precondition on t->p was
+ * lost. The branch cannot be taken under v == t->m, so t->p keeps the value
+ * the requires clause gave it and the ensures holds. */
+typedef struct { int m; int p; } T;
+
+void f(T *t, int v)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(t, sizeof(T)));
+  __ESBMC_requires(t->p == 0);
+  __ESBMC_requires(v == t->m);
+  __ESBMC_assigns(t->p);
+  __ESBMC_ensures(t->p == 0);
+  if (v != t->m) { t->p = 1; }
+}
+int main(void) { return 0; }
--- a/regression/function_contract/github_6797_is_fresh_initial_version_pass/test.desc
+++ b/regression/function_contract/github_6797_is_fresh_initial_version_pass/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--function f --enforce-contract f --unwind 6
+^VERIFICATION SUCCESSFUL$
--- a/regression/function_contract/github_6797_is_fresh_no_precondition_fail/main.c
+++ b/regression/function_contract/github_6797_is_fresh_no_precondition_fail/main.c
@@ -1,0 +1,16 @@
+/* github_6797_is_fresh_no_precondition_fail:
+ * The control for github_6797_is_fresh_initial_version_pass. With no requires
+ * establishing t->p, nothing makes the ensures hold on the path where the
+ * branch is not taken, so it must still be rejected. Giving the object an
+ * initial version must fix the lost precondition, not invent one. */
+typedef struct { int m; int p; } T;
+
+void f(T *t, int v)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(t, sizeof(T)));
+  __ESBMC_requires(v == t->m);
+  __ESBMC_assigns(t->p);
+  __ESBMC_ensures(t->p == 0);
+  if (v != t->m) { t->p = 1; }
+}
+int main(void) { return 0; }
--- a/regression/function_contract/github_6797_is_fresh_no_precondition_fail/test.desc
+++ b/regression/function_contract/github_6797_is_fresh_no_precondition_fail/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--function f --enforce-contract f --unwind 6
+^VERIFICATION FAILED$
+^.*contract ensures.*$
--- a/src/goto-programs/contracts/contracts.cpp
+++ b/src/goto-programs/contracts/contracts.cpp
@@ -1305,6 +1305,10 @@ goto_programt code_contractst::generate_checking_wrapper(
       assume_nn->guard = notequal2tc(ptr_var, gen_zero(ptr_var->type));
       assume_nn->location = location;
       assume_nn->location.comment("__ESBMC_is_fresh: pointer is non-null");
+
+      // Must follow the non-null assume, or this writes through a pointer
+      // symex may still believe is NULL.
+      emit_is_fresh_initial_version(wrapper, ptr_var, size_expr, location);
     };
 
   // is_fresh pointers whose lvalue reads through another pointer — e.g. a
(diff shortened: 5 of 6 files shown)


LukeW1999:
Superseded by #6805, closing.

@lucasccordeiro's fix is in the right layer and strictly better than this one. I had looked at `symex_mem` and talked myself out of it, on the reasoning that giving the object a record at allocation would produce the same phis and therefore the same cost as removing the skip in `phi_function`. That was wrong, and the part I missed is the one your comment states: `get_ident_name` numbers a count-0 record exactly as it numbers an absent one, so `declare()` costs no renumbering. My attempt at the goto-symex layer took `humaneval_152` from 763 MB to 7.75 GB peak; #6805 leaves it at 764 MB and runs faster than before.

It is also strictly wider than this PR. This one only handled a constant extent, because the write it emitted could go out of bounds when the extent may be zero. #6805 has no such gate, and a symbolic-extent contract of the same shape now passes where this branch still fails:

```c
void f(T *t, int v, unsigned n) {
  __ESBMC_requires(__ESBMC_is_fresh(t, n));
  __ESBMC_requires(n >= sizeof(T));
  __ESBMC_requires(t->p == 0);
  __ESBMC_requires(v == t->m);
  __ESBMC_assigns(t->p);
  __ESBMC_ensures(t->p == 0);
  if (v != t->m) { t->p = 1; }
}
```

The two regression tests here are redundant with `regression/function_contract/github_6797`, so nothing is lost by dropping them.

Thanks for turning it around overnight and for putting it where it belonged rather than taking a workaround in the contracts layer.

PR #6815, opened 2026-08-08
[solver] Stop array_conv aborting on a re-encoded or index-free equality

### LINKED ISSUE #6794: Bitwuzla aborts in array_conv on --enforce-contract with an array field
`--enforce-contract` on a function whose pointer parameter points at a struct with an array field aborts inside `array_conv` under Bitwuzla. Z3 verifies the same input successfully.

The two in-tree tests that cover this shape fail on macOS for this reason, while passing on the Linux CI leg.

### Reproducer

`regression/function_contract/enforce_ptr_param_array_field_pass/main.c`, unmodified:

```c
#include <stdint.h>
#include <stddef.h>
#define N 8
typedef struct { int16_t coeffs[N]; } poly;
void add(poly *r, const poly *b)
{
  __ESBMC_requires(r != NULL && b != NULL);
  __ESBMC_assigns(r->coeffs);
  __ESBMC_ensures(1);
  for (unsigned i = 0; i < N; i++)
    r->coeffs[i] = (int16_t)(r->coeffs[i] + b->coeffs[i]);
}
```

```
$ esbmc main.c --enforce-contract add --function add --unwind 9 --bitwuzla
Assertion failed: (start_pos < a1.size()), function add_array_equality, file array_conv.cpp, line 1052.

$ esbmc main.c --enforce-contract add --function add --unwind 9 --z3
VERIFICATION SUCCESSFUL
```

Without an explicit solver flag the run defaults to Bitwuzla and aborts.

`enforce_ptr_param_array_field_fail` aborts identically.

### The assertion

`src/solvers/smt/array_conv.cpp:1052`, in `add_array_equality`:

```cpp
const ast_vect &a1 = array_valuation[arr1_id][arr1_update];
const ast_vect &a2 = array_valuation[arr2_id][arr2_update];
...
assert(start_pos < a1.size());
```

So the caller passes a `start_pos` at or past the end of the recorded valuation vector for that array. Being an `assert`, a build with `NDEBUG` set would instead run the following loop, which is a no-op at `start_pos >= a1.size()` — the equality would be silently dropped rather than encoded.

### Isolation

1. **Not a regression from any in-flight PR.** Reproduced on unmodified master (`85bf9780ae`) with a clean working tree. Verified the binary was built from pristine sources by checking that a fix known to be absent from master is in fact absent from it.
2. **Not caused by the contracts changes under review.** First seen while testing #6788; reverting `src/goto-programs/contracts/contracts.{cpp,h}` to master's copies left the abort unchanged.
3. **Solver-dependent.** `--z3` verifies the same input; only `--bitwuzla` aborts.
4. **Passes on the Linux CI leg.** Both tests pass there in 0.42s — run 31132763053, job 92725364837, 2026-08-07. The Linux job is built by the same `scripts/build.sh -b DebugOpt -e ON` path with both solvers enabled, so the build type and the available solver set match the local one.

### What is not established

Why the Linux leg differs. I could not confirm which solver the Linux run selected for these two tests, so "Linux passes" is not by itself evidence that Linux passes *under Bitwuzla*. The remaining candidates are the host architecture (aarch64 vs x86_64) and the Bitwuzla build each leg links against. Someone able to run `--bitwuzla` on x86_64 Linux can settle this in one command.

### Environment

- ESBMC 8.4.0, 64-bit aarch64 macOS, master `85bf9780ae`
- `CMAKE_BUILD_TYPE=DebugOpt`, `ENABLE_BITWUZLA=On`, `ENABLE_Z3=On` — the same build type CI's PR legs use, so asserts are enabled in both

lucasccordeiro:
mk_array_equality re-runs add_array_equalities over every equality at the level,
so an already-encoded one was encoded again and tripped the result assertion. An
equality whose arrays are never indexed also tripped a strict start bound, though
an empty conjunction is exactly the right encoding for it.

Fixes #6794

### DIFF
--- a/src/solvers/smt/array_conv.cpp
+++ b/src/solvers/smt/array_conv.cpp
@@ -983,6 +983,13 @@ void array_convt::add_array_equalities()
 
   for (auto &it = pair.first; it != pair.second; it++)
   {
+    /* mk_array_equality() runs this over every equality at the level each time
+     * it records a new one, and it hands the caller the AST from the first
+     * encoding -- so re-encoding cannot reach the caller and only trips the
+     * assertion below (#6794). */
+    if (it->second.result != NULL)
+      continue;
+
 #ifndef NDEBUG
     assert(array_indexes_are_same(
       array_indexes[it->second.arr1_id], array_indexes[it->second.arr2_id]));
@@ -1049,7 +1056,10 @@ void array_convt::add_array_equality(
   const ast_vect &a2 = array_valuation[arr2_id][arr2_update];
 
   smt_solver_baset::ast_vec lits;
-  assert(start_pos < a1.size());
+  /* Equal is admissible: an equality between arrays that no index is ever read
+   * from constrains nothing, so the loop below is empty and make_n_ary_and
+   * returns true. A strict bound aborted on that case instead (#6794). */
+  assert(start_pos <= a1.size());
   for (unsigned int i = start_pos; i < a1.size(); i++)
   {
     lits.push_back(a1[i]->eq(ctx, a2[i]));


lucasccordeiro:
Superseded by #6801, which fixes #6794 at the source by discharging a reflexive array equality instead of relaxing the assertions downstream. Verified on current master: enforce_ptr_param_array_field_pass and _fail both run to a verdict under Bitwuzla with no abort.

Issue #6804, opened 2026-08-07
False SUCCESSFUL: write through a pointer with a top value set is lost


rafaelsamenezes:
A write through a pointer whose value set is non-exhaustive (contains `unknown`)
is absorbed by the failed symbol `dereference()` builds as a fallback, so no real
object is updated and no property reports it. Asserting the write did *not*
happen is then proved.

```c
#include <stdint.h>
#include <assert.h>
int main(void) {
  struct S { int x; } s = { .x = 42 };
  uintptr_t u = (uintptr_t)&s;
  u *= 2;
  u -= (uintptr_t)&s;   /* u == (uintptr_t)&s */
  int *p = (int *)u;
  *p = 3;               /* really writes s.x */
  assert(s.x == 42);    /* false in C */
}
```

`VERIFICATION SUCCESSFUL` under both Bitwuzla and Z3 (8.4.0). gcc aborts on the
assertion at `-O0` and `-O2`. Replacing the multiplication with an additive round
trip (`u += 8; u -= 8`) reports `FAILED` on the identical assertion, so the
verdict turns on the operator rather than on the program's semantics.

Mechanism:

1. `get_value_set_rec` records `unknown` (`value_set.cpp:776-784`).
2. `dereference()` sets `known_exhaustive = false` and builds
   `make_failed_symbol(type)` (`dereference.cpp:564-570`).
3. The `unknown` entry reaches `deref_invalid_ptr` and returns nil, contributing
   no real target (`dereference.cpp:793-797`, `579-580`).
4. In write mode the lvalue is therefore the failed symbol.

The `invalid pointer` property does not catch it: it is guarded by a
solver-decided `invalid_pointer(deref_expr)` (`dereference.cpp:951`, `968-977`),
which is false exactly when the address is recoverable. The solver layer is not
at fault — `convert_typecast_to_ptr` (`smt_casts.cpp:435-509`) resolves the
integer to `s` correctly by address range, and the `github_426_3` counterexample
shows it doing so; the static value-set layer discards that first.

Not specific to `mul2t`: a plain nondet integer cast to a pointer and constrained
to alias an object loses the write identically, masked there by the alignment
check (`--no-pointer-check` exposes the same false proof).

This refines #6545, which was closed as a stated limitation on the claim that the
gap only over-reports. Reproducer added as
`regression/esbmc/ptr_int_mul_lost_write` (KNOWNBUG);
`docs/design/pointer-integer-provenance.md` is corrected accordingly.

PR #6824, opened 2026-08-08
[pointer-analysis] Add --deref-unknown-objects for non-exhaustive writes

rafaelsamenezes:
A write through a pointer whose value set is not exhaustive was absorbed entirely
by the failed symbol `dereference()` builds as a fallback, so the real object kept
its old value and the store went unreported. Under the new flag the write splits
over every numbered object, with the failed symbol as the final else; the unknown
entry is untouched, so `invalid pointer` still fires. Opt-in because the split
reaches 104 objects and times out `github_1807`.

Refines #6545. Tracked as #6804.


lucasccordeiro (approved):
Verified independently on this branch (arm64 macOS, Bitwuzla):

| check | result |
|---|---|
| #6804 reproducer, default | `SUCCESSFUL` — unchanged |
| #6804 reproducer, `--deref-unknown-objects` | **`FAILED`** — correct |
| additive round-trip control | `FAILED` — still correct |
| `overapproximation-test`, all 6 cases | 17/17 pass |
| full unit suite | **663/663 pass** |

The fourth row is the one that matters. I attempted this in #6809 by raising a `dereference_failure` when a write resolved no target, and it was killed by *a filter that would empty the value set leaves it alone* — checks went **down**, the missed-bug direction. Instrumenting the write-mode value set since then shows why:

| program | value set at the write |
|---|---|
| plain `int *p = &a; *p = 3;` | `targets=1 exhaustive=1 : [object_descriptor]` |
| #6804 repro | `targets=1 exhaustive=0 : [unknown]` |
| that unit test | `targets=1 exhaustive=0 : [unknown]` ×4 |

The provenance case and k-induction's deliberate over-approximation are byte-identical at `dereferencet::dereference`, so no predicate there can separate them. Splitting the write behind an opt-in flag is the right shape precisely because it adds no claim; an always-on check cannot work at this layer.

One scoping point, not a change request: with the flag off by default the false `SUCCESSFUL` persists in the default configuration, so I'd keep #6804 open when this lands and let it track turning the split on (or making it affordable — the 104-object blow-up on `github_1807` is the real obstacle).

Issue #2142, opened 2024-11-13
false positive: incompatible pointer types passing 'bool *' (aka 'int *') to parameter of type '_Bool *' 


oerdnj:
This is a simple reproducer:
```
#include <stdatomic.h>
#include <stdbool.h>

int
main(void) {
        atomic_bool x = false;
        (void)atomic_compare_exchange_strong(&x, &(bool){ false }, true);

        return 0;
}
```

result is:

```
ESBMC version 7.7.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing a.c
a.c:7:43: warning: incompatible pointer types passing 'bool *' (aka 'int *') to parameter of type '_Bool *' [-Wincompatible-pointer-types]
        (void)atomic_compare_exchange_strong(&x, &(bool){ false }, true);
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/llvm-14/lib/clang/14.0.6/include/../include/stdatomic.h:135:112: note: expanded from macro 'atomic_compare_exchange_strong'
#define atomic_compare_exchange_strong(object, expected, desired) __c11_atomic_compare_exchange_strong(object, expected, desired, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)
                                                                                                               ^~~~~~~~
Converting
ERROR: Conversion of unsupported clang cast operator: "NonAtomicToAtomic" to expression
ImplicitCastExpr 0x55dc528f8a50 </tmp/esbmc.b01a-675c-cb76/headers/stdbool.h:6:15> 'atomic_bool':'_Atomic(_Bool)' <NonAtomicToAtomic>
`-ImplicitCastExpr 0x55dc528f8a38 <col:15> '_Bool' <IntegralToBoolean>
  `-IntegerLiteral 0x55dc528f8a18 <col:15> 'int' 0

ERROR: CONVERSION ERROR
```

Unfortunately, even with:
```c
#include <stdatomic.h>
#include <stdbool.h>

int
main(void) {
        atomic_bool x = false;
        bool exp = false;
        (void)atomic_compare_exchange_strong(&x, &exp, true);

        return 0;
}
```

the error remains:

```
ESBMC version 7.7.0 64-bit x86_64 linux
Target: 64-bit little-endian x86_64-unknown-linux with esbmclibc
Parsing a.c
[...]

PR #6827, opened 2026-08-08
[c2goto] Make stdbool.h's bool expand to _Bool, not int

lucasccordeiro:
ESBMC's `stdbool.h` typedef'ed `bool` to `int`, so the conversion to 0 or 1 that C11 6.3.1.2 requires never happened. The everyday mask-test idiom was a false alarm:

```c
bool found = flags & MASK;
assert(found == true);   // VERIFICATION FAILED; holds under clang
```

C11 7.18 specifies `bool` as a macro expanding to `_Bool`; C23 predefines the three names, so the header leaves them alone there. Relates to #2142, whose "'bool *' (aka 'int *')" mismatch is this typedef — the warning no longer fires on LLVM 22, so no test pins it.

Validated against a control binary: the 137 regression tests that include `stdbool.h` and a 150-test sample of the Python list/dict OM suite are verdict-identical, and both new tests flip verdict between control and fixed.


lucasccordeiro:
Worth noting a second defect this fixes, found while differentially testing the C library against clang:

```c
#include <stdbool.h>
int obj; int *p = &obj;
bool b = p;        // ERROR: PARSING ERROR on master
```

C11 6.5.16.1 explicitly admits a pointer on the right of an assignment whose left operand is `_Bool`, so this is valid C — but a constraint violation when `bool` is an `int`, which is why it failed to parse. Covered by two extra assertions in the existing test.

Issue #6831, opened 2026-08-08
SV-COMP: the 831-point drop is entirely timeouts, 60% from #6607


lucasccordeiro:
The score fell **38004 → 37173 (−831)** and `correct` fell **22791 → 22308 (−483)** between these two Benchexec runs:

| | ESBMC | sv-benchmarks | run |
|---|---|---|---|
| good | `978a007e73` | `adc21156d71` | [30680908491](https://github.com/esbmc/esbmc/actions/runs/30680908491/job/91347230245) |
| bad | `5d0876a203` | `f25da614561` | [31217688734](https://github.com/esbmc/esbmc/actions/runs/31217688734/job/93009382544) |

**Every point lost is a timeout. ESBMC did not change a single answer.** Diffing all 36,603 tasks keyed by (run set, task) gives 641 verdict changes, and not one is a `true`↔`false` flip:

| transition | n | effect |
|---|---|---|
| `true` → TIMEOUT | 319 | lost |
| `false(no-overflow)` → TIMEOUT | 148 | lost |
| `false(unreach-call)` → TIMEOUT | 15 | lost |
| `false(valid-memtrack/valid-deref/termination/no-data-race)` → TIMEOUT | 7 | lost |
| TIMEOUT → `true` / `false(...)` | 7 | gained |
| OUT OF MEMORY ↔ TIMEOUT, `unknown` → TIMEOUT | 145 | score-neutral |

## The `incorrect` delta is not ours

`incorrect false` went 6 → 8. Since no ESBMC verdict flipped between `true` and `false`, those two cannot come from ESBMC. The benchmark repository moved between the runs — `adc21156d71` → `f25da614561`, which includes a merge branch named `fix/list-racefree-unreach-verdict` — so the expected verdicts changed underneath us. That also explains `max: 61276 → 61277`.

## Two independent causes

Bucketing the 18,885 tasks that produced an identical verdict in both runs, by their runtime in the good run:

| good runtime | n | median ratio | median delta |
|---|---|---|---|
| 0–1 s | 3551 | 1.214 | 0.19 s |
| 1–2 s | 7777 | 1.142 | 0.18 s |
| 2–5 s | 3317 | 1.118 | 0.36 s |
| 5–10 s | 1906 | 1.076 | 0.51 s |
| 10–20 s | 2771 | 1.054 | 0.75 s |
| 20–40 s | 1764 | 1.043 | 1.14 s |
[...]

PR #6844, opened 2026-08-09
[docs] Plan the schedule-space work behind #6831 cause 1

rafaelsamenezes:
Adds `docs/roadmap/svcomp-6831-schedule-space-plan.md`, a plan for cause 1 of
#6831 (schedule-space explosion after #6607, 291 of 489 lost SV-COMP tasks).

Measured on the issue's own reproducer rather than assumed: MPOR prunes ~20% of
schedules, `--state-hashing` prunes none, and symex outweighs the solver by ~10x
-- so the plan instruments the reduction first, then targets POR and per-schedule
cost. Reverting #6607 is a stated non-goal.

Docs only; no code change.

### DIFF
--- a/docs/roadmap/svcomp-6831-schedule-space-plan.md
+++ b/docs/roadmap/svcomp-6831-schedule-space-plan.md
@@ -1,0 +1,291 @@
+# Plan — affording the schedule space #6607 exposed (issue #6831, cause 1)
+
+**Status:** plan only. No code shipped. Nothing here has been implemented.
+**Owner issue:** [#6831](https://github.com/esbmc/esbmc/issues/6831), *cause 1 —
+schedule-space explosion*, 291 of 489 lost SV-COMP tasks.
+**Bisected to:** `bac652b13c` — `[goto-symex] Track main-thread termination per
+state, not per search` (#6607), which fixes #4584.
+**Last updated:** 2026-08-09.
+
+**Measurement environment.** All numbers below were measured on an x86_64 Linux
+host against `build/src/esbmc/esbmc`, ESBMC 8.4.0, built from `19db2adc96`
+(a descendant of `bac652b13c`, so the post-#6607 behaviour). Solver: Bitwuzla
+0.9.0 (the default). Single run per configuration, one machine, no repetition —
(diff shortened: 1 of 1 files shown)


lucasccordeiro (commented):
Good plan, and the measurements match mine independently — 939 vs my 940, MPOR ~20%, hashing 0. Three things from the issue's data worth folding in.

**1. 287 of the 291 lost concurrency tasks are `true`; only 4 are `false`.** Split from the two Benchexec logs:

| | `true` | `false` |
|---|---|---|
| concurrency-flavoured | **287** | 4 |
| other | 32 | 166 |

This sharpens W1's priority: these tasks need the *exhaustive* space to shrink, so a reduction that merely reaches a violation sooner is worth ~4 tasks. It also forecloses context-bounding as a remedy — `--incremental-context-bound` already exists and a reader may reach for it, but it deepens until `max-context-bound` on a safe program and reports UNKNOWN, which scores the same as the timeout it replaces.

**2. MPOR prunes *post-hoc*, which strengthens the case for W1 candidate 1 (sleep sets).** The code says it outright — *"We can't tell whether or not a transition is allowed in advance because we don't know what it is"* — so the transition is taken and explored, then discarded; only its subtree is saved, never the transition. Instrumenting it: **132 of 1024 transitions blocked (~13%)**, consistent with the 20% schedule figure. So the weakness is structural, not a tuning constant, and sleep sets address the mechanism because they decide *before* exploring. Worth stating in W1 as the reason to try them ahead of DPOR.

**3. §2 re-measured on current master as the document asks.** `8ffb84b24d` includes #6793:

| configuration | plan | master now |
|---|---|---|
| baseline | 939 | 940 |
| `--state-hashing` | 939 | 940 |
| `--no-por` | 1170 | 1171 |
| `--no-por --state-hashing` | 1154 | 1155 |

All within one interleaving — #6793 did not move the counts, so §2 can be acted on as written.

W0 is cheap; I've put it up as #6852, which reports `Schedules explored: 940 (pruned by MPOR: 296, by state hashing: 0)` in one run. Happy to drop it if you'd rather own that piece.

PR #6911, opened 2026-08-10
[goto-symex] Fix --state-hashing soundness and add sleep sets

rafaelsamenezes:
`--state-hashing` collided states differing only in which thread was active, so
violating schedules were never generated; the fingerprint now mixes the active
thread in, and four CORE tests return to their expected FAILED. Adds
`--sleep-sets` (off by default), which skips a thread at a DFS node once its
subtree is exhausted and nothing dependent has run since, plus per-run counters
for what each schedule reduction prunes.

Refs #6831.


Anthonysdu:
I agree hashing active thread is sound, we might need SVCOMP runs as it will cut fewer interleavings and potentially slow down the verification.

Issue #6380, opened 2026-07-24
[contracts] __ESBMC_is_fresh does not compose across a nested --replace-call-with-contract call


adilanwar2399:
## Summary

`__ESBMC_is_fresh` does not compose across a nested `--replace-call-with-contract` / cross-contract call: a function `A` that is itself `is_fresh`-guarded cannot call, via contract replacement, another `is_fresh`-guarded function `B` on a field of `A`'s own parameter — even the simplest possible case (a single non-array struct field, no arrays, no symbolic indexing). This is confirmed as a distinct issue from #6356 (fixed by PR #6358): with #6356's fix applied, the crash this pattern used to produce is gone, but the underlying `is_fresh` composition still fails — now as a clean `contract requires` violation instead of a `SIGABRT`.

## Minimal reproducer

```c
typedef struct { int x; } Inner;
typedef struct { Inner a; Inner b; } Outer;

void touch_inner(Inner *p) {
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(Inner)));
  __ESBMC_ensures(p->x == 0);
  p->x = 0;
}

void touch_outer(Outer *o) {
  __ESBMC_requires(__ESBMC_is_fresh(o, sizeof(Outer)));
  __ESBMC_assigns(o->a, o->b);
  __ESBMC_ensures(1);
  touch_inner(&o->a);
  touch_inner(&o->b);
}

int main(void) { return 0; }
```

Run as:
```
esbmc repro.c --function touch_outer --enforce-contract touch_outer \
  --no-pointer-check --unwind 10 --replace-call-with-contract touch_inner
```

**On stock ESBMC 8.4.0**: crashes.
```
WARNING: Could not find definition for temporary variable: c:@F@touch_inner::$tmp::return_value$___ESBMC_is_fresh$1
Removed function c:@F@touch_inner (replaced with contract)
GOTO program processing time: 0.002s
Starting Bounded Model Checking
libc++abi: terminating due to uncaught exception of type irep2_cast_error: irep2: struct_union_member_names() called on incompatible type (type_id = symbol)
```

**On an ESBMC build with #6298 (fix in PR #6326), #6356 (fix in PR #6358), and #6364 (fix in PR #6365) all integrated** — i.e. the crash this pattern used to hit under #6356 is confirmed gone — it no longer crashes, but still fails, now as a clean requires violation:
```
Violated property:
  file repro.c line 14 column 3 function touch_outer
  contract requires
[...]

PR #6919, opened 2026-08-10
Hold callers to the extent `__ESBMC_is_fresh` asks for

LukeW1999:
`__ESBMC_is_fresh(p, n)` names an extent, and the documentation has always said it means "a valid block of at least `size` bytes". At a replace site only the pointer survived the lowering. The obligation put to the caller was `valid_object(p)` on its own, so any live object of any size discharged it.

```c
typedef struct { int coeffs[4]; } P;

void callee(P *p) {
  __ESBMC_requires(__ESBMC_is_fresh(p, 4096));
  __ESBMC_assigns(p->coeffs);
  __ESBMC_ensures(p->coeffs[0] == 1);
  p->coeffs[0] = 1;
}

int main(void) {
  P *o = (P *)malloc(sizeof(P));   /* 16 bytes */
  __ESBMC_assume(o != 0);
  callee(o);
}
```

`--replace-call-with-contract callee` reported SUCCESSFUL. Sixteen bytes were accepted against a request for 4096, and the callee had already been enforced against a promise the caller never made — which is the composition the two modes exist to guarantee.

## The fix

Carry the extent through `is_fresh_mapping_t` and conjoin it:

```
pointer_offset(p) + n <= dynamic_size(p)
```

`dynamic_size2t` lowers to `__ESBMC_alloc_size[POINTER_OBJECT(p)]`, an allocated byte count, which is the right currency for an `is_fresh` extent — `__ESBMC_get_object_size` counts elements for arrays and goes nondet otherwise, which is why the earlier attempts recorded on #6542 did not get there. The `pointer_offset` term is what keeps `__ESBMC_is_fresh(&a[i], n)` working.

The check is an implication about size, not a blanket rejection: an extent the caller does satisfy still composes. Both directions have a test.

## Testing

340/340 `function_contract`, including `replace_is_fresh_requires_weak_caller_fail`, which #6542 singles out as the one negative control any fix here has to preserve. Two new tests, `github_6542_is_fresh_extent_fail` and `_pass`.

## Scope

This closes the extent half of #6542. The automatic-storage half is a separate path and still reproduces on master; unchanged either way by this.

Programs that passed a smaller object than their contract asked for will now report a violated `requires`. That is the point, but it is a visible change, so the `__ESBMC_is_fresh` section of the docs now says so.

### DIFF
--- a/regression/function_contract/github_6542_is_fresh_extent_pass/main.c
+++ b/regression/function_contract/github_6542_is_fresh_extent_pass/main.c
@@ -1,0 +1,21 @@
+/* github_6542_is_fresh_extent_pass:
+ * __ESBMC_is_fresh(p, n) asks for n bytes, and a replace site has to hold the
+ * caller to it. The extent operand used to be dropped in the lowering, which
+ * The control: an extent the caller does satisfy must still compose, so the
+ * check has to be an implication about size and not a blanket rejection. */
+#include <stdlib.h>
+typedef struct { int coeffs[4]; } P;
+
+void callee(P *p) {
+  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(P)));
+  __ESBMC_assigns(p->coeffs);
+  __ESBMC_ensures(p->coeffs[0] == 1);
+  p->coeffs[0] = 1;
+}
+
+int main(void) {
+  P *o = (P *)malloc(sizeof(P));
+  __ESBMC_assume(o != 0);
+  callee(o);
+  return 0;
+}
--- a/src/goto-programs/contracts/contracts.cpp
+++ b/src/goto-programs/contracts/contracts.cpp
@@ -2258,17 +2258,23 @@ expr2tc code_contractst::replace_is_fresh_temps(
         //
         // requires side at a --replace-call-with-contract call site
         // (require_dynamic == false):
-        //   valid_object(p) only. The precondition is *asserted* against the
-        //   caller's argument here, and a real caller legitimately passes a
-        //   live stack object or an interior sub-object (e.g. &v->vec[k] of a
-        //   fresh vector) -- valid, but not heap-dynamic. Requiring is_dynamic
-        //   would reject every such caller and make is_fresh unusable under
-        //   contract replacement; the frame guarantee is carried by the
-        //   callee's assigns clause, not by heap-freshness. (#6380)
+        //   valid_object(p) and the stated extent, but not is_dynamic. The
+        //   precondition is *asserted* against the caller's argument here, and
+        //   a real caller legitimately passes a live stack object or an
+        //   interior sub-object (e.g. &v->vec[k] of a fresh vector) -- valid,
+        //   but not heap-dynamic. Requiring is_dynamic would reject every such
+        //   caller and make is_fresh unusable under contract replacement; the
+        //   frame guarantee is carried by the callee's assigns clause, not by
+        //   heap-freshness. (#6380)
+        //
+        //   plus the extent the contract asked for, which the caller has to
+        //   supply. dynamic_size is a byte count, the currency is_fresh states
+        //   its extent in, and the offset term is what keeps
+        //   is_fresh(&a[i], n) working. It says nothing about an object with
+        //   automatic storage -- __ESBMC_alloc_size is only maintained for
+        //   heap objects -- so the extent is owed only where it is meaningful,
+        //   which keeps the stack caller above working. (#6542)
         expr2tc valid_obj = valid_object2tc(mapping.ptr_expr);
-        if (!require_dynamic)
-          return valid_obj;
-
         expr2tc ptr_obj = pointer_object2tc(pointer_type2(), mapping.ptr_expr);
 
         const symbolt *dyn_sym = ns.lookup("c:@__ESBMC_is_dynamic");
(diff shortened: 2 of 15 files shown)


src/goto-programs/contracts/contracts.cpp
```diff
@@ -2258,16 +2258,32 @@ expr2tc code_contractst::replace_is_fresh_temps(
...
+        //   supply it: offset + n must fit inside the allocation. dynamic_size
+        //   is a byte count, which is the currency is_fresh states, and the
+        //   offset term is what keeps is_fresh(&a[i], n) working. (#6542)
         expr2tc valid_obj = valid_object2tc(mapping.ptr_expr);
         if (!require_dynamic)
+        {
+          if (!is_nil_expr(mapping.size_expr))
+          {
+            expr2tc n = typecast2tc(size_type2(), mapping.size_expr);
+            expr2tc off = pointer_offset2tc(size_type2(), mapping.ptr_expr);
```
lucasccordeiro: `pointer_offset2t` asserts its type is `signedbv`, but `size_type2()` is unsigned, so this aborts any assertions-enabled build — including both new tests. Use `signed_size_type2()` and cast.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -4379,6 +4395,7 @@ expr2tc code_contractst::lower_is_fresh_in_requires(
     is_fresh_mapping_t m;
     m.temp_var_name = to_symbol2t(call->ret).thename;
     m.ptr_expr = ptr;
+    m.size_expr = call->operands[1];
```
lucasccordeiro: `ptr` is rebound to the actual argument above, but `size_expr` is not, so a formal-parameter extent evaluates the callee's symbol at the call site. Apply the same `replace_symbol_in_expr` loop.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2258,16 +2258,32 @@ expr2tc code_contractst::replace_is_fresh_temps(
...
+        //   is a byte count, which is the currency is_fresh states, and the
+        //   offset term is what keeps is_fresh(&a[i], n) working. (#6542)
         expr2tc valid_obj = valid_object2tc(mapping.ptr_expr);
         if (!require_dynamic)
+        {
+          if (!is_nil_expr(mapping.size_expr))
+          {
+            expr2tc n = typecast2tc(size_type2(), mapping.size_expr);
+            expr2tc off = pointer_offset2tc(size_type2(), mapping.ptr_expr);
+            expr2tc have = dynamic_size2tc(mapping.ptr_expr);
```
lucasccordeiro: `dynamic_size` is meaningless for automatic storage, so this rejects the stack caller the comment above promises to keep working (#6380). Guard the conjunct with `is_dynamic`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2258,16 +2258,32 @@ expr2tc code_contractst::replace_is_fresh_temps(
...
+            valid_obj = and2tc(
+              valid_obj, lessthanequal2tc(add2tc(size_type2(), off, n), have));
```
lucasccordeiro: `off + n` is unsigned and can wrap, making the check vacuously true for a large extent. Compare as `off <= have && n <= have - off`.

regression/function_contract/github_6542_is_fresh_extent_pass/main.c
```diff
@@ -0,0 +1,21 @@
...
+typedef struct { int coeffs[4]; } P;
+
+void callee(P *p) {
+  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(P)));
+  __ESBMC_assigns(p->coeffs);
+  __ESBMC_ensures(p->coeffs[0] == 1);
+  p->coeffs[0] = 1;
+}
+
+int main(void) {
```
lucasccordeiro, line 16: Both new tests are heap, offset 0, constant extent. Add cases for a stack receiver, an interior pointer `&a[i]`, and an extent that is a callee parameter.

lucasccordeiro (changes requested):
Verified against this branch on aarch64 macOS. Three defects reproduce, one of them on the PR's own two tests; details are inline.

LukeW1999:
All five taken, and thank you — two of them were defects I had no way of seeing, for a reason worth stating.

**My "340/340" was measured on a build with assertions compiled out.** Local `RelWithDebInfo` carries `-DNDEBUG`, CI uses `DebugOpt`. So `pointer_offset2t`'s `signedbv` assertion could never have fired for me, and you are right that both of my own tests would have aborted in CI. I have switched my build to `DebugOpt` and re-measured everything below on it. That was a hole in how I verify, not a detail I missed.

**Signed offset.** `pointer_offset2tc(get_int_type(config.ansi_c.address_width), ...)` and a cast, following the idiom in `dereference.cpp`.

**Unbound extent.** Fixed, and it was the most consequential of the five: the motivating case is `bufs_differ(b1, b2, n)` wanting `__ESBMC_is_fresh(b1, n)`, which is precisely a parameter extent. Both my tests used constants, so they never touched it.

The discriminating test is the pass direction, not the fail direction. With the rebinding reverted, an unbound callee `n` is free, so a solver picks it large and the obligation fails either way; `github_6542_is_fresh_param_extent_fail` is FAILED with or without the fix and does not separate them. `..._param_extent_pass` is FAILED without the rebinding and SUCCESSFUL with it. Both are checked in; the pass case is the one that guards this.

**Automatic storage.** Guarded as `!is_dynamic || fits`. The emitted predicate now reads

```
VALID_OBJECT(&v) && (!is_dynamic[POINTER_OBJECT(&v)] || ...)
```

so the extent is not owed where `__ESBMC_alloc_size` is not maintained.

The stack receiver still fails, but on `valid_object` alone, and that is #6542's automatic-storage half rather than anything this PR adds: `__ESBMC_alloc` is never updated for stack pointers, so `VALID_OBJECT` of an automatic object is a free boolean. Same failure on master. I have checked the case in as `github_6542_is_fresh_stack_receiver_knownbug` with that written down, rather than assert a SUCCESSFUL that is not true yet.

**Unsigned wrap.** Your form, `off <= have && n <= have - off`. I considered the widening idiom in `add_restrict_assertions.cpp:83-95`, which carries a comment about this exact hazard, but it introduces address-width-doubled arithmetic and yours is as correct without it.

**Tests.** Now five: the two originals, `param_extent_{fail,pass}`, `interior_ptr_fail` (`&p[3]` against 8 bytes of a 10-byte object, which pins the offset term), and the stack KNOWNBUG.

344/344 `function_contract` on `DebugOpt`, `replace_is_fresh_requires_weak_caller_fail` still FAILED, clang-format clean, complexity unchanged at 570.

PR #6926, opened 2026-08-10
[unit] Remove the temporary directory goto_factory stages sources in

Yiannis128:
goto_factory staged every test source with get_unique_tmp_path(), which creates the directory but leaves removal to the caller, and neither caller removed it, so a single `ctest -LE regression` run left 322 esbmc-test-* directories in /tmp. create_tmp_dir() returns the same directory as a tmp_path, so holding it across the parse gets it removed by the destructor, and nothing in the returned program refers back to the source file.

The added test points TMPDIR at a private directory, builds a program through goto_factory, and checks that no esbmc-test-* entry survives the call. It fails on master and passes with the fix. Unit harness only, no change to ESBMC behaviour, so there is no matching pair of passing and failing regression tests.


unit/testing-utils/goto_factory.cpp
```diff
@@ -85,9 +85,11 @@ program goto_factory::get_goto_functions(
   goto_factory::Architecture arch,
   const std::string &test_name)
 {
-  std::string filename(
-    file_operations::get_unique_tmp_path("esbmc-test-%%%%%%"));
-  filename += "/" + test_name;
+  /* The directory outlives the parse and is removed by ~tmp_path(): the
+   * program built from it holds no reference to the source. */
+  file_operations::tmp_path dir =
+    file_operations::create_tmp_dir("esbmc-test-%%%%%%");
```
lucasccordeiro: create_tmp_dir() registers every path for signal cleanup and nothing unregisters it; 300+ calls saturate the 32-slot sig_tmp_paths array, so genuinely live temp paths are then lost on SIGTERM. Wrap get_unique_tmp_path in a tmp_path instead.

unit/testing-utils/goto_factory.cpp
```diff
@@ -99,9 +101,9 @@ program goto_factory::get_goto_functions(
   goto_factory::Architecture arch,
   const std::string &test_name)
 {
-  std::string filename(
-    file_operations::get_unique_tmp_path("esbmc-test-%%%%%%"));
-  filename += "/" + test_name;
+  file_operations::tmp_path dir =
+    file_operations::create_tmp_dir("esbmc-test-%%%%%%");
```
lucasccordeiro: Same registration leak as the istream overload above; both call sites need the non-registering variant.

unit/testing-utils/goto_factory.test.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  // goto_factory resolves the temp directory per call, so TMPDIR confines the
+  // staged directory to `sandbox` and keeps this off the shared /tmp, where a
+  // concurrently running unit test would stage one of its own.
+  const char *saved = getenv("TMPDIR");
+  const std::string old_tmpdir = saved ? saved : "";
+  setenv("TMPDIR", sandbox.path().c_str(), 1);
+
+  std::string code = "int main() { return 0; }";
+  program p = goto_factory::get_goto_functions(code);
```
lucasccordeiro: TMPDIR is left pointing at the removed sandbox if get_goto_functions throws. Restore it from a scope-guard destructor rather than inline after the call.

lucasccordeiro: Only the std::string overload is covered; the std::istream overload was changed identically and is untested. Add a case for it.

lucasccordeiro (changes requested):

unit/testing-utils/goto_factory.cpp
Yiannis128: Done in c71777e — both call sites now wrap get_unique_tmp_path() in a tmp_path, so nothing is added to the cleanup registry.

unit/testing-utils/goto_factory.cpp
Yiannis128: Done in c71777e — both call sites now wrap get_unique_tmp_path() in a tmp_path, so nothing is added to the cleanup registry.

unit/testing-utils/goto_factory.test.cpp
Yiannis128: Done — TMPDIR is set and restored by a tmpdir_override scope guard, so a throwing stage call cannot leave it pointing at the removed sandbox.

Yiannis128: Done — added a second case for the istream overload; both go through the same helper that stages under the sandbox and lists what survives.

unit/testing-utils/goto_factory.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <vector>
+
+#ifndef _WIN32
+namespace
+{
+/* Restores TMPDIR from the destructor: a staging call that throws would
+ * otherwise leave the rest of the binary pointing at the removed sandbox. */
+class tmpdir_override
+{
+  const char *_saved = getenv("TMPDIR");
```
lucasccordeiro, line 20: `setenv` in the constructor may invalidate this pointer, so the destructor tests an indeterminate value. Store a `bool` instead of the `const char *`.

unit/testing-utils/goto_factory.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "goto_factory.h"
+#include <util/base/filesystem.h>
+
+#include <boost/filesystem.hpp>
+#include <cstdlib>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#ifndef _WIN32
```
lucasccordeiro, line 13: The whole file is guarded, so the leak fix ships untested on Windows, where ctest also runs. Use `_putenv_s` there instead of excluding it.

unit/testing-utils/goto_factory.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+ * where a concurrently running unit test would stage one of its own. */
+template <typename F>
+std::vector<std::string> stage_and_list_leftovers(F &&stage)
+{
+  namespace fs = boost::filesystem;
+  auto sandbox = file_operations::create_tmp_dir("esbmc-test-sandbox-%%%%");
+  {
+    tmpdir_override tmpdir(sandbox.path());
+    program p = stage();
+    REQUIRE(
```
rafaelsamenezes, line 76: A staging path that stopped honouring TMPDIR would leave the sandbox empty and this test green. Assert something was staged here, before checking it is gone.

unit/testing-utils/goto_factory.test.cpp
Yiannis128: Done in 99b044f — the guard stores `bool _was_set` and the `std::string` copy, so nothing reads a pointer that `setenv` may have invalidated.

unit/testing-utils/goto_factory.test.cpp
Yiannis128: Done in 99b044f — the `#ifndef _WIN32` is gone and the override goes through a `set_tmpdir()` helper that uses `_putenv_s` there. One correction to the mechanism though: on Windows it has to override `TMP`, not `TMPDIR`. `boost::filesystem::temp_directory_path()` consults `TMPDIR, TMP, TEMP, TEMPDIR` on POSIX but `TMP, TEMP, LOCALAPPDATA, USERPROFILE` on Windows (boostorg/filesystem src/operations.cpp), so setting `TMPDIR` there would leave the staged directory in the real temp dir, the sandbox listing would come back empty either way, and the test would pass on Windows whether or not the leak is fixed.

unit/testing-utils/goto_factory.test.cpp
Yiannis128: Agreed, and it applies to the sandbox not being reached at all, not just to staging drifting away from it. Done in 7ee0d66 — inside the override the helper now claims a directory through the same `get_unique_tmp_path("esbmc-test-%%%%%%")` goto_factory uses and requires its parent to be the sandbox, so the override has to be live for the leftover check to mean anything. The probe is held in a `tmp_path`, so it is gone before the listing.

unit/testing-utils/goto_factory.test.cpp
```diff
@@ -0,0 +1,114 @@
...
+      fs::equivalent(fs::path(probe.path()).parent_path(), sandbox.path()));
+
+    program p = stage();
+    REQUIRE(
+      p.functions.function_map.find("c:@F@main") !=
+      p.functions.function_map.end());
+  }
+
+  // The frontends extract their headers into the same directory and remove
```
lucasccordeiro, line 81: C/C++ frontends serve bundled headers from `esbmc_clang_vfs()`, never TMPDIR, so this rationale is stale. Drop it and assert the sandbox is empty instead.

Issue #6938, opened 2026-08-11
[python][contracts] umbrella: bring function contracts to the Python frontend, scalars first


LukeW1999:
> **Status, re-measured on master `3f00d56ed1` (2026-09-06).** Section 1 below
> no longer holds: it records the state on `42aa5ce8be` (2026-08-10), when no
> Python function had a contract clause at all. Most of the v1 scope has since
> landed. `regression/python-contracts/` now holds 68 tests: 65 CORE, all
> passing, and 3 FUTURE marking what is still missing. Both
> `--enforce-contract` and `--replace-call-with-contract` work on Python.
>
> Still missing, each pinned by a `FUTURE` test:
> `len()` and subscripting inside a clause (#7629), and `__ESBMC_assigns`
> (#7630). Closed in favour of those two. Read section 1 as history, not as
> current behaviour.

Function contracts (`__ESBMC_requires` / `__ESBMC_ensures` / `__ESBMC_assigns`, `--enforce-contract`, `--replace-call-with-contract`) are a C/C++-only feature today. Nothing in the Python frontend produces a contract clause, so both contract modes reject every Python function.

This issue does three things: confirms the current state with commands and output, fixes a **minimal v1 scope** using the Python parameter representation as the deciding evidence, and splits the work into tasks by replaying how the C side actually landed.

Measured on `42aa5ce8be` (2026-08-10), ESBMC 8.4.0, local `RelWithDebInfo` build.

---

## 1. Confirmed state

Three distinct failure modes, none of them a working contract.

**(a) Undeclared clause becomes a reachability assertion.**

```python
def double(x: int) -> int:
    __ESBMC_requires(x > 0)
    __ESBMC_ensures(x > 0)
    return 2 * x

y: int = double(5)
assert y == 10
```

```
$ esbmc c3.py
WARNING: Undefined function '__ESBMC_requires' - replacing with assert(false)
WARNING: Undefined function '__ESBMC_ensures' - replacing with assert(false)
...
  Unsupported function '__ESBMC_requires' is reached
VERIFICATION FAILED
```

Loud rather than silent, which is the right failure, but it means the C idiom does not merely no-op in Python the way it does in C under plain BMC.

**(b) Declared as a Python stub, silently a normal call.**

Adding `def __ESBMC_requires(c: bool) -> None: pass` makes the file parse, and then both modes reject the function:

```
$ esbmc c2.py --enforce-contract double
[...]

Issue #6485, opened 2026-07-28
[contracts] umbrella: --enforce-contract must not assume a pointer extent the contract does not state


Yiannis128:
Tracks the work to make `--enforce-contract` stop assuming anything about a pointer parameter that the contract does not state. Four items, one of which is already fixed and one of which is the remaining blocker.

## Dependency tree

```
PR #6484  [MERGED 30 Jul]  give pointer parameters a nondet extent
│
├─ closes ──────────> #6215  [CLOSED]  ARRAY_ALLOC_ELEMS used as a byte count
│                            fully resolved: allocating in bytes removes
│                            the byte/element split entirely
│
└─ partly addresses ─> #6212  [OPEN]  unstated buffer extent
                       │      primitive + array params: DONE in this PR
                       │      struct params: still assume 1 element
                       │
                       └─ was blocked by > #6483  [CLOSED 7 Aug, verified]  is_fresh struct param +
                                         two __ESBMC_old field snapshots
                                         silently discharges the ensures
                                         (pre-existing, repros on 8.3.0)
```

## Which edges are real

| Edge | Kind | Where it lives |
| --- | --- | --- |
| #6483 blocks #6212 | formal, now discharged | #6483 closed 7 Aug; reproducer verified gone on master |
| #6484 closes #6215 | formal | `Fixes #6215`, appears in `closingIssuesReferences` |
| #6484 to #6212 | prose only | `Addresses #6212`, deliberately not a closing reference |
| #6215 "depends on #6212" | prose only, now stale | written when the two were expected to land separately |

## Items

- [ ] #6212 `--enforce-contract` assumes an unstated buffer extent and reports VERIFICATION SUCCESSFUL on an out-of-bounds write
- [x] #6215 `--enforce-contract` allocates `ARRAY_ALLOC_ELEMS` as a byte count, not an element count
- [x] #6483 `--enforce-contract` misses an `ensures` violation on an `is_fresh` struct param with two `__ESBMC_old` field snapshots
- [x] #6484 give pointer parameters a nondet extent in enforce harnesses

## State

[...]

Issue #6212, opened 2026-07-20
--enforce-contract assumes an unstated buffer extent and reports VERIFICATION SUCCESSFUL on an out-of-bounds write


Yiannis128:
```c
void f(int *p) {
    __ESBMC_requires(p != 0);
    __ESBMC_ensures(1);
    p[20] = 1;
}
```

```
esbmc --function f --enforce-contract f --overflow-check --unsigned-overflow-check \
      --memory-leak-check --ub-shift-check --force-malloc-success f.c
```

`VERIFICATION SUCCESSFUL` on 8.4.0 (built from b19ab45d75) and on 8.3.0. The body writes past the end of any caller buffer smaller than 21 ints, and the contract does not constrain the buffer's extent, so this is a genuine defect that the enforce check reports as clean. CBMC's equivalent enforce harness reports the same function as a bounds violation.

The cause is `add_pointer_validity_assumptions` backing every primitive pointer parameter with a fixed-size heap object (`goto-programs/contracts/contracts.cpp:4092`, via `emit_pointer_param_malloc`). The precondition `p != 0` does not imply that `p` points at that many bytes, so the harness adds an assumption the contract does not state. Enforce mode then verifies a strict subset of the inputs the contract admits, and the reported verdict is unqualified rather than conditional on the assumed extent. `--check-vacuity` does not catch this, because the path is reachable, just under-constrained.

I understand why the fixed size is there. The comment at `contracts.cpp:4095` notes that a single-element stack allocation makes out-of-bounds reads produce a fresh nondet on each access, so an `ensures` of the form `__ESBMC_return_value == arr[idx]` becomes unprovable. Reverting to a single element would trade this false pass for false failures, which is not obviously an improvement.

What would fix it without that trade is a symbolic extent: allocate a nondet size for the parameter and let the existing bounds checks decide, which is what `__ESBMC_is_fresh(p, n)` already does for an explicit `n`. Accesses the contract does not justify would then be caught, while the object stays a single consistent SMT allocation, which is the property the current fixed size is protecting. Failing that, the assumed extent should at least be reported, so that a pass is qualified rather than unconditional.

For contrast, `__ESBMC_is_fresh` already behaves correctly in both directions: `is_fresh(p, 1000)` admits `p[200]`, and `is_fresh(p, 8)` rejects `p[24]`. The problem is only the default when the contract says nothing about the extent.

[...]

PR #6940, opened 2026-08-11
[python] Lower requires/ensures clauses in the Python frontend

LukeW1999:
Function contracts were C/C++ only. A Python `__ESBMC_requires` call either hit the undefined-function stub, or, once declared as a Python `def`, reached `builtin_functions.cpp` as a callee carrying a body, which short-circuits to a plain `FUNCTION_CALL` before the clause lowering at `builtin_functions.cpp:906` can run. Both contract modes then rejected every Python function:

```
$ esbmc main.py --enforce-contract double
ERROR: --enforce-contract: cannot use 'double': that function declares no contract clauses
```

The error was never "unknown function `double`". `code_contractst::find_function_symbol` already resolved the Python symbol by short name; only clause detection failed.

Two commits: make clauses work, then stop the cases that do not work from passing silently.

## 1. Lower the clauses

Two frontend edits, ~30 lines. `src/goto-programs/contracts/contracts.cpp` is unchanged.

- `function_call/builder.cpp`: register `__ESBMC_requires` and `__ESBMC_ensures` as bodiless `void f(bool)` symbols. Folded into the existing `__ESBMC_loop_invariant` registration, since all three want the same shape: a valueless symbol whose call `goto_convert` lowers into a `LOOP_INVARIANT` or a marked `ASSUME`.
- `converter/converter_expr.cpp`: resolve `__ESBMC_return_value` to a symbol typed from the enclosing function's return annotation. The contracts pass then rewrites it to the real return value, exactly as it does for C.

Inertness under plain BMC came for free: the clause drop in `goto_sideeffects.cpp` keys on `fsym->name`. `--replace-call-with-contract` likewise needed nothing beyond the registration.

```python
def double(x: int) -> int:
    __ESBMC_requires(x > 0)
    __ESBMC_ensures(__ESBMC_return_value > x)
    return 2 * x
```

```
Violated property:
  contract ensures
  return_value > x
```

## 2. Reject clauses that do not lower to an expression

A clause becomes one ASSUME/ASSERT, so its argument has to survive as a pure expression. Three Python constructs do not, and each one silently verified a different property than the one written:

| construct | what was actually verified |
| --- | --- |
| a call, e.g. `__ESBMC_requires(len(l) > 0)` | `ASSUME <SSA temp> > 0`, where the temp is declared in the *original body* and is free in the wrapper. The precondition became arbitrary. |
| a subscript, e.g. `__ESBMC_ensures(rv == l[0])` | `ASSERT 0`. A guaranteed false positive. |
| a parameter whose type could not be inferred | `o > 0` on a `void *` compares a pointer, not the value the source reads as. |

All three now fail with a diagnostic naming the construct and the line:

```
ERROR: __ESBMC_requires clause at line 2 contains a function call; a contract clause must be a pure expression
ERROR: __ESBMC_ensures clause at line 3 contains a subscript; a contract clause must be a pure expression
ERROR: __ESBMC_requires clause at line 2 references 'o', whose type could not be determined; annotate the parameter so the clause constrains its value
```

The discriminator is the **clause expression, not the parameter type**. A `list` or `str` parameter is fine as long as the clause itself is pure, which the `typed_container_param` test pins down:

```python
def take(l: list, n: int) -> int:      # list parameter, allowed
    __ESBMC_requires(n > 0)
    __ESBMC_ensures(__ESBMC_return_value == n)
```

An unannotated parameter is likewise allowed when the frontend inferred its type from a call site; it is rejected only when the type really is `void *`.

`__ESBMC_old` is not implemented yet and is a call, so it now reports the pure-expression diagnostic rather than "Unsupported function".

## 3. Stop mypy reporting ESBMC intrinsics as undefined

Pre-existing and not specific to contracts: the frontend supplies `__ESBMC_assume`, `nondet_int` and friends, so `mypy --strict` flagged every use. A three-line file printed five `name-defined` errors before this. Those diagnostics are now filtered out. A name the user actually mistyped still reaches the report, and non-name diagnostics are untouched.

## Tests

New `regression/python-contracts` suite with its own ctest label, 15 tests, all passing. Every positive is paired with a mutation that must fail, so none of the greens can be vacuous.

| test | expected |
| --- | --- |
| `enforce_int` / `_fail` | SUCCESSFUL / FAILED on `contract ensures` |
| `enforce_float` / `_fail` | SUCCESSFUL / FAILED on `contract ensures` |
| `enforce_bool` / `_fail` | SUCCESSFUL / FAILED on `contract ensures` |
| `replace_int` | SUCCESSFUL |
| `replace_int_requires_fail` | FAILED on `contract requires` at the call site |
| `replace_int_vacuity_fail` | FAILED |
| `clauses_inert` | SUCCESSFUL with no contract flag |
| `clause_call_fail` | the call diagnostic |
| `clause_subscript_fail` | the subscript diagnostic |
| `clause_untyped_fail` | the untyped-parameter diagnostic |
| `typed_container_param` / `_fail` | SUCCESSFUL / FAILED, list parameter with a pure clause |

`replace_int_vacuity_fail` is the soundness control: the caller asserts `y == 10`, which `ensures(return_value > x)` does not entail, so the replacement must not prove it. This is the #6212 / #6542 failure class, caught cheaply rather than after the fact.

Three unit tests in `unit/python-frontend/test_typecheck_intrinsics.py` cover the mypy filter, including that a genuine typo survives it.

Also run: the five existing `regression/python` tests that assert on undefined-name output, the loop-invariant tests, and two ~85-test samples spread across the `python` suite. All passing. `git clang-format` against `origin/master` is clean under clang-format-11, and `yapf` is clean on the Python changes.

Complexity gate: `check_contract_clause` is new at CCN 11, its strict limit. `get_expr` 283 → 287 and `function_call_builder::build` 64 → 66, both already over threshold before this change and both still under their strict limits.

## Not in this PR

- `__ESBMC_old`, which needs a frontend-side snapshot since Python has no address-of
- `__ESBMC_assigns` and any frame reasoning
- a separation predicate for aliased containers
- methods and `self`, `--enforce-all-contracts`
- exceptional postconditions

These are P1.3 onward in #6938, and the container work there will hit the same extent and aliasing questions the #6485 umbrella tracks for C.

Refs #6938

### DIFF
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -864,6 +864,14 @@ static void emit_va_marker_call(
   t->location = function.location();
 }
 
+static bool contracts_enabled(const optionst &options)
+{
+  return !options.get_option("enforce-contract").empty() ||
+         !options.get_option("replace-call-with-contract").empty() ||
+         options.get_bool_option("enforce-all-contracts") ||
+         options.get_bool_option("replace-all-contracts");
+}
+
 void goto_convertt::do_function_call_symbol(
   const exprt &lhs,
   const exprt &function,
@@ -935,6 +943,14 @@ void goto_convertt::do_function_call_symbol(
       "Found __ESBMC_assigns call with {} arguments",
       arguments.size());
   }
+
+  // A contract clause states nothing outside contract mode. goto_sideeffects
+  // already drops it, but only for clauses that arrive as a side-effect
+  // expression; the Python frontend emits a direct FUNCTION_CALL, which never
+  // reaches that strip, so a live `requires` would be assumed and mask real
+  // bugs in the function it annotates.
+  if ((is_requires || is_ensures) && !contracts_enabled(options))
+    return;
 
   if (is_assume || is_assert || is_loop_invariant || is_requires || is_ensures)
   {
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -864,12 +864,9 @@ static void emit_va_marker_call(
   t->location = function.location();
 }
 
-static bool contracts_enabled(const optionst &options)
+bool goto_convertt::drop_inactive_contract_clause(bool is_clause) const
 {
-  return !options.get_option("enforce-contract").empty() ||
-         !options.get_option("replace-call-with-contract").empty() ||
-         options.get_bool_option("enforce-all-contracts") ||
-         options.get_bool_option("replace-all-contracts");
+  return is_clause && !options.contracts_enabled();
 }
 
 void goto_convertt::do_function_call_symbol(
@@ -933,6 +930,7 @@ void goto_convertt::do_function_call_symbol(
   bool is_loop_invariant = (base_name == "__ESBMC_loop_invariant");
   bool is_requires = (base_name == "__ESBMC_requires");
   bool is_ensures = (base_name == "__ESBMC_ensures");
+  bool is_clause = is_requires || is_ensures;
   bool is_assigns = (base_name == "__ESBMC_assigns");
 
   // Debug: log if we see assigns
@@ -949,10 +947,10 @@ void goto_convertt::do_function_call_symbol(
   // expression; the Python frontend emits a direct FUNCTION_CALL, which never
   // reaches that strip, so a live `requires` would be assumed and mask real
   // bugs in the function it annotates.
-  if ((is_requires || is_ensures) && !contracts_enabled(options))
+  if (drop_inactive_contract_clause(is_clause))
     return;
 
-  if (is_assume || is_assert || is_loop_invariant || is_requires || is_ensures)
+  if (is_assume || is_assert || is_loop_invariant || is_clause)
   {
     if (arguments.size() != 1)
     {
@@ -962,7 +960,7 @@ void goto_convertt::do_function_call_symbol(
 
     if (
       options.get_bool_option("no-assertions") && !is_assume &&
-      !is_loop_invariant && !is_requires && !is_ensures)
+      !is_loop_invariant && !is_clause)
       return;
 
     // Rafael's invariant merging: combine consecutive
@@ -995,7 +993,7 @@ void goto_convertt::do_function_call_symbol(
     else
     {
       // For contract functions, generate ASSUME instructions with special markers
-      if (is_requires || is_ensures)
+      if (is_clause)
       {
         t = dest.add_instruction(ASSUME);
         t->guard = guard;
--- a/src/python-frontend/converter/converter_expr.cpp
+++ b/src/python-frontend/converter/converter_expr.cpp
@@ -501,6 +501,23 @@ symbolt *python_converter::contract_return_value_symbol(
   return add_symbol_and_get_ptr(ret_symbol);
 }
 
+/// The "variable is not defined" diagnostic for a Name that resolved to no
+/// symbol, naming the enclosing function when the reference is inside one.
+static std::string undefined_variable_message(
+  const std::string &var_name,
+  const std::string &func_name,
+  const locationt &location)
+{
+  std::ostringstream error_msg;
+  error_msg << "Variable '" << var_name << "' is not defined";
+  if (!func_name.empty())
+    error_msg << " in function '" << func_name << "'";
+  if (!location.get_line().empty())
+    error_msg << " at line " << location.get_line();
+  error_msg << ".";
+  return error_msg.str();
+}
+
 exprt python_converter::get_expr(const nlohmann::json &element)
 {
   get_expr_depth_guard depth_guard(*this);
@@ -1047,27 +1064,8 @@ exprt python_converter::get_expr(const nlohmann::json &element)
           break;
         }
 
-        locationt location = get_location_from_decl(element);
-        std::ostringstream error_msg;
-        if (!current_func_name_.empty())
-        {
-          // Variable referenced inside a function
-          error_msg << "Variable '" << var_name
-                    << "' is not defined in function '" << current_func_name_
-                    << "'";
-          if (!location.get_line().empty())
-            error_msg << " at line " << location.get_line();
-          error_msg << ".";
-        }
-        else
-        {
-          // Variable referenced at global scope
-          error_msg << "Variable '" << var_name << "' is not defined";
-          if (!location.get_line().empty())
-            error_msg << " at line " << location.get_line();
-          error_msg << ".";
-        }
-        throw std::runtime_error(error_msg.str());
+        throw std::runtime_error(undefined_variable_message(
+          var_name, current_func_name_, get_location_from_decl(element)));
       }
     }
 
--- a/src/python-frontend/converter/converter_expr.cpp
+++ b/src/python-frontend/converter/converter_expr.cpp
@@ -1007,6 +1007,30 @@ exprt python_converter::get_expr(const nlohmann::json &element)
             break;
           }
         }
+        // __ESBMC_return_value names the returned value inside an
+        // __ESBMC_ensures clause. The contracts pass rewrites it to the real
+        // return value, so the frontend only has to give it the enclosing
+        // function's return type for the clause to type-check.
+        if (var_name == "__ESBMC_return_value" && !current_func_name_.empty())
+        {
+          symbol_id ret_sid = create_symbol_id();
+          symbolt *func_symbol = find_symbol(ret_sid.to_string());
+          if (func_symbol && func_symbol->get_type().is_code())
+          {
+            ret_sid.set_object(var_name);
+            symbolt ret_symbol = create_symbol(
+              current_python_file,
+              var_name,
+              ret_sid.to_string(),
+              get_location_from_decl(element),
+              to_code_type(func_symbol->get_type()).return_type());
+            ret_symbol.lvalue = true;
+            ret_symbol.file_local = true;
+            expr = symbol_expr(*add_symbol_and_get_ptr(ret_symbol));
+            break;
+          }
+        }
+
         locationt location = get_location_from_decl(element);
         std::ostringstream error_msg;
         if (!current_func_name_.empty())
--- a/src/python-frontend/converter/converter_expr.cpp
+++ b/src/python-frontend/converter/converter_expr.cpp
@@ -1015,7 +1015,15 @@ exprt python_converter::get_expr(const nlohmann::json &element)
         {
           symbol_id ret_sid = create_symbol_id();
           symbolt *func_symbol = find_symbol(ret_sid.to_string());
-          if (func_symbol && func_symbol->get_type().is_code())
+          // A None-returning function has no value to name, and an
+          // empty-typed symbol crashes the encoder rather than failing here.
+          const typet ret_type =
+            func_symbol && func_symbol->get_type().is_code()
+              ? to_code_type(func_symbol->get_type()).return_type()
+              : typet();
+          if (
+            ret_type.is_not_nil() && ret_type != none_type() &&
+            ret_type.id() != "empty")
           {
             ret_sid.set_object(var_name);
             symbolt ret_symbol = create_symbol(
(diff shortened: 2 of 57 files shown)


LukeW1999:
Code review of the branch found a soundness hole I introduced, plus two crashes. All fixed in `8d890dd6d7`; the suite is now 23 regression tests and 4 unit tests.

### The one that mattered

Contract clauses stayed live as `ASSUME` when no contract flag was passed, so a `requires` masked real bugs in the function it annotates:

```python
def f(x: int) -> int:
    __ESBMC_requires(x > 0)
    return 100 // x          # ZeroDivisionError

def main() -> None:
    n: int = nondet_int()
    y: int = f(n)
```

`VERIFICATION SUCCESSFUL` before, `VERIFICATION FAILED` after. `__ESBMC_ensures(rv != rv)` was worse: it became `assume(false)` and let `assert False` pass.

`goto_sideeffects.cpp:1495` nils the clause only along the `remove_sideeffects` path, which sees clauses arriving as a side-effect expression. Clang produces one; the Python frontend emits a direct `FUNCTION_CALL`, which never reaches that strip. The drop now happens in `goto_convert` instead, where both frontends pass through. The C `function_contract` suite (346 tests) is unaffected.

My original `clauses_inert` test could not have caught this: `double(5)` satisfies `x > 0` trivially, so it passed whether the clause was stripped or live. Replaced by `clauses_inert_masking_fail`, which is the repro above and must report FAILED. That was a real gap in how I wrote the control, not just a missing case.

### Also fixed

| defect | before | after |
| --- | --- | --- |
| `__ESBMC_return_value` in a `-> None` function | `ERROR: Can't generate zero for type empty` + core dump | rejected, naming the function |
| two-argument clause, `__ESBMC_requires(a, b)` | `abort()` + core dump | rejected with a line number and "combine conditions with 'and'" |
| `__ESBMC_return_value` inside `requires` | silently assumed over a free symbol | rejected: a precondition cannot mention the return value |
| `__ESBMC_old` rejection message | "must be a pure expression" | names the intrinsic as not yet supported, rather than blaming purity |
| `replace_int_vacuity_fail` | matched any failure | pins `^  assertion y == 10$` |
| mypy failure with no `: error:` line | swallowed | passed through unchanged |

### Verification

- `python-contracts` 23/23
- C `function_contract` 346/346, confirming the `builtin_functions.cpp` change is inert for C
- `loop-invariants` 76/76
- three ~80-test samples across the `python` suite
- 4 unit tests on the mypy filter
- clang-format-11 clean on changed lines; complexity `do_function_call_symbol` 161 → 164 and `build` 64 → 66, both reduced from the first cut by extracting `contracts_enabled` and `check_contract_call`

Three review findings I did not act on, with reasons:

- **blanket `Call` rejection bans `__ESBMC_old` / `__ESBMC_is_fresh`.** Correct, and deliberate: neither is implemented for Python yet, so rejecting is right. Only the diagnostic was wrong, which is fixed. Whitelisting them belongs with implementing them (#6938 P1.3).
- **the untyped-parameter guard passes when the symbol is not in the table.** True. It is defence in depth on top of the type check, not a claimed invariant; a name not yet registered cannot be a parameter, which is the case the guard targets.
- **the suppression regex is broader than the intrinsic set.** `nondet_\w+` matches the frontend's own `starts_with("nondet_")` dispatch, so it is not broader there. A mistyped `__ESBMC_asume` still fails loudly at verification via the undefined-function stub, so the mypy suppression hides nothing that stays hidden.

src/python-frontend/function_call/builder.cpp
```diff
@@ -88,6 +92,129 @@ const std::string kPytTerminate = "__pyt_terminate";
...
+  {
+    if (node.is_array())
+      for (const auto &child : node)
+        check_contract_clause(child, clause);
+    return;
+  }
+
+  const std::string node_type = node.value("_type", "");
+
+  if (node_type == "Call" || node_type == "Subscript")
```
lucasccordeiro: Blocklisting only `Call`/`Subscript` lets `ListComp`, `GeneratorExp` and `NamedExpr` through, which also do not lower to a pure expression. Invert this to an allow-list of pure node types.

src/python-frontend/function_call/builder.cpp
```diff
@@ -88,6 +92,129 @@ const std::string kPytTerminate = "__pyt_terminate";
...
+    }
+
+    symbol_id sid(
+      converter_.python_file(),
+      converter_.current_classname(),
+      converter_.current_function_name());
+    sid.set_object(name);
+
+    const symbolt *sym = converter_.symbol_table().find_symbol(sid.to_string());
+    if (sym && sym->get_type() == any_type())
```
lucasccordeiro: The lookup is scoped to the current function, so a clause naming a module-level `any_type()` global escapes this check. Fall back to a module-scope lookup.

src/goto-programs/builtin_functions.cpp
```diff
@@ -845,6 +845,14 @@ static void emit_va_marker_call(
   t->location = function.location();
 }
 
+static bool contracts_enabled(const optionst &options)
```
lucasccordeiro: Duplicates the four-option contract-mode predicate at `goto_sideeffects.cpp:1496`; the two copies will drift. Extract one shared helper and call it from both sites.

src/python-frontend/converter/converter_expr.cpp
```diff
@@ -1013,6 +1013,38 @@ exprt python_converter::get_expr(const nlohmann::json &element)
...
+          symbol_id ret_sid = create_symbol_id();
+          symbolt *func_symbol = find_symbol(ret_sid.to_string());
+          // A None-returning function has no value to name, and an
+          // empty-typed symbol crashes the encoder rather than failing here.
+          const typet ret_type =
+            func_symbol && func_symbol->get_type().is_code()
+              ? to_code_type(func_symbol->get_type()).return_type()
+              : typet();
+          if (
+            ret_type.is_not_nil() && ret_type != none_type() &&
```
lucasccordeiro: Re-implements `returns_no_value()` from `builder.cpp:97`. Share that helper here so the two None-return rules cannot diverge.

src/python-frontend/parser/typecheck.py
```diff
@@ -11,24 +12,40 @@
 
 __all__ = ["run_mypy_strict"]
 
+# Verification intrinsics (__ESBMC_assume, nondet_int, __ESBMC_requires, ...)
+# are supplied by the frontend rather than declared in the source, so mypy
+# reports every use as undefined. Suppress only those; a name-defined error
+# for a name the user actually mistyped still reaches the report.
+_INTRINSIC_UNDEFINED = re.compile(
+    r'error: Name "(__ESBMC_\w+|__VERIFIER_\w+|nondet_\w+|__loop_invariant)"'
```
lucasccordeiro: `__ESBMC_\w+` and `nondet_\w+` also suppress misspelled intrinsics such as `__ESBMC_assme`. Match an explicit list of known intrinsic names instead.

unit/python-frontend/test_typecheck_intrinsics.py
```diff
@@ -0,0 +1,54 @@
...
+              'main.py:3: error: Name "__ESBMC_return_value" is not defined'
+              '  [name-defined]\n'
+              'main.py:4: error: Name "nondet_int" is not defined'
+              '  [name-defined]\n'
+              'Found 3 errors in 1 file (checked 1 source file)\n')
+
+    assert _without_intrinsic_diagnostics(1, report) == (0, "")
+
+
+def test_user_typo_survives_alongside_intrinsics():
```
lucasccordeiro: `typo_name` cannot match the intrinsic regex, so this does not exercise the real gap. Add a misspelled intrinsic such as `__ESBMC_assme`.

LukeW1999:
All six addressed in `fa77c22e0c`. The allow-list one turned up something worse than the hole you named.

**Allow-list instead of deny-list.** Confirmed first: `__ESBMC_requires((m := n) > 0)` was accepted and lowered to `ASSUME m > 0`. A generator was worse. `__ESBMC_requires(all(x > 0 for x in l))` reached the wrapper as:

```
ASSUME ESBMC_all_1 // contract requires
```

The preprocessor had rewritten `all(genexp)` into a loop plus a temporary, so the clause stood over a body-local symbol the wrapper never declares — the Python counterpart of #6941, arrived at from the other direction.

The gate could not have caught it: by the time the converter ran, the argument was a plain `Name`. I found it by printing the node kinds `check_contract_clause` actually receives, which showed only the `ensures` clause arriving; the `requires` call never reached the gate at all.

So two changes rather than one. The gate is an allow-list now, covering comprehensions, the walrus operator and whatever nobody enumerated. Operator and context nodes carry their own `_type` (`And`, `Lt`, `Load`, …), so they are listed too — I checked what `ast2json` emits rather than guessing, since omitting them would have rejected every legal clause. And `_lower_listcomp_in_expr` leaves a contract clause alone, so the clause reaches the gate as written.

`clause_genexp_fail` and `clause_walrus_fail` pin both.

**Module-scope fallback**, **shared `contracts_enabled`**, **shared `returns_no_value`**: done as described. The predicate now lives on `optionst`, so goto_sideeffects and builtin_functions read the same one.

**The mypy regex** and **the unit test**: both fixed, and you were right that the test was the weaker of the two. `typo_name` could never have matched the pattern it was meant to exercise, so it proved nothing. It now uses `__ESBMC_assme` and `nondet_intt`, and the suppression matches an explicit intrinsic list. Verified end to end: a misspelled intrinsic still reaches the report, correct ones stay quiet.

**Complexity.** You were right and I had misread the report — I took the `strict` column for the limit, when the gate is `core > 15` and flags any function that *rises* while over threshold, not just new ones. `check_contract_clause` is under it now, and the additions to `get_expr`, `do_function_call_symbol`, `build` and `visit_Call` are extracted into helpers so those four return to their base counts. `get_expr` keeps +1: paying that back means refactoring code this PR does not otherwise touch, inside a 283-CCN function. Say the word if you want it at zero.

Suites: `python-contracts` 21/21, `function_contract` 346/346, ~170 tests sampled across `python` plus the 22 comprehension-specific ones, a `numpy` slice, clang-format-11 and yapf clean.

src/python-frontend/function_call/builder.cpp
LukeW1999: Inverted in e361c23359. `is_pure_clause_node` is now an allow-list, so `ListComp`, `SetComp`, `DictComp`, `GeneratorExp` and `NamedExpr` are all rejected rather than only `Call`/`Subscript`. `clause_genexp_fail` and `clause_walrus_fail` pin the two you named.

src/python-frontend/function_call/builder.cpp
LukeW1999: Added the module-scope fallback in e361c23359: the function-scoped lookup now falls back to `sid.global_to_string()` before giving up.

Leaving this unresolved because I could not pin it with a test. `clause_untyped_fail` covers an unannotated *parameter*, not a global, and I could not construct a module-level global that actually reaches `any_type()` — the ones I tried (`g = make(1)` with an unannotated `make`, and `g: object = ...`) both got a concrete type inferred, so they pass the check without exercising the fallback. If you have a case in mind that yields an `any_type` global, I will add it.

src/goto-programs/builtin_functions.cpp
LukeW1999: Extracted in e361c23359 as `optionst::contracts_enabled()` in `util/config/options.cpp`, holding the four-option predicate once. Both sites call it now: `drop_inactive_contract_clause` here and `goto_sideeffects.cpp:1495`.

src/python-frontend/converter/converter_expr.cpp
LukeW1999: Shared in e361c23359. `returns_no_value` now has one definition, in `util/lang/python_types.cpp`, and both `converter_expr.cpp:489` and `builder.cpp:213` call it, so the two None-return rules cannot diverge.

src/python-frontend/parser/typecheck.py
LukeW1999: Replaced the wildcards with an explicit `_INTRINSICS` tuple in e70829aa66, matched by exact name (`re.escape` over each entry). A misspelling such as `__ESBMC_assme` or `nondet_intt` no longer matches and stays in the report.

unit/python-frontend/test_typecheck_intrinsics.py
LukeW1999: Added `test_misspelled_intrinsic_still_reaches_the_report`, using the `__ESBMC_assme` you suggested plus `nondet_intt`. It asserts both survive the filter while a real `__ESBMC_ensures` diagnostic is suppressed, and that the exit status stays 1.

Issue #6941, opened 2026-08-11
[contracts] a clause containing a call binds to a body-local temporary, breaking both contract modes


LukeW1999:
A contract clause containing a function call binds to an SSA temporary declared in the *original function body*. The wrapper never declares that temporary, so the clause is left over a free symbol. Both contract modes then report false positives, and the clause constrains nothing it was written to constrain.

Measured on `42aa5ce8be` (ESBMC 8.4.0). Pre-existing; not related to any open PR.

## Repro

```c
int helper(int x) { return x + 1; }

int f(int x)
{
  __ESBMC_requires(helper(x) > 100);
  __ESBMC_ensures(__ESBMC_return_value > 100);
  return x + 1;
}

int main(void)
{
  int n;
  __ESBMC_assume(n > 200);
  return f(n);
}
```

The contract holds: `n > 200` gives `helper(n) = n + 1 > 100`, and the returned `n + 1 > 100`.

```
$ esbmc t.c --enforce-contract f
  return_value > 100
VERIFICATION FAILED
```

Rewriting the same condition without the call is the control, and it isolates the cause exactly:

```c
  __ESBMC_requires(x + 1 > 100);   /* same condition, no call */
```

```
$ esbmc t.c --enforce-contract f
VERIFICATION SUCCESSFUL
```

## What the wrapper actually gets

```
f (c:@F@f):
        ASSUME return_value$_helper$1 > 100 // contract requires
        DECL signed int return_value;
        FUNCTION_CALL:  return_value=contracts_original_f(x)
        ASSERT return_value > 100 // contract ensures
```

`return_value$_helper$1` is declared inside `__ESBMC_contracts_original_f`, not in the wrapper. The precondition is therefore an assumption over an unconstrained symbol, which says nothing about `x`. The body is checked with `x` free, so the `ensures` fails.

## Both modes are affected, symmetrically

| mode | clause becomes | consequence |
| --- | --- | --- |
| `--enforce-contract` | `ASSUME <free symbol>` | too weak: constrains nothing, so the body is checked under fewer assumptions than the contract states → false positive on `ensures` |
| `--replace-call-with-contract` | `ASSERT <free symbol>` | too strong: unprovable for an arbitrary value, so **every** call site is rejected |

The replace side is worth stating plainly: a caller that *satisfies* the precondition is still rejected.

```c
int main(void) { return f(500); }   /* helper(500) = 501 > 100, satisfied */
```

```
[...]

PR #6942, opened 2026-08-11
[python] Complete the scalar contract MVP

LukeW1999:
Stacked on #6940, which brings the clauses up. This one takes the result to a usable MVP: sweeping the checklist in #6938 against the running build, fixing the one real gap it found, and documenting the whole surface.

**Base is `python-scalar-contracts`, not `master`.** Review #6940 first; this PR's own diff is the last commit.

## The gap it found

A short name is resolved by the `c:@F@<name>` convention before any other mode's symbols. The operational models define plenty of ordinary names, so a Python method called `add` resolved to `umath.c`'s `add`:

```python
class Counter:
    def add(self, k: int) -> int:
        __ESBMC_requires(k > 0)
        __ESBMC_ensures(__ESBMC_return_value > 0)
        return k
```

```
$ esbmc counter.py --enforce-contract add
ERROR: --enforce-contract: cannot use 'add': that function declares no contract clauses
```

The message is doubly misleading: the function does declare clauses, and the symbol reported on is not the one the user wrote. `find_function_symbol` now prefers whichever same-named candidate actually carries a contract, which is by definition the one that was annotated.

C is unaffected. Where no candidate carries a contract the previous precedence and error apply unchanged, and the 346-test `function_contract` suite passes. Two annotated candidates sharing a name still reach the existing ambiguity report rather than being silently resolved.

## Three questions settled by measurement

**Methods already work.** Both by bare name and by full symbol ID (`py:counter.py@C@Counter@F@add`). Only the name resolution above stood in the way. This matters more than it sounds: Python code is class-heavy, and a contract feature that could not annotate a method would not be an MVP.

**`__ESBMC_old` is unnecessary for parameters.** Python passes scalars by value, so a parameter named in `ensures` already denotes its entry value, even when the body reassigns it:

```python
def bump(x: int) -> int:
    __ESBMC_requires(x > 0)
    __ESBMC_ensures(__ESBMC_return_value == x + 1)   # x is the entry value
    x = x + 1
    return x
```

`pre_state_param` pins this and its `_fail` twin proves it is genuinely checked. That removes P1.3 from the MVP path entirely rather than deferring it.

**Several clauses of each kind combine, and the logical operators lower correctly.** Two `requires`, two `ensures`, and `and` / `or` / `not` all behave, each with a control that fails when the second clause is violated.

## Also

Contract intrinsics Python does not lower yet (`__ESBMC_assigns`, `__ESBMC_old`, `__ESBMC_is_fresh`, `__ESBMC_forall`, `__ESBMC_exists`) now say so:

```
ERROR: __ESBMC_assigns at line 5 is not supported in Python contracts yet
```

Previously they reached the undefined-function stub and failed as `Unsupported function '__ESBMC_assigns' is reached`, which is loud but explains nothing about contracts.

## Documentation

A `Python contracts` section in `website/content/docs/function-contracts.md`: the syntax, methods, why `__ESBMC_old` is absent, the pure-expression rule with the three constructs it rejects and why each one had to be rejected, and a table of the supported surface including what is deliberately missing. Every claim in it was run against the build rather than inferred.

The section also points at #6941 for the call-in-clause restriction, since that is a general contract defect rather than a Python one: the same clause in C is silently mis-verified instead of rejected.

## Tests

`regression/python-contracts` goes from 19 to 28. New in this PR:

| test | pins |
| --- | --- |
| `method_contract` / `_fail` | the name-resolution fix, on a method whose name collides with a C model |
| `pre_state_param` / `_fail` | a parameter in `ensures` denotes its entry value |
| `multiple_requires` | the second `requires` is in force, not just the first |
| `multiple_ensures` / `_fail` | both postconditions are checked |
| `clause_logic_ops` | `and` / `or` / `not` |
| `unsupported_assigns_fail` | the new diagnostic |

`multiple_requires` is built so that it can only pass if the second clause is assumed, rather than asserting a verdict that holds either way.

Also run: `function_contract` 346/346, an 80-test sample across the `python` suite, clang-format-11 clean on changed lines. Complexity: `find_function_symbol` 8 → 16, within its strict limit; nothing else moved.

## MVP status

Working: `requires` / `ensures` with several of each, `__ESBMC_return_value`, `and` / `or` / `not`, both contract modes, `int` / `float` / `bool` parameters, methods, globals in clauses, and `list` / `str` / unannotated parameters where the clause does not mention them.

Absent by design, each with a diagnostic that names it: `__ESBMC_old`, `__ESBMC_assigns`, `__ESBMC_is_fresh`, quantifiers, and `--enforce-all-contracts`.

Refs #6938


LukeW1999:
Added `FUTURE` tests for every construct the frontend rejects, so the gaps are executable rather than only described in prose. `regression/python-contracts` is now 33 tests: 28 CORE and 5 FUTURE.

| test | pins |
| --- | --- |
| `future_old` | `__ESBMC_old` in an `ensures` clause |
| `future_assigns` | an `__ESBMC_assigns` clause |
| `future_len_in_clause` | `len()` in a `requires` clause, which #6944 unblocks |
| `future_subscript_in_clause` | a subscript in an `ensures` clause |
| `future_enforce_all` | `--enforce-all-contracts`, needing a Python equivalent of the `__ESBMC_contract` attribute |

`FUTURE` means the regex is the verdict wanted *after* the feature lands, and the current run must not match it. So each test flips the moment support arrives, instead of the gap being tracked only in the docs.

I checked the mode discriminates rather than passing unconditionally: pointing `future_old` at output the run does produce makes it exit 77 with `Test 'python-contracts/future_old' passed but is marked as FUTURE. Consider reclassifying it as CORE.` Reverted after checking.

The docs now say each unsupported construct has such a test, matching how the C limitations section already refers to its KNOWNBUG tests.

LukeW1999:
Correcting my previous comment: two of the five `FUTURE` tests could not have flipped, so `regression/python-contracts` is 32, not 33.

**`future_subscript_in_clause` was testing the wrong thing.** Its precondition was `len(l) > 0`, so the run failed on the *call*, not on the subscript the test is named for. Implementing subscripts alone would not have flipped it, and it duplicated `future_len_in_clause`. Its precondition is now pure, and it fails on the subscript as intended.

**`future_enforce_all` is removed.** It expected SUCCESSFUL from `--enforce-all-contracts` on a file carrying no marker, but the flag selects functions by the `__ESBMC_contract` attribute. Whatever Python surface that marker eventually gets — a decorator is the obvious candidate — the source has to carry it, so implementing the flag would leave the run reporting `no function carries the __ESBMC_contract annotation` and the test green forever. A `FUTURE` test cannot state a desired verdict before the surface is chosen; that gap stays in #6938's checklist and in the docs.

What I did check about the mode, since "all FUTURE tests pass" would otherwise mean nothing:

- it is registered on every platform in the main CI. `CORE_REGRESSION_ONLY` defaults off, and only the sanitizer workflow narrows the modes
- exit 77 is a genuine failure here, not a skip: the suite sets `SKIP_RETURN_CODE 10`
- all four remaining tests fail fast with a specific `ERROR`, none by timeout. That matters, because the harness deliberately accepts a timeout as satisfying a `FAIL_MODES` expectation, which is a quiet way for such a test to become meaningless
- the mode discriminates after the edit: pointing the test at output the run does produce makes it exit 77 with "passed but is marked as FUTURE"

src/goto-programs/contracts/contracts.cpp
```diff
@@ -175,21 +175,63 @@ bool code_contractst::is_compiler_generated(
   return false;
 }
 
+bool code_contractst::declares_contracts(const symbolt &func_sym) const
+{
+  auto it = goto_functions.function_map.find(func_sym.id);
+  if (it == goto_functions.function_map.end() || !it->second.body_available)
```
lucasccordeiro: An annotated function with no body returns false here, because the early return skips `is_annotated_contract_function`. Test the annotation before `body_available`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -175,21 +175,63 @@ bool code_contractst::is_compiler_generated(
...
+    return false;
+
+  return has_contracts(it->second.body) ||
+         is_annotated_contract_function(func_sym);
+}
+
 symbolt *code_contractst::find_function_symbol(const std::string &function_name)
```
lucasccordeiro, line 208: `find_function_symbol` is now CCN 16, over the `core` gate of 15, so the complexity job fails. Extract the annotated-candidate scan into a helper.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -175,21 +175,63 @@ bool code_contractst::is_compiler_generated(
...
-    return sym;
+
+  // A short name is resolved against every mode's symbols, and the operational
+  // models define plenty of ordinary names. Whichever candidate actually
+  // carries a contract is the one the user annotated, so prefer it over
+  // whatever the naming conventions below happen to match first: a Python
+  // `add` never has the id `c:@F@add`, but umath.c does.
+  symbolt *annotated = nullptr;
+  bool annotated_ambiguous = false;
+  forall_goto_functions (it, goto_functions)
```
lucasccordeiro: This loop duplicates the short-name scan below, differing only by the `declares_contracts` filter. Factor both into one helper taking a predicate.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -175,21 +175,63 @@ bool code_contractst::is_compiler_generated(
...
+  {
+    symbolt *candidate = context.find_symbol(it->first);
+    if (
+      candidate && candidate->get_type().is_code() &&
+      id2string(candidate->name) == function_name &&
+      declares_contracts(*candidate))
+    {
+      if (annotated != nullptr)
+      {
+        annotated_ambiguous = true;
```
lucasccordeiro: No regression test covers two annotated same-named candidates reaching the ambiguity report. Add one pinning that branch.

src/python-frontend/function_call/builder.cpp
```diff
@@ -113,10 +113,27 @@ typet function_call_builder::enclosing_return_type() const
...
+         name == "__ESBMC_is_fresh" || name == "__ESBMC_forall" ||
+         name == "__ESBMC_exists";
+}
+
 void function_call_builder::check_contract_call(
   const symbol_id &function_id) const
 {
   const std::string &clause = function_id.get_function();
+
+  if (is_unsupported_contract_intrinsic(clause))
```
lucasccordeiro: `check_contract_call` runs on every call, so this throws outside any clause while reporting "in Python contracts". Scope the check to contract clauses.

LukeW1999:
All five addressed in `abf2c58886`. The branch is rebased on the updated #6940, so its diff is the last two commits.

**`declares_contracts` and a bodyless annotated function.** Confirmed by reading: the early return on `body_available` ran before `is_annotated_contract_function`, so the annotation was never consulted. The annotation lives on the symbol, so it is tested first now and only the clause scan needs a body.

**Complexity.** You were right, and I had misread the report — I took the `strict` column for the limit rather than the `core > 15` gate, and the gate also flags any function that *rises* while over threshold, not only new ones. `find_function_symbol` is back under it.

**The duplicated scan.** Both are one `short_name_candidates(short_name, predicate)` now. That also removed the awkward `annotated_ambiguous` flag: the annotated pass returns only when it finds exactly one candidate, and an empty result is what gates the convention lookups.

**The untested ambiguity branch.** `ambiguous_contract_name_fail` covers it: a method `Counter.bump` and a free `bump`, both carrying contracts.

```
ERROR: Ambiguous function name 'bump'; use a full symbol ID to disambiguate.
Candidates: ...@C@Counter@F@bump, ...@F@bump
```

**`check_contract_call` scope.** Rather than scope it, I reworded it. The check does run on every call, but the names it rejects (`__ESBMC_assigns`, `__ESBMC_old`, `__ESBMC_is_fresh`, the quantifiers) have no meaning outside a contract either, so refusing them anywhere is right; only the wording was wrong. It now says the frontend does not support the name yet, which holds wherever the call appears. Scoping properly would need the builder to know whether the enclosing function declares clauses, which it does not track — happy to add that if you would rather have it.

Suites: `python-contracts` 35/35, `function_contract` 346/346, an 80-test sample across `python`, clang-format-11 clean. The complexity gate is clean apart from `get_expr` at +1, inherited from #6940 and explained there.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Reordered in 99e959197a. `is_annotated_contract_function` is now tested first and returns early, before any `function_map` lookup, so a bodiless annotated function is recognised. Only the clause scan needs a body.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Extracted the scan into `short_name_candidates` in 99e959197a. `find_function_symbol` is CCN 10 now, under the `core` gate of 15, and the complexity job reports "No function this PR adds or worsens exceeds its threshold".

src/goto-programs/contracts/contracts.cpp
LukeW1999: Factored into one helper in 99e959197a: `short_name_candidates(short_name, accept)` takes the predicate, and the two scans are now calls to it — one passing `declares_contracts`, the other passing an accept-all lambda.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Added `regression/python-contracts/ambiguous_contract_name_fail` in 99e959197a: an annotated `Counter.bump` method and an annotated module-level `bump`. Both are collected as annotated candidates, so the convention lookups are skipped and the ambiguity report names both:

```
ERROR: Ambiguous function name 'bump'; use a full symbol ID to disambiguate. Candidates: ...@C@Counter@F@bump, ...@F@bump
```

src/python-frontend/function_call/builder.cpp
LukeW1999: Scoped in 4b21d91db7. Only `__ESBMC_assigns` is reported from here now, because it is a statement in its own right and nothing else sees it. The other four are meaningful only inside a clause, where `check_contract_clause` already rejects them by name with a clause-scoped message that says which call was the problem.

So a stray `__ESBMC_old` outside any clause is no longer a contracts diagnostic at all; it falls through to the ordinary undefined-function path, which is what it is.

Pinned both sides, neither had a CORE test before:

- `intrinsic_outside_clause_fail` covers the case that changes. It fails on the unfixed build, which aborts conversion and never reaches `VERIFICATION FAILED`.
- `clause_old_fail` is the control, holding the clause-walker message in place.

PR #6944, opened 2026-08-11
[contracts] Report a clause call instead of lifting it

### LINKED ISSUE #6941: [contracts] a clause containing a call binds to a body-local temporary, breaking both contract modes

LukeW1999:
Fixes #6941 by refusing the case rather than handling it. The branch previously lifted a clause's calls into the wrapper; that is withdrawn, and the requirements it turned up are recorded in #6960.

## The defect

A clause is lowered into a single `ASSUME`/`ASSERT`, so a call inside it survives only as the SSA temporary the frontend bound its result to. That temporary is declared in the function body, never in the wrapper or at a replaced call site, so the clause was left over a free symbol.

```c
int helper(int x) { return x + 1; }

int f(int x)
{
  __ESBMC_requires(helper(x) > 100);   /* VERIFICATION FAILED */
  __ESBMC_ensures(__ESBMC_return_value > 100);
  return x + 1;
}
```

Writing the same condition as `x + 1 > 100` reports SUCCESSFUL. Both modes were affected: enforce assumed the free symbol and constrained nothing, replace asserted it and rejected every call site including ones that satisfy the precondition.

## What this PR does

Refuses such a contract, naming the callee:

```
ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@helper',
whose result the clause cannot name outside the function body
```

Two silent false-positive classes become one diagnostic. Contract intrinsics keep their own materialisation and are excluded, so `__ESBMC_is_fresh`, `__ESBMC_old` and the quantifiers are unaffected.

## Why not lift the call

The first version of this branch did lift it, and review found that unsound in ways that are not detail fixes.

A clause may guard its own call: `__ESBMC_requires(d != 0 && q(d) > 0)`. The lifted `q(d)` ran before the ASSUME, reporting a division by zero inside `q` that no caller can reach — a fix for one false positive introducing another. It cannot be reordered, because the ASSUME names the temporary the lift defines, so the call has to precede the assumption that would make it safe.

Lifted operands received parameter substitution only, so a clause combining a call with `__ESBMC_old` fed the wrapper an unassigned snapshot temporary; on the replace side that reached an ASSUME.

The purity gate added to justify the lifting did not hold either. A global written as `g = h()` lowers to FUNCTION_CALL rather than ASSIGN and was invisible to it; `free` and `printf` lower to OTHER and counted as pure; and it never ran for `--enforce-contract '*'` or the `*-all-contracts` flags, which is how this suite spells it throughout. It promised a guarantee it did not deliver, which is worse than not offering one. That last point is filed separately as #6959, since it affects every diagnostic behind that gate, not just this one.

## Tests

`function_contract` is 351/351. The five new tests pin the diagnostic on the enforce side, the ensures side, the replace side, and a nested call, plus one control:

| test | pins |
| --- | --- |
| `clause_call_requires` | the diagnostic, naming the callee |
| `clause_call_ensures` | same on the postcondition |
| `clause_call_replace` | same at a replaced call site |
| `clause_call_nested` | `outer(inner(x))` names the outer callee |
| `clause_call_global_ok` | a clause naming a global that a call assigns is **not** refused, since a global is nameable outside the body |

`clause_call_global_ok` asserts SUCCESSFUL, which is what proves no diagnostic fired; an earlier version of it asserted a verdict that folded to zero VCCs and so proved nothing.

No pre-existing test changed behaviour: the 346 tests that were passing before still pass, so the refusal rejects nothing the suite relies on.

Complexity gate clean, clang-format-11 clean.


LukeW1999:
The known limit in the PR description is now measured and pinned rather than only stated.

A lifted call is re-issued while the original body keeps its own copy, so a clause whose callee mutates state has that effect twice:

```c
int g = 0;
int bump(void) { g++; return g; }

int f(int x)
{
  __ESBMC_requires(bump() > 0);
  ...
}
```

Asserting `g == 1` fails; asserting `g == 2` succeeds. Exactly twice, once in the wrapper and once inside the original function.

This is not a regression. Before this PR the clause referenced a temporary the wrapper never declared, so the precondition constrained nothing at all — strictly worse. The double evaluation is newly *reachable*, not newly broken.

`regression/function_contract/clause_call_side_effect_knownbug` pins it, and #6945 records the two candidate designs, which differ in more than implementation: evaluate once by reusing the body's computation, or reject a side-effecting clause outright. I did not decide that inside this PR because the second option needs a purity analysis that does not exist at this point in the pipeline, and the first changes how `ensures` threads a value out of the callee.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -309,6 +309,86 @@ code_contractst::extract_requires_from_body(const goto_programt &function_body)
...
+
+  forall_goto_program_instructions (it, original_body)
+  {
+    if (!it->is_function_call() || !is_code_function_call2t(it->code))
+      continue;
+
+    const code_function_call2t &call = to_code_function_call2t(it->code);
+    if (is_nil_expr(call.ret) || !is_symbol2t(call.ret))
+      continue;
+    if (!referenced.count(to_symbol2t(call.ret).thename))
```
lucasccordeiro: `call.ret` can be a global — `g = compute();` binds it directly — so this matches and emits `DECL g` into the wrapper, shadowing it. Restrict lifting to temporaries DECL'd in the body.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -309,6 +309,86 @@ code_contractst::extract_requires_from_body(const goto_programt &function_body)
...
+/// which is what the __ESBMC_is_fresh path has always done for its one
+/// hard-coded intrinsic.
+void code_contractst::lift_clause_call_temps(
+  const expr2tc &clause,
+  const goto_programt &original_body,
+  goto_programt &dest,
+  const std::vector<std::pair<expr2tc, expr2tc>> &substitutions) const
+{
+  std::set<irep_idt> referenced;
+  collect_symbol_names(clause, referenced);
```
lucasccordeiro: Only clause symbols are collected, so `outer(inner(x))` lifts `outer` while its argument `return_value$_inner$1` stays undeclared. Iterate to a fixpoint over the lifted calls' operands.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -309,6 +309,86 @@ code_contractst::extract_requires_from_body(const goto_programt &function_body)
 }
 
 // Helper structure to track assignments to a variable
+/// Contract intrinsics keep their own materialisation in the wrapper, which
+/// knows how to turn them into allocations, snapshots or quantifiers rather
+/// than a plain re-issued call.
+static bool is_contract_intrinsic(const std::string &funcname)
+{
+  return is_fresh_function(funcname) ||
+         funcname.find("__ESBMC_old") != std::string::npos ||
```
lucasccordeiro: Substring matching also skips user functions whose names merely contain these tokens. Compare the mangled identifier exactly instead of `find(...) != npos`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -1542,6 +1622,7 @@ goto_programt code_contractst::generate_checking_wrapper(
   // 3. Assume requires clause (after memory allocation for is_fresh)
   // Also replace __ESBMC_old() references in the requires clause — old snapshots
   // have already been materialized above (step 2), so replacement is safe here.
+  lift_clause_call_temps(requires_clause, original_body, wrapper);
```
lucasccordeiro: A lifted `requires` call runs before the body and re-executes its side effects, so the body is checked from a state no caller can reach. Reject impure clause calls.

regression/function_contract/clause_call_requires/main.c
```diff
@@ -0,0 +1,15 @@
+int helper(int x) { return x + 1; }
+
+int f(int x)
+{
+  __ESBMC_requires(helper(x) > 100);
```
lucasccordeiro, line 5: No test covers a nested clause call or a clause naming a global assigned by a call; both are mis-lifted today. Add a case for each.

lucasccordeiro (changes requested):

LukeW1999:
All five addressed in `194b45be0d`. Each was reproduced first, and the two structural ones turned out to produce wrong verdicts, not just odd GOTO.

**`call.ret` can be a global.** Confirmed. The wrapper emitted `DECL signed int g;` and the assumption then constrained a local the body never reads:

```c
int g;
int probe(void);
int f(int x) { g = probe(); __ESBMC_requires(g > 100); __ESBMC_ensures(g > 100); return x; }
```

`VERIFICATION FAILED` where the contract holds. Selection is now restricted to symbols DECL'd in the body, so a global is left alone. `clause_call_global` pins it.

**Nested clause calls.** Confirmed, and worse than it looked: `outer(return_value$_inner$1)` was emitted with the inner temporary never declared, which is exactly the free-symbol bug this PR exists to fix, one level down. Selection now iterates to a fixpoint over the operands of what it has already selected, and emits in body order so each temporary is defined before its use.

The discriminating case needed `--function f`; with a caller present, `main`'s own constraint on the argument masked it. `clause_call_nested` reports FAILED without the fix while an otherwise identical non-nested twin passes, so nesting is what the test isolates. `clause_call_nested_fail` is its negative control.

**Substring matching in `is_contract_intrinsic`.** Fixed, it compares the base name after the last `@` now.

**Reject impure clause calls.** Done rather than deferred. `callee_is_pure` is conservative: a callee with no body available, a write to anything but its own locals, or a call to something not itself pure all count as impure, with a visited set for recursion.

```
ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@bump',
which is not provably free of side effects
```

That settles the design question #6945 raised, so its KNOWNBUG becomes `clause_call_side_effect_rejected`, a CORE test pinning the diagnostic. I will close #6945 when this merges.

**Tests.** `function_contract` is 355/355, up from 351: `clause_call_nested`, `clause_call_nested_fail`, `clause_call_global`, and the converted rejection test.

One thing your review also caught indirectly: the complexity gate is `core > 15`, and I had read the report's `strict` column as the limit. `diagnose_contract_target` went to 16 and is back to 14 by extracting `impure_clause_reason`. The gate now emits no error locally.

LukeW1999:
@lucasccordeiro this is a substantial retreat from what I told you last round, so flagging it directly rather than leaving it in the diff.

You asked me to reject impure clause calls instead of deferring the question, and I said I had done it conservatively. A review of that code found the gate did not hold in three independent ways, all reproduced:

- a global written as `g = h()` lowers to `FUNCTION_CALL`, not `ASSIGN`, so the check never saw it
- `free` and `printf` lower to `OTHER`, so a callee doing either counted as pure
- the gate never ran for `--enforce-contract '*'` or the `*-all-contracts` flags, which is how this suite spells it throughout

The same review found the lifting itself unsound for reasons that are not detail fixes: a clause guarding its own call (`__ESBMC_requires(d != 0 && q(d) > 0)`) had the call hoisted above the ASSUME and reported a division by zero no caller can reach, and lifted operands never received the `__ESBMC_old` substitution, which is unsound on the replace side rather than merely noisy.

So the lifting is withdrawn and the branch keeps only the diagnostic. That is smaller than what you asked for, and I would rather say so than ship a gate that reads like a guarantee and is not one. The requirements the attempt turned up are in #6960 so the next attempt starts from them, and the wildcard bypass is #6959 since it affects every diagnostic behind that gate, not only mine.

#6945 is closed: without lifting there is no double evaluation.

PR #6947, opened 2026-08-11
[svcomp] Falsify with one bounded round before the concurrency strategy

lucasccordeiro:
Stacked on #6939 (base is that branch, not master).

Measured on the 340 concurrent CORE tests under the wrapper's own flags rather than the tests' own — the two disagree, because `--no-por` comes from `test.desc`, not the wrapper. 22 of them strand at base; `--falsify-context-bound 1` recovers one (`03_microbenchmark`: no verdict in 120 s → FAILED in 0.71 s, matching the verdict its `test.desc` expects) and costs a median 0.02 s on the tests that already answer. No verdict moved in 683 runs across both flag sets, and the pre-pass cannot claim a proof by construction.

`W4.2` in the plan records both configurations, the cost breakdown, why N = 2 adds nothing, and that the score delta itself still needs a competition run.

Refs #6831.


lucasccordeiro:
This PR (900s): 
```
  correct:             899
    correct true:      524
    correct false:     375
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2277
  Score:              1423 (max: 5723)

https://github.com/esbmc/esbmc/actions/runs/31528824161.
```

Master (900s): 
```
  correct:             896
    correct true:      525
    correct false:     371
  incorrect:             6
    incorrect true:      0
    incorrect false:     6
  unknown:            2274
  Score:              1325 (max: 5723)

https://github.com/esbmc/esbmc/actions/runs/31528946776.
```

PR #6962, opened 2026-08-12
[contracts] Havoc what an assigns clause names, not the pointer to it

### LINKED ISSUE #6961: [contracts] --replace-call-with-contract aborts on an assigns clause naming a pointer parameter passed an array
## Summary

`--replace-call-with-contract` aborts with

```
ERROR: assignment to address_of not handled
```

(core dumped, `symex_assign.cpp:703`) whenever the callee's `__ESBMC_assigns`
clause names a **pointer or array parameter** and the caller passes an
**array**. No other flag is needed.

## Minimal reproducer

```c
#define N 4
void clr(int *p)
{
  __ESBMC_assigns(p);
  __ESBMC_ensures(p[0] == 0);
  for (int i = 0; i < N; i++)
    p[i] = 0;
}
int main(void)
{
  int b[N];
  b[0] = 1;
  clr(b);
  __ESBMC_assert(b[0] == 0, "z");
  return 0;
}
```

```
esbmc s_d.c --replace-call-with-contract clr --unwind 10
```

An array-typed parameter (`void clr(int a[N])` with `__ESBMC_assigns(a)`)
aborts identically.

## What does and does not trigger it

| `assigns` target | call argument | result |
|---|---|---|
| `p`, parameter `int *p` | array `b` | **abort** |
| `a`, parameter `int a[N]` | array `b` | **abort** |
| `p`, parameter `int *p` | `malloc`'d pointer | no abort |
| `g`, a file-scope array | — | `VERIFICATION SUCCESSFUL` |
| `x->coeffs`, struct member array | `&p` | `VERIFICATION SUCCESSFUL` |

So the trigger is specifically *parameter named in `assigns`* + *array
argument*.

## Cause

`code_contractst::replace_call_with_contract` builds the precise havoc by
substituting formals with actuals (`contracts.cpp:4705-4753`):

```cpp
expr2tc instantiated_target = target_expr;
...
instantiated_target = replace_symbol_in_expr(instantiated_target, param_symbol, actual_args[i]);
...
t->code = code_assign2tc(instantiated_target, rhs);
```

`clr(b)` decays the argument to `&b[0]`, so substituting `p := &b[0]` leaves
`instantiated_target` an `address_of2t`. That reaches `symex_assign_rec`,
whose dispatch has no `address_of` case, and it aborts.

The guard immediately above would have skipped it:

```cpp
if (config.options.get_bool_option("add-symex-value-sets") &&
    is_pointer_type(instantiated_target))
  continue;
```

but `add-symex-value-sets` is only set for `--k-induction`,
`--k-induction-parallel` and `--inductive-step`
(`command_line_options.cpp:582-584`), so it is off on a plain
`--replace-call-with-contract` run.

The sibling no-assigns path a few lines down (`contracts.cpp:4834`) gets this
right — it havocs `dereference2tc(pointee_type, actual_args[i])`, the pointee,
rather than the address.

## Not version-specific

Reproduced on `master` and on 8.1.0. This is long-standing, not a regression.

## Note on `--loop-invariant-check`

This was originally observed as "`--loop-invariant-check` +
`--replace-call-with-contract` crashes". That flag is not involved. In the
original setting the harness `main` had a 256-iteration loop, so without
`--loop-invariant-check` symex stopped at the unwinding assertion before ever
reaching the replaced call. Cutting the loop merely lets symex get far enough
to hit the bad assignment.

## Underlying question

Separately from the abort, it is not defined what `__ESBMC_assigns(p)` means
when `p` is a pointer parameter: the pointer variable itself (pass-by-value,
so invisible to the caller, and havocking it is a no-op) or the pointee region
(what mlkem-native's CBMC contracts express as `assigns(memory_slice(r, len))`).
Havocking only `*p` is still unsound for `p[1..n-1]` when the extent is
unstated, so whichever way the abort is fixed, the extent question needs its
own answer.

LukeW1999:
Fixes #6961.

## Symptom

`--replace-call-with-contract` aborted (core dumped) whenever the callee's
`__ESBMC_assigns` clause named a pointer parameter and the caller passed an
array:

```
ERROR: assignment to address_of not handled
```

The clause target was the *address* of the array rather than the array, so the
replacement tried to assign to an `address_of`.

## What the PR does

**Havoc the array, not its address.** A decayed array argument arrives as
`&b[0]`; `havoc_place` resolves it to `b`, so the whole array is havocked,
which also covers everything the callee reaches through the decayed pointer.

**Write through a pointer parameter, not to it.** A pointer parameter is
pass-by-value, so the callee cannot change the caller's pointer, only what it
points at. Havocking the argument itself both missed that write and invented a
bogus pointer for the `ensures` ASSUME to dereference.

**Do not follow a pointer that already names its place.** `&x` names the place
and `havoc_place` resolves it to `x`; following the pointer as well wrote
*through* `x`. For an `int **pp` argument that havocked `*x`, a dereference of
whatever the caller held — a null one was reported at the call site, blaming
the caller for a write the tool synthesised.

**One pointee-skip rule.** `havoc_through_pointer` and the no-assigns havoc at
2.4 each decided which pointees to leave alone, and had already drifted: 2.4
also skips a pointer pointee under `--add-symex-value-sets` and the assigns
path did not. Both call `skip_pointee_havoc` now.

**Test the value-set skip on the place actually written**, after widening and
after following a pointer parameter. The skip is about *assigning* a pointer,
not about *writing through* one; tested earlier it swallowed the whole frame,
so a pointer-variable argument was silently unhavocked. `--add-symex-value-sets`
is default-enabled by `--k-induction`, `--inductive-step` and
`--loop-invariant`, so that shape was unchecked in all three.

**Widen any constant index.** `&b[2]` is not the decay a compiler produces, but
it names a place inside `b` that the callee writes forward from, and the array
is the only object to widen to. That over-havocs `b[0..1]`, which loses the
caller information rather than granting it any.

**Say so when nothing can be written.** A frame target the contract named that
cannot be havocked used to disappear; a dropped target reads exactly like one
that was havocked, so it now warns.

## Tests

Ten, in `regression/function_contract/`.

| test | |
|---|---|
| `github_6961_assigns_array_param` / `_ptr_param` / `_ptr_param_fail` | the reported abort, now verified |
| `github_6961_assigns_ptr_param_valuesets_fail` | the value-set skip on a decayed array |
| `github_6961_assigns_ptr_var` | a pointer variable argument, written through |
| `github_6961_assigns_ptr_to_ptr` | `&r` where `r` is itself a pointer, no longer double-dereferenced |
| `github_6961_assigns_null_arg_fail` | a null argument, reported as 2.4 has always reported it |
| `github_6961_assigns_interior_ptr_fail` | `&b[2]` widened to the array |
| `github_6961_assigns_ptr_var_valuesets_fail` | a pointer variable under `--add-symex-value-sets` |
| `github_6961_assigns_ptr_var_frame_knownbug` | a pointer names no object to widen to, so only `*q` is reached |
| `github_6961_assigns_void_ptr_knownbug` | a void pointee names no object at all — **#7067** |

`function_contract` 380/380, `loop-invariants` 76/76, `python-contracts` 40/40,
and a uniform sample of the core C suite (295 of 1769) is clean.

## Note on the null-argument test

Checked against master before changing anything: the no-assigns pointer havoc
at 2.4 has always built `*actual_arg = nondet` with no null guard, so a null
argument is reported there too. The two paths now agree rather than one of them
being new. A contract meaning to accept null has to say `requires(p != 0)`.


src/goto-programs/contracts/contracts.cpp
```diff
@@ -4730,6 +4743,8 @@ void code_contractst::generate_replacement_at_call(
         is_pointer_type(instantiated_target))
         continue;
 
+      instantiated_target = havoc_place(instantiated_target);
```
lucasccordeiro: The value-set skip above tests the pre-widening target, so `&b[0]` is pointer-typed and the array is never havocked under `--add-symex-value-sets`. Call `havoc_place` before that check.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -931,6 +931,19 @@ static expr2tc unwrap_array_decay(const expr2tc &expr)
   return expr;
 }
 
+// The place a havoc writes to. Substituting a formal with an array argument
+// re-introduces the decay the collection step stripped once — `p` becomes
+// `&b[0]` — and an address is not a place to assign to. Widening back to the
+// whole array also covers everything the callee reaches through the decayed
+// pointer (#6961). Any other address_of names its own operand.
+static expr2tc havoc_place(const expr2tc &target)
```
lucasccordeiro, line 939: Only the decay pattern is widened: for `int *q = b; clr(q);` this still havocs the caller's own `q`, which is pass-by-value. Handle it, or add a test pinning the limitation.

regression/function_contract/github_6961_assigns_ptr_param/main.c
```diff
@@ -0,0 +1,19 @@
...
+  __ESBMC_ensures(p[0] == 0);
+
+  for (int i = 0; i < N; i++)
+    p[i] = 0;
+}
+
+int main(void)
+{
+  int b[N];
+  b[0] = 1;
```
lucasccordeiro, line 15: All three tests pass the array directly, so the decay is always syntactic. Add one passing a pointer variable, which is the shape that still misbehaves.

LukeW1999:
All three addressed in `f8fca50421` and `eba5d33c96`. Both were real, and the first is worse than a missed havoc.

### The value-set skip

Reproduced before touching anything. `github_6961_assigns_ptr_param_fail` is the test where the callee zeroes `b[1]` and `main` then asserts it survived, so the only correct verdict is FAILED:

```
                     plain                 --add-symex-value-sets
_assigns_ptr_param_fail   VERIFICATION FAILED   VERIFICATION SUCCESSFUL
```

The skip is not reachable only by someone who types the flag. `command_line_options.cpp:582` turns it on for `--k-induction`, `--k-induction-parallel` and `--inductive-step`, and `--loop-invariant` sets `k-induction` two lines above that. So every k-induction or loop-invariant run that also replaced a call left the array alone and reported SUCCESSFUL on it.

`havoc_place` now runs first, and the skip tests the widened place. A decayed array argument is an array again by then, so it is havocked; a pointer variable is still a pointer, so it is still skipped, which is what the comment meant. Pinned as `github_6961_assigns_ptr_param_valuesets_fail`.

### The pointer variable

`int *q = b; clr(q);` is worse than havocking the caller's `q`. It havocs `q`, and then the `ensures` ASSUME dereferences what it wrote:

```
State 1  line 17  q = invalid-object
Violated property: dereference failure: Incorrect alignment when accessing data object
VERIFICATION FAILED
```

on a program that is correct. This predates the branch — `havoc_place` is the identity on a bare symbol, and a build with master's `contracts.cpp` prints the same two lines — but the branch is where the shape belongs, so I fixed it here rather than filing it.

A pointer parameter is pass-by-value, so the callee cannot touch the caller's pointer, only what it points at. The havoc now follows the pointer, with the same void and function pointee guards the no-assigns path at 2.4 already uses. `github_6961_assigns_ptr_var` covers it: SUCCESSFUL, no dereference failure.

That reaches `*q` and stops there. Widening needs an object and a pointer names none, so `b[1]` still survives a replaced call that writes it. That one is `github_6961_assigns_ptr_var_frame_knownbug` rather than a comment, so it flips on its own the day the call site learns what the argument points at.

### The test that passes the array directly

Three tests were the decay case three times, yes. There are six now: the original three, the value-set one, and the two above.

`regression/function_contract` goes 356 to 359 and runs 358/359, with `github_4220` the only failure. It fails the same way on master (`dereference failure: array bounds violated: heap object` where the `.desc` wants `contract ensures`), so it is not from this branch. My build has Boolector and Bitwuzla off, which is the likeliest reason it does not show up for you.

LukeW1999:
The complexity gate caught the second commit: `code_contractst::generate_replacement_at_call` went 34 → 40, and it is already well past the `core > 15` threshold. Amended `d4f50177ba` rather than stacking a cleanup on top.

Parameter substitution moves out to `instantiate_assigns_target` and the new step to `havoc_through_pointer`. That takes four decision points out of `generate_replacement_at_call` and puts two back, so the gate now says "No function this PR adds or worsens exceeds its threshold" against the merge base. Behaviour is unchanged: same six `github_6961` verdicts, same 358/359 on `function_contract`.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Moved `havoc_place` above the skip in 9c1a461896, so the skip now tests the widened place. `github_6961_assigns_ptr_param_fail` reported SUCCESSFUL under `--add-symex-value-sets` while reporting FAILED without it; pinned as `github_6961_assigns_ptr_param_valuesets_fail`. Worth flagging that `--add-symex-value-sets` is default-enabled by `--k-induction`, `--inductive-step` and `--loop-invariant`, so this was silently dropping the havoc in all of those runs.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Handled the pass-by-value half in 81a72d741d: `havoc_through_pointer` writes through the pointer rather than havocking the caller's own `q`, so `int *q = b; clr(q);` no longer invents a bogus pointer for the ensures ASSUME to dereference.

Widening is the half that stays: it needs an object to name, and a pointer variable names none, so only `*q` is reached. Pinned as `github_6961_assigns_ptr_var_frame_knownbug` — `b[1]` keeps its pre-call value and the run reports SUCCESSFUL where the desc asks for FAILED.

regression/function_contract/github_6961_assigns_ptr_param/main.c
LukeW1999: Added `github_6961_assigns_ptr_var`, which passes `int *q = b` rather than the array, in 81a72d741d. Its counterpart `github_6961_assigns_ptr_var_frame_knownbug` pins the part that still misbehaves, so both the fixed and the unfixed halves of that shape are covered.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -994,6 +994,71 @@ static expr2tc unwrap_array_decay(const expr2tc &expr)
...
+{
+  expr2tc place = unwrap_array_decay(target);
+  if (is_address_of2t(place))
+    return to_address_of2t(place).ptr_obj;
```
rafaelsamenezes, line 1000: `&b[2]` is not the decay pattern, so this widens to a single element while the callee writes the whole tail through `p`. Handle it, or pin it as KNOWNBUG.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -994,6 +994,71 @@ static expr2tc unwrap_array_decay(const expr2tc &expr)
...
+// dereferences. Only the first element is reached this way; widening needs an
+// object, which only the decay case names. Nil when there is nothing to write
+// through, matching the void and function pointee skips at 2.4.
+static expr2tc havoc_through_pointer(const expr2tc &place, const namespacet &ns)
+{
+  if (!is_pointer_type(place))
+    return place;
+
+  type2tc pointee = ns.follow(to_pointer_type(place->type).subtype);
+  if (is_empty_type(pointee) || is_code_type(pointee) || is_nil_type(pointee))
```
rafaelsamenezes: These pointee skips duplicate 2.4's, which additionally skips pointer pointees under `--add-symex-value-sets`. Have 2.4 call this helper so the two cannot drift.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -994,6 +994,71 @@ static expr2tc unwrap_array_decay(const expr2tc &expr)
...
+    return expr2tc();
+
+  return dereference2tc(pointee, place);
```
rafaelsamenezes, line 1032: A null argument now gets `*p = nondet`: `clr(q)` with `q = 0` gives 0 VCCs on master, a NULL-dereference failure here. Guard the write, or add a test.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Done, thanks — and they had already drifted, exactly as you say: 2.4 also skipped a pointer pointee under `--add-symex-value-sets` and the assigns path did not. Both call `skip_pointee_havoc` now.

Chasing that down turned up a worse one next to it. `&x` names the place, and `havoc_place` resolves it to `x`, but `is_pointer_param` was still set so the pointer was followed a second time. For an `int **pp` argument that havocked `*x`, a dereference of whatever the caller held:

```
$ esbmc a2_ptrptr.c --replace-call-with-contract setp
✗ FAILED: dereference failure: NULL pointer      # blaming the caller for our own write
```

Fixed by not following a pointer that already names its place; `github_6961_assigns_ptr_to_ptr` pins it.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Checked this against master before changing anything, and it is not new: the no-assigns pointer havoc at 2.4 has always built `*actual_arg = nondet` with no null guard, so a null argument is reported there too.

```
$ esbmc nullarg_noassigns.c --replace-call-with-contract clr    # no assigns clause, on master
  dereference failure: NULL pointer
VERIFICATION FAILED
```

So the two paths now agree rather than one of them being new, and guarding only the assigns path would split them again — the opposite of your other comment. Pinned as `github_6961_assigns_null_arg_fail`, with the note that a contract meaning to accept null has to say `requires(p != 0)`.

If you would rather both paths guard the write, I am happy to do that instead — it is a one-line change in the shared helper now.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Pinned as `github_6961_assigns_interior_ptr_knownbug`, taking your second option.

Your comment sent me looking for siblings and there are two more, both of them false negatives rather than gaps:

- a pointer *variable* argument under `--add-symex-value-sets` is skipped entirely, so nothing is havocked. My `assigns_ptr_param_valuesets_fail` passes the array, which the widening rescues, so it never reached that branch — the test did not check what it claimed to.
- a `void *` parameter has no pointee to write through, and the target was dropped silently. It now warns, since a dropped frame target reads exactly like a havocked one.

All three are `KNOWNBUG` here and filed together as #7067.

PR #6963, opened 2026-08-12
[regression] Make the two audited-fix tests pass on Windows

lucasccordeiro:
github_1932 and github_2163 landed red on windows-latest and now fail every PR that merges master.

github_1932 cast argv[0] through `unsigned long`, 32-bit under LLP64, so the guard admitted non-null pointers; `uintptr_t` is lossless on every data model. github_2163 reports its assertion UNREACHED on Windows, where MSVC's `__security_cookie` has no definition, so it is marked THOROUGH until that is modelled.

Refs #6954


lucasccordeiro:
Superseded by #6958, which fixes both tests and goes further: it adds github_1932_lossy_cast_fail to pin the LLP64 truncation, and makes github_2163 target-independent rather than excluding it from Windows as this PR did.

PR #6964, opened 2026-08-12
[contracts] Snapshot the array, not one element per index

### LINKED ISSUE #4219: [contract] `ERROR: Unrecognized address_of operand` (SIGABRT) on `__ESBMC_old(arr[bound_var])` inside a quantifier ensures
ESBMC 8.2.0, Z3 v4.15.8 aborts (exit 134) during SMT encoding when an ensures predicate references the pre-state of an indexed buffer read (`__ESBMC_old(src[k])`) inside a `__ESBMC_forall` quantifier. The encoder hits an `Unrecognized address_of operand` on the chained `address_of(bitcast(concat(...)))` expression it builds for the quantified `__ESBMC_old`.

## Repro

```c
// /tmp/issue3.c
typedef unsigned long size_t;

__ESBMC_contract
void copy_buf(int *dst, const int *src, size_t n) {
    unsigned long __scribe_qv_0;
    __ESBMC_requires(__ESBMC_is_fresh(dst, n * sizeof(int)));
    __ESBMC_requires(__ESBMC_is_fresh(src, n * sizeof(int)));
    __ESBMC_requires(n <= 8);
    __ESBMC_ensures(__ESBMC_forall(&__scribe_qv_0,
        !(__scribe_qv_0 < n) || (dst[__scribe_qv_0] == __ESBMC_old(src[__scribe_qv_0]))));
    __ESBMC_assigns(*dst);

    __ESBMC_unroll(10);
    for (size_t i = 0; i < n; i++) dst[i] = src[i];
}
```

```
$ esbmc --function copy_buf --enforce-contract copy_buf --timeout 30s /tmp/issue3.c
...
ERROR: Unrecognized address_of operand:
address_of
* pointer_obj : bitcast
    * from : concat
        * forward : concat
            * forward : concat
                * forward : index
                    * source_value : symbol
                        * name : symex_dynamic::dynamic_2_array
                        ...
$ echo $?
134
```

## Expected

ESBMC should support `__ESBMC_old(expr)` where `expr` is a buffer-indexed read whose index is a quantifier-bound variable. The natural memcpy-style postcondition `forall k<n, dst[k] == old(src[k])` is unexpressible without it.

---

- Also emits `WARNING: migrate_expr: symbol 'tmp$5' missing renaming delimiters, treating as level0 with base name 'tmp$5'` during GOTO program creation — possibly related.
- Replacing `__ESBMC_old(src[k])` with `src[k]` (no `__ESBMC_old` wrap) avoids the abort. The `__ESBMC_old` wrapper is what causes the encoder to build the unrecognized expression.

LukeW1999:
Fixes #4219.

## Symptom

`__ESBMC_old(base[j])` inside `__ESBMC_forall` cannot be used. On `master` it
is rejected at GOTO conversion:

```
ERROR: cannot model call to `__ESBMC_old_raw' on a quantified variable inside
__ESBMC_forall/__ESBMC_exists
```

(#4219 recorded it as `Unrecognized address_of operand` + SIGABRT in the
encoder; the diagnosis has since moved earlier, the capability gap has not.)

That rules out the natural postcondition for any elementwise update —
`forall j < n, r[j] == old(r[j]) + b[j]` — which is exactly what
mlkem-native's `mlk_poly_add` / `mlk_poly_sub` need.

## Cause

`__ESBMC_old(x)` expands to `*(__typeof__(x)*)__ESBMC_old_raw((void*)&(x))`.
For `old(base[j])` the argument is `&base[j]`, whose value depends on `j`.
`remove_sideeffects_for_quantifier_body` hoists side effects out of the
quantifier so the body is left side-effect free; an address that only exists
under a binding cannot be hoisted, so it fails loudly rather than produce a
vacuous quantifier.

## Fix

Rewrite the operand rather than the mechanism:

```
old(base[j])   →   old(base)[j]
```

`old(base)` does not mention `j`, so the existing hoisting applies unchanged
and everything downstream — `old_snapshot` sideeffects,
`collect_old_snapshots_from_body`, `materialize_old_snapshots_at_wrapper`,
`replace_old_in_expr` — sees an ordinary snapshot that happens to be
array-typed. No contract-layer change.

Two details:

- **Snapshot the struct, not the member array.** `snap = (*r).coeffs` is a
  whole-array rvalue read through a dereference, which `dereference.cpp:1097`
  refuses ("Can't construct rvalue reference to array type"). Taking `*r` and
  re-applying `.coeffs` to the snapshot is a legal struct rvalue.
- **A pointer base is left alone.** With a symbolic extent there is no whole
  object to snapshot. That is #4219's own reproducer (`int *src`, `n <= 8`),
  so this PR does not close the general case — it closes the
  statically-sized-array case and leaves the abort in place for the rest.

## It is also much cheaper than the workaround

The only way to write these postconditions today is to expand the clause over
every index. Measured on mlkem-native's array-level polynomial functions,
`--function f --enforce-contract f --loop-invariant-check`:

| | encoding | VCCs | time | result |
|---|---|---|---|---|
| `mlk_poly_add` | 769 expanded terms | — | >900s | no result |
| `mlk_poly_add` | forall `requires`, expanded `ensures` | 9004 → 2842 | 534s | SUCCESSFUL |
| `mlk_poly_add` | **all forall (this PR)** | **3914 → 566** | **29s** | SUCCESSFUL |
| `mlk_poly_sub` | 1793 expanded terms | — | >900s | no result |
| `mlk_poly_sub` | **all forall (this PR)** | **3914 → 566** | **41s** | SUCCESSFUL |

The expansion materialises one `old` snapshot per index; this materialises
one. Both functions go from "does not terminate" to under a minute.

## Tests

- `github_4219_old_in_forall` — the reproducer, struct with a fixed array
- `github_4219_old_in_forall_fail` — the control

The control is what distinguishes a snapshot from a plain read: the same
shape with a body that adds the operand twice must still be rejected. If the
rewrite tracked the post-state instead of pinning the pre-state, the ensures
would hold and this test would go SUCCESSFUL.

350/350 `function_contract`, 76/76 `loop-invariants`, 87/87 `z3`, 35/35
`bitwuzla` — the four suites that exercise quantifiers. Complexity gate clean.

### DIFF
--- a/src/goto-programs/goto_sideeffects.cpp
+++ b/src/goto-programs/goto_sideeffects.cpp
@@ -87,6 +87,103 @@ static bool mentions_symbol(const exprt &e, const std::set<irep_idt> &ids)
   return false;
 }
 
+/// `__ESBMC_old(base[j])` with `j` bound by the enclosing quantifier takes the
+/// address of one element, which has no value outside the quantifier and so
+/// cannot be hoisted. Snapshot the array instead — `__ESBMC_old(base)[j]` —
+/// leaving an operand the bound variable does not reach, so the ordinary
+/// hoisting applies and the contract layer materialises one array snapshot in
+/// place of one per index (#4219). Only a base of array type is rewritten: a
+/// pointer with a symbolic extent has no whole object to snapshot.
+/// The `__ESBMC_old_raw` call under \p deref, which the `__ESBMC_old` macro
+/// wraps as `*(T*)__ESBMC_old_raw(&x)`, or nullptr if this is not one.
+static exprt *old_raw_call_under(exprt &deref)
+{
+  if (deref.id() != exprt::deref || deref.operands().size() != 1)
+    return nullptr;
+
+  exprt *call = &deref.op0();
+  while (call->id() == exprt::typecast && call->operands().size() == 1)
+    call = &call->op0();
+
+  if (
+    call->id() != "sideeffect" || call->statement() != "function_call" ||
+    call->operands().size() < 2 || !call->op0().is_symbol())
+    return nullptr;
+
+  const std::string callee = call->op0().identifier().as_string();
+  static const std::string old_raw = "__ESBMC_old_raw";
+  if (
+    callee.size() < old_raw.size() ||
+    callee.compare(callee.size() - old_raw.size(), old_raw.size(), old_raw) !=
+      0)
+    return nullptr;
+
+  return call->op1().operands().size() == 1 ? call : nullptr;
+}
+
+/// The array element \p addr addresses, when its index reaches \p bound_vars
+/// and its base is an array. A pointer with a symbolic extent has no whole
+/// object to snapshot, so it is left alone.
+static const exprt *
+bound_array_element(const exprt &addr, const std::set<irep_idt> &bound_vars)
+{
+  const exprt *e = &addr;
+  while (e->id() == exprt::typecast && e->operands().size() == 1)
+    e = &e->op0();
+  if (e->id() != exprt::addrof || e->operands().size() != 1)
+    return nullptr;
+
+  const exprt &target = e->op0();
+  if (target.id() != exprt::index || target.operands().size() != 2)
+    return nullptr;
+
+  if (
+    !mentions_symbol(target.op1(), bound_vars) ||
+    !target.op0().type().is_array())
+    return nullptr;
+
+  return &target;
+}
+
+static void
+lift_old_over_bound_index(exprt &expr, const std::set<irep_idt> &bound_vars)
+{
+  Forall_operands (it, expr)
+    lift_old_over_bound_index(*it, bound_vars);
+
+  exprt *call = old_raw_call_under(expr);
+  if (!call)
+    return;
+
+  const exprt *target =
+    bound_array_element(call->op1().operands()[0], bound_vars);
+  if (!target)
+    return;
+
+  // Snapshotting an array through a dereference is an rvalue array read, which
+  // the dereference layer refuses. Where the array is a struct member, take the
+  // struct — a legal rvalue — and re-apply the member to the snapshot.
+  const exprt base = target->op0();
+  const exprt index = target->op1();
+  const bool via_member = base.id() == exprt::member;
+  const exprt object = via_member ? base.op0() : base;
+
+  exprt whole_snapshot = *call;
+  whole_snapshot.op1().operands()[0] = address_of_exprt(object);
+
+  typet object_ptr("pointer");
+  object_ptr.subtype() = object.type();
+  exprt snapshot =
+    dereference_exprt(typecast_exprt(whole_snapshot, object_ptr), object_ptr);
+
+  if (via_member)
+    snapshot = member_exprt(
+      snapshot, to_member_expr(base).get_component_name(), base.type());
+
+  index_exprt element(snapshot, index);
+  expr.swap(element);
+}
+
 /// A side effect other than a nested function call (e.g. ++ on a parameter)
 /// cannot be replicated by argument substitution.
 static bool has_non_call_sideeffect(const exprt &e)
@@ -1008,6 +1105,9 @@ void goto_convertt::convert_quantifier_calls(exprt &expr)
       // Bottom-up: convert any nested quantifier calls first, then summarize
       // the remaining calls so the bound variable stays free in the body.
       convert_quantifier_calls(args[1]);
+      const irep_idt bound = quantifier_bound_var_id(args[0]);
+      if (!bound.empty())
+        lift_old_over_bound_index(args[1], {bound});
       inline_calls_in_quantifier_body(args[1], max_quantifier_inline_depth);
       if (!has_sideeffect(args[1]))
       {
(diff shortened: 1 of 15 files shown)


src/goto-programs/goto_sideeffects.cpp
```diff
@@ -87,6 +87,103 @@ static bool mentions_symbol(const exprt &e, const std::set<irep_idt> &ids)
...
+  if (e->id() != exprt::addrof || e->operands().size() != 1)
+    return nullptr;
+
+  const exprt &target = e->op0();
+  if (target.id() != exprt::index || target.operands().size() != 2)
+    return nullptr;
+
+  if (
+    !mentions_symbol(target.op1(), bound_vars) ||
+    !target.op0().type().is_array())
```
lucasccordeiro, line 144: An `int r[N]` parameter decays to a pointer, so `__ESBMC_old(r[j])` still hits the old error. Handle it, or add a test pinning the limitation.

src/goto-programs/goto_sideeffects.cpp
```diff
@@ -87,6 +87,103 @@ static bool mentions_symbol(const exprt &e, const std::set<irep_idt> &ids)
...
+    bound_array_element(call->op1().operands()[0], bound_vars);
+  if (!target)
+    return;
+
+  // Snapshotting an array through a dereference is an rvalue array read, which
+  // the dereference layer refuses. Where the array is a struct member, take the
+  // struct — a legal rvalue — and re-apply the member to the snapshot.
+  const exprt base = target->op0();
+  const exprt index = target->op1();
+  const bool via_member = base.id() == exprt::member;
```
lucasccordeiro, line 170: Both tests snapshot `r->coeffs[j]`, so the `via_member == false` branch never runs. Add a case whose base is an array directly.

src/goto-programs/goto_sideeffects.cpp
```diff
@@ -87,6 +87,103 @@ static bool mentions_symbol(const exprt &e, const std::set<irep_idt> &ids)
...
+  if (
+    call->id() != "sideeffect" || call->statement() != "function_call" ||
+    call->operands().size() < 2 || !call->op0().is_symbol())
+    return nullptr;
+
+  const std::string callee = call->op0().identifier().as_string();
+  static const std::string old_raw = "__ESBMC_old_raw";
+  if (
+    callee.size() < old_raw.size() ||
+    callee.compare(callee.size() - old_raw.size(), old_raw.size(), old_raw) !=
```
lucasccordeiro: Suffix match also accepts a user symbol ending in `__ESBMC_old_raw`. Compare the base name, or the full identifier, instead.

LukeW1999:
All three addressed in `3df4c24026` and `4a1625ec41`.

### The suffix match

Right, and it was silent. A temporary print in `old_raw_call_under` says the real id is `c:@F@__ESBMC_old_raw`, so nothing needs the suffix — and `c:@F@my__ESBMC_old_raw` matched it too. With a user function of that name inside a `forall`:

```
before:  VERIFICATION FAILED
after:   ERROR: cannot model call to `c:@F@my__ESBMC_old_raw' on a quantified variable
```

The FAILED was a verdict on a clause the user never wrote: the lift had replaced their call with a snapshot of the array being indexed. Now it splits at the last `@` and compares exactly, the same way `is_contract_intrinsic` in `contracts.cpp` does, and `github_4219_old_raw_lookalike` pins it.

### The `via_member == false` branch

It does run — the print shows `via_member=0 base.id=symbol object.type=array` for a directly named array — but you are right that nothing in the suite reached it. `github_4219_old_in_forall_array_base` does: master says `cannot model call to __ESBMC_old_raw on a quantified variable`, this says SUCCESSFUL. Its `_fail` twin adds two instead of one so the verdict turns on the snapshot rather than on the test being unfalsifiable.

First attempt used `__ESBMC_assigns(g)` and failed on `assigns compliance: c:@g not in assigns clause`, which is the gap `assigns_global_arr_elem_knownbug` already pins. Dropped the clause; the lift is what the test is about.

### The `int r[N]` parameter

Confirmed, still the old error, and the lift never fires. Pinned as `github_4219_old_in_forall_array_param_knownbug` rather than handled: the parameter is a pointer, and a pointer with a symbolic extent has no whole object to snapshot, which is the same wall the doc comment describes.

Probing that turned up a third shape worth pinning. `int (*r)[N]` does name a whole array, so the lift fires and takes the non-member path, and then the snapshot reads the array through a dereference:

```
```

Master errors too, differently, so nothing regressed — but the branch is what makes this reachable, so it is `github_4219_old_in_forall_ptr_to_array_knownbug`. The member path avoids it by taking the struct and re-applying the member; a pointer to array has no enclosing object to take. Handling it means keeping the snapshot as a pointer and indexing that instead of dereferencing to an array rvalue, which is a bigger change than this PR.

### One thing that is not mine

`clause_call_nested` fails in my tree and passes in the #6962 tree, and it is neither branch. Reverting `goto_sideeffects.cpp` to master and rebuilding, it still fails. It reports `c:@F@inner` where the `.desc` wants `c:@F@outer`, and which one you get depends on the binary:

`clause_call_callee` returns the first element of `select_clause_calls`, which is a `std::set<goto_programt::const_targett>`. That is a set of `std::list` iterators ordered by `std::less`, i.e. by node address, so with two eligible calls the diagnostic names whichever the allocator happened to place first. Ordering by position in the body instead would make it deterministic. Happy to send that separately if you want it — it is unrelated to both of these.

PR #6980, opened 2026-08-13
[contracts] Expect the inner call in the nested-clause diagnostic

lucasccordeiro:
clause_call_nested has failed on master since #6944 landed: for the clause `outer(inner(x)) > 100` the diagnostic names `c:@F@inner` (the first call the walk reaches), but the test asserts `c:@F@outer`. Reproduced deterministically, five runs out of five.

This matches the test to the shipped behaviour so the suite goes green. If naming the outermost call is what #6944 intended, the implementation should change instead and this expectation goes back.

### DIFF
--- a/regression/function_contract/clause_call_nested/test.desc
+++ b/regression/function_contract/clause_call_nested/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --enforce-contract f --function f
-^ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@outer', whose result the clause cannot name outside the function body$
+^ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@inner', whose result the clause cannot name outside the function body$


LukeW1999:
The determinism here is per-binary, not per-program, so this expectation will flip back on someone else's build. Two of mine, same input, same command line, on master's `main.c`:

```
build A -> c:@F@outer
build B -> c:@F@inner
```

They differ only in unrelated contract patches elsewhere, neither touching this path — I checked by reverting one of them to master and rebuilding, and the answer stayed put. With this `.desc` build A now fails; before it, build B failed.

The mechanism: `clause_call_callee` returns the first element of `select_clause_calls`, which is a `std::set<goto_programt::const_targett>` (`contracts.cpp:349`), and the comparator for those is

```cpp
// goto_program.cpp:215
bool operator<(const goto_programt::const_targett i1,
               const goto_programt::const_targett i2)
{
  const goto_programt::instructiont &_i1 = *i1;
  const goto_programt::instructiont &_i2 = *i2;
  return &_i1 < &_i2;
}
```

`instructiont`s live in a `std::list`, so each is separately allocated and "first" means "lowest heap address". With two eligible calls the diagnostic names whichever the allocator happened to place lower, which is stable for a given binary and arbitrary across binaries. Neither `inner` nor `outer` is the shipped behaviour; the walk order you observed is your allocator's.

`#6944` reads as intending "the first call the walk reaches", which would be program order. Iterating the body and returning the first instruction that is in `selected`, instead of taking `*selected.begin()`, makes it that. Happy to send it — the `.desc` would go back to `outer` as your commit message anticipated.

Two other target sets have the same comparator — `contracts.cpp:1194` (`instructions_to_remove`) and `remove_unreachable.cpp:27` (`reachable`). Membership is all either one uses, so neither is wrong today, but the ordering they carry is not the one it looks like.

LukeW1999:
Opened #7012. The clean reproducer is one environment variable on an unmodified master build, no rebuild:

    esbmc --enforce-contract f --function f clause_call_nested/main.c            -> 'c:@F@inner'
    MALLOC_MMAP_THRESHOLD_=1024 esbmc ... same command ...                       -> 'c:@F@outer'

The PR adds `clause_call_nested_three`, which nests one level deeper and fails on master in a plain environment (`outermost` where the data dependency says `innermost`), so the regression is red-green rather than a determinism claim. Your expectation here stays as it is: program order gives `inner` on every layout.

PR #6982, opened 2026-08-13
[goto-programs] Havoc the target of an unmodelled interval assignment

lucasccordeiro:
`interval_domaint::assign()` dispatched only on `is_bv_type`/`is_floatbv_type`, neither of which covers bool, so a write to a bool left alive the fact a guard had recorded for it.

Interval analysis then pruned live branches: 42 incorrect-true results on SV-COMP ReachSafety-Sequentialized (score 33688 -> 32499), and unsound proofs under the default symex guard domain too. The bool fold added in #6931 only routed those benchmarks into a pre-existing defect; a cast-free repro fails without it.

### DIFF
--- a/regression/esbmc/interval_bool_assign/main.c
+++ b/regression/esbmc/interval_bool_assign/main.c
@@ -1,0 +1,12 @@
+int main()
+{
+  int x = nondet_int();
+  _Bool b;
+
+  if (x > 5)
+  {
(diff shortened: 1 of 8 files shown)


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/31735319286.

PR #6987, opened 2026-08-13
[symex] Ackermannise uninterpreted functions over pointer args

### LINKED ISSUE #6965: [symex] Uninterpreted functions with pointer arguments lose functional congruence
## Summary

`symex_uninterpreted_function` models a `__ESBMC_uninterpreted_*` / `__CPROVER_uninterpreted_*` call as a genuine uninterpreted function **only when every argument and the result is a scalar**. Any pointer or aggregate operand falls back to a fresh nondeterministic result, which drops functional congruence: two calls with identical arguments may return different values.

The fallback is sound (it only removes constraints), but it makes any harness that relies on an uninterpreted predicate over pointers unprovable. This is the root cause of three SV-COMP false alarms in the aws-c-common cluster.

## Reproducer

```c
extern void abort(void);
void reach_error(void) { __assert_fail("0", "uf_ptr.c", 5, "reach_error"); }
void __VERIFIER_assert(int c) { if (!c) { reach_error(); abort(); } }

_Bool __CPROVER_uninterpreted_equals(const void *const a, const void *const b);
unsigned long __CPROVER_uninterpreted_hasher(const void *const a);

int main(void) {
  int x, y;
  const void *a = &x, *b = &y;
  _Bool r1 = __CPROVER_uninterpreted_equals(a, b);
  _Bool r2 = __CPROVER_uninterpreted_equals(a, b);
  __VERIFIER_assert(r1 == r2);          /* congruence over pointer args */
  unsigned long h1 = __CPROVER_uninterpreted_hasher(a);
  unsigned long h2 = __CPROVER_uninterpreted_hasher(a);
  __VERIFIER_assert(h1 == h2);
  return 0;
}
```

```
esbmc uf_ptr.c --64 --incremental-bmc
```

**Actual:** `VERIFICATION FAILED` — the two calls disagree.
**Expected:** `VERIFICATION SUCCESSFUL` — an uninterpreted function is by definition a function of its arguments.

Control: the same program with `int` arguments (`_Bool __CPROVER_uninterpreted_equals_i(int, int)`) returns `VERIFICATION SUCCESSFUL`, so the argument sort is the sole trigger.

Reproduced with ESBMC 8.4.0 on master `58e1ed9963`.

## Mechanism

`src/goto-symex/symex_function.cpp:427-458`:

```cpp
bool uf_encodable = is_number_type(call.ret->type);
for (const expr2tc &argument : arguments)
  uf_encodable = uf_encodable && is_number_type(argument->type);
```

`const void *` is not a number type, so the call takes the `else` branch and is assigned an unconstrained `sideeffect2t::allockind::nondet`.

The fallback was introduced deliberately in #5382 (`f3daf74742`) because `mk_smt_uninterpreted_function` cannot encode the tuple sort a pointer lowers to and aborted the solver backend (#5369). Avoiding the abort was correct; losing congruence is the collateral damage.

## Impact on SV-COMP

Three false alarms in run [31365723691](https://github.com/esbmc/esbmc/actions/runs/31365723691/job/93514135132) (ESBMC 8.4.0, 300s), all at k=1:

- `c/aws-c-common/aws_hash_table_create_harness.i`
- `c/aws-c-common/aws_hash_table_find_harness.i`
- `c/aws-c-common/aws_hash_table_remove_harness.i`

The find harness makes the inconsistency directly visible. `s_find_entry` matches a slot through `s_hash_keys_eq`, which calls `uninterpreted_equals(entry->element.key, key)`:

```
State 97  ... function uninterpreted_equals    rval = 1
State 98  ... function s_hash_keys_eq          rval = 1
```

The harness then asserts `p_elem->key == key || uninterpreted_equals(p_elem->key, key)`, and the same call on the same key pair returns the opposite:

```
State 122 ... function uninterpreted_equals    rval = 0
State 186 ... function __VERIFIER_assert       Violated property: reach_error
```

The harness's own `assume`s constrain only reflexivity and symmetry, so nothing else ties the two calls together.

## Suggested fix

Ackermannise non-scalar uninterpreted functions at symex level instead of dropping congruence: keep a per-(mangled identifier) history of `(arguments, result)` pairs and, for each new call, emit `arguments == previous_arguments => result == previous_result` using structural equality over the pointer/aggregate operands. This keeps the SMT backend free of tuple-sorted UF symbols — so the #5369 abort stays fixed — while restoring the defining property. A prototype along these lines was explored in PR #5364.

Note for whoever picks this up: restoring congruence is necessary but may not be sufficient to flip the three harnesses. The #5287 investigation found a second, stacked layer — reads of `p_elem->key` through an unresolvable pointer mint inconsistent pointer identity (#5369) — so the arguments themselves may still diverge between the two calls.

lucasccordeiro:
A pointer argument lowers to a tuple sort the SMT backend cannot use as an uninterpreted-function domain, so such calls fell back to a fresh nondet result and lost functional congruence entirely: two applications with identical arguments could disagree.

Keeps the fallback, but assumes equal arguments imply an equal result against each earlier application of the same callee, which restores the defining property without handing the solver a tuple-sorted symbol. Aggregates still fall through uncongruent, and the single-application soundness test stays refutable.

Fixes #6965.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/31831169327.

lucasccordeiro:
```
  correct:           20320
    correct true:    13470
    correct false:    6850
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16282
  Score:             33774 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/31831169327/job/94932450127
```

Issue #5393, opened 2026-06-17
[SV-COMP] false alarm on unreach-call: aws_hash_table_init_bounded_harness


lucasccordeiro:
## Wrong result in SV-COMP 26 run

**False alarm**: ESBMC reports `VERIFICATION FAILED` (FALSE) on a benchmark whose ground truth is TRUE — a spurious counterexample on a correct program (precision/incompleteness bug).

| Field | Value |
|---|---|
| Property | `unreach-call` |
| Expected verdict | `true` |
| ESBMC verdict | `false(unreach-call) (Bug found, k=129)` |
| Benchmark | [`c/aws-c-common/aws_hash_table_init_bounded_harness.i`](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_init_bounded_harness.i) |
| Run | [SV-COMP 26 benchexec run](https://github.com/esbmc/esbmc/actions/runs/27476688694) (commit `ee2c67bf32a3f0230788430c9d8c922e1ac0eefd`) |
| Reproduced on master | `74da7c0400` |
| ESBMC version | 8.3.0 |
| Strategy | `kinduction` |
| Related umbrella issue(s) | #2513, #1447 |

## Reproducer

Wrapper invocation (matches the SV-COMP run):

```
./esbmc-wrapper.py -p sv-benchmarks/c/properties/unreach-call.prp -s kinduction --arch 64 sv-benchmarks/c/aws-c-common/aws_hash_table_init_bounded_harness.i
```

Underlying ESBMC command:

```
./esbmc --no-div-by-zero-check --force-malloc-success --force-realloc-success --state-hashing --add-symex-value-sets --no-align-check --k-step 2 --floatbv --unlimited-k-steps --no-vla-size-check sv-benchmarks/c/aws-c-common/aws_hash_table_init_bounded_harness.i --64 --witness-output witness --enable-unreachability-intrinsic --no-pointer-check --interval-analysis --no-bounds-check --error-label ERROR --goto-unwind --unlimited-goto-unwind --k-induction --max-inductive-step 3
```

Fetch the benchmark from the [sv-benchmarks repo](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/aws-c-common/aws_hash_table_init_bounded_harness.i).

## Spurious counterexample

```
Violated property:
[...]

PR #6999, opened 2026-08-14
[solvers] Keep a non-null pointer off address zero

### LINKED ISSUE #7008: Nondet pointers are lost when stored into an untyped byte allocation
## Summary

A pointer with no value-set provenance — a nondet pointer — does not survive a
round trip through an **untyped (byte-array) heap object**. The store narrows to
a single byte, so an assumed non-null pointer reads back possibly null.

This is a **soundness bug**: it produces false alarms on correct programs. It is
the root cause of #5393 and #5394.

Reproduces on master (`010b25e557`) with default flags, no SV-COMP options.

## Reproducer

```c
#include <stdlib.h>
void *nondet_voidp(void);

int main(void)
{
  void *q = nondet_voidp();
  __ESBMC_assume(q != 0);

  char *raw = malloc(64);          /* byte-typed allocation */
  if (raw == 0)
    return 0;

  void **s = (void **)raw;
  *s = q;

  __ESBMC_assert(*s != 0, "a stored non-null pointer reads back non-null");
  return 0;
}
```

```
$ esbmc redP.c --no-pointer-check
VERIFICATION FAILED
```

The counterexample shows the store narrowing to one byte:

```
*s = (signed char)((unsigned long int)q)
```

## Boundary — what does and does not reproduce

All with default flags. Every row was run; the three passing shapes are the
sibling tests added in the PR below.

| Shape | Verdict |
|---|---|
| `malloc(64)` as `char *`, cast to `void **`, nondet pointer stored | **FAILED** (wrong) |
| `malloc(n)` where `n` is nondet, same store | **FAILED** (wrong) |
| `calloc(1, sizeof(void *))`, same store | **FAILED** (wrong) |
| `malloc(sizeof(void *))` — typed object | SUCCESSFUL |
| same byte allocation, **concrete** pointer (`&x`) | SUCCESSFUL |
| same byte allocation, `long` instead of a pointer | SUCCESSFUL |
| typed `malloc` + explicit `memset`, then store | SUCCESSFUL |

So it is specifically **pointers with no provenance into an untyped object**.
Integers survive because their byte stitching is exact; a pointer is an
object+offset pair rather than a bitvector. Concrete pointers survive via the
value set.

`calloc` is affected because its model
(`src/c2goto/library/stdlib.c:85`) is `malloc(total_size)` + `memset`, and
`total_size` is a runtime value — so the object never gets a type. `memset`
itself is innocent.

## Why it matters beyond these two issues

"Allocate bytes, cast, store a pointer" is the custom-allocator idiom. In
#5393/#5394 it is `aws_hash_table_init` copying nondet `hash_fn`/`equals_fn`
into a `calloc`'d `hash_table_state` via `*state = *template`
(`s_alloc_state`). The copy loses `equals_fn`, `hash_table_state_is_valid`
sees it NULL, and the harness's `__VERIFIER_assert` fires — the k=129 false
alarm, reproduced from the unmodified benchmark in 63 s.

## Ruled out

Both were rebuilt and re-run against the reproducers, not judged by inspection:

- **#6965 / PR #6987** (Ackermannised uninterpreted functions over pointer
  arguments) — same false alarm at k=129. Not the cause, despite both touching
  aws-c-common.
- **#6922 / PR #6977** (whole float element store without byte stitching) —
  minimal reproducers still FAILED. Same family, does not cover pointers.

## Where the fix probably belongs

`dereferencet::construct_from_array` (`src/pointer-analysis/dereference.cpp`,
~1355-1435). For a byte-array object `arr_subtype` is `signed char`, so an
8-byte pointer access has `deref_size > subtype_size` and takes the stitching
path, which cannot represent a pointer. The `whole_element` escape added for
#6922 is `deref_size == subtype_size && is_floatbv_type(arr_subtype)` — keyed
off the *array's* subtype, so it never fires here; the neighbouring #4435 case
is keyed the same way. A fix needs either a typed view of the object for
pointer-sized aligned accesses, or the pointer recorded in the value set on the
byte path (which #4435's comment notes the byte path destroys).

## Tests

PR #6999 pins this as `regression/esbmc/github_5393_nondet_ptr_byte_alloc`
(KNOWNBUG) with three mutation-checked CORE siblings covering the boundary
rows above.

## Version

```
ESBMC version 8.4.0 64-bit aarch64 macos, Bitwuzla 0.9.1
```

lucasccordeiro:
Object 0 is NULL and occupies the single address 0, and `finalize_pointer_chain` makes every other live object disjoint from it — so only NULL has address zero. Those constraints exist only for objects symex registered, so a pointer whose object id is a free variable escaped the invariant and the solver could place it at address 0. A `p != NULL` assumption was then lost the moment the pointer went through its numeric address, which is what any store into an untyped byte object does.

Minimal case, seven lines, default flags:

```c
void *q = nondet_voidp();
__ESBMC_assume(q != 0);
unsigned long a = (unsigned long)q;
__ESBMC_assert(a != 0, "");   /* VERIFICATION FAILED before this patch */
```

Four regression tests: the store-into-an-untyped-allocation case plus three mutation-checked boundary siblings (concrete pointer, integer, typed allocation) that passed before and still do.

Found while reducing #5393. It fixes the first of that harness's two failing conjuncts — `equals_fn` now reads back non-null — but a second, independent defect there keeps the benchmark failing, so #5393/#5394 stay open.

Fixes #7008.


lucasccordeiro:
This PR (30s): 

```
  correct:           20389
    correct true:    13521
    correct false:    6868
  incorrect:            45
    incorrect true:     42
    incorrect false:     3
  unknown:           16169
  Score:             32518 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/31831540343.
```

rafaelsamenezes:
The 45 incorrect results in the run above are not from this patch — they come from the branch's base commit, which predates the fix for a known unsoundness.

The 45 split into 42 `incorrect true` + 3 `incorrect false`. Dedup'd from the run artifact, all 42 wrong `true`s are one family: `c/seq-mthreaded-reduced/pals_lcr*` (ReachSafety-Sequentialized). The 3 wrong `false`s (`aws-c-common/aws_hash_table_{find,remove}_harness`, one ldv termination task) appear in every run before and after, so they are independent of this PR.

Three runs, same 30 s configuration:

| run | commit | has #6931 / #6982 | incorrect (true/false) | score |
|---|---|---|---|---|
| master 2026-08-13 | `fbabb08169` | yes / **no** | 45 (**42**/3) | 32499 |
| this PR | `21fed8ae79` | yes / **no** | 45 (**42**/3) | 32518 |
| master 2026-08-14 | `f2df960d08` | yes / **yes** | 3 (0/3) | 33822 |

Pre-fix master reproduces this PR's signature exactly — same 42, same category, same score band — with none of this patch in it. This branch is based on `010b25e557`, which contains #6931 but not #6982, i.e. it sits inside the regressed window.

The cause is the one 00eb9fcf6a (#6982) fixed: `interval_domaint::assign()` dispatched only on `is_bv_type`/`is_floatbv_type`, neither of which covers bool, so a write to a `_Bool` left alive a fact a guard had recorded, and interval analysis pruned live branches. That commit quotes the same number — "42 incorrect-true results on SV-COMP ReachSafety-Sequentialized (score 33688 -> 32499)". It fits the tasks: the wrapper passes `--interval-analysis` for non-concurrent unreach-call, and the `pals_lcr` programs are scalar/`_Bool` reactive code. It also explains the timing: those tasks were `unknown` (timeout) before and answer a wrong `true` in 1–29 s, because the pruning is cheap. The score arithmetic agrees — 42 × (−32) + 40 from the extra correct answers = the −1304 against post-fix master.

Rebasing onto master and re-running should bring `incorrect` down to 3.

One thing worth watching on the clean re-run, unrelated to the above: the new axiom `obj != 0 -> addr_space[obj].start + offset != 0` is asserted unconditionally. The premise holds for base addresses (object 1 starts at 1, and disjointness from object 0 forces `start >= 1` for the rest), but the axiom also excludes models where an out-of-bounds offset wraps `start + offset` to 0 mod 2^64, which is reachable under `--no-pointer-check` where offsets are unconstrained. That prunes counterexamples rather than adding them, so a post-rebase `incorrect true` count is the check for it.

lucasccordeiro:
New SV-COMP run (30s): 

```
  correct:           20310
    correct true:    13459
    correct false:    6851
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16292
  Score:             33753 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/31899726535
```

PR #7000, opened 2026-08-14
[contracts] Accept either call in the nested-clause diagnostic

lucasccordeiro:
clause_call_nested pins which call the walk names for `outer(inner(x))`, and that is not stable. The same ubuntu leg produced `inner` on one run (#6970) and `outer` on another (#6982) with src/goto-programs/contracts unchanged since #6944, so either name is flaky.

Accept both until the traversal is made deterministic. My earlier #6980 swapped the expectation from `outer` to `inner` on a single local measurement; that was treating a flake as a fixed value, and this supersedes it.

Refs #6944

### DIFF
--- a/regression/function_contract/clause_call_nested/test.desc
+++ b/regression/function_contract/clause_call_nested/test.desc
@@ -1,4 +1,4 @@
 CORE
 main.c
 --enforce-contract f --function f
-^ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@inner', whose result the clause cannot name outside the function body$
+^ERROR: --enforce-contract: cannot use 'f': a contract clause calls 'c:@F@(inner|outer)', whose result the clause cannot name outside the function body$


LukeW1999:
#7012 makes it deterministic, so this can tighten back up.

The traversal follows the heap, not the program: `clause_call_callee` takes the first element of a `std::set<goto_programt::const_targett>`, and `operator<` on those is `&*i1 < &*i2`. `instructiont` lives in a `std::list`, so "first" is "lowest address". Your #6970/#6982 disagreement on one ubuntu leg is that, and it reproduces on demand without a rebuild:

    esbmc --enforce-contract f --function f clause_call_nested/main.c   -> 'c:@F@inner'
    MALLOC_MMAP_THRESHOLD_=1024 esbmc ... same command ...              -> 'c:@F@outer'

#7012 walks the body instead, which is the order #6944 describes, and puts the expectation back to `inner`. It also adds a three-level case that names `outermost` on master in a plain environment where the data dependency says `innermost`, so there is a red-green test rather than only a determinism argument.

PR #7015, opened 2026-08-14
[solver] Lower remainder/fmod/remquo to the solver's fp.rem

### LINKED ISSUE #6896: libm models of fmod/remainder/remquo are unsound: x - y*(int)(x/y)
## Summary

The operational models of `fmod`, `remainder`, and `remquo` compute the finite case as

```c
return x - (y * (int)(x / y));   /* src/c2goto/library/libm/fmod.c:31 */
```

(`remquo` uses `llrint` for the quotient). This is wrong on ordinary inputs, for two independent reasons:

1. **Double rounding**: `x / y` rounds to the nearest double, so `y * (int)(x/y)` is not the exact multiple of `y` nearest/below `x`; the subtraction then measures the distance to the *wrong* multiple.
2. **Truncation/overflow through the `int` cast**: the quotient is truncated (wrong rounding for `remainder`, which needs nearest-even) and overflows `int` entirely for large `x/y`, after which the result is unrelated to the true remainder.

`remquo` additionally never writes `*quo`.

## Concrete counterexample (CPython-verified)

```
fmod(-524285.96875476, 1.9995117783546448) = +5.8207660913467407e-11   (model)
                                             -1.9995117783546448...    (true fmod ~ -1.99951)
```

Wrong sign *and* wrong magnitude.

## Property-level evidence

All four C17 7.12.10 properties are **disproved by the solver on every backend** against the current models, e.g.:

```c
double x = nondet(); double y = nondet();
__ESBMC_assume(isgreaterequal(x, -1e6) && islessequal(x, 1e6));
__ESBMC_assume(isgreaterequal(y, 1.0) && islessequal(y, 1024.0));
assert(isless(fabs(fmod(x, y)), fabs(y)));            /* |fmod| < |y|: FAILS */
```

and likewise `fmod` sign-of-`x`, `|remainder| <= |y|/2`, and the same bound via `remquo`.

## Fix (implemented on the camada branch, PR #4179)

Commit `5f20a6d2d0` replaces the composition with the solver's exact operation:

- New `ieee_rem2t` IR node, lowered to SMT `fp.rem` on the bit-vector path (`fp.rem` is exact — no rounding-mode dependence) and, under `--ir-ieee`, to `r = x - n*y` with a fresh nearest integer `n` (even on ties).
- `remainder()` maps to `ieee_rem` in the frontend (guarded on the float type so a user-defined `int remainder(int,int)` is untouched).
- `fmod` becomes `remainder` plus an exact sign fixup: when the signs disagree the two differ by exactly one `|y|`, both representable (C17 7.12.10).
- `remquo` computes the quotient bits from the same construction and actually writes `*quo`.
- Constant folds for the trivial cases (rem of/by 0, 1, NaN, inf operands).
- Five regression tests pin the properties: `regression/floats-regression/ieee_rem_{fmod_bound,fmod_sign,fmod_sign_fail,remainder_bound,remquo_bound}` — all four properties flip from FAILED to SUCCESSFUL, and the `_fail` twin keeps the checker honest.

As a side effect the encoding is also cheap: on the property proofs above, camada's fp.rem lowering measured 30-100x faster than bitwuzla's native fp.rem operator (the old model chain was never timed like-for-like -- the properties are simply disproved against it).

The fix rides the camada solver-backend PR; if that lands on its current schedule this issue is closed by it. Filing separately so the soundness bug is tracked on its own — it affects every current master verification touching `fmod`/`remainder`/`remquo`, and could be cherry-picked to master ahead of the backend switch if needed (the `ieee_rem2t` lowering is backend-agnostic; master's fp API has `mk_smt_fpbv_rem` equivalents).

lucasccordeiro:
The libm models computed the finite case as `x - y*(int)(x/y)`, which
double-rounds through an inexact quotient and truncates through an int cast
that overflows for large quotients: `fmod(1e18, 3.0)` returned 64.0 instead
of 1.0. Adds an `ieee_rem2t` node lowered to native `fp.rem` on z3/cvc4/cvc5,
to an exact bit-blasted remainder in `fp_convt` for the solvers without one
(bitwuzla, the default), and to `x - n*y` with a fresh nearest-even `n` under
`--ir-ieee`; `fmod` becomes `remainder` plus an exact sign fixup.

Ported from the camada branch (#4179). The symbolic property tests live in
`regression/bitwuzla/` because the same query solves in ~2s there and takes
over 200s under z3.

Fixes #6896


mikhailramalho (approved):
lgtm

XLiZHI (approved):
LGTM

PR #7018, opened 2026-08-14
[python] Extend dynamic typing to variable reassignment and arithmetic

brcfarias:
Follow-up to #6535. Extends dynamic-typing support to a variable that already had a native type before the if/else that reassigns it incompatibly (previously only variables created inside branches were covered), and adds +, -, *, / support against a literal (numeric, plus string concatenation for +). Also extracts the tagged-scalar conversion logic into a dedicated dynamic_type_handler class.


src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -0,0 +1,485 @@
...
+        new_id,
+        alias_location,
+        type_handler_.get_tagged_object_type());
+      alias_symbol.lvalue = true;
+      alias_symbol.file_local = !converter_.get_current_func_name().empty();
+      alias_symbol.is_extern = false;
+
+      symbolt *alias_symbol_ptr =
+        converter_.symbol_table().move_symbol_to_context(alias_symbol);
+      aliases_[tag_id] = new_id;
```
lucasccordeiro, line 165: The alias is never initialised from the pre-existing symbol, so a path that skips both branches (nested `if`, zero-trip loop) reads an uninitialised tagged object. Initialise it at declaration.

src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+bool dynamic_type_handler::is_tagged(const std::string &name) const
+{
+  if (tagged_names_.count(name) > 0)
+    return true;
+
+  symbol_id sid = converter_.create_symbol_id();
+  sid.set_object(name);
+  const std::string tag_id = sid.to_string();
+  if (aliases_.count(tag_id) > 0)
```
lucasccordeiro, line 219: `aliases_` is never cleared, so after the join `x = f()` now throws "assigning a non-literal value ... not yet supported" for pre-existing variables that converted fine before.

src/c2goto/library/python/scalar.c
```diff
@@ -62,3 +70,103 @@ __ESBMC_HIDE:;
...
+    tagged && tagged->type_id == type_id,
+    "TypeError: unsupported operand type(s) for /");
+  if (!tagged || tagged->type_id != type_id)
+    return 0;
+  long long tagged_val = *(const long long *)tagged->value;
+  long long divisor = tagged_is_left ? value : tagged_val;
+  __ESBMC_assert(divisor != 0, "ZeroDivisionError: division by zero");
+  if (divisor == 0)
+    return 0;
+  return tagged_is_left ? tagged_val / value : value / tagged_val;
```
lucasccordeiro: Python `/` is true division; the non-tagged path returns `double_type()` (converter_binop.cpp:2630). Truncating `long long` division makes `x = 7; x / 2` yield 3, not 3.5.

src/c2goto/library/python/scalar.c
```diff
@@ -62,3 +70,103 @@ __ESBMC_HIDE:;
...
+{
+__ESBMC_HIDE:;
+  __ESBMC_assert(
+    tagged && tagged->type_id == type_id,
+    "TypeError: unsupported operand type(s) for /");
+  if (!tagged || tagged->type_id != type_id)
+    return 0;
+  long long tagged_val = *(const long long *)tagged->value;
+  long long divisor = tagged_is_left ? value : tagged_val;
+  __ESBMC_assert(divisor != 0, "ZeroDivisionError: division by zero");
```
lucasccordeiro: ZeroDivisionError must be a catchable raise, as converter_binop.cpp:1605-1651 does for the non-tagged path; a bare assert makes `try: x / 0 except ZeroDivisionError` report FAILED.

src/c2goto/library/python/scalar.c
```diff
@@ -62,3 +70,103 @@ __ESBMC_HIDE:;
...
+
+long long __python_scalar_mult_num(
+  const PyObject *tagged,
+  size_t type_id,
+  long long value)
+{
+__ESBMC_HIDE:;
+  __ESBMC_assert(
+    tagged && tagged->type_id == type_id,
+    "TypeError: unsupported operand type(s) for *");
```
lucasccordeiro: Python allows `str * int` repetition, so asserting TypeError when the tagged value holds a str is a false positive. Model the repeat, or refuse at conversion time.

src/c2goto/library/python/scalar.c
```diff
@@ -62,3 +70,103 @@ __ESBMC_HIDE:;
...
+  const char *value,
+  size_t value_size,
+  int tagged_is_left)
+{
+__ESBMC_HIDE:;
+  __ESBMC_assert(
+    tagged && tagged->type_id == type_id,
+    "TypeError: can only concatenate str (not other types) to str");
+
+  size_t tagged_len = tagged->size - 1;
```
lucasccordeiro, line 105: Unlike the four sibling functions, no `if (!tagged || tagged->type_id != type_id) return` guard follows the assert, so the mismatch path reads a numeric payload as a string.

regression/python/dynamic_type_scalar_tag_sub_num/main.py
```diff
@@ -0,0 +1,7 @@
+cond = nondet_bool()
+if cond:
+    x = 5
+    y = x - 2
```
lucasccordeiro, line 4: Every new arithmetic test puts the tagged operand on the left, leaving the `tagged_is_left == 0` paths untested. Add cases like `2 - x`, `10 / x` and `"b" + x`.

lucasccordeiro (changes requested):

PR #7020, opened 2026-08-14
add  suport numpy arange arrays

sergillam:
- `np.arange(...)` with constant, small arguments — materialized to a literal list directly instead of routing through the operational model's while-loop implementation (`models/numpy.py`'s `arange()`), which was disproportionately expensive to symbolically execute even for a handful of elements.
- `np.arange(n)` with a non-constant argument (e.g. sourced from a function parameter) — rejected explicitly and quickly with a `TypeError` instead of hanging past any practical timeout.
- `np.arange(start, stop, step)` — start/stop/step variants (positive/negative step, empty result, `step=0` → explicit `ValueError`) and float variants, preserved from existing logic and covered by new regression tests.
- `np.arange(...)` used directly as an argument (not via a variable) to `np.sum`/`np.mean`/`np.max`/`np.flatten`/`np.ravel`/`np.greater`/`np.equal` — now resolved correctly instead of failing extraction.


src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5077,6 +5120,48 @@ exprt numpy_call_expr::create_expr_from_call()
...
+  // list-concatenation implementation (models/numpy.py's arange()), which
+  // is disproportionately expensive to symbolically execute even for a
+  // handful of elements. Although real np.arange() returns an ndarray, this
+  // frontend materializes it as a list-like runtime object for consistency
+  // with build_static_lists, disabled the same way full()/eye()/identity()/
+  // linspace() already do it -- a plain static array here would not
+  // compare equal to a `[]`-literal PyListObj.
+  if (
+    std::optional<nlohmann::json> materialized =
+      materialize_numpy_constructor_array(call_, converter_.ast()))
```
lucasccordeiro: `materialize_arange` accepts only literal nodes, so `n = 3; np.arange(n)` now throws; master verifies it SUCCESSFUL in 2.6s. Run `resolve_var` on each argument first.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5077,6 +5120,48 @@ exprt numpy_call_expr::create_expr_from_call()
...
+    call_.contains("args") &&
+    arange_constant_args_have_zero_step(call_["args"]))
+    throw std::runtime_error(
+      "ValueError: numpy.arange() step must not be zero");
+
+  // Non-constant arguments (e.g. a function parameter) cannot be
+  // materialized this way; falling back to the operational model here
+  // hangs past any practical timeout instead of producing a verdict, so
+  // this is rejected explicitly and quickly instead.
+  throw std::runtime_error(
```
lucasccordeiro, line 5173: `is_recognized_numpy_constructor_call_shape` also declines `from numpy import arange` and any keyword, so `arange(3)` and `np.arange(3, dtype=np.int32)` now throw; both verify SUCCESSFUL on master.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5077,6 +5120,48 @@ exprt numpy_call_expr::create_expr_from_call()
   return converter_.get_expr(expr);
 }
 
+exprt numpy_call_expr::get_arange_expr()
+{
+  // np.arange(...) with constant, small arguments is materialized to a
```
lucasccordeiro, line 5126: Nothing bounds the materialized element count despite "small arguments": `np.arange(1000000)` builds a million-element JSON list here. Cap the count and decline above it.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5077,6 +5120,48 @@ exprt numpy_call_expr::create_expr_from_call()
...
+    {
+      converter_.current_lhs->type() = result.type();
+      converter_.update_symbol(*converter_.current_lhs);
+    }
+    return result;
+  }
+
+  if (
+    call_.contains("args") &&
+    arange_constant_args_have_zero_step(call_["args"]))
```
lucasccordeiro: `np.arange()` and `np.arange(1,2,3,4)` fall through to the generic TypeError below. Keep the "expects 1 to 3 arguments" arity message for those.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5077,6 +5120,48 @@ exprt numpy_call_expr::create_expr_from_call()
   return converter_.get_expr(expr);
 }
 
+exprt numpy_call_expr::get_arange_expr()
```
lucasccordeiro, line 5113: This duplicates the arange block at lines 3159-3191, which is unreachable because `get()` returns here first. Delete the dead copy instead of maintaining two.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -2102,6 +2102,30 @@ materialize_arange(const nlohmann::json &args)
     to_double(start_v), to_double(stop_v), to_double(step_v));
 }
 
+// True when args are all-constant arange() arguments whose step evaluates to
+// exactly zero. Checked separately from materialize_arange() so a zero step
+// can raise NumPy's actual ValueError instead of the generic "non-constant
+// input" TypeError every other decline reason gets.
+static bool arange_constant_args_have_zero_step(const nlohmann::json &args)
```
lucasccordeiro: Re-parses arguments `materialize_arange` already parsed and duplicates its `step == 0` check. Have `materialize_arange` report the decline reason instead.

regression/numpy/arange_float_step_edge/main.py
```diff
@@ -0,0 +1,14 @@
...
+# Negative float step.
+a = np.arange(2.0, 0.0, -0.5)
+assert len(a) == 4
+assert a[0] == 2.0
+assert a[1] == 1.5
+assert a[2] == 1.0
+assert a[3] == 0.5
+
+# stop is exclusive even when the progression would land exactly on it.
+b = np.arange(0.0, 2.0, 0.5)
```
lucasccordeiro, line 12: Both float steps are exact binary fractions. `materialize_arange_float` accumulates `current += step`, so add `np.arange(0.0, 1.0, 0.1)`: it yields 11 elements, NumPy 10.

lucasccordeiro (changes requested):

Issue #7021, opened 2026-08-14
signbit of a NaN is unconstrained under native FP backends (--z3)


lucasccordeiro:
`regression/floats-regression/copysign` passes on the default solver
(bitwuzla) and **fails under `--z3`**. Minimal reproducer:

```c
#include <assert.h>
#include <math.h>
int main(void)
{
  double s = copysign(NAN, -2.0);
  assert(isnan(s));
  assert(signbit(s));   // <-- violated under --z3
  return 0;
}
```

```
$ esbmc a.c --z3             VERIFICATION FAILED   (assertion signbit(s))
$ esbmc a.c --z3 --fp2bv     VERIFICATION SUCCESSFUL
$ esbmc a.c                  VERIFICATION SUCCESSFUL   (bitwuzla)
```

`VERIFICATION SUCCESSFUL` is the correct answer. `signbit` "determines the
sign of all values, including infinities, zeros, and NaNs" (N3220 §7.12.3.7
fn. 273), and `copysign` "produce[s] a value with the magnitude of x and the
sign of y" (N3220 §7.12.11.1) with no NaN exception. Apple libm agrees: the
same program compiled natively prints `isnan=1 signbit=1`.

### Cause

`convert_signbit` (`src/solvers/smt/smt_fp_conv.cpp:841`) extracts the top
bit of the FP value. With a native FP backend that operand has an FP sort,
so `z3_convt::mk_extract` routes it through `mk_from_fp_to_bv`
(`src/solvers/z3/z3_conv.cpp:791`), i.e. `Z3_mk_fpa_to_ieee_bv`. SMT-LIB's
`fp.to_ieee_bv` is **unspecified for NaN** — the theory has a single,
sign-less NaN — so the sign bit is left free and the solver picks 0.

The software encoding builds a concrete NaN bit-pattern, so `fp.neg` flips a
real sign bit and the property holds. That is why `--fp2bv`, and every solver
without native FP (bitwuzla, boolector, yices, smtlib), are correct.

### Scope

Any bit-level observation of a possibly-NaN FP value under a native FP
backend — `signbit`, and union / `memcpy` type punning of a NaN.

cvc4/cvc5 already half-address this: their own `mk_from_fp_to_bv` pins a
single `__ESBMC_NaN` symbol so that all NaN-to-bv conversions at least agree
with one another. The bit-pattern is still unconstrained, so the sign of a
NaN should be free there too — not verified, as those backends are not built
[...]

PR #7022, opened 2026-08-14
[solver] Use Bitwuzla's native floating-point theory by default

lucasccordeiro:
Implements the `fp_convt` interface on `bitwuzla_convt` — FP and rounding-mode
sorts, constants, arithmetic, comparisons, classification, conversions and
model extraction — and makes it the default, so floats are encoded with
`fp.add` and friends instead of ESBMC's own bit-vector encoding. `--fp2bv`
opts back out. Bitwuzla has no `fp.to_ieee_bv`, so the bit pattern is reached
through a fresh bit-vector constrained by the bv->fp direction, as cvc4/cvc5 do.

`fp.rem` keeps ESBMC's lowering via a bit-vector round-trip, as the mathsat
backend does: `ieee_rem_remainder_bound` solves in 0.81s that way against 92s
with `fp.rem`. Measured on macOS/arm64, the theory is 1.2-13x faster on
symbolic-input FP queries, and 303 FP-touching CORE tests agree on every
verdict.

The theory cannot represent the sign of a NaN (#7021), so that assertion moves
from `floats-regression/copysign` to `regression/bitwuzla/fp2bv_nan_sign`
(CORE) and `native_fp_nan_sign` (KNOWNBUG).

Stacked on #7015 (base is that branch, not master) for `mk_smt_fpbv_rem`.


mikhailramalho (commented):
I suggest using bitwuzla native encoding by default. While fp2bv rem's from camada is better, other operations as faster with native fp bitwuzla encoding. And fp.rem is a very rare operation to force esbmc to use fp2bv encoding by default.

lucasccordeiro:
Done — native is now the default, `--fp2bv` opts back out, and `--bitwuzla-native-fp` is gone.

You were right that the original table was measuring the wrong thing: those benchmarks are all `fp.rem`-bound. On symbolic-input FP queries the theory wins comfortably (macOS/arm64, DebugOpt, bundled Bitwuzla):

| workload | `--fp2bv` | native |
|---|---|---|
| 12-term series over a nondet double | 4.68s | 0.34s |
| Newton sqrt over a nondet double | 1.29s | 0.22s |
| nondet `sqrt` / `signbit` / conversions | 0.16-0.49s | 0.14-0.31s |

On `fp.rem` we no longer have to choose: `mk_smt_fpbv_rem` now round-trips to bit-vectors and uses our own lowering, the way the mathsat backend does. `ieee_rem_remainder_bound` goes 92.1s (fp.rem) -> 0.81s, i.e. below the 1.76s of the old default; `ieee_rem_remquo_bound` 79.3s -> 0.97s.

Across 303 FP-touching CORE tests in floats, floats-regression, bitwuzla, esbmc, cstd, cbmc, esbmc-cpp*, k-induction, cuda, llvm and smoke-tests there are no verdict differences; total runtime 97.8s native vs 71.2s bit-vector, the gap concentrated in five constant-folding-heavy Taylor-series tests (`math_exp02` 1.7s -> 8.5s, `Float28` 0.4s -> 6.1s), all far inside the CI cap.

Two things worth flagging:

* The sign of a NaN (#7021) is now wrong on the default path, since SMT-LIB's FP theory has a single sign-less NaN. `floats-regression/copysign` keeps its finite and infinity cases; the NaN assertion moved to `regression/bitwuzla/fp2bv_nan_sign` (CORE, correct under `--fp2bv`) and `native_fp_nan_sign` (KNOWNBUG).
* No test used `--fp2bv` before, so the flip would have left our software FP encoding with no regression coverage at all. Added `fp2bv_arith`, `fp2bv_rem` and `fp2bv_nan_sign` to keep it exercised.

lucasccordeiro:
This PR (30s): 

```
  correct:           20899
    correct true:    14034
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15703

```

Master (30s):

```
  correct:           20449
    correct true:    13614
    correct false:    6835
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16153
  Score:             34047 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32578289411
```

lucasccordeiro:
This PR (100s): https://github.com/esbmc/esbmc/actions/runs/32267826941.

PR #7033, opened 2026-08-15
[contracts] Name the value a replaced call returns

### LINKED ISSUE #7009: [contracts] --replace-call-with-contract assumes ensures about the caller's variable instead of a fresh return value
## Summary

When a replaced call appears as the right-hand side of an **assignment**, `--replace-call-with-contract` introduces no temporary for the return value. It assumes the `ensures` clause directly about the caller's own variable, which is never havocked. The clause therefore constrains whatever that variable held *before* the call.

When the pre-call value contradicts the contract this is `assume false`: the path dies silently and ESBMC reports **VERIFICATION SUCCESSFUL** for a program whose assertion is plainly reachable. There is no diagnostic — the log still prints `Replacing calls with contracts for N function(s)`.

When the call instead **initialises a declaration**, the same contract is handled correctly.

## Reproducer

```c
#include <assert.h>

int f(void)
{
	__ESBMC_ensures(__ESBMC_return_value == 1);
	return 1;
}

int main(void)
{
	int r = 200;    /* stale value the call must overwrite */
	r = f();
	assert(0);      /* reachable, so the only correct answer is FAILED */
	return 0;
}
```

```
$ esbmc repro.c --replace-call-with-contract "*"
VERIFICATION SUCCESSFUL      # wrong: assert(0) is reachable

$ esbmc repro.c              # whole-program, real body
VERIFICATION FAILED          # correct
```

## Where it goes wrong

`--goto-functions-only` on the reproducer:

```
DECL signed int r;
ASSIGN r=200;
ASSUME r == 1        // contract ensures -- about the stale pre-call value
ASSERT 0
```

No temporary, no havoc. Writing the call as an initialiser instead produces the correct shape:

```c
int r = f();
```

```
DECL signed int return_value$_f$1;
ASSUME return_value$_f$1 == 1     // contract ensures
ASSIGN r=return_value$_f$1;
```

A fresh symbol of the callee's return type, the clause assumed about it, then the assignment.

## Second symptom: a truncating cast

Because there is no correctly-typed temporary, the assumption is emitted as a `(signed int)` cast of the caller's lvalue. For any return type wider than `int` that truncates to the low 32 bits and reinterprets them as signed, so the assumed postcondition is far weaker than what was written and the caller's assertion fails spuriously:

```c
#include <stdint.h>
#include <assert.h>

uint64_t f(void)
{
	__ESBMC_ensures(__ESBMC_return_value < 100);
	return 1;
}

int main(void)
{
	uint64_t r;
	r = f();
	assert(r < 100);
	return 0;
}
```

```
$ esbmc repro2.c --replace-call-with-contract "*"
VERIFICATION FAILED          # spurious

$ # change to `uint64_t r = f();`
VERIFICATION SUCCESSFUL      # correct
```

```
ASSIGN r=NONDET(unsigned long int);
ASSUME (signed int)r < 100          // contract ensures -- truncated to 32 bits
```

`int` and `unsigned short` are unaffected. `unsigned int`, `long`, `int64_t`, `uint32_t` and `uint64_t` all reproduce.

## Notes

- `requires` is emitted correctly in both forms; only the `ensures` side is affected.
- Not #6364, which is closed and drops the whole replacement (requires included) for a call at a jump-target position. Here the replacement happens, and the postcondition is bound to the wrong object.

Reproduced on ESBMC 8.4.0 built from master at `66dd7970ac`, `DebugOpt` (assertions enabled), x86_64 linux.

LukeW1999:
Fixes #7009.

## Symptom

`__ESBMC_return_value` was rewritten to the caller's own lvalue, which
`--replace-call-with-contract` never wrote, so the `ensures` was assumed about
the value that lvalue held *before* the call.

```c
int f(void) { __ESBMC_ensures(__ESBMC_return_value == 1); return 1; }

int main(void)
{
	int r = 200;
	r = f();
	assert(0);       /* reachable, so the only correct answer is FAILED */
	return 0;
}
```

```
$ esbmc repro.c --replace-call-with-contract "*"
VERIFICATION SUCCESSFUL      # wrong, and silent
```

`r` is still 200, so `ASSUME r == 1` is `assume false`: the path dies and the
reachable assertion is never reported. A false negative — it does not turn a
passing table red, it makes a failing one look green.

Where the stale value happens to satisfy the contract the path survives and the
caller simply keeps it. Writing the same call as an initialiser was already
correct, so whether the contract was honoured depended on how the call site was
spelled.

## Fix

Havocking the caller's place in situ is not enough, and the first version of
this PR did exactly that. The clause is instantiated with the caller's
*argument expressions*, so `r = f(r)` carried `r` on both sides and the ASSUME
read `r == r + 1` — the same `assume false`, now for a different reason.

The result gets a name of its own, and the caller's place is written once the
clause has been assumed:

```
DECL signed int return_value$0;
ASSIGN return_value$0=NONDET(signed int);
ASSUME return_value$0 == r + 1
ASSIGN r=return_value$0;
```

A discarded result needs the same name. The ensures is assumed either way, and
with nothing to rewrite `__ESBMC_return_value` to it was assumed over a symbol
no instruction defines, which took a reachable assertion with it. That was a
second false negative sitting next to the first, also present on master.

`create_snapshot_variable` and the new declaration share
`declare_local_symbol` rather than repeating the symbol setup, and the
write-back is its own helper, so `generate_replacement_at_call` gains no
cyclomatic complexity.

## Known limitation, pinned rather than hidden

A pointer result carries no value set, so a contract pinning it to an object
proves the equality and then loses writes through it
(`github_7009_pointer_result_knownbug`). This is not a contracts defect: the
same failure reproduces with no contract flag at all, from a plain
`__ESBMC_assume(tmp == &a)` — filed as **#7066** against goto-symex.

Omitting the declaration for pointers is not an option: that puts the `ensures`
back over the caller's stale value, which is the false negative this PR
removes. A reported failure is the safe direction of the two, and unlike the
other it is visible.

## Tests

Six, in `regression/function_contract/`. Every red-green pair was checked by
building master first, not assumed.

| test | master | patched | |
|---|---|---|---|
| `github_7009_ensures_assigned_return_fail` | SUCCESSFUL | FAILED | red-green |
| `github_7009_ensures_stale_value_fail` | SUCCESSFUL | FAILED | red-green |
| `github_7009_self_assigned_arg_fail` | SUCCESSFUL | FAILED | red-green; `r = f(r)` |
| `github_7009_discarded_result_fail` | SUCCESSFUL | FAILED | red-green; result dropped |
| `github_7009_self_assigned_arg` | SUCCESSFUL | SUCCESSFUL | control: the result is written back, not merely left unconstrained |
| `github_7009_ensures_init_return` | SUCCESSFUL | SUCCESSFUL | control: the initialiser form that already worked, so an over-havoc is caught |
| `github_7009_pointer_result_knownbug` | — | KNOWNBUG | #7066 |

`function_contract` 380/380, `loop-invariants` 76/76, `python-contracts` 55/55
— the three suites whose `.desc` files pass `--replace-call-with-contract`.
Nothing outside those paths can reach the change; it lives in
`generate_replacement_at_call`.

### DIFF
--- a/src/goto-programs/contracts/contracts.cpp
+++ b/src/goto-programs/contracts/contracts.cpp
@@ -2578,35 +2578,37 @@ bool code_contractst::is_old_call(const expr2tc &expr) const
   return false;
 }
 
+// A fresh lvalue of \p type registered under \p name.
+//
+// Note: symbolt uses IRep1 (typet) while we work with IRep2 (type2tc). This is
+// ESBMC's architecture: the symbol table is IRep1-based for global state, while
+// modern code (GOTO programs, contracts) uses IRep2 for local logic. Set the
+// symbol's IREP2 type directly via the migrate-layer chokepoint;
+// set_symbol_type stores the cache authoritatively and derives the legacy field
+// via migrate_type_back exactly once (esbmc/esbmc#4715, B2 S4b).
+expr2tc code_contractst::declare_local_symbol(
+  const std::string &name,
+  const type2tc &type) const
+{
+  symbolt symbol;
+  symbol.name = name;
+  symbol.id = name;
+  set_symbol_type(symbol, type);
+  symbol.lvalue = true;
+  symbol.static_lifetime = false;
+  symbol.file_local = false;
+
+  return symbol2tc(type, context.move_symbol_to_context(symbol)->id);
+}
+
 expr2tc code_contractst::create_snapshot_variable(
   const expr2tc &expr,
   const std::string &func_name,
   size_t index) const
 {
-  // Generate unique snapshot variable name
-  std::string snapshot_name =
-    "__ESBMC_old_snapshot_" + func_name + "_" + std::to_string(index);
-
-  // Create symbol and add to symbol table
-  // Note: symbolt uses IRep1 (typet) while we work with IRep2 (type2tc).
-  // This is ESBMC's architecture: Symbol Table is IRep1-based for global state,
-  // while modern code (GOTO programs, contracts) uses IRep2 for local logic.
-  // Set the symbol's IREP2 type directly via the migrate-layer chokepoint;
-  // set_symbol_type stores the cache authoritatively and derives the legacy
-  // field via migrate_type_back exactly once (esbmc/esbmc#4715, B2 S4b).
-  symbolt snapshot_symbol;
-  snapshot_symbol.name = snapshot_name;
-  snapshot_symbol.id = snapshot_name;
-  set_symbol_type(snapshot_symbol, expr->type);
-  snapshot_symbol.lvalue = true;
-  snapshot_symbol.static_lifetime = false;
-  snapshot_symbol.file_local = false;
-
-  // Add to context (symbol table)
-  symbolt *added = context.move_symbol_to_context(snapshot_symbol);
-
-  // Return symbol expression (IRep2)
-  return symbol2tc(expr->type, added->id);
+  return declare_local_symbol(
+    "__ESBMC_old_snapshot_" + func_name + "_" + std::to_string(index),
+    expr->type);
 }
 
 expr2tc code_contractst::replace_old_in_expr(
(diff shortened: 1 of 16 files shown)


src/goto-programs/contracts/contracts.cpp
```diff
@@ -4710,6 +4710,28 @@ static bool havoc_pointed_to_array(
...
+// `assume false`, the path dies and a reachable assertion goes unreported
+// (#7009); where it does not, the caller silently keeps a stale value. An
+// initialiser gets this right only because its variable has no prior value to
+// keep. A discarded result leaves nothing to write, and nothing to assume.
+static void havoc_return_value(
+  const expr2tc &ret_val,
+  const locationt &call_location,
+  goto_programt &replacement)
+{
+  if (is_nil_expr(ret_val))
```
rafaelsamenezes: The comment's "nothing to assume" is wrong: with nil `ret_val` the ensures is still ASSUMEd, `__ESBMC_return_value` unreplaced. Add a discarded-result test.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -4993,6 +5015,9 @@ void code_contractst::generate_replacement_at_call(
     }
   }
 
+  // 2.5. Havoc the place the call assigns its result to.
+  havoc_return_value(ret_val, call_location, replacement);
```
rafaelsamenezes: `r = f(r)` still dies: ensures is instantiated with the caller's expression, so this havoc makes the ASSUME read `r == r + 1`. Use a temporary for the result.

LukeW1999: You are right, and it is the case the PR is named after. Thanks — every test I wrote used `f(void)`, so none of them had an argument to collide with the result.

Havocking the caller's place in situ cannot work, for the reason you give: the clause is instantiated with the caller's argument expressions, so `r = f(r)` carried `r` on both sides. The result has a name of its own now, and the caller's place is written after the clause has been assumed:

```
DECL signed int return_value$0;
ASSIGN return_value$0=NONDET(signed int);
ASSUME return_value$0 == r + 1
ASSIGN r=return_value$0;
```

`github_7009_self_assigned_arg_fail` asserts `r == 5` after the call, which the previous version proved vacuously; `github_7009_self_assigned_arg` is the positive half.

One consequence worth flagging rather than hiding: a pointer result now carries no value set, so a contract pinning it to an object proves the equality and then loses writes through it. Pinned as `github_7009_pointer_result_knownbug` and filed as #7066. The alternative is the false negative this PR removes, so I took the visible failure over the silent pass.

src/goto-programs/contracts/contracts.cpp
LukeW1999: Correct — the comment claimed something the code did not do. The ensures is assumed either way, and with nothing to rewrite `__ESBMC_return_value` to it was assumed over a symbol no instruction defines, which took the reachable assertion with it:

```
$ esbmc discard.c --replace-call-with-contract g
VERIFICATION SUCCESSFUL      # assert(0) after the call, plainly reachable
```

That reproduces on master too, so it was a second false negative sitting next to the first. Declaring the result unconditionally covers both your comments with one change. `github_7009_discarded_result_fail` pins it.

PR #7038, opened 2026-08-15
[python] Lower __ESBMC_old for scalar values

LukeW1999:
Closes the scalar contract MVP: `__ESBMC_old` is the last item of #6938's v1
scope (P1.3). Everything still open on that list is P4, the Python object
model, which the umbrella gates behind the C-side pointer-extent work.

**Stacked on #7033**, which the replace side needs — see below. Base is
`contracts-havoc-return-value`; it retargets to `master` when that merges.

## Symptom

A clause could name a parameter but not the value it held before the body ran,
so no postcondition could relate the two:

```
$ esbmc main.py --enforce-contract bump
ERROR: __ESBMC_ensures clause at line 3 contains a function call;
       a contract clause must be a pure expression
       (__ESBMC_old is not supported by the Python frontend yet)
```

## Fix

Build what the C macro expands to. `__ESBMC_old(x)` in C is
`*(__typeof__(x)*)__ESBMC_old_raw((void*)(&x))` — Python cannot spell an
address-of, but the frontend emits irep and can build one, so the intrinsic is
reached by the same route and every downstream step is the C one, unchanged.

#6938 P1.3 predicted a frontend-side snapshot on the grounds that "Python has
no address-of". The frontend has one either way, and reusing `__ESBMC_old_raw`
means the snapshot instruction, the wrapper materialisation and the call-site
materialisation are shared rather than reimplemented. The lowered wrapper is
the C shape:

```
bump (py:main.py@F@bump):
        DECL signed long int old_snapshot_bump_wrapper_0;
        ASSIGN old_snapshot_bump_wrapper_0=x;
        ASSUME x > 0                                        // contract requires
        DECL signed long int return_value;
        FUNCTION_CALL: return_value=contracts_original_bump(x)
        ASSERT return_value == old_snapshot_bump_wrapper_0 + 1  // contract ensures
```

## Scope, and what is rejected by name

A variable of `int`, `float` or `bool` — a parameter or a module-level global —
in either mode. Everything else is reported rather than lowered into something
else:

| shape | why |
|---|---|
| a container | a pointer into the object model, whose extent and aliasing no clause can state yet (P4) |
| a local | it does not exist at the point a snapshot would be taken |
| inside `__ESBMC_requires` | a precondition already speaks about the pre-state, and only the ensures is rewritten, so the requires would be asserted over an undefined symbol |
| outside a clause | there is no "before" to name, and the planted instruction has no reader |

Silently accepting any of them would be a false SUCCESSFUL, which the umbrella
singles out as the worst outcome available.

## Second fix: Python module globals at a replaced call site

Found while writing the float tests. A Python module global was never havocked
under `--replace-call-with-contract`: it carries `static_lifetime=false` so the
C static-init pass will not const-propagate it, which also hid it from
`havoc_static_globals`. An `ensures` naming one therefore contradicted its
pre-call value, the path died and everything after the call went unchecked.

```
$ esbmc repro.py --replace-call-with-contract touch
VERIFICATION SUCCESSFUL      # with `assert False` after the call
```

`rw_set.cpp:180` already recognises these by `mode == "Python" && !file_local`,
so the havoc uses the same rule rather than inventing one.

## Why this needs #7033

On the replace side the snapshot is taken at the call site, and the shape that
matters is the one where the result overwrites the variable the snapshot names:
`n = bump(n)`. Without #7033 the ensures is assumed about the `n` the caller
already had, contradicting the snapshot, so the path dies and the caller's
assertion passes vacuously. `old_replace` is therefore a test of both commits.

## Tests

Thirteen, in `regression/python-contracts/` (and `regression/python/` for the
loop-invariant inertness case).

| test | |
|---|---|
| `old_enforce` / `_fail` | was `future_old`; the `_fail` twin matches `^  contract ensures$` |
| `old_replace` / `_fail` | `n = bump(n)`; vacuous without #7033 |
| `old_float` / `_fail`, `old_bool` / `_fail` | the other two scalar types |
| `old_global` / `_fail` | enforce side; checked against `--goto-functions-only`, not the verdict alone |
| `old_global_replace` | replace side, `assert False` after the call |
| `old_inert` | no contract flag: the snapshot is a planted statement, so inertness is what is most at risk |
| `clause_old_fail`, `old_local_fail`, `old_in_requires_fail`, `old_outside_clause_fail` | the four rejections |

`intrinsic_outside_clause_fail` moves to `__ESBMC_is_fresh`; `__ESBMC_old` is
no longer unsupported, so it can no longer carry that test's point.

`python-contracts` 55/55, `function_contract` 376/376, `loop-invariants` 76/76.
In `python`, the 208 tests whose `main.py` mentions an `__ESBMC_` intrinsic or
a loop invariant all pass, run in two halves because the full suite does not
finish inside this repo's five-minute cap.

Contract-intrinsic handling in `build()` is consolidated into one
`build_contract_intrinsic`, so the function gains no cyclomatic complexity.


regression/python-contracts/intrinsic_outside_clause_fail/main.py
```diff
@@ -1,7 +1,9 @@
+# An unsupported intrinsic outside any clause is reported as the unsupported
+# call it is, rather than blamed on the contract clauses above it.
 def bump(x: int) -> int:
     __ESBMC_requires(x > 0)
     __ESBMC_ensures(__ESBMC_return_value > x)
-    __ESBMC_old(x)
+    __ESBMC_is_fresh(x, 8)
```
rafaelsamenezes, line 6: This was the only test covering `__ESBMC_old` outside a clause; it now builds a snapshot with no diagnostic. Add a test pinning that behaviour.

src/python-frontend/function_call/builder.cpp
```diff
@@ -194,6 +195,118 @@ static std::string describe_clause_node(const std::string &node_type)
   return "a " + node_type + " expression";
 }
 
+/// Scalars the wrapper can snapshot by value. Anything else is a pointer into
+/// the Python object model, whose extent and aliasing the contract layer has no
+/// way to state yet (#6938 P4).
+static bool is_snapshottable_scalar(const typet &type)
```
rafaelsamenezes, line 201: float and bool are admitted here and named in the error message, but every new test snapshots an int. Add float and bool tests.

regression/python-contracts/intrinsic_outside_clause_fail/main.py
LukeW1999: Thanks — I saw the coverage go and talked myself into it being out of scope, which was wrong.

Rejected rather than re-covered, since outside a clause there is no "before" to name and the planted instruction has no reader:

```
ERROR: __ESBMC_old at line 6 is only meaningful inside a contract clause
```

`old_outside_clause_fail` pins it, and `intrinsic_outside_clause_fail` keeps its point on `__ESBMC_is_fresh`.

Looking for the same hole elsewhere found two more, both of which reached the solver instead of the user:

- inside `__ESBMC_requires`. Only the ensures is rewritten, so the requires was asserted over an undefined symbol and failed spuriously.
- on a callee local, which has no pre-call value at all.

Both now rejected by name, with `old_in_requires_fail` and `old_local_fail`.

src/python-frontend/function_call/builder.cpp
LukeW1999: Added: `old_float` / `old_float_fail` and `old_bool` / `old_bool_fail`, each `_fail` twin matching `^  contract ensures$` so it has to fail for the stated reason.

Writing the float one turned up a false negative underneath. A Python module global was never havocked at a replaced call site, because it carries `static_lifetime=false` so the C static-init pass will not const-propagate it — which also hid it from `havoc_static_globals`. An ensures naming a global therefore contradicted its pre-call value and killed the path:

```
$ esbmc p8d.py --replace-call-with-contract touch
VERIFICATION SUCCESSFUL      # assert False after the call
```

`rw_set.cpp:180` already recognises these by `mode == "Python" && !file_local`, so the havoc uses the same rule. `old_global_replace` pins it with `assert False` after the call.

PR #7059, opened 2026-08-15
[docs] Design a JavaScript frontend reusing the Python infrastructure

rafaelsamenezes:
Adds `docs/design/javascript-frontend.md`: a design for a JavaScript frontend built as a sibling of the Python frontend rather than a standalone compiler.

Recommends Acorn (ESTree, zero deps, MIT) behind a Node subprocess first and an embedded QuickJS-ng later, and promotes `PyObject`/`PyListObject` plus the list, dict, string, set, class and exception handlers into a shared dynamic-language layer that both frontends consume. The one large piece of new work — widening the tagged-value path beyond literal assignment — is a gap in the Python frontend too, so it is shared work rather than JavaScript-specific work.

Documentation only; no code or build changes.

### DIFF
--- a/docs/design/javascript-frontend.md
+++ b/docs/design/javascript-frontend.md
@@ -12,22 +12,30 @@ The short version of the conclusion:
 - **The pipeline shape is reusable verbatim.** External parser → JSON AST →
   annotation pass → C++ converter → legacy `exprt`/`codet` into `contextt` →
   `clang_cpp_adjust` → `goto_convert`. Nothing about that is Python-specific.
-- **The runtime substrate is reusable after generalisation.** `PyObject`,
-  `PyListObject`, the string handler, the dict handler, the exception lowering
-  and the class-to-`struct_typet` builder all become a shared
-  *dynamic-language layer* that Python and JavaScript both consume.
+- **The runtime substrate is reusable after generalisation — behind an
+  interface, not by relocation.** `PyObject`, `PyListObject`, the string
+  handler, the dict handler, the exception lowering and the
+  class-to-`struct_typet` builder become a shared *dynamic-language layer*, but
+  only once each sits behind a stated contract (§5.5) and the extraction's entry
+  criteria are met (§10, M3). Method resolution is explicitly *not* shared:
+  Python's MRO and JavaScript's prototype chain are different algorithms, and
+  each frontend keeps its own resolver. Anything that cannot be stated without a
+  language conditional is duplicated instead of shared.
 - **The typing strategy is not reusable as-is, and this is the crux of the
   proposal.** ESBMC's Python frontend is, in practice, a static
   monomorphising translator that leans on PEP 484 annotations; its genuinely
-  dynamic path is narrow enough that assigning a non-literal to a
-  dynamically-typed variable raises "not yet supported"
-  (`src/python-frontend/converter/converter_stmt.cpp:3243`). JavaScript has no
-  annotations. The frontend therefore needs a real inference pass and a
+  dynamic path accepts an rvalue only when that rvalue is a numeric scalar or a
+  string, and raises "not yet supported" for anything else — a container, a
+  class instance, a call whose result type is unknown
+  (`src/python-frontend/converter/converter_stmt.cpp:3062-3075`). JavaScript has
+  no annotations. The frontend therefore needs a real inference pass and a
   first-class tagged-value path — and building that is the one large piece of
   new infrastructure, which pays back into Python immediately.
 
 Everything below is evidence for those three claims and a plan that follows from
-them.
+them. §1.5 states the strongest objection to the whole approach — that the
+Python frontend is too unsettled to build a second language on — and the three
+constraints the rest of the document is written to respect because of it.
 
 ---
 
@@ -147,21 +155,34 @@ materialised by symex — the SMT-level unbounded-array trick that makes symboli
 list indices tractable.
 
 **The dynamic-variable path is deliberately narrow.** A scalar variable only
-becomes `PyObject`-typed when `scalar_tag_candidates()` flags it at an `if`
-join point, and even then:
+becomes `PyObject`-typed when the tag analysis flags it, and assignment into a
+tagged variable is gated on the *type of the rvalue*:
 
 ```cpp
-if (is_tagged) {
-  if (ast_node["value"]["_type"] == "Constant") { get_tagged_scalar_assign(...); return; }
+if (dynamic_type_handler_.is_tagged(name)) {
+  if (ast_node.contains("value") && !ast_node["value"].is_null()) {
+    exprt rhs = get_expr(ast_node["value"]);
+    if (type_handler_.is_numeric_scalar_type(rhs.type()) ||
+        type_handler_.is_string_type(rhs.type())) {
+      dynamic_type_handler_.assign(rhs, …); return;
+    }
+  }
   throw std::runtime_error(
-    "assigning a non-literal value to a dynamically-typed variable is "
-    "not yet supported");
+    "assigning a value of this type to a dynamically-typed variable "
+    "is not yet supported");
 }
 ```
-— `src/python-frontend/converter/converter_stmt.cpp:3224-3245`
+— `src/python-frontend/converter/converter_stmt.cpp:3060-3076`
 
-That is the honest state of the art: tagged scalars support literal assignment
-at branch joins, and nothing else.
+So the boundary is *by rvalue type*, not by literal-vs-computed: a computed
+numeric or string rvalue does box, while a container, a class instance, a tuple
+or a call whose result type is unknown throws. `is_numeric_scalar_type`
+additionally excludes bitvectors narrower than 16 bits
+(`type/type_handler.cpp`).
+
+That is the honest state of the art, and it is the shape JavaScript cannot live
+within: in JS the *common* tagged rvalue is an object, an array or a call
+result — exactly the three the check rejects.
 
 **Classes reuse the C++ object model.** `python_class_builder::build`
 (`src/python-frontend/class/python_class_builder.cpp`) emits a `struct_typet`
@@ -204,6 +225,45 @@ Stated plainly, because the JS design has to plan around it:
 JavaScript needs (1), (2), (3) and (4) to be *good*, not special-cased. That
 determines the roadmap ordering in §10.
 
+### 1.5 The objection this design has to answer first
+
+The Python frontend is not a settled component. It is under active development,
+its dynamic path is the narrow thing §1.3 just described, and several of its
+subsystems are still changing shape. Building a second frontend on top of it,
+and then *promoting its internals to a shared interface*, risks freezing
+immature abstractions into a contract two languages depend on — and a shared
+layer that both frontends can edit is a faster route to entanglement than two
+separate frontends would be.
+
+This is the strongest argument against the proposal and it is not answered by
+asserting that reuse is cheaper. It is answered by three constraints, which the
+rest of this document is written to respect:
+
+1. **Reuse is by interface, not by relocation.** §5.5 defines an interface
+   before any code moves; nothing is shared because it happens to be sitting in
+   `src/python-frontend/`. A component qualifies for the shared layer only if it
+   has a stated contract, its Python-specific policy has been factored out of
+   it, and a second consumer actually exercises it. Components that fail that
+   test stay in the Python frontend and are *duplicated* on the JS side — a
+   deliberate accepted cost, because two clear implementations beat one
+   entangled one.
+2. **The dependency is one-way and enforced.** `src/js-frontend/` must not
+   include anything from `src/python-frontend/`, and vice versa; both include
+   `src/dynlang/`. This is checkable in CI (an include-path lint), and it is
+   what stops "shared" from degrading into "mutually coupled".
+3. **The extraction is gated, not scheduled.** M3 does not start because the
+   roadmap says it is next. Its entry criteria are in §10; if they are unmet,
+   the JS frontend proceeds on duplicated code and the extraction is deferred.
+   A design whose first large milestone is a refactor of someone else's
+   experimental subsystem should be able to survive that refactor not happening.
+
+There is a real benefit in the other direction that the objection should be
+weighed against: every gap in §1.4 is a gap JavaScript *forces*, and the four
+listed there (dynamic scalars, iteration, closures, `try`/`finally`) are the
+Python frontend's own known weaknesses. A second consumer is the usual way an
+experimental component acquires a stable interface. But that argument only
+holds under constraint (1); without it, the objection stands.
+
 ---
 
 ## 2. Parser selection
@@ -234,6 +294,15 @@ determines the roadmap ordering in §10.
 - **`acorn-walk` and `acorn-loose` are in the same repo**, giving a supported
   traversal utility for the annotation pass and an error-tolerant mode for
   diagnostics.
+
+**What Acorn does not give you.** It produces an AST and nothing else: no scope
+tree, no binding table, no type information. Scope resolution is a separate
+analysis the annotation pass owns (§5.2), and the recommendation there is to
+vendor `eslint-scope` for it. So the vendored bundle is Acorn + `acorn-walk` +
+`eslint-scope` (+ `esrecurse`/`estraverse`), all MIT or BSD-2 and all pure JS.
+"Zero dependencies" above is a property of Acorn itself and of the *native*
+toolchain — which is what §2.4's embedding argument actually rests on — not a
+claim that the frontend vendors exactly one file.
 
 ### 2.3 Why not the others
 
@@ -292,9 +361,24 @@ choose a *pure-JS* parser over a native one: pure JS can be embedded in a 400 KB
 engine, whereas SWC or Tree-sitter would each pull their own native toolchain.
 Stage 2 is scheduled as milestone 7 so it is never on the critical path.
 
-**Contract between stages.** Both stages produce byte-identical JSON, so the
-converter is unaware of which is in use, and Stage 1 remains a supported
-fallback (`--js-parser=node|embedded`).
+**Contract between stages.** The contract is the *ESTree AST*, not its
+serialisation: both stages must produce structurally identical trees — same node
+kinds, same children, same `start`/`end`/`loc` — so the converter is unaware of
+which is in use, and Stage 1 remains a supported fallback
+(`--js-parser=node|embedded`).
+
+Byte-identical JSON is deliberately *not* the requirement. `JSON.stringify` is
+free to differ between Node's serialiser and QuickJS-ng's in key order, number
+formatting (`1e21` vs `1000000000000000000000`, `-0`), and non-ASCII escaping,
+and pinning the contract to text would make an unrelated engine upgrade look
+like a parser regression. Equality is therefore defined on the parsed tree, by a
+structural comparison that ignores key order and normalises numeric literals.
+
+Where a *textual* artefact is genuinely needed — golden-file tests, and the
+cross-backend parity check in M7 — it is produced by one canonical serialiser
+that both backends feed (recursive key sort, fixed number formatting, `\u`
+escaping above ASCII), so the canonical form is a property of the test harness
+rather than of either engine.
 
 ---
 
@@ -309,9 +393,9 @@ fallback (`--js-parser=node|embedded`).
 | `string` | `str` | `char[N]` / `char *` | `string.c`, `string_handler` | `index`, `constant_string` | **R** (UTF-16 caveat) |
 | `boolean` | `bool` | `bool` | — | `constant_bool` | **R** |
 | `null` | `None` | `pointer(bool)` sentinel | — | `isnone2t` | **R** |
-| `undefined` | *(none)* | second sentinel | — | `isnone2t` variant | **N** (small) |
+| `undefined` | *(none)* | sentinel, own pointee type (§4.1) | — | `isnone2t` variant | **N** (small) |
 | `symbol` | *(none)* | opaque unique id | — | `unsignedbv` | **N** (deferred) |
-| array | `list` | `JsArrayObject` | `list.c` | infinite `array` + `index` | **G** |
+| array | `list` | `JsArrayObject` (+ presence, `length` invariant) | `list.c` | infinite `array` + `index` | **G** (§4.2.0) |
 | object (fixed shape) | class instance | `struct_typet` | `__ESBMC_new_object` | `member`, `with` | **R** |
 | object (dynamic keys) | `dict` | `DynDict` | `python-dict/` | struct + arrays | **G** |
 | class | class | `struct_typet` + methods | `python_class_builder` | `member`, indirect call | **R** |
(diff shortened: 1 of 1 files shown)


rafaelsamenezes:
I don't really like this plan:

1. Python frontend is experimental, starting a plan over an experimental environment is a quick way to a spaghetti mess. 
2. AFAIK, JS has some properties that should be more detailed (e.g., list and values being linked lists).
3. Sharing between frontends is not as trivial. We probably should design an interface that handles all common language constructs (e.g. vtables)

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+**`number`.** One type, IEEE-754 binary64: `floatbv(52,11)`, the same
+`double_type()` the Python frontend produces for `float`. Arithmetic maps to
+`ieee_add`/`ieee_sub`/`ieee_mul`/`ieee_div` with the rounding-mode operand, and
+`Number.isNaN`/`isFinite` map to `isnan2t`/`isfinite2t`.
+
+Two JavaScript-specific obligations:
+
+- **Bitwise operators are `ToInt32`.** `a | 0`, `a << 1`, `a & 0xff` and `~a`
+  all convert through a modular 32-bit signed integer. Lower as
+  `typecast(floatbv → signedbv(32))` under JS's truncate-toward-zero-modulo-2³²
```
pmatos: The prescribed lowering for JS bitwise operators — `typecast(floatbv → signedbv(32))` plus a bit op plus a cast back — does not implement `ToInt32`. Per the ES spec, `ToInt32` is a *total* function: NaN, +Infinity, -Infinity, and ±0 must all map to +0 before truncation, and the result is then reduced modulo 2^32 (so `4294967296 | 0` must be `0`, and `2147483648 | 0` must be `-2147483648`). A bare float→int32 typecast has no modulo step and is undefined/implementation-defined for out-of-range or NaN operands (SMT-LIB's `fp.to_sbv` is explicitly underspecified there too).

This is exactly the operator the doc itself calls out two sentences later as "the single most common source of unsoundness in naive JS models" — worth spelling out the real algorithm (NaN/Inf/±0 → +0, truncate toward zero, reduce mod 2^32, wrap into the signed range) rather than a plain typecast, since an implementer following this passage literally would ship the unsoundness the doc is trying to avoid.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+
+### 4.4 Property lookup and prototype chains
+
+Property access is the hottest operation in JavaScript and needs to resolve
+statically wherever possible.
+
+**Resolution ladder**, cheapest first:
+
+1. **Known shape, own property** → `member2t`. Free.
+2. **Known shape, inherited property** → walk the `__proto__` chain *at
```
pmatos: Step 2 of this resolution ladder walks the `__proto__` chain *at conversion time* and bakes the result into a `member2t` on the resolved subobject — but §5.5, a few hundred lines later, gives exactly the opposite characterization as the reason `dyn_dispatch` can't share Python's C++-style resolver: "JavaScript walks a `[[Prototype]]` chain that is mutable at runtime" (`Object.setPrototypeOf`, reassigning `Foo.prototype`, monkey-patching a method after construction). §7 never defers prototype reassignment from MVP scope.

So step 2 statically resolves exactly the thing the document elsewhere says can't be statically resolved, with no stated precondition here (e.g. "no reachable prototype mutation reaches this call site") the way Tier-1 shape inference in §4.2.1 states one for object keys ("the inference must be conservative in both directions"). As written, an implementer following the ladder literally can silently miscompile ordinary mixin/polyfill-style JS that reassigns a prototype method after some objects already exist.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+- `dyn_binop(op, a, b)` dispatching on `(a.type_id, b.type_id)` — where the
+  inference pass has narrowed the possible tag set, emit only those arms, so a
+  variable known to be "string or number" costs a two-way branch rather than a
+  full dispatch table.
+- `dyn_unbox(v, T)` with a *checked* unbox that raises `TypeError` on tag
+  mismatch, so type confusion is a reported property rather than silent
+  reinterpretation.
+- `dyn_truthy(v)`, `dyn_typeof(v)`, `dyn_equals(v, w)` for `==` vs `===`.
+
+`typeof` is a `type_id` read; `instanceof` is `isinstance2t`, already resolved by
```
pmatos: `typeof` as "a `type_id` read" misses two ECMAScript special cases that would corrupt exactly the narrowing analysis §5.2 step 5 relies on ("`typeof x === "string"`... These are the idioms real JS uses to be safe, and honouring them is what keeps most variables monomorphic in practice"):

- `typeof null === "object"` — a preserved historical spec quirk. `null`'s type tag reads as `"object"`, not `"null"`, even though §4.1 goes to some length to give `null` and `undefined` distinct `type_id`s.
- `typeof someFunction === "function"` — closures are modelled as `{fn, env}` structs per §4.3, so a raw tag read on a closure value would yield something like `"object"`/`"closure"`, not `"function"`.

Real JS narrows on both idioms constantly (`typeof x === "function"`, `x !== null`), so a literal tag-read implementation of `typeof` would misclassify exactly the values programmers use it to distinguish, feeding wrong branches into the narrowing lattice.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+
+That is the honest state of the art, and it is the shape JavaScript cannot live
+within: in JS the *common* tagged rvalue is an object, an array or a call
+result — exactly the three the check rejects.
+
+**Classes reuse the C++ object model.** `python_class_builder::build`
+(`src/python-frontend/class/python_class_builder.cpp`) emits a `struct_typet`
+with a method table and base classes, instances are allocated through
+`__ESBMC_new_object` (`src/c2goto/library/python/list.c:14`, intercepted by
+symex's `symex_mem_inf` to produce a non-expiring typed object), and
```
pmatos: This claim — that `__ESBMC_new_object` is "intercepted by symex's `symex_mem_inf`" — doesn't match current `master`, and it's repeated as load-bearing justification in three places (here, §6.1 line 1014, §9.3 line 1137).

`symex_mem_inf` (`src/goto-symex/builtin_functions/memory_alloc.cpp:482`) is reached only via the `__ESBMC_create_inf_obj` handler (`src/goto-symex/symex_main.cpp:1118-1132`, used for the list `items` infinite array). `__ESBMC_new_object` is intercepted by a separate hand-written block at `symex_main.cpp:1058-1116`, whose own comment at line 1065 says it "Mirrors `symex_mem_inf`'s binding" rather than calling it.

(The doc's error likely traces back to a stale comment already in `src/c2goto/library/python/list.c:11`, which says the same thing — so this may be pre-existing doc/code drift in ESBMC itself rather than something this PR introduced. Still worth naming the actual mechanism, since the JS class-instance-allocation reuse argument rests on it.)

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+| `symbol` | *(none)* | opaque unique id | — | `unsignedbv` | **N** (deferred) |
+| array | `list` | `JsArrayObject` (+ presence, `length` invariant) | `list.c` | infinite `array` + `index` | **G** (§4.2.0) |
+| object (fixed shape) | class instance | `struct_typet` | `__ESBMC_new_object` | `member`, `with` | **R** |
+| object (dynamic keys) | `dict` | `DynDict` | `python-dict/` | struct + arrays | **G** |
+| class | class | `struct_typet` + methods | `python_class_builder` | `member`, indirect call | **R** |
+| inheritance | inheritance | base subobject | `get_bases`, `clang_cpp_adjust` | `member` chain | **R** |
+| prototype chain | *(none)* | `__proto__` link | new OM | convert-time walk / OM loop | **N** |
+| function | function | function symbol | — | `code_function_call` | **R** |
+| closure | nested function | `{fn, env}` pair | `__ESBMC_new_object` env | fn-pointer call | **N** (see §4.3) |
+| `this` | `self` | leading parameter | — | — | **R** |
```
pmatos: This table marks `this` as **R** ("pure reuse", modelled like Python's `self` as a "leading parameter"), but §6.2 (line 1029) lists "`this` binding" as new JavaScript-specific work scheduled for milestone 6: "conversion-time: method call, `call`/`apply`/`bind`, arrow lexical capture."

§6.2 is the one that's right: Python's `self` is bound statically at method-definition time, while JS's `this` for an ordinary function is determined by the *call site* — a detached method reference like `const f = obj.m; f()` gets `this === undefined` in the MVP's always-strict ESM modules, and any property read through it should be a reported `TypeError`. Arrow functions don't have their own `this` at all; they lexically capture the enclosing one, structurally like a closure, not a parameter.

An implementer skimming just this table would model `this` as a fixed leading parameter and silently lose the unbound/rebound cases — a common real JS bug class this tool exists to catch.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+| exception | exception | `cpp-throw`/`cpp-catch` | `remove_exceptions` | `code_cpp_throw` | **R** |
+| `finally` | `finally` (restricted) | — | — | — | **G** |
+| dynamic typing | annotations + narrow tags | `DynObject` | shared tag OM | `isinstance2t` | **G** — the big one |
+| `for...of` | `for … in range()` | desugared | — | `code_goto` loop | **G** |
+| modules (ESM) | `import` | module manager | `module_manager` | — | **G** |
+| `typeof` | `type(x)` | tag read | — | `isinstance2t` | **R** |
+| `instanceof` | `isinstance` | value-set query | `python_builtins.cpp` | `isinstance2t` | **R** |
+| `in` | `in` (dict) | key membership | `python-dict/` | `hasattr2t` | **R** |
+
+Aggregate: of 24 constructs, 11 reuse directly, 7 generalise a Python
```
pmatos: This aggregate doesn't match a literal count of the table's own verdicts (rows just above): **R appears 13 times** (`number`, `string`, `boolean`, `null`, object-fixed-shape, `class`, inheritance, `function`, `this`, exception, `typeof`, `instanceof`, `in`), **G appears 7 times** (`bigint`, array, object-dynamic-keys, `finally`, dynamic typing, `for...of`, modules), **N appears 4 times** (`undefined`, `symbol`, prototype chain, closure) — 13/7/4, not 11/7/6. The 24-row total happens to match, which masks the R/N split being off by 2 each.

Separately, "of those six, three... and two" only enumerates 5 items, and "dynamic typing" is folded into that "new" bucket even though the table marks its own verdict as **G** ("the big one"), not **N**.

```suggestion
Aggregate: of 24 constructs, 13 reuse directly, 7 generalise a Python
component, 4 are new — three (`undefined`, `symbol`, prototype chains) are
small and one (closures) is the load-bearing new construct; dynamic typing
(marked **G** above) is the other piece of load-bearing work, generalising
the Python tagged-value path rather than introducing a new one.
```

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+design, it is a `git mv`. Sharing implementation between two frontends without
+first agreeing what is shared is how the layer becomes a place where each
+language's special cases accumulate behind `if (lang == PYTHON)` — the outcome
+§1.5 exists to avoid. So the shared layer is defined as a set of **interfaces
+the two converters call**, and code moves only once it sits behind one.
+
+The layer is five interfaces. Each is stated as *what the consumer may assume*,
+because that is the part that has to survive a second consumer:
+
+| Interface | Operations | Language-specific policy it must **not** contain |
```
pmatos: This interface table's "must not contain" column is contradicted by the surrounding prose in several places — the stated exclusions and the shared-layer content disagree about where things actually live:

- **`dyn_dispatch`**'s Operations column ("resolve a method/property name on a receiver shape to a callable") *is* the thing its own exclusion column disclaims ("the resolution algorithm itself") — subtract one from the other and nothing is left. Neither language's resolver actually touches `src/dynlang/`: Python "reaches `clang_cpp_adjust`" directly (line 966), and "§4.4's ladder is the JS resolver, and it lives in `src/js-frontend/js_object/`, not in `src/dynlang/`" (line 979). §9.3's own table marks this row "**interface**" rather than "**shared**" — a tacit concession it's different from the other four.
- **`dyn_container`** excludes "hole semantics" and "OOB behaviour" as language-specific policy, but §4.2.1 puts `present`/`props` directly into the shared `JsArrayObject` struct's layout, and §6.2 rows "Array holes / OOB → `undefined`" as "array OM modification" — i.e. inside the shared layer, not behind a caller-supplied policy.
- **`dyn_value`**'s listed operations (`box`, `unbox`, `tag_of`, `truthy`, `type_id_for`) omit `dyn_binop`/`dyn_equals`, which §5.3 says must be built — and `dyn_binop` is precisely where the excluded "which tags coerce to which" policy has to live.
- **`dyn_string`** excludes "encoding (UTF-16 vs byte)", yet §4.1 claims full UTF-16 support "is a milestone-9 item and again benefits both frontends" — which only makes sense if it lands inside `dyn_string`, the interface whose contract just excluded it.
- The layer's membership is stated three different ways: **five** interfaces in this table, **seven** modules in this section's own architecture diagram just above it (adding `dyn_exception`/`dyn_scope`), and **six** in the §9.1 diagram — which drops `dyn_dispatch` entirely, the one interface this section's own prose calls "the one that decides whether this works" (line 963).

None of these individually breaks the proposal, but together they suggest the interface boundary here is aspirational rather than load-bearing as currently specified — worth tightening before M3's entry criteria (which require "no interface operation carries a language conditional") get evaluated against it.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+1. The five `src/dynlang/` interfaces (§5.5) are written down as headers, with
+   each operation's contract stated, and reviewed by a maintainer of the Python
+   frontend.
+2. The Python frontend has been rebuilt against those headers *in place* —
+   still under `src/python-frontend/`, no files moved — and `regression/python`
+   is green. This is the step that proves the interface is adequate before it
+   is load-bearing for two languages.
+3. No interface operation carries a language conditional. A component that
+   needs one is struck off the shared list and duplicated instead.
+4. `js_converter` (from M2) actually calls at least `dyn_value` and
```
pmatos: This entry criterion requires "`js_converter` (from M2) actually calls at least `dyn_value` and `dyn_container`." But M2's deliverables (§10 M2) are purely monomorphic — "numeric literals, arithmetic, comparison, `let`/`const`/`var`, `if`, `while`, `for`; `assert` and the ESBMC intrinsics; program entry synthesis" — nothing that boxes a value or touches a container. The doc's own roadmap doesn't introduce `dyn_value`/`dyn_container` traffic until **M4** ("widened `dyn_assign`/`dyn_binop`/`dyn_unbox`") and **M5** (arrays/objects over the shared array OM) — both scheduled *after* M3.

This also directly conflicts with the closing "Sequencing note" further down: "M1, M2 and M3 can be worked in parallel by separate contributors once M1's JSON contract is fixed." M3 can't finish gated on M2's converter calling into `dyn_value`/`dyn_container` while also running in parallel with M2's development.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+which is `PyListObject` plus the two fields §4.2.0 forces. The infinite-array
+trick (`__ESBMC_create_inf_obj`) carries over unchanged and is precisely what
+makes symbolic indices tractable; `present` is a second infinite array, so holes
+cost one extra bit per touched index and nothing for arrays that have none —
+the annotation pass marks an array *dense* when no hole can reach it, and the
+`present` reads fold away.
+
+JS-specific behaviour that Python's list lacks:
+
+- **OOB read yields `undefined`**, it does not raise, so the bounds *property*
```
pmatos: This default — OOB reads return `undefined` rather than raising, with bounds checking demoted to an opt-in `--js-array-oob=error` flag — is in tension with two other claims in the doc: §7 lists "array bounds" under MVP **Verification properties** without qualifying it as opt-in, and §8's reuse table credits `goto_check` with "bounds, division by zero, overflow, pointer safety, all **for free**."

If the MVP's default is `--js-array-oob=undefined` (as this paragraph implies), out-of-bounds array access is not actually flagged by default — so "array bounds" isn't the for-free, on-by-default supported property §7/§8 present it as. Worth reconciling: either state the permissive default explicitly in §7, or note in §8 that the `goto_check` bounds reuse doesn't apply as-is to JS array indexing (it's superseded by the array OM's own OOB→`undefined` behaviour).

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+*reusing* the existing mechanisms, not for *merging the two frontends' code*.
+The proposal takes reuse and declines the merge — §5.5's interface boundary and
+§10's M3 entry criteria are what keep those two things separate, and if the
+criteria cannot be met, the reuse happens by duplication and the conclusion
+above is unaffected.
+
+Where JavaScript genuinely differs — prototype chains, capturing closures, the
+`ToInt32` bitwise rules, `undefined` as distinct from `null`, `==` coercion —
+the proposal adds *JavaScript-specific components on top of the shared layer*,
+not a competing substrate. That boundary is what §9.4's list of five new
```
pmatos: This closing reference to "§9.4's list of five new components" is stale: §9.4 opens with "Seven:" and explicitly flags the correction — "Earlier drafts of this document counted five and described the array change as a field rename; that was wrong, and the count is stated here so the estimate in §10 is not read as smaller than it is." This sentence reverts to the pre-correction count the doc just spent a sentence disowning.

(Separately, this sentence's own "five" — prototype chains, capturing closures, `ToInt32`, `undefined`/`null`, `==` coercion — is a list of *differences*, not a count of §9.4's components, which compounds the confusion for a reader trying to reconcile the two.)

```suggestion
not a competing substrate. That boundary is what §9.4's list of seven new
```

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1479 @@
...
+
+Member order in the emitted struct is the §4.2.0 #5 enumeration order, so
+`Object.keys` and `JSON.stringify` are a walk over `member_names` and need no
+side table. The inference must be conservative in *both* directions: a key that
+might be added on some path, and a key whose creation order differs between
+paths, both disqualify the site from Tier 1.
+
+*Tier 2: open objects → `DynDict`.* When keys are computed (`o[k]`), added, or
+deleted, fall back to the generalised dict handler (`src/python-frontend/
+python-dict/`, ~4k lines across 8 translation units) keyed by string with
```
pmatos: Minor figure inconsistency, grouped here with a few similar ones found elsewhere in the doc: this cites `python-dict/` at "~4k lines," but §6.1's table cites the same directory at "4.8k lines" (an actual count on `master` gives ~4.77k, so §6.1's figure is the accurate one).

```suggestion
python-dict/`, ~4.8k lines across 8 translation units) keyed by string with
```

A few other small figure/citation mismatches worth a pass, none independently worth a separate comment: the `regression/python/` test count is cited as "4,590" (here, and at lines ~1323/1449) but counting `test.desc` files on `master` currently gives 4,607; the same `converter_stmt.cpp` code block is cited with two different line ranges in different places (`:3062-3075` near the top of the doc vs `:3060-3076` here and near the end); the "~0.6–1.2 s" GOTO-creation-time range in §2.3 cites "Examples 2 and 5" of the Python README, but Example 5's actual figure there is 0.896s, not ~1.2s; and the "22 mixins" figure for `preprocessor/*.py` (§1.2) doesn't match the 18 `*Mixin` classes actually composed into `Preprocessor` in `src/python-frontend/preprocessor/__init__.py`. None of these affect the design's conclusions, but they're the kind of detail a reader might quote later.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+        type_handler_.is_string_type(rhs.type())) {
+      dynamic_type_handler_.assign(rhs, …); return;
+    }
+  }
+  throw std::runtime_error(
+    "assigning a value of this type to a dynamically-typed variable "
+    "is not yet supported");
+}
+```
+— `src/python-frontend/converter/converter_stmt.cpp:3059-3076`
```
lucasccordeiro, line 175: Wrong location: `converter_stmt.cpp:3059-3076` is numpy-view rejection. The tagged-assignment gate is at `:5326-5373`. Fix here and at lines 30 and 1651.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+So the boundary is *by rvalue type*, not by literal-vs-computed: a computed
+numeric or string rvalue does box, while a container, a class instance, a tuple
+or a call whose result type is unknown throws. `is_numeric_scalar_type`
+additionally excludes bitvectors narrower than 16 bits
+(`type/type_handler.cpp`).
+
+That is the honest state of the art, and it is the shape JavaScript cannot live
```
lucasccordeiro, line 183: Stale on master: #7075 added a `mint_retyped_symbol` fallback, so a container rvalue throws only inside a class, a loop body or a nested block. Requalify this claim.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+exception hierarchy are all just Python source.
+
+### 1.4 Where the Python frontend is weak
+
+Stated plainly, because the JS design has to plan around it:
+
+1. **No general dynamic scalar.** As above.
+2. **Iteration is special-cased.** "Only `for` loops using the `range()`
+   function are supported" (`src/python-frontend/README.md`, *Limitations*); the
+   preprocessor rewrites other shapes where it can.
```
lucasccordeiro, line 229: `src/python-frontend/README.md:410` no longer says only `range()` is supported: list/tuple/string/dict-view/generator plus `enumerate`/`zip`/`reversed` all iterate. Update, since §10 ordering rests on this.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+`getprop` node is discussed in §4.4), but the MVP maps entirely onto existing
+terms.
+
+### 1.2 The Python frontend, end to end
+
+```
+  main.py
+    │  python_languaget::parse()                       python_language.cpp:92
+    │    spawns `python3 parser/__main__.py`  ← FLAIL-mangled .py files, dumped
+    │    to a temp dir at first use (dump_python_script(), :63)
```
lucasccordeiro, line 87: `dump_python_script()` does not exist. The mechanism is `python_languaget::register_bundled()` (`python_language.cpp:65`) plus `file_operations::filesystemt::materialize()` (`:95`). Name the real function.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+### 1.3 How Python values are actually represented
+
+This is where the reuse analysis has to be precise, because the headline
+"ESBMC supports a dynamically typed language" is doing more work than the
+implementation supports.
+
+**Scalars are statically monomorphised.** `type_handler::get_typet`
+(`src/python-frontend/type/type_handler.cpp:429`) maps annotation strings to
+concrete types:
```
lucasccordeiro, line 117: Line citations have drifted from master: `get_typet` is `:461` not `:429`, the `type_id` hash `:1173` not `:1137`, `build_optional_type` `:1498` not `:1451`. Re-derive them.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+  Getting the order wrong — casting first and reducing afterwards — is exactly
+  the bug this paragraph exists to prevent. It is a small, self-contained
+  helper, and it is the single most common source of unsoundness in naive JS
+  models, so it belongs in milestone 4, not later.
+- **Integer-valued fast path.** Most computational JS uses numbers that are
+  integers. The annotation pass (§5) marks a variable *integral* when every
+  reaching definition is an integer literal, an integer-preserving operation, or
+  a `| 0`; such variables are emitted as `signedbv(64)` with an overflow guard
+  rather than `floatbv`. This is exactly the widening the Python annotator does
```
lucasccordeiro, line 490: A `signedbv(64)` integral path diverges from JS above 2^53, where doubles stop being exact, and a 64-bit overflow guard does not catch that. State the 2^53 precondition.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+### 9.2 Data flow
+
+```
+ main.js
+   │ js_languaget::parse()
+   ├─▶ [Stage 1] node parse.js      ‖  [Stage 2] embedded QuickJS-ng
+   │       acorn.parse(src, {ecmaVersion:"latest", sourceType:"module",
+   │                         locations:true})
+   ├─▶ preprocess.js   desugar to a small core; lambda-lift closures
```
lucasccordeiro, line 1369: Ordering contradiction: `preprocess.js` runs before `annotate.js` here, but §4.3 step 1 does closure conversion over the scope tree `annotate.js` builds. Reorder or split scope resolution out.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+therefore sequenced after M5**, and until it lands the JS frontend calls the
+interfaces where M3a has defined them and duplicates where it has not.
+
+If M3a's criteria are never met, M3b never happens: the JS frontend proceeds
+entirely on duplicated implementations and the rest of the roadmap is unaffected.
+That is the point of §1.5 constraint 3 — this plan has to survive its own
+largest refactor not happening.
+
+**Deliverables.** *M3a:* interface headers; the Python frontend rebuilt against
+them in place; `models/errors.js`, `models/nondet.js`. *M3b:* `src/dynlang/` and
```
lucasccordeiro, line 1504: `models/errors.js` and `models/nondet.js` are JS-only but sit behind M3a's shared-layer entry criteria, which M4 then depends on. Move them out of M3a.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+### M8 — Benchmarks and evaluation
+
+**Deliverables.** A `regression/javascript-intensive/` suite mirroring
+`regression/python-intensive/`; a HumanEval-JS style corpus (the repo already
+carries `regression/humaneval`); comparison against the Python frontend on
+algorithms implemented in both languages; performance and coverage reporting.
+**Risks.** Selection bias — a corpus tuned to what the frontend supports proves
+nothing. Draw from an external, pre-existing corpus.
+**Testing.** Full suite under the project's 5-minute cap, scoped per-suite.
```
lucasccordeiro, line 1600: There is no 5-minute cap: AGENTS.md caps a full-suite run at 10 minutes and `ESBMC_REGRESS_TIMEOUT` is 1200s per test. Correct the figure.

docs/design/javascript-frontend.md
```diff
@@ -0,0 +1,1669 @@
...
+
+- A monomorphic `number` costs one `floatbv`; an integral-inferred number costs
+  one `signedbv(64)`, which is materially cheaper to bit-blast.
+- A tagged value costs a 4-field struct plus a pointer dereference per read, and
+  each dispatch adds one `if2t` arm per candidate tag. Narrowing the candidate
+  set is therefore the whole game — hence the emphasis on step 5 above.
+- Shape-inferred objects cost exactly what C structs cost; open objects cost
+  what the dict OM costs.
+- The `--ir` (integer/real) mode's float side-buffer
+  (`__ESBMC_float_buf`, `src/c2goto/library/python/list.c:27`) is a
```
lucasccordeiro, line 1018: `__ESBMC_float_buf` is a fixed 4096-entry array with a never-reused index (`list.c:27-29`). JS tags far more values, so `--ir` exhausts it; "unchanged" understates this.

Issue #4243, opened 2026-05-01
bundled <array>: value-init ({}) does not zero-initialise elements — nondet elements cause false-positive overflow VCCs


lucasccordeiro:
## Summary

When `std::array<T, N>` (or a class that contains one) is value-initialised
with `{}`, ESBMC's bundled `<array>` leaves the internal `elems[N]` storage
as nondet symbolic values instead of zero. This produces false-positive
overflow VCCs in any arithmetic that reads from the supposedly zero-initialised
array before a write.

A plain aggregate shim (`struct { T elems[N]; }`) is modelled correctly —
elements are zero. Reproducible in ESBMC 8.2.0.

---

### 1. `std::array` is required to be an aggregate

The C++ standard has required `std::array` to be an aggregate since C++11
([array.overview]). ESBMC's bundled header declares `elems` as `private`:

```cpp
template <typename T, std::size_t N>
class array {
private:
  value_type elems[N];    // private non-static data member → NOT an aggregate
  ...
};
```

A class with private non-static data members does not satisfy the aggregate
conditions in any C++ version (C++11 through C++23). The bundled header is
therefore a non-conforming implementation.

### 2. Value-initialisation must zero-initialise the elements

The deciding factor for whether the zero-initialisation step applies is
**whether the class has a user-provided constructor**, not whether the class
is an aggregate. C++20 [dcl.init.general]/8 states:

> To value-initialize an object of type T means:
> — if T is a class type with a default constructor that is neither
>   user-provided nor deleted, then the object is **zero-initialized**
>   and, if T has a non-trivial default constructor, default-initialized.

The bundled `class array` has **no declared constructor at all**, so its
implicit default constructor is neither user-provided nor deleted. This
places it squarely in the "zero-initialize first" branch.

Zero-initialisation of a class type zero-initialises every non-static data
member regardless of access specifier — private `elems[N]` included
([dcl.init.general]/6). The implicit default constructor of `class array`
is trivial, so no further default-initialisation follows.

In short: both aggregate and non-aggregate classes without a user-provided
constructor undergo the same zero-initialisation step. The non-aggregate
status of the bundled `<array>` is an independent conformance defect, not
the cause of the zero-init modelling failure.

---

## Minimal reproducer

```cpp
// esbmc_array_zeroinit.cpp  —  run with --overflow-check
#include <array>
#include <cstdint>

int main()
{
    std::array<int32_t, 4> buf{};   // value-init: all elements must be 0
[...]

Issue #4183, opened 2026-04-25
C++ frontend: std::array crashes gen_vptr_initializations on Clang 22 (split from #4180 Bug 1)


lucasccordeiro:
Split off from #4180 Bug 1 — issue #4180 is being closed by the qualified-`constexpr` fix; this tracks the remaining `std::array` crash.

## Reproducer

```cpp
#include <array>
#include <cstdint>

int main()
{
    std::array<uint8_t, 4> a{};
    return a[0];
}
```

```
$ esbmc bug.cpp --std c++20
Target: 64-bit little-endian aarch64-unknown-macos with esbmclibc
Parsing bug.cpp
Converting
Assertion failed: (ctor_class_symb), function gen_vptr_initializations,
file clang_cpp_adjust_code_gen.cpp, line 48.
```

## Root cause (under investigation)

After the Clang 22 elaborated-type fix in fcfdf1afc2 (#4144), `get_decl_name` for a `RecordDecl` prepends the kind name, so the class symbol id is `tag-struct Foo` (or `tag-struct std::__1::array<unsigned char, 4>`). But `annotate_class_method` in `src/clang-cpp-frontend/clang_cpp_convert.cpp:2031-2039` still computes `parent_class_id = tag-Foo` (no kind prefix), so the ctor component carries a `#member_name` that does not match any symbol.

Confirmed on a minimal non-virtual class with `--symbol-table-only`: symbol id is `tag-struct Foo` while the ctor's component carries `#member_name=tag-Foo`. The mismatch is silent when `need_vptr_init=false`, which is why simple classes do not trip it. For `std::array` something flips `has_vptr_component` (likely a nested class with virtuals), `need_vptr_init` becomes true, and the lookup at `clang_cpp_adjust_code_gen.cpp:46-48` asserts.

## Candidate fix

In `annotate_class_method`, under `CLANG_VERSION_MAJOR >= 22`, prepend `cxxmdd.getParent()->getKindName().str() + " "` to `parent_class_name` so it matches what `get_decl_name` produces, or rely on `get_decl_name` directly to derive the id. Needs a build + regression run to confirm.

## Why split

The qualified-`constexpr` half of #4180 was small, well-scoped, and already has regression tests (`regression/esbmc-cpp/cpp/github_4180_qualified_constexpr{,_fail}/`). The `std::array` half needs a `get_decl_name` / `annotate_class_method` audit that is independent and may surface other latent mismatches across the C++ frontend.

PR #7062, opened 2026-08-15
[clang-cpp-frontend] Zero-init trivial value-init mem-initializers

zacheryasc:
## Description

When you write `B() : bitmap_()`, the frontend turns it into a call to `array#`, a constructor that was never actually generated. Symex hits a function with no body, warns, and skips it. Nothing ever writes the zeroes and the array contents stay random. `std::array<int,4> buf{}` was already fixed by #4244: that form just becomes the constant `{0,0,0,0}` directly. The two spellings go through different code, though, and the constructor one never got the same fix.

```cpp
struct B
{
  std::array<std::uint32_t, 8> bitmap_;
  B() : bitmap_() {}
};

int main()
{
  B x;
  __ESBMC_assert(x.bitmap_[0] == 0, "elements must be zero"); // FAILED on master
}
```

## Change

In `get_function_body`, where ctor member initializers get converted, if the initializer is a `CXXConstructExpr` that value-initializes a member whose default ctor is trivial, we now assign a zero constant straight to the member instead of emitting the call. For a trivial default constructor that's exactly equivalent. It isn't user-provided, has no vptrs to set and no members to construct, so zeroing is all the call would ever have done. This matches the standard's value-initialization rule ([dcl.init.general]/8): a class type whose default ctor is neither user-provided nor deleted is zero-initialized. Copy/move inits and every ctor that has a real body still take the old path unchanged.

## Validation

- New regressions: `github_4243_mem_init` / `_fail` twin.

Refs #4243.

### DIFF
--- a/regression/esbmc-cpp/cpp/github_4243_mem_init_fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_4243_mem_init_fail/main.cpp
@@ -1,17 +1,21 @@
+// Counterpart of github_4243_mem_init: the mem-initializer zeroes the
+// elements, but an explicit write in the ctor body must win over that
+// zero-init.
 #include <array>
 #include <cstdint>
 
 struct B
 {
   std::array<std::uint32_t, 8> bitmap_;
-  B() : bitmap_() {}
+  B() : bitmap_()
+  {
+    bitmap_[0] = 7;
+  }
 };
 
 int main()
 {
   B x;
-  __ESBMC_assert(
-    x.bitmap_[0] != 0,
-    "value-init must zero-initialise std::array members");
+  __ESBMC_assert(x.bitmap_[0] == 0, "explicit write must not be zeroed");
   return 0;
 }
--- a/regression/esbmc-cpp/cpp/github_4243_mem_init_fail/main.cpp
+++ b/regression/esbmc-cpp/cpp/github_4243_mem_init_fail/main.cpp
@@ -1,0 +1,17 @@
+#include <array>
+#include <cstdint>
+
+struct B
+{
+  std::array<std::uint32_t, 8> bitmap_;
+  B() : bitmap_() {}
+};
+
+int main()
+{
+  B x;
+  __ESBMC_assert(
+    x.bitmap_[0] != 0,
+    "value-init must zero-initialise std::array members");
+  return 0;
+}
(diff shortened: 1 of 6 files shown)


src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2329,7 +2330,22 @@ bool clang_cpp_convertert::get_function_body(
...
         exprt rhs;
         rhs.set("#member_init", 1);
-        if (get_expr(*init->getInit(), rhs))
+
+        /* `m()` in a mem-initializer value-initializes m. For a class type
+         * whose default ctor is trivial (thus not user-provided) this is
+         * exactly zero-initialization, [dcl.init.general]/8; and the implicit
+         * ctor has no GOTO body, so emitting the call would havoc m and leave
+         * it nondeterministic (#4243). */
+        const auto *ctor_expr =
```
lucasccordeiro: Re-implements `zero_initialises()`'s CXXConstructExpr arm (line 635, same file). Call that helper and add the triviality check, so the value-init spellings stay in one place.

regression/esbmc-cpp/cpp/github_4243_mem_init_fail/main.cpp
```diff
@@ -0,0 +1,17 @@
...
+struct B
+{
+  std::array<std::uint32_t, 8> bitmap_;
+  B() : bitmap_() {}
+};
+
+int main()
+{
+  B x;
+  __ESBMC_assert(
```
lucasccordeiro: On unfixed master `bitmap_` is nondeterministic, so this assertion already FAILs; the twin reports FAILED with and without the fix and pins nothing.

src/clang-cpp-frontend/clang_cpp_convert.cpp
```diff
@@ -2329,7 +2330,22 @@ bool clang_cpp_convertert::get_function_body(
...
+          llvm::dyn_cast<clang::CXXConstructExpr>(init->getInit());
+        if (
+          ctor_expr && ctor_expr->requiresZeroInitialization() &&
+          ctor_expr->getConstructor() &&
+          ctor_expr->getConstructor()->isTrivial())
```
lucasccordeiro: No test covers a non-trivial or user-provided default ctor still taking the old path. Drop `isTrivial()` and nothing fails. Add that case.

zacheryasc:
Review addressed:
- call the helper in [src/clang-cpp-frontend/clang_cpp_convert.cpp](https://github.com/esbmc/esbmc/pull/7062/files/4f01f47ead4c6494e4432b5df896e38a58fc33ff#diff-784347884400427356b283749ef36cc6e458d4d4df4f8e3a2a4007244a637b1e)
- deterministic FAIL for the regression test. We now assert that user initializations are not overwritten (pattern from #4183)
- assert we do not overwrite user provided and non-trivial constructions.

PR #7068, opened 2026-08-16
[contracts] Root a multi-level assigns path at its parameter

### LINKED ISSUE #7055: [contracts] a multi-level assigns target generates no VCCs, so a frame violation reports SUCCESSFUL
## Summary

`__ESBMC_assigns` naming a path through two pointers is not classified, so the
assigns-compliance check generates **zero VCCs** for the function and a body
that writes outside its frame reports `VERIFICATION SUCCESSFUL`.

This is a false negative, and the silent kind: the run is indistinguishable
from one where the frame was checked and held.

## Reproducer

`regression/function_contract/assigns_multilevel_ptr_knownbug`, in tree since
the frame work landed and still reproducing on master today.

```c
typedef struct { int a; int b; } Inner;
typedef struct { Inner *sub; int x; } Outer;

void write_sub_a(Outer *o, int v)
{
  __ESBMC_requires(o != (void *)0 && o->sub != (void *)0);
  __ESBMC_assigns(o->sub->a);
  __ESBMC_ensures(1);
  o->sub->a = v;
  o->x = 99; /* not in __ESBMC_assigns: a frame violation */
}

int main() { return 0; }
```

```
$ esbmc main.c --enforce-contract write_sub_a --function write_sub_a --no-pointer-check
VERIFICATION SUCCESSFUL      # wrong: o->x is written and not framed
```

Single-level paths are checked as expected, so the boundary is exactly at the
second dereference.

## Cause

Phase 2C classifies `*ptr` and `ptr->field`. It has no case for
`ptr->sub->field`, and an unclassified target contributes no obligation rather
than an error, so the whole function ends up with nothing to prove.

## Why file it now

The test is marked `KNOWNBUG`, so the suite is green and nothing points at it.
It is also the prerequisite for the Python side: `__ESBMC_assigns` is #6938's
P4.4, and admitting it in Python before this is fixed imports the same class of
silent frame hole into a second frontend.

Two behaviours worth deciding on together:

- an assigns target the classifier does not recognise should be **reported**,
  not dropped. A frame the tool cannot check is not a frame that holds.
- whether multi-level paths are supported, or rejected with a diagnostic, is a
  separate call — but silently generating no VCCs is neither.

LukeW1999:
Fixes #7055.

## Symptom

An `__ESBMC_assigns` clause naming a path through more than one pointer
produced **zero** verification conditions, so a body writing outside its frame
verified — and the output looked exactly like a frame that had been checked and
held.

```c
typedef struct { int a; int b; } Inner;
typedef struct { Inner *sub; int x; } Outer;

void write_sub_a(Outer *o, int v)
{
  __ESBMC_requires(o != (void *)0 && o->sub != (void *)0);
  __ESBMC_assigns(o->sub->a);
  __ESBMC_ensures(1);
  o->sub->a = v;
  o->x = 99; /* not in __ESBMC_assigns: a frame violation */
}
```

```
$ esbmc main.c --enforce-contract write_sub_a --function write_sub_a --no-pointer-check
Generated 0 VCC(s), 0 remaining after simplification
VERIFICATION SUCCESSFUL
```

## Cause

Both halves of the check missed it, which is why nothing was reported at all:

- `classify_assigns_targets` recognises `ptr->field` by matching
  `member(dereference(symbol), f)`. For `o->sub->a` the dereference is applied
  to a member rather than a symbol, so it fell through to `direct_targets`.
- `direct_targets` is matched against the collected snapshots by expression
  equality, and `o->sub->a` is not a snapshot subject, so it matched nothing.
- Meanwhile Phase 2C skips a pointer parameter that appears anywhere in the
  assigns clause, so `o` got no snapshot either.

Instrumented output before the fix:

```
classify: direct=1 ptr=0 ptr_field=0 struct_field=0
ptr_deref_snaps=0
```

## Fix

Find the parameter the path is rooted at and the field it enters — `o->sub->a`
is rooted at `o` through `sub` — and record that in `ptr_field_targets`, so the
per-field check holds every other field of `*o` unchanged. The search recurses,
so three levels root the same way as two.

After:

```
classify: direct=0 ptr=0 ptr_field=1 struct_field=0
Generated 2 VCC(s)
VERIFICATION FAILED
  assigns compliance: ...@write_sub_a@o->x not in assigns clause
```

## Scope, stated rather than implied

Rooting the path at its parameter is weaker than the clause. It catches a write
to any *other* field of the parameter, which is the reported violation, but:

- a write **under** the named field — `o->sub->b` — is still unreported. The
  pointee of a field is not a parameter, so Phase 2C has nothing to root a
  snapshot at.
- recording `sub` as assignable permits reassigning `o->sub` itself, which a
  clause naming only `o->sub->a` does not.

Both are pinned in `github_7055_assigns_multilevel_inner_knownbug` with the
reasons in the source comment.

## Tests

| test | mode | |
|---|---|---|
| `github_7055_assigns_multilevel_fail` | CORE | red-green: 0 VCC / SUCCESSFUL on master, FAILED here, matching `o->x not in assigns clause` |
| `github_7055_assigns_multilevel` | CORE | control: a body writing only what the clause names must still verify |
| `github_7055_assigns_multilevel_three_fail` | CORE | three levels, so the root search has to recurse |
| `github_7055_assigns_multilevel_inner_knownbug` | KNOWNBUG | what the approximation does not reach |

**Mutation check.** The first control I wrote did not bite: a body writing only
`o->sub->a` passes whatever field name is recorded, because it never touches
`o->sub` itself. The discriminator is a body that writes the pointer field on
the path (`o->sub = 0`) — SUCCESSFUL with the fix, FAILED when the recorded
field name is corrupted. Reported here because the first two attempts at this
control were worthless and a reader should not assume otherwise.

`function_contract` 372/372, `loop-invariants` 76/76.


src/goto-programs/frame_enforcer.cpp
```diff
@@ -49,6 +49,32 @@ void frame_enforcert::materialize_snapshots(
...
+// what catches a write to `o->x`. What happens under `o->sub` is not covered --
+// the pointee of a field is not a parameter, so Phase 2C has nothing to root a
+// snapshot at (github_7055_assigns_multilevel_inner_knownbug).
+static bool root_pointer_field(const expr2tc &e, irep_idt &ptr, irep_idt &field)
+{
+  if (!is_member2t(e))
+    return false;
+
+  const member2t &mem = to_member2t(e);
+  if (!is_dereference2t(mem.source_value))
```
lucasccordeiro: No typecast stripping here, unlike every other pointer-chain walk in `contracts.cpp` (e.g. line 3056), so `((Inner *)o->sub)->a` still falls to `direct_targets` and 0 VCCs. Strip typecasts around `mem.source_value` and `under`.

src/goto-programs/frame_enforcer.cpp
```diff
@@ -87,8 +113,17 @@ frame_enforcert::classified_assignst frame_enforcert::classify_assigns_targets(
...
-          result.direct_targets.push_back(target);
+          // A path through more than one pointer, e.g. `o->sub->a`. Record the
+          // parameter it is rooted at and the field it enters, so the per-field
+          // check still holds every other field of that parameter unchanged.
+          // Left in direct_targets it matched no snapshot and no obligation was
+          // generated at all, so a body writing outside its frame verified
+          // (#7055).
+          irep_idt root_ptr, root_field;
+          if (root_pointer_field(target, root_ptr, root_field))
+            result.ptr_field_targets[root_ptr].insert(root_field);
```
lucasccordeiro, line 136: The root need not be a parameter: for a global `Outer *g`, `__ESBMC_assigns(g->sub->a)` still gives 0 VCC / SUCCESSFUL (checked on master), since `materialize_ptr_field_snapshots` only walks parameters. Add that shape to the pinned scope.

regression/function_contract/github_7055_assigns_multilevel_inner_knownbug/main.c
```diff
@@ -0,0 +1,23 @@
...
+typedef struct { int a; int b; } Inner;
+typedef struct { Inner *sub; int x; } Outer;
+
+void write_sub_a(Outer *o, int v)
+{
+  __ESBMC_requires(o != (void *)0 && o->sub != (void *)0);
+  __ESBMC_assigns(o->sub->a);
+  __ESBMC_ensures(1);
+  o->sub->a = v;
+  o->sub->b = 42; /* not in __ESBMC_assigns */
```
lucasccordeiro, line 17: Its `test.desc` matches only `^VERIFICATION FAILED$`, so it would flip on any unrelated failure. Add `^.*o->sub->b not in assigns clause.*$`, as the two `_fail` tests do.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Confirmed. Reproduced with `void *sub` to force a real cast: 0 VCCs and VERIFICATION SUCCESSFUL while the body writes `o->x` outside its frame. Fixed in e1625434ef by stripping typecasts either side of the dereference, matching the pointer-chain walks in `contracts.cpp`. Now 2 VCCs and the write is reported.

Tests: `github_7055_assigns_multilevel_cast_fail` (red before the fix) and `github_7055_assigns_multilevel_cast` as the positive control, so the strip does not widen the frame. Thanks for catching this.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Confirmed. `Outer *g; __ESBMC_assigns(g->sub->a)` with a write to `g->x` gives 0 VCCs / SUCCESSFUL: `materialize_ptr_field_snapshots` iterates `code_type.arguments()`, so a global matches no parameter and the recorded field target matches nothing.

Pinned as `github_7055_assigns_multilevel_global_knownbug` rather than fixed here, since extending the snapshot side to globals is wider than this PR.

regression/function_contract/github_7055_assigns_multilevel_inner_knownbug/main.c
LukeW1999: Agreed. Both KNOWNBUGs now name the message, so an unrelated failure cannot read as the bug being fixed.

One deviation worth flagging: I used `sub->b` rather than `o->sub->b`. The label is built as `ptr + "->" + field` in `emit_ptr_field_assertions`, which is single-level today, so pinning the exact two-level spelling would encode a guess about the shape of the eventual fix. The looser form still discriminates, since the only frame violation in that body is the write to `o->sub->b`.

Issue #7057, opened 2026-08-15
[contracts] __ESBMC_old in a quantifier needs a snapshot of a pointed-to region


LukeW1999:
> **Updated after an attempt, and one correction.** Re-scoping: this is a
> snapshot-mechanism issue, not a lift issue. #7070 records the evidence.

## Why a lift is needed at all

A quantified index cannot be snapshotted one element at a time — the snapshot is
taken once, before the body runs, while `j` still ranges — so
`__ESBMC_old(r[j])` has to become "copy the whole region now, index it by `j`
later". That is `lift_old_over_bound_index`, and it needs an object to copy.

## `int (*r)[N]`

`github_4219_old_in_forall_ptr_to_array_knownbug`:

```
ERROR: Can't construct rvalue reference to array type during dereference
```

**Correction to the original text here:** the array rvalue is *not* what fails.
A directly named global array takes the same route and builds the same
`dereference(typecast(__ESBMC_old_raw(&object), int (*)[N]))`, and verifies.
Instrumented:

| test | `object.id` | `object.type` | result |
|---|---|---|---|
| `github_4219_old_in_forall_array_base` | `symbol` | `array` | SUCCESSFUL |
| `github_4219_old_in_forall_ptr_to_array_knownbug` | `dereference` | `array` | the error above |

They differ in one thing: what `&object` is. For the global it is the address of
a named object; here `object` is `*r`, so it is `&*r` — an address computed
through a pointer, which the dereference layer cannot construct the rvalue for.

Two ways round it were tried; both trade the error for a crash:

| attempt | result |
|---|---|
| `((int *)snap)[j]` | `Assertion !is_scalar_type(expr)` in `dereferencet::dereference_expr_nonscalar` (dereference.cpp:529) |
| `*((int *)snap + j)` | assertion in `assert_arith_2ops_consistency` (irep2_expr.cpp:673) |

## `int r[N]` parameter

`github_4219_old_in_forall_array_param_knownbug`. `bound_array_element` requires
an array-typed base and a decayed parameter is a pointer, so the lift declines
before it starts and the quantified `__ESBMC_old` reaches the frontend
unchanged:

```
cannot model call to `c:@F@__ESBMC_old_raw' on a quantified variable
```

The lift is right to decline. The region to copy is reached through the pointer
and its extent lives in the `__ESBMC_is_fresh` clause, which the lift never
sees.

## The common cause

Both shapes need the snapshot to be taken of a **pointed-to region of stated
[...]

PR #7069, opened 2026-08-16
[contracts] Spare the element an assigns clause named

### LINKED ISSUE #7056: [contracts] an assigns clause naming a global array element is reported as a frame violation
## Summary

A body that writes only what its own `__ESBMC_assigns` clause names is reported
as a frame violation, when the frame target is an element of a **global** array.

The opposite failure to #7055: this one refuses a correct program rather than
accepting an incorrect one, but it makes the element-level frame rule unusable
for globals.

## Reproducer

`regression/function_contract/assigns_global_arr_elem_knownbug`, marked
`KNOWNBUG` and still reproducing on master.

```
$ esbmc main.c --enforce-contract <fn> --function <fn>
VERIFICATION FAILED     # the desc asks for SUCCESSFUL
```

## Cause

Phase 2B handles array-element assigns targets for locals and parameters. For a
global array the frame enforcer snapshots the whole array as one scalar and
emits `ASSERT global == snapshot`, which any write to any element falsifies.
The element the clause named is never distinguished from the rest.

## Note

Filed to give the in-tree `KNOWNBUG` an owner. It sits next to #7055 — both are
gaps in the same classifier — and the two are worth looking at together, since
the fix for one is where the other's case would naturally go.

LukeW1999:
Fixes #7056.

## Symptom

A body that writes only what its own `__ESBMC_assigns` clause names was
reported as a frame violation, when the target is an element of a global array.

```c
int global[10];

void write_global_elem(int i, int v)
{
  __ESBMC_requires(i >= 0 && i < 10);
  __ESBMC_assigns(global[i]);
  __ESBMC_ensures(1);
  global[i] = v;      /* only touches global[i], which IS in assigns */
}
```

```
$ esbmc main.c --enforce-contract write_global_elem --function write_global_elem
Violated property:
  assigns compliance: c:@global not in assigns clause
  global == frame_snap_contract_write_global_elem_0
VERIFICATION FAILED
```

The opposite failure to #7055: this one refuses a correct program rather than
accepting an incorrect one, but it makes the element-level frame rule unusable
for globals.

## Cause

Both element-level mechanisms missed it. Phase 2B recognises an array element
only as `add2t(pointer_symbol, index)` — the pointer arithmetic a *parameter*
decays to. A global array element is `index2t(symbol, index)`, which is neither
that nor a `ptr->field`, so it fell through to `direct_targets`. That is matched
against snapshots by expression equality, and the snapshot subject is the whole
array symbol, so it matched nothing and the coarse whole-array assertion stood.

## Fix

Record the indices a clause names for an array symbol, and hold every *other*
element to its pre-state rather than the array as a whole.

The classification is tested before the pointer case, because an element of an
array *of pointers* is itself pointer-typed: classed as a pointer target, the
whole array stayed asserted unchanged and even a body writing only the named
element failed.

Two encodings hold the rest of the array, chosen by extent:

**Element-wise, up to 256 elements.** One assertion per element, excused at the
named indices. It names the element that broke the frame, which is what a
counterexample is for, but it costs more than linearly: 256 solve in 0.6s, 512
in 2.4s, 1000 in 16s, and 10000 does not finish inside the regression timeout.

**A store over the snapshot, past that.** `global == (snap WITH [i := global[i]])`
— the array equals its pre-state with only the named indices replaced by their
current values. Exact, and one equality whatever the extent: a constant 5 VCCs,
0.2s at both 1000 and 10000.

The store form is deliberately not used everywhere. Three limits, each measured
rather than assumed:

- it reports the array, not the element, so below the cap the element-wise form
  earns its cost
- Bitwuzla rejects equality over constant arrays (`not fully supported yet`,
  then `ERROR: SMT solver failed`), so it cannot serve the loop rule's ASSUME
  mode; the same program answers under `--z3`
- an element that is itself an array fails on every solver — the same
  array-rvalue gap that stops `__ESBMC_old` over an array (#7057)

The struct-field case was already the same idea applied to a different part of
an object, so the two share `emit_partial_frame`, a file-static free function.
That also leaves `enforce_frame_rule` no more complex than it was.

## Tests

| test | mode | |
|---|---|---|
| `assigns_global_arr_elem` | CORE | was `assigns_global_arr_elem_knownbug`, pinned since #3702 when the frame rule landed; 14 VCCs here rather than a skipped check |
| `github_7056_assigns_global_other_elem_fail` | CORE | control: `i >= 1`, so writing `global[0]` is an element the clause does not name |
| `github_7056_assigns_global_other_sym_fail` | CORE | control: naming an element of one global does not excuse a different global |
| `github_7056_assigns_global_2d` | CORE | a row of a two-dimensional global |
| `github_7056_assigns_ptr_array` | CORE | an element of an array of pointers, which the pointer branch used to swallow |
| `github_7056_assigns_ptr_array_fail` | CORE | control: reports `pa[0]`, not `@pa` |
| `github_7056_assigns_large_array` | CORE | 1000 elements, past the cap: the store form |
| `github_7056_assigns_large_array_fail` | CORE | control: the store form still catches a write outside the frame |
| `github_7056_assigns_large_array_wrong_index_fail` | CORE | control: the body writes a *valid* element, just not the one the clause names |
| `github_7056_assigns_large_2d_knownbug` | KNOWNBUG | past the cap, an element that is itself an array keeps the fallback |
| `loop_assigns_array_elem_pass` | CORE | the loop rule's ASSUME path, which no `loop-invariants` test reached |
| `loop_assigns_array_elem_fail` | CORE | its ASSERT half |
| `loop_assigns_moving_index_knownbug` | KNOWNBUG | an index the loop advances is excused one iteration late |
| `loop_assigns_large_array_knownbug` | KNOWNBUG | past the cap in ASSUME mode, where Bitwuzla aborts |

Both KNOWNBUGs for shapes past the cap were checked against the previous commit
and behave identically there, so neither is a regression from this one.

**Mutation checks.** The `_fail` tests do not discriminate the store form on
their own -- the whole-array fallback fails them too -- so there are three:

| mutation | flips |
|---|---|
| store form disabled (fall back past the cap) | `large_array` |
| store form's guard made trivially true | `large_array_fail`, `large_array_wrong_index_fail` |
| `emit_array_elem_frame` stubbed to `return false` | `assigns_global_arr_elem` (the test this PR lifts), and 7 of 10 overall including both loop tests |

The last one is what shows the two loop tests reach the ASSUME path rather than
passing for unrelated reasons; `--goto-functions-only` also shows the eight
element-wise `frame: preserve unassigned variable` ASSUMEs they generate.

`function_contract` and `loop-invariants` 458/458.

## Note

A non-constant extent falls back to the whole-array assertion, i.e. to the old
behaviour. A C global always has a constant extent, so that branch is defensive
rather than reachable from this entry point.


src/goto-programs/frame_enforcer.cpp
```diff
@@ -96,6 +96,17 @@ frame_enforcert::classified_assignst frame_enforcert::classify_assigns_targets(
         result.direct_targets.push_back(target);
       }
     }
+    else if (
+      is_index2t(target) && is_symbol2t(to_index2t(target).source_value) &&
```
lucasccordeiro: The `is_pointer_type(target)` branch above wins for a global array of pointers, so `int *pa[10]; __ESBMC_assigns(pa[i])` still reports the #7056 false positive (verified). Handle it here or pin it as KNOWNBUG.

src/goto-programs/frame_enforcer.cpp
```diff
@@ -130,6 +141,105 @@ static void emit_frame_instruction(
...
+  auto ait = classified.array_elem_targets.find(arr_name);
+  if (ait == classified.array_elem_targets.end())
+    return false;
+
+  const array_type2t &atype = to_array_type(var->type);
+  if (!is_constant_int2t(atype.array_size))
+    return false;
+
+  const BigInt &n = to_constant_int2t(atype.array_size).value;
+  for (BigInt k = 0; k < n; k += 1)
```
lucasccordeiro, line 218: This emits one assertion per element over the whole extent, so `int buf[1000000]` yields a million assertions. Cap the extent and fall back to the whole-array assertion above it.

src/goto-programs/frame_enforcer.cpp
```diff
@@ -130,6 +141,105 @@ static void emit_frame_instruction(
...
+  {
+    expr2tc kc = constant_int2tc(atype.array_size->type, k);
+    expr2tc elem_guard = equality2tc(
+      index2tc(atype.subtype, var, kc), index2tc(atype.subtype, snap, kc));
+
+    for (const expr2tc &assigned : ait->second)
+      elem_guard = or2tc(
+        elem_guard, equality2tc(typecast2tc(assigned->type, kc), assigned));
+
+    emit_frame_instruction(
```
lucasccordeiro, line 228: `enforce_frame_rule` also runs in ASSUME mode for loops, and no `loop-invariants` test names an array element in `__ESBMC_assigns`. Add one covering that path.

src/goto-programs/frame_enforcer.cpp
```diff
@@ -130,6 +141,105 @@ static void emit_frame_instruction(
...
+      dest,
+      loc,
+      field_guard,
+      mode,
+      id2string(sym_name) + "." + id2string(field));
+  }
+  return true;
+}
+
+bool frame_enforcert::emit_partial_frame(
```
lucasccordeiro: `emit_partial_frame` touches no member state yet is public API. Make it a file-static free function like its two callees.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Confirmed, and worse than reported: the positive case fails too. `int *pa[10]; __ESBMC_assigns(pa[i]); pa[i] = v;` -- a body writing only the element the clause names -- gives VERIFICATION FAILED with `c:@pa not in assigns clause`.

Fixed in 8a1646b0cf by testing the index case before the pointer case. That case now verifies, and the violating variant reports `pa[0]` rather than `@pa`. Tests: `github_7056_assigns_ptr_array` and `_fail`.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Confirmed and measured: 256 elements solve in 0.61s, 512 in 2.38s, 1000 in 16.3s, and 10000 does not finish inside the 120s regression timeout. The growth is superlinear, so the cap matters well below a million.

Capped at 256, falling back to the whole-array assertion above it. That stays sound and only loses precision, so the loss is pinned as `github_7056_assigns_large_array_knownbug`: a correct 512-element body is now reported. Mutation-checked -- raising the cap to 1024 flips that test, so it pins the value rather than merely passing.

Lifting the cap properly wants the spared element encoded as a quantifier rather than an assertion per element.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Confirmed: the only `loop-invariants` test naming `__ESBMC_assigns` is `ring_loop_with_contract`, and it names struct fields only.

Added `loop_assigns_array_elem_pass` and `_fail`. I checked they reach the path rather than just passing: `--goto-functions-only` shows 8 element-wise `frame: preserve unassigned variable` ASSUMEs, and stubbing `emit_array_elem_frame` to return false flips both.

Writing them turned up a separate limitation -- naming an index the loop itself advances excuses the element one iteration late, so a correct body is reported. Pinned as `loop_assigns_moving_index_knownbug`.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Done -- now a file-static free function, with the declaration dropped from the header.

src/goto-programs/frame_enforcer.cpp
LukeW1999: Follow-up: the cap is still there, but what sits above it is no longer the whole-array assertion, so the KNOWNBUG I mentioned is gone.

Past the cap the array is now held as a store over its snapshot -- `global == (snap WITH [i := global[i]])` -- which is exact and costs one equality at any extent. 1000 elements go from 16s to 0.2s and 10000 from not finishing to 0.2s, at a constant 5 VCCs. `github_7056_assigns_large_array` is CORE rather than KNOWNBUG now.

It is deliberately not the encoding everywhere, for three reasons I measured rather than assumed:

- it reports the array, not the element that broke the frame, so below the cap the element-wise form still earns its cost
- Bitwuzla rejects equality over constant arrays (`not fully supported yet`, then `ERROR: SMT solver failed`), so it cannot serve the loop rule ASSUME path; the same program answers under `--z3`
- an element that is itself an array fails on every solver -- the same array-rvalue gap that stops `__ESBMC_old` over an array in #7057

Both uncovered shapes are pinned (`github_7056_assigns_large_2d_knownbug`, `loop_assigns_large_array_knownbug`) and I checked each was already the behaviour on the previous commit, so neither is a regression from this one.

Controls: two mutations, since the `_fail` tests alone do not discriminate -- the whole-array fallback fails them too. Disabling the store form flips the positive; making its guard trivially true flips both `_fail` tests, including one where the body writes a valid element that just is not the one the clause names.

PR #7089, opened 2026-08-16
numpy call-result chaining for non-constructor sources

sergillam:
- is_numpy_logical_comparison_call, evaluate_numpy_logical_call (split into evaluate_numpy_compare_call, evaluate_numpy_logical_and_or_call, evaluate_numpy_logical_not_call, evaluate_numpy_where_call), resolve_numpy_logical_arg, numpy_broadcast_binary, numpy_compare_scalar, numpy_logical_as_bool/numpy_logical_as_double, make_bool_constant_node, make_list_node — new helpers in numpy_call_expr.cpp that evaluate a numpy comparison/logical call (greater/less/greater_equal/less_equal/equal/not_equal/logical_and/logical_or/logical_not/where) chained as another numpy call's argument — nested directly or via an intermediate variable — instead of silently substituting the inner call's raw first argument.
- resolve_var and get_arg (inside numpy_call_expr::get()) now call these helpers before falling back to the old behavior.
- The top-level comparison/logical/where dispatch in get() was rewritten to reuse the same helpers, removing duplicated broadcasting/comparison logic.
- Chaining is bounded to 4 levels (max_numpy_logical_chain_depth); beyond that it raises an explicit TypeError instead of silently misreading.

### DIFF
--- a/docs/roadmap/numpy-support-assessment.md
+++ b/docs/roadmap/numpy-support-assessment.md
@@ -1,6 +1,6 @@
 # ESBMC NumPy — Remaining Work
 
-**Updated:** 2026-08-14.
+**Updated:** 2026-08-16.
 
 This file tracks only what is **not yet implemented, broken, risky, or queued
 as backlog** in the NumPy module. If an item is not listed here as a gap, TODO,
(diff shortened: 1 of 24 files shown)


src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -2251,6 +2251,305 @@ static void materialize_inline_numpy_constructor_call(
...
+  python_converter &converter,
+  int depth);
+
+// Resolves one argument of a numpy comparison/logical call: a Name is
+// followed to its declaration; a Call to a recognized constructor is
+// materialized; a Call to another comparison/logical function is evaluated
+// recursively via evaluate_numpy_logical_call(). Anything else (a function
+// parameter, a call to an unrelated numpy function) is left unchanged, so
+// the caller's own numeric/list extraction fails on it exactly as it
+// already does for any other unresolvable operand.
```
lucasccordeiro: This rationale is wrong: `numpy_logical_as_bool`/`numpy_logical_as_double` return `false`/`0.0` for an unresolved node, they do not fail. Raise instead.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -2251,6 +2251,305 @@ static void materialize_inline_numpy_constructor_call(
...
+static nlohmann::json resolve_numpy_logical_arg(
+  nlohmann::json arg,
+  python_converter &converter,
+  int depth)
+{
+  if (arg.is_object() && arg.value("_type", std::string()) == "Name")
+  {
+    const nlohmann::json decl = json_utils::find_var_decl(
```
lucasccordeiro: `find_var_decl` returns the *first* assignment, so `x = np.equal(a,b)` then `x = np.not_equal(a,b)` resolves `x` to the stale value. Reject reassigned names.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -2251,6 +2251,305 @@ static void materialize_inline_numpy_constructor_call(
...
+    resolve_numpy_logical_arg(args[0], converter, depth);
+  const nlohmann::json rhs =
+    resolve_numpy_logical_arg(args[1], converter, depth);
+  return numpy_broadcast_binary(
+    lhs, rhs, [&](const nlohmann::json &a, const nlohmann::json &b) {
+      return numpy_compare_scalar(function, a, b);
+    });
+}
+
+static std::optional<nlohmann::json> evaluate_numpy_logical_and_or_call(
```
lucasccordeiro, line 2484: No test chains a call into `logical_and`/`logical_or`, so this recursion is uncovered. Add e.g. `np.logical_and(np.greater(b, a), np.equal(a, a))`.

src/python-frontend/numpy/numpy_call_expr.cpp
```diff
@@ -5485,127 +5790,81 @@ exprt numpy_call_expr::get()
...
+        arg.value("_type", std::string()) == "Call" &&
+        is_numpy_logical_comparison_call(arg, converter_.ast()))
+      {
+        if (
+          std::optional<nlohmann::json> evaluated =
+            evaluate_numpy_logical_call(arg, converter_, 0))
+          arg = std::move(*evaluated);
+        else
+          throw std::runtime_error(
+            "TypeError: numpy call composition exceeds the supported "
```
lucasccordeiro: Third literal copy of this message (also 2423, 5768), pinned by a test regex. Extract one helper that evaluates-or-throws.

regression/numpy/chaining_depth_exceeded_fail/main.py
```diff
@@ -0,0 +1,16 @@
...
+r = np.logical_not(
+    np.logical_not(
+        np.logical_not(
+            np.logical_not(
+                np.logical_not(np.logical_not(np.equal(a, b)))
+            )
+        )
+    )
+)
+assert r[0] == False
```
lucasccordeiro: numpy gives `r[0] == True` here, so this assertion is wrong. Fix it, otherwise raising the depth bound makes this test fail for the wrong reason.

docs/roadmap/numpy-support-assessment.md
```diff
@@ -51,65 +51,87 @@ Architectural decisions that gate specific pendencies here (referenced as
...
-   resolve a `Name` argument back to a *constructor* call (`zeros`, `full`,
-   `arange`, ...); a `Name` whose declaration is itself a call to another
-   numpy function (`equal`, `greater`, ...) falls through to the old blind
-   `args[0]` read, silently substituting that inner call's first argument
-   for its actual (unevaluated) result. Found while extending
-   `np.greater`/`np.equal` to accept `np.arange(...)` directly as an
-   argument; those two now work — this is a distinct, pre-existing gap in
-   general call composition, not arange-specific.
+
+No known soundness bugs remain open (the numpy call-result chaining gap that
```
lucasccordeiro, line 55: Premature: `resolve_numpy_logical_arg`'s first-assignment lookup still silently mis-evaluates reassigned variables. Qualify this claim.

lucasccordeiro (changes requested):
Six actionable items below. The `where`/broadcast length-mismatch points already raised by Copilot are not repeated.

PR #7102, opened 2026-08-17
[docs] Scope the -only DEAD placement, and retire the hoist as its fix

lucasccordeiro:
**Master-based**, docs and one KNOWNBUG test.

§97's residue left 13 untagged tests, most showing a `DEAD` in a different place. Read properly it is a real defect plus a plausible arm that turns out not to fix it.

**The defect.** For `for (int i = 0; i < 10; ++i) ...`:

| path | placement |
|---|---|
| default | `DECL i`, the loop, `GOTO 1`, then `2: DEAD ...@i` |
| `-only` | `DECL i`, **`DEAD ...@i`**, then the loop guard |

The loop variable is marked dead *before the body that reads and writes it*. **No verdict impact is demonstrated** — an assertion over a value accumulated in such a loop verifies identically on both paths — and I've said so in §98.1 rather than letting "marks a live variable dead" imply more than was shown. It still has to close before the round-trip can be deleted.

**The arm it is not.** `adjust_for` hoists the init into an enclosing block and its own comment calls it *"the only structurally-mutating adjust_\* method"* (fenced on a non-nil init because of #5298). Unported, so it was the obvious candidate. I wrote it and measured it: it fires on 4 sample tests, converges them (`github_4978` 14 → 8 differing lines, corpus-wide 33 767 → 33 743) — and **leaves the misplaced `DEAD` exactly where it was**.

So I am not shipping it. On its own it moves location comments and nothing a test can meaningfully pin, while the defect dominating its own class survives; it belongs in the change that fixes the placement, gated together. The measurements are in §98.2 so the next attempt starts from them instead of repeating the dead end, and §98.3 says where to look (`convert_block`'s destructor unwinding, not the adjuster — the hoist changing block structure without moving the `DEAD` is the useful negative result).

The KNOWNBUG's regex is confirmed to match the **default** path, so it flips to CORE when fixed rather than being vacuously green.

Part of #4715.


lucasccordeiro:
Superseded in part by #7105. §98.2's conclusion here — that porting `adjust_for`'s hoist leaves the misplaced `DEAD` untouched — is **wrong**: the hoist is the fix, and my port had a nesting bug that made it look otherwise (it spliced nothing, so the init's own block closed the scope immediately). #7105 has the corrected arm, −8426 differing lines corpus-wide, and §100 records the correction.

The KNOWNBUG added here passes with #7105, so `testing_tool.py` would exit 77. When both land, this test should become CORE or be dropped in favour of `irep2_only_for_scope` in #7105. Happy to close this PR and fold the surviving material (the DEAD-placement description) into #7105 if you prefer that.

PR #7109, opened 2026-08-17
[python] Check a contract against every input, not one caller

LukeW1999:
Part of #6938. Adds tests, and the frontend change one of them needs.

## What was wrong

None of the 55 tests in `regression/python-contracts` passed `--function`, against the guidance the docs already give:

> Always pair `--enforce-contract func` with `--function func`. Without it, ESBMC follows the call chain from `main`, which limits the range of inputs the function is tested against.

So every enforce test checked its clause against the single argument written in `main`, never against an arbitrary input the `requires` clause admits. On the C side 187 of 405 tests pass `--function`; here it was 0 of 55.

That is not a Python defect -- C behaves identically without the flag, checked side by side -- but it means the suite could not distinguish a contract that holds from one that holds of one caller.

Three positive tests turned out to be in exactly that position. They fail once the parameter is genuinely arbitrary:

| test | postcondition | with `--function` |
|---|---|---|
| `enforce_int` | `x > 0` ⟹ `2 * x > x` | FAILED |
| `enforce_float` | `x > 0.0` ⟹ `x / 2.0 < x` | FAILED |
| `multiple_ensures` | same as `enforce_int` | FAILED |

`2 * x > x` reads as arithmetic but is false in the machine type: `2 * x` overflows a 64-bit int. Bounding `x < 1000` makes it verify, which is what confirms overflow is the cause. Each passed only because `main` calls `double(5)`; changing that to `double(7)` flips the verdict.

## Commit 1 -- the two-tier harness (tests only)

What #6938's own testing plan asked for -- literal repro plus a generalisation constrained only by `requires` -- and was never built.

| test | |
|---|---|
| `enforce_int_entry` | bounded so the postcondition holds of every admitted input |
| `enforce_float_entry` | the float half |
| `old_entry` | `__ESBMC_old` under the entry harness |
| `enforce_int_entry_unbounded_fail` | control: the unbounded claim, refuted |
| `old_entry_caller_value_fail` | control: `old(n) == 3` is true of the call and false of an arbitrary `n` |

The three existing tests keep their flags -- the caller-dependent tier is worth having -- and now carry a comment saying what they do and do not establish, with a pointer to the twin.

**Why the two `_fail` controls are the load-bearing part.** The positives alone do not prove the parameter is havoc'd: `old_entry` and `enforce_float_entry` pass with *and* without `--function`, so they would look identical if the entry harness did nothing. The controls do not: both verify without `--function` and fail with it. That difference is the only direct evidence in the suite that an entry harness makes the input arbitrary.

## Commit 2 -- let `--function` name a method

Writing the tier above surfaced a gap rather than pinning it: `--function` could not name a method under any spelling (`add`, `Counter.add`, `Counter@add` all gave `Function not found`), because it searched only the module body. A static method's signature is wholly scalar and so already inside the v1 contract scope, which left it checkable against one caller and nothing else.

`find_entry_function` now resolves a target through the class bodies as well. A module-level function still wins, `Class.method` and an unambiguous bare name resolve a method, and a bare name several classes define is refused rather than guessed at.

| test | |
|---|---|
| `staticmethod_entry` | a static method as the entry point |
| `staticmethod_entry_qualified` | the same target spelled `Counter.add` |
| `staticmethod_entry_caller_value_fail` | control: true of the call, false of an arbitrary input |
| `staticmethod_entry_ambiguous_fail` | control: two classes define the name, so it is refused |
| `method_entry_instance_fail` | an instance method is refused, not harnessed |

## Commit 3 -- decide the receiver by decorator, not by name (review round)

Commit 2 refused an instance method by looking for a parameter named `self`. Python does not fix that name, so the rule was wrong in both directions:

| input | before | after |
|---|---|---|
| `def add(this, k)` -- an instance method | harnessed over a nondet receiver, **VERIFICATION SUCCESSFUL** | refused |
| `@classmethod def add(klass, k)` | harnessed | refused |
| `@staticmethod def add(self: int)` -- legal | refused as an instance method | admitted |

The false accept is the serious half: it printed a proof for a harness built over a receiver no constructor made. The rule is now the builtin `@staticmethod` decorator, which also covers `@classmethod`, whose `cls` is a class object and equally unbuildable.

Admitting `@staticmethod def add(self: int)` then reaches a `namespacet::follow` assertion, because the frontend binds `self` and `cls` to the enclosing class wherever the names appear, independently of this path. That is reported rather than asserted on.

| test | |
|---|---|
| `method_entry_nonself_receiver_fail` | the instance method the name-based rule bypassed |
| `classmethod_entry_fail` | receiver named `klass`, so the test reaches the new branch |
| `staticmethod_entry_receiver_name_fail` | the converse direction, and the frontend limitation behind it |

Mutation-checked: reverting the discriminator to the parameter name flips all three; dropping the receiver-name guard flips exactly the third.

## Still out of scope

An instance method remains refused under `--function`, so its contract is checkable against its callers only (`method_contract`). #6938 marks P3.3 done, which is true of the caller-dependent half plus static methods.

`regression/python-contracts` 68/68, and the 14 `regression/python` tests that pass `--function` 14/14.


src/python-frontend/python_converter.cpp
```diff
@@ -507,6 +507,103 @@ void python_converter::convert_extra_translation_unit(
...
+  return owners;
+}
+
+static bool takes_self(const nlohmann::json &node)
+{
+  if (!node.contains("args") || !node["args"].contains("args"))
+    return false;
+
+  for (const auto &arg : node["args"]["args"])
+    if (arg.contains("arg") && arg["arg"] == "self")
```
lucasccordeiro: Keying on the parameter name misses instance methods: `def bump(this, n)` is harnessed with a nondet `this`, while the same method spelled `self` is refused. Decide on the @staticmethod/@classmethod decorator instead.

src/python-frontend/python_converter.cpp
```diff
@@ -507,6 +507,103 @@ void python_converter::convert_extra_translation_unit(
...
+    if (method.empty())
+      continue;
+
+    node = method;
+    owners.push_back(element["name"].get<std::string>());
+  }
+  return owners;
+}
+
+static bool takes_self(const nlohmann::json &node)
```
lucasccordeiro: No test covers an instance method whose first parameter is not named `self`, which is the case that bypasses this guard. Add one.

lucasccordeiro (changes requested):
The description says "Test-only", but this adds `find_entry_function` and method resolution for `--function`. Its "Known gap pinned rather than fixed" section also states `--function` will not name a method, which this PR implements. Please update it.

LukeW1999:
Thanks -- all three were right, and the first one was worse than it reads.

**Decorator instead of parameter name.** Reproduced both directions before changing anything:

| input | before | after |
|---|---|---|
| `def add(this, k)` -- a real instance method | harnessed over a nondet receiver, **VERIFICATION SUCCESSFUL** | refused |
| `@classmethod def add(klass, k)` | harnessed | refused |
| `@staticmethod def add(self: int)` -- legal Python | refused as an instance method | admitted |

The false accept is the serious half: it printed a proof for a harness built over a receiver no constructor made. The rule is now the builtin `@staticmethod` decorator (bare `Name` only, so a dotted `x.staticmethod` does not count), which also covers `@classmethod` -- `cls` is a class object and equally unbuildable.

One thing that fell out of admitting the third row: `@staticmethod def add(self: int)` then reaches a `namespacet::follow` assertion, because the frontend binds `self` and `cls` to the enclosing class wherever those names appear, independently of this path. Rather than trade a wrong refusal for an assertion failure I report that limitation, and pin it.

**Tests.** Added the case you named plus the two neighbours:

- `method_entry_nonself_receiver_fail` -- `def add(this, k)`, the instance method the name rule bypassed
- `classmethod_entry_fail` -- receiver named `klass`, not `cls`, so the test actually reaches the new branch
- `staticmethod_entry_receiver_name_fail` -- the converse direction and the frontend limitation behind it

`classmethod_entry_fail` was worth a second pass: written with the conventional `cls` it still passed under the old name-based rule, so it proved nothing. Mutation-checked all three -- reverting the discriminator to the parameter name flips all three, dropping the receiver-name guard flips exactly the third.

**Description.** Rewritten. It now says what each of the three commits does, drops the "Test-only" claim, and the old "Known gap pinned rather than fixed" section is gone -- the remaining scope note is that an *instance* method is still refused, which is accurate.

`regression/python-contracts` 68/68, and the 14 `regression/python` tests that pass `--function` 14/14.

PR #7110, opened 2026-08-17
[c2goto] Register bundled models through the filesystem singleton

Yiannis128:
Master does not compile. #6766 added five `ESBMC_FLAIL` blocks in `register_bundled_libc()` that call `fs.add_bundled(...)`, but `fs` is never declared; the three pre-existing blocks spell out `file_operations::filesystemt::get()` instead. Those five now use the same fully-qualified call.

Binding a local `filesystemt &fs` and using it from every block was the first attempt and reads better, but it does not survive static analysis: the flail headers are generated at build time, so an analyser that cannot resolve `#include <libmusl.h>` never sees the macro bodies that use the reference and reports it unused. Codacy flagged exactly that. Spelling the singleton out keeps the analysers quiet and matches what the file already did.

Verified with a clean `ninja -C build esbmc` (Python frontend on) on a tree that failed to compile before the change.

### DIFF
--- a/src/c2goto/cprover_libc_sources.cpp
+++ b/src/c2goto/cprover_libc_sources.cpp
@@ -61,23 +61,27 @@ void register_bundled_libc()
 #undef ESBMC_FLAIL
 
 #define ESBMC_FLAIL(body, size, ...)                                           \
-  fs.add_bundled(vfs_musl + "/" #__VA_ARGS__, body, size);
+  file_operations::filesystemt::get().add_bundled(                             \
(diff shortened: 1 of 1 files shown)


Yiannis128:
Superseded by #7123, which landed the identical fix: every `ESBMC_FLAIL` block now spells out `file_operations::filesystemt::get()`. `git diff origin/master fix/c2goto-undeclared-fs` on this file is empty, so there is nothing left here to merge. Closing.

Issue #6483, opened 2026-07-28
[contracts] --enforce-contract misses an ensures violation on an is_fresh struct param with two __ESBMC_old field snapshots


Yiannis128:
`--enforce-contract` reports VERIFICATION SUCCESSFUL on a function that violates its `ensures` clause, when the pointer parameter is heap-backed via `__ESBMC_is_fresh` and two or more `ensures` clauses use `__ESBMC_old` on different fields of it.

## Minimal reproducer

```c
typedef struct { int val; int count; } Counter;

void f(Counter *c) {
  __ESBMC_requires(__ESBMC_is_fresh(c, sizeof(Counter)));
  __ESBMC_ensures(c->val   == __ESBMC_old(c->val) + 1);
  __ESBMC_ensures(c->count == __ESBMC_old(c->count) + 1);
  c->val++;
  c->count += 2;   /* violates the second ensures */
}

int main() { return 0; }
```

```
esbmc repro.c --function f --enforce-contract f
```

`VERIFICATION SUCCESSFUL` on 8.3.0 and on current master. The body increments `count` by 2 where the contract promises 1, so the expected result is `VERIFICATION FAILED`.

## Three conditions are jointly required

Removing any single one of these makes ESBMC report `VERIFICATION FAILED` correctly, so all three are needed to trigger it:

| variant | verdict |
| --- | --- |
| as above | VERIFICATION SUCCESSFUL (wrong) |
| drop the `is_fresh`, so the param is stack-backed | VERIFICATION FAILED |
| keep only the `count` ensures clause | VERIFICATION FAILED |
| keep both ensures but write only `c->count` | VERIFICATION FAILED |

The stack-backed row is the important control: identical contract, identical body, and the only difference is whether the parameter's storage came from `__ESBMC_is_fresh` (heap) or from the harness's struct stack backing.

The same shape with a scalar `int *` parameter under `is_fresh` reports `VERIFICATION FAILED` correctly, so it is not simply "`__ESBMC_old` through a heap pointer". It needs two `old` snapshots of distinct fields of the same heap object plus writes to both.

## The GOTO program and the VCC are both correct

This is not the GOTO-level conjunct dropping of #6298. `--goto-functions-only` shows a well-formed harness, with both snapshots taken before the call and both conjuncts asserted after:

```
ASSIGN c=MALLOC(unsigned char, nil);
ASSUME c != 0                                    // __ESBMC_is_fresh: pointer is non-null
[...]

PR #7116, opened 2026-08-17
[contracts] Give a struct pointer parameter a nondet extent

Yiannis128:
Closes the struct half of #6212. #6484 gave pointer parameters a nondet extent but exempted structs, because a heap-backed struct then broke `__ESBMC_old` ensures clauses (#6483); that is fixed and closed, so the carve-out goes. A contract that dereferences a struct pointer must now state `__ESBMC_requires(__ESBMC_is_fresh(p, sizeof(T)))` or it reports `VERIFICATION FAILED`. The affected tests were migrated accordingly, `_fail` twins included, and each now pins the property it was written to test rather than only `VERIFICATION FAILED`.

Two things the carve-out was hiding come with it. First, a parameter with a nondet extent was skipped by the assigns-clause frame check entirely, on the grounds that reading `*p` against that extent is itself out of bounds. That left the frame to the bounds check alone, so `--no-bounds-check` verified an out-of-frame write. The snapshot and its assertion are now taken under the condition that the extent covers the pointee, which is sound either way and closes the same hole for scalar pointers, where it has been open since #6484. Second, a C++ reference parameter is a pointer with a struct pointee, so it would have picked up a nondet extent with no way to state one: `__ESBMC_is_fresh` needs a pointer it can point at fresh storage, and a reference gives the contract none to name. References keep the whole-object backing alongside the implicit receiver, since C++ guarantees both address one complete object.

`function_contract` is the affected suite; CI runs it.


LukeW1999 (commented):
The direction is right and #6483 really is out of the way — I re-ran the
heap-backed-struct-with-`__ESBMC_old` shape and the false negative does not
recur, and `github_6483_old_two_fields_fail` survives as a live control. But I
do not think this can land as it stands: removing the carve-out takes a frame
check with it, and it breaks C++ reference parameters with no migration
available.

Everything below is a measured A/B on `a499f7425d`: build the branch, run the
command, then restore `if (is_structure_type(pointee))` at
`contracts.cpp:5339`, rebuild, run the same command.

## 1. The assigns-clause frame check disappears for struct parameters

Struct params now take `emit_pointer_param_malloc` with `justified = false`,
and `materialize_ptr_deref_snapshots` (`contracts.cpp:3155`) skips every
`!justified` param — so no pre-state snapshot and no `assigns compliance`
assertion is emitted. With the default checks on, the bounds check hides this.
With them off it does not.

`regression/function_contract/enforce_ptr_param_array_field_fail/main.c`, whose
body writes `b->coeffs[0]` and whose clause is `__ESBMC_assigns(r->coeffs)`:

```
$ esbmc main.c --enforce-contract add --function add --unwind 9 \
      --no-pointer-check --no-bounds-check

carve-out restored:  FAILED  assigns compliance: ...@add@b->coeffs[k]
                             (k: nondet witness index) not in assigns clause
                     VERIFICATION FAILED

this branch:         VERIFICATION SUCCESSFUL
```

An out-of-frame write verifies. Those two flags are not exotic —
`regression/function_contract/conc_mutex_pass/test.desc` already runs
`--enforce-contract` with exactly them.

The hole itself predates this PR: `justified` has gated the frame check for
scalar pointers since #6484. What this PR changes is that struct pointers now
fall into it, and `__ESBMC_assigns(p->field)` is where the clauses actually
live. Worth deciding whether `justified` should gate the frame check at all,
rather than only whether the extent is assumed — they are different questions
and the flag currently answers both.

## 2. C++ reference parameters break, and the extent cannot be stated

`void f(S &s)` is a pointer parameter with a struct pointee and base name `s`,
so it now gets a nondet extent:

```c
struct S { int n; };
void f(S &s)
{
  __ESBMC_requires(s.n >= 0);
  __ESBMC_assigns(s.n);
  __ESBMC_ensures(s.n == 1);
  s.n = 1;
}
```

```
carve-out restored:  VERIFICATION SUCCESSFUL
this branch:         VERIFICATION FAILED
```

And there is no migration for it. The obvious one is rejected outright:

```
$ esbmc ref.cpp --enforce-contract f --function f
   /* __ESBMC_requires(__ESBMC_is_fresh(&s, sizeof(S))); */
ERROR: __ESBMC_is_fresh needs a pointer it can point at fresh storage, but the
contract of 'f' gives it the address of a non-pointer object.
```

The rationale in the code — the language guarantees one complete object, so
this is not an extent the contract failed to state — applies to a reference
word for word. The code applies it only to `this`. Either widen the exemption
to reference-typed parameters, or give references a way to state an extent,
before this lands. No regression test covers a reference parameter, which is
why the suite stays green.

## 3. Migrating a `_pass` test without its `_fail` twin under-determines the twin

`enforce_ptr_param_array_field_fail`'s own header says it exists to pin that
the Phase 2C frame check catches the write — "Confirms the fix is sound, not
merely crash-avoiding". On this branch, with default flags:

```
Violated property: line 15  dereference failure: array bounds violated: heap object
```

Not `assigns compliance`. Its `test.desc` greps only `^VERIFICATION FAILED$`,
so it stays green while no longer testing what it was written to test. Same for
`assume_nonnull_valid_two_ptrs_fail`, which now reports a bounds violation
where it used to report `contract ensures`.

Ten more `_fail` tests are in the same position — `assume_nonnull_valid_fail`,
`assume_nonnull_valid_old_fail`, `assume_nonnull_valid_req_content_fail`,
`cpp_wildcard_enforce_struct_ptr_fail`, `mem_struct_init_fail`,
`or_ensures_flag_fail`, `or_ensures_two_branches_fail`,
`struct_nested_field_fail`, `struct_old_field_fail`,
`struct_return_ptr_field_fail`. They still report the intended property on this
build, but each now has two independent reachable counterexamples and none of
their `test.desc` files pins a reason, so which one is reported is the solver's
choice. Migrating the twin, and adding a reason regex, closes both halves.

## 4. Two migrations give up aliasing coverage they did not need to

`enforce_ptr_param_array_field_pass` is distilled from `mlk_poly_add(r, b)` and
used to exercise `r == b` through the frame check. Unconditional
`__ESBMC_is_fresh` on both puts them in `separated_params`, so
`emit_pointer_param_aliasing` emits nothing — `may alias` goes from present to
zero occurrences. The guarded form this PR itself introduces for the
`github_6551_*` tests (`r == 0 || __ESBMC_is_fresh(r, sizeof(poly))`) still
gives `VERIFICATION SUCCESSFUL` here, so the coverage was given up for nothing.
`struct_return_ptr_field_pass` is the same shape.

## 5. The receiver exemption has a positive test but no negative controls

`github_6212_receiver_extent_pass` pins that the exemption applies. Nothing
pins its two premises: that a C-mode parameter named `this` does *not* get it —
delete `func.mode == "C++"` and no test fails — and that an access beyond the
receiver's one element is still caught. The C path has exactly that companion
in `github_6212_struct_extent_beyond_fail`; the receiver path does not. A
`this[1].n = ...` case and a C `struct S *this` case would close both.

For what it is worth I checked the identification itself and it is solid:
`clang_cpp_convert.cpp:2609` sets the implicit receiver's base name to `"this"`
unconditionally, and `this` is a reserved word in C++, so nothing else can
carry that name in that mode.

## 6. Staleness

115 commits behind master, and `contracts.cpp`/`contracts.h` conflict with
#7217 (`arr_elem_snapshot_t::declared_indices`). The conflicts are contextual —
master did not touch `add_pointer_validity_assumptions` or the `param_extentt`
producers — so it still applies semantically, but no number in the description
survives a rebase untested. Two other things in the body are stale: "will not
build in CI until #7110 lands" — #7110 is closed unmerged, #7111 landed
instead — and `403/404`, which is now `404/404` on the branch's own base, so
the description understates the result.

## Checked and cleared

The removed #6483 warning is safe to drop: I mutation-tested the heap-backed
struct plus `__ESBMC_old` shape directly and it still reports `contract
ensures`. An injected `__ESBMC_ensures(0)` probe over all 25 migrated `_pass`
tests reports `VERIFICATION FAILED` for every one, so none of them passes
vacuously. The `github_6551_*` guarded-`is_fresh` migrations do keep their
aliasing.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3220,6 +3202,16 @@ code_contractst::materialize_ptr_deref_snapshots(
...
       continue;
 
+    // A nondet extent does not cover the pointee on every path, so the reads
+    // below are taken under it rather than dropped: the body cannot
+    // dereference p where it does not hold either. Declared ahead of the skip
+    // checks below, which jump past this point.
+    auto extent_it = param_extents.find(param_id);
+    expr2tc readable;
+    if (extent_it != param_extents.end() && !extent_it->second.justified)
+      readable = greaterthanequal2tc(
```
lucasccordeiro, line 3212: `readable` uses the parameter's own extent symbol, but `emit_pointer_param_aliasing` runs earlier and may repoint it at another parameter's allocation. Derive the guard from the object actually addressed.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -5442,16 +5421,24 @@ void code_contractst::add_pointer_validity_assumptions(
...
+    // address one complete object, so one element is what the language already
+    // promises rather than an extent the contract failed to state.
+    // __ESBMC_is_fresh cannot state it either way: a reference gives the
+    // contract no pointer to name, and on `this` it would additionally assert
+    // separation from every other pointer parameter, which no contract states.
+    // `this` needs the mode test because it is a reserved word only in C++; a
+    // C parameter of that name carries no such guarantee. A function reference
+    // has no object to declare, so it keeps the pointer path.
+    if (
+      (is_reference(param.type()) ||
```
lucasccordeiro, line 5434: `is_reference` is false for `S&&`, which carries `#rvalue_reference` instead, so rvalue-reference parameters still get a nondet extent they cannot state. Use `is_lvalue_or_rvalue_reference`.

src/goto-programs/contracts/contracts.h
```diff
@@ -440,22 +442,20 @@ class code_contractst
...
-  /// \param wrapper GOTO program to append snapshot instructions to
-  /// \param location Source location
-  /// \param func_name Function name for unique snapshot naming
-  /// \param param_extents Byte extent of each harness allocation. Params whose
-  ///        backing is not justified are skipped: the snapshot dereferences
-  ///        the pointer, and against a nondet extent that harness-invented
-  ///        read fails its own bounds check, reporting a violation in a
-  ///        parameter the contract never mentions.
+  /// \param classified Classified assigns targets (provides pointer_targets,
+  /// ptr_field_targets) \param assigns_targets Full assigns target list (must
```
lucasccordeiro, line 452: Accepting clang-format's fixed point merged the `\param` tags into one paragraph, so Doxygen stops rendering them as parameters. Shorten each `\param` line under 80 columns instead.

regression/function_contract/github_6212_nondet_extent_frame_fail/main.c
```diff
@@ -0,0 +1,26 @@
...
+ * out-of-frame write. The snapshot and its assertion are now taken under the
+ * condition that the extent covers the pointee, which is sound whether or not
+ * bounds checking is on.
+ */
+typedef struct
+{
+  int x;
+} S;
+
+void f(S *r, S *b)
```
lucasccordeiro, line 14: Both new frame tests use a struct pointee, so the scalar-pointee `read_where` branch the description claims closes #6484's hole is unpinned. Add a scalar-pointer twin.

regression/function_contract/assume_nonnull_valid_no_null_req_pass/main.c
```diff
@@ -12,6 +12,7 @@ typedef struct { int x; } S;
 
 void f(S *p)
 {
+  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(S)));
```
lucasccordeiro, line 15: `__ESBMC_is_fresh` emits its own `ASSUME(p != NULL)`, so this no longer pins that the harness supplies non-null without a `requires`. Update the header comment or the premise.

PR #7117, opened 2026-08-17
[python] Extend dynamic typing to isinstance()

brcfarias:
Follow-up to #7018 (dynamic typing for scalar variables whose type diverges across an if/else branch).
Extends `isinstance()` to work on those scalar variables, which previously threw "not yet supported".


src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -538,6 +549,28 @@ exprt dynamic_type_handler::handle_arithmetic(
...
+  if (type_name == "int")
+    return type_handler_.tagged_scalar_type_matches(
+      tagged_type_id, long_long_int_type());
+
+  if (type_name == "str")
+    return build_equal(
+      tagged_type_id,
+      type_handler_.tagged_scalar_type_id(pointer_typet(char_type())));
+
+  return false_exprt();
```
lucasccordeiro: `isinstance(x, object)` now returns hard false, but `object` resolves to `any_type()` and must be true. Throw on unhandled names instead of silently answering false.

regression/python/dynamic_type_scalar_tag_isinstance_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.py
+
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: Three asserts, but only `isinstance(y, bool)` should fail; this regex still passes if a different assert breaks instead. Pin the failing line.

src/c2goto/cprover_libc_sources.cpp
```diff
@@ -60,6 +60,8 @@ void register_bundled_libc()
 #include <libm.h>
 #undef ESBMC_FLAIL
 
+  auto &fs = file_operations::filesystemt::get();
```
lucasccordeiro: Unrelated master build fix, duplicated by #7123. Drop it here and rebase once that merges, to avoid a conflict.

PR #7142, opened 2026-08-18
[symex] Reuse claims proved in an earlier run via --vcc-cache

lucasccordeiro:
Adds an opt-in cross-run cache of discharged claims: `--vcc-cache <dir>` keys
each claim on the canonical text of its sliced SSA cone plus the ESBMC build,
every option in effect and the data model, and skips the solver when a matching
proof is on disk. Only UNSAT is stored, and the stored text is compared in full
on a hit, so a digest collision degrades to a re-solve rather than a wrong
verdict. `--vcc-cache-verify` solves everything anyway and flags disagreement.

Measured on `regression/esbmc`: after editing one function, 350/385 claims in
the other functions are reused (90.9%), verdicts unchanged. `--vcc-fingerprint-dump`
is the instrument that measured it and is kept for future tuning.


src/esbmc/bmc.cpp
```diff
@@ -2448,6 +2450,43 @@ int bmct::ltl_run_thread(symex_target_equationt &equation)
...
+    ssa_cone_size(cone),
+    verdict,
+    loc,
+    msg);
+
+  static std::mutex dump_mutex;
+  std::lock_guard<std::mutex> lock(dump_mutex);
+  if (path == "-")
+  {
+    std::cout << "VCC-FP " << line << "\n";
```
rafaelsamenezes: Avoid cout

src/esbmc/bmc.cpp
```diff
@@ -2526,6 +2565,27 @@ smt_resultt bmct::multi_property_check(
...
   bool is_keep_verified = options.get_bool_option("keep-verified-claims");
+  const std::string fingerprint_dump =
+    options.get_option("vcc-fingerprint-dump");
+
+  // Cross-run cache of discharged claims. Held to the modes where a claim's
+  // sliced cone is the whole of what the verdict depends on: the k-induction
+  // phases re-use one claim across base/forward/inductive step, a coverage or
+  // dead-code probe is not a property, and past the first interleaving a claim
+  // carries a schedule the cone does not name. These are the exclusions the
+  // in-run assertion_cache already makes (see the bmct constructor).
```
rafaelsamenezes: Rambling

rafaelsamenezes (changes requested):

lucasccordeiro:
Superseded by #7143, which raises the same branch from `esbmc/esbmc` rather than from a fork. Same commits, including the fixes for the review comments left here.

PR #7143, opened 2026-08-18
[symex] Reuse claims proved in an earlier run via --vcc-cache

lucasccordeiro:
Adds an opt-in cross-run cache of discharged claims: `--proof-cache <dir>` keys
a claim on a 128-bit hash of its sliced SSA cone, the build, every option that
changes what is verified, and the data model, then skips the solver on a hit.
Only UNSAT is stored; `--proof-cache-verify` solves anyway and flags mismatches.


lucasccordeiro:
SV-COMP (30s): https://github.com/esbmc/esbmc/actions/runs/32179757667.

lucasccordeiro:
## Measurement report

Independent A/B to decide whether this is worth merging. Two git worktrees, identical
`RelWithDebInfo` configs, Bitwuzla, macOS arm64 (M-series, 14 cores), runs strictly serial:

- **base** = `db6929f438` — the merge-base, i.e. master without this PR
- **pr** = `671773f250` — this branch

Every scenario is the median of 3 repeats, and the repeats are **interleaved** (each repeat runs
base, cache-off, cold, warm in the same order) so drift and transient load hit all scenarios alike.
Harness and raw JSON available on request.

### 1. Soundness

**672 regression files, 5 470 properties, 94 153 claims, 7 scenarios each — 0 verdict
disagreements.** No scenario (cache off, cold, warm, or after any of three edit kinds) reaches a
different verdict from master on any file.

### 2. Cost when the cache is off

| | wall | vs base |
|---|---|---|
| baseline (master) | 265.08s | |
| PR, no `--vcc-cache` | 265.27s | **+0.07%** |

Per-file distribution: p05 −4.0%, p50 −1.0%, p95 +1.7%. The default path is unaffected.

### 3. Cost and gain on the regression suite

| scenario | wall | vs base | solver | vs base |
|---|---|---|---|---|
| baseline | 265.08s | | 97.75s | |
| cold cache | 286.71s | +8.16% | 91.25s | −6.64% |
| warm cache | 209.77s | **−20.86%** | 37.99s | −61.13% |
| edit: comment at top | 170.43s | −22.62% | 38.02s | −59.82% |
| edit: statement into `main` | 174.45s | −19.80% | 42.29s | −54.53% |
| edit: statement into another function | 115.03s | −21.25% | 36.95s | −49.87% |

Cold overhead is skewed, not uniform: median **+0.5%**, p75 +2.4%, p95 +11.1%, worst +95.7%. It is
paid per claim — about 0.07 ms for a typical file, 0.5 ms at p75, 1.6 ms at p95 — so files with
many cheap claims pay most and gain least.

On the 29 files where the solver is ≥0.5s of the run: cold +14.61%, warm **−34.84%**.

### 4. Claim reuse (exact counts, not timings)

| scenario | claims reused | files 100% reused |
|---|---|---|
| cold (intra-run collapse only) | 1 492 / 6 931 = 21.5% | 0 / 529 |
| warm, no edit | 6 487 / 6 931 = **93.6%** | 233 / 529 |
| comment inserted at top of file | 5 822 / 6 931 = 84.0% | 223 / 529 |
| statement inserted into another function | 3 421 / 4 029 = **84.9%** | 81 / 166 |
| statement inserted into `main` | 5 519 / 6 928 = 79.7% | 203 / 525 |

Split by verdict, warm: **SUCCESSFUL 100.0%**, FAILED 85.2%, UNKNOWN 89.7% — the shortfall is
by design, since a refuted claim is never stored.

25 329 cache entries occupied **0 bytes** (entries are empty files; the name is the key).

### 5. Where it actually pays

The suite understates the feature badly: its **median file spends 0.9% of its run in the solver**,
and only 43 of 671 spend more than half. On solver-bound modules (3 hand-written modules, 43/41/166
claims, 97% solver time on the first):

| module | strategy | base | warm | speed-up |
|---|---|---|---|---|
| bits | `--unwind 20` | 6.00s | 0.27s | **22x** |
| bits | `--incremental-bmc` | 6.10s | 0.31s | 20x |
| math | `--unwind 20` | 2.70s | 0.10s | 27x |
| math | `--incremental-bmc` | 8.38s | 2.14s | 3.9x |
| math | `--k-induction` | 1.47s | 0.71s | 2.1x |
| array | `--unwind 20` | 1.46s | 0.30s | 4.9x |
| array | `--incremental-bmc` | 0.81s | 0.28s | 2.9x |

**Edit-and-recheck sequence** — a 3-module project re-verified after each of six commits, each
editing the body of one function, cache persisted across commits:

| commit | baseline | cached | speed-up |
|---|---|---|---|
| 0 (cold) | 10.55s | 10.66s | 1.0x |
| 1 | 10.21s | 2.90s | 3.5x |
| 2 | 10.22s | 0.68s | 15.0x |
| 3 | 10.28s | 0.67s | 15.2x |
| 4 | 10.30s | 3.67s | 2.8x |
| 5 | 10.33s | 3.16s | 3.3x |
| 6 | 9.95s | 1.53s | 6.5x |
| **total** | **71.85s** | **23.27s** | **3.1x** |

77.9% of claims reused across the sequence, every verdict SUCCESSFUL.

### 6. Two corrections this turned up

- The PR body and the docs both said the stored cone text is compared in full on a hit, so a
  collision degrades to a re-solve. That stopped being true at `2dfef0f0d0`: entries are empty
  files named `<16 hex context><32 hex cone>.vcc`, so the digests **are** the comparison and the
  64-bit context half is the narrower margin. Both are now corrected.
- The docs' "around 30% of the run" cost had the dependency backwards. Cost scales with claim
  count, benefit with solver share, so solver-heavy files absorb it best, not worst.

### Recommendation

Worth merging. It is opt-in and free when off, no verdict moved across ~4 700 runs, the failure
mode is a re-solve rather than a wrong answer, and `--vcc-cache-verify` gives an audit path. The
value is narrow — the edit-recheck loop on solver-bound code — and should not be enabled for
one-shot or competition runs, which the docs now say explicitly.

One gap: I have not run a full clean regression suite on this branch. My sweep covers 672 files
with cache flags, which exercises the new path well but is not the same thing.

src/esbmc/bmc.cpp
```diff
@@ -2450,6 +2452,152 @@ int bmct::ltl_run_thread(symex_target_equationt &equation)
   return ltl_res_good;
 }
 
+/// A cached proof is only sound where the claim's sliced cone is everything its
+/// verdict depends on. The k-induction phases and --incremental-bmc qualify:
+/// each sets its phase flags and unwind on this optionst before running, so
+/// both are already part of a claim's key.
+static bool vcc_cache_usable(
```
Yiannis128: The cache is silently inert without --multi-property: it is only wired into multi_property_check, so `esbmc file.c --vcc-cache dir` runs to completion with no hint anything was skipped. The same applies to --vcc-cache-verify given without --vcc-cache: the directory is empty, make_vcc_cache returns null, and the flag evaporates. Please warn or hard-error in both cases.

src/esbmc/bmc.cpp
```diff
@@ -2450,6 +2452,152 @@ int bmct::ltl_run_thread(symex_target_equationt &equation)
...
+{
+  if (!vcc_cache_usable(
+        options, dir, is_cov_run, is_dead_code, interleaving_number))
+    return nullptr;
+  return std::make_unique<vcc_cachet>(dir, options);
+}
+
+/// A warm run solves nothing, so without this the report shows no solver
+/// activity at all and gives no sign of where the verdicts came from.
+static void log_vcc_cache_summary(const vcc_cachet *cache)
```
Yiannis128: Deactivation is also silent at runtime: under coverage modes, --ltl or past the first interleaving the flag was accepted but nothing tells the user the cache never engaged. A one-line status like "VCC cache: inactive (<reason>)" when a directory was named but the cache is disabled would save users a confusing session of unexplained zero hits.

src/util/ssa/vcc_cache.cpp
```diff
@@ -0,0 +1,105 @@
...
+std::string vcc_cache_context(const optionst &options)
+{
+  std::ostringstream ctx;
+  ctx << "esbmc " << ESBMC_VERSION << '\n';
+
+  // Every option is folded in unabridged except the cache's own controls and
+  // the source path, none of which change what is verified -- the path is
+  // stripped from symbol names before a cone is digested. Curating a wider
+  // exclusion list is how a cache like this goes unsound.
+  static const std::set<std::string> not_semantic = {
```
Yiannis128: The unsound-curation argument cuts both ways: folding every option unabridged means purely cosmetic flags (--verbosity, --json-ui, --cex-output) bust every key, and users get inexplicably cold caches they cannot connect to any cause. Could output-only options join not_semantic? The list is already curated in this direction; this just widens it to flags that provably do not change what is verified.

src/esbmc/options.cpp
```diff
@@ -388,6 +388,22 @@ const struct group_opt_templ all_cmd_options[] = {
    {{"output-goto",
      boost::program_options::value<std::string>(),
      "Export generated goto program"},
+    {"vcc-cache",
```
Yiannis128: Non-blocking nit: the flag says vcc-cache but the docs page calls the feature the Proof Cache, and "VCC" is jargon ESBMC's own output avoids (it says "claim"). Worth aligning, or at least cross-referencing the term in --help so the flag name leads users to the docs.

website/content/docs/proof-cache.md
```diff
@@ -0,0 +1,222 @@
...
+  part of what other claims downstream depend on.
+- Adding or removing a function, or anything that changes the program's set of
+  objects, can invalidate claims that touch memory even in untouched functions.
+
+Measured over 671 regression files, inserting a statement into one function
+left **84.9%** of a file's claims reusable, and a re-run with no edit at all
+reused **93.6%** — the shortfall being claims the solver refuted, which are
+never stored. Every verdict was unchanged.
+
+## What a claim is keyed on
```
Yiannis128, line 125: Suggestion for this section: note that the context half is only ESBMC_VERSION, so two different builds of the same version share keys and a locally rebuilt binary can reuse proofs computed by code that no longer exists. A sentence advising --vcc-cache-verify after rebuilding ESBMC itself would cover it.

src/util/ssa/vcc_cache.cpp
```diff
@@ -0,0 +1,105 @@
...
+  std::filesystem::create_directories(dir, ec);
+  if (ec)
+    log_warning("VCC cache: cannot create {}: {}", dir, ec.message());
+}
+
+std::string vcc_cachet::entry_path(const std::string &cone_key) const
+{
+  const std::string key =
+    fmt::format("{:016x}{}", fingerprint_hash(context), cone_key);
+  return dir + "/" + key.substr(0, 2) + "/" + key + ".vcc";
```
Yiannis128: Nit: string concatenation with "/" instead of std::filesystem::path joins. Works on Windows in practice, but path joining exists for a reason.

Yiannis128 (changes requested):
Excellent documentation and an honest cost model, and the design respects the user: per-run reuse counts, a verify mode, a fingerprint dump, entries safe to delete, only proofs stored. Requesting changes solely on the two silent-inertness issues in bmc.cpp (cache does nothing without --multi-property, deactivation has no runtime signal); the rest are suggestions and nits.

lucasccordeiro:
Thanks — all six addressed in 4b77e8b.

The two blocking ones: the flags are now refused when they could not work (`--proof-cache` without `--multi-property`, `--proof-cache-verify` with no cache to check), and every mode that deactivates the cache prints one line — `WARNING: Proof cache: inactive (<reason>); every claim will be solved and none stored`. The reason list lives in a single function used by both the driver and `make_proof_cache`, so the two cannot drift.

On the option list: yes, widened, and `--color` turned out to be the case for it. It resolves from `isatty(stderr)`, so an interactive run and the same run in a pipeline were disagreeing on every key while agreeing on every verdict. Each of the seven (`--color`, `--ascii-report`, `--verbosity`, `--quiet`, `--log-message`, `--file-output`, `--cex-output`) was checked by reading its use sites and then measured end to end; a control set of semantic flags still misses as it should.

On the context being only `ESBMC_VERSION` — rather than document that, the key now names the build: the ID `scripts/buildidobj.py` links into the binary (commit plus dirty flag), the backends this build enabled (`pick_default_solver()` never writes the implicit choice back into the option set), and a digest of the executable itself where the ID names a class of builds rather than one. Verified by rebuilding a dirty tree and confirming the earlier proofs are not reused. What is left outside the key is a second build of the same commit with the same backends from a different toolchain; the docs say so and point at `--proof-cache-verify`.

Renamed throughout to `--proof-cache` / `--proof-cache-verify` / `--claim-fingerprint-dump`, with the docs page linked from `--help`, and `std::filesystem::path` joins instead of string concatenation.

Two things a review pass afterwards caught, also fixed here: the new check was bypassed by `--skip-bmc` and by the `--falsify-context-bound` prepass, both of which reached a verdict with the flag silently inert; and under `--proof-cache-verify` any non-UNSAT result was reported as the stored proof being contradicted, so a solver error failed the run. Only a counterexample counts now. Six new regression tests cover the refusals, the two inactive paths and the bypasses.

LukeW1999 (commented):
Reviewed at `4b77e8b771`, so after the six fixes. Rebuilt that head and ran
everything below on it.

Three of the things I had written up before that commit are now closed by it:
the build identity is no longer the release triple, `--proof-cache-verify` no
longer reports disagreement on a solver `unknown`, and an inactive cache says
so. The five below survive, and the first one is a live false proof.

## 1. `strip_type_location` merges two different cones, and I can make it prove a false claim

`src/util/ssa/fingerprint.cpp:60`:

```cpp
const size_t at = name.find("_at_");
return at == std::string::npos ? name : name.substr(0, at);
```

The comment above it says anonymous types are named `..._at_<file>_<line>_<col>`,
which is true, but `find` is not anchored and this runs on every non-symbol
`irep_idt` through `rename()` — member names, type tags, string literals. Symbol
names get `assign_names`, which keeps the map bijective by appending a per-cone
index on collision. Nothing else does, so a user identifier containing `_at_` is
truncated with no disambiguation.

Two programs differing only in which field they write:

```c
/* b1.c */                             /* b2.c */
struct S { int val_at_a, val_at_b; };  struct S { int val_at_a, val_at_b; };
int nondet_int(void);                  int nondet_int(void);
int main(void) {                       int main(void) {
  struct S s; int v = nondet_int();      struct S s; int v = nondet_int();
  __ESBMC_assume(v > 0 && v < 10);       __ESBMC_assume(v > 0 && v < 10);
  s.val_at_a = v;                        s.val_at_b = v;
  __ESBMC_assert(s.val_at_a > 0, "p");   __ESBMC_assert(s.val_at_a > 0, "p");
  return 0; }                            return 0; }
```

Ground truth, no cache: `b1` PASSED, `b2` FAILED — `val_at_a` is never written
in `b2`.

```
$ esbmc b1.c --multi-property --proof-cache d
Proof cache: 0 claim(s) reused, 1 solved
VERIFICATION SUCCESSFUL

$ esbmc b2.c --multi-property --proof-cache d
Proof cache: 1 claim(s) reused, 0 solved
VERIFICATION SUCCESSFUL          <-- ground truth is FAILED
```

`--claim-fingerprint-dump` shows why: the raw digest differs
(`cd6c0c82…` / `9c1447ce…`) but both normalised digests are identical
(`0d52a915d1b2e628`, `d27c2d566b95a42b`) across the UNSAT and the SAT cone.

Control, isolating the cause to the substring — same two programs with the
fields renamed `val_x_a` / `val_x_b`:

```
VERIFICATION SUCCESSFUL     (0 reused, 1 solved)
VERIFICATION FAILED         (0 reused, 1 solved)
```

The header comment at `fingerprint.cpp:163-165` says the stripping "costs hits
rather than merging anything". A lossy map merges by construction; that is what
`assign_names` exists to prevent for symbols. Either route these through the
same disambiguation, or anchor the match to the shape the frontend actually
emits (`clang_c_convert.cpp:4893/4935/4966`) rather than any `_at_`.

## 2. The data model is keyed in part, and its default comes from the build host

`proof_cache.cpp:176` hashes `word_size`, the four integer widths,
`char_is_unsigned` and `endianess`. Not `config.ansi_c.target`,
`address_width`/`pointer_width()`, `single_width`, `double_width`,
`long_double_width`, `wchar_t_width`, `bool_width`, `char_width`. With no
`--32`/`--i386-linux`-style flag, `configt::this_architecture()` and
`this_operating_system()` (`config.cpp:211`) take the target from the build
host, so it is in neither `option_map` nor the hashed model. x86-64 Linux and
x86-64 macOS, or riscv64 and x86-64 Linux, produce a byte-identical context
while differing in libc models, predefined macros and `long_double_width`.
Sharing one cache directory across CI machines is what the docs recommend.

## 3. "Every option folded in unabridged" does not hold for repeatable options

`optionst::cmdline` (`src/util/config/options.cpp:56-67`) calls `set_option`
per value, which overwrites, so only the last value of a
`value<std::vector<std::string>>` option reaches `option_map`: `-D`, `-I`,
`--include-file`, `--idirafter`, `-f`, `--enforce-contract`,
`--replace-call-with-contract`, `--no-slice-name`, `--no-slice-id`. `-DA -DB`
and `-DB` give the same context. The comment above the loop has the right
instinct; the code gets a curated list anyway, silently. Folding
`cmdline.vm` in directly would cover it.

## 4. The cone digest does not cover every step it keeps

`ssa_cone_digest` feeds `(type, guard, cond)`. For assignments that is complete
(`symex_target_equation.cpp:194` puts `lhs == rhs` in `cond`). For `RENUMBER` it
is not: `renumber()` (`:265-282`) leaves `cond` nil and carries the symbol in
`lhs` and the new object size in `rhs`, `convert_internal_step` uses all three,
and the slicer keeps such steps (`slice.cpp:308`). Two cones differing only in a
reallocated object's size digest the same. `OUTPUT` args are likewise nil in
`cond` but walked by `pre_register_addresses`. No live caller reaches
`renumber()` today, so this is latent — but `fingerprint.h:29-31` states the
triple is "a complete account of what the step contributes to the formula",
and the day that changes there is no signal.

## 5. The hit path is not equivalent to the UNSAT path

`bmc.cpp:2859` returns at `:2870`, before the `P_UNSATISFIABLE` branch prunes
the claim via `clear_verified_claims_in_ssa` / `_in_goto` (`:3266`) when
`!is_keep_verified && !bs`. Reachable with `--diagnose-unknown-properties` and
with `--inductive-step --loop-invariant`, the other two gates on
`multi_property_check`. A warm run then leaves the goto program in a different
state than a cold one and keeps re-symexing claims a cold run had removed —
behavioural rather than unsound, but the opposite of the point.

---

The measurement report is the right shape, and the interleaved repeats are what
makes it readable. Worth noting for the record that a 672-file soundness sweep
cannot catch #1: it compares each file against itself across scenarios, and the
collision needs two different files sharing a cache. A cross-file check — one
cache directory for the whole suite, then compare against per-file verdicts —
would have surfaced it.

lucasccordeiro:
Thanks — this is a good review, and the first one was real. All five are fixed in `39998c5`..`f388bbb`, one commit each.

## 1. `strip_type_location` — confirmed, and it was a false proof

Reproduced your `b1`/`b2` before touching anything: `b1` stored, `b2` reused, `VERIFICATION SUCCESSFUL` where ground truth is `FAILED`.

The root cause is the asymmetry you named. Symbols go through `assign_names`, which stays bijective by appending a per-cone index on collision; every other `irep_idt` goes through `rename()`, which has nothing of the kind. So a strip applied there has to be **injective on its own**, and an unanchored `_at_` search is not — it takes part of a user identifier with it.

Anchored to the `__anon_` marker the frontend actually emits (`clang_c_convert.cpp:4893/4935/4966`) rather than to any `_at_`. Two anonymous types that still collide once their positions are gone have the same members, so they encode alike anyway; a user identifier never carries the marker. The comment at `fingerprint.cpp:163-165` that claimed this "costs hits rather than merging anything" was wrong and is corrected — it now says what the constraint on such a strip is.

The obvious worry is that anchoring costs reuse. It costs none. A/B against a control binary differing only in that function, 140 regression files, cold / warm / comment-at-top:

| | cold | warm | comment at top |
|---|---|---|---|
| unanchored | 42 / 225 | 147 / 225 | 147 / 225 |
| anchored | 42 / 225 | 147 / 225 | 147 / 225 |

Byte-identical. Anonymous aggregates are named by line and column rather than character offset, so they are far less mobile than locals are, and the marker is present in every name that has a position to cut.

## 2–4

- **Data model.** Was seven scalars. Now the target triple, every width, `pointer_width()`, signedness, CHERI, endianness, locale, lib and the C/C++ dialect. The docs said to share one directory across CI; they now say a Linux runner and a macOS runner populate disjoint entries rather than reusing each other's.
- **Repeatable options.** `optionst` keeps the full value list alongside `option_map`. One subtlety worth flagging: the list is folded in *on top of* the live `option_map` value, not in place of it — k-induction rewrites `unwind` per step, and keying on the stale command-line value alone would have made every k step share a key. `-DA -DB` and `-DB` now differ; the repeat still hits.
- **Cone digest.** `RENUMBER`'s `lhs`/`rhs` and `OUTPUT`'s arguments now reach the key, gated by step type so an assignment pays nothing extra. Latent today, agreed — the point of fixing it is that `fingerprint.h` claimed completeness, and it now does what it claims.

## 5. Hit path vs UNSAT path

Fixed under the same `!is_keep_verified && !bs` guard as the `P_UNSATISFIABLE` arm.

## On your closing note

You are right that a 672-file sweep cannot see this, and that is the part worth keeping: it compares each file only against itself, and the collision needs two different files sharing one cache. `proof-cache-bench.py` now takes `--shared-cache`, which puts every task in one directory the way a persisted CI cache works. On the pre-fix binary:

```
$ proof-cache-bench.py --shared-cache b1.c b2.c        # pre-fix
b2.c                                        MISMATCH FAILED/SUCCESSFUL/SUCCESSFUL
tasks measured: 1, MISMATCHES: 1

$ proof-cache-bench.py b1.c b2.c                       # pre-fix, per-task cache
tasks measured: 2, MISMATCHES: 0
```

The default mode is blind to exactly the thing the shared mode catches.

## Testing, and what is not covered

Six unit tests — four in `ssa_fingerprint.test.cpp`, two in `proof_cache.test.cpp` — each mutation-checked: all six fail on the pre-fix code and pass on the fixed. The existing "an anonymous type's location does not change the digest" case used a tag with no `__anon_` marker, which the frontend never emits; it is corrected to the real shape.

Two gaps, stated plainly:

- **No regression test for 1 or 5.** Both need two runs sharing a cache, which `test.desc` cannot express — `SEED_FILE` cannot seed an entry whose name embeds the build identity. A single-file version of 1 is not constructible either: the two claims have to write to distinguishable storage, so their cones differ in a `symex::nondet` index or a folded constant regardless of the member names. Hence the unit tests.
- **Still no full clean regression suite on this branch.** Sampled `-L esbmc -I ,,6`; the only failures are the Solidity suite (dead on macOS) and `esbmc-cpp/cpp/github_2242_2`, which is a host libc++ `<cctype>` PARSING ERROR before ESBMC runs. 16/16 proof-cache regression tests and 738/738 unit tests pass.

PR #7144, opened 2026-08-19
alias numpy 1-D views to a shared buffer via pointers

sergillam:
- Implements the first slice of the definitive NumPy view descriptor model (ADR-NP-003 etapa 2), scoped to a 1-D, unit-stride, literal-bound slice assigned to a bare name (`v = a[1:3]`). Instead of copying elements into an independent array symbol (the current behaviour for every view-like expression), `v` becomes a `pointer_typet` pointing into `a`'s own buffer (`build_address_of` over the existing pointer-indexing path in `build_index`).
- Reads and writes through `v` now observe `a`, and vice versa; `len(v)`, `v.shape`, and `v.ndim` were fixed to report the view's own logical length instead of falling through to the base buffer's remaining size (`numpy_pointer_view_lengths_`, new).
- Flips the three existing tests whose scenario is exactly this case from `_fail` to `_success` with corrected assertions (`view_slice_write_view_*`, `view_slice_write_source_*`); everything outside this slice (row/column views, transpose, reshape/ravel, `.flat`, step≠1, symbolic bounds, general escape) is unchanged and still explicitly rejected/copied.
- Add test units


src/python-frontend/python-list/list_access.cpp
```diff
@@ -1337,6 +1347,49 @@ const symbolt &python_list::get_str_slice_sym()
...
+  const exprt &array,
+  const typet &elem_type,
+  long long step_val,
+  bool needs_null_term,
+  long long literal_start,
+  long long static_slice_len)
+{
+  const namespacet ns(converter_.symbol_table());
+  if (
+    step_val != 1 || needs_null_term || ns.follow(elem_type).is_array() ||
```
lucasccordeiro, line 1404: Aliasing survives rebinding the base: `a=np.array([1,2,3,4]); v=a[1:3]; a=np.array([9,9,9,9])` makes `v[0]` read 9, so a valid assert fails. Invalidate the view on base rebind.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -1825,12 +1825,28 @@ void python_converter::reject_unsafe_numpy_view_target(
...
 
+  // A view symbol this PR's 1-D slice aliasing retyped to a pointer (see
+  // list_access.cpp's handle_range_slice, which populates
+  // numpy_pointer_view_lengths_ exactly for that case) genuinely aliases
+  // its source's storage: writing through it, or through the source while
+  // it is live, is sound pointer semantics, not the copy-divergence this
+  // guard otherwise exists to reject. Checking membership in that map
+  // (rather than just "is this symbol's type a pointer") avoids misreading
+  // some other, unrelated pointer-typed symbol as one of these views.
+  auto is_pointer_backed = [this](const std::string &id) {
```
lucasccordeiro, line 1836: This guard only runs for Subscript targets, so a whole-name rebind of a viewed array is never seen. Handle Name targets to drop views of the rebound array.

src/python-frontend/python-list/list_access.cpp
```diff
@@ -1337,6 +1347,49 @@ const symbolt &python_list::get_str_slice_sym()
...
+  // an empty slice at the very end of the source (literal_start ==
+  // source_len) computes a one-past-the-end address, legal to form but not
+  // to dereference (C 6.5.6p8). index2tc would route it through the array
+  // bounds checker as an out-of-bounds subscript; add2tc over the decayed
+  // pointer is checked under pointer, not array-index, rules.
+  const typet view_ptr_type = pointer_typet(elem_type);
+  exprt base_ptr = build_typecast(build_address_of(array), view_ptr_type);
+  exprt view_ptr = build_add(
+    base_ptr, from_integer(literal_start, size_type()), view_ptr_type);
+  converter_.current_lhs->type() = view_ptr_type;
```
lucasccordeiro, line 1420: Retyping the LHS symbol to a pointer is never undone, so rebinding that name to an array later leaves a pointer-typed symbol holding an array.

src/python-frontend/python-list/list_access.cpp
```diff
@@ -657,10 +657,16 @@ bool try_get_literal_int(const nlohmann::json &node, BigInt &out)
...
+// clamped) literal start offset -- the same value used to compute the
+// length below, exposed for callers that need the actual byte/element
+// offset into the source (e.g. building a pointer view), not just the
+// resulting slice length.
 std::optional<long long> literal_slice_length(
   const nlohmann::json &slice_node,
   long long source_len,
-  long long step)
+  long long step,
+  long long *out_start = nullptr)
```
lucasccordeiro: `literal_slice_length` rises 19->21 and `handle_range_slice` 96->98, failing the complexity gate. Extract the bound normalisation so neither grows.

lucasccordeiro (changes requested):
Rebinding the base array after a view is created makes the view observe the new buffer: `a=np.array([1,2,3,4]); v=a[1:3]; a=np.array([9,9,9,9]); assert v[0]==2` verifies SUCCESSFUL on master and FAILED on this branch. No test covers base rebind; please add one.

PR #7162, opened 2026-08-19
[python] Extend dynamic typing to function return values

brcfarias:
Functions with return statements that diverge in type across an if/else (e.g. `return 1` in one branch, `return "a"` in the other) used to crash goto-symex/SMT encoding, since the return type was worked out by just picking whichever RETURN it came across first. This extends the tagged `PyObject` mechanism already used for local variables and `isinstance()` to cover function returns too, so both branches get reconciled into one tagged type instead.


src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3370,11 +3376,14 @@ void python_converter::get_var_assign(
...
-    // Process RHS before LHS if in function scope
+    // Process RHS before LHS if in function scope, or for a global with an
+    // auto-inferred annotation (needs checking against the actual RHS type).
     exprt rhs;
     if (
-      sid.to_string().find("@F") != std::string::npos &&
-      sid.to_string().find("@C") == std::string::npos)
+      (sid.to_string().find("@F") != std::string::npos &&
+       sid.to_string().find("@C") == std::string::npos) ||
+      (ast_node.value("_inferred_annotation", false) &&
```
lucasccordeiro: `_inferred_annotation` is set on essentially every un-annotated assignment, so this flips RHS-before-LHS for all module-level assignments and regresses DebugOpt CI (`list_pop6`, `list_pop15`, `list_pop14_nondet`, `list-slice-step-reverse`, ~25 numpy tests; `z = l.pop()` now types as `signed long int`, truncating 2.5). Narrow the condition to the tagged-RHS case.

src/python-frontend/converter/converter_funcdef.cpp
```diff
@@ -1381,9 +1381,17 @@ void python_converter::get_function_definition(
...
   // Determine return type
-  if (
-    return_node.is_null() ||
-    (return_node["_type"] == "Constant" && return_node["value"].is_null()))
+  if (return_node.is_null())
+  {
+    // Detects a genuine int/str return divergence, so the function gets a
+    // tagged return type instead of the post-hoc scan below narrowing to
+    // whichever branch it sees first.
+    type.return_type() =
```
lucasccordeiro, line 1396: `already_optional` (line 1608) does not cover the tagged type, so any `return None`/bare `return` elsewhere in the body makes `body_has_none_return` wrap this in `Optional<tagged>`, silently reverting to the mixed-type RETURN. Add the tagged type to `already_optional`.

src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -582,6 +607,74 @@ exprt dynamic_type_handler::build_isinstance_check(
...
+  if (!function_body.is_array())
+    return false;
+
+  for (const auto &stmt : function_body)
+  {
+    if (!stmt.is_object() || stmt.value("_type", "") != "If")
+      continue;
+    if (!stmt.contains("body") || stmt["body"].empty())
+      continue;
+    if (!stmt.contains("orelse") || stmt["orelse"].empty())
```
lucasccordeiro: Requiring a non-empty `orelse` misses the idiomatic early-return shape (`if c: return 1` then `return "a"`) and `elif` chains, which still hit the original crash. Handle those shapes and add regression tests for them.

src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -582,6 +607,74 @@ exprt dynamic_type_handler::build_isinstance_check(
...
+  const std::string &name,
+  codet &target_block)
+{
+  symbol_id sid = converter_.create_symbol_id();
+  sid.set_object(name);
+  std::string tag_id = sid.to_string();
+  auto alias = aliases_.find(tag_id);
+  if (alias != aliases_.end())
+    tag_id = alias->second;
+  symbolt *tag_symbol = converter_.symbol_table().find_symbol(tag_id);
```
lucasccordeiro: Lines 661-671 duplicate `assign()`'s alias-resolution and lookup verbatim. Extract a shared `find_tag_symbol(name)` helper and call it from both.

lucasccordeiro (changes requested):
Cyclomatic complexity CI fails: `get_var_assign` 249→253, `get_function_definition` 101→102, `get_return_statements` 64→69, all past the gated `core` > 15. Extract the new logic into helpers in this PR.

PR #7174, opened 2026-08-20
[docs] Document the full ESBMC intrinsic surface

lucasccordeiro:
The constructs page covered assert/assume, nondet, pragma unroll and
quantifiers; the memory-model, CBMC-compatibility, bulk-memory, overflow,
rounding, offsetof and CHERI intrinsics declared in `internal_additions()` were
absent. Every example was run against ESBMC 8.4.0 before being documented.

Three behaviours the page now pins down: `__ESBMC_cover` is declared for C but
implemented only in the Python frontend, `__ESBMC_unreachable()` is a silent
no-op without `--enable-unreachability-intrinsic`, and C `offsetof` cannot be
used where a constant expression is required.

The pre-existing `#pragma unroll` examples are left untouched: they do not
compile (undeclared `y`) and their loops do not terminate, so repairing them
changes each example verdict and belongs in a separate PR.


website/content/docs/constructs.md
```diff
@@ -91,18 +184,400 @@ int main() {
...
+
+ESBMC accepts the SV-COMP vocabulary directly, so a benchmark written for the
+competition needs no adaptation:
+
+| SV-COMP | ESBMC equivalent |
+| --- | --- |
+| `__VERIFIER_nondet_X()` | `nondet_X()` |
+| `__VERIFIER_assume(int)` | `__ESBMC_assume` |
+| `__VERIFIER_atomic_begin()` / `__VERIFIER_atomic_end()` | `__ESBMC_atomic_begin` / `__ESBMC_atomic_end` |
+| `__VERIFIER_error()` | error sentinel under `--enable-unreachability-intrinsic` |
```
lucasccordeiro: `__VERIFIER_error()` already asserts false without the flag (`builtin_functions.cpp:1336`); the flag only matters when the program supplies its own body. Fix the row.

website/content/docs/constructs.md
```diff
@@ -35,6 +79,17 @@ harnesses and stubs:
 Every ESBMC run defines `__ESBMC_execution`, and the header refuses to compile
 without it — so a harness that includes `esbmc.h` is a hard error under any
 ordinary compiler instead of silently building with the intrinsics undefined.
+The same macro lets a source tree tell an ESBMC run from an ordinary compile:
+
+```c
+#ifdef __ESBMC_execution
+    ESBMC_assume(len < 16);   /* only under verification */
```
lucasccordeiro, line 91: `ESBMC_assume` requires `#include <esbmc.h>`, which `#error`s when `__ESBMC_execution` is undefined. Put the include inside the guard, or use `__ESBMC_assume` here.

website/content/docs/constructs.md
```diff
@@ -91,18 +184,400 @@ int main() {
...
+}
+```
+
+### Reinterpreting bytes
+
+```c
+void __ESBMC_bitcast(void *tgt, void *src);
+```
+
+Equivalent to `memcpy(tgt, src, n)` where `n` is the size of the type the two
```
lucasccordeiro, line 491: Mismatched pointee widths abort on `bitcast2t`'s width assertion, and a may-alias argument on `internal_deref_items.size() == 1`. Document both preconditions.

website/content/docs/constructs.md
```diff
@@ -91,18 +184,400 @@ int main() {
...
+    __ESBMC_memcpy(c, b, 4);
+    assert(__ESBMC_memcmp(c, b, 4) == 0);
+    return 0;
+}
+```
+
+### Reinterpreting bytes
+
+```c
+void __ESBMC_bitcast(void *tgt, void *src);
```
lucasccordeiro, line 488: Nothing under `regression/` calls `__ESBMC_bitcast`. Add a test pinning this example before documenting it as user-facing surface.

website/content/docs/constructs.md
```diff
@@ -91,18 +184,400 @@ int main() {
...
+    __ESBMC_overflow_result s = __ESBMC_overflow_result_plus(1, 2);
+    assert(!s.overflow && s.result == 3);
+    return 0;
+}
+```
+
+The macro declares `__ESBMC_overflow_result` as a packed
+`struct { _Bool overflow; type result; }` together with
+`__ESBMC_overflow_result_plus`, `_minus`, `_mult`, `_shl` (all binary) and
+`_unary_minus`.
```
lucasccordeiro, line 582: The typedef name is fixed, so a second `DEFINE_ESBMC_OVERFLOW_TYPE` in the same translation unit is a redefinition error. State that limit.

website/content/docs/constructs.md
lucasccordeiro: Confirmed and fixed in f79f2ba. Reproduced all three cases:

- declared, no body, no flag -> `VERIFICATION FAILED` (asserts false on its own)
- own body, no flag -> `VERIFICATION SUCCESSFUL` (the body is called)
- own body, with flag -> `VERIFICATION FAILED` (`skip_body` suppresses it)

The row now reads "Asserts false at the call", and the section below states that the flag's only effect is skipping a body the program itself supplies.

website/content/docs/constructs.md
lucasccordeiro: Confirmed and fixed in f79f2ba. With the include at top level, an ordinary `clang -fsyntax-only` gives `error: "esbmc.h can only be used for ESBMC runs"` — which is exactly the compile the guard was meant to protect. Moving the include inside the `#ifdef` compiles clean under clang and still verifies under ESBMC. Added a note that ESBMC-only code can use the `__ESBMC_`-prefixed names directly and skip the header.

website/content/docs/constructs.md
lucasccordeiro: Confirmed and documented in f79f2ba. Both reproduce as hard aborts, not diagnostics:

- 4-byte target from an 8-byte source: `Assertion failed: (type->get_width() == from_->type->get_width()), function bitcast2t, file irep2_expr.h, line 769`
- source that may address either of two objects: `Assertion failed: (internal_deref_items.size() == 1), function run_intrinsic, file symex_main.cpp, line 947`

Both are now a warning callout naming the condition and the assertion, with a steer to `memcpy` when either is not statically guaranteed.

Separately: aborting on ordinary bad input rather than emitting a diagnostic looks like a defect in its own right. I have documented current behaviour truthfully here and left the C++ alone; happy to file an issue if you agree.

website/content/docs/constructs.md
lucasccordeiro: Confirmed — zero hits for `__ESBMC_bitcast` anywhere under `regression/`. Added two tests in f79f2ba:

- `regression/esbmc/bitcast_intrinsic` — float/uint32 and double/uint64 round-trip, `VERIFICATION SUCCESSFUL`, pinning the documented example
- `regression/esbmc/bitcast_intrinsic_fail` — same bitcast asserting the wrong bit pattern, `VERIFICATION FAILED`

The `_fail` case is the mutation of the passing one, so the pair cannot pass vacuously. Both verified through `testing_tool.py`, not just a bare binary run.

website/content/docs/constructs.md
lucasccordeiro: Confirmed and documented in f79f2ba — and the limit is stricter than stated. Two *different* types collide as expected, but so does the same type twice, since each expansion defines a fresh anonymous struct:

```
DEFINE_ESBMC_OVERFLOW_TYPE(int)
DEFINE_ESBMC_OVERFLOW_TYPE(int)
-> error: typedef redefinition with different types
```

Documented as at most one per translation unit, with the reason and the suggestion to split across translation units to cover several operand types.

PR #7193, opened 2026-08-20
alias numpy 2-D row views to shared buffers

sergillam:
- Adds pointer-backed aliasing for literal 2-D NumPy row views such as row = a[0] and row = a[-1].
- Makes row-view reads and writes observe the original array storage in both directions.
- Preserves len(row), .shape, and .ndim for pointer-backed row views.
- Keeps symbolic row indexes and unsupported view cases outside this slice.
- Preserves simple named NumPy array aliases (b = a) for row-view consumers.
- Canonicalizes view sources so rebinding the original array detaches live pointer-backed views correctly.


src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3828,7 +3951,11 @@ void python_converter::get_var_assign(
   // unannotated assignment branches above -- the annotator may inject an
   // annotation onto what the user wrote as a plain `a = ...`, routing it
   // through either one.
-  if (target.value("_type", "") == "Name" && lhs_symbol)
+  const bool is_self_assignment =
+    target.value("_type", "") == "Name" &&
+    ast_node["value"].value("_type", "") == "Name" &&
```
lucasccordeiro: `ast_node["value"]` is null for a bare `x: int`, so `.value()` throws type_error.306 — 13 python tests fail on macOS CI. Check `is_null()` first.

src/python-frontend/converter/converter_funcall.cpp
```diff
@@ -369,12 +413,72 @@ std::optional<exprt> python_converter::try_get_numpy_pointer_view_len(
...
+    row_index += row_count;
+  if (row_index < 0 || row_index >= row_count)
+  {
+    std::ostringstream msg;
+    msg << "IndexError: index " << *literal_index
+        << " is out of bounds for axis 0 with size " << row_count;
+    const locationt location = get_location_from_decl(element);
+    if (!location.is_nil())
+      msg << " at " << location.get_file() << ":" << location.get_line();
+    throw std::runtime_error(msg.str());
```
lucasccordeiro: Row out-of-bounds aborts the frontend, while column out-of-bounds raises a catchable IndexError. Use `gen_exception_raise` here so `try/except IndexError` still verifies.

regression/numpy/view_mutation_slice_assign_fail/test.desc
```diff
@@ -1,4 +1,4 @@
 CORE
 main.py
 --incremental-bmc
-^ERROR: TypeError: writing through a copied numpy view is not supported
+^ERROR: Slice assignment is only supported on list targets$
```
lucasccordeiro: This target is a numpy view, not a non-list, so the new message misdiagnoses it. Keep a numpy-specific diagnostic for slice assignment through a view.

lucasccordeiro (changes requested):
Cyclomatic complexity CI fails: `get_var_assign` 237→241 and `update_numpy_array_binding` 15→16, both already over threshold. Extract helpers so neither rises.

lucasccordeiro:
Dozens of unrelated comment re-wraps in `converter_stmt.cpp` and `converter_funcall.cpp` obscure the functional change. Revert the reflow-only hunks.

PR #7206, opened 2026-08-21
[contracts] Grant the element the clause named on entry

### LINKED ISSUE #7103: [contracts] the element-level frame rule accepts a broken frame and proves a false assertion
Three defects in the element-level frame rule, introduced by #7069 (`de37320176`). Two are unsound: one accepts a body that breaks its frame, one proves an assertion that is false. Reproducers below run against `9b547f59b1`.

## 1. An assigns index is evaluated in the post-state, so a body chooses after the fact what it was granted

The array is snapshotted; the index is not. When the index is itself in the clause -- the ring-buffer idiom -- the body can move it and retroactively legalise an out-of-frame write.

```c
int buf[1000];
int head;
extern int nondet_int(void);

void push(int v)
{
  __ESBMC_requires(head >= 0 && head < 999);
  __ESBMC_assigns(buf[head], head);
  __ESBMC_ensures(1);
  head = head + 1;
  buf[head] = v;   /* writes buf[head_pre + 1]: outside the frame */
}

int main(void)
{
  head = nondet_int();
  __ESBMC_assume(head >= 0 && head < 999);
  push(7);
  return 0;
}
```

```
$ esbmc main.c --enforce-contract push
VERIFICATION SUCCESSFUL          <-- the frame violation is accepted
```

Swap the two body statements so the body is correct (`buf[head] = v; head = head + 1;`) and it is rejected:

```
Violated property: assigns compliance: c:@buf not in assigns clause
VERIFICATION FAILED
```

The verdicts are exactly inverted. Reproduces on both encodings, so at any extent. ACSL evaluates `assigns` targets in the pre-state, and every other part of the enforcer agrees -- only the index escapes the snapshot.

## 2. Above the element-wise cap the loop frame rule assumes the array wholly unchanged, which is a false hypothesis

`emit_array_elem_frame` declines extents over 256 in ASSUME mode, so `enforce_frame_rule` falls through to `assume(global == snap)` -- asserting the loop wrote *nothing* to an array whose element it was granted. A stronger hypothesis than the truth proves things that are false.

```c
int global[512];
int main()
{
  int i = 0;
  __ESBMC_loop_assigns(i, global[3]);
  __ESBMC_loop_invariant(i >= 0 && i <= 4);
  while (i < 4) { global[3] = 1; i++; }
  __ESBMC_assert(global[3] == 0, "FALSE: the loop wrote global[3] = 1");
  return 0;
}
```

```
$ esbmc main.c --loop-invariant-check --loop-frame-rule --z3
VERIFICATION SUCCESSFUL          <-- a false assertion, proved
```

Controls: the same file at `global[256]` FAILS (element-wise path, correct); dropping `--loop-frame-rule` FAILS; plain `--unwind 10` FAILS. So the frame rule is the cause.

On Bitwuzla this is masked -- the solver hits "Equality over constant arrays not fully supported yet" and aborts, which `regression/loop-invariants/loop_assigns_large_array_knownbug` currently pins. That test's comment says the solver fails outright; the more important half is that when the solver does answer, the answer is wrong.

## 3. The store form truncates an out-of-range index, sparing an unrelated element

Past the cap the array is held as `global == (snap WITH [i := global[i]])`. The solver's index domain is only as wide as the extent, so an index past the end wraps.

```c
int global[1000];
void f(int v)
{
  __ESBMC_assigns(global[1024]);
  __ESBMC_ensures(1);
  global[0] = v;
}
int main() { return 0; }
```

```
$ esbmc main.c --enforce-contract f --function f --no-bounds-check
VERIFICATION SUCCESSFUL          <-- 1024 wrapped to 0 and spared global[0]
```

`global[1025]` spares `global[1]`, and so on. The element-wise path rejects the identical shape, because it compares the index against each element as a value rather than using it to index. So crossing the cap turns a correct rejection into acceptance.

## Not part of this, but found alongside

`--replace-call-with-contract` trusts a contract without requiring the enforce side to have passed, so the reproducer in 1 proves a false property end to end even after 1 is fixed. That predates #7069 -- it reproduces on a pre-#7069 build -- and is filed separately.

LukeW1999:
Closes the pointer-parameter half of #7103.

## The defect

`__ESBMC_assigns(buf[head])` names its target in the pre-state: it grants the
element `head` denoted on entry. The Phase 2B frame check for a pointer
parameter recorded the index *expression* and read it back after the body, so a
body that moves `head` — itself in the clause, as the ring-buffer idiom needs —
chose after the fact which element it had been granted.

Both verdicts come out inverted. On `a082fe7d1d`:

```c
int head;
void push(int *buf, int v)
{
  __ESBMC_requires(__ESBMC_is_fresh(buf, 8 * sizeof(int)));
  __ESBMC_requires(head >= 0 && head < 7);
  __ESBMC_assigns(buf[head], head);
  buf[head] = v;      /* exactly the granted element */
  head = head + 1;
}
```

```
FAILED  assigns compliance: ...@push@buf[j] not in assigns clause (Phase 2B)
```

and the broken body, which writes `buf[head + 1]` — an element the clause never
named:

```c
  head = head + 1;
  buf[head] = v;
```

```
VERIFICATION SUCCESSFUL
```

The second is the unsound direction: a body writing outside its frame is
accepted, which is what #7103 named for the global-array path.

## The fix

#7103 fixed the global-array path by mapping the index through `in_pre_state`
before building the frame constraint. Phase 2B has no `active_snapshots` to map
through, so it snapshots the index into its own variable instead.
`materialize_arr_elem_snapshots` runs before the call and
`emit_arr_elem_assertions` after it, so capturing there is enough:

```cpp
didx_assign->code = code_assign2tc(declared_idx, idx_expr);
...
entry.declared_idx = declared_idx;
```

The witness clamp used the same live expression as its fallback and moves to the
snapshot with it.

## Tests

- `github_7103_assigns_index_prestate_ptr` — the ring-buffer push above, must
  verify
- `github_7103_assigns_index_prestate_ptr_fail` — the off-by-one write, must be
  caught

`regression/function_contract` 404/404, `loop-invariants` and
`python-contracts` 135/135.

https://claude.ai/code/session_01MEkhmEkbhB4tUdSnqYntPA


LukeW1999:
@lucasccordeiro rebased onto master, so this now sits on top of #7217.

Worth saying explicitly that the two are independent: #7217 widened *which*
indices a check excuses, this fixes *when* those indices are read. Re-tested on
`e77486a3af` (master with #7217 in) before rebasing, and the defect is
unchanged there:

```
buf[head] = v; head = head + 1;   /* writes exactly the granted element */
  FAILED   assigns compliance: ...@push@buf[j] not in assigns clause (Phase 2B)

head = head + 1; buf[head] = v;   /* writes buf[head + 1], never granted */
  VERIFICATION SUCCESSFUL
```

The second is the unsound direction and the shape #7103 was opened for.

The rebase changed the fix: master now groups targets per array, so each index
in `group.indices` gets its own pre-state capture rather than the single
`declared_idx` the first version snapshotted. `regression/function_contract`
416/416, including the `github_7184_mixed_index_types` cases from #7217.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3622,6 +3622,45 @@ code_contractst::materialize_arr_elem_snapshots(
...
+      didx_decl->location.comment(
+        "frame: array-elem declared index (Phase 2B)");
+
+      goto_programt::targett didx_assign = wrapper.add_instruction(ASSIGN);
+      didx_assign->code = code_assign2tc(didx, idx);
+      didx_assign->location = location;
+      didx_assign->location.comment(
+        "frame: capture declared index pre-state (Phase 2B)");
+
+      declared_indices.push_back(didx);
```
lucasccordeiro, line 3863: Nothing tests a non-first index: reverting all but `indices.front()` to the live expression keeps `function_contract` 414/414 green. Add a two-index case whose second index moves.

LukeW1999: Right, and I reproduced it: reverting all but `indices.front()` to the live expression leaves the suite green.

Added `github_7103_assigns_index_prestate_ptr_second{,_fail}`, naming
`__ESBMC_assigns(buf[head], buf[tail], tail)` and moving `tail`, the second
index on that array. 418/418.

One thing worth recording rather than leaving implied: only the `_fail`
direction discriminates. Under the front-only variant:

```
_second       (writes buf[tail] then moves tail)    PASSES  -- no signal
_second_fail  (moves tail then writes buf[tail])    PASSES  -- so the test goes red
```

The out-of-frame write is the one that gets wrongly excused when the index is
read after the body, so that is where the coverage is. I kept `_second` as the
in-frame counterpart, but it is documentation, not a guard, and I would rather
say so than have someone later trust it to catch a regression it cannot see.

I do not have a confirmed account of why the in-frame direction stays green
here when the single-index `..._ptr` case does not; forcing `head != tail` does
not change it. If you would rather not carry a test that proves nothing, I am
happy to drop `_second` and keep only `_second_fail`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3622,6 +3622,45 @@ code_contractst::materialize_arr_elem_snapshots(
     const irep_idt &arr_id = to_symbol2t(arr_ptr).thename;
     const std::string &arr_name = id2string(arr_id);
 
+    // A clause names its targets in the pre-state: `__ESBMC_assigns(buf[head])`
+    // grants the element `head` denoted on entry. Read back after the body, a
+    // body that moves `head` -- itself in the clause, as the ring-buffer idiom
+    // needs -- chooses after the fact which element it had been granted, so an
+    // off-by-one write verifies. Same defect as #7103 on the global-array path,
+    // which fixed it by mapping the index through `in_pre_state`; this path has
+    // no snapshot table to map through, so it captures each index itself.
```
lucasccordeiro: `active_snapshots` is populated here: step 3b-i snapshots every global just above this call. Either use `frame_enforcer.in_pre_state(idx)` for those, or state the real reason (non-global indices).

regression/function_contract/github_7103_assigns_index_prestate_ptr_second/main.c
```diff
@@ -0,0 +1,22 @@
+/* The moving index is the second one the clause names, so this fails unless
```
lucasccordeiro: Measured: this verifies on master too, where no index is captured. "fails unless every index is captured on entry" is wrong — correct the comment or drop the test.

regression/function_contract/github_7103_assigns_index_prestate_ptr_second_fail/main.c
```diff
@@ -0,0 +1,22 @@
...
+ * one the clause names. */
+int head;
+int tail;
+
+void push_tail_bad(int *buf, int v)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(buf, 8 * sizeof(int)));
+  __ESBMC_requires(head >= 0 && head < 7);
+  __ESBMC_requires(tail >= 0 && tail < 7);
+  __ESBMC_assigns(buf[head], buf[tail], tail);
```
lucasccordeiro, line 13: Two symbolic indices still miss out-of-frame writes: `assigns(buf[head], buf[tail])` with `head==0`, `tail==1` and body `buf[5]=v` is SUCCESSFUL. A literal second index is caught. File it.

lucasccordeiro (changes requested):

src/goto-programs/contracts/contracts.cpp
LukeW1999: You are right, and the comment was wrong: `frame_enforcer.materialize_snapshots(globals, ...)` runs at 3b-i, well before this.

Stating the real reason instead of switching to `in_pre_state`, because it does not reach every index here. It resolves a symbol against `active_snapshots`, and `collect_global_variables` (`frame_enforcer.cpp:474`) skips pointer types — so the same ring-buffer idiom with the cursor passed in rather than global resolves to nothing:

```c
void push_d(int *buf, int *p, int v)
{
  __ESBMC_requires(__ESBMC_is_fresh(buf, 8 * sizeof(int)));
  __ESBMC_requires(__ESBMC_is_fresh(p, sizeof(int)));
  __ESBMC_requires(*p >= 0 && *p < 7);
  __ESBMC_assigns(buf[*p], *p);
  *p = *p + 1;
  buf[*p] = v;          /* writes buf[*p + 1], never granted */
}
```

```
master:      PASSED  (Phase 2B)  VERIFICATION SUCCESSFUL
this branch: FAILED  (Phase 2B)  VERIFICATION FAILED
```

Added as `github_7103_assigns_index_prestate_ptr_deref_fail`, and the comment now says which symbols reach `in_pre_state` rather than claiming there is no table.

Worth noting a parameter index needs no fix at all — `assigns(buf[i], i)` with `i = i + 1; buf[i] = v;` already FAILS on master, since the body moves its own copy and the wrapper reads the caller side.

regression/function_contract/github_7103_assigns_index_prestate_ptr_second/main.c
LukeW1999: Correct, and I should have caught it — I noted in my previous reply that only the `_fail` direction discriminates, then left a test whose comment claimed the opposite.

Dropped it. `_second_fail` stays, with its comment rewritten to say the in-frame direction is not a guard.

regression/function_contract/github_7103_assigns_index_prestate_ptr_second_fail/main.c
LukeW1999: Reproduced, and it predates this branch — master `c00bdd53e7` is `VERIFICATION SUCCESSFUL` with no `assigns compliance` property emitted at all, so the check is never generated rather than discharged. One symbolic index is caught, and symbolic-plus-literal is caught; two symbolic on the same array is what loses it.

Filed as #7355.

LukeW1999: Correcting my previous reply on this thread: I confirmed the symptom without checking the reproducer was satisfiable, and it is not. The bug does not exist.

`head` and `tail` are globals with no initialiser, so they are zero, and `__ESBMC_requires(tail == 1)` cannot hold. The requires lowers to a false assumption, nothing after it is reachable, and that is the whole of the `Generated 0 VCC(s)` / `VERIFICATION SUCCESSFUL`. No frame check is missing — none of the function was reached.

It is not contract machinery either: a plain `__ESBMC_assume(other == 1)` on a zero-initialised global under `--function`, no contracts anywhere, gives the same 0 VCCs.

Retested with initialisers so the requires holds:

```c
int head = 0;
int tail = 1;
void w(int *buf, int v) {
  __ESBMC_requires(__ESBMC_is_fresh(buf, 8 * sizeof(int)));
  __ESBMC_requires(head == 0);
  __ESBMC_requires(tail == 1);
  __ESBMC_assigns(buf[head], buf[tail]);
  buf[5] = v;
}
```

```
Generated 31 VCC(s), 18 remaining
FAILED  assigns compliance: ...@w@buf[j] not in assigns clause (Phase 2B)
VERIFICATION FAILED
```

And independently, two symbolic indices with range requires zero satisfies (`head >= 0 && head < 8`, `tail >= 0 && tail < 8`) gives 14 VCCs and FAILED. Two symbolic targets do emit their check and do catch the write.

#7355 is closed as invalid with the measurements.

What the episode does show is worth its own issue: an unsatisfiable `requires` makes `--enforce-contract` report `VERIFICATION SUCCESSFUL` having proved nothing, and the only signal is the VCC count. Both of us read a vacuous run as a missing frame check, which is the failure mode a diagnostic there would prevent. Happy to file it unless you would rather it went under #6485.

LukeW1999: Filed as #7356, and the answer to where it belongs turned out to be sharper than "a diagnostic somewhere".

`--replace-call-with-contract` does catch it — but only in one of the two cases, and not the one that matters. With no caller establishing the precondition, the requires is a call-site assertion and fails:

```
FAILED  [main.assertion.1]  contract requires
```

With the caller establishing it correctly — ordinary code — the assertion is discharged, the contract is assumed, and the system level is right to trust it. The obligation to check the body is the enforce phase's, and that is the phase that went vacuous. So a broken body passes both:

```c
int other; int x; int y;
void w(int v)
{
  __ESBMC_requires(other == 1);
  __ESBMC_assigns(x);
  __ESBMC_ensures(x == v);
  x = v + 1;          /* violates ensures */
  y = 99;             /* violates assigns */
}
int main(void) { other = 1; w(7); return 0; }
```

```
--function w --enforce-contract w     0 VCC(s)  VERIFICATION SUCCESSFUL
--replace-call-with-contract "*"      PASSED    VERIFICATION SUCCESSFUL
```

Root cause is that enforce starts with globals at their static initialisers, so `other == 1` is unsatisfiable at that entry even though it is satisfiable in the program. Same principle as #6484 gave pointer extents, applied to globals — which is why I filed it next to #6485 rather than inside it.

LukeW1999:
@lucasccordeiro all three threads are addressed and the branch now sits on
current master (`74d869f97f`), which brings in #7255 -- 488 lines rewritten in
the same `contracts.cpp`. The merge is textually clean, so I checked the fix is
still load-bearing rather than trusting that.

Reverted `contracts.cpp` alone to master's version on the merged tree, rebuilt,
and ran the four tests this PR adds:

```
15167 - github_7103_assigns_index_prestate_ptr             (Failed)
15168 - github_7103_assigns_index_prestate_ptr_deref_fail  (Failed)
15169 - github_7103_assigns_index_prestate_ptr_fail        (Failed)
15170 - github_7103_assigns_index_prestate_ptr_second_fail (Failed)
```

All four flip, so none of them is passing for a reason #7255 supplied. Restored
and re-run: 4/4. Full suite on the merged tree is `function_contract` 425/425.

The red `Run tests` jobs on the previous head were the 16 `sorted_key_*` cases
from #7348 (`2e9424fd8e`), inherited from master rather than caused here; the
merge should clear them.

Ready for another look when you have a moment.

PR #7207, opened 2026-08-21
[contracts] Excuse every element the clause named

### LINKED ISSUE #7184: [contracts] two array-element assigns targets on the same array collapse into one reported property, hiding a real frame check
[contracts] two array-element assigns targets on the same array collapse into one reported property, hiding a real frame check

## Summary

When a function's `__ESBMC_assigns` names two elements of the *same* array by a
symbolic index (e.g. `arr[n1], arr[n2]`), `--enforce-contract` generates two
genuinely separate witness-based frame assertions (Phase 2B) — but both are
tagged with the identical literal property name `"assigns compliance"`, no
per-target suffix. ESBMC's property reporting collapses the two into a single
reported property, so only one of the two frame checks is ever shown. A body
that violates the *other* target's frame is either wrongly reported as failing
the one that's shown, or the violation is masked entirely depending on solve
order — confirmed by testing, not just read from source.

A single array-element target (`arr[n1]` alone) verifies correctly. The
collapse only appears once a second symbolic-indexed target on the same array
is added.

## Minimal Repro

```c
void array_swap(int* arr, int n, int n1, int n2)
{
    __ESBMC_requires(__ESBMC_is_fresh(arr, n * sizeof(int)));
    __ESBMC_requires(n >= 1);
    __ESBMC_requires(n1 >= 0 && n1 < n);
    __ESBMC_requires(n2 >= 0 && n2 < n);
    __ESBMC_ensures(arr[n1] == __ESBMC_old(arr[n2]));
    __ESBMC_ensures(arr[n2] == __ESBMC_old(arr[n1]));
    __ESBMC_assigns(arr[n1], arr[n2]);

    int temp = arr[n1];
    arr[n1] = arr[n2];
    arr[n2] = temp;
}
int main() {
    int arr[5] = {1, 2, 3, 4, 5};
    array_swap(arr, 5, 0, 4);
    __ESBMC_assert(arr[0] == 5, "a0");
    __ESBMC_assert(arr[4] == 1, "a4");
    return 0;
}
```

```
$ esbmc repro.c --function array_swap --enforce-contract array_swap --no-pointer-check --unwind 10
...
  FAILED  assigns compliance: c:repro.c@16@F@array_swap@arr[j] not in assigns clause (Phase 2B)
  NOT CHECKED  contract ensures
VERIFICATION FAILED
```

This is a correct, in-frame swap of two elements the clause explicitly grants —
it should verify. Reducing the reproducer to a single target confirms the
mechanism itself works when only one array-element target is present:

```c
void set_one(int* arr, int n, int n1)
{
    __ESBMC_requires(__ESBMC_is_fresh(arr, n * sizeof(int)));
    __ESBMC_requires(n >= 1);
    __ESBMC_requires(n1 >= 0 && n1 < n);
    __ESBMC_ensures(arr[n1] == 99);
    __ESBMC_assigns(arr[n1]);
    arr[n1] = 99;
}
```
→ `VERIFICATION SUCCESSFUL` (0 of 2 properties failed).

**Declaration order does not determine which target survives.** Swapping to
`__ESBMC_assigns(arr[n2], arr[n1])` in the repro still reports a violation
against `n2` specifically — same witness expression
(`... == n2 || arr[j] == snapshot`) either way. So this is not simple
last-declared-wins; the two ASSERT instructions are both generated (confirmed
below), but the property-reporting layer only surfaces one.

## Cause

`code_contractst::materialize_arr_elem_snapshots`
(`src/goto-programs/contracts/contracts.cpp`) iterates
`classified.pointer_targets` and calls `result.push_back(entry)` once per
target — confirmed no dedup: both `arr[n1]` and `arr[n2]` produce independent
witness/snapshot symbols (`arr_elem_snap_counter` increments each time, giving
each pair of witness/snapshot symbols a unique name).

`code_contractst::emit_arr_elem_assertions`, immediately after, emits one
`ASSERT` per snapshot with a correctly distinct *guard* per target
(`(witness_j == declared_idx) || (arr[witness_j] == snapshot)`), but the
same **property name** for all of them:

```cpp
goto_programt::targett t = wrapper.add_instruction(ASSERT);
t->guard = guard;
t->location = location;
std::string arr_name = id2string(to_symbol2t(snap.arr_ptr).thename);
t->location.comment(
  "assigns compliance: " + arr_name +
  "[j] not in assigns clause (Phase 2B)");
t->location.property("assigns compliance");   // <-- identical for every target
```

`arr_name` is the same for both entries (both targets are elements of `arr`),
and the property tag is the bare literal `"assigns compliance"` — nothing
distinguishes the two per-target assertions from each other at the property
level, only their internal guards differ. The comment string is *also*
identical between the two, since it's built from `arr_name` alone, not from
the per-target witness/counter suffix already available in the enclosing
function (`cnt_str` / `arr_elem_snap_counter`).

Confirmed via a single symbolic-index target working correctly (rules out a
witness/snapshot generation bug) and via order-independence (rules out a
simple overwrite of one target by the other before this point) — the two
ASSERT instructions coexist in the GOTO program with different guards, but
identical property identity.

## Suggested Fix Direction

Give each array-element witness assertion a distinct property name/comment,
e.g. suffixing with `cnt_str` (already computed) or the declared index
expression's source text, matching how `arr_elem_snap_counter` already keeps
the witness/snapshot *symbol* names unique per target — just not surfaced to
the property tag.

## Verification

Reproduces on `3ec89ca2f0` (current `origin/master` as of 2026-08-20,
86 commits ahead of `9b547f59b1`/#7070 — not a stale-build artifact). Tested
directly, not inferred from source reading alone:
- Two-target repro above: `FAILED`.
- Single-target reduction: `VERIFICATION SUCCESSFUL`.
- Declaration order reversed (`arr[n2], arr[n1]`): still `FAILED`, same
  witness expression referencing `n2` either way.

## Related

Distinct from #6485 (pointer-extent umbrella), #7056 (global-array element
targets specifically), #7103/#7107 (index evaluated in post-state, now
fixed) — none of those touch the property-naming collision described here.
Also distinct from #7057 (`__ESBMC_old` inside a quantifier needing a
pointed-to-region snapshot) — a separate limitation hit independently while
working through the same benchmark suite.

LukeW1999:
Closes #7184, though not for the reason the issue gives — see the comment there.

## The defect

An assigns clause naming two elements of the same array grants both. Each
Phase 2B check excused only the index it was built from:

```cpp
// Guard: (j == declared_idx) || (arr[j] == snap)
expr2tc eq_idx = equality2tc(snap.witness_idx, snap.declared_idx);
```

So for `__ESBMC_assigns(arr[n1], arr[n2])`, the check built from `n1` demands
every element other than `n1` is unchanged — including `n2`, which the clause
explicitly granted. A correct swap falsifies it.

No second write is needed to trigger it. On `a082fe7d1d`, this body writes a
single named index and is still rejected:

```c
void set_first(int *arr, int n1, int n2)
{
  __ESBMC_requires(__ESBMC_is_fresh(arr, 5 * sizeof(int)));
  __ESBMC_requires(n1 >= 0 && n1 < 5);
  __ESBMC_requires(n2 >= 0 && n2 < 5);
  __ESBMC_assigns(arr[n1], arr[n2]);
  arr[n1] = 7;
}
```

```
FAILED  assigns compliance: ...@set_first@arr[j] not in assigns clause (Phase 2B)
```

The witness of the `n2` check picks `j == n1`, sees that element changed, and
reports a violation of a frame that permits it.

## The fix

`emit_array_elem_frame` in `frame_enforcer.cpp` already implements the right
rule for a global array — it ORs over every index recorded for that array. This
does the same in `emit_arr_elem_assertions`, grouped by array pointer.

## Tests

- `github_7184_assigns_two_elems` — swap of two granted elements, must verify
- `github_7184_assigns_two_elems_one_write` — one write, two named indices, must
  verify
- `github_7184_assigns_two_elems_fail` — a write to an index the clause never
  named, must still be caught, so the widened excusal does not over-permit

`regression/function_contract` 405/405.

Independent of #7206, which fixes a second Phase 2B defect on the same code
path; each branch was built and tested on its own.

https://claude.ai/code/session_01MEkhmEkbhB4tUdSnqYntPA

### DIFF
--- a/regression/function_contract/github_7184_assigns_two_elems/main.c
+++ b/regression/function_contract/github_7184_assigns_two_elems/main.c
@@ -1,0 +1,19 @@
+/* A clause naming two elements of the same array grants both. Swapping them is
+ * a write inside that frame, so it must verify -- and used to be rejected,
+ * because each element check excused only its own index (#7184). */
+void array_swap(int *arr, int n1, int n2)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(arr, 5 * sizeof(int)));
+  __ESBMC_requires(n1 >= 0 && n1 < 5);
+  __ESBMC_requires(n2 >= 0 && n2 < 5);
+  __ESBMC_assigns(arr[n1], arr[n2]);
+  __ESBMC_ensures(1);
+  int temp = arr[n1];
+  arr[n1] = arr[n2];
+  arr[n2] = temp;
+}
+
+int main()
+{
+  return 0;
+}
--- a/regression/function_contract/github_7184_assigns_two_elems/test.desc
+++ b/regression/function_contract/github_7184_assigns_two_elems/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--enforce-contract array_swap --function array_swap
+^VERIFICATION SUCCESSFUL$
--- a/regression/function_contract/github_7184_assigns_two_elems_fail/main.c
+++ b/regression/function_contract/github_7184_assigns_two_elems_fail/main.c
@@ -1,0 +1,18 @@
+/* Widening the excusal to every named index must not excuse an index the
+ * clause never named: this body writes a third element and must be caught. */
+void set_third(int *arr, int n1, int n2, int n3)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(arr, 5 * sizeof(int)));
+  __ESBMC_requires(n1 >= 0 && n1 < 5);
+  __ESBMC_requires(n2 >= 0 && n2 < 5);
+  __ESBMC_requires(n3 >= 0 && n3 < 5);
+  __ESBMC_requires(n3 != n1 && n3 != n2);
+  __ESBMC_assigns(arr[n1], arr[n2]);
+  __ESBMC_ensures(1);
+  arr[n3] = 7;
+}
+
+int main()
+{
+  return 0;
+}
--- a/regression/function_contract/github_7184_assigns_two_elems_fail/test.desc
+++ b/regression/function_contract/github_7184_assigns_two_elems_fail/test.desc
@@ -1,0 +1,5 @@
+CORE
+main.c
+--enforce-contract set_third --function set_third
+^VERIFICATION FAILED$
+^.*arr\[j\] not in assigns clause.*$
--- a/regression/function_contract/github_7184_assigns_two_elems_one_write/main.c
+++ b/regression/function_contract/github_7184_assigns_two_elems_one_write/main.c
@@ -1,0 +1,17 @@
+/* Writing one of the two granted elements is still inside the frame. This
+ * shape failed even though the body touched a single named index, because the
+ * other element's check excused only the index it was built from (#7184). */
+void set_first(int *arr, int n1, int n2)
+{
+  __ESBMC_requires(__ESBMC_is_fresh(arr, 5 * sizeof(int)));
+  __ESBMC_requires(n1 >= 0 && n1 < 5);
+  __ESBMC_requires(n2 >= 0 && n2 < 5);
+  __ESBMC_assigns(arr[n1], arr[n2]);
+  __ESBMC_ensures(1);
+  arr[n1] = 7;
+}
+
+int main()
+{
+  return 0;
+}
--- a/regression/function_contract/github_7184_assigns_two_elems_one_write/test.desc
+++ b/regression/function_contract/github_7184_assigns_two_elems_one_write/test.desc
@@ -1,0 +1,4 @@
+CORE
+main.c
+--enforce-contract set_first --function set_first
+^VERIFICATION SUCCESSFUL$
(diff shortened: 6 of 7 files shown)


LukeW1999:
Superseded by #7217, which landed the same fix on 22 Aug and does it better.

This branch excused an index by comparing index *values*, narrowing each to the
witness's type:

```cpp
equality2tc(snap.witness_idx, typecast2tc(snap.witness_idx->type, other.declared_idx))
```

#7217 compares the addresses instead, and its comment names exactly what the
narrowing above gets wrong — indices of one clause need not share a type, and
narrowing a wide one can make it alias a witnessed element and excuse it. The
`github_7184_mixed_index_types_fail` case added there covers it; nothing here
did.

Closing. The second Phase 2B defect this branch sat next to — the declared
index being read back after the body — is not fixed by #7217 and is still open
as #7103, with the fix in #7206.

PR #7222, opened 2026-08-21
[pointer-analysis] Subtract from a void pointer in the right direction

### LINKED ISSUE #7127: [c] void* backward byte arithmetic on a member address flips sign
ESBMC stores at the wrong address when `void *` byte arithmetic moves *backwards* from a member address. The offset is applied with the wrong sign: `p - k` writes at `p + k`.

```c
#include <assert.h>
struct c { int f0, f1, f2, f3, f4, f5; };
int main(void) {
  struct c s = {0,0,0,0,0,0};
  void *p = &s.f2;         /* offset 8 */
  *(int *)(p - 8) = 42;    /* should write f0, at offset 0 */
  assert(s.f0 == 0);
  assert(s.f4 == 0);       /* offset 16 == 8 + 8 */
  return 0;
}
```

`esbmc main.c --multi-property`: `s.f0 == 0` **PASSES** and `s.f4 == 0` **FAILS** — the write landed at offset 16 instead of 0. Native clang at `-O0 -fsanitize=address,undefined` runs clean and prints `f0=42 f2=0 f4=0`. Stack object, no `malloc`, default flags.

**The pointer value itself is provably correct.** ESBMC proves all three of `__ESBMC_POINTER_OFFSET(back) == 0`, `__ESBMC_same_object(back, cc)` and `back == (void *)cc` for the same expression — and then dereferences it somewhere else. The pointer model and the dereference disagree with each other.

**Localisation.** The GOTO is byte-identical between the failing and working cases — `ASSIGN *((signed int *)(p - 8))=42;` in both — and the only difference is the declared type of `p`. So the divergence is entirely downstream of the GOTO, in the dereference of a pointer whose subtype is `void`.

| Variant | Result |
|---|---|
| `void *p = &s.f2; *(int *)(p - 8) = 42;` | **writes f4** — wrong |
| `char *p = (char *)&s.f2;` — same arithmetic | writes f0 — correct |
| `void *p = (char *)s + 8;` (byte-arithmetic base, not a member address) | correct |
| `void *p = &s.f2; *(int *)(p + 8) = 42;` (forward) | correct |

So the trigger is specifically **backwards** arithmetic on a **`void *`** whose base is a **member address**.

**Severity: unsound, not a false positive.** Where the wrong address still lands inside the object the access is silently misdirected, so a real defect at the intended address is missed. Where it lands outside, it surfaces instead as a spurious `Access to object out of bounds: heap object` whose VCC degenerates to `!SAME-OBJECT(...)`.

**Why this has gone unnoticed.** `container_of` is the common way to write this, and `offsetof` is currently expanded by `clang_c_language.cpp` to `((size_t)__ESBMC_POINTER_OFFSET(&((type*)0)->member))` (#2772). That keeps the offset opaque and dodges this path — so `regression/esbmc/github_2512_*` passes *because of* the workaround, not in spite of it. Removing the expansion turns four of those tests red, which is what surfaced this. #7126 therefore drops the expansion for C++ only and documents this issue as the reason C keeps it; fixing this is the precondition for removing it there too.

lucasccordeiro:
`get_value_set_rec` applied the sign of pointer arithmetic inside the `try` block that computes the element size, but a `void *` throws out of that block at the element-size step and the handler recovering the byte offset kept only the magnitude — so `p - k` accumulated `+k` and the store landed at the mirrored address. The sign now runs once, after the handlers, where both arms share it.

Also pins R35 as KNOWNBUG: a write below an object's base is neither performed nor flagged, while the symmetric overflow is.

Fixes #7127


lucasccordeiro:
This PR (30s): 

```
  correct:           20463
    correct true:    13626
    correct false:    6837
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16139
  Score:             34073 (max: 61277)


https://github.com/esbmc/esbmc/actions/runs/32608283898
```

Master (30s):

```
  correct:           20310
    correct true:    13459
    correct false:    6851
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16292
  Score:             33753 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/31899726535
```

PR #7228, opened 2026-08-21
[solver] Compare pointer offsets the way the model reads them

lucasccordeiro:
`convert_ptr_cmp` cast both offsets to unsigned before comparing, so a pointer
one byte below its object sorted above the base and `p >= b` stayed true for
`p = b - 1`. Compare them signed, the way `__ESBMC_POINTER_OFFSET` already
reads them. An offset at or above 2^63 now mis-orders instead; pinned as R37.

Third in a stack: #7222 → #7224 → this.


lucasccordeiro:
This PR (30s): 

```
  correct:           20904
    correct true:    14036
    correct false:    6868
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15698
  Score:             34924 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32684350512.
```

Issue #5267, opened 2026-06-09
ARM64 platform limitations: --32 flag unsupported, interval analysis edge cases affected


LukeW1999:
## Background: ARM64 vs x86-64 Architecture Differences Relevant to ESBMC

Modern ARM64 (AArch64) platforms — including Linux ARM64 (e.g. Homebrew CI, AWS Graviton) and macOS Apple Silicon — differ from x86-64 in ways that affect ESBMC's functionality:

| Aspect | x86-64 | ARM64 (Linux & macOS) |
|---|---|---|
| 32-bit ABI support | Available (`-m32`, 32-bit libc headers) | **Not available** — pure 64-bit only |
| `_BitInt` width via Clang | Up to 128+ bits | **Capped at 128 bits** |
| Default `int` size | 32-bit | 32-bit (same) |
| Default pointer/`long` size | 64-bit | 64-bit (same) |
| Scalable vectors (SVE) | Not present | Present (SVE/NEON) |

> ⚠️ The 128-bit `_BitInt` cap means Solidity verification (which requires `_BitInt(256)`
> for `uint256`/`int256`) is **not supported** on ARM64. This is an upstream LLVM/Clang
> limitation. A stub is used to allow the build to succeed.

---

### 1. `--32` Flag is Non-Functional

The `--32` flag instructs ESBMC to model a 32-bit target (ILP32 ABI).
On ARM64, 32-bit libc headers (`/usr/include/i386-linux-gnu/` etc.) are not installed, so any invocation of ESBMC with `--32` fails with a parse/include error.

**Impact:**
- Verifying software that explicitly targets a 32-bit environment is not possible on ARM64.
- All regression tests using `--32` are skipped on ARM64 CI (see PR #5252).
- Users on ARM64 machines (Apple Silicon Macs, Graviton instances) will silently lose
  32-bit verification capability.

**Affected:** Linux ARM64, macOS ARM64 (Apple Silicon)

### 2. Solidity Verification Unsupported

Solidity uses `uint256`/`int256` types, which require `_BitInt(256)` in the ESBMC model.
ARM64 Clang does not support `_BitInt` wider than 128 bits.

**Current workaround:** An empty stub replaces the Solidity model on ARM64. All
`esbmc-solidity` regression tests are skipped on ARM64 CI.

**Affected:** Linux ARM64, macOS ARM64 (Apple Silicon)

### 3. Interval Analysis — Edge Cases Tied to 32-bit Width

[...]

Issue #1400, opened 2023-10-18
Regression suite not prepared for change of architecture or environment


fbrausse:
By default ESBMC is built with 32-bit and 64-bit architecture support. This works fine when the environment provides headers that work for both architectures. Such an environment is provided by many C standard libraries such as glibc. However, this breaks down when the system does not do that, for instance on aarch64 FreeBSD-14 we have:
- <limits.h> includes <sys/limits.h>
- <sys/limits.h> includes <machine/_limits.h> and defines `LONG_MAX` (etc.) to `__LONG_MAX`
- <machine/_limits.h> defines `__LONG_MAX` to the hard-coded 64-bit constant 0x7fffffffffffffffL, which is contrary to our clang frontend's internal definition of `__LONG_MAX__` to the 32-bit constant 2147483647L when run with `-m32` (as we do for `esbmc --32`).

On such a system with the clang frontend(s) we cannot reliably assume that `esbmc --32` would do anything useful. Even overriding all their standard headers wouldn't help for programs that just check "is aarch64 && is FreeBSD" and conclude "long has 64 bit". This would be a valid assumption on such a system, but it obviously fails with `-m32`.

Now, FreeBSD and other OSes that support multi-arch capable platforms such as aarch64 and amd64 usually optionally provide the 32-bit libraries and headers as an add-on package, e.g. `lib32.txz` at [https://download.freebsd.org/ftp/releases/](https://download.freebsd.org/ftp/releases/)${ARCH}/${RELEASE}. For RISC-V, however, there is no multi-arch support in FreeBSD.

As the multi-arch depends on both, the OS and on whether the platform itself supports it, I think we should enable the 32-bit *and* 64-bit support only when supported during CMake-time. This also applies for building/running ESBMC on a 32-bit-only host.

There is another problem though: the regression suite. It assumes that `--64` and `--32` work out of the box and separating them into different directories begins to clash with the separation by solvers we've done so far (e.g. there is a `z3` directory containing only Z3-tests, etc.).

Any suggestions how we could make the regression suite more modular?

fbrausse:
> Any suggestions how we could make the regression suite more modular?

Related to this question how to modularize the regression suite by multiple orthogonal parameters of the form "X is supported" where X could be
1. a particular backend solver
2. bit-width of the architecture (--32 or --64)

there additionally are issues with the regression suite on aarch64:
- incremental-smt/incremental-28 fails there with
  ```
[...]

PR #7229, opened 2026-08-21
[build] Add armv8 CI builds and an armv8 Linux release asset

### LINKED ISSUE #7236: ARM64: host a trimmed aarch64 LLVM 22 archive so armv8 matches x86_64
The x86_64 build links `clang+llvm-22.1.6-x86_64-linux-gnu-ubuntu-22.04.tar.xz`, which the project hosts on its own v8.3 release, trimmed to ~140MB. There is no aarch64 equivalent, so the ARM64 build links Ubuntu 24.04's distribution LLVM instead, which is LLVM 18.

Why not the upstream aarch64 archives:

- LLVM stopped publishing the compact `clang+llvm-*-aarch64-linux-gnu.tar.xz` after 19.x. The last one is 19.1.7.
- The `LLVM-*-Linux-ARM64.tar.xz` archive that replaced it is the full install tree: 1.65 GiB compressed, **11.01 GiB extracted** (read from the xz index, not estimated). A public `ubuntu-24.04-arm` runner has roughly 14 GB of disk, so that plus the ESBMC build tree does not fit.
- apt.llvm.org does publish llvm-22 for arm64 and `libclang-22-dev` ships the static archives ESBMC links (`libclangAST.a`, `libclangIndex.a`, `libclangTooling.a`, ~0.7 GiB installed), but wiring that into `scripts/build.sh` means the script adds a third-party APT source and GPG key to a user's machine.

Publishing a trimmed aarch64 archive on an esbmc release, built the same way the x86_64 one was, removes the asymmetry with no build-script policy change: `Options.cmake` would just gain the aarch64 URL next to the x86_64 one and drop the distro-LLVM special case in `build.sh`.

LLVM 18 is within ESBMC's supported range (`MIN_SUPPORTED_LLVM_VERSION_MAJOR`), so this is a supported configuration rather than a defect, but the two Linux assets are built against different LLVM majors until it is done. Documented in `docs/setup.md`.

Context in #7229.

Yiannis128:
`build.yml` gains an `architecture` dropdown (x86_64 / armv8); armv8 dispatches to `ubuntu-<version>-arm`, validated to only allow ubuntu-24.04/22.04. The release publishes `esbmc-linux-armv8.zip` as a fourth canonical asset, and `pull_request.yml` adds an armv8 leg for per-PR coverage.

Both Linux assets build against LLVM 22. aarch64 downloads a trimmed `LLVM-22.1.6-Linux-ARM64` archive from [`esbmc/llvm`](https://github.com/esbmc/llvm), produced by a reproducible `tools/trim.sh` that keeps the static libraries, CMake config, Clang resource headers, tblgen binaries and the shared libraries the CMake exports import. The ARM64 package installs the whole monorepo, so the trim also drops the import checks for the ~100 tool binaries it does not keep, which `find_package` would otherwise abort on. x86_64 uses its existing trimmed archive.

armv8 is a fourth entry in the release build matrix, so `create-release` gates on it exactly like Windows and macOS. A published tag missing its advertised `esbmc-linux-armv8.zip` is worse than a blocked release, and no other architecture gets an exemption. The `build-linux-arm64` Homebrew check (shared system LLVM 22, #5252) stays on weekly cron only, a distinct axis from the static leg.

172 `test.desc` files gain `REQUIRES arch32`. They pass `--32`, which reads the 32-bit libc c2goto builds, and the aarch64 Linux build turns that off. The capability now follows `ESBMC_BUNDLE_LIBC_32BIT` rather than the host CPU, so Apple Silicon, which keeps it on, still runs them. The `REQUIRES` mechanism (#1400) already skips a test on hosts lacking the named capability; `arch32` is only in the set on x86. The declaration was unused because all CI was x86_64, so these edits have zero effect on existing CI, they just prevent spurious failures on the armv8 leg. One line per file, no logic change.

aarch64 limitations (no CVC5, goto-contractor, 32-bit libc, Solidity) predate this PR and are documented with tracking issues (#5267, #7230, #7344). Solidity is now switched off there rather than left on with a stubbed model, which is what made all 512 `esbmc-solidity` tests abort; the existing CMake gate then unregisters the suite. Twenty-one further tests gain a capability `REQUIRES`. Eight use `arch_x86` or `signed_char_host`: four reach for clang's x86 intrinsic headers or x86 asm constraints, four spell `signed char` in expected output, which plain `char` is not under the AAPCS. Five more pin the x86 type ABI in their verdict rather than their output, so `signed_char_host` grows two siblings, `signed_wchar_host` and `x87_long_double`: `overflow-result3`/`4` need `100 + 100` to overflow a `char`, `github_151-2` asserts `CHAR_MIN == (char)-128`, `github_1009_success` redeclares `wchar_t` as `int`, and `udl_operator` asserts a float identity that holds in x87 80-bit and not in IEEE binary128. ESBMC answers all five correctly on aarch64, so these are test assumptions rather than defects. The last seven need the host's C++ standard library and take a new `host_cxx_headers` probe, which asks the C++ frontend for a system `<cassert>`; installing g++ is not enough, since ESBMC's clang still fails to find the GCC C++ include tree on aarch64 (#7308).

`regression/python/dataclass_initvar_post_init_string` aborts on aarch64 because the Python frontend's string model assumes a signed `char`, so it takes `REQUIRES signed_char_host` too, with #7308 tracking the defect. Deriving `config.ansi_c.char_is_unsigned` from the target instead, so `char_type()` matches what clang hands the frontend, is not the fix: it turned 16 aarch64 failures into 292, 280 of them Python tests aborting on `base_type_eq` in `value_sett::assign`, because the frontend types strings through `char_type()` but rebuilds and compares them elsewhere against a `char` it assumes is signed. That belongs in its own PR. The two `Interval Analysis (signed)` unit tests now spell `signed char` so they test their own name on every target. `nn-tanh_5_unsafe` joins the existing 2x timeout tier: 16.7s on x86_64 but over the 120s cap on aarch64, where boolector's float solving is several times slower while the rest of the suite matches x86 to within 5%. Unsupported deps (CHERI, CVC5, MathSAT) now fail configure explicitly instead of downloading an empty URL.

The armv8 asset is a statically linked binary, so it runs on any aarch64 Linux userspace without matching the host's libc, including an Android terminal.

Fixes #7236.


.github/workflows/build.yml
```diff
@@ -46,7 +59,7 @@ jobs:
     if: ${{ !startsWith(inputs.operating-system, 'windows-') }}
     uses: ./.github/workflows/build-unix.yml
     with:
-      operating-system: ${{ inputs.operating-system }}
+      operating-system: ${{ (inputs.architecture == 'armv8' && startsWith(inputs.operating-system, 'ubuntu-')) && format('{0}-arm', inputs.operating-system) || inputs.operating-system }}
```
lucasccordeiro: `ubuntu-latest` + armv8 yields `ubuntu-latest-arm`, not a valid runner label; the job queues forever. Map `ubuntu-latest` to `ubuntu-24.04-arm` or fail fast.

.github/workflows/build.yml
```diff
@@ -27,6 +35,11 @@ on:
             description: "Targeted OS (e.g. ubuntu-latest, windows-latest)"
             required: true
             type: string
+        architecture:
+            description: "Target architecture (x86_64 or armv8)"
```
lucasccordeiro, line 39: Unvalidated string: `aarch64` or `arm64` silently falls back to an x86_64 build. Fail on values other than `x86_64`/`armv8`.

website/content/_index.md
```diff
@@ -34,7 +34,8 @@ solver.
 {{< tabs >}}
 
 {{< tab name="Linux" >}} {{< cards >}}
-{{< card link="https://github.com/esbmc/esbmc/releases/latest/download/esbmc-linux.zip" title="Download" icon="download" >}}
+{{< card link="https://github.com/esbmc/esbmc/releases/latest/download/esbmc-linux.zip" title="Download (x86_64)" icon="download" >}}
+{{< card link="https://github.com/esbmc/esbmc/releases/latest/download/esbmc-linux-armv8.zip" title="Download (ARMv8)" icon="download" >}}
```
lucasccordeiro, line 39: Unlike `esbmc-linux.zip`, this asset is a shared build (`STATIC=OFF` on aarch64) and needs llvm-18, z3 and boost installed. Document the runtime dependencies.

lucasccordeiro (commented):
Only `release.yml` builds armv8, so the new ARM path is never exercised on PRs. Add an armv8 leg to `pull_request.yml`.

.github/workflows/build.yml
Yiannis128: Fixed in 175178b. The validate-platform job now rejects anything other than ubuntu-24.04 or ubuntu-22.04 for armv8, since those are the only Ubuntu images GitHub publishes an -arm variant for. Failing rather than silently mapping ubuntu-latest to a pinned version, so the caller sees which image it actually gets.

.github/workflows/build.yml
Yiannis128: Fixed in 175178b. Same job now validates the value: anything other than x86_64 or armv8 fails with an explicit error instead of falling through to an x86_64 build. The inputs are also passed via env rather than interpolated into the script body.

website/content/_index.md
Yiannis128: This was true of the first commit but no longer is: 91d2ca6 made the armv8 asset static, so it has the same runtime dependencies as esbmc-linux.zip (none). Options.cmake now selects aarch64 prebuilts (clang+llvm-19.1.7-aarch64-linux-gnu and z3-4.13.3-arm64-glibc-2.34) when CMAKE_SYSTEM_PROCESSOR is aarch64, so BUILD_STATIC works on ARM and the leg passes the same flags as x86_64 rather than -S OFF.

One asymmetry worth knowing about: the armv8 build pins LLVM 19 while x86_64 uses 22. LLVM stopped publishing the compact clang+llvm-*-aarch64-linux-gnu tarball after 19.x, and the LLVM-*-Linux-ARM64 archives that replaced it are built with ThinLTO, so their static archives hold bitcode a GCC link cannot consume. The clean fix is a trimmed aarch64 LLVM 22 tarball hosted on an esbmc release, the way clang+llvm-22.1.6-x86_64-linux-gnu-ubuntu-22.04.tar.xz already is. Happy to do that if you can publish the asset.

Yiannis128: Correction to my previous reply: the LLVM 19 pin was based on a wrong reading of LLVM's release cache, and 788362e puts aarch64 back on LLVM 22, matching x86_64.

I had claimed the LLVM-*-Linux-ARM64 archives were unusable for a static build because the release is built with ThinLTO. That is wrong. The final stage sets LLVM_ENABLE_FATLTO=ON, and CPACK_PRE_BUILD_SCRIPTS runs release_cpack_pre_build_strip_lto.cmake, which does llvm-strip -R .llvm.lto over every lib/*.a before packaging. The shipped static archives are therefore plain native objects and link fine under GCC.

The one real cost is size: no compact clang+llvm-*-aarch64-linux-gnu tarball is published past 19.x, so this is the full install tree at ~1.8GB compressed, against ~140MB for the trimmed x86_64 tarball ESBMC self-hosts. Whether that fits a public ARM runner's disk is what the new PR leg will show. If it does not, the fallback is apt.llvm.org's llvm-22 arm64 packages, which build-linux-arm64 already installs successfully on that runner.

.github/workflows/pull_request.yml
```diff
@@ -31,6 +31,17 @@ jobs:
...
+  # Goes through build.yml rather than build-unix.yml so the architecture input
+  # and its runner-label mapping are covered too, not just the ARM build.
+  build-ubuntu-armv8:
+    uses: ./.github/workflows/build.yml
+    name: static armv8 DebugOpt
+    with:
+      operating-system: ubuntu-24.04
+      architecture: armv8
+      build-flags: '-b DebugOpt -e ON'
+      testing: true
```
lucasccordeiro, line 43: `testing: true` runs the full Linux suite on ARM64, including the 159 `--32` tests this PR documents as unsupported there. Gate or exclude them.

scripts/cmake/Options.cmake
```diff
@@ -69,6 +68,15 @@ if(WIN32)
 
   set(MATHSAT_URL "https://mathsat.fbk.eu/download.php?file=mathsat-5.6.10-win64-msvc.zip")
   set(MATHSAT_NAME "mathsat-5.6.10-win64-msvc")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
```
lucasccordeiro: `arm64` also matches macOS Apple Silicon, handing Darwin these Linux/glibc LLVM and Z3 tarballs. Restrict the branch to Linux.

scripts/cmake/Options.cmake
```diff
@@ -69,6 +68,15 @@ if(WIN32)
...
+  # No compact clang+llvm-*-aarch64-linux-gnu tarball is published past 19.x,
+  # so this is the full install tree: ~1.8GB against ~140MB for the trimmed
+  # x86_64 build above, which is a real constraint on a CI runner's disk.
+  set(DEFAULT_LLVM_URL "https://github.com/llvm/llvm-project/releases/download/llvmorg-22.1.6/LLVM-22.1.6-Linux-ARM64.tar.xz")
+  set(DEFAULT_LLVM_NAME "LLVM-22.1.6-Linux-ARM64")
+
+  set(DEFAULT_Z3_URL "https://github.com/Z3Prover/z3/releases/download/z3-4.13.3/z3-4.13.3-arm64-glibc-2.34.zip")
+  set(DEFAULT_Z3_NAME z3-4.13.3-arm64-glibc-2.34)
```
lucasccordeiro, line 85: This branch leaves `DEFAULT_CVC5_URL` and `MATHSAT_URL` unset, so `-DENABLE_CVC5=On` downloads an empty URL — the failure the CHERI guard below rejects. Add the same explicit error.

.github/workflows/release.yml
```diff
@@ -50,7 +67,10 @@ jobs:
   create-release:
     name: Upload Release Asset
     runs-on: ubuntu-24.04
-    needs: [build, build-info]
+    needs: [build, build-linux-armv8, build-info]
+    # Wait for the armv8 leg so its asset is ready, but never let it hold up the
+    # release: the other platforms publish even when the ARM build fails.
+    if: ${{ !cancelled() && needs.build.result == 'success' && needs.build-info.result == 'success' }}
```
lucasccordeiro: When the armv8 leg fails, the unreviewed weekly still publishes, leaving the new front-page ARMv8 download card 404ing. Fail or warn on the missing asset.

.github/workflows/pull_request.yml
Yiannis128: Agreed, and worse than stated: 174 test.desc files pass --32 and only 2 declared REQUIRES arch32, so 172 would have run on the ARM leg. The capability already exists (regression/CMakeLists.txt only appends arch32 on x86 hosts) and was simply not being used, so a42e61e adds the REQUIRES line to all 172 rather than gating at the ctest level. No effect on x86_64, where arch32 is always present.

scripts/cmake/Options.cmake
Yiannis128: Agreed, this was a real bug. CMAKE_SYSTEM_PROCESSOR is arm64 on Apple Silicon, so the branch caught Darwin and handed it glibc tarballs. a42e61e restricts it with NOT APPLE. macOS never actually downloaded them because build.sh sets LLVM_DIR to the Homebrew prefix, but that was luck rather than design.

scripts/cmake/Options.cmake
Yiannis128: Agreed. a42e61e replaces the CHERI-only guard with one block covering CHERI, CVC5, MathSAT, and LLVM itself, each with a message naming what is unsupported and how to turn it off, gated on DOWNLOAD_DEPENDENCIES so a caller supplying its own paths is unaffected.

This also grew while addressing it: the LLVM 22 aarch64 archive is the full install tree, 1.65 GiB compressed and 11.01 GiB extracted, which does not fit a public ARM runner. aarch64 no longer downloads LLVM at all and links the distro packages via build.sh instead. Z3 keeps its arm64 prebuilt, which is small and ships libz3.a.

.github/workflows/release.yml
Yiannis128: Half agreed. The non-blocking behaviour is deliberate and I would rather keep it: a broken ARM leg should not stop the Windows, Linux and macOS assets from publishing.

But you are right that a silently absent asset against a live download link is bad. a42e61e adds a step that emits a ::warning:: in the run summary naming the armv8 leg's result when the asset is missing. Loud, not fatal. If that turns out to be too quiet in practice, the next step would be pulling the ARMv8 card off the front page until a run publishes it.

website/content/_index.md
Yiannis128: Still static after the latest change, so the original concern stays addressed, but the mechanism moved and it is worth recording here.

aarch64 now links Ubuntu's own LLVM packages rather than a downloaded prebuilt, because the LLVM 22 aarch64 archive unpacks to 11.01 GiB and a public ARM runner has around 14 GB. Ubuntu 24.04 ships LLVM 18, so the two Linux assets are built against different LLVM majors for now. 18 is ESBMC's supported minimum and the binary is still statically linked with no runtime LLVM dependency, so there is nothing for a user to install. docs/setup.md explains the tradeoff and #7236 tracks removing it by hosting a trimmed aarch64 archive next to the x86_64 one.

regression/CMakeLists.txt
```diff
@@ -74,6 +74,9 @@ endif()
 # it does not, which is the case issue #1400 was filed about.
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|amd64|AMD64|i[3-6]86)$")
     list(APPEND ESBMC_TEST_CAPABILITIES arch32)
```
lucasccordeiro, line 78: `arch32` keys on host CPU, but these tests need the bundled 32-bit libc: they pass on macOS arm64, where `ESBMC_BUNDLE_LIBC_32BIT=ON`. Gate on that variable instead.

.github/workflows/pull_request.yml
lucasccordeiro, line 43: `testing: true` makes this leg red on every PR: #7308 lists eight aarch64 failures left unskipped. Mark it `continue-on-error` until #7308 closes.

website/content/docs/setup.md
```diff
@@ -41,3 +41,32 @@ into two directories:
...
+
+`esbmc-linux-armv8.zip` is the Linux ARM64 (aarch64) build. It is static and
+self-contained like the x86_64 build, but a few features are unavailable on
+ARM64. Each is tracked by a GitHub issue; if the issue is closed, the limitation
+has been lifted and this page is simply out of date.
+
+| Limitation                                    | Tracked by |
+| --------------------------------------------- | ---------- |
+| `--32` fails: no 32-bit libc headers on ARM64 | [#5267]    |
+| Solidity unavailable: `_BitInt` caps at 128   | [#5267]    |
```
lucasccordeiro: #5267 tracks `--32` and interval analysis, not Solidity. Point this row at an issue that tracks the `_BitInt(256)` gap.

website/content/docs/setup.md
```diff
@@ -41,3 +41,32 @@ into two directories:
...
+| Interval analysis differs from x86_64         | [#5267]    |
+| CVC5 not built in                             | [#7230]    |
+| `--goto-contractor` not built in              | [#7230]    |
+
+Solidity needs `_BitInt(256)` for `uint256`, above what ARM64 Clang supports.
+
+The first three apply to macOS Apple Silicon too; the two solver gaps are
```
lucasccordeiro: `--32` works on macOS Apple Silicon with the bundled 32-bit libc; nine of ten `--32` tests pass there. Exclude `--32` from this sentence.

scripts/build.sh
```diff
@@ -177,13 +177,16 @@ prepare_platform_config() {
...
           "-DClang_DIR=/usr/lib/cmake/clang-$CLANG_VERSION"
           "-DLLVM_DIR=/usr/lib/llvm-$CLANG_VERSION/lib/cmake/llvm"
-          "-DZ3_DIR=/usr"
         )
         log "Configuring shared Ubuntu build with Clang-$CLANG_VERSION frontend"
       else
         log "Configuring static Ubuntu build"
       fi
 
+      if [[ "$STATIC" == "OFF" ]]; then
```
lucasccordeiro: This repeats the identical `STATIC == OFF` test five lines above. Move `-DZ3_DIR=/usr` back into that block.

regression/CMakeLists.txt
Yiannis128: Agreed, and the gate was wrong for a second reason: it also caught the `arch_x86` capability I had folded into the same block, which really is about host CPU. 0a71baa splits them. `arch32` now follows `ESBMC_BUNDLE_LIBC AND ESBMC_BUNDLE_LIBC_32BIT`, since what a `--32` test reads is the libc c2goto builds rather than anything on the host, and that is the option `scripts/build.sh` turns off on aarch64 Linux. Apple Silicon leaves it on, so the `--32` tests run there again instead of being skipped by the 172 `REQUIRES` lines this PR adds. `arch_x86` keeps the `CMAKE_SYSTEM_PROCESSOR` test.

.github/workflows/pull_request.yml
Yiannis128: `continue-on-error` is not accepted on a job that calls a reusable workflow. The permitted keys there are `concurrency`, `if`, `name`, `needs`, `permissions`, `secrets`, `strategy`, `uses`, `with`, and adding it fails workflow validation rather than degrading gracefully. The two existing uses in this repo (`sanitizers.yml`, `complexity.yml`) are both on plain `runs-on` jobs.

Rather than push the flag down into `build-unix.yml` as an input, which would hide the failures instead of tracking them, I have taken the leg to zero known failures so it can stay blocking.

Run 32835499608 had 16. Twelve of those are the capability skips this PR adds, one is the timeout tier for `nn-tanh_5_unsafe`, and two were the interval-analysis unit tests, now spelling `signed char` so they pin their own name on every target. That left `regression/python/dataclass_initvar_post_init_string`, and 352b727 skips it behind `signed_char_host` with #7308 tracking the defect.

Worth recording why it is a skip and not a fix. I had tried the obvious fix, deriving `char_is_unsigned` from the target so `char_type()` matches what Clang hands the frontend. Run 32936794855 shows what that costs: 292 failures, 280 of them Python, aborting on `base_type_eq` in `value_sett::assign`. The Python frontend types strings through `char_type()` but compares and rebuilds them elsewhere against a char it assumes is signed, so flipping the flag splits the two. Untangling that is its own PR. 041ecbd reverts it.

If you would still rather the leg were non-blocking while #7308 is open, say so and I will add the input.

website/content/docs/setup.md
Yiannis128: Agreed, #5267 is about `--32` and interval analysis and says nothing about `_BitInt`. Filed #7344 for the Solidity gap and c9179b4 points the row at it.

website/content/docs/setup.md
Yiannis128: Agreed. c9179b4 rewrites the sentence: Solidity and interval analysis carry over to Apple Silicon, `--32` does not, because that build keeps the bundled 32-bit libc. 0a71baa is the matching code change, so the `REQUIRES arch32` lines no longer skip those tests there.

scripts/build.sh
Yiannis128: Fixed in d4fea94.

PR #7241, opened 2026-08-22
[c++] Displace `this` onto a non-first base under a virtual base

### LINKED ISSUE #7025: [C++] Member access via this is misresolved for a non-first base with a virtual base
A class that has a virtual base resolves its **own** data members to the wrong
offset when it is used as a **non-first** base and the access happens inside one
of its member functions.

```cpp
#include <cassert>

struct V
{
  int v;
};
struct A
{
  int a;
};
struct B : virtual V
{
  char buf[8];
  void put(char c)
  {
    buf[0] = c;
  }
};
struct D : A, B
{
};

int main()
{
  D d;
  d.buf[0] = 'x';
  d.put('A');
  assert(d.buf[0] == 'A');   // holds under clang/gcc; VERIFICATION FAILED
  return 0;
}
```

`clang++ -std=c++11` runs this to completion, so the assertion is a fact about
the program, not a property that happens to be too strong. ESBMC reports
`VERIFICATION FAILED`: the write performed through `this` inside `B::put` and
the read performed directly through `d` land on different bytes.

Three conditions are jointly required — dropping any one makes the same program
verify:

| variant | verdict |
|---|---|
| `struct D : A, B` (as above) | **VERIFICATION FAILED** |
| `struct D : B, A` — the virtual-base subobject moved to offset 0 | VERIFICATION SUCCESSFUL |
| `struct B : V` — same layout, non-virtual base | VERIFICATION SUCCESSFUL |
| `d.buf[0] = 'A';` written directly instead of via `B::put` | VERIFICATION SUCCESSFUL |

So the offset applied to a member reached through the implicit `this` of a
member function does not account for the shift that the preceding base
subobject introduces, but only when the enclosing class has a virtual base.
Access from outside the class is correct, which is what makes this quiet: the
object looks right to its users and wrong to its own methods.

This is not a contrived shape — it is `std::basic_stringstream`, which derives
from both `istream` and `ostream` over a shared virtual `ios` base. Any member
of `ostream` (its character buffer, `ios_base::__width`) read from a
`stringstream` yields another field's bytes, which is why the `<sstream>`
operational model cannot store stream state on the base and has to duplicate
the insertion operators on the derived class. Singly-derived `ostringstream` is
unaffected.

ESBMC 8.4.0, master `000dc204a6`.

lucasccordeiro:
A hierarchy containing a virtual base keeps the legacy flattened layout, where a non-first base's members sit at a non-zero displacement inside the derived object, but every derived->base conversion was left unadjusted -- so a base method, its constructor and its destructor all addressed the derived object's leading storage.

The conversion is now marked at convert time and the displacement resolved in `clang_c_adjust`, which is where padding is final. The destructor previously took its offset from clang's ABI `getBaseClassOffset`, which disagrees with ESBMC's flattened layout once the primary-base rule applies; both ends of the object lifetime now use one oracle.

Eight regression tests under `regression/esbmc-cpp/inheritance/`, including two `KNOWNBUG` pins for the shapes the flattened layout still cannot express.

Fixes #7025


lucasccordeiro:
This PR (30s): 

```
  correct:           20906
    correct true:    14040
    correct false:    6866
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15696
  Score:             34930 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32635623767.
```

Master (30s):

```
  correct:           20899
    correct true:    14034
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15703

```

src/clang-c-frontend/clang_c_adjust_code.cpp
```diff
@@ -7,6 +7,29 @@
...
+// expression adjustment: it carries the derived->base displacement marker,
+// which only clang_c_adjust::adjust_expr resolves. gen_typecast leaves it
+// under the cast to the base pointer type. See #7025.
+static bool carries_derived_to_base(const exprt &arg)
+{
+  for (const exprt *e = &arg;; e = &e->op0())
+  {
+    if (!e->get("#derived_to_base").empty())
+      return true;
+    if (e->id() != "typecast" || e->operands().size() != 1)
```
XLiZHI, line 20: IIUC, this code seems to be able to match only local structures.

It can recognize:

```
  typecast
  └── typecast
      └── marked expression (#derived_to_base)
```

But it cannot recognize a marker inside a conditional expression:

```
  if
  ├── condition
  ├── marked conversion
  └── marked conversion
```

When it examines the root node, e->id() is "if", so it immediately returns false without visiting either branch.

A fix might be to inspect all operands recursively:

```
  static bool carries_derived_to_base(const exprt &expr)
  {
    if (!expr.get("#derived_to_base").empty())
      return true;

    for (const auto &op : expr.operands())
      if (carries_derived_to_base(op))
        return true;

    return false;
  }
```

PR #7243, opened 2026-08-22
[c++] Take base-subobject displacements from ESBMC's own layout

lucasccordeiro:
The thunk adapting a `Base*` receiver to an overriding method's `Derived*`, and
both arms of `dynamic_cast`, sized their pointer adjustment with clang's
`ASTRecordLayout`. ESBMC lays classes out itself, so under the Itanium
primary-base rule dispatch operated on a different object than member access.


lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/32683577762

Master (30s):

```
  correct:           20899
    correct true:    14034
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15703

```

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
```diff
@@ -1037,66 +974,24 @@ bool clang_cpp_convertert::build_dynamic_cast(
...
+      exprt result = to_derived;
+      if (!to_void)
       {
-        // For T* the result must point to the T sub-object inside D:
-        //   result = (T*)((char*)src - off(S inside D) + off(T inside D))
-        // When both offsets are zero (single inheritance with S and T at
-        // the start of D) the structural typecast is exact; otherwise the
-        // pointer is re-based through char* byte arithmetic, mirroring the
-        // (void*) arm above.
-        auto off_S = offset_of_subobject(*ASTContext, D, S);
```
XLiZHI, line 1066: This PR removed this protection, but did not add the corresponding flattened-layout reverse calculation for `adjust_base_to_derived()`.

example:
```cpp
  #include <cassert>

  struct V
  {
    virtual ~V() {}
    int v;
  };

  struct A
  {
    int a;
  };

  struct D : A, virtual V
  {
    int d;
  };

  int main()
  {
    D object;
    object.d = 42;

    V *pv = &object;
    D *pd = dynamic_cast<D *>(pv);

    assert(pd == &object);
    assert(pd->d == 42);
  }
```

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
XLiZHI: Given that this variable is not being used, shall we remove it?

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
lucasccordeiro: Good catch — this is a real defect, and it's worse than the example shows. Your program passes on this branch, but only because ESBMC's flattened layout happens to put `V` at displacement 0 inside `D`, so the unadjusted pointer is accidentally exact. Making the source base land later in the flattened struct exposes it:

```cpp
struct AA { int aa; }; struct BB { int bb; }; struct VB { int vb; };
struct ZZ { virtual ~ZZ() {} int z; };
struct D : AA, BB, ZZ, virtual VB { int d; };
// ZZ *pz = &o; D *pd = dynamic_cast<D *>(pz);
```

`pd == &o`, `pd->d`, `pd->aa` and `dynamic_cast<void *>` all FAILED: `adjust_base_to_derived` consulted only `base_subobject_offset`, so under the flattened layout it found no `@base@` component and declined.

Fixed by factoring the oracle choice out of `adjust_derived_to_base` into `base_displacement()` and using it in both directions, so the reverse computation reads the flattened layout as well. Added `github_3894_dynamic_cast_virtual_base{,_fail}`; the passing one FAILED before the fix and the `_fail` one returned SUCCESSFUL, so both discriminate.

One shape still has no fixed displacement — a virtual base shared by two sibling bases — and there the pointer stays unadjusted. I kept that at `log_debug` rather than restoring a loud failure: raising it to a warning printed five lines on a plain `<iostream>` hello-world from pre-existing thunk paths, and changing what ESBMC prints is an `esbmc-wrapper.py` interface change.

src/clang-cpp-frontend/clang_cpp_convert_vft.cpp
lucasccordeiro: Removed — it went unused when the `ASTRecordLayout` walk was dropped. Also removed from the declaration and the caller.

lucasccordeiro:
This PR (30s): 

```
  correct:           20909
    correct true:    14042
    correct false:    6867
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15693
  Score:             34935 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32743373427.
```

PR #7250, opened 2026-08-22
[svcomp] Classify a task from the properties it actually violated

lucasccordeiro:
`parse_result()` matched its comment strings against the whole ESBMC output, which since #7064 ends with a table naming every property, checked or not. A `NOT CHECKED  ... unwinding assertion loop` row turned a violation into `Unknown`, and a `NOT CHECKED ... forgotten memory` row outranked the violated free.

Replaying the 36532 per-task logs of the 2026-08-22 30s run restores 2553 correct-false verdicts (4284 → 6837); on pre-#7064 logs only two verdicts change, both from the wrong memsafety subproperty to the right one. Also revives the `free()` checks an always-true operand had made dead, and guards the CLI so the parser can be self-tested (`test_esbmc_wrapper.py`, 3 of its 6 cases fail on master).


lucasccordeiro:
This PR (30s): 

```
  correct:           20449
    correct true:    13614
    correct false:    6835
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           16153
  Score:             34047 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32578289411.
```

PR #7255, opened 2026-08-22
[contracts] Snapshot a pointed-to region for __ESBMC_old(ptr[j]) on a pointer parameter

adilanwar2399:
# [contracts] Snapshot a pointed-to region for __ESBMC_old(ptr[j]) on a pointer parameter

Partially addresses #7057. Fixes the `int r[N]` pointer-parameter shape;
the related `int (*r)[N]` pointer-to-array shape
(`github_4219_old_in_forall_ptr_to_array_knownbug`) is a separate bug in
the dereference layer and is not touched here.

## Symptom

`__ESBMC_old(ptr[j])` inside `__ESBMC_forall` aborts GOTO conversion when
`ptr` is a pointer *parameter* (as opposed to a named array):

```c
void bump(int r[N])
{
  unsigned j;
  __ESBMC_requires(__ESBMC_is_fresh(r, N * sizeof(int)));
  __ESBMC_ensures(__ESBMC_forall(&j, !(j < N) || (r[j] == __ESBMC_old(r[j]) + 1)));
  __ESBMC_assigns(r);
  for (unsigned i = 0; i < N; i++)
    r[i] = r[i] + 1;
}
```

```
ERROR: cannot model call to `c:@F@__ESBMC_old_raw' on a quantified variable
inside __ESBMC_forall/__ESBMC_exists
```

This blocks essentially any elementwise-array-transform postcondition on a
function taking its array the normal way (a pointer with a runtime-stated
extent) — doubling, incrementing by a constant, reversal, etc.

## Cause

`bound_array_element` (`goto_sideeffects.cpp`) requires an array-typed base
to lift `__ESBMC_old` over a bound index; a decayed pointer parameter is
never array-typed, so the lift declines and the raw `__ESBMC_old_raw` call
survives into `remove_sideeffects_for_quantifier_body`, which aborts on
finding a quantifier-bound-variable-mentioning side effect it can't hoist.

The deeper reason a lift-only fix can't work: `__ESBMC_old_raw`'s snapshot
mechanism (`builtin_functions.cpp`) always uses the snapshotted lvalue's own
type. For a named array that's already the array type (the working case,
#6964). For a pointer parameter there is no such lvalue — only N bytes
reachable *through* the pointer, with N stated separately in the function's
`__ESBMC_is_fresh(ptr, N)` clause, information the lift has no access to at
its point in the pipeline (ordinary GOTO conversion, well before
`contracts.cpp`'s wrapper generation runs and resolves extents).

## Fix

Two-part fix, split exactly along that boundary:

**`goto_sideeffects.cpp`** — `bound_array_element` now also recognizes a
pointer-typed, bare-symbol base reached via `dereference(+(ptr, j))` (what
`ptr[j]` actually desugars to for a decayed pointer — not `index_exprt`,
which only a genuine array produces). `lift_old_over_bound_index` snapshots
the pointer's *value* (no extent needed for that) and defers the actual
region copy, producing `dereference(add(typecast(old_raw_temp), j))`.

**`contracts.cpp`** — `collect_old_snapshots_from_body` now classifies each
snapshot as a region access when it finds that shape; `old_snapshot_t`
gained `is_ptr_region`/`region_index`/`region_elem_type`.
`materialize_old_snapshots_at_wrapper` takes the function's already-computed
`param_extents` map and, for a region snapshot, declares a real array-typed
temp sized by the pointer's `__ESBMC_is_fresh` extent and fills it with an
explicit element-wise copy loop (hand-built GOTO IR — no reusable
loop-construction helper existed in this codebase) instead of the single
whole-value `ASSIGN` the named-array case uses. `replace_old_in_expr` gained
a matching branch so `__ESBMC_old(ptr[j])` in the ensures guard resolves to
an ordinary index into that new array.

Two defensive cases, both hard-erroring with a specific diagnostic rather
than silently mis-compiling:
- No `__ESBMC_is_fresh` extent found for the pointer (or an indirect
  lvalue's extent, which isn't tracked in `param_extents`).
- The same pointer used both as a bare `__ESBMC_old(ptr)` and a region
  `__ESBMC_old(ptr[j])` in one contract (ambiguous to materialize).

**Scope, deliberately**: enforce-mode only (`--enforce-contract`); an
unconditional, direct-symbol `__ESBMC_is_fresh` only. `--replace-call-with-contract`
has no equivalent extent-tracking machinery at a call site and is left for
a follow-up.

## Verification

- `regression/function_contract/github_4219_old_in_forall_array_param`
  (promoted from `..._knownbug`): passes, 16/16 properties, both Debug and
  Release builds.
- New `_fail` sibling (`github_4219_old_in_forall_array_param_fail`,
  deliberately wrong `+2` implementation): correctly reports
  `VERIFICATION FAILED`.
- Full `function_contract` regression directory (415 tests): 412 pass,
  identically on Debug and Release; the other 3 are a shell-quoting
  artifact of the local test runner used for this check
  (`--replace-call-with-contract *` needs to reach the process unquoted,
  which a `eval`-based bash harness can't replicate without also
  glob-expanding it) — confirmed by rerunning each manually with the `*`
  properly quoted, all pass.
- Manual smoke test against a fully symbolic extent (runtime `n`, not a
  compile-time `#define` — the regression test's own shape): 19/19
  properties pass.
- Debug build specifically exercised (assertions compiled in) to rule out
  resurrecting the two crashes already documented for the *other* shape
  (`dereferencet::dereference_expr_nonscalar`,
  `assert_arith_2ops_consistency`) — confirmed clean; this fix's
  construction never dereferences an array-typed rvalue through a computed
  pointer address, which is what those crashes are specific to.

## Known follow-ups, not attempted here

- `int (*r)[N]` (pointer-to-array) — different bug, dereference layer.
- `--replace-call-with-contract` support.
- Indirect (`this->p`-style) and conditional/guarded `__ESBMC_is_fresh` —
  both fall through to the "no extent found" diagnostic rather than being
  supported.
- Separately discovered, **not fixed by or specific to this PR**: a
  conditional `__ESBMC_is_fresh` (e.g. `flag || __ESBMC_is_fresh(ptr, N)`)
  already allows dereferencing `ptr` when `flag` is true and the `is_fresh`
  side was never actually required — confirmed with a plain dereference,
  no `__ESBMC_old` involved. `--enforce-contract` mallocs the pointer
  unconditionally regardless of the guard. This region-snapshot copy loop
  inherits that existing behavior rather than worsening it; flagging in
  case it's relevant to #6485.


src/goto-programs/contracts/contracts.h
```diff
@@ -380,6 +380,15 @@ class code_contractst
...
     expr2tc snapshot_var;  ///< Snapshot variable symbol
+
+    /// True for __ESBMC_old(ptr[j]) where ptr is a pointer parameter (not a
+    /// named array): original_expr is the pointer itself, snapshot_var
+    /// becomes an array-typed copy of its is_fresh'd extent (materialized by
+    /// a copy loop, not a single whole-value ASSIGN), and region_index/
+    /// region_elem_type describe the element access. #7057.
+    bool is_ptr_region = false;
+    expr2tc region_index;     ///< nil unless is_ptr_region: the index expr
+    type2tc region_elem_type; ///< nil unless is_ptr_region: element type
```
lucasccordeiro: These two members break the Linux build: `-Werror=missing-field-initializers` at `contracts.cpp:2974`. Give them default member initializers `{}`, as `is_ptr_region` has.

src/goto-programs/contracts/contracts.h
```diff
@@ -380,6 +380,15 @@ class code_contractst
...
     expr2tc original_expr; ///< Expression inside __ESBMC_old()
     expr2tc snapshot_var;  ///< Snapshot variable symbol
+
+    /// True for __ESBMC_old(ptr[j]) where ptr is a pointer parameter (not a
+    /// named array): original_expr is the pointer itself, snapshot_var
+    /// becomes an array-typed copy of its is_fresh'd extent (materialized by
+    /// a copy loop, not a single whole-value ASSIGN), and region_index/
+    /// region_elem_type describe the element access. #7057.
+    bool is_ptr_region = false;
+    expr2tc region_index;     ///< nil unless is_ptr_region: the index expr
```
lucasccordeiro: `region_index` is written at `contracts.cpp:3021` and never read — `replace_old_in_expr` takes the index from the expression instead. Drop the field.

src/goto-programs/contracts/contracts.h
```diff
@@ -533,16 +542,23 @@ class code_contractst
...
+  ///   needed only for region snapshots
   void materialize_old_snapshots_at_wrapper(
     std::vector<old_snapshot_t> &old_snapshots,
     goto_programt &wrapper,
     const std::string &func_name,
-    const locationt &location) const;
+    const locationt &location,
+    const std::map<irep_idt, param_extentt> &param_extents) const;
 
   /// \brief Materialize old snapshots at call site (replace-call mode)
```
lucasccordeiro, line 627: `materialize_old_snapshots_at_callsite` ignores `is_ptr_region`, so `--replace-call-with-contract` now indexes a pointer-typed snapshot instead of erroring. Add an explicit diagnostic there.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2744,6 +2748,39 @@ expr2tc code_contractst::replace_old_in_expr(
...
+      if (is_symbol2t(base) && is_old_temp_symbol(base))
+      {
+        const symbol2t &sym = to_symbol2t(base);
+        for (const auto &snapshot : snapshots)
+        {
+          if (is_symbol2t(snapshot.original_expr))
+          {
+            const symbol2t &snap_sym = to_symbol2t(snapshot.original_expr);
+            if (sym.thename == snap_sym.thename)
+              return index2tc(expr->type, snapshot.snapshot_var, add.side_2);
```
lucasccordeiro: Matching only on name, without checking `snapshot.is_ptr_region`, indexes a pointer-typed `snapshot_var` when the snapshot was not classified as a region. Add the guard.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -2820,6 +2857,58 @@ expr2tc code_contractst::replace_old_in_expr(
...
+      // is dereference2t(typecast2t(old-temp-symbol)), not just
+      // typecast2t(old-temp-symbol).
+      expr2tc base = to_add2t(ptr_expr).side_1;
+      if (is_dereference2t(base))
+        base = to_dereference2t(base).value;
+      while (is_typecast2t(base))
+        base = to_typecast2t(base).from;
+      if (is_symbol2t(base) && to_symbol2t(base).thename == old_temp_thename)
+      {
+        out_index = to_add2t(ptr_expr).side_2;
```
lucasccordeiro: This shape also matches a user-written `*(__ESBMC_old(p) + i)` outside any quantifier, reclassifying a pointer-value snapshot as a region and aborting when `p` lacks `is_fresh`. Restrict to lift-produced uses.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3958,134 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
+        {
+          log_error(
+            "{}: __ESBMC_old(...[...]) region snapshot expected a bare "
+            "pointer parameter symbol (#7057)",
+            func_name);
+          abort();
+        }
+        const irep_idt &ptr_thename = to_symbol2t(original_expr).thename;
+        auto extent_it = param_extents.find(ptr_thename);
+        if (extent_it == param_extents.end() || !extent_it->second.justified)
```
lucasccordeiro: `justified` is also true for the #6483 one-element struct stack backing, so a struct-pointer parameter with no `is_fresh` silently gets a 1-element snapshot rather than this diagnostic. Require an is_fresh-sourced extent.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3958,134 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
 
-      // Generate snapshot assignment: new_snapshot_var = original_expr
-      goto_programt::targett assign_inst = wrapper.add_instruction(ASSIGN);
-      assign_inst->code = code_assign2tc(new_snapshot_var, original_expr);
-      assign_inst->location = location;
-      assign_inst->location.comment("__ESBMC_old snapshot assignment");
+        const type2tc &elem_type = old_snapshots[i].region_elem_type;
+        type2tc idx_type = size_type2();
+
+        BigInt elem_sz = type_byte_size(elem_type, &ns);
```
lucasccordeiro: Missing the `elem_sz > 0` guard the sibling extent division at line 3812 has; a zero-size element type divides by zero here. Add it.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3958,134 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
+
+        goto_programt::targett i_decl = wrapper.add_instruction(DECL);
+        i_decl->code = code_decl2tc(idx_type, to_symbol2t(i_sym).thename);
+        i_decl->location = location;
+
+        goto_programt::targett i_init = wrapper.add_instruction(ASSIGN);
+        i_init->code = code_assign2tc(i_sym, gen_zero(idx_type));
+        i_init->location = location;
+
+        // i = 0; while (i < n_elems) { snap[i] = *(ptr + i); i = i + 1; }
```
lucasccordeiro: This loop is unbounded when the extent is symbolic, so `--unwind` truncates it. Add the symbolic-extent smoke test from the description as a regression test.

src/goto-programs/goto_sideeffects.cpp
```diff
@@ -136,15 +136,42 @@ bound_array_element(const exprt &addr, const std::set<irep_idt> &bound_vars)
...
+  if (target.id() == exprt::index && target.operands().size() == 2)
+  {
+    if (
+      !mentions_symbol(target.op1(), bound_vars) ||
+      !target.op0().type().is_array())
+      return nullptr;
+    return &target;
+  }
+
+  // Pointer-parameter case: a decayed pointer's p[j] desugars at the
```
lucasccordeiro: The Doxygen comment above (line 127) still says a pointer base "is left alone"; that is no longer true. Update it.

lucasccordeiro (changes requested):

adilanwar2399:
Thanks for the thorough review — addressed all nine points, plus one more the same testing surfaced. Pushed as three commits (build fix + complexity, review fixes, format).

**Real bugs:**

- **Replace-mode silently miscompiling** — `materialize_old_snapshots_at_callsite` now checks `is_ptr_region` and hard-errors with a specific diagnostic instead of indexing a pointer-typed snapshot. `--replace-call-with-contract` support for this shape stays a documented follow-up.
- **Legacy bare-symbol match not checking `is_ptr_region`** — guarded, so it can no longer return an array-typed `snapshot_var` where a scalar is expected.
- **False-positive region matching** (`*(__ESBMC_old(p) + i)` outside a quantifier) — `find_ptr_region_use` now only matches inside a `__ESBMC_forall`/`__ESBMC_exists` body, since the lift can only ever fire there. Verified against your exact example.
- **One more, found verifying the above**: `try_replace_ptr_region_old` had the same gap — matched by pointer name alone, no `is_ptr_region` check — and still crashed (`dereferencet::dereference_expr_nonscalar`) on your counter-example even after the classifier was fixed. Same guard applied; the repro now matches unmodified `master`'s behavior exactly (clean `VERIFICATION FAILED`, no crash).
- **`justified` conflating two sources** — added `from_is_fresh`, set only by the genuine `__ESBMC_is_fresh` producer, and required alongside `justified`. A struct-pointer parameter with no `is_fresh` now correctly hits the "no extent" diagnostic instead of silently getting the #6483 carve-out's one-element backing.
- **Missing `elem_sz > 0` guard** — added, matching the sibling division.

**Hygiene:**

- Dropped `region_index` (dead — both consumers already have their own index expression in hand).
- Added the symbolic-extent case as a real regression test (`github_4219_old_in_forall_array_param_symbolic_extent`) instead of only a manual smoke test.
- Fixed the stale "is left alone" doc comment.
- Also extracted `find_snapshot_by_symbol` as a shared helper for the two by-name lookup paths that remained, to keep `replace_old_in_expr` under the complexity gate now that it does more.

All nine repros (ambiguity guard, no-extent fallback, replace-mode diagnostic, the struct-carve-out precision case, and the hand-written-outside-quantifier false positive) individually re-verified, plus a full `function_contract` regression run on both Debug and Release.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3992,148 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
+          !extent_it->second.from_is_fresh)
+        {
+          log_error(
+            "{}: __ESBMC_old({}[...]) needs its pointer parameter's extent "
+            "stated by an unconditional, direct __ESBMC_is_fresh({}, N); "
+            "none found (#7057)",
+            func_name,
+            id2string(ptr_thename),
+            id2string(ptr_thename));
+          abort();
```
lucasccordeiro: No regression test reaches this abort. Add one: `__ESBMC_old(ptr[j])` on a pointer parameter whose contract states no `__ESBMC_is_fresh`.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3846,6 +4163,24 @@ code_contractst::materialize_old_snapshots_at_callsite(
...
+    // cleanly instead; --replace-call-with-contract support is a follow-up
+    // (#7057).
+    if (old_snapshots[i].is_ptr_region)
+    {
+      log_error(
+        "{}: __ESBMC_old(...[...]) on a pointer parameter is not yet "
+        "supported under --replace-call-with-contract (#7057)",
+        id2string(function_symbol.name));
+      abort();
+    }
```
lucasccordeiro, line 4217: This replace-mode rejection is untested. Add a `--replace-call-with-contract` test asserting the diagnostic, so the path cannot silently regress.

src/goto-programs/contracts/contracts.h
```diff
@@ -405,6 +440,20 @@ class code_contractst
     const expr2tc &expr,
     const std::vector<old_snapshot_t> &snapshots) const;
 
+  /// \brief Does this old-temp's hoisted symbol appear under a
+  /// pointer-region dereference shape somewhere in the body? Sets
+  /// is_ptr_region/region_index/region_elem_type on each entry that does.
```
lucasccordeiro: `region_index` was dropped from `old_snapshot_t`; this is its last mention in `src/`. Remove it from the comment.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3846,6 +4163,24 @@ code_contractst::materialize_old_snapshots_at_callsite(
...
+    // valid_object assertion (lower_is_fresh_in_requires). Silently
+    // proceeding would create a pointer-typed snapshot_var that
+    // replace_old_in_expr's region match then tries to index. Reject
+    // cleanly instead; --replace-call-with-contract support is a follow-up
+    // (#7057).
+    if (old_snapshots[i].is_ptr_region)
+    {
+      log_error(
+        "{}: __ESBMC_old(...[...]) on a pointer parameter is not yet "
+        "supported under --replace-call-with-contract (#7057)",
```
lucasccordeiro, line 4214: None of the five new `abort()` diagnostics has a regression test, this one included. Manual re-verification does not stop them regressing; add tests asserting the messages.

adilanwar2399:
Thanks — addressed all four:

- Removed the stale `region_index` reference from `classify_ptr_region_snapshots`'s comment.
- Added regression tests for 4 of the 5 `abort()` diagnostics, each asserting the actual message: `github_4219_old_in_forall_array_param_no_extent_fail` (no `is_fresh` extent), `github_4219_old_in_forall_array_param_replace_mode_fail` (the replace-mode rejection), `github_4219_old_in_forall_array_param_zero_elemsz_fail` (a zero-size element type via an empty struct), `github_4219_old_in_forall_array_param_ambiguous_fail` (the same pointer as both a bare and a region snapshot).

The fifth — `materialize_old_snapshots_at_wrapper`'s "region snapshot expected a bare pointer parameter symbol" check — I couldn't write a real test for, and want to flag why rather than fake one: `bound_ptr_param_offset` in `goto_sideeffects.cpp` (the lift that produces the region shape in the first place) already requires the base to be a bare symbol before it marks anything as a region access. So by the time this check runs, `original_expr` is structurally guaranteed to already be a symbol — I can't find any C source that reaches it. It looks like a defensive invariant check for a case the upstream gating already rules out, not a live path. Happy to remove it as dead code if you'd rather, or leave it documented as defensive — whichever you prefer.

Separately, unrelated to the review: caught that my branch had fallen 25 commits behind `origin/fix/7057-old-forall-pointer-region` (missing the merge-master push) partway through this — pulled up to date, rebuilt, and re-ran everything (all 4 new tests, the original target test, and the three downstream ConVer benchmark files this fix unblocks) against the current state before pushing. Full `function_contract` regression: 417/420, same 3 pre-existing local-runner artifacts already noted in earlier comments.

lucasccordeiro (commented):
`regression/function_contract/github_4219_old_in_forall_ptr_to_array_knownbug/main.c:26` still points at `github_4219_old_in_forall_array_param_knownbug`, which this PR removes. Update the reference.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3992,148 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
+            id2string(ptr_thename));
+          abort();
+        }
+        const expr2tc &bytes = extent_it->second.bytes;
+        expr2tc n_elems = typecast2tc(
+          idx_type,
+          div2tc(bytes->type, bytes, constant_int2tc(bytes->type, elem_sz)));
+        simplify(n_elems);
+
+        type2tc arr_type = array_type2tc(elem_type, n_elems, false);
```
lucasccordeiro: A struct element type keeps its unfollowed `symbol` subtype here, so symex aborts with `struct_union_packed() called on incompatible type (type_id = symbol)`. Follow the element type before building the array.

src/goto-programs/contracts/contracts.cpp
```diff
@@ -3801,22 +3992,148 @@ void code_contractst::materialize_old_snapshots_at_wrapper(
...
-        original_expr, func_name + "_wrapper", unique_snapshot_count++);
+      const size_t snap_idx = unique_snapshot_count++;
 
-      // Generate snapshot declaration
-      goto_programt::targett decl_inst = wrapper.add_instruction(DECL);
-      decl_inst->code = code_decl2tc(
-        original_expr->type, to_symbol2t(new_snapshot_var).thename);
-      decl_inst->location = location;
-      decl_inst->location.comment("__ESBMC_old snapshot declaration");
+      if (old_snapshots[i].is_ptr_region)
```
lucasccordeiro, line 4142: This ~110-line branch of hand-built loop IR makes `materialize_old_snapshots_at_wrapper` do two unrelated jobs. Extract the region case into its own helper.

adilanwar2399:
Thanks — all three addressed in 2156be3240:

1. **`struct_union_packed()` crash**: root-caused to a dead residual copy of the ensures expression that `goto_sideeffects.cpp`'s `lift_old_over_bound_index` builds directly in the renamed original function body — it never passes through `contracts.cpp`'s own `follow()` calls, so it kept the struct element's raw, unfollowed `symbol_typet`. Fixed by threading a `namespacet` through the lift and following the element type there before building the dereference node. Confirmed via rebuild+retest: the crash is gone.

   One caveat I want to flag rather than paper over: with the crash fixed, a struct-element `__ESBMC_old(ptr[j]).field` region snapshot still surfaces a genuine `contract ensures` *failure* (not a crash) — traced to the pre-existing #6483 limitation (`__ESBMC_is_fresh` heap-backs struct pointers as raw byte buffers, and the dereference-safety checker sees a real type mismatch once the comparison is actually exercised). A tautology control without `__ESBMC_old` on the identical setup passes cleanly, which is consistent with #6483 rather than a new bug from this PR. Full struct-element `__ESBMC_old` region support looks like it needs #6483 fixed first — happy to open a tracking issue if useful.

2. **Extracted the region branch**: pulled the ~110-line hand-built copy loop out of `materialize_old_snapshots_at_wrapper` into its own `materialize_ptr_region_old_snapshot` method (declared in `contracts.h`). Confirmed behavior-identical via rebuild+retest against the same repro files.

3. **Stale comment**: fixed — `github_4219_old_in_forall_ptr_to_array_knownbug/main.c:26` now points at `github_4219_old_in_forall_array_param` (its current name after promotion to CORE) instead of the old `_knownbug` name.

All 420 `function_contract` regression tests pass locally (macOS build; this suite isn't wired into the Linux-only CMake regression list on this platform, so I ran `testing_tool.py` directly per-test rather than via ctest.
EOF
)

PR #7261, opened 2026-08-23
alias numpy column, diagonal, ravel, and flat views via strided pointers

sergillam:
- **2-D column views** (`a[:, j]`) via strided pointer
- **1-D slices with step≠1 / reversed** (`a[::2]`, `a[::-1]`)
- **`np.diagonal(a, offset=k)` / `a.diagonal()`** — read-only strided
  pointer view; a nested/nowhere-assigned use (e.g. `np.diagonal(a)[i]`)
  declines explicitly instead of silently returning a wrong value
- **`np.trace()` / `np.fill_diagonal()`** — reuse the same offset/stride
  math as a scalar reduction and an in-place mutation, without needing a
  view of their own; `fill_diagonal`'s value must match the diagonal's
  length exactly (matches NumPy's own broadcasting rule — no
  repeat/truncate)
- **`np.ravel(a)` / `a.ravel()` / `a.flat[i]`** — writable, contiguous
  pointer view for fixed-shape 1-D/2-D arrays; both directions
  (read and write) observe live mutations of the source

- Reads, writes, `len()`, `.shape`, and `.ndim` all observe the real
-  aliasing for every case above. Existing regression tests that asserted
- the old copy-divergence rejection (`view_strided_write_*_fail`,
- `flat_mutation_diagnostic_fail`, `copy_unsupported_view_like_fail`) were


src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -3059,6 +3150,137 @@ bool python_converter::is_numpy_ravel_receiver(
...
+// -- a directly-written a.ravel('F')[i] or np.ravel(a, order='F')[i] must
+// not match flat_subscript_receiver_node. The `.flat` rewrite that
+// function's main target never carries an order argument itself (`.flat`
+// always iterates C-order), so this only ever declines a genuine
+// explicit-order ravel call.
+bool ravel_call_has_order_arg(const nlohmann::json &value_node)
+{
+  if (
+    value_node.contains("args") && value_node["args"].is_array() &&
+    value_node["args"].size() > 1)
```
lucasccordeiro: The method form `a.ravel('F')[i]` carries order in args[0], so `size() > 1` never fires and it aliases C-order. Count args per receiver form.

src/python-frontend/python-list/list_access.cpp
```diff
@@ -1472,18 +1655,57 @@ const symbolt &python_list::get_str_slice_sym()
...
+  // computes a one-past-the-end address, legal to form but not to
+  // dereference (C 6.5.6p8). index2tc would route it through the array
+  // bounds checker as an out-of-bounds subscript; add2tc over the decayed
+  // pointer is checked under pointer, not array-index, rules.
+  exprt base_ptr = build_typecast(build_address_of(array), view_ptr_type);
+  exprt view_ptr =
+    build_add(base_ptr, from_integer(offset, size_type()), view_ptr_type);
+  converter_.current_lhs->type() = view_ptr_type;
+  converter_.update_symbol(*converter_.current_lhs);
+  converter_.numpy_pointer_view_info_[converter_.current_lhs->identifier()
```
lucasccordeiro: Two views bound to the same name on different branches overwrite each other here, so indexing uses the last stride/length on both paths. Decline on re-registration.

src/python-frontend/converter/converter_stmt.cpp
```diff
@@ -2070,13 +2130,21 @@ void python_converter::detach_numpy_pointer_views_of(
...
-    // storage at this point in program order) into the snapshot.
+    // storage at this point in program order) into the snapshot. The
+    // source read is scaled by the view's own stride (1 for a unit-stride
+    // slice/row view, but e.g. num_cols for a column view); the snapshot
+    // itself is always densely packed, so the destination index is not.
     for (std::size_t i = 0; i < length; ++i)
     {
-      exprt idx = from_integer(i, size_type());
-      exprt src = python_expr::build_index(old_ptr, idx, elem_type);
+      // Commit 2 only ever sees stride >= 1 (unit-stride slices, row
```
lucasccordeiro: Stale: reverse slices land in this PR, so this loop does see negative strides. Fix the comment and add a rebind-detach regression for `a[::-1]`.

PR #7269, opened 2026-08-23
[util] Fold a relation between two boolean constants

lucasccordeiro:
`simplify_relations` dispatched the constant fold on the operand type with no
`bool` case, so `true != false` stood unfolded and symex could decide neither
branch of a guard over it, unwinding JBMC's `clinit_wrapper` ~1000 times. The
operand dispatch moves into its own function to stay under the complexity gate.

### DIFF
--- a/docs/roadmap/jbmc-goto-binary-poc-plan.md
+++ b/docs/roadmap/jbmc-goto-binary-poc-plan.md
@@ -963,6 +963,32 @@ programs that *use* strings, not to every program that merely links
 to fail and JBMC does not. That is a known pointer-model difference (§5), not a
 JVM-semantics gap.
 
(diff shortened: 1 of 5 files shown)


lucasccordeiro:
This PR (30s): 

```
  correct:           20914
    correct true:    14047
    correct false:    6867
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15688
  Score:             34945 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32683630308
```

Master (30s):

```
  correct:           20899
    correct true:    14034
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15703

```

PR #7281, opened 2026-08-23
[python] Extend dynamic typing to tagged arithmetic and elif chains

brcfarias:
Local-variable divergence detection only recognized a plain two-branch if/else, so an elif chain was invisible to it, and this wasn't just a coverage gap. It actually broke the join. `if c1: x = 1 elif c2: x = 2 else: x = "a"` left `x` an untagged int on the `c1` path, and only got tagged once the converter recursed into the elif's own if/else. Detection now recurses through the whole elif chain, requiring every branch to assign the name for it to qualify. `-` and `/` between two tagged operands (e.g. `x - y` where both diverge between int and str across the same if/else) now work, dispatching to new C models that read both operands' type_id/value directly and raising a TypeError when either side turns out non-numeric.

### DIFF
--- a/regression/python/dynamic_type_elif_dangling_no_tag/main.py
+++ b/regression/python/dynamic_type_elif_dangling_no_tag/main.py
@@ -1,0 +1,9 @@
+# No final else, so x should stay a plain int, not get tagged.
+cond1 = nondet_bool()
+cond2 = nondet_bool()
+if cond1:
+    x = 1
+    assert x == 1
+elif cond2:
+    x = 2
+    assert x == 2
--- a/regression/python/dynamic_type_elif_dangling_no_tag/main.py
+++ b/regression/python/dynamic_type_elif_dangling_no_tag/main.py
@@ -1,9 +1,16 @@
-# No final else, so x should stay a plain int, not get tagged.
+# No final else, so x/y should stay plain scalars, not get tagged. x + y
+# would fail to convert if both got tagged (Add between two tagged
+# operands isn't supported), so this pins the untagged case rather than
+# checking a value that would hold either way.
 cond1 = nondet_bool()
 cond2 = nondet_bool()
 if cond1:
     x = 1
-    assert x == 1
+    y = 10
+    z = x + y
+    assert z == 11
 elif cond2:
-    x = 2
-    assert x == 2
+    x = "a"
+    y = "b"
+    z = x + y
+    assert z == "ab"
--- a/regression/python/dynamic_type_scalar_tag_div_dyn/main.py
+++ b/regression/python/dynamic_type_scalar_tag_div_dyn/main.py
@@ -1,0 +1,9 @@
+cond = nondet_bool()
+if cond:
+    x = 10
+    y = 2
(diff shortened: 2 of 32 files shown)


regression/python/dynamic_type_elif_dangling_no_tag/main.py
```diff
@@ -0,0 +1,9 @@
+# No final else, so x should stay a plain int, not get tagged.
+cond1 = nondet_bool()
+cond2 = nondet_bool()
+if cond1:
+    x = 1
+    assert x == 1
```
lucasccordeiro: `assert x == 1` holds whether or not `x` is tagged, so this pins nothing. Add a second diverging name and `x + y`, which errors only if both got tagged.

regression/python/dynamic_type_scalar_tag_div_dyn/main.py
```diff
@@ -0,0 +1,9 @@
+cond = nondet_bool()
+if cond:
+    x = 10
+    y = 2
+    z = x / y
```
lucasccordeiro, line 5: `/` has no TypeError counterpart to `dynamic_type_scalar_tag_sub_dyn_typeerror`, so the div model's assert and the zero-guard's `rhs->value` deref are never exercised on a str operand. Add one.

src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -611,6 +668,72 @@ exprt dynamic_type_handler::handle_arithmetic(
...
+  // operands are numeric.
+  exprt rhs_numeric_value = build_dereference(
+    build_typecast(
+      build_member(rhs, "value", pointer_typet(empty_typet())),
+      pointer_typet(signedbv_typet(config.ansi_c.long_long_int_width))),
+    signedbv_typet(config.ansi_c.long_long_int_width));
+  exprt is_zero = build_equal(
+    rhs_numeric_value, from_integer(BigInt(0), rhs_numeric_value.type()));
+
+  exprt raise = converter_.get_exception_handler().gen_exception_raise(
```
lucasccordeiro, line 712: This ZeroDivisionError guard is a verbatim copy of `build_div_literal`'s. Extract the deref/is_zero/raise/add_instruction block into a shared helper.

src/python-frontend/dynamic_type/dynamic_type_handler.cpp
```diff
@@ -76,6 +76,44 @@ classify_branch_literal_assigns(const nlohmann::json &block)
   return types;
 }
 
+// Recursively collects, per name, the literal kinds an elif chain's leaves
+// assign it. Returns false for a dangling elif with no final else.
+bool collect_orelse_literal_kinds(
```
lucasccordeiro: Only `orelse` is recursed, so `if c1: (if c2: x=1 else: x=2) else: x="a"` stays undetected. Recurse the `then` body too, or narrow the header's "every branch" claim.

lucasccordeiro (changes requested):

PR #7283, opened 2026-08-23
[pointer-analysis] Do not ask whether a union is a subclass

lucasccordeiro:
`value_sett::assign` includes unions in its aggregate branch and then asks
`is_subclass_of`, which casts both operands with `to_struct_type` and throws.
Inheritance has no union analogue, so treat a non-`base_type_eq` union pair as
incompatible, exactly as the branch already does for a mismatched `type_id`.

Refs #4715


lucasccordeiro:
This PR (30s): 

```
  correct:           20898
    correct true:    14034
    correct false:    6864
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15704
  Score:             34916 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32767823828.
```

PR #7292, opened 2026-08-24
[c++] Find a base-adjustment marker anywhere in a call argument

lucasccordeiro:
`carries_base_adjustment` gates whether a call-statement argument reaches
`adjust_expr`, and it walked only the `op0()` typecast chain, so a marker
sitting anywhere else went unresolved and left the pointer on the base
subobject silently. Scan every operand instead, so the gate does not depend on
the shape today's producers happen to have.

No producer nests one today: an instrumented build flagging arguments where the
recursive scan finds a marker the linear walk misses reported zero over 2007
files of the C++ corpus, so this is hardening with a measured zero delta.

Addresses the review comment on #7241. Based on #7243's branch, which renamed
this predicate; merge after it.


lucasccordeiro:
This PR (30s): 

```
  correct:           20911
    correct true:    14046
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15691
  Score:             34941 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32770895033.
```

PR #7294, opened 2026-08-24
[build] Static aarch64 Linux build via trimmed LLVM 22 archive

### LINKED ISSUE #7236: ARM64: host a trimmed aarch64 LLVM 22 archive so armv8 matches x86_64

Yiannis128:
The x86_64 Linux release links a self-hosted, trimmed LLVM 22 archive so `esbmc-linux.zip` has no runtime LLVM dependency. No equivalent existed for aarch64: upstream only publishes an 11 GB full install (too big for a CI runner), so the arm64 build had nothing static to link against and `esbmc-linux-armv8.zip` could not be produced.

This adds the missing half in two repos:

**esbmc/llvm** (new): `tools/trim.sh` deterministically trims an upstream LLVM release tarball to the static archives, CMake package config, Clang resource headers, public headers and tblgen binaries that ESBMC's static link consumes. A `workflow_dispatch` workflow runs it on `ubuntu-24.04-arm` and publishes a tagged release asset. The method is committed, so the archive is reproducible rather than hand-made (the x86_64 one was trimmed locally with no recorded steps).

**esbmc/esbmc** (this PR): `Options.cmake` points Linux aarch64 at the trimmed `LLVM-22.1.6-Linux-ARM64-esbmc.tar.xz` from `esbmc/llvm` so the static link resolves against per-component `.a` files instead of a distro `libLLVM.so`. `release.yml`'s arm64 job drops the apt.llvm.org shared build and exercises this static path directly. Both Linux assets now build against LLVM 22.

Fixes #7236. Blocks #7229 (the armv8 asset PR cannot ship a working static binary without this).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,18 +31,12 @@ jobs:
       canon-names: true
 
   build-linux-arm64:
-    name: Homebrew compatibility check (Linux ARM64, LLVM 22)
+    name: Static Linux ARM64 build
     runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
-      - name: Install LLVM 22
-        run: |
-          wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-          echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-22 main" | sudo tee /etc/apt/sources.list.d/llvm.list
-          sudo apt-get update
-          sudo apt-get install -y llvm-22 clang-22 libclang-22-dev libclang-cpp22-dev
-      - name: Build ESBMC
-        run: ./scripts/build.sh -b RelWithDebInfo -S OFF -c 22 -e OFF
+      - name: Build ESBMC (static, downloaded LLVM 22)
+        run: ./scripts/build.sh -b RelWithDebInfo -e OFF
 
   build-info:
     uses: ./.github/workflows/build-info.yml
--- a/scripts/cmake/Options.cmake
+++ b/scripts/cmake/Options.cmake
@@ -59,7 +59,9 @@ option(CORE_REGRESSION_ONLY "Only add tests in the regression that are CORE (def
 # PRE-BUILT DEPENDENCIES
 #############################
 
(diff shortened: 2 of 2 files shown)


Yiannis128:
Superseded by #7229, which now incorporates the trimmed LLVM 22 aarch64 archive wiring directly. The esbmc/llvm repo (trim script + workflow) and its `llvm-22.1.6-armv8` release asset remain the deliverable for the prebuilt toolchain.

PR #7295, opened 2026-08-24
[goto-coverage] Name the dead direction in --dead-code-check advisories

lucasccordeiro:
`--dead-code-check` named the guard of the *live* branch. A branch-coverage
probe is `assert(c)`, so a probe left unviolated proves `!c` infeasible — the
dead direction is `!c`, but `report_dead_code()` printed `c`.

Record each claim's negation from the guard expression at instrumentation time
and report that. Deriving it from the expression rather than flipping the
comment text matters: `from_expr` parenthesises by precedence, so `!flag`
negates to `flag` while `a > b` negates to `!(a > b)`.


pmatos:
@lucasccordeiro You were too fast, I was just finishing a review.

PR #7301, opened 2026-08-25
[goto-coverage] Drop the unguarded lookup in report_dead_code

pmatos:
Follow-up to #7295 (already merged). Review of that PR turned up one
robustness defect and two inaccurate rationale comments.

### `report_dead_code` lookup

`report_dead_code` iterated `goto_coveraget::all_claims` and then indexed
`goto_coveraget::claim_negation` with `std::map::at()`. The two are built in
lockstep in `get_total_cond_assert()`, so the key is present today — but the
invariant is only documented in a comment, and `at()` turns any future drift
into an uncaught `std::out_of_range` (ESBMC terminating instead of printing a
report) rather than a degraded advisory.

Iterating `claim_negation` directly keys the loop off a single container. The
map's key type and comparator are identical to the set's and the key set is
`all_claims` by construction, so **iteration order and report output are
unchanged**; it also drops a per-claim temporary `pair<string,string>` copy and
a redundant O(log n) lookup.

### Comment corrections

- `goto_coverage.h`: the note said `from_expr` parenthesisation is why `!flag`
  negates to `flag`. Parenthesisation only explains the `a > b` → `!(a > b)`
  half; `!flag` → `flag` is `gen_not_expr` (goto_coverage.cpp:1222) stripping a
  leading `not2t`.
- `cwe_dead_code_guard_polarity/main.c`: the mutation note claimed flipping the
  probe's comment text would print `!flag`. A naive text flip of the surviving
  probe's comment `!flag` yields `!(!flag)`; `!flag` is what the *old* code
  printed by not flipping at all.

The test comment is held to exactly three lines so the `main.c:10` /
`main.c:13` pins in `test.desc` still resolve.

### Verification

- `git clang-format --diff upstream/master` — no changes to the touched lines.
- `-fsyntax-only -Wall -Wextra` using the real `compile_commands.json` command
  for `bmc.cpp` — clean, no warnings. This exercises the nested structured
  binding over `std::pair<const Key, T>`, the only real risk in the edit.

No local test run: this tree's build directory predates master by a month, so
CI is the verification. Behaviour is output-identical by construction, and
`cwe_dead_code_guard_polarity`, `python/dead_code_guard_polarity` and
`github_7267-genuine-dead-code` are the tests that cover it.

Follow-up issues for the findings not addressed here: #7302 (k-path guard is
double-negated), #7303 (claim_negation duplicates all_claims), #7304
(const getter rebuilds global state), #7305 (SARIF polarity untested).

### DIFF
--- a/regression/esbmc/cwe_dead_code_guard_polarity/main.c
+++ b/regression/esbmc/cwe_dead_code_guard_polarity/main.c
@@ -1,8 +1,8 @@
 extern int __VERIFIER_nondet_int(void);
 
 // Pins both polarities of the guard named in a dead-code advisory. `flag` also
-// pins that the negation is derived from the expression: flipping the probe's
-// comment text would print "!flag" here.
+// pins expression-level negation: the surviving probe's comment is "!flag", so
+// printing it gives "!flag" and wrapping it in !(...) gives "!(!flag)".
 int main(void)
 {
   _Bool flag = 0;
--- a/src/esbmc/bmc.cpp
+++ b/src/esbmc/bmc.cpp
@@ -1028,10 +1028,11 @@ static bool is_kpath_maximal(const std::string &claim_sig)
 // instrumented assertion over a branch guard) that multi_property_check never
 // violated proves `!c` infeasible up to the current unwinding bound — so `!c`
 // is the dead direction, and the advisory names goto_coveraget::claim_negation
-// rather than the claim's own comment. The dead set is therefore
-// `all_claims \ reached_claims`. Findings are advisory: they are printed as a
-// separate [Dead code] section and, when --sarif-output is set, emitted at
-// SARIF note level. They never flip the verdict (see report_result).
+// rather than the claim's own comment. claim_negation is keyed by, and rebuilt
+// with, goto_coveraget::all_claims, so the dead set is exactly the entries
+// below that reached_claims does not hold. Findings are advisory: they are
+// printed as a separate [Dead code] section and, when --sarif-output is set,
+// emitted at SARIF note level. They never flip the verdict (see report_result).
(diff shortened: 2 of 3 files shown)


lucasccordeiro (approved):

pmatos:
The two red tests on the `static llvm-22 DebugOpt / Build ESBMC (ubuntu-24.04)` job are **not caused by this PR**:

```
2731 - regression/esbmc/irep2_only_complex_compound_assign (Failed)
2732 - regression/esbmc/irep2_only_complex_compound_assign_fail (Failed)
```

Filed as #7312. Summary of the evidence:

- The abort is `assert_arith_2ops_consistency` (`src/irep2/irep2_expr.cpp:693`) firing during *Converting*, in `clang_c_adjust_irep2::lower_complex_compound_assignment` — the frontend adjust pass, before symex, coverage, or the solver. This PR touches only `report_dead_code` in `bmc.cpp` and two comments.
- Both tests were added by #7284 together with the lowering they cover, so they have never passed on `master`.
- The same two failures are present on #7298, #7294 and #7295 (this PR's predecessor), none of which touch irep2 or the C frontend.
- Root cause: `a *= 2.0f` on a `_Complex float` builds `mul2tc` at the 64-bit complex type with a 32-bit `float` RHS; the width check in the node constructor fires before `adjust_complex_arith` gets a chance to promote the scalar operand.

Only visible in assertions-enabled builds — `Release`/`RelWithDebInfo` set `-DNDEBUG`, which compiles the check out, which is why the other build jobs here are green.

The `Build and upload coverage to Codecov` failure (4s, before any build) is also pre-existing — same on #7298, #7294 and #7295.

Nothing to change on this PR as a result.

PR #7306, opened 2026-08-25
[goto-symex] Cap every allocation path at PTRDIFF_MAX

lucasccordeiro:
Above PTRDIFF_MAX an object's offset reads negative in the bounds checks,
pointer subtraction and the relational comparator, so one-past-the-end sorts
below the base. Cap every path that can size an object there, as glibc >= 2.30
does, closing R37 (malloc), R38 (alloca, realloc), R39 (a constant alloca) and
R40 (a VLA declaration).

A symbolic size is bounded by assumption where the path has no failure outcome
to report, and realloc joins the cap to its failure condition so C17 7.22.3.5
still holds. A constant request is reported instead: there the assumption would
be identically false and would prove the program rather than bound it.

R40 adds an assumption to every VLA program, so the benchexec run matters here
rather than the regression suite alone.

Refs #7228.


lucasccordeiro:
This PR (30s): 

```
  correct:           20898
    correct true:    14037
    correct false:    6861
  incorrect:            18
    incorrect true:      0
    incorrect false:    18
  unknown:           15687
  Score:             34647 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/32895985895.
```

Master (30s):

```
  correct:           20899
    correct true:    14034
    correct false:    6865
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15703

```

lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/32956947673

lucasccordeiro:
SV-COMP run for this PR: 

```
  correct:           20912
    correct true:    14040
    correct false:    6872
  incorrect:             1
    incorrect true:      0
    incorrect false:     1
  unknown:           15690
  Score:             34936 (max: 61277)

https://github.com/esbmc/esbmc/actions/runs/33011751321
```

Provenance: branch `fix/r37-ptrdiff-max-cap` @ `72eb72b9`, mode Full, timeout 30s, `ESBMC_OPTS` empty, SV-Benchmarks `main`, validation off. Compare only against a master run with the same timeout, options, and date.

Issue #7024, opened 2026-08-14
quantified variable assigned outside of quantification leads to PASS


MarkAA-uw:
In the program below, the quantified variable `idx_vfy` is assigned a value outside of the quantification.  This causes the property to pass, even though it shouldn't.

-mark

Command line: `esbmc --result-only rev-5.cpp`
Revision: `000dc204a6`

```
int main()
{
  int idx;
  int idx_vfy;

  idx_vfy = 17;
  
  __ESBMC_assert(
    __ESBMC_forall(
      &idx_vfy,
      !(0 <= idx_vfy && idx_vfy < 10) ||
      idx_vfy == 13
  ),
    "forall eq"
  );
  
}
```

PR #7317, opened 2026-08-25
[pointer-analysis] Fold symex-known index values before building references

### LINKED ISSUE #7311: Performance: ~60s symex for a single call to a callee with early return + trailing loop (minimized reproducer)
## Summary

Symex time explodes to ~60 s for a *single* call to a function whose body contains an early `return` followed later by a `while` loop, when the unwind bound is large — even when the inner loop is never executed. Removing either feature restores sub-second symex. Minimized from a real-world SHA-1 implementation (`sha1_update`'s early-return buffer path + `while (n >= 64)` flush loop, called in a loop from `sha1_final`'s padding), where this makes verification of a 170-line program intractable at any message size, including empty input.

## Environment

- ESBMC 8.3.0, 64-bit x86_64 Linux (with esbmclibc)
- Boolector 3.2.4 (also reproduced with cvc5 — solver-independent, time is in symex)

## Reproducer

```c
#include <stdint.h>
#include <stddef.h>

typedef struct { uint8_t buf[64]; size_t buflen; uint64_t len; } ctx;

static void update(ctx *c, const uint8_t *p, size_t n) {
    c->len += n;
    if (c->buflen) {
        size_t need = 64 - c->buflen;
        if (n < need) { c->buf[c->buflen] = *p; c->buflen += n; return; }
        c->buf[c->buflen] = *p;
        c->buflen = 0;
        p += need; n -= need;
    }
    while (n >= 64) { p += 64; n -= 64; }
    if (n) { c->buf[0] = *p; c->buflen = n; }
}

int main(void) {
    ctx c; c.buflen = 0; c.len = 0;
    uint8_t b = 0x80;
    update(&c, &b, 1);
    b = 0;
    while (c.buflen != 2) update(&c, &b, 1);   /* ONE iteration */
    return c.buf[0];
}
```

Command: `esbmc repro.c --boolector --unwind 260`

Result: `Symex completed in: 64.0s` — for one effective call of `update`, with the inner loop's guard always false.

## Unwind-bound sensitivity (same file, outer loop runs once)

| `--unwind` | symex time | SSA assignments |
|---|---|---|
| 10 | 0.030 s | 822 |
| 40 | 0.750 s | 9 222 |
| 260 | 64.0 s | (not reached promptly) |

Each extra unwinding of the (never-taken) inner loop adds ~280 assignments, and per-unwinding wall time grows superlinearly. The inner loop is dead (`n == 1`); its unwindings are pure guard overhead.

## Isolation matrix

All variants below differ only in the callee body; the outer structure is identical. Symex time:

| callee contains early `return` | callee contains inner `while` | symex |
|---|---|---|
| yes | yes | **64 s** |
| yes | no | 0.96 s |
| no | no | 0.66 s |
| no | yes | 0.23 s |

Both features are required; either one alone is cheap. Also verified *not* to be the cause (all fast, 0.01–0.14 s): the full SHA-1 80-round compression function on a `uint32_t w[80]` schedule array with symbolic content; byte-wise word assembly through `const uint8_t *`; a 55-iteration per-byte `memcpy` padding loop; `void *` parameters; `--overflow-check` / `--memory-leak-check` / `--ub-shift-check` instrumentation.

## Outer-loop scaling (reproducer as above, target K → K−1 iterations)

| K | symex |
|---|---|
| 2 | 64.0 s |
| 8 | 64.8 s |
| 24 | 70.9 s |
| 48 | 74.1 s (then timed out at 90 s total) |

Essentially flat: the cost is a fixed per-callee-instantiation blowup, so a real program that calls such a function in a loop (e.g. `sha1_final`'s padding loop, ~55 calls) cannot be verified within any reasonable budget.

## Suspected area

Interaction between multi-exit CFG merging (early `return`) and subsequent loop unwinding in the same callee: the guard/phi state carried into the inner loop's 260 dead unwindings appears to be re-processed at superlinear cost. With a single-exit callee the same unwindings are nearly free.

## Workaround for users

Restructure loop-called callees to single-exit form (`else` instead of early `return`), or keep `--unwind` small. Both restore sub-second symex on this pattern.

## Original context

Hit while verifying a from-scratch `sha1sum` clone (FIPS 180-1 streaming SHA-1, ~170 lines, no dynamic allocation) with `esbmc --k-induction-parallel --overflow-check --memory-leak-check --ub-shift-check`. Every configuration timed out in symex — k-induction, BMC with `--unwindset`-style bounds, and a harness with fully concrete message lengths down to `LEN=0` — always inside the unwinding of `sha1_block`/`sha1_update`. Happy to provide the full source and harnesses if useful as a larger regression test.

pmatos:
## Summary

- `dereference_expr_nonscalar` now renames and simplifies an index operand once it is free of dereferences, so an index whose value symex already knows (e.g. an array index stored in a struct field, `foo->buf[ctx->len]`) stays on the constant-offset path and produces a member/index reference.
- Previously such accesses degenerated to a whole-object `byte_extract`/`byte_update`, which constant propagation cannot see through: the assignment dropped the object's recorded constant, every later guard over the field became undecidable, and loops unwound to the bound — quadratic symex time (issue reproducer: 242 s symex, 345,537 SSA assignments at `--unwind 260` with Bitwuzla at current master).
- With this change the reproducer finishes symex in **0.001 s with 30 assignments** and `VERIFICATION SUCCESSFUL`.

Fixes #7311

## Test plan

- [x] Regression pair `regression/esbmc/github_7311` + `github_7311_fail` (`--unwind 800`): passing test expects `^VERIFICATION SUCCESSFUL$`; failing companion asserts on `c.buf[c.buflen - 1]` written through the same field-derived index, expects `^VERIFICATION FAILED$`. Both run in ~0.2 s when fixed; unfixed, the loop-bound claims pass vacuously but superlinear per-unwinding cost (2.3 s @ `--unwind 60`, 242 s @ 260, Bitwuzla) blows the 120 s CI per-test timeout.
- [x] `ctest -R "bitfield|github_7311"` — 42/42 pass (a broader variant of this change that renamed offsets in `build_reference_to` broke 10 bitfield/memset tests; the narrowed index-operand hook does not)
- [x] `ctest -R "^regression/esbmc/"` — no regressions vs. baseline: the only 11 failures (`github_536_boolector`, `github_1352*`, `math_log-rec*`, `multi_dimensional_array_1-success`, `overflow_*_fir*`, `02_eureka23`, `gcc_popcount_01`) fail identically on an unmodified build of the same base commit in this environment
- [x] Manual repro confirms the fix on HEAD (`2262116459`, == upstream master at time of testing) with Bitwuzla 0.9.0


regression/esbmc/github_7311/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unwind 260
+^VERIFICATION SUCCESSFUL$
```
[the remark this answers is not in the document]
pmatos: Addressed in 4e316c05f9: added regression/esbmc/github_7311_fail expecting ^VERIFICATION FAILED$ via assert(c.buf[c.buflen - 1] != 0) — buf[1] was written 0 through the same field-derived index, so a wrong-constant fold flips the verdict. Both tests verified: github_7311 SUCCESS (0.20s), github_7311_fail FAILED on exactly main.assertion.1 (0.17s).

src/pointer-analysis/dereference.cpp
```diff
@@ -508,6 +508,19 @@ expr2tc dereferencet::dereference_expr_nonscalar(
...
+    /* Now free of dereferences, fold in any value symex already knows for
+     * the index (e.g. an array index stored in a struct field): a constant
+     * index keeps this access on the constant-offset path, which produces a
+     * member/index reference. A symbolic index degenerates to a whole-object
+     * byte_extract/byte_update that constant propagation cannot see through:
+     * the assignment drops the object's recorded constant, later guards over
+     * it become undecidable and loops unwind to the bound (#7311). */
+    if (!is_constant_int2t(index.index))
+    {
+      dereference_callback.rename(index.index);
```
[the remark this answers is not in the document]
pmatos: Addressed in 426c011a83: dereference_callbackt::rename doc in dereference.h now states the default is a no-op while the symex override substitutes SSA names/recorded constants, invoked on index operands in dereference_expr_nonscalar; the README.md rough-edges entry says the same and keeps the build_reference_to benchmarking note (that hook remains disabled).

lucasccordeiro, line 599: This bypasses `goto_symex_statet::rename`'s `rename_quantified` path and substitutes quantifier-bound indices. Skip symbols bound by an enclosing quantifier.

Built this hunk on `2262116459` and measured: `__ESBMC_forall(&i, !(0 <= i && i < 4) || c->buf[i] == 0)` with `c->buf[3] == 1` and `int i = 2;` in scope goes `VERIFICATION FAILED` (violated property `all zero`, a `forall`) -> `VERIFICATION SUCCESSFUL`. The bound `i` is replaced by the outer value 2. A pointer-typed `p[i]` does not reproduce; the struct-field shape this PR targets does.

regression/esbmc/github_7311_fail/test.desc
```diff
@@ -0,0 +1,4 @@
+CORE
+main.c
+--unwind 800
+^VERIFICATION FAILED$
```
lucasccordeiro, line 4: `^VERIFICATION FAILED$` alone is satisfied by a bounds violation from a mis-folded index. Pin the property you verified: add `assertion c.buf\[c.buflen - 1\] != 0`.

regression/esbmc/github_7311/main.c
```diff
@@ -0,0 +1,60 @@
...
+  c.len = 0;
+  uint8_t b = 0x80;
+  update(&c, &b, 1);
+  b = 0;
+  /* Runs exactly one iteration; only verifiable quickly if the write above
+   * keeps c.buflen constant-propagated. */
+  while (c.buflen != 2)
+    update(&c, &b, 1);
+  return c.buf[0];
+}
```
lucasccordeiro, line 60: Neither test covers an index under a quantifier. Add a `__ESBMC_forall` case over a struct-field array reached through a pointer.

lucasccordeiro (changes requested):
PR carries no labels; AGENTS.md requires an area label plus `needs-svcomp-run` for pointer-analysis changes.

src/pointer-analysis/dereference.cpp
pmatos: Good catch — reproduced your forall case locally: SUCCESSFUL with the unguarded hook, VERIFICATION FAILED after this fix. In 2f114f3acb dereference_expr now tracks binder names while descending into forall/exists bodies (same shape as rename_quantified's bound set) and the index fold skips any index mentioning a bound symbol. New regression github_7311_forall pins the FAILED verdict for exactly your p->buf[i]-under-forall shape.

regression/esbmc/github_7311_fail/test.desc
pmatos: Done in 2f114f3acb: test.desc now also requires the violated-property line `assertion c\.buf\[c\.buflen - 1\] != 0`, so a bounds violation from a mis-folded index no longer satisfies the test.

regression/esbmc/github_7311/main.c
pmatos: Added in 2f114f3acb: regression/esbmc/github_7311_forall asserts __ESBMC_forall over p->buf[i] with a shadowing outer i = 2 and a violation at i == 3 — verifies the fold leaves quantifier-bound indices alone.

pmatos:
@lucasccordeiro ready for another look.

src/pointer-analysis/dereference.cpp
```diff
@@ -143,13 +143,67 @@ const expr2tc &dereferencet::get_symbol(const expr2tc &expr)
...
+  case expr2t::exists_id:
+  {
+    /* Descend with the bound variable tracked: the index-operand fold in
+     * dereference_expr_nonscalar must not substitute a quantifier-bound
+     * symbol with the like-named program variable's SSA value -- the rename
+     * hook has no quantifier context (rename_quantified handles the same
+     * concern at the renaming layer, #7024). */
+    const expr2tc &binder =
+      is_forall2t(expr) ? to_forall2t(expr).side_1 : to_exists2t(expr).side_1;
+    const irep_idt bound = quantifier_binder_name(binder);
```
lucasccordeiro: An indirect binder returns the pointer's name, not the bound variable's: `void *q = &i; __ESBMC_forall(q, ... p->buf[i] ...)` folds `i` to 2 and masks the violation (master: FAILED, here: SUCCESSFUL). Disable folding for the whole subtree unless the binder is `address_of(symbol)`.

src/pointer-analysis/dereference.cpp
```diff
@@ -143,13 +143,67 @@ const expr2tc &dereferencet::get_symbol(const expr2tc &expr)
 
 static expr2tc distribute_steps_over_if(const expr2tc &e);
 
+/// Base name of the variable a forall2t/exists2t binds (mirrors
+/// quantifier_bound_name in goto_symex_state.cpp), or an empty id when the
+/// binder is not the (typecast of) address_of(symbol) shape the solver
+/// expects.
+static irep_idt quantifier_binder_name(const expr2tc &binder)
```
lucasccordeiro: Verbatim copy of `quantifier_bound_name` in goto_symex_state.cpp:419; the two must stay in sync for #7024. Expose the existing one in a header and call it.

src/pointer-analysis/dereference.cpp
```diff
@@ -143,13 +143,67 @@ const expr2tc &dereferencet::get_symbol(const expr2tc &expr)
...
+     * dereference_expr_nonscalar must not substitute a quantifier-bound
+     * symbol with the like-named program variable's SSA value -- the rename
+     * hook has no quantifier context (rename_quantified handles the same
+     * concern at the renaming layer, #7024). */
+    const expr2tc &binder =
+      is_forall2t(expr) ? to_forall2t(expr).side_1 : to_exists2t(expr).side_1;
+    const irep_idt bound = quantifier_binder_name(binder);
+    const bool tracked =
+      bound != irep_idt() && quantifier_bound_vars.insert(bound).second;
+    expr->Foreach_operand(
```
lucasccordeiro: This recursion is byte-identical to the `default:` arm below, so future changes there must be mirrored here. Scope the bound name with an RAII guard and share one recursion.

lucasccordeiro (changes requested):

lucasccordeiro:
@pmatos: Could you also fix the C/C++ coding style CI?

pmatos:
> @pmatos: Could you also fix the C/C++ coding style CI?

Sure but you're using clang-format-11... that's errr, 5 years old. Shall we consider updating to a more recent one? 

1. Choose a stable version of clang-format
2. do a whole tree reformat
3. add the sha to git blame ignore revs
4. update CI

Done that before for FEX here: https://github.com/FEX-Emu/FEX/commit/340630a937ea7c1f2c37bc587bb282a2e6b90679

src/pointer-analysis/dereference.cpp
pmatos: Fixed in ff4e5203ca. `dereferencet::quantifier_scopet` now takes the name from `quantifier_direct_bound_name`, which is empty unless the binder is (a typecast of) `address_of(symbol)`. An empty name bumps an `opaque_binders` counter instead of recording a name, and `may_fold_index` returns false while that counter is non-zero — so the whole subtree is out of the fold, not just indices matching some name.

Your `void *q = &i;` repro is pinned as `regression/esbmc/github_7311_forall_indirect`: `VERIFICATION FAILED` with the violated property line. Mutation-checked — dropping the `opaque_binders == 0` conjunct flips it back to `VERIFICATION SUCCESSFUL`.

One caveat worth recording: the verdict there also needs `rename_quantified` to recover the bound name, which it does because `q` is constant-propagated to `&i` before renaming. A binder whose value is not known (`q = nondet ? &i : &j`) hits a separate pre-existing gap in that layer — it segfaults during VCC encoding on master too, with or without any dereference in the body, so it is not something this PR introduces or pins. I've noted that in the test's comment; happy to open a separate issue for it.

src/pointer-analysis/dereference.cpp
pmatos: Fixed in 62edd7e0e0. `quantifier_bound_name` moved out of `goto_symex_state.cpp` to `src/irep2/irep2_utils.{h,cpp}` — body unchanged, only `static` dropped — and both `rename_quantified` and `dereference.cpp` now call it, so #7024's shape is defined once.

Alongside it there is a strict companion, `quantifier_direct_bound_name`, empty unless the binder is `address_of(symbol)`; they share the typecast-stripping, so the shape logic is not written twice either. That is what the dereference side needs for the indirect-binder case above.

Two small things that fell out: `rename_quantified` now skips an empty name rather than inserting it into the bound set (no symbol carries the empty name, so the set's meaning is unchanged), and there is no `quantifier_binder_name` left anywhere in `src/`.

src/pointer-analysis/dereference.cpp
pmatos: Fixed in ff4e5203ca. The bound name and the opaque-binder count are now scoped by an RAII `dereferencet::quantifier_scopet`, and the switch reads `case forall_id: case exists_id: default:` over a single recursion — the duplicated `Foreach_operand` block is gone. Constructing the guard over a non-quantifier does nothing, so the shared arm needs no branch.

The guard is non-copyable (a copy would run the destructor twice: `opaque_binders` underflow, or `multiset::erase` on `end()`), and it erases through the iterator rather than the key so an inner scope rebinding the same name cannot drop the outer scope's entry.

pmatos:
@lucasccordeiro all three points addressed, plus the labels and the code-style CI.

- **Indirect binder** (ff4e5203ca) — folding is now disabled for a quantifier's whole body unless the binder is `address_of(symbol)`; your `void *q = &i;` case is pinned as `github_7311_forall_indirect`.
- **Duplicated helper** (62edd7e0e0) — `quantifier_bound_name` moved to `irep2_utils`; both `rename_quantified` and `dereference.cpp` call it.
- **Duplicated recursion** (ff4e5203ca) — RAII `quantifier_scopet` + one shared recursion.
- **Labels** — `pointer-analysis`, `goto-symex`, `needs-svcomp-run`.
- **Code style** — one hunk in `mentions_bound_var` from 2f114f3acb was not clang-format-11 clean; `git-clang-format` against master is clean now.

Tests, on top of the three quantifier tests being mutation-checked (each flips when the exclusion is removed):

- `esbmc/` 1909/1920 — the same 11 failures as an unmodified build of the same base in this environment (`github_536_boolector`, `github_1352*`, `math_log-rec*`, `multi_dimensional_array_1-success`, `overflow_*_fir*`, `02_eureka23`, `gcc_popcount_01`)
- `bitwuzla/` + `z3/` + `function_contract/` + `k-induction/` + `loop-invariants/` — 745/745
- unit — 708/708; `bitfield|memset` — 61/61

lucasccordeiro:
> Sure but you're using clang-format-11... that's errr, 5 years old. Shall we consider updating to a more recent one?

Definitely! Can you please submit a PR to update it, @pmatos?

lucasccordeiro:
This PR (30s): https://github.com/esbmc/esbmc/actions/runs/32985549434.

pmatos:
> This PR (30s): https://github.com/esbmc/esbmc/actions/runs/32985549434.

Seems to be stuck. What's going on?

